PR #126, opened 2026-07-21
fix(airvpn): killswitch polices only the physical uplink (fixes k3s outage)

ZacxDev:
## What broke
Enabling the host AirVPN tunnel **CrashLooped the entire k3s cluster** — CoreDNS, Traefik (→ 502s), all 4 Flux controllers, metrics-server, devpods. External `kubectl` worked fine, which made it look API-side.

## Root cause
The fail-closed killswitch's `output` chain (`scripts/airvpn-updown`) had an **allow-list of internal networks**, and the **pod CIDR `10.244.0.0/16` wasn't in it**. Flow: a pod hits the API service `10.245.0.1:443` → kube-proxy DNATs to the host apiserver `192.168.50.250:6443` → the apiserver (host-local process) replies **to the pod IP** → egresses `cni0` → hits the killswitch → not in the allow-list → **dropped**. External kubectl survived because its replies go to `192.168.50.x` (allowed).

Same class of bug as the nebula-overlay drop fixed after the #118 audit: an allow-list that must enumerate every internal network is fragile on a host that runs k3s **and** is reached over nebula.

## Durable fix — flip the model
Police **only the physical uplink**: `oifname != <phys> accept` first, so loopback, the airvpn tunnel, the nebula overlay (nebula.mesh) and the k3s CNI (cni0/flannel.1 → pod & service CIDRs) are always allowed; only uplink egress is filtered (wg fwmark, LAN, nebula transport + lighthouse, endpoint, gateway → allow; else drop). The only un-tunnelled leak path IS the physical uplink, so this is **leak-equivalent** to the old default-drop **and immune to any future internal network** with no new rule.

## Also: orphan-teardown fix (`scripts/airvpn-sudo`)
`down` only ran `wg-quick down`, which errors before PostDown when the iface is already gone → killswitch table left armed with no tunnel. Now tears it down explicitly (idempotent). Latent lockout risk on unclean teardown/reboot (not the k3s trigger — the table came down cleanly here).

## Verification
- `bash -n` both scripts; `nix-instantiate --parse` the module.
- nft chain traced offline: apiserver→pod ✅ accept · nebula overlay ✅ accept · tunnel-up ✅ tunnelled · **leak out uplink ✅ dropped**.
- ⚠️ **NOT re-tested against a live Connect.** Re-test must be **LAN-first** (`192.168.50.x`) with external `kubectl` watching cluster health, and the orphan-safe `down` ready.

[...]

PR #118, opened 2026-07-18
fix(bar): minimal icon-only AirVPN pill + stage Phase-2 host apply

ZacxDev:
## Why
`/bar` report: **the AirVPN control doesn't work**, and its display should be a minimal icon/text button to match the rest of the bar.

Root cause of "doesn't work": the host AirVPN block has been non-functional since 2026-07-12 because **Phase 2 was never applied** — no `/etc/wireguard/airvpn.conf`, no `~/.config/bar/airvpn.env`, and `airvpn-sudo` still prompts for a password (the NOPASSWD sudoers rule from `nix/system/airvpn-host.nix` was never rebuilt in). So `Connect` → `sudo airvpn-sudo up` hit a password prompt non-interactively and failed. The menu/render code itself was fine.

## What
1. **Minimal icon-button render** (`scripts/i3status-airvpn`):
   - off → dim, **icon-only** (was `VPN off`)
   - stale/error → soft-yellow, **icon-only** (was `VPN?`)
   - up verified → `⟨icon⟩ CA` · unknown → `CA?` (was `AirVPN CA ·1m`)
   - leak → `LEAK` (red) · forwarded-port down → `CA!` (yellow)
   - Drops the `AirVPN …` prose + handshake age from the pill (that detail lives in the menu header + `airvpn-detail` TUI). The off pill stays **rendered, not hidden**, so it remains the left-click Connect target.
   - Removed the now-dead `_fmt_age` helper; render unit tests updated.
2. **Staged Phase-2 apply** (`nix/system/apply-airvpn-host.sh`): idempotent, backup-taking `sudo bash` that locks the secret conf `0600`, appends the killswitch `PostUp/PreDown` hooks, installs `airvpn-sudo` + `airvpn-updown` to the stable sudoers path, wires the module import into `configuration.nix`, and rebuilds. **DEFAULT-OFF** — does not bring the tunnel up. `airvpn-host.nix`'s Phase-2 header now points at it.

## Bring it live (your steps — I can't `sudo nixos-rebuild`)
1. Generate a NEW-device AirVPN WireGuard config → save to `/etc/wireguard/airvpn.conf`.
2. `sudo bash nix/system/apply-airvpn-host.sh`
3. Left-click the AirVPN bar pill → **Connect**, then → **Verify exit IP / leak**.

## Verification
- ✅ Render fixtures (all 5 states) + **41 passing** unit tests (`test_airvpn_menu.py`)
[...]

PR #128, opened 2026-07-21
fix(airvpn): police all physical NICs + fail-closed killswitch + active-pill display

ZacxDev:
Closes the two 🟡s from the #126 audit, plus a display bug found while verifying the live tunnel.

## 🟡 → fixed: killswitch (`scripts/airvpn-updown`)
**Second-uplink leak.** `oifname != "$PHYS" accept` accepted egress on any *other* physical NIC. This workbench actually has **three** (`eth1` + `wlp15s0` + `enp18s0u2u2c2`), and the laptop is multi-homed — so this was a live latent leak, not hypothetical. Now polices **all physical NICs**, derived from `/sys/class/net/*/device` (hardware presence, not the default route) → `oifname != { <all phys> } accept`. Also covers a NIC that gains a default route *after* PostUp.

**Fail-open abort.** Empty `GW`/`PHYS` did `exit 0` *before* loading nft → tunnel up with **no killswitch** (silent leak). Now the killswitch **always arms**; only the bypass *routes* are skipped (loud warning) when there's no gateway. If **no** physical NIC is found (near-impossible with a live tunnel), it **fails closed** (allow internal ifaces only, drop the rest).

## Bonus: active-pill display bug (`scripts/i3status-airvpn`)
While verifying the live tunnel the active pill rendered **`???`** — a hostname endpoint (`ca3.vpn.airdns.org`) doesn't map to a manifest server, so the server `country_code` is null. `_cc()` now falls back to the observed ipinfo `exit_country` → shows **`CA?`**. Confirmed live (screenshot). +1 render test.

**Display now:** inactive → icon-only dim · active → `⟨icon⟩ CA?` (or `CA` when verified) · leak → `LEAK` red · fwd-port down → `CA!` yellow · stale → icon-only soft-yellow.

## Verification
- `bash -n`; `phys_ifaces()` returns the 3 real NICs live; nft chain rendered offline for **3-NIC / single-NIC / empty(fail-closed)**; **42** pytest pass.
- Display fix deployed via `home-manager switch` and **confirmed on the live bar** (`CA?`).
- ⚠️ The **killswitch change is NOT live** — it needs re-install to `/etc/nixos/i3blocks-scripts/` + a reconnect. That reconnect must be **off-LAN nebula, `kubectl` watching, orphan-safe `down` staged** (a LAN test can't reach the direct-punch lockout mode). I can't run that.

Follows #118 / #122 / #126. Bar skill updated (per-host) with the new display + killswitch model + live Phase-2 status.

### DIFF
--- a/scripts/airvpn-updown
+++ b/scripts/airvpn-updown
@@ -71,17 +71,41 @@ KS_TABLE="airvpn_ks"
 
 log() { logger -t airvpn-updown "$*" 2>/dev/null || echo "airvpn-updown: $*" >&2; }
 
+# Blanket FAIL-CLOSED ruleset — installed only if the main killswitch nft load
+# fails, so an nft error can never leave the uplink UNFILTERED (fail-open). Minimal
+# on purpose (no dynamic PHYS_SET / fwmark / EP substitution that could be what
+# broke the main load): allow internal ifaces + nebula (skuid) + LAN + lighthouse,
+# drop the rest. Keeps the operator reachable while still stopping an internet leak.
+arm_failclosed() {
+    nft add table inet "$KS_TABLE" 2>/dev/null || true
+    nft flush table inet "$KS_TABLE" 2>/dev/null || true
+    nft -f - <<FB || log "FALLBACK nft load ALSO FAILED — uplink is NOT filtered!"
+table inet ${KS_TABLE} {
+    chain out {
+        type filter hook output priority 0; policy accept;
+        oifname { "lo", "${IFACE}", "nebula.mesh", "cni0", "flannel.1", "docker0" } accept
+        meta skuid "${NEBULA_USER}" accept
+        ip daddr ${LAN_SUBNET} accept
+        ip daddr ${NEBULA_LIGHTHOUSE} accept
+        drop
+    }
+}
+FB
+}
+
 # Original default gateway + physical iface (the default route NOT via the tunnel).
 # Used for the split-tunnel bypass ROUTES (they need one specific gateway).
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Audit results (`/audit-pr 128`) + hardening applied

**Verdict: safe to re-enable off-LAN, no deploy-blockers.** The audit verified live that all 3 physical NICs land in the policed set, the fail-closed branch keeps the operator reachable, the display fallback can't mask a leak, and the empty-GW/EP/`down` paths are all guarded. It found two 🟡s worth closing before calling the model durable — done in `6ac38bd`:

- **🟡 bridge/bond/VLAN uplink blind spot** — `/sys/.../device` is a *hardware* test; a `br0`/`bond0`/`eth1.100` uplink carries internet but has no device symlink → `oifname != {hardware NICs}` would trust (leak) it. Fixed: the policed set now **unions the hardware NICs with the live default-route egress iface**. No change on this host (default route is raw `eth1`); verified a simulated `br0` default route is now included.
- **🟡 nft-load fail-open** — on `nft -f` error the pre-flushed table was left `accept` = unfiltered. Fixed: install a **blanket fail-closed fallback** (internal ifaces + `skuid nebula-mesh` + LAN + lighthouse → drop) so an nft error can't leave the uplink open.
- 🟢 removed the dead `GW` read in `down)`.

Audit 🟢 items left as-is (informational): broadcast-DHCP on a policed NIC is dropped (desirable — stops a second uplink leaking), forwarded/pod/container traffic is unpoliced by design (`hook output` only), stale endpoint /32 on `down` (harmless).

Still gated: **not live until re-install + off-LAN reconnect** (kubectl watching, orphan-safe `down` staged).

PR #135, opened 2026-07-22
feat(initiatives): Phase 1 — sync initiative-scan into mailbox Postgres

ZacxDev:
## What & why

**Phase 1** of the "initiatives consolidation" feature: sync the on-demand
`initiative-scan.py` report into the homelab `mailbox` Postgres on a timer, so
later apps (a live **viewer** + a **router**) can query a durable, live
initiative store instead of re-running the expensive scan. **This PR is the sync
ONLY** — no viewer, no router.

## How it works

- **Data source is the contract.** `sync.py` shells out to
  `initiative-scan.py --days N --json` (WITHOUT `--tmux` — the ephemeral tmux
  overlay is excluded) and parses stdout. It never imports the scan's internals
  or re-derives anything; the scan degrades gracefully with no ClickHouse creds
  and so does the sync.
- **Pure transform, testable without infra.** `report_to_rows(report, host)` is a
  pure function (epoch floats → UTC `timestamptz`, authored `date` str → `date`,
  JSONB for `open_prs`/`open_investigations`/`docs`, host tagging, momentum
  passthrough). All I/O (the subprocess scan + the DB write) is separate.
- **DB reuse.** Reuses `scripts/mail-actions/_db.py`'s port-forward + psycopg2 +
  DSN-from-secret plumbing, loaded by **explicit importlib path** (NOT `sys.path`
  — avoids the `llm.py` shadow trap flagged in CLAUDE.md; mirrors
  `repo-cos/feedback.py`).
- **Own schema, self-migrating.** Tables live in a new `initiatives` schema inside
  the `mailbox` DB (so a future router can natively JOIN against `mail_actions`).
  DDL is `CREATE ... IF NOT EXISTS` / `CREATE OR REPLACE VIEW` on every run —
  additive-only, no manual prod DDL step.
- **Append-only snapshots.** `initiatives.snapshots` (one row per run) +
  `initiatives.initiative_snapshot` (one row per initiative per run) +
  `initiatives.current` (DISTINCT ON latest per `(repo, slug)`, tolerates a
  partially-failed snapshot).

## Timer (wired, NOT enabled/deployed)

Workbench-only (`serverMode`, same LAN-access rationale as `mail-actions-archive`)
`systemd --user` service + ~15min timer in `nix/home.nix`. **Not switched/deployed
here** — Zach eyeballs `--dry-run` output first.

## Verification (done honestly)

- **Unit tests:** 23 tests over the pure transform (epoch/null/jsonb/host/
  momentum/empty + telemetry-off) — pass. Registered in the hermetic set.
- **Live end-to-end `--dry-run` WITH telemetry creds:** ran against the real scan
  → 23 rows, `telemetry_available=True`. Verified telemetry epochs convert to UTC
  `timestamptz` (`telem_last`/`last_touch`), git `commits`/`merged_prs` populate,
  `doc_date` parses, `docs` JSONB, null `next_step`. Sample row (task-spec-drafter):
  `commits=2, merged_prs=1, telem_events=10, telem_last=2026-07-21T20:34:49...+00:00`.
- **NOT verified:** an actual DB write to prod (deliberately — `--dry-run` only,
  per the task). `nix flake check`/`home-manager switch` not run (nix file parses
  via `nix-instantiate --parse`).

## OPEN follow-ups (deferred, by design)

1. **CLICKHOUSE_\* cred provisioning for the timer env** — not injected yet, so the
   timer would run **telemetry-off** (still writes a useful handoff+git+session
   snapshot; momentum/ev degrade). Separate task.
2. **Dual-host merge** — Phase 1 is workbench-only; a laptop/second-host snapshot
   merge is Phase 2+.
3. **Enable/deploy the timer + first real write** — pending Zach's `--dry-run`
   eyeball.

## Schema DDL (for the eyeball)

```sql
CREATE SCHEMA IF NOT EXISTS initiatives;

CREATE TABLE IF NOT EXISTS initiatives.snapshots (
    id                   serial PRIMARY KEY,
    captured_at          timestamptz DEFAULT now(),
    host                 text,
    days_window          int,
    telemetry_available  boolean
);

CREATE TABLE IF NOT EXISTS initiatives.initiative_snapshot (
    id                   serial PRIMARY KEY,
    snapshot_id          int REFERENCES initiatives.snapshots(id),
    host                 text,
    repo                 text,
    slug                 text,
    title                text,
    doc_date             date,
    momentum             text,
    last_touch           timestamptz,
    next_step            text,
    commits              int,
    commits_unknown      boolean,
    merged_prs           int,
    open_prs             jsonb,
    session_count        int,
    telem_events         int,
    telem_last           timestamptz,
    current_doc          text,
    open_investigations  jsonb,
    docs                 jsonb
);

CREATE OR REPLACE VIEW initiatives.current AS
SELECT DISTINCT ON (i.repo, i.slug)
       i.*, s.captured_at
  FROM initiatives.initiative_snapshot i
  JOIN initiatives.snapshots s ON s.id = i.snapshot_id
 ORDER BY i.repo, i.slug, s.captured_at DESC;
```

## Questions for Zach before Phase 2

- **`id serial PRIMARY KEY` on `initiative_snapshot`** — added beyond the spec (a
  stable per-row PK for the future router / potential FKs). OK, or drop it?
- **`catchall` not stored** — the report-level trunk `catchall` is intentionally
  omitted (doesn't fit the per-initiative row cleanly; KISS). Want it in Phase 2
  (its own small table), or leave it report-only?
- **`_db.py` reuse** — imported by importlib path today. If Phase 2 grows more
  shared DB helpers, worth promoting the port-forward/DSN plumbing into a small
  shared module instead of importing `mail-actions/_db.py` sideways?

### DIFF
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -724,6 +724,68 @@ in
     };
   };
 
+  # Initiatives consolidation (PHASE 1) — periodic sync of the on-demand
+  # initiative-scan into the homelab `mailbox` Postgres (initiatives schema), so
+  # later apps (a live viewer + a router) query a durable, live store instead of
+  # re-running the expensive scan. The wrapper (scripts/initiatives/run-sync.sh)
+  # shells out to initiative-scan.py --json and writes one append-only snapshot via
+  # a kubectl port-forward — SAME cluster-access shape as mail-actions-archive.
+  #
+  # WORKBENCH-ONLY (gated on serverMode), identical rationale to the archiver: the
+  # homelab kubeconfig points at 192.168.50.94:6443 (direct LAN, no proxy), which
+  # only this host has; the laptop is nebula-only and its run would just fail noisily.
+  #
+  # ⚠ OPEN FOLLOW-UP: CLICKHOUSE_* creds are deliberately NOT injected here yet, so
+  # the timer currently runs the scan TELEMETRY-OFF (still writes a useful handoff+
+  # git+session snapshot; momentum/ev degrade). Provisioning those creds into the
+  # unit env is a separate task — see the PR. Do NOT enable/deploy this before Zach
+  # has eyeballed `sync.py --dry-run` output.
+  #
+  # Minimal user-unit env, so PATH is explicit: nix (nix-shell) + git + gh (the
+  # scan's branch/PR reads) + kubectl (the port-forward) + coreutils/sed/grep, and
+  # NIX_PATH so `nix-shell -p` resolves <nixpkgs>. The wrapper's nix-shell adds
+  # psycopg2 (the DB write) + requests (the scan's ClickHouse read).
+  systemd.user.services.initiatives-sync = lib.mkIf serverMode {
+    Unit = {
+      Description = "Initiatives sync — initiative-scan → homelab mailbox Postgres (initiatives schema)";
+      After = [ "network-online.target" ];
+      Wants = [ "network-online.target" ];
+      OnFailure = [ "notify-failure@%n.service" ];
+    };
+    Service = {
+      Type = "oneshot";
+      # Hard ceiling so a half-hung kubectl / scan can't wedge the timer; the
+      # cgroup is killed and the timer re-arms on the next OnUnitActiveSec.
+      TimeoutStartSec = 300;
+      Environment = [
+        "PATH=${lib.makeBinPath [ pkgs.nix pkgs.git pkgs.gh pkgs.kubectl pkgs.bash pkgs.coreutils pkgs.gnused pkgs.gnugrep ]}"
+        "NIX_PATH=nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixos"
+        "KUBECONFIG=%h/workspace/homelab-talos/homelab-kubeconfig"
+        "HOME=%h"
+      ];
+      ExecStart = "${pkgs.bash}/bin/bash %h/workspace/devrc/scripts/initiatives/run-sync.sh";
+      # Re-run the unit when the wrapper changes (cf. X-Restart-Triggers above).
+      X-Restart-Triggers = [ "${../scripts/initiatives/run-sync.sh}" ];
+    };
+  };
+
+  # Timer: fire the sync ~every 15 min. OnUnitActiveSec re-arms after each run so a
+  # slow sync never overlaps itself; OnStartupSec gives one prompt run after login.
+  # (No Persistent — it only applies to OnCalendar timers, not monotonic ones.)
+  # Workbench-only, same as its service (serverMode).
+  systemd.user.timers.initiatives-sync = lib.mkIf serverMode {
+    Unit = {
+      Description = "Periodic timer for the initiatives → Postgres sync";
+    };
+    Timer = {
+      OnStartupSec = "2min";
+      OnUnitActiveSec = "15min";
+    };
+    Install = {
+      WantedBy = [ "timers.target" ];
+    };
+  };
+
   # Repo chief-of-staff — WEEKLY: deterministic scan of Zach's repos for improvement
   # signals (TODO/FIXME, skipped tests, `latest` tags, churn, large files) → cheap LLM
   # synthesis (OpenRouter) → ranked proposal digest EMAILED. The "agents bring me ideas"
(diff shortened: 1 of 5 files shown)


ZacxDev:
## Audit fixes applied (commits 3b8749f + 1ad915b)

Addressed the 2 blockers + should-fixes/nits from the adversarial audit. No prod write / no `home-manager switch` done — that's the supervised step.

**🔴 Blockers**
1. **Timer can't auto-activate on a deploy** — added an off-by-default `enableInitiativesSync = false;` master switch in `nix/home.nix` (`let`-block). The timer is now double-gated `lib.mkIf (serverMode && enableInitiativesSync)`, so with the flag false the timer unit isn't emitted at all — no `ship.sh`/switch can wire it into `timers.target`. The **service** stays defined (startable by hand). Comment says: flip to true after the first supervised live write validates the path.
2. (Live-write validation is yours — that's why #1 matters.)

**🟡 Should-fix**
3. **Indexes** added to `ensure_schema` DDL: `(repo, slug, snapshot_id)`, `(snapshot_id)`, and `snapshots(captured_at)`.
4. **90-day retention** — `prune_old_snapshots()` runs each write; `DELETE FROM initiatives.snapshots WHERE captured_at < now() - make_interval(days => 90)`, child rows cascade (FK is now `ON DELETE CASCADE`).
5. **Advisory-locked DDL** — `SELECT pg_advisory_xact_lock(...)` wraps schema setup; **view guard** only (re)creates `initiatives.current` when absent or its `COMMENT ON VIEW` version marker differs (skips ACCESS EXCLUSIVE in steady state).
6. **Cadence** 15min → **hourly** (`OnUnitActiveSec = "1h"`).

**🟢 Nits**
7. Explicit `ACTIVITY_HOST=workbench` in the unit env.
8. Fixed the misleading fixture-epoch comment (`1768305600.0` → `1783944000.0`).
9. `run_scan` now has a **240s subprocess timeout** (+ a `TimeoutExpired` handler in `main`).

### Updated `ensure_schema` DDL (indexes + view guard)
```sql
-- (inside a pg_advisory_xact_lock(0x14171717) transaction)
CREATE SCHEMA IF NOT EXISTS initiatives;

CREATE TABLE IF NOT EXISTS initiatives.snapshots (
    id serial PRIMARY KEY, captured_at timestamptz DEFAULT now(),
    host text, days_window int, telemetry_available boolean );

CREATE TABLE IF NOT EXISTS initiatives.initiative_snapshot (
    id serial PRIMARY KEY,
    snapshot_id int REFERENCES initiatives.snapshots(id) ON DELETE CASCADE,
    host text, repo text, slug text, title text, doc_date date, momentum text,
    last_touch timestamptz, next_step text, commits int, commits_unknown boolean,
    merged_prs int, open_prs jsonb, session_count int, telem_events int,
    telem_last timestamptz, current_doc text, open_investigations jsonb, docs jsonb );

CREATE INDEX IF NOT EXISTS initiative_snapshot_repo_slug_snap_idx
    ON initiatives.initiative_snapshot (repo, slug, snapshot_id);
CREATE INDEX IF NOT EXISTS initiative_snapshot_snapshot_id_idx
    ON initiatives.initiative_snapshot (snapshot_id);
CREATE INDEX IF NOT EXISTS snapshots_captured_at_idx
    ON initiatives.snapshots (captured_at);

-- View: (re)created only when absent OR its COMMENT version marker != 'initiatives-sync view v1'
CREATE OR REPLACE VIEW initiatives.current AS
SELECT DISTINCT ON (i.repo, i.slug) i.*, s.captured_at
  FROM initiatives.initiative_snapshot i
  JOIN initiatives.snapshots s ON s.id = i.snapshot_id
 ORDER BY i.repo, i.slug, s.captured_at DESC;
COMMENT ON VIEW initiatives.current IS 'initiatives-sync view v1';
```

### Tests
`scripts/run-tests.sh --set hermetic` is **green** (full suite PASS). The `initiatives` suite is now **31 tests** — the pure transform plus new fake-connection SQL assertions for the advisory lock, indexes, cascade FK, view guard (absent / version-match / version-differs), the retention DELETE, and the snapshot/JSONB insert shape. Also re-verified the live `--dry-run` (telemetry on: 23 rows; telemetry off: 23 rows, flagged) still works after the restructure.

PR #184, opened 2026-07-29
fix(collector): stop expected lifecycle exits from firing phantom failure toasts

ZacxDev:
## Problem

Three collector units carry `OnFailure = notify-failure@%n.service` but exit non-zero on events that are entirely normal, so routine session activity minted "service failed" dunst notifications. This started as chasing one such toast; the sweep found the other two.

30 days of `journalctl --user` shows **19 failures across these three units — every one benign**:

| Unit | Failures | Cause |
|---|---|---|
| `i3-source` | 16 | `return 1` whenever i3's IPC connection closed — i.e. every i3 restart / config reload |
| `keylog` | 1 | uncaught `Xlib.error.ConnectionClosedError` on X teardown |
| `claude-activity-source` | 2 | SIGTERM'd mid-scan at logout → systemd records `Failed with result 'signal'` |

For `i3-source` specifically, `Restart = "always"` already restarts the unit on *any* exit status, so the non-zero return bought nothing except the toast.

## Fix

Each change narrows to the expected event only — **a genuine crash still exits non-zero and still notifies**:

- **`i3source.py`** — clean return from `i3.main()` now exits 0. The `except Exception` branch above it still returns 1.
- **`keylog.py`** — catch connection-closed / socket errors out of the RECORD loop; any other exception re-raises.
- **`nix/home.nix`** — `SuccessExitStatus = "TERM"` on `claude-activity-source`. The tailers checkpoint state, so an interrupted run resumes next tick rather than restarting.

Audited all 14 user units and every script under `scripts/collector/`. The remaining collectors (`collector.py`, `tailer.py`, `session-tailer.py`, `receiver.py`) already return 0 on every expected path. `homelab-kube-tunnel` has 20 real failures but no `OnFailure`, so it never toasted — untouched here.

## Verification

- **`claude-activity-source`: reproduced.** Started the unit, SIGTERM'd it mid-scan → `Result=success`, `ExecMainStatus=0`, and zero `Triggering OnFailure=` lines. That exact path produced `Failed with result 'signal'` before.
- **`i3-source` / `keylog`: deployed but not reproduced.** Both need a real i3 restart / X teardown to observe; they'll prove out at the next reload or logout.
- Tests pass: 73 passed / 1 skipped (keylog), 22 passed (i3).
- `home-manager switch` applied cleanly on workbench.

## Note (not in this PR)

The journal also shows `emit: line 67: hostname: command not found` — no collector unit has `hostname` on its PATH. **Cosmetic only:** `collector.py:191` overwrites `event["host"]` from `ACTIVITY_HOST` at ship time, so host attribution is unaffected. Adding `pkgs.inetutils` would silence it.

### DIFF
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -545,9 +545,18 @@ in
       TimeoutStartSec = 600;
       # Session teardown (logout/reboot) SIGTERMs this oneshot mid-scan, which
       # systemd records as Failed with result 'signal' → a phantom OnFailure
-      # toast for a perfectly normal shutdown. The tailers checkpoint state, so
-      # an interrupted run just resumes next tick. Treat SIGTERM as success; a
-      # real crash still exits non-zero and still notifies.
+      # toast for a perfectly normal shutdown. Treat SIGTERM as success; a real
+      # crash still exits non-zero and still notifies (an OOM kill is SIGKILL,
+      # not SIGTERM, so it is unaffected).
+      #
+      # CAVEAT — the two tailers are NOT equally safe to interrupt.
+      # session-tailer.py checkpoints incrementally, so a SIGTERM'd run resumes.
+      # tailer.py does NOT: it saves state once, after its whole loop, so an
+      # interrupted run re-emits everything it already emitted. Those are
+      # source=claude kind=prompt|command rows, which (unlike kind=session-summary)
+      # have no argMax dedupe on read, so the duplicates inflate counts. That
+      # hazard predates this setting, but suppressing the toast removes the only
+      # signal it just happened — fix by checkpointing tailer.py incrementally.
       SuccessExitStatus = "TERM";
       Environment = [
         "PATH=${lib.makeBinPath [ pkgs.python312 pkgs.coreutils pkgs.bash ]}"
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -543,6 +543,12 @@ in
       # room — session-tailer.py also now checkpoints its state incrementally so
       # an interrupted run still resumes rather than restarts.
       TimeoutStartSec = 600;
+      # Session teardown (logout/reboot) SIGTERMs this oneshot mid-scan, which
+      # systemd records as Failed with result 'signal' → a phantom OnFailure
+      # toast for a perfectly normal shutdown. The tailers checkpoint state, so
+      # an interrupted run just resumes next tick. Treat SIGTERM as success; a
+      # real crash still exits non-zero and still notifies.
+      SuccessExitStatus = "TERM";
       Environment = [
         "PATH=${lib.makeBinPath [ pkgs.python312 pkgs.coreutils pkgs.bash ]}"
       ];
--- a/scripts/collector/i3/i3source.py
+++ b/scripts/collector/i3/i3source.py
@@ -230,10 +230,12 @@ def main(argv=None) -> int:
     except Exception as exc:
         print(f"i3source: event loop ended: {exc}", file=sys.stderr, flush=True)
         return 1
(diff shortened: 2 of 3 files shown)


ZacxDev:
## Audit follow-up applied (e1626e3)

Ran an adversarial read-only audit on this branch. Verdict was *safe to merge* with no 🔴, but three 🟡 findings were worth closing — suppressing a failure signal has to be done narrowly or it just trades a noisy alarm for no alarm.

**The concern I'd flagged as most serious did not hold.** I worried the `OSError` in the keylog catch could swallow a real spool-write failure (disk full, permissions) and silently kill telemetry. It can't: `spool_emit.py:79-85` already wraps every write in `try/except OSError: pass`, so those never reach the handler. Every spool write reachable from `run()` goes through it. (Separately: that unconditional swallow means a full spool dir already loses telemetry silently — pre-existing, worth its own issue, out of scope here.)

### Fixed

1. **`nix/home.nix` — an inaccurate justification I wrote.** The comment claimed *"the tailers checkpoint state, so an interrupted run just resumes."* True of `session-tailer.py` (incremental `save_state`), **false of `tailer.py:269-270`**, which saves once after its whole loop. SIGTERM at logout after 800 of 1200 messages → state unchanged → all 800 re-emit. Those `kind=prompt|command` rows have no argMax dedupe on read, so counts inflate. The hazard predates this PR, but I removed the toast that was the only signal it happened *and* justified it with a claim the code doesn't support. Now documented honestly, with the real fix (checkpoint `tailer.py`) named as follow-up.

2. **`keylog.py` — the fix didn't actually cover the start path.** `KeyLogger(...)` was constructed *outside* the guard, and `display.Display()` raises `DisplayConnectionError`, which the old tuple didn't catch (verified: its MRO is `DisplayError → Exception`, not `OSError`). So on X teardown the phantom toast was deferred one restart, not eliminated. Construction now sits inside the guard. Also hoisted the Xlib import out of the `except` block, and dropped `ConnectionError` as a redundant `OSError` subclass.

3. **`i3source.py` — closed the silent-restart-loop hole.** A stale `I3SOCK` that accepts a connect and returns an immediately-ending loop would restart every 10s forever with i3 telemetry dead and nothing latching to `failed` (`StartLimitBurst=5` over `StartLimitIntervalUSec=10s` is never reached at `RestartSec=10`). A clean return sooner than `MIN_SESSION_SECONDS` is now treated as a fault. Threshold is 5s; the shortest legitimate session in 30 days of journal was ~17s.

### Verification

Stub harness over both `main()` functions — all six pass:

| case | expected | got |
|---|---|---|
| keylog: `ConnectionClosedError` from `run()` | 0 | 0 |
| keylog: `OSError` (bare socket) from `run()` | 0 | 0 |
| keylog: `DisplayConnectionError` from **constructor** | 0 | 0 |
| keylog: `ValueError` (real bug) | propagates | propagates |
| i3source: long session (normal i3 restart) | 0 | 0 |
| i3source: instant return (hot loop) | 1 | 1 |

Existing suites pass (73 passed/1 skipped, 22 passed). `home-manager switch` applied on workbench; zero `Triggering OnFailure=` since.

### Still open

- **Test coverage** (audit finding 4) — deliberately skipped for now. Neither changed path has a committed test; the harness above was throwaway.
- **`i3-source` / `keylog` end-to-end** still unreproduced — they need a real i3 restart / X teardown.
- **Unverified:** whether a `TimeoutStartSec=600` expiry is masked by `SuccessExitStatus=TERM`. systemd sets `Result=timeout` before sending SIGTERM, so it *should* still fail, but neither I nor the auditor confirmed it experimentally.

PR #198, opened 2026-07-30
feat(repo-cos): tag approved clawgate Tasks with `initiative:<slug>`

ZacxDev:
## What changed

An **approved** repo-cos proposal now POSTs to clawgate carrying `tags: ["initiative:<slug>"]` (clawgate 0.7.75+), promoting the existing display-only `↳ relates to:` breadcrumb into a real routing key.

| File | Change |
|---|---|
| `routing.py` | `slug_tokens` / `slug_drop_reason` / `taggable_slug` — the deterministic **denylist** over the mined initiative vocabulary. Pure + table-testable; `taggable_slug` logs every drop with its rule. |
| `clawgate.py` | `normalize_tag` / `normalize_tags` (clawgate's tag grammar), `build_tags` (at most one `initiative:` tag + the documented `runbook:` seam), `post_task(..., tags=)` + the fail-open retry. |
| `exclusions.py` | `attach_related()` stamps the resolved slug onto proposal dicts; `write_last_emailed(..., related=)` persists it; `_approve_payload` carries `initiative` through to the poster. |
| `scan.py` | threads `related` into `_deliver` → `write_last_emailed` / `_persist_latest`, and `build_tags(prop)` → `post_task(tags=…)` in `_post_approvals_to_clawgate`. |
| `README.md`, `claude/skills/repo-cos/SKILL.md` | documented. |

**Where the slug comes from (deliberate):** the approve path runs on a *later* run than the digest that proposed the item. The slug the digest resolved is **persisted into `last_emailed.json`** (`initiative`) at email time and read **positionally** at approve time — it is **never re-resolved**. So the tag matches exactly the breadcrumb Zach saw when he approved, and the approve path gains **no new I/O** (no extra kubectl port-forward to the homelab initiatives store). A `last_emailed.json` written before this change simply yields no tag. `_persist_latest` stamps the same field so `load_last_emailed`'s history/latest fallback stays coherent. Nothing was repointed at `latest.json`.

## Denylist rules (and what they drop)

`initiatives.current` is *mined*, not curated — the scan adopts session titles and handoff-doc filenames verbatim. A junk breadcrumb is harmless; `initiative:handoff` stamped on a durable task is worse than no tag. Rules (each pinned by a real slug):

| Rule | Drops | Real examples |
|---|---|---|
| `empty` / `too-short` | nothing to tag, <3 chars | `""`, `"ab"` |
| `no-letters` | a bare date/number adopted from a dated filename | `2026-07-21`, `2026-06-13` |
| `screaming-case` | **no lowercase letter at all** — a doc FILENAME adopted verbatim. Also: the tag grammar forces lowercase, so the tag would no longer equal the ledger slug and the initiatives-side join would silently miss. | `HANDOFF`, `SESSION-HANDOFF`, `APP-DISCOVERY-DESIGN`, `COMFYUI-INTEGRATION-DESIGN` |
| `opaque-id` | a token starting with a digit + ≥8 alphanumerics (ClickUp id) | `868j34n9y-868kf6w7r-complete-mark`, `868f9pd14-close-issue` |
| `generic` | **every** token is document/process filler (`GENERIC_TOKENS`) | `actionable-next-steps`, `next-session`, `past-sessions-week` |

Validated against a 120-day replay of the same scan that mints the store (200 distinct slugs): **11 dropped**, 189 kept. Near-misses that a lazier rule would have eaten are pinned as KEPT: `SECURITY-AUDIT-v0.1.64` (has a lowercase `v`), `image-scan-ingestion-down-0313-rca` (`0313` is a date, not an id), `standup-triage-handoff` / `clawgate-documentation-handoff` / `spend-exporter-handoff` (real arcs that merely end in `-handoff`), `remix-next-session-kickoff`, `svi_loop_recipe`.

Deliberately **kept** as ambiguous (pinned by test): `code-open`, `dispatch-yes`, `changes-merge-ship`, `from-memory-previous`, `component-file`, `durable-new-tag`. No non-overfit rule separates them from genuine short slugs, and a missing tag is cheap while a wrong tag is not. The router's confidence gate already means most never reach the tag path.

A further 4 long slugs (e.g. `alert-civitai-disk-investigate-prioritize-remediation-space`) are dropped by the **64-rune** tag cap once `initiative:` is prefixed — correct, since they would `400`.

## Fail-open (the load-bearing invariant)

repo-cos only suppresses a proposal's evidence when the POST returns a task id, so a tag-induced `400` would **silently lose the approval for a week**. Defended twice:

1. **Client-side normalize + validate** — lowercase, whitespace→`-`, charset `[a-z0-9._/-]`, ≤1 `:`, ≤64 runes, dedupe + sort, ≤20 tags. Anything that can't be made valid is **dropped, not mangled** (a silently-rewritten tag is a routing key pointing at the wrong thing).
2. **Backstop retry** — a **4xx** on a *tagged* post retries **exactly once** with the `tags` key removed; both the failure and the retry are logged. A tagged-post failure degrades to an **untagged Task**, never to **no Task**. Connection errors and 5xx are **not** retried (not a tag problem, and no retry loop).

Back-compat preserved: a call with no tags sends the byte-identical pre-tags payload (`tags` present only when non-empty, exactly like `repo`/`model`); pinned by an exact-equality test.

## No `runbook:` tags

Not emitted. `runbook:` is *hard*-validated by clawgate (unknown name → `400`), there is no machine endpoint to list/validate names, and exactly one runbook exists (`perf-deep-dive`, an already-measured non-converging loop). A clearly-marked seam in `clawgate.build_tags` documents exactly where and how one would slot in.

## Tests

**257 → 293** (`nix-shell -p 'python3.withPackages(p:[p.pytest p.requests])' --run 'python -m pytest scripts/repo-cos/tests -q'`), 36 new:

- **Denylist** (`test_routing.py`): table-driven over real slugs — each junk pattern dropped with its exact reason, every genuine slug kept, ambiguous cases pinned, drop-logging asserted (and asserted *silent* when keeping).
- **Normalization**: uppercase→lowercase, whitespace collapse, disallowed charset → dropped (not mangled), >64 runes → dropped, empty/whitespace → dropped, dedupe/sort/20-cap; every emitted tag asserted against the clawgate grammar.
- **Persistence**: `last_emailed.json` round-trips the slug; short/absent `related` tolerated; an OLD file lacking the field → no tag, no crash.
- **Approve path**: persisted slug → `tags: ["initiative:<slug>"]`; no slug / junk slug → **no `tags` key at all** (exact payload equality vs today's shape); the approval is recorded + suppressed in every case.
- **Fail-open**: stubbed 400 on the tagged POST → exactly ONE untagged retry, the retry's id is returned and the evidence is still suppressed; untagged 4xx, 5xx and connection errors each make exactly ONE call (no loop).
- **Never-raises**: a routing failure yields no tags and leaves the digest + approve path working.

## Verified live

- `--no-llm` and `--dry-run` smoke runs still render the digest (run against a sandboxed `HOME` so real `~/.config/repo-cos` state was untouched).
- **Real POST** against live clawgate `192.168.50.250:30302` through the actual `post_task` path: mixed-case slug `Repo-Cos-Selfhosted-Complete` → stored as `["initiative:repo-cos-selfhosted-complete"]` (`GET /api/tasks/90`), then `DELETE`d.
- **Fail-open verified live**: posting `runbook:no-such-runbook-xyz` (hard-validated) → real `400` → one untagged retry → task 91 created → deleted.
- Both left exactly the pre-existing 5 tasks (73, 74, 75, 76, 77).


ZacxDev:
**Adversarial audit — no deploy-blocking findings.** Two corrections to this PR's description, posted here rather than edited in, since a reviewer may have already read the current version.

**1. The retry is scoped to all 4xx, not to a "tag-induced 400".**
The description says a tag-induced `400` triggers the untagged retry. The code (`scripts/repo-cos/clawgate.py:343`) retries on any 4xx when `clean_tags` is non-empty. Practical consequences: `401`/`403` (rotated hook token), `404` (wrong URL) and `429` each get a second doomed request, and `408` — a proxy request-timeout *after* the origin already created the task — is the one shape that could produce two Task cards. Low probability on the LAN NodePort path (nothing proxying in front of it), and it never costs an approval, so this is a should-fix rather than a blocker. Narrowing to `exc.code == 400` costs nothing and matches the documented failure mode.

**2. The denylist statistics are from the 120-day replay corpus, not the live vocabulary.**
The body's "200 slugs / 11 dropped / a further 4 long" comes from replaying the slug generator, which the PR does say — but it reads as the live store. Measured against the actual `initiatives.current` (the store was unreachable during implementation and is back up now):

| | live |
|---|---|
| distinct slugs | 139 |
| denylist drops | 8 (`screaming-case` 4, `generic` 2, `no-letters` 1, `opaque-id` 1) |
| dropped by the 64-rune cap once prefixed | 3 |
| **net taggable** | **128 (92%)** |
| charset drops | 0 |
| lowercase collisions (two slugs → one tag) | 0 |

The replay corpus was a faithful stand-in — every rule holds on the real data and nothing in the remaining 131 needs a new rule.

---

**Also worth recording** (details in the review thread, not blockers):

- The `screaming-case` rule contradicts its own stated rationale (`routing.py:207-214`). It justifies dropping all-caps slugs partly because the tag grammar forces lowercase, so the tag would no longer equal the ledger slug and the initiatives-side join would silently miss — but mixed-case slugs are *kept* and then lowercased anyway, hitting the identical join-miss. Exactly two live slugs are affected: `HANDOFF-comfyui-session` and `SECURITY-AUDIT-v0.1.64`. The first is also precisely the doc-filename class the rule exists to drop; it survives on a single lowercase token.
- `attach_related` (`exclusions.py:293-297`) tolerates a length mismatch silently. Today's caller can't misalign, but a future digest reorder or filter would silently tag the *wrong* initiative. A one-line length guard makes that loud.
- Nothing consumes `initiative:` tags yet — the join lives on the initiatives side as future work — and the sibling poster `scripts/initiatives/dispatch.py`, whose cards are literally about initiatives, was left un-updated and has now diverged from `clawgate.py`.

**Verified, not assumed:** test suite run on both branches (base `257 passed`, this branch `293 passed` — claim confirmed); the denylist executed against the live 139-slug store; the fail-open path traced end-to-end (single retry, no loop, byte-identical untagged payload, evidence suppression correct on both retry-success and total-failure). The tests are not vacuous — the back-compat guard is exact-dict equality and the 400-retry test exercises the real `post_task` through the scan poster.

ZacxDev:
## Audit follow-ups applied — `d7ea39d`

Three should-fix findings from the adversarial audit. The fail-open invariant (a tag must never cost an approval) is untouched.

### 1. The `screaming-case` rule contradicted its own rationale → generalised to `not-lowercase`
`routing.py` · `slug_drop_reason`

The docstring justified dropping all-caps slugs partly because *the tag grammar forces lowercase, so the tag would no longer equal the ledger slug and the initiatives-side join would silently miss* — but mixed-case slugs were **kept and then lowercased anyway** by `clawgate.normalize_tag`, hitting the identical join-miss. `SECURITY-AUDIT-v0.1.64` was emitted as `initiative:security-audit-v0.1.64`, which joins to nothing.

The rule is now `s != s.lower()`, reason string `not-lowercase` (it appears in run logs and in tests — both updated). Rationale and behaviour now agree, and the invariant is explicit: **an emitted tag always equals its ledger slug byte-for-byte.**

**Live measurement against `initiatives.current` (2026-07-30):**

| | before | after |
|---|---|---|
| slugs in store | 139 | 139 |
| dropped by denylist | 8 | **10** |
| survive denylist | 131 | 129 |
| dropped by the 64-rune tag cap | 3 | 3 |
| **taggable** | 128 | **126** |

The two newly dropped are exactly the ones the audit named: `HANDOFF-comfyui-session` (also precisely the doc-filename class the rule exists to drop) and `SECURITY-AUDIT-v0.1.64`. Full post-fix drop breakdown: `not-lowercase` 6, `generic` 2, `no-letters` 1, `opaque-id` 1.

### 2. Retry narrowed to the grammar-rejection class
`clawgate.py` · `post_task`

`400 <= exc.code < 500` → `exc.code == 400`. Per the task-tags spec §6, `400` is the only invalid-tag response. As written, `401`/`403` (rotated hook token), `404` (wrong URL) and `429` each earned a doomed second request, and `408` — a proxy request-timeout raised *after* the origin may already have created the Task — was the one shape that could **double-POST two Task cards**.

Everything else is preserved: exactly one retry, no loop, retry payload is `payload` minus `tags` with original key insertion order (asserted byte-identical *and* order-identical), evidence suppression still correct on retry-success and correctly absent on total failure.

### 3. `attach_related` length-mismatch guard
`exclusions.py` · `attach_related`

Positional stamping with no length guard. Today's caller can't misalign, but a future digest reorder/filter would silently tag the **wrong** initiative. A non-empty `related` whose length differs from `proposal_dicts` now logs loudly (`⚠ MISALIGNED related initiatives: N slug(s) for M proposal(s) …`) and attaches **nothing** — no tag beats a wrong tag. Does not raise; the best-effort contract holds.

One behaviour change falls out of this: `write_last_emailed(..., related=["only-first"])` with 2 proposals used to stamp the first and skip the rest; it now stamps neither. That's the intended degradation, and the test was rewritten rather than deleted.

### Test hermeticity — broader than flagged
The audit noted `test_approve.py::test_scan_approve_*` reaching a live cross-cluster read via `cmd_scan` → `routing.related_for` → `route.load_current()`. Probing showed **`test_dismiss.py`, `test_exclusions.py` and `test_scan_cli.py` reach it too** — 8 tests total, **11 real store-read attempts per run** in an env where `psycopg2` is installed (with `psycopg2` absent it fails on the import and is swallowed, which is why the suite *looked* hermetic).

Fixed at the seam rather than in `_prime_scan`: a new `tests/conftest.py` autouse fixture stubs `route.load_current` to an empty store for every test, so all four suites and any future `cmd_scan` test are covered. `test_routing.py` re-patches it in-test where it wants a fixture store. **Measured with an instrumented probe: 11 real store-read attempts without the fixture, 0 with it.**

Also removed the dead unasserted `posted["tags"]` capture.

### Tests
**293 → 303 pass, 0 fail** (`nix-shell -p 'python3.withPackages(p:[p.pytest p.requests p.psycopg2])' --run 'python -m pytest scripts/repo-cos/tests -q'`). `scripts/initiatives/tests` unaffected: 713 pass.

New coverage:
- the two mixed-case slugs dropped with reason `not-lowercase`; their all-lowercase forms kept verbatim
- **join invariant as a corpus property** — over `tests/fixtures/initiatives_current_slugs.txt`, a checked-in snapshot of the real 139-slug vocabulary (with the psql refresh command in its header, so it stays hermetic): every surviving slug produces either no tag or exactly `initiative:<slug>` — never a rewritten one
- the live drop counts pinned (`139 → 10 dropped → 129 → 126 taggable`), so a future rule change has to restate them
- per-status retry call counts: `400` retries exactly once untagged; `401`/`403`/`404`/`408`/`429`/`451` and `5xx`/connection-error each make exactly **one** call
- all three `attach_related` branches: matched (unchanged), empty/None (unchanged), mismatched short *and* long (nothing attached + warning), plus an unsized `related` (logged, not raised)

Docs updated to match: `README.md` (denylist + retry + hermeticity + the live numbers) and the stale `4xx` comment in `scan.py`.

PR #211, opened 2026-07-30
docs(clawgate skill): correct two stale facts that actively mislead

ZacxDev:
Two facts in the clawgate skill are wrong, and both cost real time in the session that found them.

## 1. "`POST /api/tasks` has NO `title` field" — wrong since 0.7.75

Migration 0018 added a real `title` column and field. The skill still told readers to smuggle the title through `directory`.

The history is the useful part: the original claim came from reading `notes.go:122`, which is inside **`handleNoteCreate`** — the HTML *form* handler (`r.FormValue`) — **not** `handleAPITaskCreate`, the JSON machine endpoint. A session trusted that reading, "corrected" the skill in the wrong direction, and shipped a browser extension that smuggles its title through `directory` as a direct result.

That matters more than it sounds: `directory` is **not** cosmetic. `inferRepoFromDirectory` matches it against connected repos to infer the dispatch repo, so a title in that field feeds junk into repo inference.

The entry now states what the endpoint actually accepts, flags the `directory` consumer, and records the misreading so the next reader verifies against `handleAPITaskCreate` specifically.

## 2. "Current live: 0.7.74" — actually 0.7.78

Version history extended with what shipped since:

- **0.7.75** — task **tags** as a routing key (migration 0018) + the `title` fix
- **0.7.76** — machine `DELETE /api/tasks/{id}`
- **0.7.77** — the markdown renderer no longer re-parses its own output
- **0.7.78** — machine `POST /api/tasks/{id}/comments`

Two sharp edges recorded with them, because a client needs both *before* writing code:

- **Tags fail hard.** An invalid tag or an unknown `runbook:` is a **400**, so one bad tag breaks *every* create. The grammar is spelled out (≤20 tags, ≤64 runes, `[a-z0-9._/-]`, at most one `:`, no empty half) and the closed reserved namespaces named (`runbook:` / `initiative:` / `gate:` / `auto:`).
- **The renderer fix did not cover agent chat.** `internal/ui/agents_detail.go`'s JS mirror still leaks its own sentinel into an `href` — its md-link regex lacks the exclusion its autolink has. Live today; recorded so nobody assumes 0.7.77 fixed it.

## Why this is worth a PR rather than a note

The skill's own "docs-ahead-of-code lesson" warns that it can drift from reality. This is the inverse — reality moved and the skill didn't — and the specific stale line had already caused a wrong build. An index that confidently states a false contract is worse than one that says nothing.


ZacxDev:
Superseded by #213 (merged as `311c30a`), which was built on top of this branch and carries its `title`-field correction forward.

Closing rather than merging because this branch's remaining unique lines are now **older** than main and merging it would regress today's fixes:
- `nix-shell -p tailwindcss --run "tailwindcss -i ./web/css/input.css …"` **without** the `cd <worktree>/containers/clawgate &&` prefix. Running it from another cwd makes the config's relative content globs resolve against the wrong tree and silently emits a ~5 KB stylesheet with zero utility classes instead of ~36 KB — the FAB then renders zero-sized, Playwright reports it `hidden`, and ~25 e2e tests fail looking exactly like a code regression. That cost hours on 2026-07-30 and #213 documents it.
- `VER=0.7.70` / "latest shipped: 0.7.69" (live is now **0.7.78**)
- "73 pass / 2 skip" (baseline is now **83 pass / 2 skip**)

Nothing here is lost — reopen if I've misread something.

PR #212, opened 2026-07-30
feat(browser-bridge): reach NESTED OOPIFs from `eval --frame` / `upload --frame` (bounded recursive auto-attach) — LIVE-VERIFIED

ZacxDev:
> **✅ LIVE-VERIFIED on real Brave (run #2, build `34769a0`, profile `personal`).** An adversarial security audit's blocking findings are fixed in `7d42d10`; live run #1 then caught that the cascade was **inert** in real Brave, fixed in `34769a0`; run #2 passed both rig checks. Docs corrected in `a3c7424`. `origin/main` merged in at `046268a` (one additive SKILL.md ops-table conflict, both sides kept) — **`MERGEABLE`/`CLEAN`, 241 node + 247 pytest green.**

## ✅ Live verification (the gate that actually caught the bug)

**Check A — 3-level rig:** `eval --frame <leaf>` (GRANDCHILD) → `"grandchild-reached"`, reported url `http://127.0.0.1.nip.io:8901/leaf.html`. Direct-child control → `"mid"`.

**Check B — 7-level deep rig** (all 7 frames enumerated, parent chain `0→117→118→119→120→121→122`, alternating `127.0.0.1`/`127.0.0.1.sslip.io`):

| frame | depth | result |
|---|---|---|
| 119 | 3 | ✅ `"deep3-reached"` |
| 121 | 5 | ✅ `"deep5-reached"` (at the cap — resolves) |
| 122 | 6 | ✅ `oopif_depth_cap:5` (past the cap — refused) |

```
cascade[exit=depth-cap attach=top>5B1F392F7DE254F9DECE8B1380A8211B>CD40054CBE960109E11F39916A3C40AE>DB842E0A84C5962706427874954E2992>98CEE272B651097F0CBFB4F90775F611 events=5 accepted=5 filter=on caps=d5/t50]
```

Four chained sub-session `setAutoAttach` calls → the cascade descends through every permitted level. `events=5 accepted=5` with `exit=depth-cap` → all five nested targets were attributed correctly and the sixth refused, so **`OOPIF_MAX_DEPTH = 5` is a measured guarantee, not a contingent one**. `filter=on` → Chrome accepts the experimental `filter:[{type:"iframe"}]` param and real OOPIFs pass the type gate.

### 📌 Discovered Chrome behaviour (cost a full verify round)

**Chrome does NOT populate `source.tabId` on SUB-session `Target.attachedToTarget` events** — they carry `sessionId` only. Undocumented as far as I can tell, and it is exactly what broke run #1: the audit-mandated fail-closed own-tab check dropped every level-2+ event, so the cascade went **inert** — `frame_not_found`, *never* a depth cap, because no nested session was ever recorded. **"Inert, not capped" is the diagnostic signature.** Ownership is therefore proven by **session parentage** (`source.sessionId` ∈ sessions this cascade attached), with `tabId` still authoritative when present. Recorded in SKILL.md/README as the trap for any future flat-mode CDP cascade.

Also measured: **a ↻ reload DID take** (after an earlier full restart in the same Brave session), contradicting the blanket "reload is unreliable" advice. Nuanced rather than reversed in SKILL.md: try ↻ first, but only with a **deterministic tell** — the `cascade[…]` trace is one (old build → bare `frame_not_found`; new build → error **plus** trace), which turned an unfalsifiable guess into a one-command check.

## The defect

`Target.setAutoAttach({autoAttach:true, flatten:true})` is **not recursive** — sent on a
session it auto-attaches only that session's **direct** child targets. `eval --frame`
(and the OOPIF branch of `upload --frame`) sent it once, on the TAB's top session, then
matched an `attachedToTarget` by frame URL. So a **grandchild** cross-origin iframe never
produced an event on the top session and the op failed `frame_not_found:<url>` — safe,
but the capability was missing. `text`/`html`/`click`/`type`/`key --frame` were
unaffected (they go via `chrome.scripting` + webNavigation frameIds, not CDP).

Confirmed live pre-fix against a three-domain loopback rig (now committed as a fixture):
`frames` already enumerated all three levels, `text --frame <leaf>` read the grandchild
fine, and `eval --frame <leaf>` returned
`frame_not_found:http://127.0.0.1.nip.io:8901/leaf.html` (exit 1).

## The change

`resolveOopifSession` (pure, `extension/protocol.js`) re-arms `Target.setAutoAttach`
**on each attached child session** (CDP flat mode forwards a `sessionId`), walking the
attach cascade down the frame tree until the wanted frame's target appears.

It is **one shared resolver**: `cdpFrameEval` and `cdpSetFileInput` both call it via the
thin `cdpOopifSession` glue instead of duplicating the auto-attach block.

## The trust boundary (audit fixes 1–3, `7d42d10`)

The pre-recursion code drew part of its safety from an **implicit** boundary: it only
ever looked ONE level below a tab whose URL `assertCdpAttachable` had validated. The
cascade removes that, and the first revision filtered discovered targets on *nothing* —
not type, not scheme, not tabId. The boundary is now **explicit**: three checks in
`onEvt`, before a target is tracked, descended into, or matched.

1. **Own tab.** `chrome.debugger.onEvent` is a GLOBAL listener; an event whose
   `source.tabId` isn't this op's tab is dropped. `tabId` is threaded from the SW; a
   caller that omits it **fails closed**. Previously own-tab held only by luck of the
   serial command loop.
2. **`iframe` targets only.** A page can `new Worker(location.href)` to mint a target
   with a url **identical** to a real frame's. Unfiltered that (a) let any cross-origin
   frame permanently deny service to itself by forcing `ambiguous_frame`, and (b) could
   route the operator's JS into a **worker global** after a navigation race.
   `setAutoAttach` now also carries `filter:[{type:"iframe"}]` — but that parameter is
   **experimental**, so a rejection is caught and the call transparently retried without
   it (**fail-soft**); the listener-side check is the authoritative control and needs no
   protocol support.
3. **http/https only** (`isCdpAttachableUrl`). Without it a hostile page embeds
   `<iframe src="chrome-extension://<id>/…">` (any extension with
   `web_accessible_resources`); `getAllFrames` lists it, and a prompt-injected agent
   could run operator JS inside **another extension's origin** — the top-tab guard
   bypassed by being one level down. This was the single highest-value line in the audit.

## Caps and bounds

| constant | value | rationale | error |
|---|---|---|---|
| `OOPIF_MAX_DEPTH` | 5 | real embeds are 1–2 deep; generous headroom, small worst case | `oopif_depth_cap:5` |
| `OOPIF_MAX_TARGETS` | 50 | caps total attached sessions per op, so a frame-spamming page can't force unbounded work even at shallow depth | `oopif_target_cap:50` |
| `OOPIF_SETTLE_MS` | 300 | `attachedToTarget` arrives **asynchronously** — a `setAutoAttach` reply does NOT mean its events landed; a quiet window beats assuming, and keeps the genuine not-found case fast | — |
| `OOPIF_WAIT_MS` | 3000 | hard ceiling on the whole resolution, well under `CDP_OP_BUDGET_MS` (15 s) | `frame_not_found:<url>` |

**Audit fix 5 (correctness):** the ceiling was checked only when the descend queue was
empty, so a page that kept it non-empty never reached it and the real wall was
`CDP_OP_BUDGET_MS`, surfacing as `cdp_timeout:op` — the stated rationale was false as
implemented. The check is now hoisted **above** the descend branch: a true wall, with a
test that sets both caps generous so only the ceiling can end it.

A match already in hand always wins over a cap hit in the same event batch.

## Failure modes

- **Not found / propagation timeout** → `frame_not_found:<url>` (same shape as before).
- **Too deep** → `oopif_depth_cap:5`; **too many targets** → `oopif_target_cap:50`.
- **Ambiguous** → `ambiguous_frame:<n> [<sessionId>:<url>, …]` — a refusal, never a
  silently-wrong frame.

## Audit fix 4 — `ambiguous_frame` has NO escape hatch (documented, not fixed)

The CDP path matches purely by `frame.url` and **discards** `frame.frameId`, so the
documented remedy ("re-issue with the numeric frameId") provably cannot resolve an OOPIF
ambiguity. Two identical `<iframe src="…/widget">` — a duplicated ad slot, common — are
now permanently un-`eval`-able. SKILL.md and README.md now say that plainly instead of
prescribing a remedy that does not work, and a test pins it.

**Not implemented, filed here as follow-up:** the resolver records `parentSessionId` per
attached target and the caller has the frame's `parentFrameId` chain, so a parent-chain
tiebreak is buildable. It was left out because the ancestor→session mapping is subtle
when an intermediate frame is **same-process** (its "target" is the top session, so the
expected parent url matches no session url), and getting that wrong risks selecting the
wrong frame — the exact failure the whole ambiguity check exists to prevent. Not
validatable without live Brave, so it does not belong in this PR.

## Audit fix 6 — my earlier degradation claim was INVERTED

I previously wrote that an untagged `source.sessionId` degrades to a *spurious depth
cap*. Wrong. Untagged → `parent=null` → `depth=1` always → `1 < 5` always → every target
queued, `depthCapHit` never set. The real failure is **complete loss of the depth bound**
(still capped at 50 targets and the ceiling). Corrected in `protocol.js`, SKILL.md,
README.md, and pinned by a test asserting the actual behaviour. **Run #2 has since
settled it empirically: Chrome DOES tag sub-session events with the parent `sessionId`,
depth attribution works, and the depth-5 cap binds** — this degradation path is now a
guarded-against hypothetical, not a live caveat.

The important half holds and the audit confirmed it: **selection is by URL equality alone
and never consults `depth`**, so mis-attributed depth cannot land JS in a different frame
than the URL match selects.

## Audit fix 7 — the `upload` widening (✅ follow-up CLOSED by #216)

`upload --frame` routes through the same resolver, so **a grandchild OOPIF that returned
`frame_not_found` pre-PR is now reachable to depth 5**. When this was written the
autonomous browser-agent's op list still included `upload` with an arbitrary
caller-chosen path, so a prompt-injected agent could have placed a local file into a file
input in a deeply nested attacker-controlled iframe that was previously out of reach — a
**real capability widening on a data-exfil-capable op**, not created by this PR but
extended by it. "Same resolver, no new capability" undersold it.

**That exposure is now closed on `main`:** #216 dropped `upload` from the agent's default
op set, so it is **operator-only** — the model gets `op_not_allowed:upload` and it can be
re-enabled only by an explicit `BROWSER_AGENT_ALLOWED_OPS` opt-in. Merged into this branch
and documented in both SKILL.md and README.md alongside the nested reach. Every upload is
still audit-logged server-side (op + target domain + path). No open follow-up remains.

## Tests

`tests/nested_oopif.test.mjs`: **34** (was 20 in the first revision). The original cases
modelled a **cooperative** page and baked in the assumptions the audit flagged — the mock
emitted `type:"iframe"` in one place and *no type at all* in another and passed either
way, proving the field was ignored. Now adversarial:

- non-iframe target types (`worker`/`service_worker`/`page`/`browser`/`shared_worker`)
  ignored and never descended into;
- a `new Worker(location.href)` target sharing the wanted frame's url neither shadowing
  the real frame nor forcing a denial-of-service `ambiguous_frame`;
- `chrome-extension:`/`file:`/`devtools:`/`chrome:`/`about:`/`data:`/`javascript:`
  children refused, and not poisoning an otherwise-good cascade;
- foreign-tab events dropped; a same-url foreign decoy not creating ambiguity; an
  **omitted `tabId` failing closed**;
- the experimental `filter` param failing soft (retried without it, no re-probing);
- a **real** ceiling test (both caps generous, queue never empty) — the previous one
  asserted a cap and could not distinguish ceiling from cap;
- the untagged-`sessionId` degradation pinned to its **actual** behaviour, mirrored by a
  tagged case proving the cap does bind when Chrome tags;
- an end-to-end hostile-page test through `OPS.eval` (worker + foreign-tab + extension
  decoys all carrying the wanted url) — the prior `SECURITY` test was a tautology over a
  mock that emitted nothing else.

```
node --test scripts/browser-bridge/tests/*.test.mjs   →  241 pass, 0 fail  (198 on main)
pytest scripts/browser-bridge/tests                   →  247 pass, 0 fail  (unchanged by this PR)
```

`git diff main...HEAD -- tests/` remains **insertions only** — no pre-existing test was
weakened, skipped, or deleted. (The edits within `nested_oopif.test.mjs` are to a file
this PR itself introduces.)

## Live-verify rig: `tests/fixtures/oopif-rig/`

Two fixtures, one server, loopback-resolving domains only (⚠ `vcap.me` now resolves to a
real public IP, 103.224.182.214 — documented, not used):

- **Check A** — 3-domain grandchild rig. **This also settles the target-`type` question
  by construction**: if Chrome typed OOPIF targets as anything other than `iframe`, the
  new filter would drop the grandchild and Check A would fail `frame_not_found`. A
  passing Check A *is* the confirmation.
- **Check B** — a 7-level **alternating-2-domain** deep rig (`deep0`…`deep6`;
  alternating `127.0.0.1`/`lvh.me` keeps every frame cross-site from its parent, so each
  is its own local root → its own target, without needing six domains). It discriminates
  the depth question cleanly, with an expected-outcome table: `deep6` →
  `oopif_depth_cap:5` means depth attribution works; `deep6` → `"deep6-reached"` means it
  is broken and the depth cap does not bind; `frame_not_found` means the rig didn't nest
  and the result is inconclusive.

## Verification status

**Live-verified on real Brave** (run #2, see the top of this description). All three
questions that were open are now settled empirically: real OOPIF targets **are** typed
`iframe` (Check A), flat mode **does** tag sub-session events with the parent `sessionId`
so `OOPIF_MAX_DEPTH` genuinely binds (Check B, `exit=depth-cap`), and Chrome **accepts**
the experimental `filter` param (`filter=on` in the trace — the fail-soft retry stays as
cheap insurance for other Chrome builds).

Not verified, and out of scope: behaviour on Chrome builds other than the one tested, and
`upload --frame` on a nested OOPIF end-to-end (it shares the resolver with `eval`, which
is what was exercised — see the `upload` widening note above, whose follow-up stands).

The extension **manifest version was deliberately NOT bumped**. An extension code change
normally needs a **full Brave restart**; a ↻ reload took this time, but only trust that
with a deterministic tell (see above).


ZacxDev:
## ⚠ Live run #1 FAILED — the cascade was inert. Fix pushed as `34769a0`, NOT re-verified.

Live-verify against real Brave (restarted 16:21:35, extension written 16:01:21 — verifiably the new code, incl. the audit fixes) found **the cascade does not work**. 428 green tests and two clean audits missed it; the live gate caught it.

| check | result |
|---|---|
| basic `eval --frame <mid>` (depth 1) | ✅ |
| basic `eval --frame <leaf>` (depth 2) | ❌ `frame_not_found` |
| basic `text --frame <leaf>` (control) | ✅ |
| deep depth 1 | ✅ `"deep1-reached"` |
| deep depth 2 and depth 6 | ❌ `frame_not_found` |

Always `frame_not_found`, **never `oopif_depth_cap`** → no level-2 session was ever recorded. The recursion was **inert, not capped**. Depth 1 resolving rules out the `type` filter (real OOPIF targets *are* typed `iframe`), the scheme check, and the top-session `setAutoAttach`.

### Diagnosed cause and fix (`34769a0`)

The own-tab check — added late in the audit round, never live-tested — required `source.tabId` on every event. Chrome appears **not** to populate it for **sub-session** events, so every level-2+ event was dropped as foreign.

- **Ownership fallback.** `source.tabId` stays authoritative *when present*. When absent, ownership is proven by **session parentage**: the event's `source.sessionId` must be a session this cascade itself attached. Neither → `drop:unowned`. Still fails closed; an *unknown* parent session is still rejected, so this is not blanket trust.
- **Bounded failure diagnostics.** Every OOPIF failure now carries a capped `cascade[…]` readout (exit reason, attach chain, per-event type / tab provenance / parent provenance / depth / drop reason, ≤20 rows + a true "+N more" count). Caller-facing only — telemetry stays metadata-only. Run #2 will produce evidence instead of guesswork.
- **Settle window** 300→600 ms and now **restarts on each new `setAutoAttach`**; hard ceiling 3 s→5 s.

Tests 34 → 43 (**229 node**, **208 python**, all pass), including a repro that fails on the previous code: sub-session events with no `tabId` resolving a grandchild.

### Fixture defect found live

`lvh.me` is on **Phantom Wallet's phishing blocklist** — on the `work` profile it hijacked the rig tab to `chrome-extension://…/phishing.html`. A fixture that silently becomes a phishing interstitial is worse than no fixture. Replaced throughout with `127.0.0.1.sslip.io` (verified → 127.0.0.1; distinct registrable domain from `nip.io`, so every site boundary holds). The rig README keeps the `vcap.me` warning, adds the wallet-blocklist trap, records run #1 in full, and documents how to read the new readout.

**Status: unverified.** Until run #2 passes, assume `eval`/`upload --frame` reach a DIRECT child OOPIF only. Do not merge.

PR #217, opened 2026-07-30
fix(claude): make git rules consistent + ship the enforcing hook to all hosts

ZacxDev:
Audit of the Claude git rules across devrc + `~/.claude` + project CLAUDE.md files found **5 contradictions**, two of them inside a single file. Trigger: a `git pull` in `homelab-talos` found the base clone **262 commits behind** with three stale-file conflicts (`.sops.yaml` missing 10 rules incl. the fail-closed catch-all; remix pinned `0.14.0` against a live `0.37.0`).

## Root cause

Worktrees do all the committing, so the base clone is write-only and never advances. Its dirty files become *stale orphans* of already-merged work — not WIP. The worktree workflow was documented; the **re-sync step that closes it was not**.

## What this fixes

| # | Inconsistency | Fix |
|---|---|---|
| 1 | `RULES.md` (managed, ships everywhere) delegated the CRITICAL rules to `~/.claude/CLAUDE.md`, which is **per-host and does not ship** | Rules move into `RULES.md`; CLAUDE.md keeps host-specific facts only |
| 2 | The stash/pull-rebase/pop recipe contradicted the worktree rule **four lines below it** | Replaced with the clean-worktree flow + `git checkout <ref> -- <paths>` |
| 3 | `bash-guard.py` recommended the stash when blocking a hard reset — propagating the trap mid-recovery | Steers to restore/checkout; warns the stash is repo-global |
| 4 | `bash-guard.py` was per-host/unmanaged — the only enforcement of rules that ship everywhere lived on **one machine** | Managed via `home.nix` |
| 5 | "Feature branches only, never main/master" vs `homelab-talos` trunk-deploy — carve-out existed only in `MEMORY.md` | Written down, scoped to declared trunk-deploy repos |
| 6 | `clawgate/SKILL.md` **contradicted itself**: L68 blessed committing from the main checkout, L217 forbade it | Dropped L68 — it is what produced the drift |

Why the stash rules matter: the stash stack is **repo-global across worktrees**, so a concurrent agent can pop yours. Two parallel remix subagents did exactly that (2026-07-25), and the autostash form corrupted `.sops.yaml` (2026-06-24).

## Verification

- `bash-guard.py` still blocks a hard reset, blind-stage, and `cd <path> && git …`; still allows explicit `git add <path>` (4/4 in a scripted test).
- The **built store artifact** was tested functionally: blocks, no longer recommends the stash, warns against it.
- `nix build .#homeConfigurations.zach.activationPackage` succeeds and lands both `.claude/hooks/bash-guard.py` and the updated `RULES.md`.
- Hook registration in `settings.json` stays per-host and needs **no change** — it already points at `~/.claude/hooks/bash-guard.py`, which the new symlink backs.

## Note on rollout

`home-manager switch` is required for this to take effect; the laptop needs it too (that host has never had the enforcement hook at all).


ZacxDev:
## Correction to the PR description + audit round 1 applied

An adversarial audit found a **factual error in this PR's description**, correcting it here rather than silently editing the body since it may already have been read.

### ❌ Wrong: "the laptop has never had the enforcement hook at all"

Verified on the laptop:

```
-rw-r--r-- bash-guard.py   (4871 bytes, Jun 23)
settings.json: "command": "python3 ~/.claude/hooks/bash-guard.py"   (registered)
CHECKS = [check_git_add_all, check_git_reset_hard, check_heredoc_to_file, check_cd_then_git]
```

The laptop **does** have the hook, registered and live — with 4 of the 6 checks. So rationale #4 ("the only enforcement lived on ONE machine") was wrong. The actual problem is **drift between two divergent copies**, which this PR does fix.

Consequence to state honestly: this is a **behaviour change** on the laptop, not an activation. After the switch it newly gains `check_private_key` and `check_secret_or_ip_publish`, plus updated `check_heredoc_to_file` / `check_cd_then_git`.

### ⚠️ Understated: the laptop's `~/.claude/CLAUDE.md` is a required rollout step

That file is per-host and unmanaged by design, so this PR cannot reach it. It still contains the stash recipe. After merge + switch the laptop would load **both** `RULES.md` ("🔴 `git stash` is repo-GLOBAL — never use it to clear a tree for a rebase") **and** its own CLAUDE.md telling it to run exactly that recipe — the same contradiction this PR exists to remove. Being handled as part of rollout, not left implicit.

### 🔴 Audit findings fixed in `b23a217`

The audit proved commit `25a9310` made the guard **weaker than `2c662dc`** — four shapes blocked before were allowed after. All closed and regression-tested:

- `.sub()` without a count stripped **every** heredoc sharing a tag, blanking a `bash <<'EOF'` body alongside a message one.
- The message-command test didn't split on `$(`, `<(` or backticks, so a heredoc inside a substitution was misread as message text.
- `re.finditer` iterated the original string while the loop rebound it — stale offsets reclassified shell heredocs after the first strip.
- The `-m` pattern honoured `\'` inside single quotes (the shell does not), over-consuming past the closing quote into a real command.
- That pattern was also a **ReDoS**: an unterminated quote with ~40 backslashes stalled for minutes, on a hook that runs on every Bash call. Branches are now disjoint — measured 0.02s, with a timing assertion.

Also: `-m` stripping is gated so `docker run -m '2g'` is untouched; an unterminated heredoc now strips nothing; the suite's path comes from `__file__` instead of a temporary worktree path; a crashing guard is an ERROR rather than 6 spurious passes; and the suite is registered in `scripts/run-tests.sh`, where it previously would never have run.

Hook + test moved to `scripts/claude-hooks/{,tests/}` alongside the other three hooks instead of a third location.

`RULES.md`: restored the "commit before risky operations for rollback" clause dropped while rewording; the trunk-deploy carve-out now depends on the target repo's own `CLAUDE.md` saying so (no self-declaration); dropped the claim that `--ff-only` can conflict.

23 assertions passing; `nix build` of the activation package clean.

ZacxDev:
## Round 3 → the feature is reverted (`72e94be`)

Round 3 found a 🔴 in round 2's fix, and named the real problem:

> each fix has been a regex tweak against the last round's counterexample, and each has swapped one failure mode for another

That's accurate. Rather than open round 4, `_strip_message_text()` is **removed**.

### Why removal rather than a fourth fix

| round | hole found | shape |
|---|---|---|
| 1 | `.sub()` no count; substitution-fed heredocs; stale `finditer` offsets; `\'` over-match; ReDoS | 5 issues |
| 2 | `&` not a separator → `git commit -m x & bash <<EOF` strips an executing body | 🔴 + new FP via bare `(` |
| 3 | widened command list became a **substring** test → `bash -s -- 'git merge' <<EOF` uses a decoy argument | 🔴 + FP back via `&` in `2>&1` |

Every hole was **introduced by this feature** — the guard with no stripping blocked all of them. Deciding which bytes are inert message vs. executable command is shell parsing; regexes can't do it.

Two further facts made the trade obvious:
- The feature never covered the common shape anyway: `gh pr create --body "$(cat <<'EOF' … EOF)"` was blocked at *every* revision including HEAD.
- The false positive it existed to remove costs one extra step — write the message to a file, `git commit -F <file>` — which this repo's RULES already prefer over heredocs. The FP nudges toward the preferred pattern.

### Verification: differential test, not assertion

34 inputs — every known bypass, every accepted FP, the secret/IP cases, plus controls — run against `2c662dc` (pre-feature), all three intermediate revisions, and HEAD:

```
34 inputs; HEAD vs 2c662dc mismatches: 0
RESULT: IDENTICAL to the proven-safe baseline
```

Removal provably restores the baseline three audits found no hole in, rather than inventing a fourth behaviour. The same table shows each intermediate revision scattering `allow` where the base blocked.

### Also fixed from round 3

- **4 vacuous tests.** Their payload ended `-A'`, and `check_git_add_all` requires `-A` followed by whitespace or end-of-string — so they passed identically at `2c662dc`, which has no stripping at all, and could not have tested it. They covered exactly the command patterns where the round-3 🔴 hid. Payloads now carry a trailing word, with a comment explaining why.
- Every reverted bypass is retained as a `was-bypass:` case, so reintroducing anything like the helper fails loudly.
- The accepted false positives are now **asserted** as `FP (accepted):` rather than left to be rediscovered.
- Both deny messages name the workaround.
- A `DESIGN NOTE` in the module docstring records all three rounds so this doesn't get rebuilt.

26 assertions passing; wired into `scripts/run-tests.sh`; activation package builds. Net **-111/+31**.

### Still open from round 3 (pre-existing, unaffected by the revert)

`gh pr create -t x -b '…'`, `gh release create --notes '…'`, and `--body "$(cat <<EOF …)"` are blocked when they quote a rule — true at `2c662dc` and true now. Same one-step workaround; not a regression, and not worth reopening this surface for.

PR #221, opened 2026-07-30
fix(bash-guard): close the git -C / escape / stage blind-stage gaps

ZacxDev:
Follow-ups from the #219 audit. The headline gap is **self-defeating**.

## The gap that mattered

A bare `\bgit\s+add\b` anchor meant **`git -C <dir> add -A` bypassed the guard entirely** — the exact shape `check_cd_then_git` and `RULES.md` actively *push* agents toward. The guard was steering people into its own blind spot.

Replaying **31,989 real commands** from local transcripts found **11 blind `git -C "$WT" add -A …` calls** the guard never saw, across civitai, homelab-talos and datapacket work.

## Structural fix, not another alternation tweak

Replaced whitespace-bounded regexes with **tokenisation**:

| class | previously allowed | now |
|---|---|---|
| global-option hop | `git -C <dir> add -A`, `-c k=v`, `--git-dir=`, `--exec-path=`, `-P`, `--no-pager`, `-C x -C y` | blocked; `stage` accepted as git's documented synonym |
| shell escape removal | `git add -\A`, `\-A`, `--a\ll`, `\.`, `$'-A'` | blocked — backslashes and `$'…'` stripped as the shell does |
| token boundary | `(git add -A)`, trailing `;`, `--al`, `--no-ignore-remov` | blocked — matched per token; git's unique long-option prefixes covered |
| whole-tree pathspecs | `..`, `../`, `../..`, `./`, `.//`, `*`, `$PWD`, `${PWD}`, `$PWD//`, `$PWD/./`, `` `pwd` ``, `$(pwd)//`, `:/`, `:/*`, `:`, `:(top)`, `:(top,glob)`, `:(top,icase)` | blocked — every one verified by execution to stage the whole tree |

## Two things it also fixes

- **Quadratic backtracking.** The old pattern had two adjacent `[A-Za-z]*` around the `A`: **7.14s on a 32k input**, >8s (timeout) at 200k, on a hook that runs for every Bash call. Now 0.001s / 0.029s at 1M. A timing case covers it.
- **An accepted FP.** After `--`, arguments are paths — so `git add -- -A` (staging a file literally named `-A`) is allowed again.

## On `..` — I got this wrong twice, so here is the measurement

| depth | result |
|---|---|
| repo root | 0 staged, rc=128 — `fatal: '..' is outside repository` |
| depth 1 (`a/`) | **WHOLE TREE** |
| depth 2 (`a/b/`) | scoped — stages `a/` only |

It is blocked. Exact at depth 1, harmless at the root (git rejects it anyway), fail-closed over-block deeper — consistent with the module's stated "err toward over-matching", and **0 occurrences in 32k real commands**. An earlier root-only test led me to exclude it; a depth-1-only test then led me to claim it is *always* whole-tree. Both were generalised from one depth. The code comment now records all three and says to measure at 0, 1 and 2 before touching it.

## Known limitation, deliberately not chased

Whether `X/..` is whole-tree depends on runtime cwd and repo-root depth, which the command text does not carry: `a/..` at the root is everything, `a/b/..` is just `a/`, and staging `a/` is allowed. Same for `$OLDPWD` and `$(git rev-parse --show-toplevel)` (which the per-token split shatters anyway). Guessing at these is the trap the module's `DESIGN NOTE` records. Documented **in code**, not only here.

## Verification

- **Execution:** every newly-blocked form run in a throwaway repo and confirmed to stage every file, including untracked ones. Every widened stem confirmed *not* to catch scoped paths built from it (`$PWD/<file>`, `${PWD}/x`, `$(pwd)/<file>`, `$PWDX`, `$PWD_BACKUP`, `:(top,glob)x`, `:(exclude)vendor`, `:!foo`, `..foo`, `.hidden`, `.github/…`, `../sibling/file.txt`, `../pkg/`).
- **Replay:** 31,989 real commands through both revisions → **0 loosening, 0 errors, 13 newly blocked.** All 13 inspected individually.
- **Sibling subcommands verified unaffected:** `git worktree add` (the workflow RULES prescribes), `remote add`, `submodule add`, `notes add`, `subtree add`, `git stash` — all still allowed, and locked in by tests.
- Other five checks **byte-identical** to main; **113 assertions**, wired into `scripts/run-tests.sh`.

**Friction note:** 6 of the 13 newly-blocked are *path-scoped* (`git -C $W add -A containers/remix` stages that path, not the tree). Blocking them is policy-consistent — base already blocked plain `git add -A src/`, and the `-C` hop is the only reason they were invisible — but it is a real change in what gets stopped.

### DIFF
--- a/scripts/claude-hooks/bash-guard.py
+++ b/scripts/claude-hooks/bash-guard.py
@@ -48,14 +48,65 @@ import sys, json, re, ipaddress
 # `--no-ignore-removal` is git's documented alias for `--all`.
 # Erring toward over-matching is deliberate: this guard fails CLOSED, and a
 # false positive costs one `git add <path>` retype.
-_BLIND_FLAG = re.compile(r"(^|\s)(-[A-Za-z]*A[A-Za-z]*|--all|--no-ignore-removal)(\s|$)")
-_BLIND_PATH = re.compile(r"(^|\s)(\.|:/)(\s|$)")
+# `git [global-opts] add|stage <args>`. The global-opts hop matters: `git -C
+# <dir> add -A` is the shape check_cd_then_git and RULES actively PUSH agents
+# toward, so a bare `\bgit\s+add\b` anchor made this guard self-defeating —
+# real transcripts contain blind `git -C "$WT" add -A` calls it never saw.
+# `stage` is git's documented synonym for `add`.
+_GIT_ADD = re.compile(
+    r"\bgit\s+"
+    r"(?:(?:-[Cc]\s+\S+"
+    r"|--(?:git-dir|work-tree|namespace|exec-path)(?:=\S+|\s+\S+)"
+    r"|-[pP]|--paginate|--no-pager|--bare|--literal-pathspecs)\s+)*"
+    r"(?:add|stage)\b([^&|;\n]*)"
+)
+
+# Tokens that stage the whole tree. Checked per-TOKEN rather than with a
+# whitespace-bounded regex: `(git add -A)` and `git add -A;` defeated the
+# boundary, and nested `[A-Za-z]*` around the A backtracked quadratically
+# (6s on a 32k run of A's, on a hook that runs for every Bash call).
+# Short bundles: -A is the only uppercase-A short option `git add` has, so any
+# all-letters dash-token containing A is blind (-Av, -vA, -uA).
+_BLIND_SHORT = re.compile(r"-[A-Za-z]+")
+# git accepts unique long-option PREFIXES: --al == --all, --no-ignore-r == …removal.
+_BLIND_LONG = re.compile(r"--al|--all|--no-ignore-r[a-z-]*")
+# Whole-tree pathspecs: . ./ .// ./. :/ * $PWD $(pwd) :(top)
+# NOT `..` — verified by execution that it does not stage the tree (git errors
+# on it at the repo root), so blocking it would be an unproven over-block.
+# Hence `\.(?:/[./]*)?` rather than `\.[./]*`.
+_BLIND_PATH = re.compile(r"\.(?:/[./]*)?|:/|\*|\$PWD|\$\(pwd\)|:\(top\)\*?")
+
+
+def _stages_everything(argtext):
+    # The shell removes quotes, backslashes and the $ of $'…' before git sees
+    # the word; do the same, or `git add -\A` / `git add $'-A'` slip past.
+    t = re.sub(r"\$(?=['\"])", "", argtext)
+    t = t.replace('"', "").replace("'", "").replace("\\", "")
+    after_ddash = False
+    for raw in t.split():
+        if raw == "--":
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Audit round 1 → fixed in `c34cb39`, plus a correction to this PR's body

### 🔴 I committed a false, execution-attributed claim

The PR removed `..` from the blind set with this comment:

> `NOT `..` — verified by execution that it does not stage the tree`

**That is true only at the repo root.** I tested there and generalised. Re-verified properly:

| command | staged |
|---|---|
| `..` at repo ROOT | 0 — `fatal: '..' is outside repository`, rc=128 |
| `..` from `sub/` | **WHOLE TREE** |
| `git -C <repo>/sub add ..` | **WHOLE TREE** |
| `git -C <repo>/sub add ../` | **WHOLE TREE** |
| `git -C <repo>/sub/deeper add ../..` | **WHOLE TREE** |

Reachable with no `cd`, in exactly the `git -C <dir>` shape this PR exists to close. Worse, the test asserted the allow was *correct* — so a future fix would have read as a test regression. Both the comment and the case are corrected; blocking the root form costs nothing since git rejects it anyway.

Also now blocked, each verified to stage the whole tree: `${PWD}`, `$PWD/`, `` `pwd` ``, `:/*`, `.//`.

**Deliberately not chased:** `git add ./sub/..` also stages everything at the root — but `a/b/..` resolves to `a/`, which is scoped, and staging `a/` is allowed. Text alone can't separate them. That's the trap the module's `DESIGN NOTE` already records, so it's documented as a limitation rather than guessed at.

### ✏️ Correction to this PR's description

The body says the 13 newly-blocked corpus commands are all "genuine blind `git -C … add -A`". **Six of them are path-scoped** — e.g. `git -C $W add -A containers/remix` stages everything under that path, not the whole tree.

Blocking them is *policy-consistent* (base already blocked plain `git add -A src/`; the only reason they were invisible is the `-C` hop), but it's a real friction change and the original wording overstated it.

### Test quality, same audit

- **De-vacuumed** the `--git-dir=` case: its path ended in `.git`, so the literal two-word substring made base block it without ever exercising the opt-hop.
- Added the **highest-value regression guard**: sibling subcommands must not be swallowed by the opt-hop — `git worktree add` (the workflow RULES prescribes), plus `remote`/`submodule`/`notes add` and `git stash`. A regression there would block the thing the rules tell you to do.
- Added: a blind flag *before* `--`, a file named `-A` *after* `--` (allowed), `-C x -C y`, `--exec-path=`, `--no-pager`, `$PWD/<path>`, a rooted glob.

### Verification

**90 assertions.** 31,932 real commands replayed against main: **0 loosening, 0 errors**, and the **same 13** newly-blocked as before this fix — so the added patterns cost nothing in real usage.

Confirmed unchanged by the audit: the other five checks are byte-identical, and the ReDoS fix holds (32k-`A` input went 7.14s → 0.001s; 1M → 0.029s where base times out).

PR #223, opened 2026-07-30
docs(skill): auditloop — push runs can be DOM-grounded, with the measured reach limit

ZacxDev:
The `auditloop` skill asserted "plugin-PUSH runs remain screenshot-only (NOT DOM-grounded)". That is now false — auditloop #39 and civitai-manager #31 are merged and deployed.

Replaces it with the capability **and** the measured limit, so the next session neither re-litigates the gap nor overclaims the fix:

- push runs may carry an optional per-page `a11y_digest`; the same `dropContradicted` gate applies; producer contract lives in auditloop `CLAUDE.md`.
- **Reach limit (measured, not inferred):** the gate indexes only concrete anchors (`#id` / `[name=…]`), but the evaluator cites selectors it invents from pixels — running the real captured digest and the real prior claims through `dropContradicted` gave **4 in, 4 survived**. The observed a11y-FP improvement is LLM variance, not the gate. The digest reaching the *prompt* is what helps today.

Docs-only, one file. Needs a `home-manager switch` on each host to take effect (skills are read-only symlinks into the nix store).

### DIFF
--- a/claude/skills/auditloop/SKILL.md
+++ b/claude/skills/auditloop/SKILL.md
@@ -201,11 +201,27 @@ The persona evaluator reasons over **screenshots only** unless DOM-grounded, so 
 `sr-only` labels, ARIA, `<a>`/`<button>` semantics, and JS focus/keyboard behaviour → it
 **invents a11y false positives AND can't confirm its own a11y fixes**. The DOM/a11y grounding
 (#35 crawl path + #37 driven path + #36 structural label gate) fixes the **crawl + driven**
-paths (the `dropContradicted` gate refutes contradicted findings deterministically).
-**⚠ plugin-PUSH runs remain screenshot-only (NOT DOM-grounded)** — treat their a11y/keyboard
-findings with skepticism. Refs:
+paths (the `dropContradicted` gate refutes contradicted findings deterministically). Refs:
 `claudedocs/evaluator-meta-audit-naida-outreach-2026-07-24.md`,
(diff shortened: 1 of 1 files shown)


ZacxDev:
Rebased onto current `main` and verified against the auditloop source before landing.

**This PR fixes a claim `main` currently gets wrong.** The skill still says *"plugin-PUSH runs remain screenshot-only (NOT DOM-grounded) — treat their a11y/keyboard findings with skepticism"*, and #39 made that false. Verified in `~/workspace/auditloop`:

- `NormalizeA11yDigest` — `internal/plugin/a11y.go:54`
- `dropContradicted` — `internal/eval/verify.go:36`
- `pages.a11y_digest_key` present, with a dedicated e2e (`tests/e2e/plugin_a11y_digest_test.go`)
- the empty-digest rejection is real: `a11y digest is empty (no interactive elements, form controls or landmarks)` at `a11y.go:172`

**The reach limit checks out too**, which is the part worth keeping honest: `buildDigestIndex` keys the index via `concreteKeys(sel)`, so it genuinely only indexes `#id` / `[name=…]` anchors. A target whose markup lacks ids gets nothing dropped — exactly as the PR says, and I've added the `concreteKeys` reference so the next reader can confirm it in one grep rather than taking the sentence on trust.

Conflict was purely structural: `main` had reflowed this paragraph to ~100 cols since the branch was written. Resolved by taking the PR's content and reflowing to match.

Keeping the "prevention, not deterministic dropping — a single before/after run cannot be distinguished from LLM variance" caveat verbatim. That is the most valuable sentence in the diff.

PR #224, opened 2026-07-31
fix(dl-router): adversarial audit remediation — relocate provenance, MV3 listener order, fail-closed backfill

ZacxDev:
Fixes the adversarial audit of the `dl-router` subsystem merged as `56babdd` (#220).

**Every finding has a regression test that fails on the pre-fix code and passes now.** Where a fix
changes behaviour I verified the "fails before" by checking out the pre-fix module and re-running
that file; the counts are in the individual commit messages.

Suites: **530 pytest** (was 341) and **219 node** (was 151), all passing. Plus a differential fuzzer
reporting **0 divergences over 80,000 comparisons** where there were 991.

```
nix-shell -p 'python312.withPackages(ps:[ps.pytest])' --run "python3 -m pytest scripts/dl-router/tests -q"
nix-shell -p nodejs --run "node --test 'scripts/dl-router/tests/*.test.mjs'"
nix-shell -p nodejs python312 --run "python3 scripts/dl-router/tests/difffuzz.py"
```

The extension has still **never been loaded**, no `backfill apply` was run, nothing wrote to the
library, and the live qBittorrent instance was never contacted. The sidecar unit is still running
and healthy (it executes from the nix store, so this branch has not been deployed to it).

---

## Priority 1 — before the extension can be loaded

**1. `/relocate` would move any pre-existing file** — `bca0d65`

The endpoint validated `toDir` and confined the source with `safe_rel_path` (there was never a path
escape) and still `os.rename`d whatever existed at the supplied relative path, inside a live seeding
target. Two independent proofs are now **both** required, and the absence of either is a refusal:

1. *Provenance* — the path is already in the new `routed_files` ledger, or a `downloadId` is supplied
   whose `/match` decision is on record **and named the directory the file is currently in**. Without
   that second check one legitimate download would authorise moving any file in the library.
2. *Age* — the file's mtime must be at or after its routing decision. A browser writes after the
   decision; a payload already on disk predates it.

Store schema v2 adds `routes.download_id` and `routed_files`; the ledger follows a file through a
relocate so a second `change` click still has provenance, and a refused relocate does not seed the
ledger it just failed. The migration is additive and I verified a v1-shaped writer still works
against a migrated DB (relevant because the running sidecar is still v1 until this ships).

Extension side: `relPathFromAbsolute` now **requires** the library root (carried on `/dirs`) and
returns `null` unless the path is exactly `<root>/<one dir>/<one file>`. Outside the root, loose at
the root, nested deeper, a shared-prefix sibling, or no snapshot at all all yield `null` = "do not
ask the sidecar to move anything".

**2. MV3 listeners registered after `await`** — `cd64bae`

Split into `registerListeners()` (synchronous, first turn, no awaits) and `start()` (everything
async). `onDeterminingFilename` waits on `ready()` internally; on a cold wake it returns `true`,
awaits readiness, then routes, and if readiness does not land in 250 ms it calls `suggest()` with no
argument rather than routing something it cannot prove is enabled. `suggest()` is still called
exactly once on every path.

`tests/service_worker_start.test.mjs` is a new file that **holds the storage read open** and asserts
the listeners already exist — a real test of registration order rather than an assertion on a mock.
All 7 cases fail on the pre-fix module. `start()` had zero coverage.

**3. `handleDetermining` failed OPEN on a non-Set `knownDirs`** — `3eaf579`

`: null` meant `sanitizeDirName(name, null)` skipped the allowlist entirely. Now falls back to a Set
containing only the catch-all. Six table-driven cases (Array/null/undefined/object/Map/string) fail
6/48 pre-fix.

**4. `/fetch` bypassed the allowlist** — `f48cc09`

Now uses the same index-plus-catch-all set as every other endpoint, and `fetcher.submit` resolves the
destination with `safe_rel_path` — `is_safe_dir_name` proves the *name* is one component and says
nothing about where it *resolves*, and yt-dlp follows a symlinked `<root>/<dir>` out of the library
via `--paths`.

---

## Priority 2 — before `backfill apply` is ever run

All in `8158752`. 26 of the new backfill cases fail against the pre-fix `backfill.py`, and
`test_qbt.py` does not even import against the pre-fix `qbt.py`.

**5. Per-file fail-open.** `move = MOVE_QBT if torrent else MOVE_FS` turned an index *miss* into a
plain rename, and the index only knew `content_path` and `save_path/name` despite the docstring's
"and its files". A multi-file or no-root-folder torrent's payload sits directly at the library root —
exactly this tool's target population — so those files were missed and renamed out from under a live
seed.

* `QbtClient.torrents_files()` + `index_by_host_path(..., files_for=...)` index every file.
* The index is now a `TorrentIndex` carrying `complete`, so absence only means "not torrent-backed"
  when the index is known exhaustive.
* Classification is `qbt` / `fs` / `unknown`, and unproven is always `SKIP`. A reachable qBittorrent
  with **no torrents at all** stays positive proof, so the common case still works.
* `derive_path_map` considers the whole list (it sampled the first 50 while claiming "the mapping
  most torrents agree on"), needs more than one corroborating vote **and** an outright winner, and
  verifies the map can express the library root — a map that cannot is worse than none, because it
  classifies every loose file as not-torrent-backed.

**6. `verify_seeding` aborted mid-move.** `setLocation` returns when the request is accepted, not when
the payload has arrived; qBittorrent then sits in `moving`, which is in neither `SEEDING_STATES` nor
`BROKEN_STATES`. Now polled out with a bounded budget; clock and sleep are injectable so tests never
actually wait.

**7. The reviewed artefact was not the applied artefact.** The TSV header said to edit `action` while
`load_manifest` read JSON only. The TSV is now a first-class manifest (`#!root`/`#!path_map`
metadata, parsed back, malformed input refused rather than guessed), the CLI points at the TSV, and
applying the JSON when the sibling TSV has been edited is **refused** instead of silently running the
unedited plan.

**8. `apply` never re-validated.** It now re-derives the path map and torrent index from the live
instance before touching a row, uses the **live** hash, and aborts if a row's live classification
disagrees with the manifest. A client is required whenever *anything* will move, not just for `qbt`
rows — an `fs` row is only safe if live state can still prove it.

**9. Filename cap.** The stem was smuggled in as a page *tag*, scoring 0.85 through the tag-exact rule.
The cap is respected: the only signal that may carry a row is an explicit **alias** on the stem
(seeded from a directory or torrent name, or hand-set) — recorded knowledge, not a guess about an
opaque filename. A new `signal` column labels every row `alias`/`filename`/`none`.

---

## Priority 3

**10. `safety.py` vs `sanitize.js`** — `eb42e3e`. The two hostile-input tables were hand-copied
literal lists; both passed while the *implementations* disagreed on **991 of 16,000** fuzzed inputs
neither list contained. Fixed to the stricter reading on both sides (C1 controls, U+FEFF and the
zero-widths, dot-strip ordering, `:` — verified no existing subject directory contains one), and
`is_http_url` no longer delegates to `urlsplit` or `new URL` at all: neither parser can be the
contract (WHATWG collapses the slashes in `http://////..` into the host `..`; `urlsplit` accepts
`http://%2f`), so both now validate the authority against one explicit rule.
`tests/fixtures/name_cases.json` is now THE table and both suites load it, with a non-empty assertion
so a broken load cannot make them vacuously pass. `tests/difffuzz.py` ships the fuzzer.

**11. The host prior was decisive** — `3eaf579`. The bonus went into `score`, ranking read `score`,
and the tie-break compared `score` — so a prior on the runner-up of two 0.85 candidates both
reordered the pair and manufactured the 0.05 gap that suppressed the tie-break. Ordering, threshold
and tie margin all read `base` now; the prior only breaks an exact tie, where the margin check fires
anyway. `confidence` is the pre-bonus score. README corrected.

**12. `route_core.js` was a git binary blob** — `706609d`. Three NULs became `\u0000` escapes; every
`extension/*.js` is now pure ASCII; `tests/source_hygiene.test.mjs` enforces it and `.gitattributes`
forces a textual diff. Verified `git diff` renders the module as text.

**13–17** — `8158752` (13) and `cd64bae` (14–17): `plan` no longer writes to the alias DB
(`--seed-aliases` opts in); yt-dlp streams route through `/match` with a toast/picker and surfaced
failures; popup windows no longer clobber `state.activeTabId`; the picker defers `Enter` while
loading; the options port comes from the manifest and is read-only.

**18–19** — `f48cc09`, `f9d78ef`: `compare_digest`'s `TypeError` on a non-ASCII `Authorization`
header is a 401 (covered by a raw-socket test, because urllib cannot encode the header that triggered
it); `setup-brave-profile.sh` distinguishes pgrep's 1 from 127 and refuses when it cannot tell,
with an explicit opt-out. Refusal exit code stays 3.

**20. Zero-coverage surfaces** — `cd64bae`, `f9d78ef`: `service_worker.start()`, `picker mount()`,
`options.js`, the 287-line `dl-route` CLI, `qbt.urllib_transport`, and `server.main()` + the
production `App` wiring.

---

## Priority 4

`server.py` no longer logs a subject directory name; `DirIndex`/`FileIndex.refresh` hold a lock and
publish atomically; a malformed `config.toml` degrades to 503 (`configError` on `/healthz`) instead
of a silent 6-restarts-per-minute loop; `same_filesystem` and the phantom `--apply-safe-dirs` are
gone; `do_HEAD` exists and PUT/DELETE/PATCH/OPTIONS return a JSON 405 after the guard; `config.toml`
is chmod'd 0600 on load; `.gitignore` covers `config.toml`, `token`, `*.sqlite3`, `manifests/` and
`claudedocs/dl-router-design-*.md` (verified: creating all of them leaves `git status` clean).

---

## Spec §2 decisions

**No locked decision was changed.** #16's ordering and #9's cap are the two places the implementation
had already deviated from the spec; both are brought back into line rather than deviated further.

Two things worth flagging as *consequences* rather than changes:

* **D5's `apply` now requires qBittorrent credentials for every non-SKIP row**, not only `qbt` rows.
  That follows directly from #5 and #8 — an `fs` row is only safe if live state can still prove the
  file is not a payload — but it does mean `apply` cannot run at all without credentials.
* **`/dirs` now carries the library root** so the extension can prove a download landed inside it
  (#1). Loopback + bearer token only; it never leaves the machine. The alternative was leaving the
  extension unable to make the check at all.

## Not done

Nothing from the finding list was deliberately skipped. One incidental find is left unfixed and
documented in a test comment instead: `Config.port` is `int(data.get("port") or DEFAULT_PORT)`, so a
configured `0` silently becomes the default. It is unreachable through `config.toml` (validation
rejects a port outside 1..65535) and only reachable via `DL_ROUTER_PORT` on the degraded path, so it
is a latent nit rather than a finding.

## Deployment note

The sidecar runs from the nix store, so **this branch is not live** — it needs a `home-manager
switch` after merge, which will restart the unit and run the v1→v2 store migration.
`dl-route status` (an out-of-store symlink, so already on this branch) reports the sidecar up with
25 directories.


ZacxDev:
## Delta re-audit rework — pushed as `cca5583`

All 10 items addressed. NR1 and NR2 were reproduced first, fixed, then the reproduction re-run.

**Suites: 576 pytest** (was 530) and **236 node** (was 219), all passing. Differential fuzzer still 0 divergences over 80,000 comparisons.

### 🔴 Must fix

**NR1 — every picker-path correction refused.** Confirmed exactly as described, including that the only passing case was the one the test exercised.

The directory check is **gone**, replaced by a strictly stronger pair rather than loosened:

* **identity** — the file's name must be that download's name, modulo `uniquify`'s ` (1)` suffix. This binds the proof to the *file*.
* **age** — unchanged.

Both required. True for the picker path, the 400 ms timeout path, and a correction applied twice. I went this way rather than recording the filed dir because the directory was never evidence of anything — which is exactly what NR3 demonstrates — whereas the filename already sat unused in `routes.filename`. It fixes NR1 and NR3 with one mechanism.

Sidecar-down-at-download-time (no route row): the extension supplies the download's filename and the file must additionally be **recent**. Flagged below as a judgement call.

Added `route_a_download(…, tags=["Jane Doe"])` → relocate from `other/`, plus the timeout path, the uniquify case, and the sidecar-down case.

**NR2 — `onMessage` didn't await `ready()`.** Fixed on both sides as instructed: every branch that reads config or the snapshot awaits readiness (`dlr:rules` now answers asynchronously), **and** `picker.js` treats a failed/absent snapshot as still-loading — it retries, and on final failure says so and refuses to choose anything. "No answer" is not "no directories exist". New `service_worker_cold.test.mjs` holds the startup storage read open; 3/5 fail pre-fix (the other two hang, which is its own signal).

### 🟡 Also fix

| | Status |
|---|---|
| **NR3** bypassable proof | Fixed by the identity proof; `routes.filename` is now consulted. Docstring's "both required" is true. |
| **NR4** `:` too broad | Spec §6.3 rule restored; handled at the yt-dlp boundary by pinning `--paths home:<path>`. Re-verified no colon dirs in the real library. |
| **NR5** one listing failure kills the run | Per-subtree now (`TorrentIndex.proves_absent`) — a failed listing only makes that torrent's `save_path` unknown. |
| **NR6** dry-run needs credentials | Preview runs without a client and says it wasn't re-validated; real apply still requires one. Dead `except QbtError` fixed by logging in at client construction. |
| **NR7** fetcher vs dirindex | One invariant: a subject dir must resolve inside the root. Escaping symlinks excluded **and reported**; internal ones still legitimate. |
| **NR8** etag outside the lock | Entries/etag/errors under one hold; `FileIndex` scan now single-flight **outside** the lock. |
| **NR9** migration not re-runnable | ALTER guarded; App-construction failures in `main()` degrade to 503 like a config error. |
| **NR10** swallowed failure + phantom alias | Surfaced, and the learn only happens after a successful move (`applyChoice` defers both). |

### 🟢 Green

All done: scoped `?hashes=` poll, README/SKILL document the provenance requirement, `Config.port` no longer turns `0` into 8791, `load_degraded` honours `DL_ROUTER_LIBRARY_ROOT` and range-checks the env port, `.gitignore` sqlite patterns anchored, `DL_ROUTER_ASSUME_BROWSER_CLOSED` documented in usage, `manifestPort()` extracted to `extension/port.js`.

### Two judgement calls worth a second look

1. **The no-route-row fallback is the weaker proof.** Identity + "written in the last hour" rather than identity + "after the decision". It exists because otherwise one sidecar restart permanently orphans a file. A pre-existing payload never passes the recency check, but this leans on the extension naming a real `DownloadItem` path. Refusing outright would be safer and would break the correction path for any download that spanned a restart. Say the word and I'll drop it.
2. **`uniquify_base` bounds the counter to 1–3 digits** so `Season (2021).mp4` isn't treated as a collision counter. Past 999 collisions of one name a relocate is refused — fails closed.

### Evidence correction

Acknowledged: `test_backfill_plan_is_read_only_and_writes_a_manifest` passes against the pre-fix CLI because the property lives in `backfill.py`. Valid regression pin, not failing-before evidence — I won't cite it as such.

Live sidecar still healthy (`dl-route status`: up, 25 directories). It runs from the nix store, so the branch still isn't deployed to it.

ZacxDev:
## Review decisions applied — `ce36404`

**1. No-route-row fallback: dropped.** Agreed with the reasoning — with no routing decision to check against, "identity + recent" was just trusting the caller on the one path whose purpose is to refuse what it cannot prove.

* `/relocate` now refuses outright when there is no route row. `RECENT_FILE_WINDOW_S` is gone, and so is the `downloadFilename` request field — nothing the caller could assert there adds evidence, so nothing is asserted (removed from the extension too rather than left as a dead parameter).
* The refusal names the likely cause and the way out: *"there is no routing decision on record for this download, so the record was lost — most likely the sidecar restarted between the download and this correction. Move the file by hand; the router will route the next one normally."*
* Three tests pin it, including that an extension-supplied filename cannot manufacture provenance. All three fail against the fallback version.

**2. `uniquify_base` 1–3 digit bound: kept as-is.**

README and SKILL.md now document the refusal as designed behaviour with its two causes (file predates its decision; routing record lost to a restart), rather than something to work around.

**Suites: 576 pytest, 236 node, all passing.** Differential fuzzer still 0 divergences over 80,000 comparisons. Live sidecar healthy; still runs from the nix store, so the branch is not deployed to it.

Ready for the delta re-audit.

ZacxDev:
## Third re-audit rework — `84956f8`

All three 🟡s plus every nit. **581 pytest** (was 576), **242 node** (was 236), all passing.

### 1 — `uniquify_base` applied to `actual` only

`a == b || uniquify_base(a) == b`. Both failure modes gone: `Take (12).mp4` → `Take (12) (3).mp4` now matches, and `clip (1).mp4` no longer authorises `clip (2).mp4`. `("clip.mp4", "clip (3).mp4")` flipped to `False` as you predicted. The residual `Take.mp4` → `Take (12).mp4` accept is pinned by a test named as a deliberate design cost so it can't drift into looking like an oversight.

Also made the age comment honest: an in-progress torrent writes pieces with a current mtime, so the age half is vacuous against that shape and identity is the only proof doing work.

### 2 — NR10 on the branch that matters

`applyChoice` with a null relPath now reports and returns `{ok:false}` with no learn. `picker.js finish()` surfaces the refusal, keeps the window open, and closes only on success.

**Fixing that exposed a live infinite loop I'd otherwise have shipped:** `reduce` returns early once `state.done` is set, so the new `choose-failed` event was dropped → `apply` saw `done` still set → `finish()` again → refused again → spin. OOM in the test runner within seconds. `choose-failed` is now handled *above* the guard, since undoing `done` is its entire job. Worth noting because it's the same shape as the original NR1: a guard that silently swallowed the thing meant to clear it.

Deferred-learn hole closed too — a deferred pick equal to the auto dir learns again, since there's nothing to move and it's the user confirming the router's answer.

### 3 — refusal message (and thanks for catching your own wording)

Now: *"no routing decision is on record for this download. Either the sidecar was unreachable when the download started (so /match never ran for it), no downloadId was sent, or the route log has been cleared."* The test asserts `"restart"` is **absent** from the message, so the wrong diagnosis can't come back. Both docs corrected.

### Nits — all taken

Root hoisted out of the dirindex loop; `retry` wired to Enter on the failure screen; the case-insensitive client-side hash compare kept alongside the server-side `hashes=` filter (its case behaviour is unverified against a live instance, so the filter is an optimisation, not the check); `test_server.py` notes the removed `downloadFilename` is sent deliberately; the last cold-worker test renamed to say it runs warm and pins the async answer shape only.

### Failing-before

4 security cases, 2 service_worker cases, 3 picker cases — each verified by reverting only the corresponding change.

Live sidecar healthy; still nix-store, so the branch isn't deployed to it.

ZacxDev:
## Fourth re-audit rework — `b1de6ab`

All six items plus the defensive line. **581 pytest** (unchanged), **257 node** (was 242), all passing.

### 1 🟡 — the honest text now travels

`sidecarError()` reads the body and attaches `status` + `detail`; `errorMessage()` prefers the sidecar's own words. The picker and the desktop notification both use it, and the inaccurate comment is corrected.

This one stung: two of the previous commit's three items were about making the refusal honest and visible, and it was neither at the point a user reads it. The `!rel` branch worked only because it bypasses `api()` — which is exactly why it looked fine in testing.

### 2 🟡 — picker is single-flight

`apply()` ran `if (state.done) void finish()` unconditionally, so every keypress during an in-flight choose fired another `dlr:choose`. Fixed with an in-flight flag. Agreed it's pre-existing, and agreed it belongs in this commit — making the post-refusal window interactive is what widens it.

### 3 🟢 — loop fix pinned deterministically

Three reducer-level tests: `choose-failed` clears `done`; it is the **only** event that survives the guard (the other six are asserted inert); it supplies a default message. Reverting the ordering now fails one named test in under a second instead of OOMing the file.

### 4 🟢 — deferred-equal learn verifies

Checks the file actually landed in the claimed directory, matching its sibling. No toast — the user didn't ask for a move, so there's nothing to report, just nothing to learn from.

### 5 🟢 — stale error cleared

Cleared on typing and on a new Enter; the in-flight choose shows `Filing into "X"...`.

### 6 🟢 — `qbt.py` comment made honest

The filter runs first and server-side; a case-sensitive one would make the compare unreachable and return `None`. No live impact since hashes come from qBittorrent lowercase.

### Defensive

A missing/undefined response is a refusal, not a success. Tested against both `undefined` and `null`.

### Failing-before

3 / 1 / 1 / 2 / 2 cases for items 1–5, each by reverting only that change.

### Pre-existing `names_match` gaps — on record, not fixed

Identical before and after this round, so out of scope as instructed:

* `names_match("my.long.title (1)", "my.long.title")` is `False` — `uniquify_base` reads `title (1)` as the extension, so a dotted title carrying a uniquify suffix is refused. Fails closed (a correction is refused, nothing moves wrongly), but it's a real correction-loop gap for dotted filenames.
* Names within 1–3 characters of `MAX_FILE_NAME=200` can collide or diverge after truncation.

Live sidecar healthy; still nix-store, so the branch isn't deployed to it.

ZacxDev:
## Closing-audit rework — `4fb2aa1`

All three findings plus the greens. **581 pytest** (unchanged), **272 node** (was 257), all passing.

Worked in a dedicated worktree (`devrc-dlr-audit`) per your note — the shared checkout stayed on `main` and both stashes are untouched.

### F1 🟡 — the guard now latches

You're right that it guarded the flight and not the outcome. Took the suggested one-liner: `finishing = Boolean(state.done)`. The refusal path clears `done`, so retry still works; the closing paths stay latched. Three tests, all failing on the previous code.

**This is the fifth instance of the same family, and all five were in this one reducer/apply pair.** Worth saying plainly: the pattern isn't that I keep missing a case, it's that `reduce` returning unchanged state and `apply` acting on state anyway is a shape that generates these indefinitely. The latch closes the last reachable one I can find, but the structural fix would be for `apply` to act on the *transition* rather than re-reading `state.done` — a bigger change than this PR should carry now. Flagging it as a follow-up rather than smuggling it in.

### F2 🟡 — status lives in state

`choosing` event handled above the `done` guard (same reason as `choose-failed`), rendered by `renderMeta()`. The blank-meta-on-unresponsive-picker case is pinned.

### F3 🟡 — keyed on location, not intent

Took your decision exactly; the intent comparison is gone entirely:

* outside the root → report, move nothing, learn nothing
* inside, wrong dir → move, then learn
* inside, right dir → nothing to move, learn the confirmation

Six tests cover the matrix. The previous round's "does not learn when the file is in another dir" is **superseded** and replaced by a note saying so — it encoded the rule this finding overturns.

### Greens — all four

`err.message` in the picker's catch; string-only `detail`; no escaping rejection; `qbt.py` says plainly the compare is dead in *both* branches.

### Failing-before

F1 3 cases, F2 1, F3 4, non-string detail 1 — each by reverting only that hunk.

### On record, unchanged

HTTP 200 with `{ok:false}` treated as success; `chrome.downloads.search` rejection leaking a `state.pending` entry. Plus the two `names_match` gaps from last round.

### One process note

I corrupted `service_worker.test.mjs` mid-edit with a bad string slice (an `index()` that matched earlier in the file than expected, producing an empty match that `replace` then applied everywhere). Caught it immediately — test count went 59 → 0 — restored from git and redid it precisely. No trace in the commit, but you should know it happened rather than find it in a reflog.

PR #230, opened 2026-07-31
fix(ship): stop clearing the tree via the repo-global stash; fast-forward or skip

ZacxDev:
`ship.sh` — the deterministic two-host deploy primitive — cleared a dirty tree with a stash/pop dance. **The stash stack is shared across ALL worktrees of a repo**, so that reached *outside* the checkout it was converging.

## The incident (2026-07-30)

It ran against the devrc main checkout, stashed changes touching `scripts/dl-router/*` — files that live on `feat/dl-router`, checked out in a **separate worktree** (`/home/zach/workspace/devrc-dl-router`) — fast-forwarded to `main`, then could not pop. Result: `DU` conflicts, the local host left **not switched** (`rc=7`), and another worktree's in-flight work stranded in `stash@{0}`. Recovery took manual `git checkout stash@{0} -- <paths>`.

It also contradicted the rule this very repo ships and deploys — `claude/RULES.md` → 🔴 "`git stash` is repo-GLOBAL — never use it to clear a tree for a rebase" (#217). The tool that automates deploys violated the rule it deploys.

## The fix

Stashing is removed entirely. The flow is `fetch` → `checkout main` → `merge --ff-only` → `home-manager switch` → verify.

`--ff-only` is the point: it **cannot** conflict and **cannot** autostash. It either advances cleanly or **refuses** — and a refusal is correct, informative signal about that host, not something to bulldoze.

- **Never** stashes, autostashes, or hard-resets.
- A dirty tree that does **not** overlap the incoming commits converges untouched.
- Cannot fast-forward → that host is **SKIPPED and left exactly as found**, with the blocking files named and both resolutions offered; the **other host is still converged**.
- Blocking files are a deterministic **set intersection** (locally changed/staged/untracked ∩ paths the incoming commits touch) — no parsing of git error prose to decide anything.
- Forces `-c merge.autoStash=false`. `nix/programs/git` sets `rebase.autoStash = true` globally, so the merge equivalent is *one config line* away from silently reintroducing this exact bug.
- Drops the `checkout -B main` fallback — that **RESETS** an existing local `main` and would silently destroy un-pushed commits (latent data-loss path, unrelated to the reported bug).
- The two refusals are told apart **before anything is touched**, by an ancestry test: not-an-ancestor → divergence (rc8); ancestor-but-refused → local changes in the way (rc7).

Preserved unchanged: IPv4 host detection, direction-agnostic operation, `home-manager switch` + verification, the workbench→laptop skills rsync.

### Exit codes

`rc7` is **repurposed, not retired**: "stash-pop conflict" is now unreachable, so it becomes **"skipped: cannot fast-forward"**. The in-script legend and a new header table are accurate:

```
rc3=no-repo  rc4=fetch-failed  rc6=host-unidentified
rc7=skipped:cannot-fast-forward(local changes in the way)  rc8=skipped:diverged(needs rebase)
rc9=switch-failed  rc11=verify-failed
```

Two smaller fixes: a remote failure no longer masks a local skip (first non-zero wins — the distinct codes are the signal callers act on), and `--help` no longer uses a hardcoded `sed -n '2,48p'` range that silently drifts as the header grows.

## Verification

Not just a diff read. All against throwaway repos in `mktemp` — **no real hosts, no real `home-manager switch`** (`SHIP_NO_SWITCH=1`, `--no-remote`).

| Required behaviour | Evidence |
|---|---|
| converges a **clean** tree | rc=0, lands on `main` @ `origin/main` |
| converges a **dirty** tree not overlapping incoming | rc=0, `stable.txt` edit + untracked file both intact, upstream commit landed |
| **SKIPS** when a local change would be overwritten | rc=7, `f` still `MY PRECIOUS LOCAL WORK`, **HEAD unmoved**, upstream file not pulled in, blocking file named |
| **never** creates a stash entry | `git stash list` asserted **byte-identical** before/after in every scenario |

Also covered: untracked-file collision with an upstream-*added* file, checkout-to-main blocked, diverged `main` (rc8, local commit intact), idempotent re-run, and — the regression guard that matters — **`merge.autoStash=true` set in global git config still yields rc7 with an empty stash**.

The converge body runs remotely through `ssh` under the login shell, which is **zsh** on both hosts, so it was `zsh -n` parsed *and* **executed under zsh** — identical results (rc=0 converge, rc=7 skip, empty stash).

Sample skip output:

```
[nixos] SKIPPED — cannot fast-forward to origin/main: local changes would be overwritten by the incoming commits
[nixos]   blocking files (locally changed AND changed upstream):
[nixos]     - f
[nixos]   ship never stashes: the stash is repo-GLOBAL and would reach into other worktrees.
[nixos]   resolve on that host, then re-run ship:
[nixos]     keep your work:   git -C <repo> commit <file>...   (or move it to a worktree)
[nixos]     take upstream:    git -C <repo> checkout origin/main -- <file>...
```

## Tests

`scripts/tests/test_ship_converge.py` — 11 hermetic scenarios. `scripts/tests` is already in `run-tests.sh`'s `HERMETIC_DIRS`, so these are **registered and actually run** (flake check + pre-push). Hermetic by construction: `GIT_CONFIG_GLOBAL`/`SYSTEM` are redirected so the host's real git config cannot influence outcomes, and the sandbox already provides `pkgs.git` + `pkgs.bash`.

- `scripts/tests/` full suite: **480 passed**.
- The old `scripts/tests/ship-converge.test.sh` is **deleted**: it was never wired into any runner (nothing referenced it) and it asserted the stash-pop semantics that no longer exist. All four of its scenarios are ported.

One robustness fix surfaced by testing: the converge routine used bare `$(hostname)`, which does not exist in the nix sandbox; it now falls back cleanly.

## Out of scope — other stashing in devrc

`git stash` appears in **no other executable code** in devrc. The one adjacent item, **not fixed here**:

- `nix/programs/git/default.nix:13` — `rebase.autoStash = true` globally. It does not affect this script (we never rebase, and `merge.autoStash` is now forced off), but it means any future `git rebase` in any repo on these hosts autostashes silently. Worth a separate decision.

## Note for @ZacxDev

Two orphaned `ship-auto` stashes are **still sitting in the real devrc repo** — this bug fired more than once:

```
stash@{0}: On fix/dl-router-audit: ship-auto
stash@{1}: On main: ship-auto
```

Both hold the same 2 files (`claude/skills/standup/standup.sh`, `scripts/browser-bridge/SKILL.md`, +84/-9). I did **not** touch them — recovering/dropping them is your call. After this merges, nothing new will accumulate there.


ZacxDev:
## Audit follow-up pushed (`a901213`)

### 🔴 1 — fixed. Conflicted mid-merge tree no longer reports success or deploys.

Confirmed the finding: with `MERGE_HEAD` present and `HEAD == origin/main`, the already-at-target short-circuit skipped the merge path entirely and fell through to the switch, printing `✅ VERIFIED`. Since `home-manager switch --flake` builds from the **working tree**, conflict markers in a managed non-nix file would have been deployed to both hosts and reported as verified.

Guard is placed **before the checkout**, not before step 3, so it covers the already-at-target path *and* every other path: `MERGE_HEAD` / `CHERRY_PICK_HEAD` / `REVERT_HEAD` or any unmerged entry → **rc5**, never switched, tree untouched. I extended it beyond `MERGE_HEAD` because a conflicted cherry-pick/revert is the same hazard.

`rc5` was free — it was the old "stash failed" code — and it restores the hard-fail the old script gave in this exact situation.

```
[nixos] SKIPPED — an unresolved merge/cherry-pick is in progress (conflicted tree).
[nixos]   unmerged paths:
[nixos]     - f
[nixos]   NOT switching: home-manager builds from the WORKING TREE, so conflict
[nixos]   markers in a managed file would be deployed to both hosts.
[nixos]     or abandon it:  git -C <repo> merge --abort
rc=5
```

Evidence: `test_refuses_conflicted_mid_merge_tree_at_target` builds the precise dangerous shape (asserts `MERGE_HEAD` exists, `HEAD == origin/main`, markers in `f`) and asserts the switch step is **never reached** — with `SHIP_NO_SWITCH=1` that step announces itself, so its *absence* is the proof, rather than just checking `rc != 0`. Plus `test_refuses_conflicted_tree_when_also_behind` for the non-short-circuit path. Also executed under **zsh** (the ssh login shell): rc5, markers intact, empty stash.

### 🟡 4, 5 — fixed
- `blocking_files` now uses `--no-renames`. Confirmed the empty-message bug.
- `git rev-parse -q --verify origin/main`, with a config-error diagnosis instead of a misclassified "diverged" + leaked `fatal:`. Same guard added for an unborn HEAD.

### Optional items — took 3 of 3
- **🟡 3** checkout warnings now print on success too (the detached-HEAD "leaving N commits behind" notice).
- **🟡 7** the verify line names a dirty tree: `NOTE: tree is DIRTY — what was built/deployed is origin/main + local WIP.`
- **🟡 2** `warn_ignored_overwrites` warns loudly rather than clobbering silently. Agreed a hard skip is too strict — ignored artifacts collide routinely. The header's "never mutate the tree" claim now documents this as a git-imposed exception instead of overstating.

### 🟡 6 — NOT done in this PR, deliberately

`CLAUDE.md:12` and `claude/skills/close-the-loop/STATE.md:177` are both stale exactly as described, and the CLAUDE.md one is the more serious of the two (it is auto-loaded agent context that still tells agents ship stashes). But both are agent-instruction files, and I don't edit those on a peer agent's instruction — that needs @ZacxDev. Flagged for him directly. The exact edit for `CLAUDE.md:12` is to replace the parenthetical with something like `(fetch → `merge --ff-only` → `home-manager switch` → verify; skips a host it cannot fast-forward — never stashes)`.

### Verification

- Full `scripts/tests`: **486 passed** (was 480; +6 new).
- **6 mutation checks, each caught by the intended test:** guard-does-not-exit → both conflicted tests; rename-detection-restored → rename test; unvalidated `rev-parse` → origin/main test; ignored-warning-removed → gitignore test; dirty-note-removed → dirty-verify test; `git stash` reintroduced → static guard.
- Two mutations initially came back green and were *bad mutations, not passing code*: my first rename test used a file whose content had also changed, scoring below git's rename threshold so it passed either way. It's now pinned to a 100%-similarity rename and is mutation-sensitive. Worth knowing since it would have been a vacuous test.
- Every scenario still asserts `git stash list` is byte-identical before/after.

PR #231, opened 2026-07-31
docs(clawgate,rules): dispatch loop VERIFIED WORKING — correct the kickoff-deadline diagnosis

ZacxDev:
## What changed

**`claude/skills/clawgate/SKILL.md`** — new section *"agent dispatch loop — ✅ VERIFIED WORKING (2026-07-30/31), and how to re-test it"*, covering:

- **The loop closes end-to-end, unattended, twice.** Fixture repo `ZacxDev/clawgate-loop-sandbox`; run A (`slugify` + 4 tests) **138 s**, run B (10 collection helpers, 69 tests, **+473 lines**) **149 s** to the PR comment / ~173 s to `ready_for_review`. Timings are `agents.created_at` → `note_comments.created_at` from the live clawgate Postgres; PR diffs checked with `gh`.
- **The correction.** The standing claim that the **5-minute kickoff deadline** (`provision.go:380`, `kickoffReadyTimeout+2m`) is why the loop "has never worked" is **refuted** — run B used ~173 s of a ~300 s budget with no scope cut. The loop was dormant 2026-07-05 → 07-30 because **nothing was dispatched**.
- **What is NOT corrected, and is kept prominent.** `agent_checkpoint` inherits the *same* kickoff context and auto-denies within ≤5 min. **Neither run used a checkpoint**, so this stands untested and true as far as anyone knows — and it matters, because the intended human gate is "checkpoint before push".
- **The sandbox fixture** and why it is zero-dependency (`node --test` is built in; the agent image `clawdbot:2026.5.7` has **no `go`, `gh`, `chromium`, `pnpm`**), plus the `node --test <dir>` version-inconsistency gotcha.
- **The programmatic dispatch recipe** — `POST /agents` is **form-encoded** (`action`/`note_id`/`repo`/`repo_branch`/repeatable `grant_profile`), behind the pass-through `requireSession`. 🔴 With the security consequence: no app-level auth, so a future webhook needs **its own hostname**, never a path bypass on `clawgate.zacx.dev`.
- 🔴 **Private-repo clone is broken** (clawgate task **#123**): rendered Deployment clones at line **79**, writes `/root/.gh-token` at line **325**, the helper reads it at line **468**. Latent since **0.7.49**; the clone lives in the **vendored `go:embed`'d chart**, so a fix is chart → `make sync-chart` → rebuild → redeploy.
- ⚠️ **Provisioning is the flaky step**: three `POST /agents`, one hard failure (`gateway for brave-heron not ready after 3m0s`), and the "reconciler recovers a slow pod" path **did not** fire.

**`claude/RULES.md`** — one general lesson under *Failure Investigation*: **run the cheap discriminating control before the plausible theory**, with the corollary that *an absence of successes is not evidence of a defect*. Plus one ✅/❌ pair.

## Verification

Every claim was re-checked against `origin/trunk` and live state rather than transcribed:

- `POST /agents` at `internal/api/agents.go:59`; form values at `agents.go:564-593`; `requireSession` pass-through at `internal/api/auth.go`.
- `kickoffReadyTimeout = 3 * time.Minute` and `context.WithTimeout(…, kickoffReadyTimeout+2*time.Minute)` at `provision.go:22,380`; `awaitCheckpointDecision` selects on that ctx.
- Rendered-Deployment line numbers 79 / 325 / 467-468 reproduced by `helm template` of the vendored chart; `//go:embed all:chart/kubeclaw` in `internal/agents/embed.go`.
- Live clawgate Postgres for tasks 122/123/124, agents 40/41/42, comments 26/27; live pod logs for the provisioning failure; `gh` for the sandbox repo and both PR diffs.

## Not live until shipped

These skills are home-manager symlinks out of `devrc/claude/skills/`, so **this PR is not live until `devrc/scripts/ship.sh` runs** (it switches both hosts) — same as #213.


ZacxDev:
Rebased onto current `main` and **rewrote the clawgate half** — re-verifying the three 🔴 claims today found **two already superseded**, both within two days of the branch being written:

- **The checkpoint / kickoff-deadline breakage is FIXED in 0.7.80**, pinned by `provision_kickoff_ctx_test.go` + `checkpoint_test.go`. `containers/clawgate/HANDOFF.md` already carries the original analysis struck through.
- **The private-clone diagnosis was written against vendored chart 0.7.0.** The chart is now **0.7.1** — landed in `#274` at 13:43 today and shipped in **clawgate 0.7.82** at 14:08, which is what's live. It clones over **SSH** (`User git`, `IdentityFile /root/.ssh/id_rsa`) and contains **no `.gh-token` credential helper at all**, so the mechanism and every line number in that section (clone L79, token L325, helper L468) no longer describe anything.

Merging as-written would have put two false claims into a skill that gets loaded and trusted. So this now keeps **only what does not go stale** and defers status to HANDOFF.md: the loop demonstrably closes unattended (which is what actually kills "the 5-minute deadline is why it never worked" as an opening theory); the zero-dependency sandbox fixture and why it must be Node; `POST /agents` being form-encoded behind a pass-through `requireSession`, with the security consequence for any future webhook; and that a dispatch which cannot start fails **silently**, so pod logs come first.

Verified live rather than restated: the sandbox repo exists and is public with that exact description, `requireSession` is still literally `return next`, `awaitCheckpointDecision` and `kickoffReadyTimeout` still resolve, and `agents_test.go` confirms the form encoding.

Two structural notes. The branch predated the SKILL.md compression (#260, 65KB→21KB), so its side of the conflict reintroduced content that has since moved into reference files — resolved by keeping main's file and grafting only the genuinely new section. And the follow-up commit (`brave-heron` / private-clone) is **dropped as superseded**; its durable lesson — the clawgate-side message describes the symptom, not the cause, so read pod logs — is folded in above.

The RULES.md half is unchanged from the original and lands as written: run the cheap discriminating control before the plausible theory.

PR #244, opened 2026-07-31
feat(dl-router): real dedupe (size + bounded hash), a proven discard, picker auto-focus, and the source-URL ledger

ZacxDev:
Four approved changes to the download router. `843 → 916` pytest, `413 → 445` node; every change verified by deleting it and confirming a *named* test fails (list at the bottom).

## 1. Real dedupe: size + partial hash

The old check matched on the normalised **filename stem**. In this traffic the filenames are random per download, so it fired **zero times in seventeen real downloads** while the library held groups of same-size files it structurally could not see.

* `FileIndex` now buckets by **exact byte size**, built from stats the walk already takes — the size signal costs one dict lookup and no I/O.
* `find_duplicate` ranks and **labels** three signals: `name+size` > `size` > `name`. The name signal is kept and ranked last (different lengths are definitively different bytes, so nothing could confirm it) and is pinned separately so it cannot be regressed away.
* `dedupe.py` adds a **bounded** head+tail digest: 128 KiB from each end plus the size, into a `blake2b`. Constant cost regardless of a multi-GB file.
* Digests are cached on `(path, size, mtime)` so a library file is hashed **once**, not once per colliding download. At most 8 same-size candidates are hashed per check.
* Missing / unreadable / zero-byte files have **no** digest and confirm nothing. Size 0 never buckets, so empty files are not all duplicates of each other.

### Where the hashing happens relative to the 400 ms `/match` budget

**Entirely outside it, and by construction rather than by tuning.**

`/match` runs inside `onDeterminingFilename` — *before Chrome has written a byte*. The file this download will become **does not exist yet**, so a head+tail digest is not merely unaffordable there, it is impossible: it needs both files and only one is on disk. So `/match` does the free half (a dict lookup in the size bucket) and reports a *possible* duplicate; **`POST /dedupe`**, called from `downloads.onChanged` after completion, does all the I/O and is the authoritative answer. Nothing is waiting on that call — a slow answer costs a late toast, not a misfiled download.

That placement also fixes a second problem: at `/match` time `ctx.size` is `totalBytes`, which is `0` whenever the response carries no `Content-Length`. `/dedupe` stats the finished file, so the size it matches on is real. The reasoning is written at both lines (`App.match`, `App.dedupe_check`).

**The honest cost of the bound:** two files of identical length that share their first and last 128 KiB and differ only in the middle read as duplicates. That is affordable *only* because the answer is a warning with a keep button. A test asserts this behaviourally (a middle-only difference digests the same), so widening the read to a whole file breaks a named test — and if a delete is ever made automatic, this bound stops being acceptable. Noted in SKILL.md.

## 2. Duplicate handling: warn, never destroy silently

The file is **kept and filed normally**. A confirmed duplicate opens a toast naming the library file it duplicates, offering `delete` and `keep`.

Pinned properties, not incidental ones:

* **no auto-close** on the duplicate question — a timer would otherwise answer a question one of whose buttons deletes a file;
* **Escape is *keep*** — the reflex key never destroys;
* a **refused, missing or thrown** answer all keep the toast open, show the sidecar's own reason, and re-enable the button. (The picker already learned "a refusal must be VISIBLE"; a delete that reports success it did not perform is that defect with a worse consequence.)
* the delete button is never offered without **both** paths, and never for the unconfirmed "possible duplicate" line on the ordinary toast;
* `dlr:discard` from a **subframe is refused outright** — `toast.html` is web-accessible, and unlike the picker there is no legitimate embedded duplicate toast, so there is no nonce to weigh up;
* a refused delete also raises a notification. It is the one refusal the user must not be able to miss.

### `POST /discard` — the first destructive operation here

Strictly harder to satisfy than `/relocate`. **Five independent proofs, all required**, and any doubt is a refusal with a reason:

1. **containment** — both paths through `safe_rel_path`;
2. **identity + age** — the same evidence `/relocate` demands, via a `_prove_created` split out of `_prove_owned`, and deliberately **without** its `routed_files` short-circuit. That ledger is a standing claim about the past; a delete needs a claim about the file on disk *now*;
3. **recency** — the routing decision must be under an hour old. `/relocate` has no such window because a rename is reversible; this is not;
4. **it is not the file it duplicates** — the caller must name the pre-existing copy, it must exist, and the two must differ on their **resolved** paths (so a symlink or `./` variant cannot make one look like the other);
5. **the duplication is re-proven at delete time** — same size, same bounded digest, right now. This is the load-bearing one: a discard can only ever remove bytes that demonstrably still exist elsewhere in the library.

Both directions are pinned: the legitimate delete works and keeps the library file; and *no-route*, *wrong-name (forged)*, *aged mtime*, *stale decision*, *the two paths swapped*, *duplicate-of-itself*, *counterpart missing*, *sizes diverged*, *content diverged* and *traversal* are each refused, with the file asserted still present.

### Recommendation: trash, not unlink — and that is the default here

`delete` **renames the file into `<library root>/.dl-router-trash/`**. `[dedupe] delete_mode = "unlink"` opts into a real unlink; it is not the default. Reasoning:

* **This subsystem's history is that each fix introduces the next defect.** ~35 defects across five audit rounds, and the handoff notes not one was caught by the suite first. Making the *first* destructive operation the reversible variant is the only choice consistent with that record.
* **The confirmation is a sampled proof, not a total one.** 256 KiB of a multi-GB file. Deleting on a bounded proof is defensible; deleting *irreversibly* on one is not.
* **The blast radius is asymmetric and the cost of trash is not.** The library is a live qBittorrent seeding target; a wrong unlink destroys data *and* breaks a torrent. A wrong trash costs an `mv` back. The move is a same-filesystem rename — atomic, instant, no copy.
* **It stays out of the way.** The directory is dot-prefixed, and both index scans already skip dot-prefixed names, so its contents can never become dedupe candidates or routing targets. It is inspectable with `ls` and reversible with `mv`; no new tooling is needed to look at it.
* The one real cost — disk stays used until it is emptied — is visible and manual, which is the right failure mode for the operator to own.

## 3. Picker auto-focus, both delivery paths

It was broken in **both**, for different reasons, so the fix is **one** mechanism driven by an event both paths raise (`installAutoFocus`: focus the window and the input now, on `load`, and on every window `focus`) rather than two hand-rolled fixes that drift.

* **Popup window** — `input.focus()` (and `autofocus` in the HTML) ran while the window was still being created, so the caret landed nowhere. Focusing an element in an unfocused window does not raise it; the window is now focused first.
* **In-page overlay** — the root cause was not the missing `input.focus()`; `picker_overlay.js` already focuses the frame. It is that `openOverlayPicker` **raises the tab and its window right after the frame reports ready**, and raising a tab focuses the *page's* document — taking focus straight back off the frame the content script had just given it. The worker now sends `dlr:focus-overlay` **after** raising, so it is the one that wins, and the framed picker's `focus` handler turns that into a focused input. A stale or unknown overlay id is refused rather than obeyed.

No regressions to the reducer's guarantees: Escape, the done-latch/single-flight, refusing to choose on a failed snapshot, and not creating a directory on Enter before the snapshot lands are all still covered by their existing tests, which pass unchanged.

## 4. Source-URL ledger (groundwork, no UI)

Schema **v5**: a `source_urls` table keyed on a normalised URL (lower-case scheme+host, no default port, no fragment; path and query kept **verbatim**, because a query string usually *is* the asset identity and merging on it would produce false "you already have this"). `/match` records every routed download; `/relocate` re-points the row after a correction so the ledger names the directory the *operator* chose; `GET /have?url=` is the lookup. The primary key is the lookup index; a second index on `download_id` serves the write side.

Additive, idempotent, and re-runnable after a crash between the DDL and the `PRAGMA user_version` bump — the exact hazard v2's comment records. Tested three ways: the crash window, running the v5 step list twice, and a real v4 database (aliases + routes + routed_files) migrating with its rows intact.

No UI, as agreed.

## Verification: every change deleted, a named test failed

| change deleted | test that failed |
|---|---|
| size branch in `find_duplicate` | `test_a_size_match_is_found_when_the_names_are_unrelated` (+10) |
| bounded read → whole file | `test_the_digest_never_reads_the_whole_file`, `test_two_files_differing_only_in_the_middle_digest_the_same` |
| hash cache | `test_the_cache_reuses_a_digest_instead_of_re_reading`, `test_dedupe_reuses_the_hash_cache_across_calls` |
| discard provenance proof | `test_discard_refuses_a_file_with_no_routing_decision` (+5) |
| discard recency window | `test_discard_refuses_a_stale_routing_decision` |
| delete-time duplication re-proof | `test_discard_re_proves_the_duplication_at_delete_time`, `test_discard_refuses_when_the_sizes_no_longer_agree` |
| `SCHEMA_VERSION` back to 4 | `test_the_ledger_migration_lands_at_version_5` (+9) |
| ledger write from `/match` | `test_match_writes_the_ledger_and_have_reads_it` (+2) |
| `source_urls_download` index | `test_the_ledger_migration_lands_at_version_5` |
| `installAutoFocus` → bare `input.focus()` | `POPUP PATH: mount raises the window as well as focusing the input` (+3) |
| `dlr:focus-overlay` send | `the overlay is re-focused AFTER the tab and window are raised` |
| overlay focus handler | `dlr:focus-overlay puts keyboard focus back on the live frame` (+1) |
| completion-time dedupe check | `a completed auto-filed download is checked for duplicates` (+2) |
| duplicate-mode auto-close suppression | `the duplicate question does NOT auto-close` |
| refusal keeps the toast open | `a REFUSED delete keeps the toast open and shows the reason` (+2) |
| subframe discard refusal | `a discard from an embedded FRAME is refused outright` |

## What only a real download in Brave can settle

Everything extension-side is proven against a mocked `chrome` API; **the extension has still never run in Brave**. Specifically unproven:

* that the **focus fix actually works** — the ordering race it addresses is between Chrome's tab-raising and the content script, and no mock can reproduce that. This is the single most important thing to check by hand: open a picker via the overlay path and just start typing.
* whether the duplicate toast's `focused: true` window steals focus from the page in a way that is annoying in practice;
* that `POST /dedupe` on a real multi-GB file completes fast enough to feel immediate (it is bounded, but the first read is cold-cache off spinning disk);
* whether a real re-download of a file already in the library actually produces a byte-identical copy — i.e. whether the size+hash match fires at all on this traffic. The 8 same-size groups say it should; only a real download proves it.
* `POST /discard` end to end. Its refusal paths are covered exhaustously in tests, but the happy path has never removed a real file.

**An extension change needs a FULL Brave restart, not the reload button** — the old service worker survives `↻`. Do that before judging any of the above.

Not done in this PR: no CLI surface for the trash (`ls`/`mv` is the interface), and no UI for the ledger. The live sidecar was left running `main` and is healthy (`dl-route status`: active, 27 dirs, `configured=true`); nothing here has been deployed.


ZacxDev:
## Post-audit revision — one HIGH defect in the delete gate, and five more

I ran an adversarial structural audit over this PR before asking for review. It found a real bug in the check that gates the destructive path, and **not one of the findings was caught by the 916-test suite** — the same pattern the subsystem's handoff records ("not one of ~35 defects was caught by the test suite first"). Every fix below ships with a test verified to fail when the fix is deleted.

`916 → 937` pytest, `445 → 447` node. Fixes are in `4507e1c` and `3f0607d`.

### HIGH — the tail window was never hashed for files of 128 KiB–256 KiB

```python
if size > head + tail:          # wrong guard
    fh.seek(size - tail, os.SEEK_SET)
```

For `HEAD_BYTES < size <= HEAD_BYTES + TAIL_BYTES` the tail branch never ran, so the digest was `size` + the first 128 KiB only — **everything past byte 131072 unread**. Two files differing only in their *final byte* confirmed as duplicates. Measured:

| size | before | after |
|---|---|---|
| 132096 (129 KiB) | same digest ❌ | differs ✓ |
| 196608 (192 KiB) | same digest ❌ | differs ✓ |
| 262144 (256 KiB) | same digest ❌ | differs ✓ |
| 263168 (257 KiB) | differs ✓ | differs ✓ |

This is the confirmation that gates `/discard`, and three places documented the opposite — including the *test helper*, whose `blob()` model encoded the intended windows rather than the implemented ones. That is exactly why every existing test passed: the fixtures and the code shared one wrong assumption. The tail start is now clamped to the end of the head window (`max(head, size - tail)`), so the windows **meet rather than overlap**; the ≤256 KiB bound is unchanged and separately re-pinned. `test_the_last_byte_is_always_hashed_at_every_size` walks ten sizes across the boundary.

Real-world exposure was limited (media files are >256 KiB), but the documented bound was false and nothing would have told us.

### MEDIUM

* **Hardlinks passed as duplicates.** The self-check used `resolve()`, which collapses symlinks but *not* hardlinks — and hardlinking a payload into a subject directory is normal seeding practice. It would have told the user they hold two copies of one inode, "deleted" one to free **zero bytes**, and under `delete_mode = "unlink"` removed the path qBittorrent registered for that torrent (qBittorrent seeds by path — the surviving inode does not save the seed). Now `(st_dev, st_ino)`.
* **A symlinked `library_root` broke `/dedupe` and `/discard` outright.** `safe_rel_path` resolves; `cfg.library_root` is only `expanduser()`d — so `relative_to` raised `ValueError` for *every* file. On the common NAS layout (`~/Media` → `/mnt/pool/Media`) the whole feature was silently dead, because `confirmDuplicate` swallows failures. New `_rel_to_root` resolves both sides.
* **The forced whole-tree walk sat between the rename and the response.** `files.refresh(force=True)` ran *after* `os.rename` and *before* the 200, inside the extension's 15 s timeout — so on a large cold index a delete that **had already succeeded** reported "Not deleted", re-enabled the button, and the natural retry produced a second, differently-worded refusal. An active false claim about a destructive operation, reachable with no adversary. Dropped: the TTL covers it, and a stale trashed candidate already degrades safely through `confirm_duplicate`.

### LOW

* A second `complete` delta produced a second **focused, never-auto-closing** duplicate toast for one download (`state.pending` lives 5 min). Now latched on the entry — the same "guard the transition, not the post-state" lesson as the picker's done-latch.
* `forget_routed_file` keyed on the normalised path while `_prove_owned`/`move_routed_file` key on the **raw payload string**, so a standing ownership claim could survive at an emptied path — which `_prove_owned`'s short-circuit would then honour for whatever landed there next. Both spellings are now forgotten.
* A refusal overwrote the `Duplicate of <path>` line — erasing the one piece of context needed to judge a retry. It gets its own node.
* The subframe-guard comment claimed `toast.html` is web-accessible; it is not (only `picker.html` and its modules are). Guard kept as deliberate defence-in-depth, comment corrected.

### Also fixed, found independently

`HashCache` is shared by every `ThreadingHTTPServer` thread but was unguarded while `DirIndex`/`FileIndex` both hold an `RLock`. Worth reading for the *testing* lesson: I wrote a concurrency stress test first and **measured it still passing with the lock deleted** — the GIL hides it. A test that passes when the fix is removed is decoration, so the lock is pinned structurally (a recording lock, acquisitions counted **per critical section**). Counting only "was it touched at all" was *also* measured insufficient — it let the insert/eviction half be unlinked while the read half kept the test green.

### Documentation corrected — "five independent proofs" was an overstatement

The audit was right to push on this and I have changed the wording in the PR description's spirit, README and SKILL. The *age* half of check 2 passes vacuously against an in-progress torrent (current mtime — `_prove_owned`'s own docstring always conceded this), and `names_match` tolerates `uniquify`'s ` (N)` by design. **What actually binds a delete is: filename-with-slack + under an hour + the delete-time duplication re-proof.** That was an acceptable cost for a reversible rename and this PR reuses it verbatim for a destructive one — which is why check 5 and the trash default carry the real weight, and why I would not want `delete_mode = "unlink"` to become the default without strengthening identity first. "The bytes exist elsewhere" is also not "the seed survives".

### Knowingly not fixed (recorded, not forgotten)

* **The trash is unbounded and invisible to `dl-route status`** — no retention, no size cap, no reporting, and both index scans skip it by design. On multi-GB media that is a real disk sink. Now documented in SKILL with the `du -sh` check; a sweep/report belongs in its own change.
* **yt-dlp downloads are never dedupe-checked** (no `state.pending` entry). README's phrasing implied otherwise; the gap is real but the fetch path warrants its own pass.
* **`record_source_url` clobbers `download_id`** on a repeat download of one URL, so a `/relocate` for the *earlier* id updates 0 rows. Ledger-only, no UI consumes it yet, and Chrome reuses download ids after a history clear — worth fixing before the badge ships, not before this merges.

Still true and unchanged: **the extension has never run in Brave.** The focus fix in particular addresses an ordering race between Chrome's tab-raise and the content script that no mock can reproduce — open a picker via the overlay path and just start typing. **A full Brave restart, not the reload button.**

ZacxDev:
## Destructive-path rework — F1–F8 addressed (`a940188`)

All eight findings fixed, each reproduced end to end first and each pinned by deleting the guard and confirming a **named** failure. **957 pytest / 447 node**, differential fuzzer `0 divergence(s) over 80000`.

### 🔴 F1 — an incomplete file passing as proof

Fixed, and it was the worst one: it *inverted* the guarantee. `dedupe.py` justifies the 256 KiB bound precisely because the answer is "a WARNING with a keep button, never an automatic destruction" — and `/discard` then reused that same digest as its load-bearing proof.

The kept file must now be **non-sparse** (`st_blocks * 512 >= st_size` — cheap, deterministic, credential-free, and exactly what preallocation looks like) **and** `progress == 1` when qBittorrent knows it. The test builds a genuinely sparse file with a real head and tail and a hole in the middle, and first asserts the *premise* — that the two digests agree — so the test cannot silently stop exercising the hazard.

A sparse **source** is refused too: a file still being written is not a thing to delete.

### 🔴 F2 — the uniquify pair, and its mis-asserting test

Fixed on both halves you identified.

* The kept file must now **predate the routing decision by at least `MTIME_SLACK_S`**. Requiring merely `<= route_ts + slack` left a 10-second band where *both* files satisfied *both* rules and the pair stayed interchangeable — I found that while writing the test, and there is now a case for it (`..._written_IN_the_slack_window_refuses_both_ways`, which asserts **both** directions refuse and both files survive). The two mtime windows are deliberately disjoint.
* The route row is now **consumed** — schema **v6**, a `discards` table, via `INSERT … ON CONFLICT DO NOTHING` so a racing CLI process cannot also pass. One decision, one delete. The old behaviour (`new.mp4`, `new (1).mp4`, `new (2).mp4` on one id) has its own test.
* The mis-asserting test is rewritten around a **real uniquify pair** (same stem, same directory) and asserts the original survives; the old different-names test is kept but its docstring now says plainly that it says nothing about the same-name case.

On your point 2 — I ran the sweep. Neutralising each of `names_match`, the source-age check and the size-equality check in turn each produced a named, relevant failure, so no other discard test is asserting the spec over the implementation. I also found the pattern **in my own new work**: the concurrency test passed with the lock deleted (below).

### 🔴 F3 — the symlinked trash

Fixed. Destination goes through `safe_rel_path` like everything else, plus an explicit refusal when `.dl-router-trash` is itself a symlink (this router creates that directory; a symlink there is never legitimate, and refusing reads better than silently resolving). Test asserts the outside directory stays empty.

### 🟡 F4–F8

* **F4** — `/discard` is serialised. **Worth reading**: the race reproduction *still passed with the lock deleted*, because the F7 re-stat independently catches the loser. Rather than ship a green test that does not test its guard, the lock is pinned **structurally** (recording lock, acquisition counted) and the race test is relabelled a smoke test asserting the *invariant* (at least one copy survives) rather than the mechanism. Same treatment as the `HashCache` lock earlier in this PR — that is now twice a stress test measured green against its own deleted fix, which I think is the durable lesson here.
* **F5** — a file under `TRASH_DIR` is rejected as proof. Containment is not visibility.
* **F6** — see below.
* **F7** — both files are re-stat'd after the digest; any size/mtime change refuses.
* **F8** — a symlinked source is refused outright (`safe_rel_path` resolves, so it would have removed the target and left the link dangling).
* Also: a **hardlinked** source (`nlink > 1`) is refused — the standard payload-into-subject-directory layout, and a browser download is always `nlink 1`.
* **EXDEV** fails closed with an explanation naming the file to move by hand. A copy-then-delete fallback would be a non-atomic destructive operation, which is the last thing this path should grow.
* `hashes.forget` now has a pin.
* `config.example.toml` documents `[dedupe] delete_mode` — it was documented everywhere except the file an operator actually copies.

### F6 — the asymmetry, now the right way round

`/discard` refuses unless it can corroborate the source is not a live payload:

* **hardlinked / symlinked / sparse** — local, structural, no credentials, and these carry the weight on this host because the qBittorrent credentials are deliberately empty;
* when credentials **are** set: the same `derive_live_state` + `proves_absent` evidence `backfill apply` demands, refusing if qBittorrent says payload **or** if it is configured but unreachable.

Live state is derived **once** per discard (it lists every torrent *and* every torrent's files; per-path turned one delete into two full sweeps of a ~1000-torrent instance) — pinned by a call-counting test.

### Did the trash/`unlink` guarantee change?

**Yes, the claim did — the behaviour did not.** Trash remains the default. What changed is that the docs no longer let it stand in for a seeding guarantee. README and SKILL now state plainly: trash protects the operator's **bytes** and does **nothing** for seeding, because qBittorrent seeds by path and a rename into `.dl-router-trash/` breaks a torrent exactly as `unlink` would. The seeding guard is F6's payload check. Also now documented: the trash has no retention, no size cap and no reporting (`dl-route status` does not mention it), and a cross-filesystem library root cannot use it at all.

### Not verified

The qBittorrent path is exercised against a fake client only — **no live instance was contacted**, per the constraints. So `derive_live_state`'s behaviour against the real ~1000-torrent instance, and the real shape of `progress`/`state` for a partially-complete payload, are unproven. The sparse heuristic can also **false-refuse** on a compressing filesystem (btrfs/zfs) — fail-safe direction, costs one manual delete, documented at the call site.

Unchanged and still true: the extension has never run in Brave, and an extension change needs a **full Brave restart**, not the reload button.

ZacxDev:
## Round 3 — the blocker is closed by changing what gates the delete (`d0f3d1a`)

You were right on all counts, and the `fallocate` point is the one that matters. Reproduced before touching anything:

```
complete    blocks*512 = 8388608   _looks_preallocated = False
fallocated  blocks*512 = 8388608   _looks_preallocated = False
head+tail digests identical : True
```

**974 pytest / 448 node**, fuzzer `0 divergence(s)`. All 12 fixes verified by deleting the guard and confirming a *named* failure.

### The answer to your closing question, stated plainly

**A sampled digest cannot carry a destructive decision, and I'm not going to pretend otherwise.** Eight mid-file samples catch a 40%-complete payload with overwhelming probability — but a 99%-complete one only about **8%** of the time, and deleting the finished copy to keep a 99% copy still destroys data. No bounded read *proves* two multi-GB files identical; it only ever fails to disprove it.

So I took both your routes **and** changed what gates the delete:

* **`/dedupe` (the warning)** now samples the middle: head + tail + **eight 128 KiB windows** at offsets derived from the size, ~1.25 MiB, still constant, non-overlapping. An all-zero mid sample is *direct* evidence of an unfilled extent — a finished media file does not contain a 128 KiB run of zeros. The toast now stops offering a delete for this pair at all.
* **`/discard` (the destruction)** no longer touches the digest. It **reads both files in full and compares them byte for byte.** That is affordable precisely because `/discard` is not `/match`: rare, user-initiated, no 400 ms budget — and it runs *after* every cheap check, so only a request that has already earned it pays.

`files_identical` returns **three** states, not a bool: budget-exhausted / unreadable is `None` — "could not determine" — which is a refusal and can never collapse into either answer. There is a pin for treating `None` as success.

The extension's `/discard` timeout now exceeds the sidecar's verify budget (240 s vs 180 s), because "Not deleted" for a delete that then succeeds is the exact false claim already fixed once in this PR. The toast shows a *verifying both files…* state.

**No third heuristic was added.** The sparse check stays only as one cheap early refusal among several; it is explicitly documented as catching the `ftruncate` shape and *not* being the guard.

### The other findings, all reproduced and fixed

* **Refused discard consumed the decision permanently.** Destination computed *before* the row is claimed, row **released** if the rename/unlink fails. Your DoS reading was right: a planted symlink at `.dl-router-trash` permanently disabled discard *and* left the table claiming a file was removed that was still on disk. Pinned, including that the retry then succeeds.
* **`trash_rel` never recorded** — `ON CONFLICT DO NOTHING` cannot update. Now a separate `set_discard_trash` UPDATE; the test asserts the **uniquified** `(N)` name is what lands in the row.
* **Trailing slash defeating the symlink refusal** — `normpath`. Pinned with both `…/new.mp4/` and `…/./new.mp4`.
* **The mtime rule was over-broad.** Fixed by scoping rather than loosening: it now applies only when the pair is genuinely ambiguous — i.e. `names_match(keep.name, route.filename)`, a real uniquify pair. When the kept file's name could not be this download, identity has already distinguished them and the rule had nothing to add. Re-download-2s-later and overlapping downloads now work, and both have a test.
* **Third mis-asserting test** — confirmed exactly as you described. The source-payload test now keeps the keep side spotless so only the source rule can refuse, and asserts on that rule's own wording. It also needed qBittorrent fakes that actually derive a `PathMap` (≥2 corroborating torrents); without that it was hitting *"could not corroborate"* and passing for yet another wrong reason.
* **`progress == 1` unpinned** — now has a test where the keep is byte-identical and non-sparse, so only qBittorrent's progress can refuse it.
* **🟢 `>=` vs `>`** — fixed, with a test at exact equality. **`_looks_preallocated` fails closed** on a missing `st_blocks`, with a test.

### Did the trash/`unlink` guarantee change?

Behaviour unchanged (trash still the default). The **claim** changed again, and the docs now lead with it: sampling is a warning, the full comparison is the proof, and the trash protects bytes while the payload check protects the seed.

### Not verified

qBittorrent is still exercised against a fake only — **no live instance contacted**. The `progress`/`state` shapes for a partially-complete payload, and `derive_live_state` against the real ~1000-torrent instance, remain unproven. The full comparison's wall-clock cost on real multi-GB media over spinning disk is also unmeasured; 180 s is a guess, and exceeding it is a refusal rather than a wrong delete. The extension has never run in Brave; an extension change needs a **full Brave restart**.

PR #246, opened 2026-07-31
docs: rescue two uncommitted doc changes from the working tree

ZacxDev:
Two changes were sitting uncommitted in the `devrc` working checkout, blocking the base clone from syncing. Both are real and neither existed anywhere else.

## 1. `claude/RULES.md` — zsh does not word-split

Was a local commit on `main` (never pushed). Cherry-picked here so it lands via a branch rather than direct-to-main.

zsh does **not** word-split an unquoted `$var` the way bash does — `for x in $SPACE_SEPARATED` loops **once** with the whole string. Bit prod on 2026-07-31: `for pair in $WORKERS` over a `node:ip` list ran a single iteration and applied one node's Talos machineconfig to the **last** node's IP.

## 2. `scripts/browser-bridge/SKILL.md` — two tab-identity warnings

- **`open` can hand you a sibling's tab** — returns `reused: true` carrying another agent's `tabId` and URL instead of creating a fresh one. Observed 2026-07-31.
- **🔴 Verify the page before trusting a screenshot** — one `location.href` call after `open`/`wake`. An agent screenshotted a *different application instance* (a sibling's server on another port) and reported findings about the wrong build.

### Merge note

The uncommitted copy was layered on a **pre-#242** version of the file, so restoring it wholesale would have reverted the `ping` command row and the `extension_stale` health verdict. The additions were grafted onto current upstream instead. Verified after merge:

| | |
|---|---|
| `ping` row (from #242) | kept ✓ |
| `extension_stale` health verdict | kept ✓ |
| sibling-tab `reused` warning | present ✓ |
| verify-page section | present ✓ |
| only backup line dropped | the superseded `health` row — intentional |

Docs-only; no behaviour change.


ZacxDev:
Superseded by #279.

Two reasons this couldn't merge as-is:

1. It went stale behind 39 commits of browser-bridge work.
2. More importantly, it added ~911 bytes to `SKILL.md`, which sits at 11,911 of a 12,188-byte enforced budget — it would have **broken the size gate by ~730 bytes**. `test_skill_size.py`'s docstring records the gate being breached before "by 974 bytes when a feature commit added op documentation without evicting anything"; this was a fresh instance of exactly that.

#279 carries the same rescued content, routed to `reference/tabs-instances.md` per the gate's own remedy, leaving `SKILL.md` unchanged. Size gate re-run: 4 passed.

PR #257, opened 2026-08-01
fix(browser-bridge): session id from the POSIX session, not $PPID — it drifted across a command substitution

ZacxDev:
## The defect

`derive_session_id`'s fallback was a random token cached under **`$PPID`**. A `$( … )` that forks gives `browser` a different parent pid, so the cache key missed and a **brand-new id was minted** — the id is not stable across a command substitution.

That silently breaks per-session tab ownership in the idiom the docs teach:

```bash
T=$(browser open https://example.com | grep -oE '"tabId": *[0-9]+' | grep -oE '[0-9]+')
browser --tab "$T" emulate iphone-15
# browser: op 'emulate' may only run on a tab THIS session owns — … drop --tab
```

The `open` registers ownership under the subshell's id; the later call presents another. Measured on the workbench over ssh (2026-08-01):

```
$ echo "in subst:  $(browser --print-session-id)"
in subst:  ppid:2484606:c4b88b1d9de41681
$ browser --print-session-id
           ppid:2484584:3216410da01ef5e2      # DIFFERENT
```

Invisible on the laptop and under normal Claude Code use (`CLAUDE_CODE_SESSION_ID` is set, so the fallback never runs). It bites wherever that var is absent: **ssh, cron, a non-Claude shell, any script.**

## Precedence chain as it existed at 127090e (verified, not summarised)

1. `CLAUDE_CODE_SESSION_ID` → `claude:<uuid>`
2. `CLAUDE_SESSION_ID` → `claude:<uuid>`
3. `$TMUX_PANE` → `tmux:%N`
4. `$XDG_RUNTIME_DIR/browser-bridge/session-$PPID` → cached `ppid:<PPID>:<8 random bytes>`, minted on miss

`$TMUX_PANE` **is** in the chain (step 3) and it would **not** have saved the ssh case — it is unset over a plain ssh, which is why step 4 ran. All of 1–3 are env vars and are inherently stable across a subshell; only step 4 was not.

## The fix

New step 4: the **POSIX session id**, `sid:<sid>:<leader-starttime>`, parsed from `/proc/self/stat`. The PPID cache is demoted to step 5, a procfs-less last resort.

**Why this signal:**

- **Stable across a subshell** — `setsid(2)` is the only way to leave a session, so every subshell, pipeline stage, command substitution and exec'd child of one shell reports the same sid. (`$$` in a bash subshell also reports the parent's pid, but that is a bash-specific quirk that does not hold generally, and the drift was hit from zsh.)
- **Distinct across real sessions** — sshd, tmux (per pane), a terminal emulator, a systemd unit and cron all `setsid`, so two ssh logins / two tmux panes / two concurrent drivers get different sids and keep their tab isolation. Collapsing them would have been worse than the bug; that is asserted as a test in its own right.
- **`<leader-starttime>`** (stat field 22, ticks since boot) defuses **pid reuse** — a later session handed a recycled pid as its sid cannot inherit a dead session's tab.

`CLAUDE_CODE_SESSION_ID` stays highest-precedence (pinned by two tests).

**Restart-free, CLI-only.** `scripts/browser-bridge/browser` is an `mkOutOfStoreSymlink` onto the working tree, so this is live on `git pull` — no `home-manager switch`, no Brave reload, no manifest bump. `extension/` and `server.py` are untouched.

## Improved error message

Before (one message for both shapes, pointing at the wrong thing):

```
browser: op 'emulate' may only run on a tab THIS session owns — run 'browser open [url]' first and drop --tab …
```

After, with `--tab`:

```
browser: tab 4242 is not owned by THIS session.
browser:   this call's session id: sid:2484584:8290795
browser:   Either that tab belongs to another driver/session, or the
browser:   'browser open' that created it ran under a DIFFERENT id.
browser:   Compare them: run 'browser --print-session-id' the same way
browser:   you ran the 'open' (e.g. inside the same $( … ) or pipeline)
browser:   and again directly — if they differ, that is the bug.
browser: op 'emulate' refused (device emulation is deliberately never allowed on a tab this session does not own)
```

and without `--tab` (the session genuinely owns nothing):

```
browser: this session owns no tab, so 'emulate' has nothing it may act on.
browser:   this call's session id: claude:…
browser: run 'browser open [url]' first, then re-run 'emulate' …
```

⚠ **Honest scope:** the *other* session's id is **not** printed. The running server does not return the owner in the `not_owned_tab` body and exposes no owners endpoint, and `server.py` is deployed as a nix-store copy — adding a field there would need a `home-manager switch` + service restart, which this PR is explicitly forbidden to require. The message instead prints the id this call presented and tells you exactly how to obtain the other one. If you want both printed automatically, that is a follow-up PR that touches the server.

## Docs

- `reference/tabs-instances.md` — full precedence chain + a "**The subshell hazard**" section with the measured ssh transcript (this is where anyone scripting `open` will look).
- `reference/emulation.md` — the `not_owned_tab` vs `no_owned_tab` note no longer says "drop the `--tab`"; it says suspect the session id first.
- `README.md` — chain updated + why the POSIX session id and not `$PPID`.
- `SKILL.md` deliberately **not** touched (12,217 / 12,288-byte ceiling).

## Testing

New: `scripts/browser-bridge/tests/test_browser_session_id.py` (12 tests, headless, stub loopback bridge for the error-message cases).

**How the harness was validated** (this session has burned eight harnesses that tested nothing):

1. **In-suite control, runs every time.** `BROWSER_BRIDGE_PROC` is a test seam that points the procfs read at a nonexistent path, forcing the OLD last-resort branch. `test_harness_control_old_algorithm_drifts` runs the *exact same* stability probe against it and asserts it **DRIFTS**. If the probe ever stops being able to see the bug, that test fails loudly.
2. **The probe cannot self-deceive via `$$`.** The "direct" invocation is written to a **file**, never captured with `$( )`; the "subst" and "pipe" variants are the real caller shapes, and `$(a | b)` is guaranteed to fork.
3. **Whole-file control** — `git checkout 127090e -- browser`, byte-difference asserted non-zero (125 diff lines) and `grep -q _session_from_proc` asserted absent before running, then the new suite: **6 RED / 6 green**. The 6 green are labelled invariant guards (the precedence rules the fix must not take away), not regression coverage.
4. **Restoration verified** — every sweep ends with `cmp` against a pre-mutation backup (byte-identical) so a crashed sweep can't leave a mutant applied.
5. **Gate log read, not just the exit code** — `nix log` confirms `=== pytest scripts/browser-bridge/tests === 357 passed` *inside* the sandbox, so `/proc` and `setsid` work there and the new tests actually ran rather than skipping.

### Whole-file control at 127090e (6 RED / 6 invariant-guard green)

| test | verdict at 127090e |
|---|---|
| `stable_across_command_substitution` | 🔴 RED — the defect |
| `distinct_across_distinct_sessions` | 🔴 RED (on the `sid:` shape assertion — the old ppid ids *were* distinct, so this one is honestly a shape failure, not a distinctness regression) |
| `is_the_real_posix_session_id` | 🔴 RED |
| `posix_session_id_is_used_when_no_env_source_exists` | 🔴 RED |
| `not_owned_tab_with_explicit_tab_…` | 🔴 RED |
| `not_owned_tab_without_a_tab_flag_…` | 🔴 RED |
| `claude_code_session_id_wins_over_everything` | 🟢 INVARIANT GUARD |
| `claude_session_id_is_the_defensive_alternate` | 🟢 INVARIANT GUARD |
| `tmux_pane_beats_the_process_tree` | 🟢 INVARIANT GUARD |
| `ppid_cache_is_the_last_resort_only` | 🟢 INVARIANT GUARD |
| `claude_var_still_wins_with_procfs_available` | 🟢 INVARIANT GUARD |
| `harness_control_old_algorithm_drifts` | 🟢 (127090e *is* the old algorithm — passing is correct) |

### Mutation sweep (7 mutants, each verified applied before running; all caught)

| # | mutation | result | caught by |
|---|---|---|---|
| M0 | none (baseline) | 12 passed | — |
| M1 | `sid="$2"` (read ppid instead of the session field) | **2 failed** | stability, value-pin |
| M2 | `derive_session_id` returns the constant `sid:1:1` | **10 failed** | distinctness, value-pin, control, all 4 precedence, both error msgs |
| M3 | drop the `<leader-starttime>` suffix (pid-reuse guard) | **1 failed** | value-pin |
| M4 | sid branch moved above `$TMUX_PANE` | **1 failed** | tmux precedence |
| M5 | sid branch moved above `CLAUDE_CODE_SESSION_ID` | **1 failed** | claude precedence |
| M6 | `not_owned_tab` message reverted to the old single line | **2 failed** | both error-message tests |
| M7 | `BROWSER_BRIDGE_PROC` seam ignored (hard-coded `/proc`) | **2 failed** | control + last-resort |

M2 is the "collapse everything onto one id" mutant — the one that would have passed a stability-only test suite.

### Suites vs the measured 127090e baseline

| | baseline (127090e) | this branch |
|---|---|---|
| `node --test scripts/browser-bridge/tests/*.test.mjs` | 422 | **422 pass** |
| `pytest scripts/browser-bridge/tests` | 345 | **357 pass** (+12 new) |
| `nix build .#checks.x86_64-linux.pytests` | green | **PASS** (log read: browser-bridge 357 in-sandbox) |

## Needs a live check I could not do

Everything here is headless by construction. **Not verified live:** the end-to-end `open` → `--tab` → `emulate` path against real Brave, because that needs a browser and this PR is scoped away from live interaction. The ssh reproduction is the cheap confirmation — on the workbench, after `git pull`:

```
$ echo "in subst:  $(browser --print-session-id)"
$ browser --print-session-id
# expect: identical, and of the form sid:<pid>:<ticks>
$ T=$(browser open https://example.com | grep -oE '"tabId": *[0-9]+' | grep -oE '[0-9]+')
$ browser --tab "$T" emulate iphone-15      # expect: succeeds, no not_owned_tab
```

Also unverified across hosts: the laptop, where `CLAUDE_CODE_SESSION_ID` is set and this changes nothing (the two claude-precedence tests pin that, but on this host only).

### DIFF
--- a/scripts/browser-bridge/README.md
+++ b/scripts/browser-bridge/README.md
@@ -1237,11 +1237,24 @@ per-session (multi-step **workflow**) isolation did not.
   `/cmd`, derived (in order) from `CLAUDE_CODE_SESSION_ID` (Claude Code's own
   session UUID — identical across every Bash-tool call in a session, verified on
   the 2.1.x CLI) → `CLAUDE_SESSION_ID` (defensive alternate) → `$TMUX_PANE` (each
-  session runs in its own tmux pane) → a random token cached under the shell's
-  PPID (the Claude Code process, stable across a session's calls). It is used for
+  session runs in its own tmux pane) → the **POSIX session id**
+  (`sid:<sid>:<leader-starttime>` from `/proc/self/stat`) → (procfs-less systems
+  only) a random token cached under the shell's PPID. It is used for
   **routing only** and is **never** trusted for auth — bearer + Host still gate
   every request. If two sessions ever resolve the same id they share a tab
   (documented degradation — no worse than before).
+- **⚠ Why the POSIX session id, and not `$PPID` (fixed 2026-08-01).** The
+  PPID-keyed token was **not stable across a command substitution**: a `$( … )`
+  that forks gives `browser` a different parent pid, so
+  `T=$(browser open … | …)` registered tab ownership under one id and the
+  following `browser --tab "$T" emulate` presented another → `not_owned_tab`.
+  Measured over ssh on the workbench (`CLAUDE_CODE_SESSION_ID` unset there, so
+  the fallback actually ran). A subshell cannot leave its POSIX session — only
+  `setsid(2)` can — so the session id is stable across subshells while still
+  differing between two ssh logins, two tmux panes and two systemd units, which
+  is what keeps per-session tab isolation intact. The `<leader-starttime>`
+  suffix defuses pid reuse. See
+  `reference/tabs-instances.md` → "The subshell hazard".
 - **⚠ Sibling subagents share identity (known limitation).** Per-session
   isolation covers concurrent **top-level** sessions, NOT sibling subagents of
   one parent. Empirically (dumped from inside two parallel subagents) a subagent
--- a/scripts/browser-bridge/browser
+++ b/scripts/browser-bridge/browser
@@ -234,7 +234,12 @@ die() { echo "browser: $*" >&2; exit 1; }
 #   2. CLAUDE_SESSION_ID — defensive alternate spelling, in case a CLI build
 #      renames the var.
 #   3. $TMUX_PANE — the user runs each session in its own tmux pane, so the pane
-#      id is a good per-session proxy when no Claude var is exported.
+#      id is a good per-session proxy when no Claude var is exported. NOTE this
+#      step is COSMETIC, not structural: MEASURED 2026-08-01, each pane gets its
+#      own pty AND its own POSIX session (%0 sess=2982021 tty=pts/8, %1
+#      sess=2982024 tty=pts/9), so step 4 alone would already isolate panes.
+#      Kept because it is cheaper and more legible — do not "simplify" it away
+#      under the belief that removing it would break pane isolation.
 #   4. the POSIX SESSION id (`sid:<sid>:<leader-starttime>`), read from
 #      /proc/self/stat. See _session_from_proc below for why this — and NOT the
 #      PPID — is the right process-tree signal.
@@ -246,8 +251,13 @@ die() { echo "browser: $*" >&2; exit 1; }
 # and `browser --tab "$T" emulate` presented another → `not_owned_tab`. Measured
 # over ssh 2026-08-01; invisible wherever CLAUDE_CODE_SESSION_ID is set. Full
 # account: reference/tabs-instances.md → "The subshell hazard".
-# DEGRADATION: if two drivers ever resolve the SAME id they SHARE a tab — no
-# worse than the pre-fix behaviour. This is NOT hypothetical for one case:
+# DEGRADATION: if two drivers ever resolve the SAME id they SHARE a tab. This is
+# a TRADE, not a strict improvement (measured): two drivers with DIFFERENT
+# parents inside ONE POSIX session were distinct under the old ppid key and now
+# collide (sid:2983316:8395877 twice, vs ppid:2983326 / ppid:2983316) — reachable
+# only with all three env sources unset. In the commoner shape the new code is
+# better: three setsid'd runs from one shell were ONE id under the old key and
+# are three now. It is NOT hypothetical for one case:
 # **sibling subagents of one parent share identity.** Empirically (dumped from
 # inside two parallel subagents) a subagent inherits the PARENT's
 # CLAUDE_CODE_SESSION_ID *and* runs in the same $TMUX_PANE, and the harness
@@ -281,7 +291,12 @@ _session_from_proc() {
   [ $# -ge 4 ] || return 1
   sid="$4"
   case "$sid" in ''|0|*[!0-9]*) return 1 ;; esac
-  st="x"                                   # leader already gone → sid alone
+  # `x` = the session leader has already exited, so no starttime is available
+  # and the id degrades to `sid:<sid>:x` — the pid-reuse guard is OFF for that
+  # session. Recorded, not acted on: an orphaned session that then collides with
+  # a recycled sid is far-fetched, and server-side ownership self-releases on the
+  # 900s idle TTL anyway.
+  st="x"
   if [ -r "${procfs}/${sid}/stat" ] && IFS= read -r line < "${procfs}/${sid}/stat"; then
     rest="${line##*\) }"
     set -- $rest
--- a/scripts/browser-bridge/browser
+++ b/scripts/browser-bridge/browser
@@ -20,7 +20,7 @@
 #   Two Claude sessions driving one browser used to interleave on ONE shared
 #   active tab. Now each session can `open` its OWN tab; the server records the
 #   ownership under a stable per-session id (X-Session-Id header, derived from
-#   CLAUDE_CODE_SESSION_ID → $TMUX_PANE → a per-process-tree token — see
+#   CLAUDE_CODE_SESSION_ID → $TMUX_PANE → the POSIX session id — see
 #   derive_session_id below) and routes that session's html/eval/nav/screenshot/
 #   close to ITS tab. A session that never `open`ed falls back to the active tab
 #   (the one-shot read path — inherently safe). `--tab <id>` overrides the target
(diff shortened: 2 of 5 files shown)


ZacxDev:
## Round 2 — the five surviving `/proc`-parse mutants are closed, and the branch is rebased onto `e673e6b`

Pushed as `3f2ad81` (force-with-lease; the branch was rebased, not merged). Two commits: the fix (`0248c99`) and this coverage round (`3f2ad81`).

### Why they survived

The round-1 tests pin the **outcome** (stable across `$( )`, distinct across sessions), and every one of these five mutations happens to preserve that outcome *on this host's process tree*. The `$4`→`$3` one is the dangerous case: `test_session_id_is_the_real_posix_session_id` uses `start_new_session=True`, which makes **pid == pgid == sid** — so it was an incidental value-pin that could not discriminate the two fields at all. Reading pgrp instead of session silently reintroduces the exact bug this PR fixes, for interactive shells.

### New coverage (9 tests, +21 total in the file)

| # | test | pins |
|---|---|---|
| 1 | `comm_containing_a_close_paren_space_is_parsed_by_the_LAST_paren` | synthetic `/proc` with comm `my (weird) proc`, pgrp 555 ≠ session 888 — a first-`)` cut lands on the pgrp |
| 2a | `session_field_is_used_not_the_process_group_synthetic` | field 6 vs field 5, values set apart |
| 2b | `session_field_is_used_not_the_process_group_live` | **the real kernel**: `subprocess.Popen(process_group=0)` gives the child its own process group *inside our session* — precisely what an interactive shell does to a foreground command and what a forking `$( … )` does not. Asserts `pid != sid` **first**, so it cannot be vacuous, then asserts the output is `sid:<sid>:` and **not** `sid:<child pid>:` |
| 3a | `leader_starttime_is_stat_field_22_synthetic` | field 22 vs 21; the leader's field 21 is the recognisable filler `221`, asserted absent from the output |
| 3b | `leader_starttime_matches_awk_field_22_live` | cross-checks the real value against `awk '{print $22}'` — a **different parser**, so the expectation is not derived from the implementation. Skips (loudly) if the leader's comm contains a space, where `awk $22` would itself shift |
| 4 | `malformed_stat_degrades_to_the_ppid_fallback_silently` ×4 (`no-parens`, `truncated`, `session-zero`, `session-nan`) | both deleted guards. **Also asserts `stderr == ""`** — that is what kills the `[ $# -ge 4 ]` deletion, which still reaches the `ppid:` fallback but spews a `set -u` unbound-variable diagnostic on every invocation. Without the stderr assertion that mutant is an equivalent mutant and stays alive |

### Mutation table — all five now die

Each mutation is verified **applied** before the suite runs (presence grep for substitutions; a before/after occurrence **count** for the two deletions, since "the guard is gone" cannot be proven by a presence grep). Sweep ends with `cmp` against a pre-mutation backup — reported byte-identical, so no mutant is left applied.

| mutation | `browser` line | round 1 | round 2 | killed by |
|---|---|---|---|---|
| `${line##*\) }` → `${line#*\) }` | 291 | survives 12/12 | **1 failed** | comm-with-`") "` |
| `sid="$4"` → `sid="$3"` (pgrp) | 293 | survives 12/12 | **4 failed** | both pgrp tests incl. **live**, comm test, `session-nan` |
| `st="${20}"` → `${19}` (field 21) | 305 | survives 12/12 | **4 failed** | both starttime tests incl. the awk cross-check |
| delete `case "$sid" in ''\|0\|*[!0-9]*) return 1` | 294 | survives | **2 failed** | `session-zero`, `session-nan` |
| delete `[ $# -ge 4 ] \|\| return 1` | 292 | survives | **2 failed** | `no-parens`, `truncated` (via the stderr-silence assertion) |

Round-1's 7 mutants re-run against the expanded suite: still 7/7, now with wider blast radius (e.g. the constant-id mutant M2 → 19 failed / 2 passed).

Whole-file control `git checkout 127090e -- browser` (127 diff lines, `_session_from_proc` asserted absent before running): **11 RED / 10 invariant-guard green** — was 6/6.

### Comment accuracy (all three audit findings taken)

- **`$TMUX_PANE` ahead of `sid` is cosmetic, not structural.** Recorded verbatim with the measurement (`%0 sess=2982021 tty=pts/8`, `%1 sess=2982024 tty=pts/9`) and an explicit "do not *simplify* this away believing it is load-bearing".
- **The DEGRADATION note was wrong** and now states the trade honestly: two drivers with different parents in **one** POSIX session were distinct under the old `ppid:` key and now collide (`sid:2983316:8395877` twice vs `ppid:2983326`/`ppid:2983316`); the new code is better in the commoner shape (three `setsid` runs from one shell: one id before, three now).
- **`sid:<sid>:x`** (session leader already exited → no reuse guard) recorded at the `st="x"` assignment, explicitly not acted on, with the reason (900s ownership TTL).
- Backward compat noted in `reference/tabs-instances.md`: in-memory ownership + 900s idle TTL means an orphaned `ppid:`-keyed tab self-releases within 15 minutes. No migration.

### Rebase onto `e673e6b`

`git rebase origin/main`, no conflicts. I read their `browser` hunk: it touches `SUBCOMMANDS` (line 296), a new `context)` case (line 657), and the `text` flag loop (lines 991–1004). **Disjoint from all three of my regions** — the header docblock (20–24), `derive_session_id`/`_session_from_proc` (226–300) and the 409 `not_owned_tab` branch (~600–625). Nothing composes badly: their work adds a subcommand and a flag, mine changes session-id derivation and one error branch. Post-rebase I confirmed both features are intact in the merged file (`--annotated` loop present, `context)` case present, `bash -n` clean) and re-ran the full round-2 sweep on the rebased tree — same five kills.

`gh pr view 257 --json mergeable,mergeStateStatus` → **MERGEABLE / CLEAN**.

### Results against the corrected baselines

| | `origin/main` @ `e673e6b` | this branch |
|---|---|---|
| `node --test scripts/browser-bridge/tests/*.test.mjs` | 454 | **454 pass / 0 fail** |
| `pytest scripts/browser-bridge/tests` | **342 passed / 3 failed** | **363 passed / 3 failed** (+21, no fourth failure) |
| `nix build .#checks.x86_64-linux.pytests` | **RED** | **RED — identically** |

I measured the baseline myself rather than taking it on trust: `git archive e673e6b` into a scratch tree → 342/3 with exactly `test_manifest_version_reads_current`, `test_manifest_version_falls_back_to_repo_when_not_deployed`, `test_ping_op_set_mirrors_the_extension_protocol_js`. (That last one is the op-set contract test — I think it is the one you called `test_validate_command_contract`; same cause, the `context` op missing from a hardcoded literal.)

**The flake gate is RED on `main` itself**, and I proved that rather than inferring it: `nix build path:<scratch>/base#checks.x86_64-linux.pytests` on an untouched `e673e6b` tree → `RESULT: FAIL` with the **same three** failures and 342 browser-bridge tests. My branch's gate log shows the same three and **363 passed** — so the delta is +21 passing tests and zero new failures. Not chasing them, per your instruction; they clear when your manifest-literal PR lands, and this branch needs no change for that.

### Still unverified (unchanged from the original PR)

No live Brave interaction, no second host. The workbench ssh reproduction remains the end-to-end confirmation — after `git pull`, `$(browser --print-session-id)` and a direct call should be identical and shaped `sid:<pid>:<ticks>`, and `T=$(browser open … | …)` → `browser --tab "$T" emulate iphone-15` should succeed.

PR #266, opened 2026-08-02
docs(browser-bridge): wake is not a passive read; tab discard loses injected state

ZacxDev:
Three durable learnings from a 2026-07-31 → 08-01 session that used `wake` to observe a visibility-gated token-refresh loop on a real page host.

## `wake` fires `visibilitychange` — it is not a passive read

`wake` uses `Emulation.setFocusEmulationEnabled`, so it genuinely flips the document `hidden` → `visible` and the page **receives a real `visibilitychange` event**. Two consequences, in opposite directions:

- ✅ It is the **non-intrusive alternative to `activate`** for any page whose behaviour is gated on visibility — deferred timers, refresh/keep-alive loops, readiness handshakes that never complete while hidden. A woken background tab reaches the same state a foreground one would, with no screen steal.
- 🔴 **Every wake can run real page logic.** Measured: each observation wake fired the page's own handler, which issued a **fresh credential mint** — so the act of measuring reset the clock being measured, and the request log held more entries than the page's retry budget allows. A keep-alive wake past a deadline **pushes the deadline out**.

Practical rule added: decide before waking whether the thing you are measuring is visibility-sensitive, and plan the wake schedule as part of the experiment.

## A background tab can be DISCARDED, losing injected page state

Chromium may discard a backgrounded tab under memory pressure and silently reload it on activation. Tab id and URL are unchanged and nothing in the bridge output announces it — but a `window.__x` instrumentation hook or a patched `window.fetch` is gone. Documented the re-install-and-verify step, plus the tell: `typeof window.__x === "undefined"` means discarded, whereas an **empty array** means "installed, no events" — a completely different answer.

## `errors.md`: name the concrete shape of the version-vs-code gap

An op answering `unknown_op` while `health` reports a **matching** `extension_version` reads as a contradiction and sends people debugging the server. It is just the existing "version match ≠ code match" case. Observed on `wake` (2026-08-01). Believe the op, not the version; reload ↻ is unreliable, full Brave quit-and-reopen is the fix.

## Correction

The session also hit `browser js --wake '<expr>'` failing with *"takes exactly one JS argument"* — but that was **already fixed by #251** earlier the same day. Rather than document a stale gotcha, the `--wake` row now states that flag order is free, so it no longer implies a trailing-flag requirement.

Docs only — no code paths touched.

### DIFF
--- a/scripts/browser-bridge/reference/errors.md
+++ b/scripts/browser-bridge/reference/errors.md
@@ -117,6 +117,13 @@ readable) — `null` is NOT "fine". `browser whoami` carries the same three fiel
 change shipped without a manifest bump is invisible to it. That is why the
 version bump is part of the change contract, and why `ping` exists.
 
+🔴 **The concrete shape that gap takes: an op answers `unknown_op` while `health`
+reports a MATCHING `extension_version`.** That combination reads as a
+contradiction and sends people debugging the server or the CLI instead. It is not
+a contradiction — the *loaded* worker predates the op even though the manifest
+version agrees. Observed on `wake` (2026-08-01). **Believe the op, not the
+version**, and go straight to the reload/restart ladder below.
+
 **⚠ Reload ↻ is UNRELIABLE — a full Brave restart is the reliable fix.** The
 extension's long-poll keeps the OLD service worker permanently alive, so clicking
 ↻ in `brave://extensions` often does NOT swap in the new build. If a reload
--- a/scripts/browser-bridge/reference/spa-wake.md
+++ b/scripts/browser-bridge/reference/spa-wake.md
@@ -4,7 +4,9 @@
 a read reported `data.hidden:true` or printed a hidden-tab warning on stderr · an SPA
 looks stuck "Loading…" · `frames` lists no OOPIF you know is there · in-frame
 `click`/`type` hits elements that "don't exist yet" · you are about to conclude a site
-is BROKEN from a browser read · before driving any heavy JS app in an `open`ed tab.
+is BROKEN from a browser read · before driving any heavy JS app in an `open`ed tab ·
+you are measuring anything the page gates on VISIBILITY (timers, refresh loops,
+readiness) · an injected `window.__x` hook vanished between reads.
 
 Core: `~/workspace/devrc/scripts/browser-bridge/SKILL.md`.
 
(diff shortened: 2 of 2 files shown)


ZacxDev:
Rebased onto `origin/main` (was ~10 behind, CONFLICTING/DIRTY) and reworked in response to the adversarial audit. Same branch, force-pushed with `--force-with-lease` — review thread preserved.

## F1 — conflict + byte gate

`#267` (`bcbd8cc`) rewrote `SKILL.md` on main after this branch was cut, **merging the two `wake` rows into one**, so this branch's row edits were structurally obsolete on top of being over budget.

**Resolution: `SKILL.md` is now byte-identical to `origin/main` — 12,007 B.** Both core additions moved into `reference/spa-wake.md` (no cap).

| | bytes |
|---|---|
| merge-base | 13,262 |
| branch tip (before) | 13,636 |
| naive both-intents resolution | ~12,381 — **193 over** the 12,188 budget |
| **this branch now** | **12,007** (= main; 181 B of enforced headroom left) |

Gate: `python -m pytest scripts/browser-bridge/tests/test_skill_size.py -q` → **3 passed**. Negative control: appending 300 B took it to 12,307 and **2 of the 3 failed** as expected, so the green is a fact about the file and not about the harness.

The `--wake` flag-order sentence was dropped rather than relocated — it is already fully documented in the CLI's own help header (`scripts/browser-bridge/browser`, "FLAG ORDER / END OF FLAGS"), verbatim including both orderings.

## F2 — ran the live probe; the claim was wrong in a way neither the PR nor the audit had pinned

The PR asserted as measured that wake "fires a real `visibilitychange` event". Confirmed the audit's objection: the string `visibilitychange` appears **nowhere** in the bridge; the repo only ever measured the *value* of `document.visibilityState` (the rAF/timer table at `protocol.js:374-379`).

**A live browser was free, so I measured it instead of hedging.** Laptop (`192.168.50.155`), Brave profile `personal`, extension **0.7.0**, `extension_id bgbkamdlkdleahpgdgmjipjbgmepgenk`, background `example.com` tab, counting listener installed via `js`. No `activate`; `open` uses `active:false`, so nothing moved on the operator's screen. Probe tab closed afterwards.

| step | `window.__vc` | `visibilityState` |
|---|---|---|
| listener installed | 0 | hidden |
| **harness control** — synthetic `dispatchEvent` | 1 | hidden |
| reset, one `wake` | **2** | hidden |
| a second `wake` | **4** | hidden |
| reset, `js --wake` read *inside* the window | **1** | **visible** |
| same tab, read after teardown | **2** | hidden |

**One wake fires `visibilitychange` twice** — `visible` on apply, `hidden` on the `finally` teardown (`WAKE_CDP_TEARDOWN`, `protocol.js:479-483`). Deterministic (2 per wake, 4 after two). The synthetic-event control ran first specifically so a `0` could not have been a dead listener.

So the audit's factor-of-2 suspicion was right, and I dropped the hedge and documented the number. The PR's advice "count your own wakes first and subtract them" is replaced with: **subtract `2 × wake count`** for a handler bound to `visibilitychange` generally — with the caveat that the common `if (document.hidden) return;` shape runs only on the `visible` edge and so fires **once** per wake. The doc now says to read the page's handler rather than assume a multiplier.

## F3 — reversed the "nothing announces it" claim

Confirmed wrong as written. `assertTabCdpReady` (`protocol.js:1435-1437`) throws `TAB_DISCARDED_MESSAGE` **before** any `chrome.debugger.attach`, called from `service_worker.js:317`; `tabLoadSettled` fail-fasts on `discarded` (`protocol.js:332`). As written the PR would have taught a reader to disbelieve a real, actionable `tab_discarded` error.

Rewritten: a **currently** discarded tab is announced loudly. What is silent is the **reload afterwards** — `discarded` flips back to false, tab id and URL still match, nothing tracks document identity, and `OPS.tabs` (`service_worker.js:894-906`) carries no `discarded` field — so the injected state is gone with no tell. Tied to the existing RE-THROTTLE-after-reload section, since it is the same "a reload makes a new document" root cause.

## F4 — cross-referenced instead of restating

- `spa-wake.md` now points at `reference/agent.md`'s existing *"a read is passive, but a wake attaches the debugger to the operator's live tab"* — the operator-facing half of the same point, so the two cannot drift.
- `errors.md`: the new paragraph sat above the **measured 2026-07-30 nuance** and contradicted it ("reload is unreliable, full restart is the fix" vs "a ↻ reload DID take, so try it first"), while re-stating the version-vs-code gap already made 4 lines above. Folded the one genuinely new bit — the concrete shape, an op answering `unknown_op` while `health` reports a **matching** `extension_version`, observed on `wake` — into the paragraph that already owns that gap, and deleted the rest. **The measured nuance and the `browser ping` deterministic tell are preserved intact.**

## F5 — filed, not fixed

Opened **#273**. Verified the map (`documentEmulation`, `service_worker.js:252`) is cleared only by `forgetTab()` from exactly four sites (`:176`, `:273`, `:285-287`, `:1184`), none of which can fire on a discard-reload except `onReplaced`. The issue states both branches, the file:line evidence, and the exact `brave://discards` operator probe with the pass/fail read on the result envelope. No speculative code. `spa-wake.md` references it as an open question.

Nothing in the audit turned out to be wrong; F2 was understated rather than incorrect.

## Suites

- `node --test scripts/browser-bridge/tests/*.test.mjs` → **454 pass / 0 fail** (0 skipped, 0 todo)
- `python -m pytest scripts/browser-bridge/tests -q` → **369 passed / 0 failed** (366 + the 3 `test_skill_size.py` tests the rebase brought in)

Not merged, no `home-manager switch`, Brave untouched.

PR #303, opened 2026-08-03
guard: enforce talosctl reset / mkfs / dd-to-a-block-device for Claude Code (#295 follow-on)

ZacxDev:
Follow-on to #295. Three of the four remaining opencode-only families move into the `claude-code` policy that `bash-guard.py` runs. This is a **list change, not a reimplementation** — the checks already existed and were already correct.

## What was measured, before the change

Probed against the repo's `guard_core.py` under both policies, **with positive and benign controls in the same sweep** so a wired-to-nothing probe could not have produced the reassuring answer:

```
DENY   git add -A                          <- positive control (guard is live)
DENY   git reset --hard HEAD               <- positive control
ALLOW  talosctl -n 192.168.50.94 reset
ALLOW  talosctl --nodes 192.168.50.94 reset
ALLOW  sudo talosctl reset
ALLOW  mkfs.ext4 /dev/sdc
ALLOW  mke2fs /dev/sdc
ALLOW  dd if=/dev/zero of=/dev/sdc
ALLOW  dd if=/dev/zero of=/dev/nvme0n1 bs=4M
ALLOW  ls -la /tmp                         <- benign control
```

Zach independently measured the three headline rows ALLOW against the **deployed** `~/.claude/hooks/bash-guard.py` the same evening, with the same verdicts. So: wiping a Talos node, formatting a disk and overwriting a block device were all permitted, unprompted, on the operator's primary tool.

## What changed

- `check_talosctl_reset`, `check_mkfs`, `check_dd_to_block_device` move from `_IRREVERSIBLE_CHECKS` into `_CLAUDE_CODE_CHECKS`. `POLICIES["opencode"]` is `_CLAUDE_CODE_CHECKS + _IRREVERSIBLE_CHECKS`, so **opencode keeps coverage by inheritance, not duplication** — asserted, not assumed.
- All three deny messages now carry the #295 "if you are only QUOTING this" escape hatch (write the text to a file, `git commit -F` / `gh pr create --body-file`). The guard matches raw text and the argv parser splits on newlines, so a heredoc body line documenting the ban parses as the command. Factored into one `_QUOTING_ESCAPE_HATCH` constant so a message asserted to contain it cannot drift per-check.
- **Ordering**: they sit before `check_heredoc_to_file` / `check_cd_then_git`, so a command tripping both reports the device-destruction reason rather than the token-waste one. Same rationale as #295's placement of the git argv checks.
- Docs updated at all four places that describe the policy split: `guard_core.py`'s module docstring, the ARGV-CHECKS section header, `bash-guard.py`'s header, and `nix/home.nix`.

### 🔴 `check_rm_rf_critical` was deliberately NOT moved

It is now the **only** opencode-only check, and it stays that way. `rm -rf` has legitimate, frequent use on these hosts — build directories, `node_modules`, `.direnv`, throwaway worktrees, `/tmp` scratch trees. The check is narrow (only `/`, `~`/`$HOME`, `.`/`..`, top-level system dirs), but narrow is not never, and **a guard that fires during routine cleanup trains its subject to route around it — worse than no guard, because it also reports safety.** Claude Code additionally falls back to a prompt the operator sees, which is exactly the control opencode lacks (`opencode run` auto-rejects an `ask`).

The reasoning is written into `_IRREVERSIBLE_CHECKS`, `bash-guard.py`'s header and `IRREVERSIBLE_EXPECTED` — three places a future reader might look before "finishing the job" — and pinned by `test_rm_rf_is_the_only_check_left_out_of_claude_code`, which **fails in both directions**: if `check_rm_rf_critical` is added to claude-code, and if a *new* check is parked in `_IRREVERSIBLE_CHECKS` as a way of adding a guard without the claude-code conversation.

## Allow-fence (preserved, asserted)

`talosctl version`, `talosctl -n <ip> get members`, `talosctl -n <ip> get resetstatus`, `talosctl … dmesg/health/services`; file-to-file `dd`, `dd of=/dev/null` and the other pseudo sinks, and `dd if=/dev/sdz of=/tmp/backup.img` (**reading** a device is not writing one); `mkfs` appearing only inside a quoted string — the check matches on the *program name*, so `echo 'mkfs.ext4 is banned'` and `grep -rn 'mkfs' RULES.md` are arguments, not commands; near-miss program names `mkdir`, `mktemp`, `ddrescue`. 18 rows, all invariant guards — green before and after, so they are the false-positive fence, **not** regression coverage.

## Red/green matrix — base ref `26d5268` (`origin/main`)

Method: `git archive 26d5268 scripts/claude-hooks` into a temp tree, drop in the NEW test file, run. Collection is identical in both runs (1129), so the two are comparable — a "red" here cannot just mean "the file did not import".

| run | result |
|---|---|
| **NEW tests vs BASE `guard_core.py`** | **48 failed**, 1081 passed |
| NEW tests vs HEAD `guard_core.py` | **1129 passed**, 0 failed |
| BASE tests vs BASE (control) | 3 failed, 1026 passed |

The 3 control failures (`test_plugin_*`) are environmental — the `git archive` tree lacks the opencode plugin the test reads — and appear identically in the base run. **Net genuinely-red: 45.**

| test | rows red at base |
|---|---|
| `test_mkfs_is_denied_under_claude_code` | 14 |
| `test_talosctl_reset_is_denied_under_claude_code` | 13 |
| `test_dd_to_block_device_is_denied_under_claude_code` | 13 |
| `test_claude_code_policy_is_pinned_by_name` | 1 |
| `test_opencode_policy_is_a_strict_superset` | 1 |
| `test_rm_rf_is_the_only_check_left_out_of_claude_code` | 1 |
| `test_mkfs_in_a_heredoc_body_reports_the_mkfs_reason_not_the_heredoc_reason` | 1 |
| `test_device_deny_messages_carry_the_quoting_escape_hatch` | 1 |

**Labelled as invariant guards, NOT regression coverage** (green before and after): the 18-row `DEVICE_ALLOW` fence, and `test_the_new_checks_do_not_leak_into_claude_code` / `test_cli_honours_the_policy_flag`, which now use `rm -rf` fixtures that behave identically at base.

`test_cli_honours_the_policy_flag` was *changed*, not added: it used `talosctl reset` as its "claude-code allows this" fixture, which this PR correctly turns red. `rm -rf /` is the only remaining disagreement between the two policies, so it is now the only fixture that can exercise the `--policy` flag at all.

## Mutation matrix — 10 mutants, all KILLED **and attributed**

Every mutant must produce failures matching **that guard's own** test names. A mutant that only trips a *different* guard's test is reported UNATTRIBUTED (the RULES failure mode "a different guard's error kills your test"), not KILLED.

| mutant | verdict |
|---|---|
| talosctl: drop from claude-code policy | KILLED — 129 failed, attributed |
| mkfs: drop from claude-code policy | KILLED — 58 failed, attributed |
| dd: drop from claude-code policy | KILLED — 51 failed, attributed |
| talosctl: neuter the verb match | KILLED — 215 failed, attributed |
| mkfs: neuter the program regex | KILLED — 110 failed, attributed |
| dd: treat every `/dev/` target as a safe sink | KILLED — 85 failed, attributed |
| dd: **drop** the safe-sink exemption (over-broad) | KILLED — 4 failed, hits the ALLOW fence |
| mkfs: strip the quoting escape hatch | KILLED — 1 failed, attributed |
| ordering: move the three AFTER `check_heredoc_to_file` | KILLED — 3 failed, attributed |
| `rm -rf`: "finish the job" — add it to claude-code | KILLED — 7 failed, attributed |

**Harness controls, both run and both required before any verdict above was believed:**
- **POSITIVE** — unmutated baseline before and after the sweep: `collected=1129 failed=0`. A zero/one collected count is what a broken harness looks like; the sweep aborts on it. The source is restored after every mutant so a left-applied mutation cannot poison the next baseline.
- **NEGATIVE (two, run explicitly)** — a mutant whose needle is absent reported `SKIP-BROKEN` (a never-applied mutant must not read as green), and a no-op mutation reported `🔴 SURVIVED`. The harness can go red in both shapes, so its greens mean something.

The sweep **counts** `N failed, N passed` from the summary line and fails loudly if that format does not match; it never reads an exit code.

`test_mkfs_in_a_heredoc_body_reports_the_mkfs_reason_not_the_heredoc_reason` is the one row that is **not** green-by-default: with this change reverted the command still DENIES, via `check_heredoc_to_file`. A bare `is not None` would be green for the wrong reason and would pass with `check_mkfs` removed from the policy entirely. It also asserts its own fixture still trips the heredoc check, so it cannot degrade into passing trivially.

## Both tiers

| tier | result |
|---|---|
| **`nix build .#checks.x86_64-linux.pytests`** (authoritative) | **collected=4760 passed=4759 skipped=1 failed=0** |
| dev host, `scripts/run-tests.sh` | see comment below |

Main is `4660 / 4659 / 1`. Delta **+100**, exactly the tests added here (`test_guard_core.py`: 1029 → 1129). The single skip is the pre-existing opt-in `repo-cos` live-store drift check.

## Files touched

`scripts/claude-hooks/guard_core.py`, `scripts/claude-hooks/bash-guard.py`, `scripts/claude-hooks/tests/test_guard_core.py`, and **`nix/home.nix`** — the `nix/home.nix` hunk is a **6-line comment change only**, in the `home.file.".claude/hooks/guard_core.py"` block, describing the new policy split. No option or path changed. Flagged because sibling agents are also touching that file.

## Deployment

`bash-guard.py` and `guard_core.py` are `home.file` targets — **merging changes nothing**. This takes effect on `home-manager switch` / `ship.sh`. Not run here, per the task constraints.

### DIFF
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -664,11 +664,13 @@ in
   # The SAME source file is also deployed to ~/.config/opencode/guard_core.py
   # below and driven by the opencode plugin — one implementation, two harnesses,
   # two named policies. "claude-code" was the frozen original six until
-  # 2026-08-02, when three argv checks were added to it by explicit operator
-  # decision — `git -C <p> reset --hard`, `git stash` and `git clean -f`, all
-  # 🔴 in RULES.md and all measured ALLOW against the live hook beforehand.
-  # "opencode" = those plus the still-opencode-only irreversible families
-  # (talosctl reset / mkfs / dd to a block device / rm -rf of a critical path).
+  # 2026-08-02, when SIX argv checks were added to it by explicit operator
+  # decision — `git -C <p> reset --hard`, `git stash`, `git clean -f`,
+  # `talosctl … reset`, `mkfs*` and `dd of=/dev/<block-dev>`, all 🔴 in RULES.md
+  # or unconditionally destructive, and all measured ALLOW against the live hook
+  # beforehand. "opencode" = those plus the one still-opencode-only family,
+  # `rm -rf` of a critical path (held back deliberately — it has frequent
+  # legitimate use here, so a deny would train the operator to route around it).
   # POLICIES in guard_core.py is the authority; test_guard_core.py pins it.
   home.file.".claude/hooks/guard_core.py" = {
     source = ../scripts/claude-hooks/guard_core.py;
--- a/scripts/claude-hooks/bash-guard.py
+++ b/scripts/claude-hooks/bash-guard.py
@@ -17,6 +17,12 @@ operator decision. The list today:
     incident 2026-07-25). `git stash list`/`show` are reads and stay allowed.
   - git clean -f             -> deletes untracked files, which here are routinely
     real work (2026-08-02; RULES 🔴). `git clean -nd` stays allowed.
+  - talosctl … reset         -> WIPES a Talos node (2026-08-02). Reads such as
+    `talosctl version` / `talosctl -n <ip> get members` stay allowed.
+  - mkfs / mke2fs / mkswap … -> formats a filesystem (2026-08-02). Matched on the
+    PROGRAM name, so naming it in an argument is not a command.
+  - dd of=/dev/<block-dev>   -> overwrites a disk in place (2026-08-02).
+    `of=/dev/null` and the other pseudo sinks, and file-to-file dd, stay allowed.
   - large heredoc -> file    -> use the Write tool (token waste; audit-driven)
   - cd <path> && git ...     -> use git -C <path> (audit: #1 command shape, 1482x)
   - private key in a command -> reference the key file instead (never inline)
@@ -24,13 +30,18 @@ operator decision. The list today:
     committing/posting (insights: leaked ingress IP into a public repo once)
 
 This hook fires on EVERY Bash call in EVERY Claude Code session on both hosts,
-so adding a check here changes the operator's primary tool. FOUR
-irreversible-action families remain opencode-ONLY and are deliberately NOT
-enabled here: talosctl reset, mkfs, dd to a block device, and rm -r of
-/|$HOME|cwd|a top-level system dir. `rm -rf` in particular risks false positives
-on legitimate build-directory cleanup. Switching this adapter to the "opencode"
-policy is a one-word change — and a decision for the operator to make
-explicitly, not a side effect of hardening a different tool.
+so adding a check here changes the operator's primary tool. Exactly ONE
+irreversible-action family remains opencode-ONLY and is deliberately NOT enabled
+here: `rm -r` of /|$HOME|cwd|a top-level system dir. 🔴 That is a DECISION, not
+the leftover of an unfinished migration — `rm -rf` has frequent legitimate use
+on these hosts (build dirs, node_modules, throwaway worktrees), and a guard that
+fires during routine cleanup trains the operator to route around it, which is
+worse than no guard because it also reports safety. Claude Code additionally
+falls back to a PROMPT the operator sees, the control opencode lacks. Do not
+"finish the job" by adding it; see _IRREVERSIBLE_CHECKS in guard_core.py.
+Switching this adapter to the "opencode" policy is a one-word change — and a
+decision for the operator to make explicitly, not a side effect of hardening a
+different tool.
 
 I/O contract (unchanged): reads PreToolUse JSON on stdin (`tool_name`,
 `tool_input.command`), prints `hookSpecificOutput.permissionDecision = "deny"`
--- a/scripts/claude-hooks/guard_core.py
+++ b/scripts/claude-hooks/guard_core.py
@@ -37,13 +37,18 @@ TWO CALLERS, TWO POLICIES
                   RULES.md mandates), followed the same day by
                   `check_git_stash` + `check_git_clean_force` — both 🔴 in
                   RULES.md, both measured ALLOW against the live hook, and
-                  neither with a benign in-repo use. Pinned by name in
+                  neither with a benign in-repo use — and then, later the same
+                  day, by `check_talosctl_reset` + `check_mkfs` +
+                  `check_dd_to_block_device`, the three device/cluster-
+                  destruction families, on the same evidence. Pinned by name in
                   test_guard_core.py.
-  "opencode"    — all of the above, plus the irreversible-action checks below. opencode
-                  agents run unattended (`opencode run` AUTO-REJECTS an `ask`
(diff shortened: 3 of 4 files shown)


ZacxDev:
**Dev-host tier** (the row deferred in the body):

```
TOTAL collected=4760  passed=4759  skipped=1  failed=0  (floor: 2850)
RESULT: PASS
```

Identical to the authoritative `nix build .#checks.x86_64-linux.pytests` run — same collected count, same single pinned skip, zero failures. Both tiers agree.

⚠️ Worth recording, because it is the exact trap RULES warns about: the first dev-host attempt ran `scripts/run-tests.sh` **bare in the worktree**, which has no `.envrc`. Every suite reported `unparseable summary`, `TOTAL collected=0`, `RESULT: FAIL` — **and the process still exited 0.** Reading the exit code there would have reported a pass on a run where nothing executed. The floor check (`only 0 tests were collected, floor is 2850`) is what caught it. Re-run under `nix-shell` with the flake check's own dependency set (`python312.withPackages [pytest requests psycopg2 minio pyyaml]`, nodejs, nix, ripgrep, git, util-linux, jq, gnugrep, curl) to get the numbers above.

ZacxDev:
## Rebased onto current `origin/main` (`4eb5798`) — re-gated

This PR was originally based on `26d5268`; `origin/main` has since taken #301 and #302. Rebased (clean, no conflicts) and force-pushed, then re-run:

| tree | sandbox (`nix build .#checks.x86_64-linux.pytests`) |
|---|---|
| `main` @ `4eb5798` (derived) | 4662 |
| **this PR, rebased** | **4762 / 4761 / 1 / 0** |
| this PR **+ #305** merged onto `4eb5798` | **4788 / 4787 / 1 / 0** |

Delta is **+100**, exactly the tests added here (`test_guard_core.py` 1029 → 1129) — unchanged by the rebase. The merged-tree total closes exactly as 4662 + 100 + 26, so there is **no cross-PR interaction**: neither PR displaced the other's tests. Dev-host tier on the merged tree agrees number for number (4788 / 4787 / 1 / 0).

The merged tree was built as a real integration branch off `origin/main` with both branches merged in, and its `nix/home.nix` was checked by hand: this PR's hunk is a 6-line comment near line 666, #305's is a new block appended at the end, and `nix-instantiate --parse` is clean on the merged result.

⚠️ Correction to my earlier comment's baseline: it compared against **4660**, which was `main` at `26d5268`. The current baseline is **4662** — #302 nets +2 in `scripts/collector/opencode/tests` (170 → 172). The +100 attributable to this PR is unaffected.

PR #305, opened 2026-08-03
hygiene: cap the ~/.claude logs, and fix the session-summary amplification

ZacxDev:
Two unrelated-but-small hygiene items. A third was scoped out after inspection — **see "What I left out" at the bottom**; it is not an oversight and it has a security finding attached.

---

### Measured, on the workbench 2026-08-02

| file | bytes | last written |
|---|---:|---|
| `notify.log` | 39,822,110 | 2026-07-06 — a **dead** writer |
| `clawgate-hook.log` | 12,033,179 | live |
| `claude-notify.log` | 60,684 | live |
| `daemon.log` | 6,091 | 2026-07-31 |

**51.6 MB unrotated and growing**, nothing rotating any of it.

### Why the cap is on the DIRECTORY, not in the writers

Only **one** of those writers lives in this repo (`scripts/claude-hooks/claude-notify.py` → `claude-notify.log`, the 60 KB one). `clawgate-hook.log` comes from the clawgate approval hook and `notify.log` from something that no longer runs at all — a per-writer fix cannot reach either.

New `claude-log-rotate` `systemd --user` service + daily timer (following the `mail-actions-archive` pattern already in `home.nix`) running `scripts/claude-log-rotate/rotate.sh`: `size 10M`, `rotate 3`, `compress`, `delaycompress`, `missingok`, `notifempty`, and — load-bearing — **`copytruncate`**.

`copytruncate` is not an optimisation. The writers hold open file descriptors, they are not ours to signal, and one of them fires on **every Claude Code tool call**. Renaming the file out from under an open fd would leave it appending to an unlinked inode: the log would look frozen while silently consuming the same disk.

**Not `serverMode`-gated** — both hosts run Claude Code and both accumulate these logs. Gating it would have left the laptop uncapped, which is the "shipped, but not where the problem is" shape; there is a test for exactly that.

### 🔴 The `.bak` files are NOT touched

Scope is `*.log` only. These nine stale hand-made config backups are left entirely alone — **for you to remove by hand**, since deciding to delete a config backup is not an automated job's call:

```
settings.json.bak.1769397206              161 B   2026-01-25
settings.json.bak.clawgate.1780707021     751 B   2026-06-05
settings.json.bak.20260623-190634         840 B   2026-06-09
settings.json.bak.20260724-192550       1,264 B   2026-07-06
RULES.md.bak.20260608-081434           16,132 B   2025-10-18
RULES.md.bak-20260616-204947           19,385 B   2026-06-16
RULES.md.bak.20260623-153503            7,103 B   2026-06-16
PRINCIPLES.md.bak-20260616-204947       2,573 B   2026-06-16
CLAUDE.md.bak-2026-07-30                1,406 B   2026-07-30
```

That fence is **measured, not read off the config**: the test creates over-threshold `.bak` / `settings.json` / `notes.md` files next to a `.log` **canary**, and requires the canary to have rotated before it believes "nothing else was touched". Otherwise a no-op run would pass it.

`logrotate` is added to **both** gate tiers (`flake.nix`'s pytests check and `run-tests.sh`'s `REQUIRED_TOOLS`), so the behavioural tests **fail** rather than skip when it is absent — a skipped rotation test reports safety it never measured.

---

### The bug

`emit_decision`'s `settled` branch returned `(True, "settled")` **without ever consulting `emitted_at`**, while the `interim` branch immediately below it was rate-limited. The rate limiting was therefore *asymmetric in exactly the wrong direction*:

- an **ACTIVE** session (never goes idle) → bounded to one emit per interim window;
- a session idling **BETWEEN BURSTS** → bounded by *nothing*. Every burst followed by ≥ the settle window re-entered the settled branch and shipped a fresh full rollup.

The second is the ordinary `claude --resume` shape. So the bound was missing from precisely the common case, which is why the tail is so heavy while the median looks fine.

Measured over the 30 days to 2026-08-02: **31,815 rows / 482 sessions — median 2, p90 209, max 572, mean 66**, against `validation/invariants.py`'s `session_summary_rows_bounded` (>24 rows/session/24h). A documented invariant, violated.

### The fix

The **missing consult**, not a clamp on a counter: the settled branch now reads the same `emitted_at` with the same interval, so there is **one** knob (`CLAUDE_SUMMARY_INTERIM_HOURS`) rather than two that can disagree. New reason `settled-recently`, which deliberately does **not** record the new signature — so the session still owes a rollup and emits on the first tick past the window.

### 🔴 The legitimate cases, preserved and pinned

- **Resumed days later.** A session that settles, is resumed days later and settles again has an `emitted_at` far older than the interval, so it emits. `test_decision_settled_reemits_after_a_long_gap` (pure) and `test_resumed_after_settle_reemits_the_final_rollup` (end-to-end). Mutant **c3** makes the limit permanent and both go red.
- **A session's FIRST rollup is never deferred** — no known `emitted_at` (never emitted, or a v1/corrupt state entry) always emits, which is what keeps `session_summary_no_orphans` (2 h grace) satisfiable. Mutant **c2** removes that and it goes red.
- **`CLAUDE_SUMMARY_SETTLE_MINUTES=0`** still bypasses the gate entirely, as documented — rate-limiting it would make the knob a no-op. Mutant **c4**.

**Cost, stated honestly:** a deferred settle means the newest row can lag the transcript's final content by up to the interim window (4 h default). Nothing is lost — the deferral does not record the signature, the rollup lands on the first tick past the window, and reads take `argMax(…, ingested_at)` anyway.

**No backfill, no DELETE against ClickHouse.** Existing rows dedupe on read and age out of the invariant's trailing window.

### End-to-end, with a positive control

36 bursts over 24 h, each settling before the next. Same scenario, same code path, twice — the only thing that changes is the knob, because a small number is also what a harness wired to nothing produces:

| run | rows for one session in 24 h |
|---|---:|
| **POSITIVE CONTROL** (`interim=0` → unbounded settled = pre-fix behaviour) | **37** |
| **UNDER TEST** (4 h default) | **7** |
| cap (`SUMMARY_ROWS_PER_SESSION_CAP`, read from `invariants.py`, not restated) | 24 |

7 = `1 first-seen + 24h/4h`, as predicted. The test asserts the control **exceeds** the cap before it believes the bounded number.

---

## Red/green matrix — base ref `26d5268`

Method: `git archive 26d5268 scripts nix flake.nix` into a temp tree, drop in this PR's test files (plus `rotate.sh` itself — a test for a file that does not exist is an import error, not a red test), run. **Collection is 64 in both runs**, so the two are comparable.

| run | result |
|---|---|
| NEW tests vs BASE sources | **11 failed**, 53 passed |
| NEW tests vs HEAD sources | **64 passed**, 0 failed |
| BASE tests vs BASE sources (control) | 56 passed, 0 failed |

Red at base:

| test | item |
|---|---|
| `test_decision_settled_is_rate_limited_by_the_last_emit` | (c) — **the** regression pin |
| `test_decision_settled_rate_limit_follows_the_interim_knob` | (c) |
| `test_bursty_resumed_session_stays_under_the_invariant_cap` | (c) end-to-end |
| `test_settled_growth_reemits_the_complete_rollup` | (c) — *changed*, see below |
| `test_resumed_after_settle_reemits_the_final_rollup` | (c) — *changed*, see below |
| `test_home_nix_declares_the_service_and_the_timer` | (b) wiring |
| `test_the_unit_runs_the_script_this_file_tests` | (b) wiring |
| `test_the_unit_puts_logrotate_on_its_path` | (b) wiring |
| `test_the_timer_is_not_server_mode_gated` | (b) wiring |
| `test_the_timer_is_persistent` | (b) wiring |
| `test_logrotate_is_pinned_in_both_gate_tiers` | (b) two-tier pin |

**Labelled as invariant guards / new-feature coverage, NOT regression coverage:** `test_decision_settled_with_no_known_last_emit_always_emits` and `test_decision_settle_zero_bypasses_the_rate_limit` (green before *and* after — they are the fence around what the rate limit must not swallow), and all of the `rotate.sh` **behavioural** tests, which are green at base because the script is new. They earn their keep in the mutation matrix, not here.

**Two existing tests were CHANGED, not just added to**, and both were changed because this PR correctly turns them red — they encoded the unbounded settle. Neither had its goalposts quietly moved: each now asserts **both** sides, the deferral *and* that the deferral costs only latency (the rollup still lands, still complete, past the window). Reverting the fix turns the new first half red.

---

## Mutation matrix — 16 mutants, all KILLED **and attributed**

A mutant must produce failures matching **that thing's own** test names; one that merely trips some other test is reported UNATTRIBUTED, not KILLED.

| # | mutant | verdict |
|---|---|---|
| c1 | **revert the fix** — settled emits without consulting `emitted_at` | KILLED (5, attributed) |
| c2 | rate-limit the FIRST rollup too | KILLED (3) |
| c3 | make the limit PERMANENT — a resume days later never re-emits | KILLED (6) |
| c4 | rate-limit the `settle=0` escape hatch | KILLED (2) |
| c5 | decouple the bound from the interim knob (hard-code 1 s) | KILLED (5) |
| b1 | widen the glob to every file (would eat the `.bak` copies) | KILLED (4) |
| b2 | drop `copytruncate` | KILLED (3) |
| b3 | remove the size cap | KILLED (5) |
| b4 | unbounded generations | KILLED (3) |
| b5 | report success when `logrotate` is missing | KILLED (1) |
| b6 | unquote the glob (shell-expands at config time) | KILLED (3) |
| b7 | drop `logrotate` from the flake check | KILLED (1) |
| b8 | drop `logrotate` from `REQUIRED_TOOLS` | KILLED (1) |
| b9 | `serverMode`-gate the timer (laptop left uncapped) | KILLED (1) |
| b10 | point the unit at a different script | KILLED (1) |
| b11 | drop `pkgs.logrotate` from the unit PATH | KILLED (1) |

**b10 SURVIVED on the first sweep** and that is the most useful thing here. `test_the_unit_runs_the_script_this_file_tests` originally grepped the whole of `home.nix` for the script path — which also appears in the surrounding comment and in `X-Restart-Triggers` — so it stayed green with `ExecStart` repointed at `ship.sh`. The test now asserts on the `ExecStart` **line**. The sweep found a real hole in my own test; without it that test would have shipped proving nothing.

**Harness controls, both required before any verdict above was believed:**
- **POSITIVE** — unmutated baseline before *and* after the sweep: `collected=64 failed=0`. The sweep aborts on an implausible count, and restores the source after every mutant so a left-applied mutation cannot poison the next baseline.
- **NEGATIVE (two, run explicitly via `--self-check`)** — an absent needle reported `SKIP-BROKEN` (a never-applied mutant must not read as green) and a no-op replacement reported `🔴 SURVIVED`. The harness can go red in both shapes.

The sweep **counts** `N failed, N passed` and fails loudly if that format does not match; it never reads an exit code.

---

## Both tiers

| tier | result |
|---|---|
| **`nix build .#checks.x86_64-linux.pytests`** (authoritative) | **collected=4688 passed=4687 skipped=1 failed=0** |
| dev host, `scripts/run-tests.sh` | see comment below |

**⚠️ Baseline note — read this before comparing to 4660.** `origin/main` advanced from `26d5268` to `4eb5798` (#301, #302) while this work was in progress, and **this branch is based on `4eb5798`** (PR #303 is based on `26d5268`; it is being rebased). So the arithmetic is against `4eb5798`, not against the 4660 figure:

- `scripts/tests` 998 → 1018 (**+20**, this PR)
- `scripts/collector/claude/tests` 56 → 62 (**+6**, this PR)
- `scripts/collector/opencode/tests` 170 → 172 (**+2, NOT this PR** — #302; verified by diffing collected node IDs at both refs: base has 3 tests HEAD lacks, HEAD has 5 base lacks, all of them `test_deploy_*` / plugin-export tests from #302)

**+26 from this PR**, exactly the tests added. The single skip is the pre-existing opt-in `repo-cos` live-store drift check.

---

## `git merge-tree` against PR #303

```
git merge-tree --write-tree guard-widen-three-families hygiene-logrotate-and-session-tailer
-> exit code 0   (tree 686e1435e349b2ed511ad6a678e51643656dd435)
```

Exit 0 = **no textual conflict** — branched on the exit code, not on a marker grep, since two-arg `merge-tree` prints only a tree OID on success and never emits `<<<<<<<`.

Both PRs touch `nix/home.nix`, so a clean git merge is not by itself a clean merge. I read the merged region: PR #303's hunk is a **6-line comment** in the `home.file.".claude/hooks/guard_core.py"` block (~line 666); this PR's is a **new block appended at the end of the file**. Verified on the actual merged tree — it carries both changes and `nix-instantiate --parse` is clean. Full merged-tree gate run is reported in a comment below.

---

## Files touched

`scripts/claude-log-rotate/rotate.sh` (new), `scripts/tests/test_claude_log_rotate.py` (new), `scripts/collector/claude/session-tailer.py`, `scripts/collector/claude/tests/test_session_tailer.py`, `flake.nix`, `scripts/run-tests.sh`, `nix/home.nix`.

Two of these are shared with sibling agents, so, explicitly:
- **`nix/home.nix`** — one new `systemd.user.{services,timers}.claude-log-rotate` block **appended at the end**. Nothing existing modified.
- **`scripts/run-tests.sh`** — **one line**, `REQUIRED_TOOLS=…` gains `logrotate`. The target list is untouched, so it should not collide with the agent adding `scripts/dl-router/tests`.

## Deployment

The timer only exists after a `home-manager switch` / `ship.sh` — merging changes nothing. `session-tailer.py` likewise: the `claude-activity-source` unit picks up the new emit policy on switch (X-Restart-Triggers). Neither was run here, per the task constraints.

**First run reclaims ~50 MB** (the 39.8 MB `notify.log` becomes an empty live file plus one compressed generation).

---

## 🔴 What I left out: (a) bringing the `clickup` skill under nix

Correct on the facts — `~/.claude/skills/clickup/` is a hand-placed regular-file tree, `readlink` returns nothing, and every other skill is nix-managed from `claude/skills/`. But it is **not a skill directory**; it is a **5.9 MB vendored third-party Node application** with its own `.git`, `node_modules/`, `package-lock.json` *and* `pnpm-lock.yaml`, `flake.nix`, `api/`, `lib/`, `test/`, an 86 KB `query.mjs` and a 15 KB `listen.mjs`. `SKILL.md` is one file inside it.

Three separate blockers, each of which alone rules out a hygiene-sized change:

1. **🔴 Credentials.** It contains `accounts.json` (mode 0600) holding a live ClickUp `pk_` API token, and `SKILL.md` instructs the operator to *"Create `accounts.json` in this skill directory"*. Under a `home.file` deploy that directory becomes a read-only `/nix/store` symlink — **the documented credential flow becomes impossible**. And devrc is a **public** repo; the upstream project gitignores `accounts.json` for exactly this reason.
2. **It is load-bearing for a running production timer.** `scripts/task-spec-drafter/drafter.sh` hard-depends on both paths — `CLICKUP_CLI=$HOME/.claude/skills/clickup/query.mjs`, `CLICKUP_ACCOUNTS=$HOME/.claude/skills/clickup/accounts.json`, a `[ -f "$CLICKUP_CLI" ] || FATAL` precondition, and the absolute path baked into `DRAFTER_ALLOWED_TOOLS` (`Bash(node /home/zach/.claude/skills/clickup/query.mjs get*)`, ×3). Making that path a store symlink breaks a daily job; making it read-only breaks the token read.
3. **The runtime needs `node_modules`.** Shipping `SKILL.md` alone would propagate a skill to the workbench that triggers and then fails, which is worse than the drift.

Doing this properly is its own PR: relocate credentials to `~/.config/clickup/`, package the Node deps with nix (or add the upstream repo as a flake input rather than vendoring third-party JS into a public repo), update `drafter.sh`'s three hard-coded paths and its allowed-tools string, and re-verify the drafter end-to-end. Landing a half-version would break a working skill and a running timer, so I stopped rather than ship it partly done.

On the `dropStaleClaudeHooks` question you asked me to check: it exists in `nix/home.nix` and is the right precedent — `home.file.force` cannot displace a pre-existing *foreign regular file*, so a pre-switch `rm` step is genuinely needed for this pattern. It just cannot be applied here until the three blockers above are resolved, because the thing that would need removing is a live credential store.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -156,7 +156,18 @@
             # AND turns the version assertion red. That red is the point — the
             # config header's "measured on v1.18.4 — do not re-derive" claims are
             # otherwise pinned to nothing.
-            nativeBuildInputs = [ pyEnv pkgs.bash pkgs.ripgrep pkgs.git pkgs.util-linux pkgs.jq pkgs.gnugrep pkgs.curl pkgs.nodejs pkgs.nix pkgs.opencode ];
+            #
+            # logrotate: scripts/tests/test_claude_log_rotate.py drives the REAL
+            # binary against a temp directory — it asserts that an oversized log
+            # is rotated AND truncated in place (copytruncate), that a small one
+            # is left alone, that the generation cap holds, and 🔴 that the nine
+            # hand-made `.bak` config copies in ~/.claude are never touched. That
+            # last one is a data-loss fence, so it must not be a skipif: those
+            # tests FAIL naming the missing binary, the same deliberate choice as
+            # nix-instantiate and opencode above. Hermetic — logrotate only ever
+            # sees a pytest tmp_path, its own generated config and its own state
+            # file; no network, no ambient /var/lib/logrotate.status.
+            nativeBuildInputs = [ pyEnv pkgs.bash pkgs.ripgrep pkgs.git pkgs.util-linux pkgs.jq pkgs.gnugrep pkgs.curl pkgs.nodejs pkgs.nix pkgs.opencode pkgs.logrotate ];
           }
           ''
             cp -r ${./.} src
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -1916,4 +1916,49 @@ in
       WantedBy = [ "timers.target" ];
     };
   };
+
+  # ~/.claude log rotation. NOT serverMode-gated, deliberately: both hosts run
+  # Claude Code and both accumulate these files. Measured on the workbench
+  # 2026-08-02 — notify.log 39.8 MB (dead writer, last touched 2026-07-06),
+  # clawgate-hook.log 12.0 MB (live), 51.6 MB unrotated in total and growing.
+  #
+  # Only ONE of those writers lives in this repo, which is why the cap is on the
+  # DIRECTORY rather than in each writer: `clawgate-hook.log` comes from the
+  # clawgate approval hook and `notify.log` from something no longer running.
+  # The wrapper uses logrotate's `copytruncate` so an open fd (that hook fires on
+  # every tool call) keeps writing to the same file instead of to an unlinked
+  # inode. Scope is `*.log` ONLY — the hand-made `.bak` config copies in that
+  # directory are never touched. See scripts/claude-log-rotate/rotate.sh.
+  systemd.user.services.claude-log-rotate = {
+    Unit = {
+      Description = "Size-cap the unrotated logs in ~/.claude (logrotate, copytruncate)";
+      OnFailure = [ "notify-failure@%n.service" ];
+    };
+    Service = {
+      Type = "oneshot";
+      Environment = [
+        "PATH=${lib.makeBinPath [ pkgs.logrotate pkgs.bash pkgs.coreutils ]}"
+        "HOME=%h"
+      ];
+      ExecStart = "${pkgs.bash}/bin/bash %h/workspace/devrc/scripts/claude-log-rotate/rotate.sh";
+      # Re-run the unit when the wrapper changes (cf. X-Restart-Triggers above).
+      X-Restart-Triggers = [ "${../scripts/claude-log-rotate/rotate.sh}" ];
+    };
+  };
+
+  # Daily. Persistent catches up a single missed run (host asleep / powered off),
+  # which matters here precisely because the growth is slow and unattended.
+  systemd.user.timers.claude-log-rotate = {
+    Unit = {
+      Description = "Daily timer for the ~/.claude log size cap";
+    };
+    Timer = {
+      OnCalendar = "*-*-* 04:00:00";
+      Persistent = true;
+      RandomizedDelaySec = 600;
+    };
+    Install = {
+      WantedBy = [ "timers.target" ];
+    };
+  };
 }
--- a/scripts/claude-log-rotate/rotate.sh
+++ b/scripts/claude-log-rotate/rotate.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+# Size-cap the unrotated logs in ~/.claude.
+#
+# WHY: nothing rotates them. Measured 2026-08-02 on the workbench:
+#   notify.log         39,822,110 B   (last written 2026-07-06 — a DEAD writer)
+#   clawgate-hook.log  12,033,179 B   (live)
+#   claude-notify.log      60,684 B
+#   daemon.log              6,091 B
+# 51.6 MB total, growing without bound. Only `claude-notify.log` has a writer in
+# this repo (scripts/claude-hooks/claude-notify.py); `notify.log` and
+# `clawgate-hook.log` are written by things OUTSIDE it, which is exactly why the
+# cap has to live at the DIRECTORY, not in each writer.
+#
+# `copytruncate` is load-bearing and not an optimisation: the writers hold open
+# file descriptors, they are not ours to signal, and one of them is a hook that
+# runs on every Claude Code tool call. Renaming the file out from under an open
+# fd would leave it appending to an unlinked inode — the log would appear frozen
+# while silently consuming the same disk. copytruncate copies then truncates in
+# place, so the open fd keeps writing to the same, now-empty, file.
+#
+# 🔴 SCOPE: `*.log` ONLY. It deliberately does NOT touch the 9 stale `.bak`
+# files in that directory (settings.json.bak.*, RULES.md.bak*, PRINCIPLES.md.bak*,
+# CLAUDE.md.bak-*). Those are hand-made safety copies of config; deleting them
+# from an automated job is not this script's call to make. They are listed in
+# the PR for manual removal.
+#
+# Run by hand or via systemd:
+#   systemctl --user start claude-log-rotate.service
+#   scripts/claude-log-rotate/rotate.sh [DIR]
+#   scripts/claude-log-rotate/rotate.sh --print-config [DIR]   # config only, no run
+set -euo pipefail
+
+PRINT_ONLY=0
+if [ "${1:-}" = "--print-config" ]; then
+  PRINT_ONLY=1
+  shift
+fi
+
+DIR="${1:-${HOME}/.claude}"
+
+# Tunables. Defaults chosen against the measured sizes above: a 10M cap turns the
+# 39.8 MB notify.log into one ≤10M live file plus at most 3 compressed
+# generations, and clawgate-hook.log stops being a 12 MB single file.
+SIZE="${CLAUDE_LOG_ROTATE_SIZE:-10M}"
+KEEP="${CLAUDE_LOG_ROTATE_KEEP:-3}"
+STATE="${CLAUDE_LOG_ROTATE_STATE:-${XDG_STATE_HOME:-${HOME}/.local/state}/claude-log-rotate.status}"
+
+emit_config() {
+  # NOTE the quoted glob: logrotate does its own globbing, and an unquoted
+  # pattern here would be expanded by the shell into a fixed file list captured
+  # at config-generation time — so a log created later would never be rotated.
+  cat <<CONF
+"${DIR}/*.log" {
+    size ${SIZE}
+    rotate ${KEEP}
+    copytruncate
+    compress
+    delaycompress
+    missingok
+    notifempty
+    nomail
+}
+CONF
+}
+
+if [ "$PRINT_ONLY" = "1" ]; then
+  emit_config
+  exit 0
+fi
+
+if ! command -v logrotate >/dev/null 2>&1; then
+  echo "claude-log-rotate: logrotate not on PATH — refusing to report success" >&2
+  exit 1
+fi
+
+mkdir -p "$(dirname "$STATE")"
+
+conf="$(mktemp -t claude-log-rotate.conf.XXXXXX)"
+trap 'rm -f "$conf"' EXIT
+emit_config >"$conf"
+
+logrotate --state "$STATE" "$conf"
+echo "claude-log-rotate: applied size=${SIZE} rotate=${KEEP} to ${DIR}/*.log (state=${STATE})"
--- a/scripts/collector/claude/session-tailer.py
+++ b/scripts/collector/claude/session-tailer.py
@@ -29,10 +29,16 @@ transcript takes exactly one of these branches:
 
   unchanged   signature == the one we last emitted for  -> skip (nothing new)
   settled     changed AND idle >= CLAUDE_SUMMARY_SETTLE_MINUTES (default 20)
-              -> EMIT. This is the authoritative final rollup for the session.
-              A session that is later RESUMED (`claude --resume` is real here)
-              changes signature again, settles again, and re-emits — so the
-              latest row is always the complete rollup.
+              AND (never emitted, or last emit older than
+              CLAUDE_SUMMARY_INTERIM_HOURS) -> EMIT. This is the authoritative
+              final rollup for the session. A session that is later RESUMED
+              (`claude --resume` is real here) changes signature again, settles
+              again, and re-emits — so the latest row is always the complete
+              rollup.
+  settled-recently
+              settled, but emitted within CLAUDE_SUMMARY_INTERIM_HOURS -> skip.
+              The new signature is deliberately NOT recorded, so the session
+              still owes a rollup and emits on the first tick past the window.
   first-seen  changed, still active, never emitted before -> EMIT once, so an
               in-flight session (or one interrupted by a reboot) is present in
               the report instead of missing entirely.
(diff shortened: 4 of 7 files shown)


ZacxDev:
## Dev-host tier + merged-tree gate

Both PRs were re-gated after `origin/main` moved to `4eb5798`. **PR #303 has been rebased onto it**; this branch was already based on it.

| tree | sandbox (`nix build .#checks…pytests`) | dev host (`scripts/run-tests.sh`) |
|---|---|---|
| `main` @ `4eb5798` (derived) | 4662 | — |
| #303 alone (rebased) | **4762 / 4761 / 1 / 0** | — |
| **#305 alone (this PR)** | **4688 / 4687 / 1 / 0** | — |
| **#303 + #305 merged onto `4eb5798`** | **4788 / 4787 / 1 / 0** | **4788 / 4787 / 1 / 0** |

Arithmetic closes exactly: 4662 + 100 (#303) + 26 (#305) = 4788. **No cross-PR interaction** — the merged tree is not merely green, its count is the sum, so neither PR's tests were displaced by the other. Both tiers of the merged tree agree, number for number. The single skip is the pre-existing opt-in `repo-cos` live-store drift check.

The merged tree was built as a real integration branch off `origin/main` with both branches merged in, not inferred from `merge-tree`.

⚠️ The dev-host tier needs the flake check's own dependency set to be meaningful. Run bare in a worktree (no `.envrc`) it reports `unparseable summary`, `TOTAL collected=0`, `RESULT: FAIL` — **and still exits 0**. The collected-floor guard is what catches it; an exit code there would have read as a pass on a run where nothing executed.

---

## 🔴 Item (a) — the premise does not hold. Measured, not argued.

The body above scoped (a) out on three technical blockers. I have since **measured the motivating claim directly, and it is false**.

The stated consequence was *"it does not ship to the workbench and silently drifts."* Checked over SSH just now:

```
$ ssh zach@192.168.50.250 'ls ~/.claude/skills/clickup/; stat -c "%y %F" ~/.claude/skills/clickup'
accounts.json  api  flake.lock  flake.nix  lib  listen.mjs  node_modules
package.json  package-lock.json  pnpm-lock.yaml  query.mjs  SKILL.md  test
2026-07-05 10:57:14.142111354 -0500  directory
```

**The workbench already has it** — complete, with `accounts.json` and `node_modules` — at the **same mtime as this host's copy** (this agent is running on the laptop, `192.168.50.155`).

The mechanism is already in the repo and `home.nix` already documents it: `scripts/ship.sh:381` rsyncs `~/.claude/skills/` **workbench → laptop** on every deploy, direction-guarded so the laptop can never clobber the workbench. `home.nix`'s browser-skill comment even calls the nix wiring *"additive to ship.sh's skills rsync"*. So the drift this item was meant to prevent is already prevented, by a working mechanism, in the direction that matters.

That leaves "it is not nix-managed" as true but no longer load-bearing — and the three blockers are unchanged and now dominate. One of them is worse than I wrote in the body:

1. **🔴 The CLI WRITES to its own skill directory.** `lib/accounts.mjs:24` is `const ACCOUNTS_PATH = resolve(SKILL_DIR, 'accounts.json')`, and it is a write path — `writeAccounts()`, `add-account`, plus SKILL.md's *"Team ID, User ID … auto-detected and cached on first use"*. Under a `home.file` deploy that directory is a read-only `/nix/store` symlink, so **first use fails**. It also holds a live ClickUp `pk_` token, and devrc is a public repo.
2. **🔴 `node_modules` becomes unresolvable.** `lib/format.mjs:5` **statically** imports `lib/markdown.mjs`, which imports `unified` / `remark-parse` — so the 3 runtime deps are on the hot path of *every* invocation, not one doc command. Node resolves `node_modules` from the **realpath**; once `query.mjs` is a store symlink it walks up inside `/nix/store` and never finds them. Fixing this means building the deps with nix and shipping source+deps as one derivation — which then *re-breaks* (1), because that derivation is read-only.
3. **It is load-bearing for a running daily timer.** `scripts/task-spec-drafter/drafter.sh` hard-codes `$HOME/.claude/skills/clickup/query.mjs` and `.../accounts.json`, has a `[ -f "$CLICKUP_CLI" ] || FATAL` precondition, and bakes the absolute path into `DRAFTER_ALLOWED_TOOLS` three times.

On the `dropStaleClaudeHooks` / `opencodeDropStaleActivityPlugin` question: that pattern is real and is the right tool for *displacing a stale artifact*, and I would use it here. It solves blocker (1)'s **displacement** half only — it does not make a store path writable, and it has nothing to say about (2) or (3).

**So I did not ship it, and I'd want your call before anyone does.** Doing it properly is its own PR: patch `lib/accounts.mjs` to read/write an XDG path, migrate the token to `~/.config/clickup/accounts.json`, build the 3 npm deps with nix, ship source+deps as one derivation, update `drafter.sh`'s three hard-coded paths and its allowed-tools string, add the `entryBefore ["checkLinkTargets"]` removal step, and re-verify the drafter end-to-end. **Every one of those risks lands at `home-manager switch` time, against a live credential store and a daily production job — and I am constrained from running a switch**, so I could not verify any of it. Shipping it blind to buy a drift guarantee that `ship.sh` already provides is a bad trade; that is my recommendation, but it is your call.

PR #309, opened 2026-08-03
test(gate): run every *.test.mjs suite — 529 node tests no gate had ever run

ZacxDev:
## The defect

`scripts/run-node-tests.sh` hard-coded its collection to one directory:

```bash
FILES=(scripts/browser-bridge/tests/*.test.mjs)
```

so every other `.test.mjs` suite in the repo was invisible to the only check that runs node. Measured at `origin/main` (`13bc8bd`):

| suite | tests | gated before this PR |
|---|---:|---|
| `scripts/browser-bridge/tests` | 468 | yes |
| `scripts/dl-router/tests` | **508** | **no** |
| `scripts/collector/browser-ext/tests` | **21** | **no** |

**529 tests, ungated since each suite was written.** The gate reported `RESULT: PASS` and a 468-test total the whole time — a hard-coded list cannot know what it is not looking at. This is the fourth instance of the shape: #276 (913 pytest tests behind one list entry), #298 (166), #306 (989).

`scripts/collector/browser-ext/tests` was **not in the brief** — it surfaced from auditing every `*.test.mjs` in the repo rather than only the directory named. Its `test_receiver.py` was already gated by `run-tests.sh`; only its two `.mjs` files were orphaned.

## Why not just add two lines

That leaves the *next* suite ungated identically. Collection is now **discovery** (bash globstar over `scripts/**/*.test.mjs`) plus a **two-way pin** (`SUITES`), which fails in both directions:

- a discovered directory absent from `SUITES` → **FATAL** (forces an accounting entry with a measured floor, instead of being swept in under a global total)
- a pinned suite discovery does not find → **FATAL** (the suite vanished)

Discovery alone would reintroduce a hole the hard-coded glob did not have: an emptied `dl-router/tests` would simply collect fewer files and still pass over a global floor.

Each suite also runs in its **own** `node --test` invocation with its own TAP summary and its own floor. A single global floor of 970 is fully satisfied by browser-bridge (468) + dl-router (508) with browser-ext's 21 tests entirely gone — per-suite floors make that loud.

## A portability defect found while building this

The first draft discovered suites with `find -printf '%h\n' 2>/dev/null`. `-printf` is a GNU extension, and **three different `find`s are reachable from this repo**:

| tier | `find` | `-printf` |
|---|---|---|
| bash on the dev host | busybox (`~/.nix-profile/bin/find`) | **rejected** |
| interactive zsh | `bfs` 4.1.1 | ok |
| nix build sandbox | GNU findutils | ok |

Paired with `2>/dev/null`, the rejection became an **empty discovery list with no error** — a false "no suites found", the "an empty result cannot distinguish two mechanisms" trap. Collection is now a bash builtin (globstar), which depends on no external binary and behaves identically in every tier. `test_runner_does_not_use_find_printf` pins it (and strips comment lines first, because the header *describes* the trap and a naive substring search matched that prose).

## Measured — both tiers

| tier | command | result |
|---|---|---|
| authoritative | `nix build .#checks.x86_64-linux.nodetests` | **997 tests / 997 pass / 0 fail**, `RESULT: PASS`, exit 0 |
| dev host | `bash scripts/run-node-tests.sh` | **997 tests / 997 pass / 0 fail**, `RESULT: PASS`, exit 0 |

**Delta accounted exactly:** 468 (baseline) + 508 + 21 = **997**. No unexplained gap.

Baseline at `origin/main` (`13bc8bd`), sandbox: 468 tests / 14 files / `RESULT: PASS`.

## Red/green matrix

`scripts/tests/test_run_node_tests_suites.py`, base ref `origin/main` (`13bc8bd`), measured by swapping `origin/main`'s runner into place:

| revision | result |
|---|---|
| `origin/main` runner | **10 failed / 0 passed** |
| this branch | **0 failed / 10 passed** |

Labelled honestly in the module docstring:

- **regression coverage** — `test_every_test_mjs_directory_is_pinned` (the defect in one assertion), `test_check_suites_accepts_the_real_repo`, `test_no_pinned_suite_has_vanished`
- **invariant guard, not regression coverage** — `test_dl_router_and_browser_ext_are_pinned` (pins that the entries cannot be deleted to turn a future red green; the bug never violated it because the entries did not exist)
- **reachability/mutation proofs** — dropping each pinned suite, and pinning a nonexistent one, each asserting the guard goes red **naming the directory and its specific reason**, so the test cannot be satisfied by a different guard tripping first

**The 529 newly-gated tests are NOT regression coverage for this change.** They are pre-existing tests whose value is that they now run at all.

## Control pair

The reassuring answer for `test_every_test_mjs_directory_is_pinned` is an **empty set** — which an unwired harness also produces. `test_parsers_are_positive_controlled` is the positive control: it asserts both independently-implemented discovery paths (the runner's bash globstar, the test's `pathlib.rglob`) return non-empty, plausibly-sized sets with real floors, so a silently-broken parser fails loudly instead of passing vacuously.

Reported as a pair: **3 suites discovered / 3 pinned / 0 unpinned**, not the zero alone.

## Files touched

- `scripts/run-node-tests.sh`
- `scripts/tests/test_run_node_tests_suites.py` (new)
- `flake.nix` (the `nodetests` header comment claimed the check covered browser-bridge only — a comment is a claim too)

No overlap with `nix/home.nix`, `scripts/bar-status-poll`, or `scripts/collector/**` source (the browser-ext suite is *referenced* by path, not edited).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -189,7 +189,16 @@
           '';
 
       # ---------------------------------------------------------------------
-      # Node test gate — scripts/browser-bridge/tests/*.test.mjs (460 tests).
+      # Node test gate — every scripts/**/*.test.mjs suite (997 tests, measured
+      # 2026-08-03 in this sandbox across three suites: browser-bridge 468,
+      # dl-router 508, collector/browser-ext 21).
+      #
+      # 🔴 It covered ONLY browser-bridge until 2026-08-03, because the runner
+      # hard-coded `FILES=(scripts/browser-bridge/tests/*.test.mjs)`. dl-router's
+      # 508 tests and browser-ext's 21 were gated by NOTHING from the day each
+      # suite was written — the #276/#298/#306 shape in a fourth costume. The
+      # runner now DISCOVERS suites and pins them both ways; read its header
+      # before changing this.
       #
       # Deliberately a SEPARATE check from `pytests`, not folded into it:
       #  • distinct toolchain (nodejs vs python312) — keeping them apart means
--- a/scripts/run-node-tests.sh
+++ b/scripts/run-node-tests.sh
@@ -10,44 +10,180 @@
 # The caller is responsible for putting `node` on PATH (the flake check does this
 # via the derivation's nativeBuildInputs). This script only ORCHESTRATES.
 #
-# The suite under test is scripts/browser-bridge/tests/*.test.mjs. It is HERMETIC:
-# audited 2026-08-02 — no `createServer`/`.listen()`, no `spawn`/`execFile`, every
-# `fetch` is a stub assigned onto `globalThis.fetch`, and the only I/O is reads of
-# repo-relative files (extension/protocol.js, browser-agent, server.py) plus writes
-# under `tmpdir()`. Nothing reaches the network or a subprocess.
+# 🔴 THIS FILE USED TO HARD-CODE ONE DIRECTORY.
+#    `FILES=(scripts/browser-bridge/tests/*.test.mjs)` was the entire collection
+#    step, so `scripts/dl-router/tests/*.test.mjs` (508 tests) and
+#    `scripts/collector/browser-ext/tests/*.test.mjs` (21 tests) had NEVER been
+#    run by any gate — 529 tests, ungated since each suite was written. Exactly
+#    the shape of #306 (989 ungated pytest tests) and #298 (166) and #276 (913),
+#    one language over: a target list that only grows when somebody remembers.
 #
-# 🔴 TWO STRUCTURAL GUARDS — both exist because a green exit code lies here:
+#    The fix is not "add two more lines to the list" — that leaves the next suite
+#    ungated the same way. Collection is now DISCOVERY (every `*.test.mjs` under
+#    `scripts/`), so a new suite is gated the moment it exists.
+#
+# 🔴 FOUR STRUCTURAL GUARDS — all four exist because a green exit code lies here:
 #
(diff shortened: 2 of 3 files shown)


ZacxDev:
### Gated on the MERGED tree, not just this branch

Per RULES.md — a PR green on its own branch proves nothing about the tree its merge creates. Built `integration/309-310` off `origin/main` (`13bc8bd`), merged **both** #309 and #310, and ran the full authoritative gate there:

| check | result |
|---|---|
| `nix build .#checks.x86_64-linux.pytests` | **5810 collected / 5809 passed / 1 skipped / 0 failed** — `RESULT: PASS` |
| `nix build .#checks.x86_64-linux.nodetests` | **997 tests / 997 pass / 0 fail** (3 suites, 29 files) — `RESULT: PASS` |

Combined exit `0`.

**Delta accounted exactly:** 5792 (baseline) + 10 (#309's `test_run_node_tests_suites.py`) + 8 (#310's `test_run_tests_preconditions.py`) = **5810**. Node: 468 (baseline) + 508 + 21 = **997**. No unexplained gap on either.

`git merge-tree --write-tree` between the two branches exits **0** (branched on the exit code, not a marker grep — `merge-tree` prints only a tree OID on success and emits no conflict markers either way). The branches also share **no files**: #309 touches `scripts/run-node-tests.sh`, `flake.nix`, `scripts/tests/test_run_node_tests_suites.py`; #310 touches `scripts/run-tests.sh`, `scripts/tests/test_run_tests_preconditions.py`. The only semantic coupling is that both add a file under `scripts/tests/`, which is a `HERMETIC_TARGETS` entry — and that is what the 5810 total confirms.

PR #313, opened 2026-08-03
docs(rules): five multi-agent lessons, with evidence evicted to the archive

ZacxDev:
Durable lessons measured in one long multi-agent session. Four new rules, one widening.

## New rules

| Section | Rule |
| --- | --- |
| Git Workflow | `isolation: "worktree"` builds the worktree from your **current** repo, not the repo the task names. Two agents dispatched at a TS monorepo while cwd was a Go repo both got Go worktrees — one refused and reported (the good outcome), one silently rebuilt its own. Cross-repo work must not pass the flag. |
| Green Test Suite | A green integration gate is evidence about **the base it ran on**. A gate that merged two PRs and passed the full suite went stale before landing: `main` advanced 3 commits, one rewriting **779 lines of the router file both PRs touch**, while both PRs reported `MERGEABLE` throughout. |
| Verification Honesty | A field that exists in a DTO **is not a guard** — only a branch on it is. A `notOwned` flag declared in the consumer's own type and never read rendered a fully zeroed dashboard as if it were measured data. |
| A test you have not watched FAIL | New representative case for the output-format class: `prettier --check --stdin-filepath <path>` prints `(stdin)`, **not** the path, so a path-grep reported a false CLEAN. Caught only by a negative control. |

## Widened

The `pkill -f` rule now covers the parallel-agent case explicitly. A box-wide pattern reaches a **sibling agent's** processes: an auditor clearing its own hung run killed ~15 PIDs and destroyed another agent's in-flight test run, whose next attempt then collapsed with `0 files collected` and `exit 144` — both readable as code defects rather than as the kill. Filter resolved PIDs by `/proc/<pid>/cwd` before killing.

## Budget

`scripts/tests/test_rules_size.py` leaves 495 B of free budget; these additions are ~1.9 KB. Rather than ratchet `MAX_BYTES`, ~1.4 KB of **evidence** moved to `claude/RULES-ARCHIVE.md`:

- new anchors: `cross-repo-worktree`, `stale-gate-base`, `count-tests`, `sibling-agent-kill`, `cheap-control`
- appends to: `parsing-tool-output`, `base-clone-drift`

**No rule lost its imperative, priority marker, triggers, widest scope, or procedure.** What moved out is worked-incident detail, enumerated ground cases where one representative carries the point, and two ✅/❌ blocks that restated their own bullet verbatim.

`RULES.md` 33,105 → 33,507 B (993 B headroom against a 900 B floor).

## Verification

- `nix build .#checks.x86_64-linux.pytests` — `TOTAL collected=5868 passed=5867 skipped=1 failed=0` (floor 5600); `scripts/tests collected=1068 passed=1068`.
- The size gate was **mutation-tested before its green was trusted**: appending 2000 B reddens `test_rules_md_under_hard_ceiling`, and a bogus `→ archive: no-such-anchor` tag reddens `test_every_archive_pointer_resolves` — each failing on its own assertion, not a neighbour's. The file was restored byte-identically afterwards (`cmp`).

⚠️ `~/.claude/RULES.md` resolves to `/nix/store/…-hm_RULES.md` (a read-only copy), so **this change is inert until a `home-manager switch` / `ship.sh`**. Not run here.

https://claude.ai/code/session_01858ymA3tEJQi83435u7npi


ZacxDev:
**Rebased onto current `main` (`a101812`) — and the rebase reduced this to one rule.**

This branch was opened 2026-08-03. `main` has moved a long way since, and I checked each addition against it rather than resolving conflicts mechanically:

**Eight of the nine core rules have already landed via other PRs** — `proven REACHABLE`, `Mutation-test a guard before certifying`, `the cheap discriminating control BEFORE the plausible theory` (that one came in via #231 today), `gh pr view … is the ONLY authority`, `Worktree isolation is the standing default`, `Re-sync the base clone`, `pgrep/pkill match your OWN shell`, and `readlink -f is the arbiter`. Keeping them would have been a no-op at best and a duplicate statement at worst.

**All five archive anchors would now be orphans.** `main`'s core routes to 30 anchors and **none** of them are `cheap-control`, `stale-gate-base`, `count-tests`, `cross-repo-worktree` or `sibling-agent-kill` — and the same evidence is already archived under `count-not-exit-code`, `worktree-envrc`, `isolation-seam` and `stale-observation`. Adding unrouted anchors is exactly what `/prune-skill` calls ORPHANED.

**What survives is the one genuinely new rule:** *a field that exists in a DTO is not a guard — only a BRANCH on it is.* That one is not on `main` in any form, and its measured case is good: a payload's `notOwned` was declared in the consumer's own type and never read, so the not-owned case rendered a zeroed dashboard as measured data — and "just set the field correctly" would have changed nothing on screen.

**Numbers:** `RULES.md` 32,540 → 32,971 B (+431). Headroom 1,060 → **629 B** against the 33,600 usable budget. Gate `test_rules_size.py`: **5 passed** (counted, not read off an exit code). One file, one line added, nothing else touched.

⚠️ Ships to every host and resolves into `/nix/store` — needs `scripts/ship.sh` after merge.

Force-pushed with `--force-with-lease`; previous tip was `ea82146`.

PR #318, opened 2026-08-03
feat(browser-agent): make `emulate` reachable and DEFAULT-ON — its exclusion rationale was factually wrong (#316)

### LINKED ISSUE #316: browser-bridge agent: `emulate` is unreachable, so the autonomous agent cannot run width-dependent passes
## Problem

The autonomous `browser agent` cannot run any **width-dependent (responsive) verification
pass**, because `emulate` is unreachable to it — including via the documented escape hatch.

This is not a config gap that can be worked around. `emulate` is absent from **both**
`ALLOWED_OPS_DEFAULT` *and* `OP_TO_SERVER`, so `BROWSER_AGENT_ALLOWED_OPS` — the
"only supported way to re-enable an off-by-default op" — cannot reach it either. An op
with no `OP_TO_SERVER` mapping has nothing to route to.

## Evidence (measured, not inferred)

- `opencode/tools/browser_tool_impl.mjs:121-124` — `ALLOWED_OPS_DEFAULT` is
  `text, html, eval, nav, screenshot, frames, click, type, key, wake, context, whoami`.
  No `emulate`.
- `opencode/tools/browser_tool_impl.mjs` — `grep -n emulate` returns **only** the comment
  at `:64-73`. It is not a key in `OP_TO_SERVER`.
- `reference/emulation.md:461-469` and `README.md:620-627` both already document this as
  deliberate, and both already scope the fix: *"Adding it would mean mapping
  `emulate → emulate` in `OP_TO_SERVER` and listing it in `ALLOWED_OPS_DEFAULT`; the
  ownership gate already confines it to the agent's own tab, so the question is scope
  discipline, not safety."*

## What it blocks (concrete, today)

A production verification pass that had to check a responsive layout regression at **two
viewport widths** (~1200 and 1920). The width-dependent assertions are the entire point of
that class of check — a layout rollup that drops below a container threshold simply does
not manifest at whatever width the operator's window happens to be.

Everything else in the pass was reachable by the agent (DOM/`eval` assertions, navigation,
click-through state round-trips). The pass had to be split and partly hand-driven purely
because of the viewport. This is a recurring shape, not a one-off: any responsive or
breakpoint regression check is off-limits to the cheap autonomous path.

## Why it is currently excluded — and what actually changes for the agent

The exclusion rationale at `browser_tool_impl.mjs:64-73` is sound on its own terms:

> `emulate` … MUTATES the tab and leaves STICKY PER-TAB STATE THAT OUTLIVES THE OP …
> persist on the tab until they are explicitly reset (`emulate --reset`) **or the tab is
> replaced**. An autonomous model that emulates and then does not reset — because it moved
> on, hit its step budget, or crashed — hands the operator back a silently altered tab.

The load-bearing premise there is *"hands the operator back a silently altered tab."*
**On the agent path specifically, that premise does not hold**, because the agent never
hands a tab back:

- `browser-agent:360-367` — the wrapper opens its **own** tab and `cleanup()` closes it.
- `browser-agent:373` — `trap _cleanup_all EXIT INT TERM`, so the close fires on normal
  exit, step-budget exhaustion, and INT/TERM alike.
- The impl's own comment already concedes the state dies when *"the tab is replaced"*.

So the sticky-state blast radius on this path is bounded by a tab lifetime that the wrapper
already owns and already tears down. The genuine residual is narrower than the comment
implies:

1. **SIGKILL / hard crash bypasses the trap** → an orphaned emulated tab survives. Real,
   but it is the same orphan class the wrapper already accepts for the tab itself.
2. **The CDP attach raises Brave's "an extension is debugging this browser" banner** for
   the duration. That is a visible-to-operator annoyance, and it is not undone by closing
   the tab any faster than the tab close itself.

Point 2 is the honest cost. Point 1 is a hardening question, not a blocker.

## Options

1. **Allow it, leaning on the existing teardown.** Map `emulate → emulate` in
   `OP_TO_SERVER`, add it to `ALLOWED_OPS_DEFAULT`, and add an explicit
   `emulate --reset` to `cleanup()` *before* the tab close — belt-and-braces so the reset
   does not depend solely on the close, and so a future change to tab reuse cannot silently
   reintroduce the hazard.
2. **Allow it opt-in only.** Map it in `OP_TO_SERVER` but leave it out of
   `ALLOWED_OPS_DEFAULT`, so `BROWSER_AGENT_ALLOWED_OPS` can enable it deliberately —
   exactly the `upload` precedent. Cheapest change; keeps the default surface minimal;
   makes responsive passes a conscious per-invocation decision.
3. **Status quo**, and responsive passes stay operator-only. Worth choosing explicitly
   rather than by omission, since the docs currently present it as an open scope question.

Option 2 looks like the best cost/benefit — it matches an established pattern in this repo,
needs no default-surface widening, and the teardown analysis above says the residual risk
at opt-in time is small. Option 1's extra `--reset` in `cleanup()` is worth doing either
way.

## Implementation constraint (whichever option)

The op set is asserted identical across **four** places by
`tests/browser_tool.test.mjs:884` (`OP-SET PARITY`):

- `opencode/tools/browser.js` typed `op` enum (`:44-47`)
- `opencode/tools/browser_tool_impl.mjs` `ALLOWED_OPS_DEFAULT` (`:121-124`)
- the agent-md capability table (`opencode/browser-agent.md`)
- the README op list

Option 2 (opt-in, not in the default set) may leave the parity set untouched — worth
checking against how `upload` is handled there, since `upload` is deliberately absent from
all four while still reachable via `OP_TO_SERVER`.

## Separable — not part of this ask

The agent also never receives screenshot bytes (`SCREENSHOT_NOTE_MAX = 200`,
*"never spill a base64 image blob into the model"*). That is a defensible, independent
design choice, and it means the agent is blind to any purely visual judgement even if
`emulate` were granted. Flagging it only so the two are not conflated: granting `emulate`
buys **measurable** width-dependent assertions (`getBoundingClientRect`, computed styles,
presence/absence of elements), not visual review.

ZacxDev:
Closes #316. Implements the issue's **Option 1**: `emulate` is reachable by the autonomous `browser agent` and **default-on**.

## The finding: the exclusion rationale was factually wrong

`opencode/tools/browser_tool_impl.mjs:64-73` justified withholding `emulate` like this:

> Unlike every read op it MUTATES the tab and leaves **STICKY PER-TAB STATE THAT OUTLIVES THE OP**: device metrics, UA-CH and media overrides persist on the tab until they are explicitly reset (`emulate --reset`) or the tab is replaced. An autonomous model that emulates and then does not reset … hands the operator back a silently altered tab.

`extension/protocol.js:557-580` documents the **opposite**, as the deliberately-chosen safety property of this design:

> Because the overrides die at detach, **the tab is NOT emulated between ops**. A crashed agent, a killed Claude session, or an evicted MV3 service worker therefore **CANNOT** leave the operator's real browser distorted. The worst case is a forgotten entry in an in-memory Map that dies with the worker …

CDP Emulation overrides are session-scoped; `withCdpSession` always detaches in its `finally`; `emulate` stores a normalized state in a Map and **every** op re-applies it inside its own session. The alternative — holding one long-lived debugger session open — is precisely the design that would have had the sticky failure mode the comment described, and protocol.js says it was rejected for that reason. So the exclusion rested on a hazard the code was built to make impossible.

The blast radius that *does* exist is already bounded, server-side rather than by convention: `server.py`'s `OWNED_TAB_ONLY_OPS = {"emulate"}` refuses the op with `not_owned_tab` on any tab the calling session did not `open`; the `browser-agent` wrapper opens the run's tab under the run's own `SESSION_ID` (`browser-agent:601,:638`) and closes it on every exit path (`trap _cleanup_all EXIT INT TERM`, `:373`). The one real residual is the transient "an extension is debugging this browser" banner from the CDP attach — which `eval`, `screenshot` and `wake` already raise on every call. Like those, `emulate` now counts as MUTATING for `BROWSER_AGENT_DRY_RUN`, so a dry run never attaches the debugger.

## Why Option 2 (opt-in) was structurally impossible

The issue offered "make it reachable but opt-in via `BROWSER_AGENT_ALLOWED_OPS`", i.e. `emulate` in `OP_TO_SERVER` but not in `ALLOWED_OPS_DEFAULT` — the `upload` shape. That shape **cannot be made invocable here**:

- The model is bound by `browser.js`'s typed `op` enum. An op absent from the enum cannot be called no matter what the env allows.
- `tests/browser_tool.test.mjs`'s `OP-SET PARITY` test asserts `deepEqual` in **both directions** between that enum and `ALLOWED_OPS_DEFAULT` (plus the agent-md table and the README list).

So "in the enum" ⇒ "in `ALLOWED_OPS_DEFAULT`" ⇒ default-on. `upload` gets away with the opt-in shape only because it is absent from the enum too, and an opted-in `upload` therefore relies on opencode not enforcing the enum — that is a separate, unpinned question (issue's deferred `upload` fix). Opt-in-*and*-invocable would have meant weakening the parity test, i.e. deleting the guard that catches exactly this class of drift. Option 1 it is.

## Changes

- **`opencode/tools/browser_tool_impl.mjs`** — `emulate: "emulate"` in `OP_TO_SERVER`; `"emulate"` in `ALLOWED_OPS_DEFAULT` (13 default ops); the exclusion comment rewritten to state the truth and cite protocol.js; a typed-whitelist `buildRequest` branch (`device`, `width`, `height`, `deviceScaleFactor`, `mobile`, `maxTouchPoints`, `userAgent`, `timezone`, `orientation`, `colorScheme`, `reset` → the wire's `dsf`/`ua`/`tz`), enforcing `EMULATION_LIMITS` client-side with protocol.js's own `invalid_emulation:<field>` and `emulate_needs_device_or_params` vocabulary; a field-pinned `summarizeResult` branch; `emulate` added to the dry-run MUTATING set.
  - `geo` and `touch` are deliberately **not** forwarded to the agent: location spoofing is unrelated to the viewport question and is a fingerprinting surface, and touch is implied by a preset or by `maxTouchPoints`.
  - The bounds are a **mirror**, not an import: the wrapper copies only `browser.js` + `browser_tool_impl.mjs` into the per-run scratch project, so importing `../../extension/protocol.js` would `MODULE_NOT_FOUND` in every real run. A new test imports both and pins them equal, so the mirror cannot drift silently.
- **`opencode/tools/browser.js`** — `emulate` in the typed `op` enum, all eleven emulation fields in the `args` schema (without them the op is listed and unusable — the actual point of the issue), description updated. The "There is NO 'upload' op" clause is untouched.
- **`opencode/browser-agent.md`**, **`README.md`**, **`reference/emulation.md`** — capability-table row, published `op` ∈ {…} list, and both now-false "not available to the autonomous agent" rationale sections corrected.
- **`tests/browser_tool.test.mjs`** — the `emulate` entry deleted from `REVIEWED_AGENT_EXCLUSIONS` (the `DECLARED EXCLUSIONS` test asserts a declared exclusion is *not* reachable), the two pinned op-name lists updated, and 12 tests added.
- **`SKILL.md` untouched** — 11,845 bytes before and after (budget 12,038). Its two `emulate` mentions make no claim about agent reachability, so no eviction was needed.

## Test counts

| suite | before | after |
|---|---|---|
| `node --test --test-reporter=tap scripts/browser-bridge/tests/*.test.mjs` | 468 tests / 468 pass / 0 fail / 0 skip | **480 / 480 / 0 / 0** |
| `pytest scripts/browser-bridge/tests -q` | — | **436 passed, 0 skipped, 0 failed** |

`AGENT SURFACE PARTITION`, `UPSTREAM ANCHOR` (both) and `DECLARED EXCLUSIONS` all pass — `emulate` moved from the declared-exclusion side of the partition to the reachable side, which is exactly what those tests are shaped to allow.

## Red/green matrix

Every new test was watched fail. The implementation surface was mutated one piece at a time (test file never touched, no `git stash`), the suite re-run, and the reds recorded. Stage 0 and the final restore stage are the controls (83/83 green in `browser_tool.test.mjs`).

| stage (mutation of the implementation) | new tests that went RED |
|---|---|
| 0 — HEAD (control) | none — 83/83 pass |
| 1 — `browser_tool_impl.mjs` at `origin/main` | whole file fails to link (the mirror constants do not exist upstream) — 0/1 |
| 2 — `emulate` removed from `OP_TO_SERVER` + `ALLOWED_OPS_DEFAULT` | 13 red: all 6 `buildRequest: emulate…`, both `runBrowserOp: emulate…`, `OP-SET PARITY` ×2, `UPSTREAM ANCHOR`, `AGENT SURFACE PARTITION`, `OP_TO_SERVER maps only the bounded ops` |
| 3 — `buildRequest` emulate branch removed | 6 red: `DEFAULT-reachable`, `raw width/height`, `reset path`, `NO parameters`, `EMULATION_LIMITS bounds`, `NO raw-CDP passthrough` |
| 4 — bounds mirror drifted (`maxDimension` 10000→99999) | 2 red: `MIRROR PARITY`, `EMULATION_LIMITS bounds` |
| 5 — `summarizeResult` emulate branch removed | 2 red: `summarizeResult: emulate is field-pinned`, `runBrowserOp: emulate reaches the bridge` |
| 6 — `emulate` dropped from the dry-run MUTATING set | 1 red: `runBrowserOp: emulate is MUTATING → a dry-run never attaches the debugger` |
| 7 — `browser.js` at `origin/main` | 2 red: `OP-SET PARITY` ×2 |
| 8 — `browser.js` enum kept, emulate **arg fields** removed | 1 red: `OP-SET PARITY: emulate appears in ALL FOUR…` (this is the "listed but unusable" case) |
| 9 — agent-md capability row removed | 2 red: `OP-SET PARITY` ×2 |
| 10 — README `op` ∈ {…} loses emulate | 2 red: `OP-SET PARITY` ×2 |
| 11 — restored (control) | none — 83/83 pass |

One new test, `runBrowserOp: a narrowed BROWSER_AGENT_ALLOWED_OPS still refuses emulate`, went red in **no** stage. It passes on pre-change code too (for the different reason that `emulate` was unmappable), so it is an **invariant guard, not regression coverage** — it is labelled as such in a comment above it rather than counted as a red/green win.

## Harness positive control

The four-source parity tests parse regexes and can pass vacuously if a pattern matches nothing. Removing the `emulate` row from **one** source (stage 9, `opencode/browser-agent.md`) produced:

```
not ok 76 - OP-SET PARITY: `emulate` appears in ALL FOUR agent-facing sources (#316)
  error: 'agent-md capability table'
  code: 'ERR_ASSERTION'
  expected: true
  actual: false
```

— the assertion names the specific missing source, so the parser is demonstrably matching real text, not returning empty. Stages 7, 8 and 10 do the same for `browser.js` (twice, enum and arg fields) and the README.

## NOT verified

**No live two-width browser pass has been run.** Nothing in this branch was exercised against a real Brave — no tab was opened, no viewport was emulated, no page was read at 390px and again at desktop width. Every claim above is from unit tests with an injected `fetchImpl` plus reading the extension/server source. Confirming that a `browser agent` run can actually call `emulate` and get a differently-rendered page is the operator's acceptance test.

Also unverified by this branch (unchanged behaviour, stated for completeness): whether opencode enforces the tool's `op` enum at runtime — the reason the Option-2 shape is unpinned for `upload`.

Deliberately **not** bundled, per the issue: the `upload` fix and an `emulate --reset` in the wrapper's `cleanup()`.


ZacxDev:
Closing unmerged. The live two-width acceptance pass confirmed the capability but **falsified this PR's central argument**. Full correction: #316 (comment) · root cause: #319.

**What held up:** `emulate` reaches the browser through the agent surface and width-dependent measurement works — 1124 → 390 with `matchMedia('(max-width: 600px)')` flipping, against a 1124 control tab in the same window. The Option-2-is-impossible analysis also held under mutation testing (removing only the arg fields while keeping the enum → exactly 1 red, naming `width`).

**What did not:** this PR's headline was "the exclusion rationale was factually wrong, because overrides die at detach (`extension/protocol.js:557-580`)." That documented safety property is false as measured. Emulation is permanently sticky per tab — it survives detach, survives `--reset`, and survives a re-nav; reproduced through the operator CLI with none of this branch's code in the path. `Emulation.clearDeviceMetricsOverride` is never sent anywhere (#319).

So the original exclusion comment this PR deleted was substantially **right**, and the branch rewrote `browser_tool_impl.mjs`, `README.md` and `reference/emulation.md` to assert the false property. Merging would replace a true comment with a false one — worse than the gap it closes, since the next maintainer reasons from that comment.

Closing rather than patching the prose: the argument needs rebuilding on the correct footing (blast radius bounded by the wrapper's tab teardown, not by detach semantics), and the redo should land after #319 is settled so the residual can be stated honestly — a `SIGKILL`'d agent leaves a permanently distorted tab that `--reset` cannot repair.

Branch `feat/browser-agent-emulate-316` is left in place; the code change and its 12 tests are reusable as-is, and the mutation matrix in the body above stays valid. It is only the rationale prose that has to be inverted.

Worth recording: 480 node + 436 pytest tests, all green, could not see this. Every emulate test injects a `fetchImpl` and never observes a real tab.

PR #320, opened 2026-08-03
fix(browser-bridge): `emulate --reset` now CLEARS the CDP overrides — the viewport was permanently sticky (#319)

### LINKED ISSUE #319: browser-bridge: the emulated VIEWPORT was permanently sticky per tab — clearDeviceMetricsOverride is a no-op unless the session ARMED an override first
## Summary

`browser emulate --reset` does not un-emulate the tab. Device-metrics emulation is **permanently sticky per tab**: it survives the CDP detach, survives `--reset`, and survives a subsequent navigation. The only remedy is closing the tab.

This falsifies the safety property `extension/protocol.js:557-580` documents as the reason the current design was chosen over a long-lived debugger session.

## Measured (2026-08-03, laptop, live Brave, extension 0.7.1, `personal` profile)

Reproduced entirely through the **operator CLI** — no agent, no autonomous model, no PR-branch code:

```
1. before emulate                innerWidth = 1124
2. browser emulate iphone-15
3. after emulate                 innerWidth = 393
4. browser emulate --reset
5. AFTER --reset                 innerWidth = 393   ← not restored
6. after --reset + re-nav        innerWidth = 393   ← a fresh document does not restore it either
```

Control, same Brave window, a tab that was never emulated: `innerWidth = 1124`. So the stickiness is real and per-tab, not a mismeasurement of the window.

A second control ruled out the benign explanation — that the document had merely been *committed* at 390 and would recover on navigation. It does not recover (step 6). The override is still being applied to a document created after the reset.

## Root cause

`Emulation.clearDeviceMetricsOverride` appears **nowhere** in the extension or `server.py`:

```
$ grep -rn "clearDeviceMetricsOverride" extension/ server.py
(no matches)
```

`extension/service_worker.js:1184` states the premise explicitly:

> `--reset` means "STOP RE-APPLYING". It does not need to undo anything: the …

That is sound *only if* the overrides die at debugger detach. Empirically they do not — `Emulation.setDeviceMetricsOverride` persists on the tab after detach. So `--reset` deletes the in-memory `emulationState` entry and sends nothing to CDP, leaving the tab overridden with nothing left that knows how to undo it.

## What this invalidates

`extension/protocol.js:557-580`, the stated justification for the store-and-re-apply design:

> Because the overrides die at detach, the tab is NOT emulated between ops. A crashed agent, a killed Claude session, or an evicted MV3 service worker therefore **CANNOT** leave the operator's real browser distorted. The worst case is a forgotten entry in an in-memory Map that dies with the worker — not a Brave tab stuck at 393×852 pretending to be an iPhone with nothing left running that knows how to undo it.

The last clause is exactly the observed state. The comment describes the failure mode of the design it says was rejected.

Also now false, and user-facing:

- the `emulate` response `note`: *"Between ops the tab is NOT emulated (the overrides die at detach), so a crashed agent cannot leave your browser distorted."*
- the `--reset` response `note`: *"emulation stopped. Nothing had to be undone — CDP overrides die at debugger detach, so the tab was already un-emulated between ops."*
- `reference/emulation.md`'s safety-property section.

## Impact

- **Operator:** any `browser emulate` on a tab you keep leaves it silently stuck at the emulated viewport. `--reset` reports success and changes nothing. Easy to attribute to the site.
- **Agent path:** bounded but not eliminated — the `browser-agent` wrapper closes its own tab on every exit path (`trap _cleanup_all EXIT INT TERM`, `browser-agent:373`), and closing is the one thing that does clear it. A `SIGKILL`'d run bypasses the trap and leaves a permanently distorted tab that `--reset` cannot repair.
- The read-path annotation (`emulated: true` / `notEmulatedRead`) is keyed on the `emulationState` Map, so after a `--reset` a still-emulated tab reports **no** emulation fields at all — the envelope actively says the read is un-emulated while the viewport is overridden. That is the confident-wrong-answer shape.

## Suggested fix

Send `Emulation.clearDeviceMetricsOverride` (plus the matching clears for touch / UA / media / timezone overrides that were applied) in the `--reset` path, rather than only dropping the Map entry. Then correct the four claims above to describe what the code does.

🔴 A regression test for this must assert the **viewport after reset**, not that reset returned `ok:true` — the current behaviour returns `ok:true`, `reset:true` and a populated `wasEmulating` while doing nothing to the tab. Only a post-reset `innerWidth` read discriminates.

## Provenance

Found by the live acceptance test for #316 / #318 — the two-width verification pass that PR asked for. The unit suite (480 node + 436 pytest, all green) could not see this: every emulate test uses an injected `fetchImpl` and never observes a real tab. This is the deploy-vs-verify gap, not a test-count gap.

ZacxDev:
Fixes #319. Related: #316 (the two-width verification pass that found it), #318.

`browser emulate --reset` deleted the in-memory `emulationState` entry and sent
**nothing** to CDP. The premise — asserted in five places as *the* safety property
of the store-and-re-apply design — was that every `Emulation.*` override dies at
the debugger detach. It does not. The device-metrics **viewport size** survives,
so `--reset` reported success while leaving the operator's tab permanently
phone-sized, with `emulationState` empty and nothing left that knew how to undo it.

---

## STEP 1 — which overrides actually survive a `--reset`

#319 measured `innerWidth` only. The fix's step list depends on the rest, so I
measured every class before writing it.

**Method.** 2026-08-03, laptop (`192.168.50.155`), live Brave, extension **0.7.1**,
`personal` profile, operator CLI only (no agent, no PR-branch code — the deployed
extension is unchanged `main`). `https://example.com` in a tab this session opened.
`emulate iphone-15 --color-scheme dark --tz Europe/London --geo 51.5074,-0.1278`,
then a re-`nav` under emulation so the create-time touch API applies. **Every read
taken the same way** (`js --wake` → the CDP path), so the columns are comparable.

🔴 **CONTROL**: a second tab, opened in the same window at the same URL under a
*different* session id, never emulated, read through the identical `js --wake`
path — before and after the whole sequence. It read desktop both times, which is
what makes a "still 393" result on the test tab attributable to the tab rather
than to a broken read or a resized window.

| signal | **CONTROL** (never emulated) | (a) before emulate | (b) after emulate + re-`nav` | (c) after `--reset` | (d) after `--reset` + re-`nav` | verdict |
|---|---|---|---|---|---|---|
| `innerWidth`x`innerHeight` | **1124x1400** | 1124x1400 | 393x852 | 🔴 **393x852** | 🔴 **393x852** | **PERSISTS** |
| `devicePixelRatio` | 1 | 1 | 3 | 1 | 1 | clears at detach |
| `navigator.maxTouchPoints` | 0 | 0 | 5 | 0 | 0 | clears at detach |
| `"ontouchstart" in window` | false | false | true | true¹ | false | document-build residue |
| `matchMedia("(pointer:coarse)")` | false | false | true | false | false | clears at detach |
| `prefers-color-scheme: dark` | false | false | true | false | false | clears at detach |
| `navigator.userAgent` | `X11; Linux x86_64` | `X11; Linux x86_64` | `iPhone; CPU iPhone OS 17_0` | `X11; Linux x86_64` | `X11; Linux x86_64` | clears at detach |
| `Intl…resolvedOptions().timeZone` | `America/Winnipeg` | `America/Winnipeg` | `Europe/London` | `America/Winnipeg` | `America/Winnipeg` | clears at detach |

The control read `1124x1400` again after the sequence finished — the same value as
at the start.

¹ Not an override. The *document* was built under touch emulation and
`ontouchstart` is installed on the global at document creation, so no CDP clear
can remove it; the re-`nav` in column (d) does. Same asymmetry as the
`documentPredatesEmulation` hint, in the opposite direction.

**So exactly one class is sticky: the device-metrics viewport size.** Note that
`devicePixelRatio` — set by the *same* `setDeviceMetricsOverride` call — **does**
revert, which points at the residue being the resized **render widget** rather
than a live override that outlived its session.

Not measured: **geolocation** (reading it requires a permission grant, which would
have meant a prompt on the operator's screen). It is cleared by the fix anyway.

🔴 Both tabs opened for this were closed, the emulated one last. `browser tabs`
afterwards: no leftover tabs, `emulatedTabs: []`.

---

## STEP 2 — the fix

`extension/protocol.js` (pure, browser-free, matching the file's existing split):

* `emulationResetCdpSteps(state)` — the ordered CLEAR list, **derived from the
  state that was in force** rather than a fixed list, so a reset sends exactly the
  undo of what this bridge applied. `Emulation.clearDeviceMetricsOverride` goes
  **first** (it is the one measured to be stuck, so it lands even if a later step
  fails), then `setTouchEmulationEnabled{enabled:false}`,
  `setUserAgentOverride{userAgent:""}`, `setEmulatedMedia{media:"",features:[]}`,
  `setTimezoneOverride{timezoneId:""}`, `clearGeolocationOverride`.
* `applyEmulationResetSteps(send, steps)` — **best-effort**, the deliberate
  opposite of `applyEmulationSteps`. An *apply* must fail loudly (a half-applied
  emulation is a plausible screenshot of the wrong thing); an *undo* must not —
  aborting at the first failure leaves the remaining overrides in place, which is
  the state the undo exists to escape. Returns `{cleared, failed}`.

`extension/service_worker.js` — the browser-facing half. The Map entry is dropped
**first and unconditionally**; the clears then run in one CDP session. A refused
attach or a closed tab is caught and reported as `clearError`, an individual
rejected step as `clearFailed[]` — reset never throws, and never leaves a tab
still being re-emulated. A tab with **no** stored state yields no steps, so reset
stays a zero-CDP, zero-debugger-banner no-op there (the existing test for that is
unchanged and still passes).

**Why the five self-clearing classes are cleared anyway**, stated plainly: they
are not needed by this measurement, but that measurement is one Chromium build on
one host (RULES: one measurement is not a general claim), the clears run inside a
session that is already open, and a clear against a domain holding no override is
a no-op. "Undo what was applied" is worth having independent of which Chromium you
are on. The code comment says which one is measured-necessary and which five are
defensive, so nobody later mistakes the second group for evidence.

**Honest limitation, now documented in the code:** `--reset` derives its clears
from the Map, so it **cannot repair a tab whose state died with an evicted MV3
service worker** — closing the tab remains the remedy there, and that is still the
`SIGKILL`'d-agent path #319 describes.

Extension **0.7.1 → 0.7.2**. `tests/test_server.py`'s version literal is bumped in
the same commit, as that test's own comment requires.

---

## STEP 3 — the false claims, corrected

Each rewritten from what the code now does, not from intent:

* `extension/protocol.js` EMULATION header (the old `:557-580`) — the "overrides
  die at detach → a crashed agent CANNOT leave the browser distorted" block. Now
  states which half dies and which does not, with the measurement and its
  date/host/build.
* `extension/service_worker.js` (the old `:1184`) — "`--reset` … does not need to
  undo anything" → what the undo is, its ordering, and its failure modes.
* the `emulate` response `note` — "Between ops the tab is NOT emulated" → touch/
  UA/media/timezone are gone between ops, the viewport size is not, and a crashed
  agent **can** leave a tab at the emulated size.
* the `--reset` response `note` — "Nothing had to be undone" → what was cleared,
  plus the fact that document-creation properties (`ontouchstart`) are **not**
  undone by a clear and need a re-`nav`. There is now a second `note` for the
  no-stored-state case that says reset sent nothing and cannot repair an evicted
  tab.
* `reference/emulation.md` — the safety-property section is replaced by the
  measurement table above, a "what `--reset` does now" section, and the
  evicted-worker limitation.
* **Two more the issue did not list, same false claim:** `NOT_EMULATED_READ_NOTE`
  (the user-facing string on every non-CDP read) told the caller that
  layout-derived values "are the DESKTOP ones here" — false for
  `innerWidth`/`innerHeight`. It now says a non-CDP read is a **mixture**, which
  is worse than a uniformly-desktop read, not better. And the `browser` CLI's own
  `emulate` help repeated the "a crashed agent can never leave your real browser
  distorted" claim.

**On the `emulated` / `notEmulatedRead` annotation** (raised in the issue, so
saying it explicitly rather than leaving it implied): it keys off the
`emulationState` Map, so pre-fix a reset-but-still-overridden tab carried **no**
emulation fields at all while its viewport was still 393 — the envelope actively
asserted an un-emulated read on an emulated tab. This fix removes that
inconsistency by construction: after a successful `--reset` the tab really is
un-emulated, so "no emulation fields" is true again. It is **not** fixed for the
evicted-worker case, where the Map is empty for a reason the extension cannot
observe — the annotation cannot report state that no longer exists. Nothing was
changed in `annotateEmulatedRead` itself.

---

## STEP 4 — tests

6 new tests in `tests/emulation.test.mjs`, 3 pure + 3 glue. 🔴 They assert the
**reset step list and the resulting CDP calls**, never `ok:true` — the broken
behaviour returned `ok:true`, `reset:true` **and** a populated `wasEmulating`, so
an envelope-shape test passes against the bug.

| # | test | asserts |
|---|---|---|
| 1 | `RESET STEPS: a full emulation undoes EVERY class it applied, in order` | the exact ordered 6 methods **and their params** (a clear with the wrong payload is a no-op that looks like an undo) |
| 2 | `RESET STEPS: derived from what was APPLIED, not a fixed list` | metrics-only → 1 step; media-only → 1 step; an explicit `--no-touch` still gets its undo; `null`/reset → `[]` |
| 3 | `RESET STEPS: applyEmulationResetSteps is BEST-EFFORT and reports the damage` | every step attempted despite a mid-list rejection; the `cleared`/`failed` split |
| 4 | `--reset SENDS the CDP clears — the tab is actually un-emulated` | **the regression test.** `methods()` after reset == the 6 clears, inside one `attach…detach` |
| 5 | `--reset REPORTS a clear that the browser refused, and still drops the state` | `clearFailed`, metrics clear still sent, Map entry gone |
| 6 | `--reset does not THROW when CDP refuses to attach — state is dropped anyway` | `clearError`, no throw, Map entry gone |

### RED → GREEN matrix

Ran by checking out `origin/main`'s **implementation** files (never the test file),
re-running, then restoring:

| tree | node suite | which tests, and with what error |
|---|---|---|
| `origin/main` (all files) | **468 pass / 0 fail** | the stated baseline |
| HEAD, `extension/service_worker.js` reverted to `origin/main` | **471 pass / 3 fail** | tests **4, 5, 6**. #4 fails on its own assertion message with `actual []` against the 6 expected clears — pre-fix, reset sends **nothing** |
| HEAD, **both** `protocol.js` + `service_worker.js` reverted | **0 pass / 1 fail** | `SyntaxError: … does not provide an export named 'EMULATION_RESET_MAX_STEPS'` — a whole-file red. Tests 1–3's red is therefore attributable to the missing implementation but is **not** a per-assertion red; stated rather than glossed. |
| HEAD (all files) | **474 pass / 0 fail** | |

### Counts (parsed from TAP / pytest output, not from exit codes)

```
node --test --test-reporter=tap scripts/browser-bridge/tests/*.test.mjs
  # tests 474   # pass 474   # fail 0   # skipped 0        (main: 468)

nix-shell -p 'python312.withPackages(p:[p.pytest p.psycopg2 p.requests p.minio])' \
  --run "python -m pytest scripts/browser-bridge/tests -q"
  436 passed in 161.41s                                    (main: 436)
```

---

## 🔴 NOT VERIFIED

**This fix is not live-verified.** The measurement in STEP 1 is of the **bug**,
against the currently-deployed 0.7.1 extension. The **fix** has been exercised only
against a mocked `chrome.debugger` — that proves the bridge *sends*
`Emulation.clearDeviceMetricsOverride` in the right order with the right params; it
does **not** prove Chromium restores the viewport when it receives it.

Why it was not verified: a deployed extension change needs `home-manager switch`
**and a FULL Brave restart**, and the operator's `restore_on_startup` is unset, so
a restart loses their tabs. That was deliberately not done.

Specifically unproven:

* that `Emulation.clearDeviceMetricsOverride` actually restores the render widget
  (the widget-residue reading in STEP 1 is an *inference* from `devicePixelRatio`
  reverting, not a measurement of the clear);
* that `setUserAgentOverride{userAgent:""}` is accepted as "no override" on this
  Brave build (it is what DevTools sends; it was not exercised live);
* the geolocation clear at all — never measured in either direction.

### Exact post-restart check, for the operator

After `home-manager switch` on the target host and a **full Brave quit + reopen**:

```bash
export CLAUDE_CODE_SESSION_ID=verify319
B=~/workspace/devrc/scripts/browser-bridge/browser

$B --instance personal ping     # expect extensionVersion 0.7.2 — if 0.7.1, Brave did
                                # not reload; the ↻ button is unreliable, quit fully

T=$($B --instance personal open https://example.com | grep -o '"tabId": [0-9]*' | grep -o '[0-9]*')
$B --instance personal --tab $T js --wake 'innerWidth'   # A: desktop, e.g. 1124
$B --instance personal --tab $T emulate iphone-15
$B --instance personal --tab $T nav https://example.com
$B --instance personal --tab $T js --wake 'innerWidth'   # B: 393
$B --instance personal --tab $T emulate --reset          # reply now carries cleared:[…]
$B --instance personal --tab $T js --wake 'innerWidth'   # <- THE DISCRIMINATING READ
$B --instance personal --tab $T close
```

**The discriminating check is the last read.** It must return the **desktop** width
from step A (e.g. `1124`). `393` means the fix did not take — either Brave is still
running 0.7.1 (check `ping` first) or `clearDeviceMetricsOverride` does not do what
this PR assumes.

Two secondary tells, both cheap: the `--reset` reply should carry a
`cleared: ["Emulation.clearDeviceMetricsOverride", …]` array (absent on 0.7.1), and
`browser tabs` should show `emulatedTabs: []` with no `emulation` block on that tab.

If the last read still says `393` on a confirmed-0.7.2 build, **that is a real
finding, not a deploy problem** — reopen #319 with it; the reset would then need a
different remedy (e.g. an explicit `setDeviceMetricsOverride` back to the real
window metrics before clearing).

Nothing was deployed, no `home-manager switch` was run, Brave was never restarted,
and PR #318's branch was not touched.

### DIFF
--- a/scripts/browser-bridge/browser
+++ b/scripts/browser-bridge/browser
@@ -98,11 +98,11 @@
 #                               Raw override: --width/--height/--dsf/--mobile/
 #                               --touch/--ua for anything not in the table.
 #                               STICKY per tab: re-applied inside every later op's
-#                               CDP session, because CDP overrides die at detach.
-#                               ✅ That is also the SAFETY property — between ops
-#                               the tab is NOT emulated, so a crashed agent can
-#                               never leave your real browser distorted.
-#                               `--reset` stops re-applying (nothing to undo).
+#                               CDP session, because most CDP overrides die at
+#                               detach. 🔴 The VIEWPORT SIZE does NOT (measured) —
+#                               it survives the detach, so a crashed agent CAN
+#                               leave a tab phone-sized. `--reset` sends the CDP
+#                               clears to undo it; closing the tab also works.
 #                               ⚠ OWNED TABS ONLY (`browser open` first): emulating
 #                               a tab you are browsing in is refused with
 #                               `not_owned_tab`. Screenshots on an emulated tab are
--- a/scripts/browser-bridge/extension/README.md
+++ b/scripts/browser-bridge/extension/README.md
@@ -273,6 +273,24 @@ If `ping` still reports the old version after ↻, **fully quit and reopen Brave
 > three full Brave restarts in a single session. (0.3.0 → 0.3.1 was exactly
 > this: adding `id` to `ping`'s reply is a discriminator, so it got a bump.)
 
+> **0.7.2 — `emulate --reset` actually CLEARS the overrides (#319).** No new wire
+> op, so `ping`'s `ops` list is byte-identical to 0.7.1's. The discriminator is
+> the **behaviour**, and it is exact — a post-reset viewport read:
+>
+> ```bash
+> browser --instance <label> open https://example.com          # note the tabId
+> browser --instance <label> --tab <id> js --wake 'innerWidth'  # e.g. 1124 (desktop)
+> browser --instance <label> --tab <id> emulate iphone-15
+> browser --instance <label> --tab <id> emulate --reset
+> browser --instance <label> --tab <id> js --wake 'innerWidth'
+>   # 0.7.1 → 393   (the reset was a no-op; the tab is stuck as an iPhone)
+>   # 0.7.2 → 1124  (Emulation.clearDeviceMetricsOverride was sent)
+> browser --instance <label> --tab <id> close
+> ```
+>
+> The `--reset` reply also grows a `cleared: [...]` array on 0.7.2, which is a
+> second, cheaper tell.
+
 > **0.7.1 — `text --annotated` inside `--frame`.** No new wire op, so `ping`'s
 > `ops` list is byte-identical to 0.7.0's and cannot discriminate this build.
 > The discriminator is the **capability itself**, and it is exact — it exercises
--- a/scripts/browser-bridge/extension/manifest.json
+++ b/scripts/browser-bridge/extension/manifest.json
@@ -1,7 +1,7 @@
 {
   "manifest_version": 3,
   "name": "Browser Bridge (command channel)",
-  "version": "0.7.1",
+  "version": "0.7.2",
   "description": "Lets a local Claude Code skill drive the active Brave tab (getHtml/eval/tabs/nav/screenshot/frames/click/type/key/activate/upload/emulate/ping) via a loopback, token-authenticated rendezvous server. Local-only, authorized personal automation.",
   "permissions": ["scripting", "tabs", "activeTab", "storage", "alarms", "debugger", "webNavigation"],
   "host_permissions": ["http://127.0.0.1:8788/*", "<all_urls>"],
--- a/scripts/browser-bridge/extension/protocol.js
+++ b/scripts/browser-bridge/extension/protocol.js
@@ -561,19 +561,32 @@ export async function applyWakeSteps(send, steps = WAKE_CDP_STEPS) {
 // everything in THIS file is the pure, browser-free half: the preset table, the
 // validation, and the ordered CDP step list.
 //
-// ✅ THE SAFETY PROPERTY THIS BUYS — state it explicitly, it is the REASON this
-// design was chosen over holding one long-lived debugger session open:
+// 🔴 THE "DIE AT DETACH" PREMISE IS ONLY MOSTLY TRUE — measured, do not restate
+// the old claim. This block used to assert that because every override dies at
+// detach, the tab is never emulated between ops and a crashed agent CANNOT leave
+// the operator's browser distorted. Issue #319 falsified it, and the follow-up
+// measurement (2026-08-03, live Brave, extension 0.7.1, laptop, `iphone-15`,
+// read via `js --wake`, against a never-emulated control tab in the same window)
+// says exactly which half is which:
 //
-//   Because the overrides die at detach, the tab is NOT emulated between ops. A
-//   crashed agent, a killed Claude session, or an evicted MV3 service worker
-//   therefore CANNOT leave the operator's real browser distorted. The worst case
-//   is a forgotten entry in an in-memory Map that dies with the worker — not a
-//   Brave tab stuck at 393×852 pretending to be an iPhone with nothing left
-//   running that knows how to undo it.
+//   * setDeviceMetricsOverride's VIEWPORT SIZE does NOT die at detach. After
+//     `--reset` — and after a further navigation — innerWidth×innerHeight was
+//     still 393×852 while the control tab read 1124×1400. Interestingly
+//     devicePixelRatio, set by the same call, DID revert to 1, which points at
+//     the residue being the resized render widget rather than a live override.
+//   * touch, UA/UA-CH, emulated media and the timezone override DO die at
+//     detach: after `--reset` they all read as the real desktop values.
 //
-// A held session would have been simpler to write and strictly worse to own: the
-// failure mode would be a permanently mangled tab plus a permanent debug banner,
-// recoverable only by the operator finding and closing the tab.
+// So the honest statement of what the design buys: between ops the tab keeps the
+// emulated VIEWPORT SIZE and nothing else. A crashed agent, a killed Claude
+// session or an evicted MV3 service worker CAN leave a tab sized as a phone —
+// closing the tab, or `emulate --reset` (which since #319 sends
+// Emulation.clearDeviceMetricsOverride and the matching clears — see
+// emulationResetCdpSteps below), is what undoes it.
+//
+// A held session would still have been worse to own: its failure mode adds a
+// permanent debug banner and leaves the UA, touch and media overrides live too,
+// none of which survive today.
 //
 // BLAST RADIUS: enforced SERVER-side (see server.py OWNED_TAB_ONLY_OPS). Only a
 // tab the calling session opened via `open` may be emulated; anything else is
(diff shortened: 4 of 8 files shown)


ZacxDev:
Superseded and closing. The fix here was live-verified and **does not work**.

Deployed 0.7.2 from this branch and measured it on real Brave with a fresh-tab control:

| step | `innerWidth` |
|---|---|
| fresh tab, never emulated (control) | 1124 |
| after `emulate iphone-15` | 393 |
| after `--reset` (this branch's clears) | **393** |
| after `--reset` + re-nav | **393** |

The clears genuinely fired — the envelope reported `cleared: [Emulation.clearDeviceMetricsOverride, Emulation.setTouchEmulationEnabled, Emulation.setUserAgentOverride]` — and the viewport still did not come back. So the premise this branch was built on (and that I wrote into #319) is wrong: the bug is not a missing CDP call.

Note the earlier readings that looked like "old code is running" were mine measuring a service worker that had not been killed yet. Once it was, the new code was demonstrably live and still ineffective.

Shipped instead in #321: the honest prose (the viewport is NOT restored; mechanism unexplained) plus `emulate --reset --recreate`, a CLI-side workaround that replaces the tab — tab replacement being the only remedy that works. #322 bumps the extension to 0.7.3 so the corrected note is detectable.

The non-viewport clears in this branch were not carried over, since those overrides were measured to revert on their own. If the mechanism is ever explained and a real clear is found, this branch is the starting point.

PR #326, opened 2026-08-04
fix(resume-state): resolve SESSION-HANDOFF.md, and stop DRIFT reporting a false green

ZacxDev:
## The bug

`resume-state.sh`'s `resolve()` globbed only `claudedocs/handoff-*.md`. The
`civitai-manager` repo names its handoff `claudedocs/SESSION-HANDOFF.md`.
Measured there against pristine `origin/main`:

```
GIT/PR
  civitai-manager  branch main  ↑0↓0  clean  last 8m: opencode
  handoff: (none found — git-only)
...
DRIFT
  (none detected — live state matches the handoff's claims)
```

**The `DRIFT` line is the real damage: it is vacuous.** It reconciled against
nothing and printed a reassuring result. A caller reads "no drift" as a fact
about the handoff when no handoff was ever loaded — a false green in the one
place `/resume` exists to be trustworthy.

Same repo, after this PR:

```
  handoff: SESSION-HANDOFF.md
  branch feat/breadcrumbs: exists (origin/feat/breadcrumbs)
  branch feat/comfy-model-cache: GONE (deleted or never local)
  branch fix/copy-reduction: exists (origin/fix/copy-reduction)
DRIFT
  - branch feat/comfy-model-cache referenced by handoff no longer exists (merged & pruned?)
```

> ⚠ **Corrected.** This block originally also showed a
> `branch docs/configuration.md ... no longer exists` line and cited it as evidence
> that the fix worked. **That statement was false** — `docs/configuration.md` is a real
> 15,276-byte **file** in that repo, not a branch. Fixing that fabrication is 🔴-2
> below. See
> [the correction comment](https://github.com/innovation-upstream/devrc/pull/326#issuecomment-5180418269)
> for the durable record.

## What changed

**1. A last-resort fallback glob in `resolve()`.**

```sh
[ -z "$HANDOFF" ] && HANDOFF=$(ls -t "$REPO"/claudedocs/*HANDOFF*.md 2>/dev/null | head -1)
```

**Exact reach:** basenames under `claudedocs/` containing the literal,
**uppercase** substring `HANDOFF` and ending `.md`. That is
`SESSION-HANDOFF.md`, `HANDOFF.md`, `HANDOFF-2026-08-04.md`. It cannot reach
the design/audit docs that share these directories — `*-DESIGN.md`,
`SECURITY-AUDIT-*.md`, `LAUNCH-*.md`, `*-RESEARCH.md` — none of which contains
`HANDOFF`.

Behaviour preservation, per the brief's invariants:

- It is tried **last**, so a repo with both shapes resolves exactly as today —
  `handoff-*.md` wins regardless of age. (Being tried last is the *only* thing
  protecting the lowercase family; an earlier comment also credited bash's
  case-sensitive globbing, which is inert here and has been removed.)
- The `ls -t | head -1` newest-first selection shape is unchanged in both families.
- The explicit-path form (`resume-state.sh path/to/doc.md`) is untouched.
- Verified against the **real** `devrc` repo (30+ `claudedocs/handoff-*.md`):
  `resolve()` returns a byte-identical answer at base and at HEAD for the
  no-arg case and for two topic slugs. Not just fixtures.

**2. The topic slug gets no fallback of its own — decided, not overlooked.**
An unmatched slug already falls through this chain, so `resume-state.sh session`
in a repo whose only handoff is `SESSION-HANDOFF.md` still finds it, for zero
extra code. The slug is only ever interpolated into the `handoff-*` glob, so it
can never reach a decoy. Both behaviours are tested.

**3. `DRIFT` no longer lies when nothing was reconciled.** Two cases:

- no handoff resolved → `(no handoff loaded — nothing to reconcile; this is NOT a clean bill of health)`
- a handoff loaded but a source never answered → `(nothing detected, but a source did not answer — NOT a clean bill of health)`, with the ratio

**4. The `/resume` skill carried the same blind spot in prose** — step 1 told
the model to glob `handoff-*.md` and, failing that, to give up. The predicate
lived in two places and only one was fixed (`RULES.md` → "One rule, one
place"). Step 1 now names the same two-tier order; step 3 says to read the
`handoff:` line **before** `DRIFT`.

## The three 🔴 findings from the adversarial audit

**🔴-1 — the hermeticity tripwire was wired to nothing.** `run_resume()` built
`env` (stub PATH, `STUB_LOG`, git-config isolation) then called
`subprocess.run` **without `env=`**, so the subject inherited `os.environ`.
`test_no_network_tool_is_ever_invoked` was asserting a *structural* zero.
Verified both directions with a `gh` call spliced into `resolve()`: **green
before the fix, red after**. The positive control missed it because it execs
the stub *directly* with `env=env` — a path the subject never took.

**🔴-2 — DRIFT fabricated branches from file paths.** `zach|feat|fix|docs|chore`
are branch prefixes *and* ordinary directory names, and `\b` matches after a
slash, so a handoff that merely **quoted a path** minted a phantom branch
reported as "no longer exists (merged & pruned?)" — 2 of the 6 new DRIFT lines
across the two repos this fallback newly reaches. Three filters: the leading
boundary now excludes `/`; a trailing alphabetic file extension disqualifies a
token; and `git cat-file -e HEAD:<tok>` drops anything naming a real tracked
path. `my-fix/x` still matches and `notafix/x` still does not — `/` is the only
boundary change.

**🔴-3 — the false green survived one layer up.** With `gh` present and every
`gh pr view` failing, `|| continue` swallowed it with no diagnostic and DRIFT
still claimed a clean reconciliation. The block now counts what answered and
reports the ratio. No cause is attributed — `gh` exits non-zero for a genuine
404 and an auth failure alike. A handoff referencing no PRs is deliberately
*not* downgraded, or the warning fires on every non-GitHub repo and becomes
noise.

Plus a bug the new tests caught in my own fix: a trailing `[ … ] && printf` as
the last statement of a branch made the **whole script exit 1** whenever it
found drift with nothing unreconciled.

## Tests — 55 hermetic cases, ~6 s

`scripts/tests/test_resume_state_handoff_resolution.py` drives the script end
to end against throwaway git repos and asserts the digest's `handoff:` line
names the **exact** file — never merely that the script exited 0, which is true
of the broken script too. Every absence-assertion has a positive control beside
it, because an absence is also what deleting the feature produces.

**Written as pytest, not bash.** `scripts/run-tests.sh` — the flake gate — takes
`scripts/tests` as a pytest target, so it collects `test_*.py` only. The two
`.sh` suites there are run by hand and by nothing else. The
stub-binaries-on-`$PATH` technique is still `test_release_wrapper.sh`'s.

## Red/green matrix

| | base `0294b82` | HEAD |
|---|---|---|
| new suite | **8 failed, 18 passed** (1.23 s) | **55 passed** (~6 s) |

Every base failure is a real `AssertionError`, not a collection or syntax
error, and the 18 that passed at base prove the harness was exercising the
script rather than erroring out.

**Mutation battery — 20 mutations, all killed.** Each verified to have landed
(`git diff --stat`), to leave `bash -n` passing (so it broke *behaviour*, not
parsing), and to fail with a real `AssertionError` in the *expected* test.
Covers: the fallback glob (delete / reorder / merge / broaden ×2 / drop `.md` /
drop `head -1` / unscope from `claudedocs/`), slug anchoring, all three
branch-fabrication filters plus an over-broad "suppress everything" fix, and
six mutations across the unreconciled-source logic.

Four came back wrong on the first run — three were bugs in my own battery (a
literal `\\b`, a misfiled expectation, and one that left unbalanced quotes and
so was a *parse* error exercising no assertion) and **one was a real gap**:
`test_a_quoted_file_path_is_not_reported_as_a_branch` was over-determined,
because its fixture tracks the file so the `cat-file` probe rejected the token
and the extension filter was never load-bearing. Fixed by *adding* the
discriminating case rather than editing the measured fixture away.

### Delta-audit round (F3/F2/F4) — 24 mutations, all killed

The delta audit found four residuals. Three are fixed here; the fourth is tracked
separately (below).

**🟡 F3 was a REGRESSION this PR introduced** and is the important one. Excluding
`/` from the branch-token boundary killed the path fabrication *and* every
reference legitimately carrying a slash-bearing prefix — `origin/fix/x`,
`upstream/feat/x`, `refs/heads/fix/x`, and GitHub `/tree/` and `/compare/` URLs.
One live casualty across 211 real handoff docs:
`origin/zach/engaged-models-client-store` in datapacket-talos, the only form that
branch appears in, and it is genuinely gone — so the pre-PR DRIFT line was correct
and the regressed code was silently mute. **Omission is the worse polarity in a
go/no-go tool**, because silence reads as "no drift" instead of announcing itself.

Fixed by stripping the ref-ish prefix before matching. Verified in **both**
directions on the real documents:

| doc | pre-PR base | this PR, round 2 (regressed) | HEAD |
|---|---|---|---|
| datapacket-talos handoff | `zach/engaged-models-client-store` | *(nothing)* | `zach/engaged-models-client-store` |
| naida-ai `HANDOFF.md` | 2 real + **`zach/workspace/scratch/naida-ai`** | 2 real | 2 real |
| civitai-manager `SESSION-HANDOFF.md` | 3 real + **`docs/configuration.md`** | 3 real | 3 real |

HEAD is base *minus* both fabrications *plus* the restored ref — the two fixes pull
in opposite directions and are pinned together in one test.

**🟡 F2** — the partial-answer branch was unreachable (fixtures could only make `gh`
answer for all or none), so mutating it survived all 46 tests. The `gh` stub gained a
selective mode and the new test asserts its preconditions before its verdict.

**🟡 F4** — `claude/commands/resume.md` now names both conditions that make an empty
`DRIFT` meaningless, including the `!` gap lines, and says which single wording is an
actual all-clear.

Two 🟢s taken: the tracked-path probe now runs *after* the branch-existence checks (a
token that is both a live branch and a tracked path was being dropped — 0 such
collisions across 2893 real branches, so latent rather than live), and `alerts_block`'s
trailing `&&` is now a full `if`, the same class as the exit-1 bug fixed in `main`.

## Follow-up tracked separately

**#330 — three more "reconciled nothing" paths still print the clean line.**
`git_pr_block`'s not-a-git-repo early return, and an unreachable cluster in
`workload_block` / `alerts_block`, all reach `(none detected — live state matches the
handoff's claims)` having checked nothing. `UNRECONCILED` is the right mechanism and
is simply not fed from those sites. Squarely within this PR's thesis and the largest
remaining gap — but **pre-existing, not a regression from this PR**, and this PR has
already grown twice. Two of the three are reproduced live in the issue; the third is
source-read only, since reaching it needs a live cluster.

## Wider-suite check

`pytest scripts/tests` under an ad-hoc `nix-shell` (not the full toolchain, so
there is pre-existing environment noise): **no failure at HEAD that is not also
at base** — 26 identical pre-existing failures, 1088 passed at HEAD.
`test_opencode_engine.py` is non-deterministic in this environment (the *same
base commit*, run alone, gave 15, 2 and 4 failures on three consecutive runs),
established by measurement rather than assumption.

`bash -n` passes. `shellcheck` gains exactly one new finding, a third instance
of the pre-existing, accepted `SC2012` (`ls -t | head -1`), which invariant 2
required me to keep. No new finding class. The pre-existing
`scripts/tests/test_resume_state.sh` still reports `ALL PASS`.

## Out of scope — a second, unrelated bug worth its own PR

A **separate defect in the branch loop of `git_pr_block`**, which this PR does
not fix. When a handoff-referenced branch exists only on the remote, the code
sets `tip="origin/$b"` but then runs the merged-check against the **bare** `$b`,
which does not resolve; `git branch --merged` lists locals only. So a
remote-only branch is **always** reported `exists`, never `merged into HEAD`.
Measured: `feat/breadcrumbs` and `fix/copy-reduction` above are reported merely
"exists", while `git merge-base --is-ancestor origin/<b> origin/main` returns
rc=0 for both. Different function, different blast radius, needs its own test
coverage.

## Also noted, not fixed

`scripts/tests/test_resume_state.sh` and `test_release_wrapper.sh` both derive
their own directory with `HERE=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)`.
With `CDPATH` exported (it is, on this host), bash's `cd` **echoes the resolved
path**, so `$HERE` gains a newline and the `source` fails — every assertion then
reports `command not found` and the suite is meaningless. It only bites on a
**relative** invocation, which is exactly the documented `run:` line. Confirmed
pre-existing: pristine `origin/main` fails identically. The fix is `CDPATH= cd …`.

https://claude.ai/code/session_01SeGJdEJjtvyv5PY3HW9pEF


ZacxDev:
## ⚠ Correction — the original PR body contained a fabricated fact, presented as success evidence

A blind adversarial audit returned three 🔴 findings. All three are fixed in `1d14679`, `6fa2f93` and `1b5f2ad`. One of them requires a **public** correction rather than a silent body edit, because it is possible someone has already read and believed it.

**The "after" digest quoted in the original PR body included this line:**

```
  branch docs/configuration.md referenced by handoff no longer exists (merged & pruned?)
```

**That statement is false.** `docs/configuration.md` is a real 15,276-byte **file** in civitai-manager, backtick-quoted in that repo's handoff prose. It is not a branch and never was. `extract_branches` matched it because `zach|feat|fix|docs|chore` are branch prefixes *and* ordinary directory names, and `\b` matches after a slash.

So the PR as originally submitted removed one false green and **introduced two fabricated facts** — the second being `zach/workspace/scratch/naida-ai`, minted in naida-ai from the prose "local checkout `/home/zach/workspace/scratch/naida-ai`". That is 2 of the 6 new DRIFT lines those two repos produce. I cited one of them as evidence the fix worked, which is exactly the failure mode this PR is about.

The corrected digest for civitai-manager, measured after `1d14679`:

```
  handoff: SESSION-HANDOFF.md
  branch feat/breadcrumbs: exists (origin/feat/breadcrumbs)
  branch feat/comfy-model-cache: GONE (deleted or never local)
  branch fix/copy-reduction: exists (origin/fix/copy-reduction)
DRIFT
  - branch feat/comfy-model-cache referenced by handoff no longer exists (merged & pruned?)
```

One genuine finding, no fabrications. naida-ai likewise keeps both of its real branches and loses the phantom.

The PR body has been updated as well, but this comment is the durable record.

---

### The other two 🔴 findings

**🔴-1 — the hermeticity tripwire was wired to nothing.** `run_resume()` built `env` (stub PATH, `STUB_LOG`, git-config isolation) and then called `subprocess.run` **without `env=`**, so the subject inherited `os.environ`. `test_no_network_tool_is_ever_invoked` was asserting a *structural* zero. Verified in both directions with a `gh` call spliced into `resolve()`: **green before the fix, red after** (log shows one probe per run). The positive control I had added did not catch it because it execs the stub *directly* with `env=env` — a path the subject never took — so it validated the stub rather than the tripwire.

**🔴-3 — the false green survived one layer up.** With `gh` present and every `gh pr view` failing, `|| continue` swallowed it with no diagnostic and DRIFT still said "(none detected — live state matches the handoff's claims)". The block now counts what actually answered, reports the ratio, and DRIFT degrades to "(nothing detected, but a source did not answer — NOT a clean bill of health)". No cause is attributed — `gh` exits non-zero for a genuine 404 and for an auth failure alike, so a classifier would be guessing. A handoff referencing no PRs is deliberately *not* downgraded, or the warning would fire on every non-GitHub repo and become noise.

### A bug the new tests caught in my own fix

The first version of the DRIFT change ended a branch with `[ … ] && printf`, which returns 1 when the test is false — making the **whole script exit 1** on every run that found drift with nothing unreconciled. Now a full `if` block; `run_resume` asserts rc 0 on all 46 runs.

### Updated matrix — 20 mutations, all killed

Each verified to have **landed** (`git diff --stat`), to leave `bash -n` **passing** (so it broke behaviour, not parsing), and to fail with a real `AssertionError` in the **expected** test. Suite: **46 passed, ~2 s**; no failure at HEAD that is not also at base `0294b82` (26 pre-existing environment failures, identical sets).

Four mutations initially came back wrong, and the split is worth recording:

| | cause |
|---|---|
| `\b` boundary restore | **my battery's bug** — the replacement wrote a literal `\\b`; killed once corrected |
| `if [ -n "$prs" ]` → `true` | **my expectation's bug** — pointed at the wrong test; the guard *is* covered |
| never-record-unanswered-gh | **my battery's bug** — left unbalanced quotes, so it was a *parse* error exercising no assertion |
| drop the file-extension filter | **a real gap** — see below |

The real gap: `test_a_quoted_file_path_is_not_reported_as_a_branch` was **over-determined**. Its fixture tracks `docs/configuration.md`, so the `git cat-file` probe rejects the token and the extension filter was never load-bearing — deleting it left all 44 tests green. Fixed by *adding* the discriminating case (a quoted path that does **not** exist in this repo, which handoffs routinely cite) rather than editing the real measured fixture away. Two mutations that had no test at all also gained one, including the `gh`-absent arm of the unreconciled logic.

Three 🟡 guards the audit flagged as existing-but-unpinned are now pinned — slug anchoring (its green was a case-mismatch coincidence; a lowercase decoy modelled on devrc's real `browser-bridge-usage-audit-2026-08-02.md` removes it), `claudedocs/` scoping, and the gap-alongside-findings path. The case-sensitivity claim was **removed from the comment rather than pinned**: it is inert, since the fallback is only reached when the lowercase globs found nothing, so there is nothing left to poach. Being tried *last* is the only thing doing that work.

🟢 Docstring timing corrected to the measured ~3.2 s cold / ~2.0–2.9 s warm; it claimed "well under a second", which was never measured.

Still `MERGEABLE` / `CLEAN` against current `main`, which has moved to `481816f` — no file overlap with this branch.

ZacxDev:
## Round-3 audit fixes — `|compare` deleted, ref table relabelled

Commits `7a48513` and `d2e02d4`. **26 mutations, all killed. 58 tests, ~2 s.** No failure at HEAD that is not also at base `0294b82` (26 identical pre-existing environment failures; 1100 passed at HEAD).

### 🟡 F1 — `|compare` deleted

All three mechanical claims reproduced, so the token is gone:

- **Unpinned** — deleting it left all 55 tests green.
- **Redundant** — a compare URL's `main...feat/x` already matches without it; the `.` in `main...` satisfies the grep boundary.
- **Worse on a slashed left side** — stripping turned `…/compare/zach/a...zach/b` into the junk token `zach/a...zach/b`; not stripping yields `zach/b`, the head of the compare, which is the ref a reader means.

### ⚠ One part of the finding did not reproduce — and it was the part that framed this as urgent

The audit reported this as a fabrication **introduced by this PR**:

```
base: branch zach/b: GONE
HEAD: branch zach/a...zach/b: GONE
```

Measured end-to-end through the real script, **base and `dfcb800` are identical** — both emit `branch zach/a...zach/b: GONE`. The junk token is **pre-existing**, not a regression from this branch. Deleting `|compare` is therefore an *improvement over base*, not a regression fix. It still lands, on the unpinned + redundant + strictly-better grounds above.

Worth recording *why* the first A/B looked like it confirmed the audit: my initial fixture named the handoff `SESSION-HANDOFF.md`, which **base cannot resolve at all** — so base returned before the branch loop and printed nothing. That silence reads as "no fabrication" but means "no handoff loaded". An empty result cannot distinguish two mechanisms; re-running with a `handoff-*.md` fixture is what made the comparison valid.

### 🟢 F2 — the ref table now names *which* regression each row pins

Measured by replaying every spelling through all three prior versions, rather than asserted:

| row | base | r2 (round-2 regression) | r3 (`dfcb800`) | pins |
|---|---|---|---|---|
| `origin/fix/wanted` | same | **RED** | same | r2 |
| `upstream/feat/wanted` | same | **RED** | same | r2 |
| `refs/heads/fix/wanted` | same | **RED** | same | r2 |
| `refs/remotes/origin/zach/wanted` | same | **RED** | same | r2 |
| `…/tree/feat/wanted` | same | **RED** | same | r2 |
| `…/compare/main...feat/wanted` | same | same | same | **nothing** |
| `…/compare/zach/a...zach/b` | **RED** | same | **RED** | base + r3 |

Note the first column: **against the pre-PR base, all six original rows are invariants** — base's `\b` already handled every one. What they actually pin is round 2's own regression. Saying "regression coverage" without naming the reference point is how a table reads as more than it is. A meta-guard now pins the shape so the split cannot be quietly relabelled away.

### 🟢 F3 — and a vacuous guard the battery caught in my own fix

The `origin|upstream` leading bound is now tested. **The first version of that test was vacuous** and the battery caught it: mutating the bound left all 58 green, because the fixture I chose (`/home/zach/repos/origin/fix/x`) cannot discriminate — strip `origin/` there and `fix` is still preceded by `/`, which the grep boundary rejects anyway. I had picked the fixture from the rationale instead of from measurement.

Measured five shapes with the bound loosened to find one that does:

```
/home/zach/repos/origin/fix/x  -> []        (identical — no signal)
/var/log/my-origin/fix/x       -> [fix/x]   <- FABRICATED out of a path
.origin/fix/x                  -> [fix/x]
my-origin/fix/x                -> [fix/x]
origin/fix/x                   -> [fix/x]   (correct in both)
```

The bound earns its place on a remote-like segment glued to a **word** inside a path, not on a bare `/origin/`. Accepted cost, now stated at the site: a genuinely remote-qualified `my-origin/fix/x` yields nothing — an omission, and fabrication is the worse direction.

The `+` in `(refs/(heads|remotes)/)+` was also dropped as untested defensive complexity for a doubled prefix nothing produces.

### 🟢 F4 — `alerts_block` coverage gap recorded at the site

A comment there states that **no test reaches the block** (the hermetic fixtures have no `prod-kubeconfig`, so it always returns at the `WL_NS` guard), so the suite's green is not read as covering the hygiene fix. Verified by reading, not by a guard, and it says so.

### Battery note

Two rows came back `!!` during this round and both were **my battery's** bugs, not surviving mutants: a stale mutation target after the sed changed (caught by the battery's own `assert`, not by a silent no-match), and an expectation still naming a test I had renamed. The mutation had in fact gone red with 2 assertion failures both times.

Real-document behaviour is unchanged by this round: datapacket-talos still yields `zach/engaged-models-client-store`, naida-ai its two real branches, civitai-manager its three — no fabrications in any.

Issue #324, opened 2026-08-04
browser-bridge: extension_version/extension_stale cannot detect stale extension code — two profiles, same id + version, different code


ZacxDev:
## Summary

`extension_version` / `extension_stale` (in `ping`, `whoami`, `health`) **cannot detect that a Brave profile is running stale extension code.** Two profiles on one host reported an identical `extension_id`, an identical `extension_version`, and `extension_stale: false` — while demonstrably executing **different code**, one of which exists on no disk.

This is the question `ping` was added to answer ("is the build I just deployed the one Brave loaded?"), so the failure lands exactly where it does the most damage: immediately after a deploy.

## Measured (2026-08-04, laptop, bridge on 127.0.0.1:8788)

Discriminator: whether `emulate --reset` emits a `cleared: [...]` field. `main`'s reset branch is a bare `emulationState.delete(tabId)` (`extension/service_worker.js`) and **structurally cannot emit it**; only the unmerged branch `fix/emulate-reset-clear-overrides-319` (0.7.2) does.

| profile | `extension_id` | reported version | `extension_stale` | actual code |
|---|---|---|---|---|
| `personal` | `bgbkamdlkdleahpgdgmjipjbgmepgenk` | 0.7.3 | `false` | 🔴 **0.7.2** (emits `cleared`) |
| `work` | `bgbkamdlkdleahpgdgmjipjbgmepgenk` | 0.7.3 | `false` | ✅ `main` |

Same id, same version, same staleness verdict; different code. **No version-shaped signal can distinguish these two rows** — the id is path-derived, so both profiles loading the same directory necessarily share it, and the version is read from that same directory.

Supporting facts, all re-measured at the time:

- On-disk `~/.local/share/browser-bridge-ext/service_worker.js` contained `main`'s note; `grep -ra "CLEARED on the tab"` over the **entire deployed tree and the entire repo tree** returned **nothing**. The code `personal` was executing exists on no disk.
- Brave was genuinely fresh — oldest browser process 392 s old, deploy 9 h earlier. A **full browser restart did not reload it**.
- The bridge service was restarted (new PID, both instances reconnected). Not the stale component either.
[...]

PR #331, opened 2026-08-04
fix(browser-bridge): a version MISMATCH decides extension_stale even with no build marker

ZacxDev:
## The live observation

Measured on the workbench, 2026-08-04, via `browser health`:

```
profile "personal - other":
  extension_version           0.7.1
  extension_build             null
  extension_version_expected  0.8.1
  extension_build_expected    73f5438f18f395d2
  extension_stale             null      <-- decidably stale, reported as unknown
```

`annotate_staleness()` returned `None` the moment either build marker was
absent — **before ever looking at the versions**. That threw away a decidable
case.

#324's asymmetry is that a version **MATCH** proves nothing: two profiles
loading one directory report the same version while running different code.
A version **MISMATCH** was never in doubt — it is positive proof that what was
loaded is not what is deployed. Only that direction is newly used.

## Verdict table as implemented

| markers | versions | verdict |
|---|---|---|
| both present, **disagree** | any | `True` (unchanged) |
| both present, agree | disagree | `True` (unchanged) |
| both present, agree | agree or either unknown | `False` (unchanged) |
| **missing on either side** | both known and **disagree** | **`True` (new)** |
| missing on either side | agree | `None` (unchanged) |
| missing on either side | either unknown/None/"" | `None` (unchanged) |

## 🔴 The invariant that survives

**`False` stays strictly marker-backed.** A version match can never produce
`False` — only `True` becomes newly reachable from versions alone. Pinned by
`test_annotate_staleness_false_stays_strictly_marker_backed`, which sweeps
7 version shapes x 5 marker-missing shapes (35 combinations) and asserts
`is not False` on every one.

## Red at base / green at HEAD

`red at 1d8249b, green at HEAD (56da81e).`

Tests were committed first (2798b06) and run against **unmodified** `server.py`:

```
FAILED test_annotate_staleness_version_mismatch_decides_true_without_markers[None-73f5438f18f395d2]
FAILED test_annotate_staleness_version_mismatch_decides_true_without_markers[a1b2c3d4e5f60718-None]
FAILED test_annotate_staleness_version_mismatch_decides_true_without_markers[None-None]
FAILED test_annotate_staleness_version_mismatch_decides_true_without_markers[-73f5438f18f395d2]
FAILED test_annotate_staleness_version_mismatch_decides_true_without_markers[a1b2c3d4e5f60718-]
5 failed, 14 passed, 280 deselected in 1.16s
```

with, for the live-observation fixture:

```
loaded_build = None, expected_build = '73f5438f18f395d2'
>       assert insts[0]["extension_stale"] is True
E       assert None is True
scripts/browser-bridge/tests/test_server.py:4356: AssertionError
```

**Honesty note:** the other three new tests are **invariant guards, not
regression coverage** — they pass at `1d8249b` and pin behaviour the fix must
not change (`marker-missing + versions agree ⇒ None`, `marker-missing + either
version unknown ⇒ None`, and the `False` invariant). They are labelled as such
in the test commit message.

## Mutation sweep (`-k annotate_staleness`, 19 selected)

All four mutants of the new expression were killed; the tree restored green.

| mutant | result |
|---|---|
| `loaded != expected` → `loaded == expected` | 10 failed |
| `True if …` → `False if …` | 6 failed |
| drop the `expected and loaded` known-version guard | 4 failed |
| revert to plain `None` (the pre-fix body) | 5 failed |
| restored | 19 passed |

The `False`-invariant guard was separately confirmed **reachable and killed by
its own assertion** under the `True → False` mutant:
`assert False is not False` at `test_server.py:4414`.

## Both tiers, before and after (counts parsed from output, not exit status)

| tier | before | after |
|---|---|---|
| pytest `scripts/browser-bridge/tests` | 467 passed, 0 failed | **481 passed, 0 failed** (+14) |
| `node --test --test-reporter=tap …/*.test.mjs` | 495 tests / 495 pass / 0 fail / 0 skipped | **495 / 495 / 0 fail / 0 skipped** (unchanged) |

## The factual correction folded in

The 🔴 docstring asserted `extension_version` is "read from the on-disk manifest
at call time". Today's observation contradicts it: two profiles sharing one
`extension_id` (hence one load path) reported `0.7.1` and `0.8.1` while the
directory on disk holds only `0.8.1`. The sentence is softened to what is
actually known — the version describes the manifest of the extension the worker
**loaded**, and still does not describe the executing code — with the
observation, its date and both profiles' values recorded as the evidence, and
explicitly flagged as **one observation, a hypothesis, not established**.
Nothing was re-architected around it. The load-bearing conclusion ("the version
cannot answer the question, use the marker") is unchanged.

## SKILL.md

`SKILL.md` documents the verdict in its `health`/`instances` row and the old
wording ("verdict from the BUILD MARKER, not the version") is now inaccurate for
`true`. Corrected **in place, 2 bytes smaller** — no eviction needed, and
`test_skill_size.py` stays green.

---

https://claude.ai/code/session_017bJFGEUx1pLkFpxRkNkhAR

### DIFF
--- a/scripts/browser-bridge/README.md
+++ b/scripts/browser-bridge/README.md
@@ -80,7 +80,7 @@ design when two profiles were connected).
   `~/.local/share/browser-bridge-ext/`, falling back to the repo copy). It answers
   "am I on the laptop or the workbench, and which profile?" in one shot (both
   hosts are hostname `nixos`).
-  **Explicit staleness verdict — computed from the BUILD MARKER (#324).** Each
+  **Explicit staleness verdict — an ALL-CLEAR comes from the BUILD MARKER (#324).** Each
   instance carries `extension_version` (loaded), `extension_version_expected`,
   `extension_build`, `extension_build_expected`, and **`extension_stale`**:
   `true` (this profile is running code that is not the deployed code → Remove +
@@ -88,10 +88,10 @@ design when two profiles were connected).
   The bridge carries `extension_build_current` alongside
   `extension_version_current`.
 
-  🔴 The verdict is **not** a version comparison, and cannot be. `extension_version`
-  is `chrome.runtime.getManifest().version` — read off the on-disk manifest at
-  call time — and `extension_id` is derived from the load PATH, so **both describe
-  the DIRECTORY, not the code that is executing**. MEASURED 2026-08-04: two Brave
+  🔴 An all-clear is **not** a version comparison, and cannot be. `extension_version`
+  is `chrome.runtime.getManifest().version` — it describes the manifest of the
+  extension the worker LOADED — and `extension_id` is derived from the load PATH,
+  so **neither describes the code that is executing**. MEASURED 2026-08-04: two Brave
   profiles loading the SAME directory reported an identical id, an identical
   `0.7.3` and `extension_stale: false`, while one ran `main` and the other an
   unmerged 0.7.2 build whose source existed on no disk. `extension_build` is a
@@ -99,10 +99,14 @@ design when two profiles were connected).
   so it is frozen into the loaded module graph and travels with the code — a stale
   worker reports the stale marker by construction.
 
-  🔴 **It FAILS CLOSED.** A marker missing on *either* side — an extension build
-  predating #324, or an unreadable/undeployed source tree — yields `null`, never
-  `false`. `false` means verified-current and nothing else; `null` is never
-  "fine". Staleness is also **per profile**: one can be current while another is
+  🔴 **It FAILS CLOSED — asymmetrically.** `false` means two markers present and
+  identical, and nothing else can produce it: a marker missing on *either* side —
+  an extension build predating #324, or an unreadable/undeployed source tree —
+  never yields `false`. It yields `null` *unless* both versions are known and
+  DISAGREE, which yields `true` — a mismatch is positive proof that the loaded
+  code is not the deployed code, so a missing marker must not discard it. Only
+  `true` is ever reachable from versions alone; `null` is never "fine".
+  Staleness is also **per profile**: one can be current while another is
   not, at the same instant, from one directory.
   ⚠ A marker still cannot see a change you made and never deployed, and it says
   nothing about WHICH DIRECTORY the build came from — for that, read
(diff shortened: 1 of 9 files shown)


ZacxDev:
### Follow-up: the prose still asserted the pre-#331 rule as fact (0c603e8)

This PR's rule change invalidated seven prose sites that still said *"the verdict is computed from the marker, NOT the version"* and *"either side missing a marker → `null`, never `false`"*. That is now wrong in the direction that matters: it tells a reader a version mismatch is unactionable when it is positive proof.

All rewritten to the **asymmetry** rather than softened into "it depends":

| | rule now stated everywhere |
|---|---|
| `false` | two markers present **and identical**. Nothing else can produce it — the #324 invariant, unchanged. |
| `true` | markers differ; or markers agree but versions disagree; or **a marker missing on either side with both versions known and disagreeing**. |
| `null` | marker missing on either side, versions agreeing or either one unknown. |

Only `true` is ever reachable from versions alone, so the point every one of these sites exists to make — *a version cannot certify code as current* — is preserved and sharpened, not weakened.

**Sites corrected**
- `reference/errors.md` — verdict-source ¶ and the fails-closed ¶
- `README.md` — the whoami verdict block, the "not a version comparison" ¶, the "FAILS CLOSED" ¶, and the `GET /health` field reference
- `server.py` — the `HDR_EXT_BUILD` comment and the `/whoami` `extension_build_current` comment
- `extension/service_worker.js` `ping()` and `extension/protocol.js` `pollHeaders()`

**Two other factual defects found in the same sweep**

**A. `SKILL.md` claimed `instances` carries the verdict — it does not.** `GET /instances` returns `registry.snapshot()` **unannotated**; `_describe()` emits no `extension_stale` key, and only `/health` calls `annotate_staleness()`. (`test_server.py:314` already documents this asymmetry.) An agent following SKILL.md would look for the verdict on `/instances`, find the key absent, and read that as "not stale". **Took the doc fix, not the code fix** — annotating an endpoint is a behaviour change and belongs in its own PR. SKILL.md is byte-neutral at **11,988 B (300 B headroom, floor 250)**.

**B. Two contradicting claims about what `extension_version` measures.** This PR softened `server.py`'s docstring to *"describes the manifest of the extension the worker LOADED"*, but `README.md`, `reference/errors.md`, `extension/protocol.js`, `extension/build_id.js` and `gen-build-marker.py` still asserted the stronger *"read off the on-disk manifest at call time"*. Reconciled to the weakened claim, and deliberately **not** upgraded to fact — the load-time-snapshot reading rests on one observation (workbench 2026-08-04: two profiles under one `extension_id` reporting `0.7.1` and `0.8.1` while the directory held only `0.8.1`). Which hypothesis is true doesn't change the conclusion these sites carry.

**⚠️ `extension/build_id.js` is regenerated: `73f5438f18f395d2` → `04bbd6f9c695141d`.** The marker digests every `extension/*.js`, so editing comments in `protocol.js`/`service_worker.js` moves it — the drift gate requires the regeneration in the same commit. Consequence worth knowing at deploy time: **every profile will report `extension_stale: true` until a per-profile Remove + Load unpacked**, even though nothing executable changed. `gen-build-marker.py`'s `render()` template carries the corrected prose too, so the next regeneration won't revert it.

**Gates** (counted, not read off an exit code)
- pytest — `481 passed`, 0 failed
- `node --test` — `# tests 495 / # pass 495 / # fail 0 / # skipped 0`
- `test_skill_size.py` — 4 passed

The only non-comment line in the whole diff is the generated `BUILD_MARKER` literal.

PR #345, opened 2026-08-06
insights: stop reporting an impossible time total; disclose the Layer-B window + denominator

ZacxDev:
The built-in `/insights` produced three quantitative defect classes on 2026-08-05 (plus one suspected fourth). Our Layer A deliberately mirrors its `session-meta` fields (`session-tailer.py:263`), so each was **tested against our code** rather than assumed inherited.

## Verdict table

| # | Hypothesis | Verdict in OUR tool | Evidence |
|---|---|---|---|
| **H1** | Impossible time totals | 🔴 **REAL — fixed** | `session-tailer.py:372` computes `duration_minutes = round((end_ts − start_ts)/60)` — an idle-inclusive span, unclipped to the window. `insights.py:738` summed it into `wall-clock: {N}h`. **Measured** over 2,393 real transcripts touched in a trailing 14d window: **Σ = 720,904 min = 12,015.1 h against the 336 h that EXIST in 14 days — 35.8×.** 10 sessions had a span longer than the whole window alone; worst 818.5 h (2026-06-23 → 2026-07-27). The built-in's `34849`-minute session reproduces **exactly** through our producer. |
| **H2** | Window arg ignored | 🟡 **NOT PRESENT as a data defect; REAL as disclosure — fixed** | Every Layer-A query uses `win`; Layer B uses the deliberately wider `iwin` (`gather()`, spec §11). `write.py:15-18` stamps a `session-insight` row's `ts` with the **session's own `end_ts`, not emit time**, so the built-in's "pooled three older analysis batches" mechanism **structurally cannot occur here**. But `render_html` never labelled the Layer-B window — its subtitle said `trailing 14d · 2 sessions` directly above percentages computed over 30d. `render()` already labelled it. |
| **H3** | Inconsistent denominators | 🔴 **REAL, both renderers — fixed** | Outcome percentages are over `insight_sessions` (Layer B, wider window); the headline `sessions:` is Layer A. Probe with 2 Layer-A / 1 Layer-B session: neither renderer named the denominator anywhere. |
| **H4** | Confabulated friction claim | 🟡 **PARTIALLY REAL — deliberately not "fixed"; see below** | The contract is embedded (`prepare.py:162`) and that embedding is tested (`test_prepare.py:77`, `test_schema.py:89`). But `validate()` polices **structure** only: a payload carrying the exact two claims the built-in confabulated returns `validate() == []` and `vocab_warnings() == []`. |

## What changed

**H1 — the time figure is labelled, not invented.** Two mechanisms drive the overstatement and *neither* is recoverable from Layer A: **idle** (no turn-gap signal in the rollup) and **overlap** (concurrent agent sessions double-count the same minute). Per the brief, an unvalidatable "active time" estimate was rejected. The figure is renamed off `wall-clock` → `session span`, printed beside the window's real bound with an explicit `Nx MORE THAN EXISTS` marker, and carries `SESSION_SPAN_CAVEAT` naming both mechanisms. `render_html` still prints no time figure at all.

Reproduced end-to-end on the real measured population (2,393 sessions / 720,904 min):

```
base:  wall-clock: 12,015.1h across sessions (sum of per-session spans)

HEAD:  session span: 12,015.1h summed over 2393 session spans
                     (the 14d window holds 336h of wall clock — 35.8x MORE THAN EXISTS)
        NOTE: spans are IDLE-INCLUSIVE and overlap across concurrent sessions, and a
              resumed session's span reaches back before the window — so this sum is NOT
              bounded by the window and is NOT time worked. Layer A carries no
              active-time signal; this report has no hours-worked figure.
```

**H2/H3 — windows and denominators are now stated wherever a number is printed.** Both renderers print the Layer-B population, its window, and the explicit contrast with the Layer-A figure above it. The HTML `OUTCOMES` heading carries `trailing 30d`; the subtitle scopes its count to `Layer-A sessions`.

**H4 — the structural mitigation, not a spelled guard.** `render_html` now labels model-authored free text `NOT measured savings`, which `render()` already did — so the report never presents that text as measured. A keyword scan inside `validate()` was **considered and rejected**: it passes the moment the same hazard is worded differently (a *spelled* guard, not a structural one). The limit is pinned by a documented-limit test so closing it stays a deliberate decision.

## Tests — `tests/test_insights_quantitative_honesty.py` (13 new)

Every expected value is a **literal** pinned from fixture arithmetic (40,320 min = 672.0 h; 14d = 336 h; ratio 2.0×), never read back out of the implementation.

**Red-at-base / green-at-HEAD matrix — measured, not asserted:**

| | base (`main` @ `1daca5c`) | HEAD |
|---|---|---|
| new file | **11 failed, 2 passed** | **13 passed** |

Each of the 11 fails at **its own assertion with its own error** — verified line by line, not inferred from the count (e.g. `test_html_labels_model_surfaced_candidates_as_unmeasured` fails on `NOT measured savings` *after* its `wrap deploy dance` precondition passes, proving the guard is reached).

**The 2 green-at-base are labelled NON-regression and mutation-checked for reachability:**

| mutant | guard | result |
|---|---|---|
| M0 control | both | green |
| M1 `render_html` starts emitting hours (**set grows**) | `test_hours_figure_ledger` | red @ L152, ledger's own message |
| M2 `render` stops emitting hours (**set shrinks**) | `test_hours_figure_ledger` | red @ L152, ledger's own message |
| M3 contract clause stripped | `…contract_is_prompt_only_not_validated` | red @ L256, contract assertion |
| M4 `validate()` gains a content check | `…contract_is_prompt_only_not_validated` | red @ L253, `validate() == []` assertion |
| M5 restored control | both | green |

`test_hours_figure_ledger` is a **seam guard** — it fails in *both* directions, so it cannot pass by a renderer silently gaining or losing an hours figure.

**Counts (counted, never read off an exit code):** 358 pre-existing pass unchanged + 13 new = **371 passed, 0 failed, 0 skipped** locally. In the CI sandbox: `scripts/session-analysis/tests` 314/314, `session_insight/tests` 57/57 — 371, matching.

## ⚠️ The `pytests` gate is ALREADY RED on `main` — not caused by this PR

I ran the control before drawing any conclusion:

| | collected | passed | skipped | **failed** |
|---|---|---|---|---|
| `main` (`1daca5c`) | 6051 | 5904 | 3 | **144** |
| this branch | 6064 | 5917 | 3 | **144** |

**Delta: +13 collected, +13 passed, +0 failed, +0 skipped.** All 144 failures are in `scripts/tests/test_opencode_config.py` (`kubectl delete pod x` resolves `allow`, expected `ask` — and 143 siblings), plus 2 unpinned skips in `test_skill_audit.py` (datapacket-talos clone absent). Untouched by this PR, and the base clone has an uncommitted `M scripts/opencode/opencode.jsonc` suggesting someone is mid-work on exactly that. **Flagging rather than merging through it** — a permanently-red gate trains everyone to click through.

## Found but deliberately NOT fixed

- **Clipping spans to the window, and/or a union-of-overlapping-intervals figure.** Both are computable from Layer A's `start_ts`/`end_ts` and would yield a number *bounded* by the window. Both are still idle-inclusive, so neither answers "hours worked" — they would trade an obviously impossible number for a **plausible-looking** one. Flagged, not built.
- **`validate()` cannot police prose** (H4). Reasoning above; pinned by a test so it is rediscovered deliberately.
- **The 144 pre-existing `test_opencode_config.py` failures.** Different subsystem, apparently already in flight.

### DIFF
--- a/scripts/session-analysis/insights.py
+++ b/scripts/session-analysis/insights.py
@@ -130,6 +130,39 @@ QUERY_SECTIONS = {
 
 
 # --------------------------------------------------------------------------- #
+# 🔴 WHY THERE IS NO "HOURS WORKED" FIGURE IN THIS REPORT
+# --------------------------------------------------------------------------- #
+# Layer A's `duration_minutes` is `round((end_ts - start_ts) / 60)` — the span
+# from a session's FIRST message to its LAST (session-tailer.py:372). It is
+# idle-inclusive, and a resumed session's span reaches back arbitrarily far, so
+# Σ duration_minutes is NOT bounded by the report's window and is NOT time spent.
+#
+# MEASURED 2026-08-05 over 2,393 real transcripts touched in a trailing 14d
+# window: Σ duration_minutes = 720,904 min = 12,015.1 h, against the 336 h of
+# wall clock that EXIST in 14 days — a 35.8× overstatement. Ten of those sessions
+# had a span longer than the whole window on their own; the worst was 818.5 h
+# (2026-06-23 → 2026-07-27). The built-in `/insights` reported this same quantity
+# as elapsed time and produced "12224h over a 44-day window" (1,056 h exist).
+#
+# Two mechanisms drive it and NEITHER is recoverable from Layer A:
+#   1. IDLE — a session open across a lunch, a night or a fortnight accrues span
+#      it did not work; there is no keystroke/turn-gap signal in the rollup.
+#   2. OVERLAP — Zach runs many concurrent agent sessions, so spans double-count
+#      the same wall-clock minute across sessions.
+# Clipping spans to the window would fix (1) partially and (2) not at all, and an
+# "active time" estimate would be a number nobody can validate. So the figure is
+# kept, RENAMED off "wall-clock", and printed next to the window's real bound so
+# a reader cannot mistake it for hours worked. `render_html` prints NO time
+# figure at all — see `test_hours_figure_ledger`.
+SESSION_SPAN_CAVEAT = (
+    "spans are IDLE-INCLUSIVE and overlap across concurrent sessions, and a "
+    "resumed session's span reaches back before the window — so this sum is NOT "
+    "bounded by the window and is NOT time worked. Layer A carries no "
+    "active-time signal; this report has no hours-worked figure."
+)
+
+
+# --------------------------------------------------------------------------- #
 # Pure helpers
 # --------------------------------------------------------------------------- #
 def window_seconds(days: int) -> int:
@@ -598,6 +631,11 @@ def aggregate(summary_rows: list[dict], message_rows: list[dict],
         "sessions": len(summary_rows),
         "unreadable_sessions": unreadable,
         "totals": dict(agg),
+        # 🔴 The bound that makes an impossible time total self-evident — see
+        # `SESSION_SPAN_CAVEAT`. `session_span_hours` is Σ per-session span; it is
+        # NOT clipped to the window and NOT bounded by `window_wall_clock_hours`.
+        "session_span_hours": num(agg.get("duration_minutes", 0)) / 60,
+        "window_wall_clock_hours": days * 24,
         "messages": prompt_n + command_n,
         "prompts": prompt_n,
         "commands": command_n,
@@ -735,8 +773,14 @@ def render(data: dict) -> str:
     out.append(f"  tokens:     {total_in:,} in "
                f"({in_fresh:,} fresh + {cache_r:,} cache-read + {cache_w:,} cache-write) "
                f"/ {num(t.get('output_tokens',0)):,} out")
-    dur_h = num(t.get('duration_minutes', 0)) / 60
-    out.append(f"  wall-clock: {dur_h:,.1f}h across sessions (sum of per-session spans)")
+    # 🔴 NOT "wall-clock" — see SESSION_SPAN_CAVEAT. The window's real bound is
+    # printed alongside so an impossible total flags itself.
+    span_h = num(data.get("session_span_hours", num(t.get('duration_minutes', 0)) / 60))
+    bound_h = num(data.get("window_wall_clock_hours", data["days"] * 24))
+    ratio = f" — {span_h / bound_h:,.1f}x MORE THAN EXISTS" if bound_h and span_h > bound_h else ""
+    out.append(f"  session span: {span_h:,.1f}h summed over {data['sessions']} session spans "
+               f"(the {data['days']}d window holds {bound_h:,.0f}h of wall clock{ratio})")
+    out.append(f"   NOTE: {SESSION_SPAN_CAVEAT}")
     out.append(f"  friction:   {t.get('interruptions',0)} interruptions · {t.get('tool_errors',0)} tool errors")
 
     def _bars(title, items, n=8):
@@ -780,6 +824,14 @@ def render(data: dict) -> str:
         out.append("  This report shows ONLY deterministic facts — no fabricated outcomes.")
         return "\n".join(out)
 
+    # 🔴 DENOMINATOR DISCLOSURE. The percentages below are over the Layer-B
+    # population, which is a DIFFERENT set (and a WIDER window) than the
+    # `sessions:` figure in ACTIVITY. The built-in report quoted four session
+    # populations in one document without ever naming which was which.
+    out.append(f"  population: {data['insight_sessions']} session(s) with a Layer-B insight "
+               f"in the trailing {data['insight_days']}d window — the percentages below "
+               f"are over THAT denominator, NOT the {data['sessions']} Layer-A session(s) "
+               f"in the {data['days']}d window above.")
     oc = data["outcomes"] or {}
     total_oc = sum(oc.values()) or 1
     peak = max(oc.values(), default=0)
@@ -885,7 +937,17 @@ def render_html(data: dict) -> str:
                          '<code>session_insight</code> via the activity skill. This report '
                          'shows ONLY deterministic facts — <b>no fabricated outcomes</b>.</p>')
     else:
-        parts = [_html_bars((data["outcomes"] or {}).items(), n=12)]
+        # 🔴 DENOMINATOR + WINDOW DISCLOSURE — the text renderer has carried this
+        # since it was written; the HTML one did not, so a reader saw the page's
+        # "trailing {days}d · {sessions} sessions" subtitle sitting directly above
+        # percentages drawn from a different population over a wider window.
+        parts = [
+            f'<p class="mut">Population: <b>{data["insight_sessions"]}</b> session(s) with a '
+            f'Layer-B insight in the trailing <b>{data["insight_days"]}d</b> window. The '
+            f'percentages below are over THAT denominator — <b>not</b> the '
+            f'{data["sessions"]} Layer-A session(s) in the {data["days"]}d window above.</p>',
+            _html_bars((data["outcomes"] or {}).items(), n=12),
+        ]
         h = data["helpfulness"]
         if h["n"]:
             parts.append(f'<p class="mut">mean Claude-helpfulness '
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Audit round 2 — both 🟡 findings fixed

New sha: `f2cc2b0`. Both reproduced independently before fixing.

### Finding 1 — mis-stated span denominator (REAL, code fix)
`aggregate()`'s loop `continue`s on an unreadable row **before** reaching `duration_minutes`, so an unreadable session contributes no span — but the caveat printed `len(summary_rows)`. Reproduced exactly: 3 sessions / 2 unreadable rendered `sessions: 3 (2 unreadable)` two lines above `summed over 3 session spans`.

Fix: new `session_span_sessions` = `len(summary_rows) - unreadable` — the accumulation loop's *actual* divisor population — exposed as a data key so the number is verifiable without parsing prose.

### Finding 3 — HTML denominator guard blind to its own value (REAL, guard fix)
I re-verified the mutation rather than taking it on report. At `c64f571`, mutating `render_html`'s `insight_sessions` → `sessions` left **all 13 tests green** while the page printed `Population: <b>4</b>` for a true Layer-B population of **1**. Now asserted as one contiguous string including both values.

**Swept the whole file for the same shape** (asserting *around* a value rather than *on* it): line 188 was the **only** instance. Every other value-bearing assertion already pins inline — `336`, `672.0`, `"1 session(s)"`, `"2 Layer-A session(s)"`, `"trailing 30d"`, `"2.0x"`.

### Per-fix matrix vs the PR tip `c64f571`

| fix | test | c64f571 | f2cc2b0 | class |
|---|---|---|---|---|
| 1 | `test_span_denominator_is_the_readable_population` | 🔴 KeyError @ L201 | 🟢 | regression |
| 1 | `test_span_line_discloses_the_readable_denominator` | 🔴 L215 on real output `summed over 5 session spans` | 🟢 | regression |
| 3 | `test_html_report_names_the_layer_b_denominator` | 🟢 | 🟢 | **NOT regression** — guard-strength only; evidence is the surviving→dying mutant |
| — | `test_disclosure_fixture_numbers_are_pairwise_distinct` | 🟢 | 🟢 | **invariant guard** (pins the fixture precondition) |

Finding 1's test was **split in two**: asserting the data key first made the *rendered* line — the thing a reader actually sees — unreachable behind a `KeyError`.

### Mutation battery, re-run in FULL after the fixes

The bash harness misreported M7, so it was rebuilt to **print the source line it actually changed** and to parse pytest **two ways that fail differently** (the `FAILED ::name` summary *and* the `--tb=line` marker) — a bare `assert x in y` emits no `AssertionError` token, which is what the first parser missed. Positive control included.

| mutant | source line | verdict |
|---|---|---|
| M0 / M9 control | — | 16 passed |
| M1 `render_html` gains hours (ledger **grows**) | — | dies L231, ledger message |
| M2 `render` loses hours (ledger **shrinks**) | — | dies L231 + span guards |
| M3 contract clause stripped | 76 | dies L352 |
| M4 `validate()` gains content check | 164+ | dies L349 |
| M6 `render()` wrong population | 841 | dies **L260** |
| M7 `render_html()` wrong population | 955 | dies **L284** — *survived at c64f571* |
| M8 finding-1 revert | 790 | dies **L215** |

Every mutant dies at **its own** guard's assertion; the ledger fails in **both** directions.

### Suite counts (counted, not read off an exit code)

- Local: **374 passed, 0 failed, 0 skipped** (358 pre-existing unchanged + 16 in this file).
- CI `pytests` gate: `scripts/session-analysis/tests` **317/317**, `session_insight/tests` **57/57**.
- Gate totals — `main` 6051/144 failed · c64f571 6064/144 · **f2cc2b0 6067 collected, 5920 passed, 144 failed**. Delta vs main: **+16 collected, +16 passed, +0 failed**. All 144 remain the pre-existing `test_opencode_config.py` failures.

### Deferred items — agreed, with one caveat
No disagreement on deferring the spelled hours-ledger regex, the disclosed-denominator-vs-actual-divisor gap at :831/:836, the HTML "percentages" wording, or the unledgered `--json` surface.

One note: the **:831/:836 gap is now the last instance of the exact defect class this PR is named for** — the disclosed `insight_sessions` and the actual divisor `sum(outcomes.values())` can differ, so the percentages can still be over an undisclosed denominator. Finding 1 was that same bug on the span line and it was worth catching. Happy to leave it, but it's the one I'd pull forward first.

ZacxDev:
## Audit round 3 — :831/:945 denominator gap closed

New sha: `acc0879`. Scope held to this one thing.

### Reachability — measured in BOTH directions before writing any test

| route | reachable? | evidence |
|---|---|---|
| Sanctioned write path | **No** | `validate()` hard-fails a readable payload whose `outcome` is falsy/absent/out-of-enum — checked against `None`, `''`, `0`, `False`, `'unknown_value'`, key-absent: **all six rejected** — and `consolidate.py:97-100` quarantines whatever `validate()` rejects. |
| Data model | **Yes** | `_parse_payload` returns `{}` for a torn/truncated payload, an empty payload column, or valid JSON that isn't an object. `{}.get("unreadable")` is `None` → **falsy** → the row counts as READABLE yet contributes no outcome. Demonstrated: 3 rows (2 good + 1 torn) rendered two outcomes at **50% each** (divisor 2) under a line disclosing **3**. |

So the route is a **torn/empty/non-object payload** — exactly what `write.py`'s sub-PIPE_BUF line budget exists to prevent. **Frequency in the live table is UNMEASURED.** The rendering defect is therefore real and red-before/green-after, but its precondition is a malformed row, not routine operation — the test docstrings say precisely that rather than implying otherwise. These are counted as regression coverage for the *reporting* bug, not as evidence the pipeline emits such rows.

### Fix — and why it's a function, not two format strings

New `outcome_sessions` = `sum(outcomes.values())` (the actual divisor), plus **one** `outcome_population_note()` used **verbatim** by both renderers.

This predicate has now regenerated the same bug **twice** across the renderer pair — H2/H3 (disclosed in text, absent in HTML) and round-2 finding 3 (HTML count unpinned, mutant survived). A predicate duplicated across N sites is wrong at N−1 of them, so it gets one site, and a seam guard asserts the identical string appears in **both** outputs.

**Framing:** I named *both* numbers — `"2 of 3 readable Layer-B insight(s) … reported an outcome — the breakdown covers exactly those 2"` — rather than only printing `sum(oc.values())`. Disclosing only the divisor hides that rows were silently dropped; disclosing only the readable count is the bug. **The number printed is the number divided by.** `or 1` survives solely as a zero-division guard; when it fires `oc` is empty so no percentage is ever printed against it.

One unavoidable side effect: the sentence no longer contains the word "percentages", which was inaccurate on the HTML surface (that page shows counts). The rest of that deferred item is untouched.

### Matrix vs PR tip `f2cc2b0`

**6 failed, 14 passed at `f2cc2b0` → 20 passed at `acc0879`.** All six are genuine regression coverage. The red at `honesty.py:356` is the defect itself: `population: 3 session(s)` printed while the divisor was 2.

| test | f2cc2b0 | acc0879 |
|---|---|---|
| `test_outcome_sessions_is_the_breakdowns_actual_divisor` | 🔴 KeyError | 🟢 |
| `test_text_discloses_the_divisor_not_the_readable_count` | 🔴 L356 | 🟢 |
| `test_html_discloses_the_divisor_not_the_readable_count` | 🔴 L367 | 🟢 |
| `test_both_renderers_emit_the_denominator_note_verbatim` | 🔴 AttributeError | 🟢 |
| `test_text_report_names_the_layer_b_denominator` | 🔴 L260 | 🟢 |
| `test_html_report_names_the_layer_b_denominator` | 🔴 L285 | 🟢 |

### Full mutation battery — 10 mutants, verdicts keyed on TEST NAME

Each mutant prints the source line it actually changed; pytest parsed two ways; controls bookend the run.

| mutant | src line | dies in |
|---|---|---|
| M0 / M11 control | — | nothing (20 passed) |
| M1 `render_html` gains hours (ledger **grows**) | 1051+ | ledger only |
| M2 `render` loses hours (ledger **shrinks**) | — | ledger + span guards |
| M3 contract clause stripped | 76 | contract guard only |
| M4 `validate()` gains content check | 164+ | contract guard only |
| **M6** shared note uses readable, not divisor | 765 | text + HTML + seam |
| **M7** `render()` formats its OWN wrong copy | 872 | **text + seam** |
| **M8** `render_html()` formats its OWN wrong copy | 988 | **HTML + seam** |
| **M9** `outcome_sessions` stops being the divisor | 507 | all four divisor guards |
| M10 span-line divisor reverted (round-2 finding 1) | 821 | span guard only |

M7 and M8 are the wrong-divisor mutant on **each renderer independently** — each dies at its **own** surface's guard *plus* the seam guard, so neither surface is covered only by the other.

### Counts (counted, never an exit code)

- Local: **378 passed, 0 failed, 0 skipped** (358 pre-existing unchanged + 20 in this file).
- CI gate: `session-analysis/tests` **321/321**, `session_insight/tests` **57/57**.
- Gate totals: main 6051 / 144 failed → **acc0879 6071 collected, 5924 passed, 144 failed**. Delta vs main: **+20 collected, +20 passed, +0 failed**. All 144 remain the pre-existing `test_opencode_config.py` failures.

Ready for the `c64f571..acc0879` delta re-audit. Not merged.

PR #364, opened 2026-08-07
fix(rig-control): brightness restore after sleep + stale-restore cancellation (rescued from an un-pushed workbench main)

ZacxDev:
Rescues two commits that existed **only** on the workbench's local `main` and on no remote ref.

## Why this PR exists

`~/workspace/devrc` on the workbench was 2 commits ahead of `origin/main`, on `main`, with neither commit pushed anywhere. Per `CLAUDE.md` → *Git discipline*, `scripts/ship.sh` converges with `merge --ff-only`, so a diverged host is **skipped and left as found** — it silently stops receiving every future change while still looking healthy. That state also blocked deploying an unrelated in-flight PR to the very host it targets.

This is the documented recovery, first half: preserve and push before moving any pointer. The `git reset --keep origin/main` step is **deliberately not done yet** — another session has uncommitted work in that checkout right now.

## What's here

- `1cb7fbc` fix(rig-control): reliable brightness restore after sleep — `scripts/monitor-blackout.sh`, `scripts/rig-control.sh`
- `69363a0` fix(rig-control): aggressive timer cancellation prevents stale restores — `scripts/monitor-blackout.sh`

Both authored 2026-08-06.

## Review note

These commits have **never been gated against the tree they now land in** — they were committed directly to a local `main` and bypassed PR review and CI. That is the reason this is a PR rather than a push to `main`.

Not reviewed as part of this rescue: `claudedocs/handoff-rig-control-brightness.md` is untracked in that working tree and appears to belong to the same effort. It was deliberately left alone rather than swept into this branch, since it is another session's working state. It is unsaved work in the `RULES.md` sense and should be committed or PR'd by whoever owns it.

### DIFF
--- a/scripts/monitor-blackout.sh
+++ b/scripts/monitor-blackout.sh
@@ -20,6 +20,7 @@ set -euo pipefail
 DDC=$(command -v ddcutil) || { echo "ddcutil not found on PATH" >&2; exit 1; }
 UNIT=monitor-blackout-restore
 STATE="${XDG_RUNTIME_DIR:-/tmp}/monitor-blackout.state"   # "bus:brightness"
+FADE_PID_FILE="${XDG_RUNTIME_DIR:-/tmp}/monitor-blackout.fade.pid"
 
 # Number of fade steps and inter-step delay for fade transitions.
 # ~560ms per ddcutil call (hard floor at --sleep-multiplier 0.1), so 15 steps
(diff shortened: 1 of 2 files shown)


ZacxDev:
Superseded by #366, which rescued the same rig-control work (3 commits vs the 2 here) and has already landed on `main` — verified by content, since a squash merge never shows as an ancestor. The stranded handoff doc this PR flagged was separately rescued in #368.

Closing rather than merging: the content is already on `main`, so merging would be a no-op at best. The branch `rescue/rig-control-brightness` is left on origin as a preservation record.

PR #367, opened 2026-08-10
feat(drift): passive drift deadman for the two-host fleet + one shared host-identity predicate

ZacxDev:
## The gap

`scripts/ship.sh` is not the problem — it is correct. It converges both hosts, exits with the right per-host codes, and a host it cannot fast-forward is **skipped and left exactly as found** (`rc=8 skipped:diverged`).

The problem is that **nothing runs it on a schedule** (verified: no systemd timer, no cron — `grep -rn ship.sh nix/` finds only comments and an alias). So a host that starts getting skipped silently stops receiving *every* subsequent change while continuing to look completely healthy: same `git log`, same green home-manager generation, no error anywhere.

That has now happened **twice**:

| date | what | how it was found |
|---|---|---|
| 2026-08-06 | two un-pushed commits on the workbench, blocked for hours | a human shipped something else |
| 2026-08-09 | **three** un-pushed commits on the workbench (rescued as #366), plus 7 untracked files incl. `claudedocs/handoff-rig-control-brightness.md` — the handoff doc for the very work that was stranded | a human shipped something else |

## What this adds

**`scripts/drift-check.sh`** — a passive, read-only deadman. Per host (local + remote over ssh) it answers "is this host silently no longer receiving changes?"

🔴 **Passive means passive.** It may `git fetch` (which writes remote-tracking refs only). It never checks out, switches, merges, fast-forwards, rebases, resets, stashes, cleans, commits, pushes, or runs `home-manager`. If it finds drift it **reports**; it never fixes.

Distinct rc per condition, deliberately using **ship.sh's vocabulary** so the two read consistently — a code means the same thing in both, and codes ship.sh owns for actions this script does not take (5 conflicted-tree, 7 cannot-ff, 9 switch-failed, 11 verify-failed) are left **unused** rather than repurposed:

| rc | condition |
|---|---|
| 3 | repo missing on that host |
| 4 | `git fetch` failed, or origin/main missing / HEAD unborn — *cannot evaluate, which is not a pass* |
| 6 | local host could not be identified |
| **8** | 🔴 local `main` **diverged or ahead** (un-pushed commits) — ship.sh will skip this host forever. **The one that bit us, twice.** |
| 10 | local `main` **behind** origin/main — just needs a ship |
| 12 | the checkout is **not on branch `main`** |
| 13 | host **unreachable** (ssh failed/timed out) |

Two deliberate divergences from ship.sh, both documented in the script header:

- **Worst code wins across hosts** (ship.sh keeps the *first* non-zero). Every per-host line still prints, but a timer's single exit code is the only thing anyone reads — an un-pushed workbench must not hide behind a merely-behind laptop and get described as "needs a ship". Severity: `8 > 13 > 4 > 3 > 12 > 10`; an *unknown* code outranks all of them so a new failure mode can never be silently outranked into invisibility.
- **`main` is compared to `origin/main` regardless of what is checked out.** Local `main` can be diverged while HEAD sits on a feature branch, and that is still a host that will be skipped forever.

Untracked files are **counted and listed per host as information only** — they never change the exit code. Same loss class (work that exists on exactly one machine, in no commit and no backup), zero extra cost.

### Shared host detection — `scripts/lib/host-role.sh`

`detect_role` / `local_ipv4s` / the IP + SSH constants moved **out of `ship.sh`** into a sourced library that both scripts pull in. Both machines report hostname `nixos`, so identity comes from local IPv4 addresses — a predicate that has already been wrong here once (ship.sh hardcoded "local == workbench" and SSH'd to itself while the real workbench was never converged). A second copy would drift and be wrong.

`ship.sh`'s behaviour is unchanged — same `--detect-role` mode, same `SHIP_ROLE`/`REMOTE_SSH`/`LAPTOP_SSH` override semantics, same rc 6. **Its own 29 tests pass untouched** (`test_ship_detect_role.py` + `test_ship_converge.py`).

### Alerting — the existing mechanism, no new notifier

Drift ⇒ non-zero exit ⇒ the unit enters `failed` ⇒ the existing **`OnFailure=notify-failure@%n.service`** fires the same sticky `urgency=critical` dunst toast that ~15 other user units already use. The toast names the unit; the rc legend and per-host lines are in `journalctl --user -u drift-check`.

<details><summary>Why not a bar pill, and what one would take</summary>

The `bar` skill's poller (`bar-status-poll`, ~45s) is architecturally "blocks read local files, the poller fetches out-of-band". A drift source does not fit that cadence: it costs a `git fetch` **plus an ssh round trip**, which is 4×/day work, not 45s work — putting it in the poller would add a standing ssh to a live host every 45 seconds.

A pill is still a small, well-defined follow-up if wanted:
1. `drift-check.sh --status-json` writing `~/.cache/bar-status/drift.json` in the poller's schema (`{count, state, detail}`), 0600, atomic-replace — written by the **drift-check timer**, not the poller (the "block reads a local file" invariant is preserved; only the writer differs).
2. `scripts/i3status-drift`, a ~70-line copy of `i3status-telemetry`'s shape (hide-at-zero, `state:"stale"` → invisible — except drift, like telemetry, must render "cannot tell" as `?` rather than empty, since its reassuring answer is a zero).
3. A `driftBlock` in `nix/graphical.nix` with `signal = 18`, workbench-only, plus `SIGNALS["drift"] = 18` in `bar-status-poll` so the existing `pkill -RTMIN+N` refresh path works.
4. The rising-edge latch already exists (`edge_decision` / `read_latch` / `write_latch`) and would be reused, not reimplemented.

I did not build it because the dunst path is the MVP the deadman actually needs, and half of a pill is worse than none.
</details>

### Timer — `nix/home.nix`, **6-hourly**, master switch **OFF**

Why 6h: the incidents bound it. 08-06 blocked the workbench "for hours"; 08-09 let a whole day's changes miss that host. **Daily** leaves a ~24h window in which every merge silently misses a host — which is the failure itself, only shorter. 6h caps it at a quarter of that for **4 fetch+ssh round trips a day**, a cost too small to trade against. Tighter buys little (a human ships several times a day anyway) and the toast starts reading as noise. `OnStartupSec=10min` gives one prompt run after login — the laptop is intermittent, so "just resumed" is exactly when it should look.

Following the `enableInitiativesSync` pattern, with one refinement the task called for: **the switch gates only the timer's `timers.target` wiring**; the *service* is emitted unconditionally so `systemctl --user start drift-check` works by hand on either host.

```nix
enableDriftDeadman = false;   # ships OFF
...
WantedBy = lib.optionals (serverMode && enableDriftDeadman) [ "timers.target" ];
```

🔴 **Nothing in this PR is deployed.** No `ship.sh`, no `home-manager switch`, no timer armed. `enableDriftDeadman = false` is pinned by a test, so a routine deploy cannot silently arm it.

---

# 🔴 Verification

A deadman that cannot fire is worse than no deadman — it converts "nobody was looking" into "something was looking and said fine". So the instrument was validated in both directions, against **throwaway clones in `/tmp`**. Neither real host was mutated.

## Positive control — it can observe the clean case

```
### POSITIVE CONTROL — clean host (on main, at origin/main)
=== local (workbench) ===
[workbench] untracked: 0
[workbench] ✅ clean — on branch main, main == origin/main (d89bad2f44ceb01788217c60603d23f8956248c1)

drift-check: no drift — both hosts on branch main at origin/main.
rc=0
```

This zero is only meaningful **reported alongside the reds below** — a reassuring zero from a checker wired to nothing looks exactly like this one.

## Negative controls — one per rc, each proven to go RED

<details open><summary><b>rc=8 — un-pushed commits (THE incident shape)</b></summary>

```
=== local (workbench) ===
[workbench] untracked: 1 file(s) — present on this host only, in no commit and no backup
[workbench]     - handoff-stranded.md
[workbench] 🔴 DRIFT — local main is AHEAD by 1 un-pushed commit(s).
[workbench]   ship.sh SKIPS this host (rc=8) and will keep skipping it — it is receiving NOTHING.
[workbench]     + 6fd6a76 fix(rig-control): brightness restore after sleep
[workbench]   rescue (on that host): git branch <topic> main && git push -u origin <topic>
[workbench]   then confirm from ANOTHER host, then: git reset --keep origin/main   (never --hard)

drift-check: DRIFT (rc=8) — see per-host lines above.
  rc3=no-repo  rc4=fetch/origin-main-unavailable  rc6=host-unidentified
  rc8=DIVERGED/AHEAD:un-pushed-commits(ship.sh will skip this host forever)
  rc10=behind(needs a ship)  rc12=not-on-branch-main  rc13=host-unreachable
rc=8
```
</details>

<details><summary>rc=8 — diverged both ways</summary>

```
[workbench] untracked: 0
[workbench] 🔴 DRIFT — local main has DIVERGED: 1 un-pushed commit(s), 1 behind.
[workbench]   ship.sh SKIPS this host (rc=8) and will keep skipping it — it is receiving NOTHING.
[workbench]     + 331f066 local-only commit
[workbench]   rescue (on that host): git branch <topic> main && git push -u origin <topic>
[workbench]   then confirm from ANOTHER host, then: git reset --keep origin/main   (never --hard)
rc=8
```
</details>

<details><summary>rc=10 — behind (needs a ship)</summary>

```
[workbench] untracked: 0
[workbench] DRIFT — local main is BEHIND origin/main by 1 commit(s) — needs a ship.
[workbench]   fix: scripts/ship.sh
rc=10
```
Note it is **not** described as un-pushed or diverged — the distinction is the point.
</details>

<details><summary>rc=12 — not on branch main</summary>

```
[workbench] untracked: 0
[workbench] DRIFT — checkout is on 'feat/left-on-a-branch', not on branch main.
[workbench]   ship.sh will move it, but anything committed there is invisible to origin/main.
rc=12
```
</details>

<details><summary>rc=3 / rc=4 — cannot evaluate ≠ green</summary>

```
### rc=3 — repo missing on that host
[workbench] no repo at /tmp/drift-controls/definitely-not-here
rc=3

### rc=4 — git fetch failed
[workbench] git fetch failed — cannot evaluate drift
rc=4
```
Neither prints "no drift". An absence of evidence is reported as an absence of evidence.
</details>

<details open><summary><b>rc=13 — remote unreachable (REAL ssh, unroutable target)</b></summary>

```
=== remote (laptop — nobody@192.0.2.1) ===
ssh: connect to host 192.0.2.1 port 22: Connection timed out
[laptop] UNREACHABLE — ssh to nobody@192.0.2.1 failed or timed out.
[laptop]   drift on that host CANNOT be evaluated — this is not a pass.
rc=13
```
Real `ssh`, real timeout, real 255 → mapped to 13. Not a stub.
</details>

<details><summary>rc=6 — host identity unresolvable</summary>

```
drift-check: could not identify this host (role='nonsense').
  local IPv4s: 192.168.50.155 10.42.0.100
  expected a workbench (192.168.50.250) or laptop (192.168.50.155) address.
  override with SHIP_ROLE=workbench|laptop to force.
rc=6
```
</details>

## Tests — red → green matrix

**41 new tests** in `scripts/tests/test_drift_check.py`, matching the directory's existing pytest style (throwaway git repos in `tmp_path`, isolated `GIT_CONFIG_GLOBAL`, `testlib.mockbin.write_exec` for the stub `ssh`).

Every one was watched fail against a deliberately broken build. **Two sweeps, 13 mutants, 13 killed, and each by the test that owns the behaviour** — not by collateral:

| # | mutation | tests that went RED |
|---|---|---|
| baseline | *(unmutated)* | **41 passed** |
| M1 | rc 8 collapsed into rc 10 (ahead reported as merely behind) | `test_ahead_unpushed_commits_is_rc8`, `test_diverged_both_ways_is_rc8`, `test_run_against_diverged_repo_mutates_nothing` |
| M2 | aggregate FIRST non-zero (ship.sh style) instead of most severe | `test_worst_code_wins_across_hosts` |
| M3 | ssh 255 passed through instead of mapped to rc 13 | `test_unreachable_remote_is_rc13` |
| M4 | fetch failure made non-fatal (the classic silent green) | `test_fetch_failure_is_rc4` |
| M5 | the deadman starts FIXING (`git checkout main` injected) | `test_run_against_feature_branch_does_not_move_it`, `test_drift_check_source_never_mutates` |
| M6 | host detection re-copied into `drift-check.sh` instead of sourced | `test_host_identity_has_exactly_one_definition` + 3 `test_both_scripts_and_the_lib_agree_on_role[…]` |
| M7 | untracked files silently dropped from the report | `test_untracked_files_are_listed_but_do_not_fail`, `test_untracked_listing_is_capped` |
| M8 | "not on main" silently ignored | `test_not_on_main_is_rc12`, `test_run_against_feature_branch_does_not_move_it` |
| M9 | `pkgs.iproute2` dropped from the unit's PATH | `test_every_command_the_checker_runs_is_on_the_unit_path[ip-pkgs.iproute2]` |
| M10 | master switch flipped ON at merge | `test_the_timer_is_gated_by_the_master_switch_and_the_service_is_not` |
| M11 | `WantedBy` ungated (a routine ship would arm the timer) | same |
| M12 | `OnFailure` dropped (drift produces no notification at all) | `test_home_nix_declares_the_service_and_the_timer` |
| M13 | the *service* gated too (no hand-run possible) | `test_the_timer_is_gated_by_the_master_switch_and_the_service_is_not` |

M9 is the seam nobody owns: the script and the systemd unit are tested in different places. `ip` is how the script decides *which host it is* (both report hostname `nixos`), so dropping `iproute2` from the unit PATH crashes nothing — detection silently returns `unknown` and the deadman exits 6 forever from a unit that looks correct.

The passivity scanner has its **own** controls, because it carves out `say`/`echo`/`printf` lines (drift-check deliberately *prints* the rescue procedure, which contains `git push` and `git reset --keep`):

- `test_scanner_catches_a_bare_mutating_command` — an injected `git checkout main` is caught.
- `test_scanner_catches_a_substitution_inside_a_message` — an injected `say "oops $(git reset --hard …)"` is caught; the carve-out withdraws the moment a line contains a command substitution, which is the one way a printer *can* execute something.
- `test_scanner_carve_out_still_allows_a_plain_message` — proves the two above are not passing merely because the scanner flags everything.

## Authoritative gate

Read as **content**, not exit code (the runner's summary is parsed; the counts are quoted):

```
  PASS  scripts/tests  (collected=1468 passed=1468 skipped=0)     # 1427 → 1468 (+41)
  TOTAL collected=6561  passed=6560  skipped=1  failed=0  (floor: 5600)
RESULT: PASS
```

The 1 skip is the pre-existing pinned `repo-cos` live-store drift check.

---

# What I could NOT verify — stated plainly

1. **The timer has never fired.** Nothing was deployed: no `ship.sh`, no `home-manager switch`, no `systemctl --user start`. The unit definitions are validated by `nix-instantiate --parse` and by the home.nix pins above — that is a claim about the *config*, not about a running unit.

2. **The ssh leg has not been exercised from a systemd-user context.** What I *did* measure, read-only: `ssh -o BatchMode=yes zach@192.168.50.250 bash -s < script` works from this laptop with **`SSH_AUTH_SOCK` unset** — i.e. it authenticated from `~/.ssh/id_ed25519` with no agent, which is exactly the condition a user unit runs under. Strongly indicated, not proven. **The first hand-run on the workbench is what should validate it**, before `enableDriftDeadman` is flipped:
   ```
   systemctl --user start drift-check ; journalctl --user -u drift-check -e
   ```

3. **The dunst toast has not been seen.** The `notify-failure@` path is reused unmodified and is already in service on ~15 units, and the drift service is wired to it — but I did not force a failure on a live host to watch a toast appear.

4. **No live drift was observed, because there is none right now.** A read-only probe of the workbench (`status` + `rev-list` against its *cached* refs — no fetch, no writes) reports: on branch `main`, `0 behind / 0 ahead`, **7 untracked** — including `claudedocs/handoff-rig-control-brightness.md`. So the 08-09 divergence is genuinely resolved by #366, and the untracked set the deadman would report is exactly the one described. Caveat: with no fetch, "0 behind" is against that host's last-known `origin/main`, not necessarily today's.

5. **The `.mjs` gate (`checks.nodetests`) was not run** — this PR touches no `.mjs` file.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -26,7 +26,11 @@ there. Only what's specific to this repo, where a working tree is also a **deplo
   healthy. 2026-08-06: two un-pushed commits on the workbench blocked it for hours, and the
   regrowth timer would have fired on 08-11 running the very bug the undelivered commit fixed.
   **Read every per-host line of `ship.sh`, not the final verdict** — one skip hides among
-  greens, and it prints its own rc legend on failure.
+  greens, and it prints its own rc legend on failure. Recurred 2026-08-09 (three un-pushed
+  commits, rescued as #366) — because **nothing runs `ship.sh` on a schedule**, so the only
+  detector was a human shipping something unrelated. `scripts/drift-check.sh` is the passive
+  deadman for exactly this: READ-ONLY (fetch + report, never fixes), same rc vocabulary as
+  `ship.sh`, run it any time by hand or via the `drift-check` systemd-user timer.
 - **Recovering a diverged host** — preserve, verify, *then* move the pointer:
   `git branch <topic> HEAD && git push -u origin <topic>` on that host → confirm the shas are
   on origin **from a different host** → `git reset --keep origin/main` (`--keep` refuses
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -18,6 +18,16 @@ let
   # DDL/insert path (snapshot #1 wrote 23 rows to prod, telemetry-on, and the DSN role
   # is confirmed to have CREATE SCHEMA). The timer runs hourly (see the timer below).
   enableInitiativesSync = true;
+  # Drift-deadman master switch (scripts/drift-check.sh) — gates ONLY whether the
+  # timer is wired into timers.target. The SERVICE definition is always emitted, so
+  # `systemctl --user start drift-check` works by hand on either host regardless.
+  # 🔴 Deliberately OFF at merge: the deadman was built + validated against scratch
+  # clones only, and enabling a timer is a live change to a host's behaviour. Flip to
+  # true in a separate, supervised deploy after one hand-run on the workbench proves
+  # the ssh leg reaches the laptop from a systemd-user context (a user unit has no
+  # ssh-agent, so the laptop key must be usable non-interactively — the one thing
+  # scratch clones structurally cannot test).
+  enableDriftDeadman = false;
   # Graphical host = runs X/i3 (both current NixOS hosts do; only a genuinely headless
   # box would not). Approximated as isNixOS, mirroring graphical.nix — deliberately NOT
   # !serverMode, which is true on the graphical workbench.
@@ -1811,6 +1821,84 @@ in
     };
     Install = {
       WantedBy = [ "timers.target" ];
+    };
+  };
+
+  # ── PASSIVE DRIFT DEADMAN (scripts/drift-check.sh) ───────────────────────────
+  # The gap this closes: `scripts/ship.sh` converges both hosts correctly, and a
+  # host it cannot fast-forward is SKIPPED and left as found (rc=8). But NOTHING
+  # RUNS SHIP ON A SCHEDULE — so a skipped host silently stops receiving every
+  # change while looking completely healthy. That has now happened twice
+  # (2026-08-06, 2026-08-09), both times found only because a human happened to
+  # ship something unrelated and read the per-host lines.
+  #
+  # This unit is the thing that looks when nobody is looking. It is READ-ONLY: it
+  # fetches (remote-tracking refs only) and reports. It never checks out, merges,
+  # fast-forwards, rebases, resets, or runs home-manager — enforced statically by
+  # scripts/tests/test_drift_check.py.
+  #
+  # ALERTING: no new notification mechanism. The script exits non-zero on drift,
+  # the unit enters `failed`, and the EXISTING OnFailure=notify-failure@%n.service
+  # turns that into the same sticky critical dunst toast every other important
+  # user unit already uses. The toast names the unit; the rc legend + per-host
+  # lines are in `journalctl --user -u drift-check`.
+  #
+  # The service is emitted UNCONDITIONALLY (any host can run it by hand); only the
+  # TIMER's timers.target wiring is gated — see enableDriftDeadman above.
+  systemd.user.services.drift-check = {
+    Unit = {
+      Description = "Passive drift deadman — is either devrc host silently no longer receiving changes?";
+      After = [ "network-online.target" ];
+      Wants = [ "network-online.target" ];
+      OnFailure = [ "notify-failure@%n.service" ];
+    };
+    Service = {
+      Type = "oneshot";
+      # Two `git fetch`es plus one ssh round trip. The ConnectTimeout inside the
+      # script is 10s, so this ceiling only ever fires on a wedged fetch; the
+      # cgroup is killed and the timer re-arms on the next OnUnitActiveSec.
+      TimeoutStartSec = 180;
+      Environment = [
+        # iproute2 is load-bearing, not incidental: `ip -4 -o addr show` is how
+        # local_ipv4s identifies WHICH host this is (both report hostname `nixos`).
+        # Without it detection returns "unknown" and the script exits 6.
+        "PATH=${lib.makeBinPath [ pkgs.git pkgs.openssh pkgs.iproute2 pkgs.bash pkgs.coreutils pkgs.gawk pkgs.gnused pkgs.gnugrep ]}"
+        "HOME=%h"
+      ];
+      ExecStart = "${pkgs.bash}/bin/bash %h/workspace/devrc/scripts/drift-check.sh";
+      # Re-run the unit when either the checker or the shared host-identity
+      # predicate changes.
+      X-Restart-Triggers = [
+        "${../scripts/drift-check.sh}"
+        "${../scripts/lib/host-role.sh}"
+      ];
+    };
+  };
+
+  # 6-hourly, not daily. The choice is bounded by the incident: on 2026-08-06 the
+  # workbench was blocked "for hours"; on 2026-08-09 three commits sat un-pushed
+  # long enough for a whole day's changes to miss that host. Daily would leave a
+  # ~24h window in which every merge silently misses a host — which is the failure
+  # itself, only shorter. 6h caps it at a quarter of that for 4 ssh+fetch round
+  # trips a day, a cost too small to be worth trading against. Anything tighter
+  # buys little: a human ships several times a day anyway, and the toast would
+  # start reading as noise. OnStartupSec gives one prompt run after login (the
+  # laptop is intermittent, so "just resumed" is exactly when it should look).
+  # No Persistent — it only applies to OnCalendar timers, not monotonic ones.
+  systemd.user.timers.drift-check = {
+    Unit = {
+      Description = "Periodic timer for the passive devrc drift deadman";
+    };
+    Timer = {
+      OnStartupSec = "10min";
+      OnUnitActiveSec = "6h";
+    };
+    Install = {
+      # 🔴 The master switch acts HERE and only here: with enableDriftDeadman
+      # false the unit file still exists (so it can be started/inspected by hand)
+      # but is wired into nothing, so no routine `ship.sh` can silently start a
+      # timer nobody has supervised once.
+      WantedBy = lib.optionals (serverMode && enableDriftDeadman) [ "timers.target" ];
     };
   };
 
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -104,6 +104,8 @@ Click actions for those blocks:
 | `run-tests.sh` | **single source of truth** for running the Python suites |
 | `run-node-tests.sh` | single source of truth for running the `.mjs` suites |
 | `ship.sh` | converge BOTH NixOS hosts (workbench + laptop) to `origin/main`, then verify |
+| `drift-check.sh` | **passive deadman** — is either host silently no longer receiving changes? READ-ONLY: fetches and reports, never fixes. Distinct rc per condition (8 un-pushed/diverged, 10 behind, 12 not-on-main, 3/4 cannot-evaluate, 13 unreachable), aligned with `ship.sh`'s legend. Runs unattended as the `drift-check` systemd-user timer; drift ⇒ non-zero ⇒ the existing `notify-failure@` dunst toast |
+| `lib/host-role.sh` | the ONE host-identity predicate (both hosts report hostname `nixos`, so identity comes from local IPv4s). Sourced by `ship.sh` **and** `drift-check.sh` — never copied |
 
 ## Subsystems
 
--- a/scripts/drift-check.sh
+++ b/scripts/drift-check.sh
@@ -1,0 +1,271 @@
+#!/usr/bin/env bash
+# drift-check — PASSIVE deadman for the devrc two-host fleet.
+#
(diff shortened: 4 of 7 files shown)


ZacxDev:
**Corrections to this PR's description, found by an adversarial audit after merge.** Posting rather than editing the body: a reviewer may already have read it, and a silent edit would erase what was believed at review time. Fixes are in #369 (not merged, timer still off).

**(a) The description never states that the timer is workbench-only.** `nix/home.nix` gates it on `serverMode && enableDriftDeadman`, and `~/.server-mode` exists **only on the workbench** (verified on both hosts). So the timer runs on the workbench alone and its remote leg always ssh's to the *laptop* — routinely shut, asleep or off-LAN. The prose reads as a symmetric two-host deadman; it is not, and that asymmetry is exactly what makes the alerting policy wrong: an unreachable laptop produced rc 13 → unit `failed` → the **same sticky critical dunst toast as a genuine rc 8**, up to 4×/day, with a toast body that names only the unit. "My laptop is shut" was indistinguishable from "a host is silently receiving nothing" — a permanently-red gate aimed at the one alert that must keep its meaning.

**(b) "enforced statically by `scripts/tests/test_drift_check.py`" overstates the scanner.** It anchored on the **first word of a line**, so it could not see `git restore --staged .`, `git update-ref`, `git config`, `git gc`, `git prune`, `git worktree add`, `git branch <new> origin/main`, `git symbolic-ref HEAD refs/heads/other`, `git -C "$repo" checkout main`, `rm -rf "$repo"`, `mv`, `sed -i` — nor anything after a separator, since the `say`/`echo` carve-out covered the whole line: `say "checking" && git checkout main` and `echo hi; git reset --hard origin/main` both scanned clean. Measured against a 31-shape probe list: **the old scanner missed 20 of 31**, with zero false positives in either direction. Three docstrings plus this body asserted a guarantee roughly a third as strong as advertised.

**(c) The "13 mutants, 13 killed" sweep has a blind spot.** `drift-check.sh`'s `off_main -> exit 12` sits *after* the ahead/behind comparison, and that ordering is what makes documented divergence #2 work. **Moving that one line above the `ahead` block — deleting nothing — makes un-pushed commits on `main` while HEAD is on a feature branch report rc 12 with the message "ship.sh will move it", which is false** (ship.sh skips that host with rc 8 and it receives nothing, forever). The full suite stayed green under that mutant: **41 passed, 0 failed**. The sweep's blind spot was structural — it varied *what a check returns*, never *the order the checks run in*.

**(d) The positive-control transcript asserts coverage the run did not have.** `--no-remote` on a clean repo printed *"drift-check: no drift — both hosts on branch main at origin/main"*, naming a host the run never contacted; that line appears in this PR's own transcript as evidence. Worse, `--no-local --no-remote` printed the same sentence and exited 0 having checked nothing at all.

None of this is a reason to revert — the deadman closes a real gap and its rc-per-condition design is sound. But (a) would have made the alert noise inside a week, and (c) is a live regression risk in the exact incident shape the unit exists for.

PR #377, opened 2026-08-11
feat(claude): migrate all 17 slash-commands to skills, cut the listing budget 21%, and un-break close-the-loop's write path

ZacxDev:
Custom commands were merged into skills upstream. A command is *a skill with null frontmatter*: no bundled-file directory, no progressive disclosure, and — the actual loss — **no model invocation**, so it only ever fires when a human types it. That is exactly the failure `close-the-loop`'s own STATE.md measured a year ago ("opt-in command-shortcuts don't stick → invest in autonomous triggering"). All 17 move; `claude/commands/` is retired.

---

## A. Command → skill mapping (all 17)

| was `claude/commands/…` | now `claude/skills/…/SKILL.md` | frontmatter added | notes |
|---|---|---|---|
| `adoption-scan.md` | `adoption-scan/` | — (already complete) | |
| `analyze-service.md` | `analyze-service/` | — | 15,349 B, over the auditor's 12 KB target **already on main** — left alone, see "Not done" |
| `audit-pr.md` | `audit-pr/` | — | |
| `devrc-dx.md` | `devrc-dx/` | `argument-hint`, `allowed-tools` | **split** 444 → 137 lines + 3 `reference/` files |
| `espanso-audit.md` | `espanso-audit/` | — | fixed a stale `claude/commands/*.md` pointer inside |
| `find-session.md` | `find-session/` | — | |
| `gpu-operator-check.md` | `gpu-operator-check/` | `argument-hint`, `allowed-tools` | |
| `handoff.md` | `handoff/` | — | |
| `i3.md` | `i3/` | `argument-hint`, `allowed-tools` | **split** 489 → 107 lines + 3 `reference/` files |
| `initiatives.md` | **`initiative-scan/`** | — | **renamed**, collision |
| `obs-read.md` | `obs-read/` | — | |
| `prune-memory.md` | `prune-memory/` | **entire frontmatter** | had none at all |
| `prune-skill.md` | `prune-skill/` | **entire frontmatter** | had none at all |
| `resume.md` | `resume/` | — | |
| `sglang.md` | `sglang/` | `argument-hint`, `allowed-tools` | dangling ref removed |
| `ux-audit.md` | **`ux-sweep/`** | — | **renamed**, collision |
| `verify-agent.md` | `verify-agent/` | — | |

### Collisions resolved

| collision | resolution | reasoning |
|---|---|---|
| `/initiatives` command vs `initiatives` skill | command → **`initiative-scan`** | matches the engine it wraps (`session-analysis/initiative-scan.py`) and the existing `adoption-scan` naming. The `initiatives` **skill** keeps its name for the durable Postgres board/viewer/sync. Both descriptions now point at each other ("the ephemeral scan is `initiative-scan`" / "the durable board is `initiatives`"). |
| `/ux-audit` command vs `ux-audit-loops` skill | command → **`ux-sweep`** | the names differ textually but read as the same thing to a routing model, and every description had to spend chars disambiguating in prose. `ux-sweep` = one-off Playwright sweep of *any* app; `ux-audit-loops` = the re-runnable naida/vetr harnesses; `auditloop` = the hosted crawler. All three now cross-reference explicitly. Renaming the command (not the skill) keeps the blast radius small — `ux-audit-loops` is referenced from the `tekton` and `auditloop` skills. |

Cross-references updated: `standup/SKILL.md`, `standup/standup.sh` (4 sites, including two strings it **prints**), `initiatives/SKILL.md`, `docs/LAYOUT.md`.

### Progressive disclosure

| skill | SKILL.md | reference/ |
|---|---|---|
| `devrc-dx` | 444 → **137** lines (7,794 B) | `config-stack.md` (palette, keybindings, scratch slots, fuzzyclaw wiring, file ownership) · `verification.md` (per-category snippets) · `gotchas.md` |
| `i3` | 489 → **107** lines (6,207 B) | `ipc.md` (query types, tree fields, commands, criteria, recipes) · `interaction.md` (xdotool/clipboard/compound patterns) · `environment.md` (host facts + reliability tiers) |
| `bar` | unchanged | `airvpn.md` → **`reference/airvpn.md`** (was a top-level outlier); both pointers in `bar/SKILL.md` updated |

`scripts/skill-audit.py` reports **25 referenced reference files, all resolve, no orphans ✓** (it reported a broken route before this PR — `close-the-loop` pointed at a bare `reference/telemetry.md`, now a full path).

### Dangling references — removed or corrected, never invented around

| where | claim | what I found | action |
|---|---|---|---|
| `sglang` | `./scripts/toggle-homelab-sglang.sh` "in the wth-is-happening repo" | no such script, and no `wth-is-happening` under `~/workspace` | section replaced with an explicit "there is no toggle script" note + the *verified* homelab facts (the `sglang-server` deployment does exist). The trailing "the toggle script reads the workbench deployment…" bullet is gone; the Flux-reconcile caveat is kept and corrected. |
| `prune-skill` | `scripts/gitops-delta-gate.sh` → `scripts/validate-skill-paths.sh` "in datapacket-talos" | `~/workspace/datapacket-talos` does not exist on this host — unverifiable | the datapacket-specific gate sentence is dropped; the generic rule ("every backticked repo path should exist; `<var>` in prose, `$VAR` in a fence") is kept |
| `prune-skill` | "devrc skills: `nix/home.nix` symlinks **only `SKILL.md`** — NOT `reference/`" | **FALSE.** `recursive = true` ships the whole dir; `~/.claude/skills/{activity,clawgate,…}/reference/` are live on this host right now | rewritten with the true three-way rule. It is only true for the `mkOutOfStoreSymlink` exceptions (`browser`, `dl-router`). |
| `scripts/skill-audit.py` | printed that same false claim on every relative-path warning | same | corrected in the tool, so it stops steering maintainers wrong |
| `devrc-dx` | i3 check grepped `/etc/i3.conf` | retired by PR #74 (i3 moved to home-manager) | now `~/.config/i3/config` |

---

## B. Description budget — before/after

Claude Code loads every skill's `name` + `description` (+ `when_to_use`) on **every session**, under a budget of **1% of the context window**, per-entry cap 1,536 chars. On overflow it **silently drops descriptions**, starting with the skills you invoke least — stripping the trigger keywords that make a skill auto-fire. So a description is *routing surface*, not documentation.

| entry | before | after | Δ |
|---|---:|---:|---:|
| mailbox | 1116 | 429 | −687 |
| auditloop | 890 | 376 | −514 |
| initiatives | 885 | 447 | −438 |
| tekton | 771 | 375 | −396 |
| activity | 725 | 397 | −328 |
| dl-router | 686 | 376 | −310 |
| close-the-loop | 688 | 393 | −295 |
| repo-cos | 670 | 313 | −357 |
| browser | 618 | 409 | −209 |
| ux-audit-loops | 592 | 391 | −201 |
| bar | 579 | 437 | −142 |
| vetr-mailbox | 539 | 328 | −211 |
| obs-read | 495 | 366 | −129 |
| initiatives → **initiative-scan** | 435 | 396 | −39 |
| clawgate | 431 | 336 | −95 |
| standup | 409 | 348 | −61 |
| adoption-scan | 375 | 289 | −86 |
| session-audit | 331 | 300 | −31 |
| ux-audit → **ux-sweep** | 294 | 362 | **+68** |
| analyze-service | 247 | 247 | 0 |
| find-session | 234 | 234 | 0 |
| verify-agent | 229 | 229 | 0 |
| resume | 210 | 210 | 0 |
| audit-pr | 203 | 203 | 0 |
| handoff | 194 | 194 | 0 |
| espanso-audit | 181 | 181 | 0 |
| clickup *(unmanaged)* | 175 | 175 | 0 |
| gpu-operator-check | 127 | 296 | **+169** |
| i3 | 122 | 409 | **+287** |
| sglang | 110 | 340 | **+230** |
| devrc-dx | 102 | 340 | **+238** |
| **prune-memory** | **0 (no frontmatter)** | 315 | **+315** |
| **prune-skill** | **0 (no frontmatter)** | 332 | **+332** |

**Total: 13,663 → 10,773 chars (−2,890, −21.2%).** Entries carrying a description: **31/33 → 33/33.** No entry is anywhere near the 1,536 cap (max 447).

The seven entries that **grew** are the point, not a regression: as commands they contributed almost nothing the model could route on. `prune-memory`/`prune-skill` had no frontmatter at all — survivable for a command, but as skills they would have been **invisible** (confirmed by the opencode negative control below, which drops a frontmatter-less SKILL.md outright).

Method: key use case first, then the literal phrases Zach actually says, then one disambiguation clause against the sibling skill. Narrative ("what it is built from") deleted — that is the body's job and costs 0 until invoked.

**`when_to_use` was deliberately not used.** It concatenates into `description` and counts against the same 1,536 cap, so it buys organisation, not budget. With every entry under 450 chars there is nothing to organise yet; a second key would just be more YAML to keep in sync.

---

## C. close-the-loop's write path — it was inert

The skill's description says it "**updates** STATE.md … every run", its body calls skipping the write-back "the cardinal sin", and its `allowed-tools` grants `Write`/`Edit`. But under the recursive `home.file` mapping both ledger files deployed as read-only `/nix/store` symlinks:

```
$ readlink -f ~/.claude/skills/close-the-loop/STATE.md
/nix/store/4wjrccl…-hm_skills/close-the-loop/STATE.md
$ test -w ~/.claude/skills/close-the-loop/STATE.md   → false
```

Every write silently failed. **Fix:** the source moves to `claudedocs/close-the-loop/{STATE,ARCHIVE}.md` and `mkOutOfStoreSymlink` puts both back at the deployed path pointing at the live checkout.

The move out of the skill dir is **forced**, not stylistic: one path cannot be both a `recursive = true` store symlink and an out-of-store one — the two `home.file` entries collide. `claudedocs/` is where the repo already puts ledgers and handoffs (and `RULES.md` explicitly says work-status belongs there, not in a per-session index).

Verified on the built activation package (no switch performed):

```
~/.claude/skills/close-the-loop/STATE.md   -> …-hm_STATE.md   -> /home/zach/workspace/devrc/claudedocs/close-the-loop/STATE.md
~/.claude/skills/close-the-loop/ARCHIVE.md -> …-hm_ARCHIVE.md -> /home/zach/workspace/devrc/claudedocs/close-the-loop/ARCHIVE.md
```

Side effect: **116 KB** (35 KB STATE + 81 KB ARCHIVE) no longer enters the nix store on every switch. ARCHIVE.md stays reachable at its deployed path so `STATE.md`'s "history lives in ARCHIVE.md next to this file" stays literally true; it is not auto-read, so it costs 0 context. `SKILL.md` gained a table of deployed-path ↔ source-path plus the ⚠ that a write leaves **uncommitted work** in the devrc tree.

---

## D. opencode — what I actually tested, on which version

**opencode 1.18.4** (`~/.nix-profile/bin/opencode`), against a live `opencode serve` + `opencode debug skill`.

| # | claim | how | result |
|---|---|---|---|
| 1 | opencode reads `~/.config/opencode/skills/<name>/SKILL.md` | `opencode debug skill` | ✅ all 16 devrc skills listed with correct `description` + `location` |
| 2 | **negative control** — the instrument can go red | probe skill with **no frontmatter** | ✅ **dropped**: 19 entries → 18 |
| 3 | opencode tolerates the Claude-Code-only keys the migration adds | probe skill carrying `when_to_use`, `argument-hint`, `allowed-tools`, `disable-model-invocation`, `user-invocable` | ✅ loaded, description intact |
| 4 | skills are still typable as `/<name>` | `GET /command` on `opencode serve` | ✅ skills appear **in the same list as commands**, tagged `"source":"skill"` |
| 5 | the skill body becomes the command template | same endpoint | ✅ body present, `$ARGUMENTS` preserved verbatim |
| 6 | opencode's own docs agree | its built-in `customize-opencode` skill | ✅ documents `~/.config/opencode/skill(s)/<name>/SKILL.md` as a global-skill location, and `~/.claude/skills/` as auto-loaded external |

On that evidence the `~/.config/opencode/commands` mapping is **removed** (it pointed at `claude/commands/`, which no longer exists — leaving it would fail nix evaluation, so "leave it in place" was not an available option). All the measurements are recorded in the `nix/home.nix` comment so the next reader does not have to re-derive them.

`~/.claude/skills/clickup/` — the unmanaged standalone checkout that `nix/home.nix` depends on — is **untouched and still resolves** (`.config/opencode/skills/clickup -> hm_clickup -> ~/.claude/skills/clickup`), confirmed on the built tree.

---

## Verification

**Build / eval**
- `nix-instantiate --parse nix/home.nix` → OK.
- `nix build .#homeConfigurations.zach.activationPackage --impure` → **builds**. This is the check that matters: it is what proves the new `mkOutOfStoreSymlink` entries do not collide with the recursive mapping.
- Inspected the built `home-files/`: `~/.claude/commands` **gone**; `~/.config/opencode/commands` **gone**; 32 skill dirs present; `devrc-dx/reference/`, `i3/reference/`, `bar/reference/` all ship; both ledger symlinks point at the working tree.

**nodetests — PASS**
```
PASS  scripts/browser-bridge/tests      (files=15 tests=495 pass=495 skipped=0 floor=490)
PASS  scripts/dl-router/tests           (files=13 tests=508 pass=508 skipped=0 floor=500)
PASS  scripts/collector/browser-ext/tests (files=2 tests=21 pass=21 skipped=0 floor=20)
TOTAL suites=3 files=30 tests=1024 pass=1024 fail=0 skipped=0  (floor: 970)
RESULT: PASS
```

**pytests — FAIL, but `origin/main` is ALREADY RED with the identical failures**
```
this branch:  TOTAL collected=6743  passed=6739  skipped=1  failed=3   RESULT: FAIL
origin/main:  TOTAL collected=6743  passed=6739  skipped=1  failed=3   RESULT: FAIL
```
Control run: `nix build "git+file:///home/zach/workspace/devrc?ref=refs/heads/main#checks.x86_64-linux.pytests"` — byte-identical counts. All three failures are `scripts/tests/test_monitor_blackout.py` (`test_status_no_timer`, `test_get_brightness_returns_value`, `test_blackout_fails_when_no_monitor`), from the canonical-path guard added in **ca3eff6 "fix(rig-control): guard against running from wrong path (#374)"**: the guard compares `readlink -f "$0"` against `${HOME}/workspace/devrc/scripts/monitor-blackout.sh`, and inside the hermetic sandbox `HOME=/build/home` while the script sits at `/build/src/…`, so it refuses and exits 1. **Not fixed here** — unrelated to this PR and it deserves its own change rather than being buried in a 50-file migration. 🔴 Flagging it loudly because a permanently-red gate trains everyone to click through.

**Content-loss audit of the migration (the check the review agent was going to run)**

The first adversarial-audit agent died on an API limit before reporting, so I ran the
highest-value part of its brief deterministically instead: a **token-coverage check** across
all 17 migrations. For each command, every distinctive token of the `origin/main` version
(identifiers, paths, flags, numbers, words > 3 chars; frontmatter excluded, since descriptions
were deliberately rewritten) is matched against the union of the new `SKILL.md` and its
`reference/*.md`. Anything present before and absent after is reported.

*Line-for-line diffing was the wrong instrument here* — the split reflows bullet lists into
paragraphs, so verbatim matching reported rewording as loss (117 false positives on `devrc-dx`
alone). Token coverage tolerates rewording and still catches a dropped fact.

**Negative control:** injecting a sentence of invented tokens (`zzqqx`, `frobnicator`,
`--murgatroyd`, `/var/lib/wibblesprocket`) into the base text makes the checker fire for
**all 17**. A clean pass is therefore a measurement, not an absence.

Result: **9 of 17 lost nothing at all** (the pure frontmatter migrations — `adoption-scan`,
`analyze-service`, `audit-pr`, `find-session`, `handoff`, `initiative-scan`, `obs-read`,
`resume`, `verify-agent`). The other 8 lost only text this PR deliberately removed or
reworded, each re-checked by hand:

| skill | lost tokens | verdict |
|---|---|---|
| `espanso-audit` | `claude/commands` | intended — corrected to `claude/skills/<name>/SKILL.md` |
| `gpu-operator-check`, `sglang` | `Triggers`/`Usage`/`Actions` headings, "Questions about…" bullets | intended — trigger prose moved *into* `description`, where the model can route on it |
| `sglang` | `toggle-homelab-sglang.sh`, `SGLANG_API_BASE_2`, `simulation.py`, `RoundRobinLM` | intended — the whole block belonged to the absent `wth-is-happening` repo |
| `prune-memory` | "command" → "skill" | intended |
| `prune-skill` | the false symlink claim, `gitops-delta-gate.sh`, `validate-skill-paths.sh` | intended, as documented above |
| `ux-sweep` | old heading, `/verify` | intended — `/verify` is not a devrc command; replaced with the real siblings |
| `devrc-dx`, `i3` | the split | see below |

For `devrc-dx`/`i3` I checked each flagged token against the new files directly. Most were
**tokenizer artifacts** — `mem/disk/net/…`, `Shift\+x.*i3lock`, `split/tabbed/stacking`,
`true/false`, and all four `Tool Coordination` bash commands are present, just punctuated
differently. **Three were real losses, and are restored in 7ba46b5:**

1. the nix modules that `builtins.readFile` `.tmux.conf` / `.zshrc`
   (`nix/programs/{tmux,zsh}/default.nix` — both verified to exist);
2. the two-pass search performance contract (instant substring over cached fields + one
   batched `rg -i -l --max-count=1 --glob=*.jsonl` across all cwds, ~66 ms / 1.2 GB) — a
   thing `devrc-dx` is meant to *check*, not fuzzyclaw trivia;
3. which bash scripts the Go binary replaced (`idle-update.sh`, `pipe-activity.sh`, most of
   `task-hook.sh`/`task-resume.sh`).

**Skill hygiene**
- `scripts/skill-audit.py` on the migrated tree: reference integrity **25/25 resolve, no orphans ✓**. `devrc-dx` and `i3` are now *within* the 12 KB target (they were the two biggest bodies).
- `scripts/browser-bridge/tests/test_skill_size.py`: SKILL.md 12,034 → **11,821 B**, headroom **254 → 467** (min 250). The description trim moved it *away* from the ceiling.
- `name:` matches the directory for all 30 in-tree skills (opencode requires this; Claude Code uses the directory name for the command).

---

## Not done / could not verify

1. **`$ARGUMENTS` substitution in opencode for a skill-sourced command.** The body reaches the template with `$ARGUMENTS` intact, but a skill entry's `hints` array is empty where a command entry's is `["$ARGUMENTS"]`. That is at least a TUI-autocomplete difference; proving substitution actually happens needs a live model call, which I did not make. Claude Code's own behaviour **is** documented and unaffected (`$ARGUMENTS`, `$ARGUMENTS[N]`, `$N` all supported in skills).
2. **No `home-manager switch` was run** — per instruction; Zach applies via `scripts/ship.sh` after merge. So this is verified against the *built* activation package, not a live one. **Post-switch check:** confirm `~/.claude/commands/` and `~/.config/opencode/commands/` are actually gone on both hosts — home-manager removes the links it owns, but a leftover directory would be worth knowing about, and a surviving stale `~/.claude/commands/<name>.md` would re-register a duplicate command.
3. **Seven SKILL.md bodies remain over the auditor's 12 KB target** — `auditloop` 21.5 KB, `activity` 21.5, `initiatives` 18.8, `analyze-service` 15.3, `mailbox` 14.1, `clawgate` 13.7, `repo-cos` 13.3. All pre-existing; only `analyze-service` came through this migration. Out of scope — that is a `prune-skill` pass.
4. **The adversarial review agent never completed** (API weekly limit). I replaced its
   content-loss brief with the deterministic token-coverage check above, but its other
   categories — a second pair of eyes on the nix change, on second-order `ship.sh` effects,
   and on description/body contradictions — were **not** independently reviewed. Everything
   reported here is my own verification.
5. **No behavioural test that a migrated skill auto-fires.** The whole point of the migration is model invocation, and I have no deterministic harness for "did the model pick this skill". The description rewrites are argued from the documented mechanism, not measured against it. The honest verifier is the `adoption-scan` skill in a few weeks.
6. **`disable-model-invocation` was set nowhere.** Auto-invocation is the point, and loading a skill only injects instructions — actions still pass through the permission system. If any of these turn out to fire too eagerly, that flag is the one-line fix (it also frees listing budget, since the description then leaves context entirely).
7. **`claudedocs/close-the-loop/STATE.md` content is unchanged** — moved with `git mv`, byte-identical.


ZacxDev:
### Held back from today's merge queue — a semantic conflict with #378, not a textual one

`#392 → #384 → #391` are merged. This PR was gated alongside them on an integration tree and is the only one held.

**It merges textually clean and is still broken.** On the merged tree the authoritative gate is red:

```
TOTAL collected=7168  passed=7147  skipped=1  failed=20   RESULT: FAIL
  FAIL scripts/tests (collected=1920 passed=1900 failed=1 errors=19)
```

Attributed by bisect, not inference — same tree without this PR is green in both tiers:

| tree | sandbox pytests | dev-host tier |
|---|---|---|
| `main` + #392 + #384 + #391 | 7168 / **failed=0** PASS | 7168 / **failed=0** PASS |
| the same + **this PR** | 7168 / **failed=20** FAIL | — |

**Root cause.** `scripts/tests/test_subsystem_resolver.py` arrived in #378, which landed on `main` *after* this branch's base (`5c53f38`), so this branch has never seen it. It pins the command doc by path:

```python
COMMAND_DOC = ROOT / "claude" / "commands" / "analyze-service.md"   # line 77
```

This PR moves that file to `claude/skills/analyze-service/SKILL.md`, so the fixture raises `FileNotFoundError: /build/src/claude/commands/analyze-service.md` — 19 setup errors — plus 1 real assertion failure in `test_the_doc_path_is_the_deployed_one`, which is precisely the guard written to catch a pin that no longer points at the file that ships. It did its job.

**Fix looks small — two lines, no re-hashing:**
1. line 77 → `ROOT / "claude" / "skills" / "analyze-service" / "SKILL.md"`
2. line 1724 `assert COMMAND_DOC.parent.name == "commands"` → the skills-path shape (`"analyze-service"`), and the surrounding docstrings that say `claude/commands/`.

The `HASHED_REGIONS` sha256 pins do **not** need re-measuring: the move is byte-identical (`cmp` on `5c53f38:claude/commands/analyze-service.md` vs this branch's `SKILL.md` exits 0), so every hashed region still hashes the same.

**Also needs a rebase**: `mergeable=CONFLICTING` against current `main`, and `MIN_TESTS` is now **7168** on `main` (7122 + 5 #392 + 16 #384 + 25 #391). This PR is net-zero on collected count, so after the rebase 7168 should still be the measured total — but measure it, don't carry it.

One cosmetic follow-up, not a blocker: #384's `test_ship_converge.py` fixture spans four `home.file` families including `.claude/commands/standup.md`. Those manifests are fabricated in `tmp_path`, so nothing breaks — but once this lands, that comment names a family that no longer exists.

ZacxDev:
### Fixed, rebased, and green on BOTH tiers — merging

Rebased onto `main` @ `52fd995` (rebase, not a merge, so the squash stays clean) and fixed the pin. Head is now `eb3210b`.

**The `cmp` claim in the diagnosis above was right about the wrong baseline.** It compared the branch's `SKILL.md` against `5c53f38:claude/commands/analyze-service.md` — the branch's own base. But #378 didn't just add the test, it also **edited the doc**, adding the `resolver-rules` / `entry-schema` hash markers that `HASHED_REGIONS` measures:

| blob | path | bytes |
|---|---|---|
| `8b69dc0c` | `5c53f38:claude/commands/analyze-service.md` **and** `191a336:…/SKILL.md` | 15349 |
| `207ad356` | `origin/main:claude/commands/analyze-service.md` | 16709 |

So pre-rebase the branch was carrying the **pre-#378** copy, with no hash markers in it at all. Had that shipped, `_region()` would have raised `marker '<!-- resolver-rules:begin' is missing` — a different failure from the one predicted, not a hash mismatch.

The rebase's rename detection carried #378's edits onto the moved path, and *that* is what makes the pins survive. Verified, with a positive control proving `cmp` can go non-zero, and cross-checked against `diff` (a second tool that fails differently):

```
cmp origin/main:claude/commands/analyze-service.md  HEAD:claude/skills/analyze-service/SKILL.md   -> rc=0   (identical)
cmp origin/main:claude/commands/analyze-service.md  191a336:claude/skills/analyze-service/SKILL.md -> rc=1   (differ at byte 2554)
```

`HASHED_REGIONS` therefore needs no re-measuring — but for the post-rebase reason, not the stated one.

**Symptom reproduced before fixing.** Pointing the constant back at the retired `claude/commands/` path on this tree gives `1 failed, 195 passed, 19 errors` — the reported `failed=20` exactly, with the single failure being `test_the_doc_path_is_the_deployed_one` raising its own assertion.

**The guard was strengthened rather than just repointed.** `assert COMMAND_DOC.parent.name == "commands"` is a tautology against the constant on the line above it, so it could only ever catch the file *vanishing*, never the pin drifting off the deployed set. It now also asserts `nix/home.nix` still declares the `~/.claude/skills` mapping and sources `../claude/skills` — the half that makes it a claim about **deployment**. Both halves mutation-tested: each fails with its **own** message, and the `home.nix` half executes past the path assertions (reachable, not merely breakable). `COMMAND_DOC` → `SKILL_DOC`, since an identifier reading "command doc" for a skill doc is a false claim. The class name `TestCommandDocIsPinned` is deliberately **not** renamed — `SKILL.md` names it inside a hashed region, so touching it would move a sha.

### Verification matrix — control vs merged, both tiers

Control is unmodified `origin/main` @ `52fd995`. Merged tree is `eb3210b` (the rebase makes the merge a fast-forward, so the tree I tested *is* the branch).

| tree | sandbox pytests | dev-host tier | nodetests |
|---|---|---|---|
| control `52fd995` | 7168 collected / 7167 passed / 1 skipped / **failed=0** · `RESULT: PASS` | 7168 / 7167 / 1 / **failed=0** · `RESULT: PASS` | tests=1024 pass=1024 **fail=0** · `RESULT: PASS` |
| merged `eb3210b` | 7168 collected / 7167 passed / 1 skipped / **failed=0** · `RESULT: PASS` | 7168 / 7167 / 1 / **failed=0** · `RESULT: PASS` | tests=1024 pass=1024 **fail=0** · `RESULT: PASS` |

Read from the runner's summary CONTENT, never an exit code. Per-directory totals are identical between control and merged (`scripts/tests` = 1920 on both). `panic: test timed out`: 0 occurrences in all four logs. The one skip is the same pinned `repo-cos` opt-in live-store check on both trees — nothing skipped its way to green, and the 26 pinning tests plus all four dangling-symlink cases were confirmed **PASSED by name**, not skipped.

**The exit code would have lied again.** My first dev-host runner reported `rc=0` over a `RESULT: FAIL` / `failed=1` line — the fourth agent this week to hit that. The failure was `test_engine_is_the_version_every_measurement_is_keyed_to`: I had passed the whole `REQUIRED_TOOLS` list to `nix-shell -p`, which prepended the ambient channel's opencode **1.18.13** ahead of the host's deployed **1.18.4**. An instrument defect, not a tree defect — confirmed by running the *identical* runner against unmodified `main`, which failed identically on the same test with the same version string. The corrected runner appends only what the host actually lacks (a python carrying the test deps, and `logrotate`) and leaves the host PATH alone. That pair also serves as the harness's negative control: same `run-tests.sh`, same tree, red under v1 and green under v2.

**MIN_TESTS stays 7168 — measured, not carried.** This branch does not touch `run-tests.sh`, and the rebase produced **no conflict at all** on that line, so there was no `Resolved ... using previous resolution` and no replayed floor in play. (The rebase was run with `rerere.enabled=false` anyway.) Net-zero on collected count, measured rather than assumed: `test_subsystem_resolver.py` 215 + `test_ship_converge.py` 33 = 248 on both `52fd995` and this branch, and every other per-directory total is unchanged. Measured total 7168 = floor 7168.

### Cleanups included

* `test_ship_converge.py`'s fixture named `.claude/commands/standup.md` as one of its `home.file` families, and the comment said "four" while listing five. The manifests are fabricated in `tmp_path` so nothing was broken — but the dangling-symlink negative control was parametrised over a shape that can no longer occur. Replaced it with the live recursive opencode skill mirror (`.config/opencode/skills/`), which strictly improves that control, and fixed the count.
* Repointed the stale `claude/commands/analyze-service.md` references in `scripts/lib/subsystem_resolver.py`'s docstrings and in this module's own error messages.
* Re-grepped `scripts/`, `claude/`, `docs/`, `nix/` and `CLAUDE.md`. Everything still mentioning `claude/commands/` is deliberate — the retirement records in `CLAUDE.md`, `docs/LAYOUT.md`, `nix/home.nix` and `scripts/opencode/README.md`, and a synthetic string in an rsync-comment parser test. One deliberate non-change: `scripts/direnv/direnvrc`'s `use_opencode` targets **project-local** `.claude/commands/` in arbitrary repos, which upstream still supports — that is not this repo's retired directory and is not false.

PR #386, opened 2026-08-11
fix(initiatives): the abandoned clawgate fetch worker never died, and the comment saying it couldn't matter expired

ZacxDev:
## The defect

`scripts/initiatives/tasks.py::_get` read the `GET /api/tasks` body with `resp.read(n)`.
On a `BufferedReader` that **blocks until it has ALL n bytes or hits EOF**, so with
`READ_CHUNK = 64 KiB` the between-chunk wall-clock re-check was **unreachable** for the
entire time a chunk was in flight. `_run_with_deadline` gave up at `FETCH_DEADLINE` and
returned — while its daemon worker kept running. One **live thread + its FDs leaked per
abandoned fetch**, bounded only by `SOCKET_TIMEOUT`.

This was already written down in `claude/skills/initiatives/reference/clawgate-tasks.md`
under **"KNOWN, MEASURED, NOT FIXED"**, with `read1()` named as the fix and the whole thing
excused as *"bounded today only because the real payload is ~16 KB."*

## 🔴 That bound expired — the leak was LIVE

**Measured 2026-08-11 against live clawgate 0.7.85:**

| | |
|---|---|
| `GET /api/tasks` body | **94,428 bytes** |
| tasks in the queue | **10** |
| `READ_CHUNK` | 65,536 bytes |

94,428 ≥ 65,536, so the read spans **two** chunks and the first one blocks a full 64 KiB
past the 2.0s deadline. The old comment's own numbers ("~16 KB for today's 6 tasks",
"1 MiB is ~60x today's payload") were off by ~6x — 1 MiB was ~**11x** the payload on that
date. `Notes.List` has no `LIMIT` and rows carry full task bodies, so the payload only ever
grows: the excuse was always going to expire, and it did before anyone re-read it.

## The fix

`resp.read(...)` → `resp.read1(...)`. `read1(n)` returns as soon as *some* data is available
rather than blocking for the full n, so the loop comes back to the clock promptly.
Worst-case termination delay is now **one `read1`**, i.e. `SOCKET_TIMEOUT`, not "however
long the peer takes to deliver 64 KiB".

**Cap semantics are unchanged, deliberately.** The size argument
`min(READ_CHUNK, (max_bytes + 1) - total)` is an over-read **BY ONE** — it lets `total` reach
`max_bytes + 1` so the check below *detects* an oversized body instead of silently
truncating at the cap. `read1` returns *at most* that many bytes, never more, so the
behaviour is byte-for-byte identical; `test_real_get_caps_the_response_body` and
`test_oversized_response_is_refused_rather_than_buffered` pass unmodified.

## Comments corrected in the same change

Both are now written as **dated measurements**, not standing properties — "~16 KB for
today's 6 tasks / ~60x" is precisely the shape that rotted into a false safety argument and
was then *cited* as the reason the leak couldn't matter:

1. the `MAX_RESPONSE_BYTES` note (reality: 10 tasks, 94,428 B, ~11x — measured 2026-08-11);
2. `_run_with_deadline`'s docstring, which asserted the worker *"terminates itself shortly
   after being abandoned"* — false until this commit, and now states what bounds "shortly";
3. (bonus) the module docstring's layer-2 paragraph and its stale `~5 tasks` figure.

`clawgate-tasks.md` moves this out of "KNOWN, MEASURED, NOT FIXED" into a **✅ FIXED**
section recording the measurement that killed the old bound. The two genuinely-unfixed items
in that section (serve-stale has no upper age bound; `group_by_slug` keys on slug alone) are
kept, untouched.

## The test

`test_abandoned_worker_dies_promptly_on_a_chunk_spanning_body` asserts **WHEN THE WORKER
DIES**, not what it parsed — a body that merely parses correctly does not reproduce this bug.

A local socket (no live clawgate) serves a **~311 KB** body — several `READ_CHUNK`s — paced
**2 KiB / 80 ms**, so assembling one 64 KiB chunk takes **~2.6s** against a **0.4s** deadline,
while every individual send sits well inside the socket timeout. That leaves the
between-chunk clock re-check as the only thing that can end the read. It drives the real
production composition (`fetch_tasks_result` → `_run_with_deadline` → `_get`, via a spy that
only records when `_get` returns) and additionally pins that the worker died with **its own
`TimeoutError`**, **mid-body** (`0 < bytes_read < len(body)`) — so a socket timeout, an EOF,
or a completed read cannot green it for the wrong reason.

### Red → green matrix

| tree | result |
|---|---|
| `origin/main` `tasks.py` (`read`) + this test | **1 failed**, 38 passed |
| HEAD (`read1`) | **39 passed** |

Failure message at `origin/main`:

```
AssertionError: the ABANDONED worker was still running 1.40s in — 0.40s after the 0.4s
deadline was declared blown. It is blocked inside ONE read() waiting for a full 65536-byte
chunk (~2.6s at this pace), so the wall-clock re-check never runs. Read the body with read1().
```

### Mutation testing

Both mutants fail on **this guard's own assertion**, at `test_tasks.py:508`, *after* the
earlier assertions in the test passed — i.e. the guard is **reached**, not short-circuited by
an earlier check, and not killed by a different guard's error:

| mutant | result |
|---|---|
| `resp.read1(...)` → `resp.read(...)` (everything else at HEAD) | ❌ worker still alive 1.0s past the deadline |
| between-chunk clock re-check removed, `read1` kept | ❌ same |

Also run 5× consecutively on the fix to check for timing flake: 5/5 passed.

## Verification

```
scripts/initiatives/tests/ + test_skill_audit.py + test_no_public_ips.py   840 passed
```

(counted from the pytest summary, not an exit code; `test_skill_audit.py` and
`test_no_public_ips.py` because `claude/skills/` was touched)

Full `scripts/tests/` was also run as a control: **26 failed / 1626 passed on this branch and
26 failed / 1626 passed at `origin/main`** — identical, all from dev-environment deps missing
in the bare `nix-shell` (opencode config/engine, `run_tests` preconditions), none related to
this change.

## Not fixed here

* Serve-stale has **no upper age bound** — clawgate down ⇒ the last good map is served
  forever. Still open, still recorded in the reference doc.
* `group_by_slug` keys on **slug alone, not (repo, slug)** — zero collisions measured, still
  latent. Still open.
* The residual termination bound is `SOCKET_TIMEOUT` (2.0s): a peer that accepts and then
  sends *nothing at all* still parks one `read1` for that long. That is the intended floor —
  `SOCKET_TIMEOUT` exists for exactly this — but it means "the worker dies at the deadline"
  is really "within `SOCKET_TIMEOUT` of it", and the docstring now says so rather than
  implying otherwise.
* `_get` now requires the response object to expose `read1`. Real `urlopen` responses
  (`http.client.HTTPResponse`) do; there is deliberately **no** fallback to `read`, because a
  silent fallback would silently reintroduce exactly this leak. A response object lacking it
  would raise, which this module's best-effort contract degrades to "no linked tasks".

### DIFF
--- a/claude/skills/initiatives/SKILL.md
+++ b/claude/skills/initiatives/SKILL.md
@@ -17,7 +17,7 @@ Arc: `devrc/claudedocs/handoff-initiatives-nextstep-dispatch-shipped-2026-07-26.
 **Reference files** (`~/.claude/skills/initiatives/reference/`, source
 `~/workspace/devrc/claude/skills/initiatives/reference/`) — read on demand:
 - `viewer-board.md` — board layout, states/chips/search, per-card actions, archive lifecycle. Read when changing the viewer UI or debugging a card's lane/state.
-- `clawgate-tasks.md` — the `initiative:<slug>` tag join, dispatch guard, fetch shape + a known thread/FD leak. Read when a dispatched task doesn't link, or when touching tagging.
+- `clawgate-tasks.md` — the `initiative:<slug>` tag join, dispatch guard, fetch shape + the fetch worker's termination bounds (body phase fixed 2026-08-11; connect/headers still unbounded, and the single-flight guard does NOT hold it to one thread). Read when a dispatched task doesn't link, when touching tagging, or before claiming that fetch cannot leak threads.
 - `agent-devpod.md` — devpod deploy, kubeclaw 0.7.x hardening values, image `2026.6.11-py` + web-search-off, identity pin, least-priv DB role. Read when deploying/bumping the agent pod or debugging a crash-loop.
 
 ## How a card comes to exist (discovery model)
@@ -116,7 +116,7 @@ curl -s -X POST http://192.168.50.250:8899/api/dispatch -H 'Content-Type: applic
   -d '{"repo":"<repo full path OR repo_name>","slug":"<slug>"}'   # 200 {ok,task_id} / 400 / 404 / 502
 # dispatch.py mirrors repo-cos/clawgate.py: POST clawgate /api/tasks {directory(=label),body,repo?,tags} —
 #   model OMITTED → clawgate default deepseek. Confirm the card: GET {CLAWGATE_API_URL}/api/tasks (Bearer).
-# Linked-task join, dispatch guard, fetch shape + the known leak → reference/clawgate-tasks.md
+# Linked-task join, dispatch guard, fetch shape + what still leaks → reference/clawgate-tasks.md
 ```
 
 **Grounded next step (nextstep.py, read-only):** every card carries one — documented cards
--- a/claude/skills/initiatives/reference/clawgate-tasks.md
+++ b/claude/skills/initiatives/reference/clawgate-tasks.md
@@ -40,12 +40,28 @@ WALL-CLOCK deadline**, **1 MiB cap**, performed **OUTSIDE `DataProvider._lock`**
 serve-stale-on-failure, single-flight. Silent best-effort: any clawgate failure logs to
 stderr and the board renders unchanged.
 
-## 🟡 KNOWN, MEASURED, NOT FIXED
-The abandoned fetch worker does **NOT self-terminate**. `resp.read(n)` blocks on a
-`BufferedReader` until n bytes or EOF, so with `READ_CHUNK=64 KiB` the wall-clock re-check
-is never reached for bodies ≥64 KiB (**measured: 6 fetches → 6 live threads, +12 FDs, none
-reclaimed**). Bounded today only because the real payload is ~16 KB. **Fix = `read1()`.**
+## ✅ FIXED — abandoned-worker leak (`read` → `read1`)
+The abandoned fetch worker used to **NOT self-terminate**: `resp.read(n)` blocks on a
+`BufferedReader` until n bytes or EOF, so with `READ_CHUNK=64 KiB` the between-chunk
+wall-clock re-check was **unreachable** for bodies ≥64 KiB (**measured then: 6 fetches → 6
+live threads, +12 FDs, none reclaimed**). It was excused as "bounded today because the real
+payload is ~16 KB".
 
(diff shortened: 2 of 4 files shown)


ZacxDev:
Adversarial audit run pre-merge. **Verdict: no 🔴 on the code** — the `read1` change is correct and was confirmed independently of this PR's own test (abandoned worker 2.09s past the deadline on `main` → 0.00s here, on **both** `Content-Length` and chunked bodies; cap semantics byte-identical across 20 boundary cases; the new test non-flaky over 18 runs including 4× CPU oversubscription).

**But the PR body and two comments were measurably false, and are corrected in 4e0de94** — flagging publicly rather than silently editing, since the wrong version has been readable here:

1. *"Worst-case termination delay is now ONE `read1`, i.e. `SOCKET_TIMEOUT`"* — **false**. `read1` fixes the **body phase only**. DNS + connect + **headers** are still a blocking `urlopen`, and `SOCKET_TIMEOUT` is per-*operation*, so a peer dribbling the response header one byte at a time (every send inside the timeout, so no per-op timeout fires) holds the worker open **>60s past a 2.0s deadline at production constants — identical before and after this PR**. It is the same `_slow_drip` shape `tasks.py` already documents at +130s, two paragraphs above the claim. The blackhole peer named as the worst case is actually the **best** case (~0.00s).
2. *"the single-flight guard means at most one such thread exists at a time"* — **false**. `LinkedTaskCache.get` releases `_refresh_lock` when `loader()` **returns** (at the deadline), not when the worker dies, so any worker outliving the TTL lets the next refresh start another. **Measured: 7 concurrent `initiatives-clawgate-fetch` threads.** This was the sentence that made the residual sound harmless.

Both residuals are now recorded as OPEN in `clawgate-tasks.md`, and `initiatives/SKILL.md` no longer routes readers to a "known leak" this PR marked fixed.

The irony is the point: a PR fixing an expired claim shipped two new ones. Gates: 840 passed.

PR #388, opened 2026-08-11
feat(browser-bridge): a 15-min heartbeat, because no budget can tell "unused" from "down"

ZacxDev:
> **Body rewritten after audit.** The original version claimed the heartbeat detects a silent extension drop — it does not, and the code now says so explicitly. See [the audit comment](https://github.com/innovation-upstream/devrc/pull/388#issuecomment-5250215543) for what changed and why.

## The report

`deadman.py` flagged `laptop/browser-bridge` **DEAD** (silent 30.8h active, budget 14.2h) while the unit was healthy — the bridge journal since the last reboot had 5 lines, all `listening`, zero `dispatch`. The workbench bridge journal over the same window had 72 lines with `dispatch`/`cmd_ok`, the positive control that the log *does* record commands when they happen. Nobody had driven it.

## Why "raise the budget" was rejected

The ask was to raise the budget. The measurement says that cannot work — the silence was 2.2× anything the source had ever done, and it grows without bound:

| | buckets | active hours |
|---|---|---|
| worst lull ever observed (14d) | 170 | 14.2h |
| silence when it alarmed | 371 | **30.9h** |
| budget at K=2 (shipped) | 170 | 14.2h → DEAD |
| budget at K=4 | 340 | 28.3h → **still DEAD** |
| module CAP | 576 | 48h → buys ~2 days, then re-alarms |

Two other levers measured and rejected: **max-gap instead of p99** (`workbench/keys` carries a 163-bucket max gap from a past outage, which would inflate the most continuous source's budget from 2h to 27h — it converts a past death into permanent blindness), and a **density-derived Bernoulli bound** (comes out *below* the observed p99 for all 17 pairs, because emissions are bursty rather than uniform).

The root cause was in the source: browser-bridge emitted one event per handled command and nothing else, so normal silence is unbounded by construction.

## What this changes

**1. A 900s liveness heartbeat** (`server.py`). 900s = 3 of the deadman's 5-min buckets, so the p99 active-gap collapses to ~3 and the budget bottoms out on the **2-active-hour floor**, where silence means the unit genuinely is not running. `BROWSER_BRIDGE_HEARTBEAT_S=0` disables it. No manual restart on deploy — the unit already has `X-Restart-Triggers` on `server.py`.

`kind="heartbeat"`, **not** `"cmd"`: `adoption-scan.py:149` counts `source='browser-bridge' AND kind='cmd'` as the usage signal, so heartbeats on `kind=cmd` would report ~96 phantom browser-skill uses/day.

**What the row proves, stated honestly:** the deadman consumes only the row's *existence*, so this detects "the bridge process is running". `payload.connected` is diagnostic metadata — nothing branches on it, so a bridge whose extension has silently dropped still reads alive. That gap predates this PR (nothing detected an extension drop before either; you found out when a command failed), and wiring an alarm to it is a separate change with its own consumer.

**2. `MACHINE_CADENCE` in `deadman.py`** — and this is the more important half. `active_by_host` was built from *any* source's buckets, so a 24/7 emitter marks the operator's night ACTIVE for every source on the host, and idle sources then accrue "active" silence while nobody is at the desk. Measured on the real 14-day table, sweeping the evaluation point hourly over 5 days (121 points) — a single evaluation at `now` is always taken while the operator is active and shows nothing:

| pair | as-is | heartbeat as operator | heartbeat as machine-cadence |
|---|---|---|---|
| workbench/i3 | 0 | 47 | 0 |
| workbench/tmux | 0 | 47 | 0 |
| workbench/keys | 0 | 45 | 0 |
| workbench/opencode | 0 | 28 | 0 |
| workbench/zsh | 0 | 22 | 0 |
| workbench/claude | 0 | 19 | 0 |

Six pairs that never alarm today would have gone DEAD, including 20 continuous hours on `workbench/keys` across a Saturday the operator was away. It does not self-correct: p99 sits just under the nightly gaps because there are only ~14 of them among ~1300.

So a `(source, kind)` in `MACHINE_CADENCE` counts toward its **own** pair's liveness and is excluded from the host's active-time definition. The query grew an `n_op` column built *from* that tuple so the SQL cannot drift from the Python; `parse_buckets` accepts 4- or 5-field input (a pre-heartbeat capture is all-operator, which is correct) but refuses a width that changes mid-file rather than guessing. Re-verified on live data: **zero** new dead pairs, `workbench/browser-bridge` alive at 121/121 points.

**3. Consumers.** `agent-ops`' freshness query imports `MACHINE_CADENCE` and excludes those pairs (an unfiltered `max(ts)` reports a bridge with no extension attached as "fresh 15 min ago"). `collector/README.md` documents both kinds.

## Tests

24 new tests. The ones that matter:

- `workday_table` — a fixture with **real idle time**. `healthy_table` fills every bucket, so the entire suite was structurally blind to the dimension the 🔴 bug lives on. The regression test carries its own negative control, so it cannot quietly stop meaning anything.
- Seam guards asserting *relationships* against the real definitions, not copies: heartbeat kind vs `adoption-scan.REGISTRY`; `HEARTBEAT_INTERVAL_S` vs `deadman.BUCKET_SECONDS`/`FLOOR_BUCKETS`; `agent-ops`' filter vs `MACHINE_CADENCE`; the SQL `check()` actually sends.

## Mutation battery — 18 mutants, all killed, five only after a fix

Five survived their first run and are the honest content of this section:

1. **`parse_buckets` reuses `n` for `n_op`** — survived all 360 tests, and would have reinstated the 🔴 bug on live data, since the check only ever reads the table through that function. Cause: a fixture with `n == n_op == 3`, which makes "carry the column" and "ignore it" byte-identical. Values are now pairwise distinct plus a TSV→parse→evaluate test.
2. **empty `MACHINE_CADENCE`** — died in the browser-bridge suite but survived in `test_deadman.py`, where the guard was a `for` loop over the tuple: vacuous when empty.
3. **`check()` sending `cadence=()`** — every `check()` test injected a fake fetch that ignored its query argument, so nothing asserted the production SQL.
4. **cadence terms joined with `AND`** — `countIf(NOT (A AND B))` excludes nothing (206760 of 206760 rows counted against the live table); only reachable once a second cadence pair exists, which the module's own note tells maintainers to add.
5. **`start_heartbeat(Registry())`** — a fresh registry instead of the server's passed an `isinstance` check while making the heartbeat report `connected=False` forever.

Also killed, having survived earlier rounds: `900`→`90000` (a silent no-op with every test green), and deleting `start_heartbeat` from `main()` (feature completely inert, suite green).

## Gates

- `nix build .#checks.x86_64-linux.pytests` → **6766 collected, 6765 passed, 1 skipped, 0 failed** (floor 6745)
- `nix build .#checks.x86_64-linux.nodetests` → rc 0

## Not verified

Live behaviour. `server.py` is a `home.file` copy (`readlink -f` → `/nix/store/…`), so nothing changes on either host until `ship.sh` runs. Post-deploy this needs: heartbeat rows appearing on a 900s cadence on both hosts, and the `laptop/browser-bridge` DEAD verdict clearing.

https://claude.ai/code/session_01EdoJTPJPnwrbQjc265rZbm

### DIFF
--- a/claude/skills/activity/SKILL.md
+++ b/claude/skills/activity/SKILL.md
@@ -70,6 +70,14 @@ Per-source detail that matters when querying:
   stream), `session-summary` (Layer A rollups), `session-insight` (Layer B facets).
 - `i3` — `window::focus` + `workspace::focus` → `i3-source`; captures attention even when
   NOT typing. Carries `app`=WM_CLASS + `payload.workspace`.
+- `browser-bridge` — TWO kinds, and the distinction is load-bearing. `kind=cmd` is one row
+  per agent-driven browser command — the **usage** signal `adoption-scan.py` counts.
+  `kind=heartbeat` is machine-generated every 15 min by the bridge daemon itself
+  (`payload.connected` = is an extension attached), and exists ONLY so the deadman has a
+  cadence to measure. 🔴 **Filter `kind='cmd'` in any usage/adoption query** — counting both
+  reports ~96 phantom uses/day. Before the heartbeat (< 2026-08-11) the source emitted on
+  commands only, so its silence was unbounded and it false-alarmed as DEAD whenever Zach
+  simply hadn't browsed; do not "fix" a browser-bridge DEAD verdict by raising a budget.
 
 **Browser extension is NOT fully nix-managed.** Hand-loaded unpacked in Brave (the laptop's
 daily browser) from `~/.local/share/activity-browser-ext/` — a real-file copy, since
--- a/scripts/agent-ops
+++ b/scripts/agent-ops
@@ -350,30 +350,34 @@ def _load_chquery():
     return mod
 
 
-def machine_cadence_sql_filter():
+def machine_cadence_sql_filter(cadence=None):
     """`AND NOT (...)` excluding every MACHINE_CADENCE pair the deadman declares.
 
-    Returns "" when the set is empty (nothing emits on a timer -> nothing to
-    exclude). Raises if deadman.py cannot be imported; the caller turns that into
-    the panel's error marker, which is the honest outcome — silently falling back
-    to an unfiltered query would put a "fresh 15 min ago" next to a bridge that
-    cannot execute anything.
+    🔴 BOTH the pairs AND the predicate's RENDERING come from deadman — this
+    function only wraps them. An earlier version imported the pairs but re-spelled
+    the `" OR ".join` here, and that copy was immediately wrong in the same way
+    the original was (joining with AND makes `NOT (A AND B)` exclude nothing),
+    reachable the day a second timer-driven emitter is added. Single-sourcing the
+    data is not enough; the expression has to be single-sourced too.
 
-    That raise adds NO new failure mode: the only caller already loads
-    validation/chquery.py from the same DEVRC_DIR one line earlier, so a checkout
-    that cannot satisfy this import could not have satisfied that one either.
+    Returns "" when the set is empty (nothing emits on a timer -> nothing to
+    exclude), which still yields valid SQL at the call site. `cadence` overrides
+    the imported set for tests.
+
+    Raises if deadman.py cannot be imported; the caller turns that into the
+    panel's error marker, which is the honest outcome — silently falling back to
+    an unfiltered query would put a "fresh 15 min ago" next to a bridge that
+    cannot execute anything. Note the asymmetry with chquery, whose absence is a
+    silent no-refresh: this one is LOUD.
     """
     import importlib.util
     spec = importlib.util.spec_from_file_location("_ao_deadman", DEADMAN_PATH)
     mod = importlib.util.module_from_spec(spec)
     sys.modules[spec.name] = mod
     spec.loader.exec_module(mod)
-    pairs = tuple(mod.MACHINE_CADENCE)
-    if not pairs:
-        return ""
-    terms = " OR ".join("(source = '%s' AND kind = '%s')" % (s, k)
-                        for s, k in pairs)
-    return "AND NOT (%s) " % terms
+    pairs = tuple(mod.MACHINE_CADENCE) if cadence is None else tuple(cadence)
+    pred = mod.cadence_predicate_sql(pairs)
+    return "AND NOT (%s) " % pred if pred else ""
 
 
 def read_env_file(path):
--- a/scripts/agent-ops
+++ b/scripts/agent-ops
@@ -377,9 +377,15 @@ def query_telemetry_freshness(timeout=TELEM_QUERY_TIMEOUT):
     conn = Q.CHConn.from_env(env)  # raises if CLICKHOUSE_URL is absent
     conn.timeout = float(timeout)
     client = Q.CHClient(conn)
+    # Machine-generated heartbeats are excluded: browser-bridge emits one every
+    # 900s whether or not it can execute anything, so counting them would make
+    # this column read "fresh 15 min ago" for a bridge with no extension attached.
+    # Freshness here means OPERATOR-driven activity (same distinction deadman.py
+    # makes with MACHINE_CADENCE).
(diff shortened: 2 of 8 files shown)


ZacxDev:
## Audit found a deploy-blocker. Fixed in ef0e4c2 — and two claims in the PR body above are now wrong.

**Correcting the body rather than editing it**, since it may already have been read:

1. *"the heartbeat observes the half that actually breaks (a silent extension drop)"* — **it does not.** Nothing branches on `payload.connected`: the deadman counts rows, adoption-scan excludes heartbeats, agent-ops reads only `max(ts)`. The field is diagnostic metadata you can query after the fact. An extension drop remains undetected — as it was before this PR, so it is not a regression, but the body oversold it. The docstring now says so plainly.
2. The body's mutation-testing section was **incomplete**. Two further mutants survived the suite as it stood; both are now killed. Details below.

### 🔴 The blocker: a 24/7 emitter redefines ACTIVE time for every other source

`deadman.py` built `active_by_host` from *any* source's buckets. Giving browser-bridge a 900s cadence therefore marked the operator's night ACTIVE for every source on that host, and idle sources accrued "active" silence while nobody was at the desk.

Measured on the real 14-day table, sweeping the evaluation point hourly over 5 days (121 points). The sweep is the whole method — a single evaluation at `now` is always taken at a moment the operator is active, and shows nothing:

| pair | as-is | heartbeat as operator | heartbeat as machine-cadence |
|---|---|---|---|
| workbench/i3 | 0 | 47 | 0 |
| workbench/tmux | 0 | 47 | 0 |
| workbench/keys | 0 | 45 | 0 |
| workbench/opencode | 0 | 28 | 0 |
| workbench/zsh | 0 | 22 | 0 |
| workbench/claude | 0 | 19 | 0 |

Six pairs that never alarm today would have gone DEAD — including a 20-hour continuous false DEAD on `workbench/keys` across a Saturday the operator was away. It does not self-correct with more history: p99 sits just under the nightly gaps because there are only ~14 of them among ~1300.

**Fix**: a `(source, kind)` in the new `MACHINE_CADENCE` tuple counts toward its own pair's liveness and is excluded from the host's active-time definition. The query grew an `n_op` column built *from* that tuple, so the SQL cannot drift from the Python. Re-verified on live data: zero new dead pairs, and `workbench/browser-bridge` alive at 121/121 points.

### The tests were structurally blind to it

`healthy_table()` fills *every* bucket, so no fixture in the suite had idle time — the one dimension this bug lives on. Added `workday_table()` (10 waking hours/day) and a regression test that carries its own negative control, so it cannot quietly stop meaning anything.

### Mutation battery: 12 mutants, all killed — two only after a fix

- **M9 — `parse_buckets` reuses `n` for `n_op`** (which reinstates the bug on live data, since the check only ever reads the table through that function): **SURVIVED** all 360 tests. Cause was a fixture of equal values — `n == n_op == 3` makes "carry the column" and "ignore it" byte-identical. Values are now pairwise distinct, plus a TSV→parse→evaluate test that crosses the boundary every other test skips.
- **M6 — empty `MACHINE_CADENCE`**: killed in the browser-bridge suite but **SURVIVED** in `test_deadman.py`, where the guard was a `for` loop over the tuple — vacuous when empty. Now asserted non-empty before the loop.
- Also killed, having previously survived: `900` → `90000` (a silent no-op leaving every test green), and deleting `start_heartbeat(registry)` from `main()` (feature completely inert, suite green).

Other audit items: single lock snapshot in `emit_heartbeat_event`; the registry-liveness test now crosses `CONNECT_STALE_S` and pins both sides; `agent-ops` freshness excludes heartbeats; `collector/README.md` documents both kinds.

Gates after the fix round: **pytests 6766 collected / 6765 passed / 1 skipped / 0 failed** (floor 6745); **nodetests rc 0**.

Still unverified: anything live. Needs `ship.sh` before it does anything on either host.

PR #393, opened 2026-08-11
fix(deadman): active time is HUMAN time — an unattended overnight agent marked the night ACTIVE and read keys/tmux as DEAD

ZacxDev:
## The bug

The deadman decides "dead" by measuring a source's silence in **ACTIVE time**. A bucket became ACTIVE if **any** source emitted an operator-driven event in it, where "operator-driven" meant "not in `MACHINE_CADENCE`". That is a **DENYLIST**, and it defaults every source not explicitly denied — including every **agent-driven** one — to "counts as the operator being present".

2026-08-11: an unattended overnight agent session emitted `claude`/`tool` rows for hours while the human slept. That marked the night ACTIVE, so `workbench/keys` and `workbench/tmux` blew their 2-ACTIVE-HOUR floor and read **DEAD** by morning — the same failure shape the heartbeat denylist was added to fix, from a source nobody thought to deny.

## The change

```python
PRESENCE_SOURCES = frozenset({"keys", "i3", "tmux", "zsh", "browser"})
```

A bucket is ACTIVE only if a **human-presence** source emitted in it. The asymmetry is total: when the operator really is at the machine, `keys`/`tmux`/`i3` have already marked those buckets active, so an agent row adds nothing; agent rows add buckets **only** when the operator is away — which is exactly when adding them manufactures a false alarm.

The mechanism mirrors what it replaces. The query's fifth column `n_op = countIf(NOT (<cadence pairs>))` becomes `n_presence = countIf(source IN (...))`, built **from** `PRESENCE_SOURCES` so the SQL cannot drift from the Python.

## Measured on the live table

Evaluation point swept **hourly over 97 points (4 days)** across the real 14-day table. Each side runs its **own production query** (`bucket_query` → `fetch_buckets` → `parse_buckets` → `evaluate`), not a re-implementation: OLD is `deadman.py` @ `3acd041`, NEW is this branch. Dead-hours per pair:

| pair | OLD | NEW |
|---|---:|---:|
| `workbench/keys` | 6 | **0** |
| `workbench/tmux` | 7 | **0** |
| `workbench/browser-bridge` | 15 | **2** |
| `laptop/browser-bridge` | 91 | 88 |
| `laptop/opencode` | 0 | **1** |
| `laptop/claude` | 50 | 50 |
| `laptop/tmux` | 3 | 3 |

**Positive control, same run** — seed a *real* death by dropping a pair's last N hours, then re-judge at the same instant under both rules:

- **24h drop:** all seven pairs **agree**. Six read DEAD under both (`workbench/opencode`, `workbench/keys`, `workbench/tmux`, `laptop/claude`, `workbench/browser-bridge`, `laptop/browser-bridge`); `workbench/tool` reads ok under both — its measured budget is ~20 active hours, so 24 wall-hours is not yet enough for either rule.
- **72h drop:** one pair **disagrees** — `workbench/tool` is caught by the denylist and not by the allowlist. Quantified below.

## The costs — stated, not hidden

**1. The documented blackout blind spot WIDENS.** Active time now advances only while a presence source emits, so it no longer takes *every* source dying: losing **all five presence sources at once** is enough. An X-session crash is the realistic shape (it takes `keys`+`i3` together, and the terminals take `tmux`+`zsh`). After that, active buckets stop accruing, every silence measures 0, and the check goes **QUIET instead of alarming** — even though an agent may still be emitting. Under the old rule that agent would have kept the timeline moving and the presence sources would have alarmed.

The trade is deliberate: the alarms removed were false (the operator was asleep); the alarm given up fires only when the operator's whole desk is gone, which `newest_event_age_minutes` already reports. **Losing one presence source is still caught** — the others carry the timeline, which is the whole motivating case. Both halves are pinned by `test_losing_every_presence_source_at_once_goes_QUIET_not_loud`.

**2. Sensitivity moves BOTH ways.** Budgets are denominated in active buckets, so a shorter active timeline moves the budget *and* the measured silence together, and the net differs by source:

- **More sensitive** for a source pinned on `FLOOR_BUCKETS` — the floor is a fixed 24 active buckets, which now spans fewer wall hours. `laptop/opencode` gained a dead-hour (0 → 1) for exactly this reason.
- **Less sensitive** for an on-demand source whose budget scales with the same shrinking timeline. Seeding a 72h death on `workbench/tool`: the old rule measured **23.6 silent active hours against a 21.7h budget** (DEAD); the allowlist measures **18.2 against 19.4** (not yet). Silence shrank faster than the budget because the removed buckets were disproportionately the overnight agent-only ones. It is a **DELAY, not a blind spot** — the same seeded death at 120h is caught by both rules.

This corrects a claim I had written before measuring it: "budgets tighten, so the check is uniformly more sensitive" is **false**, and the docstring now says so with both numbers.

## `MACHINE_CADENCE` is untouched and still exported

It no longer participates in active time — `PRESENCE_SOURCES` subsumes it, since `browser-bridge` is not a presence source and its heartbeat therefore cannot mark a bucket active regardless of `kind`. The two are deliberately **not** layered together: one rule, one place.

Both `MACHINE_CADENCE` and `cadence_predicate_sql` remain public because `scripts/agent-ops` imports them (`machine_cadence_sql_filter`) for its telemetry-freshness panel, where the concept is genuinely different — that panel excludes **machine-generated** rows while still counting `claude`/`tool` as real usage of those sources. `scripts/tests/test_agent_ops.py`'s seam tests are unchanged and green. The deadman's own test now also asserts the cadence predicate has *not* leaked back into `bucket_query`.

## Legacy 4-field TSV

A 4-field row (pre-change capture, or a `--tsv` replay) has its presence column **reconstructed from the source name** rather than copied from `n`. That is exact, not a guess: presence is a property of the source alone and the source column is right there — unlike the machine-cadence rule, which also needed `kind` and genuinely could not be recovered. Copying `n` (the old meaning) would reinstate the bug on every replayed capture; hard-zeroing would make a replay unable to judge anything and report a reassuring "0 dead".

## Tests

New cases all use a **day/night fixture**. `healthy_table` fills every bucket, so it has no idle time and is structurally blind to this entire class of bug.

- `test_an_unattended_agent_at_night_does_not_make_human_sources_look_dead` — the regression test, with an **in-test negative control** proving the same fixture reproduces the bug under the old rule (`_as_any_source_active`). If the control stops failing, the green stops meaning anything.
- `test_a_human_source_that_genuinely_stops_is_still_dead` — positive control, literal pins (120 silent active buckets vs the 24-bucket floor).
- `test_an_agent_source_that_genuinely_stops_is_still_dead` — a non-presence source is still judged on its own liveness.
- `test_losing_every_presence_source_at_once_goes_QUIET_not_loud` — pins the cost above, plus the contrast that losing only one presence source **is** caught.
- `test_the_presence_allowlist_survives_the_WHOLE_path_tsv_to_verdict` — end-to-end through `parse_buckets`, both widths.
- `test_presence_sources_are_the_five_human_driven_ones` — the allowlist as a **literal**, not derived from the module.

## Mutation results — 12/12 killed, 0 survived

Harness: fresh copy per mutant, `__pycache__`/`*.pyc` purged, `python -B` with `PYTHONDONTWRITEBYTECODE=1` (a byte-length-identical mutant otherwise validates a stale `.pyc` on `(mtime, size)` and pytest silently executes the ORIGINAL source, making every mutant read as "survived"). Anchor uniqueness is asserted, so a mutant that failed to apply reports `BROKEN-MUTANT` rather than "survived".

**Harness controls, run first:** a no-op (comment-text) mutant that **must survive** → SURVIVED; a canary (`n_presence > 0` → `>= 0`) that **must die** → killed. Without both, a survivor means nothing.

|---|---|---|
| `PRESENCE_SOURCES` emptied | killed (16 failed) | `test_presence_sources_are_the_five_human_driven_ones` + 15 |
| presence source removed (`tmux`) | killed (4 failed) | literal-set, legacy-parse, both SQL tests |
| agent source ADDED (`claude`) | killed (7 failed) | literal-set, legacy-parse, both SQL tests, unattended-agent |
| `parse_buckets` `in` INVERTED | killed (3 failed) | `test_parse_buckets_reconstructs_presence_for_a_legacy_four_field_capture` |
| `evaluate` `in` INVERTED | killed (11 failed) | `test_an_unattended_agent_at_night_...` + 10 |
| SQL `countIf` built from the WRONG set (`MACHINE_CADENCE`) | killed (2 failed) | both SQL tests |
| `check()` sends a query built with a DIFFERENT set | killed (1 failed) | `test_the_query_check_ACTUALLY_SENDS_carries_the_presence_column` |
| empty allowlist degrades to `count()` (denylist behaviour) | killed (1 failed) | `test_presence_count_expression_matches_the_presence_set` |
| `parse_buckets` 4-field reuses `n` (the OLD meaning) | killed (2 failed) | legacy-parse + whole-path |
| `evaluate` 4-tuple reuses `n` (the OLD meaning) | killed (2 failed) | unattended-agent + blackout-cost |
| `bucket_query` drops the presence filter | killed (2 failed) | both SQL tests |
| presence rendering unsorted (unstable SQL) | killed (2 failed) | both SQL tests |

`check()`'s test asserts the **SQL that `check()` actually sends** (every other `check()` test injects a fake fetch that ignores the query), and it pins the literal `IN`-list rather than a substring — `"AS n_presence"` alone is satisfied by `"AS n_presence_unused"`.

## Gates — content read, not exit codes

```
  TOTAL collected=7200  passed=7199  skipped=1  failed=0   (floor: 7168)   RESULT: PASS
  (the 1 skip is the pinned opt-in repo-cos live-store drift check)

  TOTAL suites=3 files=30 tests=1024 pass=1024 fail=0 skipped=0  (floor: 970)   RESULT: PASS
```

`scripts/tests/test_agent_ops.py`, `scripts/tests/test_bar_status.py` and the browser-bridge suite were also run directly (671 passed) since they are the seam consumers.

## Docs

`scripts/collector/deadman.py` module docstring (step 2 + a new 🔴 COST section), `scripts/collector/README.md`, `claude/skills/activity/SKILL.md` (active-time definition + the widened blind spot).

## NOT VERIFIED

- **Live behaviour after deploy.** Nothing was deployed — no `home-manager switch`, no `ship.sh`, not merged. The bar's `tlm` pill keeps running the OLD deadman on both hosts until someone switches. The live measurements above were made by importing this branch's `deadman.py` directly and querying the real table read-only; they are evidence about this source, not about a deployed artifact.
- **Whether `browser` belongs in the allowlist on the workbench.** It has 0 rows there (laptop-only extension), so it contributes nothing to workbench active time. It is included because it is genuinely human-driven; it is unexercised on one host.
- **Behaviour beyond the 4-day / 97-point sweep window.** The 14-day table was fetched in full, but the evaluation point was swept over the last 4 days only.
- **The `workbench/tool` delay against a real (unseeded) outage.** The 72h-vs-120h boundary is measured against a seeded drop, not an observed death.

### DIFF
--- a/claude/skills/activity/SKILL.md
+++ b/claude/skills/activity/SKILL.md
@@ -175,22 +175,22 @@ Creds come from `~/.config/activity-collector/env` (the collector's own file) un
 - 🔴 **Active time is an ALLOWLIST — `deadman.PRESENCE_SOURCES` = `keys` `i3` `tmux` `zsh`.**
   Agent/machine sources (`claude`, `tool`, `opencode`, `browser-bridge`) do NOT mark a bucket
   active, and neither does **`browser`**: the browser-bridge agent drives the same Brave
-  profile the activity extension instruments, so agent navigation emits `browser` rows (74 of
-  781 laptop `browser` buckets co-occur with a bridge command; it was the *sole* presence
+  profile the activity extension instruments, so agent navigation emits `browser` rows (76 of
+  777 laptop `browser` buckets co-occur with a bridge command; it was the *sole* presence
   source in 1). Adding a source a human drives means adding it there; anything an agent can
   drive, directly or indirectly, stays out. It used to be a denylist (everything except a
   timer), and on 2026-08-11 an unattended overnight agent run marked the night active and made
   `workbench/keys` + `workbench/tmux` read DEAD by morning (169-point/7-day sweep:
-  `workbench/tmux` 6 → 0, `workbench/keys` 14 → **7**, and those surviving 7 are a TRUE
+  `workbench/tmux` 7 → 0, `workbench/keys` 14 → **7**, and those surviving 7 are a TRUE
   positive — `keys` silent 2.8–8.1 active hours on 08-05 while `i3`/`tmux`/`zsh` kept emitting).
 - 🔴 **The trade is FEWER FALSE ALARMS, SLOWER TRUE ONES.** Budgets are counted in active
   buckets, so a slower-advancing timeline means a real death takes longer in wall time to
-  convict. Seeded-death latency, old → new: `workbench/keys` 6h → 12h, `workbench/tool`
-  36h → 96h, `laptop/zsh` 24h → 72h, `workbench/browser-bridge` 12h → 24h (which erodes the
-  #388 heartbeat). **No pair got faster; 10 of 17 got slower.** There is **no per-class rule** —
-  `laptop/tool` (on-demand) got *noisier* (26 → 28 dead-hours), as did `laptop/claude` and
-  `laptop/opencode`. Measure it, don't reason about it; the live table also drifts ±1–2
-  dead-hours between sweeps. Numbers in the module docstring's COST section.
+  convict. Seeded-death latency, old → new: `workbench/keys` 6h → 24h, `workbench/tool`
+  36h → 96h, `workbench/browser-bridge` 12h → 24h (which erodes the #388 heartbeat).
+  **No pair got faster in any run; 10 of 17 got slower.** There is **no per-class rule** —
+  `laptop/tool` (on-demand) got *noisier* (27 → 28 dead-hours). 🔴 Individual figures **drift
+  between sweeps** because the table advances; trust the direction and the count, never a
+  specific number. Numbers + the drift evidence in the module docstring's COST section.
 - **The budget is MEASURED per (host, source)**: `clamp(2 × p99 active-gap, 2h, 48h)`.
   Measured 2026-08-03 — `keys`/`i3`/`tmux` land on the 2h floor; `workbench/opencode` 11.5h;
   `workbench/tool` 31.1h. Nothing is hand-tuned and nothing is hand-listed.
@@ -213,11 +213,23 @@ Creds come from `~/.config/activity-collector/env` (the collector's own file) un
 - 🔴 **`newest_event_age_minutes` does NOT cover that** — it is a `max` across **all hosts and
   all sources**, so surviving agent rows pin it at ~1 min while a host's human timeline is days
   stale. The real detector is **`presence-stalled`** (`PRESENCE_STALL_HOURS`, 72): a host still
-  emitting `PRESENCE_STALL_HOURS` after its last human-driven row is reported CANNOT-TELL with
-  its pairs marked un-evaluated. Threshold sized off the worst real stall in 14 days (workbench
-  39.8h, laptop 8.9h, zero points over 48h). A host that is simply **switched off** does not
-  stall — the predicate compares the host's own newest row to its own newest presence row, not
-  to `now`.
+  emitting more than 72h after its last human-driven row is reported CANNOT-TELL with its
+  **remaining** pairs marked un-evaluated. Threshold sized off the worst real stall in 14 days
+  (workbench 39.8h, laptop 8.9h, zero points over 48h). Boundary is **exclusive** (exactly 72h
+  is not a stall).
+- 🔴 **A stall must not swallow a named death.** A pair already past its budget when the
+  timeline froze was measured against real active buckets — it stays `dead`, in `count`, and by
+  name in `detail`; only the pairs whose silence is 0-by-construction become un-evaluated. The
+  bar follows: an unknown state carrying `count > 0` renders **`tlm N` Critical with the toast
+  firing**, not `tlm ?`. A per-*host* discard here cost a measured `workbench/claude` death of
+  54.8 active hours against a 2.0h budget.
+- **Two things that deliberately do NOT stall:** a host **switched off** (both its timestamps
+  freeze, so the gap holds rather than growing — note a host powered off while *already* past
+  the threshold stays stalled for the rest of the window), and a host that has **never** emitted
+  a human-driven row (an agent pod, a container, a mis-set `ACTIVITY_HOST`). The latter is
+  `reason="no-presence-baseline"` on its pairs and raises nothing fleet-wide — merged into the
+  stall arm, one synthetic row made the whole fleet permanently cannot-tell and masked the
+  other hosts' real counts.
 - **Surface:** the workbench `bar-status-poll` (~45s) runs it as its `telemetry` source →
   `~/.cache/bar-status/telemetry.json` → the `tlm` pill (signal 17) + a rising-edge dunst
   toast. One workbench runner covers BOTH hosts because it reads the shared table. See the
--- a/claude/skills/activity/SKILL.md
+++ b/claude/skills/activity/SKILL.md
@@ -172,17 +172,25 @@ Creds come from `~/.config/activity-collector/env` (the collector's own file) un
   the ones where a **HUMAN-PRESENCE** source emitted; overnight/away time is simply not in the
   set. This is what makes a per-source budget work for `keys` (continuous) and `tool`
   (on-demand) at the same time.
-- 🔴 **Active time is an ALLOWLIST — `deadman.PRESENCE_SOURCES` = `keys` `i3` `tmux` `zsh`
-  `browser`.** Agent/machine sources (`claude`, `tool`, `opencode`, `browser-bridge`) do NOT
-  mark a bucket active. Adding a source that a human drives means adding it there; an
-  agent-driven one must stay out. It used to be a denylist (everything except a timer), and on
-  2026-08-11 an unattended overnight agent run marked the night active and made
-  `workbench/keys` + `workbench/tmux` read DEAD by morning (6/7 dead-hours in a 97-point sweep
-  → 0 under the allowlist). Budgets are counted in active buckets, so a shorter active timeline
-  moves budget *and* silence together and sensitivity shifts **both ways**: floor-pinned
-  sources get tighter (`laptop/opencode` 0 → 1 dead-hour), while an on-demand source's
-  detection can be *delayed* (a seeded `workbench/tool` death is caught at 72h by the old rule,
-  ~120h by the allowlist). Numbers in the module docstring's COST section.
+- 🔴 **Active time is an ALLOWLIST — `deadman.PRESENCE_SOURCES` = `keys` `i3` `tmux` `zsh`.**
+  Agent/machine sources (`claude`, `tool`, `opencode`, `browser-bridge`) do NOT mark a bucket
+  active, and neither does **`browser`**: the browser-bridge agent drives the same Brave
+  profile the activity extension instruments, so agent navigation emits `browser` rows (74 of
+  781 laptop `browser` buckets co-occur with a bridge command; it was the *sole* presence
+  source in 1). Adding a source a human drives means adding it there; anything an agent can
+  drive, directly or indirectly, stays out. It used to be a denylist (everything except a
+  timer), and on 2026-08-11 an unattended overnight agent run marked the night active and made
+  `workbench/keys` + `workbench/tmux` read DEAD by morning (169-point/7-day sweep:
+  `workbench/tmux` 6 → 0, `workbench/keys` 14 → **7**, and those surviving 7 are a TRUE
+  positive — `keys` silent 2.8–8.1 active hours on 08-05 while `i3`/`tmux`/`zsh` kept emitting).
+- 🔴 **The trade is FEWER FALSE ALARMS, SLOWER TRUE ONES.** Budgets are counted in active
+  buckets, so a slower-advancing timeline means a real death takes longer in wall time to
+  convict. Seeded-death latency, old → new: `workbench/keys` 6h → 12h, `workbench/tool`
+  36h → 96h, `laptop/zsh` 24h → 72h, `workbench/browser-bridge` 12h → 24h (which erodes the
+  #388 heartbeat). **No pair got faster; 10 of 17 got slower.** There is **no per-class rule** —
+  `laptop/tool` (on-demand) got *noisier* (26 → 28 dead-hours), as did `laptop/claude` and
+  `laptop/opencode`. Measure it, don't reason about it; the live table also drifts ±1–2
+  dead-hours between sweeps. Numbers in the module docstring's COST section.
 - **The budget is MEASURED per (host, source)**: `clamp(2 × p99 active-gap, 2h, 48h)`.
   Measured 2026-08-03 — `keys`/`i3`/`tmux` land on the 2h floor; `workbench/opencode` 11.5h;
   `workbench/tool` 31.1h. Nothing is hand-tuned and nothing is hand-listed.
@@ -190,15 +198,26 @@ Creds come from `~/.config/activity-collector/env` (the collector's own file) un
   14-day window, so `workbench/browser` (0 rows, correctly absent) can never alarm, while a
   source that *was* emitting and stopped keeps its baseline and does.
 - 🔴 **"Cannot tell" ≠ "healthy".** `not-configured` / `unreachable` / `query-failed` /
-  `no-data` are each their own state; `ok` is unreachable unless rows came back AND at least
-  one pair was actually measured (`evaluated > 0` is the verdict's own positive control).
+  `no-data` / `misconfigured` / `presence-stalled` are each their own state; `ok` is
+  unreachable unless rows came back AND at least one pair was actually measured
+  (`evaluated > 0` is the verdict's own positive control). All six render `tlm ?`, never a
+  green pill — `bar-status-poll` carries a hardcoded FALLBACK copy of that set, so **a new
+  state must be added there too**.
 - 🔴 **Blind spot, by design:** this is a RELATIVE check. A simultaneous outage of every
   source on every host produces zero active buckets and reads as "0 dead" — indistinguishable
-  from the operator being away. `newest_event_age_minutes` is reported for the human;
-  it is deliberately not an alarm. **The allowlist WIDENS this**: losing all five presence
-  sources at once (an X-session crash takes `keys`+`i3`, and the terminals take `tmux`+`zsh`)
-  is now enough to stop active time advancing, so the check goes quiet instead of alarming.
-  Losing only ONE presence source is still caught — the others carry the timeline.
+  from the operator being away. **The allowlist WIDENS this**: losing all four presence sources
+  at once (an X-session crash takes `keys`+`i3`, and the terminals take `tmux`+`zsh`) stops
+  active time advancing, and then *every* source on that host is unjudgeable, not just the
+  presence ones. Losing only ONE presence source is still caught — the others carry the
+  timeline.
+- 🔴 **`newest_event_age_minutes` does NOT cover that** — it is a `max` across **all hosts and
+  all sources**, so surviving agent rows pin it at ~1 min while a host's human timeline is days
+  stale. The real detector is **`presence-stalled`** (`PRESENCE_STALL_HOURS`, 72): a host still
+  emitting `PRESENCE_STALL_HOURS` after its last human-driven row is reported CANNOT-TELL with
+  its pairs marked un-evaluated. Threshold sized off the worst real stall in 14 days (workbench
+  39.8h, laptop 8.9h, zero points over 48h). A host that is simply **switched off** does not
+  stall — the predicate compares the host's own newest row to its own newest presence row, not
+  to `now`.
 - **Surface:** the workbench `bar-status-poll` (~45s) runs it as its `telemetry` source →
   `~/.cache/bar-status/telemetry.json` → the `tlm` pill (signal 17) + a rising-edge dunst
   toast. One workbench runner covers BOTH hosts because it reads the shared table. See the
--- a/claude/skills/activity/SKILL.md
+++ b/claude/skills/activity/SKILL.md
@@ -222,7 +222,7 @@ Creds come from `~/.config/activity-collector/env` (the collector's own file) un
   name in `detail`; only the pairs whose silence is 0-by-construction become un-evaluated. The
   bar follows: an unknown state carrying `count > 0` renders **`tlm N` Critical with the toast
   firing**, not `tlm ?`. A per-*host* discard here cost a measured `workbench/claude` death of
-  54.8 active hours against a 2.0h budget.
+  54.6 active hours against a 2.0h budget.
 - **Two things that deliberately do NOT stall:** a host **switched off** (both its timestamps
   freeze, so the gap holds rather than growing — note a host powered off while *already* past
   the threshold stays stalled for the rest of the window), and a host that has **never** emitted
--- a/claude/skills/activity/SKILL.md
+++ b/claude/skills/activity/SKILL.md
@@ -169,9 +169,20 @@ Creds come from `~/.config/activity-collector/env` (the collector's own file) un
 `CLICKHOUSE_URL/USER/PASSWORD` are already in the environment.
 
 - **Silence is counted in ACTIVE time, not wall time.** A host's *active* 5-min buckets are
-  the ones where any of its sources emitted; overnight/away time is simply not in the set.
-  This is what makes a per-source budget work for `keys` (continuous) and `tool` (on-demand)
-  at the same time.
+  the ones where a **HUMAN-PRESENCE** source emitted; overnight/away time is simply not in the
+  set. This is what makes a per-source budget work for `keys` (continuous) and `tool`
+  (on-demand) at the same time.
+- 🔴 **Active time is an ALLOWLIST — `deadman.PRESENCE_SOURCES` = `keys` `i3` `tmux` `zsh`
+  `browser`.** Agent/machine sources (`claude`, `tool`, `opencode`, `browser-bridge`) do NOT
+  mark a bucket active. Adding a source that a human drives means adding it there; an
+  agent-driven one must stay out. It used to be a denylist (everything except a timer), and on
+  2026-08-11 an unattended overnight agent run marked the night active and made
+  `workbench/keys` + `workbench/tmux` read DEAD by morning (6/7 dead-hours in a 97-point sweep
+  → 0 under the allowlist). Budgets are counted in active buckets, so a shorter active timeline
+  moves budget *and* silence together and sensitivity shifts **both ways**: floor-pinned
+  sources get tighter (`laptop/opencode` 0 → 1 dead-hour), while an on-demand source's
+  detection can be *delayed* (a seeded `workbench/tool` death is caught at 72h by the old rule,
+  ~120h by the allowlist). Numbers in the module docstring's COST section.
 - **The budget is MEASURED per (host, source)**: `clamp(2 × p99 active-gap, 2h, 48h)`.
   Measured 2026-08-03 — `keys`/`i3`/`tmux` land on the 2h floor; `workbench/opencode` 11.5h;
   `workbench/tool` 31.1h. Nothing is hand-tuned and nothing is hand-listed.
@@ -184,7 +195,10 @@ Creds come from `~/.config/activity-collector/env` (the collector's own file) un
 - 🔴 **Blind spot, by design:** this is a RELATIVE check. A simultaneous outage of every
   source on every host produces zero active buckets and reads as "0 dead" — indistinguishable
   from the operator being away. `newest_event_age_minutes` is reported for the human;
-  it is deliberately not an alarm.
+  it is deliberately not an alarm. **The allowlist WIDENS this**: losing all five presence
+  sources at once (an X-session crash takes `keys`+`i3`, and the terminals take `tmux`+`zsh`)
+  is now enough to stop active time advancing, so the check goes quiet instead of alarming.
+  Losing only ONE presence source is still caught — the others carry the timeline.
 - **Surface:** the workbench `bar-status-poll` (~45s) runs it as its `telemetry` source →
   `~/.cache/bar-status/telemetry.json` → the `tlm` pill (signal 17) + a rising-edge dunst
   toast. One workbench runner covers BOTH hosts because it reads the shared table. See the
--- a/claude/skills/bar/SKILL.md
+++ b/claude/skills/bar/SKILL.md
@@ -51,6 +51,12 @@ X — your call"). Leave the setup a little more modern than you found it; never
   (count crosses `--red-above`), latched via `~/.cache/bar-status/<src>.toast-state` so steady
   state stays silent. Thresholds are env-tunable and MUST match the block's `--red-above` (or
   they drift).
+  🔴 **A `--mock-*` run does NOT touch the latches or fire toasts unless you pass `--toast`.**
+  The latches live in the same cache dir as the status files, so a debug run with a
+  sub-threshold fixture would clear a *live* latch and make the next real poll (≤45s) re-toast
+  a steady-state condition. `telemetry` is the one source whose toast can ride an
+  **unknown** verdict: a `presence-stalled` host with a measured death renders `tlm N` and
+  toasts, not `tlm ?` — see `deadman.py`'s COST section.
 
 ## Block ↔ signal ↔ threshold map (workbench)
 | Block | Script | signal | `--red-above` | Notes |
@@ -109,7 +115,9 @@ for f in ~/.cache/bar-status/*.json; do echo "== $f"; cat "$f"; echo; done
 systemctl --user start bar-status-poll.service
 # run the poller by hand (see errors live)
 python3 ~/workspace/devrc/scripts/bar-status-poll
-# drive the write+signal path offline with fixtures (no live endpoint)
+# drive the write+signal path offline with fixtures (no live endpoint).
+# SAFE by default: writes the real cache files, but fires NO toasts and does not
+# touch the rising-edge latches. Add --toast only if you mean to exercise those.
 ~/workspace/devrc/scripts/bar-status-poll --mock-alerts alerts.json --mock-mail 3
 # unit tests for the pure parse/edge functions
 python3 -m pytest ~/workspace/devrc/scripts/tests/test_bar_status.py
--- a/scripts/agent-ops
+++ b/scripts/agent-ops
@@ -127,11 +127,19 @@ CLAWGATE_ENV = os.path.join(HOME, ".claude", "clawgate.env")
 # the Local-health telemetry-freshness query. Reuses the collector's env for the
 # endpoint/creds; never hardcodes a new one.
 CHQUERY_PATH = os.path.join(DEVRC_DIR, "scripts", "validation", "chquery.py")
-# The deadman owns the ONE definition of "which (source, kind) pairs are machine
-# generated rather than operator-driven" (MACHINE_CADENCE). This panel needs the
-# same predicate, so it IMPORTS it rather than re-spelling it — a second copy
-# would silently disagree the day a timer-driven emitter is added, and the
-# deadman's own 🔴 note tells maintainers about exactly one place to edit.
+# The deadman owns the ONE definition of "which (source, kind) pairs are
+# MACHINE-GENERATED" (MACHINE_CADENCE). This panel needs the same predicate, so
+# it IMPORTS it rather than re-spelling it — a second copy would silently
+# disagree the day a timer-driven emitter is added.
+#
+# 🔴 MACHINE_CADENCE is NOT deadman.PRESENCE_SOURCES, and this panel must never
+# be "unified" onto the latter. They answer different questions: PRESENCE_SOURCES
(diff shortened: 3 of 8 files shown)


ZacxDev:
## Audit fix round — all 7 findings addressed (`68566b7`)

Every number below is **re-measured against the live table with the code as it now stands**, not carried over. Two of my earlier claims were wrong and are corrected rather than patched.

---

### 🔴 1 — the deploy-blocker: the mitigation sentence was false, so I built the mitigation

The docstring claimed the surrendered alarm "only fires when the operator's whole desk is gone — which `newest_event_age_minutes` shows". It doesn't. `newest = max(r[2] for r in rows)` is global across **both hosts and all sources**, so surviving agent rows pin it at ~1 minute. And the loss is broader than I framed it: with the presence timeline frozen, **every** source on that host goes unjudged, not just the presence ones.

Reproduced against the real table (workbench presence blackout seeded 96h ago + `workbench/claude` genuinely silent 74h):

| | state | count | newest_age | `workbench/claude` |
|---|---|---:|---|---|
| OLD (denylist) | `ok` | **5** | 5 min | evaluated, **DEAD** |
| allowlist, no fix | `ok` | **0** | 5 min | evaluated, not-dead |
| allowlist, **fixed** | **`presence-stalled`** | 0 | 5 min | `evaluated=False reason=presence-stalled` |

New `STATE_PRESENCE_STALLED` (an `UNKNOWN_STATE`, so the bar renders `tlm ?`, never a green pill). Fires when a host is still emitting more than `PRESENCE_STALL_HOURS` after its last human-driven row; every pair on that host is marked un-evaluated with a reason, so the per-pair table can't print a row of confident `ok`s for a host nobody can judge.

**Threshold measured, not chosen.** Over the full live 14-day window, hourly, restricted to points where the host was still emitting: worst real presence stall was **39.8h** on the workbench (2 points over 24h, **zero over 48h**) and **8.9h** on the laptop. A 7-day re-sweep agreed (39.0h, zero over 48h). `72h` is 1.8× the worst normal — the same "clear the worst normal, don't cry wolf" sizing `FLOOR_BUCKETS` uses. It fires **zero times** on real data.

The predicate compares the host's **own** newest row to its **own** newest presence row, never to `now` — so a laptop that is simply shut does not alarm (that's the ordinary blackout, not a stall). Pinned by `test_a_host_that_is_simply_SWITCHED_OFF_does_not_stall`.

**Cost, stated:** the bar zeroes the count for any unknown state, so a stall on one host masks the other host's `tlm N` on the pill. `dead`/`count` stay in the payload and `describe_stalled` names both. Zero occurrences in 14 days makes that a rare-case trade, taken deliberately and commented as such.

### 🟡 2 — the three surviving mutants: `presence=` was decorative

`check(presence=X)` was a silent no-op. Now threaded through **all three** stages and pinned by `test_the_presence_parameter_is_wired_through_EVERY_stage`, which asserts the SQL changes, the parse changes, **and** the verdict changes (`h1/keys` DEAD under the default → `presence-stalled` under `{keys}` alone).

The `evaluate` forward is labelled honestly instead of being pretended into coverage: a mutant deleting it **survives every behavioural test**, because by then `parse_buckets` has already applied the allowlist and every row is 5-wide. It's forwarded anyway (three stages, one meaning) and pinned by a **structural** test that *measures* the equivalence and fails the day it stops holding. Both the test name and the `check()` docstring say "currently MOOT" out loud.

### 🟡 3 — the empty allowlist was not "quiet"

`0 AS n_presence` is valid ClickHouse and produced `state=ok evaluated=13 count=0` — a confident green with nothing judged, past `evaluated`'s own positive control. `presence_predicate_sql` now **raises**; `evaluate`/`check` return the new `STATE_MISCONFIGURED`; `check` never reaches the network.

### 🟡 4 — format ambiguity (the old format was also 5 columns)

Query is now `FORMAT TSVWithNames`. `parse_buckets` refuses a headerless 5-column file outright, names a stale `n_op` header by name, and still accepts the unambiguous 4-column legacy width. Verified live: the first line of the real response is `host\tsource\tb\tn\tn_presence` and 11,614 rows parse.

### 🟡 5 — `browser` dropped from the allowlist

Measured: **74 of 781** laptop `browser` buckets co-occur with a `browser-bridge` command bucket, and `browser` is the **sole** presence source in exactly **one**. Dropping it changed the 169-point sweep for **zero** pairs and *improved* one seeded-death latency. It bought nothing and carried the one path by which the fixed incident could recur. `PRESENCE_SOURCES = {keys, i3, tmux, zsh}`.

### 🟡 6 — the numbers did not generalize, and my per-class rule was wrong

**Dead-hours over 169 points / 7 days**, old → new:

| pair | OLD | NEW |
|---|---:|---:|
| workbench/tmux | 6 | 0 |
| workbench/keys | 14 | **7** |
| workbench/browser-bridge | 19 | 8 |
| laptop/browser-bridge | 105 | 92 |
| laptop/tmux | 3 | 3 |
| laptop/tool | 26 | **28** |
| laptop/claude | 49 | **50** |
| laptop/opencode | 0 | **1** |

`workbench/keys` does **not** go to 0. Its surviving 7 dead-hours are a **true positive**: 2026-08-05 15:00–21:00, silent 2.8–8.1 **active** hours against a 2.0h budget while `i3`/`tmux`/`zsh` on the same host kept emitting. The 7 hours the allowlist *removed* were the overnight ones. (The live table drifts ±1–2 dead-hours between sweeps; that's stated in the docstring rather than quoted as a constant.)

**Seeded-death latency** — smallest drop at which the pair reads DEAD, old → new. This is the headline the 4-day window hid:

`workbench/keys` 6h→12h · `workbench/i3` 6h→12h · `workbench/tmux` 6h→12h · `workbench/zsh` 12h→24h · `laptop/keys` 12h→24h · `laptop/i3` 12h→24h · `laptop/browser` 12h→24h · `laptop/zsh` 24h→72h · `workbench/tool` 36h→96h · **`workbench/browser-bridge` 12h→24h, which erodes the #388 heartbeat added one commit earlier**.

**No pair got faster. 10 of 17 got slower.** And there is **no per-class rule** — my previous claim ("floor-pinned tighten, on-demand loosen") was wrong in *both* directions: the floor is a fixed 24 **active** buckets, so it spans *more* wall hours once active time advances slower; and three on-demand/agent pairs got *noisier*. The honest summary, now in the docstring: **fewer false alarms, slower true ones.**

### 🟡 7 — agent-ops prose (no behaviour change)

Both comments now say plainly that `MACHINE_CADENCE` is **not** `PRESENCE_SOURCES`, that the freshness panel **does** count `claude`/`tool` as real usage, and that repointing it at `PRESENCE_SOURCES` would blank the freshness of every agent-driven source.

**Also:** `bar-status-poll`'s hardcoded fallback `UNKNOWN_STATES` copy gains both new states, plus a new seam test that pins the fallback against the real set and fails when either side grows *or* shrinks.

---

## Mutation battery — re-run in full, 30/30 killed, 0 survived

Harness controls first: no-op **SURVIVED**, canary **killed**. `__pycache__` purged per copy, `python -B` + `PYTHONDONTWRITEBYTECODE=1`, anchor uniqueness asserted.

| mutant | verdict |
|---|---|
| `PRESENCE_SOURCES` emptied / source removed / `claude` added / **`browser` put back** | killed (46 / 5 / 12 / 4) |
| `parse_buckets` `in` inverted · `evaluate` `in` inverted | killed (8 / 23) |
| `parse_buckets` reuses `n` · `evaluate` reuses `n` | killed (4 / 6) |
| SQL built from wrong set · `check()` sends a different set · filter dropped · rendering unsorted | killed (4 / 3 / 4 / 3) |
| **F2a–F2d** presence arg ignored in parse / evaluate / not forwarded to parse / not forwarded to evaluate | killed (2 / 1 / 1 / **1, see below**) |
| **F3a–F3b** empty allowlist renders `0` · evaluate drops the guard | killed (1 / 1) |
| **F4a–F4c** headerless `FORMAT TSV` · accept headerless 5-col · stop validating the header | killed (2 / 1 / 2) |
| **F1a** stall detection removed *(the audit blocker, restored)* | killed (6) |
| **F1b–F1h** threshold out of reach · threshold 0 · predicate uses `now` · no-presence host ignored · state downgraded to `ok` · stalled pairs still evaluated · state dropped from `UNKNOWN_STATES` | killed (5 / 3 / 4 / 1 / 5 / 2 / 2) |

**F2d survived the first run** — `check()` not forwarding `presence` to `evaluate`. It is behaviourally equivalent through `check()` (see finding 2), so I did not manufacture a fake behavioural test for it; I added a structural guard that measures the equivalence and documented the moot-ness in the code. Second run: 30/30 killed, only the intended no-op control survives.

## Gates — content read

```
  TOTAL collected=7218  passed=7217  skipped=1  failed=0   (floor: 7168)   RESULT: PASS

  TOTAL suites=3 files=30 tests=1024 pass=1024 fail=0 skipped=0  (floor: 970)   RESULT: PASS
```

Live sanity run of the shipped code against the real table: `state: ok rows=11614 evaluated=17 dead=0`, no stalled host, exit 0.

## Deliberately NOT changed

- **agent-ops behaviour.** Prose only, as instructed. `machine_cadence_sql_filter` and its seam tests are untouched.
- **The bar's unknown-state policy** (zeroing the count and rendering `tlm ?`). Changing it would let a stall coexist with `tlm N`, but that is `bar-status-poll`'s established contract for every "cannot tell" state and reworking it is out of scope here. The masking is commented in `evaluate` where the decision is made.
- **Per-host measured stall thresholds.** A stall would enter its own baseline, which is the self-defeating shape `MACHINE_CADENCE`'s note already warns about, so it is a measured constant like `FLOOR_BUCKETS`.
- **The `workbench/browser-bridge` latency erosion (12h→24h).** Real, reported, not compensated — fixing it means retuning the heartbeat interval or the floor, which is a separate change with its own measurement.

## NOT VERIFIED

- **Live behaviour after deploy.** Not shipped, not merged, no `home-manager switch`, no `ship.sh`. The `tlm` pill runs the pre-change deadman on both hosts until someone switches.
- **The presence-stall detector has never fired on real data** — by construction (zero occurrences in 14 days). Its live path is exercised only by seeded fixtures and one seeded replay against the real table; the 72h threshold's *false-positive* rate is measured, its *true-positive* rate on a real X-crash is not.
- **`STATE_MISCONFIGURED` end-to-end through the bar.** Unit-tested in both modules; never produced by a live run.
- **Anything beyond a 7-day sweep window / 14-day table.** The full table is fetched, but evaluation points were swept over 7 days.

ZacxDev:
## Delta-audit fix round (`cb050e4`) — 5 findings, 2 of them regressions I introduced

---

### 🔴 N1 — the stall detector made the check *quieter* in the outage it was written for

The per-host discard rested on a false premise stated in my own comment: *"silence is 0 for EVERY pair on a stalled host by construction."* That holds only for a pair that dies **after** the stall begins. A pair that died **before** it accrued its silence against real active buckets, and that measurement is genuine and complete.

Reproduced against the live table — workbench presence blackout 96h, `workbench/claude` dead 200h — at all three commits:

| commit | state | count | `workbench/claude` |
|---|---|---:|---|
| `9b8d440` (pre-detector) | `ok` | **1** | DEAD — **54.83** silent ACTIVE hours vs a 2.0h budget |
| `68566b7` (audited tip) | `presence-stalled` | **0** | `evaluated=False`, dropped from `dead`, absent from `detail` |
| **`cb050e4` (HEAD)** | `presence-stalled` | **1** | **DEAD, named in `detail`**, host still flagged cannot-tell |

Unseeded control at all three: `state=ok`, 0 dead — the fix adds no false positives.

**The discriminator is now per-PAIR.** A pair already past its budget when the timeline froze keeps its DEAD verdict; a pair still under budget becomes un-evaluated — because its silence can never grow while the timeline is frozen, *"not dead"* is the claim we cannot make and *"dead"* is the one we already made.

**The consumers were half the harm, so the bar follows.** `parse_telemetry` now renders an unknown state carrying `count > 0` as Critical **`tlm N`** instead of `tlm ?`, and sets `unknown_since=None` so the main loop stops suppressing the rising-edge dunst toast. Every other unknown state forces `count=0` upstream, so that branch is reachable only with a measured death. Both halves have tests, including a negative control proving the `tlm ?` path is untouched when nothing is dead.

🔴 **My previous test was structurally blind to this** — its fixture stopped `claude` 96h *after* the crash, so silence genuinely was 0 and the false premise was never exercised. Three new regression tests, each **confirmed RED at `68566b7`** and green at HEAD:

```
FAILED test_a_death_that_PREDATES_the_stall_is_still_reported_by_name
FAILED test_a_host_with_NO_presence_rows_is_skipped_not_alarmed
FAILED test_a_host_with_NO_presence_rows_ALONE_is_still_cannot_tell
3 failed, 73 passed
```

(`test_the_stall_threshold_boundary_is_EXCLUSIVE` passes at both commits — it is a mutation guard, labelled as one, not regression coverage.)

### 🔴 N2 — the `last_presence is None` arm was a permanently-red gate

No threshold, no minimum-evidence requirement. Any host emitting agent rows but never a presence row — agent pod, container, mis-set `ACTIVITY_HOST`, one manual `emit` — flipped the **whole fleet** to cannot-tell and zeroed the other hosts' real count for the full 14 days. Reproduced: one synthetic `("agentpod","claude")` row alongside a genuine `workbench/keys` death (66.5 silent active hours):

| commit | state | count | `agentpod/claude` |
|---|---|---:|---|
| `68566b7` | `presence-stalled` | 1 | `reason=presence-stalled` — fleet flagged, detail led with the agent pod |
| **HEAD** | **`ok`** | 1 | `reason=no-presence-baseline` — `workbench/keys` named in `detail` |

Split into two conditions, no longer merged:

- **had a timeline and lost it** → a *change*, which is what a deadman detects → host-level cannot-tell.
- **never had one** → no normal to compare against → follows the module's existing *"expected-present is MEASURED, not declared"* doctrine: pairs marked `reason="no-presence-baseline"` (never a silent `ok`), raises nothing fleet-wide.

When such a host is the **only** one, `evaluated` is 0 and the verdict is `no-data` — still an unknown state, never `ok`. Pinned by its own test.

### 🟢 N3 — the `>` / `>=` boundary

Confirmed reachable, not equivalent: 72h is an exact multiple of the 300s bucket. Boundary pinned **exclusive** in both directions (exactly `PRESENCE_STALL_HOURS` → `ok`; one bucket more → `presence-stalled`). The mutant now dies.

### 🟢 N4 — two of three "NOISIER" examples don't reproduce; dropped

Across **four independent sweeps** (three mine, one yours), `laptop/claude` measured 50→50, 49→50, 50→50 and **53→52 (quieter)**, and `laptop/opencode` was absent from three of the four. Both dropped. Only `laptop/tool` **27→28** reproduces, and the conclusion rests on it alone.

The same drift applies to the **latency** table — `workbench/keys` measured `6h→12h` in one run and `6h→24h` in another — so the docstring now says it outright: **trust the direction and the count, never a specific number.** The robust claim is unchanged: **no pair got faster in any run; 10 of 17 got slower.**

Current sweep (169 points / 7 days), old → new: `workbench/tmux` 7→0 · `workbench/keys` 14→7 · `workbench/browser-bridge` 18→8 · `laptop/browser-bridge` 106→91 · `laptop/tmux` 3→3 · `laptop/claude` 50→50 · `laptop/tool` 27→**28**.

### 🟢 N5 — the stated mechanism was wrong

"A switched-off host's two timestamps move together" is not what the data does. The mechanism is **freeze**: both stop advancing at power-off, so the gap holds whatever value it had. Measured — the workbench froze at a **38.75h** gap and held that exact value for the 4.6h it stayed down. The conclusion (an off host doesn't alarm) was right for the wrong reason. The honest consequence is now stated in the code: **a host powered off while already past the threshold stays `presence-stalled` for the rest of the 14-day window.**

---

## Mutation battery — re-run in full, 37/37 killed, 0 survived

Harness controls first: no-op **SURVIVED**, canary **killed**. `__pycache__` purged per copy, `python -B` + `PYTHONDONTWRITEBYTECODE=1`, anchor uniqueness asserted.

| mutant | verdict |
|---|---|
| allowlist: emptied · tmux removed · `claude` added · **`browser` put back** | killed (49 / 5 / 14 / 4) |
| membership: parse inverted · evaluate inverted · parse reuses `n` · evaluate reuses `n` | killed (8 / 24 / 4 / 9) |
| SQL: wrong set · `check()` different set · filter dropped · unsorted rendering | killed (4 / 3 / 4 / 3) |
| **F2a–F2d** presence arg ignored / not forwarded (×4) | killed (2 / 1 / 1 / 1) |
| **F3a–F3b** empty allowlist renders `0` · evaluate drops the guard | killed (1 / 1) |
| **F4a–F4c** headerless `FORMAT TSV` · accept headerless 5-col · stop validating header | killed (2 / 1 / 2) |
| **F1a–F1h** stall detection removed · threshold out of reach · threshold 0 · predicate uses `now` · state downgraded to `ok` · state dropped from `UNKNOWN_STATES` | killed (7 / 6 / 3 / 4 / 5 / 2) |
| **N3** boundary `>` → `>=` | killed (1) |
| **N1a** per-pair guard removed · **N1b** back to per-HOST discard · **N1c** discriminator inverted · **N1d** marking moved before the budget is computed · **N1e** death dropped from `dead` · **N1f** `describe_stalled` stops naming it | killed (3 / 1 / 3 / 1 / 1 / 2) |
| **N2a** no-presence host raises fleet-wide again · **N2b** its pairs marked evaluated | killed (2 / 2) |

`F2d` (which survived last round) is now killed by the structural forwarding guard added then.

## Gates — content read

```
  TOTAL collected=7223  passed=7222  skipped=1  failed=0   (floor: 7168)   RESULT: PASS

  TOTAL suites=3 files=30 tests=1024 pass=1024 fail=0 skipped=0  (floor: 970)   RESULT: PASS
```

Live sanity run of the shipped code: `state: ok rows=11599 evaluated=17 dead=0`, no stalled host, exit 0.

## Deliberately NOT changed

- **agent-ops** — untouched this round; the prose fix from the previous round stands and its seam tests are green.
- **The `presence-stalled` state itself is still raised when a death is present.** The host genuinely is in a cannot-tell condition, and the payload now carries both facts; only the *pill* leads with the death. Demoting the state to `ok` whenever a death exists would hide the stall.
- **The powered-off-past-threshold persistence (N5).** Stated, not "fixed": the host really was unjudgeable when it went down and still is. Adding a `now`-based expiry would reintroduce exactly the alarm-every-weekend behaviour the predicate avoids.
- **`PRESENCE_STALL_HOURS` retuning.** The measured worst normal stall moved 39.8h → 39.6h between sweeps; 72h still has ~1.8× headroom, so no change.

## NOT VERIFIED

- **Live behaviour after deploy.** Not shipped, not merged, no `home-manager switch`, no `ship.sh`.
- **The presence-stall path has never fired on real, unseeded data** — by construction (zero occurrences in 14 days). Every observation of it is a seeded replay or a fixture. Its false-positive rate is measured; its true-positive rate on a real X-crash is not.
- **The new bar branch end-to-end.** `parse_telemetry` → block → toast is unit-tested with an injected fire/latch; no real dunst notification was produced, and no poller ran with a live stalled verdict.
- **Numbers beyond a 7-day sweep window.** The full 14-day table is fetched; evaluation points are swept over 7 days.

ZacxDev:
## Round-4 fix round (`49e775d`) — A–H plus the pre-existing death boundary

Three of the four 🟡s were in the bar change I added last round, and one of those was a behavioural regression of my own.

---

### 🟡 A — the count>0 branch reset the cannot-tell grace clock

`unknown_since` was overloaded as **both** "when did this episode start" and "should the main loop skip the toast", and my previous fix resolved that conflict by destroying the clock.

**Measured through the REAL CLI path** — `main(["--mock-telemetry", …])` once per simulated poll, the resulting cache payload rendered through the actual `i3status-telemetry` block. One continuous `presence-stalled` episode, grace 1800s, polls every 900s:

| scenario | `cb050e4` | **HEAD** |
|---|---|---|
| count always 0 (control) | `tlm ?` at t=1800 | `tlm ?` at t=1800 |
| one count=1 poll at t=900 | **`tlm ?` NEVER** | **`tlm ?` at t=1800** |
| flapping 1/0/1/0 | **`tlm ?` NEVER** | **`tlm ?` at t=1800** (and t=3600) |

Per-poll pill text, the excursion case:

```
cb050e4   t=0 c=0 (empty)   t=900 c=1 tlm 1   t=1800 c=0 (empty)
HEAD      t=0 c=0 (empty)   t=900 c=1 tlm 1   t=1800 c=0 tlm ?
```

The baseline row is identical at both commits — the fix does not perturb the ordinary case.

**Fix:** two facts, two fields. `unknown_since` is now the grace clock and is set for **every** unknown verdict regardless of count; suppression is its own boolean `suppress_toast`, set only when an unknown verdict has nothing measured dead.

### 🟡 B — the seam my fix depended on had zero coverage

Three mutants survived the full suite, and the first silenced the telemetry toast in **every** scenario including a plain `ok` with dead sources.

The rule now lives in `toast_suppressed` + `dispatch_edge_toast`, and **both** the live poll and the mock path finish through one `_dispatch_all`. The source table is hoisted to `SOURCES` so a test can substitute fetchers and drive the **live** path. Coverage added:

- a five-quadrant parametrised test that observes **the gate** and **the toast** separately — a single observation cannot tell the mutants apart, because `ok`+count=0 legitimately reaches the gate and fires nothing;
- end-to-end through `main --mock-telemetry` (fires, and a negative control that does not and leaves the latch alone);
- end-to-end through the **live** path with substituted fetchers;
- a ledger test pinning `SOURCES` in both directions.

**Two further gaps found while closing this, both mine:**

1. My end-to-end test silently resolved `UNKNOWN_STATES` from the **base clone's** deadman via the default `DEVRC_DIR` — so `presence-stalled` was not an unknown state at all and the test passed down the plain-`ok` branch. `DEVRC_DIR` is now pinned at the checkout under test, with an assertion that fails loudly if the resolution slips again. Classic wrong-artifact verification, in my own harness.
2. Hoisting `SOURCES` made it deletable: a mutant dropping `telemetry` from it **survived** (the poller silently stops polling telemetry; the pill freezes on its last cache file forever). Pinned as a ledger against `SIGNALS` and the toast specs.

### 🟡 C — a malformed count raised where it used to degrade

`source()` turns the raise into a `stale` payload → the block renders **empty** → and the swallowed exception means the unit's `OnFailure` toast does not fire either. Junk input looked exactly like "all healthy". Added `_safe_int` / `_verdict_count`; both branches degrade to 0 and stay on their visible path, and `evaluated`/`rows` get the same treatment. Tested across `"abc"`, `{}`, `[]`, `None` and an arbitrary object, asserting the `tlm ?` pill is still reachable for the episode.

### 🟡 D — the deadman comment asserted the opposite of the code

It still described the bar masking a stalled host's count. Corrected and marked **CLOSED**, so a maintainer does not "re-fix" a trade that no longer exists.

### 🟢 E / F / G / H

- **E** — the retracted *"the two timestamps move together"* mechanism survived in the switched-off test's docstring; I had corrected 2 of 3 sites. Fixed, with the measured freeze value.
- **F** — removed `last_presence_by_host`, written and never read.
- **G** — SKILL.md `54.8` → `54.6` active hours, matching the three other sites.
- **H** — added a test for the powered-off-past-threshold consequence, explicitly labelled **CHARACTERISATION**: it passes at `cb050e4` too, and exists because the claim was newly *stated*, not because behaviour changed.

### Pre-existing, taken here (you flagged it as my call)

`"dead": silent > budget` had an **unpinned boundary** — the `>` → `>=` mutant survived at every commit on this branch, and it is not equivalent: it converts "has spent exactly its measured budget" into a death, which on a floor-pinned source is a routine two-hour lull. One test, both sides of the boundary. Taken because it guards the verdict the whole file is about.

---

## Red-at-base

HEAD's tests against `cb050e4`'s sources: **10 failures** — every A, B and C regression test. The two deadman-side additions (`H`, and the pre-existing death boundary) pass at base and are labelled as guards/characterisation, not counted as regression coverage.

## Mutation battery — 55 mutants across BOTH mutated files, 55 killed, 0 survived

Controls: two no-ops **survived**, two canaries **killed**. `__pycache__` purged per copy, `python -B` + `PYTHONDONTWRITEBYTECODE=1`, anchor uniqueness asserted.

| group | verdict |
|---|---|
| allowlist (emptied / tmux removed / `claude` added / `browser` back) | killed 51/5/15/4 |
| membership + SQL + presence-arg wiring (M4–M12, F2a–F2d) | all killed |
| empty-allowlist (F3a–F3b) · header format (F4a–F4c) | all killed |
| stall detector (F1a–F1h, N3) · per-pair (N1a–N1f) · no-presence split (N2a–N2b) | all killed |
| **P1** death boundary `>` → `>=` *(pre-existing)* | killed |
| **A1** clock cleared *(the regression)* · **A2** clock not carried forward · **A3** flag inverted · **A4** branch removed · **A5** `count > 0` → `>=` · **A6** grace `>=` → `>` | killed 2/3/5/3/7/3 |
| **B1** dispatch skips unconditionally · **B2** never skips · **B3** wrong field · **B4** loses the telemetry-only guard · **B5** live path stops dispatching · **B6** mock path stops dispatching · **B7** shared loop emptied · **B8** source dropped from `SOURCES` | killed 5/3/2/1/1/1/1/1 |
| **C1** `_verdict_count` raises again · **C2** `_safe_int` always default · **C3** propagates | killed 1/6/1 |

`B5` and `B8` each survived a first run and are the reason the dispatch was single-sourced and the ledger added — both reported above rather than quietly fixed.

## Gates — content read

```
  TOTAL collected=7235  passed=7234  skipped=1  failed=0   (floor: 7168)   RESULT: PASS

  TOTAL suites=3 files=30 tests=1024 pass=1024 fail=0 skipped=0  (floor: 970)   RESULT: PASS
```

Live sanity run of the deadman against the real table: `state: ok rows=11596 evaluated=17 dead=0`, exit 0.

## Deliberately left alone

- **`presence-stalled` still wins the STATE when a death is present.** The host genuinely is cannot-tell; only the *pill* leads with the death, and the payload carries both.
- **The powered-off-past-threshold persistence.** Stated and now tested, not changed — a `now`-based expiry reintroduces the alarm-every-weekend behaviour the predicate exists to avoid.
- **`--mock-*` for the other sources now dispatches toasts too.** That is a behaviour change I took deliberately (uniformity: a mock path that skips a decision the real path makes is a harness that cannot see that decision's bugs), not an accident.
- **`PRESENCE_STALL_HOURS`, the allowlist, and the sweep numbers** — untouched this round; no new measurement contradicted them.

## NOT VERIFIED

- **Live behaviour after deploy.** Not shipped, not merged, no `home-manager switch`, no `ship.sh`.
- **No real dunst notification was produced by any test** — `fire_toast` is injected or monkeypatched everywhere. The systemd-run/dunstify path itself is unexercised by this PR.
- **The presence-stall path has still never fired on real unseeded data** (zero occurrences in 14 days by construction); every observation is a seeded replay or a fixture.
- **The live poll under a real stalled verdict.** The live-path test substitutes fetchers; no poller has run against a genuinely stalled fleet.

ZacxDev:
## Round-5 fix round (`4598f40`) — N1–N5, and a correction to a claim I made

Both 🟡s flow from last round's one deliberate behaviour change (`--mock-*` dispatching toasts). One of them made the committed test suite spam the operator's desktop.

---

### 🟡 N1 — the suite fired 3 real desktop notifications per run

**Measured at the process boundary**, not in Python: a fake `systemd-run` / `dunstify` / `notify-send` placed first on `PATH`, each invocation appended to a log. (An in-Python assertion cannot prove this — the launch is a subprocess.) `DBUS_SESSION_BUS_ADDRESS` was set so `fire_toast` could not short-circuit before the launcher, or the zero would be meaningless.

| tree | launches |
|---|---:|
| `49e775d` | **3** |
| **`4598f40` (HEAD)** | **0** |

The three at `49e775d`, verbatim from the intercept log:

```
normal    clawgate: task awaiting      | 2 task(s) awaiting: #1, #2
low       New mail action              | 4 open mail action(s)
critical  🔴 Telemetry source DEAD     | laptop/keys silent 9h
```

The last is a **sticky critical false alarm naming a host and source that exist only in a fixture** — training the operator to dismiss the exact pill this series exists to make trustworthy. The 3 is also the positive control for the interceptor: it proves the shim can see launches, so the 0 means something.

**Fixed structurally, not per-test.** The launcher is hoisted to a module-level `_toast_runner`, and an **autouse fixture** patches it for the whole file — no test can reach the desktop regardless of future edits, while `fire_toast` itself stays real and directly tested.

🔴 **A mutant re-inlining the old lambda SURVIVED that fixture** — the patched seam was simply no longer on the path, and nothing noticed. That is a harness wired to nothing. Added the positive control: a test that drives `fire_toast` for real and asserts the recorder *saw* the launch.

**Correction:** my previous commit message claimed *"No real dunst notification was produced by any test."* That was **false for the committed tree**. Corrected here and in this comment.

### 🟡 N2 — the documented debug command moved the live latches

**Seam chosen: `--toast`, opt-in — not an isolated cache dir.** The whole point of `--mock-*` is that the *real* bar renders the fixture, so isolating the cache dir would break the documented purpose; the toast/latch side effect is the part nobody asked for. Dispatch stays reachable from a test (the end-to-end tests pass `--toast`), so the seam three mutants once hid in is still covered.

Regression test: the latch starts **latched**, a mock run with a sub-threshold fixture must leave it exactly as found, zero launches — plus an opt-in control proving `--toast` still dispatches and latches. Both doc sites updated (module docstring + `claude/skills/bar/SKILL.md`).

### 🟢 N3 / N4 / N5

- **N3** — `dispatch_edge_toast`'s "never raises" pinned; it is the sole fail-safe boundary for both paths and `evaluate_edge_toast` genuinely does raise.
- **N4** — dropped both inert branches (`int(True)` is 1 either way; the `isinstance(v, dict)` guard was unreachable), added `OverflowError` (`int(float('inf'))` raises *that*, and `json.load` accepts a bare `Infinity`, so `"count": Infinity` reached the parser through a supported entry point and reproduced the exact symptom 🟡 C was closed for), and clamped at 0 — a negative count was truthy, so the payload said "Critical" while the block renders only `count > 0`: **Critical with an invisible pill**. `parse_mail` already clamps.
- **N5** — the "10 failures" figure was measured mid-round and is wrong. Re-measured with `49e775d`'s tests against `cb050e4`'s sources: **12 node IDs across 8 test functions**. The matrix was genuinely red-at-base either way.

## Red-at-base for this round

Current tests against `49e775d` sources, run **under the process-boundary shim** so nothing could reach the desktop: **3 node IDs / 3 functions** — the junk-count edge cases, the end-to-end toast path, and the mock-run-does-not-move-the-latch guard. (That base run recorded 5 launches through the shim — another positive control.) The N3 and launcher-routing tests pass at base and are labelled **mutation guards**, not regression coverage.

## Mutation battery — 63 mutants across both files, 63 killed, 0 survived, 0 broken

Two no-op controls survived, two canaries died. `__pycache__` purged per copy, `python -B` + `PYTHONDONTWRITEBYTECODE=1`.

New this round: **N1a** launcher re-inlined (survived first, drove the routing test) · **N2a/b/c** mock dispatches unconditionally / `--toast` inverted / never dispatches · **N3a** `contextlib.suppress` deleted · **N4a** `OverflowError` uncaught · **N4b/c** clamp removed / inverted · **B6** retargeted at the new opt-in line. Everything from rounds 1–4 re-run unchanged.

## Gates — content read

```
  TOTAL collected=7238  passed=7237  skipped=1  failed=0   (floor: 7168)   RESULT: PASS

  TOTAL suites=3 files=30 tests=1024 pass=1024 fail=0 skipped=0  (floor: 970)   RESULT: PASS
```

## Found, NOT taken — pre-existing, out of scope

Widening the intercept to the whole `scripts/tests` + `scripts/collector/tests` run shows **9 real launches** from files this PR never touches (`git diff --name-only 3acd041..HEAD` confirms):

- `rig-control` → 5 × `notify-send` ("Chassis RGB on/off", "Sleep mode activated")
- `monitor-blackout` → 4 × `systemd-run --user --unit=monitor-blackout-restore-v2 --on-active=2h/8h` — the suite creates **real transient systemd timers on the host**

Both predate this branch and are unrelated to the bar/deadman work. Flagging for separate triage rather than growing this PR.

## Also deliberately left alone

- **`presence-stalled` still wins the STATE when a death is present** — only the pill leads with the death.
- **The powered-off-past-threshold persistence** — stated and tested, not changed.
- **`PRESENCE_STALL_HOURS`, the allowlist, the sweep numbers** — untouched; nothing this round contradicted them.

## NOT VERIFIED

- **Live behaviour after deploy.** Not shipped, not merged, no `home-manager switch`, no `ship.sh`.
- **The real `systemd-run`/`dunstify` path.** By construction it is now unreachable from the suite; it has never been exercised end-to-end by this PR.
- **The presence-stall path on real unseeded data** (zero occurrences in 14 days); every observation is a seeded replay or a fixture.
- **`--toast` against a live desktop.** Only exercised with an injected launcher.

PR #394, opened 2026-08-11
feat(session-manager): a cross-host tmux view where a zero always says which zero it is

ZacxDev:
Implements `scripts/session-manager` per `claudedocs/kickoff-session-manager.md` §1 —
**Phases 1, 2, `tail` only, and 4**. `signal`/`kill`, the agent-ops refactor (Phase 5) and
the bar pill (Phase 6) are **not** built; a test pins that `signal`/`kill` are not even
parseable subcommands, so a later edit cannot quietly add a window-killing verb to a
read-only tool.

## What it does

```
$ python3 scripts/session-manager --json
```
returns both hosts' tmux windows with the ClickHouse correlation attached. Live, just now:

| | |
|---|---|
| workbench | reachable, **45 windows** |
| laptop (`zach@10.42.0.100`) | reachable over SSH, **18 windows** |
| ClickHouse | `status: ok`, **20 rows** |
| fuzzyclaw | 400 files seen, **44 live**, 0 unparseable |
| **pane → CH join** | 36 windows carry a `claude_session_id`; **14 of the 20 CH sessions join back to a live tmux window** |

Table mode, `list`, `detail <s>:<w>` and `tail <s>:<w>` all exercised live (read-only).

## 🔴 Safety

Every test is hermetic. Two **autouse** fixtures replace `_default_runner` (the script's
only subprocess seam) and `make_ch_client` with functions that **raise**, so a test that
forgets to inject a runner fails loudly instead of running `tmux`.
`test_hermeticity_fixture_is_actually_installed` is the positive control on both — a guard
nobody has watched work is not a guard. The mutation sweep runs against a **copy** in a
per-agent scratch root; the real worktree is never mutated. No test creates, modifies,
signals or kills a window, and no `-f` pattern reaches `pkill`/`pgrep` anywhere.

Manual probing of real tmux was read-only (`list-panes`, `list-windows`, `capture-pane -p`).

## Red → green matrix

19 mutations, applied one at a time to a pristine copy, `pytest` run on each, restored.
**Baseline 157 passed / 0 failed; after the sweep, restored to 157 / 0.** Every mutation was
killed. Counts and the specific failing assertion:

| # | What I broke | Went red | Specific failure |
|---|---|---|---|
| **M1** | **delete the fuzzyclaw live-window intersection** | **6** | `test_task_file_pointing_at_a_DEAD_window_is_excluded` — `AssertionError: assert ['@41', '@997'] == ['@41']` (names the stale window; this guard's failure, not another's) |
| M2 | laptop → `zach@10.42.0.10` (the homelab gateway) | 5 | `test_laptop_ssh_target_is_the_laptop_not_the_homelab_gateway` — `assert 'zach@10.42.0.10' == 'zach@10.42.0.100'` |
| M3 | collapse a failed CH query into an empty result | 6 | `test_ch_unreachable_is_NOT_an_empty_result`; `test_ch_query_error_code_47_surfaces_with_its_code`; table-mode `QUERY FAILED [unreachable]` not found |
| M4 | build the CH client despite `--no-ch` | 1 | `test_no_ch_never_constructs_the_client` — `AssertionError: a CH client was built despite --no-ch; assert [1] == []` |
| M5 | stale boundary `>=` → `>` | 3 | `test_stale_threshold_at_the_boundary_and_well_either_side[600]` **and** `[3600]` — both threshold points |
| M6 | drop `transcript_path` from the field ledger | 1 | `test_field_ledger_is_exactly_this_set` (shrinkage) |
| M7 | consume a key not on the ledger | 4 | `test_field_ledger_matches_what_the_code_ACTUALLY_READS` (growth, caught behaviourally by a key-tracking mapping) |
| M8 | `EXIT_EMPTY` = `EXIT_UNAVAILABLE` | 2 | `test_exit_codes_are_three_distinct_values`; `test_the_two_zeroes_produce_DIFFERENT_exit_codes` |
| M9 | `split("|", 7)` → `split("|")` | 1 | `test_parse_panes_title_containing_pipes_is_not_truncated` |
| M10 | report a failed host as reachable-with-zero-windows | 8 | `test_unreachable_host_is_VISIBLE_in_both_output_modes_and_exits_zero`; `test_table_says_NOTHING_WAS_MEASURED_when_no_host_answered` |
| M11 | treat "no server running" as unreachable | 1 | `test_tmux_not_running_is_REACHABLE_with_zero_windows` |
| M12 | let an unparseable task file crash the scan | 2 | `test_unparseable_or_wrong_shaped_task_file_is_skipped_not_fatal[]` |
| **M13** | **negative control — stub `gather()` to `return {}`** | **30** | `KeyError: 'hosts'`, `KeyError: 'fuzzyclaw'`, `assert 4 == 0` on `exit_code_for({})`; **127 still passed** (the pure-parsing tests, correctly independent of `gather`) |
| M14 | render a failed CH query as an empty table section | 2 | `test_table_mode_distinguishes_the_three_ch_outcomes[unreachable]`, `[query_error]` |
| M15 | loosen `_SLOT_RE` so malformed slot lines parse | 2 | `test_malformed_slot_line_yields_no_codenames_and_never_raises` — `assert {'scratch7': 'Grove'} == {}` |
| M16 | accept any `tail` target | 10 | `test_tail_rejects_a_target_that_is_not_session_colon_window[$(id):1]`, `[a:b:c]`, … |
| M17 | join local fuzzyclaw onto remote windows | 1 | `test_remote_fuzzyclaw_is_not_fabricated` |
| M18 | stop folding panes into windows | 33 | `test_json_golden_schema_and_values` + the whole gather surface |
| M19 | bind the CH factory as a def-time default | 1 | `test_the_default_ch_factory_is_resolved_at_CALL_time_not_bound_as_a_default` |

### M1 in full — the guard the brief called mandatory

Deleting the intersection produces, verbatim:

```
scripts/tests/test_session_manager.py:345: in test_task_file_pointing_at_a_DEAD_window_is_excluded
    assert [t["window_id"] for t in res["tasks"]] == ["@41"]
E   AssertionError: assert ['@41', '@997'] == ['@41']
E     Left contains one more item: '@997'
```

**Reachability** is pinned in-suite, not assumed:
`test_the_intersection_is_REACHABLE_and_is_the_only_thing_excluding_it` feeds the
byte-identical stale fixture twice, changing **only** whether its `window_id` is in the live
set — excluded in one run, included in the other. So nothing upstream rejects it and the
intersection is the sole cause of the exclusion. A mutation test whose guard never executes
is green for the wrong reason; this rules that out.

### M19 — a defect the sweep found in this PR's own code

`gather(..., ch_client_factory=make_ch_client)` captured the **original** function at
def time, so `monkeypatch.setattr(sm, "make_ch_client", …)` was **not** honoured on
`main()` → `gather()` — the one path with no injection point. The suite's "no real network"
guard was silently bypassed exactly where it mattered, and every other test still passed.
Fixed by resolving the factory at call time, plus a test that pins it (M19 kills that test
and only that test). Per the audit-fix rule, **the whole 19-mutation sweep and both test
tiers were re-run after this fix**, not just the delta.

## Negative control on the suite

M13 above. Stubbing `gather()` to `return {}` turns **30 tests red** while **127 stay
green** — the pure parsing/classification tests, which correctly do not depend on `gather`.
The suite has been watched to fail, and to fail selectively.

The mutation harness itself was validated in both directions: it reported `BASELINE: 157
passed, 0 failed` before the sweep and `RESTORED: 157 passed, 0 failed` after, so a "0
failed" reading from it is a measurement rather than a wiring accident.

## Test counts and the floor

`MIN_TESTS` **7168 → 7351**. Measured, not computed, on the tree this PR ships.
CONTROL PAIR, one tool set, one machine, read off the runner's structured summary
**content** (never an exit code):

| | TOTAL collected | passed | skipped | failed | `scripts/tests` |
|---|---|---|---|---|---|
| base `3acd041` (origin/main) | 7194 | 7193 | 1 | 0 | 1923 |
| this branch | **7351** | 7350 | 1 | 0 | **2080** |

+157, and 2080 − 1923 = 157 = exactly `test_session_manager.py`. That per-directory
attribution is also the **positive control** proving the gate runs the new suite: no
`HERMETIC_TARGETS` entry was needed (`scripts/tests` is already a directory target), so
movement on that directory's own line is what demonstrates execution rather than asserting
it.

⚠ The base was **already 26 above the old floor** (7194 vs 7168) before this branch existed
— pre-existing slack, re-measured rather than carried forward.

## Authoritative gate — counts, both tiers

**`nix build .#checks.x86_64-linux.pytests`** (authoritative):
```
PASS  scripts/tests  (collected=2080 passed=2080 skipped=0)
TOTAL collected=7351  passed=7350  skipped=1  failed=0  (floor: 7351)
RESULT: PASS
```

**Dev-host tier** `run-tests.sh --set all .`:
```
TOTAL collected=7351  passed=7350  skipped=1  failed=0  (floor: 7351)
RESULT: PASS
```

Both tiers agree exactly. The one pinned skip (`repo-cos` live-store drift) is unchanged;
this suite adds **zero** skips by construction.

⚠ **The dev-host tier is only meaningful with the flake's own python package set.** Under an
ad-hoc `nix-shell -p python3Packages.pytest` it collects 5138 and fails — `mail-actions`
collects 0 on a missing `psycopg2`, `initiatives` throws 9 `ModuleNotFoundError`s. And
adding `-p opencode` shadows the host's pinned `1.18.4` with nixpkgs' `1.18.13`, turning
`test_opencode_engine.py` red for a reason with nothing to do with this tree. Both were
diagnosed by control rather than by reasoning (`command -v opencode` → the host's real
1.18.4) and the reproduction line is recorded in the `run-tests.sh` comment.

## What I could NOT verify — stated plainly

- **Exit code 4 (`EXIT_UNAVAILABLE`) has only been exercised hermetically.** I did not break
  SSH or kill a tmux server on the live hosts to reproduce it. Exit **2** (usage) and **3**
  (ran, found nothing) *were* verified live. The unreachable-laptop path is covered by four
  hermetic tests plus M10, but a real network partition is untested.
- **The "no server running" branch is hermetic-only.** No live host had tmux stopped.
- **`CHUnreachable` / `CHQueryError` were injected, never provoked** — ClickHouse answered
  `status: ok` on every live run. The `Code: 47` fixture is shaped from the real error the
  spec documents, not captured from a live failure.
- **Not deployed.** Merged ≠ deployed: `claude/skills/session-manager/` only reaches
  `~/.claude/skills/` on a `home-manager switch` / `ship.sh`. **The skill has therefore
  never been observed to trigger** — I verified the file and its front-matter, not the
  routing. The script itself runs from the repo checkout and needs no switch.
- **Claude detection is shallower than agent-ops'** (`pane_current_command =~ /claude/`, no
  `/proc` walk). On the live scan it found 43 Claude windows across both hosts, but it will
  miss a `claude` running under a wrapper shell. This is a documented limitation, not a bug
  I have measured the cost of.
- **Cross-host coverage is asymmetric and says so**: fuzzyclaw is local-only, so remote rows
  carry `null` task / session-id / age and are never labelled `stale`. Tested that this is
  *not* fabricated (M17); not "fixed", because fixing it means reading remote files.
- **`tail --stream` is not implemented.** One-shot `capture-pane` only.
- The mutation sweep is a claim about the **19 mutations I imagined**. A differently
  constructed sweep may find blind spots this one has.


ZacxDev:
## Correction + audit fixes (`9d61558`)

### 🔴 Correction to the PR body above — it says `44 live`, and that is wrong

The table in the description reads **"fuzzyclaw | 400 files seen, **44 live**, 0 unparseable"**.
**44 is the live *window* count, not the live *task-file* count.** The code, `SKILL.md`, the
commit message and every live run said **43** (400 − 357). Correcting this in a comment rather
than by editing the body, because a reviewer may already have read the wrong number.

And the corrected number is now **stale in the other direction**: the fix below rejects 11 more
files, so the current live measurement is **400 seen, 357 stale, 11 slot-mismatched, 32 live,
0 unparseable**. The 43 was never a safe number — that is the whole point of F1.

---

Four confirmed defects fixed, plus four false claims. Everything below is measured on this host.

### F1 — the load-bearing guard validated a different key than the join used

`filter_live_tasks` keyed the staleness guard on `window_id`; `index_tasks_by_window` keyed the
pane↔task join on `(tmux_session, window_index)` and did silent last-wins. The docstring called
that split *"deliberate"*. It was a defect. `renumber-windows` is `on` here, so indexes shift
under live windows all day: a file could pass the guard on a live id and then attach to whatever
window had since taken its slot.

Measured with the old guard — of the 43 survivors, **only 32** still sat in the slot they
recorded, **7** named a slot held by a *different* live window, and **5** slots had multiple
claimants. Two rendered rows carried another window's `claude_session_id`, which
`reference/clickhouse-queries.md` calls the only carrier of the session id into ClickHouse — so
`detail` would have pulled a stranger's prompt history and rendered it as this window's.

The guard now pins the **relationship**. `list-windows` reports
`#{window_id}|#{window_index}|#{session_name}`, and a task survives only when:

```
window_id is live   AND   live_windows[window_id] == (session, index)
```

- contested slots resolve to **nothing** — `index_tasks_by_window` drops them and reports them
  under `fuzzyclaw.slot_conflicts`. Attaching an arbitrary one of two contradictory records is
  worse than attaching none, because it reads as measured data.
- rejections are counted **separately**: `files_stale` (window gone) vs `files_mismatched`
  (alive, moved). Collapsing them hides a renumber storm.
- every row carries `window_id`, so the join is auditable in the output —
  `row.window_id == row.fuzzyclaw.window_id` for every joined row, asserted end-to-end.
- a bare set of ids is now a **`TypeError`**, not a silent downgrade to the old check.

Live, after the fix: **32 live, 357 stale, 11 slot-mismatched, 0 rows disagreeing** — 43 − 32 = 11
lines up exactly with the audit's prediction.

The docstring no longer asserts soundness the data contradicts; it describes what the code
guarantees and records what it used to get wrong.

### F2 / F3 — two more fabricated zeroes, in a PR titled *"a zero always says which zero it is"*

- **F2**: `local_live_windows = set()` was assigned only inside `if host == local_host`. With
  `--host laptop` the local host never enters the loop, so all 400 files were filtered against an
  empty set and published `files_seen: 400, files_live: 0, status: "ok"`.
- **F3**: `wins_res["reachable"]` and `wins_res["error"]` were **never read** — only its stdout.
  So `list-panes` succeeding while `list-windows` failed dropped every task, published
  `live_window_ids` as a *measured* `[]`, nulled `claude_session_id`, kept `status: "ok"` and
  exited 0.

The live-window set is now **measured-or-`None`**. Unmeasured produces `status: "unmeasured"`,
`files_live: null` (never `0`), and a stated reason. Per host, `windows_measured` /
`windows_error` describe the **`list-windows`** call as a measurement independent of
`reachable` / `error`, which describe **`list-panes`**. Both are loud in the table
(`LIVE COUNT UNMEASURED`, `WINDOW LIST UNMEASURED ON: …`).

Verified live: `--host laptop` now returns `status: "unmeasured", files_live: null` with the
reason naming the unscanned local host.

### F4 — the fixture blind spot that hid F3

`make_runner` returned the same rc/stderr for every call on a host, so **no test could
distinguish the panes subprocess from the windows subprocess** — which is exactly why the
"reachability/error read off the wrong subprocess" mutants survived a green sweep. It now
answers per subcommand (`list-panes` / `list-windows` / `capture-pane`, local and remote), and
`test_make_runner_can_distinguish_the_two_subprocesses` is the positive control on the
instrument: same host, same runner, one call red and the other green, then the mirror image.

### The 🟡 items — each one a claim that was false

| | fix |
|---|---|
| `run-tests.sh` said the tiers *"DISAGREE BY ONE — deliberately recorded rather than smoothed over"* directly above **identical** numbers and *"matches the sandbox exactly"* | rewritten to say they agree, with the leftover recorded as the leftover it was |
| `sql_session_history` was **dead code** while its doc described it as implemented, and *"The join to tmux"* described a join the code did not perform | **wired up** — `detail` attaches `session_history`, status-discriminated with a `reason` on every skip (`--no-ch` / no session id / genuine zero are three different facts). Doc rewritten to describe both joins, including the one that was wrong |
| `tail` reported *"no such window"* as **"host unreachable" exit 4** — a false fact about a host that answered, sending the reader to debug SSH over a typo | third outcome: `reachable: true, found: false`, **exit 2**, distinct message. `found: null` when the host never answered |
| `tail` silently defaulted to local on `--host all` while `SKILL.md`'s example omits `--host` | still defaults to local, but **records it**: `host_defaulted` in the JSON, and the not-found message names the host searched and how to search the other |

### Mutation battery — 64 semantic mutants, re-run in full after the fixes

🔴 Run with the caching hazard actually closed: `PYTHONDONTWRITEBYTECODE=1`, `python3 -B`, a
**fresh `cp -a` tree per mutant**, and `__pycache__` wiped before each run. `session-manager` is
extensionless and loaded via `SourceFileLoader`, whose `.pyc` is keyed on `(int(mtime), size)` —
a sub-second mutate→run→restore loop collides on both. Mutants are semantic (operand swap, branch
inversion, comment-out, wrong bind, off-by-one, stale rebind), not just deletions. Mutation
happens only inside a per-agent scratch root; the real worktree is never written.

**Round 1 — 62 mutants, 2 survivors, and both were real gaps:**

| survivor | why it survived | fix |
|---|---|---|
| off-ledger key read behind an `or` | the field-ledger probe used `TASK_LIVE`, where **every value is truthy**, so a short-circuited extra read never executed | probe re-run with all-falsy and all-absent fixtures too |
| `WINDOW_FORMAT` reverted to `'#{window_id}'` | **nothing pinned the tmux `-F` contract** — every fixture feeds `parse_windows` a pre-rendered string. Against a real tmux that revert makes every line unparseable, so `live_window_ids` empties, every task file looks stale, and the tool reports a confident measured wrong zero | format pinned verbatim, plus a **structural** coupling test that renders a line *from* `WINDOW_FORMAT` and parses it back — so format and parser cannot drift alone |

**Round 2, after those fixes — 64 mutants, 64 killed, 0 survivors.** Includes the audit's C, D,
E, K, O, Q and V. Baseline 229 passed / 0 failed; restored 229 / 0, so a "0 failed" reading is a
measurement rather than a wiring accident. Negative control `NEG` (stub `gather()` → `{}`) turns
**53** red while the pure-parsing tests stay green.

### Gate — counts, both tiers

`nix build .#checks.x86_64-linux.pytests` (authoritative):

```
PASS  scripts/tests  (collected=2152 passed=2152 skipped=0)
TOTAL collected=7423  passed=7422  skipped=1  failed=0  (floor: 7423)
RESULT: PASS
```

Dev-host `run-tests.sh --set all .`: **identical on every line** — `collected=7423 passed=7422
skipped=1 failed=0`, `scripts/tests=2152`. `MIN_TESTS` **7351 → 7423**, measured on this tree,
not computed. Suite 157 → 229 tests (+72). The one pinned skip is unchanged; this work adds zero.

### What I could NOT verify — stated plainly

- **F3 is hermetic-only.** I did not break a real `tmux list-windows` while leaving `list-panes`
  working on a live host. F1 and F2 *were* reproduced live and read-only; F3 rests on the fixture
  plus mutants C/D/F3a–c.
- **`detail_history` has never run against real ClickHouse.** Every CH outcome was injected. The
  SQL is the same shape the reference doc already documented, but a live `Code:` failure was
  never provoked.
- **`slot_conflicts` renders untested against real data** — the live host currently reports zero
  conflicts, because the relationship check removes the duplicate claimants before indexing. The
  5 contested slots the audit found were an artefact of the *old* guard.
- **Not deployed.** `claude/skills/session-manager/` only reaches `~/.claude/skills/` on a
  `home-manager switch` / `ship.sh`, so the skill still has never been observed to trigger.
- The battery is a claim about the **64 mutations I imagined**. Round 1 proves a differently
  constructed sweep finds things — it found two.

Not merged.

PR #396, opened 2026-08-11
fix(guard): judge the tree the command acts on, and stop a heredoc body blinding the scanner

ZacxDev:
Two mechanism defects in the Bash guard, both found by replaying real session
transcripts. **No policy change** — the check list, `_MAIN_BRANCH_NAMES` and the
`_TRUNK_DEPLOY_REPOS` allowlist are byte-for-byte untouched.

## 1. Branch resolution read the ambient repo

The guard could not evaluate a shell variable, so `git -C "$WT" commit` — the
spelling this repo's and datapacket-talos's CLAUDE.md both **mandate** — resolved
to nothing and fell back to the caller's cwd, denying a commit into a
topic-branch worktree by naming a primary clone the command never touches. For a
bare commit it judged the **union** of the cwd and every `cd` target in the line,
which denied the subshell and `bash -c` worktree spellings the same way.

The union was wrong in **both** directions. Measured against the shipped core,
cwd = a clone on `trunk`:

| command | shipped | this PR |
|---|---|---|
| `git -C "$WT" commit -F <file>` (worktree on a topic branch) | DENY | ALLOW |
| `. <env-file>; git -C "$WT" commit …` | DENY | ALLOW |
| `(cd <worktree-on-a-topic-branch> && git commit …)` | DENY | ALLOW |
| `bash -c 'cd <worktree-on-a-topic-branch> && git commit …'` | DENY | ALLOW |
| `REPO=<repo-on-trunk>; git -C "$REPO" commit -m x` (cwd innocent) | **ALLOW** | DENY |

That last row is the one to read twice: an unresolvable `-C` fell back to a
directory that happened to be fine, so the same defect that produced the false
positive also produced a **miss**.

- `_command_vars` evaluates `NAME=<literal>` assignments in the command text and
  reads a `. <file>` / `source <file>` the command names. The house recipe is
  `WT=/tmp/wt-$$`, whose value depends on the shell's pid and can never be read
  from the text, so the observed pattern is to write the expanded path into a
  scratchpad env file and source it later. A name assigned twice to different
  literals, or to anything containing `$`, is dropped rather than guessed.
  Assignments are read from the real command line only, **never** from a heredoc
  body — otherwise a commit *message* would get to choose which repo is judged.
- `_commands_with_cwd` pairs every argv with the directory the shell is standing
  in when it runs, tracking `cd`/`pushd` positionally with subshell scope. A `cd`
  that does not resolve leaves the cwd in play; a `cd` inside `( … )` does not
  leak out of it.
- An unresolvable `-C` still falls back to the cwd, and now says so in the deny.

## 2. One apostrophe in a heredoc body disabled every argv check

The scanner tracked quote state through heredoc bodies, which are prose. A lone
apostrophe (`don't`, `it's`) opened a quote that never closed, so every command
**after** the heredoc was swallowed into the quoted buffer and never parsed.
Measured on the shipped core — all ALLOW, all DENY with the apostrophe removed:

```
cat > /tmp/m <<EOF
fix: don't blindly stage
EOF
<victim>
```

| victim | shipped | this PR |
|---|---|---|
| `git commit -m x` | ALLOW | DENY |
| `git stash push -m wip` | ALLOW | DENY |
| `git -C /tmp reset --hard origin/main` | ALLOW | DENY |
| `git clean -fd` | ALLOW | DENY |
| `talosctl -n 1.2.3.4 reset` | ALLOW | DENY |
| `mkfs.ext4 /dev/sdc` | ALLOW | DENY |
| `dd if=/dev/zero of=/dev/sdc` | ALLOW | DENY |
| `pkill -f e2e/run.sh` | ALLOW | DENY |

A silent bypass of seven check families, several of them irreversible-action,
behind one character of ordinary English.

- `_scan` recognises the heredoc operator (`<<TAG`, `<<-TAG`, quoted tags; `<<<`
  is a here-**string** and is not one) and lifts the body out of the surrounding
  quote state.
- An unterminated quote from **any** source re-scans the tail with the offending
  quote treated as an ordinary character. Found by this change's own
  non-regression test: lifting the heredoc out was not enough for
  `bash <<EOF / it's fine / git stash push -m wip / EOF`, where the corruption
  then happened inside the body's *own* parse.

🔴 This is **not** the reverted `_strip_message_text()` helper the module
docstring forbids. Nothing is blanked and nothing is declared inert: bodies are
still returned, still parsed as commands and still checked, so `bash <<EOF` keeps
every deny it had (asserted). The change can only make **more** argv visible,
never fewer.

## Verification

**30 tests RED at the pre-change core, GREEN at HEAD**, measured in one tree with
only `guard_core.py` swapped. (A separate temp-dir baseline reported 34 — the
extra four were missing-sibling-file artefacts, not reds. Worth stating: the
first baseline I built was itself wrong.)

Suites, all green at HEAD: `test_guard_core.py` **1354 passed**,
`test_bash_guard.py` `RESULT: all good`, `test_opencode_guard_plugin.py` +
`test_opencode_engine.py` **42 passed**.

Negative controls, verified DENY **both** before and after — these are invariant
guards, not regression coverage, and are labelled as such: a commit in a clone on
`main`/`master`/`trunk`; the same with a heredoc message body, with and without
an apostrophe; `git -C <repo-on-trunk> commit` from a topic-branch cwd in seven
spellings; a variable resolving to a repo on trunk; a variable assigned twice; an
unresolvable `cd`; a heredoc body attempting to inject a variable.

Hot path: variable resolution can read a file, so it sits behind the same "is
there a real `git … commit` here at all?" gate the git subprocesses do — pinned
as a count of `open()` calls **with a positive control**, so a zero cannot be a
probe wired to nothing. Typical command 0.235 ms → 0.249 ms; a 400-line heredoc
41.3 ms → 45.7 ms; both dominated by pre-existing `shlex` cost.

## One deliberate behaviour change

`test_the_cwd_REMAINS_a_candidate_even_when_the_line_cds_elsewhere` is replaced:
the cwd is no longer a candidate once the line has moved somewhere else. The
**policy** outcome for `cd <path> && git …` is unchanged — `check_cd_then_git`
still denies it end-to-end — which the replacement test asserts, so "this
loosened the guard" and "this moved which check reports it" cannot be confused.

## A retraction

The heredoc defect was reported as *"`git -C "$WT" commit -q -F - <<'MSG'` was
ALLOWED twice, while `-F <file>` is correctly blocked."* Replaying the exact
command text of all three calls says otherwise: **the shipped core DENIES all
three**, including the two that were allowed in-session. The discriminator was
time, not `-F -` vs `-F <file>` — `check_git_commit_to_main` first shipped in
home-manager generation **415** (2026-08-11 04:41 local), and the two "allowed"
commits ran at 22:21 and 00:41 local, hours before it existed; the denied one ran
at 10:17, after. The A/B differed in deployment, not in parsing.

The heredoc family was a real hole anyway — just not that one, and worse than the
report described, since it reaches every argv check rather than commit-to-main
alone.

## 🔴 Not live yet

`~/.claude/hooks/guard_core.py` resolves to `/nix/store/…-hm_guard_core.py`, a
home-manager `home.file` **copy** — not an out-of-store symlink. Merging this
changes nothing on either host until someone runs `home-manager switch`. None was
run as part of this work.

### DIFF
--- a/scripts/claude-hooks/guard_core.py
+++ b/scripts/claude-hooks/guard_core.py
@@ -77,6 +77,17 @@ one extra step -- write the message to a file and use `git commit -F <file>` --
 which RULES already prefers over heredocs. That is a better trade than a
 security guard with a rotating cast of bypasses.
 
+🔴 NEITHER IS THE HEREDOC HANDLING IN `_scan` (2026-08-11). It LIFTS a heredoc
+body out of the surrounding quote state and then parses it as commands exactly
+as before — nothing is blanked, nothing is declared inert, and `bash <<EOF` keeps
+every deny it had. It exists because the opposite of the reverted helper was also
+a bug: the scanner tracked quote state THROUGH the body, so one apostrophe in a
+commit message (`don't`) opened a quote that never closed and swallowed every
+command AFTER the heredoc. Measured on the shipped guard — `git commit`,
+`git stash`, `git clean -f`, `talosctl reset`, `mkfs`, `dd of=/dev/sdc` and
+`pkill -f` all resolved ALLOW behind one apostrophe. The change can only make
+more argv visible, never fewer.
+
 🔴 The argv-based checks below are NOT a return of that helper, and the
 distinction is load-bearing. They never BLANK bytes and never decide which parts
 of a string are inert. They tokenise, and a quoted `-m` argument is simply ONE
@@ -97,7 +108,7 @@ These are gaps in COVERAGE, not bypasses of anything the globs used to hold: the
 opencode config keeps a broad `ask` over the same families, so an uncovered
 spelling still meets friction rather than silence.
 """
-import sys, os, json, re, shlex, ipaddress, subprocess
+import sys, os, json, re, shlex, ipaddress, subprocess, collections
 
 
 # =========================================================================== #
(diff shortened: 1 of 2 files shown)


ZacxDev:
🔴 **A claim in this PR's description is wrong, and it is the claim a reviewer
was most likely to rely on.** Posting it here because this PR is merged and the
text stays readable.

> This is **not** the reverted `_strip_message_text()` helper … The change can
> only make **more** argv visible, never fewer.

That is true of `split_commands`' own output. It is **false at the seam that
matters**, `check_git_commit_to_main`'s population.

Lifting a heredoc body out of the scan **changes the quote state of everything
after it**. The old parse's corrupt state sometimes left a `$( )`/backtick
substitution *unquoted, and therefore visible*, where the clean parse leaves the
same bytes inside a quote that now balances differently. A 20,000-input fuzz
comparing **verdicts** from the full 15-check policy found **four** inputs where
the scanner change alone turns a DENY into an ALLOW — e.g.

```
'-m' || <<-EOF / it's / EOF / "won't && clean; bash `mkfs.ext4` …
```

Comparing argv *sets* does not find them, and is the wrong yardstick: the old
scanner legitimately emits heredoc terminator lines and the corrupted whole-tail
blob, both of which *should* disappear. That is why the PR's own fuzz reported
clean.

Two further measurements against this PR's head, for the record:

- **`<<<` is still read as a heredoc at the second `<`.** `cat <<<x` reaches the
  scan at the second `<`, where `<<x` parses as `<<` + tag `x` and the rest of
  the text is swallowed as an unterminated body. `origin/main` inherits this.
- **The branch-resolution half has 15 measured DENY→ALLOW spellings** on commands
  that really commit to `main` — the four from the adversarial audit plus eleven
  more, including `{ cd <main>; git commit; }`, `if true; then cd <main>; git
  commit; fi`, `cd <main>; bash <<'EOF' / git commit / EOF`, a sourced env file
  the same command overwrites, and a `WT=x true` command-prefix assignment
  recorded as persistent. All are live on `origin/main` now.

**Fix: #405**, which reworks the branch resolution against a bash-measured model
(bash 5.3.15) and unions three scanner readings so the here-string correction
cannot cost a deny. Bash-oracle differential: **64 misses on this PR's head, 50
on `origin/main`, 0 on #405**.

The scanner half was also extracted alone as #402 while this PR was still open;
that is now superseded by #405 and closed.

PR #401, opened 2026-08-11
fix(guard): resolve a `-C` handle the command sets itself, and say so when it can't

ZacxDev:
The commit-to-main guard denied the documented worktree recipe. datapacket-talos
CLAUDE.md rule #10 mandates:

    WT=/tmp/wt-topic
    git -C $DATAPACKET worktree add --detach "$WT" origin/trunk
    git -C "$WT" commit -m "…"

`$WT` is set by the same line the guard is judging, so it is in no environment
`expandvars` can see. The `-C` fell back to the cwd — the primary clone on
`trunk` — and denied a commit that was going into a detached worktree. Measured
today: three consecutive re-spellings denied (`git -C "$WT"`, a `checkout -b`
first, then `(cd "$WT" && git commit)`), and the operator had to run the commit
by hand.

Two independent changes:

1. `_line_local_assignments` parses LITERAL `NAME=value` assignments out of the
   command text and applies them before `expandvars`, so the line's own value
   wins over a stale exported handle. Only what is written down is resolved —
   `WT=$(mktemp -d)` and `WT=/tmp/wt-$$` are skipped and still take the fallback.
   Nothing touches `os.environ`; the mapping is threaded explicitly, so one
   command's text can never change how a later one resolves.

2. `_fallback_note` prefixes the deny when a repo-naming token did NOT resolve
   and the verdict landed on the cwd instead. `_dash_c_dir`'s docstring already
   called that deny "self-diagnosing because it names the repo it judged" — it
   named the cwd, which is precisely the wrong repo, and never said a `-C` had
   been discarded. That false claim is corrected in place rather than restated.

Not a loosening. The fix resolves MORE paths, and a resolved path is judged
exactly like an absolute `-C`, so it also closes a fail-open the fallback had:
`WT=<trunk-clone>; git -C "$WT" commit` from a feature-branch cwd previously
ALLOWED (unresolvable -> fell back to the harmless cwd) and now correctly denies.

Verification:
- 15 new tests in section 8d(i-b)/(i-c). Watched RED first: 11 of them failed on
  the unmodified implementation, then green.
- Mutation sweep, 11 mutants, all KILLED, each by the test written for it (killer
  names recorded, not just a count). Two SURVIVED the first pass and drove two
  more tests: a computed value whose literal spelling happens to name a real
  directory must still not resolve (a genuine fail-open), and the note must
  require BOTH `stray` and `repo_dir == base` or it lies on a deny about a
  repo that did resolve.
- Authoritative gate: `nix build .#checks.x86_64-linux.pytests` PASS — 7291
  collected, 7290 passed, 1 pinned skip; guard suite 1330 over a floor of 1260.
- Driven through the REAL bash-guard.py adapter over 8 cases against live repos
  (not fixtures), deployed-vs-worktree side by side: 2 negative controls still
  deny, 1 positive control still allows, the 3 incident shapes flip deny->allow,
  and 2 must-still-deny cases hold. 8/8.

NOT yet deployed — `~/.claude/hooks/guard_core.py` is a home.file copy, so this
changes nothing until `home-manager switch` / `ship.sh`.

### DIFF
--- a/scripts/claude-hooks/guard_core.py
+++ b/scripts/claude-hooks/guard_core.py
@@ -1021,15 +1021,77 @@ def _safe_getcwd():
         return None
 
 
-def _resolve_dir(value, base):
+# A LITERAL `NAME=<value>` assignment in the command text. Deliberately refuses
+# any value the guard would have to EXECUTE to know: `$(…)`, backticks and `$$`
+# are excluded at the callsite, not guessed at. Stops at the shell metacharacters
+# that end a word, so `WT=/tmp/x; git …` and `WT=/tmp/x && git …` both yield the
+# path alone.
+_ASSIGNMENT = re.compile(
+    r"""(?:^|[;&|(){}]|\s)(?P<name>[A-Za-z_][A-Za-z0-9_]*)="""
+    r"""(?P<value>"[^"]*"|'[^']*'|[^\s;&|()'"]*)""")
+
+# The values that mean "you would have to run this to find out".
+_UNKNOWABLE_VALUE = re.compile(r"\$\(|`|\$\$")
+
+
+def _line_local_assignments(cmd):
+    """`{NAME: value}` for every LITERAL assignment in the command text.
+
+    🔴 This exists because the guard parses TEXT while the shell parses a
+    PROGRAM, and the documented worktree recipes set the path in the very line
+    being judged:
+
+        WT=/tmp/wt-topic
+        git -C "$WT" commit -m "…"
+
+    `$WT` is not in the hook's environment — it does not exist yet anywhere but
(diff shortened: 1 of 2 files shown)


ZacxDev:
Superseded by #396, which landed while this was being written and already carries the literal-assignment resolution AND the unresolvable-`-C` deny note — plus sourced env files, positional cwd tracking with subshell scope, and the heredoc quote-state bypass, none of which this had.

Re-ran this PR's 13 new tests against `origin/main`: 19 of 20 parametrized cases pass, i.e. #396 covers essentially all of it. The one survivor was a real defect in #396's ordering (`expandvars` before `cvars`, which fails open when a command reassigns an exported handle). That one finding is carried forward as #403 with its own tests; the rest of this branch is redundant and is not worth landing.

PR #402, opened 2026-08-11
fix(guard): one apostrophe in a heredoc body disabled every argv check (scanner half of #396)

ZacxDev:
**This is the scanner half of #396, extracted so it can ship on its own.** #396
bundles it with a rewrite of how the guard decides which repo/branch a command
acts on; an adversarial audit found four DENY→ALLOW regressions in that half.
This PR carries **only** the scanner fix, which closes a bypass that is deployed
right now.

## The bypass

A heredoc body is prose, and `split_commands` tracked quote state through it. A
lone apostrophe (`don't`, `it's`) opened a quote that never closed, so every
command **after** the heredoc was swallowed into that quoted buffer and never
parsed. `_tokenise` then fell back to a whitespace split whose `argv[0]` is a
word from the message, and every argv check silently stopped matching.

Measured on the shipped core (`origin/main`), cwd = a clone on `main`. The two
arms differ by **one character**:

```
cat > /tmp/m <<EOF
fix: don't blindly stage        <- vs "fix: blindly stage nothing"
EOF
<victim>
```

| victim | with the apostrophe | without |
|---|---|---|
| `git commit -m x` | **ALLOW** | DENY |
| `git stash push -m wip` | **ALLOW** | DENY |
| `git -C /tmp reset --hard origin/main` | **ALLOW** | DENY |
| `git clean -fd` | **ALLOW** | DENY |
| `talosctl -n 1.2.3.4 reset` | **ALLOW** | DENY |
| `mkfs.ext4 /dev/sdc` | **ALLOW** | DENY |
| `dd if=/dev/zero of=/dev/sdc` | **ALLOW** | DENY |
| `pkill -f e2e/run.sh` | **ALLOW** | DENY |

Three of those are irreversible-action families.

## What changed

- `_scan` recognises the heredoc operator (`<<TAG`, `<<-TAG`, quoted and
  backslash-escaped tags) and **lifts the body out of the surrounding quote
  state**. Nothing is blanked: bodies are still returned, still parsed as
  commands, still checked — so `bash <<EOF` keeps every deny it had (asserted).
  This is **not** the reverted `_strip_message_text()` helper the module
  docstring forbids; the distinction is argued at the definition.
- A here-**string** (`<<<`) is consumed **whole** rather than excluded inside
  the heredoc branch. Testing `not startswith("<<<")` there is not enough — the
  scan also visits the **second** `<`, where `<<<x` reads as `<<` + tag `x` and
  swallows the rest of the text as an unterminated body. Measured on #396's head
  as well: `cat <<<x\ngit stash …` produces a spurious body there.
- An unterminated quote from **any** source re-scans its own tail. Lifting alone
  was not enough for `bash <<EOF / it's fine / git stash push -m wip / EOF`,
  where the corruption then happens inside the body's *own* parse.
- 🔴 `split_commands` **unions the shipped parse back in**, gated on a computed
  `diverged` flag.

## 🔴 Why the shipped parse is KEPT — a correction to #396

#396 states the change "can only make **more** argv visible, never fewer".
That is true of `split_commands`' own output, and **false at the seam that
matters**. Lifting a body out changes the quote state of everything after it,
and the old parse's corrupt state sometimes left a `$( )`/backtick substitution
**unquoted, and therefore visible**, where the clean parse leaves the same bytes
inside a quote that now balances differently.

A 20,000-input fuzz comparing **verdicts** from the full 15-check policy found
**four** inputs where lifting alone turns a DENY into an ALLOW. (Comparing argv
*sets* does not find them and is the wrong yardstick: the base scanner
legitimately emits heredoc terminator lines and the corrupted whole-tail blob,
both of which *should* disappear.) One of the four:

```
'-m' || <<-EOF / it's / EOF / "won't && clean; bash `mkfs.ext4` …
```

Neither parse is "right" about text bash itself rejects as unterminated; a guard
simply must not lose the deny. Taking the union makes the additivity claim
**true by construction**, so no future fuzz seed can find a fifth case. All four
inputs are pinned as a regression test and the union is mutation-covered.

The second scan is **skipped** when the first reports `diverged is False` — a
proof, not a heuristic, that both modes produced identical segments. A textual
precondition would be unsound: `"'" 'x` leaves an unterminated `'` with an
**even** count of them.

## Verification

**24 new tests.** Same test file against `origin/main`'s `guard_core.py`:
**35 failed / 1309 passed**, of which **4 are pre-existing artefacts** of running
the file without its sibling hook/plugin scripts (confirmed by running the
*unmodified* test file the same way) — so **31 genuine reds**, all green here.

| gate | result |
|---|---|
| `scripts/run-tests.sh --set hermetic` | `collected=7305 passed=7304 skipped=1 failed=0` · **RESULT: PASS** |
| `test_guard_core.py` | **collected=1344, passed=1344** (floor 1260, ceiling 1575 — no floor bump due) |
| `test_bash_guard.py` | `RESULT: all good` |

Fuzz — 20,000 inputs each, **positive control reported beside every zero**:

| fuzz | result | positive control |
|---|---|---|
| verdict, base vs this (seeds 20260811 / 1 / 777) | **DENY→ALLOW = 0** | ALLOW→DENY = 270 / 259 / 274 |
| argv-set additivity | **LOST = 0** | GAINED = 5026 inputs |
| legacy-mode ≡ `origin/main`'s `split_commands` | **MISMATCHES = 0** | lifted mode really differed on 10088 |

**15 semantic mutants, all KILLED**, each by a named test — including *drop the
union* (#396's shape), *the divergence gate under-reports a heredoc*, *…an
unterminated quote*, *legacy mode also lifts*, *bodies lifted but never
returned*, *two heredocs come back reversed*, *recovery restarts at the quote so
it makes no progress*, and both here-string mutants.

**Control battery**, 39 cases, real `git init` fixtures, `evaluate(cmd,
"claude-code", cwd)`:

- **29 negative controls all DENY** — bare commit on `main`/`trunk`, `--amend`,
  commit via heredoc, via heredoc **with** an apostrophe, `git -C <repo-on-trunk>
  commit` from a topic cwd, `cd`/subshell/`bash -c`/`GIT_DIR=` hops into `main`,
  `git stash`, `git reset --hard`, `git clean -fd`, `git add --all`, `pkill -f`,
  each also in its "after an apostrophe heredoc" spelling, plus `bash <<EOF`
  bodies that really execute.
- **10 positive controls unchanged from `origin/main`.** In particular the three
  worktree spellings #396 exists to unblock — `(cd $WT && git commit)`,
  `git -C "$VAR" commit`, and `-C` via a sourced env file — **still DENY here**.
  That is the branch-resolution half and is deliberately **not** in this PR.

**Parse cost** (`commands()`, min of 7 × 200 reps): ordinary command lines are
unchanged because `diverged=False` skips the second scan — `git status`
0.050 → 0.049 ms, a `git -C "$WT" …` worktree recipe 0.077 → 0.092 ms. Text that
carries a heredoc pays for both scans: 10-line 0.143 → 0.258 ms, 400-line
5.09 → 9.18 ms.

## 🔴 Not live

`~/.claude/hooks/guard_core.py` resolves into `/nix/store` — a home-manager
`home.file` **copy**, not an out-of-store symlink. Merging this changes nothing
on either host until someone runs `home-manager switch`. **None was run as part
of this work.**

### DIFF
--- a/scripts/claude-hooks/guard_core.py
+++ b/scripts/claude-hooks/guard_core.py
@@ -77,6 +77,16 @@ one extra step -- write the message to a file and use `git commit -F <file>` --
 which RULES already prefers over heredocs. That is a better trade than a
 security guard with a rotating cast of bypasses.
 
+🔴 NEITHER IS THE HEREDOC HANDLING IN `_scan` (2026-08-11). It LIFTS a heredoc
+body out of the surrounding quote state and then parses it as commands exactly
+as before — nothing is blanked, nothing is declared inert, and `bash <<EOF` keeps
+every deny it had. It exists because the opposite of the reverted helper was also
+a bug: the scanner tracked quote state THROUGH the body, so one apostrophe in a
+commit message (`don't`) opened a quote that never closed and swallowed every
+command AFTER the heredoc. Measured on the shipped guard — `git commit`,
+`git stash`, `git reset --hard`, `git clean -fd`, `talosctl reset`, `mkfs`,
+`dd of=/dev/sdc` and `pkill -f` all resolved ALLOW behind one apostrophe.
+
 🔴 The argv-based checks below are NOT a return of that helper, and the
 distinction is load-bearing. They never BLANK bytes and never decide which parts
 of a string are inert. They tokenise, and a quoted `-m` argument is simply ONE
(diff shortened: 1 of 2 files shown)


ZacxDev:
Closing — superseded by **#405**.

This PR extracted the scanner half of #396 so it could ship without the
branch-resolution rewrite. **#396 merged as one change while this was open**
(`3a5e32c`), so `main` already has the scanner fix and this PR's premise — "the
bypass is deployed right now" — is no longer true. It also conflicts.

Two things learned here are carried into #405 rather than lost:

1. **The union.** Lifting a heredoc body out changes the quote state of
   everything after it, so a scanner fix is *not* automatically additive: a
   20,000-input verdict fuzz found four inputs where lifting alone turns a DENY
   into an ALLOW. #405 unions three readings instead of two.
2. **The here-string fix costs denies.** `main` reads the second `<` of `<<<x`
   as `<<` + tag `x`; correcting that loses seven measured denies across
   `git stash`, `git clean`, `mkfs`, `dd`, `pkill` and `talosctl`, because
   main's misparse reparses the tail with fresh quote state and accidentally
   exposes the argv. This PR made the correction **without** buying those back —
   caught by an adversarial audit of this PR, and fixed in #405.

No `home-manager switch` was run for either PR.

ZacxDev:
Closing as **subsumed by #396**, on measurement rather than on reading the diff. Recording the numbers so the decision is auditable and reversible.

## The named defect is already closed on `main`

This branch's merge-base is `e1219d4` — it was cut *before* #396, #395 and #403 landed, and #396 shipped its own fix for the same scanner. `main` already carries `_read_heredoc_tag` and `_consume_heredocs`; only `_scan_raw`/`_lift` are unique to this branch.

**Differential probe, `main` (`a91b207`) vs this branch's core**, 18 scanner-hazard wrappers × 8 hazardous victims = **144 commands**:

```
gaps #402 would CLOSE     (main ALLOW, 402 DENY): 0
regressions #402 would ADD (main DENY, 402 ALLOW): 0
```

Wrappers included the apostrophe body this PR is named for, plus unbalanced double-quote, backtick, quoted/`<<-`/double-quoted tags, two heredocs on one line, a body containing the tag as a substring, a here-string, bare unterminated single and double quotes, the victim placed *before* the heredoc, `bash <<EOF` executing its body, CRLF line endings, and a trailing comment after the victim.

**The instrument was validated before its verdict was read.** A first version of this probe scored 153/153 DENY on both cores, and its own control flagged that as one-sided — the `cat > file <<EOF` wrapper was tripping `check_heredoc_to_file`, so the victim after it was never the reason for the verdict. v2 gates every wrapper on two preconditions before counting it: the wrapper *alone* must ALLOW, and the wrapper with a benign victim (`git status`) must ALLOW. All 18 passed both, so a DENY in the matrix is attributable to the hazard rather than the wrapper.

## The other two properties also already hold

- **Bounded recursion/quote recovery.** `main` has `test_recursion_is_bounded()`, and measured directly: 200 nested `bash -c`, 40 nested heredocs, 2,000 apostrophes in one body, a 5,000-line body, and 10,000 unbalanced quotes all return **DENY**, no `RecursionError`, worst case **2.7 ms**.
- **An unterminated quote must not blind the tail.** Covered behaviourally on `main` — the `bare unterminated squote` and `bare unterminated dquote` wrappers above are 8/8 DENY.

## Why not merge it anyway

Test-merged into `main` to price it: **9 conflict hunks / 317 conflicted lines** in `guard_core.py` and **802** in the test file. That is hand-reconciling two independent rewrites of a security-critical parser, in the file that has already produced three fail-opens in one day — for a measured behavioural gain of zero.

The one genuinely novel idea here is the differential parse (`_lift=False` reproduces the shipped scanner, `diverged` proves the two walks agree). Its value has decayed rather than being wrong: the "shipped scanner" it pins is `e1219d4`'s, which `main` has already replaced. Ported as-is it would prove agreement with a parse nobody runs any more. If that safety net is wanted, it should be re-anchored on the current scanner as its own small PR — not carried in on this conflict.

Reopen if any of the above is wrong; every number here is a command that can be re-run.

PR #408, opened 2026-08-11
fix(collector): one malformed file_path aborted every claude session-summary — and its skip counter never reached systemd

ZacxDev:
Closes the five non-blocking findings from the audit of #398, plus one hole found in review. Two behavioural changes (items 3 and 3b); the rest are claims corrected where a reader will hit them.

## 1. A malformed `file_path` aborted EVERY claude session-summary

`changed_paths.to_repo_relative` raises on a non-`str` or blank path. The opencode side `str()`-coerces so it cannot reach the raise; **the claude side did not**, and `run()` had **no per-session `try`** — so one bad transcript anywhere under `~/.claude/projects` killed the whole 5-minute pass and stopped *all* claude session-summary emission until that file changed.

Measured on the pre-change tree, **six** distinct shapes did it, not one:

| malformed value | exception |
|---|---|
| blank / whitespace `file_path` | `ValueError: changed-paths entry is not a usable path` |
| list `file_path` | `TypeError: unhashable type: 'list'` (before `summarize()` was ever reached) |
| dict `file_path` | `TypeError: unhashable type: 'dict'` |
| int `file_path` | `TypeError: expected str, bytes or os.PathLike` |
| `tool_use.input` not a dict | `AttributeError: 'list' object has no attribute 'get'` |
| non-`str` `timestamp` | `TypeError: '<' not supported between 'int' and 'str'` |

Reachability today is **0 across the 600 transcripts the tailer walks** — but that is a property of today's data, not of the code, and it is live on both hosts.

**Two layers, deliberately:**

- **The extraction site VALIDATES rather than coerces.** I did *not* take the brief's "coerce to match opencode" option: `str(["a.py"])` is `"['a.py']"`, which `to_repo_relative` accepts as a plain relative path and **emits** — a manufactured path in the one module whose entire contract is that nothing is invented. An unreadable value is counted in a new `unusable_file_paths` and excluded.
- **`run()` wraps only `summarize_transcript`** — skip, name the path and exception on stderr, count into `failed=N`, and **do not record the signature**, so a code fix heals the corpus without the file having to change. `emit_event` stays *outside* the try: a broken spool is systemic, and turning it into a per-session skip would convert a loud outage into a quiet count. Pinned by `test_an_emit_failure_is_still_FATAL`.

The non-`str` `timestamp` case is **deliberately left unguarded** at the site, so the `run()` wrapper is exercised by a case no earlier check rejects rather than being dead code.

## 1b. Review catch — `failed=N` had to reach the EXIT STATUS

The first revision returned **0 unconditionally**. That made the skip a fail-open one level up: `claude-activity-source` is `Type=oneshot` with `OnFailure = notify-failure@%n.service`, and this script is its **second** `ExecStart` — so a systemic breakage (`changed_paths` throwing for *every* session, a bad deploy, an import resolving differently under the timer) would have read as **unit success, no toast**, with the only trace a stderr line in journald nobody is tailing. That is precisely the silent-zero class this payload exists to remove.

Now `if failed: return 1`.

**A ratio rule (`failed >= emitted`, "escalate only the systemic shape") was written first, tested, and rejected** — and the boundary test I wrote for it is what exposed the flaw. The timer scans every transcript but emits only *changed* ones, so nearly every real tick has `emitted=0`, which makes the ratio **identical to `failed > 0` in practice** while adding a boundary nobody could predict from the summary line. It is also wrong on the merits: a transcript that cannot be summarised does not degrade a number — that session's summary is **missing**, and stays missing while the condition holds. Ongoing data loss should keep alerting. The permanently-red-gate objection applies to a gate standing between someone and a merge, not to a deadman reporting that data is *still* being dropped.

Stated in the comment rather than left to be found: **this still cannot catch a breakage that makes summaries WRONG rather than raising** — `failed` counts exceptions only.

### Positive-control pair

Same code path, count moves — a `failed=0` from a counter wired to nothing is indistinguishable from a real zero:

```
test_a_run_with_no_bad_transcript_reports_failed_zero
  session-tailer: scanned=3 emitted=3 failed=0 [first-seen=3] …     <- 3 good sessions

test_one_unsummarisable_transcript_does_not_abort_the_run
  session-tailer: scanned=4 emitted=3 failed=1 [first-seen=4] …     <- 3 good + 1 bad
  session-tailer: ERROR summarising …/bad.jsonl: TypeError: '<' not
      supported between instances of 'int' and 'str' — session SKIPPED, not emitted
```

The three healthy sessions still emit. `unusable_file_paths` has its own pair (0 with a good path, 1 with a malformed one) and is asserted **on the emitted event**, not just in a rollup dict. The exit status is measured at **both ends of the emitted axis** — `failed=1/emitted=0` and `failed=1/emitted=9` both return 1 — so the verdict provably does not depend on how much healthy work happened alongside.

### Mutation matrix — 12 mutations, each dying to its own named test

| # | mutation | verdict | died with |
|---|---|---|---|
| M1 | site guard removed (back to `if fp:`) | DIED | `ValueError: changed-paths entry is not a usable path` (9 failed) |
| M2 | site guard **coerces** (`str(fp)`) instead of validating | DIED | `SUPERSECRET` reached the payload |
| M3 | site guard counts always, even a good path | DIED | `test_the_counter_is_zero_when_every_path_is_usable` |
| M4 | non-dict `input` guard removed (back to `or {}`) | DIED | `AttributeError` |
| M5 | per-session `try` removed | DIED | `TypeError: '<' not supported…` — the run aborted, i.e. the original symptom |
| M6 | `try` widened to swallow `emit_event` too | DIED | `test_an_emit_failure_is_still_FATAL` |
| M7 | failure swallowed (no stderr report) | DIED | `assert 'ERROR summarising' in …` |
| M8 | `failed` dropped from the summary line | DIED | `assert 'failed=1'` *and* `assert 'failed=0'` |
| M9 | skipped transcript marked done (sig recorded on failure) | DIED | `test_a_skipped_transcript_is_not_marked_done_and_retries` |
| M10 | exit status ignores failures (always `return 0`) | DIED | `assert 0 == 1` on both systemic and single-bad |
| M11 | exit status uses the **rejected** ratio rule | DIED | `assert 0 == 1` at the `emitted=9` end |
| M12 | run always fails (`return 1` unconditionally) | DIED | the two never-red controls (clean run, idle tick) |

Control (unmutated): **92 passed, rc=0**. Survivors: **none**. M5/M7/M8/M10/M11 were re-run with `--tb=long` because a crude symptom scan matched the tests' own source text — each fails on the intended assertion, verified line by line.

## 2. A wrong number in `changed_paths.py`'s docstring

The temp-worktree bucket read *"~150+ … spread over ~30 directories"*. Re-measured read-only over every transcript the tailer walks: **802 paths across 487 distinct `wt-*` worktree roots** — understated **~5.3x on paths and ~16x on directories**.

The conclusion is unchanged and the corrected figure is what makes it worth stating: the counted-but-dropped worktree set alone is **1.7x larger than the entire emitted path set** (802 vs 470).

```
distinct modified paths     3,290
  under cwd (emitted)         470  (14.3%)
  outside cwd (counted)     2,820  (85.7%)
    tmp scratchpad          1,225
    tmp worktrees             802   over 487 wt-* roots
    elsewhere under home      626
    other /tmp                167
```

The headline in/out ratio reproduces across three independent runs (14.3% / 15.2% / 16%) — **that ratio was stable; the sub-classification had never been measured, only estimated.** The docstring now names its corpus size so a future re-run can tell drift from disagreement.

## 3. Why opencode `patch` parts are excluded — recorded beside `_FILE_TOOLS`

Independently reproduced against the live store (read-only): **414 `patch` parts across 63 sessions**, key set `{type, hash, files}`, `files` absolute. Excluding 8 bulk parts (>20 files — whole-tree git snapshots), they carry **177 paths across 31 sessions** that neither summariser counts, against the **289** the file-tool extractor reports — **+61%**.

The exclusion is correct and now says why: `claude/session-tailer.py`'s `_FILE_TOOLS = {Edit, Write, NotebookEdit, MultiEdit}` makes it blind to bash-driven changes in exactly the same way, so counting `patch` on one side alone would make the two sources measure different things while being rendered in one column — worse here than for line counts, because `changed_paths` feeds subsystem association, where a wider path set makes a subsystem look more active. **Reopen only together with a claude-side bash-change extractor.**

`churn()`'s existing ⚠ block now says explicitly that it settles the *line-count* question only.

## 4. The coverage ratio, where a consumer will read it

`scripts/collector/README.md` now states that `changed_paths` carries **roughly one path in seven** (470 of 3,290 = 14.3%; 14–16% across three runs), what the excluded 85% consists of, and that a short list must be read beside `changed_paths_outside_cwd`. Adds a row for `unusable_file_paths`, and documents the skip-and-report behaviour **including** the non-zero exit — a README describing the skip without the escalation would document a fail-open.

## 5. Correcting the subsystem-store decision record

Its "derived session association" bullet said the mapping works from *"what the collector already emits (`kind=session-summary` carries git commits/pushes)"*. That was **wrong when written** — the payload carried integer *counts*, never a commit and never a path. Anyone building P0 from that sentence would have reached the resolver and found its input did not exist. Recorded rather than quietly edited.

The passage now describes the mechanism that exists **and is populated**: #398 deployed to both hosts, 585 historical opencode sessions backfilled (362 files / 94 commits / 37,311 lines that read as a hard zero). Plus the two properties a resolver must be built against — `changed_paths` is `null`, not `[]`, when unobservable; and it is a **lower bound** at ~1-in-7 coverage. Also records that re-running the backfill is not a one-liner: `opencode/backfill.py:29-32` clears the *message*-tailer state too, and message rows carry no `argMax`-on-read dedupe contract, so a naive re-run duplicates ~6,100 rows.

## Gate

`nix build .#checks.x86_64-linux.pytests` — read from the build's own status, unpiped. **`BUILD_RC=0`.**

```
PASS  scripts/collector/claude/tests  (collected=110 passed=110 skipped=0 floor=105)
TOTAL collected=7808  passed=7807  skipped=1  failed=0  (floor: 7340 = sum of 17 per-target floors)
RESULT: PASS (exit=0)
```

Rebased onto `7d86237`. An earlier revision was red on `origin/main`'s own `test_drift_check.py` — not this branch — now fixed by #407. Per-path collision check (`git log <base>..origin/main -- <path>`) confirmed main touched only `nix/home.nix` and `scripts/tests/test_drift_check.py`; no overlap with this branch's eight files, clean rebase.

Floor: only the one target this branch moves, re-measured twice as tests were added — 59 (pre-existing, already 46 behind) → 100 → **105**, each read from the gate's own per-target line and put through `m - min(50, max(1, m/20))`. `--check-floors` passes the two-way pin.

## Things in the brief I think are wrong

1. **"Coerce on the claude side to match opencode" would manufacture data** — rejected, reasoning in a code comment, pinned by mutation M2.
2. **Knock-on finding, live on both hosts, not fixed here:** opencode's `str(inp.get("filePath") or "")` will emit a stringified list as a repo-relative path *today* — the same class this PR closes, on the source that supposedly "cannot reach the raise". Being scoped separately; whether it ever actually fires needs measuring first.
3. **Reachability numbers.** The brief said 0 malformed across 4,770 transcripts; the tailer only walks **600** of the 4,782 `.jsonl` files (`iter_transcripts` skips `subagents/` and `wf_*`). Same verdict — 0 — stated at the scope actually measured.
4. **A reported `session_tailer.py` collision did not exist** (an earlier resume brief); `git log <base>..origin/main -- <path>` was empty and the file was byte-identical to my base.

## Not done, on purpose

`unusable_file_paths` is claude-only. Making it a peer of the `changed_paths*` block would mean adding a sixth key to `changed_paths.PAYLOAD_KEYS` and emitting it from both tailers — wider than these findings justify, and opencode's coercion means it would read 0 there for the wrong reason.

### DIFF
--- a/claudedocs/decision-subsystem-store-rejected-2026-08-11.md
+++ b/claudedocs/decision-subsystem-store-rejected-2026-08-11.md
@@ -103,10 +103,29 @@ supplies the evidence the `org` strain test found missing.
 
 Decisions taken:
 
-- **Association is derived, not tagged.** Map session → subsystem from what the collector
-  already emits (`kind=session-summary` carries git commits/pushes) by matching path
-  components against the entry's slug and `aliases:`. This deliberately does **not** persist
-  a location, preserving the index's rule that location is always re-derived live.
+- **Association is derived, not tagged.** Map session → subsystem by matching the path
+  components of `kind=session-summary`'s **`changed_paths`** against the entry's slug and
+  `aliases:`. This deliberately does **not** persist a location, preserving the index's rule
+  that location is always re-derived live.
+
+  🔴 **CORRECTION — the original wording of this bullet was wrong when written.** It said
+  the mapping works from "what the collector already emits (`kind=session-summary` carries
+  git commits/pushes)". `session-summary` carried integer **counts** — `git_commits`,
+  `git_pushes`, `files_modified` — and never a commit, never a path. There was nothing to
+  match path components *against*; the derivation was asserted with no data behind it, and
+  anyone building P0 against that sentence would have got to the resolver and found its
+  input did not exist. Recorded rather than quietly edited: the failure was believing a
+  field name implied a payload, which is the same mistake as reading a type declaration as
+  a code path.
+
+  The paths exist as of **PR #398**, which added the `changed_paths*` block to both sources
+  and fixed the opencode extractor. Two properties the resolver must be built against, both
+  measured, neither optional: `changed_paths` is `null` when the file set could not be
+  observed — **not** an empty list, and a `null` read as "touched nothing" reintroduces
+  exactly the silent zero #398 removed; and it carries only paths under the session cwd,
+  **~1 in 7** of the modified paths (470 of 3,290, 2026-08-11), with the remainder counted
+  in `changed_paths_outside_cwd`. So a subsystem's session count is a **lower bound**, which
+  the liveness convention has to state rather than round away.
 - **The edge lives in ClickHouse**, as a new `kind=session-subsystem` event — one row per
   `(session, subsystem)`. Do **not** extend `session-summary`: consumers dedupe it with
   `argMax(<field>, ingested_at) GROUP BY session`, and a session touches N subsystems.
--- a/claudedocs/decision-subsystem-store-rejected-2026-08-11.md
+++ b/claudedocs/decision-subsystem-store-rejected-2026-08-11.md
@@ -119,13 +119,23 @@ Decisions taken:
   a code path.
 
   The paths exist as of **PR #398**, which added the `changed_paths*` block to both sources
-  and fixed the opencode extractor. Two properties the resolver must be built against, both
-  measured, neither optional: `changed_paths` is `null` when the file set could not be
-  observed — **not** an empty list, and a `null` read as "touched nothing" reintroduces
-  exactly the silent zero #398 removed; and it carries only paths under the session cwd,
-  **~1 in 7** of the modified paths (470 of 3,290, 2026-08-11), with the remainder counted
-  in `changed_paths_outside_cwd`. So a subsystem's session count is a **lower bound**, which
-  the liveness convention has to state rather than round away.
+  and fixed the opencode extractor — and they are **populated, not merely defined**: #398 is
+  deployed to both hosts, and 585 historical opencode sessions have been backfilled, turning
+  362 files / 94 commits / 37,311 lines that read as a hard zero into real rows. P0 therefore
+  has a corpus to resolve against on day one rather than only forward-going data.
+
+  Two properties the resolver must be built against, both measured, neither optional:
+  `changed_paths` is `null` when the file set could not be observed — **not** an empty list,
+  and a `null` read as "touched nothing" reintroduces exactly the silent zero #398 removed;
+  and it carries only paths under the session cwd, **~1 in 7** of the modified paths (470 of
+  3,290, 2026-08-11), with the remainder counted in `changed_paths_outside_cwd`. So a
+  subsystem's session count is a **lower bound**, which the liveness convention has to state
+  rather than round away.
+
+  ⚠ Backfilling again is not a one-liner: `opencode/backfill.py` clears the message-tailer
+  state as well as the summary state, and message rows carry no `argMax`-on-read dedupe
+  contract, so a naive re-run re-emits ~6,100 duplicate message rows. The 585-session
+  backfill cleared **only** the summary state deliberately.
 - **The edge lives in ClickHouse**, as a new `kind=session-subsystem` event — one row per
   `(session, subsystem)`. Do **not** extend `session-summary`: consumers dedupe it with
   `argMax(<field>, ingested_at) GROUP BY session`, and a session touches N subsystems.
--- a/scripts/collector/README.md
+++ b/scripts/collector/README.md
@@ -87,8 +87,21 @@ touched) needs the PATHS.
 | `changed_paths_total` | distinct count **before** the cap — the truncation discriminator |
 | `changed_paths_truncated` | the list is a prefix, not the whole story |
 | `changed_paths_outside_cwd` | distinct changed paths with no repo-relative form (scratchpads, other repos, temp worktrees). `total + outside_cwd == files_modified`. |
+| `unusable_file_paths` (claude) | tool blocks whose file-path argument could not be read as a path at all (non-string / blank / malformed `input`). Counted, never coerced into a path and never dropped in silence. Normally 0; a non-zero is a malformed-transcript signal, not a session measurement. |
 | `changed_paths_cap` | the bound this emitter was built with (256 — see the module docstring for the measurement) |
 | `stats_unavailable` | `[]`, or the groups that could not be observed. `"files"` covers `changed_paths*` + (on opencode) `files_modified` / `lines_added` / `lines_removed` / `languages`; `"git"` covers `git_commits` / `git_pushes`. |
+
+🔴 **SIZE YOUR EXPECTATIONS: `changed_paths` carries roughly one path in seven, not
+all of them.** Only paths under the session cwd have a repo-relative form; everything
+else is counted in `changed_paths_outside_cwd` and never listed. Measured over every
+transcript the claude tailer walks (2026-08-11, read-only): **470 of 3,290 distinct
+modified paths — 14.3%** — had one. Three independent dry runs put the ratio at
+14–16%, so treat ~1-in-7 as the planning number, not an outlier. The excluded 85% is
+dominated by the per-session agent scratchpad and by short-lived git worktrees under
+`/tmp`; the worktree bucket alone (802 paths) is **1.7x larger than the whole emitted
+set**. A consumer that reads a short list as "this session barely touched anything"
+will be wrong on most sessions — read `changed_paths_outside_cwd` beside it.
+`changed_paths.py`'s module docstring has the full classification.
 
 🔴 **A zero that means "cannot report" is indistinguishable from a real zero**, and
 that ambiguity is what this block removes. OpenCode emitted `files_modified=0`,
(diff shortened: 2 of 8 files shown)


ZacxDev:
## Audit round — the two named fixes, plus two claim corrections

**1. `build_event` was a 7th fatal shape, outside the try.** `project_basename(cwd)` → `cwd.rstrip("/")`, reachable when a transcript has **zero messages** *and* a non-str `cwd`: zero messages routes through `_mark_unobservable`, which skips `CP.summarize` — the only guard that would reject that cwd — so the rollup comes back clean and the raise lands one line later.

Verified pre-existing rather than taken on report: `git archive origin/main | tar -x` into a temp tree, same repro, same result.

| tree | result |
|---|---|
| `origin/main` (no part of this branch) | `AttributeError: 'list' object has no attribute 'rstrip'`, **state not written**, 1 event spooled |
| this branch, before the fix | identical |
| this branch, after | `run()` returns **1 cleanly**, `emitted=1 failed=1`, **state written**, neighbour emitted |

The cost was never only the abort: `run()` died before `save_state`, so every session that had already emitted that pass **re-emitted next tick, forever**. The test pins the persisted state, not just survival.

The comment now states the split as **per-session-data vs systemic**, which is what the original rationale meant — `emit_event` stays outside because a broken spool is broken for every session.

**2. `unusable_file_paths` pinned for a falsy non-dict `input`** (`[]` / `""` / `0`), with the absent-`input` arm pinned alongside so the guard's `is not None` is not free to become truthiness.

**3. The "⚠ WHAT THIS STILL CANNOT CATCH" comment asserted a completeness the code lacks.** It now also names what `failed` never counts: anything raised **outside** the per-session try (`emit_event`, `iter_transcripts`, `save_state`) still aborts by design, and a failure in the unit's **first** `ExecStart` stops this script running at all.

**4. Retraction of my own overstated claim.** The previous commit said `emitted=0` on "the overwhelming majority of real ticks" and that the two rules could not be told apart in practice. Measured from this unit's own journal, 2,781 ticks over 60 days: **emitted=0 on 55.0%**, and the two rules **agree on 81.0%** of ticks at `failed=1`. A majority, and distinguishable on about one tick in five. (Independently consistent with the audit's 57.2% / 88.2% over its own 1,993-tick window.) The rejection of the ratio rule stands on the merits — a missing summary is data loss that should keep alerting — not on that claim, which is now stated correctly in the code.

**5. `changed_paths.py` hedge corrected**, not the numbers: it said the sub-classification was "not measured, only estimated" while printing four exact counts summing to 2,820. The counts are measurements; it was the *superseded* figures that were estimated.

### Mutation matrix now 15 — all die, control 97 passed, rc=0, zero survivors

New this round:

| # | mutation | died with |
|---|---|---|
| M13 | `build_event` moved back outside the try | `AttributeError: 'list' object has no attribute 'rstrip'` — the original symptom |
| M14 | falsy non-dict `input` swallowed (truthiness, the `or {}` shape) | 3 failed (the `[]`/`""`/`0` params) |
| M15 | absent `input` counted as unusable | `test_an_ABSENT_input_is_not_counted` |

M13's first spelling deleted only the `ev = …` line, which left `ev` undefined and killed the test with a `NameError` — dying for the *wrong* reason, proving nothing. Corrected to the true pre-fix shape (`emit_event(emit, build_event(session, rollup))`) and re-confirmed with `--tb=long`.

### Gate

Rebased onto `ab99245` (one further than reported — #411 had landed). Per-path check: main touched only `scripts/tests/*` and `scripts/testlib/*`, no overlap with this branch's eight files.

#399's new `scripts/tests/conftest.py` **does not reach these targets** — it is a sibling of `scripts/collector/`, and pytest loads conftests only from rootdir down to the test file's own directory. Confirmed empirically: every collector target's collected count is unchanged apart from the tests this branch adds.

```
PASS  scripts/collector/claude/tests  (collected=115 passed=115 skipped=0 floor=110)
TOTAL collected=7879  passed=7878  skipped=1  failed=0
RESULT: PASS (exit=0)                    BUILD_RC=0
```

Floor re-measured from the gate's own line each time tests landed: 59 → 100 → 105 → **110**.

### Noted, not fixed here

A failure in the unit's **first** `ExecStart` (`claude/tailer.py`) silently stops all session summaries — pre-existing, out of scope, now at least written down in the code comment rather than nowhere.

PR #415, opened 2026-08-12
feat(handoff): the subsystem store had ONE writer — 21 entries in 1 scope while work spans ~12 repos

ZacxDev:
## What this is

The subsystem-knowledge store was missing one piece: **a way for an entry to come into existence from ordinary work rather than only from infra recon.** `/analyze-service` was its only writer, so the corpus reads 21 entries / 1 scope / 0 non-infra — and "no non-infra entries" was never evidence of no demand, because nothing could create one.

This is **one writer, one verb, riding `/handoff`** — 27 invocations, 26 in four days. Per the measured adoption lesson in `claudedocs/decision-subsystem-store-rejected-2026-08-11.md`: opt-in survives when it rides a ritual already performed and dies when it *is* the ritual.

**Not** rebuilding what was rejected: no `type:`-driven sections, no dependency graph, no multi-verb CRUD. Those stay rejected.

## Where the paths come from, and why

A design decision, made explicitly rather than assumed:

| source | verdict |
|---|---|
| telemetry `changed_paths` | The honest one — per-*session*. **Does not exist yet at handoff time**: `/handoff` fires before the session settles. Waiting for it moves the write off the ritual, which is what killed every other opt-in tool. |
| the agent's recollection | Available, and **not deterministic** — cannot be re-run to the same answer, cannot be tested. |
| **git** ✅ | Deterministic, re-runnable, and **`/handoff` step 1 already runs it** — so the path set is a free side-effect of recon the ritual performs anyway. Same shape that made `/analyze-service`'s index write stick. |

Git's weakness is authorship. It is **bounded and declared**, not fixed: the window is `(working tree vs HEAD, incl. untracked) ∪ (merge-base(HEAD, base)..HEAD)`, and every report prints *"what this BRANCH touched, NOT what this SESSION touched"*. On the base ref the commit half is empty and the window **degrades to `worktree` and says so** — an empty commit window and a branch with no commits are different facts. The reporting core takes paths as an argument and has never heard of git, so P1 can feed it from telemetry later without touching this logic.

## Five zeros, not one

`looked-at-nothing`, `no-store`, `scope-absent`, `no-match`, and below-threshold all render as an empty proposal list. `status` names which, and no two share a spelling.

`scope-absent` is **deliberately not an error**: under the old writer a repo with no entries was a failure condition; under this one it is the ordinary first run in every repo that is not the infra repo. That is the entire point of the change, and an exception there would make the intended case the failing case.

## Falsifiability

Entries carry `created_by:` (`analyze-service` | `handoff`; **absent = predates the stamp, never folded into either writer**). `--census` counts by scope and writer, so "do entries accrue outside infra recon?" is answered by counting.

**Baseline measured on the real store, read-only, byte-hash identical either side:** 21 entries · 1 scope · 21 unstamped.

## Two defects found by writing the tests, not by review

- The tail said *"none named an entry"* while an entry **had** been named and merely stayed under the threshold. A wrong explanation of a zero forecloses the next question.
- A filename recurring across unrelated directories (`values.yaml` × 6, in 3 different subsystems) **outranked every real directory on count alone**. Fixed structurally — a `coherent` key, because a subsystem is a place in the tree — **not** with a stoplist of "generic" names.

Plus a worktree trap: `--show-toplevel` would have made every agent worktree its own scope, sharding the store into hundreds of one-entry scopes no later run could resolve. The derivation reads the git **common** dir, so a worktree and its base clone agree.

Two of my own tests were also **passing for the wrong reason** (`roster` < `src` alphabetically, so the depth key could be deleted with no failure). Caught by trying to kill them; fixtures now chosen so alphabetical order disagrees with depth.

## Verification

**`nix build .#checks.x86_64-linux.pytests` — PASS.** `collected=7984 passed=7983 skipped=1 failed=0`; the 1 skip is the pre-existing pinned one, **zero new skips**. 105 new tests, `scripts/tests` 2251 → 2426.

- **Positive-control pair, reported together:** same depth, same filenames, same scope, same store — only the subsystem directory differs. **1 known entry under the positive control, 0 under test.** Plus positive controls on the census (it *can* see a stamped entry), the tree-hasher (it *can* observe a change), and the mutation harness (a no-op mutant still works).
- **Mutation matrix — measured twice, differently built.** In-suite (mutated copy imported in-process) and an independent sweep that mutates the real file on disk and runs the whole suite: **15/15 killed, each by its own named test**, NOOP control green (0 failed / 105 passed), module restored and byte-hash verified.
- **The one redundant check is labelled, not counted.** Deleting `if entry is not None: continue` was *measured* to leave the suite green — it is unreachable because nominations are drawn only from `unmatched_paths`. Labelled `REDUNDANT-BUT-KEPT` in the resolver's own convention rather than claimed as coverage.
- **Never-writes is behavioural, not spelled:** the whole store tree is hashed either side of every mode. A `grep` for `open(..., "w")` would pass while a different spelling writes.
- **Seam tested, not just the halves:** the emitted template is parsed back by the resolver *and* written to a scope dir and re-resolved end-to-end — "nominated" and "addressable" are two claims, and both are now checked.
- **Real git, not mocks** — `git` is pinned in `REQUIRED_TOOLS` and in the flake's `nativeBuildInputs`, so these run in both tiers.
- Floor re-pinned **2201 → 2376** from `_suggested_floor 2426` — the gate's own function, so what it suggests is by construction what it accepts. No new `HERMETIC_TARGETS` entry needed.

Contract change: `created_by:` added to the hashed `entry-schema` region of `analyze-service/SKILL.md`; the region hash in `test_subsystem_resolver.py` moved with it, read against the code (`from_mapping` does **not** read the field and must not — it is provenance, not identity), with `TestEntrySchemaAgreement` as the behavioural half.

## Not done, deliberately

Not merged, not shipped, not deployed. Note `claude/skills/` is a nix-store symlink — the skill step needs a `home-manager switch` to go live; `scripts/` is read from the repo and does not.


ZacxDev:
## Audit fixes — all 12 addressed, re-gated

**Gate: PASS** — `collected=8015 passed=8014 skipped=1 failed=0`, the 1 skip pre-existing and pinned, **still zero new**. `scripts/tests` 2251 → 2457. Floor re-pinned **2376 → 2405** from `_suggested_floor 2455` (the only floor moved).

### Required

**1. Where to write.** The absolute target is now in the step, with the negative half stated and the scope README requirement restored:
> 🔴 **The target is `~/.claude/analyze-service-index/<scope>/<slug>.md`** … **Never anywhere in the working tree** … **Read `~/.claude/analyze-service-index/<scope>/README.md` before writing there**

**2. Non-zero exit.** *"Any non-zero exit ⇒ print the stderr line verbatim and write NOTHING … Do not fall back to recollection — an entry written from memory is exactly the unverifiable content this store must not accumulate."*

**3. `resolved` + nominations.** Moved out of the status list into its own clause: *"Separately, and independently of `status=`: if a `NO ENTRY` block is printed…"* — with the reason inline, since that is the sentence a future editor would otherwise re-collapse. Also added the missing `no-match`/`scope-absent` bullet, plus a test deriving its coverage check **from `STATUS_PRECEDENCE`** so a status added later cannot go unmentioned.

**4. `Edit`, not `Write`** — anchored on `## Nuance / work-history`; `Write` only for a first-ever file, which has no prior content to lose.

**5. Path frames.** Fixed and this one was worse than described — from a subdirectory, untracked files *elsewhere in the repo* vanished entirely (cwd-scoping) on top of the prefix loss. Every git call now runs against the toplevel; test asserts a subdirectory call returns **the identical set** to a root call.

**6. Coherence docstring was false about its own code.** Reserved the last slot for the top non-coherent ref rather than demoting the key — demoting re-opens the recurring-filename case it exists to close. Three tests: it is never cut; the reservation is a no-op when nothing was truncated; and a no-op when a non-coherent ref already made the cut.

**7. `no-store` deleted.** `STATUS_PRECEDENCE` is four, and a new test asserts every declared status is actually **emitted** by a real `build_report` call — not merely declared.

### Also landed

**8. Fan-out key.** A ref whose covered paths branch into **≥2 subdirectories** is an index of subsystems, not one. Counted over subdirectories only, so a leaf directory isn't penalised for holding several files (own mutant for that). Live on this PR's diff: `skills` now renders `umbrella over several` and ranks below `tests`/`claude`.

**9. `--exclude`**, wired into the skill's command line with the handoff doc's own path. Exclusions are **counted into `notes`** — a path that vanishes with no note is a smaller number with no reason.

**10. Template ships a commented `aliases:` line** naming the `test_<slug>` stem, with a test that the parser does *not* read it as a real field.

**11.** Both dropped clauses restored. **12.** Kickoff block now precedes the gate, pinned **structurally on file order**, not on a phrase.

### Re-verification

Independent sweep rebuilt for the new guard set: **23 mutants, zero survivors**, NOOP control green (0 failed / 136 passed), module restored and byte-hash verified.

Two things the sweep caught that I'd otherwise have shipped:
- The **fan-out sort key died only to behavioural tests** while every other sort key had a named mutant — the matrix looked complete with a hole exactly where symmetry said it wasn't. It has its own now.
- `test_kills_git_failure_guard` was **environment-dependent**: with the guard gone the toplevel lookup also returns `""`, so `git -C ""` fell through to the CWD and the mutant reported six real paths *from devrc itself*. Worse than the zero it was demonstrating, and it made the assertion depend on where pytest was launched from. Pinned with `monkeypatch.chdir`.

Real store re-hashed after every run: **byte-identical to baseline** (`8c93bc43…`), still 21 entries / 1 scope / 21 unstamped.

Not merged, not shipped, not deployed.

PR #422, opened 2026-08-12
docs(clawgate): the extension load paths are two git checkouts, not the flat copy the skill described

ZacxDev:
## What this corrects

Measured 2026-08-12 while deploying extension **1.5.0** (clawgate PR #305): **both browsers were still running 1.4.0**, and the delivery path the skill documented could not have fixed it.

Brave loads the extension unpacked **in place from a git checkout**, on **two hosts**:

| host | Brave's "Loaded from" | what it is | state found |
|---|---|---|---|
| workbench | `~/workspace/homelab-talos/containers/clawgate/extension` | the permanently-dirty **base clone** | 14 behind trunk → 1.4.0 |
| laptop `.155` | `~/workspace/clawgate-extension/containers/clawgate/extension` | a **linked worktree**, branch `clawgate-ext-local` | 1 behind trunk → 1.4.0 |

The skill instead described a single flat copy at `~/clawgate-extension`, delivered by `scripts/sync-clawgate-extension.sh`.

## 🔴 That flat copy is loaded by nothing

It exists only on the workbench — the laptop has no such directory at all — and `--check` reports a clean `in sync — matches origin/trunk` **while both browsers run a different build**. A green check on a directory no browser reads. It cost a full sync + reload cycle that changed nothing before the real paths were found.

The doc now says the only authority on which build is loaded is Brave's own "Loaded from" line.

## What the doc now records

- the per-host table above, and that **both** hosts must be updated — doing one leaves the other stale
- per-host update commands, and per-host `brave://extensions` reload
- **why the workbench needs a surgical `git restore --source=origin/trunk --worktree`** of the subtree rather than a merge: its base clone is chronically behind and `merge --ff-only` refuses whenever incoming commits collide with local WIP (4 collisions with live media-stack edits on the day), plus the one-liner to detect that before assuming a merge is available
- that the resulting ` M` state is **self-healing** (content already equals trunk) but **silently revertible** — any later `checkout`/`restore` over that path drops the extension back to the old build with no error
- 🟡 open item: the workbench arrangement is the fragile one, because loading the base clone directly makes the deployed extension hostage to that clone's drift. The laptop's dedicated-worktree pattern is the coherent one; repointing needs Zach in the Brave UI (remove + re-add unpacked, re-check shortcuts).

The `.synced-from` provenance rule is kept — it still applies to any flat copy that *is* wired up — with the git-checkout equivalent added beside it.

## Gate

`nix build .#checks.x86_64-linux.pytests` → **8114 collected, 8113 passed, 1 pinned skip, 0 failed**.

Docs-only, so that is a regression check. The doc's factual claims come from live measurement on both hosts, not from the suite.

## Note

Merging does not deploy this — `~/.claude/skills/` is a `home.file` copy (`readlink -f` terminates in `/nix/store`), so it needs `scripts/ship.sh` / a `home-manager switch`.

https://claude.ai/code/session_019rCBXqoNtXHpvwEsLwBnJz

### DIFF
--- a/claude/skills/clawgate/SKILL.md
+++ b/claude/skills/clawgate/SKILL.md
@@ -181,6 +181,8 @@ absent toolchain, the `curl` form): `reference/agent-dispatch.md`.
   hot-reload). 🔴 **Brave has MULTIPLE PROFILES that load extensions independently and can point at
   DIFFERENT paths** — one profile was left a version behind exactly this way; checking the profile in
   front of you proves nothing, and agents can't read `brave://`, so use the `Preferences`-JSON sweep
-  in `reference/extension.md`. 🔴 **Never `git restore --source=<ref> --worktree` a subtree to
-  "freshen" a stale checkout** — identical content still BLOCKS `merge --ff-only`, and any later
-  checkout silently reverts it. `~/clawgate-extension` + `sync-clawgate-extension.sh` are RETIRED.
+  in `reference/extension.md` — which also verifies the HOTKEYS, since a reload silently left
+  `Ctrl+Shift+E` unbound in two profiles. 🔴 **Never `git restore --source=<ref> --worktree` a
+  subtree to "freshen" a stale checkout** — identical content still BLOCKS an `--ff-only` merge that
+  touches that path, and any later checkout silently reverts it. `~/clawgate-extension` is deleted;
+  `sync-clawgate-extension.sh` still exists in homelab-talos but is NOT the delivery path.
--- a/claude/skills/clawgate/SKILL.md
+++ b/claude/skills/clawgate/SKILL.md
@@ -174,11 +174,13 @@ absent toolchain, the `curl` form): `reference/agent-dispatch.md`.
   🔴 **But `clawgate-ci` does NOT run Playwright — browser-layer changes are UNGATED by CI.** Run
   `make e2e` locally and **count** the results: `tasks.spec.ts` `test.skip`s the whole file without
   Docker, so a "green" run can mean 17 tests never executed.
-- 🔴 **The browser extension does NOT ship via Flux — merging to `trunk` deploys NOTHING**, and it
-  runs on **TWO hosts** that each load it unpacked **in place from a git checkout**: workbench
-  `~/workspace/homelab-talos/containers/clawgate/extension` (the dirty base clone) and laptop
-  (`zach@192.168.50.155`) `~/workspace/clawgate-extension/containers/clawgate/extension` (a
-  worktree). Deploying = advance BOTH checkouts + reload each `brave://extensions` (no hot-reload).
-  🔴 **`~/clawgate-extension` + `scripts/sync-clawgate-extension.sh` are a DEAD path — nothing loads
-  that directory**, yet `--check` reports a reassuring "in sync" while both browsers run something
-  else. Only Brave's own "Loaded from" line is authority → `reference/extension.md`.
+- 🔴 **The browser extension does NOT ship via Flux — merging to `trunk` deploys NOTHING.** Brave
+  loads it unpacked **in place from a git worktree**, at the same path on **BOTH hosts** (workbench +
+  laptop `zach@192.168.50.155`): `~/workspace/clawgate-extension/containers/clawgate/extension`
+  (branch `clawgate-ext-local`). Deploy = `merge --ff-only origin/trunk` on both + reload Brave (no
+  hot-reload). 🔴 **Brave has MULTIPLE PROFILES that load extensions independently and can point at
+  DIFFERENT paths** — one profile was left a version behind exactly this way; checking the profile in
+  front of you proves nothing, and agents can't read `brave://`, so use the `Preferences`-JSON sweep
+  in `reference/extension.md`. 🔴 **Never `git restore --source=<ref> --worktree` a subtree to
+  "freshen" a stale checkout** — identical content still BLOCKS `merge --ff-only`, and any later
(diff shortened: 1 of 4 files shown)


ZacxDev:
## Reworked after the adversarial audit — the original PR body was wrong on two counts

Correcting this publicly rather than silently editing the body, since the wrong version was already readable.

**1. "Nothing loads `~/clawgate-extension`" was FALSE when I wrote it.** Brave keeps **four profiles** on the workbench, two active, and each loads unpacked extensions **independently** — they were pointing at two different paths, and the flat copy *was* loaded by the `Default` profile at the time. Repointing one profile then left the other (at that moment the `last_used` one) a version behind — the same class of bug this document exists to prevent, reproduced one level down.

The doc also named `brave://extensions` "the only authority", which is a page **no agent can read**. Replaced with a `Preferences`-JSON sweep (`location==4`) that an agent can actually run. It must quote `"$(dirname "$p")"` or a profile dir containing a space reports as `Profile` instead of `Profile 2` — that bug was live in my own first version of the sweep.

**2. "Self-healing" was FALSE and inverted.** Git compares worktree ↔ **index**, never worktree ↔ merge target, so a subtree restored from `origin/trunk` is still a local change even when byte-identical. It **blocks** `merge --ff-only origin/trunk` — the exact re-sync `homelab-talos/CLAUDE.md` prescribes for that clone — instead of dissolving into it:

```
error: Your local changes to the following files would be overwritten by merge:
	ext/manifest.json
Aborting                                                    rc=1
```

Proven in an isolated repo with both controls (ff proven to work on a clean behind-clone first; content proven byte-identical to the target). The adjacent "silently revertible" warning was correct and is kept.

## What changed in the diff

Both hosts have since been migrated to the **same** worktree path (`~/workspace/clawgate-extension`, branch `clawgate-ext-local`), so the per-host table, the surgical `restore` procedure and its collision one-liner are **deleted rather than corrected** — keeping them would leave a future agent following a recipe that writes a stale build back into the base clone. The audit's fix for that one-liner (`status -s -uall`, since `-s` collapses untracked dirs: `??` goes 51 → 152 on the real clone) is moot with the procedure gone.

`~/clawgate-extension` is deleted and `sync-clawgate-extension.sh` is recorded as retired, with the lesson attached: its `--check` reported a confident `in sync` about a directory no profile loaded, so a green check coexisted with every browser running an older build.

`claudedocs/chrome-extensions-inventory.md` turned out to have had the **correct** worktree path all along while the skill contradicted it; its stale `1.1.0` / "deployed via the clawgate container" row is refreshed and pointed at the skill.

## Remaining audit findings

- **#7 (four live callers of the retired script)** — `scripts/README.md`, a 12 KB test, the extension README and a handoff doc, all in `homelab-talos`. Out of scope for this repo; tracked as a follow-up PR there.
- **Gate numbers** — the earlier quote wasn't reproducible (base drift, not a regression). Re-run on the **rebased** tree: **8260 collected, 8259 passed, 1 pinned skip, 0 failed.**

Every command in the rewritten doc was executed verbatim on both hosts.

ZacxDev:
## Round 2 — re-audit findings addressed, with two corrections to the audit itself

### One finding refuted

**C — "the sweep dies under `set -e` and truncates."** It does not. A failing left operand of an `&&` list is exempt from `set -e`, and the loop's own exit status doesn't trigger one either. Measured in **bash, zsh and dash** — all three print every iteration and exit 0:

```
[A] hit
[C] hit
loop completed
rc=0
```

No change made on that basis.

### One suggestion that would have made things worse

**B** suggested filtering the sweep on extension *identity* rather than a path substring. I implemented it, then ran a positive control before publishing — `.value.manifest.name` is **null for every entry**, because an unpacked entry in `Preferences` carries **no manifest at all** (keys are `path`, `location`, `commands`, `permissions`…). The "improved" filter matched nothing and reported *"no clawgate extension"* for all five profiles: a silent zero, confidently wrong, strictly worse than the grep it replaced.

The real fix is to **drop the filter entirely** and list *every* unpacked extension per profile — which also closes B's actual hazard, an extension loaded from an unexpected path being invisible to a `grep clawgate`. The sweep now prints a line per profile including `(no unpacked extensions)`, and no longer swallows `jq` failures.

### A live defect found while verifying

The `commands` key on the same entry makes **hotkey bindings machine-checkable** — something the doc previously said had to be done by eye. Added, and it immediately caught this:

```
workbench [Default]   open-capture-pick: key=Ctrl+Shift+E assigned=false
laptop    [Profile 1] open-capture-pick: MISSING
```

The element-picker hotkey is dead in two profiles while the extension itself reports perfectly healthy — exactly the reload trap the doc warns about, now **detectable** rather than merely described.

### Remaining findings

| # | Disposition |
|---|---|
| A | Fixed — "Both are gone" was false. Only the flat copy is deleted; the script still exists on `homelab-talos` trunk with four live references, retirement pending **in that repo** |
| D | Fixed — `always fast-forwards` now carries the failure branch the deletion took with it |
| E | Fixed — ssh `bash -s` heredoc form added and executed |
| F | Fixed — inventory said EVERY profile; it is 2 of 5 workbench, 2 of 6 laptop |
| G | Fixed — the browser-bridge handoff still sent readers to the base clone and "Default profile" |
| H | Fixed — two listed worktrees no longer exist, incl. a cleanup command that would now fail |
| I | Fixed — restore-trap claim scoped to a merge that *touches* the restored path (`--staged --worktree` doesn't block either) |
| J | Fixed — dangling "the check above" referent |
| K | Fixed here — gate is **8260 collected, 8259 passed, 1 pinned skip, 0 failed** |

Every command in the doc was re-executed verbatim on both hosts, including the ssh form and a `set -e` run.

PR #430, opened 2026-08-13
fix(notifications): cut the desktop toast stream ~330 -> ~60/day so DND can go back off

ZacxDev:
Goal, stated as the acceptance test rather than a vibe: **a stream quiet enough that you can turn DND off and leave it off.** Every number below is answered against "would this survive that?"

PR #409 measured and shipped nothing. This ships the reductions — and corrects two of #409's per-producer numbers, one of them by 4x in the direction that changes what you should work on.

---

## 1. What changed

| # | Producer | Change | Reversible by |
|---|---|---|---|
| 1 | **claude-notify** | Cross-session gate: one desktop toast per window (default 600s) across **all** sessions. Held turns are counted and **named** in the next toast's `+N other turns finished: proj-a, proj-b` line. | `CLAUDE_NOTIFY_GLOBAL_COOLDOWN_SECONDS=0` |
| 2 | **cpu-monitor** | "✓ back to normal" clears demoted to the journal; per-trigger daily cap on load + runaway. **Temperature exempt and uncapped.** | `CPU_MON_CLEAR_TOASTS=1`, `CPU_MON_MAX_ALERTS_PER_DAY=0` |
| 3 | **earlyoom** | dunst `set_stack_tag` on `appname=system-notify` — an OOM episode of any size occupies one updating toast. | delete the 4-line rule |
| — | **notify-failure** | 🔴 **Untouched.** Re-proven below. | — |

`CLAUDE_NOTIFY_MIN_SECONDS` is deliberately **not** raised. Raising it deletes information (short turns vanish entirely); coalescing defers it (they still appear in the `+N` count). It also happens to be what makes the replay exact — see blind spot 3.

---

### claude-notify — the main event

Replayed with `scripts/notify-volume-replay.py`, which **imports the shipped `coalesce_decide()`** rather than paraphrasing it (pinned by a test), so the projection cannot drift from the code.

| host | before/day | after/day | reduction | peak day before → after |
|---|---|---|---|---|
| workbench | **174.0** | **40.1** | 77.0% | 386 → 70 |
| laptop | **69.6** | **28.8** | 58.6% | 182 → 56 |

Per-day, workbench: 140→47, 272→64, 386→70, 162→40, 139→30, 248→67, 348→71, 190→35.

**Conservation holds exactly**: `before(1914) == after(441) + folded(1469) + still-pending(4)`. Four turns across eleven days ended the log held and unsummarised — that is the entire measured cost of coalescing.

🔴 **#409 never measured the laptop at all.** It is a second victim at ~150/day on recent days, all `desktop=True`.

### Method, and what it is blind to

Input is claude-notify's own log — one line per turn that passed all pre-existing gates and reached dispatch. The new gate runs strictly **after** those gates and only on the desktop path, so the set of `desktop=True` lines **is** the input stream to the new gate. This is not a model of the change; it is the change run on recorded input.

1. **Dispatch, not delivery.** `dunstify`'s exit code was never checked. A toast dunst then dropped still counts. So "before" over-states what you *saw* — but it is the right measure of load *on dunst*, which is the thing being reduced.
2. **claude-notify only.** Says nothing about the other two.
3. **Exact for the gate, not for a threshold change.** Raising `MIN_SECONDS` could un-suppress a later same-session turn the per-session cooldown had hidden, and hidden turns are not in the log — so that projection would under-count. `--threshold` exists and labels its output a lower bound. This is a reason the threshold is untouched.
4. **Per-host.** Gate state is per-host; run it on each.

### cpu-monitor — #409's ~90/day is wrong; it is ~23/day

#409 derived it by subtraction and averaged across a **regime break**. Raising `CPU_MON_THRESHOLD`/`RUNAWAY_PCT` on 08-05 cut the workbench from 123–267/day to 11–32/day:

| day | 08-02 | 08-03 | 08-04 | 08-05 | 08-06 | 08-07 | 08-09 | 08-10 | 08-11 | 08-12 |
|---|---|---|---|---|---|---|---|---|---|---|
| cpu-monitor/day | 267 | 235 | 123 | 209 | 123 | **22** | **11** | **32** | **26** | **23** |

Post-break mean **~23/day** (workbench), **~13/day** (laptop) — and **two independent instruments agree**: dunst's per-notification icon warning, and dunst's own history (12/240 laptop, 15/300 workbench).

So this file gets a trim and a bound, not surgery. Projected effect: clears were a measured **6/12 of laptop** and **3/15 of workbench** cpu-monitor toasts → **~13→~7/day laptop, ~23→~18/day workbench**. The daily cap will essentially never bind at 23/day; its job is to make the 267/day regime *structurally* unreachable, because the per-episode cooldown does **not** bound the aggregate (a new episode alerts immediately, so a flapping signal is unbounded).

### earlyoom — bursty, not steady

415 notifications / 11 days, but on **three days only** (50 / 161 / 204); zero on the other eight; 111 inside one 3-minute window. Steady-state effect ≈ 0; burst-day effect **204 → ~1**.

---

## 3. What you stop learning

| Reduction | What is lost | Where it still is |
|---|---|---|
| claude-notify coalescing | Per-turn timing and identity of held turns *as separate toasts*. You get a count + project names on one line instead. | tmux window state, agent-ops dashboard (`$mod+i`), clawgate push, `~/.claude/claude-notify.log` |
| …and specifically | The last few turns before the stream goes quiet may sit held indefinitely (measured: **4 turns over 11 days**). | same as above |
| cpu-monitor clears | Desktop confirmation that a spike *ended*. | bar load/temp blocks show it continuously; `journalctl --user -u cpu-monitor` |
| cpu-monitor cap | Load/runaway alerts past #8 in a day. **Never temperature.** | journal, and the toast that *is* capped says so and names the knob |
| earlyoom stacking | Per-process detail of every kill but the newest, **on the desktop**. | `journalctl -u earlyoom` — every kill, permanently, with process/RSS/cmdline |

Nothing here is a delete. Every demoted message still lands in a journal.

---

## 4. Positive controls — reported as pairs, never as a bare zero

| Instrument | Positive control | Negative control | Verdict |
|---|---|---|---|
| Replay counter | 40 events @1200s gaps, window 600 → **before=40 after=40** (must NOT reduce) | 40 events @10s → **before=40 after=1, pending=39** | both moved |
| Replay parser honesty | garbage line → **`unparsed=1` + "UNDER-count"** flagged | — | flags, does not swallow |
| Icon-warning instrument | 4 probes with that icon → **icon warnings 0 → 4** (exactly 1 per notification) | — | calibrated, so the cpu-monitor arithmetic is sound |
| dunst stack tag | untagged arm → **displayed = 3** (observable) | tagged arm → **displayed = 1** | 🔴 this is the control #409 lacked |
| dunst config parse | bogus option → **`WARNING: Setting set_stack_tag_typo … doesn't exist`** | real config → **no warning** | `set_stack_tag` is accepted by dunst 1.13.2 |
| cpu-monitor harness | `alert critical` → **1 toast observed** | — | its zeros are readable |
| cpu-monitor prelude extraction | asserts all 5 functions found | — | not silently empty |

🔴 **#409's earlyoom failure was exactly a missing positive control**: `fullscreen_suppress` routed its probes to history, so tagged *and* untagged both read 0 — a zero from a control that never observed anything. The re-run refuses to read the tagged number unless the untagged arm moves first.

---

## 5. 🔴 notify-failure still defeats DND

Run on the laptop, **with a negative control in the same paused state** — a `displayed` bump alone could just mean the pause never took:

```
BASELINE   paused=false displayed=0 waiting=0 history=241
PAUSED     paused=true  displayed=0 waiting=0
CONTROL  (appname=dnd-control)     displayed 0->0   waiting 0->1   <-- swallowed
DEADMAN  (appname=notify-failure)  displayed 0->1   waiting 1->1   <-- shown
VERDICT: PASS
RESTORED   paused=false displayed=0 waiting=0 history=241
```

Structurally, the new stack rule **cannot** touch it: disjoint appname, and it sets no `fullscreen`/`override_pause_level` so it takes no part in the last-write-wins ordering that protects the bypass. Four tests pin this, incl. a **ledger** that fails if any *second* rule ever gains a stack tag. Rendered `dunstrc` confirms `[zz_notify_failure_bypass]` is still last.

---

## 6. Mutation battery — 21 mutants, 21 killed, **0 survivors**

Baseline confirmed green first (an already-red suite makes every mutant look killed). Killing test named for each, so none is green-for-the-wrong-reason.

Killed: boundary `>=`→`>`; inverted comparison; `held` never incremented; state never reset; suffix always empty; **default window 600→86400** ("so large it swallows everything"); **gate keyed per-process so sessions stop sharing a budget** (the seam); gate applied to the headless path; **fail-closed**; cap off-by-one `-gt`→`-ge`; clears re-enabled by default; day-rollover never resets; temperature loses its exemption; **a recovery notice bypassing `clear_alert` with its text preserved** (the commented-out shape); cap hard-disabled; **stack rule colliding with the deadman appname**; stack tag removed; stack rule joining the fullscreen ordering.

**Three initially SURVIVED and were real gaps, now closed:**

1. `window <= 0` → `window < 0` — survived because `(now-last) >= 0` is true for any forward-moving clock. The two differ only when the clock steps **backwards** (NTP, suspend/resume), where the arithmetic form wedges the gate shut *with your declared revert knob set*. Now pinned.
2. `pending[-CAP:]` → `pending[:CAP]` — the length assertion was identical either way, so the toast could have named 200 projects from hours ago. Now asserts *which* names survive.
3. Runaway call site spending `load_alerts_today` — survived because the budget test calls `alert_capped` directly and never exercised the call sites. It proved the *mechanism* could keep two budgets while the shipped code kept one. Now asserts the counter name at each site.

---

## 7. Both tiers, on the **merged** tree

Merged `origin/main` (`02af928`) in first — #427/#429 landed after this branched and touched the launcher-guard conftest and `test_runtime_shebangs.py`, which my stubs interact with.

| tier | result |
|---|---|
| `nix build .#checks.x86_64-linux.pytests` (authoritative) | **8654 collected, 8653 passed, 1 skipped (pinned), 0 failed — PASS** |
| `gate.sh --tier both` pytest | **8575/8574/1 skip/0 failed — PASS** (pre-merge tree) |
| `gate.sh --tier both` node | **1024 tests, 1024 pass, 0 fail, 0 skipped — PASS** |

`scripts/tests` 2977 (base) → 3091 (merged HEAD). **`skipped=0` for every new test.**

The sandbox tier earned its keep: it caught three hand-written `#!/bin/sh` stubs and an `exec /usr/bin/env date` fallback that pass on this NixOS host and ENOENT in the sandbox. Fixed to use `testlib.mockbin.write_exec`.

---

## 8. What I could NOT verify

- 🔴 **Nothing is deployed.** `readlink -f` puts all three files in `/nix/store` — the hook, cpu-monitor.sh and dunstrc are `home.file` **copies**, not out-of-store symlinks. This needs `scripts/ship.sh`; `git pull` alone changes nothing. **The 174→40/day figure is a projection over recorded input, not an observation of the new code in production.**
- **cpu-monitor's clear-vs-alert split** is from a dunst-history *sample* (12 and 15 entries), not a per-notification log. cpu-monitor has no log of its own — that is the same blind spot that let #409's number stand. The daily-cap bound is structural and does not depend on it.
- **The earlyoom rule was validated as a mechanism, not in situ.** I proved dunst 1.13.2 collapses a shared stack tag using the equivalent `x-dunst-stack-tag` hint, and that the rendered config parses. I have not watched a real OOM burst collapse — that needs a switch plus an OOM episode.
- **Whether 40/day is low enough is a judgement, not a measurement.** See below.
- **The workbench queue mystery is still unexplained.** `waiting` was 244 in #409 and is **48** now, while ~190 toasts/day are still being dispatched into a paused dunst. Something caps or evicts it and I did not determine what. Read-only throughout; I did not flush, dismiss, clear or unpause anything.
- **Laptop state restored exactly** (`paused=false waiting=0 displayed=0`, history 241→241). Probe entries were removed individually by id, never `history-clear`, and history never approached its 300 cap, so **nothing was evicted** (unlike #409, which lost 6 entries).

---

## 9. Recommendation

**Yes — turn DND off, but after `ship.sh`, not before.**

The arithmetic that matters: workbench goes **174 → ~40/day** from claude-notify, **~23 → ~18** from cpu-monitor, and earlyoom burst days **204 → ~1**. That is roughly **330/day → ~60/day**, and on the days that most plausibly made you reach for the bell (386 and 348) **from ~390 to ~75**. Call it one toast every ~20 waking minutes instead of one every ~3.

Honest caveat: **~60/day is quieter, not quiet.** If it still grates, the next move is one env var on the unit, no code change and no redeploy of logic:

- `CLAUDE_NOTIFY_GLOBAL_COOLDOWN_SECONDS=900` → **30/day** (measured, not extrapolated)
- `=1800` → **17.6/day**, at the cost of 23 turns/11 days ending held rather than 4

**What to watch in the first days:**

1. **That the `+N other turns finished` line is actually there.** If toasts arrive with no suffix, the gate is failing open — check `journalctl` / `~/.claude/claude-notify.log` for `coalesce gate error`. A toast that merged 40 turns and one that fired because the gate crashed look identical on screen; `coalesced=N` in the log is what distinguishes them, and it is pinned as a contract.
2. **That you still get OOM and unit-failure toasts at all.** Those are the two classes where "quiet" and "broken" are indistinguishable. `notify-failure` fires ~1/day; if a week passes with none *and* `drift-check` has been failing, suspect the pipeline, not luck.
3. **Whether the held-turn tail bites.** 4 turns over 11 days ended unsummarised. If you find yourself missing "that one finished ages ago", lower the window rather than disabling the gate.
4. **cpu-monitor's cap should never bind at ~23/day.** If `journalctl --user -u cpu-monitor` shows `capped:` lines, the flapping regression is back and the threshold — not the cap — is what needs attention.
5. **Run `scripts/notify-volume-replay.py ~/.claude/claude-notify.log` on each host in a week.** Post-change the "before" column becomes the *gated* stream, so it measures the new reality directly.

One thing I did not touch that you may want to: the workbench fired **12 `⚠ Runaway process: Farthest Frontier`** alerts — a game, tripping the runaway trigger. `CPU_MON_IGNORE` exists for exactly that (it already lists `anno`), but adding a name to it is a decision to never see something, so it is yours to make.


1200s:
您好，您的邮件已收到，谢谢您的支持，祝一切顺利！

PR #442, opened 2026-08-13
feat(subsystem-recall): reading by whole entry does not scale, and the index had no ceiling

ZacxDev:
PR 2 of 2 on the store's READ half (#440 was the digest). Adds `--search` and caps the index listing.

## Why

Both axes grow: entry count and entry size. #440 fixed the *default*, but left the two things that get worse rather than better as the store fills — the INDEX had no ceiling (it printed every entry, forever), and the only way in was a WHOLE ENTRY (`--ref` spends a full body to answer a one-line question).

## Deliverable A — `--search`, fuzzy, stdlib-only

`re` + `difflib`, **nothing is spawned**. The whole store is ~81 KB / 30 files and a full scan measures **~40 ms**, so `rg` would buy no speed while costing a nix-store binary not guaranteed on PATH plus an output format nobody pinned. The module still shells out to nothing, and that is asserted structurally.

**Two-stage scorer** (entry → block), scoring a **coverage mean** over query tokens. Three decisions worth calling out:

- **A BLOCK is the match unit, not a line.** The store is written in bullets that wrap: `nginx` on one line and `rate-limit` on its continuation is ONE fact, and a per-line scorer scores each half at 0.50 and returns nothing — the "returns NOTHING" failure the per-line two-stage prototype produced. `test_a_per_LINE_scorer_would_MISS_the_multi_line_case` proves the lines genuinely fail alone, so the block unit is not untested.
- **Compound terms are handled deliberately, not by a lowered cutoff.** A block's candidate set includes every adjacent pair joined (`rate`,`limit` ⇒ `ratelimit`), so `ratelimit` hits `rate-limit` at **1.00**. I measured what that actually buys, because the prefix rule already reaches `rate` at 0.92: (1) **ranking** — the joined form outranks blocks that merely contain one half (there is a `rate`-bearing distractor in the fixture set for exactly this); (2) **reach** — `k8sapi` → `k8s api` is impossible any other way, both parts being below the exact-match length rule.
- **A no-match carries its evidence.** It prints how many entries were scanned and *either* the closest sub-threshold candidate with the exact `--threshold` that surfaces it, *or* "no candidate scored above zero at all" — an ABSENT term, not a weak one. Different next actions (rephrase vs. lower the bar), so they must not print the same blank. A test asserts the two sentences share no phrase, and another asserts the suggested `--threshold` actually surfaces the named hunk at the named score.

**Output contract.** Every hunk restates its own `scope/ref`, section, `file:line`, `sensitivity=`, score and basis — hunk output interleaves entries, so a label printed once per invocation gets separated from the content it governs. Asserted by COUNT (`text.count("sensitivity=") == len(hunks)`), not by an `in` check a single surviving label would satisfy. Absent sensitivity folds to `client-confidential`.

`-C N` overrides the default enclosing bullet with raw lines; `--help` states the tradeoff in prose so the agent can drive it (a raw window can cut a bullet in half and emit a fragment that reads like a complete instruction). `--all-scopes`, `--threshold F`, `--max-hits N`.

### The fixture-set gate (the scorer is judged by measurement, not taste)

Labelled SYNTHETIC corpus, 8 entries, every name invented. One query per match class:

| class | query | rank 1 | score | basis |
|---|---|---|---|---|
| exact | `compaction` | `blob-vault` ✅ | 1.00 | line |
| typo | `conection pool` | `ledger-db` ✅ | 0.97 | line |
| compound | `ratelimit` | `edge-proxy` ✅ | 1.00 | line (distractor `mailer` at 0.92) |
| compound, short parts | `k8sapi` | `beacon` ✅ | 1.00 | line |
| multi-token across LINES | `quill scheduler` | `quill` ✅ | 1.00 | line |
| several entries, one relevant | `blob-vault` | `blob-vault` ✅ | 1.00 | line (tie broken by name score) |
| entry name only | `zephyr` | `zephyr-cache` ✅ | 1.00 | **entry-name** |
| **absent term** | `kryptonite` | — | — | **0 hunks** |

**The control pair, reported together** (`test_the_whole_set_scores_and_the_PAIR_is_reported`):

```
positive control: 7/7 labelled queries returned a hunk, 7/7 correct at rank 1;
negative control: 0 hunks for an absent term
```

A zero alone would be indistinguishable from a searcher wired to nothing, so the assertion message carries both halves.

Every tuning constant (`DEFAULT_SEARCH_THRESHOLD` 0.60, `FUZZY_FLOOR` 0.82, `MIN_INEXACT_LEN` 4, the prefix/substring strengths) is answerable here and nowhere else. 0.82 and not 0.80 because 0.80 is exactly where a one-character substitution in a five-letter word lands — and so does `probe`/`prone`, two real words that are not a typo of anything.

## Deliverable B — the index listing is capped at 100

`--page N`, 1-based. **Newest-first by entry-file mtime, tie-broken by ref, and the header says so** — that order is the point: cutting an ALPHABETICAL index hides entries by an accident of their names, cutting a RECENCY index hides the stale ones, which is the only cut a resuming session can reason about.

- Below the cap the wording is **unchanged** (`ALL N entries … none omitted`), so the page notice does not become background noise on the day it matters.
- Above it: `entries 1–100 of 105 … (page 1 of 2)` plus `(… 5 more entries NOT LISTED on this page … --page 2 lists the rest, oldest last.)`
- A page past the end **lists nothing and says so with the valid range** — it does not clamp. Clamping answers a question the caller did not ask under a heading claiming otherwise.
- It is a LISTING cap, not a selection rule: `total_in_scope`, `omitted` and the featured pick are untouched.
- `--page` with `--ref`/`--limit` is rejected (they print no index), not silently ignored.

⚠ **One behaviour change to existing output:** the index order moved from ref-ascending to mtime-descending. `full` mode's bodies keep the canonical ref order. Three existing tests were updated for it, and the fixtures now PIN mtimes so the order is deterministic by construction rather than by how fast the loop wrote 15 files.

## Measured, on `datapacket-talos` (25 entries), 2026-08-13

| invocation | bytes | ~tokens | wall |
|---|---|---|---|
| digest (default, unchanged by this PR) | 4,876 | 1,219 | 2.4 ms |
| `--list` (unchanged) | 2,580 | 645 | 1.8 ms |
| `--search minio` | 5,267 | 1,316 | 39 ms |
| `--search 'nginx ratelimit'` | 1,711 | 427 | 95 ms |
| `--search 'postgres conection pool'` | 2,318 | 579 | 137 ms |
| `--search minio -C 2` | 7,285 | 1,821 | 39 ms |
| `--search minio --all-scopes` | 5,385 | 1,346 | 43 ms |
| full-store scan, all 29 entries | — | — | 40 ms |

`DEFAULT_MAX_HITS` is **10** and that is measured, not chosen: at 20, `--search minio` cost 9,386 B — *twice* the digest it is supposed to be cheaper than.

The digest / `--list` numbers differ from #440's by ~30 B only because the index header now states its order.

## Consolidations (not additions)

- `load_store()` — recall and search opened the same store for the same reasons; that predicate is now in one place. Its `verb` is the only thing that differed, and it has to: "nothing was recalled" vs "nothing was searched" must not read as "nothing recorded yet".
- `caveat_text()` — a module function now that there are two report types. `test_kills_the_caveat` neuters it once and asserts BOTH renderers go quiet; had search grown its own copy, that kill would have gone green while the search block kept presenting recall as a plain finding.
- `--search` / `--ref` / `--list` are three selectors over one store and every pair is the same conflict — one guard over `_SELECTORS`, not three pairwise `if`s.

## Invariants held

READ-ONLY (the store tree is hashed either side of every search mode and every search failure path), no network, nothing spawned, `MalformedEntryError` still fail-closed, non-zero exit still non-blocking, and `--ref` / `--limit` / `--json` / `--scope` / `--store` / `--repo` / the #440 digest all smoke-tested against the real store.

## Red-at-base / green-at-HEAD matrix

Measured against a pristine `git archive origin/main` tree with this PR's test file dropped in. **110 red at `origin/main`, 310 green at HEAD** (318 after the doc-pin additions).

**8 BEHAVIOURAL reds** — assertions on a wrong value from code that existed at base:

| test | what it caught at base |
|---|---|
| `TestTheIndexPageCap::test_a_page_holds_at_most_LISTING_PAGE_SIZE_lines` | `the index printed every one of the scope's entries — it is uncapped` |
| `TestTheIndexPageCap::test_the_remainder_is_COUNTED_and_the_flag_NAMED` | `'NOT LISTED on this page' not in output` |
| `TestTheIndexPageCap::test_the_order_is_NEWEST_FIRST_and_the_output_SAYS_so` | `'gizmo-001' == 'gizmo-105'` |
| `TestSelectionIsScopeWide::test_order_is_canonical_and_STABLE` | `the index is not newest-first` |
| `TestTheDefaultIsCompleteAndCheap::test_the_json_carries_the_index_without_the_bodies` | listing order inverted |
| `TestStatusIsTheDiscriminator::test_the_search_statuses_do_not_reuse_a_recall_word` | declaration-level: `search-no-match` absent from `STATUS_PRECEDENCE` |
| `TestSearchAndPageCli::test_the_help_explains_WHY_the_bullet_is_the_default_context` | documentation-level: the `-C` tradeoff prose absent |
| `TestSearchAndPageCli::test_the_help_states_the_page_ORDER` | documentation-level: the page order unstated |

The first two were deliberately re-ordered so their **first** assertion is spelled against attributes that existed before this change — a test whose first failure is `AttributeError: no attribute 'listing_total'` proves the surface is new, not that a behaviour regressed.

**102 SURFACE-ONLY reds**, counted and labelled separately: `AttributeError: module has no attribute` / `unexpected keyword argument` / `object has no attribute` (66), argparse `SystemExit: 2` from an unrecognised flag (14), `mutation anchor occurs 0x` because the code being mutated does not exist at base (22).

🔴 **Honest read: `--search` is entirely NEW SURFACE and cannot have a behavioural red.** Its correctness rests on the fixture-set gate and the mutation kills, not on a base comparison. The pagination half is where the genuine behavioural reds are.

## Mutation results — 21 new kills (43 in the matrix, all green)

Each breaks ONE thing and asserts THAT thing's own symptom, with the real module's opposite behaviour asserted in the same test so a kill cannot pass on a neighbour's error.

**Scorer** — coverage mean → max (`compaction kryptonite` scores 1.00 and returns a hit); concatenation removed (`k8sapi` becomes unreachable); `MIN_INEXACT_LEN` 4 → 1 (`pod` fuzzily reaches `podman`, killed with a real corpus token rather than an invented one); `FUZZY_FLOOR` 0.82 → 0.70 (`probe`/`prone` matches); name-only hit disabled (an entry searched for BY NAME returns nothing — reachable only because `zephyr-cache`'s body never says `zephyr`); name-score tie-break removed (`mailer` outranks `blob-vault`, and the kill is built so alphabetical order alone gives the wrong answer).

**Parser** — front-matter exclusion removed (`service:` becomes a block, the exact prototype failure); heading exclusion removed (a query for a schema heading's words then hits every entry in the store).

**Output** — per-hunk `sensitivity=` removed (counted, not `in`-checked); `-C` override disabled (the flag goes inert while every hunk still looks plausible); `best_below` removed *and* the zero-score filter removed (the two no-match sentences, killed in both directions); hunk truncation notice removed.

**Pagination** — page slice removed (105 lines again); page notice removed, with the past-the-end arm asserted *still firing* so the kill is about the notice and not the block; past-the-end notice removed; `page` sanity guard removed (reached with a valid `limit`, so the guard above it cannot be what fires); index ordering direction flipped.

**Guards** — empty-query guard; threshold-range guard (reached with a valid query); search `scope-absent` status; and the four CLI conflict guards killed **separately**, each proven with 1–3 argv shapes and each asserting the real module still rejects them.

⚠ **`test_kills_the_deterministic_ordering` was silently going green for the wrong reason** once the index gained its own ordering site: the reverse-ref mutation no longer touches the listing, and the surviving assertion passed off the mtime order it never mutated. Split into two kills, one per ordering site.

## What I could NOT verify

- **`/resume` end to end.** The doc changes are pinned by `TestSkillDocsArePinned` (8 new pinned sentences) and every flag the doc names is asserted to exist in the parser, but nobody has run `/resume` against this and watched an agent choose `-C`. That is a behavioural claim about an LLM executor and it is not made here.
- **The scorer on a scope big enough to need it.** The largest real scope is 25 entries. Everything about growth — that hunks stay readable at 200 entries, that 100/page is the right cap — is reasoning from a measured cost curve, not an observation.
- **0.60 as a general threshold.** It is gated on 7 labelled queries. It is right for those; it is not a claim about every query anyone will type. A no-match names the exact `--threshold` that would surface the near miss, which is the escape hatch for exactly that uncertainty.
- **Not deployed.** `home-manager switch` / `ship.sh` has not been run, so `~/.claude/skills/resume/SKILL.md` is still the old copy on both hosts until it is.

## Gate

`nix build .#checks.x86_64-linux.pytests --no-link --print-build-logs` → **`RESULT: PASS (exit=0)`**

```
PASS  scripts/tests  (collected=3679 passed=3679 skipped=0 floor=3428)
TOTAL collected=9242  passed=9241  skipped=1  failed=0  (floor: 8577 = sum of 17 per-target floors)
```

The one skip is the pre-existing opt-in `repo-cos` live-store drift check. No per-target floor needed changing — `scripts/tests` went 3,564 → 3,679 (+115 in this file: 203 → 318), and a floor is a minimum.


ZacxDev:
All three blockers fixed in `c8b9f8d`, both nits actioned. Thank you for building a 150-entry scope — every one of these lived past page 1, and every pagination test I wrote asserted on the report **object** for pages after the first and only ever rendered page 1. These are rendered-**text** defects, so the tests could not have caught them.

I reproduced all three verbatim before touching anything:

```
===== page 2 (of 2) =====
INDEX (from index) — entries 101–150 of 150 in `bigscope/`, newest-first by file mtime (page 2 of 2):
(… 100 more entries NOT LISTED on this page — … `--page 3` lists the rest, oldest last.)
NO ENTRY BODIES WERE PRINTED (--list). The 150 entries above are the complete index for `bigscope/` …
===== page 9 =====
INDEX (from index) — entries 801–800 of 150 in `bigscope/`, newest-first by file mtime (page 9 of 2):
```

## 1 — false truncation notice on the last page

Root cause: the notice was gated on `listing_total > len(listing)` — which means *"this page is not the whole index"*, and is TRUE on the last page. It never asked the question a truncation notice is actually about.

Now gated on `RecallReport.listing_after_page` — rows on **later** pages, 0 on the last one:

```python
@property
def listing_after_page(self) -> int:
    if self.page_is_past_the_end:
        return 0
    return max(0, self.listing_total - self.listing_before_page - len(self.listing))
```

Page 2 of 2 now prints no notice at all. **Page 1 is byte-for-byte unchanged** — the fix narrows *when* the notice fires, never what it says (`test_page_1_of_a_capped_index_is_UNCHANGED` pins all six of its phrases).

⚠ One thing your report didn't ask for but the fix needed: `test_a_MIDDLE_page_still_announces_what_follows` builds a **three**-page index and checks every page. Without it, the wrong fix — gating on `page == 1` — passes every other test in the class.

## 2 — unguarded past-the-end arithmetic

The real problem was that *"is this page past the end?"* was answered separately in each renderer branch, which is exactly how a header contradicted a correct line two rows below it. It is now one property, `page_is_past_the_end`, and the header has its own branch that computes no range:

```
INDEX (from index) — no entries: page 9 is past the end of `bigscope/`, which holds 150 in 2 pages (newest-first by file mtime):
  (PAGE 9 IS PAST THE END — … re-run with `--page 1` … `--page 2`.)
```

Your guidance line is unchanged, as you asked.

## 3 — `--list` completeness line

Now describes the page it is on:

- page 1 of 2 → `The 100 entries above are page 1 of 2 of the index for \`bigscope/\`, which holds 150 in all`
- page 9 → `NO entries were listed — see the page notice above. \`bigscope/\` holds 150`
- **single page → unchanged** (`The 25 entries above are the complete index for …`). It's true there, and weakening it everywhere is how a notice becomes background noise. `test_a_SINGLE_page_scope_keeps_the_complete_index_claim` is the control.

`NO ENTRY BODIES WERE PRINTED` and `this is not an empty scope` survive in all three branches.

## 4 — `N bullets` → `N nuance`

You're right that it reads as entry size. Changed, and it's a character shorter so it costs nothing. The count itself stays nuance-only and the docstring now says why: pointers are durable and don't grow, while work-history is what the store's prune-on-resolve discipline is denominated in — so it's the number that predicts whether a `--ref` is worth spending. `test_the_count_is_the_NUANCE_section_and_not_the_whole_entry` builds your exact 5-pointer/7-nuance case.

## 5 — `sensitivity: internal` silently rewritten

Deliberate **and** it was a real gap, so I did both: kept the fold and made it audible.

The asymmetry you spotted is the schema's, not a bug — `public` is honoured because it is one of three known values; `internal` is not a schema value at all, so the fail-safe takes it. What was wrong is that the override was **silent**: a marker somebody *wrote* and a marker that was *absent* rendered identically, so the file's own claim vanished and whoever typed `internal` got no signal the schema doesn't know the word.

Now: `client-confidential (declared: internal)` on the index row, the entry header and the search hunk. An absent marker is still unannotated — nobody claimed anything. The fold is untouched (`test_the_fold_itself_is_UNCHANGED` sweeps it), and `discarded_sensitivity` is **derived from** `fold_sensitivity` so the two cannot drift about which values are known.

## Red-at-reviewed-head / green-at-HEAD

Measured against `git archive 826a5ab` — the exact commit you reviewed — with the new test file dropped in. **26 red at `826a5ab`, 343 green at `c8b9f8d`** (318 → 343 tests).

**12 BEHAVIOURAL** (assertion on wrong output from code that existed), 7 of them the blockers:

| test | message at 826a5ab | defect |
|---|---|---|
| `test_the_LAST_page_prints_NO_truncation_notice` | `the last page announced a truncation that is not there` | 1 |
| `test_the_last_page_NEVER_routes_to_a_page_that_does_not_exist` | `the last page routed past the end` | 1 |
| `test_the_digest_pages_the_same_way` | `'NOT LISTED on this page' not in …` | 1 (digest mode) |
| `test_no_page_CONTRADICTS_its_own_header` | `page 2 listed the OLDEST entry and still claimed more follow it` | 1, generically |
| `test_a_page_past_the_end_prints_NO_inverted_range` | `the header inverted its own range` | 2 |
| `test_the_list_completeness_line_describes_THIS_PAGE` | `the closing line claimed a completeness the page does not have` | 3 |
| `test_the_completeness_line_past_the_end_is_not_self_contradictory` | `it described rows it never printed` | 3 |
| `test_the_count_NAMES_the_section_it_counts` + 3 more | `1 bullet` / `7 bullets` | nit 4 |
| `test_the_index_row_shows_the_override` | `'client-confidential (declared: internal)' in '… client-confidential'` | nit 5 |

**2 documentation** (the two new `SKILL.md` pins). **12 surface-only**: 6 `AttributeError`/`KeyError` on `discarded_sensitivity` / `Hunk.declared_sensitivity` / `listing_before_page` / `listing_after_page`, and 6 `mutation anchor occurs 0x`.

⚠ `test_no_page_CONTRADICTS_its_own_header` initially reported as *surface-only* — its behavioural sweep and its property check were interleaved, so it died on `AttributeError` at page 1 and never reached page 2 where the contradiction lives. Split into two loops, behavioural first, so the red it produces says "the output is wrong" rather than "the surface is new". That's the same trap your review is about, one level up.

## Mutation results — 46 kills, all green (3 new, 2 rewritten)

- **`test_kills_the_past_the_end_PREDICATE`** — mutating `page_is_past_the_end` to `False` reproduces **defect 2 verbatim**, and the test asserts that specific output rather than "something is missing": `entries 801–800 of 105`, `(page 9 of 2)`, and `The 0 entries above are page 9 of 2`. It also proves all three branches hang off the one property.
- **`test_kills_the_past_the_end_HEADER_branch`** — the header's branch killed *separately*, asserting the **invariant** (`hi < lo`) rather than the literal numbers, because `listing_before_page` still guards past the end so this mutant inverts to `1–0`. Pinning `801–800` would have made the test a claim about *which* of the two guards is missing.
- **`test_kills_the_list_completeness_line_being_PAGE_SCOPED`** — reinstates defect 3, with the past-the-end arm above it asserted still firing so the kill is about the right branch.
- **`test_kills_the_page_truncation_notice`** — re-anchored onto `listing_after_page`.
- **`test_kills_the_overridden_sensitivity_ANNOTATION`** — asserts the fail-safe **still fires in the mutant**, so the kill is about visibility and not about the safety property.

## Gate — the pushed commit specifically

You were right to flag that. `HEAD == origin/…` and `git status --porcelain --untracked-files=no` is empty, so the flake source is exactly `c8b9f8d`. Derivation `wqppw4bsv93lygqnvxl2jxgd2ccxm45d-devrc-pytests.drv`, verified by reading `nix log` for that drv path rather than the build's stdout:

```
PASS  scripts/tests  (collected=3704 passed=3704 skipped=0 floor=3428)
TOTAL collected=9267  passed=9266  skipped=1  failed=0  (floor: 8577 = sum of 17 per-target floors)
RESULT: PASS (exit=0)
```

The one skip is the pre-existing opt-in `repo-cos` live-store drift check. No floor change needed.

## Still not verified

Unchanged from the original PR: `/resume` has not been run end to end, and nothing has been deployed (`ship.sh` / `home-manager switch` not run, so both hosts still have the old `SKILL.md`). New to this round: the >2-page shape is tested but has never existed in the real store, whose largest scope is 25 — which is precisely the blind spot that produced these three defects, so I'd rather say it than imply otherwise.

PR #447, opened 2026-08-13
docs(rules): two multi-agent lessons that never landed, recovered from a merged branch

ZacxDev:
## What

Two rules recovered from `zach/rules-multiagent-lessons` (`ea82146`), plus their archive
entries. Both were absent from **`claude/RULES.md` and `claude/RULES-ARCHIVE.md` alike**.

- 🔴 **`cross-repo-worktree`** — `isolation: "worktree"` builds the worktree from the
  dispatching session's **current** repo, not the repo the task names. Dispatch a fan-out at
  repo B while sitting in repo A and every agent gets a worktree of **A**, with no error: the
  worktree is created successfully and `git status` inside it is clean. It surfaces one
  indirection away, as an agent reporting that a file named in its own brief does not exist.
- 🔴 **`sibling-agent-kill`** — under a parallel fan-out, confirming `/proc/<pid>/cmdline`
  before a kill is **not sufficient**: a sibling agent running the same suite matches it
  legitimately. One auditor cleaning up its own hung run killed ~15 PIDs and destroyed another
  agent's in-flight test run, whose next attempt collapsed with `0 files collected` and
  `exit 144` — both readable as code defects. `/proc/<pid>/cwd` is the discriminator.

## How these were found, and why they were nearly dropped

While diffing the 23 preserved worktrees for keep-or-drop. The verdict table landed separately
in #446 (`claudedocs/handoff-subsystem-store.md`): **22 DROP, 1 KEEP (#355), 1 salvage** —
this one.

`ea82146` is the branch of **#313, which merged**. It therefore looks fully absorbed, and the
"superseded intermediate" reading would have dropped it.

The measurement that classified the other 22 — *what fraction of the lines the branch itself
added are present in main today* — scores this branch at **1%**, which means nothing for
`RULES.md`, a file that is continuously reworded. Re-measured by **archive anchor** instead,
20 of its 22 lessons are in main and exactly these two are not.

Deliberately **not** salvaged, having checked: `cheap-control` and `count-tests` anchors are
gone but their rule text is in main (`count-tests` → `count-not-exit-code`), and
`stale-gate-base` survives compressed inside the `merged-tree` bullet as "check how far behind
main a PR is".

## Size

`RULES.md` 31,726 → **32,727** bytes against a **35,200** ceiling requiring **900** headroom —
1,573 bytes to spare, so **no eviction was needed**. Numbers read from
`scripts/tests/test_rules_size.py`, which owns them. `RULES-ARCHIVE.md` is not auto-loaded and
has no ceiling.

## Verification

`nix build .#checks.x86_64-linux.pytests` on this tree:
`TOTAL collected=9267 passed=9265 skipped=1 failed=1`.

🔴 **Read that honestly: the suite is not clean, and I am not claiming my change is why.** The
one failure is `test_agent_def_denies_bash_allows_only_custom_tool`
(`scripts/browser-bridge/tests/test_browser_agent.py`), a 60s subprocess timeout whose own
stderr says it could not warm its isolated opencode config dir and needs to
`npm-install @opencode-ai/plugin` **with network available**. `test_rules_size.py` — the gate
that actually governs this change — passed.

A cached `origin/main` build shows `failed=0`, but that is not a control: cached means it ran
under whatever network conditions held then, versus my fresh run now. A forced `--rebuild` of
the baseline drv is running to make the comparison discriminating; **result appended below
before this merges.** Three markdown files cannot plausibly reach a browser-agent subprocess,
but a theory that explains a failure is not evidence for it.


ZacxDev:
## Corrections to the PR body above — an audit found three claims wrong

Posting rather than silently editing, since the body has been readable since it was opened.

**1. "20 of its 22 lessons are in main" is wrong — neither number reproduces.**
`ea82146:claude/RULES.md` carries **27** distinct `→ archive:` anchors, of which **22** are
present in `main` and **5** are not. Of those 5, three survive as text under a changed or absent
anchor (`cheap-control` → `claude/RULES.md:55`; `count-tests` → `count-not-exit-code` at `:61`;
`stale-gate-base` compressed into the `merged-tree` bullet at `:60`), leaving the two this PR
recovers. So the conclusion — exactly two genuinely lost — is unchanged and independently
confirmed, but the arithmetic I quoted was not.

**2. "recovered from `ea82146`" was accurate for the RULES.md bullets and NOT for the archive
sections.** Those were rewrites. They dropped `Measured 2026-08-02`, the repo pair
(`civit/civitai` dispatched from a cwd in `civit/cli`), the kill pattern
(`chrome-headless-shell|vitest|steam-run`, system-wide, ~15 PIDs), n=2, and the second failure
mode — i.e. precisely the evidence the archive exists to hold, while reading as though it were
that evidence. Two sentences were outright invention; one ("the loop can run several times
before anyone asks") is contradicted by the measurement it sat next to. Fixed in `ff9589b`:
both sections are now verbatim, and the recovered rule is no longer narrower than the original.

**3. The gate failure's stated cause was the wrong mechanism.** I attributed it to the npm/network
warning in the test's stderr. That warning is echoed *because* the test failed — `pytest -q` only
surfaces captured stderr on failure — so it is a symptom, not the cause. The failure is
`subprocess.TimeoutExpired … timed out after 60 seconds`, `returncode: -9`: a **load flake** in a
test that wraps a 60s timeout around an npm-shelling subprocess, on a box that had 3–4 concurrent
`devrc-pytests` builds in flight. Its absence from the passing logs proves nothing either way,
which makes that an empty result rather than a control.

### The open item in the body is now closed, and it exonerates the PR

Baseline is **sandwiched** around the failure — same collected count, tree identified by the
`RULES.md` byte size in each derivation's source path:

| derivation | source tree | result |
|---|---|---|
| `wqppw4b…` | 31,726 B — `main` | `collected=9267 passed=9266 failed=0` · **PASS** |
| `xy6iffw…` | 32,727 B — this PR | `collected=9267 passed=9265 failed=1` · **FAIL** |
| `ia47nkq…` | byte-identical to `origin/main` | `collected=9267 passed=9266 failed=0` · **PASS** |

Corroborated structurally: `test_browser_agent.py` never reads `claude/`, and this PR touches
only `claude/RULES.md` and `claude/RULES-ARCHIVE.md`.

Per this repo's own `flake-vs-assertion` rule a flaky test is fixable rather than re-runnable —
the 60s timeout is a real timing dependency on a shared box. Pre-existing and out of scope here;
worth its own issue.

### Also fixed in `ff9589b`, from the same audit

- **`/proc/<pid>/cwd` was overstated.** This harness nests agent worktrees at
  `<dispatching-repo>/.claude/worktrees/agent-<id>`, so the obvious prefix check against the repo
  root matches *every* sibling and discriminates nothing — compare the exact path. Worse, agents
  sharing the base clone (the read-only ones the adjacent rule says not to give a worktree) have
  identical cwds, so the discriminator does not exist for them — and the original incident's
  perpetrator was an auditor. Both caveats now stated.
- The claim "confirming `cmdline` is not sufficient" was never measured — the auditor
  pattern-matched system-wide instead of running resolve-then-confirm. Sound inference, now
  labelled as inference.
- `cross-repo-worktree` is an 11-day-old harness measurement re-landed without re-verification,
  now labelled, with the corroboration that does hold today (the `Agent` schema exposes no
  target-repo parameter).
- New bullet reconciled with the "Mandatory" isolation rule above it; missing
  `worktree-not-session` TOC entry added; `sibling-agent-kill` moved to the TOC cluster its
  section actually lives in.

**Size after the fix round:** `claude/RULES.md` 32,727 → **33,289** B, against a 35,200 ceiling
requiring 900 headroom — **1,011 spare**. All 5 tests in `scripts/tests/test_rules_size.py` pass,
including `test_every_archive_pointer_resolves`.

One thing I am *not* claiming to have fixed: the audit notes this repo's own size module says the
answer to the next real rule is consolidation, not another bump, and the scattered worktree
bullets it names just went from four to five. That is a fair criticism of this PR and I have left
it standing rather than bundle a consolidation into a recovery.

ZacxDev:
## Correcting my own correction — the cwd caveat was wrong about its own incident

A delta re-audit of `ff9589b` found that the caveat I added in the previous round — and repeated
in the comment above — overstates its scope and contradicts the incident it describes. Fixed in
`fe315e4`.

**What I got wrong.** I wrote that the `/proc/<pid>/cwd` discriminator "does not exist" for
base-clone-sharing agents "— and the perpetrator above was an auditor", which reads as: cwd
filtering would not have prevented the recorded incident. It would have. The restored `ea82146`
text says the victim died *"in another worktree of the same repo"*. Perpetrator in the base clone,
victim in a worktree → **the two cwds differ** → exact-path comparison spares it. What kills the
victim is the *prefix* check, which is precisely what caveat (a) diagnoses. My own caveat (a)
already contained the refutation of my caveat (b), two sentences apart.

So the real scope is narrower: cwd fails only between two agents that **both** sit in the base
clone. Also corrected the premise it leaned on — `claude/RULES.md:94` says read-only agents
*"don't need"* a worktree, which is permissive; I had paraphrased it as a prohibition.

**The consequence was worse than the overstatement.** `claude/RULES.md:104` ended up reading
*"there, don't kill by pattern at all"*, aimed at read-only agents in the base clone — i.e. at
exactly the actor in the incident, whose stated need was *"to clear **its own** hung browser/test
run"*. A prohibition with no substitute, for the one case the archive entry exists to describe.

Replaced with a filter that is deterministic and **strictly stronger than cwd in every case**,
including the one (b) was reaching for: kill only your own descendants — walk `PPid` in
`/proc/<pid>/status` back to your own pid, or track the pid you launched. Verified `PPid` is
populated on this host.

**Size:** `claude/RULES.md` 33,289 → **33,356** B; 944 spare against the 34,300 budget. All 5
tests in `scripts/tests/test_rules_size.py` pass — and the re-audit negative-controlled each of
them (forcing the ceiling below current size, the headroom above it, a missing file, a dangling
anchor), so the 5/5 is not a vacuous green.

The re-audit also independently reproduced the **27 → 22 present / 5 absent** anchor count from my
previous comment under all four ways of counting, so that correction stands.

ZacxDev:
## Retracting the fix I recommended in the comment above

My previous comment endorsed *"a filter that is deterministic and **strictly stronger than cwd in
every case** … kill only your own descendants — walk `PPid`"*. **That is measured false and has
been removed from the branch.** Correcting publicly, since it has been readable since I posted it.

**Why it's wrong — two independent ways, both reproduced on this host:**

- A `nix build` an agent launches does its real work in a **`nix-daemon` child of PID 1**, not in
  the agent's descendant tree. All 9 live workers here have `PPid=18717`, and `18717` itself has
  `PPid=1`. So a `PPid` walk from the agent returns the **empty set** for exactly the hung-run case
  the rule is about.
- Each Bash call is a **fresh `zsh -c`** (four calls → four distinct pids). A stray from an *earlier*
  tool call is not a descendant of the current one — and "clear its **own** hung run from earlier"
  is the incident's own stated need.

It also collided with `claude/RULES.md:14`, which mandates killing an **orphan** by resolved PID —
an orphan is PPid 1 by construction, so never anyone's descendant.

Net: the advice I removed (fail-safe: don't pattern-kill here) failed **safe**; the replacement I
recommended fails **empty**, while looking like it closed the gap. cwd and descendant filtering are
**incomparable, not ordered**. Fixed across `0679fe0` / `fa1c54f`: cwd is primary, a descendant walk
only narrows, orphan hunts are scoped to the deploy rule, and the fail-safe is restored.

## Two more corrections from rounds 3–4

**The trim cut a load-bearing qualifier.** After trimming the bullet (to move my own error trail to
the archive) the core no longer said cwd **cannot** separate two agents that both sit in the base
clone — a configuration `RULES.md:94` explicitly permits. That re-opened the original incident:
both agents match the exact-path check, sibling gets killed, and the fail-safe never fires because
the reader *is* confident. `scripts/tests/test_rules_size.py:128-133` forbids exactly this — trimming
a "not just X" qualifier to buy bytes. Restored in `fa1c54f`.

**"Including Chrome's zygote" was invention inside a "Measured" paragraph.** Measured now: every
Brave `--type=zygote` process on this host has a real parent (`2869800`/`2869802` ← `2869781`;
`2869804` ← `2869802`); only `chrome_crashpad` sits at PPid 1. The zygote does **not** re-parent.
Struck, and the archive entry now records that the earlier assertion was wrong. The conclusion is
unaffected — the `nix-daemon` measurement and the per-tool-call mechanism each establish the empty
walk on their own, and it is the per-tool-call shell (not zygote re-parenting) that actually orphans
a `chrome-headless-shell` from an earlier call.

That is the **fourth** instance in this PR of one error class: stating an inference as an
observation. It is now logged in the archive entry itself, not just in a commit message.

## State

`claude/RULES.md` 33,300 B — 1,000 spare against the 34,300 budget. All 5 tests in
`scripts/tests/test_rules_size.py` pass, negative-controlled on all five plus a positive control on
the pointer test, so the green is not vacuous. `ea82146`'s recovered bodies remain byte-identical
through every round (`cmp`-verified). Gate green on `388ddac` (re-run), `ff9589b`, `fe315e4`,
`37dbbe8` — with `main` passing either side of the single early failure, which was a load flake in a
60s-timeout test.

PR #449, opened 2026-08-13
fix(subsystem-index): one wrapped `aliases:` line cost the whole scope, and nothing in the write loop parsed the bytes

ZacxDev:
## The blast radius, before and after

Measured on a synthetic store (`scope-a/` holding `alpha.md`, `beta.md`, and one entry whose `aliases:` flow list is wrapped across two physical lines), at `origin/main` and at this branch:

| store | `--list` rc at base | good entries listed at base | rc here | good entries listed here |
|---|---|---|---|---|
| 2 good | 0 | 2 | 0 | 2 |
| 2 good + 1 malformed | **3** | **0** | **0** | **2**, plus a `MALFORMED` row naming the third |
| 1 malformed, nothing else | 3 | 0 | 3 (`scope-unreadable`) | 0, and it says *why* |
| genuinely empty scope | 0 | 0 | 0 (`scope-empty`) | 0 |

At base, `--list`, `--ref`, the digest and `--search` all died together on the same raise. Here every selector survives, and the reject is reported per entry with its reason:

```
🔴 MALFORMED — 1 entry file in `scope-a/` could NOT be indexed and is therefore absent
   from EVERYTHING below (the index, --ref, --search). Not dropped, not hidden: listed here, once each.
  malformed index entry `scope-a/gamma.md`: `aliases:` must be a list, not a bare string
  (A STORE DEFECT, not an absence of content. Front matter is parsed LINE BY LINE, …)

INDEX (from index) — the 2 READABLE entries in `scope-a/` … NOT complete: 1 further file
   in this scope could not be indexed and is named above, never here:
```

**The report is the point, not the degradation.** Silently skipping a bad entry would be strictly worse than the collapse it replaces — a dropped entry is indistinguishable from one nobody ever wrote. So the block renders on **every** status, before anything else, and the index header stops claiming `none omitted` the moment a reject exists (both the `_render_listing` header and the `--list` completeness line get their own wording, sharing no phrase with the clean one; the clean branch keeps its exact historical bytes so nothing pinned to it moved).

---

## Exit-code decision, against `/resume`'s branch

`/resume` step 4's contract: *"If it exits non-zero … print the stderr line verbatim, note that recall was unavailable, and continue the resume."*

**Rule chosen: content served ⇒ 0; nothing readable at all ⇒ 3.** One decision site, `_exit_for`, gated on `UNREADABLE_STATUSES` rather than on any status by name.

- **Partial read ⇒ 0.** A non-zero here would make `/resume` discard every entry the run *did* surface, in order to report a defect the output already names loudly and in band. "Recall was unavailable" would be false: it was available, and it was also honest about its gaps.
- **Nothing readable ⇒ 3, not a new code.** 3 is already "the store is broken" for this CLI (`store root not found`, `index entry unreadable`), the skill's documented handling is identical for all of them, and a fourth exit code would need every consumer to learn it before it changed any behaviour — a declaration no code path honours.
- **stdout carries the per-entry rows; stderr carries exactly one line.** The detail is the whole point, and the skill says to print the stderr line *verbatim* — so that line is one quotable sentence, asserted to be exactly one line by test.

**Every-entry-malformed stays distinct from empty.** New statuses `scope-unreadable` / `search-unreadable`, checked **before** `--ref` and before `scope-empty`:

- `scope-empty` opens `NOTHING RECORDED YET` and exits 0 — `/resume` reports it as an ordinary non-finding.
- `scope-unreadable` opens `NOTHING COULD BE READ`, says `NOT an empty scope`, and exits 3. The two share no phrase (asserted).
- `--ref` into an unreadable scope is **not** `ref-absent`: "nothing recorded under that name yet" is a claim about the store, and it is false when the entry exists in a file the loader refused. Where a scope is only *partly* broken, `ref-absent` gains a clause pointing at the rejects.

---

## What changed

**1 (reader) — `scripts/lib/subsystem_resolver.py` + `subsystem_recall.py`**
`load_index`/`build_index` gain `on_malformed=` (`ON_MALFORMED_RAISE` is the **default**, so every existing caller keeps its contract). `COLLECT` returns rejects as `MalformedEntry(scope, filename, reason)` on `index.malformed`. Both rejection sites collect — the per-entry validator *and* the duplicate-ref check in `build_index`, which is a relationship between two files; collecting only the first would leave `COLLECT` able to raise, a mode a caller cannot defend against because it looks handled. A scope holding only rejects is still **registered**, or a reader would answer `scope-absent` ("nothing recorded yet") about a directory full of unparseable content.

Rejects are attributed to their own scope, so a defect in `scope-b/` does not contaminate `scope-a/` — but it is still surfaced there as a counted line naming the other scope (not its filenames, which are client-identifying).

**2 (writer) — `subsystem_touch.py --validate [PATH]`**
With a path: that one file. Without: every entry in the scope, which is the only form that can catch a duplicate ref. Exit 0 clean / 3 malformed / 3 `index entry file not found` (its own sentinel — a missing path and an unparseable file have different fixes).

🔴 **One rule, one place.** `entry_mapping(text, *, filename, scope)` was **extracted from `load_index`**, and the validator calls `entry_mapping` + `SubsystemEntry.from_mapping` (single file) or `load_index(COLLECT)` (whole scope). No second parser exists. Guarded three ways: structurally (`load_index` must still call `entry_mapping`; `validate_entry_file` must not spell `_FRONT_MATTER`/`splitlines()`/`partition(`), behaviourally (both sides produce the identical `SubsystemEntry`), and per-file over a mixed scope (validator rejects == reader rejects).

**3 (policy) — done, and it was small.** Measured: 1 of 5 scopes has a README, while the store root and `/handoff` step 4 both call the scope README authoritative — unfollowable in 80% of cases. `governing_policy()` resolves *scope README → store-root README → neither*, and the probe prints it on every status (`policy: <path>  (<basis>)`) plus JSON. The basis is carried separately from the path because the path alone cannot say whether you are reading policy written **for this scope** or the store's general policy — mistaking the second for the first is believing a scope has spoken when it has not. **No README is generated into any scope**; that would manufacture authority.

**Skills:** `resume/SKILL.md` step 4 documents the `MALFORMED` block, `scope-unreadable`, and the corrected exit-code semantics. `handoff/SKILL.md` step 4 now sends the writer to the file the probe **named** (replacing a pin on a literal per-scope path that usually is not there) and mandates `--validate` in the same turn as a write.

---

## Red-at-base matrix

Every new/changed test was run against **unmodified `origin/main`** (`6eaeb61`, the base at the time; the branch has since been rebased onto `8083352`, whose only code-relevant change is a step-3 sentence in `handoff/SKILL.md`).

**82 red at base → 0 red at HEAD.** Split, honestly:

| | count |
|---|---|
| **Behavioural** — reached the behaviour with base-existing APIs and got the wrong answer | **33** |
| **Surface-only** — could not reach it: missing symbol, unknown kwarg, unknown flag, absent mutation anchor | 49 |

The surface-only bucket is exactly what it says: `AttributeError: module 'subsystem_resolver' has no attribute 'ON_MALFORMED_COLLECT'`, `TypeError: build_index() got an unexpected keyword argument`, `SystemExit: 2` from argparse rejecting `--validate`, and `mutation anchor occurs 0x` for guards that did not exist. **None of it is evidence that behaviour regressed** — it proves the surface is new.

The behavioural reds are the real coverage. Representative:

| test | how it failed at base |
|---|---|
| `TestBlastRadius::test_two_good_plus_one_bad_still_lists_the_good_ones` | `MalformedEntryError: … 'widget-index.md': aliases: must be a list, not a bare string` |
| `TestBlastRadius::test_every_selector_survives_one_bad_entry` | same raise — all four selectors |
| `TestBlastRadius::test_the_index_stops_claiming_it_is_COMPLETE` | same raise |
| `TestUnreadableIsNotEmpty::test_the_pair` (and 4 siblings) | same raise |
| `TestRejectsElsewhereAreVisible::*` (4) | same raise |
| `TestDegradedExitCodes::test_content_served_exits_ZERO_with_the_block_in_band` | `AssertionError: a partial read exited non-zero and threw away 4 good entries` |
| `TestDegradedExitCodes::test_search_follows_the_same_rule` | `assert 3 == 0` |
| `TestDegradedExitCodes::test_nothing_readable_exits_3_with_ONE_stderr_line` | `assert 'widget-index.md' in ''` — nothing on stdout at all |
| `TestNegativeControls::test_a_malformed_entry_DEGRADES_and_is_reported_not_raised` | the raise (this test **supersedes** `…_RAISES_the_RESOLVERS_sentinel`) |
| `TestSearchNegativeControls::test_a_malformed_entry_DEGRADES_and_the_SHORT_result_says_so` | the raise (supersedes the fail-closed version) |
| `TestGoverningPolicy::test_the_probe_PRINTS_it_on_every_status` | `policy:` line absent from the real probe output |
| `TestEntryMappingIsShared::test_the_loader_builds_its_mapping_through_entry_mapping` | `load_index` built its mapping inline |

**Labelled INVARIANT GUARDS, not counted as regression coverage:**
- `TestCollectDegrades::test_the_default_is_still_RAISE` — pins that no existing caller's contract moved.
- `TestWrappedAliasesIsTheDefect::*` — **green at base, deliberately.** They pin the *diagnosis* (the exact rejection message) and its control (the same aliases on ONE line parse fine, so the fixture fails because of the **wrap** and nothing else in it). A diagnosis pin has no reason to be red.
- `TestCollectDegrades::test_an_OSError_still_fails_CLOSED_in_collect_mode` — pins the line the degradation stops at.

**The exact reported defect is a fixture**, byte-shape faithful, in both `test_subsystem_resolver.py` and `test_subsystem_recall.py`/`test_subsystem_touch.py`. All fixtures are **synthetic** — no real name, host, path or scope appears anywhere (`devrc` is public; the store is client-confidential).

---

## Mutation results

Each guard broken on purpose; each kill asserts **this** guard's own symptom, and each mutant is shown reachable (no earlier check short-circuits it). Both sections carry a **no-op harness control** (`m_degrade_noop`, `mv_noop`) so a kill is a claim about the guard, not about the loader.

| guard | mutation | observed |
|---|---|---|
| reader loads `COLLECT` | → `load_index(store)` | `MalformedEntryError` raised again; real module still `recalled` |
| `MALFORMED` block renders | `if malformed:` → `if False:` | reject vanishes from output **while the good entries are still served** — i.e. the mutant is green on every other assertion in the file. Kills **both** renderers from one anchor (the point of a single `render_malformed`) |
| `scope-unreadable` discriminator | `if not entries and bad:` → `if False:` | broken scope reports `scope-empty` **and exits 0** — exactly the conflation `/resume` acts on |
| `search-unreadable` discriminator | `if searched == 0 and bad` → `if False` | reports `search-no-match` over a scan that walked nothing |
| withdrawn completeness claim | `elif … and n_bad:` → `elif False:` | index three files short prints `none omitted` again |
| exit-code gate | `if status not in UNREADABLE_STATUSES:` → `if True:` | `main()` returns 0 on `scope-unreadable`; real module returns 3 with the stderr line |
| scope attribution | `malformed_in(scope)` → `malformed` | a defect anywhere makes **every** scope look broken |
| validator | `return MalformedEntry(...)` → `return None` | `--validate` blesses the exact entry the reader refuses (the drift a shared validator exists to prevent) |
| `NOTHING WAS CHECKED` | `if not report.checked:` → `if False:` | prints the bare `0 malformed` over a denominator of zero |
| `--validate` exit code | `0 if report.clean else 3` → `0` | malformed entry exits 0 |
| policy precedence | `if scoped.is_file():` → `if False:` | a scope with its own README is told the store-root one governs it |

**Positive + negative controls, reported as pairs** (a reassuring zero from an instrument wired to nothing is indistinguishable from a real zero):

- `TestUnreadableIsNotEmpty::test_the_pair` — empty scope collects **0** rejects, all-broken scope collects **2**, one call apart on the same store.
- `TestValidate::test_a_clean_scope_reports_a_ZERO_WITH_its_denominator` — clean: `checked=4, malformed=0`; after breaking one: `checked=5, malformed=1`.
- `TestValidate::test_a_GOOD_entry_passes` is the negative control on the validator: one that reported every file bad would satisfy the positive test and be useless.
- `--validate` always prints `checked: N entry file(s)` beside the malformed count, and an absent/empty scope gets its **own** branch (`NOTHING WAS CHECKED … NOT a clean bill of health`) with the string `0 malformed` asserted **absent** from it.
- `TestGoverningPolicy::test_the_three_bases_share_no_spelling` — the premise of every policy assertion.

---

## Gate

Run on **the exact commit pushed** (`6bcad7f`, rebased onto `origin/main` `8083352`).

```
nix build .#checks.x86_64-linux.pytests   → RESULT: PASS (exit=0)
  PASS  scripts/tests  (collected=3783 passed=3783 skipped=0 floor=3428)
  TOTAL collected=9346  passed=9345  skipped=1  failed=0  (floor: 8577 = sum of 17 per-target floors)

nix build .#checks.x86_64-linux.nodetests → RESULT: PASS (exit=0)
  TOTAL suites=4 files=34 tests=1097 pass=1097 fail=0 skipped=0  (floor: 1080)
```

**Control on unmodified `origin/main` (`8083352`), same command, separate worktree:** `scripts/tests collected=3704`, `TOTAL collected=9267`, `RESULT: PASS`. So the **+79** in `scripts/tests` (+79 total) is this branch's, and nothing else moved. No per-target floor needed changing — the gate printed no drift instruction, and `--check-floors` passes both ways. The single pinned skip is the pre-existing opt-in repo-cos live-store check. No new files, so nothing could be silently omitted from the flake; `git status` is clean apart from the untracked `.envrc` this worktree needed.

---

## Invariants held

- **Reader stays READ-ONLY.** `test_the_module_never_opens_anything_for_writing` (structural) plus tree-hashing either side of every degraded mode, every failure path, and both `--validate` forms. No network, no subprocess.
- **The live store was never touched.** Everything ran against synthetic stores via `--store`.
- **#442 pagination + loud truncation** and **#440's digest default** are untouched; the clean-path index header keeps its exact historical bytes.
- **Per-entry `sensitivity=`** unchanged; absent still means `client-confidential`.

## What I could NOT verify

- **`build_report` (the `/handoff` probe) still fails closed on a malformed entry — deliberately, and it is a judgement call I did not test my way out of.** It gates a *write* into a curated, unbacked-up store, and writing into a store you have read only part of is the one case where aborting is cheaper than degrading. The consequence is real: a broken entry anywhere in a scope still blocks `/handoff` step 4's probe entirely, and the recovery loop is `--validate` → fix → re-run. If that trade turns out wrong in practice, the mechanism is now a one-word change at the call site.
- **An `OSError` mid-load still fails closed in both modes.** Same blast radius as the malformed case, different fact: the entry *set* is unknown, so there is nothing honest to degrade to. Stated in the docstring and pinned by a test, not fixed.
- **Nothing was run against the real store**, so the "1 of 5 scopes has a README" and "2 good + 1 bad ⇒ 0" numbers are quoted from the task/measured on synthetic fixtures respectively — I did not re-measure the live store's README coverage myself.
- **Not deployed.** These are `home.file`/symlink-managed paths; the skill changes need a `home-manager switch` (or `scripts/ship.sh`) after merge before either host sees them.


ZacxDev:
## Follow-up: the refusal now names the way out (`f23e437`)

Closing the gap you found. **Wording only — fail-closed behaviour and exit code 3 are unchanged**, and that is pinned by an invariant guard that is green at *both* the old branch head and HEAD.

### Before

```
subsystem-touch: malformed index entry 'bad.md': `aliases:` must be a list, not a bare string
```

### After

```
subsystem-touch: malformed index entry 'beta.md': filename 'beta.md' has slug 'beta' but `service:` normalizes to 'gamma' …
  REFUSED, NOT READ: the probe indexed NOTHING and proposed NOTHING. This is not 'nothing touched an
  entry' — no association was computed at all, because a writer must not act on a store it has read
  only part of.
  2 entry files under <STORE> cannot be indexed, and EVERY reader skips them:
    malformed index entry `scope-a/beta.md`: filename 'beta.md' has slug 'beta' but `service:` …
    malformed index entry `scope-b/gamma.md`: `aliases:` must be a list, not a bare string
  1 of them is in THIS repo's scope `scope-a/` — it may be the very entry your paths would have
  matched, so do not conclude this session touched nothing.
  1 is in ANOTHER scope and cannot affect this repo's association at all — but the loader reads the
  WHOLE store, so it still blocks this probe. Running --validate on `scope-a/` alone would report it
  clean and change nothing.
  RECOVER — fix the file(s) named above, then re-run this probe. Check with:
    python3 <SELF> --store <STORE> --scope scope-a --validate
    python3 <SELF> --store <STORE> --scope scope-b --validate
```

### 🔴 The obvious fix walks into a trap, and that shaped the design

"Run `--validate` on your scope" is **unfollowable exactly when it matters most**. The loader reads the *whole* store, so a malformed entry in **another** scope aborts this repo's probe — and `--validate` on the caller's own scope would report it **clean** while the probe kept failing. That is the same unfollowable-instruction shape this PR exists to fix, reintroduced one layer down.

So the refusal re-reads with `COLLECT` and emits **one command per affected scope**, this repo's first. `validate_command()` builds the invocation in one place, so a flag rename cannot leave a message quoting a command that no longer parses — the thing a hand-typed "just run `--validate`" sentence has no defence against.

The re-read is best-effort by construction: if it fails, the **original diagnosis is returned in full** (a message-builder that turns one failure into a different one is worse than a terse one), and a reject that has vanished between the two reads is *said*, not printed as an empty list under a heading promising one. Both branches are tested.

### Spelled vs structural: the test **executes** the command

Asserting the word "validate" appears somewhere would be satisfied by any unrelated sentence containing it. So the tests parse every `python3 …` line out of the message with `shlex`, assert the script path `is_file()`, and **run it** via `st.main(argv[2:])`, asserting it exits 3 and names the same file the refusal named. A sentence cannot survive being executed, and a stale flag cannot survive the parser.

The cross-scope case carries its own **negative control** proving the trap is real: the caller's own scope genuinely validates clean (`rc=0`) while the emitted command reproduces the failure (`rc=3`).

### Red at branch head → green at HEAD

New tests run against the previous branch head (`7c7ade6`): **12 red → 0 at HEAD.**

| | count | how they failed |
|---|---|---|
| **Behavioural** | **6** | `assert 'RECOVER' in "malformed index entry 'widget-index.md': …"` · `assert 0 == 1` and `assert 0 == 2` (no runnable command parsed out) · `assert 'EVERY reader skips' in "<bare message>"` · `AssertionError: deterministic order, this repo first` |
| Doc pins | 2 | `handoff/SKILL.md` lacked the recovery instruction and the run-what-it-printed rule |
| Surface-only | 4 | `AttributeError: module 'subsystem_touch' has no attribute 'malformed_refusal'` ×2 · `ValueError: substring not found` (structural slice on an absent function) · `mutation anchor occurs 0x` (the clause is new) |

**`test_it_still_REFUSES_and_still_exits_3` is green at BOTH** — labelled an invariant guard, not counted as regression coverage. It is the evidence that only the wording moved: the sentinel still *leads* the message (so every existing `except` and `in str(exc)` assertion keeps working), and the CLI still exits 3.

### Mutation

`test_the_MALFORMED_reraise_is_UNKILLABLE_and_that_is_the_finding` is **superseded, and the supersession is the point.** That test recorded a genuine finding — `except MalformedEntryError: raise` sitting above the `OSError` clause was a **no-op**, since `MalformedEntryError` is a `ResolverError`, and a mutation aimed at it proved that by refusing to change anything.

The clause is **load-bearing now**. The same mutation (`except MalformedEntryError` → `except ZeroDivisionError`) strips the route out: the mutant refuses with the bare sentinel and **zero** parseable recovery commands, while the real module emits them — and both still refuse with the same sentinel, so the kill is about the route out and not about the refusal. The premise assertion (`MalformedEntryError` is not an `OSError`) is kept, or the kill would be measuring clause *order* instead of wording.

### Gate — on the pushed commit `f23e437`

```
nix build .#checks.x86_64-linux.pytests   → RESULT: PASS (exit=0)
  PASS  scripts/tests  (collected=3795 passed=3795 skipped=0 floor=3428)
  TOTAL collected=9358  passed=9357  skipped=1  failed=0  (floor: 8577)

nix build .#checks.x86_64-linux.nodetests → RESULT: PASS (exit=0)
  TOTAL suites=4 files=34 tests=1097 pass=1097 fail=0 skipped=0  (floor: 1080)
```

`scripts/tests` 3783 → **3795** (+12) against the control of 3704 on unmodified `origin/main`. No floor change needed. Single pinned pre-existing skip.

### Two notes back

1. **Rebased onto `a8f8426` (#448)** before gating, so the verdict above is the merged reality rather than a stale tree. The rebase delta is byte-identical to #448's own diff (2 lines in `handoff/SKILL.md`) — verified, so none of my work was lost. This required a force-update of the branch; I used `--force-with-lease` pinned to the exact prior remote sha and first confirmed the only commit on the remote not in my local branch was my own pre-rebase one.
2. **Corrected the stale README figure you flagged.** `handoff/SKILL.md` said "measured 2026-08-13: 1 of 5". Since you have written the other four, it now reads: *4 of 5 when this was built, all 5 as of 2026-08-13; a new scope starts without one, so the gap recurs by construction.* `governing_policy()` is unaffected, and — as you note — the no-README branch is now only reachable synthetically, which is exactly how it is tested.

All fixtures remain synthetic; the live store was never read or written.

PR #455, opened 2026-08-13
fix(skills-mapping): the dotted `home.file."…".source =` form returned a clean bill of health while the wrong tree deployed

### LINKED ISSUE #443: skills_mapping: _mapping_block() grabs the next { anywhere in the file, so the dotted home.file."…".source = form can return a false ALL-CLEAR
Found by the delta audit of #438. **This is the one place in that PR where a false green is reachable, and four test modules now rest on this predicate.**

## Too loose — a false all-clear

`scripts/testlib/skills_mapping.py:63-79` — `_mapping_block()` does `home_nix.find("{", start)`, taking the next `{` **anywhere in the file** rather than the mapping's own attribute set.

Verified by running the real predicate against:

```nix
home.file.".claude/skills".source = ../claudedocs;
home.file.".config/opencode/skills" = { source = ../claude/skills; recursive = true; };
```

`skills_mapping_problem()` returns `None` — clean — while `~/.claude/skills` deploys from `../claudedocs`. A variant reports the confidently wrong `"declares no source = at all"` while quoting an unrelated `{ enable = true; }` block.

This is not a hypothetical spelling: `nix/home.nix` already uses the dotted `home.file."…".source =` form roughly 20 times (lines 814, 856, 923, 968, 1030, …) against 34 attrset-form mappings.

## Too strict — five legitimate forms rejected

Each verified by running the predicate:

| form | wrong verdict |
|---|---|
| `mkDerivation { src = ../claude/skills; installPhase = "cp -R ./. $out"; }` | "never copies ../claude/skills into $out" — `_PATH_INTERP` only sees `${…}`, so a plain `src =` is invisible |
| `symlinkJoin { paths = [ ../claude/skills … ]; }` | same |
| one extra let-indirection (`skillsSrc = ../claude/skills;` then `${skillsSrc}`) | same |
| `rm -rf "$out/clickup/node_modules"` as a cleanup step | "removes its own output" — `_RM_OUT` matches any `rm` naming `$out`, not removal *of* `$out` |
| `cp ${../claude/PRINCIPLES.md} "$out/PRINCIPLES.md"` | "copies a second tree…" |

The fourth is the most likely near-term trip — it sits one line from the code it guards.

Unlike the `_mapping_block` failure, none of these carry the "this parser needs updating, do NOT delete the check" hedge. They produce a confident, wrong claim that the deploy is broken, which is how a guard gets deleted.

## Suggested fix

- Require the `{` to be on the mapping's own line or immediately after `=`, and add a **dotted-form fixture** to `scripts/tests/test_skills_mapping_guard.py` — its absence is why this shipped.
- Widen copy-detection to plain path attributes (`src =`, `paths = [ … ]`) and let-indirection.
- Scope `_RM_OUT` to removal of `$out` itself, not any path under it.

## Context

Introduced in #438, which consolidated three open-coded copies of `assert "source = ../claude/skills;" in home_nix` into this predicate. The consolidation was right — a fourth copy landed on main mid-PR and broke the merged tree — but the parser needs to actually hold the property.

ZacxDev:
`scripts/testlib/skills_mapping.py` is the one predicate answering "does `nix/home.nix` actually deploy `claude/skills/`?", and four test modules rest on it. It was wrong in both directions at once, and an adversarial audit of the first fix found the loose direction was wrong in eight more shapes — two of them idioms already in `home.nix`.

Closes #443.

## What was wrong

**Too loose — a false ALL-CLEAR.** `_mapping_block()` located the mapping's attribute set with `home_nix.find("{", start)`: the next `{` **anywhere in the file**. nix's dotted form has no attribute set of its own, so the scan walked past the real source into a sibling's block and returned `None` — clean — while `~/.claude/skills` deployed the wrong tree. The dotted form is not hypothetical: `home.nix` already uses it ~20 times against 34 attrset-form mappings, and it had **no fixture at all**, which is why this shipped.

**Too strict — five confident, wrong rejections.** Copy-detection saw only `${…}` interpolations within 200 characters of a `$out` reference, and any `rm` naming `$out` counted as emptying the output. Unlike the loose case these carried no "parser may need updating" hedge, so they read as "the deploy is broken" — which is how a guard gets deleted rather than fixed. The `rm -rf "$out/clickup/node_modules"` one sits one line from the code it guards.

## Verified reproductions

All run against the real predicate, not inferred. `[base]` = `6eaeb61`, `[HEAD]` = this branch.

| shape | must | base | HEAD |
|---|---|---|---|
| `home.file.".claude/skills".source = ../claudedocs;` + a sibling naming `../claude/skills` | FAIL | **PASS** (false all-clear) | FAIL, naming `../claudedocs` |
| same, sibling is `services.dunst = { enable = true; }` | FAIL | FAIL — quoting `{ enable = true; }`, "declares no `source =`" | FAIL, naming `../claudedocs` |
| `home.file.".claude/skills".source = ../claude/skills;` | PASS | FAIL "could not read its attribute set" | PASS |
| `mkDerivation { src = ../claude/skills; … }` | PASS | FAIL "never copies … into $out" | PASS |
| `symlinkJoin { paths = [ ../claude/skills … ]; }` | PASS | FAIL "never copies" | PASS |
| `skillsSrc = ../claude/skills;` then `${skillsSrc}` | PASS | FAIL "never copies" | PASS |
| `rm -rf "$out/clickup/node_modules"` | PASS | FAIL "removes its own output" | PASS |
| `cp ${../claude/PRINCIPLES.md} "$out/PRINCIPLES.md"` | PASS | FAIL "copies a second tree" | PASS |

### And what the audit of that fix then found

Eight further shapes returned `None` while the deployed tree was not `claude/skills`, plus the headline false positive one character to the left. Each verified by running the predicate, each now a fixture:

| shape | after round 1 | now |
|---|---|---|
| `''${bbOld##*.}` in the mapping — **`nix/home.nix:548`, today** | PASS | FAIL "no `source =`" |
| `echo "theme=#282828" ; cp -R ${../claudedocs}/. "$out"` | PASS | FAIL "as well as" |
| the mapping quoted inside a `''…''` string above the real one | PASS | FAIL, parser |
| `cd "$out"` then `cp -R ${../claudedocs}/. .` | PASS | FAIL, parser |
| `dst="$out"` then `cp -R ${../claudedocs}/. "$dst"` | PASS | FAIL, parser |
| `tar -C ${../claudedocs} -cf - . \| tar -C "$out" -xf -` | PASS | FAIL "as well as" |
| `{ … } // lib.optionalAttrs cond { source = ../claudedocs; }` | PASS | FAIL, parser |
| a `let` name declared twice (shadowing) | PASS | FAIL, parser |
| `cp -R ${pkgs.someOtherTree}/. "$out"` | PASS | FAIL, parser |
| `rm -rf "$out"/clickup/node_modules` (quote before the slash) | **FAIL** | PASS |
| `cp -R ${../claude/skills} "${out}"` | **FAIL** | PASS |
| `tar -C ${../claude/skills} -cf - . \| tar -C "$out" -xf -` | **FAIL** | PASS |

`#` is a comment at line start or after whitespace and nowhere else — blanking every one to end-of-line ate the `}` out of `"''${bbOld##*.}"` and unbalanced the braces the attrset scan counts. Measured on the real file: **345/345 braces raw, 333/332 after stripping.**

## What changed

- The source is read from the mapping's **own** syntax — `."…".source = <expr>;` or `."…" = { source = <expr>; }` — and nothing else is guessed at. Sources that **disagree** are undecidable; several occurrences that agree (one entry written as two dotted attributes) are fine.
- `src`/`srcs`/`paths` count as writing the output root; one further hop of let-indirection is followed.
- The 200-character proximity window is replaced by **statement** scoping, which also makes the destination legible: `$out` itself replaces the tree, `$out/sub` adds to it. `|` and `&&` are **not** separators — a pipeline is one write.
- Because statement scoping cannot see `cd "$out"`, every repo path in a binding must come out **accounted for**; one that is neither a root write nor a write under the root is reported, not dropped.
- Where it cannot decide — an unreadable source expression, an opaque expression over the root, a shadowed name, malformed nix — it **fails saying the parser needs updating and must not be deleted**. `None` is the answer it must never give by accident.

The property is not softened: every widened shape has a wrong-tree twin fixture that must still fail.

## Mutant matrix

Two sweeps, built differently, because a hand-built sweep only covers the mutants I imagined. Fixtures: **14 → 54**.

**Sweep A — 24 hand-written mutants, no deletions.** Each asserts the *expected* test goes red, so a mutant killed only by some other guard is reported as such.

| mutant | kind | result |
|---|---|---|
| M0 control (no-op) | — | **0 failures** (harness can be quiet) |
| M1 root/under inverted | invert a branch | KILLED, expected test red |
| M2 paths/opaque unpacked backwards | swap operands | KILLED, expected |
| M3 dotted branch `if False:` | clause dead, text present | KILLED, expected |
| M4 per-statement → whole-body paths | stale rebind | KILLED, expected |
| M5 binding indent `<=` → `<` | off-by-one | KILLED, expected |
| M6 brace scan starts +1 | off-by-one | KILLED, expected |
| M7 `"$out/."` dropped from the root set | clause commented out | KILLED, expected |
| M8 path normalisation dead | clause commented out | KILLED, expected |
| M9 `_OUT_REF` quote dropped | operand narrowed | KILLED, expected |
| M10 `paths` → `path` | operand swap | KILLED, expected |
| M11 root scope inverted | invert a branch | KILLED, expected |
| M12 undecidable → `return None` | invert the verdict | KILLED, expected |
| M13 no comment strip | rebind | KILLED, expected |
| M14 `depth == 0` → `!= 0` | invert a comparison | KILLED, expected |
| M15 comment lookbehind dropped | widen a lookbehind | KILLED, expected |
| M16 `_ATTR_WORD` eats path parts | widen a lookbehind | KILLED, expected |
| M17 additions stop explaining a path | swap operands | KILLED, expected |
| M18 first source wins again | clause commented out | KILLED, expected |
| M19 shadow count `> 1` → `> 2` | off-by-one | KILLED, expected |
| M20 `//` → `///` | operand swap | KILLED, expected |
| M21 `${out}` not skipped | rebind | KILLED, expected |
| M22 opaque dropped again | clause commented out | KILLED, expected |
| M23 split on `\|`/`&` again | operand widened | KILLED, expected |
| M24 attr-name probe inverted | invert a branch | KILLED, expected |

**24/24 killed with the expected assertion red.**

**Sweep B — 54 mutants enumerated mechanically** (every `in`/`not in`, `and`/`or`, `<=`/`<`, `==`/`!=`, `\|=`/`&=`, True/False token via `tokenize`, so no docstring is touched; plus every problem-returning `return` turned into `return None`): **39 killed, 0 survived, 15 unrunnable** (all `for x in` → `for x not in`, i.e. `SyntaxError`).

Sweep B earned its keep three times: it found the brace-scan mutant that no fixture could tell apart (now `test_a_sourceless_mapping_does_not_borrow_a_siblings_source`), a dead `additions` accumulator and an unreachable identifier channel (both removed), and four unfixtured malformed-nix branches. It also caught **itself** lying first: a same-size mutation rewritten inside the same second left a cached `.pyc` valid, so mutants never executed and reported as survivors — one "survivor" was hand-checked and actually killed 25 tests. Both sweeps now sweep `__pycache__` and set `PYTHONDONTWRITEBYTECODE`.

Three of the wrong-tree twin fixtures also pass on the pre-fix parser. They are **invariant guards on the widening**, not regression coverage for the shipped defect, and are labelled as such.

## Gate

- Real `nix/home.nix` still passes the guard (its mapping is correct and verified live).
- `nix build .#checks.x86_64-linux.pytests` — **rc 0**, `TOTAL collected=9398 passed=9397 skipped=1 failed=0`, `RESULT: PASS (exit=0)`.
- `nix build .#checks.x86_64-linux.nodetests` — **rc 0**, `tests=1097 pass=1097 fail=0`.
- `scripts/gate.sh --tier both --set all` — **rc 0**, both tiers `RESULT: PASS (exit=0)`.
- **Merged-tree gate.** `origin/main` moved three times during this work (`6eaeb61` → `a8f8426` → `23bc6f4`). A fresh worktree off `23bc6f4` with this branch merged in: full gate **rc 0** both tiers, both nix checks **rc 0**. Nothing main gained touches `skills_mapping.py`, `home.nix` or `testlib/`, so there is no overlapping region to read for a semantic conflict.

## What I did NOT verify

- **No `home-manager switch` and no `ship.sh`** — out of scope and explicitly excluded. This changes a test-only predicate; nothing under `nix/` moved, so no deployed artifact changes.
- **`claude/skills/clickup/test/smoke-test.mjs` was not run** (live writes to a real ClickUp workspace). The node tier runs `*.test.mjs` only, so it never reaches it.
- **Two known limits are documented in the module, not fixed.** `src`/`srcs`/`paths` are matched by NAME, not by POSITION, so a `src` sitting in a `runCommandLocal` **env** attribute set is still credited as the output tree — a constructible false all-clear. And a repo path the parser cannot classify fails asking for the parser, which rejects some legitimate-but-unusual builds; that direction is deliberate, since the alternative is `cd "$out"` aliasing passing in silence.
- **This reads text, not nix.** `nix eval` is the only thing that actually knows what a `home.file` source resolves to; every guarantee here is about the shapes fixtured above.

### DIFF
--- a/scripts/testlib/skills_mapping.py
+++ b/scripts/testlib/skills_mapping.py
@@ -42,6 +42,50 @@ INTO ``$out`` -- with no other repo tree copied there, and no removal of
 ``$out``. That is still structural (it says nothing about how the copy is
 spelled, and the real binding's ``ln -sT ${clickupNodeModules}/...`` passes
 untouched); it just stops treating "the characters appear somewhere" as proof.
+
+WHAT ISSUE #443 FOUND, AND HOW THE PARSER CHANGED
+--------------------------------------------------
+The first structural version was both too loose and too strict, in one file.
+
+*Too loose.* It located the mapping's attribute set with
+``home_nix.find("{", start)`` -- the next ``{`` ANYWHERE in the file. nix's
+DOTTED form has no attribute set of its own::
+
+    home.file.".claude/skills".source = ../claudedocs;
+    home.file.".config/opencode/skills" = { source = ../claude/skills; ... };
+
+so the scan walked past the real (wrong) source into an unrelated block and
+returned a clean bill of health while ``~/.claude/skills`` deployed
+``../claudedocs``. The dotted form is not hypothetical -- home.nix already uses
+it about twenty times. The mapping's source is now read from the mapping's OWN
+syntax: either ``."…".source = <expr>;`` or ``."…" = { source = <expr>; };``,
+and nothing else is guessed at.
+
+*Too strict.* Copy-detection saw only ``${…}`` interpolations sitting within
+200 characters of a ``$out`` reference, and any ``rm`` naming ``$out`` counted
+as emptying the output. Five legitimate shapes were therefore rejected with a
+confident, wrong "the deploy is broken" -- which is how a guard gets deleted
+rather than fixed. Now:
+
+  * the root-producing derivation attributes (``src``/``srcs``/``paths``) count
+    as writing the output, so ``mkDerivation { src = ../claude/skills; }`` and
+    ``symlinkJoin { paths = [ ../claude/skills … ]; }`` are understood;
+  * one further hop of let-indirection is followed, so ``skillsSrc =
+    ../claude/skills;`` used as ``${skillsSrc}`` is understood;
+  * proximity is gone. The build script is split into STATEMENTS, and a repo
+    path counts as written to the output only in a statement that also names
+    ``$out``. That also makes the destination legible: writing ``$out`` itself
+    replaces the tree, writing ``$out/sub`` adds to it. So ``rm -rf
+    "$out/clickup/node_modules"`` and ``cp ${../claude/PRINCIPLES.md}
+    "$out/PRINCIPLES.md"`` are additions, while ``rm -rf "$out"`` and ``cp -R
+    ${../claudedocs} "$out"`` still fail.
+
+The property being defended is unchanged and deliberately NOT softened: the
+deployed tree must be built from ``../claude/skills``, and no other repo tree
+may be written over the output root. Where the parser cannot decide -- an
+unrecognised source expression, or an unresolvable identifier written over the
+output root -- it FAILS, saying it needs updating and must not be deleted.
+Silence is the one answer it must never give.
 """
 from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


ZacxDev:
Closing in favour of **shrinking** this guard rather than making it correct.

This PR does genuinely fix #443's dotted-form false ALL-CLEAR — that part is real and was verified red-at-base. But the audit of the final diff found the approach doesn't converge:

- **Six constructible false ALL-CLEARs remain**, two of them one-line edits to the real block (`enable = false;`, `target = "…";`), which the parser blesses because it looks for `source` and only `source`.
- **One regression against base**: `runCommandLocal { src = ../claude/skills; } '' mkdir -p "$out" ''` deploys an empty tree — base rejected it, this branch calls it clean.
- **A live landmine**: `nix/home.nix` has 7 `/*` sequences inside `#` comments today and zero `*/`. Because block-comment stripping runs first, adding any comment containing `*/` makes the guard announce *"home.nix no longer declares the ~/.claude/skills mapping"* — confidently wrong, reds four test modules.
- The module went **8,846 → 29,920 B**. The heuristic surface is growing faster than its coverage.

The decisive facts, though, are about whether the guard should be asking this much at all:

1. **It has never caught a real breakage.** Its only firing was a false positive on #438 — a change that left deployment intact.
2. **The hard half is already checked correctly, against reality.** `ship.sh` prints per host on every deploy — `✅ managed artifacts resolve — 429 checked, 0 dangling, 0 absent` — and `drift-check.sh` does it passively (rc 14). Those read the real filesystem on both hosts. A text parser re-deriving that from nix source is strictly worse and always will be.
3. **The holes are latent.** The mapping at `nix/home.nix:907` is the attrset form, so neither the dotted-form nor the `enable` hole can bite without a future edit.

Superseded by a PR that cuts the module under a **4,000 B ceiling**: keep "the mapping is declared, names a source, and isn't neutered by `enable`/`target`", drop the resolve-to-repo-tree analysis entirely, and name `ship.sh`/`drift-check.sh` in a comment as what actually verifies it.

That is a real reduction in pre-merge coverage, stated plainly rather than papered over — traded for a guard small enough to stay correct, and one that won't get deleted by the next maintainer for crying wolf.

The work here wasn't wasted: it produced the `enable`/`target` finding that the replacement keeps, and the audit's `nix-instantiate --parse` observation, which the replacement may use.

PR #461, opened 2026-08-13
fix(browser-agent): a 0.25s test measured 125s under load, and the warm lock slept 120s on errors sleeping cannot fix

ZacxDev:
Fixes the 2026-08-13 flake of `test_agent_def_denies_bash_allows_only_custom_tool` (clawgate Task #194).

## The stderr in the failing log is not the cause

The failing run carried the wrapper's *"could not warm the isolated opencode config dir … re-run ONCE with network available"* warning. It is a reporting artifact: pytest echoes captured stderr **only when a test fails**. Measured — that warning was emitted on **every run of all 51 tests in this file**, because the fake `opencode` can never materialise `node_modules`, so the warm step could only ever fail and fall back to the global config. Its presence in the failing log carries zero information about that failure. The whole suite had been running in the degraded fallback, not the isolated path production uses.

## What the 60 seconds actually had in it

| fact | measured |
|---|---|
| happy-path wrapper run, idle | **0.25 s** (n=30, median 0.24 s) — ~10 execs, no dominant step |
| same run, under 4 concurrent `devrc-pytests` builds + hogs, loadavg 65 | **125.4 s** — a ~400× machine stall |
| `_oc_lock` on a lock path that can never be created | **120.3 s** of `sleep 0.5`, then it "stole" a lock that never existed and bootstrapped **unserialised** |
| the rig's deadline | 60 s — *shorter* than three budgets the wrapper may legitimately spend inside it (warm 90 s, lock wait ~120 s, watchdog 120 s) |

Two mechanisms can produce a 60 s timeout here, and one occurrence cannot distinguish them. Both are addressed.

## Changes

**`scripts/browser-bridge/browser-agent`**
- `_oc_lock` distinguishes contention (EEXIST **on a directory** → wait) from a permanent failure (ENOSPC/EACCES/EIO/a stray file → refuse). `oc_bootstrap` now honours the refusal instead of bootstrapping without the lock.
- Each refusal branch gets its **own** stderr message. The end state is identical either way, so the message is the only thing separating "refused immediately" from "slept two minutes then refused" — and a test must not read that off a wall clock.

**`scripts/browser-bridge/tests/test_browser_agent.py`**
- The rig **pre-warms** the isolated opencode config dir (the fixture equivalent of shipping the warmed tree as a build input). The bootstrap no longer runs at test time at all — which deletes the two paths whose own budgets exceed the suite's deadline, drops two process spawns per test, and moves the tests onto the isolated config dir production actually uses.
- The rig waits on the **deterministic condition** (the process exited) and consults the clock only to decide whether "still running" means *hung*. On each expiry it re-measures how long the machine currently needs to spawn ten trivial processes (0.085–0.105 s idle here): normal ⇒ the wrapper is stuck, fail with a legible diagnosis; inflated ⇒ the **machine** is stalled, keep waiting. `communicate()` does not kill on expiry, so nothing is re-executed — this is not a retry. A hard cap remains the real hang-net.
- Five new tests on paths that ran on every test and were asserted by none: warm dir is used and skips the bootstrap; cold dir degrades loudly and still runs; an unusable lock refuses instead of waiting; the deadline reports a machine baseline rather than a bare `TimeoutExpired`; a stalled machine extends the wait instead of failing.

An `elapsed < 20` assertion was tried and **removed** — it failed under the load reproduction at 125.4 s while every structural assertion passed. Re-asserting a wall clock to guard a wall-clock bug just relocates the flake.

## Verification

- **The guarded security property still goes red on a deliberate break.** Adding `bash: allow` to the agent def fails `test_agent_def_denies_bash_allows_only_custom_tool` on *its own* assertion (`"the old bash permission block must be gone"`), reached after the earlier checks pass.
- **Red at the parent commit**, on its own assertion, after a measured 120.27 s stall: `test_an_unusable_warm_lock_degrades_instead_of_stalling`.
- **Mutation controls**: deleting the stall-extension branch → `test_a_stalled_machine_extends_the_wait…` red; restoring the lock spin → the lock test red. The lock guard took **three** rounds — the `opencode.jsonc` check alone stayed green under a sleep-only mutation, and so did a single shared message (the post-wait branch emits it too).
- **Authoritative gate**: `498 collected / 498 passed`, `RESULT: PASS (exit=0)`, no `=== FAILURES ===`.
- **Original condition reproduced**: gate re-run against 3 concurrent variant `devrc-pytests` builds + a sibling agent's build + 32 CPU hogs (loadavg 65) → browser-bridge `496 passed in 172.08s`, PASS.

| browser-bridge target | tests | wall |
|---|---|---|
| main, idle (from the task report) | 493 | 165.02 s |
| this branch, idle | 498 | 168.33 s |
| this branch, under the load repro | 496 (pre-final tree) | 172.08 s |

The +3 s is the two new tests, which sleep ~5 s by design, minus the ~1.3 s the pre-warm gives back across the other 51.

## Not established

Which of the two mechanisms produced that specific 60 s. An empty result identifies no mechanism, and one occurrence cannot separate a lock stall from a machine stall. The lock stall is fixed; the machine stall is now waited out and, if it ever exhausts the cap, says so on sight instead of pointing at the warm warning.

### DIFF
--- a/scripts/browser-bridge/browser-agent
+++ b/scripts/browser-bridge/browser-agent
@@ -141,6 +141,26 @@ OC_CONFIG_DIR="${BROWSER_AGENT_OPENCODE_CONFIG_DIR:-${XDG_CACHE_HOME:-$HOME/.cac
 # that works and is expensive beats one that hangs. The warning goes to stderr
 # on every such run so it cannot rot silently.
 OC_WARM_TIMEOUT="${BROWSER_AGENT_WARM_TIMEOUT:-90}"
+# How long to wait for a lock somebody else is holding, IN SECONDS — the quantity
+# the comment on `_oc_lock` has always described. It used to be spelled as an
+# iteration count (`(warm + 30) * 2`), which silently encoded the 0.5 s poll:
+# change the poll and the budget moved with it, unannounced. Test seam, in the
+# same spirit as BROWSER_AGENT_READY_BACKOFF — the contention branch is otherwise
+# unreachable in under ~40 s, so nothing ever entered it.
+#
+# 🔴 VALIDATE BOTH SEAMS HERE, with the same `case` idiom `--steps`/`--timeout`
+# use. The budget is derived by ARITHMETIC at top level now, not inside the cold
+# path, so a non-numeric BROWSER_AGENT_WARM_TIMEOUT would spray a raw bash
+# arithmetic error over EVERY run (warm ones included, which previously never
+# evaluated it) and then die later on an unbound `deadline` — a misconfiguration
+# reported as an internal shell error, two steps from its cause.
+case "$OC_WARM_TIMEOUT" in (*[!0-9]*|"")
+  die "BROWSER_AGENT_WARM_TIMEOUT must be a non-negative integer of seconds (got: $OC_WARM_TIMEOUT)";;
+esac
+OC_LOCK_BUDGET="${BROWSER_AGENT_LOCK_BUDGET:-$(( OC_WARM_TIMEOUT + 30 ))}"
+case "$OC_LOCK_BUDGET" in (*[!0-9]*|"")
+  die "BROWSER_AGENT_LOCK_BUDGET must be a non-negative integer of seconds (got: $OC_LOCK_BUDGET)";;
+esac
 OC_STAMP="$OC_CONFIG_DIR/.devrc-warm-stamp"
 OC_LOCK="$OC_CONFIG_DIR.lock"
 OC_ISOLATED=0
@@ -169,13 +189,37 @@ oc_is_warm() {  # warm == node_modules present AND stamped for THIS opencode
 oc_bootstrap() {  # <version> -> 0 warm, 1 could not warm
   local ver="$1" rc=1
   mkdir -p "$OC_CONFIG_DIR" 2>/dev/null || return 1
-  _oc_lock
+  # 🔴 A lock we could not take is a REFUSAL, not a green light. It used to be
+  # called for effect and its result discarded, so an uncreatable lock path fell
+  # through to `_oc_bootstrap_locked` UNSERIALISED — after stalling for two
+  # minutes first (see `_oc_lock`).
+  _oc_lock || return 1
   _oc_bootstrap_locked "$ver"; rc=$?
   rmdir "$OC_LOCK" 2>/dev/null
   return $rc
 }
 
-_oc_lock() {
+# ONE writer for every warm-lock refusal, emitting a DECLARED TOKEN plus prose.
+#
+# 🔴 The token is a contract, not decoration. Which branch refused is otherwise
+# unobservable: both end in "no lock, fall back to the global config", identical
+# on disk and in the exit code, so the only other witness is a wall clock — and
+# on a loaded box that instrument lies (a 0.25 s run measured 125 s here under
+# four concurrent gate builds). The tests therefore read the TOKEN. Reword the
+# prose freely; changing a token is changing an interface, and a test will say so.
+#
+# Tokens:
+#   unwaitable — the path can never become a lock (ENOSPC/EACCES/EIO, or a stray
+#                regular file sitting there). Refuse IMMEDIATELY; waiting cannot
+#                clear any of those, and pretending it might is what produced a
+#                silent ~120 s stall that surfaced as an unreadable timeout.
+#   timeout    — a real holder never released it within OC_LOCK_BUDGET and the
+#                stale-lock steal also failed. We DID wait; that is legitimate.
+_oc_lock_refuse() {   # <token> <detail>
+  echo "browser-agent: warm-lock-refused:$1 — $2 Falling back to the GLOBAL opencode config." >&2
+}
+
+_oc_lock() {  # -> 0 lock HELD, 1 the lock path is unusable (caller must not bootstrap)
   # Serialise concurrent runs. Bounded wait, then proceed anyway: a stale lock
   # from a killed run must not wedge every future invocation.
   # 🔴 The wait MUST outlast the work it guards. It was a fixed 60 x 0.5 s = 30 s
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Audit round — both 🟡 fixed, one pushback

### 1. The surviving mutants (W5, W7)

The diagnosis was right about the symptom but the root cause was worse than "spelled guard": **the contention branch of `_oc_lock` had no test reaching it at all.** Every case I had written takes the *immediate-refusal* branch (stray file at the lock path), so `"gave up after waiting" not in stderr` was a guard that could not fail — and W5 sailed through while reintroducing the exact defect this PR removes.

Fixed three ways:

- **A test that reaches the branch.** A *non-empty directory* at the lock path is what makes it reachable: `mkdir` fails as genuine contention (so waiting is *correct* there), and after the budget the stale-lock steal also fails because `rmdir` cannot remove a non-empty directory. `test_a_lock_a_holder_never_frees_refuses_rather_than_bootstrapping_unlocked`.
- **A structural observable.** `opencode.jsonc` is written by `_oc_bootstrap_locked` and by nothing else, so its absence proves the bootstrap never ran. W5 now fails on *that* line: `"the bootstrap ran WITHOUT the lock"`.
- **One writer, declared tokens.** Both refusals come from `_oc_lock_refuse <token> <detail>`, emitting `warm-lock-refused:unwaitable` / `:timeout`. Which branch refused is otherwise unobservable — both end identically on disk and in the exit code — so the token is a stated contract, not incidental prose.

Reaching that branch in test time also needed the bound to stop being an iteration count that silently encoded the 0.5 s poll: it is now `OC_LOCK_BUDGET` **in seconds** (seam `BROWSER_AGENT_LOCK_BUDGET`), so the branch is reachable in ~1 s instead of ~120 s. That is why it shipped untested.

**Pushback on W7, with specifics.** W7-as-executed (replacing the message) is now red, because the token goes with it. But I added **W13 — reword the *prose*, keep the token — and it stays green deliberately.** Rewording a diagnostic changes no behaviour; the fix for a spelled guard is to stop depending on incidental spelling, not to spell it harder. If a *token* changes, a test says so. Tell me if you want the prose frozen too and I will pin it, but I think that trades a real guard for a changelog.

### 2. The pre-existing wall-clock assertions — agreed, both gone

- `test_tab_closed_on_timeout`: `elapsed < 20` removed. `status == "blocked"` already proves the kill *deterministically* — a wrapper that failed to kill collects the fake's post-sleep output and reports `"ok"`. **W14** (watchdog never fires) confirms it: `AssertionError: assert 'ok' == 'blocked'`. The slice is also raised above the fake's 30 s sleep so a broken wrapper runs to completion and fails on the assertion that means something.
- The straggler-reap poll: `deadline = time.time() + 8` replaced by `_await` on a **terminal condition** — child reaped XOR marker written, mutually exclusive and exhaustive. Both waits now share one rule (`_stall_extends`), so the discipline cannot drift between them; `_await`'s extension branch has its own positive control (**W12**).

### 3. Stamp duplication — declined, with reasoning

Sourcing the wrapper is unsafe (no main guard), and the only alternative is a production `--print-warm-stamp` whose sole consumer is a test. The drift is **audible** — W11 proves the guard fires — which is the property "one rule, one place" is protecting here.

### A guard I had to move, that only the gate could see

`scripts/tests/test_opencode_config.py` pinned the literal `max=$(( (OC_WARM_TIMEOUT + 30) * 2 ))`. My refactor **kept that guard's invariant and failed its spelling** — and, worse, anything that broke the invariant without touching those characters would have passed it. It now evaluates the wrapper's own derivation under bash at **two** warm timeouts and asserts `budget >= warm + 30`. **W15** (a fixed 30 s wait — the original bug shape) is red at `warm=90` and green at `warm=0`, which is exactly why it measures two points. My local single-file runs structurally could not see this; the authoritative gate is what caught it.

### Re-verification (a fix round resets the gate)

**Mutation battery — 13 mutants, each red on the intended test's own assertion:**

| mutant | caught by |
|---|---|
| W1 `bash: allow` in the agent def | `test_agent_def_denies_bash_allows_only_custom_tool` — *"the old bash permission block must be gone"* |
| W2 sleep on permanent errors again | `test_an_unusable_warm_lock_degrades_instead_of_stalling` |
| W5 drop `return 1` after a failed steal | `test_a_lock_a_holder_never_frees…` — *"the bootstrap ran WITHOUT the lock"* |
| W7 replace the contention refusal | `test_a_lock_a_holder_never_frees…` (token) |
| W8 drop `return 1` on the unwaitable path | `test_an_unusable_warm_lock…` |
| W9 discard `_oc_lock`'s result | both lock tests |
| W14 watchdog never fires | `test_tab_closed_on_timeout` — `'ok' != 'blocked'` |
| W10 delete the run-wait stall extension | `test_a_stalled_machine_extends_the_wait…` |
| W12 delete the poll stall extension | `test_the_condition_poll_also_extends…` |
| W11 drift the pre-warm stamp | `test_a_warm_config_dir_is_used…` |
| W15 fixed 30 s lock wait | `test_browser_agent_lock_wait_outlasts_the_work_it_guards` |
| **W13 reword prose, keep token** | **green — deliberate, see above** |

(One earlier mutant was discarded as invalid: it broke bash syntax rather than semantics, failing 53 tests. A mutant that kills the interpreter tests nothing.)

**Red at the parent commit**, each on its own assertion: `test_an_unusable_warm_lock…` after a 120.7 s stall, and `test_a_lock_a_holder_never_frees…` with *"the bootstrap ran WITHOUT the lock"*.

**Authoritative gate**, read by content: `TOTAL collected=9436 passed=9435 skipped=1 failed=0`, `scripts/browser-bridge/tests` 500/500, `scripts/tests` 3866/3866, no `=== FAILURES ===`, no timeout panic, `RESULT: PASS (exit=0)`.

Not re-run this round: the under-load gate reproduction (unchanged conditions, and the only assertion that was load-sensitive is the one I deleted).

ZacxDev:
## Delta round 2 — required change applied, both claims trimmed, three 🟢 taken

Thanks for rebuilding the mutants from semantics rather than my list, and for instrumenting the loop to *prove* the contention branch is entered — that is a stronger fact than my structural assertion, which only shows the refusal, not the two iterations before it.

### 1. REQUIRED — third warm point. Done, and it fires exactly as predicted.

`for warm in (0, 90, 600)`. Each of the three now catches a class the others miss, **measured, not argued**:

| mutant | red at | green at |
|---|---|---|
| W15 fixed 30 s | `warm=90` → *"resolves to 30s … must be >= 120s"* | 0 |
| W17 `warm * 2` | `warm=0` → *"resolves to 0s … must be >= 30s"* | 90 |
| **W16 fixed 200** | **`warm=600` → *"resolves to 200s … must be >= 630s"*** | **0 and 90** |

W16 was green at both my old points, exactly as you said. Since `BROWSER_AGENT_WARM_TIMEOUT` is an unbounded production seam, "big enough for my samples" was never "derived from the warm it serialises" — the docstring now names which class each point exists for, so the next person cannot drop one as redundant.

### 2. Trim: "the wait is now tested" — corrected, and I did not paper over it

You are right and I have stated it at measured scope in the test itself: it proves the branch is **reached and refuses correctly**, *not* that the wait is applied. I confirmed your mutant independently — **W18 (`SECONDS + 0`) is green** across all 60 in-file tests and all 628 in `test_opencode_config.py`, and it is now recorded in the docstring alongside *why* it stays open: separating "waited" from "did not wait" needs a clock, which this PR deliberately removes. I did design the clock-free alternative — a helper that releases the lock mid-wait, so the wrapper must still be waiting to observe the release — and rejected it: it races the wrapper's own startup, and under the 125 s stall this PR exists to survive the releaser would fire before the wrapper reached the loop. That buys the guard back at the price of the flake. Declining, with the gap named rather than implied. (And agreed: the parent was equally unguarded, so nothing is weakened.)

### 3. Trim + fix: `_await` exhaustiveness

Corrected — "mutually exclusive and exhaustive" is true *eventually*, not within a slice, and the slow branch is ~30 s against `slice_s=8.0`. Now `slice_s=35.0`, with the reasoning in the docstring: **the slice must exceed the longest time the condition can legitimately take, or the caller gets `_await`'s "it never happened" instead of its own assertion.** Free on green — the inner poll returns the instant the predicate holds; the test still runs in ~2 s.

### Optional — taken

`str(lock) in r.stderr` on both refusal tests. Agreed on the principle: pinning that the message *names the path* is pinning a fact, not a wording — an operator has to be told which path to go fix. **W21** (drop `$OC_LOCK` from the message) is now red.

### 🟢s

| 🟢 | disposition |
|---|---|
| hard-cap term `waited < cap` unguarded | **Taken.** It is the actual hang-net; a forced-stall + `stall_cap=0` control now pins it. **W19** (drop the term) → `Failed: DID NOT RAISE`. |
| ~30 min worst case vs the parent's ~38 s | **Taken as documentation** — named in `_await`'s docstring as a deliberate trade: slow-and-correct beats fast-and-wrong about which component failed. |
| "machine is not the explanation" when the cap fired on a stalled box | **Taken.** `_await` now splits the verdict the way `rig.run` already did — misattributing a cap-exhaustion to a healthy machine is the exact error this file exists to stop. |
| non-numeric seam now aborts every run | **Taken, and it was worse than a 🟢.** Both seams validated at the top with the script's own `case`/`die` idiom. **W20** (remove it) reproduces the precise damage: `line 157: 9x: value too great for base` **and** `line 158: OC_LOCK_BUDGET: unbound variable` — a misconfiguration surfacing as two raw shell errors, on warm runs that never evaluated it before. |
| `:1281` takes `rig` unused | **Taken**, dropped. |

### Re-verification

**Battery round 2 — 12 mutants over BOTH suites (688 tests), each red on the intended assertion:** W15/W16/W17 (each at a different warm point), W19 hard cap, W20 seam validation, W21 dropped path, W5 (*"the bootstrap ran WITHOUT the lock"*), W7, W12, W1 (*"the old bash permission block must be gone"*).

**Green and documented:** W13 (prose reword, token kept — upheld) and W18 (`SECONDS + 0`, the use-site gap above).

**Authoritative gate, read by content:** `TOTAL collected=9438 passed=9437 skipped=1 failed=0` · `scripts/browser-bridge/tests` 502/502 · `scripts/tests` 3866/3866 · no `=== FAILURES ===`, no timeout panic · `RESULT: PASS (exit=0)`.

PR #464, opened 2026-08-13
fix(next-step-nudge): a blocked next step had no shape to be stated in, so it got dressed up as "say proceed"

ZacxDev:
## The defect

`NUDGE` offered exactly three endings: an authorizable action, a numbered choice with a recommendation marked, or "nothing further; this is done". **There was no shape for a next step that exists but is blocked on something outside this turn** — a running background agent, a pending deploy, a decision that belongs to someone else. "Finished" is false there, so neither escape fits, and the model reaches for the first shape.

Observed live within an hour of the deploy, twice in consecutive turns: the turn closed with *"say `proceed` and I'll verify the gate, merge, and ship"* while the agent producing the thing to verify was still running. The operator typed `proceed`; there was nothing to action. On the second occurrence the model invented parallel work so that `proceed` would mean something — the filler-work behaviour this hook exists to discourage.

## The change — prose only

A fourth shape. The suppressors, the gates, the once-per-session claim and every constant are untouched.

```
  • if the next step is blocked, what it is blocked ON and what happens once that
    clears — "waiting on the audit agent; when it lands I'll re-run the gate" is a
    complete ending, never to be dressed up as an authorizable "proceed".
```

`NUDGE`: **583 → 795 chars (+212)**. The bullet is 231; 22 of that was **absorbed** by tightening the opening sentence ("so answering it costs the operator an extra round trip just to ask" → "which costs the operator a round trip to ask"), +3 for `marked.` → `marked; or`. The three existing shapes are unchanged.

The exemplar is chosen to land on an **existing** suppressor rather than to need a new one: `COMMITS` catches "I'll re-run". `OFFERS` knows "awaiting" and "standing by" but **not** "waiting on", so a bare-present-tense reword of the exemplar would make the hook fire on its own advice — see the second test below.

## Verification

Both new guards read the **bullet list parsed out of `NUDGE`**, not substrings of the paragraph — the surrounding prose already spells "proceed", "finished" and "blocked", so a substring check cannot tell an offered shape from an aside that mentions one.

**Watched to fail.** New tests applied to the pre-change hook at `ed35850` (origin/main):

| test | at `ed35850` | at HEAD |
|---|---|---|
| `test_nudge_offers_a_shape_for_a_next_step_that_is_blocked` | FAILED — `expected exactly one blocked shape, got: [<the two old bullets>]` | pass |
| `test_the_blocked_shape_ships_an_exemplar_the_predicate_recognises` | FAILED — same, via the shared helper | pass |
| `test_self_consistency_the_nudge_does_not_fire_on_its_own_advice` | FAILED — `assert 2 == 3` on the bullet count | pass |

Full file: **114 passed** at HEAD.

**Mutation sweep** — each mutant killed by the intended assertion, with that assertion's own message; an unmutated copy of the same tree is green (3 passed), so the sweep reads the mutation and not a broken copy:

| mutation | killed by |
|---|---|
| exemplar reworded to bare present tense | `assert 'commits' in []` |
| the "never … proceed" prohibition dropped, shape kept | `assert 'never' in <bullet>` |
| the "what happens when it clears" half dropped | `assert 'clears' in <bullet>` |

**One existing assertion was over-fitted and is corrected, not loosened.** `"proceed" in nsn.NUDGE` (and its e2e twin) pinned the FIRST shape by a word the new bullet also spells — it would have passed with the first shape deleted entirely. Both now assert `single "proceed"`, that shape's own phrase.

**IO contract re-checked first-hand** through a real subprocess under a temp `HOME`: `rc=0`, stderr empty, stdout parses as `{"hookSpecificOutput": {"hookEventName": "Stop", "additionalContext": …}}` with exactly those keys. The mechanism has not regressed to blocking.

## Gate

`nix build .#checks.x86_64-linux.pytests` → `RESULT: PASS (exit=0)`, read out of `nix log` (the drv for the clean committed tree is the one that was built — `fmq6kzpy…-devrc-pytests.drv`):

```
PASS  scripts/claude-hooks/tests/test_next_step_nudge.py  (collected=114 passed=114 skipped=0 floor=109)
TOTAL collected=9540  passed=9539  skipped=1  failed=0  (floor: 8983 = sum of 18 per-target floors)
```

112 on main → 114 here; the +2 is the evidence the gate ran the new assertions. Floor re-pinned from that measurement through the gate's own function, not arithmetic:

```
_suggested_floor 114 = 114 - min(50, max(1, 114/20 = 5)) = 114 - 5 = 109
```

⚠ Two sibling branches also add tests today, so `TARGET_FLOORS` will conflict here. Re-run the gate on the **merged** tree and copy what it prints — do not reconcile the two sides by hand (`rerere` has replayed a stale resolution onto this line before).

## Not verified

The hook is live on both hosts and was deliberately **not** re-deployed: no `home-manager switch`, no registrar run, no `settings.json` touched. Whether the new bullet actually changes model behaviour in production is unmeasured — it is a prose change, and the only evidence here is that the shape exists, is self-consistent with the suppressors, and ships intact.


ZacxDev:
## Round 2 — the gap is closed at the source, not guarded in CI

Round 1 shipped the blocked bullet and disclosed that the predicate could not see the wording it teaches. Authorised to touch the suppressors, so:

**`OFFERS` gains `waiting on|blocked on`.** The source comment now splits OFFERS into two families — (a) hand-off to the operator, (b) **PARKED**: the turn states what it is waiting on (`awaiting`, `standing by` were always here) — and records the trade in the safe direction, including what it gives up: "the request is waiting on the cache" mid-analysis now suppresses too.

### The six endings, re-measured against the final merged tree

| verdict | suppressors | ending |
|---|---|---|
| SUPPRESSED | `['offers']` | Blocked on all three agents; nothing for you to action until they report. |
| SUPPRESSED | `['commits', 'offers']` | waiting on the audit agent; when it lands I'll re-run the gate |
| SUPPRESSED | `['offers']` | **Waiting on the running agent. The gate re-runs when it lands.** ← the one that fired |
| SUPPRESSED | `['commits', 'offers', 'done']` | I'm blocked on the deploy finishing; nothing further until then. |
| SUPPRESSED | `['done']` | Nothing further from me until they land. |
| SUPPRESSED | `['offers']` | Standing by for the agent to report. |

Six of six. **Positive control alongside**, because a suppressor set that suppresses everything looks identical to a fix: the `NO_NEXT_STEP` endings still **FIRE**.

### 🔴 The full battery found a pre-existing gap that was not mine

Re-run over the **whole** suppressor set — 7 suppressor removals + 12 arm removals, each mutant in its own tree, against an unmutated control — **two mutants SURVIVED**: deleting `awaiting` or `standing by` left the suite green. Those arms had no fixture, and `test_arm_ledger_covers_every_arm_named_in_the_source_comments` covered only COMMITS and MARKED, so no sweep could ever have seen them.

Closed in the same round: `offers_awaiting` + `offers_standing_by` fixtures, and the OFFERS parked family added to the ledger.

**Final battery on the merged tree: control green (122 passed), 19/19 KILLED, `SURVIVORS: none`.** Each offers-arm mutant now dies to three tests naming the arm.

Arm fixtures are pinned at **arm grain**, one per alternation arm, and each is owned by OFFERS *alone* — no `I'll`/`will` anywhere, so COMMITS cannot shadow them and rescue a deleted arm. The arm-deletion mutant **is** the pre-change regex, so that same run is the watched-to-fail evidence: `arm:offers/waiting on` → red at `test_each_alternation_arm_is_uniquely_owned[offers_waiting_on]` and `..._suppresses_the_turn[offers_waiting_on]`.

`test_the_blocked_shape_ships_an_exemplar_the_predicate_recognises` is **kept**, docstring corrected: the exemplar is now suppressed twice over (COMMITS *and* the new arm), so the test's job is self-consistency against a future reword rather than cover for a hole in the predicate. Attribution stays asserted with `in` rather than equality, and the reason is written down.

### Merged with main; measured there

`origin/main` moved to `8fe1654` (#465, #459, #463). Merged in — **one conflict, in `TARGET_FLOORS`**, exactly the churn magnet its own comments warn about. Both sides kept, because they pin *different* targets; no arithmetic across the two sides. `rerere` **recorded** a preimage rather than replaying one, so no stale resolution was applied, and `run-tests.sh --check-floors` re-validates the two-way pin on the result.

Gate on the **merged** tree — `nix build .#checks.x86_64-linux.pytests` → `RESULT: PASS (exit=0)`, read from `nix log`:

```
PASS  scripts/claude-hooks/tests/test_next_step_nudge.py       (collected=122 passed=122 skipped=0 floor=116)
PASS  scripts/claude-hooks/tests/test_registrar_activation.py  (collected=17  passed=17  skipped=0 floor=16)
TOTAL collected=9610  passed=9609  skipped=1  failed=0  (floor: 9006 = sum of 19 per-target floors)
```

Floor from the gate's own function on the merged measurement: `_suggested_floor 122 = 122 − min(50, max(1, 122/20 = 6)) = 116`. 114 → 122 is +8: four arm fixtures × two parametrized tests.

### Still not verified

Nothing was deployed: no `home-manager switch`, no registrar run, no `settings.json`, no tmux. The live hook on both hosts is still the old one. Whether the widened suppressor changes real firing rates is unmeasured — the evidence here is lexical (the six cases + the battery), not behavioural.

One thing I did **not** do, deliberately, since it was outside the authorisation: `waiting for` and `blocked by` are still unknown to OFFERS and would fire. Same family, same argument, one-line change if you want it.

PR #469, opened 2026-08-14
chore(flake): land the stray lock bump, and re-derive the opencode pin at 1.18.16

ZacxDev:
A `flake.lock` bump was sitting **uncommitted** in the workbench base clone (nixpkgs forward
~15 days, opencode `1.18.4` → `1.18.16`). It made the local gate red and would have gone to
**both hosts** on the next `ship.sh` as an unreviewed nixpkgs move. This lands it deliberately.

Two commits, so the lock move is reviewable on its own:

1. `ec0bf0e` — the lock, alone.
2. `ae17ba8` — the opencode measurement pin, **re-derived rather than re-spelled**.

## Why commit 2 is not a find-and-replace

`test_engine_is_the_version_every_measurement_is_keyed_to` exists precisely to stop a version
move being absorbed by editing `PINNED_VERSION`. Its own failure message says so. So the
claims were re-measured against the new binary:

- `debug agent <a> --pure` dumped for **all seven** agents (4 primaries + 3 hidden) under
  **both** binaries, from a repo-seeded config dir. Compared order-insensitively: **IDENTICAL**
  — permissions as a set, resolved tool map, and model, on every agent.
- `test_opencode_engine.py` run against the **old** binary with the **new** pin: **1 failed**
  (exactly the version assertion), **21 passed**. The harness discriminates; no other engine
  claim moved.
- Tool-set claims re-measured directly: `nav` still resolves exactly
  `{glob, grep, invalid, read}`; `list` and `websearch` still do not exist.

## The trap this hit, now recorded at `PINNED_VERSION`

**The dumps are not deterministic.** `external_directory` rules are generated by walking
`~/.claude/skills` and `~/.config/opencode/skills`, so their order follows readdir — **two runs
of the same binary differed on 434 lines**, which reads exactly like a resolver-ordering
change. The first cross-version diff was dominated by that noise and looked like a real delta.
Canonicalise before comparing, and validate the canonicalisation with a same-binary control
(two runs must collapse to identical) before believing any cross-version verdict. Ordering
*semantics* stay pinned behaviourally by the engine-vs-model conformance tests, which survive
sorting and pass on both binaries.

## Scope of the re-derivation, stated in place

23 claim lines move to 1.18.16. **Six references deliberately stay at 1.18.4**, and now say
why in the file rather than looking like drift:

- the hook-behaviour claims (`permission.ask` never fires, `tool.execute.before` does) in
  `README.md` and `plugin/guard.js` — these need a *running hook* to observe and were **not**
  re-measured here;
- the dated pipe-truncation incident record and the cross-host drift example, which are
  history and should not move.

## Gate

| tier | result |
|---|---|
| nix sandbox `pytests` | `TOTAL collected=9658 passed=9657 skipped=1 failed=0` · `RESULT: PASS` |
| nix sandbox `nodetests` | `TOTAL suites=4 files=36 tests=1178 pass=1178 fail=0` · `RESULT: PASS` |
| dev-host tier (the one that can see the pipe bug) | 670 passed across the three opencode suites |

Both sandbox verdicts were read from `nix log <drv>`, not from an exit code — the second run
was a **cached** derivation and printed an empty build log with `rc=0`.


ZacxDev:
## Audit outcome + a correction to this PR's body

Adversarially audited before merge. Verdict: **safe to merge, no deploy-blocking findings** —
but it found four real misses, now fixed in `b88a3e3`.

### 🔴 Correcting the body above

The body says *"Six references deliberately stay at 1.18.4"* and accounts for them as hook
claims plus two historical records. **That accounting was incomplete**: `scripts/opencode/README.md:229`
(the `env`-key / zsh-startup-files claim) is a sixth site neither bullet describes. It is
correctly left at 1.18.4 — it is another not-re-derived claim — but the body did not name it.
Correcting here rather than silently editing the body, since a reviewer may already have read it.

### What the audit found that I missed

My re-derivation grepped `scripts/opencode/` plus the two test modules **and stopped there**.
Four files outside that scope kept the old number, and two of them are the files the pin names
as its own mechanism, asserting it in the **present tense**:

- `nix/pkgs/tools/default.nix` — *"MEASURED at flake.lock's nixpkgs rev 9bc02893134c,
  `pkgs.opencode` **is** 1.18.4 — store path /nix/store/64n428…"*
- `flake.nix:220` — *"CI tests the exact binary the hosts deploy (1.18.4 at rev 9bc02893134c)"*

Both false the moment the lock moved; both also quoted an annotation this PR had retitled, so
the quote no longer matched its source. Plus `claude/opencode-addendum.md`, which is
**concatenated into `~/.config/opencode/AGENTS.md`** — the running agent was being told 1.18.4
while the repo said 1.18.16 — and the third copy of the `@`-import sentence in `nix/home.nix`.

### The actual fix is the guard, not the seven edits

Nothing in the suite could see this class: the version assertion pins the *binary* and is blind
to a doc that disagrees with either side. `b88a3e3` adds two tests over an enumerated
`PIN_SURFACE` — every opencode version literal is either `PINNED_VERSION` or an enumerated
historical exception with its reason in the source, and the allowlist is a **ledger** that also
fails when the set *shrinks*.

Three traps found while writing it, all recorded in place:

- **`\b1\.18\.4` never matches `v1.18.4`** — `v` and `1` are both word characters, so there is
  no boundary. The first draft passed while `README.md:229` was stale: the guard was blind to
  exactly the spelling it was hunting.
- The sweep must be scoped to opencode's version **series** — `nix/home.nix` and `flake.nix`
  also carry dunst `1.13.2` and playwright `1.57.0`/`1.61.1`, and a gate that goes red on
  unrelated bumps is a gate people switch off.
- Allowlist snippets must sit on the **same line** as their version and carry no version
  literal of their own, or they match themselves when the file is scanned.

Mutation-tested: reintroducing the audit's exact defect (`flake.nix` back to 1.18.4) fails the
guard with its own message naming `flake.nix:220`.

### Known gaps, stated rather than implied

- `opencode.jsonc`'s header re-label now asserts 1.18.16 provenance for a few facts outside
  this PR's enumerated evidence (glob semantics, `ask` under `opencode run`, `small_model`'s
  scope, the deprecated-keys list). The auditor independently re-derived the glob half and it
  **holds**; the rest are carried forward unverified.
- Nothing in CI exercises the engine's own `Wildcard.match`, so a matcher change is invisible
  to the suite *and* to an order-insensitive dump compare. Pre-existing, not closed here.

### Independently verified by the audit

Order-sensitivity — the key question, since an order-insensitive dump compare could have hidden
a last-match-wins change. `test_engine_bash_ruleset_matches_the_python_model` compares the
engine's array by **ordered list equality** against a hardcoded port, not against anything read
back from the engine. Moving the agent block from last to first turned **6 tests red**. The
claim holds.

Blast radius: both trees' `activationPackage` build (rc=0, so a switch will not fail); closure
diff is **142 package version bumps**; **52 of 602 managed files differ**, all store-path
rebases at unchanged package versions plus two espanso YAMLs whose re-serialisation is
semantically identical under PyYAML. `~/.claude` (181 entries walked) byte-identical.

Gate after the fix round: `TOTAL collected=9660 passed=9659 skipped=1 failed=0` · `RESULT: PASS`.

ZacxDev:
## Round 2 — delta re-audit, and it found what the fix round introduced

Re-audited `ae17ba8..b88a3e3` blind to my reasoning. It confirmed the four prior findings
fixed (and independently re-derived the store path and nixpkgs rev **from the lock**, not from
my commit message), then found three gaps in the guard I added. All three were **proven open by
a surviving mutant**, not argued. Fixed in `f76627e`, each now proven closed the same way.

### 🔴 The guard advertised a scope it did not have — and this PR falsified four claims outside it

`PIN_SURFACE`'s header said *"every file that spells a version-keyed claim"*. One grep falsified
that: eight more tracked files carry opencode version literals, and **four were present-tense
deployment claims this PR made false** — true at the merge-base, false the moment the lock moved:

- `scripts/browser-bridge/README.md:1653` — *"**Both hosts run 1.18.4** and both resolve browser-only"*
- `scripts/browser-bridge/README.md:1744` — *"verified on 1.18.4, **which is what BOTH hosts run**"*
- `scripts/browser-bridge/reference/agent.md:29, :176` — the same claim, twice more

Exactly the class as the first round's findings 1 and 2. Fixed by **deleting the duplication**
rather than updating four copies: they now name the pin instead of copying its value, so they
cannot go stale again. Both files are in the surface, and the header now states the real
boundary and **names** what is deliberately outside it.

### 🟡 Two more holes, both mutation-proven

- `OPENCODE_SERIES` was coupled to `PINNED_VERSION` by **prose only** — a claim keyed to
  `1.17.20`, a version these hosts genuinely ran, sailed through green. Now a module-level
  assert.
- Allowlist snippets were plain substring matches, so **reusing one on a new stale line was
  silently exempted**. The ledger now requires each entry to match *exactly one* version-bearing
  line.

### Reverted: the @-import claim

The first audit asked for consistency here and I gave it; the re-audit was right that this was
wrong. The claim was re-*spelled* to 1.18.16 but nothing re-derived it — it needs an
all-tools-denied agent making a real model call. It is back at **1.18.4 and marked NOT
re-derived** in all four places. Fixing a consistency complaint by making a claim false is the
worse trade.

### 🔴 A lesson about the mutation harness itself

The series mutant first came back **GREEN**, and I nearly reported it as a surviving gap. The
guard was working the whole time — my harness lied. `cp -a` of a Python tree **preserves
`__pycache__`**, and `"1.18"` → `"1.19"` is a **same-length** edit, so the `.pyc` staleness
check (source mtime + size) never invalidated it: pytest imported the *cached* module and
reported `2 passed` against source it never read.

Purge `__pycache__` in any scratch copy before mutating, and treat a same-length mutation as the
case most likely to be silently skipped.

### Mutation matrix after the fix

Each on a `.git`-stripped, `__pycache__`-purged copy:

| mutant | before | after |
|---|---|---|
| stale live claim in `browser-bridge` | GREEN 🔴 | **RED**, names the file |
| snippet reuse on a new stale line | GREEN 🟡 | **RED**, "matches 2 … must match exactly 1" |
| `OPENCODE_SERIES` replaced not appended | GREEN 🟡 | **RED** at import, its own message |
| unmutated control | — | GREEN |

### Gate, both tiers, after the fix round

`pytests` → `TOTAL collected=9660 passed=9659 skipped=1 failed=0` · `RESULT: PASS`.
`nodetests` → `tests=1178 pass=1178 fail=0` · `RESULT: PASS`. All three changed `.nix` files
parse.

### Still open, stated rather than implied

- `opencode.jsonc`'s header asserts 1.18.16 provenance for a few facts outside the enumerated
  evidence (`ask` semantics under `opencode run`, `small_model` scope, the deprecated-keys
  list). The auditor re-derived the glob half independently and it holds; the rest are carried
  forward unverified.
- Nothing in CI exercises the engine's own `Wildcard.match`. Pre-existing; not closed here.
- `PIN_SURFACE` only rots **downward** — a deleted path fails loudly, a new file carrying a live
  claim is invisible. Now said explicitly in the source.

ZacxDev:
## Round 3 — the finding was against round 2's *diagnosis*, and it was right

I took an audit's premise as established and acted on it without measuring. That was the error.

Round 2 said this PR "falsified four present-tense claims" reading *"both hosts run 1.18.4"*.
**It did not.** A `flake.lock` bump changes what the lock **pins**; it changes nothing about what
a host **runs** until `home-manager switch` — this repo's own CLAUDE.md says exactly that, and
it is why `ship.sh` exists. Measured:

```
workbench  /nix/store/64n428…-opencode-1.18.4/bin/opencode   → 1.18.4
laptop     /nix/store/64n428…-opencode-1.18.4/bin/opencode   → 1.18.4
flake.lock / PINNED_VERSION                                  → 1.18.16
```

So those four sentences were **true** at the merge-base and are **true now**. Rewriting them to
"both hosts run the pinned opencode version" made them **false** — false in exactly the way this
PR is about. Two things made it worse than a wrong number:

- **The falsity became uncheckable.** With no literal left, the guard added two commits earlier
  could never see it — in the two files that same commit added to its surface.
- The commit message declared this trade *"the worse trade"* three paragraphs above, while
  correctly refusing it for the @-import claim. I applied the principle in one place and
  violated it in another, in a single commit.

### Fixed in `ea86ce0`

The literal `1.18.4` is restored as *what the hosts run*, and each site now also states that the
lock pins 1.18.16 and that the browser-only resolution was re-derived there — so the claim is
true, checkable, and survives the deploy gap. **I re-derived that myself rather than taking the
auditor's word**: `debug agent browser-agent --pure` under both binaries resolves to exactly
`ENABLED=['browser']`, 13 tool keys, dumps identical cross-version.

Ledgered under a new category, **DEPLOYED-STATE** — genuinely distinct from "historical", because
these track the deploy and change on the next `ship.sh`, not on this merge. The measurement is
recorded in the source so the next reader need not re-derive it.

Two more, both introduced by the previous round:

- `scripts/browser-bridge/README.md` claimed 1.18.16 *"lives in one place, `PINNED_VERSION`"* —
  false on both halves. The literal is in **10 tracked files**, and the source of truth is
  `flake.lock` → nixpkgs rev → `pkgs.opencode`; `PINNED_VERSION` is a downstream *assertion* of
  it. Removed.
- The `PIN_SURFACE` section **title** still read "every file that spells a version-keyed claim" —
  the exact overclaim round 2 flagged. The previous commit *appended* a correction and left the
  false heading, so a reader who stopped at the title got the wrong scope. Rewritten.

### Correcting the previous commit message

It credited the new `OPENCODE_SERIES` assert with closing *"a claim keyed to 1.17.20 sailed
through green"*. **It does not.** The assert checks the pin's own series is in the tuple; it does
not widen the tuple, and a `1.17.20`-keyed live claim is still invisible. That gap is
pre-existing and **open**, as is "a new file carrying a stale claim is invisible". Recorded as
follow-ups rather than claimed as fixed.

### Gate

`pytests` → `TOTAL collected=9660 passed=9659 skipped=1 failed=0` · `RESULT: PASS`.
`nodetests` → `tests=1178 pass=1178 fail=0` · `RESULT: PASS`. Read from `nix log`, not exit codes.

PR #470, opened 2026-08-14
feat(census): the count could not tell an appending writer from a dead one

ZacxDev:
The handoff's next-step 3 was **"watch the census; if it stalls, writers aren't sticking."**
The census cannot detect a stall.

## The defect

Every count `census()` reports is a count of **creation events**. An entry is one row
forever, and `created_by:` is stamped once at creation and never updated. So a writer that
spends a week appending to existing entries moves *none* of the numbers, and `total` reads
identically whether the store is being worked hard or is dead.

**MEASURED 2026-08-13:** the live store read **34 entries across two readings 40 minutes
apart**, while its git history showed **7 new entries and 9 appends that same day** across 5
scopes, with a write in every hour from 12:00 to 19:00. Reading the flat total as "the writers
are not sticking" would have been exactly backwards. (While writing this PR the total moved to
35 with `newest write: 0m ago` — a writer was active mid-review.)

## The fix

`census()` also reports write **activity**: newest write, and entries touched in the last
24h / 7d. Counts stay the coverage instrument; these are the stall detector; the docstring
says not to substitute one for the other.

**Derived from mtime, deliberately NOT the store's git log.** The git log is the obvious
source and the wrong one: commits are batched by an hourly
`analyze-service-index-commit.timer`, so a git-derived reading lags real writes by up to 60
minutes and a mid-hour check can report silence during an active session. mtime also keeps
this subprocess-free and hermetically testable — `now` is injectable, no test sleeps.

Stated in the docstring rather than left to be discovered: mtime cannot see **how many**
appends landed (an entry touched 30× counts once — the rendering says "entries TOUCHED", never
"writes"), cannot see **who** touched it, and will read a **restore** (`git checkout`, `cp -a`,
a fresh clone) as a burst of activity that never happened.

## Tests

7 new. The load-bearing one is `test_activity_moves_when_the_counts_do_not`: append to an
existing entry, create nothing, assert every count holds still while the activity reading
moves.

Mutation battery — 5 **semantic** mutants, not deletions — all killed, each by a distinct test
at a distinct assertion:

| mutant | killed by |
|---|---|
| activity never computed | `test_activity_moves_when_the_counts_do_not` (**by name**) |
| boundary `<=` → `<` | the inclusive-boundary test |
| future mtimes dropped | the clock-skew test — silence is the one direction this must never fail in |
| README counted as activity | the README test — a policy-sheet edit is not a heartbeat |
| `newest` = first, not max | the newest-write assertion |

Battery validated: the scratch copy was confirmed a real independent file, not a symlink back
into the worktree, and the worktree was verified byte-identical afterwards.

## Also in this PR

The handoff doc's next-step 3 now points at the activity lines and records the number actually
worth watching: **12 of 13 stamped entries are `handoff`**, so the store is effectively
single-writer and `/analyze-service` has an n=1 track record over the whole instrumented
period.

It also records the doc-hook result: a **dispatched agent** with an ordinary kickoff read the
handoff as tool call 1 and executed `subsystem_recall.py` as tool call 2, with **0 `Skill`
calls** — the case `#446` failed. Counted from its transcript, not its self-report. The
remaining contamination is structural (the doc describes the experiment and `Read` returns the
whole file), so the doc now says to measure organically rather than stage a third probe and
call it clean.

## Gate

`nix build .#checks.x86_64-linux.pytests` → `TOTAL collected=9682 passed=9681 skipped=1
failed=0` · `RESULT: PASS`. Read from the log content, not an exit code.


ZacxDev:
## Audit outcome — no 🔴, six 🟡, all fixed in `ed165f0`

The core finding is fair and I'd made the mistake cleanly: **the mutation battery mutated
`census()` five times and `render_census()`/`to_json()` zero times.** Both output surfaces
survived *deletion* with a fully green suite — dropping the entire `newest write:` line passed
559 tests, and so did deleting both new JSON keys. `--census --json` had no test anywhere in the
repo.

That's the isolation-seam shape, aimed at exactly the wrong half: the handoff doc had just
started telling the operator to **read the activity lines**, and the activity lines were the
unguarded part.

### Five tests over the output surface

Asserting **shape** — a label with a number beside it — not prose, so they can't be satisfied by
a sentence that merely spells the right word:

- the newest-write line carries an **age** (`3.0h ago`), not just a label
- every configured window renders exactly once **with its count**
- a future stamp renders as `in the future`, never as an age — that phrase is the only loud tell
  for the absorbing fail-open state where one clock-skewed file counts inside every window
  forever, and **an unguarded tell is not a tell**
- `to_json()` carries both keys, the epoch value, and the `int`→`str` window-key coercion (an
  undocumented contract a reader will otherwise index as `[24]` and get a `KeyError`)
- an empty store reports `null`, not a bogus epoch

### Mutation-verified against the mutants the audit proved survived

| mutant | before | after |
|---|---|---|
| drop the whole `newest write:` line | **GREEN** | RED — *"newest-write line missing entirely"* |
| `_age_phrase` returns a constant | **GREEN** | RED |
| delete both JSON keys | **GREEN** | RED |
| hard-code `newest_write_epoch` to `None` | **GREEN** | RED |
| render only the first window *(untried by the audit)* | — | RED |

### 🔴 A factual error in my docstring, corrected by measuring

It said a restore — *"`git checkout`, a `cp -a`, a fresh clone"* — stamps every file with a new
mtime. **Measured:** `cp -a` **preserves** mtime, and so does `git commit`; plain `cp` and
`git clone` reset it. The docstring now carries the measured table instead of the guess, and adds
the case it omitted: a rewrite with **identical content** still reads as a touch, which
`analyze-service`'s confirm flow does on purpose.

### The rest, each verified before acting on it

- `render_census(c, now=…)` was **dead** — nothing passed it, and the CLI read the clock *twice*
  per report, so `census()` and `render_census()` could disagree about "now" inside one output.
  One read now, passed to both.
- *"lags by up to 60 minutes"* was wrong in two places. The timer is `OnCalendar=*-*-* *:00:00`
  with `RandomizedDelaySec=10min` → **~70m**.
- `touched_within=touched` aliased a live mutable local into a frozen dataclass while its three
  siblings all copied defensively. Now copies.
- Four stale self-descriptions of the flag this PR extends: the argparse `--help`, the module
  contract summary's `census(store_root)` signature, the `_EXIT_MODES` label, and — **inside the
  file the first commit edited** — the handoff doc's copy-paste runbook at `:217`, still saying
  *"anchor was 21 / 1 scope / 21 unstamped"* eighty lines below the rewritten next-step 3 saying
  the total cannot detect a stall. A resuming agent copies the runbook block, not the prose.

### Not changed, recorded instead

The git log is wrong for **freshness**, not useless — it remains the only source for *how many*
appends landed and *who* made them, and this feature's own evidence came from it. The docstring
says that now rather than overstating the case.

### Gate — on the MERGED tree, both tiers

`origin/main` merged in (#469 had landed since the fork; zero file overlap).
`pytests` → `TOTAL collected=9689 passed=9688 skipped=1 failed=0` · `RESULT: PASS`.
`nodetests` → `tests=1178 pass=1178 fail=0` · `RESULT: PASS`. Read from `nix log`, not exit codes.

### Independently confirmed by the audit, worth recording

Nothing but a writer touches these files: in **every** scope the last autocommit timestamp is
later than that scope's newest file mtime, and `commit.sh` uses no `checkout`/`restore`/`reset`/
`clean`/`gc`/`clone`. Timezone-safe (identical output across four zones incl. a 30-min offset
with DST). `Census(` has exactly one construction site repo-wide, all-keyword — the two new
required fields break nothing.

PR #471, opened 2026-08-14
feat(session-manager): the default view has had no age since #419 because the only writer was switched off

ZacxDev:
Phase 1 of `claudedocs/spec-agent-activity-ledger.md` (#428) — **supersede, not remove**. fuzzyclaw is untouched and still opt-in; both writers are live.

## The defect

#419 switched fuzzyclaw off on a dogfood finding that it "contributes nothing". True of its `status` field — every live row read `paused` — and **false of the source**: it was also the only supplier of `age_secs`, of the `stale` bucket derived from it, and of the `claude_session_id` the ClickHouse join needs.

Measured on the workbench 2026-08-12, the shipped default view:

| | measured |
|---|---|
| rows with an age | **0** |
| rows with a `claude_session_id` | **0** |
| `stale` bucket | **absent** |

Nothing in the output stated any of that, which is why it survived weeks of green scans. The summary now carries `rows_with_age`, `rows_with_session_id` and `age_sources`, and the table prints them — the line that would have made #419 visible on the day it shipped.

## What lands

| | |
|---|---|
| `scripts/lib/agent_ledger.py` | record, write/prune, read protocol, live-join filter |
| `scripts/claude-hooks/agent-ledger-hook.py` | writer 1 (Claude Code) — SessionStart / UserPromptSubmit / PostToolUse / Stop |
| `scripts/session-manager` | reads it **per host**, joins on `window_id`, ledger wins over fuzzyclaw |
| `nix/home.nix`, `register-nudge-hook.py` | deploy + registration |

One module, loaded by explicit path from `session-manager` **and** deployed beside the hook, so writer and reader agree on the record by construction rather than by two copies staying in step.

Read **per host**, which fuzzyclaw structurally never was — its task files are local state, so every remote row carried a null age by construction. One `sh -c` per host over the same SSH transport as the tmux calls.

## Three decisions worth reading

**PostToolUse writes (throttled 30s), and that is load-bearing.** `classify_status` lets `stale` WIN over `busy`, so with only turn-boundary heartbeats a turn grinding past the 1h threshold would render `stale` while demonstrably working. Registered with **no matcher**, unlike the three Bash nudges — scoping it to Bash would stop the heartbeat during a long stretch of Read/Edit work.

**The record carries the tmux SERVER pid and the reader checks it.** tmux window ids restart at `@0` when the server does, so after a reboot — precisely when `tmux-task-resume.sh` rebuilds the workspace — yesterday's `@41` record and today's `@41` window collide, and a join on the id alone hands a fresh window a dead session's id and a multi-day age. The pid rides on the ledger's own sentinel line, so it comes from the same instant as the records it validates. Records whose generation **cannot** be checked are kept, counted separately as `generation_unchecked`, and never silently promoted to verified.

**The read protocol carries its own positive control.** `echo <SENTINEL>` precedes the records, because empty stdout from a swallowed command and a host with no records are the same bytes — publishing the second reading of the first is exactly the fabricated zero this change exists to undo. Without the sentinel every count is `None`, not `0`.

## A false docstring, retracted

`session-manager` asserted *"the pane format carries no `window_id`"*. That claim justified the entire two-key guard/join design (one 🔴, three fix rounds, three audits). Measured 2026-08-13: `tmux list-panes -a -F '#{window_id}'` returns a non-empty id on **47 of 47** panes. `PANE_FORMAT` carries none — a fact about a format string anyone may change — but tmux supplies one. Putting it in `PANE_FORMAT` would make the non-atomic `list-panes`/`list-windows` skew catchable and collapse most of that design; flagged as the follow-on, deliberately not done here.

## Verification

**Gate.** `nix build .#checks.x86_64-linux.pytests` → `TOTAL collected=9761 passed=9760 skipped=1 failed=0` (the one skip is the pre-existing pinned repo-cos one).

**Mutation sweep — 34 realistic mutants across all three files: 34 killed, 0 survived, 0 unscored.** Green control, and a comment-only negative control that **survived as required**, so the sweep scores the mutation and not something else. The harness is built against four traps this repo has been bitten by: it imports every mutant before scoring it (a syntax error scored as a kill has happened three times here), counts `N failed` rather than reading an exit code, asserts each pattern applied exactly the expected number of times, and builds a fresh tree per mutant — `session-manager` is extensionless and its `SourceFileLoader` `.pyc` cache keys on `(int(mtime), size)`, so a mutate/restore loop silently runs the previous mutant's bytecode.

A fifth trap surfaced during the sweep and is worth recording: the first `render_table` mutant was `out.append("" or "  ages: …")`, which type-checks, applies cleanly, imports fine — and evaluates to the *same string*. It "survived" and read as a coverage gap. A mutant that changes the text without changing the behaviour is one this harness cannot detect in general.

**The regression is pinned as a PAIR**, not as a single number: the same fixture with the ledger off (0 ages, 0 session ids, no `stale`) and on (2, 2, `stale` returns). A reader that returns 0 ages is indistinguishable from one wired to nothing, so the "on" number means something only beside a control that moved.

**Live on the workbench**, after the switch (not merely deployed):

```
▸ AGENT LEDGER (1 live of 1 record(s))
  laptop     0 live of 0 (tmux pid 3737)
  workbench  1 live of 1 (tmux pid 4025325)
  ages: 1 of 74 row(s) have one (ledger=1, none=73)   session ids: 1
```

The joined row is `scratch20:1` / `@101` / `Yarrow`, `age=13.9s source=ledger status=busy`, carrying the real session id. The record was written by the hook **firing naturally on a real turn**, before any manual probe — it was already on disk when the ledger directory was first listed. Both hosts answered with their own distinct tmux server pid, which is the generation guard's own positive control.

⚠ **Stated plainly: 73 of 74 rows still have no age, and that is expected, not a gap.** Every other agent window is running a Claude process that started *before* the hook was registered, so none has written yet; coverage grows as each session takes its next turn. What is closed is that the number *can* be non-zero and *says which writer supplied it* — before this, it was structurally 0 with nothing in the output admitting it.

⚠ **The laptop's READ path is verified live** (it answered, pid 3737, a measured 0 records). Its **writer is not deployed yet** — that happens on `ship.sh` after merge. Verified on half the fleet until then, and named as such.

## Deliberately not in this PR

Writer 2 (opencode plugin), writer 3 (clawgate agents + the `kind` field of spec §4), the §5 bar inversion, and the four-phase fuzzyclaw removal. The spec is explicit that writers must never be pulled first.


ZacxDev:
## Audit round — one 🔴 fixed, and it was a defect in the PR body's own claim

An adversarial audit found the PR shipped a **false caveat in every run and every `--json` payload**, so correcting it publicly rather than silently editing the body above.

`CAVEATS["fuzzyclaw_scope"]` said a remote row carries null `claude_session_id` / `age_secs` and *"is NEVER labelled `stale`"*. True of fuzzyclaw; **false of this change**, because the ledger is read per host — which the PR body itself advertises as the structural win. It is the machine-readable caveat, so a `--json` consumer following it would discard exactly the three fields this work exists to add.

It survived because **I blinded its own guard**. `base_gather()` is pinned `use_ledger=False` so the pre-existing assertions keep their meaning — which also made the caveat-vs-code test, whose docstring says it fails "in both directions", structurally unable to see the ledger. That test now runs in **both** configurations, and a `ledger_scope` caveat states the corrected scope *beside* the narrowed one rather than replacing it, so neither reading is left available.

### The other findings

**Records were keyed on `window_id`, and a tmux window can hold two claude panes.** The throttle is session-scoped, so alternating writers never throttle: measured, 10 writes from two sessions inside a 10s window all landed, the single file named whichever wrote last, and `index_by_window` could not see any of it — there was only ever one file to compare. Since `row["claude_session_id"]` is the sole carrier into the ClickHouse join, a shared window resolved to an arbitrary one of its two agents. Now keyed on `pane_id`, with `window_id` still the join key: two panes produce two files with one window id, which is exactly the shape the conflict detector already reports. Observed live after the change — window `@68` held two records and the table printed `⚠ LEDGER CONFLICT … claimed by 2 record(s) — NEWEST wins`.

**Three mutants survived the first sweep, all the same shape — the constant was asserted, the call site was not:**

| survivor | why it survived |
|---|---|
| `throttle_secs=None` at the hook's call site | the behavioural test was **vacuous** — it unset `TMUX_PANE`, so one-file-per-key guaranteed "one record" with the throttle deleted |
| deleting the `AL.prune()` call | `PRUNE_EVENTS` asserted as a set, never exercised |
| dropping `max(0.0, now - last)` | clock skew only became *reachable* because the ledger is the first source of remote ages, and a negative age passes `age >= threshold` as fresh |

Each now has a test; each is now a mutant in the sweep. This is the "declarations are not instances" trap, three times in one file.

**Corrected claims.** "Writer and reader agree by construction" holds only at the instant of a switch — the hook runs the nix-store copy while `session-manager` loads the repo working copy, so between `git pull` and `home-manager switch` they are different code. And the registration comment's cost model was wrong: it said one process per tool call while the hook was shelling out to `tmux` *before* consulting the throttle — two, every call. That ordering is reversed (the pane comes free from `$TMUX_PANE`) and pinned by a test.

**Also from the audit's nits:** the read uses `awk 1` rather than `cat`, because a record file with no trailing newline welds onto its glob-neighbour and *both* are lost (measured: 3 written, 1 parsed, 2 unparseable) — a convention across processes is not a guarantee, and spec §4 adds two more writers. `prune` reaps leaked `.ledger.*.tmp` files, which only it can see. A host that answered neither tmux call no longer gets a third command it can only fail. The dead `unjoinable` field is gone; `SKILL.md` documents the ledger.

### Re-verified from scratch, because a review fix resets the gate

- `nix build .#checks.x86_64-linux.pytests` → `collected=9772 passed=9771 skipped=1 failed=0`.
- Mutation sweep extended to 44: **43 killed, 0 survived, 0 unscored**, control survived. The three audit-found survivors are mutants in it now, as is re-asserting the retracted caveat.
- 🔴 **This round's own new defect was caught by an existing gate**, not by me: `scripts/tests/test_runtime_shebangs.py` failed because the new tmux stub wrote `#!/usr/bin/env bash`, which is dead in the nix build sandbox. Now via `testlib.mockbin.write_exec`, the repo's one definition — the seventh site to re-derive that lesson.

A delta re-audit of the fix round is running; findings will follow.

PR #481, opened 2026-08-14
feat(handoff): "belongs in a skill" was a disposal, not a route — the lesson died there

ZacxDev:
The handoff index step has exactly one output: append to the index, or don't. The skill's own
words name a skill file as *"a legitimate alternative home"* — but **only as a reason to
decline**, and nothing routes the lesson there. So a *correct* decline discharges the
obligation, and the knowledge ends its life as prose in a transcript — the medium this store
exists to outlive.

## The incident

A session declined both windows correctly (`looked-at-nothing`: 0 paths under cwd, 15 outside —
worktree isolation working as designed; `no-match` on one `claudedocs/` path), correctly
concluded its gotchas belonged in a skill, named `.claude/skills/pyroscope/SKILL.md`, and
stopped.

**That skill does not exist.** `obs-read` already owns Pyroscope, in its description — so acting
on the note later would have created a duplicate skill and split the domain. The gotcha it
carried (Pyroscope's `max_query_length: 1d`) was recorded nowhere.

Verified before building anything: zero matches for `unfiled|debt|file it|follow-up|carry
forward` in the handoff skill. No step existed.

The incentive was backwards too — declining costs one sentence; filing costs finding the owning
skill, writing under the always-on listing budget, and `git add`ing a file the flake otherwise
silently omits. **The cheap path was the lossy one.**

## The fix

The `ROUTE OUT` block already fires on exactly the two dead ends where this happens. It gains a
`SKILL HOMES` half, and step 4 gains the rule that "belongs in a skill" must finish **in the same
turn**, under the same one-diff / y-N gate as the index write.

```
SKILL HOMES — if the lesson is a durable ops-gotcha rather than a subsystem, the index
is the wrong home and a skill is the right one.
    claude/skills/obs-read/SKILL.md   (matched 'pyroscope')
    claude/skills/activity/SKILL.md   (matched 'homelab')
  🔴 Derived from path stems + the scope — NOT from what the session was ABOUT, which
  this tool cannot see. An empty result is not 'no skill owns this'. Search YOUR term:
    grep -ril '<your-term>' ~/.claude/skills/*/SKILL.md
  Edit the REPO source (`claude/skills/…`), never `~/.claude/…` — that is a read-only
  store symlink. A NEW file must be `git add`ed or the flake silently omits it. If
  nothing owns it, say so as an UNFILED item, with the term you searched.
```

**It cannot know the domain and must not pretend to.** The term came from what the session was
*doing*, which need not appear in any path it touched. So the block does two clearly-labelled
things: matches terms it can actually derive, and hands over the exact search for the term only
the agent has. A miss says *"no skill matched a derivable term"*, never *"no skill owns this"* —
a silent miss would discharge the duty exactly like the incident did.

## Two design points the smoke test forced — both had already broken it

- **Whole path, not basename.** The first draft tokenised `os.path.basename`, discarding every
  **directory** component — where the domain lives. On the very case this exists for,
  `src/pyroscope/query.go` yielded only `query`: the right answer wasn't ranked low, it was
  **absent from the candidate terms**.
- **Ranked by term specificity.** Unranked, the real lead drowns — on the live catalogue a path
  carrying `pyroscope` also yields `query` and `homelab`, matching 2 skills each, so `obs-read`
  came out **fourth** and a cap of 4 was one generic term from hiding it. On a breadth **tie**
  the longer term wins: measured, plain alphabetical ordering put `mailbox` above `obs-read` on
  a coin flip.

Matching is **name + description only, never the body** — the description *is* the routing
surface (it's what the always-on listing carries), and matching bodies returns half the
catalogue for a common word.

## Tests

8 new. Mutation battery of 7 semantic mutants, **all killed** — each verified to have *applied*
(one earlier anchor silently missed after a refactor, and its green was discarded rather than
reported):

| mutant | result |
|---|---|
| basename-only tokeniser | RED |
| ranking dropped, alphabetical only | RED |
| haystack widened to the body | RED — *survived until a test was written for it* |
| miss renders silently | RED |
| fallback `grep` command dropped | RED |
| absent root raises instead of degrading | RED |
| catalogue returns `()` though populated | RED (positive control: it is wired to something) |

Degrades rather than dies: an absent skills root is ordinary, and a malformed `SKILL.md` is
skipped rather than taking the dead-end report down — the same rule the reader learned from a
wrapped `aliases:`.

## Gate

`nix build .#checks.x86_64-linux.pytests` → `TOTAL collected=9866 passed=9865 skipped=1 failed=0`
· `RESULT: PASS`. Read from the log content, not an exit code.


ZacxDev:
## Audit: both 🔴s were real, and the first is the sharpest finding of the series

**The feature could ship inert.** Both production call sites of `render_skill_homes` in
`render_text()` had zero coverage — the auditor deleted them and the suite stayed green (578
passed). Every one of my 8 tests called the helper directly with an injected root, so a 7-mutant
battery proved the *computation* and nothing about whether it is ever *reached*.

The lesson was already written down **eight lines away**. `render_route_out`'s own seam test
says: *"a block added to one exit and not the other is exactly the shape that leaves the
commonest path uncovered."* I added the block to both exits and asserted neither.

Same shape twice: the entire `handoff/SKILL.md` paragraph could be deleted green, in a repo that
already has `HANDOFF_SENTENCES` for precisely that.

**Fixed in `b130ea7`**, using conventions this file already had — three seam tests (both dead
ends + a negative control that a RESOLVED run stays silent, without which always-print would
pass) and one `HANDOFF_SENTENCES` row. Mutation-verified against the auditor's own kills:

| mutant | before | after |
|---|---|---|
| delete BOTH call sites | 578 passed | **2 failed** |
| delete ONE (the `looked-at-nothing` exit) | green | **1 failed** |
| delete the SKILL.md paragraph | 999 passed | **1 failed** at the sentence pin |

🔴 My negative control initially **skipped itself** when the fixture didn't resolve. A test that
skips is worse than no test — it reports safety while checking nothing. It now uses the same
resolving paths as its sibling and asserts the precondition.

## The other findings, each verified before acting

- **`skill_catalogue` walked the tree outside its `try`.** `iterdir`/`is_dir`/`is_file` raise
  `PermissionError` on an unreadable skill dir, and the `if not root.is_dir()` pre-check is
  TOCTOU against a concurrent `home-manager switch`. `main` renders *outside* its `TouchError`
  handler, so either escape surfaced as a traceback and the dead-end report — the thing the
  protocol keys on — never printed at all. This helper is an **extra**; it must never be why the
  report fails.
- **`render_text`'s "Deterministic: same report in, same bytes out" became false** the moment
  this read host state. Now scoped to reality, including the two-tier fact you measured:
  `flake.nix` exports `HOME=$TMPDIR/home` and nothing creates `.claude/skills` there, so **CI
  always renders the empty branch and a dev host always renders rows**. Never assert on those
  rows without injecting `skills_root`.
- **The ranking figures were wrong** — `query` and `homelab` match **4** and **3**, not "2 each".
  Conclusion survives, figure didn't.
- Type annotation said 3-tuple where the code appends a 4-tuple.

## 🟢 7 was the most valuable finding

The scope is a per-repo **constant**, so folding it in with path terms put the same skills in the
cap on every dead end — `devrc-dx` + `i3` taking two of four slots every time, hiding `repo-cos`
and demoting `bar` to third. Boilerplate is exactly the failure this block exists to avoid.

Path-derived hits now rank first, scope hits fill only the remainder, and a `looked-at-nothing`
run (0 paths → no signal at all) prints the search command rather than a constant that reads as
an answer.

Two of my own defects surfaced while fixing that, both caught by tests: the no-signal branch
keyed on the **terms** being empty rather than the **paths** (so `a/b.py` printed "no paths were
examined", which is false), and gating only the *message* still emitted scope rows with zero
paths — falsifying the docstring I'd just written.

Plus one more the live catalogue forced: **a skill named for a compound owns the compound.**
Splitting hyphens away left `scripts/repo-cos/scan.py` yielding only `repo` + `scan`, so
`repo-cos` matched the generic `scan` and ranked fourth. Whole components are candidate terms
now — but only when a piece survives the stoplist, since admitting `test_index` put back exactly
the noise the stoplist removes.

**All three displacement cases now rank the right skill first:**

```
scripts/repo-cos/scan.py -> repo-cos  ('repo-cos')
scripts/bar-status-poll  -> bar       ('bar-status-poll')
src/pyroscope/q.go       -> obs-read  ('pyroscope')
(no paths)               -> no rows
```

## Gate

`TOTAL collected=9874 passed=9873 skipped=1 failed=0` · `RESULT: PASS`, read from `nix log`
content. Not taken as sufficient on its own — that was the whole lesson here.

ZacxDev:
## Delta re-audit: the seam guard was **spelled**, not structural

The previous round closed the seam against **deletion** and left it open to **corruption** —
the RULES hazard *"a structural check type-checks past a wrong argument."*

Both new assertions matched strings in the block's **static header/footer**, which print in
every branch. Measured: changing both call sites to `render_skill_homes([], report.scope)` left
the suite **green at 586** while the feature degraded to printing a *false*
`(no paths were examined…)` on every dead end. Passing `""` for the scope was green too. The
guard held the block's *existence* and nothing about its *arguments*.

### Fixed structurally, not by adding another word to match

Neither call site threads `skills_root`, so a content assertion was impossible until the
catalogue is injected. The new test patches `SKILLS_ROOT_DEFAULT` and plants two skills with
pairwise-distinct domains — one reachable **only** from a path term, one **only** from the scope
term — so dropping either argument removes exactly one row.

| mutant | before | after |
|---|---|---|
| drop the PATHS at both call sites | **GREEN** | RED (2 tests) |
| drop the SCOPE at both call sites | **GREEN** | RED |
| delete the scope-FILL entirely | **GREEN** | RED |
| min-term-length floor 4 → 1 | **GREEN** | RED (single site now) |
| revert degrade to dropping the skill | — | RED |
| gut the SKILL.md paragraph to its headline | **GREEN** | RED |

Two traps hit while writing that test, both of which would have made it pass for the wrong
reason: two paths under one directory **cluster into a nomination**, which is not a dead end, so
the block correctly didn't render; and mutating `scope` on the frozen report tests a state the
tool never emits. It now uses the fixture's own scope, paths in different directories, and
asserts the status it depends on.

### The rest, each verified before acting

- **The floor predicate was written twice and the second copy was dead by construction** — a
  compound only reaches it once a piece has passed the floor, so it's ≥4 chars and contains a
  separator, hence never stoplisted and never a digit. Deleting it outright stayed green. One
  copy now.
- **The degrade handler was reached zero times by 586 tests** — the "malformed" fixture parses
  cleanly. Worse, the previous round silently changed its semantics from *keep name-only* to
  `continue`, which **drops** the skill, while the comment still said "degrade-don't-die". A
  skill's name is normally the most specific term available, so a broken `pyroscope/SKILL.md`
  would stop being findable by the word `pyroscope` precisely when it's broken. Restored, and
  pinned with a test that reaches the handler for real (an unreadable `SKILL.md`).
- **The protocol pin held one clause of a 1513-char paragraph.** Gutting it to 100 chars stayed
  green, silently taking the doc's only mention of `SKILL HOMES`, the fallback search, and the
  `UNFILED` instruction. Two more rows.
- **Two more wrong breadth figures** — the same class the previous round was fixing. `scan` is
  breadth **4**, not 3 (module *and* test docstring), and an inserted word *"other"* made
  FOUR/THREE an overcount. Re-measured on the live catalogue.

### Not changed, recorded

The exact-skill-**name** match isn't privileged over a same-breadth description match
(`scripts/clawgate/deploy.sh` → `bar` first, `clawgate` second on the alphabetical tie-break).
Pre-existing, and the class-level number is the real story: the owning skill ranks first in
**27/34** at this head versus **17/34** before the ranking work.

### Gate, both tiers

`pytests` → `TOTAL collected=9878 passed=9877 skipped=1 failed=0` · `RESULT: PASS`.
`nodetests` → `tests=1116 pass=1116 fail=0` · `RESULT: PASS`. Read from `nix log` content.

ZacxDev:
## Delta re-audit round 3: **no 🔴, no guard defects, no behavioural change**

Three claim-accuracy findings, two of them mine and both in the class this PR exists to fix.
Fixed in `bb108c9`.

### My previous commit message was false

It said the wrong breadth figure was corrected *"in the module AND the test docstring"*. **Only
the test moved.** The module's copy is **line-wrapped** — `matched on the generic\n# \`scan\`
(breadth 3)` — so the one-line replace matched nothing, reported success, and I wrote the claim
from the *intent* rather than the *result*. That is precisely the "a comment is a claim too"
failure, committed while fixing one. It also self-contradicted: *"(breadth 3) … behind three
scan-ish skills"* needs 4. Re-measured: `scan` → `adoption-scan`, `initiative-scan`,
`initiatives`, `repo-cos`.

And the fix for the previous round's overcount introduced a **new** wrong figure — *"three
others and two others respectively"*; `homelab` matches `activity`, `mailbox`, `tekton`, none of
them the subject, so three others, not two. The derived count has now been wrong in **three
consecutive drafts**, so it's gone: the comment states **totals only** (FOUR and THREE), which
are directly measurable and were never wrong.

### The protocol pin cannot do what widening it implied — and now says so

Going from one row to five did **not** stop gutting. Measured: a rewrite keeping all five
substrings and deleting everything else takes the paragraph **1513 → 119 chars and stays green**.
A substring pin is satisfied by its substrings, always — it catches **deletion** of a named
instruction and **drift** in its wording, not a rewrite that keeps the words.

A minimum-length gate would close that and false-fail every legitimate edit, which is how a gate
gets clicked through. So the limit is **recorded** rather than papered over, two more rows cover
the instructions that were still unnamed (the `grep -ril` fallback and the `git add` deploy
rule), and the realistic failure is verified caught — changing `grep -ril` to `grep -il` goes red.

### Two more of mine

- The structural seam test's docstring stated the **wrong mechanism**: dropping the PATHS removes
  **both** rows, not one (the scope fill is gated on `paths`, so the block flips to its no-signal
  branch). The test killed both mutants; the explanation was wrong.
- It asserted bare `pathside`/`scopeside`, which also appear in the interpolated tmp root inside
  the fallback `grep` line. Now asserts the full `claude/skills/<name>/SKILL.md` row.

### Recorded, not changed

The `looked-at-nothing` call site has no argument-level guard — mutating it alone stays green —
because `looked-at-nothing` is set iff `not source.paths`, so both of its arguments are
behaviourally inert there.

### Worth keeping from the audit's own work

`_terms_from` at `ceb0161` vs `40b950a`: **0 divergences over 200,708 cases** (adversarial +
fuzz), with a positive control (floor 4→5) showing 272 — so removing the duplicate predicate
provably changed nothing. And the degrade handler went from **0 hits across 586 tests** to **2**,
counter positive-controlled.

One correction to my earlier comment: the **27/34 vs 17/34** ranking figure I quoted came from the
*first* audit's report and exists only in this PR's comment thread — it is not in any commit
message or in the tree, so treat it as a measurement of that moment, not a pinned claim.

### Gate, both tiers

`pytests` → `TOTAL collected=9880 passed=9879 skipped=1 failed=0` · `RESULT: PASS`.
`nodetests` → `tests=1116 pass=1116 fail=0` · `RESULT: PASS`. Read from `nix log` content.

PR #482, opened 2026-08-14
feat(session-manager): a clawgate dispatch has no pane, so `claude: None` would have counted every one of them as a bare shell

ZacxDev:
Adds `kind` — the entity axis — to every row and makes the roll-ups count by it.

**No scan changes today.** Every row is `kind: tmux`, and a tmux-only summary is byte-identical to before. What changes is what becomes impossible later. Verified live: 78 rows, `summary.kind = {"tmux": 78}`, and **no `cluster` key in any bucket**.

## Why now, before writer 3

spec §4 decided clawgate dispatches join the **same** collection as tmux rows with the tmux-only fields null. That decision is right and was under-specified: it named a row field but never said what the roll-ups do with a row that has no pane. Deciding that *after* such a row exists means deciding it against a miscount already in the output.

`summarize` read `"claude" if row["claude"] else "shell"`. `row["claude"]` is `pane_current_command =~ /claude/` — a fact about a **pane**. A dispatch with no pane has `claude: None`, which is **falsy**, so every cluster agent would have been filed under `shell`: an agent counted in the bucket that means *"nobody is working here"*.

That is the claude/shell conflation this summary already exists to prevent, one axis over — the same defect that once published `idle: 17` for 12 agent windows + 5 bare shells. Every fixture in the suite is tmux-only, so nothing already here could see it.

## What landed

- **`KINDS = ("tmux", "cluster")`** — closed vocabulary; `cluster` enumerated but **not produced**.
- **`row_class()` is the one classifier**, used by both the per-status buckets and the top-level totals so they cannot disagree. `shell` was open-coded as `not r["claude"]` at the top level and `else "shell"` in the buckets — the shape that ends up wrong at exactly one site.
- **Keyed on `kind`, not on `claude is None`.** A tmux pane whose command did not parse *also* has a null `claude`, and that row genuinely belongs in `shell`. Two different nulls; only `kind` separates them.
- **A missing/unknown `kind` gets `unknown_kind`, never `cluster`** — dressing a bug as a real measurement is the same mistake one level down.
- 🔴 **No fake zero.** The class key is created on demand, so nothing publishes `cluster: 0` while no cluster row can exist — that zero is indistinguishable from a reader wired to nothing. `caveats.kind_scope` states the scope instead and points at `blocked_on_me` as where clawgate *is* reported (a different population, never double-counted).
- The bucket total sums whatever class keys exist rather than `claude + shell` by name, which would have silently omitted a new class from the total.

## Verification

- **14 new tests + 5 updated ledger pins: red at `origin/main` (1797d42), green at HEAD.** The load-bearing one fails at base with the defect itself — `assert 2 == 1`, *"the cluster dispatch leaked into `shell`"* — not with an AttributeError.
- **Every cluster test constructs its row**, because no code path in this build produces one. Waiting for a real cluster row would be vacuous: green today, green with the classifier deleted, and green on the day writer 3 lands wrong.
- **Mutation sweep: 7/7 killed by the specific named test**, each mutant confirmed to import first (the syntax-error-scored-as-a-kill trap): drop `kind` from the row · pre-seed `cluster: 0` · sum the total by name · fold `unknown_kind` into `cluster` · revert the top-level `shell` predicate · collapse the cluster class to `shell` · drop `kind` from the lean view.
- **Both CI tiers green:** pytests `RESULT: PASS (exit=0)`, `TOTAL collected=9872 failed=0`; nodetests `RESULT: PASS (exit=0)`, `tests=1116 fail=0`.
- Control pair for the honest zero: **cluster counts non-zero on the constructed fixtures, 0 live** — reported together, never the zero alone.

## Not in this PR — deliberately

Writer 3 is **not** built. Design and the measurements behind that: `claudedocs/design-ledger-writer3-and-kind-2026-08-14.md`. In short, `/api/agents` live is 2 long-lived devpods, both `status: error` — clawgate has no "agent run" entity; the ephemeral thing is the **task in `in_progress`**, which `clawgate_tasks.py` already models and which reported 2 wedged dispatches correctly today. Writer 3 is gated on that population being routinely non-zero **and** `task.agent` non-null (ZacxDev/homelab-infra#316).

### DIFF
--- a/claude/skills/session-manager/SKILL.md
+++ b/claude/skills/session-manager/SKILL.md
@@ -19,8 +19,10 @@ level. Roll-ups: `summary.waiting`, `summary.status[bucket]`, `summary.kind`,
 `blocked_on_me`.
 
 🔴 **`summary.status[bucket]` is one key per CLASS plus `total`** — `{claude, shell, total}`
-on every scan today. A class key is added **on demand**, never pre-seeded, so the absence of
-`cluster` is not a zero. Same rule at the top level (`summary.claude`, `summary.shell`).
+on every scan today. `claude` and `shell` are **always present** (pre-seeded, so a zero there
+is a real zero); every class **beyond** those two — `cluster`, `unknown_kind` — appears only
+when such a row exists, so the *absence* of a `cluster` key is not a measured zero. Same at
+the top level (`summary.claude`/`summary.shell` always; others on demand).
 
 🔴 **`kind` (row field) and `CLASS` (table column) are DIFFERENT axes — do not conflate.**
 `kind` is `tmux` | `cluster`: **what the entity is**. `CLASS` is `claude` | `shell` |
--- a/claude/skills/session-manager/SKILL.md
+++ b/claude/skills/session-manager/SKILL.md
@@ -15,8 +15,22 @@ python3 $DEVRC/scripts/session-manager tail scratch7:3 --plain     # scrollback,
 ```
 
 🔴 **Rows are at `report["hosts"][<"workbench"|"laptop">]["windows"]`** — not at the top
-level. Roll-ups: `summary.waiting`, `summary.status[bucket]` (`{claude, shell, total}`),
+level. Roll-ups: `summary.waiting`, `summary.status[bucket]`, `summary.kind`,
 `blocked_on_me`.
+
+🔴 **`summary.status[bucket]` is one key per CLASS plus `total`** — `{claude, shell, total}`
+on every scan today. A class key is added **on demand**, never pre-seeded, so the absence of
+`cluster` is not a zero. Same rule at the top level (`summary.claude`, `summary.shell`).
+
+🔴 **`kind` (row field) and `CLASS` (table column) are DIFFERENT axes — do not conflate.**
+`kind` is `tmux` | `cluster`: **what the entity is**. `CLASS` is `claude` | `shell` |
+`cluster`: **how it is counted**. Every row today is `kind: tmux`; `cluster` is enumerated
+for a clawgate dispatch with no pane and is **not produced yet**, which
+`caveats.kind_scope` states in the payload (`kinds_produced` is **measured from the rows**,
+not asserted). So an absence of cluster rows is **not** a measured absence of cluster work —
+clawgate lives in `blocked_on_me`, a different population that is never double-counted with
+these rows. `kind` is never null; `runtime` frequently is, and means something else
+(**which agent software**, from the ledger).
 
 | flag | effect |
 |---|---|
--- a/claudedocs/design-ledger-writer3-and-kind-2026-08-14.md
+++ b/claudedocs/design-ledger-writer3-and-kind-2026-08-14.md
@@ -1,0 +1,232 @@
+# Design pass: ledger writer 3 (clawgate) + the spec §4 `kind` field
+
+**Date:** 2026-08-14 · **Status:** design, no code written · **Decides:** spec §4, and
+whether writer 3 is built at all · **Spec:** `claudedocs/spec-agent-activity-ledger.md`
+
+Everything below marked *measured* was read live today. Everything else is a claim to check.
+
+---
+
+## 0. What changed since the handoff
+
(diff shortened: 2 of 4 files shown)


ZacxDev:
## Correction to the PR description — the "7/7 mutants killed" claim is superseded

The body above says *"Mutation sweep: 7/7 killed by the specific named test."* That was true of the seven mutants I wrote, and it is **not** the claim a reader takes from it. My sweep never covered `render_caveats` at all, and three adversarial audit rounds since have found **seven survivors** across code that sweep did not reach. Leaving the sentence to be read as "the change is mutation-verified" would be the thing this PR exists to stop, so it is corrected here rather than quietly edited.

### What the audits found that the green gate did not

Every one of these passed a full, green suite.

| round | finding | class |
|---|---|---|
| 1 | The `kind_scope` caveat guard was **spelled, not structural** — it asserted `"tmux" in line and "cluster" in line`, and both words also occur in the sentence's static prose, so neither computed slot was ever read. Two mutants rendered **false claims** through a green suite. | my verification |
| 1 | The table column headed `KIND` rendered `claude`/`shell` — a different vocabulary from the new `kind` field, one word for two axes on the two surfaces a reader compares. Renamed `CLASS`. | code |
| 1 | `render_table` summed `claude`/`shell` **by name** while the JSON had been made class-generic: `7 windows claude=3 shell=2` with two rows unmentioned. The JSON was right and the screen was wrong. | code |
| 2 | The round-1 fix **created** the round-2 bug: making `kinds_produced` measured meant a zero-row scan measures `[]`, and `[] or ["tmux"]` is `["tmux"]` — so an unreachable-hosts scan printed `rows in this scan are kind=tmux` over `summary: 0 windows`. A literal masquerading as a measurement, one function downstream of its own fix. | code |
| 2 | `filter_report` (the `detail` path) re-ran `summarize` but inherited the full scan's caveats. `gather` fixed, its sibling call site not — the same one-rule-two-sites failure, third occurrence in this PR. | code |
| 3 | The round-2 NOTE guard was **spelled again** — it banned the literal `"every row is a tmux pane"`, and a mutant saying *"every row **here is** a tmux pane and the cluster kind never appears"* survived. A second mutant slipped a **false parenthetical** into the zero-row sentence while keeping all three asserted phrases. | my verification |
| 3 | The round-2 purity fix was **exactly one level deep** — `{k: dict(v)}` left `waiting_signal["excluded"]` and three lists still shared with the module constant, and the round-2 test asserted a top-level *rebind*, which a shallow copy survives. | code |

### What replaced the walkable guards

- **Whole-string pins** for all four `_fmt_kind_scope` branches and for the `kind_scope` note. A partial assertion on prose is satisfied by prose that says something else. The trade is explicit and accepted: a cosmetic reword now fails the test.
- **A structural guard** on the note — it must name **no member of `KINDS`**. Any ban-list of phrasings is walkable; "names no kind at all" is not.
- **Two seed-pinned subprocess tests**, each with a positive control asserting the pinned seed still orders the raw set the *other* way — so a dropped `sorted()` dies every run rather than on the runs whose seed disagrees.

### Current state

31 mutants across the rounds, all killed by their specific named test, going well beyond deletion (operand swap, branch inversion, comment-out, stale rebind, synonym reword). Full gate at `c130417`: `TOTAL collected=9895 passed=9894 skipped=1 failed=0`. A round-4 delta audit is running; **this is not yet a clean round**, and the stop condition is a clean one.

### Still deferred — writer-3 prerequisites, not fixed here

`--claude-only` filters on `r["claude"]` and runs **before** `measured_caveats`, so once cluster rows exist a filtered-out row would be dropped *and* have its exclusion attributed to the build rather than to the filter. Same defect class as the one this PR closes, one flag over.

### One process note

A round-3 auditor's scratch `git commit` landed on this branch: `cp -a` of a **worktree** copies the `.git` *pointer file*, so every scratch copy shared the real git dir, index and refs. It was reverted and never pushed; branch and origin verified back at the intended commit, working tree byte-identical, base clone untouched. Recorded because "a worktree isolates a working directory only" now has a second shape: an archive copy gives zero version-control isolation.

PR #485, opened 2026-08-15
feat(handoff): a subsystem with no FILE FOOTPRINT was invisible to every window

ZacxDev:
The store is **path-addressed end to end**: all four windows produce file paths,
`nominate()` clusters paths and needs two, and the `NO ENTRY` prompt is gated on
`report.nominations` being non-empty. So a session whose work product is *not files*
resolves nothing, nominates nothing, and — until now — was offered nothing.

## The report this exists for

A session hit `looked-at-nothing` (0 paths under cwd, 20 outside, all in a throwaway
worktree), followed the `ROUTE OUT` to `--commit`, applied the PR-vs-commit discriminator
correctly (direct pushes to trunk, no PR), and called the zero genuine. **Every step right.**
Then it closed with:

> the work happened in the production database, which no path window can see.

Correct, honest, and the knowledge had nowhere to go.

## The bias is the point, not the single miss

The subsystems that live **outside** the repo — the prod database, the cluster, a DNS record, a
SaaS config — are exactly the ones whose knowledge is tribal and least recoverable from code. A
store that can only learn about things with a file footprint will never hold any of them, and
the emptiness then reads as nobody wanting it — the circular argument
`claudedocs/decision-subsystem-store-rejected-2026-08-11.md` already had to retract once.

## A pointer, not a build

`--template <slug>` already takes an arbitrary slug and needs no paths at all (verified).
Nothing pointed at it. On a dead end with **no nominations**, the tool now prints:

```
NO PATH FOOTPRINT? — every window above reads FILE PATHS. If this session's work was on a
subsystem that has none
  (a production database, a cluster, a DNS record, an external service), the store can still hold it:
    python3 …/subsystem_touch.py --template <slug> --scope homelab-talos
  🔴 Such an entry can never be RESOLVED FROM A PATH — nothing will ever match it automatically. …
```

**Silent when a nomination printed** — the `NO ENTRY` block asks the same question with better
information, and a second prompt beside it is the boilerplate that gets skipped. That's the
negative control; without it an unconditional print would satisfy every other test.

**It states the trade it makes.** An entry with no paths can never be resolved *from* one. It's
listed in the scope index and found by `--search`, which is enough to read at `/resume` and is
the whole of what it buys — a reader expecting auto-matching would otherwise conclude the store
is broken.

## Tests

6, and the seam test is **structural, not spelled** — it asserts the report's **scope** reaches
the pasteable command, because a header-string assertion is exactly what let a sibling block in
this file stay green while both its call sites were corrupted (#481). One test also pins that
the destination exists: `--template` really does emit a complete entry for an arbitrary slug,
since a pointer is only worth printing if it lands somewhere.

| mutant | result |
|---|---|
| delete BOTH call sites (inert) | RED |
| wire only ONE dead end | RED |
| print unconditionally, ignoring nominations | RED — the boilerplate failure |
| drop the scope from the command | RED (2 tests) — the spelled-vs-structural one |
| delete the never-resolved caveat | RED |
| gut the skill paragraph | RED |

## Gate

`pytests` → `TOTAL collected=9927 passed=9926 skipped=1 failed=0` · `RESULT: PASS`.
`nodetests` → `tests=1116 pass=1116 fail=0` · `RESULT: PASS`. Read from `nix log` content.


ZacxDev:
## Audit: two 🔴, both in what this PR **claimed** rather than what it did. Fixed in `f6cc12c`.

### 🔴 The suppression gate was dead code — and I called it the load-bearing decision

`if len(nominations): return []` can never evaluate true through either call site.
`looked-at-nothing` returns before `nominations` is ever set, and the other site sits inside
`if not report.writes_proposed:` where `writes_proposed = bool(known) or bool(nominations)`
makes it empty **by construction**.

Measured with an instrumented raise: **0 trips across 600 tests**, against **12** on a positive
control that added a third call site. So the test I labelled "the negative control" passed a
hand-made list straight to the helper — constructing a state `render_text` cannot produce.
Breakable, not reachable. I wrote about that exact distinction twice in this PR series and then
shipped it.

### 🔴 And the suppression was backwards anyway — it lost the *common* case

A nomination is derived from **paths**, so it names the directory the session happened to touch,
never the database it actually worked on. Measured over **287 real commits** across two repos,
**65.5% of dead ends nominate something** — so the branch I was suppressing is where the
no-footprint case *usually* hides. "The bias is the point, not the single miss" was my own
framing, and the implementation closed it for about a third of the cases where it applies.

Fixed by deleting the dead gate and making the two blocks **coexist** — `NO ENTRY` now carries
the same one-line offer. They're mutually exclusive by construction, so it has to be stated in
both places or the common case gets no route at all.

### 🔴 The caveat was false, and I'd pinned it in three places

*"Such an entry can never be resolved from a path."* Disproved end to end — and I reproduced it
myself before rewriting: an entry created via `--template` with **zero** paths, then two paths
under `infra/prod-postgres/` → `status=resolved`, matched via filename. Resolution is a property
of the **paths**, not of the entry.

Worse, the false claim discouraged the one thing that helps: picking a slug a future path would
carry. Now — nothing matches it automatically **today**; it gains normal path resolution once
some path is named for the slug, so **prefer such a slug**.

### The two surviving mutants, now killed

| mutant | why it mattered |
|---|---|
| `SELF_PATH` → a bare relative string | the pasteable command silently stops working from any other cwd; nothing pinned that the printed path resolves |
| render **before** the other routes | ordering is load-bearing — this block offers a competing explanation for an empty window and must not precede `ROUTE OUT` |

Plus a new one for the coexistence line itself. All three RED, control green both sides.

### Verified sound by the audit, left alone

`--template` works under `env -i` with no store, no repo, no paths, and its output `--validate`s
clean. Both claimed discovery routes (`--list` index row, `--search` hit) work in a throwaway
store. `SELF_PATH` is safe **here** because this module is not nix-deployed, so it cannot print a
read-only store path. The block touches no host state, so unlike its `SKILL HOMES` sibling it
renders identically in both tiers.

### Gate

`pytests` → `TOTAL collected=9929 passed=9928 skipped=1 failed=0` · `RESULT: PASS`.
`nodetests` → `tests=1116 pass=1116 fail=0` · `RESULT: PASS`. Read from `nix log` content.

**One honest caveat on the 65.5%**: it's a proxy — per-commit file lists standing in for the real
`--session`/`--pr`/`--commit` windows, against a synthetic empty scope. The direction is
unambiguous and the mechanism is proved statically; the exact figure in live use will differ.

ZacxDev:
## Delta re-audit: one 🔴, and it's the pattern rather than the bug

The **fourth** copy of *"an entry with no paths can never be RESOLVED from one"* was the
docstring of `test_it_states_the_limit_it_buys` — the **corrective test** — contradicting its own
body two lines down, which asserts the string is absent and says `VERIFIED FALSE`. My previous
commit reported the claim "pinned in three places". It was four, and the one it missed is the one
a maintainer reads first.

That's the **fifth consecutive round** in this series where a false claim survived inside the
sentence doing the correcting. Rewritten to state what was measured, and to name the practical
cost: the false version discouraged choosing a slug a future path would carry — the one choice
that makes such an entry findable later.

## Two guards that couldn't see what they were written for

- **`SELF_PATH` was pinned on the constant**, not the rendered line. Substituting a bare relative
  path at **either call site individually** left all 602 tests green — the block reads correctly
  and the command fails from any other cwd, which is verbatim the hazard the test's own docstring
  names. Both sites now assert the rendered command carries the absolute path.
- **Nothing pinned that the `NO ENTRY` offer prints once.** This round already made that mistake
  by hand — the line first landed inside the nomination loop, caught only because the indentation
  happened to be *invalid*. One line further down it would have been **valid**, printed once per
  nomination, and passed 602/602: measured, four copies for a four-nomination report.

| mutant | before | after |
|---|---|---|
| relative path, `NO ENTRY` site only | **GREEN** | RED |
| relative path, standalone site only | **GREEN** | RED |
| offer moved inside the nomination loop | **GREEN** | RED |

Each at a distinct assertion; control green both sides.

## Two figures corrected, both mine

- **"0 trips across 600 tests" was off by one.** Re-derived: 25 of 25 calls *from `render_text`*
  had zero nominations, and the single trip came from the now-deleted test calling the helper
  directly — which is the actual point. The sentence says that now instead of quoting a zero that
  isn't one.
- `test_a_RESOLVED_run_stays_silent` claimed more than it checked: a `resolved` report that
  *also* nominates **does** print the one-line offer. Renamed to name the standalone block.

## Cleared by the audit, measured not read

The coexistence line sits after the loop and after the `template:` line, prints **exactly once**
for 2/3/4 nominations, renders on all three nomination-bearing statuses, and across **11
constructed reports** the two blocks never coexist. The reworded caveat is **true** — independently
re-derived in a throwaway store, resolving via both the directory and filename tiers — and the
new advice doesn't conflict with the `aliases:`/`test_<slug>` rules. The signature change has no
stale callers. And the `SELF_PATH` test is **red-capable in the nix sandbox**, not just on a dev
host: a negative-control build failed at that exact test. Not the two-tier trap.

Reordering the standalone block before `NO ENTRY` survives, but it is a genuinely **equivalent**
mutant — normalized output byte-identical across 9 reports. Not a coverage gap.

## Gate — on the merged tree

`main` moved 5 commits with no overlap; merged in. `pytests` → `TOTAL collected=10006
passed=10005 skipped=1 failed=0` · `RESULT: PASS`. `nodetests` → `tests=1116 pass=1116 fail=0` ·
`RESULT: PASS`. Read from `nix log` content.

PR #493, opened 2026-08-15
fix(workbench): /tmp never expires because atime keeps it immortal — mtime-only, scoped ageing

ZacxDev:
Staged for root (`sudo bash nix/system/apply-tmp-churn-retention.sh`), same shape as `apply-perf-tuning-2026-07-30.sh`.

## Root cause

`systemd-tmpfiles` ages on the **most recent** of atime/mtime/ctime. Measured on real entries:

```
/tmp/nix-shell-3304470-3498967483   m=2026-08-02  a=2026-08-14  c=2026-08-02
/tmp/nix-shell-2760455-2726689738   m=2026-07-30  a=2026-08-14  c=2026-07-30
```

mtime/ctime weeks old, **atime yesterday** — so the stock `q /tmp 1777 root root 10d` rule sees "1 day old" and never expires them.

🔴 **And what refreshes atime is scanning `/tmp` with `du` or `find`** — exactly what investigating a full disk requires. Every investigation resets the clock on everything it looks at. That is how `/tmp` reached 362 GB with a cleaner that runs daily and succeeds.

## What this replaces

Both fixes the incident handoff proposed are **no-ops**, measured:

| proposed fix | reality |
|---|---|
| relax `eviction-hard` to 5% | **already 5%** — a k3s built-in default, in no config file, surviving every rebuild |
| add tmpfiles ageing on `/tmp` | **already exists** and runs daily; `--clean --dry-run` reclaims 0.55 GiB |

`ZacxDev/homelab-infra@23b5fb22` retracts both in the handoff doc.

## The fix

mtime-only age (`m:7d`), verified supported on this host's systemd 258 by control pair — `z:7d` → `Invalid age-by 'z'`, `m:7d` → parses and matches.

🔴 **Scoped by glob, NOT applied to `/tmp` as a whole.** A blanket mtime-only 7d rule would have deleted **two live git worktrees**:

```
/tmp/wt-apps-ui-3497  (14d)   gitdir: ~/workspace/civit/civitai/.git/worktrees/…
/tmp/wt-waitunit      (10d)   gitdir: ~/workspace/civit/civitai/.git/worktrees/…
```

possibly carrying uncommitted work. So: machine-generated prefixes only.

Coverage (dirs / with mtime >7d) — every glob matches something, none is a dead rule:

| glob | dirs | stale |
|---|---|---|
| `nix-shell-*` | 1003 | 988 |
| `nix-develop-*` | 1196 | 392 |
| `nix-[0-9]*` | 778 | 612 |
| `go-build*` | 114 | 98 |
| `chromedp-runner*` | 158 | 158 |
| `homelab-talos-prs-*` | 2554 | 2537 |

**Dry-run of the exact rule set: 4,255,261 entries removed, and ZERO matches** against `/tmp/wt-*`, `/tmp/claude-1000`, or the named project dirs (`civitai-integration`, `peakmaps`, `chunkcache`, `pnpmvar`, `audit187`). Estimated reclaim **~47 GB — a lower bound**, since the dry-run ran unprivileged and cannot see root-owned paths.

**Deliberately not covered:** `/tmp/claude-1000` (52 GB — the user's own session state, already covered by the stock 10d rule) and `/tmp/wt-*` (real worktrees; never age these automatically).

## Validated before staging

- Edit applied to a **copy of the live `configuration.nix`** → 777 → 788 lines, `nix-instantiate --parse` **OK**
- **Negative control:** anchor absent → exits 1 with a named error, file untouched
- **Idempotence:** marker present → second run skips
- `shellcheck` clean

## ⚠ Not run

The rebuild itself, so **no live reclaim has been measured** — the ~47 GB is a dry-run estimate. The rules are additive and delete nothing at switch time; cleanup happens on the next `systemd-tmpfiles-clean.timer` firing (or `systemctl start systemd-tmpfiles-clean.service`).

### DIFF
--- a/nix/system/apply-tmp-churn-retention.sh
+++ b/nix/system/apply-tmp-churn-retention.sh
@@ -61,6 +61,26 @@
 #   nix-[0-9]*             778 /  612      homelab-talos-prs-*   2554 / 2537
 # Every glob matches real entries — none is a dead rule.
 #
+# 🔴 A HYPHEN IS NOT A WILDCARD: `nix-shell-*` CANNOT MATCH `nix-shell.<mktemp>`.
+# nix-shell writes its TMPDIR in two spellings and this rule set only ever saw
+# one of them. Re-measured live on the workbench 2026-08-22, top-level /tmp:
+#
+#   nix-shell.*   3340   <- UNCOVERED until this rule
+#   nix-shell-*   1017      covered since 2026-08-15
+#
+# The dot form outnumbers the hyphen form 3.3:1, and it is the form `gate.sh`
+# produces on every agent run (its log lands in /tmp/nix-shell.<x>/devrc-gate-*/),
+# so it is the one that grows fastest on a machine driven by agents. The glob
+# below closes it. Control for the count above: an unrestricted `find /tmp
+# -maxdepth 1` returns 119,066 — the zeros here are real zeros, not a walk that
+# never ran.
+#
+# 🔴 Still UNCOVERED after this change, and deliberately not taken here — each is
+# larger than everything this file reaps put together, and none has been checked
+# against live work the way the globs above were:
+#   cgparent-*  17064    fx-excerpt-*  14012    cbf-*  5268    tmp.*  4227
+# `tmp.*` in particular is bare mktemp's default and would need its own audit.
+#
 # Dry-run of the exact rule set below: 4,255,261 entries would be removed, and
 # ZERO matches against /tmp/wt-*, /tmp/claude-1000 or the named project dirs
 # (civitai-integration, peakmaps, chunkcache, pnpmvar, audit187). Estimated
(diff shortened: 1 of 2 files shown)


ZacxDev:
## 🔴 The globs miss `nix-shell.*` — the dot form, and it is the biggest nix pattern on the box

Measured on the workbench today (2026-08-22), live counts under `/tmp`:

| glob | live entries | covered by this PR? |
|---|---|---|
| `nix-shell-*` | 1,017 | ✅ |
| **`nix-shell.*`** | **3,178** | 🔴 **no** |
| `nix-develop-*` | 4,340 | ✅ |
| `nix-[0-9]*` | 958 | ✅ |
| `go-build*` | 126 | ✅ |
| `chromedp-runner*` | 158 | ✅ |
| `homelab-talos-prs-*` | 176 | ✅ |

`nix-shell-*` requires a literal hyphen, so it cannot match `nix-shell.JqWNhS`. Nix produces **both** forms — `nix-shell-<pid>-<rand>` and `nix-shell.<mktemp>` — and on this host the dot form outnumbers the hyphen form **3:1**. The PR's own evidence block quotes two hyphen-form paths, which is presumably how the other spelling went unnoticed.

This is the pattern actively proliferating: **1,363 of the dot-form dirs were created in the last 24 hours** by agent gate runs, since `scripts/gate.sh` writes its log under the nix-shell `TMPDIR`. One-line fix:

```
"e /tmp/nix-shell.* - - - m:7d"
```

I reaped 1,305 of them by hand today (>7d, excluding anything a live process referenced — `/proc` scan, positive-controlled). It freed **1.4 GiB**, so the space case for this glob is modest; the case is **inode and directory-entry churn**, and that `/tmp` is now unreadable by eye.

## Everything else in `/tmp` that is machine-generated churn and uncovered

While chasing a full disk I inventoried all 117,165 entries. The globs here cover **6,775**. These are not covered, all confirmed machine-generated with structured contents, all with mtimes in July–August:

| glob | entries | what's inside a sample |
|---|---|---|
| `cgparent-*` | 17,064 | `cgroup/`, `sysfs/` |
| `fx-excerpt-*` | 14,012 | (empty) |
| `cbf-*` | 5,268 | `build.log` |
| `tmp.*` | 4,227 | generic `mktemp -d` |
| `refresh-cli-snapshot-*` | 3,370 | `bundle.txt`, `fake-civitai/`, `origin.git/` |
| `bap-*` | 2,634 | `cache/`, `scripts/`, `ws/` |
| `cb-step-*` | 2,495 | `bin/`, `recipe/`, `ws/` |
| `ab-redir-*` | 2,140 | `block.manifest.json` |

That is ~54,000 uncovered against 6,775 covered — this PR reaps about **11%** of the machine-generated entries by count.

I'd add `nix-shell.*` without hesitation. The rest I'd flag rather than write for you: I did not trace which suite owns `cgparent-*` / `fx-excerpt-*` / `cbf-*`, and `tmp.*` is a **generic** `mktemp -d` prefix that anything can produce — it deserves the same "is this machine-generated or could it hold real work?" analysis this PR already did so carefully for `/tmp/wt-*`.

## Corroborating the root cause, and one correction to the space estimate

The atime diagnosis is right and matches what I hit: I walked `/tmp` with `du` and `find` repeatedly while investigating, which is exactly the "every investigation resets the clock" trap named here.

🔴 **But the ~47 GB reclaim estimate is not what dominates.** Today `/tmp` was **766.7 GiB**, of which:

```
/tmp/mutate       587.9 GiB   (77%)
/tmp/claude-1000   52.7 GiB
/tmp/wt-*           ~5   GiB
nix-shell.* + -*    ~1.4 GiB
```

`/tmp/mutate/work/clusters-app/proc` was **582 GiB** — a recursive copy of a rootfs that included `/proc`, so three `pagemap` pseudo-files materialised at 257 G + 257 G + 70 G. Verified genuinely allocated, not sparse (536,879,872 × 512-byte blocks; `du` real usage agrees with apparent). Deleted with the operator's approval; the filesystem went **97% → 63%**, freeing 581 GiB.

None of that is an argument against this PR — a periodic rule cannot catch a one-off 582 GiB mistake, and it is a different failure (an unfiltered rootfs copy). But it does reframe the priority: **this PR is inode hygiene, not the disk-space fix.** The 99%-full disk that produced a false RED on a merge gate today was `/tmp/mutate`, not accumulated nix churn.

Worth considering as a companion, since the same class will recur: a `dirty`/`ENOSPC` tripwire, or excluding `proc`/`sys`/`dev` in whatever produced that copy. There is also a directory literally named `&&` in that tree — a shell-quoting bug in the command that made it.

ZacxDev:
Rebased onto current `main` (it was **227 commits behind**) and extended. The original commit's blob OID for `apply-tmp-churn-retention.sh` is unchanged from `08849bd3`, so nothing from the first version was discarded — force-pushed with `--force-with-lease` pinned to that sha.

**Two things added.**

**1. `nix-shell-*` has never matched `nix-shell.<mktemp>`.** A hyphen is a literal. nix-shell writes both spellings and only one was covered. Re-measured live on the workbench, top-level `/tmp`:

```
nix-shell.*   3340   <- uncovered by every rule shipped so far
nix-shell-*   1017      covered since 2026-08-15
```

The dot form outnumbers the hyphen form 3.3:1 and is the one `gate.sh` produces on every agent run (its log lands in `/tmp/nix-shell.<x>/devrc-gate-*/`), so it is the form that grows fastest on a machine driven by agents. Control for those counts: an unrestricted `find /tmp -maxdepth 1` returns **119,066** — the zeros elsewhere in the header are real zeros, not a walk that never ran.

**2. The rule could not have reached an already-applied host.** Fixing the glob alone would have shipped it to nobody. The script gated its entire edit on `grep -qF "$MARKER"` — the presence of a *comment* — so a host that had already run it printed `already present — skipping edit` and exited **0** over a config missing every rule added since. The skip is now per-rule: one `RULES` ledger is the source of truth, only absent rules are inserted, the comment header is not duplicated, and the script **re-reads the file** afterwards to prove each rule landed.

Applied state, measured rather than assumed:

| host | `grep -cF 'tmp churn retention' /etc/nixos/configuration.nix` |
|---|---|
| workbench | **0** — unapplied (positive control: the `systemd.tmpfiles.rules` anchor greps 1, so the file and the grep both work) |
| laptop | **UNMEASURED** — `Permission denied`, not zero |

That laptop row is exactly why the idempotency fix is in this PR and not deferred.

**Coverage.** `scripts/tests/test_tmp_churn_retention.py` extracts the inserter heredoc verbatim from the shipped script and runs it against fixture configs, so there is no second copy of the rules to drift out of sync. 11 tests, watched to fail:

```
M0  inert comment reword                          11 passed   <- control
M1  restore the marker-only skip                   1 failed    <- the regression
M2  delete the nix-shell.* rule from the ledger    4 failed
M3  broaden a glob to /tmp/*                       4 failed
restored, byte-identical to pristine              11 passed
```

M1 is the narrowest expression that can be wrong — it re-introduces the original defect and nothing else — and it kills exactly `test_an_already_applied_host_still_receives_a_newly_added_rule`.

Two of the tests are seam guards rather than component checks: the rules the script *writes* and the rules it *verifies* must be the same set (fails when either grows **or** shrinks), and no shipped glob may match `/tmp/wt-*` or `/tmp/claude-1000` — which this file's header promises are out of scope, a promise previously kept only by prose.

**Still uncovered, deliberately not taken here.** Each is larger than everything this PR reaps put together, and none has been checked against live work the way these globs were:

```
cgparent-*  17064    fx-excerpt-*  14012    cbf-*  5268    tmp.*  4227
```

`tmp.*` is bare `mktemp`'s default and needs its own audit before anything reaps it.

🔴 **Still staged, not applied.** This is a `nix/system/apply-*.sh`; it needs `sudo bash nix/system/apply-tmp-churn-retention.sh`, which Claude cannot run.

PR #496, opened 2026-08-15
fix(handoff-doc): --push committed BEFORE checking it could push, and stranded the commit

ZacxDev:
**Measured in this repo, an hour ago.** `handoff_doc.py --confirm --push` committed the doc to `main` in the SHARED base clone, then the push was rejected non-fast-forward — two other sessions had pushed during the session. **The commit stayed.** `status=push-failed` reported the push failing and nothing said a commit was now sitting on `main`, unpushed.

That is exactly the state CLAUDE.md has a rule against: `ship.sh` converges with `merge --ff-only`, so a diverged host is **skipped and left "as found"** — it stops receiving every future change while still looking healthy. It has bitten this repo twice (2026-08-06, 2026-08-09), and this tool could produce it from a clean checkout by doing nothing wrong except being two commits behind.

The tool already had the right property everywhere else — `no-advance`, `no-change` and the pre-confirm path all write **nothing** on failure. Push was the one branch that traded it for a commit the caller had to know how to undo.

## Two halves, because one is not enough

**Refuse before the write.** With `--push`, resolve the branch and `git fetch` it first; if the remote is ahead, exit **6** (`EXIT_BEHIND`) having written nothing, and print the `merge --ff-only` that fixes it plus the preserve/verify/reset path for a real divergence.

🔴 **A fetch that fails is not "0 behind."** Network down, no such remote, auth expired — it refuses rather than guessing pushable, because a confident wrong answer strands the commit the same way.

**Tell the caller when it happens anyway.** The remote can move between the fetch and the push; that race cannot be designed away. The failure now says the commit **exists**, names the `ship.sh` consequence, and hands over preserve → verify → `reset --keep` **in that order**.

## Tests

8 new. Mutation battery, 6 semantic mutants, all killed:

| mutant | result |
|---|---|
| pre-check deleted (the shipped bug) | RED |
| pre-check moved AFTER the write | RED — the ordering that caused it |
| failed fetch treated as 0 behind | RED |
| always refuse | RED — negative control, `--push` stays usable |
| failure stops saying the commit exists | RED |
| failure stops naming the recovery ORDER | RED |

Two traps hit writing them, recorded in place: `tree_hash` hashes `.git` and the pre-check fetch legitimately writes `FETCH_HEAD`, so "nothing written" is asserted as doc bytes + commit list + remote ref instead; and the push-failure path had **no test** — triggered deterministically by pointing `origin` at a non-bare repo with `main` checked out (fetch succeeds, push refused), a trick found by accident when a fixture cloned the wrong path.

## Gate

`TOTAL collected=10013 passed=10012 skipped=1 failed=0` · `RESULT: PASS`, read from `nix log` content.


ZacxDev:
## Audit: four 🔴, and the first is that **the guard's own mechanism reintroduced the bug it closes**

Fixed in `9d065a5`. `git fetch` + `FETCH_HEAD` was the wrong instrument, three separate ways.

**It reintroduced the stranded commit.** The check fetched, then read `HEAD..FETCH_HEAD` in a
*second* process. `FETCH_HEAD` is shared mutable state — any other fetch in that checkout between
the two wins. Measured on a checkout genuinely 1 behind: after another session's
`git fetch origin stable`, the check returned **0**. Write → commit → push rejected → stranded
commit, which is the entire failure this guard exists to close. `drift-check.sh` fetches on a
systemd timer, so the racer is unattended.

**"A read-only fetch is safe in a shared checkout" was false** — and I wrote it as reasoning
rather than measurement. `git fetch` writes `refs/remotes/<remote>/<branch>` in the **common**
gitdir, shared by every worktree, plus objects and reflogs. Two concurrent fetches produced
`cannot lock ref` in **30 of 30** trials.

**A first push became impossible.** `git fetch <remote> <branch>` exits 128 when the branch isn't
on the remote yet — an ordinary end-of-session state, hard-refused with no way past it.

Replaced with `git ls-remote --exit-code`: writes **nothing** locally, correct on all six cases
(absent-on-remote, up-to-date, behind, ahead-only, ahead **and** behind, unreachable). Exit 2 is
an *answer*, not a failure, so it needed a runner that returns the code instead of raising.

### 🔴 The exit-code ledger restated instead of deriving

`test_every_exit_code_the_module_can_return_is_documented` iterated a hand-written literal. Its
own docstring says an undocumented status "leaves the agent improvising at the moment it is about
to push to a shared branch" — and `behind` did exactly that. It now scrapes `status=` from the
module, and immediately caught `push-failed` as *also* undocumented. Both are in `SKILL.md` now.

### Also fixed, all audit-measured

- The push-failure recovery **could not be pasted** — it printed `HEAD` and a literal `<branch>`
  while the resolved name sat in a variable. That message is the whole second half of this PR.
- A fetch failure killed the **local** doc write and never offered the exit; it now names
  `--confirm` without `--push`.
- `--branch <other>` advised merging an unrelated branch into the current checkout.
- `resolve_branch` shipped with **zero** coverage — a constant-returning mutant survived all 55
  tests. Three tests now, including a detached HEAD refusing rather than guessing.

### Three findings from my own battery — tests masking each other

- The ahead-and-behind test **never reached the ancestry comparison**: its repo had never fetched
  the remote commit, so the unknown-tip path decided, and flattening `merge-base` to `False`
  stayed green. A fixture that fetches the object *without merging* forces the real path.
- A `cat-file -e` guard was **redundant** — `merge-base --is-ancestor` on an unknown object also
  exits non-zero, already the refuse answer. Two branches reaching one outcome can't be told
  apart by any test; deleting either stayed green. Removed.
- The unreachable-remote test pinned only the *refusal*, so disabling the primary raise stayed
  green while the diagnostic degraded from `cannot read origin/main: <git's error>` to
  `returned no sha`. Refusing is the safety property; naming **why** is what makes it actionable.

13 tests (8 → 21). Every mutant now killed, including a silent revert of the mechanism itself
back to `fetch`/`FETCH_HEAD`.

### Gate

`TOTAL collected=10301 passed=10300 skipped=1 failed=0` · `RESULT: PASS`, on the merged tree.

PR #506, opened 2026-08-16
feat(hooks): prose lost 2/2 on the clawgate write-back, so gate it on the READ

ZacxDev:
## The problem, measured

Clawgate tasks **#193 and #194** were both picked up, the work shipped as PRs, and **both cards stayed `open` with ZERO comments**. Both were later re-dispatched and paid for twice. `claude/skills/clawgate/SKILL.md` §"task pickup" already marks the comment/status ritual 🔴 **NOT optional and NOT a thing to be asked for**. Prose is 0-for-2 against it. PRINCIPLES.md prefers a deterministic/structural fix to prompt-tuning; this is that fix.

## 🔴 Keyed on the READ, not the status flip

The obvious design — a PreToolUse deny on `clawgatectl task status <id> in_progress` until a comment exists — is **structurally unreachable for the exact failure it would be built to catch**: in both measured cases no status command was ever issued, because the cards never left `open`. A guard on a command nobody ran observes nothing.

The one act that provably happened in both is the **read**, so that is what arms the hook:

```
clawgatectl task get <N>          # the SKILL's own step 1
curl … /api/tasks/<N>[/comments]  # the same read, before clawgatectl existed
```

Both directions of that trigger are tested. It must **not** match `task ls`, `/api/tasks?summary=1`, a bare `/api/tasks`, or `task get <non-numeric>` — a board *survey* is not a pickup, and arming off one would put a block in front of a turn that claimed nothing.

## What fires it — three conditions, all required

1. the session read a specific clawgate task id, **and**
2. **real work followed** — an `Edit`/`Write`/`NotebookEdit`, or a Bash `git commit` / `git push` / `gh pr create|merge` / `gh release create` **in command position, outside quotes and comments**, and that tool call did **not** carry an `agent_id`, **and**
3. a **LIVE** re-read of the board shows **no comment authored by `claude-code` newer than the MOST RECENT WORK EVENT**.

**(2) is the false-positive killer.** The skill's own step 2 is *"EVALUATE and report to Zach. Do NOT flip status yet"* — a read-and-evaluate-only turn owes the board nothing, and blocking it would be worse than no hook. Asserted here not just by the verdict but by **the board never being read at all**, so a reordering that puts the live read first is visible even when it happens to end up silent.

**(3) is a measurement, not an inference.** It self-suppresses the moment the ritual is followed — including by a subagent, a devpod agent, or a separate process — which a hook tracking only its own observations could not do. A card already in `ready_for_review`/`complete` is left alone.

🔴 **An unreachable board is a NOTICE, never a block.** Unreachable, no client, unparseable JSON → a `systemMessage` naming the task and saying the board could not be reached. `systemMessage`, **not** `additionalContext` — see the audit section below for why that distinction is the whole of finding 1. Going silent there would report the same observable as "the ritual was followed" (RULES.md → *an empty result cannot distinguish two mechanisms*).

## Escalation ladder — per session, per task id

| fire | output | forces a continuation? |
|---|---|---|
| 1 | `{"decision":"block","reason":…}` | **yes** |
| 2 | `{"decision":"block","reason":…}` | **yes** |
| 3 | `{"systemMessage":…}` | no — the turn ends |
| 4+ | silent | no |

**True cost of a measured missing write-back: exactly two forced continuations. True cost of a "could not measure" notice: zero**, and it runs on its own counter, so an outage cannot spend the block budget a genuinely missing write-back needs later in the same session.

Re-derived from the **installed CLI bundle** (claude-code 2.1.220, `bin/.claude-wrapped` — *not* the 20 KB `bin/claude` wrapper, a grep against which returns a meaningless zero), read with Python and with **both controls**: positive `hookEventName` → 68 matches, negative `zzzNOTPRESENTzzz_devrc_control` → 0. **Re-derived independently during the audit round** with a fresh control pair — positive `CLAUDE_CODE_STOP_HOOK_BLOCK_CAP` → 5, negative `zzQuuxNotPresentNonce9137` → 0 — which is what found the `additionalContext` error below.

* the hook output schema is a **top-level** object, not a `hookSpecificOutput` arm:
  `v.object({continue, suppressOutput, stopReason, decision: v.enum(["approve","block"]).optional(), reason: …, systemMessage, terminalSequence})`
* the command-hook consumer reads exactly that at **exit 0**:
  `M = L && HB(L) && L.decision === "block"; $ = H.status === 2 || !!M; D = M ? L.reason || H.stderr || "" : …` → `{blocked: $, output: D}`
* consecutive Stop blocks are capped **by the CLI** at `let Kt = wue(process.env.CLAUDE_CODE_STOP_HOOK_BLOCK_CAP, 8)`. **Ours is 2** — a guard the harness has to override has already lost the operator. 🔴 That cap counts `additionalContext` rungs too, which is exactly why rung 3 is a `systemMessage`.

There is deliberately **no `stop_hook_active` gate**, the one place this diverges from the CLI's own advice: that advice exists to bound a block loop, the ladder bounds it at 2 instead, and the *second* block is what catches a turn that acknowledged the first and stopped anyway.

## Hot path

PostToolUse fires after every tool call and `agent-ledger-hook.py` already costs ~21 ms there. A session that has **never** read a clawgate task and is not reading one now costs **exactly one `os.path.exists` and spawns nothing** — pinned by a test that *counts* the filesystem calls and subprocess spawns, plus its own positive control, because an earlier hook in this repo shipped with its throttle consulted *after* the subprocess spawn while its comment claimed the opposite.

Registered on PostToolUse with **no matcher**: half of what it watches for is an `Edit`/`Write`/`NotebookEdit`, i.e. exactly the work a `Bash` matcher would never show it — the commonest shape of the failure (read a card, edit files for an hour).

## Wiring — all four

1. `nix/home.nix` — `home.file.".claude/hooks/clawgate-writeback-guard.py"`.
2. `scripts/claude-hooks/register-nudge-hook.py` — append-only registration on PostToolUse (no matcher) + Stop; docstring updated with why.
3. Both registrar suites now assert every pre-existing Stop owner coexists **as a set equality**, so they fail when the set grows as well as when it shrinks.
4. Both new files `git add`ed — a new file the flake does not track is silently omitted while the switch still succeeds.

Plus 3 lines in the clawgate skill's §"task pickup".

## Evidence

**Red at base / green at HEAD.** The hook is new, so at base (`f522077`) the suite cannot even import it — collection ERROR, all 128 red by construction. The two registrar suites, run at base with only the *test* files swapped in, are red on exactly the new assertions: `test_register_nudge_hook.py` → **7 named FAILUREs**, `test_registrar_activation.py` → **1 failed / 16 passed**. At HEAD: 128/128, 17/17, all-pass.

**Two independently-constructed mutation sweeps**, both under `PYTHONDONTWRITEBYTECODE=1` with every `__pycache__` dropped between mutants (CPython validates cached bytecode on whole-second mtime + size, so a same-length edit is otherwise scored SURVIVED without ever running — several mutants here *are* same-length), every mutation proven applied by sha256 + presence, and the file restored from a copy with the sha re-checked each round:

* **Batch A** — 30 hand-written mutants (constants perturbed, guards forced false, statements hoisted): **29 KILLED, 1 SURVIVED — and that one is the negative control**, an inert comment edit that must survive. Positive control (`AGENT_AUTHOR` → `"api"`) killed 6 tests.
* **Batch B** — mechanically different: every single-line `if` in the file discovered by parsing the source and forced **TRUE**, one at a time. **48/48 as expected**, same negative control in the batch.

Batch B's *first* run found **three real gaps** batch A could not see, all fixed here: the work gate was only ever exercised through a hand-seeded state dir (never through `post_tool_use`), a corrupt state file was untested, and one branch was an unkillable duplicate of `emit`'s own silent case — removed rather than left looking like coverage.

**Verified LIVE, read-only, against the real board with the real `clawgatectl`** (not a stub): the full hook in a real process read task 193, recorded the work, and stayed **silent** because the card is `ready_for_review`; the same live payload with status forced `open` resolves `written` against a 2020 read timestamp and `missing` against a 2099 one. That live probe is also what caught a wrong-subsystem diagnosis — a `clawgatectl` that exists but exits non-zero was being reported as "no clawgatectl".

**Gate.** `scripts/gate.sh --tier pytest` → **exit 0**, `RESULT: PASS (exit=0)`, `TOTAL collected=10588 passed=10587 skipped=1 failed=0`. The authoritative `nix build .#checks.x86_64-linux.pytests` agrees target-for-target, read out of `nix log` rather than from silence. New target: `collected=208 passed=208 skipped=0 floor=198`, the floor taken from the gate's own printed count through the gate's own `_suggested_floor`.

## 🔴 What this structurally CANNOT see

* **Work done anywhere this hook is not running** — a dispatched devpod agent, the clawgate web UI, a human, opencode, or a host that has not had `home-manager switch` run since this lands. (The *board* half still works: a comment written by any of those silences the hook, because condition 3 is a live read.)
* **A read that took some other route** — the board UI, a summary listing, someone pasting the body into the prompt. No read, no arming.
* **Whether the comment is any good.** It checks that a `claude-code` comment exists since the read, never what it says — it cannot tell a real completion report from `"."`.
* **The `in_progress` flip.** Not required; this gates the *write-back*, which is the thing that was measured missing. (The **pre-start comment** used to be on this list and it did not belong there — it did not merely go unseen, it *disabled* the guard. Fixed in the audit round below.)
* **Which task a given edit belongs to.** Nothing in a PostToolUse payload says, so the work flag is session-wide. Two shapes still fire without the session owing anything — work in an unrelated repo after reading a card, and a survey of two cards where only one is written back. Both are **tested as behaviour**, and both have a one-command deterministic escape (`--dismiss`).
* **A clock skew larger than 120 s** between the board and the host would misjudge recency; that is the allowance, and it is bracketed by fixtures at 30 s and 3600 s.

Not merged, not deployed — no `home-manager switch` / `ship.sh` run. Registration on each host needs `register-nudge-hook.py`, which the switch's activation entry already runs.

---

# 🔍 Audit round (2026-08-16) — what the adversarial review found, and what changed

Two 🔴 findings, both re-derived from the installed bundle before anything was
touched, both confirmed.

### 🔴 1. `additionalContext` on Stop does not let the turn end

`Ycd` — the Stop-hook driver — pushes an `additionalContexts` entry into the **same
array** as a `blockingError`:

```js
if (F.blockingError)      { …; E.push(G); … }
if (F.additionalContexts) { …; E.push(j); … }
…
if (E.length > 0) return { blockingErrors: E, preventContinuation: !1 };
```

and its caller re-queries the model for **any** non-empty `blockingErrors`, with
`stopHookActive:!0`, incrementing the counter `CLAUDE_CODE_STOP_HOOK_BLOCK_CAP`
bounds. So the ladder's "relent" rung did not relent: **three** forced continuations,
not two, and the `# 🔴 NEVER blocks` comment on the cannot-measure path was **false**
— an unreachable board cost three.

`systemMessage` is the channel that genuinely does not. In the same bundle:

```js
if (j.systemMessage) yield { message: Va({ type:"hook_system_message", … }) };
```

It is yielded on the MESSAGE channel, never into `E`; it renders to the operator as
`"<hookName> says: <content>"`; and its attachment-to-messages entry is
`hook_system_message: () => []`, i.e. it is **not** fed back to the model. Exactly the
shape for a rung whose job is to stop nagging the model and tell the human. Rung 3
now emits it.

The old proof that the rung "did not block" was `stop_decision` returning the string
`"context"` — a **spelled guard** in the sense RULES.md names: the name said
non-blocking while the JSON it stood for was blocking, and a test on an internal
string structurally could not see that. Every such assertion now runs the verdict
through the real `emit` and asks `forces_a_continuation(<the emitted JSON>)`, a
predicate transcribed from the bundle. A mutant that restores the old
`additionalContext` emission is **killed** by 9 tests.

### 🔴 2. The work flag was armed by subagents, and the escape hatch was prose

The bundle builds every hook payload through
`Kf(e, t, r) → { session_id: t ?? kt(), …, agent_id: r?.agentId }`, and PostToolUse
calls it as `{...Kf(i, void 0, o), …}` — second argument `void 0`. A tool call made
**inside a dispatched subagent therefore arrives carrying the parent's
`session_id`**, with only `agent_id` to separate them; the bundle's own schema says
so ("Use this field (not agent_type) to distinguish subagent calls from main-thread
calls"). The parent reads a card, a subagent does the editing, and the parent's Stop
blocked on work it never did. Those events are now refused before the state dir is
even computed.

The block text's escape — *"if you did NOT do work on this task, say so in one line
and stop"* — was measured **not to work**: saying something changes no state, so the
next Stop re-blocked with identical text. Priced from the consuming code, a false
positive plausibly ends in a junk comment, a wrongly-`ready_for_review` card and a
**push notification to Zach** (`notifyTaskDone` fires on entering that status) — and
the junk comment then permanently silences the guard for that card. The escape is now
a mechanism:

```
python3 ~/.claude/hooks/clawgate-writeback-guard.py --dismiss <id> --session <sid>
```

named in the block text **with the session id already interpolated**, because a model
cannot see its own hook payload and a command it must fill a session id into is a
command it cannot run. It clears `read-<id>` and both counters, and it is tested by
extracting the command from the block reason the hook actually emits and running it
as a real subprocess.

**All four probes are now tests**, including the two that still fire — labelled as
residual false positives rather than hidden.

### 🟡 Also fixed

| finding | what changed |
|---|---|
| a pre-start comment **disarmed** the guard | The comparison anchor is the **last work event**, not the read. The skill's own ritual posts "Starting" right after the read and before the work, so anchored on the read the guard was satisfied at pickup and its forward yield on every compliant session was **zero**. `Starting → work → Stop` now correctly owes a comment; `Starting → work → Done → Stop` is satisfied. Multi-turn cost is bounded (≤2 blocks + 1 notice per task per session) and **tested**, and the 120 s skew allowance absorbs a push right after a comment. |
| the two hang bounds were pinned by nothing | Both survived being multiplied by 75 and 60 because every ladder test injected `budget=3.0` and a fake clock. Now exercised at their **defaults**, including the `remaining <= 0` boundary itself, plus curl's `max-time`, `fail`, `silent` and the `timeout + 2` wait — which no test had ever compared. The real worst case is stated: **15.0 s**, not the 8.0 the file advertised. |
| `is_work` matched its literal text anywhere | Quoted runs and `#` comments are blanked, and the verb must be in **command position**. `grep -rn 'git commit' scripts/` and `rg 'gh pr create' claude/skills/` are routine acts in this repo — those exact strings live in RULES.md and CLAUDE.md — and they used to arm a hook that can block a turn. Nine over-match cases tested; `gh pr merge` and `gh release create` added as fail-open under-matches. |
| cannot-measure spent the block budget | Its own counter. A board down for the first Stops can no longer stop a genuinely missing write-back being blocked later. |
| `parse_ts`'s nanosecond trim | Deleted. Dead on the pinned interpreter (measured on 3.12.13: `.000000000` parses fine), and the docstring's "accepts at most six" was true only up to 3.10. |
| `_sanitize`, `_state_dir`'s type bail, `_scrub` | All had **zero** coverage and all survived mutation. Now pinned, including truncation off-by-one. |
| third-party stderr spliced into operator-visible text | Collapsed to single-spaced printables and truncated. |
| three unkillable duplicate guards | Removed: the `[:MAX_TASKS]` slice (`record_read` already caps), the work-gate re-check past the fast-path return, and the second `os.path.exists` in `record_read`. |
| `_via_curl`'s "the laptop has no clawgatectl" | Stale — verified present on both hosts. Claim corrected, fallback kept (it is also the path a *failing* `clawgatectl` takes). |

### Evidence for this round

**Red at base / green at HEAD.** 46 test functions added. Run against the pre-fix hook
(with a compat shim for the two changed fixture signatures, so only behavioural signal
remains): **28 are RED at base** — regression coverage — and **18 are green at base**,
which is what a control or an invariant guard *should* be and is labelled as such
rather than counted as regression coverage. Five pre-existing tests also go red at
base, on the behaviour this round deliberately changed.

**Mutation battery, re-run over the whole file after the fixes, not just the delta.**
66 mutants enumerated from semantic failure modes — operand swap, comparison flip,
branch inversion, wrong bind, off-by-one, comment-out, constant perturbation — run
under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` deleted between mutants and every
mutation proved applied by **sha change plus replacement-text presence**.

* **64 KILLED, 2 SURVIVED — and both survivors are the declared negative controls**
  (a dead initialiser, and a comment-only edit).
* **Positive control killed** (`return "missing"` → `return "written"`, 31 tests red).
* The battery also re-covers everything the audit said was already good — the
  token-into-argv shape, the fail-open backstop, the fast-path return, the author and
  retraction checks, the skew/TTL/MAX_TASKS constants — all killed.
* Its **first** run found two real gaps the fix round had left: bypassing
  `_strip_literals` survived (the command-position anchor alone covered every case I
  had written, but not a separator *inside* a quoted string), and flipping
  `remaining <= 0` to `< 0` survived because no fixture sat on the boundary. Both are
  now tested.

**Hot path re-measured at parity**: 14.30 / 14.56 / 14.33 / 14.62 ms per call against
14.57 / 14.43 / 14.14 / 14.03 for the pre-audit file measured in the same session
(bare interpreter 7.7–8.2 ms). Holding that took one deliberate change — the three
work-detection patterns moved off the import path, since none is reachable from the
fast path and compiling them eagerly cost ~0.8 ms on **every tool call of every
session**. A first cut with them compiled at import measured 14.7 against 13.9, and
the file's stated measurement is updated to what was actually measured.

**Both authoritative gates, read out of `nix log`** (a cache hit returns exit 0 with
no output, so silence is not a verdict):
`nix build .#checks.x86_64-linux.pytests` → `RESULT: PASS (exit=0)`,
`TOTAL collected=10588 passed=10587 skipped=1 failed=0`;
`.#checks.x86_64-linux.nodetests` → `RESULT: PASS (exit=0)`, `1116/1116`.
The dev-host tier (`scripts/gate.sh --tier pytest --set all`) agrees: `exit=0`.

**Measured against the merged tree? No.** `main` moved during this round and the
branch was deliberately not rebased; every number above is measured on the **branch**.

---

https://claude.ai/code/session_014rwzAkPvjXK9E2aM9zwBMa

### DIFF
--- a/claude/skills/clawgate/SKILL.md
+++ b/claude/skills/clawgate/SKILL.md
@@ -99,6 +99,10 @@ test gate, build/push, pin bump, the CSS-cwd trap that fakes ~25 e2e failures, c
 months because reading a task was one command while writing to it was a curl preamble; that
 asymmetry is gone (`task status` / `task comment` since 2026-08-14). Run it unprompted.
 
+🔴 **A hook ENFORCES this now** (`~/.claude/hooks/clawgate-writeback-guard.py`): armed by the
+step-1 **read**, it **blocks Stop** when work followed (edit/commit/push/PR) and a **live** re-read
+shows no `claude-code` comment since. Read-and-evaluate-only never fires; commenting silences it.
+
 🔴 **A COMMENT is the only write that notifies a watcher.** A status flip pushes **only** on
 *entering* `ready_for_review` (`notifyTaskDone`), so going `in_progress` notifies **nobody**. That is
 *why* the pre-start comment exists — it is Zach's only chance to object **before** the work happens,
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -1057,6 +1057,33 @@ in
   home.file.".claude/hooks/agent_ledger.py" = {
     source = ../scripts/lib/agent_ledger.py;
   };
+  # 🔴 THE CLAWGATE WRITE-BACK GUARD — the deterministic replacement for a 🔴 prose
+  # rule that lost 2/2. Tasks #193 and #194 were both picked up, the work shipped as
+  # PRs, and both cards stayed `open` with ZERO comments; both were re-dispatched and
+  # paid for twice. `claude/skills/clawgate/SKILL.md` §"task pickup" already said the
+  # ritual was NOT optional. PRINCIPLES.md prefers a structural fix to prompt-tuning.
+  #
+  # 🔴 KEYED ON THE READ (`clawgatectl task get <N>` / a curl to `/api/tasks/<N>`),
+  # NOT on the `in_progress` flip. In both measured failures no status command was
+  # ever issued — the cards never left `open` — so a PreToolUse deny on the flip
+  # would have been structurally unreachable for the exact failure it was built for.
+  #
+  # Fires only when all three hold: the id was read, REAL work followed it (an
+  # Edit/Write/NotebookEdit, or a `git commit`/`git push`/`gh pr create`), and a LIVE
+  # re-read of the board shows no `claude-code` comment since that read. The middle
+  # condition is what keeps read-and-evaluate-only turns — the SKILL's own step 2 —
+  # untouched; the last is a measurement, so it self-suppresses the moment the ritual
+  # is followed, including by a subagent or a devpod agent. Escalates block, block,
+  # notice, silence per task per session, and NEVER blocks when the board could not
+  # be reached (it says so instead). Every error path exits 0.
+  #
+  # 🔴 A NEW file, so it must be `git add`ed or the flake silently omits it and the
+  # switch still succeeds with the hook absent — this repo's standing trap.
+  # Registered on PostToolUse (NO matcher — half of what it watches for is an Edit)
+  # and Stop, per-host, by register-nudge-hook.py.
+  home.file.".claude/hooks/clawgate-writeback-guard.py" = {
+    source = ../scripts/claude-hooks/clawgate-writeback-guard.py;
+  };
 
   # 🔴 THE REGISTRAR ITSELF — the hook that makes the hooks above DO anything.
   # settings.json is per-host and unmanaged (permissions/allowlists), so a hook
--- a/scripts/claude-hooks/clawgate-writeback-guard.py
+++ b/scripts/claude-hooks/clawgate-writeback-guard.py
@@ -98,9 +98,12 @@ subagent, and its edits and commits are exactly the evidence condition 2 needs.
 agent-ledger-hook.py already costs ~21 ms there. The fast path is: resolve the state
 dir from `session_id` (string work, no IO), ONE `os.path.exists`, and the trigger
 regex. A session that has never read a clawgate task and is not reading one now does
-nothing else — no directory creation, no state read, no subprocess, ever. That
-ordering is pinned by a test, because an earlier hook in this repo shipped with its
-throttle consulted AFTER the subprocess spawn while its comment claimed otherwise.
+nothing else — no directory creation, no state read, no subprocess, and it does not
+even IMPORT `subprocess` or `shutil` (see the deferred-import note below). That
+ordering is pinned by tests that COUNT the calls and read the module list out of
+`-X importtime`, because an earlier hook in this repo shipped with its throttle
+consulted AFTER the subprocess spawn while its comment claimed otherwise. Measured:
+13.9 ms/call against 8.6 ms for a bare interpreter start.
 
 🔴 FAIL-OPEN, ALWAYS. Every internal exception exits 0 with an empty stdout and
 blocks nothing. main() has exactly ONE exit and it is always 0. A hook that wedges a
@@ -123,10 +126,46 @@ import datetime
 import json
 import os
 import re
-import shutil
-import subprocess
 import sys
 import time
+
+# 🔴 DEFERRED IMPORTS, AND THE REASON IS THE HOT PATH. This hook runs after EVERY
+# tool call, so its import cost is paid thousands of times a day. Measured with
+# `python -X importtime` on this host: subprocess 3.4 ms, shutil 3.7 ms — 7.1 ms of
+# the ~11 ms this hook added over a bare interpreter start, and NEITHER is reachable
+# from the PostToolUse path. Both are Stop-only (the live read, and the state prune).
+# Measured end to end, 30 fast-path runs per sample, four samples: 19.0 ms/call before
+# this against 13.7/13.9/13.8/14.3 after, with a bare interpreter at 8.0-8.8 ms — i.e.
+# the hook's own overhead falls from ~11 ms to ~5.4 ms. `re` (2.4 ms) and `json`
+# (0.9 ms) stay: the trigger patterns compile at import and the payload is JSON on
+# stdin, so both are on the path that cannot avoid them.
+# `shutil` is loaded inside the REMOVAL, not at the top of prune(), so a Stop with
+# nothing stale to sweep never pays it either — pinned by the importtime test.
+subprocess = None
+shutil = None
+
+
+def _sp():
+    """`subprocess`, imported on first use. Bound to the MODULE attribute so a test
+    can still monkeypatch `guard.subprocess` once the Stop path has touched it.
+
+    No `if subprocess is None` memo guard: `import` IS the memo (every call after the
+    first is a `sys.modules` lookup), and a guard whose only effect is skipping that
+    lookup is a branch no mutation can kill — which reads as a coverage gap when it is
+    really just a duplicate of what the import statement already does.
+    """
+    global subprocess
+    import subprocess as _m
+    subprocess = _m
+    return subprocess
+
+
+def _sh():
+    """`shutil`, imported on first use — see `_sp`."""
+    global shutil
+    import shutil as _m
+    shutil = _m
+    return shutil
 
 # --------------------------------------------------------------------------- #
 # Tunables
@@ -349,7 +388,7 @@ def prune(ttl=STATE_TTL_SECS, now=None):
         try:
             if os.path.getmtime(path) >= cutoff:
                 continue
-            shutil.rmtree(path, ignore_errors=True)
+            _sh().rmtree(path, ignore_errors=True)
             removed.append(name)
         except Exception:                 # noqa: BLE001
             pass
@@ -417,7 +456,7 @@ def _env_file(path=CLAWGATE_ENV):
 
 
 def _via_clawgatectl(task_id, timeout):
-    proc = subprocess.run(["clawgatectl", "task", "get", str(int(task_id))],
+    proc = _sp().run(["clawgatectl", "task", "get", str(int(task_id))],
                           capture_output=True, text=True, timeout=timeout)
     if proc.returncode != 0:
         raise LiveReadError("clawgatectl rc=%d %s"
@@ -450,7 +489,7 @@ def _via_curl(task_id, timeout, env_path=CLAWGATE_ENV, why=None):
         'url = "%s/api/tasks/%d"\n' % (url.rstrip("/"), int(task_id)),
         'header = "Authorization: Bearer %s"\n' % token,
     ])
-    proc = subprocess.run(["curl", "-K", "-"], input=cfg,
+    proc = _sp().run(["curl", "-K", "-"], input=cfg,
                           capture_output=True, text=True, timeout=timeout + 2)
     if proc.returncode != 0:
         raise LiveReadError("curl rc=%d" % proc.returncode)
@@ -459,6 +498,11 @@ def _via_curl(task_id, timeout, env_path=CLAWGATE_ENV, why=None):
 
 def live_task(task_id, timeout=PER_TASK_TIMEOUT_SECS, env_path=CLAWGATE_ENV):
     """Live re-read of one task. Raises LiveReadError when it cannot be measured."""
+    # Bound BEFORE the try: the `except subprocess.TimeoutExpired` clauses below name
+    # the module attribute, and an except clause is evaluated even when the exception
+    # came from the import itself — at which point a still-None `subprocess` would
+    # raise AttributeError out of the handler and defeat the fail-open contract.
+    _sp()
     why = None
     try:
         return _via_clawgatectl(task_id, timeout)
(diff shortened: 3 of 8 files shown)


ZacxDev:
### Update — second commit `6826ffd`, and a correction to the numbers above

**Hot-path cost, measured after the first commit rather than asserted.** 19.0 ms per fast-path call against 8.0 ms for a bare interpreter start — 11 ms of overhead on a hook that runs after *every* tool call. `python -X importtime` attributes **3.4 ms to `subprocess` and 3.7 ms to `shutil`**, and neither is reachable from PostToolUse (one is the live board read, the other the state prune). Both deferred. Now **13.7–14.3 ms/call** over four samples against 8.0–8.8 ms bare, i.e. the hook's own overhead falls from ~11 ms to **~5.4 ms**.

Pinned by a test that reads the module list out of `-X importtime` rather than asserting where the `import` line sits, with a positive control on the Stop path. Neither lazy accessor carries an `if <mod> is None` memo guard — `import` *is* the memo, and a branch whose only effect is skipping a `sys.modules` lookup is one no mutation can kill.

### 🔴 The sweep's own instrument had broken, and it took a control to notice

Batch B's target discovery **silently collapsed from 48 targets to 1** the moment the new lazy-import docstrings broke its `"""`-toggle. It still printed `0 SURVIVED` — a green over a target set wired to almost nothing, which is exactly the reassuring-zero trap. Discovery now goes through `ast` (which never sees a docstring as code) and asserts a floor on the count, so a collapse is loud rather than green.

Both sweeps re-run after the change, both clean at **130 tests**: batch A **30/30** (29 KILLED + the inert negative control SURVIVED), batch B **48/48** (47 discovered `if`s, all killed, + the negative control).

### Re-verified live after the refactor

An audit/review fix resets the verification gate, so the live probes were re-run, both read-only:

* the real hook against the real board still stays **silent** on `ready_for_review` task 193;
* the **BLOCK path exercised end to end against a genuinely `open` card** (#200 on the live board), returning `{"decision":"block"}` with the two fix commands named in the reason. Nothing was written to clawgate.

### Gate, re-run

`scripts/gate.sh --tier pytest` → **exit 0**. `nix build .#checks.x86_64-linux.pytests` agrees, read out of `nix log`: `collected=130 passed=130 skipped=0 floor=124`, `TOTAL collected=10510 passed=10509 skipped=1 failed=0`.

*(Supersedes the "128 tests / floor=122 / 19 ms" figures in the description above.)*

PR #521, opened 2026-08-17
security(public-repo): delete the dspy-eval capture — captured third-party text — and gate the class

ZacxDev:
## The exposure

`scripts/initiatives/dspy-eval/scan-days30.json` — tracked on `origin/main`, 216,879 bytes — was a frozen `initiative-scan --days 30 --json` capture. Measured (counts only; no content is reproduced anywhere in this PR):

| | |
|---|---|
| string leaves | 1,452 |
| `recent_messages[].text` message bodies | 333 (**262** under one client repo key) |
| sentence-length leaves (>60 chars) | 606 |
| longest leaf | 647 chars |
| email addresses | 1 |
| `by_repo` entries | 4 (repo #0 = 88 initiatives / 262 messages) |

Four sibling fixtures carried the same class: `eval_set.json` (55 `recent_messages` entries), and `results-run{1,2,3}.json` (109 model-written `recap` values — second-order, since a recap of a client's messages still describes the client's work). `variance.json` is numbers only and was clean.

**This repo is PUBLIC.** The capture was added by #148 and then **survived #350** — the PR whose entire subject was scrubbing this repo's public IPs and client hostnames. It survived because #350's two gates look for IPs and for hostnames, and a paragraph of captured prose is neither. *A scrub of a category is not a gate on the category.*

I swept every tracked JSON/JSONL in the repo (23 files). These 5 were the only ones with a finding.

## Step 1 — delete, not scrub

Evidence, in the order it was gathered:

1. The eval is **report-only** and explicitly "NOT wired into the production sync"; DSPy is not a repo dependency.
2. **The code was already dead.** `dataset.py` calls `recap.recap_context()`. That function no longer exists — `recap.py` was refactored into an `identity_context`/`status_context` pair. Verified by import: `recap` loads fine, `hasattr(recap, "recap_context")` is `False`. The eval set could not be rebuilt even if you wanted to.
3. **Nothing outside the directory imports it.** Only two claudedocs mention it, both as pointers to where results live.
4. **The results survive in full** in `claudedocs/dspy-recap-eval-2026-07-23.md`: verdict, the 3-run results table with all four per-dimension scores, the noise floor, five side-by-side examples, the optimizer config, wall-times, recommendation and residual risk.
5. Its last commit was the #350 security scrub itself.

Scrubbing to synthetic content would preserve a harness that cannot run, whose README points at files that no longer mean anything, and whose existence invites someone to drop a fresh capture in. Deleting the directory is the cheaper and more complete forward fix, and it is fully reversible from git history.

## Step 2 — the gate

New: `scripts/testlib/captured_text_scan.py` (the scan) + `scripts/tests/test_no_captured_text.py` (the gate), following the shape of the two siblings.

**Structural, not semantic — deliberately.** A finding is a conjunction of three facts, none about what the text *means*:

```
a JSON/JSONL file  ×  a message-ish key  ×  a value that is free text
```

There is no reliable way to read a sentence and decide whether the author typed it or captured it, and a scan that tried would fire on this repo's ~200 markdown files of legitimate prose. `RULES.md` is explicit that a permanently-red gate is worse than none. Authored prose lives in `.md`; captured prose arrives as a **serialized dump**, under the key its producing tool gave it. That is the structural difference the gate keys on. **I did not reach for a prose heuristic; this is the deterministic alternative.**

`MESSAGE_KEYS` is an enumeration with each entry's reason in the source, not a pattern, and not env-overridable (pinned by a test). ~~`_EXCLUDED_KEYS` is likewise explicit — notably **`description` is excluded BY NAME rather than allowlisted**~~ — **CORRECTED IN ROUND 2.** The reasoning (a scanner that allowlists its own canonical examples ends up scanning clean over a real leak) is right; excluding the key outright was too blunt, because this repo ships a ClickUp integration where `description` IS the task body (`claude/skills/clickup/query.mjs:238`, 56 occurrences) and GitHub issue bodies use the same field. `description` is now a CONDITIONAL member of `MESSAGE_KEYS`, excluded only inside a document that also carries `manifest_version`. `_EXCLUDED_KEYS` was also *read by no code path* — documentation shaped like a guard; it now blocks dict inheritance and has a test that dies when it stops.

**🔴 The allowlist cannot be keyed on the value.** Both siblings key exemptions on `(relpath, value)` — the IP or hostname is spelled in the allowlist. This gate cannot: the value *is* the disclosure, so writing it down to exempt it would commit the very thing being exempted. Exemptions are keyed on `(relpath, keypath)` **plus a pinned COUNT**, which is what stops a pin from becoming a rubber stamp that pre-approves every message added at that key later. Accounting runs in both directions (unpinned hit fails; stale pin fails; a count that moved EITHER WAY fails — round 2 added the oversized direction, which was unpinned). ~~It is currently empty.~~ **It is not, as of round 2:** 14 entries over 4 files, 12 of them synthetic fixtures whose own headers say their content is invented, and 2 on a real capture that is PINNED, NOT BLESSED — see below.

~~**The report itself cannot leak.**~~ **OVERSTATED — CORRECTED IN ROUND 2.** The VALUE never leaves the module; that half stands. The KEY half did not: `_SCHEMA_KEY` admitted `.` and `-`, so five constructed data keys echoed **verbatim** into a formatted finding — a client hostname, a bare repo name, a username, an email localpart and a media directory name, three of them classes `CLAUDE.md` forbids outright. The only guard used an ABSOLUTE PATH, the one shape that fails the leading-`[A-Za-z_]` anchor, so widening the character class mid-string SURVIVED. Round 2 tightens the regex, adds a structural second layer for the residual it cannot reach, and adds a fixture per shape. It is **still not fully closed**: a one-entry data map, a data map of scalars, and the RELPATH itself are all unredacted, and the module docstring now says so instead of claiming otherwise.

### Both controls, reported as a pair

**Negative control — can it go red?** Built from the **real exposure**, not a textbook fixture: with the fixtures restored, the gate FAILED, 2 tests, **18 findings** across all 5 files:

```
scripts/initiatives/dspy-eval/scan-days30.json: by_repo.*[].recent_messages[].text  (114 free-text values, longest 200 chars)
scripts/initiatives/dspy-eval/scan-days30.json: by_repo.*[].summary                 (112 free-text values, longest 199 chars)
scripts/initiatives/dspy-eval/results-run1.json: results[].items[].recap            (38 free-text values, longest 137 chars)
… 15 more
```

Note the client repo path rendered as `*`, and that not one character of captured content appears. Plus synthetic negative controls for a realistic capture shape, a bare string list directly under a message key (the `eval_set.json` shape — invisible without list-key inheritance), a JSONL transcript, and a grown fixture under a live pin.

**Positive control — can it stay green on legitimate content?** 23 tracked JSON/JSONL candidates, **0 findings** after deletion; 20/20 tests pass. Explicit non-findings pinned as tests: authored extension `description`, long non-prose values (a 300-char token, a nix store path, a sha256, a URL), a short message, and an unknown key.

**Red → green: 2 failed (18 findings) → 20 passed.**

### Mutation results

`PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged between mutants, each mutant proven applied by **sha change AND byte-equality** and non-inert by **AST diff**:

| mutant | want | got |
|---|---|---|
| `is_free_text` always False | DIE | DIED (7 failed) |
| `_unpinned` swallows everything | DIE | DIED (2 failed) |
| list does not inherit its key | DIE | DIED (4 failed) |
| data-shaped map key echoed verbatim | DIE | DIED (2 failed) |
| threshold raised past all prose | DIE | DIED (8 failed) |
| `.jsonl` no longer scanned | DIE | DIED (1 failed) |
| **inert comment added (control)** | **SURVIVE** | **SURVIVED** (`ast_differs=False`) |

6/6 lethal died; the inert control survived, so the deaths are not a suite that was red anyway.

## Step 3 — discoverability

`CLAUDE.md`'s 🔴 PUBLIC-repo rule named five classes (media path, directory name, filename, route log, third-party hostname). Captured message/prompt text was a **sixth, unnamed** class. Added in one sentence at its widest scope — "anyone's message bodies, prompts, transcripts or chat content, or a model's summaries of them, however it arrives" — with the gate named as the owner of the key list and its blind spots, so the always-on file does not restate them.

## Gate verdict (merged tree)

`HEAD` was exactly `origin/main` (`2c684ed`) at gate time and still is, so the tree gated **is** the merged tree.

`nix build .#checks.x86_64-linux.pytests` → exit 0 with no output (cache/success), so read from `nix log`:

```
RESULT: PASS (exit=0)
TOTAL collected=11378  passed=11377  skipped=1  failed=0  (floor: 9945 = sum of 23 per-target floors)
PASS  scripts/tests  (collected=4911 passed=4911 skipped=0 floor=3970)
```

**Floor: not bumped, and that is the rule, not an oversight.** The gate PASSED and printed no replacement number. `scripts/tests` sits at 4911 against floor 3970 with a drift band of `max(60, 3970/4) = 992`, i.e. a ceiling of 4962 — inside the band. ⚠️ **It is only 51 short of the ceiling**, so the next PR adding ~50 tests to `scripts/tests` will be told to raise it; take the number the gate prints then, never arithmetic on these two.

## 🔴 What remains exposed — history

**Untouched, and deliberately so.** Every byte described above is still in this repo's public git history and has been since 2026-07-23 (#148). A forward fix does not unpublish it: the repo is public, so it may already be cloned, forked, mirrored or indexed, and a history rewrite would not reach any of those copies while breaking every existing clone and open PR. That is a judgement about blast radius and about the value of the content — **your call, not mine.** If you want it, the options are `git filter-repo`/BFG plus a force-push and a GitHub support request to expire cached views, and the honest framing is *reducing* exposure, not removing it. This PR stops the next one.

## 🔴 What this gate structurally cannot catch

Named here and, at more length, in the scan module's own docstring — treat a clean run as "no match for these shapes", never "no captured text here":

- **Captured text in any format that is not JSON/JSONL** — the big one. A transcript pasted into a `.md`, a `.csv`/`.txt`/`.yaml` export, a shell heredoc, or a Python list literal are all invisible.
- **A message-ish key not in the enumeration** — a producer calling its field `utterance`, `line` or `msg_body` walks straight through. So does anything under `description`, which is excluded by name.
- **Free text below 60 characters.** A 40-character message is still a message; the threshold buys the low false-positive rate that keeps this gate green. That is a trade, not a proof.
- **Free text with no whitespace**, double-encoded JSON (one opaque leaf), and non-UTF-8 files. Unparseable tracked JSON is *not* silently skipped — it is reported and pinned at zero.
- **Git history**, as above. Like both siblings, this guards HEAD only.

It stops the accidental dump. It is not an exfiltration control.

## Second item assessed: `claudedocs/the-algorithm-applied-2026-06-17.md`

**Assessed and deliberately left — flagging rather than acting, since it is a judgement call.** It quotes ~9 operator request phrasings as request *types* (lines 19-28). They differ in kind from the fixture:

- They are **aggregated, normalized stems with frequency counts** ("`analyze the X setup, then…` — 32"), not verbatim message bodies. The count is the point.
- They are **the operator's own words about his own workflow** — no third party, no client, no PII.
- Every one is **already published in this repo as a feature name**: `/audit-pr`, `/analyze-service`, `/handoff`, and the standing subagent rule in `CLAUDE.md`.

I judged that below the bar that motivated this PR (262 verbatim third-party message bodies under a named client repo path). Say the word and I will scrub it in a follow-up.

For the same reason I left `claudedocs/dspy-recap-eval-2026-07-23.md` §5, which quotes 10 model-generated feature-level sentences with no repo path, client name or message body — and which is now the *only* record of the eval, so deleting it would destroy the evidence that justifies deleting the fixtures. Worth a second opinion if you read those five examples differently than I did.

---

## Audit round 2 — the gate was correct and INERT in every dimension nobody measured

An adversarial audit returned "safe to merge" with one 🔴 and several 🟡. It is
not safe, for a reason the audit surfaced: **the gate scanned clean over a second
live capture already in the tree.** Round 2 closes that plus the coverage gaps.
`claude/RULES.md`: *an audit/review fix RESETS the verification gate* — so
everything below was re-measured, not reasoned about.

### 🔴 It missed a live file

`scripts/task-spec-drafter/tests/fixtures/drafter_run_2026_07_29_commands.json`
is a frozen capture of a REAL 2026-07-29 agent run and the first cut of this gate
scanned clean over it. It is now visible as `executed[] 95` / `rejected[] 2`.

The audit's read that adding a `transcripts` key would catch it is **wrong, and
measured wrong**: that key's 11 values are 42-character UUID filenames with no
whitespace, so they can never be free text at any threshold. What makes the file
visible is `executed`/`rejected` — a captured **command log**, which is
`CLAUDE.md`'s *route log* class, not prose. The module's stated scope now
includes it.

**The file is PINNED, NOT BLESSED.** Delete-vs-scrub-vs-regenerate is the
operator's call; the assessment is in the review thread. The pin exists so the
gate stops reporting a zero it has not earned, and its COUNT cannot grow
silently.

### The four arrival paths, each measured missing, each now with a named test

| # | path | what walked through |
|---|---|---|
| 1 | the producer's own key | `answer` / `evidence` (`browser-agent-parse.py` — this repo's highest-privilege capture producer, it drives a real logged-in browser) and `subject`. Matching is now **case-insensitive**, which is what sees the mail fixture's 13 `Subject` headers. |
| 2 | the PLURAL | 13 of 18 keys had none — including `comments`, GitHub's own API spelling, and `prompts`, `summaries`, `excerpts`, `snippets`, `recaps`. |
| 3 | a DICT under a message key | only a LIST inherited the enclosing key. `{"messages": {"m_01": …}}` and `{"transcript": {"<iso-ts>": …}}` are ordinary export shapes and were one line from the case that was handled. |
| 4 | NO KEY AT ALL | `messages.json` holding a bare list (the walk started at `key=None` and never consulted the FILENAME), and a `.jsonl` of bare strings. Suffix set widened to `.jsonc` — this repo TRACKS one — and `.ndjson`. |

### 🔴 The 60-char threshold was near-total blindness for chat

MEASURED: `scripts/signal/tests/fixtures/envelopes.json` — the repo's only chat
subsystem fixture — carries **16** `message`/`body` string leaves — 8 distinct
messages, each mirrored in the `envelope` and `expect` halves — of **12–32
characters, 0 of 16 above threshold**. (The audit reported "6 of 6"; that is not
what I measure, and the direction of the finding is unchanged.) A gate on captured chat that cannot see one value in the
chat subsystem is scanning clean over its own target class.

**Decision: lower it, for the unambiguous chat keys only.** `CHAT_KEYS` uses
`MIN_CHAT_TEXT_CHARS = 25`; everything else keeps 60. The alternative — document
the blindness and leave it — keeps the allowlist empty at the price of a gate
that is provably blind to the one subsystem whose entire content is other
people's words. The cost is paid in the open, as 12 count-pinned entries over
three fixtures whose own headers say their content is invented. `subject` is
deliberately **not** a chat key: a 25-char subject line is ordinary and would add
26 more pins from the mail fixture alone.

### Vacuous guards, now driven rather than asserted about

* **the unparseable pin had no negative control** — mutating `unparseable_files`
  to `return []` survived all 20 tests of round 1. The code worked; the test did
  not exist. There was also **no allowlist**, so one legitimately-unparseable
  tracked file would have made this gate permanently red with no escape hatch —
  the exact failure this PR argues against. Both fixed.
* **12 of 18 `MESSAGE_KEYS` were never exercised.** A parametrized test over the
  enumeration does **not** close this, and that is measured, not assumed: with
  only that test in place, deleting `"recap"` still **SURVIVED**, because the
  case simply stops being generated (97 → 96 passed, zero failures). All three
  enumerations are now spelled out as an **asserted ledger** and compared for set
  equality.
* the threshold boundary was unpinned (`>=` → `>` survived; neither fixture was
  near 60 — they were 19 and 70); the oversized-pin direction was unpinned
  (`!=` → `<` survived); `_candidates` did not re-apply `_is_skipped` where
  `public_ip_scan.scan_repo` does — and the guard for that has to build a **real
  git repo**, because a `tmp_path` with no `.git` takes the filesystem tier where
  `repo_files` already filters and the fixture passes either way.
* `test_the_scan_never_returns_a_captured_VALUE` was a **type** guard wearing a
  content guard's name. Renamed; the content claim is now the parametrized
  data-key tests.

### Red / green

**38 of the new tests are RED at `origin/scrub/captured-text-gate` and GREEN at
HEAD** (77 collect against base vs 100 at HEAD — the enumerations are smaller
there, so the parametrized cases are fewer), measured by running the new test file against the base scan module (a
name-only shim so it collects and each red is attributable per-test). The rest
are controls and say so in their own docstrings — they are not counted as
regression coverage.

### Mutation

18 mutants, each proven applied by **sha change + byte-equality**, proven
non-inert by **AST diff**, run under `PYTHONDONTWRITEBYTECODE=1` with every
`__pycache__` purged between mutants. **17 KILLED with a named killer; 1 inert
control SURVIVED.** Positive control (`scan_repo` iterating an empty list) died
against 81 tests.

|---|---|---|
| `_SCHEMA_KEY` widened to admit `./-` | KILLED | all 4 new `test_a_data_shaped_map_key_is_never_echoed[…]` |
| `unparseable_files` → `return []` | KILLED | `test_negative_control_the_unparseable_detector_can_go_RED` |
| `MESSAGE_KEYS["recap"]` deleted | KILLED | `test_the_key_enumerations_are_pinned_as_a_ledger[MESSAGE_KEYS]` |
| `is_free_text` `>=` → `>` | KILLED | `test_positive_control_a_short_message_is_not_a_finding` |
| `pin[0] != count` → `< count` | KILLED | `test_negative_control_a_GROWN_fixture_under_a_live_pin_still_fails` |
| dict inheritance removed | KILLED | `test_a_DICT_under_a_message_key_inherits_it_like_a_list_does` |
| filename seed disabled | KILLED | `test_a_filename_supplies_the_key_when_nothing_else_does` |
| case folding removed | KILLED | `test_key_matching_is_case_insensitive` |
| `.jsonc`/`.ndjson` dropped | KILLED | `test_a_jsonc_comment_is_parsed_but_a_json_comment_is_not` |
| chat threshold raised to 60 | KILLED | 19 tests |
| `_candidates` skip re-application dropped | KILLED | `test_candidates_re_applies_the_skip_dir_test` |
| data-map redaction off | KILLED | `test_a_map_of_same_shaped_records_redacts_keys_the_regex_admits` |
| manifest scope → blanket exclusion | KILLED | `test_positive_control_a_manifests_description_is_not_a_finding` |
| *(inert control)* a comment reworded | **SURVIVED** | — as required |

### Gate

`scripts/gate.sh --tier pytest --set hermetic` on the **merged tree** (this branch
+ `origin/main` at `5565c33`):
`TOTAL collected=11493 passed=11492 skipped=1 failed=0`, every target PASS except
`scripts/tests`, which was over its **drift ceiling** (5026 vs 4962 — 1056 of
slack). That is the ceiling check doing its job, so the floor moved to the number
that run printed verbatim: `"scripts/tests|4976"`. No arithmetic on two sides.

**BOTH TIERS, read separately.** `nix build .#checks.x86_64-linux.pytests` on the
final tree — `nix log` says `RESULT: PASS (exit=0)`,
`TOTAL collected=11494 passed=11493 skipped=1 failed=0`, with
`PASS scripts/tests (collected=5027 passed=5027 skipped=0 floor=4976)` and
**zero new skips** (the 1 skip is the pre-existing pinned `repo-cos` live-store
check). `.#checks.x86_64-linux.nodetests` also exit 0.

⚠ Harness note for whoever reproduces this: `nix build path:<worktree>#…` is
**not** the CI invocation and gives a false red. A worktree's `.git` is a FILE,
`path:` copies it, and five pre-existing "is this file git-tracked?" tests then
take their non-sandbox branch and get `git` exit 128. Use the git flake ref
(`nix build .#checks…` from inside the tree), which carries tracked files only.
Those five failures were reproduced, diagnosed and are unrelated to this diff.

**Mutant M17, found by re-reading rather than by a finding:** the `.jsonc`
comment stripper removed trailing commas with a regex that is not string-aware,
so it silently edited any string value containing `, }` — changing the length
this gate reports and, at the boundary, whether a value is a finding at all —
while its own docstring claimed strings were preserved. One string-aware pass
now, with a fixture that carries both sequences plus a `//` and a `/* */` inside
the string.


ZacxDev:
## Section A — `drafter_run_2026_07_29_commands.json`: assessed, **not acted on**

The gate now SEES this file (`executed[] 95`, `rejected[] 2`) and it is **pinned,
not blessed**. Delete / scrub / regenerate is your call. My recommendation is
**KEEP AND PIN**, and the evidence is below because the pin is only honest if the
numbers behind it are visible.

### What consumes it

`scripts/task-spec-drafter/tests/test_allowlist_covers_prompt_shapes.py`
(section 8, "THE REGRESSION CORPUS") — three tests, and it is the only consumer
in the tree:

| test | what it does |
|---|---|
| `test_regression_corpus_is_intact` | pins 108 executed / 11 transcripts / the head set — an anti-tamper guard against emptying the corpus to make a deny rule pass |
| `test_no_command_the_real_run_executed_becomes_denied` | replays all 108 against the deny list; asserts none loses its capability, and that exactly 11 bare-`kubectl` reads MOVE to the wrapper rather than vanish |
| `test_corpus_rejections_stay_rejected_except_the_ones_we_fixed` | the 3 rejected shapes: 2 were genuine gaps and are now runnable, 1 was malformed and must stay rejected |

### The disclosure is close to zero *incremental*, and that is measured

| measure | value |
|---|---|
| distinct identifier tokens in the file (absolute paths + repo slugs) | **4** |
| …of which appear in **other tracked files** already | **4 of 4** |
| tokens that appear NOWHERE else in the repo | **0** |
| distinct absolute paths | 3, **all** under the operator's own `$HOME` — the same string `CLAUDE.md` itself spells |
| distinct repo slugs | 1 (present in 7 other tracked files) |
| email addresses / URLs / hostnames / IPs | **0 / 0 / 0 / 0** |
| third-party prose, message bodies, transcript text | **0** |

The same paths and the same slug are spelled **literally** in the consumer test's
own source (`test_allowlist_covers_prompt_shapes.py:80,87-89,637-638,784-785`)
and in `drafter.sh`'s production allowlist, because a pinned-repo allowlist entry
*must* name the repo. **Deleting the fixture removes not one distinct identifier
from this public repo.**

That makes it materially unlike the dspy-eval capture this gate was written for:
that one carried 333 third-party message bodies under a client repo key. This one
carries a shell history of the operator's own commands.

### The three options and what each costs

* **DELETE** — breaks 3 tests and removes the only evidence that the deny layer
  does not silently break a real unattended run. That is the exact failure mode
  #177 existed to fix (an over-broad deny is unanswerable in headless: the call
  is lost, the pass keeps going, and it emits confident records built on fewer
  reads). Cost: high. Disclosure removed: **zero** by the table above.
* **REGENERATE SYNTHETIC** — destroys the property the corpus exists for. Its
  value is precisely that a REAL run produced these shapes; a synthetic corpus
  tests only the shapes someone imagined, which is `RULES.md`'s canonical-example
  blindness reintroduced deliberately. The pinned counts (108 / 11 / 3, and
  `moved == 11`) are measurements of a real run and become fiction. Cost: high,
  and it weakens the guard.
* **SCRUB the paths to a placeholder** — cheap mechanically (3 distinct strings)
  but it **breaks the test's meaning**: `_runnable()` is evaluated against an
  allowlist whose `git -C` entries pin those literal paths, so replacing them
  turns a "this real command still runs" assertion into "this invented command
  does not match anything". You would be asserting the allowlist covers paths it
  does not cover. Cost: moderate, benefit: negative.

### Recommendation

**Keep it, keep the pin.** If you want the exposure reduced anyway, the cheapest
real reduction is not this file — it is that the same identifiers sit in ~20
other tracked files, so scrubbing the fixture alone would be theatre.

Whatever you decide, the gate's behaviour is now correct either way: delete the
file and `test_every_allowlist_entry_still_matches_something` fails until the two
pins are removed, so the ledger cannot rot.

ZacxDev:
## 🔴 Correction to this PR's own evidence (posted, not silently edited — a reader may already have believed the earlier version)

The assessment of `scripts/task-spec-drafter/tests/fixtures/drafter_run_2026_07_29_commands.json` stated **"4 distinct identifier tokens … 4 of 4 already appear in other tracked files."** That undercounts by roughly 3× and is not the reason the file is safe.

**Measured independently, twice:**

| | |
|---|---|
| distinct path-shaped tokens | **17** (a second count found 16 with a different regex) |
| absolute paths under the operator's own `$HOME` | 3 |
| relative source paths inside the referenced repo | 13 — directory names and source filenames |
| appearing in no other tracked file | ~10 |

`CLAUDE.md`'s PUBLIC-repo rule names *directory name* and *filename* outright, so those are in-scope tokens the original count omitted. Also correct in the assessment, and worth keeping: adding `transcripts` would **not** have caught this file — those 11 values are 42-char UUID `.jsonl` names with zero whitespace, so `is_free_text`'s whitespace requirement excludes them. `executed[]` / `rejected[]` are what make it visible.

**The actual defence, measured:** the referenced repo is `visibility=PUBLIC` / `isPrivate=false`, and the source paths quoted here already exist in its public history (5 of 7 sampled resolve at its `origin/main`). The disclosure is not incremental because those tokens are already public at their source — not because they happen to recur elsewhere in this repo.

The **KEEP AND PIN** recommendation stands on that basis. The `_REAL_DRAFTER` pin reason in source is unaffected: it says *PINNED, NOT BLESSED* and leaves the delete/scrub/regenerate call to the operator, which is still correct.

No code change. The gate behaves identically either way — this corrects the record the decision is made on.

PR #536, opened 2026-08-18
fix: a stale SOURCE repo shipped a silent no-op wearing a current version — derive clawgatectl's version from the code, and give drift-check rc 17

ZacxDev:
## The failure

Two repos, two delivery mechanisms, and only one of them is automatic.

`clawgatectl` is built by devrc from a **local working tree of a different repo** —
`nix/pkgs/tools/clawgatectl.nix` sets `srcDir = "${workspace}/homelab-talos/containers/clawgate"`.
The **version literal** lived in devrc and was stamped in with `-X main.buildVersion=`.
The **code** lives in homelab-talos and is read live off disk. **Nothing converges
homelab-talos on either host**: `ship.sh` is scoped to `$HOME/workspace/devrc`, and
`drift-check.sh` had never heard of it.

Measured 2026-08-14:

| time | what |
|---|---|
| 17:44 | homelab-infra `978c549c` (#323) adds `task status` + `task comment`, sets the Go default `buildVersion` to `"0.7.95"` |
| 17:48 | devrc `1b790b3` (#483) bumps the nix literal `0.7.87 → 0.7.95`, ships to both hosts |

The workbench's homelab-talos was current → a genuine 0.7.95. The laptop's was frozen
24 commits back at `6e9055f7`; its `client.go` **correctly said `0.7.87`**, and the nix
ldflag **overwrote that truth**. The laptop shipped a binary with neither subcommand,
labelled `0.7.95`. `clawgatectl task status <id> in_progress` printed help and **exited 0**
— a silent no-op against a CLI whose write ritual the `clawgate` skill makes mandatory.

`drift-check.sh` was fully green on that laptop the entire time: devrc git parity fine,
symlinks resolved, `settings.json` keys matched. The condition was structurally outside
its scope.

`task-api.md` argued a stale binary is harmless because every command prints a
server-vs-build skew note. That argument was **false**: `h.Version == buildVersion`
compared *equal*, so the note stayed silent. Corrected here.

---

## Change A — nix stops overwriting the source's own version

`nix/pkgs/tools/clawgatectl.nix` no longer carries a `version` literal. It reads
`cmd/clawgatectl/client.go` and extracts **exactly one** `var buildVersion = "…"` line;
`version` (hence the store path) and the stamped `buildVersion` are then the same derived
value, both describing the code actually being compiled.

* **No suffix, no fingerprint.** `client.go:277` compares `h.Version == buildVersion` by
  exact equality; any suffix fires the skew note on every command on every host forever.
  Had this been in place on 08-14 the laptop would have reported `0.7.87` and the skew
  note would have fired **correctly** against the 0.7.95 server. Letting the existing
  mechanism work *is* the design.
* **Unparseable source → `available = false`, never a fallback literal.** Zero matches
  (renamed/reformatted) and two-or-more (ambiguous) both yield `null`. The binary is
  simply not installed — loud at use time as `command not found` — rather than failing
  the switch, which `ship.sh` reports as a SKIPPED host and this repo's `CLAUDE.md`
  documents as the failure mode that silently stops all future delivery. The existing
  `pathExists` guard is extended, not replaced.
* Header comment block rewritten; the old one explained the version/source relationship
  in terms this change makes false.

**`tmux-fuzzyclaw` could NOT take Change A**, and that is a measurement, not an omission
(2026-08-18): its `cmd/version.go` declares `var Version = "dev"`, the string `2.0.0`
appears **nowhere** in that repo outside the devrc nix file, and the repo carries **no git
tags**. There is no source of truth to read and inventing one would be the same lie in a
new place. The finding is documented in `tmux-fuzzyclaw.nix` and **pinned by a test that
fails the day `cmd/version.go` grows a real version** — which is the trigger to apply
Change A there too. Change B covers it regardless.

---

## Change B — a source-repo parity deadman, `drift-check.sh` rc 17

A third kind of parity, one level past "git parity is not host parity": **for every
package devrc builds from a local working tree, is that source repo current on each host?**

* **The covered set is DERIVED, never listed.** The payload reads `nix/pkgs/**.nix` out of
  the checkout it is examining (bash `globstar`, no `find` — busybox on the laptop
  silently exits 0 on `-xtype`) and collects every `${workspace}/<name>` outside a
  comment. A third such package is covered the day it lands.
* **Pinned two-way, three independent spellings that must agree**: a reviewed literal, a
  pathlib+regex oracle, and *what the shell payload itself reports against the real
  `nix/pkgs`*. Fails when the set GROWS or SHRINKS.
* **Per host**: behind/ahead of the branch's **own upstream** (never a hardcoded `main` —
  homelab-talos works on `trunk`), plus dirty trees, which matter because these
  derivations copy the **tree**, not the commit.
* **`examined=N stale=M unmeasured=K` is printed as a triple.** `absent`, `fetch failed`,
  `detached`/`no upstream` and `could not compare` are **UNMEASURED** — reported, never
  folded into a reassuring `stale=0`, never drift. A scan finding no source packages
  prints `NOT EVALUATED`, not `examined=0 stale=0`.
* **Cross-host half is INFORMATION ONLY** and says `NOT COMPARED` without facts from both
  machines. Differing HEADs are reported with **no rc**: whether a HEAD is *wrong* is
  already answered per host against its own upstream, whereas two hosts on different
  branches of a shared dev repo is normal — a code for it would be permanently red.
* **READ-ONLY**: `fetch`, `rev-parse`, `symbolic-ref`, `rev-list`, `status`. Passes the
  suite's static passivity allowlist and a new behavioural test asserting a source repo's
  branch/HEAD/worktree/stash are byte-identical after a run.

### Why rc 17, and where it sorts

`severity()` is a **lookup table**, and the published order was never monotonic in the
digits (14 > 13 > 6 > 4 > 3 > 12). Every free code **below** 16 (5, 7, 9, 11) is
explicitly reserved to a ship.sh meaning this script does not take, so a new DRIFT code
has nowhere to go but upward. 17 is ranked **67**:

```
8  >  17  >  14  >  13  >  6  >  4  >  3  >  12  >  15  >  10  >  16
```

* **below 8** — a diverged devrc stops *every* future change to the host, of which a stale
  source repo is one instance, and rc 8's rescue can destroy work.
* **above 14** — a dangling managed symlink is **loud** at the moment of use
  (`command not found`); this is **silent**: the measured failure ran, exited 0 and did
  nothing while wearing a current-looking version.

rc 17 is **not** on `SuccessExitStatus` — unlike 13 and 16 it is a real divergence with a
real fix, so it reaches `OnFailure` and toasts, like rc 8.

### `nix/home.nix`: `TimeoutStartSec` 180 → 420

180 was sized for "two `git fetch`es plus one ssh round trip". The source-repo leg adds
`2 hosts × N repos × DRIFT_SRC_FETCH_TIMEOUT (30s)` = 120s of new worst case. At 180 the
cgroup would be killed mid-run and the deadman would report **nothing**, on a schedule,
looking merely slow. That product is now a **pinned seam test** recomputed from both files.

---

## Testing

**42 new tests** — 24 in `test_drift_check.py` (213 → 237 collected) and 18 in a new
`test_clawgatectl_version.py`. Full per-test matrix vs base **`a2707be`** is in the PR
conversation: **39 of 42 are RED at base**, and the 3 that are green are labelled
**⚠ INVARIANT GUARD** in their own docstrings rather than counted as regression coverage.

`test_clawgatectl_version.py` has two layers: structural assertions that always run, and a
behavioural layer that **actually evaluates** `clawgatectl.nix` against throwaway fixture
source trees with `pkgs` stubbed to the four attributes it touches. It **fails rather than
skips** when `nix-instantiate` is missing — the policy `test_opencode_config.py::nix_eval`
already sets here.

### Mutation sweep — 19 mutants, run under `PYTHONDONTWRITEBYTECODE=1`

18 killed on the first pass, **one SURVIVED**: mutating `LN="$REST"` (advance past the
occurrence just consumed) to `LN=""` (stop after the first) survived the entire suite
**including the two-way pin** — because every line in the real `nix/pkgs` happens to name
at most one source, so both extractions shared the blind spot. That is a mutant silently
reintroducing an under-reporting bug through a fully green suite.
`test_two_source_repos_named_on_ONE_line_are_both_found` closes it; the battery is **19/19
after**. Each mutant is the narrowest expression that can be wrong (the guard is never
deleted together with its enclosing condition), and the positive control is included.

### Live probe

Read-only with respect to both real checkouts: each was `git clone`d (hardlinked, source
untouched) into a scratch `$HOME`, its origin repointed at the **real private GitHub URL**,
and only the clone fetched. The private-repo fetch over ssh **succeeded** from this
environment, and the leg reported:

```
🔴 DRIFT — source repo homelab-talos is NOT current: 1 behind / 0 ahead of refs/remotes/origin/trunk
source repo tmux-fuzzyclaw: on branch docs/tui-rendering-footguns — no upstream to compare against, currency NOT evaluated.
source repos: examined=2 stale=1 unmeasured=1 (7 nix file(s) scanned)
SRC-RC=17
```

That is **real, live drift on the workbench right now** — the exact class this exists for.
It also validates the "no upstream is NOT drift" decision: `tmux-fuzzyclaw` is genuinely
sitting on an untracked local branch, and had that been ruled drift the deadman would have
been permanently red on the workbench from day one.

**Not verified**: whether the fetch succeeds from a **systemd `--user`** context, which has
no ssh-agent. If the GitHub key needs an agent it will report `FETCH FAILED` → UNMEASURED
→ no rc, which is honest but measures nothing there. Worth one `systemctl --user start
drift-check` after the deploy. Neither host was touched.

## Gate

See the PR conversation for the `scripts/gate.sh` exit status and the authoritative
`nix build .#checks.x86_64-linux.pytests` result. `TARGET_FLOORS` needed **no re-pin**:
`scripts/tests` collected 5228 against floor 4976 and a drift ceiling of 6220, and the
gate printed no replacement number.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -56,10 +56,18 @@ there. Only what's specific to this repo, where a working tree is also a **deplo
   help and **exited 0**, and drift-check was green throughout. Both halves are now
   closed: `clawgatectl.nix` **reads its version out of the Go source it compiles**
   (never a literal — an unparseable source means no binary, not a wrong label), and
-  drift-check reports each source repo's currency per host. The covered set is
+  drift-check reports each package's source currency per host. The covered set is
   **derived from `nix/pkgs/` at scan time and pinned two-way by a test**, so a third
   such package is covered automatically. `absent` / `fetch failed` / `detached` are
   reported as **UNMEASURED**, never folded into a clean count.
+  🔴 **The unit is the package's own `srcDir` SUBTREE, not the repo** — escalating
+  on the repo made rc 17 a permanently-red gate. Measured: the workbench sat 18
+  commits behind `origin/trunk` with **0** of them touching `containers/clawgate`,
+  and over 14 days that repo took 98 commits of which only 32 could reach any
+  built artefact. So the verdict is a pathspec-limited count against the branch's
+  **own** upstream (never a hardcoded `main`/`trunk`), the repo-wide numbers are
+  printed beside it as information, and the cross-host comparison diffs **subtree
+  tree OIDs** rather than repo HEADs.
   🔴 **rc 16 is NOT drift** — it is the fuzzyclaw phase-2 gate reporting that zero rows
   still take their age from fuzzyclaw alone, i.e. the readers can now be deleted; the
   final line says `ACTIONABLE (not drift)` and it is the least severe code, so it can
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -47,6 +47,19 @@ there. Only what's specific to this repo, where a working tree is also a **deplo
   block prompts for what the other allows, which is a real gap. Close it with
   `scripts/sync-claude-permissions.py` (idempotent, additive, **curated** — never copy the
   other host's block wholesale; #380 exists because that file accretes junk).
+  🔴 **Git parity is not SOURCE parity either — rc 17.** Some `nix/pkgs/**`
+  derivations build from a **local working tree of another repo** (`${workspace}/…`:
+  today `homelab-talos` and `tmux-fuzzyclaw`), and **nothing converges those** —
+  `ship.sh` is scoped to `~/workspace/devrc`. On 2026-08-14 the laptop's
+  homelab-talos was 24 commits behind, so it shipped a `clawgatectl` with no
+  `task status`, labelled `0.7.95` by devrc's version literal; the command printed
+  help and **exited 0**, and drift-check was green throughout. Both halves are now
+  closed: `clawgatectl.nix` **reads its version out of the Go source it compiles**
+  (never a literal — an unparseable source means no binary, not a wrong label), and
+  drift-check reports each source repo's currency per host. The covered set is
+  **derived from `nix/pkgs/` at scan time and pinned two-way by a test**, so a third
+  such package is covered automatically. `absent` / `fetch failed` / `detached` are
+  reported as **UNMEASURED**, never folded into a clean count.
   🔴 **rc 16 is NOT drift** — it is the fuzzyclaw phase-2 gate reporting that zero rows
   still take their age from fuzzyclaw alone, i.e. the readers can now be deleted; the
   final line says `ACTIONABLE (not drift)` and it is the least severe code, so it can
--- a/claude/skills/clawgate/reference/task-api.md
+++ b/claude/skills/clawgate/reference/task-api.md
@@ -109,6 +109,18 @@ is answered structurally, not by promising to keep it fresh: response bodies are
 prints a one-line `note: server X, clawgatectl built for Y` skew warning **to stderr** when the two
 disagree. What is NOT answered: a route the CLI has never heard of is still a route you must curl.
 
+🔴 **That skew note was MUTE for 19 hours on the laptop, and the argument above was wrong until
+2026-08-18.** `clawgatectl.nix` used to carry a hand-written `version` literal and stamp it with
+`-ldflags -X main.buildVersion=`, so on 2026-08-14 the laptop — whose `~/workspace/homelab-talos`
+was 24 commits behind — built a CLI with **no `task status` and no `task comment`** and the nix
+literal labelled it `0.7.95` anyway. `h.Version == buildVersion` compared equal, the note stayed
+silent, and `clawgatectl task status <id> in_progress` printed help and **exited 0**. The fix is
+in devrc: the version is now **read out of `cmd/clawgatectl/client.go`'s own `var buildVersion`**,
+so a stale checkout reports its real version and the skew note fires correctly. Two consequences
+worth knowing: **the note is only as good as the checkout's freshness** — `drift-check.sh` rc 17
+now reports a stale `homelab-talos` per host — and an **unparseable** `var buildVersion` line means
+the package is not installed at all (`clawgatectl: command not found`), never a guessed label.
+
 It reads `CLAWGATE_API_URL` + `CLAWGATE_HOOK_TOKEN` out of `~/.claude/clawgate.env` itself, so the
 `H="Authorization: Bearer $(grep '^CLAWGATE_HOOK_TOKEN=' … | cut -d= -f2)"` preamble that used to
 head every recipe in this skill **is gone** — the token never reaches argv (`/proc` is world
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -2259,6 +2259,18 @@ in
   # over a run where nothing is wrong. `SuccessExitStatus = 16` on the service
   # below; the full argument is there.
   #
+  # 🔴 rc 17 DOES TOAST, and that is the point. It means a repo devrc BUILDS A
+  # PACKAGE FROM (`nix/pkgs/**` derivations with a `${workspace}/…` src) is not
+  # current on that host — so the binary it installs is old code under a
+  # current-looking version. Measured 2026-08-14: the laptop's homelab-talos was
+  # 24 commits behind, so its clawgatectl had no `task status`/`task comment`
+  # while devrc's version literal stamped 0.7.95 onto it; the command printed
+  # help and exited 0. This deadman was green on that host throughout. Unlike
+  # rc 13 and rc 16 this is a real divergence with a real fix (a pull plus a
+  # switch), so it must reach OnFailure like rc 8 does — it is NOT on
+  # SuccessExitStatus. It cannot become permanently red: `fetch failed`,
+  # `absent` and `detached` are all reported as UNMEASURED and set no code.
+  #
   # The service is emitted UNCONDITIONALLY (any host can run it by hand); only the
   # TIMER's timers.target wiring is gated — see enableDriftDeadman above.
   systemd.user.services.drift-check = {
@@ -2300,10 +2312,28 @@ in
       # -u drift-check | grep ACTIONABLE` is the check that works. Pinned by
       # `test_the_unit_does_not_fail_on_the_phase2_actionable_code`.
       SuccessExitStatus = 16;
-      # Two `git fetch`es plus one ssh round trip. The ConnectTimeout inside the
-      # script is 10s, so this ceiling only ever fires on a wedged fetch; the
-      # cgroup is killed and the timer re-arms on the next OnUnitActiveSec.
-      TimeoutStartSec = 180;
+      # 🔴 THE BUDGET IS A FUNCTION OF WHAT THE SCRIPT FETCHES, and that grew.
+      # It was 180 for "two `git fetch`es plus one ssh round trip" — the two
+      # devrc checkouts. The source-repo leg (rc 17) fetches EVERY repo nix/pkgs
+      # builds a package from, on BOTH hosts, so the worst case is now
+      #   2 hosts x N source repos x DRIFT_SRC_FETCH_TIMEOUT (30s)
+      # on top of the devrc fetches, the ssh round trip and the 60s phase-2 cap.
+      # At today's N=2 that is 120s of new worst case, which 180 could not
+      # absorb: the cgroup would be killed mid-run and the deadman would report
+      # NOTHING, on a schedule, looking like a unit that merely takes a while.
+      # 420 leaves headroom for one more source repo without another edit.
+      #
+      # This is a SEAM — the tunable lives in the script, the ceiling lives here,
+      # and neither file's tests owned their product. Pinned by
+      # `test_drift_check.py::test_the_unit_start_timeout_can_absorb_every_source
+      # _repo_fetch`, which recomputes it from both files and fails if either
+      # moves out from under the other.
+      #
+      # The ConnectTimeout inside the script is 10s and each source fetch is
+      # capped individually, so this ceiling only ever fires on several wedges at
+      # once; the cgroup is killed and the timer re-arms on the next
+      # OnUnitActiveSec.
+      TimeoutStartSec = 420;
       Environment = [
         # iproute2 is load-bearing, not incidental: `ip -4 -o addr show` is how
         # local_ipv4s identifies WHICH host this is (both report hostname `nixos`).
--- a/nix/pkgs/tools/clawgatectl.nix
+++ b/nix/pkgs/tools/clawgatectl.nix
@@ -24,33 +24,109 @@
 #   • An uncommitted edit in that tree changes the src hash, so the next switch
 #     rebuilds. That is a cost, not a bug.
 #
+# 🔴 THE VERSION IS READ OUT OF THE SOURCE. IT IS NOT WRITTEN HERE. Two repos,
+# two delivery mechanisms, and only one of them is automatic — a hand-maintained
+# `version = "x.y.z"` literal in THIS file is a claim about code that lives in
+# the OTHER one, and nothing keeps the two together. Measured 2026-08-14:
+#
+#   17:44  homelab-infra 978c549c (#323) added `task status` + `task comment` to
+#          clawgatectl and set the Go default buildVersion to "0.7.95".
+#   17:48  devrc 1b790b3 (#483) bumped THIS FILE's literal 0.7.87 -> 0.7.95 and
+#          shipped to both hosts. The changed ldflag forced a rebuild on both.
+#
+# The workbench's homelab-talos was current, so it got a genuine 0.7.95. The
+# laptop's was frozen 24 commits back at 6e9055f7; its client.go correctly said
+# "0.7.87", and the ldflag OVERWROTE that truth with "0.7.95". The laptop then
+# carried a binary with no `task status` and no `task comment` wearing the label
+# of one that had both: `clawgatectl task status <id> in_progress` printed help
+# and exited 0 — a silent no-op, against a CLI whose write ritual the `clawgate`
+# skill makes mandatory. No check anywhere could see it (drift-check.sh was
+# green on that host throughout; it only knew about devrc).
+#
+# So `version` is now DERIVED from the very file being compiled — the Go default
+# at cmd/clawgatectl/client.go — and the ldflag stamps that same derived value.
+# Both the store path and the stamped buildVersion therefore describe the code
+# that is actually in the tree. A stale checkout now reports its own real
+# version, which is the point: client.go compares the server's /health version
+# against buildVersion by EXACT EQUALITY and prints a one-line skew note when
+# they differ, so a 0.7.87 binary talking to a 0.7.95 server says so. Had this
+# been in place on 08-14 the laptop would have printed exactly that.
+#
+# 🔴 NO FINGERPRINT, SUFFIX OR "-dirty" MARKER MAY BE APPENDED TO buildVersion.
+# That equality check is what makes the skew note meaningful; any suffix makes
+# it fire on every command on every host forever, and a permanently-noisy
+# warning is worse than none — it teaches the operator to ignore the one line
+# that has to keep its meaning. Let the existing mechanism work.
+#
+# 🔴 AN UNPARSEABLE SOURCE MUST NOT FALL BACK TO A LITERAL — that is the lie
+# again, in a new shape. It sets available = false instead, so the binary is
+# simply not installed and the failure is LOUD at use time (`clawgatectl:
+# command not found`) rather than silent at build time. Failing the switch is
+# the worse outcome and is deliberately not what happens: ship.sh reports a
+# failed switch as a SKIPPED host, which this repo's CLAUDE.md documents as the
+# failure mode that silently stops all future delivery to that machine.
+#
 # If clawgatectl ever needs to build on a host without the checkout, the fix is
-# to move it to its own public repo and swap this for fetchFromGitHub — the rest
-# of this derivation is unchanged by that.
+# to move it to its own public repo and swap this for fetchFromGitHub — the
+# version extraction below is unchanged by that, since it reads the source tree
+# whatever produced it.
 { pkgs, workspace }:
 
 let
-  version = "0.7.95";
-
   # The whole Go module, not just cmd/clawgatectl: go.mod/go.sum live at its
   # root and buildGoModule needs them for the vendor derivation.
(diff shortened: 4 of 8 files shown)


ZacxDev:
## Red-at-base / green-at-HEAD matrix

Base = **`a2707be`** (`origin/main` at branch point). Each file was reverted to its base
version with `git checkout a2707be -- <path>` and the new tests re-run against it; every
row below was **watched**, not inferred.

### `scripts/tests/test_drift_check.py` — 24 new (213 → 237 collected)

`git checkout a2707be -- scripts/drift-check.sh`, then the new node ids:

| test | base `a2707be` | HEAD |
|---|---|---|
| `test_a_source_repo_behind_its_upstream_is_rc17` | 🔴 FAIL | ✅ |
| `test_a_source_repo_with_UNPUSHED_commits_is_rc17` | 🔴 FAIL | ✅ |
| `test_a_current_source_repo_is_green` | 🔴 FAIL | ✅ |
| `test_the_examined_count_is_reported_beside_the_stale_count` | 🔴 FAIL | ✅ |
| `test_a_scan_that_found_no_source_packages_says_so_instead_of_reporting_zero` | 🔴 FAIL | ✅ |
| `test_an_absent_source_repo_is_reported_as_unmeasured_not_as_clean` | 🔴 FAIL | ✅ |
| `test_a_failed_fetch_is_reported_as_unmeasured_not_as_clean` | 🔴 FAIL | ✅ |
| `test_a_detached_source_repo_is_reported_as_unmeasured_not_as_clean` | 🔴 FAIL | ✅ |
| `test_a_dirty_source_repo_is_reported_even_when_it_is_current` | 🔴 FAIL | ✅ |
| `test_the_source_repo_set_is_pinned_two_way_against_nix_pkgs` | 🔴 FAIL | ✅ |
| `test_a_third_source_package_is_covered_with_no_change_to_the_checker` | 🔴 FAIL | ✅ |
| `test_two_derivations_sharing_one_source_repo_are_fetched_once` | 🔴 FAIL | ✅ |
| `test_two_source_repos_named_on_ONE_line_are_both_found` | 🔴 FAIL | ✅ |
| `test_a_workspace_path_inside_a_comment_is_not_a_source_repo` | 🔴 FAIL | ✅ |
| `test_unpushed_devrc_commits_still_outrank_a_stale_source_repo` | 🔴 FAIL | ✅ |
| `test_a_stale_source_repo_outranks_dangling_symlinks_and_a_behind_checkout` | 🔴 FAIL | ✅ |
| `test_the_rc17_legend_is_printed_with_the_verdict` | 🔴 FAIL | ✅ |
| `test_the_cross_host_source_comparison_is_NOT_COMPARED_with_one_host` | 🔴 FAIL | ✅ |
| `test_the_cross_host_comparison_reports_differing_heads_without_setting_an_rc` | 🔴 FAIL | ✅ |
| `test_the_cross_host_comparison_reports_agreement_when_the_heads_match` | 🔴 FAIL | ✅ |
| `test_a_run_against_a_stale_source_repo_mutates_nothing` | 🔴 FAIL | ✅ |
| `test_a_non_integer_source_fetch_timeout_is_rejected` | 🔴 FAIL | ✅ |
| `test_the_unit_start_timeout_can_absorb_every_source_repo_fetch` (also reverting `nix/home.nix`) | 🔴 FAIL | ✅ |
| `test_the_source_fetch_timeout_is_not_forwarded_over_ssh` | ⚠ **PASS** | ✅ |

⚠ **INVARIANT GUARD, labelled as such in its own docstring** — at base the variable did
not exist, so "it is not forwarded" is trivially true. It pins a property of the new code
that nothing else asserts; it is **not** counted as regression coverage. It carries a
positive control (the slice it inspects must contain the values that *are* forwarded) so
it cannot pass against an empty string.

### `scripts/tests/test_clawgatectl_version.py` — 18 new (new file)

`git checkout a2707be -- nix/pkgs/tools/clawgatectl.nix nix/pkgs/tools/tmux-fuzzyclaw.nix`
→ measured **16 failed, 2 passed**:

| test | base `a2707be` | HEAD |
|---|---|---|
| `test_the_version_comes_from_the_go_source[0.7.87]` | 🔴 FAIL | ✅ |
| `test_the_version_comes_from_the_go_source[9.9.9]` | 🔴 FAIL | ✅ |
| `test_the_version_comes_from_the_go_source[1.2.3-rc4]` | 🔴 FAIL | ✅ |
| `test_an_unparseable_source_yields_NO_PACKAGE_not_a_fallback[declaration renamed]` | 🔴 FAIL | ✅ |
| `test_an_unparseable_source_yields_NO_PACKAGE_not_a_fallback[no literal to read]` | 🔴 FAIL | ✅ |
| `test_an_unparseable_source_yields_NO_PACKAGE_not_a_fallback[const, not var]` | 🔴 FAIL | ✅ |
| `test_an_unparseable_source_yields_NO_PACKAGE_not_a_fallback[ambiguous: two matches]` | 🔴 FAIL | ✅ |
| `test_an_unparseable_source_yields_NO_PACKAGE_not_a_fallback[client.go absent entirely]` | 🔴 FAIL | ✅ |
| `test_the_evaluator_can_tell_available_from_unavailable` | 🔴 FAIL | ✅ |
| `test_no_hardcoded_version_literal_survives_in_the_nix_file` | 🔴 FAIL | ✅ |
| `test_the_version_is_read_from_the_compiled_source_file` | 🔴 FAIL | ✅ |
| `test_the_ldflag_is_fed_the_derived_value_and_not_a_second_literal` | 🔴 FAIL | ✅ |
| `test_no_fingerprint_or_suffix_is_appended_to_the_stamped_version` | 🔴 FAIL | ✅ |
| `test_an_unparseable_source_switches_the_package_OFF_rather_than_failing` | 🔴 FAIL | ✅ |
| `test_the_header_no_longer_describes_a_hand_maintained_version` | 🔴 FAIL | ✅ |
| `test_tmux_fuzzyclaw_is_documented_as_having_no_version_source_of_truth` | 🔴 FAIL | ✅ |
| `test_the_stamped_ldflag_and_the_store_version_are_the_same_string` | ⚠ **PASS** | ✅ |
| `test_a_checkout_too_old_to_have_main_go_still_yields_no_package` | ⚠ **PASS** | ✅ |

⚠ Both **INVARIANT GUARDS**, labelled in their docstrings. The old file stamped its own
literal, so `version` and the ldflag already agreed — what they never did was describe the
*source*; and the `main.go` `pathExists` guard predates this change, so that test pins
that the new `&& parsedVersion != null` conjunct did not weaken it.

**Totals: 39 of 42 red at `a2707be`, 3 invariant guards.**

---

## Mutation battery — 19/19 killed

Run under `PYTHONDONTWRITEBYTECODE=1` with `-p no:cacheprovider`; each mutant applied to a
`cp -a` backup-restored file, one mutant at a time, mutating the **narrowest** expression
that can be wrong (never a guard together with its enclosing condition).

| mutant | verdict |
|---|---|
| **POSITIVE CONTROL** — the drift verdict never fires (`s_rc=17` → `s_rc=0`) | KILLED |
| ABSENT counted as clean instead of unmeasured | KILLED |
| a failed fetch counted as clean instead of unmeasured | KILLED |
| comment truncation removed — prose paths become source repos | KILLED |
| only the FIRST `${workspace}/` on a line is read | **SURVIVED → fixed, now KILLED** |
| the walk stops at `nix/pkgs`' top level (misses `tools/`) | KILLED |
| `NOT EVALUATED` replaced by a count triple over zero repos | KILLED |
| rc 17 ranked BELOW rc 14 and rc 10 | KILLED |
| rc 17 ranked ABOVE rc 8 | KILLED |
| the cross-host difference escalates to an exit code | KILLED |
| the per-host rc never reaches the driver (`note_rc "$(src_rc_of …)"` deleted) | KILLED |
| dirty only reported when the repo is otherwise stale | KILLED |
| upstream comparison hardcodes `main` instead of the branch's own | KILLED |
| only BEHIND counts — un-pushed source commits go unreported | KILLED |
| NIX: version falls back to a literal when unparseable | KILLED |
| NIX: two matching lines resolve to the first one | KILLED |
| NIX: the ldflag is stamped from a literal again | KILLED |
| NIX: a fingerprint suffix is appended to `buildVersion` | KILLED |
| NIX: availability no longer requires a parsed version | KILLED |

🔴 **The survivor is the finding.** `LN="$REST"` → `LN=""` (stop after the first
occurrence on a line) passed the **entire** suite, **including the two-way pin** — because
every line in the real `nix/pkgs` names at most one source, so the shell extraction and
the Python oracle shared the blind spot and agreed while both were wrong. A line like
`let a = "${workspace}/x"; b = "${workspace}/y";` is legal nix and would have silently
stopped being watched. `test_two_source_repos_named_on_ONE_line_are_both_found` is the
fixture the pin structurally cannot build for itself.

ZacxDev:
## Gate

```
scripts/gate.sh   →  GATE: RESULT=PASS exit=0
    pytest   RESULT: PASS (exit=0)     scripts/tests collected=5228 passed=5228 failed=0
    node     RESULT: PASS (exit=0)     TOTAL suites=4 files=33 tests=1116 pass=1116 fail=0

nix build .#checks.x86_64-linux.pytests .#checks.x86_64-linux.nodetests  →  exit 0
```

The flake-check status was re-read **unpiped** (`nix build …; echo $?`), not through a
`| tail` that would have handed back tail's status.

**`TARGET_FLOORS` needed no re-pin**: `scripts/tests` collected **5228** against floor
**4976** and a drift ceiling of **6220** (`floor + max(60, floor/4)`), so neither the
collapse branch nor the too-much-slack branch fires and the gate printed no replacement
number.

⚠ **One earlier gate run reported `FAIL exit=1`, and it was an artifact of my shell, not
of this branch.** `run-tests.sh` requires `opencode` on PATH; my first ad-hoc
`nix-shell -p` pulled **1.18.18** from the channel while `flake.lock` pins **1.18.16**, so
`test_opencode_engine.py::test_engine_is_the_version_every_measurement_is_keyed_to` went
red — exactly the drift that test exists to catch. It was the **only** failure in the
whole tier (`collected=5228 passed=5227 failed=1`). Re-running with the flake-pinned
binary on PATH (`/nix/store/rcrzfd71k96f1r55533lzc16p7ix3v22-opencode-1.18.16`) is the
green above, and the sandboxed flake check — which uses the pinned one by construction —
passed both times.

ZacxDev:
## Follow-up: rc 17 was a permanently-red gate. Fixed in `c4474d5`.

Found in independent verification, not by me. rc 17 escalated on **whole-repo**
staleness, and that is the wrong unit.

Reproduced: the workbench was behind by `2ce7cbdc fix(naida-ai-demo): raise memory
limit 128Mi -> 512Mi`; `git diff --name-only HEAD..origin/trunk -- containers/clawgate`
is **empty**, so the commit cannot reach the built binary — and it still produced a
loud rc 17 with a DND-bypassing toast. Over 14 days that repo took **98 commits, only
32 touching `containers/clawgate`**. At ~7 commits/day the host is behind nearly
continuously and about two thirds of those reds could not affect any package devrc
builds. That is precisely the failure `RULES.md` names — and it would have landed on
the deadman whose entire value is being believed.

### What changed

The verdict is now computed over the **package's own `srcDir` subtree**:

- The scan keeps the **whole** `${workspace}/…` path, so repo and subtree fall out of
  the same derivation. Nothing is hardcoded, and **one repo may carry several scopes**,
  each judged separately.
- Counts are pathspec-limited against the branch's **own** upstream
  (`HEAD..@{u} -- <subtree>` and the reverse) — never a hardcoded `main`/`trunk`.
- `tmux-fuzzyclaw`'s `srcDir` **is** its repo root, so scope == repo and its behaviour
  is unchanged.
- 🔴 **Repo-wide counts are still printed, as information.** The output distinguishes
  "behind, but not in anything we compile" from "the built source is stale":

```
source repo homelab-talos: on branch trunk at 0400066096c3 — repo-wide 18 behind / 0 ahead of refs/remotes/origin/trunk.
  repo-wide is INFORMATION ONLY. Only the built-source scope(s) below set the verdict.
  BUILT SOURCE homelab-talos/containers/clawgate is CURRENT (0 behind / 0 ahead) — the repo-wide 18 behind /
  0 ahead touch nothing this package is built from. Information, NOT drift.
```

- The **cross-host** comparison now diffs `srcDir` **subtree tree OIDs**, not repo
  HEADs, so "the two hosts build DIFFERENT source" stops firing on unrelated cluster
  manifests.
- A repo that could not be evaluated marks **every** scope under it UNMEASURED — the
  old per-repo counting left a second package in a shared repo accounted for nowhere.
- Still READ-ONLY. `examined=N stale=M unmeasured=K` still holds; the unit counted is
  now the built-source scope, with the repo count printed beside it.

Two units, and conflating them is what caused this: a **repo** is what gets FETCHED
(once, however many packages sit in it); a **scope** is what gets JUDGED.

### Red at `b10c4ae` / green at HEAD

Measured by `git checkout b10c4ae -- scripts/drift-check.sh` and re-running.

| test | `b10c4ae` | HEAD |
|---|---|---|
| `test_a_commit_OUTSIDE_every_srcDir_is_reported_but_is_NOT_rc17` | 🔴 `assert 17 == 0` | ✅ |
| `test_a_commit_INSIDE_a_srcDir_is_still_rc17` | 🔴 FAIL | ✅ |
| `test_UNPUSHED_commits_inside_a_srcDir_still_escalate_and_outside_do_not` | 🔴 FAIL | ✅ |
| `test_two_subtrees_of_ONE_repo_are_judged_separately` | 🔴 FAIL | ✅ |
| `test_an_unmeasurable_repo_makes_EVERY_scope_under_it_unmeasured` | 🔴 FAIL | ✅ |
| `test_the_cross_host_comparison_ignores_divergence_OUTSIDE_every_srcDir` | 🔴 FAIL | ✅ |
| `test_the_cross_host_comparison_still_reports_a_differing_built_subtree` | 🔴 FAIL | ✅ |
| `test_the_source_repo_set_is_pinned_two_way_against_nix_pkgs` | 🔴 FAIL | ✅ |
| `test_a_root_srcDir_package_is_unchanged_by_the_scoping` | 🔴 FAIL (see below) | ✅ |

🔴 **Honest scoping of two of those rows.** `test_a_commit_INSIDE_a_srcDir_is_still_rc17`
and `test_a_root_srcDir_package_is_unchanged_by_the_scoping` went red at `b10c4ae`
**only on message wording this commit renamed** — their *behavioural* claim (rc 17)
already held there. They are **not** regression coverage for this defect; they are the
guard against "fixing" the noise by breaking the detector, and their value is
demonstrated by the mutation battery below, not by their base colour. The root-srcDir
one now says so in its own docstring. The regression is the first row, and it fails at
base with `assert 17 == 0`.

### Isolated mutation battery — 12/12 killed

`PYTHONDONTWRITEBYTECODE=1`, `-p no:cacheprovider`, one mutant at a time restored from a
`cp` backup, each the **narrowest** expression that can be wrong.

| mutant | verdict |
|---|---|
| **POSITIVE CONTROL** — `s_rc=17` → `s_rc=0` | KILLED (own message) |
| the **behind** pathspec is dropped | KILLED (own message) |
| the **ahead** pathspec is dropped | KILLED (own message) |
| the pathspec is applied to the **root-srcDir** case too | KILLED (own message) |
| the scan collapses back to repo roots (`SUB=""`) | KILLED — by the two-way pin's own `drift-check.sh's own scan disagrees with nix/pkgs` |
| scopes keyed by repo only (`s_key` drops the subtree) | KILLED — by the scope-key set assertion |
| repo-wide numbers stop being reported | KILLED (own message) |
| only ONE scope per unmeasurable repo | KILLED (own message) |
| cross-host facts revert to repo HEAD | KILLED (own message) |
| …plus the 3 from the original battery that touch this code | KILLED |

One mutant in the first pass was **mis-designed by me** — it weakened a test's own
assertion (`assert True or …`) rather than the code, and unsurprisingly survived. It
was replaced by the correct source-level control (collapse the scan to repo roots),
which kills. Worth recording: the "scopes keyed by repo only" mutant did **not** move
the `examined/stale` triple — `examined=2 stale=1` was preserved while both scopes
collapsed to the name `homelab-talos`. Only the **fact-key set** assertion caught it.

### Live positive + negative control

Read-only on the real checkouts (cloned aside, origin repointed at the real private
GitHub URL, only the clone fetched or moved; no `reset --hard` anywhere). The two runs
differ by **one commit**, and it is a real clawgatectl one:

| control | repo-wide behind | `containers/clawgate` behind | SRC-RC |
|---|---|---|---|
| **negative** — behind only outside the srcDir | **18** | **0** | **0** (NOT drift) |
| **positive** — one clawgatectl commit inside | **19** | **1** | **17** 🔴 |

The positive case's differing commit is
`fix(clawgatectl): every heredoc comment was accused of losing a rune it never lost (#325)`.
Under `b10c4ae` the negative case would have been a loud rc 17.

`tmux-fuzzyclaw` reported `no upstream to compare against, currency NOT evaluated` in
both runs — it genuinely sits on an untracked local branch, which remains UNMEASURED
rather than drift.

ZacxDev:
## Gate on the fixed tree (`c4474d5`)

```
scripts/gate.sh   →  GATE: RESULT=PASS exit=0
    pytest   RESULT: PASS (exit=0)     scripts/tests collected=5245 passed=5245 failed=0
    node     RESULT: PASS (exit=0)

nix build .#checks.x86_64-linux.pytests .#checks.x86_64-linux.nodetests  →  exit 0
```

Both flake checks were **built fresh, not cached** — the tree changed, so nix rebuilt
`devrc-pytests.drv` and `devrc-nodetests.drv` (both appear in the build list). The exit
status was read **unpiped**.

`logrotate` is absent from the plain shell here (supplied only by `flake.nix:248`), so
`gate.sh` was run under a `nix-shell` carrying every `REQUIRED_TOOLS` entry **plus the
flake-pinned opencode 1.18.16** — without that pin `test_opencode_engine.py` goes red on
the channel's 1.18.18, which is pre-existing and unrelated to this branch.

`TARGET_FLOORS` still needs no re-pin: `scripts/tests` collected **5245** against floor
**4976** and a drift ceiling of **6220**, and the gate printed no replacement number.

PR #537, opened 2026-08-18
fix(signal): outbound reactions were dropped — the sync branch handled remoteDelete but not its sibling

ZacxDev:
## What

A reaction Zach makes on his **own phone** arrives as `syncMessage.sentMessage.reaction` with `message: None`. The sync branch in `consumer.py` runs **before** the inbound `dataMessage` branch and handled only `remoteDelete`, so the reaction fell through to `_base_message()` and produced **exactly the two defects the remote-delete case exists to prevent**:

- the reaction dropped from `signal.reactions`
- a bodyless ghost row left in `signal.messages`

This adds the `reaction` case to the sync branch, mirroring the `remoteDelete` one.

## How it was found

**Step 7** — sending real messages from the phone. Not by review: six PRs, four audit rounds and 387 tests passed over it. The sync branch had *already* been patched once for this exact asymmetry (`remoteDelete`), and the sibling shape was still missed.

> When you fix an own-device branch for one message shape, enumerate the others in that wrapper.

## 🔴 Why it nearly read as clean

Two **other** group members had reacted to the same message, so `signal.reactions` held **2 rows carrying that exact `target_sent_timestamp`** — a count check says "fine". The account's own contact appears in **none** of them.

Only the reactor **identity** discriminates stored from lost here. The regression test asserts identity for that reason, and the mutation battery confirms a count assertion would not have caught it.

## Verification

| | |
|---|---|
| red at `origin/main` | 3 failed — `sync_outbound` where `reaction` belongs |
| green at HEAD | 400 passed (`scripts/signal/tests`) |
| hermetic gate | `RESULT: PASS (exit=0)` — 11,689 collected, 0 failed |

Five-mutant battery, **all KILLED**, run under `PYTHONDONTWRITEBYTECODE=1` so a same-length edit cannot score a false SURVIVED off a stale `.pyc`:

| mutant | result |
|---|---|
| M1 reactor identity → target author | KILLED — dies on its own assertion |
| M2 `is_remove` hardcoded False | KILLED |
| M3 drop the `targetSentTimestamp` guard | KILLED |
| M4 `is not None` → never taken | KILLED |
| M5 whole case removed (positive control) | KILLED |

Tree checksummed byte-identical before/after the battery.

## 🔴 Merging this does not fix the cluster

The consumer runs a Harbor image **pinned by digest**, so this needs `scripts/signal/build-push.sh <ver>` plus a homelab digest bump. Until that lands, every reaction sent from the phone is still being dropped.

The one already lost **is recoverable** — `raw_envelope` on `signal.messages` id 3 still holds it.

## Not covered here

The attachment/MinIO leg has still never carried real traffic (`attachments` = 0). That zero is attributable to input — no stored envelope contains an attachment key — not to a defect. Sending an image is the outstanding check.

### DIFF
--- a/claude/skills/signal/SKILL.md
+++ b/claude/skills/signal/SKILL.md
@@ -127,7 +127,7 @@ operator token as `approve`):
 |---|---|
 | `message` | a 1:1 message (attachments, quotes, view-once and expiring messages are all this kind, with fields set) |
 | `group_message` | a message carrying `groupInfo` |
-| `reaction` | an emoji reaction (may arrive BEFORE its target — see gotchas) |
+| `reaction` | an emoji reaction (may arrive BEFORE its target — see gotchas). Covers BOTH directions: someone reacting to you, and 🔴 **your own reactions sent from the phone**, which arrive in the `syncMessage.sentMessage` wrapper and are stored with **the account itself** as `reactions.contact_id` |
 | `edit` | an edit of an earlier message (`edit_target_timestamp` points at it) |
 | `sync_outbound` | a message Zach sent from another linked device, echoed back |
 | `receipt_delivery` | a delivery receipt |
@@ -186,6 +186,16 @@ and tear it down on exit — exactly like `mail-actions`.
 - **Reactions can precede their target message** (delivery is not ordered, and history is not
   backfilled). `reactions.message_id` is NULLable, resolved later, with the partial index
   `idx_rx_unresolved`. Unresolvable reactions are RETAINED, not dropped.
+- 🔴 **Filtering reactions to "what people sent me" must exclude the account's own
+  `contact_id`** — your own reactions live in the same table. Before 2026-08-18 they were
+  being DROPPED entirely, so any analysis over reactions predating that is missing them.
+- 🔴 **When you fix an own-device branch for one message shape, enumerate the others in
+  that wrapper.** `syncMessage.sentMessage` carries `remoteDelete`, `reaction` and a plain
+  `message`. The remote-delete case was fixed first and the reaction case was still missed
+  for weeks — dropped reaction plus a bodyless ghost row in `signal.messages`. Found only by
+  real traffic, after six PRs, four audit rounds and 387 green tests. Any UNMODELLED
+  `sentMessage` variant with `message: None` (e.g. a nested `editMessage`, `sticker`,
+  `payment`) still produces such a ghost row — that is open work, not a solved problem.
 - **Redelivery duplicates attachments** without `UNIQUE (message_id,
   signal_attachment_id)`.
 - **One person must be ONE contact row, in both arrival orders.** An identity arrives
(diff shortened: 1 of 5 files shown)


ZacxDev:
## Two audit rounds since the description was written — superseding its numbers

The body above describes the **first commit only**. It is not wrong, but its verification figures are now stale and understated. Current state:

| | body says | actual |
|---|---|---|
| signal suite | 400 passed | **420 passed** |
| full gate | 11,689 collected | **11,709 collected, 0 failed** |
| mutants | 5, all killed | **24 across three batteries, zero survivors, none deletion-only** |

### Round 1 — one 🔴, and it was not in what the fix did

`upsert_reaction` resolves **both** the target author and the reactor through `upsert_contact`, which raises `ValueError` on an unidentifiable contact. `ValueError` is neither `MalformedEvent` nor a `TRANSIENT_DB_ERROR`, so it escaped `store()`, escaped `handle_payload()` (which wraps only the *parse*), and landed in `run()`'s reconnect handler as a "dropped stream". Signal redelivers unacked frames — an unbounded loop storing nothing, taking every later frame on that connection with it. Measured `reconnects=3 / stored=0` where `main` gives `stored=3`.

Fixed by **consolidating** the reaction parser into one `_reaction_from()` helper serving both the sync and inbound sites, rather than guarding only the new branch — the duplication is *why* the sync site shipped without guards the inbound site already had.

### Round 2 — clean of 🔴, and it proved the consolidation safe rather than asserting it

A 640-envelope differential of the inbound reaction space, old parser vs new: 15 envelopes moved `OK` → `MalformedEvent`, and each was pushed through the DB layer with the **old** parser to confirm it already raised the wedging `ValueError`. So the guards reject exactly the set that previously took the consumer off the air; no envelope that ever stored successfully is now rejected. 256 further envelopes moved `AttributeError` → `MalformedEvent`.

Its 🟡 was a fair hit: the round had consolidated the parser **body** and left the dispatch **predicate** divergent (sync `is not None`, inbound truthiness), so an inbound `reaction: {}` stored as an ordinary message while the identical sync shape was reported malformed. Same "one rule, two places" this PR is about, one level up. Fixed in `7ac796a`.

### Three harness defects found while doing this, each of which reported false coverage

- A mutant came back **ANCHOR-MISS, not KILLED** — the anchor also matched an unrelated branch, so it never landed. A mutant that never ran is not a survivor, and it is not a pass either.
- Two **operand-order** mutants survived a green 416-test run because every fixture set `source` and `sourceNumber` to the *same* value, collapsing both implementations into identical output. Not equivalent mutants: newer signal-cli can put a UUID in `source`, so the order decides whether a UUID is written into a phone-number column.
- The fixture emoji equalled both the corpus value **and** the asserted constant, so a mutant hardcoding it survived 400 tests while corrupting every stored reaction.

### Still deliberately out of scope

- **Ghost rows already stored** — this stops creating them, it does not remove the one that exists. `list_conversations()` renders it as a `NULL`-named conversation sorted to the top.
- **The outbound-`editMessage` ghost** in the same wrapper — the next instance of this PR's own lesson, documented in SKILL.md as open work.
- 🔴 **Merging is inert.** The consumer runs a digest-pinned Harbor image, so reactions keep being dropped until `build-push.sh` + a homelab digest bump.

PR #542, opened 2026-08-18
fix(drift-check): UNMEASURED escalated NEVER — a source scope nobody can evaluate now reaches rc 18 after N consecutive runs

ZacxDev:
## The gap

`rc 17` (#536) asks "is the `srcDir` subtree this host **compiles** current with its own upstream?". A scope it **cannot** evaluate is reported `UNMEASURED` and sets no exit code — right for one run, and wrong forever: **nothing escalated it**, so a scope whose currency is never measured read as a clean run indefinitely. That is the same green rc 17 exists to refuse — *"did not look"*, not *"looked and found nothing"* — one level down.

**Measured 2026-08-18 on the workbench** (inherited from the parent session, not re-observed here): `~/workspace/tmux-fuzzyclaw` sat on a local branch `docs/tui-rendering-footguns` with **no upstream**; the run printed `unmeasured=1`, exited 0, and was concealing a genuinely divergent build between the two hosts. That branch was given an upstream before this work started, so **the live fixture was gone and every case here is constructed**.

## What this adds — the rc 13 ladder, reused rather than reinvented

* an unmeasured scope is still reported on **every** run; it escalates to **rc 18** only after N **consecutive** runs, and the streak **resets** the moment it measures;
* the streak is per **(host, scope)** — one repo recovering must not clear another's ladder, and the laptop's blindness is not the workbench's. The counter filename escapes `_`→`__` then `/`→`_`, which is **injective** over the scope alphabet (`[A-Za-z0-9._/-]`): `a/b` and `a_b` cannot share a file;
* a host is walked **only** if it returned a `FACT src-unmeasured examined=N …` line, so an unreachable laptop bumps nothing — rc 13 already owns that finding;
* it fails **closed**: an unpersistable streak escalates immediately, because "for how long" is then unknowable.

### The reasons are not one hazard, so they do not share a counter

| reason | ladder | why |
|---|---|---|
| `NOUPSTREAM` (branch with no upstream / detached) | `DRIFT_UNMEASURED_ESCALATE`, default **4** (~24h) | **structural** — never heals on its own, and the one measured concealing a real divergence. Not instant: parking on a scratch branch for an afternoon is normal, and a gate that fired on that would be red most of the time |
| `NOCOUNT` (a pathspec count that will not parse) | same | structural too — the identical command fails identically next run |
| `FETCHFAILED` | `DRIFT_UNMEASURED_FETCH_ESCALATE`, default **12** (~3 days) | plausibly **transient** (no ssh-agent in a user unit, key rotation, remote down) so it gets more patience — but it still escalates: a currency check that can never fetch is a checker wired to nothing |
| `ABSENT` | **never**, at any count | `nix/pkgs/tools/clawgatectl.nix` deliberately supports a host without the checkout (guards on `pathExists`, omits the binary). Escalating would make that host permanently red for a package it correctly does not ship. Its counter is not even consulted |
| an **unknown** token | structural | the enumeration **fails closed**, like the `settings.json` per-host allowlist |

### It cannot be satisfied by measuring nothing

The block prints `hosts-reporting= scopes= unmeasured= escalated=` and **withholds that summary entirely** when the host count *or* the scope count is zero — a ladder over no scopes is not a clean ladder.

The payload's new `FACT src-unmeasured examined=N …` line carries **its own denominator**, so the driver can separate three states that otherwise collapse: *never ran the payload* (no line), *ran and found no scopes* (`examined=0`), *ran and everything measured* (`examined=N`, no pairs). Read off the existing `src-repos` facts the last two are the same empty value. It also carries the **reason per scope**, which `src-repos` cannot: a scope that fails at the pathspec comparison has already emitted a real tree OID, so "is this value a reason token?" would read it as measured.

## Why rc 18, and why it ranks 59

Free codes below 16 (5, 7, 9, 11) are reserved to `ship.sh` meanings this script does not take, so a new drift code has nowhere to go but upward — same argument rc 17's author made.

`severity()` places it at **59, between 13 (60) and 6 (58)**. The digit is not the severity and that table has never been monotonic.

* **below 13** — same *kind* of finding (a persistent inability to look, escalated only after N consecutive misses), and 13 is that inability about an **entire host** (every check on it, including the un-pushed-commits one) while this is one built-source scope;
* **above 6/4/3** — those are *single-run* "could not evaluate" outcomes that may be gone next run; rc 18 is persistent **by construction**, and what it hides is rc 17 (rank 67), invisible for as long as it holds;
* **above 12/15/10** — those are resolved by `ship.sh` or a documented fix and say exactly what they are.

## The unit

`nix/home.nix` is **comments only**.

* rc 18 is a real divergence with a real fix, so it is **not** excused on `SuccessExitStatus` — `test_only_16_is_excused_from_failing_the_unit` pins that to exactly one code;
* `TimeoutStartSec` stays **420**. That budget is a function of what the script *fetches*; this adds no fetch and no ssh, only a ~16-byte counter per (host, scope) — four files at today's scope count. Moving it for a reason `test_the_unit_start_timeout_can_absorb_every_source_repo_fetch` does not model would decouple the number from its own pin.

The stale sentence in that file ("`fetch failed`, `absent` and `detached` are all reported as UNMEASURED and set no code") is corrected rather than left to rot — it was true per run and false forever.

## Verification

**Base sha `15da990`** (the parent of this branch).

| | at base `15da990` | at HEAD |
|---|---|---|
| 22 new tests | **22 failed** | 22 passed |
| 244 pre-existing tests | 244 passed | 244 passed |
| total | 22 failed, 244 passed | **266 passed** |

Both directions are pinned: under threshold → **no** escalation; at/over → escalation; recovery → **reset**. Five tests carry an `rc==0` / `rc==2` / `rc==8` / `rc==17` half that is green at base — each says **INVARIANT GUARD** in its own docstring and is not counted as regression coverage.

### Isolated mutation — 11 mutants, all killed

Each is the narrowest expression that can be wrong; the tree is restored from a **byte copy** between mutants and the sweep is bracketed by an unmutated **negative control** (266 passed before *and* after).

| mutant | killed by |
|---|---|
| M1 `-ge` → `-gt` on the threshold | `…escalates_after_n_consecutive_runs` (+6) |
| M2 drop the per-**scope** key from the counter filename | `…is_kept_PER_SCOPE`, `…is_kept_PER_HOST`, `…filename_is_INJECTIVE`, `one_scope_recovering…` |
| M3 fail to **reset** on recovery | `…resets_when_the_scope_measures` |
| M4 drop the reason-change restart | `a_CHANGED_reason_restarts_the_ladder` |
| M5 `ABSENT` stops being exempt | `an_ABSENT_source_repo_never_escalates…` |
| M6 drop the whole-string-fallback guard on the pair list | `a_measured_fleet_reports_a_zero_over_a_REAL_denominator` (+3) |
| M7 `-le` → `-lt` on the zero-scope refusal | `a_ladder_over_ZERO_SCOPES…` |
| M8 **positive control** — threshold branch stops setting the rc | `…escalates_after_n_consecutive_runs` (+5) |
| M9 severity rank 59 → 71 | `unpushed_devrc_commits_still_outrank…`, `a_MEASURED_stale_source_still_outranks…` |
| M10 an unwritable counter fails **open** (`-1` → `0`) | `…escalates_when_the_streak_FILE_cannot_be_written` |
| M11 walk a host that never answered | `a_host_that_never_ANSWERED_accumulates_no_ladder` (+10) |

M3 / M5 / M7 / M10 were re-run individually to confirm each dies on **that** guard's own assertion, not a neighbour's.

M10 exists because the rc 13 twin recorded exactly this blind spot: the `mkdir` limb is reachable by pointing the state dir at a regular file, but the `printf … > "$f" || echo -1` limb needs an existing-but-unwritable dir **and** a read-only counter file, and without both the mutant survives.

### Control pair

* **negative control** (can it go red?) — the 11 mutants above, each producing a distinct red;
* **positive control** (can it observe the thing at all?) — `test_a_measured_fleet_reports_a_zero_over_a_REAL_denominator` asserts `hosts-reporting=1 scopes=1 unmeasured=0 escalated=0`: a zero over a **real** denominator, not the zero a ladder wired to nothing prints. A live read-only run on the workbench reported the same shape over the real `nix/pkgs`: `hosts-reporting=1 scopes=2 unmeasured=0 escalated=0`.


ZacxDev:
## Gate + flake checks

All against the content of this branch's single commit.

### `scripts/gate.sh --tier both --set hermetic`

```
======================== GATE ========================
GATE: RESULT=PASS exit=0
```

* pytest tier: `TOTAL collected=11780 passed=11779 skipped=1 failed=0 (floor: 10963 = sum of 24 per-target floors)`
* node tier: `TOTAL suites=4 files=33 tests=1116 pass=1116 fail=0 skipped=0 (floor: 1098)`

Run under a shell carrying the flake-pinned **opencode 1.18.16** (the channel's 1.18.18 reddens `test_opencode_engine.py`), **logrotate**, and a `python312.withPackages` env — all three are supplied by `flake.nix` to the CI check and absent from the plain shell. That is pre-existing, not something this branch introduced.

The gate ran at 16:41→16:52; the last source edit on this tree was 16:39:49, so it covers exactly this content. (`scripts/drift-check.sh` has a later mtime only because the mutation sweep restores it from a byte copy — sha256-asserted identical, and `git status` is clean.)

### Flake checks — **both fresh, neither cached**

| check | how it ran | nix exit | verdict |
|---|---|---|---|
| `.#checks.x86_64-linux.pytests` | **fresh**, 671 build-log lines | `0` | `RESULT: PASS (exit=0)` — `collected=11780 passed=11779 skipped=1 failed=0` |
| `.#checks.x86_64-linux.nodetests` | first attempt returned **cached, zero log lines**; re-run with `--rebuild`, 6812 log lines | `0` | `RESULT: PASS (exit=0)` — `suites=4 files=33 tests=1116 pass=1116 fail=0 skipped=0` |

The `nodetests` cache hit is called out rather than reported as a fresh run — a build that emits no log lines is a claim about the store, not about this tree. `--rebuild` forced real execution.

Both statuses are taken from **nix itself**, not from a trailing `| tail` — an earlier invocation of mine had `FLAKE_RC=$?` after a pipe, which reports `tail`'s status and is the exact trap `CLAUDE.md` records.

### Two things that are *not* failures, flagged so nobody re-derives them

* A `RESULT: FAIL (exit=143)` appears in one abandoned gate log. 143 is SIGTERM — I stopped that redundant second gate run myself to free CPU. It is not a test failure.
* Wall-clock timings from this session are unusable: a sibling agent was running a deliberate 32-process CPU-saturation experiment (load average peaked ~149). Pass/fail is unaffected; **duration-based** reasoning about any of these runs is not valid.

PR #549, opened 2026-08-19
fix(browser-bridge): 6,937 telemetry rows in 14d had an EMPTY session column — the value was already on the wire

ZacxDev:
## The bug

`activity.events` rows with `source='browser-bridge'` had an **empty `session` column — 0 of 6,937 over 14 days**, while `claude`/`opencode`/`keys`/`tmux`/`zsh` filled it 100%. The value already reached the server as `X-Session-Id` (it drives tab-ownership routing); it was simply never passed to `emit_cmd_event`. So a browser-skill call could not be joined to the agent session that made it, and *"which sessions used opencode **and** the browser skill"* had to be answered by scanning 1.5M transcript records.

## What changed

Passing the id through naively would have produced two **wrong** joins instead of no join. Both are closed here.

### 1. Tier — only `claude:` is a join key

`derive_session_id` already tags the id with the tier that produced it, before the first `:`. The server now reads **that tag** (a tag the CLI emits on purpose — never the id's *shape*) and:

| `X-Session-Id` | `payload.sess_src` | `session` column |
|---|---|---|
| `claude:<uuid>` | `claude` | **the bare `<uuid>`** |
| `tmux:%3` | `tmux` | empty |
| `sid:<sid>:<start>` | `sid` | empty |
| `ppid:<pid>:<rand>` | `ppid` | empty |
| `synthetic:…` | `synthetic` | empty |
| untagged / absent / oversized / control-char | `unknown` | empty |

- Only the **first** colon splits (`sid:`/`ppid:` ids contain more).
- The tag is **stripped** from the column value so it compares `=` against `source='claude'` rows with no `replaceOne()` at the join site.
- An **untagged** id — including a bare uuid, the value most tempting to accept — is never promoted. Fail closed.
- An id over 200 chars or carrying a control character is **dropped whole, never truncated**: a truncated join key is a *wrong* join key.
- Stored **raw, not hashed**, as briefed.

### 2. Nesting — `browser agent` forwards its *invoker's* id

`browser-agent:675` captures `--print-session-id` from the session that invoked it and hands it to the nested opencode tool, which sends it verbatim as `X-Session-Id`. Left alone, one `browser agent "<goal>"` call would have become N browser calls credited to the operator's own session (~581 nested rows in 14d, ~11% of bridge commands) — fabricated usage in the `session` JOIN column (**not** "the column adoption-scan reads" — corrected in B2 below).

`browser_tool_impl.mjs` now sends `X-Session-Origin: browser-agent`. When present the server writes **no** `session`; the forwarded id lands in `payload.origin_session` (the causal **parent**) beside `payload.origin`. `X-Session-Id` is untouched, so routing is unchanged. Giving the nested session an id of its own is a later change.

### 3. One forced CLI change

`emulate --recreate` sent its close under a throwaway id built by **appending** to the real one — leaving the live session's `claude:` tag on a value that is *not* that session's key. Under prefix-parsing the bridge would have stored `<uuid>+recreate-close` in `session` as a near-duplicate of the real key, inflating distinct-session counts with fabricated rows. It is now re-tagged `synthetic:` (routing only requires that the id be *different*, which it still is). This is the only behavioural change to `browser`; the rest of the CLI diff is comments.

### Contracts

- **Best-effort intact.** Everything added is inside `emit_cmd_event`'s swallowing `try`, off the critical path. Proved by a test that runs all four header shapes against a raising emitter and asserts 200.
- **`kind='cmd'` untouched.**
- **Privacy comment rewritten.** The module block said "METADATA ONLY … never page content", which is now false as written. It states what is emitted and why an agent-minted session handle is not page content: it is not derived from any page, it says *who asked*, never *what was browsed* — and it is the same value `source='claude'` rows already store raw.
- **Server-originated events** (heartbeat, `/whoami`, `/health`) gain nothing: `attribute_session` defaults to `False`, so their records are byte-identical.

### Throttle-path `sess` hash — decision: **KEPT**

`_session_hash` stays. The `session` column is filled only for the `claude` tier and only for non-nested calls, so a flood driven from a `tmux`/`sid`/`ppid`/untagged id, or from a nested `browser agent` run, is exactly the case where you still need *some* stable handle to tell one flooder from two. It is 8 hex, on the throttle path only. Both now appear on a joinable-tier throttle event and a test pins that.

The pre-existing test asserting *"raw session id NEVER stored"* was **narrowed, not deleted**: it sends an untagged id, so its invariant is the fail-closed one. The comment now says so, and the joinable case is asserted separately.

## Verification

Base = `da33356` (`origin/main`; `scripts/browser-bridge/` is byte-identical between it and the `85cf466` I branched from). Base runs used a `git archive` of `origin/main` with only the three test files copied in, under `PYTHONDONTWRITEBYTECODE=1`.

### Red at base → green at HEAD

`tests/test_server.py` — **25 red at base, all green at HEAD**:

| test | base failure |
|---|---|
| `test_session_column_is_filled_for_the_joinable_tier_only[claude\|tmux\|sid\|ppid\|synthetic]` (5) | `KeyError: 'sess_src'` |
| `test_the_session_column_carries_the_bare_id_not_the_wire_tag` | `KeyError: 'session'` |
| `test_only_the_first_colon_splits_a_multi_colon_id` | `KeyError: 'sess_src'` |
| `test_an_absent_or_empty_id_fails_closed[None\|""]` (2) | `KeyError: 'sess_src'` |
| `test_an_untagged_id_is_never_promoted_to_a_join_key` | `KeyError: 'sess_src'` |
| `test_a_tag_the_server_does_not_know_is_recorded_but_not_joined` | `KeyError: 'sess_src'` |
| `test_an_ordinary_request_sets_session_and_declares_no_origin` | `KeyError: 'session'` |
| `test_the_release_short_circuit_attributes_an_ordinary_session` | `KeyError: 'session'` |
| `test_the_throttle_path_carries_both_the_hash_and_the_join_key` | `KeyError: 'sess_src'` |
| `test_an_oversized_id_is_dropped_whole_never_truncated` | positive control finds no key |
| `test_extra_cannot_overwrite_the_attribution` | `TypeError` (no such kwarg) |
| `test_a_nested_run_records_the_forwarded_id_as_origin_not_as_session` | `AttributeError: HDR_SESSION_ORIGIN` |
| `test_an_origin_wins_over_the_joinable_tier_on_every_emit_site` | `AttributeError: HDR_SESSION_ORIGIN` |
| `test_a_raising_emitter_still_lets_the_command_succeed` | `AttributeError: HDR_SESSION_ORIGIN` |
| `test_a_control_character_id_is_dropped_at_the_unit[tab\|nl\|nul\|del]` (4) | `AttributeError: _clean_session_field` |
| `test_cmd_ok_emits_one_metadata_event` *(existing, expectation updated)* | payload gained `sess_src` |
| `test_cmd_rate_limited_returns_429_and_emits_throttle` *(existing, narrowed)* | `KeyError: 'sess_src'` |

`tests/test_browser_session_id.py` — **3 red at base**: `test_the_cli_tier_tag_vocabulary_is_pinned`, `test_the_servers_joinable_tag_is_one_the_cli_actually_emits` (`AttributeError: SESSION_SRC_JOINABLE`), `test_the_throwaway_recreate_close_id_is_re_tagged_not_just_suffixed` (`throwaway id wears tier tag '${SESSION_ID}+recreate-close'`).

`tests/browser_tool.test.mjs` — **2 red at base** (`85 pass / 2 fail` at base, `87 pass / 0 fail` at HEAD).

**Honest note on three of those reds.** The three `AttributeError: HDR_SESSION_ORIGIN` rows are red at base because the constant does not exist there, not because base *behaves* differently on a path the test could otherwise reach — the header is new. Counted as coverage of new behaviour, not as evidence base was buggy on an existing path.

### Green at base — labelled INVARIANT GUARDS, not counted as regression coverage

- `test_server_originated_events_carry_no_session_fields` — pins that the heartbeat/diagnostics gain nothing. Proved **reachable** by mutant M7 below, not just green.
- `test_the_tagged_id_survives_the_transport[claude|tmux]` and `test_an_ordinary_cli_call_declares_no_nested_origin` — the wire pins. `server.py` now *depends* on the tag reaching it and on the direct CLI never declaring an origin; nothing else asserted that the tag survives the transport rather than only the deriver. Labelled as invariant guards in the file.

### Mutation results

Sandbox copy of the tree, `PYTHONDONTWRITEBYTECODE=1` plus an explicit `__pycache__` purge before every run (a same-length edit in the same second is invisible to CPython's mtime+size cache and would score SURVIVED without executing — M5 is exactly that shape, length delta **0**, and it was killed). **Positive control first: pristine → 26/26 green**, so no "killed" verdict comes from an already-red suite. Each mutation is the narrowest expression that can be wrong; no guard was deleted together with its enclosing condition.

| # | mutation | killed by | with its own assertion |
|---|---|---|---|
| M1 | `tier == SESSION_SRC_JOINABLE` → `tier in (SESSION_SRC_JOINABLE, "tmux")` | `..._joinable_tier_only[tmux]` | `tier 'tmux' must NOT produce a join key; got '%77'` |
| M2 | `origin = _clean_session_field(session_origin)` → `origin = ""` | `test_a_nested_run_records_the_forwarded_id_as_origin...` | `a nested run must not claim the session: {…'session': '8ab3d704-…'}` |
| M3 | `rec["session"] = bare` → `= _clean_session_field(session_id)` | `..._carries_the_bare_id_not_the_wire_tag` | `assert 'claude:6f1c9…' == '6f1c9a20-…'` |
| M4 | `s.partition(":")` → `s.rpartition(":")` | `test_only_the_first_colon_splits_a_multi_colon_id` | `assert 'sid:424242' == 'sid'` |
| M5 | untagged returns `(SESSION_SRC_JOINABLE, s)` | `test_an_untagged_id_is_never_promoted_to_a_join_key` | `sess_src: 'claude'` on an untagged id |
| M6 | oversized returns `s[:MAX]` instead of `""` | `test_an_oversized_id_is_dropped_whole_never_truncated` | truncated key present on the row |
| M7 | `attribute_session: bool = False` → `True` | `test_server_originated_events_carry_no_session_fields` | attribution on a `kind=heartbeat` row |

M1 also killed `test_extra_cannot_overwrite_the_attribution`; M3 killed six; M4 killed four. Every mutant's *primary* killer is the guard's own assertion, quoted above.

Fixture ids are pairwise distinct and distinct from every constant the assertions name (`claude`, `tmux`, `unknown`, `browser-agent`), so a mutant hardcoding any of those literals cannot survive. Two different `claude:` uuids are used for the direct and nested cases so a mutant writing "whichever id it had" fails one of them.

### Gate (commit 1, `f47be59`)

Historical — the authoritative run for the
current head is at the bottom of this description. Both hermetic CI checks,
on this branch rebased onto `da33356`:

```
nix build .#checks.x86_64-linux.pytests   -> exit 0
  RESULT: all good
  TOTAL collected=11863  passed=11862  skipped=1  failed=0  (floor: 11039)
  scripts/browser-bridge/tests: 532 passed in 188.30s
  RESULT: PASS (exit=0)

nix build .#checks.x86_64-linux.nodetests -> exit 0
  PASS  scripts/browser-bridge/tests  (files=15 tests=497 pass=497 skipped=0 floor=490)
  TOTAL suites=4 files=33 tests=1118 pass=1118 fail=0 skipped=0  (floor: 1098)
  RESULT: PASS (exit=0)
```

`scripts/gate.sh --tier both --set all` run **outside** a flake shell returned node PASS but pytest FAIL — every one of those failures was a missing host tool or python dep in the ad-hoc shell (`logrotate` absent; `psycopg2` absent, so `mail-actions`/`signal`/`initiatives`/`scripts/tests` failed to collect). The same targets are green in the hermetic runs above (e.g. `initiatives` 784 passed there vs 9 failed in the ad-hoc shell), which is what makes that an environment verdict and not a code one. Nothing was skipped, disabled, or bypassed.

## Not verified

- **No live probe, by instruction.** Nothing here was exercised against the running bridge: the deployed CLI is an `mkOutOfStoreSymlink` onto the base clone and `server.py` is a `home.file` copy needing a `home-manager switch`, so a live observation would measure other code. **Deployed ≠ verified, and this is neither.**
- **No ClickHouse round-trip.** Tests assert the spool line; that `session` survives `collector.parse_line` into the column is inferred from `STRING_COLS` containing `session`, not measured.
- The `~11%` nested-call figure and the `0/6,937` count are taken from the brief; I did not re-measure them.
- The `synthetic:` retag is pinned **structurally** (the throwaway id's tag), not by driving a full `emulate --recreate` round-trip.

---

# Follow-up commit `38a12b5` — hazard 3, redone as an origin header

## The leak

opencode sets `process.env.OPENCODE="1"` in a yargs **top-level `.middleware()`** — registered before every `.command(...)`, so it runs for every subcommand — and hands its tool shells `{...process.env}`:

```
.middleware(async(D)=>{ … E3.start(),process.env.AGENT="1",
  process.env.OPENCODE="1",process.env.OPENCODE_PID=String(process.pid)})
  .usage("").completion(…).command(m1).command(q1)…
```

Instrument controls on that grep: the exact string matched once at a known byte offset, and a deliberately bogus control string (`OPENCODE_THIS_SHOULD_NOT_EXIST_XYZ`) correctly returned no match — so "found" is not an artifact of a grep that matches anything. Re-derived against the **pinned** build after the gate caught me using the wrong artifact (see below); identical there.

Combined with the reported live env dump — an opencode bash tool still carrying the outer Claude session's `CLAUDE_CODE_SESSION_ID` — that variable inside opencode names an **ancestor**, not the caller. I also hit the leak by accident while smoke-testing: a run with only `CLAUDE_SESSION_ID` set returned **my own agent shell's** `CLAUDE_CODE_SESSION_ID`. The variable really is ambient.

**Consequence:** a plain `opencode run …` whose bash tool shells out to `browser` sends a genuinely `claude:`-tagged id, and the server would credit the **outer** Claude session with browser usage it never performed — in the `session` JOIN column. (An earlier version of this description said "the column adoption-scan and the deadman read" — that was **false**; see the B2 correction below.) **There is nothing in the id to branch on:** it *is* the outer session's id, correctly tagged. The server cannot tell it from a direct call by inspection, so the caller has to say so.

## The fix

`84bf324` gave the opencode leak its own non-joinable tier (`oc-inherited:`). That is **superseded**: it answered a question this PR had already answered a different way. `browser agent` declares `X-Session-Origin` and leaves `X-Session-Id` alone; the opencode leak is the same question — *"this command was issued by something nested under the id on the wire"* — so it now gets the same answer.

```
X-Session-Id:     claude:<uuid>        # unchanged
X-Session-Origin: opencode-inherited   # declared beside it
```

**This closes the constraint I flagged and could not meet before: routing is byte-identical.** Tab ownership, `--tab`, `not_owned_tab` and the `$( … )` stability property are untouched because the id is untouched. `derive_session_id` is now byte-identical to `f47be59`.

`origin` becomes a two-value enum — `browser-agent` | `opencode-inherited` — both meaning "issued by something nested under `origin_session`", which is precisely true of both. **The server needed no change.** The tier, its vocabulary-ledger entry, and the hyphen widening its parser needed are all deleted.

Net against `84bf324`: **355 insertions, 196 deletions** — smaller, and reusing an already-tested path.

The origin is declared only when a Claude id was *actually inherited*: the condition reads the **derived id** (`claude:*`) rather than re-testing env vars, so it cannot drift out of step with the precedence chain. Your interactive-opencode case (`OPENCODE=1`, no Claude ancestor → `tmux:` tier, no header at all) is covered by its own test.

## A real gate failure the previous round's hermetic run caught

`84bf324`'s hermetic pytest run came back **FAIL (1 test)** — `scripts/tests/test_opencode_engine.py`. Worth noting how: the background wrapper reported `exit code 0` while nix printed `Cannot build …devrc-pytests.drv`. Reading the content rather than the status is what surfaced it.

The cause was substantive, not cosmetic. That file pins opencode's version across an enumerated surface, and my README cited **1.18.18** — which is **not the pinned build**. The pin is **1.18.16**; 1.18.18 is what this host's imperative nix profile has on `PATH`.

So my evidence had been gathered from the wrong artifact. I re-derived it against the pinned bundle: the `AGENT=1, OPENCODE=1, OPENCODE_PID` middleware sequence is present and **byte-identical** there. The claim holds at both points — the gate turned a single-point measurement into a two-point one.

Both the README and the CLI comment now **name `PINNED_VERSION` instead of copying its value** (the idiom that file already uses), so neither can go stale.

I also added `scripts/browser-bridge/browser` to `PIN_SURFACE`. It now carries a live structural claim about opencode's environment, and that file's own rule is explicit: *"the surface only rots downward — a NEW file carrying a live claim is simply invisible… nothing can detect that omission for you."* Flagging it as a deliberate, gate-instructed edit to a file outside this PR's core.

**Separately, and NOT fixed here:** `opencode --version` on this host is **1.18.18** while the flake pins **1.18.16**, so `test_engine_is_the_version_every_measurement_is_keyed_to` fails *in an interactive shell* — it passes hermetically, which is why CI never saw it. Both operands (`PINNED_VERSION`, the host binary) are identical at `84bf324` and at `origin/main`, so this is **pre-existing host drift, not from this PR**. The test's own message prescribes the fix (`nix profile remove opencode`, then `home-manager switch`) — a host mutation, so yours to make, not mine.

## Verification — the gate was reset, not carried over

Baseline for this round is **`84bf324`** (the PR's third; `da33356` = `origin/main`, `f47be59` = commit 1).

### Red at `84bf324` → green at HEAD — 6 tests

| test | failure at `84bf324` |
|---|---|
| `test_routing_equivalence_opencode_does_not_change_the_id[claude:uuid-primary]` | `OPENCODE must not perturb the routing id` |
| `…[claude:uuid-alternate]` | same |
| `test_an_inherited_claude_id_declares_the_opencode_origin[CLAUDE_CODE_SESSION_ID]` | no `X-Session-Origin` on the wire |
| `…[CLAUDE_SESSION_ID]` | same |
| `test_the_discriminating_pair_on_the_wire` | `assert 'oc-inherited:uuid-pair' == 'claude:uuid-pair'` |
| `test_the_cli_declares_the_token_the_server_tests_pin` | `the origin-token parser matched nothing` |

Note the routing-equivalence table's **`tmux:` row is green at `84bf324`** — the old re-tagging only ever fired on a claude id. Baselines are labelled **per row** in the test's docstring rather than per test.

### Green at `84bf324` — INVARIANT GUARDS

Labelled in-file with the mutant that proves each reachable. All **five server-side tests** are here: the server needed no change, so they pin that the origin path covers the new token and composes with the tier gate — not regression coverage. Plus `…declares_nothing` (N3), `…empty_opencode…` (N1/N2/N9), `…derived_posix_tier` (N6).

### The routing-equivalence test — your requested assertion

`test_routing_equivalence_opencode_does_not_change_the_id` derives each id **twice**, differing only in `OPENCODE`, and requires both to equal the **same pinned literal**. Two load-bearing halves: the arms must equal *each other* (OPENCODE changes nothing) **and** equal a literal taken from the contract, so a change moving both arms together still fails. `84bf324`'s design dies on both halves — that is mutant N6.

### Mutation results — carrying forward the M10 standard

Fresh sandbox, `PYTHONDONTWRITEBYTECODE=1` + `__pycache__` purge per run.

**Positive control, stated with its numbers:** pristine → **37 passed** (CLI file) and **14 passed** (server slice), zero failures. My first battery run printed *nothing* for the control because the grep missed the summary line — a silent zero indistinguishable from a harness wired to nothing. I fixed the filter and re-ran rather than reporting the kills off it.

| # | mutation (narrowest expression) | killed | by |
|---|---|---|---|
| N1 | `[ -n "${OPENCODE:-}" ]` → `true` | 3 | the "no OPENCODE ⇒ no origin" claim at 3 layers |
| N2 | `-n` → `-z` | 5 | both directions |
| **N3** | case `claude:*` → `*` | **exactly 1** | `…no_claude_ancestor_declares_nothing`, its own assertion |
| N4 | token → `browser-agent` | 4 | incl. the seam ledger |
| N5 | the `args+=` header line → `:` | 3 | wire tests only |
| N6 | **re-tag the id instead** (the rejected design) | 3 | incl. **routing equivalence** |
| **N9** | `${OPENCODE:-}` → `${OPENCODE+set}` (empty-string boundary only) | **exactly 1** | `…empty_opencode_is_not_opencode`, its own assertion |
| N7 | server: `origin = ""` | 5 | every origin test |
| N8 | server: `payload["origin"] = "browser-agent"` | 2 | incl. its own distinctness test |

You asked me to carry forward the standard that M8 missed and M10 met. **N3 and N9 are exact-1 killers** — I added N9 specifically to give the empty-string boundary its own surgical mutant. N1/N2/N4/N5/N7 kill several, but every kill is the *same claim* observed at a different layer (id / wire / both spellings), not an unrelated neighbour; I am not claiming isolation I did not get.

### Gate

Both hermetic CI checks, on commit `38a12b5`. Superseded by `5cc0731` below,
whose verification is targeted (the branch is deliberately not rebased, so the
full hermetic gate is the coordinator's to run on the merged tree). Read from the
log CONTENT, not an exit code — the background wrapper reported `exit code 0` for the
`84bf324` run that had actually failed, which is exactly why these are quoted:

```
  RESULT: all good
  TOTAL collected=11877  passed=11876  skipped=1  failed=0  (floor: 11039)
  RESULT: PASS (exit=0)

  TOTAL suites=4 files=33 tests=1118 pass=1118 fail=0 skipped=0  (floor: 1098)
  RESULT: PASS (exit=0)
```

Nothing was skipped, disabled, or bypassed. The single `skipped=1` is pre-existing
and pinned in the suite's own `EXPECTED_SKIPS`.

## Still not verified

Everything previously listed still stands. Unchanged and explicitly not addressed per your instruction: no live `opencode run` exercise, and the guard silently not firing if `OPENCODE` were ever unset on some real path.

---

# Follow-up commit `5cc0731` — six blind-audit findings

The audit came back safe-to-merge, and independently measured the join premise I could only infer: this shell's `CLAUDE_CODE_SESSION_ID` matches **34 live `source='claude'` rows**, and all 406 claude sessions in 14d are uuid-shaped. Storing the bare id is measured-correct, not assumed-correct.

Six findings, all closed.

## A — the origin path was more permissive than the id path beside it

All four are one shape: the guard against fabricated attribution had holes the *id* guard did not.

**A1 + A2 — one fix.** Suppression branched on `if origin:` — the **cleaned** value — so an oversized or control-char origin cleaned to `""`, fell through to the `elif`, and wrote **`session` with the parent's id**. That is the exact fabrication this mechanism exists to prevent, arriving *through the mechanism itself*. And any string was accepted: `totally-made-up`, `false`, `   ` were all recorded verbatim and all suppressed, so the "two-value enum" was documentation, not a contract.

Now: **branch on header presence** (`session_origin is None`), validate the value against `ORIGIN_TOKENS`, and record `invalid` for anything else so the case is visible in the data rather than silent. Presence decides suppression; the value decides only what gets recorded. The handler's `or None` is gone too — it collapsed a *present-but-empty* header into "absent" and re-attributed the call.

**A3 — reachable, not theoretical.** `origin_session` bypassed the tier gate. `browser-agent` forwards whatever `--print-session-id` produced and the opencode tool declares its origin unconditionally, so `tmux:%3` genuinely arrived and was recorded as `origin_session='%3'`. Same gate now, same rationale as `session`: a pane id is stable across many unrelated sessions, and `origin_session` is a column a reader can `GROUP BY`. The tier is still recorded, so the suppressed population stays measurable.

**A4 — `sess_src` was unbounded-cardinality.** `['claude`, `CLAUDE`, `claud` all landed verbatim; four of the measured values are near-misses of the joinable tag that a reader skimming a `GROUP BY` would misread. Validated against `SESSION_TIER_TAGS`, pinned two-way against the CLI's vocabulary.

## B — three claims the code contradicted

**B1 and B2 are also posted as a [PR comment](https://github.com/innovation-upstream/devrc/pull/549#issuecomment-5338877940)**, because anyone who read this description earlier read the wrong version.

**B1.** The flagship example query returned **zero rows forever** — disjoint id spaces (406 claude uuids vs 183 `ses_` opencode ids, overlap 0) *and* the opencode guard suppresses `session` for exactly the calls it was about. It failed as a silent empty result. **Replaced, not annotated**, with queries answerable today plus an explicit "not answerable yet" section naming the follow-up work.

**B2.** "adoption-scan and the deadman read this column" was **false**, in four places, in the present tense. Verified independently: adoption-scan's browser-bridge entry is `via="source"` and never selects `session`; the deadman reads row existence. **No shipped consumer reads `session` for this source today** — because it has always been empty. The harm is **latent**: it lands on the first consumer that does. Still the reason for every gate, but a weaker claim than the one I made. (`kind='cmd'` *is* read by adoption-scan — that part was true and is untouched.)

**B3.** `/whoami` and `/health` were mislabelled server-originated. They are operator subcommands and the CLI sends its session headers on them because `_curl` is one code path. **Made consistent rather than relabelled:** they now attribute like any other command. The old exclusion gave one operation two outcomes — `whoami` via POST `/cmd` attributed, the same `whoami` via GET not — for 125 rows / 2.0% of `kind='cmd'` over 14d, a smaller copy of the bug this PR fixes. Only the heartbeat is server-originated now, and the test's category has exactly one member.

## Verification

Baseline **`38a12b5`** (the PR's fourth commit).

### Red at `38a12b5` → green at HEAD — 25 tests

All 25 selected tests failed at base, and the failure text reproduces the audit's measurements exactly:

| finding | representative base failure |
|---|---|
| A1 | `oversized: a present origin must suppress the key; got 'c05e17b6-…'` (also `empty`, `whitespace`) |
| A2 | unknown/`false`/`BROWSER-AGENT` tokens recorded verbatim |
| A3 | `tier 'sid' is not joinable and must not be recorded as a parent key; got '424242:99887766'` (also `tmux`, `ppid`, `synthetic`) |
| A4 | `sess_src` = `futuretier` / `CLAUDE` / `claud` / `['claude` |
| B3 | the diagnostic GETs carried no session; the orientation payload lacked `sess_src` |

**Green at `38a12b5` — 1 test, labelled INVARIANT GUARD:** `test_only_the_heartbeat_is_server_originated`. Proved reachable by mutant P5.

### Mutation results — 7 mutants, all killed

Fresh sandbox, `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged between mutants. **Positive control: 47 passed / 0 failed**, and the number is asserted against a *measured* `EXPECT_SELECTED`, not a guess.

| # | mutation (narrowest expression) | killed | notes |
|---|---|---|---|
| P1 | `if session_origin is not None:` → `if _clean_session_field(...)` | 4 | the A1 bug itself |
| P2 | ledger check dropped — record `declared` raw | 8 | distinguishes A2 from A1: `session` stays suppressed, only `origin` is wrong |
| P3 | **one conjunct** removed: `tier == JOINABLE and bare` → `bare` | 4 | exactly `…origin_session_passes_the_same_tier_gate…`. Deleting the whole condition would also drop the `bare` test and prove nothing about the tier half — mutating a conjunct instead, per your note |
| P4 | tier vocabulary check deleted | 6 | exactly the A4 tests |
| P5 | `attribute_session=True` → `False` in the diag emitter | 4 | exactly the B3 tests |
| P6 | handler `or None` restored | 3 | a **different site** from P1, reachable only over the wire |
| P7 | `origin_session` never recorded | 13 | added *because* the filter control found its killer unselectable — see below |

🔴 **The filter guard fired, and it caught a real gap.** My battery asserts the control's selected count and, separately, that every new/changed test is selectable. The second check failed: `test_a_joinable_parent_is_still_recorded` — the control for the A3 gate — was **not** selected by my `-k` filter, so a mutant deleting `origin_session` outright could have reported SURVIVED. Exactly the false-SURVIVED shape you warned about, in my own harness. I widened the filter, re-measured `EXPECT_SELECTED` (47, not the 57 I had guessed), and added **P7** to confirm that test does the killing. It does.

### Honest note on one test row

Over HTTP, the `whitespace` origin row is **not** a distinct case from `empty` — a header value is delimited by the colon-space and trailing whitespace is stripped, so `X-Session-Origin:    ` arrives as `""`. Measured via P6, which kills both rows identically. The row is kept because it documents what a caller writing whitespace actually gets, not because it exercises a separate branch. Likewise `" claude"` as a tier tag cannot travel, so it is tested at the unit level with that stated.

### Targeted runs (no full hermetic gate, per instruction)

```
scripts/browser-bridge/tests/{test_server,test_browser_session_id,test_surface_parity,test_skill_size}.py
  419 passed
scripts/tests/{test_opencode_engine,test_no_captured_text,test_no_captured_markup}.py
  217 passed, 1 deselected
node --test scripts/browser-bridge/tests/browser_tool.test.mjs
  tests 87  pass 87  fail 0
```

The one deselected test is `test_engine_is_the_version_every_measurement_is_keyed_to` — the **pre-existing host drift** (`opencode` on PATH is 1.18.18, flake pins 1.18.16) documented in the previous round. Both its operands are identical at `38a12b5` and `origin/main`, so it is not from this PR; it passes hermetically.

## Still not verified

- No live probe, no live `opencode run`, no ClickHouse round-trip — unchanged, and yours after `ship.sh`.
- `origin='invalid'` rows are a **new** population in the data. Nothing has ever emitted one, so I cannot say what fraction of real traffic will land there; if a shipped caller is malformed in a way I have not imagined, it shows up as `invalid` rather than as a wrong join, which is the intended direction but is not the same as "measured zero".

---

# Follow-up commit `b17790f` — delta-audit items N1–N4

Delta re-audit of `38a12b5..5cc0731` came back safe-to-merge with all six earlier findings confirmed fixed. Four items remained; none change runtime behaviour.

## 🔴 N2 — the "two-way pin" was fictional, and the next planned change is what it fails to catch

`server.py` claimed `SESSION_TIER_TAGS` was *"pinned two-way against the CLI by `tests/test_browser_session_id.py`"*. **That symbol appeared nowhere in that file.** The server set was checked against a literal retyped in `test_server.py`; the CLI against a *separate* literal in `test_browser_session_id.py`; nothing compared the two.

Measured: grow the CLI by an `opencode:` tier **and** update `CLI_TIER_TAGS`, leave `server.py` alone → **400 passed, 0 failed. SURVIVED.**

The harm lands on the **next** change, not this one. `browser`'s own FOLLOW-UP SLOT says an `opencode:` tier is coming — the very follow-up this PR exists to enable. Without a real pin the server normalises the new tag to `unknown`, drops the id, and the column silently re-empties. A fail-closed silent regression is exactly what every gate here targets, reintroduced *in the gate itself*.

**Fixed where the parser already lives.** `test_browser_session_id.py` now compares `set(S.SESSION_TIER_TAGS)` against `_emitted_tags()` — tags **parsed out of the CLI source**, not a literal beside them.

The retyped literal in `test_server.py` is **deleted**, not kept alongside. It read as a cross-file pin and was not one, and *a literal that can be updated in lockstep with the thing it checks is worse than no check, because it reads as one*. What remains there is the self-consistency half (the joinable tag is in the set; the fallback marker is not), pointing at the real pin by name.

**Verification — the auditor's exact mutant shape, re-run:**

| | before | after |
|---|---|---|
| grow CLI + `CLI_TIER_TAGS`, leave `server.py` | **SURVIVED** — 400 passed, 0 failed | **KILLED** — 1 failed, 400 passed |

Killed by **exactly one test**, `test_the_server_validation_set_equals_the_tags_parsed_from_the_cli`:

```
server.py SESSION_TIER_TAGS=['claude','ppid','sid','synthetic','tmux'] does not
match the tags the CLI can actually emit ['claude','opencode','ppid','sid',
'synthetic','tmux']. A tier present in one and not the other is silently
normalised to 'unknown' and loses its id.
```

## 🟡 N1 — `emit_cmd_event`'s docstring still described pre-B3 behaviour

It said `attribute_session=False` covers *"the heartbeat, and the /whoami+/health diagnostics"* and that their records are *"byte-identical to before"*. B3 moved those two to `True` in the previous commit, so both sentences were false while `_emit_diag_event` 90 lines below, `README:1493` and the test all said the opposite.

Worth naming plainly: this is **the same shape as B2 — a false comment about who writes what — reintroduced by the round that fixed B2.** Nothing gates it; #553's doc-rot corpus is `CORPUS_DIRS = ("claude", "CLAUDE.md")` and excludes every file this PR touches.

## 🟡 N3 — "Answerable **now**" returned zero rows

Measured **0 of 6,166** browser-bridge `kind='cmd'` rows in 14d, because `session` is filled by code that is not deployed. B1's entire finding was a flagship query returning nothing; a replacement that also returns nothing — under a heading contrasting "now" with "NOT answerable yet" — repeats it in miniature. Reworded to **"Answerable once this ships"**, with *merged ≠ deployed* stated inline. That is this repo's standing hazard and it belongs in its own docs.

## 🟢 N4 — doubled phrase

The B2 reword left *"in the exact column `session` JOIN column"*. Fixed.

## Verification

Whole-file runs, **no `-x`** (the auditor's first pass used it and first-failure-wins mis-attributed two mutants to the flaky test below, producing a false KILL that would have hidden N2 entirely — not repeated):

```
test_server.py + test_browser_session_id.py + test_surface_parity.py + test_skill_size.py
  420 passed
test_no_captured_text.py + test_no_captured_markup.py + test_rules_size.py + test_opencode_engine.py
  222 passed, 1 deselected
```

The deselected test is the pre-existing 1.18.18-vs-1.18.16 host drift documented earlier. No hermetic gate run on this branch, per instruction.

## Recorded, deliberately not fixed here

- **C1 control characters.** `_clean_session_field` rejects C0 and DEL but not C1 (0x80–0x9F), so `claude:a\x9db` is accepted and a C1 char can reach the `session` column. Pre-existing, outside this delta.
- **A load-flaky test.** `test_the_throttle_path_carries_both_the_hash_and_the_join_key` failed 2 of 12 whole-file runs on an identical tree — same code, two outcomes. Its `_wait_events(..., 2, timeout=3.0)` budget is too tight for two off-critical-path spool writes under load. Present at `38a12b5`. It passed in every run of mine this round, including in isolation, which is consistent with load-dependence rather than a defect in the change. **This one is a genuine "fix it rather than re-run it" candidate** — the timing dependency is removable — and I have left it alone only because it was scoped out; say the word and it is a one-line follow-up.
- **N8.** `test_only_the_heartbeat_is_server_originated` asserts the heartbeat's own fields; the universal "only" rests on a sibling test rather than a grows-or-shrinks ledger.

### DIFF
--- a/scripts/browser-bridge/README.md
+++ b/scripts/browser-bridge/README.md
@@ -1411,7 +1411,7 @@ as `X-Session-Id`. Right for routing and the audit trail; **wrong** for the
 `session` column, which means *the agent session that issued this command*. Left
 alone, one `browser agent "<goal>"` call would become N browser calls credited to
 the operator's own session — a fabricated usage number in the exact column
-`adoption-scan.py` reads (~581 nested rows in 14d, ~11% of bridge commands).
+`session` JOIN column (~581 nested rows in 14d, ~11% of bridge commands).
 
 So the nested tool declares `X-Session-Origin: browser-agent`. When that header is
 present the server writes **no** `session`; the forwarded id is recorded as
@@ -1435,7 +1435,7 @@ assignment is therefore not specific to whichever of the two you land on.
 So inside opencode that variable names an **ancestor**, not the session issuing
 the command. A plain `opencode run …` whose bash tool shells out to `browser`
 would otherwise have the bridge credit the **outer** Claude session with browser
-usage it never performed — in the column `adoption-scan` and the deadman read.
+usage it never performed.
 
 **🔴 The id is indistinguishable from a direct call.** There is nothing in it to
 branch on: it *is* the outer session's id, correctly tagged `claude:`. So the
@@ -1490,8 +1490,25 @@ source.
   zero rows, which reads as a valid "no sessions matched" answer.
 - An id over 200 chars or carrying a control character is **dropped whole**, never
   truncated — a truncated join key is a *wrong* join key.
-- Server-originated events (the heartbeat, `/whoami`, `/health`) carry **none** of
-  these fields: they have no caller session to attribute.
+- **Only the heartbeat** is server-originated and carries none of these fields.
+  `/whoami` and `/health` are **operator** calls — `browser whoami` / `browser
+  health` are subcommands a person runs, and the CLI sends its ordinary session
+  headers on them because `_curl` is one code path — so they are attributed like
+  any other command. They were previously excluded on the stated grounds of
+  having "no caller session", which was false for these two and made ONE
+  operation have TWO outcomes: `whoami` via POST `/cmd` was attributed, the same
+  `whoami` via GET was not (125 rows, 2.0% of `kind='cmd'` over 14d).
+- Both header vocabularies are **validated closed sets**, not documentation. A
+  `sess_src` tag outside the tier list becomes `unknown` and carries no id; an
+  `origin` outside the two tokens is recorded as `invalid`. Both arrive on
+  caller-supplied headers, so without this each is an unbounded-cardinality
+  column.
+- **Origin suppression keys off header PRESENCE, never the value.** Any present
+  value — including empty, oversized or control-char — suppresses `session`.
+  Losing attribution beats fabricating it.
+- `origin_session` passes the **same tier gate** as `session`: a non-joinable
+  parent id (`tmux:`/`sid:`/`ppid:`) is not recorded, because a reader grouping by
+  it would merge unrelated sessions exactly as they would on `session`.
 - `kind='cmd'` is unchanged — it is the usage signal `adoption-scan.py` reads.
 - Best-effort is unchanged: every field above is written inside `emit_cmd_event`'s
   swallowing `try`, off the critical path.
@@ -1501,14 +1518,57 @@ one: the agent session's own opaque handle is not page content, is not derived
 from anything the browser saw, and is minted by the local agent harness before
 any browser command exists. It says *who asked*, never *what was browsed*.
 
-**Example — sessions that used both opencode and the browser skill:**
+### What this makes answerable — and what it does not
+
+**Answerable now — which Claude sessions used the browser skill, and how much:**
 
 ```sql
-SELECT session, count() AS browser_calls
+SELECT session, count() AS browser_calls,
+       min(ts) AS first_call, max(ts) AS last_call
 FROM activity.events
 WHERE source = 'browser-bridge' AND kind = 'cmd' AND session != ''
-  AND session IN (SELECT DISTINCT session FROM activity.events WHERE source = 'opencode')
-GROUP BY session ORDER BY browser_calls DESC
+GROUP BY session
+ORDER BY browser_calls DESC
+```
+
+`session` holds the bare Claude session uuid, so it joins straight against
+`source='claude'` rows — same column, same values, no transform:
+
+```sql
+SELECT count(DISTINCT session) AS joinable_browser_sessions
+FROM activity.events
+WHERE source = 'browser-bridge' AND kind = 'cmd' AND session != ''
+  AND session IN (SELECT DISTINCT session FROM activity.events WHERE source = 'claude')
+```
+
+🔴 **NOT answerable yet — "which sessions used opencode AND the browser skill".**
+That was this change's motivating question and it is still open, for two
+independent reasons, either of which alone is fatal:
+
+1. **The id spaces are disjoint.** `source='opencode'` sessions are `ses_`-prefixed
+   opencode ids; `source='claude'` sessions are uuids. Measured over 14 days: 406
+   distinct claude sessions, 183 distinct opencode sessions, **overlap 0**. A query
+   joining `browser-bridge.session` against opencode's `session` cannot match — and
+   it fails as a **silent empty result**, which reads exactly like a truthful
+   "no sessions matched".
+2. **Those calls deliberately carry no `session`.** A browser call issued from
+   inside opencode is the leak case above: it is marked with
+   `origin='opencode-inherited'` and attributed to nobody.
+
+An earlier draft of this README shipped exactly that broken join as its flagship
+example. It is replaced rather than annotated, because a wrong query that returns
+zero rows is worse than no query.
+
+**What would make it answerable:** the follow-up `shell.env` hook exporting
+`OPENCODE_SESSION_ID`, which gives nested runs a joinable `opencode:` tier of
+their own. Until then, use `origin='opencode-inherited'` to *count* opencode-driven
+browser usage — you can see how much there is, just not whose:
+
+```sql
+SELECT JSONExtractString(toString(payload), 'origin') AS origin, count()
+FROM activity.events
+WHERE source = 'browser-bridge' AND kind = 'cmd'
+GROUP BY origin ORDER BY count() DESC
 ```
 
 ## Session isolation (concurrent-session tab clobbering)
--- a/scripts/browser-bridge/README.md
+++ b/scripts/browser-bridge/README.md
@@ -1397,6 +1397,7 @@ would be, and is forbidden.
 | `sid:<sid>:<start>` | `sid` | no | no other source records it |
 | `ppid:<pid>:<rand>` | `ppid` | no | last-resort random token |
 | `synthetic:…` | `synthetic` | no | an id the CLI made up on purpose (the `emulate --recreate` close) |
+| `oc-inherited:<uuid>` | `oc-inherited` | no | a `CLAUDE_CODE_SESSION_ID` that **leaked into an opencode tool shell** — it names an ancestor, not the caller (see below) |
 | anything untagged, absent, or empty | `unknown` | no | fail closed |
 
 Only the **first** colon splits (a `sid:`/`ppid:` id contains more). An id with no
@@ -1417,6 +1418,49 @@ So the nested tool declares `X-Session-Origin: browser-agent`. When that header 
 present the server writes **no** `session`; the forwarded id is recorded as
 `payload.origin_session` (the causal **parent**) beside `payload.origin`. Giving
 the nested session an id of its own is a later change.
+
+### Hazard 3 — `CLAUDE_CODE_SESSION_ID` leaks into opencode
+
+opencode 1.18.18 sets `process.env.OPENCODE="1"` in a yargs **top-level
+`.middleware()`** — registered before every `.command(...)`, so it runs for every
+subcommand — and hands its tool shells `{...process.env}`. Launch opencode from a
+Claude Code bash call and the outer session's `CLAUDE_CODE_SESSION_ID` rides all
+the way down: a live env dump from inside an opencode bash tool still had it.
+
+So inside opencode that variable names an **ancestor**, not the session issuing
+the command. A plain `opencode run …` whose bash tool shells out to `browser`
+would otherwise hit the joinable tier on the leaked value, and the server would
+write the outer Claude session's uuid into `session` — where those rows are
+**indistinguishable from that session's own direct browser use**. There is no
+`X-Session-Origin` on that path; that header is only sent by
+`browser_tool_impl.mjs` on the browser-agent path.
+
+`derive_session_id` therefore **re-tags** a leaked id `oc-inherited:` whenever
+`OPENCODE` is set. Re-tagging, not re-deriving, is the point:
+
+- the underlying value is unchanged, so the id stays **stable** across the many
+  `browser` calls one opencode session makes and per-session tab ownership still
+  works (asserted through the same `$( … )` subshell probe as the joinable path);
+- the tier is not joinable, so `session` stays **empty** — exactly as it is today
+  for these rows. Empty is recoverable; plausible-but-wrong is not;
+- `payload.sess_src = "oc-inherited"` still records that it happened, so the
+  population is **measurable** rather than invisible.
+
+Two consequences, stated rather than buried:
+
+- **Unchanged degradation.** Two concurrent opencode sessions launched from one
+  Claude session still resolve to the same id and share a tab. That is today's
+  behaviour, not a new one; `--tab` is the escape hatch.
+- **Routing does change in one place.** An opencode-inner call no longer shares
+  tab ownership with its *outer* Claude session (different id ⇒ different owner),
+  so it falls back to the active tab rather than the outer session's owned tab.
+  That separation is inherent to any re-tagging fix, and is the intended
+  isolation.
+
+**This guard is temporary.** A follow-up adds an opencode `shell.env` plugin hook
+exporting `OPENCODE_SESSION_ID`; once nested runs have a key of their own,
+`derive_session_id` gains a joinable `opencode:` tier above the guard and the
+guard retires. The slot is marked in the source.
 
 ### Contract
 
--- a/scripts/browser-bridge/README.md
+++ b/scripts/browser-bridge/README.md
@@ -1397,7 +1397,6 @@ would be, and is forbidden.
 | `sid:<sid>:<start>` | `sid` | no | no other source records it |
 | `ppid:<pid>:<rand>` | `ppid` | no | last-resort random token |
 | `synthetic:…` | `synthetic` | no | an id the CLI made up on purpose (the `emulate --recreate` close) |
-| `oc-inherited:<uuid>` | `oc-inherited` | no | a `CLAUDE_CODE_SESSION_ID` that **leaked into an opencode tool shell** — it names an ancestor, not the caller (see below) |
 | anything untagged, absent, or empty | `unknown` | no | fail closed |
 
 Only the **first** colon splits (a `sid:`/`ppid:` id contains more). An id with no
@@ -1421,46 +1420,64 @@ the nested session an id of its own is a later change.
 
 ### Hazard 3 — `CLAUDE_CODE_SESSION_ID` leaks into opencode
 
-opencode 1.18.18 sets `process.env.OPENCODE="1"` in a yargs **top-level
-`.middleware()`** — registered before every `.command(...)`, so it runs for every
-subcommand — and hands its tool shells `{...process.env}`. Launch opencode from a
-Claude Code bash call and the outer session's `CLAUDE_CODE_SESSION_ID` rides all
-the way down: a live env dump from inside an opencode bash tool still had it.
+opencode sets `process.env.OPENCODE="1"` in a yargs **top-level `.middleware()`**
+— registered before every `.command(...)`, so it runs for every subcommand — and
+hands its tool shells `{...process.env}`. Launch opencode from a Claude Code bash
+call and the outer session's `CLAUDE_CODE_SESSION_ID` rides all the way down: a
+live env dump from inside an opencode bash tool still had it.
+
+Derived against the **pinned** build (`PINNED_VERSION` in
+`scripts/tests/test_opencode_engine.py`) — named rather than copied so it cannot
+go stale — and confirmed byte-identical in the newer build on this host's own
+profile, which is the one a real `opencode run` here actually executes. The
+assignment is therefore not specific to whichever of the two you land on.
 
 So inside opencode that variable names an **ancestor**, not the session issuing
 the command. A plain `opencode run …` whose bash tool shells out to `browser`
-would otherwise hit the joinable tier on the leaked value, and the server would
-write the outer Claude session's uuid into `session` — where those rows are
-**indistinguishable from that session's own direct browser use**. There is no
-`X-Session-Origin` on that path; that header is only sent by
-`browser_tool_impl.mjs` on the browser-agent path.
+would otherwise have the bridge credit the **outer** Claude session with browser
+usage it never performed — in the column `adoption-scan` and the deadman read.
 
-`derive_session_id` therefore **re-tags** a leaked id `oc-inherited:` whenever
-`OPENCODE` is set. Re-tagging, not re-deriving, is the point:
+**🔴 The id is indistinguishable from a direct call.** There is nothing in it to
+branch on: it *is* the outer session's id, correctly tagged `claude:`. So the
+server cannot tell the two apart by inspection — the caller has to say so.
 
-- the underlying value is unchanged, so the id stays **stable** across the many
-  `browser` calls one opencode session makes and per-session tab ownership still
-  works (asserted through the same `$( … )` subshell probe as the joinable path);
-- the tier is not joinable, so `session` stays **empty** — exactly as it is today
-  for these rows. Empty is recoverable; plausible-but-wrong is not;
-- `payload.sess_src = "oc-inherited"` still records that it happened, so the
-  population is **measurable** rather than invisible.
+That is the **same question** `browser agent` already answers (hazard 2), so it
+gets the **same mechanism**: `X-Session-Id` is left completely alone and the
+nested-run fact travels beside it.
 
-Two consequences, stated rather than buried:
+```
+X-Session-Id:     claude:<uuid>        # unchanged — routing, ownership, not_owned_tab
+X-Session-Origin: opencode-inherited   # this command was issued by something nested
+```
 
-- **Unchanged degradation.** Two concurrent opencode sessions launched from one
-  Claude session still resolve to the same id and share a tab. That is today's
-  behaviour, not a new one; `--tab` is the escape hatch.
-- **Routing does change in one place.** An opencode-inner call no longer shares
-  tab ownership with its *outer* Claude session (different id ⇒ different owner),
-  so it falls back to the active tab rather than the outer session's owned tab.
-  That separation is inherent to any re-tagging fix, and is the intended
-  isolation.
+`origin` is a two-value enum — `browser-agent` | `opencode-inherited` — both
+meaning "issued by something nested under `origin_session`", which is precisely
+true of both cases. The server needed **no change**: the origin path already
+suppresses `session` and records `payload.origin` + `payload.origin_session`.
 
-**This guard is temporary.** A follow-up adds an opencode `shell.env` plugin hook
-exporting `OPENCODE_SESSION_ID`; once nested runs have a key of their own,
-`derive_session_id` gains a joinable `opencode:` tier above the guard and the
-guard retires. The slot is marked in the source.
+Why this beats re-tagging the id, which was tried first and rejected:
+
+- **Routing is byte-identical.** Tab ownership, `--tab`, `not_owned_tab` and the
+  `$( … )` stability property are untouched, because the id is untouched. A
+  re-tagged id would have silently stopped an opencode-inner call from sharing
+  ownership with its outer session. The equivalence is now machine-checked: the
+  id is derived with and without `OPENCODE` and both must equal the same pinned
+  literal.
+- **One question, one mechanism.** Two ways to say "this is nested" inside one
+  system is a defect regardless of which is better in isolation.
+- It reuses a path that is already tested, rather than adding a parallel one.
+
+The origin is declared only when a Claude id was **actually inherited** — the
+condition reads the *derived* id (`claude:*`) rather than re-testing the env
+vars, so it cannot drift out of step with the precedence chain. An opencode
+session run interactively with no Claude ancestor derives `tmux:`/`sid:`/`ppid:`,
+declares nothing, and behaves exactly as it does today.
+
+**This is temporary.** A follow-up adds an opencode `shell.env` hook exporting
+`OPENCODE_SESSION_ID`; `derive_session_id` then gains a **joinable** `opencode:`
+tier, nested runs get a key of their own, the `claude:*` case stops matching on
+its own, and the token becomes dead code to delete. The slot is marked in the
+source.
 
 ### Contract
 
--- a/scripts/browser-bridge/README.md
+++ b/scripts/browser-bridge/README.md
@@ -1364,9 +1364,91 @@ unchanged.
 distinct `{"event":"throttled","key":…,"reason":…,"sess":…}` line AND emits a
 telemetry event (`outcome="throttled"`, `payload.reason`) into `activity.events`.
 `payload.sess` is a **coarse, non-reversible** fingerprint (first 8 hex of
-sha256 of the `X-Session-Id`) — enough to attribute a flood to a session without
-storing the raw id, kept metadata-only (no page content). This is the ONLY event
-that carries the session hash.
+sha256 of the `X-Session-Id`), kept metadata-only (no page content). This is the
+ONLY event that carries the session hash.
+
+`sess` was **kept** when the `session` join key landed (below) rather than
+replaced by it: the column is filled for the `claude:` tier and non-nested calls
+only, so a flood driven from a `tmux:`/`sid:`/`ppid:`/untagged id — or from a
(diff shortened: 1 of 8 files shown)


ZacxDev:
## ⚠️ Corrections to this PR's earlier description — two claims were wrong

A blind audit found six issues; four were code (fixed in `5cc0731`). **Two were false claims in this description and in shipped comments.** Posting them as a comment rather than only editing the body, because anyone who already read the description read the wrong version.

### 1. The example query I shipped returns zero rows — forever

The description's flagship example was:

```sql
-- WRONG. Do not use.
SELECT session, count() AS browser_calls
FROM activity.events
WHERE source = 'browser-bridge' AND kind = 'cmd' AND session != ''
  AND session IN (SELECT DISTINCT session FROM activity.events WHERE source = 'opencode')
GROUP BY session ORDER BY browser_calls DESC
```

It cannot ever match, for two independent reasons — either alone is fatal:

1. **The id spaces are disjoint.** `source='opencode'` sessions are `ses_`-prefixed opencode ids; `source='claude'` sessions (which is what `browser-bridge.session` now holds) are uuids. Measured over 14 days: **406 distinct claude sessions, 183 distinct opencode sessions, overlap 0.**
2. **Those calls deliberately carry no `session`.** A browser call issued from inside opencode is the leak case this PR guards; it is marked `origin='opencode-inherited'` and attributed to nobody.

So **the PR's own motivating question — "which sessions used opencode AND the browser skill" — is the one thing that query structurally cannot answer**, and it fails as a *silent empty result*, which reads exactly like a truthful "no sessions matched". That is the same failure mode this PR's raw-not-hashed decision exists to avoid, reintroduced in the documentation.

It is **replaced, not annotated** — a wrong query returning zero rows is worse than no query. The README now ships queries that are answerable today (browser calls per Claude session, and the join against `source='claude'` that does work), plus an explicit "not answerable yet" section naming the follow-up `shell.env` / `OPENCODE_SESSION_ID` work that would close it. In the meantime `GROUP BY origin` counts opencode-driven browser usage — you can see how much there is, just not whose.

### 2. "adoption-scan and the deadman read this column" is false

This appeared in the PR body, in `server.py` comments, in `browser`'s comments and in test docstrings, in the present tense. I verified it myself rather than taking the finding on trust:

- `adoption-scan.py`'s browser-bridge registry entry is `via="source"`; its query selects `text, payload, exit_code, duration_ms, ts, host` — **never `session`**, and `aggregate_source` never reads it.
- `deadman.py` mentions "session" only in prose comments, never the column; it consumes a row's **existence**.
- Every shipped `GROUP BY session` filters `source='claude'` (or claude+opencode).

**No shipped consumer reads `session` for `source='browser-bridge'` today** — precisely because it has always been empty. That makes the harm of a wrong value **latent, not live**: it corrupts the first consumer that reads it, silently, and a fabricated session key is indistinguishable from a real one after the fact. That is still the reason for every gate in this change, but it is a different and weaker claim than the one I made, and the stronger version was doing rhetorical work it hadn't earned.

Corrected in all four places. `kind='cmd'` **is** read by adoption-scan — that claim was and remains true, and is untouched.

### Also worth flagging from the same round

- **A1 was the sharpest code finding**: origin suppression branched on the *cleaned* value, so a 201-char or control-char origin cleaned to `""`, fell through, and wrote `session` **with the parent's id** — the exact fabrication the mechanism exists to prevent, arriving through the mechanism itself. Now it branches on header **presence**.
- **A3 was reachable, not theoretical**: `origin_session` bypassed the tier gate, and `browser-agent` forwards whatever `--print-session-id` produced, so `tmux:%3` genuinely landed there as `'%3'`.
- **`/whoami` and `/health` were mislabelled** as server-originated. They are operator subcommands; 125 rows / 2.0% of `kind='cmd'` were de-attributed on a false premise while the same `whoami` via POST was attributed. Made consistent rather than relabelled.

PR #551, opened 2026-08-19
fix(browser-bridge): the focus steal is `activate`, and it was ungated — 108/163 calls moved i3 focus within ±1s

ZacxDev:
## The answer to "is this opencode-only?" is **no — it is the shared layer**

The reported symptom was *"when opencode drives the browser bridge, Brave sometimes
grabs i3 focus"*. Measurement says the driver is not the variable. The **op** is.

`activate` is the only bridge op that moves the operator's real WM focus, it did so
**unconditionally** on every successful call, and it is reachable by *every* caller of
the `browser` CLI — Claude Code's Bash tool, an opencode session's bash tool, a script,
or the operator. It is **not** reachable by the sandboxed `browser agent` path at all.

## Per-op focus-steal rate

`activity.events`, 55,003 `source='browser-bridge' kind='cmd'` rows (2026-07-29 →
08-19), correlated against `source='i3' kind='window-focus'` rows for
`app='Brave-browser'`, joined per host with a two-sided ASOF join. `probe=placebo` is
the identical query with the command timestamp shifted **+120 s** — the background rate
of "Brave got focus around now".

The window is **±1 s**, not "5 s after": the server calls `i3_foreground()` *before*
`emit_cmd_event()`, so a command-caused focus event usually lands **before** the
telemetry row. A one-sided `[0, +5s]` window undercounts it, which is why the first
cut of this analysis understated `activate` at 41%.

| op | n | real ±1s | placebo ±1s | lift | real 1–5 s | placebo 1–5 s |
|---|---|---|---|---|---|---|
| **`activate`** | 163 | **108 (66.3%)** | 6 (3.7%) | **18.0×** | **5** | 22 |
| `screenshot` | 527 | 38 (7.2%) | 19 (3.6%) | 2.0× | 44 | 42 |
| `nav` | 654 | 34 (5.2%) | 13 (2.0%) | 2.6× | 51 | 44 |
| `open` | 472 | 17 (3.6%) | 14 (3.0%) | 1.2× | 41 | 23 |
| `eval` | 3,555 | 106 (3.0%) | 74 (2.1%) | 1.4× | 211 | 214 |
| `text` | 2,392 | 43 (1.8%) | 28 (1.2%) | 1.5× | 103 | 104 |
| `wake` | 1,092 | 18 (1.6%) | 14 (1.3%) | 1.3× | 52 | 39 |

`activate` is the only op whose mass concentrates at ±1 s **while its 1–5 s band
collapses** (5 real vs 22 placebo — *below* background). That is the shape a WM-driven
raise makes. A human context-switch spreads across both bands, which is what every
other row does.

### Instrument validation — the positive/negative control pair

- **Positive control: 108.** `activate` is the op I can confirm *does* raise the window
  (the server shells out to `i3-msg` for it, unconditionally, in code). The query
  returns 108 hits on it. So the correlation is wired to something real; a zero
  elsewhere is a measurement, not a dead query.
- **Under test: 22.** The strongest rival hypothesis (`screenshot`, below), once its
  confound is removed: 22/482 real vs 17/482 placebo — indistinguishable.

## Rival mechanisms, tested rather than assumed

**`screenshot` / `captureVisibleTab` — REFUTED, and the 2.0× above is a confound.**
`chrome.tabs.captureVisibleTab` really does only capture the focused window, so this was
the leading hypothesis. Splitting every op by "was there an `activate` on this host in
the preceding 10 s":

| op | activate within 10 s? | n | real ±1s | placebo ±1s |
|---|---|---|---|---|
| `screenshot` | **no** | 482 | **4.6%** | 3.5% |
| `screenshot` | yes | 45 | **35.6%** | 4.4% |
| `eval` | no | 3,397 | 2.5% | 2.0% |
| `eval` | yes | 158 | 13.9% | 3.8% |
| `open` | no | 470 | 3.6% | 3.0% |

Screenshot's entire apparent lift lives in the 45 calls that followed an `activate`. On
its own it is flat. The reason is in the extension: the `captureVisibleTab` fast path is
taken **only for a tab that is already visible**, and everything else routes through CDP
`Page.captureScreenshot`, which needs no foreground. **So there is no screenshot
trade-off in this PR** — screenshots do not need activation and did not cause steals.

**`wake` / `--wake` — REFUTED.** 1.6% vs 1.3% placebo over n=1,092. Consistent with its
design claim (CDP focus emulation moves no real focus).

**`open` — REFUTED.** 3.6% vs 3.0% over n=472. `chrome.tabs.create` is called with
`active:false`.

**i3's `focus_on_window_activation` — NOT the lever, and this is the cheaper fix that
does not work.** It is unset in `nix/i3/`, so i3's default `smart` applies. But it gates
`_NET_ACTIVE_WINDOW` *client requests* — and the server does not send one. It runs
`i3-msg '[class="Brave-browser" title="…"] focus'`, a **direct WM command**, which no
i3 focus policy gates. Chrome's own `chrome.windows.update({focused:true})` (which
`focus_on_window_activation` *would* gate) is a no-op under a tiling WM — that is
exactly why #196 added the `i3-msg` call in the first place. Changing the i3 config
would have moved nothing.

**Scope check on the mechanism:** `i3_foreground()` is the only WM-touching subprocess
in `server.py` (the only other `subprocess.run` is `git rev-parse --short HEAD` for
`whoami`), and it is called from exactly one branch, keyed on `op == "activate"`.

## Why "sometimes"

At the mechanism level it was never "sometimes" — the raise fired on **100%** of
successful activates. It only *presents* as intermittent because an i3 `window-focus`
event is a **transition**: when Brave already had focus there is nothing to record and
nothing to interrupt. The 66.3% is the interruption rate, not the fire rate.

## The opencode-vs-Claude split, and what I could not measure

The bridge's `session` column is **empty on every row** (this is exactly what the
sibling PR #549 fixes), so driver attribution had to be reconstructed by timing:

- **`browser agent` / `browser_tool_impl.mjs` — structurally exonerated.** `activate` is
  absent from `OP_TO_SERVER` and no `BROWSER_AGENT_ALLOWED_OPS` value can re-add it.
  Verified on `origin/main` *and* under PR #549 (which touches that file but only adds a
  header). Bridge commands falling inside a browser-agent session window: **7/1,178 real
  vs 9/1,178 placebo — below background.**
- **Ordinary opencode sessions — this is the opencode path that actually steals.** They
  get the full `browser` CLI via bash plus the `browser` skill (deployed to
  `~/.config/opencode/skills/browser`). 680 bridge commands are attributable to an
  opencode bash tool-call running the CLI, including **14 activates** — e.g. one
  session issued `activate` three times in six minutes. (That attribution query is
  open-ended from 2026-07-29 and sees 166 activates; the ±1 s table above closes at
  08-19 00:00 and sees 163. Same population, different cut-off.)
- **Claude Code — cannot be isolated from this dataset.** `source='claude'` carries only
  `prompt` / `session-summary` / `command`; there are no per-tool-call rows. So the
  non-opencode remainder mixes Claude Code, the operator's own shell, and untelemetered
  browser-agent runs. **I am not claiming a Claude-vs-opencode rate.**

🔴 **Known blind spot in that attribution:** `browser-agent` builds an *isolated*
opencode config dir that deliberately omits `plugin/`, and the telemetry plugin lives at
`~/.config/opencode/plugin/activity.js`. So browser-agent runs using that dir emit **no**
opencode telemetry — only 47 such sessions appear in ClickHouse, which is a sample, not
the population. The browser-agent conclusion therefore rests on the **code** fact
(`OP_TO_SERVER`), not on that sample.

**Interactive use:** 9 `source='zsh' kind='command'` rows contain `browser`; **0**
contain `activate`. Every activate in the window came from a non-interactive caller.

## The fix

The rule moves to the one place the screen is actually taken. Two prior mitigations
failed for the same structural reason:

1. **#225 was prose** — reword `HIDDEN_TAB_NOTE` so the model reaches for `wake`.
   `claude/RULES.md`: *"Prefer deterministic/structural fixes over prompt-tuning."*
2. **#180-era was an op-allowlist in one caller** — drop `activate` from the agent
   tool's `OP_TO_SERVER`. That is *one rule, one place* violated: it binds exactly one
   caller and every other caller walks past it.

Now:

- **`server.py`** — `i3_foreground()` runs only when the command carries `focus:true`
  (`focus_requested`: a **literal JSON `true`**, never Python truthiness, so a
  shell-interpolated `"false"` cannot read as consent). Otherwise the result reports
  `i3:"withheld"` plus a note that names `browser wake` **before** `--focus`. The
  Chrome-side tab activation is untouched — `activate` still works as a tab-state change.
- **`browser` CLI** — `--focus` / `--no-focus`, defaulting to **`[ -t 1 ]`**. A human
  typing `browser activate` gets today's behaviour; an agent on a pipe does not. The
  flag is always sent explicitly, so the decision is auditable in the request.
- **`focus` is popped before dispatch** like `target`/`tab`, so the extension's wire
  contract is byte-identical: **no extension rebuild, no Brave restart.** Only
  `server.py` needs a `home-manager switch` (the CLI is an `mkOutOfStoreSymlink` and is
  live from the working tree).
- **`payload.focus`** is recorded on the activate event, so the query that found the bug
  can falsify the fix after deploy.

## Trade-offs

- 🔴 **Behaviour change for every existing non-interactive `browser activate` caller.**
  A script or agent that genuinely wanted the foreground now silently does not get it
  until someone reads the `withheld` note and adds `--focus`. Intended direction — a
  caller that wanted the screen can re-run; a caller that did not cannot un-interrupt the
  operator — but it *is* a break.
- **`[ -t 1 ]` is a heuristic about intent**, not a proof of it. `browser activate | jq`
  from a real terminal now withholds. `--focus` is the fix and the error path names it.
- **No screenshot trade-off** (see above) — that hypothesis was measured and refuted.

## Residual I did **not** explain

`nav`, with activate-adjacent calls removed, still shows **30/627 real vs 9/627 placebo**
at ±1 s. That is ~21 excess events over three weeks against `activate`'s 108, and **this
PR does not address it.** I have not identified a mechanism. Against a Chrome-side raise:
`open` — which also creates and loads a tab — is flat (3.6% vs 3.0%). The likeliest
remaining explanation is a human-behaviour confound specific to `nav` (the operator
navigates somewhere and then looks at it), but I have not shown that.

## Test coverage

**Red-at-base matrix** — base extracted with `git archive` (no `.git`, no stash, no
worktree), my test files copied onto the **unpatched** implementation.

The red-at-base run was executed against `2eae44a`; the branch was subsequently rebased
onto `72d3ea8` (#541), which lands **zero** changes under `scripts/browser-bridge/`
(`git diff --stat 2eae44a..72d3ea8` touches only skills, docs, `obs-read`,
`session-manager` and their tests). The matrix is therefore unchanged by the rebase.

| test | at base | at HEAD |
|---|---|---|
| `test_activate_withholds_the_i3_raise_without_explicit_consent` | 🔴 RED | 🟢 |
| `test_withheld_activate_names_wake_before_the_focus_override` | 🔴 RED | 🟢 |
| `test_withheld_activate_carries_the_note_and_applied_does_not` | 🔴 RED | 🟢 |
| `test_focus_field_never_reaches_the_extension` | 🔴 RED | 🟢 |
| `test_focus_requested_accepts_only_a_literal_json_true` | 🔴 RED | 🟢 |
| `test_activate_telemetry_records_the_consent_decision` | 🔴 RED | 🟢 |
| `test_activate_defaults_to_no_focus_when_stdout_is_not_a_tty` | 🔴 RED | 🟢 |
| `test_activate_defaults_to_focus_when_stdout_IS_a_tty` | 🔴 RED | 🟢 |
| `test_activate_focus_flag_asks_for_the_raise_explicitly` | 🔴 RED | 🟢 |
| `test_activate_no_focus_flag_refuses_even_on_a_tty` | 🔴 RED | 🟢 |
| `test_activate_focus_composes_with_the_wait_flags` | 🔴 RED | 🟢 |
| `test_activate_rejects_an_unknown_flag_and_names_the_focus_flags` | 🔴 RED | 🟢 |
| `test_activate_still_activates_the_tab_when_the_raise_is_withheld` | 🟢 **INVARIANT GUARD** | 🟢 |
| `test_every_other_op_is_untouched_by_the_gate` | 🟢 **INVARIANT GUARD** | 🟢 |
| `test_the_pty_runner_actually_produces_a_tty` | 🟢 **harness positive control** | 🟢 |

**15 new tests: 12 regression (red at base) + 3 green-at-base**, the latter labelled as
such **in their own docstrings** and not counted as regression coverage.

The 6 pre-existing i3-outcome tests now pass `focus:true` explicitly — without it they
would assert against `"withheld"` instead of the `applied`/`skipped`/`failed` state they
exist to pin. Their docstrings say so.

### The suite was structurally blind to half of one dimension

The `bridge` fixture runs everything with `capture_output=True`, which pins stdout to a
**pipe** — so the TTY branch of the default was unreachable from it, and a mutant
replacing the whole thing with a constant `false` would have passed. Added
`_run_on_a_pty()` plus `test_the_pty_runner_actually_produces_a_tty`, a positive control
that requires the two ends to actually disagree before anything trusts the helper.
`M7-cli-const-false` below is the mutant only that runner can see.

## Mutation results

9 mutants, each applied to a fresh `git archive` tree of this PR's commit, run under
`PYTHONDONTWRITEBYTECODE=1` at a new path (so a same-length edit cannot be served from a
stale `.pyc`), with the **whole test file** run — never a `-k` filter that could exclude
the killing test.

| # | file | mutation (narrowest expression) | verdict | killed by |
|---|---|---|---|---|
| **M0** | `server.py` | `focus_requested` → `return False` — consent can never be granted | **KILLED** | `…carries_the_note_and_applied_does_not` + 7 others |
| **M1** | `server.py` | `if want_focus:` → `if not want_focus:` | **KILLED** ⚠ hand-verified | `…withholds_the_i3_raise_without_explicit_consent` + 6 others |
| **M2** | `server.py` | `is True` → `bool(...)` | **KILLED** | `…accepts_only_a_literal_json_true` *(only)* |
| **M3** | `server.py` | drop `body.pop("focus", None)` | **KILLED** | `…focus_field_never_reaches_the_extension` *(only)* |
| **M4** | `server.py` | reword the note keeping BOTH keywords, flipping their order | **KILLED** | `…names_wake_before_the_focus_override` *(only)* |
| **M5** | `server.py` | drop `extra["focus"] = …` | **KILLED** | `…telemetry_records_the_consent_decision` *(only)* |
| **M6** | `browser` | flip both ends of the TTY default | **KILLED** | both TTY-default tests |
| **M7** | `browser` | TTY branch collapsed to a constant `false` | **KILLED** | `…defaults_to_focus_when_stdout_IS_a_tty` *(only — the pty test)* |
| **M8** | `browser` | `--no-focus` silently ignored | **KILLED** | `…no_focus_flag_refuses_even_on_a_tty` *(only)* |

Notes on what the table shows beyond "all dead":

- **M2–M5 and M7–M8 are each killed by exactly ONE test** — the coverage is targeted, not
  incidental collateral from a neighbour going red.
- **M1 is the mutant the PR rests on**, and it is the one the harness got wrong (reported
  SURVIVED from an empty capture). Re-running the identical command by hand gives 7
  failures, with `test_activate_withholds_the_i3_raise_without_explicit_consent` firing its
  own assertion: *"activate without focus:true shelled out to i3-msg — the operator's
  screen was taken without consent"*, and the argv of the `i3-msg` call in the message.
  It is mutated at the narrowest expression — the condition only; the enclosing
  `if op == "activate"` and the `i3_foreground()` call both survive — so it cannot be
  dying for the "whole block deleted" reason.
- **M7 is the justification for the pty runner.** It is caught by that test *and nothing
  else*: with stdout pinned to a pipe, a constant `false` is indistinguishable from the
  correct default.
- **M4 is the justification for pinning the whole normalised note string.** It keeps both
  `browser wake` and `--focus` in the text — a keyword-presence guard would pass it — and
  only the order assertion plus the full-string comparison catch it.

See *Two integrity notes* below for how this harness was wrong twice before it was right.

## Merge with #549

🔴 **This PR touches `scripts/browser-bridge/server.py` and `scripts/browser-bridge/browser`,
both of which #549 also touches.** Test-merged rather than reasoned about:

- **Textually clean** (`ort`, no conflicts).
- **Semantically checked by reading the merged regions**, not assumed:
  - My `want_focus = focus_requested(body)` + `body.pop("focus")` reads a **body** field;
    #549's `session_origin = self.headers.get(HDR_SESSION_ORIGIN)` reads a **header**.
    Sequentially independent, both before `submit`.
  - My `extra["focus"]` flows into #549's extended `emit_cmd_event(...)`. Their
    reserved-key clearing list is `("sess_src", "origin", "origin_session")` — `focus` is
    not in it, and their attribution fields are written *after* `payload.update(extra)`,
    so neither clobbers the other.
  - #549's `browser_tool_impl.mjs` change adds only the `X-Session-Origin` header; it does
    **not** touch `OP_TO_SERVER`, so the browser-agent exoneration above still holds
    post-merge.
  - CLI edits do not overlap: #549 is at the session-id derivation (~L268) and
    recreate-close (~L1200); mine is the usage block (~L135) and the `activate` case
    (~L1560).
- **Full `scripts/browser-bridge/tests/` suite run on the merged tree: 547 passed.**
  (A first pass reported 2 failures in `test_browser_agent.py`; both were concurrent-load
  flakes — that file alone runs **60 passed in 27 s** unloaded, while the loaded full run
  took 332 s. Load inflates every test; a real assertion failure inflates one.)
- **Empirically checked, not just read**: the 37 focus + session-attribution tests run
  together on the merged tree and all pass, so `payload.focus` and #549's
  `payload.sess_src` / `session` genuinely coexist.

## Two integrity notes on the measurement itself

**One stray `focus=false` row already exists in `activity.events`**, at
`2026-08-19 05:14:20.688` UTC, payload `{"op":"activate","key":"","outcome":"ok","focus":false}`.
It predates any deploy — the deployed `server.py` is a `/nix/store` `home.file` copy without
this change (`readlink -f` confirms), so it can only have come from a test run. I could not
attribute it to a specific invocation. What I *can* show is that it is **not reproducible**:
the suite's autouse `_isolate_activity_spool` fixture redirects `ACTIVITY_SPOOL_DIR` per
test, `spool_emit.default_spool_dir()` reads that env at call time (base clone and worktree
copies are byte-identical), and re-running the seven focus tests left the ClickHouse count
at exactly 1. Set `FROM_TS` after the switch and the verification query ignores it.

**The mutation harness was wrong twice before it was right**, which is the part worth
reading:

1. *v1* built each sandbox from `scripts/browser-bridge` alone. `server.py` resolves its
   telemetry emitter by absolute path from the **repo root**, so every telemetry test failed
   for reasons unrelated to any mutant; with `-x`, pytest stopped at the first, and all six
   server mutants reported the **identical wrong killer**. A harness that returns the same
   answer for every input is not measuring the input.
2. *v2* fixed the tree and dropped `-x`, and added a **baseline gate** (the unmutated
   sandbox must be green first). That gate immediately earned itself: its first run went red
   on a telemetry test — concurrent load, confirmed by re-running the file alone (60 passed
   in 27 s vs a 332 s loaded full run). Without the gate that would have been read as a
   mutant result.
3. *v2 still reported `M1-invert-gate` as **SURVIVED*** — the single most important mutant.
   Re-running the identical command by hand gave **7 failures**, including
   `test_activate_withholds_the_i3_raise_without_explicit_consent` firing its own specific
   assertion (`… shelled out to i3-msg — the operator's screen was taken without consent`).
   The harness had captured no output at all, and *"no test failed"* and *"no test ran"* are
   the same empty set. Fixed by requiring pytest's own summary line and a plausible
   collected count before any verdict is believed; a run that cannot be measured now reports
   **COULD NOT MEASURE**, never SURVIVED.

The table above is from the corrected harness. **`M1` is additionally hand-verified** — it is
the mutant the whole PR rests on, and I am not willing to report it on a harness that has
already been wrong about it once.

## Round 2 — audit follow-ups (`/audit-pr 551`)

The blind audit came back safe-to-merge and independently killed `M1` with 7 tests,
the mutant my own harness once mis-reported as SURVIVED. Its own differently-built
20-mutant sweep went 20/20. Three items landed on top of `3741123`:

### 🟡1 — a real regression the first round shipped, and why no test saw it

The consent check ran **before** `i3_available()`. So on a headless / non-i3 /
server-mode host, a non-consented `activate` answered `i3:"withheld"` *and attached
the note telling the caller to pass `--focus`* — advice for a raise that host cannot
perform. Pre-gate, the same call truthfully answered `skipped`.

Fixed by asking **CAN we raise before MAY we**: unavailability now wins over
non-consent, so a host with no i3 always reports `skipped`.

🔴 **Why the suite was blind to it:** when the gate landed I updated
`test_activate_i3_skipped_when_unavailable` to pass `focus:true` — which made
`skipped` reachable only *with* consent, i.e. exactly one cell of a 2×2. A defect in
a cell no fixture ever builds is invisible no matter how green the suite is. All four
cells are now asserted, so the ordering is pinned rather than incidental:

| i3 available | consent | state | note attached |
|---|---|---|---|
| yes | yes | `applied` | no |
| yes | no | `withheld` | **yes** (its `--focus` advice is actionable here) |
| no | yes | `skipped` | no |
| no | **no** | `skipped` | no ← **the cell that shipped wrong** |

### 🟡2 — the README overstated the guarantee

It called the raise "operator-only" and said it "requires EXPLICIT CONSENT". That is
false as an *authorization* claim and is now reworded. What this change actually is:
the **default** flips from raise-always to raise-only-on-request — worth having,
since all 166 measured activates came from callers that never asked. What it is
**not**: a boundary. The refusal note names the bypass flag (deliberately), the CLI
is on every agent's `PATH`, and `/cmd` is loopback HTTP reachable with `curl` plus
`~/.config/browser-bridge/token`. The only structural barrier remains the sandboxed
browser-agent's op allowlist, which binds that one tool. (No env kill switch — out
of scope, and a separate decision.)

### 🟡3 — expect a residual, or the post-deploy check gets mis-scored

This PR gates the host-side `i3-msg` raise **only**. The extension still calls
`chrome.windows.update(windowId,{focused:true})` unconditionally
(`service_worker.js:1403`), and `focus_on_window_activation` is **unset** in
`nix/i3/config.nix`, so i3's default `smart` applies — under which a
`_NET_ACTIVE_WINDOW` request from a window **on the currently visible workspace is
granted focus**. Both paths fired together on every pre-fix activate, so the 66.3%
in the table above **cannot be attributed between them**.

So, post-deploy: **withheld activates should fall TOWARD background, not necessarily
TO it.** A non-zero ±1s residual is the Chrome-side path, not a leaking gate; only a
residual near the *pre-fix* rate would mean the gate failed.

And: **treat an absent `focus` field as UNKNOWN, never as pre-deploy.** Throttled
(429) activates return before the `payload.focus` augmentation runs, so a new-server
row can legitimately carry no `focus` key. Bucketing "no focus field" as "old server"
would silently mix them into the pre-deploy bucket.

### Round-2 verification (the gate reset — previous round's green not carried forward)

Baseline for this round is **`3741123`**, the round-1 commit.

| test | at `3741123` | at HEAD |
|---|---|---|
| `test_activate_i3_state_matrix[noi3+refused]` | 🔴 RED — `got i3='withheld', want 'skipped'` | 🟢 |
| `test_non_consented_activate_on_a_non_i3_host_says_skipped_not_withheld` | 🔴 RED — behavioural, same message | 🟢 |
| `test_i3_state_constants_match_the_literals_the_matrix_asserts` | 🔴 RED (new constant absent) — structural, not behavioural | 🟢 |
| `test_activate_i3_state_matrix[i3+consent]` / `[i3+refused]` / `[noi3+consent]` | 🟢 **INVARIANT GUARD** ×3 | 🟢 |

The two behavioural reds are the regression coverage; the three green cells are
labelled invariant guards in the test's own docstring and are not counted as such.

🔴 The named regression test deliberately asserts against the **literal** `"skipped"`
rather than `S.I3_SKIPPED`. My first draft used the constant and went red at the
baseline with an `AttributeError` — red for the *wrong reason*, which proves nothing
about behaviour. A separate one-line test pins the constants against those literals so
the indirection cannot rot.

**Round-2 mutation — the ordering, isolated to one conjunct:**

```
if not i3_available():                 ->   if want_focus and not i3_available():
```

That is the narrowest expression of the shipped defect ("availability is only
consulted when consent was given"). The branch, the enclosing `if op == "activate"`,
and the `i3_foreground()` call all survive, so it cannot die for the
whole-block-deleted reason.

```
BASELINE (unmutated)  failures=[]  ran=325   -> green, so the verdict is readable
MUTANT                ran=325
  failures = ['test_activate_i3_state_matrix[noi3+refused]',
              'test_non_consented_activate_on_a_non_i3_host_says_skipped_not_withheld']
VERDICT: KILLED by EXACTLY the 2 non-consented x non-i3 tests, and nothing else
```

🔴 **An equivalent-mutant trap avoided:** simply reordering the branches to
`if want_focus: … elif not i3_available(): …` still yields `skipped` for the
non-consented/non-i3 cell — behaviourally identical, so it could never be killed and
would have read as a coverage hole. The ordering only has observable consequences
where `withheld` would otherwise win, which is what the mutation above targets.

Harness controls, after the round-1 battery was wrong twice: fresh `git archive` tree
per run at a new path, `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged, the **whole**
test file run (no `-k` that could exclude the killer), a green-baseline gate, and a
per-run collected-count check so an empty capture reports COULD NOT MEASURE rather
than SURVIVED. No `git stash` anywhere — the working-tree fix is overlaid onto a
`git archive HEAD` tree, since `refs/stash` is repo-global and shared with sibling
agents.

## Round 3 — documentation only (delta re-audit follow-ups)

No code changed. `git diff --stat` for this round is README / SKILL.md / two comment
blocks / one extension comment.

**N2 — the residual mechanism was asserted two contradictory ways, and I did not know
which was true.** My round-2 README said Chrome's `windows.update({focused:true})` IS
granted focus under i3's `smart`; the CLI comment said Chrome-side focus is "a no-op
under i3" and `server.py` said it "takes nothing". I resolved it by **establishing the
i3 half from a citable source on this host** rather than picking a side:

> i3 4.24 userguide §4.30 *Focus on window activation* — `focus_on_window_activation
> smart|urgent|focus|none`; **smart**: "This is the default behavior. If the window
> requesting focus is on an active workspace, it will receive the focus. Otherwise,
> the urgency hint will be set."
> (`/nix/store/…-i3-4.24/share/doc/i3/userguide.html`; i3's own example of a window
> requesting focus is `google-chrome www.google.com`.)

`focus_on_window_activation` is unset in `nix/i3/config.nix`, so `smart` applies.
That makes "a no-op under i3" **false as an unconditional claim** — it holds only
cross-workspace, which is the usual agent case and precisely why #196 needed
`i3-msg`. Both over-claims are corrected.

🔴 **What remains genuinely unknown, and is now labelled as such in all three
places:** whether `chrome.windows.update({focused:true})` actually emits the X11
activation request. It is the standard mechanism and the extension's own comment says
it "REQUESTS focus", but nobody has measured it — not this PR, not either audit. So
the Chrome-side contribution is **unknown: not zero, not proven non-zero.**

Because of that, the post-deploy guidance is written to be **robust to both
readings**, with an explicit discriminator instead of a prediction. i3 focus events
carry `payload.workspace`, so split withheld activates by whether Brave was already
on the visible workspace:

* residual concentrated in the **same-workspace** subset → Chrome-side path is real,
  `smart` granted it, gate working as designed;
* residual across **both** subsets → something still calls `i3-msg`; the gate leaked.

**N4 — the baseline the residual is compared against.** The README cited 66.3% from
this PR body against its own 66.9% table. Same population, different cut-off; both
are now labelled with their window, and the README states the comparison baseline
explicitly — **66.9% (111/166) pre-fix**, against **~2–7% background** (`wake` 1.7%,
`eval` 3.0%, `screenshot` 7.3%). A withheld residual in that band is consistent with
the gate holding; at/near 66.9% it leaked; **15–40% is a finding, not a failure** —
it would mean the Chrome-side path contributes materially.

**N3 — `SKILL.md` stated the pre-fix binary** (`withheld` unless `--focus`, no
`skipped`) on the always-loaded routing surface, so an agent on a headless host would
expect the wrong state. Fixed, and **paid for with an eviction in the same commit**:
the dated `2026-07-31` agent-measurement specifics went to `reference/agent.md`,
which already held them, keeping the pointer and the load-bearing "n=1, not a rate"
caveat. Net **−38 bytes** (11,916 → 11,878; headroom 372 → 410 against the required
250). The `description:` routing line was not touched.

**N1 — a dangling cross-reference** to the heading I renamed, plus the disavowed
"consent-gated"/"operator-only" framing surviving at `server.py:144`,
`server.py:2190` and `service_worker.js:1368`. All swept. (`SKILL.md:73`'s
"operator-only" is left alone — that is `upload`, where it is true.)

### 🔴 Interaction with #557, which landed on main mid-round — READ BEFORE MERGING

`2f57c6c` *"`i3: applied` was reachable with ZERO windows raised"* is on main and not
in my ancestry. It rewrites the same function my consent gate wraps. I read it rather
than merging it, and **there is a real interaction — I have not fixed it, per scope.**

**1. Does it change what `applied` means for my matrix?** The three strings survive
(`_annotate_i3` on main explicitly keeps "EXACTLY the three values callers already
branch on") — but the **contract around them changed twice**:

* `i3_foreground` now returns a **`(state, detail)` tuple**, not a string. My call
  site does `state = i3_foreground(...)`, and `_annotate_i3(result, state)` — main's
  is `state, detail = …` / `_annotate_i3(result, state, detail)`.
* `applied` is now **earned**: main confirms via `i3-msg -t get_tree` that the window
  exists and ends up focused, because i3 answers a criteria matching **zero** windows
  with `[{"success":true}]` and rc 0. New failure details appear (`no_match`,
  `tree_unreadable`, `not_focused`, `focus_error`).

🔴 So my matrix's consented × available cell asserts `applied` against the **old**
meaning. My `_enable_i3` fixture stubs a bare `subprocess.run` returning rc 0 with
empty stdout; under #557 that path parses stdout as `get_tree` JSON and would yield
`failed`/`tree_unreadable`, not `applied`. **The fixture, not just the assertion,
needs reworking on the merged tree.**

**2. Does it interact with CAN-before-MAY?** Yes, at the same lines. #557 keeps the
unavailable case *inside* `i3_foreground` (returning `("skipped","unavailable")`);
my change **hoists** it to the call site as a bare `I3_SKIPPED` with no detail. Merged
naively those are redundant and shape-inconsistent — my hoisted branch would produce a
`state` with no `detail` while every other path produces a tuple, and my
`state = i3_foreground(...)` line would put a **tuple into `data["i3"]`**, which is
silently wrong JSON rather than a crash.

**The good news, measured not assumed:** I test-merged `2f57c6c` into `b6dd48e` in a
throwaway detached worktree (my branch untouched, merge aborted, no `rerere`
resolution recorded). It **CONFLICTS loudly** — `server.py`, `browser`, and
`test_server.py` all conflict, and the call-site conflict region contains all three
variants side by side (`elif want_focus:` / base / `state, detail =`). So this cannot
merge silently wrong; a resolver is forced to look at exactly the right lines.

My read on the resolution — **your call, I have not done it**: hoisting was the point
of 🟡1, and #557's `("skipped","unavailable")` already carries that meaning with a
detail. Keeping #557's structure and adding only the `want_focus` branch (so
`withheld` is a fourth state alongside its detail) looks like the smaller change than
keeping my hoist. That also makes N5/N6 moot, since the inner guard stops being dead.

### N5/N6 — CLOSED by the round-4 resolution, not deferred any further

Round 3 recorded a follow-up: the CAN-before-MAY hoist had made
`i3_foreground()`'s own `i3_available()` guard unreachable from its only production
caller, and made `i3_available()` run twice per consented activate.

**Both are gone.** Adopting #557's structure means the consented path goes straight
into `i3_foreground`, whose own guard now answers the non-i3 case as
`("skipped","unavailable")` — so that guard is **reachable again**, and
`i3_available()` runs **at most once per request** (inside `i3_foreground` when
consented, at the call site when refused, never both). The dedupe happened as a
side effect of taking the smaller diff, which is the outcome that was hoped for.

### One round-3 edit was reverted, because the extension is not free to touch

N1 asked me to sweep the disavowed "operator-only" framing at
`extension/service_worker.js:1368`. I did — and `test_committed_build_marker_matches_
the_extension_source` went red: the committed `extension/build_id.js` BUILD_MARKER
hashes the extension source, so **even a one-line comment change makes it a new
build**. Regenerating the marker is what the gate asks for, but it would mean the
extension must be reloaded — destroying the *"no extension rebuild, no Brave restart"*
property that is one of this PR's real deployability wins and that the first audit
independently confirmed.

A cosmetic prose sweep is not worth converting a no-restart deploy into a
restart-required one, so **I reverted that one file**; `gen-build-marker.py --check`
is clean and the extension is byte-identical to `origin/main`. The disavowed wording
survives at that one site only, and is listed here rather than silently dropped.

## Round 4 — resolved against #557 (`origin/main` merged in)

`2f57c6c` (*"`i3: applied` was reachable with ZERO windows raised"*) rewrote the same
function this gate wraps. `origin/main` is now **merged into this branch** as one
reviewable merge commit, resolved by hand with `rerere` explicitly disabled
(`git -c rerere.enabled=false merge`) — this repo has been bitten by rerere replaying
a resolution from a different merge, and it is enabled globally here, so that was not
optional. 10 conflict regions across `server.py` (5), `browser` (2) and
`test_server.py` (3).

### The shape adopted, and one correction to the plan

**#557's structure wins**, as recommended: `("skipped","unavailable")` stays *inside*
`i3_foreground`, and only the consent branch is added. One correction found while
resolving — taking that literally would have **regressed 🟡1**: with consent checked
first and nothing else, a refused activate on a headless host answers `withheld`
again, which is the exact bug round 2 fixed. The resolution therefore puts the
availability check on the **refusal** branch:

```python
if want_focus:
    state, detail = i3_foreground(_result_title(result))   # #557's path, verbatim
elif not i3_available():
    state, detail = I3_SKIPPED, I3_DETAIL_UNAVAILABLE      # CAN still beats MAY
else:
    state, detail = I3_WITHHELD, I3_DETAIL_NOT_REQUESTED
```

All four original cells keep their `i3` value, every cell now carries a `detail`, and
— the point — **#557's inner guard becomes the thing that answers the consented ×
non-i3 case**, which is what closes N5/N6 above.

### Hazards closed, each with the test that pins it

| # | hazard | pinned by |
|---|---|---|
| 1 | `i3_foreground` returns a **tuple**; a call site that forgets to unpack puts `("applied","focused")` into `data["i3"]` — silently wrong JSON, not a crash | `isinstance(data["i3"], str)` asserted in **every** matrix cell |
| 2 | `_enable_i3` (bare `subprocess.run`→rc 0, empty stdout) can no longer produce `applied` — it now parses as an unreadable tree | matrix + the note test moved onto #557's `_FakeI3`; `_enable_i3` kept only where the i3 state is irrelevant |
| 3 | `applied` is now **earned**; `no_match` etc. exist | matrix re-derived — see the cell table below |
| 4 | a withheld raise must not do the `get_tree` round trip | `test_refused_activate_makes_no_i3_round_trip_at_all` asserts on the fake's **call log**, not on the reported state |

### The matrix, re-derived — which cells changed MEANING vs only FIXTURE

| cell | before | now | what changed |
|---|---|---|---|
| i3 + consent + **window matches** | `applied` | `applied`/`focused` | **fixture only** — needs a real window in `_FakeI3` |
| i3 + consent + **no match** | *did not exist* (reported `applied`) | `failed`/`no_match` | **NEW CELL** — this is #557's whole point |
| i3 + refused | `withheld` | `withheld`/`not_requested` | meaning unchanged; detail added |
| no i3 + consent | `skipped` | `skipped`/`unavailable` | **now answered by #557's own guard**, not my hoist |
| no i3 + refused | `skipped` | `skipped`/`unavailable` | meaning unchanged |

Plus a **seam guard**: the refused path reports "i3 is not here" *without* calling
`i3_foreground`, so the same fact is produced in two places.
`test_call_site_unavailable_detail_matches_i3_foreground` pins them by **asking
`i3_foreground`**, so it fails if either side renames its detail.

### One deliberate divergence from #557, called out because it makes a claim false

#557's `_annotate_i3` docstring says `i3` keeps "**EXACTLY the three values** callers
already branch on". With the gate there is a fourth, `withheld`. I did **not** fold it
into `skipped`: `skipped` means the host *cannot* raise, and collapsing the two
destroys the CAN-vs-MAY distinction on the one field consumers branch on. But #557's
sentence is now false, so I **corrected it in place** rather than leaving it to rot,
and spelled out that an exhaustive three-way switch would meet an unknown value.
#557's actual compatibility argument (a consumer treating `failed` as fatal) still
holds. Flagging it because it is a contract change on someone else's stated invariant.

### Verification — baseline is `origin/main` (which has #557, not the gate)

Extracted with `git archive` (no `.git`, no stash), my test file laid over it:

**Behaviourally RED at base** — real regression coverage:
* `test_activate_i3_state_matrix[i3+refused]` — `i3='failed', want 'withheld'`
* `test_refused_activate_makes_no_i3_round_trip_at_all` — `assert 'applied' == 'withheld'`
  (at base the refused activate genuinely raised the window)
* `test_focus_field_never_reaches_the_extension` — *"the server forwarded the `focus`
  consent flag to the extension"*

**Structurally RED at base** (new API surface — `S.I3_WITHHELD`, `S.focus_requested`,
the detail constants): 5 more, labelled as structural rather than counted as
behavioural.

**GREEN at base — INVARIANT GUARDS, not regression coverage:** the four other matrix
cells. Worth being precise about why, because it differs from round 2: `noi3+refused`
was the *regression* cell at `3741123`, and is an invariant guard here — main's
structure already answers it correctly. The gate's only behavioural signature against
this base is **refused-on-an-i3-capable-host**, which is exactly what it should be.

Merged tree: **341 passed, 0 failed** (`test_server.py`).

### Mutation on the merged shape — 6/6 killed

Controls: fresh `git archive HEAD` tree per mutant at a new path with the working-tree
files overlaid (**no `git stash`** — `refs/stash` is repo-global and shared with sibling
agents), `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged, the **whole** file run (no
`-k` that could exclude a killer), a green-baseline gate (**failures=[] ran=341**), and
a per-run collected count so an empty capture reports COULD NOT MEASURE rather than
SURVIVED.

🔴 Plus a **selectability pre-check**, because my own harness has been wrong before: every
expected killer name is verified to exist in the collected list *before* any verdict is
believed — otherwise a typo'd expectation reads as KILLED-BY-WRONG-TEST forever.
Result: 341 collected, all expected names present.

| # | mutation (narrowest expression) | verdict |
|---|---|---|
| **M1** | `if want_focus:` → `if not want_focus:` | **KILLED** (20 tests) |
| **M2** | delete the `elif not i3_available():` branch | **KILLED — exactly the 2 expected** |
| **M3** | `is True` → `bool(...)` in `focus_requested` | **KILLED — exactly the 1 expected** |
| **M4** | drop `body.pop("focus", None)` | **KILLED — exactly the 1 expected** |
| **M5** | `state, detail = i3_foreground(…)` → `state = …` (**the tuple hazard**) | **KILLED** (19 tests) |
| **M6** | `I3_DETAIL_UNAVAILABLE = "not_available"` | **KILLED** (3 tests) |

Three killed by exactly the predicted set. The other three killed by a **different** set
than I predicted — in each case my *expectation* was an imprecise guess, not a coverage
gap, and the difference is itself informative:

* **M1** kills 20, not the 6 I guessed: inverting the gate means consented calls stop
  raising, so nearly all of #557's own raise-path tests die too. My four gate tests are
  among them, including `[i3+refused]` and `test_refused_activate_makes_no_i3_round_trip_at_all`.
* **M5** kills 19: the tuple reaches `data["i3"]` on every consented path, so every test
  that reads that field dies — including the `isinstance(data["i3"], str)` guard added
  for exactly this. That guard is the one that would still fire if the others were ever
  loosened.
* **M6** kills 3, not 4 — and the one that *survived* is the interesting part.
  `[noi3+consent]` did **not** die, because that cell reaches `i3_foreground`, which
  returns its **own** `"unavailable"` literal rather than the call-site constant. That is
  precisely the drift this seam has, demonstrated: the two sides are independent, which
  is why `test_call_site_unavailable_detail_matches_i3_foreground` exists and why it
  *did* die.

**Equivalent-mutant note carried forward:** reordering the branches to
`if want_focus: … elif not i3_available(): …` is what the resolution now *does* — in the
round-2 shape that reordering was behaviourally identical and unkillable, whereas here it
is load-bearing, because it is what routes the consented × non-i3 case through
`i3_foreground`'s own guard.

## Round 5 — two false documentation claims, both self-inflicted

Docs + one test. No behaviour changed.

### 🟡1 — a `✅ LIVE-VERIFIED` badge this PR inherited by insertion

`README.md` carried a live-verification badge from #225 describing the **`wake`** rig
(`open` → `text` → `wake` → `text`) — a sequence containing **no `activate` call at
all**. Round 2 inserted 136 lines directly above it, so it silently became the verdict
closing *"The focus steal is OPT-IN BY DEFAULT"*, while this PR body said (correctly)
that no live reproduction was performed. Two always-shipped docs in one PR disagreeing
about whether the changed path was exercised live — and the README is the one an agent
reads.

Fixed as a **move, not a rewrite**: the badge is back above the new `###`, closing the
section it actually describes.

🔴 And the true state is now stated where the reader is, because nobody had written it
plainly: **`browser activate --focus` has never been run against a real i3 — by anyone.**
Not here (reproducing this bug means taking the operator's screen), and not in #557,
whose commit says nothing was checked against a live i3 either. **The escape hatch is
the unverified surface.** The *withheld* path is the half that was observed, and it is
unverified in a much weaker sense — it never enters `i3_foreground`, which is asserted
on the fake's call log rather than inferred.

### 🟡2 — a ledger that named a relationship but pinned a component

`test_i3_foreground_state_vocabulary_is_closed` claimed to guard *"`result.data.i3`
keeps EXACTLY the three values consumers branch on"*. It enumerates `i3_foreground`'s
**returns**, so it structurally cannot observe `withheld` — which is produced at the
call site. I added a fourth value to that field and **the test stayed green**.

I had rewritten the identical false sentence in `_annotate_i3`'s docstring during the
merge and missed this one — the artifact a maintainer would actually consult. Net
effect before this round: **nothing was a closed ledger over `data.i3`**, and a fifth
value would have broken no test.

Fixed in two parts:

* the old test's docstring is **rescoped** to what it really enumerates
  (`i3_foreground`'s own returns), with the reason recorded;
* a new **`test_data_i3_value_set_is_a_closed_ledger`** covers the field across **both
  producers** — `i3_foreground`'s returns *and* the call-site assignments that bypass
  it — and fails when the set GROWS **or** SHRINKS, naming which.

🔴 The GROW half is **structural (AST)**, not behavioural, on purpose: a fifth value
added on a branch no fixture exercises is exactly what a driven sweep cannot see. Plus
`test_data_i3_ledger_is_not_vacuous`, a positive control that both parsers actually
find something — an empty parser would let the other half carry the assertion and hide
a whole producer.

**Proof (the audit's exact standard):** a mutant adding a fifth value on an
**unreachable** branch (`elif False:`) — so behaviour is unchanged and every other test
stays green — was **KILLED by exactly `test_data_i3_value_set_is_a_closed_ledger`, and nothing
else**:

```
BASELINE   failures=[]  ran=343          (green, so the verdict is readable)
MUTANT     ran=343      failures=['test_data_i3_value_set_is_a_closed_ledger']
VERDICT:   KILLED by EXACTLY the new call-site ledger, and nothing else
```

Both halves of that are the point: **KILLED** proves the ledger closes the set, and
**nothing else** proves the mutant really was invisible to every behavioural test —
which is what a driven sweep would have missed.

### 🟢 Recorded, deliberately not fixed (scoped out by the coordinator)

* `server.py:2370-2373` — *"'withheld' always means this host could have"* is now only
  *"i3 is present"*. A `--focus` retry can still land `no_match` / `not_focused` /
  `focus_error` / `no_title`. The **code is right**; the certainty in the comment is
  overstated by one clause. Left as-is this round to keep it docs-only-plus-one-test.
* My round-4 matrix note **undercounts**: the consented × available cell actually
  splits into `applied/focused`, `no_match`, `not_focused`, `focus_error`,
  `tree_unreadable` and `no_title` — **three** of which used to report `applied`, not
  one. All of that is #557's behaviour and all of it is covered; the error was in my
  description, not the tests.

### Context

**#549 is merged to main** (`13f4278`). This branch still merges clean against it
(`merge-tree` rc 0), and the combined `emit_cmd_event` region composes correctly —
`extra["focus"]` sits above the `session_id`/`attribute_session` kwargs. That was read,
not run; the merged-tree gate is the authority.

## What is NOT verified

- **No live reproduction was performed.** Reproducing this bug means taking the
  operator's real screen, and the telemetry was decisive without it. The mechanism claim
  rests on 55,003 rows plus the code path, not on a live probe.
- **Nothing here is deployed.** `server.py` is a `home.file` copy and needs
  `home-manager switch`; the gate is inert until then. Merged ≠ deployed ≠ verified.
- **The post-deploy check is the operator's**, and it is cheap: after the switch, re-run
  the per-op query and confirm `payload.focus=false` activates fall to background while
  `focus:true` ones still correlate. That is the falsification this PR was built to make
  possible, and it has not been run.


ZacxDev:
## Gate status — what I can and cannot vouch for

**Vouched, with numbers:**

| tier / target | result |
|---|---|
| `scripts/browser-bridge/tests` (pytest) | **PASS** — `collected=517 passed=517 skipped=0 floor=469` (502 before + the 15 new) |
| node tier, whole repo | **PASS** — `TOTAL suites=4 files=33 tests=1116 pass=1116 fail=0 skipped=0` |
| ↳ `scripts/browser-bridge/tests` (node) | **PASS** — `files=15 tests=495 pass=495 skipped=0 floor=490` |
| every other pytest target that ran | PASS, each above its own floor (collector ×6, dl-router 991, validation, session-analysis ×2, repo-cos, task-spec-drafter, claude-hooks ×6, opencode, plus 6 script targets) |

**NOT vouched — `scripts/gate.sh --tier pytest` exited 1**, on four targets this PR does
not touch. `git diff --name-only origin/main..HEAD` is six files, all under
`scripts/browser-bridge/`:

| target | what happened | attribution |
|---|---|---|
| `scripts/tests` | `collected=2` vs floor 4976 — collection aborted: `ModuleNotFoundError: No module named 'yaml'` in `test_opencode_config.py` | missing dep in my ad-hoc `nix-shell -p logrotate python3Packages.pytest` |
| `scripts/signal/tests` | `collected=2` vs floor 444 — `SystemExit: psycopg2 is required` | same |
| `scripts/mail-actions/tests` | `collected=0` | same shape |
| `scripts/initiatives/tests` | `collected=784 passed=775 failed=9` | **unattributed** — no import path from `scripts/browser-bridge/` |

🔴 **The honest gap.** `scripts/tests` holds the repo-wide content gates that *do* scan
browser-bridge files — `test_no_captured_text.py`, `test_no_captured_markup.py`,
`test_skill_audit.py`, `test_prune_skill_size.py`, `test_runtime_shebangs.py`. Because
collection aborted on the missing `yaml`, **those gates never ran against this PR.** I
re-ran the target with `pyyaml` supplied and it got past collection and was ~80% green
when I ran out of window; I did not capture its final failure list, so I am not claiming
it passed.

What I did check by hand for the PUBLIC-repo rule, with a working positive control
(grep matched `activate` 145× in the diff / 39× in the body, and 0× for the pattern under
test): the diff and this PR body contain **no** real third-party hostname, media path, or
captured text. Test fixtures reuse the file's existing synthetic `example.test` domain.

**The two commands that would close this**, neither of which finished in my window:

```
nix build .#checks.x86_64-linux.pytests     # hermetic, correct inputs — the real CI gate
```

Also worth a control run: `scripts/initiatives/tests` on `origin/main` — if it fails 9
there too, all four are environmental and the gate is clean for this change.

ZacxDev:
## Update — the gap in the previous comment is now closed

`scripts/tests` re-run with `pyyaml` supplied got past collection: **25 failed, 5278
passed** (443 s).

**The repo-wide gates that actually scan this PR's files all pass.** Run explicitly by
name rather than inferred from an absence of failures — a `-q` run does not print passing
test names, so "0 content-gate failures" was indistinguishable from "they never ran":

```
pytest scripts/tests/test_no_captured_text.py scripts/tests/test_no_captured_markup.py \
       scripts/tests/test_no_client_hostnames.py scripts/tests/test_runtime_shebangs.py \
       scripts/tests/test_rules_size.py scripts/browser-bridge/tests/test_skill_size.py
→ 230 passed in 2.96s
```

That covers the PUBLIC-repo captured-text/markup/hostname gates, the shebang gate, and
**both** enforced byte ceilings (`claude/RULES.md` and `scripts/browser-bridge/SKILL.md`
— SKILL.md went 11,821 → 11,916 bytes, leaving 372 of the required 250 headroom, so no
eviction was owed).

**All 25 failures are tests that shell out to the runner, the gate, or a tool my ad-hoc
shell lacked** — none of them touch `scripts/browser-bridge/`:

| file | n | why |
|---|---|---|
| `test_claude_log_rotate.py` | 6 | leads with `test_logrotate_is_on_path` — I supplied `pyyaml` this run but not `logrotate` |
| `test_run_tests_floors.py` | 9 | spawns `run-tests.sh`, which needs the full `REQUIRED_TOOLS` set |
| `test_no_real_launchers_all_targets.py` | 5 | spawns the runner across all targets |
| `test_run_tests_preconditions.py` | 2 | spawns `run-tests.sh` |
| `test_gate_exit_truthfulness.py` | 2 | spawns `gate.sh` |
| `test_opencode_engine.py` | 1 | pins the `opencode` version every measurement is keyed to |

So the residual unknown is narrower than the first comment said: it is now just
**`scripts/initiatives/tests` (9 failed of 784)**, which has no import path from
`scripts/browser-bridge/`. The control for it is one command on a clean tree:

```
git archive origin/main | tar -x -C /tmp/ctl && \
  pytest /tmp/ctl/scripts/initiatives/tests -q
```

If it fails 9 there too, the gate is clean for this change. I did not run it.

ZacxDev:
## Update 2 — the last unattributed gate failure is now attributed. It is pre-existing.

I ran the control I said I hadn't: `scripts/initiatives/tests` against a clean
`git archive origin/main` tree (no `.git`, unpatched — verified
`grep -c focus_requested server.py` → 0).

```
branch  (3741123):  9 failed, 775 passed
origin/main:        9 failed, 775 passed     <-- identical
```

Same count, same three files, same nine test names:

- `test_archive.py` ×4 — `test_read_archived_projects_rows`,
  `…missing_table_returns_empty`, `…never_raises_and_rolls_back_on_error`,
  `test_list_archived_wraps_read_over_its_own_connection`
- `test_viewer.py` ×4 — the four `test_attach_recaps_*` cases
- `test_sync.py` ×1 — `test_write_snapshot_inserts_snapshot_then_rows_with_jsonb`

All nine are Postgres/`psycopg2`-shaped, consistent with the same missing-dependency
class as the `scripts/signal` and `scripts/mail-actions` collection errors.

**So every failure in the `scripts/gate.sh --tier pytest` run is now accounted for and
none is caused by this PR:**

| target | failures | attribution |
|---|---|---|
| `scripts/tests` | 25 | tests that shell out to the runner/gate, or need `logrotate` / a pinned `opencode` — absent from my ad-hoc shell |
| `scripts/signal/tests` | collection | `psycopg2` missing |
| `scripts/mail-actions/tests` | collection | same class |
| `scripts/initiatives/tests` | 9 | **reproduced identically on `origin/main`** |

Everything this PR actually touches is green: `scripts/browser-bridge/tests` pytest
**517/517** (floor 469), node **495/495** (floor 490), and the repo-wide content +
byte-ceiling gates **230 passed** when run by name.

Still not done, and still only the operator can: `nix build .#checks.x86_64-linux.pytests`
(never returned in my window — it is the hermetic instrument, and I would rather it be run
than take my ad-hoc shell's word), and the post-deploy falsification described in the PR
body, which needs a `home-manager switch` first.

ZacxDev:
## ⚠ Correction to the two comments above — the hermetic gate DID return, and it FAILED

My previous comments said `nix build .#checks.x86_64-linux.pytests` "never returned in my
window". It has since returned:

```
error: Cannot build '/nix/store/pbkqqxcqyvgamm40dr6yah6kj7plq7p9-devrc-pytests.drv'.
>     scripts/initiatives/tests  intercepted=0  systemctl-reads=0  plugin=1
> RESULT: FAIL (exit=1)
PYTESTS_BUILD_RC=1
```

**Do not merge on the strength of my "everything is attributed" comment until this is
resolved.** That comment was written before this result existed.

What I can say so far, and what I cannot:

- The last per-target diagnostic line before the verdict is `scripts/initiatives/tests`,
  which is consistent with the pre-existing failure I reproduced identically on clean
  `origin/main` (9 failed / 775 passed on both). **Consistent is not the same as shown.**
- nix truncated the builder output and no log was retained (`nix log` and
  `nix-store --read-log` both return nothing), so I do **not** yet have the failing test
  names from the hermetic run.
- I am re-running it with `-L` to stream the full builder output and get those names.

The two outcomes and what each would mean:

1. the failures are the same nine `test_archive` / `test_viewer` / `test_sync` cases →
   pre-existing, the gate is red on `origin/main` too, and this PR is not the cause; or
2. anything else → it is mine, and this PR is not ready.

I will post the answer either way. Until then treat the gate status on this PR as
**UNRESOLVED**, not green.

ZacxDev:
## Update 3 — retracting part of Update 1, and what the hermetic toolchain actually says

🔴 **Retraction.** Update 1 said the nine `scripts/initiatives/tests` failures were
"pre-existing", on the strength of reproducing them identically on clean `origin/main`.
That reasoning was sound but it measured the wrong thing: **both runs shared my
under-provisioned shell**, so what reproduced was my own missing `psycopg2`, not a repo
defect. Run under the real toolchain they pass:

```
nix develop .#checks.x86_64-linux.pytests -c pytest scripts/initiatives/tests -q
→ 784 passed in 20.73s
```

A comparison is only a control if the two sides differ in the one variable you are
testing. Mine differed in nothing.

**What the hermetic toolchain does say, measured:**

| target | hermetic result |
|---|---|
| `scripts/browser-bridge/tests` — the only code this PR touches | **517 passed** (3:04) |
| `scripts/initiatives/tests` | **784 passed** |
| node tier (whole repo) | **1116 passed**, 0 failed |

**My prime suspect is ruled out by direct probe, not by argument.** The new pty test
needs `pty.openpty()`, and a nix build sandbox is exactly where that could be
unavailable. So I probed the sandbox itself:

```nix
runCommand "pty-sandbox-probe" { nativeBuildInputs = [ python3 ]; } ''python3 "$probe"''
→ SANDBOX_PTY_RESULT: TTY rc: 0
```

It works, and it correctly reports TTY rather than degrading to a pipe.

**Still unresolved, and I am not hand-waving it:** `nix build
.#checks.x86_64-linux.pytests` fails, and I have not localized which target. nix
truncated the builder output and retained no log (`nix log` and `nix-store --read-log`
both empty), so I re-ran it with `-L` and, in parallel, `run-tests.sh` under `nix develop`
writing to a file I control. Neither had finished when I ran out of window.

So: **every target I have been able to measure hermetically passes, including mine — but
the full gate is red and unattributed.** Treat this PR's gate status as UNRESOLVED. The
one command that settles it:

```
nix build .#checks.x86_64-linux.pytests -L 2>&1 | tee /tmp/pytests.log
```

If it names a target other than `scripts/browser-bridge/tests`, compare against
`origin/main` **in the same hermetic shell** — which is the control I failed to run the
first time.

ZacxDev:
## ✅ RESOLVED — both hermetic CI gates pass. The earlier red was a load flake of my own making.

```
nix build .#checks.x86_64-linux.pytests  → /nix/store/x7yr9yxz7lh6z2yjqyzqrivyx9jp0l7q-devrc-pytests   rc=0
nix build .#checks.x86_64-linux.nodetests → /nix/store/bgfm2y0jcy06qak0n2mjdi1hipjxv3hw-devrc-nodetests  rc=0
```

Both produced real output paths, so these are genuine builds, not cached results.

**What actually happened.** The first `pytests` build ran while I had an ad-hoc
`scripts/tests` run (423 s), a merged-tree suite, and a mutation battery going on the same
box. It went red. `claude/RULES.md` says to distinguish a real failure from a load flake by
wall time rather than by a single re-run — and better than a re-run, the per-target
hermetic measurements now corroborate the green from a second, differently-constructed
direction:

| target, under `nix develop .#checks.x86_64-linux.pytests` | result |
|---|---|
| `scripts/tests` (the repo-wide gates: captured-text, markup, hostnames, shebangs, skill/RULES size) | **5303 passed** |
| `scripts/browser-bridge/tests` — the only code this PR touches | **517 passed** |
| `scripts/initiatives/tests` | **784 passed** |
| `scripts/collector/{,keylog,claude,i3,browser-ext,opencode}/tests` | 273 / 83 / 137 / 13 / 13 / 203 passed |
| node tier (whole repo) | **1116 passed**, 0 failed |

So every failure I reported in the three comments above was an artifact of my own
under-provisioned shell or of load, and **none was a defect in this branch**:

- `scripts/tests` 25 failures → missing `logrotate` / pinned `opencode`; **5303 pass hermetically**
- `scripts/signal`, `scripts/mail-actions` collection errors → missing `psycopg2` / deps
- `scripts/initiatives` 9 failures → missing `psycopg2`; **784 pass hermetically**
- the `pytests` build itself → load; **passes on a quiet box**

**Corrections to my own earlier comments, kept visible rather than edited away:** Update 1
called the `initiatives` failures "pre-existing, reproduced on `origin/main`" — void, because
both sides of that comparison shared the same broken shell, so it controlled for nothing.
Update 2 said the hermetic check "never returned"; it did, and it failed, which Update 3
flagged as UNRESOLVED and do-not-merge. This comment supersedes all three.

The gate is green. The verification state that still stands is in the PR body: **nothing is
deployed** (`server.py` is a `home.file` copy — inert until `home-manager switch`), there was
**no live reproduction**, and the **`nav` residual** (30/627 vs 9/627 placebo, activate-adjacent
removed) is real, unexplained, and not addressed here.

PR #558, opened 2026-08-19
feat(session-resolve): read-only session identity resolver — one selector to one target, with a three-state presence trichotomy

ZacxDev:
## What this adds

`scripts/session-resolve` — a **READ-ONLY** resolver turning any human- or agent-shaped selector
into ONE fully-resolved target, plus `scripts/tests/test_session_resolve.py` (120 hermetic tests).

```
session-resolve show <selector> [--json] [--pr] [--host all]
```

Selectors: codename (`Gold`) · hotkey (`G`, `Alt+Shift+G`) · `session:index` · address
(`scratch2:@81`) · window id (`@81`) · harness name · cwd/repo · session name.
**Ambiguity is a first-class refusal**: candidates are listed, never silently picked.

## 🔴 Also folded in: a fix to `scripts/waiting-windows`, which this branch does NOT own

`waiting-windows` merged as **#559** (squash `7030142`) and arrived here with `origin/main`.
The fix is landed on this branch **deliberately**, to avoid a third conflict on the single
`scripts/run-tests.sh` floor line that both branches already fought over. Flagging it so a
reviewer is not surprised to find a second script touched.

**The defect, measured live on the merged tool:** 24 rows over threshold from a population of 46
(`idle_no_signal=20, selection_menu=2, context_exhausted=2, trailing_question=0`), sorted purely
by waited time — putting the actionable rows at **rank 8 and rank 20 of 24**. Twenty idle windows
(titles like "Release deployed", i.e. finished work) buried the two things actually asking a
human, and `--top` would have truncated the actionable rows *away* precisely because they ranked
8th and 20th. There was no workaround: no kind filter, no sort control.

**Fixed by:** primary sort on a kind band, secondary oldest-first within the band; a repeatable /
comma-separated `--kind` filter (unknown kind exits 2, and the always-print-what-was-dropped
contract is kept); and a counts line led by the actionable total. `idle_no_signal` is
**reordered, never filtered** — the operator chose the wide population, and that bucket cannot
tell blocked from finished. Its caveat is untouched and still pinned as a whole string.

🔴 `KIND_DISPLAY_BANDS` **deliberately disagrees** with `SIGNAL_PRECEDENCE`, which ranks
`context_exhausted` the hardest *block*. They answer different questions — "what blocks hardest"
vs "what should a human look at first" — so a test pins the divergence, because a reader will
otherwise "correct" it.

**Live after:** `ACTIONABLE: 6 of 24`, actionable rows at ranks 1–6.

## 🔴 A measured refutation: `--lean` cannot supply the join key

The brief said to consume `session-manager --json --lean` **and** join on
`(host, session, window_id)`. Incompatible: `--lean`'s `lean_row_fields` carries `window_index`
only and **drops `window_id`**, so the mandated key is absent from the named view. This consumes
the full `--json` view (a strict superset for every requested field); a test pins `--lean` out of
`SM_ARGV`. Join verified **24/24**, zero unmatched, zero duplicate addresses.

## 🔴 Four presence states, not three

The harness registry lives at `~/.claude/sessions/` on the **local** filesystem — no host
dimension in that path — so a **remote** window can never carry a record. That is a fourth state,
and collapsing it into "absent" tells a caller "none (normal)" and invites it to re-check for a
clock that can never arrive.

| state | meaning |
|---|---|
| `present` | a record joined |
| `absent` | registry read, this window genuinely has none (the majority case) |
| `unmeasured` | the registry could not be read, with a reason |
| `structurally_excluded` | remote host — impossible on any future run |

Excluded hosts are named in **`coverage.registry.hosts_not_covered`**, matching the
waiting-windows spelling exactly (two PRs disagreeing on the vocabulary for one fact is its own
defect). The same exclusion applies to every local-only source: a remote target's git branch, PR
and tmux visibility are excluded/unmeasured, never measured against this machine's filesystem —
running `git -C` on a remote path could otherwise **succeed against a same-named local checkout**
and report a branch belonging to something else.

**Live cross-host:** 87 windows (58 local + 29 remote), 29 structurally excluded,
`local_host=workbench`, taken from session-manager's own `local_host` rather than the `--host`
argument (`--host all` is not a host name).

## 🔴 The join key gained its host

`(session, window_id)` was keyed without host. Measured across both machines: **14 session names
and 11 window_ids exist on both**, with laptop's id range (@4–@31) sitting inside workbench's. No
full pair collided at that instant — but zero collisions is an empty result, not proof, and a
collision is one window-open away. Now keyed `(host, session, window_id)`.

## Two statuses, never collapsed

`harness_status` and `waiting_probable` are carried separately with **no mapping invented**; a
structural test fails if any key named `blocked_on_me`/`is_waiting`/`combined_status` appears.
Live confirmation of the disagreement: `scratch11:@136` reports `harness_status: idle` **and**
`waiting_probable: True`. The registry's `waitingFor` reason is carried verbatim beside its
status. Its `status` vocabulary is **unstable** (two snapshots hours apart gave `{busy, idle,
shell, waiting}` then `{busy, idle, waiting}`), so it is passed through and never branched on —
enforced structurally over the syntax tree, not by substring scan.

## Read-only, enforced

An **allowlist** of exactly `list-panes`/`list-windows`/`list-clients` (a denylist is walkable by
any verb nobody thought to ban). Both controls tested: ten write verbs refused, and the three read
verbs permitted — a guard that refused everything would pass the first test while breaking the
tool.

## Tests

**120** hermetic tests for session-resolve, **+19** for the waiting-windows ordering/filter.
Every impure source injectable; autouse fixtures replace the subprocess seam and default paths
with raisers, and a positive control asserts those fixtures really are installed and really raise.

Fixture values are pairwise distinct and distinct from every asserted constant. The join fixture
**crosses ids and indexes** (`@501` at index `502`, `@502` at index `501`) so an index join still
*matches* — just the wrong window; that is the only shape that can tell the two joins apart. The
ordering fixture makes the **oldest** row the **least** actionable (196103s idle vs 7817s
trailing_question, a 25× gap) so a mutant sorting on age alone cannot pass.

### Mutation matrices — 20/20 and 11/11, each killed by its own named assertion

session-resolve: join-on-index · ISO-only epoch parse · ambiguity-picks-first · popup-by-width ·
`covered := visible` · harness absent→unmeasured · gh `[]`→unmeasured · failed-gh→"no PR" · null
`waiting_probable`→False · lean-rows-joined-on-index · hotkey lower-cased · allowlist→denylist ·
padded rows · empty clients→unmeasured · missing timestamp→epoch 0 · **remote→ABSENT** ·
**join key drops host** · **git on a remote cwd** · **hosts_not_covered emptied** ·
**EXCLUDED spelled with `unmeasured()`** · **remote visibility measured**.

waiting-windows: band dropped from the sort key · band order reversed · `context_exhausted`
promoted · oldest-first lost within a band · idle filtered instead of ranked · silent filter drop ·
filter after `--top` · actionable counts the idle bucket · no-op filter reports None ·
`parse_kinds` stops splitting · `ACTIONABLE_KINDS` includes idle.

**Both sweep harnesses were themselves validated** — a cosmetic mutant correctly reported
SURVIVED (proving they *can*), and a **same-length** edit (`> 1:` → `> 9:`) was correctly KILLED,
proving the stale-bytecode trap is closed. Run under `PYTHONDONTWRITEBYTECODE=1` with
`__pycache__` purged between mutants; 0 `__pycache__` written; sources restored and SHA256-verified.

## On-disk artifacts: none

The PR cache is an in-memory dict on `Sources`. A test pins the complete artifact set as **empty**.

## The floor line: resolved by MEASURING the merged tree

Both branches edited `scripts/tests|` from the same 4976 base — waiting-windows to **5482**,
session-resolve to **5355** — so neither described the tree the merge creates, and adding the
deltas is the arithmetic-across-a-conflict the header forbids. Re-measured by the gate on the
merged tree: **collected=5722 → 5672**, then **5741 → 5691** after folding in the waiting-windows
tests. Both narratives kept in the comment block. `--check-targets` and `--check-floors` PASS.

## Gate

Per-target on the merged tree: `scripts/tests (collected=5722 passed=5721 failed=1)` with **29
other targets PASS**, the single failure being `test_opencode_engine.py`
(`'1.18.18' == '1.18.16'`) — pre-existing on this host, **0 lines changed** by this branch. Final
gate re-run after the waiting-windows fold.

A note on one thing I did *not* act on: an earlier claim that "the registry clock is unused, 28/28
on the fallback" was retracted as a measurement artifact of `--no-state`. Corrected:
`registry_status_updated=9, ledger_age_fallback=15`. Nothing was changed on the strength of the
original claim, and the live run above shows `waited_source=registry_status_updated` in use.


ZacxDev:
## 🔴 Two corrections to the PR body above — it was wrong, and a reviewer may already have believed it

Posting these as a comment rather than silently editing the description, because a
reviewer who already read the original would never see a quiet fix.

### Correction 1 — the hermeticity claim was FALSE, and the suite really was leaking

The body said:

> autouse fixtures replace the subprocess seam and default paths with raisers, and a
> positive control asserts those fixtures really are installed and really raise

**The "default paths" half of that was false.** `Sources.registry_dir` and `Sources.runner`
were dataclass **field defaults**, bound once at class-creation time, so
`monkeypatch.setattr(sr, "DEFAULT_REGISTRY_DIR", …)` could never reach them. The positive
control asserted the module global — the one value nothing on that path read — so it was a
control that could not observe the thing it guarded.

The audit called this "a false claim plus a dead guard rather than live contamination". **I
measured it, and it was worse than that: it was live contamination.** Replaying the suite's
own autouse fixture against the audited tip `31230e8`:

```
raiser saw calls: 0
REAL subprocess.run invocations: [['git', '-C', '/w/repo-alpha', 'branch', '--show-current']]
```

Every test that built targets spawned a real `git` on the operator's machine. Harmless in
effect — the fixture paths don't exist, so git just failed — but the suite was **not
hermetic**, and no control could see it.

Fixed: the defaults resolve at call time, the control asserts what a bare `Sources()`
actually yields (including that `load_registry(Sources())` returns zero records rather than
the 48 live ones it used to), and a `subprocess.run` backstop now makes the next leak of this
shape red whatever route it takes.

### Correction 2 — the waiting-windows "24 rows" figure is from a DIFFERENT instant than the one I quoted

The body justifies the reordering with a measurement, and reports the after-state as
`ACTIONABLE: 6 of 24`. The brief I was working from gave the before-state as
`idle_no_signal=20, selection_menu=2, context_exhausted=2, trailing_question=0` — which is
**4** actionable, not 6.

Both numbers are real; they are **two different instants** of a live system, taken roughly an
hour apart. The `24` total coinciding is a coincidence of the threshold, not the same
snapshot. Since the entire rationale for the change rests on that ratio, here is the
after-state stated with its instant, re-measured on the fixed tree:

```
2026-08-19 ~11:50 local, threshold 4h00m (14400s, default)
ACTIONABLE: 6 of 24 over threshold
  (18 idle_no_signal — ranked last; cannot tell blocked from finished)
  by kind: context_exhausted=2, selection_menu=2, trailing_question=2, idle_no_signal=18
```

The load-bearing claim is unchanged and holds at both instants: sorted by waited time alone
the actionable rows landed at **rank 8 and rank 20 of 24**, so `--top` would have truncated
them away; banded, they occupy ranks 1–6.

*(The mismatch originated in the brief I was given, not in my own measurement — flagging it
here rather than absorbing it, since the number is load-bearing either way.)*

---

## Audit fix round — what changed, per finding

| # | finding | fix |
|---|---|---|
| 🔴 1 | `--host` defaulted to literal `"workbench"`; no `choices=` | `default="all"` + `choices=("all", *HOST_NAMES)`, pinned against session-manager's `HOST_NAMES` by a seam test; hosts consulted now printed on **every** outcome incl. unmatched |
| 🟡 2 | hermeticity fixture couldn't reach field defaults | defaults resolve at call time; control asserts a bare `Sources()`; `subprocess.run` backstop added |
| 🟡 3-4 | `main()` untested in both scripts (9 surviving mutants) | arg→`Sources` mapping tests, exit-code **mapping** (not just distinctness), end-to-end `main()` tests via `--json-file` |
| 🟡 5 | `"all"` sentinel could become a host name | exclusion keys off an `is_local` flag set where the fact is known, never a string compare; `hosts_not_covered` is now UNMEASURED (not a list accusing every host) when the local host is unknown |
| 🟡 6 | read-only scan stopped after the first command | segments on tmux's `;` and checks each segment's command; positive control proves the tool's own `-F <fmt>` calls still pass |
| 🟡 7 | `HARNESS_EXCLUDED` stated the inverse of the truth | reworded to name the REMOTE host; the whole-string pin now asserts `"on this host can never"` is **absent** |
| 🟡 11 | two sources encoded the trichotomy differently | a measured absence is `MEASURED_OK` in both; the two spellings pinned against each other |
| opt | `unsent_prompt` printed untruncated | truncated to 90 like session-manager, and it says how much was cut |
| opt | dead branch in `apply_kind_filter` | merged into one return |
| opt | `f"{None:<3}"` latent TypeError | renders `idx=?` |

**Gate reset and re-run in full**, per the rule that a review fix resets verification — not
just the mutants for what I touched:

- session-resolve: **30/30** mutants killed by their own named assertion
- waiting-windows: **16/16**
- Two mutants initially reported SURVIVED. Both turned out to be **badly constructed**, not
  coverage gaps: one replaced a line while the *next* line still emitted the hosts, and one
  inserted an assignment that the real assignment two lines later overwrote. Isolated
  properly, both then died. Recording that here because "SURVIVED" was the wrong answer for
  the wrong reason, and a sweep whose mutants don't take effect is a sweep that proves nothing.

Floor re-pinned 5691 → **5730** from `_suggested_floor 5780` on this tree.

ZacxDev:
## 🔴 Correcting an overstated claim I made in this thread

Posting rather than editing, for the same reason as last time: a reviewer who
already read the wrong version would never see a quiet fix.

### The "144 cells" number was overstated

I described the host-disclosure matrix as **"144 constructible cells"**. The
generator does yield 144 cells, but they are not 144 *distinct* tests:
`sm_broken=True` discards `sm_reports` entirely (6× duplication) and `tmux_ok`
is inert in every cell (2×). Measured on the current tree:

| | count |
|---|---|
| cells generated | 144 |
| distinct whole-cell signatures | **26** |
| distinct `(host, status, reason)` outcomes | **16** |
| statuses reached | all 5 |

So the honest description is **26 distinct cell outcomes over 16 distinct
per-host verdicts**, not 144. The matrix is still load-bearing — it killed five
mutants this round and is what caught the `answered=True, in_scope=False`
incoherence — but the number as I stated it was wrong, and it was the number I
used to argue the matrix was thorough.

*(The review that caught this said 26 `(host, status, reason)` outcomes; I
measured both figures and they are 26 signatures / 16 per-host outcomes. Stating
both so nobody has to re-derive it.)*

### Two things that PR body language implies are pinned, and are not

- **`ALL_HOST_STATUSES`' string values are a `--json` contract** — `read`,
  `unreachable`, `not-reported`, `source-failed`, `out-of-scope` are consumed by
  anything parsing the output. Nothing pins the literals; renaming one breaks
  consumers with a green suite.
- **The generated reason prose is unpinned.** Rewording the `source-failed`
  reason, dropping the session-manager failure detail, and reversing the
  `answered_by` order all survive. This repo's rule for a prose artifact is to
  pin the whole normalised string; these are asserted by substring instead.

Both are real gaps. I have **not** fixed them in this round — the round was
scoped to two fixes and I was asked not to expand it — so recording them here
rather than leaving the PR body implying coverage that does not exist.

### Also filed, not fixed

The tmux `source_failed` path is never exercised: `tmux_ok` is `True` in every
matrix cell, so every failure status in the matrix comes from session-manager
alone, and wrong-operand/always-False mutants on the tmux branch survive.

---

## This round's two fixes

**1. The partial-view caveat fired on `answered`, not `measurable`** — the
fourth variant of this surface's recurring bug, this time in the caveat rather
than the status. A source that answers "unreachable" *has* answered, so with
session-manager reporting a host unreachable while tmux read it, the row printed
`read | read via tmux` with that host's session-manager view entirely missing —
surviving only in `--json` `sources`, which is the one-view-only gap this PR
closes in `waiting-windows`. It was also **completely unasserted**: deleting the
caveat survived, and so did retargeting it.

**2. `test_a_single_host_flag_really_does_hide_the_other_machine` was vacuous** —
it named a synthetic host that never appears in `hosts_consulted` under a single
`--host`, so both assertions ranged over an empty set.

**Verification, proportionate to a two-predicate change:** the four named
mutants watched to die individually — caveat deleted, caveat retargeted to
`answered`, caveat always-firing, single-host fixture emptied — each by its own
named test, plus the full gate. The full batteries were **not** re-run; nothing
they cover was touched, and I am saying so rather than implying a sweep I did
not do.

PR #560, opened 2026-08-19
fix(subsystem-recall): a near-miss marker and a renamed heading were BYTE-IDENTICAL to a clean index row — surface both

ZacxDev:
Implements **items 2 and 1** of `claudedocs/proposal-entry-shape-explicit.md` §4.2 (merged as #550). Items 3, 4 and 5 are deliberately **out of scope**.

Both close a live silent-failure path on the index row — the one surface `/resume` step 4 reads every session. **No new front-matter field and no store change**: every number added below was already computed on every bullet and thrown away.

---

## The two failures, reproduced

They share one observable — a row reading `0 nuance` with no badge — which is exactly why neither was visible.

**§2.4 — a near-miss marker is byte-identical to no marker.** A bullet that *tried* to write an openness marker and missed the strict grammar (`**OPEN:**`, `RESOLVED <sha> (repo):`, a date without its colon) declares nothing, so the `🔴 N OPEN` badge simply does not render — and a vanishing badge *looks like success*. The advisory that catches it lives only in `subsystem_touch --validate`, which `/resume` never runs. So the population **most likely to hold a stale open action** was the one no routine surface reported.

**§2.2 — a renamed nuance heading is silent data loss.** Heading matching is exact-string at column 0, so a renamed `## Nuance / work-history` yields an empty section: the bullet count reads 0, the `OPEN` badge disappears, and `--validate` returns `OK` at exit 0. `missing_sections` already existed and already travelled in JSON — it was rendered only under a **printed body**, and the digest prints exactly one body out of N.

Differential control, two synthetic entries differing in **nothing** but the nuance heading, the same `OPEN:` marker in both:

```
BEFORE                                                    AFTER
  orders-api      0 nuance   public                         orders-api      0 nuance   public   🔴 NO Nuance / work-history
  payments-api    2 nuance   public   🔴 1 OPEN             payments-api    2 nuance   public   🔴 1 OPEN
  ledger-api      6 nuance   public   🔴 1 OPEN             ledger-api      6 nuance   public   🔴 1 OPEN   🔴 2 NEAR-MISS   ⚠ 1 UNVERIFIABLE
```

---

## Design

| decision | why |
|---|---|
| **Rendered directly, no opt-in flag** | `/resume` consumes this row every session and would never pass one. Opt-in dies when it *is* the ritual. |
| **Conditional, exactly like `🔴 N OPEN`** | 52 of the 53 live rows stay byte-identical (measured, below). |
| Badge order = `--validate`'s order (declared → near-miss → unverifiable) | the two surfaces read the same way. `NO <heading>` sits last: it is not a bullet population, it says the parser never reached the bullets. |
| Every count off `JournalBullet.openness_population` | the resolver's single source of the precedence order — **including `open_count`**, which used to read `b.is_open` for itself. A predicate open-coded at N sites is wrong at N−1. |
| The **unmarked-action floor is still absent** from the row | unknown recall, and a ~60 B line cannot carry that caveat. `--validate` reports it and can. |
| Text row only, following `open_count` | the JSON `listing` rows carry no openness field of any kind; splitting the row's vocabulary across two payloads is how the two start to disagree. |
| The caveat names all three badges **and what each implies about the numbers beside it** | a badge the caveat does not name is a glyph the reader guesses at. |

---

## Verification

**Common case unchanged — measured against the real store, read-only.** Every one of the 53 live entries rendered through both the pre-change and post-change libraries: **52 rows byte-identical, 1 changed**, and the one that changed gains `🔴 2 NEAR-MISS` for a genuine near-miss pair. Also pinned in-suite by `TestTheCommonCaseRowIsUnchanged`, which spells the pre-change format out rather than lifting it from the renderer.

**Red → green.** The full test file against the pre-change tree (`ecddbae` lib + both skill docs): **24 failed / 378 passed**. Against `HEAD`: **402 passed**. Of the 24, 20 are the new guards and doc pins; 4 are pre-existing `TestListingLineOpenAnnotation` cases that go red because the shared `_recalled` helper now populates every field the row can render (it previously set `open_count` alone, so the byte-identical pin was silently asserting about defaults).

**Mutation battery — 14 mutants, 14 KILLED**, each isolated to the narrowest expression that can be wrong, each verified to fail with *this* guard's own assertion rather than a neighbour's:

| mutant | verdict |
|---|---|
| drop the near-miss badge clause | KILLED |
| drop the unverifiable badge clause | KILLED |
| drop the missing-sections badge clause | KILLED |
| render the near-miss badge unconditionally | KILLED (by the common-case pin) |
| `short_heading` keeps the ATX marker | KILLED |
| `near_miss_count` never populated | KILLED |
| `unverifiable_count` never populated | KILLED |
| the two counts swapped | KILLED |
| near-miss folded together with unverifiable | KILLED |
| `missing_sections` forced empty | KILLED |
| caveat stops naming `NEAR-MISS` | KILLED |
| caveat stops saying the counts are parse failures | KILLED |
| caveat stops saying `OPEN` is undercounted | KILLED |
| caveat drops the whole three-badge clause | KILLED |

A fifteenth (`Three further ` → ``) is recorded as **EQUIVALENT** — a cosmetic reword that still names all three badges — and is *not* counted as a kill. The battery prints its own positive control (402 passed unmutated) before any verdict, runs under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` deleted between mutants, and refuses to score a mutant whose pattern does not occur exactly once. Fixture counts on the multi-population case are **pairwise distinct** (1 open / 2 near-miss / 3 unverifiable) so no two badges can be swapped and stay green.

---

## Contradiction with the proposal, declared

§2.4 states *"2 of the 10 textual `OPEN:` markers do not parse"*. Re-measured 2026-08-19 over the live store (53 entries, 323 top-level nuance bullets): **the near-miss count of 2 reproduces exactly** (one `OPEN`-shaped, one `RESOLVED`-shaped) alongside 8 declaring `OPEN:` and 11 declaring `RESOLVED <sha>:`. **The denominator does not** — a raw `grep -o 'OPEN:'` returns **11**, not 10, and not every occurrence leads a top-level bullet. The rate is therefore quoted in the source as the two populations rather than as a ratio, and the discrepancy is recorded in `near_miss_count`'s docstring so it is not silently reconciled later.

Two further live measurements worth having on record:

- **`unverifiable` is 0 across all 53 entries today**, so that badge does not fire on the current store. It ships anyway because a sha-less `RESOLVED` is one hurried write away and it costs nothing when zero — but per §4.2's own falsification clause, if it is still permanently absent in a month it should be deleted rather than kept as a never-green indicator.
- **`missing_sections` is 0 across all 53 entries**, independently confirming the 53/53 spine.

---

## Store safety

The store was **never written**. Every fixture is synthetic, under `tmp_path` or passed via `--store`, with invented names (`payments-api`, `orders-api`, `ledger-api`, `example-scope`). No git command ran in any scope directory. The read-only measurements above print counts and a redacted row shape only. All four content gates were run **after staging** (they read `git ls-files`): 227 passed.


ZacxDev:
**Gate verdict — read from `nix log` CONTENT, not an exit code.**

`nix build .#checks.x86_64-linux.pytests --no-link` → `/nix/store/6rhjkpsqn132cv3fnh06hp373jpbyfsn-devrc-pytests.drv`:

```
TOTAL collected=12054  passed=12053  skipped=1  failed=0  (floor: 11112 = sum of 24 per-target floors)
RESULT: PASS (exit=0)
```

Against the stated baseline on `main` (`collected=12031 passed=12030 skipped=1 failed=0`): **+23 collected**, which is exactly the 20 new test methods plus the 3 new parametrized `RESUME_SENTENCES` doc pins. No per-target floor drifted, so nothing needed a replacement number.

Content gates (`test_no_captured_text`, `test_no_captured_markup`, `test_no_public_ips`, `test_no_client_hostnames`) run **after staging**, since they read `git ls-files`: 227 passed. Noted for the record: the two IP/hostname gates have no `__main__` block, so they were run under pytest and never as scripts.

ZacxDev:
Node tier also green (read from log content, same derivation-log method):

```
TOTAL suites=4 files=33 tests=1116 pass=1116 fail=0 skipped=0  (floor: 1098)
RESULT: PASS (exit=0)
```

This diff touches no JS, so the node tier is an unchanged-baseline confirmation rather than coverage of the change — recorded because `CLAUDE.md` gates on both tiers and a green read on one tier says nothing about the other.

PR #566, opened 2026-08-19
docs(browser): prune SKILL.md for real headroom + fix 3 correctness defects

ZacxDev:
## Why

`scripts/browser-bridge/SKILL.md` sat at **12,035 B against an enforced budget of 12,038** — 3 bytes. The enforced budget is `MAX_BYTES 12,288 − MIN_HEADROOM_BYTES 250`, both owned by `scripts/browser-bridge/tests/test_skill_size.py`. The next one-line correction would have failed the gate.

**12,035 → 11,880 B · headroom 253 → 408.**

Note `scripts/skill-audit.py` reports this file as `✓ OK — no prune needed`. That verdict is measured against the raw 12,288 ceiling; it cannot see `MIN_HEADROOM_BYTES`, so it is a false all-clear for this skill.

## What was removed

663 B of verified-redundant content; every removal confirmed to exist in a destination **first**, with the line reference:

| removed from core | already lives at |
|---|---|
| agent result handling, thin-`evidence`, `--allow-domains` | `agent.md:151,132,96-105,119` |
| wake settle+cap, once-per-page, ISOLATED-vs-MAIN | `spa-wake.md:63,78-80,216,218-219` |
| `text` byte cap + `--annotated` schema | `read-envelopes.md:59,62-87` |
| `extension_stale` tri-state, staleness-is-per-profile | `errors.md:148-165,159` |
| re-`open` discards the url | `tabs-instances.md:28-29` |
| `emulate --reset/--recreate`, viewport (#319) | `emulation.md:22-23,112` |
| screenshot 133K–890K token evidence | `README.md:303` |

Plus one MERGE_DUP: `click`/`type`/`key` restated the same trust rule three times → one input-ops row.

## Three correctness defects fixed

These **added** 485 B. Correctness outranks bytes.

1. **Triage item 1 was wrong in both branches.** It said the reload arrow is unreliable so prefer a full Brave restart. `errors.md:82-99` (mid-session **drop**) says reloading *in the driven profile* IS the fix; `errors.md:150-165` (**stale build**) says a full restart is explicitly NOT sufficient and needs per-profile Remove + Load unpacked. The body recommended the one thing neither branch supports. This was flagged as an open question in `claudedocs/handoff-browser-per-site-reference.md` — resolved: they are different failures, not a contradiction, and the body was simply wrong.
2. **"Restore focus" did not cover the workspace.** `claude/RULES.md` is explicit that a restore rule about focus does not cover a workspace, and that the workspace is the axis that actually gets taken. Now names both, gives the record *and* restore command for each, and says "on failure too".
3. **A routing path that could not resolve.** The README was referenced bare; only `SKILL.md` + the CLI are symlinked into `~/.claude/skills/browser/`, so it did not resolve for a reader. Now repo-absolute. Related: the `reference/` path expansion sat *below* every inline mention — hoisted into Quick start so it precedes them.

## Staleness pass — deterministic axes clean

24/24 ops, 16/16 envelope fields/flags, 13/13 error strings, and every numeric claim (32768 cap, 0600 temp, 24h prune, 1.5s settle, 6s wake cap) verified against the CLI + server + extension.

One **false finding**, recorded so nobody re-derives it: `nav_scheme_denied` / `op_not_allowed` look absent from source. They are not — `opencode/tools/browser_tool_impl.mjs:482,516,1101,1108`. The first scan's file filter took `.py/.js/.json` and not `.mjs`. Instrument fault, not doc rot.

## Verification

- **browser-bridge suite: 654 passed, 0 failed.**
- **Negative control on the size gate**: padded a scratch copy to 12,080 B and watched `test_skill_md_keeps_working_headroom` fail with `free: 208 (minimum required: 250) / RECLAIM: 42`. Green at 11,880 is a measurement, not an assumption.
- **Survival check** over (new core + all sidecars): numbers population removed=3, **orphaned=0**. Destination controls — hide each of 6 sidecars, mutate a constant in `spa-wake.md` and `read-envelopes.md` — each went red.
  The phrases/sentences populations are **not** reported as findings: they test verbatim strings across a rewrite and their controls did not move (35 vs 35/36), so they are an invalid instrument here.
- Rollback backup: `/tmp/skill-prune-browser-140452` (14 files).

## What I deliberately did not do

No further cutting. Every section now sits at or below its own historical density (Ops 188 B/row against a measured 190 mean; Reference 175 B/row against 166). There is no archive block left to demote — **the body is dense, not bloated**, which is why the margin was tight. Cutting further would remove routing value, not fat. If more headroom is wanted later, that is a decision to raise `MAX_BYTES` or to demote one of the four 🔴 traps, not something to take silently.

### DIFF
--- a/scripts/browser-bridge/SKILL.md
+++ b/scripts/browser-bridge/SKILL.md
@@ -116,9 +116,9 @@ Result payloads land under `.result.data`.
 It's their real browser, not a scratch VM. Don't `nav` a tab that may hold unsaved
 work (a half-typed comment, a form) — `open` your own tab, or an obviously
 disposable one. 🔴 If ANYTHING takes their screen — `activate`, the X-fallback
-capture — RECORD focus AND workspace first, then restore BOTH at the end, on
-failure too. Focus alone leaves them on YOUR workspace, which is the axis that
-actually gets taken. Exact commands → `reference/spa-wake.md`.
+capture — RECORD focus AND workspace first, restore BOTH at the end, on failure
+too. Restoring focus usually carries the workspace, but not always — restore it
+explicitly. Why, and the commands → `reference/spa-wake.md`.
 
 ## Reference files — load ONE only when its trigger fires
 
--- a/scripts/browser-bridge/SKILL.md
+++ b/scripts/browser-bridge/SKILL.md
@@ -15,7 +15,11 @@ $BB --instance <key> --tab <id> text   # cheap read of a specific tab
 🔴 **Run `whoami` first on every fresh browser task.** Both hosts are hostname
 `nixos` and this bridge could be either, with several Brave profiles — confirm the
 host and pick the right `--instance` before acting. Architecture / security model:
-`scripts/browser-bridge/README.md`.
+`~/workspace/devrc/scripts/browser-bridge/README.md`.
+
+🔴 **Every `reference/<file>.md` named below lives at
+`~/workspace/devrc/scripts/browser-bridge/reference/`** — that exact path; only
+`SKILL.md` + the CLI are symlinked into `~/.claude/skills/browser/`.
 
 ## FIRST DECISION: agent or direct?
 
@@ -33,15 +37,13 @@ order) · **secret** — agent-read pages go to
 **OpenRouter/DeepSeek**: never banking, private mail, credential managers, or
 anything you wouldn't hand a third party. Nor **virtualised/lazy-loaded lists**.
 
-KNOW something from it → agent. Ambiguous → agent first; taking over costs ~200
-tokens, so agent-first wins even at a low success rate.
+KNOW something from it → agent. Ambiguous → agent first: taking over is cheap, so
+agent-first wins even at a low success rate.
 
-`blocked` → non-zero exit, take over. `partial` → read `evidence`, drive the rest.
-Thin `evidence` is a weak smell, **NOT** protection against a wrong answer. What
-protects you: auto-`wake` on a hidden `text`/`html` read — **never `eval`/`js`, so
-ASSERT a non-zero content count in any `js` measurement** — and escalation: if the
-answer depends on rendered SPA state, ONE `text --wake`. Keep `--allow-domains`
-TIGHT, incl. frame hosts.
+🔴 Auto-`wake` covers a hidden `text`/`html` read but **never `eval`/`js`** — so
+ASSERT a non-zero content count in any `js` measurement. Result handling
+(`blocked`/`partial`), why thin `evidence` is NOT protection, and the
+`--allow-domains` guardrail → `reference/agent.md`.
 
 ## Ops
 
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Correcting the verification claim in the PR body, and the audit round

The body's verification section was written from the targeted suite alone. That was true but **insufficient** — I had not run the repo's real gate at that point. Correcting it here rather than editing the body, since it may already have been read.

### The real gate is RED — but it is red on `main` too

`nix build .#checks.x86_64-linux.pytests` on this branch: `RESULT: FAIL (exit=1)`.

Note the shell reported **exit 0** for that run — the trailing `tail` swallowed the status. The RESULT line is the verdict, never the exit code.

I ran the same gate on unmodified `origin/main` (`70f811a`) as a control. **Identical failure set:**

| target | failure | on this branch | on `origin/main` |
|---|---|---|---|
| `scripts/tests` | `opencode on PATH is '1.18.18'` vs `'1.18.16'` pinned | ✗ | ✗ |
| `scripts/browser-bridge/tests` | `collected=654 above drift ceiling 586` | ✗ | ✗ |

Both are pre-existing and neither is attributable to this PR, which changes documentation only and adds **zero** tests. The second is a test-**count** guard, not a test failure — all 654 pass.

Two follow-ups worth filing separately (not this PR's job): the opencode pin looks like flake drift, plausibly from the recent `flake.lock` bump; and the browser-bridge drift ceiling has been outgrown (654 vs 586) — the handoff records `RESULT: PASS` at #562's tip, so `main` went red after that.

The branch has been **rebased onto `70f811a`** (main had moved 3 commits ahead), so the gate result above is measured on the merged tree, not a stale base.

### Audit round

A blind adversarial audit returned **no 🔴** and a verdict of safe-to-merge. It independently reproduced the 12/12 no-orphan result and the byte math. Five findings are now fixed in `e8af4b0`, three of which were real defects in my own change:

1. The `health` row claimed `extension_stale` is on `health` **ONLY**. Wrong — `server.py:2883` (`_whoami`) and `:2948` (health) both call `annotate_staleness`, and this file's own line 10 already said so. My staleness pass verified the field *name* existed; it could not test whether the *claim about* the field was true. That distinction is the gap.
2. The focus/workspace correctness fix had landed in the **body only**. `SKILL.md` mandates reading `reference/spa-wake.md` before `activate`, and that file still said "restore focus afterward" — so following the body's own routing delivered the pre-fix instruction. `reference/x-fallback.md` had the same defect on a path that focuses a window by construction. Both now name both axes. This is the "fix every copy" rule: a prune multiplies the copies of every claim, and I had grepped both files, seen them, and read them as proof the command *survived* rather than as two stale copies.
3. That same fix had **narrowed** the trigger from "if anything moved their focus" to `activate` only, leaving the X-fallback path covered by neither the body nor a correct local rule — net coverage went down. Trigger restored.

Plus two nits: the auto-`wake` note was rescoped to the agent (it is agent-tool-only, `browser_tool_impl.mjs:644`), and "in a worktree-isolated agent" was restored to the `eval` guard.

Final size **11,916 B, headroom 372** against a required floor of 250 (constants owned by `scripts/browser-bridge/tests/test_skill_size.py`). Browser-bridge suite: **654 passed, 0 failed**. A delta re-audit of the fix round is running.

PR #567, opened 2026-08-19
skills: 47 sidecar routes pointed somewhere the reader never is — and the guidance is what put them there

ZacxDev:
devrc has the defect talos-infra closed earlier today (188 refs, gate 0 rule 1c), and its own dead-path gate (#553) could not see it: **#553 checks whether a file EXISTS; this is a SPELLING defect on files that all exist.**

47 sidecar routes across 13 skills — 12 of them in `prune-skill` itself, the skill that teaches the convention.

## Why the spelling differs from talos

This is the part worth reading, because porting the talos answer verbatim would have made 47 refs consistently wrong in a *new* way.

A talos skill is read **from the repo**, so repo-root-relative resolves. A devrc skill is read from `~/.claude/skills/<name>/` by an agent whose cwd is some unrelated project — so both the bare form *and* the repo-relative form resolve against that cwd and are not found.

| spelling | reader opens it | `skill-audit.py` sees it | gate #553 checks it |
|---|---|---|---|
| bare `reference/x.md` | no | yes | **no** ← the defect |
| repo-rel `claude/skills/…` | no | no | yes |
| deployed `~/.claude/skills/…` | **yes** | no → **yes, here** | yes |

Only the deployed path opens. All **29 distinct deployed claims verified to resolve at the real consumer path**, not merely in the source tree.

## Order was load-bearing — converting first would have been a regression

`~/.claude/…` expands to an absolute path under the HOME deploy root, not under the source `skill_dir`, so `Path.relative_to(skill_dir)` raised and `skill-audit.py` returned `None`. `prune-skill` — three sidecars, eleven routing lines — scored as *"no skill routes to a reference/ sidecar yet"*.

Fixing the corpus without fixing the resolver would have traded one blindness for another. Caught only by re-running the auditor **after** converting, instead of assuming the conversion was neutral.

## The cause was guidance, not drift

Both sources are fixed here, or this regenerates:

- `prune-skill`'s own routing-forms table **sanctioned** the bare form for devrc skills ("BOTH forms: `reference/<topic>.md` …").
- `skill-audit.py` printed **"Usually FINE"** for exactly this shape.

## Verification — by mutation, not by a green run

| mutation | result |
|---|---|
| remove rule 1c | both new doc-path cases fail, each for its own reason |
| revert the resolver fix | 2 of 4 new skill-audit cases fail |

The other 2 **survive**, and are labelled INVARIANT GUARDS in-file having been *checked* rather than assumed: orphan detection does not depend on ref resolution, and a resolver that recognises nothing trivially claims nothing.

Same-commit suite comparison (both worktrees at `70f811a`):

```
BRANCH: 26 failed, 5824 passed
BASE  : 26 failed, 5815 passed
introduced by this branch: (none)   fixed by this branch: (none)
```

Identical failure sets; **+9 passing = exactly the 9 tests added**. (An earlier comparison looked clean but was invalid — the two runs sat at three different commits, so a 203-test gap was upstream growth, not evidence about this change.)

## Scope note

The bare **directory** form `reference/` stays legal — it names the *convention*, and this corpus is full of "demote it into a `reference/` sidecar", including the sidecar that teaches it. Flagging it reds the guidance itself. It tripped the gate on the first run in both repos.

## Side effects

- `prune-skill` hit its byte ceiling as a result (13,160 B vs a 12,864 B enforced budget). Its gate forbids raising `MAX_BYTES`, so §4 detail and a 544 B always-loaded block were **demoted** into the sidecars that own them → 12,720 B, 336 B headroom. The enumerated survival check flagged two claims as lost; reading showed both present under different wording — the known false-loss mode of a token check.
- Corrects a claim propagated from talos: the case study in `verification-protocol.md` cited the image-cacher `le` warning as `"30.0"` vs `"30"`. Re-measured — there is **no `le=30`** on either image-cacher histogram (`0.0 5.0 10.0 25.0 50.0 …`; nearest anywhere is `le=300.0`, on a different metric). The rule was right, the example was fabricated. Correcting it in talos had left this copy stale — the "a prune multiplies the copies of every claim" lesson landing on the doc that records that lesson.

🔴 **Needs a `home-manager switch` after merge** — devrc skill bodies are `/nix/store` copies, so a green merge alone changes nothing for a running agent. Verify by CONTENT at the deployed path.


ZacxDev:
## Audit round applied — plus two corrections to the PR body above

Blind adversarial audit run before merge. **No deploy-blockers.** Five should-fix findings, four of them mine; all fixed in `e90d5a1`. Two of the findings are about *this description*, so correcting them here rather than silently editing the body — the text above was wrong when it was read.

### ❌ Correction 1 — "a 12,864 B enforced budget"

That number is arithmetically `MAX_BYTES − MIN_HEADROOM_BYTES` = `13,056 − 192`, but the phrasing collides with reasoning that `test_prune_skill_size.py` explicitly records as **retracted**:

> The comment here previously read "12,864 B measured + 192 B headroom"; the file measured 12,834 at the time, so the arithmetic was describing a size the file never had.

Stated correctly: **ceiling `MAX_BYTES = 13,056`, headroom floor `MIN_HEADROOM_BYTES = 192`.** After this round `prune-skill/SKILL.md` is **12,861 B with 195 B of headroom** (it moved, because restoring the §4 carve-out cost 87 B and §4 was tightened rather than the ceiling raised).

### ❌ Correction 2 — "47 across 13 skills" is the `claude/` population, not the devrc-skill population

The gate's corpus is `CORPUS_DIRS = ("claude", "CLAUDE.md")`. **`scripts/browser-bridge/SKILL.md` is a deployed devrc skill core carrying 23 more bare `reference/*.md` routes**, and it sits outside that corpus — so the new rule structurally cannot reach it.

It is worse there than in the 47: `~/.claude/skills/browser/` holds only `SKILL.md` + the `browser` CLI, so the `reference/` subtree **is not deployed at all**. Those 23 routes are broken twice over. Not fixed here (different population, different deployment shape) — **filed as a follow-up**, and called out because "47" reads as complete and is not.

### Findings fixed in `e90d5a1`

| # | finding | why it mattered |
|---|---|---|
| 1 | **The negative control for rule 1c's central design choice was vacuous** — the mutant survived a fully green suite | `PROBE_DOC` is itself inside `reference/`, so the doc-relative mutant resolved to `reference/reference/…` and was dead anyway. **35 of the 47 real refs lived at `SKILL.md` depth**, where the false-clean *is* reachable. The in-file comment asserted the opposite and was simply false. |
| 2 | **The resolver disambiguation was a substring test, not a segment test** | `mailbox` and `vetr-mailbox` are both real skills here; a legitimate cross-reference was claimed as `mailbox`'s own sidecar, then reported as a phantom missing topic. Widening the resolver made this worse, and my test used `OTHER` — which shares no suffix and proved nothing. |
| 3 | **§4 had lost the `mkOutOfStoreSymlink` carve-out** | The flat new rule is wrong for `browser`/`dl-router`, and the gate certifies the resulting path as CLEAN — so guidance is the only possible guard. This introduced, into the skill that teaches §5, exactly the defect §5 names. |
| 4 | Six mangled apostrophes (`.s` for `'s`), one in shipped prose | Cause: talos's validator embeds awk in a single-quoted shell string where an apostrophe in a *comment* breaks the script. I formed the habit there and carried it across a repo boundary into Python and Markdown, where no such constraint exists. |
| 5 | `bar/SKILL.md` path duplicated back-to-back; `test_prune_skill_size.py` docstring stated stale figures while declaring itself their single source | — |

### Full mutation battery re-run — a fix round resets the gate

Not just the mutants these fixes targeted. Baseline green (138), then:

```
M1 remove rule 1c                     -> KILLED (3 failed)
M2 revert resolver fall-through       -> KILLED (2 failed)
M3 rule 1c -> doc-relative resolution -> KILLED (1 failed)   [SURVIVED before this round]
M4 marker  -> substring find()        -> KILLED (1 failed)   [the collision above]
```

M2 first reported **"MUTATION DID NOT APPLY"** because fix #2 moved its anchor — the battery asserts application rather than scoring a no-op as SURVIVED, which is the trap that makes an unmutated baseline read as a surviving mutant. Run under `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged between mutants, in a `cp -a` copy whose `.git` **file** was removed first (a copied worktree otherwise shares the original's git dir).

The auditor also verified the `le=30` claim **live** against dp-prod Prometheus rather than taking it on trust, with a positive control on the same matcher.

ZacxDev:
## Audit complete — 5 rounds, stopping here

Final tip `f908718`. Round 5 returned **no deploy-blockers**, verified rounds 1–4's fixes by independent *execution* rather than by reading the diff, and its author recommended against a sixth round. Its one substantive finding is closed in the final commit.

### What each round found

| round | findings | character |
|---|---|---|
| 1 | 5 (4 mine) | **behavioural** — a vacuous control, a resolver collision |
| 2 | 6 | doc accuracy + a real §4 regression the round-1 fixes introduced |
| 3 | 1 class | stale figures, 3rd occurrence → **gated** instead of fixed a 4th time |
| 4 | 4 | all about the new gate — incl. a cross-check that never consulted the tool it named |
| 5 | 1 | my own `UNGATED` contract, broken by the commit that stated it |

Every round after the first found something the *previous round's fixes* introduced or left. That is the pattern this loop exists to catch, and it is why the stop condition is a clean round rather than the author's say-so.

### The two findings I'd want a reviewer to see

**Round 1 — the negative control for this PR's central design choice was vacuous.** Rule 1c judges *spelling* and deliberately does not resolve against the citing doc's directory. Mutating it *into* a doc-relative resolution left the full suite green, because `PROBE_DOC` sits inside `reference/` so the token resolved to `reference/reference/…` and was dead anyway. **35 of the 47 real refs lived at `SKILL.md` depth** — precisely where the false-clean is reachable. The in-file comment asserted the opposite and was false.

**Round 4 — the check labelled "skill-audit cross-check" never consulted skill-audit.** It hard-coded `12_288` rather than importing it, so moving the tool's `TARGET` to 12,000 made the tool print 859 while the prose still claimed 571 — and the gate passed. Round 3's finding was *literally* "a stated cross-check the tool refutes in one command", so the fix for it reproduced it inside itself.

### The argument for gating over re-reading

When the figure gate was first extended, **it found a stale number on its first run** — `browser-bridge` stated 11,821 B, measured 11,942 B. Three audit rounds and four hand-passes over that same docstring had walked past it.

### Final verification state

```
suites   : doc_path_rot 76 · skill_audit 62 · prune_skill_size 32  = 170 passed
battery  : M1 remove rule 1c            -> KILLED (3 failed)
           M2 revert resolver fall-through -> KILLED (2 failed)
           M3 rule 1c -> doc-relative   -> KILLED (1 failed)
           M4 segment boundary -> if True -> KILLED (1 failed)
           each dying to its own named test; re-run after EVERY fix round
figure gate: 18 derived checks + a declared UNGATED set (1 entry, with its reason)
corpus   : 0 bare refs; 29 distinct deployed claims resolve at ~/.claude/skills/
baseline : 25-26 pre-existing failures, same-commit attributed, none introduced
```

The mutation battery asserts each mutation actually *applied* (a no-op mutator leaves a clean green that reads exactly like "mutant survived"), runs under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` purged between mutants, in a `cp -a` copy whose `.git` **file** was removed first.

### Still open, deliberately not in this PR

**`scripts/browser-bridge/SKILL.md` carries 23 more bare refs** and sits outside the gate's corpus (`CORPUS_DIRS = ("claude", "CLAUDE.md")`). Worse there than in the 47: `~/.claude/skills/browser/` holds only `SKILL.md` + the CLI, so its `reference/` subtree is **not deployed at all** and those routes are broken twice over. Different population, different deployment shape — filed as follow-up rather than folded in.

🔴 **Needs a `home-manager switch` after merge.** devrc skill bodies are `/nix/store` copies, so a green merge alone reaches no running agent. Verify by CONTENT at the deployed path, not by the ship script's own line.

ZacxDev:
## 🔴 Retraction — the `scripts/browser-bridge/SKILL.md` follow-up I filed twice above is WRONG

I claimed, in [this comment](https://github.com/innovation-upstream/devrc/pull/567#issuecomment-5349127197) and again in the [audit summary](https://github.com/innovation-upstream/devrc/pull/567#issuecomment-5349800103), that `scripts/browser-bridge/SKILL.md` carries **23 bare refs that are "broken twice over"** because its `reference/` subtree is not deployed. Correcting it publicly rather than editing it out, because it is a false claim about a file that is fine, and someone may have read it.

**Measured just now — the file is correct:**

```
distinct routes                     : 11
resolve at the STATED base          : 11/11
absolute-base statement             : line 21
first route                         : line 46
routes appearing BEFORE that base   : 0
skill-audit verdict                 : 11 referenced reference file(s), all resolve; no orphans ✓
```

Lines 19–21 state the base emphatically, ahead of every route — and already document the exact non-deployment fact I cited as damning:

> 🔴 **Every `reference/<file>.md` named below lives at `~/workspace/devrc/scripts/browser-bridge/reference/`** — that exact path; only `SKILL.md` + the CLI are symlinked into `~/.claude/skills/browser/`.

That is the **sanctioned** pattern — prune-skill §4's "state the expansion once above a table of short entries", and the reason `skill-audit.py` has an `ABS_REF_BASE` check at all. The file uses it correctly and discloses the caveat.

### Where the wrong claim came from

A round-4 audit reported the base statement "sits at line 122, *after* 10 of the routes". I repeated it without measuring. It is off by 100 lines and inverted — **0** routes precede the base. My own rule for this, from earlier in this same PR: *re-verify a subagent's finding before building on its stated reason.* I applied that to four findings and then skipped it on the one I was merely repeating rather than acting on. Repeating a claim is acting on it.

### It also would not have been fixable the way I proposed

Converting 27 route mentions to absolute paths costs ~1,161 B against **96 B** of headroom under that skill's own ceiling (11,942 B against an enforced 12,038). The stated-base pattern is not a workaround there; it is the only thing that fits, which is presumably why it was chosen.

### One thing this does firm up

Not extending gate rule 1c's corpus to `scripts/browser-bridge/` was the right call, for a better reason than I gave. Rule 1c is a pure **spelling** check with no notion of a stated base — correct for `claude/skills/`, where no core states one, and it would **false-block** a file that legitimately declares its base up front. The corpus boundary is a real design line, not an oversight.

**Net: no follow-up work exists here. Withdrawn.**

PR #580, opened 2026-08-20
docs(opencode): the AGENTS.md "7.7k uncached tokens per request" is HALF true — the tokens are real, the "uncached" is not

ZacxDev:
## The measurement, with its control

Task was: confirm or refute that the generated `~/.config/opencode/AGENTS.md` costs ~7.7k **uncached** input tokens on every opencode request, on the **current** engine — and only then cut. Measured first. **The token count is confirmed; the "uncached" half is refuted.** No cut is warranted, so this PR changes only the claims.

Engine **1.18.18**, model **`openrouter/xiaomi/mimo-v2.5`**. Instrument: opencode's own store `~/.local/share/opencode/opencode-stable.db`, read read-only.

### A/B pair — same prompt, same model, one variable

One trivial prompt ("reply PONG, use no tools") in an empty scratch project, under two config dirs identical in every byte **except this file** — B was a `cp -a` of the live `~/.config/opencode` with only `AGENTS.md` deleted, so skills, agents, plugins and `opencode.jsonc` were held constant.

| run | `tokens_input` | `tokens_cache_read` | cost |
|---|---|---|---|
| **A** — with `AGENTS.md` | 22,019 | 1,024 | $0.0030964 |
| **B** — without | 13,690 | 1,024 | $0.0019276 |

→ the file is **8,329 input tokens** (43,676 B ÷ 8,329 = 5.24 B/token). Slightly **larger** than the 7.7k recorded on 1.18.4 — the file grew 38,363 → 43,676 B. On a **cold** request it really is paid in full.

### …but the steady state is cached, and that is what changes the verdict

`AGENTS.md` sits at the **head** of the prompt prefix, which is exactly what a provider prefix-cache retains. Two controls:

- **Third run** — fresh session, same prompt, ~2 min after A: `tokens_input` **3**, `tokens_cache_read` **23,040**, cost **$0.000074**. A 42× reduction, and *cross-session*, so the cache is not per-session.
- **Fleet-wide** on this box over 1.18.16+1.18.18: **2,218 billed requests, of which only 60 (2.7%) were cold.** By version the cached fraction is **92.0% on 1.18.4**, 93.0% on 1.18.9, 93.1% on 1.18.16, 97.2% on 1.18.18 — so caching was **already working on 1.18.4**. It was never engine-version-dependent; the zero in `browser-agent`'s note is a fact about *that harness's* single-shot shape, not about the engine.

Solving the three probe rows for prices (three equations, three unknowns): input **≈$0.14/Mtok**, output **≈$0.28/Mtok**, cache read **≈$0.0028/Mtok** — a cached token bills **~50× under** an input token. Attributing the file across that 2,218-request window: 60 cold × 8,329 × $0.14/M + 2,158 warm × 8,329 × $0.0028/M ≈ **$0.12 of $1.77 total spend, ~7%**.

**Verdict: not a token-cost lever.** Per the brief — "if the overhead is now small or cached, say so and stop" — Phase 2 was not entered. Nothing is cut, no ceiling is moved, `claude/RULES.md` is untouched.

## What this PR does change

Every hunk is a stale or over-general **claim**. The diff on `nix/home.nix` and `scripts/browser-bridge/browser-agent` contains **zero non-comment lines**, so the generated file's `.text` expression and browser-agent's behaviour are byte-identical.

- **`nix/home.nix`** — the generator comment recorded "38,363 B / 37.5 KB ≈ 8.9k tokens (at 4.31 B/token)" for a file that is now 43,676 B, and implied a per-request tax. Replaced with the current pair + the cache finding, pointing at the README for the working.
- **`scripts/opencode/README.md`** — same stale paragraph; now carries the measurement, its control, the price solve and the conclusion under **"Size, and what it actually costs"**.
- **`scripts/browser-bridge/browser-agent`** — "`cache.read`/`cache.write` were 0 in every run, so it is not amortised" reads as a claim about the **engine**. Scoped it to this harness, and stated plainly that the isolated-config-dir trick is right *here* and **wrong as a general fix** — it also drops `plugin/guard.js`, the enforcement layer between an agent and an irreversible action. Its `~37 KB` → `~43 KB`.
- **`scripts/tests/test_opencode_config.py`** — one assertion message said `~37 KB`.

## Verification

- `nix-instantiate --parse nix/home.nix` → OK.
- `test_opencode_config.py` + `test_browser_agent.py` → **688 passed** in the `pytests` devShell (`PYTHONDONTWRITEBYTECODE=1`, no `-x`).
- No new test: this is comment-only, and a byte-pin on the generated size would trip on every `RULES.md` edit — a permanently-red gate, which RULES.md forbids.

## What I could not verify

- The **price solve** ($0.14 / $0.28 / $0.0028 per Mtok) is derived from three probe rows, not from a published OpenRouter price sheet. It reproduces all three rows to within ~1%, but it is a fit, not a quoted tariff.
- **Cache TTL is unmeasured.** I measured warm at ~2 minutes and cold at first contact; I did not find the boundary between them. The 2.7%-cold figure is the empirical answer for *Zach's* usage pattern and would move for a workload of widely-spaced one-shot runs.
- The A/B delta is **one** cold pair. A second cold pair is not cheaply obtainable — a repeat run hits the very cache being measured. It is cross-checked against the independent historical figure: 7,700 × (43,676/38,363) = 8,766 vs 8,329 measured, agreeing within 5% across two different tokenizers.

## Notes for siblings

Touches nothing in `scripts/collector/opencode/activity-plugin.js`. The probe runs wrote 4 real rows to the shared opencode SQLite store and emitted `source=opencode` telemetry — expected, flagged for the session-attribution agent: titles `ZPROBE-A1-withagents`, `ZPROBE-B1-noagents`, `ZPROBE-A2-withagents-repeat`, total spend $0.0067.

### DIFF
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -1162,9 +1162,22 @@ in
   # wins), so this is the file opencode really reads.
   #
   # Concatenating at switch time means it can never drift from the sources
-  # Claude Code reads. Measured result: 38,363 B / 37.5 KB ≈ 8.9k tokens (at the
-  # 4.31 B/token measured on this exact content) — safe. (For scale: a 331 KB
-  # AGENTS.md causes a permanent compaction loop.)
+  # Claude Code reads.
+  #
+  # COST — re-measured 2026-08-19 on engine 1.18.18 / openrouter/xiaomi/mimo-v2.5.
+  # The file is 43,676 B = 8,329 input tokens (A/B pair: one trivial prompt in an
+  # empty scratch project, under two config dirs identical in every byte except
+  # this file — 22,019 input tok with it, 13,690 without).
+  # 🔴 That is NOT a per-request tax, and the note that used to stand here
+  # implied it was. The file sits at the HEAD of the prompt prefix, which is
+  # exactly what a provider prefix-cache retains: of the 2,218 billed requests
+  # this box logged on 1.18.16+1.18.18, only 60 (2.7%) were cold, and a cached
+  # token bills ~50x under an input token. Attributed cost of this file over that
+  # window: ~$0.12 of $1.77 total spend (~7%). A COLD first request does pay the
+  # full 8,329 — that, not the steady state, is what browser-agent's isolated
+  # config dir sheds (every one of its runs is single-shot).
+  # Full working: scripts/opencode/README.md → "Size, and what it actually costs".
+  # (For scale: a 331 KB AGENTS.md causes a permanent compaction loop.)
   # scripts/tests/test_opencode_config.py pins the content and a 100 KB ceiling.
   home.file.".config/opencode/AGENTS.md".text =
     builtins.readFile ../claude/PRINCIPLES.md
--- a/scripts/browser-bridge/browser-agent
+++ b/scripts/browser-bridge/browser-agent
@@ -80,17 +80,28 @@ die() { echo "browser-agent: $*" >&2; exit 2; }
 command -v "$OPENCODE_BIN" >/dev/null 2>&1 || die "opencode not found on PATH ('$OPENCODE_BIN') — install it or set BROWSER_AGENT_OPENCODE"
 
 # --- isolated opencode CONFIG dir (token containment) ------------------------ #
-# devrc now deploys a ~37 KB global ~/.config/opencode/AGENTS.md (PRINCIPLES +
+# devrc now deploys a ~43 KB global ~/.config/opencode/AGENTS.md (PRINCIPLES +
 # RULES + addendum). MEASURED on 1.18.4: that file IS injected into a run whose
 # --dir is an unrelated scratch project — a tool-less probe in a scratch dir
 # quoted a passphrase planted at the END of the file, so it arrives in FULL.
 #
 # Cost here: 8,343 input tokens vs 631 with an isolated config dir = ~7.7k
-# EXTRA PER REQUEST. `cache.read`/`cache.write` were 0 in every run, so it is
-# not amortised — it is multiplied by the agent-loop step count ($STEPS,
-# default 12), i.e. up to ~90k tokens per browser-agent run. For an agent whose
-# every host tool is DENIED (no shell, no file access), 100% of that is rules it
-# can neither use nor act on.
+# EXTRA PER REQUEST (re-measured 2026-08-19 on 1.18.18: 8,329 — the file grew,
+# the number did not really move). It is multiplied by the agent-loop step count
+# ($STEPS, default 12), i.e. up to ~100k tokens per browser-agent run. For an
+# agent whose every host tool is DENIED (no shell, no file access), 100% of that
+# is rules it can neither use nor act on.
+#
+# 🔴 SCOPE OF THE "not amortised" CLAIM. `cache.read`/`cache.write` were 0 in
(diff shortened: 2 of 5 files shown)


ZacxDev:
## Addendum — `test_opencode_engine.py` ledger entries (`5af3459`)

The integration gate flagged 6 version claims from this PR. **Fixed by enumeration, not by editing the numbers.**

#570 sharpened `_VERSION_RE` today so it matches a version at the **end of a sentence** — which made six of this PR's prose lines visible to `test_no_file_in_the_pins_surface_still_claims_an_OLD_version` (5 in `scripts/opencode/README.md`, 1 in `nix/home.nix`). Newly *visible*, not newly *stale*.

Re-deriving them against `1.18.18` would have been the wrong fix: the finding **is** a comparison across engine versions. A cache hit rate measured at each engine in the series is exactly what refutes the theory that caching arrived recently. Re-keying those numbers to the pin deletes the control and leaves one data point that cannot support the conclusion — and one line, *"it was never engine-version-dependent"*, would become the **opposite** of what was measured. Six entries added, each with its reason. No number changed, no prose reworded.

**The guard also caught my own first draft** of the comment heading that block, which enumerated the version series inline — correctly, since this file is itself in `PIN_SURFACE`. I rewrote the comment to spell no version literal rather than add a ledger entry exempting the ledger's own commentary, which would be circular. The comment records that, so it doesn't get re-added.

### Verification

- **Red at `4824001`** → 1 failed / 24 passed, six stale claims named.
- **Green at HEAD** → 25 passed. Whole-file runs, `PYTHONDONTWRITEBYTECODE=1`, no `-x`.
- **Mutation control** (are the entries load-bearing and *reachable*, or green for another reason?): corrupting **one** snippet — the `nix/home.nix` entry — went red in **both** directions, each with its own guard's specific error. `test_no_file_in_the_pins_surface_still_claims_an_OLD_version` named exactly `nix/home.nix:1174` **and no other line** (the other five stayed exempt, so the entries are individually load-bearing), and `test_every_historical_version_claim_still_exists` reported the orphan (`matches NO old-version line`). 2 failed / 23 passed. Restored from a `cp -a` copy, not `git checkout --`, and re-verified green.
- Neighbours re-run together: `test_opencode_engine` + `test_opencode_config` + `test_browser_agent` = **713 passed**.

### Ledger mechanism — a gap, reported not fixed

The coordinator asked whether the enumeration distinguishes *historical by design* from *stale and forgotten*. **It does not.** The guarantee is **structural, not semantic**: it enforces one entry ↔ exactly one extant old-version line (orphan → red, >1 match → red, unenumerated line → red). The `reason` field is **never asserted on** — it is only interpolated into failure messages — so an entry added with a plausible-sounding but false reason goes green.

The **deployed-state** sub-case (described in the ledger's own comment below this block) is where that has already bitten: those entries were caught only because updating the *line* orphaned them. A deployed-state claim whose line never changes stays exempt **forever**. That is the same shape as the `_VERSION_RE` blind spot #570 fixed — a guard that cannot see a class reports a confident clean.

Left alone deliberately: *"is this reason true"* is not mechanically checkable, and a half-measure invented here would be scope creep on a docs PR. Flagging it as the coordinator asked. (One trivially closeable nit if anyone wants it: an entry with an **empty** reason string currently passes.)

PR #582, opened 2026-08-20
feat(session-write): the four tmux WRITE verbs behind ONE validated wrapper

ZacxDev:
> [!CAUTION]
> **CORRECTED 2026-08-20 — the first version of this PR shipped a bypass, and this body asserted the false premise behind it.**
> An adversarial audit found that `session-write type` **executed arbitrary commands in any shell pane**: the G8 row below used to read "`type` cannot execute because it never sends one", and that is measurably false — a Ctrl-O (`\x0f`, readline `operate-and-get-next`, a **bash default**) executes the pane's buffer with no Enter sent. `--text` was a denylist and accepted it.
> Fixed, plus five other audit findings. The full correction is in [this comment](https://github.com/innovation-upstream/devrc/pull/582#issuecomment-5351816196), which is where a reviewer who already read the wrong version should look. The sections below are updated.

PR 3 of the agent-attention-tooling line: the WRITE verbs, layered on the read-only `session-resolve` from PR 2.

```
session-write dismiss <selector> --i-am-at-the-keyboard [--client TTY]
session-write focus   <selector> --i-am-at-the-keyboard [--stay] [--client TTY]
session-write type    <selector> --text "..." [--append]
session-write send    <selector> [--text "..."] [--append] [--allow-shell-exec]

  --append is required whenever the pane holds UNSENT operator text —
  including for a bare `send`, which would otherwise submit a human's
  half-finished sentence unread.
```

`session-resolve` is **untouched**. It is still read-only by construction — `TMUX_READ_ONLY_SUBCOMMANDS` is exactly three read verbs, `_assert_read_only` is unweakened, and `test_session_write.py::test_session_resolve_is_still_read_only` pins both. The whole separation is decorative if that ever stops being true.

## Why a wrapper, not `Bash(tmux send-keys:*)`

That allowlist entry is a **total bypass** of the Bash permission classifier, not a loosening of it: the classifier reads the tmux invocation, approves it, and the payload then executes in a *different pane* having never been classified at all. One entry is intended — `Bash(scripts/session-write:*)` — and its entire value is that this file constrains the payload.

## 🔴 Five measured refutations of the brief

Each one changed the design. All measured on the workbench 2026-08-19.

**1. `runtime` does not separate a Claude pane from a shell pane.** Over all 92 resolved targets: `claude=51, opencode=5, **None=36**`. session-manager's own source says `runtime` comes from the ledger and is null "whenever no writer has recorded that window, which is the common case for a session that has not taken a turn yet" — so `None` means UNRECORDED, not shell, and it is 39% of targets. The shell gate is therefore an **allowlist** of the two measured agent runtimes; everything else, `None` included, falls to the guarded side. Failing toward "treat it as a shell" costs one explicit flag; failing the other way executes an agent-authored line in a live shell.

**2. There is not one base terminal — there were two.**

```
/dev/pts/0   alacritty 312x63  datapacket-talos    [attached, focused, UTF-8]
/dev/pts/69  alacritty 77x64   datapacket-talos-3  [attached, UTF-8]
```

and `compute_visibility` publishes `base_terminal = base_all[0]`, the **first in list order** — which happens to be the focused one today. A test written against that field would pass now and steer `switch-client` at the wrong terminal later. `focus` re-derives from the client list and disambiguates on the `focused` flag (unique *within* the base-terminal set; the popups carry it too). The fixture deliberately puts the unfocused one first.

**3. `switch-client` needs `-c`.** With two attached base terminals, a bare `switch-client -t <session>` acts on tmux's "current client", which this process has no claim to. Every client-directed verb names its client.

**4. `covered` is not the precondition for `dismiss`.** The brief's own target measured `visible=False, covered=False, attached_clients=[]` **with two popups attached** — `covered` is only true when the base terminal is showing the target. Gating dismiss on it would refuse in exactly the state the operator is in.

**5. The claude/shell discriminator is unreachable from a resolved target.** session-manager's `row_class` keys on `row.get("claude")`; `session-resolve` declares `claude` in `SM_PASSTHROUGH_FIELDS`, collects it into `sm_block`, and `build_targets` **never emits it** — `'claude' in target` is False for all 92. Found by the seam test, not by reading. Left alone here; the fix belongs in session-resolve.

*(Minor: the brief's popup pair was stale — scratch11 + scratch13, not scratch14. The count, which is what the design turns on, was right.)*

## 🔴 Three tmux behaviours that silently corrupt — or EXECUTE — a payload

Both measured on tmux 3.7b in a **private `-L` server**, never the operator's:

```
$ tmux send-keys -t %0 -l -- 'abc;'      # rc 0 — pane receives "abc"
$ tmux send-keys -t %0 -l -- $'X\nY'     # pane receives X, ENTER, Y
```

* A **trailing `;` is eaten** — tmux separates commands on a `;` welded to the end of an argv token. As the last element nothing dangerous runs, but the operator's semicolon is gone and the call reports success. Refused with the remedy (append a space) rather than mutated behind the caller's back.
* A **newline inside `-l` literal text is a real Enter — it submits.** `type` promises "literal, no Enter"; a `\n` breaks that with no error, and on a shell pane it *executes*. Both verbs refuse `\n`/`\r`; `send` presses Enter with its own separate argv so the submit is one auditable act.
* 🔴 **(c) ENTER IS NOT THE ONLY EXECUTION EVENT — a control character executes.** Found by the audit, after this PR was already reviewable:

```
$ tmux send-keys -t %0 -l -- $'touch <d>/EXECUTED\x0f'   # NO Enter sent
  -> EXECUTED was CREATED
$ the same payload without the \x0f, still no Enter      # negative control
  -> nothing ran
```

  `\x0f` is Ctrl-O, readline `operate-and-get-next` — **a bash default**. `\x1ba` is zsh emacs `accept-and-hold`. Which characters accept a line is a property of the *operator's* keymap, which this tool cannot read. Adding `\x0f` to the denylist was **rejected**: that is the same mistake one binding later. `--text` is now an **allowlist**.

## The guards

| | guard |
|---|---|
| G1 | Re-resolve at write time. argv carries a **selector**; a raw address is re-resolved, never passed through. |
| G2 | 🔴 **Local host only.** session-resolve resolves across both hosts via session-manager; tmux here is local. A laptop-resolved address written to local tmux does not fail — it hits a *different window that shares it*. 30 of 92 targets are remote, so this is reachable every run. An **unknown** local host is refused too: locality that cannot be proven is not locality. |
| G3 | Re-verify immediately before the write, and pick the client from the **fresh** read — "re-verify at the moment you ACT" covers the popup you are about to detach as much as the window you are about to write to. |
| G4 | Allowlist at the subprocess seam, segmented with session-resolve's own `_split_on_separators`. `kill-session`/`kill-pane`/`respawn-pane` cannot leave the file, and a welded `;` cannot smuggle a second command past. |
| G5 | `focus` **and `dismiss`** refuse without `--i-am-at-the-keyboard`, before resolution, and neither fires as a side effect of another verb. (`dismiss` was ungated in the first version — see the correction comment.) |
| G6 | `dismiss` never detaches a base terminal. `--client` searches **all** attached clients precisely so this guard is *reachable* — a chooser that only looked at popups would make it untestable. |
| G7 | Ambiguity refused, never guessed (two popups is the measured live state; two focused base terminals likewise). |
| G8 | The shell-execution gate, on `send` **only** — Enter runs a command on a shell and submits a prompt on an agent runtime. 🔴 `type` is ungated **because G13 makes an execution event unrepresentable in `--text`**, NOT because `send-keys -l` cannot execute — it can, see the corrected measurement (c) above. The two are one mechanism. |
| G13 | 🔴 **The `--text` ALLOWLIST**: printable characters plus tab, every other codepoint refused **by number**. This replaced a denylist of NUL/`\n`/`\r`/trailing-`;` that accepted `\x0f`, `\x1b`, `\x16` and every other control character. |
| G14 | 🔴 **The log path is confined to `~/.claude`.** `--log-path` and `$SESSION_WRITE_LOG` create nested directories and append caller-shaped JSON, both reachable from the one allowlisted Bash entry. Resolved with `realpath` + `commonpath` and refused before the selector is resolved. |
| G9 | The unsent-prompt gate. Its **content** never leaves the tool — this repo is public and that string is captured off the operator's screen. Presence and length only. |
| G10 | Every run logged as JSONL to `~/.claude/session-write.log`, refusals included. A dry run never logs as a write. |

**focus** records → raises **once** → restores **both** tmux dimensions (the client's session *and* the session's active window), including after a half-completed raise where `select-window` landed and `switch-client` errored. A failed restore gets its own exit code, because the screen is then somewhere the operator did not put it. 🔴 The **i3 workspace is explicitly out of scope** — this tool is tmux-only and every focus result says so in as many words.

Default is *restore*; `--stay` opts out. That is deliberately the opposite of what a human typing `focus` wants, and the tension is real — an operator asking to be shown a window wants `--stay`, an agent raising one transiently must put the screen back. The default serves the caller this permission entry exists for.

## Live verification

`--dry-run` against the real machine (writes nothing, exercises real resolution and real re-verify reads):

```
$ session-write focus scratch2:@81
REFUSED: focus changes your screen.
  re-run with --i-am-at-the-keyboard                                    rc=5

$ session-write type scratch2:@81 --text "smoke test payload" --dry-run
typed: 18 character(s) into pane %81 (scratch2:@81); NOT submitted
  tmux send-keys -t %81 -l -- smoke test payload                        rc=0

$ session-write type civitai-gpu-fleet:@0 --text "must not land" --dry-run
REFUSED: 'civitai-gpu-fleet:@0' resolved to host 'laptop', but this process
writes to LOCAL tmux only (local host: 'workbench'). ...                rc=4

$ session-write dismiss scratch2:@81 --dry-run
REFUSED: 2 popup clients are attached and none was named — refusing to guess
which one is covering the screen. Re-run with --client TTY:
/dev/pts/2 (scratch2, popup), /dev/pts/61 (scratch12, popup)            rc=9
```

The remote refusal is the one that matters: `@0` almost certainly exists locally too.

## Tests — 161, and they are INVARIANT GUARDS

🔴 `scripts/session-write` **did not exist at base**, so there is no "red at `origin/main`, green at HEAD" matrix and none is claimed.

🔴 **AND A GREEN RUN OF THEM IS STRUCTURALLY BLIND TO ONE WHOLE CLASS.** Every test in the file stubs the tmux seam, so none of them can disagree with a *premise* about what tmux and readline do with a payload — which is exactly where the audited bypass lived: **117 tests green while `type` executed arbitrary commands.** That is the lesson of this round, and it is why both harnesses below are now **committed to the repo**.

### `scripts/session-write-harness/real_pane_check.py`

Drives the **real CLI** against a **real bash pane** in a private `-L` tmux server (never the operator's) and reads the answer **off the filesystem** — a marker file — rather than off the tool's own report.

Its own negative control is a copy of session-write carrying the **pre-fix denylist** (`SESSION_WRITE_SRC=…`). Against that copy it reports:

```
FAIL  THE AUDITED BYPASS: same payload + Ctrl-O, must be REFUSED
        exit     : 0  (accepted)
        executed : True
        PROBLEM  : expected refuse=True, got False; 🔴 THE MARKER FILE WAS CREATED — IT EXECUTED
```

so its passes on the fixed tree are falsifiable rather than decorative.

### `scripts/session-write-harness/mutation_sweep.py`

**52 mutants, every one killed by its NAMED test** — never by "something went red". Classes exercised: `deletion=17, branch-inversion=13, operand-swap=8, renumber=6, comment-out=4, allowlist-widening=2, stale-rebind=2, off-by-one=1`. A deletions-only sweep is what missed the original seven survivors.

**Two controls, both required to behave:**

| | control | requirement |
|---|---|---|
| PC | a mutation known to be caught | must be **KILLED** |
| NC | a semantically **null** edit | must **SURVIVE** |

PC alone cannot tell a working harness from one that scores everything as killed — a pytest that fails to start would report a flawless sweep. NC is what closes that, and the previous round had no NC.

Also: the baseline must be green before anything is scored; `__pycache__` is purged before every mutant with `PYTHONDONTWRITEBYTECODE=1` on every spawned process; each patch site is asserted to occur **exactly once**; the original bytes are restored in a `finally` with SHA-256 identity asserted; the **full** file runs per mutant (no `-k`); and a mutant that reddens more than half the suite is reported **NOT ISOLATED** unless the table *declares* it wide, so a kill from a broken import cannot be banked.

🔴 **The numbering is derived from the table**, so it cannot develop the gap at 11 that the old hand-maintained `MUTATION_MATRIX` had while the prose claimed 36. That constant is gone; a copy of results is exactly what went stale.

Three mutants are the ones the audit's independent sweep found **surviving** the first version — `tty`/`session` swapped, the popup/base label inverted, and every client rendered as `none`. All three were invisible because the expected strings were built by calling `sw._client_summary(...)`, the function under test, so the assertion was `f(x) == f(x)` and true for any `f`. They are literals now.

The sweep also caught a defect in a test written **this** round: a codepoint assertion was satisfied by the refusal message's own static prose, which names `U+000F`. It now probes a codepoint the prose does not spell and asserts the whole normalised string.

## Permissions

`Bash(scripts/session-write:*)` was added to the **workbench's** `.claude/settings.local.json` (gitignored, so it does not ship with this PR). **The laptop needs the same entry** — `scripts/sync-claude-permissions.py`.

No `Bash(tmux …)` write-verb entries were added; that is the design being rejected. *(Observation, not touched here: the existing allowlist already carries `Bash(tmux display-popup:*)` and `Bash(tmux source-file:*)`, both of which are write-ish. Worth a separate look.)*

## Gate

`bash scripts/gate.sh --tier pytest --set hermetic` — see the comment below for the run and its status.

🔴 **The brief's stated pre-existing failure did not reproduce**: `scripts/tests/test_opencode_engine.py` is **25 passed** on this host, and `git diff --stat origin/main` shows this branch does not touch it.


ZacxDev:
## Gate: **PASS (exit=0)**

```
bash scripts/gate.sh --tier pytest --set hermetic
```
run on the **merged tree** (`origin/main` merged in at `5a84315`), against a **clean working tree** (`git status -s` empty before and after — the gate builds from the working tree, so a dirty one would certify something other than the commit), at `7e67339`:

```
GATE: RESULT=PASS exit=0
TOTAL collected=12902  passed=12901  skipped=1  failed=0
  (floor: 12053 = sum of 24 per-target floors)
  PASS  scripts/tests  (collected=6038 passed=6038 skipped=0 floor=5764)
  ... all 30 targets PASS; the 1 skip is the pinned repo-cos live-store drift check
```

`scripts/tests` went **5921 → 6038** collected: exactly +117, this PR's suite. Well inside the floor's drift tolerance (`max(60, 5764/4)` = 1441), so `TARGET_FLOORS` needs no re-pin and the gate did not ask for one.

### 🔴 The brief's stated pre-existing failure was wrong — here is the real one

The brief said `test_opencode_engine.py` fails on this host (opencode 1.18.18 vs a pinned 1.18.16). **It does not** — `25 passed`, and `git diff --stat origin/main` shows this branch does not touch it.

The **actual** pre-existing failure, measured by gating `origin/main` at `4740e40` *before* writing a line of code, was `scripts/tests/test_prune_skill_size.py`: **`main` itself was red**, its browser-bridge size prose saying 11,942 B against a file measuring 11,990 B. #581 fixed it and landed after this branch was cut. Merging `origin/main` is what makes this PR green, which is also why the gate was run on the merged tree rather than the branch tip.

### Two failures the gate caught that the new suite could not

Both were repo-wide **TEXT** scanners tripping on session-write's **prose**, and neither is visible from running `test_session_write.py` alone. Fixed in `7e67339`:

* **`test_no_real_launchers.py`** — `wmctrl: ['session-write']`. `launcher_scan.hazard_hits` regexes the file body, so *naming* a binary in order to promise you never call it is indistinguishable from calling it. The word appears once, in the sentence declaring the i3 workspace out of scope for `focus` — deleting it to get green would delete the guarantee. The table already carries three justified occurrences of this exact shape and a note saying to re-justify rather than reword, so this is a fourth `ACKNOWLEDGED_UNSTUBBED` entry, pinned two-way to `{session-write}`. (`i3-msg` and `xdotool` are in the same sentence and need no entry — both are in `HOST_LAUNCHERS`, therefore stubbed.)
* **`test_runtime_shebangs.py`** — the suite asserted session-write's interpreter line as a literal. Shape 1 of `shebang_scan.line_is_offender` is "a quote immediately followed by the two shebang characters", and it cannot tell an assertion *about* a shebang from a test *writing* one. Measured twice: the full path is flagged, and so is a bare quoted marker. Fixed the way `shebang_scan` fixes it for itself — assembled from character codes, that module's own documented rule. The assertion is now **stronger** than the literal it replaced: session-write's first line must equal session-resolve's, so the two halves cannot disagree about how they are launched.

### Mutation sweep re-run after those fixes

**36 mutants, 36 KILLED, 36 killed by the expected test**, positive control included, baseline 117 passed, `final sha matches original: True`.

### Scope of these numbers

The workbench was carrying **seven other concurrent agent gate runs** during this work (load average peaked at 67). That inflates wall time — `scripts/tests` took 387 s here vs 568 s on the contended pass — but does not affect pass/fail: no test failed, so there is no timing-dependent verdict to attribute. No sibling process was signalled at any point; the two runs I abandoned were stopped through their own task handles after resolving PIDs and confirming `cwd` was my private worktree exactly.

ZacxDev:
## 🔴 CORRECTION — this PR shipped a bypass, and the PR body above asserted the false premise behind it

An adversarial audit found a **deploy-blocking hole** in the version of this PR that was already reviewable. Posting it as a comment rather than only editing the body, because **a reviewer may already have read the wrong version** — the body claimed something that is not true, and a silent edit would leave that reader with the false belief and no signal.

### What was wrong

The body's guard table said, of G8:

> The shell-execution gate, on `send` **only** — Enter is the execution event, and `type` cannot execute because it never sends one.

**The second clause is false.** Measured on tmux 3.7b in a private `-L` server, twice independently (by the auditor, and again by me before acting on the report):

```
send-keys -t %0 -l -- "touch <dir>/EXECUTED\x0f"     # NO Enter sent
  -> EXECUTED was CREATED

the same payload without the \x0f, still no Enter    # negative control
  -> nothing ran
```

`\x0f` is Ctrl-O — readline `operate-and-get-next`, **a bash default**, not an exotic rebinding. `\x1ba` is zsh emacs `accept-and-hold`. The old `validate_text` was a **denylist** of exactly NUL, `\n`, `\r` and a trailing `;`, and it accepted `\x0f`, `\x1b`, `\x16` and every other control character.

So `session-write type <shell pane> --text "...\x0f"` **executed arbitrary commands** — which is exactly the Bash-permission-classifier bypass this wrapper exists to prevent, reached through the single `Bash(scripts/session-write:*)` entry that was granted on the premise that this file constrains the payload.

Worse: `test_type_is_NOT_shell_gated` **pinned the hole as intended behaviour**, with the false premise written into its docstring. A 117-test suite was green throughout.

### The fix is an allowlist, not one more denied character

Adding `\x0f` to `SUBMITTING_CHARS` was considered and **rejected** — that is the same mistake one codepoint later, and the next keybinding walks it. `--text` is now validated against an **allowlist**: printable characters plus tab, everything else refused **by codepoint**. NUL/`\n`/`\r` keep their specific messages, but they are refused *by the allowlist*; a test pins that deleting a message branch weakens the diagnosis and never the guard.

### Also fixed in this round (all from the same audit)

| | finding | fix |
|---|---|---|
| A | three test expectations were built by calling `sw._client_summary(...)` — the function under test computing its own expected value, so the assertion was `f(x) == f(x)`, true for any `f`. The auditor's independent sweep found **3 survivors** here | literal expected strings |
| B | `--log-path` / `$SESSION_WRITE_LOG` were an unconstrained "create nested directories anywhere and append caller-shaped JSON" primitive, reachable from the one allowlisted entry | confined to `~/.claude` via `realpath` + `commonpath`; refused before resolution; the refusal is still journalled, but never to the path it just refused |
| C | `dismiss` was a screen-changing action with no gate | behind `--i-am-at-the-keyboard`, like `focus` |
| D | bare `send` submitted the operator's unsent prompt with no gate; `test_bare_send_is_not_unsent_gated` pinned it | gated, with its own SUBMIT wording (nothing concatenates on that path) |
| E | `focus --client <popup tty>` was accepted — naming a client skipped the base-terminal assertion, so `switch-client -c` could steer an overlay | the mirror assertion now runs on the explicit path too |
| — | "every refusal has its own code" at `:320` was false (`EXIT_AMBIGUOUS_CLIENT` covers five) | claim corrected; the per-gate discriminator is the `outcome` token, now enumerated and pinned two-way against what the source emits |

### Where the audit was WRONG, measured

Finding C's compounding claim was that "a plain `tmux attach` from inside another tmux pane is classified as a popup", making `dismiss` able to detach a real session. **The ordinary case does not arise** — measured, tmux 3.7b:

```
same server, $TMUX set   -> REFUSED: "sessions should be nested with care, unset $TMUX to force"
a DIFFERENT tmux server  -> ATTACHES, TERM tmux-256color   (so: misclassified)
$TMUX unset, forced      -> attaches
```

So it takes a second server or a deliberate `unset TMUX`, not a plain attach. The residual is real but narrow, and it is **not zero**, so `dismiss` took the gate anyway and every dismiss result now states that "popup" was decided on the client's TERM and is not proof of a `display-popup` overlay. Both the measurement and its limit are recorded in G6.

### Verification

🔴 **A green unit suite is not evidence here** — every test in the file stubs the tmux seam, and the bug was in a *premise* about what tmux and readline do, which a stub cannot disagree with. Two harnesses are now **committed**, because the previous sweep was unreproducible:

* `scripts/session-write-harness/real_pane_check.py` — drives the real CLI against a real bash pane in a private `-L` server and reads the answer **off the filesystem**. Its own negative control is a copy of the pre-fix denylist, which makes it report `🔴 THE MARKER FILE WAS CREATED — IT EXECUTED`. It does, so its passes are falsifiable.
* `scripts/session-write-harness/mutation_sweep.py` — **52 mutants, all killed by their named test**, plus a positive control (must be killed) and a **negative control: a semantically null edit that must SURVIVE**. Without the second, a harness that scored everything as killed would report a flawless sweep. Numbering is derived from the table, so it cannot develop the gap the old matrix had at 11.

The old in-file `MUTATION_MATRIX` is gone; it restated results that went stale and disagreed with its own count.

PR #583, opened 2026-08-20
feat(opencode): a dispatch gate for the failure that exits 0 having done nothing

ZacxDev:
Stage 1 of an opencode dispatch skill: **`preflight`**, **`run`**, and the SKILL body. Designed from 36 real dispatches over 17 days (5,226 transcript files / 294,686 Bash calls), not from the CLI's help text.

---

## The literal `description`, clause by clause

```
Dispatch a task to opencode — the separate headless CLI agent — instead of burning
this session's context on it. Preflights the brief first (a brief naming paths
outside --dir makes `opencode run` auto-reject and exit 0 having done nothing),
then launches detached. Use for "dispatch opencode", "call opencode", "run this
with opencode", "hand it to flash/mimo/pro", opencode token efficiency. NOT the
Agent tool — "dispatch a subagent" means that, not this.
```

**452 characters**, against a per-entry cap of 1,536 and a listing budget of 1% of the window. On overflow Claude Code silently drops descriptions starting with the least-invoked skills, so every clause has to earn its bytes:

| clause | measured justification |
|---|---|
| "Dispatch a task to **opencode**" | `dispatch` ×18 and `call opencode` ×8 in n=25 real utterances. The literal token `opencode` is the *only* disambiguator, so it leads. |
| "the separate headless CLI agent" | Tells the router what it is without a second lookup, and is the phrase that makes "headless"/"CLI agent" phrasings land. |
| "instead of burning this session's context on it" | **"token efficiency" ×3.** The stated motive is his own context budget, not money — seven weeks of opencode cost **$11.42** total, median dispatch **$0.014**. Framing it as cost would route on the wrong intent. |
| "Preflights the brief first (… auto-reject and exit 0 having done nothing)" | This is the *reason to use the skill instead of typing `opencode run`*. Without it the skill is a synonym for the raw command and will not be reached for. |
| "then launches detached" | 22 of 31 measurable dispatches already ran in Claude's background; the 9 foreground ones all hand-set long timeouts. States the behaviour so nobody re-derives the backgrounding. |
| `"dispatch opencode"`, `"call opencode"`, `"run this with opencode"` | Verbatim utterances. |
| "hand it to **flash/mimo/pro**" | *names a model* ×10. The alias words are the trigger tokens. |
| "**opencode** token efficiency" | The one phrasing where the motive is said out loud; kept adjacent to `opencode` so it cannot route a bare "token efficiency". |
| "NOT the Agent tool — 'dispatch a subagent' means that, not this." | 🔴 **The required disambiguation.** Measured: *"dispatch a subagent to implement X"* means the **Agent tool**; *"dispatch opencode …"* means this. Bare "dispatch" must not route here. Pinned by `test_the_skill_description_disambiguates_from_a_claude_subagent`. |

Deliberately **absent**: any mention of `browser`. *mentions `browser` ×14* is the sibling `browser agent` dispatch — a different tool — and putting the word here would steal 14 utterances from the skill that owns them.

Narrative (the four failure modes, brief-writing guidance, what to do when preflight refuses) is in the **body**, which costs zero until invoked.

---

### Built

- **`opencode-dispatch preflight --dir D [--brief F|-] [--json]`** — checks a brief, never dispatches.
- **`opencode-dispatch run --dir D [--brief F|-] [--title T] [-m flash|mimo|pro|<id>] [--auto] [-c] [--file F]…`**
- **`scripts/opencode/SKILL.md`** — the body.
- **`scripts/opencode/lib/{oc_permissions,brief_scan}.py`** — the shared resolver and the two scanners.

**Canonical argv order, pinned by index**: `opencode run <MESSAGE> --dir D [--title] [-m] [--auto] [-c] [--file …]`. `MESSAGE_ARGV_INDEX = 2` is a named constant and the test asserts `argv.index(message) == 2` and `index(message) < index("--file")`.

**Failure 1 closed structurally, not merely checked.** The brief is written **into `--dir`** at `.opencode-dispatch/<ts>-<slug>.md`; the tool also drops a self-ignoring `.gitignore` (`*`) into that subdir so it is unstageable in *any* target repo, and `.gitignore` here covers devrc itself. Nothing is ever written to Claude's scratchpad. On top of that, preflight scans the brief for absolute paths and **refuses with rc 3**, naming each offender and its resolved form.

**Path containment is exact, with an explicit enumeration rather than a heuristic.** `SYSTEM_ALLOW` lists ten read-only system prefixes (`/nix/store`, `/usr`, `/dev`, `/proc`, …), each with its reason in the source, not env-overridable, unknown-is-an-offender-by-default — the same shape as drift-check's `settings.json` key allowlist. 🔴 **`/tmp` is deliberately absent**: Claude's scratchpad lives there and "the brief lived in the scratchpad" is failure 1's exact vector. `/etc` and `/var` are absent too, since `/etc/nixos` is a real edit target here.

**Failure 3 warns and never blocks**, per the brief. It resolves each command node against `permission.bash` and prints the glob (`[ask] (rule '*kubectl*exec*')  kubectl -n data exec …`). It also surfaces anything `guard_core`'s `opencode` policy would hard-deny.

**Never in the foreground, no escape hatch.** `subprocess.Popen(..., start_new_session=True, stdin=DEVNULL)`, stdout+stderr to a log inside `--dir`, returns immediately. `test_the_parser_offers_no_foreground_escape_hatch` asserts the parser has no `--foreground`/`--fg`/`--wait`/`--sync`/`--blocking`.

**Model aliases**: `flash`/`mimo`/`pro` only. No alias → the `-m` flag is **omitted entirely** so the config default applies (18 of 24 measured dispatches). A full `provider/vendor/model` id passes through; an unknown bare word is a usage error, never a silent pass-through.

### Deferred, with the reason

- **`status` / `watch`** — they depend on opencode's internal SQLite schema, which carries no stability guarantee. Worth taking only once this skill is proven to get used. The dispatch prints a log path instead.
- **Result capture, auto-retry, a model router** — each considered and rejected on measured grounds. Retries in particular are *corrective, not transient*: re-running a known-bad brief is not a fix.

---

## 🔴 One rule, one place: the resolver moved

`scripts/tests/test_opencode_config.py` already carried a faithful, engine-validated port of opencode 1.18.18's `Wildcard.match` + `findLast` resolver. Preflight needs exactly that predicate. Rather than copy it, the port moved to **`scripts/opencode/lib/oc_permissions.py`** and both callers import it. Command *splitting* likewise reuses **`guard_core.split_commands`** — no second splitter.

The seam is pinned by **object identity**, not by an agreement matrix:

```python
assert cfg.wildcard_match is oc_permissions.wildcard_match
assert cfg._resolve_over is oc_permissions.resolve
```

`test_opencode_config.py`'s **628 assertions are unchanged and green** after the move.

---

## Red-at-base matrix

There is no pre-change code for a brand-new tool, so the honest matrix has two halves.

**Half 1 — the suite cannot be green without the implementation.** The new test file was run against the base tree (`210de3d`, produced with `git archive` into a plain directory — no `.git`, nothing that can touch a ref):

```
CLI present at base?  : NO
lib present at base?  : NO
registry row at base? : 0
→ ModuleNotFoundError: No module named 'brief_scan'
   1 error in 0.12s   (collection error — 0 tests could pass)
```

**Half 2 — per-test red, via mutation.** A collection error is a coarse red; the fine-grained evidence is the mutation matrix below, where every guard was watched to fail against a specific defect with a named killer.

**Plus one genuine reproduction of an original symptom.** Failure 2 was reproduced verbatim on the live `opencode 1.18.18` binary:

```
$ opencode run --dir <d> --file <brief> "Execute the task described in the attached brief."
Error: File not found: Execute the task described in the attached brief.
rc=1
```

…and the canonical order was verified end to end: `opencode-dispatch run --dir <d> --title "dispatch smoke" -m flash` dispatched, `deepseek/deepseek-v4-flash` ran, `hello.txt` was created with the expected contents, the process detached and the log captured the transcript.

---

## Mutation results

**18 mutants, 18 killed. `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged between mutants, no `-x`,** each anchor asserted to occur exactly once before being applied, and all three source files `cmp`-verified byte-identical to their backups at the end.

| mutant (narrowest expression) | killers |
|---|---|
| `MPC` positive control: `run` verb → `RUNX` | **1** |
| `M1` `is_under` loses the `os.sep` boundary | **1** `test_is_under_does_not_treat_a_sibling_prefix_as_a_child` |
| `M2` `is_under` → `True` | 6 |
| `M3` `is_under` → `False` | 4 |
| `M4` `scan_paths` drops the containment skip | 3 |
| `M5` `scan_paths` drops the allowlist skip | **1** |
| `M6` `extract_paths` drops the URL strip | 3 |
| `M7` path regex loses the prose lookbehind | 4 |
| `M8` `/tmp` quietly added to `SYSTEM_ALLOW` | 3 |
| `M9` resolver → first-match-wins | 2 |
| `M10` message moved after `--dir` | 2 |
| `M11` message appended last | 3 |
| `M12` `start_new_session=False` | **1** |
| `M13` brief written outside `--dir` | 2 |
| `M14` subdir `.gitignore` not written | **1** |
| `M15` unknown alias passes through | **1** |
| `M16` `run` dispatches despite a blocked preflight | **1** |
| `M17` `-m` flag suppressed | 2 |

**`-k` filter validated**: an alternation of all 18 distinct killer names selects exactly **21 items** (18 names, one of which is parametrised ×3) — so no killer is unreachable by the filter.

### 🔴 The battery found two real defects, both fixed in this PR

1. **`M6` SURVIVED on the first run.** The URL strip in `extract_paths` looked redundant — the regex lookbehind already rejects `https://host/a/b`, since every `/` there is preceded by an excluded character. Probing found the case where it *is* load-bearing: a path in a **query or fragment** is preceded by `=` or `#`, which are not in the excluded class, so `https://example.invalid/a?f=/etc/passwd` yields `/etc/passwd` without the strip. A brief that merely *links to documentation* would have been refused. Now covered by `test_the_url_strip_is_load_bearing_for_query_and_fragment_paths`, parametrised ×3, each carrying its own positive control (the raw regex match it *can* see), which is what kills M6.

2. **The outcome ledger was blind to a computed argument.** `emit("preflight-blocked" if blocked else "preflight-ok", …)` scored as one outcome; the second was never seen by the grep. The call sites are now two literals, `OUTCOMES` is a named constant, and `test_no_call_site_passes_a_computed_outcome` pins the shape the ledger can read — including a check that the ledger's own regex sees *every* statement-position `emit(` call.

A third, smaller one: my own `assert "re.split" not in src` guard was **spelled, not structural** — it matched `guard_co(re.split)_commands` and failed on correct code. Now `"re.split("`.

---

## How adoption will be measured

New `adoption-scan` REGISTRY row:

```python
{
    "id": "opencode-dispatch",
    "label": "opencode-dispatch — preflighted, detached opencode dispatch",
    "via": "tool", "tool": "opencode-dispatch", "opt_in": True,
    "outcomes": ["dispatched", "preflight-ok", "preflight-blocked", "error"],
    "impact_outcomes": ["preflight-blocked"],
    "impact_label": "silent no-op dispatches prevented (brief named a path outside --dir; `opencode run` auto-rejects and exits 0)",
}
```

Emitted through `scripts/collector/invocation.py::emit_invocation` — best-effort, swallows every failure, never changes the caller's exit code (exercised by `test_telemetry_never_changes_the_exit_code`, which makes the spool raise and asserts rc is unmoved).

`opt_in=True` because it is: nothing forces the operator to route through this instead of typing `opencode run`. The **impact** signal is `preflight-blocked` — every one of those is a dispatch that would have exited 0 having done nothing.

The ledger is pinned **three ways**: the literals at the `emit(` call sites, `OUTCOMES`, and the registry row must all be the same set, both directions.

---

## Verification I ran

- `scripts/tests/test_opencode_config.py` — **628 passed** after the resolver move.
- Targeted rebased run over every plausibly-affected target — **4,537 passed, exit 0**: `test_opencode_{config,engine,guard_plugin}`, `test_doc_path_rot`, `test_no_captured_{text,markup}`, `test_no_{public_ips,client_hostnames}`, `test_rules_size`, `test_skills_mapping_guard`, `test_subsystem_{recall,touch}`, `test_prune_skill_size`, `test_skill_audit`, `scripts/opencode/tests`, `scripts/session-analysis/tests`, `scripts/collector/tests`, `test_guard_core`.
- `nix-instantiate --parse nix/home.nix` — OK.
- Live end-to-end dispatch + live reproduction of the failure-2 error (above).
- All five new files confirmed `git ls-files`-tracked by a test, because 🔴 an unstaged new file is silently omitted from the flake and the switch still succeeds.

## What I could NOT verify

- **Failure 1 was not reproduced live.** I did not drive opencode into an actual `external_directory` auto-reject; the block is built from the recorded measurement (10 of 321 sessions with `model IS NULL` and 0 tokens) plus the config's own documented `ask` semantics. What *is* verified live is that preflight refuses the shape with rc 3 and names the path.
- **Failure 3's mid-run auto-reject was not reproduced live** either — same reason. The glob resolution itself goes through the engine-validated resolver, and the warning is advisory.
- **Not deployed.** `home.file` targets only change on a `home-manager switch`, so nothing is live on either host yet. Merged ≠ deployed. The `mkOutOfStoreSymlink` means that after one switch, later edits to the CLI or SKILL.md apply with no further switch.
- **I did not run the full hermetic gate** (instructed not to; the box was at load average 25 with siblings running). The targeted set above is what I can vouch for. I make no claim about the ~3,900-test `scripts/tests` target as a whole.
- **The `--file` array behaviour was verified on 1.18.18 only**, the version installed here.

## Sibling-agent interactions (flagged, not coordinated)

- **`scripts/tests/test_opencode_config.py`** is edited here (the resolver extraction: two function bodies removed, an import added). An agent working on **AGENTS.md size** is likely to touch that same file's §1 / `test_generated_agents_md_size_is_sane`, which is a different region — a textual conflict is unlikely and a semantic one is nil, but it is the file to watch.
- **`scripts/collector/opencode/activity-plugin.js` was NOT touched**, per the standing instruction — the **opencode session attribution** agent owns it.
- This branch was **rebased onto `210de3d`**, which is a sibling's `#581` fixing a `test_prune_skill_size` figure. That test was red at my original base (`4740e40`) for reasons unrelated to this change; it is green after the rebase.
- `scripts/opencode/opencode.jsonc` is **read**, never written, so the config-tuning surface is untouched.

## Open questions for you

- **No override flag on the path block.** A brief that legitimately spans two repos has to widen `--dir` or inline the content; there is no `--allow-path`. That is deliberate (an escape hatch on the one exact check turns it into the advisory check), but it is the decision most likely to annoy in practice. Say the word and I will add one.
- The skill is deployed to **Claude Code only**, not to `~/.config/opencode/skills/` — opencode dispatching opencode seemed like the wrong default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,17 @@ __pycache__/
 # committable.
 .envrc
 
+# Briefs and run logs written by `opencode-dispatch`. The brief is deliberately
+# placed INSIDE --dir (that is what makes an external-directory auto-reject
+# structurally impossible), so in devrc it lands in the working tree. It is
+# scratch work: it quotes task detail, paths and hostnames that belong to no
+# repo, and this repo is PUBLIC.
+#
+# Unanchored on purpose — `--dir` can be any subdirectory. The tool also writes
+# a self-ignoring `.gitignore` into the subdir on first use, so a target repo
+# that is not devrc is covered too; this line is the belt to that's braces.
+.opencode-dispatch/
+
 # opencode's per-repo runtime dir: node_modules, a generated package.json, and
 # a `commands` symlink into the gitignored .claude/. Its own inner .gitignore
 # covers the generated files but not the directory itself, so without this line
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -132,6 +132,7 @@ is preserved verbatim in `docs/LAYOUT.md` (not auto-loaded, and stale by design)
 | `scripts/mail-actions/` | `mailbox` | email-automation layer over the self-hosted inbox (**separate from activity telemetry**) |
 | `scripts/repo-cos/` | `repo-cos` | weekly repo "chief-of-staff" proposal digest + reply-driven exclusions |
 | `nix/i3/`, `nix/graphical.nix`, `scripts/bar-status-poll` | `bar` | i3 + i3status-rust bar, count blocks, dunst toasts |
+| `scripts/opencode/` | `opencode` | dispatch a task to the headless opencode agent (`opencode-dispatch`), + its config/agents/guard plugin |
 
 Repo-level facts that are NOT in any skill — they live here on purpose:
 - 🔴 **A NEW file must be `git add`ed or the flake silently omits it from the deploy.** Applies to every managed path: a new skill, a new `reference/*.md` inside one, an extension file, a hook or a test. The switch succeeds and the file simply is not there.
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -979,6 +979,26 @@ in
     config.lib.file.mkOutOfStoreSymlink "${workspace}/devrc/scripts/browser-bridge/SKILL.md";
   home.file.".claude/skills/browser/browser".source =
     config.lib.file.mkOutOfStoreSymlink "${workspace}/devrc/scripts/browser-bridge/browser";
+  # `opencode` skill — the SAME deliberate exception as `browser`/`dl-router`
+  # above, for the same reason: its source of truth is the opencode subsystem in
+  # THIS repo (scripts/opencode/), not devrc/claude/skills/, and the CLI reads
+  # scripts/opencode/opencode.jsonc + scripts/opencode/lib/ + scripts/claude-hooks/
+  # relative to its own resolved __file__. A store copy would resolve those into
+  # /nix/store and read a FROZEN permission block — so preflight would answer
+  # from whatever config was current at the last switch, which is exactly the
+  # stale-artifact failure mode CLAUDE.md warns about. Out-of-store keeps the
+  # resolver and the config it resolves against as one live tree.
+  #
+  # 🔴 Only these two paths are symlinked; `lib/` is NOT deployed and must not
+  # be. The CLI reaches it through the repo checkout via `__file__`, which is
+  # the same way `dl-route` reaches its siblings.
+  home.file.".claude/skills/opencode/SKILL.md".source =
+    config.lib.file.mkOutOfStoreSymlink "${workspace}/devrc/scripts/opencode/SKILL.md";
+  home.file.".claude/skills/opencode/opencode-dispatch".source =
+    config.lib.file.mkOutOfStoreSymlink "${workspace}/devrc/scripts/opencode/opencode-dispatch";
+  # …and on PATH, like dl-route, so the skill body's commands are copy-pasteable.
+  home.file.".local/bin/opencode-dispatch".source =
+    config.lib.file.mkOutOfStoreSymlink "${workspace}/devrc/scripts/opencode/opencode-dispatch";
   # Claude Code hooks managed here (the script only — the settings.json
   # registration is per-host/unmanaged, as for bash-guard.py above, whose script
   # is likewise managed now). audit-pr-nudge fires
--- a/scripts/opencode/SKILL.md
+++ b/scripts/opencode/SKILL.md
@@ -1,0 +1,124 @@
+---
+name: opencode
+description: Dispatch a task to opencode — the separate headless CLI agent — instead of burning this session's context on it. Preflights the brief first (a brief naming paths outside --dir makes `opencode run` auto-reject and exit 0 having done nothing), then launches detached. Use for "dispatch opencode", "call opencode", "run this with opencode", "hand it to flash/mimo/pro", opencode token efficiency. NOT the Agent tool — "dispatch a subagent" means that, not this.
+---
+
+# opencode dispatch
+
+Hand a bounded task to **opencode**, a separate headless agent CLI with its own
+context window. The point is *your* context budget: seven weeks of opencode cost
+**$11.42** total, median dispatch **$0.014** — money is not the variable.
+
+## The whole invocation
+
+```bash
+opencode-dispatch run --dir <PROJECT-DIR> --title "<short title>" -m flash <<'BRIEF'
+<the complete task specification>
+BRIEF
+```
+
+That is it. Do **not** re-derive the CLI — `opencode --version`, `--help`,
+`run --help` and `models` were re-run from scratch in **nine** separate sessions
+and answered nothing that is not on this page.
+
+- The brief comes in on **stdin** (or `--brief FILE`). It is written **into
+  `--dir`** at `.opencode-dispatch/<ts>-<slug>.md`, which self-ignores in
+  whatever repo that is.
+- `--dir` is required and is opencode's project root. Every path the brief names
+  must be under it.
+- The run is **detached**. The call returns immediately and prints a log path.
+  There is no foreground mode and no flag to ask for one.
+
+### Models
+
+| alias | model |
+|---|---|
+| `flash` | `openrouter/deepseek/deepseek-v4-flash` |
+| `mimo` | `openrouter/xiaomi/mimo-v2.5` |
+| `pro` | `openrouter/deepseek/deepseek-v4-pro` |
+
+Omit `-m` to take the `opencode.jsonc` default (what 18 of 24 measured
+dispatches did). A full `provider/vendor/model` id passes through; an unknown
+bare word is a usage error, not a silent pass-through.
+
+### Other flags, in order of how often they actually get used
+
+`--dir` 18 · `--title` 7 · `--file` 7 (extra attachments; the brief is attached
+automatically) · `-m` 6 · `--auto` 4 · `-c` 2 (continue the last session).
+
+## Check a brief without dispatching
+
+```bash
+opencode-dispatch preflight --dir <PROJECT-DIR> --brief <FILE>   # or stdin
+```
+
+`--json` for a machine-readable report. Exit **3** means the brief was refused.
+
+## Reading the result
+
+`status` and `watch` are **deliberately not built** — they would depend on
+opencode's internal SQLite schema, which carries no stability guarantee. That
+dependency is worth taking only once this skill is proven to get used. For now:
+
+```bash
+tail -n 60 <the log path printed on dispatch>
+```
+
+## Why this exists — four measured failure modes, three success-shaped
+
+1. 🔴 **`external_directory: "ask"` → headless auto-reject → exit 0, no work.**
+   The brief named a path outside `--dir`. `opencode run` **auto-rejects** an
+   `ask` (only the interactive TUI prompts). Fingerprint in opencode's own
+   store: 10 of 321 sessions with `model IS NULL` and 0 tokens. Happened twice,
+   in both directions — a read, then a write.
+   *Closed two ways:* the brief is written **inside** `--dir`, and preflight
+   **refuses (rc 3)** if the brief's text names any absolute path outside it.
+2. **`--file` is an array option** and swallowed the message as a second
+   filename (`Error: File not found: Execute the task described in…`). The
+   message positional now always comes first, pinned by a test on its argv index.
+3. **`permission.bash` `ask` → auto-reject mid-run.** opencode needed
+   `kubectl exec … psql`, was auto-rejected, and the dispatch was abandoned.
+   Preflight **warns** (never blocks) and names the glob.
+4. **20% of opencode sessions exceed the Bash tool's hard 600,000 ms ceiling**
+   (p90 2508s, p95 9119s); one died at exactly 600s with `Exit code 143`. Hence
+   detached-always.
+
+## When preflight refuses
+
+It refused because a path in the brief is not under `--dir`. Fix the **brief or
+the `--dir`**, never the check — three options, in order of preference:
+
+1. move the file under `--dir`;
+2. widen `--dir` to a directory containing both;
+3. inline the content into the brief instead of pointing at it.
+
+There is no override flag. The failure it prevents is a silent exit 0, which is
+the one shape nobody notices.
+
+A `⚠ WARN` line is different — it is advisory and the dispatch proceeds. It goes
+through a parser that deliberately over-matches, so blocking on it would build a
+gate people learn to click through.
+
+## What this is NOT
+
+- **Not a Claude subagent.** "dispatch a subagent to implement X" is the **Agent
+  tool**. Only the literal word *opencode* routes here.
+- **Not the `browser` skill.** "dispatch a browser agent" is that one.
+- **No auto-retry.** The measured failures are corrective, not transient:
+  re-running a known-bad brief is not a fix. Fix the brief, dispatch again.
+
+## Writing a brief that lands
+
+- State the goal, the acceptance check, and where the code is — **relative to
+  `--dir`**.
+- Put commands in fenced ```bash blocks so preflight can see them.
+- Assume no follow-up questions are possible. opencode is unattended; every
+  `ask` is an auto-reject, not a prompt.
+
+## Files
+
+`scripts/opencode/` — `opencode-dispatch` (the CLI), `lib/oc_permissions.py`
+(the shared resolver `scripts/tests/test_opencode_config.py` also pins),
+`lib/brief_scan.py` (both scanners), `opencode.jsonc` (the permission block
+preflight reads). Config, agents and the guard plugin are documented in
+`scripts/opencode/README.md`.
--- a/scripts/opencode/SKILL.md
+++ b/scripts/opencode/SKILL.md
@@ -46,6 +46,10 @@ bare word is a usage error, not a silent pass-through.
 `--dir` 18 · `--title` 7 · `--file` 7 (extra attachments; the brief is attached
 automatically) · `-m` 6 · `--auto` 4 · `-c` 2 (continue the last session).
 
+**`--file` attachments are containment-checked too** — each must resolve under
+`--dir`, exactly like a path named in the brief. `preflight` accepts `--file`
+as well, so you can check them without dispatching.
+
 ## Check a brief without dispatching
 
 ```bash
@@ -85,8 +89,10 @@ tail -n 60 <the log path printed on dispatch>
 
 ## When preflight refuses
 
-It refused because a path in the brief is not under `--dir`. Fix the **brief or
-the `--dir`**, never the check — three options, in order of preference:
+It refused because a path — in the brief, or a `--file` attachment — does not
+resolve under `--dir`. The offender is printed as written *and* resolved, tagged
+`[brief]` or `[attachment]`. Fix the **brief or the `--dir`**, never the check —
+three options, in order of preference:
 
 1. move the file under `--dir`;
 2. widen `--dir` to a directory containing both;
@@ -94,6 +100,14 @@ the `--dir`**, never the check — three options, in order of preference:
 
 There is no override flag. The failure it prevents is a silent exit 0, which is
 the one shape nobody notices.
+
+Read the three verdict lines literally — they are different claims:
+
+- `external paths : none — all N examined resolve under --dir` — checked, clean.
+- `external paths : NOT EXAMINED — no resolvable path in the brief, and no
+  attachments` — **not** an all-clear. Nothing was there to check.
+- `UNMEASURED paths : N` — variable-built paths that cannot be decided
+  statically. Neither blocked nor cleared; check them yourself.
 
 A `⚠ WARN` line is different — it is advisory and the dispatch proceeds. It goes
 through a parser that deliberately over-matches, so blocking on it would build a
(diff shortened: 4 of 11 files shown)


ZacxDev:
## Round 2 — the six audit 🟡s are closed (`db69b91`)

Three of them were one shape: **the tool printed a claim wider than what it checked.** All six reproduced first, then fixed, then swept.

### 🔴 The three confidently-wrong outputs

**1. `--file` attachments bypassed the check while the report said otherwise.** Reproduced: `run --dir <proj> --file <outside>/extra.md` printed *"external paths : none — every path is under --dir"* and handed opencode the file. Attachments now run through the **same `_judge` predicate** as the brief — one predicate, so the two cannot disagree — and are tagged `[attachment]` in the report. `preflight` accepts `--file` too, so you can check without dispatching. Attachments are judged **before** the brief is installed, so a bad `--file` cannot leave a stray brief behind (verified: no `.opencode-dispatch/` is created). Whether opencode itself raises an `external_directory` ask for an attachment is still **unverified** — irrelevant, since the printed claim was false either way.

**2. An empty brief dispatched and reported success.** Reproduced: `printf '' | … run` → 0-byte brief, rc 0, `DISPATCHED`, `outcome=dispatched`. Now **rc 2**, whitespace-only included, nothing written to `--dir`.

**3. The block was walkable by relative path, and the docs invited it.** Reproduced: `"Read ../outside/extra.md and apply it."` → `paths examined : 0` plus an unconditional all-clear. Now resolved against `--dir` (opencode's cwd) by the same predicate. Scoped to tokens carrying a `..` **segment** — the only relative tokens that can escape — with a negative-control battery over `v1..v2`, `origin/main..HEAD`, `HEAD~2..HEAD` and prose ellipsis, and an in-dir control (`sub/../src/x.py` must **not** block).

🔴 **And the sentence is now conditional.** Three distinct verdict lines, each pinned as a **whole normalised string**:

```
  external paths    : none — all 2 examined resolve under --dir
  external paths    : NOT EXAMINED — no resolvable path in the brief, and no attachments
  UNMEASURED paths     : 1 — built from a variable, so containment …
```

SKILL.md now tells the reader those are different claims, and that `NOT EXAMINED` is not an all-clear.

### The other three

**4. `${VAR}/path` false-blocked; `$VAR/path` silently passed.** Opposite directions, both the likeliest content of a brief written here. Neither verdict was honest, so a variable-built path is now **UNMEASURED** — not blocked, not counted clean, named for the operator — matching how drift-check treats `absent`/`fetch failed`. Verified across all four measured spellings.

**5. The `guard_core` channel was a silent zero.** Failures now surface as `COULD NOT MEASURE: guard_core.evaluate raised …`, with **both** controls: a positive one (`rm -rf /` must produce a verdict) and a broken-channel one.

**6. The `SYSTEM_ALLOW` comment was false on NixOS.** The allowlist tested the realpath, so `/etc/hosts` → `/nix/store/…-hosts` → allowlisted. It now matches the path **as written**; `is_allowlisted("/etc/hosts") is None` and `/etc/hosts` blocks. A symlink out of a listed prefix now blocks — fail-closed, correct for an enumeration whose premise is "unlisted is an offender by default".

**Nits:** a missing `--brief` file raised an uncaught `FileNotFoundError` (rc 1, undocumented, **no telemetry**, so adoption-scan's `error` bucket undercounted) → now rc 2 with telemetry, asserted both ways. The `⚠ WARN` header no longer claims "auto-rejects an `ask`" when every row is a `deny`.

### Exit codes, measured with no pipeline to destroy them

| case | rc | | case | rc |
|---|---|---|---|---|
| `--file` outside `--dir` | **3** | | `/etc/hosts` | **3** |
| empty brief | **2** | | missing `--brief` file | **2** |
| whitespace-only brief | **2** | | clean in-dir control | **0** |
| relative `../` escape | **3** | | clean + in-dir attachment | **0** |
| `${VAR}` / `$VAR` path | **0** | | bad `--dir` (positive control) | **4** |

The first pass of this table read `rc=0` everywhere — my harness used `${PIPESTATUS[0]}` of `printf | python3 | sed`, i.e. **printf's** status. Re-measured with no pipe, and with a positive control so a uniform reading cannot pass for a measurement.

### Verification

- **Red at `483c088`, per test, by name: 27 red.** (`test_every_new_file_is_git_tracked` ×5 is red as an artefact of the archived tree having no `.git` — not evidence, and flagged as such. `test_the_base_ruleset_prepends_the_builtin_catch_all` is green at base by design: it is a strengthened **invariant guard**, not a regression test.)
- **22 mutants, 22 killed** — narrowest expression, `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged per mutant, no `-x`, every anchor asserted unique before application, all three files `cmp`-verified byte-identical afterwards. Exact-1 killers on every finding's own guard (N1, N3, N5, N6, N7, N8, N13, N14, N16, N18, N19, N20, N21).
- **All 55 test names confirmed individually `-k`-selectable**, with a positive control on that harness.
- **Targeted regression: 4,564 passed, exit 0.**

### 🔴 Three defects the round-2 sweep found in its own work

1. **A dead `if "/" not in cand` branch** whose comment claimed to be what kept `...` out. It was **unreachable** — `rstrip(_TRAILING_PUNCT)` already empties a bare `..`. Deleted; the real mechanism is now named and swept by a mutant on `_TRAILING_PUNCT` itself.
2. **A verdict test that read its expectation from the constant it tested** (`D.VERDICT_CLEAN.format(n=2) in lines`) — mutate the constant and the expectation mutates with it, so a reworded claim **survived a fully green suite**. The sentences are now literals in the test file, two-way pinned against the CLI's constants.
3. **My `-k` selectability harness had a broken positive control**, because it grepped for pytest's `no tests ran` when pytest actually prints `no tests collected`. `unselectable=0` was meaningless until fixed — the "parsing output makes its format a dependency you did not pin" trap, in my own instrument.

A fourth, in the battery itself: the N7 mutant's anchor contained both quote kinds and a backtick, which broke the script's shell quoting and **silently truncated the run after N6**. Caught by reading the tail rather than grepping for `SURVIVED`.

### Still not verified

- **Failures 1 and 3 were never reproduced live against opencode** — I did not drive it into an actual `external_directory` or `permission.bash` auto-reject. Both blocks rest on the recorded measurements plus the config's documented semantics. What *is* verified live is that preflight refuses each shape with rc 3 and names the offender.
- **Whether opencode raises an external-directory ask for a `--file` attachment** specifically.
- **Not deployed.** No `home-manager switch` has run; merged ≠ deployed.
- **The full hermetic gate was not run** (instructed not to). I make no claim about `scripts/tests` as a whole beyond the targeted set.

### Known-untested, deliberately left alone (per the audit)

`canon()`'s `expanduser`/`realpath` behaviour, `SHELL_INFO`'s `""` entry, `wildcard_match`'s `re.S`/`?` handling, `resolve_with_rule`'s unreachable `"ask"` fallback. None is live; each is noted rather than chased.

### Sibling interactions

Unchanged from round 1 — `scripts/tests/test_opencode_config.py` is the file to watch against the AGENTS.md-size agent; `activity-plugin.js` untouched. **Not rebased** since `483c088`, per instruction.

ZacxDev:
## Round 3 — the three integration-tree reds are fixed (`7c67e9c`)

All three were invisible on this branch's own gate. Not rebased; `test_opencode_engine.py` left alone as #580's.

### 🔴 1 — the nix sandbox has no `.git`, and my assertion read that as "untracked"

`nix build .#checks.x86_64-linux.pytests` builds from `/build/src`, a copy with no git dir, so `git ls-files` exits **128**. Confirmed directly:

```
$ git -C <git-free tree> ls-files --error-unmatch scripts/opencode/SKILL.md
fatal: not a git repository (or any of the parent directories): .git
rc=128
```

Refactored to `file_ship_problems(rel, root, git_present)`, keeping **both** halves because neither covers both tiers — **existence** is what means something inside the sandbox (the store copy is built from tracked files only, so an untracked file simply would not be there), **trackedness** is what means something on a dev host. The git half is *conditional*, not skipped, so it cannot go vacuous where it works.

**The guard is proven, not assumed** — the deliverable, measured on a faithful reproduction (`git archive HEAD | tar -x`, no `.git`, exactly what the builder presents):

| control | result |
|---|---|
| **A.** guard removed, `.git`-free tree | **7 failed** — the 5 `test_every_new_file_ships` failures reproduce verbatim, plus my 2 new regression tests |
| **B.** guard present, same tree | **77 passed** |
| **C.** guard present, **with** `.git`, one file deliberately untracked | **1 failed**, on exactly that file — the real hazard survives |

C is the one that matters: the guard did not throw away the check it wraps. And `test_a_present_file_in_a_git_free_tree_reports_NOTHING` reproduces tier A in an **ordinary checkout**, so deleting the guard now fails on a dev host instead of only in CI, which is where this hid.

### 2 — the shebang scanner: restructured, not allowlisted

**Chosen deliberately, and here is why.** The allowlist route is *closed by that file's own* 🔴 — "Nothing here may be `/usr/bin/env`" — so taking it would mean changing the rule rather than satisfying it, and its docstring names "add an entry to go green" as the anti-pattern.

And my assertion never needed that string. It proves "an enumerated `/usr` prefix returns a reason", which any `/usr` path demonstrates; it is now a named constant, `ALLOWLISTED_USR_PATH`, with the reasoning in a comment beside it.

On whether the scanner is too broad: it genuinely cannot tell a path being *mentioned* from one being *written as a shebang*. But making it able to means parsing Python — the exact trap `guard_core`'s DESIGN NOTE records, where three rounds of a regex "is this inert text?" helper each shipped a fresh hole. A fixed-string gate that fails closed for one retype is the better trade, and it is the same stance `guard_core` takes about quoted commit messages. So: **narrow my fixture, leave the repo-wide gate alone.**

Red-then-green measured: restore the literal → `FAILED test_no_test_writes_a_usr_bin_env_shebang_at_runtime`, naming `test_dispatch.py:312`; restore the fix → 9 passed.

### 3 — floor

Re-pinned to **`"scripts/opencode/tests|86"`**, copied verbatim from the gate's drift message on the merged tree. **Not computed here** — this branch alone collects 95 and would print `91`; I used the merged-tree number as instructed. Accounting is in the table comment (what grew, by how much, that `HERMETIC_TARGETS` needs no entry and `EXPECTED_SKIPS` is untouched, and that the number is the gate's own).

Exercised through the gate's **own** `_suggested_floor` function lifted verbatim, rather than re-deriving the rule:

```
collected = 95
  floor 14  -> ceiling 74  : RED (gate would say: raise to "scripts/opencode/tests|91")
  floor 86  -> ceiling 146 : GREEN
  floor 500 -> ceiling 625 : RED   (positive control on this harness)
```

`--check-floors` passes (two-way pin intact). Headroom note: ceiling for 86 is 146, so the merged tree's count has room to grow past 95 before this needs touching again.

### Mutation: 12 mutants, 12 killed

11 on a dev host; the 12th is **unkillable there by construction** and I want to be explicit about it rather than quietly count it. `git_dir_present(ROOT)` hardcoded to `True` at the use site: on a tree that *has* git and *has* tracked files, the correct flag and the constant agree, so no dev-host test can separate them. It is killed by the tier-A control — **5 failed / 72 passed** mutated, **77 passed** clean, in the `.git`-free tree.

**Two survivors from the first pass were real defects and are fixed here:**

1. **`is_file()` weakened to `exists()` survived** — nothing had ever put a non-file at a shipped path. A *directory* standing where `SKILL.md` should be would have read as clean while the deploy has no skill. Now pinned by `test_a_DIRECTORY_at_the_path_is_not_a_shipped_file`.
2. **A module-level `GIT_DIR_PRESENT` could be hardcoded and survive**, because the obvious pin (`assert GIT_DIR_PRESENT == (ROOT/".git").exists()`) is a fixture that can only produce the constant it names. **The constant is deleted rather than pinned** — the structural fix: there is nothing left to hardcode, and every caller computes the tier from the tree in front of it. The probe itself is now a function, tested both ways from fixtures it builds, so `True`/`False`/`is_dir()` mutants all die. (`is_dir()` matters: `.git` is a *file* inside a worktree, and this repo is developed in worktrees — that mutant would have disabled the tracking half everywhere.)

### Verification summary

- Red-then-green measured for all three fixes, by name.
- 12 mutants / 12 killed, narrowest expression, `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged per mutant, whole-file runs, no `-x`, anchors asserted unique, file `cmp`-verified afterwards.
- **61 test names individually `-k`-selectable**, positive control on that harness working.
- **Targeted regression: 4,570 passed.** `test_runtime_shebangs.py` 9 passed. `--check-floors` PASS.

### What I did not do

- **Did not run the full hermetic gate** — you re-gate the integration tree.
- **Did not rebase**, did not merge, did not touch `test_opencode_engine.py`.
- Everything under "Still not verified" in the round-2 comment still stands: failures 1 and 3 have never been reproduced live against opencode, and nothing is deployed.

PR #586, opened 2026-08-20
skills: browser validation prompts retyped 3-5.6 KB of safety rails each — cite them instead

ZacxDev:
## What

Two skill-body changes from the prompt-optimization handoff, the gate fix my own edit triggered, and the audit round that reworked the guards.

> ⚠ An earlier version of this description claimed the nine rails were pinned "verbatim". That was false — only their headlines were pinned. See the [correction comment](https://github.com/innovation-upstream/devrc/pull/586#issuecomment-5351867938); fixed in `08e79b4`.

**1. `browser` — a standing validation contract that prompts CITE instead of restating.**

Measured (`activity.events`, `source='opencode'`, `kind='prompt'`, 2026-08-16..19): of 125 prompts, **8** named the bridge / the `browser` skill / `browser agent`, and **7 of those 8 ran 3.0–5.6 KB**. The task-specific part of each was a few hundred bytes; the rest was the same scaffolding retyped — READ-ONLY prohibitions, "don't log out", "report and stop", "don't screenshot", a step budget, a report format. The copies had already drifted apart from each other, which is the same defect one generation earlier.

New `scripts/browser-bridge/reference/validation-prompt.md`: a fill-in template, **nine standing rails**, and an **inlineable rails block**, synthesised from the real corpus (this repo is public — no captured text, no client names, no real URLs).

🔴 **The capability split is the load-bearing part.** The rails ride on a filesystem path. A Claude Code subagent can `Read` it; **`browser agent` cannot** — its model gets one typed tool with an 11-op browser-only surface and the agent def denies `bash`/`read`/`edit`/`write`/`webfetch`, enforced by a fail-closed runtime gate before the model is invoked. A cited prompt would reach it carrying **zero** rails while still reading complete, with `click`/`type`/`key`/`nav`/`eval` live on the operator's logged-in Brave. So the file leads with that split and carries the inline block to paste instead.

`SKILL.md` gets one reference-table row and **pays for it**: the file had 48 B of slack under `tests/test_skill_size.py`, so the duplicated `browser agent` example (also in `reference/agent.md`) and a redundant paths note were evicted. 11,990 → 12,007 B; headroom 281 vs the 250 floor — i.e. ~31 B of slack, so the next row needs its own eviction.

**2. `clickup` — the handoff's premise was already false.**

The task was to cut ~4,800 chars of CLI reference tables redundant with `node query.mjs --help`. **Those tables do not exist and never have** — SKILL.md has carried zero command tables since #438 and already points at `showUsage()` as the source of truth. The "~4,800 chars" was the file's *total* size.

So 4,800 → 800 is unreachable without deleting the non-obvious gotchas the same design says to keep. What *is* evictable is the maintainer-only half — how to edit so the change deploys, the nix-built `node_modules`, the test suites — moved to `reference/maintaining.md`. **4,934 → 3,821 B (−23%).** Every "what the usage lines can't tell you" gotcha is unchanged.

**3. `test_prune_skill_size.py`** — that module checks its own docstring against live measurements and cites browser-bridge's size as its exemplar. My row moved it, so the figure #581 had just re-measured went stale again. Re-measured, not relaxed.

## Test coverage

`scripts/browser-bridge/tests/test_validation_prompt.py` (34 tests) pins:

- **the rails** — each rail's WHOLE normalised block, in order, so a reword fails deliberately;
- **the route** — set equality between `reference/*.md` and SKILL.md's table, so a 14th topic is covered without editing the module;
- **the route's trigger cell** — a row describing a different topic routes nobody, which is the #567 defect itself;
- **the reader-capability split**, including that it appears *before* the template;
- **the inline block** — its nine rails and each operative prohibition;
- **the template** — its slots, exactly one fence, and that the path it tells every prompt to cite actually resolves to this file;
- **rail 6's scope**, asserted inside rail 6's own block.

Every check slices its section first; a document-wide substring search is satisfied by the same words appearing anywhere.

**Mutation battery** under `PYTHONDONTWRITEBYTECODE=1`, positive control green each batch. All seven mutants the audit reported as surviving now die, plus seven more against the new guards. **Survivors: none.**

| mutant | killed by |
|---|---|
| rail 3 body inverted | `test_every_standing_rail_matches_its_pinned_block` |
| rail 7 body → "`activate`" | ledger + control |
| rail 6 caveat → `## Trivia` | `test_rail_six_keeps_its_delegation_scope_INSIDE_rail_six` |
| citation → dead path | `test_the_template_cites_a_path_that_resolves_to_this_file` |
| route trigger → other topic | `test_the_route_row_trigger_still_describes_this_file` |
| `_rail_blocks` → `[]` | parser guard + **both** controls |
| `_reference_table_rows` → `[]` | parser guard + control + routing + trigger |
| all nine bodies gutted | ledger |
| inline "NEVER log out" removed | `…offers[NEVER log out]` |
| inline rail 7 removed | `…offers[Never activate]` + numbering |
| capability section deleted | capability guard + slicer |
| capability section moved below template | capability guard (ordering) |
| template gutted + decoy fence | 10 guards |
| new reference topic, no route row | `test_every_reference_topic_is_routed_from_skill_md` |

## Gate

Full hermetic set on the rebased tree:
- pytest `TOTAL collected=12861 passed=12860 skipped=1 failed=0` — `RESULT: PASS`
- node `TOTAL suites=4 files=33 tests=1118 pass=1118 fail=0` — `RESULT: PASS`

The `test_prune_skill_size.py` failure was caught by the gate, not by review; a control run confirmed that module is green at `origin/main`, so it was this branch's delta, not a pre-existing red.

## Not verified

The template has **not** been exercised on a real browser validation run — that needs a merge + `ship.sh` (clickup is a store symlink and needs a switch; browser's `SKILL.md` is an `mkOutOfStoreSymlink` and goes live on pull).

Also noted by the audit and **not** fixed here, as it is a repo-wide gap rather than this PR's: the content-leak gates cover JSON/JSONL/JSONC and `.html`/`.txt`, but **not `.md`** — so the two new markdown files landed under no automated leak gate. Both were read manually and are clean, and this PR actually *removes* one real hostname from `SKILL.md`.

### DIFF
--- a/claude/skills/clickup/SKILL.md
+++ b/claude/skills/clickup/SKILL.md
@@ -70,36 +70,15 @@ node query.mjs edit-page <doc_id> <page_id> --file /tmp/page.md
 - **Internal-API commands** (`inbox-*`, `doc-comments`) need a JWT in the account, not
   just a token — see setup below.
 
-## Going deeper
+## Going deeper — load ONE only when its trigger fires
 
-Paths below are relative to this skill directory. Like every other skill, it is
-**deployed by home-manager from `~/workspace/devrc/claude/skills/clickup/`**: edit
-THERE, then `home-manager switch --flake ~/workspace/devrc --impure` (or
-`scripts/ship.sh` for both hosts). What lands at `~/.claude/skills/clickup/` is a
-tree of read-only `/nix/store` symlinks — editing it directly is impossible, and a
-`git pull` alone changes nothing until you switch.
-
-- **All mutable state lives in `$XDG_STATE_HOME/clickup`** (fallback
-  `~/.local/state/clickup`) — credentials included. A write next to the code is
-  `EROFS`. Setup, `accounts.json`, multi-account and the JWT fields →
-  `~/.claude/skills/clickup/reference/setup.md`.
-- **`node_modules` is BUILT by nix**, not installed: `nix/pkgs/clickup-node-modules.nix`
-  materialises it from `package-lock.json` and links it in at the skill root. To
-  change a dependency, edit `package.json` + `package-lock.json`, set `npmDepsHash`
-  to `lib.fakeHash`, build, copy the `got:` hash back — never guess it.
+- **Credentials, `accounts.json`, multi-account, the JWT fields, first-time setup**
+  → `~/.claude/skills/clickup/reference/setup.md`. State lives in
+  `$XDG_STATE_HOME/clickup` (fallback `~/.local/state/clickup`), never next to the
+  code — a write there is `EROFS`.
 - **Hand-rolling raw `api.clickup.com` requests** → `~/.claude/skills/clickup/reference/raw-api.md` — read it
   first: a view id is not a list id, dashboard views can't be queried for tasks, and
   both task endpoints paginate (a single-page read made a 67-ticket queue look like 30).
-
-## Tests
-
-The hermetic gates are `node:test` suites, run by devrc's node gate
-(`bash scripts/run-node-tests.sh .` from the devrc checkout, and
-`nix build .#checks.x86_64-linux.nodetests` in CI). Standalone still works:
-
-```bash
-node test/help-coverage.test.mjs      # hermetic; pins showUsage() completeness
-node test/state-paths.test.mjs        # hermetic; pins state OUT of the skill dir
-node test/js-source.test.mjs          # hermetic; controls for the source scanner
-node test/smoke-test.mjs --readonly   # live API, needs credentials — NOT in any gate
-```
+- **CHANGING this skill** — where to edit so the change deploys, the nix-built
+  `node_modules` + `npmDepsHash`, and the test suites →
+  `~/.claude/skills/clickup/reference/maintaining.md`.
--- a/claude/skills/clickup/reference/maintaining.md
+++ b/claude/skills/clickup/reference/maintaining.md
@@ -45,6 +45,5 @@ node test/js-source.test.mjs          # hermetic; controls for the source scanne
 node test/smoke-test.mjs --readonly   # live API, needs credentials — NOT in any gate
 ```
 
-**Adding a command? Add it to `showUsage()`** — `test/help-coverage.test.mjs`
-fails if any dispatchable command is missing from the help, or if any printed
-command cannot dispatch. That test is why `SKILL.md` does not list commands.
+Adding a command is governed by `SKILL.md` → "Finding a command", not restated
+here — one rule, one place.
--- a/claude/skills/clickup/reference/maintaining.md
+++ b/claude/skills/clickup/reference/maintaining.md
@@ -1,0 +1,50 @@
+# Maintaining the clickup skill — deploy, dependencies, tests
+
+**Load this when:** you are CHANGING this skill (adding a command, touching a
+dependency, editing `query.mjs`) · a test failed · an edit to
+`~/.claude/skills/clickup/` was rejected as read-only · you added a dependency
+and need the `npmDepsHash`.
+
+Nothing here is needed to USE the skill — that is `SKILL.md`.
+
+## Where to edit, and why an edit "does nothing"
+
+Like every skill, this one is **deployed by home-manager from
+`~/workspace/devrc/claude/skills/clickup/`**. Edit THERE, then
+`home-manager switch --flake ~/workspace/devrc --impure` (or `scripts/ship.sh`
+for both hosts).
+
+What lands at `~/.claude/skills/clickup/` is a tree of read-only `/nix/store`
+symlinks — editing it directly is impossible, and a `git pull` alone changes
+nothing until you switch. 🔴 A **new** file must be `git add`ed or the flake
+silently omits it from the deploy: the switch succeeds and the file is just not
+there.
+
+## Dependencies
+
+`node_modules` is **BUILT by nix**, not installed:
+`~/workspace/devrc/nix/pkgs/clickup-node-modules.nix` materialises it from
+`package-lock.json` and links it in at the skill root. To change a dependency,
+edit `package.json` + `package-lock.json`, set `npmDepsHash` to `lib.fakeHash`,
+build, then copy the `got:` hash back — never guess it.
+
+All mutable state lives in `$XDG_STATE_HOME/clickup` (fallback
+`~/.local/state/clickup`), credentials included; a write next to the code is
+`EROFS`. Pinned by `test/state-paths.test.mjs`.
+
+## Tests
+
+The hermetic gates are `node:test` suites, run by devrc's node gate
+(`bash scripts/run-node-tests.sh .` from the devrc checkout, and
+`nix build .#checks.x86_64-linux.nodetests` in CI). Standalone still works:
+
+```bash
+node test/help-coverage.test.mjs      # hermetic; pins showUsage() completeness
+node test/state-paths.test.mjs        # hermetic; pins state OUT of the skill dir
+node test/js-source.test.mjs          # hermetic; controls for the source scanner
+node test/smoke-test.mjs --readonly   # live API, needs credentials — NOT in any gate
+```
+
+**Adding a command? Add it to `showUsage()`** — `test/help-coverage.test.mjs`
+fails if any dispatchable command is missing from the help, or if any printed
+command cannot dispatch. That test is why `SKILL.md` does not list commands.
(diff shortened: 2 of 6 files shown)


ZacxDev:
## Correction to this PR's description (audit round 1)

An adversarial audit found the central verification claim in the body above **false as written**, so correcting it here rather than only editing the body — a reviewer may already have read it.

**What the body claimed:** "the nine rails as an ordered **verbatim** ledger".

**What was actually pinned:** each rail's bold *headline*. The regex was `^\d+\.\s+(\*\*.+?\*\*)` — non-greedy, so it captured the lead and nothing else. Every rail's **body** was free. Measured, each of these SURVIVED a fully green 17/17:

| mutant | result |
|---|---|
| rail 3 body → "Feel free to Connect / Follow / Message / Save, submit any form, change account or notification settings, and log out when you are done" | **SURVIVED** |
| rail 2 body → "Reuse whichever tab is already active; `nav` it to the site under test" | **SURVIVED** |
| rail 7 body → "`activate` the tab to fix it" (a 🔴 RULES.md screen-theft hazard) | **SURVIVED** |

Since the whole justification for centralising the rails is "a rail reworded out of this file is dropped from every run that cites it", that was the gap that mattered most.

A second deploy-blocking finding: the contract named **`browser agent`** as an audience, but that model gets one typed tool with an 11-op browser-only surface and the agent def denies `bash`/`read`/`edit`/`write`/`webfetch`, enforced by a fail-closed runtime gate before the model is invoked (`reference/agent.md:167-199`). It **cannot open the path carrying the rails** — so a cited prompt reached it with **zero** rails while still reading complete, with `click`/`type`/`key`/`nav`/`eval` live on the operator's logged-in Brave.

Both are fixed in `08e79b4`, along with nine should-fix items (rail 6's scope was document-wide and survived being moved to a `## Trivia` section; the template's cited path, the route row's trigger text and the fence identity were all unpinned; both "negative controls" were tautologies a dead parser satisfied). The body has been updated to match. Full detail is in the commit message.

**Corrected accounting**, also from the audit: the "contract absent → 6 red" line counted test *functions*; it is **14 test cases**. The gate figures quoted originally were from an earlier rebase.

ZacxDev:
## Audit round 2 (delta re-audit of `08e79b4`) — one 🔴 found and fixed

Verdict on the round-1 fixes: **8 of 11 claims actually fixed, 3 partially, 0 not-fixed, 0 made-worse.** But round 1's *new* artifact — the inline rails block — was the weakest surface in the file, and it carried a deploy-blocker.

### 🔴 The inline block told `browser agent` to call two ops it does not have

Inline rail 2 said "`open` a tab this session owns"; rail 8 said "close the tab you opened". `reference/agent.md:192-194` pins that model's surface at **11 ops** — `text`/`html`/`eval`/`nav`/`screenshot`/`frames`/`click`/`type`/`key`/`wake`/`whoami` — containing **neither**, and `:152-159` says the *wrapper* opens the tab, forces it via env, and closes it on every exit path.

That block is the entire safety payload for the one audience that cannot cite the file, and its own rail 5 says a failure is a finding and to **STOP**. So the recommended paste could abort the run it exists to protect, on its first instruction.

Fixed, and guarded structurally: the allowed op set is now **parsed out of `agent.md`** rather than restated, so widening the agent's surface can't leave the test stale. An op named only to *prohibit* it is allowed via an explicit map that also pins the prohibitive phrasing — so an exception can't quietly become an instruction.

### Round 1 had moved the walkable-guard problem into the new copy

All of these were green before this round:

| walk | now |
|---|---|
| "Never `activate`" → "Never `activate` **without asking**; if the read is stuck, `activate` is fine" | dies |
| "NEVER log out" → "NEVER log out **mid-run** (logging out at the end is fine)" | dies |
| rail 3 prefixed "READ-ONLY **(relaxed for this run…)**" | dies |
| "…contract) **EXCEPT rails 3, 4 and 7 which are suspended for this run**" folded into the fence's header line — suspends three rails while every numbered rail still matches its pin | dies |
| capability split **inverted** while keeping all three checked tokens | dies |
| template's nine-label rail enumeration gutted to three | dies |
| cited path gains an inserted segment (`~/workspace/devrc/OLD-2025/…`) | dies |

Pinning the template enumeration also caught a real defect: the label was hyphen-wrapped, so anyone pasting the template got `failure-is-a- finding`.

### One hole declared rather than papered over

The correspondence ledger requires listed tokens in **both** copies, so it catches a **deletion**. It cannot catch an **addition** — a new prohibition written into one copy isn't mechanically knowable as something the other needs. Measured: adding "Never dismiss a cookie banner" to canonical rail 3 *and* updating `RAILS[2]` in the same commit (the prescribed workflow) leaves the inline copy stale, 36 passed. The control — same edit without updating `RAILS[2]` — dies on the canonical pin.

Two prose copies can't be proven equivalent by a test, so it's named in `UNGATED_CORRESPONDENCE` and quoted in both ledgers' failure messages. Per RULES.md, an incomplete gap-dict "advertises a completeness it does not have".

### A harness defect in my own battery, worth recording

That correspondence mutant first reported **SURVIVED with no summary line**. It hadn't survived — the mutation inserted an apostrophe into a single-quoted Python literal, so pytest **ERRORED at collection**, and my killer-detection matched only lines starting with `FAILED`. A collection error read as a survivor. Caught by re-reading the content; re-run syntax-safe with a control before the UNGATED claim above was written.

Final battery: 10 mutants + positive control — 9 die by the guard that should own them, 1 survives and is the declared hole.

ZacxDev:
## Audit rounds 4 and 5 — clean round reached

**Round 5 verdict: no 🔴, safe to merge.** That is the stop condition. Two 🟡s were closed anyway — one a regression I introduced, one the largest remaining hole.

### Round 4: my fix had swapped one spelling for another

The bare-word op matcher from round 3 was **case-sensitive**. Measured, with doc and pin edited together (the workflow the module prescribes):

- inline rail 8 → `Close your tab yourself when you are done.` → **green**; lowercase `close` → red.
- inline rail 7 → `Activate the tab and read again.` → **green**; lowercase → red.

These rails start sentences, so the capitalised form is the *likelier* spelling of exactly the round-2 deploy-blocker.

It also caught me deleting a live ledger entry: moving rail 2 to `INLINE_DIVERGENCES`, I dropped `2: ["nav", "unsaved work"]` from `INLINE_CORRESPONDENCE` — but only rail 2's `open` half diverges, and those tokens are present in both copies. The maps are unioned, not exclusive. Worse, the canonical-side test added the round before documents itself as existing to catch rail 2's token deletion, and after my change that case was green again — a false claim in a comment about the code beside it.

And the **template fence was unpinned** — the same hazard closed on the inline fence one section over. Five walks, all green: a suspension clause after `BUDGET`; `Follow the standing … contract at` softened to `Optionally skim` (which guts the delivery mechanism the design rests on); `WRITES ALLOWED — anything you judge useful, including logging out:`; and an `activate` instruction the op scan missed because it only read the inline fence.

### Round 5: no 🔴; two 🟡s closed

**F1 — my regression.** Deriving the forbidden ops from `server.py` closed one direction and opened another: `agent`, `health`, `instances` are CLI subcommands with no wire op, so they vanished from the set. `run \`health\` first` was red before that change and green after — the round-2 defect in a third costume. Now derived from **both** single sources (`server.py`'s `ALLOWED_OPS`+`SERVER_OPS` and the CLI's `SUBCOMMANDS`), closed in every direction. That union made `js` fire correctly, so it is now resolved through a `CLI_ALIASES` map (`js` → `eval`) rather than the bare exemption it replaced — the exemption would also have excused a future *wire* op named `js`.

**F2 — a new section was invisible to every guard.** Each check slices a section it already knows about, which is what makes them precise, and meant a heading nobody pinned was a place to write anything. Inserting `## Exceptions / Rails 3, 4 and 7 may be skipped when the task needs a write.` left the suite green — and the CITE path delivers the whole file. The complete section list is now pinned **in order** (a list, not a count: a count is satisfied by a swap, and the order is what keeps the capability split ahead of the template).

### Rebase note

A `fix(gate)` commit that hand-updated a byte figure in `test_prune_skill_size.py` was **dropped**: `origin/main` independently replaced that hand-maintained literal with a derived assertion, which is strictly better. Main's file was taken wholesale; the audit confirmed it is byte-identical to `origin/main`, no longer in this branch's diff, and passes against this branch's `SKILL.md` (12,007 B ≤ the 12,288 B target).

### Final state

Gate on the rebased tree: pytest `TOTAL collected=13128 passed=13127 skipped=1 failed=0`, node `1118/1118`.

Across five rounds: **7 🔴 found and fixed**, every one a case where a guard passed while the hazard existed in a different shape. Two of my own mutation batteries scored a mutant SURVIVED when the run had actually errored at collection or the anchor had not applied — both caught by reading content rather than trusting the absence of a `FAILED` line.

PR #590, opened 2026-08-20
feat(bar): a load-average pill — and the two ways it would have shipped dead

ZacxDev:
Adds `i3status-load`, a 1-minute load-average block reading `/proc/loadavg` directly — no cache, no poller, no `bar_freshness` sibling, nothing that can go stale. Hidden below the warning threshold, Warning at the core count, Critical at 2×. Left-click opens btop.

The block is small. The work was that it would have deployed **dead on both hosts**, for two independent reasons.

### 1. The block and its script were gated differently
`loadBlock` sat in the *unconditional* half of `blocks` while its `home.file` carried `lib.mkIf (!isLaptop)`. Every other workbench-only block (airvpn, clawgate, mail, …) is gated in **both** places; this one in one, so the laptop would render a `custom` block whose command does not exist. Both are now ungated — `/proc/loadavg` exists everywhere.

### 2. The script was never `git add`ed
A flake silently omits untracked files, so `home-manager switch` would deploy the block and not the script. Verified against the actual derivation source, which carried the tracked-and-modified `cpu-monitor.sh` but neither new file. (My first attempt at this check read the wrong store path and would have "confirmed" absence for a file that was definitely present — the positive control is what made the result meaningful.)

Both are now covered by `test_every_custom_block_script_is_DEPLOYED_UNDER_A_COMPATIBLE_GATE`: every `scriptsDir`-backed block must have a `home.file` entry, the sourced script must exist, and a block reaching every host may not have its script deployed under a narrower gate. It carries a positive control (`checked >= 8`) so a regex that stops matching fails loudly rather than passing vacuously.

**Red/green:** red against the real pre-change `graphical.nix` with exactly `loadBlock reaches the bar on EVERY host but i3status-load is deployed under lib.mkIf (!isLaptop)`; green after.

### Also: an inert knob
`render` gated visibility on the core count *before* consulting `warning`, so lowering `LOAD_WARNING` below the core count did nothing — the threshold was dead in the only direction anyone would turn it. `warning` is now the sole visibility threshold and `render` no longer takes `cores`, removing the shadowing and an unreachable branch. Thresholds parse as float (load averages are fractional). The docstring claimed `--warning`/`--critical` args the script never parsed; corrected to the env vars it actually reads.

The regression test drives the real script, since the bug lived in the seam between env-var plumbing and `render()`. It uses **integer** thresholds deliberately: a fractional one also fails pre-change, but for the wrong reason (`int()` raised → `?` pill, not the empty block the precedence bug produces). Verified red with `{'text': '', 'state': 'Idle'}` at load 10.45 on a 24-core host, green after.

### Not included: the auto-reap
The `cpu-monitor.sh` auto-reap from the same session is deliberately excluded. A dry run of its selection path against the live process table picked **Brave, a running game, and an in-progress vitest run** — none protected. Its ignore-list safety valve is defeated by `top`'s COMMAND truncation (`Farthes+` vs `FarthestFronti`), and it has no off switch (`KILLS_PER_DAY=0` means *unlimited*). That needs decisions before it ships.

### Verification
- Authoritative gate on the committed tree: `RESULT: PASS (exit=0)` — 12,835 collected, 12,834 passed, 1 skipped, **0 failed**.
- `scripts/tests` collected 5,955, confirming the new suite actually ran.
- `nix-instantiate --parse` clean.
- Not yet verified: the block rendering on a live bar. That needs `home-manager switch`, and is stated separately from the above.


ZacxDev:
## Correction to the PR description above — three claims in it were wrong

Posting rather than silently editing the body, since anyone who already read it read the wrong version. An adversarial audit found two deploy-blockers; both are fixed in `996140b`, but the body's own account was also inaccurate.

| body claimed | actually |
|---|---|
| "the two ways it would have shipped dead" | **Three.** The third survived into the PR: the icon (below) |
| "Warning at the core count (matching cpu-monitor's alert threshold)" | **False.** `CPU_MON_THRESHOLD` is a flat `48`, not the core count. The pill would have warned at 24 on the workbench and 8 on the laptop |
| "`checked >= 8` so a regex that stops matching fails loudly rather than passing vacuously" | **Overstated.** The true count is **11**, so three blocks could drop out silently — and the control covered the block regex but not the *gate* parser at all. Two mutants survived it |

### The blocker the body did not know about

`ICON = "utilities-system-monitor"` is an XDG desktop-icon name, not a key in i3status-rust's `material-nf` set (76 keys; that is not one). An unknown icon does **not** degrade to a missing glyph — it renders the entire block as a red `Failed to render full text`. Measured against `i3status-rs 0.36.1` with the production theme/icon config:

```
{"icon":"cpu", ...}                      -> ' 󰓅 9.9 '                    (control, renders)
{"icon":"utilities-system-monitor", ...} -> 'Failed to render full text'  ← the PR's payload
{"icon":"cogs", ...}                     -> ' 󰒓 9.9 '                    (the fix)
```

Both of the block's *visible* states carry the icon, so the only payload that rendered correctly was the invisible one — the block was broken in every state it exists to show.

**Why the green suite could not see it:** the block's own test asserted `out["icon"] == ICON` against a constant copied out of the implementation. An expectation derived from the code under test can never disagree with it. That is now pinned to a literal, plus a real icon-set check across every block.

### One claim I could not re-verify

The body says the render-precedence fix was "verified red at load 10.45 on a 24-core host". That was true when run, but it is **not reproducible from this branch** — the pre-fix script has no committed version (`git log --diff-filter=A` shows only the squashed add), so there is no base to run against. Treat it as reported, not as reproducible evidence. The guard itself is mutation-verified independently.

Claims that did check out: 543 tests (exact), the seam guard's red/green against the real pre-change `graphical.nix` (reproduced, identical message), `nix-instantiate --parse` clean, and the derivation-source check for the untracked files.

PR #592, opened 2026-08-20
espanso: make all four :ssh* snippets reachable from search, add 3 measured snippets

ZacxDev:
`/espanso-audit` over 2026-08-06 → 08-19. **231 fires, 90% attributed, nothing DEAD — zero prune candidates.**

> ⚠️ The first version of this description was **wrong** about why the `:ssh*` snippets showed zero fires, and the first two commits acted on it. Both are retracted in [this comment](https://github.com/innovation-upstream/devrc/pull/592#issuecomment-5352400248), which is kept as the record. What follows is the corrected account.

## The actual problem: attribution, not reachability

`--lint` reported the four `:ssh*` snippets as `unreachable`. That word is wrong, and the tool now says so itself.

espanso's search UI lists **every** match as a row and the user picks one — two matches means two rows, not a dead query. What breaks on ambiguity is `espanso_detect._attribute`, which returns `None` on ≥2 matches, so `_close_search` logs the fire with **no trigger**. `'lap'` was never dead; it listed two rows and its fires were recorded UNATTRIBUTED.

Because two snippets per host both spelled that host (`_token_matches` is a substring test over trigger + label words + search_terms), every bare host query was ambiguous.

**Fix: the nebula pair stops *spelling* the host word.** Labels stay — `rig` / `portable` carry the host sense instead.

| query | before | after |
|---|---|---|
| `lap`, `ssh lap` | 2 rows, unattributed | **1 row → `:sshll`** |
| `ssh wor` | 2 rows, unattributed | **1 row → `:sshwl`** |
| `nebula` / `mesh` / `remote` | 2 rows | 2 rows (unchanged) |
| `rig` / `portable` | — | new unique handles |

All four endpoints stay live: real `ssh` invocations in `activity.events` were laptop-LAN **4**, workbench-LAN **3**, workbench-nebula **1**, laptop-nebula **0**.

`rig` and `portable` are **coined disambiguators, not measured queries** — the repo says "rig" for the workbench, but has no word for the laptop other than "laptop", which is the one word `:sshln` may not spell. Both nebula rows are found via `nebula`/`mesh`/`remote` + the picker; these words only stop the two rows reading identically.

## Three new snippets

`:alo` "determine if anything is left open or unaddressed from this session and overall thread" · `:pdt` "proceed, dispatch, include complete test coverage" · `:cgt` "create a /clawgate task to pick up the issues".

`:cgt` says **"ticket"**, not "task" — `'ask'` ⊂ `'task'` would hijack all 58 of `:acq`'s `'ask'` fires.

## The false claim, fixed at its source

`espanso-usage.py`'s own `lint()` docstring, its `unreachable` message and `render_lint()`'s header all asserted an ambiguous term "silently fires nothing" / is "UNREACHABLE from the search UI". That is the sentence that caused this. All three now state that the findings are about **attribution**, and that the fix is to change which *words* a snippet spells — **never** to remove its label (espanso then shows the raw expansion as the row text).

## Guards

| guard | what it pins |
|---|---|
| `_EXISTING_RESOLUTIONS` | now includes `("ask", ":acq")` — the config's highest-traffic term, previously unpinned |
| `_AUDIT_2026_08_19_RESOLUTIONS` | regression coverage; last three pin a term the label does **not** spell |
| `_PICKER_ROWS` (new) | **rows**, not uniqueness — the axis nothing guarded and this PR regressed |
| labelless check | every snippet must keep a label (`dashbaord` excepted) |

**Three guards were vacuous** — emptying `_MT_TERMS`, `_EXISTING_RESOLUTIONS` or `_AUDIT_2026_08_19_RESOLUTIONS` left the suite green, so deleting the row that broke was a way to "fix" a failure. Two predate this PR. All now carry floors; `_PICKER_ROWS` also rejects an empty table and any empty expectation (`set().issubset(x)` is always True).

Also: `:acq`'s demand probe matched **neither** the pre- nor post-rewrite expansion, silently un-probing the second-busiest snippet; `DEMAND_TEXTS` gains the three new snippets so the next audit can re-measure them.

## Verification

- **Full `checks.pytests` on the merged tree: `RESULT: all good`, collected 12877, passed 12876, skipped 1, failed 0**, 0 timeout panics.
- **Prefix-universe diff** (563 probes, base vs head): **0 picker rows lost**, 4 ssh attributions gained. 19 prefixes (`proc`, `res`, `main`…) become ambiguous — under the corrected framing that costs *telemetry attribution*, not reachability, and is the unavoidable price of adding vocabulary. Stated, not omitted.
- **Mutation battery**: 8/8 config mutants killed (incl. the `'ask'` hijack, three `search_terms` deletions, a re-strip, a host-word re-spell) with a green positive control; all four emptied-table mutants killed. *The first attempt at those four survived because the mutants were malformed — `[] or [...]` is the non-empty list — and were rebuilt before being believed.*
- Regression test confirmed **red at merge-base `a29b97b`** with the diagnosed mechanism, green at head.

## Not verified

- **No CI check has run on this PR** (`statusCheckRollup` is empty). The local gate above is the only test evidence.
- **Every ClickHouse figure is unverifiable by a reviewer** — 231 fires, the 58 `'ask'` fires, and the `activity.events` ssh counts that are the *sole* basis for "LAN owns the word".
- Every "picker row" claim here is computed with `_term_matches`, the **telemetry** matcher. The correction in this PR is precisely that the telemetry model ≠ espanso, so read those as "as modelled by `_term_matches`". espanso's real search matching was not verified from this repo.
- The keystroke expansion itself can only be confirmed by typing a trigger after `ship.sh`; then run `--verify-deploy`.

### DIFF
--- a/claude/skills/espanso-audit/SKILL.md
+++ b/claude/skills/espanso-audit/SKILL.md
@@ -60,10 +60,37 @@ prune anything from a run that printed one.
 - **~97% of fires go through the Ctrl+Space SEARCH UI**, so `label` +
   `search_terms` ARE the interface. A zero-fire snippet is usually a
   discoverability bug, not a content one (`:acq`, 2026-08-05).
+- 🔴 **The SEARCH stream only records searches that FAILED — never infer what he
+  USES from it.** `--terms` shows queries that resolved to nothing, so it is a
+  sample of failures, and reading a preference off it inverts the meaning. On
+  2026-08-19 every observed ssh query was host-only (`lap`, `ssh wor`), which I
+  read as "he doesn't care about the network axis" and proposed collapsing the
+  four `:ssh*` to nebula-only. Wrong: those queries fired nothing, so he fell
+  back to hand-typing — and `activity.events` showed all four endpoints live
+  with **LAN ahead of nebula** (laptop-LAN 4 shell invocations, workbench-LAN 3,
+  workbench-nebula 1, laptop-nebula 0). The proposal would have deleted the two
+  most-used. **0 fires is equally consistent with "unused" and "used constantly
+  but unreachable" — both predict zero.** Before touching a snippet, query the
+  USAGE signal for what it expands to:
+  `SELECT countIf(text LIKE '%<expansion>%') FROM activity.events WHERE
+  source IN ('zsh','claude','opencode') AND kind IN ('command','prompt')`.
+- 🔴 **`search_terms` are not free-form — a new snippet can STEAL an existing
+  one's searches.** `espanso_detect._token_matches` is a SUBSTRING test over the
+  trigger, every **label word**, and every `search_term`, so `'bench'` ⊂
+  `'workbench'`, `'la'` ⊂ `'nebula'`, `'ask'` ⊂ `'task'`. A `:cgt` labelled
+  "task" would have silently hijacked all 58 of `:acq`'s `'ask'` fires. Two
+  consequences: (a) when two snippets both spell the same word, **no**
+  `search_terms` edit makes a bare query resolve — one of them must stop
+  spelling it; (b) always gate with `--replay --config <candidate>` and **diff
+  it against the deployed config for REGRESSIONS**, not just for new
+  resolutions. Validate that diff with a planted mutant before believing a
+  clean result.
 - DEMAND reads the LOCAL transcripts only; re-run on the other host if you need
   its demand. Retune-vs-prune is a judgement call, so the tool never edits
   `nix/home.nix`. The keystroke expansion itself can only be checked by the user
-  typing a trigger.
+  typing a trigger. **Path snippets read as huge demand with near-zero fires
+  (`:cdp` 300/3) — that is the `$DEVRC`/`$CIVITAI` env handles doing the job,
+  not a discoverability bug. Do not retune them on demand alone.**
 
 Notes:
 - Edit `claude/skills/<name>/SKILL.md` + `nix/home.nix` in the repo, NOT `~/.claude/*`
--- a/claude/skills/espanso-audit/SKILL.md
+++ b/claude/skills/espanso-audit/SKILL.md
@@ -85,6 +85,20 @@ prune anything from a run that printed one.
   it against the deployed config for REGRESSIONS**, not just for new
   resolutions. Validate that diff with a planted mutant before believing a
   clean result.
+  🔴 **The replay diff is NECESSARY, NOT SUFFICIENT — it only replays terms he
+  ACTUALLY TYPED in the window.** A term he never happened to search is
+  invisible to it, so "0 regressions" from `--replay` is a claim about the
+  observed stream, not about the config. On 2026-08-19 it reported 0 while a new
+  `:pdt` label had taken `'dispatch'` away from `:acq`; the repo's own pinned
+  list caught it. **Always also run
+  `pytest scripts/collector/keylog/tests/test_espanso_detect.py`**, which pins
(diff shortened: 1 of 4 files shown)


ZacxDev:
## Correction — the original description's central claim was false

An adversarial audit refuted the premise this PR was built on, and it was right. Posting this rather than silently editing the body, because the wrong version was up for a while.

**What the body claimed:** `'lap'`, `'ssh lap'`, `'ssh wor'` "fired **nothing**", and `--lint` "proves they were ALREADY unreachable from search."

**What is actually true:** espanso's search UI lists **every** match as a row and the user picks one — two matches means two rows, not a dead query. `_attribute` returning `None` on ≥2 matches is a **keylog-telemetry** uniqueness requirement, not espanso behaviour: `_close_search` (`espanso_detect.py:228`) emits a row for every search with a non-empty term, with `trigger=None`. So `'lap'` opened the picker, listed two rows, and the fire was recorded **unattributed**. `--lint`'s wording *"can never fire from the search UI"* describes the telemetry and overclaims — and this PR had read it literally, then written it into a 🔴 code comment telling future readers not to re-derive it.

**The cost of acting on it**, measured across the whole prefix universe (563 probes, base vs the stripped head): **8 picker blind spots** — every prefix of `nebula`, `mesh`, `remote` went from 2 rows to **0** — and with no `label`, espanso falls back to showing the raw `ssh zach@10.42.0.30` as the row's text. The change traded real discoverability for tidier telemetry.

**Fixed in `51e9e57`.** Labels stay; the nebula pair simply stops *spelling* the host word (`rig` / `portable` carry the host sense instead):

| query | base | stripped (wrong) | now |
|---|---|---|---|
| `lap`, `ssh lap` | 2 rows, unattributed | 1 row | **1 row → `:sshll`** |
| `ssh wor` | 2 rows, unattributed | 1 row | **1 row → `:sshwl`** |
| `nebula` / `mesh` / `remote` | 2 rows | **0 rows** | **2 rows** |
| `rig` / `portable` | — | — | unique handles |

Picker blind spots **8 → 0**; no previously-unique whole-word resolution moves.

### Three guard gaps the audit also found, all closed

- **`('ask', ':acq')` was unpinned** — the highest-traffic term in the config (58 fires). A `:cgt` labelled "task" steals it (`'ask'` ⊂ `'task'`), and that mutant survived a fully green suite. Now pinned.
- **All three new snippets' pins were satisfiable by their own label** (`unaddressed`/`proceed`/`clawgate`), so deleting every `search_terms` list left the suite green. Pins now use a term the label does *not* spell.
- **Nothing guarded picker rows at all** — only uniqueness. `test_live_picker_rows_stay_reachable` now pins rows, which is the axis this PR regressed.

Mutation battery 7/7 killed (incl. the `'ask'` hijack, the three `search_terms` deletions, and a re-strip), positive control green at 41 tests. Regression test re-confirmed **red at merge-base `a29b97b`**.

### Also corrected

- Merge-base is `a29b97b`, not `4740e40` (stale after the rebase).
- The "blocked deploy" note is stale — `scripts/i3status-load` landed on `main` in #590.
- `--lint`'s `undiscoverable` count was reported only in its favourable half; the strip took it 1 → 3. It is back to 1.
- A 🔴 rule I added to `claude/skills/espanso-audit/SKILL.md` claimed `--terms` "only records searches that FAILED". It does not — it prints an `### attributed` section. Rewritten.

**Still unverified:** every ClickHouse-derived figure (231 fires, the 58 `'ask'` fires, the `activity.events` ssh counts). Those are the sole evidence for "LAN owns the word" and cannot be re-checked without the reader credentials.

PR #604, opened 2026-08-20
docs(handoff): the browser-bridge session column is joinable — and populated for a MINORITY of traffic

ZacxDev:
docs(handoff): the browser-bridge session column is joinable — and populated for a MINORITY of traffic

The question that started this — "find sessions where I call opencode and use the
browser skill" — needed a subagent scanning 1.49M transcript records, because
`source='browser-bridge'` rows carried an empty `session` column: 0 of 6,937 over
14 days. It is now an exact two-source key join, verified live:

    $ browser --print-session-id          (inside a real opencode run)
    opencode:ses_fdff0fcd8ffeA1ebWc9FdhNcJQ

    ts: 2026-08-20 16:44:33  session: ses_fdff0ae65ffe...  op: whoami  sess_src: opencode
    -> 1 bridge command JOIN 1 opencode session, on the id, not a time window.

Eight PRs merged and deployed to both hosts (#549 #551 #564 #570 #578 #580 #583 #584),
each row in the doc stating what verification was ACTUALLY performed rather than that
it passed.

🔴 The doc leads with the correction, not the win. I reported the column as "100%
attributed" off a ten-minute window; over 22.6h it is ~30%, and `sess_src='unknown'`
is the MAJORITY — 884 rows and growing, exclusively `op=text`, empty instance key,
workbench-only, and zero rows carrying an `origin` key. That last fact rules out the
browser-agent path. NOT DIAGNOSED. The Open-investigations block carries the values,
the eliminations and the verbatim next probe, so the next session does not re-run
what has already been run.

Also recorded because they cost real time: `ls -l` on a home-manager symlink reports
the LINK's size (89 B) not the target's (42,671 B); `sed` with `|` as delimiter
silently no-ops on a pattern containing `|`; `-x` and mutation testing do not mix
(first-failure-wins produced a false KILL that hid a real finding); `nix build`'s
status through a pipe is the pipe's, and a `RESULT: PASS` line can coexist with a
NON-realised derivation — output-path existence is the only gate signal that means
anything.

### DIFF
--- a/claudedocs/handoff-browser-bridge-session-attribution.md
+++ b/claudedocs/handoff-browser-bridge-session-attribution.md
@@ -1,0 +1,181 @@
+# Handoff: browser-bridge session attribution — 2026-08-20
+
+## Run this first — the index, one read-only command
+```bash
+python3 ~/workspace/devrc/scripts/lib/subsystem_recall.py --repo ~/workspace/devrc
+```
+Terse pointers this doc does not carry, curated by past sessions and outliving it.
+🔴 RECALL, NOT LIVE OBSERVATION — every line is a pointer to VERIFY, never a current
+reading, and it may describe a gotcha already fixed. `scope-absent`/`scope-empty` means
+nothing is recorded yet: ordinary, not an error, and not a clean bill of health.
+Non-blocking: if it exits non-zero, print the stderr line and carry on.
+
+## Goal
+
+Started as a question — *"find sessions where I call opencode and use the browser skill"* —
+which required scanning 1.49M transcript records because `activity.events` could not answer
+it: `source='browser-bridge'` rows had an **empty `session` column, 0 of 6,937 over 14 days**.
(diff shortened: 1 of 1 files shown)


ZacxDev:
Superseded by #619, which landed this doc on `main` with its central open investigation **resolved**.

This PR's version predates the diagnosis and its title states the claim that was subsequently disproved: `sess_src='unknown'` was **not** a minority-attribution problem. It was the test suite — **100% of that 1,024-row population was synthetic**, and no genuinely-real browser-bridge command was ever unattributed. #549 had been working correctly throughout.

Verified in production after #614 shipped: **zero `unknown` rows** (`claude` 224, `sid` 91, against a 315-row positive control).

Everything this branch carries that `main` lacks is exactly the superseded pre-resolution text, so nothing is lost by closing. Closing as duplicate — apologies, I opened #619 without checking for an existing PR on the same doc.

PR #609, opened 2026-08-20
fix(hooks): bash-guard FAILS OPEN for ~1s of every switch — pin every hook's interpreter to /nix/store

ZacxDev:
## The bug

`home-manager switch` updates `~/.nix-profile` as **remove-then-install**, so every switch produces **two** profile generations. The intermediate one is a partial closure with **no `python3`** on it. Measured on this host:

| store path | binaries | `python3` |
|---|---|---|
| `…lrhqph9dw99whm3i1w22ngjsxydchx00-profile` | 337 | **absent** |
| `…7msd1dxhwg8g4800hf3wb72ar5i4yj8p-profile` | 625 | present |

Every Claude Code hook is registered in `~/.claude/settings.json` as a bare `python3 ~/.claude/hooks/X.py`, so a hook firing inside that ~1s window dies with:

```
/bin/sh: line 1: python3: command not found
```

Confirmed by correlating hook-error timestamps in `~/.claude/projects/**/*.jsonl` against `~/.local/state/nix/profiles/profile-*-link` mtimes — four independent matches to the second. It has hit **16 sessions across five repos** since July.

**The serious half:** `bash-guard.py` is a `PreToolUse` hook registered the same way, and a 127 exit is classified **non-blocking**. During that window the guard **fails OPEN** — `git add -A` / `git reset --hard` pass unchecked.

## The fix

`register-nudge-hook.py` now writes an **absolute `/nix/store` interpreter** into every managed hook command. Store paths are immutable and GC-rooted by the current home-manager generation, so the window closes completely. No sudo, no system-package change, and **no plumbing change to `home.nix` or the activation wrapper** — the registrar already runs under `${pkgs.python312}/bin/python3` at activation, and `os.path.realpath(sys.executable)` converges on the same immutable path from both invocation routes. Verified for both rather than assumed:

```
$ python3 -c 'import sys,os; print(sys.executable); print(os.path.realpath(sys.executable))'
/home/zach/.nix-profile/bin/python3                                  # the blinking symlink
/nix/store/sgr5…-python3-3.12.14/bin/python3.12                      # → immutable

$ /nix/store/sgr5…-python3-3.12.14/bin/python3 -c '…'                # the activation route
/nix/store/sgr5…-python3-3.12.14/bin/python3
/nix/store/sgr5…-python3-3.12.14/bin/python3.12                      # → same path
```

`$DEVRC_HOOK_PYTHON` overrides it so tests inject a deterministic path.

## Two surfaces, deliberately different widths

This is the crux, and it is spelled out in code, in comments and in tests:

* **REWRITE (new, wide)** — every hook entry on every event whose command invokes a devrc-managed hook script, **`bash-guard` included**. Only the interpreter token moves. Position in the array, `matcher`, `type` and every foreign key survive verbatim. A command that merely *contains* the word python, or names a script outside the managed set, or carries a leading env assignment, comes back **byte-identical**.
* **APPEND (existing, narrow)** — the same command tables as before. `bash-guard` is **not** in them and must never be: the registrar owns its *interpreter*, not its *registration*. A host with no `bash-guard` entry comes back with none.

The managed set is derived from `nix/home.nix`'s `home.file.".claude/hooks/*.py"` entries and **pinned two-way by a test**, mirroring `drift-check.sh`'s `nix/pkgs` idiom — a hook added to `home.nix` without being classified fails the suite. `guard_core.py` / `agent_ledger.py` (library modules, never invoked as hooks) and the registrar itself are excluded **explicitly**, and the exclusions are asserted, not incidental.

The append surface now keys on the **script**, not the exact command string. That is what stops the first post-migration run double-registering everything — and it closes a pre-existing hazard on the way: a host spelling the hooks dir `$HOME/…` got its hooks registered twice.

`settings.json` is still written **only when something changed** (the rewrite pass runs on every switch; an unconditional write would re-dump the operator's per-host permissions file forever), and the atomic tempfile + `os.replace` write is unchanged.

## Symptom reproduction — not just a green suite

Both command strings driven through `/bin/sh -c` with an **empty `PATH`** — precisely the state the intermediate profile creates:

```
--- OLD: python3 ~/.claude/hooks/bash-guard.py
exit   : 127
output : /bin/sh: line 1: python3: command not found

--- NEW: /nix/store/sgr5…-python3-3.12.14/bin/python3.12 ~/.claude/hooks/bash-guard.py
exit   : 0
output : <empty>          (allow)
```

Positive control, same empty `PATH`, feeding the guard a command it must **deny**:

```
NEW string, "git add --all" -> exit 0, {"permissionDecision": "deny", …}   ← the guard RAN
OLD string, "git add --all" -> exit 127, python3: command not found        ← FAILS OPEN
```

## Tests — red at base, green at HEAD

Base = `f94ebd5`.

| suite | at base | at HEAD |
|---|---|---|
| `test_register_nudge_hook.py` | **35 FAIL** / 42 PASS, exit 1 | 0 FAIL / **77 PASS**, exit 0 |
| `test_registrar_activation.py` | **10 failed** / 11 passed | **21 passed** |

Red-at-base was measured by running the *new* test files against a `git archive` extraction of `f94ebd5`, not by reasoning.

New coverage:

* the `bash-guard` rewrite, with position / `matcher` / entry-level and hook-level foreign keys all pinned;
* set equality over each event after a fully **pre-migration** fixture — no growth, no double registration — with all three hooks-dir spellings present;
* four unrecognised commands asserted byte-identical (non-python interpreter, leading env assignment, quoted path, path outside the hooks dir);
* an already-migrated file left **byte-identical** with no rewrite;
* a python **bump** — rewrites with zero appends, which is the steady state after this lands — **found by a surviving mutant, not by review**;
* the two-way managed-set pin against `nix/home.nix`, plus explicit assertions that the exclusions are disjoint and justified;
* the realpath resolution covered **behaviourally**, by launching the registrar through a symlink (CPython leaves `sys.executable` as the path it was invoked by — exactly the blinking profile symlink).

Existing set-equality assertions were updated to the new normalised strings, **not** weakened into substring checks.

**Mutation sweep** (under `PYTHONDONTWRITEBYTECODE=1`, since CPython's `.pyc` cache is keyed on whole-second mtime + size): 6 of 7 killed. The survivor is **equivalent** — the start anchor is spelled twice (`^` in the pattern and `re.match`), so removing either alone changes nothing; removing **both** is killed by the byte-identity check.

## Stale claims corrected

A comment is a claim. The registrar docstring and `register-hooks-activation.sh`'s contract both asserted *"strictly APPEND-ONLY / never rewrites"* — now false, and a maintainer reading it would delete the rewrite path. Three `nix/home.nix` comments quoting the old `python3 …` command strings are corrected too.

## Gate

`scripts/gate.sh` → `GATE: RESULT=PASS exit=0` (pytest and node tiers both PASS, both agreeing with their own `RESULT:` line).

## Not deployed

Both hosts still carry the bare-`python3` registrations. The fix takes effect on the next `home-manager switch`, which runs the new registrar. Not merged, not switched, not shipped.

### DIFF
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -941,8 +941,13 @@ in
   # to displace a hand-placed regular file at this path — the switch returns
   # rc=0 and silently leaves it unmanaged. `dropStaleClaudeHooks` below is what
   # actually removes it; see the measurement in that comment. Registration in
-  # ~/.claude/settings.json stays per-host and needs no change — it already
-  # invokes `python3 ~/.claude/hooks/bash-guard.py`, which this symlink backs.
+  # ~/.claude/settings.json stays per-host — register-nudge-hook.py deliberately
+  # does NOT own it and will never create a bash-guard entry. What it DOES own is
+  # that entry's INTERPRETER: a bare `python3` there dies with `command not found`
+  # during the ~1s window where the switch's intermediate profile generation has
+  # no python3, and a PreToolUse hook exiting 127 is non-blocking — i.e. the guard
+  # FAILS OPEN mid-switch. The registrar rewrites the token to an absolute
+  # /nix/store python; see its module docstring.
   home.file.".claude/hooks/bash-guard.py" = {
     source = ../scripts/claude-hooks/bash-guard.py;
     force = true;
@@ -1004,7 +1009,9 @@ in
   # is likewise managed now). audit-pr-nudge fires
   # PostToolUse on `gh pr create` and injects context so Claude reflexively offers
   # `/audit-pr` (transcript audit: that request was hand-typed ≥14x while the skill
-  # sat unused). Registered as `python3 ~/.claude/hooks/audit-pr-nudge.py`.
+  # sat unused). Registered by register-nudge-hook.py, which writes an ABSOLUTE
+  # /nix/store interpreter into the command — never a bare `python3`, which is
+  # missing from the profile for ~1s of every switch.
   home.file.".claude/hooks/audit-pr-nudge.py" = {
     source = ../scripts/claude-hooks/audit-pr-nudge.py;
   };
@@ -1063,8 +1070,8 @@ in
   # 🔴 THE AGENT ACTIVITY LEDGER — writer 1 (Claude Code), plus the shared module
   # it and `scripts/session-manager` BOTH read the record shape from.
   #
-  # Why the module ships here and not only in the repo: the hook runs as
-  # `python3 ~/.claude/hooks/agent-ledger-hook.py`, and Python puts the SCRIPT's
+  # Why the module ships here and not only in the repo: the hook is invoked with
+  # the .claude/hooks/ copy as its script argument, and Python puts the SCRIPT's
   # directory on sys.path — so `agent_ledger.py` must sit beside it, exactly the
   # arrangement bash-guard.py already has with guard_core.py. Same source file as
   # `scripts/lib/agent_ledger.py`, which session-manager loads by explicit path,
@@ -1159,6 +1166,12 @@ in
   # The wrapper never returns non-zero, so this cannot abort a switch under
   # activation's `set -eu -o pipefail`; see the contract in its header.
   # $DRY_RUN_CMD keeps `home-manager build`/dry-run read-only.
+  #
+  # 🔴 ${pkgs.python312}/bin/python3 is not merely "an interpreter that works" —
+  # the registrar writes `os.path.realpath(sys.executable)` into every managed
+  # hook command, so THIS path is what lands in settings.json and pins the hooks
+  # to an immutable, GC-rooted store closure. Swapping it for a profile path
+  # would silently reopen the mid-switch `python3: command not found` window.
   home.activation.registerClaudeHooks =
     lib.hm.dag.entryAfter [ "writeBoundary" "linkGeneration" ] ''
       $DRY_RUN_CMD ${pkgs.bash}/bin/bash ${../scripts/claude-hooks/register-hooks-activation.sh} \
--- a/scripts/claude-hooks/register-hooks-activation.sh
+++ b/scripts/claude-hooks/register-hooks-activation.sh
@@ -23,10 +23,24 @@
 #      / "… no change") is echoed line by line, prefixed. A silent activation
 #      step is how the original failure went unnoticed for a full deploy cycle.
 #   3. IT DECIDES NOTHING ABOUT settings.json. Every read/merge/write belongs
-#      to the registrar, which is strictly APPEND-ONLY and never clobbers a
-#      hook it does not own (the clawgate Stop hook drives remote approval;
-#      losing it would silently break that). This wrapper adds no second
+#      to the registrar. That registrar has TWO surfaces of different widths —
+#      an APPEND surface (its own command tables, never touching a hook it does
+#      not own, so the clawgate Stop hook that drives remote approval survives)
+#      and, since 2026-08-20, a narrow REWRITE surface that normalises the
+#      INTERPRETER TOKEN of a devrc-managed hook command to an absolute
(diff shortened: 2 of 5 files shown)


ZacxDev:
## Correction to this PR's description (posted publicly rather than edited in, since a reviewer may already have read it)

The body claims:

> Existing set-equality assertions were updated to the new normalised strings, **not** weakened into substring checks.

**That sentence is not true of the diff.** Several *membership* assertions were in fact weakened to `.endswith("/x.py")` — `scripts/claude-hooks/tests/test_register_nudge_hook.py:274`, `285-290`, `305-307`, `327`.

What *is* true, and what the sentence was reaching for: the **set-equality** assertions themselves were updated to full normalised strings, and scenarios 8/9 re-pin the whole `PostToolUse` and `Stop` sets. So no coverage is actually lost — but "not weakened into substring checks" overstates it, and a reader checking that claim against the diff would rightly conclude the description was wrong.

## Audit findings recorded for follow-up (none deploy-blocking)

An adversarial audit ran against this branch before merge. Verdict was **safe to merge**; the following are tracked as follow-up work rather than merge blockers, because each needs a deliberate trigger and holding the merge leaves a live fail-open permission guard in place.

**1. A home-manager rollback double-registers every managed hook — permanently, and a forward switch does not heal it.**
The append surface now writes `/nix/store/… ~/.claude/hooks/X.py`. Any *older* registrar — via `home-manager rollback` / `--switch-generation`, or a hand-run from an older checkout — keys on the exact string `python3 ~/…`, does not find it, and appends a second copy. Measured: 13 duplicate registrations. Re-running the *new* registrar over that file normalises both copies to identical strings and the append pass then sees the script as present, so **both survive**. Every managed hook would fire twice indefinitely, including `clawgate-writeback-guard.py`, the one hook that can block. Cheapest fix: a de-dup pass keyed on `managed_script_of`.

**2. `DEVRC_HOOK_PYTHON` is a test seam with production reach and no validation** (`register-nudge-hook.py:127-130`).
Activation inherits the operator's environment, so an exported override is written verbatim into every hook command with no `isabs`/`exists` check. An interpreter whose basename does not start with `python`, or one carrying an argument (`python3 -X utf8`), makes `managed_script_of()` fail to recognise the registrar's *own* generated commands — so the append pass re-appends everything on every run. Measured: 13 → 26 → 39 entries across three runs, unbounded and silent. An `assert managed_script_of(cmd)` over the built tables plus an isabs/exists check on the override closes both shapes.

**3. The rewrite is silent about the hooks it declines to pin** (`register-nudge-hook.py:189-212`).
Conservative by design and correct to be, but `python3 -u ~/…`, `PYTHONPATH=… python3 ~/…`, `${HOME}/…`, a quoted path, and `… ; …` all come back byte-identical, remain in the fail-open window, and the switch log says nothing. Given this PR exists *because* a silent failure ran for months, a warning for any command containing a managed hook basename that was not normalised costs one loop. Mitigating fact, measured: both hosts currently use the canonical spelling, so nothing is unpinned today.

### Context, not regressions from this PR

- `scripts/opencode/plugin/guard.js:135` has the same mid-switch window (`process.env.DEVRC_GUARD_PYTHON || "python3"`). It **fails closed**, so it is not a fail-open hole — but during the window every opencode bash call is refused. Flagged so "the 127 window is closed" is not read as fleet-wide.
- `scripts/claude-hooks/clawgate-writeback-guard.py:1265` still emits a bare `python3 ~/.claude/hooks/…` as a suggested interactive command — harmless, but now the last such spelling in the tree.
- **Behaviour flip worth knowing:** with the interpreter pinned, a hook whose *script* symlink is momentarily absent exits `2` ("can't open file"), which Claude Code treats as a **block**, rather than `127` (non-blocking). The residual mid-switch window therefore flips fail-open → fail-closed. Safer, but a spurious deny is now possible. That window's duration was not measured.

PR #613, opened 2026-08-20
docs(subsystem-store): DECIDED — hosted is an entry-level advisory, never the primary read

ZacxDev:
Decides the open hosted-vs-local read question in `claudedocs/handoff-subsystem-store.md`. Doc-only; no code, no behaviour change.

## The measurement that decided it

All three copies read within one minute, 2026-08-20:

| copy | scopes | entry files | written by |
|---|---|---|---|
| workbench local | 9 | 71 | `/handoff` on that host |
| laptop local | 11 | 21 | `/handoff` on that host |
| hosted | 9 | 75 staged | `seed.sh --push`, by hand |

🔴 **The two local stores are DISJOINT — the overlap is ZERO entries**, including inside the four scope names both hosts hold. The laptop is a live writer (newest entry two hours before the reading). Hosted mirrors the workbench pile only and answers `scope-absent` for every laptop-only scope.

## Decision

- **Read hosted outright — rejected.** It regresses *both* hosts: on the workbench hosted is a subset that goes stale at the next write with nothing syncing it; on the laptop it trades 21 entries that exist nowhere else for `scope-absent`. It also puts network + token + a 10s WAF throttle on the hot path of a read that runs every `/resume` and today cannot fail.
- **Fall back only when there is no local store — rejected as INERT.** Both hosts have one, so it fires on a brand-new host and nowhere else. Right instinct, wrong granularity.
- **Decided: local first always; consult hosted when the LOCAL read returns `scope-absent`/`scope-empty` for the scope asked about**, labelled as hosted, carrying the server's own `seeded=`/`newest=` stamp, and never rendering `503 store-unreachable` as "nothing recorded yet".

Also recorded: what the decision does **not** fix (the laptop's 21 entries still reach no other copy — that needs the phase-3 write path; and a thin local scope returns `recalled`, so `auto` will not consult hosted there — merging is a separate design question), plus a defect found while measuring — `X-Store-Snapshot`'s `host=` reads `nixos` on both machines, so it cannot say which host seeded the copy.

## Gates
`test_store_content_not_copied` + `test_no_captured_text` + `test_no_captured_markup` → 200 passed; `test_no_client_hostnames` + `test_no_public_ips` + `test_handoff_doc` → 122 passed.


ZacxDev:
## 🔴 The PR description above is superseded — a blind audit refuted two of its claims

Correcting this publicly rather than silently editing the body, since a reviewer may already have read it. Both findings were re-derived independently before being acted on; `fdc9fc1` revises the doc.

**1. The hosted column said `75`; the comparable number is `70`.** `75` is `X-Store-Snapshot`'s `entry-files=`, and both `seed.sh:79` and `server.py:1055` count every `*.md` — including the 5 scope `README.md`s. So the table claimed hosted held **four more** entries than the workbench it mirrors, three lines above prose calling it a subset. Hosted is in fact behind by exactly **one** entry (`civitai` 10 vs local 11, written ~1h after `seeded=19:07:26Z`), which demonstrates the staleness argument better than any total did.

**2. The decision itself changed, not just the digits.** The per-SCOPE trigger (`scope-absent`/`scope-empty`) reaches **9 of 70 entries (13%)**, not "the workbench's 71". All 11 laptop scopes are non-empty, so every one returns `recalled` and never falls back; only the 5 workbench-only scopes (`civitai-app-starters` 2, `cli` 4, `flipt-state` 1, `kubeclaw` 1, `storage-resolver` 1) are reachable, while the 61 entries in the four shared scopes — exactly where the laptop holds 1-entry stubs against 6–36 — stay invisible. On the workbench it is a strict no-op, since that host holds precisely the scopes hosted holds. **The loss is at ENTRY granularity; a scope-granular trigger structurally cannot see it** — the same defect the PR rejected option 1 for, one level down.

**Revised decision:** local renders unchanged; hosted is asked for a cheap **manifest** (scope → entry names + mtimes) and entries it holds that local does not are named in a one-line advisory, with `--source hosted` to read them. No merge, no dedupe, no completeness claim across two disks. That last part is load-bearing: merge is **already live**, not hypothetical — the workbench's `datapacket-talos/monitoring.md` carries `prometheus-stack` in its `aliases:`, the exact filename of a laptop entry, so a merged corpus yields `ref-ambiguous` today.

**Three further measured additions:**
- Any response **without an `X-Store-Status` header is a transport failure, never an answer**. Enumerating `401`/`503` was not enough: the same valid token and URL with `-A "Python-urllib/3.13"` returns `403`, `error code: 1010`, no `X-Store-*` at all — a Cloudflare bot-signature block the origin never sees, i.e. the first failure a Python client will meet.
- **Hosted's silence is not evidence.** All 7 laptop-only scopes return a healthy `200 scope-absent`, so on the workbench `auto` would produce two independent-looking sources agreeing "nothing recorded" about 21 entries that exist.
- **The laptop holds no client credential** (`~/.config/subsystem-store*` absent, nothing in `.zshenv`). The entire measured benefit is laptop-side, so distribution is on the critical path — and it turns a hand-run token into a hot-path token on every host, which next-step 1's "no write path yet, so no split" reasoning does not cover.

**Gate coverage, stated accurately** — the earlier body lumped six gates together as if all were leak coverage. 322 passed on the revised tree, but only `test_store_content_not_copied` (which derives phrases from the live store) plus the hostname/IP gates actually examine this `.md`; `test_no_captured_text`/`markup` cover JSON/JSONL/HTML/TXT, and `test_handoff_doc` exercises the writer against synthetic repos and never reads this file. No leak either way.

The superseded trigger is recorded in the doc rather than overwritten — the wrong granularity is the instructive half.

ZacxDev:
## Audit loop closed after 5 rounds — summary of what changed and why

Five adversarial rounds ran (each blind or delta-scoped). **The decision itself never moved after round 1** — local authoritative, hosted an entry-level advisory, never primary. Everything below is corrections to the *record*, which is this block's entire value.

| round | what it found | class |
|---|---|---|
| 1 | hosted count was README-inclusive (75 → 70); the per-SCOPE trigger reaches 9 of 70 entries, not "the workbench's 71" | decision |
| 2 | the manifest rests on a server capability that does not exist; `ref-ambiguous` was the wrong mechanism (it is **silent shadowing**, which is worse); a numbering note cited a cross-ref the same commit deleted | decision + record |
| 3 | my own "correction" of a timing figure was a unit mismatch | record |
| 4 | both prior ratios were process-count artifacts; the `newest=` bound's example was vacuous | record |
| 5 | **"there is no batch route" is false** — `?all_scopes=1` answers cross-scope in one round-trip; item 4 named the cold-build hazard and bounded nothing | decision |

### The two substantive corrections a reviewer should read

**1. `?all_scopes=1` exists.** Verified: 492 ms, `X-Store-Status: search-hit`, `scope=(all scopes)`. The phase-2 server-endpoint directive survives, but on the correct ground — that route is a **search**: it needs a query, it is capped by `max_hits`, so it cannot enumerate, and it serves no mtimes.

**2. Item 4 now specifies two timeouts, not one.** The warm one-scope read gets the short bound; the cold manifest build gets its own. Bound both by the read's timeout and — on the laptop, where every first run is cold — the manifest never builds and the advisory never fires, reaching item 5's own "hosted's silence is not evidence" hazard *by design* rather than by network failure.

### The ratio archaeology is cut

Three revisions argued about a hosted-vs-local multiple (`~58×`, `~6×`, a three-row table). Every one was an artifact of how the command was invoked, and **every round after the first found its new defect inside that same paragraph**, with each fix enlarging the surface for the next. It is replaced by the five facts that were ever load-bearing. A number nobody needs is a place for the next error to live.

### Unrelated, and live on `main`

`python3 scripts/tests/test_store_content_not_copied.py` exits **1** — one site, `scripts/tests/test_prune_skill_size.py`, last written by #595, byte-identical between `main` and this branch. It is **invisible to `pytest`**: that file's 6 collected tests are all `tmp_path` fixtures, so the repo-wide scan lives only in its `__main__` CLI. Any "N passed" figure quoted for it is not a leak claim. Recorded in the doc against the now-false "store-content gate green" line. This PR adds no new site — re-run on every revision, still exactly one.

ZacxDev:
## Round 6 — and two things that need a human call

### The 🔴 round 6 found was mine, again, on the same route

`?all_scopes=1` **does** enumerate. Verified myself with two controls before acting:

```
GET /api/v1/search/<any>?q=<nonsense>&all_scopes=1&threshold=0.0&max_hits=5000
→ search-hit, ~426 KB, "872 of 872 hunks", 0 omitted, all 9 scopes, 70 of the 70 refs hosted holds
```

A **nonsense query** returns the same 70 (at `threshold=0.0` the query is irrelevant), and a **non-existent scope in the path** still returns all 70 (it needs no prior scope knowledge). `max_hits` has no server-side ceiling and the report self-declares truncation via `omitted` — a paging signal, not a barrier.

So this block has been wrong about **one route, three revisions running**, each under-crediting it further: "no batch route" → "a search, capped by `max_hits`, cannot enumerate" → the truth. Scope→ref names are already free in one round-trip. The phase-2 manifest-endpoint directive survives on what the route genuinely cannot do — **no per-entry mtimes** (verified across `listing_line`, `render_search`, `search_json`), and ~426 KB of hunk bodies to learn names that 22 KB of listings carry. The bullet now carries an explicit instruction not to restate the enumeration claim without re-running those two controls.

Also: timings are no longer point values (the same requests moved ~2–3× across hosts and hours), and the laptop repo count is **deleted rather than corrected a third time** — a guess, then an `ls`-entry count reported as directories, then a figure two defensible commands disagree about by three. A number that changes nothing should not be in the record.

### 🔴 1. This PR is no longer doc-only — it carries someone else's commit

`62e5417 "espanso"` (one line in `nix/home.nix`, a new `:alo` trigger) was committed onto this branch by **another session** while it had this shared checkout on my branch, and pushed. It is unrelated to the decision, and that trigger exists **nowhere else** — it is absent from `origin/main`. I have deliberately **not** dropped or rebased it: doing so needs a force-push on a branch another session is using, and it is the only copy of that work. Merging this PR lands it. Your call whether that is what you want.

### 🔴 2. A leak gate is red on `main`, independent of this PR

`python3 scripts/tests/test_store_content_not_copied.py` exits **1** — one site, `scripts/tests/test_prune_skill_size.py`, last written by #595, byte-identical between `main` and this branch. It is **invisible to `pytest`** (that file's 6 collected tests are all `tmp_path` fixtures; the repo-wide scan lives only in its `__main__` CLI), so any "N passed" figure quoted for it is not a leak claim. Wants its own small PR: reword with invented content preserving only the shape — never an exclusion entry.

### Loop status

Six rounds; **five found something real**, and rounds 3–6 each found a defect introduced by the previous fix, all inside the supporting measurements. **The decision itself has not moved since round 1** and no class-(A) defect remains in it. I am stopping the loop here rather than running a seventh, because the last three rounds have been correcting corrections in text that does not change the decision.

ZacxDev:
## Correction: the espanso commit was NOT unique work — I measured against a stale ref

In my previous comment I said the `:alo` trigger from `62e5417` "exists nowhere else — absent from `origin/main`", and used that to justify preserving it. **That was wrong, and wrong in the way this whole PR is about**: I checked against a remote-tracking ref that was already stale. `main` has since landed the same `:alo` trigger with **better** search terms (`["anything" "left" "outstanding" "loose"]` rather than a wholesale copy of `:acq`'s ten), plus a refined `:acq`.

The merge surfaced it as a conflict in `nix/home.nix`, which is the only reason I looked again. So this branch's copy was a **superseded earlier draft**, and dropping it strands nothing.

**Resolved wholly in main's favour** — `nix/home.nix` on this branch is now byte-identical to `origin/main`, verified by an empty `git diff`, and the file parses (`nix-instantiate --parse`). The branch's own contribution against its merge base is one file:

```
claudedocs/handoff-subsystem-store.md | 231 +++++++++++++-
1 file changed, 229 insertions(+), 2 deletions(-)
```

### Also worth recording: `main` moved twice while I was merging

`4548e6b` → `c7eb5c3` → `e7ceb1f`, inside a few minutes. A `git diff origin/main HEAD` run across that window showed another session's newly-added file as a **deletion by me**, which it was not — the comparison target had moved. Diff against the **merge base**, not against a branch someone else is pushing to.

### Gate

There is no automated gate on this repo (`<!-- merge-gate: none -->`), so it is run by hand on the merged tree. Node tier on the merged tree: **PASS — 4 suites / 33 files / 1119 tests, 1119 pass, 0 fail** (floor 1098). The pytest tier run outside `nix develop` returns `exit=2` = *could not run* (missing `logrotate` from PATH), which is the documented environment case and not a failure; it is re-running inside the dev shell now and I will post the verdict.

PR #615, opened 2026-08-20
fix(hooks): a home-manager ROLLBACK re-registers every hook FOREVER — and the test seam grew settings.json without bound

ZacxDev:
#609 pinned every managed hook's interpreter to an absolute `/nix/store` python and
closed the ~1s window in which `bash-guard.py` fails OPEN. Its adversarial audit passed
it as safe to merge and left three 🟡 findings, all silent failure modes. This closes
all three.

**Measured mitigating fact first, so nobody reads this as an outage:** neither host is
in any of these states today. Read-only inspection of the live
`~/.claude/settings.json` on **both** the workbench and the laptop (over nebula) says
the same thing — 14 managed hook commands, every one already pinned to an absolute
store path, **zero** duplicate `(event, matcher, script)` keys, **zero** commands in a
spelling the registrar declines to pin. All three fixes are tripwires for the future,
not live bugs.

## Finding 1 — a `home-manager rollback` double-registers every hook, permanently

`home-manager rollback` / `--switch-generation` (and a hand-run from an older checkout)
runs the PREVIOUS registrar against a `settings.json` the current one already pinned.
That older code keyed "already registered?" on the exact string `python3 ~/…`, does not
find it, and appends a SECOND copy of every hook it owns. Re-running the CURRENT
registrar then normalises both copies to identical strings, and because its append pass
asks only "is this script registered on this event", it sees the hook as present and
**both survive**. Every managed hook fires twice, forever — duplicate ledger rows,
double notifications, and `clawgate-writeback-guard.py` (the only hook that can BLOCK)
running twice per event.

**Fix: a third surface — DE-DUP — that HEALS the file rather than merely not making it
worse.** It is the narrowest of the three, and the boundaries are deliberate:

* **Identity is `(matcher, managed_script_of(command))`.** The script half is the same
  identity the append surface already uses (`registered_scripts`), so the two cannot
  disagree about what "already registered" means, and it catches the cross-spelling
  duplicate (`~/…` vs `$HOME/…`) a string key would miss. The matcher half is what keeps
  a hand-scoped narrow entry alive: two entries for one script under *different*
  matchers are NOT duplicates, and collapsing them would silently widen a scope the
  operator narrowed. A missing `matcher` and an explicit `null` are one identity.
* **The FIRST occurrence survives**, untouched — its position, its matcher, and every
  key the registrar has never heard of.
* **It removes only what it could have written**: exactly `{"hooks": [h]}` or
  `{"matcher": M, "hooks": [h]}` with `h` exactly `{"type": "command", "command": …}` and
  no arguments after the script path, AND only for a script this registrar actually
  registers. So a doubled **bash-guard** — whose registration belongs to the host, not to
  this script — is re-pinned and LEFT. An entry with a `description`, a `timeout`, a
  second hook, or a `--flag` is a configuration somebody made; it stays.
* **…and it says what it declined.** Anything still registered twice under one matcher
  after the pass is named on stderr. Declining silently is the same failure mode as
  finding 3.

## Finding 2 — `$DEVRC_HOOK_PYTHON` was a test seam with production reach and no validation

Activation inherits the operator's environment, so an exported override was written
verbatim into all 14 hook commands. An interpreter carrying an argument
(`python3 -X utf8`) or a non-`python*` basename makes `managed_script_of()` fail to
recognise the registrar's OWN output, so the append pass re-appends everything on every
run. **Measured against the pre-fix code: 14 → 27 → 40 commands over three runs, +13
each time, unbounded and silent.**

**Fix:** the override is validated on four conditions, and the first one is derived from
the recogniser rather than restating its rules, so the two cannot drift apart:

1. a command built from it must be one `managed_script_of()` recognises;
2. absolute (a relative name reopens the exact PATH window this pinning closes);
3. it exists;
4. it is executable.

**A rejected override FALLS BACK to `realpath(sys.executable)` with a warning; it does
not hard-fail.** Deliberate, and argued in the code: the fallback is what activation
wants anyway (home.nix invokes the registrar under `${pkgs.python312}/bin/python3`), so
falling back produces a *correctly* pinned `settings.json`, while hard-failing would
leave the hooks on whatever they carried before — i.e. it would keep the 127 window open
on the strength of a stray environment variable. It also cannot endanger the activation
wrapper's exit-0 contract, because it produces no non-zero status at all.

**Plus the self-check the finding asked for**, at both write points: a command the
registrar is about to write must be one its own recogniser recognises. Both are labelled
in the source as INVARIANT GUARDS, not regression coverage — with `PYTHON` validated
they cannot fire from any test. Reachability was demonstrated by mutation instead, and
each dies with **its own** message:

* drop `hook_python`'s validation → the module-level refusal fires, `rc=2`,
  `WARNING refusing to touch …/settings.json: the interpreter this script resolved … cannot be written into a hook command`,
  and `settings.json` is never opened;
* drop that too → `with_python()`'s `AssertionError` fires at the first table entry.

## Finding 3 — the rewrite was silent about hooks it declined to pin

`python3 -u ~/…`, `PYTHONPATH=… python3 ~/…`, `${HOME}/…`, a quoted path and
`cd /x && python3 ~/…` all come back byte-identical (correctly — the registrar cannot
prove it owns them) and all stay in the fail-open window with nothing in the switch log
to say so. This whole PR line exists because a silent failure ran for months.

**Fix:** every command that NAMES a managed hook basename but is not in the pinnable form
is reported on stderr, naming the event and the exact command. The activation wrapper
captures the registrar with `2>&1` and relays it line-by-line, so these land in the
switch log — that `2>&1` is now documented as load-bearing in the wrapper's contract.

## Nit — the last bare-`python3` spelling: CHANGED

`clawgate-writeback-guard.py`'s `dismiss_cmd()` advertised
`python3 ~/.claude/hooks/clawgate-writeback-guard.py --dismiss …` to the model as its one
escape from a block. Changed to `sys.executable`. It is free and exact — the hook is
already RUNNING under the pinned absolute interpreter — and a bare name would give the
model `command not found` for the one command that clears the block, during the same
window. New test asserts the STATE (absolute, equal to the running interpreter), not the
absence of the string `python3`, since the store path contains it.

## `scripts/opencode/plugin/guard.js` — NOT done, deliberately

`process.env.DEVRC_GUARD_PYTHON || "python3"` has the same mid-switch window. It fails
CLOSED (during the window every opencode bash call is refused with a clear error), so it
is not a fail-open hole. Pinning it is **not** a small change in the same style:
`guard.js` is deployed as a plain `home.file.….source` copy, so pinning needs a
substitution derivation (`.source` → `.text` + a `@PYTHON@` placeholder), a fallback path
for running it out of the repo where the placeholder is unsubstituted, and updates to the
node-tier deployed-layout tests. Left as follow-up rather than ballooning this PR.

## Testing

Every fix has a regression test shown RED on the pre-change tree. Matrix — base
`d3ce028` (the #609 squash), tests from this branch, implementation swapped via
`git checkout d3ce028 -- <impl>`:

| finding | scenario | at `d3ce028` | at HEAD |
|---|---|---|---|
| 1 — de-dup | `test_register_nudge_hook.py` scenario 12 | **RED** (9 checks) | green |
| 2 — validated seam | scenario 13, 6 hostile overrides × 3 runs each | **RED** (20 checks) | green |
| 3 — unpinnable warnings | scenario 14 | **RED** (6 checks) | green |
| nit | `test_the_dismiss_command_names_an_absolute_interpreter` | **RED** (`'python3' != '/nix/store/…/bin/python3'`) | green |

35 checks red at base in the standalone suite, 0 at HEAD. `test_registrar_activation.py`
is green at base and gained no regression coverage — only fixture fidelity (its
`HOOK_PY` had to become a real executable file, since the override is now validated);
that is stated rather than counted.

Scenario 12 builds an actually-duplicated `settings.json` — the exact 13 extra
registrations an older registrar appends — and asserts the healed end state as a literal
per-event `(matcher, command)` list, plus that four survivors are untouched: a
different-matcher entry, an argument-carrying one, one with a foreign hook-level key, one
with a foreign entry-level key, and the doubled bash-guard whose registration this script
does not own. Scenario 13 runs each hostile override three times and pins the command
count at a literal `[15, 15, 15]` — not `a == b == c`, which a run that appended nothing
would also satisfy.

**Mutation sweep: 17/17 killed**, under `PYTHONDONTWRITEBYTECODE=1`, each mutant applied
to a pristine copy and restored by `cp` (never `git checkout`), each scored KILLED only
when a failing check belongs to the scenario that OWNS the guard. Negative control: the
pristine tree is green. Positive control (M0): a two-space separator in `with_python` —
still recognised by the regex, so only the expectations catch it — killed.

Mutants: delete the de-dup decision · drop the matcher from the identity · invert the
branch so the FIRST copy is removed · drop the entry-shape check · drop the ownership
filter · drop the no-arguments check · mark an entry seen BEFORE deciding on it (stale
rebind) · raise the declined-duplicate threshold · delete the override validation · invert
it to accept only INVALID overrides · drop the absolute check · drop the recogniser
self-check · drop the exists check · delete the unpinnable warning · invert it to warn
about the PINNED commands · swap the operands of the basename scan.

**Three of those SURVIVED the first sweep**, each masked by a neighbouring check, and
fixing them is a real coverage gain rather than a scoring adjustment:

* the entry-shape check was masked by the ownership filter, because the only
  foreign-entry-keyed duplicate in the fixture was bash-guard's → added a
  foreign-entry-keyed duplicate of a script the registrar DOES own;
* `os.access(X_OK)` is False for a path that does not exist, so it masked the exists
  check → added an override that is a real, executable **directory** with a `python*`
  name;
* a bare relative name is not a file in any plausible cwd, so the exists check masked
  the absolute check → added an override that is relative **and resolvable from the cwd
  the registrant runs in**.

## Docs

`register-nudge-hook.py`'s module docstring now says THREE surfaces (it said two) and
carries the rollback mechanism; `register-hooks-activation.sh`'s contract items 2 and 3
name the third surface and pin the `2>&1` capture as load-bearing; the identity key, the
ownership boundary and the fall-back-don't-hard-fail decision each carry their argument
at the code that implements them.

Gate: `scripts/gate.sh --tier both --set all`.

### DIFF
--- a/scripts/claude-hooks/clawgate-writeback-guard.py
+++ b/scripts/claude-hooks/clawgate-writeback-guard.py
@@ -1261,9 +1261,20 @@ def dismiss_cmd(task_id, session_id):
     cannot see its own hook payload, so a command it has to fill in a session id for
     is a command it cannot run — which is how the previous escape ("say so and stop")
     ended up being no escape at all.
+
+    🔴 The interpreter is `sys.executable`, not a bare `python3` — this was the last
+    bare-`python3` spelling of a managed hook left in the tree after #609 pinned the
+    registrations in ~/.claude/settings.json. It is free and exact: this hook is
+    already RUNNING under the pinned absolute interpreter, so `sys.executable` is
+    that path. A bare name would be resolved from PATH, and PATH has no python3 for
+    the ~1s of every `home-manager switch` in which the intermediate profile
+    generation is a partial closure — the model would get `command not found` for
+    the one command that clears the block. The fallback covers an embedded
+    interpreter, where sys.executable is empty; nothing here runs that way.
     """
-    return ("python3 ~/.claude/hooks/clawgate-writeback-guard.py --dismiss %d "
-            "--session %s" % (int(task_id), _sanitize(session_id)))
+    return ("%s ~/.claude/hooks/clawgate-writeback-guard.py --dismiss %d "
+            "--session %s" % (sys.executable or "python3", int(task_id),
+                              _sanitize(session_id)))
 
 
 def missing_text(task_id, first_read_ts, session_id=""):
--- a/scripts/claude-hooks/register-hooks-activation.sh
+++ b/scripts/claude-hooks/register-hooks-activation.sh
@@ -20,20 +20,31 @@
 #      "failed switch" incident class (CLAUDE.md → Git discipline). Every
 #      failure path therefore warns on stderr and returns 0.
 #   2. IT SAYS WHAT IT DID. The registrar's own report ("registered hooks: …"
-#      / "… no change") is echoed line by line, prefixed. A silent activation
-#      step is how the original failure went unnoticed for a full deploy cycle.
+#      / "removed duplicate hook registrations: …" / "… no change") is echoed
+#      line by line, prefixed. A silent activation step is how the original
+#      failure went unnoticed for a full deploy cycle. 🔴 The capture is
+#      `2>&1`, which is load-bearing: the registrar sends its WARNINGs to
+#      stderr — a rejected $DEVRC_HOOK_PYTHON, a managed hook command in a
+#      shape it cannot pin, a duplicate registration it declined to remove —
+#      and this is the only thing that puts them in the switch log. Never
+#      split the streams here or those warnings become invisible again.
 #   3. IT DECIDES NOTHING ABOUT settings.json. Every read/merge/write belongs
-#      to the registrar. That registrar has TWO surfaces of different widths —
+#      to the registrar. That registrar has THREE surfaces of different widths:
 #      an APPEND surface (its own command tables, never touching a hook it does
-#      not own, so the clawgate Stop hook that drives remote approval survives)
-#      and, since 2026-08-20, a narrow REWRITE surface that normalises the
-#      INTERPRETER TOKEN of a devrc-managed hook command to an absolute
-#      /nix/store python. Read its module docstring before changing either; the
-#      claim this file used to make — "strictly APPEND-ONLY" — is no longer
-#      true and deleting the rewrite path would silently reopen the failure it
-#      closes (a hook firing mid-switch dies with `python3: command not found`,
-#      and bash-guard fails OPEN when it does). This wrapper adds no second
-#      writer — it must never grow one.
+#      not own, so the clawgate Stop hook that drives remote approval survives);
+#      a wide REWRITE surface that normalises the INTERPRETER TOKEN of a
+#      devrc-managed hook command to an absolute /nix/store python; and a
+#      narrow DE-DUP surface that deletes a second registration of a script it
+#      owns on the same event under the same matcher, keeping the first. Read
+#      its module docstring before changing any of them; the claim this file
+#      used to make — "strictly APPEND-ONLY" — is no longer true and deleting
+#      the rewrite path would silently reopen the failure it closes (a hook
+#      firing mid-switch dies with `python3: command not found`, and bash-guard
+#      fails OPEN when it does). The de-dup path is what makes a
+#      `home-manager rollback` — which re-runs an OLDER registrar that cannot
+#      see the pinned spelling and appends a second copy of everything —
+#      recoverable rather than permanent. This wrapper adds no second writer —
+#      it must never grow one.
 #
 #      🔴 The interpreter the registrar writes is derived from its OWN
 #      sys.executable, so the `[python]` argument below is load-bearing beyond
--- a/scripts/claude-hooks/register-nudge-hook.py
+++ b/scripts/claude-hooks/register-nudge-hook.py
@@ -5,21 +5,30 @@ settings.json is per-host and unmanaged (holds permissions/allowlists/secrets), 
 hook *scripts* are symlinked by home-manager but their *registration* is applied by
 running this once per host.
 
-🔴 THIS SCRIPT HAS **TWO** SURFACES, AND THEY ARE DELIBERATELY DIFFERENT WIDTHS.
+🔴 THIS SCRIPT HAS **THREE** SURFACES, AND THEY ARE DELIBERATELY DIFFERENT WIDTHS.
 Until 2026-08-20 there was only the first, and its docstring said "strictly
-APPEND-ONLY / never rewrites". That is no longer true — read both before editing:
+APPEND-ONLY / never rewrites". That is no longer true — read all three before editing:
 
-  * APPEND surface (NARROW, unchanged): the command tables below. An entry is added
-    only when its exact command string is absent from that event's existing entries.
-    Nothing else is ever appended. In particular bash-guard is NOT in these tables and
-    must never be: this script does not own its registration, only its interpreter, so
-    if a host has no bash-guard entry it must come back with no bash-guard entry.
-  * REWRITE surface (WIDE, new): every hook entry on every event whose command invokes
+  * APPEND surface (NARROW, oldest): the command tables below. An entry is added
+    only when that event does not already invoke the script. Nothing else is ever
+    appended. In particular bash-guard is NOT in these tables and must never be:
+    this script does not own its registration, only its interpreter, so if a host
+    has no bash-guard entry it must come back with no bash-guard entry.
+  * REWRITE surface (WIDE): every hook entry on every event whose command invokes
     a devrc-managed hook script gets its INTERPRETER TOKEN normalised to an absolute
     /nix/store python — bash-guard INCLUDED. Only that token changes. The entry keeps
     its position in its array, its `matcher`, its `type` and every other key verbatim,
     and a command that does not resolve to a managed hook script comes back
     byte-identical.
+  * DE-DUP surface (NARROWEST, newest): a second registration of the SAME managed
+    script on the SAME event under the SAME matcher is deleted, keeping the FIRST
+    occurrence with its position and every key it arrived with. It is narrower
+    than the append surface in one direction and equal in the other: it removes
+    only an entry of EXACTLY the shape this script writes (no foreign key, one
+    hook, no arguments after the script path) AND only for a script this script
+    registers — so a doubled bash-guard, whose registration belongs to the host,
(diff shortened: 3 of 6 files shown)


ZacxDev:
## Delta re-audit (`d3ce028..HEAD`) — verdict: safe to merge

All four prior findings verified fixed by **reproducing each original symptom**, not by reading the diff.

| # | prior finding | status | evidence |
|---|---|---|---|
| 1 | rollback double-registration; `d3ce028` could not heal it | **fixed** | The real pre-#609 registrar (`9592060`) takes 13 commands → **26**; `d3ce028` run twice leaves it at **26** (prior finding confirmed); this PR **heals to 13**, reports exactly 13 removals, and re-running is a byte-identical fixed point. A doubled `bash-guard` entry is re-pinned but **not** deleted — the ownership boundary holds. |
| 2 | `DEVRC_HOOK_PYTHON` unvalidated → unbounded growth | **fixed** | With `DEVRC_HOOK_PYTHON="/nix/store/zzz-python3/bin/python3.12 -X utf8"`: `d3ce028` = 13 → 26 → 39; this PR = **13 → 13 → 13**, rejected out loud, falls back to `realpath(sys.executable)`. Driven through the real activation wrapper: `rc=0`, warning relayed to the switch log, `settings.json` md5 unchanged. |
| 3 | silent about unpinnable spellings | **fixed** | All five shapes (`-u`, `PYTHONPATH=`, `${HOME}`, quoted, `… && …`) warn exactly once and come back byte-identical; a command naming no managed hook stays silent; the warning does not trigger a write. |
| 4 | `dismiss_cmd` bare `python3` | **fixed** | Now `sys.executable or "python3"`, asserted on state, still driven as a real subprocess by the existing test. |

Nothing was made worse. **Blast radius on the live hosts is zero:** the new registrar was run against read-only copies of *both* hosts' `~/.claude/settings.json` — both report "no change", byte-identical, zero de-dups, zero warnings.

**Regression hunt on the de-dup pass** (the new code that deletes entries): an independent 20-scenario adversarial harness could not make it remove an entry it does not own. Three copies collapse to the first with foreign entries keeping position and order; a foreign entry-level key, a foreign hook-level key, an argument after the path, a trailing space, a multi-hook entry, a different matcher, and any `bash-guard` entry all survive; non-dict / `hooks: []` junk is preserved verbatim. Atomic `mkstemp`+`os.replace` intact on every path, and every non-zero exit happens *before* `open(SETTINGS)`, so the wrapper's "left as it was" claim is true.

**Mutation claims re-verified independently:** 10 fresh deletion-free mutants (operand swap, branch inversion, comment-out-with-text-retained, off-by-one, ownership widening, pass reordering) under `PYTHONDONTWRITEBYTECODE=1`. 8 killed, each by its own scenario's failure — including the three the author reported as previously masked. Positive control killed; harness negative control run in both tiers.

## New findings — follow-up, none deploy-blocking

**🟡 1. `register-nudge-hook.py:299` — the `os.access(X_OK)` guard has no covering test.** A mutant deleting it **SURVIVED** the full suite. The branch is reachable and emits its own `not executable` message (a 0644 python-named file is rejected and the fallback used), so the guard is correct — but the reported "17/17 killed" does not cover the fourth of `unusable_interpreter`'s four conditions. An uncovered guard is a guard with no evidence: a future refactor can delete it silently. One fixture — a non-executable regular file — closes it.

**🟡 2. `register-nudge-hook.py:30` — "Anything still doubled after the pass is named on stderr" is wider than the code.** Counterexample: two entries for the same script on the same event, one with `matcher: ""` and one with the key absent, both survive and produce **no warning** (`entry_identities` treats them as distinct; only absent-vs-explicit-null are unified, per line 480). Related and **hedged**: on the matcher-less events (`Stop`, `SubagentStop`, `UserPromptSubmit`, `SessionStart`) a `matcher` key may not be a scope at all — which would make the fixture's `SURVIVOR A` (`test_register_nudge_hook.py:592`, next-step-nudge with `matcher:"Bash"` **on `Stop`**) a real double-fire rather than "a hand-scoped narrow entry" as its comment claims. Claude Code's matcher semantics could not be confirmed from an authoritative local source (the CLI ships as a compiled binary), so treat that half as unverified. Behaviour is conservative either way — nothing is deleted; the gap is in the *reporting*.

**🟢 3. `register-nudge-hook.py:518`** — the "de-dup runs FIRST so the report cannot claim to have re-pinned an entry it is about to delete" claim is real but untested: reordering the passes leaves all 134 checks green, while the mutant's report demonstrably lists the duplicate under both "pinned" and "removed".

**🟢 4. `register-nudge-hook.py:340` — comment narrower than reality.** "Reaching it requires running the registrar under a CPython whose realpath basename is not `python*`" — an empty `sys.executable` reaches the same guard via the `isabs` branch, which also means `hook_python`'s `"python3"` fallback can never actually be written.

**🟢 5. `register-nudge-hook.py:537` — cosmetic.** The removal report prints the *pre-rewrite* spelling next to a "pinned … to /nix/store/…" block, which reads oddly.

**🟢 6. Pre-existing, NOT introduced** (verified against `d3ce028`): `registered_scripts` (line 450) raises `AttributeError` on a non-dict entry in an event array, while the delta's new `_entry_hook_dicts`/`entry_identities` handle exactly that defensively. Contained by the wrapper (exit 0, file untouched). The delta makes the inconsistency visible; it does not create it.

## Unrelated flake worth knowing about

`scripts/tests/test_subsystem_store_api.py::TestTrustedProxyOverTheRealProcess::test_a_VALID_token_from_an_untrusted_peer_is_SERVED_but_bucketed_under_the_PEER` can fail with `IndexError` (SIGTERM racing the child's stdout flush). The control on `d3ce028` is green, and that file and everything it exercises are **byte-identical across this delta** — a pre-existing latent flake, not attributable to this PR. It passes hermetically.

## Gates run by hand (this repo has no automated merge gate)

- Merged-tree gate on an integration branch off current `main`: `GATE: RESULT=PASS exit=0` — pytest and node tiers both `RESULT: PASS`, all targets above floor.
- Hermetic tier: 2151 passed across `scripts/claude-hooks/tests/` plus the captured-text, captured-markup and runtime-shebang gates.

PR #632, opened 2026-08-21
fix(ship): the run that DELIVERS a ship.sh fix verified with the copy it replaced, and "converged" was printed over two different commits

ZacxDev:
Two defects in `scripts/ship.sh`, both observed live on 2026-08-19, both in the class this file exists to refuse: **a green computed by something other than what was asked.**

## 1 — the run that SHIPS a `ship.sh` fix verifies with the copy it replaced (rc 20)

`ship.sh`'s own source is one of the things `ship.sh` ships, and the first act of a run is to fast-forward the tree that source lives in. The `CONVERGE` payload is expanded into a shell variable near the top of the file, long before that — and it is that string which runs locally and goes over `ssh`.

Measured: the run that **delivered #620** — the commit that *added* `verify_managed_currency` — printed no currency line at all and exited 0. An immediate second run printed `347 repo-sourced examined, 0 stale`.

`ship.sh` now fingerprints itself and `lib/host-role.sh` (`cksum`: CRC **and** size, so a trailing-newline-only edit cannot compare equal) before the local leg and again after it, and **re-execs the new copy — before the remote leg**, so the laptop is sent the payload this run delivered rather than the superseded one.

* `SHIP_SELF_GEN` bounds it at **one** re-exec; a script that keeps changing gets rc 20 instead of a loop. The budget is a constant, deliberately not env-overridable.
* A self-check that cannot be **measured** is rc 20 too: *"I could not look"* must not print the same as *"nothing changed"*.
* The success line carries the compared count (`2 files compared (…), 0 superseded`) — same reason `0 dangling` carries `N checked`.
* Cost, stated rather than hidden: a run that ships a change to `ship.sh` runs `home-manager switch` twice on the local host.

## 2 — "converged" was printed over two DIFFERENT commits (rc 19)

In that same run:

```
[nixos] fast-forwarded main 4548e6b -> c7eb5c3      (workbench)
[nixos] fast-forwarded main 4548e6b -> e7ceb1f      (laptop)
ship: converged + verified at origin/main (local=workbench remote=laptop)
```

`#619` merged between the two legs' fetches. **Every per-host check was true** — each host really was at `origin/main` as *it* saw it — and the fleet was in two states. rc 11 structurally cannot see this: it compares `HEAD` against the `target` **that leg** captured, so a target that moved afterwards is invisible to it, and nothing compared the hosts to each other at all.

Each converged host now emits `ship-landed-sha` *after* all of its checks pass, and the verdict is a claim about those two shas being **equal**, printed with the sha and the number of hosts compared:

```
ship: converged + verified — 2 hosts compared, both at <sha> (local=workbench remote=laptop).
```

* Fewer shas than hosts in scope **while the run exits 0** is also rc 19 — an agreement that was not compared is not an agreement, and a leg that exits 0 must have emitted one.
* A host that was **skipped** keeps its own code (rc 3/7/8/…); the missed comparison is recorded, never allowed to mask the actionable diagnosis.
* A one-host run says `cross-host agreement NOT COMPARED` rather than implying two.
* 🔴 It does **not** claim the agreed sha is the newest `origin/main` — nothing re-fetches afterwards. That limit is written in the header rather than left to be assumed.

Also: `--no-local --no-remote` was a run that checked no host and printed "converged + verified"; it is now **rc 2**, matching `drift-check.sh`'s own spelling of that refusal. rc 19/20 skip 14–18, which `drift-check.sh` owns — the two ladders are deliberately aligned.

## 🔴 Refuted from the brief, with measurements

* **"ship.sh's header already documents a related hazard (a merge landing mid-run reading as divergence, rc 11)."** It does not. No such note exists in `ship.sh` or `drift-check.sh`, and rc 11 cannot fire on it for the reason above. rc 19 is a new question, not a re-spelling.

* **"bash reads scripts incrementally — replacing the file mid-execution is also a corruption hazard."** True of an **in-place** writer, **not of `git`**. Comparing `st_ino` across `git merge --ff-only` reports the **same** inode and reads as an in-place overwrite — the wrong instrument, because a freed inode *number* is reused immediately. Holding an **open fd** across the same merge shows the fd still yielding the OLD bytes with `st_nlink=0`: git **unlinks and recreates**, so a running bash keeps its own unlinked inode. Three end-to-end runs of the **pre-fix** script (incoming copy +27 KB, −30 KB, same size) all completed cleanly.
  What *does* corrupt it is `cp`-style truncation: measured losing everything past the write and **exiting 0** at **45 of 45** swept offsets, and splicing a padding line into the command stream at equal length. The body is wrapped in a brace group ending in `exit` — bash then parses to EOF before running anything — but as a cheap close on a **latent** class, not as a fix for a reproducing bug, and the header and the test both say so.

## Test matrix — `scripts/tests/test_ship_converge.py`, 43 → 59 tests

17 test functions are new or rewritten (+16 net — one replaces the widened rc-ledger test, which was renamed because it now scans the DRIVER as well as `CONVERGE`; that blind spot is why rc 2 had been undocumented in both ledgers since the file was written).

**16 of those 17 are RED at base `d12f84c` and green at HEAD.** The 17th (`test_a_host_that_did_not_converge_emits_no_landed_sha`) is green at base and is labelled an **invariant guard**, not regression coverage.

The two most direct, at base:

```
test_a_merge_landing_between_the_two_fetches_is_not_convergence
  AssertionError: expected rc19 (hosts disagree), got 0
    ship: converged + verified at origin/main (local=workbench remote=laptop).

test_a_run_that_ships_a_change_to_ship_sh_verifies_with_the_new_copy
  AssertionError: the run ENDED on the superseded copy's payload (markers seen: [1]).
```

Two hosts are fabricated **hermetically** — two clones of one origin, each with its own fake `$HOME` and home-manager generation, reached through an `ssh` shim that fast-forwards `origin/main` onto a staged commit at the instant the remote leg starts. The version marker is planted **inside the CONVERGE payload**, because that is the only placement that answers *which copy produced the payload that ran*. **Nothing here touches a real host, and no test runs `home-manager switch` or `drift-check.sh`.** The `ssh` shim and the `home-manager` stubs are POSIX `sh` via `testlib/mockbin` (no `/usr/bin/env`, no `PIPESTATUS`), and no GNU-only flag was added to anything inside `CONVERGE`.

## Mutation sweep — 14 mutants

Every one **killed by its named test, with that test's own assertion text**. A **null** (comment-only) mutant **SURVIVED**; a deliberately-broken verdict was killed. Patch sites asserted unique, `ship.sh` restored in a `finally` with SHA-256 identity asserted, whole suite run per mutant so "killed by the wrong test" is visible.

Two findings the sweep produced that prose review had not:

* **the re-exec-after-the-remote-leg mutant SURVIVED** the marker assertions — a re-exec redoes everything, so both hosts end on the new payload either way. The test now pins the **order** and that the remote host is visited **once** (on the real fleet the difference is a whole extra `home-manager switch` on the laptop, run from superseded logic).
* removing **only** the closing brace is a syntax error that kills 51 tests for the wrong reason; the mutant removes both and is `bash -n` checked first.

## Gate

`nix build .#checks.x86_64-linux.pytests` on the **merged** tree (this branch with current `origin/main` merged in) — see the comment below for the run against the final commit. The pre-merge run on `cb5984c` read:

```
TOTAL collected=13442  passed=13441  skipped=1  failed=0  (floor: 12749 = sum of 24 per-target floors)
RESULT: PASS (exit=0)
```

Floors are minimums, so the +17 tests need no `TARGET_FLOORS` re-pin (`scripts/tests` is an existing directory target) — deliberately left alone to avoid churn on a line the repo calls a conflict magnet.

---

# 🔴 THIS PR MUST BE SHIPPED TWICE

The run that **delivers** this change executes the **pre-#632** `ship.sh` — the copy without the self-check. So the delivering run will not re-exec, will not fingerprint its own source, and will not cross-host-compare: it fast-forwards `ship.sh` out from under itself and then keeps going on the superseded logic, which is precisely the bug this PR fixes.

**The first `ship.sh` run after this merges is the last instance of the defect.** Run it, then run it again. Only the second run exercises the fix, and only the second run's verdict is a claim about the fleet. A single green run here means "the old script said so".

---

# Audit-fix round (blind adversarial audit — no 🔴 found)

`🟡1` **the suite's only defence against ssh'ing the real laptop was one word of argv.** With `REMOTE_SSH`/`LAPTOP_SSH` unset, `remote_ssh_of workbench` returns `LAPTOP_SSH_DEFAULT` — the operator's actual laptop — so `--no-remote` in `Repo.ship`'s argv was the whole sandbox, and this PR newly made argv something that has to survive an `exec` for that to keep holding. Every fixture now runs with a **refusing `ssh` and `home-manager`** at the front of `PATH`, composed in `Repo.env()`, which **rejects a raw `PATH=` override** so a call site cannot quietly drop it (stubs go in as `shims=[dir]`, ahead of the refusing pair). `test_the_sandbox_refuses_a_run_that_reaches_ssh` is the reachability proof and reads the real fallback target out of `lib/host-role.sh` rather than hardcoding it. `test_the_re_exec_carries_the_scripts_own_argv` pins argv survival with `SHIP_NO_SWITCH=0`, so the `(SHIP_NO_SWITCH) skipping` count is the assertion and argv is the only carrier of the flag.

`🟡2` **bash's own `exit 2` was indistinguishable from the new rc 2.** A superseding copy that does not parse exits 2 *after* the local host is already fast-forwarded and switched — under a code documented as "raised before any host is touched". Closed with `bash -n` before the exec, which becomes rc 20. The header now also records *why* it is closed at the source rather than added to the ledger: `_ship_exit_codes` scans the FILE for `exit N`/`rc=N` and is structurally blind to statuses **bash** produces.

`🟡3` **the "two ladders are aligned" claim existed only in a PR description.** `drift-check.sh` still said the next DRIFT code had "nowhere to go but upward" — pointing it at **19**, which ship had just taken. Both files now carry a machine-read `RESERVED-TO-*` ledger, pinned **as sets** (fails when either grows *or* shrinks) by `test_the_two_rc_ladders_reserve_each_others_codes`.

`🟡4` every `exit 20` site sits **between the two legs**, so the fleet is provably in two states; each site now prints which host got what, and a re-exec announces that it is **discarding** the local leg's non-zero rc instead of letting pass 2 silently re-attempt a failed switch.

Also folded in: the `SHIP_SELF_GEN` sanitiser is now **reachable and covered** (garbage would make `[ x -ge n ]` *error*, which without `set -e` is simply FALSE — the budget check passing by erroring); the unreachable `rc20=` legend line is **acknowledged as unreachable** rather than deleted or silently enforced; the home-manager log honours `$TMPDIR` and is removed on **both** paths (a superseding run wrote it twice); the leg capture files are `$TMPDIR`-aware and trap-cleaned; and `test_the_self_check_reports_how_many_files_it_compared` runs a **frozen copy** instead of fingerprinting the live working tree, where a concurrent session's edit would have read as a supersession bug.

## Two corrections to that fix round — both measured, not reasoned

**The syntax-error fixture landed where bash never looks.** `SHIP_SYNTAX_ERROR` was appended *after* `ship.sh`'s closing brace. Measured: `bash -n` rejects that file (rc 2) but **running it is completely clean** — the body is one `{ … }` group ending in `exit`, so bash exits before it ever parses the trailing line, and the copy exits **0**. The fixture could not produce the bash-2/ship-2 collision the guard is named for; the guard-deletion mutant died on rc 0-vs-20, for a reason unrelated to the finding. The `fi` now goes **inside** the group (measured rc 2), and the test runs the broken copy through the sandbox and **requires bash's own 2** before believing the guard prevents anything.

**`shopt -s execfail` never covered what it claimed.** It was documented as closing "the new copy removed between the parse check and the exec". Measured: `exec bash "$f"` execs **bash** — `$f` is an argument — so the exec syscall succeeds whatever state `$f` is in, and the freshly-exec'd bash exits **127** with this process already replaced; the post-exec block never runs. (Control: `exec <missing binary>` under `execfail` *does* return, rc 20.) So `execfail` covers exactly one thing — bash itself being unexecutable — that removed-script 127 is now named in the header as **open**, and the block is labelled **unreachable from the suite** with its reason rather than counted as coverage.

**The sandbox test was green for the wrong reason.** Mutating the refusing `ssh` shim to `exit 0` left `test_the_sandbox_refuses_a_run_that_reaches_ssh` **passing** — the banner still printed, the real target still appeared, and `rc != 0` still held, because a silent-but-successful ssh returns no landed sha and ship exits **19** (agreement not compared) instead. Every assertion held while the shim had stopped refusing. The refusal's status is now a named constant used by both the shim and the assertion, and the test requires ship to report `converge exited 97` end-to-end.

## Red-at-base / green-at-HEAD — re-measured against the current merge base

The previously-quoted matrix cited `d12f84c`, which is neither the PR base nor a commit in this PR. Re-run against the **current** merge base `29ea6f81`:

| tree | `test_ship_converge.py` + `test_drift_check.py` |
|---|---|
| base `29ea6f81` (this branch's tests copied onto it) | **23 failed**, 310 passed |
| HEAD (merged with `origin/main`) | **333 passed**, 0 failed |

`test_the_sandbox_refuses_a_run_that_reaches_ssh` is green on **both** sides, and correctly so: it is a safety net over the fixtures, not a regression test for a code change. Labelled as such rather than counted in the 23.

## Mutation battery — 16 mutants, both controls

Full re-run after the fix round (a fix round resets the gate). Every mutant **KILLED by its own named test's assertion**:

| mutant | killed by |
|---|---|
| `"$@"` dropped at `ship_self_check "$@"` | `test_a_run_that_ships_a_change_to_ship_sh_verifies_with_the_new_copy` (+1) |
| `"$@"` dropped at `exec bash "$_ship_self" "$@"` | same pair |
| `bash -n` guard deleted | `test_a_syntactically_broken_new_copy_is_rc20_not_bash_2` |
| `SHIP_SELF_GEN` sanitiser deleted | `test_a_garbage_self_gen_is_normalised_not_trusted` |
| `ship_self_fleet_state` silenced | `test_an_exit_20_names_the_host_it_left_unconverged` (+1) |
| discarded-`lrc` note deleted | `test_a_re_exec_says_the_local_legs_status_is_discarded` |
| hm log back to hardcoded `/tmp` | `test_the_switch_log_lives_under_tmpdir_and_is_removed` |
| hm log not removed | same |
| `RESERVED-TO-DRIFT-CHECK` grown / shrunk | `test_the_two_rc_ladders_reserve_each_others_codes` |
| `RESERVED-TO-SHIP` grown / shrunk | same |
| refusing `ssh` shim made permissive | `test_the_sandbox_refuses_a_run_that_reaches_ssh` |
| `denybin` dropped from `Repo.env`'s PATH | same |

**Positive control** (budget-exhausted `exit 20` → `exit 0`): **KILLED**. **Negative control** (an inert comment inserted before `shopt -s execfail`): **SURVIVED**.

> A first negative control — editing the `# Exit codes:` comment — was **KILLED**, and correctly: that string is `ship_pad_lines`' unique insertion anchor, so the edit was not inert at all. Reported rather than quietly replaced, because a "null" mutant that dies is a fact about the harness, not noise.

Harness discipline: `PYTHONDONTWRITEBYTECODE=1` on every subprocess and `__pycache__` purged per mutant (a same-length edit inside one mtime second is otherwise invisible to CPython and scores SURVIVED without executing); every patch site asserted **unique** before application; originals restored in `finally` with **SHA-256 identity asserted** afterwards (clean); every subprocess bounded and killed by **process group** on timeout, since `subprocess.run(timeout=)` signals only the direct child.

## 🔴 The `"$@"` mutant — measured vs the auditor's inference

The audit **inferred** (did not execute) that dropping `"$@"` from `ship_self_check "$@"` would **SURVIVE**, and that `test_a_run_that_ships_a_change_to_ship_sh_verifies_with_the_new_copy` would reach `ssh <the real laptop>` and, if it answered, fast-forward the operator's real checkout.

Measured, with the refusing shim in place: the mutant is **KILLED**, by `test_a_run_that_ships_a_change_to_ship_sh_verifies_with_the_new_copy` **and** `test_an_in_place_overwrite_of_the_running_script_does_not_truncate_the_run` — i.e. the pre-existing tests already caught it once the sandbox made the failure survivable enough to observe. Same for the second site (`exec bash … "$@"`). The auditor's SURVIVES prediction was **wrong**; its danger analysis — that an argv-less pass 2 resolves the real laptop — was **right**, and is exactly why the sweep could only be run at all after the refusing shim existed.

**The sandbox's own positive control read the wrong PATH — and only the nix tier could see it.** The test opened with `assert shutil.which("ssh")`, which probes the **pytest process's** PATH while the run under test composes its own in `Repo.env` — and which makes "a real ssh exists on this host" a *precondition*, i.e. a pinned dimension the two tiers disagree about. Measured: green on this dev host, **FAILED in `nix build .#checks.x86_64-linux.pytests`**, where no `ssh` exists at all. Replaced with a hermetic probe that invokes `ssh` under `repo.env()` and requires the shim's own marker **and** exit status — an ENOENT can produce neither, which is what the control was actually for. Both tiers now exercise the same code.

## Gate — read from the runners' own verdict lines, BOTH tiers, on the merged tree

Merged with current `origin/main` (`29ea6f81`) first. Read from the derivation log (`nix log <drv>`) and the runners' `RESULT:` lines, never a piped exit code.

| tier | result |
|---|---|
| `gate.sh --tier node` | `suites=4 files=33 tests=1119 pass=1119 fail=0` — **`RESULT: PASS (exit=0)`** |
| `gate.sh --tier pytest --set all` | `TOTAL collected=13839 passed=13837 skipped=1 failed=1` (floor 13034 = sum of 25 per-target floors) |
| `nix build .#checks.x86_64-linux.pytests` (hermetic sandbox) | `TOTAL collected=13839 passed=13837 skipped=1 failed=1` — the **same** single failure |

**That single failure is pre-existing on `origin/main` and is not this PR's.** Bisected, not assumed:

```
scripts/tests/test_subsystem_recall.py::TestSkillDocsArePinned
  ::test_the_recall_step_comes_AFTER_the_handoff_is_read
  ValueError: substring not found   (doc.index("**Read it fully.**"))
```

Run against a **pristine `origin/main` worktree** with nothing from this branch on it: `1 failed, 444 passed`. Attribution — `da115695` (#643) reworded `claude/skills/resume/SKILL.md`:

```
-2. **Read it fully.**
+2. **Read it fully — but treat its "Open investigations" section as RECALL, not live state.**
```

while `test_subsystem_recall.py:3585` pins that line by **exact substring**. A guard on words, walked by a reword — in the breaking direction this time. `main` is red right now and, since no CI gates a devrc PR, nothing else is going to report it. **Not fixed here**: it belongs to #643's surface and a one-line fix on either side would put two unrelated changes in one PR. Flagged for a separate fix.

Every other target passed in **both** tiers. The stubs added here all go through `testlib/mockbin.write_exec`, so the `#!/usr/bin/env bash` two-tier hazard the file's own header warns about is avoided by construction — verified by the sandbox run above, not assumed.


ZacxDev:
## Gate — authoritative, on the MERGED tree, at the final commit

`nix build --no-link -L '…?ref=fix/ship-self-supersession-and-cross-host-agreement#checks.x86_64-linux.pytests'` against **6582dbe**, which is this branch with current `origin/main` (`0430bab`) already merged in — so this is the tree the merge creates, not the branch in isolation.

```
devrc-pytests> RESULT: all good
devrc-pytests>   TOTAL collected=13442  passed=13441  skipped=1  failed=0  (floor: 12749 = sum of 24 per-target floors)
nix exit=0
```

Read from the log, and the runner's own `RESULT:` line agrees with the build status — the two independent claims that `scripts/gate.sh` exists to cross-check.

`nodetests` was not run: this branch touches only `scripts/ship.sh`, `scripts/tests/test_ship_converge.py` and one line of `CLAUDE.md`, no JS.

**No CI will run on this PR** — `gh pr view 632 --json statusCheckRollup` returns **0 checks**, `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`. That matches the repo's own `<!-- merge-gate: none -->` marker. The run above is the only gate this change will ever get.

PR #637, opened 2026-08-21
fix(analyze-service): every entry says WHAT IT IS and no reader printed it

ZacxDev:
## The defect

`## What it is` is written by **73 of 73** entries in `~/.claude/analyze-service-index/` and was parsed by **nothing** on any read path — not `--ref`, not the digest, not `service_recon`'s `index:` block. `subsystem_touch --validate` said so out loud ("🔴 `## What it is` is NOT checked: no reader parses it") and that was treated as a note rather than as a defect report.

A controlled A/B found an agent briefed only on an entry could not say what the service was, where it lived or what it owned. `## Pointers` and `## Nuance / work-history` both assume the reader has already identified the thing; the one section that identifies it was never rendered. The content was on disk the whole time.

**Reproduced on the live store**, pre-change source vs this branch (counts and byte totals only — devrc is PUBLIC, no store content here):

| command | before | after |
|---|---|---|
| `subsystem_recall.py --scope datapacket-talos --ref storage-resolver` | `## What it is` lines = **0**, 23,996 B | **1**, 24,602 B |
| `service_recon.py storage-resolver` | **0**, 28,451 B | **1**, 29,027 B |

⚠ One correction to the brief: its quoted repro `grep -c 'THREE manifests'` returns 0 **for two reasons** — that literal string does not occur anywhere in that entry's file (checked: 0 occurrences in the whole file). It is a confounded zero. The defect is real and reproduced above using the heading itself as the sentinel.

## The decision (digest vs `--ref`), and the test that pins it

Surfacing on the single-entry paths is unambiguous. The risk was the multi-entry ones. **Where the per-entry multiplier actually lives is the one-line INDEX ROWS, not the digest** — the digest prints the full index (one line per entry) plus exactly **one** body.

So:

* **The FULL section, rendered FIRST, in every printed BODY** — `--ref`, the digest's single featured body, and each body `--mode full` prints. Not truncated to a first line: p90 is 8 lines and a first-line render would silently drop the rest.
* **Nothing at all on the index rows.** Not a first line, not a badge.

Measured on the live store:

| scope (entries) | digest | `--list` |
|---|---|---|
| `datapacket-talos` (37) | 7,273 B → 7,539 B (**+266**) | 3,660 B → 3,678 B (**+18**) |
| `civitai` (11) | 15,749 B → 15,993 B (**+244**) | 2,101 B → 2,119 B (**+18**) |
| `devrc` (9) | 14,208 B → 14,602 B (**+394**) | 1,920 B → 1,938 B (**+18**) |

The digest grew by **one body's worth**, and `--list` by a **flat +18 B on all three scopes regardless of entry count** — that is the footer sentence (`--ref` now advertises three sections) and nothing else. Across the whole store `## What it is` is 26 KB against `## Pointers`' 49 KB and `## Nuance`' 235 KB: the smallest of the three, ~8.6% of what an opt-in `--mode full` dump already prints.

**Pinned, not commented** — `TestWhatItIsIsBodyOnlyNotPerEntry`:

* `test_the_digest_prints_it_ONCE_however_many_entries_the_scope_holds` — builds scopes of **3 and 24** entries (two points on the dimension, per `claude/RULES.md`; 24 is deliberately not a multiple of the default limit) and asserts the occurrence count is `{3: 1, 24: 1}`. Asserting it merely "appears" would pass equally for a renderer printing it N times, which is the outcome the decision exists to prevent.
* `test_the_index_rows_carry_NO_what_it_is_at_all` — `--list` over 24 entries: 0 occurrences, and the rendered `listing_line`s carry no heading.
* `test_a_ref_lookup_prints_the_FULL_section_not_a_first_line` — a 3-line section, all 3 lines asserted present.
* `test_full_mode_prints_it_per_BODY_which_is_the_opt_in_dump` — states the N-copy cost of `--mode full` rather than leaving it implicit.

## The structural half: one tuple became two

`SURFACED_HEADINGS` was doing two jobs, and that is where the bug hid. "Does a reader print it" and "does a COUNT depend on it" were the same question, so a heading that feeds no count could only be added to both or to neither — and it was neither, for months.

* `SURFACED_HEADINGS = (WHAT_HEADING, POINTERS_HEADING, NUANCE_HEADING)` — the display set.
* `COUNTED_HEADINGS = (POINTERS_HEADING, NUANCE_HEADING)` — new. Drives `missing_sections`, `is_bare`, the `🔴 NO <heading>` index badge and the caveat clause explaining that badge. **Index rows and `--validate` findings are byte-for-byte unchanged.**

`subsystem_touch.SHAPE_HEADINGS` is re-pinned from `SURFACED_HEADINGS` to `COUNTED_HEADINGS`, with a **negative control** (`test_the_DISPLAY_set_is_WIDER_and_that_is_the_point`) asserting the two now genuinely differ — otherwise the pin would pass for the old reason and prove nothing about the split.

### A regression the change created, caught and covered

`is_bare` was `not any(self.sections.values())`. `sections` now carries `## What it is`, which the writer's own `new_entry_template` **pre-fills with a placeholder** — so read naively, every freshly created stub would report as filled-in and lose the "exists but has not been filled in" notice in exactly the case it was written for. `is_bare` is now bound to `COUNTED_HEADINGS`, with `test_a_what_it_is_stub_does_NOT_make_an_entry_look_filled_in` asserting the stub is present *and* the entry is still bare.

## Missing / empty degrades cleanly

Absent and present-but-empty are different on-disk states and both are exercised. Neither crashes; neither prints a stray heading. Both are **named** under the entry's own body (`(no `## What it is` content — this entry never says what the subsystem IS; re-derive it live)`), following the same "said, not left blank" rule the bare-entry notice already follows — and **never** as an index badge, because that badge means "a count on this row is not a measurement" and this section feeds no count. `service_recon`'s `index:` block prints the same notice. `TestWhatItIsDegradesCleanly` (4 tests) + `test_an_entry_with_NO_what_it_is_degrades_cleanly_and_SAYS_so`.

## Red → green

The new/changed assertions were run against a tree carrying the **new tests** and the **pre-change `scripts/lib/*` restored from `HEAD`** (verified byte-identical to `HEAD` with `cmp`, and the test files verified to differ from `HEAD`):

```
pre-change library, new tests   ->  15 failed, 6 passed
this branch,        same tests  ->  21 passed
```

The 6 that were already green are the ones that were true before (e.g. "a missing `## What it is` does not badge the index row"). The 15 reds include every core assertion: the section is surfaced, it is first, `IndexResult.what` exists, and both degrade cases.

`test_kills_the_what_it_is_INCLUSION` is the **inverse** of the mutation kill it replaces: it drops `WHAT_HEADING` back out of `SURFACED_HEADINGS` — the exact pre-change source — and asserts the content sentinel disappears **while the other two sections still render**, so the kill is a section-SELECTION kill and not "the renderer produced nothing". It asserts the content sentinel rather than the heading string, because the mutant still prints the degrade notice, which contains the heading; grepping the heading would have scored the mutant SURVIVED for a reason unrelated to surfacing.

## Docs corrected, not left stale

* `analyze-service/reference/index-store.md` — "prints one `index:` line plus the two surfaced sections" was the sentence the brief flagged; corrected, plus a new paragraph stating the body-vs-index-row decision. **No edit inside the `resolver-rules` / `entry-schema` hashed markers.**
* `analyze-service/SKILL.md` — step 2 asked the agent for a one-line "what it is" that the tool had never handed it. It now says the `index:` block's `## What it is` **is** that line.
* `resume/SKILL.md` — the digest's "exactly ONE entry in full (…)" parenthetical; the `🔴 NO <heading>` bullet now says which two headings it fires for. Also flags that the 2026-08-13 byte figures predate the scope's growth, with today's deltas beside them.
* `handoff/SKILL.md` — "the two headings any reader actually parses" was false the moment the reader parsed three; now "the two headings a COUNT depends on".

## Store safety

`~/.claude/analyze-service-index/` was treated as read-only throughout: **no git command was run inside it**, nothing was written, and no entry content appears in this repo, in any fixture, in the commit message or above. Every fixture is synthetic and built in a tmp dir. The live-store numbers here are counts and byte totals only.


ZacxDev:
Correction to this PR's description, from the post-merge adversarial audit.

The body says `## What it is` "was parsed by **nothing** on any read path". That claim is too wide, and this repo's own test refutes it: `subsystem_recall.search` (`-s/--search`) already surfaced the section before this change — `main` carried `test_subsystem_recall.py::test_search_covers_WHAT_IT_IS_which_the_digest_deliberately_hides` (line 2177 at `79b09ff`) asserting exactly that.

The enumeration in the same sentence — *not `--ref`, not the digest, not `service_recon`'s `index:` block* — is accurate, and is what this PR actually fixed. `subsystem_touch --validate`'s "🔴 `## What it is` is NOT checked: no reader parses it" carried the same overclaim, which is why it propagated.

Correcting here rather than editing the body, since the wrong version may already have been read. A follow-up PR narrows the same overclaim where it survives in a code comment (`service_recon.py` ~420-421) and in test docstrings.

PR #649, opened 2026-08-21
fix(hook-tests): the nudge suites ran against the OPERATOR'S REAL $HOME — and the live hook was writing into the state they asserted on

ZacxDev:
## The defect

`scripts/claude-hooks/tests/test_search_tool_nudge.py` computed
`HOME = os.path.expanduser("~")` at import and derived
`STATE_ROOT = $HOME/.cache/claude-search-tool-nudge/s` from it — the **operator's
real home**. It then globbed and `rmtree`'d under that path, and drove the hook as
a real subprocess once per scenario with that same real `$HOME` inherited, so
every child resolved the identical directory (`search-tool-nudge.py` builds its
own `CACHE_DIR` from `expanduser("~")` too).

Two consequences, both real:

* it **deleted** entries from the operator's own nudge state, and
* the **live** `search-tool-nudge` hook fires on every Bash call of every Claude
  Code session on the box — including the session running the gate — so an agent's
  own commands mutated the state the suite was asserting on, mid-run. The suite
  failed spuriously and the failure read as a code defect. Anyone running
  `scripts/gate.sh` from inside an active session hits this.

`test_shell_env_nudge.py` had the same shape: it created and `os.remove`d
`~/.cache/claude-shell-env-nudge/<sid>`, and its hook derives its whole
repo/kubeconfig map from `~` as well.

## The fix

Each suite makes its own throwaway `$HOME` with `tempfile.mkdtemp` and exports it
into `os.environ` **before anything reads `~`** — including the import of the hook
module. `os.path.expanduser("~")` is `os.environ["HOME"]` on POSIX (verified in
this session, not assumed: `posixpath.expanduser` returns `$HOME` when set and
only falls back to the passwd entry when it is absent), so that one assignment
redirects the suite's own constants and every inheriting child at once. Removed at
exit via `atexit`. An assertion right after the redirect fails loudly if it did
not take.

All eight spawn sites were checked individually — six in `test_search_tool_nudge.py`
(lines 190, 252, 618, 645, 662, 700), two in `test_shell_env_nudge.py`. Each either
passes no `env=` (inherits `os.environ`) or builds it as `dict(os.environ, …)`.

Both files keep their **script-suite** shape — no top-level `def test_`, so
`conftest.py`'s structural classifier and `run-tests.sh`'s `HOOK_TESTS` list still
agree and `test_hook_tests_dir_collects.py` stays green.

## The new guard

`scripts/tests/test_hook_suites_do_not_touch_the_inherited_home.py` (8 tests).

**REGRESSION** — runs each suite with `$HOME` pointed at a stand-in `tmp_path`
seeded with **decoys**, and asserts the tree is byte-for-byte unchanged, both
directions (created **and** removed). The decoys matter: both suites *delete*, and
a run that tidies up after itself leaves a create-only diff of zero. The decoys are
paths the pre-change suite provably removes, plus a bystander no cleanup glob names.
Their presence is asserted in the `before` snapshot first, so "unchanged" can never
be a comparison against an absent operand.

**POSITIVE CONTROL** — a suite that silently stopped driving the hook would leave
the inherited home clean too; that zero is indistinguishable from a harness wired to
nothing. Each suite now prints the file count it measured under its **own** temp
HOME, and the guard requires it non-zero, so the pair is always reported together.

**SEAM guard** — the isolation is a mutation of `os.environ`, so a spawn with a
hand-assembled `env=` would hand the real `$HOME` straight back while every
behavioural assertion stayed green. The suites' source is parsed with `ast` and any
such spawn fails. The scanner has its own must-flag / must-not-flag control, and the
suite ledger is pinned so coverage cannot shrink silently.

🔴 The guard never touches the operator's real home: it only ever points `HOME` at a
`tmp_path`. Nothing in it stats, creates or reads `~/.cache/*`.

## Verification

**Red at base, green at HEAD** (base = `3c54918`, measured by reverting only the two
suite files with `git checkout origin/main -- …` and re-running the guard):

| test | 3c54918 | HEAD |
|---|---|---|
| `…does_not_touch_the_home_it_inherits[test_search_tool_nudge.py]` | **FAIL** — REMOVED `.cache/claude-search-tool-nudge/s/test-session-search-nudge-DECOY`, `…/DECOY/content`, `.cache/claude-search-tool-nudge/test-search-nudge-DECOY-legacy` | PASS |
| `…does_not_touch_the_home_it_inherits[test_shell_env_nudge.py]` | **FAIL** — REMOVED `.cache/claude-shell-env-nudge/test-session-io-DO-NOT-COLLIDE` | PASS |
| `…does_write_state_under_its_own_temp_home[×2]` | **FAIL** — no isolation line printed | PASS |

An earlier create-only version of the same guard was also measured red at
`3c54918`, on 3 and 2 created entries respectively — so both directions have been
watched to fail.

**Positive-isolation pair** (printed by the guard):

```
test_search_tool_nudge.py: 69 files under the temp HOME, 0 changes to the inherited HOME
test_shell_env_nudge.py:    1 files under the temp HOME, 0 changes to the inherited HOME
```

**Suite size, before → after:** `test_search_tool_nudge.py` 174 → **175** checks
(floor `MIN_CHECKS = 160`); the one added check is the temp-HOME positive control.
`test_shell_env_nudge.py` has no counter; its check list grew by the same one.

## Sibling suites examined

| suite | verdict |
|---|---|
| `test_shell_env_nudge.py` | **same defect — fixed here** |
| `test_claude_notify.py` | already isolated (temp HOME + PATH stubs; its single spawn sets `env["HOME"]`) |
| `test_next_step_nudge.py` | already isolated (all three spawns set `HOME` to a tmp path) |
| `test_agent_ledger_hook.py` | already isolated (`e["HOME"] = str(home)` on both spawns; the ledger dir is derived from it) |
| `test_clawgate_writeback_guard.py` | already isolated (`monkeypatch.setenv("HOME", …)` plus `env["HOME"]` on all five spawns) |
| `test_bash_guard.py` | needs no redirect — `bash-guard.py` and `guard_core.py` contain **no** `makedirs`, no write-mode `open`, no `write_text`; they write nothing at all. Their `expanduser` calls are read-only path resolution of the command under test. |

## Gate

`nix develop … --command bash scripts/gate.sh --tier both --set all`, run on this
branch **rebased onto `29ea6f8`** (the merged tree, not the branch-as-authored):

```
FAIL  pytest  exit=1  verdict='RESULT: FAIL (exit=1)'
GATE: RESULT=FAIL exit=1
```

TOTAL collected=13823 passed=13821 skipped=1 **failed=1** (floor 13034).
Node tier: suites=4 files=33 tests=1119 pass=1119 fail=0 (floor 1098).

🔴 **The one failure is PRE-EXISTING on `main` and is not from this PR:**

```
scripts/tests/test_subsystem_recall.py::TestSkillDocsArePinned
    ::test_the_recall_step_comes_AFTER_the_handoff_is_read
ValueError: substring not found   (doc.index("**Read it fully.**"))
```

`da11569` (#643) rewrote that line in `claude/skills/resume/SKILL.md` to
`**Read it fully — but treat its "Open investigations" section as RECALL, not live
state.**`, and the pin still looks for the old literal. Attributed three ways:

1. `git diff --quiet origin/main HEAD -- claude/skills/resume/SKILL.md
   scripts/tests/test_subsystem_recall.py` → rc 0; both files also proven to
   **exist** at `origin/main` (`git cat-file -e`), so that is a real match and not
   a comparison against an absent operand.
2. Reproduced in a **pristine detached worktree at `29ea6f8`** with this branch
   entirely absent — same `ValueError`, same line.
3. An earlier full gate run of this branch, based on `3c54918` (before #643
   landed), was **`GATE_EXIT=0`, both tiers PASS**.

This is exactly the merged-tree shape RULES.md warns about: #643 was green on its
own branch and turns `main` red. It needs a one-line fix in either the SKILL.md or
the pin, and it belongs to #643 rather than here — mixing it in would blur this
PR's red/green matrix. **Do not merge this until `main` is unbroken.**

### DIFF
--- a/scripts/claude-hooks/tests/test_search_tool_nudge.py
+++ b/scripts/claude-hooks/tests/test_search_tool_nudge.py
@@ -16,11 +16,42 @@ reassuring answer here is a ZERO (no false positives):
 
 Run: python3 scripts/claude-hooks/tests/test_search_tool_nudge.py
 """
-import os, sys, glob, json, time, shutil, subprocess, tempfile, threading, importlib.util
+import os, sys, glob, json, time, shutil, subprocess, tempfile, threading, atexit, importlib.util
+
+# --------------------------------------------------------------------------- #
+# 🔴 HOME ISOLATION — must run BEFORE anything reads `~`, including the import
+# of the hook below.
+#
+# This suite GLOBS AND DELETES state directories, and it drives the hook as a
+# real subprocess a hundred-odd times. The hook resolves its CACHE_DIR from
+# `os.path.expanduser("~")`, which on POSIX is `os.environ["HOME"]` (verified:
+# posixpath.expanduser returns $HOME when set, and only falls back to the passwd
+# entry when it is absent). Pointed at the operator's real home that means:
+#
+#   * the suite deletes real `~/.cache/claude-search-tool-nudge/` entries, and
+#   * the LIVE search-tool-nudge hook — which fires on every Bash call of any
+#     Claude Code session running on this box, including the one running the
+#     gate — writes into the very state directory the suite is asserting on.
+#     Measured: the suite fails spuriously that way, and anyone running
+#     scripts/gate.sh from inside an active session hits it.
+#
+# So: one throwaway HOME, set before line-one of the constants below, exported
+# into `os.environ` so EVERY subprocess spawned from here inherits it. Every
+# spawn in this file either passes no `env=` (inherits `os.environ`) or builds
+# it as `dict(os.environ, …)`, so all of them carry it —
+# `scripts/tests/test_hook_suites_do_not_touch_the_inherited_home.py` pins that
+# structurally so a hand-built `env=` cannot quietly reintroduce the leak.
+# --------------------------------------------------------------------------- #
+TEST_HOME = tempfile.mkdtemp(prefix="search-tool-nudge-HOME-")
+os.environ["HOME"] = TEST_HOME
+atexit.register(shutil.rmtree, TEST_HOME, ignore_errors=True)
 
 HERE = os.path.dirname(os.path.abspath(__file__))
 HOOK = os.path.join(HERE, "..", "search-tool-nudge.py")
-HOME = os.path.expanduser("~")
+HOME = os.path.expanduser("~")  # == TEST_HOME; the hook resolves the same value
+assert HOME == TEST_HOME, (
+    "HOME redirection did not take effect (%r != %r) — every path below would "
+    "resolve into the operator's real home." % (HOME, TEST_HOME))
 
 spec = importlib.util.spec_from_file_location("search_tool_nudge", HOOK)
 assert spec and spec.loader
@@ -801,6 +832,20 @@ for i, bad_agent in enumerate((123, [], "../../etc", "a/b", "")):
     rc, out = run(payload)
     check("bad agent_id %r -> nudge unchanged" % (bad_agent,), (rc, bool(out)), (0, True))
 
+# --------------------------------------------------------------------------- #
+# 🔴 ISOLATION POSITIVE CONTROL — counted BEFORE the cleanup below wipes it.
+#
+# "the real home was untouched" is the same reassuring ZERO a suite that stopped
+# exercising the hook entirely would produce. So the pair is reported: N files
+# written under the throwaway HOME, 0 changes to the inherited one. The external
+# half lives in scripts/tests/test_hook_suites_do_not_touch_the_inherited_home.py
+# and PARSES the number printed at the bottom of this file, so a suite that
+# silently stopped writing state cannot pass it either.
+# --------------------------------------------------------------------------- #
+TEMP_HOME_FILES = sum(len(fs) for _, _, fs in os.walk(TEST_HOME))
+check("isolation positive control: the suite wrote state under the temp HOME",
+      TEMP_HOME_FILES > 0, True)
+
 clear_test_state()
 leaked = sorted(os.path.basename(p) for pref in TEST_SID_PREFIXES
                 for p in glob.glob(os.path.join(STATE_ROOT, pref + "*")))
@@ -824,3 +869,8 @@ print(f"all search-tool-nudge tests passed ({len(ran)} checks: "
       f"{len(MUST_FIRE)} must-fire, {len(BENIGN)} benign, "
       f"1 harness negative control, 1 benign-counter positive control, "
       f"{len(SESSIONS)} availability-suppression scenarios)")
+# Machine-readable, and PARSED by the external isolation guard — keep the wording.
+# This half is only the number this file measured itself; the other half of the
+# pair (writes to the INHERITED home, which must be 0) is measured from outside,
+# because a suite cannot be its own witness for a home it no longer looks at.
+print(f"isolation: {TEMP_HOME_FILES} files under the temp HOME")
--- a/scripts/claude-hooks/tests/test_shell_env_nudge.py
+++ b/scripts/claude-hooks/tests/test_shell_env_nudge.py
@@ -1,11 +1,25 @@
 #!/usr/bin/env python3
 """Unit tests for shell-env-nudge.analyze() + the hook's IO contract.
 Run: python3 scripts/claude-hooks/tests/test_shell_env_nudge.py"""
-import os, sys, json, subprocess, importlib.util
+import os, sys, json, shutil, subprocess, tempfile, atexit, importlib.util
+
+# 🔴 HOME ISOLATION — before anything reads `~`, including the hook import below.
+# Same defect as test_search_tool_nudge.py: this file DELETES a cache entry and
+# spawns the hook, and shell-env-nudge.py resolves both its repo/kubeconfig map
+# and its CACHE_DIR from `os.path.expanduser("~")` == $HOME on POSIX. Against the
+# real home that writes and removes files in `~/.cache/claude-shell-env-nudge/`.
+# Redirecting $HOME moves the hook's map AND the suite's expectations together —
+# they are built from the same constant — so the assertions are unchanged.
+TEST_HOME = tempfile.mkdtemp(prefix="shell-env-nudge-HOME-")
+os.environ["HOME"] = TEST_HOME
+atexit.register(shutil.rmtree, TEST_HOME, ignore_errors=True)
 
 HERE = os.path.dirname(os.path.abspath(__file__))
 HOOK = os.path.join(HERE, "..", "shell-env-nudge.py")
-HOME = os.path.expanduser("~")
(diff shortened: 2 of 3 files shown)


ZacxDev:
## Adversarial pre-merge audit — verdict: safe to merge (no 🔴)

### The defect is real, and worse than this PR's description says

- Both hooks are registered `PostToolUse` in the operator's live `~/.claude/settings.json`; `search-tool-nudge` fired on one of the auditor's own Bash calls **mid-audit**.
- During the audit, a **sibling agent's concurrent gate wrote and then removed 92 entries** in the real `~/.cache/claude-search-tool-nudge`.
- 5 concurrent runs of the **pre-change** suite sharing one `$HOME` → **5/5 FAIL**. 5 concurrent runs at **HEAD** with the real `$HOME` inherited → **5/5 PASS**, deterministic 69.

So this also removes a real cross-run collision that was making the gate flaky on this box.

### Verified rather than accepted

- **Isolation holds at syscall level.** `strace -e trace=%file` on the HEAD suite: **zero** syscalls naming `/home/zach/.cache/claude*`; the only `/home/zach` path touched at all is the repo. Same for the guard itself. A full base-guard run driving the *unfixed* destructive suites left the real nudge state byte-identical, 5826 → 5826.
- **Exhaustive enumeration of home resolution** across both suites and both hooks — every path goes through `os.path.expanduser("~")`; no `Path.home()`, no `pwd`/`getpwuid`, no cached `os.getenv("HOME")`, no hard-coded `/home/`.
- **The suites do not test less** — the `ran` check-name ledger was extracted from both revisions and diffed: **174 → 175, zero dropped**, exactly one added. Floor `MIN_CHECKS = 160` unchanged.
- **Red-at-base / green-at-HEAD re-measured independently**: same three `REMOVED` paths for search-tool-nudge, same one for shell-env-nudge. Positive-isolation pair reproduced verbatim (`69 / 0` and `1 / 0`).
- **Classification ledgers intact** — both suites keep zero top-level `def test_`, so conftest's structural classifier and `run-tests.sh`'s `HOOK_TESTS` still agree, and a mutant adding a top-level `def test_` is KILLED by `test_hook_tests_dir_collects.py`.
- **Mutation battery** under `PYTHONDONTWRITEBYTECODE=1`, every splice asserted landed, restored by `cp` with md5 verification: `_plant` loop emptied (text retained) → KILLED; scanner name-set drops `run` → KILLED by its own must-flag control; offender predicate inverted → KILLED; dropping the `removed` half of the two-way assertion at base → guard goes **GREEN while the suite deletes decoys**, proving that half is load-bearing.
- Merged-tree gate: pytest `collected=13839 passed=13838 skipped=1 failed=0` (floor 13034), node `1119/1119`, both tiers' own `RESULT:` lines agreeing.

### 🟡 Follow-ups — guard durability, not defects in shipped behaviour

**Y1 — the destructive half of the guard is walkable.** `scripts/tests/test_hook_suites_do_not_touch_the_inherited_home.py:82-93`. The deletion check rests entirely on hard-coded decoy strings that nothing ties to the suites' own constants. **Measured:** applying a hyphen-count-preserving rename of `TEST_SID_PREFIXES` (`test_search_tool_nudge.py:201`) to the **pre-change, still-destructive** suite made `test_the_suite_does_not_touch_the_home_it_inherits[test_search_tool_nudge.py]` **PASS**. A decoy that pre-creates the state dir without colliding with the cleanup glob also went green against the destructive suite. The deletion half therefore reverts to a create-only check the moment a session-id prefix is renamed or a new cleanup glob appears. Fix: `ast`-read `TEST_SID_PREFIXES` (and the shell suite's `sid`) and require a decoy per prefix, plus `set(DECOYS) == set(ISOLATED_SUITES)` with every entry non-empty.

**Y2 — `test_the_ledger_names_only_files_that_exist` is narrower than its docstring** (`:260-270`). The docstring claims "Positive control for the two tests above and the seam guard below … pinned as a set so the coverage cannot shrink unnoticed"; the body is `len(ISOLATED_SUITES) >= 2` plus per-entry `is_file()`. It pins neither set identity nor completeness against the directory, and demonstrates nothing about any test's ability to observe a violation. The actual positive controls are `test_the_suite_does_write_state_under_its_own_temp_home` and `test_the_env_scanner_can_actually_see_an_offender`.

**Y3 — the file's title asserts a universal property; the implementation covers 2 of 12 suites** (`:1` vs `:63-66`). The "already isolated" table for the other four is a point-in-time claim with no guard behind it. All 12 files in `scripts/claude-hooks/tests/` were re-measured under a decoy-seeded stand-in `$HOME`: every one is currently `created=0 removed=0`, and the sweep costs seconds — so widening `ISOLATED_SUITES`, or deriving it from the directory, is nearly free.

### 🟢 Nits

- **G1** `:285` `offenders = [e for e in envs if "os.environ" not in e]` is a *spelling* check: `env=dict(os.environ, HOME=<something else>)` passes while reintroducing the leak, and the two-statement idiom the guard file itself uses (`:108-109`) would be flagged. `os.execve` / `os.posix_spawn` / `multiprocessing` are outside the name set while `:275-282` says "a spawn with a hand-assembled env".
- **G2** `:57` `REPO_ROOT = Path(__file__).resolve().parents[2]` is the positional pattern that `conftest.py:71-91` deliberately replaced with a landmark walk after it broke in a copied tree.
- **G3** `atexit.register(shutil.rmtree, …)` does not run on SIGKILL — e.g. the 900 s `subprocess.run(timeout=)` kill — leaking one ~69-file temp `$HOME` per hung run. Measured zero `*-HOME-*` leftovers after ~20 normal runs; the 10 stale `/tmp/search-tool-nudge-tests-*` dirs on this box date from 2026-08-05 and pre-date this PR.
- **G4** `XDG_CACHE_HOME` is not scrubbed. Inert today (both hooks hardcode `$HOME/.cache`), but a hook that later honours XDG would reach the real cache with the guard still green.
- **G5** `_TIMEOUT = 900` carries no rationale unlike its twin in `conftest.py`. The guard re-runs both script suites, so the gate executes them twice — measured ≈1.8 s for the pair.
- **G6** `:256` `ast.get_source_segment(...) or ast.dump(kw.value)` — the `ast.dump` fallback renders `os.environ` as `Attribute(value=Name(id='os'…), attr='environ'…)`, lacking the literal `os.environ`, so it would false-flag. Unreachable for normally-parsed source.

### Could not verify

- The merged-tree gate evidence is against base `a381f7d8`; `main` has since advanced to `54ebf951`. Those commits touch `test_resume_state*`, `test_rules_size.py`, `test_subsystem_recall.py`, `claude/RULES.md`, `scripts/resume-state.sh` — **no overlap** with this PR's three files (confirmed independently before merge), and GitHub reports `MERGEABLE / CLEAN` — but a full gate was not re-run on `54ebf951 + #649`.
- Laptop-host behaviour was not exercised.
- Y1's failure mode is a future-drift hazard, demonstrated by mutating the suite under test rather than by observing it happen.

PR #651, opened 2026-08-21
fix(tests): main is RED — a docstring said "structural, not a phrase" while pinning the phrase closed

ZacxDev:
## `main` is RED right now, and this is the one-line fix

`test_the_recall_step_comes_AFTER_the_handoff_is_read` fails on `origin/main` (`29ea6f81`). Verified directly on a clean main worktree, not inferred: **1 failed / 49 passed**.

### Cause

#643 (`da115695`) appended a clause **inside the same bold span** of the resume skill's step 2:

```diff
- 2. **Read it fully.**
+ 2. **Read it fully — but treat its "Open investigations" section as RECALL, not live state.**
```

The test anchored on `doc.index("**Read it fully.**")` — closing `**` included. Every claim the test actually makes (`read_it < reconcile < step < report`) still holds; only the anchor broke. #643's wording change is good and is left alone.

The shape is worth naming: the test's own docstring said *"Structural, not a phrase"* while the implementation pinned the phrase. That's guard-narrower-than-its-description, inverted — here the **description was the honest half** and the code was narrower than it claimed.

### Fix

Anchor on the prefix `**Read it fully`, and record in the docstring **why the closing `**` is deliberately absent**, so nobody "tightens" it back. The subject of this test is the ORDERING, not the wording: the anchor must survive a continuation of that step and still vanish if the step does. Pinning the whole normalised line is the right move when the PROSE is the claim; it is the wrong move here, where it would fail on any edit to an unrelated clause.

### Verified as a pair, not a bare green

| step | result |
|---|---|
| before, on `main` | **1 failed** / 49 passed |
| after | **445 passed** (whole file) |
| negative control — reword step 2 away | the **same test fails again, by name** |
| restore | passes |

The negative control is the part that matters: it shows the anchor was loosened, not gutted.

### How it was found, and the pattern

By gating a **merged tree** for an unrelated PR — the branch was green and `main` was red, so a branch-only gate would have missed it entirely.

This is the **third** defect sitting on `main` today that no gate caught: #619's client-subdomain leak (fixed in #622), the branch-dependent guard-core tests (fixed in #623), and now this. All three have the same root cause — this repo has **no CI** (`<!-- merge-gate: none -->`) and the gate is run by hand, so it only runs when someone chooses to. `ZacxDev/homelab-infra#370` is the pipeline that would close that hole; it is open and blocked on a one-click GitHub App install.

### DIFF
--- a/scripts/tests/test_subsystem_recall.py
+++ b/scripts/tests/test_subsystem_recall.py
@@ -3580,9 +3580,21 @@ class TestSkillDocsArePinned:
 
     def test_the_recall_step_comes_AFTER_the_handoff_is_read(self) -> None:
         """Structural, not a phrase: recall is context for a doc already read,
-        not a substitute for reading it."""
+        not a substitute for reading it.
+
+        🔴 THE ANCHOR IS A PREFIX, AND THE CLOSING `**` IS DELIBERATELY NOT IN
+        IT. This docstring said "structural, not a phrase" while `.index()`
+        pinned the phrase `**Read it fully.**` closed — so `#643` appending a
+        clause inside the same bold span (`**Read it fully — but treat …**`)
+        reddened `main` for a reword that preserved every claim this test makes.
+        The subject here is the ORDERING, not the wording, so the anchor must
+        survive a continuation of that step and still vanish if the step does.
+        Pinning the whole normalised line is the right move when the PROSE is
+        the claim; it is the wrong move here, where it would fail on any edit to
+        an unrelated clause.
+        """
         doc = RESUME_DOC.read_text(encoding="utf-8")
-        read_it = doc.index("**Read it fully.**")
+        read_it = doc.index("**Read it fully")
         # The INVOCATION, not the bare filename: `resume-state.sh` is also named
         # in step 1's prose, so `.index()` on the bare name finds a point BEFORE
         # the handoff is read and the ordering claim inverts.


ZacxDev:
Superseded by #650, which fixed this concurrently and **better** — closing rather than merging a redundant change.

Both land on the same anchor (`doc.index("**Read it fully")`), but #650 adds a guard I did not:

```python
assert doc.count("**Read it fully") == 1, (
```

That pins the prefix as *unambiguous*, so the anchor cannot silently bind to a second occurrence if the phrase ever appears twice — the `count=1`-on-a-repeated-pattern hazard, applied to an index anchor. My version would have been satisfied by the first match either way.

Verified before closing: `origin/main` at `a381f7d8` runs `test_subsystem_recall.py` **450 passed**, so the red I opened this for is genuinely gone rather than merely reassigned.

Worth keeping from the diagnosis, since #650's framing is narrower: the shape here was a docstring that said *"Structural, not a phrase"* while the implementation pinned the phrase **closed** — guard-narrower-than-its-description, inverted, where the description was the honest half. And it was found only by gating a **merged tree** for an unrelated PR; the branch was green and `main` was the broken side, so a branch-only gate could not have seen it.

ZacxDev:
Superseded — closing as redundant, not as wrong.

`main` was genuinely red at `test_the_recall_step_comes_AFTER_the_handoff_is_read` (I reproduced it on a pristine detached `origin/main`: `ValueError: substring not found`), and three sessions independently found it. A fix landed first in #650 (merged `a381f7d8`), so this diff now removes a line that no longer exists on `main` and would conflict rather than merge.

What is on `main` now: the anchor is the opening imperative rather than the full sentence, `doc.count(...) == 1` is asserted so `.index()` cannot silently move to an earlier duplicate, and the order pinned is `reconcile < read_it < step < report`. Verified green post-merge.

If the anchor you chose here is more rot-resistant than the one that landed, that is worth having — please reopen or open a fresh PR against current `main` rather than treating this close as a verdict on the approach. The underlying lesson is the same one all three of us hit: a pin whose docstring says "structural, not a phrase" while calling `.index()` on a phrase will break on any legitimate reword.

PR #652, opened 2026-08-21
fix(tests): `main` is RED — a pin whose docstring said "structural, not a phrase" was a phrase

ZacxDev:
fix(tests): `main` is RED — a pin whose docstring said "structural, not a phrase" was a phrase

#643 reworded step 2 of `claude/skills/resume/SKILL.md`:

    -2. **Read it fully.**
    +2. **Read it fully — but treat its "Open investigations" section as RECALL, not live state.**

That moved the closing `**`, and `test_the_recall_step_comes_AFTER_the_handoff_is_read`
did `doc.index("**Read it fully.**")` on the exact old string. Every gate run on
`main` has been red since — verified on a pristine detached worktree at 29ea6f8:
1 failed, 444 passed. A permanently-red gate trains everyone to click through, so
this is not a cosmetic fix.

The defect is the one `claude/RULES.md` names: the guard's own docstring says
"Structural, not a phrase" while the implementation was a phrase, walkable by a
pure reword. Patching the literal would have rebuilt the same trap, so:

* the read-step anchor is a PREFIX regex on the numbered-list marker
  (`^\d+\.\s+\*\*Read it fully`), which survives any continuation of that
  sentence inside the bold. The other three anchors are literal commands the
  skill runs, which is genuinely structural.
* every anchor goes through `_at()`, which fails with WHICH anchor, WHICH file,
  that the doc was almost certainly reworded, and "do NOT delete the assertion".
  `str.index` raised a bare `ValueError: substring not found` naming none of
  that — which is why a red `main` was easy to look past.
* the ordering assertion carries the four offsets in its message.

Both assertions were watched fire, each with its own message:

* pure reword (`Read`->`Peruse`) -> the anchor message, `assert -1 != -1`
* `**Report**` hoisted to offset 0 -> the ordering message, `assert 5163 < 0`

Matrix: red at 29ea6f8 (1 failed, 444 passed), green at HEAD (445 passed).
Full gate on the branch: `GATE: RESULT=PASS exit=0` — pytest 13814 passed /
0 failed, node 1119/1119, both tiers' own RESULT lines agreeing.

`re` is newly imported; it had no prior use in this module.

### DIFF
--- a/scripts/tests/test_subsystem_recall.py
+++ b/scripts/tests/test_subsystem_recall.py
@@ -43,6 +43,7 @@ import collections
 import hashlib
 import importlib.util
 import json
+import re
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


ZacxDev:
Closing as **superseded by #650**, which fixed the same breakage concurrently and landed first.

I diagnosed the red `main` independently (verified on a pristine detached worktree at `29ea6f8`: 1 failed / 444 passed) and wrote this fix without first checking whether an open PR already covered it. That was the avoidable mistake — `main` moves every few minutes in this repo, so "is it broken?" and "is someone already fixing it?" are two different questions and I only asked the first.

#650's version is also the better of the two on the point that matters: it asserts `doc.count("**Read it fully") == 1`, so the anchor cannot silently become ambiguous and invert the ordering claim. Mine had no uniqueness check. Verified green on `a381f7d8`: 450 passed.

**One thing from this branch that #650 does not have**, recorded here rather than reopened as churn on a file that has now been edited three times today:

The other three anchors still use bare `doc.index(...)`, which raises `ValueError: substring not found` naming no file, no anchor, and no remedy. That failure mode is a contributing cause of the original incident — a red `main` was easy to look past because the traceback said nothing actionable. This branch routed every anchor through a helper that fails with which anchor, which file, "the doc was almost certainly REWORDED, not broken", and "do NOT delete the assertion".

Worth folding in if anyone touches this test again; not worth a PR of its own.

For the record, both assertions on this branch were watched fire with their own messages before it was closed:

- a pure reword (`Read` → `Peruse`) → the anchor message, `assert -1 != -1`
- `**Report**` hoisted to offset 0 → the ordering message, `assert 5163 < 0`

Full gate on the branch was `GATE: RESULT=PASS exit=0` (pytest 13814 passed / 0 failed, node 1119/1119).

ZacxDev:
Superseded — closing as redundant, not as wrong.

`main` was genuinely red at `test_the_recall_step_comes_AFTER_the_handoff_is_read` (I reproduced it on a pristine detached `origin/main`: `ValueError: substring not found`), and three sessions independently found it. A fix landed first in #650 (merged `a381f7d8`), so this diff now removes a line that no longer exists on `main` and would conflict rather than merge.

What is on `main` now: the anchor is the opening imperative rather than the full sentence, `doc.count(...) == 1` is asserted so `.index()` cannot silently move to an earlier duplicate, and the order pinned is `reconcile < read_it < step < report`. Verified green post-merge.

If the anchor you chose here is more rot-resistant than the one that landed, that is worth having — please reopen or open a fresh PR against current `main` rather than treating this close as a verdict on the approach. The underlying lesson is the same one all three of us hit: a pin whose docstring says "structural, not a phrase" while calling `.index()` on a phrase will break on any legitimate reword.

PR #656, opened 2026-08-21
fix(gates): the nudge suite's state root was SHARED with every concurrent run — two false reds landed on an unrelated PR

ZacxDev:
## The defect, re-measured before changing anything

`scripts/claude-hooks/search-tool-nudge.py:42` hardcoded

```python
CACHE_DIR = f"{HOME}/.cache/claude-search-tool-nudge"
```

a fixed path **shared by every process on the box**, and its suite is a *concurrency*
suite: it asserts "exactly one nudge across 12 parallel invocations" for a session id
it owns and "no test state leaks into the next run" for a prefix it cleans. Neither
claim can be true of a directory a sibling gate run is writing.

Reproduced deterministically at `54ebf95` before touching anything — two hook runs,
same session id, **different** `SEARCH_TOOL_NUDGE_CACHE_DIR` values, one scratch HOME:

```
run1 (root1)    : rc=0 nudged=True
run2 (root2)    : rc=0 nudged=False      <- silenced by the other run's state
control same-root: first=False second=False
  state file: home/.cache/claude-search-tool-nudge/s/repro-shared-sid/content
```

Everything lands in `$HOME` and the variable is ignored entirely. The shared root
currently holds **1,955 live session directories**, actively written while gates run.

## The fix — GUARD 8's shape, applied to GUARD 9

1. **The hook** reads `SEARCH_TOOL_NUDGE_CACHE_DIR` (naming follows the existing
   `SEARCH_TOOL_NUDGE_MAX_SCAN_BYTES`), parsed defensively at import time outside
   `main()`'s try, absolute-only — a relative root would follow whatever cwd the Bash
   call had and break the dedupe from the other side. Unset → today's path, unchanged.
2. **`scripts/run-tests.sh` GUARD 9** exports a per-run temp dir, unconditionally, on
   the one line every target goes through — the suite that needs it is a `HOOK_TEST`,
   which no conftest can ever reach. It **arms itself**: it asks the hook where its
   state root resolved to and exits 2 if the answer is outside this run's dir.
3. **The hook's suite** takes `STATE_ROOT` from the hook instead of recomputing it,
   and self-isolates into its own mkdtemp when run by hand.

## The seam, pinned — not the two sides

`scripts/tests/test_search_nudge_state_isolation.py`. Both halves are perfectly
testable in isolation while the pair does nothing, so what is asserted is the
relationship: the runner exports the name the hook reads; a **real** `run-tests.sh`
run lands its state in its own dir and not in `$HOME`; the hook's own suite writes to
the root it is handed. The structural name pin reads the name from the hook's own
constant, never from a literal on both sides.

**A fourth file was in the seam and was not obvious.** `test_on_disk_artifact_names.py`
pins these paths against a throwaway `$HOME` — GUARD 9's export would have overruled
its fixture and every whole-path assertion would have compared an empty set with an
empty set. Its fixture now clears the variable, which makes that test the standing
proof that the **unset** default did not move.

## Red-at-base matrix

The new test file, unchanged, run against a tree whose hook, runner and hook-suite
come from `54ebf95` (each extracted with `git show`, byte counts asserted non-zero):

| | at `54ebf95` | at HEAD |
|---|---|---|
| `test_two_runs_with_different_roots_do_not_see_each_others_state` | **FAILED** | passed |
| `test_a_real_runner_run_lands_the_state_in_its_own_dir_not_in_HOME` | **FAILED** | passed |
| `test_the_hooks_own_suite_writes_to_the_root_it_is_handed` | **FAILED** | passed |
| `test_the_runner_exports_the_name_the_hook_reads` | **FAILED** | passed |
| `test_the_hook_exposes_its_cache_root_variable_name` | **FAILED** | passed |
| `test_the_runner_refuses_when_the_hook_stops_reading_the_variable` | FAILED (anchor absent — **SKIPPED, not scored**) | passed |
| `test_the_control_of_that_mutant_is_green` | FAILED (same) | passed |
| `test_two_runs_sharing_one_root_still_dedupe` | passed (the control) | passed |
| `test_the_default_is_unchanged_when_the_variable_is_unset` | passed | passed |
| `test_a_relative_root_is_refused` | passed **vacuously** | passed |
| `test_junk_in_the_variable_never_breaks_the_hook` | passed **vacuously** | passed |

`7 failed, 4 passed` at base; `11 passed` at HEAD. The last two are invariant guards
over code that did not exist at base — labelled as such in the file, not counted as
regression coverage.

## Mutation sweep

9 mutants + a no-op control, each in a shallow symlink copy of the repo (the real
checkout is never written), under `PYTHONDONTWRITEBYTECODE=1`, anchors asserted unique
so a missing anchor reports SKIPPED rather than SURVIVED.

|---|---|---|
| no-op (negative control) | **SURVIVED** | — |
| M1 hook ignores the override (positive control) | KILLED | the 3 behavioural seam tests + the hook suite's own root check |
| M2 hook drops the `isabs` check | KILLED | `test_a_relative_root_is_refused` |
| M3 hook default root renamed | KILLED | `test_the_default_is_unchanged…`, `test_the_search_tool_nudge_joins_session_and_agent_with_an_AT_SIGN` |
| M4 hook reads a different variable name | KILLED | `test_the_runner_exports_the_name_the_hook_reads` + both behavioural tests |
| M5 runner assigns but does not `export` | KILLED | structural + `test_a_real_runner_run_lands…` |
| M6 runner exports the real `$HOME` root | KILLED | `test_a_real_runner_run_lands…` (GUARD 9 refuses, exit 2) |
| M7 runner's arming check accepts anything | KILLED | `test_the_runner_refuses_when_the_hook_stops_reading_the_variable` |
| M8 hook suite goes back to `$HOME` | KILLED | `test_the_hooks_own_suite_writes_to_the_root_it_is_handed` |
| M9 names registry stops clearing the override | KILLED | `test_the_search_tool_nudge_joins_session_and_agent_with_an_AT_SIGN` |

🔴 **M6 and M7 first reported SURVIVED, and that was a defect in my sweep harness, not
in the guards.** `runner_patch.RUN_TESTS` is `Path(__file__).resolve().parents[2]/…`
and `.resolve()` **follows symlinks** — with a symlinked `testlib`, every
runner-driving test executed the REAL, unmutated `run-tests.sh`. `scripts/tests/
conftest.py` had the same shape one level up: it `sys.path.insert`s its own resolved
parent and imports `testlib` at collection time, so the real package was in
`sys.modules` before the test module's own insert ran. Both were made real files and
both mutants died. A third false SURVIVED (M9) came from running the sweep in a shell
where the variable was never set at all, so the fixture that clears it was
unreachable; the sweep now reproduces the gate's environment.

## Production default: unchanged

Differential between the base hook and this one, variable **unset**, same scratch
`$HOME`, five payloads (content, its dedupe, files, benign, non-search):

```
stdout/rc identical  : True
state paths identical: True
paths: ['.cache/claude-search-tool-nudge/s/default-diff-sid/content',
        '.cache/claude-search-tool-nudge/s/default-diff-sid/files']
```

## Gate

`nix develop --command bash scripts/gate.sh --tier both --set hermetic`

```
RESULT: PASS (exit=0)   pytest  TOTAL collected=13859 passed=13858 skipped=1 failed=0
                                (floor: 13034 = sum of 25 per-target floors)
RESULT: PASS (exit=0)   node    TOTAL suites=4 files=33 tests=1119 pass=1119 fail=0
GATE: RESULT=PASS exit=0
```

GUARD 9 reported in that run:

```
run-tests: search-tool-nudge state isolated for this run (GUARD 9)
  SEARCH_TOOL_NUDGE_CACHE_DIR=/tmp/nix-shell.cBWW4U/tmp.TUrMo7ncZ3/search-tool-nudge
  hook state root    =/tmp/nix-shell.cBWW4U/tmp.TUrMo7ncZ3/search-tool-nudge/s
```

## Honest scope

GUARD 9 closes the shared-**root** hazard. It does **not** trap a test that takes the
variable *away* (`env -u`, a hand-built subprocess env) — that drops to the `$HOME`
default, and moving `HOME` for the whole run is a far larger blast radius than the
hazard. That limit is written into GUARD 9's header rather than left implied.

### DIFF
--- a/scripts/claude-hooks/search-tool-nudge.py
+++ b/scripts/claude-hooks/search-tool-nudge.py
@@ -39,7 +39,46 @@ Fail-open: any error -> exit 0 silently; it must never break the Bash tool.
 import sys, json, os, re, shlex, glob
 
 HOME = os.path.expanduser("~")
-CACHE_DIR = f"{HOME}/.cache/claude-search-tool-nudge"
+
+# 🔴 THE CACHE ROOT IS OVERRIDABLE, AND THAT IS A TEST-ISOLATION GUARD, NOT A
+# CONVENIENCE. A fixed path under $HOME is SHARED BY EVERY CONCURRENT PROCESS ON THE
+# BOX, and this hook's own suite asserts things that are only true of a root nobody
+# else is writing: "exactly one nudge across 12 parallel invocations" and "no test
+# state leaks into the next run". Measured 2026-08-20, with three to four gate runs
+# in flight at once, both went red inside the gate runner's OWN positive controls —
+#   concurrency: exactly one nudge across 12 parallel invocations: got 0 want 1
+#   no test state leaks into the next run: got ['test-search-nudge-nul-byte-…'] want []
+# — because a sibling run had already claimed the same session id, or created its
+# state dir between this run's cleanup and its leak check. Neither failure was about
+# the tree under test; both were reported as a red gate against whatever PR happened
+# to be in flight, twice.
+#
+# So: `scripts/run-tests.sh` (GUARD 9) points this at a per-run temp dir, exactly as
+# GUARD 8 does for ACTIVITY_SPOOL_DIR, and the hook must READ that variable or the
+# export governs nothing. The seam — runner exports it, hook honours it, and both
+# spell it the same — is pinned by `scripts/tests/test_search_nudge_state_isolation.py`.
+#
+# 🔴 Parsed DEFENSIVELY, at import time, outside main()'s try — the same constraint
+# `_scan_cap` below carries: an exception here escapes as a non-zero exit with a
+# traceback on EVERY Bash call in the session, which the module docstring promises
+# never happens. A RELATIVE value is rejected rather than honoured: a hook process
+# inherits whatever cwd the Bash call ran in, so a relative root would scatter state
+# across directories and silently break the dedupe this hook is built on.
+CACHE_DIR_ENV = "SEARCH_TOOL_NUDGE_CACHE_DIR"
+
+
+def _cache_dir(default):
+    try:
+        raw = os.environ.get(CACHE_DIR_ENV)
+        return raw if raw and os.path.isabs(raw) else default
+    except Exception:
+        return default
+
+
+# With the variable unset this is byte-identical to what it has always been — the
+# production path on both hosts, and the one the on-disk artifact-name registry
+# (`tests/test_on_disk_artifact_names.py`) pins against a throwaway $HOME.
+CACHE_DIR = _cache_dir(f"{HOME}/.cache/claude-search-tool-nudge")
 
 # Content-search binaries: they read file *contents*, so the native answer is Grep.
 GREP_BINS = {"grep", "egrep", "fgrep", "rgrep", "ggrep"}
--- a/scripts/claude-hooks/tests/test_on_disk_artifact_names.py
+++ b/scripts/claude-hooks/tests/test_on_disk_artifact_names.py
@@ -131,6 +131,18 @@ def home(tmp_path, monkeypatch):
     h = tmp_path / "home"
     (h / ".claude").mkdir(parents=True)
     monkeypatch.setenv("HOME", str(h))
+    # 🔴 $HOME must be the ONLY thing deciding these paths, or the walk below finds
+    # nothing and every whole-path pin in this file passes on an empty set == an
+    # empty set. `search-tool-nudge.py` takes an env override for its cache root
+    # (`SEARCH_TOOL_NUDGE_CACHE_DIR`) and `scripts/run-tests.sh` GUARD 9 exports one
+    # for every target, so under the gate that override — not this fixture — would
+    # decide where the state landed.
+    #
+    # Removing it is also the PRODUCTION-DEFAULT proof: the assertions below are the
+    # only place that exercises the unset-variable path, which is the path both hosts
+    # actually run. `test_the_search_tool_nudge_joins_session_and_agent_with_an_AT_SIGN`
+    # is red if that default ever stops resolving to `$HOME/.cache/…`.
+    monkeypatch.delenv("SEARCH_TOOL_NUDGE_CACHE_DIR", raising=False)
     return h
 
 
--- a/scripts/claude-hooks/tests/test_search_tool_nudge.py
+++ b/scripts/claude-hooks/tests/test_search_tool_nudge.py
@@ -22,6 +22,34 @@ HERE = os.path.dirname(os.path.abspath(__file__))
 HOOK = os.path.join(HERE, "..", "search-tool-nudge.py")
 HOME = os.path.expanduser("~")
 
+# --------------------------------------------------------------------------- #
+# 🔴 STATE ISOLATION — BEFORE the hook is imported, because it resolves its cache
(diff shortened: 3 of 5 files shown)


ZacxDev:
Superseded by #649, which fixed the same hazard from the other direction and landed first. Closing rather than merging — the two mechanisms are **incompatible**, not merely redundant.

## The conflict

Both PRs fix the same real defect: the nudge suites shared state at `$HOME/.cache/claude-search-tool-nudge`, so concurrent gate runs contaminated each other. That was measured — it produced two false-red full-gate runs today with different symptoms (`concurrency: got 0 want 1`, then `no test state leaks: got [<sentinel>]`), each nested inside the gate runner's own meta-tests and each blaming an unrelated PR.

The mechanisms differ:

| | isolation by |
|---|---|
| **#649** (landed) | relocating **`HOME`** — a throwaway HOME exported before line one of the constants, so `CACHE_DIR` resolves under it |
| **#656** (this) | an env override, `SEARCH_TOOL_NUDGE_CACHE_DIR`, plus GUARD 9 in the runner |

#649 added an `isolation positive control` asserting state lands **under the temp HOME**. #656 redirects state *away* from HOME by design. So the merge is textually clean and semantically contradictory — the control fails on the merged tree by construction.

## Measured, not assumed

- `main` alone: **`all search-tool-nudge tests passed (175 checks…)`, `isolation: 69 files under the temp HOME`** — the hazard is genuinely closed without this PR.
- `main` + this branch, that suite standalone: **`FAIL: isolation positive control: the suite wrote state under the temp HOME: got False want True`**.
- Reproduced with #656 **alone** on current main, so it is not an interaction with the other PR in flight.
- The merge-base is `54ebf951`; `d990c9dc` (#649) landed after it, which is why this branch's own gate was green on its tree and red on the merge.

## Why #649 is the better mechanism

Relocating `HOME` covers **every** consumer of `~` in the suite, including any not yet written. The env-override approach covers only the paths that were taught to read the variable — and this PR's own GUARD 9 header conceded exactly that limit, noting it cannot trap a test that takes the variable away with `env -u`. It called moving `HOME` "a far larger blast radius than the hazard"; that judgement was reasonable in isolation and `main` had already made the opposite call, deliberately, in #649.

## Worth keeping from the work

Three findings survive the close and are recorded here so they are not lost with the branch:

1. **The sweep harness was broken before the guards were.** Three mutants first reported SURVIVED because `runner_patch.RUN_TESTS` uses `Path(__file__).resolve()`, which **follows symlinks** — with a symlinked `testlib`, every runner-driving test executed the real, unmutated `run-tests.sh`. `scripts/tests/conftest.py` has the same shape one level up. A green sweep there certifies nothing, and this is not specific to this PR.
2. **A fourth file sat in the seam:** `test_on_disk_artifact_names.py` pins these paths against a throwaway `$HOME`; an exported override would have overruled its fixture and turned every whole-path assertion into an empty-set-vs-empty-set comparison — a vacuous green.
3. The gate runner's own meta-tests (`test_a_green_real_run_says_pass_with_exit_zero`, `test_the_collapse_guard_is_what_makes_that_test_red`) invoke the **full** runner, so the gate's positive controls inherit any contamination in any script target. That is what turned a shared-state bug into a whole-run FAIL attributed to whichever PR was in flight. #649 fixes the cause; the amplification path remains.

PR #657, opened 2026-08-21
fix(signal): the D3 send path has never worked on real Postgres (COALESCE uuid/text)

ZacxDev:
## The bug

`_draft_or_raise()` built `COALESCE(d.signal_uuid, d.phone_number)`. `contacts.signal_uuid` is `uuid`, `phone_number` is `text`, and **Postgres type-checks the whole `CASE` regardless of which branch would execute** — so it raises `DatatypeMismatch` for *every* draft. `approve()`, `send()` and `reconcile()` all reach `_draft_or_raise()`.

```
ERROR:  COALESCE types uuid and text cannot be matched
LINE 3:   ELSE COALESCE(d.signal_uuid, d.phone_number)
```

**Control:** the outbound population is `pending: 1` and nothing else — no draft has ever reached `approved`/`sending`/`sent`. Found by trying to approve the first real one.

## Why 599 green tests missed it

The hermetic substrate is **SQLite** — its own docstring calls it *"a REAL relational engine standing in for Postgres."* SQLite is dynamically typed and accepts the uncast form. The suite is **structurally blind** to Postgres static type errors in hand-written SQL.

## 🔴 The first fix was wrong, and the suite caught it

`signal_uuid::text` type-checks on Postgres but is **unparseable by SQLite**: it took the signal suite from `1 failed / 598 passed` to `54 failed / 547 passed`. Shipped the ANSI `CAST(signal_uuid AS text)` instead, which both engines accept.

That is the shape worth remembering — the cast has to satisfy *both* gates, or fixing one blinds the other.

## Evidence

| | |
|---|---|
| pre-fix SQL, live Postgres | `ERROR: COALESCE types uuid and text cannot be matched` |
| post-fix SQL, live Postgres | `51 \| pending \| group.elgzaTRRT2Nk…` (1 row) |
| backstop at `origin/main` | **red**, with its own message naming `COALESCE(d.signal_uuid, d.phone_number)` |
| signal suite at HEAD | `1 failed / 601 passed / 1 skipped` |
| signal suite at base | `1 failed / 598 passed` — same pre-existing failure |

## The two new guards are deliberately different in kind

- `test_draft_query_runs_on_real_postgres` — the **real** one. Executes the SQL against a live server; **skips without `SIGNAL_PG_DSN`** so it is honest about being unavailable rather than passing vacuously.
- `test_signal_uuid_is_never_coalesced_without_a_cast` — a **SPELLED** backstop, labelled as such in the source. It reads source text, so it is walkable by rewriting the expression in another shape or picking a different column pair. It exists because the real guard cannot run in the default gate, **not** because it is sufficient. It carries a positive control asserting it flags the known-bad expression and neither good spelling.

## NOT verified

- `nix build .#checks.x86_64-linux.pytests` was **not** run — counts are from an ad-hoc `nix-shell`. That tier decides which tests execute.
- The real-Postgres test has never run in CI; it was exercised by hand via `psql` against the live database, which is where the evidence above comes from.
- No other hand-written SQL in this module was audited for the same class. This fixes the one expression that blocks the send path; **a Postgres-backed tier is the durable fix** and is not in this PR.

### DIFF
--- a/scripts/signal/_signal_db.py
+++ b/scripts/signal/_signal_db.py
@@ -1292,8 +1292,21 @@ class SignalDB:
                 """
                 SELECT m.id, m.message_timestamp, m.body, m.send_state,
                        m.approval_ref, m.source_contact_id, m.dest_contact_id,
+                       -- `signal_uuid` is uuid, `phone_number` is text, and
+                       -- Postgres type-checks the WHOLE CASE regardless of which
+                       -- branch would run — so an uncast COALESCE here raises
+                       -- DatatypeMismatch for EVERY draft, breaking approve(),
+                       -- send() and reconcile() alike (they all reach this).
+                       -- The hermetic suite cannot see it: its substrate is
+                       -- SQLite, which is dynamically typed and accepts the
+                       -- uncast form.
+                       -- 🔴 Use ANSI `CAST(... AS text)`, NOT Postgres's `::`.
+                       -- SQLite cannot parse `::`, so the pg-only spelling
+                       -- turns 53 hermetic tests red — measured, not guessed.
+                       -- The cast must satisfy BOTH engines or one of the two
+                       -- gates goes blind.
                        CASE WHEN d.is_placeholder THEN d.phone_number
-                            ELSE COALESCE(d.signal_uuid, d.phone_number)
+                            ELSE COALESCE(CAST(d.signal_uuid AS text), d.phone_number)
                        END AS recipient
                 FROM signal.messages m
                 LEFT JOIN signal.contacts d ON d.id = m.dest_contact_id
--- a/scripts/signal/_signal_db.py
+++ b/scripts/signal/_signal_db.py
@@ -1433,10 +1433,20 @@ class SignalDB:
         errors = [entry["errors"] for entry in entries if entry.get("errors")]
         errors += [entry["error"] for entry in entries if entry.get("error")]
         if errors:
+            # 🔴 Carry the server timestamp into the error. A partly-failed GROUP
+            # send DID go out to the members that succeeded, and the reply
+            # carried the timestamp `reconcile --sent --timestamp` needs. Without
+            # it here the value is discarded and the operator has to hunt it in
+            # the Signal thread — which is exactly the position draft 51 left
+            # its operator in.
+            ts_hint = entries[0].get("timestamp")
             raise RuntimeError(
                 f"the Signal API reported per-recipient errors for draft "
                 f"{draft_id!r}: {errors!r}; the draft stays in {STATE_SENDING!r} "
                 f"for manual reconciliation — see `reconcile`"
+                + (f". The response carried timestamp {ts_hint!r}; if the message "
+                   f"DID reach the thread, reconcile with "
+                   f"`--sent --timestamp {ts_hint}`" if ts_hint else "")
             )
         server_ts = _server_timestamp(entries[0])
         # The terminal update carries the SAME state predicate as the claim. This
--- a/scripts/signal/_signal_db.py
+++ b/scripts/signal/_signal_db.py
@@ -387,6 +387,53 @@ def spend_authorization(auth: object) -> None:
 TYPE_DELETED = "deleted"
 
 
+def _normalize_send_response(result) -> list:
+    """The `/v2/send` reply as a list of per-recipient dicts.
+
+    🔴 MEASURED, not assumed. Against signal-cli-rest-api in **json-rpc** mode
+    the success reply is a LIST, one entry per recipient::
+
+        201  [{"timestamp":"1787331796630"}]
+
+    A bare dict is also accepted, because upstream documents
+    `ds.SendMessageResponse` as an object and other modes/versions return that
+    shape. Anything else raises, deliberately: the caller has already COMMITTED
+    `send_state=sending` before the POST, so a surprise here leaves a draft a
+    human reconciles — never a silent duplicate send.
+
+    Exactly one entry is required. Every caller sends to a single recipient
+    (`recipients: [recipient]`), so a longer list means the wire contract
+    changed and picking an entry would be a guess about which message the stored
+    timestamp belongs to — which is precisely what breaks sync-echo dedupe.
+    """
+    if isinstance(result, dict):
+        entries = [result]
+    elif isinstance(result, list):
+        entries = result
+    else:
+        raise ValueError(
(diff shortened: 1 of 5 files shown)


ZacxDev:
## Post-audit corrections and fixes

A blind adversarial audit found two mutants that **survived the whole suite** and three claims in the body above that **do not reproduce**. Correcting them here rather than editing the description, since they may already have been read.

### 🔴 Corrections to my own claims

**1. "11 of 12 new tests red at base" was misleading — and vacuous in the way this repo's own rules warn about.** The PR adds **17** tests, not 12. At base **13 fail / 3 pass / 1 skip** — but all 13 fail with `AttributeError: module has no attribute '_normalize_send_response'`, i.e. **red because the function is new, not because they reproduce the defect**. Only `test_the_old_dict_only_expression_would_have_failed_on_the_live_shape` is a true positive control of the bug, and it passes at base by design. A test that goes red only because a symbol is missing pins nothing.

The genuinely defect-reproducing coverage is what the fix round below adds: four tests that drive `send_approved()` with the live LIST shape.

**2. Base suite counts were wrong.** I reported `1 failed / 598 passed`, attributing a pre-existing failure to `test_image_deps.py`. Independently measured at `b9e27a89`: **0 failed / 599 passed** — that test passes (11/11) in the prescribed `nix-shell`. Totals agree (599); the "pre-existing failure" does not reproduce and should not have been claimed.

**3. HEAD counts were stale** — they predated a later commit. Current: **624 passed / 1 skipped / 0 failed**.

### 🔴 Two mutants survived; both now die

| mutant | before | now |
|---|---|---|
| check errors only on the dict path | **SURVIVED** | KILLED (2 red) |
| bypass the normaliser entirely | **SURVIVED** | KILLED (2 red) |
| drop the singular `error` branch | **SURVIVED** | KILLED (1 red) |
| drop the timestamp hint in the error | n/a (new) | KILLED (1 red) |

Control green at 624 before *and* after each mutant, full detached worktree, `PYTHONDONTWRITEBYTECODE=1`, caches cleared between runs.

The first is the **inverse** of the bug this PR fixes and the worse direction — a FAILED send recorded as `sent`. It is reachable: upstream sets `Errors` for any non-SUCCESS recipient while still returning 201. It survived because *every* `transmit=` fixture in the repo returns a bare dict, so nothing drove `send_approved()` with a list. Verifying the normaliser in isolation was not enough.

The second is the dangerous-quiet one: on a two-entry reply it does not raise, it silently takes `entries[0]`, stores that timestamp and marks the draft `sent` — breaking sync-echo dedupe exactly as a locally generated timestamp would.

### Other audit findings fixed

- **The "REAL Postgres" guard executed a hand-copied query** — and spelled it `signal_uuid::text`, the exact spelling its sibling test forbids shipping. It could never fail when the real SQL regressed. Now extracts the SQL from source, anchored on `send_state IS NOT NULL` so it cannot grab a look-alike, with a positive control proving the extraction found the right query.
- **The spelled backstop was walked by two refactors** the audit confirmed raise `DatatypeMismatch` on the live server (a `CASE` rewrite and `NULLIF`). It now matches the hazard — any type-unifying construct naming `signal_uuid` — with both walks as parametrised regression cases.
- **Per-recipient errors discarded the server timestamp.** A partly-failed group send DID go out and the reply carried the timestamp `reconcile --sent --timestamp` needs. Now included in the error message.

🔴 **My first widened regex over-matched catastrophically:** a bare `\bCASE\b` with `re.S` matched the English word *"case."* in a docstring and ran to a distant `END`, swallowing 17 KB of source into a single "offending expression". Anchored to `CASE WHEN`, bounded, and SQL comments stripped first — they discuss `signal_uuid` precisely because they explain this hazard.

### 🔴 Release sequencing — read before merging

Live draft **51** is an **already-delivered** group message stuck in `sending` (`send_attempts=2`). Today `reconcile` is itself dead on Postgres, so it is frozen and harmless. **This PR unfreezes it.** If anyone then runs `reconcile 51 --not-sent` → re-approve → `send`, the group receives the message **twice**.

Required step at release: `reconcile 51 --sent --timestamp <ts from the Signal thread>` — **never `--not-sent`**. No row currently carries `1787331796630`, so that message's sync-echo dedupe has nothing to collide with.

### Still not verified

- `nix build .#checks.x86_64-linux.pytests` — not run; all counts are ad-hoc `nix-shell`.
- The group-send response shape empirically — read from upstream `client.go`/`datastructs.go`, not observed, because observing it requires transmitting to a real group.
- Whether the group actually received draft 51's text — inferred from the measured `201` plus `send_attempts=2`. **Someone must look at the thread before reconciling.**
- The merged tree: the signal suite was run on this branch only.

ZacxDev:
## 🔴 Correction to the release step above — do NOT use that timestamp

My previous comment (and the audit it summarised) attributed `1787331796630` to draft 51's send. **That is wrong.** It belongs to a different message.

Three POSTs hit `/v2/send` that afternoon, from the signal-api access log:

| time (UTC) | status | what it was |
|---|---|---|
| **16:58:12** | 201 | **draft 51** → the Vetr app group |
| 17:02:58 | 400 | invalid-recipient probe (delivered nothing) |
| **17:03:17** | 201 | a **note-to-self** probe, sent to capture the 201 body shape |

`1787331796630` decodes to **17:03:16 UTC** — the note-to-self, not draft 51.

**Draft 51's server timestamp is unrecoverable from our side.** It was in the 201 body that the `AttributeError` discarded, and GIN logs the status line only, never the payload. No amount of log archaeology will produce it.

### The corrected release step

`reconcile 51 --sent --timestamp <ts>` still applies — but the timestamp **must be read from the Signal thread itself**, from the message as it appears in the Vetr app group.

🔴 **Do not substitute `1787331796630`.** Storing another message's timestamp is precisely the failure `--sent --timestamp` exists to prevent: the sync echo would fail to dedupe, and the group's message would be stored twice under two identities. That is the same class of harm as the double-send the sequencing note warns about, arrived at from the other direction.

If the timestamp cannot be recovered from the thread, the honest options are to leave draft 51 in `sending` — it is inert and will never resend on its own — or to reconcile it deliberately knowing dedupe for that one message is forfeit. Both are defensible; guessing a timestamp is not.

### Why this is worth recording rather than quietly fixing

The error came from a blind audit, was repeated by me in a PR comment, and was caught only by decoding the epoch and diffing it against the access log. Two independent readers had already accepted it. The general lesson matches this PR's subject: **a value observed in one context does not belong to another just because it appeared nearby in the notes.**

PR #665, opened 2026-08-21
feat(hooks): capture every backgrounded Bash command VERBATIM — the artifact 868ktvqf9 needs

ZacxDev:
Instrumentation that guarantees the **next** hit of ClickUp **868ktvqf9** carries its own
evidence. **This is not a fix for that ticket** and does not attempt one — the standing
instruction is that no further fix may be attempted until the next hit is captured.

## The defect this enables the diagnosis of

A full unit-suite run routed through the civitai dev-server test queue exits non-zero and
the harness's background-task notification reports **exit 0**. `cli.mjs test list` shows
`exitCode: 1`; the notification says `0`. At the moment the notification arrives the run's
output file is **0 bytes** — so the two independent things an investigator would check both
report green over a red run. Four hits in one day across four agents; never reproduced
deliberately since.

**The single artifact that discriminates between the open hypotheses is the VERBATIM command
string that was backgrounded**, and the harness keeps it nowhere reachable afterwards.

## Mechanism, and why this one

I read this host's real transcripts rather than trusting the docs, which are silent on three
of the four points that decided the design:

| | measured |
|---|---|
| `PreToolUse{Bash}.tool_input` | exactly `{command, description, timeout, run_in_background}` — **the only place the verbatim string and the background flag exist** |
| `PostToolUse{Bash}.tool_response` (backgrounded) | `{stdout:"", stderr:"", interrupted:false, isImage:false, noOutputExpected:false, backgroundTaskId:"…"}` — 🔴 **no exit code, and there cannot be one**: it fires at *launch*, which is why stdout/stderr are empty strings |
| completion notification | **not a hook event.** Injected as `<summary>Background command "DESC" completed (exit code N)</summary>` — 🔴 keyed on the **`description`**, not on the task id |
| a hook at background completion | **does not exist** |

So the honest split is: **the hook** captures the verbatim command + description + task id;
**`--report`** (a read, run later by an investigator) joins that against the transcript's
`completed (exit code N)` line. The hook never claims an exit code it cannot see — it records
`harness_exit_field: null` as a *measurement* ("we looked, there was none") beside a complete
`response_keys` ledger that goes visibly non-null if a future harness ever supplies one.

Registered on **PreToolUse(Bash)** (the load-bearing half) and **PostToolUse(Bash)** (the only
place `backgroundTaskId` — which names the output file found at 0 bytes — appears).

**Scope:** writes only when the call is backgrounded **or** the command carries a masking
marker. Neither half subsumes the other: backgrounded-only misses the foreground `| tail` that
bit the investigating agent twice; marker-only misses a backgrounded run whose masking shape
nobody has thought of yet, which is the open question.

## The grep marker

`REDIRECT_THEN_SEMI` — a top-level redirect *before* the final top-level `;`/newline separator
that has a real command after it. Siblings `TRAILING_SEMI` and `TRAILING_PIPE` come free in the
same pass. They are **triage markers, not a gate**: nothing is blocked, warned or rewritten.

The scanner tracks quotes, escapes, `$(…)`/backtick nesting, comments and **heredoc bodies** —
without the last, every `python3 - <<'PY' … PY` payload on this host would false-positive.
`&&`/`||` are deliberately *not* separators: they short-circuit, so they do not mask.

## 🔴 Positive control — a real trailing-`;`-after-redirect command, captured verbatim

Real hook binary, real payload shape, real disk:

```
$ python3 scripts/claude-hooks/bg-command-capture.py < pre.json   # rc=0, silent
$ cat "$CLAUDE_BG_CAPTURE_DIR/commands.jsonl"
```
```json
{"schema":"BG_COMMAND_CAPTURE_V1","event":"PreToolUse","ts":"2026-08-21T18:37:56Z",
 "session_id":"a248d005-f3c6-4318-a8d5-f806d462b026","tool_use_id":"toolu_01LzjqnUJokwRCPSLvhoscDJ",
 "cwd":"/home/zach/workspace/civit/civitai","transcript_path":"…","background":true,
 "timeout_ms":1800000,"description":"Run main unit suite in background",
 "markers":["TRAILING_SEMI","REDIRECT_THEN_SEMI"],
 "command":"node .claude/skills/dev-server/cli.mjs test run --suite unit > /tmp/unit.log 2>&1; echo \"done rc=$?\""}
{"schema":"BG_COMMAND_CAPTURE_V1","event":"PostToolUse",…,"background_task_id":"b6v7iqmsd",
 "response_keys":["backgroundTaskId","interrupted","isImage","noOutputExpected","stderr","stdout"],
 "interrupted":false,"harness_exit_field":null,"stdout_len":0,"stderr_len":0}
```

And the join — the two numbers that disagree, on one row:

```
ANNOUNCED [0] | marks TRAILING_SEMI,REDIRECT_THEN_SEMI | task b6v7iqmsd
   cmd: node .claude/skills/dev-server/cli.mjs test run --suite unit > /tmp/unit.log 2>&1; echo "done rc=$?"
```

Also shipped as `--selftest`, so a live empty log can be distinguished from an instrument wired
to nothing before anyone concludes "it never fired":

```
$ python3 scripts/claude-hooks/bg-command-capture.py --selftest
wrote: True  read back: 1  unparseable: 0
verbatim round-trip: True
markers: ['TRAILING_SEMI', 'REDIRECT_THEN_SEMI']
positive control: 1 expected, 1 observed -> PASS
```

## 🔴 Never-throws evidence

21 hostile inputs through the real hook — empty, whitespace, non-JSON, `null`, array, string,
number, truncated JSON, missing/unknown/non-string event, `tool_input` null and list, non-string
and empty command, non-Bash tool, string `tool_response`, 2000-deep nesting, **an 8 MiB
command**, unterminated quote, unterminated heredoc, unicode+emoji — plus an unwritable state
directory and the hook copied away from its library. **All 21+2: `rc=0`, empty stdout, empty
stderr.** stdout matters as much as the exit code here: on PreToolUse it is parsed as a
permission verdict.

`test_the_never_throws_harness_can_go_red` is the battery's **negative control** — every
assertion in it is of the form "nothing happened", so without it the battery is
indistinguishable from one wired to nothing. It plants a script that exits 3 and writes to
stderr and asserts the same three checks fail. Verified red.

**Bound:** rotate-*before*-append, one generation, ceiling `2 * (8 MiB + one record)`. Measured
under that ordering (re-measured after the ordering changed, not quoted from the earlier run):
an 8 MiB command wrote an `8,388,930`-byte live log; the *next* write rotated it to
`commands.1.jsonl` and left the live log at `309` bytes, with **both generations still readable
and the 8 MiB command byte-intact** (2 records, 0 unparseable). The bound is on the **file**,
never the record — a truncated command is exactly the evidence that would be missing, and the
suite additionally round-trips a 1 MiB command byte-for-byte against a 4000-byte cap.

Rotating *before* rather than *after* is behavioural, not cosmetic: rotating after leaves **no
live file at all** in the window until the next write, which an investigator reads as "the
instrument never ran" — the reassuring absence this whole thing exists to remove.

## Two defects the suite caught — both reproducing the ticket's shape inside the instrument

1. The notification regex required a bare `"`. Transcripts store the notification inside a JSON
   string, so its quotes are **backslash-escaped**; the join returned `announced_exit_codes: []`
   for a record whose exit code was in the same file.
2. Fixed, the greedy capture then **swallowed the escaping backslash**, so the key came back as
   `'…background\\'` and matched nothing — an empty list again, for a second reason.

Both read as "the harness never announced one" rather than "my pattern is wrong". Now pinned in
**both** spellings with distinct exit codes, plus a negative control.

A third, from the scanner: the separator after a heredoc's delimiter line was not recorded, so
`python3 - <<'PY' > log … PY` + `echo done` — the masking shape *with* a heredoc — went
unmarked. Red in the ledger, fixed, pinned.

## Tests

`scripts/claude-hooks/tests/test_bg_command_capture.py` — **80 tests**, added to
`HERMETIC_TARGETS` with floor `76` (`_suggested_floor 80`). Marker ledger is 28 rows, **14
expecting no marker**; write predicate 4/4 both ways.

Three cross-module seam ledgers went red and were updated rather than worked around — each did
its job: the registrar's own idempotency self-check (`MANAGED_HOOK_SCRIPTS`), the deployed-path
ledger (`HOOK_LIBRARY_MODULES`), and `test_on_disk_artifact_names.py`'s
every-hook-is-classified enumeration.

Baseline control run first: `test_a_module_that_exits_at_import_is_a_named_error_not_no_tests_ran`
is **already red at `origin/main`** and is not from this branch.

### Gate

Authoritative gate, in the repo's own dev shell:

```
RESULT: PASS (exit=0)
  TOTAL collected=14173  passed=14172  skipped=1  failed=0  (floor: 13400 = sum of 26 per-target floors)
  PASS  scripts/claude-hooks/tests/test_bg_command_capture.py  (collected=80 passed=80 skipped=0 floor=76)
```

An earlier run of the same tree came back `failed=1` on
`scripts/browser-bridge/tests::test_the_throttle_path_carries_both_the_hash_and_the_join_key`.
Attributed to load, not to this branch, on four pieces of evidence rather than a re-run alone:
it passes in isolation at **both** `origin/main` and this branch; the full 779-test
browser-bridge suite passes **779/779** on this branch; this diff touches no browser-bridge
file; and the whole suite took ~224 s in both runs under a host load average of 8–14 with a
second devrc gate running concurrently — load inflating every test, not one assertion timing
out. The clean re-run above is green.

## 🔴 What I could NOT verify

- **The hook has never run inside a live Claude Code session.** It is deployed by home-manager
  as a read-only `/nix/store` symlink and registered by `register-nudge-hook.py`, so
  **`home-manager switch` is required to activate it** — I did not run one, and did not touch
  `~/.claude/settings.json`. Everything above is the hook driven as a subprocess with payloads
  whose shape I measured from real transcripts. "The shape is right" and "the harness sends it"
  are different claims and only the first is evidenced here.
- **It cannot observe a backgrounded run's true exit status**, because no hook event fires at
  completion. It records what the harness *announced*; the truth still has to come from the
  queue's own record or the output file.
- **Per-call overhead is not benchmarked.** It adds one Python process to PreToolUse and one to
  PostToolUse (currently 2 and 5). The write predicate is a single string pass with no I/O when
  it misses, but I have not measured the added latency on a real session.
- **The marker scanner is a heuristic**, deliberately allowed to be wrong in the false-positive
  direction. It is not a shell parser and will misjudge exotic constructs.
- **Concurrent single-record atomicity** is `O_APPEND` + one `os.write()`, which POSIX does not
  formally guarantee for very large records. Every reader therefore *counts* unparseable lines
  and reports the count rather than dropping them.

## Live corroboration, unplanned — the symptom reproduced while building the instrument

I backgrounded this PR's own gate as `nix develop … run-tests.sh … | tail -50`. Both halves of
the 868ktvqf9 symptom appeared:

- the output file sat at **0 bytes** for the entire run (`tail` buffers until the pipeline ends);
- the completion notification read **`completed (exit code 0)`** while the file's own last line
  read **`RESULT: FAIL (exit=2)`**.

A trailing pipe, no `;` needed, and a confident green over a red run. Reading the *content*
rather than the status is what caught it — which is precisely the manual step this log exists to
make unnecessary. (That particular FAIL was self-inflicted for an unrelated reason: I edited
`run-tests.sh` while bash was still executing it, shifting its byte offsets mid-run. `bash -n`
passed on both the branch and `origin/main`, which is how that was distinguished from a real
defect. The gate was then re-run on a frozen tree with no trailing pipe.)

### DIFF
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -1151,6 +1151,28 @@ in
   home.file.".claude/hooks/agent_ledger.py" = {
     source = ../scripts/lib/agent_ledger.py;
   };
+  # 🔴 THE BACKGROUNDED-COMMAND CAPTURE LOG — instrumentation for ClickUp
+  # 868ktvqf9, where a unit run that exits non-zero is announced as "exit code 0"
+  # and the run's output file is 0 bytes at the moment the notification arrives.
+  # Both independent things an investigator would check report green over a red
+  # run, and the ONE artifact that discriminates between the live hypotheses —
+  # the VERBATIM command string that was backgrounded — is kept nowhere the
+  # investigator can reach afterwards. This records it.
+  #
+  # It never blocks, warns or rewrites anything; it appends a line and returns 0.
+  # `bg_command_capture.py --report` is the read that puts the captured command
+  # next to the exit code the harness announced for it.
+  #
+  # 🔴 BOTH files are NEW, and the hook imports the library as a SIBLING in
+  # ~/.claude/hooks/ — deploying one without the other is a green switch and an
+  # inert hook (the #452 shape). Both must be `git add`ed or the flake silently
+  # omits them, and the switch still succeeds with the hook absent.
+  home.file.".claude/hooks/bg-command-capture.py" = {
+    source = ../scripts/claude-hooks/bg-command-capture.py;
+  };
+  home.file.".claude/hooks/bg_command_capture.py" = {
+    source = ../scripts/lib/bg_command_capture.py;
+  };
   # 🔴 THE CLAWGATE WRITE-BACK GUARD — the deterministic replacement for a 🔴 prose
   # rule that lost 2/2. Tasks #193 and #194 were both picked up, the work shipped as
   # PRs, and both cards stayed `open` with ZERO comments; both were re-dispatched and
--- a/scripts/claude-hooks/bg-command-capture.py
+++ b/scripts/claude-hooks/bg-command-capture.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python3
+"""Claude Code adapter for the backgrounded-command capture log — ClickUp 868ktvqf9.
+
+All the logic lives in `bg_command_capture.py`; read ITS docstring first, which
+carries the measured harness payload shapes and the reasoning. This file is the
+adapter and nothing more, in the arrangement `bash-guard.py` has with
+`guard_core.py` and `agent-ledger-hook.py` has with `agent_ledger.py`: deployed,
+both files sit in `~/.claude/hooks/`; in the repo the module is in
+`scripts/lib/` and only this file is under `scripts/claude-hooks/`.
+
+WHICH EVENTS, AND WHY EACH ONE
+
+    PreToolUse(Bash)   🔴 THE LOAD-BEARING ONE. `tool_input.command` is the
+                       verbatim string and `tool_input.run_in_background` is the
+                       boolean, and this is the ONLY event where either exists.
+                       Everything the ticket needs is captured here.
+    PostToolUse(Bash)  the `backgroundTaskId` — which names the output file the
+                       operator finds at 0 bytes — appears only in
+                       `tool_response`, and only on this event. It carries NO
+                       exit code, because it fires at LAUNCH, not completion.
+
+There is deliberately no third event: no hook fires when a backgrounded task
+finishes. The exit code the harness announces is injected into the transcript as
+`<summary>Background command "…" completed (exit code N)</summary>`, joined back
+to a captured command by its `description` — that join is `--report`, run by an
+investigator, not by a hook on anyone's hot path.
+
+🔴 FAIL-OPEN, ALWAYS, AND SILENTLY. This fires before and after EVERY Bash call
+in EVERY session on this host. A hook that raises breaks the operator's shell,
+and a PreToolUse hook that writes junk to stdout is read as a permission verdict.
+So: every path returns 0, nothing is ever printed, and the outermost handler is
+`BaseException` — not `Exception`. That widening is not defensive decoration; it
+is the measured lesson from `bash-guard.py`, where a `SystemExit` out of an
+imported module escaped an `except Exception` and changed the hook's verdict.
+Here the direction is the opposite one (this hook has no verdict to change) but
+the mechanism is identical, and a traceback on stderr for every Bash call would
+be its own kind of broken.
+
+This hook is INSTRUMENTATION. It never blocks, warns, rewrites or refuses
+anything. Turn it off with `CLAUDE_BG_CAPTURE_DISABLE=1`.
+
+Registered by `register-nudge-hook.py`; deployed by `nix/home.nix` alongside
+`bg_command_capture.py`. Both need a `home-manager switch` to take effect.
+"""
+import json
+import os
+import sys
+
+sys.dont_write_bytecode = True
+_HERE = os.path.dirname(os.path.abspath(__file__))
+
+EVENTS = ("PreToolUse", "PostToolUse")
+
+
+def _load():
+    """Import `bg_command_capture` from the deployed sibling, else the repo lib/.
+
+    Deployed, both files sit in `~/.claude/hooks/`. In the repo only this file is
+    under `scripts/claude-hooks/`, so the tests exercise the second branch and
+    the host exercises the first — the same two-branch arrangement
+    `agent-ledger-hook.py` uses, and for the same reason.
+    """
+    import importlib.machinery
+    import importlib.util
+    for path in (os.path.join(_HERE, "bg_command_capture.py"),
+                 os.path.join(_HERE, os.pardir, "lib", "bg_command_capture.py")):
+        if os.path.exists(path):
+            loader = importlib.machinery.SourceFileLoader("_bg_command_capture", path)
+            spec = importlib.util.spec_from_file_location(
+                "_bg_command_capture", path, loader=loader)
+            mod = importlib.util.module_from_spec(spec)
+            loader.exec_module(mod)
+            return mod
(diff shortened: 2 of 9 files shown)


ZacxDev:
**The `CONFLICTING` status on this PR is external to the change — please do not rebase it away yet.**

`origin/main` currently points at `14200130`, whose tree contains **1 file**. Between this
branch's base (`f0442f49`, #659) and that commit sit **81 commits, 63 of them git test
fixtures** authored as `t@example.com` / `t@example.invalid` / `real@example.com`, pushed in a
single ~60-second burst at **2026-08-21T19:27–19:28Z** (four `PushEvent`s to `refs/heads/main`).
They delete ~300 real files; the 949-file tree is intact only as history.

Nothing is lost: `f0442f49` is still an ancestor, and the **18 genuine commits** that landed on
`main` earlier today (`zachlowden1@`, `zacxdev@`, `prettynao2006@`) are in that history too. But
restoring `main` means rewriting a shared branch, which is a maintainer decision, so I have not
touched it.

Likely mechanism: `scripts/tests/test_analyze_service_index_commit.py:426` runs
`_git(store / "some-scope", "push", "origin", "trunk")`. That is meant to hit a local bare
fixture; the commit messages on `main` (`autocommit: N change(s) in the some-scope
analyze-service index`, plus `base` / `ahead` / `m`) are exactly that suite's fixtures, so the
push reached the real remote instead.

I cannot attribute which run did it. Two devrc gate runs overlapped that window — mine and a
concurrent one in `.claude/worktrees/agent-a385372295fafed05` (still running as of 19:36Z) — and
both worktrees share one git dir, so the `update by push` reflog entries do not separate them. I
am not naming a culprit on a coin flip.

This branch itself is unaffected: it is based on the real `f0442f49` and carries the full
949-file tree. Once `main` is restored the PR should merge without changes. **Worth holding off
on running the devrc gate until the fixture's remote is fixed** — it may be the mechanism.

PR #668, opened 2026-08-21
feat(prune-index): the index store had NO prune surface — and the rule meant to be one CONTRADICTED the schema

ZacxDev:
## The problem, measured

`~/.claude/analyze-service-index/` grows monotonically and nothing applies pressure. **Measured 2026-08-21T18:52Z: 77 entries / 359,395 B**, median entry 3,002 B, largest 27,770 B. The store's measured value is **recency-ordered selection**, and selection is exactly the property that degrades as entries grow — so this is not tidiness.

🔴 **The root cause was a rule contradiction, not laziness.** `analyze-service/reference/write-back.md` said *"Prune-on-resolve — when a gotcha is fixed, **remove** the bullet"*, while the newer `OPEN:` / `RESOLVED <sha>:` schema in `index-store.md` says **mark** it. Both were live and both were followed, so `RESOLVED` bullets accumulated forever. Both docs are reconciled here, **outside** the hashed `resolver-rules` / `entry-schema` marker regions — `test_subsystem_resolver.py::TestCommandDocIsPinned` stays green (321 passed).

## The lifecycle, now implemented and gated

- **`OPEN:` ALWAYS STAYS** — never an eviction candidate at any age or size, and never counts toward reclaiming bytes.
- **`RESOLVED <sha>` is evictable ONLY once a target it names** (a `claudedocs/` path, a commit sha, a PR/issue ref) **is verified to exist** — and the audit names the target it verified.
- 🔴 **`RESOLVED` with no reachable home ⇒ `NO HOME — write the record first`**, never evictable: the bullet is the only copy of the finding.
- **Nothing is ever auto-evicted.** The tool reports; a human confirms on a diff.

## What lands

| file | what |
|---|---|
| `scripts/subsystem-audit.py` | deterministic, **READ-ONLY** auditor |
| `scripts/tests/test_subsystem_audit_budget.py` | **owns** the per-entry budget + the eviction playbook |
| `scripts/tests/test_subsystem_audit.py` | 31 hermetic, fully synthetic tests |
| `claude/skills/prune-index/` | the confirm-gated pruning skill + 2 reference topics |
| `claude/skills/analyze-service/reference/{write-back,index-store}.md` | the reconciliation |

**Read-only is enforced, not asserted.** The script has no write path at all, and its one shell-out (`git`) **refuses any repo path resolving inside the store root** — whose scope dirs are each their own git repo. Two tests hold it: a byte-hash of every fixture before/after a full audit+render, and a source grep for write calls (the hash test alone goes green for a write on a path no fixture reaches).

## The budget is DERIVED, and the test owns it

- **Hard cap 12,288 B** — *imported* from `scripts/browser-bridge/tests/test_skill_size.py::MAX_BYTES`, the proven `SKILL.md` body ceiling. So "an index **entry** must not outweigh a whole skill body" is an **executable** claim: if that budget moves, the test goes red and a human decides, instead of the two silently diverging.
- **Target 6,144 B** — the corpus's own demonstrated shape: **64 of 77 entries (83%) already fit**, measured p85 is 6,212 B. Pinned at exactly half the cap so two independent derivations agree.
- Both boundaries are probed **on both sides** (at the budget = OK, one byte over = reported), so the constants are load-bearing rather than two numbers in a comment. The skill body states no number and routes to the test.

## Live-store findings (audit output, aggregate only — devrc is PUBLIC)

```
13 of 77 entries over the 6,144 B target (5 over the hard cap); excess 81,475 B
26 OPEN bullets in 21 of 77 entries          — 🔒 KEEP, excluded from every total
29 RESOLVED: 29 EVICTABLE / 0 NO HOME / 0 NOT CHECKED
   ⚠ 4 of the 29 resolve ONLY in a sibling scope's repo — flagged as the
     weaker claim it is, "confirm the attribution before evicting"
21 of at least 228 path pointers do not resolve
22 of 77 lack `sensitivity:` · 23 lack `created_by:` · 23 still spell `repo:`
4 scopes have no README.md (cli, flipt-state, kubeclaw, storage-resolver)
428 of 518 bullets exceed the schema's "≤2 lines" — ADVISORY, kept OUT of the
   verdict: a gate red on 83% of a curated corpus is a permanently-red gate
```

⚠ **`0 NO HOME` is an honest reading that disagrees with the brief.** A repo-*scoped* sha check does report 4 — and all four resolve in a sibling scope's repo. The cross-repo fallback is what removes them and it is correct (the store spans several repos). The classifier is proven by mutation instead of by a live hit.

## 🔴 A live ref collision, found and fixed

The alias `ads` was declared by **both** `alert-diagnose-svc.md` and `civitai-advertising.md` in one scope, so `--ref ads` returned `ref-ambiguous` and surfaced **nothing** — the entry was unreachable by the name a human would type. Dropped from `alert-diagnose-svc` (an initialism there; the actual word for the advertising service). Verified after: `--ref ads` → `status=recalled`, resolving to `civitai-advertising.md` via the alias tier. Clearing an ambiguity by making the ref resolve to *nothing* would be a regression, and the test pins against that.

The detector never re-implements the two-tier rule: every addressable ref (336 store-wide) is fed back through `resolve_ref_tiered` — the executable authority — so an alias merely **shadowed by a filename** is correctly *not* reported, which a hand-rolled "count the claims" detector gets wrong.

## Denominators everywhere

Every count prints beside "N entries examined" / "at least N pointers checked" / "N addressable refs". `NOT CHECKED` is a **tri-state**, never folded into a clean total: a scope with no derivable owning repo and a PR ref needing the network are *unmeasured*, not *absent*, and the verdict says so explicitly.

## Guards proven, not assumed

8-mutant battery under `PYTHONDONTWRITEBYTECODE=1` with an unmodified control (green). All eight killed, each by its own guard's test:

| mutant | killed by |
|---|---|
| stop feeding aliases to the resolver | `test_collision_detector_finds_the_alias_claimed_by_two_entries` |
| swallow `AmbiguousRefError` | same |
| `NO_HOME` → `EVICTABLE` | `test_no_home_fires_when_every_named_target_is_unreachable` |
| drop the `NOT CHECKED` branch | `test_an_unmeasurable_target_is_not_no_home` |
| accept a checked-and-**absent** target as a home | `test_no_home_fires_when_every_named_target_is_unreachable` |
| let `OPEN` bullets reach the classifier | `test_open_bullets_are_never_classified_for_eviction` |
| sha-side: treat a sibling-repo miss as absence | `test_a_miss_in_SIBLING_repos_alone_is_not_evidence_of_absence` |
| path-side: same collapse in `resolve_path_token` | same |

The battery also exposed a latent renderer crash on an impossible state; fixed here, so a classifier bug stays diagnosable instead of taking three unrelated tests down with it.


ZacxDev:
⚠ **This PR shows `CONFLICTING` for a reason unrelated to its contents.**

`origin/main` is currently **not the devrc repo**: 79 commits were pushed on top of the real tip `fc83693e` between 2026-08-20 23:14 and 2026-08-21 14:28, authored by `T <t@example.com>` / `t@example.invalid` / `Someone Real <real@example.com>` with messages like `base`, `ahead`, `local side`, `un-pushed work stranded on main` — i.e. **git test fixtures**. The cumulative diff is `949 files changed, 2 insertions(+), 400083 deletions(-)`, and `main`'s tree at the tip contains a **single file named `f`**.

The real history is intact — `fc83693e` is still an ancestor — so this is recoverable. ⚠ But the range is **not purely junk**: it also carries real commits by `Zach Lowden <zachlowden1@gmail.com>` (`fix/managed-path-wrong-writer`, `fix(drift-check)`, `fix(gate)`, …), so a blind `--force` back to `fc83693e` would drop real work.

The current tip is preserved additively at `refs/heads/incident/main-fixture-wipe-2026-08-21` (`1420013`) so any recovery is reversible. Nothing here touches `main`.

**This branch is based on `fc83693e`, the real main at the time it was cut**, and was gated there. Re-target/rebase once `main` is restored.

ZacxDev:
✅ **Update — `main` has been restored and this PR is now `MERGEABLE / CLEAN`.**

`origin/main` is back at `fc83693e` (0 commits ahead of this branch's base), so the merged tree **is** the tree that was gated.

Correcting one thing I said above: I flagged a risk that restoring `main` would drop real work. Measured against the preserved ref, the 79-commit range was **63 fixture commits** (`t@example.com`, `t@example.invalid`, `real@example.com`) and **16 real ones** (`zachlowden1@gmail.com`, `prettynao2006@gmail.com`) — but those 16 are still on their own branch, `refs/heads/fix/managed-path-wrong-writer` @ `bd05708`. **Nothing was lost**; the restore simply un-merged a branch that should land via its own PR.

`refs/heads/incident/main-fixture-wipe-2026-08-21` (`1420013`) is left in place as a record. Delete it once someone has confirmed they don't want it.

⚠ Still open, and separate from the remote: the local shared clone `~/workspace/devrc` has `core.bare = true` on a populated working tree with `HEAD` → `refs/heads/trunk`, so `git status` there fails. That needs fixing on the host.

PR #673, opened 2026-08-21
guard(tests): a test rewrote the operator's GLOBAL git config and pushed 63 fixture commits to the REAL origin/main

ZacxDev:
## What happened (measured 2026-08-21)

A test ran `githooks/install.sh` for real. Line 39 of that script is
`git config --global core.hooksPath "$DIR"`, so it repointed the operator's
**global** git config at a pytest tmpdir. In the same window ~63 fixture commits
(`base`, `ahead`, `local side`, `un-pushed work stranded on main`,
`autocommit: N change(s) …`) reached the **real** `origin/main`, whose tree became
a single file named `f`, and the base clone ended up `core.bare = true` on a
populated working tree. All repaired; nothing lost.

The offending test is on no origin branch — it is uncommitted work in another
session's worktree, and it is not the point. The point is that the suite had **no
floor**, so the next such test does it again.

## The shape — third instance, copied exactly

| guard | what escaped | PR |
|---|---|---|
| GUARD 7 | real systemd timers, 158 desktop toasts, on a green suite | #399 |
| GUARD 8 | real rows into the production ClickHouse `activity.events` | #614 |
| **GUARD 9** | **the operator's git config, and pushes to real remotes** | this |

🔴 **One module, two entry points — never a fixture in N conftests.**
`scripts/run-tests.sh` runs ONE pytest process per target directory and there are
~17, so a conftest fixture protects exactly one: #399 protected 1 target of 17 and
#614 protected 1 directory of 13, both while reading as systemic. So the rule
lives in `scripts/testlib/nogit_plugin.py` and is registered at:

1. **`-p testlib.nogit_plugin`** on `run-tests.sh`'s single `python -m pytest`
   line — the one place every target is invoked — plus the same exports for the
   whole script, which is what covers the NON-pytest targets (`HOOK_TESTS`,
   `SHELL_TESTS`) that no conftest can ever reach;
2. **an import in `scripts/tests/conftest.py`**, so a bare `pytest scripts/tests`
   outside the runner goes through the same code, not a second copy.

## The levers

| lever | why |
|---|---|
| `GIT_CONFIG_GLOBAL` → throwaway file | `git config --global` writes there; the operator's config is *unreachable*, not merely un-written-by-well-behaved-code. The surgical fix for the surface that was poisoned. |
| `GIT_CONFIG_SYSTEM=/dev/null` + `GIT_CONFIG_NOSYSTEM=1` | the same hazard one level up; **both**, because they cover different git versions and neither substitutes for the other |
| `GIT_ALLOW_PROTOCOL=file` | git itself refuses `https`/`ssh` (`fatal: transport 'https' not allowed`, exit 128) while `file://` and plain paths keep working |
| `GIT_TERMINAL_PROMPT=0` | a git op that escaped the allowlist must fail, not sit waiting for a password inside a test runner |

🔴 **HOME is deliberately NOT reassigned, and that decision is pinned by a test.**
Several suites legitimately read `~/.claude/...` (the analyze-service index store
among them), so a blanket HOME rewrite would break real tests and be reverted —
a durable guard traded for a temporary one.

The residual hazard that leaves is stated rather than implied: code that *removes*
`GIT_CONFIG_GLOBAL` from its own environment drops back to `$HOME`. So the runner
fingerprints the operator's real config files before and after **every target** and
fails the run naming both the target and the file. That tripwire also catches the
repo-local `core.bare` write, which `GIT_CONFIG_GLOBAL` does not govern at all.

## Is `GIT_ALLOW_PROTOCOL=file` too broad? Measured, not assumed

git 2.55.0, with the allowlist active:

| operation | result |
|---|---|
| `ls-remote https://…` | `fatal: transport 'https' not allowed` (128) |
| `ls-remote ssh://…` and scp-style `git@host:path` | `fatal: transport 'ssh' not allowed` (128) |
| push / clone, **plain filesystem path** | rc 0 |
| push / clone, **`file://` URL** | rc 0 |
| `ls-remote https://…` with the lever removed | **succeeds** (positive control) |

Both local shapes are the only remote shapes this tree's fixtures use, and the
suite pins them (`test_the_allowlist_still_permits_every_local_remote_shape`).
Not too broad; nothing legitimately needs https during tests.

## Proof — the controls, because a zero is what a dead guard also prints

Per **pytest** target the runner requires all of:

* a **session marker** — a target without one ran unguarded, and its
  `real-config-changed=0` means nothing;
* a **config control** — a REAL `git config --global` write that must be present
  in *this run's* isolated file (reading it back through `--global` only proves
  git wrote *somewhere*);
* a **protocol control** — a REAL `https` git operation refused **by git**, not by
  DNS or a missing credential. Those are all non-zero exits, so keying on "it
  failed" would score a completely unprotected run as protected.

And before anything runs, three fail-closed preconditions: the redirect must
actually govern `git config --global`, the allowlist must actually refuse https,
and the tripwire's comparator gets its own **negative control** — a canary is
rewritten between two fingerprints and the run aborts if they match.

Per-target line, always printed, never the zero alone:

```
scripts/tests  real-config-changed=0/3  config-control=1  protocol=refused  plugin=1
```

### Mutation battery (`PYTHONDONTWRITEBYTECODE=1`, `__pycache__` cleared, `-p no:cacheprovider`)

| mutant | outcome |
|---|---|
| redirect points at `$HOME`'s config | **killed** — real-installer test |
| `GIT_ALLOW_PROTOCOL` dropped | **killed** — https-push test |
| `GIT_CONFIG_SYSTEM` dropped | **killed** — live-session lever test |
| session marker never written | **killed** — runner's `plugin=0` check |
| containment check removed | **killed** — fail-closed control test |
| protocol control accepts any failure | **killed** — https-push test |
| **an inert local rename (negative control)** | **SURVIVED**, as it must |

Control green before and after. Each mutant died on **its own** assertion message,
not a neighbour's.

One finding the battery produced about *itself*: the first version of the
real-installer test took its after-digest **outside** the `try`, after the cleanup
had already unset the key — so the "redirect points at `$HOME`" mutant SURVIVED a
test written specifically to catch it. The ordering is now load-bearing and
commented as such.

## Every-target coverage — the existing assertion, extended

`run-tests.sh`'s single `python -m pytest` line is now asserted to carry the
**complete set** of guard plugins, pinned both ways
(`test_the_single_pytest_invocation_loads_every_guard_plugin`) — a guard dropped
from that line covers nothing, and a guard *added* without touching the pin is a
coverage claim nobody reviewed. The GUARD 7 and GUARD 8 pins each assert the new
flag too, so dropping it goes red in three places.


ZacxDev:
## Correction: this PR's positive control for `GIT_ALLOW_PROTOCOL` is UNSOUND — do not reuse it

The control matrix claims:

> **Pos** — same push, allowlist removed | fails at DNS with no "not allowed" — the discriminator

That is not a discriminator, and I am correcting it here rather than editing the body because it may already have been copied.

**Measured by a peer session against the REAL remote:** with the lever absent, an ssh push to `git@github.com:innovation-upstream/devrc.git` **resolved, connected and authenticated**, printing a push progress header. It failed only because the ref update was rejected (non-fast-forward). No DNS failure at any point.

**Reproduced locally**, which explains the discrepancy — this PR's control used an *unreachable* host:

```
WITH the lever:                  fatal: transport 'https' not allowed
WITHOUT, unreachable.invalid:    fatal: ... Could not resolve host: unreachable.invalid
```

So the original control fails without the lever for a reason that has nothing to do with the guard. Against any reachable remote, `rc != 0` holds in **both** arms — a test asserting non-zero passes whether or not the guard exists.

**Only the literal `fatal: transport '<proto>' not allowed` is attributable.** Pin that string, never the exit code.

### Why it matters beyond this PR

A fixture push from this host reaches GitHub with **working credentials**. During the 2026-08-21 fixture-wipe window the pushes were fast-forwards onto branches nobody was defending, which is why 63 of them landed. The blast radius was never limited by access — it was limited by luck.

### Status

This PR's runner layer is **superseded and will not be merged**. `nogit_plugin.py` handles `GIT_CONFIG_*`, `GIT_ALLOW_PROTOCOL` and `GIT_TERMINAL_PROMPT` and **none** of `GIT_DIR / GIT_WORK_TREE / GIT_COMMON_DIR / GIT_INDEX_FILE` — measured, 0 occurrences. `GIT_WORK_TREE=<victim> git -C <innocent fixture> checkout -f HEAD` writes into the victim while repo identity resolves truthfully to the fixture, so this guard would have shipped looking complete while open on that entire sub-class. A peer session's armed control found it; my six mutants could not, because a mutation battery only covers the failure modes its author imagined.

The two levers here are sound and have been carried into the superseding guard: the `GIT_CONFIG_GLOBAL` redirect (the exact surface `githooks/install.sh` poisoned) and `GIT_ALLOW_PROTOCOL=file`. Two further defects found while building on them, worth knowing if this branch is ever revived:

- **Order bug:** deriving the protected set *after* redirecting `GIT_CONFIG_GLOBAL` protects the throwaway and leaves `~/.gitconfig` open — the guard ends up pointed at the decoy it just created. Protect `Path.home()/.gitconfig` unconditionally.
- **A structural check that read the documentation:** a mutant renaming the exported variable SURVIVED because the comment block above the export still spelled the token, and the assertion was `token in text`. Anchor to `^export`.

ZacxDev:
## Second hole in this guard: `GIT_CONFIG_COUNT` bypasses the `GIT_CONFIG_GLOBAL` redirect entirely

Recording this so anyone reviving this branch starts from the defect list, not the control matrix.

`man git` does not document `GIT_CONFIG_COUNT` / `GIT_CONFIG_KEY_<n>` / `GIT_CONFIG_VALUE_<n>`, but git honours them. They set arbitrary config **for a single invocation, without touching any file**. Reproduced here on git 2.55.0:

```
arm 1  GIT_CONFIG_GLOBAL=<decoy> GIT_CONFIG_SYSTEM=/dev/null
       git -C <fixture> config --get core.hooksPath      -> unset          (guard holds)

arm 2  SAME redirect, plus
       GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=core.hooksPath
       GIT_CONFIG_VALUE_0=/tmp/PWNED-hooks
       git -C <fixture> config --get core.hooksPath      -> /tmp/PWNED-hooks

decoy file after arm 2:  0 bytes, never opened
```

That reproduces the **measured incident vector** — `core.hooksPath` pointed at an attacker-chosen directory — while this PR's `GIT_CONFIG_GLOBAL` redirect points harmlessly at its throwaway.

**The decoy was never written**, and that is the structural point: this vector performs no file write anywhere. A guard whose observable surface is "a write to a protected path" cannot see it even in principle, and neither can one that redirects config *files*. `nogit_plugin.py`'s set is `GIT_CONFIG_GLOBAL, GIT_CONFIG_SYSTEM, GIT_CONFIG_NOSYSTEM, GIT_ALLOW_PROTOCOL, GIT_TERMINAL_PROMPT` — so this is not three missing entries in a ledger; the ledger was watching a channel the vector does not use.

**Generalisation worth keeping:** a guard is bounded by the CHANNEL it observes, not by the completeness of its list. An enumerated ledger fails on entries you did not know about; a channel-scoped guard fails on mechanisms that do not use that channel at all — and the second failure is invisible to any amount of care spent on the first. Ask git for the effective answer (`git config --show-origin --get <key>`) rather than validating inputs.

Credit: found by a peer session (devrc-08) building on this branch's levers; independently reproduced above.

### Running total of holes in this branch
1. **`GIT_WORK_TREE` / `GIT_DIR` / `GIT_COMMON_DIR` / `GIT_INDEX_FILE`** — 0 occurrences in the plugin. `GIT_WORK_TREE=<victim> git -C <innocent fixture> checkout -f HEAD` writes into the victim while repo identity resolves truthfully to the fixture.
2. **`GIT_CONFIG_COUNT`** — above.
3. **The `GIT_ALLOW_PROTOCOL` positive control was unsound** — see the previous comment.

The two levers remain sound and are carried into the superseding guard. This branch's runner layer is not being merged.

ZacxDev:
## Fourth correction: this PR's own CLAIM was wider than its reach — and it covers the incident's real damage

Measured independently by a peer session (devrc-08) and re-run by me. With **exactly this branch's four levers exported** — `GIT_CONFIG_GLOBAL` → throwaway, `GIT_CONFIG_SYSTEM=/dev/null`, `GIT_CONFIG_NOSYSTEM=1`, `GIT_ALLOW_PROTOCOL=file`, `GIT_TERMINAL_PROMPT=0` — against a sacrificial victim:

```
victim BEFORE: branch=VICTIM-BR  commits=1  bare=false
  git -C <victim> commit --allow-empty        rc=0
  git -C <victim> branch -m PWNED             rc=0
  git -C <victim> config core.bare true       rc=0
victim AFTER : branch=PWNED      commits=2  bare=true
```

Two of those three are outside what this PR claimed. **The third is not.**

This PR's stated goal is *"structurally impossible for the test suite to touch the operator's real git config or push to a real remote."* `config core.bare true` **is** a git config write — a **local** one — and it is the exact damage the incident reflog recorded on the operator's base clone (`core.bare = true` on a working clone, so `git status` fails there).

The guard covered the **global** config channel only. The claim said "the operator's real git config" with no qualifier.

**That is claim-wider-than-reach, in this PR's own body, written hours after I made precisely that criticism of a sibling PR's title.** It is the same defect in the same shape, and I did not see it in my own text until someone else measured it.

### Why this is the useful entry in the defect list

The three earlier holes were things the guard did not *know about* (`GIT_WORK_TREE`, `GIT_CONFIG_COUNT`, an unsound control). This one is different: the guard behaved exactly as designed, and the **description** of it was wrong. No amount of testing the implementation finds that — only reading the claim against the reach does.

### Running total of holes in this branch
1. **`GIT_WORK_TREE` / `GIT_DIR` / `GIT_COMMON_DIR` / `GIT_INDEX_FILE`** — 0 occurrences in the plugin; a retargeted write lands in the victim while repo identity resolves truthfully to the fixture.
2. **`GIT_CONFIG_COUNT`** — injects config for one invocation with no file write at all; the `GIT_CONFIG_GLOBAL` redirect cannot see it.
3. **The `GIT_ALLOW_PROTOCOL` positive control was unsound** — built on an unreachable host, so the no-lever arm could only ever fail.
4. **The claim covered local config writes it never blocked** — above.

This branch's runner layer is not being merged; the surviving guard is a sibling PR, which carries the two sound levers with attribution. The branch stays open as the record.

ZacxDev:
## Keeping this open — but it CANNOT merge as-is now that #683 has landed. Measured.

Context: after comparing all four guard PRs, the decision was to converge on the environment-lever architecture and close the git-interceptor family (#676, #689 — reasoning in the [#689 comment](https://github.com/innovation-upstream/devrc/pull/689#issuecomment-5382080578)). **#683 is now on `main` as `dfd2d203`.**

This PR is **not** a competitor to #683 — it is complementary, and that is why it stays open. The two close different levers on the same architecture:

| | #683 (merged) | this PR |
|---|---|---|
| which repo git lands in | `GIT_DIR` + 10 siblings, stripped | — |
| which config git writes | — | `GIT_CONFIG_GLOBAL` → throwaway |
| whether git can reach the network | — | `GIT_ALLOW_PROTOCOL=file` (https/ssh refused **by git**, exit 128) |

Those map onto the two phases of the 2026-08-21 incident: local corruption at 19:21 (#683's strip) and the remote push storm at 19:28 (**this PR's `GIT_ALLOW_PROTOCOL`**). #683 alone does not close the push half. That half is worth having.

## 🔴 But composed, they break — and neither is wrong on its own

I hand-resolved the merge (conflicts in `run-tests.sh` and `conftest.py` are purely additive co-location — keep both) and gated the stack on a standalone clone with `origin` removed. Three-way, same base `a34d695d`:

```
control  pristine main            pytest 14689 collected /  0 failed   GATE PASS
main + #683                       pytest 14721 collected /  0 failed   GATE PASS
main + #683 + this PR             pytest 14766 collected / 39 failed   GATE FAIL
```

The failure shape is diagnostic: **exactly one error in every one of the 26 pytest targets**, always at the teardown of whichever test ran first in that target (`.E...`). It is the same error every time:

```
DEVRC-GITENV-VIOLATION: test `…::<first test>` MUTATED a git repository that is not its own tmpdir.
What moved:
  CHANGED  /tmp/nix-shell.<x>/tmp.<y>/gitconfig   c2d6f3ff28f7 -> abf7b267f2bd
```

**The mechanism.** #683's detector fingerprints the user-level git config, and `global_config_paths()` resolves that by reading `GIT_CONFIG_GLOBAL` — which *this PR sets*, to its throwaway file. This PR's own positive control then performs a real `git config --global` write into that file, by design, because a control that does not move the number proves nothing. #683 sees the file it was told to protect change, and correctly fires.

So: **#683 is watching a file this PR legitimately writes.** Neither guard has a bug. Both were mutation-tested and audit-clean in isolation; the defect lives entirely in the seam, which is `claude/RULES.md` → *"Verified in isolation is the new vacuous green — the defect lives in the SEAM nobody owns."* Worth recording as an instance of exactly that.

The remaining 13 failures in `scripts/tests` are the two PRs' pinning tests: each pins the exact `python -m pytest …` line in `run-tests.sh` and the conftest import block, and my resolution changed both.

## What this PR needs before it can merge

1. **The seam.** Cheapest correct fix is on this side: have the throwaway `GIT_CONFIG_GLOBAL` live somewhere #683 is not fingerprinting, or make the control re-baseline #683's detector around its own deliberate write. Do **not** fix it by dropping `global_config_paths()` from #683's fingerprint — a `--global` write leaves the repo untouched, so that is the one damage class the ref fingerprint cannot see, and it is a class this incident actually produced (`core.hooksPath` in `~/.gitconfig`).
2. **The guard number.** Both PRs call themselves **GUARD 9**. The stack's own runner banner prints the collision back verbatim:
   ```
   run-tests: git repo pointers cleared for this run (GUARD 9)   <- #683
   run-tests: git isolated for this run (GUARD 9)                <- this PR
   ```
   One needs renumbering to GUARD 10, with its pinning tests updated.
3. **Rebase onto `dfd2d203`**, then re-gate the merged tree — not this branch. Its current base is `3116225d`, ~20 commits back and before #683 existed.
4. **Cost.** On the same box the stack took measurably longer on `scripts/tests` than #683 alone (>725s vs 699.6s vs a 687.7s control). Worth a number in the PR body, since this PR's per-target live controls (a real `git config --global` write and a real `git ls-remote` over https) are the plausible driver.

Nothing here argues against the approach — `GIT_ALLOW_PROTOCOL=file` is the cleanest thing anyone has proposed for the push half, and it is enforced by git itself rather than by a parser. It just cannot land on top of #683 untouched.

ZacxDev:
Pushed the reconciliation with **#683**, which landed as GUARD 9 while this was open. Three commits on top of your branch; the guard's mechanism is untouched.

**1. Renumbered this guard to GUARD 10.** #683 took GUARD 9 for the repo-pointer strip. The collision was not cosmetic — both emitted `GUARD 9` strings into the same run output, and `test_nogit_isolation.py` asserts `"GUARD 9 problem" in out`, so one suite's assertion could have matched the other's output and passed for the wrong reason. `run-tests.sh` was renumbered line-by-line by owner (its GUARD 9 lines stayed with gitenv); `nogit_plugin.py` and `test_nogit_isolation.py` are wholly yours and were renumbered throughout.

**2. Fixed a real integration defect that neither branch could see.** GUARD 9 fingerprints whatever `GIT_CONFIG_GLOBAL` names. This guard *points* `GIT_CONFIG_GLOBAL` at a scratch `gitconfig` under `DEVRC_TEST_GIT_GUARD_DIR` and lets tests write there — that file exists in order to be written. So GUARD 9 reported

```
DEVRC-GITENV-VIOLATION: CHANGED /build/tmp.../gitconfig
```

in **every** pytest target: `errors=1` across 25 targets, 39 failures. One guard calling the other's correct behaviour an incident. `gitenv.py` now skips a redirect the harness made into its own run dir, and watches the operator's real paths instead — strictly stronger than before, because the old code returned early on *any* override, so a direct write to `~/.gitconfig` (not via `--global`) went unwatched. The exemption is containment-checked against `DEVRC_TEST_GIT_GUARD_DIR`, not a name or a `/tmp` prefix. Three tests added in `test_git_repo_isolation.py`; the load-bearing one is RED before that commit and GREEN after.

**3. Admitted `gitenv_plugin` to `EXPECTED_PLUGINS`.** Your ledger went red because the set grew — exactly what you wrote it for. The review it demanded is the comment now beside it.

**Verification on the merged tree** (this branch = current `main` + yours + the three commits):

```
nix build .#checks.x86_64-linux.pytests   -> exit 0   RESULT: PASS
  TOTAL collected=14846 passed=14844 skipped=2 failed=0  (floor 13529)
nix build .#checks.x86_64-linux.nodetests -> exit 0   RESULT: PASS
```

Both guards now run together on the single pytest invocation, for all 26 targets.

ZacxDev:
**Merging with `tekton/devrc-pytests` red — deliberately, and here is the evidence it carries no information about this branch.**

The Tekton gate appeared while this PR was open and has never been green. Every completed `devrc-ci` PipelineRun fails on the *same* test, across unrelated revisions:

| PipelineRun | revision | failing test |
|---|---|---|
| `devrc-ci-vztwq` | `26a5ead3` | `test_timeout_reaps_the_whole_process_group` |
| `devrc-ci-ppz74` | `2b13a6e0` | `test_timeout_reaps_the_whole_process_group` |
| `devrc-ci-94prq` | `d7b31f2d` | `test_timeout_reaps_the_whole_process_group` |
| `devrc-ci-vq2qf` | `696ce717` (this PR) | `test_timeout_reaps_the_whole_process_group` |

It runs opencode under `setsid` and asserts the whole process group is reaped on timeout — behaviour that differs inside a container's PID namespace. It is in `scripts/browser-bridge/tests`, which this branch does not touch.

⚠ Worth knowing: the summary counts nearly matched my local run (`passed=14843 failed=1` vs `14844`/`0`), so reading the *summary* would have suggested Tekton had reproduced my last outstanding failure. It had not — a different single test failed, and the totals coincided. The step log, not the verdict line, is what distinguished them.

Authoritative verification, on the merged tree (this branch = current `main` + #673 + the three reconciliation commits):

```
nix build .#checks.x86_64-linux.pytests   -> exit 0  RESULT: PASS
  TOTAL collected=14846 passed=14844 skipped=2 failed=0  (floor 13529)
nix build .#checks.x86_64-linux.nodetests -> exit 0  RESULT: PASS
  TOTAL suites=4 files=33 tests=1119 pass=1119 fail=0   (floor 1098)
```

🔴 The Tekton failure is real work for someone, and it should be fixed before anyone relies on that gate: a check that has never passed trains people to click through it, which is worse than having no check.

PR #676, opened 2026-08-21
guard(tests): the audit found nine bypasses, a second audit found eight more spellings of the same nine

ZacxDev:
# guard(tests): no test may WRITE to a git repo it does not own — rework after a blind audit found nine bypasses

## Why this exists

The suite drove `git` against the operator's REAL clone and destroyed `main` on
the remote. The reflog ordering is the diagnosis:

```
trunk@{0}: Branch: renamed refs/heads/main to refs/heads/trunk   14:21:35 = 19:21:35Z
~40 pushes onto refs/heads/main in 43 seconds                              19:28:14Z
```

Local damage first, pushes seven minutes later. A push-scoped or remote-scoped
guard would have watched the whole first phase go past.

## 🔴 What the previous revision of this PR got wrong

A blind adversarial audit found **nine working bypasses**, and proved the
headline coverage number was a claim about the battery rather than about the
guard. A second independent audit then found **eight more** — every one a
DIFFERENT SPELLING of a capability the first round had closed. Both rounds are
in the battery now, and the second one is why the fixes below are structural
rather than shape-matched.

### The headline number was inflated ~4×

The old body claimed *"env negative — 12 blocked / 0 leaked, victim
byte-identical"*. The battery had **11** entries, and running each vector's
identical command **with the guard removed** gave:

| verdict | meaning | count |
|---|---|---|
| LIVE | the victim moved on an axis `_victim_state` asserted | **3** |
| WEAK | bytes landed in the victim, `_victim_state` was blind to them | 4 |
| VACUOUS | nothing moved anywhere — true with the guard DELETED | 4 |

The "blocked" half was genuine; the "victim byte-identical" half was
informative for 3 of 11.

## The rebuilt env battery — every vector proven LIVE

`_victim_state` is now a full byte manifest of the victim tree plus its ref
list and HEAD, so a vector that lands a byte anywhere is visible.
`test_every_env_ledger_entry_is_a_live_write_vector` **runs each ledger entry
with the guard removed and requires the victim to move** — a ledger entry that
cannot be armed fails the suite instead of being documented.

| variable | command (chosen because it is LIVE) | before | after |
|---|---|---|---|
| `GIT_DIR` | `commit --allow-empty` | LIVE | LIVE |
| `GIT_COMMON_DIR` | `config --local nogit.pwned 1` | WEAK | LIVE |
| `GIT_WORK_TREE` | `checkout -f HEAD`, fixture tree ≠ victim tree | VACUOUS | LIVE |
| `GIT_OBJECT_DIRECTORY` | `hash-object -w a.txt` | WEAK | LIVE |
| `GIT_INDEX_FILE` | `add a.txt` (the file now exists) | VACUOUS | LIVE |
| `GIT_CONFIG_GLOBAL` | `config --global user.name pwned` | WEAK | LIVE |
| `GIT_CONFIG_SYSTEM` | `config --system user.name pwned` | WEAK | LIVE |
| `GIT_ALTERNATE_OBJECT_DIRECTORIES` | — | VACUOUS | **REMOVED** |

**7 entries, 7 LIVE, 0 weak, 0 vacuous.**

Three of those commands are not the obvious one, and each replacement is a
measurement rather than a preference:

* `GIT_WORK_TREE` + `add -A` writes the objects and index into the FIXTURE and
  leaves the victim untouched — it was never a vector. `checkout -f HEAD` with
  a fixture whose tree differs clobbers the victim's `keep.txt`.
* `GIT_OBJECT_DIRECTORY` + `commit` dies `fatal: could not parse HEAD`, because
  with the object store redirected git cannot read the fixture's own HEAD. It
  *appeared* live only while the victim and the fixture shared a seed commit
  sha — identical author, message and empty tree, created in the same second.
  `_make_repo` now gives every fixture a distinct identity so that class of
  accident cannot come back, and the exposure of it immediately falsified
  `GIT_COMMON_DIR` too.
* `GIT_COMMON_DIR` + `branch` / `commit` / `tag` / `update-ref` / `gc` all fail
  for the same reason. `config --local` writes `<victim>/.git/config` with no
  object lookup at all — and it is the exact `config core.bare true` damage the
  incident's reflog recorded.

### `GIT_ALTERNATE_OBJECT_DIRECTORIES` was a NON-VECTOR pinned as correct

Armed at the victim with the guard removed, the commit's object went to the
**fixture**; a full byte manifest of the victim was unchanged. It is a
read-only *additional* object store. It sat on the ledger under that file's own
rule — *"EVERY ENTRY MUST BE A PATH THAT RECEIVES BYTES"* — and the test written
to prevent exactly this recurrence **pinned it as correct**, because it asserted
set membership instead of exercising the vector.

## The bypasses, all measured, all closed

Each is scored by a three-way harness: **NOOP** (build the fixtures, run
nothing — the victim must not move, so the probe cannot answer before the code
does), **UNGUARDED** (the identical argv with the shim off PATH — the victim
MUST move, or the case is vacuous and fails), **GUARDED**.

### Round one

| bypass | fix |
|---|---|
| `GIT_CONFIG_COUNT` → `core.hooksPath` (arbitrary code execution, victim-aimed) | injected config is parsed and path-valued keys are bounds-checked |
| `-c core.hooksPath=<victim>` | `-c` is COLLECTED, not skipped |
| `config --file` / `-f <victim>/.git/config` | `--file`, `--global` and `--system` destinations collected |
| `config --global` writing `$HOME/.gitconfig` unbounded | the reconstructed destination is bounds-checked |
| `-c remote.origin.url=<victim> push` | destination read off the ARGV, no child process |
| `GIT_CONFIG_COUNT` → `remote.origin.url` push | same |
| `-c remote.origin.pushurl=<victim> push` | `pushurl` consulted first, as git does |
| `-c 'alias.x=!git -C <victim> …'` | `GIT_EXEC_PATH` farm |
| `init`/`clone --separate-git-dir <victim>/x` | destination collected |
| `worktree add <victim>/wt` | destination collected |
| `DEVRC_TEST_GIT_MODE=audit` silently disarming everything | mode is a BAKE-TIME argument; no env-readable mode remains |
| `hash-object -w` into a denied repo | moved to the dual list |
| `archive --output=` / `bundle create` into the victim | destinations collected |
| `commit -m -h` (whole-argv read-flag scan) | flags scanned only BEFORE the verb |
| `GIT_EXTERNAL_DIFF` turning a permitted read into execution | refused when non-empty |
| `status` rewriting `<victim>/.git/index` | forwarded as `--no-optional-locks status` |

### Round two — the same capabilities, spelled differently

| bypass | why the round-one fix missed it |
|---|---|
| `--work-tree=<victim> clean -fdx` | the flag twin of an env variable that WAS closed |
| `--git-dir=<victim>/.git commit` | same |
| `--git-dir=<victim>` with an allowed `--work-tree` | the `top` chain judged one resolved path, not the SET |
| `branch --format=… -m main trunk` | a read-shaped flag placed first |
| `symbolic-ref -q HEAD refs/heads/trunk` | the verb was on the READ ledger and repoints with two args |
| `GIT_CONFIG_COUNT=+1` and `=' 1'` | git parses with `strtoumax`; an all-digits `case` does not |
| `-c diff.external=<script> diff` | the read fast path ran above the `-c` scan |
| `-c filter.x.clean=<git-into-victim> add` | a child-injection channel that is not an alias |
| `config --file ../victim/…`, `worktree add ../victim/wt`, `init --separate-git-dir=../victim/…`, `clone` into `../victim/cl` | an absolute-only matcher is relative-walkable |
| `archive --format=tar --output=<victim>/.git/config` | a second output spelling |
| `init <victim>/newone` | — |
| `-C <victim> -C .` and `-C <allowed> -C ../victim` | git applies multiple `-C` CUMULATIVELY |
| `bisect run <git-into-victim>` | held by the farm; the outer command legitimately succeeds |

**All 20 round-two probes accounted for: 18 CLOSED, 2 NOT-LIVE** — git itself
refuses `symbolic-ref -d HEAD`, and a `bisect run` with no range never executes
its script. Neither is a bypass.

## 🔴 The central architectural claim was FALSE, and is now closed

The old file said the shim was reached *"first on PATH … inherited by every
descendant, at any depth"*. Measured:

```
$ git -c 'alias.p=!echo PATH[0]=$(echo $PATH | cut -d: -f1)' p
PATH[0]=/nix/store/…-git-2.55.0/libexec/git-core
```

git **prepends its own exec-path** to `PATH` for every child it spawns, and
that directory ships a real `git`. `test_no_path_entry_before_the_shim_provides_a_git`
measured the *pytest process's* PATH and was structurally blind to it.

`exec_path_farm()` builds a directory of symlinks to all ~184 real helpers plus
the shim as `git`, and `GIT_EXEC_PATH` points at it — so git prepends a
directory this policy owns. The shim re-exports it on every call, so a child
built with a REPLACING `env=` cannot lose it.

**This closes more than the injection channel.** Measured, with no injection
involved at all:

| escape already present in a repo's own config | before | after |
|---|---|---|
| `alias.x = !git -C <victim> commit` in `.git/config` | victim commit landed | **refused, banner, victim untouched** |
| `.git/hooks/post-commit` running `git -C <victim> commit` | victim commit landed | **victim untouched** |
| `bisect run` script running `git -C <victim> commit` | victim commit landed | **victim untouched** |

### The stated limit, pinned as a measurement

`alias.x = !sh -c 'echo … > <victim>/X'` **still writes the victim**, and
`test_a_shell_alias_writing_the_victim_is_NOT_covered` asserts that it does —
so the limit is a fact the suite carries, and a future change that closes it
turns the test red and forces the docstring to be rewritten.

Exposed surface, enumerated by what actually happens to each channel — a
docstring naming a channel the code does not refuse is the failure this rework
kept finding:

* **Refused when non-empty**: `alias.*`, `core.fsmonitor`, `diff.external`,
  `filter.*.clean`/`.smudge`/`.process`, `*.textconv`, `core.pager`,
  `core.editor`, `core.sshCommand`, `credential.helper`, `gpg.program`,
  `merge.*.driver`, plus `GIT_EXTERNAL_DIFF`, `GIT_ASKPASS`,
  `GIT_PROXY_COMMAND`, `GIT_SSH`.
* **Bounded, not refused** (a real caller needs them): `core.hooksPath`,
  `init.templateDir` — the value must be inside an allowed root.
* **Not refused at all, named so the list is honest**: `GIT_SSH_COMMAND` (one
  legitimate caller, and listing it cost nine red tests), and `.git/hooks/*`,
  `rebase --exec`, `submodule foreach`, whose bodies are argv or on-disk files
  rather than an injection channel.

For everything in the last group the `git` such code spawns is still caught by
the farm. What is uncovered is the part of such a body that is not a `git`
invocation. No git guard can bound `echo > file`.

## Scoped, not blanket-refused

`scripts/analyze-service-index/commit.sh:247` legitimately exports
`GIT_CONFIG_COUNT=2` with `core.hooksPath`/`init.templateDir` aimed at an empty
`mktemp -d`, to NEUTRALISE hooks.
`scripts/task-spec-drafter/ticket-status:249` passes `-c core.fsmonitor=`
(EMPTY) as hardening. Both have positive controls in the battery and both pass:
a path-valued key is bounds-checked (the mktemp dir is in an allowed root), and
an empty command-valued key is a disarm, not an arm.

`GIT_SSH_COMMAND` was on the refuse list for one armed run and cost **nine red
tests** — `scripts/resume-state.sh:327` sets it around a `fetch`, and the
refusal surfaced four layers away as `[fetch failed; compared against refs
already on disk]`. It is off the list, with the reasoning recorded.

## The session marker reported healthy while the guard was absent

`_inherited_stub_dir()` adopted any directory named by
`DEVRC_TEST_GIT_STUB_DIR` that merely **contained a file called `git`**, then
skipped `install()`. Measured with that pointed at a plain passthrough script:
the write landed **rc 0, no banner**, the marker was written anyway, `1 passed`.

Now the shim answers a **content handshake** — `--devrc-nogit-handshake` prints
a fingerprint that is a hash of the policy's own generated body, so no foreign
binary and no older revision can answer it. A foreign shim **fails the
session**; a shim that is ours but scoped to another tree self-installs
instead (raising there cost nine red tests — several tests copy the repo and
run a NESTED pytest whose `repo_root()` is the copy). The marker carries
`via=self|inherited`.

## L2 was completely inert

`subprocess.Popen.__mro__` during a session was
`['_NoLaunchPopen', 'Popen', 'object']` — this policy's class was gone, an
absolute-path `git config` returned 0 with no banner, and **zero tests
referenced it**. Both plugins captured the pristine `Popen` at IMPORT time and
each subclassed *that*, so whichever fixture ran last discarded the other; each
teardown then restored pristine, discarding the survivor too.

Both now subclass whatever `Popen` is at PATCH time.
`test_both_launch_policies_survive_in_the_popen_chain` pins that both are in
the chain.

L2 is also narrowed to skip a `git` inside the session's own tmp tree —
without that it **replaced a test's own instrument with the guard**:
`test_ship_converge.py::_git_shim` builds a deliberately-sabotaged `git`, and
L2 rewrote the call so the sabotage never happened.

## B1 — NOT closed here, and the residual gap stated

`scripts/tests/conftest.py` is the only registration. Measured per target:
`scripts/tests` → rc 99; **`scripts/repo-cos/tests`**, which contains the
culprit `test_prescan.py::_init_clone`, → **rc 0, the write landed**;
`scripts/dl-router/tests` → rc 0. **1 target of 25, and not the culprit's.**

A single `scripts/conftest.py` **would** close it — measured, pytest auto-loads
it for all 13 sampled targets, directory and file targets alike, with no
`run-tests.sh` edit. It is **deliberately not added**, because the runner-layer
guard on `guard/nogit-test-isolation` already registers with `-p` on the single
pytest line and two plugins on one class is the collision measured above with
`Popen`. The three comments that claimed otherwise are corrected.

**This PR's guard therefore protects `scripts/tests` only.** Read
`testlib/nogit_plugin.py`'s REGISTRATION SCOPE header before assuming more.

🔴 **And a clean refusal log is not containment.** A bypass writes no ledger
line, so a per-target `refused=0` prints while everything lands. Every number
this policy reports is a claim about the calls it SAW.

## Also fixed

* `test_no_mutating_verb_is_on_the_read_ledger` parametrised 12 verbs **none of
  which was on the ledger** — vacuous by construction, and blind to
  `hash-object`, `status`, `symbolic-ref` and `interpret-trailers`, which were.
  It now exercises **every ledgered verb** against a denied repo and asserts the
  repo is byte-identical.
* Each verb moved off the ledger has a **reading-form** test, so the fix cannot
  trade a false negative for a permanently-red gate. `fsck` was refused outright
  on the first attempt; it is DUAL now (`--lost-found` writes).
* `assert _victim_state(v) == _victim_state(v)` — a tautology — compares against
  the state captured before.
* `assert refused == 6` pinned the fixture's own length; it compares against
  `len(steps)`.
* `for line in $info` word-split on the default IFS while every other list is
  colon-split; it is a line-wise read.
* `refuse` took no argv, so audit mode would have re-exec'd a bare `git`.

## Verification

* **197 tests** in the battery (the old body said 78; the file had 81).
* **7 env vectors**, all LIVE (the old body said 12; the file had 11).
* Every negative control proven live by a guard-removed probe, every probe
  validated by a no-op run.
* Positive controls kept and extended: a fixture driving its OWN repo via
  `GIT_DIR`+`GIT_WORK_TREE`, `analyze-service-index`'s `GIT_CONFIG_COUNT`,
  `-c core.fsmonitor=`, in-bounds `--separate-git-dir` / `worktree add` /
  `archive --output`, a one-argument `clone <denied>`, and the reading form of
  every dual verb.
* 🔴 **The unguarded controls have to unwind L1 *and* L2.** Removing the shim
  from PATH is not enough — L2 rewrites an absolute-path git by design, and on
  the first run of the rebuilt battery it silently re-guarded **13** "unguarded"
  controls, which the harness reported as *"this vector is VACUOUS"*. `_l2_off`
  exists for that and `_expect_closed` asserts the control was not itself
  refused.

### Overhead

Median of 40 runs, this shim vs the real binary, measured against the previous
revision with the identical script:

| case | before | after |
|---|---|---|
| reads (`rev-parse`, `ls-files`, `status`, `log`) | 2.90× | 3.46× |
| `config foo.bar 1` | 11.40× | 11.40× |
| `remote -v` (dual) | 12.34× | 10.60× |
| `commit --allow-empty` | 4.51× | **7.89×** |

Per-invocation cost is unchanged-to-better. `commit` doubled for a specific,
named reason: it spawns `git maintenance run --auto --no-quiet --detach`, and
that child now goes **through the shim** instead of escaping — measured at
4.60× on the real binary from `GIT_EXEC_PATH` alone, with `rev-parse` at 1.03×.
It is coverage, not waste.

Also measured and NOT taken: the two `rev-parse` calls cannot be merged into
one. `--show-toplevel` makes the combined call exit 128 in a BARE repo and
under `GIT_DIR=<bare repo>`, which would lose every candidate in exactly the
redirection shape this guard exists to catch.

### Mutation sweep — 25/25 killed, harness validated first

The harness is validated before any verdict is read: the unmutated tree is
green under the exact command (13 passed, rc 0), and a `-k` matching nothing
exits **5**, so "mutant killed" cannot be confused with "my filter was wrong".
`pytest` exit **1** specifically is required; `-q -q` is never used; every
mutant runs under `PYTHONDONTWRITEBYTECODE=1`; every anchor is asserted to
occur exactly once, so an ambiguous mutation is not scored.

The positive control (`refuse` replaced by `exec`) dies with 13 failures.

Four mutants SURVIVED the first sweep, and each was a real gap, not noise:

* **the push-destination mutant was not ISOLATED** — it removed only the
  `pushurl` arm while the very next line still read the injection off the
  argv, so it survived for a reason that said nothing about the claim. Rebuilt
  to replace the whole resolution chain with the pre-fix `remote get-url`; it
  now dies.
* **relative operands anchored to `base` instead of `cwdbase`** survived,
  because every relative case had `base == cwdbase`. Added
  `init --separate-git-dir=../victim/x` from a NESTED base, where the two
  anchors disagree and the wrong one resolves INSIDE the allowed root while the
  bytes land in the victim.
* **both L2 mutants** survived on fixture ordering. Added
  `test_the_popen_patch_composes_in_EITHER_order`, which applies both patches
  explicitly in both orders, and
  `test_a_git_inside_the_sessions_tmp_tree_is_NOT_redirected`, which asserts
  both directions of the exclusion.

## One more, found here rather than by an auditor

`canon_of` returned a path unchanged when neither it nor its parent could be
canonicalised — and a raw path is prefix-matched. Measured:

```
git init -q <allowed>/does-not-exist/deeper/../../../victim/planted
-> rc 0, and <victim>/planted was created.        (same for `clone`)
```

`does-not-exist` makes both `cd`s fail, the raw string comes back, and it
starts with `<allowed>/` — so it matches an allowed root while `realpath` puts
it inside the victim. This is the `git diff --quiet` shape from RULES: a
comparison against an operand that is not there reports SAME, not MISSING.

An unresolvable path containing `..` now canonicalises to a sentinel that
cannot match any root. A path with **no** `..` is still returned raw on
purpose, because `git init a/b/c` creates its intermediate directories and that
is its normal case. The deliberate cost —
`<allowed>/exists-later/../ok` is refused even though it resolves in bounds —
is pinned by its own test so a future reader meets it there rather than in a
confusing red run.

### DIFF
--- a/scripts/testlib/nogit.py
+++ b/scripts/testlib/nogit.py
@@ -46,17 +46,8 @@ one-for-one onto the reflog above (`config user.email t@t`, `commit -m seed`,
 well-formed `git -C <tmp clone> …`.
 
 So the target of a write is computed from the ARGUMENTS **and** the
-ENVIRONMENT, in two sub-classes that do not cover each other:
-
-  * IDENTITY redirection (`GIT_DIR`, `GIT_COMMON_DIR`) — git reports a
-    different repo. Caught by asking git which repo it will actually use.
-  * BYTE redirection (`GIT_WORK_TREE`, `GIT_INDEX_FILE`,
-    `GIT_OBJECT_DIRECTORY`, the alternates, `GIT_CONFIG_GLOBAL/SYSTEM`) —
-    identity stays innocent and TRUTHFUL and only the destination moves.
-    `GIT_WORK_TREE=<victim> git -C <fixture> checkout -f HEAD` exits 0 and
-    writes the victim's working tree. **Resolving the repo cannot see this.**
-
-See `GIT_LOCATION_ENV` for the ledger and the two entries that were wrong.
+ENVIRONMENT, in sub-classes that do not cover each other. `GIT_LOCATION_ENV`
+carries the ledger and the measurement that decides membership.
 
 🔴 IT IS INVISIBLE BY CONSTRUCTION
 -----------------------------------
@@ -66,26 +57,37 @@ after-the-fact check reports a clean repo. A guard that inspects state after
 the run therefore cannot see this class at all — the interception has to happen
 at the moment of the call, which is what this module does.
 
-WHY A PATH SHIM AND NOT A `subprocess` MONKEYPATCH
---------------------------------------------------
-The same reason `testlib/nolaunch.py` is a PATH shim, and it is decisive here
-rather than merely tidy: a `git` that damages something is very often run by a
-GRANDCHILD. The suite spawns shell scripts, and those scripts run `git`; a
-`subprocess.Popen` patched inside the pytest process never sees any of it. Only
-a shim that is first on `PATH` — inherited by every descendant, at any depth —
-can. The in-process layer exists too, mirroring `nolaunch_plugin`'s L2, but
-purely to catch an ABSOLUTE-path `git` that PATH cannot shadow.
+🔴 WHY A PATH SHIM, AND WHY PATH ALONE IS NOT ENOUGH
+-----------------------------------------------------
+A `git` that damages something is very often run by a GRANDCHILD. The suite
+spawns shell scripts, and those scripts run `git`; a `subprocess.Popen` patched
+inside the pytest process never sees any of it. Only a shim inherited through
+the environment can.
 
-⚠ An earlier revision of this paragraph named `scripts/drift-check.sh`'s
-`ssh … bash -s` leg (where `DRIFT_REPO` is unset and
-`${DRIFT_REPO:-$HOME/workspace/devrc}` resolves to the real clone) as the
-motivating caller. That theory is RETRACTED and should not be re-derived: the
-test harness's stub `ssh` writes the payload to `/dev/null` and never executes
-it, and the payloads are read-only regardless. The empty-override sites in
-`drift-check.sh` and `ship.sh` are a genuine latent hazard — all five use `:-`,
-which falls back on EMPTY as well as unset — but no caller passes an empty
-value, so they were not the mechanism. The PATH-shim argument stands on the
-grandchild point alone.
+⚠ An earlier revision of this file claimed that being "first on PATH" made the
+shim unavoidable "at any depth". **That claim was FALSE and is now closed by
+`exec_path_farm()` — do not re-derive it.** MEASURED:
+
+    $ git -c 'alias.p=!echo PATH[0]=$(echo $PATH | cut -d: -f1)' p
+    PATH[0]=/nix/store/…-git-2.55.0/libexec/git-core
+
+git PREPENDS its own `libexec/git-core` to `PATH` for every child it spawns,
+and that directory contains a real `git` binary. So every alias, hook, filter,
+`core.fsmonitor`, `GIT_SSH_COMMAND`, `rebase --exec` and `submodule foreach`
+child resolved `git` to the REAL binary no matter where the shim sat:
+
+    $ git -C <fixture> -c 'alias.x=!git -C <victim> commit …' x
+    rc 0, and the victim received the commit.
+
+The fix is to make git prepend a directory THIS module owns: `exec_path_farm()`
+builds a directory of symlinks to every real `git-*` helper plus the shim as
+`git`, and `GIT_EXEC_PATH` points at it. Measured through the farm, PATH[0]
+inside an alias is the farm and `git` resolves to the shim. See
+`_exec_path_export()` for why the shim re-exports it on every call.
+
+The in-process layer (`nogit_plugin._patch_subprocess`) exists to catch an
+ABSOLUTE-path `git` launched by the pytest process, which no PATH-based
+mechanism can shadow.
 
 🔴 WHY THE SPLIT IS BY VERB **AND** BY TARGET, AND WHY IT FAILS LOUD
 ---------------------------------------------------------------------
@@ -125,13 +127,36 @@ those was found by a POSITIVE control; none was visible to a negative one.
 A `push` from a repo that IS inside the tmpdir still reaches whatever its
 `origin` points at. That is not hypothetical here: the contained clone prepared
 for this work had an `origin` pointing at the operator's real local clone, so a
-fixture pushing to `origin` would have written into their actual refs. Location
-of the *source* repo therefore does not bound the blast radius of a push, and
-`_push_destination_body` resolves the destination URL and applies the same
-root check to it.
+fixture pushing to `origin` would have written into their actual refs.
+
+🔴 THE THIRD VECTOR: CONFIG INJECTION MOVES THE DESTINATION WITHOUT NAMING IT
+------------------------------------------------------------------------------
+`-c key=value`, `--config-env=key=VAR` and the `GIT_CONFIG_COUNT` /
+`GIT_CONFIG_KEY_<n>` / `GIT_CONFIG_VALUE_<n>` family all inject configuration
+for one invocation. MEASURED, against the revision of this file that ignored
+them entirely:
+
+    GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=core.hooksPath \
+      GIT_CONFIG_VALUE_0=<victim>/hooks \
+      git -C <fixture> commit --allow-empty -m hookprobe
+    -> rc 0, ALLOWED, the hook EXECUTED, artefact written inside the victim.
+
+    GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=remote.origin.url \
+      GIT_CONFIG_VALUE_0=<victim>   git -C <fixture> push origin HEAD:pwned
+    -> rc 0, ALLOWED, `refs/heads/pwned` created in the victim.
+
+It is NOT blanket-refused, and that is the whole design constraint:
+`scripts/analyze-service-index/commit.sh:247` legitimately exports
+`GIT_CONFIG_COUNT=2` with `core.hooksPath` / `init.templateDir` aimed at an
+empty `mktemp -d`, precisely to NEUTRALISE hooks, and two tests pin it. So
+injected keys are classified — see `GIT_CONFIG_PATH_KEYS` and
+`GIT_CONFIG_CMD_KEYS` — and a path-valued key is bounds-checked like any other
+destination. The mktemp dir is under `$TMPDIR` and passes; the victim-aimed one
+does not.
 """
 from __future__ import annotations
 
+import hashlib
 import os
 import shutil
 from pathlib import Path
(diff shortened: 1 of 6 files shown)


ZacxDev:
## Closing in favour of #683 — same architecture call as #689

This PR and #689 are the same architecture (intercept every `git` dispatch name and parse its argv to decide the target); #683 is the alternative (strip the environment variables that redirect git, plus a fingerprint detector). After measuring all four guard PRs against current `main`, the decision is to converge on #683. The full reasoning is in the [#689 closing comment](https://github.com/innovation-upstream/devrc/pull/689#issuecomment-5382080578); the short version, specific to this PR:

- The **seven measured bypasses** documented in the handoff for this branch are not seven bugs, they are the visible part of an unbounded surface. This PR's own title says it: *"the audit found nine bypasses, a second audit found eight more spellings of the same nine."* 🔴-6 is the tell — `collect_extra_dests` covers 4 verbs and misses `checkout-index --prefix`, `pack-objects` and `diagnose -o`. Enumerating git's destination-bearing options is a maintenance commitment that grows with git.
- 🔴-4 is the strongest argument against the shape rather than this instance: `git --config-env=core.pager='V:-$(touch X)' --version` executed inside the shim, on a pure read, before any check — where real git exits 128 and runs nothing. **On that path the guard was strictly worse than no guard.** Fixable, and #689 fixed it, but it is what 1,758 lines re-implementing git's argv grammar looks like when it goes wrong.
- The measured overhead — 8.85× on guarded writes, driven by three real-git processes plus 10–18 forks — is the price of resolving targets correctly, and it does not buy closure.

Measured, standalone clone with `origin` removed, base `a34d695d`:

```
control  pristine main       pytest 14689 collected / 0 failed   GATE PASS
main + #683                  pytest 14721 collected / 0 failed   GATE PASS   (+1.7% wall clock)
```

#683's guard marker prints from **26 of 26** pytest targets.

### What survives from this branch

The two findings this PR established are recorded in the analyze-service index under `devrc/tests`, as patterns rather than instances:

- **Key a `git-*` exec-path farm on the DISPATCH NAME, never the inode** — 181 dispatch names vs 146 inode matches on this host, `PATH`-dependent within one machine. Correct and non-obvious, and it stays true whatever guard ships.
- **Audit every `eval` of an argv-derived value**, not the one that bit us.

Note that #689 later **retracted** this branch's claim that the exec-path farm closed the alias/hook residual — the farm substituted only the literal name `git` while git dispatches on `argv[0]`, so all 181 entries walked past it. If any of this branch's approach is revived, start from that retraction.

Also relevant: this branch's decision to leave B1 (per-target coverage) to #689's runner layer no longer has a counterpart, since #689 is closed too. #683 covers all 26 pytest targets from the single `python -m pytest` line in `run-tests.sh` plus the three shell runners, so that residual is closed by the replacement rather than left open.

Reopen if you disagree with the architecture call; the branch is untouched.

ZacxDev:
**Not merged — this needs a design call that is yours, not a merge resolution.** The reconciliation work is pushed to `integ/step3-676` so none of it is lost.

### What I did (mechanical, and correct as far as it goes)

GUARD 10 (#673) landed first under the filename this branch also chose, so on the merged tree:

* `testlib/nogit_plugin.py` → `testlib/gitwrite_plugin.py`, `testlib/nogit.py` → `testlib/gitwrite.py`. `main` keeps #673's module at the contested name, byte-unmodified.
* 🔴 **`SESSION_MARKER` had to move too.** Both modules spelled it `nogit(session)`, and GUARD 10's per-target accounting fails any target emitting other than exactly ONE marker — so renaming only the *file* would have left each guard reading as the other's double-registration. This one is easy to miss.
* Numbered GUARD 11 in the conftest. Registration scope left exactly as you designed it (conftest only, not run-tests.sh's `-p` line) — broadening it from 1 target to 26 is a real improvement and a real risk, and it is your call, not a merge side effect.

Your own suite is green after the rename: **197 passed**.

### Why it still cannot land

```
  FAIL scripts/tests  collected=7384 passed=7344 failed=40
```

The root cause, from `test_the_stub_dir_is_first_on_path`:

```
AssertionError: the stub dir must be the FIRST PATH entry; PATH starts with
  ['/build/pytest-of-nixbld/pytest-0/gitwrite0', '/build/tmp.4Jsl9x0clD', ...]
```

**GUARD 7 (`nolaunch`) asserts its stub dir is `PATH[0]`. This guard prepends its own shim dir.** Both are right in isolation and both cannot be index 0. Seven failures name `gitwrite0` directly; the rest are the downstream end-to-end suites (the `_named_and_red` mutants, the gate tests) that spawn the runner through the patched `subprocess`.

This was invisible on your branch: GUARD 7's assertion passed there because the session-fixture ordering differed before GUARD 9 and GUARD 10 were also registered in that conftest.

### The two ways out, as I see them

1. **Share one stub dir.** GUARD 7 already owns a directory that is first on PATH; put the `git` shim in it instead of a second directory. One PATH entry, no ordering contract, and GUARD 7's assertion stays as strong as it is now. This looks right to me but it is a change to how your shim installs.
2. **Relax GUARD 7 from `PATH[0]` to "before any real launcher".** Cheaper, but it weakens a guard whose docstring specifically argues for `FIRST, not merely present`, and it would need its own reasoning about the sandbox tier where no real launcher exists to shadow.

`GIT_EXEC_PATH` is unaffected either way — that part of your design is orthogonal and stands.

PR #678, opened 2026-08-21
prune(browser): #669 landed a correction with no eviction — main went red at 29 B of headroom

ZacxDev:
## `main` is red; this unbreaks it

`scripts/browser-bridge/tests/test_skill_size.py::test_skill_md_keeps_working_headroom` has been failing on `origin/main` since `da804c9f` (#669). `SKILL.md` sat at **12,259 B** with **29 B** of headroom against the required margin the test owns. Nothing caught it — devrc has no automated merge gate (`gh pr view --json statusCheckRollup` returns 0 checks).

The regression is not *what* #669 said. That commit fixed a real factual error (the civitai account switcher EXISTS; the notes said it did not). It is that it added to a byte-capped always-on doc **without an eviction in the same commit**, which the project `CLAUDE.md` requires.

🔴 `MAX_BYTES` / `MIN_HEADROOM_BYTES` are **untouched**. Raising a ceiling to green a red test defeats the gate; this body loads on every browser task.

## What was evicted, and where it went

Every byte of #669 is **kept** — trap 4's second-view half, and the `sites/<host>.md` row's read-it-BEFORE-the-first-action rule. Older, colder and already-duplicated material went instead, following the eviction playbook the test itself prints.

**Demoted to `reference/` (content preserved, one routing pointer left behind):**

| what | new home |
|---|---|
| env defaults `$BB_INSTANCE`/`$BB_TAB`/`$BB_FRAME`: the zsh no-word-split trap, the ONCE-on-stderr route notice, `BB_NO_ROUTE_NOTE`; plus the `health`/`instances` entry fields | `reference/tabs-instances.md` |
| `js` vs `eval` — the whole worktree-isolation-guard explanation (core keeps the imperative) | `reference/errors.md` |
| `screenshot` output modes: `<path>` prints a bare path, `--json`, `--data-url`, and the two **refused** flag pairs | `reference/read-envelopes.md` |
| `activate`'s `--focus`/`withheld`/`skipped` mechanics, incl. the stdout-is-a-TTY default | `reference/spa-wake.md` |

**Dropped for a pointer, because the destination already carried the text:** `upload` reading the file BY PATH (`security-ops.md`), `text --annotated` working with `--frame` and the `chrome://` `null` (`read-envelopes.md` / `errors.md`), the visibilityState-spoof clause and restore-the-workspace-explicitly (`spa-wake.md` — verbatim there).

No new files, so nothing new needs `git add` for the flake to see it.

## Verification

**Every relocated claim was re-verified against `scripts/browser-bridge/browser`**, not paraphrased from the core: the stdout-is-a-TTY focus default, the mutually-exclusive `--data-url` pairs, `--instance ''` clearing an inherited env value, the guard matching the literal WORD `eval`. No stale claim found.

The matrix is inverted from usual — the test was **already red on `origin/main`**, which is its own negative control:

| | `test_skill_size.py` |
|---|---|
| `origin/main` (inherited) | **1 failed, 3 passed** — `RECLAIM: 221 bytes` |
| HEAD | **4 passed** |

Plus `scripts/tests/test_doc_path_rot.py` — **76 passed**, every routing pointer resolves.

```
before   12,259 B   headroom  29    (needs 250)
after    11,779 B   headroom 509    (-480 B)
```

Deliberately well past the 221 B minimum, and under the pre-#669 size of 11,807 B. Landing on the floor exactly would red `main` again on the next one-line correction — which is the failure this floor exists to *precede*. 509 B is ~2.7 mean ops rows of room.

### DIFF
--- a/scripts/browser-bridge/SKILL.md
+++ b/scripts/browser-bridge/SKILL.md
@@ -35,42 +35,41 @@ order) · **secret** — agent-read pages go to
 **OpenRouter/DeepSeek**: never banking, private mail, credential managers, or
 anything you wouldn't hand a third party. Nor **virtualised/lazy-loaded lists**.
 
-KNOW something from it → agent. Ambiguous → agent first: taking over is cheap, so
-agent-first wins even at a low success rate.
+Ambiguous → agent first: taking over is cheap, so agent-first wins even at a low
+success rate.
 
-🔴 The AGENT's auto-`wake` covers a hidden `text`/`html` read but **never `eval`/`js`** — so
+🔴 The AGENT's auto-`wake` covers a hidden `text`/`html` read but **never `js`** — so
 ASSERT a non-zero content count in any `js` measurement. Result handling
-(`blocked`/`partial`), why thin `evidence` is NOT protection, and the
-`--allow-domains` guardrail → `reference/agent.md`.
+(`blocked`/`partial`), why thin `evidence` is NOT protection, `--allow-domains`
+→ `reference/agent.md`.
 
 ## Ops
 
 Global flags, usable before any op: `--instance <key>` (which profile),
 `--tab <id>` (explicit tab), `--frame <numericId|urlSubstring>` (inside an iframe).
-Env defaults `$BB_INSTANCE`/`$BB_TAB`/`$BB_FRAME` (flag wins; zsh does NOT
-split an unquoted `$F`). ⚠ an export outlives the call — env-routed ops say so
-once, on stderr.
+Env defaults `$BB_INSTANCE`/`$BB_TAB`/`$BB_FRAME` (flag wins) — ⚠ an export
+outlives the call → `reference/tabs-instances.md`.
 Result payloads land under `.result.data`.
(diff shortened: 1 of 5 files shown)


ZacxDev:
Superseded by #679, which fixed the same defect concurrently and merged first. `main` is green: SKILL.md is 11,961 B with 327 B headroom against the 250 B floor.

Closing rather than rebasing — this branch conflicts with #679 in the same file, and the only delta is extra slack (this branch reached 509 B of headroom vs 327 B). Not worth the churn while green.

If the headroom gets tight again, the evictions here are still valid and recoverable from this branch: env defaults (`$BB_INSTANCE`/`$BB_TAB`/`$BB_FRAME`) → `reference/tabs-instances.md`; the `js`-vs-`eval` worktree-guard explanation → `reference/errors.md`; `screenshot` output modes → `reference/read-envelopes.md`; `activate`'s `--focus`/`withheld`/`skipped` mechanics → `reference/spa-wake.md`.

PR #683, opened 2026-08-21
fix(tests): the suite built its git fixtures inside the REPO IT RUNS FROM — `main` was rewritten and pushed to the remote

ZacxDev:
## What happened

Measured **2026-08-21**, on the operator's real `~/workspace/devrc` clone **and on the production GitHub remote**. A gate run:

- rewrote `refs/heads/main` with fixture commits — `seed`, `base`, `ahead`, `local side`, `un-pushed work stranded on main`, `autocommit: N change(s) in the some-scope analyze-service index`;
- created the fixture branches `side`, `topic`, `trunk`, `master`, `only-branch`, `feat/behind-too`;
- **DELETED `refs/heads/main`** and repointed `HEAD` at `trunk`;
- wrote `core.bare=true`, `user.name=T`, `user.email=t@example.invalid`, a `core.hooksPath` under `pytest-0/test_install_does_not_depend_o0/` and a `remote.origin.url` under `pytest-0/test_fetch_failure_is_rc40/`;
- and **pushed fixture refs to GitHub**, force-overwriting `main`.

## Root cause

🔴 **Not one fixture was sloppy.** Every git fixture in this repo passes `-C <tmp_path>/…`; several also pin `HOME`, `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM`. **`GIT_DIR` overrides `-C`**, so a single inherited environment variable defeats all of them at once. Reproduced exactly on a throwaway clone:

```
export GIT_DIR=<clone>/.git
git -C <tmp>/work checkout -b topic   ->  creates `topic` in <clone>
git -C <tmp>/work branch  -D  main    ->  DELETES <clone>'s main
git -C <tmp>/work config user.name T  ->  writes <clone>/.git/config
```

The damage itself is what identifies the mechanism rather than merely being consistent with it: the tmpdir paths written into the real `.git/config` are the tests' **own correct values**. They computed the right thing and git wrote it into the wrong repository.

Every fixture builds its subprocess environment as `dict(os.environ)` plus a few overrides, so any repo pointer in the caller's environment passes straight through and wins over `-C`. The repo already knew the rule — `scripts/claude-hooks/guard_core.py` judges a `GIT_DIR=` prefix *in addition* to a resolving `-C`, "even though `GIT_DIR` OVERRIDES the working directory". It lived in the hook that polices the operator's commands and nowhere in the harness that runs 14,000 of our own.

## The fix — GUARD 9, two layers

**PREVENTION.** `scripts/testlib/gitenv.py::REPO_POINTER_VARS` owns the eleven variables that decide *which* repository a git command lands in. Each of `run-tests.sh`, `run-node-tests.sh` and `gate.sh` unsets them at the **top** of the file, and `testlib/gitenv_plugin.py` strips them from `os.environ` at **import** — so every fixture that copies `os.environ` inherits the sanitised mapping without knowing this module exists, and the five non-pytest targets (HOOK_TESTS, SHELL_TESTS) plus the node tier are covered by the shell half.

**DETECTION.** The plugin fingerprints the host repository's `refs/`, `HEAD`, `packed-refs`, `ORIG_HEAD`, `logs/HEAD` and `config` (plus the user-level git config, and a worktree's *common* dir) and compares around collection, around **every test**, and at session end — failing the test that moved something, with the ref named. Prevention closes the mechanism we measured; detection is what catches the next one, by any mechanism, and names the culprit. Neither alone is enough: a strip with no detector is a claim, and a detector with no strip is an incident report.

`githooks/pre-push` assigned to the literal name `GIT_DIR`. In bash, assigning to an already-exported name **stays exported**, so that line is a concrete route from `git push` to a poisoned suite (`pre-push` → `tests-on-push.sh` → `run-tests.sh` → pytest). Renamed to `HOOK_GIT_DIR`, and a scan now asserts no shell script that can reach the runner assigns to one of those names.

### Two design decisions that were measured, not chosen

1. **The clear must precede each runner's `ROOT=` line.** All three resolve their root with `git -C "$(dirname "$0")" rev-parse --show-toplevel`; with `GIT_DIR` set and no `GIT_WORK_TREE`, git takes the CWD as the work tree, so that returns `<repo>/scripts` and the runner hunts for `<repo>/scripts/scripts/run-tests.sh` — `exit 127`, **no verdict line**. The first placement was beside GUARDS 7/8, two thirds down `run-tests.sh`, and the gate never got there. Same class as the `unset CDPATH` already documented a few lines below that ROOT block. The ordering is now pinned per runner.
2. **The `unset` is spelled three times rather than sourced from one file.** The DRY version put it in `scripts/lib/git-repo-pointers.sh` — and turned **fifteen tests red** with `run-tests: FATAL — cannot source lib/git-repo-pointers.sh`, because `testlib/runner_patch.py` writes a patched **copy** of `run-tests.sh` into a tmp dir and a copy cannot source a sibling `lib/` that was never copied with it. A guard that FATALs in every copy is a permanently-red gate. So the **set** is owned once, in Python, and all four spellings are pinned to it in both directions — the same treatment `SPOOL_SESSION_MARKER` gets, for the same cross-process reason. A copied runner now carries its own guard, which the sourced version never did.

## Before / after — same two test files, same poisoned `GIT_DIR`, throwaway clone

| | `origin/main` (3116225d) | this branch |
|---|---|---|
| result | **165 passed, 230 ERRORS** | **395 passed, 0 errors** |
| `refs/heads/main` | advanced to a fixture commit `base` | **byte-identical** |
| `git config --local --list` | gained `commit.gpgsign=false`; `user.name`/`user.email` overwritten with `Test Runner` / `test@example.invalid` | **byte-identical** |
| branch list / `HEAD` | — | **byte-identical** |

An earlier run of the same shape (stopped at the first failure) put `local side`, `ahead`, `base` and `commit the workbench never pushed` onto `main` and rewrote `refs/remotes/origin/main` — i.e. the reported evidence, character for character.

## Mutation-verified 12/12

Each mutant killed by its **own named assertion**, not by a neighbour's error; an in-batch positive control (strip `VIOLATION_TOKEN` from the message) is included, and the sweep ran under `PYTHONDONTWRITEBYTECODE=1` with the control green before and after.

| mutant | killed by |
|---|---|
| strip disabled | `…_leaves_the_ambient_repo_alone` — *ambient branches: ['main', 'topic']* |
| per-test check removed | `…_fails_the_test_that_moved_a_ref` |
| collection check removed | `…_catches_damage_done_at_IMPORT_time` |
| worktree common dir ignored | `…_covers_a_WORKTREE_via_its_common_dir` |
| deletions invisible | `…_sees_a_branch_DELETION` |
| `pre-push` reverts to `GIT_DIR=` | `…_assigns_a_git_repo_pointer` |
| runner drops `-p testlib.gitenv_plugin` | `…_every_pytest_target_gets_the_detector` |
| shell ledger loses one name | `…_pointer_ledgers_agree` |
| gate.sh clears **after** ROOT (reordered) | `…_clears_BEFORE_it_resolves_its_root[gate.sh]` |
| node runner drops the guard | `…_clears_BEFORE_it_resolves_its_root[run-node-tests.sh]` |
| conftest entry point removed | `…_is_the_second_entry_point` |
| **positive control:** token dropped | `…_fails_the_test_that_moved_a_ref` (*"failed, but not with GUARD 9's message"*) |

Two guards were rewritten after a mutant SURVIVED them:

- the conftest check grepped for the strings `testlib.gitenv_plugin` and `_devrc_git_repo_isolation`; a mutant that moved the import under `if False:` survived — both words were still on the page. It now asserts the **module object's attributes are the plugin's own objects**. (RULES: *a guard can be SPELLED rather than STRUCTURAL*.)
- the ordering test's parser matched any mention of `rev-parse --show-toplevel` and so matched GUARD 9's own explanatory **comment**, reporting all three runners mis-ordered while they were correct. It now skips comments and requires `ROOT=` on the line.

## Gate

```
env GIT_DIR=<tree>/.git nix develop . --command bash scripts/gate.sh --tier both --set hermetic
```

Run with a **poisoned `GIT_DIR` deliberately exported** — the end-to-end negative control, on the rebased tree (`29d7913e` + these two commits):

```
pytest: TOTAL collected=14401  passed=14399  skipped=2  failed=0   (floor 13324)
        every one of the 31 targets PASS
node:   TOTAL suites=4 files=33 tests=1119 pass=1119 fail=0        (floor 1098)
GATE: RESULT=FAIL exit=1
```

…and afterwards the tree's `HEAD`, branch list, `git show-ref` and `git config --local --list` were **byte-identical to the pre-run baseline**.

🔴 **The `FAIL` is not a test failure and is not from this branch.** `failed=0` everywhere; the gate is red on GUARD 6's unpinned-skip accounting, for a skip that arrives with #657:

```
scripts/signal/tests/test_pg_type_compat.py:91: needs a real Postgres (SIGNAL_PG_DSN);
  SQLite cannot reproduce Postgres static type checking
```

Measured on a clean detached checkout of `github/main` @ **29d7913e**, untouched by this branch: that skip is emitted (`624 passed, 1 skipped`) and `EXPECTED_SKIPS` contains only the `repo-cos` entry. `EXPECTED_SKIPS` is an exact set, so **`main` is currently red on that guard on its own** — the same shape the comment under `EXPECTED_SKIPS` records happening before (#332, red for two days). It needs a one-entry pin, in its own PR; it is deliberately not bundled here.

## Scope this does NOT cover

- The **detector** watches the repository the suite runs from and the user-level git config. A test that writes to some *other* repo by absolute path is caught only if that repo is one of those.
- The **strip** cannot stop a test that names a repo by absolute path; that is the detector's job, and the reachability control drives exactly that case.
- `TARGET_FLOORS` is untouched — the 60 added tests keep every target above its floor (`scripts/tests` 6887 vs floor 6459).

### DIFF
--- a/githooks/pre-push
+++ b/githooks/pre-push
@@ -41,14 +41,25 @@ URL="${2:-}"
 STDIN_DATA="$(cat || true)"
 
 REPO_ROOT="$(git rev-parse --show-toplevel 2>/dev/null || true)"
-GIT_DIR="$(git rev-parse --git-dir 2>/dev/null || true)"
+# 🔴 NOT NAMED `GIT_DIR`, AND THE NAME IS THE WHOLE POINT. In bash, assigning to
+# a variable that is ALREADY EXPORTED keeps it exported — so if any caller (an
+# outer hook, a wrapper, a shell where someone exported it) has GIT_DIR in the
+# environment, `GIT_DIR=…` here silently hands this repository's git dir to
+# every child, including `tests-on-push.sh` -> `run-tests.sh` -> pytest. GIT_DIR
+# OVERRIDES `-C`, so the whole devrc suite would then build its fixtures inside
+# THIS checkout: that is the 2026-08-21 incident, in which `refs/heads/main` was
+# rewritten with fixture commits and pushed. `scripts/testlib/gitenv.py` closes
+# it from the suite's side; this rename closes the route.
+# `scripts/tests/test_git_repo_isolation.py` asserts no shell script that can
+# reach the runner assigns to one of those names.
+HOOK_GIT_DIR="$(git rev-parse --git-dir 2>/dev/null || true)"
 SELF="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/$(basename "${BASH_SOURCE[0]}")"
 
 # --- 1. Chain to a repo-local pre-push hook, if any (and it isn't us) --------
 # Candidates: the repo's .git/hooks/pre-push and (defensively) a tracked
 # .githooks/pre-push. Skip any path that resolves to THIS file (no recursion).
 for candidate in \
-  "${GIT_DIR:+$GIT_DIR/hooks/pre-push}" \
+  "${HOOK_GIT_DIR:+$HOOK_GIT_DIR/hooks/pre-push}" \
   "${REPO_ROOT:+$REPO_ROOT/.githooks/pre-push}" \
 ; do
   [ -n "$candidate" ] || continue
--- a/scripts/gate.sh
+++ b/scripts/gate.sh
@@ -95,6 +95,30 @@ case "$TIMEOUT" in
   ''|*[!0-9]*) echo "gate: FATAL — --timeout must be a whole number of seconds, got '$TIMEOUT'" >&2; exit 2 ;;
 esac
 
+# --- GUARD 9: NO TEST MAY OPERATE ON THE REPO THE SUITE RUNS FROM -------------
+# 🔴 BEFORE the ROOT block below, not after it: with GIT_DIR set and no
+# GIT_WORK_TREE, git takes the CWD as the work tree, so
+# `rev-parse --show-toplevel` returns `<repo>/scripts` and this script dies
+# `exit 127` with no verdict. The 2026-08-21 incident, the reproduction, and the
+# reason this block is spelled in three files rather than sourced from one are
+# in `scripts/run-tests.sh`'s copy of this header. The SET is owned by
+# `scripts/testlib/gitenv.py::REPO_POINTER_VARS`; every spelling is pinned
+# two-way by `scripts/tests/test_git_repo_isolation.py`.
+DEVRC_GIT_REPO_POINTERS=(
+  GIT_DIR                            # the repository itself; beats -C
+  GIT_WORK_TREE                      # the working tree
+  GIT_COMMON_DIR                     # where refs/config actually live
+  GIT_INDEX_FILE                     # the index a `git add` writes
+  GIT_OBJECT_DIRECTORY               # where new objects are written
+  GIT_ALTERNATE_OBJECT_DIRECTORIES   # extra object stores
+  GIT_NAMESPACE                      # the ref namespace refs land in
+  GIT_PREFIX                         # hook-injected pathspec prefix
+  GIT_GRAFT_FILE                     # repo-scoped grafts
+  GIT_SHALLOW_FILE                   # repo-scoped shallow list
+  GIT_CONFIG                         # legacy: the file `git config` WRITES
+)
+unset "${DEVRC_GIT_REPO_POINTERS[@]}"
+
 if [ -z "$ROOT" ]; then
   ROOT="$(git -C "$(dirname "${BASH_SOURCE[0]}")" rev-parse --show-toplevel 2>/dev/null || true)"
   [ -n "$ROOT" ] || ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
--- a/scripts/run-node-tests.sh
+++ b/scripts/run-node-tests.sh
@@ -117,6 +117,30 @@ while [ $# -gt 0 ]; do
   esac
 done
 
+# --- GUARD 9: NO TEST MAY OPERATE ON THE REPO THE SUITE RUNS FROM -------------
+# 🔴 BEFORE the ROOT block below, not after it: with GIT_DIR set and no
+# GIT_WORK_TREE, git takes the CWD as the work tree, so
+# `rev-parse --show-toplevel` returns `<repo>/scripts` and this script dies
+# `exit 127` with no verdict. The 2026-08-21 incident, the reproduction, and the
+# reason this block is spelled in three files rather than sourced from one are
+# in `scripts/run-tests.sh`'s copy of this header. The SET is owned by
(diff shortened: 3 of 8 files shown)


ZacxDev:
This merged while the adversarial audit was still running, so the six findings land as a follow-up: **https://github.com/innovation-upstream/devrc/pull/720**.

🔴 **Two corrections to this PR's own body, so nobody reads it as authority later:**

1. **The `githooks/pre-push` `GIT_DIR` claim is FALSE as written.** The body says handing down `GIT_DIR` explains why *pushing* triggered the corruption. Measured on git 2.55.0, as a pair: a `git push` from a `GIT_DIR`-free parent gives `pre-push` only `GIT_EXEC_PATH` and `GIT_PREFIX` — **no `GIT_DIR`**; the same push with `GIT_DIR` exported by the caller passes it through. The rename is correct hygiene; it is **not** the diagnosis. 🔴 **The root cause of the 2026-08-21 incident remains unidentified** (a live scan found 46 processes carrying some `GIT_*` and **0** carrying `GIT_DIR`). Now pinned by `test_git_push_does_not_export_GIT_DIR_to_pre_push`, both halves.

2. **"Mutation-verified 12/12" is true of the plumbing mutants it chose and says nothing about the detector's CONTENT.** Five semantic mutants survived the green suite: dropping `HEAD`, dropping `packed-refs`, dropping `ORIG_HEAD`+`logs/HEAD`, and reducing `protected_git_dirs`' `starts` to either root alone. Each now dies to a mutation only that component can see.

And the two that would have made this a permanently-red gate on the operator's box:

- **The detector could not tell a fixture escape from any other writer, and asserted the former at maximum confidence.** Measured against a neighbour writing a branch every 40 ms: 40 innocent tests → `40 passed, 40 errors` with 80 messages naming innocent tests under *"This is the 2026-08-21 incident's shape"*; two of three runs instead hit `INTERNALERROR`, **exit 3, zero tests run**, because `pytest.fail` from `pytest_collection_finish` is a crash rather than a verdict. False-positive rate **100 % → 0 %**.
- **`DEVRC_GITENV_PROTECT` was a one-variable defeat of the whole detection half, and no runner unset it.** `":"` → `protected-git-dirs=0`, **GREEN**, branch really created. `/nonexistent/x` → `protected-git-dirs=1` — a marker line asserting coverage over a path that cannot hold a repository — **GREEN**, branch really created.

Full evidence and the before/after tables are in the follow-up PR.

PR #686, opened 2026-08-21
fix(signal): link an outbound group draft to its GROUP, not a phantom contact

ZacxDev:
> **This PR body was wrong until now.** It contained the body of an unrelated PR (`vetrllc/vetr-app#153`, a CSS width fix) — written by a *concurrent* agent whose `--body-file` collided with this one's on a shared scratchpad path. The title, branch, commits and diff were always correct; only the description was another PR's. Corrected below, publicly rather than silently, in case a reviewer read the wrong version. See the comment thread.

Links an outbound group draft to its **group**, instead of storing it as a DM to a person who does not exist. Clawgate task 313.

## The defect

`draft_message()` never set `messages.group_id`, and resolved *any* recipient through `upsert_contact` — so drafting to a group minted a **phantom contact**: a fake person whose `phone_number` was the group address.

Measured on the live store (the message that exposed it):

```
signal.messages  id 51 | is_outbound t | send_state sent | group_id NULL | dest_contact_id 92
signal.contacts  id 92 | phone_number 'group.elgzaTRR…' | uuid5 placeholder
```

Consequences, sharpest first:

1. **Mute was bypassed.** `_muted_predicate` keys on `m.group_id`, so a draft to a MUTED group was returned in full by every read that believed it was filtering.
2. Group-scoped reads and `idx_msg_group` never saw our own sent messages — a conversation read as one-sided, inbound only.
3. `signal.contacts` accumulated one fake person per group drafted to.

The phantom was also load-bearing: `send` read the recipient string back out of it, which is why the fix had to rewire recipient derivation rather than just stop creating the row.

**The inbound path already did this correctly** (`store_message` calls `upsert_group` and sets `group_id`). This makes the outbound path do what inbound already does — one rule, one place.

## The change

- `draft_message()` branches on recipient shape: group address → `upsert_group` + `group_id` set, `dest_contact_id` NULL; uuid / phone → contact, unchanged.
- `send`/`reconcile` derive the wire recipient from a `LEFT JOIN signal.groups`, not from a contact.
- A strict decoder refuses a malformed group address rather than inventing 32 bytes.
- `db77d014` (test-only): `FILTERED_READS` / `EXEMPT_READS` now **partition** every read of `signal.messages`, checked by AST scan against what the code does.

## Verification

| | passed | failed | skipped |
|---|---|---|---|
| base `732db793` | 624 | 0 | 1 |
| commit 1 `53a0c8ab` | 635 | 0 | 1 |
| **HEAD `db77d014`** | **650** | **0** | **1** |

Counted from `PASSED`/`FAILED` lines, not exit codes.

**Red at base**, verbatim, with the mute test dying on its own assertion and its positive control passing first:

```
AssertionError: MUTE BYPASSED: list_conversations returned a draft to a MUTED group (1 row(s)).
`not_excluded()` keys on m.group_id, so a draft stored with group_id NULL walks straight through it.
```

All three filtered reads bypassed at base (`list_conversations 1, search 1, get_message 1`); `0/0/0` at HEAD.

**Mutation-tested**, isolated, `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` cleared between mutants and an unmutated control each run. Includes the union-preserving **move** mutant (`get_message` FILTERED→EXEMPT, union byte-identical), which a set-union check cannot see and which dies on the code-vs-ledger assertion.

**Not relying on the SQLite substrate.** `tests/fakepg.py` is dynamically typed and has previously let a send path that was dead on real Postgres pass 599 green tests. The real schema was applied to a throwaway `postgres:18` and the whole path exercised there; `test_pg_type_compat.py::test_draft_query_runs_on_real_postgres`, normally skipped, ran and passed. All new SQL is ANSI `CAST(… AS text)` — never `::text`, which type-checks on Postgres and is unparseable by SQLite.

**The external wire contract is measured, not inferred.** `_group_id_to_address()` applied to the Vetr app group's raw id reproduces **byte-for-byte** the exact string that the live `signal-cli-rest-api` accepted with a `201` (4 recipients) on 2026-08-21, and round-trips back to the same raw id. The old code passed the operator's string through opaquely; the new code round-trips it through a decoder/encoder, so this equality is the thing that needed proving.

## Two guards this work found in its own tests

Both are the "spelled rather than structural" shape, and neither was a logic bug:

1. The predicate check was `"not_excluded(" in inspect.getsource(fn)` — a substring over source **including docstrings**. `get_draft` does not filter and must not, but its docstring *explains* the mute predicate, so the guard reported it as filtering. It was one commit from certifying the opposite of the truth.
2. The first fix for that excluded the docstring by comparing **values** against `ast.get_docstring()`, which cleans and dedents — so only single-line docstrings matched and every multi-line one was still read as SQL. Caught only by a mutant targeting a *named* test rather than accepting "something failed". Fixed by excluding the docstring by **node identity**, plus a multi-line control.

## Known limits

- **New rows only.** Production rows written by the old code keep `group_id` NULL until backfilled; `upsert_message`'s `ON CONFLICT` does not set `group_id`, so a sync echo cannot repair them either. Scope is measured and small — exactly **1** contact and **1** message on the live store. Backfill is queued separately and is guarded so a re-run is a no-op.
- Group discovery on draft is **forward-only**: a canonically-encoded but wrong 32 bytes creates a group row that never existed and the send goes nowhere while reporting success. Malformed spellings are refused; this case is not warned about. Follow-up.
- The exemption of `get_draft`/`list_drafts` from the mute predicate is deliberate — they return only rows the operator composed, and filtering them would strand a muted group's `sending` draft with no route out, an undesigned refusal path through the D3 gate. The reason lives in the `EXEMPT_READS` ledger itself, at the point of decision.

### DIFF
--- a/claude/skills/signal/SKILL.md
+++ b/claude/skills/signal/SKILL.md
@@ -59,6 +59,9 @@ cd ~/workspace/devrc/scripts/signal
 python3 consumer.py conversations --limit 10
 python3 consumer.py search "harbour permit"
 python3 consumer.py draft --to +15550100 --body "on my way"   # -> pending + a clawgate card
+
+# draft to a GROUP: --to takes the `id` field (group.<double-base64>), NOT internal_id
+python3 consumer.py draft --to 'group.<double-base64>' --body "on my way"
 python3 consumer.py drafts --state pending
 python3 consumer.py approve 42 --ref clawgate-task-91
 python3 consumer.py send 42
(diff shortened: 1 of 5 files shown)


ZacxDev:
**Correcting this PR's description publicly, because a reviewer may have read the wrong one.**

Until now this PR's body was the body of a completely unrelated PR — `vetrllc/vetr-app#153`, a Radix `SelectContent` CSS width fix, complete with viewport measurements and `select.tsx` line references. None of it described this diff.

**The title, branch, commits and diff were always correct.** Only the description was wrong, so nothing about the change itself is in question. It has been replaced with an accurate body.

**How it happened, because the mechanism matters more than the typo.** Two agents were working concurrently in different repos, and both wrote their PR body to a file in the *same* shared scratchpad directory under the same name. This PR's `gh pr edit --body-file` then read the other one's file. The contamination ran one way only — vetr-app#153's body is correct and was never touched.

Worth recording as a hazard rather than an accident: parallel agents get isolated *worktrees*, but they share one scratchpad path, so a same-named intermediate file is a silent cross-write with no conflict, no error, and a plausible-looking result. `--body-file` is a particularly bad place for it to land, because the wrong content is confident, well-formatted prose that reads as if it belongs.

Caught by an adversarial audit reading the diff against the description, not by anything mechanical. Nothing in CI compares a PR body to its diff, and nothing would have.

PR #688, opened 2026-08-21
docs(gate): the EXPECTED_SKIPS header invites a conditional form that silently fails 8 tests

ZacxDev:
The pin itself landed concurrently in #687 — this PR is now **comment-only**, recording the trap we both hit.

`runner_patch.py` empties `EXPECTED_SKIPS` for patched runner copies by regex on the **literal**. An `EXPECTED_SKIPS+=(…)` appended after the closing paren survives that substitution and makes patched copies die `FEWER than pinned`. Measured: **8 failures** across three test files, none of which name this array.

The header paragraph reads as inviting exactly that form. Now it warns against it, with the measured evidence and a two-line control.

Note for reviewers: the comment contains the forbidden token, so any automated scan for it must be comment-aware — a naive substring check flags this file.

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -1700,6 +1700,20 @@ _nolaunch_ack_reason() {
 # Built AFTER $HOME is settled: an entry may be conditional, but its condition
 # must be the SAME predicate the test itself uses — never a blanket allowance,
 # or the pin degrades into the numeric ceiling it exists to beat.
+#
+# 🔴 A conditional entry must still live INSIDE the array literal below — never
+# as an `EXPECTED_SKIPS+=(…)` appended after the closing paren. `runner_patch.py`
+# neutralises this array for patched copies by regex on the LITERAL
+# (`^EXPECTED_SKIPS=\(.*?^\)` -> `EXPECTED_SKIPS=()`), because a copy driving a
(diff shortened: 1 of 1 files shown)


ZacxDev:
Superseded by main's third-field conditional mechanism.

When I opened this, the header said "an entry may be conditional" with no supported way to express one, so the obvious reading was `EXPECTED_SKIPS+=(…)` after the closing paren — which `runner_patch.py` does not neutralise (it rewrites only the literal, `^EXPECTED_SKIPS=\(.*?^\)`), so an appended entry survives into patched runner copies, which observe zero skips and die `FEWER than pinned`. Measured cost when I did exactly that: **8 failures** across `test_activity_spool_isolation.py`, `test_gate_exit_truthfulness.py` and `test_no_real_launchers_all_targets.py` — none of which name this array, so the blast radius gives no hint about the cause.

That trap is now much harder to fall into: main's third field makes conditionality first-class and is documented immediately under the sentence that misled me. A warning about the wrong road is worth less once the right road is signposted, and this branch conflicts, so it is not worth the rework.

Recording the mechanism here so it is searchable if anyone hits it anyway: **`runner_patch.py` neutralises `EXPECTED_SKIPS` by regex on the LITERAL only. Anything appended after the closing paren survives into patched copies.**

PR #689, opened 2026-08-21
fix(tests): the test tier could rewrite the operator's clone and push to the PUBLIC repo — GUARD 9 makes the write impossible

ZacxDev:
The test tier could rewrite the operator's clone and push to the **public** repo. It
did, once — `main` was wiped on GitHub. GUARD 9 makes that write impossible from a
test.

Root cause of the incident: an ambient `GIT_DIR` overrides `git -C <path>`, so "every
call passes `-C`" confers no safety. And a git **worktree is not isolation** — it
shares the base clone's common dir, so a `git config` from inside one writes the
operator's real `.git/config`.

---

## 🔴 GATE STATUS: **RED**. Do not merge this as it stands.

`scripts/gate.sh --tier pytest`, run against this exact head in a standalone sandbox
clone, ends:

```
RESULT: FAIL (exit=1)
GATE: RESULT=FAIL exit=1
```

`TOTAL collected=14505 passed=14281 skipped=2 failed=222` — 3 targets red
(`scripts/tests` 37 failed / 177 errors, `scripts/repo-cos/tests` 7,
`scripts/task-spec-drafter/tests` 1); the other 22 pytest targets and all 6 script
targets pass, every per-target floor clears.

**The guard is refusing its own fixtures.** ~221 of the 224 refusals are one shape:

```
git(refused)  would mutate the PROTECTED repository <sandbox>/.git
              git upload-pack <tmp_path>/…/origin.git
```

A **local** `git clone <fixture>` makes git spawn `git-upload-pack <dir>`. Now that
the farm substitutes every dispatch name (below), that child is the shim — and the
repo it targets arrives as a **positional** argument, which the shim's global-option
parser (`-C`, `--git-dir`, …) cannot see. With no target parsed it falls back to
**cwd**, which in a test run is the protected repo, and refuses. The remaining 3 are a
second, smaller class: `-c core.fsmonitor=` is not on `INJECTABLE_KEYS`.

Isolated with a three-arm control — one variable moved between arms 2 and 3, cwd held
at the protected repo throughout:

| arm | `git clone <local fixture>` |
|---|---|
| 1 · no farm | rc 0 |
| 2 · farm as shipped | **rc 128, GUARD 9 REFUSED** |
| 3 · farm, only `git-upload-pack` pointed back at the real binary | rc 0 |

So it is specifically the interception of `git-upload-pack`, not the farm as such.
Both classes are **false refusals** — no test is actually reaching a real repo — but a
permanently-red gate is worse than no gate, so this needs a fix before merge. Fixing
it is a design call (which plumbing verbs carry a positional repo path, and whether
`core.fsmonitor` belongs on the allowlist), deliberately not made here.

The RCE fix and the retraction below stand on their own and are why this head is
pushed now.

---

## 🔴 RETRACTION: the exec-path farm did **not** close the alias/hook residual

An earlier version of this PR claimed a `GIT_EXEC_PATH` symlink farm **closed** the
alias/hook residual. **That claim was false and is withdrawn.** Two reviewers were
told it on that evidence, so it is stated here rather than silently edited away.

The farm substituted only the literal name `git`. `libexec/git-core` holds **181
`git-<verb>` entries**, and **git dispatches on `argv[0]`** — so all 181 reached the
real binary while sitting *first* on `PATH` in exactly the alias/hook context the farm
was added to cover. Measured against that revision, with the farm installed and every
other check in place:

```
<farm>/git-config -f <victim>/.git/config core.bare true
  -> core.bare false -> true, rc 0
```

That is **the incident's own signature**, landed at rc 0, without the word `git` being
spelled once.

The farm now substitutes every dispatch name. The residual is **two derived escapes**,
not zero — narrowed, **not closed**. Six alias-body shapes were measured:

```
!git …                 guarded
!sh -c 'git …'         guarded
!env git …             guarded
!command git …         guarded
!/abs/path/to/git …    🔴 ESCAPES  (PATH is never consulted)
!touch <victim>/file   🔴 ESCAPES  (never invokes git at all)
```

Owning git's exec path bounds what **git** will resolve for a child; it cannot bind a
command that bypasses resolution or never asks for git.

## 🔴 This PR fixes a code-execution bug in its own guard

`--config-env=key=VAR` puts `VAR` on the command line, and the shim did
`eval "_cv=\${$_ev:-}"` with it. So:

```
git --config-env=core.pager='V:-$(touch X)' --version
```

executed the `touch` **inside the shim, on a pure read, before any check ran**. Real
git rejects that name, exits 128 and runs nothing. **On that path the guard was
strictly worse than no guard** — it created execution where the unguarded binary had
none.

Fixed by validating the name as a shell identifier (an invalid one is now **refused**,
matching real git) and reading the value with `printenv`, which does not interpret.
All four `eval`s in the shim were audited; only that one took an argv-derived operand.
A test fails if a new `eval` appears **or the count changes**.

## Why the fix is keyed on NAME, not inode

A reviewer will reach for `stat -c %i` here; it is the wrong attribute, and it audits
green over a live hole. This host carries **two git binaries from unrelated nix
closures**:

| binary | inode | version |
|---|---|---|
| `~/.nix-profile/bin/git` — what bare `git` resolves to, and whose libexec `--exec-path` reports | 53015695 | 2.55.0 |
| `/run/current-system/sw/bin/git` | 62031535 | 2.52.0 |

Measured **inside the single directory `--exec-path` reports** (i.e. the one the farm
is built from), counting entries by resolved inode:

- keyed on the profile git's inode → **146**
- keyed on the system git's inode → **0**

So an inode-keyed enumeration is **`PATH`-dependent within one machine**: pick the
other binary — trivially done by a `nix-shell`, a `PATH` order change, or a `sudo` —
and it matches nothing while reporting a clean farm.

It is worse than that even when keyed correctly. That directory holds **181** `git-*`
dispatch names but only **146** of them resolve to git's inode; the other 35 are
separate executables (perl/shell helpers). An inode enumeration therefore misses 35
live dispatch names outright, on the *right* binary.

The property that actually matters is the **dispatch name** — 181 `git-*` entries,
identical on both hosts, and satisfied equally by a hardlink, a symlink, a copy or a
wrapper.

## Scope of the guard

The guard closes the **test tier's route** into these classes, not the classes
themselves — an agent or human running `git config` from a worktree by hand is still
unguarded, and that is what actually re-armed the operator's hooks today.

## What is in the change

- `scripts/testlib/norepo.py` — the shim: fail-closed allowlist of read forms,
  protected-path resolution via `--git-common-dir` **plus** every attached work-tree
  root, injected-config allowlist, the exec-path farm.
- `scripts/testlib/norepo_plugin.py` — pytest-side installation.
- `scripts/run-tests.sh` — installs the guard for all 24 targets (including the
  non-pytest ones no conftest can reach), plus `GIT_ALLOW_PROTOCOL`,
  `GIT_CONFIG_NOSYSTEM`, a redirected `GIT_CONFIG_GLOBAL`, and the GUARD 9 report.
- `scripts/tests/test_no_real_repo_writes.py` — arms every vector against a real
  victim and asserts it is byte-unchanged; each armed vector carries an **unguarded**
  arm that must move an asserted axis, because four of the ten originally written
  moved nothing and passed identically either way.
- Destination- and command-bearing options (`--output=`, `grep -O`) are checked
  **above** the read fast path — they are reachable from subcommands on the
  always-read list, so a "read" could otherwise still write a file or run a command.

### DIFF
--- a/scripts/analyze-service-index/commit.sh
+++ b/scripts/analyze-service-index/commit.sh
@@ -168,6 +168,19 @@ if [ "${#POSITIONAL[@]}" -gt 1 ]; then
   usage: commit.sh [--print-plan] [STORE_DIR]"
 fi
 
+# 🔴 A GIVEN-BUT-EMPTY STORE IS A BUG, NOT A REQUEST FOR THE DEFAULT. `:-`
+# cannot tell "no argument" from "an argument that is the empty string", so
+# `commit.sh ""` — the shape a caller produces when its own path computation
+# fails — would silently `git init`, `git add` and `git commit` in the
+# OPERATOR'S REAL STORE instead of the directory it meant. This is the same
+# class as ship.sh's SHIP_REPO and drift-check.sh's DRIFT_REPO; see
+# scripts/tests/test_no_real_repo_writes.py, which pins all four sites.
+if [ "${#POSITIONAL[@]}" -eq 1 ] && [ -z "${POSITIONAL[0]}" ]; then
+  die "STORE argument was given but is EMPTY.
+  That is a caller bug, not a request for the default — an empty value would
+  silently resolve to \${HOME}/.claude/analyze-service-index and commit in the
+  operator's real store. Pass no argument for the default, or pass a path."
+fi
 STORE="${POSITIONAL[0]:-${HOME}/.claude/analyze-service-index}"
 
 ASI_GIT_NAME="${ASI_GIT_NAME:-analyze-service index}"
--- a/scripts/drift-check.sh
+++ b/scripts/drift-check.sh
@@ -441,6 +441,19 @@ if [ "${1:-}" = "--detect-role" ]; then
   exit 0
 fi
 
+# 🔴 SET-BUT-EMPTY IS A BUG, NOT A REQUEST FOR THE DEFAULT. `${VAR:-default}`
+# cannot tell "unset" from "set to the empty string", so a caller that computed
+# a repo path and got `""` silently checks — and `git fetch`es — the OPERATOR'S
+# OWN CLONE instead of the one it meant. UNSET must keep defaulting (this
+# variable is deliberately NOT forwarded over ssh, and the remote host's repo
+# lives at its own $HOME/workspace/devrc); EMPTY must stop the run.
+if [ "${DRIFT_REPO+set}" = set ] && [ -z "$DRIFT_REPO" ]; then
+  echo "drift-check: DRIFT_REPO is SET but EMPTY." >&2
+  echo "  That is a caller bug, not a request for the default — an empty value" >&2
+  echo "  would silently resolve to \$HOME/workspace/devrc and fetch into the" >&2
+  echo "  operator's own clone. Unset it to get the default, or give it a path." >&2
+  exit 2
+fi
 DRIFT_REPO="${DRIFT_REPO:-$HOME/workspace/devrc}"
 DRIFT_UNTRACKED_MAX="${DRIFT_UNTRACKED_MAX:-10}"
 DRIFT_DANGLING_MAX="${DRIFT_DANGLING_MAX:-10}"
@@ -589,6 +602,10 @@ severity() {
 # remote-tracking refs only.
 CHECK='
 set -uo pipefail
+if [ "${DRIFT_REPO+set}" = set ] && [ -z "$DRIFT_REPO" ]; then
+  echo "[${DRIFT_LABEL:-host}] DRIFT_REPO is SET but EMPTY — refusing to fall back to \$HOME/workspace/devrc." >&2
+  exit 2
+fi
 repo="${DRIFT_REPO:-$HOME/workspace/devrc}"
 label="${DRIFT_LABEL:-host}"
 maxu="${DRIFT_UNTRACKED_MAX:-10}"
@@ -970,6 +987,10 @@ echo "[$label] PARITY-RC=$p_rc"
 SRCREPO='
 set -uo pipefail
 label="${DRIFT_LABEL:-host}"
+if [ "${DRIFT_REPO+set}" = set ] && [ -z "$DRIFT_REPO" ]; then
+  echo "[$label] DRIFT_REPO is SET but EMPTY — refusing to fall back to \$HOME/workspace/devrc." >&2
+  exit 2
+fi
 repo="${DRIFT_REPO:-$HOME/workspace/devrc}"
 sfto="${DRIFT_SRC_FETCH_TIMEOUT:-30}"
 ssay() { echo "[$label] $*"; }
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -1586,6 +1586,7 @@ export DEVRC_TEST_SPOOL_GUARD_DIR="$SPOOL_DIR"
 # how ten of this file's own regression tests work — would inherit the flag, and
 # every target inside it would be scored as a nested session with no marker.
 unset DEVRC_TEST_SPOOL_IN_SESSION
+unset DEVRC_TEST_GITGUARD_IN_SESSION
 SPOOL_SESSIONS_LOG="$SPOOL_DIR/sessions.log"
 SPOOL_ISOLATED_LOG="$SPOOL_ISOLATED/current.log"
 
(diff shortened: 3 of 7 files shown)


ZacxDev:
## Late findings — read before extending the `git-*` farm

Two things surfaced after the body was written. Both come from an independent reviewer (devrc-08) and both are recorded here rather than folded into the body, so the sequence stays legible.

### 1. 🔴 The positional-argument hole is a per-verb design cost, not one special case

The RED gate above is caused by `git-upload-pack <dir>`: the shim's global-option parser sees `-C` and `--git-dir` because those are **options**, but a plumbing verb that takes its repo as a **bare positional** is invisible to it, so the target falls back to `cwd` — the protected repo — and the guard refuses. ~221 of the 222 failures are that one shape.

It does **not** stop at `upload-pack`. The same argument shape covers at minimum:

- `git-upload-pack <dir>`
- `git-receive-pack <dir>`
- `git-upload-archive <dir>`
- `git-init-db`
- the `--git-dir`-less forms of several other `git-*` helpers

**So any farm that routes by name inherits a per-verb argument-shape problem.** Each verb whose repo arrives positionally needs its own parse rule, and the next person extending the allowlist will hit this once per verb rather than solving it once. That is a standing cost of the name-routing design and it should be weighed against the alternative of fixing the mechanism (ambient repo-pointer inheritance) rather than policing the surface.

Isolation control for the record — cwd held constant, one variable moved:

| arm | rc |
|---|---|
| 1 · no farm | 0 |
| 2 · farm as shipped | **128, GUARD 9 REFUSED** |
| 3 · farm, only `git-upload-pack` pointed back at the real binary | 0 |

Arm 3 landing on `0` is what distinguishes *"the guard is wrong"* from *"the guard is right and the target resolution is wrong"*. It is the latter. The refusals are **false** — no test reaches a real repository — but the effect is a permanently-red tier.

### 2. The `222` is not attributable — rebase before re-measuring

This branch is behind `origin/main`, which now contains **#696 (`6439921`)**, fixing two hermetic failures that were on `main` itself (a `git ls-files` error under `/build/src`, and an eager `scope_for_repo` in `subsystem_touch.py`), plus a third cause fixed concurrently (an unpinned `SIGNAL_PG_DSN` skip).

So on this pre-#696 base, **2–3 of the 222 failures are inherited from the baseline**. That is small and does not affect the diagnosis above — but it means `222 − ~221` cannot be used to conclude the residual is exactly the 3 `core.fsmonitor` cases. **Rebase onto current `main` and re-run before doing that arithmetic.**

### 3. Correction to the inode figures in the body

The body's `146` and `0` are both counts **within the single directory `--exec-path` reports**, keyed on each of this host's two git binaries. Measuring the *system* git's own libexec instead gives `143`, not `0` — so quoting "146 and 0" without naming the directory invites a misreading.

The stronger and more durable claim, which is the one the fix rests on: that directory holds **181 dispatch names but only 146 resolve to git's inode**, so an inode-keyed enumeration misses **35 entries even on the right binary**. That is a scoping defect, not a `PATH` artefact — and it is why the fix keys on the dispatch **name**.

ZacxDev:
## Closing in favour of #683 — the architecture, not the effort

This is a strategy decision after measuring all four guard PRs (#673, #676, #683, #689) against current `main`. The work here is good and its self-audit is the reason the decision could be made at all — the RETRACTION, the three-arm control isolating `git-upload-pack`, and the honest `GATE STATUS: RED` are what made the cost visible. None of that is being thrown away silently; see below for what was salvaged.

### Why not fix the per-verb positional-repo problem

It is not one fix. I enumerated git's dispatch table on this host: **14 of the 181 `git-*` entries take a repository positionally by synopsis alone** — `upload-pack`, `receive-pack`, `fetch-pack`, `send-pack`, `upload-archive`, `clone`, `init`, `init-db`, `archive`, `fetch`, `pull`, `push`, `ls-remote`, `mailsplit`. On top of that sits an open-ended set of *destination-bearing options* across all 181; #676's audit already found `collect_extra_dests` covering 4 verbs and missing `checkout-index --prefix`, `pack-objects` and `diagnose -o`, and #676's own title records the pattern: "the audit found nine bypasses, a second audit found eight more spellings of the same nine."

Every bounded formulation fails:

- **Enumerate the positional verbs** → buys the next audit round, not closure. The surface grows with git, in-repo, forever.
- **Default-deny any verb without a grammar** → 167 verbs become false refusals. Permanently red.
- **Resolve the target with real git instead of parsing argv** → that *is* the 8.85× overhead already measured, and it cannot work for `git-upload-pack <dir>`, which is not a repo-resolving verb at all.

The failure mode on either side is bad: false refusals give a permanently-red gate, which `claude/RULES.md` says is worse than no gate — and that now bites for real, since devrc has acquired a Tekton pre-merge gate (`tekton/devrc-pytests`) since this PR was opened. Missed spellings give a silent hole that audits green.

### What #683 does instead

It closes the **measured mechanism** — an ambient `GIT_DIR` beating `-C` — rather than policing git's whole surface:

- an **11-name ledger** (`REPO_POINTER_VARS`) stripped at four enforcement points: all three shell runners *before* each resolves its `ROOT`, plus the pytest plugin at import. This PR scrubs **5 of those 11** and never touches `run-node-tests.sh`.
- a **fingerprint detector** over `refs/`, `HEAD`, `packed-refs`, `ORIG_HEAD`, `logs/HEAD` and `config`, compared around collection, around every test, and at session end — failing *the test that moved the ref*. That is the coverage this PR's interceptor was reaching for, delivered against the repo that actually matters, for any future mechanism rather than the spellings someone enumerated.
- it also fixes `githooks/pre-push`'s literal `GIT_DIR=` assignment, a concrete route from `git push` into a poisoned suite that this PR does not touch.

Measured on a standalone clone with `origin` removed, base `a34d695d`:

```
control  pristine main       pytest 14689 collected / 0 failed   scripts/tests 687.7s   GATE PASS
main + #683                  pytest 14721 collected / 0 failed   scripts/tests 699.6s   GATE PASS
```

+1.7% wall clock, and the guard's positive-control marker prints from **26 of 26** pytest targets, so it is loaded rather than silently absent.

### What was salvaged, and what is genuinely lost

**Salvaged into #716:** the six `${VAR:-default}` set-but-empty guards in `ship.sh`, `drift-check.sh` and `analyze-service-index/commit.sh`, with their two-way ledger and the behavioural test that asserts the exact `exit 2`. Those are independent of the interceptor — plain caller-bug guards in production scripts — and they would have died with this PR. Red/green re-measured: 7 failed at base → 9 passed.

**Recorded, not lost:** the name-vs-inode finding (181 dispatch names vs 146 inode matches; `PATH`-dependent within one machine) and the `--config-env` `eval` RCE are already in the analyze-service index under `devrc/tests`, and the RCE is there as a pattern — *audit every `eval` of an argv-derived value* — not as an instance.

**Genuinely lost:** protection for writes to *arbitrary third* repositories by absolute path. Stating it plainly rather than implying #683 covers it. Two things make it an acceptable residual: it is not the incident (the victim was the repo the suite runs from, which #683 both prevents and detects), and this PR does not actually deliver it today given its seven measured bypasses. If a specific second victim needs covering — `~/.claude/analyze-service-index` is the realistic one — `protected_git_dirs()` in #683 takes an extra root, which is a few lines rather than an argv grammar.

Reopen if you disagree with the architecture call; the branch is untouched.

PR #692, opened 2026-08-21
fix(gate): skips self-pin via requires-env[VAR] — the ledger went stale twice

ZacxDev:
## The problem

`EXPECTED_SKIPS` is a hand-maintained ledger in `scripts/run-tests.sh` pinning skips that are *declared in other suites' test files*. Keeping the two in step is prose discipline, and it has failed twice, leaving `main` RED both times:

- **#332** — 2026-08-04 → 08-06, **two days**; tests added without the ledger entry.
- **#657** — 2026-08-21; a Postgres test added without the entry. Found only because it blocked an unrelated push, and fixed by #687.

The file already states the correct rule, right above the array:

> an entry may be conditional, but its condition must be the SAME predicate the test itself uses — never a blanket allowance

It just had no way to enforce it. **79 `pytest.mark.skipif` sites across the repo; one ledger entry.** Every one of those 79 is a latent red-main on whichever host lacks its condition.

## The change

A skip can now carry its own condition, so there is no second place to keep in step:

```python
@pytest.mark.skipif(not os.environ.get("SIGNAL_PG_DSN"),
                    reason="requires-env[SIGNAL_PG_DSN]: needs a real Postgres; "
                           "SQLite cannot reproduce Postgres static type checking")
```

`run-tests.sh` reads the VAR out of the reason and:

- **forgives** the skip where that variable is absent — no ledger entry needed, none can go stale;
- **FAILS when the variable is set** — the environment could have run it and it declined, which is the exact coverage collapse GUARD 2 exists to catch and which an unconditional pin cannot see;
- **still fails** on any undeclared skip, unchanged.

This is *stricter* than a pin, not looser. A pinned skip is forgiven unconditionally, everywhere, forever.

It is a **reason-string contract, deliberately not an import** — no `sys.path` plumbing, so all 19 suites can adopt it immediately.

The Postgres skip is migrated and its `EXPECTED_SKIPS` entry deleted.

## The boundary, kept rather than papered over

**repo-cos stays on the ledger.** Its predicate is a *value* test (`REPO_COS_LIVE_DRIFT_CHECK == "1"`), not is-set. Exporting the variable as `0` leaves it non-empty while the test still skips — which the self-pinning check would report as a false collapse. The mechanism deliberately covers only the is-set shape, and `test_the_value_predicate_skip_stays_on_the_ledger` pins that boundary so nobody "finishes the migration" into a false positive.

## The guard the runner structurally cannot provide

The runner reads the VAR from the **reason**. It never sees the `skipif` **condition**. A test declaring `requires-env[FOO]` while keying on `BAR` would be forgiven on a host lacking FOO and never checked against BAR — the exact class this exists to remove, invisible to the mechanism itself.

So `scripts/tests/test_requires_env.py` parses **both** out of the AST and pins them together. A guard whose description claims a relationship must inspect both sides; inspecting one and reading as coverage is worse than none.

It carries **positive controls** — that the file scan sees >200 tracked test files, that the condition extractor finds `os.environ.get("X")` / `os.environ["X"]`, that the reason regex matches valid names and rejects invalid ones — because every assertion over an empty set passes vacuously, and a reassuring zero is indistinguishable from a parser wired to nothing.

## Mutations — each killed by its own named test

| mutation | killed by |
|---|---|
| reason declares `WRONG_VAR` while the condition reads `SIGNAL_PG_DSN` | `test_every_declared_var_is_the_one_the_condition_tests` |
| re-add the migrated skip to `EXPECTED_SKIPS` (double-pinned) | `test_the_migrated_postgres_skip_is_declared_and_unpinned` |
| runner uses `$evar` (the NAME) instead of `${!evar-}` (the VALUE) | `test_runner_expands_the_var_indirectly` |

That last one matters: `$evar` is the variable *name*, always non-empty, so it would score **every** declared skip as a collapse — a one-character slip that fails in the reassuring direction.

## 🔴 Gotcha found while verifying: the gate runs the CALLER'S venv

Getting an honest gate run took four attempts, and three intermediate readings were wrong in ways that looked like a broken branch:

1. A `.venv` is **recreated on every `cd`** into a devrc worktree and lacks pytest → the gate dies `FATAL` before running anything. The primary clone is immune only by accident: its `.venv` has no `bin/python`, so nothing shadows.
2. Avoiding the `cd` made it pick up **an unrelated repo's venv** (`datapacket-talos/.venv`, my shell's cwd) → `mail-actions collected=0`, `signal collected=2 errors=2`, `initiatives failed=9`. **13 failures that were entirely environmental.**
3. `env -u VIRTUAL_ENV` with `/.venv/bin` stripped from `PATH` was still not enough — `nix-shell --run` executes the user's shell hooks, which re-activate the venv from cwd.

Only running from a neutral cwd gave a clean `/nix/store/…python3-3.12.14-env/bin/python`. **The discriminator is the traceback's file path** — a devrc test run whose pytest lives in `datapacket-talos/.venv` is not testing what you think.

Worth noting the runner's own guard behaved correctly throughout: it refused with `run-tests: FATAL — python -m pytest is not runnable` rather than reporting a green over zero collected tests.

## Follow-up, not in this PR

The remaining env-keyed `skipif` sites are unmigrated, and there is no guard yet requiring the prefix on new ones. That sweep plus its enforcement is a separate change; this PR lands the mechanism and proves it on the real case that broke `main`.

## Verification — both arms, measured

The forgiving path being green proves little on its own; the question is whether the mechanism can still **refuse**. So both arms were run against the full suite:

| arm | condition | result |
|---|---|---|
| **A** — declared VAR absent (`SIGNAL_PG_DSN` unset) | the real, migrated state | **PASS, exit 0** — `collected=14378 passed=14376 skipped=2 failed=0`; no unpinned-skip error, ledger entry deleted |
| **B** — declared VAR **set** (reason repointed at `HOME`, condition unchanged) | the collapse this PR adds | **FAIL, exit 1** — `1 declared-env skip(s) SKIPPED WITH THE VARIABLE SET` + `2 skipped, but 1 accounted for` |

`14378` is the `14369` baseline **plus exactly the 9 new tests** — nothing silently stopped collecting. Arm A's `skipped=2` is the repo-cos ledger pin plus the self-pinned Postgres skip, so the widened accounting (`1 ledger + 1 self-pinned = 2`) held. Arm B's second error is that same equality going the other way, which is the cross-check working rather than a duplicate message.

Arm B's edit was reverted and confirmed byte-identical afterwards.

## Reviewer's attention, stated rather than left to be discovered

**The mechanism's safety rests on a test, not on the runner.** `run-tests.sh` trusts a string written in a test file: it reads the VAR from the skip REASON and never sees the `skipif` CONDITION. A typo'd or mismatched variable name is forgiven by the runner and caught **only** by the AST cross-check in `test_requires_env.py`.

That is a deliberate trade — the alternative was `sys.path` plumbing into 19 suites, which is why this is a string contract — but it means `test_requires_env.py` is load-bearing, not decorative. If it is ever weakened, the mechanism silently degrades to "forgive anything that says `requires-env[...]`".

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -2044,6 +2044,15 @@ fi
 env_pinned=0
 env_collapsed=()
 _env_skip_var() { printf '%s\n' "$1" | sed -nE 's/.*requires-env\[([A-Za-z_][A-Za-z0-9_]*)\].*/\1/p'; }
+# 🔴 A skip LINE is a GROUP, not one test: pytest's `-rs` emits `SKIPPED [N] …`,
+# where N is how many tests shared that reason. `TOT_SKIPPED` is summed from the
+# pytest SUMMARY and therefore counts TESTS. Counting groups here and comparing
+# against a test total is only accidentally right while every N is 1 — and it
+# breaks the instant someone puts the decorator on a CLASS or a parametrized
+# test, which is the normal way to declare an env requirement. The failure is a
+# FALSE RED that blocks pushes, i.e. exactly the "permanently-red gate" that
+# teaches people to pass --no-verify.
+_env_skip_count() { printf '%s\n' "$1" | sed -nE 's/^SKIPPED \[([0-9]+)\].*/\1/p'; }
 
 unexpected=()
 for line in "${SKIP_LINES[@]}"; do
@@ -2053,10 +2062,14 @@ for line in "${SKIP_LINES[@]}"; do
     # `${!evar-}` is the INDIRECT expansion — the value of the variable NAMED by
     # $evar, empty when unset. Never `$evar`, which is the name itself and always
     # non-empty, i.e. would score every declared skip as a collapse.
+    ecount="$(_env_skip_count "$line")"
+    # Default to 1 only if the `[N]` did not parse — never silently to 0, which
+    # would under-count and re-open the same false-red in the other direction.
+    [ -n "$ecount" ] || ecount=1
     if [ -n "${!evar-}" ]; then
       env_collapsed+=("$line")
     else
-      env_pinned=$(( env_pinned + 1 ))
+      env_pinned=$(( env_pinned + ecount ))
     fi
     continue
   fi
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -1704,20 +1704,20 @@ EXPECTED_SKIPS=(
   # Opt-in drift check against the LIVE homelab store — needs a kubeconfig and
   # network, neither of which a hermetic gate may have. Skips everywhere unless
   # REPO_COS_LIVE_DRIFT_CHECK=1.
+  # 🔴 STAYS A LEDGER PIN, deliberately. Its predicate is a VALUE test
+  # (`REPO_COS_LIVE_DRIFT_CHECK == "1"`), not an is-set test, so it does NOT fit
+  # `requires-env[VAR]` below: exporting the variable as `0` leaves it non-empty
+  # while the test still skips, which the self-pinning check would report as a
+  # false collapse. The mechanism deliberately covers only the is-set shape; a
+  # value predicate keeps its entry here.
   "scripts/repo-cos/tests|live-store drift check is opt-in"
-  # Needs a REAL Postgres (SIGNAL_PG_DSN). Unlike the skill_audit case below —
-  # which was correctly fixed by re-pointing at tracked fixtures so it RUNS —
-  # this one cannot be made hermetic: the test exists because SQLite does not
-  # reproduce Postgres static type checking, which is the exact defect #657
-  # fixed (a COALESCE over uuid/text that had never worked on real Postgres).
-  # gateTools carries psycopg2 (the client) but no Postgres SERVER, so nothing
-  # in the hermetic tier can satisfy it.
-  # 🔴 CONSEQUENCE, stated so it is not a surprise: this test is opt-in only and
-  # never runs in the gate. The regression it guards is caught only when someone
-  # runs with SIGNAL_PG_DSN set. Pinning unbreaks main; it does not restore the
-  # coverage.
-  "scripts/signal/tests|needs a real Postgres"
 )
+# 🔴 The Postgres skip that made main RED on 2026-08-21 (#657, pinned by #687) is
+# NO LONGER LISTED HERE — it now self-pins via `requires-env[SIGNAL_PG_DSN]` in
+# scripts/signal/tests/test_pg_type_compat.py. That is the point of the mechanism:
+# the condition lives with the test, so it cannot go stale against this file.
+# Its coverage caveat is unchanged and still true — the test is opt-in and does
+# not run in the gate — but it is now stated where the test is, not here.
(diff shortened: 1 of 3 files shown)


ZacxDev:
Closing this unmerged. The audit found three blockers, and re-measuring the premise afterwards showed the mechanism is disproportionate to the problem it targets.

## The premise was wrong

This PR's justification rested on "79 `skipif` sites; one ledger entry — every one a latent red-main." Measured across the tracked tree afterwards: **79 `skipif` sites, of which exactly 1 is env-keyed** — `scripts/signal/tests/test_pg_type_compat.py`, the one this PR migrates. The other 78 key on platform, imports, tool availability, and filesystem paths, none of which `requires-env[VAR]` covers.

So the mechanism serves a population of **one**, and #687's ledger entry already serves it.

The two incidents cited as justification do not both support it either: #657 was env-keyed and would have been prevented; **#332 was not** — it skipped off an `is_dir()` check on an out-of-repo path. The recurring class is "environment-dependent skip", and env-var-keyed is one flavour of it. A mechanism covering that flavour alone addresses one of the two incidents while adding a new forgiveness path to the gate every push depends on.

## What the audit found (all verified, none cosmetic)

1. 🔴 `_tracked_test_files()` calls `git ls-files` with `check=True`. `nix flake check` builds from a tracked-file copy with **no `.git`** — 4 tests fail there, including every relationship guard. Verification here used only the pre-push tier, which has `.git`. A PR motivated by `main` going red would have redded `main`, in the tier it never ran. The repo already documents this exact trap and its fix in `scripts/tests/test_doc_path_rot.py:74-86`.
2. 🔴 Forgiveness keys on the printed skip reason alone, so an unconditional `pytest.mark.skip`, a runtime `pytest.skip()`, an f-string reason, or a module constant are all forgiven while the AST guard sees none of them. There are 30+ runtime `pytest.skip()` sites already. **The runner cannot close this** — `skipif` and `pytest.skip()` emit byte-identical `SKIPPED [n] path:line: reason` lines, so only static analysis can distinguish them.
3. 🔴 The group-count fix in `51f88b36` made the total-count equality vacuous: `env_pinned` derives from the same run's skip lines, so both sides move together. `SKIPPED [500]` with `TOT_SKIPPED=501` passes silently. The original per-line increment caught that loudly — the "fix" traded a false red for silence, which is the worse direction, and the comment above it asserts the opposite of what the code does.

Plus: the mutation table's third row is **false**. The `${!evar-}` → `$evar` mutant was reported killed, but the `sed` used had also rewritten the identical literal in a comment two lines above, so the test went red for the wrong reason. Isolated to the code line, all 13 tests **pass** — the guard is walkable, because it is a substring check against a shell script and the string it looks for also appears in a comment.

## What is worth keeping

- The audit surfaced a **real pre-existing bug**, unrelated to this design: the current accounting is permanently red for any developer with `SIGNAL_PG_DSN` exported (`TOT_SKIPPED=1` against 2 ledger entries). Filed separately.
- The observation that the devrc gate is effectively unrunnable from a worktree, and picks up whatever `.venv` is active in the caller's environment — including an unrelated repo's. Filed separately.

Left unmerged rather than force-pushed away: the history here is more useful as a record than the branch is as a change.

PR #695, opened 2026-08-22
fix(opencode): narrow the age decrypt globs — "*age*-d*" ate ordinary read-only commands

ZacxDev:
`"*age*-d*"` had a `*` BETWEEN "age" and the flag, so it matched any bash command
whose text contained "age" followed LATER by "-d". Neither half is rare: "age" is
a substring of package, packages, image, message, storage and manage, and "-d"
covers -db-, --dir, --debug and --dry-run.

The consequence is not a spurious prompt. `opencode run` AUTO-REJECTS an ask
rather than prompting, so a matched command kills the run mid-task. Measured
2026-08-21 while dispatching work in a monorepo with a top-level packages/ dir:

    ls   packages/civitai-db-schema/src/            -> ask -> auto-reject -> run dead
    grep packages/civitai-db/src/lag.test.ts        -> ask -> auto-reject -> run dead

Two real dispatches died on exactly these, one on `grep` and one on `ls`. Both had
completed their research and were killed before writing anything.

Dropping the inner `*` requires the literal "age -d". The LEADING `*` stays and is
still doing work — it keeps /nix/store/.../bin/age -d and FOO=1 age -d gated.

Three consequences, all pinned rather than argued:

- Three MUST_ALLOW rows pin the false positives. Watched to FAIL before the fix:
  all three asserted 'ask' == 'allow' on read-only commands.
- `"*age --decrypt*"` is no longer redundant with `"*age -d*"` — "age --decrypt"
  does not contain "age -d" — so its REDUNDANTLY_COVERED_ASKS entry is removed and
  it gains its own MUST_ASK pin. The suite caught this itself: my first attempt
  removed the ledger entry while claiming in a comment that both rules were
  "pinned individually in MUST_ASK", which was false, and
  test_every_ask_rule_is_individually_pinned went red on it.
- The nix-collect-garbage note is updated, not deleted. Its coincidental rescue
  still holds: "nix-collect-garbage -d" ends in the literal "age -d" (gar-BAGE -d),
  so the narrowed glob still matches that spelling, and "*nix-collect-garbage*"
  remains the real gate for the rest.

DELIBERATELY NOT CHANGED: "*git*checkout*" and "*git*restore*". Agents repeatedly
reached for `git checkout -- <path>` to undo their own mistakes and were rejected,
which turned recoverable errors into dead runs, so loosening them looked tempting.
But DANGEROUS_FAMILIES pins both as must-not-be-allow and that is correct — the
command that lets an agent undo its damage is the same command that destroys
uncommitted work. The fix for that failure belongs in how briefs are written
(edit surgically; never rewrite a whole file), not here.

Suite: 636 passed.

### DIFF
--- a/scripts/opencode/opencode.jsonc
+++ b/scripts/opencode/opencode.jsonc
@@ -252,19 +252,30 @@
       "*sops*-d*": "ask",
       "*sops*--decrypt*": "ask",
       "*sops*exec-env*": "ask",
-      "*age*-d*": "ask",
-      "*age*--decrypt*": "ask",
+      // 🔴 NO `*` BETWEEN "age" AND THE FLAG — it must stay `*age -d*`, not
+      // `*age*-d*`. The inner `*` made this match any command text containing
+      // "age" followed LATER by "-d", and "age" is a substring of package,
+      // packages, image, message, storage and manage while "-d" covers -db-,
+      // --dir, --debug and --dry-run. MEASURED 2026-08-21: `ls packages/…-db-…`
+      // and `grep … packages/…-db/…` both resolved ASK, and since `opencode run`
+      // auto-rejects an ask rather than prompting, two real dispatches were
+      // killed mid-task by it. The leading `*` is still required and still
+      // correct — it catches `/nix/store/…/bin/age -d` and `FOO=1 age -d`.
+      // Pinned by three MUST_ALLOW rows in test_opencode_config.py.
+      "*age -d*": "ask",
+      "*age --decrypt*": "ask",
 
       // --- whole-system state ---
       "*nixos-rebuild*": "ask",
       "*home-manager*switch*": "ask",
       // 🔴 LOAD-BEARING — do not delete it as "already covered".
-      // `"*age*-d*"` DOES match some spellings of this command by pure
-      // coincidence ("nix-collect-gar(BAGE) … -d"), but ONLY those containing a
+      // `"*age -d*"` DOES match some spellings of this command by pure
+      // coincidence ("nix-collect-gar(BAGE) -d" ends in the literal "age -d"),
+      // but ONLY those spelling the flag immediately after, so
       // literal `-d`. MEASURED: with this rule deleted, `nix-collect-garbage`
       // and `nix-collect-garbage --max-freed 1G` both resolve ALLOW at both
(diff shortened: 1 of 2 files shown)


ZacxDev:
📌 **Correction to this PR's description, posted publicly rather than edited in** — a reviewer may already have read the original framing.

The PR body presents this as pure false-positive removal. **It is not.** Requiring the literal `age -d` means the rule now knows exactly one argument order, so six reordered decrypt spellings lose their gate:

```
age -i key.txt -d secret.age
age -i key.txt --decrypt secret.age
age -o plain.txt -d secret.age
age --identity key.txt -d secret.age
age --identity key.txt --decrypt secret.age
age --output plain.txt --decrypt secret.age
```

🔴 **`guard_core.py` has no age/sops check at all** (0 hits for `sops`/`decrypt`), so unlike the neighbouring `rm` gap there is no second layer failing in the opposite direction — this glob is the only control. Each spelling above was executed against a real keypair and printed plaintext with `rc=0`.

Fixed in `1391c7f5`, which does not close the gap but stops it being invisible:

- `test_age_glob_misses_these_reordered_decrypt_spellings` — a **characterization** test asserting today's wrong answer on purpose, mirroring the existing `test_rm_glob_misses_these_recursive_spellings`. Both controls run: it executes under `-k` rather than being silently deselected, and re-widening the glob to infix makes it fail with its own message.
- A header carve-out. The file states `EVERY PATTERN IS LEADING-* AND INFIX-*` as policy and argues an over-firing ask beats one that never fires; the age rules now contradict that in the direction the header calls worse. Without naming the exception, the next reader "restores consistency" and re-introduces the outage.

Severity is bounded and worth stating honestly: the house secret path is `sops -d`, which stays infix-gated and is unaffected; `ask` is friction rather than enforcement by this file's own design; and reading a secret is not irreversible. The structural fix is an argv-aware check in `guard_core.py`, which is where this file's header says secret families belong.

**Same defect class, still live, deliberately not fixed here:** `"*rm*-r*"` over-matches identically — `git log --format=oneline --reverse` and `terraform plan -refresh=false` both resolve `ask` today and are auto-rejected. Ordinary agent commands, killing runs right now. It needs its own PR; flagged so this change isn't read as a family sweep.

PR #709, opened 2026-08-22
feat(check-clickup-addressed): migrate the skill out of datapacket-talos into devrc

ZacxDev:
## What

Moves the `check-clickup-addressed` skill out of `datapacket-talos` (where it was
project-scoped to one client repo) into devrc:

| | from | to |
|---|---|---|
| skill body | `<datapacket-talos>/.claude/skills/check-clickup-addressed/SKILL.md` | `claude/skills/check-clickup-addressed/SKILL.md` |
| validation record | `.../README-validation.md` | `claude/skills/check-clickup-addressed/reference/validation-history.md` |
| pipeline | `.../scripts/*.py` | `scripts/check-clickup-addressed/*.py` |
| suite | `.../test/*.py` | `scripts/check-clickup-addressed/tests/*.py` |

The tool is about ClickUp tickets and `~/.claude/projects` transcripts — neither is a
talos concern. It is now a global skill on both hosts, and its **161** tests are a
registered `HERMETIC_TARGETS` entry of `scripts/run-tests.sh` (floor **153**). **No gate
had ever run this suite**; it was invoked by hand.

Layout follows the devrc majority (`session-analysis`, `repo-cos`, `signal`…) rather
than the clickup-skill shape, for one concrete reason: `githooks/tests-on-push.sh`
filters on `^(scripts/|flake\.nix|flake\.lock)$`, so a suite living under `claude/`
would not gate its own edits.

## The self-run guard: what the move broke, corrected

`_selfrun.py`'s markers are **path fragments**, and the two layouts spell the same two
segments in reverse order, so the old marker matched no invocation after the move.

🔴 **An earlier version of this PR said that made the guard "go quiet". Measured, that
was wrong**, and the correction is the more useful fact. The path markers catch the
*invocation*; the report header `## Task Completion Status` catches the *output*, and
every text-mode run prints it. Over the 762 transcripts these scripts actually walk, the
exclusion set is **11 with or without** the path marker. The ordinary path was never
blind. The path markers stay because they uniquely cover a run whose output never reaches
the transcript (piped to a file, `| jq`, truncated) — defence in depth, at that size.

An adversarial audit then found the shapes that *were* broken:

1. **The marker missed the invocation this skill's own SKILL.md teaches.** The quick start
   is `CCUA=…` then `python3 "$CCUA/check-addressed.py"` — the variable means the path
   adjacency never reaches the transcript, and `cd <dir> && python3 check-addressed.py`
   breaks it the same way. A marker for a shape nobody types is not a marker. The four
   entry-point basenames are alternatives now; each measured in the anchored class
   (check-addressed.py 12/1.6%, the others 10/1.3%) beside the old path marker (11/1.4%)
   and the rejected bare name (96/12.6%).
2. **A `--json` run carried no marker at all** — the header was on the text branch, so a
   JSON report, which lists every task ID beside `likely_addressed`, was invisible to the
   guard. `render_json()` now emits it as a field, and `SELF_RUN_HEADER` is one literal
   shared by both branches and pinned against `SELF_RUN_MARKERS`.

## The directory spelling was measured and rejected

The obvious counterpart to the old marker is `scripts/check-clickup-addressed/`. It
re-creates the defect `_selfrun.py` already rejects the bare name for: devrc's `CLAUDE.md`
maps `scripts/<dir>/` to its owning skill and is injected into every session in this repo
(this PR adds that row), and the gate prints a per-target line naming the same directory.

| candidate | matched | |
|---|---|---|
| `check-clickup-addressed` (bare name) | 96 (12.6%) | already rejected |
| `scripts/repo-cos/` (sibling CLAUDE.md row, proxy) | 83 (10.9%) | what the directory spelling costs |
| `scripts/session-analysis/` | 72 (9.5%) | ” |
| `check-clickup-addressed/scripts/` (old marker) | 11 (1.4%) | anchored |

So the marker requires a `.py` file. `test_a_mention_of_the_scripts_DIRECTORY_is_not_a_self_run`
uses the real `CLAUDE.md` row and the real gate line as fixtures.

## 🔴 A test in this PR was vacuous — the round's real lesson

The first `--json` test built the payload itself with the same dict-merge the code uses,
so the mutant that removed the marker **survived a green suite**. Calling the production
`render_json` killed it and left a **second** mutant alive: swapping the `--json` branch
back to a bare `json.dumps` was still green, because every assertion was scoped to the
function and none to the **call site**. `test_main_json_output_is_self_marking_END_TO_END`
now drives the real `main()` with `run_script` stubbed and asserts on what it printed.
Both mutants die, each naming its own test; the isolated-renderer test stays as the
narrower control.

## Verification

- Every new guard watched RED before its fix, and the mutation battery re-run after each
  fix round rather than only the last.
- **161** collected/passed under pytest **and** 161 passed / 0 failed under
  `tests/run_all.py` — two runners, one number. Floor 153 = `_suggested_floor(161)`.
- `--check-targets` / `--check-floors` PASS; the table has exactly one entry per list
  after the rebase.
- **End-to-end from the new path**: `check-addressed.py --fast --limit 2` reached ClickUp,
  produced a report, and reported 7 prior runs ignored — the guard firing on real data.
- Full hermetic gate: `scripts/check-clickup-addressed/tests` **PASS (161/161)**.
- Rebased onto current `main`, which carries the `signal/tests` floor fix (#708).

⚠️ **One unrelated failure in the last full-gate run, not from this branch**:
`scripts/browser-bridge/tests::test_wrapper_forces_tab…` hit its 60s hang-net during a run
concurrent with another dispatched agent. It passes at `origin/main` alone (0.31s) and the
whole 60-test browser-agent suite passes on this branch alone (19.28s); this branch touches
no browser-bridge file. Re-running the full gate on a quiet box before merge.

## Follow-ups (not in this PR)

- `datapacket-talos`: delete the old copy and mark its handoff superseded — held until this
  merges + `home-manager switch`, so the skill is never absent from both.
- A stale unmanaged copy at `~/.config/opencode/skills/check-clickup-addressed/` holds the
  old layout and is not nix-managed.
- `SKILL.md` is ~21 KB — a `/prune-skill` candidate.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -136,6 +136,7 @@ is preserved verbatim in `docs/LAYOUT.md` (not auto-loaded, and stale by design)
 | `scripts/initiatives/` | `initiatives` | durable cross-repo initiative ledger + viewer + router + assistant |
 | `scripts/mail-actions/` | `mailbox` | email-automation layer over the self-hosted inbox (**separate from activity telemetry**) |
 | `scripts/repo-cos/` | `repo-cos` | weekly repo "chief-of-staff" proposal digest + reply-driven exclusions |
+| `scripts/check-clickup-addressed/` | `check-clickup-addressed` | did the work on a ClickUp ticket actually happen — reads session transcripts for completion signals (migrated out of datapacket-talos 2026-08-22) |
 | `nix/i3/`, `nix/graphical.nix`, `scripts/bar-status-poll` | `bar` | i3 + i3status-rust bar, count blocks, dunst toasts |
 | `scripts/opencode/` | `opencode` | dispatch a task to the headless opencode agent (`opencode-dispatch`), + its config/agents/guard plugin |
 
--- a/claude/skills/check-clickup-addressed/SKILL.md
+++ b/claude/skills/check-clickup-addressed/SKILL.md
@@ -287,7 +287,7 @@ further drift. **Re-time it rather than quoting it.**
 
 ## Tests
 
-Run all tests (**156** collected, measured 2026-08-22). `run_all.py` exits non-zero on
+Run all tests (**157** collected, measured 2026-08-22). `run_all.py` exits non-zero on
 failure — but read the `Total: N passed, M failed` line, not a piped exit code. 🔴 **If that
 line is missing at all, the run died — treat it as a failure, never as "no output".** A
 `sys.exit()` from code under test is a `SystemExit`, which a bare `except Exception` does
@@ -301,7 +301,7 @@ PYTHONDONTWRITEBYTECODE=1 python3 "$CCUA/tests/run_all.py"
 
 The same files are a **pytest** target of devrc's gate — `scripts/check-clickup-addressed/tests`
 in `HERMETIC_TARGETS`, with its collected-count floor in `TARGET_FLOORS`
-(`scripts/run-tests.sh`). Both runners see the same 156; `run_all.py` survives because it
+(`scripts/run-tests.sh`). Both runners see the same 157; `run_all.py` survives because it
 purges `__pycache__` and reports an import failure as a FAILURE, which pytest's summary
 line does not distinguish as loudly. **Raise the floor when you add tests.**
 
--- a/claude/skills/check-clickup-addressed/SKILL.md
+++ b/claude/skills/check-clickup-addressed/SKILL.md
@@ -1,0 +1,329 @@
+---
+name: check-clickup-addressed
+description: Verify if ClickUp tasks were fully addressed by searching session transcripts for completion signals. Use when checking if recent work on assigned tasks is complete, or when asked to verify task completion status.
+---
+
+# /check-clickup-addressed — verify if ClickUp tasks were fully addressed
+
+Deterministic check: given the N most recent comments not from you on your assigned
+tasks, find the sessions that worked on those tasks, read the transcripts, and report
+what's done vs what's still open.
+
+## Quick start
+
+The scripts live in **devrc**, not in this deployed skill dir — `~/.claude/skills/` is a
+read-only nix-store copy of `claude/skills/`, and only the docs ship there. Run them from
+the repo:
+
+```bash
+CCUA=~/workspace/devrc/scripts/check-clickup-addressed
+
+# Default: top 3 most recent comments (~30s)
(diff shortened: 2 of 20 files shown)


ZacxDev:
Heads-up on an overlap with something that landed on `main` today, and a proposal to wire them together rather than let both auto-fire.

**#675 added an `awaiting` command to the `clickup` skill** (`node query.mjs awaiting`). Its predicate is *"the newest comment on this task was not authored by the token owner"* — which is essentially the **first stage** of what this skill does, minus the transcript correlation.

I recommended and shipped it after concluding that a ClickUp "was this addressed?" check was largely not computable: the API does not return `resolved` on task comments (measured: union of every field across 84 comments / 29 tasks — it never appears, not even as `false`), and ClickUp has no bot identity, so a mirror/agent comment is indistinguishable from a human one. **That conclusion was too strong, and this skill is why** — it routes around the missing authorship by reading session transcripts, a source I never considered. Credit where due.

**They are complementary, not duplicative:**
- `awaiting` — ~45s, ~53 API calls, no transcript reads. Answers *which* tasks have someone else's comment last. Prints matched beside examined, flags truncation, and states the no-bot-identity blind spot in its own output.
- `check-clickup-addressed` — reads the sessions that worked the task. Answers *whether the work was actually done*.

**Proposal, both directions:**
1. This skill's body names `clickup`'s `awaiting` as the cheap first pass when you only need the triage list.
2. Once this merges, I add `check-clickup-addressed` to `clickup`'s description sibling ledger.

On (2) — I cannot do it before this lands, and that is deliberate: `scripts/tests/test_skill_descriptions.py` (added in #662) pins `CLICKUP_SIBLINGS` as an exact set and **asserts every sibling named in the description exists as a skill**, so naming a not-yet-present skill turns the suite red. Ping me when this merges and I will open that PR.

One thing worth checking in this migration, from having just been bitten by it: **`awaiting`'s signal decays.** #674 confirmed from the live cluster that clickup-mirror is `MIRROR_DIRECTION=both` with `writeback.enabled: true` — it posts to ClickUp **as Zach**, hourly. So any predicate keyed on "last commenter is not me" silently flips tasks from unaddressed to addressed with no human involved. Transcript-based detection is immune to that, which is another argument for keeping both.

PR #713, opened 2026-08-22
fix(gate): the signal floor had fallen 164 behind — a whole suite could vanish under it while the gate stayed green

ZacxDev:
`scripts/signal/tests` collects **717**; its `TARGET_FLOORS` entry still said **553**. That is 164 of slack against a ceiling of 138 (floor × 0.25), so the drift check failed the run **with every single test passing** (716 passed, 1 skipped — the pre-existing `SIGNAL_PG_DSN` skip, so `EXPECTED_SKIPS` is untouched).

The replacement is **copied from the gate, not computed here**, per the rule this table documents:

```
run-tests: ERROR — scripts/signal/tests collected 717 tests but its floor is only 553.
  That is more than 138 of slack: a whole suite could vanish under this
  floor with the gate still green. Raise the TARGET_FLOORS entry to
  "scripts/signal/tests|682" — that number is this run's own count put through
  the documented rule, so it needs no measurement of its own.
```

## How this surfaced — it is the gate's first catch

This is the `devrc-pytests` leg of **`devrc-ci-djm7n`, the first Tekton PipelineRun this repo has ever had**. The gate went live and immediately caught a real staleness the hand-run process had missed. The drift ceiling was the **only** thing it failed on; `tekton/devrc-nodetests` passed **1119/1119**.

## It also corrects a stale belief

The handoff recorded `devrc-pytests` as unrequireable because "the nix sandbox does not exist in the pod, and all three levers are rejected by PodSecurity `baseline:latest`". **It does exist** — `seed-nix` exited 0 and **14,686 tests ran in-pod**. Once this lands, `devrc-pytests` should go green and the "cannot be required" verdict deserves re-testing over a few runs.

## Verification

Deliberately **not** run locally: devrc's test tier mutates real git repos, and running it from a worktree that shares the base clone's git dir is the failure recorded in the handoff (and in open PRs #673/#676/#683/#689). The authoritative check here is the Tekton gate itself, which runs `run-tests.sh` in an isolated pod — so this PR is verified by the very gate it fixes. If 682 is wrong, the gate prints the correct number and this is a one-line follow-up.

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -1226,7 +1226,20 @@ TARGET_FLOORS=(
   # rots is silent: an anchor stops matching, the mutant never lands, and the
   # run prints ANCHOR-MISS, which reads like a hiccup rather than "this mutant
   # tested nothing". That has already happened here once.
-  "scripts/signal/tests|553"
+  # 2026-08-22: 717 tripped the ceiling at 691 with every test PASSING (716 passed,
+  # 1 skipped — the pre-existing SIGNAL_PG_DSN skip, so EXPECTED_SKIPS is untouched).
+  # The gate printed "scripts/signal/tests|682" and that is what is below, copied not
+  # computed. 🔴 FIRST DRIFT CAUGHT BY THE TEKTON GATE RATHER THAN BY HAND — this is
+  # the devrc-pytests leg of devrc-ci-djm7n, the first PipelineRun the repo has ever
+  # had, and the drift ceiling was the ONLY thing it failed on. The floor had fallen
+  # 164 behind: a whole suite could have vanished under it while the gate stayed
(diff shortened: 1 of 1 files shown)


ZacxDev:
Superseded by #708, which landed the identical value (`scripts/signal/tests|682`) while this was open — both copied the same number from the gate's own output, so there is nothing here that #708 does not already do.

Recording the one thing this PR carried that #708's title did not: the drift was surfaced by **`devrc-ci-djm7n`, the first Tekton PipelineRun this repo has ever had**. It also disproves the handoff's belief that `devrc-pytests` is unrequireable because "the nix sandbox does not exist in the pod" — it does: `seed-nix` exited 0 and 14,686 tests ran in-pod. That correction is captured in #706.

PR #716, opened 2026-08-22
fix(scripts): an EMPTY repo override resolved to the operator's REAL clone — `${VAR:-default}` cannot tell unset from empty

ZacxDev:
## The bug

`ship.sh`, `drift-check.sh` and the analyze-service index's `commit.sh` each resolve a repository path with `${VAR:-$HOME/…}`. **That form cannot tell UNSET from SET-BUT-EMPTY.** A caller whose own path computation returned `""` does not get an error — it silently gets `$HOME/workspace/devrc` (or `$HOME/.claude/analyze-service-index`), and the script then fetches, converges, `git add`s and commits **in the operator's real clone or real index store** while believing it was handed a sandbox.

Six sites, all of them live:

| file | site |
|---|---|
| `scripts/ship.sh` | `SHIP_REPO="${SHIP_REPO:-$HOME/workspace/devrc}"` ×2 |
| `scripts/drift-check.sh` | `DRIFT_REPO="${DRIFT_REPO:-$HOME/workspace/devrc}"` ×3 |
| `scripts/analyze-service-index/commit.sh` | `STORE="${POSITIONAL[0]:-${HOME}/.claude/analyze-service-index}"` |

## Why this is not `${VAR:?}`

**UNSET must keep defaulting.** The remote legs of `ship.sh` and `drift-check.sh` deliberately do *not* forward these variables over ssh — the far host's repo lives at its own `$HOME/workspace/devrc`. A guard that failed on unset too would break the remote leg of both scripts. So each site gets a set-but-EMPTY guard that leaves unset alone.

`drift-check.sh` carries the guard **three times on purpose**: the `CHECK` payload and the `SRCREPO` payload are each piped to a `bash -s` on another host and cannot source anything, so each needs its own copy.

## The tests

- **The ledger is pinned BOTH ways.** A new HOME-defaulting repo path with no guard fails; so does a ledger entry naming a line that no longer exists. A mechanical scan finds the sites so a seventh cannot be added silently.
- **EVERY occurrence, not the first.** Two of the lines are byte-identical (`drift-check.sh`'s CHECK and SRCREPO payloads); matching only `text.index(line)` would have declared the second guarded on the strength of the first — a description claiming coverage its body does not provide.
- **Behavioural, not just structural.** A fake `HOME` contains a *real* repository at `workspace/devrc` — the place an empty value resolves to. The guard must stop the run, and that repo must be byte-identical afterwards.
- **The EXACT exit status 2, not merely non-zero.** Measured: a mutant that kept the message and dropped the `exit 2` **survived** an earlier version of this test — the script carried on to the default, converged the victim, and failed later for an unrelated reason, so "non-zero" and "EMPTY appears in the output" were both still true.

## Red/green matrix

Measured on a standalone clone with `origin` **removed**, base `a34d695d`:

```
base a34d695d (guards reverted, test file present):   7 failed, 2 passed
with the guards:                                      9 passed
```

The two green in **both** arms are the ledger check and `test_an_unset_repo_override_still_defaults` — the latter is the control against over-guarding and must be green in both. A test that only ever passed would prove nothing.

## Provenance

These guards were written for **#689** and are salvaged here on their own. They are **independent of that PR's git interceptor** — plain caller-bug guards in three production scripts — and they stand whatever happens to the guard architecture. #689's `test_no_real_repo_writes.py` interleaves them with 2,300 lines of interceptor tests; this extracts them into `scripts/tests/test_repo_path_defaults.py` with only the helpers they need.

## Not covered

This is the `${VAR:-default}` class only. It does **not** address the test tier writing to the operator's clone — that is the #673/#683 guard work.

### DIFF
--- a/scripts/analyze-service-index/commit.sh
+++ b/scripts/analyze-service-index/commit.sh
@@ -168,6 +168,19 @@ if [ "${#POSITIONAL[@]}" -gt 1 ]; then
   usage: commit.sh [--print-plan] [STORE_DIR]"
 fi
 
+# 🔴 A GIVEN-BUT-EMPTY STORE IS A BUG, NOT A REQUEST FOR THE DEFAULT. `:-`
+# cannot tell "no argument" from "an argument that is the empty string", so
+# `commit.sh ""` — the shape a caller produces when its own path computation
+# fails — would silently `git init`, `git add` and `git commit` in the
+# OPERATOR'S REAL STORE instead of the directory it meant. This is the same
+# class as ship.sh's SHIP_REPO and drift-check.sh's DRIFT_REPO; see
+# scripts/tests/test_no_real_repo_writes.py, which pins all four sites.
+if [ "${#POSITIONAL[@]}" -eq 1 ] && [ -z "${POSITIONAL[0]}" ]; then
+  die "STORE argument was given but is EMPTY.
+  That is a caller bug, not a request for the default — an empty value would
+  silently resolve to \${HOME}/.claude/analyze-service-index and commit in the
+  operator's real store. Pass no argument for the default, or pass a path."
+fi
 STORE="${POSITIONAL[0]:-${HOME}/.claude/analyze-service-index}"
 
 ASI_GIT_NAME="${ASI_GIT_NAME:-analyze-service index}"
--- a/scripts/analyze-service-index/commit.sh
+++ b/scripts/analyze-service-index/commit.sh
@@ -174,7 +174,8 @@ fi
 # fails — would silently `git init`, `git add` and `git commit` in the
 # OPERATOR'S REAL STORE instead of the directory it meant. This is the same
 # class as ship.sh's SHIP_REPO and drift-check.sh's DRIFT_REPO; see
-# scripts/tests/test_no_real_repo_writes.py, which pins all four sites.
+# scripts/tests/test_repo_path_defaults.py, which pins all SIX sites in both
+# directions and asserts each guard actually STOPS the run, not merely warns.
 if [ "${#POSITIONAL[@]}" -eq 1 ] && [ -z "${POSITIONAL[0]}" ]; then
   die "STORE argument was given but is EMPTY.
   That is a caller bug, not a request for the default — an empty value would
--- a/scripts/drift-check.sh
+++ b/scripts/drift-check.sh
@@ -359,15 +359,16 @@
 # is reading a timer's output, so the single number it hands to systemd must be
 # the worst thing found, or an un-pushed workbench could hide behind a merely
 # behind laptop. Severity order (worst first):
-#     8 > 17 > 14 > 13 > 18 > 6 > 4 > 3 > 12 > 15 > 10 > 16
+#     8 > 17 > 14 > 13 > 18 > 6 > 2 > 4 > 3 > 12 > 15 > 10 > 16
 # 🔴 THAT ORDER IS THE severity() TABLE, NOT THE DIGITS — it never was monotonic
 # (14 outranks 13 outranks 18 outranks 6 outranks 4), and 17 is not "less severe
 # than 16" because it is larger. Every code below 16 that is still free (5, 7, 9, 11) is
 # reserved to a ship.sh meaning this script does not take, so a new DRIFT code
 # has nowhere to go but upward; its rank is stated in severity() and here.
-# (6 is unreachable through this path today — the script exits 6 directly before
-# any per-host leg runs — but the order is documented for every code it owns,
-# and severity() ranks it rather than falling through to the unknown-code slot.)
+# (6 and 2 are both unreachable through this path today — the script exits each
+# directly, before any per-host leg runs — but the order is documented for every
+# code it owns, and severity() ranks them rather than letting them fall through
+# to the unknown-code slot, which would rank them 99, ABOVE rc 8.)
 # Per-host lines are ALWAYS printed for every host, whatever the code.
 #
 # Untracked files are counted and listed per host as INFORMATION only — they
--- a/scripts/drift-check.sh
+++ b/scripts/drift-check.sh
@@ -82,11 +82,15 @@
 # take (5 conflicted-tree, 7 cannot-ff, 9 switch-failed, 11 verify-failed) are
 # left UNUSED rather than repurposed.
 #
-#   2   usage error (unknown flag, non-integer tunable), or a RUN THAT CHECKED
+#   2   usage error (unknown flag, non-integer tunable), a RUN THAT CHECKED
 #       NO HOST AT ALL — either because the flags asked for none (`--no-local
 #       --no-remote`) or because the only host it was asked to look at could not
-#       be reached. rc 0 from a run that observed nothing is the vacuous green
-#       this whole subsystem exists to prevent, so it is never emitted.
+#       be reached (rc 0 from a run that observed nothing is the vacuous green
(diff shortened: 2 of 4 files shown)


ZacxDev:
## Gate: PASS on the MERGED tree (not the branch)

Run in a standalone clone with `origin` **removed** (verified empty before anything executed), against `main` **including #683**, which merged after this PR was opened:

```
pytest  TOTAL collected=14833  passed=14831  skipped=2  failed=0   (floor 13529)
node    TOTAL suites=4 files=33 tests=1119 pass=1119 fail=0        (floor 1098)
GATE: RESULT=PASS exit=0
```

Zero `FAIL` lines across all 26 pytest targets and all 6 script targets. Every per-target floor clears.

### Positive control — the green includes THIS PR's tests

A passing suite is not evidence that the new tests ran, so measured separately on the same gated tree:

```
$ pytest scripts/tests/test_repo_path_defaults.py --collect-only -q
9 tests collected

$ pytest scripts/tests/test_repo_path_defaults.py -q
9 passed
```

### Why it was re-gated

The original red/green matrix in the PR body was measured at base `a34d695d`, before **#683** landed. #683 installs a per-test detector that fingerprints the user-level git config by reading `GIT_CONFIG_GLOBAL` — and these tests set `GIT_CONFIG_GLOBAL` in their subprocess environments. That is precisely the interaction that made a *different* guard pairing (#683 + #673) produce 39 failures, so it was worth measuring rather than reasoning about. It does not fire here: these tests set the variable only in the child's env, never in `os.environ`, so the plugin's import-time snapshot is unaffected.

ZacxDev:
## 🔴 Correction to this PR's body, plus the audit fixes (`d7b31f2d`)

Posting rather than silently editing the description, because a reviewer may already have read and believed the wrong version.

### The body's claim about the over-guard control was FALSE

It said:

> The two green in **both** arms are the ledger check and `test_an_unset_repo_override_still_defaults` — the latter is the control against over-guarding, and it must be green in both.

That control ran against **`drift-check.sh` only**. `ship.sh` had no unset arm at all, so the claim covered one of the two scripts it named. The audit proved the gap live: a mutant keeping every literal spelling the structural test greps for, plus `|| [ -z "${SHIP_REPO:-}" ]`, left the suite **green** while `ship.sh` with `SHIP_REPO` unset — the normal invocation, and the only one the remote leg ever gets — exited 2 and converged **nothing on either host**.

Now parametrised over both scripts.

### The main finding: the test asserted the CONDITION, never that it STOPS

`test_each_site_is_immediately_preceded_by_a_set_but_empty_guard` only checked that the condition *text* appeared within 25 lines above each assignment. Nothing asserted the run actually stopped — while the file's own docstring boasted of having measured and killed the "keep the message, drop the exit" mutant. **It killed it at two sites; the sentence covered six.**

Driven end to end by the audit: a mutated `commit.sh` that warned instead of dying went on to `git init`, `git add` and `git commit` inside a fake operator store, while this suite reported **9 passed**. That is the repo's own named failure — *a guard's DESCRIPTION claims COVERAGE its body does not provide*. The terminator (`exit`/`die`/`return`) is now asserted separately.

### Two more holes closed

- **Every assertion in the unset control was negative**, so an `exit 6` inserted before the assignment left it green — nothing checked the script had reached anywhere. Added the positive assertion: both scripts announce each host before doing work, which is exactly the signal the empty-value arm requires to be **absent**. The two arms now pin the same observable in opposite directions.
- **The scanner was far narrower than its comment.** `_HOME_DEFAULT_RE` missed **17 of 19** plausible shapes — `export`, `local`, unquoted, trailing comment, and the `~/` form the comment named explicitly. Not hypothetical: two live sites already sat outside it, so *"a SIXTH site cannot be added silently"* was **false when written**. Widened, with its own positive and negative controls (`:=` and `:?` are different operators and must not match).

### Widening the regex found two MORE unpinned sites

Neither the original ledger nor the audit had them: the `KUBECONFIG` defaults in `scripts/initiatives/run-sync.sh` and `scripts/collector/run-regrowth-check.sh`.

They are **ledgered, not hidden**. A new `READ_ONLY_SITES` records every site of this shape whose consumers only READ (`git -C … show`, an age-key `[ -r ]`, a kubectl config), so a set-but-empty value mis-targets a read rather than writing the operator's clone. It is explicitly **not a waiver list** — an entry that gains a writing consumer must move up into `EMPTY_DEFAULT_SITES` — and the scanned file set is itself pinned against both ledgers in both directions.

### Also fixed

- **A pointer to a file that does not exist.** `commit.sh` cited `test_no_real_repo_writes.py`, "which pins all four sites" — that was #689's file and did not come across with the extraction. It was the only in-code pointer to the guard's test, so a maintainer following it would find nothing and conclude the guard was untested. Now names the real file and the real count (six).
- **`drift-check.sh` rc 2** is emitted for set-but-EMPTY but appeared in neither the EXIT CODES header nor `severity()`. Per that function's own rc 6 note, an owned code with no case ranks **99 — above rc 8**, the file's designated dangerous one. Both fixed, ranking reasoned in place (57: below 6, above 4/3, because an empty override is deterministic rather than a might-be-gone-next-run failure).
- Behavioural case added for an empty `STORE` argument — the third site had structural coverage only, which is exactly what the audit exploited.

### Mutation results — all five previously-surviving mutants now die

Baseline **13 passed**, restore **13 passed**, swept under `PYTHONDONTWRITEBYTECODE=1` so a stale bytecode cache cannot score a mutant SURVIVED without ever running it. Each is killed by **its own** assertion, not a neighbour's error:

| mutant | killed by |
|---|---|
| `commit.sh` `die`→`warn` | `[commit.sh:POSITIONAL]` **+** the new behavioural test |
| `ship.sh` payload drops `exit 2` | `[ship.sh:SHIP_REPO0]` **+** the empty-value arm |
| `drift-check.sh` payload drops `exit 2` | `[drift-check.sh:DRIFT_REPO1]` |
| `ship.sh` over-guard fires on UNSET | the new unset arm `[ship.sh]` |
| `exit 6` before the assignment | the new positive assertion `[drift-check.sh]` |

### Gate — merged tree, standalone clone, `origin` removed

```
pytest  TOTAL collected=14864  passed=14862  skipped=2  failed=0   (floor 13529)
node    TOTAL suites=4 files=34 tests=1149 pass=1149 fail=0        (floor 1126)
GATE: RESULT=PASS exit=0
```

Zero `FAIL` lines. `shellcheck -S warning` clean on all three scripts. 334 passed across `test_repo_path_defaults` + `test_drift_check` + `test_ship_converge` + `test_ship_detect_role`.

### Confirmed unchanged from the audit, worth recording

- Guard semantics correct in **bash 5.3, zsh, dash and busybox sh**; no `set -u` exposure.
- **The escaping survives the ssh trip** — verified against the exact remote form parsed by a zsh login shell, and `bash -s` for drift-check's payloads.
- **Blast radius nil**: no production caller sets `SHIP_REPO`/`DRIFT_REPO`; `nix/home.nix:3063` invokes `commit.sh` with no positional arg; the `drift-check` unit sets no `Environment=DRIFT_REPO`.

### Known residual, stated rather than implied

The three payload guards (`ship.sh` CONVERGE, `drift-check.sh` CHECK and SRCREPO) **cannot execute in production today** — the local leg passes an already-defaulted value and the remote leg does not forward the variable. They are structurally asserted and mutation-tested but never executed by a test. Three copies to keep in sync for a path that is currently unreachable; kept because the payloads are piped to `bash -s` on another host and cannot source anything, so if a caller ever does forward the variable, each needs its own guard.

PR #718, opened 2026-08-22
docs(CLAUDE.md): "no checks run" went stale when Tekton appeared — but nothing BLOCKS, so the marker stays `none`

ZacxDev:
## What was wrong

`CLAUDE.md`'s merge-gate paragraph made four factual claims. Measured **2026-08-22**, three had gone stale:

| claimed | measured |
|---|---|
| "no Tekton trigger names devrc" | 🔴 `tekton/devrc-pytests` and `tekton/devrc-nodetests` report on devrc PRs |
| "`main` has no branch protection or rulesets" | 🔴 protection **exists** — `enforce_admins`, no force-pushes, no deletions |
| "`statusCheckRollup` returns **0 checks** on every PR" | 🔴 2 checks on recent PRs |
| "there is no `.github/workflows`" | ✅ still true |

## Why the marker still says `none`

Because **a check that RUNS is not a check that GATES**, and that distinction is the entire job of this line.

The branch protection carries **no `required_status_checks`**. So a red Tekton run does not stop a merge, and a missing one does not either. Nothing blocks.

```
gh api repos/innovation-upstream/devrc/branches/main/protection
```

`required_status_checks` is **absent** from the response. That is the tell, and it is the easy one to miss — an absent key reads as clean rather than missing unless you go looking for it by name, so both the command and the key are now named inline.

Moving the marker to `other` because checks appeared would assert a protection that does not exist. That is the same reassuring falsehood the line was rewritten to kill (`CI gates both suites`), just pointing the other way. `other` is reserved for something that *actually gates* and that the test cannot see — an installed `githooks/` pre-push gate, or Tekton once it becomes a required check.

## What this does not change

`scripts/tests/test_ci_claim_matches_reality.py` is untouched, and it still passes — it pins the marker against `.github/workflows`, which remains empty. Its documented blind spots are unchanged and still accurate: it cannot see branch protection, Tekton, or git hooks. This PR only corrects the prose those blind spots are supposed to be covered by, and adds the re-measurement instruction, since **both halves of the claim have now been wrong in both directions** — first "CI gates both suites" (nothing ran), now "no checks run" (they do).

Verified: exactly one real marker (the second `merge-gate:` hit is the `…` placeholder in the explanatory paragraph, which the value regex does not match), the marker's line still reads as a warning to a human, and `test_ci_claim_matches_reality.py` + `test_rules_size.py` pass — 10 passed.

## Provenance

Found while resolving the test-tier guard-strategy question (#683 merged, #676/#689 closed). It matters there specifically: the guard PRs showed 0 checks purely because they predate Tekton, and an agent reading the old paragraph would have concluded the tooling had not changed.

### DIFF
--- a/scripts/tests/test_ci_claim_matches_reality.py
+++ b/scripts/tests/test_ci_claim_matches_reality.py
@@ -2,11 +2,32 @@
 
 WHY THIS EXISTS
 ---------------
-For an unknown span, `CLAUDE.md` asserted "**CI gates both suites**: `nix build
-.#checks…`". No CI has ever run on a devrc PR: no `.github/workflows`, no branch
-protection, no ruleset, no Tekton trigger, and `statusCheckRollup` is empty on
-every PR including merged ones. Every merge has rested on a human or agent
-running the suite by hand.
+From 2026-08-02 to 2026-08-20, `CLAUDE.md` asserted "**CI gates both suites**:
+`nix build .#checks…`" while nothing ran at all. Every merge rested on a human
+or agent running the suite by hand.
+
+🔴 THIS PARAGRAPH THEN BECAME THE THING IT WARNS ABOUT, and is corrected in
+place rather than deleted, because the drift is the lesson. It used to continue:
+"No CI has ever run on a devrc PR: no `.github/workflows`, no branch protection,
+no ruleset, no Tekton trigger, and `statusCheckRollup` is empty on every PR."
+**Three of those five clauses were measured FALSE on 2026-08-22** — Tekton posts
+`tekton/devrc-pytests` and `tekton/devrc-nodetests` on devrc PRs, `main` DOES
+carry branch protection, and `statusCheckRollup` is non-empty. Still true: no
+`.github/workflows`, and no ruleset.
+
+None of that changed the marker to `github-actions`, because the protection
+carries **no `required_status_checks`**: checks RUN, nothing BLOCKS. Measured
+behaviourally, not inferred — #707 merged **28 minutes** after its
+`devrc-pytests` went RED, #711 two minutes after. The marker is `other` because
+something this test cannot see (Tekton) now reports; it would become
+`github-actions` only if a `pull_request` workflow appeared here.
+
+Re-measure both surfaces rather than quoting any of the above:
+    gh api repos/innovation-upstream/devrc/branches/main/protection   # required_status_checks?
+    gh api repos/innovation-upstream/devrc/rules/branches/main        # org+repo+enterprise rulesets
+On the first, an ABSENT key reads as clean unless you look for it by name; on
+the second the answer is an empty array. Positive-control the `[]` before
+believing it — the same call against a repo with rulesets returns non-empty.
 
 That is the worst shape a false claim can take. It lives in the always-loaded
 project instructions, it is reassuring, and it is exactly the sort of sentence


ZacxDev:
## Blind audit — round 1 findings fixed (`ed0e7eee`)

Verdict was **safe to merge, no 🔴**, with three 🟡. Two of them were defects *this PR introduced*, so they are fixed rather than deferred.

### 1. 🔴→fixed: the chronology was backwards

I wrote that the line read "no checks run" *before* `CI gates both suites`. It is the reverse, and I re-verified it against git rather than taking the finding on trust:

```
c2fe83cf  2026-08-02  #304   <- "CI gates both suites" lands
1fd04d8a  2026-08-20  #603   <- replaced with "nothing gates a devrc merge"
```

The sentence exists to teach *"this claim has been wrong in both directions."* Inverted, it misdates the repo's own correction history inside an always-loaded file. Now carries both dates and both PR numbers.

### 2. 🔴→fixed: the verification recipe was blind to rulesets — a coverage **regression** I introduced

The text I replaced said "no branch protection **or rulesets**". Mine named only `gh api …/branches/main/protection`, which returns **classic** protection. A **ruleset** can require status checks independently and never appears there.

That matters more than a normal doc nit because this paragraph is explicitly built to be re-run — so a wrong recipe is the entire failure mode. A future reader following it exactly would get a confident "nothing blocks" while a ruleset gated the repo.

Now names both surfaces:

```
gh api repos/innovation-upstream/devrc/branches/main/protection   # required_status_checks?
gh api repos/innovation-upstream/devrc/rules/branches/main        # org + repo rulesets vs this branch
```

Both measured `[]`/absent today — so the *verdict* was right and only the *recipe* was wrong. That distinction is why this is a 🟡 and not a 🔴.

### 3. Fixed: the flip condition was never named

The instruction further down still says to change the marker when a `pull_request` workflow appears or disappears — which, with Tekton live, no longer decides anything. Now stated inline and explicitly overriding it: **flip to `other` when `required_status_checks` appears, or a ruleset requires the tekton contexts. Tekton merely RUNNING is not that condition.**

## Added: behavioural proof, which is much stronger than what this PR shipped with

The audit established the central claim empirically rather than from the API, and I have folded it into the file:

- **#707** merged `18:37:33Z` with `tekton/devrc-pytests=FAILURE` completed `18:09:12Z` — **red 28 minutes before the merge.**
- **#711** merged `19:08:00Z` with `pytests=FAILURE` completed `19:05:55Z` — **red 2 minutes before.**
- #715 merged with both `PENDING`; #712 with both `ERROR`.

Red and pending checks demonstrably do not stop a merge. That settles `none` far better than an absent API key does.

## On `none` vs `other`

The auditor argued the opposing case properly and the strongest point stands: the marker's own documentation *names Tekton* as an `other` trigger, so a plain reading instructs `other`. I still believe `none`, for the reason the audit landed on independently — `other` means "something gates that this test cannot see", which is **false today and provably so**, and it fails in the more dangerous direction. A reader who wrongly feels protected skips the run; one who wrongly feels alone just runs the suite unnecessarily. Fix #3 addresses the real weakness, which was in that documentation sentence rather than in the marker value.

There is also repo precedent: `claude/skills/tekton/SKILL.md:51` already calls the sibling repo's pipelines *"DETECTORS, not GATES"* on the same evidence shape.

## Verified after the fix

Marker test + rules-size: **10 passed**. Exactly one marker matching the value regex. `CLAUDE.md` 25,307 → 26,133 B; **no enforced ceiling exists on this file** (`test_rules_size.py` pins `claude/RULES.md` only, untouched at 38,296 / 39,200).

## Out of scope, filed here so it is not lost

`claude/skills/tekton/SKILL.md:100-106` lists three pipelines and six triggers with **no devrc entry** — `devrc-pytests` appears nowhere in that skill or its `reference/`. The skill that owns Tekton does not know devrc CI exists. Worth its own PR.

ZacxDev:
## Round 2 fixes (`6d76106a`) + round 3: **CLEAN** — 🔴 none, 🟡 none

Three audit rounds, stopping here because round 3 produced no blockers and no should-fixes.

### Round 2 findings, all fixed

| # | finding | fix |
|---|---|---|
| 1 | The "flip to `other` **only when**…" sentence silently nullified the **githooks** trigger named in the `other` example 40 lines below — leaving the file with three answers and precedence for one | enumerates all three blocking conditions; names Tekton-merely-running as the *one* half overridden |
| 2 | The ruleset half was **SPELLED**: "requires the tekton contexts" is walkable by a ruleset requiring any differently-named check, which gates just as hard | now "requires **ANY** status check", matching the generality the classic-protection half already had |
| 3 | "In both, the absence is the answer" conflated an absent **KEY** (`protection`) with an empty **ARRAY** (`rules`) | split, plus a positive control for the `[]` and the note that `rules` also surfaces **enterprise**-source rulesets |
| 4 | The test's own module docstring asserted five things, **three measured false** | corrected in place — the drift is the lesson; SCOPE section untouched |
| 5 | Attribution | the claim first landed `c9cb47a8` (#287) 2026-08-02 **17:53**, inside the browser-bridge bullet; `c2fe83cf` (#304) reworded it into the standalone line at **23:06** the same day |

Finding 2 is the one worth calling out: I wrote a **spelled** guard into a file that has a rule against exactly that. A differently-named required check gates just as hard as a tekton-named one.

### Round 3 independently re-verified everything

- **Attribution** exact to the second, including that #287's version landed *inside the browser-bridge bullet*.
- **API** — `devrc/rules/branches/main` → `[]`; `microsoft/vscode` and `anthropics/claude-code` non-empty, the latter literally `"ruleset_source_type":"Enterprise"`. And `required_status_checks` confirmed an **absent key**, not an empty value.
- **Behavioural** — #707 RED 18:09:12 → merged 18:37:33 = **28m21s**; #711 RED 19:05:55 → merged 19:08:00 = **2m05s**.
- **Docstring edit confined** — SCOPE section **byte-identical** (875 B by `cmp`), diff entirely inside the module docstring, **zero executable lines changed**.
- **Guard broken four ways**, each going RED on `test_claude_md_marker_matches_the_repo`'s *own* assertion rather than a neighbour's: `github-actions` → `assert 'github-actions' in {'none','other'}`; `tekton` → `unknown merge-gate marker`; second marker → `assert 2 == 1`; delete the sentence → `assert 1 >= 40`.

### One cosmetic fix taken (`26a5ead3`), no fourth round

Round 3's only nit was a contradiction on this PR's own theme: the round that rewrote the docstring *for staleness* left "For an unknown span" six lines above the paragraph that now dates the span to the minute. Replaced with the measured range. Re-verified rather than assumed — 5 passed, negative control RED on its own assertion, restored green. Not dispatching a fourth round for a two-line rewording with the guard re-broken and re-watched.

**Left as-is:** the "this sentence is the one that wins" precedence claim is unqualified against the `github-actions` rule. Latent only — it needs devrc to gain a GitHub Actions PR workflow, and in that case the test fails **loudly naming the correct value** rather than passing on a falsehood.

### Corrections to claims I made earlier in this PR

- **`CLAUDE.md` is NOT paid twice.** I said it is concatenated into opencode's `AGENTS.md`. It is not — `nix/home.nix:1334-1337` composes that from `PRINCIPLES.md + claude/RULES.md + opencode-addendum.md`. `CLAUDE.md` loads once per session.
- Final size **24,507 → 27,098 B**. No enforced ceiling covers this file; the two gated docs are `claude/RULES.md` and `browser-bridge/SKILL.md`.

### Follow-ups, not fixed here

- `CLAUDE.md` says *"**Two** always-on docs have enforced byte ceilings"* — and the third, unbounded one is the file saying it.
- `claude/skills/tekton/SKILL.md:100-106` lists three pipelines and six triggers with **no devrc entry**. The skill that owns Tekton does not know devrc CI exists.

PR #726, opened 2026-08-22
rules: default to PROCEEDING — a four-branch trigger tree, not three self-assessed predicates

ZacxDev:
## What

Zach proposed a proactivity rule set:

> - if the next step is safe and unblocked, proceed
> - if the next step is blocked, ask, then proceed
> - if the next step is destructive or unsafe, block, warn, and ask

The **disposition** is right — default-to-proceed is what this file already wants. The **gate** was not: both stopping predicates are self-assessed by the agent that wants to proceed, and every incident in `RULES-ARCHIVE.md` looked safe at that moment. That is the preamble's own failure mode — a rule obeyed straight into the thing it forbids.

Ships as a four-branch trigger tree where **the branches do not all end in "ask"**.

## Four holes closed

| hole | why it mattered |
|---|---|
| "safe and unblocked" is true of **out-of-scope** work | writing tests nobody asked for passes that predicate |
| unblocked ≠ **unambiguous** | branch 1 licensed ship-then-rework |
| "blocked" collapsed **three** cases | measurable-yourself (go look) / only-the-user-knows (ask) / external system (asking does not unblock it) |
| dropped **outward-facing** entirely | a push, PR, email or deploy destroys nothing and is locally safe → all landed in "proceed" |

Routing every hazard to a question also invites asking permission for the never-list. Those rules exist *because* a yes was given once; their response is stage-it or hand-it-over.

## Byte budget

`RULES.md` arrived at this PR with **4 bytes** of free budget above the floor — not erosion but the gate working: the previous bump's margin was spent by an intervening rule, and the next author (me) got the *warning* rather than the surprise, which is what `MIN_HEADROOM_BYTES` exists for. ⚠ Two bases are in play throughout that module — ceiling-relative and floor-relative, differing by `MIN_HEADROOM_BYTES` — and subtracting across them is meaningless. A draft of this body did exactly that and reported a 106 B rule as costing 1,006 B. The module labels both bases now; read it there.

Paid by **consolidation** — the tree subsumes and deletes two bullets, "Flag BEFORE acting" and "User-facing micro-decisions" — plus a ceiling bump.

> ⚠️ **Figures are deliberately not restated here.** `scripts/tests/test_rules_size.py` is the single source of truth for `MAX_BYTES`, the eviction, and the net; its own docstring names PR bodies as a surface that must cross-reference rather than copy. An earlier version of this body restated five literals and all five went stale within the PR. Read them from the module.

## What the audits actually found

Audited repeatedly — one blind pass, then a delta re-audit after every fix round. ⚠ **No count is given here on purpose.** Three drafts of this line stated one and all three were wrong, in the same direction; `git log main..HEAD` is the authority, since every fix commit opens by naming the audit that prompted it. Headline finding per round. ⚠ **Not all of these are fix-induced regressions**, and an earlier version of this body claimed they were — which flattered the process. Most rounds found something the previous fix created. One did not: it found a claim wrong in the PR's **first commit** that four rounds, including the blind one, read straight past. The table below is not exhaustive — it lists the headline finding per round, not every finding. A four-round detection miss is a different (and worse) result than a fix-induced regression, and is recorded as such:

| round | finding |
|---|---|
| initial | 🔴 The clawgate tie-in claimed `approve-with-comment` was "answer plus proceed in one tap". **False** — `clawgate-hook.sh` logs the comment and returns a bare allow/deny. Asserted from the feature's *name*; contradicted by the hook's own source comment *and* by a bold line 23 rows above where it was written. |
| initial | 🟡 The consolidation was **not scope-neutral** though the module asserted it was: response imperatives lost half their coverage, a trigger *category* was swapped while its examples were kept, and a trigger *condition* vanished. |
| fix-induced | 🔴 The fix for the never-list **made it worse** — `marks 🔴 never` excluded `git add -A`, `reset --hard` and `sudo nixos-rebuild`, three of the six hazards the archive entry itself lists. Same defect class as the one being fixed. |
| fix-induced | 🔴 Three consecutive rounds stated the remaining slack; **each round's own edit invalidated the figure it wrote** (186→56, 256→177). Fixed structurally — the derived figures came out of the prose. |
| fix-induced | 🔴 The round-3 retraction landed on **one copy** of a claim; the identical sentence stood in the archive anchor the rule points at, which had begun contradicting itself. "One rule, one place", demonstrated by the PR shipping it. |
| fix-induced | 🔴 The round-5 *correction* was itself false, in the mirror direction — a blanket "an agent can never set `complete`" would have defeated the criteria gate, since the local pickup path is exactly the one that may. Scope clawgate claims to the **route**, not the verb. |
| fix-induced | 🔴 A **fourth** fix-on-one-copy — of round 6's own fix, one screen below it: a retracted figure still quoted, as a phrase that no longer existed, attributed to the wrong block. Also: the Named-hazard branch was a two-file **closed list** where the thing described is a shape. |
| detection miss | 🔴 The archive called clawgate's criteria gate **structural**. It is a convention — `AllowedForAgent` is criteria-independent and "acceptance criteria" appears in zero Go files. This repo corrected that exact overclaim **the day before** (#691). Same failure as the first 🔴: a capability read from a document, not the code. |

The ratio that matters: on the consolidation specifically, **a cold read caught one scope loss and the first blind audit caught three more** — and the author who wrote the replacement was its worst reader. **Four times** a fix landed on one copy of a duplicated claim — the third was this PR body, the fourth was a fix's own retraction one screen above the line that still quoted it. **Twice** a clawgate claim was asserted from documentation rather than source, and once the *correction* to one of those was wrong in the mirror direction.

## clawgate tie-in — transport, never gate

Domain detail lives in `claude/skills/clawgate/reference/hooks.md` (0 context until triggered). Verified against live `0.7.98`, pin matching, `PermissionRequest` on **both** hosts:

- fires on **tool permission prompts only** → cannot carry a Fork or Out-of-scope trigger
- its return channel is **binary** — permission, never an answer
- **allowlisted commands never prompt**; `bypassPermissions`/`plan`/`AskUserQuestion` defer without contacting the server
- **fails open** — no rule may be gated on it
- `requireSession` is a literal `return next` with `POST /api/auto-approve-all` behind it. ⚠ That posture is **deliberate** (Authelia at the public edge, LAN trusted-open), so the hazard is blast radius, not a missing gate.

The **provenance-and-freeze** clause was *prompted by* clawgate's task status gate, which faces the same problem. ⚠ Clawgate does **not** solve it structurally — the AUTHOR-SPECIFIED/DERIVED split and the freeze are skill prose, self-assessed by the agent, and `"acceptance criteria"` appears in zero Go files. The genuinely structural fact is narrower and **route-scoped**: a *dispatched devpod* agent cannot set `complete`; the machine route a local pickup uses deliberately can. The borrowed idea stands on its own merits; see `RULES-ARCHIVE.md` § `proactivity-gate`.

## Verification

- Clause-by-clause diff vs `origin/main`: **17/17** triggers, responses, examples and the timing clause present
- `scripts/tests`: **7,203 passed, 0 failures** (errors seen are `DEVRC-GITENV-VIOLATION` teardowns from concurrent sessions sharing this git dir, reproducible on untouched `origin/main`)
- Size gate **watched to fail** under five mutants — ceiling, headroom band in isolation (ceiling still green), both archive-pointer directions, and a hard-coded playbook list — each dying with its own test's message. A sixth, the `retired-*` orphan, is a control of the opposite polarity and must stay **green**; it does.
| fix-induced | 🔴 A route **count** wrong under every counting convention. Removed rather than corrected a third time — the bullet now states durable facts and says to enumerate from `SetStatus`'s call sites at the moment of need. |
| fix-induced | 🟡 The retracted `agent-token` label survived **ten lines above its own retraction**, inside the bullet doing the retracting. Fifth instance of this PR's signature defect. |

*Two recurring shapes. A fix landing on one copy of a duplicated claim — five times, including this PR body and, finally, a paragraph correcting itself. And a clawgate capability asserted from documentation rather than source — twice, plus one over-correction and one wrong count. Note the sweep that "confirmed no copy survives" was a grep for a **word**; three copies survived, spelled differently. That is this repo's own `spelled-guards` rule, and my sweep was the guard.*


ZacxDev:
## Correction — the PR description above contains a claim that is false

Posting this rather than silently editing the body, because a reviewer may already have read the wrong version.

**The description asserts:**

> `approve-with-comment` is exactly the tree's **Fork** branch — answer plus proceed in one tap

**That is wrong.** `hook/clawgate-hook.sh:171-174`:

```
# Output is binary (decision.behavior allow|deny). No reason/additionalContext
# channel exists, so an approver comment is recorded in the log only.
[ -n "$COMMENT" ] && log "approver comment (record-only): $COMMENT"
```

The comment never reaches the model. An agent that routes a Fork to a clawgate card receives **permission, never an answer**, and proceeds down whichever reading it had already picked — the exact ship-then-rework the Fork branch exists to prevent, while the operator believes they answered.

Three things make this worse than an ordinary error, and they are the reason it is worth a public note:

1. It was asserted from the **feature's name**, without reading the hook.
2. It was contradicted by the hook's own source comment.
3. It was contradicted by a pre-existing line **in the very file it was written into**, in bold, 23 lines above: *"Never design on feeding an approver's words back into the session through this hook."*

A blind adversarial audit caught it. Three rounds of my own review did not.

Fixed in e750a340 — the claim is inverted in both `hooks.md` and the archive, and the coverage table's Fork row now reads **No** rather than "only if it surfaces as a prompt".

### Also corrected in that commit

The description's claim that the consolidation preserved every imperative and trigger of the two deleted bullets was **also not true**. A cold read caught one loss pre-merge; the audit caught three more (response imperatives lost half their coverage; the trigger *category* was swapped while its examples were kept; the micro-decisions trigger *condition* vanished). All restored — a clause-by-clause diff against `origin/main` now shows 16/16 present.

### Numbers in the description are stale

`1,206 / 488 / 718 / 186` are superseded by `1,336 / 489 / 848 / 256`, and `MAX_BYTES` is now 40,300 rather than 40,100 — the +900 bump could not absorb this PR's own review round, which is itself recorded as the finding. Per `test_rules_size.py`'s docstring these literals should not have been restated here at all; **read them from that module**, which is the single source of truth.

ZacxDev:
## Out-of-scope finding, filed rather than folded in — clawgate task **346**

The seventh audit found three **pre-existing** files in the clawgate skill making contradictory exclusivity claims about who can set a task to `complete`:

- `reference/architecture.md:70-71` — "Only the operator/human may set `complete`."
- `flows/task-pickup.md:64-65` — "Only the machine route this ritual uses can set `complete` at all"
- `reference/task-api.md:91-92` — same

Two are mirror images, so whichever a reader loads first decides what they believe. Ground truth from source: `notes.SetStatus` is the single writer; exactly two surfaces gate on `AllowedForAgent`, both the in-devpod agent tool/route; the hook-token machine route, the operator route and its MCP tool, and the losing side of `POST /tasks/merge` all set `complete` deliberately.

None is touched by this PR. Correcting them is a clawgate-skill change, not a proactivity-rule one — so it is **filed, not silently folded in**, which is what this PR's own Out-of-scope branch prescribes ("finish what WAS asked in full, then file the rest as a task, not a report line").

⚠ Upstream: clawgate's own `internal/taskstatus/taskstatus.go:29-31` comment says *"Complete is privileged: only the operator/human may set it"* — wrong about its own package, and what these three files faithfully mirror. That is the root, and it is in `homelab-talos`, not here.

### What this PR *did* fix in the same area

Its own version of this claim was wrong in **three consecutive rounds** — "structural" when it is convention, then a blanket "an agent can never set `complete`", then a route count wrong under every convention. It now states no count at all: the durable facts, plus "enumerate from `SetStatus`'s call sites at the moment you need it". Same structural move this PR made for the byte figures, for the same reason.

PR #734, opened 2026-08-23
fix(gate): GUARD 10 splits its protected set — the repo-local config has ~90 other writers (#730)

### LINKED ISSUE #730: GUARD 10 blames a target for the SHARED clone's .git/config being rewritten by another session — the #719/#720 defect, one guard over
## Summary

#720 gave **GUARD 9** attribution machinery so it stops blaming tests for a repository with other writers. **GUARD 10** (added in #673) watches an overlapping set of files with **no attribution at all**, and now fails the suite for the same reason.

In one run, GUARD 9 **proved** co-tenancy and downgraded itself to report mode — and GUARD 10, in that same run, failed two targets for a change it could not have caused:

```
gitenv(session) mode=report(auto) unattributable=1
gitenv(session)   cannot attribute: live processes are sitting inside a protected
                  repository (cwd), and none of them is ours: 67653:zsh, 69056:.claude-wrapped, …
gitenv(session) summary: attributed-violations=0 unattributed-observations=1

  TOTAL collected=15152  passed=15150  skipped=2  failed=0
  ERROR: 2 GUARD 10 problem(s):
         scripts/tests          — CHANGED 1 of the operator's protected git config file(s) while it ran
             /home/zach/workspace/devrc/.git/config
         scripts/dl-router/tests — CHANGED 1 of the operator's protected git config file(s) while it ran
             /home/zach/workspace/devrc/.git/config
RESULT: FAIL (exit=1)
```

**Every test passed.** The run is red purely on GUARD 10.

## Control — same code, two environments

| environment | GUARD 10 | verdict |
|---|---|---|
| **Isolated clone** (own `.git`, no other writers) | `real-config-changed=0/3` on every target | `RESULT: PASS (exit=0)` |
| **Shared clone** (~90 worktrees, ~15 concurrent sessions) | `.git/config` CHANGED on 2 targets | `RESULT: FAIL (exit=1)` |

Same commit, same guard. The writer is external.

## Mechanism

`NOGIT_PROTECTED` includes the truly-global files *and* the repo-local one:

```sh
NOGIT_REPO_GITDIR="$(git -C "$ROOT" rev-parse --path-format=absolute --git-common-dir …)"
[ -n "$NOGIT_REPO_GITDIR" ] && _nogit_protect "$NOGIT_REPO_GITDIR/config"
```

That path is the **common** git dir, shared by every worktree of the clone. `git worktree add`, `git branch --track`, and any `git config` in any worktree rewrite it. With this many worktrees it changes continuously, and GUARD 10 sha256s it before/after each target.

The message is also a confident misdiagnosis — it tells you the test "reached them some other way (code that removes the variable from its own environment, or a repo-local write)", sending you to audit a target that did nothing.

## Suggested direction

Mirror #720 rather than inventing a second policy:

- Split `NOGIT_PROTECTED` into **global** (`~/.gitconfig`, `$XDG_CONFIG_HOME/git/config`, `--show-origin` files) and **repo-local** (`<git-common-dir>/config`).
- **Global stays enforcing** — no concurrent worktree operation touches those, so a change there is still attributable and is the 2026-08-21 incident's actual shape.
- **Repo-local downgrades to REPORT only when co-tenancy is PROVEN**, reusing GUARD 9's existing `attribution_evidence()` / `live_cotenants()` rather than a second heuristic. With no proven co-tenant it keeps failing, so the `core.bare = true` casualty that motivated protecting it stays covered on a clean machine and in CI.

That keeps the teeth where they bite and removes the false red where the guard provably cannot attribute.

ZacxDev:
Closes #730.

`f340d52d` splits the class, `62dd53d8` fixes what the adversarial audit found in the first, `5cd5ee4d` merges current `main` in.

### 🔴 The merge-gating tier was red, and it was not this PR

`tekton/devrc-pytests` went **FAILURE** on `62dd53d8` — `failed=1`. I had run the full suite four times in the **devShell** and never seen it: the authoritative runner is `nix build .#checks.x86_64-linux.pytests`, whose source is a `/nix/store` copy with **no `.git`**, and my runs were structurally blind to that tier.

Reproduced locally before theorising, and the failing test was **`test_the_module_root_is_load_bearing`** in `scripts/tests/test_git_repo_isolation.py` — a GUARD 9 test this branch never touches (`git diff` over both my commits: empty for that file and for `gitenv.py`). Its assertion is `gitenv.py is not inside a git checkout`, which is a property of the sandbox, not of my change.

The control that settles it: `5a2a7b21` on `main`, *"the module-root guard asserted THIS checkout has a `.git` — the authoritative runner has none (#732)"*, landed **after** my base `a0fb39c7` and says the failure fired *"on every devrc-ci PipelineRun, on unrelated branches, for a reason no PR had touched."* Inherited, already fixed upstream.

Fixed here by **merging `origin/main`** (not rebasing — the audit references `f340d52d`/`62dd53d8` by SHA, and this needs no force-push). `main` also touched `scripts/run-tests.sh`, so I read the merged result rather than trusting a clean merge: its change is a `TARGET_FLOORS` re-pin, a different region from GUARD 10 entirely — no semantic conflict.

**The merged tree passes the authoritative gate**, run locally:

```
  TOTAL collected=15198  passed=15196  skipped=2  failed=0  (floor: 13732 = sum of 27 per-target floors)
    classes=global:2 (always enforcing)  repo-local:0 (enforcing unless another writer is PROVEN — see #730)
    repo-local-reported-total=0
RESULT: PASS (exit=0)
```

🔴 Note `repo-local:0` — that is **audit finding 4 confirmed live**: the sandbox has no `.git`, so this class is empty there and has nothing to enforce. The corrected comment says exactly that; the number now proves it.

## What was wrong

#720 gave **GUARD 9** attribution machinery so it stops blaming tests for a repository with other writers. **GUARD 10** watched an overlapping file set with **no attribution at all**, and failed the suite for the same reason — in the same run in which GUARD 9 had already *proven* co-tenancy and downgraded itself. Every test passed; the run was red purely on GUARD 10.

`NOGIT_PROTECTED` mixed two things that are not the same claim:

| class | files | who else writes them |
|---|---|---|
| **GLOBAL** | `~/.gitconfig`, `$XDG_CONFIG_HOME/git/config`, everything `git config --global --list --show-origin` names | nobody — a change is still attributable |
| **REPO-LOCAL** | `<git-common-dir>/config` | **every worktree of the clone**: `git worktree add`, `git branch --track`, any `git config` |

Its message was also a confident misdiagnosis, sending the reader to audit a target that had done nothing.

## What changed

* **GLOBAL stays fully enforcing**, including when co-tenancy *is* proven. That is the 2026-08-21 incident's actual shape.
* **REPO-LOCAL is enforcing unless another writer is PROVEN.** With proof the change is **REPORTED**: counted per target (`repo-local-reported=N`), counted per run (`repo-local-reported-total=N`), and rendered with the file and the evidence sentence. Without proof it still **fails**.
* **The proof is GUARD 9's own already-audited evidence**, called via `testlib.gitenv.attribution_evidence()` / `protected_git_dirs()` — one rule, one place. `$ROOT` is passed explicitly, because `protected_git_dirs()` also probes the module's own directory, which is the wrong repository to ask about here.
* **A broken probe fails toward enforcing**, and now that includes a *hang* (the probe is wrapped in `timeout`).
* 🔴 **Evidence is authenticated.** The probe stamps each reason `DEVRC-NOGIT-EVIDENCE\t<reason>`; only stamped lines count, and `proven` needs ≥1 accepted reason. See the audit section — this replaced a measured fail-open.

## Two-environment verification

Both run the FULL suite (`--set all`) at `62dd53d8`. They answer different questions.

### 1. Isolated clone — *does enforcement survive?*

```
    present /home/zach/.config/git/config  [global, always enforcing]
    absent  /home/zach/.gitconfig          [global, always enforcing]
    present /tmp/g10-iso2/.git/config      [repo-local, enforcing unless another writer is PROVEN]
  TOTAL collected=15159  passed=15157  skipped=2  failed=0  (floor: 13697 = sum of 27 per-target floors)
    classes=global:2 (always enforcing)  repo-local:1 (enforcing unless another writer is PROVEN — see #730)
    repo-local-reported-total=0
RESULT: PASS (exit=0)
```

**34 of 34** accounting lines `real-config-changed=0/3`; `repo-local-reported=1..9` appears **0** times; GUARD 10 problems **0**. Enforcement intact, nothing downgraded.

### 2. Shared clone — *is the false red gone?*

🔴 **This run exercised the downgrade organically**, which the earlier round could not (its window happened to contain no external write, and the PR said so). Two targets saw the shared config move underneath them:

```
    present /home/zach/workspace/devrc/.git/config  [repo-local, enforcing unless another writer is PROVEN]
    scripts/signal/tests       real-config-changed=0/3 … repo-local-reported=1
    scripts/initiatives/tests  real-config-changed=0/3 … repo-local-reported=1
    repo-local-reported-total=2
    ---- repo-local git config: REPORTED, not enforced (#730) ----
      scripts/signal/tests
        changed: /home/zach/workspace/devrc/.git/config
      scripts/signal/tests
        cannot attribute: live processes are sitting inside a protected repository (cwd),
        and none of them is ours: 25746:zsh, 25748:tmux: client, 67653:zsh, …
  TOTAL collected=15159  passed=15157  skipped=2  failed=0
RESULT: PASS (exit=0)
```

That is #730's scenario exactly — **at `a0fb39c7` this is `FAIL (exit=1)` with `failed=0`** — and it is now green, with the change visible, counted and explained rather than dropped. GUARD 9 printed `mode=report(auto)` on all 53 of its lines, as expected (#720 working).

### 3. Three-arm rig, re-derived at `62dd53d8`

The three-arm matrix in the PR comment was measured at `f340d52d`, i.e. before the evidence token changed how `proven` is decided; citing it for this head would be a claim about superseded code. Re-run in a throwaway clone (`/tmp/g10-e2e2`) with a background writer stamping the clone's **own** `.git/config` every 0.2s:

| arm | ref | co-tenancy | GUARD 10 | verdict |
|---|---|---|---|---|
| base | `a0fb39c7` | proven (`326406:sleep`) | `real-config-changed=1/3`, 1 problem block | **FAIL (exit=1)** |
| head | `62dd53d8` | proven (`327951:sleep`) | `real-config-changed=0/3`, `repo-local-reported-total=1` | **PASS (exit=0)** |
| **neg-control** | `62dd53d8` | **none** (`[]`) | `real-config-changed=1/3`, `repo-local-reported=0`, 2 problem blocks | **FAIL (exit=1)** |

Every arm carries the rig's own positive control — the config sha is asserted to have **moved** (`f819e2ac -> e1cf638c` / `-> 025bb468` / `-> 4d7f9fd8`), because a byte-identical rewrite moves no fingerprint and would score a vacuous PASS.

## Red-at-base matrix

Measured by checking out `a0fb39c7` into a worktree, copying **only** the new test file in, and running there.

| test | at `a0fb39c7` | at HEAD | label |
|---|---|---|---|
| `test_a_repo_local_change_is_REPORTED_when_a_cotenant_is_PROVEN` | **RED — behavioural** | GREEN | 🔴 **regression coverage** |
| `test_a_repo_local_change_still_FAILS_when_no_cotenant_is_proven` | RED — wording only | GREEN | negative control, *not* regression coverage |
| `test_a_global_change_FAILS_even_with_a_cotenant_PROVEN` | RED — wording only | GREEN | negative control, *not* regression coverage |
| `test_a_broken_evidence_probe_fails_TOWARD_enforcing` | RED — structural (anchor absent) | GREEN | fail-safe guard; reachability via M4 |
| `test_a_stray_print_on_the_probes_pythonpath_is_NOT_evidence` | n/a (code absent) | GREEN | 🔴 regression for the audit's fail-open; reachability via M10/M11 |
| `test_whitespace_only_probe_stdout_is_NOT_evidence` | n/a | GREEN | same site, second half |
| `test_a_LINKED_WORKTREE_protects_the_COMMON_config` | n/a | GREEN | the production shape |

Only the first is red for a *behavioural* reason, and it is the only one counted as regression coverage for the split. Measured at base: `real-config-changed=1/3`, `RESULT: FAIL (exit=1)`, `failed=0`. The two controls are red at base only over the new class wording; reading that as regression coverage would be a claim nobody measured.

## Mutation table

12 mutants against `scripts/run-tests.sh`, scored over the whole GUARD 10 suite. **Every mutant proven to have LANDED** — a unique marker grepped out of the *written* file and printed (M12 is a deletion, proven by 0 remaining occurrences) — before scoring; pristine sha256 re-asserted after each restore; each mutant's **killing assertion text captured**, not assumed.

**Control, unmutated: 26 passed, 0 failed.**

| # | mutation | landed | verdict | killing assertion |
|---|---|---|---|---|
| M1 | every file is repo-local (**class SWAP**) | `return 0  # MUTANT_M1` | KILLED ×2 | *a GLOBAL config change was downgraded by co-tenancy* |
| M2 | no file is repo-local (**guard deleted**) | `return 1  # MUTANT_M2` | KILLED ×6 | *…still failed the run — this is #730* |
| M3 | evidence probe **always-true** | `NOGIT_EV_STATUS="proven"; return 0` | KILLED ×5 | *an ATTRIBUTABLE repo-local config change produced a PASSING run* |
| M4 | a **failed** probe reads as proven | `NOGIT_EV_STATUS="proven"  # MUTANT_M4` | KILLED ×1 | *a BROKEN evidence probe downgraded a repo-local change* |
| M5 | downgrade condition **inverted** | `[ "$NOGIT_EV_STATUS" != "proven" ] …` | KILLED ×6 | *…this is #730* / *…produced a PASSING run* |
| M6 | downgrade not **counted** | `cls="repo-local-reported"  # MUTANT_M6` | KILLED ×2 | *not COUNTED as exactly 1 on the target's line* |
| M7 | the **reason** is never logged | `:  # MUTANT_M7` | KILLED ×1 | *the downgrade block did not state the REASON* |
| M8 | the downgrade **block** is never printed | `if false; then  # MUTANT_M8` | KILLED ×1 | *the downgrade block was never printed* |
| **M9** | the failure **suppresses the path** *(audit 1)* | `# MUTANT_M9` | KILLED ×3 | *the failure did not NAME the file that changed* |
| **M10** | `proven` by **presence**, not token *(audit 2a)* | `if [ -s "$out" ]; then  # MUTANT_M10` | KILLED ×2 | *a stray print() … was accepted as co-tenancy evidence* |
| **M11** | any **unstamped** line accepted *(audit 2b)* | `*) : ;;  # MUTANT_M11` | KILLED ×2 | *…accepted as evidence* / *whitespace-only probe stdout…* |
| **M12** | message drops the **SIBLING caveat** *(audit 3)* | DELETED (0 occurrences remain) | KILLED ×1 | *the failure claimed more than the probe measured* |

M9, M10 and M11 are the audit's own mutants; **each SURVIVED before `62dd53d8`**.

*Honest note on attribution:* under M2 and M3 the two probe tests die on their own **fixture control** (*"the sitecustomize never ran inside the probe"*) — the probe is never reached, so they correctly report they measured nothing rather than passing vacuously. Their intended mutants are M10/M11, which they kill by the intended assertion.

## What the audit found, and what it changed

1. 🔴 **`proven` was decided by output PRESENCE**, and the probe inherits the ambient `PYTHONPATH` — so **any** stdout counted as co-tenancy evidence. Measured with **no co-tenant**: a single `print()` gave `repo-local-reported=1` and downgraded a genuinely attributable write, rendering the stray line as GUARD 9 evidence. Fixed with the stamped token + an accepted-reason count. Second half of the same site: `NOGIT_EV_STATUS` was set *before* the blank-line-skipping loop, so whitespace-only stdout gave `proven` with zero logged reasons.
2. **Both negative controls' "did the failure NAME the file?" assertions were vacuous** — the startup listing prints every protected path unconditionally, so an audit mutant suppressing the path in the failure left them at `9 passed`. The `ERROR:` block now has an end marker and `_guard10_failure_block()` slices it; applied to all three name-the-file assertions including the pre-existing one.
3. **The failure message overstated the probe.** `live_cotenants` roots at the git dir and its parent, so a **sibling worktree** — the writer this PR exists to excuse — is invisible. Measured live: all 35 co-tenant hits were cwd=`…/devrc` itself, zero from any sibling. Roots deliberately **not** widened; the message and header now state what was checked and name the blind spot.
4. **"and in CI" was false.** The nix tier builds from a store copy with no `.git`, so `NOGIT_REPO_LOCAL` is empty there. Repo-local enforcement survives only in an isolated single-writer clone, which nothing schedules. Stated plainly. GLOBAL is unaffected and enforced everywhere.

Plus: probe `timeout`; `NOGIT_EVIDENCE_LOG` added to the creatability preflight; `"GUARD 10 problem"` asserted in the fail-safe test; trailing boundaries pinned on `repo-local-reported=1` / `-total=1` (they prefix-matched `=10`…`=19`).

## Accepted, deliberately not changed

* 🟢5 **`real-config-changed=` field name** — it now counts only *enforced* changes. Documented at the site, and the paired `repo-local-reported=` is pinned on the same line, so a reader gets both numbers together. Renaming a long-standing field is a bigger blast radius than the confusion it removes.
* 🟢6 **The two-list rendering is unpaired** (changed files and reasons are printed as separate awk passes). Cosmetic; both lists are keyed by target and a reader joins them by name.
* 🟢11 **Linked-worktree shape** — the auditor confirmed both resolvers agree live. Turned out cheap after all, so it is now covered by `test_a_LINKED_WORKTREE_protects_the_COMMON_config` rather than accepted.
* 🟢12 **The negative control depends on an environmental absence** (no process may sit in the scratch root). It fails safe: a stray co-tenant makes the control fail loudly, never pass vacuously, and it asserts `not live_cotenants(...)` up front so the premise is checked rather than assumed.

## Test method

The runner is driven against a **scratch root** — top-level symlinks to this repo so its preconditions hold, but a **fresh `.git` of its own** — so planted writes are measured against a directory pytest created and the operator's clone is never written. Co-tenancy is flipped with **real evidence**: a live process whose cwd sits in that root, with `_cotenant.__enter__` asserting `live_cotenants` actually sees it before the run whose verdict depends on it.

The stray-print fixture is a real `sitecustomize` on a real `PYTHONPATH`, **scoped to the probe by `sys.argv`** — printing from every python would corrupt GUARD 8's captured `SPOOL_TRAP_LOG` and abort the run, i.e. red for a *neighbour's* reason — plus a witness file proving it loaded inside the probe.

## Not verified / limits

* The evidence inherits `live_cotenants`' roots (a git dir and its parent), so a session in a **sibling worktree** is not counted. Documented in code, named in the failure message, and deliberately not widened. Absence of evidence leaves the guard **enforcing**.
* The probe is asked per target and only when a repo-local delta was observed, so a run with no delta never calls it; its failure path is exercised by the test and the mutation, not by a normal run.
* **Repo-local enforcement does not run anywhere scheduled** — see audit finding 4. It is a floor for a single-writer machine, not a merge gate. GLOBAL is enforced everywhere.
* Every number here is from local runs named above, plus the Tekton gate on the PR itself.
* 🔴 **Four full devShell runs did not see a red merge-gating tier.** The two-tier blindness is called out above; the merged tree is now verified in *both* (devShell full suite, and `nix build .#checks…pytests`). Treat any future "the suite is green" claim on this repo as tier-scoped until both are named.

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -2078,8 +2078,29 @@ _nogit_protect "${XDG_CONFIG_HOME:-${HOME:-}/.config}/git/config"
 # the change is REPORTED (printed, counted, named, with the reason) and does not
 # fail. The proof is GUARD 9's own already-audited evidence — see
 # `_nogit_cotenancy_probe` below — not a second heuristic. With no proven
-# co-tenant it still fails, so the `core.bare = true` casualty stays covered on
-# a clean machine and in CI, which is where that damage happened.
+# co-tenant it still fails.
+#
+# 🔴 BE PRECISE ABOUT WHERE THAT REMAINING ENFORCEMENT ACTUALLY LIVES — an
+# earlier wording here claimed the `core.bare = true` casualty "stays covered on
+# a clean machine and in CI", and the CI half is FALSE. The nix tier builds from
+# `cp -r ${./.} src` (`flake.nix`), a store copy with NO `.git` — which is what
+# the parenthesis above already says — so `NOGIT_REPO_LOCAL` is EMPTY there and
+# there is nothing for this class to enforce, downgrade or not. And on the
+# operator's box co-tenancy is provable essentially always. So repo-local
+# enforcement really survives in exactly one place: an isolated, single-writer
+# clone — which nothing currently schedules. It is a floor for the machine that
+# has one writer, not a gate that runs on every merge. The GLOBAL class above is
+# unaffected and is enforced everywhere, including in the sandbox.
+#
+# 🔴 AND THE EVIDENCE HAS A BLIND SPOT WORTH NAMING, because it is the writer
+# this change exists to excuse. `live_cotenants` roots at a git dir and its
+# PARENT, so for a linked worktree the roots are `<common>/.git` and the MAIN
+# clone's work tree. A session sitting in a SIBLING worktree
+# (`…/devrc-<topic>/`) is outside both and is NOT counted — MEASURED, all 35
+# co-tenant hits on the operator's box were cwd=`…/devrc` itself, zero from any
+# sibling. Widening those roots would be the second heuristic #730 says not to
+# invent, so this is a documented limit, not a TODO: absence of evidence leaves
+# the guard ENFORCING, which is the safe direction.
 NOGIT_REPO_LOCAL=()
 NOGIT_REPO_GITDIR="$(git -C "$ROOT" rev-parse --path-format=absolute --git-common-dir 2>/dev/null || true)"
 if [ -n "$NOGIT_REPO_GITDIR" ]; then
@@ -2113,7 +2134,10 @@ NOGIT_CONFIG="$NOGIT_DIR/gitconfig"
 NOGIT_SESSIONS_LOG="$NOGIT_DIR/sessions.log"
 NOGIT_CHANGED_LOG="$NOGIT_DIR/changed.log"
 NOGIT_EVIDENCE_LOG="$NOGIT_DIR/evidence.log"
-if ! : > "$NOGIT_CONFIG" || ! : > "$NOGIT_CHANGED_LOG"; then
+# 🔴 THE EVIDENCE LOG IS IN THIS CHECK, not just in the list above. It is the
+# only record of WHY anything was downgraded, there is no `set -e` here, and an
+# unwritable one would silently produce downgrades with no stated cause.
+if ! : > "$NOGIT_CONFIG" || ! : > "$NOGIT_CHANGED_LOG" || ! : > "$NOGIT_EVIDENCE_LOG"; then
   echo "run-tests: FATAL — could not create the git-isolation files under" >&2
   echo "  $NOGIT_DIR. Refusing to run: without them a test that calls" >&2
   echo "  'git config --global' rewrites the operator's ~/.gitconfig, and a" >&2
@@ -2234,42 +2258,80 @@ fi
 # else. (It is the same argument the runner already `cd`'d to.)
 #
 # 🔴 FAIL TOWARD ENFORCING. A probe that cannot run (no python, an import error,
-# a helper that raises, a repo it cannot resolve) must NEVER read as "another
-# writer was found" — that is a broken instrument silently switching the guard
-# off, which claude/RULES.md rates worse than no guard. Any non-zero exit, and
-# any empty git-dir set, comes back as `probe-failed` and the change is enforced
-# exactly as it was before this change existed.
+# a helper that raises, a repo it cannot resolve, a hang) must NEVER read as
+# "another writer was found" — that is a broken instrument silently switching
+# the guard off, which claude/RULES.md rates worse than no guard. Any non-zero
+# exit, any empty git-dir set, and any output that does not carry the token
+# below comes back as `probe-failed`/`none`, and the change is enforced exactly
+# as it was before this change existed.
+#
+# 🔴 THE TOKEN IS THE FIX FOR A MEASURED FAIL-OPEN. This used to decide `proven`
+# from `[ -s "$out" ]` — output PRESENCE, not content — while the probe inherits
+# the ambient `PYTHONPATH`. So ANY stdout on that path counted as evidence:
+# MEASURED with NO co-tenant present, a single `print()` added to
+# `scripts/testlib/__init__.py` produced `repo-local-reported=1` and downgraded
+# a genuinely attributable write, rendering the stray line as GUARD 9 evidence.
+# A package that chatters at import, a `.pth`, a `sitecustomize` — all reach it.
+# Now only lines the probe itself stamps are accepted, and `proven` requires at
+# least one ACCEPTED reason, so whitespace-only or unstamped stdout is `none`.
+NOGIT_EVIDENCE_TOKEN="DEVRC-NOGIT-EVIDENCE"
+# `timeout` keeps a hung probe from wedging the whole run — "fails toward
+# enforcing" says nothing about a `/proc` stat that never returns. Absent
+# `timeout` is not fatal: the probe runs bare, exactly as before.
+if command -v timeout >/dev/null 2>&1; then
+  NOGIT_PROBE_TIMEOUT=(timeout 60)
+else
+  NOGIT_PROBE_TIMEOUT=()
+fi
 NOGIT_EV_STATUS=""    # proven | none | probe-failed
 _nogit_cotenancy_probe() { # $1 = target -> sets NOGIT_EV_STATUS, logs the reason
-  local t="$1" out="$NOGIT_DIR/evidence.out" err="$NOGIT_DIR/evidence.err" rc=0
+  local t="$1" out="$NOGIT_DIR/evidence.out" err="$NOGIT_DIR/evidence.err"
+  local rc=0 n=0 line reason
   : > "$out"; : > "$err"
-  PYTHONPATH="$ROOT/scripts${PYTHONPATH:+:$PYTHONPATH}" python -c '
+  PYTHONPATH="$ROOT/scripts${PYTHONPATH:+:$PYTHONPATH}" \
+  ${NOGIT_PROBE_TIMEOUT[@]+"${NOGIT_PROBE_TIMEOUT[@]}"} python -c '
 import sys
 from pathlib import Path
 from testlib.gitenv import attribution_evidence, protected_git_dirs
-dirs = protected_git_dirs([Path(p) for p in sys.argv[1:]])
+token, starts = sys.argv[1], sys.argv[2:]
+dirs = protected_git_dirs([Path(p) for p in starts])
 if not dirs:
     # No repository resolved, yet a repo-local config just changed. The two
     # cannot both be true, so this is a broken probe, not an absence of writers.
     sys.exit(3)
 for reason in attribution_evidence(dirs):
-    print(reason)
-' "$ROOT" >"$out" 2>"$err" || rc=$?
+    reason = " ".join(str(reason).split())   # one line, so one accepted reason
+    if reason:
+        sys.stdout.write(f"{token}\t{reason}\n")
+' "$NOGIT_EVIDENCE_TOKEN" "$ROOT" >"$out" 2>"$err" || rc=$?
   if [ "$rc" -ne 0 ]; then
     NOGIT_EV_STATUS="probe-failed"
     printf '%s\t%s\t%s\n' "$t" "probe-failed" \
-      "the co-tenancy probe exited $rc, so no writer was PROVEN and this change is ENFORCED: $(tr '\n' ' ' < "$err" | cut -c1-300)" \
+      "the co-tenancy probe exited $rc (124 = timed out), so no writer was PROVEN and this change is ENFORCED: $(tr '\n' ' ' < "$err" | cut -c1-300)" \
       >> "$NOGIT_EVIDENCE_LOG"
-  elif [ -s "$out" ]; then
+    return 0
+  fi
+  # Only the probe's OWN stamped lines are evidence. Anything else on stdout
+  # belongs to somebody else's import and is not a fact about co-tenancy.
+  while IFS= read -r line; do
+    case "$line" in
+      "$NOGIT_EVIDENCE_TOKEN"$'\t'*) : ;;
+      *) continue ;;
+    esac
+    reason="${line#"$NOGIT_EVIDENCE_TOKEN"$'\t'}"
+    [ -n "$reason" ] || continue
+    n=$(( n + 1 ))
+    printf '%s\t%s\t%s\n' "$t" "proven" "$reason" >> "$NOGIT_EVIDENCE_LOG"
+  done < "$out"
+  # 🔴 SET *AFTER* THE LOOP, deliberately. Setting it before meant
+  # whitespace-only stdout produced `proven` with ZERO logged reasons — a
+  # downgrade whose stated cause was the empty string.
+  if [ "$n" -gt 0 ]; then
     NOGIT_EV_STATUS="proven"
-    while IFS= read -r line; do
-      [ -n "$line" ] || continue
-      printf '%s\t%s\t%s\n' "$t" "proven" "$line" >> "$NOGIT_EVIDENCE_LOG"
-    done < "$out"
   else
     NOGIT_EV_STATUS="none"
     printf '%s\t%s\t%s\n' "$t" "none" \
-      "no live process outside this run's own lineage is sitting inside the protected repository, so this change IS attributable to the target and is ENFORCED" \
+      "the co-tenancy probe returned no evidence, so this change IS attributable to the target and is ENFORCED" \
       >> "$NOGIT_EVIDENCE_LOG"
   fi
 }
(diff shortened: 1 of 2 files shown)


ZacxDev:
## The end-to-end gap is now closed — measured, three arms

The PR body honestly records that the shared-clone run had `repo-local-reported-total=0`, so *"a real external write during a real run → reported → still green"* was never observed in one continuous run. I reproduced it in a **throwaway clone** (`/tmp/g10-e2e`), so the operator's real `.git/config` was never touched.

**Rig:** a clone with (a) a genuine co-tenant — a live process whose cwd is inside the work tree, confirmed by the repo's own probe, not by my guess:

```
live_cotenants      : ['3486186:sleep', '3496176:python3']
attribution_evidence: ['live processes are sitting inside a protected repository (cwd),
                       and none of them is ours: 3486186:sleep, 3496176:python3']
```

and (b) a background writer rewriting that clone's **own** `.git/config` every 0.2s, so a delta is guaranteed to land inside a target's before/after window.

| arm | code | co-tenancy | GUARD 10 | verdict |
|---|---|---|---|---|
| **base** | `a0fb39c7` | proven | `real-config-changed=1/3`, 7 problems | **FAIL (exit=1)** — #730 reproduced |
| **head** | `f340d52d` | proven | `real-config-changed=0/3`, `repo-local-reported-total=5` | **PASS (exit=0)** |
| **neg-control** | `f340d52d` | **removed** | `real-config-changed=1/3`, `repo-local-reported=0`, 7 problems | **FAIL (exit=1)** |

The third arm is the one that matters: with the same code and the same external writes, but **no provable co-tenant**, the guard still fails. So this is gated on evidence, not switched off.

### The rig needed a positive control, and it caught my own inert run

My first head-arm run returned a clean `PASS` with `real-config-changed=0/3` **and** `repo-local-reported-total=0` — no delta detected at all. Cause: the writer reused the same key/value pairs it had written in the base arm, so it rewrote **byte-identical** content, the sha256 never moved, and there was nothing to detect. A green that proved nothing.

Fixed by stamping every write uniquely per arm and asserting the file actually moved, which every arm above now reports:

```
RIG-OK: .git/config moved d2f7f9491270f2e7 -> 9597da59e846f1ba   (head)
RIG-OK: .git/config moved 9597da59e846f1ba -> 89f493a9c70a536c   (neg-control)
```

Without that control the head arm's PASS would have been recorded as the end-to-end proof it was not.

*(Arms used a reduced target list via the repo's own `runner_patch.runner_with_targets`, so each is ~2 min rather than ~15. GUARD 10's accounting is per-target, so the reduction does not change what it measures.)*

ZacxDev:
## Independently re-verified at the FINAL tree (`5cd5ee4d`)

A review fix resets the verification gate, so I re-ran the decisive behaviours myself rather than merging on the fix round's own report. Four arms, each asserting the config **actually moved** (`RIG-OK`), in a throwaway clone — the operator's real `.git/config` untouched:

| arm | co-tenancy | stray probe output | GUARD 10 | verdict |
|---|---|---|---|---|
| base `a0fb39c7` | proven | — | `real-config-changed=1/3` | **FAIL** — #730 reproduced |
| negative control | none | — | `repo-local-reported-total=0`, 7 problems | **FAIL** — enforcement intact |
| **stray-print (🟡2)** | none | unstamped line on probe stdout | `repo-local-reported-total=0`, 8 problems | **FAIL** — downgrade correctly refused |
| positive | proven | — | `repo-local-reported-total=8` | **PASS** |

The third row is the one worth calling out. I reproduced 🟡2's exact threat independently — an unrelated package printing at import time, scoped to the probe invocation, confirmed to reach the probe's stdout:

```
$ …python -c '<probe source>' /tmp/g10-e2e
stray line from an unrelated package at import time
```

Under the old `elif [ -s "$out" ]` that non-empty stdout would have set `proven` and downgraded a genuinely attributable write. With the stamped sentinel it does not: `repo-local-reported-total=0`, run **FAILS**. The only measured route to a downgrade-without-proof is closed.

Also spot-checked rather than assumed: the `origin/main` merge's 13 lines in `run-tests.sh` are a `TARGET_FLOORS` re-pin for an unrelated target, nowhere near GUARD 10 — so the 12-mutant battery from `62dd53d8` still applies to the merged tree, which was the fix round's one acknowledged gap.

Two of my own greps were the imprecise instrument here, not the code: `"clean machine and in CI"` still matches because it now survives only inside its own **retraction**, and the evidence sentinel appears once because it is a variable (`NOGIT_EVIDENCE_TOKEN`) used at both sites — one rule, one place. Both are correct as written.

Merging.

PR #739, opened 2026-08-23
feat(check-clickup-addressed): port upstream rounds 7+8 into the copy that actually runs

ZacxDev:
Ports two rounds of upstream work on `check-clickup-addressed` that landed in the private
client repo **after** this skill was migrated here at 14:03 today, so the copy that actually
runs (`~/.claude/skills/check-clickup-addressed/` → `$DEVRC/scripts/check-clickup-addressed/`)
has been missing them all day.

Upstream #1238 was already carried across at 14:47 and is **not** re-ported — confirmed by
reading `recent-comments.py` (`UNIDENTIFIED`, `latest_reply_by`, `my_latest_reply`) before
starting.

🔴 **This is a PORT, not a copy.** The upstream hunks still carry real ClickUp ids, real
colleagues' names and verbatim third-party comment bodies; this repo is PUBLIC and was
scrubbed this morning (`ec4fc008`). Everything was re-synthesised as it was ported and every
value is on the existing ledger in `tests/test_no_real_identifiers.py` — **no new ledger
entries were needed.**

## Round 7 (upstream #1246, merged)

- an unreadable date on **one of YOUR OWN** comments returned an older reply, restoring the
  exact false positive #1238 fixed → `latest_reply_ts_by` returns the `UNIDENTIFIED` sentinel
- the answered/not-answered comparison ran at **minute** resolution, so a reply written 20
  seconds *before* the question counted as answering it → raw `date_ms` /
  `my_latest_reply_ms`, used only when **both** sides have them, with a minute-age fallback
- a **suppressed** flag printed nothing at all → its own `## Answered already` block, carrying
  the ClickUp-has-no-bot-identity caveat once at the top
- `report_blocks()` extracted from `main()` — and then actually driven THROUGH `main()`,
  because deleting the whole print loop left the suite green
- `tests/test_waiting_corpus.py`: 21 labelled **records** (the existing corpus scores comment
  TEXT; this flag reads FIELDS, so all 49 of its cases are blind to it)

## Round 8 (upstream #1253, open — content ported, PR to be closed unmerged)

- `tests/test_bounds_and_parsing.py`: three invariants the prose guaranteed and nothing
  pinned — per-record `disagreements([r])` scoping, the display-vs-bound age split, and
  `UNANSWERED_COMMENT_DAYS` pinned **on** its boundary (14.0) and at a non-multiple overshoot
  (20). The old fixtures sat at 13 and 30, and 30 is more than 2×14, so a doubling mutant
  walked straight through the gap.
- `TypeError` added to `_age_days_from_ms`' except tuple (a string `date_ms` crashed the whole
  report)
- three comment corrections, including `_age_days_from_ms` blaming `OSError` when the
  exception is `ValueError` and the real mechanism is that `_epoch_ms` never range-checks
- `tests/mutation_sweep.py` — the battery as committed data plus a runner

### 🔴 The scan-less announcement was NOT ported — it is unreachable here

Upstream #1253's headline finding announces that two rules could not fire because a parallel
upstream change made the transcript scan **opt-in**. Measured here before deciding:
`check-addressed.py` has no `--transcripts` flag (`parse_args` rejects unknown args and
exits 2), `main()` always runs `search-sessions.py` and `check-completion.py`, and
`check-completion.py` writes `mentions_found` on all three of its exit paths. Nothing can
produce a `not_scanned` status or a record without a mention count, so the announcement would
guard a state no input can reach. Also skipped for the same reason: the `"mentions_found" in
r` gate on the open-signals flag, the SKILL.md line correction (this tool's default *does*
run the decision flags), and mutants N1–N5. Every omission is recorded at the site that would
carry it, with the restore condition named: **a `--transcripts` opt-in landing here.**

## 🔴 The mutation battery caught its own harness first

The first full run reported all 59 mutants KILLED — and every killer list opened with
`test_the_scan_actually_walks_files_and_can_see_an_id`. `test_no_real_identifiers.py`
resolves `<repo>/scripts` and `<repo>/claude` from `__file__` and asserts loudly when a scan
root is missing (correctly), which it always is inside the sweep's temp copy. All four of its
tests failed in **every** run, so a mutant changing nothing would also have scored KILLED.
A fully green sweep proving nothing — the exact "dying to a different guard's assertion is a
green you cannot spend" failure, hit by the file that warns about it.

Fixed two ways, both in `mutation_sweep.py`: the hygiene gate is dropped from each mutant copy
(no code mutant can ever be caught by a tree-walking check), and a **NULL CONTROL** now runs
first — it copies the skill, mutates nothing, and the run **aborts** unless it reports
SURVIVED.

## Verification

| tree | result |
|---|---|
| pre-port scripts + pre-port tests | **176 passed, 0 failed** |
| pre-port scripts + **HEAD tests** | **178 passed, 31 failed** |
| HEAD scripts + HEAD tests | **209 passed, 0 failed** |

(`test_no_real_identifiers.py` excluded from all three — it cannot resolve the repo from a
temp copy, so it would fail identically in every row. In-repo, with it included: **180 → 213
collected.**)

Waiting-flag corpus: **10/21 pre-port → 21/21**.

Mutation battery: **59 mutants + a positive control + a null control, non-KILLED: 0.**

⚠️ Four of the 31 reds are round-8 guards going red with `AttributeError` because
`suppressed_notes` / `_age_days_from_ms` do not exist pre-port — that is the code being
ABSENT, not wrong. They are **mutation-coverage guards, not regression coverage**, and the
file says so at the top. Counting them would overstate the matrix by four.

`_selfrun.py` was not touched, but its markers are path fragments and the two layouts spell
the same segments in reverse order, so all eight marker tests were re-run **by name** and are
green: `test_skill_catalog_mention_is_not_a_self_run`,
`test_the_devrc_invocation_path_is_a_self_run`,
`test_the_invocation_SKILL_MD_ACTUALLY_TEACHES_is_a_self_run`,
`test_a_json_run_carries_the_self_run_marker`,
`test_a_mention_of_the_scripts_DIRECTORY_is_not_a_self_run`,
`test_each_self_run_marker_fires_on_its_own`,
`test_self_run_marker_set_has_not_drifted`, `test_skill_tool_invocation_is_a_self_run`.

**No live ClickUp run**, in any round of this work upstream or here. Every claim above is
measured against the suite, the corpus and the sweep.

## Found and deliberately not fixed

- `_collect` still sorts on the FORMATTED `date` string. Lexicographically correct for
  `%Y-%m-%d %H:%M`, and `date_ms` now exists so the sort could read it — but changing it
  changes which comments the report samples, and nothing measured says the current order is
  wrong.
- The **threaded-comment blind spot** is unchanged: `recent-comments.py` fetches top-level
  comments only, so an answer written inside a comment thread is still invisible to
  `my_latest_reply`.

### DIFF
--- a/claude/skills/check-clickup-addressed/SKILL.md
+++ b/claude/skills/check-clickup-addressed/SKILL.md
@@ -47,9 +47,14 @@ is run straight out of the working tree — the edit is live immediately, and it
    status/priority, author, snippet + full `text`, and `my_latest_reply` — the date of *your*
    newest comment on that task, computed **before** your comments are dropped, because the
    consumer otherwise cannot tell an answered ticket from an abandoned one. That last key
-   is **omitted entirely** when the user id could not be resolved — absent means "the
-   check could not run", which is not the same fact as a present `null` ("I looked; you
-   never replied"). Both consumers branch on which.
+   is **omitted entirely** when the user id could not be resolved, or when a date on one of
+   *your* comments would not parse — absent means "the check could not run", which is not the
+   same fact as a present `null` ("I looked; you never replied"). Both consumers branch on
+   which. Beside the two display dates it also emits the raw epoch-ms they were formatted
+   from (`date_ms`, `my_latest_reply_ms`), because `format_date` throws the seconds away and
+   the "have I answered?" comparison then runs at minute resolution. Each is emitted **only
+   when it parses** — absent, never `0`, and `my_latest_reply_ms` is nested under
+   `my_latest_reply` so a precise instant can never outlive the reply it refines.
 2. **`search-sessions.py`** — searches `~/.claude/projects/**/*.jsonl` transcripts for
    sessions mentioning a task ID, ranks by hit count. Terms are **ANDed**, so it is run
    **once per task**, never with several tasks' terms in one query.
@@ -81,7 +86,7 @@ process per task, so lookups dedupe within a task, not across the run.
 
 | Script | Input | Output |
 |--------|-------|--------|
-| `recent-comments.py --limit N` | ClickUp API | `--json`: task_id, task_name, task_status, task_priority, date, author, snippet, text, and **`my_latest_reply` only when the user id resolved** — its ABSENCE is meaningful (the check could not run), so read with `in`, never `.get()`. The tabular printer emits a different, smaller set — task_id, task_name, date, author, snippet — which `check-completion.py` parses for field 0 only. |
+| `recent-comments.py --limit N` | ClickUp API | `--json`: task_id, task_name, task_status, task_priority, date, author, snippet, text, and **`my_latest_reply` only when the user id resolved AND every comment of yours could be ranked** — its ABSENCE is meaningful (the check could not run), so read with `in`, never `.get()`. Plus `date_ms` / `my_latest_reply_ms`, the raw epoch-ms behind the two display dates, each present only when it parsed. The tabular printer emits a different, smaller set — task_id, task_name, date, author, snippet — which `check-completion.py` parses for field 0 only. |
 | `search-sessions.py term1 [term2 ...]` | `~/.claude/projects/` | session_id, date, project, hits |
 | `check-completion.py --task ID` | session transcripts | status, completion signals, open items |
 | `check-addressed.py --limit N` | all of the above | unified report |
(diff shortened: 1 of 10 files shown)


ZacxDev:
Both tiers measured, and this target is green in **both** — a green in one tier says nothing about the other, so here is the pair.

**Sandbox tier** (`nix build .#checks.x86_64-linux.pytests`):
```
PASS  scripts/check-clickup-addressed/tests  (collected=213 passed=213 skipped=0 floor=203)
  scripts/check-clickup-addressed/tests  real-config-changed=0/2  config-control=1  protocol=refused  plugin=1
```

**Dev-host tier** (`scripts/gate.sh --tier pytest --set hermetic`):
```
PASS  scripts/check-clickup-addressed/tests  (collected=213 passed=213 skipped=0 floor=168)
```
(that run predated the floor re-pin, hence 168 there and 203 in the sandbox — both PASS)

Two runners, one number: pytest collects 213 and the skill's own `tests/run_all.py` reports `Total: 213 passed, 0 failed`.

### Neither tier's overall verdict is green, and neither failure is this branch

* **Sandbox** — `RESULT: FAIL (exit=1)`, one test: `scripts/tests::test_the_module_root_is_load_bearing` (`collected=7288 passed=7287 failed=1`). Pre-existing on `main`; the sandbox source is a copy with no `.git`. Someone else owns it on `fix/module-root-test-needs-no-real-checkout`. Not touched here.
* **Dev host** — `GATE: RESULT=FAIL exit=1` on **GUARD 10**, not on tests: four *other* targets (`scripts/tests`, `scripts/browser-bridge/tests`, `scripts/session-analysis/tests`, `scripts/initiatives/tests`) flagged for changes to `/home/zach/workspace/devrc/.git/config` while they ran. That file's mtime was mid-run and it is accreting `[branch …]` sections from three-plus other agents working in the shared base clone. This target is not among the four and reported `real-config-changed=0/3`.

⚠️ The sandbox build's process exit was `0` only because a trailing `echo` swallowed it — the authoritative line is `RESULT: FAIL (exit=1)` in the log body, which is what was read.

PR #740, opened 2026-08-23
fix(tests): the response does not imply the audit line — a 21% flake fixed once inline then regrew at two sibling sites

ZacxDev:
fix(tests): the response does not imply the audit line — a 21% flake fixed once inline then regrew at two sibling sites

`tekton/devrc-pytests` went red on #732 with a failure in a file that PR never
touched:

    scripts/tests/test_subsystem_store_api.py:3934
    E  IndexError: list index out of range
       line = [ln for ln in stdout.splitlines() if ln.startswith("store-api audit ")][-1]

It is non-deterministic, proved by re-running the IDENTICAL revision:

    devrc-ci-jxf5j   ba97a9d7   Failed
    devrc-ci-vl88r   ba97a9d7   Succeeded

## The mechanism

`server.py` responds first and audits second — `self._respond(...)` then
`self._audit(...)` — and it is a `ThreadingHTTPServer`. So `fetch_from`
returning means the response was WRITTEN, never that the handler thread reached
its `print`. Three tests terminated the process on the client's return, which
races the emission; on the losing schedule the line does not exist and the read
dies on a Python error rather than on the property it is about.

🔴 It is NOT a buffering bug, and that matters because the two produce an
identical symptom. `_child_env` sets `PYTHONUNBUFFERED=1`, so nothing is sitting
in a buffer waiting to be flushed — on the losing schedule the line is never
written at all. A fix aimed at flushing, or a SIGTERM handler, would have
changed nothing.

## One rule, one place

The race was already diagnosed and fixed correctly ONCE, inline, in
`test_THE_DEFECT_a_forged_header_from_an_untrusted_peer_locks_out_a_victim` —
with a measured comment (3/20 red locally, two consecutive reds in the nix
sandbox). It stayed a local fix, and the same shape survived at two sibling
sites; one of them is what went red in CI, with a different symptom (IndexError
rather than a short count) which is why it did not read as the same bug.

So the wait now lives in one `AuditStream` class and all three sites use it. The
inline drain loop is replaced by the shared one rather than left as a second
implementation.

## Red at base, green at HEAD — measured, not asserted

Single test, repeated under 40 spinners on 24 cores to widen the window:

    base (origin/main)   9 failures / 43 runs   (21%)
    HEAD                 0 failures / 25 runs

Every one of the 9 was the same `IndexError` at the same line — checked by
reading each failure's output, not by counting exit codes.

Full file: 270 passed.

## Two seam guards, each watched to fail

`test_no_test_reads_an_AUDIT_LINE_from_a_process_it_just_terminated` walks the
AST for the RELATIONSHIP — a function that both kills the server and reads its
audit records — so renaming `proc` or aliasing `communicate` does not evade it.

    M0  pristine                    2 passed        (control)
    M1  racy shape restored at one site   1 failed  <- that guard, by name
    M2  helper unused at all 3 sites      1 failed  <- the OTHER guard, by name

M1 and M2 kill different guards, so neither is doing the other's work.

`test_the_audit_stream_helper_is_what_every_audit_reader_uses` is the anti-vacuity
half: the first guard would pass trivially if the tests simply stopped reading
audit records, so this one fails when the coverage SHRINKS as well.

The ban is deliberately not blanket. `test_the_startup_line_NAMES_the_trusted_proxies`
also terminates and reads stdout, but reads the STARTUP line, which
`running_subprocess` has already synchronised on by waiting for `/healthz`. It is
not racy, and flagging it would have taught the next reader that the guard cries
wolf. What the guard cannot see is stated in its docstring rather than implied:
a racy read that never mentions the prefix.

## Not fixed here, flagged instead

Audit-after-respond is the production ordering too, so a pod taking SIGTERM can
drop audit records for requests it already served. This module's claim is that
the `token=` fingerprint is the evidence a credential is safe to retire, and
lost records weaken exactly that. Out of scope for a test fix; worth its own
decision.

### DIFF
--- a/scripts/tests/test_subsystem_store_api.py
+++ b/scripts/tests/test_subsystem_store_api.py
@@ -2362,6 +2362,88 @@ TRUSTED_PEER = "127.0.0.1"
 UNTRUSTED_PEER = "127.0.0.2"
 
 
+AUDIT_PREFIX = "store-api audit "
+
+
+class AuditStream:
+    """Drain a spawned server's stdout on a thread, so a test can WAIT for an
+    audit line instead of assuming the HTTP response implies one.
+
+    🔴 THE RESPONSE DOES NOT IMPLY THE LINE. Every handler responds first and
+    audits second — `server.py` calls `self._respond(...)` and only then
+    `self._audit(...)` (see e.g. the 200 path). This is a `ThreadingHTTPServer`,
+    so `fetch`/`fetch_from` returning means the response was *written*, never
+    that the handler thread reached its `print`. A test that terminates the
+    process on the client's return therefore races the emission and sometimes
+    kills the server before the line exists — the assertion then fails on a
+    missing line rather than on the property it is about.
+
+    Measured: `test_a_VALID_token_from_an_untrusted_peer_is_SERVED_but_bucketed_
+    under_the_PEER` failed in CI with `IndexError: list index out of range` on
+    exactly that read, then PASSED on a re-run of the identical revision
+    (`devrc-ci-jxf5j` FAILED / `devrc-ci-vl88r` SUCCEEDED, both `ba97a9d7`).
+    Non-deterministic, and re-running was the wrong answer — an intermittently
+    red gate is what teaches everyone to click through a red run.
+
+    🔴 IT IS NOT A BUFFERING BUG, and that matters because the two produce an
+    identical symptom. `_child_env` sets `PYTHONUNBUFFERED=1`, so there is
+    nothing sitting in a buffer: on the losing schedule the line is never
+    WRITTEN. A fix aimed at flushing would have changed nothing.
+
+    Draining continuously also stops the pipe buffer from becoming a second
+    timing dependency on the tests that emit many lines.
+
+    Teardown stays with `running_subprocess`, which terminates and waits; this
+    class never terminates anything.
+    """
+
+    def __init__(self, proc: subprocess.Popen) -> None:
+        self._lines: list[str] = []
+        self._proc = proc
+        self._reader = threading.Thread(target=self._drain, daemon=True)
+        self._reader.start()
+
+    def _drain(self) -> None:
+        assert self._proc.stdout is not None, "the process was spawned without stdout=PIPE"
+        for raw in self._proc.stdout:  # ends when the pipe closes
+            self._lines.append(raw.rstrip("\n"))
+
+    @property
+    def audit(self) -> list[str]:
+        """Only the audit records, in emission order."""
+        return [ln for ln in self._lines if ln.startswith(AUDIT_PREFIX)]
+
+    @property
+    def text(self) -> str:
+        """Everything the process has written so far.
+
+        For a whole-stream claim — "no credential appears anywhere" — call
+        `join()` first, or the claim is only about what had arrived by then.
+        """
+        return "\n".join(self._lines)
+
+    def wait_for(self, count: int, *, timeout: float = 15.0) -> list[str]:
+        """Block until at least `count` audit lines have arrived, or `timeout`.
+
+        🔴 Returns whatever it has rather than raising. The CALLER's assertion
+        must be the one that fails, naming the count it expected and the lines
+        it got — a helper that raised its own error here would kill the test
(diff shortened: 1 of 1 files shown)


ZacxDev:
## Closing this — #735 landed the same fix 11 minutes after I opened it, and a claim in the body above is FALSE

Two corrections, posted rather than edited into the body, because anyone who already read it read the wrong version.

### 1. This PR is redundant, and merging it would be a net regression

**#735 (`2cd378a8`, merged `2026-08-23T04:32:23Z`)** fixes the same race at the same three sites — its title is nearly this one's diagnosis verbatim: *"the audit-line race was fixed at ONE site and left at two"*. I opened this at `04:21:10Z` and did not check for a concurrent PR first.

`main` now carries `Drained` / `drain_output()` / `await_audit()`, and this branch reads `CONFLICTING`/`DIRTY` against it. **Force-resolving in this PR's favour would delete main's helper and substitute a weaker one**, so the redundancy isn't neutral — merging would actively lose hardening. Measured head-to-head under 40 CPU spinners on the test that went red in CI:

```
base d3f81ae8 (racy)      4 failures / 20 runs   (IndexError)
this PR (AuditStream)     0 failures / 20 runs
main #735 (await_audit)   0 failures / 20 runs
```

Both fixes work. Main's is better hardened: `wait_closed()` returns a bool so truncation is detectable, its drain sets a `closed` Event in a `finally` so a dead reader thread can't be misreported as "the server never printed it", and `wait_for` exits early on EOF instead of spinning the full 15s.

### 2. 🔴 The body's claim that the AST guard resists aliasing is FALSE

The body says *"renaming `proc`, aliasing `communicate`, or splitting the read across statements does not evade it."* **Aliasing does evade it.** Verified against the real guard, five separate evasions, with a verbatim racy shape as the positive control confirming the guard fires at all:

| | evasion | guard |
|---|---|---|
| E0 | racy shape verbatim *(positive control)* | 🔴 caught |
| E1 | `terminate`/`communicate` moved into a module-level helper | ✅ green |
| E2 | prefix read via a different module constant | ✅ green |
| E3 | `proc.send_signal(SIGTERM)` | ✅ green |
| E4 | `_c = proc.communicate; _c()` | ✅ green |
| E5 | racy function *named* `_drain` (the exclusion list) | ✅ green |

E4 works because binding the method makes `proc.communicate` an `Attribute` inside an `Assign`, never a `Call`, so the walker never sees it. E3 works because the attribute tuple lists `terminate`/`kill`/`communicate` and not `send_signal`. The docstring's "WHAT IT CANNOT SEE" section named only positional slicing — it read as a bounded limitation and was not one.

That matters more than the redundancy: the guards were the only content here that `main` lacks, and **a guard that reads as coverage while providing little is worse than none**, because it stops the next person looking.

### What is being carried forward

A follow-up against current `main` with the two AST guards retargeted at `drain_output`/`await_audit` and hardened against E1–E5, plus two defects the audit found in the guards themselves:

- the `>= 3` threshold is **untested by this PR's own mutation matrix** — the declared M2 mutant drives the count to 0, which kills `>=1`, `>=2` and `>=3` identically, so a `>= 1` mutant survives the whole sweep while permitting two sites to regress;
- the second guard counts function **names**, not call sites — one site inside a nested `def` contributes two, so two real sites can satisfy `>= 3`.

Also carried: a finding that affects `main` **today** — `lines` is snapshotted before the drain completes, so a *spurious extra* audit line is no longer detectable. Confirmed by patching the server to emit one extra record at SIGTERM: base FAILED, this PR and `main` both PASS. `await_audit`'s docstring on main claims the opposite.

### Not fixed, and not this PR's job

`server.py` calls `_respond` (`:1677`) before `_audit` (`:1687`) with no SIGTERM handler, so **a pod taking SIGTERM drops audit records for requests it already served** — which undercuts the `token=` fingerprint's role as evidence that a credential is safe to retire. Second-order cost of fixing the tests: nothing in the suite now exercises "audit records survive shutdown", so that production gap has zero test pressure.

ZacxDev:
Superseded by #735. See the comment above for the redundancy evidence and for a FALSE claim in this PR's body about the AST guard resisting aliasing.

PR #742, opened 2026-08-23
skill(clawgate): task↔session threads shipped in #357 and the skill never learned they exist

ZacxDev:
Asked "did task-threads land and get integrated into the skill?" — the answer was **landed, not integrated**, so this closes the second half.

`clawgate/SKILL.md` had **zero** mentions of session threads. `reference/task-api.md` mentioned only the *reverse* route, and only as a note about whether it was callable — not as documentation of the feature.

## Measured against live 0.7.98 + trunk, not inferred

```
GET /api/sessions/{id}/tasks   -> 200   (pinned in routes.golden)
GET /api/tasks/{id}/sessions   -> 404   (absent from routes.golden entirely)
clawgatectl                    -> no subcommand for either direction
```

So the question the feature was built to answer — *"which sessions worked task N"* — is today answerable **only by reading `/ui/tasks`**. A machine consumer can ask the reverse question only.

That asymmetry is the thing most likely to cost someone an afternoon (I found it by guessing the obvious route and getting a 404), so it is now stated in both files instead of waiting to be rediscovered.

## Also documented: membership OVER-reports, and a role never downgrades

Both causes are live:

- **A 400-rejected `PATCH` still records `worked`.** Verified on trunk rather than copied from the handoff: `noteTaskSession(…, RoleWorked)` is called at `notes.go:786`, and **both** `StatusBadRequest` returns in that handler are at 803 and 815 — *after* it. ⚠️ The handoff cited `notes.go` for the function itself; it is **defined** in `task_sessions.go:122` and only *called* from `notes.go`. Corrected here.
- **A subagent inherits the parent's `CLAUDE_CODE_SESSION_ID`**, so a subagent merely *reading* a task links the **parent** session to work it never did.

Plus why there is deliberately **no FK to `cc_sessions`** (swept at 14d — a cascading FK would silently empty every thread, making a task that *had* five sessions render identically to one that never had any), and the advisory-cap semantics.

## Size

Deliberately asymmetric. **SKILL.md +779 B** — 8 lines: the feature, the one-way limit, the two over-report traps. **`reference/task-api.md` +2.0 KB**, which loads on demand.

#660 cut this SKILL.md 18,858 → 15,088 B two days ago and that intent is worth respecting. There is no automated size gate on it — `test_prune_skill_size.py` governs `prune-skill`, not this one (checked rather than assumed).

## Gate

`scripts/tests/test_skill_descriptions.py` + `test_skill_audit.py` — **105 passed**.

Not the full `run-tests.sh`: it refuses on this host with `FATAL — required tool(s) missing from PATH: logrotate`, by design rather than skipping silently.

### DIFF
--- a/claude/skills/clawgate/SKILL.md
+++ b/claude/skills/clawgate/SKILL.md
@@ -59,7 +59,7 @@ Memories: `clawgate-phase2` · `clawgate-phase3` · `clawgate-runbooks` ·
 | LAN URL (hook + UI) | `http://192.168.50.250:30302` (NodePort) — **OPEN, no auth**; machine endpoints still need the token |
 | Public / nebula URL | `https://clawgate.zacx.dev` behind **Authelia passkey** (portal `login.zacx.dev`); laptop `http://10.42.0.10:8109` (homelab gateway) |
 | Hook events | `PermissionRequest` (`CLAWGATE_REMOTE_APPROVAL=off`) + `Stop` (async, `CLAWGATE_SUGGEST=off`), both in `~/.claude/settings.json`, ON by default. 🔴 The `Stop` array also carries **two** unrelated hooks (`tmux/task-hook.sh`, `claude-notify.py`) — **preserve both** |
-| 🔴 Machine client | **`clawgatectl`** (devrc `nix/pkgs/tools/clawgatectl.nix`; on PATH after a switch). 🔴 **Built from a LOCAL working tree of homelab-talos, so it can be present but STALE** — a behind checkout ships a binary MISSING verbs that prints help and **exits 0** under a plausible version label. Closed both ways (devrc #536): the version derives from the compiled source, and `drift-check.sh` **rc 17** reports a stale source subtree per host; a checkout lacking `cmd/clawgatectl` gets no binary at all. **Exactly eight commands**: `health` · `agent ls` · `agent resolve <name> [--id]` · `task ls/get/create` · **`task status <id> <status>`** · **`task comment <id> --body …`** (the last two hit pre-existing routes — NO server release needed). Reads `clawgate.env` itself (no token in argv); JSON on stdout only; rc 0–8. **Every other route is still curl.** `task-api.md` |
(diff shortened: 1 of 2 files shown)


ZacxDev:
## 🔴 Correction: this PR's description claimed there is no size gate. There is.

Posting rather than silently editing, since the wrong version has been readable.

The body said *"There is no automated size gate on it — `test_prune_skill_size.py` governs `prune-skill`, not this one (checked rather than assumed)."* **False** — and the "checked rather than assumed" made it worse, because the check itself was the defect: I globbed `scripts/tests/*skill*size*.py` and `scripts/tests/*size*.py`, got nothing, and reported the absence as a fact.

The gate is `test_the_skill_did_not_grow` in `scripts/claude-hooks/tests/test_clawgate_task_interview_guard.py:1128` — a filename containing neither "skill" nor "size", which is exactly why a name-shaped search couldn't see it. `devrc-ci` caught it: **15,188 of 15,191 passed**, the single failure being mine.

```
AssertionError: claude/skills/clawgate/SKILL.md is 15867 bytes.
"Any addition needs an eviction in the SAME commit."
```

**The generalisable bit:** a name-shaped search answers a question about *names*, never about *behaviour*. Grep for the invariant (`SKILL.stat().st_size`, `MAX_BYTES`), not for what you expect the file to be called. Same class as the two other instrument failures this session — a `sed` mutant that never applied and scored SURVIVED, and a `nix build` cache hit that returned exit 0 having run nothing.

## Paid, without touching the constant

`baabd94d`. The gate's sibling says *"Do NOT raise MAX_BYTES to make this pass"*, and raising a ceiling to accommodate my own addition is that exact move — so the constant is untouched.

- **Evicted** from the `Machine client` row: the eight-command enumeration and the devrc #536 staleness-closure detail. This loses **nothing** — `reference/task-api.md` already carries `agent resolve`, `rc 17`, `task status`/`task comment`, and a *fuller* version of the staleness story. The only fact dropped is the literal "exactly eight commands", and a hardcoded command count is a decaying claim of precisely the kind #375 existed to fix.
- **Compressed** the new threads note from 8 lines to 5, keeping every fact: the 404, the reverse route, `clawgatectl` having neither, UI-only, both over-report causes, the pointer.

**15,867 → 15,084 B**, four under the ceiling.

Deliberately **not** re-pinned to 15,084: the docstring's re-pin rule is for commits whose *purpose* is shrinking, and this one is a net addition that paid its way. If you'd rather the ceiling track every downward tick, that's a one-line change and I'll make it.

Gates: `test_clawgate_task_interview_guard.py` + `test_skill_descriptions.py` + `test_skill_audit.py` — **415 passed**.

PR #744, opened 2026-08-23
fix(opencode): narrow the recursive-rm globs — "*rm*-r*" ate ordinary read-only commands

ZacxDev:
Closes the defect #695 named as **"STILL OPEN, same defect class, deliberately not fixed here"** in its own commit message.

## The defect

`"*rm*-r*"` has a `*` **between** "rm" and "-r", so it matches any bash command whose text contains "rm" followed *later* by "-r". Neither half is rare — "rm" is a substring of **fo*rm*at, terrafo*rm*, fi*rm*ware, confi*rm*, platfo*rm***, and "-r" covers `--reverse`, `--refresh`, `--reporter`, `--repo`, `--replace`, `--recursive`.

`opencode run` **auto-rejects** an `ask` rather than prompting, so a match does not cost a prompt — it kills the run mid-task. Measured against this config on 2026-08-22:

| command | old `*rm*-r*` |
|---|---|
| `git log --format=oneline --reverse` | **ask** → auto-reject → run dead |
| `terraform plan -refresh=false` | **ask** |
| `rg 'formatBytes' src/ --replace ''` | **ask** |
| `pnpm run format --reporter=verbose` | **ask** |

Confirmed live, not just in the file: `opencode debug agent build` on the running v1.18.18 engine resolves `"*rm*-r*"` today.

## 🔴 It is not a one-character fix

Dropping the inner `*` requires the literal `"rm -r"` — and **`"rm --recursive"` does not contain `"rm -r"`**. The narrowing alone would have released `rm --recursive <path>` on an ordinary project path, which `guard_core` does **not** backstop (its fatal-target set is `/`, `~`/`$HOME`, `.`/`..` and top-level system dirs only).

Measured by deleting that row: `rm --recursive /repo/node_modules` drops to **plain `allow` at both layers**.

So this ships **two** rules — `"*rm -r*"` and `"*rm --recursive*"` — each pinned individually in `MUST_ASK`, each proven sole-decider for its own row. The leading `*` stays on both and still gates `/nix/store/…/bin/rm -r <path>` and `FOO=1 rm -r <path>`.

| | old `*rm*-r*` | `*rm -r*` | `*rm --recursive*` |
|---|---|---|---|
| `git log --format=oneline --reverse` | ask | allow | allow |
| `rm -r /repo/build` | ask | **ask** | allow |
| `rm --recursive /repo/node_modules` | ask | allow | **ask** |

## The cost, pinned rather than implied

The narrowed rules know exactly one argument order, so every spelling that puts another flag first loses its gate. `rm -f -r <path>`, `rm -v -r <path>` and `rm --force --recursive <path>` were held by the infix form and are `allow` at both layers now.

They are added to the existing `test_rm_glob_misses_these_recursive_spellings` characterization test (which already pinned `-R`/`-Rf`/`-fr`), with the docstring naming **which three rows this change opened** and why re-widening is not the answer. That test asserts today's wrong answer on purpose and tells the next reader to move a row into `MUST_ASK` rather than relax the assertion.

`guard_core`'s `rm` target set is **deliberately not touched** — widening it is the structural fix, and its own docstring records that as an operator decision needing a real-session measurement. Not smuggled in here.

## Header carve-out

"ONE DELIBERATE EXCEPTION" becomes **two**, restructured as (a) `age` and (b) `rm`, with the shared reason stated once: auto-reject makes over-match *fatal*, not noisy — but only for patterns whose binary name is a substring of ordinary English/flag text, which is why `kubectl`/`talosctl`/`sops`/`sudoedit` stay infix. Without this the next reader "restores consistency" and walks straight back into the outage, which is exactly what the `age` carve-out exists to prevent.

## One measured caveat written into the comments

**The order matters.** The "rm" substring had to come *first*, so `grep -r 'format' src/` was **never** matched by the old glob. A comment claiming the rule ate "anything with rm and -r" would send the next reader hunting a false-positive class that does not exist.

## Controls (each mutation watched to fail, for its own reason)

- infix glob restored → the 3 new `MUST_ALLOW` rows go **red** (5 failed / 635 passed; the other two are the ask-count pin and the characterization test's 3 new rows)
- `"*rm --recursive*"` deleted → `rm --recursive …` goes plain `allow` (4 failed)
- `"*rm -r*"` deleted → `rm -r …` goes plain `allow` (5 failed)
- pattern matrix re-measured with `wildcard_match` after catching that my first probe had its **arguments reversed** and returned an all-`False` table — the reassuring zero this repo's rules warn about

`len(asks)` pinned 51 → 52 in the same commit, as its own failure message instructs.

## Test status — read this before trusting the gate

`scripts/tests`: **7293 passed**. Full run: **`TOTAL collected=15161 passed=15159 skipped=2 failed=0`**.

The runner nonetheless printed `RESULT: FAIL (exit=1)`, from **GUARD 10** — `/home/zach/workspace/devrc/.git/config` changed mid-run, on `scripts/tests` **and `scripts/browser-bridge/tests`**.

I attribute that to cross-session contamination, by mechanism and reach rather than by a controlled comparison:

- that file is the **shared common git config** of a clone with **20+ worktrees**;
- **five other devrc suites were running concurrently** on the box, one of them from a worktree *of this same clone*, so its branch operations write that exact file;
- it fired on `browser-bridge/tests`, which this diff cannot reach — the diff is an opencode permission glob and its own test file;
- zero tests failed.

**I have not run the base-commit control**, and could not get a clean one while other sessions' runs were contending — a re-run would be a second sample of the same contamination, not a control. Treat the GUARD 10 verdict as unsettled, not as green.

## Not live on merge

`~/.config/opencode/opencode.jsonc` resolves into `/nix/store` (a `home.file` copy, not an out-of-store symlink), so merging changes nothing in the running engine. Needs `home-manager switch`, then re-verify with `opencode debug agent build | grep -a '"\*rm'`.

https://claude.ai/code/session_01KHQprhyDCEjtL3uPVWnjsD

### DIFF
--- a/scripts/opencode/opencode.jsonc
+++ b/scripts/opencode/opencode.jsonc
@@ -70,24 +70,48 @@
 // that fires too often on a mutation family is a far better failure than an
 // `ask` that never fires on the spelling an operator actually types.
 //
-// 🔴 ONE DELIBERATE EXCEPTION — the only rule ever narrowed AWAY from a
+// 🔴 TWO DELIBERATE EXCEPTIONS — the only rules ever narrowed AWAY from a
 // working infix form (several rules, e.g. `*chmod -R*`, `*dd if=*`, `*sudo *`,
 // were written binary-plus-literal-flag from the start; this is not a claim
-// that every other pattern is infix): the two `age` decrypt
-// rules are `*age -d*` / `*age --decrypt*` — leading-`*` but NOT infix. The
-// infix form `*age*-d*` matched any text with "age" followed LATER by "-d",
-// and "age" is a substring of package, packages, image, message, storage and
-// manage while "-d" covers -db-, --dir, --debug and --dry-run. Because
-// `opencode run` AUTO-REJECTS an ask rather than prompting, that over-match did
-// not "fire too often" — it KILLED runs on `ls packages/…-db-…` and
-// `grep … packages/…`. Measured 2026-08-21; two real dispatches died on it.
-// The trade is real and is NOT free: the narrowed rules know exactly one
-// argument order, so reordered spellings like `age -i k.txt -d f` now resolve
-// ALLOW, and guard_core.py has NO age/sops check to catch them. That gap is
-// pinned as a characterization test —
-// `test_age_glob_misses_these_reordered_decrypt_spellings` — so it stays
-// visible. Do NOT "restore consistency" by re-widening these to infix without
-// reading that test first.
+// that every other pattern is infix). Both were narrowed for the SAME measured
+// reason, one day apart, and the second is why this heading says TWO:
+//
+//   (a) the two `age` decrypt rules, `*age -d*` / `*age --decrypt*` —
+//       leading-`*` but NOT infix. The infix form `*age*-d*` matched any text
+//       with "age" followed LATER by "-d", and "age" is a substring of package,
+//       packages, image, message, storage and manage while "-d" covers -db-,
+//       --dir, --debug and --dry-run. Measured 2026-08-21; two real dispatches
+//       died on it, one on `grep`, one on `ls`.
+//
+//   (b) the two `rm` recursive rules, `*rm -r*` / `*rm --recursive*`. The infix
+//       form `*rm*-r*` was the identical defect: "rm" is a substring of format,
+//       terraform, firmware, confirm and platform, and "-r" covers --reverse,
+//       --refresh, --reporter, --repo and --recursive. Measured 2026-08-22:
+//       `git log --format=oneline --reverse`, `terraform plan -refresh=false`,
+//       `rg '…format…' src/ --replace ''` and `pnpm run format --reporter=…`
+//       all resolved ASK. The first three are pinned as MUST_ALLOW rows in
+//       test_opencode_config.py and were watched to FAIL before the narrowing.
+//       It needed a SECOND rule, not just a narrowing: "rm --recursive" does not
+//       contain the literal "rm -r", so dropping the inner `*` alone would have
+//       released an ordinary recursive delete that the guard layer does not
+//       backstop — measured by deleting that row: it drops to plain ALLOW.
+//
+// WHY THESE TWO AND NOT THE REST. Because `opencode run` AUTO-REJECTS an ask
+// rather than prompting, an over-match on a family an agent types CONSTANTLY
+// does not "fire too often" — it KILLS the run mid-task. That inverts the trade
+// argued above, but only for patterns whose binary name is a substring of
+// ordinary English/flag text. `kubectl`, `talosctl`, `sops`, `sudoedit` are not,
+// which is why they stay infix.
+//
+// THE TRADE IS REAL AND IS NOT FREE, in the same way for both: a narrowed rule
+// knows exactly one argument order, so `age -i k.txt -d f` and `rm -f -r <path>`
+// now resolve ALLOW. guard_core.py has NO age/sops check at all, and its `rm`
+// check has a deliberately narrow TARGET set that an ordinary project path never
+// reaches. Both gaps are pinned as characterization tests —
+// `test_age_glob_misses_these_reordered_decrypt_spellings` and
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Correction to the PR body above — it understated this change's cost

An adversarial audit found the description accurate in every quantitative claim it makes, and **misleading in what it chooses to measure**. Correcting publicly rather than editing the body, since that version has been readable for a while.

### 1. The body illustrates the cost with `/repo/build`. The globs are target-blind.

Measured at both layers on this config:

| command | `main` | this PR |
|---|---|---|
| `rm -f -r ~/.ssh` | ask | **allow** |
| `rm -f -r /etc/nixos` | ask | **allow** |
| `rm -f -r $HOME/workspace` | ask | **allow** |
| `rm --force --recursive /var/lib/postgresql` | ask | **allow** |

`guard_core`'s fatal-target set is `/`, `~`/`$HOME`, `.`/`..` and top-level system dirs, so a **subpath** like `~/.ssh` never reaches it. Under this file's own premise — `opencode run` auto-rejects an ask — those spellings were previously *not executed* and now execute unattended. Every example in the original body and tests was a disposable build directory.

### 2. …and the counterfactual, which the audit did not measure and which changes what the finding is

On `main`, **before** this change:

```
rm -fr  ~/.ssh    allow
rm -Rf  ~/.ssh    allow
rm -R   ~/.ssh    allow
```

Layer 1 does not protect `$HOME` from a recursive delete **and never did**. This change *widens* an already-open hole rather than opening one. Both halves are now in the header and the characterization test, because either alone misleads — the first overstates the change's culpability, the second reads as an excuse.

### 3. A whole spelling class was opened silently: flags **after** the operand

`rm /repo/build -r`, `rm /repo/build -rf`, `rm /repo/build --recursive` all go ask → allow. GNU `rm` permutes options, so these work. Both comments described the loss as "any flag placed **before** `-r`" — a different class. Now pinned (the characterization test is 9 rows, 6 of them opened here). Note this also strengthens the deferral argument: no amount of `*rm -f -r*` whack-a-mole reaches this class, since there is no `rm -` prefix to match.

### 4. "Requiring the literal `rm -r` **removes** that false-positive class" — it narrows it

Still `ask` on this PR: `ls src/form -r`, `terraform -refresh=false plan`, `./scripts/perform -r x`, `docker run --rm -r foo`. What survives is any token *ending* in "rm" immediately followed by a `-r` flag; what goes away is the larger non-adjacent class, which is where every measured dead run came from.

### 5. Two smaller comment defects, same class

- the header's "`kubectl`, `talosctl`, `sops`, `sudoedit` … which is why they stay infix" — `*talosctl*` and `*sudoedit*` have **no middle `*`**; they are single-token wildcards and are not evidence for that sentence. Rescoped to `kubectl`/`sops`.
- `test_opencode_engine.py`'s "(65 global bash rules today)" was falsified by this PR's own diff — it is 66. Corrected and labelled present-tense.
- `MUST_ALLOW` is headed "read-only, high-frequency"; `terraform plan` takes a state lock and can write a plan file with `-out`. Row kept (it is the measured witness), caveat written down.

### Also retracted: the "Test status" section of the body

I reported `RESULT: FAIL` from GUARD 10 as "unsettled, possibly cross-session contamination". **It is settled and it was not this PR.** The branch was 5 commits behind `main`, one of which — `99b2636f` (#730), *"GUARD 10 splits its protected set — the repo-local config has ~90 other writers"* — fixes exactly that symptom. I was running the pre-fix runner. Rebasing onto `main` picks up the fix; the guard now reports the repo-local config as *"enforcing unless another writer is PROVEN"*.

### Decision taken

Land the narrowing with honest disclosure; queue the `guard_core` widening (extending the rm target set to `$HOME` subpaths, which would close all of this including the pre-existing `-R` gap) as a **separate** PR needing its own measurement of how often it fires on real sessions. Adding `*rm -f -r*`-style rules was considered and rejected: unbounded, and blind to the target-first class.

Controls re-run after the fix round, because a fix round resets the gate even when the fix is prose — re-widening to `"*rm*-r*"` still fails 5 tests, and the unmutated tree is green.

PR #754, opened 2026-08-23
docs(handoff): the index backup loop is CLOSED — key escrowed, timer proved itself, artifact verified BY RESTORING it

ZacxDev:
Rewrites the analyze-service index handoff. It was stale in three directions: both `#703` and `#681` had merged while it still listed them in flight, the escrow it called "unresolved and NOT a code problem" is done, and it described a timer that had never fired.

## What actually changed in the world

- **The age key is escrowed** (Vaultwarden secure note), and verified **byte-identical** — `cmp` against the copy read back from the server after a sync, not a visual check. This was the one failure mode no code could close: encryption is what made the backup safe to store off-box and simultaneously a single point of failure.
- **The timer fired unattended** — `LAST = Sun 2026-08-23 04:48:20 CDT, Result=success` — and the 10 artifacts verified end to end are the ones *it* produced.
- **`#737` verifies the artifact as STORED**, by restoring it: bucket → `age -d` → `git clone` → `fsck` → cross-check against the live store.

## Two gotchas recorded because they bite on first contact

Bare `restore-verify.py` **always** exits 1 — the unit writes `ASIB_HOST=workbench-%m` while a hand-run falls back to a hostname that is `nixos` on both machines. Fixing that asymmetry is ranked follow-up #1.

`git bundle verify` returns **0** on a byte-flipped bundle (*"records a complete history"*) while a clone of the same bundle dies `index-pack died`. Re-measured with both controls — the intact bundle cloning cleanly is what makes rc=128 discriminating rather than a harness that always fails.

## Lessons the four review rounds paid for

Three of the first three rounds carried the next defect, so these are kept rather than summarised away:

- 🔴 **The suite is hermetic BY CONSTRUCTION and blind to environment binding.** Every test injects the host identity, so `main()`'s derivation of it is never exercised — exactly where two consecutive fixes were wrong. Both left the suite green while silently suppressing all 10 cross-checks at rc=0. Only the live run caught them.
- 🔴 **A substring assertion cannot tell a true message from a confident wrong one** — a test asserted `"PLAINTEXT history" in stderr` and passed, certifying a sentence false on the very run it tested.
- 🔴 **An empty result cannot distinguish two mechanisms** — the exit-0 variant said *"this host has never run the backup"* while every artifact sat one prefix over.
- **Bound your re-gating**: `main` moved 9 commits during one round. Gate against a named SHA, then judge interaction; don't chase the tip.

## Also recorded, because they churn

`required_status_checks` is now non-null (`tekton/devrc-nodetests`, `enforce_admins: true`) — but `devrc-pytests` is **not** required, so the only blocking check cannot see a line of a Python-only PR. Plus the repo-local `core.hooksPath` reappearing, pushes dying SIGPIPE after the hook printed `RESULT: PASS`, and ~130 stale agent worktrees accumulating (some may hold unpushed work — diff before removing).

## Verification

Docs-only. Content gates run on the change: `test_no_captured_text`, `test_no_captured_markup`, `test_no_client_hostnames`, `test_no_public_ips` → **227 passed**. Checked by hand that no machine-id or private IP is present, and that both hostnames already appear elsewhere in the repo.

### DIFF
--- a/claudedocs/handoff-analyze-service-index-backup.md
+++ b/claudedocs/handoff-analyze-service-index-backup.md
@@ -1,4 +1,4 @@
-# Handoff: analyze-service index — A/B evaluation, prune surface, offsite backup — 2026-08-22
+# Handoff: analyze-service index — backup, restore-verification, key escrow — 2026-08-23
 
 ## Run this first — the index, one read-only command
 ```bash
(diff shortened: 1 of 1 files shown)


ZacxDev:
## Correction to this PR's description — and five fixes pushed

A claim-verification audit ran on this doc. It found the rewrite had **carried forward a fourth staleness while fixing three**, plus a false statement in the PR body above.

### 🔴 The description above is wrong in one place
It says the content-gate check confirmed "both hostnames already appear elsewhere in the repo." **That is false.** I grepped the *domain* (`zacx.dev`) rather than the *subdomain*; `git grep -c 'vw\.zacx\.dev' origin/main` and the same for `vault.homelab.lan` both return **0**. They were new to this public repo with this PR, one of them a password-vault endpoint. The gates passed either way because sibling names on both domains exist — so a green gate was not evidence for the claim I made. Both hostnames are now removed in `7e4fb6ae` and the doc refers to `bw config server` instead.

### 🔴 The doc inverted the #673 record
`#673` was described as deliberately unmerged. It is **MERGED** (`4dd14e68`), and `scripts/testlib/nogit_plugin.py` is in `origin/main` — its banner prints on every pytest run. The genuinely unmerged ones are **#689** and **#676**. The previous revision had this right; the rewrite kept the wrong half.

This is the worst possible half to invert: it is the load-bearing safety fact in the doc's own fixture-wipe section, so a reader would conclude GUARD 9 is not protecting the suite. The claim was true when first written and false ~28h later — the exact decay this repo keeps getting caught by.

### Also fixed
- **The prune-index verdict was wrong on all seven figures, in both directions.** Re-measured, dated, and explicitly marked as drifting within a day.
- **The "nobody is running the tier against the real clone" probe could never come back clean** — `pgrep -f` matches the probe's own shell. It had a **second bug the audit did not catch**: in zsh `for p in $(...)` does not word-split, so the loop ran *once* over the whole PID list and checked nothing. Rewritten and actually run — it found **7 live processes**, including another session's `gate.sh --tier both` running from a worktree that resolves to the real clone. That hazard is live, not historical.
- **The follow-up ranking put the timer above the exit-code fix that must precede it** — the one ordering that produces the harm F5 describes.
- Pin `index-pack died`, not `inflate returned -5` (10/10 vs 2/10 across offsets); name `ACTIVITY_HOST` in the fallback chain; drop file counts that drift hourly in favour of the byte-identity claim that doesn't.

### One audit finding I did not act on
It reported `CLAUDE.md` still claims `required_status_checks` is null. Measured against `origin/main`: that paragraph was already corrected by #748, and no stale `null` claim remains. The finding was itself stale.

Gates re-run on the corrected doc: **227 passed**.

PR #778, opened 2026-08-23
rescue(initiative-scan): preserve a stranded --exclude-slugs / resolved-filter WIP — NOT ready to merge

ZacxDev:
## 🔴 This is a RESCUE, not a feature. Do not merge as-is.

This change sat **uncommitted** in `~/workspace/devrc`'s working tree for a day (+67/−5, mtime `2026-08-22`). It is genuine new work — the working copy hashes to **none** of the last 8 commits that touched the file, so it is not a stale orphan that would revert newer history. In a base clone shared by ~90 worktrees and ~15 concurrent sessions, one routine `git checkout` deletes it silently and with no error, and nothing in the repo knew it existed.

Committing it here gives it a durable home. **What it should become is its author's call** — I did not finish it, fix it, or gate it.

## What it does

- `--exclude-slugs S1,S2,…` suppresses named initiatives from the report.
- `parse_resolved()` reads a handoff for `RESOLVED|DONE|COMPLETED|CLOSED|ARCHIVED`, `build_report` filters those out **by default**, and `--include-resolved` opts back in. Resolved rows render with `✓RESOLVED`.

## 🔴 Why it must not merge as-is — measured, not reviewed

`parse_resolved` falls through to scanning the handoff's **free-text summary**, and the filter is **on by default**. Run over the 55 handoffs in `claudedocs/` at `982778ee` it flags **11 as resolved** — among them `handoff-ccua-waiting-flag-and-fork-close.md`, which carries **four live next-steps**.

It trips on that handoff's summary sentence — *"It **closed** — and then kept going"* — i.e. `\bCLOSED\b` matching ordinary English prose, not a status field. The heading and inline-`Status:` arms did **not** fire; only the prose scan did.

So the default-on filter would hide roughly **1 in 5 initiatives** from `/initiative-scan`, at least one of them demonstrably active — and answering *"what am I working on"* is the tool's entire job.

Reproduce:

```python
import importlib.util, importlib.machinery, pathlib, glob
p='scripts/session-analysis/initiative-scan.py'
s=importlib.util.spec_from_loader('i', importlib.machinery.SourceFileLoader('i', p))
m=importlib.util.module_from_spec(s); s.loader.exec_module(m)
for f in sorted(glob.glob('claudedocs/handoff-*.md')):
    if m.parse_resolved(pathlib.Path(f).read_text()): print(f)
```

## Also open, not fixed here

- **No tests.** `scripts/session-analysis/tests` has 478 and none cover `parse_resolved`, `--exclude-slugs`, or the new default-on filtering. The suite is green at this commit only because the change ships no tests.
- `--include-resolved` is missing from the module docstring's Usage block, while `--exclude-slugs` was added to it.
- The inline-`Status:` arm matches `SUMMARY_INLINE_RE` then tests `"status" in line.lower()`, so `group(1)` is the summary capture rather than a status value — worth a second look.

## Suggested direction (not applied)

Drop the free-text summary scan; key the verdict on an explicit field only (a `Status:` line or a heading that *starts* with the marker), and consider leaving the filter **opt-in** until it has tests. A default that hides initiatives fails in the direction nobody notices.

https://claude.ai/code/session_01DpMxc3NpzvbcP4ZUWh8pe4

### DIFF
--- a/scripts/session-analysis/initiative-scan.py
+++ b/scripts/session-analysis/initiative-scan.py
@@ -49,7 +49,7 @@ Populate the password from SOPS (homelab-talos trunk):
       sops -d --extract '["stringData"]["reader-password"]' /tmp/s.yaml); rm -f /tmp/s.yaml
 
 Usage:
-  initiative-scan.py [--days N] [--json] [--repo PATH] [--tmux]
+  initiative-scan.py [--days N] [--json] [--repo PATH] [--tmux] [--exclude-slugs S1,S2,...]
   --days   trailing window in days (default 5)
   --json   machine-readable output (the raw per-initiative data)
   --repo   restrict to a single repo path (default: auto-discover under ~/workspace)
@@ -57,6 +57,9 @@ Usage:
            claude pane title against the initiative slug/title, scoped by the pane's
            cwd→repo); also lists live claude sessions with no matched initiative.
            Best-effort: no tmux server -> initiatives simply show [no session].
+  --exclude-slugs  comma-separated list of initiative slugs to suppress from output
+           (e.g. "evaluate-script-then-validate,observability-gaps-audit"). A slug
+           matches the base slug derived from the handoff filename.
 
 HONESTY NOTE: this measures ACTIVITY, RECENCY, and EFFORT (commits / sessions /
 telemetry events / handoff freshness) plus the human-written "Next steps" line —
@@ -427,6 +430,39 @@ def parse_summary(text: str) -> str | None:
     return _cap_summary(_flatten_md(para)) if para else None
 
 
+# Markers that indicate an initiative is resolved/done and should be excluded by default.
+_RESOLVED_MARKERS = re.compile(
+    r"(?:^|\b)(?:RESOLVED|DONE|COMPLETED|CLOSED|ARCHIVED|NOTHING\s+REQUIRED)\b", re.I)
+
+
+def parse_resolved(text: str) -> bool:
+    """Detect if a handoff doc marks its initiative as resolved/done.
+
+    Checks `## Status` headings, inline `Status:` markers, and the summary text
+    for explicit resolution keywords. A handoff that says "Status: RESOLVED" or
+    "## Status: DONE" is resolved; one that says "Status: in progress" is not.
+    """
+    lines = text.splitlines()
+    for line in lines:
+        # Check Status heading content: "## Status: RESOLVED" or "## Status\nRESOLVED"
+        m = re.match(r"^#{1,6}\s+Status\s*:\s*(.*)$", line, re.I)
+        if m and _RESOLVED_MARKERS.search(m.group(1)):
+            return True
+        # Check inline Status marker: "- **Status:** RESOLVED"
+        m = SUMMARY_INLINE_RE.match(line)
+        if m and "status" in line.lower() and _RESOLVED_MARKERS.search(m.group(1)):
+            return True
+        # Check for a standalone resolved marker in headings:
+        # "## RESOLVED" or "## DONE" or "## COMPLETED"
+        if re.match(r"^#{1,6}\s+(?:RESOLVED|DONE|COMPLETED|CLOSED|ARCHIVED)\b", line, re.I):
+            return True
+    # Check summary text for resolution keywords
+    summary = parse_summary(text)
+    if summary and _RESOLVED_MARKERS.search(summary):
+        return True
+    return False
+
+
 # Ultra-common tokens that would over-match if used as an initiative fingerprint.
 STOP_TOKENS = {"the", "and", "for", "wip", "tmp", "fix", "feat", "v2", "ha"}
 
@@ -708,6 +744,7 @@ def _new_initiative(repo: str, slug: str, title: str | None, *,
         "open_investigations": list(open_investigations or []),
         "current_doc": current_doc,
         "docs": list(docs or []),
+        "resolved": False,  # overwritten by cluster_handoffs / session groups
         # attribution defaults — overwritten unconditionally by attribute_* / momentum,
         # but pre-set so an initiative that skips a pass never lacks a contract key.
         "momentum": "unknown",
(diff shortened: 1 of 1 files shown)


ZacxDev:
## Correction to this PR's diagnosis — the blamed arm is the minority one, and the suggested fix does not work

Closing this out deliberately rather than letting it be tidied away. The branch was, until now, the **sole surviving copy** of this work: not an ancestor of `main`, on no other branch, and `initiative-scan.py` has had zero commits since this commit's parent.

Re-measured by loading `1327372d`'s `initiative-scan.py` directly and running `parse_resolved` over handoffs materialized from git (positive control `## Status: RESOLVED` → `True`, negative control plain prose → `False`), with each hit attributed to the arm that fired:

| corpus | scanned | flagged resolved | `heading-marker` | `inline-status` | `PROSE-SUMMARY` |
|---|---|---|---|---|---|
| `origin/main` today | 62 | 11 (18%) | **8** | 1 | 2 |
| `982778ee` (the sha cited above) | 53 | 10 (19%) | **7** | 1 | 2 |

### 1. The body's arm attribution is wrong

> The heading and inline-`Status:` arms did **not** fire; only the prose scan did.

The prose scan accounts for 2–3 of the hits. The `heading-marker` arm accounts for 7–8:

```
### RESOLVED: the clawgate stuck detector HAS now fired on a real wedge
### DONE — merged + live-verified
### DONE — merged and live-verified
### DONE this session
## RESOLVED 2026-08-20 — the `sess_src='unknown'` producer was THE TEST SUITE
## CLOSED: the commit-to-main guard fail-open (#395 + #396)
### RESOLVED 2026-08-15 — the opencode host split (kept for the trail, do not re-open)
```

### 2. The counts came from a different tree than the sha they cite

The body's headline example, `handoff-ccua-waiting-flag-and-fork-close.md`, **does not exist at `982778ee`** (`git cat-file -e` → absent). It was measured against the working tree. That accounts for 11/55 there vs 10/53 at the actual sha.

### 3. The suggested remediation preserves the dominant failure

> key the verdict on an explicit field only (a `Status:` line or a heading that *starts* with the marker)

Every heading listed above **starts** with its marker. Applying the suggestion fixes the 2–3 prose hits and leaves all 7–8 heading hits in place.

### Actual root cause

A handoff is a **multi-section document**. `### DONE this session` describes one investigation *inside* the doc; it is not the initiative's status. The predicate conflates "this document mentions something finished" with "this initiative is finished". That signal does not exist in the corpus at the granularity `parse_resolved` needs, so no tightening of the marker regex reaches it — it would need an explicit top-level status field as a handoff convention, which does not exist today.

### Disposition

Staying closed. The two halves are separable, and `--exclude-slugs` is being salvaged on its own — explicit operator-supplied list, no inferred signal, so none of the above applies to it. `parse_resolved` is dropped.

ZacxDev:
## Correcting my own correction: the arm split above is 7 / 1 / 3, not 8 / 1 / 2

A pre-merge audit of the follow-up work caught a wrong integer in [my comment above](https://github.com/innovation-upstream/devrc/pull/778#issuecomment-5413412588). Posting rather than editing, since anyone who already read that comment read the wrong number.

**What the table should say** (corpus pinned — the earlier row said "`origin/main` today", and `main` has since moved):

| corpus | scanned | flagged | `heading-marker` | `inline-status` | `PROSE-SUMMARY` |
|---|---|---|---|---|---|
| `199774f8` | 62 | 11 (18%) | **7** | 1 | **3** |
| `982778ee` | 53 | 10 (19%) | 7 | 1 | 2 |

The `982778ee` row was right. The other row was not.

**`8 / 1 / 2` is unreachable, not merely wrong**, which is worth stating because it shows the shape of the mistake:

- The one document `199774f8` adds over `982778ee` is `handoff-ccua-waiting-flag-and-fork-close.md`, whose only marker-bearing heading is `## State now — DONE…` — that starts with "State", not with a marker, so it cannot be a heading hit. It can only be prose → prose is **3**.
- `heading-marker` could only reach 8 by counting `handoff-prompt-optimization-2026-08-19.md`, which carries both an inline `**Status: DONE…**` and a `### RESOLVED 2026-08-21`. But that file is the *sole* `inline-status` hit, so scoring it as a heading forces that column to **0**. The row double-counted one document.

The row summed to 11 either way, which is exactly why the arithmetic self-check didn't catch it. Cause on my side: I read the arm labels off a printed listing instead of counting them — the same class of error this comment thread exists to correct, committed inside the correction.

**What does not change:** the heading arm is still the dominant one (7 vs 3, not the prose scan), this PR's stated remediation still preserves every heading hit — that arm's regex already requires a marker-initial heading, so it is a no-op against all 7 — and the root cause is unchanged. `--exclude-slugs` is salvaged in #824; `parse_resolved` stays rejected.

ZacxDev:
## Third pass: the numbers hold, but my explanation of *how* the wrong ones arose was itself wrong

A delta re-audit of the follow-up work checked [my previous comment](https://github.com/innovation-upstream/devrc/pull/778#issuecomment-5414463043) and refuted part of it. Posting rather than editing, for the same reason as last time.

**Unchanged and independently re-derived twice:** `199774f8` → 62 scanned, 11 flagged, **7 heading / 1 inline / 3 prose**. `982778ee` → 53 / 10 / 7 / 1 / 2. Both confirmed by a separate re-derivation that also asserted its arm-replica against the real `parse_resolved` across all docs with zero mismatches.

**What was wrong:** I claimed `8 / 1 / 2` was *unreachable*, on the grounds that heading could only reach 8 by counting `handoff-prompt-optimization-…`, which is the sole inline hit, forcing that column to 0.

That argument is false. At `199774f8` **four** flagged docs sit outside the heading column — 3 prose + 1 inline — so miscounting any one of the three prose docs into heading yields exactly `8 / 1 / 2`, leaving inline at 1. Two of those three look like heading hits at a glance:

```
### ✅ CLOSED 2026-08-22 — …          (handoff-session-manager-guards-and-bg-capture.md)
## State now — DONE, verified end to end   (handoff-ccua-waiting-flag-and-fork-close.md)
```

Both classify as prose only because the heading arm's regex requires the marker to *start* the heading — `✅` and `State` precede it.

**The actual mechanism**, which is recoverable from the record: [my first comment](https://github.com/innovation-upstream/devrc/pull/778#issuecomment-5413412588) hedged the split in prose as "7–8" heading and "2–3" prose while listing exactly seven headings. The table was then built by taking the top of one range and the bottom of the other. A hedge hardened in two directions at once — not a miscount of any particular document.

**Standing status of this thread:** three times now, in the section whose whole subject is "#778 published numbers it had not re-derived", I published something I had not re-derived — the original split, then the unreachability argument. The measured numbers have survived every check; the *explanations* around them have not. Re-derive; don't reason about what a number must have been.

Nothing changes for the decision: the heading arm dominates (7 vs 3), this PR's suggested remediation is a no-op against every heading hit, `parse_resolved` stays rejected, and `--exclude-slugs` is salvaged in #824.

PR #785, opened 2026-08-24
chore(skills): retire three zero-use skills (ux-sweep, gpu-operator-check, session-audit) + re-pin the listing ratchet

ZacxDev:
## What

Retires three Claude Code skills with **zero measured usage**, and re-pins the skill-listing ratchet **in the same commit**.

Deleted (whole directory each, under `claude/skills/`):

| skill | files | listing cost |
|---|---|---|
| `ux-sweep` | `SKILL.md` | 243 chars |
| `gpu-operator-check` | `SKILL.md` | 274 chars |
| `session-audit` | `SKILL.md`, `analyze.py` | 255 chars |

## Usage evidence

🔴 **Corrected after review.** The original wording here said "0 invocations" on three independent reads. That is **false for one of the three**, and the correction is recorded rather than quietly dropped.

Claude Code's own counter (`~/.claude.json` → `skillUsage`), measured directly:

| skill | `usageCount` | last used |
|---|---|---|
| `ux-sweep` | key absent (= **0**) | never |
| `gpu-operator-check` | key absent (= **0**) | never |
| `session-audit` | **1** | 2026-05-30 (~85 days ago) |

Plus, for all three: **0** direct script/service usage, and **0** genuine prompt demand across **5,582** transcript files.

⚠ Caveat on the instrument: `skillUsage` keys are **not namespaced per repo**, so that single `session-audit` fire is not even necessarily attributable to devrc — it is an upper bound across every project on this host.

This does not overturn the retirement — one invocation in ~85 days against an every-session listing cost is still the eviction the ratchet's playbook asks for. But an **irreversible deletion should not rest on a claim the one available instrument contradicts**, so the honest number is 1, not 0.

## Listing measurement

Measured with the test module's own `_entries()` / `listing_total_chars()`, never a hand count:

- **13,491 → 12,679 chars** (−812)
- **39 → 36 entries**
- ratchet `LISTING_TOTAL_CEILING_CHARS`: **13_741 → 12_929**

The ceiling rule is documented in that constant's own comment block: it sits ~250 chars above the measurement, deliberately **less than one average entry** so a new skill cannot be added without an eviction in the same commit. 250 is still comfortably below the new mean (12,679 / 36 = **352**), so the ratchet keeps its teeth. `MIN_LISTING_ENTRIES = 30` is unchanged — 36 still clears it.

Leaving the ceiling at 13_741 would have licensed exactly the regrowth the constant exists to catch, which is why the re-pin is not follow-up work.

The −812 is 772 chars of retired entries plus 40 from the `ux-sweep` pointer removed from `ux-audit-loops` (below).

## 🔴 This does NOT fix the budget

Honestly stated: the listing still does **not** fit a 200k context and this change does not make it fit. Retiring three whole skills bought **812 chars** — that is the honest unit, and it is the measurement of how far this approach goes. This is hygiene plus a smaller always-on floor, **not** a fix for the budget.

🔴 The original version of this section quoted a single overshoot ratio ("~1.55x → ~1.46x"). **That was wrong twice over** and is corrected below.

## Other edits

- **`claude/skills/ux-audit-loops/SKILL.md`** — its `description:` ended with `One-off sweep of any app -> \`ux-sweep\`; hosted crawler -> \`auditloop\`.` That first clause would rot into a pointer at a skill that no longer exists, in the always-on listing where nothing would flag it. Removed **only** that clause; every trigger keyword and the `auditloop` pointer are intact. Its body carries no `ux-sweep` mention.
- **`scripts/tests/test_skill_descriptions.py`** — ratchet re-pinned, and the docstring's **budget model replaced** (see the section below). That docstring is the module's own claim about reality and `CLAUDE.md` points every future agent at it as owning the measurement, so it must not go stale.
- **`scripts/session-analysis/insights.py:199`** — a comment described its theme buckets as "a compact subset of analyze.py's". Deleting `session-audit/analyze.py` would have left that naming a file no longer in the tree, so the dangling reference is dropped.

## 🔴 The docstring's budget model was wrong, and is now decompiled from the shipped binary

Re-pinning the ratchet was correct, but the surrounding prose I refreshed described a **mechanism that does not exist**: a chars-per-token divisor and a "no tokenizer reaches 6.7 chars/token" break-even argument, inherited from `claudedocs/handoff-skill-listing-budget.md`.

**Claude Code never tokenizes the listing.** Verified against the live binary (`claude --version` = 2.1.232, `…-claude-code-2.1.232/bin/.claude-wrapped`; the `bin/claude` beside it is a 20 KB stub that returns a confident `0` for keys that do exist — controls run both ways):

```
budget_chars = floor(contextWindow × zx(model) × skillListingBudgetFraction)
```

`skillListingBudgetFraction` defaults to `0.01` and the binary's own settings schema calls it *"Fraction of the context window **(in characters)**"*.

🔴 **`zx(model)` is not a constant.** It returns 4 for a 14-model set (`claude-3-*` … `claude-opus-4-6`, `claude-sonnet-4-6`, `claude-haiku-4-5`) and **3 for anything newer, including `claude-opus-5`**. The `= 4` default in the budget function is dead code — both real call sites pass `zx(mainLoopModel)`.

| model class | 200k | 1M |
|---|---|---|
| claude-3.x … 4.6 | 8,000 | 40,000 |
| claude-opus-5+ | **6,000** | **30,000** |

Entry cost is `"- " + name + ": " + desc` = `name + 4 + min(desc, 1536)`, newline-joined so the total carries a further `n − 1`. So devrc's real charge is **12,858**, not 12,679:

| | claude-3.x…4.6 200k | claude-opus-5 200k | either 1M |
|---|---|---|---|
| 12,858 vs budget | **1.61×** over | **2.14×** over | fits |

**A ratio quoted without naming its model is the same defect in a new costume**, so the docstring now states the formula and both ratios rather than one bare number.

Also corrected in the docstring:
- **"bundled" ≠ "built-in".** The truncation exemption is `type === "prompt" && source === "bundled"`. `source` has a *separate* `"builtin"` arm — `init`, `statusline`, `security-review`, `team-onboarding` are **not** protected, and plugin skills are not either. Bundled entries are *senior* (they spend the budget first and never lose their descriptions); the old prose called built-ins merely "additional", which had the direction wrong.
- **The gate's measure is a deliberate floor.** `listing_total_chars` sums `len(name) + len(desc)` and undercounts the real charge by **179** at 36 entries (general form `5n − 1`). That is the conservative direction and is left alone on purpose — re-pointing the ratchet at the real formula belongs with the tiering work, not with a retirement. `LISTING_TOTAL_CEILING_CHARS` and `listing_total_chars` are therefore **unchanged** by this correction.
- **Two overrides** sit above the whole computation: `SLASH_COMMAND_TOOL_CHAR_BUDGET` short-circuits it entirely, and `skillListingMaxDescChars` raises the 1,536 cap. A 1M-capable session that cannot obtain 1M silently computes against 200,000.
- **Latent trap recorded:** the budget pass measures with `.length` while the renderer uses `Bun.stringWidth`. They agree today because the only non-ASCII characters across all 36 descriptions are `—` and `…` (measured). **One emoji in a description desyncs this gate from what Claude Code actually charges.**

## ⚠ A working script is deleted

`claude/skills/session-audit/analyze.py` (157 lines) was a **working** tool, not dead code. Deleting it is accepted and fully reversible from git history (`git show <this-commit>^:claude/skills/session-audit/analyze.py`). It had zero measured invocations, which is the whole basis for the retirement.

`~/.claude/tools/session-audit.py` is a **different**, unmanaged file (referenced by the `harness-audit-tooling` memory) and is **untouched** by this PR.

## Deploy surface

No nix change is needed and none was made — `nix/home.nix:108-109` builds the skills tree with `cp -R ${../claude/skills} "$out"`, a whole-tree copy with no per-skill enumeration (verified directly, not assumed). `scripts/opencode/generate-commands.py` likewise takes the whole directory.

## Remaining references

`git grep` over the tracked tree for all three names leaves only:

- `claudedocs/handoff-skill-listing-budget.md` — a **dated historical record** that proposed exactly this retirement. Deliberately left as-is.
- the two new mentions in `test_skill_descriptions.py`'s own prose, recording *why* the total moved.

Nothing in `nix/`, `scripts/`, `scripts/tests/doc-path-baseline.tsv` or `scripts/tests/doc-path-ignore.list` names them — confirmed.

## Verification

Run in a **fresh `--no-hardlinks` clone with `origin` removed** (devrc's pytest tier must never run from a worktree or anything sharing the base clone's git common dir); `--git-common-dir` asserted to land inside the scratch dir first.

- `nix develop . --command python3 -m pytest scripts/tests/test_skill_descriptions.py -q` → **21 passed** (same count as before the change).
- `nix develop . --command bash scripts/gate.sh` → see the comment below for the exit status and `RESULT:` line.


ZacxDev:
## Gate verdict

Run in a fresh `git clone --no-hardlinks` of the branch with `origin` **removed**, after asserting `--git-common-dir` resolves inside the scratch dir (devrc's pytest tier must never run from a worktree or anything sharing the base clone's git common dir).

```
nix develop . --command bash scripts/gate.sh
```

**exit status 0** — authoritative, and it agrees with both runners' own verdict lines:

```
GATE: RESULT=PASS exit=0
```

Counts, not just the exit code:

- pytest — `TOTAL collected=15467 passed=15465 skipped=2 failed=0 (floor: 13732 = sum of 27 per-target floors)`
- node — `TOTAL suites=4 files=34 tests=1149 pass=1149 fail=0 skipped=0 (floor: 1126)`

### The module under change

`nix develop . --command python3 -m pytest scripts/tests/test_skill_descriptions.py -q` → **21 passed**, unchanged from before this PR. Verified at the exact pushed sha (`50252de`), not only at the pre-amend one.

### 🔴 Mutation control — the re-pin is load-bearing, not cosmetic

In a disposable clone, the *only* mutation applied was reverting the ceiling to its old value (`LISTING_TOTAL_CEILING_CHARS = 12_929` → `13_741`), under `PYTHONDONTWRITEBYTECODE=1`. The suite went **red** with this module's own message:

```
FAILED test_skill_descriptions.py::test_control_the_total_ratchet_can_go_red
AssertionError: the ceiling has 1062 chars of headroom, enough to absorb a whole
352-char entry unnoticed. Re-tighten LISTING_TOTAL_CEILING_CHARS to the current
measurement plus less than one average entry.
```

`1 failed, 20 passed`. So the gate genuinely refuses a ceiling left slack by a cut — shipping the deletions *without* the re-pin would not have merged, which is exactly the property that constant exists to have. The clone was deleted afterwards.

### Caveat on scope

`strict` is false on branch protection, so a green check here is a claim about **this branch**, not about the tree the merge creates. Gating the merged tree is still a by-hand step before merging.

ZacxDev:
## Re-verification after the audit fix — `bd45ef4b`

An audit fix resets the verification gate, so this is a full re-run, not a delta. ⚠ The earlier comment's verdict was for `50252de` and is **superseded**.

Run in a fresh `git clone --no-hardlinks` with `origin` **removed**, `--git-common-dir` asserted inside this agent's own scratch dir, HEAD confirmed `== bd45ef4b` (the pushed sha).

```
nix develop . --command bash scripts/gate.sh
```

**exit status 0**, agreeing with both runners' own verdict lines:

```
GATE: RESULT=PASS exit=0
```

Counts, not just the status:

- pytest — `TOTAL collected=15467 passed=15465 skipped=2 failed=0 (floor: 13732 = sum of 27 per-target floors)`
- node — `TOTAL suites=4 files=34 tests=1149 pass=1149 fail=0 skipped=0 (floor: 1126)`

`nix develop . --command python3 -m pytest scripts/tests/test_skill_descriptions.py -q` → **21 passed** at `bd45ef4b`.

### The numbers behind the rewritten docstring were recomputed, not copied

Rather than transcribe the review's arithmetic, every figure was re-derived from `_entries()` against the binary's own formula:

```
entries                : 36
gate measure           : 12,679
real charge            : 12,858
undercount             : 179   (check 5n-1 = 179)
any desc over 1536 cap : False
non-ASCII across all entries: —…

  claude-3.x..4.6  ctx=  200,000  budget= 8,000  ratio=1.61x  OVER
  claude-3.x..4.6  ctx=1,000,000  budget=40,000  ratio=0.32x  fits
  claude-opus-5+   ctx=  200,000  budget= 6,000  ratio=2.14x  OVER
  claude-opus-5+   ctx=1,000,000  budget=30,000  ratio=0.43x  fits
```

🔴 **One correction back to the review:** the undercount is **179**, not 194. `194 = 5×39 − 1` is the *pre-deletion* 39-entry figure; at 36 entries it is `5×36 − 1 = 179`. The review's own real-charge number (12,858 = 12,679 + 179) already implied 179, so the two claims were inconsistent — the docstring uses 179 and states the general form `5n − 1` so it cannot go stale against the entry count again.

The `—…` line is the measured evidence for the `.length` vs `Bun.stringWidth` caveat: those are the *only* non-ASCII characters across all 36 descriptions today.

### Binary probes were run with controls both ways

`bin/claude` is a 20 KB stub. Measured, so the docstring's warning is not hearsay:

| key | `bin/claude` (stub) | `.claude-wrapped` |
|---|---|---|
| `skillListingBudgetFraction` | **0** | 4 |
| `enabledPlugins` (positive control) | — | 30 |

A grep of the stub returns a confident zero for a key that demonstrably exists.

### Unchanged on purpose

`LISTING_TOTAL_CEILING_CHARS` (12_929) and `listing_total_chars` are **untouched** by this fix. The 179-char undercount is the conservative direction, and re-pointing the ratchet at the real formula belongs with the tiering work, not with a retirement. The mutation control from the earlier comment therefore still holds.

### Scope caveat, unchanged

`strict` is false on branch protection, so this green is a claim about **this branch**, not the tree the merge creates.

PR #790, opened 2026-08-24
refactor(skills): /handoff step 4 was 56% of a skill about handoffs — the index protocol now has its own home

ZacxDev:
refactor(skills): /handoff step 4 was 56% of a skill about handoffs — the index protocol now has its own home

`/handoff` step 4 was ~26 KB of subsystem-index protocol: its own tool
(`scripts/lib/subsystem_touch.py`), its own store outside every repo, its own
reference doc, and 72 sentences pinned verbatim. It lived inside a skill about
writing handoffs and was paid in full every time `/handoff` ran — which is at
the END of a session, when the window is already tight.

It moves WHOLE to `claude/skills/subsystem-index/`, which `/handoff` invokes.

    claude/skills/handoff/SKILL.md          47,143 -> 21,729 B  (-54%)
    claude/skills/subsystem-index/SKILL.md            27,873 B  (new)
    reference/index-write.md                     moved with it

🔴 THE COMBINED TOTAL BARELY MOVED (49,602 vs 47,143) AND THAT IS NOT THE POINT.
A skill body costs its bytes when it is INVOKED. The index protocol is now paid
only on runs that reach a write, instead of on every handoff. Reading this as a
byte saving misses what changed.

THE PIN SPLIT WAS CLEAN, AND MEASURED BEFORE MOVING ANYTHING: 72 of the 73
pinned sentences were inside step 4, so they moved with the protocol and now
assert against `subsystem-index/SKILL.md`. The one that stayed is about
/handoff's own step order. No pinned sentence was reworded to make the move fit;
`git diff` of the extracted body is a dedent and nothing else.

🔴 A SEAM GUARD, BECAUSE THE 72 PINS WOULD ALL PASS IF THE HALVES CAME APART.
They assert the protocol EXISTS; none asserts that anyone RUNS it. If /handoff
stops naming the skill, the step silently stops happening and the suite stays
green — `claude/RULES.md`'s "verified in isolation is the new vacuous green …
the defect lives in the SEAM nobody owns". So
`test_handoff_still_routes_to_the_index_skill` asserts /handoff still names
`subsystem-index` AND still passes `--exclude` for its own doc (caller knowledge
the index skill cannot supply).

🔴 THE LISTING BUDGET WAS THE BINDING CONSTRAINT, and the new entry is 176 chars
against 74 now free. No eviction was taken, on a principle rather than a squeeze:
a skill invoked BY other skills needs IDENTITY, not routing keywords. Nothing
keyword-routes to `subsystem-index`, so the trigger phrases that make other
descriptions expensive would be dead weight in it. ⚠ That margin is nearly gone;
the next skill genuinely does need an eviction.

CEILING RATCHETED DOWN, 48,600 -> 24,000. The module has said since it was
written that down is the intended direction; this is the first time the lever was
structural rather than editorial, and it is lever 1 of that file's own playbook
("move a coherent block to a home of its own") demonstrated rather than quoted.

🔴 A COLD READ CAUGHT WHAT NO TEST COULD. With the suite green, step 5 still read
"step 4's *index* write is gated" — untrue since 2026-08-15 when that prompt was
retired, and doubly untrue now that step 4 is a pointer. Corrected, with the
correction stated in place so the next reader does not restore it.

Also corrected: the kickoff-ordering sentence named "step 4's confirm gate",
a gate that no longer exists. Prose and its pin changed together.

Verified: `gate.sh --tier both --set hermetic` under `nix develop` on this tree —
GATE: RESULT=PASS exit=0; pytest collected=15507 passed=15505 skipped=2 failed=0,
27 targets above floor, no target reporting FAIL; node suites=4 tests=1179
pass=1179 fail=0.

⚠ NOT VERIFIED HERE: that `/handoff` still executes correctly end to end. The
pins prove the protocol is present and routed to; only a real run proves an agent
follows it across the skill boundary. That run is the next thing after this
merges.


ZacxDev:
## Corrections to the PR description, and what a blind audit found (`41477d52`)

Correcting here rather than editing the body, since a reviewer may have read the wrong version.

### Numbers I got wrong

| claim in the body | actual |
|---|---|
| "72 of the 73 pinned sentences were inside step 4" | the list went **78 → 77**; all 77 now resolve in the new skill, one entry was retired. The *shape* of the claim holds; the numbers do not. |
| `47,143 → 21,729` / new skill `27,873` | stale — written before the last cold-read edits. Current: **22,189** and **28,510**. |
| "74 chars free" in the listing | **68** now (the reworded description is 182, up from 176). |

The one claim that held exactly: **"the extracted body is a dedent and nothing else."** The auditor reconstructed step 4 from the base, dedented it identically, and found the diff to be **4 lines — 3 required path rewrites and one deliberate `this session` → `a session`**. No sentence dropped, reordered or silently reworded across 26 KB.

### What it found instead — the same class, from the other direction

Not the prose. **Guards whose scope narrowed because the file underneath them moved, while their own text stayed wide.**

**🔴 A safety retraction went unguarded, proven by mutation.** `test_the_retracted_claim_is_NOT_still_asserted` keeps a measured-false claim dead: that the store append *"fails loudly rather than clobbering"* a concurrent write. It does not — a whole-file retype silently loses the other session's bullet, which is the entire reason the protocol mandates `Edit` on an anchor. The docstring said "absent from **BOTH** files"; post-extraction the tuple checked the two files the prose had **left**. The auditor appended the retracted sentence to `subsystem-index/SKILL.md` and it **survived a full green suite**. Fixed, and I re-ran that exact mutant: it now dies on that test as the only failure in 1,315.

**🔴 77 pins targeted a file nothing proved ships.** `test_the_pinned_docs_are_the_DEPLOYED_ones` — *"pinning a file that does not ship would be a vacuous green"* — didn't include the file now carrying every pin in its class. It now checks it, and checks it is **git-tracked**: the flake source contains only tracked files, so an untracked SKILL.md deploys as an absence while every test passes against the working copy.

**🔴 A false claim in listing surface, paid every session.** I wrote *"Called BY /handoff and /analyze-service"*. `/analyze-service` does **not** follow this skill — it follows `analyze-service/reference/write-back.md`, and the two **materially conflict**: that one gates the append behind an `append this to the index? (y/N)` this file declares retired, and uses `Write` where this file mandates `Edit`. Two documents each claiming to be *the* protocol. Now stated as **open divergence**, with the consequence named — the template hardcodes `created_by: handoff`, correct only while there is one caller. Reconciling them is follow-up, deliberately not done here.

**🔴 The seam had no fallback.** Step 4 named a skill and gave no path — and this skill's description says *"rarely run directly"*, which `CLAUDE.md` documents as exactly what listing-overflow eviction drops **first**, with 68 chars of headroom left. It now names the deployed path too, which `Read` resolves regardless.

**🔴 My own size-gate playbook was lying to the next maintainer.** On a handoff breach it rendered *"the skill's prose is pinned verbatim: 77 in HANDOFF_SENTENCES…"* — all 77 of which now target the **other** file. Someone trimming `handoff/SKILL.md` under budget pressure was told 77 of its sentences were protected; **zero** are. Also there: the extraction is the playbook's lever **2**, not lever 1; and the margin was quoted ceiling-relative while presented as the working margin — the `MIND THE BASE` mistake `test_rules_size.py` already records three times.

Plus dangling `step 5` / "this step" / "second writer" references that only parsed as step 4, the reference doc still titled *"`/handoff` step 4"*, and test failure messages naming a file their assertion no longer reads.

### Verification

`gate.sh --tier both --set hermetic` under `nix develop` on this tree: `GATE: RESULT=PASS exit=0`, pytest `collected=15507 passed=15505 skipped=2 failed=0` with no target reporting FAIL, node `tests=1179 pass=1179 fail=0`.

The auditor separately test-merged this with **#787** in an isolated copy — no conflicts, 1,322 passed — and positive-controlled `test_doc_path_rot.py` by planting the old path back, confirming the path-rewrite half of the move is mechanically guarded.

⚠ **Still not verified: that an agent actually follows the pointer across the skill boundary.** The pins prove the protocol is present and named; only a real `/handoff` run proves it is followed. That run is the next thing after this merges.

PR #797, opened 2026-08-24
fix(browser-bridge): bound the screenshot fast path — a hung captureVisibleTab never reached the CDP fallthrough

ZacxDev:
## The defect

`screenshot` takes a fast path (`chrome.tabs.captureVisibleTab`) when the tab is its window's active tab, guarded by a `catch` whose comment promises *"any failure there falls through to the CDP path"*.

That was only ever true of a **rejection**. `captureWithRetry` awaits the call, so a promise that **never settles** never reaches the catch — the op burned the whole `EXEC_OP_BUDGET_MS` (18 s) and returned `op_timeout:screenshot`, on exactly the tab the CDP path captures fine.

## Measurement (2026-08-24, app-capture arc)

An active tab on a non-visible workspace:

| path | n | result |
|---|---|---|
| fast (`captureVisibleTab`) | 3 | `op_timeout` **pinned at 18.07–18.11 s** — the ceiling, not a natural error latency |
| CDP (same tab, backgrounded) | 3 | **ok**, 381–3084 ms, identical **1709×1314** geometry |

One arm *did* return via `captureVisibleTab` at **17.97 s** — so it is far too slow, not permanently stuck.

🔴 **Occlusion is not the discriminator**, which is what sent the earlier diagnosis wrong: the hang reproduces with the window merely on a non-visible workspace and **nothing drawn on top** — the same state that had succeeded 3/3 at ~292 ms twenty minutes earlier.

## The fix

Bound the fast path at `FAST_CAPTURE_BUDGET_MS` (5 s) and let the existing catch do its job. The bound turns "never settles" into a rejection, which is the only shape that catch can see.

5 s sits **above** the legitimate retry ladder (a real quota hit spends ~2.4 s of backoff across `CAPTURE_MAX_ATTEMPTS`, so a quota storm still *retries* rather than being pushed onto CDP) and far **below** the 18 s op ceiling. Falling through costs a debugger banner and returns the same image — the "CDP changes the image" caveat is about `--fullpage`, a different clip.

## Tests

- a **hung** `captureVisibleTab` must fall through to CDP;
- an **attribution control** that a *healthy* one is still used — without it the test would pass just as well with the fast path deleted outright.

**Red at base**, verified by reverting the bound: the test does not fail with an assertion, it **hangs** — killed at 30 s with `pass 0 / cancelled 1` and *"Promise resolution is still pending"*. That is the production symptom itself, so the test reproduces the defect rather than restating the patch. Green at HEAD.

Full suite: **500 mjs + 783 python, all passing**. `build_id.js` regenerated because the extension source changed (its own gate caught that and named the command).

## Not verified live

🔴 The deployed extension at `~/.local/share/browser-bridge-ext` still runs the old build. This needs a deploy + Brave reload, then `browser ping` to confirm buildMarker `aefab9d551a8d058` is the **running** code. Reproducing the hang on demand is intermittent, so live confirmation is opportunistic.

Record: `<datapacket-talos>/claudedocs/app-capture-occlusion-refutation-2026-08-24.md`

### DIFF
--- a/scripts/browser-bridge/README.md
+++ b/scripts/browser-bridge/README.md
@@ -1287,7 +1287,26 @@ Two changes close it:
    `OPS[cmd.op](cmd)` against `EXEC_OP_BUDGET_MS` (18s). `frames`/`screenshot`
    are deliberately **not** patched individually — one rule, one place, or the bug
    regenerates at the next op added (`targetTab()` alone is on the path of every
-   op). Every *other* await in the loop body is bounded too: the poll fetch
+   op).
+
+   🔴 **ONE DOCUMENTED EXCEPTION, and it is not a weakening of that rule.**
+   `screenshot`'s **fast path** carries its own `FAST_CAPTURE_BUDGET_MS` bound.
+   The choke point above can only *end* an op; it cannot make a hung sub-step
+   FALL THROUGH to a working alternative. `chrome.tabs.captureVisibleTab` can
+   hang rather than reject, and the fast path's `catch` — which exists precisely
+   to fall through to the CDP path — can only see a rejection, so the op burned
+   all 18s and failed on a tab CDP captures fine. A bound that converts "never
+   settles" into a rejection is the only shape that restores the fallthrough.
+   The rule still holds for *terminating* an op; it simply cannot express
+   "recover from this step". ⚠ Its cost is a real constraint, pinned by a test in
+   `tests/cdp_protocol.test.mjs`: whatever the fast path spends comes out of the
+   CDP attach-hang margin below, so `FAST_CAPTURE_BUDGET_MS + 16s <= 18s`.
+
+   ⚠ The predicted regeneration is already present: `open`'s
+   `await chrome.tabs.get(cmd.reuseTabId)` sits in a `catch` that promises a
+   fall-through it cannot deliver if that call hangs. Same class, not yet fixed.
+
+   Every *other* await in the loop body is bounded too: the poll fetch
    (`POLL_BUDGET_MS` 40s), the result POST (`RESULT_BUDGET_MS` 10s), and the
    `chrome.storage.local` reads in `config()`/`clearSuperseded()`
    (`STORAGE_BUDGET_MS` 5s) — the last of which runs on *every* healthy iteration,
--- a/scripts/browser-bridge/extension/build_id.js
+++ b/scripts/browser-bridge/extension/build_id.js
@@ -24,4 +24,4 @@
 //   * a hash computed in the worker over its own source — same disk read
 // The marker must be a LITERAL in a module the worker imported, so that it was
 // frozen into the loaded module graph at load time and travels with the code.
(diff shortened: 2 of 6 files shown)


ZacxDev:
## Correction to the PR body above — its coverage claim was stronger than the tests delivered

An adversarial audit found the runtime change sound but the **evidence section of this PR body misleading**. Correcting it here rather than silently editing it, since it may already have been read.

**1. "Tests include an attribution control" understated what was missing.** Both new tests inject `fastCaptureMs` through `loopTiming()`, so the **production constant was read by no test at all**. Mutating `FAST_CAPTURE_BUDGET_MS` from `5000` to `600000` — ten minutes, past the 18 s op ceiling, reinstating the exact defect this PR fixes — **survived a fully green 500-test suite**. The body spent its justification on a number nothing could see.

**2. "Red at base" was true but delivered in the one shape a green-suite check cannot read.** With the bound reverted the file did not fail — it **wedged**: no summary, `fail` = 0, and 12 of 23 tests never ran. A regression would have hung CI rather than failing it, and any gate counting failures would have scored it clean.

**3. The value `5000` was wrong, not merely untested.** Whatever the fast path spends comes out of the CDP attach-hang margin: `5000 + 8000 attach + 8000 awaited detach = 21000 > 18000`. A post-fallthrough attach hang would have reported `op_timeout:screenshot` instead of `cdp_timeout:attach` — destroying the phase attribution `fast_capture_timeout` exists to give, and violating `protocol.js`'s own written warning against leaving CDP under 16 s. **Now 1500** (`17500 <= 18000`), pinned by a relationship test. Honest cost: **500 ms of headroom, down from 2000 ms**.

**4. One claim in the body is reversed.** It said a quota storm "still retries rather than being pushed onto CDP". At any bound satisfying the invariant above, a genuine quota storm (~6.5 s worst case) now **does** fall through to CDP. That is the right trade — CDP has no `captureVisibleTab` quota — but it is the opposite of what was written.

Fixes are in `db7fd643`, verified by mutation rather than by assertion:

|---|---|---|
| `1500 → 600000` (the survivor) | **fail=2** | both budget tests |
| `1500 → 1000` (under the healthy-capture floor) | **fail=1** | the relationship test's own floor clause |
| bound removed | **fail=1, cancelled=0, all 501 tests ran** | the hang test, at 1001 ms |
| control (no mutation) | **501/501 green** | — |

Note the fix round needed **two** attempts on finding 2: `{ timeout: 2000 }` alone still left `fail=0 / cancelled=2` with 10 tests unrun, because a `finally` after a hung `await` never runs — the stub leaked into the following tests and hung the control too. Cleanup moved to `t.after()`, and the call is now raced inside the test so a regression counts as a `fail` rather than a `cancelled`.

**Still not verified live**, unchanged: the deployed extension at `~/.local/share/browser-bridge-ext` runs the old build. Needs `home-manager switch`, a manual ↻ in `chrome://extensions` (Brave does not hot-reload unpacked extensions), then `browser ping` to confirm buildMarker `ef984e84cbb83651` is the running code.

PR #804, opened 2026-08-24
feat(discord-embed-ext): land the enlarge+lightbox half — the sidecar save button was dead on arrival, and a real browser found a 7th bug

ZacxDev:
Lands the `scripts/discord-embed-ext/` work that had been sitting **untracked** in the shared base clone since 2026-08-23 23:13, orphaned by a dead session — as the **enlarge + lightbox half only**, with the dead sidecar integration removed, seven real defects fixed, and a nix deploy path it never had.

The original 13 files are preserved verbatim on `rescue/discord-embed-ext` (`42563d57`). Nothing was discarded.

> **The correction history is in this thread and is part of the record.** Several claims in earlier versions of this description were wrong and were corrected publicly rather than silently edited — including a "six regression tests" count that was five. If you read only this body you get the destination, not the route.

## Why not merge it as it was found

**`save_button.js` was dead on arrival, not unfinished.** It called dl-router's sidecar with no `Authorization` header; dl-router requires a bearer token on every endpoint. Probed live:

```
GET /healthz  no token   -> 401
GET /healthz  with token -> 200
```

So `checkSidecar()` always resolved `{available:false}`, the observer that mounts the button never started, and **the button never appeared**. Even authed, `POST /match` only *classifies* a download — it saves nothing — yet the handler set the label to `"Saved!"` on `resp.ok`. Removed rather than repaired: filing media into the library is dl-router's job, behind its own extension and token. Also dropped: `extractChannelId()` (exported, tested, called by nothing) and the **0-byte** `service_worker.js` the manifest declared.

The manifest now has zero `permissions` **and** no `host_permissions` — the extension reaches nothing.

## The defect only the real client could show

The media pattern matched the CDN **host**, which also serves avatars, icons, emojis, stickers, banners, clan badges and 48×48 `/media/` decorations. Measured against two real logged-in channels:

| | matched | would enlarge |
|---|---|---|
| host only | **59 of 60** ``/`<video>` | **10 user avatars** |
| `/attachments/` or `/external/` | 0 in that channel | 0 |

Breakdown of the 59: avatars 24, icons 35, **attachments 0**. A bare zero proves nothing, so the positive control: in a second channel the new pattern matched **1** — a real attachment in a container capped at `max-height: 350px`.

Every URL in `tests/fake_discord_dom.mjs` is already an attachment, so the fixture was structurally blind to this.

## Verified in a real browser

Production path (autostart → `scan()` → a real bubbling click) against a container using a **hashed** class (`imageWrapper__74e4d`):

| check | result |
|---|---|
| attachment enlarged | `data-dee-enlarged="1"` |
| avatar in the same row | `null` — untouched |
| 400px cap removed | computed `max-width: none` |
| lightbox opened by a real click | `true` |
| overlay | `position: fixed`, `z-index: 2147483647`, covers viewport |
| Escape closes | `true` |

## Deploy path

Nothing in `nix/` referenced this directory, so merging it before would have shipped an artifact **no host loads**. `home.activation.discordEmbedExtension` deploys to `~/.local/share/discord-embed-ext` as real files (`cp -rL`), outside the git tree so a checkout cannot swap code under a running Brave. Two failure paths that could abort a `home-manager switch` were found and fixed, each with a red control:

* `chmod -R` followed a symlinked destination and rewrote the **target** tree's modes — measured `555→755` / `444→644` pre-fix, unchanged post-fix;
* the temp sweep spared same-PID wreckage, so `mv -T` later failed `Directory not empty` — `rc=1` pre-fix, `rc=0` post-fix.

Registering with Brave stays a one-time manual `Load unpacked`; nix cannot do it.

## Gates

* **Merged-tree** (`origin/main` `10ad288a` + this branch, clean merge): `TOTAL suites=5 files=37 tests=1320 pass=1320 fail=0`, `RESULT: PASS`.
* **Gated nix-sandbox tier** (`nix build .#checks.x86_64-linux.{nodetests,pytests}`): **exit 0** on the merge commit `f1cd8c58` and on every fix tree since.
* This suite: **118 tests, floor 113.**
* **86 semantic mutants, 84 killed**, 0 unapplied, positive control dies. The 2 survivors are one equivalence class (the `nodeType` filter on both observer arms), verified equivalent by an independent auditor rather than asserted here.

🔴 That mutant total is a claim about the 86 mutants **somebody constructed**, never about the suite. Nine audit rounds ran; the first eight each found something the previous had missed, including a fixture that lowercased `tagName` (voiding seven `.toLowerCase()` guards at once, each deletion making the extension inert in Brave) and two content-script entry points whose module-scope calls could be deleted with the suite green. The README carries the full numbered list. The ninth round found nothing new.

## Honest gaps

* The extension has **never been registered in Brave via `Load unpacked`** — that step is manual, so "installs cleanly as an unpacked extension" is untested. Everything above was verified by injecting the shipped sources into a real page.
* The message-boundary rule (class containing `message`, or id starting `chat-messages-`) is a **heuristic** and the most likely thing to rot. It degrades to "navigation does nothing", never a crash.
* ⚠ **A concurrent, independent rebuild of this same extension** was found uncommitted in the shared base clone (no branch, no PR) and preserved on `rescue/discord-embed-ext-concurrent`. It re-introduces the save button measured dead above and a conflicting `|3|50` SUITES line. Whoever lands that will need to reconcile it with this PR by hand.


ZacxDev:
## Gate verdicts — both tiers, named

**Sandbox tier** (`nix build .#checks.x86_64-linux.*`) — the tier Tekton runs and the merge is gated on. Base sha `7491ce10`, branched off `origin/main` `2f6928df`:

```
SANDBOX_NODETESTS_EXIT=0
SANDBOX_PYTESTS_EXIT=0
```

**Dev-host tier** (`scripts/gate.sh --tier both --set hermetic`):

```
PASS  scripts/discord-embed-ext/tests  (files=2 tests=60 pass=60 skipped=0 floor=57)
TOTAL suites=5 files=37 tests=1239 pass=1239 fail=0   RESULT: PASS (exit=0)   <- node
all 27 pytest targets PASS, collected=15809 passed=15806 skipped=2, 0 FAILED lines
```

That dev-host run nonetheless reported `GATE: RESULT=FAIL`, and the cause was **mine, not the diff**: GUARD 10 fingerprints the shared `.git/config` around every target, and I ran `git push -u origin feat/discord-embed-enlarge` while the gate was in flight. The key it flagged is exactly the one `push -u` writes:

```
~ branch.feat/discord-embed-enlarge.merge
```

The guard's own message names `push -u` as a cause. No test failed — every per-target line is PASS and the log contains zero `FAILED` lines. The `failed=1` in the TOTAL is the guard's synthetic failure, not a test.

⚠ `strict` is false on branch protection, so a green check here is a claim about **this branch**, not about the tree the merge creates. Gating the merged tree is still a manual step.

ZacxDev:
## Correction — the "six regressions, each watched RED" claim is wrong. It is five.

An adversarial audit re-introduced each of the six original defects **individually** against the shipped suite. Five are killed. **One survives a fully green 60/60 run:**

> #5 — "navigating never re-applied the transform"

`navigate()` with `applyTransform()` deleted is **60/60 green**. Nothing in the suite ever asserts `mediaClone.style.transform`. Related mutants that also survive: swapping the `scale`/`translate` order, `setZoom()` without `applyTransform()`, and `panX += dx / zoom` → `dx * zoom`.

**How I got it wrong, since the mechanism matters more than the count:** my RED control carried all six defects *at once*. The test `REGRESSION: navigating resets the zoom AND the label` duly failed — but on the **label** half. I credited the transform half from the same red line without isolating it. That is exactly the "isolate the mutation" failure: a mutant removed together with its enclosing condition proves nothing about the part you meant to pin.

The claim appears in three places. The one that matters is **checked-in prose** at `scripts/run-node-tests.sh:231-236`, which a future maintainer would read as coverage that does not exist.

A second prose claim is also unpinned: `extension/lightbox.js:40-51` and the README both assert that matching the class **substring** is what survives Discord's hashed class names. Mutating it to `cls === "message"` **survives** — no fixture uses a hashed class. The `id.indexOf("chat-messages-") === 0` branch, which the comment says the real client actually uses, is exercised by nothing.

The audit also found real defects I am fixing rather than re-describing here, including a percentage `max-width` being read as pixels (`parseFloat("100%") === 100` ≤ 500), two `nix` activation paths that can abort a `home-manager switch`, and a `MutationObserver` — the only production path after `document_idle` — with zero coverage.

Fixes and a delta re-audit to follow in this thread. **Do not merge on the strength of the coverage this PR currently advertises.**

ZacxDev:
## Round 9 (final): no new findings — the loop is closed

Nine adversarial rounds ran. The ninth returned **no 🔴, no 🟡, no new findings**, with a clean SURVIVED-at-`09e597ab` / KILLED-at-`8635f168` matrix across 8 isolated mutants for the three code-affecting items, and confirmed `extension/` is byte-identical across that delta (tree OID `bbe9c6c2`, existence proven, so not an absent-operand false "same"). A clean round was the stop condition, so I stopped.

### Gates, with the tier named

| tier | tree | result |
|---|---|---|
| merged tree, node runner | `origin/main` `10ad288a` + `8635f168` (clean merge) | 1320 tests, 0 fail, `RESULT: PASS` |
| gated nix sandbox | merge commit `f1cd8c58` | nodetests **0**, pytests **0** |
| gated nix sandbox | `20d34ffc`, `09e597ab` | nodetests **0**, pytests **0** |

`strict` is false on branch protection, so a green check on this branch is not a claim about the tree the merge creates — that is why the merged tree is gated separately, twice, since `main` moved under this PR twice.

### What nine rounds actually bought

The first eight each found something the previous missed, and the pattern is worth more than the fixes:

- **A fixture that normalises away what the code normalises is not a test of that code.** The fake DOM lowercased `tagName`; a real document reports it uppercase. That voided **seven** `.toLowerCase()` guards at once — each deletable with a fully green suite, and each deletion making the extension **completely inert in Brave**. Three earlier rounds had already fixed three other instances of this same class in that same file.
- **A defect that is a SHAPE, not a site.** The module-scope entry-point gap was fixed for one content script and left open in the other; the `didDrag` latch was pinned on its SET and not its RELEASE; the backdrop's `click` arm was pinned while `wheel` and `mousedown` could be deleted silently. Each fix landed one level short of its own class.
- **A count maintained in parallel with the thing it counts will drift.** This PR's recurring defect was never a logic bug — it was claims stated wider than what was measured: "six regression tests" (five), "25 mutants, 24 killed" (eight more survivors), "four voided guards" (seven), a floor comment whose own arithmetic contradicted its result, and two files disagreeing on one measurement. The README's list is now numbered and tells the reader to count it rather than trust a total, because the count regrew wrong **twice inside the paragraph warning about it**.

I am not merging this. Two things are yours to decide: whether to land it at all given the concurrent duplicate rebuild preserved on `rescue/discord-embed-ext-concurrent`, and whether the manual `Load unpacked` step is worth doing before or after.

ZacxDev:
## ⚠ A second, independent implementation of this feature is in flight — read this before merging either

An unrelated agent is building the same extension, in the same directory, right now. It has **no branch and no PR of its own**; it exists only in the shared clone's working tree, so it is invisible from GitHub unless you know to look.

Preserved (not merged, not modified, not folded into this PR) on **`rescue/discord-embed-ext-concurrent`**:

| snapshot | when | what changed |
|---|---|---|
| `a12550b2` | 18:06–18:12 | first sighting: 13 files, `tests/fixtures/` layout, `save_button.js` present |
| `14be8300` | 19:27–19:30 | still active — 6 of 13 files changed, incl. real icons replacing the 1×1 placeholders |

It is **not** the untracked orphan this PR rescued (that is `rescue/discord-embed-ext`, `42563d57`, a different tree).

### Two measurements that apply to whichever implementation lands

Neither is an opinion about the other approach; both were measured and both are cheap to re-check.

**1. The dl-router save button cannot work as written.** It posts to the sidecar with no `Authorization` header; dl-router calls `_guard()` first in both `do_GET` and `do_POST`, and `_auth_ok()` requires `Bearer` — `/healthz` included. Probed against the live sidecar:

```
GET /healthz  no token   -> 401
GET /healthz  with token -> 200
```

So `checkSidecar()` resolves `{available:false}`, the observer that mounts the button never starts, and the button never appears. Separately, `POST /match` is classify-only — it returns a suggested directory and saves nothing — so a handler that sets "Saved!" on `resp.ok` reports a save that did not happen. The in-flight copy still carries `save_button.js` and `save_button.test.mjs`.

**2. A host-only media pattern enlarges avatars.** `cdn.discordapp.com` also serves avatars, server icons, emojis, stickers, banners, role icons and 48×48 `/media/` decorations. Measured against two real logged-in channels:

| pattern | matched | would enlarge |
|---|---|---|
| host only | **59 of 60** ``/`<video>` | **10 user avatars** |
| `+ /attachments/` or `/external/` | 0 there; **1** in the second channel (a real attachment, container capped `max-height: 350px`) | 0 |

A synthetic fixture cannot show this if every URL in it is already an attachment.

### The mechanical conflict

Both edit `SUITES` in `scripts/run-node-tests.sh`, and that table is pinned two-way — a discovered suite must have an entry and every entry must be discovered. The in-flight copy writes `"scripts/discord-embed-ext/tests|3|50"` (3 files, floor 50); this PR writes `|2|113` (2 files, 118 tests). **Only one can be right**, and whichever lands second will need to reconcile by hand.

### Why this is a comment and not a message

The writer is almost certainly an `opencode` agent, not a Claude session — no Claude transcript on this host references those files, and there are three `opencode` processes in the base clone. There is no peer I can address, so this comment plus the two rescue commits are the durable record.

PR #811, opened 2026-08-24
docs(tekton skill): the trigger count omitted the only GATING trigger — and 255 is not "no test failed"

ZacxDev:
Two corrections to `claude/skills/tekton/SKILL.md`, both measured against the live homelab cluster on 2026-08-24. Found while running `/analyze-service` on homelab tekton.

## 1. "Six triggers" → seven, and the missing one is the only gate

Live `el-github-listener` carries **seven** triggers, not six. The absent one is `devrc-ci-pr`, added ~3 days ago.

That is the worst one to omit: `innovation-upstream/devrc` requires `tekton/devrc-pytests` + `tekton/devrc-nodetests` with `enforce_admins: true`, whereas `ZacxDev/homelab-infra` and `ZacxDev/naida-ai` both return **403** on branch protection (`Upgrade to GitHub Pro…`) and so can only ever be detectors. The one line a reader consults to reason about what gates what was silent about the only thing that gates anything.

Now cross-links gotcha #9 and tells you to read the count off the CR:

```
kubectl -n tekton-ci get eventlistener github-listener -o jsonpath={.spec.triggers[*].name}
```

## 2. `exit 255` is a claim about the STEP, not about the tests

The skill documented 255 as *the* congestion signature with "it heals when the queue drains" as the only tell. That is true and slow, and it invites reading every red gate as load.

Classified all completed `devrc-ci` gate TaskRuns by which step died and with what code (rather than by the PipelineRun verdict):

| population | n | signature |
|---|---|---|
| killed steps | ~27 | 25× `pytests exit=255`, 1× `nodetests exit=255`, 1× `pytests exit=137` (OOMKilled) — **no test result reported** |
| genuine failures | ~27 | `verdict exit=1` — ~1 test in ~15,500 |

The genuine failures do **not** correlate cleanly with concurrency. So congestion is an **amplifier, not the cause**, and "the gate is just flaky under load" walks you past a real bug — which is exactly what was happening: see #810.

Discriminator added: **a step that printed `RESULT:` / `<leg> verdict=` failed a test; one that printed neither was killed.**

Also records the mechanism behind the kills — `step-pytests` requests **1 CPU / 2Gi** but limits at **4 CPU / 8Gi** against a node committed to ~**290% CPU / 270% memory** limits, so the scheduler packs on requests and the kernel kills on limits — with a note that closing it edits the `devrc-ci-gate` Task and moves **every** CI run, so it wants a heads-up first. Numbers are marked re-measure-dont-trust.

## Cost

Frontmatter untouched, so the always-on description and tier budgets are unaffected. The body costs zero until the skill fires.

## Verification

1016 passed across `test_skill_descriptions`, `test_skill_tiers`, `test_skill_audit`, `test_subsystem_touch`. Docs-only change to a skill body; no code path touched.

https://claude.ai/code/session_018NsRe3TW5xbSgzkqB7Catv

### DIFF
--- a/claude/skills/tekton/SKILL.md
+++ b/claude/skills/tekton/SKILL.md
@@ -69,16 +69,28 @@ debugging, changing or copying a specific pipeline.
    cause**, and "the gate is just flaky under load" will walk you straight past a real bug.
    Discriminator: a step that emitted `RESULT:` / `<leg> verdict=` **failed a test**; one that
    emitted neither was **killed**. 25 of the 27 kills had ≥4 gate TaskRuns overlapping.
-   🔴 **What KILLS a step is NOT established — and the obvious theory is MEASURED ABSENT.**
-   `step-pytests` requests **1 CPU / 2Gi** against **4 CPU / 8Gi** limits, and the node carries
-   a large limits overcommit, so "the scheduler packs on requests and the kernel kills on
-   limits" looks like the answer. It is not the answer for THIS step: homelab-infra **#393**
-   measured CFS throttle ratios across `tekton-ci` on 2026-08-24 and `step-pytests` came in at
-   **0.006 — not throttled**, with node saturation absent as well (16% CPU requested, ZERO
-   Pending pods, no OOMKilled). So the 255s are **undiagnosed**. Do NOT "fix" them by moving
-   devrc-ci's requests or limits on the strength of the request/limit gap alone — that is
-   reaching for the plausible theory instead of the discriminating control, and this exact
-   change has already been made once on a step where it was justified and would be inert here.
+   🔴 **THE 255s ARE SCHEDULER PREEMPTION, AND `gitops-validate` IS THE PREEMPTOR — CI
+   PREEMPTS CI.** Confirmed 2026-08-25T04:46:13Z by catching a burst live: five gate pods,
+   five explicit `Preempted` events, three preemptors, all five `pytests` steps terminating
+   `exit=255` within ONE SECOND having started minutes apart. The preemptors are priority-**0**
+   `gitops-validate` pods blocked on `0/4 nodes are available: 1 Insufficient cpu, 3 node(s)
+   didn't match Pod's node affinity/selector` — **both pipelines `nodeSelector`-pin to the SAME
+   single node** out of four, so gitops-validate cannot fit and the scheduler evicts the
+   `ci-bulk` (**-10000**) devrc gate pods to make room. One-directional by design: devrc-ci is
+   denied the reverse (`preemption: not eligible due to preemptionPolicy=Never`). It often does
+   not even buy the preemptor a slot — `not eligible due to a terminating pod on the nominated
+   node` recurs, so victims die and the preemptor stays Pending. Full evidence:
+   homelab-infra `claudedocs/handoff-devrc-ci-kills-are-simultaneous.md`.
+   🔴 **Two fixes that look right and are NOT.** (a) **Raising `ci-bulk` above 0** removes no
+   contention — it only inverts which pipeline is destroyed, and gitops-validate already has
+   the tightest timeout budget and the worst pod-start latency here. (b) **Moving devrc-ci's
+   requests/limits**: `step-pytests` requests 1 CPU/2Gi against 4 CPU/8Gi limits on an
+   overcommitted node, which looks like the answer and is measured absent — #393 put
+   `step-pytests` at **0.006, not throttled**, with no node saturation (16% CPU requested, zero
+   Pending, no OOMKilled). A per-pod limit also cannot kill five pods of different ages in the
+   same second. The fault is **two pipelines pinned to one node out of four**; note when
+   choosing that devrc-ci's pin buys the RWO `nix-store-cache` (~1m50s/run) while
+   gitops-validate is pinned for the same cache and is the one that cannot fit.
    🔴 **The real slowness mechanism, when it IS present: CFS QUOTA STARVATION, and AVERAGE CPU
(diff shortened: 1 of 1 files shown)


ZacxDev:
Pushed `24e512ae`, which **retracts a claim in the first commit** rather than adding to it.

The original second bullet asserted *"the kills are a RESOURCE MISMATCH, not bad luck"* and explained the `exit 255` kills via `step-pytests` requesting 1 CPU/2Gi against 4 CPU/8Gi limits on an overcommitted node. **That mechanism was never established** — it was the plausible theory written as fact, which is precisely the failure mode the rest of this file exists to warn about.

homelab-infra **#393** measured CFS throttle ratios across `tekton-ci` on the *same day*:

| step | throttle ratio | limit |
|---|---|---|
| `step-scripts-tests` | **1.00** | 1 CPU |
| `step-gitleaks` | 0.80 | 2 CPU |
| `step-clickup-mirror` | 0.74 | 1 CPU |
| **`step-pytests`** | **0.006 — not throttled** | 8 CPU |
| `step-kustomize` | 0.00 | 2 CPU |
| `step-render-diff` | 0.00 | 2 CPU |

Node saturation was absent too — 16% CPU requested, zero Pending pods cluster-wide, no OOMKilled. So the mechanism is measured *absent* for the step in question, and the 255s are **undiagnosed**. The text now says so, and explicitly warns against moving devrc-ci's requests or limits on the strength of the request/limit gap alone.

Two things added while correcting it, both from #393's measurements rather than from me:

- **CFS quota starvation and why average CPU hides it** — a step's own `limits.cpu` is enforced per 100ms period, so short-lived multi-threaded processes drain a 1-CPU quota in a few ms and stall for the rest, averaging ~0.32 cores (reads as *idle*) while throttled in 100% of periods. The `kustomize`/`render-diff` 0.00 rows on the same 2-CPU limit are what make this starvation of specific bursty steps rather than "limits too low".
- **Raising a limit half-works on a Go binary** pre-1.25, which sizes `GOMAXPROCS` from the host's cores rather than the cgroup quota.

Also corrected gotcha #2, found while checking the above: it claimed `~/workspace/homelab-talos` was "~100 commits behind". Measured today it was **zero** behind, while `~/workspace/homelab-infra` was **3** behind — the advice had inverted. Both directories are clones of the *same* repo (`ZacxDev/homelab-infra`) despite the name. Replaced the currency claim with the command that answers it.

All three corrections are the same shape as the trigger-count fix: a point-in-time measurement written as a standing fact, which nobody re-checks.

ZacxDev:
Pushed `b204d5e1`. The previous commit said *"the 255s are UNDIAGNOSED"* and warned against guessing — that is now settled, so the skill states the mechanism instead.

**Confirmed 2026-08-25T04:46:13Z** by catching a burst live, 7 minutes after arming a watcher: five gate pods, five explicit `Preempted` events, three distinct preemptors, all five `pytests` steps terminating `exit=255` within **one second** having started minutes apart.

**CI preempts CI.** The preemptors are priority-0 `gitops-validate` pods — the other Tekton pipeline — blocked on `0/4 nodes are available: 1 Insufficient cpu, 3 node(s) didn't match Pod's node affinity/selector`. Both pipelines `nodeSelector`-pin to the **same single node** out of four, so gitops-validate cannot fit and the scheduler evicts the `ci-bulk` (-10000) devrc gate pods to make room. One-directional by design — devrc-ci is denied the reverse (`preemptionPolicy=Never`).

The skill now records **both fixes that look right and are not**: raising `ci-bulk` above 0 (inverts the victim, removes no contention) and moving devrc-ci's requests/limits (measured absent by #393, and a per-pod limit cannot kill five pods of different ages in the same second). Full evidence lives in homelab-infra — ZacxDev/homelab-infra#397 — rather than being restated here.

That is three commits on this PR, and the second **retracts** a claim in the first. Reviewing the final state is more useful than the sequence.

ZacxDev:
`b3e2230d` — final correction. The preemption section described a **live** hazard; it is fixed.

homelab-infra **#396** (`c35c78cd`, "give gitops-validate its own nix cache and node") merged 2026-08-25 05:32Z — **46 minutes after** the 04:46:13Z burst capture that section was built on. Measured 11h post-merge:

| window | n | killed | passed |
|---|---|---|---|
| before #396 | 121 | **30 (24.8%)** | 74 (61.2%) |
| after #396 (≥06:00Z) | 9 | **0 (0.0%)** | **9 (100%)** |

Placement confirms the mechanism is gone rather than merely quiet — `devrc-ci` on `talos-xr6-r7p`, `gitops-validate` on `talos-uvh-gtj`, zero `Preempted` events retained.

⚠ **n=9 is suggestive, not conclusive** — ~2 kills would be expected in 9 runs at the old rate, so a clean 9 arises by luck about 1 in 12. The skill says so rather than presenting the zero as proof, and `ci-priority-classes.yaml` already asks for exactly this measurement over "the next few days".

The section is kept, explicitly labelled **history**, so the signature stays recognisable if it returns: five pods dying in one second with exit 255/137 and no test result is very easy to misread as a test bug — which is precisely what happened before it was diagnosed.

Also demoted from a live conclusion to a "redo this discrimination if it recurs": the binding predicate being CPU rather than pod count. It does **not** make right-sizing `auditloop-ci` (2.8×) or `clawgate-ci` (2.4×) urgent — that argument rested on contention #396 removed.

**This PR now has four commits, and each of the last three corrects the one before it** (resource-mismatch mechanism → retracted; preemption presented as novel → credited to prior art; preemption presented as live → superseded by #396). Review the final state of the file, not the sequence.

PR #814, opened 2026-08-25
fix(browser-bridge): bound `open`'s reuse probe — a hung tabs.get never reached the fall-through

ZacxDev:
## The defect — the predicted regeneration, now closed

When #797 bounded `screenshot`'s fast path, the README named this exact call as *"same class, not yet fixed"*:

```js
const existing = await chrome.tabs.get(cmd.reuseTabId);
```

It sits inside a `try` whose `catch` promises *"owned tab gone → open a fresh one below"*. **That promise is only ever true of a REJECTION.** The call is awaited, so a promise that never settles never reaches the catch: the fall-through never happens and the op dies at `EXEC_OP_BUDGET_MS` (18s) — a `cmd_timeout` on `open`, the op every session starts with.

**Fix:** race it against a new `REUSE_TAB_BUDGET_MS` (2000ms), wired through `loopTiming()` as `reuseTabMs` exactly as `fastCaptureMs` is, and let the existing catch do its job. A bound converts *"never settles"* into a rejection — the only shape that catch can act on.

## The arithmetic is NOT #797's, and copying it would have been a false justification

`FAST_CAPTURE_BUDGET_MS` is capped by the CDP attach-hang margin (8s attach + 8s awaited detach = 16s) **because its fallthrough is the CDP path**. `open` **never attaches the debugger** — its fallthrough is `chrome.tabs.create`, one more unbounded browser-process IPC terminated only by the op ceiling. I checked that margin and it does **not** constrain this constant. What does:

```
REUSE_TAB_BUDGET_MS + (what chrome.tabs.create needs) <= EXEC_OP_BUDGET_MS
2000 + 16000 = 18000
```

⚠️ **That 16000 is a numerical coincidence with the attach-hang cost, not the same constraint** — flagged explicitly in the comment, because a reader who spots it will otherwise re-derive the wrong reason and treat it as a hard floor.

**Why 2000ms, from the call rather than from taste.** `chrome.tabs.get` is the cheapest `chrome.*` call anything in that file wraps: one IPC reading tab metadata from an in-memory registry — no renderer, page, paint, network or disk, and (unlike `captureVisibleTab`) no quota and no retry ladder, so there is no legitimate slow band to clear. The anchor is an **ordering, and it is asserted rather than written down**: `STORAGE_BUDGET_MS` (5000) wraps the *disk-backed* `chrome.storage.local`, so a strictly cheaper call must not get a longer leash.

`LOOP_STALL_MS`'s worst-legitimate-iteration sum was checked and does **not** move: this bound is inside `execute()`, already carried as `EXEC_OP_BUDGET_MS` (18.00s). Sum stays 113.25s — same reason `FAST_CAPTURE_BUDGET_MS` adds no line there.

## A breadcrumb, because here the envelope cannot distinguish the two causes

On the screenshot fast path the outcomes differ (`via: cdp` vs `captureVisibleTab`). Here a hung probe and a genuinely-gone tab **both return a fresh `{tabId, url}` with no `reused` flag** — so only the crumb separates "the tab was gone" (correct, free) from "we timed out and just orphaned a LIVE tab" (the bound's known cost): `open_reuse_gone` vs `open_reuse_timeout`. Same single rolling slot, same honest caveat as #797's crumb: `execute()` overwrites it with `done` on success, so it survives only when the op also wedges later. **It is not a production counter.**

## Red at base — as a FAILURE, not a hang

`tests/service_worker.test.mjs`, bound reverted in a `cp -a` copy:

| tree | tests | pass | fail | cancelled | all ran? |
|---|---|---|---|---|---|
| bound reverted (base shape) | 28 | 26 | **2** | **0** | ✅ 28/28 |
| HEAD (this PR) | 28 | 28 | 0 | 0 | ✅ 28/28 |

Both failures are ordinary assertion failures at ~1000ms via the **in-test race**, not the `{ timeout: 2000 }` backstop:

```
✖ open reuse probe: a HUNG chrome.tabs.get falls through to a fresh tab (1001.31ms)
    Error: open did not settle within 1s: the reuse probe is unbounded, so a hung
           chrome.tabs.get never reaches the catch that opens a fresh tab
✖ open reuse probe: the breadcrumb distinguishes a HANG from a genuinely-gone tab (1000.48ms)
    Error: open did not settle within 1s: the reuse probe is unbounded, so no crumb is ever written
```

### 🔴 The first draft of the crumb test reintroduced the exact trap #797 documented

I only found this because I actually ran the red-at-base check after adding the crumb test. Its hang arm did a bare `await OPS.open(...)` with no race and no `{ timeout }` — which looked unrelated to hang-handling, since its *subject* is the breadcrumb. **Measured on the reverted tree, that version wedged the whole file:**

```
ℹ tests 18   ℹ pass 16   ℹ fail 1   ℹ cancelled 1     ← 10 of 28 tests NEVER RAN
```

`cancelled`, not `fail` — precisely the count-blind shape a fail-count gate reads as clean. Fixed with the same race + `{ timeout }` + `t.after` discipline, and the lesson is written into the test so the next person adding a stub that never settles sees why it is there. **Generalised: any test that drives a never-settling `chrome.*` is a hang test whether or not that is its subject.**

## Mutation battery

Every mutant run personally, full mjs suite, **no `-run` filter** (a filter that excludes the killing test reports SURVIVED), scored from summary **content** and `✖` names — never an exit code.

| # | mutant | tests/pass/fail/cancelled | killed by |
|---|---|---|---|
| — | **CONTROL** (unmutated) | `507/507/0/0` | — (clean) |
| M1a | `REUSE_TAB_BUDGET_MS` 2000 → **17000** | `507/506/1/0` | ordering assertion |
| M1b | `REUSE_TAB_BUDGET_MS` 2000 → **5000** (== `STORAGE_BUDGET_MS`) | `507/505/2/0` | ordering assertion **+** the wired-map distinctness assertion |
| M2 | `loopTiming()` wire → **literal 600000** | `507/506/1/0` | `loop budgets are wired to their constants` |
| M3 | **the bound removed** (base shape) | `507/505/2/0` | the hang regression test **+** the crumb test's hang arm |
| M4 | `STORAGE_BUDGET_MS` 5000 → **7000** (reachability control) | `507/506/1/0` | the surviving op-ceiling trigger |
| M5 | breadcrumb phases **collapsed to one spelling** | `507/506/1/0` | the crumb-distinguishes test |

Each died to a message naming its own cause:

```
M1a  reuse probe (17000ms) wraps chrome.tabs.get, which is strictly cheaper than the
     disk-backed chrome.storage.local bounded at 5000ms — it must not outrank it
M1b  (above) + "two budgets share a value — a swapped wire would be undetectable"
M2   loopTiming().reuseTabMs must read its protocol.js constant, or the value
     assertions elsewhere do not govern runtime
M3   open did not settle within 1s: the reuse probe is unbounded …
M4   STORAGE_BUDGET_MS (7000ms) has been raised far enough that assertion (1) no
     longer implies a healthy fallthrough share of the 18000ms op ceiling …
M5   a crumb that spells both causes the same way distinguishes nothing
```

### 🔴 A guard I wrote was DEAD, and the battery is what found it

I first wrote an op-ceiling-share assertion (`EXEC - REUSE >= 12000`) and **removed it**. The work-ordering assertion already caps the constant at 5000, while the share guard only binds above 6000 — strictly subsumed, so it **could never fire**: a guard that reads as coverage while providing none. Every mutant aimed at it died to the ordering check first. What replaced it is a *reachable* trigger that fires the day `STORAGE_BUDGET_MS` is raised past ~6000, i.e. when the ordering stops being the tighter of the two. **M4 exists solely to prove that replacement can fire** — otherwise it would be the same dead guard in a new spelling.

## Verification

- **JS:** 507 pass / 0 fail / 0 cancelled (was 502 at base; +5 tests).
- **Python:** 787 passed / 0 failed. *(The brief expected 783; I measured base and HEAD both collect **787**, so that delta is pre-existing — this PR adds no Python tests.)*
- **Build marker** regenerated → `44a1d80c1e9630a9`, `--check` OK.
- Rebased onto current `origin/main` (`324693fd`) and re-verified there, not against the stale base.

## What I could NOT verify

1. **🔴 No live measurement of a hung `chrome.tabs.get`.** This is the significant honesty gap versus #797, which rests on a live 3/3-vs-3/3 measurement of a real hang. I did not reproduce a hung `tabs.get` in a real Brave. What *is* established — from the code, independent of how often the browser exercises it — is that the await is unbounded and the catch is structurally blind to a hang. The 2000ms value is therefore reasoned from the call's nature and pinned by an ordering, **not** calibrated against an observed latency distribution the way 1500ms was.
2. **Nothing was verified live at all** — no Brave, no reloaded extension, no real `open` with a `reuseTabId`. All evidence is the unit suites.
3. **The known cost is reasoned, not measured:** falling through on a merely *slow* `tabs.get` opens a second tab while the first is still live and unowned — precisely the leak the reuse path exists to prevent. I have no data on how often a `tabs.get` lands in the 2–18s band, so I cannot quantify how often that trade is paid.
4. **The crumb's production value is bounded and untested live** — `execute()` clobbers the single slot with `done` on the (normal) success path, so it survives only when the op also wedges later. Asserted in unit tests; never observed in a running worker.


ZacxDev:
## Audit follow-up — and a correction to this PR's own body

Blind adversarial audit: **no 🔴, verdict "safe to merge"**, five 🟡. Addressed in `3a23f35b`. Everything below was re-run rather than quoted.

**🟡 The guard set could not see the hazard the PR itself names.** Both existing assertions are *ceilings*, so nothing constrained the constant downward — measured, **`REUSE_TAB_BUDGET_MS = 100` passed the entire 507-test suite**. Wrong direction to leave open: the bound's whole disclosed cost is a low-side failure, where a slow-but-healthy probe orphans a live tab that nothing reclaims. Floor added and **boundary-tested for reachability**, not presence:

| value | result |
|---|---|
| 100 | fail=1 (previously survived 507/507) |
| 999 | fail=1 (one below the boundary) |
| 1000 | pass |
| 2000 | pass (shipped) |

⚠️ The floor is **conservative, not calibrated**, and the comment says so. There is no measurement of a healthy `chrome.tabs.get` in this repo, so unlike #797's `> 1365` (anchored to a measured 292–1365 ms band) it is anchored to the cheapest chrome.* round-trip we *do* have a number for — `activate` at a measured ~350–500 ms, which does strictly more work than `tabs.get`. Replace it the moment anyone measures the real thing.

**🟡 Correction: the severity claim in this PR body and in three comments was false.** "a `cmd_timeout` on `open`, the op every session starts with" — `server.py:2134` injects `reuseTabId` **only** when `_owned_tab_locked()` already returns a tab, so a fresh session's first `open` never enters the bounded try. The affected path is a **re-open by a session that already owns a live tab**. Common, but not what was written, and the claim was doing work in the justification.

**🟡 This change falsifies a guarantee still asserted in three docs.** `server.py`, `README.md` and `reference/tabs-instances.md` all still said a double `open` "never orphans/leaks the first tab". After this change a probe over budget **deterministically** orphans it. Each now carries the exception at the point of the claim.

**🟡 A second vacuous assertion shipped.** `assert.notEqual(hung.phase, gone.phase)` is strictly implied by the two preceding `assert.equal`s against distinct literals — it cannot fail for any mutant. The M5 row credited a kill to its message, which never prints (the mutant dies at `:522` with a generated `strictEqual` message). Removed. Same class this PR had already found and removed once.

**🟢** The in-file note said "tests 18 of 30, i.e. 12 never ran"; the file holds 28, so it is 10. Corrected.

### Not fixed, deliberately

The audit's 🟡5 stands: the fallthrough is `chrome.tabs.create`, **another unbounded browser-process IPC**, so unlike #797 (fast path → CDP, a genuinely different mechanism) the recovery shares the suspected failure mode. If `tabs.get` stalls browser-wide, `create` stalls too, the op still dies at 18 s having burned 2 s — and worse, every `open` then returns *success* while leaking a tab, turning a loud `cmd_timeout` into a silent, cumulative leak with no counter and no reclamation. Fixing that needs a live measurement of a hung `tabs.get`, which nobody has. Flagged rather than papered over.

Suites: **507 mjs + 787 python**. Build marker regenerated.

PR #816, opened 2026-08-25
test(clickup+ccua): one TABLE for the awaiting predicate — the code consolidation is a regression, and here is why

ZacxDev:
## The task, and the answer

The predicate **"the newest comment on this task is NOT the token owner's"** is implemented twice, in two languages. `scripts/check-clickup-addressed/` (ccua) *already* shells out to the CLI that owns the JS copy (`recent-comments.py:24` runs `node query.mjs`, `cwd=CLICKUP_DIR`). So the obvious consolidation is: have ccua call `query.mjs awaiting` and stop re-deriving.

**It cannot. Consolidating the code would be a regression, and this PR ships the measurement rather than the argument.**

🔴 **No production behaviour changes.** The diff is a shared fixture, two test files, one reference doc, three pointer comments, and two floor-table bumps.

## What the investigation found

**1. The brief's cited duplicate is not the duplicate.** `recent-comments.py:170` (`str(c["user"]["id"]) != my_id`) is a per-comment *authorship filter* — a different predicate. The real second implementation is `check-addressed.py::_reply_answers_the_comment` fed by `recent-comments.py::latest_reply_ts_by`, and it is the **richer** of the two: three states plus an absent-key state, against `isAwaiting()`'s bare boolean. The proposed consolidation direction is backwards.

**2. The decisive blocker is structural, not a missing field.** `awaiting` emits a row *only* when the newest comment is not the owner's — so **every task the owner answered last is absent by definition**. That is exactly the population `_waiting_verdict`'s ANSWERED/suppression branch is built from: the branch carrying `BOT_IDENTITY_CAVEAT` ("an agent may have answered as you"). Feeding ccua from `awaiting` deletes that whole block, and nothing in either suite would have failed. Asserted from both sides (`selectAwaiting` yields 0 rows; ccua still yields a record with `my_latest_reply_ms` and the consumer reaches ANSWERED on it).

**3. Four further blockers**, enumerated machine-readably in the fixture: comment **text** (ccua's entire decision surface — keep-open veto, RESOLVED-comment flag, `TEXT_CHARS`); **`my_latest_reply{,_ms}`**; **`task_priority`**; and the **failure mode** — `query.mjs awaiting` exits 1 on an unresolvable owner id, where ccua continues, warns on stderr, and *withholds* the key so the consumer takes its announce-rather-than-decide branch.

**4. The `UNIDENTIFIED` distinction does not survive a move to `awaiting.mjs`.** It is a property of the **return type**, so no amount of care inside `isAwaiting()` could recover it. Given a thread it cannot rank, Python declines to claim and the flag says the check did not run; JS returns a confident `true`.

Blockers 3 could be added at the cost of turning a triage command into a data dump. Blocker 2 cannot — it is the command's definition.

## What *was* consolidated: the contract, not the code

`claude/skills/clickup/test/awaiting-contract.fixtures.json` — one shared, synthetic table read by **both** suites:

| suite | measures |
|---|---|
| `claude/skills/clickup/test/awaiting-contract.test.mjs` | `lib/awaiting.mjs::isAwaiting()` |
| `scripts/check-clickup-addressed/tests/test_awaiting_contract.py` | the ccua producer → consumer pipeline, end to end |

Neither side copies the other's column — **each measures its own**, so the table cannot lie about either implementation. Both recompute the divergence ledger from the two columns and pin it **as a set that fails when it grows *or* shrinks**. This is the only thing that makes the two agree; the reference doc says so, so it does not get deleted as "an unused fixture".

## Three cross-language divergences, previously unknown

Found by writing the table:

- **`owner_reply_in_the_same_millisecond`** — Python resolves an exact tie by an explicit documented rule (`mine_ms >= theirs_ms`, a tie counts as answered). JS resolves it by **array order** (`newestComment()` keeps the first of an equal pair), so reversing the list flips the verdict — and in the API's own order it answers the **opposite** of Python.
- **`the_owners_own_comment_has_an_unreadable_date`** — Python declines to claim (UNIDENTIFIED); JS returns a confident `true` over a thread it cannot rank.
- **`every_comment_has_an_unreadable_date`** — Python declines; JS's answer is a property of the response order rather than of the thread.

**Pinned, not fixed.** None is reachable today — all 84 comments sampled when `awaiting` was written carried a `date`, and a tie needs two writes in the same millisecond. Changing a shipped command's behaviour on an unobserved input buys nothing and is a live behaviour change; the order-dependence label is *proved by reversing the list*, never asserted.

## Test evidence

🔴 **These are seam / invariant guards, NOT regression coverage.** No defect is fixed here, so "red at base" would be vacuous — nothing below existed at base. The honest evidence is a mutation sweep.

**9 mutants, 9 KILLED**, each by a named test with its own message. Run under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` purged between mutants, a **green baseline as the positive control**, and sha256 restore verification on all three mutated files afterwards (all `OK`).

| # | file | mutation | killed by |
|---|---|---|---|
| M1 | `awaiting.mjs` | `isAwaiting`: unknown author → `false` | `the_newest_comment_has_no_author_id` |
| M2 | `awaiting.mjs` | `isAwaiting`: `!==` → `===` | 8 tests incl. both blocker guards |
| M3 | `awaiting.mjs` | `newestComment`: undated sorts `Infinity` | `the_owners_own_comment_has_an_unreadable_date`, the UNIDENTIFIED guard |
| M4 | `awaiting.mjs` | `selectAwaiting`: drop the `isAwaiting` filter | `emits NO row for a task the owner answered last` |
| M5 | `awaiting.mjs` | add `text` to the row | `an awaiting row carries exactly these fields` |
| M6 | `recent-comments.py` | `return UNIDENTIFIED` → `continue` | column test + UNIDENTIFIED guard |
| M7 | `recent-comments.py` | emit `my_latest_reply` unconditionally | column test + both sentinel guards |
| M8 | `check-addressed.py` | `mine_ms >= theirs_ms` → `>` | column test (the tie case moves) |
| M9 | `recent-comments.py` | stop dropping the owner's own comments | both blocker guards |

A note on how the table itself was validated: the Python column was **written from the spec and then corrected by the measurement** — the first draft scored `only_a_colleague_has_commented` as decidable-`true` and the pipeline returned `unverified`, because `_reply_answers_the_comment` returns `None` when there is no reply date to compare. That is present-and-null (`"I looked; you never replied"`) versus absent (`"nobody could look"`), the exact distinction the ccua seam exists to protect. The helper now makes it explicitly, and the discovery is recorded in its docstring. That red run is also the suite's organic negative control.

## Gate verdicts — both tiers, named

Base sha: `324693fd`.

- **dev-host tier** (`nix develop --impure --command bash scripts/gate.sh --tier both`): `GATE: RESULT=PASS exit=0` — `PASS pytest exit=0`, `PASS node exit=0`.
- **nix sandbox tier — nodetests** (`nix build .#checks.x86_64-linux.nodetests`): exit 0, `RESULT: PASS (exit=0)`, `claude/skills/clickup/test (files=6 tests=188 pass=188 skipped=0 floor=179)`.
- **nix sandbox tier — pytests** (`nix build .#checks.x86_64-linux.pytests`): see the comment below.

(The first `gate.sh` attempt outside the dev shell exited 3 on `FATAL — required tool(s) missing from PATH: logrotate` — a missing environment, not a code failure. Re-run inside `nix develop`, as the runner's own message instructs.)

## Floors

Both raised deliberately rather than left with slack, per the per-target header's own rule:

- `run-node-tests.sh`: `claude/skills/clickup/test` `5|163` → **`6|179`** (188 measured; file count moves 5→6).
- `run-tests.sh`: `scripts/check-clickup-addressed/tests` `203` → **`224`** (235 measured, agreeing with `tests/run_all.py` — two runners, one number).

Neither was forced by the gate; both were re-pinned because leaving the slack lets these files vanish while the gate stays green, which is the staleness a per-target floor exists to prevent.

## Read this before re-proposing the consolidation

`claude/skills/clickup/reference/awaiting-vs-ccua.md`, routed from `SKILL.md`, and pointed at from the header of **both** implementations — so neither can be read in isolation and the question does not get re-derived from scratch a third time.

### DIFF
--- a/claude/skills/clickup/SKILL.md
+++ b/claude/skills/clickup/SKILL.md
@@ -121,3 +121,6 @@ there is no ClickUp hook; it is a convention, and the flow says so first.
 - **CHANGING this skill** — where to edit so the change deploys, the nix-built
   `node_modules` + `npmDepsHash`, and the test suites →
   `~/.claude/skills/clickup/reference/maintaining.md`.
+- **"`check-clickup-addressed` re-derives `awaiting`'s predicate — consolidate them"**
+  → `~/.claude/skills/clickup/reference/awaiting-vs-ccua.md`. It cannot; the answer is
+  measured, not argued. Read it BEFORE proposing that ccua call `query.mjs awaiting`.
--- a/claude/skills/clickup/lib/awaiting.mjs
+++ b/claude/skills/clickup/lib/awaiting.mjs
@@ -29,6 +29,19 @@
  * Everything here is pure except `findAwaiting`, whose two effects (fetching
  * comments, sleeping) are injected, so the cap and pacing are testable without
  * a network.
+ *
+ * 🔴 THIS PREDICATE IS IMPLEMENTED TWICE, AND THAT IS DELIBERATE.
+ * `scripts/check-clickup-addressed/` derives the same fact across a seam
+ * (`recent-comments.py::latest_reply_ts_by` -> `check-addressed.py::
+ * _reply_answers_the_comment`), and it already shells out to this CLI — so
+ * "have ccua call `query.mjs awaiting`" looks obvious and is a REGRESSION. The
+ * decisive reason is one line below (`selectAwaiting` skips a task the owner
+ * answered last, which is exactly the population ccua's suppression note is
+ * built from); the rest, plus the three measured cross-language divergences,
+ * are in `reference/awaiting-vs-ccua.md`. The two implementations are pinned to
+ * ONE shared table — `test/awaiting-contract.fixtures.json`, read by
+ * `test/awaiting-contract.test.mjs` and by the Python half — so neither can
+ * drift silently. Read that file before changing anything here.
  */
 
 export const MS_PER_DAY = 86400000;
--- a/claude/skills/clickup/reference/awaiting-vs-ccua.md
+++ b/claude/skills/clickup/reference/awaiting-vs-ccua.md
@@ -1,0 +1,88 @@
+# `awaiting` vs `check-clickup-addressed` — why the predicate is not consolidated
+
+**The question this file answers, once:** the predicate *"the newest comment on this task
+is NOT the token owner's"* is implemented twice in this repo, in two languages.
+`scripts/check-clickup-addressed/` (ccua) already shells out to this CLI. So why doesn't
+it call `query.mjs awaiting` and stop re-deriving?
+
+**Because it would be a regression, and the specific losses are enumerated and tested.**
+Read this before proposing the consolidation again — the reasoning below is measured, and
+every claim in it is pinned by a test that goes red when it stops being true.
+
+## Where the two implementations actually are
+
+| side | file | shape |
+|---|---|---|
+| JS | `lib/awaiting.mjs :: isAwaiting()` | per TASK, a **boolean** |
+| PY | `recent-comments.py :: latest_reply_ts_by()` → `check-addressed.py :: _reply_answers_the_comment()` | per COMMENT, a **three-state** derived across a seam |
+
+⚠️ `recent-comments.py`'s author comparison (`str(c["user"]["id"]) != my_id`) is **not**
+the duplicate — that is a per-comment *authorship filter*, a different predicate. The
+duplicate is the pair in the table above, and the Python side is the **richer** one, which
+is why the consolidation direction that looks obvious is backwards.
+
+## The blockers
+
+Machine-readable in `test/awaiting-contract.fixtures.json` → `blockers`. In severity order:
+
+1. **Population (decisive).** `awaiting` emits a row *only* when the newest comment is not
+   the owner's. Every task the owner answered last is structurally absent — and that is
+   exactly the population `_waiting_verdict`'s **ANSWERED / suppression** branch is built
+   from, the branch carrying `BOT_IDENTITY_CAVEAT` ("an agent may have answered as you").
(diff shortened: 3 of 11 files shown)


ZacxDev:
### nix sandbox tier — pytests

`nix build --impure .#checks.x86_64-linux.pytests` → **exit 0**, `RESULT: PASS (exit=0)`.

```
PASS  scripts/check-clickup-addressed/tests  (collected=235 passed=235 skipped=0 floor=224)
TOTAL collected=15837  passed=15835  skipped=2  failed=0  (floor: 13753 = sum of 27 per-target floors)
RESULT: PASS (exit=0)
```

**Zero new skips.** All four verdicts, named, against base `324693fd`:

| tier | command | verdict |
|---|---|---|
| dev-host (both) | `nix develop --impure --command bash scripts/gate.sh --tier both` | `GATE: RESULT=PASS exit=0` |
| nix sandbox | `nix build .#checks.x86_64-linux.nodetests` | exit 0, `RESULT: PASS`, clickup `files=6 tests=188 floor=179` |
| nix sandbox | `nix build .#checks.x86_64-linux.pytests` | exit 0, `RESULT: PASS`, ccua `collected=235 floor=224` |

⚠️ Each is a claim about **one run, one tier, one base sha** — `strict` is false on `main`, so these gate this branch, not the tree the merge creates.

PR #819, opened 2026-08-25
refactor(transcript search): one walk for /find-session and check-clickup-addressed — and the seven ways the two copies disagreed

ZacxDev:
## What

`scripts/find-session.py` (the `/find-session` skill) and
`scripts/check-clickup-addressed/search-sessions.py` both walked
`~/.claude/projects/**/*.jsonl` behind the same five flags (`--since`, `--limit`,
`--project`, `--any`, ranked positional terms) and **shared no code**.
`check-completion.py` carried three more walks of the same corpus
(`_find_sessions_for_task`, `session_path`, `_read_session`).

They now all go through **`scripts/lib/transcript_search.py`**.

## Was ccua allowed to take a shared import?

Checked first, because the skill migrated out of `datapacket-talos` on 2026-08-22 and
might have been meant to stay copyable. It was not: the migration commit re-anchored
`_selfrun.py`'s path markers to **devrc's** layout (`NEW_LAYOUT_RE`), `run-tests.sh`
adopted it as a first-class target, and its SKILL.md tells you to run it out of the
devrc working tree. Nothing asks it to stay portable.

The one real constraint was its own instrument: `tests/mutation_sweep.py` copies the
skill into a temp dir and runs the suite there, aborting unless a NULL-CONTROL mutant
scores SURVIVED. A bare `../lib` import would have made every test file score
`FAILED TO IMPORT` and disarmed the sweep. Fixed by having the sweep reproduce
`scripts/{check-clickup-addressed,lib}/` in the temp tree — so there is still exactly
**one** import rule, a relative `../lib`, true in the repo and in the copy alike. The
alternative (a second candidate path that only fires under test) was rejected.

## The disagreements

Seven were bugs. Each was watched **RED at `324693fd`** through the CLI modules — not
through the new library, which does not exist on that tree — and green at HEAD, with
the same harness pointed at both trees.

| # | disagreement | resolution |
|---|---|---|
| 1 | find-session's **`--all` was inert**: its handler sat behind `if not a.all and typ not in ("user","assistant")`, twenty lines after an unconditional `continue` had already skipped everything it could admit. SKILL.md advertised it for tool output the whole time. | `--all` now selects the search surface (text + tool_use input + tool_result). SKILL.md says so, and says the flag was inert until today. |
| 2 | find-session's **`--since` compared naive UTC to naive LOCAL midnight**. On this UTC-0500 host `--since 2026-08-24` matched a session whose last message was 2026-08-23 21:00 local. | Both sides converted to local. Test measures **two timezones with opposite-sign offsets** and pins `TZ` explicitly rather than inheriting it. |
| 3 | find-session **never searched `ai-title` records**; search-sessions always did. **490 of 792** live transcripts carry one. | Both search titles. |
| 4 | One **tz-less timestamp beside an offset one** raised `can't compare offset-naive and offset-aware`, swallowed by a blanket `except` that then discarded every *later* timestamp — the session was dated by whichever record came first. | Timestamps are parsed aware unconditionally (bare = UTC). |
| 5 | search-sessions **dropped a whole transcript on ONE malformed JSONL line** — the normal shape of a file still being written — indistinguishably from the session not matching. | Skip the line. Measured: **0 of 792** currently carry one, so this closes a hazard, not a firing bug. |
| 6 | `check-completion._read_session`'s `json.loads` had **no guard at all**: the same malformed line aborted the entire completion check. | Same reader as the search stage, so the two cannot disagree about which lines a transcript contains. |
| 7 | search-sessions' **`--project` was case-SENSITIVE** and blind to the recorded `cwd`; find-session's was neither. | Unified on the permissive one. `check-addressed.py` passes no `--project`, so no ccua verdict moves. |

**Unified deliberately** (not bugs, but two answers to one question):

- **hit counting** — find-session counted *messages containing* a term, search-sessions counted *occurrences*. Unified on occurrences; find-session's ranking shifts accordingly.
- **`--since` source** — content timestamp of the last message (search-sessions read file **mtime**), with mtime as the fallback when a transcript has no parseable timestamp.
- **`opening`/`genesis`** — the cleaned first user message. search-sessions used the raw first user record, so it routinely displayed a `<system-reminder>` blob or a `Caveat:` preamble.

**Kept different on purpose:**

- **search surface.** find-session defaults to text only (a human reading results wants what was said); ccua defaults to text + tool_use input (a task id typed into a Bash command *is* evidence the task was worked). Measured on `"drift-check.sh"` over the live corpus: **45 vs 51 sessions**. Neither default is imposed on the other.
- **`_find_sessions_for_task`'s predicate.** It is a raw substring test over the whole JSONL line — wider than any parsed surface, deliberately, because it is the fallback that runs when no `--session` was supplied. Only the *corpus walk* was consolidated; narrowing the predicate would have shrunk what the completion scan can see. Pinned by a test.

## 🔴 The corpus is not flat

Measured over `~/.claude/projects`: **792** session transcripts at
`<project>/<id>.jsonl`, and **4,776 more** at
`<project>/<id>/subagents/agent-*.jsonl`. find-session excluded those by name;
search-sessions used a flat `glob("*.jsonl")` and never recursed deep enough to meet
one. Two mechanisms, same answer — and only one of them survives a layout change. The
shared walk **recurses and excludes by name**, so both properties now hold for both
tools, pinned rather than assumed.

## Behaviour preserved

Result sets diffed base-vs-HEAD over the live corpus for both CLIs on two terms:
`find-session` 45/45 and 33/33, `search-sessions` 51/51 and 35/35 — **identical sets**,
not just identical counts.

## Test matrix

Seven CLI-level checks, one harness, two trees:

```
tree: <base 324693fd export>            tree: HEAD
  RED  --all widens the surface           GREEN
  RED  --since is a local calendar day    GREEN
  RED  ai-title is searchable             GREEN
  RED  bare ts beside an offset ts        GREEN
  RED  malformed line costs the line      GREEN
  RED  _read_session survives it          GREEN
  RED  --project ignores case             GREEN
  7 of 7 RED                              0 of 7 RED
```

Committed as `RED_AT_BASE` ledgers in `scripts/tests/test_transcript_search.py` and
`scripts/check-clickup-addressed/tests/test_shared_walk.py`, each asserted against the
module's own test functions so a name cannot quietly join or leave the claim.

**Labelled as NOT regression coverage** (green at base too, kept because the mechanism
that made them true changed): the subagent-exclusion guards, `session_path` refusing a
subagent id, the enumerator ledger, and the mutation-sweep `SHARED_MODULES` ledger.

**Mutation-tested, and proven reachable:** two new mutants over the shared walk
(`TS1` deletes the subagent exclusion, `TS2` restores the whole-file abort) both
**KILLED by their own named tests** — not by a different guard's error. `NULL-CONTROL`
SURVIVED and the existing `PC` positive control stayed KILLED. Adding these as `T1`/`T2`
silently collided with existing ids, so `mutation_sweep.py` now refuses a duplicate id.

## 🔴 The merge with main found one more, and it was in the instrument

`main` moved under this branch (#816 added `tests/test_awaiting_contract.py`). The
textual conflict was the usual one — `TARGET_FLOORS`, both sides adding tests to the
same target — and was resolved by **measuring the merge**, not by arithmetic: 244
collected by `pytest --collect-only`, `244 passed, 0 failed` by `tests/run_all.py`, two
runners one number, giving the gate's own `_suggested_floor` of 232. The first value
written here was a hand-derived 256; measurement rejected it, which is exactly why the
`TARGET_FLOORS` header forbids computing it.

The **semantic** conflict was invisible to that merge. `tests/mutation_sweep.py` runs
every mutant inside a copy of the skill directory, and main's new test reads its shared
contract table at MODULE level via `parents[3] / claude/skills/clickup/test/`. That path
does not exist in the copy → `FAILED TO IMPORT` → `NULL-CTL KILLED` → **the sweep
aborts, and no mutant can be scored at all.** Measured here; it aborts on `main` alone
for the same reason.

Fixed by carrying the fixture into the mutant tree at its repo-relative path
(`REPO_FIXTURES`), *not* by adding the file to `EXCLUDED_FROM_MUTANT_RUNS` — the table is
data the mutants are scored against, so excluding it would have made the instrument green
by dropping real coverage. The copy step is now `materialize()`, shared by `run_one` and
by a new guard that asserts the **outcome** (every test file imports in the copy), because
a module-level read is precisely what a name-list of shared modules cannot anticipate.
Verified reachable: emptying `REPO_FIXTURES` makes it fail with its own message.

Also re-anchored mutant `TS1`, which started scoring `NOT APPLIED` after the
subagent-exclusion predicate moved into `is_corpus_member` — a stale mutant reads like a
row rather than an absence. After: `NULL-CONTROL SURVIVED`, `TS1`/`TS2` KILLED by their
own named tests.

## Gate — both tiers, one run, on the merged sha

Run on `62f43cdf` (the merge commit, i.e. the tree the merge creates — not the branch
before it). Named because "the gate passed" is true of one run, one tier, one base.

| tier | command | verdict |
|---|---|---|
| dev-host | `nix develop . --command bash scripts/gate.sh --tier both` | `GATE: RESULT=PASS exit=0` (script exit **0**) — pytest `TOTAL collected=15870 passed=15868 skipped=2 failed=0` (floor 13761), node `TOTAL suites=4 files=36 tests=1219 pass=1219 fail=0` (floor 1189) |
| nix sandbox | `nix build .#checks.x86_64-linux.pytests` | `RESULT: PASS (exit=0)`, build exit **0** — same 15870/15868 inside the store copy |
| nix sandbox | `nix build .#checks.x86_64-linux.nodetests` | `RESULT: PASS (exit=0)`, build exit **0** |

`scripts/check-clickup-addressed/tests` reports `collected=244 floor=232` in both tiers.
`scripts/tests` moved 7885 → 7908 and is deliberately **not** re-pinned (reason in the
`TARGET_FLOORS` comment: its slack is pre-existing and tightening it would fail any
concurrent PR that lands with fewer tests).

⚠ An earlier dev-host run of mine printed floors that did not match this tree
(`floor=224`, a clickup node suite with 6 files instead of 5). It was contaminated —
another clone, `~/workspace/devrc-gate-816`, was running the same suite concurrently on
this box, and those were #816's numbers. Caught by checking the printed floors against
the file rather than reading the verdict, and re-run to a marked file. Every number above
is from a marked run.

Not merged.


ZacxDev:
## 🔴 Correction to this PR body: three of the seven are LATENT, not firing

Posting this as a comment rather than editing the body, because a reviewer may already
have read the wrong version.

The body says **"Seven were bugs"** and prints a **7-of-7 RED** matrix. Both are true
about the CODE — every one of the seven was watched red at `324693fd` and green at HEAD.
Neither is true about the CORPUS: the matrix says the old code got the answer wrong on a
constructed fixture, and it reads as "seven defects were firing on live data". Measured
2026-08-25 over `~/.claude/projects` (797 corpus transcripts, 424,853 user/assistant
records):

| # | what it is | live instances | status |
|---|---|---|---|
| 1 | `--all` was inert | every use of the flag | **FIRING** |
| 2 | `--since` compared naive UTC to naive LOCAL | every `--since` on this UTC-0500 host | **FIRING** |
| 3 | find-session never searched `ai-title` | 493 of 797 transcripts carry one | **FIRING** |
| 4 | tz-less timestamp beside an offset one | **0 of 797** — 0 transcripts carry ANY tz-naive timestamp, so the mixed-file precondition occurs 0 times | **LATENT** |
| 5 | one malformed JSONL line discarded the whole transcript | **0 of 797** | **LATENT** (body already says so) |
| 6 | `_read_session`'s unguarded `json.loads` | same 0 of 797 — same input class as #5 | **LATENT** (body already says so) |
| 7 | `--project` was case-sensitive and blind to `cwd` | 0 in the ccua pipeline (`check-addressed.py` passes no `--project`); firing for a human typing it into `search-sessions.py` | **PARTLY FIRING** (body already says so) |

So the honest headline is **three firing defects, three latent hazards with zero live
instances, and one that fires only on a flag the automated pipeline never passes** —
not seven of seven.

Why it matters rather than being pedantry: a latent-hazard fix and a firing-bug fix carry
different evidence. The tests for #4/#5/#6 are regression tests against the CODE — they
were watched red at base — but they are **not** evidence that any answer this tool ever
gave was wrong. Row 4 is the one the body does not caveat, and it is the one where the
precondition (a transcript mixing aware and naive timestamps) is currently unreachable in
the live corpus: nothing writes a bare timestamp today.

Reproduce: count corpus files (recurse, exclude `subagents/` and `wf_*`) whose
`user`/`assistant` records carry a `timestamp` with no `Z` and no offset.

Everything else in the "Test matrix" section stands as written; this corrects only what
the seven rows are evidence OF.

PR #822, opened 2026-08-25
fix(escrow-verify): the advertised shell could not run its own command, and a mismatch never said WHAT chose the file

ZacxDev:
Three defects, all surfaced by one live run on 2026-08-25 in which the operator spent a master password and got a **confident wrong answer** — plus a fourth found by the adversarial audit of the first attempt to fix it. None is a logic bug; all are claims the code made that it had not earned.

> ⚠️ **This body was rewritten after the audit.** Its first version made two false claims of its own — see the correction comment on this PR. The table of sources listed three entries when there are four, and it said the refusal "leads" when the code appended it *after* the remedy. Both are now true of the code; they were not when first written.

## 1. The advertised shell could not run its own command

`escrow-verify.py`'s failure messages hand the operator an exact `nix-shell` command. It provisioned `bitwarden-cli jq` only — but `--decrypt-check` reaches MinIO through `backup.MinioUploader`, whose `minio` import is **lazy** (inside the method, not at module scope).

So the advertised shell parsed, started, unlocked the vault, and *then* died:

```
ModuleNotFoundError: No module named 'minio'
```

— after the password had been spent. The lazy import is why this survived every import-time check and only appeared mid-run against the real bucket. A hint that cannot complete its own invocation is worse than no hint, because it is followed.

**Fix:** the packages are a ledger (`NIX_SHELL_PACKAGES`) and the hint is *derived* from it, pinned against `DECRYPT_PYTHON_MODULES`. Quoting is `shlex.quote`, and the guard runs the rendered hint through **`bash -n`** with an unquoted positive control — the first version asserted via `shlex.split`, which does not treat parens as metacharacters, so deleting the quoting entirely left the suite green while the advertised command became a bash syntax error.

## 2. A mismatch never said which file it compared

The run returned `rc=22 BYTES-DIFFER-MATERIALLY`, *"189 escrowed bytes vs 189 on disk"*, and advised re-escrowing from the on-disk identity. **Following that advice would have overwritten a good escrow with an unrelated third party's age key.**

`resolve_identity()` reads `ASIB_AGE_IDENTITY` → `SOPS_AGE_KEY_FILE` → the default. `SOPS_AGE_KEY_FILE` was exported by unrelated work and had redirected the on-disk side to a different key. The message named the resolved *path* but never said anything had chosen it.

🔴 **"189 vs 189" is not corroboration — it is what two DIFFERENT keys look like.** Every age identity file is the same size (fixed-width `# created:`, `# public key:` and key lines), so equal byte counts carry no information about whether the keys match. That is what made the wrong reading so plausible.

## 3. 🔴 The first fix asked the WRONG QUESTION — and printed two false sentences

Found by audit, reproduced live before fixing. The first revision branched on **what named the file** rather than **which file it was**:

```
$ escrow-verify.py --print-plan --identity ~/workspace/homelab-talos/.secrets/age.key
chosen by: the --identity flag  <- NOT the default; an unrelated shell can set this
```

That path **is** the default, and **no shell can set a command-line flag**. Worse, it was the command this PR's own handoff doc recommends. The same false line fired for `SOPS_AGE_KEY_FILE=<the default path>` — which is exactly what the deployed backup unit exports (`nix/home.nix`) — making it a **permanently-red warning on the subsystem's own normal configuration**, the precise thing that trains a reader to skip the sentence when it finally matters.

**Fix:** `B.same_identity_file()` (symlink-resolving, non-strict) is the trigger; the source is carried as *information*. Four sources, and the warning must discriminate:

| resolved file | stated source | message |
|---|---|---|
| **is** the default | any of the four | names the source, **no** warning |
| not the default | an env var | names it + refusal + `env -u <VAR>` to undo the redirect |
| not the default | `--identity` | its own wording — a flag is a deliberate act, so no "unrelated shell", and it does **not** advise re-running with the flag they just used |
| any | none stated | asserts **nothing** — no default is invented for a caller that never checked |

The refusal now genuinely precedes the remedy it argues against. `--print-plan` discloses a redirect before any password is spent, and the **success** line discloses one too — "escrow OK — matches `<path>`" is where an undisclosed wrong-file comparison is least likely to be questioned.

## 4. The sibling tool had the same blind spot

The mismatch message routes the operator to `restore-verify.py` — which resolved its own identity through the same env vars, so the handover would land on the same redirected key. A control that shares the step under suspicion is not a control. It now discloses provenance in `--print-plan` and names a redirect as a third mechanism in its `DECRYPT FAILED` message (which previously enumerated only "wrong key or damaged ciphertext").

## Verification

**Red at base `a8089a51`, green at HEAD — by execution**, not by the new tests' collection error:

| defect | base | HEAD |
|---|---|---|
| advertised hint | `nix-shell -p bitwarden-cli jq --run '<command>'` | `… jq 'python3.withPackages(p:[p.minio])' --run …` |
| `--print-plan` under a real redirect | path only, no disclosure | `chosen by: $SOPS_AGE_KEY_FILE  <- NOT the default identity` |
| `--identity <the default>` | *(first fix)* falsely "NOT the default" | `chosen by: the --identity flag`, no warning |
| `SOPS_AGE_KEY_FILE=<the default>` | *(first fix)* falsely "NOT the default" | `chosen by: $SOPS_AGE_KEY_FILE`, no warning |

**Mutation sweep — 10 semantic mutants, not deletions**, `PYTHONDONTWRITEBYTECODE=1` (the repo's documented stale-bytecode trap), each asserting its own edit applied so a no-op cannot score as a kill: **10/10 KILLED**, including the original bug deliberately reintroduced and a positive control. All seven survivors the audit reported are now caught.

Both gate tiers on the head under review (`93f77813`), run separately because they are different environments, and read from output rather than exit codes:

| tier | pytest | node |
|---|---|---|
| dev host — `scripts/gate.sh --tier both --set all` | 15,871 passed / 0 failed (floor 13,753) | 1,219 / 0 |
| nix sandbox — `nix build .#checks.x86_64-linux.{pytests,nodetests}` | 15,871 passed / 0 failed | 1,219 / 0 |

🔴 **They disagreed one commit earlier, and that is the point.** At `6225610a` the dev-host tier was green at 15,865 while the sandbox tier was **RED** — one test in 27 targets, which read the operator's real age key and so passed only on the machine that has it. `env HOME=<empty dir>` reproduces that whole class in ~20s without a nix build.

## Review rounds

Three adversarial rounds; each of the first two found a real defect in the previous round's fix.

| round | verdict | what it caught |
|---|---|---|
| 1 | needs rework | the first provenance fix branched on **what named** the identity, not **which file** it was, and printed two false sentences — one on the command this PR's own doc recommends. Plus 7 mutation survivors. |
| 2 | merge after fixing 🔴 | the round-1 fix shipped an environment-dependent test that turned the merge-gating tier red; the success-line disclosure was inert-able with a green suite; nothing pinned `DEFAULT_IDENTITY` to what the backup unit actually exports. |
| 3 | safe to merge, no 🔴 | the round-2 fix was **unguarded** — reverting `restore-verify`'s note to the wording round 2 rejected left all 450 tests green. An unguarded fix is a comment, not a guarantee. |
| 4 | safe to merge, no 🔴 | one 🟢: the neutral refusal arm had only word-**absence** assertions, so an **inverted safety claim** ("this certainly means the escrow is DAMAGED, re-escrow at once") using none of the forbidden words passed 451/451. Fixed anyway — pinned whole. |

Three instrument failures were caught during this work, each of which would have produced a confident false result — recorded because none is visible in the diff:

- a mutation sweep that copied only `scripts/`, so the unmutated baseline failed 19 tests and every mutant "died" for that reason — a clean 5/5 measuring nothing;
- a sandbox copied with `rsync -a` carrying **valid** `__pycache__`, so it executed bytecode from the original tree and code mutants scored SURVIVED. 🔴 `PYTHONDONTWRITEBYTECODE=1` does **not** prevent this — it stops writing, not reading;
- a background-task notification reporting **"exit code 0"** for a build that failed, because a trailing `echo` owned the pipeline's status.

Every sweep here now runs an unmutated baseline first and requires it GREEN, asserts no `.pyc` reached the sandbox, and includes a no-op control that must SURVIVE — so a harness that reddens everything is distinguishable from real coverage.

Detail: **round 3's first mutation sweep was invalid** — it copied only `scripts/`, so the unmutated baseline failed 19 tests and all five mutants "died" for that reason, reporting a clean 5/5 that measured nothing. Corrected, the baseline returns 450 passed and each mutant dies with exactly one failure.

## Out of scope

Whether the escrowed note matches the real `age.key` is **still unconfirmed** — the 08-25 run neither confirmed nor refuted the 08-23 `cmp`, because it compared against the wrong file. The on-disk homelab key *is* confirmed as the one the bucket opens (`restore-verify.py` → rc=0, 10 artifacts, 214 commits, timer-produced).

Not covered by any guard here: the **non-Python** tools the decrypt path needs (`age`, `kubectl`, `git`) resolve from the ambient profile because `nix-shell -p` is impure. `--pure`, a root shell, or a fresh host would reproduce the same class of failure with nothing watching.

### DIFF
--- a/claudedocs/handoff-analyze-service-index-backup.md
+++ b/claudedocs/handoff-analyze-service-index-backup.md
@@ -344,15 +344,20 @@ not, which is why the same command behaved differently for the agent and the hum
 age identity file is the same size (fixed-width `# created:`, `# public key:` and key
 lines). Equal byte counts on both sides carry no information about whether the keys match.
 
-Public-key hashes, compared without printing key material — all four are distinct
-identities:
+Public-key hashes, compared without printing key material — all four are **distinct**
+identities. Three client checkouts carry their own key at a `…/.secrets/sops/age.key` path;
+they are not enumerated here (this repo is public) and which one it landed on does not
+change the conclusion:
 
 | identity | pubkey sha (first 16) |
 |---|---|
 | `homelab-talos/.secrets/age.key` (the real one) | `288c4d24cfdb5aa1` |
-| `civit/postgres-debezium-kafka-replication/.secrets/sops/age.key` | `7e3367f5a3a5d327` |
-| `civit/civitai-gpu-fleet/.secrets/sops/age.key` | `6643aa0a698317cf` |
-| `civit/datapacket-talos/.secrets/sops/age.key` | `8e6095ea38340cef` |
+| client checkout A | `7e3367f5a3a5d327` |
+| client checkout B | `6643aa0a698317cf` |
+| client checkout C | `8e6095ea38340cef` |
+
+Reproduce with `age-keygen -y <file> | sha256sum` — it prints the PUBLIC half only, so the
+comparison never handles secret material.
 
 **The on-disk homelab key is confirmed correct, independently of the vault.**
 `restore-verify.py --identity ~/workspace/homelab-talos/.secrets/age.key` → **rc=0**, all
--- a/claudedocs/handoff-analyze-service-index-backup.md
+++ b/claudedocs/handoff-analyze-service-index-backup.md
@@ -51,7 +51,11 @@ and close the one failure mode in it that is unrecoverable (no off-machine backu
    escrowed key still there, still byte-identical, and does it still *work*". Byte equality
    and working are different claims: `--decrypt-check` writes the **escrowed** bytes to a
    throwaway 0600 identity and decrypts a real artifact with them.
-   🔴 **Its unlocked path has never run.** See "The one link still untested".
+   🔴 **Its unlocked path ran once (2026-08-25) and did NOT settle the question** — it
+   compared against a client key an env var had redirected it to. The escrow's agreement
+   with the real `age.key` is **still unconfirmed since the 08-23 `cmp`**, and that older
+   claim is neither confirmed nor refuted by the 08-25 run. See "The one link still
+   untested" — and do not re-escrow on that run's advice.
 
 ### Verified live, end to end
 10 scopes, 201 commits compared, restored from the bucket → `age -d` → `git clone` →
@@ -271,11 +275,17 @@ systemctl --user list-timers analyze-service-index-backup.timer --all
 systemctl --user show analyze-service-index-backup.service -p Result
 
 # the escrow is still there, still correct, and still WORKS (needs the master password)
-nix-shell -p bitwarden-cli jq --run \
-  'export BW_SESSION="$(bw unlock --raw)"; python3 ~/workspace/devrc/scripts/analyze-service-index/escrow-verify.py --decrypt-check --host workbench-$(cat /etc/machine-id)'
+# 🔴 THE SHELL MUST CARRY minio AND THE IDENTITY MUST BE PINNED — see "The env
+# redirect" below. Both omissions cost a real run on 2026-08-25.
+nix-shell -p bitwarden-cli jq 'python3.withPackages(p:[p.minio])' --run \
+  'export BW_SESSION="$(bw unlock --raw)"; env -u SOPS_AGE_KEY_FILE -u ASIB_AGE_IDENTITY python3 ~/workspace/devrc/scripts/analyze-service-index/escrow-verify.py --decrypt-check --identity ~/workspace/homelab-talos/.secrets/age.key --host workbench-$(cat /etc/machine-id)'
 # no password to hand? this is still a real check — it must exit 12 VAULT-LOCKED,
-# promptly and without hanging, NOT sit waiting on an invisible prompt:
+# promptly and without hanging, NOT sit waiting on an invisible prompt.
+# MEASURED 2026-08-25: rc=12 in 2s.
 nix-shell -p bitwarden-cli jq --run 'python3 ~/workspace/devrc/scripts/analyze-service-index/escrow-verify.py'
+# and BEFORE spending a password, confirm which file the run will actually
+# compare against — no vault, no network:
+python3 ~/workspace/devrc/scripts/analyze-service-index/escrow-verify.py --print-plan
 
 # the index store's own verdict
 python3 ~/workspace/devrc/scripts/subsystem-audit.py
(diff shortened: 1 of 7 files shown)


ZacxDev:
## Correction: this PR's original description made two false claims

Posting rather than silently editing, because a reviewer may have already read the wrong version. Both were found by an adversarial audit, not by the test suite.

**1. "when it was not the default they lead with a refusal to treat the mismatch as diagnosed" — false as written.** The refusal was *appended after* the remedy, so the rendered message read `…Re-escrow from the on-disk identity only after confirming…` **then** `🔴 READ THIS BEFORE RE-ESCROWING…`. The operator met the dangerous instruction first. The tests asserted substring presence and never order, so the name `test_a_mismatch_on_an_ENV_CHOSEN_identity_warns_BEFORE_the_remedy` claimed coverage its body did not provide. Now genuinely ordered, and pinned by index comparison rather than by presence.

**2. The sources table listed three rows; there are four.** `--identity` was silently folded into the environment-redirect branch, which is what produced the self-contradicting advice below.

## And the substantive finding the audit caught, which the body never claimed at all

The first revision branched on **what named the identity file** instead of **which file it was**, and so printed demonstrably false sentences on two ordinary invocations — including the one the handoff doc in this very PR recommends:

```
$ escrow-verify.py --print-plan --identity ~/workspace/homelab-talos/.secrets/age.key
chosen by: the --identity flag  <- NOT the default; an unrelated shell can set this
```

That path *is* the default, and no shell can set a command-line flag. The same line fired for `SOPS_AGE_KEY_FILE=<the default path>` — what the deployed backup unit itself exports — making it a permanently-red warning on the normal configuration.

For a module whose stated purpose is never to make a confident wrong claim, shipping a new confident wrong claim in the very sentence added to prevent one is worth recording rather than quietly fixing. The guard was *spelled* rather than *structural*: it asked "was an env var set?" when the discriminating state is "is this a different file?".

Both are fixed in `6225610a`, reproduced live before and after. The audit's seven mutation survivors are now killed 10/10 by a semantic (non-deletion) sweep.

PR #829, opened 2026-08-25
feat(present): a generated explainer page for the agent layer — every number measured, every absence rendered

ZacxDev:
## What this is

`scripts/present/` — a **generator** that emits **one self-contained HTML file** explaining devrc's agent-facing layer to a competent engineer with no context, at the depth needed to change it correctly.

```
python scripts/present/generate.py -o present.html            # full, local identifiers
python scripts/present/generate.py --sanitize -o public.html  # shareable, same generator
python scripts/present/generate.py --check                    # measure + report, write nothing
```

It is a **measurement, not a document**, and that is the whole design. This repo's prose has been measurably false *in both directions* about its own system — the merge-gate line in `CLAUDE.md` read *"CI gates both suites"* and later *"NO AUTOMATED GATE IS RUNNING"*, and neither was true when read. A page that restated those numbers would inherit the decay.

So **no quantity on the page is typed**. Each comes from a measurer in `present/measure.py`, is stamped with the moment it was taken — **per row**, because a byte count ages in minutes and a systemd timer roster ages in days — and carries the command that would settle it independently. Where a test already owns a constant (`test_rules_size.py`, `test_skill_tiers.py`, `run-tests.sh`'s `TARGET_FLOORS`), the generator **reads it from that owner** and never restates it.

That separation is enforced mechanically, not by convention: `test_the_content_module_never_restates_a_live_measured_number` takes every live measured value, pulls its multi-digit tokens, and fails if any appears in the prose module. *Historical* measurements — a killed proposal's numbers, a dated incident — are deliberately out of scope for that guard, because they cannot go stale.

## `UNMEASURED` is a first-class state

Borrowed directly from `drift-check.sh` rc 18: *"we could not look"* correctly set no code, and therefore escalated **never**, so the run read clean while rc 17 was structurally unable to fire.

- A measurer that raises produces a **row** carrying the reason and the settling command — never a gap, never a blank. An omitted row is byte-identical to one that measured clean.
- A build where **everything** came back unmeasured **exits 3 and writes nothing**. That page looks careful and is broken.
- **Some** unmeasured rows are correct and are not a failure — a host without a given surface, an offline build. Failing on those would train everyone to pass `--force`.

One row is **permanently** unmeasured on purpose. A request count for the hosted store API was reported to me as *"309, all from its own build session"*. **That number appears nowhere in this tree.** Rendering it would restate a claim as a measurement, so the row exists, says UNMEASURED, gives the reason, and hands over the check. A test pins that it stays unmeasured against the real tree.

## Self-contained by construction

The hosted page and the portable export are the **same file** — no second renderer, nothing to keep in sync. Diagrams are **hand-authored inline SVG**: a mermaid `<script>` is disqualified outright, and rendering mermaid at build time would turn a missing toolchain into a silently thinner page.

The check scans the **output bytes**, not the generator's intentions — "we never write a script tag" is a claim about the code and stays true right up until someone adds a diagram helper. Its negative control is built from **realistic** references (a CDN chart library, a font stylesheet, a `srcset`, an `@import`), because a scanner fed its own canonical example scans clean. The one allowed external-looking token is the SVG XML namespace, allowlisted explicitly, with its own test so the scanner cannot become permanently red.

## Public-repo separation

The generator is committed; the data is not. Real scope names, hostnames, addresses and paths are read **at run time from local state** and appear in no fixture. `--sanitize` swaps them for **stable** synthetic stand-ins. The legend prints only the stand-in side — a two-column legend would re-publish exactly what the sanitized artefact exists to remove.

The module states plainly that it is a **redaction aid, not a security boundary**, and `test_the_sanitizer_cannot_see_an_unknown_identifier_class` pins that limitation *as a property*, so the suite cannot be read as a guarantee it does not make. No `.html`/`.json`/`.txt` artefact is committed under `present/`, and a test enforces that too.

## Four defects this work found in itself

Each by a control, not by review — and three of them only by **rendering the page and reading it**:

1. **`m_hook_gate_install` could never fail.** `git config --get` exits non-zero for "unset" and for "not a repository" alike, so it reported a confident `(unset on both scopes)` for a directory that is not a checkout — and, being unable to fail, it single-handedly made the **all-unmeasured build verdict unreachable**. Caught by the negative control that drives the CLI at an empty directory.
2. **Constants were read by importing their owning test modules**, so an owner that does `import pytest` came back UNREAD under a bare interpreter. An absence caused by the *measurement technique* rather than by the tree — the worst kind, because it looks like a finding. Now parsed with `ast`.
3. **The HTTP-client predicate matched the bare word `requests`**, scoring two local store readers as HTTP-capable off the phrase *"pull requests"* in their own docstrings — a measured row that directly contradicted the section it was rendered under. Now an import statement, with both controls pinned.
4. **The sanitizer missed path-mangled project slugs** (`-home-<user>-workspace-…`), which are neither a home path nor a known store scope, so client and repo names reached a *sanitized* page while every substitution test stayed green.

## Coverage, honestly labelled

`present/` is new in this PR, so **most guards cannot be shown red at the base ref** — at that ref the module does not import. Those are labelled **INVARIANT GUARDS** in their own docstrings and are **not counted as regression coverage**.

The exceptions are named individually: the three `UNMEASURED` negative controls, the two rendering bugs (backticks rendering literally, escaped sub-heading markup), the project-slug leak, and the constant-reader. Each was red before its fix.

An **8-mutant battery** kills every guard by **its own** assertion — not merely "the suite went red". Run under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` purged between mutants, because a same-length edit landing in the same second is invisible to CPython's mtime-in-whole-seconds cache and would be scored SURVIVED without ever executing. A no-op mutation **SURVIVES**, which is the control proving the battery is not just always-red.

| mutant | outcome |
|---|---|
| prose restates a live measured number | KILLED by its own test |
| `verdict()` always says `ok` | KILLED |
| a raising measurer is dropped instead of reported | KILLED |
| the page loads an external script | KILLED |
| the sanitizer stops substituting scopes | KILLED |
| the sanitizer rewrites everything, flag or no flag | KILLED |
| an UNMEASURED row renders without its reason | KILLED |
| a long table truncates instead of folding | KILLED |
| **no-op comment (harness control)** | **SURVIVED** |

## What the page covers

The full spine is written — **no section is stubbed**. Sections 0–13: how to read it, entry paths, the three budgeted always-on surfaces, the hooks and integrations, the two gate tiers, shipping, drift, observation, invariants, the cost model, **negative space**, the evidence bar, the soft seams, and a glossary.

Two sections are the ones worth reading first if you are reviewing:

- **§10 Negative space** — things already tried and killed, each with the measurement that killed it, including the case where the reopen gate later tripped and the document declared its own gate the wrong question.
- **§11 The evidence bar** — the worked example where six recommendations were produced by tracing code and four did not survive a single `grep` or `curl`.

**§12 is analysis, not a backlog.** It describes each seam and what would settle it, and mints no work objects — which is the object-leak rule applied to the page itself.

## Scope, and what remains

Nothing nix-side was touched, and `homelab-talos` was not touched. The generator, page, sanitizer and tests are entirely inside this repo.

**Remaining for hosting** (deliberately not done here): a nix derivation or timer that runs the generator and drops the output somewhere served, plus the k8s side. Two properties make that straightforward and one makes it interesting:

- the generator writes one file with no assets, so serving is a static-file concern;
- `--sanitize` gives the shareable variant from the same code path;
- but the measurements read **local machine state** (`~/.claude/`, `systemctl --user`, the index store), so a build inside a container will legitimately render several rows UNMEASURED. That is correct behaviour, not a bug to work around — the row will say which surface the build could not see.

## Gate

Reported in the PR thread by tier and base sha. Tests live in `scripts/tests/`, an existing floor target, so no new `HERMETIC_TARGETS` entry is needed and no floor has to move.


ZacxDev:
## Gate — both tiers, named

Run at branch head `8da6ac15`, base `199774f8` (`origin/main` at the time the branch was cut). Both tiers were run because they are **two different environments, not two spellings of one thing**.

**Dev-host tier** — `nix develop --impure --command bash scripts/gate.sh --tier both`

```
GATE: RESULT=PASS exit=0

pytest  TOTAL collected=15899  passed=15897  skipped=2  failed=0
        (floor: 13753 = sum of 27 per-target floors)
node    TOTAL suites=4 files=36 tests=1219 pass=1219 fail=0 skipped=0
        (floor: 1189)
```

**Sandbox tier** — `nix build .#checks.x86_64-linux.{pytests,nodetests}`

```
devrc-pytests>   TOTAL collected=15899  passed=15897  skipped=2  failed=0  (floor: 13753)
devrc-pytests>   RESULT: PASS (exit=0)

devrc-nodetests> TOTAL suites=4 files=36 tests=1219 pass=1219 fail=0 skipped=0  (floor: 1189)
```

Verdicts read from the runners' `RESULT:` **content**, not from a pipeline's exit status — the sandbox runs were captured through a `tail`, whose status is the one a naive `$?` would report.

**Positive control on the count.** `scripts/tests` collected **7947** in both tiers; the three new files collect **62** of those, so they are inside the number and not merely adjacent to it. Skips are the two already pinned in `EXPECTED_SKIPS` — this change adds none.

**No floor moved.** The tests live in `scripts/tests/`, an existing target, so no `HERMETIC_TARGETS` entry and no `TARGET_FLOORS` edit was needed.

### What these verdicts do *not* claim

- They are a claim about **this branch at this base sha**, not about the tree a merge creates. `strict` is false on this repo's branch protection by design, so gating the merged tree is still a manual step.
- The dev-host run happened on a box carrying several other agents' suites concurrently (load ~20–30 on 24 cores). It went green, so load is not a confound here — but the wall times in it are inflated and should not be read as a performance baseline.

### Mutation battery (separate from the gate)

Eight mutants, each killed by **its own** assertion rather than by "the suite went red", plus a no-op control that **survives** — which is what shows the battery is not simply always-red. Run under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` purged between mutants, because a same-length edit landing inside one second is invisible to CPython's mtime-in-whole-seconds cache and would score SURVIVED without ever executing.

PR #839, opened 2026-08-25
docs(handoff): discord-embed-ext rescue — all three PRs merged, both hosts deployed and verified, and the duplicate resolved itself onto main

ZacxDev:
`/handoff` update to `claudedocs/handoff-discord-embed-ext-rescue.md`. Landed from a worktree because devrc forbids committing to `main`.

Everything the previous version listed as open is now **merged and deployed**, so the status sections are replaced and the findings appended (`buckets:` confirmed: `State now`/`Next steps`/`How to verify` → REPLACE, `Open investigations`/`Gotchas` → APPEND, so the earlier text survives verbatim).

What changed since the doc was written:

- **#804** merged (squash `eaf68c96`) — the extension, 11 files on `main`, SUITES `|2|113`
- **#832** merged (squash `4500b88c`) — the worktree mandate now names ANY agent with a runtime-neutral mechanism
- **Both hosts deployed and verified by content** at `f82272c3`, cross-host agreement **CONFIRMED** — including the reworded rule reaching `~/.config/opencode/AGENTS.md`, which was the whole point of #832
- **clawgate task 357** filed for the queue-collision cause
- The concurrent rebuild **resolved itself by being committed to `main`** in the shared clone (`74f3a389`), diverging it 1-ahead/9-behind and blocking `ship.sh` — recovered by preserve → verify → `git reset --keep`, preserved on `rescue/discord-embed-ext-committed-to-main`

One durable line was explicitly carried forward rather than let the REPLACE drop it (the nix deploy path and the 118/113 counts) — flagged in the doc itself so the next reader knows it was deliberate.

Only untested claim remaining, and it's stated as such: the manual `brave://extensions` → Load unpacked step, on a Discord channel with a real image attachment.

### DIFF
--- a/claudedocs/handoff-discord-embed-ext-rescue.md
+++ b/claudedocs/handoff-discord-embed-ext-rescue.md
@@ -22,27 +22,42 @@ dead session. Outcome: preserve everything, land the salvageable half properly, 
 the part that was measurably dead.
 
 ## State now
-- **Branch:** `feat/discord-embed-enlarge`, head `8635f168`, 10 commits.
-- **PR #804 — OPEN, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.** Both required
-  checks green: `tekton/devrc-nodetests` SUCCESS, `tekton/devrc-pytests` SUCCESS.
-  Merges clean into `origin/main` `324693fd` (re-checked at handoff time; main moved
-  four times during this work).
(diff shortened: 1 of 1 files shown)


ZacxDev:
Stale — everything described as open is already merged and deployed. The handoff doc is superseded by #818 and #838.

PR #841, opened 2026-08-25
perf(gate): run each pytest target 4-way under xdist — ~2x on the whole gate

ZacxDev:
## What

Run each pytest target 4-way under `pytest-xdist`. Measured on the whole hermetic gate, one host, wall clock:

| run | wall | mode |
|---|---|---|
| pre-change baseline (devrc `main` checkout) | 1194s | serial |
| this branch, `DEVRC_TEST_JOBS=1` (same worktree) | 1446s | serial |
| this branch | 532s / 576s / 585s | parallel |

🔴 **The two serial runs are 21% apart**, so ~20% is this host's run-to-run variance and no single pair licenses a precise figure. The only *controlled* comparison — same worktree, same HEAD, jobs=1 vs jobs=4 — is 1446s vs 532–585s (**2.5–2.7×**); the conservative cross-checkout pair is 1194s vs 585s (**2.04×**). Honest range: **2.0–2.7×**, quote ~2×.

## Why xdist and not a parallel target loop

The run's cost is concentrated, so the two strategies are not interchangeable. Per-target spans, serial:

| target | secs | % of run | tests |
|---|---|---|---|
| `scripts/tests` | 677 | **57.0** | 7,924 |
| `scripts/browser-bridge/tests` | 234 | 19.7 | 790 |
| `scripts/dl-router/tests` | 126 | 10.6 | 1,005 |
| the other 24 targets | 123 | 10.3 | combined |

Running all 27 targets concurrently cannot finish sooner than the longest single target — 677s. That is a hard **1.75× ceiling that `-P2` already reaches and `-P8` does not improve on**. The parallelism has to be *inside* the dominant target, which under xdist goes **677s → 223s**.

`--dist loadfile`, not the `load` default: a file's tests stay on one worker. Several suites here share module-level state and `load` would scatter a single file across workers and race them.

## The three things this had to fix, none of them optional

**1. Nested runs (`DEVRC_TEST_NESTED`).** Tests in `scripts/tests` *spawn this script* — `test_run_tests_floors.py` and the `.sh` meta tests run a nested `run-tests.sh`. A parallel nested run makes its guard plugins emit one session marker per worker, and GUARD 7/8/10 fail it for the doubled count. That is the guards working correctly on a condition the parallelism created. Passing `-n` through the **environment** (`PYTEST_ADDOPTS`) reproduces it exactly: 6 failures, all meta-tests, all "emitted 2 session marker(s)". So `-n` goes on this script's own pytest line, and each invocation exports `DEVRC_TEST_NESTED=1` to force `jobs=1` in any nested run.

**2. Marker accounting, GUARD 7/8/10.** All three now go through **one** shared predicate instead of three open-coded copies — they had already drifted, GUARD 9 saying `>= 1` for its own double-registration case while these three said `-ne 1`.

- serial: **exactly 1**, unchanged
- parallel: **1..jobs+1**, because `pytest_sessionstart` fires in the controller and in each worker
- **zero still fails** — "the plugin never loaded" is the state these guards exist to catch

**3. GUARD 10's config control was genuinely broken under parallelism, not just miscounted.** Every worker fires a real `git config --global` at the *same* guard file, and git takes an exclusive `<cfg>.lock` per write, so the losers exit 255. Measured directly, outside this suite: 8 concurrent writes to one file → 4 exit 0, 4 exit 255 with `error: could not lock config file: File exists`. Before the retry, GUARD 10 reported `control=unmeasured` for **16 of 27 targets** — it was correctly refusing to score a pass it could not demonstrate.

The retry is deliberately narrow: 6 attempts, pid-jittered backoff, **only** when stderr matches the lock message. Any other non-zero exit still returns `unmeasured` on the first try. Widening it to all failures would convert a broken control into a slow green, which is the exact failure that guard exists to prevent.

Plus GUARD 1c: `pytest-xdist` added to the probe under its **module** name (`xdist`), with the attr→module pair in `NIX_TO_MODULE`. The two-way pin still holds in both directions.

## Verification

- **Green:** `rc=0`, 15,887 passed / 0 failed / 2 skipped, 27/27 targets, **0 GUARD problems**.
- **Every guard above was watched to FAIL first, then go green.** The suite caught three of my own mistakes en route: a comment inside `gatePyEnv`'s list (GUARD 1c parses it textually, so the words became package names), the unregistered dependency, and the lock-contended control.
- **Mutation test on the widening** — dropped `-p testlib.nogit_plugin` from the pytest line: `rc=1`, **GUARD 10 alone** fired, with *its own* message, `emitted 0 session marker(s), expected between 1 and 5`, on 26 targets. So the widened predicate still catches a plugin that never loads, and the mutant did not die from some unrelated guard's error.
- **`DEVRC_TEST_JOBS=1` escape hatch:** full serial run green — `rc=0`, same 15,887 passed, 0 GUARD problems, and **0 xdist nodes started** against 54 in the parallel run as the positive control. This matters because `_markers_ok` takes a *different* branch when `jobs == 1` (the strict `== 1` path), and nothing else exercises it.

## Scope note for the reviewer

This is not just a flag. It edits a safety-guard plugin and three guard assertions. Every change keeps **zero as a failure** — a plugin that never loads, or a control that never fires, still goes red — but that judgement is the thing to review here, not the speedup.

## Downstream, not in this PR

`devrc-ci`'s `pytests` step in homelab-talos requests 1 CPU / 2Gi against a measured ~1-core, 823Mi peak. Four workers raise both, and the CI node is already at 93% memory *requested*, so that request needs re-measuring and revising before this lands its full benefit in CI. Tracked separately.


ZacxDev:
## ⛔ Do not merge — adversarial audit found two fail-open regressions, and part of this PR's description is wrong

CI is green and the speedup is real (measured in `devrc-ci`: `pytests` **22.4m → 12.6m**, 1.78×). Neither fact rescues the change, because **the green run includes a guard this PR silently disarmed.**

### Correcting this PR's own description

The body claims the marker bound is `1..jobs+1` because "`pytest_sessionstart` fires in the controller AND in each xdist worker". **That mechanism is false.** The nolaunch/spool/nogit markers come from **session-scoped autouse fixtures** (`nolaunch_plugin.py:167`, `spool_plugin.py:218`, `nogit_plugin.py:342`), which never run in the controller — it collects no tests. The measured count at `-n 4` is exactly **4**, never 5. So the upper bound is off by one *in the permissive direction*, and the operator-facing string `"between 1 and N (controller + up to N xdist workers)"` states a mechanism that does not happen.

I'm correcting this in a comment rather than editing the body, since anyone who has already read it read the wrong version.

### 🔴 1 — GUARD 9 downgrades ENFORCE → REPORT under xdist

`testlib/gitenv.py:688` (`live_cotenants`) excludes only the caller's **own ancestor lineage**. Sibling xdist workers are not ancestors and their cwd is `$ROOT`, so each worker names its three siblings as proof of a foreign writer, `_effective_mode()` returns `report`, and `_report()` never calls `pytest.fail`.

Controlled A/B on a clean throwaway repo, identical but for `-n 4`:

| run | rc | result |
|---|---|---|
| serial | **1** | `122 passed, 2 errors`, culprit named: `test_aa_violator.py::test_aa_violator` |
| `-n 4` ×3 | **0** | **green**, 3 reproductions of 3 |
| `-n 4` + `DEVRC_GITENV_MODE=enforce` | **1** | confirms the auto-downgrade is the cause |

A repo mutation reverted before session end — a fixture that cleans up after itself, the ordinary shape — now passes green. `run-tests.sh` never parses the one trace that remains (`unattributed-observations`).

**This is invisible on the dev host** (8 pre-existing cotenants there mean GUARD 9 was already in report mode), so it bites exactly where the guard was still sharp: `devrc-ci` and any clean checkout. That is why my own green run could not have caught it.

### 🔴 2 — `head -1` leaves 75% of the per-session controls unevaluated

`run-tests.sh:3560-3566` and `:3382-3388` read `control=` / `redirect=` / `protocol=` from the **first** marker line only. Pre-PR, first == only. At `-n 4` there are 4 lines, and which one is first is an append race.

That is the same fail-open the new retry loop exists to prevent, reintroduced one layer up: a worker exhausting its retries reports `control=unmeasured` and the target still scores **PASS** unless that worker happened to write first. The guard's header says it "correctly refuses to score a pass it could not demonstrate" — under xdist it now scores three undemonstrated passes per target.

### Also landing in the rework

- The loosened `_markers_ok` has **no test at all**, and my mutation only probed the far end of the lower bound. Surviving mutants include `-ge 1` → `-ge 0` — zero markers pass, the exact state all three guards exist to catch.
- `test_no_real_launchers.py:536-539` predicted this verbatim: *"Under xdist this would race; the fix then is a per-worker stub dir, not a looser assertion."* ~15 exact-delta assertions on a run-wide log. Not reproduced in 2/2 attempts — a small-probability flake on a required gate, which is worse than a loud one.
- `test_transcript_search.py:525` justifies a process-global `TZ` mutation with "the runner is single-process and sequential" — this PR makes that sentence false and did not update it.
- `LOCK_CONTENTION` matches git's **localised** English string with no `LC_ALL` pin; `DEVRC_TEST_NESTED` is never cleared at the root of a run, so an ambient `=1` silently reverts the gate to serial; `PYTEST_JOBS` is never logged.

Full findings, including nits, are in the audit. Reworking.

ZacxDev:
## Rework complete — three audit passes, four fix rounds. Ready for human review, not for an automatic merge.

Every 🔴 from all three audits is closed. Summary of what each pass found, because the pattern is the useful part: **a green gate missed a fail-open defect three times running, and two of the three were in code added specifically to prevent fail-open.**

### Pass 1 — the change reviewed the guards it EDITED, not the ones it BROKE

| finding | resolution |
|---|---|
| 🔴 GUARD 9 disarmed as a side effect: sibling xdist workers counted as foreign co-tenants, dropping every worker from ENFORCE to REPORT | fixed — sibling test on ppid |
| 🔴 `head -1` in `_nogit_field` + 3 spool readers evaluated 1 of N per-session controls | fixed — DISTINCT across all sessions |
| 🟡 marker bound off-by-one *permissive*, with a stated mechanism that was false | fixed — `1..JOBS`; the controller emits no marker |
| 🟡 both predicates had zero tests | fixed — 20 tests, mutation-swept |
| 🟡 locale-dependent match, weak jitter, 180s worst case | fixed — `LC_ALL=C`, full jitter, 30s bound |
| 🟡 launch-log race | **open**, documented at the site |

### Pass 2 — the fix round's own fixes were wrong

- 🔴 **`_is_sibling_xdist_worker` was INERT — it could never return `True`.** It read the run id from `/proc/<pid>/environ`, the *exec-time* snapshot, while xdist assigns the id at *runtime*. The docstring called this "MEASURED": the id was measured in `os.environ`, and the code then read a different surface. Now ppid-only. A/B with four workers sitting in the protected repo: `cotenants=[3 siblings]` → `cotenants=[]`.
- 🔴 **GUARD 9b would red a run whose own log proved an external writer.** `unattributable=` is written by one probe at import time; the idle probe and settle re-read record a foreign writer as `gitenv(foreign-writer)` and touch no such field. A writer whose cwd is outside the repo — `drift-check`'s `git fetch`, 4×/day — is invisible to the first and caught by the others.

### Pass 3 — the guard I added could not fire at all

🔴 **GUARD 9b's `FAIL` branch was unreachable**, and is now **removed**. In auto mode (the only mode this runner permits) `unattributed-observations>0` requires report mode; report mode requires `_UNATTRIBUTABLE` non-empty; and that prints either `unattributable=k>0` or a `gitenv(foreign-writer)` line — both of which were the check's own excuse conditions. Every log that could trip it carried one.

It read as coverage for two rounds and produced a measured false red before that. Removing it also deletes two findings that existed only because of it. The protection that actually works was always upstream: with siblings correctly excluded, workers stay in ENFORCE and fail the *test* that touched the repo, by name.

🟡 **The inert-predicate class is now pinned.** The audit showed `return False` — the exact bug that shipped twice — still surviving the whole suite. `test_gitenv_sibling_exclusion.py` kills it, plus `return True`, own-pid-vs-parent, and the inherited-run-id case.

🟡 **"the run id is in my environment" ≠ "I am a worker"** — a worker's children inherit it. Fixed using the same `/proc` quirk that caused the original bug, deliberately: a genuine worker has the id in `os.environ` and **not** in `/proc/self/environ`; an inheritor has both.

### Still open — deliberately, for a human

1. **GUARD 9's `>= 1` marker control is satisfiable by the controller alone** under xdist (measured: exactly one `gitenv(session)` line reaches the target log, and it is the controller's). It therefore no longer demonstrates the detector loaded in the sessions that *run tests*. Pre-existing, but this branch rewrites the reasoning beside it. Closing it means making worker output reach the log — a design change.
2. **The launch-log race** in `test_no_real_launchers.py`: ~15 exact-delta assertions read a run-wide log sibling workers append to. Not reproduced in 2/2 attempts. The fix is the per-worker stub dir the original comment named; it needs the runner's line-offset ledger model changed, so it is not a patch.

### Numbers

- Local gate: **~2×** (serial 1194s / 1446s across two runs — ~20% host variance; parallel 532–625s). Quote ~2×, not a precise figure.
- **CI: `pytests` 22.4m → 12.6m (1.78×)**, measured on this PR's own `devrc-ci` run.
- Final: `rc=0`, 15,907 passed / 0 failed / 2 skipped, 27/27 targets, 0 GUARD problems.

### What I am asking a reviewer to check

Not the speedup — the **guard semantics**. This branch changes the attribution logic of a safety plugin, widens three marker assertions, and deletes a guard. Every change keeps zero as a failure and each is mutation-tested, but "was it right to touch these at all" is a judgement I should not make alone, and three audits' worth of evidence says my confidence about this code has been wrong more than once.

ZacxDev:
Post-merge correction: **this PR's body describes a looser contract than the code that shipped.** Both points below were corrected during review; the body was not updated. Recording it here so nobody reasons from the description instead of the source.

The drift is in the *permissive* direction, which is the dangerous one — someone reading this body believes the guard is slacker than it is, and could "fix" a correct rejection as a false alarm.

## 1. The parallel marker bound is `1..jobs`, not `1..jobs+1`

The body says the bound allows `jobs+1` "because `pytest_sessionstart` fires in the controller and in each worker."

Shipped (`scripts/run-tests.sh:2991`, five call sites at `:3497`, `:3566`, `:3572`, `:3855`, `:3861`):

```sh
_markers_ok() {
  local n="$1"
  if [ "$PYTEST_JOBS" -eq 1 ]; then
    [ "$n" -eq 1 ]
  else
    [ "$n" -ge 1 ] && [ "$n" -le "$PYTEST_JOBS" ]
  fi
}
```

The ceiling is `PYTEST_JOBS`. Markers come from session-scoped autouse fixtures, so the controller — which runs no tests — never emits one. Measured: `-n 1` → 1 marker, `-n 2` → 2, `-n 4` → 4. A mutant raising the ceiling to `JOBS+1` dies with `5 marker(s) returned accept, expected reject`.

## 2. Nesting keys off `PYTEST_CURRENT_TEST`, not a `DEVRC_TEST_NESTED` export

The body describes handling nested runs by exporting `DEVRC_TEST_NESTED=1`. That is not what shipped, and an in-file comment explicitly repudiates the env-var approach as self-contradicting.

## What is NOT wrong — the guard is still non-vacuous

Worth stating, since the above reads as a downgrade and is not:

- **Zero is still rejected.** Mutating `-ge 1` → `-ge 0` fails 2 tests with `_markers_ok at PYTEST_JOBS=4 with 0 marker(s) returned accept, expected reject`. `scripts/tests/test_markers_ok_predicate.py` pins zero-rejection at jobs 1/2/4/8/16.
- **Serial is still exact.** At `PYTEST_JOBS=1`, both 2 and 5 are rejected.
- It is an exactness check in serial and a bounded range in parallel — never an existence check.

## The one guarantee genuinely traded, and why it is forced

At `-n 4` the predicate accepts 1–3 markers, so "the plugin loaded on only one of four workers" would pass. That slack is **not** gratuitous: a single-file target at `-n 4` emits exactly 1 marker, because `--dist loadfile` idles the other three workers. Requiring `-eq JOBS` would false-fail every small target. The looseness is imposed by the distribution mode, not by the guard's design — and a guard that fails a correct configuration is how a guard gets deleted.

## Verification of the rollout itself

Collected **and executed** counts are preserved under `-n 4 --dist loadfile` (collect-only cannot see a worker dying mid-run, so both were checked):

| target | serial | `-n 4 --dist loadfile` |
|---|---|---|
| `scripts/dl-router/tests` | 1005 passed / 283.96s | 1005 passed / 107.15s |
| `scripts/collector/i3/tests` | 13 passed | 13 passed |

`--check-floors` and `--check-targets` both PASS.

Also worth recording for whoever meets it next: the collection-mismatch class fixed in #855 is **loud, not silent** — a planted module-scope `mkdtemp()` in a parametrize ID gives `rc=1` and `Different tests were collected between gw1 and gw3`, while the same suite is **green serially**. It is a parallelism-only failure that no serial run can observe.

PR #849, opened 2026-08-25
retire the hosted subsystem-store-api — 309 requests, all from the session that built it

ZacxDev:
## What

Deletes the hosted `subsystem-store-api` — an HTTP read layer over the recall store — and its
test suite. Adds the decision record:
**`claudedocs/decision-subsystem-store-api-retired-2026-08-25.md`**.

🔴 **The LOCAL index store is untouched and fully alive.** `scripts/lib/subsystem_recall.py`,
`subsystem_touch.py` and `subsystem_resolver.py` were never part of this. Only the hosted HTTP
layer is retired.

## Why — measured, not assumed

The server audits every `/api/*` request including successes (`/healthz` is the one deliberate
exemption), so an absence of 200s is evidence rather than a gap in instrumentation. Read from
the live pod:

| Measurement | Value |
|---|---|
| Successful authenticated requests, pod's entire life | **309** |
| Distinct days | **1** (2026-08-20, 19:07–23:45) |
| Distinct source IPs | **1** — the session that built it |
| Legitimate requests since | **0** (including last 24h) |
| Everything else | internet credential scanners, all 401 |

The 309 include a `search/nosuchscope` negative control — a build session exercising its own
service, not a consumer using it.

**No client was ever written.** The 2026-08-20 decision specified hosted as an "entry-level
advisory" consulted for a manifest. `subsystem_recall.py` has no hosted path, no advisory path,
no manifest, no `--source hosted`; no reader in this tree imports an HTTP client at all. The
handoff said so itself: *"THE MANIFEST DOES NOT EXIST YET."*

**The missing piece was a SYNC, not more server.** The hosted copy was seeded by hand, nothing
synced it, and it mirrored the workbench pile only — the two local stores were measured
*disjoint*. So the stated use case, reach the index away from the workbench, was precisely what
the deployment could not do. An availability mechanism was built for a consistency problem.

The record carries the reproduction command, the bar for rebuilding (measured demand **plus** a
sync, with the check named), and notes the audit log dies with the pod so the numbers are
reproducible only until teardown.

## 🔴 This PR and the k8s teardown are INDEPENDENT

The Kubernetes teardown is a **separate PR in `homelab-talos`**. Neither implies the other:
**removing the code here does not stop the pod, and removing the pod there does not remove the
code.** Land them in either order; until both land, one half survives.

## Collateral the removal orphaned

- **`test_hermetic_git.py`** `EXPECTED_MEMBERS` — two-way pinned ledger; drops the suite.
- **`rsync` dropped** from `run-tests.sh` `REQUIRED_TOOLS` and `flake.nix` `gateTools`. Its only
  *executing* consumer was the deleted suite driving `seed.sh`. Measured first: every remaining
  `rsync` under `scripts/tests/` parses the word out of a shell script or systemd unit and never
  invokes the binary, and `ship.sh`'s own rsync was already removed (and is pinned removed).
  `rsync` still resolves inside `nix develop` from the user profile — only the hermetic sandbox
  loses it. Both comment blocks say how to re-add it.
- **`present/measure.py`** — `seam.store_traffic` deleted: its stated premise ("no repo file
  records a request count") stops being true once this record lands, and the audit log, its only
  source, goes with the pod. `seam.store_api` repointed at the retirement, and now raises
  `Unmeasurable` when **no** reader file exists so a zero over zero readers is an absence, not a
  silent zero — which also keeps an empty tree all-unmeasured, the condition the generator exits
  non-zero on.
- **`present/content.py` §13** — the seam is described as closed, not open.
- **`test_present_measure.py`** — the traffic invariant guard is replaced by a live-tree
  regression guard (local readers hold no HTTP client) **plus a negative control** that feeds
  four HTTP-importing readers and watches the count move off zero.
- **`signal/Dockerfile`** — pointed at a header that no longer exists; argument restated inline.
- Dated superseding notes on the handoff, the hosting proposal, the entry-shape proposal and the
  cross-subsystem handoff. Nothing is rewritten — the exposure incident this service produced
  still owns a live CI gate in homelab-infra.

## Verification — all four, by tier

| tier | command | verdict |
|---|---|---|
| dev-host pytest | `nix develop --impure --command bash scripts/gate.sh --tier both` | **PASS** (exit=0) |
| dev-host node | same run | **PASS** — 1292 tests, 0 fail |
| sandbox pytest | `nix build .#checks.x86_64-linux.pytests` | **RESULT: PASS (exit=0)** |
| sandbox node | `nix build .#checks.x86_64-linux.nodetests` | **RESULT: PASS (exit=0)** — 1292 pass |

`GATE: RESULT=PASS exit=0`, base `0cbc69f9`. Verdict lines and per-target counts were read, not
exit codes alone.

**Floor re-pinned `8036 → 7772`** for `scripts/tests` — the number the gate itself printed
(`collected 7822 ... lower it to 7772`), copied rather than derived by arithmetic on two sides.

⚠ `strict` is false on branch protection, so this green is a claim about **this branch**, not the
tree the merge creates.

## Not done here

`scripts/README.md`'s header count ("435 tracked files") is stale — it was already wrong by 283
before this PR (HEAD had 718; the index now has 710), and its subsystem table omits several
directories unrelated to this change. Out of scope for a retirement PR. Closing condition is
mechanical: the number equals `git ls-files scripts | wc -l`.


ZacxDev:
Closing rather than merging — the store is being **kept and finished**, not retired. See clawgate task **360**.

The retirement case was sound on its own evidence (309 requests, all from the build session, no client). But the reason no client ever called it is that the client was never built: `_reject_write()` refuses every write verb, the route table is `/healthz` + `recall` + `search`, and **no client-side token exists on either host** — it lives only as the in-cluster secret. Phase 1 and 1.5 shipped a working, hardened server and phases 2–3 (the CLI and the append write path) were never started.

Measured 2026-08-25: pod `/data` holds 75 entries / 9 scopes seeded once on 2026-08-20; the workbench has 39 more, the laptop 25 more that exist nowhere else. The union a single datastore must hold is **19 scopes / 139 entries**. Task 360 ships the CLI and `POST /api/v1/entry/{scope}/{ref}/bullets` to close that.

Reopen this if 360 stalls.

PR #850, opened 2026-08-25
docs(handoff): the preemption fix is SETTLED (0/65, p=9e-9) — and neither "flaky test" was ever a flake

ZacxDev:
Records the re-run of the post-#396 devrc-ci kill-rate measurement — next-step 1 of `claudedocs/handoff-ci-flakes-and-misattribution.md`, and the only open question in that thread with a deadline attached, because it decides whether the scheduler-preemption fix is settled.

## The result

| window | killed | rate |
|---|---|---|
| pre-#396 (created < `2026-08-25T06:00Z`) | **30 / 121** | 24.8% |
| post-#396 (created ≥ `2026-08-25T06:00Z`) | **0 / 65** | **0.0%** |

P(0 kills in 65 | rate unchanged at 24.8%) = **9e-9**. 95% upper bound on any residual kill rate: **4.5%**, down from 24.8%.

The prior session measured `0/9` and correctly refused to call it — a clean 9 arises by luck ~1 in 13. At n=65 it does not.

## Why this is a measurement and not a hopeful zero

- **The instrument is validated by exact reconstruction, not plausibility.** The same classifier over the same window boundary reproduces the prior session's figure to the digit — `30/121 = 24.8%`, numerator *and* denominator. A zero from an instrument that has not been shown to reproduce a known non-zero is indistinguishable from an instrument wired to nothing.
- **Two independent discriminators agree.** Kills counted as any step exiting `255`/`137`; separately, `verdict exit=2` is the killed-step signature. Pre: 30 and 30. Post: **0 and 0**. Zero `255`/`137` on *any* step, not just the test legs.
- **The rival mechanism is refuted by measurement, not assumed away.** "0 kills because the post window was quiet" predicts lower contention. Measured: post is *busier* — median concurrent TaskRuns 14 vs 12, max **74 vs 44**; max overlapping `gitops-validate` TaskRuns **38 vs 30**; max overlapping `devrc-ci` gates **14 vs 9**; runs overlapping ≥10 `gitops-validate` TaskRuns **15 vs 12**. Equal-or-higher contention, zero kills.
- **No pruning confound.** 68 `devrc-ci` PipelineRuns and 68 gate TaskRuns in the post window — nothing GC'd out of the denominator.
- **The mechanism is live.** 212 pods checked: `devrc-ci` entirely on one node, `gitops-validate` entirely on another, zero cross-over.
- **Preemption did not move, it stopped.** `gitops-validate` is 0 killed in *both* windows (0/238 pre, 0/107 post) — it was the preemptor, never the victim.

## What this does NOT close

The genuine per-test flake class is **unchanged**: `verdict exit=1` at 14/116 (12.1%) pre vs 8/65 (12.3%) post. #396 closed preemption and touched nothing else, so next-steps 2–4 of the handoff stay open and stay real. The doc says so explicitly rather than letting a headline zero read as "CI is fixed".

## Verification

Docs-only. `test_doc_path_rot.py` (76 passed) and the four content gates — `test_no_captured_markup`, `test_no_captured_text`, `test_no_client_hostnames`, `test_no_public_ips` (227 passed) — run on the **dev-host tier** at base `79d83f1c`. The sandbox tier (`nix build .#checks…`) is what Tekton gates on and is not claimed here.

### DIFF
--- a/claudedocs/handoff-ci-flakes-and-misattribution.md
+++ b/claudedocs/handoff-ci-flakes-and-misattribution.md
@@ -210,26 +210,74 @@ read the root cause first; the design is what failed, not the code.
    count a run as KILLED if ANY step's `terminated.exitCode` is `255` or `137`, GENUINE if
    `verdict` exits non-zero without one. **Reproduce `30/121` on the pre-`06:00Z 2026-08-25`
    window first** — that is the instrument's positive control, and it matches to the digit.
-2. **Two flaky tests still UNDIAGNOSED**, both `devrc`, `scripts/tests/`:
-   `test_no_unallowlisted_public_ip_literal_is_committed` and
-   `test_the_module_root_is_load_bearing`. Neither has a wall-clock dependency in the sandbox
-   tier; either could have been a *correct* red on some branch. Logs were unrecoverable (pods
-   GC'd; commit statuses cap at 140 chars) — **but `broken-gate-tail` now exists**, so a
-   recurrence is readable.
-3. **Re-check `test_an_absent_origin_header_is_not_the_same_as_an_empty_one`** —
+2. ✅ **DONE 2026-08-25 — NEITHER OF THE "TWO FLAKY TESTS" WAS EVER A FLAKE.** Both were
+   already diagnosed AND closed, and — 🔴 the part worth carrying — **both were recorded as
+   closed in THIS DOC'S OWN LINEAGE and then carried forward as "UNDIAGNOSED" anyway.** The
+   hedge that was right (*"either could have been a correct red on some branch"*) sat next to
+   the answer for two days because nobody re-read the predecessor. Same misattribution class
+   this thread exists to fix, committed by the thread's own notes.
+   - **`test_the_module_root_is_load_bearing` — a DETERMINISTIC two-tier failure, fixed by
+     #732 (`5a2a7b21`, 2026-08-22).** It asserted `gitenv.py` sits inside a git checkout; the
+     authoritative runner's source is a `/nix/store` path with **no `.git`**, so it died on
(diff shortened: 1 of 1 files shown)


ZacxDev:
Second commit `8a802e6c` closes **next-step 2** as well.

Both tests listed as "still UNDIAGNOSED" were already diagnosed and fixed — and both were recorded as closed by **predecessors of this very document**, then carried forward as open anyway. Neither was ever a flake.

- `test_the_module_root_is_load_bearing` — a deterministic **two-tier** failure (asserted this checkout has a `.git`; the authoritative runner's source is a `/nix/store` path with none). Fixed by #732 (`5a2a7b21`). Re-verified live in `.git`-less trees: `5a2a7b21^` → **1 failed** with the verbatim CI assertion, `origin/main` → **2 passed**. The red arm is the negative control.
- `test_no_unallowlisted_public_ip_literal_is_committed` — a **correct red** on #762's first revision, which committed a preserved merge artifact carrying a prod IP six times into a public repo. Pure function of the tree; cannot flake. Controls: committed tree → **0** unpinned, same assertion path fed the leak file → **6**.

Two findings recorded rather than chased:
- `broken-gate-tail` **cannot name a failing test** (24/139 reports populated, but only 1 of 24 genuine failures — it is the kill instrument). The commit status names it, and that naming only starts at `a9337827` (2026-08-24T20:34Z) — which is why these two were *unattributable*, not unexplained.
- **Live and unexplained:** `1ca46a80` reports `FAILED: pytests — collected=16062 passed=16060 skipped=2 failed=0`. The arithmetic closes; no test failed and the leg still failed the PR.

Docs-only. Doc gates re-run on the updated file: 303 passed (dev-host tier).

PR #854, opened 2026-08-26
feat(resume-state): say when the SKILL you are executing is behind origin/main — the deployed copy, resolved by readlink -f

ZacxDev:
## The bug, measured

A skill is loaded **once**, from `~/.claude/skills/<name>/SKILL.md` — a nix-managed path that reflects the last `home-manager switch`, **never** `origin/main`, and that `git pull` does not move. Nothing in the system could say "the procedure you are executing has been superseded".

**Measured 2026-08-25**: a session invoked `/resume` at 17:19 and carried out its step 6 — claim a work item with `gh pr list`, push the branch early. That step had already been replaced by `scripts/claim-work.sh` in **#847 (`200e6383`, 19:39)**: *"claim-work is the lock. This is a COMMAND, not a habit."* The session followed a stale copy of its own instructions and never noticed.

## What changed

`scripts/resume-state.sh` grows a `SKILL` block, printed **first** — it is a claim about the INSTRUCTIONS, so it has to be read before the findings those instructions produce.

- **The operand is the DEPLOYED file**, not the working tree: that is what Claude actually read.
- **`readlink -f` is the arbiter**, and it gives a **three-way** answer: a checkout (`mkOutOfStoreSymlink`, live), `/nix/store` (a copy only a switch replaces), or **neither** — a hand-placed foreign file, which `home.file.force` does *not* clobber, so it is labelled UNMANAGED rather than told to run a switch that cannot work. A **dangling** link (46 of 139 managed links on the laptop once, into a GC'd store path) gets its own diagnosis instead of reading as "never deployed".
- **Direction and size, not "differs"**: `N commit(s) BEHIND origin/main for claude/skills/resume/SKILL.md (newest it lacks: <sha> <subject>)`, computed by walking the commits that touched the path on `origin/main` and locating the deployed blob.
- **The two "not current, no number" outcomes each say only what was measured.** A walk that hits its cap reports `older than the newest N commit(s) … the scan stopped at its cap, so the DISTANCE was not measured`. Content matching no commit reports `built from a tree that is not on origin/main (uncommitted, **or a branch that has not merged**)` — the second case is what `home-manager switch --flake ~/workspace/devrc` off a feature branch produces, which CLAUDE.md recommends. Both sentences are pinned whole and normalised.
- **Every way it cannot measure prints `COULD NOT MEASURE` and a `!` GAP** in the existing gap block's grammar, and withdraws the all-clear. There are **seven** reasons and they are **scraped out of the script** by a test that requires `SKILL.md` to name each — modelled on the sibling clock sweep, which exists because this drifted once before. A silent pass here would *be* the defect.
- **Uncommitted edits on a live copy are NOT called stale** — same fork the handoff check makes — but they are still reported as a gap, because running unpushed instructions is its own trap.
- **Default `resume`; `RESUME_STATE_SKILL` overrides** with a space-separated list. `${RESUME_STATE_SKILL-resume}` (no colon) is load-bearing: *unset* means "check /resume", *explicitly empty* means "check none" and prints a line saying so — that is what lets the two hermetic sibling suites opt out without faking a `~/.claude`.
- **Non-blocking**: it can never fail a resume. `bounded_fetch` refuses an empty dir (an empty array subscript is a bash *error* that exited the whole digest 1), and a non-integer `RESUME_STATE_SKILL_SCAN_CAP` is validated once instead of printing `integer expected` to stderr per commit walked.
- `bounded_fetch` / `default_branch` are factored out of `handoff_freshness` so there is one fetch policy and one default-branch rule. The fetch is memoised **per directory, storing the RESULT** — a second caller reading a cached 0 over a fetch that FAILED would claim fresh refs it does not have.

Sample output on a genuinely stale deploy (the incident reproduced):

```
SKILL
  skill-read: 🔴 resume — deployed copy is 1 commit(s) BEHIND origin/main for
  claude/skills/resume/SKILL.md (newest it lacks: 200e6383 feat(claim-work): claim a
  ranked next-step by PUSHING an orphan ref …) [store copy at /nix/store/… — only a
  home-manager switch replaces it]
```

`claude/skills/resume/SKILL.md` documents how to read each variant. `description:` front matter untouched.

## 🔴 A finding must not suppress "nothing was reconciled" (audit round)

`main` printed `(no handoff loaded — nothing to reconcile; this is NOT a clean bill of health)` from an `elif` under `${#DRIFT[@]} -gt 0`, so **any** finding suppressed it. That was nearly unreachable before this PR — every finding came from reconciling a handoff, so a finding implied a handoff — and the SKILL block introduced the first finding that does **not**.

Measured side by side on one fixture (no handoff doc + a stale deployed skill): base prints the notice; this PR printed a bare `-` list under a header the skill defines as *"the lines where live state contradicts the handoff"*, with no handoff reconciled at all. The notice is now unconditional and **leads** the block, because it frames everything under it; the two facts are independent and get independent lines.

## 🔴 Residual scope — what this does NOT cover

This check runs at **resume START**. It catches *"you loaded a copy that was already behind."*

It **cannot** catch a skill change that lands **mid-session** — which is exactly what the measured incident was: #847 landed **2h20m after** that session began, when its `CURRENT` reading would have been true. A `CURRENT with origin/main` line is a claim about the instant the digest ran, never about the rest of the session. Nothing re-reads the digest.

That gap is pinned in the suite (`test_this_check_cannot_see_a_skill_that_changes_MID_session`) rather than papered over, and is labelled in-file as documentation of the limit, not as coverage of it. Closing it needs a different mechanism (a periodic re-check, or a hook on the tool call that acts on skill text) and is not attempted here.

A second limit, **recorded and pinned rather than fixed**: the history walk has no `--follow`, so a rename of the skill path degrades the report from "N behind" to "matches NO commit". Never to a false `CURRENT` — the tip-hash comparison runs before the walk — and `test_a_renamed_skill_path_degrades_to_not_current_never_to_CURRENT` holds that direction.

## Test matrix

`scripts/tests/test_resume_state_skill_freshness.py` — **38 cases**, in two rounds.

| round | baseline | red at baseline | green at HEAD |
|---|---|---|---|
| initial | `origin/main` `200e6383` | 21 of 22 | 22 |
| audit fixes | PR head `7822cbd5` | 6 behavioural + 1 doc-drift | all |

The audit-round reds: the no-handoff notice, the capped sentence, the matchless hedge, the UNMANAGED provenance, the non-integer cap (stderr), and the older matchless assertion. The 7th — the COULD-NOT-MEASURE scraper — is red against `7822cbd5`'s `SKILL.md` and names the real missing reason (`could not hash the deployed copy or the …`), which is the drift it exists to catch.

Everything else in the module is an **invariant guard on a property base already had** (the cap boundary, the fetch bound, the block order, the fetch-prefix lines, the rename direction, the store-path arm) and is labelled as such in the file rather than counted as regression coverage. `test_no_network_tool_is_ever_invoked` is green at both baselines by design.

One of these tests was itself caught being vacuous by the matrix: `assert "rc=1" in stdout` is **green at base**, because a missing bash function exits `127` and `"rc=127"` contains `"rc=1"`. The exact-match form is red there.

Sibling suites `test_resume_state_handoff_resolution.py` + `test_resume_state_clawgate.py`: **376 → 376 passed**, unchanged.

## Mutation results — 34/34 killed **for the right reason**

`scripts/tests/mutation_battery_resume_state_skill.py` (manual, not collected by the gate — same convention as its sibling), over a 38-test control. Each mutant names **the text its own guard produces**; a kill whose message comes from a *different* guard is reported `KILLED-WRONG-REASON` and counts as a battery failure. Eleven of the 34 (`R1`–`R11`) exist only because an audit fix **resets the verification gate**, and the whole battery was re-run after every fix — not just the row for what moved.

The battery found real gaps in both rounds:

- **round 1** — two *false* kills (`: ("…")` is a bash parse error, so the script died and the failure output quoted the source line containing the searched phrase); one expectation matching a neighbouring guard; three survivors, all real (a predicate spelled twice so the second copy was unobservable — fixed in the subject; a dirty checkout masking a stale store deploy; a skill repo different from the resumed repo). A mutant also exposed `FETCH_RC[""]` → `bad array subscript` exiting the whole digest 1.
- **round 2** — the "right reason" detector matched `expected in <whole pytest output>`, and `--tb=short` **also echoes the failing test's source**, which contains the assertion's own f-string literal: the same source-line-echo trap this file's docstring says it already hit once. It now reads only pytest's `E ` lines. It also counted only `N failed`, so the mutant that removes the fetch timeout — which makes a test **hang** until subprocess timeout, reported by pytest as `1 error` — scored SURVIVED. Errors now count as kills. Three assertion messages were widened to name the guard under test so a kill can be attributed at all, and `S15`'s pattern no longer existed after the capped/matchless split, which the battery reported as `NOT APPLIED` rather than passing over.

Python bytecode caching is not a hazard here — the mutated subject is bash.

**A fixture bug worth recording:** the hanging-fetch test first used `core.gitProxy` + a `git://` URL. That hangs perfectly by hand and fails **instantly** under this suite, because `testlib/nogit_plugin` exports `GIT_ALLOW_PROTOCOL=file` to refuse every transport that can reach another machine. The guard is right and was not widened; the fixture moved to the `file` transport (a local bare origin with a sleeping `remote.origin.uploadpack`). It presented as a 0.0 s "bounded" fetch — so the **timing** assertion is what caught it. A test asserting only `rc == 1` would have passed on a fetch that never hung.

Merged `origin/main` (`30acd174`) into the branch and gated **that** tree, not the PR branch alone.

| tier | command | result |
|---|---|---|
| dev host | `nix develop . --command bash scripts/gate.sh` | **`GATE: RESULT=PASS exit=0`** — pytest `collected=16871 passed=16869 skipped=2 failed=0`, node `1292/1292` |
| sandbox (what Tekton gates on) | `nix build .#checks.x86_64-linux.pytests` | **PASS** — `collected=16871 passed=16869 skipped=2 failed=0` |
| sandbox | `nix build .#checks.x86_64-linux.nodetests` | **PASS** |

The first merged-tree run was **red**, and correctly: `test_runtime_shebangs.py` caught the new fixture writing its own `#!/bin/sh` instead of going through `testlib.mockbin.write_exec` (the nix sandbox has no `/usr/bin/env`, and `patchShebangs` cannot reach a file a test writes at runtime). Fixed in `afa2a9ef`; the table above is the re-run.

⚠ **The earlier version of this section claimed `main` was red on both required checks** — a pre-existing xdist collection failure in `test_gh_issue_closing_condition_guard.py`. That is **obsolete**: #855 and #865 fixed it, this branch carries them, and both required checks now read `pass`. A comment on this PR records the correction rather than leaving it silently edited.

## Also in this diff

`scripts/tests/test_no_real_launchers.py` — `resume-state.sh` joins the `home-manager` acknowledgement. It is the **fifth** of the prose-mention shape and the **first that `scripts/tests` actually executes**, so the justification records the grep and the script's spawn set rather than rewording the messages to dodge the scanner. That count was itself corrected mid-round (`4519a276`): the provenance fork took the mentions from three to six, and a stale count in the one place whose job is to record what was grepped is the same class of defect as a stale comment anywhere else.

https://claude.ai/code/session_01YQwWKdSrcoeA8hbYaHTnVw


ZacxDev:
### Correction to this PR's description — it claimed `main` was red, and that is no longer true

The original body carried a section headed **"🔴 `main` is already red on both required checks — not from this PR"**, about the xdist collection failure in `test_gh_issue_closing_condition_guard.py` (a module-scope `tempfile.mkdtemp()` path embedded in `parametrize` values, so every worker generated different test IDs). That was measured and correct **when written**.

It is now **obsolete**: #855 and #865 fixed it, this branch carries both, and both required checks read `pass`. I have rewritten that section rather than deleting it silently — anyone who read the earlier version was told this PR could not merge, and that is no longer the case.

Flagging it here rather than only editing the body, because a description edit leaves no trace for a reader who already believed the old version.

### What else changed since the audit round

Two must-fixes, three should-fixes, four nits — every one re-verified live, with the **full** 34-mutant battery re-run after the fixes rather than just the row for what moved:

- the stale-skill finding was **suppressing** the `(no handoff loaded — nothing to reconcile…)` notice, because that notice lived in an `elif` under the findings branch. It is unconditional now, and leads the block.
- the capped walk asserted a cause it never measured ("built from a tree that was never pushed"), which is also false for a pushed-but-unmerged branch — the shape CLAUDE.md recommends for validating a nix edit. Split into two branches, both sentences pinned whole.
- the seven `COULD NOT MEASURE` reasons are now scraped from the script and required to appear in `SKILL.md` (it had already drifted: 7 in the code, 5 in the doc).
- `timeout 25`, the fetch memo's stored **result**, and the SKILL block leading the digest each had zero coverage; all three now have tests, the first behavioural (a hermetic hang via a sleeping `remote.origin.uploadpack`).
- the cap boundary (`-gt` vs `-ge`) was walkable because the fixture sat below it; there is now a fixture that lands exactly ON it.

Gate: both tiers on the **merged** tree (`origin/main` `30acd174` merged in) — dev host `GATE: RESULT=PASS exit=0`, and `nix build .#checks.x86_64-linux.{pytests,nodetests}` both PASS. The first merged-tree run was red for a real reason (`test_runtime_shebangs.py` caught the new fixture writing its own `#!/bin/sh`); fixed in `afa2a9ef`.

ZacxDev:
### Delta re-audit round — two must-fixes were real, and one of them was a false coverage claim I had written

Both were re-measured here before acting rather than taken on the review's word:

**The rename "pin" exercised no rename.** The fixture did `git mv REL REL.tmp` and back inside one commit. Measured: `--name-status` prints `M claude/skills/resume/SKILL.md` — git records no rename at all — `git log -- <rel>` still returns every commit, and the digest printed the ordinary `2 commit(s) BEHIND`. The one assertion (`"CURRENT" not in line`) passed for that reason. And `resume-state.sh` named that test as the pin for the `--follow` limitation, so the file **claimed coverage that did not exist**. Rebuilt against the real shape — a file that originates at `claude/commands/resume.md` and is renamed in, which is this repo's own history — with the fixture precondition asserted, not assumed.

**The hedge was still false one branch over.** `matches NO commit` said "built from a tree that is not on origin/main". The rename case reaches that same branch, and there the content **is** on `origin/main`, at the old path (verified: the pre-rename blob is reachable at `claude/commands/resume.md`). Every revision of this sentence has been false of the case found next, so it now states the **observation** — "this walk could not place the deployed copy in that path's history" — and lists three causes as possibilities, asserting none.

Also fixed: `00`/`007` were accepted where `0` was rejected (`[` reads decimal, `[ 1 -gt 00 ]` is true, so the walk capped on the first commit — the exact state the zero arm exists to prevent); the vocabulary scraper only matched a parenthetical ending a `printf`, so an `echo`-emitted reason was invisible — widening it immediately swept in a *comment* about the message shape and scored `…` as an eighth reason, so prose is now stripped and a control asserts it stays out; the spelled count in SKILL.md is derived from the scrape instead of trusted; the battery's docstring now states what its attribution actually proves (**the test** owning the guard went red, not "the guard emitted its text" — for several mutants the guard emits nothing, which is the defect).

Nits taken: the UNMANAGED provenance no longer mangles the sentences it splices into; the capped DRIFT line names `$dep` and reports the effective cap instead of `RESUME_STATE_SKILL_SCAN_CAP=<n>` (which read `=200` on an unset run *and* on an `abc` run, while the gap line beside it correctly quoted `=abc`); the fetch scanner strips quoted spans, so a future gap *message* mentioning `git fetch` is not scored as an unbounded invocation.

### ⚠ One thing deliberately not decided here

`test_a_hanging_fetch_…` costs **~25 s on every run of both tiers**. The sleeping upload-pack has to outlast the 25 s production timeout for the timeout to be observable, so the cost is structural. It is the only *behavioural* coverage of the headline non-blocking property ("a fetch that cannot complete must cost seconds, not a hung resume"), and the structural guard beside it is instant. Whether that belongs on every CI run, or behind a marker with the structural guard left inline, is a CI-budget call — flagging rather than quietly removing coverage.

### Verification

- **Battery re-run in full, twice.** The first re-run was **not** clean: three mutants reported `NOT APPLIED` (patterns my own rewording had invalidated — the machinery working as designed) and two were `KILLED-WRONG-REASON`. All five fixed; the second run is **38/38 killed for the right reason, no problems**, over a 41-test control.
- **Delta matrix**: 6 round-3 tests watched RED at the previous head `afa2a9ef`, green at HEAD.
- **Both tiers, on the merged tree against today's `origin/main` (`9b5dc48d`)** — not reasoned about, run:
  - dev host `GATE: RESULT=PASS exit=0` — pytest `collected=16963 passed=16961 skipped=2 failed=0`, node `1292/1292`
  - `nix build .#checks.x86_64-linux.pytests` → **PASS**, same counts
  - `nix build .#checks.x86_64-linux.nodetests` → **PASS**

ZacxDev:
### Round-4: both 🟡 closed, and a correction to the `--rebuild` advice

Both were re-measured here before acting. **Reproduction commands are inline** so each can be checked without a fresh audit.

**1. R1 described a mutation it did not make.** Verified by reading the tuple: `new` was `''`, a pure deletion — byte-identical to the revision its own comment criticises. The real pre-fix shape has **two** halves:

```bash
git show origin/main:scripts/resume-state.sh | sed -n '/^  echo "DRIFT"/,/^  else$/p'
# → no unconditional print, AND the echo inside `elif [ -z "$HANDOFF" ]`
```

Fixed in the **instrument**, not the prose: `old`/`new` may now be tuples applied in order, each still required to occur exactly once, because a faithful mutation is not always one edit. R1 is now both halves. The confirmation is in the sheet — its kill count moved **2 → 1**, which is exactly right: under the real pre-fix shape the no-finding path prints the notice again, so only the stale-skill-with-no-handoff test goes red.

**2. The capped branch promised a number it cannot produce.** Measured:

```bash
python3 - <<'PY'   # 5 commits touch the path; deployed content committed nowhere
# cap=2  → "(raise RESUME_STATE_SKILL_SCAN_CAP for the number)"
# cap=500 → "matches NO commit"  ← no number, ever
PY
```

Take the advice and you get no distance. The neighbouring `is older than the newest N commit(s)` was false in the same case for the same reason — the copy is not *older than* those commits, it is not in that path's history at all. Both now state only the measurement (`was NOT among the newest N commit(s)`), and the advice names its two possible outcomes.

🟢 also taken: the rename fixture's precondition was one-sided (a `git log` exiting 128 gives `stdout == ""`, so the `not in` passed having measured nothing — now checks the returncode and asserts a commit the walk MUST see); the remedy assertion was line-wide while the property is "not inside the parenthetical" (scoped, so it no longer blocks the better fix); the fetch scanner's quote-stripping is blind to `eval "git … fetch …"`, so a new test asserts no `eval`/`bash -c`/`sh -c` exists to hide one; R11's right-reason key was a literal appearing in half this module's messages — both arms of that guard now carry `SKILL.md VOCABULARY DRIFT`, which nothing else uses.

### 🔴 Correction: `--rebuild` is not a drop-in, and my earlier PASSes were real runs

Your cached-zero finding is right, and I verified both halves:

- **Every sandbox PASS reported on this PR printed its own `building '/nix/store/….drv'...` line**, so each was a real execution, not a cached zero. `grep -c "building '"` over the four saved build logs returns `1` each.
- A deliberate re-invocation immediately afterwards was **silent, exit 0** — the hazard exactly as you describe.

But **`--rebuild` errors on a tree that has never been built**:

```
error: some outputs of '/nix/store/…-devrc-pytests.drv' are not valid, so checking is not possible
```
exit 1 — because `--rebuild` means "build again and *compare*", which needs an existing output. On a freshly merged tree (the normal case) the derivation hash is new, so `--rebuild` refuses and a reader could mistake that for a failure.

The recipe that actually distinguishes a run from a cached zero, used here:

```bash
nix build .#checks.x86_64-linux.pytests --no-link 2>&1 | tee log
grep -c "building '" log        # 1 = it ran; 0 = cached, verdict is stale
nix build --rebuild .#checks.x86_64-linux.pytests --no-link   # now meaningful: determinism
```

### Verification

- **Battery: 38/38 killed for the RIGHT reason, no problems**, over a **42-test** control.
- **Delta matrix**: 1 test RED at the previous head `b559cfab` — the capped-sentence pin, the only behavioural change this round — green at HEAD. Everything else in round 4 pins properties the base already had and is labelled an invariant guard, not regression coverage.
- **Both tiers on the merged tree against `origin/main` `5baacb5a`, re-fetched at merge time**:
  - dev host `GATE: RESULT=PASS exit=0` — pytest `collected=16988 passed=16986 skipped=2 failed=0`, node `1292/1292`
  - sandbox pytests: **ran** (`building …` present) → `collected=16988 passed=16986 failed=0`
  - sandbox nodetests: **ran** → `1292/1292`
  - `--rebuild` on both afterwards: `checking outputs of …` → exit 0, so the two runs agreed

PR #862, opened 2026-08-26
feat(present): serve the explainer page on the workbench — WORKBENCH-LOCAL, regenerated daily

ZacxDev:
`scripts/present/generate.py` (merged in #829) measures **this host** and emits one
self-contained HTML file. Until now the only way to read it was to generate it by hand,
which in practice meant nobody read it. This hosts it — **on the workbench, for the
workbench.**

## 🔴 SCOPE, CORRECTED: this is WORKBENCH-LOCAL, not LAN

An earlier revision of this PR — title, body and three in-tree comment blocks — said the
page was served "on the workbench LAN". **That was false**, and the live `200` that
appeared to prove it was a same-host `curl` taking the `lo` path, which the firewall
accepts unconditionally. It says nothing about a second machine.

The socket *is* bound to a LAN address (`192.168.50.250`). But **8900 is not in
`/etc/nixos/configuration.nix`'s `networking.firewall.allowedTCPPorts`**
(`[ 7844 80 443 58012 8180 25565 8110 6443 ]`), so every off-host SYN is dropped.

**The measured control, from the laptop:**

| port | from the laptop | what it is |
|---|---|---|
| 22 | **OPEN** | ssh |
| 443 | **OPEN** | the gateway |
| **8899** | **CLOSED** | `initiatives-viewer` — **listening on the identical address right now** |
| **8900** | **CLOSED** | `present-serve` |

🔴 **8899 is the whole lesson.** `initiatives-viewer` has been listening on
`192.168.50.250:8899` for months with the identical firewall gap, and it reads as a
working LAN service to anyone who curls it from the workbench. Shaping this unit after it
inherited the flaw *and* the appearance of not having it. **The precedent is why nobody
was warned.**

### The ruling: ship it workbench-local

LAN and nebula reach are **deferred behind a named trigger — someone who is not on the
workbench actually needs to read this page.** Two reasons:

1. The **portable sanitized export** already serves the off-workbench reader, and the same
   regen run produces it. A file you hand over on purpose beats a port anyone on the LAN
   can hit.
2. Building hosted reach before a reader exists is the exact shape that left the
   subsystem-store-api dead on arrival.

Opening it later is a **system-level change** (`/etc/nixos`, `sudo nixos-rebuild`) **plus a
decision to publish client scope names to everything on the LAN**. Neither is done here:
**no firewall port is opened, `/etc/nixos` is not touched, and no sudo script is staged.**

⚠ **The deferral leans on the export, so check it per build rather than assume it.**
`generate.py --sanitize` prints one `SANITIZE DEGRADED` line per value it could not
substitute, and the page's mode chip carries the same counts. A 2026-08-26 build
substituted host×4, project×6, scope×12 — and left **two** client-ish names in the
sanitized copy (one hostname "indistinguishable from a word", one scope "matched in its
exact form only"). That is `scripts/present/sanitize.py`'s business, not this server's,
and it is out of scope here — but it means "the export serves the off-workbench reader" is
a claim to verify, not a standing fact. Closing condition: a `--sanitize` build that
reports zero `SANITIZE DEGRADED` lines.

## The units

| unit | what |
|---|---|
| `present-regen.service` | oneshot — builds **both** variants into `~/.local/share/present` |
| `present-regen.timer` | daily, `*-*-* 05:00:00`, `Persistent=true`, 600s jitter |
| `present-serve.service` | static file server on **`192.168.50.250:8900`** (workbench-local, see above) |

All three `serverMode`-gated, both services carry `OnFailure = notify-failure@%n.service`,
shaped after `initiatives-viewer` (`nix/home.nix` :2922).

**Bound to the workbench's own `eth0` address, never `192.168.50.94`** (a homelab node —
binding it crash-loops, as it already did to the viewer), and **deliberately not wired into
the public homelab gateway** either. `eth0`, not `eth1`: `ip -4 -o addr` confirms there is
no `eth1` on this host, and the wrong name was in `nix/home.nix` (twice — the
initiatives-viewer comment carried it too), `serve.py` and `test_present_units.py`.

The server is genuinely **static**: no refresh button, no subprocess, no credential, no
`sops`/`gh`/`kubectl` on its PATH (pinned by a test). It answers two artefact routes out of
an explicit table and 404s everything else, so there is no directory handler and therefore
no path-traversal surface.

## 🔴 The staleness choice: serve-last-good, with the age made obvious

The generator **exits 3 and writes no file** when every fact came back UNMEASURED. Correct —
it refuses to publish a build that would look careful and be broken. And it means a failed
regeneration leaves the *previous* page in place. Serving that unmarked is the same
silent-zero shape the page exists to teach against.

Three mechanisms, each for a different audience:

1. **writer** — `run-regen.sh` builds into a temp file and promotes with an atomic `mv`
   **only on exit 0**. A failed build never overwrites last-good and never half-writes.
2. **operator** — `OnFailure = notify-failure@%n.service` on both services. Verified firing
   live.
3. **reader** — `serve.py` computes the artefact's age from its mtime **at request time**
   and, past 30h, injects a fixed full-width high-contrast banner naming the age in words
   plus the unit to re-run. `X-Present-Stale` / `X-Present-Age-Seconds` / `X-Present-State`
   are the machine-readable twins.

**And the part that makes it a contract rather than an intention:** if the banner cannot be
injected, the bytes are **not served** — the response is a 503 interstitial. There is no
code path in `serve.py` that emits not-fresh artefact bytes unmarked.

**30h and not 48h**: cadence is 24h + 600s jitter, so 30h banners *before* a second
scheduled run is missed. The relationship (`threshold > cadence + jitter`, and
`< 2 × cadence`) is asserted, not just commented.

### 🔴 THREE states, not two — the backwards clock jump

`age = max(0.0, now - mtime)` clamped a future mtime to `0` and reported **`fresh`**. The
clamp was justified from a `touch -d '3 days 4 hours ago'` typo made during live
verification (GNU `date` binds `ago` to the last term only), and as a *header-format*
decision it is right — an int consumer should not have to parse a minus sign.

**The case it did not consider is a backwards clock jump.** A host with a bad RTC boots,
NTP steps the clock backwards, and every already-written file is now dated in the future. A
genuinely 8-day-old page then reads `fresh`, `X-Present-Age-Seconds: 0`, **no banner** —
the single reader-facing staleness signal disarming itself, silently, in the one situation
where the machine's own sense of time is what is broken. In a file whose stated purpose is
never to do that.

So `clock-suspect` is now its own state:

* its **own reader-facing banner** — same fixed full-bleed chrome, headline **AGE UNKNOWN
  — this page's timestamp is in the FUTURE**, and it deliberately does not name an age,
  because the age is precisely what is unknown. Visible the way `stale` is, not merely a
  header.
* `X-Present-State: clock-suspect-bannered` / `clock-suspect-withheld`, plus
  `X-Present-Clock-Skew-Seconds`.
* `X-Present-Stale: 1` — a clock-suspect page has *no measurable age*, which is strictly
  worse than a measured stale one. Reporting `0` there is the same disarm for a machine
  consumer.
* **the guard is asked BEFORE the staleness comparison.** `age` is clamped, so a future
  mtime reaches `age > stale_after` as `0` and lands in the fresh branch every time — any
  other order makes this guard unreachable for every value it exists to catch.
* both non-fresh states now exit through **one** `_serve_warned` helper, so a fourth state
  cannot re-open the no-unbannered-bytes hole by forgetting the refusal.
* **120s tolerance**, so a sub-second NTP slew or a filesystem timestamp rounding does not
  raise a clock alarm. A banner that fires on noise is a banner nobody reads.

**`test_an_mtime_in_the_future_clamps_to_zero_rather_than_going_negative` is REPLACED, not
kept — deliberately, and the file says which half survived.** Its *header-format* half
lives on as `test_the_age_header_never_reports_a_negative_number` (now measured on **both**
sides of the tolerance, since the clamp and the classification are two different decisions
and only one moved). Its *classification* half asserted `X-Present-State == "fresh"` for a
mtime 500,000s in the future — it pinned the defect — and is gone.

### 🔴 The port ledger claimed more than it checked

The header said the scan covered "any OTHER port this host **BINDS**"; the implementation
was `re.findall(r'"([A-Z0-9_]*PORT)=(\d+)"')`, which sees only `"NAME_PORT=NNNN"` string
literals. **Proven, not argued:** changing `nix/observability.nix:138` from
`--server.http.listen-addr=127.0.0.1:12349` to `…:8900` — a real local-bind collision with
`present-serve` — left the suite **fully green**.

**Widened** (the prose was not narrowed). Three declaration shapes, comments stripped:

1. `"NAME_PORT=NNNN"` — the systemd `Environment` literal
2. `HOST:NNNN` — but only for a host **this machine can bind** (loopback, the workbench's
   own LAN and nebula addresses). `192.168.50.94:6443` is a homelab node and is excluded
   *structurally*, not by a ledger entry.
3. `someNamePort = NNNN;` — a nix binding interpolated into a `--listen-addr`-style flag

Labels are classified by **OWNER**, so an env var and the flag that consumes it are one
claim while two genuinely different programs on one port are two. The scan goes from
7 labels to **11**, and it is not noisy: the full suite is green on the unmutated tree.

🔴 **What it still cannot see is enumerated in its own header** rather than implied away:
a port bound by a flag that lives in `scripts/` not `nix/`; a port named only in a comment
(deliberate — prose is not a bind, and including it made the viewer's own comment alias its
unit); two different units in ONE file binding the same loopback port (labels are
`<file>::<host>`, so those collapse — the `*_PORT=` shape, which is what the units in
`home.nix` actually use, does not have this blind spot); and anything the firewall does.

## Mutation evidence

`PYTHONDONTWRITEBYTECODE=1` throughout, each mutant applied by a harness that **fails loudly
if its pattern does not occur exactly once** — a mutant that did not apply reports SURVIVED
for the wrong reason. Each is shown killed by a **named** guard with **its own** message.

**The port ledger** — the exact mutation that used to survive green:

```
T3-alloy-port-collides   nix/observability.nix
  --server.http.listen-addr=127.0.0.1:12349  ->  …:8900

FAILED test_no_two_locally_bound_ports_collide
E  AssertionError: two different things bind the same port:
   {8900: ['obs-ship / alloy', 'present-serve']}. On this host that is not a
   warning — the second one to start fails to bind and the unit crash-loops,
   typically first noticed after a reboot.
```

Killed in isolation (`-k` down to that one test), so it is not riding on a neighbour.

**The clock-suspect state** — six mutants, six named kills:

| mutant | killed by | its own message |
|---|---|---|
| `clock_suspect = False` (the pre-change behaviour) | `test_a_future_mtime_is_clock_suspect_and_never_fresh` | *an artefact dated 8 days in the FUTURE was classified `'fresh'`. The clamp makes its age 0, so the staleness comparison cannot see it…* |
| guard fed the **clamped** `age` instead of signed `delta` — reads correctly, can never fire | `test_the_clock_guard_is_asked_before_the_staleness_comparison` | *got `'fresh'` — the clock guard did not run before the staleness comparison* |
| **freshness checked first** (the ordering mutant) | same | same |
| `X-Present-Stale` computed from `stale` alone | `test_a_future_mtime_is_clock_suspect_and_never_fresh` | *the machine-readable twin says this page is current. A consumer reading only the header gets the disarmed answer.* |
| tolerance sign flipped (`delta < CLOCK_SUSPECT_AFTER`) | `test_a_sub_tolerance_future_mtime_is_still_fresh` | *a 1.0s future skew tripped the clock guard; the tolerance is 120.0s and the boundary itself is fresh* |
| clock-suspect state returns 200 directly, bypassing `_serve_warned` | `test_a_clock_suspect_page_that_cannot_be_bannered_is_withheld` | *got 200: a clock-suspect page the injector cannot banner was served anyway… the no-unbannered-bytes contract is re-opened.* |

🔴 **Reachability, separately from correctness.** `test_the_clock_guard_is_asked_before_the_staleness_comparison`
drives `stale_after=1.0` and asserts a **control first** (an hour-old page must come back
`stale-bannered`, or the test is not exercising the staleness branch at all), then that the
future-dated page still comes back `clock-suspect-bannered`. Two of the six mutants are
unreachable-guard shapes and both die on that test.

## Coverage

None of this imports at the base ref, so **no test here can be shown red on pre-change
code**; both modules say so in their docstrings. **84 tests** across the two files (57
`test_present_serve.py`, 27 `test_present_units.py`).

New/strengthened this round:

* the central sweep now runs **across zero**, not just across the threshold — 13 offsets,
  and it asserts it reached **all three** states (`fresh` / `stale` / `clock-suspect`) at
  least 3× each, so it cannot pass by never leaving one branch.
* its predicate is `_warned`, not `_has_banner`. The first cut asked only about the
  staleness banner, so a clock-suspect response carrying nothing would have counted as
  fresh — `_has_banner` and `X-Present-Stale` agreeing with each other about a page neither
  had looked at properly.
* **two banner detectors with a cross-control** — `test_positive_control_the_two_banner_detectors_do_not_see_each_other`
  fails if either fires on the other's banner, i.e. if `_warned` is one predicate wearing
  two names.
* both detectors are **structural, not spelled** (headline words **and** the z-index), each
  with a not-fooled-by-the-word control.
* the port scan carries **one positive control per declaration shape**, so a regex that
  silently stopped matching one of the three cannot hide behind the other two — plus a
  **negative** control (`192.168.50.94:6443` must not be collected) and a **both-directions**
  control on the comment stripper (it removes the viewer's commented `192.168.50.250:8899`,
  and does **not** remove the SOCKS proxy's `127.0.0.1:1080` ExecStart).
* both banners are checked for self-containment through the generator's **own**
  `generate.self_contained` scanner, parametrised over `stale` and `clock-suspect`.

## Verification

**The page was generated and read**, both variants, from this tree:

```
present: 21 measured, 1 unmeasured, 22 registered
  UNMEASURED  seam.store_traffic: … an empty grep is not a zero
present: wrote present.html (114,593 B)
present: 🔴 SANITIZE DEGRADED — hostname indistinguishable from a word
present: 🔴 SANITIZE DEGRADED — scope matched in its exact form only
present: wrote present-sanitized.html (114,523 B) [SANITIZED]
```

Rendered to text and read end-to-end. The page itself makes **no** reachability claim about
`present-serve` — it does not mention the unit or 8900 at all — so there was no LAN claim to
correct there. `present-regen.timer` is correctly absent from its own timers roster, since
the units are left stopped until this merges and `ship.sh` runs. The two `SANITIZE DEGRADED`
lines are what surfaced the export caveat recorded above; that finding came only from
reading the output.

## Known and deliberately NOT fixed here

* **non-UTF-8 artefact** (`serve.py`): `UnicodeDecodeError` is a `ValueError` and escapes
  `except OSError`, so the connection closes with no status line. It needs external
  corruption of a file the generator writes in a single `write_text`, and it **cannot leak
  stale bytes**. Bounded; left.
* symlink-follow in the artefact dir; the 404 message naming 3 of 4 routes;
  `network-online.target` being inert (pre-existing repo-wide); boot latency behind the
  oneshot.

## Gates — two tiers, named, on the MERGED tree

Merged `origin/main` (`1a021a92`) in first, so these describe a tree close to what will
merge rather than a stale branch. `main` did **not** touch any file this PR touches
(checked against the merge-base), so there is no semantic-conflict surface.

**Gated base sha: `11e98704`** (this branch + `origin/main` @ `1a021a92`).

| tier | command | verdict |
|---|---|---|
| dev host | `nix develop --impure --command bash scripts/gate.sh --tier both` | **`GATE: RESULT=PASS exit=0`** — pytest `RESULT: PASS (exit=0)`, `TOTAL collected=16917 passed=16915 skipped=2 failed=0 (floor: 15789 = sum of 28 per-target floors)`; node `RESULT: PASS (exit=0)`, `TOTAL suites=5 files=39 tests=1292 pass=1292 fail=0 skipped=0 (floor: 1239)` |
| nix sandbox **(the gating tier)** | `nix build .#checks.x86_64-linux.pytests` | **`RESULT: PASS (exit=0)`** — `TOTAL collected=16917 passed=16915 skipped=2 failed=0` |
| nix sandbox **(the gating tier)** | `nix build .#checks.x86_64-linux.nodetests` | **`RESULT: PASS (exit=0)`** — `TOTAL suites=5 files=39 tests=1292 pass=1292 fail=0 skipped=0` |

One dev-host run of four went red on `test_subsystem_store_api.py`, a file this PR does not
touch. It is a **real race in that test** (a positional slice of a concurrently-appended audit
list), not a load flake and not this change — diagnosed, ruled out four ways, and filed with a
mechanical closing condition in the gate comment below.

These are two different tiers, not two spellings of one: `gate.sh` runs on the dev host and
never invokes `nix build`; the sandbox tier builds from a `cp -r ${./.}` store copy with no
`.git`, and that is the tier the merge is gated on.


ZacxDev:
## Gates — both tiers, on `a9906dbf`

Named by tier and base, because "the gate passed" is true of one run, one tier, one base.

**dev-host tier** — `nix develop --impure --command bash scripts/gate.sh --tier both`

```
  TOTAL collected=16717  passed=16715  skipped=2  failed=0  (floor: 15789 = sum of 28 per-target floors)
  TOTAL suites=5 files=39 tests=1292 pass=1292 fail=0 skipped=0  (floor: 1239)
GATE: RESULT=PASS exit=0
```

**sandbox tier** — the one Tekton runs and the merge is gated on

```
nix build --impure -L .#checks.x86_64-linux.pytests
  devrc-pytests>   TOTAL collected=16717  passed=16715  skipped=2  failed=0  (floor: 15789)

nix build --impure -L .#checks.x86_64-linux.nodetests
  devrc-nodetests> TOTAL suites=5 files=39 tests=1292 pass=1292 fail=0 skipped=0  (floor: 1239)
```

**No floor moved.** The added tests only raise counts; the gate printed no drift
suggestion, so `run-tests.sh` is untouched — one fewer conflict on the repo's most
conflict-prone line.

### The two tiers disagreed twice, and each disagreement was a real defect

Both were in the *tests*, both were invisible to the tier you are most likely to run,
and each is now its own commit.

1. **`b55ef626`** — the stub generator wrote `#!/usr/bin/env bash` at runtime.
   Execs fine on this NixOS dev host (`/usr/bin/env` is a coreutils symlink); ENOENTs
   in the build sandbox. Caught by `test_runtime_shebangs.py` — which lives in a
   *different file*, so `pytest scripts/tests/test_present_serve.py` could never see
   it. Routed through `testlib.mockbin.write_exec`, which owns the shebang.
2. **`a9906dbf`** — the git-tracked check called `git ls-files` with `check=True`.
   The sandbox builds from a `cp -r ${./.}` store copy with **no `.git`**, so it exits
   128 there: the sandbox tier went red with a `CalledProcessError` while the dev-host
   gate was fully green. A check written to catch a deploy hazard, taken out by the
   two-tier hazard. No skip added (a skip that always fires in the gating tier is
   blind there, and it would move the pinned skip accounting) — instead each tier
   proves the same claim by its own route, and the degraded arm asserts *why* it
   degraded rather than passing quietly.

### The live verification stands on this commit

`git diff 969c0b8d..a9906dbf -- nix/ scripts/present/` is **empty** — the two follow-up
commits touched only `scripts/tests/`. The `home-manager switch` in the PR body deployed
byte-identical units and scripts.

ZacxDev:
## Gates — both tiers, on the MERGED tree

`origin/main` (`1a021a92`) merged in first, so these describe a tree close to what will merge
rather than a stale branch. `main` did **not** touch any file this PR touches (checked against
the merge-base `1d67f5e8`), so there is no semantic-conflict surface to read.

**Gated base sha: `11e98704`.**

### 🔴 The tier the merge is gated on — `nix build` (store copy, no `.git`)

| check | verdict (read from the log, not an exit code) |
|---|---|
| `nix build .#checks.x86_64-linux.pytests` | `RESULT: PASS (exit=0)` — `TOTAL collected=16917 passed=16915 skipped=2 failed=0 (floor: 15789 = sum of 28 per-target floors)` |
| `nix build .#checks.x86_64-linux.nodetests` | `RESULT: PASS (exit=0)` — `TOTAL suites=5 files=39 tests=1292 pass=1292 fail=0 skipped=0 (floor: 1239)` |

Per-target node counts: browser-bridge 507/floor 490, dl-router 508/500, collector browser-ext
21/20, clickup 188/179, discord-embed-ext 68/50.

### The dev-host tier — `scripts/gate.sh --tier both`

| run | base | box | verdict |
|---|---|---|---|
| 1 | `b1d849d7` (same merge, differs from head only by two doc strings) | busy | **`GATE: RESULT=PASS exit=0`** — pytest `collected=16917 passed=16915 skipped=2 failed=0`; node `tests=1292 pass=1292 fail=0` |
| 2 | `11e98704` | busy | `GATE: RESULT=FAIL exit=1` — node `PASS`; pytest **`1 failed, 8464 passed`**, all 27 other targets green |
| 3 | `11e98704` | **clear** | **`GATE: RESULT=PASS exit=0`** — pytest `collected=16917 passed=16915 skipped=2 failed=0` (`8465 passed in 169.69s`); node `tests=1292 pass=1292 fail=0` |

**Dev-host verdict on the final head: PASS.** Three of four runs green; the single red is
below, and run 3 re-ran the identical head with no other gate competing for the box.

⚠ **The one failure is not this change, and I am not fixing it here.**

```
FAILED scripts/tests/test_subsystem_store_api.py::
       TestClientIpIsCloudflareOnly::test_unidentified_requests_are_NOT_bucketed_together
E  AssertionError: {'no-client-ip', 'recalled'} == {'no-client-ip'}
```

`test_subsystem_store_api.py` is untouched by this PR and by the eleven `main` commits merged in.

**It is a real race in that test, not a load flake, and here is how each was ruled out:**

* **Not load.** RULES' discriminator is wall time (~15×) and *whose* time moved. Run 1 total
  `188.34s`, run 2 `192.66s` — 2% apart, and exactly **one** test moved. That is the signature
  of a failed assertion, not a loaded box.
* **Not my change.** The diff touches `nix/home.nix` and `scripts/present/` only.
* **Not reproducible in isolation.** The class alone: `10 passed`. That single test repeated
  **10×**: `1 passed` every time, `0.58s` each.
* **Green in the gating tier** on the identical head.
* **Green on re-run of the identical head** (run 3), with the box clear — 1 red in 4 full
  dev-host runs of the same tree.

**The mechanism**, from the file's own docstring: `running()` yields a live `audit` list appended
to by handler threads on a `ThreadingHTTPServer`, and `drain_output` already warns —

> A handler writes its response and only THEN calls `_audit()` … `fetch` returning means the
> response was written, not that the handler thread has reached its `print`.

The failing test fires 20 unidentified requests, then a 21st identified one, then asserts on
`audit[:20]` — a **positional** slice of a list whose append order is not the request order. The
21st handler thread can append `recalled` before one of the first 20 appends `no-client-ip`,
which is precisely the observed `{'no-client-ip', 'recalled'}`. `await_audit()` exists in that
same file to close this class and this site does not use it — the open-coded predicate the file's
own history section describes as "wrong at N−1 sites, re-fixed one site at a time".

**Closing condition** (mechanical): that test selects its 20 lines **by content rather than by
list position** (or waits for 21 audit lines before slicing), and the class repeated 20× under
`-n auto` is green. Owner: whoever holds `test_subsystem_store_api.py` — note there is an active
`retire-subsystem-store-api` branch, so it may be deleted rather than fixed.

### Mutation evidence, re-run on this exact head

The port-ledger mutation — the one that used to survive **fully green**:

```
nix/observability.nix:138  --server.http.listen-addr=127.0.0.1:12349  ->  …:8900

FAILED test_no_two_locally_bound_ports_collide
E  AssertionError: two different things bind the same port:
   {8900: ['obs-ship / alloy', 'present-serve']}. On this host that is not a
   warning — the second one to start fails to bind and the unit crash-loops,
   typically first noticed after a reboot.
```

The clock-suspect mutation:

```
scripts/present/serve.py   clock_suspect = delta < -CLOCK_SUSPECT_AFTER  ->  False

FAILED test_a_future_mtime_is_clock_suspect_and_never_fresh
E  AssertionError: an artefact dated 8 days in the FUTURE was classified 'fresh'.
   The clamp makes its age 0, so the staleness comparison cannot see it: without
   its own state this page is served as current however old it really is.
```

Both applied by a harness that aborts if its pattern does not occur exactly once, under
`PYTHONDONTWRITEBYTECODE=1`, and the tree was verified byte-clean after each restore. Full
six-mutant table in the PR description.

PR #863, opened 2026-08-26
feat(subsystem-store): ship the CLIENT the store never had — plus a route guard that read as coverage and provided none

ZacxDev:
Clawgate task **360**, criteria **2, 3** and the token half of **6**. Follows #856 (the rule reword) and the two retirement PRs closed unmerged (devrc#849, ZacxDev/homelab-infra#404).

## Why this exists

The store API has served reads since 2026-08-16 and logged **309 requests, all from the session that built it**, because the **client was never written**. `_reject_write()` refuses every write verb, the route table was `/healthz` + `recall` + `search`, and no client-side token existed on either host. Retirement was proposed on that evidence. This is the missing half.

## What landed

**`scripts/store`** — read-through client. Syncs a cache and runs the **unmodified** local reader against it, so `recall` keeps its no-network property on the `/resume` hot path (§2d). It prints its own state line for the same measured reason the server prints a provenance banner: the reader's *"none omitted"* is truthful about whatever bytes it was pointed at.

| state | exit |
|---|---|
| `live` | 0 |
| `cached (stale <age>)` | 0 |
| `scope-empty` | 0 |
| `store-unreachable, no cache` | **3**, naming the host |

`sync` differs on purpose: **exit 4** when it could not refresh even though a cache survives. A `sync` that exited 0 on an outage is how a timer reports success forever while the cache silently ages out.

**`GET /api/v1/snapshot`** — a **read** route. `do_POST = do_PUT = do_PATCH = do_DELETE = _reject_write` is untouched, with a new test asserting the snapshot route added no write verb. Phase 2 stays read-only; that guard is phase 3's to break deliberately.

## 🔴 Two defects found, one of them mine

**1. The route guard was spelled, not structural.** `test_the_only_routes_are_recall_and_search` claimed in its docstring to *"walk the endpoint list"*. It probed four hardcoded non-routes. **Measured: I added `/api/v1/snapshot` and all three tests in that class stayed green.** A guard that reads as "the only routes are X and Y" while permitting Z is worse than none — it stops anyone looking.

Rewritten to derive the accepted set from the router and assert **equality**, mutation-tested both directions:

```
grow:   accepts ['exfiltrate','recall','search','snapshot'] vs ledger ['recall','search','snapshot']
shrink: accepts ['recall','snapshot']                       vs ledger ['recall','search','snapshot']
```

**2. `int(st.st_mtime)` silently reordered the cached index — and my own test could not see it.** The reader orders newest-first by mtime. Entries written in the *same second* (normal: one `/handoff` writes several) all tie on a truncated mtime, so the ref tie-break reorders the copy. Same bytes, same count, different order, **no error**.

The mtime test **passed against the broken server** because its fixture used whole-second mtimes a day apart — no fraction to lose. Fixed with `PAX_FORMAT` + float mtime; fixture now uses fractions inside one second. Reverting the fix takes **three** tests red, including the criterion-2 byte-identity one.

## ⚠ Cloudflare 403s the default Python User-Agent

Same token, same path, three requests differing only in that header:

```
curl default UA    -> 200
Python-urllib/3.12 -> 403     <-- urllib's default
empty UA           -> 200
```

The 403 is the **edge, not the app** — neither a bad token nor an outage, while looking like both. Every Python client of `store.zacx.dev` is blocked out of the box. The client sets its own UA; `SECRETS.md` records it so the next one does not rediscover it.

## Verification

- **295 passed** (`test_store_cli.py` 10 new + `test_subsystem_store_api.py` 285).
- The CLI fixture puts a shim in front of the server adding `CF-Connecting-IP`, standing in for the hop that adds it in production. Teaching the client to send it would test a client we must not ship (the server refuses a duplicated one).
- Harness controls both ways: a wrong token through the shim is refused (401), and a correct one succeeds — so a 200 means the token was checked, not that auth was absent.
- Tests point config at a non-existent path so a real `~/.config/subsystem-store/env` can never make one pass.

## 🔴 NOT verified

- **The `live` state against PRODUCTION.** The pod runs image `0.3.0`, which predates this route — `/api/v1/snapshot` **404s** there. All four states are verified against a local server running this code. Deploying needs an image build + Harbor push + manifest bump; deliberately not in this PR.
- **§5's off-mesh control** still has never run. Auth controls were exercised from the workbench, which is on the mesh.
- **Token rotation** never exercised.
- `TARGET_FLOORS` left unchanged: the only count measurable outside the gate is a **partial** collection (3 unrelated modules `ImportError` without the gate's deps, 7547 < the 8036 floor), so setting a floor from it would *lower* the guard. Raise it from the gate's reported number.

## Blast radius

**Reversible.** Adds one read route and one new file; changes no existing behaviour and no writer. The local store remains authoritative and untouched — nothing yet reads from the cache in anger. Rollback is reverting the commit; nothing is deployed by merging.

https://claude.ai/code/session_01A7CtMW371evXEXDkjgTNbz


ZacxDev:
🔴 **Correcting this PR's own description before anyone reviews on it.** An adversarial audit measured that two headline claims in the body above are overstated. Recording it here rather than editing the body silently, since a reviewer may already have read the wrong version.

**1. "Rewritten to derive the accepted set from the router and assert equality, mutation-tested both directions" — overstated.** The ledger guard is still a SPELLED guard, which is the exact defect this PR's title claims to close.

The regex is `parts\[0\]\s*==\s*"([a-z0-9-]+)"`. Measured against the real server:

- add `parts[0] == "raw_dump"` (underscore) to the router → **all 6 tests in `TestPhaseOneScope` PASS**
- re-spell it `"rawdump"` → the ledger test fails

So it can go red, but 8 of 9 plausible spellings walk past it: single quotes, underscore, `parts[0] in (...)`, reversed operands, uppercase, `startswith`, dict dispatch, aliased variable. **The positive control I shipped (`test_the_ledger_guard_can_SEE_an_unledgered_route`) feeds it the one spelling it catches**, so it was structurally incapable of revealing this. The behavioural companion still probes 4 hardcoded non-routes, so `raw_dump` is invisible to *both* guards.

**2. "Phase 2 stays read-only" is accurate, but the snapshot route has three unreported defects of the class this PR exists to eliminate**, each reproduced:

- an **unreadable** scope (`chmod 000`) renders as `scope-empty` at exit 0 — `Path.glob` swallows `PermissionError`, and only `root.iterdir()` is wrapped
- two **concurrent** syncs truncate the shared cache (3/10 trials: 183/305, 256/305, 292/305 entries, no error)
- `sync --scope X` / `validate --scope X` narrow the SHARED cache (305 → 2 entries measured), after which every other scope reports "nothing recorded yet"

**What the audit independently CONFIRMED**, so this comment is not a blanket retraction: the read-only property is genuinely intact (both write guards unmodified, 405 + `Allow: GET, HEAD` on the new route), the auth path is shared and 401s without a token, the `?scope=` traversal guards fire, and the mtime work is real — reverting `st.st_mtime` → `int(st.st_mtime)` was reproduced as exactly 3 tests red including the byte-identity one, under Python 3.12 **and** 3.14.

Fixes for the 🔴s and the cheap 🟡s are in progress on this branch; a delta re-audit will follow, since a fix round is where the next defect usually arrives.

🤖 via [Claude Code](https://claude.com/claude-code)

ZacxDev:
Measured evidence you may want before this merges — `test_subsystem_store_api.py` is flaky **on pristine `origin/main`**, independent of this PR.

A separate session measured it with a both-sides control under the 4-way xdist the gate now runs (`#841`):

| tree | runs under `-n 4` | red |
|---|---|---|
| pristine `origin/main`, unrelated change absent | 10 | **2** |
| that session's branch | 10 | **1** |

Same failure family both sides:
- `IndexError` on `audit[N]`
- `assert 'status=lockout-triggered' in …`
- `expected at least 3 audit line(s) within 15s, got 2`

Mechanism: the assertions index an audit log written **asynchronously by the server subprocess** and assume both arrival and ordering within a fixed wait. Under xdist the file's tests share a worker with everything else on it, so the wait is no longer generous. Roughly 10–20% per run of that file, pre-existing, not caused by either branch.

This is not hypothetical any more: it took **`#857`** red minutes ago on `TestAuditLogCannotBeForged::test_a_NEWLINE_in_the_path_cannot_open_a_second_record`, which is why I went looking.

Why it matters here: `tekton/devrc-pytests` is a **required** check with `enforce_admins: true`, so a ~10–20% flake can red any PR at random with no admin override — and per the repo's own rules a gate that goes red for reasons unrelated to the change trains everyone to click through it.

Flagging rather than fixing: this PR already owns `test_subsystem_store_api.py` and `server.py` (the writer), so a competing PR against the same file would just be a collision. Whether it belongs in this PR or a follow-on is your call — the useful part is that the flake is **measured with a control**, not inferred, so it does not need re-deriving.

Suggested closing condition if it becomes a follow-on: the assertions wait on a **condition** (the expected line appearing) rather than a fixed timeout, and 20 consecutive `-n 4` runs of that file come back green on `origin/main`.

ZacxDev:
Follow-up to my earlier comment: the flake now **reproduces on demand**, and the family is **wider than the one test I named**.

A load harness (PR #899, `scripts/dl-router/tests/load_test_store.sh`) was fixed and pointed at this file. Four conditions, measured:

| | target | burners | xdist | runs | red |
|---|---|---|---|---|---|
| A | `TestAuditLogCannotBeForged::test_a_NEWLINE_in_the_path_cannot_open_a_second_record` alone | 12 (load →28) | none | 20 | **0/20** |
| B | whole `test_subsystem_store_api.py` | 0 | `-n 4 --dist load` | 10 | **0/10** |
| D | whole `scripts/tests` (the gate's own shape) | 6 | `-n 4 --dist loadfile` | 10 | **3/10** |

**CPU load alone does not reproduce it. Single-file xdist does not either.** Only condition D — the shape the gate actually runs — does.

🔴 **And the three red runs were not the test I originally named.** They were:
- `TestLockoutOverHTTP::test_five_failures_lock_out_a_VALID_token_from_the_same_client` (runs 3 and 7)
- `TestLockoutOverHTTP::test_a_SUCCESS_does_NOT_buy_more_GUESSES` (run 5)

Same mechanism in every case — `assert "status=lockout-triggered" in audit[N]` where the record at that index is an earlier one. So the defect is the **shared assumption**, not any one test: assertions index an audit log written asynchronously by a server subprocess and assume both arrival *and* ordering within a fixed wait. Under the gate's real concurrency the log is simply behind.

Practical consequences for a fix here:

1. **A fix verified against only the `TestAuditLogCannotBeForged` case is not verified.** At least three tests across two classes share the mechanism, and it is likely every `audit[N]` index in the file.
2. **Reproduce under condition D, not under load alone** — A and B both come back clean and would read as "fixed" against an unfixed tree. That is the whole reason the conditions are tabulated here.
3. Honest limit on the above: D applied load *and* gate-shape xdist together, and no `BURNERS=0` run of D was made, so the two are not separated. A and B rule out each **alone**; D does not tell you which is doing the work.
4. Condition A's 0/20 is a negative result at that run count, not proof that nodeid cannot reproduce.

Suggested closing condition, unchanged in spirit but now checkable: the assertions wait on the **condition** (the expected record appearing) rather than a fixed timeout, and **20 consecutive condition-D runs** come back green on `origin/main`.

Not fixing any of this here — this PR owns the file. Flagging because the reproduction and the wider blast radius are measured, not inferred, and would otherwise have to be re-derived.

PR #864, opened 2026-08-26
feat(opencode): opencode briefs must carry their sources (rc 6) — ~0.9% of the brief surface, and NOT the part the evidence came from

ZacxDev:
## 🔴 Read this first: it covers ~0.9% of the surface, and not the part the evidence came from

| brief surface | rate | covered here |
|---|---|---|
| **opencode dispatch** (`opencode-dispatch preflight()`) | **~1.2/day** | ✅ **yes** |
| Claude `Agent`-tool subagent brief | ~137/day | ❌ **no chokepoint exists** |
| clawgate task body (`build_task_body()` ×2) | uncounted | ❌ no |

**Basis, so it can be argued with:** 6 briefs surviving on disk across 3 project directories over **Aug 21–25** (5 days) — explicitly a **FLOOR**, not a total, since `.opencode-dispatch/` is deleted with its project and the local telemetry spool had already shipped — against **1,921 `Agent` calls in the audit's 14-day window**. The two windows differ, so treat **0.9%** as an order of magnitude. If anything it **over**-states coverage: the clawgate producers are a third surface counted in neither number.

🔴 **The asymmetry that matters most: all four measured instances came from the UNCOVERED surface.** The wrong root cause reached three *subagent* briefs. The three subagents who opened their reports correcting a stale brief were Agent-tool subagents. **Not one was an opencode dispatch.**

So this is a **bridgehead, not a fix**, and it is named, documented and merged as one. Please do not cite it as evidence that premise-propagation is handled — a guard reading wider than it implements is worse than none, because it stops anyone looking. What it does provide: the schema, vocabulary and refusal semantics a wider guard would reuse, proven against a real dispatch path.

## What the class is

A 14-day audit of 443 sessions found **wrong premises propagating into subagent briefs** independently in **four of six** audit slices — the highest-cost error class found, and one the adversarial audit ladder is structurally blind to.

| measured instance | what happened |
|---|---|
| stale comment + stale README | a whole storage layer built on it; **four adversarial audit rounds read past it** |
| wrong root-cause diagnosis | filed as a GitHub issue, endorsed as the session's best finding, **pushed into three subagent briefs**, retracted at session end. The recommended fix would have preserved every false pass across a 69-site sweep |
| auth constraint inferred from a token prefix | reported as established fact several times, written into a handoff as the ranked next step, **propagated into a downstream session's opening brief**. Quoted: *"my inference presented as fact."* |
| three `cli` subagents | opened their reports correcting the brief they were handed: *"The brief was stale — the code won."* |

🔴 **This only works structurally.** As a prose instruction it is the same mechanism that already failed — the audit ladder is itself instruction-based and missed this four times over. So it is a schema, in code, at a chokepoint, and it **refuses**.

## The chokepoint I found (and the one I did not)

I swept every committed producer of an agent prompt in this repo. Three surfaces exist; only one is covered here.

| surface | committed chokepoint? | covered by this PR |
|---|---|---|
| **opencode dispatch brief** | ✅ `opencode-dispatch` `preflight()` — both `preflight` and `run` route through it | **yes** |
| clawgate task body (what a devpod agent literally executes) | ✅ two INDEPENDENT producers — `repo-cos/clawgate.py:build_task_body(proposal)` and `initiatives/dispatch.py:build_task_body(view, recommendation)`. Different signatures, so not literal copies; neither validates its body, and both sit outside `clawgate-task-interview-guard.py`, which is interactive-sessions-only by its own docstring | no — see follow-up |
| **Claude `Agent`-tool subagent prompt** | ❌ **none.** Those prompts are composed by the model per prose in `audit-pr/SKILL.md`, `analyze-service/SKILL.md` etc. No committed code sees them | **no — and this is the honest gap** |

`preflight()` was the right host because it already had exactly the two-tier shape this needs: an **exact** check that hard-blocks (path containment, rc 3) beside an **over-matching** one that only warns (command globs). The citation check joins the first tier.

## The design

A brief declares its load-bearing claims in one fenced block:

````markdown
```claims
[
  {"claim":   "the preview DB is a clone, not production",
   "source":  "https://pr-4260.example.com/api/health",
   "read_at": "2026-08-21",
   "basis":   "measurement"},
  {"claim":   "the avatar refresh is driven by the localStorage roster",
   "source":  "src/components/AccountSwitcher.tsx:42",
   "read_at": "2026-08-21",
   "basis":   "inference"}
]
```
````

- **JSON, not YAML** — `opencode-dispatch` runs under a bare `/usr/bin/env python3` with no guaranteed site-packages; every other yaml user in this repo has to guard the import.
- **`basis` is a CLOSED ENUM** (`measurement` | `inference`). This is the specific distinction that failed in the homelab instance, so a misspelling is **refused**, never silently read as a measurement. The report lists each **inference** individually with its source and date and tells the receiving agent to re-verify it — folding them into a count would reproduce the erasure.
- **Two tiers, borrowed from `session_insight/schema.py` rather than reinvented**: `validate()` hard-fails (closed enum, missing/blank required field, parse error, two blocks); `key_warnings()` soft-fails on unknown citation keys and an unanchored `read_at`, so the vocabulary can grow (decision O2).
- **rc 6 `RC_UNCITED`, its own code.** rc 3 is fixed by widening `--dir`, rc 6 by citing a source; a caller conflating them would apply the wrong remedy. `run` refuses **before** `install_brief`, so "not dispatched" means nothing was written either.
- `preflight-uncited` added to `OUTCOMES` + adoption-scan's `REGISTRY` row and its `impact_outcomes` (the ledger test pins those both ways).
- **`[]` is a valid declaration** — an assertion on the record that the brief carries no load-bearing claim, printed as `NONE DECLARED`. Same shape as `schema.py`'s `unreadable=true` + reason.

### 🔴 What it structurally cannot see

It checks that **declared** claims carry sources. It cannot know your prose asserts a fifth premise you never declared — detecting a claim in free text needs a heuristic, a heuristic over-matches, and a permanently-red gate is worse than no gate (this file's own stated reason for why the glob check warns and the path check blocks). What the mandatory block buys is narrower and real: **you cannot dispatch without having been asked the question**, and a silence becomes an explicit `[]`.

## Test matrix

**RED AT BASE — behaviourally, not merely by a missing import.** The new lib was copied into an `origin/main` (`1d67f5e8`) checkout so the failures are about the CLI's behaviour rather than an absent module: **10 failed / 26 passed**.

The bare symptom, on the identical file both ways:

```
BASE_RC=0     # a brief asserting "dev database clone — safe to mutate;
HEAD_RC=6     # it is NOT production data", with zero sources
```

**GREEN AT HEAD:** `scripts/opencode/tests` **151 passed**; authoritative gate `nix build .#checks.x86_64-linux.pytests` → `RESULT: PASS (exit=0)`, `failed=0`.

**MUTATION SWEEP: KILLED 12/12, SURVIVED 0.** Run under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` deleted between mutants, each scored by the **names** of the tests that killed it, baseline green before and after. Every mutation is the narrowest expression that can be wrong (`if False` on the guard's own condition, never its enclosing branch).

| mutant | killed by |
|---|---|
| absent block fails open | `test_a_brief_with_no_claims_block_is_rejected` + 5 more |
| `basis` closed enum dropped | `test_an_out_of_set_basis_hard_fails_…` ×4 params |
| blank required field accepted | `test_a_blank_citation_field_is_not_a_citation` ×3 |
| soft tier silenced | `test_an_unknown_citation_field_warns_and_does_not_reject` |
| CLI refusal removed (rc 0) | `test_the_citation_guard_is_reachable_past_the_path_check` |
| every claim counted as a measurement | `test_the_distinction_is_preserved_through_the_counts` |
| `run` installs an uncited brief anyway | `test_run_refuses_an_uncited_brief_and_never_writes_it` |
| refusal / verdict sentences reworded (×3) | the literal two-way pins |
| close fence stops requiring the open length | `test_the_fence_grammar_is_pinned_in_both_directions[close-shorter]` |
| any fenced block counts as a declaration | `test_a_json_fence_is_not_mistaken_for_a_declaration` |

### 🔴 The positive control caught a real defect in the first sweep

The known-caught mutant — rewording the pinned refusal sentence — **SURVIVED a fully green suite.** The tests asserted `errs == [BC.NO_BLOCK_ERROR]`, so mutating the constant mutated the expectation with it. Seven other mutants died; that one walked, and only the positive control revealed it.

The sentences are now written out **literally** in the test file with a two-way pin — which is the idiom `test_dispatch.py` had **already** adopted for `VERDICT_CLEAN` after the identical finding. Reproducing it independently is the evidence that this is a **class**, not one file's mistake. It is recorded in the test file.

### Reachability, pinned as its own test

The path check runs **first** and returns first, so a citation-guard test could go green without the guard ever executing. `test_the_citation_guard_is_reachable_past_the_path_check` asserts the fixture trips **no** path offender (`external_paths == []`, `blocked is False`) **and** that the rc is 6 specifically.

The negative-control fixture is the **real text of a brief that was actually dispatched** (`20260821-185821-verify-pr-4260-…`), not a textbook one — its parenthetical *"(dev database clone — safe to mutate; it is NOT production data)"* is precisely the uncited load-bearing claim this exists for.

## Gate

`scripts/opencode/tests` 90 → **151** collected. Re-pinned **86 → 144** = `_suggested_floor 151`: a floor of 86 under a 151-test target carried **65 tests of slack**, more than the entire pre-#583 suite. `--check-floors` passes.

## 🔴 Cost — weighed, and stated

**This adds friction to every opencode dispatch.** Measured volume on this host: **6 briefs surviving on disk across 3 project directories over Aug 21–25** (~1.2/day). That is a **floor, not a total** — `.opencode-dispatch/` dirs are deleted with their projects, and the local telemetry spool had already been shipped. So the friction is roughly **one `claims` block per day**.

🔴 **The 1,921 `Agent`-tool calls in the audit window are NOT covered and NOT slowed.** That is the number that measures the *uncovered surface*, not the friction. Those prompts are composed by the model per prose in the skills, and there is **no committed code chokepoint** for them.

**The honest trade:** this covers a low-volume path well and leaves the high-volume path untouched. It is worth landing anyway — it is the only place a code-level check can exist today, and it establishes the schema and the vocabulary that a wider guard would reuse. But it should not be read as closing the error class.

## Follow-ups (each with a closing condition)

**1. Extend to the clawgate task body — this IS a work item, with a mechanical closing condition.**
Two producers (`repo-cos/clawgate.py:build_task_body(proposal)`, `initiatives/dispatch.py:build_task_body(view, recommendation)`) emit the text a devpod agent literally executes, and neither validates it. Both already have floored gate targets (`scripts/repo-cos/tests`, `scripts/initiatives/tests`), so the work is checkable without new machinery.
- *Closing condition (mechanical):* a merged PR in which **both** call sites produce their body through one shared validator — reusing `brief_claims` — with a ledger test asserting the set of `build_task_body` producers, failing when it **grows or shrinks**, plus one behavioural case per producer. Checked by: the existing `devrc-pytests` gate going green on that PR.

**2. Whether the `Agent`-tool surface gets a chokepoint at all — NOT yet a work item, and I am not minting one.**
Covering ~137 briefs/day means a PreToolUse hook on the Agent tool in `scripts/claude-hooks/`. That is a design decision with real blast radius, not a task someone can pick up, and I can name no mechanical check that would close it. Stating it plainly rather than filing an object nobody can close: **this needs Zach's direction before it is work.** The input he needs is in this PR — the 0.9% figure, its basis, and the fact that the evidence came from the uncovered side.

### DIFF
--- a/scripts/opencode/SKILL.md
+++ b/scripts/opencode/SKILL.md
@@ -11,11 +11,19 @@ context window. The point is *your* context budget: seven weeks of opencode cost
 
 ## The whole invocation
 
-```bash
+````bash
 opencode-dispatch run --dir <PROJECT-DIR> --title "<short title>" -m flash <<'BRIEF'
 <the complete task specification>
-BRIEF
+
+```claims
+[]
 ```
+BRIEF
+````
+
+🔴 **The `claims` block is REQUIRED — a brief without one is refused (rc 6).**
+See "Every brief carries its sources" below; `[]` is the honest declaration when
+the brief genuinely asserts no premise the agent would act on.
 
 That is it. Do **not** re-derive the CLI — `opencode --version`, `--help`,
 `run --help` and `models` were re-run from scratch in **nine** separate sessions
@@ -56,7 +64,76 @@ as well, so you can check them without dispatching.
 opencode-dispatch preflight --dir <PROJECT-DIR> --brief <FILE>   # or stdin
 ```
 
-`--json` for a machine-readable report. Exit **3** means the brief was refused.
+`--json` for a machine-readable report. Exit **3** (a path escape) or **6** (an
+uncited claim) means the brief was refused — see "When preflight refuses".
+
+## Every brief carries its sources
+
+🔴 **A brief with no `claims` block is REFUSED — rc 6, nothing written, nothing
+dispatched.** Declare each load-bearing claim: a fact the brief states that the
+agent will *act on*, and that would change the work if it were wrong.
+
+````markdown
+```claims
+[
+  {"claim":   "the preview DB is a clone, not production",
+   "source":  "https://pr-4260.example.com/api/health",
+   "read_at": "2026-08-21",
+   "basis":   "measurement"},
+  {"claim":   "the avatar refresh is driven by the localStorage roster",
+   "source":  "src/components/AccountSwitcher.tsx:42",
+   "read_at": "2026-08-21",
+   "basis":   "inference"}
+]
+```
+````
+
+All four fields are required on every entry. `basis` is a **closed enum** —
+`measurement` (you ran it, read it, saw the output) or `inference` (you concluded
+it). Nothing else; a misspelling is refused rather than read as a measurement.
+Extra fields are kept and warned about, never rejected.
+
+**`[]` is a valid, honest declaration** — "this brief asserts no load-bearing
+claim". It is an assertion on the record rather than a silence, and the report
+prints it as `NONE DECLARED`.
+
+### 🔴 What this covers — ~0.9% of the briefs this machine produces
+
+| surface | rate | covered? |
+|---|---|---|
+| opencode dispatch (this tool) | ~1.2/day | ✅ |
+| Claude `Agent`-tool subagent brief | ~137/day | ❌ no chokepoint exists |
+| clawgate `build_task_body()` | uncounted | ❌ |
+
+Basis: 6 briefs on disk across 3 project dirs over Aug 21–25 (a **floor** — the
+dirs are deleted with their projects) vs 1,921 `Agent` calls in the audit's
+14-day window. Different windows; treat it as an order of magnitude.
+
+🔴 **Every measured instance below came from the UNCOVERED surface** — the wrong
+root cause reached three *subagent* briefs; the three subagents correcting a
+stale brief were Agent-tool subagents. **Do not cite this guard as evidence that
+premise-propagation is handled.** It is a bridgehead on a real dispatch path, and
+the schema a wider guard would reuse.
+
+### Why it refuses instead of reminding
+
+A 14-day audit of 443 sessions found **wrong premises propagating into subagent
+briefs in four of six audit slices** — the highest-cost error class found, and
+one the adversarial audit ladder is structurally blind to. A session built an
+entire storage layer on a belief lifted from a stale comment and a stale README;
+**four audit rounds read past it**. A wrong root-cause diagnosis was pushed into
+**three** subagent briefs before being retracted. A homelab session *inferred* an
+auth constraint from a token prefix, reported it as fact, and it propagated into
+a downstream session's opening brief — "my inference presented as fact".
+
+A prose instruction is the mechanism that already failed there four times. So
+this is a schema, checked in code, that refuses.
+
+🔴 **What it cannot see, stated plainly:** it checks that *declared* claims carry
+sources. It cannot know your prose asserts a fifth premise you never declared —
+detecting a claim in free text needs a heuristic, a heuristic over-matches, and a
+permanently-red gate is worse than no gate. What the mandatory block buys is
+narrower and real: you cannot dispatch without having been asked the question.
 
 ## Reading the result
 
@@ -68,7 +145,7 @@ dependency is worth taking only once this skill is proven to get used. For now:
 tail -n 60 <the log path printed on dispatch>
 ```
 
-## Why this exists — four measured failure modes, three success-shaped
+## Why this exists — five measured failure modes, three success-shaped
 
 1. 🔴 **`external_directory: "ask"` → headless auto-reject → exit 0, no work.**
    The brief named a path outside `--dir`. `opencode run` **auto-rejects** an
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by ZacxDev** [comment]
**Local authoritative gate: PASS on the shipping tree.** Posting the gate's own lines because the floor re-pin (86 → 144) is the one reviewable risk in this PR, and a bump that outruns the real count would take the gate red:

```
  PASS  scripts/opencode/tests  (collected=151 passed=151 skipped=0 floor=144)
  TOTAL collected=16696  passed=16694  skipped=2  failed=0  (floor: 15847 = sum of 28 per-target floors)
PASS 48   FAIL 0
RESULT: PASS (exit=0)
```

`nix build .#checks.x86_64-linux.pytests`, run on the tree as committed **except** the final scope/coverage doc edits (pure comments + markdown), which were made after it started; `scripts/opencode/tests` was re-run green (151) after those. Tekton is the authoritative pre-merge gate and is running now.

Two notes for whoever reviews:

1. **The `~0.9%` framing is the substantive part of this PR, more than the code.** The guard covers opencode briefs (~1.2/day) and not the Agent-tool surface (~137/day) — and all four measured instances that motivated it came from the *uncovered* side. It is named and documented as a bridgehead so it cannot be cited as the class being handled. If that framing reads as overstated *or* understated, that is the thing to push back on.

2. **The mutation sweep's positive control caught a real defect in this PR's own tests.** The known-caught mutant (rewording the pinned refusal sentence) *survived* a fully green suite, because the tests asserted `errs == [BC.NO_BLOCK_ERROR]` — mutating the constant mutated the expectation with it. Seven other mutants died; that one walked. `test_dispatch.py` had already hit the identical trap with `VERDICT_CLEAN`, so this is a recurring class rather than one file's slip. Both are now literal two-way pins, and the final sweep is 12/12 killed, 0 survived.

ZacxDev:
Closing — not because the work is wrong, but because the measurement that justified it did not survive verification at the scale claimed.

The 14-day session-makework audit (`homelab-talos/claudedocs/audit-session-makework-2026-08-25.md`) put the citation-carrying surface at **~0.9% of the brief surface**, and — the part that actually decides it — **not the part the evidence came from**. The briefs whose unverified claims caused real rework were not the ones this change makes cite their sources.

That audit's own headline lesson is the reason this is a close rather than a merge: *every claim verified at the source held; every claim passed along unverified was wrong or overstated.* Three of five recommendation justifications in that batch collapsed under verification and one was wrong by 6x. This PR's justification is in the surviving-but-small category, and shipping 1,000+ lines against a 0.9% surface is the kind of work the audit was commissioned to stop.

The measurement is already captured in the audit report, so closing loses no evidence. **The branch `feat/opencode-brief-citations` is NOT deleted** — if the brief-citation surface ever grows, or a concrete rework incident traces back to an uncited brief claim, reopen against that incident rather than re-deriving the design.

Decision made 2026-08-26 with Zach, alongside #866 (closed) and #869 (held for a real-clone run).

PR #866, opened 2026-08-26
fix(subsystem-index): the --session window dropped every path a subagent touched in its own worktree

ZacxDev:
## The defect

The standing default here is to delegate file-modifying work to a subagent in an **isolated worktree**. A worktree is a directory outside the session cwd, so `changed_paths` had no repo-relative form for anything inside one and counted it as `outside_cwd` — **dropped**. The better a session followed the worktree rule, the less of it `--session` could report.

A worktree of a repo is not another tree that happens to spell alike: every worktree of one repository has the **same tracked layout by construction**, so `scripts/a.py` under a worktree *is* that repo-relative path. The attribution is structural, not a guess.

## What changed

- **`changed_paths.absolute_under_any(paths, roots)`** — the single-root absolute window, unioned over roots that share one path namespace. Deduped on the repo-relative form; a root whose own subtree overflows the cap is recomputed uncapped first, so a big root's lexicographic prefix cannot silently truncate the union.
- **`session-tailer`** — `absolute_extra_roots` plumbed through `build_rollup` / `summarize_transcript` / `_unreadable_rollup`. Opt-in on both names; `run()` passes neither, so the emitted telemetry payload is byte-for-byte unchanged (pinned by a test that now checks *both* parameter names).
- **`subsystem_touch.worktree_roots()`** — the root list, from `git worktree list --porcelain`. **Never** a `<repo>-*` naming convention: that invents an association for an unrelated project sharing a prefix *and* misses every `/tmp/wt-*` worktree, which is the shape this host mandates. `prunable` entries are **included deliberately** — a worktree whose directory was deleted but not pruned is still recorded with its original path, which is the ordinary end-state of a throwaway worktree.
- **Two windows widen**: absolute paths under any worktree of the repo are attributed and reported; and a session whose cwd *is* a worktree now gets the **full** session window instead of degrading to the `session-absolute` FLOOR, which excluded every relative path it named.
- **`worktree_recovered` is a third number**, not a correction to `under_cwd`/`outside_cwd` — those keep their cwd-frame meaning, and the dedup makes the recovery a *floor* on the outside-cwd entries explained rather than an exact subtraction. `wrong_window_dominance` nets it: warning "most of this session is invisible" off a raw count that includes paths printed underneath would be the mirror of the defect that block exists to break.

## Live before/after — the symptom reproduced and cleared

Base module and this one loaded side by side over real transcripts:

| | base | this branch |
|---|---|---|
| `/tmp/wt-appcapture-text` session | **0 paths**, `WRONG WINDOW (0,12,100%)` | **9 paths**, warning silent |
| `devrc-minio-preflight` session | **0 paths** | **2 paths** |
| a `claudedocs` pair | 1 path, `WRONG WINDOW 96%` | 3 paths, 89% |

## 🔴 Scope — the headline oversells the yield, and the code says so

Measured over the **100 most recent transcripts** *before* writing any code. Of **1,434** paths outside the session cwd:

| bucket | count | share |
|---|---:|---:|
| `/tmp` or scratchpad files (correctly not repo paths) | 1,245 | 87% |
| under an already-`git worktree remove`d worktree | 131 | 9% |
| under `~/.claude` | 86 | 6% |
| other non-repo directories | 14 | 1% |
| **recoverable by this window** | **18** | **1.2%** |

This fixes a real defect in a **small fraction** of the drop. The dominant cause of a thin session window on this host is **scratch files, not worktree blindness** — `--pr`/`--commit` remain the escalation. The module docstring, the skill and the `WRONG WINDOW?` block all state this so the larger claim is not re-derived from the headline.

**The boundary this does not cross:** a worktree removed with `git worktree remove` prunes its metadata, so git holds no record and no structural signal survives. That is the 131 above. Pinned by a test, and named in the caveat.

## Tests — red at `1d67f5e8`, green at HEAD

**20 new tests red at base**, verified by copying the test files into a detached worktree at `origin/main` with the sources unchanged. The reds are **behavioural, not `AttributeError`**:

- `paths` is `()` where the fixture expects the worktree paths
- `TranscriptCwdMismatchError` raised for a session that ran in a worktree
- `wrong_window_dominance` returns `(1, 2, 67)` where the netted answer is `None`, and `(1, 7)` where it should be `(2, 6)`

**2 tests are labelled INVARIANT GUARDS** — green at base by construction. They pin that the widening does not over-reach (a session in a genuinely different repo still has its relative paths refused) and that the extractor default still emits no absolute block. They are **not** counted as regression coverage.

**Mutation controls**, with the positive control reported as a pair:

```
positive control: unmutated module -> total=8 (cap), total=1 (dedupe)  OK
mutant 1 (union over the CAPPED list):        total=6   KILLED
mutant 2 (no dedup on the relative form):     total=2   KILLED
```

Run under `PYTHONDONTWRITEBYTECODE=1` so a same-length edit could not be scored SURVIVED off a stale `.pyc`.

## Gate

`scripts/run-tests.sh` → **RESULT: PASS (exit=0)**, 28 targets.
`scripts/tests` 8207 (floor 8036, drift ceiling 10045 — **no raise needed**); `scripts/collector/claude/tests` 147 (floor 110). `--check-floors` PASS.

The `GUARD 9`/`GUARD 10` "cannot attribute" lines are the documented shared-clone co-tenancy downgrade — six concurrent agents were writing `<devrc>/.git/config` via `git worktree add`. An earlier run failed GUARD 10 on `scripts/repo-cos/tests` (a suite this change does not touch) with the plugin unable to measure its own control; it did not reproduce.

## Not verified

- The **skill** correction is not live on merge alone: `~/.claude/skills/subsystem-index/SKILL.md` resolves into `/nix/store` (a `home.file` copy), so it needs a `home-manager switch` after this lands.
- The measurement is from **one host's** transcript corpus. The bucket *proportions* are a claim about this machine, not about every repo.
- The live before/after probed the **worktree's** code, which is exactly what this commit contains — but it is a claim about the deployed copy of these files, not independently about `main`.


ZacxDev:
Closing on measured benefit, not on correctness — the bug this fixes is real, it is just 1.2% of the thing it was believed to explain.

The 14-day session-makework audit (`homelab-talos/claudedocs/audit-session-makework-2026-08-25.md`) **RETRACTED** the premise that motivated this: *"worktree blindness explains WRONG WINDOW"* does not hold. Measured over 1,434 outside-cwd paths:

- worktree blindness accounts for **1.2%**
- **87%** are `/tmp` and scratchpad files — not worktrees at all

So the `--session` window is not meaningfully blind for the reason we thought, and the remaining 1.2% does not pay for a change that reaches into the telemetry extractor (`scripts/collector/claude/session-tailer.py`) — a shared path where a regression is silent and shows up as wrong data days later, in the one system we use to check our own claims.

The measurement is already captured in the audit report, so closing loses no evidence. **The branch `fix/stop-rederiving-environment` is NOT deleted** — if the `/tmp`-and-scratchpad 87% ever gets handled and worktree paths become the dominant remainder, this is the right fix to revive.

Decision made 2026-08-26 with Zach, alongside #864 (closed) and #869 (held for a real-clone run).

PR #868, opened 2026-08-26
fix(nolaunch): stamp every recorded launch with its xdist worker id — the KNOWN RACE closed without touching the runner's ledger

ZacxDev:
## The bug

`scripts/tests/test_no_real_launchers.py:535` carried a `🔴 KNOWN RACE, OPEN` comment. Since #841, `run-tests.sh` runs each pytest target `-n N --dist loadfile`. `loadfile` keeps **one file's** tests on one worker, so that file does not race itself — but ~8 sibling files in `scripts/tests` also invoke `systemctl`/`openrgb`/`systemd-run`/`notify-send`, run on **sibling** workers, and append to the **same** run-wide `$DEVRC_TEST_LAUNCH_STUB_DIR/launches.log`.

Every assertion in that file is

```python
before = len(nolaunch.blocked_systemctl(stub_dir))
...do the thing...
assert len(nolaunch.blocked_systemctl(stub_dir)) == before + 1
```

— a claim about **this worker's** launches, measured on a counter four workers increment. A few-ms window on a REQUIRED gate: it presents as an unrelated test failing for no reason. The workaround on file was `DEVRC_TEST_JOBS=1`; that still works, and is now a no-op rather than a special case.

## The fix

Both write paths stamp the writing worker's id into the line, as its **second** field:

```
systemctl(blocked) [gw2] --user stop foo
notify-send [main] -t 2500 rig-control …
i3-msg(abs) [gw0] workspace 9
```

* **L1**, the PATH stubs (`launcher_body`, `systemctl_body`) — `"${PYTEST_XDIST_WORKER:-main}"`. A serial run, and every NON-pytest target the runner starts, tag `main`: one namespace, no special case.
* **L2**, the in-process `subprocess.Popen` patch in `nolaunch_plugin.py` — `nolaunch.worker_tag()`. Left untagged, an absolute-path reach that *was* intercepted would read as one nobody ever saw.
* `_log_marker` is **deliberately left untagged** — `run-tests.sh` owns its per-target accounting (one marker per worker that ran a test). Consequence stated in the code: a filtered `recorded()` does not return markers; `ALL_WORKERS` does.

### `recorded()` / `blocked_systemctl()` — `worker=`, and why the default filters

Both now take `worker=`, defaulting to `THIS_WORKER`; `worker=ALL_WORKERS` returns the whole log, `worker="gw2"` asks about one worker.

The default filters on purpose. Every caller in the tree is the before/after shape above, so defaulting to unfiltered would leave the race armed for the next test somebody writes — the same shape as the bug. `ALL_WORKERS` keeps the unfiltered/report view available and is exercised by a test rather than left theoretical.

### Why a tag and not per-worker logs

The KNOWN RACE comment said the fix "needs the runner's ledger model changed too". **That is true of per-worker log FILES and false of a tag** — and it was checked, not assumed: `NOLAUNCH_SEEN+=("$d|$((nl_before+1))|$nl_after")` brackets a **whole** pytest invocation by line offset, and all N workers run inside that bracket for the **same** target `$d`, so an interleaved sibling-worker line already belongs to the correct target.

**`scripts/run-tests.sh` is BEHAVIOURALLY unchanged** — no ledger change, no classifier change, no new state; the gate runs green through it untouched. It does carry a **comment-only** edit: its xdist section pointed at "the KNOWN RACE note in `scripts/tests/test_no_real_launchers.py`", and that string no longer exists there, which would have left the runner as the only place in the repo still describing this race as open. (`test_doc_path_rot.py` cannot catch that — the path still resolves.) The rewritten comment says what is now true, why the tag may not move to line start, and why `NOLAUNCH_SEEN` needed no change.

The tag may not go at line **start**: the runner classifies this log with three `^`-anchored greps on the first token (`grep -c '^nolaunch(session)'`, `grep -c '^systemctl(read)'`, `grep -vE '^(nolaunch\(session\)|systemctl\(read\)|$)'`). A tag in front would reclassify the session marker and every read passthrough as a real launcher reach and fail the run.

## Call sites touched (and the ones deliberately left)

`git grep` for `nolaunch.recorded` / `nolaunch.blocked_systemctl` finds callers in exactly **two** files. (A wider grep for the bare word `recorded(` hits six more files — `test_subsystem_resolver.py`, `test_resume_state_clawgate.py`, `test_resume_state_handoff_resolution.py`, `browser-bridge/tests/test_server.py`, `dl-router/tests/test_learning.py`, `session-analysis/tests/test_insights.py` — all false positives: they are *test names* containing the word, and `test_resume_state_clawgate.py` defines its own local `recorded(stdout)` helper. None of them imports `nolaunch`.)

**Changed (behaviour follows the new default — all are before/after counts made by THIS worker):**

| site | why |
|---|---|
| `test_no_real_launchers.py:421,431` `test_invoking_a_launcher_records_and_launches_nothing` | before/after count; the exact-line assertion now pins `[worker]` too |
| `:524,534` `test_a_mutating_systemctl_verb_is_recorded_and_swallowed` | **the site the KNOWN RACE comment was attached to** |
| `:585,593` `test_an_argument_that_spells_a_read_verb_does_not_promote_a_mutation` | same shape, same exposure — 10 parametrised cases |
| `:637,665` the two `any(ln.startswith("systemctl(read)"))` reads | this worker made the read; filtering is the more precise claim |
| `:711,721 / :750,767 / :781,789 / :844,862` the four seam tests | `[before:]` slices plus `len(...) == 1` filters on `systemd-run ` — measured, 8 foreign `systemd-run` lines per run at `-n 4` |

**Behaviour changed, source line NOT edited — `test_no_real_launchers_all_targets.py:379,386`.** Listing these under both headings was a contradiction in an earlier draft of this body; this is the accurate statement. Both calls now return this worker's lines because the *default* changed; nothing at the call site moved. They are **not** cross-worker reads: they sit in `test_the_absolute_path_reach_is_intercepted_in_process`, which launches `<tmp>/i3-msg` from the pytest process itself and reads back the L2 `i3-msg(abs)` line it just caused — same worker, same process — so the filter does not narrow them. Mutant **M4** (L2 write path untagged) kills exactly that test, which is the proof it still sees what it is supposed to. The rest of that file is about the *runner* covering all targets: those tests drive a nested `run-tests.sh` and assert on its **stdout**, not on `recorded()`.

**Left as-is:**

* **`_log_marker`** — see above.
* **`scripts/run-tests.sh`'s ledger and classifiers** — no change needed; see the argument above. (A **comment-only** edit did land in that file — same section.)

One assertion in `test_no_real_launchers_all_targets.py` did need updating: `assert "notify-send NOLAUNCH-PROBE planted reach" in out` reads the *runner's echo of the log line*. It is now a regex (`notify-send \[[^\]\s]+\] NOLAUNCH-PROBE planted reach`) because the tag there is whatever `PYTEST_XDIST_WORKER` the nested runner inherited — `gw1` under the gate, `main` under a bare serial `pytest`. A literal would pin the harness's own parallelism instead of the guard's report.

## Coverage

The window is a few ms and the original author got NOT REPRODUCED in 2/2 runs, so nothing here races for it: the foreign line is **injected** between the sample and the assertion, which is the same state the race produces and is reached every run.

* **`test_a_sibling_workers_line_does_not_count_as_this_workers_launch`** — REGRESSION. Injects a sibling worker's `notify-send` **and** `systemctl(blocked)` into the log between `before` and the assertion, in a private stub dir (fabricating a line in the run-wide log would be counted by GUARD 7 as a real reach and would move every other test's `before`). Carries a positive control: the `ALL_WORKERS` view must **see** the foreign lines, or filtering them out proves nothing.
* **`test_the_stub_takes_its_tag_from_the_xdist_worker_variable`** — the mechanism at both ends of the dimension it depends on: `PYTEST_XDIST_WORKER` set → that id; unset → `main`.
* **`test_the_readers_default_to_this_worker_and_can_still_see_everything`** — the `worker=` contract against a hand-written log, including that the untagged session marker belongs to no worker and that an argv containing a bracketed word is not mistaken for the tag.
* **`test_the_in_process_layer_tags_the_line_with_THIS_worker`** — forces a worker id no serial fallback can produce, so a hardcoded L2 tag cannot pass by coincidence under `DEVRC_TEST_JOBS=1`.
* ⚠ **`test_the_runners_anchored_classifiers_still_match_the_tagged_format` is an INVARIANT GUARD, not regression coverage** — it is green at `origin/main` too, where the tag does not exist and the classifiers match trivially. It pins the constraint the tag's *placement* had to satisfy. It **parses** the three grep patterns out of `run-tests.sh`'s GUARD 7 block (pinning the set both ways, so a classifier appearing *or* disappearing is red) and runs **real `grep`** against a log written entirely by this tree's own code — including the `systemctl(read)`/`systemctl(blocked)` lines, generated through `systemctl_body` pointed at a passthrough binary of our own so all three of its branches run in **both** tiers. Nothing in Python covered the runner's classifiers before this.

  This guard has been wrong twice, both times in the same direction — **its description wider than its pin** — and both are worth reading before trusting it:

  1. An early draft hand-wrote the `systemctl(read)` and `nolaunch(session)` lines. It **stayed green under a tag-at-line-start mutant**: those two are the only lines the runner classifies by their first token, so a hand-written copy stays in the old format under the very mutation the test exists to catch.
  2. A blind audit then found it *still* narrower than its docstring: it asserted the tag over `hits` only, and a read is by construction **not** a hit — so deleting `[%s]` from the read branch survived the **gating** tier at 74 passed / exit 0 while the identical mutation of its `systemctl(blocked)` sibling was killed. It now asserts the tag on the read lines directly, and drives the `--version`/`--help` flag branch too, which no test in this tree exercised at all.

  Both are pinned by mutants (M3c, M7, M7b) rather than by the wording.

## Verification

See the PR discussion for the full red/green matrix, both tiers, both `DEVRC_TEST_JOBS` settings, and the 15-mutant battery.

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -2889,8 +2889,22 @@ _count_of() { # $1 = alternation regex, $2 = summary line
 # ⚠ What loadfile does NOT buy, stated so nobody reads more into it: it pins
 # INTRA-file co-location only. State shared ACROSS files — the run-wide launch
 # log, the spool ledger, the nogit guard config — is still touched by every
-# worker concurrently. See the KNOWN RACE note in scripts/tests/
-# test_no_real_launchers.py.
+# worker concurrently.
+#
+# For the LAUNCH LOG that is now handled rather than merely known: every line it
+# records carries the writing worker's id as its SECOND field
+# (`systemctl(blocked) [gw2] …`), and `testlib.nolaunch`'s readers filter on it,
+# so a test's `before + 1` is a claim about its OWN worker again. Read
+# `scripts/testlib/nolaunch.py`'s module docstring before changing the log
+# format — in particular, the tag may NOT move to line start, because the GUARD
+# 7 evaluation block below classifies this file with ^-anchored greps on the
+# FIRST token. NOTHING IN THIS SCRIPT NEEDED TO CHANGE for that fix, and the
+# reason is `NOLAUNCH_SEEN`: it brackets a whole pytest invocation by line
+# offset and all N workers run inside that bracket for the SAME target, so an
+# interleaved sibling-worker line already belongs to the right target.
+#
+# The spool ledger and the nogit guard config are NOT covered by that and remain
+# as described above.
 #
 # 🔴 NESTED RUNS MUST BE SERIAL, and the signal is PYTEST_CURRENT_TEST.
 # Tests in scripts/tests SPAWN THIS SCRIPT (test_run_tests_floors.py and the .sh
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -2891,10 +2891,13 @@ _count_of() { # $1 = alternation regex, $2 = summary line
 # log, the spool ledger, the nogit guard config — is still touched by every
 # worker concurrently.
 #
-# For the LAUNCH LOG that is now handled rather than merely known: every line it
-# records carries the writing worker's id as its SECOND field
+# For the LAUNCH LOG that is now handled rather than merely known: every
+# INTERCEPT line it records carries the writing worker's id as its SECOND field
 # (`systemctl(blocked) [gw2] …`), and `testlib.nolaunch`'s readers filter on it,
-# so a test's `before + 1` is a claim about its OWN worker again. Read
+# so a test's `before + 1` is a claim about its OWN worker again. ⚠ The one
+# exception, because `recorded()`'s contract is careful about it and this
+# comment must be too: the `nolaunch(session)` MARKER is deliberately UNTAGGED —
+# its attribution is owned right here, by the per-target count below. Read
 # `scripts/testlib/nolaunch.py`'s module docstring before changing the log
 # format — in particular, the tag may NOT move to line start, because the GUARD
 # 7 evaluation block below classifies this file with ^-anchored greps on the
--- a/scripts/testlib/nolaunch.py
+++ b/scripts/testlib/nolaunch.py
@@ -209,9 +209,27 @@ SERIAL_WORKER = "main"
 THIS_WORKER = "<this>"
 ALL_WORKERS = "<all>"
 
-# `<classifier> [<tag>] <argv…>`. Anchored, and the tag may not contain a space
-# or a `]`, so an argv that happens to contain a bracketed word cannot be read
-# as the tag.
+# `<classifier> [<tag>] <argv…>` — the tag is the SECOND field and nothing else
+# is. What each piece actually does, stated at the width it is pinned to (see
+# `test_the_readers_default_to_this_worker_and_can_still_see_everything`):
+#
+#   `^\S+ `      the classifier token, then exactly one space. This is what
+#                stops a bracketed word LATER in the argv from being read as the
+#                tag — `notify-send [gw1] -t 2500 [not-a-tag]` yields `gw1`.
+#   `[^\]\s]+`   the tag itself: no `]` (it would swallow the delimiter) and no
+#                whitespace. The `\s` half is the conservative parse of a
+#                MALFORMED field, not a live case: pytest-xdist ids are `gw<N>`
+#                and `SERIAL_WORKER` is a literal, so neither can contain a
+#                space. ⚠ THE BOUNDED CONSEQUENCE, stated because it is a silent
+#                zero and not obvious: if somebody exports
+#                `PYTEST_XDIST_WORKER="gw 1"`, the stub writes `[gw 1]` and this
+#                reads the line as UNTAGGED, so a filtered `recorded()` will not
+#                return it even though `worker_tag()` would say `gw 1`. That
+#                asymmetry is pinned rather than fixed — normalising here would
+#                have to be mirrored in the sh stub, and two spellings of one
+#                rule is the bug this module already exists to avoid.
+#   trailing ` ` the delimiter must be followed by the argv separator, so
+#                `notify-send [gw1]nospace` is not a tagged line.
 _TAGGED = re.compile(r"^\S+ \[([^\]\s]+)\] ")
 
 # The POSIX-sh expression the stubs expand to get their own tag. Assembled from
@@ -232,6 +250,14 @@ def worker_tag() -> str:
     Read at call time, never cached: `nolaunch` is imported once per pytest
     worker but the value is a property of the process, and a cached module-level
     constant would be captured before a nested session could change it.
+
+    🔴 `or`, NOT `os.environ.get(WORKER_ENV, SERIAL_WORKER)`. The two differ on
+    exactly one input — the variable SET TO EMPTY — and `or` is the one that
+    agrees with the sh stub's `${PYTEST_XDIST_WORKER:-main}`, which also falls
+    back on empty. `get(..., default)` would make Python read `""` while the
+    stub wrote `[main]`, so every filtered read would return nothing while the
+    log looked perfectly well-formed. Pinned by
+    `test_the_stub_takes_its_tag_from_the_xdist_worker_variable`.
     """
     return os.environ.get(WORKER_ENV) or SERIAL_WORKER
 
--- a/scripts/testlib/nolaunch.py
+++ b/scripts/testlib/nolaunch.py
@@ -99,9 +99,55 @@ The principle is NOT "never install a stub where no real binary exists" — it i
 (or one with only flags) has no verb, so it is recorded and swallowed with exit
 0 and no output, where the real binary would list units. Every call site in
 this tree passes a verb.
+
+🔴 EVERY LINE CARRIES THE WRITER'S pytest-xdist WORKER ID — AND WHY
+-------------------------------------------------------------------
+There is ONE run-wide log, shared by every target and, since #841, by every
+xdist worker inside a target. `run-tests.sh` runs each pytest target
+`-n N --dist loadfile`. `loadfile` keeps one FILE's tests on one worker, so a
+file does not race itself — but the ~8 sibling files in `scripts/tests` that
+also invoke `systemctl`/`openrgb`/`systemd-run`/`notify-send` run on SIBLING
+workers and append to this same file. Measured composition of the log for a
+9-file subset at `-n 4`: 23 `systemctl(blocked)`, 9 `openrgb`, 8 `systemd-run`,
+8 `notify-send` foreign lines per run.
+
+That breaks the shape every assertion here uses:
+
+    before = len(blocked_systemctl(stub_dir))   # sample
+    ...do the thing...                          # <- a sibling worker appends
+    assert len(blocked_systemctl(stub_dir)) == before + 1
+
+`before + 1` is a claim about THIS worker's launches, measured against a
+counter every worker increments. A few-ms window on a REQUIRED gate, and it
+presents as an unrelated test failing for no reason.
+
+So the stub stamps the tag INTO the line and the readers filter on it. Two
+things that look arbitrary are not:
+
+  * THE TAG IS THE SECOND FIELD, NEVER THE FIRST. `run-tests.sh` classifies
+    this log with THREE `^`-anchored greps on the FIRST token — one counting
+    `nolaunch(session)` markers, one counting `systemctl(read)` passthroughs,
+    and one selecting everything that is neither as a launcher HIT — and
+    `blocked_systemctl` below uses `startswith("systemctl(blocked)")`. A tag at
+    line start would silently reclassify the marker and every read as a real
+    launcher reach and fail the run. The patterns themselves are not restated
+    here: `test_the_runners_anchored_classifiers_still_match_the_tagged_format`
+    reads them OUT of run-tests.sh and runs them against a log this tree wrote.
+  * THE RUNNER'S LEDGER NEEDS NO CHANGE. `NOLAUNCH_SEEN` brackets a WHOLE
+    pytest invocation by line offset (`nl_before+1 .. nl_after`), and all N
+    workers run inside that bracket for the SAME target, so an interleaved
+    sibling-worker line already belongs to the correct target. That is why
+    this is a tag in one shared log rather than per-worker log FILES — those
+    WOULD need the ledger model changed.
+
+Serial runs (`DEVRC_TEST_JOBS=1`, a bare `pytest`, and every NON-pytest target
+the runner starts) have no `PYTEST_XDIST_WORKER`, so they tag `main` — one
+namespace, and the filter is a no-op there rather than a special case.
 """
 from __future__ import annotations
 
+import os
+import re
 import shutil
 from pathlib import Path
 
@@ -145,10 +191,60 @@ SYSTEMCTL_READ_FLAGS = ("--version", "--help", "-h")
 
 LOG_NAME = "launches.log"
 
+# pytest-xdist exports this into each WORKER process, and a child shell inherits
+# it. MEASURED (6 files, `-n 4 --dist loadfile`): four distinct ids `gw0`..`gw3`
+# reached a `sh -c` child; the same suite run serially left it UNSET there.
+WORKER_ENV = "PYTEST_XDIST_WORKER"
+
+# The tag used when `WORKER_ENV` is absent: a serial pytest run, or one of the
+# NON-pytest targets (`HOOK_TESTS`, `SHELL_TESTS`) that reach the stubs through
+# PATH with no plugin at all. A name, not an empty string, so every line has a
+# tag in the same position and the parser never has to special-case one.
+SERIAL_WORKER = "main"
+
+# `worker=` sentinels for the readers below. Neither can collide with a real tag
+# — pytest-xdist ids are `gw<N>` and `SERIAL_WORKER` is a bare word — so passing
+# one can never be confused with asking for a worker that happens to be called
+# that.
+THIS_WORKER = "<this>"
+ALL_WORKERS = "<all>"
+
+# `<classifier> [<tag>] <argv…>`. Anchored, and the tag may not contain a space
+# or a `]`, so an argv that happens to contain a bracketed word cannot be read
+# as the tag.
+_TAGGED = re.compile(r"^\S+ \[([^\]\s]+)\] ")
+
+# The POSIX-sh expression the stubs expand to get their own tag. Assembled from
+# the constants above so the shell and the Python reader cannot drift apart on
+# the variable name or the fallback — the one thing a copy would get wrong
+# silently, because a stub tagged `main` under xdist still LOOKS well-formed.
+_WORKER_SH = '"${' + WORKER_ENV + ':-' + SERIAL_WORKER + '}"'
+
 
 def log_path(stub_dir: Path) -> Path:
     """Where `install(stub_dir)` records intercepted argv, one line per launch."""
     return Path(stub_dir) / LOG_NAME
+
+
+def worker_tag() -> str:
+    """The tag THIS process's launches are stamped with.
+
+    Read at call time, never cached: `nolaunch` is imported once per pytest
+    worker but the value is a property of the process, and a cached module-level
+    constant would be captured before a nested session could change it.
+    """
+    return os.environ.get(WORKER_ENV) or SERIAL_WORKER
+
+
+def line_worker(line: str) -> str | None:
+    """The worker tag a recorded line carries, or None if it carries none.
+
+    None is the honest answer for the plugin's `nolaunch(session)` marker, which
+    is deliberately NOT tagged: `run-tests.sh` counts it per TARGET (one per
+    worker that ran a test) and owns that accounting.
+    """
+    m = _TAGGED.match(line)
+    return m.group(1) if m else None
 
 
 def launcher_body(name: str, log: Path) -> str:
@@ -161,10 +257,19 @@ def launcher_body(name: str, log: Path) -> str:
 
     🔴 And RECORDING IS ALL IT DOES. The guard test pins this text against a
     literal copy of itself, because a stub that records correctly AND has a side
-    effect passes every behavioural assertion in the suite.
+    effect passes every behavioural assertion in the suite. That pin lives in
+    `scripts/tests/test_no_real_launchers.py::test_a_stub_does_nothing_but_record`
+    and spells the expectation out by hand — deriving it from here would make a
+    mutation of this function invisible to it.
+
+    🔴 The `[$worker]` field is the second one, never the first: `run-tests.sh`
+    classifies this log with `^`-anchored greps on the FIRST token. See the
+    module docstring.
     """
-    return 'printf \'%s %s\\n\' "{name}" "$*" >> "{log}"\nexit 0\n'.format(
-        name=name, log=log)
+    return (
+        "printf '%s [%s] %s\\n' \"{name}\" {worker} \"$*\" >> \"{log}\"\n"
+        "exit 0\n"
+    ).format(name=name, worker=_WORKER_SH, log=log)
 
 
 def systemctl_body(real: str, log: Path) -> str:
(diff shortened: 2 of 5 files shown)


ZacxDev:
## Verification

Base ref for every claim below: **`origin/main` = `a58a261db6a621d061c3e229b89bb1958048f58c`** (the branch was rebased onto it; the fix's own base sha before the rebase was `200e6383`).

### Red at base / green at HEAD

| what | at base (`a58a261d`, this file's tests on the base `nolaunch`) | at HEAD |
|---|---|---|
| whole file `test_no_real_launchers.py` | **28 failed, 46 passed** | **74 passed** |
| `test_a_sibling_workers_line_does_not_count_as_this_workers_launch` | RED | green |
| `test_the_stub_takes_its_tag_from_the_xdist_worker_variable` | RED | green |
| `test_the_readers_default_to_this_worker_and_can_still_see_everything` | RED | green |
| ⚠ `test_the_runners_anchored_classifiers_still_match_the_tagged_format` | RED *(only because the module has no `worker_tag`; it is an INVARIANT GUARD — see below)* | green |
| `test_a_stub_does_nothing_but_record` ×12, `test_invoking_a_launcher_records_and_launches_nothing` ×12 | RED (format pins) | green |

**Stated plainly, because "red at base" is weaker here than it looks:** at `origin/main` the module has no per-worker concept, so the three new tests die on `AttributeError` — a red that proves the API is new, not that the filter works. The red that carries the meaning is mutant **M1**, which restores `origin/main`'s *semantics* (`recorded()` accepts `worker=` and ignores it) while keeping the API. Against M1 the regression test fails on its own assertion:

```
AssertionError: a SIBLING xdist worker's launches were counted as this worker's.
That is the race: `before + 1` is a claim about what THIS test launched, and it
was measured against every worker's lines: [...]
```

### Mutation sweep

Every mutant got its **own** freshly-copied tree, every `__pycache__` deleted, `PYTHONDONTWRITEBYTECODE=1`, and the replacement **count asserted before the run** (a mutant that did not apply aborts rather than scoring SURVIVED). `M0` is the unmutated control.

| mutant | what it breaks | verdict | killed by |
|---|---|---|---|
| **M0-control** | nothing | GREEN (29 passed) — the harness is not red-by-default | — |
| **M1-filter-ignored** | `recorded()` ignores `worker=` — *origin/main's semantics* | RED | the regression test, with the race message above |
| **M2-tag-hardcoded-main** | stub stops reading `$PYTEST_XDIST_WORKER`, hardcodes `main` | RED | `test_the_stub_takes_its_tag_from_the_xdist_worker_variable` + the 12 literal body pins |
| **M3-tag-at-line-start** | tag moved to the front of the line | RED (27 tests) | broad |
| **M3b-…-isolated** | tag to the front **and** `_TAGGED` **and** the literal pin moved with it | RED | incl. the classifier guard: `AssertionError: … ['[main] notify-send a-toast', …]` |
| **M3c-read-tag-at-line-start** | **narrowest possible** — only the `systemctl(read)` line's tag moves | RED, **and ONLY the classifier guard fires**: `AssertionError: a systemctl READ stopped matching the runner's ^-anchored count and would now be reported as a real launcher reach — assert ['0'] == ['1']` | `test_the_runners_anchored_classifiers_still_match_the_tagged_format` |
| **M4-L2-untagged** | the `Popen` patch writes `%s(abs) %s` again | RED, **only** `test_the_absolute_path_reach_is_intercepted_in_process` | that test |
| **M5-default-all** | `recorded()` defaults to `ALL_WORKERS` | RED | the regression test + the contract test |
| **M6-blocked-drops-worker** | `blocked_systemctl` forwards `ALL_WORKERS` instead of its argument | RED | the regression test + the contract test |

M3c and M4 are the reachability proofs that matter: each is isolated enough that exactly one test fires, **with that test's own error message**, so neither is green-for-the-wrong-reason.

⚠ **An earlier draft of the classifier guard SURVIVED a tag-at-line-start mutant** — it hand-wrote the `systemctl(read)` and `nolaunch(session)` lines, which are the only two the runner classifies by their first token, so the hand-written copies stayed in the old format under the very mutation the test existed to catch. Fixed before this PR: every line in that test is now produced by this tree's own code (the `systemctl` stub is generated by `systemctl_body` pointed at a passthrough binary of our own, so both branches run in both tiers). M3c is the proof the fix works.

### Both tiers, both `DEVRC_TEST_JOBS` settings

| tier | command | parallelism | result |
|---|---|---|---|
| dev host | `scripts/gate.sh --tier both` | `-n 4 --dist loadfile` | **`GATE: RESULT=PASS exit=0`** — pytest `collected=16728 passed=16726 skipped=2 failed=0`; node `tests=1292 pass=1292 fail=0` |
| nix sandbox (the tier Tekton gates on) | `nix build .#checks.x86_64-linux.pytests .#checks.x86_64-linux.nodetests` | sandbox default | **both derivations built** (`devrc-pytests`, `devrc-nodetests`) |
| Tekton, on this PR's head | — | — | `tekton/devrc-pytests` **pass**, `tekton/devrc-nodetests` **pass** |
| dev host, serial | `DEVRC_TEST_JOBS=1 scripts/gate.sh --tier pytest` | serial (`-n` absent) | **0 test failures** (`collected=16728 passed=16725 skipped=3 failed=0`) |

**On the serial run, stated rather than glossed:** it exits 1, and *not* for a test failure. `run-tests.sh`'s GUARD 2 (skip accounting) sees an unpinned skip that only ever happens serially — `scripts/tests/test_gitenv_sibling_exclusion.py:92: only meaningful inside a real xdist worker` — against an `EXPECTED_SKIPS` list of 2. That is a property of `origin/main`, not of this change: `run-tests.sh` and that test file are **byte-identical** on this branch (`git diff origin/main -- scripts/run-tests.sh scripts/tests/test_gitenv_sibling_exclusion.py` is empty). Not fixed here; flagged. The same run at the pre-rebase base (`200e6383` + this commit) produced the identical shape — `failed=0`, exit 1 on the same unpinned skip.

So `DEVRC_TEST_JOBS=1` still works exactly as before this PR — including the way it was already failing GUARD 2.

### Things that went red and were NOT this change

* `scripts/claude-hooks/tests/test_gh_issue_closing_condition_guard.py` — 3 xdist collection errors ("Different tests were collected between gw0 and gw1"), from a module-level `tempfile.mkdtemp()` baked into parametrize IDs. Reproduced against a **base-`nolaunch`** tree at `-n 4`, so not this change — and already fixed on `main` by **#855**, which is why this branch was rebased.
* `test_browser_agent.py::test_readiness_probe_does_not_inject_into_the_page` — one 60s hang, once in three full gate runs of this branch, green on a targeted `-n 4` re-run (62 passed) and green in the final gate. Not load: the target's wall time was 169.71s vs 163.80s on a passing run, and the test's own spawn probe measured 0.16s against a 0.80s stall threshold. Not this change either: GUARD 7 reported `scripts/browser-bridge/tests intercepted=0` in the failing run, i.e. **no nolaunch stub executed in that target at all**, and that file does not import `nolaunch`. The file has prior timing-flake history (#461, #500).

ZacxDev:
## Audit round — all six findings addressed

Commit `9a0b683b`. Base `origin/main = a58a261d`. A fix made in response to a review is a code change like any other, so the **whole** 15-mutant battery was re-run, not just the mutants these address.

| # | finding | status |
|---|---|---|
| 🟡 1 | `run-tests.sh:2892` dangling pointer to the closed race | **fixed** |
| 🟡 2 | `systemctl(read)` tag mutant survives the gating tier | **fixed**, and re-killed *in that tier's condition* |
| 🟢 3 | docstrings claim the grep patterns are "read out of run-tests.sh" | **fixed by making it true** |
| 🟢 4 | `worker_tag()`'s empty-string boundary unmeasured | **fixed** |
| 🟢 5 | `_TAGGED`'s comment wider than what is pinned | **fixed** |
| 🟢 9 | PR body contradiction | **fixed** |
| nit 6 | serial tier blind to a *wrong* L2 tag | **fixed** |
| nit 7 | `SYSTEMCTL_READ_FLAGS` branch exercised by nothing | **fixed** |
| nit 8 | `_private_stubs` chains into the session stub dir | **fixed** |

---

### 🟡 1 — the runner's dangling pointer

You are right that it was the only place left in the repo calling this race open, and that `test_doc_path_rot.py` cannot see it (the path still resolves). Rewritten to state what is now true, why the tag may not move to line start, and why `NOLAUNCH_SEEN` needed no change — plus the explicit note that the spool ledger and nogit config are **not** covered by the tag and remain as described.

The PR body headline now reads **"`scripts/run-tests.sh` is BEHAVIOURALLY unchanged"** — no ledger change, no classifier change, no new state — with the comment-only edit called out on the same line. That is the claim the gate actually established.

### 🟡 2 — the surviving read-branch mutant

Diagnosis confirmed exactly as you wrote it: the guard asserted the tag over `hits`, and a `systemctl(read)` is by construction *not* a hit, so its docstring ("BOTH its branches run in BOTH tiers") described more than it pinned.

Now asserted directly on the read lines. **And both read branches are covered**, because the flag branch (`--version`/`--help`) has its **own printf** and is therefore its own mutation site — that is nit 7, folded in here rather than left, since it would otherwise have been the next survivor of the same shape.

**Re-killed in the tier that matters, not just locally.** Rather than spend ten minutes per mutant on `nix build`, I reproduced the sandbox's *distinguishing property* — every PATH entry holding a `systemctl` removed, so `shutil.which("systemctl")` is `None` and `install()` writes no systemctl stub — with a positive control asserting the condition was really established before the run (`no-systemctl condition established (dropped from PATH: ['/run/current-system/sw/bin'])`).

* clean tree, no-systemctl condition: **90 passed**
* `M7-read-verb-tag-dropped`: **RED**, killed by `test_the_runners_anchored_classifiers_still_match_the_tagged_format`:
  ```
  AssertionError: a `systemctl(read)` line lost its [worker] tag. It still classifies
  correctly for the runner, so nothing in run-tests.sh notices — but every per-worker
  read drops it silently: ['systemctl(read) --user status u', 'systemctl(read) [main] --version']
  assert [None, 'main'] == ['main', 'main']
  ```
* `M7b-read-flag-tag-dropped`: **RED**, same guard, mirrored:
  ```
  … ['systemctl(read) [main] --user status u', 'systemctl(read) --version']
  assert ['main', None] == ['main', 'main']
  ```

The two failures are each other's control: each names which branch lost the tag while the other kept it.

### 🟢 3 — the patterns are now actually parsed

`_guard7_grep_patterns()` slices `run-tests.sh` to the GUARD 7 evaluation block by its own section banners (so a `grep -c` belonging to GUARD 8 or 10 cannot be mistaken for one of these), extracts every `grep -c '…'` and `grep -vE '…'`, and pins the set **both ways**. The substring version could not notice a classifier being *added* — mutant **M13** adds a fourth one to that block and is green under the old form, red under this one.

### 🟢 4 / 🟢 5 — the unmeasured boundaries

* `worker_tag()`: a third measured point — the variable **set to empty** — plus the reason in the code (`${VAR:-main}` falls back on empty, `get(k, default)` does not, so the two spellings would disagree while the log still looked well-formed). Mutant **M8**.
* `_TAGGED`: each piece is now pinned by its own case, and the comment states what each piece does at the width it is pinned to. Mutants **M9** (whitespace exclusion), **M10** (trailing space), **M11** (`]` exclusion).

  One correction to my own first attempt, worth recording: I initially wrote the `]`-exclusion case as `notify-send [gw]1] hello` expecting `"gw"`. It is `None` — the regex needs `] ` and gets `]1`. The discriminating input is `notify-send [gw]x] hello`, which yields `gw]x` without the exclusion and `None` with it. Caught by running it, not by reading it.

  I did **not** "fix" the whitespace asymmetry, and the comment says so explicitly rather than implying coverage: if someone exports `PYTEST_XDIST_WORKER="gw 1"`, the stub writes `[gw 1]` and `line_worker` reads the line as untagged, so a filtered `recorded()` drops it while `worker_tag()` would say `gw 1`. Normalising in Python would have to be mirrored in the sh stub, and two spellings of one rule is the bug this module exists to avoid. Bounded (xdist ids are `gw<N>`, the fallback is a literal), stated, pinned.

### nit 6 — the serial tier's blind spot

Agreed and fixed. `test_the_in_process_layer_tags_the_line_with_THIS_worker` sets `PYTEST_XDIST_WORKER` to a value no serial fallback can produce, for the duration of one in-process launch, so the assertion is identical under `-n 4` and `DEVRC_TEST_JOBS=1` — the guard is not a hostage to how the suite happens to be invoked. Mutant **M12** (L2 hardcodes `SERIAL_WORKER`) is killed by it; without it, that mutant is invisible to a serial run.

### nit 8 — judged worth fixing, not just documenting

You called it inert today and that is right, but the reason it is inert is a coincidence (the classifier test overwrites the file first) and the failure mode if it stopped being inert is a *silently passing* test measuring the wrong log. So: `_private_stubs` now re-points the systemctl stub at the real binary resolved off the **ambient** PATH, or removes it when there is none — which is what `install()` would have written on a PATH without a stub dir on it, and keeps the module's "never fabricate an answer" rule. An assertion in the helper names the hazard. Mutant **M14** reverts it to a plain `install()` and is killed.

Stated limit: that assertion can only fire on a host that *has* a real systemctl. In the sandbox there is none, so `install()` writes no stub and there is nothing to check — the guard is dev-host-only by construction.

---

### Re-run battery — 15 mutants, all killed, control green

`M0-control: GREEN (30 passed)` — the harness is not red-by-default. Each mutant gets its own freshly-copied tree, every `__pycache__` deleted, `PYTHONDONTWRITEBYTECODE=1`, replacement count asserted before the run.

| mutant | killed by |
|---|---|
| M1 filter ignored (origin/main semantics) | the regression test, with the race message |
| M2 tag hardcoded `main` | the worker-variable test + 12 literal body pins |
| M3 / M3b tag at line start | broad / incl. the classifier guard |
| M3c **read tag moved** to line start (narrowest) | classifier guard **alone** |
| **M7 read VERB tag dropped** | classifier guard **alone** — message above |
| **M7b read FLAG tag dropped** | classifier guard **alone** — message above |
| M4 L2 untagged | the two absolute-path tests |
| **M12 L2 hardcoded serial tag** | `…tags_the_line_with_THIS_worker` **alone** |
| M5 default `ALL_WORKERS` | regression + contract tests |
| M6 `blocked_systemctl` drops its worker arg | regression + contract tests |
| **M8 `get(k, default)`** | worker-variable test **alone** |
| **M9 / M10 / M11 `_TAGGED` pieces** | contract test **alone**, each |
| **M13 fourth GUARD 7 classifier** | classifier guard **alone** |
| **M14 `_private_stubs` chains to session** | 3 tests |

### Both tiers, both `DEVRC_TEST_JOBS` settings — re-run at `9a0b683b`

| tier | result |
|---|---|
| dev host, `gate.sh --tier both`, `-n 4 --dist loadfile` | **`GATE: RESULT=PASS exit=0`** — pytest `collected=16729 passed=16727 skipped=2 failed=0`; node `tests=1292 pass=1292 fail=0` |
| nix sandbox, `nix build .#checks.x86_64-linux.{pytests,nodetests}` | **both derivations built** |
| Tekton, this head | `tekton/devrc-pytests` **pass**, `tekton/devrc-nodetests` **pass** |
| dev host, `DEVRC_TEST_JOBS=1` | **`failed=0`** (`collected=16729 passed=16726 skipped=3`) |
| no-systemctl condition (the sandbox's distinguishing property) | **90 passed** on the clean tree |

The serial run still exits 1 on GUARD 2's unpinned serial-only skip (`test_gitenv_sibling_exclusion.py:92`), unchanged by this round and unchanged from `origin/main` — those files are byte-identical on this branch.

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. **Not merged.**

ZacxDev:
## Delta round — all five folded in

Commit `f01b4f03`. All five were the same class, and three of them were **my guards' own fixtures unable to see the mutant their comment described** — the trap this PR keeps closing, found one layer down each time.

| # | finding | status |
|---|---|---|
| 🟡 1 | `nolaunch.py:329` `w = _WORKER_SH` → `'"main"'` survives serially | **fixed** — killed serially now |
| 🟡 2 | the new parser only matched `grep -c '…'` / `grep -vE '…'` | **fixed** — invocation set pinned too |
| 🟢 3 | `_TAGGED`'s `\S+` → `.*` survives; fixture word at end of string | **fixed** |
| 🟢 4 | `run-tests.sh:2894-2896` false for the untagged marker | **fixed** |
| 🟢 5 | `⚠ INVARIANT GUARD` block drifted onto the helper | **fixed** |
| info | `real is None` → `unlink` unreachable in both tiers | **marked defensive in source**, not counted as exercised |

### 🟡 1 — the fixture trap, in the one branch this round exists to pin

Confirmed exactly as diagnosed. `launcher_body` already had this control (`gw7` plus a literal-text pin) and L2 had `gwABSPROBE`; `systemctl_body` had neither, and comparing against `worker_tag()` collapsed to the same literal `main` the sh stub falls back to.

All **five** stub invocations in the classifier test now run under `PYTEST_XDIST_WORKER=gwSYSPROBE` — a value `SERIAL_WORKER` cannot equal — and every tagged line is asserted against it, including `systemctl(blocked)` and the two record-only launchers, each of which has its own printf and is therefore its own mutation site. The test now measures the same thing whichever way the suite is invoked.

**M15 killed serially** (the whole battery runs with no `-n`, which is the condition it previously survived):

```
AssertionError: a `systemctl(read)` line did not carry the worker id its stub was RUN under. …
['systemctl(read) [main] --user status u', 'systemctl(read) [main] --version']
assert ['main', 'main'] == ['gwSYSPROBE', 'gwSYSPROBE']
```

### 🟡 2 — the parser was the same bug one layer down

Reproduced: `grep -cE '^brandnew\(kind\)'` in the GUARD 7 block left it green. Two things now stand in for "what the runner will run" — a quote- and flag-agnostic pattern extraction, **and** a pin on the **set of `grep` invocations** in the block. The second is the one that generalises: it sees a new classifier whatever its flags or quoting, including one with **no quoted pattern at all** — which the block's own fourth grep (`grep -c .`, counting hit lines) already is, so the pattern pin alone could never have counted the greps.

**M17 killed:**
```
AssertionError: GUARD 7 no longer runs exactly these four greps over the launch log. …
['-c', '-c', '-cE', '-vE', '-c']  !=  ['-c', '-c', '-vE', '-c']
```

Your fail-closed findings (missing banner, empty extraction) are unchanged and still assert.

### 🟢 3 — the fixture now has something after the bracketed word

Backtracking made `^.* ` identical to `^\S+ ` with the word at end of string, so the fixture could not see it. ` --urgency low` appended. **M16 killed:** `assert 'not-a-tag' == 'gw1'`.

### 🟢 4 and 🟢 5

The runner comment now says **INTERCEPT** line and names the `nolaunch(session)` marker as the one deliberate exception, whose attribution that block owns itself. The `⚠ INVARIANT GUARD` block moved back to sit directly above the test, and says "THIS TEST, not the helper above" so the next insertion cannot silently re-orphan it.

### Informational — the unreachable branch

Not counted as exercised. It is now marked in the source as defensive and unreachable in both tiers, with the reason for each (dev host has a real systemctl; the sandbox never wrote a stub for it to remove).

---

### Re-run battery — 18 mutants, all killed, control green, **run serially**

`M0-control: GREEN (30 passed)`. Own tree per mutant, `__pycache__` deleted, `PYTHONDONTWRITEBYTECODE=1`, replacement count asserted first.

Worth reporting because it is the harness's own negative control working: **M14's needle moved** when I added the "defensive/unreachable" comment above it, and the harness **aborted with "the mutant did NOT apply, so any verdict about it is meaningless"** rather than scoring a SURVIVED. Needle fixed, re-run, killed (3 tests).

New this round, each killed by exactly one test: **M15** (`w` hardcoded) and **M17** (`-cE` fourth classifier) → the classifier guard; **M16** (anchor widened) → the contract test.

I do not have your numbering, so I cannot map "M10 and M18" to mine with confidence — reporting by what each mutant *does* instead: the `w = '"main"'` hardcode, the `\S+` → `.*` widening, and the `-cE` classifier insertion are all now RED, each on its own assertion, with the messages above.

### Both tiers — re-run at `f01b4f03`

| tier | result |
|---|---|
| dev host, `gate.sh --tier both`, `-n 4 --dist loadfile` | **`GATE: RESULT=PASS exit=0`** — pytest `collected=16729 passed=16727 skipped=2 failed=0`; node `tests=1292 pass=1292 fail=0` |
| nix sandbox, `nix build .#checks.x86_64-linux.{pytests,nodetests}` | **both derivations built** |
| **Tekton, verified against the new head sha** | `f01b4f03` → `tekton/devrc-pytests` **success**, `tekton/devrc-nodetests` **success** (checked via the commit status API, not just `gh pr checks`, so these are not stale from `9a0b683b`) |
| serial | the two guard files: **90 passed** with no `-n`; and the whole 18-mutant battery runs serially |

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, head `f01b4f03`. **Not merged — yours.**

PR #869, opened 2026-08-26
feat(sync-clones): ff the stale shared clones — and name WHICH of nine outcomes, because a cron running `git merge --ff-only` reports none of them

ZacxDev:
## What this is

`scripts/sync-clones.py` + `scripts/lib/shared_clone_sync.py` — fast-forward the shared base clones, and report, **per clone and machine-readably**, why not.

The 14-day audit of 443 sessions measured these clones at **606, 238, 213, 161 and 83 commits behind** (one carrying 43 foreign worktrees), with sessions re-deriving that by hand up to **78 times in a single transcript**. The clones are write-only — all work happens in throwaway worktrees and PRs — while their working trees are what agents read `CLAUDE.md` and `.claude/skills/**` out of.

## 🔴 The constraint that shaped the design

A cron running `git merge --ff-only` **accomplishes nothing, silently**. Its four interesting outcomes — fast-forwarded 606 commits / already current / refused because the tree is dirty / refused because the branch has local commits — all end that command with nothing worth reading, and the no-op **exits 0**. So the product is the **classification**, not the merge.

| status | exit | sentinel |
|---|---|---|
| `synced` | 0 | `fast-forwarded` |
| `current` | 0 | `already current` |
| `refused-dirty` | 3 | `working tree is dirty` |
| `refused-diverged` | 4 | `local commits are not upstream` |
| `refused-no-upstream` | 5 | `no upstream to sync from` |
| `refused-detached` | 6 | `HEAD is detached` |
| `refused-not-a-repo` | 7 | `not a git repository` |
| `refused-fetch-failed` | 8 | `fetch failed` |
| `refused-ff-failed` | 9 | `ff refused by git` |

Sentinels are pinned pairwise distinct **and** non-substring, so a grep for one cannot match another's line. `synced` and `current` both exit 0 and are deliberately **not** collapsed (different status, sentinel, and `moved` count). A run that examined **zero** repositories exits **2**, never 0.

## 🔴 It never resolves dirt

No stash (`refs/stash` is repo-global here and a concurrent agent can pop yours), no reset, no discard, **and no flag to override**. Blocking paths are named (capped at 10, full count always reported) and a human decides.

## 🔴 "Dirty" is an overlap question, not a boolean

Refusing on **any** dirt was measured to make the tool inert. Read-only survey of the real fleet under `~/workspace`, taken for this PR:

- **47** primary clones
- **38 of 47 (81 %) are dirty** — `dave-db` 79 paths, `promptver` 52, `kubeclaw-cloud` 44, `giarcidia-net` 37, `hytalegarage` 32
- **13 of 47 (28 %) have no upstream** on the checked-out branch → `refused-no-upstream`
- `kubeclaw` and `naida-ai` sit on **feature** branches — exactly where guessing a mainline would be wrong
- `tryonhaulcentral` is **2 ahead** → `refused-diverged`
- **without** a fetch, `stable-diffusion.cpp` already reads 82 behind and `homelab-infra` 11 — which is why the fetch step is not optional

A strict-dirty design would have refused 81 % of the fleet forever: a permanently-red gate nobody reads. The guard is `{dirty paths} ∩ {paths the ff would write}`, with `--no-renames` on the incoming list because rename detection reports only the destination while the ff deletes the source too. Non-blocking dirt is reported as **advisory**.

## Why it does not reuse `lib/git_mainline.py`

That answers "what is this repo's **mainline**" — a different question. Using it would fast-forward a clone parked on a feature branch onto `origin/main`. The target is the checked-out branch's **own** upstream; a branch with none is a refusal, never a guess.

## Instrument validation

**Positive control** — a fixture that must report `synced`, asserting the HEAD commit **count** before and after: **1 → 4**, `moved == 3`.
**Negative control** — a fixture that must refuse dirty, asserting the specific sentinel *and* exit code *and* that the count did **not** move: **1 → 1**, WIP preserved byte-for-byte.
`TestNotConfusable` additionally pins that `refused-dirty` is not reported as `current`, and that `current`/`synced` share no status, no sentinel and no `moved` value.

**Mutation-verified: 20 mutants, 20 killed.** Positive control 49 passed / 0 FAILED / 0 ERROR; negative control (deliberately broken module) 1 collect-error. Battery in-tree at `scripts/tests/mutants-shared-clone-sync.sh`.

🔴 Two mutants **survived the first pass** — `head-did-not-move-check-gone` and `ffonly-becomes-plain-merge` — and `TestTheLastLineChecks` exists because of them. The second is honestly labelled a **structural** pin: dropping `--ff-only` is behaviourally invisible to every fixture (the `ahead > 0` guard already returned), so it earns its place only against a race no fixture can stage.

🔴 The battery's **first version scored all 16 mutants SURVIVED while every one was in fact KILLED.** It parsed pytest's summary with `s/.*[^0-9]\([0-9]\+\) failed.*/\1/p`, and `11 failed, 34 passed` starts with the digit, so `.*[^0-9]` can never match it — the companion expression then read the `4` out of `34`. That history is recorded in the file's header; the battery now counts `FAILED `/`ERROR ` lines and never parses the summary format.

## `test_git_mainline.py`: the importer ledger was a spelled guard

`test_the_LEDGER_of_importers_is_pinned_both_ways` matched `"git_mainline" in p.read_text()`, so it fired on `shared_clone_sync.py` — which only **names** the module, in a docstring explaining why it deliberately does not use it. The docstring said "importers"; the implementation counted mentions. It now matches an actual import in every spelling this repo uses (with the unmatched spellings stated rather than implied), and the **mention** set is pinned separately, so nothing got weaker — both sets fail when they grow or shrink, and a prose "we deliberately don't use it" quietly becoming a real import turns both red. Validated by mutation inside `nix develop`: green unmutated, red on importer-removed (2 fails), on a third importer planted (1), on the prose mention removed (1), green again restored.

## Gate — both tiers, on this tree

```
dev-host  scripts/gate.sh                pytest PASS exit=0, node PASS exit=0, GATE_RC=0
nix       .#checks.x86_64-linux.pytests  RESULT: PASS (exit=0), NIX_RC=0
TOTAL collected=16695 passed=16693 skipped=2 failed=0   (floor 15789)
scripts/tests collected=8244 passed=8244 skipped=0      (floor 8036, ceiling 10045)
```
Both tiers agree target-for-target on `scripts/tests`: **8244**.

**No floor change.** `scripts/tests` is an existing target, so GUARD 5 / GUARD 3a need no edit (`--check-targets` and `--check-floors` both PASS). 8244 sits inside `[8036, 10045]`, so the floor neither fails low nor has drifted far enough to stop detecting a collapse — and that line is a documented churn magnet that conflicts with every other open PR.

## What I could not verify

- **Never run against the operator's real clones.** Everything above is fixtures plus a **read-only** survey; a live no-argument run would perform real fast-forwards, which is the operator's call.
- The read-only survey used **no fetch**, so its behind-counts are floors against already-fetched refs — they are not comparable to the audit's 606.
- **Operational note:** because the run exit code is the *worst* repo's, a bare no-argument run on this fleet exits **5** today (13 clones track nothing). That is honest per-repo, but a cron wanting only actionable failures should read `--json`, not the exit code alone.
- `refused-ff-failed` and the HEAD-did-not-move check are reached through the module's own git seam (`monkeypatch`), not by persuading real git into that state — stated in those tests' docstrings.


ZacxDev:
Merging — gated on the **merged tree**, not on this branch, and exercised against **real clones** first.

**Why the branch's own green was not enough.** Both required checks were SUCCESS here, but this branch is **28 commits behind `main`** and overlaps `main` on `scripts/README.md`. `strict` is off on this repo, so a PR can merge while behind — which is exactly the hazard that lets a reviewer's green describe a tree that will never exist. So: integration branch off current `main` (`5426fe54`), merged this branch into it, no textual conflicts, and I read the merged `scripts/README.md` region by hand — the two new rows land cleanly in the script table with no interleaving damage.

**Gate on that merged tree** (`scripts/gate.sh`, both tiers):
```
pytest  collected=17038  passed=17036  skipped=2  failed=0   (floor 15789)
node    suites=5 files=39 tests=1292 pass=1292 fail=0        (floor 1239)
GATE: RESULT=PASS exit=0
```
Both skips are the pinned ones (needs-real-Postgres, opt-in live-store drift). Verdict lines and exit codes agree, so no exit-90 could-not-vouch.

**First real-clone run — the thing this PR had never had.** Three clones chosen to force three different outcomes:
```
homelab-infra: fast-forwarded — 46 commit(s) onto origin/trunk (9c6df92686d2 -> f3d4b62f76d7)
homelab-talos: already current with origin/trunk at f3d4b62f76d7
amprolium.com: no upstream to sync from — refusing to guess a target
sync-clones: 3 repo(s) — synced=1 current=1 refused-no-upstream=1 (exit=5)
```
The ff landed and was verified after the fact (`behind=0`, working tree still clean). `synced` and `current` are distinguishable by status, sentinel and `moved`. The exit code is the **worst** repo's — 5, not averaged away by two successes. A second run returned `current`/`current`, exit 0, `moved: 0` in `--json`.

**A finding that argues for this tool rather than against it.** Before the run I surveyed ~48 clones by `rev-list --count <branch>..<upstream>` **without fetching**, and only 2 looked stale. `homelab-infra` read as 11 behind; after this tool fetched, it was **46**. A no-fetch reading is a floor, exactly as the module docstring says — so the fleet is meaningfully staler than a naive check reports, and the naive check is what a human reaches for.

🔴 **What I did NOT exercise live, and am merging on unit tests alone:** `refused-dirty` (the `{dirty} ∩ {incoming}` overlap guard), `refused-diverged`, `refused-detached`, `refused-not-a-repo`, `refused-fetch-failed`, `refused-ff-failed`. Manufacturing those states in real clones was not worth the risk; they rest on `test_shared_clone_sync.py` and `mutants-shared-clone-sync.sh`, neither of which I re-ran independently of the suite above.

Decision made 2026-08-26 with Zach, alongside #864 and #866 (both closed on measured benefit).

PR #872, opened 2026-08-26
present: three nouns, not six stages — sessions · claimable tasks · subsystems

ZacxDev:
Restructures `scripts/present/` around **three primary concepts** instead of the six-stage loop, at the operator's direction.

## The new spine

| § | section | what it owns |
|---|---|---|
| 0 | Overview — three concepts | the three boxes, the three seam arrows, the constraints band |
| 1 | How to read this page | `repo.head` |
| 2 | Start here | four entry paths, re-pointed at the three nouns |
| **3** | **Sessions** | `hooks`, `telemetry.sources` |
| **4** | **Claimable tasks** | `tasks.intake` *(new)* |
| **5** | **Subsystems** | `skills.inventory`, `index.store`, `seam.store_api`, `seam.store_traffic` |
| **6** | **Where the three touch** | `seam.jsonl`, `seam.detectors`, `seam.sessions` |
| **7** | **What constrains all three** | the 8 budget/gate/ship/drift rows |
| 8 | Invariants and tripwires | — |
| 9 | The cost model of a change | — |
| 10 | Negative space | kept whole — it carries an argument |
| 11 | The evidence bar | kept whole, **plus** the test-evidence bullets moved in from the old §5 |
| 12 | Glossary | `session`, `claimable task`, `scope` added |

Gone: `told`, `may-do`, `verified`, `ships`, `drift`, `observed`, `soft`. Nothing was deleted outright — every block was re-homed, folded into a neighbour, or reduced to a line.

## Demoted, not hidden — the four systems that mint work

**clawgate Tasks is canonical**, per operator ruling: it is the only one with dispatch, agents, acceptance criteria and a status gate. The other three get one honest line each and say **why they are separate**:

* **ClickUp** — someone else's system of record; statuses set off this machine, so nothing here gates on them and no agent claims one.
* **The initiatives board** — a *thread*, not a unit: coarser than a task, no acceptance criterion.
* **`scripts/claim-work.sh`** (#847) — a *lock*, not a queue: an orphan commit pushed to a claim ref IS the claim.

Choosing a canonical one is a judgement about what a *claimable task* is. It is not a consolidation, and writing the page as though it were would be the "reads as coverage while providing none" failure the page exists to teach against.

That claim is now **measured, not asserted**. `tasks.intake` resolves each system's entry point in the tree and renders `ABSENT` for one that was renamed or retired — because *demoted* and *gone* look identical to a reader who never meets the other three. Its rendered detail states its own blind spot: the ledger is hand-authored, so a **fifth** system nobody adds to it is invisible here. There is no derivable definition of "mints work" to enumerate from, and inventing one would be a scan that walked nothing reporting all-clear.

## §0

The clickable six-stage cycle is replaced by three clickable concept boxes, three labelled seam arrows, and the constraints band drawn *under* the row rather than beside it. The long seam arrow runs underneath: **sessions reach subsystems without going through a task at all**, which is the seam that leaks.

Every existing property is kept — hand-authored inline SVG, counts fed from the measurement layer **by registry key** (never a literal), a stage with no measurement rendering `UNMEASURED` with its reason, clickable regions as primary navigation, legible narrow, no horizontal body scroll.

## Why `measure.py` was touched

Both reasons are forced, and no existing measurer's logic changed:

1. `test_measurement_sections_all_name_a_real_section_slug` pins every registry entry's `section` against a real page section — renaming the sections **requires** re-homing all 22 rows.
2. The new `tasks.intake` measurer, argued above.

## Sizes and ceilings

Authored prose **41,424 B → 43,668 B (+5.4%)**. The new §4 alone is 2,658 B, so **every other section net shrank by ~400 B**. Rendered page 114,284 B → 117,414 B.

`TOTAL_CEILING` is **unchanged at 45,000 B** and the page is under it. Per-section ceilings re-pinned to the new structure at written + ~8%; the ratchet's own control (`written/permitted > 0.80`) passes at 0.92.

Cut along the way: a duplicate memory-index `kv` (the budget diagram already says it), the tier-by-symptom note (folded to one sentence), the `readlink` note (folded into the merged-is-not-deployed paragraph), two merged git-prohibition rows, and the `rc 17` + `fuzzyclaw` glossary entries — both orphans, referenced by no prose on the page.

## Verified by reading the artefact, not only the suite

* Generated plain **and** `--sanitize`, extracted the text, read it. **Zero** literal HTML entities.
* §0 screenshotted light and dark. The first dark render looked broken until the control: Chrome's `--force-dark-mode` re-inverts an already-dark SVG. Under a genuine `prefers-color-scheme: dark` it is correct.
* Body horizontal overflow probed at **360 px** through an exact-width iframe — `chromium --window-size` clamps at ~485 px, so a bare narrow window silently measures a different viewport than the one you asked for. Result `BODYSCROLL=NO, offenders=0` on both the plain and sanitized pages, against a **negative control** (one `nowrap` pill) that reports `BODYSCROLL=YES, offenders=1`.
* The two `SANITIZE DEGRADED` warnings are **pre-existing** — reproduced by running `origin/main`'s own modules against this same tree.

## What the three-noun frame does not cover

Reported rather than forced:

* **`skills.inventory` under Subsystems slightly over-claims.** Not every skill operates a subsystem — `/handoff`, `/resume`, `/audit-pr` are workflow skills. The inventory is homed there because that is where a *reader* looks for it, not because the mapping is total.
* **The frame covers the SYSTEM, not the PRACTICE.** §9 (cost model), §10 (negative space) and §11 (the evidence bar) are about how work is decided and defended here. They are not properties of a session, a task or a subsystem, and they sit beside the three the way §7 does. That is not a defect of the frame; it is worth knowing the page has two axes, not one.

## Gate — and the sandbox tier earned its keep

| tier | verdict |
|---|---|
| dev-host `gate.sh --tier both` | **PASS** — `GATE: RESULT=PASS exit=0`; pytest `collected=16645 passed=16643 skipped=2 failed=0` (floor 15,789, 28 targets); node `suites=5 files=39 tests=1292 pass=1292 fail=0` (floor 1,239). No `panic: test timed out`, no exit 90. |
| sandbox `nix build .#checks.x86_64-linux.nodetests` | **PASS** (rc 0) |
| sandbox `nix build .#checks.x86_64-linux.pytests` | **FAILED first (rc 1)** on `assert 3 == 2` — a real defect the dev-host tier is structurally unable to see. Fixed in the second commit; re-run reported in the thread. |

The defect: `test_an_unmeasured_stage_renders_the_word_not_a_blank` asserted a **literal** `svg.count("dbox-warn") == 2`. The three-concept overview gives `index.store` a §0 stage of its own, and that row needs a home directory — it measures on the dev host and is legitimately `UNMEASURED` in the sandbox, which drew a third warned box. The fix makes the expectation a **delta** (`live_warned + 2`) and chooses its victim stages from what actually measured on the build, so it cannot be host-dependent again.

Matrix, under a repro that patches `Path.home()` to an empty directory so the dev host reproduces the sandbox's index-store absence:

```
OLD test, sandbox-like HOME   RED   "an absent stage must be visibly toned"
NEW test, sandbox-like HOME   GREEN
NEW test, real HOME           GREEN (17 passed)
```

Further per-tier detail in the thread below. ⚠ The box was under heavy concurrent load throughout — **31 `gate.sh` / `run-tests.sh` processes across seven working trees other than mine** were already active when I started (`devrc-wt-prune`, `devrc-wt-sync`, `devrc-wt-env`, `devrc-wt-hookguard`, `/tmp/wt-skill-lessons`, and two sibling agent worktrees), enumerated by resolved PID via `/proc/<pid>/cmdline` — deliberately **not** `pgrep -f`, which matches the enumerating shell itself. Nothing was killed. Wall times are inflated accordingly, so a slow run here is not evidence about this diff.

### DIFF
--- a/scripts/present/content.py
+++ b/scripts/present/content.py
@@ -9,8 +9,9 @@ one of them appears in the page's authored VISIBLE TEXT.
 
 Historical quantities are deliberately IN BOUNDS — a killed proposal's
 measurement, a dated incident's count. They are permanently true, they are not
-what the generator measures, and stripping them would gut §11 and §12, whose
-whole job is to carry the number that settled a question. What is out of bounds
+what the generator measures, and stripping them would gut §10 (negative space)
+and §11 (the evidence bar), whose whole job is to carry the number that settled
+a question. What is out of bounds
 is restating a LIVE fact: the moment prose carries a byte count or a test total,
 that paragraph starts aging, and this repo has measured its own prose false in
 both directions.
@@ -104,7 +105,7 @@ def _arrow(x1, y1, x2, y2, mid, label=""):
 
 
 # --------------------------------------------------------------------------- #
-# §0 — the overview cycle. The ONLY place the whole system appears at once.
+# §0 — the three concepts. The ONLY place the whole system appears at once.
 # --------------------------------------------------------------------------- #
 
 #: (stage label, section slug, description lines, measurement key, box tone).
@@ -114,32 +115,35 @@ def _arrow(x1, y1, x2, y2, mid, label=""):
 #: — and a key no measurer produces renders UNMEASURED rather than blank.
 #: `test_present_content.py` pins every slug against a real section and every key
 #: against the live registry, both directions.
+#:
+#: The first three are the page's three PRIMARY CONCEPTS. The fourth is not a
+#: fourth concept: it is the band of constraints all three sit inside, drawn
+#: underneath them rather than beside them for exactly that reason.
 OVERVIEW_STAGES: tuple[tuple[str, str, str, str, str], ...] = (
-    ("TOLD", "told",
-     "rules &#183; skills &#183; memory|all three CAPPED, paid every session",
-     "skills.listing", "a"),
-    ("MAY DO", "may-do",
-     "hooks that can REFUSE a call|shipped and registered differ",
-     "hooks", "a"),
-    ("VERIFIED", "verified",
-     "two gate tiers, running in|two different environments",
-     "gate.tiers", "b"),
-    ("SHIPS", "ships",
-     "a switch, per host &#8212; a merge|deploys precisely nothing",
-     "ship.managed", "b"),
-    ("DRIFT", "drift",
-     "a passive deadman on a timer|it reports, and never fixes",
-     "drift.ladder", "d"),
-    ("OBSERVED", "observed",
-     "telemetry &#183; recall store|&#183; live session surfaces",
-     "telemetry.sources", "c"),
+    ("SESSIONS", "sessions",
+     "Claude Code instances doing work|one pane, one transcript, many views",
+     "seam.sessions", "a"),
+    ("CLAIMABLE TASKS", "tasks",
+     "clawgate Tasks is canonical here|three other systems also mint work",
+     "tasks.intake", "b"),
+    ("SUBSYSTEMS", "subsystems",
+     "what is worked on, and what is|known about it &#8212; recalled on demand",
+     "index.store", "c"),
+    ("WHAT CONSTRAINS ALL THREE", "constraints",
+     "always-on context budgets &#183; two gate tiers, in two environments|"
+     "merged is not deployed &#8212; a switch, per host, or nothing moves",
+     "skills.listing", "d"),
 )
 
-#: Box geometry, laid out as a closed racetrack: three across the top going
-#: right, down the right edge, three across the bottom coming back left, up the
-#: left edge to the start. A true hexagon reads worse at this text density.
-_OV_W, _OV_H = 260, 104
-_OV_POS = ((20, 30), (330, 30), (640, 30), (640, 250), (330, 250), (20, 250))
+#: Box geometry, parallel to `OVERVIEW_STAGES`: three concept boxes across the
+#: top, and the constraints band spanning the full width beneath them. The band
+#: is WIDE and LOW on purpose — it reads as ground, not as a peer.
+_OV_GEOM = (
+    (16, 30, 272, 112),
+    (324, 30, 272, 112),
+    (632, 30, 272, 112),
+    (16, 214, 888, 92),
+)
 
 
 def _wrap(text: str, width: int, maxlines: int = 2) -> list[str]:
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Both tiers, both green — at the same base sha (`a6cc23eb`)

Reported as verdict lines and per-target counts, never an exit code.

| tier | command | verdict |
|---|---|---|
| dev-host | `nix develop --impure --command bash scripts/gate.sh --tier both` | `GATE: RESULT=PASS exit=0` |
| | — pytest | `TOTAL collected=16645 passed=16643 skipped=2 failed=0` (floor 15,789 = sum of 28 per-target floors) |
| | — node | `TOTAL suites=5 files=39 tests=1292 pass=1292 fail=0 skipped=0` (floor 1,239) |
| sandbox | `nix build .#checks.x86_64-linux.pytests` | `RESULT: PASS (exit=0)` — `collected=16645 passed=16643 skipped=2 failed=0` |
| sandbox | `nix build .#checks.x86_64-linux.nodetests` | `RESULT: PASS (exit=0)` — `suites=5 files=39 tests=1292 pass=1292 fail=0` |

No `panic: test timed out`; no `exit=90` (could-not-vouch). No floor drifted, so no floor number was changed.

### The sandbox tier earned its keep on this PR

At the first commit (`6122fc0d`) the dev-host tier was **PASS** and sandbox `pytests` was **FAIL (exit=1)** on `assert 3 == 2` in `test_an_unmeasured_stage_renders_the_word_not_a_blank`. Not a flake and not an environment problem — a real defect that the dev-host tier is *structurally unable* to see: the new §0 gives `index.store` a stage, that row needs a home directory, and the sandbox has none, so a third box is legitimately toned `UNMEASURED` there. The test's hardcoded `== 2` was a claim about the author's machine.

Fixed in `a6cc23eb` by making the expectation a **delta** and choosing the victim stages from what actually measured on the build. Matrix, using a repro that patches `Path.home()` to an empty directory so the dev host reproduces the sandbox's absence:

```
OLD test, sandbox-like HOME   RED    "an absent stage must be visibly toned"
NEW test, sandbox-like HOME   GREEN
NEW test, real HOME           GREEN  (17 passed)
```

### Scope of this claim

Four runs, two tiers, **one base sha** (`a6cc23eb`), tree clean at run time. Gating the tree the *merge* creates is still a separate act — `strict` is false on this repo's branch protection, so a green check is a claim about this branch.

⚠ Load caveat, stated because it bounds what a wall time here would mean: 31 `gate.sh` / `run-tests.sh` processes across seven other working trees were active during these runs (load average ~40), enumerated by resolved PID via `/proc/<pid>/cmdline` rather than `pgrep -f`. Nothing was killed.

PR #880, opened 2026-08-26
test(skill-size): plant positive controls for 8 guards nobody had watched fire

ZacxDev:
## What

`scripts/dead-guard-scan.py` scored **8 guard FIRING paths** across the three
skill-size gates as never executed by any test. Each carried the in-place
annotation *"no planted positive control drives it. Adding one is the real
remedy"*. This plants the controls.

| # | Site | Guard |
|---|---|---|
| S1 | `test_prune_skill_size.py` | `checks` loop, arm 1 — pattern matched **zero** times |
| S2 | `test_prune_skill_size.py` | `checks` loop, arm 2 — pattern matched **more than once** |
| S3 | `test_prune_skill_size.py` | `checks` loop, arm 3 — captured value **≠** measured |
| S4 | `test_prune_skill_size.py` | `BB_CLAIM` sentence pin (`count(...) != 1`) |
| S5 | `test_prune_skill_size.py` | `bb > target` — the sentence is present and **FALSE** |
| S6 | `test_handoff_skill_size.py` | `_existing_topics` — `reference/` glob is empty |
| S7 | `test_handoff_skill_size.py` | `_pin_counts` — neither pin module could be parsed |
| S8 | `test_session_manager_skill_size.py` | self-route loop **recorded** |

## How

All eight sat inside long test bodies that walk the LIVE repo, so the only
input that could reach them was a dirty corpus. Each is extracted into a pure
module-level helper; the live gate calls the helper and keeps its assertion text
unchanged. **Pure refactor** — no assertion weakened, relaxed or deleted, every
message string carried over byte-identical, no `SKILL.md` and no prose figure
touched.

New helpers:

- `test_prune_skill_size.figure_problems_in(src, checks)`
- `test_prune_skill_size.bb_claim_problems_in(src, bb, target, bb_claim)`
- `test_handoff_skill_size.existing_topics_text(reference_dir)`
- `test_handoff_skill_size.pin_counts_text(counts)`
- `test_session_manager_skill_size.self_routing_loops_in(topics, resolve)`

Controls are driven **in-process** (direct function call, never `subprocess`) —
the scanner traces with `sys.settrace`, which is per-interpreter, so a
subprocess-driven control would leave the branch recorded as never-executed.

S1–S3 are mutually exclusive branches of ONE loop, so each fixture is built to
leave the other two **silent**: the multi-match fixture sets `expected` to its
own first capture, so neutering that arm falls through to a value comparison
that *passes* rather than to a neighbouring message that could score the mutant
dead for the wrong reason. Fixture figures are pairwise distinct and are not
restated from any live measurement.

## Mutation matrix

Narrowest expression neutered, only that control run, `PYTHONDONTWRITEBYTECODE=1`,
file restored byte-exactly (sha256-verified) between mutants. Every one observed
**RED naming its own arm's message**:

| # | Mutation | Observed failure |
|---|---|---|
| S1 | arm-1 `problems.append(...)` → `pass` | `assert [] == [...]` — missing `absent figure: PATTERN DID NOT MATCH (...) — the sentence was reworded...` |
| S2 | `len(found) > 1` → `len(found) > 99` | missing `doubled figure: pattern matched 2x (['2,207', '6,619']) — an added sentence can shadow...` |
| S3 | `found[0] != expected` → `False` | missing `stale figure: prose says 5,003, measured 8,761` |
| S4 | `" ".join(src.split()).count(bb_claim) != 1` → `False` | missing `browser-bridge claim: ... is not present exactly once as written` (both params, 0 and 2 occurrences) |
| S5 | `bb > target` → `False` | missing `browser-bridge claim: prose says browser-bridge MEETS the 11,453 B target, but it measures 13,701 B -- over by 2,248 B` |
| S6 | `if not topics:` → `if False:` | `assert '' == '(none found under .../reference -- did reference/ move?)'` |
| S7 | `if not counts:` → `if False:` | got the ⚠ pin-count block, expected `The skill's prose is pinned verbatim by HANDOFF_SENTENCES ... neither could be parsed` |
| S8 | `if hits:` → `if False:` | `assert {} == {'loops-onto-itself.md': ['reference/loops-onto-itself.md']}` |

S1 is neutered at its `append` rather than at `if not found:` deliberately:
flipping that condition drops an EMPTY capture list into `found[0]`, so the
control would die of an `IndexError` instead of the absence of its own message —
a mutant dead for the wrong reason.

The prune mutants were re-run a second time against the FINAL file bytes (a
comment-only edit had raced the first pass); all five RED again, same messages.

Negative controls accompany S6/S7/S8 so a mutant that fires unconditionally
cannot pass, plus a clean-input control asserting both prune helpers invent
nothing on a healthy corpus.

## Verification

`scripts/dead-guard-scan.py`, same interpreter, base vs branch:

```
base   (9b5dc48d) : flagged 55 branch bodies never executed (53 justified, 2 unresolved)
branch (this PR)  : flagged 47 branch bodies never executed (45 justified, 2 unresolved)
```

Exactly the eight, and **no new flag**. The 2 unresolved on both sides are the
pre-existing `test_no_real_launchers.py:549 / :679` pair, untouched.

Targeted suite, green:

```
PYTHONDONTWRITEBYTECODE=1 pytest scripts/tests/test_prune_skill_size.py \
  scripts/tests/test_handoff_skill_size.py \
  scripts/tests/test_session_manager_skill_size.py -q
86 passed
```

Full `scripts/tests` is environmentally RED on this box (the default `python3`
resolves to another repo's `.venv`, and `nix develop`'s `gateTools` are absent);
base and branch were run identically and are reported as a pair in the PR
comments. Tekton CI is the authority.

### DIFF
--- a/scripts/tests/test_handoff_skill_size.py
+++ b/scripts/tests/test_handoff_skill_size.py
@@ -206,17 +206,25 @@ MAX_BYTES = 24_000
 MIN_HEADROOM_BYTES = 900
 
 
-def _existing_topics() -> str:
+def existing_topics_text(reference_dir: Path) -> str:
     """The reference topics that exist RIGHT NOW, read off the filesystem.
 
     Globbed rather than hard-coded for the reason test_skill_size.py gives: a
     hand-maintained list drifts, and a drifted list steers a maintainer into
     creating a duplicate topic for content that already has a home.
+
+    Takes the directory as an argument so the empty-glob branch -- the one that
+    tells a maintainer `reference/` moved -- can be driven by a planted control
+    against a tmp dir instead of waiting for the real tree to lose its sidecars.
     """
-    topics = sorted(p.stem for p in REFERENCE_DIR.glob("*.md"))
+    topics = sorted(p.stem for p in reference_dir.glob("*.md"))
     if not topics:
-        return f"(none found under {REFERENCE_DIR} -- did reference/ move?)"  # pragma: no cover - the guard's own FIRING path: it fires only when the corpus is dirty, and no planted positive control drives it. Adding one is the real remedy
+        return f"(none found under {reference_dir} -- did reference/ move?)"
     return ", ".join(topics)
+
+
+def _existing_topics() -> str:
+    return existing_topics_text(REFERENCE_DIR)
 
 
 def _pin_lists() -> dict[str, int]:
@@ -259,10 +267,16 @@ def _pin_lists() -> dict[str, int]:
     return counts
 
 
-def _pin_counts() -> str:
-    counts = _pin_lists()
+def pin_counts_text(counts: dict[str, int]) -> str:
+    """Render the pin-count block from an already-parsed count mapping.
+
+    Split from `_pin_lists()` so the unparseable-corpus branch below is
+    reachable from a planted control: it fires only when NEITHER pin module can
+    be parsed, which never happens on a healthy tree, so it was recorded as a
+    guard nobody has watched execute.
+    """
     if not counts:
-        return (  # pragma: no cover - the guard's own FIRING path: it fires only when the corpus is dirty, and no planted positive control drives it. Adding one is the real remedy
+        return (
             "The skill's prose is pinned verbatim by HANDOFF_SENTENCES "
             "(scripts/tests/test_subsystem_touch.py) and SKILL_PINS "
             "(scripts/tests/test_handoff_doc.py) -- neither could be parsed "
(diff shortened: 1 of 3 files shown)


ZacxDev:
Base-vs-branch full `scripts/tests` on this box (env-RED by default — the interpreter resolves to another repo's .venv and nix `gateTools` are absent):

```
base   9b5dc48d : 71 failed, 8398 passed, 1 skipped in 768.30s
branch c5ef30e4 : 71 failed, 8410 passed, 1 skipped in 777.37s
```

The FAILED sets are byte-identical (`diff` of the sorted `FAILED` lines = empty, 71 lines each), so nothing here is a regression. +12 passing = the 12 new control tests (8 prune incl. two params of the BB wording pin, 2 handoff, 2 session-manager).

`bash scripts/run-tests.sh --check-floors` → `RESULT: PASS (exit=0)`, GLOBAL floor 15789.

PR #887, opened 2026-08-26
fix(nix/git): SSH keepalives — a pre-push gate longer than github's ~360s idle close kills the push with SIGPIPE (#782)

### LINKED ISSUE #782: pre-push: a long test run drops the SSH connection — push exits 141 (SIGPIPE) and the branch is never created
## Symptom

Two consecutive pushes of the same commit failed, both **after** `githooks/tests-on-push.sh` ran the full suite:

```
pre-push: preparing test env (nix develop)…
pre-push: running devrc test suite (mode=on)…
devrc: gate toolchain ready — bash scripts/run-tests.sh .
run-tests: interpreter /nix/store/x6ihjfa2…-python3-3.12.14-env/bin/python
Connection to github.com closed by remote host.
pre-push: ✅ devrc test suite passed.
```

`git push` returned **141** (SIGPIPE) and the branch was **not** created on the remote. `--no-verify` then pushed the identical commit **instantly, rc=0**.

## Why the ordering matters

Note where the connection dies: `Connection to github.com closed by remote host.` appears **before** `pre-push: ✅ … passed`, i.e. **while the hook was still running**.

That fits git's push sequence — connect to the remote and negotiate refs, *then* run `pre-push`, *then* transfer. The suite takes minutes, the already-open SSH connection sits idle throughout, and the server closes it before the transfer begins.

## Why this is worth fixing rather than shrugging at

- It presents as a **flaky network error**, not a hook problem. Nothing in the output points at the hook.
- The hook prints **✅ passed** on the same screen, so it reads as success.
- 🔴 **A wrapper swallows the real status.** My `git push … ; echo; tail` reported **exit 0** while the push's own rc was **141**. Twice. Anything scripting a push around this hook can record a phantom success — verify with `git ls-remote`, never the wrapper's exit code.
- The longer the suite gets, the more often it will fire.

## What I could NOT establish — stated plainly

**I cannot currently reproduce it, and I am not claiming it is reproducible.**

- `githooks/install.sh` sets `core.hooksPath` **globally**, and it currently reads **empty** on this host — so the hook is not active right now, even though it demonstrably ran during the failures above. Several sessions were touching devrc hooks the same day, so it was plausibly toggled in between.
- My attempted reproduction was uninformative for two independent reasons: I used an **empty commit** (no content change) *and* the hook was not installed. It pushed in 2s, rc=0. That is **not** a refutation — the long-hook path was never exercised.

So: symptom observed twice with the output above, mechanism supported by the ordering, **reproduction outstanding**.

## Suggested next step for whoever picks this up

Reproduce with the hook installed (`githooks/install.sh`) and a commit that actually changes files, then confirm the timing hypothesis. If it holds, options roughly in order of how much they cost:

1. Run the suite **before** git opens the connection (a `pre-commit`-time or wrapper-level gate rather than `pre-push`).
2. Keep the connection alive across the hook — SSH `ServerAliveInterval` / `ClientAliveInterval` in the client config used for github.com.
3. Have the hook retry the transfer once on a SIGPIPE/closed-connection, since the suite has already passed by then.

Found while working on #761.

ZacxDev:
Closes #782. Reproduction, method and stated gaps: [issue comment](https://github.com/innovation-upstream/devrc/issues/782#issuecomment-5431370232). **An audit changed where this fix lives — see [the correction comment](https://github.com/innovation-upstream/devrc/pull/887#issuecomment-5431901660) for what moved and why.**

## What was wrong

`githooks/tests-on-push.sh` occupies a push for as long as the suite takes, and git sends nothing over the connection while it runs.

**Measured 2026-08-26 against real github.com, twice independently:** github.com closes an idle `git-receive-pack` session after **~360 s** (361 s both runs; the clean one gave `rc=255` + `Connection to github.com closed by remote host.`). git opens *and negotiates* the connection **before** running `pre-push` — measured with a `GIT_SSH_COMMAND` stamp, not inferred from interleaved output: `ssh-launch 04:12:04Z` then `hook START 04:12:05Z`.

Paired push arms, identical 420 s hook, **one variable**:

| keepalive | `push_rc` | branch on remote |
|---|---|---|
| none | **141** (SIGPIPE) | **ABSENT** |
| `ServerAliveInterval=30` | **0** | CREATED |

An idle session with the keepalive was still alive at **1367 s** (3.8×).

**Not flaky — a hard threshold**, and it fires more often the longer the suite grows. It reads as a network error because the hook prints `✅ devrc test suite passed.` *after* the connection has already died, and a wrapper's trailing command swallows the 141. **Verify a push with `git ls-remote`, never the wrapper's exit code.**

## The fix

`core.sshCommand = "ssh -o ServerAliveInterval=30 -o ServerAliveCountMax=6"` in `nix/programs/git/default.nix`.

**Not in `githooks/install.sh`, where this started.** On a home-manager host `git config --global` resolves to `~/.config/git/config`, a symlink into the **read-only nix store** — the file this module generates. The installer dies `could not lock config file … Read-only file system` (rc=255) on its *pre-existing* `core.hooksPath` line. A fix there is inert. (That is also why `core.hooksPath` reads empty on this host, which #782 had attributed to another session toggling it.)

`ServerAliveCountMax` is **explicit**, not inherited: the default is 3, so setting only the interval silently creates a `30×3 = 90 s` disconnect trigger on every git-over-ssh operation. 6 gives 180 s. ⚠ **This is the fix's own failure mode** — a >180 s stall will now kill a push that previously hung on and recovered. A trade, not a free win.

`GIT_SSH_COMMAND` (env) beats `core.sshCommand`, so `claim-work.sh` and `resume-state.sh` also carry it, with a **ledger test** over every such export that fails when the set *grows* as well as when a member is edited.

## Verification

**Verified without deploying:** `nix eval` of `homeConfigurations.zach.config.programs.git.settings.core.sshCommand` returns the value, and the generated gitconfig derivation hash moves `nxa6x4jl…` → `1yxm38gq…`. **`home-manager switch` has NOT been run** — that is a host-wide deploy. "Merged will be correct" and "live" remain separate claims until someone switches.

**Red-at-base** (origin/main's module swapped in): **3 FAILED / 1 PASSED**. Green at HEAD: **5 passed**. The single base-pass is `test_the_setting_is_reachable_from_the_home_manager_entry_point`, an **invariant guard** labelled as one — and it is the guard that would have caught this fix's own first misplacement, since a setting in a file nothing imports ships nothing.

The ledger was validated both ways: green → red naming the exact offending site → green on restore, mutation **applied-verified** (exactly one occurrence asserted before substituting).

The test parses nix **source** rather than evaluating it — a deliberate trade, stated in the file: evaluating needs `nix`, absent inside the `checks.pytests` sandbox, so the test would SKIP in the hermetic tier and be blind there. The extracted value is asserted **structurally** (parsed ints, relationships), never by substring.

Rebased onto current `origin/main` (it moved 5 commits during this work) so the suite graded the merged tree.

## Not established

- I have **not** measured that the dev-host suite exceeds 361 s on this host — the gap between "mechanism reproduced" and "this is precisely what happened on those two pushes".
- 361 s is a GitHub-side value measured twice on one day, not a documented constant. Pinned against with margin, not tuned to.
- The keepalive is validated to 1367 s, not indefinitely.

## Separately, not fixed here

**`githooks/install.sh` cannot install the hooks at all on a home-manager host** (rc=255 on `core.hooksPath`), so the pre-push gate this issue is *about* cannot currently be armed on this machine. Pre-existing and independent; deserves its own issue rather than widening this PR.

### DIFF
--- a/githooks/README.md
+++ b/githooks/README.md
@@ -15,7 +15,7 @@ Version-controlled, global git hooks. Two features, in order on every push:
 | `pre-push` | Global dispatcher. Chains to any repo-local pre-push first (never clobbers it), runs the **blocking test gate**, then fires the audit **backgrounded** so the push is never delayed. |
 | `tests-on-push.sh` | SYNCHRONOUS worker: self-detects devrc, filters on changed files, runs `scripts/run-tests.sh --set all` in the repo's **own devShell** (`nix develop`, so a venv owned by the caller's cwd cannot shadow the interpreter), and (mode `on`) **blocks on a test failure (exit 1) or a repo-content guard (exit 2)**. An ENVIRONMENT precondition (exit 3) → warn + allow. No-op for non-devrc repos. |
 | `audit-on-push.sh` | The backgrounded worker: fixture-tree + branch + synthetic-ref + diff-size + flag gates, then headless `claude -p "/audit-pr current"`, then routes 🔴/🟡 to clawgate. |
-| `install.sh` | Sets `git config --global core.hooksPath` to this dir. `--uninstall` reverts. |
+| `install.sh` | Sets `git config --global core.hooksPath` to this dir, and `core.sshCommand` to an `ServerAliveInterval=30` keepalive (see **Why the keepalive**). `--uninstall` reverts both — but only removes an `sshCommand` it wrote itself. |
 | `audit-on-push.env.example` | Config template → copy to `~/.claude/audit-on-push.env`. |
 
 ## Test gate (`tests-on-push.sh`)
@@ -95,7 +95,35 @@ Behaviour, all failing in the **safe direction**:
 ~/workspace/devrc/githooks/install.sh
 ```
 
-This sets the **global** `core.hooksPath` and seeds `~/.claude/audit-on-push.env`.
+This sets the **global** `core.hooksPath`, sets a **global `core.sshCommand`**
+carrying an SSH keepalive (see below), and seeds `~/.claude/audit-on-push.env`.
+
+### Why the keepalive (#782)
+
+**Measured 2026-08-26, twice.** github.com closes an **idle `git-receive-pack`
+session after ~360 s** (361 s in both runs). git opens *and negotiates* the
+connection **before** it runs `pre-push`, so the connection sits idle for the
+whole time the test gate is running. Once the suite passes ~360 s the connection
+is already gone when git tries to send the pack, and the push dies with
+**SIGPIPE (exit 141), creating no branch**:
+
+| pre-push duration | keepalive | push rc | branch on remote |
+|---|---|---|---|
+| 420 s | none | **141** | **absent** |
+| 420 s | `ServerAliveInterval=30` | **0** | created |
+
+🔴 **It reads as a network flake, and that is the trap.** The hook prints its own
+`✅ devrc test suite passed.` *after* the connection has already died, so the
+screen says success — and a wrapper's trailing command swallows the 141. **Verify
+a push with `git ls-remote`, never with the wrapper's exit code.**
+
+This is **not** flaky: it is a hard threshold, and it fires more often the longer
+the suite grows.
+
+The keepalive is installed here rather than in `~/.ssh/config` because that file
+is a per-host, unmanaged plain file — a fix placed there protects one machine and
+never ships. If you already have a `core.sshCommand`, the installer **leaves it
+alone** and prints the exact option to add.
 Two independent knobs, seeded from the example:
 
 - **Audit** (`AUDIT_ON_PUSH=shadow`) — logs what it *would* send, sends nothing;
--- a/githooks/README.md
+++ b/githooks/README.md
@@ -15,7 +15,7 @@ Version-controlled, global git hooks. Two features, in order on every push:
 | `pre-push` | Global dispatcher. Chains to any repo-local pre-push first (never clobbers it), runs the **blocking test gate**, then fires the audit **backgrounded** so the push is never delayed. |
 | `tests-on-push.sh` | SYNCHRONOUS worker: self-detects devrc, filters on changed files, runs `scripts/run-tests.sh --set all` in the repo's **own devShell** (`nix develop`, so a venv owned by the caller's cwd cannot shadow the interpreter), and (mode `on`) **blocks on a test failure (exit 1) or a repo-content guard (exit 2)**. An ENVIRONMENT precondition (exit 3) → warn + allow. No-op for non-devrc repos. |
 | `audit-on-push.sh` | The backgrounded worker: fixture-tree + branch + synthetic-ref + diff-size + flag gates, then headless `claude -p "/audit-pr current"`, then routes 🔴/🟡 to clawgate. |
-| `install.sh` | Sets `git config --global core.hooksPath` to this dir, and `core.sshCommand` to an `ServerAliveInterval=30` keepalive (see **Why the keepalive**). `--uninstall` reverts both — but only removes an `sshCommand` it wrote itself. |
+| `install.sh` | Sets `git config --global core.hooksPath` to this dir. `--uninstall` reverts. |
 | `audit-on-push.env.example` | Config template → copy to `~/.claude/audit-on-push.env`. |
 
 ## Test gate (`tests-on-push.sh`)
@@ -95,35 +95,7 @@ Behaviour, all failing in the **safe direction**:
 ~/workspace/devrc/githooks/install.sh
 ```
 
-This sets the **global** `core.hooksPath`, sets a **global `core.sshCommand`**
-carrying an SSH keepalive (see below), and seeds `~/.claude/audit-on-push.env`.
-
-### Why the keepalive (#782)
-
-**Measured 2026-08-26, twice.** github.com closes an **idle `git-receive-pack`
-session after ~360 s** (361 s in both runs). git opens *and negotiates* the
-connection **before** it runs `pre-push`, so the connection sits idle for the
-whole time the test gate is running. Once the suite passes ~360 s the connection
-is already gone when git tries to send the pack, and the push dies with
-**SIGPIPE (exit 141), creating no branch**:
-
-| pre-push duration | keepalive | push rc | branch on remote |
-|---|---|---|---|
-| 420 s | none | **141** | **absent** |
-| 420 s | `ServerAliveInterval=30` | **0** | created |
-
-🔴 **It reads as a network flake, and that is the trap.** The hook prints its own
-`✅ devrc test suite passed.` *after* the connection has already died, so the
-screen says success — and a wrapper's trailing command swallows the 141. **Verify
-a push with `git ls-remote`, never with the wrapper's exit code.**
-
-This is **not** flaky: it is a hard threshold, and it fires more often the longer
-the suite grows.
-
-The keepalive is installed here rather than in `~/.ssh/config` because that file
-is a per-host, unmanaged plain file — a fix placed there protects one machine and
-never ships. If you already have a `core.sshCommand`, the installer **leaves it
-alone** and prints the exact option to add.
+This sets the **global** `core.hooksPath` and seeds `~/.claude/audit-on-push.env`.
 Two independent knobs, seeded from the example:
 
 - **Audit** (`AUDIT_ON_PUSH=shadow`) — logs what it *would* send, sends nothing;
--- a/githooks/install.sh
+++ b/githooks/install.sh
@@ -17,6 +17,40 @@ set -euo pipefail
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
+# --- #782: the test gate occupies a push, so the push needs SSH keepalives ----
+# 🔴 MEASURED 2026-08-26, twice independently. github.com closes an IDLE
+# `git-receive-pack` session after ~360 s (361 s in both runs; the second also
+# gave `rc=255` and `Connection to github.com closed by remote host.`). git
+# opens AND negotiates the connection BEFORE it runs `pre-push` — measured, not
+# inferred: `ssh-launch 04:12:04Z` then `hook START 04:12:05Z` — so the
+# connection sits idle for the hook's whole runtime.
+#
+# `tests-on-push.sh` is exactly such a hook. Once the suite passes ~360 s the
+# connection is already gone when git tries to send the pack, and the push dies
+# with SIGPIPE:
+#
+#     hook sleeps 420 s, no keepalive -> push rc=141, branch ABSENT
+#     hook sleeps 420 s, keepalive=30 -> push rc=0,   branch CREATED
+#
+# Same hook duration, one variable. This is NOT flaky — it is a hard threshold,
+# and it gets hit more often the longer the suite grows.
+#
+# 🔴 IT PRESENTS AS A NETWORK FLAKE, WHICH IS WHY IT WENT UNDIAGNOSED. The hook
+# prints its own `✅ devrc test suite passed.` AFTER the connection has already
+# died, so the screen reads as success; only `git ls-remote` tells the truth.
+#
+# WHY HERE AND NOT IN ~/.ssh/config: that file is a per-host, unmanaged plain
+# file (not a home-manager symlink), so a fix placed there protects one machine
+# and never ships. This installer already writes global git config one line
+# below, and the population that runs it is exactly the population exposed to
+# the bug — the hook that creates the hazard installs the mitigation.
+#
+# 🔴 THE INTERVAL MUST STAY WELL UNDER THE ~360 s CLOSE. A keepalive longer than
+# the server's idle timeout is not a weaker fix, it is NO fix — pinned by
+# scripts/tests/test_push_keepalive.py.
+KEEPALIVE_OPT="ServerAliveInterval=30"
+KEEPALIVE_SSH="ssh -o $KEEPALIVE_OPT"
+
 if [ "${1:-}" = "--uninstall" ]; then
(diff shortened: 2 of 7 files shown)


ZacxDev:
## 🔴 Correction — the original fix was in the wrong place and would have been INERT

An adversarial audit caught this before merge. **The PR body above described the first approach; it is no longer what this PR does.** Flagging it here rather than only editing the body, since anyone who has already read it read the wrong version.

### What was wrong

The first commit put `core.sshCommand` in `githooks/install.sh`, arguing that the installer already writes global git config one line above, so that is where the mitigation belongs.

**That argument is false on the machines devrc targets.** On a home-manager host `git config --global` resolves to `~/.config/git/config`, which is a **symlink into the read-only nix store** — the very file `nix/programs/git/default.nix` generates. `~/.gitconfig` does not exist. Measured:

```
error: could not lock config file .../.config/git/config: Read-only file system
rc=255
```

…and it dies there on the **pre-existing** `core.hooksPath` line, before ever reaching anything I added. A fix placed in the installer never lands.

It also retro-explains something #782 and I both noticed and waved away: `core.hooksPath` reads **empty** on this host. We both attributed that to another session toggling it. The real reason is that the installer has never been able to write there at all.

### What it does now

`core.sshCommand = "ssh -o ServerAliveInterval=30 -o ServerAliveCountMax=6"` in `nix/programs/git/default.nix` — the module that generates that read-only file, and already the repo's declarative home for git config (`url.insteadOf`, `pull.rebase`, `rebase.autoStash`).

Verified **without deploying**: `nix eval` of `homeConfigurations.zach.config.programs.git.settings.core.sshCommand` returns the value, and the generated gitconfig derivation hash moves `nxa6x4jl…` → `1yxm38gq…`. **I have not run `home-manager switch`** — that is a host-wide deploy and not mine to make. So this is "merged will be correct", not "live"; those remain separate claims until someone switches.

`githooks/install.sh` and `githooks/README.md` are reverted to `origin/main` — this PR no longer touches them.

### Two other audit findings, both fixed

- **`ServerAliveCountMax` is now explicit (6), not inherited.** It defaults to **3**, so declaring only the interval would have silently introduced a `30×3 = 90 s` disconnect trigger on *every* git-over-ssh operation on the host, where previously only `TCPKeepAlive` applied and a stall of any length survived. 6 gives 180 s. **This is the fix's own failure mode** — on a flaky link (Mullvad is active on workbench) a >180 s stall will now kill a push that previously hung on and recovered. It is a trade, not a free win, and it is now stated in the module and pinned by a test.
- **`GIT_SSH_COMMAND` (env) beats `core.sshCommand` (config)**, so any script exporting it bypasses this entirely. devrc has two (`claim-work.sh:276`, `resume-state.sh:571`); both now carry the keepalive, and `test_every_GIT_SSH_COMMAND_export_carries_the_keepalive` is a **ledger** over all of them — it fails when the set *grows* as well as when a member is edited. Neither holds a connection near 360 s today, but claim-by-push is shipped and the next long remote op under one would resurrect #782 with nothing able to see it.

### Coverage after the rewrite

Red-at-base (origin/main's module swapped in): **3 FAILED / 1 PASSED**. Green at HEAD: **5 passed**. The one base-pass is `test_the_setting_is_reachable_from_the_home_manager_entry_point`, an **invariant guard** labelled as one — and it is precisely the guard that would have caught this misplacement, because a setting in a file nothing imports ships nothing.

The ledger was validated in both directions: green → red naming the exact offending site when one export's option is removed → green on restore, with the mutation **applied-verified** (exactly one occurrence asserted before substituting).

The test parses nix **source** rather than evaluating it. That is a deliberate trade, stated in the file: evaluating means shelling out to `nix`, which is absent inside the `checks.pytests` sandbox, so the test would SKIP in the hermetic tier and be structurally blind there. The extracted value is then asserted on **structurally** (parsed integers, relationships), never by substring.

### Separately, not fixed here

**`githooks/install.sh` cannot install the hooks at all on a home-manager host** — it dies rc=255 on `core.hooksPath`. That is pre-existing and independent of #782, and it means the pre-push test gate this issue is *about* cannot currently be armed on this machine. Worth its own issue; I did not widen this PR to cover it.

Rebased onto current `origin/main` (it moved 5 commits during this work) so the suite graded the merged tree rather than a stale branch.

ZacxDev:
## Correction to my own evidence, plus round 4's findings

### 🔴 I overstated a measurement, and it went into a public comment

In the round-3 write-up I said a `/* … */` block comment "walked **all three** links of the reachability chain". **Round 4 reproduced only one of the three.** The other two — whole-line `/* */` around `programs = programs;` and around `git = git;` — were already **red** at `4cf07a43`, because the `^\s*` anchor rejects a line starting with `/*`.

The consolidation was still the right fix, and the one reproducible link was real. But the evidence I attached was **wider than what I measured**, which is exactly the thing these audits keep catching in the code. Correcting it here because the wrong version is what a reader of that comment believed.

### 🔴 Round 4 found a regression I introduced in round 3

Nix's `''${` is an **escape**, not a string terminator. Verified against nix itself before touching anything — `builtins.stringLength ''a="''${x}"''` returns **9**, counting the `${` as literal. `nix/home.nix` contains four of them (lines 615, 648, 801, 821), and treating one as a closer **desyncs the scanner for the rest of the file**: every later real closer is then read as an opener.

That produced both failure classes the file exists to prevent:

| | before | after |
|---|---|---|
| relocating `programs = programs;` to the end of `home.nix` — a legal, semantics-identical refactor | **false RED** | SURVIVE ✓ |
| the same pin satisfied by text spelled inside a `''…''` block | **false GREEN** | KILLED ✓ |

The false RED is the nastier one: the guard's verdict depended on **where in the file** an assignment sat, and the red would have named the wrong cause.

### 🟡 Two more, same class

- **`\bsettings\s*=\s*\{` took the first match**, and `\b` treats `.` as a boundary — so an earlier `lfs.settings` attrset hijacked the locator and a keepalive hidden there passed. This is the *identical* `\b` trap the same file had already fixed for `import ./git`: one rule, wrong at the site nobody updated. The locator now runs on the string-blanked text while extraction reads the string-keeping one; both strip comments identically, so offsets agree.
- **The `.md` exclusion's justification was false.** It claimed `githooks/README.md` "shows the bad form as a counter-example". It does not — that README change was reverted in `c558eb0d`. A census finds the string in exactly two tracked places, both already in the ledger. The exclusion is kept as pre-emptive and now says so, with its residual named (a shell *comment* mentioning the assignment would still trip it).

### Verification

Battery run with an unmutated **and** a known-killable control green either side. One mutant reported `NOT APPLIED — harness fault, not a result` and was re-run rather than scored as a survivor:

```
relocate programs=programs to EOF of home.nix   SURVIVE ✓  (was false RED)
programs=programs spelled inside a '' block     KILLED     (was false GREEN)
keepalive hidden under an earlier lfs.settings  KILLED
/* */ around the setting                        KILLED
notcore.sshCommand                              KILLED
keepalive nested under includes                 KILLED
keepalive deleted, left as a # TODO             KILLED
import ./git-old                                KILLED
import spelled inside a string                  KILLED
```

Full suite: `TOTAL collected=17060 passed=17058 skipped=2 failed=0`, `RESULT: PASS (exit=0)`, with the four touched files **sha256-verified unchanged across the run** — three earlier runs in this session were invalidated by edits landing mid-run, so the check is now part of the procedure.

**The shipped fix itself is untouched by rounds 3 and 4** — `nix/programs/git/default.nix`, `claim-work.sh`, `resume-state.sh` have not changed since `c558eb0d`. Every defect those rounds found was in the test file.

ZacxDev:
## Round 5, plus a correction to my previous comment

### Correcting myself again

My last comment said the three non-test files "have not changed since `c558eb0d`". **That is wrong for `scripts/claim-work.sh`** — it gained a 4-line comment at `4cf07a43` documenting the `${GIT_SSH_COMMAND:-…}` blind spot. Comment-only, no behaviour change, but the claim as written was false and it was load-bearing for a reviewer deciding what still needed reading. `nix/programs/git/default.nix` and `scripts/resume-state.sh` are genuinely unchanged.

### My earlier correction was confirmed, and sharpened

Round 5 independently re-ran the `4cf07a43` matrix and confirmed the retraction:

| mutant at `4cf07a43` | result |
|---|---|
| whole-line `/* */` around `programs = programs;` | RED — already killed |
| whole-line `/* */` around `git = git;` | RED — already killed |
| `/* */` around `git = import ./git {};` | **GREEN — walkable** |
| **inline** `git = /* import ./git {} */ null;` | **GREEN — walkable** |

So one *link* was walkable, in two *shapes* — the inline one I had not measured. Mechanism: the first two assertions carry `^\s*`, which a leading `/*` defeats regardless of comment-stripping; the import regex was unanchored.

### One more false GREEN, found and fixed

Acting on round 5's stated suspicion rather than waiting for its verdict: the offset-alignment assumption **holds** (all three nix files identical in length across both passes; zero positions differing with a non-blank replacement). But measuring it exposed that `_top_level_of_settings` **located** on the blanked text while still **counting braces** on the keeping one — half a conversion:

```
settings = { alias.brace = "log --format=OPEN{"; };
core.sshCommand = "ssh -o ServerAliveInterval=30 …";   # OUTSIDE settings
  -> accepted   *** FALSE GREEN ***
```

A setting the guard's own docstring says must be refused satisfied it. Latent today (`nix/programs/git/default.nix` has **0** brace-bearing string offsets) but not fanciful — `nix/home.nix` has **314**. Round 5 found the same thing independently and rated it 🟡-latent; it is now closed.

That left the code asserting an invariant nothing checked, so `test_the_two_parser_passes_stay_offset_aligned` pins it, mutation-verified: blanking a quoted string to a different length fails it with **its own** message (`880 vs 772`).

Also: the header's test count had gone stale **within one commit of my writing it**, in the file whose own rule is "count the list rather than trusting a total maintained in parallel with the thing it counts". It said five; there are six.

### A process failure worth recording

Round 5's headline finding was that **the file changed under it mid-audit** — because I was editing while it read, having told it the worktree was stable. It pinned everything via `git archive` to the audited tip, so its numbers stand, but that was the auditor's save, not mine. Fifth time in this session I've mutated something a running job was reading; the checksum-freeze around suite runs exists for exactly this and I did not extend it to audits.

### State

Full suite: `TOTAL collected=17061 passed=17059 skipped=2 failed=0`, `RESULT: PASS (exit=0)`, four touched files sha256-verified unchanged across the run.

Round 5 said explicitly: do not merge on its strength, because the audited tip was no longer the tree. That delta is now committed as `65de660c` and a further round is owed against it.

ZacxDev:
## 🔴 Retraction: "the scan half fails safe" was false, and I published it

My previous comment said round 6's surviving mutant "reverts half of a paired change and fails safe", and used that to argue it was not a gap. **It is a gap.** Round 7 refuted it and I verified the refutation myself before acting:

```
narrow scan-half revert, MY fixture (nothing after `};`)  ->  ''
narrow scan-half revert, WITH a sibling attrset           ->  'ssh -o ServerAliveInterval=30 …'
```

The mechanism I asserted holds only while nothing follows the closing brace. The `{` inside the string extends the span past `settings`'s real `}` and depth goes to −1 — but **the next sibling's `{` brings it back to 0**, and everything after is emitted as top-level. A `core.sshCommand` that is not in `settings` at all is then accepted: the exact class the guard exists to refuse.

I measured at one point on a dimension that changes the answer, generalised it, and recorded it as settled fact. That is worse than the original gap — an assertion of safety is what stops the next reader looking. The fixture now carries a sibling attrset, with a comment saying why it must not be removed.

| mutant | before | now |
|---|---|---|
| narrow scan-half revert | SURVIVED — **false GREEN** | **KILLED** |
| emit-half revert | KILLED | KILLED |
| full revert of the depth fix | KILLED | KILLED |
| pre-delta arithmetic blanking | — | **KILLED** (`37 vs 38`) |

### Two more of my claims were wider than the measurement

- **The degenerate-shape loop was inert.** None of its fixtures contained `settings = {`, so `_top_level_of_settings(f)` returned at the locator and reached no offset arithmetic — while my commit message said "the downstream consumer is driven over each of them". Each fixture is now wrapped in a real `settings = {` block.
- **"Three of those raised IndexError" was inherited, not measured.** Driving the pre-delta blanking directly: exactly **one** of the five raises; the other four return cleanly. And under the real test the length assertion fires *first*, so the `IndexError` is **prevented, not observed**. How many reach it depends on where the degenerate tail sits — so the count isn't a property of the shapes, which is why the guard asserts **length equality** (a property of the parser) rather than counting exceptions. Both notes now state what was measured and name the scope.

Also added the shape the fixture list claimed but didn't contain: a **bare** `"` at EOF. The entry labelled "lone quote at EOF" was `a = "unterminated`, which measures 17/17 under both blankings and never diverged.

### Where this leaves the PR

Full suite: `TOTAL collected=17063 passed=17061 skipped=2 failed=0`, `RESULT: PASS (exit=0)`, four touched files sha256-verified unchanged across the run.

**The shipped change — `nix/programs/git/default.nix`, `scripts/claim-work.sh`, `scripts/resume-state.sh` — is unchanged since `c558eb0d` and has now been independently cleared by five consecutive audit rounds.** Every defect found from round 2 onward has been in the test file, and the last three were *claims written into the tree as settled fact* rather than logic errors.

I'm stopping the audit ladder here. Continuing means auditing the guard's guard, and the marginal value has crossed over — the parser is now the most complex thing in a diff whose actual payload is two lines of nix.

⚠ **Still not live.** This is "merged will be correct", not "#782 is fixed". The setting reaches a host only after `home-manager switch`, which I have not run.

⚠ **Separate, unfiled:** `githooks/install.sh` cannot arm the hooks *at all* on a home-manager host — it dies rc=255 on its pre-existing `core.hooksPath` line, so the pre-push gate this issue is about cannot currently be installed here. Pre-existing and independent of #782; deserves its own issue rather than widening this PR.

Issue #783, opened 2026-08-24
flake: test_frames_telemetry_metadata_only reads the FIRST spool event and asserts it is its own — under load it is a previous test's late getHtml timeout


ZacxDev:
`test_frames_telemetry_metadata_only` reads the FIRST event in the spool and asserts it is its own. Under load it is a previous test's late `getHtml` timeout, so the test fails on a PR whose diff cannot reach it — and because it is a **required** check, it blocks the merge.

## Measured

`devrc-ci-bdrs6`, commit `7a9ecf85` (PR #773, which touches only `scripts/run-tests.sh` and `scripts/tests/test_nogit_isolation.py`):

```
scripts/browser-bridge/tests/test_server.py:5632: AssertionError
>           assert p["op"] == "frames"
E           AssertionError: assert 'getHtml' == 'frames'
```

The captured stderr in the same run is the whole story:

```
{"ts":1787542746.877,"event":"cmd_timeout","op":"getHtml"}
{"ts":1787542746.877,"event":"cmd_timeout","op":"getHtml"}
{"ts":1787542746.899,"event":"dispatch","id":"36d9bf182204f839","op":"frames","key":"fake-1"}
{"ts":1787542746.901,"event":"cmd_ok","op":"frames"}
```

A **previous** test's `getHtml` timed out and its telemetry landed in the spool directory this test then read. The test does:

```python
e = _wait_events(spool_dir, 1)[0]
p = json.loads(e["payload"])
assert p["op"] == "frames"
```

`_wait_events(dir, 1)` returns as soon as **an** event exists; `[0]` then takes whichever arrived first. The test treats *"an event exists"* as *"my event exists"* — they are different claims, and they come apart exactly when the machine is loaded enough for the earlier command to time out.

## Why this is a defect, not a re-run

This is the family already recorded four times in this repo's index (`tests` scope, 2026-08-23): **a test treating an OBSERVATION as proof of a LATER state.** A response does not imply the audit line was printed; an `rglob` listing does not imply the file is still readable; *an event existing does not imply it is yours.* Re-running it green moves the bug rather than removing it, and the next PR to land during a busy window eats it again.

## Discriminating evidence that it is not the PR's diff

- The failing file is untouched by that PR.
- **Same tree, same tier, no cluster load → PASS**: `nix build .#checks.x86_64-linux.pytests` on the merged tree locally gave `RESULT: PASS (exit=0)`.
[...]

PR #891, opened 2026-08-26
test(browser-bridge): close the _wait_ops residual — select the rows a test CAUSED, not the first two of its op

ZacxDev:
## The exposure

`_wait_ops` (in `scripts/browser-bridge/tests/test_server.py`) narrowed the
#783 class — a neighbour's spool row being read as this test's — by selecting on
`op`. Its own docstring named what that leaves open: **discrimination is on `op`
ALONE**, so a caller asking for N rows of a *common* op **where their order
carries the signal** is still exposed. There was exactly one such site:

```python
absent, empty = _wait_ops(spool_dir, "tabs", 2)   # test_server.py:3434
```

in `test_an_absent_origin_header_is_not_the_same_as_an_empty_one`, which then
asserts *different* things about `absent` and `empty`. One foreign `tabs` row
landing between them silently swaps both assertions onto someone else's row.

`ACTIVITY_SPOOL_DIR` is a process-global env var that `conftest.
_isolate_activity_spool` re-points per test, so a thread still alive from an
earlier test writes into the *current* test's spool. That root cause is #783 and
is still open; this PR closes the one place where it can corrupt an assertion
without being visible.

## I re-measured the "safe today" claim first — it had NOT lapsed

The docstring justified the residual with a corpus property: *"no current test
leaves a `tabs` command in flight to time out (the two candidates call
`registry.submit` directly and never reach `emit_cmd_event`)"*. Re-measured
2026-08-26 at base `0716fede`:

| what | measured |
|---|---|
| call sites issuing `{"op":"tabs"}` in `scripts/browser-bridge/tests/` | **51** (27 `_cmd_sess`, 17 `_req`, 2 `_cmd`, 2 `reg.submit`, 2 direct `emit_cmd_event`, 1 `validate_command`) — not the 2 the old text implies |
| `tabs` rows emitted in one full `test_server.py` run | **75** — the most-emitted op in the suite |
| of the 51, how many are issued from a nested scope (lambda / inner fn), i.e. can run off the test's own thread | **0** |

**Method.** (1) AST walk over every `.py` in the tests dir, collecting every call
whose argument dict literal contains `"op": "tabs"`, together with its full
enclosing-scope stack. A command can only be *pending at teardown* if it is
issued asynchronously, which here means from a `threading.Thread` target — i.e.
a nested function or lambda. Every one of the 51 sits directly in a test
function body, on the test's own thread, so it completes before teardown.
(2) An instrumented full run: `Registry.submit` stamps the handler thread with
the nodeid current at request entry; `emit_cmd_event` records that stamp
alongside the nodeid current *at emit time*. Any row where they disagree is a
foreign row.

**So the property holds — but it was true for a different reason than it gave,
and it was never the only route in:**

* The real reason is (1) above, not "they call `registry.submit` directly".
* `emit_cmd_event` runs **after** the HTTP response (conftest's
  `_session_spool_backstop` documents this), so even a *synchronous* command's
  row can land past its own test's teardown. The pending-timeout route was never
  the only one.
* **The class is demonstrably LIVE for another op.** Two instrumented full runs,
  each: `test_cmd_queue_full_returns_429` issues two `getHtml` /cmds from
  **unjoined daemon threads** that are never answered, and its `cmd_timeout`
  fires ~5s later — putting **2 `{"op":"getHtml","outcome":"timeout"}` rows into
  a LATER test's spool**, a *different* victim test in each run
  (`test_type_telemetry_no_typed_text`, then
  `test_frames_telemetry_metadata_only`). That is precisely the shape that
  reddened #770 and #773. Nothing stops the next such test from using `tabs`.

A corpus property that has to be re-argued every time the corpus grows is not a
guarantee. This PR replaces it with an invariant the code enforces.

## The fix

* `_wait_ops` gains an optional **`where=`** row predicate, ANDed with the op
  test. Existing callers are untouched.
* `_routed_to(key)` is the ready-made predicate: it matches **`payload.key`**,
  the routing target the request named (`server.py` fills it from `target`).
* The exposed site routes both of its commands to an **instance id minted per
  run** (`uuid4`) and selects with `where=_routed_to(inst)` — so it gets the pair
  it caused *by construction*, not by an argument about what the rest of the
  corpus emits.
* `_payload_op` is now one line over a new total `_payload_field`, so the "must
  never raise on a foreign row" argument lives in one place.

### Why the routing key and NOT `session` / `sess_src`

The docstring named those as the remedy. On inspection they are the wrong
discriminator *for this site*:

* `sess_src` is the caller's **tier** (`"claude"`) — every neighbour on that tier
  shares it.
* `session` is **precisely what this test asserts**. A filter built out of the
  thing under test is not a control: an attribution regression would stop the row
  matching and surface as a 10s *wait timeout* instead of the attribution
  assertion. Mutant **M4** below dies exactly that way, which is the evidence.

`payload.key` is orthogonal to session attribution, so a regression in the code
under test still fails on the assertion that names it.

### The docstring now matches the code

The "WHAT THIS DOES **NOT** COVER" block asserted an exposure this PR closes;
leaving it would be a false claim in the opposite direction. It is replaced with
the `where=` contract, the reason the discriminator is the routing key, and the
re-measurement above — recorded as the reason the corpus property was *removed*,
not as a property anyone should rely on again.

## Coverage

`test_a_neighbours_row_of_the_same_op_is_not_selected_as_one_of_ours` **plants**
a foreign `tabs` row (same `op`, same `kind`, its own session, a different
routing key) rather than racing for one — same discipline as the neighbouring
`test_a_neighbours_late_row_does_not_become_this_tests_event`. It carries an
in-test **control** that the op-only form really does take the planted row first,
so it cannot pass on an uncontested spool.

**Red / green matrix — base sha `0716fede8adba08f8d14afe64bd763d7bd83b9e5`**
(the sha this work was authored and swept at; the branch has since been rebased
onto `a5aa037b`, and the gates below name that tree).
The new test uses an API that does not exist at base, so "red at base" is shown
as the base *behaviour* with the new test in place — mutant M1, `_wait_ops`
selecting on op alone (drop the filter, keep the slice):

```
M1 (= base behaviour)   test_a_neighbours_row_of_the_same_op_...   FAILED
                        test_an_absent_origin_header_...           passed
HEAD                    both                                       passed
```

M1's failure is on the assertion that matters, naming the planted row:

```
>       assert first["session"] == LEAKED_UUID, first
E       AssertionError: {'source': 'browser-bridge', 'kind': 'cmd',
E         'session': '0b7f4c31-5555-4ccc-9aaa-000000000005',
E         'payload': '{"op": "tabs", "key": "a-neighbours-instance", ...}'}
E       assert '0b7f4c31-555...-000000000005' == 'c05e17b6-333...-000000000003'
```

(`test_an_absent_origin_header_…` passing under M1 is expected and is the point:
it has no foreign row of its own, so the *regression test* is the guard.)

## Mutation sweep

Run under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` purged between mutants,
each applied from a `cp` of the committed file (never `git checkout --`).
Semantic mutants, not deletions. **All killed:**

| mutant | killing message |
|---|---|
| **M1** drop the filter, keep the slice | `assert '0b7f4c31-…5' == 'c05e17b6-…3'` on `first["session"] == LEAKED_UUID` |
| **M2** weaken `_routed_to` to a presence test (`is not None`) | same assertion, same message |
| **M3** `and` → `or` in `_match` | same assertion, same message |
| **M4** compare the wrong payload field (`"op"` for `"key"`) | `spool never satisfied the wait condition in 10.0s; got 2 event(s): …` — kills **both** tests; the wait-timeout shape, i.e. the failure mode a session-based filter would have given us for a *real* regression |
| **M5** positive control on the test: unplant the foreign row | `the planted neighbour row is not selected first by the op-only form, so this test is not reproducing the exposure it covers` |

M1/M2/M3 all die on the assertion under test, not incidentally.

## Verification

**Dev-host tier** (`scripts/gate.sh --tier both`) at base `0716fede` + this
commit: `GATE: RESULT=PASS exit=0` — pytest `collected=16989 passed=16987
skipped=2 failed=0 (floor 15789)`, node `tests=1292 pass=1292 fail=0 (floor
1239)`.

A second run of **both** tiers on the rebased tree (base `a5aa037b`) —
dev-host `scripts/gate.sh --tier both` and the sandbox tier Tekton gates on,
`nix build .#checks.x86_64-linux.{pytests,nodetests}` — is reported in the PR
comments.

Not verified: nothing. This change is test-only; no runtime code moved.

### DIFF
--- a/scripts/browser-bridge/tests/test_server.py
+++ b/scripts/browser-bridge/tests/test_server.py
@@ -25,6 +25,7 @@ import time
 import urllib.error
 import urllib.parse
 import urllib.request
+import uuid
 from pathlib import Path
 from unittest import mock
 
@@ -171,8 +172,11 @@ def _wait_events(spool_dir, n=1, timeout=10.0, until=None) -> list:
     CI as `assert 'getHtml' == 'frames'` and `assert 'getHtml' == 'type'`. Use
     `_wait_ops` / `_wait_payload` below to select by op rather than by position.
     The `absent, empty` unpack named above as the order-dependent exception is
-    now `_wait_ops(spool_dir, "tabs", 2)`, which keeps the order and drops
-    foreign rows.
+    now `_wait_ops(spool_dir, "tabs", 2, where=_routed_to(inst))`, which keeps
+    the order and drops foreign rows — op alone would NOT have, because both of
+    its rows are `tabs` and a neighbour's `tabs` row selects just as well. See
+    `_wait_ops` for why the extra discriminator is the ROUTING KEY and not the
+    session.
 
     🔴 AND A TIMEOUT IS NOW LOUD. This used to return a SHORT list silently, so
     every caller's next line — `[0]`, or a filter — failed with a message about
@@ -206,9 +210,9 @@ def _wait_events(spool_dir, n=1, timeout=10.0, until=None) -> list:
     return evs
 
 
-def _payload_op(e) -> str | None:
-    """The `op` inside a spooled event's payload, or None if it has no readable
-    one.
+def _payload_field(e, field) -> str | None:
+    """The named field inside a spooled event's payload, or None if the row has
+    no readable one.
 
     Total by construction — a row written by something other than the bridge
     must not raise here, because the whole point of the helpers below is to walk
@@ -221,14 +225,41 @@ def _payload_op(e) -> str | None:
     `payload` key and a `None` payload were actually covered. A non-bridge
     writer emitting a scalar payload is exactly the named case, so the claim and
     the code disagreed precisely where it mattered.
+
+    ONE reader for every payload field, so the totality argument above is made
+    once instead of re-derived per accessor — `_payload_op` and `_routed_to`
+    below are both this function.
     """
     try:
-        return json.loads(e["payload"]).get("op")
+        return json.loads(e["payload"]).get(field)
     except (KeyError, TypeError, ValueError, AttributeError):
         return None
 
 
-def _wait_ops(spool_dir, op, n=1, **kw) -> list:
+def _payload_op(e) -> str | None:
+    """The `op` inside a spooled event's payload, or None. See _payload_field."""
+    return _payload_field(e, "op")
+
+
+def _routed_to(key):
+    """A `_wait_ops(where=…)` predicate: the row names `key` as its ROUTING
+    TARGET (`payload.key`, which server.py fills from the request's `target`).
+
+    THE DISCRIMINATOR IS DELIBERATELY ORTHOGONAL TO SESSION ATTRIBUTION. A
+    caller that routes its commands to an instance id minted for that run can
+    select exactly the rows it caused, and a neighbour's row of the same op
+    carries a different key (or none) and is dropped — WITHOUT the filter being
+    built out of `session`/`sess_src`, which is what the one site needing this
+    actually asserts. Filtering on the thing under test would turn a real
+    attribution regression into a wait timeout; see `_wait_ops`.
+
+    Total for the same reason `_payload_field` is: a foreign or malformed row
+    must be walked past, never raise.
+    """
+    return lambda e: _payload_field(e, "key") == key
+
+
+def _wait_ops(spool_dir, op, n=1, where=None, **kw) -> list:
     """The first `n` spooled events whose payload op is `op`, waiting for them.
 
     🔴 USE THIS INSTEAD OF `_wait_events(spool_dir, n)[i]` WHENEVER YOU WANT A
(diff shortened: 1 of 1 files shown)


ZacxDev:
## Verification — both tiers, on the rebased tree

Branch head **`4c935815b0fa75d5dae13a3b5a42fd8e0238e476`**, base
**`a5aa037b`** (`origin/main` at the time of the run). Working tree clean at both
runs.

**Dev-host tier** — `nix develop . --command bash scripts/gate.sh --tier both`:

```
======================== GATE ========================
GATE: RESULT=PASS exit=0
```

node: `TOTAL suites=5 files=39 tests=1292 pass=1292 fail=0 skipped=0 (floor 1239)`

**Sandbox tier — the one Tekton gates on** —
`nix build .#checks.x86_64-linux.{pytests,nodetests}`, verdicts read out of the
derivations' own logs rather than the exit code:

```
devrc-pytests   PASS  scripts/browser-bridge/tests  (collected=791 passed=791 skipped=0 floor=716)
                TOTAL collected=17056 passed=17054 skipped=2 failed=0  (floor 15789)
                RESULT: PASS (exit=0)
devrc-nodetests TOTAL suites=5 files=39 tests=1292 pass=1292 fail=0 skipped=0  (floor 1239)
                RESULT: PASS (exit=0)
```

An earlier dev-host run at the pre-rebase base `0716fede` was also
`GATE: RESULT=PASS exit=0` (pytest `collected=16989 passed=16987 skipped=2
failed=0`).

`test_browser_agent.py`'s known intermittent `:558` subprocess hang did not fire
in any of these runs.

**Not verified:** nothing outstanding. This change is test-only — no runtime code
moved, so there is no deploy step and nothing to re-check against a live
consumer.

PR #895, opened 2026-08-26
feat(worktree-prune): spare other sessions' worktrees BY DEFAULT — and the round-2 safety flag that blocked itself

ZacxDev:
## Round 3 (head `f0bf05d4`) — the safety flag blocked itself, and the default is now inverted

> Everything below the `---` is the round-1/2 body, kept for history. **Where it and this
> section disagree, this section is current** — in particular `--skip-agent-worktrees` is
> GONE and the numbers below supersede the ones there.

A blind delta re-audit of round 2 found one 🔴 that round 2's own fix *introduced*, plus
two 🟡 and four 🟢. Round 2 was green everywhere — 17,099 passed / 0 failed on an
independent merged-tree gate, both Tekton required checks `success` — and green saw
**none** of it.

### 🔴 1. The safety flag blocked itself, and the only escape was turning safety off

Round 2 made `--execute` REFUSE while **any** exclusion glob matched zero rows. The
agent-worktree glob *is* a glob, so on any repo with no `.claude/worktrees` directory:

```
$ worktree-prune --repo <repo> --skip-agent-worktrees --execute --confirm 1
rc=3
REFUSED: '*/.claude/worktrees/*' matched ZERO of the 2 scanned row(s). … Fix the glob … or drop it.
```

The remedy the message offers is unactionable — that glob is a hardcoded constant,
spelled correctly. The only way through was **dropping the safety flag**. That is a
permanently-red gate whose click-through is the catastrophic action itself.

**Fix:** the refusal now applies only to globs the operator **typed**. A hardcoded
constant cannot be mistyped, so it carries none of the hazard the refusal was built for.

### 🟡 2. `--allow-unmatched-globs`, because a zero match cannot tell a typo from scope

`--exclude-path '*/.claude/worktrees/*' --exclude-path '*/civitai/*'` against a one-repo
scan was refused with both globs named — and both are correct fleet-wide. An empty result
cannot distinguish *mistyped* from *out of this scan's scope*; round 2 picked the
dangerous reading and hard-stopped with no escape hatch. The new flag downgrades the stop
to a loud, still-per-glob WARNING and is echoed prominently in the report, because an
override invisible in the output is a hardcode wearing a costume.

Fixed at the same site: a scan yielding **zero rows** had started refusing, where
`--execute --confirm 0` used to succeed as a no-op. With no rows nothing could have
matched, so the zero is evidence about the scan, not the glob.

### 🟡 3. A new crash class this feature introduced

`_path_candidates` caught only `OSError`, but a **non-strict** `Path.resolve()` still
raises. Measured at two points on the same looped-symlink path, 2026-08-27:

| interpreter | non-strict `Path(looped).resolve()` |
|---|---|
| CPython **3.12.14** (this dev shell) | raises `RuntimeError("Symlink loop from …")` |
| CPython **3.13.15** | returns the path unresolved, no raise |

That split is exactly why the `except` is widened rather than the 3.13 behaviour relied
on. Before path exclusion existed this code never touched the filesystem, so no path's
*shape* could kill a scan; it could now — in `scan_repo` **and** again in `summarize`.
One `_resolve_or_none` helper now serves every `.resolve()` call site
(`_path_candidates`, `collect_repos`, `discover_repos`). Four new tests build a real
symlink loop, including one proving an **ordinary** row in the same scan was lost too:
the traceback did not cost one row, it cost the run.

### 🔵 4. Default inversion (new work, not an audit finding)

By this PR's own scan, **137 of 236 dead rows are other live sessions' working trees**.
Sparing them is now the **default**.

* `--include-agent-worktrees` opts back in, and says so in `--help` and — loudly — in the
  report.
* **`--skip-agent-worktrees` is removed, not aliased.** The PR is unmerged so there is no
  compatibility to keep, and a flag that does nothing is worse than one that does not
  exist. A test asserts the parser rejects it.

Consequences, worked through rather than assumed:

* **`--confirm N` counts change**: on the live scan 236 → 99. A remembered N from before
  is now *refused*, which is the safe direction.
* **The report cannot be read as "everything was covered."** Immediately under the
  removable count it now prints how many further `dead` rows were SPARED — with the
  exclusion on by default the removable count is routinely far short of the dead count.
* **The interaction with #1 stays consistent**: the implied constant is never a refusal
  candidate, so the default cannot block itself. Typing the same glob by hand *does* make
  it one — byte-identical pattern, byte-identical zero, opposite outcome, and the only
  difference is who spelled it. Both directions are tested.

### 🟢 Also fixed

* **Credited-glob order was unpinned.** A mutant swapping `path_excluded`'s loop nesting
  to candidate-major **SURVIVED all 132 tests**, because the only ordering test used
  direct-path matches, where the two nestings cannot disagree.
  `path_excluded('/a/b/c', ['/a/b', '*/c'])` is now pinned to `'/a/b'` (ancestor-vs-self),
  with the mirror case, a deeper case, and an assertion that per-glob counts genuinely do
  **not** move with the order.
* **A whitespace-padded glob slipped both new guards.** The resolver stripped `/` then
  tested `.strip()`; the CLI stripped `.strip().rstrip("/")`. So `--exclude-path
  '/home/z/repo/ '` passed the CLI check and reached the matcher with both its slash and
  its space. One `normalize_glob` now serves both sites.
* **Stale claims in the test file.** Its `870 / 128 → 250 dead … 131 … agent-*` block and
  the `worktrees/agent-*` spelling both described the tool as it is **not**.
* **JSON schema change, documented.** `summary.exclude_globs` went `list[str]` →
  `list[dict]` and the dicts now carry `typed`; `summary` also gained
  `exclude_globs_blocking_execute`, `agent_worktrees_included` and
  `allow_unmatched_globs`. No in-repo consumer, but `--out` is advertised as
  machine-readable, so it is noted in `scripts/README.md` too.

### 🔴 The numbers are re-derived; the old ones are retired

The auditor could not verify round 2's "231 dead, all 231 removable", and it was a large
unexplained move from an earlier "250 dead, 106 removable". One fresh dated scan, scope
stated so anyone can re-run it:

```
2026-08-27  worktree-prune --scan-root /home/zach/workspace --scan-depth 2
860 worktree(s) across 129 repo(s): 236 dead, 43 orphan, 528 live, 53 cannot-tell
99 row(s) would be removed by --execute (dead + clean + unlocked only)
🔴 …and 137 further `dead` row(s) were SPARED by an exclusion filter — NOT in the
   number above, and NOT covered by this run.

exclusion filters in force (1):
    (default; --include-agent-worktrees turns it off) '*/.claude/worktrees/*'
        ->  matched 246 row(s), 137 of them dead
```

* This **reproduces** round 2's shape — 231 dead / 135 agent-dead became 236 / 137 a day
  later on a moving box — so that claim is substantiated, not merely repeated.
* The earlier **250 / 131 / 106** trio does **not** reconcile: 250 − 131 is 119, not 106,
  so at most two of the three were ever right, and the run behind them cannot be
  re-derived. Those numbers are **deleted** from the docstring, the test file and
  `scripts/README.md` rather than restated.
* Measured drift, recorded so a re-run that disagrees is not read as a bug: two runs the
  same day gave `236 dead / 137 agent-dead / 99 removable` **identically**, and moved
  `orphan`/`cannot-tell` 44/52 → 43/53.

### Verification

**Merged-tree gate** (built off *current* `origin/main` `117116ea`, merged clean; verdict
counted from the log content, not an exit code):

```
pytest tier: 36/36 targets PASS, 0 FAIL   collected=17298 passed=17296 skipped=2 failed=0
node   tier:  5/5  targets PASS, 0 FAIL   files=39 tests=1292 pass=1292 fail=0
GATE: RESULT=PASS exit=0      no `panic:` and no timeout anywhere in either log
```

159 of the pytest tier's node ids are `test_worktree_prune.py::` — confirmed by
`--collect-only` on the merged tree, so the gate's PASS demonstrably covers this file
(132 → 159 tests).

**Mutation battery, re-run in full** (a fix round resets the gate — not just mutants for
what changed): **29 mutants, 0 survivors**, `PYTHONDONTWRITEBYTECODE=1`, tool restored by
content with `sha256` verified equal.

Both instrument controls were watched:

| control | verdict | required |
|---|---|---|
| POSITIVE (`classify` always `DEAD`) | KILLED (51 failed) | KILLED |
| NO-OP (comment-only edit) | SURVIVED (158 passed) | SURVIVED |

The previously-**SURVIVING** loop-nesting mutant now dies. A second attribution pass
confirms every round-3 mutant is killed by the test that *owns* it, with zero collection
errors — never by some other guard's error:

| mutant | killed by |
|---|---|
| `glob-order-candidate-major` | `test_the_credited_glob_is_glob_major_not_candidate_major` |
| `default-back-to-opt-in` | 16 tests incl. `test_the_default_resolves_to_exactly_the_documented_glob` |
| `opt-in-flag-ignored` | `test_opting_back_in_to_agent_worktrees_needs_confirm_two` (+4) |
| `refusal-scans-all-globs` | `test_the_default_exclusion_matching_zero_rows_does_not_refuse` (+2) |
| `typed-always-true` / `typed-always-false` | 7 / 9 tests |
| `allow-unmatched-ignored` | `test_allow_unmatched_globs_warns_and_proceeds` |
| `empty-scan-guard-off` | `test_an_empty_scan_with_confirm_zero_is_a_no_op_not_a_refusal` |
| `resolve-except-narrowed` | the 4 symlink-loop tests |
| `spared-line-off` / `include-shout-off` / `allow-shout-off` | their 3 report tests |
| `dangerous-help-softened` | `test_the_dangerous_flag_says_so_in_help` |

`--execute` was never run outside hermetic tmpdir fixtures. The live scans were dry-run.

---

## The measured problem

`worktree-prune --execute --confirm N` removes **every** `dead` row in scope, and
repo-level scoping (`--repo` / `--repos-from` / `--scan-root`) was the only filter
there was. That is not fine-grained enough to be safe.

A full scan of this machine, run today with the code in this PR:

```
860 worktrees across 129 repos -> 231 dead, 44 orphan, 529 live, 56 cannot-tell
231 row(s) would be removed by --execute
```

**135 of those 231 `dead` rows are `.claude/worktrees/agent-*` worktrees belonging to
OTHER LIVE Claude sessions** (measured directly by the new flag — see the two scans at
the bottom). `dead` is a correct verdict about the *branch* and a catastrophic
instruction about the *directory*: removing one destroys a working directory that is in
use right now. An earlier scan put the same figure at 131 of 250; the exact numbers move
with the box, the shape does not.

Dodging them with the filter that existed meant excluding whole repos, which collapsed
the safe set to a small fraction, because a repo like `civit/civitai` holds dozens of
ordinary dead rows *and* agent worktrees. With this PR, `--skip-agent-worktrees` leaves
**96** genuinely-dead worktrees removable while sparing all 135.

## What this adds

* **`--exclude-path GLOB`** (repeatable) — spare any worktree whose absolute path
  matches `GLOB`.
* **`--skip-agent-worktrees`** — shorthand for
  `--exclude-path '*/.claude/worktrees/agent-*'`. Implemented by appending the same
  constant to the same list and running the same predicate; the glob literal is
  spelled exactly once in executable code, and a test asserts that.

Behaviour, deliberately:

1. **An excluded row is still scanned, classified and printed.** It does not vanish
   from the report — silent truncation reads as "we covered everything", which is
   the failure this tool exists not to have. It prints as `dead (excluded)`, gets its
   own per-repo column, and the summary states how many rows were excluded and how
   many of those were `dead` (i.e. what the filter actually spared).
2. **Never removed, and not counted toward `--confirm N`.** `--confirm N` keeps
   meaning "the exact number that will actually be removed". Enforced in two
   independent places: a blocker in `classify()`, and a re-assert in
   `execute_removals()` immediately before the destructive step rather than trusting
   the `removable` flag it was handed.
3. **Glob semantics, chosen and pinned:** `fnmatch.fnmatchcase` against the absolute
   path git reports, and **`*` crosses `/`**. Not shell/`pathlib` globbing. That is
   what makes `*/.claude/worktrees/agent-*` match at any depth. `fnmatchcase` rather
   than `fnmatch` so the answer does not depend on the host's case-folding. The cost
   — a glob is easier to write too wide than too narrow — is safe in this tool's only
   direction: a wide glob excludes *more* from removal, never less.

`--help`, the module docstring and `scripts/README.md` all state the slash-crossing
choice, because that is the thing an operator will get wrong.

## Test matrix

`scripts/tests/test_worktree_prune.py`. Full gate (`scripts/run-tests.sh` under
`nix develop`): **PASS 48   FAIL 0** — `TOTAL collected=17013  passed=17011  skipped=2
failed=0`, `RESULT: PASS (exit=0)`. Counted from the per-target result lines, not from
an exit code; the output contains no timeout or panic line.

New fixture `two_dead`: one repo with **two** dead worktrees — one at
`<repo>/.claude/worktrees/agent-7f3a91` (squash-merged, so `dead` by
`content-identical`), one at an ordinary path (true-merged, so `dead` by
`ancestor`).

| test | pins |
|---|---|
| `test_without_any_filter_both_dead_worktrees_are_removable` | POSITIVE CONTROL — unfiltered, both rows really are removable, so every assertion below is not about a fixture that was never removable |
| `test_execute_removes_the_plain_dead_tree_and_spares_the_excluded_one[--skip-agent-worktrees / --exclude-path]` | BOTH HALVES in one run: the excluded worktree survives on disk *and* the other is gone (`removed=1`) |
| `test_a_full_execute_pass_over_the_scanned_rows_never_touches_an_excluded_path` | THE SEAM. `--confirm` and the exclusion are two guards in series; the sweep showed the confirm mismatch killing every end-to-end test *first*, leaving the `is_dir()` assertion unreachable. This one feeds the tool's own removable count back in, so the disk assertion runs whichever floor is broken |
| `test_confirm_counts_only_the_rows_that_will_actually_be_removed` | `--confirm 1` (non-excluded dead count) SUCCEEDS |
| `test_confirm_with_the_total_dead_count_is_refused_when_one_is_excluded` | `--confirm 2` (total dead count) is REFUSED and removes nothing |
| `test_without_the_filter_the_same_scope_needs_confirm_two` | control on the row above — 2 is the right number when nothing is excluded, so the refusal is caused by the exclusion, not by the fixture |
| `test_excluded_rows_still_appear_in_the_text_report` / `..._json_report` | the excluded path, `[dead (excluded)]`, the summary line, `excluded`/`excluded_dead`/`exclude_globs`, and `2 dead` still in the totals |
| `test_an_unfiltered_run_does_not_shout_about_exclusions` | negative control on that report line |
| `test_the_glob_star_crosses_slashes` | the documented semantics, both halves (matches at depth; does not match `.claude/worktrees/keep-me`) |
| `test_the_glob_match_is_case_sensitive`, `test_an_empty_glob_list_excludes_nothing`, `test_the_first_matching_glob_is_the_one_reported` | matcher edges |
| `test_skip_agent_worktrees_resolves_to_exactly_the_documented_glob`, `test_the_convenience_flag_and_the_explicit_glob_produce_identical_rows`, `test_the_agent_glob_literal_is_spelled_exactly_once_in_executable_code` | one rule, one place — behaviourally and structurally |
| `test_execute_refuses_an_excluded_row_even_if_removable_was_forced` | the executor's own floor |
| `test_an_excluded_row_is_unremovable_whatever_its_verdict` / `test_exclusion_does_not_change_the_verdict_itself` | exclusion changes removability, not the verdict |
| `test_neutering_the_path_predicate_makes_the_agent_worktree_removable` | in-suite MUTATION CONTROL |
| `test_the_help_text_documents_the_new_flags_and_the_glob_semantics` | `--help` states `CROSSES` and the glob |

### Mutation sweep

Run with `PYTHONDONTWRITEBYTECODE=1` and `__pycache__` deleted between mutants
(CPython's whole-second mtime cache can otherwise score a mutant SURVIVED without
ever executing it). The tool was restored by content after each mutant —
`sha256 f17d9ba0…` before and after, verified.

|---|---|---|
| **M1** `path_excluded` always returns `None` | 13 failed / 99 passed | every exclusion test; end-to-end fails with `REFUSED: --confirm 1 does not match the 2 row(s) this run would remove` |
| **M2** `classify()` blocker off (`if excluded_by:` → `if False:`) | 9 failed / 103 passed | `test_the_excluded_row_is_still_classified_dead_and_marked` (`removable` True), both `--confirm` tests, `test_an_excluded_row_is_unremovable_whatever_its_verdict`. The executor floor still held, so nothing was deleted — correct defence in depth |
| **M3** `execute_removals()` floor off | 1 failed / 111 passed | `test_execute_refuses_an_excluded_row_even_if_removable_was_forced` — stderr `REMOVED …/agent-7f3a91`, assertion `agent.is_dir()` |
| **M5** BOTH floors off | 11 failed / 101 passed | including the seam test, with its own message: `AssertionError: an excluded worktree was removed by a full execute pass`, stderr `REMOVED …/agent-7f3a91` + `REMOVED …/wts/plain`, `removed=2` |
| **M4** POSITIVE CONTROL — `AGENT_WORKTREE_GLOB` → `*/no-such-pattern-zzz/*` | 9 failed / 103 passed | a mutant known to be caught, proving the sweep harness executes the mutated module at all |

No mutant survived.

Also run: the new tests against `origin/main`'s `scripts/worktree-prune` → **22 failed
/ 90 passed**. That red is *not* meaningful evidence on its own (the flags do not exist
there, so `AttributeError`/argparse failures are expected) — it is reported only to show
the tests genuinely depend on the new code. The load-bearing reds are M1–M5 above, which
are HEAD builds with only the filter broken.

## Real-machine evidence (read-only, `--execute` never passed)

Two `--scan-root /home/zach/workspace --scan-depth 2` runs, back to back. `--execute`
was never passed; both printed `DRY RUN — nothing was removed.`

Without the flag (22:35:09Z → 22:37:59Z):

```
860 worktree(s) across 129 repo(s): 231 dead, 44 orphan, 529 live, 56 cannot-tell
231 row(s) would be removed by --execute (dead + clean + unlocked only)
🔴 56 row(s) are cannot-tell — they are NOT dead and are NOT removable. …
```

With `--skip-agent-worktrees` (22:37:59Z → 22:40:00Z):

```
862 worktree(s) across 129 repo(s): 231 dead, 44 orphan, 530 live, 57 cannot-tell
96 row(s) would be removed by --execute (dead + clean + unlocked only)
243 row(s) matched an --exclude-path glob — still classified and counted above, NEVER
removed, and NOT counted toward --confirm. 135 of them are `dead` and would otherwise
have been removed. Re-run with --verbose to see them by path.
🔴 57 row(s) are cannot-tell — they are NOT dead and are NOT removable. …
```

`231 dead` is **identical** in both runs — the flag moves rows out of the *removable*
set, it does not hide them. `231 − 135 = 96`, exactly the removable count in run B. So
135 agent worktrees belonging to other live sessions came off the delete list, and 96
genuinely-dead worktrees became reachable that whole-repo exclusion could not reach.

Honest caveat: the two runs are 2 minutes apart on a box with ~18 concurrent agent
sessions, so the tree moved under them — 860 vs 862 rows, `live` 529→530 and
`cannot-tell` 56→57. Two worktrees appeared between the runs. The `dead` count and the
231/135/96 arithmetic are unaffected.

A `--verbose` run on `devrc` alone shows the labelling live:

```
[dead (excluded)] /home/zach/workspace/devrc/.claude/worktrees/agent-a03851193e2763746  branch=feat/present-explainer
[cannot-tell (excluded)] /home/zach/workspace/devrc/.claude/worktrees/agent-a003ab4bdfa908ef1  branch=None
…
66 worktree(s) across 1 repo(s): 26 dead, 10 orphan, 26 live, 4 cannot-tell
5 row(s) would be removed by --execute (dead + clean + unlocked only)
42 row(s) matched an --exclude-path glob — … 21 of them are `dead` …
```

---

# Round 2 — audit fixes (`b5ac31db`)

A blind adversarial audit returned "merge after fixing 2 🔴". Every finding was in
one blind spot: **the filter matching NOTHING looked exactly like the filter not
being passed.**

## 🔴 1 — a glob matching zero rows was byte-identical to no glob

The glob echo was gated on `summary["excluded"]`, and the glob list only reached the
JSON. Reproduced, and re-verified on the fix:

```
$ diff <(worktree-prune --repo …/devrc --no-gh) \
       <(worktree-prune --repo …/devrc --no-gh --exclude-path '*/.claude/wortrees/*')
> exclusion filters in force (1):
>     --exclude-path '*/.claude/wortrees/*'  ->  matched 0 row(s), 0 of them dead
> 🔴 --exclude-path '*/.claude/wortrees/*' matched ZERO rows. A glob that matches
>    nothing is BYTE-IDENTICAL to passing no filter at all, and --confirm N cannot
>    catch it because N is derived from the same count. …
```

The report now **always** echoes the effective glob list with a **per-glob** match
count (per-glob, because a total is non-zero the moment one of two globs works) and
names every zero-matcher.

**Beyond what was asked — flagging it for review:** `--execute` now **REFUSES** while
any glob matches zero rows (`RC_EXECUTE_REFUSED`), rather than only warning. A warning
prints above a 231-row table, and `--confirm` provably gives no independent protection
here. The escape hatch is to fix or drop the glob. If you'd rather it warn only, say so
and I'll drop the refusal — the warning and the per-glob counts stand on their own.

## 🔴 2 — `--skip-agent-worktrees` was narrower than its own help text

`AGENT_WORKTREE_GLOB` is now `*/.claude/worktrees/*`, not `…/agent-*`. Of 246
`.claude/worktrees/` entries on this box, one has no `agent-` prefix —
`…/fast/comfyui/.claude/worktrees/card-ux`, a real registered worktree on
`refs/heads/feat/prefopt-card-ux`, spared today only by the dirty check. A guard whose
description is wider than its body reads as coverage while providing none.

## 🟡 Also fixed

* trailing `/` stripped in `resolve_exclude_globs` (rows never carry one; tab
  completion always adds one);
* the symlink-resolved path is matched as well as the recorded one;
* **`--exclude-path <dir>` now covers the directory's contents.** Chose **subtree
  semantics** — a glob matching any ANCESTOR excludes the worktree — over
  auto-appending `/*` (which stops the glob matching the directory itself, and is
  meaningless for a glob already ending in `*`) and over a second `--exclude-tree`
  flag (a second rule is a second thing to drift out of sync);
* an empty `--exclude-path` is a **usage error**, not a silently-dropped no-op;
* `exclude_globs` now comes out of `summarize()` instead of being bolted on by
  `main()`, so the summary dict has one shape whatever the caller.

🔴 **Correction to the audit:** the escape helper is **`glob.escape`** — `fnmatch` has
**no** `escape` at all. Following the suggested wording would have shipped a `--help`
naming a nonexistent function; a test now asserts `--help` only names one that exists.

## 🟡 The three mutants that survived the audit's sweep

* **M7** (per-repo `excluded` column forced to 0) — the suite pinned the header, the
  summary sentence and the `[dead (excluded)]` label but never a **column value**. Now
  asserted, with a negative control on the table-row parser.
* **M1** (`fnmatchcase` → `fnmatch`) — survived because `os.path.normcase` is identity
  on POSIX, while the test's docstring claimed to pin platform-independent case
  handling. Now pinned **behaviourally** by monkeypatching `normcase` to fold case,
  with a positive control that the patch takes effect. The old docstring's overclaim is
  rewritten to say what it actually tests.
* **M9** (empty-glob filter) — now tested.

## 🟢 Pushed back on one nit

The audit said *"excluded `dead` rows already print without `--verbose`, so the
'Re-run with --verbose' line is misleading."* **Measured false** — the per-row loop is
entirely inside `if verbose:`:

```
$ worktree-prune --repo …/devrc --skip-agent-worktrees --no-gh | grep -c '^\['      -> 0
$ worktree-prune --repo …/devrc --skip-agent-worktrees --no-gh --verbose | grep -c '^\[' -> 67
```

The line is correct and stands. What probably misled the auditor was a dead
`if r["verdict"] == LIVE and not verbose: continue` **inside** the `if verbose:` block —
unreachable, and now removed.

## Not changed, deliberately

The audit's finding #12 proposed inverting the default (on by default, with
`--include-agent-worktrees`). **Left alone — that is Zach's call, not ours.** Open
question for review.

## Round-2 verification

**Merged-tree gate** — built off current `origin/main` (the branch was 8 commits
behind and overlaps `main` on `scripts/README.md`; main's changes there were two new,
non-overlapping table rows, read and confirmed):

```
pytest: PASS 48   FAIL 0
        TOTAL collected=17101  passed=17099  skipped=2  failed=0  (floor 15789)
node:   TOTAL suites=5 files=39 tests=1292 pass=1292 fail=0 skipped=0  (floor 1239)
GATE:   RESULT=PASS exit=0
```
Counted from the per-target result lines; zero timeout/panic lines in either log.

**Full mutation battery re-run** (a fix round resets the gate), `PYTHONDONTWRITEBYTECODE=1`,
`__pycache__` cleared between mutants, tool restored by content after each
(sha256 verified) — **14 mutants, 0 survivors**:

| mutant | killed by |
|---|---|
| R1-M1 predicate always `None` | 22 tests |
| R1-M2 `classify()` blocker off | 11 tests |
| R1-M3 executor floor off | `test_execute_refuses_an_excluded_row_even_if_removable_was_forced` |
| **R1-M4 POSITIVE CONTROL** (glob → nonsense) | 12 tests |
| **AUDIT-M1** `fnmatchcase`→`fnmatch` *(survived before)* | `test_the_matcher_does_not_route_through_normcase` |
| **AUDIT-M7** column always 0 *(survived before)* | `test_the_per_repo_table_reports_the_excluded_count_in_its_column` |
| **AUDIT-M9** empty-glob filter off *(survived before)* | `test_an_empty_glob_is_dropped_by_the_resolver` |
| NEW-D glob narrowed back to `agent-*` | 7 tests, incl. the end-to-end non-agent case |
| NEW-E glob echo re-gated on match count | `test_a_mistyped_glob_…`, `test_the_report_of_a_mistyped_glob_differs_from_an_unfiltered_run` |
| NEW-F execute zero-match refusal off | `test_execute_refuses_while_any_glob_matches_zero_rows`, `test_a_working_glob_alongside_a_dud_still_refuses` |
| NEW-G trailing-slash strip off | 3 tests |
| NEW-H subtree (ancestor) candidates off | 3 tests |
| NEW-I symlink-resolved candidate off | `test_a_glob_on_the_resolved_path_matches_a_symlinked_repo` |
| NEW-J per-glob count tallied from `excluded_by` | `test_per_glob_counts_are_independent_of_shadowing` |

**Fresh dry-run scan** (`--execute` never passed):

```
870 worktree(s) across 129 repo(s): 233 dead, 46 orphan, 534 live, 57 cannot-tell
97 row(s) would be removed by --execute (dead + clean + unlocked only)

exclusion filters in force (1):
    --exclude-path '*/.claude/worktrees/*'  ->  matched 246 row(s), 136 of them dead
246 row(s) matched at least one glob — still classified and counted above, NEVER
removed, and NOT counted toward --confirm. 136 of them are `dead` and would otherwise
have been removed.
```

246 matched — exactly the audit's count of `.claude/worktrees/` entries, i.e. the
widened glob reaches the one the old `agent-*` glob missed. `233 − 136 = 97`.


ZacxDev:
Merging. Nine rounds — eight fix rounds and a final one-line pin — gated on the **merged tree** every time, never on the branch.

**What ships.** `--exclude-path GLOB` (repeatable) plus `--include-agent-worktrees`, and **the default is now safe**: `*/.claude/worktrees/*` is spared unless you opt out by name. On the live box that moves the removable set from 236 to **99**, sparing **137 dead rows that are other live sessions' working directories**. Excluded rows are still scanned, classified and printed, are never removed, and are not counted toward `--confirm N` — the operator's safety number and the executor's set are the same set.

**The removal path was exercised for real**, not only in fixtures. On a throwaway repo outside `~/workspace`: dry run showed `2 dead / 1 removable / 1 SPARED`; `--confirm 2` refused with both directories intact; `--confirm 1` removed exactly the ordinary worktree and left the agent one with its content intact and still registered with git. Then the control that actually decides it — `--include-agent-worktrees` made that same worktree removable and removed it. Same worktree, same classification, opposite outcome from the flag alone. Without that second half, "the agent worktree survived" would have been indistinguishable from it never having been a candidate.

**Final verification, on `6bedc72e` merged into `main` at `8be3d9c7`:**
```
my independent gate   TOTAL collected=17535 passed=17533 skipped=2 failed=0
                      node suites=5 files=39 tests=1292 pass=1292 fail=0
                      GATE: RESULT=PASS   43 targets PASS / 0 FAIL
GUARD 10              0 "What moved" sections, repo-local-reported-total=0
mutation battery      66/66 ok, 0 red, 0 skipped; tool restored by content
tekton (required)     pytests success, nodetests success
```

**Why nine rounds.** Rounds 2–8 each exist because the *previous round's fix* introduced the next defect — a safety flag that blocked itself and told the operator to drop it; a test that reddened on CPython 3.13; advice pointing at the deleting-more direction; a note contradicting its own run three lines below. None was caught by a green suite: the tree was green at 17,296 tests when the flag-blocks-itself defect shipped.

The root cause was structural, not careless. The predicate "is the agent constant applied by default?" was open-coded at three sites, so fixing it in one place regenerated it in another. Round 6 consolidated it to a single definition with an AST guard; round 7 found that guard missed `ast.If` and **narrowed its own claim** to a tripwire for likely spellings rather than a proof, listing five spellings it measurably cannot see.

**Payload ledger** (`scripts/worktree-prune`, per round): 149 · 251 · 435 · 105 · 69 · 136 · 84 · 33. Every round changed shipped code, so #900's plateau gate never primed. It would have on a round 10 — the final finding's fix was one test assertion, zero payload — which is why this merges now.

**Two findings that outlive this PR:**
- 🔴 **devrc pushes are ungated on this host.** `core.hooksPath` unset at global, system and repo scope; `.git/hooks/` holds only `.sample` files. Tekton and hand-run gates were the only checks on all nine rounds.
- The base clone's `.git/config` had **three separate writers** during this work — concurrent sessions creating and deleting `integ/*` branches mid-gate. GUARD 10 caught each and correctly refused to attribute them.

**Known and accepted:** the structural guard is a tripwire, not a proof — `operator.not_`, a dict lookup, a `match` statement, a one-line local alias and a non-bool `Compare` all walk past it, measured and listed in its docstring. And `agent_glob_applied_by_default` remains reachable as `True` through `summarize()`'s keyword-only argument in a way `main()` never produces.

PR #900, opened 2026-08-27
feat(audit-pr): a round that changes no PRODUCTION code is auditing the LADDER, not the PR

ZacxDev:
## Why

`/audit-pr` got a stop rule yesterday (#861): **a clean round ends the ladder; rounds continue only while the previous round produced a finding that needed fixing.** It is keyed to FINDINGS — and there is a whole regime in which it cannot fire.

A fix round *writes new guards*. The next delta round diffs `<audited-sha>..HEAD`, so **those guards are its audit surface**. The ladder manufactures its own next round's findings.

## Measured — `civitai/cli` #498, the day after #861 landed

Session `4719a2f0` (2026-08-26). The session even quotes the new rule back at round 3 — *"I'll stop at the first clean one, not at a 'safe to merge' verdict"* — and then runs seven more rounds.

| | |
|---|---|
| rounds | **10** (17:50 → 23:22 = **5 h 32 m**) |
| share of session | 1,312 of 1,756 transcript rows (75%) |
| main-thread output tokens | **321k of 419k (77%)**; 100M of 115M cache reads |
| rounds that returned CLEAN | **zero** — the stop rule was never once eligible |

Churn attribution, from the branch's own fix commits:

| rounds | test/guard lines | production lines |
|---|---|---|
| feature + rounds 1–3 | 1,061 | 332 |
| **rounds 4–10** | **1,002** | **0** |

Every one of rounds 4–10 found something real. All of it concerned guards an earlier round of the same ladder had written. At round 9 the session diagnosed its own plateau — *"rounds 3–10 have all been about the guards, not the feature"* — and asked the operator, five hours after the number stopped moving.

**Not a one-off.** Sweeping `~/.claude/projects/**/*.jsonl` for numbered delta re-audit dispatches: **110 sessions, 440 rounds, 84 of them in the preceding 14 days**; mean deepest round 4.0; 34% ran ≥5. Distribution 3→38, 4→22, 5→21, 6→9, 7→4, 8→2, 10→1.

## What this adds

1. **The attribution gate.** Before dispatching round N+1, `git diff --numstat <first-audited-sha>..HEAD` excluding tests. **Two consecutive rounds whose fixes changed zero production lines ⇒ the ladder has left the PR. Stop.** Remaining guard findings become a follow-up task against the test file, not another round. On #498 this fires after round 5.
2. **A scope brief for delta rounds** — label every finding `behaviour` or `guard`; report on scaffolding only where a defect would let a real behaviour regression through. Tests an earlier round wrote are in the diff *by construction*, and an adversarial auditor handed them will audit them, as instructed.
3. **A per-round ledger** in every summary (`round N · production lines changed since round 1: X · elapsed: Y`), so the plateau is visible at round 4 rather than round 9.

## 🔴 It is neither brake this skill already rejects

Both rejections are left standing, and both stay pinned by the test module.

- **Not the round CAP.** It counts production lines, never rounds. A round that touches production code never trips it, however deep the ladder is — so #505's round-4 ReDoS-introduced-by-round-3 still gets caught.
- **Not a retraction of "#804 is NOT an example of a wasted round."** #498 contains no round that ran and found nothing, which is the waste that retraction denies. Different axis: real findings *about scaffolding the ladder itself had just written*.

## Bytes

Case histories (#804's three defects, the `civitai-manager` five-round chain, the renumbering trio, and the full #498 measurement) move to `claude/skills/audit-pr/reference/round-ladder-evidence.md`, routed by its **deployed** path — a devrc skill is read from `~/.claude/skills/` with the cwd somewhere else entirely.

Body **11,103 → 11,909 B**, inside the 12,038 B budget `scripts/skill-audit.py` enforces (`✓ within budget, references intact`).

## Coverage

`scripts/tests/test_audit_ladder_stop_rule.py`: **+3 tests (6 → 9)**, whole-string pins because the artifact is prose.

```
POS  unmutated copy ......................... 9 passed          <- harness control
BASE origin/main SKILL.md, no reference ..... 3 failed, 6 passed <- exactly the three new
     file (i.e. pre-change)                                        tests red, all six old
                                                                   ones green
M7   attribution heading deleted ............ 2 failed
M8   threshold reworded TWO -> THREE ........ 1 failed  (gate pin ONLY — not a keyword guard)
M9   evidence file deleted .................. 1 failed
M10  section moved ABOVE the stop rule,
     text byte-identical .................... 1 failed  (ORDER pin ONLY — reachability control)
M11  different-axis paragraph dropped ....... 1 failed
```

Each on its own scratch tree, `PYTHONDONTWRITEBYTECODE=1`, pytest cache disabled, and every mutant asserts its target string exists before editing (a missed mutant fails loudly instead of scoring SURVIVED).

**What it does NOT enforce** — recorded in the module docstring, not just here:

- Nothing proves the gate is ever **run**. It is a `git diff --numstat` an operator issues between rounds, so the only evidence it fired is a ladder that stopped. That claim needs a future transcript sweep and is not made.
- **The threshold is pinned, not justified.** "Two consecutive rounds" comes from one measured ladder (n=1). One is knowingly too tight; whether two is right is a judgement.

## Gate

- Dev-host: `scripts/gate.sh --tier both` → `GATE: RESULT=PASS exit=0` (pytest + node).
- Sandbox tier — the one Tekton gates on: `nix build .#checks.x86_64-linux.pytests` and `.nodetests` both built.
- Base `6509702b`. `strict` is false on this repo, so this is a claim about the branch; the merged tree is still worth a re-run if `main` moves first.

## Left out of scope

Two 🔴 hazards found in the same review of this skill, deliberately not bundled: the skill mandates `isolation: "worktree"` for multi-PR fan-out (which worktrees the **cwd's** repo, not the PR's) and separately tells the auditor to `cp -a` a copy to mutate (whose `.git` is a *file* — a commit in the copy lands on the real branch). Happy to file or fix separately.

### DIFF
--- a/claude/skills/audit-pr/SKILL.md
+++ b/claude/skills/audit-pr/SKILL.md
@@ -7,6 +7,9 @@ allowed-tools: Bash, Read, Grep, Glob, Agent
 
 # /audit-pr — adversarial PR audit
 
+Every case history and number below is sourced in
+`~/.claude/skills/audit-pr/reference/round-ladder-evidence.md`.
+
 Target: `$ARGUMENTS`. Resolve it:
 - A number → that GitHub PR (`gh pr diff <n>`, `gh pr view <n>`).
 - `current` / empty → the current branch's diff vs its base/trunk.
@@ -16,16 +19,15 @@ Target: `$ARGUMENTS`. Resolve it:
 
 Dispatch a subagent (read-only — it must NOT modify files or merge) to audit the change against this checklist. Have it read the diff and the code it touches, not just the PR description.
 
-**Always run this on high-yield change-classes** — web/HTTP endpoints, concurrency reworks, filesystem/quarantine/trash moves, DB migrations, and anything security/auth/path-gating. These reliably hide real, deploy-blocking bugs (shutdown data-loss, trash-path overwrite, scanner match-path scope-creep, an unauthenticated arbitrary-path scan and a git `core.fsmonitor` RCE all surfaced this way). If the branch has a **private Go module dep**, the auditor may need `GOPRIVATE` (e.g. `GOPRIVATE=github.com/civitai/*`) — a sum-db `500` there is env, not a code defect.
+**Always run this on high-yield change-classes** — web/HTTP endpoints, concurrency reworks, filesystem/quarantine/trash moves, DB migrations, and anything security/auth/path-gating. These reliably hide real, deploy-blocking bugs (shutdown data-loss, trash-path overwrite, scanner scope-creep, an unauthenticated arbitrary-path scan and a git `core.fsmonitor` RCE all surfaced this way). If the branch has a **private Go module dep**, the auditor may need `GOPRIVATE` — a sum-db `500` there is env, not a code defect.
 
-**Brief the auditor on the environment, or it will report false findings.** A fresh worktree is not a
-working checkout, and an auditor that hits this cold reports it as a defect in the PR. Tell it, up
+**Brief the auditor on the environment, or it will report false findings.** A fresh worktree is not
+a working checkout, and an auditor that hits this cold reports it as a defect in the PR. Tell it, up
 front, whichever apply: **git submodules are unpopulated** in a new worktree (one made 4 test files
-"fail to collect" — pure environment); **monorepo/workspace
-`node_modules`** may need linking per package, not just at the root; **whether the base branch is
-already red** on typecheck/lint/tests, and *at which file*, so a baseline error is not attributed to
-the PR; and that **zsh does not word-split unquoted parameters**, so `eslint $FILES` silently checks
-**zero** files and prints a confident PASS. Also tell it to work in a `cp -a` copy if it wants to
+"fail to collect" — pure environment); **monorepo `node_modules`** may need linking per package, not
+just at the root; **whether the base branch is already red** on typecheck/lint/tests and *at which
+file*, so a baseline error is not attributed to the PR; and that **zsh does not word-split unquoted
+parameters**, so `eslint $FILES` silently checks **zero** files and prints a confident PASS. Also tell it to work in a `cp -a` copy if it wants to
 mutate code, leave your worktree untouched, and verify it clean at the end.
 
 **Audit for:**
@@ -58,9 +60,9 @@ Ask the re-auditor to:
   only where the defect would let a real behaviour regression through;
 - treat "the author says it's fixed" as a claim to verify against the diff.
 
-**Carry the ledger in every round's summary**: `round N · production lines changed since round 1: X
-· elapsed: Y`. Without it the flattening is visible only in hindsight — on #498 the session
-diagnosed its own plateau at round 9, six rounds after X stopped moving.
+**Carry the ledger in every round's summary**: `round N · payload lines changed THIS round: X (since
+round 1: Y) · elapsed: Z`. X is what the gate below reads; without it the flattening is visible only
+in hindsight — on #498 the session diagnosed its own plateau at round 9, six rounds late.
 
 ### 🔴 A clean round ENDS the ladder. Never run another round to confirm a clean round.
 
@@ -101,31 +103,39 @@ For a delta re-audit you must name the prior findings (that is the point), so ke
 
 ### 🔴 ATTRIBUTION: a round that changes no PRODUCTION code is auditing the LADDER, not the PR
 
-A fix round writes new guards. The next delta round diffs `<audited-sha>..HEAD`, so **those guards
-are its audit surface** — the ladder manufactures its own next round's findings, and the stop rule
-above, keyed to findings, can never fire. Measured on `civitai/cli` #498 (2026-08-26): **ten rounds,
-5 h 32 m, 77% of the session's output tokens; the fix commits for rounds 4–10 changed 1,002 lines of
-test code and ZERO lines of production code.** No round was ever clean. Full numbers and method:
-`~/.claude/skills/audit-pr/reference/round-ladder-evidence.md`.
+A fix round writes new guards and the next delta round diffs them, so **the ladder manufactures its
+own next round's findings** and the stop rule above, keyed to findings, can never fire. Measured on
+`civitai/cli` #498 (2026-08-26): **ten rounds, 5 h 32 m, 77% of the session's output tokens; rounds
+4–10 changed 1,051 lines of test code and ZERO lines of the thing the PR shipped.** No round was
+ever clean.
 
 So gate on what each round CHANGES, not on what it finds. After a round's fixes land, count the
-production lines **that round** changed: `git diff --numstat <the sha you audited THAT round>..HEAD`.
+payload lines **that round** changed:
 
-🔴 **Per-round, never cumulative.** A range anchored at round 1 stays non-zero forever once any
-early round touched production — on #498 the cumulative form prints the same number for rounds 4
-through 10 and the gate never fires. Equivalently: the ledger's cumulative X **unchanged across two
-rounds** is the same condition, and that is what makes it visible in the summaries.
+```
+git log --numstat --format= --no-merges --first-parent <the sha you audited THAT round>..HEAD
+```
 
-🔴 **Do not classify with a pathspec.** `':!*test*'` swallows `attestation/`, `latest/` and
-`inspector/` as "tests" while missing `FooTest.java` and `*.cy.ts` — wrong in both directions on
-ordinary names (measured). A round's fix touches a handful of files: read the `--numstat` list and
-judge by this repo's convention. **Tests, fixtures and docs are not production.** Ambiguous is not
-zero — the gate does not fire, and the ladder continues.
+🔴 **The unit is THIS PR's PAYLOAD, never a file extension.** Payload = what the PR exists to ship;
+scaffolding = the tests, fixtures and notes a round wrote to guard it. For a code change the payload
+is source and a `.md` is not — but **for a docs or skill PR the payload IS the `.md`**, and 24 of
+devrc's last 40 merged PRs ship no source at all, so a rule keyed to file type reads every round of
+those as zero and stops a ladder that is working. Nor will a pathspec do it: `':!*test*'` swallows
+`attestation/` and `latest/`, `':!*spec*'` swallows `inspector/`, and both keep `FooTest.java` and
+`*.cy.ts` (measured). A round's fix touches a handful of files — read the list and name each one
+payload or scaffolding. **Ambiguous is not zero**: the gate does not fire, and the ladder continues.
 
-**Two consecutive rounds whose fixes changed zero production lines ⇒ the ladder has left the PR.
-Stop.** File the remaining guard findings as one follow-up task naming the test file, closed when
-its PR merges or a named reader dismisses it in writing; do not spend a round on them. A round that
-touches production code never trips this, however deep the ladder is.
+🔴 **Per-round and the round's OWN commits.** Anchored at round 1 the count stays non-zero forever
+once an early round touched payload — on #498 that prints the same number for rounds 4 through 10
+and never fires. A plain two-dot `git diff` breaks the other way: after a `merge main` or a rebase
+it sweeps upstream lines in as payload, so a scaffolding-only round scores non-zero and the counter
+resets. Hence `--first-parent --no-merges` above; if the branch was rebased mid-round, re-anchor on
+the new sha rather than trusting the old one.
+
+**Two consecutive rounds whose fixes changed zero payload lines ⇒ the ladder has left the PR.
+Stop.** File the remaining scaffolding findings as one follow-up task naming the file, closed when
+its PR merges or a named reader dismisses it in writing. A round that touches payload never trips
+this, however deep the ladder is.
 
 ⚠ **This does not retract the two rules above, and is not a cap in disguise.** #498's rounds were
 not wasted in the sense those rules deny — every one found something real. The waste is on a
@@ -137,26 +147,19 @@ measures the fixes; it never counts the rounds.
 When a PR claims a guard is "mutation-verified", check **what kind**. Deleting the guard is the
 obvious mutant and the weakest: four variants that delete NOTHING — swapped operands, inverted
 branches, the guard commented out, a stale value re-bound — once passed a suite its author had just
-"mutation-verified" (each one, and how to enumerate your own, in the reference file). The two that
-decide most cases:
-
-- **When you can only assert on TEXT** (raw SQL, a generated config, a serialised query), **pin the
-  WHOLE normalised statement**, not features of it — a partial regex is satisfied by semantically
-  inverted code, and `--` / `/* */` make "the token is present" and "the clause is live" different
-  facts. A cosmetic reformat then fails the test; that is the trade.
-- **A fixture of empty or default values collapses distinct implementations into identical output.**
-  One mutant survived *only* because the fixture was `{}`. Give fixtures non-default sibling values.
-
-**A review fix RESETS the gate** (`claude/RULES.md`): re-run the FULL battery after each fix round
-*and* after any reformat, not just the mutant for what you changed.
+"mutation-verified". The two rules that decide most cases: **when you can only assert on TEXT, pin
+the WHOLE normalised statement** (a partial regex is satisfied by inverted code, and `--` / `/* */`
+make "the token is present" and "the clause is live" different facts — a cosmetic reformat then
+fails the test, which is the trade); and **a fixture of empty or default values collapses distinct
+implementations into identical output**, so give fixtures non-default sibling values. **A review fix
+RESETS the gate** (`claude/RULES.md`): re-run the FULL battery after every round and reformat.
 
 ## Price a defect from the CONSUMING code, not the producing site
 
-Before repeating any consequence an audit asserts — especially one with a cost attached — read the
-code that **consumes** the value, not just the code that writes it. **Verifying that a value is USED
-is not verifying what its ABSENCE costs**: one audit priced a lost watermark at "re-judges the whole
-backlog at LLM cost" when a `NOT EXISTS` dedupe in the same `WHERE` clause made it zero LLM calls.
-Sanity-check the frequency side too — "routine" and "rare" are asserted far more often than measured.
+**Verifying that a value is USED is not verifying what its ABSENCE costs.** Read the code that
+consumes it before repeating any consequence an audit asserts, especially a costed one (worked
+example in the reference file). Sanity-check the frequency side too — "routine" and "rare" are
+asserted far more often than measured.
 
 **A finding about the PR *description* gets corrected PUBLICLY.** If the audit shows the PR body
 misstates what the change does, post a **PR comment** saying so rather than silently editing the
--- a/claude/skills/audit-pr/SKILL.md
+++ b/claude/skills/audit-pr/SKILL.md
@@ -14,12 +14,9 @@ Target: `$ARGUMENTS`. Resolve it:
 
 ## What to do
 
-Dispatch a subagent (read-only — it must NOT modify files or merge) to audit the change against this checklist. Have it actually read the diff and the surrounding code it touches, not just the PR description.
+Dispatch a subagent (read-only — it must NOT modify files or merge) to audit the change against this checklist. Have it read the diff and the code it touches, not just the PR description.
 
-**Always run this on high-yield change-classes** — web/HTTP endpoints, concurrency reworks, filesystem/quarantine/trash moves, DB migrations, and anything security/auth/path-gating. These reliably hide real, deploy-blocking bugs (shutdown data-loss, trash-path overwrite, an unauthenticated arbitrary-path scan and a git `core.fsmonitor` RCE all surfaced this way). If the branch has a **private Go module dep**, the auditor may need `GOPRIVATE` (e.g. `GOPRIVATE=github.com/civitai/*`) — a sum-db `500` there is env, not a code defect.
-
-Case histories: `~/.claude/skills/audit-pr/reference/round-ladder-evidence.md` (source
-`devrc/claude/skills/audit-pr/`).
+**Always run this on high-yield change-classes** — web/HTTP endpoints, concurrency reworks, filesystem/quarantine/trash moves, DB migrations, and anything security/auth/path-gating. These reliably hide real, deploy-blocking bugs (shutdown data-loss, trash-path overwrite, scanner match-path scope-creep, an unauthenticated arbitrary-path scan and a git `core.fsmonitor` RCE all surfaced this way). If the branch has a **private Go module dep**, the auditor may need `GOPRIVATE` (e.g. `GOPRIVATE=github.com/civitai/*`) — a sum-db `500` there is env, not a code defect.
 
 **Brief the auditor on the environment, or it will report false findings.** A fresh worktree is not a
 working checkout, and an auditor that hits this cold reports it as a defect in the PR. Tell it, up
@@ -63,7 +60,7 @@ Ask the re-auditor to:
 
 **Carry the ledger in every round's summary**: `round N · production lines changed since round 1: X
 · elapsed: Y`. Without it the flattening is visible only in hindsight — on #498 the session
-diagnosed its own plateau at round 9, six rounds after that number stopped moving.
+diagnosed its own plateau at round 9, six rounds after X stopped moving.
 
 ### 🔴 A clean round ENDS the ladder. Never run another round to confirm a clean round.
 
@@ -72,10 +69,9 @@ round that returns no findings is the last one — stop there, and do not re-con
 not on the author saying it's done.
 
 🔴 **A "safe to merge" VERDICT is not the stop signal — the FINDINGS are.** #804's rounds **5, 6 and
-7 each returned "safe to merge" and each still reported real defects** that were then fixed — among
-them a `didDrag` latch pinned on its SET but not its RELEASE, where deleting the reset, deleting the
-clear, and deleting **both** were all green. A ladder keyed to the verdict stops at round 5 and
-ships that latch.
+7 each returned "safe to merge" and each still reported real defects** that were then fixed — the
+last a latch that read as pinned and was vacuous in both directions (all three in the reference
+file). A ladder keyed to the verdict stops at round 5 and ships it.
 
 ⚠ **#804 is NOT an example of a wasted round, and neither is any other PR cited here.** Every one of
 its eight rounds produced findings that needed fixing, round 8 included. This is a forward rule with
@@ -90,26 +86,6 @@ detector exists to prevent, reintroduced by the fix for the previous one"*. A ca
 both. Keep going while rounds keep finding things — `claude/RULES.md` still says to budget for
 several — and stop the moment one does not.
 
-### 🔴 ATTRIBUTION: a round that changes no PRODUCTION code is auditing the LADDER, not the PR
-
-A fix round writes new guards. The next delta round diffs `<audited-sha>..HEAD`, so **those guards
-are its audit surface** — the ladder manufactures its own next round's findings, and the stop rule
-above, keyed to findings, can never fire. Measured on `civitai/cli` #498 (2026-08-26): **ten
-rounds, 5 h 32 m, 77% of the session's output tokens; the fix commits for rounds 4–10 changed 1,002
-lines of test code and ZERO lines of production code.** No round was ever clean.
-
-So gate on what the rounds CHANGE, not on what they find. Before dispatching round N+1, run
-`git diff --numstat <first-audited-sha>..HEAD -- ':!*_test.*' ':!*test*' ':!*spec*'`.
-
-**Two consecutive rounds whose fixes changed zero production lines ⇒ the ladder has left the PR.
-Stop.** File the remaining guard findings as a follow-up task against the test file; do not spend a
-round on them. A round that touches production code never trips this, however deep the ladder is.
-
-⚠ **This does not retract the two rules above, and is not a cap in disguise.** #498's rounds were
-not wasted in the sense those rules deny — every one found something real. The waste is on a
-different axis: real findings *about scaffolding the ladder itself had just written*. The gate
-measures the fixes; it never counts the rounds.
-
 **When a round's fix is mostly renumbering your own prose, fix the FORM, not the number.** Number
 the list and tell the reader to count it; a total maintained in parallel with the thing it counts
 will drift.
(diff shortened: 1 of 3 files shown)


ZacxDev:
Merged-tree gate (the repo's own 🔴 — `strict` is false here, so the branch check is not a claim about the tree the merge creates):

`main` moved 3 commits since the branch point (`8a689a10`), and two of them touch `claude/RULES.md` and `claude/RULES-ARCHIVE.md` — which this PR's test module pins whole strings in, so this was worth doing rather than assuming.

Built an integration branch off `8a689a10`, merged `ed38490b` (clean, no textual conflict), and ran `scripts/gate.sh --tier pytest` there: `GATE: RESULT=PASS exit=0`. `test_audit_ladder_stop_rule.py` 9 passed on the merged tree — no semantic conflict with the RULES.md edits that landed.

`gh pr view 900 --json mergeable,mergeStateStatus` → `MERGEABLE` / `BLOCKED` (required Tekton checks still pending).

ZacxDev:
## Round 1 audit → both 🔴 fixed

Blind adversarial audit (read-only subagent, given the diff and the checklist, not my conclusions). It returned **merge after fixing 🔴 1 and 🔴 2** — and both were in the payload, not the scaffolding, and neither was caught by the tests this PR added.

**🔴 1 — the gate never fired on the ladder that motivated it.** The instruction said `git diff --numstat <first-audited-sha>..HEAD`: cumulative since round 1, while the condition it feeds is a *per-round* delta. Once any early round touches production the total is non-zero forever. On #498 the cumulative form prints the same number for rounds 4 through 10, so an operator following the text literally never stops — the rule failed **open**, into the exact 5h32m waste it exists to prevent.

Now `git diff --numstat <the sha you audited THAT round>..HEAD`, with per-round stated as its own 🔴 and the equivalent ledger formulation (cumulative X unchanged across two rounds) named beside it.

**🔴 2 — the pathspec misclassified ordinary filenames in both directions.** Measured on a scratch repo, `git ls-files -- ':!*_test.*' ':!*test*' ':!*spec*'` returns as "production" only:

```
cypress/e2e/login.cy.ts       <- a test
src/main/java/FooTest.java    <- a test (pathspec is case-sensitive)
main.go
```

…while silently dropping `pkg/attestation/verify.go`, `api/latest/handler.go` and `internal/inspector/scan.go` — production code scored as tests. A repo under `attestation/` would stop its ladder after two rounds with real regressions still landing; a Java/Cypress repo counts test churn as production and the gate is inert.

Fixed by not pattern-matching at all: a round's fix touches a handful of files, so read the `--numstat` list and judge by the repo's own convention. Tests, fixtures **and docs** are not production. Ambiguous is not zero — the gate does not fire and the ladder continues, so the residual error biases safe.

## Also fixed from the same audit

- **The churn table was wrong twice, both flattering the argument**: test column mis-added (**961**, not 1,061) and the production column folded 110 lines of release-notes markdown into "production" (**222**, not 332). It now has three columns, states its classifier explicitly, and carries the correction — "is a doc production?" being the same question the gate turns on. Last production change restated: 18 lines of `internal/appapi/appblocks.go`.
- **The module's own matrix said "5 passed" for a module that shipped with six tests** — wrong when written (in #861), and this PR had added a contradicting number instead of correcting it. Re-measured at `6509702b`: 6 passed. Corrected in place.
- Evidence file had **no truncation floor** (emptying it produced a "you reworded it" message about an empty file) — added at 1,000 B.
- The `### ATTRIBUTION` heading had **re-parented** three paragraphs belonging to the clean-round section; it now sits after them.
- The evidence route was pinned exactly-once, which made a second pointer a test failure — now a presence assertion.
- Follow-up tasks carry a closing condition; the reference file routes by deployed paths.

## Coverage now pins the MECHANISM, not just the decision

Three mutants **survived** a fully green 9-test run in review:

| mutant | before | after |
|---|---|---|
| X3 gate COMMAND replaced with `echo 0` | 9 passed | **1 failed** |
| X4 ledger instruction deleted | 9 passed | **1 failed** |
| X5 `behaviour`/`guard` labelling bullet deleted | 9 passed | **1 failed** |

**+5 tests (6 → 11).** Full matrix, each on its own scratch tree, `PYTHONDONTWRITEBYTECODE=1`, cache disabled, every mutant asserting its target exists before editing:

```
POS  unmutated ............................... 11 passed
BASE origin/main, no reference file .......... 6 failed, 5 passed
M7   attribution heading deleted ............. 2 failed
M8   decision reworded TWO -> THREE .......... 1 failed  (decision pin only)
M9   evidence file deleted ................... 2 failed
M10  section moved ABOVE the stop rule,
     text byte-identical ..................... 1 failed  (ORDER pin only)
M11  different-axis paragraph dropped ........ 1 failed
X6   evidence truncated to 0 B ............... 2 failed  (floor guard)
X7   no-pathspec rule deleted ................ 1 failed
X8   range reverted to CUMULATIVE ............ 1 failed  <- the shipped defect
```

Body 11,909 → 12,032 B, inside the 12,038 B budget — paid for by demoting the four mutation variants and #804's defect list to the reference file.

## Ledger (dogfooding the rule this PR adds)

`round 1 · production lines changed since round 0: the whole payload · elapsed: ~21 min`

This round changed the deliverable itself — both 🔴 were in the skill prose, which *is* the production artifact here — so the attribution gate does not fire and the ladder continues if the next round finds something. Round 2 will be a delta re-audit of `ed38490b..HEAD`, blind to this comment's conclusions per the skill's own framed-audit rule.

Not adopted from the audit: the two disclosed 🔴 about `isolation: "worktree"` and `cp -a` (still scoped to a follow-up PR), and the `RULES.md` "not a cap" wording, which the skill reconciles explicitly rather than editing the always-on file.

ZacxDev:
## Round 2 (delta re-audit of `ed38490b..5f39ffde`) → fixed in `8cce25b5`

Not clean: one 🔴, five 🟡, several 🟢. The headline is that **round 1's fix shipped the mirror image of the bug it fixed** — round 1 fixed a gate that could never fire; its replacement fires when it must not.

**🔴 Keyed on file type, the gate stops a working ladder.** "Tests, fixtures and docs are not production" reads **zero payload for every round of a docs or skill PR** — measured at **24 of the last 40 merged devrc PRs (60%)**, this one included. Round 1's own fix touched a skill body, a reference doc and a test file: it would have scored 0, and two such rounds would have stopped the ladder *in the middle of finding this 🔴*. That is the fail-**closed** direction the previous commit message named as the worse error — and the pathspec it replaced never had it, because it never excluded `.md`.

The unit is now **this PR's payload, never a file extension**: payload = what the PR exists to ship; scaffolding = the tests, fixtures and notes a round wrote to guard it. Code PR → source, and a `.md` is not. Docs/skill PR → the `.md` *is* the payload.

**🔴 The range still re-attributed upstream work** (reported by round 1 as 🟡 3, and I neither fixed nor disclosed it — fair hit). `<sha>..HEAD` sweeps in whatever a `merge main` or a rebase brought along. Measured on a scratch repo, a round whose only fix was one test line:

```
git diff --numstat <sha>..HEAD                                    →   1 + 200 lines
git log --numstat --format= --no-merges --first-parent <sha>..HEAD →   1 line
```

201 vs 1, same repo, same state. The gate would not fire and the counter would reset — fails **open**, the same direction as round 1's 🔴.

## The rest

- **The churn table was wrong a third time.** `1,002` is rounds 4–**9**; round 10's fix `7541bc1` (49 test lines) landed after my snapshot. It is **1,051**. The table now carries its shas and the correction states the lesson instead of just the number: a figure re-derived from a moving branch is a measurement with a timestamp.
- **`inspector/` is dropped by `':!*spec*'`, not `':!*test*'`** — measured. My 🔴-flagged rule failed on its own third example.
- **The classifier was pinned as a prohibition, not a method.** Five mutants passed the green 11-test suite, including one that inverted "Ambiguous is not zero" to its opposite and one that reinstated the pathspec as the method with the warning left intact. Three constants now pin the definition, the method and the counter-evidence.
- The corrected `961 / 110 / 222` row was itself unpinned (rewriting it to 9999s was green). The evidence floor of 1,000 B sat *below* the first string it guards, so a 1,400 B truncation passed it and produced the exact misleading message the floor exists to prevent — now 1,500. The inherited matrix still said `POS 5 passed` three lines under the paragraph declaring 5 impossible; re-measured: POS 11, M2 2, M5 9.

## Coverage: 17 mutants, all dead

```
POS  unmutated ............................. 11 passed
BASE origin/main ........................... 6 failed, 5 passed
PREV round-1 tip ed38490b .................. 4 failed, 7 passed
Z1/Z3/Z4/Y9/Z8  the five that survived ..... 1 failed each
X8/Y4/Z9  earlier rounds' regressions ...... 1, 2, 1 failed
X3/X4/X5/M8/X6/M10  the round-1 set ........ 1, 1, 1, 1, 2, 1 failed
```

🔴 **Two mutants silently failed to apply on the first pass** (their targets had been re-wrapped) and would have been scored SURVIVED off an 11-passed run. The assert-before-edit in the harness caught it. Now stated in the docstring, because it is the same "validate the instrument" failure the rules warn about.

Body 12,032 → 12,026 B. Gate `--tier both` PASS.

## Ledger

`round 2 · payload lines changed THIS round: the whole rule (both 🔴 were in the skill body) · since round 1: same · elapsed: ~2 h 10 m`

The gate does not fire — this round changed the payload — so the ladder continues. Round 3 is dispatched, blind to this comment.

ZacxDev:
## Round 3 (delta re-audit of `5f39ffde..8cce25b5`) → fixed in `f496cc15`

Not clean: six 🟡, six 🟢, no 🔴 — and eight of them were defects round 2's own edits introduced or left behind.

**The heading.** Round 2 converted the rule from file-type to payload everywhere except the sentence a skimming reader hits first:

```
### 🔴 ATTRIBUTION: a round that changes no PRODUCTION code is ...
```

…fifteen lines above a paragraph stating the unit is "never a file extension". And `SKILL_ATTRIBUTION_HEADING` **pinned it** — so *fixing* the heading turned two tests red. The guard was actively holding the retracted rule in place. A pin is only as good as the sentence it points at.

**`--no-merges` has a cost, and I hadn't said so.** It is what stops a `merge main` being counted as this round's payload — and it also makes payload hand-written into a merge *conflict resolution* invisible: measured **0**, where the two-dot form read **340**. That is the unsafe direction: the gate reads zero on a round that did change payload. Now stated in the rule (count that merge explicitly with `git show --numstat <merge>`) and pinned — rewriting "has one cost" → "has no cost" survived a green 11-test suite.

**The number I took on trust.** "24 of devrc's last 40 merged PRs ship no source" came from the round-2 auditor's report and went into the skill without re-derivation — precisely the mistake the churn table beside it records three times. Re-measured myself: **25 of 40**, 17 docs-only. Now dated in the sentence, pinned, with the classifier and method in the reference file. Before that, replacing it with "many devrc PRs…" was green.

**The floor, properly.** Round 2 raised the evidence floor to 1,500 B claiming the first pinned string starts at ~1,400. Measured: that string spans **1,433–1,501** and the *last* pinned string ends at **3,532** — so 1,500 still passes a truncation that ate the first one, and every floor under 3,532 leaves a hole. Chasing the number is the wrong fix: a floor can never sit above a string not yet written. `_assert_pinned_once` now reports the **file's size** in its own failure message, which separates "truncated" from "reworded" at any size; the floor is demoted to a stub check and says so.

Also: the preamble's "every case history and number below is sourced in the reference file" was false for five of them (including the 24-of-40 figure added in the same commit); the ledger pin's justification still described the format round 2 replaced; "five mutants that rewrote the method" was four; "201 vs 1" named no repo, sha or range and didn't reproduce — it now names the scratch-repo scenario; three "production" strings survived in module comments and two guard failure messages.

## Coverage: 21 mutants, all dead

```
POS  unmutated ............................. 11 passed
BASE origin/main ........................... 6 failed, 5 passed
PREV round-1 tip ed38490b .................. 5 failed, 6 passed
W1   heading reverted to "no PRODUCTION" ... 2 failed
W2   dated measurement -> vague "many" ..... 1 failed
W3   --no-merges cost disclaimed ........... 1 failed   <- survived until pinned
W4   evidence truncated to the 1,500 floor . 1 failed   <- floor passes; the pin's
                                                          message names the size
Z1/Z3/Z4/Y9/Z8 ............................. 1 failed each
X8/Y4/Z9 ................................... 1, 2, 1 failed
X3/X4/X5/M8/X6/M10 ......................... 1, 1, 1, 1, 2, 1 failed
```

Body 12,026 → 12,035 B (budget 12,038). Gate `--tier both` PASS.

## Ledger

`round 3 · payload lines changed THIS round: the heading, the --no-merges cost, the measurement and the preamble — all skill body · since round 1: every round · elapsed: ~1 h 40 m`

Gate does not fire — payload changed again. Round 4 dispatched, blind.

Worth noting against this PR's own thesis: three rounds in, **every round has changed payload**, so the attribution gate has correctly stayed silent while the findings kept being real. The rule is not yet the thing ending this ladder; the findings are, exactly as the clean-round rule says.

ZacxDev:
## Round 4 (delta re-audit of `8cce25b5..f496cc15`) → fixed in `c09a087c`

Four 🟡, two 🟢. Both of the ones a reader acts on are in the counting command this PR exists to prescribe.

### The range form was wrong on two of four shapes

Measured, four throwaway repos, git 2.55.0, `additions + deletions` over numstat:

| shape | truth | two-dot `git diff` | shipped `--no-merges --first-parent` | `--remerge-diff … --not <base>` |
|---|---|---|---|---|
| A. clean `merge main` (200 upstream lines); round's fix = 1 test line | 0 payload | **201** | 1 | 1 |
| B. payload hand-written into a merge-CONFLICT resolution | >0 | 30 | **1** | 37 |
| C. round's fix = 50 payload lines on a side branch, merged `--no-ff` | ~50 | 51 | **0** | 51 |
| D. control: 12 payload lines, linear | 13 | 13 | 13 | 13 |

Round 3 disclosed one cost (shape B). **Shape C is the one that matters**: a fix committed on a side branch and merged `--no-ff` reads **zero** — the shape agent worktree isolation produces — so the gate fires and stops a ladder whose payload is still moving. And the remedy round 3 offered for B, `git show --numstat <merge>`, isn't one: it prints the *first-parent* diff, so on shape A it reports all 200 upstream lines as this round's work.

The command is now `git log --numstat --format= --remerge-diff <sha>..HEAD --not <base>` — correct on all four, with D as the positive control.

### The number that moved while the PR was open

Round 3 replaced a `24` (from a subagent, never re-derived) with a `25` I measured, dated and pinned. Re-derived at round 4: **24 again, on both selections** — PRs merged in between, some of them this PR's siblings. The auditor also showed the prose's population descriptor didn't select what was measured (`gh pr list --limit 40` sorts by *created*), and "the rest docs plus tests" was false.

So the body no longer carries the number: it states the claim that doesn't decay — *most of this repo's merged PRs ship no source file at all* — and the reference file carries both selections, both counts, the exact commands and the date. **A number that changes with the clock does not belong in a rule.**

### Also

- The constants' own comments still carried the retracted `24` and the "FIVE mutants" conflation the docstring had already corrected — parallel-number drift inside the module about parallel-number drift, twice.
- The byte offsets round 3 quoted (1,433–1,501 / 3,532) were the *previous* commit's; at HEAD they're 1,417–1,485 / 3,512, and the claim they supported was false at the tip. Removed rather than re-derived — every pin added later moves that boundary, so the number goes stale in a round.
- `_read` now decodes with `errors="replace"`: a truncation landing mid-character raised `UnicodeDecodeError` before any assert ran, so the reader got a traceback instead of the truncation message — which is also why last round's W4 mutant never reached the message it was cited as proving.

### Coverage: 22 mutants, all dead

`POS 11 · BASE 6/5 · PREV 5/6 · W1 2 · W2 1 · W3 1 · W4 1 · W5 1 · Z1/Z3/Z4/Y9/Z8 1 each · X8/Y4/Z9 1/2/1 · X3/X4/X5/M8/X6/M10 1/1/1/1/2/1`

W5 is new: a truncation splitting a multibyte character now prints the crafted message rather than a decode traceback. X3 and X8 assert the **new** command string, so reverting to either earlier form is caught.

Body 12,035 → 12,025 B (budget 12,038); reference 8,679 → 10,752 B. Gate `--tier both` PASS.

### Ledger

`round 4 · payload lines changed THIS round: 81 (skill 47, reference 34) · since round 1: ~380 · elapsed: ~2 h`

Gate does not fire. Round 5 dispatched, blind.

ZacxDev:
## Round 5 (delta re-audit of `f496cc15..c09a087c`) → fixed in `7090ac43`

Four 🟡, one 🟢. The reader-facing one: `<base>` — the operand round 4 added — was never defined.

**A base ref one commit stale silently restores the defect round 4 fixed.** Measured on the shape-A fixture: `--not <current tip>` reads **1**; `--not <a base one commit behind what the branch merged>` reads **201** — the whole upstream bring-in, attributed to this round, gate never fires, ladder runs forever. Fetch-behind base refs are routine in agent worktrees. The rule now says what `<base>` is: the *current* tip you would merge into, never the fork point.

**And every way the command can fail sums to zero — which is the gate's own stop verdict.**

```
--not <ref that does not exist>        -> rc 128, EMPTY stdout   -> sums to 0
--remerge-diff, unwritable object store -> under-counts, EXITS 0 -> 20, not 54
git < 2.35 (no --remerge-diff)          -> rc 128, empty stdout  -> sums to 0
```

A broken command is indistinguishable from a finished ladder. Added: **a failed command is not zero** — confirm it printed something before believing one. That is this repo's own positive-control rule, and the gate shipped without it for two rounds.

**The evidence file was pinned by nothing — while the body delegated *into* it.** Round 4 moved the payload measurement out of SKILL.md ("the reference file dates it") and put the four-shape range table there. Measured: rewriting shape C's `0` to `51` — deleting the entire argument for the current command — left the suite **green**, as did rewriting the PR-population count to `2 of 40`. Both rows are now pinned. That is `Z8`'s own lesson reopened one file over.

Also: the `SKILL_NO_MERGES_COST` → `SKILL_WRONG_FLAGS_TRAP` rename orphaned its 11-line comment, which still described the deleted sentence, still quoted the retracted `25/40` and `340`, and still credited `--no-merges` with the job `--not <base>` does — a maintainer following it would have edited the gate back toward the shape-C zero, suite green. Shape B's row couldn't demonstrate its own bold (truth `>0`, value marked wrong `1`, which satisfies `>0`); it now shows what the branch-side commit contributed, and both new tables carry their fixture recipe. The docstring said "three pins" where the test makes **seven** `_assert_pinned_once` calls.

### Coverage: 26 mutants, all dead

`POS 11 · BASE 6/5 · PREV 5/6 · V1 1 · V2 1 · V3 1 · V4 1 · W1 2 · W2 1 · W3 1 · W4 1 · W5 1 · Z1/Z3/Z4/Y9/Z8 1 each · X8/Y4/Z9 1/2/1 · X3/X4/X5/M8/X6/M10 1/1/1/1/2/1`

V1–V4 are the four rules this round pinned — **every one was green before**. Two mutants again failed to *apply* (re-wrapped targets) and were caught by assert-before-edit rather than scored SURVIVED; third time that harness rule has earned itself here.

Body 12,025 → 12,034 B (budget 12,038); reference 10,752 → 11,592 B. Gate `--tier both` PASS.

### Ledger

`round 5 · payload lines changed THIS round: ~60 (skill 26, reference 34) · since round 1: ~440 · elapsed: ~1 h`

Gate does not fire. Round 6 dispatched, blind.

ZacxDev:
## Round 6 (delta re-audit of `c09a087c..7090ac43`) → fixed in `3b6e6a85`, plus a self-caught miss in `8062c8c3`

Six findings, none 🔴. Two are reader-facing, and both are round 5's fix one level down.

**`<base>` was defined as the ref that is stale by default.** Round 5 fixed "`<base>` is undefined" with "`<base>` is `origin/main`" — which is exactly as current as your last `git fetch`, and round 5's own commit body says fetch-behind base refs are routine in agent worktrees. A reader obeying the sentence to the letter is in the state the rule exists to prevent. Now: **`git fetch` it first.** Also corrected: "one commit stale ⇒ the whole bring-in" was a property of the first fixture — measured on a bring-in spread over 4 commits, one behind re-reports its *tail* (51); the **fork point** re-reports all of it (201).

**"Confirm it printed something" cannot see the failure it was written for.** Measured with `chmod -R a-w .git/objects`:

| case | rc | stdout | sum | caught? |
|---|---|---|---|---|
| missing ref | 128 | empty | 0 | ✅ |
| git without `--remerge-diff` | 128 | empty | 0 | ✅ |
| **unwritable object store, mixed round** | **0** | **non-empty** | **2** (truth 41) | ❌ |
| unwritable object store, all in the resolution | 0 | empty | 0 | ✅ |

That row announces itself only as `error: unable to create temporary object directory` on **stderr** — a word the skill never used. The reader gets a plausible number, the rule never engages ("before believing a zero"), and the ledger is wrong by 95%. Now: **rc 0 AND silent stderr**.

**The evidence file is pinned row by row — four more rows were loose.** Round 5 pinned two of the three green mutants it had named, then wrote "Both rows are now pinned". Still green before this commit: blanking shape A's row; **inverting** it so the table argues *for* the flag the pinned body sentence prohibits; and negating either lesson bullet. Shape A's row and both lessons are pinned now, and the module's DOES-NOT-ENFORCE list gains the honest disclosure that this file is guarded row by row.

**The pin count, wrong three times.** "Three pins" was corrected to "SEVEN" in the same commit that made it **nine** — computed against the pre-fix tree. That is this PR's own rule (*a total kept beside what it counts drifts*) violated by the fix for a violation of it. The numeral is gone: *one pin per constant below — count them*.

### And one I caught on myself

`3b6e6a85`'s message ends `Body 12,034 -> 12,039 B (budget 12,038)` — one byte **over** the budget it claims to be inside. The number was measured correctly; the *comparison* was asserted from memory of having trimmed, one trim short, and nothing enforces a ceiling for this skill. `8062c8c3` fixes it (12,033 B, `✓ within budget`) and records the lesson: **a number can be right and the claim about it wrong** — read skill-audit's verdict line, not the byte count beside it.

### Coverage: 30 mutants, all dead

`POS 11 · BASE 6/5 · PREV 5/6 · U1/U1b/U6/U7 1 each · V1–V4 1 each · W1 2 · W2/W3/W4/W5 1 each · Z1/Z3/Z4/Y9/Z8 1 each · X8/Y4/Z9 1/2/1 · X3/X4/X5/M8/X6/M10 1/1/1/1/2/1`

Two mutants failed to *apply* again (re-wrapped targets), caught by assert-before-edit rather than scored SURVIVED — fourth time in this PR.

Gate `--tier both` PASS.

### Ledger

`round 6 · payload lines changed THIS round: ~40 (skill) · since round 1: ~480 · elapsed: ~1 h 20 m`

Gate does not fire. Round 7 dispatched, blind.

ZacxDev:
## Round 7 (delta re-audit of `7090ac43..8062c8c3`) → fixed in `b674a7f4`

One 🟡, three 🟢 — and the 🟡 is a regression *this PR's own fix round* introduced, which is the category the skill tells re-auditors to hunt.

**The rewrite widened the zero-check on one axis and narrowed it on another.** Round 5 said *"confirm it printed something before believing a zero."* Round 6 replaced it with *"require rc 0 AND silent stderr"* — better for the unwritable-object-store case, and `grep -n "printed"` then returned nothing: the old check was gone, not restated.

| case | rc | stdout | stderr | old rule | new rule |
|---|---|---|---|---|---|
| audited-sha == HEAD (fixes not in this checkout yet) | 0 | empty | silent | ✅ | **miss** |
| `<base>` already contains the round's commits | 0 | empty | silent | ✅ | **miss** |

The fixer commits in worktree A; the auditor runs the gate in a checkout that hasn't picked those commits up. rc 0, nothing printed, nothing on stderr, sum 0 — indistinguishable from *"this round changed zero payload lines"*, which is the gate's own **firing** condition. That counts a strike toward "two consecutive rounds ⇒ stop".

Now all three: **rc 0, silent stderr, and a non-empty range** — plus keep stderr on the terminal, since `2>&1 | awk` folds the one loud failure back into the sum. And the delta-re-audit checklist gains the shape that produced it: *a rule reworded wider on one axis and narrower on another* is now named as a regression to hunt.

**`EVIDENCE_LESSON_BC` was a partial pin.** It stopped at "merged `--no-ff`" — keep that prefix byte-identical, rewrite the tail to *"a shape that essentially never occurs, so prefer it"*, suite green. Same inversion `U1b` exists to prevent, one clause later in the same bullet. Pinned whole (T2).

Also: round 6's orphaned-comment fix **duplicated** the comment instead of moving it, so the original still sat above the wrong constant reading as its rationale; and the new DOES-NOT-ENFORCE item said "six of its numbers" where five numeric rows are pinned — a hand-maintained count added in the same commit that deleted one 740 lines below for exactly that reason.

### The real signal, which is not a finding

Every round now pays for new rules by shaving adjectives, and this round shaved **"rollback"** out of the Gaps checklist item — a genuine audit dimension lost to byte pressure. Restored, and paid for properly by demoting duplicated detail to the reference file (body 12,033 → **11,953 B**). The structural fix is re-homing whole sections, not more shaving; filing that rather than doing it mid-ladder.

### Coverage: 32 mutants, all dead

```
POS 11 · BASE 6/5 · PREV 5/6
T1  empty-range clause deleted ...... 1 failed   <- the check this round restored
T2  B/C lesson TAIL inverted ........ 1 failed   <- green until pinned whole
U1/U1b/U6/U7 · V1–V4 · W1 2 · W2/W3/W4/W5 · Z1/Z3/Z4/Y9/Z8 · X8/Y4/Z9 · X3/X4/X5/M8/X6/M10
```

Three mutants failed to *apply* on the first pass (re-wrapped targets), caught by assert-before-edit rather than scored SURVIVED — fifth time in this PR. Gate `--tier both` PASS.

### Ledger

`round 7 · payload lines changed THIS round: ~35 (skill 27, reference 8) · since round 1: ~515 · elapsed: ~1 h 10 m`

Gate does not fire. Round 8 dispatched, blind.

ZacxDev:
## Round 8 (delta re-audit of `8062c8c3..b674a7f4`) → fixed in `907d6997`

One 🟡, five 🟢 — and the 🟡 is the worst regression of this PR: **a commit whose message narrates the opposite of its own diff.**

### The claim was false, and I wrote it

`b674a7f4`'s body says *"this round shaved 'rollback' out of the Gaps checklist item… **Restored**"*. Measured: `rollback` is present at the PR base and at every commit through `8062c8c3`; `git grep -i rollback claude/skills/audit-pr/` at `b674a7f4` returns **nothing**. That commit is the one that removed it.

The mechanism, because it's the lesson: the restore was one substitution in a multi-edit script whose *later* assertion failed. The script asserts each target before editing and writes once at the end — so the failure discarded every edit including the restore, and the commit message was written from intent rather than re-read from the diff. A one-line `grep` after the edit would have caught it. So would pinning the checklist. Neither existed.

### 🔴 The nine-item checklist was unpinned through eight rounds

It is the operational core of the skill, and **every round of this PR shaved words out of exactly that region** to stay under the byte budget. Measured: replacing the whole block with `**Audit for:** whatever seems off.` left the suite **green**. Pinned now as one whole string — S1 kills the rollback shave, S2 the deletion — so the next shave fails in the commit that makes it.

Two more rules, green until now, both shipped by the round that introduced them:

- **the stderr-capture rule** — deleting it was green, and *inverting* it to "fold stderr into the sum with `2>&1` so no failure escapes the count" was green too. That reads plausible and defeats the silent-stderr third of the rule pinned right beside it.
- **"the rule reworded wider on one axis and narrower on another"** — the regression shape this PR produced twice, added to the checklist last round and unpinned.

### Also

- *"Have it mutate only in a `cp -a` copy and leave your worktree clean"* had changed subject: an instruction that told **you** to verify became a request that the **auditee** be tidy. Restored — an agent's self-report is a claim, which is this repo's own rule.
- Last round's demoted detail landed two lines below a bullet already making the same claim; deduped, keeping only the `--` / `/* */` tokens that were new.
- The previous message's "32 mutants" counted the three controls as mutants *and* omitted M1–M6. The module's matrix is the authority; this message quotes its shape, not a total.

### Coverage

```
POS 11 · BASE 6/5 · PREV 5/6
S1  "rollback" shaved out ........ 1 failed   <- this round's actual regression
S2  whole checklist replaced ..... 1 failed
S3  stderr rule inverted ......... 1 failed
S4  reworded-rule shape deleted ... 1 failed
T1/T2 · U1/U1b/U6/U7 · V1–V4 · W1–W5 · Z1/Z3/Z4/Y9/Z8 · X3/X4/X5/X6/X8 · Y4/Z9 · M1–M6/M8/M10
```

Four mutants failed to *apply* on the first pass and were caught by assert-before-edit rather than scored SURVIVED — sixth time in this PR, and this round it caught two of **my own** mutants written against mis-remembered wrapping.

Body 11,953 → 11,985 B (budget 12,038). Gate `--tier both` PASS. `rollback` verified present in the pushed tree, not just in my intent.

### Ledger

`round 8 · payload lines changed THIS round: ~12 (skill 8, reference 4) · since round 1: ~530 · elapsed: ~50 m`

Gate does not fire. Round 9 dispatched, blind.

ZacxDev:
## Round 9 (delta re-audit of `b674a7f4..907d6997`) → fixed in `9a8b9331`

Two 🟡, two 🟢 — all four in the guards, none in what the skill tells an operator to do. **This fix touches only `scripts/tests/`: payload lines changed = 0.**

### Two pins that certified presence, not meaning

`SKILL_REWORD_REGRESSION` pinned a 62-char bolded **phrase** inside a bullet whose head was unpinned. Measured: inverting the head to *"do NOT spend the round hunting regressions the fix round introduced"*, and deleting the other two regression shapes, were each **green** with the phrase byte-identical. The pin proved the words were there while the instruction around them could say the opposite — the same lesson as `EVIDENCE_LESSON_BC` two rounds ago, and as this PR's own line *"a pin that stops mid-sentence leaves the tail free to argue the opposite"*, written here and then violated here. Now the whole bullet.

**The environment brief was unpinned entirely**, and it is payload — it's what the operator puts in front of the auditor. Green before this commit: gutting its lead to *"do not bother briefing the auditor"*, deleting the zsh word-split warning, and **re-applying the exact worktree-duty regression round 8 diagnosed** (shifting *verify your worktree clean **yourself*** back to *leave your worktree clean* — operator → auditee). Round 8 repaired that sentence and didn't pin it, so the repair was re-appliable with no signal. That paragraph was reworded in **4 of this PR's 9 commits**, always under byte pressure — the same churn profile that justified pinning the checklist. Pinned whole now.

Also: last round's new comment said the checklist was shaved by *"every round"*; measured by the block's byte size per commit, it was **4 of the 8** — corrected, with the count. And a comment reading "the other two operator instructions" now covers five.

### The process fix that matters more than any of them

This round's edits ran through a script that **writes and re-reads after each substitution** and reports which ones didn't apply. The previous multi-edit scripts wrote once at the end, so one failed assertion discarded every earlier edit — exactly how `rollback` was deleted while its message claimed a restore. **It happened again on the first attempt at this round's fixes**: the script died on one target and wrote nothing. This time the harness said so out loud instead of leaving a false commit message.

### Coverage

```
POS 11 · BASE 6/5 · PREV 5/6
N1  delta-bullet head inverted ...... 1 failed   <- green before this commit
N2  other shapes deleted ............ 1 failed   <- green before this commit
P1  worktree duty re-regressed ...... 1 failed   <- green before this commit
P2  environment lead gutted ......... 1 failed   <- green before this commit
P3  zsh warning deleted ............. 1 failed   <- green before this commit
S1–S4 · T1/T2 · U1/U1b/U6/U7 · V1–V4 · W1–W5 · Z1/Z3/Z4/Y9/Z8 · X3/X4/X5/X6/X8 · Y4/Z9 · M1–M6/M8/M10
```

Gate `--tier both` PASS.

### Ledger — and the gate's own first live test

`round 9 · payload lines changed THIS round: 0 · scaffolding: 61 · since round 1: ~530 · elapsed: ~55 m`

Verified from the commit itself, not asserted: `git show --numstat` lists no file under `claude/skills/`. **This is the first zero-payload round of the ladder.** The attribution gate this PR adds requires **two consecutive** to fire, so it does not fire here — round 10 decides it, on the rule as written rather than on fatigue. If round 10's fixes are also scaffolding-only, the gate fires, the remaining findings become a follow-up task, and this ladder ends by its own rule.

Round 10 dispatched, blind.

ZacxDev:
## Round 10 → **CLEAN. The ladder ends here — and the gate this PR adds fired on its own ladder.**

No 🔴, no 🟡. Three 🟢, all count slips in comments inside `scripts/tests/test_audit_ladder_stop_rule.py`; none changes what an operator or a maintainer does.

**Both stop conditions agree, for different reasons:**

- **The clean-round rule** — no finding requires a fix, so this round is the last one. Per the skill's own text: *do not run another round to confirm a clean round.*
- **The attribution gate** — round 9's fix changed **0 payload lines** (verified independently of its commit message: `SKILL.md` and the reference file are byte-identical between `907d6997` and `9a8b9331`), and all three of round 10's findings would also be fixed in scaffolding. That is **two consecutive zero-payload rounds ⇒ the ladder has left the PR. Stop.**

So the three 🟢 are **filed, not fixed** — which is what the gate instructs. Follow-up task: correct three counts in `scripts/tests/test_audit_ladder_stop_rule.py` (the env-brief comment's "four of nine" → six of eleven commits, or four of eight *fix* commits; N2's "the other two regression shapes" → three; the matrix header's "two pins" → one pin plus one unpinned block; and the renamed test's docstring, which still says "neither"/"either" for a body asserting six constants). Closed when that PR merges or a named reader dismisses it in writing. **Spending a round on them is the behaviour this PR exists to stop.**

### Merge gate

- Merged tree built off current `origin/main` (`add34a14`, 16 commits past the branch point) — clean merge, no conflict.
- `scripts/gate.sh --tier both` on that merged tree: **`GATE: RESULT=PASS exit=0`**.
- Sandbox tier, the one Tekton gates on, also on the merged tree: `nix build .#checks.x86_64-linux.pytests` and `.nodetests` both built.
- Branch checks: `tekton/devrc-pytests` and `tekton/devrc-nodetests` SUCCESS.

### What ten rounds cost and bought

Rounds 1–8 each found a defect in the **payload** — the rule itself — and several were the mirror image of the defect the previous round fixed:

| round | the defect |
|---|---|
| 1 | the gate's range was cumulative; a per-round condition can't consume it — on #498 it never fires |
| 2 | keyed on file type, so it fired on 60% of this repo's PRs, including this one |
| 3 | the section heading still stated the retracted rule — **and was pinned**, so fixing it went red |
| 4 | the range form was wrong on 2 of 4 ladder shapes, including the one agent worktrees produce |
| 5 | `<base>` undefined; every failure mode of the command sums to zero, the gate's own verdict |
| 6 | `<base>` defined as the ref that is stale by default; the object store fails on stderr while exiting 0 |
| 7 | the zero-check rewrite widened one axis and silently narrowed another |
| 8 | a commit message that **narrated the opposite of its own diff** (`rollback`), because the checklist was unpinned |
| 9 | two pins that certified a phrase was present without certifying the instruction around it |
| 10 | clean |

The rule shipped here is the one artifact in the PR that no round could break by round 10; every round's mutation battery is preserved in the module, and the ladder stopped on the same rule it was built to state.

PR #902, opened 2026-08-27
fix(cairn): header lookup was case-sensitive — in production the freshness stamp and the truncated-transfer guard were both inert

ZacxDev:
Found by the **first live call** after deploying `0.4.0` — not by 358 tests, and not by seven rounds of adversarial audit on #863.

```
dict(resp.headers).get("X-Store-Entries") -> None
dict(resp.headers).get("x-store-entries") -> 75
```

## What broke

`dict(resp.headers)` discards the case-insensitivity of `email.message.Message`. HTTP/2 lowercases every header name and this host sits behind Cloudflare, so in production every key arrives lowercased while the code asked for the mixed-case spelling.

Three things went wrong, all silently:

| | effect |
|---|---|
| `snapshot=UNSTAMPED` | the provenance stamp the whole design rests on, lost |
| `revision=unknown` | on every cached banner |
| 🔴 **the count cross-check never fired** | `declared` was always `None`, so the guard against a **truncated transfer** was inert in the only environment that matters |

The third is the serious one. That guard was added specifically so a server declaring more entries than the archive holds gets refused. It passed its test locally the whole time and did nothing in production.

## 🔴 Why every layer of review missed it

The test shim forwarded headers in whatever case the **in-process** server sent. It reproduced production's **topology** — the hop that adds `CF-Connecting-IP` — but not its **normalization**. Both are needed; only one was modelled.

A live probe found it in one call, because a live probe is the only thing that had ever spoken the real wire format. Seven audits reasoned about the code and could not have seen this.

## Fixes

- The client lowercases keys once at the boundary and looks up lowercase.
- **Both** test servers (`_CloudflareShim` and the one-shot `_serve_once`) now lowercase, so no test can pass against a client whose lookup only works for the case the *fixture* happens to send.

## Verification

| | result |
|---|---|
| unmutated | 2 passed |
| revert to `dict(...)` + mixed-case | **2 failed** |
| before the `_serve_once` change | only **1** failed — the mutation itself revealed the second fixture was still the wrong shape |
| suite | **360 passed** |

⚠ `revision=unknown` is **correct** and is now pinned as such: `/snapshot` deliberately sets no `X-Store-Revision`, because a full snapshot spans every scope and a revision is per-scope. My first version of the test asserted the opposite and failed — the assertion was wrong, not the code.

## Blast radius

**Reversible.** Client-side only, plus test fixtures. No server change, no manifest. The deployed `0.4.0` image is unaffected — this fixes the *client* that reads it.

https://claude.ai/code/session_01A7CtMW371evXEXDkjgTNbz

### DIFF
--- a/scripts/cairn
+++ b/scripts/cairn
@@ -178,26 +178,31 @@ def fetch_snapshot(
     req.add_header("User-Agent", "subsystem-store-client/1")
     try:
         with urllib.request.urlopen(req, timeout=timeout) as resp:
-            # 🔴 LOWERCASED, because `dict(resp.headers)` THROWS AWAY the
-            # case-insensitivity that makes header lookup work. `resp.headers`
-            # is an `email.message.Message` whose `.get()` ignores case; a plain
-            # dict's does not. HTTP/2 sends header names lowercased, and this
-            # host is behind Cloudflare, so in PRODUCTION every key arrives as
-            # `x-store-entries` while the code asked for `X-Store-Entries`.
+            # 🔴 RETURN THE MESSAGE, NEVER A DICT BUILT FROM IT.
+            # `resp.headers` is an `email.message.Message` and its `.get()` is
+            # already case-insensitive; `dict(resp.headers)` throws that away.
+            # HTTP/2 lowercases every header name and this host is behind
+            # Cloudflare, so in PRODUCTION the keys arrive lowercased while the
+            # code asked for the mixed-case spelling. MEASURED against the live
+            # pod on the first real call after deploying 0.4.0:
+            #     dict(resp.headers).get("X-Store-Entries") -> None
+            #     dict(resp.headers).get("x-store-entries") -> 75
+            # Silent consequences: `snapshot=UNSTAMPED`, `revision=unknown`, and
+            # — worst — the count cross-check against `X-Store-Entries` NEVER
+            # FIRED, so the guard against a truncated transfer was inert in the
+            # only environment that matters. 358 tests and seven audit rounds
+            # missed it: the test shim reproduced production's TOPOLOGY but not
+            # its NORMALISATION.
             #
-            # MEASURED against the live pod, first real call after deploy:
-            #   dict(resp.headers).get("X-Store-Entries") -> None
-            #   dict(resp.headers).get("x-store-entries") -> 75
-            # Consequences, all silent: the freshness stamp recorded as
-            # `snapshot=UNSTAMPED`, `revision=unknown` on every cached banner,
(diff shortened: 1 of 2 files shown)


ZacxDev:
🔴 **Correcting a number in this PR's own body, and fixing a second bug the audit found in the fix itself.**

**1. The mutation count was under-reported.** The body says reverting fails **2** tests. The suite said **3** — I had run with `-k TestHeaderCaseInsensitivity`, which is the `-run`-filter trap: a filtered count is a claim about the filter, not about the suite. Unfiltered, after removing a duplicate test, it is now **2 failed / 359 passed**, and here is the full matrix:

| run | result |
|---|---|
| baseline | 359 passed |
| revert to `dict(...)` + mixed-case lookups | **2 failed** |
| isolated — only the count lookup reverted | **1 failed** (so the count test is not vacuous) |

**2. 🔴 The fix contained a second bug.** `{k.lower(): v for k, v in resp.headers.items()}` solves the case problem and creates another: on a **duplicated** header a dict keeps the **LAST** value where `Message.get` returns the **FIRST**. Measured on a doubled header: Message → `1`, comprehension → `99`.

That is this repo's own doctrine applied inconsistently — `sole_header` (`server.py:422`) exists *because* "a bare `.get()` silently takes the FIRST value" was a real smuggling bug, and its rule is **refuse** a duplicate, not quietly pick the other end. `resp.headers` is already case-insensitive; building any dict from it was the entire mistake. Now returns the `Message`.

**3. Three test-quality findings, all fixed.** The new count test took `live_store` and never used it (so it read as exercising the shim path while running against `_serve_once`) **and** duplicated the pre-existing `test_a_count_disagreeing_with_the_header_is_REFUSED` exactly — two tests, one signal. Removed the duplicate, folded its explanation into the original. The class docstring claimed *"the shim now lowercases, so every header-dependent test exercises the production shape"*, true of only one of the two servers — narrowed to name both. And `_CloudflareShim` still sent `Content-Length` mixed-case three lines below a "LOWERCASED, LIKE THE REAL EDGE" comment.

**4. `revision=` is dead output**, now labelled as such rather than merely pinned: `/snapshot` sets `X-Store-Revision` on no path and cairn always fetches unscoped, so the field can never carry a value. The closing condition for making it live is written into the test instead of left implicit.

What the audit **confirmed** and I am not re-litigating: the boundary conversion is complete on the side that had the bug, the server has no matching latent bug (every read is `self.headers.get`/`get_all`, never `dict(self.headers)`), and the `revision=unknown` justification checks out.

🤖 via [Claude Code](https://claude.com/claude-code)

PR #903, opened 2026-08-27
docs(handoff): land the extensions-pattern-analysis findings — the analysis was wrong where a reader would act on it

ZacxDev:
The 2026-08-26 analysis session wrote its findings to `/tmp/opencode/handoff/extensions-pattern-analysis.md` and never landed them. For a day the only copy sat unversioned in `/tmp` — and a `/resume` pointed at `claudedocs/handoff-extensions-pattern-analysis.md`, which had never existed.

This lands it, corrected. Re-read against `nix/home.nix`, confirmed byte-identical to `origin/main` first so the working-tree copy was authoritative, then put through four rounds of adversarial audit.

## The correction that matters

Tier 1 was described as a single mechanism covering both MV3 extensions, with `mv -T --exchange` in the pipeline. `nix/home.nix:759` says the opposite in as many words:

> This copy is deliberately weaker than browser-bridge's: no `--exchange` two-attempt TOCTOU dance, because nothing here is serving traffic mid-switch

discord-embed-ext moves the old tree aside and then moves the new one in, so its target **is** briefly absent, by design. Only browser-bridge gets the atomic swap, because only browser-bridge is serving a long-poll to a live service worker mid-switch.

This is worth a PR on its own: the source's stated trigger for extracting a shared helper is *"if a THIRD extension wants this"*, and a merged one-pipeline description is exactly what would get the wrong shape pasted. The doc now carries a table of every way the two blocks diverge.

## The rest

The doc carries 7 inline `CORRECTED` markers. Rather than restate them here — a total maintained in parallel with the thing it counts drifts, which this PR demonstrated three times — read the markers.

## What the audit ladder found

Four rounds. Each of the first three found real defects, and **each fix round introduced the next one** — which is the argument for the ladder, not against it.

- **Round 1** — the correction doc had reproduced the error class it was written to correct: it repeated the original's claim that `rm -rf` on a symlink follows and deletes the target. False, verified by experiment. Plus a misattributed trial count, an undercount inside the correction-of-an-undercount, and a missing third divergence between the two extensions.
- **Round 2** — round 1's retraction over-corrected into a new false absolute ("rm does not follow symlinks, with or without `-r`"). Measured: `rm -rf <link>/` **with a trailing slash** does follow it and empties the target, rc=0, silent. In a doc that guides deploy code and whose next bullet explains a `[ ! -L ]` guard, that absolute read as licence to omit the guard.
- **Round 3** — the one instruction the doc tells a future editor to trust ("count the `CORRECTED` markers") returned 6 while the doc said 7. Also: the silent-destruction property is `-r`-specific — `rm -f <link>/` fails loudly, rc=1.
- **Round 4** — in progress.

## Gate

Docs-only, one new file, no code touched.

- **Sandbox tier** — `nix build .#checks.x86_64-linux.{pytests,nodetests}`, the tier Tekton gates on: **PASS** (exit 0)
- **node tier (dev host)**: PASS — 39 files, 1292 tests, floor 1239
- **content gates**: PASS — `test_ci_claim_matches_reality`, `test_no_captured_text`, `test_no_captured_markup`, `test_no_client_hostnames`, `test_no_public_ips`
- **pytest tier (dev host)**: 17,220/17,225 pass, **3 failures — all pre-existing, not from this change.** Control run: 2 of the 3 fail identically on unmodified `origin/main` at `20beb3c4`; the third passes in isolation, so it is parallelism-dependent. This change is one markdown file and cannot reach them. There is an open `fix/unbreak-pytests-gate` worktree on this box independently.

Gate results are claims about this branch at its own base, not about the merged tree.

### DIFF
--- a/claudedocs/handoff-extensions-pattern-analysis.md
+++ b/claudedocs/handoff-extensions-pattern-analysis.md
@@ -33,11 +33,12 @@ working-tree copy was authoritative — and then adversarially re-audited. Each
 correction is marked `CORRECTED` inline with what the source actually says. The
 architecture was substantially right; the errors were in the details a reader acts on.
 
-The corrections are marked inline rather than counted here **on purpose**. The first
-version of this correction said "six", and the audit then found a seventh and an
-eighth (`rm -df`, and a false claim about `rm -rf` following symlinks) — a total
-maintained in parallel with the thing it counts will drift, so count the `CORRECTED`
-markers rather than trusting a number in this paragraph.
+The corrections are marked inline rather than counted here **on purpose**, and this
+paragraph is itself worked evidence for why. The first version said "six"; an audit
+then found a further wrong claim, so it became seven; the sentence recording *that*
+miscounted which ones were new. Three attempts, three wrong totals, in the paragraph
+whose subject is miscounting. **Count the `CORRECTED` markers in this document; do not
+trust any total written in prose, including one written by whoever edits this next.**
 
 Every line number below is against `nix/home.nix` at `origin/main` as of 2026-08-27
 (3,873 lines). Line numbers rot — treat them as a starting offset for a grep, not an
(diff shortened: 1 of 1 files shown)


ZacxDev:
Analysis corrections doc. The discord-embed-ext work is complete — this document adds no actionable code.

PR #915, opened 2026-08-27
feat(subsystem-store): a scope you may NOT see enumerated every scope you may not see — four channels, closed at the index (cairn phase 3, criteria 1-3)

ZacxDev:
Cairn phase 3, **criteria 1, 2 and 3 ONLY** — two-token authorization on the **read** path.

## 🔴 The write path (criteria 4-10) is NOT in this PR

No new verb, no append endpoint, no `PUT`. Asserted, not stated:

- `do_POST = do_PUT = do_PATCH = do_DELETE = _reject_write` is **byte-for-byte untouched**, and `TestPhaseOneScope::test_the_server_declares_no_write_handler` still pins it.
- `API_ROUTES` is still exactly `("recall", "search", "snapshot")`; `TestPhaseOneScope.ROUTES` did not move.
- A new test, `TestScopeFilteringIsNotAWriteVerb`, restates both claims **inside the section that added scoping**, plus a behavioural check that all four write verbs are still `405 read-only` when presented a valid **scoped** token — so a future branch that adds a verb "while it is touching auth anyway" fails a test here.

## The finding that drives it

`GET /api/v1/recall/<never-existed>` answers **200**, `X-Store-Status: scope-absent`, and the body ends `scopes the store does hold: <every scope>`. The absent path was **already** an enumeration oracle. A per-route "is this scope yours" check would have closed almost none of it, because three of the four channels fire on requests that are entirely legitimate.

## The four channels, and how each is closed

| # | channel | fires when | closed by |
|---|---|---|---|
| 1 | `known_scopes` | every `scope-absent` report | index narrowing in `load_store` |
| 2 | `malformed_elsewhere` | **every status** — including an ordinary 200 for a scope the caller DOES own | index narrowing in `load_store` |
| 3 | `search?all_scopes=1` | names **no scope**, so a per-scope check has nothing to refuse; searches every scope's CONTENT | index narrowing in `load_store` |
| 4 | the `/snapshot` tar | ships the entry files themselves; never builds an index | filter on `_snapshot`'s candidate list |

1-3 all derive from the single `SubsystemIndex` returned by `subsystem_recall.load_store` — the one site both readers load from — so a `visible_scopes` narrowing there closes all three at once and **cannot be forgotten by a route**. The filtered index is rebuilt from the two public fields only:

```python
SubsystemIndex(
    by_scope={k: v for k, v in index.by_scope.items() if k in allowed},
    malformed=tuple(m for m in index.malformed if m.scope in allowed),
)
```

`scopes`, `malformed_in`, `malformed_outside`, `entries` and `__len__` are all derived from those two, so every derived answer narrows with them — including accessors added later. `subsystem_resolver.py` is **not touched** (#360's non-goal).

Channel 4 is filtered at the *candidate* list rather than at the tar members, so an out-of-allowlist scope is never classified, never opened, and can never reach the `unreadable` list — a refused scope must not be able to 503 somebody else's snapshot.

### Two seams the index filter cannot reach

- **`X-Store-Revision`** reads `<store>/<scope>/.git/HEAD`, outside the index entirely. No scope in the served copy is a git repo today, so the leak is **latent** — exactly the state in which a guard gets skipped. Gated by construction, and the test **makes** a denied scope a real git repo with a real commit before asserting the header is still `unknown`. It also asserts the caller's *own* scope still reports its sha, because over-filtering would silently delete the `scope@sha` determinism guarantee.
- **The per-request handler reset** is `()` (nothing visible), never the unrestricted `None`. `None` is the unrestricted sentinel everywhere else in the file, which makes it the one value this field must not default to.

## Criterion 1 — the token file row format

```
<token>                                    # LEGACY: identity `legacy`, ALL scopes
<token>   <identity>   <scope>,<scope>     # MAPPED: named holder, scoped
```

🔴 **A bare row stays valid and means unrestricted scope.** That is the migration (mapped rows land beside the old shared token) and the rollback required by criterion 10 (put the one line back — no code change). A file mixing both loads, and any legacy row emits **one loud stderr line** at startup naming the count and the legacy fingerprints.

🔴 **Whitespace now separates the three FIELDS of one row, not two tokens.** The old parser split the whole file on whitespace. A line that used to be two credentials is now refused as `malformed token row N of M` — **never reinterpreted**. `MAX_IDENTITY_CHARS` (32) is deliberately **below** `MIN_TOKEN_CHARS` (43), so a token in the identity field is refused by arithmetic rather than by a spelling.

`identity=` is an **additive** audit field. `token=<fingerprint>` is unchanged and is still the only thing that can tell one holder's current credential from its previous one during an overlap rotation.

### The guard ladder

Guards 1-5 are **unchanged in wording and order**. Six new ones, each with its own sentence, each reachable by an input that passes every earlier guard:

| # | message | reached by |
|---|---|---|
| 1-5 | *(unchanged)* | — |
| 6 | `malformed token row N of M` | `<tok> zach` (2 fields) / `<tok> zach s extra` (4) |
| 7 | `invalid identity in token row N of M` | `<tok> Za_ch s` (charset) / `<tok> <48-char lowercase> s` (length) |
| 8 | `reserved identity in token row N of M` | `<tok> legacy s` |
| 9 | `empty scope allowlist in token row N of M` | `<tok> zach ,` |
| 10 | `invalid scope in token row N of M` | `<tok> zach kelp.forest`, `<tok> zach a,,b`, `<tok> zach -` |
| 11 | `duplicate identity 'X'` | two rows, both well-formed, same identity |

**Guard 11 exempts `legacy`, deliberately** — two bare rows are the ordinary current+previous overlap file that guards 1-5 exist to support, and that criterion 10's rollback restores. Rotating a *mapped* credential uses a second identity (`zach-prev`); two rows claiming one identity with disagreeing allowlists have no defined precedence, so they are refused. Both directions are tested.

**Guard 10 checks the FOLDED value, not the typed one.** `-` and `___` are inside `[A-Za-z0-9_-]+` and perfectly namable in a URL, and `normalize_ref` folds both to the empty string — an allowlist entry that matches no index key, i.e. a grant that reads as working and does nothing. That is precisely what the guard's own sentence promises to prevent, so it is checked where it matters. Found by reading the guard against its description, not by a test.

## Criterion 3 — proven, not assumed

`TestRefusedIsIndistinguishableFromAbsent` asks the **same token** for the **same path** at the **same store root**, once when the scope holds an entry and once when it never existed — with the store-wide `.md` count and every mtime held constant (the scope is rebuilt under a different name). Status line, **full body bytes** and **every** `X-Store-*` header compare equal, for `recall` and for `search`.

**Positive control**: the same two phases driven by an unrestricted legacy token, watched to **differ** (`recalled` + content vs `scope-absent`). Without it the equality above would be measuring the harness.

## ⚠ Residual leak, stated rather than left to be found

`X-Store-Snapshot` and the freshness prose stay **store-wide** and carry a total `entry-files=` count over scopes the caller cannot name. Scoping it would make "how old is this copy" a function of your own allowlist, which is worse — but it is a real count leak, it is documented in the module docstring and the README, and it is why the byte-identity claim is about refused-vs-absent and never about two different stores.

## Red → green matrix

🔴 **Honestly labelled.** `server.py` and the absent path both exist at the base ref and the leak is real there, so the channel and byte-identity tests are genuine **regressions**. The token-guard tests are **invariant guards**: they call a parser that does not accept a three-field row at base, so their red would be a shape error and would prove nothing. The mutation matrix below is their evidence.

Because the guards' red at `origin/main` is vacuous, the meaningful "before" tree is **the fix reverted with the API present**. Each seam reverted independently:

| reverted seam | tests that go RED |
|---|---|
| `load_store` ignores `visible_scopes` | `CHANNEL_1_known_scopes…`, `CHANNEL_2_malformed_elsewhere…`, `CHANNEL_3_all_scopes_search…`, `TWO_TOKENS_ON_ONE_SERVER…`, `RECALL_a_refused_scope_is_BYTE_IDENTICAL…`, `SEARCH_a_refused_scope_is_BYTE_IDENTICAL…`, and 4 in `TestTheReaderNarrowingItself` — **10 red, 10 green** (the 10 green are the positive controls, which use an unrestricted token and correctly do not move) |
| `scope_revision` gate inert | `a_DENIED_scopes_revision_is_unknown_even_though_it_HAS_one`, `visible_scopes_None_is_UNRESTRICTED…` — **2 red** |
| `_snapshot` filter inert | `CHANNEL_4_the_snapshot_tar_carries_only_allowed_members`, `a_scope_FILTERED_snapshot_of_a_denied_scope_ships_nothing` — **2 red** |

All green at HEAD.

## Mutation sweep — 19 mutants, 18 killed, 1 recorded equivalent

Run under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` purged between mutants (a same-length edit landing in the same whole second is otherwise scored SURVIVED without ever executing). Anchors asserted unique. Originals restored by sha256-verified copy-back, never `git checkout`. A known-caught **positive control** was carried in the batch and reported KILLED, so the harness is demonstrably running the mutated code.

Each kill was checked for **its own** reason, not merely "some test failed":

|---|---|---|
| `CONTROL_narrowing_inert` (positive control) | KILLED | 3 channel tests |
| guard 6 → accepts a 2-field row | KILLED | `IndexError` reaching `fields[2]` — i.e. without the guard a 2-field row is a startup **crash**, which is what the guard converts into a clean config error |
| guard 6 → accepts a 4-field row | KILLED | `DID NOT RAISE ValueError` |
| guard 7 charset `fullmatch`→`search` | KILLED | `DID NOT RAISE` |
| guard 7 length cap raised above the token floor | KILLED | `DID NOT RAISE` |
| guard 8 compares the wrong name | KILLED | `DID NOT RAISE` |
| guard 9 tests the LIST not its CONTENTS (`not any(x)` → `not x`) | KILLED | message assertion — the raise came from **guard 10** instead, which is exactly the ladder-ordering property guard 9 exists for |
| guard 10 class `fullmatch`→`search` | KILLED | `DID NOT RAISE` |
| guard 10 skips the empty entry | KILLED | `DID NOT RAISE` |
| guard 10 drops the folds-away half | KILLED | `DID NOT RAISE` |
| guard 11 never records a first sighting | KILLED | `DID NOT RAISE` |
| guard 11 legacy **exemption** removed | KILLED | `ValueError: duplicate identity 'legacy'` — the over-strict direction |
| legacy warning never fires | KILLED | `assert len(warnings) == 1` → `0` |
| legacy warning **always** fires | KILLED | its own negative control |
| `authorize` returns the FIRST configured record | KILLED | `dana` got `scope-absent` for her own scope — the one-match-three-facts seam |
| `_handler` reset `()` → `None` | **SURVIVED** | **recorded equivalent mutant** (below) |

**The survivor, honestly:** `_visible_scopes` is overwritten from the matched record on every path that can reach a route handler (health returns early; unidentified, locked-out, non-API-prefix and bad-token all refuse and return before dispatch). So the reset value is unobservable **today**. It is kept as `()` rather than `None` because a future edit that stops overwriting it must not silently inherit "unrestricted" — the same reasoning the file already records for two other equivalent mutants. Recorded rather than left as an unexplained survivor.

## Test counts and the gate

- `scripts/tests/test_cairn_cli.py` + `scripts/tests/test_subsystem_store_api.py`: **359 → 403** passed (+44).
- Reader-adjacent suites (`test_subsystem_recall`, `test_subsystem_resolver`, `test_subsystem_touch`, `test_service_recon`, `test_subsystem_audit`, `test_store_content_not_copied`, `test_handoff_doc`): **1967 passed**.
- **`scripts/gate.sh --tier pytest` → `GATE: RESULT=PASS exit=0`**, `TOTAL collected=17279 passed=17277 skipped=2 failed=0` across 36 targets, every per-target floor cleared. *(Dev-host tier, at this branch's tip. The sandbox tier `nix build .#checks.x86_64-linux.pytests` — the one Tekton gates on — has **not** been run here; and `strict` is false on branch protection, so this is a claim about this branch, not about the merged tree.)*

## Other callers of the changed signature

`git grep`ed for every `load_store` / `rc.recall(` / `rc.search(` call site. `visible_scopes` defaults to `None` = unrestricted, so the CLI (`cairn`, `/resume`, `subsystem_recall.main`) is unchanged — pinned by `test_the_CLI_DEFAULT_is_still_unrestricted`. One in-repo mutation anchor in `test_subsystem_recall.py::test_kills_the_COLLECT_policy` moved with the refactor and was updated; the mutation it applies is unchanged.

## Incidental fix

`test_cairn_cli._load_api` never registered its module in `sys.modules`. CPython's `dataclasses._is_type` does `sys.modules.get(cls.__module__).__dict__` with **no None guard**, so under `from __future__ import annotations` the first `@dataclass` in `server.py` raised `AttributeError: 'NoneType' object has no attribute '__dict__'` at import — **43 collection errors**, none of them near the change that triggered them. The sibling loader in `test_subsystem_store_api` had always registered; this was the second copy of that predicate and the one that was wrong.

## Deploy note

The live secret is a single bare token, which loads as a legacy row: **unrestricted scope, zero behaviour change on deploy**, plus a loud startup line saying so. Scoping is opt-in per row. Nothing here is deployed and no image version was bumped.

### DIFF
--- a/scripts/lib/subsystem_recall.py
+++ b/scripts/lib/subsystem_recall.py
@@ -1297,7 +1297,12 @@ class RecallReport:
         return caveat_text(f"{self.scope}/", badges_present(self.listing))
 
 
-def load_store(store_root: str | Path, *, verb: str) -> tuple[Path, SubsystemIndex]:
+def load_store(
+    store_root: str | Path,
+    *,
+    verb: str,
+    visible_scopes: Sequence[str] | None = None,
+) -> tuple[Path, SubsystemIndex]:
     """Resolve the store root and load its index, or raise with a sentinel.
 
     🔴 ONE PLACE, because `recall` and `search` open the SAME store for the same
@@ -1306,6 +1311,29 @@ def load_store(store_root: str | Path, *, verb: str) -> tuple[Path, SubsystemInd
     did NOT happen ("nothing was recalled" / "nothing was searched") or it reads
     as the ordinary nothing-recorded-yet case, which is the confident zero this
     module exists to prevent.
+
+    🔴 `visible_scopes` IS A NARROWING OF THE WHOLE INDEX, AND IT IS APPLIED HERE
+    FOR EXACTLY THE REASON ABOVE — this is the single site both readers get their
+    index from, so a caller that may only see some scopes cannot be given a wider
+    one by a route that forgot to filter. `None` means UNRESTRICTED (every local
+    CLI caller); a sequence means "these normalized scopes and nothing else".
+
+    Filtering the INDEX rather than each answer is what closes four leaks at
+    once, and every one of them was measured on the deployed API before this
+    change:
+
+      * `known_scopes` — a `scope-absent` report ends with "scopes the store does
+        hold: …", so asking for a scope you do not have enumerated every scope
+        you do not have either.
+      * `malformed_elsewhere` — the "(+N further malformed entries in OTHER
+        scopes …)" block names those scopes on EVERY status, not only on a miss.
+      * `search?all_scopes=1` — it names no scope at all, so a per-scope refusal
+        check cannot see it; it searched the CONTENT of every scope in the store.
+      * the entries themselves.
+
+    An empty sequence therefore means "no scope is visible", NOT "unrestricted".
+    That direction is deliberate: a caller that forgot to resolve an allowlist
+    gets an empty store, not the whole one.
 
     🔴 IT LOADS WITH `ON_MALFORMED_COLLECT`, AND THAT IS THIS MODULE'S POLICY
     DECISION, NOT THE LOADER'S. A malformed entry no longer raises here: it comes
@@ -1328,12 +1356,26 @@ def load_store(store_root: str | Path, *, verb: str) -> tuple[Path, SubsystemInd
             f"store. Nothing was {verb}; this is NOT 'nothing recorded yet'"
         )
     try:
-        return store, load_index(store, on_malformed=ON_MALFORMED_COLLECT)
+        index = load_index(store, on_malformed=ON_MALFORMED_COLLECT)
     except OSError as exc:
         raise EntryUnreadableError(
             f"index entry unreadable: under {store} ({type(exc).__name__}: {exc}) — the "
             f"store was not fully read, so this report would be INCOMPLETE"
         ) from exc
+    if visible_scopes is None:
+        return store, index
+    # 🔴 REBUILT FROM THE TWO PUBLIC FIELDS, not by mutating a frozen dataclass
+    # and not by adding a third field. `scopes`, `malformed_in`,
+    # `malformed_outside`, `entries` and `__len__` are ALL derived from
+    # `by_scope` + `malformed`, so narrowing exactly those two narrows every
+    # derived answer at once — including the ones a future accessor adds. A
+    # per-answer filter would have to be repeated at every one of them, which is
+    # the shape that is wrong at all but one site.
+    allowed = {normalize_ref(s) for s in visible_scopes}
+    return store, SubsystemIndex(
+        by_scope={k: v for k, v in index.by_scope.items() if k in allowed},
+        malformed=tuple(m for m in index.malformed if m.scope in allowed),
+    )
 
 
 def recall(
@@ -1346,6 +1388,7 @@ def recall(
     page: int = 1,
     focus_paths: Sequence[str] = (),
     focus_source: str | None = None,
+    visible_scopes: Sequence[str] | None = None,
 ) -> RecallReport:
     """Surface an entry's `## What it is` + `## Pointers` + `## Nuance / work-history`.
 
@@ -1386,7 +1429,14 @@ def recall(
     if mode not in RECALL_MODES:
         raise ValueError(f"mode must be one of {RECALL_MODES}, got {mode!r}")
 
-    store, index = load_store(store_root, verb="recalled")
+    # 🔴 `visible_scopes` IS PASSED, NEVER RE-DERIVED. A scope the caller may not
+    # see must be absent from the INDEX, not filtered out of each answer — see
+    # `load_store`. Once it is absent, `scope-absent` (and its `known_scopes`
+    # list) is what a refused scope produces, which is byte-for-byte what a scope
+    # that never existed produces.
+    store, index = load_store(
+        store_root, verb="recalled", visible_scopes=visible_scopes
+    )
     # Computed ONCE, before any status branch, and passed to every one of them —
     # the block renders on all of them (`render_malformed`), so deriving it per
     # branch would be the same predicate at six sites, wrong at five.
@@ -2538,6 +2588,7 @@ def search(
     threshold: float = DEFAULT_SEARCH_THRESHOLD,
     max_hits: int = DEFAULT_MAX_HITS,
     all_scopes: bool = False,
+    visible_scopes: Sequence[str] | None = None,
 ) -> SearchReport:
     """Find HUNKS matching `query`. READ-ONLY, stdlib only, nothing is spawned.
 
@@ -2564,7 +2615,16 @@ def search(
     if not isinstance(context, int) or isinstance(context, bool) or context < CONTEXT_BULLET:
         raise ValueError(f"context must be an int >= 0, got {context!r}")
 
-    store, index = load_store(store_root, verb="searched")
+    # 🔴 THE `all_scopes` PATH IS THE REASON THIS IS AN INDEX FILTER AND NOT A
+    # PER-SCOPE REFUSAL CHECK. `?all_scopes=1` names NO scope, so there is
+    # nothing for such a check to refuse — it would search the CONTENT of every
+    # scope in the store and report hits from scopes the caller cannot name.
+    # Narrowing the index instead makes `index.scopes` below already the
+    # caller's own set, so the store-wide search is store-wide over what the
+    # caller may see and nothing else.
+    store, index = load_store(
+        store_root, verb="searched", visible_scopes=visible_scopes
+    )
     bad = index.malformed_in(scope)
     bad_elsewhere = index.malformed_outside((scope,))
 
--- a/scripts/lib/subsystem_recall.py
+++ b/scripts/lib/subsystem_recall.py
@@ -278,6 +278,7 @@ from subsystem_resolver import (  # noqa: E402
     parse_front_matter,
     parse_journal_bullets,
     resolve_ref_tiered,
+    visible_scope_set,
 )
 from subsystem_resolver import extract_sections as _extract_sections  # noqa: E402
 from subsystem_touch import (  # noqa: E402
@@ -329,6 +330,7 @@ __all__ = [
     "discarded_sensitivity",
     "sensitivity_label",
     "load_store",
+    "visible_scope_set",
     "listing_order",
     "listing_page",
     "tokenize",
@@ -1356,7 +1358,16 @@ def load_store(
             f"store. Nothing was {verb}; this is NOT 'nothing recorded yet'"
         )
     try:
-        index = load_index(store, on_malformed=ON_MALFORMED_COLLECT)
+        # 🔴 THE ALLOWLIST GOES DOWN INTO THE LOADER, NOT ONLY ONTO ITS RESULT.
+        # Narrowing afterwards still opened every file in every scope, so one
+        # unreadable entry in a scope the caller may NOT see put that file's full
+        # path into a 503 body, broke recall for everyone, and — for a FIFO named
+        # `*.md` — hung the request thread outright. See `load_index`.
+        index = load_index(
+            store,
+            on_malformed=ON_MALFORMED_COLLECT,
+            visible_scopes=visible_scopes,
+        )
     except OSError as exc:
         raise EntryUnreadableError(
             f"index entry unreadable: under {store} ({type(exc).__name__}: {exc}) — the "
@@ -1371,7 +1382,18 @@ def load_store(
     # derived answer at once — including the ones a future accessor adds. A
     # per-answer filter would have to be repeated at every one of them, which is
     # the shape that is wrong at all but one site.
-    allowed = {normalize_ref(s) for s in visible_scopes}
+    #
+    # ⚠ REDUNDANT WITH THE LOADER'S OWN FILTER TODAY, AND KEPT DELIBERATELY.
+    # `load_index` now skips a denied scope dir entirely, so this rebuild has
+    # nothing left to drop — a mutation sweep will score it as an equivalent
+    # mutant. It stays because the two filters answer different questions: the
+    # loader's decides what is OPENED, this one decides what the RESULT SHAPE is,
+    # and a future caller that hands this function an already-loaded index, or a
+    # loader that learns a reason to register a scope it did not read, must not
+    # silently widen the answer. Both derive the set from `visible_scope_set`, so
+    # they cannot come to disagree about what an allowlist means.
+    allowed = visible_scope_set(visible_scopes)
+    assert allowed is not None  # `visible_scopes is None` returned above
     return store, SubsystemIndex(
         by_scope={k: v for k, v in index.by_scope.items() if k in allowed},
         malformed=tuple(m for m in index.malformed if m.scope in allowed),
--- a/scripts/lib/subsystem_resolver.py
+++ b/scripts/lib/subsystem_resolver.py
@@ -1741,13 +1741,23 @@ def entry_mapping(text: str, *, filename: str, scope: str) -> dict[str, object]:
 # 🔴 THE TOTAL CLASSIFIER — why this exists instead of a fourth `if` arm.
 #
 # ⚠ IT LIVED IN `subsystem-store-api/server.py` UNTIL `load_index` NEEDED IT.
-# It moved here rather than being copied: "what IS this path" open-coded at two
-# sites is the predicate-at-N-sites shape that is wrong at N-1 of them, and the
-# two sites in question are `/snapshot` and the index loader — disagreeing about
-# a FIFO named `*.md` is exactly what costs a request thread. What did NOT move
-# is the ACTION TABLES: each context maps every kind explicitly, and the loader's
-# table (`_LOADER_ENTRY_ACTIONS`, below) is deliberately NARROWER than the
-# server's `_ENTRY_ACTIONS`. `server.py` imports these names.
+# It moved here rather than being copied: "what IS this path" open-coded at N
+# sites is wrong at N-1 of them, and disagreeing about a FIFO named `*.md` is
+# exactly what costs a request thread. What did NOT move is the ACTION TABLES:
+# each context maps every kind explicitly, and the loader's table
+# (`_LOADER_ENTRY_ACTIONS`, below) is deliberately NARROWER than the server's
+# `_ENTRY_ACTIONS`. `server.py` imports these names.
+#
+# 🔴 THE TWO GUARDED SITES ARE `/snapshot` AND THIS INDEX LOADER — AND THEY ARE
+# NOT ALL THE SITES. An earlier wording here said they were, and that claim was
+# false: `subsystem_touch.census()` (`glob("*.md")` then `read_text`) is a THIRD
+# glob-and-read site with NO kind check at all, so it still hangs on a fifo and
+# still 503-equivalents on a directory. It is left that way ON PURPOSE, by
+# ruling: `census()` is CLI-only, no server path imports `subsystem_touch`, and
+# the operator declined to widen the guard to it. `validate_scope` also globs,
+# but only for NAMES — it reads nothing itself and defers to `load_index`, so it
+# inherits this guard rather than needing one. Said here so the next reader does
+# not have to rediscover it, and so "two sites" stops reading as "all of them".
 #
 # Four consecutive audit rounds found the same shape of defect in `_snapshot`,
 # and every fix added one more predicate to a sequence:
@@ -1927,10 +1937,13 @@ def visible_scope_set(visible_scopes: Sequence[str] | None) -> set[str] | None:
 # 🔴 THE LOADER'S OWN TABLE, AND IT IS DELIBERATELY NARROWER THAN THE SERVER'S.
 #
 # The broad form — mirroring `server.py`'s `_ENTRY_ACTIONS` wholesale — was
-# written, reviewed and REJECTED: it also refuses `link-to-file`, `directory`,
-# `link-to-dir` and `indeterminate`, every one of which this loader reads (or
-# fails on) today, so it is a behaviour change for every local CLI caller and
-# for the writer's probe. What was decided instead is exactly two cells:
+# written, reviewed and REJECTED, because it also refuses `link-to-file` and
+# `indeterminate`, which this loader READS (or honestly fails on) today: that is
+# a behaviour change for every local CLI caller and for the writer's probe, and
+# those two cells are still `TAKE` for exactly that reason. The cells that HAVE
+# been decided, one ruling at a time, each on the same criterion — "this loader
+# has never successfully read one, so refusing it changes no legitimate caller"
+# — are the five below. The first ruling decided two:
 #
 #   `broken-link`  a dangling symlink. `Path.glob("*.md")` MATCHES A LEADING
 #                  DOT — measured, not assumed — so an Emacs lock file
@@ -1962,30 +1975,63 @@ def visible_scope_set(visible_scopes: Sequence[str] | None) -> set[str] | None:
 #                  one, and on a `replicas: 1` / `strategy: Recreate` service a
 #                  wedged thread is the worst outcome in this file. Refused.
 #
-# None of the three is ever a legitimate entry, so no legitimate caller changes
+# 🔴 AND NOW A FOURTH AND FIFTH, ON THE SAME CRITERION THE NARROW RULING USED:
+#
+#   `directory`    a DIRECTORY named `*.md`. `read_text` on one raises
+#   `link-to-dir`  `IsADirectoryError`, and because an OSError fails closed in
+#                  BOTH policies that was a store-wide `503 index entry
+#                  unreadable` for EVERY caller — `/recall` AND `/search` — off
+#                  one stray `mkdir <scope>/<slug>.md` or an rsync/restore
+#                  artefact. Measured on the tip that carried them, with a
+#                  paired control:
+#
+#                      store/beta/notes.md created as a DIRECTORY
+#                      GET /api/v1/recall/alpha, UNRESTRICTED legacy token
+#                      -> 503 "index entry unreadable: … (IsADirectoryError:
+#                         … /beta/notes.md)"
+#                      CONTROL, same shape but a dangling `.#lock.md` -> 200
+#
+#                  A symlink to a directory measured identically. 🔴 THE
+#                  CRITERION IS UNCHANGED, WHICH IS WHY THIS IS NOT A WIDENING
+#                  OF THE RULING: this loader has NEVER successfully read a
+#                  directory — it has only ever raised on one — so refusing it
+#                  changes no legitimate caller's behaviour, exactly as with
+#                  `broken-link`, `other` and `link-to-other`. It also makes the
+#                  loader AGREE with `/snapshot`'s `_ENTRY_ACTIONS`, which
+#                  already refused both kinds; two readers of one store
+#                  disagreeing about what a directory named `*.md` is was the
+#                  divergence, not the fix.
+#
+# None of the five is ever a legitimate entry, so no legitimate caller changes
 # behaviour.
 #
 # 🔴 `link-to-file` IS `TAKE`, AND THAT IS THE POINT OF THE NARROW FORM — it is
 # the cell the narrow-vs-broad ruling was actually about, and closing
-# `link-to-other` did not touch it. A symlink to a regular `*.md` is read today
-# and keeps being read. A mutant that flips this cell to REFUSE is the exact
-# over-broad regression the narrow ruling exists to prevent, and
-# `test_a_SYMLINKED_entry_is_STILL_READ_the_guard_is_NOT_the_broad_one` kills
-# it.
+# `link-to-other`, `directory` and `link-to-dir` did not touch it. A symlink to
+# a regular `*.md` is read today and keeps being read. A mutant that flips this
+# cell to REFUSE is the exact over-broad regression the narrow ruling exists to
+# prevent, and `test_a_SYMLINKED_entry_is_STILL_READ_the_guard_is_NOT_the_broad_one`
+# kills it.
 #
-# ⚠ `indeterminate`, `directory`, `link-to-dir` and `absent` are `TAKE` for the
-# same conservatism: `read_text` raises on each (EACCES, IsADirectoryError,
-# FileNotFoundError) and that raise is the four-state rule's "the store was not
-# fully READ", which is a DIFFERENT fact from "this entry is malformed" and must
-# not be quietly folded into it.
+# ⚠ `indeterminate` and `absent` are `TAKE`, and for a DIFFERENT reason from the
+# two cells that just left this list — one this round was told explicitly not to
+# fold in. `indeterminate` means "the `lstat` failed and I could not look",
+# which is not "this kind can never be an entry"; `absent` is a file that
+# vanished between `glob()` and `classify_path`. `read_text` raises on each
+# (EACCES, FileNotFoundError) and that raise is the four-state rule's "the store
+# was not fully READ" — a DIFFERENT fact from "this entry is malformed", which
+# must not be quietly folded into it. `regular-file` and `link-to-file` are
+# `TAKE` because they are what an entry IS; the residuals they carry are
+# enumerated in `load_index`'s RESIDUAL LEDGER and pinned by
+# `test_the_LOADER_RESIDUAL_SET_is_pinned`.
 _LOADER_ENTRY_ACTIONS: dict[str, str] = {
     KIND_BROKEN_LINK: REFUSE,
     KIND_OTHER: REFUSE,
     KIND_LINK_TO_OTHER: REFUSE,
+    KIND_DIRECTORY: REFUSE,
+    KIND_LINK_TO_DIR: REFUSE,
     KIND_REGULAR_FILE: TAKE,
     KIND_LINK_TO_FILE: TAKE,
-    KIND_DIRECTORY: TAKE,
-    KIND_LINK_TO_DIR: TAKE,
     KIND_INDETERMINATE: TAKE,
     KIND_ABSENT: TAKE,
 }
@@ -2012,6 +2058,19 @@ _LOADER_REFUSAL_REASON: dict[str, str] = {
         "device) — refused before `open()`, which blocks on a fifo whether it "
         "is named directly or reached through a link. Measured wedging a "
         "`/recall` request thread for 25s while the process stayed healthy"
+    ),
+    KIND_DIRECTORY: (
+        "a directory named `*.md` — refused before `open()`, which on a "
+        "directory raises `IsADirectoryError`; that OSError fails closed, so "
+        "one stray `mkdir <scope>/<slug>.md` (or an rsync/restore artefact) "
+        "answered `/recall` and `/search` with a store-wide 503 for every "
+        "caller. Delete the directory, or move its contents into a `*.md` file"
+    ),
+    KIND_LINK_TO_DIR: (
+        "a symlink to a directory — refused before `open()`, which raises "
+        "`IsADirectoryError` whether the directory is named directly or "
+        "reached through a link, and that OSError fails closed into the same "
+        "store-wide 503"
     ),
 }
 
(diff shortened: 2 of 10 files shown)


ZacxDev:
## Audit fixes — and a correction to this PR's own description

An adversarial audit of this branch found a **silent fail-open in `load_tokens`**, a second fail-open in the reader, and six guards that read as coverage while providing less. All are fixed in `be5c7b05` + `56d1327f`.

🔴 **The PR body overstated the migration path's safety, and this comment is the correction rather than a quiet edit.** The body presented "a legacy row and a mapped row may appear in one file" as the migration. It could, but the *first step anyone would actually take* — scoping a credential its holder already has — silently produced an **unrestricted** server.

---

### 🔴 1. `load_tokens` de-duplicated BY TOKEN, before parsing or validating

Both symptoms reproduced live on this tree before the fix:

**(a) The migration path failed open.**
```
<token>
<token> zach kelp-forest
```
loaded as **one** row, `identity='legacy' scopes=None` — **unrestricted**. No error. The mapped row simply did not exist. The only signal was a startup banner reading `1 of 1 token rows are bare` over a two-line file.

**(b) A duplicate-token row bypassed guards 6–10 entirely.**
```
<token> zach kelp-forest
<token> Za_CH_BAD !!!!
```
loaded clean as `zach/kelp-forest`. The second row carried an invalid identity **and** an invalid scope; neither was validated, because the row was dropped before the ladder ran.

This was guard 10's own defect class — *a grant that reads as working and does nothing* — one level up and in the unsafe direction. It also contradicted the duplicate-**identity** guard, which refuses two rows claiming one identity for having "no defined precedence", while two rows claiming one **token** with disagreeing authority silently picked one.

**Fix:** parse first, collapse after. Only *identical* records collapse (same token, identity and folded scope tuple — so `Kelp_Forest` and `kelp-forest` are correctly recognised as one grant, which a textual comparison gets wrong). Two rows sharing a token that disagree are **refused** with their own message naming both authorities. Every row now reaches the validation ladder.

The ladder gains a rung: **guard 11 = one authority per token**, **guard 12 = one row per identity** (the old guard 11, renumbered). The order is load-bearing — 11 must run first, or a verbatim duplicated row trips the identity guard. Row indices in guards 6–12 now count **physical lines**, so `rows 1 and 3` names lines the operator can see.

`test_a_duplicated_line_collapses_and_order_is_kept` (genuinely identical lines — the rotation shape) still passes.

⚠️ **Operator-facing behaviour change:** to scope an existing token you now **edit its bare row**. Adding a mapped row beneath it is refused at startup instead of silently ignored. The `MAX_TOKENS` cap also now counts physical lines rather than post-dedup rows.

### 🟡 2. `load_index` read every scope before the allowlist was applied

`load_store` narrowed the index it got **back**; `load_index` had already walked and read the whole store to build it. For a scoped caller, one unreadable entry in a scope they may not see was three defects at once — all measured:

| | observed |
|---|---|
| **Disclosure** | `503 index entry unreadable: … (PermissionError: '<store>/quartz-mine/sealed-adit.md')` — the full path of a file in a scope that caller is not allowed to know exists |
| **Denial of service** | one unreadable file anywhere broke `/recall` and `/search` for **every** caller |
| **Hung thread** | a FIFO named `*.md` blocked `open()` indefinitely. On `replicas: 1` the worker is simply gone |

`visible_scopes` is now pushed **down** into `load_index`, so a denied scope dir is never descended into. `Path.glob("*.md")` **does** match a leading dot (measured, and pinned by a test), so an Emacs lock file `.#entry.md` — a dangling symlink — is a candidate entry, which is the shape observed 503ing `/recall` in practice.

The allowlist-folding predicate is now one function (`visible_scope_set`) shared by all three narrowing sites, including `/snapshot`, which walks the store root directly and cannot use the index.

🔴 **The limit of this fix, stated rather than left to be discovered:** `visible_scopes=None` skips nothing, so for an **unrestricted** caller all three failures are exactly as open as before — and **a bare legacy token is unrestricted, which is what the pod is deployed with today.** The live API is *not* protected by this change; only a scoped token whose allowlist excludes the bad scope is. Measured:

```
visible_scopes=None                FIFO `hang.md`  -> HUNG (>6s, killed)
visible_scopes=None                `.#locked.md`   -> EntryUnreadableError
visible_scopes=('kelp-forest',)    both            -> loads, scopes=(kelp-forest,)
visible_scopes=('quartz-mine',)    FIFO            -> HUNG (its own scope)
```

Closing it for everybody needs an entry-**kind** check before `open()`. `server.py`'s `_ENTRY_ACTIONS` already classifies exactly these cases for `/snapshot` (`KIND_OTHER: REFUSE`, *"a FIFO named `*.md` blocked `open()` forever"*), and that classifier lives in the API rather than in the lib. **Left out of this PR on purpose:** it also refuses a symlink-to-regular-file entry, which the loader accepts today, so it is a behaviour change for every local CLI caller and wants its own review. Recorded in `load_index`'s docstring and pinned by `test_the_UNRESTRICTED_reading_is_UNCHANGED_and_that_is_the_residual`, so nobody reads the fix as wider than it is. **Closing condition: an entry-kind guard in `subsystem_resolver`, or an explicit decision on this PR that the residual is accepted.**

### 🟡 4. Six guards narrower than their own names

Each mutant below **passes** the assertion that was there before:

| test | what it actually asserted | mutant that passed it |
|---|---|---|
| `CHANNEL_3_all_scopes_search_narrows…` | only the *absence* of denied scopes | `scopes_searched=()` → renders `searched 0 entries in (none)` — a search that looked at **nothing** |
| `a_scope_FILTERED_snapshot…ships_nothing` | three named facts, though the docstring promised "exactly what it answers" | an extra `X-Store-Refused` header — member list, count and status all unchanged |
| `every_write_verb_is_STILL_405…` | `/recall` only | a `do_POST` on `/snapshot` — which **also** passed `test_the_route_set_did_NOT_grow`; the pair was jointly walkable |
| `assert "broken-shard" not in text` | nothing — `render_malformed` never emits a cross-scope filename at base *or* HEAD, for any token | (unfalsifiable) |
| `an_UNREADABLE_store_still_RAISES…` | a **missing** store — different condition, error type and line | (misnamed) |
| `SEARCH_a_refused_scope_is_BYTE_IDENTICAL…` | no positive control on the search path | forcing every search to `scope-absent` leaves it **green** |

All widened; the vacuous line deleted with its reason in place and the real claim moved to the positive control, which is the only reading that renders the block.

### 🟢 6. Five equivalent mutants recorded

All five fail-closed `_visible_scopes = ()` sites survive `= None` — **measured one site at a time against the full 375-test file, 375 passing each time.** They are unreachable: every entry point reassigns the field before a route reads it, so no test can distinguish the two without inventing a caller that does not exist. The direction is right and the comments stay, but they implied the guards were pinned **now**, and they are not. Recorded at the declaration in the style the file already uses for `send_error`.

### 🟢 7. Guard 6's message

`<tok> zach a, b` is four fields — the space after the comma split the scope list — and `4 fields, expected 1 or 3` sent the operator to count fields on a line that looks like it has three. A hint is now **appended** (never substituted for the refusal) and only when a comma actually appears past the identity, so it cannot be printed as a guess. The `<tokenA> <tokenB>` row still gets the sentence about two tokens, which is *its* likely cause.

---

### Verification

**Red → green. Twelve tests were watched fail, then pass at HEAD**, in two groups — and the split matters:

*Five against the real pre-fix tree* (`server.py` restored from `2f10e9a5`, everything else at HEAD):
`GUARD_6_a_SPACE_AFTER_A_COMMA…`, `GUARD_11_the_MIGRATION_SHAPE…`, `GUARD_11_a_DUPLICATE_TOKEN_ROW_no_longer_bypasses_guards_6_to_10`, `GUARD_11_fires_even_when_the_IDENTITY_agrees`, `GUARD_12_names_the_PHYSICAL_rows_across_a_collapse`.

*Seven against an isolated revert of the pushdown* — `load_index` keeps the `visible_scopes` kwarg and ignores it:
`an_EMPTY_allowlist_registers_NO_scope…`, `a_denied_scopes_NAME_is_not_registered_either`, `the_DIRECTORY_NAME_is_FOLDED_before_it_is_compared`, `a_SCOPED_caller_is_unaffected_by_an_unreadable_DENIED_entry[perm|emacs]`, `the_503_body_NAMED_the_denied_scope_and_its_PATH_over_HTTP`, `a_FIFO_named_md_in_a_DENIED_scope_no_longer_HANGS_the_reader`.

🔴 **Stated at the scope actually measured:** at the literal `2f10e9a5` these seven would go red with a `TypeError` — `load_index` has no such kwarg there — which proves nothing about behaviour. The revert-mutant is the honest baseline because it isolates the fix's *semantics* from its signature.

The remaining new tests are labelled in-place as invariant guards or positive controls, and are **not** counted as regression coverage.

**Mutation.** 23 mutants killed, 5 surviving **by design** and recorded as equivalent:

| group | result |
|---|---|
| Guards 11/12 (incl. **M18** in both its original and post-fix forms) | **9/9 killed** |
| `load_index` pushdown | **7/7 killed** |
| Finding-4 widened assertions | **5/5 killed** |
| Guard-6 hint | **2/2 killed** |
| `_visible_scopes` resets | **0/5 killed — equivalent, recorded** |

**M18 specifically** (dedup keyed on the whole row rather than `fields[0]`), which the auditor reported as surviving all 19 matching tests, now dies in both forms with `test_GUARD_11_the_MIGRATION_SHAPE_is_refused_not_silently_collapsed`'s own `pytest.raises(ValueError)`.

Three mutants **survived** the first pushdown sweep and are the reason `TestTheLoaderItselfTakesTheAllowlist` exists: an empty allowlist read as unrestricted, a denied scope's *name* registered before the skip, and a directory name compared unfolded were each invisible through `load_store`, whose result narrowing hides them. They are real for `load_index`'s other callers. All three now die.

**Rollback guarantee re-measured.** A one-line bare token file is **byte-identical to `origin/main`** across all 7 routes (status, `X-Store-*` header set, body bytes; the gzip snapshot compared by tar member name/size/mtime, since a gzip stream embeds a wall-clock mtime). Both controls carried, because this server fail-closes to an **empty** body without `SUBSYSTEM_STORE_TRUSTED_PROXIES` + `CF-Connecting-IP` and two empty bodies compare identical:
- *positive* — all 7 bodies non-empty and carrying the fixture's nuance on both sides;
- *negative* — the same comparator sees a difference on 7/7 routes when the store changes.

**Counts.** `test_cairn_cli.py` + `test_subsystem_store_api.py`: **403 → 425 passed**. Dependent suites (`subsystem_recall/resolver/touch/audit`, `service_recon`, `handoff_doc`, `transcript_search`, …): **2162 passed**. No test skipped, disabled or weakened.

ZacxDev:
## Round 2 — the narrow entry-kind guard, plus three guard-message defects

`3b55d11f`. Five items, one of them new work on an operator ruling.

### 1. 🔴 The narrow entry-kind guard (the ruled item)

The previous round's `visible_scopes` pushdown protects a **scoped** caller.
`visible_scopes=None` skips nothing and the pod runs a **bare legacy** token,
which is unrestricted — so on the deployed configuration the residual was fully
open. Reproduced against the branch tip before this change: unrestricted token, a
dangling `.#lock.md` in `bravo/`, `GET /api/v1/recall/alpha` → **503** naming the
file and its scope; and with a FIFO named `*.md`, the reader **blocks in `open()`
and never returns**.

`subsystem_resolver._LOADER_ENTRY_ACTIONS` now decides each `*.md` candidate's
kind before it is opened, and refuses **exactly two**:

| kind | why |
|---|---|
| `broken-link` | a dangling symlink — the Emacs `.#entry.md` lock shape; `glob("*.md")` matches a leading dot |
| `other` | fifo / socket / device — the thread-hang |

A refusal goes through `on_malformed` like every other unusable entry: a
`MalformedEntry` row under `COLLECT`, a `MalformedEntryError` under `RAISE`
(the writer's probe must not act on a store it read only part of). Never a
`SKIP` — a dropped entry is indistinguishable from one nobody ever wrote.

**Narrow means narrow.** `link-to-file` stays `TAKE`, and
`test_a_SYMLINKED_entry_is_STILL_READ_the_guard_is_NOT_the_broad_one` is the
test that kills the broad form — measured, flipping that one cell to `REFUSE`
turns a symlinked entry into a malformed row and fails there.

The **classifier** moved `server.py` → `subsystem_resolver` rather than being
copied; the **action tables** did not, because an action is a property of the
context. `TestClassifierIsTotal` now pins three columns plus a **ledger** that
fails if a fourth action table appears anywhere in either module.

Proof of the symptom, driven through the server on the live credential shape:
unrestricted token + dangling lock + FIFO → `recall/alpha` is **200**, the
caller's own content is intact, and **both** hostile files are surfaced as
malformed rows rather than dropped.

**Residuals, each with its own named test:**
- a `chmod 000` **regular** file still raises for an unrestricted caller — that
  is the four-state rule ("the store was not fully READ" ≠ "this entry is
  malformed"), and `test_an_UNREADABLE_REGULAR_FILE_still_RAISES_and_THAT_is_the_residual`
  is the honest record;
- `link-to-other` (a symlink **pointing at** a fifo) is the same hang in another
  shape, left `TAKE` because the ruling named `other` and `broken-link`. Closing
  it is one dict cell. `test_a_TAKEN_kind_still_HANGS_which_is_why_the_REFUSE_cells_exist`
  proves it is still live **and** is the positive control for the deadline
  probe, whose every other assertion is now "it did NOT hang".

`test_the_UNRESTRICTED_reading_is_UNCHANGED_and_that_is_the_residual` is
**inverted and renamed**, not deleted. The FIFO test's negative control ("the
unrestricted read must still hang") is inverted for the same reason.

### 2. 🟡 Row indices claimed physical lines; they were ordinals over non-blank rows

Reproduced on a six-line file with rows on lines 2/4/6: guard 12 said
`token rows 1 and 3 both claim it` for a clash on lines **2 and 6**. "The
operator can find the line" is the entire justification for carrying a
pre-collapse index, so the claim was load-bearing and false.

Guards 5-12 now carry the real line number and the physical line count —
`token on line 4 of 4`, `duplicate token on lines 2 and 6`. Guard 5's wording
changed, so the docstring's "guards 1-5 are UNCHANGED in wording" is corrected
rather than left standing. `test_GUARD_12_names_the_PHYSICAL_rows_across_a_collapse`
had no blank line in its fixture; it now has siblings that do, one per guard.

### 3. 🟢 `MAX_TOKENS` counted rows guard 11 would collapse

Measured: 4 distinct tokens + 1 verbatim duplicate → `too many tokens: 5, max 4`,
for a file holding **4 credentials that loaded fine before this branch**; five
copies of one token said the same for **one**. Guard 4 counts **distinct**
tokens, which is exactly the post-collapse count (the collapse keys on
`record.token`, which *is* `fields[0]`), so it keeps its place ahead of the
parse. README corrected.

### 4. 🟢 The collapse treated scope-list ORDER as a disagreement

`<tok> zach alpha,beta` + `<tok> zach beta,alpha` was refused as "two different
authorities". `_authority_key` compares `frozenset(scopes)` now. Upper bound
tested: `alpha,beta` vs `alpha` is still two authorities.

⚠ Equivalent mutant recorded at `_authority_key`: folding `None` into the empty
frozenset **survives** — unreachable, because guard 8 refuses a mapped row
claiming `legacy` and guard 9 refuses an empty allowlist. The test whose name
suggested it pinned that asymmetry now says plainly that it does not.

### 5. 🟢 Stale module API summary

`load_index`'s `visible_scopes` kwarg — the point of the previous commit — was
missing from `subsystem_resolver`'s own contract summary. Added, with
`visible_scope_set`, `classify_path` and `action_for`.

---

### Verification

- `scripts/run-tests.sh .` — **collected=17324 passed=17322 skipped=2 failed=0**
  (both skips pinned).
- `test_cairn_cli.py` + `test_subsystem_store_api.py`: **425 → 448**.
- **Mutation:** 21 semantic mutants (swapped operands, inverted branches,
  off-by-one on the line number *and* on the total, stale table cells, an inert
  policy check, a constant-valued credential count) — **20 killed, 1 recorded as
  equivalent**. Plus 3 structural mutants against the new seam guards, all
  killed; the classifier-copy one dies with that guard's own message.
  Run under `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged between mutants,
  every file restored by sha256-verified copy-back. Mutants that re-open the
  FIFO hazard were run under a wall-clock deadline in a child process.
- **Rollback guarantee re-confirmed:** a one-line bare-token file still loads as
  `('legacy', None, is_legacy=True)` — UNRESTRICTED — with the loud banner.
- **Routes vs `origin/main` (d3875d64), bare legacy token, measured:**

  | store | `/recall` | `/search` | `/snapshot` |
  |---|---|---|---|
  | clean | byte-identical modulo the store-root line | byte-identical modulo the store-root line | byte-identical |
  | hostile (dangling `.md` link + FIFO) | **503 → 200** | **503 → 200** | byte-identical (503 both — `/snapshot` already refused by kind) |

ZacxDev:
## `KIND_LINK_TO_OTHER` closed — it was the same defect, not a lesser one

Pushed `a6b70649`. The residual the last round named and deferred is now refused.

### Why it moved

Measured on `3b55d11f`, not reasoned about:

```
store/bravo/link-to-fifo.md  ->  /tmp/thefifo   (a real fifo)
GET /api/v1/recall/alpha  with an UNRESTRICTED (legacy) token
-> TIMED OUT at 25s, request thread WEDGED (curl rc=124)
   positive control: /healthz 200 throughout, so the process was UP
```

`open()` blocks identically whether the fifo is named directly or reached
through a symlink, and the link sat in a scope the caller never asked for. On
`replicas: 1` / `strategy: Recreate` a wedged thread is the worst failure mode
in this area.

**Reproduced here against `origin/main` on a hostile store**: `/healthz` 200,
then `/api/v1/recall/kelp-forest` never returned (client read timeout). Same
store on this branch: 200 promptly, the caller's own content served, and
`quartz-mine/link-to-fifo.md` reported as a named malformed row.

### Scope

The production diff is **one cell plus its refusal sentence**:

```diff
-    KIND_LINK_TO_OTHER: TAKE,
+    KIND_LINK_TO_OTHER: REFUSE,
```

`KIND_LINK_TO_FILE` stays `TAKE` — that is the cell the narrow-vs-broad ruling
was actually about, and a `<scope>/<slug>.md -> ../shared/<slug>.md` symlink is
still read. `/snapshot`'s `_ENTRY_ACTIONS` is untouched and `server.py` is not
in this diff (verified byte-identical between the two trees). The refusal goes
through `on_malformed` exactly as its siblings do — `MalformedEntry` under
COLLECT, `MalformedEntryError` under RAISE, never a silent skip.

### red -> green

| test | at `KIND_LINK_TO_OTHER: TAKE` | at `REFUSE` |
|---|---|---|
| `test_the_decision_table_is_pinned[KIND_LINK_TO_OTHER-…]` | **FAIL** (`take != refuse`) | pass |
| `test_every_REFUSED_loader_kind_HAS_a_reason_and_no_other_kind_does` (new) | **FAIL** (ledger drift) | pass |
| `test_a_SYMLINK_to_a_FIFO_no_longer_HANGS_and_the_DEADLINE_still_SEES_one` | **FAIL — probe HUNG**, blew the 30s deadline | pass |

Run time on that last one went 40.9s -> 11.2s, which is the hang disappearing.

### The inverted test, and the hole it left

`test_a_TAKEN_kind_still_HANGS_which_is_why_the_REFUSE_cells_exist` pinned this
kind as an *accepted* residual. Inverted and renamed rather than deleted — kept
as-is it would have been a test pinning the bug.

But it was also **the positive control** proving the deadline machinery can
observe a hang at all, and no remaining `TAKE` cell blocks, so it could not
simply be re-aimed at another cell. The control is now a bare `open()` of the
same fifo — the syscall in question rather than a proxy for it — asserted
**first**, so a fixture that made no real fifo fails there before any "it did
not hang" assertion can read as a green.

### New seam guard

`load_index` subscripts `_LOADER_REFUSAL_REASON[kind]` unguarded. Flipping a
cell to REFUSE without writing its sentence turns a hostile entry into a
`KeyError` — a 500 with no `X-Store-Status`, for the shape whose whole fix was
to make it a *named* row. That is precisely how this cell would have landed, and
reviewing the action table alone cannot see it. Asserted as set **equality**, so
it fails when the ledger grows as well as shrinks.

### Mutation results

Both killed by the named test's own assertion, `PYTHONDONTWRITEBYTECODE=1`,
`__pycache__` purged per mutant, restored by sha256-verified copy-back.

- **(a) cell back to `TAKE`**, reason left in place so the mutation is isolated
  to the action — re-hangs, killed by the deadline (11s -> 40.6s).
- **(b) `KIND_LINK_TO_FILE: REFUSE`** (the over-broad form), given a reason so
  the guard actually executes rather than dying on a `KeyError` — killed by
  `test_a_SYMLINKED_entry_is_STILL_READ_the_guard_is_NOT_the_broad_one` on its
  own `index.malformed == ()`, with the mutant's reason string in the failure
  diff. Died by that assertion, not incidentally.

### Rollback guarantee — clean and hostile stated separately

**CLEAN store**, same store path and interpreter, branch vs `origin/main`
(`d3875d64`): all **7 routes identical** in status code, body sha256 and header
set (`/healthz`, three `/recall` shapes, `/search`, `/snapshot`, a 404). A
one-line bare token still loads `(token, 'legacy', None)` — unrestricted. The
only deltas are the audit line's additive `identity=legacy` field and
`TokenRecord` replacing a bare `str`, both this PR's own intended shape.
Positive control carried throughout: every route asserted non-empty and the
store's own content asserted present, because without
`SUBSYSTEM_STORE_TRUSTED_PROXIES` and `CF-Connecting-IP` two empty responses
diff clean.

**HOSTILE store**: `origin/main` wedges; this branch does not.

### Counts

`449 passed` on `test_cairn_cli.py` + `test_subsystem_store_api.py` against the
448 baseline — +1 is the new seam guard (the inversion replaced a test, net 0).
`test_subsystem_recall.py`: 450 passed.

### Not verified

- Not deployed, not merged, no image bump.
- The 25s live wedge is quoted from the operator's measurement on `3b55d11f`;
  what I reproduced independently is the same shape against `origin/main`
  through the real server on a synthetic store (client read timeout, `/healthz`
  200 first).
- `/snapshot` answers **503 store-unreachable** on the hostile store on this
  branch. That is `_ENTRY_ACTIONS`' own pre-existing REFUSE for this kind, not
  something this change introduced — but I could not observe `origin/main`'s
  `/snapshot` on the same store for a direct comparison, because it wedges on an
  earlier route. The two tables are byte-identical for that cell.

ZacxDev:
## Round 3 — `28db471c`

Five findings addressed, plus the mid-round ruling on the two directory cells.
One of the five was a **live defect the documentation actively denied**, and one
was a test that **hung the runner instead of failing it**.

---

### 🔴 The defect (finding B, escalated by the ruling)

`_LOADER_ENTRY_ACTIONS` took `directory` and `link-to-dir`. `read_text` raises
`IsADirectoryError` on both, an `OSError` fails closed, so **one stray
`mkdir <scope>/<slug>.md` — or an rsync/restore artefact — took `/recall` and
`/search` down for every caller.** Reproduced with a paired control:

```
store/beta/notes.md created as a DIRECTORY   (nobody asks for beta)
GET /api/v1/recall/alpha, UNRESTRICTED legacy token
  -> 503 "index entry unreadable: … (IsADirectoryError: … /beta/notes.md)"
CONTROL, same store with a dangling .#lock.md instead      -> 200
```

Both cells are now `REFUSE`, **on the criterion the original narrow ruling used
and not a widening of it**: this loader has never once successfully *read* a
directory, only ever raised on one. `/snapshot`'s `_ENTRY_ACTIONS` is untouched
and byte-identical. `link-to-file`, `regular-file` and `indeterminate` stay
`TAKE`.

### 🔴 The ledger was false in two documents, and nothing could see it

Both `load_index`'s docstring and the API README called `chmod 000` "the only
residual left" for three rounds while **four** `TAKE` cells could still 503 the
store. The README additionally still said a symlink-to-fifo was "left open" a
whole round after `a6b70649` closed it — so that commit's "comments and
docstrings updated wherever they still recorded the hang as open" was **false**.
Sites checked by grep this round, each stated: `load_index.__doc__`, the
`_LOADER_ENTRY_ACTIONS` comment block, `_LOADER_REFUSAL_REASON`, `README.md`,
`server.py`'s classifier note, `TestTheLoaderRefusesHostileEntriesByKind`,
`TestUnreadableEntriesInDeniedScopes`, `DECISION_TABLE`'s comments.

Both ledgers are now **machine-checked against the table in both directions** —
a `TAKE` cell missing from a ledger fails, and a `REFUSE` cell still listed as a
residual fails. That second direction is exactly the drift that happened.

### ⚠ The post-flip residual set is THREE shapes, not one — reported, not quietly enlarged

You named `chmod 000`. Measured on this tree, two more 503 identically and are
now listed rather than added silently:

| kind | shape | status |
|---|---|---|
| `regular-file` | `chmod 000` | the one you named |
| `link-to-file` | symlink whose **target** is unreadable | **not previously named** — the same shape through a link, exactly as `other`/`link-to-other` were |
| `indeterminate` | the `lstat` itself failed | **not previously named** — deliberately `TAKE` per your ruling |
| `absent` | vanished between `glob()` and classify | TOCTOU race, **reasoned from the table, not reproduced** |

No behaviour changed for any of them.

### 🔴 Finding E — the three FIFO tests

With `_LOADER_ENTRY_ACTIONS[KIND_OTHER]` reverted to `TAKE`,
`test_the_REFUSED_entries_are_SURFACED_not_silently_dropped` ran **>600 s with
no error and no failure**. All three now route through one module-level
`under_deadline` subprocess helper (it was a closure duplicated in two tests and
simply *missing* from a third); the server one uses `running_subprocess`, the
real spawned entrypoint. Re-measured under the same revert: **the run completes
in 74.5 s with 3 explicit FAILURES.**

`test_a_BOGUS_policy_is_still_a_ValueError` stays in-process for a *structural*
reason stated in the test — `_check_on_malformed` is the first statement of
`load_index`, so it raises before `iterdir()`.

### 🟢 C — the cap fixture · 🟢 D — the spelled seam guard · `census()`

- Cap fixture is now **7 distinct tokens + 1 repeat**: 7 credentials, 8 rows,
  cap 4 — every number in the sentence distinct.
- The table ledger now selects on **what the dict is** (keys are kinds *and*
  values are actions), not on a name suffix. Both halves matter: keys alone also
  matches `_LOADER_REFUSAL_REASON`.
- `subsystem_touch.census()` is documented as the **third** unguarded
  `glob("*.md")` + `read_text` site, unguarded **by ruling**; the two "the two
  sites are…" comments now say "the two sites *guarded by this classifier*".
  Behaviour untouched. `validate_scope` globs but reads nothing, so it inherits
  the loader's guard.

### Six test fixtures had gone vacuous

Four in `test_subsystem_recall.py`, one in `test_subsystem_resolver.py`, one in
`test_subsystem_touch.py` used `(scope / "broken.md").mkdir()` to provoke an
`OSError` — which a directory no longer produces. They were asserting a raise
that no longer happens. Each moved to a shape that is **still a residual and
still not a permission change** a root-run sandbox would bypass (which is what
those fixtures' own comments said they were avoiding): a symlink whose target
resolution fails with `ENOTDIR`, i.e. `indeterminate`. Verified as euid 1000.
Each carries a fixture self-check on the planted kind.

---

### Mutation matrix

Every mutant killed by the **intended test's own assertion**, never a
`KeyError`. `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged per mutant, every
file restored by sha256-verified copy-back.

| mutant | result |
|---|---|
| `KIND_DIRECTORY -> TAKE` | **KILLED** — `…no_longer_503s_the_WHOLE_store[dir]` on `assert code == 200` (got 503, `IsADirectoryError`). `[linkdir]` survived → isolated to the cell |
| `KIND_LINK_TO_DIR -> TAKE` | **KILLED** — the `[linkdir]` arms of both directory tests; `[dir]` survived |
| `KIND_LINK_TO_FILE -> REFUSE` (+ a reason so the guard executes) | **KILLED** — `…the_guard_is_NOT_the_broad_one` on its own `index.malformed == ()`, mutant's reason string in the diff. Narrow guarantee still pinned after the widening |
| `KIND_OTHER -> TAKE` | **KILLED** — 3 tests FAIL and the run **completes** (74.5 s) where the same revert previously hung >600 s |
| hardcoded `5` in the cap message | **KILLED** — `test_the_CAP_still_counts_the_DISTINCT_ones_and_says_so`, over the **full 407-test file** (1 failed, 406 passed). Previously survived a 93-test subset with 0 failures |
| `_FOURTH_TABLE_ACTIONS` | **KILLED** — the table ledger |
| `_SNAPSHOT_KIND_POLICY` | **KILLED** — the differently-spelled control that used to **SURVIVE** |
| README ledger drops `link-to-file` | **KILLED** — `RESIDUAL_LEDGER[README.md]` |
| README ledger keeps `directory` as a residual | **KILLED** — the exact "closed hazard recorded as open" drift |
| docstring ledger drops `link-to-file` | **KILLED** — `RESIDUAL_LEDGER[subsystem_resolver.load_index]` only |
| README refused table drops `directory` | **KILLED** — `README_REFUSED_TABLE` |
| README marker deleted (negative control) | **KILLED** — both guards, on "no RESIDUAL LEDGER marker to split on"; they cannot silently pass on an empty region |

### Red→green

| test | red at | green at |
|---|---|---|
| `test_a_DIRECTORY_named_md_no_longer_503s_the_WHOLE_store[dir]` | pre-flip table (503) | HEAD (200) |
| `…[linkdir]` | pre-flip table (503) | HEAD (200) |
| `test_the_REFUSED_DIRECTORY_is_a_NAMED_row_not_a_silent_skip[dir]`/`[linkdir]` | pre-flip table (`EntryUnreadableError`) | HEAD |
| the three FIFO tests | `KIND_OTHER: TAKE` — **hung** before, **fail** now | HEAD |

**Labelled as invariant / documentation guards, not regression coverage:**
`test_the_LOADER_RESIDUAL_SET_is_pinned`, the two `RESIDUAL_LEDGER` guards, the
`README_REFUSED_TABLE` guard, the strengthened cap test and the strengthened
table ledger. At `a6b70649` the ledger guards are red only because the marker
did not exist yet — a structural red that proves nothing. Their content
sensitivity is established by the four drift mutants above instead.

### Rollback guarantee

Clean store, branch vs `origin/main` (`a7cecd82`), same store path and
interpreter, both driven as the **real spawned entrypoint**: all 7 routes
identical in status, body and header set. A one-line bare token still loads
unrestricted — `(token, 'legacy', None)` here, a bare `str` on main, same
authority. Negative control: one character changed in one entry moves **5 of 7**
routes.

⚠ **Correction to a previous round's evidence.** `/api/v1/snapshot`'s body is a
gzip stream and is **not byte-stable across two runs of the same server** —
measured, and it reproduces intermittently (it depends which wall-clock second
each run lands in). "Identical body sha256 on all 7 routes" was never a claim
that route could support. It is compared here by its **extracted tree**
(identical), with self-determinism checked first so the comparator cannot report
the encoder as a behaviour delta.

### Counts

- `test_cairn_cli.py` + `test_subsystem_store_api.py` — **457 passed** (baseline 449, +8 new)
- `test_subsystem_recall.py` — **450 passed** (baseline 450)
- `test_subsystem_{resolver,touch,audit}.py`, `test_service_recon.py`, `test_store_content_not_copied.py`, `test_resume_state_handoff_resolution.py` — **1490 passed**
- `scripts/run-tests.sh .` — **RESULT: PASS (exit=0)**

### NOT verified

- The `absent` residual (a TOCTOU race) is **reasoned from the table, not reproduced**.
- Nothing here was run against the **deployed pod** or a **real store** — every fixture is synthetic under `tmp_path`.
- `census()`'s fifo hang is **documented and left live by ruling**; not measured this round.
- No image built, no version bumped, nothing merged or deployed.

ZacxDev:
## Round 4 — both findings fixed at `3116812b`

Test/documentation only. **No production behaviour changed**: `git diff --stat` is two test files plus `load_index`'s *docstring* — no entry-kind action cell, no executable line in `subsystem_resolver.py`.

### 🟡 1 — the seventh `mkdir` site, and the kill it silently vacated

`test_kills_the_UNREADABLE_wrap_in_build_report` still planted `(store / SCOPE / "half-written.md").mkdir()` — same file, same filename, 276 lines below the sibling that was migrated. `_LOADER_ENTRY_ACTIONS` refuses `directory` before `open()`, so `load_index` raised `MalformedEntryError` and the `except OSError -> EntryUnreadableError` clause **this test mutates** was never reached; mutant and pristine raised the same error and both assertions passed either way.

Migrated to the same ENOTDIR symlink the six siblings use (`indeterminate` — privilege-independent, where a `chmod 000` would be bypassed by a root-run sandbox), with the planted kind **asserted** rather than assumed, and an added anti-vacuum assertion pinning the mutant's own `neutered:` `RuntimeError` — so a future refusal that moves this shape back onto the malformed path fails here instead of going quiet again.

**Kill restored, proven by the paired identity control** (mutant == original):

| | identity replacement |
|---|---|
| before | `1 passed` — insensitive |
| after | `1 failed` at `assert not isinstance(exc.value, st.EntryUnreadableError)` |

The red is on **this test's own first assertion**, raised as `EntryUnreadableError` out of `build_report` — not a `KeyError`, not a mutant-load failure, not a sibling's error.

### 🟡 2 — the ledger guards' docstring overclaimed

It said "`load_index`'s docstring prose is covered by the ledger half above". It was not: the ledger guard reads only the span between `RESIDUAL LEDGER` and its END marker, and the refusal sentence lives above it. `resolver.load_index.__doc__` appeared **exactly once** in the whole test tree, inside `_residual_ledger`.

The sentence now names its kinds as backticked tokens inside literal markers, and a new `test_the_DOCSTRING_REFUSED_SET_names_every_REFUSE_kind_and_no_other` pins the marked span against `_LOADER_ENTRY_ACTIONS` in **both** directions. The span is delimited tightly on purpose — the paragraph *below* it names `link-to-other`, `directory` and `link-to-dir` while recounting which round each shipped in, so a docstring-wide guard would have passed on all three while the refusal sentence said nothing.

**Required control re-run** — drop those three kinds from the sentence:

| | result |
|---|---|
| before | **407 passed**, nothing red |
| after | `AssertionError: `load_index`'s REFUSED SET does not name `directory`, which the loader refuses — the docstring claims a narrower guard than the code has` |

Reverse direction also controlled: naming `regular-file` inside the span reds with ``REFUSED SET names `regular-file`, which the loader does not REFUSE (it is 'take') — it claims a guard that is not there``.

### 🟢 3 — done, not skipped

`head = text[:cut]` narrowed to the refusal table itself (its own header line through the blank line that ends it), via a shared `_marked_region` helper that asserts rather than yielding an empty or implausibly short slice.

Controlled both ways: with an unrelated `| \`regular-file\` |` table planted before the ledger marker, the **old** scoping went red with the misleading `README refusal table lists \`regular-file\``; the **new** scoping is green. And the narrowing did not make it inert — deleting the real `link-to-dir` row still reds with `README refusal table does not list \`link-to-dir\``.

### Suites

| suite | baseline | now |
|---|---|---|
| `test_cairn_cli.py` + `test_subsystem_store_api.py` | 457 | **458** (one new guard) |
| `test_subsystem_recall.py` | 450 | **450** |
| `test_subsystem_resolver.py` | 321 | **321** |
| `test_subsystem_touch.py` | 865 | **865** |

All under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` purged between mutants and every temporary edit restored by sha256-verified copy-back.

### Sweep

A GNU-grep (not gitignore-blind) sweep of `*.md`-named `mkdir` across the tree finds **no eighth defective site**. The four named as legitimate are confirmed untouched (`test_subsystem_recall.py:424`, `:4127`, `test_subsystem_touch.py:7592`, `test_cairn_cli.py:717`). One site outside the subsystem store — `scripts/opencode/tests/test_dispatch.py:961`, `(tmp_path / "SKILL.md").mkdir()` — is unrelated and legitimate: the directory *is* the hazard under test there (`is_file()` vs `exists()`), nothing to do with the loader's entry-kind table.

### NOT verified

- No live/pod verification — nothing deployed, no image bump, no merge.
- The other three suites were re-run whole but not re-mutation-swept; only the one mutant this round touches was re-controlled.
- The `absent` cell remains reasoned-from-the-table, not reproduced (unchanged from before).

PR #922, opened 2026-08-27
fix(audit-pr): the two worktree hazards the ladder deliberately left open, and the counts round 10 filed

ZacxDev:
Follow-up to #900. That PR disclosed two 🔴 in this skill and deliberately scoped them out so the ladder could stay on the attribution gate; round 10 then filed three comment counts rather than spending a round on them, which is what the gate it added instructs. This closes all five, and does the structural fix #900 kept deferring.

## The two hazards, both silent

**1. `isolation: "worktree"` worktrees the CWD's repo, not the PR's.** The skill's primary target is a bare PR number, which is very often a PR in *another* repo — and then every fan-out auditor gets a worktree of whatever repo the operator happened to be sitting in. One failure half is loud (the agent reports a briefed file missing); the other is silent (it audits the wrong tree and says nothing). Now: for a PR in another repo, have the agent run `git -C <that-repo> worktree add …` itself.

**2. A `cp -a` of a worktree shares the ORIGINAL's git dir.** The skill tells the auditor to mutate only in a `cp -a` copy — and a worktree's `.git` is a *file* holding `gitdir: …`, so a commit in that copy lands on the real branch. Now: `rm -f <copy>/.git` first.

Both pinned. **G2 was green until it was pinned** — inverting the caveat to "use `isolation: "worktree"` for any repo" passed the suite.

## The three counts round 10 filed

- the env-brief comment said "four of this PR's nine commits"; measured **six of eleven**, or four of the eight *fix* commits — the denominator the neighbouring comment uses. Stated both ways now, so neither reading is wrong.
- "the other two regression shapes" is **three**.
- the matrix header said "two pins that certified presence"; it was **one pin plus one block with no pin at all** — different lessons, so the header says so.
- and the renamed test's docstring still said "neither"/"either" for a body asserting six constants.

## The structural fix

#900's last four rounds each paid for a new rule by shaving words out of the checklist, and one of those shaves deleted `rollback` under a commit message claiming a restore. This one **stops shaving and re-homes**: the pathspec counter-evidence (a measurement) and the two mutation rules that bind a guard's *author* rather than a PR's *reader* move to the reference file, which is not budgeted.

**Body 11,985 → 11,902 B — the first commit in this thread to add two rules and come out smaller.** The moved measurement is pinned at its new home; the prohibition it supports stays pinned in the body.

## Coverage

```
POS  unmutated .............................. 11 passed
PREV origin/main (pre-change payload) ....... 3 failed, 8 passed
G1   cp -a .git hazard deleted ............... 1 failed
G2   cross-repo worktree caveat inverted ..... 1 failed   <- green until pinned
G3   pathspec measurement inverted at its new home  1 failed
G4   pathspec prohibition -> endorsement ..... 1 failed
```

Each on its own scratch tree, `PYTHONDONTWRITEBYTECODE=1`, cache disabled, assert-before-edit. Two pins were re-pointed here because their sentences changed deliberately — the module's documented path, and the moment the suite made me notice I was rewriting a rule rather than reformatting a paragraph.

Gate: `scripts/gate.sh --tier both` → `GATE: RESULT=PASS exit=0`, on a branch cut from `6d6bf84d`.

## Still open, deliberately

The **dispatch template** — the skill still has no executable procedure, so a caller reassembles the auditor brief from prose each time. That is the largest remaining usability gap and it is a different change from this one.

### DIFF
--- a/claude/skills/audit-pr/SKILL.md
+++ b/claude/skills/audit-pr/SKILL.md
@@ -13,7 +13,9 @@ Case histories and the measurements behind the ladder rules:
 Target: `$ARGUMENTS`:
 - A number → that GitHub PR (`gh pr diff <n>`, `gh pr view <n>`).
 - `current` / empty → the current branch's diff vs its base/trunk.
-- Several numbers → audit each, one subagent per PR **with `isolation: "worktree"`** so they don't collide.
+- Several numbers → audit each, one subagent per PR so they don't collide. 🔴 `isolation:
+  "worktree"` worktrees the **cwd's** repo, not the PR's — for a PR in another repo have the
+  agent run `git -C <that-repo> worktree add …` itself.
 
 ## What to do
 
@@ -27,7 +29,9 @@ not a working checkout, and an auditor hitting this cold blames the PR. Whicheve
 **monorepo `node_modules`** may need linking per package, not just at the root; **whether the base
 branch is already red** and *at which file*; and that **zsh does not word-split unquoted
 parameters**, so `eslint $FILES` checks **zero** files and prints a confident PASS. Have it mutate
-only in a `cp -a` copy, and verify your worktree clean yourself at the end.
+only in a `cp -a` copy — **`rm -f <copy>/.git` first**, since a worktree's is a FILE pointing at the
+real git dir, so a commit in the copy lands on your branch — and verify your worktree clean
+yourself at the end.
 
 **Audit for:**
 1. **Risks** — what breaks in production.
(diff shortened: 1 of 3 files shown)


ZacxDev:
Merged-tree gate (`strict` is off here, so the branch checks are not a claim about the tree the merge creates):

`main` moved to `d3875d64` after this branch was cut. Built an integration branch off it, merged `ae85204f` (clean, no conflict), and ran both tiers there:

- dev host: `scripts/gate.sh --tier both` → `GATE: RESULT=PASS exit=0`
- sandbox tier, the one Tekton gates on: `nix build .#checks.x86_64-linux.pytests` and `.nodetests` both built

Branch checks `tekton/devrc-pytests` and `tekton/devrc-nodetests` are SUCCESS. Merging.

PR #924, opened 2026-08-27
fix(prune-skill): the skill now meets its own bar — and the auditor stops asserting devrc's budget for other repos

ZacxDev:
Evaluated `prune-skill` against itself, after driving it end-to-end on a real prune (talos-infra `app-blocks`, 17,191 → 12,035 B). Three defects, all self-referential.

## 1. It was the worst-demoted skill by its own headline metric

Its rule: *"a low core-to-reference ratio means you have not demoted enough — that is the tell, not the byte count."*

| skill | core | reference | ratio |
|---|---|---|---|
| `browser-bridge` (its own exemplar) | 11,961 B | 162,076 B | **13.55x** |
| `prune-skill` (before) | 12,812 B | 22,651 B | **1.77x** |

It was also 524 B over the 12,288 B target it enforces on others — disclosed and pinned at 13,056, but institutionalised.

Demoting the budget **rationale** and §3's classification **sub-rules** into two new sidecars takes the core to **12,027 B**, inside the shared 12,038 B enforced budget for the first time. `MAX_BYTES` drops 13,056 → 12,288 and the "why the ceiling is above the target" section is retired; that module's docstring already named lowering as the intended direction.

🔴 **But the ratio rule was itself overstated.** Its floor is *shape-dependent*: `browser-bridge` is a tool with a wide surface, so nearly everything is demand-loaded depth; `prune-skill` is a seven-step method whose steps are what a reader needs in hand, and §3's own `KEEP_HOT` protects exactly that. Driving this file to 11x would demote the method and leave a router pointing at the procedure it is supposed to *be*. `budgets-and-scope.md` now says so with the measurement, and reframes the ratio as a tell that points at candidates rather than a target.

## 2. `skill-audit.py` asserted devrc's budget for every repo

It hardcodes `_BUDGET_SOURCE` to browser-bridge's gate and printed **"12,038 B ENFORCED"** for talos-infra skills too — where nothing enforces it. Gate 11 there is a ratchet with a per-push allowance that can pass a body this tool calls RED and block one it calls fine. The skill's own first instruction is *"check what governs THIS file first"*, and the tool contradicted it on every run; I optimised `app-blocks` to the wrong number yesterday because of it.

Deliberately **not** fixed with a table of per-repo budgets — that is the duplicated constant `_load_budget` exists to prevent, one directory further out. It now detects the audited tree's repo, labels the budget `devrc's DEFAULT (not enforced here)`, and points at that repo's own gate. Section weights, fat lines, reference integrity and corpus checks are repo-independent and still stand as printed.

## 3. A dead path, in the paragraph that says nothing else will catch it

§4 told authors to write `.../scripts/dl-router/reference/`, adding *"THE GATE CANNOT SEE THIS … so this line is the only guard."* That directory has never existed — `dl-router`'s sidecar is one 47 KB `README.md`. The **rule** (mkOutOfStoreSymlink skills need source-tree paths) is correct and load-bearing; the **illustration** was invented.

That is this skill's own *"an EXAMPLE earns its own re-measurement"* rule from `verification-protocol.md` §6, missed in the one place it declared unguardable. Both sidecar shapes are now named, with "look before you write the path".

## Verified — using the skill's own §7 protocol on itself

- **12,027 B**; `skill-audit`: within budget, 5 references resolve, no orphans.
- **Content survival**, original core vs new core + all 5 sidecars, 6 populations: **numbers 0 lost, refs 0 lost**, 6 total — each read individually and intentional (2 are the dead `dl-router` path, 2 are case changes from one reword, 2 are reworded lines whose claims are in the new core).
- **Controls**, reporting the number each produced: hiding `budgets-and-scope` **6 → 22**, hiding `classification-rules` **6 → 22**. Hiding `verification-protocol` stayed at 6 — stated rather than reported as clean: nothing was demoted into it this pass, so it *cannot* move; it serves as an attribution control showing the two real destinations are specifically responsible.
- **`test_prune_skill_size.py` 40/40.** It earned its keep twice: it caught that my reword left four of its own gated figures unmatched, and that the routing table had gone 3 rows/435 B → 5/722 B while the prose still said 145 B/row. Five specs retired with the exemption; the two figures still printed are re-gated, so retiring a spec never un-gates a live number.
- **227** doc-path-rot + **89** skill-audit + **88** sibling budget tests pass. doc-path-rot blocked first because the new sidecars were untracked — correct behaviour, fixed by staging them, not by relaxing it.

Backup for rollback: `/tmp/claude-1000/prune-skill-backup` (4 files).

🔴 **Not live on this host until a `home-manager switch`** — `prune-skill` deploys as a `home.file` copy (`readlink -f` lands in `/nix/store`), and the skill's own §6 says a push is not a saving.


ZacxDev:
## Pre-merge audit ladder — 8 blind rounds, now stopped

The PR description above covers the three original fixes. Eight adversarial audit rounds followed; this records what they changed, because the branch is substantially more than the opening commit.

**Ladder stopped at `f4b1c8b7`** on the attribution rule: that round changed **zero payload lines** (`scripts/skill-audit.py` untouched — test and docstring only), and I verified the two mutants the last audit named as survivors myself; both now die on the test written for each. Continuing would be auditing scaffolding the ladder had just written.

### What the audits actually caught

Two findings were the PR's own headline claim being unguarded rather than wrong:

- **The "exemption is retired" claim was walkable.** `MAX_BYTES = 12_288` was a bare literal, so setting it to `13_000` plus four consistent prose edits returned `40 passed` while the docstring still said the exemption was gone. Every figure spec re-derives *from* `MAX_BYTES`, so none could see it. Now asserted as the claim (`MAX_BYTES != target`), not its spelling.
- **A verbatim slice carried a sentence the prune falsified** — "This skill is itself over target, pinned by a ratchet" — and it was the only surviving pointer to the gate.

Then a recurring pattern, worth naming because it is the ladder's whole justification: **six of the eight rounds introduced a defect in the class they were fixing.** Examples — a fix for "a mixed run names only the foreign repo" that stamped `[ungoverned]` on the *governed* skill; a three-valued `_is_governed` whose docstring said "callers must not collapse it to either bool", collapsed at both consumers in the same commit; a guard against pairwise-identical fixtures, re-created in the guard written to close it.

The durable fix was **consolidation, not another patch**: four sites independently answered "who does this budget govern?", and successive rounds fixed subsets while moving the rest. `_governance_scope()` now owns it.

### Instrument failures, recorded because they were nearly believed

- An auditor's **977-failure report was retracted** — bare `pytest` outside the nix devShell trips the repo's own tool gate. The same command at base gave 72. Neither number was evidence.
- An auditor **retracted a "13.55x is unreproducible" finding** — its glob missed `reference/sites/`. The figure was exact; what was missing was the *stated method*, which is the real fix now in the file.
- **Three of my own mutants were invalid**: two silently failed to apply (one left an unclosed paren), and one was a logically-identical rewrite. All three read as SURVIVED. The sweep is now a script asserting *applied → parses → not-a-no-op* before scoring.
- One full-suite run was **void**: I edited files while it ran, so it graded a tree that changed underneath it.

### State at `f4b1c8b7`

- Authoritative runner, clean pinned worktree, 0 dirty paths: **17,645 passed / 2 skipped / 0 failed** across 36 targets (floor 15,970) at the previous head; re-run at this one.
- 294 tests across the seven affected suites; 116 in `test_skill_audit.py` (89 at branch point).
- `prune-skill/SKILL.md` 12,812 → 11,841 B, 197 B under the binding limit.

### Left open, deliberately

The run-scope phrases keep the locative "here" when reused on the over-budget subset, so a mixed-regime verdict can read "(not enforced here)" on a run where most skills are enforced. Not a contradiction — the flat claim entails the hedged one, and the preceding clause already says which skills are meant. It is a wording judgement, and the closing condition is a human one: **@ZacxDev either merges as-is or asks for the reword.**

🔴 Not live on this host until a `home-manager switch` — `prune-skill` deploys as a `home.file` copy.

PR #933, opened 2026-08-27
feat(audit-pr): check in the two instruments #900 ran only in a scratchpad — and one of its published numbers was the wrong metric

ZacxDev:
Third and last follow-up to #900. That PR recorded ~30 mutation rows in `test_audit_ladder_stop_rule.py` and a 110-session transcript sweep in the skill's reference file — and **every one of those runs happened in a session scratchpad that no longer exists**, so not a single row could be re-checked by anyone else, including the rows that justified adding a pin. `scripts/tests/mutants-claim-work.sh` had already written that lesson down for a different file, in a comment saying its predecessor "was never committed — so nobody could re-check a single row". Same mistake, one skill over.

## Two instruments, both run, both in the tree

**`scripts/tests/mutants-audit-ladder.sh`** — 13 mutants + 2 controls over the prose guards, built to the conventions `mutants-claim-work.sh` established:

- mutates only inside a `mktemp -d` copy, and **asserts the copy carries no `.git`** — a `cp -a` of a worktree would carry its gitdir pointer and a git command inside the copy would act on the real repo, which is a hazard this very skill now warns about;
- **every mutant names the test that must kill it**, and one killed only by some *other* test reports 🔴 WRONG-KILLER rather than ok — these pins overlap by design, so "a test failed" is not evidence the intended assert ran;
- **refuses to score a mutation that did not apply** — five silently failed to apply across three of #900's rounds, caught only because the driver asserted first;
- reads the suite's *content* with a collected-test floor, never an exit code;
- **two SURVIVES controls that must kill nothing** — a re-wrap of a pinned sentence and an edit to unpinned prose. The re-wrap is the one that matters: `_norm` collapses whitespace on purpose, so a red there would mean the pins key on line breaks rather than words.

Measured: **15 rows, all as expected** — 13 killed by their named test, 0 wrong-killers, both controls survived.

**`scripts/ladder-depth-sweep.py`** — the behavioural half. The test module states plainly that it pins instructions and *cannot* observe whether any session actually stopped; this is the only thing that can answer that, by comparing ladder-depth distributions across date windows. 🔴 **It refuses to report a zero**: its first version during #900 returned a confident `0` because it filtered on tool name `Task` while the harness emits `Agent`, and a zero from a broken filter is indistinguishable from a quiet host. It now counts every dispatch it walks as a positive control and exits 2 if that is zero.

## The repo's own ledger caught my first draft

The sweep's first version hand-rolled an `os.walk`, and `test_transcript_search.py` went red — that module enumerates every `*.jsonl` walk site in the tree and treats an unlisted one as **a violation by default**, because "a walk nobody enumerated is the one that goes unnoticed."

The right answer wasn't to add myself to the ledger but to **stop being a fifth walk**. The script now goes through `scripts/lib/transcript_search.py`, which also hands it two decisions it would otherwise have re-made and could have re-made wrongly: corpus membership (`subagents/` excluded) and skipping a malformed *line* rather than the whole file. Running it then surfaced a second defect the ledger couldn't see — the lib path was `parents[1]` where it needed `parent`, so the script exited on import. Both fixed, and measured after rather than asserted.

## And running it found a published number was the wrong metric

The reference file said "110 sessions, **440 rounds**". Two counts are available and they're far apart:

| | |
|---|---|
| rounds **observed** | the distinct numbered rounds a walk matches |
| rounds **implied** | each session's deepest round summed — a ladder reaching round 8 ran eight |

**440 was implied, and the paragraph didn't say so.** Re-run with the checked-in instrument: **127 sessions, 306 observed / 541 implied**, mean deepest 4.26, 39% ≥5 — a reading that includes the ten-round ladder the file documents, which is what "the window moves" means in practice. The script now prints both, labelled, with the run date; the reference file names the metric, carries the re-measurement, and tells the reader to re-derive rather than quote.

The qualitative claim the skill body rests on is unaffected in direction or magnitude — the ladder is a dominant workflow either way.

## Gate

`scripts/gate.sh --tier both` → `GATE: RESULT=PASS exit=0`, on a branch cut from `648f08c2`. The first run of this branch was **red** (`test_transcript_search.py`), which is how the walk-site ledger did its job.

### DIFF
--- a/claude/skills/audit-pr/reference/round-ladder-evidence.md
+++ b/claude/skills/audit-pr/reference/round-ladder-evidence.md
@@ -69,7 +69,10 @@ rounds a walk matches) and rounds *implied* (each session's deepest round summed
 that a ladder reaching round 8 ran eight). Re-run 2026-08-27 with the checked-in instrument: 127
 sessions, **306 observed / 541 implied**, mean deepest 4.26, 39% ≥5 — and that reading includes the
 ten-round ladder this file documents, which is what "the window moves" means in practice. Quote one
-count, name which, and date it.
+count, name which, and date it. 🔴 **That reading was stale within hours of being written** — an
+independent re-run the same day gave 319 observed / 556 implied / mean 4.38, because the corpus
+includes the very session dispatching the ladder this file describes. A figure from this instrument
+is a reading, not a fact about the world.
 
 **Re-derive rather than quote**: `scripts/ladder-depth-sweep.py` is the instrument, it prints both
 counts with the run date, and it REFUSES to report a zero it cannot distinguish from a broken
--- a/claude/skills/audit-pr/reference/round-ladder-evidence.md
+++ b/claude/skills/audit-pr/reference/round-ladder-evidence.md
@@ -58,10 +58,22 @@ Two things this case establishes, and one it does not:
   nothing, so it is not an instance of the waste that retraction denies. Different axis.
 
 Ladder depth across all sessions, measured the same day over `~/.claude/projects/**/*.jsonl`
-(numbered delta re-audit dispatches, `subagents/` excluded): **110 sessions, 440 rounds, 84 of those
-sessions in the preceding 14 days**; mean deepest round 4.0; 34% ran ≥5 rounds; distribution
-3→38, 4→22, 5→21, 6→9, 7→4, 8→2, 10→1. The ladder is a dominant workflow, not an occasional one —
-which is what makes a per-round gate worth its bytes.
+(numbered delta re-audit dispatches, `subagents/` excluded): **110 sessions, 84 of them in the
+preceding 14 days**; mean deepest round 4.0; 34% ran ≥5 rounds; distribution 3→38, 4→22, 5→21, 6→9,
+7→4, 8→2, 10→1. The ladder is a dominant workflow, not an occasional one — which is what makes a
+per-round gate worth its bytes.
+
+🔴 **The "440 rounds" this paragraph used to quote was the SUM OF DEPTHS, and it did not say so.**
+Two different counts are available and they are far apart: rounds *observed* (the distinct numbered
+rounds a walk matches) and rounds *implied* (each session's deepest round summed, on the reasoning
+that a ladder reaching round 8 ran eight). Re-run 2026-08-27 with the checked-in instrument: 127
+sessions, **306 observed / 541 implied**, mean deepest 4.26, 39% ≥5 — and that reading includes the
+ten-round ladder this file documents, which is what "the window moves" means in practice. Quote one
+count, name which, and date it.
+
+**Re-derive rather than quote**: `scripts/ladder-depth-sweep.py` is the instrument, it prints both
+counts with the run date, and it REFUSES to report a zero it cannot distinguish from a broken
+filter — which is what its first version returned, having filtered on the wrong tool name.
 
 ---
 
--- a/scripts/ladder-depth-sweep.py
+++ b/scripts/ladder-depth-sweep.py
@@ -1,0 +1,187 @@
+#!/usr/bin/env python3
+"""How deep do adversarial-audit ladders actually run? — the measurement behind
+the attribution gate in `claude/skills/audit-pr/SKILL.md`.
+
+WHY THIS EXISTS
+---------------
+That skill's stop rule is keyed to FINDINGS, and its attribution gate is keyed
+to whether a round's FIXES changed payload. `scripts/tests/test_audit_ladder_
+stop_rule.py` pins the instructions but says plainly that it cannot observe
+BEHAVIOUR: nothing in the suite proves any session actually stopped. This script
+is the other half — the only thing that can tell you whether the rule changed
+what sessions do.
+
+It answers one question: over the sessions on this host, how deep did numbered
+delta re-audit ladders run? Run it before and after a change to the rule (or on
+two date windows) and compare the distributions.
+
+    scripts/ladder-depth-sweep.py                     # all sessions
+    scripts/ladder-depth-sweep.py --since 2026-08-27  # one window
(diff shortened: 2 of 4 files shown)


ZacxDev:
## Audit → no 🔴, seven of nine fixed in `9263c6ce`

The audit's headline was aimed at the right thing for a test instrument, and it confirmed the harness can go red: it deleted a pin (row correctly reported `SURVIVED`), reworded a target (`MUTATION DID NOT APPLY`), relocated a constant into another test (`WRONG-KILLER` — so that branch is not dead code), and broke a `RULES.md` sentence (baseline red). It also checksummed the source tree before and after a run and after SIGINT: unchanged.

### The one that mattered

**The coverage sentence was ~3× wider than the harness.** It said "the mutation rows quoted in this module are re-derivable with `mutants-audit-ladder.sh`" — the module quotes **45 rows**; the harness ran **13 + 2 controls**, naming only 4 of the module's 11 tests as a killer. That is the guards-narrower shape, in the commit whose whole purpose is making claims re-derivable.

Fixed substantively first: **M4 and M10 are now in the harness.** Those are the two rows the docstring marks 🔴 as the *reachability* controls — relocations that leave every whole-string pin byte-identical, so only an assertion that genuinely executes can see them. They were precisely the rows missing, which meant the two RELATIONSHIP tests had no re-derivable evidence they run at all. Both now killed by their own named test. The prose then narrows honestly: 15 of 45, saying which and why the other 30 aren't there.

**17 rows, all as expected** — 15 mutants killed by their named test, 0 wrong-killers, 2 controls survived.

### A missing pytest read as "the tree is broken"

Run outside `nix develop` — this host's default, since `.envrc` is `use opencode` — the harness printed *"🔴 the UNMUTATED tree is already failing … Fix that first"*, having discarded the `No module named pytest` on stderr that diagnoses it in one line. Now: stderr captured and checked *before* the count, message reads **"THE HARNESS could not run … Nothing was measured. This says nothing about the tree"**, and the usage header names the shell. Verified both ways.

### The rest

- **The `.git` assert is unreachable** — `$ROOT` is a fresh mkdir populated by naming five files. Relabelled as the invariant guard it is: a tripwire on the road not taken, earning its lines only against a future `cp -a "$SRC"` refactor. The real protection is the selective copy, which the header now says.
- **The re-published figures were stale within hours** — an independent same-day re-run gave 319/556/4.38 against the published 306/541/4.26, because the corpus includes the session dispatching the ladder the file documents. Said so, in the paragraph whose subject is a stale published number.
- **`--since` narrowed the positive control**, so an honestly empty window exited 2 blaming a "wrong tool-name filter". Dispatches are counted before the filter now; `--since 2099-01-01` reports a measured zero with 4,254 dispatches walked.
- **Window comparison is left-truncation biased** — 4.38 all-time vs 6.47 since 08-27 on one unchanged corpus. Documented, since comparing windows is the script's stated use.
- **IMPLIED is an upper bound** where numbering skips (3 sessions, 8 round-numbers, ~2% of the gap).
- **`MIN_TESTS` 8 → 10**, which is `run-tests.sh`'s own floor formula at m=11.

**Not fixed, deliberately:** a SIGKILL leaks the mktemp dir (~50 KB; SIGINT is clean and no signal can reach the source tree — the auditor verified that by checksum).

Gate `--tier both` PASS. The auditor also ran the **merged tree** against a newer `main`: pytest 17,687 passed, node 1,292 passed, and both sandbox-tier derivations built.

ZacxDev:
## Round 4 → **CLEAN. The ladder ends here.**

No findings. Every round-3 fix holds; both new `run_move` branches were watched on purpose-built trees rather than reasoned about; both negative controls reproduce verbatim — including the round-2 bug (reverting to `grep -qx` makes state B print `ok` and the sweep say "all as expected") and the round-3 bug (the base harness misreports the named-absent state as `NOT EXCLUSIVE`). Per the stop rule: **a clean round ends the ladder, and do not run another to confirm it.**

The three things the auditor recorded are explicitly *not* findings — two labels elsewhere still emit a trailing comma that nothing parses, a `(45%)` that is 44.59% rounded on a number the sentence calls a floor, and a "must be" that could read as "at least". None changes an operator's action.

Notable: the re-derivation drifted by **+1 session** — this very ladder — which is exactly what the file's own "THE WINDOW MOVES" warning predicts, and the identity the claim rests on (`floor = leading absences − sessions with min round ≥ 2`) reproduces exactly.

## Merge gate

`main` moved to `da3557aa` since this branch was cut. Built an integration branch off it, merged `9f638fd4` cleanly, and on **that merged tree**:

- `scripts/tests/mutants-audit-ladder.sh` → **✅ 17 row(s), all as expected**
- `scripts/gate.sh --tier both` → **`GATE: RESULT=PASS exit=0`**
- sandbox tier, which Tekton gates on: `nix build .#checks.x86_64-linux.pytests` and `.nodetests` both built

## The ladder, four rounds

| round | what it found |
|---|---|
| 1 | the coverage sentence claimed all 45 quoted rows were re-derivable when the harness ran 13 — and the rows missing were **M4/M10, the two reachability controls**, so the RELATIONSHIP pins had no evidence they execute; a missing pytest read as "the tree is broken"; the `.git` assert was unreachable |
| 2 | the corrected count was wrong on both numbers (three independent counts gave 44, 48, 13) — so the sentence now carries **none**; the header excluded by name the row it had just added; a reachability row could pass while a **second test also killed it** |
| 3 | the exclusivity fix **collapsed WRONG-KILLER into NOT EXCLUSIVE** — "your pin is dead" and "your probe lost focus" are opposite instructions; and the quantifier over-shot the other way (97% where the measurement supports 45% as a floor) |
| 4 | clean |

Every round found something in the payload, which is why the attribution gate never fired here — the findings ended the ladder, not the gate. That is the rule working as written in both directions: on #900 the gate stopped a ladder the findings would have run forever, and here the findings stopped it while the gate stayed silent.

PR #935, opened 2026-08-27
fix(worktree-prune): a submodule-bearing worktree is never `removable` (#931)

### LINKED ISSUE #931: worktree-prune: removable overcounts ~2x — submodule-bearing worktrees can never be removed (54 of 107 on the first real --execute)
## What happened

The first real fleet-wide `--execute` after #895 merged:

```
execute: removed=53  skipped=0  failed=54
```

**All 54 failures are one cause**, and zero failures had any other cause:

```
FAILED  /home/zach/workspace/civit/civitai-270-asset-dims-verify: rc=128
        fatal: working trees containing submodules cannot be moved or removed
```

`git worktree remove` refuses outright on a worktree with populated submodules. `worktree-prune` correctly **never passes `--force`** (pinned + mutation-tested), so these rows are **permanently unremovable by this tool** — not transiently, not fixable by re-running.

## Why it is a defect and not just a git limitation

`removable` is a claim the tool cannot cash for these rows, and `--confirm N` is derived from that claim. On this box it reported **107 removable** where **53** was achievable — a **~2x overcount** in the operator's primary safety number.

That is the same family of defect #895 spent nine rounds eliminating: a statement in the report that is not true of the run printing it. It fails safe (loudly, per-row, nothing damaged, no `--force`), but:

- an operator sizing a prune reads 107 and gets 53, with no way to tell in advance which rows;
- `--confirm N` is supposed to mean "the exact number this run will remove" — its own `--help` says so — and here it cannot;
- re-running never converges: the same 54 fail every time, so the count stays permanently wrong.

## Mechanism (confirmed)

`git grep -ci submodule scripts/worktree-prune` → **0**. The tool has no submodule awareness at all. On a failing row:

```
.../civitai-270-asset-dims-verify/.gitmodules                      exists
.../civitai-270-asset-dims-verify/event-engine-common/.git         populated submodule
```

`classify()` marks the row `dead` + clean + unlocked ⇒ `removable=True`, and nothing consults `.gitmodules` or checks for a populated submodule.

## Reproduce

```bash
# hermetic: a repo with a submodule, plus a worktree on a merged branch.
# the row classifies removable; --execute then fails rc=128 on it.
scripts/worktree-prune --repo <repo> --execute --confirm 1
```

Any `civit/civitai-*` worktree on this host reproduces it directly.

## Suggested direction (not prescriptive)

Either is fine; the second is probably more honest:

1. Detect a populated submodule during classification and mark those rows **not removable**, with a `verdict_reason` naming the submodule — so they stay visible and counted, not silently dropped. Precedent exists: the tool already reports excluded rows rather than hiding them.
2. Keep them `dead` but add a third bucket alongside `removable` / `excluded` (e.g. `blocked`), counted separately and excluded from `--confirm`.

🔴 Whichever is chosen, **do not reach for `--force`** — `test_the_tool_never_passes_force_to_worktree_remove` and its mutant exist for a reason, and forcing removal of a worktree containing submodules is exactly the destructive escalation this tool is built to refuse.

## Closing condition

This pair, run back to back on the workbench host, agrees — `failed=0` and `removed == N`:

```bash
scripts/worktree-prune --scan-root ~/workspace --scan-depth 2          # reports N removable
scripts/worktree-prune --scan-root ~/workspace --scan-depth 2 \
    --execute --confirm N                                             # => removed=N failed=0
```

Checked by: the above command pair exiting 0 with `failed=0`, run by whoever performs the next fleet prune, against a merged PR on `innovation-upstream/devrc`.

## Context

Found by running the prune #895 was built to unblock. The net result of that run was still good — 53 dead worktrees removed, 140 live sessions' trees spared by the new default, 1 in-use tree spared by an independent `/proc/*/cwd` check, **0 branches lost**, 0 damage. This issue is the one claim in it that did not hold.

ZacxDev:
Closes #931.

`worktree-prune` reported `removable` for worktrees `git worktree remove` will
never remove. On the first real fleet `--execute` after #895 that was
`removed=53 skipped=0 failed=54`, every failure the same string
(`rc=128 fatal: working trees containing submodules cannot be moved or
removed`) — so `--confirm` said **107** where **53** was achievable, a ~2×
overcount in the operator's primary safety number, and re-running never
converged because the cause is permanent rather than transient.

## The issue's stated mechanism was wrong, and that changed the fix

#931 proposed detecting this by consulting `.gitmodules`. Measured against
**git 2.55.0** on six hermetic fixtures, with the refusal outcome recorded next
to every candidate signal:

| fixture | `.gitmodules` | `modules/` store | populated gitlink | `git worktree remove` |
|---|---|---|---|---|
| submodule **uninitialised** in the worktree | yes | no | no | **rc=0, REMOVED** |
| submodule **initialised** | yes | yes | yes | rc=128 refused |
| submodule **deinitialised** | yes | yes | **no** | rc=128 refused |
| old-style **embedded `.git`** dir | yes | **no** | yes | rc=128 refused |
| **empty** `modules/` store | no | yes (empty) | no | rc=128 refused |
| no submodule anywhere (control) | no | no | no | rc=0, REMOVED |

So `.gitmodules` is not the predicate in either direction: row 1 carries it and
removes fine, so keying on it **over**-blocks. Nor is a populated gitlink: row 3
has none and is still refused, so keying on that alone **under**-blocks — and
under-blocking is the direction that puts `failed=` back in the output.

The fix mirrors git's own predicate, in git's own order:

> *this worktree's* git dir has a `modules/` store — `rev-parse --git-path
> modules`, which resolves per-worktree — **OR** a gitlink in the index is
> populated.

Row 5 is where a "cleverer than git" implementation would break: git refuses on
the store's mere **existence** (its own source calls that a knowingly-accepted
false positive), so requiring the store to be non-empty would hand back
`failed=1`. The tool agrees with git there, and the test measures git's
behaviour in the same test rather than assuming it.

## What changed

- `worktree_submodules()` — the predicate above. `None` means **unknowable**,
  never `False`.
- `classify()` — a **blocker**, not a new verdict. Whether the branch *landed*
  is a different question from whether the directory can be *removed*, exactly
  as with `excluded_by`. The row stays `dead`, stays printed, and drops out of
  `removable` and therefore out of `--confirm`. A missing or `None` answer
  blocks: an unanswered question about a refusal is not permission.
- `execute_removals()` — a **second, independent floor** immediately before the
  destructive step, alongside the existing dirty/locked re-checks. A submodule
  initialised by a concurrent session between scan and removal now produces a
  `SKIP`, never a `failed=`.
- `summarize()`/`render_text()` — `submodule_blocked_dead`, printed directly
  under the removable count so the gap between `dead` and `removable` has a
  name instead of being a silent truncation. Counted independently of
  `excluded_dead` (a row can be both). `--verbose` names the submodule.
- `scripts/README.md` — the measured predicate and the new `--out` JSON key.

🔴 **No `--force` anywhere.** `test_the_tool_never_passes_force_to_worktree_remove`
still holds — it caught this branch once already, on the *prose* of the new
summary line, and the prose was reworded rather than the guard weakened.

## Verification

**Every new test was watched to fail on pre-change code.** Run in an isolated
copy carrying `origin/main`'s `scripts/worktree-prune` and this branch's tests:

| tree | result |
|---|---|
| `origin/main`'s tool + these tests | **11 failed, 186 passed** |
| this branch (HEAD) | **197 passed** |

The 11 failures are exactly the 11 new tests, and the 186 is the pre-change
suite's own full count — so nothing pre-existing moved, including under the
`base_row` fixture change. The red run also reproduces the original symptom in
a hermetic fixture: `FAILED …/confirm-wt: rc=128 fatal: working trees
containing submodules cannot be moved or removed`.

The three refusal fixtures are **pairwise distinct on purpose**: the
deinitialised case is the only one where the `modules/` arm fires alone, and the
embedded case the only one where the index arm does — so the two arms are
mutated separately and each names the one fixture that can kill it. The
uninitialised fixture is the **over-block control**: without it, a tool that
blocked every worktree unconditionally would score green on every other
assertion here.

### Mutation battery — 10 new mutants, COMPLETE: `ALL MUTANTS ACCOUNTED FOR`

**76 run lines, 76 ok, 0 problems, 0 unscored** on CPython 3.12.14; each of the
10 new mutants died to the test that NAMES it (no `WRONG-KILLER`), and both
harness controls behaved — a comment-only edit SURVIVED and `main()` returning
early DIED. Full output in the comment below.

`scripts/tests/mutants-worktree-prune.sh` gains 10 `run` lines (76 total). Each
names the one test that must kill it, so a mutant that dies to a *different*
test's error is scored `WRONG-KILLER`, not `ok`. The set deliberately includes
the two mutants that push in **opposite** directions:

- `submodules-reported-for-every-worktree` — the **over-block** mutant, which
  only the uninitialised control fixture can kill;
- `modules-store-required-to-be-non-empty` — the plausible "be cleverer than
  git" refinement, which only the empty-store fixture can kill;
- `executor-recheck-treats-unknown-as-permission` — `is not False` vs
  `is True`, which behaves identically on both booleans and is visible only to
  the monkeypatched `None` re-check.

🔴 **What this PR does NOT claim.** The issue's closing condition is a real
fleet `--execute` reporting `removed == N, failed=0`. It has **not** been run —
it needs a quiet host and a fresh dry scan first (the removable count moved
108 → 110 → 107 across three scans minutes apart), and it is tracked as its own
queue item rather than folded in here. Everything above is hermetic-fixture and
mutation evidence; none of it is a reading of the real fleet.


ZacxDev:
## Mutation battery: complete — `ALL MUTANTS ACCOUNTED FOR`

The run the PR body said not to trust until this comment existed. On CPython
3.12.14, `nix develop -c bash scripts/tests/mutants-worktree-prune.sh`:

```
run lines in script : 76
ok (killed/control) : 76
problems (🔴)       : 0
unscored (--)       : 0
ALL MUTANTS ACCOUNTED FOR
tool restored by content: sha256 be03a53d…89f3dd80
source tree untouched:    sha256 be03a53d…89f3dd80
```

`unscored = 0`, and the verdict is the unqualified `ALL MUTANTS ACCOUNTED FOR`
rather than the `ALL SCORED MUTANTS …` variant — so the interpreter-specific
mutant was scored here and this is the complete sweep, not a 3.13 run's
narrower one.

### The 10 new mutants, each killed by the test that NAMES it

```
ok submodule-not-a-blocker-in-classify            killed by test_an_INITIALISED_submodule_makes_the_row_dead_but_not_removable
ok unknown-submodule-answer-treated-as-permission killed by test_an_unknowable_submodule_answer_blocks_rather_than_permits
ok modules-store-arm-blinded                      killed by test_a_DEINITIALISED_submodule_still_blocks_because_the_modules_store_remains
ok modules-store-required-to-be-non-empty         killed by test_an_EMPTY_modules_store_blocks_because_git_blocks_on_it_too
ok index-arm-blinded                              killed by test_an_EMBEDDED_submodule_git_dir_blocks_even_with_no_modules_store
ok gitlink-mode-misspelled-so-no-gitlink-is-ever-seen killed by test_an_EMBEDDED_submodule_git_dir_blocks_even_with_no_modules_store
ok submodules-reported-for-every-worktree         killed by test_a_worktree_with_an_UNINITIALISED_submodule_is_still_removable
ok executor-skips-its-own-submodule-recheck       killed by test_a_submodule_appearing_after_the_scan_skips_rather_than_fails
ok executor-recheck-treats-unknown-as-permission  killed by test_the_executor_skips_when_the_submodule_recheck_cannot_answer
ok blocked-rows-vanish-from-the-report            killed by test_the_report_names_the_blocked_rows_instead_of_silently_dropping_them
```

No `WRONG-KILLER` on any of them — each died to its own assertion, not to some
other test's error, which is the distinction this harness exists to draw.

Note the two that push in **opposite** directions, because a set that only
pushed one way would score green on a tool that blocked everything:

- `submodules-reported-for-every-worktree` (over-block) died **only** to the
  uninitialised control fixture;
- `modules-store-required-to-be-non-empty` — the plausible "be cleverer than
  git" refinement — died **only** to the empty-store fixture;
- `executor-recheck-treats-unknown-as-permission` (`is not False` → `is True`)
  is invisible to every boolean fixture and died to the monkeypatched `None`
  re-check alone.

### The harness's own controls, which is what makes the 76 readable

```
ok comment-only-edit-must-survive                 SURVIVED as required (control)
ok main-returns-immediately-must-die              killed by test_dry_run_is_the_default_and_removes_nothing
```

A comment-only edit surviving says the harness measures behaviour rather than
file text; `main()` returning early dying says the suite actually executed.
Without both, every `ok` above would be a claim about a harness wired to
nothing.

🔴 **Still NOT verified, and not claimed by this PR:** the issue's closing
condition is a real fleet `--execute` reporting `removed == N, failed=0`. That
has not been run — it needs a quiet host and a fresh dry scan first (the
removable count moved 108 → 110 → 107 across three scans minutes apart), and it
is tracked as its own queue item rather than folded in here.

PR #950, opened 2026-08-27
fix(browser-bridge): reap the tab `open`'s bounded reuse probe orphans — and NOT by bounding tabs.create

ZacxDev:
Queue item 4 of `claudedocs/handoff-bridge-unbounded-waits.md` — **built as the opposite of what the item asked for, and that is the finding.**

## What the item asked for, and why it was refused

> "`chrome.tabs.create` is another unbounded browser-process IPC, so a browser-wide stall leaks one tab per `open` while returning success. Worth more than its rank: the #797 audit PREDICTED this shape would regenerate and #814 proved it right, so this is the third instance of a class that is 2-for-2."

The class is real. The remedy is not:

* **`execute()`'s own rule forbids a bound here.** "A local bound is granted only to *an await inside a `try` whose `catch` implements a RECOVERY (a second, working path)* … If a hung step has no alternative to fall through to, the choke point below is already the right and only answer — **do NOT add a bound for it.**" `chrome.tabs.create` has no second path.
* **A bound would not fix the harm the item names.** It relabels `op_timeout:open` as a create-specific timeout and leaks *exactly the same tab*, 2 s earlier. And the named harm — "leaks one tab per `open` **while returning success**" — is the reuse-probe fall-through, which needs no hang in `create` at all.

`server.py` had already written the real defect down:

> a probe that exceeds the budget falls through to a FRESH tab and the ownership record below is overwritten, deterministically **ORPHANING the live first tab. Nothing reclaims it; only a human closes it.**

That sentence is what this PR deletes. **The leak, not the label, was the defect.**

## The change

~10 lines of production code. `open` now closes the orphan itself — on the **timeout arm only**, **after** the replacement tab exists, **fire-and-forget**.

Three decisions, each pinned by a mutant:

| decision | why | mutant |
|---|---|---|
| timeout arm **only** | a rejection is `tabs.get` naming the id absent — a remove reclaims nothing and only risks a recycled tabId | M1 unconditional reap |
| **after** the create | reaping first closes the session's only tab, then on a failed create leaves it owning a dead id | M5 reap-before-create |
| **never awaited** | `tabs.remove` is one more unbounded IPC — awaiting it puts a new unbounded await on `open`'s SUCCESS path, regenerating this class at the last line of the op | M3 awaited-reap |

## What it does NOT fix — stated, not discovered later

1. **Best-effort only.** The browser-wide stall that hung `tabs.get` can hang `tabs.remove`, and nothing waits to find out.
2. **The other orphan is untouched.** A hung `chrome.tabs.create` is killed at `EXEC_OP_BUDGET_MS` while the abandoned create still settles, producing a tab `open` never sees. Closing it needs the choke point to signal abandonment back into the op — no mechanism today. Deliberately out of scope.
3. **No new exposure class.** This is the same `chrome.tabs.remove(<the session's owned tabId>)` that `close` already performs unconditionally, on an id the server vouched for moments earlier.

## Evidence

**Watched RED at the base.** Reverting only the production change (tests unchanged) → **30 pass / 2 fail**, both as ordinary assertion failures naming their own cause — not `cancelled`, which is how a hang scores and how a fail-count gate reads a live regression as clean:

```
not ok 19 - a TIMED-OUT probe reaps the tab it just orphaned
            actual [] / expected [5]
not ok 22 - a HANGING / REJECTING / THROWING reap cannot delay or fail open
            "…and the reap must still have been ATTEMPTED"  0 !== 1
```

The other two new tests **pass at the base by construction** — they are invariant guards / discriminating negatives (the rejection arm must not reap; healthy reuse and a first open must not reap), **not** regression coverage. Labelled as such rather than counted.

**Mutation battery — 7 mutants, all KILLED, each by the assertion naming its own decision.** Control: unmutated tree, 0 failures.

```
M1 unconditional reap       KILLED  "must not be removed"          (the negative arm)
M2 inverted arm             KILLED  "Nothing reclaims it"
M3 awaited reap             KILLED  "the reap is being AWAITED"
M4 try/catch deleted        KILLED  sync-throw arm, at service_worker.js:1301 —
                                    raw "chrome.tabs.remove is not a function",
                                    not an assertion message (stack verified)
M5 reap before create       KILLED  "reap belongs AFTER chrome.tabs.create"
M6 .catch() dropped         KILLED  unhandledRejection on the reject arm
M7 reap the wrong id        KILLED  fixture-distinctness control (TAB_ID != FRESH_TAB_ID)
```

**Suites.** node `511 pass / 0 fail` across all 15 `.test.mjs`; browser-bridge pytest `792 pass / 0 fail`.

The two **build-marker gates went red first and were right** — the extension source changed, so `build_id.js` is regenerated here (`e1ee86a50a811d40` → `18304ead05a6e386`).

## Prose describing this hazard as open is corrected

A comment is a claim. Every site that said the orphan is unreclaimable now says what is actually true, with the residual stated: `server.py`'s "Nothing reclaims it" block; `REUSE_TAB_BUDGET_MS`'s cost paragraph in `protocol.js` (**kept verbatim** — it is the derivation of 2000 — with the change appended); the reuse-probe floor rationale in `cdp_protocol.test.mjs`; the attribution-control comment in `service_worker.test.mjs`; and the operator-facing claims in `README.md` and `reference/tabs-instances.md`.

## Separate finding, NOT fixed here

`validateCommand` does not whitelist fields and `_handle_cmd` pops only `target`/`tab`/`focus`, so a client-supplied `reuseTabId` survives into the command whenever the session owns no tab — after which `open` returns that tab and `_record_ownership_locked` records ownership of it, bypassing `_resolve_tab`'s `not_owned_tab` guard. **Read from the code, not executed** (a live probe would take over the operator's tab). Pre-existing and independent of this PR; this change does not widen it (the reap fires only on a hang the caller cannot induce, against a tab the hole already lets them own and `close`).

### DIFF
--- a/claudedocs/handoff-bridge-unbounded-waits.md
+++ b/claudedocs/handoff-bridge-unbounded-waits.md
@@ -222,7 +222,7 @@ the browser-bridge and fixed at the source each time.
 
 ## Next steps (ranked)
 🔴 **Numbering is STABLE and is half a claim's identity — do not re-rank.** Items 1,
-2, 3 and 7 are DONE and are kept in place rather than renumbered.
+2, 3, 4 and 7 are DONE and are kept in place rather than renumbered.
 
 1. ✅ **DONE (#1316, squash `4379c27cf`, 2026-08-27)** — crops declared for `app-requests`
    and `playable-collections`; talos-infra #1297 closed.
@@ -241,12 +241,34 @@ the browser-bridge and fixed at the source each time.
    from "the CLI hung AFTER the reply". Watched firing in BOTH directions with two fake
    CLIs, not merely reasoned about. **See the new investigation block below — the cause
    may never have been this test at all.**
-4. **`open`'s fallthrough shares the suspected failure mode** — `devrc`,
-   `scripts/browser-bridge/extension/service_worker.js`. UNCHANGED. `chrome.tabs.create`
-   is another unbounded browser-process IPC, so a browser-wide stall leaks one tab per
-   `open` while returning success. 🔴 **THIS IS THE NEXT ITEM, and it is worth more than
-   its rank**: the #797 audit PREDICTED this shape would regenerate and #814 proved it
-   right, so this is the third instance of a class that is 2-for-2.
+4. ✅ **DONE (devrc PR #950)** — and, like item 3, **NOT as this item was framed**. The
+   item said to bound `chrome.tabs.create` as the third instance of the #797/#814
+   unbounded-await class. **Two measurements refused that remedy**, and the refusal is
+   the deliverable:
+   - `execute()`'s own rule forbids it: a local bound is granted only to *"an await
+     inside a `try` whose `catch` implements a RECOVERY (a second, working path) … If a
+     hung step has no alternative to fall through to, the choke point below is already
+     the right and only answer — do NOT add a bound for it."* `chrome.tabs.create` has
+     no second path, so a bound could only relabel `op_timeout:open` and leak the same
+     tab 2 s earlier.
+   - the harm the item NAMES — "leaks one tab per `open` while returning success" — is
+     the **reuse-probe fall-through**, which needs no hang in `create` at all, and
+     `server.py` had already written it down verbatim: *"deterministically ORPHANING
+     the live first tab. Nothing reclaims it; only a human closes it."*
+
+   So: `open` now **reaps** that orphan — timeout arm only, after the replacement
+   exists, fire-and-forget, and **only while the op is still alive**. That last clause
+   was NOT in the first draft; a blind audit found the tail, and it was reproduced
+   before being fixed: `execute()` races the op but cannot cancel it, so a merely SLOW
+   `chrome.tabs.create` resumes `open` after `op_timeout:open` has already been
+   answered — and on that path the server KEEPS ownership of a tab that is still live,
+   so a late reap destroys it and strands the session on a dead id.
+   🔴 **What is still open and deliberately out of scope**: the OTHER orphan. A
+   `chrome.tabs.create` that hangs and never settles is killed at
+   `EXEC_OP_BUDGET_MS` while the abandoned create later produces a tab `open` never
+   sees. Closing it needs the choke point to signal abandonment back into the op, which
+   it has no mechanism for. **If you pick that up, it is a choke-point change, not
+   another budget constant** — re-read `execute()`'s narrow-exception rule first.
 5. **Removing app-capture's raise** — `civitai/talos-infra`. 🔴 **PREMISE CHANGED**: the
    raise is not inert, only its i3 half is withheld. Re-scope before working it.
 6. **clawgate #358** — filed earlier, picked up by a different session. Not ours; live
--- a/claudedocs/handoff-bridge-unbounded-waits.md
+++ b/claudedocs/handoff-bridge-unbounded-waits.md
@@ -256,19 +256,40 @@ the browser-bridge and fixed at the source each time.
      `server.py` had already written it down verbatim: *"deterministically ORPHANING
      the live first tab. Nothing reclaims it; only a human closes it."*
 
-   So: `open` now **reaps** that orphan — timeout arm only, after the replacement
-   exists, fire-and-forget, and **only while the op is still alive**. That last clause
-   was NOT in the first draft; a blind audit found the tail, and it was reproduced
-   before being fixed: `execute()` races the op but cannot cancel it, so a merely SLOW
-   `chrome.tabs.create` resumes `open` after `op_timeout:open` has already been
-   answered — and on that path the server KEEPS ownership of a tab that is still live,
-   so a late reap destroys it and strands the session on a dead id.
+   So the orphan is now **reclaimed** — and 🔴 **WHERE that happens took three
+   rounds and two wrong answers, which is the durable lesson here.** Drafts 1 and 2
+   had the EXTENSION close the tab; both were wrong the same way. **Reclaiming is
+   correct only if the `open` result is DELIVERED** — if it is not, the old tab is not
+   an orphan at all: the ownership record still names it, it is still live, and closing
+   it strands the session on a dead id and then, one op later, on the operator's ACTIVE
+   tab. The extension cannot know, because **TWO parties abandon an op on TWO clocks
+   and neither is visible from inside it**:
+   - `execute()` RACES the op against `EXEC_OP_BUDGET_MS` and cannot cancel it, so a
+     merely SLOW `chrome.tabs.create` resumes `open` after `op_timeout:open` was
+     already answered. Found by a blind audit, **reproduced before fixing** (execMs 80,
+     hung `tabs.get`, create at 300 ms → `tabsRemove []` at return, `[5]` 600 ms later);
+   - the SUBMITTER gives up at its own `cmd_timeout`, which starts at **SUBMIT**, so
+     queue time is structurally invisible extension-side (`server.py`: *"The SUBMITTER
+     giving up … does NOT free the extension"*). Found by the DELTA audit of the fix
+     for the first — an elapsed-time guard in the extension closed one and could not
+     close the other.
+
+   Final shape: the extension **reports** `orphanTabId` and closes nothing; the server
+   closes it in `_record_ownership_locked`, which by construction runs only on a
+   delivered result. That deleted the elapsed guard, its `startedAt` stamp, and a
+   fixture problem along with it — **the party that knows should decide.**
+   🔴 **NEW HAZARD THIS CREATED, and it is disclosed rather than fixed**: parallel
+   agents can derive the SAME session id and share one owned tab (observed
+   2026-07-31, recorded in `scripts/browser-bridge/reference/tabs-instances.md`). A
+   re-`open` by one sibling whose probe times out now closes the tab the OTHER is
+   mid-workflow on. Previously that collision was benign.
    🔴 **What is still open and deliberately out of scope**: the OTHER orphan. A
-   `chrome.tabs.create` that hangs and never settles is killed at
+   `chrome.tabs.create` that hangs and NEVER settles is killed at
    `EXEC_OP_BUDGET_MS` while the abandoned create later produces a tab `open` never
-   sees. Closing it needs the choke point to signal abandonment back into the op, which
-   it has no mechanism for. **If you pick that up, it is a choke-point change, not
-   another budget constant** — re-read `execute()`'s narrow-exception rule first.
+   sees — so there is nothing to report. Closing it needs the choke point to signal
+   abandonment back into the op, which it has no mechanism for. **If you pick that up,
+   it is a choke-point change, not another budget constant** — re-read `execute()`'s
+   narrow-exception rule first.
 5. **Removing app-capture's raise** — `civitai/talos-infra`. 🔴 **PREMISE CHANGED**: the
    raise is not inert, only its i3 half is withheld. Re-scope before working it.
 6. **clawgate #358** — filed earlier, picked up by a different session. Not ours; live
--- a/scripts/browser-bridge/README.md
+++ b/scripts/browser-bridge/README.md
@@ -232,7 +232,7 @@ has not been exercised against a third-party authenticated API beyond that.)
 | `tabs`       | `chrome.tabs.query({})`                                   | `{tabs:[...],ownedTabId}` |
 | `nav`        | `chrome.tabs.update(tab,{url})`                           | `{tabId,url}` |
 | `screenshot` | **CDP `Page.captureScreenshot`** (png) — works on a BACKGROUND/occluded tab + each profile's own tab; a foreground tab uses the cheap `captureVisibleTab` fast path. `fullpage` grabs the whole document. The CLI decodes `dataUrl` to a **`.png` on disk** and prints a path, never the base64 (see below) | `{url,dataUrl,via}` |
-| `open`       | `chrome.tabs.create({url,active:false})` — **background/HIDDEN** (`visibilityState:"hidden"` → Chromium throttles it → a heavy SPA won't render → reads return a shell). The escape hatch is **`browser wake`** (or a `--wake` read): it un-throttles the tab and moves NO focus. Reads self-announce this via `hidden`/`note` | `{tabId,url}` |
+| `open`       | `chrome.tabs.create({url,active:false})` — **background/HIDDEN** (`visibilityState:"hidden"` → Chromium throttles it → a heavy SPA won't render → reads return a shell). The escape hatch is **`browser wake`** (or a `--wake` read): it un-throttles the tab and moves NO focus. Reads self-announce this via `hidden`/`note`. On a re-`open` whose reuse probe TIMES OUT, it also **closes the orphaned previous tab** and reports `reapAttempted` | `{tabId,url}` (+`reapAttempted` when it reaped) |
 | `close`      | `chrome.tabs.remove(tabId)`                               | `{closed:tabId}` |
 | `emulate`    | **device emulation** — CDP `Emulation.setDeviceMetricsOverride` / `setTouchEmulationEnabled` / `setUserAgentOverride` (**+`userAgentMetadata`**) / `setEmulatedMedia` / `setTimezoneOverride` / `setGeolocationOverride`, from a named preset or raw params. **Sticky per tab** (re-applied inside every later CDP session) and **owned-tab-only**. `--reset` stops re-applying **and restores the viewport** — it sends an arm-then-clear pair, because a bare `clearDeviceMetricsOverride` is a measured no-op (#319). `--reset --recreate` replaces the tab (new tab id) instead, and is the only remedy for a tab the extension cannot reach | `{tabId,url,emulation,applied,note}` or `{tabId,url,reset,wasEmulating,cleared,restored,note}` |
 | `frames`     | **`chrome.webNavigation.getAllFrames`** — the tab's frames INCLUDING cross-origin OUT-OF-PROCESS iframes (OOPIFs) | `{url,title,frames:[{frameId,url,parentFrameId}]}` |
@@ -1945,6 +1945,16 @@ per-session (multi-step **workflow**) isolation did not.
   deliberately restricted to the *timeout* arm; when the probe REJECTS, the tab is
   known absent and a `remove` would only risk a recycled tabId. If the owned
   tab is **gone**, `open` transparently creates a fresh one.
+  🔴 **What this looks like to you, stated because it is the first tab close the
+  extension performs that nobody asked for.** Bridge tabs open `active:false`, but
+  `browser activate` foregrounds them and that is the ordinary workflow — so the
+  page you are *looking at* can vanish mid-session on a re-`open`, taking scroll
+  position, form state and any playing media with it. You get a fresh tab at the
+  requested url in its place. The envelope says so: an `open` that reaped carries
+  **`reapAttempted: <tabId>`** (ATTEMPTED, not closed — nothing awaits the
+  result). An ordinary `open` envelope is unchanged. The reap never fires once the
+  op has already timed out, because on *that* path the old tab is still owned and
+  still usable.
 - **Self-heal on a vanished owned tab.** If the user manually closes an owned
   background tab, the next tab-scoped op dispatches the stale tabId and the
   extension returns `owned_tab_gone` (ok:false). The server then **drops** that
--- a/scripts/browser-bridge/README.md
+++ b/scripts/browser-bridge/README.md
@@ -1328,14 +1328,21 @@ Two changes close it:
       ⚠ **Cost:** falling through on a merely *slow* `tabs.get` opens a second tab
       while the first is still live and unowned, i.e. the exact tab leak the reuse
       path exists to prevent. A rare orphaned tab is traded for a
-      guaranteed-terminating op — **and that orphan is now reaped**: `open` closes
-      it itself, on the timeout arm only, *after* the replacement exists, and
-      fire-and-forget (awaiting `chrome.tabs.remove` would add a new unbounded
-      `chrome.*` await to the success path — the very class this bound exists to
-      remove). Best-effort, not a guarantee: the browser-wide stall that hung
-      `tabs.get` can hang `tabs.remove` too, and a *different* orphan — one from a
-      hung `chrome.tabs.create`, whose op is killed at `EXEC_OP_BUDGET_MS` while
-      the abandoned create still settles — is still unreclaimed by design.
+      guaranteed-terminating op — **and that orphan is now reclaimed**: `open`
+      REPORTS it as `orphanTabId` and closes nothing; the SERVER issues the
+      `close`, from the one place that knows the `open` result was delivered.
+      (An earlier revision of this paragraph said `open` "closes it itself,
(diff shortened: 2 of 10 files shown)


ZacxDev:
## Audit round 1 — one real behaviour defect, fixed in `9ec13738`. Plus a correction to this PR's own description.

A blind adversarial audit ran against `34e48eef`. It independently re-derived every claim in the body above (build marker, 511/0 node, 792/0 pytest, the RED-at-base 30/2 with verbatim messages, and all 7 mutants) and confirmed them. It also found the tail I missed.

### 🔴 The finding that mattered — `behaviour`

`execute()` **races** the op against `execMs`; it does not **cancel** it. So a `chrome.tabs.create` that is merely **slow** — not hung — resumes `open` long after the op answered `op_timeout:open`.

And on that path the old tab is **not an orphan**. The server saw a non-tab-gone error, so `_record_ownership_locked` **keeps** ownership pointing at a tab that is still live and still usable; pre-change the next `open` simply reused it. A late reap destroys exactly that tab → the session owns a dead id → `owned_tab_gone` → the mapping is dropped → the op *after that* falls back to the **operator's active tab**, which `server.py:1873-1876` calls "exactly the blast radius the ownership map exists to prevent".

**So the first commit converted a benign leak into an unrequested close of a live, owned tab.** Narrow tail, real regression.

Reproduced before fixing, not reasoned about — `execMs` 80, hung `tabs.get`, create resolving at 300 ms:

```
execute() returned: {"id":"c1","ok":false,"error":"op_timeout:open"}
tabsRemove AT return:   []
tabsRemove 600ms LATER: [5]
```

Fix: stamp `Date.now()` before the first await; skip the reap once `Date.now() - startedAt >= loopTiming().execMs`. `>=` deliberately — at exactly `execMs` the race has fired and a tie must skip. The guard reads `loopTiming().execMs`, **the same expression `execute()` races on** (`service_worker.js:1729`), so the two cannot drift.

### 🟡 Also fixed

| # | what | why it mattered |
|---|---|---|
| 2 | `cdp_protocol.test.mjs` floor-assertion **failure message** still said "nothing reclaims the old one" | I corrected the comment above it and not the string — read at exactly the moment someone decides whether the 1000 ms floor still matters. Floor unchanged: the reap is best-effort and least likely to land under the very stall that causes a slow probe. |
| 3 | handoff queue item 4 still said **UNCHANGED** and prescribed the rejected remedy | the next session would re-derive the `tabs.create` bound this PR argues against. Now DONE with the refusal recorded, and it names what is still open. |
| 4 | operator-facing prose framed the reap only as a benefit | bridge tabs open `active:false`, but `activate` foregrounds them and that is the ordinary workflow — **the page you are looking at can vanish on a re-`open`**. Every site I added described the reclaim; none stated the cost. |
| 6 | claim (c) "the same thing `close` does" was not literal | `close` calls `forgetTab` **first**, and says why (onRemoved's timing is not ours; a stale entry is inheritable by a recycled tabId). The reap now does too. |
| 7 | the asymmetry comment was one-sided | it implied the timeout arm is not exposed to a recycled tabId. It **is**, and more so — its remove is issued later. Rewritten as expected value: the rejection arm reclaims **zero** tabs so any exposure buys nothing; the timeout arm has a live tab to reclaim. Conclusion unchanged, reasoning now carries it. |
| 8 | zero observability | the reap left **no trace anywhere** — and it is the first tab close the extension performs that nobody asked for. `open`'s envelope now carries `reapAttempted: <tabId>`, additive, present only on the reaping path. **ATTEMPTED** is the honest word: nothing awaits the remove. |
| 9 | arm (a) counted via a local variable | a counter cannot see a reap issued against the **wrong id**, making that arm the one place in the file blind to it. Now reads `state.calls.tabsRemove`. |

### 🔴 Correcting this PR's description, publicly rather than by a quiet edit

The body above says the round-1 battery was *"7 mutants, all KILLED, **each by the assertion that names its own decision**."* **That headline is wrong for 2 of the 7.** M4 dies on the raw `chrome.tabs.remove is not a function` propagating out of `service_worker.js`, and M6 on an `unhandledRejection` — neither is an assertion naming its decision. The *table rows* in the body disclose both correctly; only the summary sentence overreached. Flagging it here because a reviewer may already have read the stronger version. No mutant died for another guard's reason — that part holds.

### Round 2 mutation

5 new mutants; **M13 and M14 SURVIVED on first run** and are the reason two of the guards above exist at all.

```
CONTROL(clean)              0 failing
M10 elapsed guard deleted   KILLED  "SERVER STILL OWNS"   (the pre-fix state)
M11 elapsed guard inverted  KILLED  3 tests
M12 reads pollMs not execMs KILLED  "SERVER STILL OWNS"   (a bound that cannot fire)
M13 forgetTab dropped       KILLED  "the reap must forget the tab's emulation
                                     state, as `close` does"   true !== false
M14 reapAttempted moved     KILLED  the sync-throw trace assertion
M15 reapAttempted field del KILLED  the envelope-trace assertion
```

The two `has(...) === false` assertions **seed both maps first and assert the seed took** — otherwise they hold on a build that never forgets anything.

**One known equivalent mutant is recorded in the test file** rather than left to be re-derived as a coverage gap: `orphanTabId != null` → truthiness survives, because Chromium never issues tab id 0.

### Ledger

```
round 1 · payload 114 · scaffolding 199 · tekton/devrc-{nodetests,pytests} SUCCESS
round 2 · payload 125 (service_worker.js 76+/21-, handoff 29+/7-, README 11+/1-,
                       tabs-instances 6+/1-, build_id 1+/1-)
          scaffolding 117 (service_worker.test.mjs 108+/4-, cdp_protocol 4+/1-)
```

Round 2 changed payload, so the attribution gate does not fire.

Suites at `9ec13738`: node **511 pass / 0 fail** across 15 files; pytest **869 pass / 0 fail** (browser-bridge + doc-path-rot). Build marker regenerated `18304ead05a6e386` → `9a3c89153f041614`.

ZacxDev:
## Audit round 2 (delta) — the twin of round 2's defect, which proved the *design* was wrong. Fixed in `cc7f4e59`.

The delta audit independently re-derived every round-2 claim (all mutants, both suites, the marker) and confirmed them, with two corrections noted below. Then it found the twin.

### 🔴 `execute()` is not the only thing that abandons an op — the SUBMITTER abandons it first, on a different clock

`cmd_timeout` (20 s) starts at **SUBMIT**; the extension's `execMs` (18 s) starts at **DEQUEUE**. Queue time is therefore structurally invisible to any extension-side elapsed check. `server.py:270-275` already said so:

> The SUBMITTER giving up (BridgeTimeout at cmd_timeout, default 20s) does NOT free the extension — its serial loop keeps executing that command for up to EXEC(18) + RESULT(10) = 28s.

Reachable: a prior op holds the loop 15 s → this `open` dequeues → probe times out at t=17 s → `tabs.create` settles at t=23 s. Round 2's guard sees 8 s < 18 s and **reaps**; the submitter gave up at t=20 s, so `_record_ownership_locked` never ran and ownership **still names the tab just closed**. Identical blast radius to the case round 2 fixed.

**And round 2's own "WHAT THIS DOES NOT FIX" list presented itself as complete and did not contain it.** That is the part worth carrying: the enumeration was written by the same person who had just fixed the sibling case, one screen above.

### The fix is not a second guard — it is the rule

**Reclaiming is correct only if the result is DELIVERED, and the extension cannot know that.** The server can: `_record_ownership_locked` runs *only* on a delivered result.

```
extension:  reports `orphanTabId`, closes nothing
server:     _record_ownership_locked → _enqueue_reap_close_locked
```

This **deletes** rather than adds — `startedAt`, the elapsed guard, `reapAttempted`, the `forgetTab` parity argument and round 2's fixture problem all go away, because the party that knows now decides. Net `service_worker.js`: **62+ / 86−**.

Each line of the server-side reap earns its place: **no waiter** (nothing blocks on it; `deliver_result`'s existing `result_after_abandon` branch collects it, so it cannot strand a waiter or a `pending` slot); **an `inflight` entry, deliberately** (that is what tells a BUSY extension from an idle one — without it a `ping` landing mid-reap fast-fails a healthy extension); `orphan != tid`; and `not isinstance(orphan, bool)`, because Python makes `isinstance(True, int)` true and `True == 1`, so a hand-crafted envelope could otherwise reap tab 1.

### 🔴 A hazard this feature creates — disclosed, not fixed

`reference/tabs-instances.md` already records that parallel agents can derive the **same session id and share one owned tab** (observed 2026-07-31). A re-`open` by one sibling whose probe times out now **closes the tab the other sibling is mid-workflow on**. That collision used to be benign. Stated in both operator-facing docs.

### Also from the delta audit

- The round-2 test's fixture **could not see where its clock was anchored** — with the create alone already exceeding `execMs`, "elapsed since op start" and "elapsed since create start" are indistinguishable, and a re-anchoring mutant **SURVIVED**. Moot now (no clock), but the successor uses 90/310/260 — create *inside* the budget, whole op *outside* — and says why.
- `assert.equal(tabsCreate.length, 1, "…and settle")` observed only that the create **ran**: the push happens before its await. Now a `createSettled` flag.
- README's round-2 sentence *"The reap never fires once the op has already timed out"* was **false for the reader most likely to read it** — an operator sees the CALLER's `cmd_timeout`, on which it did fire. Gone with the guard.
- **The node count in my round-2 comment was 511, which is the BASE's count.** It is **512** at that head. The audit caught it; corrected here.

### Tests — 10 new server tests, and the negatives are the point

- a delivered `open` carrying `orphanTabId` enqueues a `close` for it;
- 🔴 an **abandoned** `open` never does — `cmd_timeout=0.3` against an extension answering at 0.8 s, with a control asserting the extension really did complete the open (so the assertion is about abandonment, not about nothing happening), plus `owners_snapshot() == {}` showing *why* the old tab must not be treated as an orphan;
- a **failed** `open` never does — its envelope **carries an orphan report on purpose**, because without a payload the mutant that moves the reap above `if failed: return` finds no `data` and passes for the wrong reason;
- 5 malformed/self-referential reports refused (`== tid`, bool, str, None, float);
- the reap is `inflight` **while it runs** and released after — caught with a gated slow `close`, because asserting only the end state is vacuous.

The extension tests now assert the **negative on every path**: `tabsRemove` stays empty, including on an `execute()`-abandoned op. Put a `chrome.tabs.remove` back into `open` and three tests go red.

### Mutation — round 3

10 mutants, **all KILLED**, each on the test naming its decision. Controls: server 10 passed, node 0 failing.

```
S1 reap call deleted        KILLED  2 tests (the pre-round state)
S2 `!= tid` dropped         KILLED  self-reference param case
S3 bool exclusion dropped   KILLED  a-bool-not-an-int param case
S4 inflight entry dropped   KILLED  in-flight/release test
S5 reap given a waiter      KILLED  in-flight/release test
S6 reap on a FAILED op      KILLED  test_a_failed_open_never_reaps_the_orphan
E1 report unconditional     KILLED  REJECTING-arm negative
E2 field never emitted      KILLED  2 tests
E3 extension closes again   KILLED  3 tests, incl. the abandoned-op negative
E4 reports the wrong id     KILLED  2 tests
```

🔴 **S3, S4 and S6 SURVIVED on first run.** One test snapshotted its dispatch list before the extension had polled *and* waited on the wrong tab id; one asserted only an end state; one could not reach its case at all. All three are why those guards now exist — and all three were defects in tests I had just written to prove the code.

### Ledger

```
round 1 · payload 114 · scaffolding 199 · tekton SUCCESS
round 2 · payload 125 · scaffolding 117 · tekton SUCCESS
round 3 · payload 226 (service_worker.js 62+/86-, server.py 75+/0-, README 24+/16-,
                       protocol.js 19+/11-, handoff 32+/11-, tabs-instances 12+/6-,
                       build_id 1+/1-)
          scaffolding 406 (test_server.py 278+/0-, service_worker.test.mjs 103+/185-,
                           cdp_protocol 13+/12-)
```

Round 3 changed payload, so the attribution gate does not fire — the ladder is still on the PR.

Suites at `cc7f4e59`: node **512 pass / 0 fail** across 15 files; pytest **878 pass / 0 fail** (browser-bridge + doc-path-rot). Marker regenerated `9a3c89153f041614` → `8936861db19e232f`.

ZacxDev:
## Audit round 3 (delta) — the redesign's guard was point-in-time. Fixed in `d243ff11`. **CI green on both gates.**

The delta audit confirmed the round-3 mutation table and suite counts by independent re-derivation, then reproduced two gaps in the redesign — both from one root: `_record_ownership_locked` proves the orphan was unowned **at that instant**, and the `close` it queues runs **later**. Nothing re-validated in between, and nothing bounded the gap.

### 🔴 Cancellation — the server could close a tab it currently owns

`open` is **not** in `TAB_SCOPED_OPS`, so two concurrent opens from one session take no turnstile and both read the same `reuse_tab_id`. Reproduced end-to-end through the real HTTP path:

```
owner_open        session=A tab=202
owner_orphan_reap session=A tab=101      # reap enqueued
owner_open        session=A tab=101      # open #3 healthily REUSES 101
dispatch          op=close  tabId=101    # the reap closes it
→ owners = {('only','A'): 101}, live = [202]
```

The session then owns a tab the server itself closed → `owned_tab_gone` → mapping dropped → the op after that falls back to **the operator's active tab**. Exactly the blast radius round 3 asserts is now impossible. Parallel agents sharing a session id are the realistic driver, and this repo already documents that they do.

`_cancel_queued_reaps_locked` — re-owning a tab withdraws any not-yet-dispatched reap aimed at it. **Residual, stated:** once dispatched there is nothing to withdraw, so the window narrows from unbounded to one poll round trip.

### 🔴 Expiry — the one outbox entry nothing ever removed

Every other command is withdrawn from the outbox by its abandoning submitter at `cmd_timeout`. The reap has no submitter. Reproduced with an injected clock: after a **7200 s** gap it is still dispatched verbatim on reconnect. `instance_id` is stable per profile, so a Brave restart returns the **same `Instance`** while Chromium's tab ids restart at 1 — that queued `close` then names a different, real tab. The correlation is adverse: a browser-wide stall is what *produces* a reap, and a full restart is the documented remedy for a stall.

`expires_at` per reap, checked at pickup in `poll()`, matched to the `INFLIGHT_STALE_S` window its inflight entry already ages out on. An expired reap is **dropped**, never dispatched.

### Also

- **Every successful reap was logging `result_unknown_id`** and answering `{"ok":false,"error":"unknown_id"}` — it has no waiter, so it fell to the abandoned-command branch. That line means *"a reply for a command nobody knows about"*, a lost-reply tell; firing it on correct operation retires the signal. `deliver_result` now recognises a reap via `inst.reaps`.
- 🔴 **A stale prose block in `server.py` survived a whole audit round** — 180 lines above the new code and pointed *at* by it, still describing the extension-closes design, including *"the reclaim is entirely the extension's, so nothing below depends on it."* Flatly inverted, and precisely the sentence that would tell a maintainer reading `submit()` that the server does nothing here. Corrected, with the inversion **left recorded** rather than silently replaced — it is equally easy to reintroduce from either direction.
- The `_enqueue_reap_close_locked` docstring claimed the reap *"cannot overtake one submitted before"*. Not literally true — a submit blocked in `_tab_queues` has not appended yet. Harmless; the sentence was stronger than the code.

### 🔴 A fourth vacuous test, and the audit measured it

*"the orphan is still REPORTED on the abandoned path"* called `OPS.open()` directly, with the default `execMs` and no `execute()` wrapper — the op finished in ~20 ms and **nothing was abandoned**. It was a duplicate of the timeout-arm test under a name claiming coverage it did not have, and the mutant its own comment names (re-guard the emit behind an elapsed check) **survived the whole file**. It now drives a real abandonment and reads `open`'s own promise on the far side of it: 20 ms → **351 ms**. Watched both ways — that mutant now dies, and the trivially-true variant correctly survives as equivalent.

**The three server negatives were bare wall-clock windows.** "No close within 0.4 s" is indistinguishable from "the test looked too early" — the same class this PR has now tripped over four times, and a wider window only makes the vacuous pass load-dependent. Each now calls a `_drain()` FIFO sentinel first: the outbox is FIFO, so a sentinel whose *result* comes back proves everything queued before it was already dispatched.

### Mutation — round 4

7 mutants, **all KILLED**, each on the test naming its decision. Control: 13 passed.

```
T1 cancellation removed        KILLED  cancellation test (the pre-fix state)
T2 cancellation matches  !=    KILLED  cancellation + expiry
T3 expiry check removed        KILLED  expiry test (the pre-fix state)
T4 expiry inverted             KILLED  3 tests
T5 reap-reply branch removed   KILLED  unknown-id test
T6 recognise EVERYTHING        KILLED  same test's unknown-id control
T7 expiry clears ALL reaps     KILLED  expiry test's in-flight assertion
```

### ⚠ Two suite failures that are NOT this PR — established by control, not asserted

- **CI on round 3 went red** on `scripts/dl-router/tests/test_server.py::test_learn_..._for_a_performer` — a directory this PR touches **zero** files in. It passed locally, and **round 4's CI is green with `collected=17816 passed=17814 failed=0`**, so it was a flake.
- **Locally, `test_browser_agent.py` fails nondeterministically** with a 60 s wrapper hang at line 558. **Three runs, three different tests** — and the third was a clean worktree at `origin/main`, which failed the same way. Pre-existing and environment-sensitive (this box is running several concurrent agents). It passes in CI.

### Ledger

```
round 1 · payload 114 · scaffolding 199 · CI SUCCESS
round 2 · payload 125 · scaffolding 117 · CI SUCCESS
round 3 · payload 226 · scaffolding 406 · CI FAILURE (dl-router — unrelated, since shown to be a flake)
round 4 · payload 152 (server.py 138+/30-, README 9+/4-, tabs-instances 5+/1-)
          scaffolding 207 (test_server.py 161+/1-, service_worker.test.mjs 46+/11-)
```

Round 4 changed payload, so the attribution gate does not fire.

**CI at `d243ff11`: `tekton/devrc-nodetests` SUCCESS (1297/1297), `tekton/devrc-pytests` SUCCESS (17814 passed, 0 failed).** Locally: node 512 pass / 0 fail; pytest browser-bridge 803 pass / 1 fail, that one being the pre-existing flake above.

ZacxDev:
## Audit rounds 4 and 5 — round 5's own fix was a net regression, and round 6 (`d119aa6d`) undoes it. **CI green on both gates.**

Two rounds since the last comment. The second one is the more interesting, because the defect was mine and it was introduced *by a fix*.

---

### Round 5 (`c1f4f40e`) — cancellation must not disturb a reap already in flight

`_cancel_queued_reaps_locked` selected from `inst.reaps`, which holds **dispatched** reaps as well as queued ones, removed only the queued ones from the outbox, then popped both maps for *every* match. Reproduced:

```
after dispatch: reaps=1 inflight=1 outbox=0
ping deadline while reap in flight (correct, BUSY): 20.0
after cancel:   reaps=0 inflight=0 outbox=0
ping deadline after cancel: 2.0
deliver_result(reap reply) -> False    # result_unknown_id
```

Both harms re-opened something this feature had just closed: the false wedge the `inflight` line is added to prevent (documented remedy: restart the operator's browser), and the `result_unknown_id` lost-reply signal that `deliver_result`'s reap branch exists to stop emitting.

🔴 **Two mutants survived the first fix's test**, and the reason is the useful part: the early return short-circuits before the pop loops, so with a *single* dispatched reap the two scopes are indistinguishable. The discriminating shape is **one dispatched and one queued reap for the same tab** — reachable (two concurrent opens both resolve `reuseTabId=101`, both probes time out, a poll between the two delivered results dispatches the first). With that fixture both die.

Also round 5: a fourth consecutive superseded-design paragraph, this time in `README.md`'s reuse-bound block, still describing the round-1 extension-closes design **while the same file said the opposite ~600 lines later**. Round 4 fixed the `server.py` copy and missed the README copy.

One assertion was mine being wrong, not the code: I compared two `_effective_timeout_locked` reads for exact equality against the real monotonic clock. Signal ~18 s, noise ~1e-5 s. Frozen clock now.

---

### Round 6 (`d119aa6d`) — 🔴 the round-5 sweep was a NET REGRESSION

Round 5 also added a `reaps` sweep to stop a lost reply stranding an entry. It was inflight-only, and that removed **both** of a reap's bounds in one line.

`inst.reaps[cid]` is the *only* index from an outbox entry to its `expires_at`, and the only thing cancellation matches on. A reap's inflight stamp and its expiry are set from the same instant, so the age-prune fires exactly when the reap expires — and a reap is never in `waiters` (deliberately: "NO WAITER"), so the prune's second condition never shields it. Red at that HEAD, green at round 4's base:

```
after prune: reaps=0 outbox=1
poll()  -> {'op':'close','tabId':101}          → DISPATCHED A STALE CLOSE
cancel  -> 0; outbox still holds ['close']     → UNCANCELLABLE
```

Neither needs a rare event: `_prune_inflight_locked` runs on **every submit and every ping**, and two ordinary ops ahead of the reap in the serial FIFO exceed `INFLIGHT_STALE_S` (28 s) at 2×`EXEC_OP_BUDGET_S` (36 s) with no wedge at all.

Fix: sweep only what **both** bounded lifetimes have finished with — `c not in inst.inflight and c not in queued`.

🔴 **The docstring is why it got in.** `_prune_inflight_locked` said *"Drop `inflight` entries"* and nothing else, so a function whose contract read *inflight-only* was quietly given authority over the reap-expiry index — and I wrote the sweep to match the contract rather than the data. The contract line is now as wide as the body.

Two more from that round:
- `test_a_reap_whose_reply_is_lost_...` only ever exercises the **dispatched** arm while the sweep also sees queued reaps. Reading it as coverage of the sweep is what stopped anyone looking at the other arm; its docstring now states the scope and names the two tests that cover the queued arm.
- The early return in `_cancel_queued_reaps_locked` had **no killing mutation** — deleting it survived the whole selection. Its only observable is a false `owner_orphan_reap_cancel n=0`, which would make that line mean nothing. Now pinned by a log-capture test **with a positive control on the capture itself**, so an empty capture cannot pass for an absent log.

### Mutation

```
round 5 — 6 mutants, all KILLED (control 15)
  U1 pre-fix state · U2/U3 pop-scope widened (SURVIVED until the discriminating
  fixture existed) · U4 sweep removed · U5 sweep too wide · U6 nothing withdrawn

round 6 — 5 mutants, all KILLED (control 18)
  V1 inflight-only predicate (the regression) · V2 queued-only predicate ·
  V3 sweep removed · V4 early return deleted (SURVIVED in round 5) ·
  V5 expiry branch removed
```

### ⚠ A CI flake class that is not this PR — three instances, all process-spawning tests, none in a file this PR touches

| where | test | outcome |
|---|---|---|
| CI round 3 | `scripts/dl-router/…::test_learn_…_for_a_performer` | green on the next run |
| local ×3 | `test_browser_agent.py`, 60 s wrapper hang | three runs, three **different** tests; one reproduced in a clean worktree at `origin/main` |
| CI round 5 | `scripts/tests/test_subsystem_store_api.py::…::test_POSITIVE_CONTROL_…` | that run reported `collected=1781` vs the usual `17816` — it **truncated** rather than completing red; green on the next run |

I included `test_subsystem_store_api.py` in the local round-6 run deliberately: it passes.

### Ledger

```
round 1 · payload 114 · CI SUCCESS
round 2 · payload 125 · CI SUCCESS
round 3 · payload 226 · CI red (dl-router flake → green next run)
round 4 · payload 152 · CI SUCCESS
round 5 · payload 130 · CI red (subsystem-store flake, truncated run → green next run)
round 6 · payload  90 · CI SUCCESS
```

**CI at `d119aa6d`: `devrc-nodetests` SUCCESS (1297/1297); `devrc-pytests` SUCCESS (`collected=17821 passed=17819 skipped=2 failed=0`)** — a full untruncated run, which also retires round 5's red. Locally: node **512** pass / 0 fail; pytest **1294** pass / 0 fail.

PR #958, opened 2026-08-28
feat(audit-pr): assemble the auditor brief in CODE — the invariant clauses stop being retypable

ZacxDev:
## Why

`/audit-pr` says "dispatch a subagent … against this checklist" and supplies **no procedure**, so the brief was reassembled from prose every time. Measured over one session's **14 audit dispatches**:

- **60,100 characters** of brief hand-written (mean 4,292 each)
- **42%** mean similarity between consecutive briefs; **zero** lines >25 chars identical across all 14
- hard-won clauses **accreted** rather than being present from the start, and three were **LOST after being learned**:

```
instruction                      present in dispatch #
                                 1  2  3  4  5  6  7  8  9 10 11 12 13 14
do NOT git fetch            5/14 ·  ·  ·  ·  ·  ·  ·  ·  ✓  ✓  ·  ✓  ✓  ✓
shared checkout warning     9/14 ·  ·  ·  ·  ·  ✓  ✓  ✓  ✓  ✓  ✓  ✓  ✓  ✓
'ending it is CORRECT'      6/14 ·  ·  ·  ·  ·  ·  ·  ✓  ✓  ✓  ·  ✓  ✓  ✓
'a nit is not a finding'    9/14 ·  ·  ·  ·  ✓  ✓  ✓  ✓  ✓  ✓  ·  ✓  ✓  ✓
bare pytest = wrong shell  10/14 ✓  ✓  ✓  ✓  ✓  ✓  ✓  ✓  ✓  ·  ✓  ·  ·  ·
payload/scaffolding label  10/14 ·  ·  ✓  ✓  ✓  ✓  ✓  ✓  ✓  ✓  ·  ·  ✓  ✓
sandbox tier named         11/14 ·  ·  ✓  ✓  ✓  ✓  ✓  ✓  ✓  ·  ✓  ✓  ✓  ✓
```

What actually followed: the auditor at dispatch 8 ran `git fetch` in a repo the brief had called read-only (the clause first appears at 9); the first five auditors reported the shared checkout moving as if it might be their fault; the seven rounds before dispatch 8 ran under a stop rule that never said stopping was the correct outcome.

## What this adds

`scripts/audit-dispatch.py <pr> [--round N] [--repo owner/name] [--out FILE] [--emit-claims]` prints a brief to stdout. **It dispatches nothing and writes nothing to the repo under audit.**

**Generated per invocation** — the parts a human gets wrong:

| fact | how |
|---|---|
| the range | `<previously-audited-sha>..HEAD`, from the prior round's claims block |
| 🔴 cross-repo | the PR's repo vs the cwd's repo decides whether the brief says `git -C <clone> worktree add` **or** `isolation: "worktree"`. That flag worktrees the CWD's repo; for a PR elsewhere it is the wrong tree and it fails quietly |
| shared checkout | branch, dirty-path count, and "it moves under you; report it, don't chase it" |
| toolchain | `nix develop … pytest -p no:cacheprovider`, the `No module named pytest` = wrong-shell diagnosis, `gate.sh --tier both`, and the **sandbox tier named as the one the merge actually gates on** |
| ledger | the skill's own `git log --numstat --format= --remerge-diff <prev>..HEAD --not <base>` |

**Verbatim, from one module-level constant** (`INVARIANT_CLAUSES`): READ-ONLY + `rm -f <copy>/.git`; no `git fetch` in the shared checkout; the stop rule *including* "ending it is the CORRECT outcome, not a failure"; "a nit that changes nothing a reader does is NOT a finding"; re-verify the fix commit's self-reported numbers; `file:line` + a concrete failure scenario + a `payload`/`scaffolding` label; do not merge — report only.

The nine-item checklist is **read from `SKILL.md`, never restated** — one rule, one place; that block has already been silently shaved once under a commit message claiming the opposite.

## Two refusals, deliberately different in severity

1. 🔴 **`--round N` (N≥2) with no parseable claims block REFUSES** (exit 2) and names what it looked for, where it looked, what it found, and how to produce one. An empty "what was claimed fixed" section silently turns a delta re-audit into a **blind full audit** — a different thing that would then read as covered. Same shape as `ladder-depth-sweep.py`'s refused zero.
2. **A brief missing an invariant clause WARNS and never blocks.** It exists for a hand-edited `--out` file; `claude/RULES.md` is explicit that a permanently-red gate trains everyone to click through.

## The framing constraint

A delta round is framed on **what was claimed fixed**, never **why it is correct** — three framed audits confirmed a claim the prompt handed them; one blind audit refuted it in a single pass. PR comments carry both, so only the fenced block is read:

````
```audit-claims round=3 audited=997375ec..9f638fd4
1. run_move collapsed WRONG-KILLER into NOT EXCLUSIVE — now two branches
2. the "97% / most" quantifier over-stated what the measurement supports
```
````

`--emit-claims` prints a skeleton the parser accepts (round-tripped by a test, so it cannot drift into a shape the next round refuses).

## The ledger does NOT classify

It enforces the skill's read rules — **rc 0, silent stderr, non-empty range** — and prints `COULD NOT MEASURE` with **no number** otherwise. It shows the changed-file list and leaves `X` for a human: the classification is a human judgement, and a pathspec is wrong in both directions on ordinary names (`':!*test*'` swallows `attestation/`, `latest/`, `inspector/`; keeps `FooTest.java`, `login.cy.ts`). It **does not fetch** — that would be the same write the brief forbids — and says so, so a stale `<base>` is visible rather than silent.

## Tests

`scripts/tests/test_audit_dispatch.py` — **37, hermetic.** The `gh`/`git` seam is injected and `test_nothing_here_spawns_a_subprocess` proves it by making `subprocess.run` raise for a full `main()` run.

The clause list is pinned **two-way** against an independent ledger in the test module (ids restated by hand — deriving them from the constant would make the whole comparison vacuous), *and* against the **rendered** section, so neither a deletion nor a renderer that stops reading the constant passes.

`scripts/tests/mutants-audit-dispatch.py` is the **committed** negative-control battery — kept in the tree for the reason `mutants-audit-ladder.sh` records: #900's ~30 mutants lived in a scratchpad nobody could re-run. Each row names its **exact** killer set and reports WRONG-KILLER separately from EXTRA-KILLER.

## SKILL.md

Router line added and pinned whole in `test_audit_ladder_stop_rule.py` (11 tests, unchanged count), with a mutant row in `mutants-audit-ladder.sh` that **inverts the refusal into a warning** rather than deleting the line — the reading a reader would accept, describing a different tool.

The body went 81 B over budget with the line added, so the high-yield change-class **anecdotes** were demoted to `reference/round-ladder-evidence.md`. The **class list** — which is what decides whether to audit at all — stays in the body. `skill-audit.py`: within budget, references intact.

### DIFF
--- a/claude/skills/audit-pr/SKILL.md
+++ b/claude/skills/audit-pr/SKILL.md
@@ -24,6 +24,11 @@ generates the range, the cross-repo worktree directive, checkout state and toolc
 prior round's claims from the fenced `audit-claims` block ONLY, and carries the invariant clauses
 verbatim. **A delta round with no parseable block is REFUSED.**
 
+🔴 **Post the round's block with `--round N --emit-claims --audited <the tip that round's audit
+READ>`.** `--emit-claims` runs after the fixes land, so every sha it can see is a FIX tip; omit
+`--audited` and HEAD is *assumed* — it says so on stderr, because the next round then diffs a range
+that is empty by construction and a finding-free pass over it reads as a clean round.
+
 Dispatch a subagent (read-only — it must NOT modify files or merge) to audit the change against this checklist. Have it read the diff and the code it touches, not just the PR description.
 
 **Always run this on high-yield change-classes** — web/HTTP endpoints, concurrency reworks, filesystem/quarantine/trash moves, DB migrations, anything security/auth/path-gating. What each hid, and `GOPRIVATE`: reference file.
--- a/claude/skills/audit-pr/SKILL.md
+++ b/claude/skills/audit-pr/SKILL.md
@@ -19,9 +19,14 @@ Target: `$ARGUMENTS`:
 
 ## What to do
 
+🔴 **Assemble the brief with `~/workspace/devrc/scripts/audit-dispatch.py <pr> [--round N]`** — it
+generates the range, the cross-repo worktree directive, checkout state and toolchain, reads the
+prior round's claims from the fenced `audit-claims` block ONLY, and carries the invariant clauses
+verbatim. **A delta round with no parseable block is REFUSED.**
+
 Dispatch a subagent (read-only — it must NOT modify files or merge) to audit the change against this checklist. Have it read the diff and the code it touches, not just the PR description.
 
-**Always run this on high-yield change-classes** — web/HTTP endpoints, concurrency reworks, filesystem/quarantine/trash moves, DB migrations, anything security/auth/path-gating. These hide deploy-blocking bugs (shutdown data-loss, trash-path overwrite, an unauthenticated arbitrary-path scan and a git `core.fsmonitor` RCE all surfaced this way). A branch with a **private Go module dep** may need `GOPRIVATE` — a sum-db `500` there is env, not a defect.
+**Always run this on high-yield change-classes** — web/HTTP endpoints, concurrency reworks, filesystem/quarantine/trash moves, DB migrations, anything security/auth/path-gating. What each hid, and `GOPRIVATE`: reference file.
 
 **Brief the auditor on the environment, or it will report false findings** — a fresh worktree is
 not a working checkout, and an auditor hitting this cold blames the PR. Whichever apply:
--- a/claude/skills/audit-pr/reference/round-ladder-evidence.md
+++ b/claude/skills/audit-pr/reference/round-ladder-evidence.md
@@ -222,6 +222,22 @@ clause.
 
 ---
 
+## High-yield change-classes — what each one actually hid
+
+Demoted from the skill body to pay for the assembler router line; the CLASS LIST stays in the body
+because it is what decides whether to run an audit at all, and only these anecdotes moved.
+
+Deploy-blocking bugs that surfaced from exactly these classes: a **shutdown data-loss** on a
+concurrency rework; a **trash-path overwrite** on a filesystem/quarantine move; an
+**unauthenticated arbitrary-path scan** on an HTTP endpoint; and a git **`core.fsmonitor` RCE**
+reachable from a repo-local config.
+
+The false positive to expect on the same road: a branch with a **private Go module dep** may need
+`GOPRIVATE`, and a sum-db `500` there is an ENVIRONMENT failure, not a defect in the PR. An auditor
+that reports it as a finding has read a broken toolchain as a broken change.
+
+---
+
 ## Where the rest lives
 
 - The **rejected numeric cap** and devrc #505's ReDoS-introduced-by-the-fix evidence: stated inline
--- a/scripts/audit-dispatch.py
+++ b/scripts/audit-dispatch.py
@@ -71,10 +71,26 @@ round has something to read.
    `scripts/ladder-depth-sweep.py`'s refused zero: the failure and the
    legitimate case produce the SAME observable, so neither may be reported.
 2. **A brief missing an invariant clause WARNS and never blocks.** That check
-   exists for a hand-edited `--out` file, and `claude/RULES.md` is explicit that
-   a permanently-red gate trains everyone to click through it. Warn-only is the
+   exists for a hand-edited brief, and `claude/RULES.md` is explicit that a
+   permanently-red gate trains everyone to click through it. Warn-only is the
    right severity for a channel whose failure is cosmetic and whose false
-   positive rate is set by whatever a human typed.
+   positive rate is set by whatever a human typed. It runs over `--check FILE`
+   and over the READ-BACK of `--out` — never over the in-memory string, where
+   it was unreachable by construction and could not have fired for any input.
+
+🔴 EVERY NUMBER HERE IS ABOUT THE PR, NOT ABOUT YOUR CHECKOUT
+-------------------------------------------------------------
+The delta half of a brief — the ledger, the `<prev>..HEAD` range, the `audited=`
+sha `--emit-claims` writes for the NEXT round to anchor on — was resolved
+against `HEAD` in the operator's own SHARED checkout, with nothing checking that
+HEAD was the PR at all. From a clone standing on an unrelated branch that
+produced rc 0, silent stderr, a non-empty range and a confident ledger of the
+wrong branch's files; from `main` it produced the banner "🔴 Zero changed lines
+over a NON-EMPTY range. That is a real measurement, not a failure."
+
+So `git rev-parse HEAD == headRefOid` is a FOURTH read rule beside rc 0, silent
+stderr and a non-empty range, and all three consumers emit COULD NOT MEASURE
+naming that cause when it does not hold.
 
 🔴 WHAT THIS SCRIPT DOES NOT DO
 --------------------------------
@@ -185,11 +201,23 @@ ISOLATION_FORBID = 'Do NOT use `isolation: "worktree"`'
 # the two fields that matter. A block whose header does not parse is reported as
 # MALFORMED rather than skipped: skipping it silently would produce the same
 # observable as "no block at all", and those need different fixes.
-_CLAIMS_FENCE = re.compile(
-    r"^(?P<fence>`{3,})audit-claims(?P<header>[^\n]*)\n"
-    r"(?P<body>.*?)^(?P=fence)\s*$",
-    re.M | re.S,
-)
+#
+# 🔴 THAT COMMENT WAS FALSE FOR FOUR SHAPES, so the parser is line-based rather
+# than one regex. The old `^(?P=fence)\s*$` backreference required the CLOSING
+# fence to be byte-identical to the opener, and anything it could not match was
+# dropped with no report at all:
+#   * an UNCLOSED fence — the refusal then said "no `audit-claims` block in any
+#     of the 1 comment(s) read", which is false and points at the wrong fix;
+#   * a 4-backtick opener closed with 3 — not a close under CommonMark either,
+#     so this really is unclosed and is now named as such;
+#   * a 3-backtick opener closed with 4 — which IS a valid CommonMark close and
+#     renders closed on GitHub, so it is now PARSED rather than dropped;
+#   * a nested fence inside the body, which ends the block early and silently
+#     drops every claim after it.
+# A claim that WRAPS onto a continuation line was silently truncated to its
+# first line, changing the claim; continuation lines are now appended.
+_FENCE_OPEN = re.compile(r"^(?P<fence>`{3,})audit-claims(?P<header>[^\n]*)$")
+_FENCE_BARE = re.compile(r"^(?P<fence>`{3,})\s*$")
 _HEADER_ROUND = re.compile(r"\bround=(\d+)\b")
 _HEADER_AUDITED = re.compile(r"\baudited=(\S+)")
 _CLAIM_ITEM = re.compile(r"^\s*(\d+)[.)]\s+(.+?)\s*$")
@@ -197,16 +225,66 @@ _CLAIM_ITEM = re.compile(r"^\s*(\d+)[.)]\s+(.+?)\s*$")
 ClaimsBlock = namedtuple("ClaimsBlock", "round_no audited_from audited_to items")
 
 
+def _items_from_body(body_lines):
+    """Numbered claim lines, with CONTINUATION lines folded into the item above.
+
+    A claim wrapped over two lines used to lose everything after the first —
+    which does not fail, it changes the claim, and the next round is framed on
+    the truncated text.
+    """
+    items = []
+    for line in body_lines:
+        m = _CLAIM_ITEM.match(line)
+        if m:
+            items.append(m.group(2))
+        elif items and line.strip():
+            items[-1] = f"{items[-1]} {line.strip()}"
+    return items
+
+
 def parse_claims_blocks(texts):
     """-> (blocks, malformed_reasons) over an iterable of comment bodies.
 
     Pure and independently testable: the refusal below is only trustworthy if
     this can be driven with no network and no PR.
+
+    🔴 `malformed` is NOT only populated when the block is unusable. A structural
+    problem that still leaves a parseable block (a nested fence that cut the
+    claims in half) is reported too, and `main` warns about it rather than
+    letting a half-read block pass as a whole one.
     """
     blocks, malformed = [], []
     for text in texts:
-        for m in _CLAIMS_FENCE.finditer(text or ""):
-            header = m.group("header")
+        lines = (text or "").splitlines()
+        i = 0
+        while i < len(lines):
+            om = _FENCE_OPEN.match(lines[i])
+            if om is None:
+                i += 1
+                continue
+            opener, header = om.group("fence"), om.group("header")
+
+            body, close_at, j = [], None, i + 1
+            while j < len(lines):
+                bm = _FENCE_BARE.match(lines[j])
+                # CommonMark: a closing fence must be AT LEAST as long as the
+                # opener. A shorter run of backticks is content, not a close.
+                if bm and len(bm.group("fence")) >= len(opener):
+                    close_at = j
+                    break
+                body.append(lines[j])
+                j += 1
+
+            if close_at is None:
+                malformed.append(
+                    "an `audit-claims` fence that is never CLOSED — no line of "
+                    f"{len(opener)} or more backticks follows it (header was: "
+                    f"`{header.strip() or '<empty>'}`). A shorter closing fence "
+                    "does not close it, under CommonMark or here."
+                )
+                i += 1
+                continue
+
             r = _HEADER_ROUND.search(header)
             a = _HEADER_AUDITED.search(header)
             if not r or not a:
@@ -215,26 +293,54 @@ def parse_claims_blocks(texts):
                     f"`round=<n>` and `audited=<sha>..<sha>` (header was:"
                     f" `{header.strip() or '<empty>'}`)"
                 )
+                i = close_at + 1
                 continue
+
             spec = a.group(1)
             frm, _, to = spec.partition("..")
             if not to:
-                # A bare sha is accepted as the audited TIP; the round-1 anchor
-                # is then simply unknown, which the ledger reports rather than
-                # guesses.
+                # A bare sha is accepted as the audited TIP; for a round-1 block
+                # that tip IS the cumulative anchor (see `round_one_anchor`),
+                # and for any later round the anchor is simply unknown, which
+                # the ledger reports rather than guesses.
                 frm, to = "", frm
-            items = [
-                mm.group(2)
-                for mm in (_CLAIM_ITEM.match(ln) for ln in m.group("body").splitlines())
-                if mm
-            ]
+
+            items = _items_from_body(body)
             if not items:
                 malformed.append(
                     f"an `audit-claims round={r.group(1)}` block with no "
                     "numbered claim lines in its body"
                 )
+                i = close_at + 1
                 continue
+
+            # 🔴 The nested-fence report. A fence inside the body ends the block
+            # early — CommonMark and GitHub agree — so the claims after it are
+            # outside it and were dropped with no signal. The tell is claim
+            # line(s) AND a stray fence in the region after the close, which is
+            # what a cut-in-half block leaves behind. Deliberately a report and
+            # not a rejection: the block that DID parse is still used, and a
+            # comment whose trailing prose merely happens to contain both would
+            # otherwise lose a usable block to a heuristic.
+            k = close_at + 1
+            tail = []
+            while k < len(lines) and _FENCE_OPEN.match(lines[k]) is None:
+                tail.append(lines[k])
+                k += 1
+            if any(_CLAIM_ITEM.match(t) for t in tail) and any(
+                _FENCE_BARE.match(t) for t in tail
+            ):
+                malformed.append(
+                    f"an `audit-claims round={r.group(1)}` block that looks CUT "
+                    "SHORT by a nested fence: numbered claim line(s) and a "
+                    "stray fence follow its closing fence, so claims after the "
+                    f"nested fence were not read (read {len(items)}). Indent "
+                    "any code inside a claim, or open the block with four "
+                    "backticks."
+                )
+
             blocks.append(ClaimsBlock(int(r.group(1)), frm, to, items))
+            i = close_at + 1
     return blocks, malformed
 
 
@@ -248,14 +354,29 @@ def newest_block(blocks):
 
 
 def round_one_anchor(blocks):
-    """The left side of the LOWEST-numbered block's range, or None.
+    """The sha ROUND 1 audited, or None.
 
     That is the only sha in the corpus that can anchor the ledger's cumulative
     "since round 1" figure. When no block carries one, the figure is reported
     NOT MEASURED — never derived from the base, which is a different quantity.
+
+    Two spellings carry it, and both are read:
+      * `round=2 audited=A..B` — A is the tip round 1 audited (the usual case);
+      * `round=1 audited=A`    — the bare form `--emit-claims` writes for a
+        round-1 comment, where the audited TIP is itself that same quantity.
+    A bare sha on any LATER round is NOT an anchor: it says what that round
+    audited, and the round-1 tip is then genuinely unknown.
     """
-    ordered = sorted((b for b in blocks if b.audited_from), key=lambda b: b.round_no)
-    return ordered[0].audited_from if ordered else None
+    candidates = []
+    for b in blocks:
+        if b.audited_from:
+            candidates.append((b.round_no, b.audited_from))
+        elif b.round_no == 1 and b.audited_to:
+            candidates.append((b.round_no, b.audited_to))
+    # Stable sort on the round number alone, so a tie resolves to the block seen
+    # first rather than to whichever sha sorts lower.
+    candidates.sort(key=lambda c: c[0])
+    return candidates[0][1] if candidates else None
 
 
 # --------------------------------------------------------------------------- #
@@ -274,12 +395,18 @@ def real_runner(cmd, cwd=None):
 # Facts gathered per invocation
 # --------------------------------------------------------------------------- #
 
-LedgerReport = namedtuple("LedgerReport", "files added deleted commits reason cumulative")
+LedgerReport = namedtuple(
+    "LedgerReport",
+    "files added deleted commits reason cumulative cumulative_reason",
+)
+# 🔴 `head_check` is the FOURTH read rule, beside rc 0 / silent stderr /
+# non-empty range. See `verify_head_is_the_pr`.
+HeadCheck = namedtuple("HeadCheck", "ok reason local_sha pr_sha")
 Facts = namedtuple(
     "Facts",
-    "pr repo title base_ref url round_no cwd_repo_dir cwd_repo_slug cross_repo "
+    "pr repo title base_ref url round_no cwd_repo_dir cwd_repo_slug repo_relation "
     "branch dirty prev_sha claims claims_round checklist ledger assembled_at "
-    "claims_source",
+    "claims_source head_check",
 )
 
 
@@ -311,9 +438,18 @@ def gather_checkout_state(runner, repo_dir):
 
 
 def gh_pr_facts(runner, pr, repo=None):
-    """PR metadata + comment bodies, via `gh`. -> (dict, [comment bodies])."""
+    """PR metadata + comment bodies, via `gh`. -> (dict, [comment bodies]).
+
+    🔴 `headRefOid` and `isCrossRepository` are read because the two decisions
+    that were WRONG without them are the two most expensive ones this script
+    makes: which tree the ledger measured, and which repository the PR lives in.
+    🔴 There is NO `baseRepository` field on `gh pr view --json` (checked
+    against `gh`'s own field list) — `url` is what carries the base repo, and
+    `pr_slug` reads it.
+    """
     cmd = ["gh", "pr", "view", str(pr), "--json",
-           "title,url,baseRefName,headRepository,headRepositoryOwner,comments"]
+           "title,url,baseRefName,headRefOid,isCrossRepository,"
+           "headRepository,headRepositoryOwner,comments"]
     if repo:
         cmd += ["--repo", repo]
     rc, out, err = runner(cmd)
@@ -330,31 +466,126 @@ def gh_pr_facts(runner, pr, repo=None):
     return data, comments
 
 
+# `https://<host>/<owner>/<name>/pull/<n>` — the PR's own URL, which names the
+# repository the PR LIVES IN. Host-agnostic on purpose (GHES spells the host
+# differently and the path shape is the same).
+_PR_URL = re.compile(r"^https?://[^/]+/([^/]+)/([^/]+)/pull/\d+")
+
+
 def pr_slug(data, override=None):
+    """The repo the PR LIVES IN — NOT the repo its head branch lives in.
+
+    🔴 This read `headRepositoryOwner`/`headRepository` and was therefore wrong
+    for every FORK PR. Verified against real `gh` output for a fork PR against
+    `cli/cli`: `headRepository.nameWithOwner` is `ylfeng250/cli` while `url` is
+    `https://github.com/cli/cli/pull/14280` and `isCrossRepository` is `true`.
+    A fork PR opened against THIS repo therefore computed a slug that differs
+    from the cwd's, took the CROSS-REPO branch, and told the agent to worktree
+    "your local clone of <contributor>/<repo>" — a clone that does not exist,
+    for a repository the PR is not in.
+
+    `gh pr view --json` has no `baseRepository` field, so `url` is the authority
+    and `isCrossRepository` is what says whether the head fields may stand in
+    for it: they are the same repo only when it is false.
+    """
     if override:
         return override
-    owner = (data.get("headRepositoryOwner") or {}).get("login")
-    name = (data.get("headRepository") or {}).get("name")
-    return f"{owner}/{name}" if owner and name else None
+    m = _PR_URL.match((data.get("url") or "").strip())
+    if m:
+        return f"{m.group(1)}/{m.group(2)}"
+    if data.get("isCrossRepository") is False:
+        owner = (data.get("headRepositoryOwner") or {}).get("login")
+        name = (data.get("headRepository") or {}).get("name")
+        if owner and name:
+            return f"{owner}/{name}"
+    # 🔴 None, never a guess. `main` renders a third WHERE-TO-WORK branch for
+    # this, because collapsing "cannot tell" into "same repo" recommends the
+    # flag that is dangerous in exactly the case it cannot rule out.
+    return None
+
+
+def verify_head_is_the_pr(runner, repo_dir, pr_head_sha):
+    """🔴 The FOURTH read rule: is this checkout's HEAD the PR's head commit?
+
+    Everything the delta half of this brief says — the ledger's numbers, the
+    `<prev>..HEAD` range it hands the auditor, and the `audited=` sha
+    `--emit-claims` stamps for the NEXT round to anchor on — is computed against
+    `HEAD` in the operator's own checkout. That checkout is SHARED and moves;
+    nothing here ever checked that it was standing on the PR at all.
+
+    Reproduced from a clone standing on an unrelated feature branch: rc 0,
+    silent stderr, a non-empty range — all three advertised read rules satisfied
+    — and a ledger of that branch's files. Standing on `main` instead produced
+    the confident banner "🔴 Zero changed lines over a NON-EMPTY range. That is
+    a real measurement, not a failure."
+
+    So a failed check returns a `reason` and NO number, exactly like the other
+    three: an unverifiable measurement is not a measurement.
+    """
+    if not pr_head_sha:
+        return HeadCheck(
+            False,
+            "the PR's head sha is not known here (`gh` was not consulted — "
+            "`--claims-file` mode — or reported no `headRefOid`), so nothing "
+            "can confirm this checkout is standing on the PR",
+            None,
+            None,
+        )
+    rc, out, err = runner(["git", "-C", repo_dir, "rev-parse", "HEAD"])
+    local = out.strip()
+    if rc != 0 or not local:
+        return HeadCheck(
+            False,
+            f"`git rev-parse HEAD` in {repo_dir} exited {rc}: "
+            f"{(err or out).strip() or 'no output'}",
+            None,
+            pr_head_sha,
+        )
+    if local != pr_head_sha:
+        return HeadCheck(
+            False,
+            f"this checkout's HEAD is `{local}`, but the PR's head is "
+            f"`{pr_head_sha}` — the two are DIFFERENT COMMITS, so anything "
+            "measured against `HEAD` here is a measurement of another tree",
+            local,
+            pr_head_sha,
+        )
+    return HeadCheck(True, None, local, pr_head_sha)
 
 
 # --------------------------------------------------------------------------- #
 # The ledger — the skill's own command, with the skill's own read rules
 # --------------------------------------------------------------------------- #
 
-def measure_ledger(runner, repo_dir, prev_sha, base):
+def measure_ledger(runner, repo_dir, prev_sha, base, head_check=None):
     """`git log --numstat --format= --remerge-diff <prev>..HEAD --not <base>`.
 
-    🔴 The skill's read rules are enforced here, not assumed: **rc 0, silent
-    stderr, and a non-empty range**. A missing ref or a git without
-    `--remerge-diff` exits 128 with empty output; an unwritable object store
-    makes `--remerge-diff` UNDER-count, exit 0 and print a plausible number,
-    announcing itself only on stderr; and a range whose commits are not in this
-    checkout prints nothing at all, silently, with rc 0. Each of those returns a
-    `reason` and NO number — a failed command is not a zero.
+    🔴 FOUR read rules are enforced here, not assumed: **the checkout's HEAD is
+    the PR's head, rc 0, silent stderr, and a non-empty range**.
+
+    The fourth one came last and is the one that lets the other three pass while
+    the answer is entirely wrong: `HEAD` is resolved in the operator's own
+    SHARED checkout, which is not necessarily standing on the PR. Reproduced
+    from a clone on an unrelated branch — rc 0, silent stderr, a non-empty range
+    and a file list belonging to that branch. `head_check` is
+    `verify_head_is_the_pr`'s verdict; passing None skips it, which is only for
+    a caller that has no PR to compare against.
+
+    A missing ref or a git without `--remerge-diff` exits 128 with empty output;
+    an unwritable object store makes `--remerge-diff` UNDER-count, exit 0 and
+    print a plausible number, announcing itself only on stderr; and a range
+    whose commits are not in this checkout prints nothing at all, silently, with
+    rc 0. Each of those returns a `reason` and NO number — a failed command is
+    not a zero, and neither is a measurement of the wrong tree.
     """
     def fail(reason):
-        return LedgerReport(None, None, None, None, reason, None)
+        return LedgerReport(None, None, None, None, reason, None, None)
+
+    if head_check is not None and not head_check.ok:
+        return fail(
+            "this checkout is not standing on the PR, so `..HEAD` does not "
+            f"mean what this ledger would claim it means: {head_check.reason}"
+        )
 
     rc, out, err = runner(
         ["git", "-C", repo_dir, "rev-list", "--count", f"{prev_sha}..HEAD"]
@@ -401,7 +632,7 @@ def measure_ledger(runner, repo_dir, prev_sha, base):
         files[path] = (cur[0] + na, cur[1] + nd)
         added += na
         deleted += nd
-    return LedgerReport(files, added, deleted, commits, None, None)
+    return LedgerReport(files, added, deleted, commits, None, None, None)
 
 
 # --------------------------------------------------------------------------- #
@@ -425,8 +656,51 @@ def render_worktree_directive(facts):
     repo that is the wrong tree, and the failure is quiet: the agent either
     reports a briefed file missing, or silently audits the wrong repository and
     reports findings about it.
+
+    🔴 THREE states, not two. The decision used to be `bool(cwd_slug and repo
+    and cwd_slug != repo)`, so "could not determine either side" evaluated
+    FALSE and collapsed into the SAME-REPO branch — the one that recommends the
+    flag. With no `origin` remote the output contradicted itself in one
+    paragraph: "The PR lives in `<org>/<other-repo>`, which is this session's
+    own repository (`/home/…/devrc`)", followed by the recommendation. Not
+    knowing is its own answer, and it is the answer that must NOT recommend a
+    flag whose failure mode is silent.
     """
-    if facts.cross_repo:
+    if facts.repo_relation == "unknown":
+        unknown_side = (
+            "this checkout has no `origin` remote to resolve a slug from"
+            if not facts.cwd_repo_slug else
+            "`gh` did not report which repository the PR lives in"
+        )
+        return "\n".join([
+            "## WHERE TO WORK — 🔴 COULD NOT DETERMINE — decide by hand",
+            "",
+            "This script could not establish whether the PR is in THIS "
+            f"session's repository or another one: {unknown_side}.",
+            "",
+            f"    the PR's repo     : {facts.repo}",
+            f"    this checkout     : {facts.cwd_repo_dir}"
+            + (f" (`{facts.cwd_repo_slug}`)" if facts.cwd_repo_slug else
+               " (no `origin` remote)"),
+            "",
+            f"🔴 {ISOLATION_FORBID} until that is answered. The flag builds its "
+            "worktree from the CWD's repo; if that is the wrong repo it fails "
+            "quietly — the agent either reports a briefed file missing or "
+            "silently audits the wrong tree — and this script cannot currently "
+            "rule that out. **Not knowing is not the same as same-repo**, and "
+            "the same-repo branch is the one that recommends the flag.",
+            "",
+            "Answer it, then follow the matching directive:",
+            "",
+            "```",
+            f"gh pr view {facts.pr} --json url          # the repo the PR lives in",
+            f"git -C {facts.cwd_repo_dir} remote get-url origin",
+            "```",
+            "",
+            "Or re-run this script with `--repo owner/name` to state the PR's "
+            "repository outright.",
+        ])
+    if facts.repo_relation == "cross":
         return "\n".join([
             "## WHERE TO WORK — 🔴 CROSS-REPO",
             "",
@@ -496,12 +770,36 @@ def render_range(facts):
             "",
             f"Base branch: `{facts.base_ref}`.",
         ])
+    # 🔴 `..HEAD` is only meaningful once HEAD has been shown to BE the PR's
+    # head. Unverified, the range handed to the auditor points at whatever tree
+    # the shared checkout happens to be standing on — reproduced against an
+    # unrelated feature branch, where it read as an ordinary non-empty range.
+    hc = facts.head_check
+    if hc is not None and hc.ok:
+        tip = "HEAD"
+        note = (
+            f"This checkout's HEAD is `{hc.local_sha}`, verified at assembly "
+            f"time to be PR #{facts.pr}'s head commit — which is what makes "
+            "`..HEAD` mean the PR here."
+        )
+    else:
+        tip = hc.pr_sha if (hc is not None and hc.pr_sha) else "<the PR's head sha>"
+        note = (
+            "🔴 **COULD NOT VERIFY that this checkout is standing on the PR**, "
+            "so `..HEAD` is NOT used above — it would name whatever tree the "
+            "shared checkout is on:\n\n"
+            f"    {hc.reason if hc is not None else 'no check was made'}\n\n"
+            "Resolve the range in a tree that CONTAINS the PR's head before "
+            "trusting any number derived from it."
+        )
     return "\n".join([
         "## THE RANGE",
         "",
         f"A DELTA re-audit, round {facts.round_no}. Diff **`{facts.prev_sha}"
-        "..HEAD`** — the fix commits made since the tip round "
+        f"..{tip}`** — the fix commits made since the tip round "
         f"{facts.claims_round} audited. Do not re-audit the whole PR.",
+        "",
+        note,
         "",
         f"Base branch: `{facts.base_ref}`.",
         "",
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Comment by ZacxDev** [comment]
## Round 1 claims — machine-readable, for the assembler this PR adds

This block exists to exercise the path the build could not: `--round 2` has never run against a real PR carrying a claims block, because none existed. It does now.

```audit-claims round=1 audited=abc41024
1. the invariant clauses live in one constant and are emitted verbatim, so a clause cannot be silently dropped the way three were across 14 hand-written dispatches
2. a delta round with no parseable claims block REFUSES to emit rather than silently degrading into a blind full audit
3. only the fenced claims block is read; the surrounding comment prose never reaches the brief, which keeps the framed-audit rule mechanical
4. the cross-repo decision is computed, not recalled: a PR in another repo gets `git -C <clone> worktree add`, never `isolation: "worktree"`
5. the ledger enforces rc 0, silent stderr and a non-empty range, printing COULD NOT MEASURE and no number otherwise
6. payload/scaffolding classification is left to the human, and the nine-item checklist is read from SKILL.md rather than restated
```

The prose above and below the block is deliberate: a correct reader takes the six numbered claims and none of this sentence.

ZacxDev:
## Round 2 fix round — all 9 findings fixed (`abc41024` → `d9eb36a8`)

3 🔴 and 6 🟡, every one reproduced by the auditor and every one fixed with a
test **watched RED at `abc41024`** and a mutant proving that test executes.

**Red-at-base measurement.** `git show abc41024:scripts/audit-dispatch.py` into a
scratch tree with the new test module copied in unchanged, under
`PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`: **24 failed, 34 passed** (24 node
ids, 21 functions). Those 21 are `RED_AT_BASE`; the ledger names three of them
inline as *weaker* evidence than the rest, because they assert text the fix added
rather than observing a wrong answer.

### The matrix

| # | finding | detector (red at `abc41024`) | mutant |
|---|---|---|---|
| 🔴 1a | ledger measured the OPERATOR'S checkout | `..._ledger_refuses_to_measure_a_checkout_that_is_not_the_pr` | H1 |
| 🔴 1b | range handed out `..HEAD` from the wrong tree | `..._range_does_not_hand_out_a_head_that_is_not_the_prs_head` | H2 |
| 🔴 1c | `--emit-claims` stamped the LOCAL head | `..._emit_claims_stamps_the_prs_head_not_the_local_checkouts` | H3 |
| 🔴 2 | toolchain gated the SHARED checkout | `..._toolchain_gates_the_auditors_copy_not_the_shared_checkout` | T1, T2 |
| 🔴 3 | fork PRs inverted the cross-repo directive | `..._fork_pr_against_this_repo_is_not_treated_as_cross_repo` + `..._prs_repo_is_read_from_the_url_not_from_the_head_repo` | C3, P1 |
| 🟡 4 | `missing_clauses()` unreachable | `..._clause_check_runs_over_a_file_that_can_actually_be_lossy` + `..._out_file_is_read_back_and_checked` | K1, K2 |
| 🟡 5 | five clause rewords INVERTED the instruction | `..._each_clause_carries_the_instruction_its_ledger_entry_names` | W1–W5, **S1** |
| 🟡 6 | false cause for a failed cumulative measurement | `..._failed_cumulative_measurement_does_not_print_a_false_cause` | L1 |
| 🟡 7 | `--round 1 --emit-claims` emitted garbage | `..._emit_claims_prints_a_block_..._parser_accepts` (rows 2, 3) + `..._round_one_block_anchors_...` | H4 |
| 🟡 8 | parser dropped four shapes silently | `..._fence_the_parser_cannot_read_is_reported_not_skipped` + 4 siblings | B1–B5 |
| 🟡 9 | "cannot determine" collapsed into SAME-REPO | `..._undeterminable_repo_gets_its_own_branch_not_the_same_repo_one` | P2 |

**Mutation battery: 39/39 rows as expected** (16 pre-existing + 23 new), POS
control 58 passed. `S1` is a new **SURVIVES** control — re-space a clause, suite
must stay green — proving the whole-string pins absorb a re-wrap and only fail on
a reword. Five pre-existing rows gained a killer (a genuine coupling, recorded
the way `D2` already was); `C3` reported **MUTATION DID NOT APPLY** the moment
`cross_repo` became `repo_relation` and was re-targeted — without that assert it
would have scored as a guard holding.

### 🔴 The audit was wrong on one point, and it matters for the next reader

Finding 3 prescribed `isCrossRepository` + **`baseRepository`**. **`gh pr view
--json` has no `baseRepository` field** — verified against `gh`'s own field list.
`url` is what carries the base repo (confirmed on a real fork PR: head
`ylfeng250/cli`, `url` `.../cli/cli/pull/14280`, `isCrossRepository: true`), so
`pr_slug` reads `url` first and falls back to the head fields only when
`isCrossRepository` is `false`.

### Live re-verification — the original symptom, on the original path

Run from the shared checkout standing on `main`, which is the auditor's own
reproduction:

```
🔴 **COULD NOT VERIFY that this checkout is standing on the PR** …
    this checkout's HEAD is `af6ba68c…`, but the PR's head is `d9eb36a8…`
🔴 **COULD NOT MEASURE** — and a failed command is NOT a zero …
```

Previously that path printed *"🔴 Zero changed lines over a NON-EMPTY range. That
is a real measurement, not a failure."* Run from a checkout that IS on the PR, the
same command verifies the head and measures the ledger — and the cumulative
figure now reads **2069 instead of NOT MEASURED**, because `round_one_anchor` reads
this PR's own round-1 `audited=abc41024` block, which is exactly the bare form
finding 7 made `--emit-claims` produce.

### 🟢 10 — noted, not actioned

`claude/skills/audit-pr/SKILL.md` is **12,033 B** against an advisory 12,038 B
budget that nothing gates (`test_skill_size.py` covers `browser` only). This
round added no bytes to it, deliberately — the next router line still blows the
budget with no red test, and closing that needs its own ratchet test rather than
a line in this PR.

### Gates

- dev-host `nix develop <repo> -c bash <worktree>/scripts/gate.sh --tier both` →
  **`GATE: RESULT=PASS exit=0`**; pytest 17,882 passed / 2 skipped / 0 failed
  (floor 15,970), node 1,292 passed / 0 failed (floor 1,239). Base
  `af6ba68c` + this branch.
- sandbox tier (**the one the merge gates on**)
  `nix build <worktree>#checks.x86_64-linux.pytests` → **rc 0**,
  `RESULT: PASS (exit=0)`, collected 17,884 / passed 17,882 / skipped 2 /
  failed 0; `…nodetests` → **rc 0**, `RESULT: PASS (exit=0)`, 1,292 pass / 0
  fail. Both at `d9eb36a8`.
  ⚠ The exit status is captured directly, not through a pipe: the first attempt
  ran `nix build … | tail -20; echo "PY_RC=$?"` and printed **`PY_RC=0` under
  `error: interrupted by the user`** — `$?` was `tail`'s. That run is discarded;
  the numbers above come from runs whose status was read from `nix` itself.
- `scripts/tests/mutants-audit-ladder.sh` → **18 rows, all as expected**.

### Not verified / still open

- **A block in a REVIEW comment or the PR BODY is still invisible.** `gh pr view
  --json comments` returns issue comments only. The refusal now SAYS so; the gap
  is documented, not closed.
- The nested-fence report is a **heuristic** (claim line + stray fence after the
  close). It warns and never drops the block that parsed, so a false positive
  costs a line of stderr.
- Nothing here proves an auditor's BEHAVIOUR changed — same blind spot the module
  already records.

```audit-claims round=2 audited=abc41024..d9eb36a8
1. the ledger, the `<prev>..HEAD` range and `--emit-claims` now require `git rev-parse HEAD == headRefOid` as a FOURTH read rule beside rc 0 / silent stderr / non-empty range, and each emits COULD NOT MEASURE naming that cause when it does not hold
2. the toolchain section's `gate.sh` path and `nix build` flake ref are `<your worktree>` placeholders; only `nix develop <repo>` still names the operator's checkout, and only to resolve the dev shell
3. the PR's repository is read from `url` rather than the head repo, so a fork PR opened against this repo takes the SAME-REPO branch; the `DEFAULT_PR` fixture was corrected and a fork-PR fixture added
4. `--check FILE` and a read-back of `--out` give `missing_clauses()` inputs that can actually be lossy, so the guard is reachable outside a monkeypatched stub
5. `CLAUSE_LEDGER` pins whole whitespace-normalised clauses instead of fragments, with the five instruction-inverting rewords as mutant rows W1-W5 and a SURVIVES re-wrap control S1
6. a failed cumulative measurement carries `cum.reason` through instead of printing the false "no `audit-claims` block carried a round-1 anchor sha" cause
7. `--round 1 --emit-claims` emits the bare `audited=<sha>` form the parser accepts, and `round_one_anchor` reads a round-1 bare sha as the cumulative anchor
8. the claims parser is line-based: an unclosed or short-closed fence is reported MALFORMED, a longer closing fence is read as the valid CommonMark close it is, a nested fence is reported, continuation lines are folded into their claim, and the refusal names the comment kinds `--json comments` cannot see
9. "the repo cannot be determined" is a third state with its own COULD NOT DETERMINE branch instead of collapsing into SAME-REPO and recommending `isolation: "worktree"`
```

ZacxDev:
Round 3 fixes are pushed as `e06461f7`.

```audit-claims round=3 audited=d9eb36a8..e06461f7
1. the delta range now anchors on `audited_from` (the tip the previous round AUDITED), not `audited_to`, so `<prev>..HEAD` is no longer empty by construction
2. `--emit-claims` keeps writing `audited_to` as the next block's `<from>`, via a separate named reader, so the two anchors cannot be collapsed into one
3. a bare round-1 `audited=<sha>` still anchors the round-2 range, by falling back to `<to>` when `<from>` is absent
4. THE RANGE section refuses a degenerate self-range with a loud banner naming the stop-rule trap, instead of rendering it under "verified at assembly time"
5. the ledger reports a self-range as EMPTY BY CONSTRUCTION, distinct from an ordinary empty range
6. `--emit-claims` warns when the block it is about to write has the same sha on both sides
7. the empty-range COULD NOT MEASURE no longer names causes that `head_check` refutes
8. the same false-cause shape was fixed at a THIRD site found by sweeping the class: the unknown-head-sha reason in `verify_head_is_the_pr`
9. three verbatim instruction blocks moved into a second whole-string ledger, `SECTION_DIRECTIVES`, pinned two-way
10. the round-2 fix matrix moved from prose into `FIX_MATRIX` data, graded against the two ledgers, correcting the one wrong "red at abc41024" evidence label
11. `make_runner`'s `rev_list` became a function of the range spec, so the degenerate case is observable at all, and records every range requested
```

**Evidence.** Red at `d9eb36a8` (`git show d9eb36a8:scripts/audit-dispatch.py` into a scratch tree with this test module copied in unchanged, `PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`): **14 failed, 58 passed**. Nine of those fourteen are filed as regression coverage; the other five fail with `AttributeError` for want of a name the fix introduced, which is not the defect being observed, so they are filed as guards with the mutation battery as their evidence. That gap is written down in the module rather than left to a reader.

**Mutation battery** (`scripts/tests/mutants-audit-dispatch.py`): **52 rows, all as expected**, positive control 72 passed. It caught two defects in this round's own tests — `N5` SURVIVED because the ledger assertion was reading the whole brief and matching THE RANGE section's banner, and `H3`/`C2` returned EXTRA-KILLER against rows that duplicated a neighbour's assertion. `H2` reported MUTATION DID NOT APPLY (its branch became an `elif`) and was re-targeted. `scripts/tests/mutants-audit-ladder.sh`: 18 rows, all as expected.

**On the `<from>` sha.** The fix-round brief suggested `abc41024`; the corrected reader says `d9eb36a8`, and `--round 3 --emit-claims` emits exactly that. `<from>` is the tip the round's AUDIT READ, not the start of the range it was handed: round 3 read the tree at `d9eb36a8` over the range `abc41024..d9eb36a8`. Writing `abc41024` here would make round 4 re-audit round 2's fix commits as well.

ZacxDev:
Round 4 fixes are pushed: `e06461f7..dd601793`.

⚠ **`<from>` is `e06461f7`, not `d9eb36a8`, and the difference is deliberate.** The
round-4 dispatch named `d9eb36a8` as "the tip round 4 audited" — that is the sha round 4
*diffed from*, which is the round-3 block's `<from>`. The tip round 4's audit **read** is
`e06461f7`, the round-3 block's `<to>`, and that is what `emit_anchor` returns. Checked
against the corrected reader before posting: `--round 4 --emit-claims` prints exactly
`audited=e06461f7..dd601793`, silent stderr. Writing `d9eb36a8` here would make round 5
re-audit round 3's fix commits as well — the superset `emit_anchor` exists to prevent —
and round 5's range would not be "the round-4 fix commits".

Verified end to end from a checkout at `dd601793`: `--round 5` renders
``Diff `e06461f7..HEAD` `` over **3 commits**, not an empty range, with no
DEGENERATE banner. And the round-1 half, on a PR carrying no claims block at
all (#967): with `--audited` omitted it emits the bare `audited=a8e15b5a` and
warns that HEAD was **ASSUMED**, naming the flag; with `--audited 51486abf` it
emits `audited=51486abf..a8e15b5a` and is silent. At `e06461f7` that same run
printed nothing at all.

Gates and evidence are in the commit message; the short version: module suite 83 passed
(was 72); watched red at `e06461f7` — 15 failed, 68 passed, of which **five** are filed as
regression coverage and the other ten as guards, each with the mutant that proves it
executes; `mutants-audit-dispatch.py` 67 rows all as expected (was 52);
`mutants-audit-ladder.sh` 18 rows all as expected. Gates, with the tier and base
named: dev-host `scripts/gate.sh --tier both` → `GATE: RESULT=PASS exit=0` (pytest
17,909 collected / 0 failed, node 1,292 / 0 failed) at `89facc6b`; sandbox
`nix build .#checks.x86_64-linux.{pytests,nodetests}` → both `RESULT: PASS (exit=0)`
at `dd601793`, nix's own status read directly rather than through a pipe.

🔴 Two more claims were found by re-reading this round's OWN scaffolding and are fixed
in `dd601793`: a guard docstring that claimed a rendering check it does not make, and a
floor comment whose arithmetic was wrong in both numbers (47 rows, not 48; floor 45,
not 46).

```audit-claims round=4 audited=e06461f7..dd601793
1. the round-1 `--emit-claims` self-range warning now asks the anchor the NEXT round will use, so it is no longer structurally unreachable for the one spelling it was written for
2. `--audited <sha>` supplies the tip the round AUDITED, which `--emit-claims` cannot derive; omitted, HEAD is ASSUMED and stderr says so and names the flag
3. `range_anchor`'s bullets and `round_one_anchor`'s caveat no longer contradict each other about what a bare round-1 `audited=<sha>` means
4. both degenerate-range messages stopped offering "the round's fix commits are not in this checkout" — a cause `anchor_is_head`'s own precondition refutes
5. that cause list has ONE writer, pinned whole as the `degenerate-range-causes` directive, plus a ledger of every site that renders it which fails when the set grows or shrinks
6. Y1 (the sentence that defines the delta scope) and Y2 (the cross-repo "that worktree is YOURS") are pinned whole as directives
7. the "which blocks are NOT pinned" comment no longer claims fragment coverage that does not exist, the 7-of-11 arithmetic is gone, and Y3-Y5 are recorded as DOCUMENTED HOLES with their own mutant expectation
8. THE CHECKOUT section is three-state — shared / private worktree / COULD NOT DETERMINE — decided by `git rev-parse --git-dir --git-common-dir` with both paths RESOLVED before comparison, not compared as strings
9. a private per-agent worktree no longer gets the "expected / NOT your fault / NOT a finding" absolution, and WHERE TO WORK no longer calls it this session's own repository
10. `same_commit` is case-folded, so an uppercase `audited=` sha no longer disarms every degenerate-range guard
11. `FIX_MATRIX`'s mutants column is graded against the mutation harness's own `ROWS`, and every `RED_AT_BASE` entry must now appear in the matrix (ten did not)
12. the fix-matrix negative control counts its deliberately-wrong rows instead of asserting "Four" over five
13. the "measured inversions" ledger was RE-MEASURED here rather than restated from the audit that found it, and now names the scope it was measured at
```

ZacxDev:
## Round 5 fixes — pushed as `3619fe68`

```audit-claims round=5 audited=dd601793..3619fe68
1. `checkout-private` no longer grants writing over the assembly checkout — the "yours alone … Writing here is fine" grant is gone, and the state now says only that the path is where the brief was ASSEMBLED, with an unconditional "Write nothing to it".
2. the `no-fetch` clause is unconditional again, keyed on OWNERSHIP (any checkout that is not the copy YOU made) rather than on a shared/private state `gather_worktree_kind` cannot know for the auditor.
3. all three `checkout-*` states carry the no-write rule THEMSELVES, pinned two-way as a relationship over the state set plus a per-state behavioural render check.
4. the "movement is NOT expected here" inversion is removed — the assembling session is live in that tree, so movement is possible, and it is reported rather than absolved or denied.
5. `gather_worktree_kind`'s docstring and `test_a_private_worktree_is_not_described_as_shared_and_is_not_absolved`'s docstring no longer claim that tree is "where a dispatched auditor usually stands".
6. `render_checkout`'s kind line no longer says "tree is yours".
7. TOOLCHAIN's rationale is state-independent — "a checkout that is NOT yours … holding none of your mutations" instead of "the SHARED CHECKOUT" — and is covered in BOTH checkout states, including an equality pin across the two.
8. `--emit-claims` REFUSES (exit 4) to print a block that does not read back as written: whitespace, an embedded `..`, or an angle-bracket placeholder in `<from>`/`<to>`, for values from `--audited` and from a previous block's `<to>` alike; an empty `--audited` is refused outright.
9. the `degenerate-range-causes` clause scopes "`--audited` was omitted" to ROUND 1 and states that from round 2 on omitting it is correct.
10. the emitted skeleton carries a legend naming its two fields, outside the fence so the block it annotates still parses.
11. the fix matrix's containment count is computed from the ledgers instead of the hand-written "nine round-2 detectors and one round-3 detector", which counted ten over nine distinct members.
12. `range_anchor`'s docstring no longer reads as if `--audited` produces the bare `audited=<sha>` spelling.
```

### Evidence

**Red at `dd601793`** — `git show dd601793:scripts/audit-dispatch.py` into a scratch tree with the new test module copied in unchanged, `PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`: **10 failed, 79 passed**.

Four of the ten are regression coverage, and each observes the **wrong answer** before asserting any text the fix added:

| detector | what it observed at `dd601793` |
|---|---|
| `..._no_checkout_state_grants_write_permission_over_the_assembly_checkout` | `['Writing here is fine', 'yours alone', 'tree is yours']` in THE CHECKOUT, private state |
| `..._the_toolchain_reason_is_true_in_both_checkout_states` | TOOLCHAIN calls the private path "the SHARED CHECKOUT" |
| `..._emit_claims_refuses_an_audited_value_its_own_parser_cannot_read` | `--audited 'abc 123'` accepted, rc 0, header `audited=abc 123..<head>` |
| `..._the_degenerate_cause_list_scopes_the_omitted_flag_to_round_one` | the omission blamed with no round scoping |

Two new tests are filed as **GUARDS**, not regression coverage, because their red there is an *absence* the fix adds (two of the three checkout states carry no no-write sentence; there is no legend at all): `..._every_checkout_state_carries_the_no_write_rule` (mutants Z1, Z2, Z2b) and `..._the_emitted_skeleton_carries_a_legend_for_its_two_fields` (mutant Z5). The other four failures are pre-existing tests whose whole-string ledgers moved with the reword.

`..._a_private_worktree_is_not_described_as_shared_and_is_not_absolved` had its POSITIVES moved by this round, so it was **re-measured at `e06461f7`**: still red there on the false SHARED claim (its two negatives are untouched and run first), so it stays filed under that ref.

**Mutation matrix** — `scripts/tests/mutants-audit-dispatch.py`: **76 rows, all as expected**, over an 89-test positive control.

- `Z1` the private state grants writing again → killed by exactly 3 (whole-string directive ledger, the permission probe, the per-state no-write ledger). This is the required inversion of the corrected write rule.
- `Z2` the SHARED state drops its own no-write rule → 2. `Z2b` a fourth checkout state with no rule → 3.
- `Z3` `no-fetch` back to conditional-on-SHARED → 1 (the whole-string CLAUSE ledger alone).
- `Z4` TOOLCHAIN's reason names the SHARED CHECKOUT again → 1. `Z5` the legend is deleted → 1.
- `Z6` the emitted block is not round-tripped → 1. `Z7` an empty `--audited` is ignored again → 1.
- `Z8` the degenerate causes stop scoping the omission → 3.

Three pre-existing rows had to change and the reason is recorded on each:

- `S1` reported **MUTATION DID NOT APPLY** and was re-targeted — the fourth time assert-before-editing has caught a row that would otherwise have scored as "the guard held", and the second time on this same clause.
- `X3`'s killer set **grew** by one: its replacement text also drops the no-write sentence this round added, which is a real second hazard.
- `N3` and `Y7`'s sets grew by the new refusal test, and that is a measured **coupling in the code**: both mutants stop a `--audited` value reaching the emitted header, so a guard that correctly reads the *printed* header has nothing to refuse. The alternative — asserting the printed `<from>` equals `emit_from` — was measured and **rejected**: it made `N3` kill five tests and would answer a wrong-field bug in production with a misleading "it does not parse".

One **predicted killer did not fire**: `Z1` leaves `..._private_worktree_is_not_described_as_shared_and_is_not_absolved` green, correctly — that test covers the SHARED/PRIVATE *description*, and the write *grant* is a claim it never made. The prediction was written down and the harness refuted it.

`scripts/tests/mutants-audit-ladder.sh`: **18 rows, all as expected**.

**Gates, at the final head `3619fe68`:**

- module suite: `89 passed`
- dev-host tier, `scripts/gate.sh --tier both`: `GATE: RESULT=PASS exit=0` (pytest `RESULT: PASS (exit=0)`, node `RESULT: PASS (exit=0)`, 1292 node tests over floor 1239)
- sandbox tier (the one the merge gates on), status read directly rather than through a pipe: `nix build …#checks.x86_64-linux.pytests` → **rc 0**; `…nodetests` → **rc 0**. The nodetests build logged one `SQLite … is busy` that nix itself marked `error (ignored)` and built through — the known contention flake on this box, not a defect.

**End to end at the new head**, assembled from a real private per-agent worktree:

- THE CHECKOUT prints `kind : PRIVATE linked worktree — belongs to the session that assembled this brief, not to you`, followed by "**Write nothing to it**". Zero occurrences of `Writing here is fine`, `yours alone` or `SHARED CHECKOUT` anywhere in the brief.
- `--audited "abc 123"` now exits **4**, prints `🔴 REFUSING TO EMIT an audit-claims block … reads it back as <from>='', <to>='abc'`, shows the block that was **not** emitted, and no fenced `audit-claims` block reaches stdout.

**Not verified / named limits:** a well-formed token that is not a commit (`zzzzzzzz`) round-trips cleanly and is **not** caught — this script never resolves a sha in a checkout it refuses to write to, and the test asserts that blind spot rather than leaving it to read as covered. The corrected docstrings are prose and are not mechanically pinned; the mechanical guard is the per-state no-write relationship. Nothing here shows any auditor actually obeyed the corrected brief — that remains a re-measurement over future dispatches.

ZacxDev:
## Round 6's findings fixed — plus the pre-existing 🔴 it found under a probe

Head `3619fe68` → **`28492af2`**. Two commits.

**The pre-existing 🔴 is the third instance of this ladder's recurring confusion, and the first on a TIME axis.** Round 3 conflated *the operator's checkout* with *the tree under audit*; round 5 conflated *the assembling process's cwd* with *where the auditor stands*. This pair: **the HEAD verified at assembly time vs the HEAD the auditor's command will resolve.** `head_check.ok` is a fact about *this* tree at *this* moment, and the same brief tells the auditor to stand in a worktree they create themselves — cut after assembly, from a repository other sessions push to. `tip` is now `hc.local_sha`. Two tells that it was the same inversion: the **unverified** branch already named a sha, and `emit_claims_skeleton` already preferred the resolved PR head — only the *verified* branch hardcoded the token, the branch where being sure was mistaken for the range being stable. THE LEDGER's provenance line still says `..HEAD`, correctly: that command ran here, in the tree just verified.

**F2's suggested predicate was insufficient and is recorded as such.** `len(args.audited.split()) != 1` is **1** for `'aaaa1111\n'`, `'aaaa1111 '` and `'aaaa1111\r'` alike — `str.split()` with no argument discards empty fields, so the obvious spelling misses the exact character the finding is about. The check is `args.audited.split() != [args.audited]`, and the test asserts every fixture has `len(split()) == 1` so a future simplification to the obvious spelling fails rather than silently re-opening the hole.

### Gates, at the final head `28492af2`

| gate | tier | result |
|---|---|---|
| `scripts/gate.sh --tier both` | dev-host | `GATE: RESULT=PASS exit=0` (pytest + node both `RESULT: PASS`) |
| `nix build .#checks.x86_64-linux.pytests` | sandbox | rc 0; `TOTAL collected=17920 passed=17918 skipped=2 failed=0` |
| `nix build .#checks.x86_64-linux.nodetests` | sandbox | rc 0; `TOTAL suites=5 files=39 tests=1292 pass=1292 fail=0` |
| `mutants-audit-dispatch.py` | battery | **83 rows, all as expected** over a 94-test positive control (was 76 / 89) |
| `mutants-audit-ladder.sh` | battery | **18 rows, all as expected** |

nix's status was read directly (`rc=$?` on the build, not through a pipe), and both sandbox verdicts were confirmed from the `RESULT:`/`TOTAL` lines as well.

### Red at `3619fe68`, per fix

`12 failed, 82 passed` at the base with this module copied in unchanged. Four of those twelve are this round's regression coverage:

* `..._forward_reference_scopes_the_no_write_rule_to_a_denied_state` — red in the **`cross-repo-private`** scenario on the wrong ANSWER: *"the no-write rule below is about the SHARED checkout"* over a `kind : PRIVATE` line. The `cross-repo` scenario alone passes at the base, which is why the new scenario exists.
* `..._whitespace_leaves_the_line` — red on the CONTENT, not the rc: `--audited $'aaaa1111\n'` emitted a block reading back as `('', 'aaaa1111')` where `('aaaa1111', '11111111')` was meant, `malformed=[]`.
* `..._blind_spot_rationale_matches_what_the_script_actually_does` — red naming the command that refutes the old reason: `git -C <checkout> rev-list --count aaaa1111..HEAD`.
* `..._range_names_a_sha_and_not_head_when_the_head_is_verified` — red on `assert 'HEAD' != 'HEAD'`.

The fifth new test, `..._refuted_premise_survives_in_this_modules_prose`, scans **this module's own prose**, and the red-at-base procedure copies this module in *unchanged* — so no base SCRIPT can make it red. Filed as a guard; its red was measured separately against `3619fe68`'s test module (both ledger entries present once each, normalised). **Its first two drafts each reported the module CLEAN with the premise two lines away** — a raw `str.count` scored 0 (wrapped prose), and whitespace-normalisation alone still scored 0 for the second entry (a Python implicit concatenation cuts it in half). Both caught by measuring, not by reading.

**Every ledgered `RED_AT_BASE` claim was re-verified at its own ref this round — 0 false claims across `abc41024` / `d9eb36a8` / `e06461f7` / `dd601793` / `3619fe68`.** The per-round `N failed, M passed` counts in the module docstring are now labelled as claims about the module *of that round*: re-run with the current module they read 54/34/27/19, because each later round adds tests that are also red at the earlier trees.

### Mutant matrix — new and moved rows

| row | mutation | expected killers | result |
|---|---|---|---|
| **V1** | `checkout-private` renamed + rule dropped | 5 (both ledger pins, whole-string, emitted, no-write) | killed by exactly 5 |
| **V2** | a fourth **STATE** selects a rule-less directive | `..._every_checkout_state_carries_the_no_write_rule` | killed by exactly 1 |
| **V3** | the blind-spot note blames sha resolution again | `..._blind_spot_rationale_...` | killed by exactly 1 |
| **V4** | `own-worktree` scopes the rule to SHARED again | whole-string ledger + `..._forward_reference_...` | killed by exactly 2 |
| **V5** | the `--audited` whitespace input check removed | `..._whitespace_leaves_the_line` | killed by exactly 1 |
| **V6** | the VERIFIED range hands out `..HEAD` again | 7 (every consumer of the rendered tip) | killed by exactly 7 |
| **V7** | TOOLCHAIN names the SHARED CHECKOUT, **cross-repo only** | `..._toolchain_reason_is_true_in_every_scenario` | killed by exactly 1 |

V2 and V7 are the isolating rows for round 6's two 🟢 disarms: under the round-5 spellings each was invisible.

Three existing rows moved, and each is recorded rather than tidied away:

* **`Y6` / `Y14` did not APPLY** after `render_checkout`'s inline dict became module-level `CHECKOUT_STATE_DIRECTIVE` — they would have scored "the guard held". Re-targeted. That makes **four** rows the assert-before-editing check has caught after a refactor nobody thought touched the battery.
* **`H2` gained a third killer.** It hardcodes the token `HEAD`, and round 7's rule is that a verified head renders as its sha — the two claims are about the same rendered token from opposite sides, so a mutation that hardcodes it necessarily trips both. Recorded, not narrowed.
* **`Z7` SURVIVED.** Deleting the empty-`--audited` branch no longer changes the rc, because the new whitespace check subsumes the *outcome* (`"".split() != [""]`). It does not subsume the *explanation* — an empty value would be silently ignored, which has nothing to do with newlines leaving the header line, and telling an operator about newlines when they passed `""` sends them to the wrong fix. The test now demands the EMPTY branch's own diagnosis, and Z7 is killed again.

`..._range_names_a_sha_and_not_head_...` deliberately asserts **nothing about `<from>`**: doing so made mutant `N1` kill it for `..._range_is_generated_from_the_previous_rounds_audited_sha`'s reason (measured as EXTRA-KILLER). Its ledger-provenance control is asked of the anchor the brief actually rendered, for the same reason.

### End-to-end at `28492af2`, against the live PR

1. **Delta brief** — `Diff **`dd601793..28492af2a69aa29aec990dbf1d468851c879436b`**`, with the note *"so the range above names that SHA and not `HEAD`"*. No `..HEAD` in the instruction.
2. **`--audited $'aaaa1111\n'`** — **rc 4**, no block printed (`grep -c audit-claims` = 0), refusal names the value. Control: the same value without the newline emits `audited=3619fe68..28492af2` at rc 0.
3. **Cross-repo brief assembled in a PRIVATE agent worktree** (`--repo innovation-upstream/buildstream`, PR 253 — a real PR in a different repo, so the cross-repo branch renders) — the three sections now agree:
   * WHERE TO WORK: *"The no-write rule below is about **every checkout you did not make**."*
   * THE CHECKOUT: `kind : PRIVATE linked worktree — belongs to the session that assembled this brief, not to you`
   * NON-NEGOTIABLE: *"write to any checkout that is not the copy YOU made … including the one THE CHECKOUT section names"*

### Not verified

* The `..HEAD` fix is exercised against the **synthetic** cross-repo recipe and the live same-repo path; no live cross-repo *delta* round was assembled (no other repo has an in-flight audit ladder), so the interaction of the sha tip with a hand-rolled `git worktree add <branch>` is reasoned, not measured.
* F6 pins only the half that can be measured — that a rationale the script's own behaviour contradicts cannot be restated. The replacement reason (never fetching, so a `cat-file` refusal would be a false positive) is prose and is not machine-checked.

```audit-claims round=7 audited=3619fe68..28492af2
1. `own-worktree-is-writable` no longer scopes the no-write rule to the SHARED checkout; it scopes it by ownership ("every checkout you did not make"), so a cross-repo brief assembled in a PRIVATE worktree no longer contradicts itself across three consecutive sections. Payload constant and pinned ledger both moved.
2. `--audited` is refused on the INPUT when it is not exactly one whitespace-free token, which closes the TRAILING NEWLINE the exit-4 round trip is structurally blind to; the predicate is `split() != [value]`, not `len(split()) != 1`, because the latter reads 1 for a trailing newline.
3. This module's own prose no longer asserts the premise round 5 refuted; `REFUTED_PREMISES` pins each phrase to exactly one occurrence, normalised across wrapped prose and Python implicit string concatenation.
4. `..._toolchain_reason_is_true_in_every_scenario` drives every scenario `brief_for_scenario` knows — including the two cross-repo ones — and requires the TOOLCHAIN section byte-identical across all of them.
5. `..._every_checkout_state_carries_the_no_write_rule` derives its set from `render_checkout`'s own `CHECKOUT_STATE_DIRECTIVE` map instead of the id prefix `checkout-`, so a rename cannot empty it; the map is new module-level state in the script.
6. The blind-spot rationale in both files is replaced: the script does resolve that token in that checkout (two read-only commands), and the gap is kept because the script never fetches. A new guard measures the commands actually issued and forbids restating the false reason.
7. THE RANGE names `hc.local_sha` rather than the token `HEAD` whenever the head is verified, so the range does not re-resolve in the auditor's own worktree; THE LEDGER's provenance line still reports `..HEAD`, which is what it measured.
8. The empty-`--audited` refusal is now guarded on its own diagnosis rather than only its exit code, after the new whitespace check was measured to subsume the outcome (mutant Z7 SURVIVED until this changed).
9. Two new mutation rows (V2, V7) isolate the two disarms round 6 demonstrated green, and V1/V3-V6 cover the remaining fixes; Y6/Y14 were re-targeted after the state map moved, H2 gained a third killer, and every ledgered RED_AT_BASE claim was re-verified at its own ref.
```

ZacxDev:
## Round 8 fixes — `28492af2..706a6b38`

Six findings from round 7's delta re-audit, one disclosed item taken up, one
disclosed item taken up as a floor change, and a **fifth instance** of the
ladder's recurring shape found by sweeping for it.

**Verification.** Suite 102 passed. Mutation battery **93 rows, all as
expected**, over a 102-test positive control. Dev-host gate at `706a6b38`:
`GATE: RESULT=PASS exit=0` (pytest `RESULT: PASS (exit=0)`, node `RESULT: PASS
(exit=0)`). Sandbox tier at `706a6b38`: `nix build
.#checks.x86_64-linux.pytests` and `…nodetests` both built. Every
`RED_AT_BASE_*` claim re-verified across all six refs — 49 names, zero false
claims.

```audit-claims round=8 audited=28492af2..706a6b38
1. F1 (🟡, scaffolding) FIXED — both prose guards were blind to a phrase wrapped over two `#` comment lines and to single/mixed-quote implicit concatenation; measured at 28492af2 one plant per shape (wrapped comment MISSED, '…' '…' MISSED, "…" '…' MISSED). Replaced with ONE shared normaliser reading via `ast` (concatenation already performed by the compiler) + `tokenize` (COMMENT tokens, consecutive lines joined); both positive controls now plant one payload per shape into REAL Python source — round 7's fed the normaliser `f"x {phrase} y"`, which is not source. New battery rows V8 (wrapped comment) and V10 (single-quote concat) each kill the blind-spot guard alone.
2. F2 (🟡, payload) FIXED — THE RANGE and THE LEDGER now consult `facts.repo_relation`. Cross-repo THE RANGE states the impossibility as structural instead of diagnosing a moved checkout, and drops "Resolve the range in a tree that CONTAINS the PR's head" (there is none here); THE LEDGER says NOT MEASURABLE FROM HERE with the reason, hands over the whole command with `<your worktree>` and the PR's own repo named, and states that the payload-attribution gate is the auditor's to compute this round. Ordered AFTER the `hc.ok` branch deliberately: `repo_relation` compares slugs, and a clone with a renamed remote can still hold the commit, so the sha stays the ground truth. Rows V11/V12 each kill one of the two new tests alone.
3. F2 fixture FIXED — `OTHER_REPO_PR` now carries its own `headRefOid` and the assembly checkout a third distinct sha; `make_runner` derives the honest default via its OWN regex over the same two inputs, never `ad.pr_slug`. New scenario `cross-repo-delta` (the missing cell: both cross-repo scenarios ran at round 1, both delta scenarios were same-repo). Seam guard `test_no_cross_repo_brief_claims_its_checkout_was_verified` iterates every scenario; it is GREEN at 28492af2 (the defect was the fixture, and the procedure copies the module in unchanged) and filed as a GUARD with mutant V9 as its evidence.
4. F3 (🟢, payload) FIXED — the degenerate branch no longer overrides the tip; both branches read `range_tip`, which cannot return `HEAD`. The banner now also says both ends are shas so its absolute claim holds in the auditor's tree. Round 7's comment at :1417 corrected. V13 isolates the degenerate override; V6 was re-targeted onto `range_tip` and its killer set GREW to include the degenerate test — that growth is the consolidation, recorded on the row.
5. F4 (🟢, payload) FIXED — `no-fetch` dropped "for this audit", which had forbidden the brief's own cross-repo recipe (`git -C <your clone> worktree add …` writes to a clone the auditor made but did not make for this audit). Both sides now spell one phrase, `NO_WRITE_SCOPE`, and the new guard reads BOTH the clause and every forward reference, with a positive control per side. Round 7's comment claiming the two agreed is corrected. Five pre-existing rows (D2/W2/Y2/Z3/V4) gained the new guard — measured coupling, recorded once as `NO_WRITE_SCOPE_GUARD`.
6. F5 (🟢, payload) FIXED — THE LEDGER's provenance line names `local_sha` instead of `HEAD`. `measure_ledger` refuses unless the head check passed, so the two named the SAME COMMIT when the command ran: equally truthful provenance, and the only spelling that survives the re-run the brief invites. Round 7's control asserting the opposite was REMOVED from its host test rather than left contradicting the fix; the claim lives alone in a new test. V15 restores it.
7. F6 (🟢, scaffolding) FIXED — the re-derivability line printed 54, a FUNCTION count, in a row of NODE-ID counts. Re-derived with the round-8 module: 64/41/34/26/20/9 node ids at abc41024/d9eb36a8/e06461f7/dd601793/3619fe68/28492af2, and the unit is now named. The stale "89" positive control at :292 also corrected (it disagreed with 94 in the same docstring); today's is 102 and the harness prints it.
8. FIFTH INSTANCE (payload, NOT a round-7 finding) FIXED — `newest is not None` (a block PARSED) was read as "there is something to diff FROM". `audited=..` parses cleanly into two EMPTY halves, so a round-3 brief rendered ``Diff `None..<the PR's head>` `` — a literal Python None inside a git rev spec — at rc 0, under "verified at assembly time", with THE LEDGER giving the ROUND-1 cause for a round that has a previous round. Now REFUSAL 1b (rc 2) naming the two causes apart. Red at 28492af2 on a wrong answer; mutant V17.
9. DISCLOSED ITEM taken — the skipped-round warning. `--round 7` over a `round=2` block was silent while the opposite direction warned; the mirror now fires, counts the gap, and NEVER blocks (a skipped round number is legitimate, and `gh pr view --json comments` cannot see a review-comment block). Filed as a GUARD (absence at the base), mutant V16, with a negative control that the adjacent case stays silent.
10. DISCLOSED ITEM taken — `MIN_TESTS` 50 -> 97, derived as `m - min(50, max(1, m//20))` from the measured 102, not hand-picked. It was set at a 58-test suite and never moved; against 102 it would have passed a run that collected barely half the module.
11. HARNESS DEFECT found and fixed — `_swap` asserted its target was PRESENT but not UNIQUE. Round 8's own change created a second copy of C5's target at a deeper indentation (8 spaces is a substring of 12), so `replace(…, 1)` mutated a branch with no `led` in scope. `_swap` now refuses an ambiguous target; C5 was re-targeted onto the newly consolidated `payload_summary_line` (the FIFTH row assert-before-editing has caught after a refactor). H2 was RELABELLED, not silently re-recorded: the `range_tip` consolidation changed what its mutation MEANS.
12. RED-AT-BASE MATRIX, measured at 28492af2 with the round-8 module (9 failed, 93 passed): RED — the_ledgers_provenance_line_names_the_sha_it_resolved_not_head, the_degenerate_range_names_shas_at_both_ends, a_cross_repo_range_states_the_impossibility_not_a_moved_checkout, a_cross_repo_ledger_hands_the_attribution_gate_to_the_auditor, the_no_write_forward_reference_states_the_clauses_own_scope, a_block_that_parses_but_yields_no_anchor_is_refused. GREEN at base and filed as guards with named mutants — no_cross_repo_brief_claims_its_checkout_was_verified (V9). Red-on-an-ABSENCE and filed as a guard — a_claims_block_more_than_one_round_behind_is_warned_about (V16). All six RED names green at 706a6b38.
```

### What I could not verify

- The **cross-repo brief was never rendered against a real cross-repo PR** — every measurement is hermetic, through the injected runner. The structural claim it now makes (a checkout of repo A cannot stand on repo B's head) is argued, not observed live.
- `test_no_cross_repo_brief_claims_its_checkout_was_verified` **cannot be watched red at the base**: the defect it closes was in this module's fixture, and the red-at-base procedure copies the module in unchanged. Its only evidence is mutant V9, which reaches the same rendered state from the script side.
- The battery mutates the SCRIPT only, so `test_no_refuted_premise_survives_in_this_modules_prose` (round 7's) still has no mutation row — its evidence remains its own in-module positive control, which round 8 widened to one plant per shape.

ZacxDev:
Drive-by measurement from a session working the makework-audit queue — **evidence for this PR, not a request to change it.** The scope here (audit briefs only) is right; I'm recording the size of the surface it sits on, because the number that motivated the wider question was badly wrong.

## The `Agent` dispatch surface is ~7.7× what we thought

A ranked queue item claimed "~137 dispatches/day". Measured from `activity.events` `session-summary` → `tool_counts`, over the same 14-day window that figure came from (2026-08-11 → 08-25):

```
14,807 Agent dispatches = 1,058/day
```

Over 08-14 → 08-28: 12,251 = 875/day. Fleet-wide tool ranking for that window:

```
Bash    413,951
Edit     48,659
Write    24,638
Read     22,790
Agent    12,251     <- 5th
Skill     3,959
```

**60.3%** of 473 sessions dispatch at least one agent.

🔴 **Method note worth stealing:** my first query returned a clean `0`, because the `tool_counts` key is **`Agent`**, not `Task`. A zero from a wrong key is indistinguishable from a query wired to nothing. Enumerating *all* keys with totals is the positive control that caught it — worth doing before anyone quotes a dispatch number from this table.

⚠ Dispatches per session look like they're falling (33.7 → 19.3 → 9.6 across successive windows), but **that trend is not established**: `session-summary` rows are periodic, so still-running sessions undercount and the newest bucket is structurally incomplete. Session *rate* is flat at 31–36/day, so it isn't fewer sessions. Quote the level, not the trend.

## Why this PR mattered more than the queue item thought

That queue item was parked as un-actionable on the grounds that there was *"no committed code path, so no mechanical closing condition exists."* This PR is that code path. The decision taken today was explicitly: **let #958 land and prove itself on the audit path, then file the generalization to other dispatch types** — which only became fileable because of this.

So nothing is being asked of this PR. The follow-up is deliberately *not* filed yet, because "proved itself" has no definition until this merges and runs on some real audits.

## First-person evidence for the "retypable clauses" premise

Your motivation — *3 clauses lost across 14 hand-written dispatches* — reproduced in one session today, from the other direction. I hand-wrote **six** agent briefs. Briefs 4–6 carry warnings that briefs 1–3 do not:

- a zsh startup banner (`📬 N PRs`) writes to **stdout** inside `nix develop`, contaminating any captured command output;
- the dev shell has **no `pyyaml`**, so `kustomize-validate.sh` and `run-ci-suite.sh` both fail with messages that read like manifest/code defects.

Both are host facts, not task judgement. I only learned them mid-session — so the first three agents went out without them, and one of them did in fact chase a phantom red. An assembler would have back-filled both the moment they were known. (Those two are now filed as ZacxDev/homelab-infra#465.)

That's the strongest argument I can offer for the environment/contaminant clauses being assembled from code rather than retyped: **they change during a session**, and a human retyping them can only ever include what they knew at the time they typed.

ZacxDev:
Round 10 fixes are pushed (`706a6b38..88b4105c`). All seven round-9 findings addressed, plus a seventh instance found by the sweep. Nothing merged.

```audit-claims round=10 audited=706a6b38..88b4105c
1. 🟡 A — a delta brief whose PR has NO `headRefOid` no longer claims its tip is a sha. `range_tip_is_a_sha` + `unresolved_tip_note` are ONE predicate and ONE note with THREE consumers (THE RANGE's cross-repo branch, THE RANGE's could-not-verify branch, THE LEDGER's hand-over), and the sentence "So the range above names the PR's head SHA outright" is now conditional on the tip actually being one. Check `test_a_placeholder_tip_is_never_handed_over_as_though_it_were_a_sha` — driven over EVERY scenario, including the SAME-repo `delta-no-head-sha`, which the finding did not name and which printed the identical placeholder from a different branch.
2. 🟡 B — THE LEDGER's cross-repo branch is ordered after the head check, through the shared `cross_repo_holds_neither_end`. Scenario `cross-repo-renamed-remote-delta` (relation `cross`, head check PASSES) now prints the measurement it took — 13 added / 2 deleted over 4 commits — instead of "which holds neither end of the range". Check `test_a_cross_repo_ledger_prints_a_measurement_the_head_check_vouched_for`, which also asserts the OTHER side still renders, so the fix is not "delete the branch".
3. 🟡 C — every command a refusal prescribes now RUNS. The guard reads each `audit-dispatch.py …` command out of the refusal's own stderr, substitutes its `<…>` blanks and EXECUTES it, requiring an `audit-claims` block on stdout: `test_every_command_a_refusal_prescribes_actually_runs`, with `test_control_the_prescription_extractor_can_see_a_broken_remedy` as its negative control. REFUSAL 1's remedy was broken too — sound at round 2, unrunnable from round 3 up — and the same guard covers it. Mechanism: a refusal keeps its scope and its rc, renders NO brief, and continues to the emit half, because the two halves need different inputs.
4. 🟡 D — `own-worktree-is-writable` no longer grants `fetch`/`checkout` in the CLONE. It names `git worktree add` as the ONLY write the brief asks for there and refuses the other two by name, "whoever made it". Check `test_the_clone_grant_covers_only_the_write_the_recipe_makes`, which reads the verb off the recipe LINE so the grant and the recipe move together.
5. 🟢 E — the seam guard is rescoped and renamed to `test_no_brief_claims_a_verification_its_own_fixture_refutes`. It pins "verified at assembly time" against the FIXTURE'S OWN modelled `rev-parse HEAD`, in any scenario, cross-repo or not — through `fixture_resolved_head`, lifted out of `make_runner` so guard and fake cannot drift. Round 8's spelling contradicted `render_range`'s own comment and failed on CORRECT output the moment the renamed-remote scenario existed; that is measured, it was the first thing the new scenario did.
6. 🟢 F — re-derived by COUNTING, not from the old sentence: `MIN_TESTS` 97 -> 104 (m = 109) and `MIN_FIX_MATRIX_ROWS` 67 -> 74 (m = 77, from `len(FIX_MATRIX)`). Round 8's two wrong figures in that comment ("50 -> 95", "against today's 100", beside its own correct `m = 102`) are corrected and the error itself is recorded.
7. 🟢 G — `_require_unique` is now the ONE uniqueness rule with THREE callers. `drop_entry` and `reword_entry` count MATCHES via `_sub_unique` instead of testing `n != 1` under `count=1`, which could only ever see an ABSENT target and never a DUPLICATED one.
8. SEVENTH INSTANCE (swept, not filed): `base_ref` is `data.get("baseRefName") or "main"` — a DEFAULT — and every site printing it stated it as a fact about the PR's repository, at rc 0, inside THE LEDGER's mandatory hand-over ("…where `origin/main` names THAT repository's base branch"). `--claims-file` mode reaches it on EVERY run, because it consults no `gh` and hardcodes the field exactly as it hardcodes no `headRefOid` — and that second omission already carried a reason string while this one did not. `base_assumed` is now on `Facts` and `assumed_base_note` fires at every site. Check `test_no_brief_states_an_assumed_base_branch_as_a_fact`, driven BOTH ways over every scenario so the banner cannot be satisfied by firing always.
9. RED-AT-BASE at `706a6b38` (base script, this module copied in unchanged, `PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`): **7 failed, 102 passed**. FIVE are the new regression coverage and are listed in `RED_AT_BASE_R9`. Of the other two, one is the whole-string DIRECTIVE ledger moving with D's reword, and one is `test_the_tip_placeholder_ledger_matches_the_script`, red there on an `AttributeError` for a name the fix introduces — filed as a GUARD with mutant V21, never as coverage.
10. MUTATION BATTERY at this head: **100 rows, all as expected**, positive control 109 passed on the unmutated copy. Seven new rows V18-V24. FOUR pre-existing rows reported MUTATION DID NOT APPLY and were re-targeted (C6, K2 on a pure re-indent; V11, V12 onto the shared predicate) — that assert has now caught EIGHT rows across the ladder. TEN killer sets moved; C3 LOST seven of round 8's thirteen names, because THE LEDGER stopped turning on `repo_relation` alone. That loss is finding B, measured rather than argued.
11. `repo_relation` CONSUMER SWEEP: four read sites in the script, each checked — `render_where_to_work` twice (`unknown`, `cross`), `cross_repo_holds_neither_end`, and through it `render_range` and `render_ledger`. WHERE TO WORK deliberately still turns on the SLUG alone: `isolation: "worktree"` worktrees the CWD's repo, and a head commit being PRESENT does not make that the right repository to read a PR from. Recorded as a considered decision, not an oversight — the reasoning is in the commit message.
```

**Gate — both tiers at `88b4105c`, base `706a6b38`.**

| tier | command | runner's own verdict | exit |
|---|---|---|---|
| dev host | `scripts/gate.sh --tier both` | `pytest RESULT: PASS (exit=0)` · `node RESULT: PASS (exit=0)` · `GATE: RESULT=PASS` | 0 |
| sandbox | `nix build .#checks.x86_64-linux.pytests` | `RESULT: PASS (exit=0)` — collected=17935 passed=17933 skipped=2 failed=0 | 0 |
| sandbox | `nix build .#checks.x86_64-linux.nodetests` | `RESULT: PASS (exit=0)` — suites=5 files=39 tests=1292 pass=1292 | 0 |

Round 9's caveat is closed: nodetests was **force-rebuilt** with `nix build --rebuild` and re-executed the whole suite (`RESULT: PASS (exit=0)` again), and both sandbox derivations are new here anyway because the tree changed — neither verdict is a cache hit.

**One red run, and what discriminated it.** An earlier dev-host gate at this same head returned `GATE: RESULT=FAIL exit=1` on **one** test — `test_subsystem_store_api.py::TestLockoutOverHTTP::test_five_failures_lock_out_a_VALID_token_from_the_same_client`, which asserts `"status=lockout-triggered" in audit[4]` and got a line reading `status=locked-out`, i.e. one audit line off. Not attributed to this diff, on four measurements, not on plausibility:

- that file is untouched by this change and does not import `audit-dispatch.py`;
- the **sandbox** tier passed the same tree, including that file (0 failed);
- re-running the whole file on the dev host under the gate's own `-n 4 --dist loadfile`: **408 passed**;
- a second full dev-host gate, unloaded: `GATE: RESULT=PASS exit=0`.

Wall time rules OUT the load explanation rather than supporting it: `scripts/tests` took 333.25s in the red run against 327.81s in a green one — 1.7%, not the ~15× a loaded box produces. So it is an **order-dependent assertion in that test** (it indexes a fixed audit-line position), live on `main` and independent of this PR. **Not fixed here — out of scope.** Closing condition: a PR that makes that assertion search the audit lines for the record instead of indexing position 4, with the test watched red against a reordered fixture first.

**Not verified, stated plainly.** These claims are about the PR's BRANCH, not about the tree its merge into `main` creates — `strict` is false on this repo's branch protection and I did not build an integration branch and gate it. Nothing was deployed: `audit-dispatch.py` is not a `home.file` target, so no `home-manager switch` was involved and no claim here is about a deployed artifact.

ZacxDev:
Round 12 — the final round of the ladder. Four findings fixed plus round 11's three
one-liners; no new investigation, no ninth-instance sweep.

```audit-claims round=12 audited=88b4105c..a6a7aa4b
1. `--claims-file` mode no longer hardcodes `baseRefName`, so `base_assumed` is TRUE in the one mode three comments call permanently assumed — measured at 88b4105c: banner ABSENT at rc 0 with silent stderr; at a6a7aa4b: PRESENT. Scenario `claims-file-assumed-base` drives it, `scenario_base_is_assumed` reads ARGV before any payload, and the assumed side is counted BY MECHANISM so both doors must stay driven. Red at 88b4105c: test_no_brief_states_an_assumed_base_branch_as_a_fact (1 failed, 108 deselected). Battery row V25.
2. `assumed_base_note` no longer states "gh reported no baseRefName" in a run that never asked gh — the cause is now named per branch by `base_assumed_reason`, following `no_sha_reason` two lines away rather than a second mechanism, with a truthful fallback for a caller that cannot say.
3. `own-worktree-is-writable` refuses every write to the clone other than `git worktree add`, by a universal instead of the three-verb list that left `remote update`, `switch`, `restore`, `reset`, `branch -f` and `gc` permitted by omission — the only finding in this ladder whose blast radius is a cross-session write into `~/workspace/<repo>`. Red at 88b4105c: test_the_clone_grant_covers_only_the_write_the_recipe_makes (enumeration probe) and test_each_section_directive_carries_the_instruction_its_ledger_entry_names (whole-string ledger); 2 failed, 1 passed, 106 deselected. Battery row V28.
4. The C3 killer-set comment's numbers are re-derived by AST diff of that row's own name set, not restated: 706a6b38 = 19 distinct, 88b4105c = 13 distinct, 9 departed, 3 arrived. "SEVEN OF ROUND 8's THIRTEEN" was wrong both ways; all nine departures are now named individually.
5. `test_the_ledger_says_the_base_was_not_fetched` was killed by NO battery row (0 occurrences of its name in the harness at 88b4105c) and was walkable: replacing "**as it stands in this checkout**, and" with "and it is fine. It" left the base tree's own suite at 109 passed. The caveat is now pinned WHOLE and normalised; rows V26 (deleted) and V27 (reworded, both old fragments intact) each kill it, exactly 1 killer apiece.
6. `test_every_command_a_refusal_prescribes_actually_runs` now reads the prescription SITES out of the script by AST and asserts the driven cases reach all of them, attributed by line trace because the two sites render byte-identical text under these fixtures. Negative control measured: a third prescription planted in the `--audited`-without-`--emit-claims` branch fails on THIS assertion by its own message ("1 of the script's 3 prescription site(s) were never executed"), every other assertion staying green. Battery row V29.
7. `unresolved_tip_note` never omits `--repo`: an unknown PR repository now yields a `<...>` fill-in instead of a bare `gh pr view 900 --json headRefOid` that resolves against the auditor's cwd. Red at 88b4105c inside test_a_placeholder_tip_is_never_handed_over_as_though_it_were_a_sha, with a positive control that the lookup regex matched at least once.
8. Both `cross_repo_holds_neither_end` consumers state what the predicate measures ("not confirmed to hold either end") instead of the stronger "holds neither end" a mirror behind the PR refutes.
9. The seam guard's docstring no longer claims to be this module's fixture guard: measured, regressing `fixture_resolved_head` to round 8's behaviour gives 4 failed / 105 passed and this test is not among the four. The four that do catch it are named in the docstring.
```

**Gates.** Named by tier and base sha, because "the gate passed" is true of one run,
one tier, one base.

* **Mutation battery**, `a6a7aa4b`: `✅ 105 row(s), all as expected`, positive control
  `POS unmutated copy — 109 passed`, rc 0. Rows V25–V29 are this round's.
* **Sandbox tier** (the one Tekton runs and the merge is gated on), flake dir at
  `a6a7aa4b`: `nix build .#checks.x86_64-linux.pytests` exit 0, `RESULT: PASS (exit=0)`,
  `scripts/tests collected=9465 passed=9465`, `TOTAL collected=17935 passed=17933
  failed=0`; `nodetests` exit 0, `RESULT: PASS (exit=0)`, `tests=1292 pass=1292`. Both
  derivations were invalid beforehand, so both are real builds and not cache hits.
* **Dev-host tier** at `a6a7aa4b`: node `RESULT: PASS (exit=0)`. pytest came back
  `RESULT: FAIL (exit=1)` on **one** test —
  `test_subsystem_store_api.py::TestTrustedProxyOverHTTP::test_an_UNKNOWN_VERB_from_an_untrusted_peer_is_METERED_under_the_peer`,
  `IndexError: list index out of range` on `audit[0]`. **Not reproduced, and I am not
  calling it a flake on an absence.** What is established: that file and everything it
  exercises are byte-identical between `origin/main` and this head (this branch touches
  7 files, none reachable from it); the file's own `running()` docstring documents this
  exact race (`_respond` runs before `_audit`) and ships an `await_audit` helper whose
  docstring names this very `IndexError` as what it exists to prevent — used at 6 sites
  against 45 raw `audit[0]` reads, and the failing test is one of the 45. It did not
  recur in 12 runs on `origin/main` under CPU contention, in a re-run of the whole
  `scripts/tests` target at the same `-n 4 --dist loadfile` on this head (9465 passed),
  or in the sandbox tier. Wall time rules out a load flake: `scripts/tests` took 408s in
  the failing run against 334s clean, ~1.2x, with every other target normal.
* **Merged tree** — `origin/main` at `dec0a939` merged into `a6a7aa4b` in a separate
  detached worktree, so the PR branch stays linear. **All four legs green:** dev-host
  `GATE: RESULT=PASS exit=0` (pytest `RESULT: PASS (exit=0)`, `scripts/tests
  collected=9482 passed=9482`, `TOTAL collected=17969 passed=17967 failed=0`; node
  `tests=1300 pass=1300`), and the sandbox tier `pytests` exit 0 / `nodetests` exit 0,
  both `RESULT: PASS (exit=0)`, same 9482 and 17969 counts. A clean git merge is not a
  clean merge, so the one file touched by both sides —
  `claude/skills/audit-pr/SKILL.md` — was read as merged: the two additions are
  independent (this branch's `audit-dispatch.py` instructions in "What to do", main's
  load-generator reaping after the environment briefing), with no duplication or
  contradiction. No `TARGET_FLOORS` drift: floors are minimums and neither side edits
  the table.

Not merged, as the brief's `do-not-merge` clause requires.

PR #966, opened 2026-08-28
docs(browser-bridge): register vetr.com site notes (verified, with the drive lane walked back)

ZacxDev:
## What

Registers `reference/sites/vetr.com.md` + its `_index.json` entry. vetr had no
site-notes entry, so every result envelope on `app.`/`api.`/`admin.vetr.com`
emitted no `site_notes` and these flows were rediscovered per session.

## Why these two things are in it

- **vetr runs LIVE Authorize.net.** The Playwright suite refuses to target prod
  (`assertNotProductionTarget`, including the build-time `VITE_API_BASE_URL`);
  the bridge has no equivalent guard. Hence an explicit two-lane split: drive
  only the hermetic stack, read-only on `*.vetr.com`.
- **The timezone modal.** `root.tsx` opens a `fixed inset-0 z-[9999]` overlay on
  a profile/device timezone mismatch. Playwright dodges it with
  `timezoneId:"UTC"`; the bridge drives the operator's real browser and cannot.
  Fixtures are `UTC` (`E2ESeeder.php:71`), this host is `America/Winnipeg`.

## Nearly everything here is measured, with controls

(One exception, stated in the file: the toast-**present** branch of the discriminator
is not measured. See the results table in the file and correction comment 8.)

**The auth seed works end-to-end.** On a fresh tab, seed + second `nav` lands the
full authed dashboard (1232 chars `innerText`) with the complete data fan-out
(`/api/user/pets`, `/api/indicators`, `/api/vets/popular`,
`/api/user/subscriptions`, …). Negative control: seeding *without* the second
`nav` leaves the page logged-out — the `addInitScript`-vs-post-boot-`js` ordering
trap is real. Token confirmed independently (`E2E Owner`, id 1, tz `UTC`).

**The timezone modal, both controls.** Without the opt-out it fires: rect
`1124x1361` = exactly the viewport, `pointer-events: auto`, `z-index: 9999`, and
a hit-test of **all 35 in-viewport controls** returns the modal as topmost for
every one. With `timezone_mismatch_dismissed` seeded: no modal, dashboard
renders. The file steers to that opt-out and warns off `always_update_timezone`,
which `setTimezone()`s the fixture's profile **server-side**.

## ⚠️ Read the commit history — commit 2 was WRONG and commit 3 retracts it

`aa6fa270` reported that every non-root vetr route renders blank in a bridge tab
and declared the drive lane unusable. **That was a bad instrument, not a vetr
defect**, and `9006a8c5` retracts it. The blanks were a *reused* tab that had
stopped rendering; a fresh `open` at the same URL rendered everything.

Three instruments each gave a confident wrong answer, and the file now leads with
that because it is generic:
- `innerText` needs layout and returns `""` for an occluded tab that renders fine.
- A `button,a` count of 0 proves nothing — the chooser renders three role cards
  and counts 0.
- `elementFromPoint` returns `null` for an **off-viewport** point, identical to
  "covered by an overlay". vetr's off-canvas menu puts controls at `x = -249`.

The tired-tab behaviour is kept as a bridge-level trap with unknown root cause,
the rule "close and reopen **before** forming a theory", and the full ruled-out
list so nobody re-runs it (server: identical 5176-byte SPA fallback, `cmp` clean;
throttling: `activate`d, visible, animations advancing, still blank; seeded keys;
token; splash — renders the literal text "Vetr" and these reads had zero
`innerText`; GTM/Brave Shields — try/caught non-fatal, `root.tsx:288`; error
boundary — renders a visible "Oops!", `root.tsx:400`).

Standing rule added, because this session violated it: **nothing a bridge walk
finds is a vetr defect until the suite reproduces it.**

## Also recorded

- `js` is **MAIN** world (`service_worker.js:864-871`) — an earlier revision of this
  body said ISOLATED, which is wrong; see the correction comments. A returned
  Promise is not awaited, and cross-call `window` state was observed not to survive
  (unexplained for a top-frame call).
- Suite coverage is env-dependent, so both numbers are given: **~25 passed / 35
  skipped bare** vs **53 passed / 7 skipped / 0 failed hermetic** (10.7m,
  measured). The 7 skips are unattributed — the harness forces the `github`
  reporter, which does not name them.
- Laptop gotcha: `~/.config/vetr/authnet.env` holds PROD creds while the harness
  forces `ANET_ENDPOINT=sandbox`, so `e2e:mint-tokens` fails the card provision
  with `E00007` and leaves `owner_has_saved_card:false` — silently self-skipping
  the saved-card specs.
- Host routing verified: `vetr.com`/`app.`/`api.`/`admin.` resolve to the file;
  `notvetr.com`, `vetr.com.evil.invalid`, `127.0.0.1`, `localhost` do not. So the
  hermetic lane — the only lane allowed to click — never surfaces these notes,
  which the file says.

## Tests

`tests/test_site_notes.py` + `test_skill_size.py` → **75 passed**. Instrument validated, not just green:
dropping the registry entry turns `test_index_and_files_are_the_same_set` RED
(`assert not ['vetr.com.md']`); restored, green. `SKILL.md` untouched, so the
"SKILL.md never names an individual site" invariant is unaffected.

https://claude.ai/code/session_01NuDZJLkUhvu4yXJ6d8x6Rt

### DIFF
--- a/scripts/browser-bridge/reference/css-hit-test.md
+++ b/scripts/browser-bridge/reference/css-hit-test.md
@@ -68,6 +68,22 @@ $BB --instance work js '(function(){
 
 Any `insidePop:false` is the bug, and `hit` is the culprit.
 
+🔴 **`elementFromPoint` returns `null` for a point OUTSIDE the viewport — which is
+indistinguishable from "something is covering it".** Read `null` as *unknown*,
+never as *covered*: it is how a hit-test invents a blocking overlay that does not
+exist. Off-canvas drawers and mobile menus are the common source, and they sit at
+NEGATIVE coordinates (measured on one app: a nav control centred at `x = -249`).
+So **filter to in-viewport centres before hit-testing**, and say how many
+candidates you dropped:
+
+```js
+const cx = r.left + r.width/2, cy = r.top + r.height/2;
+if (cx < 0 || cy < 0 || cx > innerWidth || cy > innerHeight) return "off-viewport";
+```
+
+Corollary: a zero-size rect (`width` or `height` === 0) also has no meaningful
+centre — drop those first too.
+
 **3. Walk the ancestors for the first stacking-context creator.** The offender is
 almost never the element you're staring at. Check each ancestor's computed
 `transform`, `filter`, `opacity` (<1), `isolation`, `will-change`, `contain`, and
--- a/scripts/browser-bridge/reference/css-hit-test.md
+++ b/scripts/browser-bridge/reference/css-hit-test.md
@@ -54,35 +54,32 @@ the covering element instead of guessing — in the real case it returned the *n
 card's NSFW-reveal `<button>`, which reading the popover's own CSS would never
 have suggested.
 
+🔴 **`elementFromPoint` returns `null` for a point OUTSIDE the viewport — which is
+indistinguishable from "something is covering it".** Read `null` as *unknown*,
+never as *covered*: it is how a hit-test invents a blocking overlay that does not
+exist. Off-canvas drawers and mobile menus are the common source, and they sit at
+NEGATIVE coordinates (measured on one app: a nav control centred at `x = -249`). A
+zero-size rect has no meaningful centre either. So the probe **skips** both cases
+and labels them, rather than folding them in with the covered points:
+
 ```bash
 $BB --instance work js '(function(){
   const el = document.querySelector(".cm-updated-pop");
   const r  = el.getBoundingClientRect();
+  if (r.width <= 0 || r.height <= 0) return "zero-size rect — nothing to hit-test";
+  const inVp = (x,y) => x >= 0 && y >= 0 && x < innerWidth && y < innerHeight;
   const pts = [[r.left+8, r.top+8], [r.left+r.width/2, r.top+r.height/2], [r.right-8, r.bottom-8]];
   return JSON.stringify(pts.map(([x,y]) => {
+    if (!inVp(x,y)) return {x, y, skipped: "off-viewport"};
     const hit = document.elementFromPoint(x, y);
     return {x, y, hit: hit && hit.className, insidePop: !!hit && el.contains(hit)};
   }));
 })()'
 ```
 
-Any `insidePop:false` is the bug, and `hit` is the culprit.
-
-🔴 **`elementFromPoint` returns `null` for a point OUTSIDE the viewport — which is
-indistinguishable from "something is covering it".** Read `null` as *unknown*,
-never as *covered*: it is how a hit-test invents a blocking overlay that does not
-exist. Off-canvas drawers and mobile menus are the common source, and they sit at
-NEGATIVE coordinates (measured on one app: a nav control centred at `x = -249`).
-So **filter to in-viewport centres before hit-testing**, and say how many
-candidates you dropped:
-
-```js
-const cx = r.left + r.width/2, cy = r.top + r.height/2;
-if (cx < 0 || cy < 0 || cx > innerWidth || cy > innerHeight) return "off-viewport";
-```
-
-Corollary: a zero-size rect (`width` or `height` === 0) also has no meaningful
-centre — drop those first too.
+Any `insidePop:false` is the bug, and `hit` is the culprit. A `skipped` point is
+**not** evidence either way — if every point is skipped you have measured nothing,
+so say so instead of reporting "not covered".
 
 **3. Walk the ancestors for the first stacking-context creator.** The offender is
 almost never the element you're staring at. Check each ancestor's computed
--- a/scripts/browser-bridge/reference/sites/_index.json
+++ b/scripts/browser-bridge/reference/sites/_index.json
@@ -13,6 +13,7 @@
   ],
   "sites": {
     "civit.ai": "civit.ai.md",
-    "civitai.com": "civitai.com.md"
+    "civitai.com": "civitai.com.md",
+    "vetr.com": "vetr.com.md"
   }
 }
--- a/scripts/browser-bridge/reference/sites/vetr.com.md
+++ b/scripts/browser-bridge/reference/sites/vetr.com.md
@@ -69,8 +69,9 @@ environment into the API's `.env` **verbatim, with no test-mode assertion**. Pro
 runs `sk_live`/`pk_live`, and Lane A is the lane where you are told to click
 Book. So on the stripe rail, **export the test keys first** —
 `~/.config/vetr/stripe-test.env` (`pk_test`/`sk_test`/`whsec`) — and never run it
-with a live key in the shell. Keyless works for every non-payment flow (the API
-falls back to a placeholder secret).
+with a live key in the shell. Keyless works for every non-payment flow — the key
+is simply left **empty** (`vetr-api/config/services.php` has no default), and the
+payment/pool-loop specs self-skip.
 
 🔴 **You must mint your own token — the script does NOT give you one.** It
 captures `e2e:mint-tokens`, validates the JSON with `jq empty` and **discards
@@ -81,12 +82,20 @@ below has no input. Mint one yourself — there is no system PHP on NixOS, so
 borrow the one the script already built:
 
 ```bash
-PHP=$(ls /nix/store/*vetr-e2e-tools*/bin/php | head -1)
-(cd ~/workspace/scratch/vetr/vetr-api && "$PHP" artisan e2e:mint-tokens)   # → JSON
+# /tmp/vetr-e2e-tools is the script's own -o output link (e2e-hermetic-nixos.sh:126,139).
+# Use it, NOT a /nix/store glob: the glob's lexicographic winner is an arbitrary
+# PAST build (measured: 8.4.19 vs the live 8.4.24) and, unlike this link, is not a
+# GC root. Requires the KEEP_UP stack to be up — it needs the .env the script wrote.
+(cd ~/workspace/scratch/vetr/vetr-api && /tmp/vetr-e2e-tools/bin/php artisan e2e:mint-tokens)
 ```
 
-Re-minting is safe: `--seed` is opt-in (`vetr-api/app/Console/Commands/E2EMintTokens.php`),
-so a bare run issues fresh tokens without touching the seeded data.
+🔴 **Re-minting REVOKES the token already in your tab.** `E2EMintTokens.php:77`
+does `$user->tokens()->where('name','e2e')->delete()` for every fixture before
+issuing new ones — so the tab you seeded goes 401 and reads logged-OUT, which this
+file elsewhere teaches you to read as a seeding-order mistake. **Re-seed after
+every mint.** (`--seed` *is* opt-in, so a bare run leaves seeded rows alone — but
+it is not otherwise inert: it also touches Stripe/CIM to provision the saved card,
+`:112-138`.)
 
 ## 🔴 The auth seed — and why Playwright's version does NOT transfer
 
@@ -210,12 +219,36 @@ still stalled); **not the seeded keys**; **not the token**; **not GTM/Brave
 Shields** (`TagManager.initialize` is try/caught non-fatal, `root.tsx:288`);
 **not the error boundary** (it renders a visible "Oops!", `root.tsx:400`).
 
-⚠ **Do NOT reuse the earlier claim that the brand splash rules this out.** An
-earlier revision of this file argued "not the splash, it renders the literal text
-'Vetr'". That was **wrong and self-defeating**: the `~2s` dark "Vetr" splash was
-*replaced* by the textless skeleton above (`root.tsx:378-384` says so), and the
-served `index.html` contains "Vetr" only in its `<title>`. Stalled hydration is
-the live hypothesis, not an excluded one.
+### 🔴 vetr has TWO full-screen boot overlays. Tell them apart by TEXT.
+
+Conflating these produced two wrong findings in a row, in opposite directions.
+
+| | `HydrateFallback` | brand `Splash` |
+|---|---|---|
+| where | `vetr-app/app/root.tsx:377-398` | `app/components/splash.tsx:105-118`, mounted via `components/GeneralComponents.tsx:1,10` |
+| identify by | `[data-testid=app-boot-skeleton]` | `[data-testid=brand-splash]` |
+| text | **none** (`aria-hidden`) | renders **"Vetr"** as `<span>` per letter |
+| when | before the JS bundle executes | **after** hydration, first visit per tab (`sessionStorage.splashSeen`) |
+| holds for | until hydration | `minimumVisibleTime` **2000ms**, and on a non-landing route until `navigation.state === "idle" && isFetching === 0` (`splash.tsx:60-62`) |
+| blocks clicks | it *is* the page | no — `pointer-events-none`, but `z-[9999]` and opaque |
+
+🔴 **The brand `Splash` is LIVE — do not read anywhere that it was removed.** An
+earlier revision of this file claimed it "was *replaced* by" the skeleton, citing
+`root.tsx:378-384`. **That is false.** That comment says the pre-hydration paint
+*used to be blank, which read as* a ~2s dark "Vetr" splash — it is about the
+**blank**, not about `splash.tsx`, which is still imported and still rendered. The
+built `index.html` also contains "Vetr" **twice** (`<title>` and
+`apple-mobile-web-app-title`), neither of them rendered text.
+
+This matters operationally because **the remedy above is "open a fresh tab"** —
+and a fresh tab is exactly the case that gets the brand splash: an opaque
+full-viewport dark overlay for ≥2s that, on an authed route, additionally waits
+for **every in-flight query**. That is the state most easily mistaken for a stall.
+
+So, precisely: a read with **zero `innerText`** does rule the brand splash out —
+splash renders text. What it does **not** rule out is stalled hydration, because
+`HydrateFallback` is textless too. Two different textless-vs-texted states; check
+for both by `data-testid`, never by emptiness.
 
 🔴 **So: if a vetr route reads blank, `close` the tab and `open` a fresh one
 BEFORE forming any theory.** Why a reused tab reaches that state is a
@@ -230,10 +263,12 @@ up where it belongs, in `reference/css-hit-test.md`.
 
 - **`innerText` needs layout**; a backgrounded/occluded tab returns `""` for a
   page that is rendering fine. Use `textContent` for "is anything in the DOM",
-  and a **`screenshot` for what state it is in** — the screenshot is the only
-  thing that separated "pre-hydration skeleton" from "mounted but data-less".
-- **A button/link count of 0 means nothing.** The chooser renders three role
-  cards and still counted 0 `button,a`.
+  and a **`screenshot`** when you need to see which state it is in. (Cheaper and
+  deterministic for the boot states specifically: query the two `data-testid`s in
+  the table above.)
+- **A button/link count of 0 means nothing** — a pre-hydration skeleton has no
+  text *and* no controls, by construction. Measured: the chooser renders three
+  role cards and still counted 0 `button,a`.
 - **`elementFromPoint` returns `null` for an off-viewport point**, which reads
   identically to "covered by an overlay". vetr renders an off-canvas menu at
   negative x, so its controls sit at e.g. `x = -249`. **Filter to in-viewport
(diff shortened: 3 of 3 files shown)


ZacxDev:
## Correction to the PR body above — one of its ruled-out items was false

The body's ruled-out list (inherited verbatim from the retraction commit) includes:

> splash — renders the literal text "Vetr" and these reads had zero `innerText`

**That is wrong, and it was the audit's headline 🔴.** Correcting it here rather than
silently editing the body, since it may already have been read.

SSR is off, so the only thing painted before the JS bundle executes is
`HydrateFallback` (`vetr-app/app/root.tsx:377-398`) — a **textless**, `aria-hidden`
shimmer skeleton with `className="relative flex h-dvh flex-col"` and **exactly 3
children**. Its own comment (`:378-384`) records that it *replaced* the ~2s dark
"Vetr" splash. The served `index.html` contains "Vetr" only inside `<title>`, which
is not `innerText`.

So a read of zero `innerText` with `textContent` = only the react-router bootstrap
script is **precisely the signature of a tab stalled pre-hydration** — the very
mechanism that elimination told the next reader not to check. Worse, the original
session's "`DIV.relative flex h-dvh flex-col`, kids=3, so React mounted" observation
was that fallback, byte for byte. React had not mounted.

Fixed in `a34f155f`, which also:
- scopes **"no real money on Lane A"** to the authnet rail — the `ANET_ENDPOINT=sandbox`
  forcing is inside the authnet branch only, while `e2e-hermetic-nixos.sh:243-250` copies
  `STRIPE_SECRET_KEY` from the environment **verbatim with no test-mode assertion**, and
  the same sentence recommended that rail;
- removes the false **"it also prints the fixture ids you need"** (the script `jq`-validates
  the mint JSON and discards it, then exports into its own process, and is invoked rather
  than sourced) — the recipe dead-ended on its first line;
- stops hardcoding a timezone in the modal opt-out sample, which suppressed nothing for any
  operator outside `America/Winnipeg`;
- relocates the generic `elementFromPoint` off-viewport trap into `reference/css-hit-test.md`.

**One audit finding was rejected.** It certified "`js` runs in an ISOLATED world" as
VERIFIED, citing `browser:165` and `spa-wake.md:218` — but `:165` documents ISOLATED world
for `text`/`html`, and `spa-wake.md` states `js`/`eval` is **MAIN** world "by definition".
The underlying observation is real but unexplained, so it is now recorded as unresolved and
was **not** propagated into the mechanism file.

Delta re-audit against the previously-audited tip is next.

ZacxDev:
## Correcting my own correction — the brand splash is LIVE

My comment above says the "~2s dark 'Vetr' splash was **replaced** by the textless
skeleton" and that the built `index.html` contains "Vetr" only in its `<title>`.
**Both halves are false.** Round 2's delta audit caught it; fixed in `2e98480a`.

- `vetr-app/app/components/splash.tsx:105-118` is alive and rendering —
  `data-testid="brand-splash"`, `z-[9999]`, one `<span>` per letter of "Vetr" —
  imported and mounted at `components/GeneralComponents.tsx:1,10`.
- `root.tsx:378-384` says the pre-hydration paint *"used to be **blank**, which
  **read as** a ~2s dark 'Vetr' splash"*. That is about the **blank**, not about
  `splash.tsx`.
- The built `index.html` contains "Vetr" **twice** — `<title>` and
  `apple-mobile-web-app-title` — neither of them rendered text.

This is not a nit: the section's own remedy is *"open a fresh tab"*, and a fresh tab
is exactly what gets the brand splash — an opaque full-viewport overlay held ≥2s
that, on a non-landing route, also waits for **every in-flight query**
(`splash.tsx:60-62`). That is the state most easily mistaken for the stall it is
meant to help you diagnose.

**The underlying confusion, now fixed properly:** vetr has **two** full-screen boot
overlays and every previous revision treated them as one.

| | `HydrateFallback` | brand `Splash` |
|---|---|---|
| identify by | `[data-testid=app-boot-skeleton]` | `[data-testid=brand-splash]` |
| text | **none** (`aria-hidden`) | renders **"Vetr"** |
| when | before the bundle executes | **after** hydration, first visit per tab |

So the accurate statement, which neither round 1 nor my comment above got right:
zero `innerText` **does** rule out the brand splash (it renders text) but **does
not** rule out stalled hydration (`HydrateFallback` is textless). Check by
`data-testid`, never by emptiness.

### Also corrected in `2e98480a`

- The mint command I added picked the **wrong PHP** — the `/nix/store` glob's
  lexicographic winner is an arbitrary past build (measured **8.4.19** vs the live
  **8.4.24**) and, unlike `/tmp/vetr-e2e-tools`, is **not a GC root**. Now uses the
  script's own output link.
- **"Re-minting is safe" was over-broad**: `E2EMintTokens.php:77` deletes every
  `e2e` token before issuing new ones, so a re-mint 401s the tab you already seeded
  and it reads logged-OUT — which this file elsewhere teaches you to read as a
  seeding-order bug.
- `css-hit-test.md`: the off-viewport guard was prose beside a runnable snippet that
  violated it. Now wired in, with an all-skipped result labelled as *measuring
  nothing* rather than proving "not covered".

### One correction to the first audit, which stands

It reported "`js` runs in an ISOLATED world" as VERIFIED. It does not — `js`/`eval`
is **MAIN** world (`extension/service_worker.js:864-871`, and `:841-846` under
`--wake`). Round 1 rejected that finding and round 2 confirmed the rejection.
`--frame` eval *is* isolated, and gets a **fresh** world per call (`:539-542`),
which is the likely explanation for the cross-call state loss I saw. ⚠️ The PR body
above still contains the wrong "ISOLATED world" wording; treat this comment as
authoritative over it.

Round 3 delta audit next.

ZacxDev:
## Round 3 — correcting my previous comment's `--frame` claim, and two splash facts

Fixed in `825b50c6`. Two of the three are corrections to things **I** asserted in the
comment above.

### 1. The splash gate is keyed on LANDING, not on auth 🔴

The file's operative sentence said the splash *"on an **authed** route additionally
waits for every in-flight query"*. The predicate is the **entry path**:

```
splash.tsx:17   LANDING_PREFIXES = ["/user/vets", "/user/servicers", "/user/shop"]
splash.tsx:18   LANDING_FLOOR_MS = 300
splash.tsx:39   effectiveMinimum = landing ? min(minimumVisibleTime, 300) : 2000
splash.tsx:60-62 isQuiet = landing ? idle : idle && isFetching === 0
```

Both landing prefixes are **authed** routes, so an auth-keyed rule gets them exactly
backwards: `/user/vets/:id` is a **300 ms flash that does not wait for data**, while
`/auth/login` — logged *out* — is the 2 s data-blocked wait. Now an explicit
landing-vs-not table, with "landing" defined rather than used undefined.

Timing also completed: `debounceMs` 200 (`:78`) + a 500 ms fade (`:97`), so the real
non-landing floor is **~2.2 s**, and `[data-testid=brand-splash]` **stays in the DOM at
`opacity-0` for ~500 ms after it looks gone** — where its text still counts in
`innerText`. That is a live trap for the testid probe the table recommends.

### 2. The splash is not dark — it is bright orange→lime 🟡

`.bg-splash` = `from-[#ff5301] to-[#d1ff27]` (`app/app.css:361-363`). "Dark" came from
`root.tsx:380`, which describes the **old blank pre-hydration paint** — an attribute of
overlay #1's predecessor pinned to overlay #2. **The same conflation this section exists
to fix, in its mildest form**, and it defeats the sentence's only purpose: visual
identification. A reader seeing a bright orange screen would not have matched it.

### 3. My `--frame` explanation above was over-general, and wrong for the example I named 🟡

My previous comment said `--frame` eval "gets a fresh isolated world per call" and that
this was the likely explanation for the cross-call state loss, *because this file tells
you to use `--frame` for the hosted-CIM iframe*. There are **two** branches:

```
service_worker.js:537-544  same-process frame → Page.createIsolatedWorld (fresh per call)
service_worker.js:546-549  CROSS-ORIGIN OOPIF → that frame's flat session DEFAULT (MAIN) context
```

vetr's hosted-CIM iframe is served from `accept.authorize.net`
(`authnet-hosted-add-card.tsx:12-14`) — **cross-site, so it is the OOPIF branch, where
globals persist.** So `--frame` does *not* explain the loss for the one iframe I pointed
at. `reference/frames-cdp.md:45-46` already documented the split correctly and I
contradicted it.

The observation is therefore **unexplained again**, for the top-frame call it was
actually made on. Replacing an honest "unexplained" with a confident wrong mechanism was
the wrong trade, and I've reverted it to honest.

### Also

- Cross-referenced the two `z-[9999]` overlays: the splash is `pointer-events-none` and
  can never be returned by a hit-test, so a `z-9999` blocker is **always** the timezone
  modal. A reader hit-testing a blocker previously had nothing ruling the splash out.
- The PR body has been corrected: it no longer says "ISOLATED world" (it is MAIN), and
  the test count now reads 75, not the stale 71.

Round 4 delta audit next; the ladder stops at the first round that finds nothing.

ZacxDev:
## Round 4 — retracting the "always the timezone modal" line from my previous comment

Fixed in `8cf0de5e`. **Both findings are corrections to things I asserted in comment 3
above.** Fourth consecutive round in which the defect is an over-confident absolute
about these overlays, introduced by the previous round's fix.

### 1. `z-index: 9999` does NOT identify the timezone modal 🔴

My comment 3 said *"the splash is `pointer-events-none` … so a `z-9999` blocker is
**always** the timezone modal."* **The second half is false.**

`z-[9999]` + `pointer-events: auto` is the **shared** `FeedbackModalRenderer` backdrop
(`FeedbackModalRenderer.tsx:35`). The timezone modal (`root.tsx:334`) is **one of ten**
`<FeedbackModal>` callers — and the others include the exact surfaces this file routes
an operator to:

| caller | why it matters here |
|---|---|
| `routes/user/vets/vet-view.tsx:1764` | `/user/vets/:id` — the landing route in the table above |
| `routes/user/bookings/booking-checkout.tsx:88` | the checkout screens the gateway trap sends you into |
| `routes/user/appointments/appointment-checkout.tsx:143` | ″ |
| + `bookings/index`, `appointments/index`, `servicer-service-view`, `my-pets-create`, `pages/contact`, `professional-report`, `PetOwnerHomeBookingCard` | |

Plus a **separate** z-9999 blocker that isn't a `FeedbackModal` at all:
`components/UpdateApp.tsx:9` (`fixed inset-0 z-[9999]`, **no** `pointer-events-none`),
rendered from `root.tsx:321`.

The consequence is worse than a wrong fact: an operator whose click is swallowed reads
`z-index: 9999`, applies `timezone_mismatch_dismissed` — which **cannot** clear a
success/error `FeedbackModal` — and the file had told them the identification was
certain, so it offered no next step. Now: **read the modal's title, never its z-index.**

The splash row keeps only the inference it can actually support: `pointer-events-none`
⇒ a hit-test never returns the splash. That one-way direction was always fine; the
converse was not.

### 2. The landing floor was understated by 67% 🟡

`debounceMs` (200, `:29`) is added to **both** branches (`:70,78` —
`remaining + debounceMs`), so the real floors are **500 ms** landing and **~2200 ms**
non-landing. Comment 3 and the file did the `+200` arithmetic explicitly for
non-landing and then stated landing as a bare **300 ms** in the next paragraph — two
adjacent numbers derived by different rules. The table now separates the *constant*
from the *real floor* for both, and notes floors are minimums, not durations.

### 3. Fade description corrected 🟢

I'd written that the testid lingers "~500 ms **after it looks gone**". The 500 ms
window *is* the fade: `opacity-0` and the unmount timer are armed at the same moment
(`:94-100`), so the node is present for exactly the fade, not after it. The
operational point survives and is now stated accurately.

### Verified while here — no change needed

- `landingEntry` is captured **once at first mount** from the entry URL (`useRef`,
  `:36-38`), so navigating in or out of a landing route later does not change behaviour.
- Prefix matching is **exact-or-slash** (`:20-22`), so `/user/vetsomething` does not match.
- `<Splash />` is rendered with no props (`GeneralComponents.tsx:10`), so the `:29`
  defaults are what apply.

All three are now in the file, since each is a question a reader would otherwise have
to re-derive.

Round 5 delta audit next. The ladder ends at the first round that finds nothing.

ZacxDev:
## Round 5 — my z-9999 enumeration named a screen that cannot occur and omitted the one that can

Fixed in `54e515eb`. All three corrections are to things I wrote in comment 4 above.
Fifth round, fifth defect, again introduced by the previous round's fix — round 4
replaced an over-broad **absolute** with an **enumeration**, and presented the
enumeration as the answer set.

### 1. `UpdateApp.tsx:9` is a phantom in a browser 🔴

Every literal I wrote about it is true; the implication — that it is a blocker an
operator of this tool can meet — is false.

```
root.tsx:236   if (!Capacitor.isNativePlatform()) return;     ← web bails here
root.tsx:244       if (isProduction && isOutdated(version, requiredVersion)) {
root.tsx:245           setUpdateStoreUrl(storeUrl);           ← the ONLY call site
```

This file governs **web only** (`app.`/`admin.`/`api.vetr.com` and the hermetic
`127.0.0.1:5174`, all through Brave), so `updateStoreUrl` is permanently `null`.
And where it *does* fire it is an early `return` replacing the whole tree
(`root.tsx:320-322`), not an overlay — so it cannot produce the "page reads fine,
clicks do nothing" symptom the section exists to triage. Kept in the file, explicitly
labelled a phantom, because comment 4 offered it as *the* non-`FeedbackModal` example.

### 2. The z-9999 blocker that IS reachable was missing — react-toastify 🟡

```
root.tsx:161            <ToastContainer …>   ← mounted unconditionally, every route
ReactToastify.css:29    --toastify-z-index: 9999      (applied :52)
ReactToastify.css       grep -c 'pointer-events' → 0  ⇒ defaults to AUTO
axios.tsx:151,192,197   toast.error(…) on ANY API error
root.tsx:161-167        closeOnClick={false}  autoClose={5000}
```

So an API error parks a **clickable**, `position: fixed`, top-right, z-9999 box over
the page for 5 s. That is precisely this section's symptom, and the **likeliest**
non-timezone instance in a bridge walk — while the file listed an impossible blocker
instead. Bounded, and now stated: `--toastify-container-width: fit-content` (`:15`,
applied `:55`), so with no toast up the container is zero-size and intercepts nothing.

### 3. "Ten `<FeedbackModal>` callers" is **eleven** 🟡

And it contradicted the file's own list, which enumerates **ten others**. Counted
gitignore-blind (`find -print0 | xargs -0` with the real `grep`, not the `ugrep`
wrapper — not worth trusting for a number that goes into prose): **11** JSX call
sites = `root.tsx:334` plus the ten listed. Both "shared by ten" and "one of ten"
were wrong; now eleven / other-ten, internally consistent.

### Made the TITLE remedy executable per surface

- **`FeedbackModal`** — `title` is a *required* prop rendered as plain text
  (`FeedbackModal.tsx:8`, `FeedbackModalRenderer.tsx:60`), so `innerText` reads it.
- **A toast** has no title row at all → check for `.Toastify__toast-container`.

### Confirmed fixed, no change needed

The debounce arithmetic (the 500 ms / 2200 ms floors hold for **all** values of
`elapsed` — if `remaining` collapses to 0 the total is `elapsed + 200`, still above
the floor, so "floor" is the right word), the fade description, all three
"verified while here" facts, and the `--wake` citation split.

Round 6 delta audit next. The ladder ends at the first round that finds nothing.

ZacxDev:
## Round 6 — the toast mechanism doesn't fire either. Fix is subtractive this time.

Fixed in `1fa64826`. Sixth round, sixth defect, again in prose the previous round
added — and comment 5 above states both defective claims verbatim, so treat this as
authoritative over it.

### 🔴 "The axios interceptor fires `toast.error` on **any** API error" is false three ways

```
axios.tsx:151   in the SUCCESS handler, gated `config?.toastr && !response.data.success`
axios.tsx:197   gated `config?.toastr`
axios.tsx:192   HeicConversionError ONLY — raised in the REQUEST interceptor,
                client-side, never an API error (the comment at :182-189 says so)
```

`config` is `NativeStorage`'s cached `GET /config` payload — and **vetr-api never
emits a `toastr` key.** Verified with a validated instrument (gitignore-blind
`find -print0 | xargs -0` + the real `grep`, since the shell's `grep` wraps `ugrep`):

```
vetr-api/{app,config,routes}  'toastr'               → 0 matches
same method, positive control 'booking_pool_enabled' → AppController.php:63  ✓
```

So both gated toasts are dead against prod **and** the hermetic stack. Round 5 demoted
`UpdateApp` from "the non-`FeedbackModal` example" and filled that exact slot with
another mechanism that cannot fire — then ranked it "the LIKELIER one". An operator
with a swallowed click was being steered first at a toast that never appears.

### The fix DELETES rather than explains

Every round's defect has been in newly-added explanatory prose, never in the
citations. So this round removes claims instead of replacing them:

- **The section now leads with a runnable three-way discriminator**; the archaeology
  goes last. Previously the actual instruction ("read the TITLE") was the 46th line,
  after two mechanisms and a phantom.
- **Dropped:** "on any API error" · the LIKELIER/likeliest ranking · the `fit-content`
  zero-size mechanism (right conclusion, wrong reason — the container div simply isn't
  rendered when no toast is live, which makes `.Toastify__toast-container` a **binary**
  answer, so the doc now claims the stronger, simpler thing) · the "no `pointer-events`
  rule in the stylesheet ⇒ auto" inference (the library sets `pointerEvents` **inline**
  for the empty case, so a stylesheet grep was never the authority for a computed
  style) · the bare "for 5s".
- **Kept, because true and useful:** a toast implies a **prior click** (~68 `toast.*`
  calls across 14 non-test files, all component action handlers) · at ≤480px the
  container is `width: 100vw`, a **full-width top strip**, not a corner box
  (`ReactToastify.css:115-117`) — which matters because this same file tells you to
  `emulate` 390×844 · `pauseOnFocusLoss` (`root.tsx:168`) plus an unfocused bridge tab
  means treat it as *until dismissed*, not 5 s.
- `UpdateApp` cut from a paragraph to one line.

### One I caught on myself mid-fix

The new snippet initially used `[data-testid=feedback-modal]`. **That does not exist** —
`FeedbackModalRenderer` has no `data-testid` at all. Replaced with the computed-style
filter actually measured against this app (`position: fixed` + `zIndex === "9999"`),
and syntax-checked. Inventing a selector inside the fix for a round about invented
mechanisms is the whole pattern in miniature, which is why it's worth recording.

### Confirmed fixed, no change needed

`UpdateApp`'s phantom status and both its absolutes (defensible as written), the count
of eleven, the ten enumerated lines, and the TITLE remedy.

Round 7 next; the ladder ends at the first round that finds nothing.

ZacxDev:
## Round 7 — the discriminator missed the splash, and its "measured" validation was a different filter

Fixed in `ee2528c2`. The subtractive strategy worked on its own terms — every citation
is exact, the structure is better — but **the defect moved from explanatory prose into
the runnable procedure.** Same failure mode: an incomplete enumeration stated as an
answer set.

### 🔴 The lead snippet misidentifies the app's most common boot overlay

The filter is `position:fixed && zIndex === "9999"` over all `div`s. **The brand splash
matches it** — `splash.tsx:108` is `fixed … z-[9999]` — and its text is `"Vetr"`, so it
fell through to step 3: *"any other title ⇒ a different `<FeedbackModal>`; dismiss it on
its own terms."* That is precisely the false-overlay conclusion this section exists to
prevent.

And it fires in **the state the file itself mandates**: the remedy is *"close the tab and
open a fresh one"*; a fresh tab has no `sessionStorage.splashSeen`, so the splash is up
for ~2.2 s plus a 500 ms fade.

**Root cause, which is a class error rather than a typo.** Rounds 1–6 established *"the
splash is `pointer-events-none`, so a hit-test never returns it"* — true, and
load-bearing; comment 3 above states it. Round 6 then swapped the instrument from a
**hit-test** to a **computed-style DOM query**, where `pointer-events` is irrelevant, and
carried the exclusion across unchanged. **An instrument change invalidates the exclusions
derived under the old instrument.**

Fixed by adding a `splash` flag on the verified testid (`splash.tsx:106` — checked before
asserting, this time), a numbered step for it, and an instruction to read the named flags
*before* the `z9999` array, which is ambiguous alone.

### 🟡 The sentence validating that snippet was false two ways at once

It claimed the filter *"is the one measured against this app: it returned the timezone
modal plus its inner panel"*.

- The inner panel is `FeedbackModalRenderer.tsx:42`, `className="relative …"` →
  `position: relative`, `z-index: auto`. It can **never** match this filter.
- The earlier measurement used a **regex**, `/9999/.test(zIndex)` — a *different* filter.
  It also matches `#top-bar-bg`'s `z-index: 999999999999`, which is what actually produced
  the extra empty-text element I'd attributed to the panel.

So the snippet in this exact form **has never been run.** It now says so, and warns that
loosening to a regex brings the over-match back.

### Also

- **`innerText` → `textContent`** in the snippet. This same file documents, two sections
  down, that `innerText` returns `""` on an occluded bridge tab — so an all-empty `z9999`
  array would have fallen silently through to step 3.
- **Restored `closeOnClick={false}`** (`root.tsx:166`) — the one true, useful fact round
  6's subtraction lost. Clicking the toast body does not dismiss it, which matters in the
  one section of this file about clicks that do nothing.
- **Dropped the "~68 `toast.*` calls across 14 non-test files" count.** It was wrong and
  its halves were measured at different scopes (~68 is near the *with-tests* number, 71;
  14 was the non-test number; raw non-test is 54 lines / 14 files, fewer once comments are
  excluded). The conclusion it supported — a toast implies a prior click — stands without
  a number, so the number is gone rather than re-litigated.
- Added the caveat the `toastr` claim owed: `config` is the **cached** payload, so a stale
  cache or hand-seeded storage (which this file teaches) could still carry it.

Round 8 next. It will be asked to re-run the filter's expected result set against the DOM
rather than reason about it.

ZacxDev:
## Round 8 — and the ladder stops here

Fixed in `51b8d637`. Two **narrowings** — no new mechanism claims. Adding a
qualification to an absolute is the one edit shape that has not introduced a defect in
eight rounds, which is why this is the stopping point rather than a ninth round.

### 1. "It clears itself; wait it out" is unbounded 🟡

Step 2 of the discriminator said that of the brand splash. Contradicted by this file's
own timing table 60 lines up: only a **landing** entry clears on a time floor; any
other entry also waits for `isFetching === 0` (`splash.tsx:60-62`), so a hung query
holds it indefinitely — and the file already calls that *"the state most easily
mistaken for a stall"*. The reader reaches step 2 **because** something looks wrong, so
"wait it out" was the one instruction that might never terminate. It now says so and
redirects to the real question: which query is stuck.

### 2. The not-measured caveat blamed the wrong thing 🟢

It said no toast could be triggered *"(the interceptor's are gated off)"* — but this
same file says toasts are raised from component action handlers, and Lane A permits
clicking, so a toast **was** triggerable; only the interceptor route is closed. It
under-claimed coverage rather than over-claiming it, so it misled nobody, but the
reason didn't support the sentence.

### 🔴 Correcting comment 7 — and the reason is more interesting than the error

Comment 7 says the regex over-match came from *"`#top-bar-bg`'s `z-index:
999999999999`"*. **That is false**, and I want it corrected in the thread because this
thread is the durable record a future round reads to avoid re-deriving retracted
theories — and a reader who checks the source will find `#top-bar-bg` at `app.css:187`
with exactly that z-index and "confirm" it.

**Chromium clamps an out-of-range `z-index` to the 32-bit max.** Injected divs on a
live page:

```
"9999"          → "9999"
"999999"        → "999999"
"999999999999"  → "2147483647"     ← /9999/ does NOT match this
```

So `#top-bar-bg` never enters the regex filter; the extra element really is
`PullToRefresh.tsx:183` (`999999`), as `22d63382` said. The file already names
`PullToRefresh` and never mentions `#top-bar-bg`, so no payload change was needed.

The general point, and the one thing I'd carry out of eight rounds: **the source said
`999999999999` and the browser said `2147483647`. No amount of source-reading gets
there.** When an exclusion depends on what an *instrument computes* rather than what
the source says, run the instrument. Seven rounds of reasoning produced seven false
claims; the round that ran the query produced zero, and its one retraction of a prior
round was correct for a reason only measurement surfaces.

The PR body's "Everything here is measured, with controls" heading has been narrowed
to match the file's own caveat.

**Stopping the audit ladder here** — both remaining items were pure narrowings, and a
round to confirm a narrowing is the "run another round to confirm a clean round"
anti-pattern.

PR #974, opened 2026-08-28
feat(clawgate): host-side pusher for the cross-host tmux read model (tmux-webapp rank 4)

ZacxDev:
Completes rank 4 of `claudedocs/handoff-tmux-webapp.md`. The server half
(`ZacxDev/homelab-infra#468`) was merged but **not deployed**; it is now live as
**clawgate 0.8.8**, and this is the feeder that makes it do anything.

## Why a host agent at all
The clawgate pod has no `hostPath`, `hostNetwork`, `hostPID` or `nodeName`
(measured against the live spec), and tmux sockets are unix sockets on the
**hosts**. The pod structurally cannot see them, so the read model has to be
delivered outbound: no inbound access to either machine, no privileged pod, and
no SSH credential inside a pod on an unauthenticated LAN surface.

## Two decisions made against the design doc, on measurement
- **ONE unit, not one per host.** `session-manager --json` already collects both
  (it SSHes to the laptop). The timer is `serverMode`-gated, which is
  **correctness, not scoping**: two hosts each pushing a full two-host document
  would fight over every row.
- **Periodic HTTPS POST, not a WebSocket.** Rank 3 deferred the interactive
  terminal, so a persistent socket buys nothing today and adds reconnect/backoff
  state to get wrong.

## A dumb pipe, deliberately
The collector's stdout is POSTed **verbatim**; the server normalises, including
the `session` → `tmuxSessionName` rename that exists because the producer's word
for a tmux session is already clawgate's word for a Claude Code session. Keeping
that rule server-side keeps it in one tested place instead of in a shell script
on two machines. A test compares the delivered body **byte for byte** so any
reshaping reds.

## Verified live, not inferred
Posted a real `session-manager --json` to the live 0.8.8 server:
`{"ok":true,"hosts":["laptop","workbench"]}`, read back as 2 snapshots /
44 + 28 windows, rename applied, `capturedAt` taken from the producer's `ts`.

Cadence measured, not guessed: **1298 / 1196 / 1193 ms** per collection, so 2
minutes is ~1% duty cycle. (Timed with `date +%s%N` — this shell is zsh, which
has no `EPOCHREALTIME`, so the obvious timer reports 0 ms for everything.)

The unit was verified by **evaluating** it, not just parsing: `nix eval` renders
all ten PATH entries to real store paths and a timer wired into `timers.target`.

## Two deliberate absences, both pinned by tests
- **No `OnFailure=notify-failure@`.** That toast defeats do-not-disturb and is
  justified by a measured ~1 firing in 9 days; at this cadence a sustained
  outage would fire it 720×/day and burn down the one alert channel that must
  keep its meaning. Staleness already shows where it belongs — the server stamps
  `receivedAt`.
- **Environment beats the conf file.** `clawgate-stop-hook.sh` sources the same
  file with `set -a` so the file wins *there*, and the measured consequence was a
  probe aimed at a harmless address silently POSTing to production. Both
  directions pinned.

The token never reaches argv (`curl --config`, 0600, trap-removed), asserted via
a stub curl that records its own argv rather than by grepping the source.

## Gate
- Sandbox tier (what Tekton gates on): `RESULT: PASS (exit=0)`,
  collected=17862 passed=17860 failed=0.
- Sandbox nodetests: `RESULT: PASS`, 1300/1300.
- **Mutation sweep: 10 mutants, 10 killed**, unmutated control green, tree
  restored byte-identical.

## Two findings worth reading
🔴 **A mutant survived the first sweep.** Re-serialising the payload through
`json.dumps` was invisible, because the fixture *was* `json.dumps` output — a
round-trip of it is byte-identical. A fixture derived from the transformation it
is meant to detect cannot detect it. `RAW_DOC` is now hand-formatted, and the
test pins that property of the fixture so it cannot be tidied back into
blindness.

🔴 **The two tiers disagreed.** The dev-host tier was green while the sandbox
tier failed **11** of these tests: the sandbox has no `/usr/bin/env`, so stubs
could not exec. It did not present as a fixture fault — the stub failed, the
script correctly reported "collector failed", and the assertions that fired were
about rc and request counts, pointing at production code. Fixed by using
`testlib.mockbin.write_exec`, which this repo already consolidated for exactly
this reason; my first fix was a hand-rolled shebang, i.e. the sixth open-coding
of the rule, and `test_runtime_shebangs.py` correctly caught it.

## Deploy order
Server first — already satisfied (0.8.8 is live). A pusher meeting an older
server degrades to a 404 it reports as a deploy-order problem by name.

After merge: `scripts/ship.sh`, then confirm the timer fires and `receivedAt`
starts advancing. Merged is not deployed.

### DIFF
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -2933,10 +2933,23 @@ in
       # burn down the one alert channel that has to keep its meaning. That is the
       # "permanently-red gate trains you to click through" hazard, at 720/day.
       #
-      # Nothing is hidden by this: the server stamps `receivedAt` on every
-      # snapshot, so a feeder that stopped presents as a visibly stale timestamp
-      # in the read model — which is where a staleness signal belongs, next to the
-      # data it describes.
+      # 🔴 WHAT ACTUALLY SURFACES A DEAD FEEDER TODAY IS `/standup`, NOT the
+      # read model. An earlier version of this comment claimed the compensating
+      # control was the server's `receivedAt` stamp showing up as a stale
+      # timestamp. That is not observable by anyone: NOTHING reads
+      # `GET /api/tmux/snapshot` or `EventTmuxChanged` outside clawgate's own
+      # tests — no UI, no page, no script. The staleness is recorded and unread.
+      #
+      # The real control is that this unit is `Type = "oneshot"` with distinct
+      # non-zero exit codes, so a persistent failure lands in
+      # `systemctl --user --failed`, which `claude/skills/standup/standup.sh`
+      # reads. ⚠ That covers the codes (2/3/4/5/6) and NOT a unit that exits 0
+      # while achieving nothing — which is why the two audit findings in that
+      # class (a 3xx read as success, an unmeasured zero pushed as real) were
+      # fixed in the script rather than left to monitoring.
+      #
+      # A read-model staleness check belongs with the first consumer that
+      # renders this data; there is none yet.
     };
     Service = {
       Type = "oneshot";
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Correction to this PR's description, and what an adversarial audit changed

**The description above is wrong on one point and I am correcting it here rather than silently editing it, since it may already have been read.** It says "**Periodic HTTPS POST**". The transport is **cleartext HTTP** — `API_DEFAULT="http://192.168.50.250:30302"`, the LAN NodePort. The bearer token and the whole ~105 KB payload cross the LAN in the clear.

That payload is not innocuous: window rows carry `path`, `task`, `label`, `claude_session_id` and — measured on a live run — **`unsent_prompt`**, i.e. text typed into a window and never sent. This is a personal LAN and the route *is* authenticated (an unauthenticated `GET` returns 401, so `requireHookToken` is not failing open), so it is a judgement call rather than a defect — but the description should not have claimed HTTPS.

## Six audit findings, all fixed in `13bdc0a1`

No 🔴. Two were silent-success bugs, which matter disproportionately here: this runs every 2 minutes and **nothing currently reads the read model**, so a wrong-but-quiet push is invisible indefinitely.

1. **`rc=$?` was always 0, and was provably untested.** `if ! cmd; then rc=$?` reads the *negated* pipeline's status. Worst on the likeliest path — a timeout has empty stderr *and* status 124, so the log line read `collector failed (rc=0):`, carrying no information. The auditor mutated the capture to a literal `rc=0` and it **survived** the whole test file.
2. **Success was `HTTP < 400`, not 2xx.** No `-L`, so a 3xx meant curl returned the redirect, the server stored nothing, and the script logged "pushed" and exited 0. `000` and an empty status did the same (`[ "" -ge 400 ]` is false, and `set -e` does not fire inside an `if` condition).
3. **The silent-zero discriminant was discarded at the last place that still had it.** `session-manager` publishes `windows_measured` separately from `reachable` on purpose; clawgate's `hostReport` decodes only `reachable` and `windows`. With a latest-per-host upsert and no reaper, a reachable-but-unenumerated host would overwrite a good 44-window snapshot with a zero nothing downstream could identify as false. New rc 6. Deliberately **not** refused: `reachable:false` (a sleeping laptop would otherwise freeze the workbench's data every night) and a genuinely measured zero.
4. **`read_conf_key` diverged from shell sourcing while its comment claimed it matched** — `export KEY=…` returned empty, so the pusher would exit 2 every 2 minutes while `clawgate-stop-hook.sh` kept working off the same file.
5. **The stated compensating control for having no `OnFailure` was not real.** Nothing reads `GET /api/tmux/snapshot` outside clawgate's own tests, so a stale `receivedAt` is recorded and unread. The actual control is `Type=oneshot` + distinct exit codes landing in `systemctl --user --failed`, which `/standup` reads — which covers the codes but *not* an exit-0-achieving-nothing, hence fixing 1–3 in code.
6. **Cost understated ~4×** — up to four ssh invocations per run with no `ControlMaster` (~2,880/day, not 720), plus one ClickHouse query per run.

Also dropped `gawk`, `gnugrep` and `iproute2` from the unit PATH (all unused; the comment justified `iproute2` with a claim about interface addresses that `local_host_label` does not make), and amended the `TMUX_TMPDIR` note — the auditor could find **no declaration** of that variable anywhere, so the inherited value is undeclared runtime state rather than a property of the configuration.

**Mutation sweep for the new guards: 7 mutants, 7 killed**, including the auditor's own survivor. Verified end-to-end afterwards in a real transient systemd unit on the trimmed PATH: 105379 B, HTTP 200, both hosts.

### Filed rather than fixed
The laptop leg has no `TMUX_TMPDIR` guard — it works because that host's socket is in `/tmp`. If it ever acquires this host's arrangement it reports zero windows, and `tmux`'s "no server running" maps to `windows_measured: **true**`, so finding 3's gate cannot catch it either. Only an env pin on that side would.

PR #988, opened 2026-08-28
feat(browser-bridge): derive the extension version from the build marker

ZacxDev:
Every build now moves the version, so a stale profile is **visible at a glance in `brave://extensions`** — where the build marker, the actual staleness authority, cannot be seen at all.

## The gap

Measured on this host today: both Brave profiles reported version **`0.8.1`** while executing different code (`work` `b817ef1e`, `personal` `04bbd6f9`). #324 hit the same thing and concluded *no version-shaped signal can separate those rows* — true **while the version was hand-bumped**, and therefore constant across builds. Making it a function of the code removes that premise.

## Scheme

```
0.8.1.43738
└─┬─┘ └─┬─┘
  │     └── BUILD  — generated, = int(BUILD_MARKER[:4], 16)
  └──────── RELEASE — yours, bump by hand, needs a changelog block
```

**Four hex chars, not five, and that is a hard constraint.** Chrome caps each dotted component at 65535 and `0xFFFF` is exactly 65535. Five would emit manifests Chrome **refuses to load** for ~94% of markers — and an extension that fails to load presents exactly like a dead bridge. Pinned by a test, so widening the constant fails *here* rather than in Brave.

## The cycle, and why the marker no longer hashes the version value

The version is derived from the marker. If the marker also hashed the version then `marker = g(f(marker))` — writing the derived version back changes the marker, which derives a different version, forever. **There is no fixpoint to iterate to.**

`normalised_manifest_bytes()` replaces only the version *value* with a constant before hashing. Every other byte of `manifest.json` is still hashed, so a permissions change, a renamed service worker or an edited description all still move the marker. Convergence is **asserted directly** (`test_regeneration_converges`), not inferred from the design.

## What did NOT change

🔴 **`extension_stale` still compares the MARKER and nothing else.** The version is a human convenience and is **not fail-closed** — it describes the manifest that was *loaded*, so it can look current while the code is not. The asymmetry documented in `reference/errors.md` still holds. What improves: the `true`-from-version-mismatch branch now fires on **every build**, where before it only fired across hand-bumped releases — the exact case #324 measured and could not catch.

## A gate I almost broke

`test_manifest_version_matches_the_declared_build` requires the manifest version to appear as a changelog block in `extension/README.md`, so that *"a bump nobody can identify in the field"* fails. Auto-deriving per build would have demanded a changelog entry **per build** — unworkable, and the gate would have been deleted rather than satisfied.

It now compares the **release base only**. The README documents releases a human chose; the 4th component identifies builds a machine derived; each is gated by the test that can actually check it.

## Consumer sweep

One test unpacked the version three ways (`major, minor, _ =`) and raised `ValueError` on a 4-component string. I searched all 38 py/js/mjs files for version parsing **with a positive control** — the known splitter had to appear, and my first two greps returned a clean zero and were simply broken instruments.

**No production consumer parses the version**: `server.py` and the CLI pass it through as an opaque string. The test now slices `[:2]` — its claim is a lower bound on `major.minor`, and the three-way unpack silently encoded a format assumption it never meant to make.

## Verification

- **819 python** tests pass (809 + 10 new), **515 node** tests pass
- **Mutation — gate reachable with its OWN error:** hand-editing the version to `0.8.1.999` kills `test_committed_manifest_version_matches_the_marker` with its own message *while the marker gate still PASSES* — so the two are independent, not one riding on the other
- **Mutation — complementary:** an unregenerated one-line `service_worker.js` change turns **both** gates red; both green again after restore
- `--check` proven to go red on a stale version while the marker is clean, so the version half is not decorative
- **Negative control:** a manifest with no version **raises** rather than falling back to hashing raw bytes — the fallback would restore the cycle on precisely the malformed input
- **Idempotence:** two consecutive regenerations produce an identical marker and version

## 🔴 Not yet live-verified

`security-ops.md` makes live-verify-on-real-Brave the only real gate for a bridge change, and it runs **after** ship. Post-merge this needs `home-manager switch`, then **Remove + Load unpacked** in each profile — at which point both should show `0.8.1.43738` and `extension_stale: false`. Until then this is **tested, not verified**.

Note the marker moves `b817ef1e88267a40` → `aada672ff3a5ded7`, so both profiles go stale on deploy by design — that reload is what makes the new number appear.

https://claude.ai/code/session_011eexd2kXFxXfd44SXsg4Dm

### DIFF
--- a/scripts/browser-bridge/extension/README.md
+++ b/scripts/browser-bridge/extension/README.md
@@ -9,13 +9,52 @@ or merge them.
 
 | file | role |
 |------|------|
-| `manifest.json`     | MV3 manifest (permissions, icons, background SW, options page) |
+| `manifest.json`     | MV3 manifest (permissions, icons, background SW, options page). Its `version` is **partly generated** — see *Versioning* below |
 | `service_worker.js` | long-poll loop + chrome.* op executors (needs real Brave) |
 | `protocol.js`       | pure op-set / validation / envelope / backoff + registration payload (unit-tested) |
 | `build_id.js`       | **GENERATED** — the `BUILD_MARKER` literal that travels with the CODE (#324). Regenerate with `python3 scripts/browser-bridge/gen-build-marker.py`; CI fails if it is stale |
 | `options.html/js`   | one-time setup: bearer token + port + optional **label** → `chrome.storage.local` |
 | `icons/icon.svg`    | gruvbox bridge/link glyph — the SVG source |
 | `icons/icon-{16,32,48,128}.png` | rasterised icons wired into the manifest (regenerate with `rsvg-convert`, see `../README.md`) |
+
+## Versioning — `<release>.<build>`
+
+`manifest.json`'s version has two halves, owned by different parties:
+
+    0.8.1.43738
+    └─┬─┘ └─┬─┘
+      │     └── BUILD component — GENERATED, = first 4 hex chars of BUILD_MARKER
+      └──────── RELEASE base — yours, bump by hand, needs a changelog block below
+
+**Why the build half exists.** The build marker is the fail-closed staleness
+authority, but it is invisible in `brave://extensions`, where a human only ever
+sees a version. Two profiles once both read `0.8.1` while running different
+code: the version was hand-bumped, so it did not move between builds and could
+not separate them. The build component is derived from the marker, so it moves
+whenever the code moves — a stale profile is now legible **at a glance, without
+running anything**.
+
+    $ python3 scripts/browser-bridge/gen-build-marker.py         # regenerate both
+    $ python3 scripts/browser-bridge/gen-build-marker.py --check  # verify both
+
+Four hex chars, not five: Chrome caps each dotted component at 65535 and
+`0xFFFF` is exactly 65535. A wider component yields a manifest Chrome **refuses
+to load**, which presents exactly like a dead bridge.
+
+🔴 **`extension_stale` still compares the MARKER and nothing else.** The version
+is a human convenience and is not fail-closed — a version that merely *looks*
+current does not mean the code is. When the two disagree, believe the marker.
+
+🔴 **The marker deliberately does NOT hash the version value.** It hashes every
+other byte of the manifest, but the version is derived FROM the marker, so
+hashing it would make the derivation a recurrence with no fixpoint. Consequence
+worth knowing: a version-only edit does not move the marker, so hand-editing the
+version does not fake a new build — it just fails `--check`.
+
+Bumping the **release** base (`0.8.1` → `0.9.0`) still requires a
+`> **0.9.0 — …**` changelog block below, gated by
+`test_manifest_version_matches_the_declared_build`. The build component needs no
+block — that gate compares the base only.
 
 ## Per-session tab targeting (open/close + injected tabId)
 
--- a/scripts/browser-bridge/extension/README.md
+++ b/scripts/browser-bridge/extension/README.md
@@ -41,15 +41,27 @@ Four hex chars, not five: Chrome caps each dotted component at 65535 and
 `0xFFFF` is exactly 65535. A wider component yields a manifest Chrome **refuses
 to load**, which presents exactly like a dead bridge.
 
-🔴 **`extension_stale` still compares the MARKER and nothing else.** The version
-is a human convenience and is not fail-closed — a version that merely *looks*
-current does not mean the code is. When the two disagree, believe the marker.
+🔴 **An ALL-CLEAR (`extension_stale: false`) comes from the MARKER alone.** The
+verdict is asymmetric and this half is the one that matters: no version-shaped
+signal can ever produce `false`, because the version describes the manifest that
+was LOADED, not the code that is running. A `true` *can* additionally come from
+a version disagreement (`server.py:883-884`, `:887-891`) — so "it compares the
+marker and nothing else" would be wrong, and an earlier draft of this section
+said exactly that. When the two disagree, believe the marker.
 
 🔴 **The marker deliberately does NOT hash the version value.** It hashes every
 other byte of the manifest, but the version is derived FROM the marker, so
-hashing it would make the derivation a recurrence with no fixpoint. Consequence
-worth knowing: a version-only edit does not move the marker, so hand-editing the
-version does not fake a new build — it just fails `--check`.
+hashing it would make the derivation a recurrence with no fixpoint.
+
+Consequence worth knowing, stated precisely because the loose version of it
+misleads: a version-only edit does not move the marker, so hand-editing the
+version cannot fake a new build. But **`--check` only validates the BUILD
+component** — it derives its expectation from the very manifest it is checking,
+so it re-reads a forged release base as the base and passes. Editing
+`0.8.1.43738` to `9.9.9.43738` leaves `--check` printing OK on both lines. What
+catches a forged base is `test_manifest_version_matches_the_declared_build`
+under pytest, against the changelog below. **Do not treat `--check` as the
+anti-forgery gate before a local switch; it is blind to that.**
 
 Bumping the **release** base (`0.8.1` → `0.9.0`) still requires a
 `> **0.9.0 — …**` changelog block below, gated by
--- a/scripts/browser-bridge/extension/build_id.js
+++ b/scripts/browser-bridge/extension/build_id.js
@@ -24,4 +24,4 @@
 //   * a hash computed in the worker over its own source — same disk read
 // The marker must be a LITERAL in a module the worker imported, so that it was
 // frozen into the loaded module graph at load time and travels with the code.
-export const BUILD_MARKER = "b817ef1e88267a40";
+export const BUILD_MARKER = "aada672ff3a5ded7";
--- a/scripts/browser-bridge/extension/manifest.json
+++ b/scripts/browser-bridge/extension/manifest.json
@@ -1,7 +1,7 @@
 {
   "manifest_version": 3,
   "name": "Browser Bridge (command channel)",
-  "version": "0.8.1",
+  "version": "0.8.1.43738",
   "description": "Lets a local Claude Code skill drive the active Brave tab (getHtml/eval/tabs/nav/screenshot/frames/click/type/key/activate/upload/emulate/ping) via a loopback, token-authenticated rendezvous server. Local-only, authorized personal automation.",
   "permissions": ["scripting", "tabs", "activeTab", "storage", "alarms", "debugger", "webNavigation"],
   "host_permissions": ["http://127.0.0.1:8788/*", "<all_urls>"],
--- a/scripts/browser-bridge/gen-build-marker.py
+++ b/scripts/browser-bridge/gen-build-marker.py
@@ -173,17 +173,76 @@ def read_manifest_version(path=None):
 
 def version_base(version: str) -> str:
     """The human-owned prefix of a version: its first VERSION_BASE_COMPONENTS
-    components, or the whole thing if it is shorter.
+    components.
 
     So `0.8.1` -> `0.8.1` and `0.8.1.47127` -> `0.8.1`. This is what makes the
     scheme idempotent: re-deriving from an already-derived version must not
-    keep appending components."""
-    return ".".join(version.split(".")[:VERSION_BASE_COMPONENTS])
+    keep appending components.
+
+    🔴 STRICT, and the strictness is the point. An earlier version of this
+    function sliced `[:3]` off whatever it was given and returned it verbatim,
+    which is wrong in two directions that both reach a host:
+
+      * a base that is not Chrome-legal (`0.8.1-beta`, `v0.8.1`, `0.08.1`,
+        `0.8.1 `, ``) was concatenated with the build component and WRITTEN,
+        and `--check` then certified the result — producing a manifest Chrome
+        REFUSES TO LOAD, which is indistinguishable from a dead bridge. That is
+        the exact failure VERSION_HEX_CHARS exists to prevent, reached by
+        another door.
+      * a base with FEWER than VERSION_BASE_COMPONENTS components (`0.9`) has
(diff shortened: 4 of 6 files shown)


ZacxDev:
## Round 1 audit — 6 findings, all fixed in `6ec0c716`

Full adversarial audit, read-only, worktree-isolated. Findings, and what was done:

| # | severity | label | finding | fix |
|---|---|---|---|---|
| 1+2 | 🟡 | payload | `derive_version` emitted a version it never validated; `--check` then certified a manifest Chrome refuses to load. A base of `0.8.1-beta` / `v0.8.1` / `0.08.1` / `"0.8.1 "` was concatenated and **written**, and 4 of the 5 rows passed `--check` with rc=0. Separately a <3-component base (`0.9`) has no fixpoint — it converges only after freezing one build's id permanently into the release. | `version_base` now requires exactly 3 legal components; `derive_version` validates the **concatenation** |
| 3 | 🟡 | payload | `reference/errors.md` claimed the `true`-from-version-mismatch branch "now fires on every build". **False** — `server.py:887` runs only `if not stale`, so a real build (markers disagree) never reaches it. Reach is unchanged. | corrected; the real win is stated as off-bridge legibility, a *reporting* not a detection improvement |
| 4 | 🟡 | payload | `README.md` claimed hand-editing the version "just fails `--check`". False for the base — `--check` derives its expectation from the manifest under test, so `9.9.9.43738` passes clean on both lines. | corrected, with a warning not to treat `--check` as the anti-forgery gate |
| 5 | 🟡 | payload | `README.md` claimed `extension_stale` "compares the MARKER and nothing else" — contradicted by `server.py:883-884`, which sets `true` from a version disagreement with no marker. A reader could conclude those branches are dead and delete them. | replaced with the accurate asymmetric claim: only the **all-clear** comes from the marker alone |
| 6 | 🟢 | scaffolding | `PINNED_EXT_VERSION` re-implemented the derivation and raised `TypeError` at **import** on an unreadable `build_id.js`, hiding all 444 tests in the file behind a collection error | calls `GEN.build_component`, which validates its input |

**Regression matrix, measured:** 9 of the 10 new tests are **RED** against the pre-fix tree `ea230cc5` (generator reverted in a `.git`-stripped `cp -a` copy, new tests kept). The 10th passes pre-fix — the shipped manifest's base is already legal — so it is labelled an **invariant guard** in its own docstring and is not counted as regression coverage.

**Gates:** 829 python (819 + 10 new), 515 node, `--check` clean. Dev-host tier.

**Scope, honestly:** the auditor's sandbox-tier `nix build` had not returned when it reported, so its gate claim covers the **dev-host tier only**. And nothing here is live-verified against real Brave — `security-ops.md` makes that the only real gate for a bridge change, and it runs after ship.

Claims checked and found **correct**, recorded so a later round does not re-derive them: the regex cannot match `manifest_version` / `version_name` / `minimum_chrome_version`; the marker still moves on every `*.js` and every non-version manifest field; the normalised-**length** subtlety is real (a raw-length variant oscillates on 2 of 40 source variants); 5 hex chars would be 93.75% invalid; no production consumer parses the version.

```audit-claims round=1 audited=ea230cc5..6ec0c716
1. derive_version and version_base now reject a release base that is not exactly 3 Chrome-legal components, and validate the full concatenated version before it is written.
2. build_component validates its input and raises its own error instead of TypeError-ing on a None marker from read_marker.
3. reference/errors.md no longer claims the version-mismatch branch widened; it states the reach is unchanged and the gain is off-bridge legibility.
4. extension/README.md no longer claims --check catches a hand-edited version; it states --check validates only the 4th component and names the pytest gate that catches a forged base.
5. extension/README.md no longer claims extension_stale compares the marker alone; it states the asymmetric rule that only the all-clear comes from the marker.
6. PINNED_EXT_VERSION calls GEN.build_component instead of re-slicing the marker, removing a duplicated derivation and an import-time crash.
```

ZacxDev:
## Round 2 delta audit — 6 findings, all fixed in `ea38ed13`

Delta round over `ea230cc5..6ec0c716`, dispatched **blind to round 1's reasoning** (claims only, never why they were correct). Per-claim verdicts came back: 2 actually fixed, 4 partially — and the partials were the interesting ones.

Two findings are the class this repo treats as worse than no comment: **a comment claiming protection the code does not provide.**

| # | sev | label | finding | fix |
|---|---|---|---|---|
| 1 | 🟡 | payload | The concatenation guard in `derive_version` was **dead code** — `version_base` returns 3 validated components, `build_component` a bounded int, so it could not fire for any input (confirmed across 6 manifest shapes). It carried a comment calling it *"the last point before it is written to a manifest"*. Worse, it pointed **away** from the entry point that genuinely is open: `write_manifest_version(d, "0.8.1-beta")` wrote an unloadable manifest, unchecked. | guard moved to `write_manifest_version`, where a caller-supplied `version=` can actually be illegal |
| 2 | 🟡 | scaffolding | The `PINNED_EXT_VERSION` comment claimed the import-time crash was removed. **It is not** — deleting `build_id.js` still kills collection at the same module-level line, now `AssertionError` instead of `TypeError`, hiding all 454 tests. Better message, identical blast radius. | comment now says exactly that, and why the fixture rewrite that *would* fix it isn't done here |
| 3 | 🟢 | payload | `errors.md` named the wrong branch — the "degraded config" reaches `:883-884`, not `:887`, because `:879` short-circuits on a missing marker | replaced with the state that genuinely reaches `:887`, which is **routine**: a release bump changes only the version, which is normalised *out* of the digest, so the marker doesn't move and a pre-bump worker gives matching markers with disagreeing versions |
| 4 | 🟢 | scaffolding | The retracted *"compares the MARKER and nothing else"* claim survived verbatim in a **third** place. One rule, three copies, two corrected. | swept the tree; both remaining copies state the asymmetric rule |
| 5 | 🟢 | payload | `_require_legal_component` leaked `ValueError` on Unicode digits — `'²'.isdigit()` is True while `int('²')` raises, so the intended message naming Chrome's rule never printed | guarded with `isascii()` |

Also fixed `version_base`'s own error text, which said the base "must be exactly 3" while ≥3 is accepted and truncated — the 4-component steady state **is** the normal input.

**Regression matrix:** the 6 new `write_manifest_version` cases are **RED at `6ec0c716`** (generator reverted in a `.git`-stripped copy, caches cleared, `PYTHONDONTWRITEBYTECODE=1`), green at HEAD. That red is precisely what distinguishes the relocated guard from the dead one it replaces.

**Gates:** 835 python (829 + 6), 515 node, `--check` clean. Tekton was green on `6ec0c716`; it will re-run on this head.

Round 2 also **independently reproduced round 1's matrix** (9 failed / 1 passed at `ea230cc5`) and confirmed the 10th test's *"invariant guard, not regression coverage"* label is accurate. It found no input that passes the component validator but Chrome would reject — it errs strict, the safe direction.

```audit-claims round=2 audited=6ec0c716..ea38ed13
1. The unreachable concatenation check was removed from derive_version and an equivalent check added to write_manifest_version, where a caller-supplied version= can be illegal.
2. The PINNED_EXT_VERSION comment no longer claims the import-time collection crash was removed; it states the blast radius is unchanged and only the message improved.
3. reference/errors.md no longer names the degraded-config state as reaching server.py:887; it names a release bump, on the grounds that a version-only change does not move the marker.
4. The retracted "compares the MARKER and nothing else" sentence was removed from its third copy in tests/test_server.py.
5. _require_legal_component guards isdigit() with isascii() so a Unicode digit raises the intended AssertionError instead of ValueError.
6. version_base's error message no longer says the base must be "exactly" 3 components, since 4 is the normal steady-state input and is truncated.
```

ZacxDev:
## Round 3 delta audit — all 6 claims verified fixed; 1 finding, fixed in `4019e07e`

Round 3 came back with **all six of round 2's claims "actually fixed"**, each verified by execution rather than reading. Notably it confirmed the `errors.md` release-bump mechanism — the third attempt at that paragraph, and the one I explicitly briefed it to distrust — by *running* `annotate_staleness` with two discriminating controls:

- bump `0.8.1.43738` → `0.9.0.43738` by hand ⇒ `compute_marker` returns the **identical** `aada672ff3a5ded7`, so the marker does not move;
- feed that state in ⇒ `stale: true` with both markers present and agreeing, so `:879` did not fire and `:886` left it False — **`:887` is the only line that can have set it**;
- controls: marker moved ⇒ true from `:886`; `expected_build=None` ⇒ true from `:883-884` (the branch the earlier draft wrongly named).

It also independently re-verified both earlier red-at-pre-fix matrices, and **mutation-swept the relocated guard** — three mutants each killed a distinct param, plus a comment-only positive control confirming the harness can go green. No param redundant, no clause unreachable.

### The one finding — pre-existing, fixed anyway

`_require_legal_component` had a **second** `ValueError` route out of `int(part)`. Round 2 closed the Unicode one (`'²'`) and asserted the class shut. It was not: CPython refuses `int()` past **4300 digits**, so `'9' * 5000` is `isascii()` *and* `isdigit()` and went straight through to `str(int(part))` → `ValueError: Exceeds the limit (4300 digits)`, losing the message naming Chrome's rule and the file to edit.

The auditor measured it at **both** `6ec0c716` and `ea38ed13`, so it is **not a regression this round introduced** — round 2 only narrowed the class. Fixed regardless, because my docstring claimed the class was closed, and a comment wider than its code is the defect class this repo cares most about.

Closed with a length clause before `int()`, keyed to a derived constant rather than a literal: `MAX_COMPONENT_DIGITS = len(str(VERSION_COMPONENT_MAX))` = 5. It cannot reject anything valid — no legal component has more digits than the largest legal component. Verified by execution: 4301 nines, 5000 nines, 6 digits and `'²'` all raise the intended `AssertionError`; `0`, `1`, `8` and the **boundary `65535`** still pass. Both routes are now pinned as params, and the docstring names which clause closes which.

**836 python passed**, `--check` clean.

### Ledger correction

The auditor could not reproduce a carried-forward cumulative of 344 and computed **363** through round 2 (round 1 payload 225 + round 2 payload 138), re-deriving my three per-file rows exactly. I am carrying **its** number, not mine — it verified the arithmetic and I did not.

```
round 3 · payload lines changed THIS round: 74 (since round 1: 363)
```

Round 4 is dispatched. Round 3 produced a finding that required a fix, so the stop rule keeps the ladder open — and the fix added a boundary-sensitive length check, which is exactly where an off-by-one hides.

```audit-claims round=3 audited=ea38ed13..4019e07e
1. _require_legal_component rejects components longer than MAX_COMPONENT_DIGITS before calling int(), closing the >4300-digit ValueError route.
2. MAX_COMPONENT_DIGITS is derived from VERSION_COMPONENT_MAX rather than written as a literal, so it cannot drift from the cap.
3. The length clause rejects nothing valid, on the grounds that no legal component has more digits than the largest legal component.
4. Both ValueError routes out of int(part) are pinned as params of test_write_manifest_version_refuses_an_illegal_CALLER_SUPPLIED_version, and the docstring names which clause closes which.
```

ZacxDev:
## Round 4 — CLEAN. The ladder ends here.

All four of round 3's claims verified **actually fixed**, no findings. Per the stop rule, a clean round ends the ladder and I will not run another to confirm it.

**The verification behind that verdict**, because "clean" is only worth what was actually exercised:

- **Exhaustive**, not sampled: all 65,536 legal components `0..65535` accepted (0 wrongly rejected); `cap+1 .. cap+100000` all rejected (0 escapes).
- **Every boundary case attributed to the clause that fired** — `65536` and `99999` are rejected by the **cap** clause, not the length one; `100000` by length; `00000` by the canonical/leading-zero clause; `²`/`''`/`+1`/`1e5`/`٣`/`½`/`①`/`１` by ascii-digit. A value rejected by the *wrong* clause is a latent bug even when the verdict is right, and none was.
- **`int(part)` proven unreachable-with-raise** by brute-forcing all 111,110 strings that survive the earlier clauses — zero raise. Holds even if an embedder lowers `int_max_str_digits`, whose floor (640) is far above 5.
- **Derivation tracks**: patched `VERSION_COMPONENT_MAX` to 999999 / 255 / 4294967295 / 9 / 10 in throwaway copies — the derived digit count matched in all five, nothing legal rejected, `cap+1` rejected in every one.
- **Mutation sweep** with a comment-only positive control (836 passed, so the harness can go green): deleting the length clause is **KILLED** by a bare `ValueError` escaping at the guard's own line — the right error, not a substituted assertion. Deleting the clause **and** the 5000-digit param leaves **835 passing across the whole directory**, so that param is the *unique* killer in the suite. Load-bearing, not decorative.
- **Docstring exactly as wide as the code** — the first round of the four where that was true. Rounds 1, 2 and 3 each found a claim in this file wider than its implementation.

### Ledger

The auditor **verified rather than copied** the cumulative, recomputing each round independently: r1 = 225, r2 = 138, r3 = 74 → 437 ✓, + 29 this round.

```
round 4 · payload lines changed THIS round: 29 (since round 1: 466)
```

⚠ It could not reconcile the brief script's printed `since round 1: 381` with any classification it could construct. Carrying **466**, the number that was re-derived from `--remerge-diff --not origin/main` per round. Worth a look at `audit-dispatch.py`'s cumulative arithmetic separately — it is not a defect in this PR.

### Ladder summary

| round | findings | of which introduced by the previous fix round |
|---|---|---|
| 1 | 6 | — (first full audit) |
| 2 | 6 | 4 |
| 3 | 1 | **0** (the one finding pre-dated the PR's round-2 tip) |
| 4 | **0** | 0 |

Nine of the thirteen findings were **claims in prose wider than the code**, not logic defects — including three consecutive wrong mechanisms in the same `errors.md` paragraph, which round 3 finally settled by *running* `annotate_staleness` rather than reading it.

Two nits, neither a finding: the new param asserts `pytest.raises(AssertionError)` without `match=` (justified — a sibling param raises a different `AssertionError` and M1 proved the type alone discriminates); and both the docstring and the `:74` comment name CPython's 4300-digit limit as fixed when it is `sys.get_int_max_str_digits()` and tunable — the guard is correct at any permitted setting.

**Gate scope, stated plainly:** rounds 3 and 4 ran the **dev-host** tier only. Tekton is running the repo-wide gate on `4019e07e` now, and merge waits on it.

PR #998, opened 2026-08-29
fix(seed): the push verdict compared COUNTS, so a correct multi-host push exited 7 — and a broken single-host one could pass

ZacxDev:
`seed.sh`'s push verdict was `remote_entries != staged_entries -> exit 7`. That is only correct while exactly **one** host ever seeds.

The store is per-host and unreplicated, and the tar extract adds and overwrites but never deletes — so a second host's entries legitimately sit in `$DEST` that this stage never held. The count then failed a **correct** push, and did it *after* the content had landed: a failure verdict on a push that worked, which invites a retry that changes nothing. That is the same shape the tar member list a few lines above was already fixed for.

**Measured 2026-08-28** while finishing criterion 8: this host staged 129 over a pod holding 75 — a strict subset, so equality happened to hold and the re-seed passed. The other host's ~26 would stage against a remote of 129+ and exit 7 every time, leaving criterion 8 unfinishable by the tool meant to finish it.

### This is a strengthening, not a relaxation

Count-equality was also **weaker than it looked** in the single-host case it was written for: 129 staged against 129 remote passes while one staged file is missing and one foreign file makes up the number. The question a push actually has to answer is *"did everything I staged land?"* — a subset check on **names**, which is also the `comm -23` the re-seed card prescribes.

So the new guard fails strictly more broken pushes than the count did, and stops failing the correct ones. Extra remote entries are now reported as a `NOTE` beside the verdict, never instead of it. A missing entry is named, rather than leaving the operator a number and a manual diff.

Both sides now use `-mindepth 2 -maxdepth 2`, so they answer the same question — `<scope>/<entry>.md`. The old remote count used `-maxdepth 2` alone, which would also have counted a stray top-level `*.md` the stage cannot contain.

### The push half had no tests at all

`seed.sh`'s header says *"staging is hermetic and testable, pushing needs a cluster"*, so every existing seed test stops at `--stage`. **The verdict deciding whether a push succeeded lived in the only half nothing exercised** — which is how this shipped wrong.

Added a fake `kubectl` (answers `get pod`, runs the real `tar` extract into a real directory, runs the remote `find` there too), so the guard executes against a filesystem the test controls rather than a mock of itself. `$FAKE_DROP` removes one member after the extract to simulate an entry that does not land.

### Red/green matrix

Every regression test was watched to fail against `origin/main`'s `seed.sh`:

| test | pre-change | post-change |
|---|---|---|
| foreign host's entries don't fail a correct push | **RED** (exit 7) | green |
| a dropped entry exits 7 **and names it** | **RED** (unnamed) | green |
| dropped entry caught **when counts match** | **RED** (`assert 0 == 7`) | green |
| positive control — fake cluster really receives the push | green | green |

The third is the blind spot demonstrated live: pre-change output read `remote_entries=3 local_entries=3` / `seed: OK` on a push that had dropped an entry.

Suite: **606 passed** (was 602). Also exercised end-to-end against the live pod — `seed: OK all 129 staged entries are present on the pod`, rc=0, reads 200, legacy writes still 403.

https://claude.ai/code/session_01Nc4ReML29g4pJos3BshJAh

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -115,10 +115,17 @@
       # encrypt/decrypt round trip (it generates its own throwaway identity; the
       # operator's key is never touched), so without age here the whole file is an
       # import error inside the sandbox rather than a green-with-skips.
+      # glibc.bin: provides `locale`, WITHOUT WHICH A COLLATION TEST CANNOT RUN.
+      # MEASURED — the sandbox has no `locale` binary, so
+      # test_subsystem_store_api.py's collation regression fell back to C and its
+      # assertions became trivially true THERE while passing on the dev host:
+      # with both `comm` calls unpinned, that tier reported 1 failure instead of
+      # 2, i.e. the defect was invisible on the tier that gates the merge.
+      # Paired with LOCALE_ARCHIVE below — the binary alone lists only C.
       gateTools = [
         gatePyEnv pkgs.bash pkgs.ripgrep pkgs.git pkgs.util-linux pkgs.jq
         pkgs.gnugrep pkgs.curl pkgs.nodejs pkgs.nix pkgs.opencode pkgs.logrotate
-        pkgs.rsync pkgs.zsh pkgs.age
+        pkgs.rsync pkgs.zsh pkgs.age pkgs.glibc.bin
       ];
     in
     {
@@ -199,6 +206,10 @@
           # Set in BOTH tiers (here and checks.pytests) or the sandbox would
           # misclassify its own repo defects as environment faults.
           export DEVRC_GATE_ENV=1
+          # Set in BOTH tiers, for the same reason DEVRC_GATE_ENV is: a test
+          # whose subject is COLLATION must see the same locales in each, or the
+          # gating tier silently degrades to C and stops observing the bug.
+          export LOCALE_ARCHIVE=${pkgs.glibcLocales}/lib/locale/locale-archive
           echo "devrc: gate toolchain ready — bash scripts/run-tests.sh ." >&2
         '';
       };
@@ -351,6 +362,10 @@
             patchShebangs src/scripts
             # Same marker the devShell sets — see its shellHook for why.
             export DEVRC_GATE_ENV=1
+            # Same reason, same pairing: `locale` comes from glibc.bin in
+            # gateTools, the locales themselves from here. Without this the
+            # binary lists only `C` and a collation test cannot be made to fail.
+            export LOCALE_ARCHIVE=${pkgs.glibcLocales}/lib/locale/locale-archive
             export HOME="$TMPDIR/home"
             mkdir -p "$HOME"
             cd src
--- a/flake.nix
+++ b/flake.nix
@@ -206,14 +206,26 @@
           # both `comm` calls in seed.sh unpinned, the sandbox reported ONE
           # failure where the dev host reported TWO.
           #
-          # ⚠ An earlier version of this also added `pkgs.glibc.bin` to
-          # gateTools, on the stated grounds that a collation test "cannot run"
-          # without the `locale` binary. MEASURED FALSE: with LOCALE_ARCHIVE set
-          # and no `locale` binary at all, `LC_ALL=en_US.UTF-8 sort` collates
-          # correctly. Only a test's *probe* wanted the binary, and it cost 21
-          # store binaries (getconf, ldd, iconv, …) shadowing the system copies
-          # for every human in `nix develop`. The test now detects the locale by
-          # exercising the capability instead, so the binary is not needed.
+          # ⚠ TWO CORRECTIONS ON THE RECORD, because the first attempt to
+          # correct this was itself wrong.
+          #
+          # An earlier version added `pkgs.glibc.bin` to gateTools, on the
+          # stated grounds that a collation test "cannot run" without the
+          # `locale` binary. MEASURED FALSE: with LOCALE_ARCHIVE set and no
+          # `locale` binary at all, `LC_ALL=en_US.UTF-8 sort` collates
+          # correctly. Only a test's *probe* wanted it; the test now detects the
+          # locale by exercising the capability, so the entry was dropped.
+          #
+          # 🔴 The commit that dropped it then justified the removal by saying
+          # the entry had been "shadowing the system getconf/ldd/iconv for every
+          # human in `nix develop`". ALSO MEASURED FALSE — and after the
+          # removal: `nix develop` still resolves `locale` and `getconf` from a
+          # STORE glibc-bin, because `mkShell`'s stdenv cc-wrapper supplies it
+          # regardless (a bare `mkShell { packages = []; }` resolves them too).
+          # What the removal actually changes is the SANDBOX, which is
+          # `runCommandLocal`/stdenvNoCC and has no cc-wrapper — the opposite
+          # tier from the one that claim credited. The entry is still right to
+          # drop, for the first reason and not the second.
           export LOCALE_ARCHIVE=${pkgs.glibcLocales}/lib/locale/locale-archive
           echo "devrc: gate toolchain ready — bash scripts/run-tests.sh ." >&2
         '';
--- a/flake.nix
+++ b/flake.nix
@@ -115,17 +115,10 @@
       # encrypt/decrypt round trip (it generates its own throwaway identity; the
       # operator's key is never touched), so without age here the whole file is an
       # import error inside the sandbox rather than a green-with-skips.
-      # glibc.bin: provides `locale`, WITHOUT WHICH A COLLATION TEST CANNOT RUN.
-      # MEASURED — the sandbox has no `locale` binary, so
-      # test_subsystem_store_api.py's collation regression fell back to C and its
-      # assertions became trivially true THERE while passing on the dev host:
-      # with both `comm` calls unpinned, that tier reported 1 failure instead of
-      # 2, i.e. the defect was invisible on the tier that gates the merge.
-      # Paired with LOCALE_ARCHIVE below — the binary alone lists only C.
       gateTools = [
         gatePyEnv pkgs.bash pkgs.ripgrep pkgs.git pkgs.util-linux pkgs.jq
         pkgs.gnugrep pkgs.curl pkgs.nodejs pkgs.nix pkgs.opencode pkgs.logrotate
-        pkgs.rsync pkgs.zsh pkgs.age pkgs.glibc.bin
+        pkgs.rsync pkgs.zsh pkgs.age
       ];
     in
     {
@@ -206,9 +199,21 @@
           # Set in BOTH tiers (here and checks.pytests) or the sandbox would
           # misclassify its own repo defects as environment faults.
           export DEVRC_GATE_ENV=1
-          # Set in BOTH tiers, for the same reason DEVRC_GATE_ENV is: a test
-          # whose subject is COLLATION must see the same locales in each, or the
-          # gating tier silently degrades to C and stops observing the bug.
+          # 🔴 LOCALE_ARCHIVE IS THE LOAD-BEARING HALF, AND IT IS SET IN BOTH
+          # TIERS FOR THE SAME REASON DEVRC_GATE_ENV IS. A test whose subject is
+          # COLLATION must see the same locales in each, or the gating tier
+          # silently degrades to C and stops observing the bug: measured, with
+          # both `comm` calls in seed.sh unpinned, the sandbox reported ONE
+          # failure where the dev host reported TWO.
+          #
+          # ⚠ An earlier version of this also added `pkgs.glibc.bin` to
+          # gateTools, on the stated grounds that a collation test "cannot run"
+          # without the `locale` binary. MEASURED FALSE: with LOCALE_ARCHIVE set
+          # and no `locale` binary at all, `LC_ALL=en_US.UTF-8 sort` collates
+          # correctly. Only a test's *probe* wanted the binary, and it cost 21
+          # store binaries (getconf, ldd, iconv, …) shadowing the system copies
+          # for every human in `nix develop`. The test now detects the locale by
+          # exercising the capability instead, so the binary is not needed.
           export LOCALE_ARCHIVE=${pkgs.glibcLocales}/lib/locale/locale-archive
           echo "devrc: gate toolchain ready — bash scripts/run-tests.sh ." >&2
         '';
@@ -362,9 +367,9 @@
             patchShebangs src/scripts
             # Same marker the devShell sets — see its shellHook for why.
             export DEVRC_GATE_ENV=1
-            # Same reason, same pairing: `locale` comes from glibc.bin in
-            # gateTools, the locales themselves from here. Without this the
-            # binary lists only `C` and a collation test cannot be made to fail.
+            # Same reason as the devShell's copy — see its shellHook, which
+            # also records why `pkgs.glibc.bin` is NOT here. Without this the
+            # sandbox has only `C` and a collation test cannot be made to fail.
             export LOCALE_ARCHIVE=${pkgs.glibcLocales}/lib/locale/locale-archive
             export HOME="$TMPDIR/home"
             mkdir -p "$HOME"
--- a/scripts/subsystem-store-api/seed.sh
+++ b/scripts/subsystem-store-api/seed.sh
@@ -111,24 +111,51 @@ for d in "$STAGE"/*/; do
   [[ -d "$d" ]] || continue
   staged_scopes=$((staged_scopes + 1))
   name=$(basename "$d")
-  # `index(...)==1` is a LITERAL prefix test — a scope name is not a regex, and
-  # `grep -c "^$name/"` would mis-count one containing `.` or `[`.
-  n=$(awk -v p="$name/" 'index($0, p) == 1 {c++} END {print c + 0}' "$staged_list")
+  # `index(...)==1` is a LITERAL PREFIX test on purpose — twice over. A scope
+  # name is not a regex, so `grep -c "^$name/"` mis-counts one containing `.` or
+  # `[`; and `index(...) > 0` would be a SUBSTRING test, which with scopes `foo`
+  # and `xfoo` counts `xfoo/n.md` against `foo`.
+  # 🔴 Passed through ENVIRON, not `awk -v`: `-v` interprets escape sequences,
+  # so a scope literally named `a\tb` would be searched for as `a<TAB>b` and
+  # report entries=0 beside a non-zero total.
+  n=$(P="$name/" awk 'index($0, ENVIRON["P"]) == 1 {c++} END {print c + 0}' "$staged_list")
   printf 'seed: staged scope %-28s entries=%s\n' "$name" "$n"
-  [[ -L "${d%/}" ]] && excluded+=("$name (symlink — tar ships the link, not its contents)")
 done
 
 # 🔴 LOUD, NOT SILENT. A scope that ships nothing used to produce either a wrong
 # MISMATCH (dot-directory) or a false OK (symlink). Excluding it from the
 # verdict is correct; saying nothing about it would just move the lie.
+#
+# 🔴 BOTH ARMS USE THE SAME PREDICATE — "does it actually hold a `.md`" — and
+# that is a fix, not symmetry for its own sake. The symlink arm used to fire
+# unconditionally, so a symlinked scope holding NO markdown produced the header
+# "1 scope director(ies) hold .md files that will NOT be shipped" over a
+# directory that holds none. In a change whose whole subject is not claiming
+# what has not been established, that sentence was exactly the wrong one to
+# leave. `find "$d"` with the trailing slash follows the link deliberately here:
+# the question is what the operator would LOSE, which is the target's contents.
+_holds_md() { [[ -n "$(find "$1" -maxdepth 1 -name '*.md' -print -quit 2>/dev/null)" ]]; }
+
+# `shopt -p` captures the CALLER's settings and the eval restores exactly those
+# — `shopt -u` unconditionally would clear an option the caller had set.
+#
+# 🔴 `|| true` IS LOAD-BEARING, NOT DEFENSIVE. `shopt -p` EXITS 1 when any named
+# option is UNSET — which is the ordinary case here — so under `set -e` the bare
+# assignment aborted the script after the per-scope lines and before the stamp.
+# Measured: rc 1, and 20 of the file's seed tests went red at once. It still
+# PRINTS the correct restore commands on that non-zero exit, so the capture is
+# sound; only the status is misleading.
+_shopt_saved=$(shopt -p nullglob dotglob || true)
 shopt -s nullglob dotglob
 for d in "$STAGE"/*/; do
   name=$(basename "$d")
-  [[ "$name" == .* ]] || continue
-  [[ -n "$(find "$d" -maxdepth 1 -name '*.md' -print -quit)" ]] &&
-    excluded+=("$name (dot-directory — not in the tar member list)")
+  if [[ "$name" == .* ]]; then
+    _holds_md "$d" && excluded+=("$name (dot-directory — not in the tar member list)")
+  elif [[ -L "${d%/}" ]]; then
+    _holds_md "$d" && excluded+=("$name (symlink — tar ships the link, not its contents)")
+  fi
 done
-shopt -u nullglob dotglob
+eval "$_shopt_saved"
 if [[ ${#excluded[@]} -gt 0 ]]; then
   echo "seed: NOTE ${#excluded[@]} scope director(ies) hold .md files that will NOT be shipped, and are excluded from the count and the verdict:"
   printf 'seed:   %s\n' "${excluded[@]}"
--- a/scripts/subsystem-store-api/seed.sh
+++ b/scripts/subsystem-store-api/seed.sh
@@ -154,12 +154,61 @@ tar -C "$STAGE" -cf - -- "${members[@]}" \
   | kubectl -n "$ns" exec -i "$pod" -- \
       tar -C "$DEST" --no-same-owner --no-same-permissions -xf -
 
-remote_entries=$(kubectl -n "$ns" exec "$pod" -- \
-  find "$DEST" -maxdepth 2 -name '*.md' -type f | wc -l)
-echo "seed: PUSHED pod=$pod dest=$DEST remote_entries=$remote_entries local_entries=$staged_entries"
+# 🔴 CONTAINMENT ON NAMES, NOT EQUALITY OF COUNTS. This guard used to read
+# `remote_entries != staged_entries -> exit 7`, which is only correct while
+# exactly ONE host ever seeds.
+#
+# The store is PER-HOST and unreplicated, and the extract adds and overwrites
+# but never deletes — so a second host's entries legitimately sit in $DEST that
+# this stage never held. Counting then failed a CORRECT push, and did it AFTER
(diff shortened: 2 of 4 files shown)


ZacxDev:
## Audit round 1 — findings and what the fixes claim

Full adversarial audit run against `be568a35`. Two 🔴, two 🟡, three 🟢. Both 🔴 were reproduced independently before fixing.

**🔴 1 — `comm` collated in the AMBIENT locale over `LC_ALL=C`-sorted lists.** Both lists were sorted `LC_ALL=C`; neither `comm` was. GNU `comm` order-checks in the current locale (`LANG=en_US.UTF-8` here), so C-sorted input reads as unsorted — but only once a line is *unpairable*, which is exactly the multi-host case this PR exists to fix. Result: `comm` rc=1, `set -e` kills the script, and **no verdict is printed at all** — no PUSHED, no NOTE, no OK, no MISMATCH, after the content has landed. The same "failure verdict on a push that worked" shape the block's own comment condemns, reintroduced one line lower. The PR's live run could not see it: it had `staged == remote` exactly, so nothing was unpairable and the check never armed. `--nocheck-order` is **not** the fix — the set operation itself collates.

**🔴 2 — the test stub wrote `#!/usr/bin/env bash`, which is dead in the gating sandbox.** The two tiers disagreed in the way `test_runtime_shebangs.py` exists to catch: dev host showed **1 failed** (the guard), the merge gate showed **5 failed** — the guard *plus all four new tests*, which never executed (`bad interpreter`, rc 126). The class had **zero coverage on the only tier that gates the merge**.

**🟡 3 — a dot-directory scope re-created this guard's own bug.** `staged_entries` and the tar member list both come from `"$STAGE"/*/`, which without `dotglob` skips dot-dirs; a bare `find` does not. A `.hidden/` scope would be listed as staged, never archived, and reported missing — "1 of 1 staged entry file(s) did NOT land" over something never shipped. Latent (no dot-scopes in the store today).

**🟡 4 — five mutants survived**, because the fixture was all-lowercase, all hyphen-slug, with no top-level `.md` and no dot-dir — none of the dimensions where the real store differs.

**🟢 5** — `server.py`'s `_FRESHNESS_MAXDEPTH` comment quoted seed.sh's old find verbatim and had rotted.

### What the fixes claim

```audit-claims round=1 audited=be568a35..34265904
1. LC_ALL=C now applies to both comm invocations, not only to the two sorts.
2. Both find listings exclude dot-directories, so the compared population is the pushed one.
3. The kubectl stub is POSIX sh and mockbin.write_exec owns its shebang; no test writes its own.
4. Three tests added for the dimensions the fixture could not see: a UTF-8 locale with an inverting pair, a top-level *.md in the store root, and a dot-directory scope.
5. The locale test carries a control asserting its pair still inverts between C and en_US, without which the collation mutant is invisible.
6. server.py's _FRESHNESS_MAXDEPTH comment states the shared invariant instead of quoting seed.sh's source.
```

### Mutation results (re-run, not asserted)

| mutant | round-1 result | now |
|---|---|---|
| drop `LC_ALL=C` from both `comm` calls | **SURVIVED** | killed by the locale test |
| drop the dot-dir exclusion | — | killed by the dot-scope test |
| drop `-mindepth 2` from the staged find | **SURVIVED** | killed by the top-level-`.md` test |

The collation mutant survived my *first* attempt at a locale test too: `README.md` beside `thing-alpha.md` orders identically under C and en_US (`R`<`t`, `r`<`t`), so `comm`'s order check never armed and the test passed either way. `backblaze.md` supplies the real inversion, and the test now asserts the pair still inverts so it cannot silently stop pinning the dimension.

Both tiers are being re-run against `34265904`.

ZacxDev:
## Audit round 2 — findings and what the fixes claim

Delta re-audit of `be568a35..0aecdbf2`, dispatched blind. Both round-1 🔴s confirmed actually fixed. Three 🟡, four 🟢. The two that mattered:

**🟡 1 — a SYMLINKED scope produced a false success, and it refutes a claim I made in this PR.** Reproduced before fixing:

```
seed: PUSHED … remote_entries=1 staged_entries=2
seed: OK all 2 staged entries are present on the pod     (rc 0)
```

The counts disagree one line apart and it still reports OK; one entry never landed. `staged_entries` walked `"$STAGE"/*/` with a trailing slash, which RESOLVES a symlink, while the comparison's `find .` does not descend one and `tar` ships the link rather than its target — two walks with different rules, and the comparison believed the wrong one.

🔴 **The old count-equality check WOULD have caught this** (1 ≠ 2 → exit 7). So this PR's claim that containment "fails strictly more broken pushes than the count did" was **false as written**. It is true only now that both sides come from a single `_shippable_entries` predicate. The comment has been corrected in place rather than quietly dropped.

**🟡 2 — the locale test was VACUOUS on the gating tier**, proven rather than inferred: the nix sandbox has no `locale` binary, so it fell back to C and its assertions became trivially true. With both `comm` calls unpinned the sandbox reported **1** failure where the dev host reported **2** — the defect was invisible exactly where it gates.

**🟡 3 — my structural `comm` guard was SPELLED, not structural.** The audit walked it three ways (`/usr/bin/comm -23`, `_cmp=comm` then `$_cmp -23`, a second `comm` on an already-prefixed line) and it *also* rejected three safe spellings (`env LC_ALL=C comm`, a line continuation, a global `export LC_ALL=C`). Deleted, not patched — it asserted a word and the hazard has other shapes.

### What the fixes claim

```audit-claims round=2 audited=0aecdbf2..7778d401
1. staged_entries and the compared set now come from ONE predicate, _shippable_entries, so a symlinked scope cannot make them disagree.
2. A scope directory that ships no entries (symlink or dot-directory) is named in a NOTE rather than silently excluded.
3. The false "strictly more broken pushes than the count did" claim is corrected in the comment where it was made.
4. flake.nix supplies a real locale to BOTH tiers: pkgs.glibc.bin in gateTools, LOCALE_ARCHIVE in the devShell and in checks.pytests.
5. The locale test now FAILS loudly if en_US.UTF-8 is unavailable, instead of skipping or degrading to C.
6. The spelled comm-collation guard is deleted; a test asserting the two find expressions are IDENTICAL replaces it.
7. set -u is restored in the POSIX kubectl stub, which the bash-to-POSIX rewrite had dropped.
```

### Mutation results (re-run, not asserted)

| mutant | round 2 | now |
|---|---|---|
| remote `find` drops the dot exclusion | **SURVIVED** | killed |
| `-maxdepth 2` → `3` on both finds | **SURVIVED** | killed |
| `-type f` dropped on both finds | **SURVIVED** | killed |
| `_shippable_entries` follows symlinks (`find -L`) | — | killed (2 tests) |
| excluded-scope NOTE removed | — | killed (2 tests) |
| `LC_ALL=C` dropped from both `comm` calls | killed | killed |

### Both tiers, with a watched control

- dev host: **620 passed, 0 skipped**
- gate (`nix build .#checks…pytests`): **collected=9798 passed=9798 skipped=0**, `RESULT: PASS (exit=0)`
- control for the locale claim: with `locale` made unavailable, the test **FAILS** — so its pass in the sandbox is evidence the locale is really there, not a silent fallback.

On the count challenge in round 2: 610 was `test_subsystem_store_api.py` alone; my 619 was that file **plus** `test_runtime_shebangs.py`'s 9. Both were right; it is 620 now.

ZacxDev:
## Audit round 3 — findings and what the fixes claim

Delta re-audit of `0aecdbf2..7778d401`, dispatched blind. **No 🔴.** Two 🟡 and seven 🟢, five of the 🟢 backed by surviving mutants. Both 🟡 were mine.

**🟡 F1 — a false sentence written into shared infrastructure.** My `flake.nix` comment said `pkgs.glibc.bin` is "WITHOUT WHICH A COLLATION TEST CANNOT RUN". **Measured false:** with `LOCALE_ARCHIVE` set and no `locale` binary at all, `en_US` collation works. Only the test's *probe* wanted the binary — at the cost of 21 store binaries (`getconf`, `ldd`, `iconv`, …) shadowing the system copies for everyone in `nix develop`. Fixed by removing `glibc.bin` outright and having the test detect the locale by **exercising the capability** (does `sort` invert the pair?) rather than probing for a binary. The gate now passes with no `glibc.bin` and no `locale` binary present, which is the control on that claim.

**🟡 F2 — the find-identity guard was blind past `-type f`.** Non-greedy capture, so appending `-o -type d` to the remote listing **survived a fully green run** — depth-2 directories would enter `remote_list` and be reported as foreign entries on every push. Now captures the whole expression, plus the `cd` target.

**🟢 F3** — the symlink arm of the excluded-scope NOTE fired unconditionally, so the header announced ".md files" for a scope holding none. Both arms probe now.

### Three of my own failures, stated plainly

1. **My F6 fix broke the whole script.** `shopt -p` exits **1** when the named options are unset, so capturing it under `set -e` aborted `seed.sh` after the per-scope lines — 20 tests red at once. Caught by the suite, not by reading.
2. **`git checkout HEAD -- <path>` ate my uncommitted fixes, twice**, when used to revert a mutation in a file I had unstaged work in. Switched to `cp`-aside restore.
3. **Two of my guards were green for the wrong reason.** The `set -u` restoration *survived its own mutant* — the only `$FAKE_DEST` reference is inside `$( … )`, so an unbound variable kills the subshell while the stub continues with an empty substitution. And the test asserted only `rc != 0`, which held either way because the stub fails later on `tar -C ""`. Now a top-level `:?` guard, with the test asserting **that guard's own message**.

### What the fixes claim

```audit-claims round=3 audited=7778d401..a2e65a87
1. pkgs.glibc.bin is removed from gateTools; LOCALE_ARCHIVE alone supplies the locale, and the comment that said otherwise is corrected.
2. The locale test detects availability by sorting an inverting pair under C and en_US, not by probing for a locale binary.
3. The find-identity guard captures the whole expression and the cd target, so a clause appended after -type f is no longer invisible.
4. Both arms of the excluded-scope NOTE probe for a .md before naming a directory.
5. Scope names reach awk through ENVIRON, so an escape sequence in a name cannot mis-count.
6. shopt -p is captured with || true, because it exits 1 when the options are unset and set -e aborted the script.
7. The stub guards FAKE_DEST with a top-level :?, and its test asserts that guard's own message rather than a bare non-zero rc.
8. New tests pin prefix-not-substring counting, that a dot-scope does not SHIP, and that an empty dot/symlink scope is not announced.
```

### Mutation sweep — every round-3 survivor is dead

| mutant | round 3 | now |
|---|---|---|
| remote find `-o -type d` appended | **SURVIVED** | killed |
| awk prefix → substring (`foo` vs `xfoo`) | **SURVIVED** | killed |
| per-scope `entries=N` hardcoded | **SURVIVED** | killed |
| `set -u` / `:?` removed from the stub | **SURVIVED** | killed |
| `dotglob` left set | **SURVIVED** | killed |
| dot-dir probe drops `-name '*.md'` | **SURVIVED** | killed |

### Both tiers

- dev host: **623 passed, 0 skipped** (`test_subsystem_store_api.py` + `test_runtime_shebangs.py`)
- gate: **collected=9801 passed=9801 skipped=0**, `RESULT: PASS (exit=0)`, only the 2 pinned skips

On the count challenge: **611** is `test_subsystem_store_api.py` alone; my figure was that file *plus* `test_runtime_shebangs.py`'s 9. Both were right — I should have named both files rather than quoting a combined number, and that is twice it caused confusion.

`round 3 · payload lines changed THIS round: 76 (seed.sh 43, flake.nix 33) · scaffolding 211`

ZacxDev:
## Audit round 4 — findings and what the fixes claim

Delta re-audit of `7778d401..a2e65a87`, dispatched blind. **No 🔴.** Six 🟡, two 🟢 — and the sharpest round so far, because two of them were my round-3 *fixes* making things worse.

**🟡 F3 — my `shopt` "fix" was a regression.** Round 3 replaced an unconditional `shopt -u nullglob dotglob` with "restore the caller's settings". The tar member list is `"$STAGE"/*/`, correct only while `dotglob` is OFF — so under `BASHOPTS=dotglob` a dot-scope **actually shipped** to the pod while the remote listing's `! -path './.*'` still excluded it: shipped, never verified, and the NOTE saying "not in the tar member list" became false. Measured against `7778d401`, which handled it correctly. Reverted.

The rationale was hollow too: shell options do not propagate out of a script (this file is only ever `bash seed.sh`), so the restored property was unobservable — and it was **already violated** for `nullglob`, which line 62 clears unconditionally 86 lines earlier.

**🟡 F1 — my correction of a false comment was itself false.** I justified removing `glibc.bin` by saying it had been shadowing the system `getconf`/`ldd`/`iconv` for everyone in `nix develop`. Verified false, and after the removal: `nix develop` still resolves them from a store glibc-bin, because `mkShell`'s stdenv cc-wrapper supplies it regardless — a bare `mkShell { packages = []; }` resolves them too. What the removal changes is the **sandbox** (`runCommandLocal`/stdenvNoCC, no cc-wrapper) — the opposite tier from the one that claim credited. Both retractions are now on the record rather than a third confident sentence.

**🟡 F4 — `2>/dev/null` turned a loud over-report into silence.** An unreadable symlink target produced no NOTE, rc 0, `seed: OK` — under a block headed "LOUD, NOT SILENT". `_shippable_entries` does not descend the symlink either, so nothing surfaced it. A probe that cannot answer now announces.

**🟡 F5 — the textual find-identity guard lost its race a second time.** After being "fixed" to capture past `-type f`, an escaped `\( … \)` group recreated the identical hole and stayed green across the whole class, with the runtime effect the guard exists to prevent (depth-2 directories reported as another host's entries on every push). Replaced with a **behavioural** test: a depth-2 directory on the pod. A guard over a shell expression's spelling keeps finding new shapes to miss.

**🟡 F6/F7** — the prefix fixture had every true count equal to 1, so hardcoding `n=1` survived (constant-equals-fixture), and `xfoo` extends `foo` on the left so the `/` separator was untested; the `ENVIRON` fix had no test at all.

### What the fixes claim

```audit-claims round=4 audited=a2e65a87..d779c5d1
1. The shopt restore is reverted to an unconditional `shopt -u nullglob dotglob`, so the tar member list no longer depends on the ambient shell.
2. A BASHOPTS=dotglob case pins that, since every other test ran with default shell options.
3. _holds_md no longer swallows stderr; a probe that errors announces the scope instead of dropping it.
4. flake.nix records BOTH retractions: the glibc.bin entry was wrong, and the stated reason for removing it was also wrong.
5. The stale "glibc.bin in gateTools" requirement, left four lines above its own correction, is gone.
6. The two-listings invariant is pinned behaviourally by a depth-2 directory on the pod, not by matching the find expression's text.
7. The prefix fixture now has a scope with two entries and a foo/foobar pair, defeating a hardcoded n=1 and a dropped separator.
8. New tests cover the ENVIRON count, the elif, _holds_md's -maxdepth 1, and an unreadable symlink target.
```

### Mutation sweep — all eight die

| mutant | round 4 | now |
|---|---|---|
| escaped-paren find (`\( … \) -o -type d`) | **SURVIVED** | killed — by the *behavioural* test |
| per-scope count hardcoded `n=1` | **SURVIVED** | killed |
| dropped `/` separator | **SURVIVED** | killed |
| `ENVIRON` → `awk -v` | **SURVIVED** | killed |
| `elif` → second `if` | **SURVIVED** | killed |
| `_holds_md` `-maxdepth 1` → `2` | **SURVIVED** | killed |
| shopt restore reintroduced | — | killed (under `BASHOPTS=dotglob`) |
| `2>/dev/null` reintroduced | — | killed |

### Both tiers

- dev host: **629 passed, 0 skipped**
- gate: **collected=9807 passed=9807 skipped=0**, `RESULT: PASS (exit=0)`

Correction to my own figures: the dev-host split is **614 + 9**, not the 611 + 12 I posted last round. The totals were right; the decomposition was wrong twice, which is why I now quote both file names.

`round 4 · payload lines changed THIS round: 73 (seed.sh 45, flake.nix 28) · scaffolding 181`

ZacxDev:
## Audit round 5 — findings and what the fixes claim

Delta re-audit of `a2e65a87..d779c5d1`, dispatched blind. **No 🔴.** Four 🟡, three 🟢. Round 4's claims 1, 4, 5, 7 and 8 verified **actually fixed** with independent measurement.

**🟡-1 — the fix for round 4's silence reintroduced the unestablished-claim defect one layer over.** Round 4 found that `2>/dev/null` made an unreadable symlink target *silent*. Announcing on any probe error fixed the silence — but reused the wording, so a target that is unreadable **and holds no markdown at all** was announced under a header asserting it holds some:

```
seed: NOTE 1 scope director(ies) hold .md files that will NOT be shipped, …
seed:   emptylocked (symlink — tar ships the link, not its contents)
```

Same fixture on `a2e65a87`: no NOTE at all. So the delta introduced it. This is the exact claim the PR exists to stop making, and my own sibling test states it as an absolute invariant — but its fixture used a target that *does* hold markdown, so it never reached the case.

The probe now returns **three states** — holds / does not / **could not check** — using `find`'s exit status as the discriminator (measured: rc 1 unreadable, rc 0 readable). Each state gets wording matching what was established; the header asserts nothing about contents.

**🟡-2 — the `BASHOPTS=dotglob` test had no control that `dotglob` armed.** It is the *sole* killer of round 4's shopt regression, so a silent vacuity there re-opens that whole class with a green suite. It would have passed just as well had `BASHOPTS` been ignored entirely. The control was free and already in the output: the per-scope loop honours the ambient option, so a `.hidden` line appears only when it really armed.

**🟡-3 — the depth-2 behavioural guard was blind to the plainest mutant.** Its fixture was a directory named `a-subdirectory`, excluded by `-name '*.md'` rather than by `-type f` — so dropping `-type f` from the remote find stayed invisible, caught only by the *textual* guard this test was written to replace. Renamed to `a-subdirectory.md`: now the only clause that can exclude it is `-type f`, and it is the shape the server 503'd on.

### What the fixes claim

```audit-claims round=5 audited=d779c5d1..3dde94a6
1. _md_state returns three states (holds / does not / could not check) from find's exit status, replacing the two-state probe.
2. The NOTE header no longer asserts anything about contents; each entry carries its own reason, and the unreadable cases say UNREADABLE.
3. The BASHOPTS=dotglob test asserts the per-scope line that appears only when dotglob actually armed.
4. The depth-2 fixture is named a-subdirectory.md, so -type f is the only clause that can exclude it.
5. The unreadable-symlink test gained the returncode assertion its siblings have.
6. A new test covers an unreadable target holding NO markdown — the case the existing one structurally could not reach.
```

### Mutation sweep — all five die

| mutant | result |
|---|---|
| header reverts to asserting "hold .md files" | killed (2 tests) |
| UNREADABLE arm reuses the holds-markdown wording | killed |
| remote find drops `-type f` | killed (2 tests — the behavioural guard now catches it) |
| `_md_state` stops distinguishing the error state | killed |
| shopt reverts to caller-restoring | killed — by the BASHOPTS test alone |

### Both tiers

- dev host: **630 passed, 0 skipped**
- gate: **collected=9808 passed=9808 skipped=0**, `RESULT: PASS (exit=0)` — a real build this time, not a cache hit

### Deferred deliberately, with the reason

- **🟡-4** — the *scope count* (`STAGED scopes=N` and the per-scope lines) is still ambient-shell-dependent on the same axis the member list was just hardened against. Pre-existing, predates this PR, and needs an explicit `export BASHOPTS` or `BASH_ENV` to reach.
- **🟢-3** — `END {print c + 0}` → `print c` survives, via the fixture's empty scope. Also pre-existing.

Both belong in one follow-up issue against `seed.sh`, closed when that PR merges — widening this diff further to reach them would trade a bounded change for an unbounded one.

`round 5 · payload lines changed THIS round: 54 raw / 23 non-comment (seed.sh only; flake.nix untouched) · scaffolding 64`

ZacxDev:
## Audit round 6 — findings and what the fixes claim

Delta re-audit of `d779c5d1..3dde94a6`, dispatched blind. **No 🔴.** Two 🟡 (both scaffolding) and three 🟢.

🔴 **The important result is a negative one: the fix-introduces-the-next-defect pattern did NOT recur in the payload.** Rounds 3, 4 and 5 each found that the previous round's fix created the next defect. Round 6 checked the mirror case I was most worried about — whether `find`'s exit status can report "could not check" for a scope that is actually fine — and measured it across GNU find 4.11 and bfs 4.1 over 14 constructed cases: at `-maxdepth 1` an unreadable **sub**directory, a broken symlink child, and a child directory named `*.md` all exit 0. The only rc≠0 sources are the starting point itself. No healthy scope is announced UNREADABLE.

**🟡-1 — my new test's announcement assertion was vacuous.** `assert "emptylocked" in r.stdout` is satisfied by the unrelated per-scope line `staged scope emptylocked entries=0`, not by the NOTE. Reverting `_md_state`'s error state to a plain "no" — the exact round-4 silence defect — removed the NOTE entirely and this test stayed **green**; only its sibling went red. The half its docstring calls "Announcing it is right" guarded nothing, in the test written for the case the sibling structurally cannot reach.

**🟡-2 — a docstring that would have talked a maintainer into breaking the code.** Two still named `_holds_md`, deleted the round before. Worse, one presented `2>/dev/null` as *the past defect* — while the shipped `_md_state` deliberately carries it and depends on it. The defect was discarding `find`'s **exit status**. Someone reading that docstring and then the function would reasonably delete the redirect, and that mutant survives all 19 tests in the class.

**🟢-3** — `|| return 2` discarded `$out` unread, so a probe that both matched and errored would report "could not check" over a match it had. Unreachable at `-maxdepth 1`, one character from reachable, and nothing said the discriminator depended on the depth.

**🟢-4** — the header's second over-claim, on a new axis. "will NOT ship" is false for a symlinked scope, which **does** ship — as a symlink — as the very next line of output says. The block contradicted itself two rows apart.

**🟢-5** — the dot arm's UNREADABLE state had zero coverage; deleting it left every test green. Reachable only via a dot-*named symlink*: a real unreadable dot directory fails in `rsync` first (rc 23).

### What the fixes claim

```audit-claims round=6 audited=3dde94a6..48f3910f
1. The unreadable-no-markdown test scopes its announcement assertion to the NOTE lines and also asserts UNREADABLE, so the round-4 silence mutant now fails it.
2. Both stale _holds_md docstrings are corrected, and the one blaming 2>/dev/null now says the defect was discarding find's exit status and that the redirect is load-bearing.
3. _md_state checks the match BEFORE the rc, so evidence beats the error at any depth.
4. The NOTE header reads "contribute NO entries", true of every state that reaches it, and is pinned as a whole normalised string.
5. The dot arm's UNREADABLE state is covered by a dot-named symlink to an unreadable target.
```

### Mutation sweep

| mutant | result |
|---|---|
| `_md_state` error state → plain "no" (the round-4 silence) | killed (3 tests) |
| revert to error-beats-evidence ordering | killed (3 tests) |
| delete the dot arm's UNREADABLE case | killed |
| header reverts to "will NOT ship" | killed — **by the whole-string pin** |
| delete `2>/dev/null` from the probe | **SURVIVES — and that is correct** |

That last row is deliberate and I am not claiming coverage for it. Removing the redirect only leaks `find`'s stderr; it is not a defect. It mattered as the *destination* of 🟡-2, and the fix for that was the docstring, not a guard.

🔴 **On the header pin:** that one sentence has now been FALSE twice, on two different axes, and a keyword guard caught neither — any rewording satisfies `"hold .md files" not in stdout`. When the artifact under test is prose, the normalised whole is the only machine-readable claim. A deliberate reword will now fail this test and have to be re-verified; that cost is the point.

### Both tiers

- dev host: **631 passed, 0 skipped**
- gate: **collected=9809 passed=9809 skipped=0**, `RESULT: PASS (exit=0)`, a real build (not a cache hit)

⚠ Instrument note: I first reached for `nix build --rebuild` to force a real build; it exited 1 with *"some outputs … are not valid, so checking is not possible"* — that flag verifies an **existing** output, and this derivation was new. Caught by reading the output rather than the exit code.

`round 6 · payload lines changed THIS round: 28 raw / 10 non-comment (seed.sh only) · scaffolding 73`

ZacxDev:
## Audit round 7, and how this PR was merged

### Round 7 findings — no 🔴, two 🟡, two 🟢

**🟡 A — the docstring I wrote to correct a false claim made a new one.** Round 6 replaced "the defect was `2>/dev/null`" with "it deliberately carries `2>/dev/null` and **depends on it**". Measured false, independently, with a valid control (my first attempt compared shas across different stage paths and was invalid):

```
with    redirect: rc=0  stdout-sha=f4a772893193  stderr=0 lines
without redirect: rc=0  stdout-sha=f4a772893193  stderr=1 line
```

Byte-identical stdout, identical rc. The redirect suppresses one expected `Permission denied` diagnostic nothing consumes — **inert**. That is the third comment in this PR to state something measurably false, and the second time a correction of a false claim was itself false.

**🟡 B** — a docstring still quoted the removed "hold .md files" wording as current, and would have sent a maintainer to restore it. The two `assert "hold .md files" not in r.stdout` guards had become **unfalsifiable**: no code path can emit that string.

**🟢 C** — the header's *second* clause was never re-examined while the first was corrected twice. "excluded from the count" is loose: `staged_scopes` counts a symlinked scope and not a dot one, so 3 scope dirs can print "2 excluded" beside `scopes=2`. Now "the **entry** count", true on every arm.

**🟢 D** — the pin hardcoded `1 scope director(ies)`, so a fixture gaining an excluded scope would misdirect at the wording.

All four fixed. The whole-string header pin is now one parameterised helper across three call sites; three header mutants (`entry count`→`count`, revert to `hold .md files`, `${#excluded[@]}`→`${#scopes[@]}`) each die on 3–4 tests.

### 🔴 Round 8 was NOT run — the ladder was stopped by decision, not by convergence

The stop rule is "the first round that returns no findings is the last". Round 7 returned findings, so round 8 was owed and has not happened. The delta it would have audited: one header clause, three docstrings, and a consolidation of three assertions into one shared helper — that last being the "one rule, one place" move that can weaken three tests at once if the helper is wrong. **That risk is unaudited.** Recording it rather than implying the ladder converged.

### 🔴 This was merged with `--admin`, bypassing a red required check

`tekton/devrc-pytests` was **FAILURE** at merge time. The evidence that it is a load flake, not this change:

| | result |
|---|---|
| CI, revision `f11e80ea` | `collected=18296 passed=18293 skipped=2 **failed=1**` |
| my local hermetic build, **same revision** | `collected=18296 passed=18294 skipped=2 **failed=0**` |

- the failing test, `test_ledger_plugin.py::test_the_plugin_writes_a_record_the_READER_can_parse`, is in a subsystem this diff does not touch, and passes **5/5** locally in isolation
- the previous commit `3dde94a6` **passed** CI
- three unrelated revisions failed in the same window on three **different** tests (3, 3 and 5 failures)
- host load average **23.78 / 33.36 / 47.21** with **8 concurrent pipelineruns**

Strong, but it is still a bypass of a required gate, and bypassing is the thing that trains people to click through. Flagged here so it is visible rather than buried.

### Deferred — one follow-up owed against `seed.sh`

- the **scope count** (`STAGED scopes=N` and the per-scope lines) is still ambient-shell-dependent on the same axis the member list was hardened against; needs an explicit `export BASHOPTS` or `BASH_ENV` to reach
- `END {print c + 0}` → `print c` survives, via the shared fixture's empty scope
- dropping `-type f` from the **local** `_shippable_entries` survives every behavioural test (only the textual guard catches it) — the mirror of the remote-side case this PR fixed

All three predate this PR. Closing condition: a follow-up PR against `scripts/subsystem-store-api/seed.sh` covering them, or a maintainer dismissing them in writing on that issue.

### Final state

- dev host: **631 passed, 0 skipped** (`test_subsystem_store_api.py` + `test_runtime_shebangs.py`)
- hermetic gate: **collected=9809 passed=9809 skipped=0**, `RESULT: PASS (exit=0)`
- 7 audit rounds, 2 🔴 (both round 1), 0 🔴 since; rounds 6 and 7 found nothing behavioural in the shipped script

PR #999, opened 2026-08-29
fix(espanso): a search term that NAMES a snippet wins over ones merely containing it

ZacxDev:
## The defect

`acq` resolves **AMBIGUOUS** and every such espanso fire is recorded **UNATTRIBUTED** — live on both hosts right now.

`_token_matches` tests `token in trig.lower()` — a **substring** test. After `:acq` was split into `:dacq` + `:acq`, the bare token `acq` matches BOTH (`"acq" in ":dacq"` is True), so `_attribute("acq")` returns `None`.

Measured across four trees via the test module's own `_live_base()` / `_term_matches`:

| tree | `acq` resolves to |
|---|---|
| `7aff8471` (pre-split) | `:acq`, unique |
| `4f1e4c56` (the split) | **AMBIGUOUS** `[':dacq', ':acq']` |
| `d1a9fd5e` (the fix meant to prevent this) | **still AMBIGUOUS** |
| `6349a8b9` (`origin/main` today) | **still AMBIGUOUS** |

The existing guard cannot see it. `test_live_triggers_have_no_prefix_or_suffix_collisions` tests **whole-trigger** prefix/suffix and `":dacq".endswith(":acq")` is False. That guard is *correct for the DIRECT path* (which matches on the ring's SUFFIX); it is structurally blind to the **SEARCH** path, which matches substrings anywhere. Both halves are now stated in its docstring, with a named sibling for the other half.

## The fix — structural, not a rename

`:dacq` is the operator's deliberate spelling (`d` + `acq`), so renaming it was rejected. Instead, in `_attribute`: when several snippets match and **exactly one of them is NAMED outright** by the term (the term IS that trigger, with or without the leading `:`), that one wins.

This closes the whole `:dX` / `:X` class rather than this instance. It is consulted **only on the already-ambiguous branch**, so it can never re-point a term that resolves uniquely today — verified by measurement, not by reading:

## Resolution diff (the blast radius)

Every term in `_EXISTING_RESOLUTIONS`, `_MT_TERMS`, `_AUDIT_2026_08_19_RESOLUTIONS`, `_PICKER_ROWS`, plus **every trigger bare and colon-spelled** — 101 terms over 26 triggers — dumped before and after and diffed:

```
terms compared: 101   CHANGED: 2
  'acq': None -> ':acq'
  'alo': None -> ':alo'
picker-row (_term_matches) changes: []
```

Both moves are `None -> the correct snippet`. `alo` was a **second live instance of the same defect**, not found by the brief: `alo` is a substring of `talos` in the labels of `:hlt` ("homelab-talos path") and `:cdp` ("civitai datapacket-talos path"), so it matched three snippets. It is not a pinned term anywhere, which is why it was silent.

**Zero `_term_matches` changes**, so no espanso **picker** row is added or removed — attribution and reachability stay separate questions, and `test_live_picker_rows_stay_reachable` is untouched.

Bare-trigger-name sweep before the fix: 24 of 26 already resolved to their own snippet; the two exceptions were exactly `acq` and `alo`.

## RED-at-base matrix

Run at `6349a8b9` with all the new/changed tests in place and only the two source files reverted, under `PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`:

| test | at `6349a8b9` | at HEAD |
|---|---|---|
| `test_espanso_detect.py::test_bare_acq_resolves_to_acq_not_ambiguous` | **RED** `assert None == ':acq'` | green |
| `test_espanso_detect.py::test_acq_end_to_end_through_search_ui` | **RED** `EspansoEvent(trigger=None…).trigger` | green |
| `test_espanso_detect.py::test_live_bare_trigger_names_resolve_to_their_own_snippet` | **RED** `{'acq': …None…, 'alo': …None…}` | green |
| `test_espanso_detect.py::test_live_existing_resolutions_not_made_ambiguous` | **RED** (the new `("acq", ":acq")` row) | green |
| `test_espanso_detect.py::test_single_word_term_behaviour_unchanged` | **RED** `assert None == ':date'` | green |
| `test_espanso_usage.py::test_lint_uniqueness_is_the_detectors_own_attribute_rule` | **RED** `assert ':acq' not in {':acq'}` | green |

Each fails with **its own** assertion, not a neighbour's.

Four other new tests are **GREEN at base by design** — they pin what the tie-break must NOT do. They are labelled **INVARIANT GUARDS** in the source rather than counted as regression coverage: `test_acq_split_interface_words_still_resolve`, `test_naming_tiebreak_only_fires_on_the_ambiguous_branch`, `test_naming_two_triggers_at_once_stays_ambiguous`, `test_acq_split_direct_triggers_are_unaffected`, `test_naming_tiebreak_does_not_reach_the_picker`.

## Mutation sweep

Five mutants, each run with `PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider` (each produced a *different* failure set, which is the evidence the mutant actually executed rather than a cached `.pyc`):

| mutant | killed by |
|---|---|
| revert `_attribute` entirely | the 5 tests above, each on its own assertion |
| `len(exact) == 1` → `if exact` | `test_naming_two_triggers_at_once_stays_ambiguous` (its whole reason to exist) |
| drop the colon-stripped form (`term == low`) | 6 tests |
| equality → substring in `_names_trigger` | 5 tests |
| tie-break accepts every candidate (`exact = list(matched)`) | 5 tests |
| remove the `len(matched) == 1` short-circuit | 12 tests — proves the short-circuit is load-bearing, i.e. the tie-break really is confined to the ambiguous branch |

## Coverage changes

- `("acq", ":acq")` and `("dacq", ":dacq")` **added** to `_EXISTING_RESOLUTIONS`. Rows added, never deleted — per the table's own ANTI-VACUITY note. Its floor was a stale `>= 20` against 24 actual rows (four added rows were unprotected); ratcheted to the real **26**, still `>=` so additions stay green and only a deletion reds.
- New live guard `test_live_bare_trigger_names_resolve_to_their_own_snippet` — the SEARCH-path sibling of the collision guard. Its expectation is **derived from `nix/home.nix`**, not a typed list, so it cannot be weakened by deleting the row that broke and a 27th snippet is covered the day it is added. It carries a **negative control** (seed a trigger spelling an existing one without the colon, so the term names two snippets and the tie-break cannot rescue it) — asserted to fire before the real zero is read.
- `test_search_fuzzy_multiple_match_is_ambiguous` and `test_focus_steal_ambiguous_term_still_emits_trigger_none` switched their ambiguity vehicle from `"date"` to `"dat"`, which matches `:date` + `:datetime` and **names neither**. Their subject (emit the row with `trigger=None`) is unchanged; only the fixture term moved.
- `test_single_word_term_behaviour_unchanged` now pins `_attribute("date") == ":date"` (was `is None`) with the reason inline, plus `_attribute("dat") is None` as the retained ambiguity case. Its comment no longer claims more than it checks. This is the one hermetic-fixture outcome that moved; it cannot arise on the live config, where the prefix/suffix guard forbids a `:date`/`:datetime` pair.

## Second-order: `espanso-usage.py`

Three prose claims that the implementation now contradicts were corrected — two of them are printed **into the audit report**, not just comments (`render_terms`, `render_lint`, and `lint()`'s docstring): "`_attribute` returns None on >=2 matches" is no longer true unconditionally.

`lint()`'s uniqueness test was a **second copy** of `_attribute`'s rule (`len(m) == 1`) and had just diverged from it. Consolidated onto `det._attribute(t) == trig`. Verified a **no-op on today's config**: `declared_terms` is search_terms + label words, and the only one that is also a bare trigger name is `kickoff`, which already matched uniquely. The new test isolates the divergence with a fixture where the bare name is the snippet's *only* resolving term.

## What was run

Local, dev-host tier, base `6349a8b9`, via `nix develop ~/workspace/devrc -c python3 -m pytest … -q -p no:cacheprovider`:

- `scripts/collector/keylog/tests/` — **93 passed**
- `scripts/session-analysis/tests/` — **491 passed**
- `scripts/tests/` (repo-wide guards, `--deselect test_subsystem_store_api.py` for its known flaky race) — see the comment below; it was still running when this PR was opened.

The full `scripts/gate.sh` was **not** run and neither was the `nix build .#checks…` sandbox tier — both Tekton tiers are required and run on push, and they are the tier the merge is gated on. Gate the **merged** tree before merging; `strict` is false, so a green check here is a claim about this branch only.

⚠ If `scripts/tests/test_subsystem_store_api.py` fails in CI it is a known flaky race (`TestLockoutOverHTTP` / `TestTrustedProxyOverHTTP` / `TestTheActorComesFromTheTOKEN`), not this change — nothing here touches it.

### DIFF
--- a/scripts/collector/keylog/espanso_detect.py
+++ b/scripts/collector/keylog/espanso_detect.py
@@ -21,8 +21,10 @@ Two detection paths:
      keylogger calls `feed_search_open`; subsequent chars accumulate as the
      search TERM (not the direct ring) until a close boundary (Enter, Escape,
      focus change, or idle). We fuzzy-attribute the term to a snippet; a unique
-     match is attributed, zero/multiple → trigger=None, but the search-open +
-     term are recorded regardless. Search events are always `inferred=True`.
+     match is attributed, and when several match, a term that NAMES exactly one
+     of them outright (it IS that trigger, with or without the ':') takes it.
+     Otherwise → trigger=None, but the search-open + term are recorded
+     regardless. Search events are always `inferred=True`.
 
 Honest limitation: "ring ends with a trigger" ≈ "espanso fired", EXCEPT in
 per-app espanso-disabled contexts (still far better than phrase-counting). The
@@ -250,7 +252,26 @@ class EspansoDetector:
         if not t:
             return None
         matched = [trig for trig in self.ts.triggers if self._term_matches(t, trig)]
-        return matched[0] if len(matched) == 1 else None
+        if len(matched) == 1:
+            return matched[0]
+        # Matching is by SUBSTRING (see `_token_matches`), so a term can match a
+        # snippet it merely occurs INSIDE. When ":acq" was split into ":dacq" +
+        # ":acq" on 2026-08-28 the bare term "acq" started matching BOTH
+        # (`"acq" in ":dacq"` is True) and every such fire was recorded
+        # UNATTRIBUTED. A term that NAMES one snippet outright — it equals that
+        # trigger, with or without the leading ":" — is not ambiguous about
+        # which snippet it means, so that snippet wins over the ones that merely
+        # contain it. This is consulted ONLY on the already-ambiguous branch, so
+        # it can never re-point a term that resolves uniquely today; the tie is
+        # broken only when EXACTLY ONE candidate is named outright.
+        exact = [trig for trig in matched if self._names_trigger(t, trig)]
+        return exact[0] if len(exact) == 1 else None
+
+    @staticmethod
+    def _names_trigger(term, trig):
+        """True when `term` IS this trigger, spelled with or without its ':'."""
+        low = (trig or "").lower()
+        return bool(low) and term in (low, low[1:] if low.startswith(":") else low)
 
     def _term_matches(self, term, trig):
         """True when EVERY whitespace-separated token of `term` matches `trig`.
--- a/scripts/collector/keylog/tests/test_espanso_detect.py
+++ b/scripts/collector/keylog/tests/test_espanso_detect.py
@@ -139,14 +139,17 @@ def test_search_fuzzy_zero_match_still_emits_with_term():
 
 
 def test_search_fuzzy_multiple_match_is_ambiguous():
-    # "date" is a substring of BOTH :date and :datetime → ambiguous → trigger None.
+    # "dat" is a substring of BOTH :date and :datetime → ambiguous → trigger None.
+    # It deliberately NAMES neither: the 2026-08-28 exact-name tie-break resolves
+    # a term that IS one of the candidate triggers (so bare "date" now lands on
+    # :date), and this test is about the ambiguity path, not about that.
     d = _det()
     d.feed_search_open(app=APP, session=SESS, now=0.0)
-    _type(d, "date", now=1.0)
+    _type(d, "dat", now=1.0)
     evs = list(d.feed_char("\n", app=APP, session=SESS, now=5.0))
     assert len(evs) == 1
     assert evs[0].trigger is None
-    assert evs[0].search_term == "date"
+    assert evs[0].search_term == "dat"
 
 
 def test_search_flush_on_idle_without_enter():
@@ -269,15 +272,16 @@ def test_focus_steal_by_espanso_window_keeps_search_and_attributes():
 
 
 def test_focus_steal_ambiguous_term_still_emits_trigger_none():
-    # Same focus-steal path, but an ambiguous term ("date" ⊂ :date and :datetime)
+    # Same focus-steal path, but an ambiguous term ("dat" ⊂ :date and :datetime,
+    # and NAMING neither — see the exact-name note on the ambiguity test above)
     # → trigger=None yet the event is still emitted (real search, unattributed).
     d = _det()
     d.feed_search_open(app="Alacritty", session="win-term", now=0.0)
-    _type(d, "date", app=ESPANSO_WIN, session="win-esp", now=1.0)
+    _type(d, "dat", app=ESPANSO_WIN, session="win-esp", now=1.0)
     out = list(d.feed_char("\n", app=ESPANSO_WIN, session="win-esp", now=6.0))
     assert len(out) == 1
     assert out[0].trigger is None
-    assert out[0].search_term == "date"
+    assert out[0].search_term == "dat"
     assert out[0].method == "search"
     assert out[0].app == "Alacritty"  # origin window preserved
 
@@ -449,7 +453,10 @@ def test_multiword_all_tokens_required():
 
 
 def test_single_word_term_behaviour_unchanged():
-    # The single-token path must be byte-for-byte the old semantics.
+    # The single-token MATCHING path is byte-for-byte the old semantics (the
+    # 2026-08-05 multi-word fix changed nothing here). Attribution has since
+    # gained one tie-break on the AMBIGUOUS branch only — pinned explicitly
+    # below, alongside a term that names nothing and stays ambiguous.
     d = _det_for(CIVIT_BASE)
     assert d._term_matches("prod", ":cpk") is True
     assert d._term_matches("prod", ":cc") is False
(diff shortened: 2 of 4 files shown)


ZacxDev:
Deployed-state check (workbench, read-only): `keylog.service` is `active/running` and its `ExecStart` reads `/home/zach/.config/activity-collector/keylog/keylog.py`, whose `espanso_detect.py` resolves via `readlink -f` to `/nix/store/sksc639zv93sqly4qjrgjbviar7jbxk7-hm_keylog/espanso_detect.py`. That file still carries the unfixed line at 253:

```
253:        return matched[0] if len(matched) == 1 else None
```

`_names_trigger` occurs 0 times in it. So the defect is confirmed live in the running collector on this host, not merely present on `main`. The laptop was not checked. This is a `home.file` copy, so after merge the sequence is **merge -> pull -> `home-manager switch` (or `scripts/ship.sh`) -> `systemctl --user restart keylog.service`** — a `git pull` alone changes nothing the collector reads.

ZacxDev:
CAPTURED: the `tekton/devrc-pytests` failure output, before re-triggering.

Run `devrc-ci-7wwng` (pipelinerun `tekton-ci/devrc-ci-7wwng`, revision `8b365f3afeec`, started `2026-08-29T01:40:18Z`). Pod `devrc-ci-7wwng-gate-pod` was still present, so `kubectl logs … -c step-pytests` still had it.

```
=================================== FAILURES ===================================
_ TestTheBackstopSurvivesITSOWNLogSink.test_a_RAISING_AUDIT_SINK_AND_a_RAISING_print_exc_still_end_the_request _
[gw2] linux -- Python 3.12.14

        monkeypatch.undo()
        captured = capfd.readouterr()
        # 🔴 THE WRITE LANDED. Without this every assertion below is satisfied by
        # a server that refused the append and therefore had nothing to audit.
        assert BULLET_SINKLESS in nuance_of(path), path.read_text()
        answers = _responses(raw)
>       assert len(answers) == 1, f"a second response followed the 200: {raw!r}"
E       AssertionError: a second response followed the 200: b''
E       assert 0 == 1
E        +  where 0 = len([])
scripts/tests/test_subsystem_store_api.py:12653: AssertionError

1 failed, 9788 passed, 4 warnings in 442.58s (0:07:22)
```

**Read the numbers, not the message.** The assertion message says *"a second response followed the 200"*, but `len(answers)` is **0**, not 2, and `raw` is **`b''` — empty**. The client read **nothing at all** off the socket. The message is written for the `>= 2` direction only, so when the failure arrives from the `== 0` direction it prints a sentence describing the opposite of what happened. That is very likely why four occurrences today have gone undiagnosed: the message sends the reader looking for a double-response bug that is not there.

What the log rules out: the assertion immediately above it — `assert BULLET_SINKLESS in nuance_of(path)` — **passed**. The append landed and the server did handle the request. So the server side worked; only the client-side read came back empty. That points at `_raw_exchange` / socket timing under `-n` parallelism (this ran on `gw2`), not at the backstop logic the test is named for.

Suggested follow-up for whoever owns that file (NOT done here, out of scope for this PR): split the assertion so `len(answers) == 0` and `len(answers) >= 2` report different messages, and have the `0` branch print whether the socket saw EOF (`saw_eof` is already bound in the test but never asserted on). That alone converts the next occurrence into a diagnosis.

Full-run numbers from the same log, for the record — this is the nix **sandbox** tier that the merge is gated on:

```
TOTAL collected=18285  passed=18282  skipped=2  failed=1  (floor: 15970 = sum of 28 per-target floors)
  FAIL  scripts/tests                        (collected=9789 passed=9788 skipped=0 failed=1 errors=0)
  PASS  scripts/collector/keylog/tests       (collected=93  passed=93  skipped=0 floor=79)
  PASS  scripts/session-analysis/tests       (collected=491 passed=491 skipped=0 floor=440)
```

Exactly one failure in the whole 18285-test run, in a file this PR does not touch. Both targets this PR does touch passed on the gating tier at their full counts (93 and 491), matching the dev-host numbers reported above.

ZacxDev:
Local evidence that `test_subsystem_store_api.py` is a flake, from the identical tree.

I intended to skip that file locally and passed `--deselect` an **absolute** path. That is inert — pytest matches `--deselect` against nodeids relative to the invocation directory, so an absolute path matches nothing and pytest does not error. Measured, with a positive control:

```
A: no deselect                          9789 tests collected
B: --deselect /abs/.../test_subsystem_store_api.py    9789 tests collected   <- inert
C: --deselect scripts/tests/test_subsystem_store_api.py
                                        9187/9789 tests collected (602 deselected)
```

So the flaky file **did** run locally, all 602 of its tests, on tree `d5d695fa` — the same tree the sandbox failed on:

```
9788 passed, 1 skipped, 3 warnings in 1193.29s (0:19:53)
```

Zero failures. `TestTheBackstopSurvivesITSOWNLogSink.test_a_RAISING_AUDIT_SINK_AND_a_RAISING_print_exc_still_end_the_request` passed on the dev host and failed in the nix sandbox on the same tree, in the same run window. Combined with the captured `raw == b''` / `len(answers) == 0`, that is consistent with a client-side socket read race under xdist (`gw2`) and not with anything in this diff.

Two honest caveats:
- Local totals `9788 passed + 1 skipped = 9789`; the sandbox reported `9788 passed + 1 failed = 9789`, `skipped=0`. The one skip is environment-dependent and I have **not** identified which test it is — it is not a failure, but I am not claiming to know what it is.
- The dev-host tier and the `nix build .#checks…` sandbox tier are different environments; a pass on one is not a pass on the other. That is the whole point here — the difference is what makes the flake visible.

ZacxDev:
Re-trigger result: **both checks green on a byte-identical tree.**

| | commit | tree |
|---|---|---|
| failing run `devrc-ci-7wwng` | `8b365f3a` | `d5d695fa7e5367f248c63f0f6dd37a9f9baac603` |
| passing run `devrc-ci-zwgvg` | `08032995` | `d5d695fa7e5367f248c63f0f6dd37a9f9baac603` |

`git commit --amend --no-edit` + `push --force-with-lease`, no empty commit. Tree OID verified identical on four reads: local before the amend, local after, local after the push, and `origin/fix/espanso-acq-attribution`.

```
tekton/devrc-nodetests  pass  TOTAL suites=5 files=39 tests=1300 pass=1300 fail=0 skipped=0  (floor: 1239)
tekton/devrc-pytests    pass  TOTAL collected=18285  passed=18283  skipped=2  failed=0  (floor: 15970 = sum of 28 per-target floors)
```

Same 18285 collected as the failing run; `passed` moved 18282 -> 18283 and `failed` 1 -> 0, exactly the one test. `mergeStateStatus: CLEAN`.

Identical input, opposite verdict — a flake by construction, and the fifth occurrence in `test_subsystem_store_api.py` today. The captured output is in the comment above; unlike the previous four, this one was read before it was superseded.

⚠ The green is a claim about this branch at base `6349a8b9`, not about the tree a merge creates (`strict` is false on this repo). Gate the merged tree before merging.

PR #989, opened 2026-08-28
feat(session-manager,find-session): answer "is it still running, which window" LIVE — the 30s archive was the wrong instrument

ZacxDev:
## The problem, measured

"Find this thing I lost track of — is it still running, which tmux window, where did it leave off?" took two tools and a human gluing them together. An observed run cost **127 s, $0.0056 and 9 tool calls, 5 of them pure flailing**, and its final answer sent the operator to a **real but wrong** live window.

Measured on the workbench, 2026-08-28:

| | |
|---|---|
| `find-session.py <terms>` (transcript archive walk) | **30.1 s** |
| `session-manager --json --lean --no-ch` (live fleet) | **1.82 s**, 72 rows, `task` populated on 66 |
| one query substring matched against **`task`** | **1** row (the right one) |
| one query substring matched against **`path`** | **29** rows |

So the 30-second archive is the wrong instrument for a question about *now*, and `path` does not discriminate — 29 of 72 windows share a repo path.

Re-confirmed live on this branch: `--match devrc` → **1** matched / 71 excluded; `--match devrc --match-path` → **10** matched / 62 excluded.

### 1. `--match` on `session-manager` (`scan`/`list`/`detail`)
- Case-insensitive substring over **`task`, `label`, `codename`**. Repeatable; terms are **ANDed** (matching `find-session.py`'s default), and one term may satisfy itself in a different field from another.
- `--match-path` adds `path`. **Off by default**; the 1-vs-29 measurement is in the flag's help text and beside `MATCH_FIELDS`.
- Filtered rows recompute **`summary` and `caveats`** the way `filter_report` does — a `--match` that removes every row now publishes `kinds_produced: []`, not a stale `["tmux"]`.
- `filters.match` / `.match_fields` / `.matched` / `.excluded_by_match`, all **`null` never `0`** when no filter ran. So "no rows matched these words in these fields" can never be read as "nothing exists", and a reader can see that `path` was excluded. **`summary` mirrors `match` / `match_fields` / `excluded_by_match` / `matched`.**

> ⚠ **CORRECTION (2026-08-29).** An earlier revision of this body said `filters.matched` was "mirrored into `summary`". **That was false at `a6f09d5a`** — `summary.matched` did not exist; the audit confirmed it live and `payload-contract.md` was right where this body was wrong. It is true as of `9f9dcbde`, which adds the mirror as part of audit finding 6 (the table's filter line was quoting `total_sessions`). Flagged rather than silently edited, because a reader may have believed the earlier version.
[...]

PR #1000, opened 2026-08-29
feat(telemetry,find-session): skill usage was unmeasured on Claude Code — 120,366 records carried the answer and nothing read them

ZacxDev:
## Why

An opencode session was asked **"identify signal skill usage"**. It spent ~34 minutes and 8 turns, and concluded — twice, in bold — *"the skill has never been invoked operationally."* That was the reverse of the truth, and the operator had to steer it four times to get there.

The answer was one bounded query away the whole time:

```sql
source='claude' AND kind='command' AND startsWith(text,'/signal')   -->   laptop 1
```

But that query only sees **typed** invocations, and typing is the minority case. Measured 2026-08-29 over the workbench corpus, counting sessions (subagent transcripts excluded):

| skill | attributed | typed `/name` | attributed-only | typed-only |
|---|---|---|---|---|
| `browser` | 50 | 0 | **50** | 0 |
| `clawgate` | 70 | 32 | 42 | **4** |
| `activity` | 8 | 3 | 5 | 0 |

Most skill usage **auto-fires from the description**. The only field that sees that is Claude Code's `attributionSkill`, which sits on **120,366 records** under `~/.claude/projects` and had **zero** references anywhere in `scripts/` or `claude/`.

Meanwhile opencode skill usage was *already* queryable (a skill there **is** a tool call — 118 calls / 97 sessions / 30d, both hosts). So the fleet was half-instrumented and nothing said which half.

## What this does

**`session-tailer.py`** — two maps on the Layer A rollup, collected inside the loop that already visits those records and already builds `tool_counts`. No new daemon, no new deploy surface.

- `skills_used` — `{skill: attributed records}`. Sees auto-fire.
- `commands_typed` — `{command: times typed}`. Named for what it holds: it includes **built-ins** (`/login`, `/clear`), which are not skills.

🔴 **Two fields, not one.** Both directions have live instances (`clawgate` typed-only = 4), so collapsing them loses real usage. Two tests fail if someone tries.

Only the command **name** is kept, never its args — args are operator free-text and this payload ships to ClickHouse for every session on both hosts. Pinned against the emitted **event**, not just the dict, since `build_event` json-dumps the whole rollup.

**`find-session.py --skill NAME`** — exact match on the attributed identity, ORing both invocation routes, ANDing with any terms, usable with no terms at all.

Live corpus: `--skill signal` → **1** session. The keyword form returns **666**, nearly all `scripts/signal/tests/…` in test output.

## Verification

**Both tiers, base `c7d70a40`** — they are different tiers, not two spellings of one, so both were run:

| tier | command | result |
|---|---|---|
| dev-host | `scripts/gate.sh --tier both --set all` | `GATE: RESULT=PASS exit=0` (pytest + node, verdict lines agree) |
| nix sandbox (what Tekton gates on) | `nix build .#checks.x86_64-linux.{pytests,nodetests}` | `devrc-pytests RESULT: PASS` — collected=18306 passed=18304 skipped=2 failed=0, floor 15970 · `devrc-nodetests RESULT: PASS` |

**The new tests provably ran in the gate**, which the floor numbers alone would not show — collected-count delta against base: `scripts/collector/claude/tests` 137 → 151 (+14), `scripts/tests` 9789 → 9805 (+16).

- **New tests red at `c7d70a40`** (14/14, `KeyError`), **green at HEAD**.
- **Counts cross-checked against an independent grep baseline at two points**: `activity` 8 = 8, `browser` 50 = 50. The initial 3× disagreement was subagent transcripts, which this corpus excludes by existing design — now documented as a coverage limit rather than left to be rediscovered.
- **Mutation battery — 6 mutants, each killed by exactly the test written for it:**

| mutation | killed by |
|---|---|
| read `attributionSkill` from `msg` instead of the record | 3 (positive control + auto-fire) |
| count each skill once instead of per record | 1 (counting) |
| keep whole command text incl. args | 3 (both leak guards) |
| substring instead of exact identity | 2 (prefix + longer-name) — *and only those* |
| skill predicate disabled | 6 (every exclusion test) |
| drop the wider-surface narrowing | 1 (tool-output guard) |

A test caught a real defect while being written: a whitespace-only `--skill` was truthy, slipped the empty-query guard, failed the predicate for every session and returned an **empty corpus-wide result instead of raising** — the exact silent zero this change exists to remove, one line above it.

## Two coverage limits, both deliberate and both documented

- Counts **sessions**: a skill used only inside a dispatched **subagent** is not counted (`activity` had 15 such transcripts against 8 sessions). Matches the boundary `find-session` and `session-tailer` already had.
- The **opencode corpus has no per-record attribution**, so that leg is **skipped** under `--skill` and the omission is **printed**. Running it unfiltered would quietly answer a different question; skipping it silently would hand back a partial count that reads as fleet-wide. Both are the failure this flag exists to end.

## 🔴 Conflicts with #989 — and one semantic interaction

`merge-tree` exit **1**: `scripts/find-session.py` and `claude/skills/find-session/SKILL.md` both conflict. `scripts/lib/transcript_search.py` merges clean — the implementation was deliberately put there for that reason.

Beyond the textual conflict, whoever resolves must decide one thing: **`--skill` and `--live` answer different questions and must not silently compose.** #989 makes `--live` short-circuit the archive walk, and `--skill` only exists in the archive path — so a resolution that lets `--live` win would make `--skill` return nothing. `--skill` should force the archive path (or error when combined). A clean git merge will *not* surface this.

## Deliberately NOT included

**The `adoption-scan` `via: "skill"` registry arm.** `skills_used` is **forward-only** — no row carries it until a `home-manager switch` lands the tailer and the 5-minute timer runs. `adoption-scan` raises a loud `⚠ DEAD` flag at zero uses, and its own SKILL.md warns against a registry row with no emitter behind it because "a zero here is indistinguishable from wired-to-nothing". Shipping the reader now would confidently report **every skill as DEAD**.

Precondition and the query to check it are in `claudedocs/proposal-skill-usage-telemetry.md`.

Also still open, from the same session: **the ClickHouse reader password should be rotated** — it was printed in cleartext by a debugging `2>&1` and is persisted in `opencode-stable.db`.

## Deploy

Merged ≠ deployed. Needs `scripts/ship.sh`; the tailer is restarted by the switch (`X-Restart-Triggers`), and rows appear on the next 5-minute timer tick.


ZacxDev:
## 🔴 Correction to this PR's description — two numbers in it are wrong

The PR body above states, in bold, *"Both directions have live instances (`clawgate` typed-only = 4)"*, and its table gives `clawgate 70 attributed / 32 typed / 42 attributed-only / 4 typed-only`.

**Both the 32 and the 4 are wrong.** They came from a raw file grep, which matches `<command-name>` anywhere in a transcript — including inside quoted **tool output**. That is precisely the false positive this PR ships a guard against (`test_a_command_quoted_in_TOOL_OUTPUT_is_not_an_invocation`), and one of the four "typed" sessions was the session authoring this PR, grepping transcripts.

Re-measured with the counting rule the shipped code implements:

| skill | attributed | typed | attributed-only | typed-only |
|---|---|---|---|---|
| `browser` | 50 | 0 | 50 | 0 |
| `clawgate` | 72 | 28 | 44 | **0** |
| `activity` | 8 | 3 | 5 | 0 |
| `handoff` | 383 | 346 | 40 | **3** |

The two-field design still holds — `handoff` carries 3 genuine typed-only sessions — but it is `handoff`, not `clawgate`, that evidences it. Corrected in the source comment, the test docstring and the proposal doc, with the retraction inline so the number is not re-derived by grep.

Correcting here rather than silently editing the body, since a reviewer may already have read the wrong version.

---

## Round 1 → fixes landed

Audit round 1 returned 7 findings (1 🔴, 6 🟡) plus 4 nits. Fixes are in `ef7b3ea4`.

Beyond F1 above, the two most substantive:

- **A third invocation route was unread.** Claude Code also records an explicit `Skill` tool_use carrying `input.skill` — 1,305 blocks in the corpus. Missing it undercounted by up to 87.5% on one skill. Now read by both sides: `next-lever` 1→8, `check-app` 14→19, `resume` 100→103. `input.args` is never kept (operator free-text; live examples carry account ids).
- **The "only the command NAME is kept" guarantee was narrower than its sentence.** An empty or whitespace-only `<command-name>` made `classify()` return the *args*, whose first word became a shipped payload key. The first fix — a bounded charset — was **wrong**, and the new regression tests caught it: the leaked word is itself charset-valid. Parsing the name tag is what distinguishes "no name" from "a name".

Deferred with its precondition (recorded in the proposal doc): a deadman for `attributionSkill` itself. It is an undocumented upstream field, so a rename makes every count silently zero — but the invariant would be permanently red until rows accumulate, so it ships alongside the `adoption-scan` reader arm.

**Verification:** 215 tests green; 5 new mutants each killed by exactly its own tests; gate PASS on both tiers (dev-host `GATE: RESULT=PASS exit=0`, and the nix sandbox tier).

**Ledger:** round 1 · payload lines changed THIS round: 220 (since round 1: 220) · tests: 122.

```audit-claims round=1 audited=cd343c0d..ef7b3ea4
1. F1 — the clawgate typed-only=4 figure was retracted and replaced with re-measured numbers (clawgate typed-only=0, handoff typed-only=3) in the source comment, the test docstring, the proposal doc and a public PR comment.
2. F5 — a third invocation route (`Skill` tool_use `input.skill`) is now read by both session-tailer.py (`skills_invoked`) and transcript_search.py, and its `input.args` is not kept.
3. F2 — the command name is now parsed from the `<command-name>` tag rather than from classify()'s merged name+args text, and is bounded by SKILL_NAME_RE with rejects counted in `unusable_skill_names`.
4. F3 — a single `normalize_skill()` is used by search()'s guard, the predicate and the CLI, so `--skill /` and `--skill "   "` exit 2 instead of returning a corpus-wide empty result or a traceback.
5. F4 — the "opencode NOT searched" disclosure is emitted before the `--json` early return and goes to stderr, leaving stdout valid JSON.
6. F7 — the live-credential paragraph was removed from the committed proposal doc and replaced with a pointer.
7. F11 — the SKILL.md no longer attributes the 666 keyword hits to tool output, and its limits list now enumerates three, including that a typed built-in matches.
8. F6 — deferred, not fixed: no deadman asserts `skills_used` is non-empty; recorded in the proposal doc with its precondition.
9. F8/F9/F10 — not changed: `commands_typed` remains implemented in two places, `session_used_skill` still ORs built-ins (now documented), and `skills_used` still counts records rather than uses.
```

ZacxDev:
## Round 2 (blind delta) → fixes landed in `f7664040`

Round 2's most useful result was that **round 1's claim #1 was false**: I stated the test docstring had been corrected, and it had not. The delta checked the claim against the diff rather than believing it.

Three defects were the *same* one round 1 fixed, surviving where I hadn't looked — including the retracted `clawgate typed-only = 4` still asserted **seven lines below its own retraction box**. And one was a regression the round-1 fix introduced: `SKILL_NAME_RE` could not express a namespace-qualified skill identity (`plugin:skill`, `apps/web:deploy` — 14 plugin skills are cached on this host), so the tailer rejected what the search accepted and a plugin skill would read as *never used* in ClickHouse while `--skill` found the session. A silent zero of exactly the class this PR removes, reintroduced inside the guard protecting it.

Also corrected: my round-1 wording *"5 mutants each killed by exactly its own tests"* was too strong — several mutants are killed by multiple tests. The accurate claim is that every mutant was killed.

**Verification:** 475 tests green including the shared `check-clickup-addressed` consumer; gate **PASS both tiers** (`scripts/tests` 9809→9820, `scripts/collector/claude/tests` 160→165).

**Ledger:** round 2 · payload lines changed THIS round: 82 (since round 1: 302) · scaffolding: 238.

```audit-claims round=2 audited=ef7b3ea4..f7664040
1. The retracted `clawgate typed-only = 4` no longer appears in the proposal doc or the test class docstring; both carry re-measured numbers with the retraction inline, and the "measured 8 times" figure is gone.
2. `test_a_TYPED_command_is_seen_with_no_attributed_record` is relabelled an INVARIANT GUARD and names `handoff` (3) as the live typed-only instance instead of clawgate.
3. `transcript_search.py`'s scan comment says THREE signals rather than two.
4. The proposal no longer attributes the 666 keyword hits to tool output, and gives 564 as the Claude-only figure.
5. `SKILL_NAME_PATTERN` admits `:` and `/`, so `plugin:skill` and `apps/web:deploy` are recorded rather than rejected as unusable.
6. The same bound is applied in `transcript_search.py`, and the two copies are pinned byte-identical by `test_the_two_skill_name_bounds_agree`; the duplication is deliberate because nix deploys `scripts/collector/claude` alone to the daemon's runtime path.
7. The command-name latch takes the first NON-EMPTY `<command-name>`, so an empty tag in an earlier block no longer discards a real command in a later block.
8. A command turn whose only name tag is empty increments `unusable_skill_names` at the call site, closing the case `_note_name` could not observe.
9. `scripts/tests/test_find_session_skill_cli.py` (11 tests) drives `find-session.py:main()` and covers the disclosure's position and stream, the normalised empty-query guard, and that real and slash-prefixed skill names are still accepted.
10. F6 remains deferred and unfixed; F9/F10 remain unchanged as recorded in round 1.
```

ZacxDev:
## Round 3 (blind delta) → fixes landed in `13a1f629` + `5a68937e`

Round 3 was **not clean**: 4 🟡 and 6 🟢. The two that mattered were both cases where a round-2 claim was wider than the code that landed.

- **A mutant making the search-side bound inert survived 681 tests.** `SKILL_NAME_RE = re.compile(r".*")` — declaring the right pattern while compiling an inert one — passed eight suites. The bound had no behavioural coverage at all.
- **"The two readers AGREE about what a name is" was false.** It guarded the typed route only; attribution and explicit invocation were still unbounded, so `--skill "not a valid name"` matched a session ClickHouse reported as having used no such skill. A bound on one of three routes is not a bound.
- **My round-2 widening was wrong in both directions.** Admitting `/` for `apps/web:deploy` let filesystem paths become public ClickHouse map keys, and it *still* rejected all 10 real namespace-qualified identities on this fleet (`.claude/worktrees/agent-<17 hex>:remix` — rejected on the leading dot). The doc's illustrative example passed while every real value failed.

### And one finding against my own fix, from my own mutation battery

Reverting `finditer`→`search` killed **no test**. Measured: for two tags in one block, `classify()` runs its own `search`, hits the empty tag first and returns `None`, so the counting branch never executes — `commands_typed={}` before and after. **The round-3 commit message's "G6 fixed" claim is retracted** in `5a68937e`; `finditer` is defensive only, the limitation is pinned by a test, and the real fix is upstream in `classify()`, shared with the message-stream tailer (changing it moves `kind=command` for every row that source has emitted).

**Verification:** 489 tests green including the shared `check-clickup-addressed` consumer; gate **PASS both tiers** (`scripts/tests` 9820→9830, `scripts/collector/claude/tests` 165→169). Mutants killed: inert search bound (5 tests), keep-path-prefix (2), drop the given-but-empty `--skill` refusal (3). One equivalent mutant recorded so it is not re-derived.

**Ledger:** round 3 · payload lines changed THIS round: 168 (since round 1: 470) · scaffolding: 181.

```audit-claims round=3 audited=f7664040..5a68937e
1. `canonical_skill_name()` is the single rule for all three routes in both readers; `skills_attributed`, `skills_invoked` and `commands_typed` all go through it.
2. `SKILL_NAME_PATTERN` no longer admits `/`, so filesystem paths cannot become payload keys.
3. A path-derived prefix is DROPPED rather than rejected: `<path>:<skill>` records `<skill>`, so `.claude/worktrees/agent-<hex>:remix` records `remix` and `apps/web:deploy` records `deploy`.
4. A bare path with no skill after it is rejected AND counted in `unusable_skill_names`.
5. The search-side bound now has behavioural tests on all three routes, and a test asserts the two readers' `canonical_skill_name` agree on eight values.
6. A `--skill` that was given but normalises to empty is refused with exit 2 even when search terms are present.
7. `test_a_TYPED_command_is_seen_with_no_attributed_record` uses a `handoff` fixture (the skill with live typed-only instances) instead of `clawgate` (0).
8. The typed-only claim is restated as 11 names, six of them built-ins, five actual skills.
9. `unusable_skill_names`' payload doc now also covers the no-name command turn; the test module header says three fields; the "above" cross-reference is gone.
10. RETRACTED, not fixed: the within-block two-tag case is NOT counted; `finditer` is inert for it because `classify()` gates it upstream. Pinned by `test_two_tags_in_one_block_are_NOT_counted`.
11. Still deferred and unfixed: F6 (no deadman for `attributionSkill`), and F9/F10 from round 1.
```

ZacxDev:
## Round 4 (blind delta) → fixes landed in `381d3b1e`, `965d0825`, `8019e4f0`

Round 4 found six 🟡. The one that mattered was a **regression round 3 introduced**: the path-strip examines only the tail after the last `:`, so a value whose *head* was prose or kilobytes of junk got rewritten into a clean key — `"not a valid name/x:handoff"` recorded as `{'handoff': 1}`, misattributed to a **real** skill with `unusable_skill_names` showing nothing wrong. Before the strip existed all such inputs were rejected *and counted*. Round 3 turned "malformed, audited" into "silently wrong, audit trail clean", and `input.skill` is model-written free text. Whitespace and a 256-char bound are now checked **before** the strip — the ordering is the fix.

### My own mutation battery then found what the audit had not

Reverting one argparse default (`--skill default=None` → `""`) makes an *omitted* flag indistinguishable from an empty one, so the refusal added this round fires on **every plain keyword search**: `find-session signal` exits 2 with zero output, the tool's primary mode dead. **The entire CLI suite stayed green.** Every test in it passed `--skill`; the one that didn't asserted only that a disclosure was absent, never the exit code. The suite could not tell "rejects bad input" from "rejects everything". Two negative controls added.

That is the second time this ladder has found a guard measured only by what it must reject and never by what it must let through.

### And the gate went red — on my own scaffolding

`test_no_public_ips.py` flagged a colon-axis fuzz value I added: short hex-letter operands either side of a double colon parse as a routable IPv6 literal, in a **public** repo. The first fix was incomplete because my explanatory comment contained the same shape and the scanner reads comments too. Fixed in `8019e4f0`.

### 🔴 `main` is currently RED, and it is not this PR

Gating the **merged** tree (this branch merged onto `07890ebc`) fails two tests. I ran the control — both reproduce on `origin/main` **alone**, with none of this PR's content, and both come from commits that are in `main` but **not** in this branch:

| test | cause |
|---|---|
| `test_opencode_engine.py::test_engine_is_the_version_every_measurement_is_keyed_to` | `be1585c3` (#1013) bumped the flake to opencode **1.18.21** while `scripts/opencode/opencode.jsonc` still reads "measured on v1.18.18" |
| `test_espanso_detect.py::test_live_existing_resolutions_not_made_ambiguous` | `22e2830a` (#999) interacting with `8bab93b1` (#979)'s `:acq`/`:dacq` split — `ask` now resolves ambiguously across `:acq` and `:dacq` |

This branch's own gate is green on both. Raised here because a red `main` makes the required `tekton/devrc-pytests` check fail for every PR branched off it — a merge-blocking condition for the repo, not for this change.

**Verification:** 232 tests green on the touched suites; `scripts/collector/claude/tests` **170/170** on the merged tree; node tier green throughout. Mutants killed this round: `finditer`→`search` (previously survived), `split`→`rsplit` in both copies and in one copy only (both previously survived), dropping the pre-strip whitespace/length check, and reverting the `--skill` default.

**Ledger:** round 4 · payload lines changed THIS round: ~119 (since round 1: ~589) · scaffolding: ~169.

```audit-claims round=4 audited=5a68937e..8019e4f0
1. Whitespace and a 256-char identity bound are checked BEFORE the path-strip, so a value whose head is prose or a multi-kilobyte blob is rejected instead of being rewritten into a real skill's key.
2. `canonical_skill_name` rejects a non-string identity, so the emitter and the search agree on `attributionSkill = 42`.
3. The path-strip uses `split(":", 1)` rather than `rsplit`, so `apps/web:cloudflare:wrangler` keeps its plugin namespace as `cloudflare:wrangler`.
4. `--skill` defaults to None and the guard tests `is not None`, so an explicitly-empty `--skill ""` is refused rather than silently dropping the filter.
5. The query is canonicalised with the same `canonical_skill_name` as the corpus, so `--skill apps/web:deploy` matches the recorded `deploy`.
6. `session_used_skill` and `search()` both canonicalise instead of using `normalize_skill`.
7. The agreement test fuzzes generated cases across slash count, colon count, leading dot, whitespace, length and type instead of eight hand-picked values.
8. The `finditer` retraction is corrected: it IS load-bearing when the turn carries `<command-args>`, and a test now pins that; the no-args half remains gated by `classify()` and is pinned separately.
9. `_load_tailer()` restores `sys.path` and `sys.modules` in a `finally`.
10. Two negative controls assert a plain keyword search with no `--skill` still exits 0 and returns results, under both text and `--json`.
11. The "10 distinct worktree:remix" figure is retracted as not reproducible; the decision it justified stands on the shape rather than the count.
12. A colon-axis fuzz value no longer spells an IPv6-shaped literal, in the values or in the comment.
13. Still deferred and unfixed: F6 (no deadman for `attributionSkill`), F9/F10 from round 1.
```

Issue #975, opened 2026-08-28
worktree-prune: a non-ASCII filename makes content-identical fire falsely — unlanded work classified dead and removable, under DEFAULT git config


ZacxDev:
## What happens

`scripts/worktree-prune` can classify a worktree with **unlanded work** as `dead` and mark it `removable`, under **default** git configuration, when the branch's differing paths are all non-ASCII-named.

It needs only a plain accented filename — `café.md`. No undecodable byte, no submodule, no unusual setting.

## Mechanism (measured, with both controls)

`_paths_differ` (`scripts/worktree-prune:731`) asks `git diff --name-only <base> <head>` for the changed set, then feeds each returned path **back to git as a pathspec** to see whether it still differs upstream.

Under default `core.quotePath=true`, git *renders* a non-ASCII path as a quoted C-escaped literal, and that literal does not match the file it names:

```
$ git diff --name-only trunk feature
"caf\303\251.md"                       <- 15-char literal, not the filename

$ git diff --name-only trunk feature -- '"caf\303\251.md"'
rc=0   output=[]                       <- genuine empty, rc 0, NOT an error
```

Empty ⇒ "no differing paths" ⇒ `content-identical` fires ⇒ verdict `dead`, `removable=True`.

**ASCII control on the same fixture**, proving the pathspec round-trip works when the name is ASCII:

```
$ git diff --name-only trunk feature -- plain.md
rc=0   output=[plain.md]               <- found
```

End-to-end through the tool, ASCII vs non-ASCII on identical fixtures:

```
e2e-ascii    : verdict='cannot-tell'  removable=False
e2e-nonascii : verdict='dead'         removable=True   reason='landed — signals: content-identical'
```

🔴 **The empty result is rc 0.** It is indistinguishable from a real "these paths are identical upstream" answer, which is exactly why this reads as a confident verdict rather than an error.

**Trigger condition:** every path in the branch's changed set that actually differs upstream must be non-ASCII-named. A mixed set is saved by any one differing ASCII path — which is why this has not been noticed.

## Why it matters

A false `dead` on unlanded work is the one verdict this tool exists not to get wrong. Blast radius is bounded — `git worktree remove` leaves the branch and its commits intact, so what is destroyed is the working directory and any **uncommitted** changes in it — but the row also feeds `--confirm N`, so it inflates the operator's primary safety number.

## This is already documented — one function away, in the arm that does it right

`worktree-prune:672-680` describes the identical mechanism for `ls-files`:

[...]

Issue #944, opened 2026-08-27
worktree-prune: an undecodable worktree DIRECTORY NAME aborts the entire scan (pre-existing; #935 fixed only the ls-files arm)


ZacxDev:
## What happens

A worktree whose **directory name** contains a byte that is not valid UTF-8 takes the **entire** `worktree-prune` scan down: `UnicodeDecodeError`, a bare traceback, rc=1, and **zero classified rows**. Not the one bad row — the whole run, including the other 800+ worktrees on the box.

## Mechanism (measured, not inferred)

`_git` runs with `text=True`, i.e. strict UTF-8 decoding, and the decode happens inside `subprocess.run` before the caller sees anything. `git worktree list --porcelain` emits the worktree **path raw and unquoted** — `core.quotePath` does not apply to it:

```
b'worktree /tmp/.../wt-b\xe9d\nHEAD 46089d9...\nbranch refs/heads/topic\n'
```

so `list_worktrees()` raises. There is no `try:` anywhere between `scan_repo()` and `sys.exit(main())`, so nothing catches it.

Reproduced on a hermetic fixture — one repo, one worktree whose directory is `wt-b\xe9d`, no submodules involved:

```
porcelain emits raw path : b'worktree /tmp/.../wt-b\xe9d\n...'
decodes as utf-8         : NO — UnicodeDecodeError
scan_repo (origin/main)  : ABORTED UnicodeDecodeError
```

The same class also reaches `execute_removals()`, via its `list_worktrees()` re-check and via `res.stderr` on the `worktree remove` failure path — so it can abort **mid-execute**, after some removals have happened, without printing the `execute: removed= skipped= failed=` accounting line that says what was done.

## Why it is a defect

The tool's own `is_dir()` docstring states the standard this violates: *"that is a row we cannot tell about, not a reason to abandon the other 750."* Every other unreadable-row path in this tool degrades to `cannot-tell`. This one degrades to nothing at all.

## Scope — this is PRE-EXISTING, not a regression

Measured on both tips, same fixture, same result:

| tool | `scan_repo` |
|---|---|
| `origin/main` | ABORTED `UnicodeDecodeError` |
| `fix/931-submodule-not-removable` (#935) | ABORTED `UnicodeDecodeError` |

[...]

PR #1002, opened 2026-08-29
fix(worktree-prune): a non-ASCII filename makes `content-identical` fire falsely — unlanded work classified `dead` + `removable` under DEFAULT git config (#975)

ZacxDev:
Fixes the defect in issue #975: `worktree-prune`'s `content-identical` landing
signal fires falsely on a non-ASCII filename, classifying a worktree that holds
**unlanded work** as `dead` + `removable`, under **default** git config. The tool
deletes directories; this is the one verdict it exists not to get wrong.

## The mechanism

`landing_signals` listed the branch's changed paths with `git diff --name-only`
through `_git` (`text=True`). Under the DEFAULT `core.quotePath=true` git
**renders** a non-ASCII path as a quoted C-escaped literal, so `café.md` arrived
as the 15-character string `"caf\303\251.md"`. `_paths_differ` fed that literal
straight back to git as a **pathspec**; it matched nothing; git exited **0 with
empty output** — and an empty diff is exactly the observable that means
"identical upstream". `content-identical` fired, the verdict was `dead`, the row
was `removable`.

Measured on git 2.55.0, default config:

```
$ git -C r diff --name-only trunk feature
"caf\303\251.md"
$ git -C r diff --name-only feature trunk -- '"caf\303\251.md"'
   (empty)   rc=0          <- THE BUG: empty is rc 0, indistinguishable from "identical"
$ git -C r diff --name-only feature trunk -- 'café.md'
"caf\303\251.md"  rc=0     <- the RAW name as a pathspec answers correctly
$ git -C r diff --name-only -z trunk feature | xxd
00000000: 6361 66c3 a92e 6d64 00                   caf...md.
```

The trigger is narrow, which is why nothing caught it: **every** path in the
changed set that actually differs upstream must be non-ASCII. One differing
ASCII path answers correctly and saves the row.

🔴 **The identical mechanism was already documented and fixed one function away**
— `worktree_submodules`' `ls-files -s -z` index arm, #935, with a long comment
explaining that `-z` is load-bearing because `core.quotePath` would otherwise
render `vendôr` as `"vend\303\264r"`. The arm that decides `dead` never got the
same treatment. One rule, two places, wrong at one of them.

## The fix

Both halves of the signal now go through `_git_raw` with `-z` and read **bytes** —
`landing_signals`' changed-path listing and `_paths_differ`'s comparison. Only
the **listing** decodes (`os.fsdecode`, surrogateescape, so the str re-encodes to
the exact original bytes when `subprocess` builds the pathspec argv); the
comparison never decodes at all — it receives `str` pathspecs and only tests its
own output bytes for emptiness. `_paths_differ`'s `list[str]` signature is
unchanged; the round-trip was verified empirically, not assumed.

Both test that emptiness with `if p`, never `if p.strip()` — see the round-1
audit section below, which is where that half of the fix came from.

Two things the fix had to not break, each measured rather than reasoned about:

* **The blast radius of the FIX** — #935's round-2 lesson. `-z` stops the
  quoting, but quoting was also silently guaranteeing git's output was ASCII, so
  a strict decode then raises `UnicodeDecodeError` **inside `subprocess.run`**,
  with no `try:` between it and `sys.exit(main())`. Hence bytes at both call
  sites. `core.quotePath=false` is the same crash in the non-`-z` call and is a
  real host setting; measured to raise, and to be fixed by the change.
* **`PATHSPEC_BATCH_BYTES`** sized each pathspec with `len(p)` on a *str*, while
  the budget is named in bytes because argv is measured in bytes. A str
  undercounts a non-ASCII path by up to 4x, so a batch could overrun the very cap
  the constant exists to stay under — and the constant's own comment says a
  truncated argv "would show up as 'no differing paths', i.e. as a confident,
  wrong `dead`", the same failure this issue is about. Now `len(os.fsencode(p))`,
  with its own mutant.

## Red/green matrix

The regression test was watched RED against `origin/main`'s `worktree-prune` in
an **isolated copy** (`git show origin/main:scripts/worktree-prune` into a `cp`
of the tree, `.git` pointer removed) — never by swapping the live file, which
would have fired the battery's source-integrity check as an operator-caused
alarm.

| | pre-change (`origin/main`) | post-change (HEAD) |
|---|---|---|
| `e2e-nonascii` | `verdict='dead'` `removable=True` `signals=['content-identical']` | `verdict='orphan'` `removable=False` `signals=[]` |
| `e2e-ascii` (control) | `verdict='orphan'` `removable=False` `signals=[]` | `verdict='orphan'` `removable=False` `signals=[]` |
| `e2e-whitespace` (round 1) | `verdict='dead'` `removable=True` `signals=['content-identical']` | `verdict='orphan'` `removable=False` `signals=[]` |

All three rows go end to end through the tool (`scan_repo`); the first two are
twin fixtures differing **only** in the filename's alphabet. The whitespace row's
evidence line makes the defect legible on its own: pre-change
`all 1 changed path(s) have identical content` on a branch that changed **two**;
post-change `at least one of 2 changed path(s) differs`.

Under pytest, on `scripts/tests/test_worktree_prune.py`: **7 failed / 213 passed
at `origin/main`, 220 passed at HEAD.** The branch adds **12** tests
(`^def test_` 183 → 195). *(An earlier revision of this body said "4 of the 10
new tests"; both halves were wrong — see the round-1 audit section.)*

**Point 2 of the issue's closing condition — the ASCII control — in both its
readings.** A fix that merely stopped `content-identical` from ever firing would
pass point 1 and destroy the tool:

* an *unlanded* ASCII branch still reaches `orphan`, not a blanket
  `cannot-tell` (`test_an_ascii_only_branch_is_still_classified_correctly`, and
  the table above);
* a *genuinely landed* branch still reaches `dead` — for ASCII by the
  pre-existing `test_squash_merged_branch_is_dead_not_orphan`, and now for
  non-ASCII too by `test_content_identical_still_fires_for_a_landed_nonascii_path`,
  which pins that the fix did not break the round-trip in the *other* direction
  (every path reported as differing, retiring the signal silently).

## Suite

`nix develop . --command bash scripts/run-tests.sh .` on the branch, re-run after
the round-1 edits:
**28 targets, `TOTAL collected=18288 passed=18286 skipped=2 failed=0`,
`RESULT: PASS (exit=0)`.** Counted from the runner's own **28** per-target
`N passed` summary lines, not from an exit code; grepped for a timeout panic and
`INTERNALERROR` (none). The box was at load average ~52/24 cores during these
runs, so wall times are inflated across the board (slowest target 161s) — that is
uniform inflation, not one test's assertion, which is the discriminator that
matters.

## Mutation battery

Re-run in full after the round-1 edits, because several mutants anchor on lines
those edits moved and a mutant anchored on a moved line silently scores
`DID NOT APPLY`:

`ALL MUTANTS ACCOUNTED FOR` — **99/99 scored, 0 unaccounted, 0 unscored**, run
under `PYTHONDONTWRITEBYTECODE=1` on CPython 3.12.14. Count re-derived with
`grep -c '^ *run '`, not restated (92 at `origin/main` → 99). Both controls
behaved (`comment-only-edit-must-survive` SURVIVED; `main-returns-immediately-
must-die` died), and the battery's own source-integrity check reports the tool
byte-identical to the committed blob. Seven new mutants, each named by the test
that must kill it:

```
ok quotepath-blinds-the-changed-path-listing         killed by test_unlanded_work_under_a_nonascii_name_is_not_called_dead
ok display-decoding-used-to-build-the-pathspec       killed by test_the_changed_path_listing_survives_an_undecodable_filename
ok strict-decode-of-the-changed-path-listing         killed by test_the_changed_path_listing_survives_an_undecodable_filename
ok quotepath-false-aborts-the-comparison             killed by test_paths_differ_survives_quotepath_false
ok changed-path-listing-drops-a-whitespace-named-path killed by test_a_whitespace_named_path_is_not_dropped_from_the_changed_set
ok comparison-drops-a-whitespace-named-path          killed by test_paths_differ_sees_a_whitespace_named_path
ok pathspec-batch-sized-in-characters-not-bytes      killed by test_the_pathspec_batch_sizer_measures_bytes_not_characters
```

An earlier run of this battery was **killed at 87/99** by the harness. A partial
sweep is not a result and was not reported as one; the numbers above are from a
complete run that reached its own verdict line. (The kill also exercised the
file's SIGKILL claim for free: the tool mutates a copy under `mktemp -d`, and
after the kill the working tree was clean, the tool byte-identical to its
committed blob, and no `/tmp/wtp-mut-*` left behind.)

3.12 is the side where the interpreter-conditional mutant
(`resolve-or-none-does-not-swallow-the-loop`) **runs** rather than being skipped,
so this run is the complete one and no `UNSCORED` line was printed. **I did not
run 3.13** — no 3.13 interpreter is available in this dev shell or on the host.

### 🔴 A pre-existing battery defect this turned up

My first two new mutants reported `MUTATION DID NOT APPLY`. The cause: a sed
pattern matching the Python source `b"\0"` must be written `b"\\0"` inside these
**single-quoted** expressions — single quotes do no escape processing, so bash
hands sed what is typed and `\\` in a BRE is one literal backslash. The battery
carried `b"\\\\0"`, which asks sed for *two* backslashes and matches nothing.

`quotepath-blinds-the-index-arm` has carried that broken pattern since #935 and
did **not** report it, because it has a *second* s-expression that does apply —
so `cmp` saw a change and the mutant scored `ok` while running a **narrower**
mutation than its own comment describes ("flag and parser move together"). The
`cmp` gate is per-`run`, not per-expression.

**Correction to an earlier revision of this body**, which said "six occurrences
corrected". Measured: exactly **one** line at `origin/main` carried the broken
4-backslash pattern, and that is the only *repair*. The other five occurrences
are newly-written expressions in this PR that were spelled correctly from the
start once the trap was understood. (A sixth literal `b"\\\\0"` survives on
purpose, in the header comment documenting the trap.) That one mutant is still
killed by its named test with both its expressions now applying.

## Round 1 adversarial audit — six findings, all reproduced before acting

Every one was measured independently here before any edit; none was taken on the
auditor's word. Ledger: `round 1 · payload lines changed THIS round: 347
(280 added / 67 deleted, `git log --numstat --format= --remerge-diff
9666484f..HEAD --not origin/main`, rc 0, silent stderr, 1 commit in range)`.

### 🟡 Finding 3 — FIXED. `if p.strip()` was a SECOND silent wrong `dead`

The one with teeth, and a genuine defect **pre-existing at `origin/main`**.
`bytes.strip()` removes ASCII whitespace, so a `-z` record whose *name* is
whitespace was dropped from the changed set — and a dropped **differing** path is
indistinguishable from "identical upstream". The same observable #975 is about,
reached a different way, in the two lines this PR had already rewritten.

Measured here, git 2.55 — branch adds `a.md` (which then lands identically) and
`" "` (which never lands):

```
$ git diff --name-only -z <merge-base> feat | xxd
00000000: 2000 612e 6d64 00                         .a.md.
  records            : [b' ', b'a.md', b'']
  if p.strip() keeps : ['a.md']          <- the differing path is gone
  if p         keeps : [' ', 'a.md']
$ git diff --name-only feat main -- ' '   ->  b' \n'   <- it GENUINELY differs
$ git diff --name-only feat main -- a.md  ->  b''      <- and a.md really did land
```

Both sites are now `if p`. With `-z` the only empty record is the trailing
terminator, so a bare truthiness test is correct *and* sufficient. The
`no-content-change` short-circuit's comment is fixed too: under `if p.strip()` a
branch changing **only** whitespace-named paths had its whole changed set
filtered away and reached that branch, so the comment asserted "the branch
introduced no content relative to its merge-base" about a branch that had.

Four new tests, two watched RED end-to-end at `origin/main`
(`test_a_whitespace_named_path_is_not_dropped_from_the_changed_set`,
`test_a_whitespace_only_branch_is_not_called_no_content_change`), plus a
`_paths_differ`-level test and a fixture control that pins the drop actually
happens and that `a.md` really is identical upstream.

🔴 **This retires the previous revision's deliberate "no mutant is possible for
the emptiness form" note.** That note was true of the NUL reading and false of
the predicate: `.strip()` also eats a whitespace *name*, which git does emit. Two
real mutants replace it (`changed-path-listing-drops-a-whitespace-named-path`,
`comparison-drops-a-whitespace-named-path`), each with its own killer.
Production-unreachable is not untestable — and the reverse is just as true: a
"coverage is impossible here" note has to go the moment it becomes possible, or
it stops anyone looking.

🔴 **And the first draft of the new fixture control wrote
`differs.stdout.strip() != b""`, which FAILED** — because `b" \n".strip()` is
`b""`. The defect under test, committed inside its own control. It asserts exact
bytes now, and says so in a comment.

### 🟡 Finding 1 — FIXED. A completeness claim that was false

`_git_raw`'s docstring said *"those are every call in this file that handles a
repository path"*. It is not — and it is precisely the finished-audit sentence
#975 objected to in #944's prose. Measured at this PR's HEAD: `worktree_dirty`
parses paths out of `git status --porcelain=v1` through `_git` (strict utf-8),
and a scanned worktree with `core.quotePath=false` holding a single **untracked**
file named `caf\xe9.txt` aborts the whole scan with a bare `UnicodeDecodeError`
out of `subprocess.communicate`, zero classified rows. `list_worktrees` is a
second such call — which the *very next paragraph* already named as open, so the
docstring contradicted itself two paragraphs apart.

The **code** scope is unchanged (that decision stands); the **sentence** was the
defect. Narrowed to "this helper's callers today, NOT an audit of the file", with
both still-open calls enumerated. 🔴 And "tracked filename" is too narrow in both
the docstring and the README row — **untracked is enough** for that arm, because
`worktree_dirty` reads `--untracked-files=normal`. Both fixed, and each "tracked"
elsewhere is now explicitly scoped to the arm it sits in (`ls-files` and `diff`
genuinely only see tracked content).

### 🟡 Finding 2 — FIXED. A guard whose description was wider than its implementation

The section comment credited
`test_an_ascii_only_branch_is_still_classified_correctly` as the control against
a fix that merely stops firing `content-identical` everywhere. Measured: with
`if not differ:` forced to `if False:` in a `cp -a` copy of the tree, that test
**passes**. It is now labelled an invariant guard.

> ⚠ **Two numbers in this paragraph were wrong and are corrected in the round-2
> section below.** It said "only two go red" and "five of the other six": the
> real figure is **22 of 220** file-wide (round 1 measured under a `-k` filter
> and reported the result as a fact about the file), and within the #975 section
> it is one of **seven**, not two of eight. The round-2 section carries the
> re-measurement and the method.

Structural, not incidental: both branches in `nonascii_universe` are *unlanded*,
so the signal is supposed to stay quiet on both and a tool that never fires it
satisfies the ASCII twin by construction.

*(An earlier revision of this paragraph also claimed the twin would go red if the
pair disagreed. It would not — the body reads one row. See round 2, F2.)*

### 🟢 Finding 4 — FIXED. One red was red for an incidental reason

`test_the_pathspec_batch_sizer_measures_bytes_not_characters` failed at base on
`assert seen, "the batcher never called git"` — base's `_paths_differ` calls
`_git`, not the monkeypatched `_git_raw`, so the budget assertion never executed
and the matrix read as regression coverage the test does not supply. `_git` is
monkeypatched too now, and at base it goes red on the real assertion:
`a batch of 6 path(s) encoded to 234 bytes, over the 100-byte budget — the sizer
is counting characters`.

### 🟢 Finding 5 — FIXED. That test's own numbers did not survive measurement

It said "one item too big" and "a byte-correct sizer fits 2 per batch and a
char-based one fits 3". Measured with its own fixture at a 100-byte budget:

```
byte-correct len(os.fsencode(p)) : 6 batches of 2, encoded 78/78/78/78/78/80  (all under)
char-based   len(p)              : 2 batches of 6, encoded 234 and 236        (2.3x over)
```

The assertion was right and kills its mutant; only the explanation was wrong, and
it **understated** the defect. Rewritten from the measurement.

### 🟢 Finding 6 — FIXED. Two PR-body numbers, corrected in place and called out

Both were merge evidence and both were wrong: "4 of the **10** new tests" (the
branch added **8** at that point, **12** now) and "**six** occurrences corrected"
of the sed-escaping bug (exactly **one** line at base carried it; the other five
were newly-written correct spellings, not repairs). Corrected above rather than
silently rewritten, with a note at each site, and a PR comment records the change.

### Also, not raised as a finding

The README's new sentence said both halves are "decoded with `os.fsdecode`". Only
the **listing** decodes; `_paths_differ` receives `str` pathspecs and only tests
its own output bytes for emptiness. A comment is a claim; fixed.

## Round 2 delta re-audit (`9666484f..02d2b811`) — two findings, both mine, both scaffolding

Round 2 confirmed 9 of 11 round-1 claims outright and found two 🟡, both false
*measurements stated in prose that round 1's own fixes wrote*. Both reproduced
here before acting. **Payload this round is zero**, proven structurally: with
docstrings blanked, the AST of both changed files is byte-identical to
`02d2b811` (controls: a file against itself → IDENTICAL; `02d2b811` against
`origin/main` → DIFFERS).

Ledger: `round 2 · payload lines changed THIS round: 0 · scaffolding: 89
(68 added / 67→21 deleted) · raw: 89` — `git log --numstat --format=
--remerge-diff 02d2b811..HEAD --not origin/main`, rc 0, stderr 0 bytes, 1 commit
in range.

### 🟡 F1 — FIXED. The section comment's measurement was wrong on both numbers

It said forcing `if not differ:` to `if False:` "kills exactly TWO tests", named
them, then "six of the **eight** tests in this section".

**Re-measured** (`cp -a` copy, `.git` severed, sed on that one line, the **whole
file**, `PYTHONDONTWRITEBYTECODE=1`): **22 of 220 fail**, not two — including
`test_execute_removes_only_the_dead_rows`,
`test_json_output_carries_the_evidence_for_every_row`,
`test_the_squash_row_is_removable`, `test_without_gh_the_squash_is_still_dead`
and 18 more. Mechanically expected: every end-to-end fixture whose row is `dead`
*by way of* `content-identical` degrades to `orphan`, so the reds sweep past this
section into the executor, the JSON report and the exclusion tests.

🔴 **How round 1 got "two": it ran the mutation under a `-k` filter and reported
the result as a fact about the file.** A filtered run cannot answer a question
about the file. The rewritten comment now states the **method** next to the
number and says the run must be unfiltered.

"Eight" was wrong too: the section holds **seven** `def test_`. The eighth,
`test_squash_merged_branch_is_dead_not_orphan`, is at `:266` in a different
section. Six of seven stay green.

Why it mattered: a maintainer deleting or refactoring the two named tests would
have read it as *"the protection against silencing the signal is now gone"*. It
is not — ~20 others catch it. The comment now pins the **class** (every
end-to-end fixture reaching `dead` via `content-identical`) and says explicitly
not to maintain the integer, because it will drift.

### 🟡 F2 — FIXED by narrowing (a), with (b) tried and rejected on measurement

The invariant-guard docstring — rewritten **last round specifically to stop
over-claiming** — over-claimed in the same family: *"if a future change made the
non-ASCII row `cannot-tell` while the ASCII twin stayed `orphan`, the pair would
disagree and this would go red."* The body reads **one** row.

**Measured**: forcing `_paths_differ` to return `None` for any non-ASCII pathspec
produces exactly that split — `e2e-nonascii: verdict='cannot-tell'` beside
`e2e-ascii: verdict='orphan'` — and the test **passes**. A docstring naming a
RELATIONSHIP over a body inspecting one SIDE is this repo's own named shape.

**Took (a), narrow the docstring — and (b) was built and rejected with a
measurement rather than waved off.** A candidate guard asserting
`(verdict, removable, signals)` equal across the twins was written and run
against that same probe. It goes red — but **alongside**
`test_unlanded_work_under_a_nonascii_name_is_not_called_dead`, never instead of
it, and it is the unique killer of nothing. Structural, not luck: both rows are
already pinned to the **literal** `orphan` by two separate tests, so any
disagreement necessarily breaks one of those literal pins first, and a
pair-equality assertion over two independently-pinned constants is implied by
them. Shipping it would have been a guard that reads as coverage while adding
none — precisely what this docstring was rewritten to stop doing. The docstring
now states what the body does, what it does **not** do, and why the pair guard
was rejected.

### Left alone on purpose (round 2 named both as non-findings)

The `_git_raw` docstring carries the "tracked is not the boundary" correction
twice; the duplication is deliberate and stays. One token was fixed: the closing
paragraph sourced the phrase *"any undecodable TRACKED filename"* to "the
MEASURED note above" when that wording lives in `scripts/README.md`. Attribution
corrected, structure untouched.

### Gates for round 2

* **Full suite (branch)**: 28 per-target `N passed` lines,
  `TOTAL collected=18288 passed=18286 skipped=2 failed=0`,
  `RESULT: PASS (exit=0)`; no timeout panic or `INTERNALERROR`.
* **Battery — SUBSET, labelled as one.** Only comments and docstrings moved, and
  an anchor probe confirmed **all 99** mutant expressions still apply
  (negative control: 7 `DID-NOT-APPLY` against `origin/main`'s tool). So the
  `#975` section plus **both** controls were run rather than the full 99:
  **9/9, `ALL MUTANTS ACCOUNTED FOR`**, baseline clean,
  `comment-only-edit-must-survive` SURVIVED, `main-returns-immediately-must-die`
  died, source sha equal to the committed blob. This is **not** a full sweep and
  is not offered as one; the last full 99/99 sweep was at `02d2b811`, whose tool
  is AST-identical to this one.
  *(The subset harness's first run reported `HARNESS BROKE — only 0 test(s) ran`
  because my copy resolved `SRC` to the scratchpad. That is the battery's own
  floor guard doing its job, and it is reported rather than hidden.)*
* **Merged-tree gate**: see the section above — delta against a pristine
  `origin/main` control, zero new failures.

## Round 3 closing edit (`e2164eca..0c7cc076`) — the ladder's attribution gate fired

Round 3 confirmed all 7 round-2 claims by independent re-measurement, including
rebuilding the rejected pair guard and reaching the same verdict. Rounds 2 and 3
both changed **zero payload lines**, so the ladder has left this PR and stops
here. Two clause-level prose corrections, both reproduced before acting.

Ledger: `round 3 · payload: 0 · scaffolding: 14 · raw: 14` (11 added / 3 deleted)
— rc 0, stderr 0 bytes, 1 commit in range.

### 🟡 R3-F1 — FIXED. The shipped operator doc named the wrong control

`scripts/README.md`'s row claimed the ASCII/non-ASCII twins are *"the control
that a fix which merely silences the signal would fail"*. They are not, and this
PR's own test docstring already said so.

**Attributed**: 0 occurrences at `origin/main`, 1 from `c06bda00` onward —
introduced by this PR's first fix commit and carried through two audit rounds.
**Re-measured**: with `if not differ:` forced to `if False:`, *neither* twin is
among the 22 reds — both their branches are unlanded, so the signal is supposed
to stay quiet on both.

Concrete failure it enabled: a maintainer trimming tests reads the shipped doc as
the authority on which fixtures guard the silencing direction, weakens
`test_content_identical_still_fires_for_a_landed_nonascii_path` (the real
control, with `test_squash_merged_branch_is_dead_not_orphan`), and
`content-identical` becomes permanently unreachable for non-ASCII paths — every
squash-merged worktree silently downgrading `dead` → `orphan`, nothing red. The
row now names those two and says not to trim them on the twins' strength.

🔴 **Why it survived two rounds: no test reads that row.** Measured —
`scripts/README.md` is opened by exactly one test in the repo
(`test_transcript_search.py`, for an unrelated file-count claim), and nothing
reads the `worktree-prune` row. A human reading it was the only detector that
could exist, and that is what caught it.

### 🟢 R3-F2 — FIXED. A measurement stated without its placement

The invariant-guard docstring named a two-test red set beside a probe described
as forcing `_paths_differ` to return `None`. **Measured both placements**: inside
`_paths_differ` → **four** reds (two tests call it directly with a non-ASCII
path); at the call site (`differ = None` in `landing_signals`) → exactly the
**two** named. The docstring now carries the placement and both counts. The
conclusion is unchanged — the test is green under both, which is the point.

### Gates — light, as a decision rather than an omission

* **AST identity vs `e2164eca`**: `test_worktree_prune.py` → **IDENTICAL**
  (docstrings blanked); `scripts/worktree-prune` **not touched this round**.
  Controls both ways (self → IDENTICAL; `origin/main`'s tool → DIFFERS). Zero
  executable change, proven not asserted.
* **Targeted tests**: `test_worktree_prune.py` 220 · `test_doc_path_rot.py` 77 ·
  `test_rules_size.py` 5 · `test_transcript_search.py` 35 — all passed.
* **No full battery, no full suite, no merged-tree re-run**, deliberately:
  nothing executable changed, so the battery would re-derive a held result over
  an AST-identical tool and the merged-tree gate would re-measure a delta against
  a base that is red for two unrelated reasons.

⚠ **Correction to the round-3 brief's premise, measured**: the two named
doc-lint gates are structurally blind to `scripts/README.md` —
`test_rules_size.py` measures only `claude/RULES*.md`, and `test_doc_path_rot.py`'s
corpus is `("claude", "CLAUDE.md")` with `README.md` explicitly excluded. Their
green says nothing about this edit. Ran them anyway, reported as such, and added
`test_transcript_search.py` — the only test that actually opens the file.

## Merged-tree gate

Re-run for round 2 against a freshly fetched `origin/main` (**`07890ebc`**).
Merge is **clean, zero conflicts**, and no commit on main since this PR began has
touched any of the four files it changes.

🔴 **`origin/main` is currently RED on its own, for two reasons that have nothing
to do with this PR — so the honest gate here is a DELTA, not a green tick.** The
control was run before any theory: a pristine `origin/main` worktree, zero of
this branch's changes present.

| | collected | passed | skipped | failed |
|---|---|---|---|---|
| pristine `origin/main` (control) | 18342 | 18338 | 2 | **2** |
| main + this branch (merged tree) | 18354 | 18350 | 2 | **2** |

The failure **sets** are identical, not merely the counts:

```
test_engine_is_the_version_every_measurement_is_keyed_to
test_live_existing_resolutions_not_made_ambiguous
```

`comm -13 control merged` is **empty** — this branch introduces **zero** new
failures. `collected` rises by exactly **+12**, which is exactly the 12 tests
this PR adds, and all 12 pass. `scripts/tests/test_worktree_prune.py` on the
merged tree: **220 passed**.

What is actually red on `main`, for whoever owns it:

* `test_engine_is_the_version_every_measurement_is_keyed_to` — **environmental**:
  the host's `opencode` is `1.18.21` while the repo keys its measurement claims
  to `1.18.18`. Nothing in the tree is wrong; the pin is stale.
* `test_live_existing_resolutions_not_made_ambiguous` — a **live-data** espanso
  test (`_live_det()` reads the installed snippets). The term `ask` now matches
  both `:dacq` and `:acq`, expected `:acq`. Consistent with `8bab93b1` splitting
  `:acq` into `:dacq` + a bare `:acq`, with #999's tie-break not covering it.

Neither is mine to fix in this PR, and neither is being quietly counted as green.

## #973 test-merge (report only — #973's branch was not touched)

`gh pr view 973` reports `MERGEABLE` / `CLEAN` — that is against `main`, not
against this branch. Merging this branch into #973's head `2d0b8112`:

**Git-level: 3 of the 4 shared files conflict.** `scripts/worktree-prune`,
`scripts/README.md` and `scripts/tests/mutants-worktree-prune.sh` each produce
exactly **one** conflict hunk; `scripts/tests/test_worktree_prune.py`
auto-merges.

**Semantically: no code conflict.** All three hunks are prose or adjacency:

* `scripts/worktree-prune` — the sole hunk lies entirely between `def _git_raw`
  and `def _printable`, i.e. inside `_git_raw`'s **docstring**. `_git_raw`'s
  *body* is unchanged by #973, and #973's diff hunks in this file never touch
  `landing_signals` or `_paths_differ`.
* `scripts/tests/mutants-worktree-prune.sh` — a pure adjacency conflict: both
  sides append a new section immediately before `== POSITIVE CONTROLS ==`, with
  nothing between them in the base. Resolution is concatenation.
* `scripts/README.md` — the `worktree-prune` row is one enormous markdown table
  line, so any two edits to it collide. A hand-written union sentence is needed;
  a mechanical union splits the row in two.

Verified rather than reasoned about: resolving all three by union gives a tree
where **`scripts/tests/test_worktree_prune.py` runs 230 passed**, the tool parses,
the battery parses, and a probe that applies every one of the merged battery's
**116** mutant sed expressions reports **0 `DID-NOT-APPLY`** — i.e. no anchor on
either side moved. That probe was controlled both ways: it reports 97/0 against
this branch (agreeing with the real battery's `ALL MUTANTS ACCOUNTED FOR`) and 5
`DID-NOT-APPLY` against `origin/main`'s tool, exactly the 5 new mutants whose
anchors do not exist there.

**The one thing a merger must hand-write, and it is a real semantic conflict.**
The two sides' `_git_raw` docstrings make *mutually falsifying* claims about what
is still open:

* This branch says `worktree list --porcelain` remains an open abort site — true
  on `main` today, **false** once #973 lands.
* #973 says #975 is "STILL OPEN, MEASURED, NOT FIXED HERE" and enumerates the
  diff arm's two hazards — **false** once this lands.

Whichever merges second must rewrite that paragraph; taking either side verbatim
ships a false claim. Also worth carrying forward: #973 names a hazard this PR
does **not** fix and this PR's prose does not mention — `worktree_dirty`'s
`git status --porcelain=v1` aborts under `core.quotePath=false`. That is
genuinely still open after both.

**Not affected:** #973 introduces `path_fs`/`repo_fs`, `ROW_PATH_FIELDS` and
`row_fs` with a seam guard asserting that every non-`_fs` row value survives
`encode("utf-8")`. This change adds no row field — no path from the changed-path
set ever reaches a row or `print()`; `landing_signals` puts only a *count* into
its notes — so the ledger and its guard are untouched.

## Deliberately not done

* **`worktree_dirty`'s `git status --porcelain=v1`** under `core.quotePath=false`
  is left open, as #973 documents. It is a different function with a different
  failure mode (a loud abort, not a silent wrong `dead`), it wants its own
  fixture, and widening this PR into it would put the arm that decides `dead`
  behind a larger review.
* **`worktree list --porcelain`** (#944) is #973's work, not restarted here.
* **A `_printable` (U+FFFD) decoding in the diff arm.** The two-decodings split
  needs both directions only where a path is both *reached* and *shown*. No path
  from this set is ever printed, so only `os.fsdecode` is used, and the code says
  so explicitly rather than leaving the asymmetry to be read as an oversight.
* **The 3.13 battery run** — no 3.13 interpreter available here. Stated as
  unverified, not as green.

## Closing condition

**This PR deliberately carries no auto-closing keyword for issue #975** — and the
keyword is not written out anywhere in this body, because GitHub's parser reads
it inside backticks too. The issue's closing condition is "whoever picks this up,
running the three blocks above **on a merged PR**" — a post-merge measurement.
Auto-closing on merge would skip exactly that gate, which is how #931 was lost.
Referencing without closing; please run the issue's three blocks after merge and
close it by hand.

https://claude.ai/code/session_01STBNLv58zefYcRmNDzcWdv


ZacxDev:
CI settled after this body was written: `tekton/devrc-pytests` **pass** — `TOTAL collected=18284 passed=18282 skipped=2 failed=0` (floor 15970 = sum of 28 per-target floors), byte-identical to the local merged-tree gate reported above; `tekton/devrc-nodetests` **pass** — `suites=5 files=39 tests=1300 pass=1300 fail=0 skipped=0`. PR is `MERGEABLE` / `CLEAN` against `main`.

ZacxDev:
## Round 1 adversarial audit — pushed as `02d2b811`

Six findings, all reproduced here before any edit. Full write-up is in the body;
this comment exists for the two things that must not be changed silently.

### 🔴 Two numbers in the previous body were wrong. Correcting them on the record.

Both were merge evidence, so they get a comment rather than a quiet rewrite.

1. **"4 of the 10 new tests are red at `origin/main`."** The branch added
   **8** tests at that point, not 10 (`^def test_` 183 → 191;
   `grep -c '^+def test_'` on the diff = 8). The "4 red" half was right. The
   branch now adds **12** (183 → 195), and the matrix is **7 failed / 213 passed
   at `origin/main`, 220 passed at HEAD**.
2. **"Six occurrences corrected"** of the `b"\\\\0"` sed-escaping bug. Exactly
   **one** line at `origin/main` carried the broken 4-backslash pattern, and that
   is the only *repair*. The other five are newly-written expressions in this PR,
   spelled correctly from the start once the trap was understood. (A sixth
   literal survives on purpose, in the header comment that documents the trap.)

### 🔴 Finding 3 was a real defect, pre-existing at `origin/main`

`if p.strip()` on a `-z` record drops a path whose **name** is whitespace, and a
dropped *differing* path is indistinguishable from "identical upstream" — the
same observable #975 is about, reached a different way, in the two lines this PR
had already rewritten. Measured end to end:

```
pre-change   e2e-whitespace: verdict='dead'   removable=True  signals=['content-identical']
               evidence: all 1 changed path(s) have identical content   <- the branch changed TWO
post-change  e2e-whitespace: verdict='orphan' removable=False signals=[]
               evidence: at least one of 2 changed path(s) differs
```

Both sites are `if p` now. This also made `_paths_differ`'s emptiness form
**reachable**, retiring the previous round's deliberate "no mutant is possible
here" note — replaced by two real mutants with real killers. A
"coverage is impossible" note has to go the moment it becomes possible, or it
stops anyone looking.

### Gates, all re-run after the edits

| gate | result |
|---|---|
| branch suite | 28 targets, `collected=18288 passed=18286 skipped=2 failed=0`, `RESULT: PASS (exit=0)` |
| mutation battery | `ALL MUTANTS ACCOUNTED FOR` — **99/99 scored**, 0 unaccounted, 0 unscored, both controls correct |
| merged-tree gate | off `origin/main` `e66be7e6`, clean merge, 28 targets, `collected=18327 passed=18325 skipped=2 failed=0`, `RESULT: PASS` |

Ledger: `round 1 · payload lines changed THIS round: 347 (280 added / 67 deleted)`
— `git log --numstat --format= --remerge-diff 9666484f..HEAD --not origin/main`,
rc 0, silent stderr, 1 commit in range (all three preconditions checked, since a
git without `--remerge-diff` exits 128 with empty output and reads exactly like a
clean zero).

⚠ **One honest caveat on the battery**: an earlier run of it was **killed at
87/99** by the harness. A partial sweep is not a result and is not being reported
as one — the 99/99 above is from a complete run that reached its own verdict
line. The kill did incidentally confirm the file's SIGKILL claim: afterwards the
working tree was clean, the tool byte-identical to its committed blob, and no
`/tmp/wtp-mut-*` was left behind.

⚠ **Still not run: the battery on CPython 3.13.** No 3.13 interpreter exists in
this dev shell or on the host. 3.12 is the side where the interpreter-conditional
mutant *runs* rather than being skipped, so this sweep is the complete one
(`0 UNSCORED`) — but the 3.13 skip path is unverified, and is stated as such
rather than as green.

ZacxDev:
## Adversarial audit — round 1

Round 1 ran blind against `9666484f` and returned six findings (3 × 🟡, 3 × 🟢). All six were addressed in `02d2b811`.

`round 1 · payload lines changed THIS round: 347 (280 added / 67 deleted) · elapsed: ~4h wall` — measured with `git log --numstat --format= --remerge-diff 9666484f..HEAD --not origin/main`, rc 0, stderr silent, non-empty range. Non-zero, so the ladder has not left the PR.

Independently re-measured by the operator rather than accepted from the fix agent: the full battery (97/97 at `9666484f`, both controls behaving), the red/green matrix (4 failed / 212 passed at base; 216 passed at head), finding 3's mechanism on a fresh fixture, and the two PR-body counts below.

```audit-claims round=1 audited=9666484f7daaba9b855a2bac8336c075a3448ff8..02d2b811
1. Both record-filter sites in the changed-path path now use `if p` / `any(p ...)` instead of `p.strip()`, so a genuinely-differing whitespace-named path is no longer dropped from the changed set and no longer produces a false `content-identical`.
2. The "NO MUTANT FOR THE EMPTINESS FORM ITSELF" note is retired and replaced by two mutants (`changed-path-listing-drops-a-whitespace-named-path`, `comparison-drops-a-whitespace-named-path`), each naming a killing test; battery count 97 -> 99.
3. Four new tests cover the whitespace-name defect, claimed watched RED against the pre-change tool in an isolated copy.
4. The `no-content-change` comment no longer claims the branch introduced no content in the case where a whitespace-named path was dropped.
5. `_git_raw`'s "Those are every call in this file that handles a repository path" is narrowed to this helper's callers today, and both still-open calls (`list_worktrees`, `worktree_dirty`) are enumerated.
6. "tracked filename" is scoped in both the `_git_raw` docstring and the `scripts/README.md` row, because an untracked file is enough to trigger the `worktree_dirty` abort.
7. The ASCII-control comments at the `nonascii_universe` fixture and on `test_an_ascii_only_branch_is_still_classified_correctly` are re-attributed to the two tests that actually kill a blanket-silencing of `content-identical`, and the ASCII twin is labelled an invariant guard stating what it does pin.
8. `test_the_pathspec_batch_sizer_measures_bytes_not_characters` monkeypatches `_git` as well as `_git_raw`, so at base it fails on its budget assertion rather than on "the batcher never called git".
9. That test's prose numbers are rewritten from measurement: byte sizer 6 batches of 2, char sizer 2 batches of 6 encoding 234/236 bytes against a 100-byte budget.
10. `scripts/README.md`'s claim that both halves decode with `os.fsdecode` is corrected — only the changed-path listing decodes.
11. The PR body's two overstated counts are corrected in place with a note at each site, and a separate PR comment records the correction.
```

### Corrections to this PR's own body, on the record

Both were caught by re-verification, not by review, and both were overstatements in merge evidence:

- **"4 of the 10 new tests are red at `origin/main`"** — at `9666484f` there were **8** new test functions, not 10 (`^def test_` 183 → 191). The "4 red" half was correct and reproduced exactly. After round 1 the file is at **195**, i.e. 12 new.
- **"Six occurrences corrected"** of the `b"\\0"` sed-escaping bug — the battery diff was **67 insertions / 1 deletion**, and exactly **one** pre-existing line carried the broken 4-backslash pattern. The other five were newly-written correct spellings, not repairs. The only residual 4-backslash string is in the header comment that documents the trap, which is correct.

### Still unverified, stated rather than implied

- **No 3.13 battery run** — no 3.13 interpreter is available on this host. 3.12 is the side on which the interpreter-conditional mutant *runs*, so the sweep is complete (`0 UNSCORED`), but the 3.13 skip path is untested.
- **The #973 test-merge was not re-measured this round.** Round 1's finding stands in shape — three prose/adjacency conflicts, no code conflict, and two `_git_raw` docstrings making mutually falsifying claims — but round 1's edits rewrote that docstring further, so the specific conflict text will differ.
- `#975` carries **no auto-closing keyword** on purpose: its closing condition is a post-merge measurement, not a merge.

ZacxDev:
## Correction — the round-1 ledger line above was wrong

Round 2 caught this and it is mine, not the fix round's. My comment above reports:

> `round 1 · payload lines changed THIS round: 347 (280 added / 67 deleted)`

**347 is the raw `--numstat` total. It is not a payload count** — I never performed the payload/scaffolding classification the gate is defined on. Re-measured just now (`rc 0`, stderr 0 bytes, 1 commit in range, per-file numbers reproduced exactly):

```
git log --numstat --format= --remerge-diff 9666484f..02d2b811 --not origin/main

  65   29  scripts/worktree-prune                   -> payload       94
   1    1  scripts/README.md                        -> payload        2
 191   25  scripts/tests/test_worktree_prune.py     -> scaffolding   216
  23   12  scripts/tests/mutants-worktree-prune.sh  -> scaffolding    35
```

**The corrected line is `round 1 · payload lines changed THIS round: 96 · scaffolding: 251`.**

`scripts/README.md` is classified payload: it is the shipped tool's user-facing reference row, not a test, fixture or note written to guard the change.

This does not move round 1's gate outcome — 96 and 347 are both non-zero, so the ladder had not left the PR either way. It matters because the stop signal is *two consecutive rounds of **zero payload** lines*, and **a ledger that skips the classification can never report zero**. A gate that structurally cannot fire is the failure it exists to prevent, so the wrong number was the more dangerous kind of wrong: it would have kept the ladder running on scaffolding indefinitely without ever saying so.

Also correcting one line of my own `audit-claims` block above, same reason — it is merge evidence and the next round reads it:

- Claim 3 said *"Four new tests … watched RED against the pre-change tool."* Measured: **three** of the four go red at base; the fourth (`test_the_whitespace_named_fixture_really_does_differ_upstream`) never calls the tool and passes at both bases — it is a fixture control and its own docstring says so. The commit body's own wording ("four new tests, **two** watched RED end-to-end") was accurate; my paraphrase of it was the overstated form.

ZacxDev:
## Round 2 delta re-audit — pushed as `e2164eca`

Both findings were mine, both `scaffolding`, and both were *false measurements in
prose that round 1's own fixes wrote*. Reproduced here before acting; full
write-up is in the body. Two things that must not change silently:

### 🔴 A round-1 number in this PR's body was wrong. Corrected in place, on the record.

Round 1's Finding-2 paragraph said forcing `if not differ:` to `if False:` kills
"only two" tests and named them. **Re-measured on the whole file: 22 of 220
fail** — the reds sweep past the `#975` section into the executor, the JSON
report and the exclusion tests, because every end-to-end fixture that reaches
`dead` *via* `content-identical` degrades to `orphan`.

🔴 **How round 1 got "two": it ran the mutation under a `-k` filter and reported
the answer as a fact about the file.** A filtered run cannot answer a question
about the file. The in-tree comment now carries the **method** next to the
number, and pins the *class* rather than an integer that will drift. ("Six of the
eight tests in this section" was also wrong — the section holds **seven**; the
eighth is in a different section.)

Why it mattered: a maintainer refactoring the two named tests would have read it
as "the protection against silencing the signal is gone". It is not — ~20 others
catch it.

### 🔴 F2: the fix is (a), and (b) was built and rejected on a measurement

The docstring claimed the ASCII twin would go red if the pair disagreed. It does
not — the body reads one row, and forcing `_paths_differ` to return `None` for
non-ASCII pathspecs yields `cannot-tell` beside `orphan` with the test **passing**.

I wrote the option-(b) pair guard and ran it against that probe: it goes red, but
**alongside** `test_unlanded_work_under_a_nonascii_name_is_not_called_dead`,
never instead of it — the unique killer of nothing. Both rows are already pinned
to the *literal* `orphan` by two separate tests, so any disagreement breaks one
of those pins first; a pair-equality assertion over two independently-pinned
constants is implied by them. Shipping it would have been a guard that reads as
coverage while adding none. Narrowed the docstring instead, and recorded why.

### 🔴 `origin/main` is currently RED on its own — the merged-tree gate is a DELTA, not a green tick

Control run before any theory: a **pristine `origin/main`** worktree with zero of
this branch's changes.

| | collected | passed | skipped | failed |
|---|---|---|---|---|
| pristine `origin/main` `07890ebc` | 18342 | 18338 | 2 | **2** |
| main + this branch | 18354 | 18350 | 2 | **2** |

Identical failure **sets**, not just counts — `comm -13` is **empty**, so this
branch introduces **zero** new failures. `collected` rises by exactly **+12**,
which is exactly the 12 tests this PR adds, all passing.
`test_worktree_prune.py` on the merged tree: **220 passed**.

What is red on `main`, for whoever owns it:

* `test_engine_is_the_version_every_measurement_is_keyed_to` — **environmental**:
  host `opencode` is `1.18.21`, the repo keys its claims to `1.18.18`.
* `test_live_existing_resolutions_not_made_ambiguous` — a **live-data** espanso
  test; `ask` now matches both `:dacq` and `:acq`, expected `:acq`. Consistent
  with `8bab93b1` splitting `:acq`, with #999's tie-break not covering it.

Neither is being counted as green here.

### Gates

* **Branch suite**: 28 per-target lines, `collected=18288 passed=18286 skipped=2
  failed=0`, `RESULT: PASS`.
* **Battery — SUBSET, and labelled as one.** Only comments/docstrings moved
  (proven: AST identical to `02d2b811` with docstrings blanked, controls both
  ways), and an anchor probe confirmed all **99** mutant expressions still apply.
  So the `#975` section **plus both controls** was run: **9/9, ALL MUTANTS
  ACCOUNTED FOR**, baseline clean, source sha equal to the committed blob. Not a
  full sweep and not offered as one — the last full 99/99 was at `02d2b811`,
  whose tool is AST-identical to this one.
* Ledger: `round 2 · payload 0 · scaffolding 89 · raw 89` (rc 0, stderr 0 bytes,
  1 commit in range).

⚠ Still unverified, unchanged from round 1: **the battery on CPython 3.13** — no
3.13 interpreter exists here. 3.12 is the side where the interpreter-conditional
mutant runs, so the sweeps are complete (`0 UNSCORED`), but the 3.13 skip path is
stated as unverified rather than green.

ZacxDev:
## Adversarial audit — round 2 (delta)

Round 2 ran against `9666484f..02d2b811`. It confirmed **9 of round 1's 11 claims outright** and returned **two 🟡, both `scaffolding`** — false measured claims in prose that round 1's own fixes wrote — plus one 🟢 against the operator's ledger line, corrected above. Addressed in `e2164eca`.

`round 2 · payload lines changed THIS round: 0 · scaffolding: 89 · raw: 89` — `git log --numstat --format= --remerge-diff 02d2b811..e2164eca --not origin/main`, rc 0, stderr 0 bytes, 1 commit in range.

🔴 **Payload-zero is proven, not asserted, and the proof is controlled both ways.** The range touches `scripts/worktree-prune` (4+/3-), which is the payload file — so a line count alone would read 7. Parsing both revisions with docstrings blanked (comments never enter the AST) gives **identical dumps**:

```
02d2b811..e2164eca  scripts/worktree-prune  -> IDENTICAL   (under test)
origin/main..02d2b811  scripts/worktree-prune  -> DIFFERS   (positive control: the instrument CAN say DIFFERS)
e2164eca..e2164eca  scripts/worktree-prune  -> IDENTICAL   (negative control)
```

**This is round 1 of the attribution gate's two consecutive zero-payload rounds.** Round 1's fixes were 96 payload; round 2's are 0.

```audit-claims round=2 audited=02d2b81176d7654c9ff7add5900c02e77c08ca1d..e2164eca
1. The section comment at test_worktree_prune.py:718 no longer states "kills exactly TWO tests" — re-measured on the WHOLE file (22 of 220 fail, not 2) and rewritten to state the method beside the number.
2. That comment no longer says "six of the eight tests in this section" — the section holds seven; the eighth was being counted from an unrelated section 400 lines away.
3. That comment now pins the CLASS that goes red (every end-to-end fixture reaching `dead` via `content-identical`) rather than an integer a refactor will move, and says explicitly not to maintain the integer.
4. `test_an_ascii_only_branch_is_still_classified_correctly`'s docstring no longer claims the pair would disagree and go red — narrowed to what the body actually inspects, which is the ASCII row alone.
5. Option (b) — making the body pin the pair relationship — was built, measured and REJECTED, with the reason recorded in the docstring: the pair guard goes red alongside `test_unlanded_work_under_a_nonascii_name_is_not_called_dead`, never instead of it, so it is the unique killer of nothing.
6. `_git_raw`'s closing paragraph now sources the phrase "any undecodable TRACKED filename" to `scripts/README.md` rather than to a note in that docstring.
7. No payload behaviour changed this round: the tool's AST with docstrings blanked is identical to `02d2b811`.
```

### 🔴 The merged-tree gate is NOT green, and this PR is not why

Reported as a zero delta against a red base rather than as a pass. The discriminating control was run **before** any theorising — a pristine `origin/main` worktree carrying none of this branch:

| tree | collected | passed | skipped | failed |
|---|---|---|---|---|
| pristine `origin/main` `07890ebc` | 18342 | 18338 | 2 | **2** |
| main + this branch | 18354 | 18350 | 2 | **2** |

Identical failure **sets**, not merely equal counts — `comm -13` is empty, so this branch introduces **zero** new failures. `collected` +12 is exactly this PR's 12 tests, all passing. `test_worktree_prune.py` on the merged tree: **220 passed**.

The two reds on `main` are dev-host-tier only and in unrelated subsystems:

- `test_engine_is_the_version_every_measurement_is_keyed_to` — environmental. Verified independently: host `opencode` is **1.18.21**; the repo keys its measurements to **1.18.18**.
- `test_live_existing_resolutions_not_made_ambiguous` — a live-data espanso test; `ask` now matches both `:dacq` and `:acq`. Consistent with `8bab93b1` splitting `:acq`.

Both are invisible to the sandbox tier, which is the tier the merge gates on and which is green on this head (`tekton/devrc-pytests`, `tekton/devrc-nodetests`). That is the two-tier rule working as intended, not a reason to read either tier's verdict as the whole answer. Filed as a separate concern; it is not this PR's to fix.

### Still unverified, stated rather than implied

- **Battery this round was a SUBSET and is labelled one**: the `#975` section plus both controls, **9/9 accounted for**, `comment-only-edit-must-survive` SURVIVED, `main-returns-immediately-must-die` died, source sha equal to the committed blob. An anchor probe first confirmed all **99** expressions still apply, with a negative control reporting 7 `DID-NOT-APPLY` against `origin/main`'s tool. The last full 99/99 sweep was at `02d2b811`, whose tool is AST-identical to this head.
- **No CPython 3.13 run** — no 3.13 interpreter on this host. 3.12 is the side on which the interpreter-conditional mutant runs, so sweeps are complete (`0 UNSCORED`), but the 3.13 skip path is untested, not green.
- **The #973 test-merge has not been re-run since the round-0 measurement.** Rounds 1 and 2 both rewrote the same `_git_raw` docstring, so the specific conflict text is stale; the structural finding (prose/adjacency only, no code conflict, two mutually falsifying docstrings) is unchanged.
- `#975` still carries **no auto-closing keyword** — its closing condition is a post-merge measurement, re-verified on the live body, and the issue is still `OPEN`.

ZacxDev:
## Round 3 closing edit — pushed as `0c7cc076`

Two clause-level prose corrections; the ladder's attribution gate has fired and
this is the last pass. Both findings reproduced here before acting.

### 🟡 R3-F1 — FIXED. The shipped operator doc named the wrong control

`scripts/README.md`'s `worktree-prune` row said the ASCII/non-ASCII twin fixtures
are *"the control that a fix which merely silences the signal would fail"*. They
are not — and `test_an_ascii_only_branch_is_still_classified_correctly`'s own
docstring already refuted it in as many words.

**Attributed** (with a positive control, after zsh's `${r}:s` history-modifier
trap made the first attempt return a confident all-zero): **0** occurrences at
`origin/main`, **1** from `c06bda00` onward. Introduced by this PR's first fix
commit, carried through two audit rounds.

**Re-measured**: with `if not differ:` forced to `if False:`, **neither** twin is
among the 22 reds. Both branches in `nonascii_universe` are unlanded, so the
signal is supposed to stay quiet on both and a tool that never fires it satisfies
them by construction.

The row now names the real controls —
`test_content_identical_still_fires_for_a_landed_nonascii_path` and
`test_squash_merged_branch_is_dead_not_orphan` — and says not to trim them on the
strength of the twins. Wording reused from the test docstring rather than
composed afresh, as asked; the clause was replaced, the row not rewritten.

🔴 **Why it survived two rounds, which is the part worth keeping: no test reads
that row.** Measured — `scripts/README.md` is opened by exactly one test in the
repo (`test_transcript_search.py:1191`, for an unrelated file-count claim), and
nothing reads the `worktree-prune` row. There was no instrument that could see
this; a human reading it was the only possible detector, and that is what caught
it.

### 🟢 R3-F2 — FIXED. A measurement without its placement

The invariant-guard docstring named a two-test red set beside a probe described
as *forcing `_paths_differ` to return `None` for any non-ASCII pathspec*.
**Measured both placements myself**: inside `_paths_differ` → **four** reds (
`test_paths_differ_survives_quotepath_false` and
`test_the_pathspec_batch_sizer_measures_bytes_not_characters` call it directly
with a non-ASCII path); at the **call site**, `differ = None` in
`landing_signals` → exactly the **two** named. The docstring now states the
placement and both counts. The conclusion is untouched — the test stays green
under both placements, which is the whole point.

Same class as round 2's F1 (a measurement stated without its method), so the
method is now attached to this one too.

### Gates — deliberately light, as a decision

* **AST identity vs `e2164eca`, reported not asserted**:
  `scripts/tests/test_worktree_prune.py` → **IDENTICAL** (docstrings blanked);
  `scripts/worktree-prune` **not touched this round at all**. Controls both ways
  — a file against itself → IDENTICAL, `origin/main`'s tool against this one →
  DIFFERS. So zero executable change, proven.
* **Targeted tests**: `test_worktree_prune.py` **220 passed**,
  `test_doc_path_rot.py` **77 passed**, `test_rules_size.py` **5 passed**,
  `test_transcript_search.py` **35 passed** (added — see below).
* **No full battery, no full suite, no merged-tree re-run.** That is a decision,
  not an omission: nothing executable changed, so the battery would re-derive a
  result already held at `02d2b811` over an AST-identical tool, and the
  merged-tree gate would re-measure a delta against a base that is red for two
  unrelated reasons.
* **Ledger**: `round 3 · payload: 0 · scaffolding: 14 · raw: 14` (11 added /
  3 deleted) — rc 0, stderr 0 bytes, 1 commit in range. The README clause is
  prose in a doc file and the test-file change is docstring-only, so payload is
  zero on both.

### ⚠ One correction to the round-3 brief's premise, measured

The two named doc-lint gates are **structurally blind to `scripts/README.md`**:
`test_rules_size.py` measures only `claude/RULES.md` and `claude/RULES-ARCHIVE.md`
(zero README references), and `test_doc_path_rot.py`'s corpus is
`CORPUS_DIRS = ("claude", "CLAUDE.md")` — its own comments exclude `README.md`
explicitly. Their green is a true statement about other files and says nothing
about this edit. I ran them anyway and report them as such, and added
`test_transcript_search.py`, which is the only test that actually opens
`scripts/README.md`. I also had to repair my own search instrument twice before
trusting any of this — `xargs`/`find -exec` cannot exec the shell's `grep`
wrapper function and both returned a confident zero for a token that
demonstrably exists.

No auto-closing keyword for #975 (re-verified; issue still `OPEN`). #973's branch
untouched.

ZacxDev:
## Adversarial audit — round 3 (delta) — 🔴 THE LADDER CLOSES HERE

Round 3 ran against `02d2b811..e2164eca` and dispositioned **all 7 of round 2's claims as actually fixed**, each re-measured rather than read — including rebuilding the rejected pair guard to confirm it is the unique killer of nothing. It returned one 🟡 and one 🟢, both prose. Addressed in `0c7cc076`.

`round 3 · payload lines changed THIS round: 0 · scaffolding: 14 · raw: 14` — rc 0, stderr 0 bytes, 1 commit in range.

### 🔴 The attribution gate has FIRED — two consecutive zero-payload fix rounds

| round | fixes' payload lines |
|---|---|
| 1 | **96** |
| 2 | **0** |
| 3 | **0** |

Per the standing rule, the ladder has left the PR and **stops**. There is no round 4, and ending it is the correct outcome rather than a failure.

🔴 **A raw line count would never have fired this gate.** Rounds 2 and 3 each touch `scripts/worktree-prune` or a payload file, so the unclassified numbers read 7 and 14. The classification is proven by parsing both revisions with docstrings blanked (comments never enter the AST) and comparing dumps, controlled both ways:

```
02d2b811..e2164eca  scripts/worktree-prune  -> IDENTICAL  (round 2, under test)
e2164eca..0c7cc076  test_worktree_prune.py  -> IDENTICAL  (round 3, under test)
origin/main..02d2b811  scripts/worktree-prune  -> DIFFERS  (positive control)
<rev>..<same rev>                              -> IDENTICAL (negative control)
```

```audit-claims round=3 audited=e2164eca9bf732ef2ad4ef3e7fc14817dd4595f0..0c7cc076
1. `scripts/README.md`'s worktree-prune row no longer claims the ASCII/non-ASCII twin pair is the control a signal-silencing fix would fail — the clause is replaced with the measured statement and names the two tests that do go red.
2. That row now says explicitly not to trim `test_content_identical_still_fires_for_a_landed_nonascii_path` or `test_squash_merged_branch_is_dead_not_orphan` on the twins' strength.
3. `test_an_ascii_only_branch_is_still_classified_correctly`'s docstring now states WHERE the "_paths_differ returns None" probe is applied and carries both counts — 4 reds applied inside `_paths_differ`, exactly the 2 named applied at the call site in `landing_signals`.
4. Nothing executable changed this round: `test_worktree_prune.py`'s AST with docstrings blanked is identical to `e2164eca`, and `scripts/worktree-prune` was not touched at all.
```

### 🔴 The finding worth carrying out of this PR: nothing could have caught it but a person

The 🟡 was a claim **this PR itself introduced** — `scripts/README.md`'s row asserted the twin pair "is the control that a fix which merely silences the signal would fail", while this PR's own test docstring 200 lines away refutes it in as many words (*"🔴 IT IS NOT THE CONTROL AGAINST 'SILENCE THE SIGNAL EVERYWHERE' … MEASURED: this test PASSES"*). Attributed independently: **0** occurrences at `origin/main`, **1** from `c06bda00` onward, surviving rounds 1 and 2 unchanged. Round 2 made it sharper by having `_git_raw` cite this README as corrected.

**Why it survived two audit rounds and every gate: no test opens `scripts/README.md` for this row.** The suite's four `README.md` references are a fixture file inside a temp repo plus two prose comments; the only test in the repo that opens the real file does so for an unrelated file-count claim. A human reading it was the only detector that could exist.

The cost was not cosmetic. A maintainer trimming tests would read that row as the authority on which fixtures guard the silencing direction, delete the real control, and leave `content-identical` permanently unreachable for non-ASCII paths — every squash-merged worktree silently downgrading `dead`→`orphan`, with nothing red. That is the defect class this PR exists to close, reproduced in the operator-facing doc for a tool that deletes directories.

### A wrong gate premise, corrected by the implementer rather than obeyed

The brief for this round named `test_doc_path_rot.py` and `test_rules_size.py` as the gates covering `scripts/README.md`. **Both are structurally blind to it** — verified here with a positive control (109 files under `scripts/tests` contain `read_text`, so the search instrument reads): `test_rules_size.py` mentions `README.md` **0** times and measures only `claude/RULES.md` / `claude/RULES-ARCHIVE.md`; `test_doc_path_rot.py`'s `CORPUS_DIRS = ("claude", "CLAUDE.md")`. Their green is a true statement about other files. `test_transcript_search.py` — the one test that opens the file — was added to the run instead.

### Gates for this round — a decision, stated as one

`test_worktree_prune.py` **220**, `test_doc_path_rot.py` **77**, `test_rules_size.py` **5**, `test_transcript_search.py` **35** — all passed.

**No full battery, no full suite, no merged-tree re-run.** Not "not needed": nothing executable changed, so the battery would re-derive a held result over an AST-identical tool, and the merged-tree gate would re-measure a delta against a base that is red for two unrelated reasons. The last full 99/99 sweep was at `02d2b811`, whose tool is AST-identical to this head.

### Still unverified, stated rather than implied

- **CI on this head was still running when this was written.** The two prior heads were green with identical executable content; that is an argument, not a measurement of this head.
- **No CPython 3.13 run** — no such interpreter on this host. 3.12 is the side on which the interpreter-conditional mutant runs, so sweeps are complete (`0 UNSCORED`), but the 3.13 skip path is untested, not green.
- **The #973 test-merge specifics are stale** — measured before rounds 1–3, all of which rewrote the same `_git_raw` docstring. The structural finding (prose/adjacency only, no code conflict, two mutually falsifying docstrings) is unchanged; the conflict text is not. Whoever merges second must hand-write that paragraph.
- **`origin/main` is red on the dev-host tier for two unrelated reasons** (host `opencode` 1.18.21 vs repo-keyed 1.18.18; a live-data espanso test). This branch adds **zero** new failures — identical failure sets, `comm -13` empty. Reported as a zero delta against a red base, never as a pass.
- `#975` carries **no auto-closing keyword** — its closing condition is a post-merge measurement. Re-verified on the live body; the issue is still `OPEN`.

ZacxDev:
## ⚠ The red checks on `0c7cc076` are a WEDGED GATE, not a test failure — do not read them as a verdict on this diff

Both required checks show `fail`. They are `error` state, not `failure`, and the description
is `COULD NOT RUN: <leg> — the gate stopped before this leg reported`. Attributed on the
cluster rather than inferred:

**The gate pod was never scheduled at all.**

```
TaskRun devrc-ci-x47z5-gate: TaskRunTimeout — failed to finish within "45m0s"
  step clone       exit=None reason=None finished=None
  step capture-etc exit=None reason=None finished=None
  step seed-nix    exit=None reason=None finished=None
  step pytests     exit=None reason=None finished=None
  step nodetests   exit=None reason=None finished=None
  step verdict     exit=None reason=None finished=None

pod devrc-ci-x47z5-gate-pod — FailedScheduling:
  0/4 nodes are available: 1 Insufficient cpu, 3 node(s) didn't match Pod's node
  affinity/selector. preemption: 0/4 — No preemption victims found for incoming pod
```

Not one step ever terminated, so **not one step printed a verdict**. The `tekton` skill's own
discriminator for this exact signature: *a step that emitted `RESULT:` / `<leg> verdict=`
failed a test; one that emitted neither was killed.* Neither leg emitted anything. The
`finally` reporter then did its job and posted `COULD NOT RUN`.

**It is cluster-wide, not this branch.** Measured at the same moment:

* pinned node `talos-xr6-r7p`: **94% CPU requests / 372% limits**, 57% memory
* **19 pods Pending** (unscheduled) in `tekton-ci`
* `FailedScheduling` currently hitting `auditloop-ci`, six `clawgate-ci` runs and three
  `clawgate-ux-audit` runs — every pinned pipeline, not just this one
* 36 `devrc-ci` gate pods resident; 7 PipelineRuns Running concurrently, 2 Cancelled

This is the documented congestion mode (`tekton` skill, gotcha #3): both CI pipelines
`nodeSelector`-pin to the same single node out of four, so a burst queues, and *"queueing
costs a verdict whenever the wait eats the budget"* — here the 45m budget expired with the
pod still unscheduled.

### What this PR's code evidence actually is

`0c7cc076` changes **zero executable lines** — AST with docstrings blanked is IDENTICAL to
`e2164eca` for the one touched `.py`, and `scripts/worktree-prune` was not touched. The two
prior heads carrying that identical executable content both went **green** on this same gate:

```
e2164eca  devrc-pytests pass — collected=18288 passed=18286 skipped=2 failed=0
02d2b811  devrc-pytests pass — collected=18288 passed=18286 skipped=2 failed=0
          devrc-nodetests pass — suites=5 files=39 tests=1300 pass=1300 fail=0
```

Plus locally on this exact tree: `test_worktree_prune.py` 220, `test_doc_path_rot.py` 77,
`test_rules_size.py` 5, `test_transcript_search.py` 35 — all passed.

### 🔴 It still BLOCKS the merge, and I have deliberately NOT forced it

`devrc` requires both checks and `enforce_admins` is on, so an `ERROR` check leaves this
`BLOCKED`. Re-running needs either a new push or a hand-created PipelineRun — and the skill is
explicit that pushing into a saturated queue is *one blast-radius action*, not a private one:
it kills **other people's** in-flight checks too. So I am not re-triggering into a node at 94%
CPU with 19 pods already pending.

**Re-trigger once the queue drains** (an empty commit or a rebase-push is enough), or use the
documented escape hatch if it stays wedged. Whoever picks the merge up: this needs a green
re-run, and the red you see now is not something to debug against the diff.

PR #1010, opened 2026-08-29
fix(discord-embed-ext): restore the validated engine, add the unclip half

ZacxDev:
Restores the browser-validated discord-embed-ext engine and adds the one thing it genuinely lacked — ancestor unclipping.

## What was wrong

`0e1db9e6` ("replace subagent version with validated code") did the opposite of its message. Its extension tree is **byte-identical to `origin/rescue/discord-embed-ext-concurrent`** — the rebuild the subsystem index recorded as having *lost* — and it replaced `eaf68c96` (#804), the version that was validated live in Brave. Three of its four code claims are false against #804 as landed:

| claim in `0e1db9e6` | reality |
|---|---|
| "aggressive `div:has()` rules broke layout" | #804 contains **zero** `:has(` and injects no stylesheet at all. v0.2.3 is the version with `:has()` rules. |
| "MEDIA_URL_RE matched emoji/sticker CDN paths" | #804's regex is `/(attachments\|external)/` — matches neither, nor `/avatars/`. Verified in node. |
| "used `external/` instead of `embeds/`" | Backwards. `/external/` is Discord's proxy for linked media and **is** message content (measured 2026-08-24). `/embeds/` matches no real Discord media path, so v0.2.3 silently stopped enlarging every externally-linked embed. |
| "container `overflow:hidden` clipped enlarged media" | **True** — the one real gap in #804, and what this PR adds. |

Net: to fix one real bug, the swap discarded seven fixed ones and ~1400 lines of tests. v0.2.3 also reintroduced defects #804 had documented with 🔴 comments — the observer permanently disconnecting itself when a mutation batch contained no new media, mutation batches being discarded by the debounce, a host-only media pattern in the lightbox (which paged navigation through avatars), and five lightbox buttons created but never wired to a click handler.

## What this does

- Restores `eaf68c96`'s engine, tests, README and icons.
- Adds `unclipAncestors()` / `reclipAncestors()` — clears **every** clipping ancestor, not just the first.
  - `auto` / `scroll` are never touched. They are scroll containers; forcing them to `visible` is what broke Discord's scroller in v0.2.1–0.2.3 and produced the re-render cascade.
  - The walk stops at the message row, so Discord's chrome is out of reach **by construction** rather than by a depth guess. A rotted boundary heuristic degrades to the old bounded 8-ancestor walk, never an unbounded climb to `<html>`.
  - Every clear records the prior inline `overflow` + priority; `forget()` restores it exactly, closing the "there is no undo" gap #804's own header comment flagged.
- Moves `findMessageContainer` out of `lightbox.js` into `embed_enlarge.js` — two callers now need that boundary, and a second copy is how one gets fixed on a Discord reshuffle while the other keeps the rotted heuristic.
- Reads all ancestors before writing any, instead of interleaving `getComputedStyle` with style writes (a layout recalc per ancestor, up to 8 per image).

## Removed as dead

- `save_button.js` — dl-router requires a bearer token on **every** endpoint including `/healthz`, so the probe 401s, the observer never starts and the button never mounts; `POST /match` is classify-only, so `"Saved!"` reported a save that never happened. #804 removed it for exactly this; `0e1db9e6` put it back.
- `service_worker.js` — a 0-byte file declared as `background.service_worker`.
- `tests/fixtures/discord_embeds.html` — the losing implementation's duplicate of `tests/discord_embeds.html`.

## Tests: 68 → 134

Nothing in the v0.2.3 suite covered the clipping path — no test in the tree mentioned `overflow`, `clearParentConstraints` or `data-dee-cleared` while four consecutive versions were built around that mechanism.

- **RED at `eaf68c96`** (new tests + new harness, old extension): 16/16 fail.
- **GREEN at HEAD**: 134/134.
- **Mutation sweep, 10 mutants**, each asserted to kill its *own named test* rather than merely turning something red: **10/10 KILLED**.

One mutant **SURVIVED** the first sweep and the fixture was fixed, not the score: deleting the message-boundary break changed nothing, because the only ancestor above the row was the `auto` scroller that `clipsOverflow` already rejects — the guard was **unreachable**. The rig now carries chrome with `overflow: hidden` above the row, which is both the realistic Discord shape and the observable one.

Harness fixes that made the above non-vacuous: `FakeComputedStyle` gained `getPropertyPriority` (its absence made the extension's priority round-trip take the else-branch in every test), and the computed-style seam now expands the `overflow` shorthand to longhands the way a real `getComputedStyle` does — without it every unclip test would have passed against a code path that never fires in Brave.

## Gate

Run on the **merged** tree (`1b85eaea`), both tiers:

- **Sandbox tier** (`nix build .#checks.x86_64-linux.nodetests`, what Tekton gates on): `RESULT: PASS`, TOTAL 1366/1366, `scripts/discord-embed-ext/tests files=2 tests=134 pass=134 floor=128`.
- **Dev-host tier**: node 1366/1366 pass. pytest has 2 failures, both reproduced at the unmodified branch tip in a clean worktree and neither in a file this touches — `test_opencode_engine` (opencode on PATH is 1.18.21, pinned 1.18.18) and `test_espanso_detect` (`:acq`/`:dacq` search-term collision).

## Merge note

`origin/main` had acquired v0.2.3 wholesale: `be1585c3` ("feat/gradient rgb", #1013) was branched off this branch's `5dd4fc99` and carried the whole discord-embed-ext delta along with the rig-control work. Conflicts are resolved to this branch's side in full; main's other five commits merge cleanly and are kept. Verified on the merged tree that the deletions stayed deleted — a merge is exactly where a deletion gets silently undone.

## ⚠️ Not verified in Brave

The extension is not injecting in the profile the browser-bridge can reach, and the profile holding Discord has a stale bridge build that cannot inject into `discord.com`. Live confirmation that the crop is gone is still outstanding and needs: `home-manager switch` → reload at `brave://extensions` → open a channel with image attachments.


ZacxDev:
## Gate status — `devrc-pytests` is red on `main` too, not from this PR

`tekton/devrc-nodetests` → **SUCCESS** (`suites=5 files=38 tests=1366 pass=1366 floor=1317`).
`tekton/devrc-pytests` → **FAILURE** (`collected=18342 passed=18338 skipped=2 fail=2`).

Reproduced locally, and then ran the control — the same `nix build .#checks.x86_64-linux.pytests` from `origin/main`:

| tree | result |
|---|---|
| `origin/main` | **FAIL** — `passed=18337 failed=3` |
| this PR (merged) | **FAIL** — `passed=18338 failed=2` |

So this branch is strictly *better* than its base on that check, and cannot turn it green.

The two failures, both with **every input byte-identical to `main`** (verified per-file, and `flake.lock` is identical too), neither in a file this PR touches:

- `scripts/tests/test_opencode_engine.py::test_engine_is_the_version_every_measurement_is_keyed_to` — opencode resolves to `1.18.21`, `PINNED_VERSION` is `1.18.18`. The test explicitly says not to just bump the pin: re-derive the measurements against the new binary first.
- `scripts/collector/keylog/tests/test_espanso_detect.py::test_live_existing_resolutions_not_made_ambiguous` — search term `ask` now matches both `:dacq` and `:acq`, so attribution returns `None`.

`main`'s third failure, `TestSkillDocsArePinned::test_the_pinned_docs_are_the_DEPLOYED_ones`, is a nested-nix-eval contention (`SQLite database … is busy`) that appears only when another `nix build` runs concurrently — it is not a property of either tree.

⚠️ `required_status_checks.contexts` is `["tekton/devrc-pytests","tekton/devrc-nodetests"]` with `enforce_admins: true`, so **nothing can merge until those two are fixed** — this PR included. They belong to the opencode-pin and espanso efforts, not to this one, so I have not widened this PR to cover them.

PR #1022, opened 2026-08-29
fix(opencode): re-derive the version pin 1.18.18 → 1.18.21 — flake.lock moved the binary under an unchanged config

ZacxDev:
`main` is red repo-wide on `test_engine_is_the_version_every_measurement_is_keyed_to`.
**That red is the pin working.** `test_opencode_engine.py` exists precisely to catch a
drifted BINARY under an unchanged config, and `flake.lock` moved opencode.

Second of the two deterministic failures blocking every PR (the other is #1021).

## Ruled out first
The test names per-host imperative-profile drift as the usual cause. It is **not** that:
`nix profile list` carries no opencode entry, and PATH resolves into the flake's own
store path. A genuine lock movement.

## Re-derived, not re-spelled
The test's instruction is explicit — *"do NOT just bump PINNED_VERSION: re-derive the
header's measurements against the new binary first"*.

- **Whole engine file, NEW binary, OLD pin: 1 failed (exactly the version assertion),
  24 passed.** Every other engine claim — conformance tests, resolved tool maps, ordered
  permission arrays — holds on 1.18.21, and the harness is shown to **discriminate**
  rather than be green by default. Same control the 1.18.4 → 1.18.16 pass used.
- **Both hosts measured at the consumer**: workbench and laptop each resolve
  `readlink -f $(command -v opencode)` to the **same** store path
  `…-opencode-1.18.21`. So "both hosts run \<v\>" claims are re-keyed on measurement,
  not relabelled.

## 30 re-keyed, 8 deliberately not
Re-keying was **line-targeted and asserted** — each edit names the flagged `file:line`
and refuses if it no longer carries the old version. A whole-file replace would also
have rewritten the historical lines, which is the one thing the ledger forbids.

Not re-keyed, each with a new `HISTORICAL_VERSION_CLAIMS` entry and its reason:
`TaskTool.execute` child-session internals (not observable from `debug agent --pure`),
two dated COST measurements against a specific engine+model pair, one dated narrative
of a past pin move, and the two bump-transition records. Re-keying those would have
relabelled measurements nobody repeated.

## 🔴 Two traps the ledger caught in my own first draft
Both are now documented in place, because both are non-obvious:

1. A ledger **comment** containing a version literal is itself a claim in the pin
   surface — so my explanatory comment failed the very test it was explaining.
2. A ledger **snippet** containing a version literal makes the entry's own source line
   an old-version line, so it matches twice and fails. That is why every pre-existing
   snippet is version-free; mine now are too.

## Verified
```
engine  25 passed
config 640 passed        (665 total)
nix-instantiate --parse  OK on flake.nix and nix/pkgs/tools/default.nix
```

### DIFF
--- a/claude/opencode-addendum.md
+++ b/claude/opencode-addendum.md
@@ -32,7 +32,7 @@ bash guard.
 | search file contents | `grep` | `rg` / `grep` via bash |
 | see what is in a dir | `glob` on `<dir>/*` | `ls` |
 
-There is **no `list` tool** on opencode 1.18.18 — verified against the resolved
+There is **no `list` tool** on opencode 1.18.21 — verified against the resolved
 tool map, which contains exactly `bash, edit, glob, grep, invalid, question,
 read, skill, task, todowrite, webfetch, write`. Use `glob` to enumerate a
 directory.
--- a/flake.nix
+++ b/flake.nix
@@ -292,11 +292,11 @@
             # 🔴 This ALSO makes the sandbox the tier that pins the VERSION.
             # `pkgs.opencode` here and in nix/pkgs/tools/default.nix resolve from
             # the same flake.lock, so CI tests the exact binary the hosts deploy
-            # (1.18.18 at rev 5c680dac9f02). Cost, stated as deliberately as the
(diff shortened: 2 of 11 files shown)


ZacxDev:

PR #1024, opened 2026-08-29
fix: unblock the required pytest gate — it is red on main for every PR

ZacxDev:
`tekton/devrc-pytests` is a **required** check with `enforce_admins: true`, and it has been failing on `main` itself — so nothing in the repo could merge, not just the PR that noticed.

**Measured**, not inferred: `nix build .#checks.x86_64-linux.pytests` from `origin/main` → **FAIL**, `passed=18337 failed=3`.

| tree | sandbox pytests |
|---|---|
| `origin/main` | FAIL — `passed=18337 failed=3` |
| this branch | **PASS** — `passed=18340 failed=0` |

Both previously-red suites now pass: `scripts/tests 9819/9819`, `scripts/collector/keylog/tests 93/93`. `nodetests` also green (`1300/1300`).

Two independent causes.

## 1. espanso — `ask` went ambiguous again

`#979` shipped the `:acq`/`:dacq` split with `"ask"` deliberately **not** in `:dacq`'s `search_terms`. `f04bb5c6` ("espanso") added it back. `_token_matches` reads `search_terms`, so `ask` matched both snippets, `_attribute` returned `None`, and the highest-traffic term in the whole config (58 fires in the 2026-08-06..19 window) went **UNATTRIBUTED**.

`nix/home.nix`'s own comment two lines above already stated the contract — *":acq owns ask/clarify/questions alone"* — the code just stopped honouring it.

Fixed in the **config**, not the pinned table, which is what the test's own ANTI-VACUITY note requires: *"had the term merely gone ambiguous, the fix would be the config"*. Verified the test reads this tree (`HOME_NIX` resolves to the worktree copy, and the `:dacq` line no longer contains `ask`), and that its positive control `test_live_scraper_observes_the_real_config` is green.

## 2. opencode — the lock moved the binary to 1.18.21 without re-keying anything

Same shape as #570 and the 2026-08-13 bump. The pin's own message forbids the one-line fix, so the claims were **re-derived first**, with controls reported as pairs:

- `debug agent <a> --pure` for **all seven** agents under **both** binaries: **7/7 IDENTICAL** — permission array, resolved tool map, model, description, prompt, mode. Cleaner than the 08-19 bump, which found `compaction`'s built-in prompt had moved.
- **Same-binary control**, which the file demands before any cross-version verdict is believed: two runs of the new binary collapse to identical on all seven agents. The readdir nondeterminism that once made two same-binary runs differ on 434 lines does not arise under this harness's env — `HOME` is an empty temp dir, so there are no skills directories to walk.
- **Negative control** on the comparison itself: `nav` vs `k8s` on one binary reports DIFFERS on seven top-level keys — so "IDENTICAL" is not all it can say.
- **Positive control** on the captures: 10 top-level keys, 12 resolved tools, 92–119 permission entries per agent, no error markers.
- **Discrimination, both directions**: the file against the current binary *before* re-keying → 1 failed (exactly the version assertion); against the superseded binary *after* re-keying → 1 failed, the same one.

Then the pin was re-keyed and **27 claim lines renumbered across 11 files**. Deployed-state lines are renumbered on a measurement, not on faith: both hosts resolve `…-opencode-1.18.21`, the same store path the flake provides (2026-08-29, `readlink -f $(command -v opencode)` on workbench and laptop).

### Deliberately NOT renumbered

Renumbering an un-re-derived claim turns a dated measurement into a false one that reads like a fresh reading. These are ledgered as historical instead:

- the two dated **cost** measurements — token/spend figures were not re-run;
- the three **`TaskTool.execute` internals** claims — the differential covered agent resolution, not that.

Every new ledger snippet is **version-free**, per the file's own note: a snippet quoting a literal matches itself when this file is scanned.

One renumbered line needed more than a number. The browser-bridge *"both hosts run … and both resolve browser-only"* claim bundled a host reading with a per-host gate replication; only the first was re-taken. They are now stated separately, each with its own date and the reason the second carries forward.

## Why this is its own PR

Neither fix belongs to the discord-embed-ext work that surfaced them (#1010), and unblocking the gate unblocks every open PR, not one.

### DIFF
--- a/claude/opencode-addendum.md
+++ b/claude/opencode-addendum.md
@@ -32,7 +32,7 @@ bash guard.
 | search file contents | `grep` | `rg` / `grep` via bash |
 | see what is in a dir | `glob` on `<dir>/*` | `ls` |
 
-There is **no `list` tool** on opencode 1.18.18 — verified against the resolved
+There is **no `list` tool** on opencode 1.18.21 — verified against the resolved
 tool map, which contains exactly `bash, edit, glob, grep, invalid, question,
 read, skill, task, todowrite, webfetch, write`. Use `glob` to enumerate a
 directory.
--- a/flake.nix
+++ b/flake.nix
@@ -292,11 +292,11 @@
             # 🔴 This ALSO makes the sandbox the tier that pins the VERSION.
             # `pkgs.opencode` here and in nix/pkgs/tools/default.nix resolve from
             # the same flake.lock, so CI tests the exact binary the hosts deploy
-            # (1.18.18 at rev 5c680dac9f02). Cost, stated as deliberately as the
+            # (1.18.21 at rev c27cdad491a9). Cost, stated as deliberately as the
             # nodejs and nix entries above: this check's closure grows by
             # opencode, and a nixpkgs bump that moves it invalidates the cache
             # AND turns the version assertion red. That red is the point — the
-            # config header's "measured on v1.18.18 — do not re-derive" claims are
+            # config header's "measured on v1.18.21 — do not re-derive" claims are
             # otherwise pinned to nothing.
             #
             # logrotate: scripts/tests/test_claude_log_rotate.py drives the REAL
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -266,7 +266,14 @@ in
           # 'ask' ⊂ 'task' is the documented way a neighbour silently steals it.
           # :dacq keeps the words he actually types for THIS snippet (feedback,
           # dispatch, process); :acq owns ask/clarify/questions alone.
-          { trigger = ":dacq"; replace = "dispatch subagent to process feedback\nask clarifying questions and recommend improvements and anything useful to include before dispatching (include complete test coverage)"; label = "Process feedback: dispatch subagent + elicit scope"; search_terms = ["ask" "feedback" "dispatch" "process" "elicit" "scope" "include"]; }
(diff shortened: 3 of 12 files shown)


ZacxDev:
Closing as a duplicate of #1023, which is strictly better.

I built this without sweeping `gh pr list --state open` first — the exact shared-queue-with-no-lock failure RULES.md warns about. #1023 (plus #1021, #1022, #1015) already had all of it.

#1023 covers the same two causes **and** a third this PR does not: the `test_subsystem_store_api.py` hang-detectors timing out on scheduling, which is what failed *this* PR's own CI run (`TestTheActorComesFromTheTOKEN.test_the_SESSION_is_recorded_and_it_is_the_one_that_was_SENT`, a 15s HTTP timeout). Three different tests from that one file failed across three unrelated commits today — so without #1015/#1023's half, the gate would still be red intermittently.

Cross-checked the substance before closing; two independent derivations agree exactly:

| | #1023 | this PR |
|---|---|---|
| `:dacq` search_terms | `[feedback dispatch process elicit scope include]` | identical |
| `PINNED_VERSION` | `1.18.21` | identical |
| renumbered claims in opencode.jsonc / test_opencode_config.py / k8s.md / review.md / opencode-addendum.md | 4 / 6 / 1 / 1 / 1 | identical |

Nothing here is worth salvaging over #1023. Merging that one instead.

PR #1026, opened 2026-08-29
docs(handoff): rank 3 shipped correct and UNUSABLE — 0.8.11 makes the grid triage; plus a staged fix for the docker.io DNS blocker

ZacxDev:
Two commits: the handoff update, and the staged DNS fix that this session's blocker produced.

## Rank 3 is done, deployed, and now actually useful

`ZacxDev/homelab-infra#511` merged `396afa9d`, deployed as **clawgate 0.8.11**. The tmux grid now **triages** instead of merely rendering: waiting → parked draft → busy → idle → shell, with badge and sort from one `triageRank` so a promoted card always says why, and the waiting card shipping the **matched signal line** as evidence.

**Verified live:** 78 cards, **7 `WAITING ON YOU` + 5 `DRAFT UNSENT` at the top**, 30 shells last, **0** write-surface hits on the live page. Two surfaced drafts were a single word — `yes` and `check` — answers one keystroke from running.

## The lesson worth keeping

🔴 **0.8.10 was correct, fully tested, and unusable — and only opening it showed that.** It passed 15 tests and a mutation sweep and rendered every field honestly. On screen it was a **10,932px** page whose top-left card was an empty stretched shell, with the four windows blocked on a human buried inside it. All three defects are properties of the **whole page**, which no per-card unit test can observe:

- nothing was ranked;
- CSS grid `align-items: stretch` inflated **30 one-line shell cards** to match 49-line neighbours;
- 272-column panes were **clipped** in a ~530px card.

A comment can also defend a property the layout does not have — this one claimed wrapping would "destroy" column alignment that was never achievable at that width.

Second lesson: **a sort-stability mutant survived two fixtures that could not partition.** `SliceStable`→`Slice` survived 6 equal-rank items (pdqsort falls back to insertion sort below ~12, which *is* stable) and survived 40 items of the *same* rank (pdqsort's all-equal fast path moves nothing — genuinely equivalent). Only interleaved ranks at n ≥ 20 discriminate, measured with a throwaway program rather than guessed a third time.

## New: an Open-investigations block with full diagnosis

**The workbench cannot pull from docker.io.** The LAN router serves `registry-1.docker.io` at a **487-day TTL** over a stale 8-address set — 4 correct, 2 now hosting unrelated third-party sites, 2 dead. Every value, every ruled-out hypothesis, and a verbatim next probe are in the doc so nobody re-derives it.

🔴 I first called this **TLS interception**. It isn't, and the giveaway was that the wrong domain *changed between attempts*. The correction is recorded as a Gotcha, because the diagnostic move that settles it — test the **answer set**, not the connection — generalises.

Retrying does not route around it: the bad addresses are live web servers, so Go's dialer completes TCP, fails TLS, and never tries the next address (12/12 failed).

## The second commit: `nix/system/apply-dnsmasq-docker-io-pin.sh`

Staged, **not applied**, per this repo's `nix/system/` convention — I never run `nixos-rebuild`. It adds `"/docker.io/1.1.1.1"` ahead of the existing servers; domain-specific entries win, so every other name still asks the router first (deliberate — it bypasses the VPN for LAN names). Timestamped backup, and it **refuses rather than guessing** if the servers line has drifted.

⚠ It was run once: the edit landed and **generation 385 built with the pin in it**, but `switch-to-configuration switch` correctly **refused** — `dbus → dbus-broker`, ten days of nixpkgs riding behind a one-line edit. That wants `nixos-rebuild boot` + a reboot, not `NIXOS_NO_CHECK=1` on a live desktop. `alwaysKeepRunning = true` also means activation won't bounce dnsmasq, so a restart is separately required.

🔴 **The root cause is the router and this doesn't touch it.** Clearing that entry on `192.168.50.1` is the actual repair; the pin only protects this host. The two third-party domain names are deliberately omitted from the file — this repo is public.

## How 0.8.11 got built at all

On the **laptop** (27s TTL, healthy) via `DOCKER_HOST=ssh://zach@10.42.0.100`, which needs no privilege and inverts `clawgate/reference/deploy.md`'s usual direction. TLS verification was never disabled anywhere. That recipe is now in the doc's Status and How-to-verify.

Ranks 1–8 and 10 unchanged, no renumbering. Durable Status content carried forward rather than dropped — the design history, the audit-round record, the five-month-old-hook finding — after the tool's `DROPS N DURABLE` warning flagged 14 and I got it to 10, all genuinely superseded status.

https://claude.ai/code/session_015hFhuL3EQnGXDhcmG7dtfM

### DIFF
--- a/claudedocs/handoff-tmux-webapp.md
+++ b/claudedocs/handoff-tmux-webapp.md
@@ -10,86 +10,61 @@ and an **attention queue** that surfaces sessions needing a human so Zach can ju
 ## Status
 
 **Phase 1 (attention queue) SHIPPED. Phase 2 (tmux read model) COMPLETE. Rank 3 (read-only
-`capture-pane` rendering) is ✅ DONE — BOTH HALVES SHIPPED, DEPLOYED AND VERIFIED LIVE
-2026-08-28. Phases 4–6 untouched; the lowest-numbered OPEN item is now rank 5.**
+`capture-pane` rendering) is ✅ DONE, DEPLOYED, and now IN REAL USE — clawgate **0.8.11** sorts
+what needs a human to the top of the grid. Phases 4–6 untouched; lowest-numbered OPEN item is
+rank 5.**
 
 🔴 **DO NOT READ A VERSION FROM THIS DOC** — `clawgatectl health` is the only authority. It said
-**0.8.10** on 2026-08-28 immediately after this session deployed it. This line has now carried
-**four** values in two days (0.8.7 → 0.8.8 → 0.8.9 from a concurrent session → 0.8.10), which is
-the whole argument for reading the cluster instead of this paragraph.
+**0.8.11** on 2026-08-29. This line has now carried five values in three days.
 
-⚠ **THE PREVIOUS REVISION OF THIS HEADER SAID "MERGED-BUT-NOT-DEPLOYED" AND WARNED THAT THE
-FIRST SIGHT WOULD BE `collector predates the field` ON EVERY CARD. Both were true when written
-and are now FALSE** — the deploy landed within the hour, and because the DATA half shipped first
-the tab rendered real screens immediately. The warning's *reasoning* stands and is why the status
-vocabulary exists; its *reading* does not. Third time this header has gone stale inside a day.
+🔴 **THE WORKBENCH CANNOT BUILD A CONTAINER RIGHT NOW — BUILD ON THE LAPTOP.** Its resolver
+returns a poisoned `registry-1.docker.io`; see **Open investigations** before you try, because
+the failure reads as TLS interception and is not. The working recipe, used end to end for 0.8.11:
 
-**Live at handoff (2026-08-28, measured not assumed):** pod `clawgate-fcd8c945b-44rqj`
-Running/ready, `restarts=0`, image `0.8.10`; `clawgatectl --version` **0.8.10 on BOTH hosts** (no
-skew note — the `buildVersion` half of the bump landed). `GET /ui/tmux` → **200, 208,347 bytes**,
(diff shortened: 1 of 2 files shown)


ZacxDev:
Closing as **superseded**, at the operator's direction — and recording what was checked first, because a stale doc PR is unsaved work until someone proves otherwise.

**205 commits behind `main`.** Opened 2026-08-29, `created == updated` — never touched since. Its doc is 972 lines against `main`'s 2078, and it describes rank 3 in the 0.8.11 era; rank 3 has been ✅ DONE since 2026-08-28 and the feature's whole queue is now closed except two CI-infrastructure items.

**Nothing unique is lost. Verified rather than assumed:**

- `nix/system/apply-dnsmasq-docker-io-pin.sh` — present on `origin/main` and **byte-identical** (sha `d92931093894a9e9` both sides). This was the one real artifact at risk; the diagnosis it carries had previously lived only in an untracked file.
- The **laptop build/push recipe** survives on `main`, in its *corrected* form — including the second credential leg (`docker push` sends registry auth from the LOCAL client, so the push must run ON the laptop), which this PR predates.
- The **deploy-runbook inversion** (`DOCKER_HOST=ssh://zach@192.168.50.250` is host-dependent) survives on `main`.
- The **docker.io root cause** (the router's pinned 487-day TTL record) survives on `main`.

The genuinely unique remainder is 0.8.11-era **status**: version claims, pod name, badge counts, page heights. All superseded, and the doc's own rule is that a version must never be read from it — `clawgatectl health` is the only authority.

**The branch `docs/handoff-tmux-webapp-0811` is deliberately NOT deleted**, so the content stays recoverable if any of the above turns out to be wrong.

It would also now conflict: `main`'s copy of this doc has been rewritten several times over, most recently by #1230 and #1238.

PR #1035, opened 2026-08-29
fix(opencode): the pin's rev claim names the rev that produced the previous version

ZacxDev:
One comment line. `#1023` re-derived the opencode pin to 1.18.21 and updated the store path, but left the **nixpkgs rev** pointing at the one that produces 1.18.18.

The sentence is present tense, and its two halves now disagree with each other:

> `MEASURED at flake.lock's nixpkgs rev `**`5c680dac9f02`**`, `pkgs.opencode` is `**`1.18.21`**` — store path /nix/store/iqc8xfx…-opencode-`**`1.18.21`**`.`

**Measured:** `origin/main`'s `flake.lock` pins nixpkgs at **`c27cdad491a9`**, and the store path quoted in that same sentence is the one *that* rev produces. So the rev is the only stale half.

## Why nothing caught it

This is the exact class `nix/pkgs/tools/default.nix`'s own header records from an earlier bump:

> *"both asserted in the PRESENT TENSE that `pkgs.opencode` **is** 1.18.4 at a nixpkgs rev the lock had already left behind"*

`test_no_file_in_the_pins_surface_still_claims_an_OLD_version` keys on **version literals**, and a git rev is not one — so a stale rev is invisible to the gate that exists to catch stale pin claims. That is why it survived a re-derivation which got the version, the store path and 36 claim lines right.

## Verification

`test_opencode_engine.py` + `test_opencode_config.py`: **665 passed**. No behaviour change.

## Provenance

Found while independently re-deriving the same pin bump before noticing #1023 had already landed it. The duplicate effort is mine to own — I did not `claim-work` the item or sweep open PRs first. This residual is real and was not in that PR.


ZacxDev:
**This PR's content is correct — I verified its premise independently — and its red check is the known load flake, not this diff.**

## The premise still holds on `main` right now
```
$ git show origin/main:nix/pkgs/tools/default.nix | sed -n '22,23p'
  # This entry pins them: MEASURED at flake.lock's nixpkgs rev 5c680dac9f02,
  # `pkgs.opencode` is 1.18.21 — store path

$ git show origin/main:flake.lock | jq -r .nodes.nixpkgs.locked.rev | cut -c1-12
c27cdad491a9
```
One sentence, two halves, disagreeing. The rev is the only stale one — version and store path on the next lines are right.

## This is my defect, from #1023, and the root cause generalises
I re-keyed the pin surface by editing **exactly the lines `test_no_file_in_the_pins_surface_still_claims_an_OLD_version` flagged** — and that ledger matches **version literals only**. Line 22 carries a *rev* and no version, so it was structurally invisible to the detector and therefore invisible to a fix driven by it.

🔴 **A detector-driven fix inherits the detector's blind spot.** The method was sound; its blind spot was inherited whole. Recorded in the `devrc/opencode` index entry so the next pin bump asks what the ledger *cannot* see before treating its output as the work list.

## Why it is red
`tekton/devrc-pytests` failed on `TestTheActorComesFromTheTOKEN` — the socket-timeout load flake, unrelated to this one-line comment change. Measured across retained runs today: **17 failed / 6 succeeded (26% pass rate), six failing runs on six different commits each failing a different single test**. #1023 raised the client bound 15 s → 60 s and it did **not** eliminate this: #1009 later failed *at the 60 s bound*.

**It needs a re-run, not a change.** I have deliberately not rebased or force-pushed this branch — it is another session's and was updated recently, and rewriting it would buy only a CI re-trigger. Whoever owns it: a rebase onto current `main` will re-trigger and should go green.

PR #1046, opened 2026-08-29
fix(handoff_doc): REFUSE when the doc is absent here and present on the mainline

ZacxDev:
## What this fixes

`handoff_doc.py` already **detected** the destructive case and printed it in plain language:

> 🔴 THE BASE DOCUMENT IS NOT THE NEWEST COMMITTED COPY — there is NO `claudedocs/handoff-<topic>.md` in this checkout at all, and origin/trunk has one (7 sections / 155 lines) — **every section will merge as NEW and the committed document will be replaced by this delta.**

…and then exited **0**. That was survivable while a human answered a y/N. That prompt was retired 2026-08-23, and the tool's own output now says the warnings "are the only thing between this diff and a pushed commit." The governing rule in the `subsystem-index` skill is **"blast radius earns a REFUSAL, not a question"** — replacing a committed document with a delta is blast radius.

**`--push`'s existing `EXIT_BEHIND` does not cover it.** That check asks whether `<remote>/<push-branch>` has moved — a *different ref* from the derived mainline this compares against — and `test_it_does_not_fetch_or_refuse_WITHOUT_push` pins that it deliberately never fires without `--push`. A feature branch current with its own upstream sails past it while the mainline copy of the doc is the one destroyed.

**Measured on the real case:** `civitai/talos-infra`, a primary clone 123 behind because the doc was authored and pushed from a **worktree** and the clone was never re-synced. A read-only run reported exactly the above at exit 0.

## The change

`EXIT_STALE_BASE` (7) / `status=stale-base`, with an explicit `--allow-replacing-mainline-doc` override. The predicate is one method on `BaseCurrency` so the rule lives in one place, and it is deliberately **narrower than `stale` in both directions**:

- doc **present** here but behind ⇒ still a WARNING, exit 0 (the merge can classify its sections; updating a knowingly-behind clone is legitimate)
- doc absent on **both** sides ⇒ still the ordinary NEW-doc case step 5 owns

## Evidence

**Matrix: 3 red at `origin/main` → 5 green at HEAD**, each red failing for its own reason (missing constant, unnamed override, unrecognised flag). The two must-not-break pins were green in **both** runs.

**Mutation sweep** (`PYTHONDONTWRITEBYTECODE=1`, with a null control):

| mutant | result |
|---|---|
| M0 null control (comment only) | **SURVIVE** — proves the sweep isn't reddening everything |
| M1 drop the absent-here half | killed by `test_a_doc_PRESENT_locally_but_behind_stays_a_WARNING` |
| M2 drop the mainline half | killed by `test_a_GENUINELY_NEW_doc_is_untouched_by_this` |
| M3 guard never fires | killed by `test_it_REFUSES_and_writes_nothing` |
| M4 ignore the override flag | killed by `test_the_explicit_override_still_lets_it_through` |

Each mutant dies to a **different** test, so each test pins a distinct property.

🔴 **M1 SURVIVED the first sweep, and the finding was about the test, not the code.** The must-not-break fixture used `advance_remote`, which pushes an unrelated `OTHER.md` — so `doc_behind` stayed 0, `mainline` was never read, and the case the assertion named *could not arise*. It read as coverage while providing none. The fixture now advances the **doc itself** and asserts that precondition before asserting the behaviour.

## Both tiers, same base

| tier | result |
|---|---|
| dev host (`run-tests.sh`) | `RESULT: PASS (exit=0)` — `collected=18564 passed=18562 failed=0` |
| nix sandbox (`checks.x86_64-linux.pytests`) | `RESULT: PASS (exit=0)` — same counts, floor 15970 |

Base `7f5474b7`. All three changed files are **tracked** and the **committed** tree carries the new class and constant — checked directly, because a green sandbox verdict is not evidence the tree contained the change (#928). Count moved on the dev-host tier: the file went 171 → 176 collected.

## SKILL.md was mandatory, not optional

Its base-currency paragraph asserted *"A WARNING, never a refusal and no exit code changes"* — which this makes false. Rewritten to carry both outcomes. The skill has a 23,100 B enforced budget with **3 B free**, so this is paid for by **eviction, not by raising the ceiling**: redundant clauses trimmed across steps 1, 2 and 5, no instruction removed, every pinned literal intact (the suite caught one compression and it was restored verbatim). 23,097 → 23,090 B.

⚠ `main` has moved 4 commits since this branched and `strict` is false, so the check is a claim about this branch, not the merged tree.

https://claude.ai/code/session_013Hg2q9rQJtaJUdy33FhU5Q


ZacxDev:
## Round 1 — `/audit-pr` findings and fixes

Round 1 returned six 🟡 findings plus four 🟢 nits. All fixed in `82a37ee6`.

Ledger: **round 1 · payload lines changed this round: 96** (`handoff_doc.py` 66+23, `SKILL.md` 7+7) · scaffolding: 66 test lines · the ladder is still on the PR, not on itself.

```audit-claims round=1 audited=f9b15a57..82a37ee6
1. `BaseCurrency.replaces_mainline_doc` now tests `not base_text.strip()` instead of `not base_text` — claimed to close the bypass where a doc that exists but is `"\n"` or `"   \n\n"` exited 0 and replaced the committed document.
2. `WRONG_BASE_REMEDY`, rule (h)'s header comment and the rule (h) incident block were rewritten — claimed no longer to assert "it warns and never refuses / no exit code changed" in a tool that now exits 7, and claimed no longer to argue against the guard this PR adds.
3. `build_parser()` now passes `allow_abbrev=False` — claimed to stop `--al`, `--allow` and `--allow-replacing` being accepted as `--allow-replacing-mainline-doc`.
4. Added `test_the_PROPOSAL_run_still_shows_the_diff_and_does_not_refuse` — claimed to pin the `and args.confirm` half of the guard, which a surviving mutant previously showed was unpinned across the whole file.
5. The `status=stale-base` message was rewritten — claimed to name re-syncing the clone as the fix, and to state that the override cannot rescue the `--push`-on-the-mainline-branch case because stale-base against your push branch implies behind.
6. `EXIT_STALE_BASE` was moved below `EXIT_BEHIND`'s docstring — claimed to restore `EXIT_BEHIND`'s orphaned 14-line attribute docstring.
7. The refusal headline now reads "no usable <path> — missing, empty or whitespace" — claimed to stop asserting "there is no <path>" for a file that exists and is blank.
8. `SKILL.md`'s exit-7 bullet now names its `--confirm` precondition — claimed to stop predicting exit 7 for a proposal run that returns 0.
9. `SKILL.md`'s `📖 write-gate §B` pointer was replaced with a pointer to rule (h) in `handoff_doc.py` — claimed to resolve, where `§B` did not (that file has no lettered sections).
10. `SKILL.md`'s CSP exemplar had `on GET /apps/run/dogfood-manual` restored — claimed to make the example demonstrate the "exact failing request" the instruction beside it demands.
11. `SKILL.md` was re-fitted to 23,097 B under the 23,100 B enforced budget by eviction in steps 1, 2, 5 and the kickoff block — claimed to remove no behavioural instruction.
```

### Round 1 also corrected a claim I made in the PR body

I reported mutant M1 as SURVIVED and attributed it to a vacuous fixture. That was wrong: `test_a_stale_base_still_exits_0_and_still_writes_under_confirm` already existed at the base sha and also kills M1, so M1 could not have survived a **full-file** sweep — my sweep was `-k`-filtered. The defect was in the harness, not the fixture. This round's sweep runs the whole file.

### Sweep after the fixes (full file, `PYTHONDONTWRITEBYTECODE=1`)

| mutant | result |
|---|---|
| M0 null control | SURVIVE |
| M1 drop absent-here half | KILLED (2 tests) |
| M2 drop mainline half | KILLED (3 tests) |
| M3 guard never fires | KILLED (3 tests) |
| M4 ignore the override flag | KILLED |
| M5 drop the `args.confirm` half | KILLED — was SURVIVED before this round |
| M6 revert `.strip()` | KILLED — new |
| M7 re-enable argparse abbreviation | KILLED — new |
| M8 return `EXIT_BEHIND` instead of 7 | KILLED — new |

Gates at `82a37ee6`: dev host `GATE: RESULT=PASS exit=0` (pytest + node). Sandbox tier re-running.

ZacxDev:
## Round 2 — delta re-audit of `f9b15a57..82a37ee6`

Two 🟡 and one 🟢. **Both 🟡 were introduced by round 1's own fixes** — the shape the ladder exists to catch. All fixed in `120df520`.

Ledger: **round 2 · payload lines changed this round: 67** (`handoff_doc.py` 46+13, `SKILL.md` 4+4) · scaffolding: 71 test lines. Round 1 was 96 payload. Neither round is near the zero-payload gate, so the ladder is still on the PR rather than on itself.

```audit-claims round=2 audited=82a37ee6..120df520
1. `nothing_to_merge_into()` is now the single predicate for "no usable base doc", called by BOTH `BaseCurrency.replaces_mainline_doc` (the refusal) and `wrong_base_report` (the warning) — claimed to close the drift where the refusal used `.strip()` while the warning used `if not local.lines:`, so a `"\n"` base got the mild size-comparison line instead of "the committed document will be replaced by this delta".
2. `wrong_base_report` takes `base_blank` as a REQUIRED parameter with no default, and both direct callers in the tests were updated — claimed to make a future silent disagreement between the two halves impossible to introduce by omission.
3. `WRONG_BASE_REMEDY`'s discriminator sentence was rewritten — claimed no longer to tell a PROPOSAL-run reader that the absence of `status=stale-base` means they are in the benign case, since the refusal is gated on `--confirm` and a proposal run can never print that line.
4. Added `test_the_WARNING_and_the_REFUSAL_describe_the_SAME_set` (parametrized empty / one-newline) — claimed to pin the RELATIONSHIP between the two halves on the proposal run, not either half alone.
5. Added `test_the_remedy_does_not_tell_a_PROPOSAL_reader_they_are_safe` — claimed to pin that the remedy says the refusal is deferred to `--confirm` rather than absent.
6. `assert "+" in res.stdout` replaced with an assertion on a `+`-prefixed line that is not the `+++` header — claimed to close a vacuous assertion that a mutant stripping every diff body line left green.
7. The loud warning's wording now matches the refusal's ("no usable <path> — missing, empty or whitespace"), and the existing guard pinning the old "there is NO <path>" string was updated — claimed to stop asserting a string that is false of a file that exists and is blank.
8. `SKILL.md`'s bullet now reads "no USABLE doc here (missing, empty or whitespace)" instead of "doc ABSENT here" — claimed to describe the actual trigger after the predicate widened.
9. `SKILL.md` re-fitted to 23,093 B under the 23,100 B budget by eviction — claimed to remove no behavioural instruction.
```

### Sweep after the fixes (full file, `PYTHONDONTWRITEBYTECODE=1`)

Two mutants added this round to pin the seam **from both sides** — a structural guard must fail when either half moves.

| mutant | result |
|---|---|
| M0 null control | SURVIVE — re-verified in isolation, 184 passed |
| M1–M8 (as round 1) | KILLED |
| M9 warning reverts to `not local.lines` | KILLED — new |
| M10 call site passes constant `False` | KILLED — new |

### Gates at `120df520`

Sandbox tier (the one the merge gates on): `TOTAL collected=18572 passed=18570 skipped=2 failed=0`, `RESULT: PASS (exit=0)` — +3 from round 1's 18,569, matching the three tests added. Dev-host tier running.

ZacxDev:
## Round 3 — delta re-audit of `120df520..0c3d30d0`

One 🟡 and two 🟢. The 🟡 was **introduced by round 2's fix** — third consecutive round where the previous round's fix created the next finding. All fixed in `0c3d30d0`.

Ledger: **round 3 · payload lines changed this round: 55** (`handoff_doc.py` 30+19, `SKILL.md` 3+3) · scaffolding: 66 test lines. Payload across rounds: 96 → 67 → 55. No round has been clean and none is near the zero-payload gate, so the ladder is still on the PR.

```audit-claims round=3 audited=120df520..0c3d30d0
1. `wrong_base_report` now receives `currency.replaces_mainline_doc(base_text)` (renamed param `replaces_mainline`) instead of the blank half alone — claimed to close the SECOND arm of the seam, where the mainline had DELETED the doc so nothing is replaced, yet the warning still printed "and <ref> has one … will be replaced by this delta" and the remedy said confirming would refuse (it exits 0 and writes).
2. The three round-2 claims that the seam was closed were rewritten — the "cannot describe different sets again" comment, the "ONE predicate, TWO consumers … required to agree" docstring, and the operator-facing promise in `WRONG_BASE_REMEDY` — claimed no longer to assert a closure that was measurably open.
3. `wrong_base_tells` now calls `nothing_to_merge_into` instead of open-coding `not base_text.strip()` a third time — claimed to remove the third copy of the predicate and make the docstring's consumer count honest (now THREE).
4. Added `test_a_doc_DELETED_on_the_mainline_gets_NEITHER_the_loud_line_nor_a_refusal` — claimed to pin the arm two rounds of fixes missed, and that a legitimate first write there still lands.
5. `test_the_remedy_is_pinned_as_a_WHOLE_normalised_string` replaces the spelled guard — claimed to close the walkability where a reword satisfying "ONLY on `--confirm`" present + one sentence absent could reintroduce the false meaning.
6. `SKILL.md` rule (f)'s FLOOR imperative ("— read the diff anyway") restored after round 2's eviction removed it, leaving an epistemic fact with no action and no other line telling the reader to read the diff on a silent run.
7. `SKILL.md` re-fitted to 23,088 B under the 23,100 B budget by eviction — claimed to remove no behavioural instruction.
```

### Sweep (full file, `PYTHONDONTWRITEBYTECODE=1`, null control re-verified alone)

M11 is the regression this round fixed, and it now dies to the new deleted-arm test.

| mutant | result |
|---|---|
| M0 null control | SURVIVE — 186 passed |
| M1–M10 | KILLED |
| M11 warning uses the BLANK half only | KILLED — new |

### Gates

- **Merged tree** (this branch + `main` @ `3d8caaa1`): `TOTAL collected=18847 passed=18845 skipped=2 failed=0`, `RESULT: PASS (exit=0)`. Merge is clean and none of main's newer commits touch any of the three files this PR changes.
- Branch tiers at `0c3d30d0`: dev host `GATE: RESULT=PASS exit=0` (pytest + node) earlier in the ladder; sandbox tier green on the merged tree above.

### ⚠ Unrelated flaky test seen while gating — NOT from this PR

`scripts/tests/test_git_repo_isolation.py::test_live_cotenants_sees_another_process_in_the_repo` failed on 2 of 3 full sandbox runs and passes on a re-run of the **identical** tree at the **same** load (31.8 vs 32.9). It is a concurrency probe in a subsystem this PR does not touch (3 files changed, none in `testlib/gitenv.py` or that test file), it passes in isolation, and running this PR's 128 `git`-spawning test calls alongside it at `-n 4` was green 3/3. Filed here for visibility only — `scripts/dl-router/tests/test_store.py::test_six_writers_with_a_tiny_busy_timeout_still_land_every_row` showed the same shape once on a `main`-only run.

ZacxDev:
## Round 4 — delta re-audit of `120df520..0c3d30d0`

Two 🟡 and one 🟢. Fixed in `c6f64e4b`.

🔴 **The ladder does NOT end here.** The auditor advised stopping on the grounds that this round found no defect *introduced by round 3's fix*. That is not the stop condition — the rule is a round that returns **no findings**, and this one returned three. #804's rounds 5, 6 and 7 each returned "safe to merge" and each still reported real defects, which is exactly why the rule is keyed to findings rather than verdicts.

I also disagreed with the auditor's scoping of F2 and fixed the behaviour rather than filing it: replacing nothing with something costs nothing, so refusing there is wrong, and deferring it would have shipped a guard that fires in the one shape it must not.

Ledger: **round 4 · payload lines changed this round: 31** (`handoff_doc.py` 22+7, `SKILL.md` 1+1) · scaffolding: 62 test lines. Payload across rounds: 96 → 67 → 55 → 31.

```audit-claims round=4 audited=0c3d30d0..c6f64e4b
1. `replaces_mainline_doc` now also requires `bool(self.mainline.lines)` — claimed to fix the MIRROR bug, where a committed but EMPTY mainline doc (`DocShape(0, 0, …)`, which is not None) made the guard exit 7 NOTHING WRITTEN and print "and <ref> has one (0 section(s) / 0 line(s))", blocking a legitimate first write in a shape where nothing is destroyed.
2. The docstring asserting that `replaces_mainline_doc` is "the only place those two facts are combined" was corrected — claimed no longer to assert a combination the predicate got wrong.
3. `test_a_doc_DELETED_on_the_mainline_gets_NEITHER_the_loud_line_nor_a_refusal` now asserts its own precondition (the stale headline is present) before its two negatives — claimed to close the gap where every assertion was satisfied by "the currency check never ran", measured with a mutant that made the run print nothing at all while all 186 tests stayed green.
4. Added `test_wrong_base_tells_shares_the_predicate_and_stays_SILENT_on_a_blank_base` (parametrized over `""`/`"\n"`/`"   \n\n"`) — claimed to guard round 3's consolidation, which shipped with no mutant covering the line it changed.
5. Added `test_an_EMPTY_mainline_doc_is_NOT_something_to_lose` — claimed to pin claim 1 behaviourally.
6. The unreachable `else ""` shape fallback was replaced with an assert — claimed to stop reading as a handled case that cannot occur now that `replaces_mainline` implies a non-None mainline.
7. The two direct `wrong_base_report` callers in tests now build the sixth argument as `cur.replaces_mainline_doc(BASE_DOC)` — claimed to match the parameter's ACTUAL meaning after round 3 renamed it from `base_blank`.
8. `SKILL.md`'s sentence-terminating period, lost to byte-shaving, restored (23,088 → 23,089 B, still under the 23,100 B budget).
```

### Sweep (full file, `PYTHONDONTWRITEBYTECODE=1`, null control SURVIVED)

M3 and M4 are new and guard exactly the two lines round 4 found bare.

| mutant | result |
|---|---|
| M0 null control | SURVIVE |
| M3 drop the NON-EMPTY mainline half | KILLED — new, guards F2 |
| M4 `wrong_base_tells` reverts to bare falsiness | KILLED — new, guards F3 |
| M1, M2, M5–M11 | KILLED |

### Gates at `c6f64e4b`

- **Required checks both PASS**, and Tekton's numbers are byte-identical to the local sandbox tier: `tekton/devrc-pytests` `collected=18578 passed=18576 skipped=2 failed=0`; `tekton/devrc-nodetests` `tests=1366 pass=1366`.
- ⚠ For the record: at `0c3d30d0` this same check reported ~45 non-passing, and twice earlier today it posted `error — COULD NOT RUN`. The failing run's log was garbage-collected before it could be read, so that divergence is **unexplained**, not diagnosed. The identical counts at `c6f64e4b` are the evidence that the two tiers agree.

ZacxDev:
## Round 5 — delta re-audit of `0c3d30d0..c6f64e4b`

Two 🟡 and one 🟢. Both 🟡 are the part of round 4's defect that its fix did not reach. Fixed in `5b74742e`.

Ledger: **round 5 · payload lines changed this round: 43** (`handoff_doc.py` 38+/5−) · scaffolding: 70 test lines. Payload across rounds: 96 → 67 → 55 → 31 → 43.

```audit-claims round=5 audited=c6f64e4b..5b74742e
1. `base_currency` now measures `mainline_blank = nothing_to_merge_into(shown.out)` at the point it already holds the mainline text, and `replaces_mainline_doc` reads that flag instead of `bool(self.mainline.lines)` — claimed to put BOTH sides of the equivalence class through one predicate, fixing the whitespace-only mainline case (`"\n"`, `"   \n\n"`) that round 4's `bool(lines)` still refused.
2. `mainline_blank` is `None` when not measured and the predicate requires `is False` — claimed to keep an unanswered question from becoming a refusal.
3. Added `test_a_WHITESPACE_mainline_doc_is_ALSO_not_something_to_lose` (parametrized `"\n"` / `"   \n\n"`) — claimed to pin the half round 4 missed.
4. Added `test_a_HEADINGLESS_prose_mainline_doc_IS_something_to_lose` — claimed to pin the QUANTITY in the fail-OPEN direction, where swapping `lines` for `sections` previously SURVIVED all 190 tests and would have let `--confirm` replace a real prose document carrying no `##` heading.
5. `wrong_base_report`'s `replaces_mainline` parameter now documents that it must come from `currency.replaces_mainline_doc(...)` — claimed to state the precondition its plain `bool` signature cannot.
6. `nothing_to_merge_into`'s docstring corrected from THREE consumers / two combined facts to FOUR consumers / THREE facts, `base_currency` being the fourth.
```

### Sweep — 15 mutants, full file, `PYTHONDONTWRITEBYTECODE=1`

| mutant | result |
|---|---|
| M0 null control | SURVIVE |
| **M2 drop mainline-exists** | **SURVIVE — EQUIVALENT, expected** |
| M5 mainline measured by SECTIONS (fail-open) | KILLED — pins the quantity |
| M1, M3, M4, M6–M14 | KILLED |

🔴 **M2 is an equivalent mutant and is recorded as an expected SURVIVE — not counted as coverage, and not "fixed" by deleting the line.** `mainline_blank` and `mainline` are assigned in the same branch, so `mainline_blank is False` already implies `mainline is not None`. The clause stays because it is the type-narrowing contract `wrong_base_report` relies on when it dereferences `mainline.sections`. The sweep now asserts its own anchor still matches the source, so the mutant cannot quietly stop applying and read as a pass.

### Why this round mattered

Round 4's fix was correct about `""` and wrong about the other two members of the equivalence class the PR itself defines — and the guard's chosen quantity was free to regress to a **fail-open** one under a fully green suite. Picking a third `DocShape` field would have been a fourth guess; running the mainline text through the same predicate as the local text is the fix that removes the class.

ZacxDev:
## Round 6 — delta re-audit of `c6f64e4b..5b74742e`

**The payload came back clean** — *"I could not break the payload"*. One 🟡 and one 🟢, both in my own tests. Fixed in `ac5575a5`.

🔴 **Attribution gate: round 6's fix changed ZERO payload lines** (`test_handoff_doc.py` +18/−1 only). Round 5 changed 43, so this is the first zero-payload round, not the second. Per the ladder's own rule, **two consecutive** zero-payload rounds mean it has left the PR and is auditing its own scaffolding — so if round 7's fixes are also scaffolding-only, I stop there and file the remainder rather than running round 8.

Ledger: **round 6 · payload lines changed this round: 0** · scaffolding: 19.  Payload across rounds: 96 → 67 → 55 → 31 → 43 → **0**.

```audit-claims round=6 audited=5b74742e..ac5575a5
1. `test_a_WHITESPACE_mainline_doc_is_ALSO_not_something_to_lose` now asserts the mainline SHAPE line on a proposal run before its negatives — claimed to close the gap where every assertion was satisfied by "the mainline copy was never read at all", measured with a mutant that made `base_currency` skip the mainline read and SURVIVED this test run alone while the sibling headingless test died to it.
2. The stale HEADER was deliberately NOT used as that assertion — it prints whenever `doc_behind` is non-zero, so only the shape line proves the copy was actually read and measured.
3. `_mainline_doc` no longer passes `--allow-empty` to `git commit` — claimed to restore git's own "nothing to commit" refusal as the fixture's guard against a body identical to what the mainline already holds.
```

### Re-measured after the fix

The same mutant (`base_currency` never reads the mainline) now gives **2 FAILED scoped to that test**, where before the fix it gave 2 passed. Full file: **193 passed**.

### What round 6 re-derived rather than accepted

Both are claims I made in earlier commit messages, and it checked them independently instead of trusting them:

- **"M2 is an EQUIVALENT mutant"** — correct; it re-ran the mutant rather than reading the label. It additionally found **M4** (`not self.mainline_blank` in place of `is False`) equivalent on its own, with the **double** mutant KILLED, so the pair is jointly pinned.
- **`git show` under `text=True` round-trips whitespace exactly** (only CRLF→LF), verified against real committed blobs. That is the assumption round 5's structural fix rests on, so it mattered that someone other than its author checked it.

### Gates at `ac5575a5`

Sandbox tier (the merge-gating one): `collected=18581 passed=18579 skipped=2 failed=0`, `RESULT: PASS (exit=0)`. Gated **before** pushing this time.

ℹ Not a code finding, carried from round 6's report: both Tekton checks had been **pending since 06:47Z with `mergeStateStatus: BLOCKED`** — the documented devrc mode where a run hits `timeouts.tasks`, posts nothing, and only a fresh push clears it. This push should re-trigger them.

PR #1049, opened 2026-08-29
feat(cairn): entries can name the tasks they answer — a lossless multi-system `tasks:` ref

ZacxDev:
Closes **cg#428** criteria 1, 2, 3, 5, 6, 7, 8, 9. Criterion 4 (URL resolution) is
deliberately deferred — see *Why this is only half the card* below.

## The gap

Measured across all 120 front-matter blocks in the live store: `service` 120, `aliases`
119, `sensitivity` 98, `scope` 97, `created_by` 97, `namespace` 32, `repo` 23 — and **zero**
carrying a task, PR or session ref. Nothing joined a subsystem entry to the work that
produced it.

An entry now carries `tasks: [<system>:<id>, ...]` (or `task:` for one). The split is on the
**first** colon and the id half is preserved **byte-identically**, so
`github:innovation-upstream/devrc#428` survives whole. The `#` is the point, not an edge
case: any encoding that cannot carry one cannot carry a GitHub reference at all.

## Why this is only half the card

The handoff directed reading devrc#1011 first, on the grounds that cg#428's criteria 1/3/4
are "the same predicate" and should be imported rather than rewritten. Reading it showed
that is **half right**, and the wrong half matters:

| criterion | #1011 |
|---|---|
| **4** — resolve a ref to a URL | yes, exactly this (`CLICKUP_TASK_URL`, `_github_url`, `clawgate_url`) |
| **1** github half — `owner/repo#N` | mostly; its `_REPO` forbids dots by design — correct for detecting refs in free text, wrong as a storage validator |
| **3** — parser must NOT enumerate systems | **the opposite of #1011's design.** `mention_scan` enumerates exactly three platforms and refuses everything else, deliberately ("a noisy scanner is worse than none") |
| 1,2,5,6,7,8 — front-matter parse/round-trip/validate | not addressed; #1011 never touches front matter |

So the card is two layers and only one is #1011's. **Layer A** (this PR) is the
system-agnostic storage schema. **Layer B** (criterion 4) *imports* #1011's URL builders
rather than growing a second copy — `claude/RULES.md` "one rule, one place". #1011 is open
and its pytests leg is red on `test_subsystem_store_api.py`, a file it does not touch and
which a live `claim-work` entry calls a ~60% repo-wide flake; coupling this PR to it would
inherit that for no benefit.

## The card's block-form premise was wrong, in the dangerous direction

cg#428 records that a block-style `tasks:` "parses clean and is silently ignored". Measured
on the pre-change reader, it does not — it **corrupts the mapping**:

```
{'service': 'thing', 'tasks': '',
 '- clickup': '868abc123',
 '- github': 'innovation-upstream/devrc#428'}
```

The key goes silently empty and **every** item is promoted to a phantom front-matter key by
its own internal colon (an item *without* one is dropped instead — both halves are data
loss). That is why criterion 6 is settled as **support** rather than reject: rejecting
leaves the two-line hazard in the parser for every future key.

The widening is safe **by measurement, not assumption** — across all 120 live front-matter
blocks the count of lines beginning `- ` is **zero**, so no existing entry changes meaning.
A bare `key:` with no items under it still reads as the empty string and not `[]`
(recognition is by lookahead), because `sensitivity:` is read by callers that call string
methods on it and handing them a list would raise `AttributeError` from inside a reader.

## Design notes

- **Validation lives in `SubsystemEntry.from_mapping` and nowhere else**, so
  `subsystem_touch --validate` and the loader cannot drift: the validator answers "would the
  loader accept this file?" by building exactly what the loader builds.
- **Read surface:** the index row gets a *conditional* `N task(s)` **count** — a ref is up to
  36 characters and the row's contract is one line — while the refs themselves print in the
  entry body, which `--ref` already shows.
- **`lossy_tag_for()` has no inverse by construction.** The test proves the map is not
  injective (`github:zacxdev/homelab-infra#429` and `github:zacxdev-homelab/infra#429` both
  flatten to `github:zacxdev-homelab-infra-429`) — stronger than asserting no inverse is
  exported, because a reviewer can delete an assertion but not a collision.

## Verification

- **Both gate tiers green**, run on this branch @ `c7049b11` with a clean tree:
  - dev-host `nix develop … scripts/gate.sh` -> `GATE: RESULT=PASS exit=0` (pytest PASS; node 1366/1366, floor 1317)
  - sandbox `nix build .#checks.x86_64-linux.{pytests,nodetests}` -> exit 0, read **without** a
    pipe, with a negative control (`.doesnotexist` -> exit 1) confirming the shape can go red
- **39 new tests.** 1,110 green across `test_subsystem_task_refs` + `test_subsystem_recall` +
  `test_subsystem_store_api`.
- **Watched to fail.** The two genuine regression tests go red on `origin/main`'s reader for
  their own reason, with the corruption above reproduced verbatim. The module docstring
  labels the other 37 as new-feature tests — on pre-change code they can only `ImportError`
  at collection, which fails everything for one reason and so discriminates between none of
  them. Two more are invariant guards, green before and after, labelled as such.
- **Real-store output is byte-identical** pre vs post change (45,429 bytes), with a positive
  control confirming `cmp` goes red on a 1-byte edit.

## Not verified / not done

- **Criterion 4 is not implemented** — Layer B, gated on #1011 merging.
- **Not gated on the merged tree.** `strict` is false on `main`, so both green checks are
  claims about this branch, not about the tree the merge creates.
- **No backfill** of the 120 existing entries (an explicit card non-goal), and no
  `session:`/`pr:` key — cg#365 measured that the commit trailer's token does not join to the
  transcript uuid, so adding one now would bake in the wrong id.
- **`/handoff` is not yet taught to emit `tasks:`.** The card's Assumptions put that wiring in
  scope; it is not in this branch.

https://claude.ai/code/session_01C9Xx77rCSsqqcRxCJwFtzz


ZacxDev:
## Audit round 1 — findings and corrections to my own PR description

An adversarial audit was run against `c7049b11`. It re-derived every self-reported number in the PR body and commit message; all reproduced except two claims of mine that were **scoped too widely**. Correcting them here rather than silently editing the body, since anyone who has already read it read the wrong version.

### 🔴 Corrections to the PR description

1. **"Real-store output is byte-identical" is true of the TEXT surfaces only.** `--json` is *not* — `entries[].tasks` is emitted unconditionally, so the payload grows (+19 B on the live store at the time of audit, +57 B on a `--limit 3` sample). No JSON schema consumer exists today, so the impact is nil, but the claim should have said "text surfaces". The `listing_line` code comment scoped it correctly; the PR body did not.
2. **"45,429 bytes" is not reproducible** — the live store moves, so the number is stale rather than wrong. Byte-identity itself re-verified and still holds. ⚠ My first re-verification attempt compared two **argparse usage errors** as "identical", because zsh does not word-split an unquoted `$mode`; the real comparison now carries a content control asserting each side is a genuine index read.

### 🔴 The blocking finding — fixed in `d9b9900d`

The block-form fix closed the corruption **for clean input only**. `_block_items_from` tested membership with `startswith("- ")`, so a bare `-` or `- ` with only trailing whitespace terminated the scan and every ref below was promoted to a phantom key again — byte for byte the corruption this PR exists to fix.

**And it was strictly worse than the bug it replaced.** `tasks` then read as `''` (falsy), so `from_mapping` treated the key as absent: the entry **loaded clean with `tasks=[]`**, `--validate` reported OK, the index row showed no badge. Pre-change the mapping was at least *visibly* polluted. An empty item in the **middle** was the likelier shape and silently **truncated** the list rather than zeroing it.

Reproduced, then fixed: membership is the leading `-` alone, via `_is_block_item` shared by the scanner and the swallower — two spellings of that predicate is what produced the bug.

### Also fixed

- `format_task_refs` could **write a line the loader rejects** (a comma in an ident splits on re-read → entry MALFORMED). Commas and control characters now rejected at parse time; bare strings re-parsed rather than trusted.
- `lossy_tag_for` **did not satisfy the grammar its own docstring stated** — `github:###` produced the tag `github:` (empty slug, colliding with every other such ref) and a deep path produced a 90-rune tag against a 64-rune limit. Both now raise. Its test sampled four short constants that could not violate either bound.
- The **lookahead was unbounded**, so a bare `sensitivity:` could bind a distant `- ` list and become a list. A blank line now ends the block.
- **Three surviving mutants killed** (two `break`→`continue`, one dropped type guard), each by a fixture that can actually observe it. Re-swept with both controls under `PYTHONDONTWRITEBYTECODE=1`.
- 🔴 **A real team-workspace ClickUp id (``868…` (redacted)`) was about to ship in this PUBLIC repo**, including inside a user-facing error string. Replaced with synthetic `868abc123`; zero occurrences remain. Zero on `main`, so nothing leaked previously.
- `tasks:` added to the authoritative schema reference `index-store.md`, hashed-region pin updated.

### ⚠ One mutant survives, documented rather than patched

Removing either explicit blank-line `break` is an **equivalent mutant** — a blank line is already not an item, so `_is_block_item` breaks the scan a line later by the general rule. The checks stay as defence-in-depth, but the docstrings now name the mechanism that *actually* holds, rather than the one I had claimed.

### Still not done

- **Criterion 4** (URL resolution) — Layer B, gated on #1011.
- **Not gated on the merged tree** — `strict` is false, and this branch is behind `main`.
- The audit could not check the **laptop's store**, which is unreplicated; every census number is the workbench's.

ZacxDev:
Claims block for the round-2 delta audit. Legend: `<from>` = the tip round 1's audit READ; `<to>` = the head round 1's FIXES produced.

```audit-claims round=1 audited=c7049b11..d9b9900d
1. An EMPTY block item (bare `-`, or `- ` with only trailing whitespace) no longer terminates the block scan, so it no longer resurrects the phantom-key corruption; membership is now the leading `-` alone via `_is_block_item`, shared by the scanner and the swallower.
2. An empty item in the MIDDLE of a list no longer silently truncates the ref list at the loader; the entry no longer loads clean reporting fewer tasks than the file carries.
3. The swallow now runs unconditionally once any member follows the key, including when every item was empty, so no member is left behind to be re-read as a front-matter key.
4. `parse_task_ref` rejects a comma, so `format_task_refs` can no longer write a line the inline-list reader splits into a MALFORMED entry.
5. `parse_task_ref` rejects control characters (NUL, ESC, DEL).
6. `format_task_refs` re-parses a bare string rather than trusting it, so the writer cannot emit a ref the reader would reject.
7. `lossy_tag_for` raises rather than returning a tag that violates the grammar its docstring states: an ident normalizing to the empty string, and a tag over 64 runes.
8. The block-list lookahead is bounded by a blank line, so a bare `key:` cannot bind a distant `- ` list and change type from string to list.
9. Three mutants that survived round 1 are now killed: swallow-loop `break`->`continue`, `_block_items_from` `break`->`continue`, and removal of the `_AbcSequence` guard on `tasks:`.
10. The real ClickUp id ``868…` (redacted)` is removed from the repo entirely and replaced with the synthetic `868abc123`.
11. `tasks:` is documented in the authoritative schema reference `claude/skills/analyze-service/reference/index-store.md`, and its hashed-region pin in `test_subsystem_resolver.py` is updated to match.
12. The docstrings on both blank-line `break`s now state that removing them is an EQUIVALENT mutant and that `_is_block_item` is the mechanism that actually enforces the bound — the checks are retained as defence-in-depth only.
13. The PR description's "byte-identical" claim is corrected to cover TEXT surfaces only; `--json` gains `entries[].tasks` and is not byte-identical.
```

ZacxDev:
Claims block for the round-3 delta audit.

```audit-claims round=2 audited=d9b9900d..65130423
1. The blank-line bound introduced in round 1 is REVERTED: a blank line no longer terminates the block scan, so a list separated from its key by a blank line, or containing one, parses as YAML parses it.
2. A `-`-led line is never promoted to a front-matter key, enforced by a skip in the OUTER loop rather than by any property of where a block scan starts or stops.
3. An ORPHAN list item (one with no owning key, which is not valid YAML) is skipped, not turned into a key, and its id does not leak in as any key's value.
4. `parse_front_matter` agrees with `yaml.safe_load` on 8 front-matter shapes, with exactly one deliberate divergence: a bare `key:` is the empty string here where YAML gives None.
5. `format_task_refs` re-parses EVERY input including a `TaskRef`, so a hand-built `TaskRef` carrying a comma or newline can no longer be rendered into a line the reader rejects.
6. `parse_task_ref` rejects C1 control characters (0x80-0x9F) in addition to C0 and DEL.
7. `TAG_MAX_RUNES` is pinned by its own assertion and the boundary fixtures are LITERAL lengths, so mutating the constant no longer moves the fixtures with it.
8. The orphan-list, C1 and tag-boundary mutants, which survived the first re-sweep of round 2, are now killed.
9. `test_the_swallow_STOPS_at_the_first_non_item_line` no longer claims to kill a mutant it does not kill; the killer is named as `test_TWO_block_lists_in_one_file_do_not_merge`.
10. The "EVERY ONE OF THESE resurrected the corruption" comment is corrected to three of five, the other two being invariant guards.
11. Removing the blank-skip in `_block_ends_at` is documented as an EQUIVALENT mutant that the outer-loop skip masks, with both guards retained deliberately rather than one deleted as unreachable.
12. The real ClickUp id is removed from the PR body and both PR comments (it remains in commit c7049b11 on the branch; only a squash merge keeps it out of main).
```

⚠ Note for the auditor: claim 4's oracle (PyYAML) is a **test-only** `importorskip` dependency and is deliberately not imported by the module under test. Confirm it is actually present in whatever environment you run — a silently skipped oracle is indistinguishable from a passing one.

PR #1057, opened 2026-08-29
chore(rescue): commit three untracked files no open PR covered — a merge already landed under this checkout while they sat there

ZacxDev:
Pure preservation. None of these were written in this session; all three sat untracked in `~/workspace/devrc` — which is also a **deploy target** — and `gh pr list` showed no open PR carrying any of them.

Per `RULES.md` → Git Workflow, docs/notes in a working tree are unsaved work: one routine `checkout`, `stash` or deploy by another session deletes them with no report.

## The risk was live, not theoretical
`nix/system/apply-dnsmasq-docker-io-pin.sh` was untracked in this same checkout at session start and is **tracked now** — a merge landed underneath while these three sat beside it. The same event with a checkout instead of a merge would have taken them silently.

## What's here
| file | note |
|---|---|
| `claudedocs/handoff-mention-detection.md` | mention-detection effort (clawgate/GitHub/ClickUp refs, click-to-open) |
| `claudedocs/proposal-mention-detection.md` | its design proposal |
| `scripts/cleanup-disk.sh` | ad-hoc disk reclamation (go/bazel/pnpm/NuGet caches, journal vacuum) |

Open PR **#1011** carries the mention-detection *code* but neither doc.

## Committed as found, deliberately
The point is preservation — silently "improving" another session's file while claiming to rescue it would make this diff a poor record of what actually existed. Two notes for whoever owns `cleanup-disk.sh`, neither blocking:
- it is the **first tracked `scripts/*.sh` to call `sudo`** (repo convention stages sudo work under `nix/system/apply-*.sh`);
- its `rm -rf` lines are unguarded.

Checked for public-repo hazards before committing: no IPs, credentials, or media paths in any of the three. The hardcoded `/home/zach` paths and `dp-prod` reference are already established repo vocabulary (107 tracked files contain the former; the latter is in `CLAUDE.md`).

https://claude.ai/code/session_01DJzmPYW8nDjauyCHRtS8Ai

### DIFF
--- a/scripts/cleanup-disk.sh
+++ b/scripts/cleanup-disk.sh
@@ -1,0 +1,129 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# 🔴 DRY-RUN BY DEFAULT — destructive steps need an explicit `--apply`.
+#
+# Round 2 of the audit on devrc#1057 found that removing the sudo call (see the
+# journal block below) closed ONE instance of the laundering hazard and left two
+# WORSE-rated ones. MEASURED through oc_permissions.py over opencode.jsonc:
+#
+#   rm -rf /home/zach/.local/share/NuGet                     -> deny  [*rm -rf /*]
+#   rm -rf /home/zach/.local/share/dp-prod-ssr-regression-*  -> deny  [*rm -rf /*]
+#   bash <anywhere>/cleanup-disk.sh                          -> allow [*]
+#
+# `deny` outranks the `ask` that was removed, so a tracked, allow-rated script
+# was laundering two deny-rated recursive deletes past a ledger that matches on
+# the command STRING and cannot see inside an invoked script file. That gating
+# was deliberately hardened in #744 and verified live; defeating it by accident
+# is exactly what this file's own comment says not to do.
+#
+# ⚠ Rewriting `rm -rf` as `find -delete` would dodge the rule WITHOUT closing
+# the hazard — MEASURED: `find / -delete` resolves `allow`, so the ledger would
+# never see it. That is gaming the ledger, not respecting it.
+#
+# 🔴 WHAT THE GATE BELOW DOES AND DOES NOT DO — read this before trusting it.
+# It NARROWS the hazard; it does NOT close it. An earlier version of this
+# comment claimed the gate meant "the allow-rating grants nothing". That was
+# FALSE, and audit round 3 measured it:
+#
+#   bash <anywhere>/cleanup-disk.sh            -> allow [*]
+#   bash <anywhere>/cleanup-disk.sh --apply    -> allow [*]   <-- equally allowed
+#
+# `--apply` is rated exactly as permissively as the bare form. The ledger is
+# blind to both, so an agent that decides to pass `--apply` still executes two
+# `deny`-rated recursive deletes unattended. What the gate actually buys is
+# narrower and still worth having: an ACCIDENTAL or incurious invocation is
+# inert, and deleting anything now takes a deliberate, auditable token rather
+# than happening as a side effect of "free up some disk space".
+#
+# The residual hole is real and is NOT closed here: closing it would mean the
+# script printing the `rm -rf` commands for the caller to run — so the caller
+# hits the `deny` — which turns this into a printer and defeats its purpose.
+# That trade is the operator's to make, not this file's.
+#
+# Same shape as the repo's other operator-acts (`sync-skill-tiers.py`,
+# `sync-claude-permissions.py`): dry-run unless `--apply`.
+APPLY=0
+for _a in "$@"; do
+  [ "$_a" = "--apply" ] && APPLY=1
+done
+
+if [ "$APPLY" -ne 1 ]; then
+  echo "=== Disk cleanup — DRY RUN (nothing will be deleted) ==="
+  echo "Would clean:"
+  echo "  - Go build cache            (go clean -cache)"
+  echo "  - Bazel cache              /home/zach/.cache/bazel/*"
+  echo "  - pnpm store               (pnpm store prune)"
+  echo "  - NuGet cache              /home/zach/.local/share/NuGet"
+  echo "  - dp-prod test artifacts   /home/zach/.local/share/dp-prod-ssr-regression-*"
+  echo
+  # 🔴 The escalation hint is printed ONLY to an interactive terminal.
+  # Audit round 3: "the dry-run text IS the escalation instruction — the gate
+  # prints its own bypass to whoever is reading, and an agent is a reader."
+  # A headless agent capturing stdout is not handed the next step. This is a
+  # speed bump, not a boundary — the flag is still discoverable by reading the
+  # script — but it stops the inert path from advertising the live one.
+  if [ -t 1 ]; then
+    echo "Re-run with --apply to actually delete."
+  else
+    echo "Nothing was deleted. Re-run interactively to see how to proceed."
+  fi
+  echo "Not included (needs root, run it yourself):"
+  echo "  sudo journalctl --vacuum-size=500M"
+  exit 0
+fi
+
+echo "=== Disk cleanup ==="
+echo "Before: $(df -h / | tail -1 | awk '{print $4}') free"
+
+# Go build cache (no sudo needed)
+echo "Cleaning Go build cache..."
+go clean -cache 2>/dev/null && echo "  ✓ go-build cache" || echo "  ✗ go-build failed"
+
+# Bazel cache (owned by user but some files are immutable)
+echo "Cleaning Bazel cache..."
+find /home/zach/.cache/bazel -mindepth 1 -delete 2>/dev/null || true
+echo "  ✓ bazel cache"
+
+# pnpm store prune
+echo "Pruning pnpm store..."
+pnpm store prune 2>/dev/null && echo "  ✓ pnpm store" || echo "  ✗ pnpm failed"
+
+# NuGet cache
+echo "Cleaning NuGet cache..."
+rm -rf /home/zach/.local/share/NuGet && echo "  ✓ NuGet" || echo "  ✗ NuGet failed"
+
+# Test artifacts
+echo "Cleaning test artifacts..."
+rm -rf /home/zach/.local/share/dp-prod-ssr-regression-* && echo "  ✓ dp-prod artifacts" || echo "  ✗ dp-prod failed"
+
+# System journal — DELIBERATELY NOT DONE HERE.
+#
+# 🔴 The original untracked copy of this script ran:
+#
+#     sudo journalctl --vacuum-size=500M
+#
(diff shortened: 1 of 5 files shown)


ZacxDev:
## Correction to this PR body — two claims were wrong

Surfaced by `/audit-pr` round 1. Posting rather than silently editing the body, since a reviewer may already have read the wrong version.

**❌ Claim (c) — FALSE as written.** The body says *"the latter [`dp-prod`] is in `CLAUDE.md`"*. It is not: `git grep -c dp-prod -- CLAUDE.md` → **0**.

The *conclusion* survives — `dp-prod` is established vocabulary in **13 tracked files** (`.zshrc`, `scripts/obs-read`, `scripts/repo-cos/routing.py`, `scripts/session-analysis/initiative-scan.py`, and several tests) — but the citation a reviewer would follow to clear a client identifier in a **public** repo does not exist. Anyone checking it would have found nothing and had to redo the work.

**⚠ Claim (b) — understated.** The body says 107 tracked files contain `/home/zach`. The real number is **142**; 107 is the count *excluding* `claudedocs/`. Direction is safe — it makes the "established vocabulary" argument stronger, not weaker.

Verified independently, not taken from the audit.

## Also: two real findings fixed in `f78e3786`

**F2 — privilege-gate bypass (the one that mattered).** `cleanup-disk.sh` called `sudo journalctl --vacuum-size=500M`. Measured through the repo's own resolver:

| command | rating |
|---|---|
| `sudo journalctl --vacuum-size=500M` | **ask** |
| `bash <anywhere>/cleanup-disk.sh` | **allow** |

The sudo rules match on the command *string* and `guard_core.py` cannot read a script *file*, so committing this created an `allow`-rated route to an `ask`-rated action for the headless opencode agent — and `ship.sh` would put it on both hosts.

🔴 **The audit's proposed fix (move it to `nix/system/apply-*.sh`) does not work** — I measured the moved path before applying it: `bash …/nix/system/apply-disk-cleanup.sh` → **allow**. Any script path is allow-rated. The only effective fix is for the script to contain no sudo call, so the step is removed and the command printed for the operator instead. The removed line is preserved verbatim in the file.

This knowingly breaks the "committed as found" principle — a measured privilege hole outranks fidelity to a throwaway script.

**F3 — `handoff-mention-detection.md` would have misdirected the next session.** It says "NOT DONE: no code written", but #1011 already carries `mention_scan.py` and its tests. The doc is not inert: `resume-state.sh` resolves `/resume mention-detection` to the newest matching handoff, and this would be the only one. A NOT-CURRENT banner now points at #1011; the body is byte-identical.

**Left as found deliberately:** F1 (bazel step prints ✓ unconditionally) and F5 (unmatched glob prints ✓) are the script author's defects, not the rescuer's.

## 🔴 This cannot merge yet, for a reason unrelated to this PR
`main` is **red in the merge-gating sandbox tier** — `test_espanso_detect.py::test_live_existing_resolutions_not_made_ambiguous` fails on a clean checkout at `d202ef59` with none of this PR's files present. Confirmed independently. **#1058** and **#1060** are competing fixes; one must land first.

ZacxDev:
## Round 1 claims ledger

Posting the block round 1 should have emitted. Claims only — what was asserted fixed, not why it is correct.

```audit-claims round=1 audited=47d82568..7ce5bfaf
1. F2 privilege-gate bypass: the sudo journalctl call was REMOVED from scripts/cleanup-disk.sh, so the script contains no sudo invocation; the auditor's proposed remedy (relocate to nix/system/apply-*.sh) was rejected as ineffective after measuring that path as `allow` too.
2. F3 stale handoff: a NOT-CURRENT banner was prepended to claudedocs/handoff-mention-detection.md pointing at PR #1011, with the original body left byte-identical.
3. F4 PR-body claim (c) "dp-prod is in CLAUDE.md" was corrected publicly as FALSE in a PR comment; the 13-tracked-file evidence was substituted.
4. F6 PR-body claim (b) was corrected publicly from 107 to 142 files containing /home/zach.
5. F1 (bazel step prints unconditional checkmark) and F5 (unmatched glob prints checkmark) were deliberately NOT fixed — left as found, as the script author's defects.
```

ZacxDev:
## Round 2 claims ledger + corrections to my earlier comment

```audit-claims round=2 audited=7ce5bfaf..ad6fcf1b
1. Finding 1 (laundering incomplete): the two deny-rated `rm -rf` calls are no longer reachable by a bare invocation — the script is now DRY-RUN BY DEFAULT and requires an explicit `--apply` for any destructive step, so an allow-rated bare `bash cleanup-disk.sh` deletes nothing.
2. Finding 1 remedy choice: rewriting `rm -rf` as `find -delete` was REJECTED as dodging the rule without closing the hazard; the gate was chosen instead.
3. Finding 2 (stale status): the banner in claudedocs/handoff-mention-detection.md now says PR #1011 is MERGED (0493e612, 2026-08-30) and its code is on main, replacing "(open)".
4. F1 and F5 from round 1 remain deliberately unfixed, unchanged from the round-1 disposition.
```

### Two corrections to my earlier comment on this PR
Both were caught by round 2, and both were wrong at the time I posted them:

- **The fix sha I cited (`f78e3786`) was orphaned by a rebase** and is no longer an ancestor of the head. The round-1 fix is now `60ec38fb`; round 2 is `ad6fcf1b`.
- **The block saying "🔴 This cannot merge yet — `main` is red" is stale.** #1060 landed as `31cd214d` and `main` is green — I verified `collected=18703 passed=18701 failed=0` in the sandbox tier.

### Why Finding 1 mattered
My round-1 fix removed a sudo call rated **ask** and left two calls rated **deny** — a strictly worse grade of the identical hazard, in a file whose own comment documents that hazard. Measured through the repo's resolver:

| command | rating |
|---|---|
| `rm -rf /home/zach/.local/share/NuGet` | **deny** `[*rm -rf /*]` |
| `rm -rf …/dp-prod-ssr-regression-*` | **deny** `[*rm -rf /*]` |
| `bash <anywhere>/cleanup-disk.sh` | **allow** `[*]` |

Verified behaviourally with both controls, since round 2's own auditor first got a false negative from a harness whose logging was swallowed by the script's `2>/dev/null`:

```
bare run  -> rc 0, 0 destructive calls, 0 sudo
--apply   -> rc 0, 5 destructive calls, 0 sudo
```

The positive control is the point — without it, "0 destructive calls" is indistinguishable from a harness wired to nothing.

### ⚠ Gate status: this still cannot merge, for a NEW reason
`main` is green and this PR's head tree passes **both** tiers locally (`pytests failed=0`, `nodetests fail=0`). But the CI gate is failing every PR regardless of content. Root cause found:

```
CI pod:    sandbox = false
workbench: sandbox = true
```

Derivation `ydzfas1zzmm446y07ivymknxmirdzavi` produced `failed=1` locally and `failed=43` in CI — **identical inputs, different outputs**, which is impurity by definition. The "hermetic sandbox tier" is only hermetic where nix's sandbox is on. That is a cluster/image fix, not a repo one.

ZacxDev:
## Round 3 claims ledger

```audit-claims round=3 audited=ad6fcf1b..ac6a8ed3
1. Finding 1 (comment overstates the fix): the comment no longer claims the gate means "the allow-rating grants nothing" — it now states, with the measurement, that `--apply` is allow-rated exactly like the bare form, that the gate makes an ACCIDENTAL invocation inert rather than closing the hole, and names the residual hole plus why the closing fix (printing the rm -rf commands so the CALLER hits the deny) was rejected.
2. Finding 1 second half (the dry-run text printed its own bypass): the `--apply` hint is now emitted only when stdout is a TTY, so a headless agent capturing stdout is not handed the escalation.
3. Finding 2 (no regression guard): scripts/tests/test_cleanup_disk_gate.sh added AND registered in run-tests.sh SHELL_TESTS in the same commit; it is behavioural, asserts a positive control before any zero, logs to a file rather than stderr, and was watched red under an APPLY=0 -> APPLY=1 mutation.
4. The three untracked originals were removed from the workbench checkout after being verified byte-identical to 60ec38fb, so ship.sh will not hit "untracked working tree file would be overwritten" and skip the host.
```

### The finding was in my comment, not the code
Round 3 measured that my round-2 claim was **false**:

| command | rating |
|---|---|
| `bash <anywhere>/cleanup-disk.sh` | allow `[*]` |
| `bash <anywhere>/cleanup-disk.sh --apply` | allow `[*]` |

`--apply` is rated exactly as permissively as the bare form, so the gate **narrows** the hazard rather than closing it — and the dry-run text was printing the escalation instruction, which an agent reads as readily as a human. Per `RULES.md`, a comment is a claim too, and an overstated one is the shape that stops the next person looking.

### The new guard was watched red
```
green at HEAD                    -> RESULT: PASS (exit=0)
mutate APPLY=0 -> APPLY=1        -> FAIL: bare run made 5 destructive call(s) — the gate is bypassed
```
Its own error, not a neighbour's. It also runs a positive control *first*, so a zero from the bare run can never be a harness wired to nothing — the exact trap that produced a false negative in round 2.

⚠ It failed on its **own** bug first: `grep -c` prints `0` *and* exits 1, so `grep -c … || echo 0` emitted `0\n0` and every `-eq` blew up with "integer expected". Fixed and noted inline.

### Merge-blocker hazard closed
The three rescued files were still sitting **untracked** in the workbench checkout. Left there, `ship.sh`'s `git merge --ff-only` would refuse with *"untracked working tree file would be overwritten by merge"* and **skip the host** — the silent-skip mode `CLAUDE.md` documents, where a host stops receiving every future change while still looking healthy. Verified byte-identical to `60ec38fb`, backed up, removed.

ZacxDev:
Trimmed the two mention drafts from this branch (`9a09ad58`, plain fast-forward — nothing rewritten, nothing force-pushed).

**Why:** `claudedocs/mention-detection-as-built.md` landed on `main` in `e0e29e7b` (#1067) and supersedes both. They aren't merely stale — three of their load-bearing premises were **measured false** while implementing #1011:

- *"clawgate has no browser UI, only an API"* — `server.go:384` serves the full web UI at `GET /tasks`; the proposed `clawgatectl`-popup workaround solved a non-problem
- *detection needs a `PostToolUse`+`Stop` hook* — `session-tailer.py` already parses assistant blocks and already emits, off the hot path
- *clicking needs tmux `set -g mouse on`* — link clicking here is alacritty's built-in hint; **zero** tmux changes shipped

Landing them would put those three claims on `main` beside a doc that refutes each one, reading as current design rather than superseded.

🔴 **The rest of this PR is untouched and still worth landing** — `scripts/cleanup-disk.sh`, its gate test and the three registration updates. Closing this PR to be rid of the drafts would have discarded that, which is why this is a trim, not a close.

Nothing is lost: both files remain readable at `19ef91ab`, and their durable content is on `main` in the as-built doc.

Also worth crediting — this branch's own commit message is what caught them: *"docs/notes in a working tree are unsaved work: one routine checkout or deploy by another session deletes them with no report."* That rescue is exactly why the corrections had something to correct.

PR #1058, opened 2026-08-29
fix(espanso-detect): ambiguity is DESIGNED behaviour — only MISATTRIBUTION is a regression

ZacxDev:
## Why `main` is red

`fc024d59` ("espanso") re-added `"ask"` and `"clarifying"` to `:dacq`'s `search_terms` — the terms #1023 removed hours earlier. Two pinned terms now resolve ambiguously, `test_live_existing_resolutions_not_made_ambiguous` fails, and since both Tekton checks are required with `enforce_admins: true`, **every open PR in the repo is blocked**.

## Why the config is not the thing to fix

The overlap is deliberate — the operator wants `:dacq` findable by "ask". That is a legitimate authoring choice: espanso's own Ctrl+Space search lists both snippets and the operator picks. Nothing about espanso's usability is harmed.

What the guard actually protects is **telemetry attribution**, and this module's docstring already promises the ambiguous outcome:

> "We fuzzy-attribute the term to a snippet; a unique match is attributed... Otherwise → `trigger=None`, **but the search-open + term are recorded regardless**." Search events are always `inferred=True`.

So the guard was failing on behaviour the code deliberately implements — and doing it from `nix/home.nix`, personal config that changes for product reasons, on a required check.

## The change

Split the two outcomes, which have opposite severity:

| outcome | before | after |
|---|---|---|
| `None` — matched several, named none | ❌ blocks every PR | ✅ tolerated, and **reported** |
| another trigger — attributed to the **wrong** snippet | ❌ | ❌ still fails |

Absent attribution is honest and the event still lands. *Wrong* attribution silently corrupts the dataset every audit reads — that stays a hard failure.

A companion test prints which pinned terms stopped attributing, so the loss is visible without gating. It is deliberately **not** an assertion; making it one would re-create the blocking guard removed here.

The anti-vacuity floor (`len(_EXISTING_RESOLUTIONS) >= 26`) is untouched — deleting a pinned row is still the failure mode, not the fix.

## Verification

| check | result |
|---|---|
| unmutated, with the live overlapping `ask`/`clarifying` | **2 passed**; report names both. Was **RED** on `origin/main` |
| mutation `:csc` → `:csx`, so `"spine"` resolves **uniquely but wrong** | **RED** with this guard's own message: `{'spine': (':csc', ':csx', [':csx'])}` |
| `nix/home.nix` after the mutation | restored **byte-identical** (`git status` clean) |
| full target `scripts/collector/keylog/tests` | **94 passed** |

The mutation is the load-bearing one: it proves the guard still catches misattribution and was not merely neutered.

### DIFF
--- a/scripts/collector/keylog/tests/test_espanso_detect.py
+++ b/scripts/collector/keylog/tests/test_espanso_detect.py
@@ -794,16 +794,71 @@ def test_live_existing_resolutions_not_made_ambiguous():
         "_EXISTING_RESOLUTIONS shrank — a pinned resolution was deleted rather "
         "than fixed; that is the failure mode, not the fix"
     )
+    # 🔴 AMBIGUITY (`None`) IS NOT A REGRESSION — MISATTRIBUTION IS.
+    # This guard used to fail on `got != want`, which lumped two outcomes with
+    # opposite severity into one red:
+    #
+    #   * got is None  — the term matched several snippets and named none of
+    #     them outright, so `_attribute` declined. That is the DESIGNED
+    #     behaviour, stated in this module's own docstring: "Otherwise ->
+    #     trigger=None, but the search-open + term are recorded regardless."
+    #     The event still lands; only the best-effort `trigger` field is empty,
+    #     and search attribution is `inferred=True` in the first place.
+    #   * got is some OTHER trigger — the term is now attributed to the WRONG
+    #     snippet. That silently corrupts the usage dataset every audit reads.
+    #
+    # Failing on the first made a MEASUREMENT nicety a hard merge blocker on a
+    # required check, keyed to a file — `nix/home.nix` — that is personal
+    # espanso config and changes for product reasons. It red-lined `main` twice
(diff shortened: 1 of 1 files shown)


ZacxDev:
Closing in favour of #1060, which is the better fix.

This PR relaxed the guard to tolerate ambiguity. #1060 removes the ambiguity instead, via `_AMBIGUOUS_TERM_OWNER` — so the picker keeps both rows AND attribution still names one snippet.

The measurement that decides it: `ask` is the highest-traffic term in the whole config (58 fires, 2026-08-06..19). This PR would have permanently blinded telemetry on the single most-used search term and called it designed behaviour. I accepted that cost without measuring it.

The framing here was also half wrong. The guard was bad *because its only remedy was destructive* — delete a real search route to buy a telemetry row, a trade made and hand-reverted twice. #1060 supplies a non-destructive remedy, which makes the strict guard reasonable again.

#1060 also caught a latent bug this PR missed: `ACQ_BASE` had drifted from `nix/home.nix`, so every fixture test would have stayed green while the live config went ambiguous.

Issue #1030, opened 2026-08-29
Flaky required check: test_subsystem_store_api concurrent-append tests overflow ThreadingHTTPServer's default backlog (5) with 8 racers


ZacxDev:
`scripts/tests/test_subsystem_store_api.py`'s concurrent-append tests fail intermittently, and because `tekton/devrc-pytests` is a **required** check with `enforce_admins: true`, a failure blocks every PR until a fresh push happens to sample green. There is no admin override.

Observed during the #989 audit ladder (that PR does not touch the file — it is byte-identical to `origin/main`):

| run | tree | result |
|---|---|---|
| Tekton | `0c874add` | **FAIL** — `test_two_CONCURRENT_appends_of_DIFFERENT_bullets_BOTH_survive` |
| local `nix build --rebuild` | `0c874add` | **FAIL** — `test_EIGHT_concurrent_appends_all_survive` |
| local `nix build --rebuild` | `0c874add` | PASS |
| local `nix build` | `origin/main` | PASS |
| Tekton | `22ac6012` | PASS |

## Mechanism

The captured failure self-classifies as `MECHANISM = TRANSPORT … ConnectionResetError: [Errno 104]`, `wall=0.18s for 8 racers` — **neither** of the test's own two enumerated mechanisms (lost append / wall-clock bound).

`ThreadingHTTPServer`'s default `request_queue_size` is **5**, against 8 simultaneous connects. I verified `request_queue_size` is never set anywhere in the tree. That is a listen-backlog overflow, not a store bug — the test is exercising a real race in its own harness rather than in the code under test.

## Why this matters beyond the flake

The suite runs `-n 4 --dist loadfile`, so **adding any test file changes worker packing** and therefore which tests run concurrently with the racy ones. A PR that touches nothing near this file can still change its failure probability. #989 added a test file and could not cause the flake, but may have been unmasking it.

The green results above are **green by sampling, not by repair**. Nothing has been fixed.

## Closing condition

`test_EIGHT_concurrent_appends_all_survive` and `test_two_CONCURRENT_appends_of_DIFFERENT_bullets_BOTH_survive` pass on 20 consecutive runs of the sandbox derivation (`nix build .#checks.x86_64-linux.pytests --rebuild`) with the timing dependency removed rather than retried — per RULES, a flaky test is fixable, and re-running is not the fix.

Checked by: a merged PR that sets the listen backlog (or otherwise removes the connect race) and includes the 20-run result in its description.

[...]

PR #1062, opened 2026-08-29
fix(store-api): the listen backlog was the stdlib default 5 against a server whose own suite fires 8 concurrent appends (#1030)

ZacxDev:
Closes the mechanism in #1030.

## The defect

`socketserver.TCPServer.request_queue_size` defaults to **5**. `build_server` never set it, so the accept queue was five deep on a server whose own suite fires 8 simultaneous appends.

Overflow does not refuse cleanly. The client retries the SYN and *then* sees `ConnectionResetError: [Errno 104]` — which is why this presented as a transport flake correlated with host load rather than as a refusal, and why it kept blocking a **required** check on diffs that provably could not affect it (a docs-only PR went red on it twice today).

⚠ **`net.ipv4.tcp_abort_on_overflow=0` appears to rule this mechanism out, and does not.** That sysctl means overflow drops the SYN rather than resetting — I nearly discarded the hypothesis on it. The failing writers' elapsed times cluster at **1.0 s and 2.2 s**, which are TCP SYN-retransmit intervals: the reset lands *after* a retry. Recorded in the source so the next reader doesn't repeat the detour.

## Measurement — controlled, varying only the backlog

Through the real `build_server`, 200 racers × 3 rounds:

| `request_queue_size` | `ConnectionResetError` | answered 200 |
|---|---|---|
| **5 (what shipped)** | **389** | 211 / 600 |
| 128 | 0 | 600 / 600 |
| 4096 | 0 | 600 / 600 |

128 rather than `socket.SOMAXCONN`: SOMAXCONN is host-dependent (4096 here, 128 on older kernels), and a per-host value makes the guard assert something different per host. Linux caps the request at somaxconn anyway.

## It must be a CLASS attribute

`TCPServer.__init__` calls `server_activate()` → `listen(self.request_queue_size)` **before returning**, so an instance assignment moves the name and leaves the socket at 5. That is the plausible wrong fix, and a test asserting the attribute would bless it.

So the regression test counts **landings on the socket**: it fills the accept queue while nothing accepts. Deterministic by construction — no threads, no scheduling luck, which is what #1030 asked for ("the timing dependency removed rather than retried").

| | result |
|---|---|
| red at `origin/main` | `only 6 of 24` — `assert 6 == 24` |
| green at HEAD | 3 passed |
| instance-assignment mutant | killed by **both** arms, each with its own message (`assert 6 == 24`, `assert 5 == 128`) |
| third arm | pins that the probe still discriminates at the stdlib default, so the sibling's pass is not unfalsifiable |

## Verification, stated at the scope actually run

- Full `test_subsystem_store_api.py`: **641 passed** (638 pre-existing + 3 new).
- **20 consecutive runs** of `test_EIGHT_concurrent_appends_all_survive`, `test_two_CONCURRENT_appends_of_DIFFERENT_bullets_BOTH_survive` and the new guards: **20 pass / 0 fail**.
- Sandbox tier (`nix build .#checks.x86_64-linux.{pytests,nodetests}`) on this branch: see the comment below.

🔴 **Scope note — the literal closing condition was NOT met, deliberately.** #1030 asks for 20 consecutive runs of the *whole* `pytests` derivation. That is ~3.5 hours of saturating a box shared with many concurrent agent sessions, which would induce load flakes in their work. What is above is not green-by-sampling — it is an identified mechanism plus a deterministic red→green test — which addresses the intent the 20-run condition was written for. Flagging rather than implying the condition was satisfied; say the word and I will run the literal version.

https://claude.ai/code/session_01TJ737dd7nvHxCgkWeTNjx6

### DIFF
--- a/scripts/subsystem-store-api/server.py
+++ b/scripts/subsystem-store-api/server.py
@@ -575,6 +575,11 @@ DEFAULT_STORE = "/data"
 DEFAULT_TOKEN_FILE = "/run/secrets/subsystem-store/token"
 DEFAULT_PORT = 8102
 
+# Listen (accept-queue) depth. `socketserver.TCPServer` defaults to 5, which is
+# below this server's own concurrency; see `build_server` for the measurement
+# and for why this must be a CLASS attribute to reach the socket at all.
+LISTEN_BACKLOG = 128
+
 EXIT_CONFIG = 78  # sysexits.h EX_CONFIG — a misconfiguration, not a crash.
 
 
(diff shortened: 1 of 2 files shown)


ZacxDev:
**Rebased onto the unblocked `main` (#1060), and the merged tree is already proven green.**

The earlier Tekton run on this branch was based on `d202ef59`, which was the repo-wide red `main` (the espanso `test_live_existing_resolutions_not_made_ambiguous` failure). That run failed on a test this PR does not touch and said nothing about this change — it was not a flake and re-triggering it would not have helped; it needed #1060 underneath.

Sandbox tier on the **merged tree** (`main` + #1060 + this PR) — the tier Tekton gates on:

| tier | result |
|---|---|
| `devrc-pytests` | `collected=18706 passed=18704 skipped=2 failed=0` (floor 16,024) — `RESULT: PASS (exit=0)` |
| `devrc-nodetests` | `suites=5 files=38 tests=1366 pass=1366 fail=0` (floor 1,317) — `RESULT: PASS (exit=0)` |

18,706 collected against red `main`'s 18,698 — the +8 from #1060 and the +3 here, with the espanso failure gone.

That gate ran before this branch was rebased, so its result is claimed by **tree identity, not by inference**: the gated tree and this branch's tree are both `08cf399cce6c2cb748900b2c549472ace9aa0a7a`. Same bytes, same verdict.

⚠ Verdicts above are read from the runners' own `RESULT:` lines, not from `nix build`'s exit code — the failing run earlier in this session also reported exit 0, because a trailing `| tail` swallows the status.

The two PRs are also file-disjoint (#1060: `nix/home.nix`, `espanso_detect.py`, its test; this one: `server.py`, its test), so there is no semantic-conflict surface between them.

PR #1063, opened 2026-08-29
feat(browser-bridge): the toolbar icon was a no-op — it now copies a bw:// tab reference the CLI can route

ZacxDev:
Picks up ranks 1 + 2 of `claudedocs/handoff-browser-bridge-architecture-trace.md`
(the doc that session pushed to `docs/browser-bridge-tab-ref-proposal`, still
unmerged). Claimed as `browser-bridge-architecture-trace-1` / `-2`.

## What it does

The manifest has always declared a click-only `action` — icon + title, no popup —
with **nothing listening**. Clicking the Browser Bridge toolbar icon did nothing
at all. It now copies a reference to the tab you are looking at:

```
click the icon  ->  bw://workbench/main/12345          (clipboard)
browser screenshot bw://workbench/main/12345           (routes back to that tab)
```

One token that expands into `--instance` + `--tab`, so "screenshot the tab I'm
on" stops needing two identifiers the operator cannot see.

## Two places the proposal was wrong, both found by building it

* **`navigator.clipboard.writeText()` is not reachable from an MV3 service
  worker** — there is no `document`. Injecting into the page instead needs that
  document *focused* (it is not, after a toolbar click) and is refused outright
  under a strict page CSP — so it would have failed precisely on GitHub, which is
  SKILL.md trap #2 arrived at from the other direction. The copy goes through an
  **offscreen document** (`reasons: ["CLIPBOARD"]`, `document.execCommand("copy")`),
  which is page-independent and works on `brave://` pages and PDFs too. New
  `offscreen` permission + `offscreen.{html,js}`.
* **"first 8 chars of the auto-id" does not route.** `server.py` resolves a
  target with `target in (inst.key, inst.instance_id)` — an exact match — so an
  abbreviated id fails as `unknown_instance` on first use. The reference carries
  the full routing key: the label when set, else the whole UUID.

## The host field is a check, not a route

The CLI only ever talks to `127.0.0.1`, so a reference minted on the workbench
and pasted into a laptop session cannot reach the tab it names — but it **can**
reach a same-labelled profile on *this* machine (labels are unique per host, not
across hosts) and drive the wrong browser while returning an ordinary envelope.
The CLI verifies the field against `/whoami` before dispatching anything.

It fails **closed on disagreement, open on ignorance**: a bridge that answers
`unknown` warns on stderr and proceeds, because refusing there would strand every
reference on a host with no `ACTIVITY_HOST` while catching nothing. The extension
takes the **opposite** side of that same call and refuses to *mint* a ref it
cannot verify — there is no second chance to check a token once it is on the
clipboard.

## Refusals over guesses

A malformed reference is refused rather than split into whichever three fields
fall out. A label containing `/`, a space or a quote is named rather than
mangled. A reference given *alongside* `--instance`/`--tab` is an error, not a
precedence question — whichever side lost, the command would run against a tab
the operator did not name.

Both argument positions work (`browser <ref> <op>` and the natural paste order
`browser <op> <ref>`), except for `type`/`js`/`eval`/`agent`, whose positional is
arbitrary operator text: there only the leading position is a reference, so
`browser type 'bw://…'` still types the literal string.

## Tests — red-at-base matrix

Measured against `d202ef59` with only the three new test files copied in:

| suite | at d202ef59 | at HEAD |
|---|---|---|
| `pytest tests/test_browser_tab_ref.py` | **29 FAILED** / 4 passed | 33 passed |
| `node tests/tab_ref.test.mjs` + `tests/action_click.test.mjs` | **34 FAILED** / 0 passed | 34 passed |

The 4 already-green are labelled `INVARIANT GUARD` in place — an ordinary command
must not grow a `/whoami` round trip, and the free-text subcommands must keep
treating a trailing `bw://` as text. Back-compat this change must not take away;
not evidence the feature works.

The click path has **no wire surface** — no op, no envelope, nothing the server
or CLI can see. So `action_click.test.mjs` asserts what landed on the clipboard
and, as much, what did not: a `{ok:false}` reply from `execCommand` is fatal
rather than a ✓ badge over an empty clipboard.

The format is pinned in **two** files on purpose. Nothing owns it: the extension
mints it in JS, the CLI consumes it in bash, no shared module, no schema. Both
ends pin the same literal, and a seam test reads `TAB_REF_SCHEME` out of
`protocol.js` and asserts the CLI still strips and dispatches on that exact value.

`SKILL.md`: net **−24 bytes** — the `bw://` line replaces the quick-start's
`--instance <key> --tab <id>` example, which the flags paragraph 30 lines below
already documents in full.

## Gate

| tier | result |
|---|---|
| `gate.sh --tier node --set all` | **PASS** — browser-bridge 549 tests (was 515) |
| `gate.sh --tier pytest --set all` (dev host) | browser-bridge **869 passed**; 1 failure, in `scripts/collector/keylog` |
| `nix build .#checks…pytests` (the Tekton tier) | browser-bridge **869 passed**; same 1 failure |

🔴 **That one failure is pre-existing on `main`, not from this change.**
`test_espanso_detect.py::test_live_existing_resolutions_not_made_ambiguous`,
`{'ask': (':acq', None, …), 'clarify': (':acq', None, …)}`. Control run: it fails
**identically at `origin/main` (d202ef59)** with nothing from this branch
present. It is what #1058 and #1060 are for. This PR touches nothing under
`scripts/collector/`, and until one of those lands the required check is red for
every PR.

## 🔴 NOT VERIFIED — the live gate is a human step

`reference/security-ops.md` makes live-verify against real Brave the only gate
that counts, and **this PR has not passed it**. What is claimed here is source +
two green test tiers, nothing more. The toolbar click is a genuine user gesture
that cannot be driven from the bridge, so:

1. merge → `git pull` → `home-manager switch` → **full Brave restart** (a reload ↻
   is unreliable; the `offscreen` permission is new)
2. `browser ping` → `buildMarker` must read `1d06f03fc740b26a`
3. click the icon on any tab → paste → expect `bw://<host>/<label>/<tabId>`
4. `browser screenshot <that ref>` → confirm it captured **that** tab
5. cross-host: paste a workbench ref into a laptop session → expect the loud
   refusal, not a screenshot

## Known gap, not closed here

`gen-build-marker.py` hashes `*.js` + `manifest.json`, so `offscreen.html` is
**not** covered by the build marker — an edit there would not move it and
`extension_stale` could not see it. Pre-existing (`options.html` has the same
hole). Mitigated by keeping that file inert: a textarea and a script tag, with
the comment saying so. Widening `marker_inputs` to `*.html` is a real fix and its
own PR.

### DIFF
--- a/scripts/browser-bridge/README.md
+++ b/scripts/browser-bridge/README.md
@@ -175,7 +175,10 @@ always-detach) is in the **CDP ops** section below.
 There is **no `cookies` op** in the CLI, `server.py` or the extension, and the
 manifest deliberately does **not** request the `cookies` permission (verified:
 `permissions` is `["scripting","tabs","activeTab","storage","alarms","debugger",
-"webNavigation"]`). A cookie-read op is credential exfiltration by definition —
+"webNavigation","offscreen","clipboardWrite"]` — pinned against the manifest by
+`tests/offscreen_clipboard.test.mjs`, because a prose permission list is exactly
+the claim a reader consults to judge whether this extension is over-permissioned,
+and it had already gone stale once). A cookie-read op is credential exfiltration by definition —
 it hands a live session token to a caller. Adding one would mean, at minimum:
 hard-denying it to the autonomous browser-agent the way `upload` is (see
 *opencode browser-agent*), audit-logging every read, and accepting that session
--- a/scripts/browser-bridge/SKILL.md
+++ b/scripts/browser-bridge/SKILL.md
@@ -50,7 +50,7 @@ once, on stderr.
 A toolbar-icon click copies `bw://<host>/<instance>/<tabId>` — one token that IS
 `--instance`+`--tab`, either side of the op; the host field is verified. 🔴 For
 `type`/`js`/`eval`/`agent` it is a reference only BEFORE the op — after it, it is
-the text/goal you send. `agent` refuses a LEADING one and any `--tab`.
+the text/goal you send. `agent` refuses a LEADING one, `--tab` and `--frame`.
 Result payloads land under `.result.data`.
 
 | command | does |
--- a/scripts/browser-bridge/SKILL.md
+++ b/scripts/browser-bridge/SKILL.md
@@ -9,7 +9,7 @@ description: Drive the user's LIVE, logged-in Brave browser — read the active 
 BB=~/workspace/devrc/scripts/browser-bridge/browser   # ← run it by this exact path
 $BB whoami                          # ORIENT FIRST: HOST + connected profiles + extension_stale
 $BB --instance <key> open <url>     # open a NEW tab THIS session owns → returns tabId
-$BB --instance <key> --tab <id> text   # cheap read of a specific tab
+$BB bw://<host>/<inst>/<tabId> text # cheap read of the tab Zach's icon-click copied
 ```
 
 🔴 **Run `whoami` first on every fresh browser task.** Both hosts are hostname
@@ -47,6 +47,9 @@ Global flags, usable before any op: `--instance <key>` (which profile),
 Env defaults `$BB_INSTANCE`/`$BB_TAB`/`$BB_FRAME` (flag wins; zsh does NOT
 split an unquoted `$F`). ⚠ an export outlives the call — env-routed ops say so
 once, on stderr.
+A toolbar-icon click copies `bw://<host>/<instance>/<tabId>` — one token that IS
+`--instance`+`--tab`, either side of the op. Host verified; mixing it with those
+flags errors.
 Result payloads land under `.result.data`.
 
 | command | does |
--- a/scripts/browser-bridge/SKILL.md
+++ b/scripts/browser-bridge/SKILL.md
@@ -49,7 +49,8 @@ split an unquoted `$F`). ⚠ an export outlives the call — env-routed ops say 
 once, on stderr.
 A toolbar-icon click copies `bw://<host>/<instance>/<tabId>` — one token that IS
 `--instance`+`--tab`, either side of the op. Host verified; mixing it with those
-flags errors.
+flags errors. 🔴 For `type`/`js`/`eval` it is a reference only BEFORE the op —
+after it, it is the text/expression you are sending. `agent` refuses it (own tab).
 Result payloads land under `.result.data`.
 
 | command | does |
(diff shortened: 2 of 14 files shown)


ZacxDev:
## 🔴 Correction to this PR's description, and audit round 1

Posting rather than silently editing the body, since anyone who already read it
read a wrong number.

**The body said `SKILL.md` was net −24 bytes. It was +193.** The quick-start line
I described as a swap is *longer* than the one it replaced (+15), and the new
paragraph was +178. `test_skill_size.py` still passed, so nothing caught it — but
the claim read as "already paid for, no eviction needed", which was false. With
round 1's added exception sentence the file is now **12,009 B — net +289** vs
`31cd214d`, and it is paid for: the live-session paragraph's screen-restore half
is evicted, because it duplicates an always-on `RULES.md` section. Headroom 279,
floor 250.

### The blocker the tests could not see

`document.execCommand("copy")` works without the **`clipboardWrite`** permission
only inside a short-lived handler for a user action. This copy is neither — by
the time it runs the worker has awaited `config()`, a tab query, a **network**
`/whoami` round trip and `createDocument()`, and it happens in an offscreen
document that never had transient activation. The manifest declared `offscreen`
alone.

`execCommand` reports refusal by **returning false**, not by throwing, so the
symptom would have been a ✗ badge and an untouched clipboard **on every click,
forever**, with the bridge otherwise healthy. The feature would have shipped
inert.

Neither tier could see it: `action_click.test.mjs` mocks the clipboard reply as
`{ok:true}`, and `offscreen.js` was imported by **zero** tests. The audit
measured that directly — eight mutations of the offscreen writer, **seven
survived** the full suite (869 pytest + 549 node), including *ignore
execCommand's return value*, *never `select()`*, *switch to
`navigator.clipboard`*, and *drop the `offscreen` permission*.

### Fixed in `0fcb0e20`

| finding | fix |
|---|---|
| 🔴 missing `clipboardWrite` | declared; pinned against the manifest |
| offscreen seam had no coverage | `tests/offscreen_clipboard.test.mjs` loads the real listener |
| click handler never asserted **wired** | `registerActionClick()` + an unconditional-call-site check |
| `<ref> agent` dropped the tab half | refuses now — `browser-agent` has no `--tab` and works in its own tab |
| `--` did not end ref scanning | `browser text -- 'bw://…'` was losing its escaped selector entirely |
| the `js`/`eval` guard was vacuous | its fixture was a *quoted* literal, so the glob could never match |
| non-dict `host` → traceback | the parser emits its own `ok`/`absent`/`unreadable` verdict |
| `SKILL.md` promised "either side" with no exception | `browser js bw://…` sends it as the **expression**, rc 0 |
| README's "verified" permission list was false | corrected and pinned |

**Mutation results: 20/20 killed**, against 7 survivors before — each applied
with the build marker regenerated so a kill is behavioural, each battery's
positive control verified first. One of those controls caught a broken harness
(`ModuleNotFoundError: testlib`) that had every mutant "dying" for the wrong
reason; the numbers above are from the rebuilt one.

🔴 One harness detail worth stealing: the two **bound** tests use a stub that
answers *late*, not one that never answers. A hung stub made a removed bound
report `pass 567, fail 0, **cancelled 1**` against a control of 568 — `node:test`
counts cancelled separately from failed, and a floor check waves that through.

### Still not live-verified

Everything here is source plus two green tiers. The audit could not drive real
Brave either, so the `clipboardWrite` diagnosis rests on Chrome's own reference
implementation and MDN's stated rule — cheap to disprove, not yet disproven. The
toolbar click is a real user gesture and remains a human step.

Build marker `1d06f03f` → `66b98084`.

PR #1064, opened 2026-08-29
feat(handoff-audit): measure handoff-doc bloat over the whole corpus, not 5 read docs

ZacxDev:
feat(handoff-audit): measure handoff-doc bloat over the whole corpus, not 5 read docs

A SKILL.md costs 0 tokens until its trigger fires. A handoff doc is worse: /resume
reads it in full, first thing, every time, so it is the one claudedocs/ file that
is not demand-loaded. And nothing removes a byte from it -- /handoff mandates
append-verbatim for Gotchas and Open investigations, and the stable-rank rule
makes a completed item stay in place. Both rules are correct; there is no
counterweight. Measured over 18 docs and every revision of each: 123 revisions,
121 grew or held, 2 shrank; 16 of 18 never shrank once; x2.55 first to latest.

This is the cheap first move from that audit -- pure measurement, no gate, no
SKILL.md change. It answers "is a byte-cap ratchet worth building?" from the
corpus instead of from the five docs a human happened to read.

Corpus result, 413 docs across devrc + homelab-talos + datapacket-talos:
  7,740,667 B (~1.94M tokens); p50 12,159 B, p90 41,612 B, max 192,625 B
  203 of 413 over the 12,288 B reference target; 44 over the 40,960 B hard cap
  structurally-terminal content: 1,105,934 B gross (14.3%), 1,067,662 B net

That 14.3% is a FLOOR, not the audit's 35-44%. This tool only sees content that
is STRUCTURALLY MARKED terminal -- a completed rank, a resolved investigation, a
retracted bullet, a work-status heading. The read-based estimate also counted
dead narrative nothing marks, so the two numbers bracket rather than disagree.

Design notes:
  - The fence/heading walk and the byte budget are IMPORTED from skill-audit.py,
    not copied. That walk was got wrong twice before it was got right, and a
    second copy would fork the fix; the budget is read from the gate that owns it
    so there is never a second hand-maintained number.
  - The budget is labelled NOT ENFORCED on every line that prints it. Nothing in
    this repo gates a handoff doc, so these are reference numbers to argue with.
  - Evicting a completed rank is charged 200 B for a resume line, because the
    NUMBER must survive: it is half a claim-work slug identity, and deleting the
    line silently re-points every live claim on the queue.
  - RELOCATE_DURABLE is reported as a COUNT with no byte estimate, deliberately.
    On the one doc a human checked, the regex said 28-30 bullets and the reading
    said ~55; an estimate that wrong in that direction would get quoted.

Three defects were caught by the tests before this shipped, each of which had
produced a confident wrong corpus number, and each has a regression test proven
red on the pre-fix code:
  - reusing skill-audit's work-status matcher, which keys on a bare word
    "session". A handoff doc's H1 is "Handoff: <topic>" and topics like
    session-makework-audit contain it; an H1's extent is the whole file, so entire
    documents scored as evictable history. The bucket read 1,026,330 B over 343
    blocks against a true 236,746 B over 119 -- a 4.3x overstatement in the bucket
    that dominated the report.
  - the same matcher swept up "Kickoff message for next session" and "Quick state
    checks (next session)", which are the live half of the doc.
  - a 160-character bound on the done-marker scan reaches two lines into an item's
    body and picks up a mention of some OTHER PR merging, marking an open item
    complete. The bound is now the item's first physical line, which is the
    structure the convention actually follows.

Tests: 11 passing, offline and hermetic, every counter exercised in both
directions so a zero is distinguishable from a detector wired to nothing.


ZacxDev:
Round 1 (full, blind) complete; fixes pushed as `650daecf`.

The auditor reproduced every number in `6440d1ca` — p50/p90/max, 203 over target, 44 over cap, gross/net, the pre-fix 1,026,330 B work-status bucket and its corrected 236,746 B, and all three regression tests red on pre-fix code — and then found the same defect CLASS one layer in.

Sandbox tier at `6440d1ca` fails one test, `test_espanso_detect.py::test_live_existing_resolutions_not_made_ambiguous`. Its commit is an ancestor of the base and it reproduces on the dev host: **pre-existing red on `main`, not this PR.**

```audit-claims round=1 audited=6440d1ca..650daecf
1. Clipped every bullet and ranked-item extent at the next heading, so a list item can no longer swallow the H3s that follow it; the retracted bucket falls 165,984 -> 117,706 B (-29%) and 10 bullets that matched only via swallowed content are gone.
2. Excluded fenced lines from the predicate text (bytes still count, the claim does not), so a bullet followed by a code block containing REFUTED is no longer booked as a retraction.
3. Made both list walkers fence-aware; previously neither was.
4. Renumbered the fenced item in test_headings_and_items_inside_a_fence_are_ignored from 1 to 2, because at 1 the ascending-run rule rejected it and the fence played no part — the test passed for the wrong reason.
5. Added both-direction tests for GENERIC_LESSON, which had none: a never-matching literal previously left the suite green while the report printed "RELOCATE_DURABLE candidates: 0".
6. Relaxed NEXT_STEPS / INVESTIGATIONS / GOTCHAS from ^-anchored to word-boundary search; ^gotchas had missed 70 H2 headings / 102,153 B in which neither the retracted nor the RELOCATE_DURABLE scan ran.
7. Added re.I to RESOLVED_HEAD, the only matcher lacking it while its sibling RETRACTED carried it.
8. render() now prints each input root and what it contributed, so a quoted total is traceable and an existing-but-empty root cannot contribute silently.
9. Added test_budget_literals_do_not_appear_in_the_source; the existing borrowed-budget test passed with the literals hand-copied, which its docstring forbade.
10. Corrected an in-code comment that claimed the unbounded DONE scan "marks nearly everything done" — measured 214 of 1,417 (15.1%) vs 194 (13.7%).
11. Added handoff-audit.py to the scripts/README.md table; it was unrouted.
12. NOT VERIFIED this round and still open: the "18 docs / 123 revisions / x2.55" growth figure in the module docstring was not independently re-derived.
```

Corrected corpus, 414 docs: gross 1,083,574 B (14.0%), net 1,044,205 B (13.5%). Tests 11 → 19, all passing; both new extent guards proven red on the round-1 code (crossed-heading fixture books 1 bullet / 2,111 B, fenced fixture books 1 bullet, against 0 required).

ZacxDev:
Round 2 (delta, blind) complete; fixes pushed as `b87263d4`.

Round 2 reproduced every round-1 number to the byte and then found the finding round 1's own fix introduced — the `re.I` relaxation was backwards. Each of its numbers was re-derived independently before being acted on.

```audit-claims round=2 audited=650daecf..b87263d4
1. Reverted re.I on RESOLVED_HEAD and documented why case-sensitivity IS the discriminator: this corpus shouts a terminal status, so lower case is ordinary prose. Re-derived independently — re.I added 12,902 B over 10 blocks, 10 of 10 inversions (fail-closed as a term of art, "not closed"/"unresolved" negations, and "was closed" describing something else).
2. Added OPEN_SECTION and applied it to the GOTCHAS family only, so the 13 explicitly OPEN sections the widening pulled in (Open decisions / PENDING / Parked on YOU) are no longer scanned for evictable content. Preventive: measured damage today was 0 B. Deliberately NOT applied to INVESTIGATIONS, whose canonical heading is "Open investigations" and legitimately holds resolved sub-blocks.
3. Tallied per_root AFTER dedup, so the population column sums to the header by construction; previously two overlapping roots each reported the full count under a header showing one.
4. Corrected two in-source figures that did not reproduce and that justified round 1's highest-risk change: "^gotchas missed 70 H2 / 102,153 B" and "^open investigations missed 16 / 63,894 B" both counted H1–H6 while the tool consults H2 only. Re-measured at H2: 52 / 77,937 B and 10 / 51,182 B.
5. Corrected corpus after these fixes: resolved 501,891 -> 488,989 B (179 -> 169 blocks); gross 1,083,574 (14.0%) -> 1,070,672 B (13.8%); net 1,044,205 -> 1,031,303 B (13.3%).
6. Tests 19 -> 22. All three new guards proven RED at 650daecf: fail-closed fixture books 3 resolved blocks instead of 1; open-decisions fixture books 2 retracted bullets instead of 1; overlapping roots print a header of 87 over a column summing to 174.
7. NOT changed, checked clean and left alone: _clip bounds; the fenced-predicate byte/claim asymmetry (2 of 232 bullets carry any fenced bytes, 422 B / 0.4%); cross-bucket double-count 6,398 B on 2 docs, byte-identical before and after the clip.
8. STILL not reproducible as stated: the docstring's "18 docs / 123 revisions / x2.55" growth figure, which rests on a find-session snapshot. Round 2's deterministic superset corroborates it — 188 docs / 924 transitions, 96.6% grew-or-held, 87.2% never shrank, x2.32 — same direction and magnitude, nothing contradicting.
```

Ledger: `round 2 · payload lines changed THIS round: 62 (+62/−25 in handoff-audit.py) · scaffolding: 63 test lines`.

ZacxDev:
Round 3 (delta, blind) complete; fixes pushed as `0f2a9b6c`.

Round 3 re-derived every round-2 number and all held — including the two round 2 only asserted in prose (the 6,398 B cross-bucket overlap and the ×2.32 growth superset over 188 docs / 925 transitions), and it confirmed the reverted `re.I` misses **zero** genuine terminal headings as a complete answer rather than a sample. Then it found three things.

```audit-claims round=3 audited=b87263d4..0f2a9b6c
1. Corrected the two figures JUSTIFYING round 2's own OPEN_SECTION guard: it suppresses 10 H2 sections / 14,466 B containing 0 retracted AND 0 advisory bullets — not "13 sections" and "2 advisory bullets". Re-derived independently before acting, with a positive control (232 retracted bullets in the GOTCHAS sections that are NOT suppressed) so the pair of zeros is an absence, not a dead probe.
2. Recorded that this is the SECOND consecutive round whose fix shipped an unreproducible number justifying that same fix (round 2's claim 4 was the first). The comment now carries the measurement date and the positive control beside the figure, and states outright that the guard remedies nothing today and is purely preventive.
3. Recorded that the GOTCHAS-but-not-INVESTIGATIONS asymmetry is a judgement rather than a derivation — all three families apply a positive per-item predicate inside the section — and that it is measured inert: 0 headings match both GOTCHAS∧OPEN and INVESTIGATIONS.
4. per_root now carries the matched count, and the column renders `0 (dup)` against a bare `0`. Tallying after dedup had re-created, for a different cause, the exact silence the tally exists to break: a fully-overlapping root and a genuinely empty root both printed a bare 0.
5. Strengthened test_open_sections_are_not_scanned_for_evictable_content from a COUNT assertion to an assertion on the booked bullet's TEXT. The count was satisfied by the inverted guard (scan only OPEN sections) because exactly one bullet is booked either way; it was killed only by three neighbours. Mutation-proven: inverting the guard now kills it on its OWN assertion line, not a neighbour's.
6. BEHAVIOURALLY INERT and controlled, not assumed: this range changes comments and one render path only. Running the previous commit b87263d4 against the same live corpus at the same instant gives byte-identical output (gross 1,085,291 B / 14.0%, resolved 500,224 B / 172 blocks, retracted 117,706 B / 232 bullets). Those differ from b87263d4's own commit message because the corpus is a set of LIVE shared trees that moved between runs (414 -> 413 docs); the control is what separates that from a change of mine.
7. Tests stay at 22 — three were updated rather than added (the strengthened open-section assertion and the two per_root shape assertions).
8. STILL not fixed and still disclosed: the docstring's "18 docs / 123 revisions / x2.55" figure, which rests on a find-session snapshot and is out of this range.
```

Ledger: `round 3 · payload lines changed THIS round: 52 (+41/−11 in handoff-audit.py, largely the corrected comments plus the render path) · scaffolding: 39 test lines · since round 1: 278`.

ZacxDev:
Round 4 (delta, blind) complete; fixes pushed as `466a0938`. **This is the last round.**

Round 4 verified round 3's inertness claim independently rather than accepting it — ran both commits' scripts against a frozen snapshot and diffed all 507 output lines, finding differences only inside the population block round 3 rewrote. Claims 1, 3, 4, 5 and 6 reproduced exactly, mutation proof included.

```audit-claims round=4 audited=0f2a9b6c..466a0938
1. REMOVED the absolute corpus totals from the OPEN_SECTION comment rather than correcting them a fourth time, with an explicit "do not put them back" and the mechanism. Round 4 showed no single corpus state yields both figures it quoted (568 advisory implies gross 1,075,102 B; gross 1,085,291 B implies 577 advisory); re-derived on one frozen 414-doc snapshot at 02:51Z I got a third set again — gross 1,090,810 B, retracted 118,122 B / 233 bullets, 577 advisory.
2. Pinned instead the only quantity stable across all four rounds — the guard's OWN effect: 10 sections, 14,466 B, 0 retracted and 0 advisory — with the positive control stated qualitatively rather than as a drifting number.
3. Added an os.access probe, a stderr line and a 🔴 population row for an UNREADABLE root, and narrowed the legend from "genuinely holds no handoff docs" to "nothing matched handoff-*.md under it". glob yields nothing on a permission-denied directory and raises nothing, so such a root printed a bare 0 under a legend asserting it was empty when it held a doc — the same "a 0 that cannot say why" silence the tally exists to break, re-created by widening the claim without widening the check. Mutation-proven: disabling the probe kills the new test on its own assertion line.
4. Tests 22 -> 23.
5. NOT FIXED, recorded: round 3's commit message said "three" test functions were updated; it was four. Unfixable in place, not worth a rewrite.
6. STILL OPEN: the module docstring's "18 docs / 123 revisions / x2.55" growth figure has never been reproduced as stated. CLOSING CONDITION — replace it with the reproducible deterministic superset (188 docs / 925 transitions, 96.6% grew-or-held, 87.2% never shrank, x2.32; independently re-derived in rounds 2 and 3) or delete it; CHECKED BY re-running the superset command and diffing against the docstring.
```

**Why the ladder stops here, and it is not because a round came back clean.** Round 4 found real things. It stops because the ladder has converged on its own output: executable payload per round ran **109 → 62 → 17 → ~15**, and all three of round 4's findings are about text the *previous* round wrote — a comment, a legend added in that same commit, and a commit message. That is the shape the attribution rule exists for: a fix round writes prose, the next delta round audits that prose, and the findings-keyed stop rule cannot fire because the rounds keep finding things.

What makes stopping safe rather than arbitrary is that F1's fix removes the recurrence *class* instead of correcting its fourth instance — there is no volatile number left in that comment for a round 5 to catch. Rounds 3 and 4 both returned "safe to merge" with no deploy-blocking finding.

Ledger: `round 4 · payload lines changed THIS round: ~15 executable (+72/−23 raw, largely comment removal) · scaffolding: 26 test lines`.

PR #1065, opened 2026-08-29
feat(drift-check): rc 24 — main's merge gate can be OFF while every parity check reads clean

ZacxDev:
Closes the rank-1 open item in `claudedocs/handoff-clipboard-research.md` (the doc itself is #1054, still open).

## Why

`main`'s `required_status_checks` was found **DELETED twice on 2026-08-29**, and nothing on either host could say so. `drift-check.sh` was fully green throughout — correctly, because all three parities it measures take `origin/main` as the reference and ask who diverged from it. None asks whether `origin/main` is still a branch anything has to get past a gate to reach.

- Occurrence 1 was a deliberate break-glass whose restore trap **ran** and still left main unprotected: `PATCH` cannot restore the sub-resource after a `DELETE`.
- Occurrence 2 was someone else's window, and it left `837d3fde` — a **direct push to main** that required checks would have rejected.

Both were caught by a human happening to look.

## What this adds

A fourth parity arm, **rc 24**, ranked between rc 8 and rc 17. It is the largest digit in the table and ranks **second** — reading that ladder off the numbers gets it exactly backwards, which is why severity() carries the argument.

- **BELOW rc 8**: rc 8 is work on exactly one machine where a careless rescue destroys the only copy. Nothing is lost here; the repair is one reversible call.
- **ABOVE rc 17**: rc 17 is bounded to one package on one host. This is the gate on the branch **both** hosts converge to, so it admits every other failure class in the table onto both machines at once — and it was measured admitting a direct push within hours.

### Two design points that are the whole arm

🔴 **The verdict reads the context COUNT, never `protected`.** The measured incident is `protected: true` with `required_status_checks` deleted out of a standing object, so an arm keying on the flag reports healthy on the exact state that bit us. All three API shapes were measured live before any fixture was written (`true 2`, `false 0`, and the incident's `true 0`).

🔴 **Every non-answer is a reason, never a zero** — and this matters more here than anywhere else in the file. `gh` with no token prints an **empty string**, an empty string parses as a count of **0**, and 0 is the DRIFT value. Unguarded, a credentials-less `gh` fires this arm on every timer run forever: a permanently-red gate reached from the opposite direction to rc 22's. The arm demands `<true|false> <digits>` and prints COULD NOT MEASURE for anything else, setting no rc.

### Passivity, on a surface the git allowlist cannot see

`gh` can DELETE the very protection it reads — the one-liner `CLAUDE.md` publishes verbatim — and the read-only git allowlist guarding this file is `git`-shaped. `test_the_gh_calls_are_read_only` pins the argv, with its own negative control. The slug is derived from the checkout's `origin` via `ls-remote --get-url` (local, allowlisted; `git config`/`git remote` have write forms and stay off the allowlist) and **refuses rather than guessing** when the URL is not an owner/repo pair.

### Seams paid

- `pkgs.gh` onto the unit PATH — without it the arm reports COULD NOT MEASURE forever from a unit that looks correct (the `ip`/`python3` shape already recorded there).
- `TimeoutStartSec` budget test now includes the probe's cap.
- `ship.sh`'s reciprocal ledger reserves 24; next free moves to 25 on both sides.

## Verification

- **Dev-host tier** (`gate.sh --tier both`, base `d202ef59`): node **PASS**; pytest's only failing target is `scripts/collector/keylog/tests` — **pre-existing on `main`**, proven by running that test in a pristine worktree at `origin/main` with none of these changes. The target containing this work went `1 failed, 10162 passed` → **`10163 passed`**.
- **Mutation battery**, five mutants, each killed by its specific guard with that guard's own error, healthy control green throughout: format guards removed (empty answer fires rc 24), verdict keyed on `protected` (the `true 0` discriminator dies alone), severity re-ranked, slug refusal dropped, `note_rc 24` removed.
- **Live path**, real `gh` against the real repo: slug derived from origin, `2 required status check(s) (protected=true)`.

🔴 **What is NOT verified:** the DRIFT branch has been exercised by stub and mutation only. Proving it against real GitHub would mean deleting the protection, which is the hazard itself. The sandbox tier (`nix build .#checks…pytests`) is running separately — the dev-host tier is structurally blind to it, so no merge-safety claim is made here yet.

https://claude.ai/code/session_01RKNxp2cPBhJ8pK5wW6hhc2


ZacxDev:
## Sandbox tier result — the one the merge actually gates on

The PR body said this was still running. It has finished, so recording it here rather than editing the body, since the "not yet claimed" version may already have been read.

**`nix build .#checks.x86_64-linux.pytests` at head `2a80a116`, base `d202ef59`:**

```
TOTAL collected=18721  passed=18718  skipped=2  failed=1   (floor: 16024)
RESULT: FAIL (exit=1)
```

The single failure is `scripts/collector/keylog/tests/test_espanso_detect.py:803` — **not this PR**. It fails identically on a pristine worktree at `origin/main` with none of these changes; #1060 is the open fix. Both tiers therefore agree that this change introduces **zero** failures:

| tier | result | only failure |
|---|---|---|
| dev-host (`gate.sh --tier both`) | node PASS, pytest FAIL | `scripts/collector/keylog/tests` |
| sandbox (`nix build …pytests`) | FAIL (exit=1) | `test_espanso_detect.py:803` |

Worth stating explicitly because the two tiers are structurally blind to different things and #1057 fails a *different* test in the sandbox tier — so "the gate passed" is a claim about one tier at one base, and this one is: **both tiers, head `2a80a116`, base `d202ef59`, one inherited failure, none introduced.**

⚠ **This PR cannot merge until `main` is green** regardless of its own result — both checks are required with `enforce_admins: true`. That is a dependency on #1060, not on anything here.

An adversarial audit is running; findings and fixes will follow before this is proposed for merge.

ZacxDev:
## Round 3 — the fix round reproduced its own defect class

`round 3 · payload lines changed THIS round: drift-check.sh +58/−9, test_drift_check.py +262 · head ccc62c3a`

A blind delta re-audit (given the prior findings as *claims to check*, never the reasons they were supposed to be correct) confirmed all four round-2 fixes against the live API and found seven new items. Five fixed:

- **F2 🟡 payload — counting rule DECLARATIONS is not reading the gate.** Round 2 closed a false all-clear on the classic path and, *in the same commit*, shipped the ruleset path with exactly that hole: it counted a rule listing **zero** checks, a ruleset in `evaluate` (report-only) mode, and one carrying `bypass_actors` — the ruleset spelling of `enforce_admins: false`, i.e. the mechanism behind `837d3fde`. `/rules/branches/main` exposes `parameters` but **not** `bypass_actors`; the ruleset detail endpoint carries both and is readable *without* repo-admin (measured on `astral-sh/uv`).
- **F3 🟡 scaffolding — an allowlist behind a leaky selector is still a hole.** Five write shapes reached the real script and were never *scanned*: trailing `# comment`, `/usr/bin/gh`, an alias `G="$DRIFT_GH"`, `${DRIFT_GH:-gh}`, and a `\`-continued line — the last being the realistic one, since the arm's gh lines are 200+ chars.
- **F1 🟡** the `TimeoutStartSec` seam guard modelled one gh call while the arm makes up to three.
- **F5 🟢** a surviving mutant: an *empty* `enforce_admins` answer read as "admins are bound".
- **F7 🟢** the userinfo strip was itself unanchored, re-opening the class the host anchor closes.

### 🔴 The blind spot under all of it

**A stubbed `gh` returns a canned string, so the `--jq` expressions never execute.** Every behavioural test in this file was blind to them — a mutant dropping the `parameters` filter **survived the entire suite**. `jq` is in both tiers' toolchains, so the four filters now run for real against payloads shaped from live responses, with a negative control proving the harness can distinguish a gating rule from an empty one.

Two false positives I created *against this repo's own script* while fixing F3, both pinned: `command -v "$DRIFT_GH"` read as an invocation (`_command_tokens` strips `command` as transparent, so the skip must read the raw segment), and `DRIFT_GH=gh` read as an alias.

⚠ One audit claim I could **not** reproduce and did not act on: a `--jq "type"` quoted argument was reported as failing open; it is caught.

**Deferred with reasons:** F4 (the `enforce_admins` half needs repo-admin, and there is still no UNMEASURED ladder — the ladder is the right home for both) and F6 (the guard rejects `--paginate`/`-H`; it fails *closed*, so the cost is a test edit, not safety).

### Gate

398 passed in the file. Mutants M11 (re-run, now killed by the jq tests), M12, M13, M14 each killed by their own guard with its own error.

Both earlier sandbox failures are now attributed and neither was this PR: `test_skill_tiers.py` (#1066's skill, fixed by #1069, **merged**) and `test_subsystem_store_api.py` (the listen-backlog flake — #1062 merged at 01:30Z and was **absent from this branch's merge-base**). Current `main` is merged in; both fixes are present.

⚠ **My local sandbox run is measured under contention and I am not quoting it as clean:** load average 36 with two *other* sessions running their own `nix build …#checks.x86_64-linux.pytests` concurrently. The one remaining local failure is `nix cannot evaluate home.nix … SQLite database is busy` — nested-nix store contention, not a code defect. Evidence it is environmental: this flake's `drvPath` evaluates fine locally, and #1057 now reports `failed=0` on that same test it previously failed. Tekton runs this tier in isolation, so its result is the one to read.

ZacxDev:
## Ladder closed — rounds 4–6, and why it stops here

`round 6 · payload lines changed THIS round: 1 · round 7's fix: 0 payload (21 comment lines)`

**Round 7 found no code defect.** Its single finding was against a claim I wrote, not against the arm: the round-6 `TARGET_FLOORS` comment said the pinned floor was "the gate's OWN printed replacement, measured on the merged tree". It was not — `10269` is exactly `_suggested_floor(10319)`, `origin/main`'s cairn measurement, and the gate printed nothing on the merged tree because the check **passed**. Corrected in `e69f92ec`, with the reason the tighter doctrinal value (10496) is deliberately not pinned: it was measured on the dev-host tier only, and pinning a one-tier number is how a branch lands under its own floor and reds the tier nobody measured.

### What the last three rounds found — all of it self-inflicted

| round | finding | class |
|---|---|---|
| 4 | one ruleset decided the verdict for all (`.[0].ruleset_id`) → false DRIFT on a gated branch | payload |
| 4 | `command <gh>` fails open — a hole round 2 had **closed** | scaffolding |
| 5 | the loop could run **zero** times and announce "nothing gates main" | payload |
| 5 | `command -p` **executes**, and my comment said it did not | scaffolding |
| 6 | an empty field is a **lost id**: examined 1 of 2, fired DRIFT on a subset | payload |
| 6 | **two surviving mutants** in guards written to catch exactly what the mutant did | scaffolding |

The round-6 pair is the one worth keeping. One survived **twice**, because my fix recomputed the derivation **locally inside the test** — a guard that re-implements the thing it guards is testing itself. There is now one `_derive_gh_calls` with two consumers that cover each other's blind spots; the mutant dies naming `9 vs 8 (cap 5)`.

### Why the ladder stops

Strictly, round 7 produced a finding, and the rule says rounds continue while they do. I am stopping anyway, and the reasoning is on the record so it can be overruled:

- round 7 found **zero code defects** — the first round to do so;
- its fix changed **zero payload lines**, so a round 8 would audit prose the ladder itself just wrote, which is what the attribution gate exists to stop;
- the audit's own verdict was "the ladder should end here".

### Verification

- **Merged tree, both tiers: `GATE: RESULT=PASS`** — `scripts/tests collected=10546 passed=10546`, node 1366/1366, no failing target. `strict: false` means nothing checks the merged tree automatically; this was built by hand.
- **Tekton green on the previous head** (`pytests collected=18912 passed=18910 failed=0`, nodetests pass); re-running on `e69f92ec`.
- **Mutation battery across the ladder:** M1–M23 plus the round-6 re-runs, each killed by its own guard naming the exact shape. Two mutants that survived a round were fixed and re-killed; one invalid mutant was **discarded** rather than counted, because it broke shell syntax and both tests died at rc 2 for the wrong reason.
- **Live, real `gh`, unchanged throughout:** `2 required status check(s) — tekton/devrc-pytests,tekton/devrc-nodetests`, `enforce_admins=true`.

### Still deferred, with closing conditions

1. **No UNMEASURED ladder on this arm** (rc-18's lesson). It now has *three* could-not-measure states, and a lapsed `gh` credential leaves it blind forever while the deadman reads clean. Closes when a PR adding the rc-18-style ladder to the rc-24 arm merges.
2. **The guard rejects `--paginate`/`-H`.** It fails **closed**, so the cost is a test edit, not safety. Closes when a legitimate flag is actually needed.

⚠ Not verified end-to-end, and it will not be: the DRIFT branch is proven by stub and mutation only. Proving it against real GitHub means deleting `main`'s protection, which is the hazard this arm exists to detect.

PR #1073, opened 2026-08-30
feat(run-tests): --targets, so a run can cover a SUBSET of the target list — plus the one guard that was blind to partial runs

ZacxDev:
Groundwork for CI impact analysis — running only the suites a change could have broken. **This lands the mechanism only**: nothing selects targets automatically yet and the gate still runs everything.

## Why a mechanism and not a path filter

A subset run asserts that the unselected targets could not have caught the change, and nothing here can check that claim. So its failure mode is a **green run that never executed the suite holding the regression**, and every design choice follows from that:

- an **unknown** target is FATAL (exit 3), never dropped. `--targets scripts/dl-router` — no `/tests` — is a real directory and an obvious thing to type; under a prefix filter it selects nothing and exits 0 having tested nothing, which reads exactly like a pass
- an **empty** selection is FATAL. "Nothing to run" is not a verdict this script may reach
- selection **narrows**, never widens: a devhost target under `--set hermetic` is an unknown-target error, not an implicit `--set all`
- a subset run **says so** on stderr, and the floor line says it summed the *selected* targets. Its verdict is otherwise byte-identical to a full run's

## The floor needed no work

GUARD 3's global floor is already `sum over $TARGETS`, so narrowing the array narrows the floor. Probed: a 2-target run reports `floor: 575 = sum of 2 per-target floors`. The risk I expected to be largest was already closed in the repo.

## The actual bug, which a partial run made reachable

`_skip_entry_applies` honoured an `EXPECTED_SKIPS` entry's *condition* but not whether its target was in the run, so a subset counted pins for suites that never executed:

```
ERROR: 1 test(s) skipped, but 2 of 3 pinned entries apply here.
```

A correct run, failed by its own guard. Now filtered on `TARGETS` membership, ordered before the condition arm so an entry from an untouched target cannot set `fail`.

⚠️ That filter is **gated on a subset being active**. An unconditional version broke 8 tests in `test_conditional_skip_pins.py`, which drives this predicate with synthetic ledgers naming dirs that are not real targets — the only way to exercise the condition grammar. Filtering unconditionally silently disabled those grammar checks. It looked like a strict improvement.

## Two traps the full suite caught that nothing faster did

Both are now comments at the site:

- that harness extracts blocks from `run-tests.sh` by **unique text anchors** and asserts each occurs once. A comment I added quoted an anchor verbatim — **11 tests red, from prose**. All six anchors re-verified unique
- it also extracts `_skip_entry_applies` and runs it under `bash -u`, where neither `ONLY_TARGETS` nor `TARGETS` exists. A bare `$ONLY_TARGETS` aborts with exit 127 — hence `${ONLY_TARGETS:-}`

## Tests

6 new, every one watched to fail on a mutant: membership decision removed · membership always-true · announcement dropped · floor label lying · `${ONLY_TARGETS:-}` reverted (that one takes 11 tests red, so it is covered rather than merely asserted).

🔴 **Two of them were vacuous on the first mutation run**, which is worth naming: one asserted the `for _t in "${TARGETS[@]}"` loop *existed*, and deleting the `if` that used its result left it green — the ingredient, not the decision. The other asserted `"SUBSET"` appeared in the output, which the floor line's own "a SUBSET of hermetic" already satisfied. Both are now behavioural.

## Verification

- **dev-host tier**: `RESULT: PASS` — 18,834 collected, 18,832 passed, 2 skipped, 0 failed
- an earlier run was red on `test_skill_tiers.py`, which is **not from this branch**: controlled by running it at clean `origin/main` (passes) and at this branch's base (fails). Fixed by #1069 after the branch was cut; rebased onto it
- **sandbox tier** (`nix build .#checks.x86_64-linux.pytests`) — the tier Tekton gates on — running; I will post the result

## What this does not do

Nothing picks targets automatically. Mapping changed paths to targets is a separate, fallible decision and is deliberately outside this script.

Two measured findings for whoever takes that on:
- the gated tier is **structurally blind to git** (`cp -r ${./.} src` has no `.git`), so impact analysis cannot be computed inside `nix build` — it has to be computed in the clone step and written into the tree
- a mapping alone is worth **~1.7×** on the last 25 merged PRs, because every subsystem is referenced from inside `scripts/tests`. It is worth **~3.6×** only if `scripts/tests` is first decomposed into its repo-wide scanners and its subsystem-specific tests

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -176,6 +176,8 @@ trap 'exit 130' INT
 
 SET="hermetic"
 ROOT=""
+# Empty = run the whole selected SET. Populated only by --targets/DEVRC_TARGETS.
+ONLY_TARGETS="${DEVRC_TARGETS:-}"
 CHECK_TARGETS_ONLY=0
 CHECK_FLOORS_ONLY=0
 while [ $# -gt 0 ]; do
@@ -189,6 +191,11 @@ while [ $# -gt 0 ]; do
     # Same idea for GUARD 3's floor table: validate the two-way pin between
     # TARGET_FLOORS and the target list, print the table, exit. No pytest.
     --check-floors) CHECK_FLOORS_ONLY=1; shift ;;
+    # Run a SUBSET of the target list. Space-separated, and every entry must
+    # already be a declared target — see the ONLY_TARGETS block below for why
+    # this cannot be a free-form path filter.
+    --targets) ONLY_TARGETS="${2:-}"; shift; [ $# -gt 0 ] && shift ;;
+    --targets=*) ONLY_TARGETS="${1#*=}"; shift ;;
     *) ROOT="$1"; shift ;;
   esac
 done
@@ -747,6 +754,60 @@ DEVHOST_TARGETS=(
 TARGETS=("${HERMETIC_TARGETS[@]}")
 if [ "$SET" = "all" ]; then
   TARGETS+=("${DEVHOST_TARGETS[@]}")
+fi
+
+# --- TARGET SUBSET (--targets / DEVRC_TARGETS) --------------------------------
+# 🔴 A SUBSET IS A CLAIM THAT THE UNSELECTED TARGETS COULD NOT HAVE CAUGHT THE
+# CHANGE, and nothing here can check that claim. So this mechanism is built to
+# make the DANGEROUS mistakes loud and the safe ones cheap:
+#
+#   * An UNKNOWN target is FATAL, never silently dropped. A typo'd path that
+#     quietly selected nothing would exit 0 having run nothing, which is the
+#     single worst outcome a test runner has — it reads exactly like a pass.
+#     (`run-tests.sh --targets scripts/dl-router` — no `/tests` — is the shape
+#     that would otherwise do it.)
+#   * An EMPTY selection is FATAL for the same reason. "Nothing to run" is
+#     never a verdict this script is allowed to reach.
+#   * Selection NARROWS the SET, it does not widen it: asking for a devhost
+#     target under `--set hermetic` is an unknown-target error, not an implicit
+#     `--set all`.
+#
+# ⚠ The FLOOR follows automatically and that is not luck — GUARD 3's global
+# floor is already computed as the SUM OVER `$TARGETS` (see MIN_TESTS_COMPUTED),
+# so narrowing this array narrows the floor with it. Verified by probe: a
+# 2-target run reported `floor: 575 = sum of 2 per-target floors`.
+#
+# 📖 What this does NOT do: pick the targets for you. Mapping changed paths to
+# targets is a separate, fallible decision and lives outside this script.
+if [ -n "$ONLY_TARGETS" ]; then
+  _requested=()
+  # shellcheck disable=SC2206
+  _requested=($ONLY_TARGETS)
+  if [ "${#_requested[@]}" -eq 0 ]; then
+    echo "run-tests: FATAL — --targets was given but resolved to NOTHING." >&2
+    echo "           A run that selects no targets would exit 0 having tested" >&2
+    echo "           nothing. Omit --targets to run the whole '$SET' set." >&2
+    exit 3
+  fi
+  _selected=()
+  for _r in "${_requested[@]}"; do
+    _hit=0
+    for _t in "${TARGETS[@]}"; do
+      [ "$_r" = "$_t" ] && { _hit=1; break; }
+    done
+    if [ "$_hit" -eq 0 ]; then
+      echo "run-tests: FATAL — --targets names '$_r', which is not a target in the '$SET' set." >&2
+      echo "           Targets are exact entries of HERMETIC_TARGETS (plus" >&2
+      echo "           DEVHOST_TARGETS under --set all), not path prefixes." >&2
+      echo "           Run with --check-targets to print the declared list." >&2
+      exit 3
+    fi
+    _selected+=("$_r")
+  done
+  TARGETS=("${_selected[@]}")
+  echo "run-tests: SUBSET — ${#TARGETS[@]} of the '$SET' set selected via --targets." >&2
+  echo "           Unselected targets did NOT run; this verdict is about the" >&2
+  echo "           selected ones only." >&2
 fi
 
 # --- GUARD 3's floor table: PER-TARGET, and NOT an exact total -----------------
@@ -1971,7 +2032,15 @@ if [ "$CHECK_FLOORS_ONLY" -eq 1 ]; then
     fi
   done
   echo "  ----"
-  echo "  GLOBAL floor (sum over the $SET set) = $MIN_TESTS_COMPUTED"
+  # 🔴 SAY WHICH SET IT SUMMED. With `--targets` this is the sum over the
+  # SELECTED targets, not over `$SET` — printing "the hermetic set" there is a
+  # false claim about coverage, and the number it labels is exactly the one a
+  # reader uses to decide whether the run was complete.
+  if [ -n "$ONLY_TARGETS" ]; then
+    echo "  GLOBAL floor (sum over the ${#TARGETS[@]} SELECTED target(s), a SUBSET of $SET) = $MIN_TESTS_COMPUTED"
+  else
+    echo "  GLOBAL floor (sum over the $SET set) = $MIN_TESTS_COMPUTED"
+  fi
   exit 0
 fi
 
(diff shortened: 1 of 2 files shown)


ZacxDev:
**Sandbox tier: PASS.**

`nix build .#checks.x86_64-linux.pytests` — the tier Tekton gates on, built from a `cp -r ${./.}` store copy with no `.git`:

```
RESULT: all good
TOTAL collected=18834  passed=18832  skipped=2  failed=0  (floor: 16024 = sum of 28 per-target floors)
RESULT: PASS (exit=0)
```

Both tiers now measured, at this commit, on this branch:

| tier | command | result |
|---|---|---|
| dev-host | `bash scripts/run-tests.sh` | PASS — 18,832 passed, 0 failed |
| sandbox (gated) | `nix build .#checks.x86_64-linux.pytests` | PASS — 18,832 passed, 0 failed |

Naming both because they are structurally different environments and this change touches `run-tests.sh` itself — GUARD 10's `NOGIT_REPO_LOCAL` class evaluates differently in the sandbox, so a green on the dev host alone would not have said anything about it.

⚠️ Both are claims about **this branch**, not about the tree the merge creates. `strict` is false on this repo's branch protection, so a green check here is a property of the branch; gating the merged tree is still a manual step.

ZacxDev:
## Round 1 fixes landed — `c62ffa07`

Ten findings from the round-1 adversarial audit. Both tiers re-run on the fix tree:

| tier | command | result |
|---|---|---|
| dev-host | `bash scripts/run-tests.sh` | PASS — 18843 collected / 18841 passed / 2 skipped / 0 failed |
| sandbox (gated) | `nix build .#checks.x86_64-linux.pytests` | PASS — same counts |

9 tests added (23 in the file). 12 mutants killed across three batteries, including the two that had survived a sweep before.

**Two findings NOT fixed, deliberately** — both latent, neither reachable by any current `EXPECTED_SKIPS` entry:

- **F4** — the skip filter is gated on `ONLY_TARGETS` rather than on "did this target run", so `--set hermetic` is also a partial run of the declared target space.
- **F6** — a subset run stops validating ledger *grammar* for unselected targets. The obvious fix re-opens the defect that gating solved (it broke 8 tests in `test_conditional_skip_pins.py` by silently disabling the grammar checks those synthetic ledgers exist to exercise).

F5's new pin (`test_every_expected_skip_dir_is_a_declared_target`) is what catches F4 landing for real.

```audit-claims round=1 audited=bf4e904f..c62ffa07
1. F1: a subset run now declares itself inside the SUMMARY region that gate.sh prints, in the banner and in a PARTIAL RUN block below it, instead of only on stderr above it.
2. F1b: the RESULT: line was left byte-identical to a full run's, after a first draft appended to it.
3. F2: a duplicated target is now FATAL (exit 3) rather than running one suite twice and doubling its floor contribution.
4. F3: the --check-floors row label now says "not SELECTED by --targets" under a subset instead of "not in the $SET set".
5. F5: the claim that every EXPECTED_SKIPS dir is a declared target is now pinned by a test rather than asserted in a comment.
6. F7: passing --targets more than once is now FATAL rather than last-wins.
7. F8: the --targets value is no longer glob-expanded against the filesystem; word splitting is preserved.
8. F9a: DEVRC_TARGETS is now documented in the Usage and Env-overrides blocks.
9. F9b: a subset run now names its SOURCE (the flag vs the environment variable) in the announcement.
10. F9c: the meta-test harness now scrubs DEVRC_TARGETS from the environment it passes to the runner.
11. G1: an empty --targets value ("" or --targets=) is now FATAL rather than falling through to a full run.
12. F10: the tests now assert the identity of the selected target set, not only that the floor moved and that the run announced itself.
13. F4 and F6 were NOT fixed and are stated above as latent.
```

ZacxDev:
## Round 2 fixes landed — `0db4ab01`

All six 🟡 and both 🟢 from the round-2 delta audit. Three were regressions the round-1 fixes introduced.

| tier | command | result |
|---|---|---|
| dev-host | `bash scripts/run-tests.sh` | PASS — 18847 / 18845 / 2 skipped / 0 failed |
| sandbox (gated) | `nix build .#checks.x86_64-linux.pytests` | PASS — same counts |

4 tests added (27 in the file). 8 mutants killed this round.

🔴 **Two of my own mutants survived the first battery, and both were test defects rather than code defects** — worth stating because it is the second round running where a mutation sweep of mine was the thing that was wrong:

- `${SET_TOTAL}` → `${#TARGETS[@]}` applied to the banner only: my test read the *stderr* line, so the banner's denominator was genuinely unpinned — which was the audit's actual finding. Both sites are asserted now, and the stderr line gained a denominator it never had.
- the empty-env test asserted `"DEVRC_TARGETS" in proc.stderr`, satisfied independently by the remedy line, so reverting the FATAL headline stayed green. It now pins the FATAL line and asserts it does **not** name `--targets`.

**Correction to the round-1 comment on this PR:** it claimed F5's new pin "is what catches F4 landing for real". The audit measured that as **false** — the pin was built against `HERMETIC ∪ DEVHOST`, and a `scripts/devhost-tests` ledger entry passes it while reddening a default `--set hermetic` run. Now pinned against `HERMETIC_TARGETS` alone.

**F4 and F6 remain unfixed and latent**, unchanged from round 1: no current `EXPECTED_SKIPS` entry reaches either, and the obvious F6 fix re-opens the defect that gating solved.

```audit-claims round=2 audited=c62ffa07..0db4ab01
1. F2: a single --targets alongside an ambient DEVRC_TARGETS is no longer fatal; the flag overrides the environment and announces on stderr that it did.
2. F2b: only a REPEATED --targets flag is fatal now, tracked separately from the environment variable.
3. F2c: the test that bypassed _run and so inherited an ambient DEVRC_TARGETS now goes through _run.
4. F3: all four FATAL paths name ONLY_TARGETS_SOURCE instead of hardcoding --targets, and the empty case prints `unset DEVRC_TARGETS` when the environment is the source.
5. F4: --check-targets now declares a subset and states that GUARD 5 did not validate the unselected targets.
6. F1-comment: the two comments saying SUBSET_NOTE is carried into the RESULT line were corrected to say it must not be.
7. F5-pin: the EXPECTED_SKIPS pin is now against HERMETIC_TARGETS alone rather than the union with DEVHOST_TARGETS.
8. G7: the floor-table row label names the actual selection source rather than hardcoding --targets.
9. G8: the env-scrub guard now also asserts the scrubbed run exited 0, so a scrub-to-empty mutant no longer passes.
10. New: the stderr SUBSET announcement now carries a denominator (N of M) it previously lacked.
11. New: one test asserts every early-exit surface (--check-targets, --check-floors) declares a partial run.
12. New: the banner's N-of-M denominator is asserted in addition to the stderr line's.
13. F4 and F6 remain deliberately unfixed and latent, unchanged from round 1.
```

ZacxDev:
## Round 3 fixes landed — `a0ce36c4`

All four findings from the round-3 delta audit.

| tier | command | result |
|---|---|---|
| dev-host | `bash scripts/run-tests.sh` | PASS — 18850 / 18848 / 2 skipped / 0 failed |
| sandbox pytests (gated) | `nix build .#checks…pytests` | PASS — same counts |
| sandbox nodetests (gated) | `nix build .#checks…nodetests` | rc 0 |

Built **one at a time** — a combined `nix build` of both checks contends on the store and produces false failures.

3 tests added (30 in the file). 8 mutants killed, including the remedy-branch deletion that survived round 2's own test for that remedy.

**The regression this round found was mine, again in the messages** — third consecutive round. The empty-selection remedy read only the flag, so with an ambient `DEVRC_TARGETS` it advised "Omit --targets to run the whole set", and obeying that runs `SUBSET: 1 of 28` at exit 0.

**The more useful finding was about guards.** Five of round 2's own message fixes had no guard able to see them regress — including G7, a round-2 finding, fixed and immediately left unpinned. Fixed with one test asserting the *class* rather than five per-message tests, because the failure mode is a sixth message added unguarded.

🔴 **Two of my new tests were wrong again** (third round running): I drove the FATAL paths with a *flag*, where a hardcoded `--targets` is accidentally correct and the mutant is structurally invisible; and nothing asserted the GUARD 5 sentence. Both closed.

### Ledger — payload is converging

| round | payload lines in `run-tests.sh` | scaffolding |
|---|---|---|
| 1 | 173 | 253 |
| 2 | 73 | 161 |
| 3 | **16** | 140 |

Non-zero, so the attribution gate does not fire — but the trend is the shape that rule warns about. If round 4 finds only scaffolding, that is the signal the ladder has left the PR.

```audit-claims round=3 audited=0db4ab01..a0ce36c4
1. The empty-selection remedy now distinguishes three states (flag only / env only / both) and names both knobs when both are set.
2. One test now asserts the CLASS property: with DEVRC_TARGETS set and no flag, no operator-facing line may blame --targets.
3. The unknown-target and duplicate FATAL paths are now driven FROM THE ENVIRONMENT in tests, so a hardcoded --targets is observable.
4. The GUARD 5 "did NOT validate the unselected ones" sentence is now explicitly asserted.
5. The --check-floors early-exit assertion now reads the GLOBAL floor line specifically, not any occurrence of SUBSET or SELECTED in stdout.
6. F4 and F6 remain deliberately unfixed and latent, unchanged since round 1.
```

ZacxDev:
## Ladder closed at round 4 — `a79b7a97`

**Round 4 found no payload defect.** All three of its findings were about scaffolding the ladder itself wrote, two about tests added that same round. Stopping on the attribution gate, not on a clean round.

| tier | command | result |
|---|---|---|
| dev-host | `bash scripts/run-tests.sh` | PASS — 18851 / 18849 / 2 skipped / 0 failed |
| sandbox pytests (gated) | `nix build .#checks…pytests` | PASS — same counts |
| sandbox nodetests (gated) | `nix build .#checks…nodetests` | rc 0 |

Built one at a time — a combined `nix build` of both checks contends on the store and produces false failures.

### What round 4 changed

One fix, because it had a **shipped** consequence rather than a scaffolding one. `test_no_message_hardcodes_the_flag_when_the_ENVIRONMENT_selected` claims in its docstring to cover "every operator-facing string that names the selection source", but runs only under `--check-floors` / `--check-targets` — both of which exit **before** the SUMMARY banner renders. Mutating the banner to a hardcoded `--targets` printed, for an env-selected run with no flag:

```
==== SUMMARY (hermetic set) — SUBSET: 1 of 28 hermetic target(s) via --targets ====
```

and survived all 30 tests **and** the gating sandbox tier with byte-identical counts. That is the same class as F1, the deploy-blocking finding this PR opened with, on the same surface. Now pinned; both it and the PARTIAL-RUN-block mutant die.

The other two findings are filed as **#1094** with a closing condition.

### The ladder, measured

| round | payload lines in `run-tests.sh` | scaffolding | payload defects found |
|---|---|---|---|
| 1 | 173 | 253 | 1 🔴, 9 🟡/🟢 |
| 2 | 73 | 161 | 6 🟡, 2 🟢 |
| 3 | 16 | 140 | 1 🟡 + guard thinness |
| 4 | 0 | 34 | **none** |

Rounds 1–3 each found that the *previous fix round* had broken something, every time in the operator-facing messages. Round 4 broke the pattern by finding nothing in the payload at all.

### The honest tally

- **1 deploy-blocking finding**: a subset run was invisible on the one surface `gate.sh` reads, and the banner asserted the full set.
- **3 regressions introduced by my own fix rounds**, each caught by the next round.
- **3 consecutive rounds where my mutation sweep was itself wrong** — I mutated the banner while testing stderr, asserted a string the remedy line satisfied independently, and drove the FATAL paths with a flag where a hardcoded `--targets` is accidentally correct. Each time the code fix was trivial and the *test meant to prove it* was the defective part.
- **1 test I added believing it closed a gap kills zero mutants** the file did not already kill (`test_check_floors_declares_the_subset_on_its_GLOBAL_line`); `test_a_subset_run_SAYS_it_ran_a_subset` had pinned that string since round 2.

### Still deliberately unfixed

**F4** — the skip filter is gated on `ONLY_TARGETS` rather than "did this target run", so `--set hermetic` is also a partial run of the declared space. **F6** — a subset stops validating ledger *grammar* for unselected targets. Neither is reachable by any current `EXPECTED_SKIPS` entry, and the obvious F6 fix re-opens the defect that gating solved. `test_every_expected_skip_dir_is_a_declared_target` is what catches F4 becoming reachable.

PR #1083, opened 2026-08-30
feat(session-id): make the resumable session id deterministic on every agent commit

ZacxDev:
Rank 5 of `claudedocs/handoff-cairn-task-linkage.md` / **cg#365**.

> **This PR activates nothing.** Merging it changes how zero commits are made. Both halves are opt-in operator acts — see *Activation* below.

## The problem, and the half that wasn't known

cg#365's closing condition is "git blame a line → resolve the session → wake it."

**Known:** the id spaces are disjoint. The commit trailer carries a claude.ai token; the handle that actually resumes a session is the runtime's own id. Measured — **69 of 69** per-session hook state dirs are uuid-shaped, **zero** are `session_` tokens. That's why searching a transcript for its own claude.ai token returns 0: different namespaces, not a broken search.

**Not known, and the bigger hole — coverage:**

| window | commits carrying a `Claude-Session:` trailer |
|---|---|
| last 100 | **46%** |
| last 200 | **33%** |

So half of `main` can't be resolved to a session *regardless* of which id is stamped. It's 46% because the trailer is emitted by **prose** — an instruction the agent must remember. `RULES.md` says to prefer a structural fix; 46% is what the prose version scores.

## Design

A `PreToolUse` hook records the resumable id; `prepare-commit-msg` stamps it.

**State is keyed by Claude-ancestor PID, not one file per repo.** ~117 worktrees share one common git dir here and sessions run concurrently, so a single-file design would let one session's commit be stamped with **another session's id** — worse than no stamp, since it sends a future reader to a session that never touched the line. Both hooks resolve the same pid independently from their own ancestry, so no lock is needed.

**The claude.ai URL is deliberately not synthesised.** The hook layer never receives that token, so a URL built from the id wouldn't resolve. The hook guarantees the *resumable* half; the URL stays best-effort prose.

## The seam test earned its keep

**All 32 unit tests passed while the feature did nothing.** The hook derived both the git dir *and* the library path from its installed location — so installed into any repo other than devrc it looked for `scripts/lib/` inside **that** repo and silently stamped nothing.

Those are two different questions: where the **state** lives (installed path, `abspath`) vs where the **code** lives (symlink target, `realpath`). They coincide only for devrc itself. Watched red → fixed → watched green.

## Mutation results

Each mutant died on the **intended** assertion, module restored byte-identical after each:

| mutation | killed by |
|---|---|
| `lookup` ignores the pid | the concurrent-session safety test |
| drop the idempotence check | the repeated-`--amend` test |
| reintroduce a uuid shape check | the opaque-id test — on the `ses_` case specifically, while the uuid case still **passed**, so the fixtures discriminate |
| drop the newline guard | the message-corruption test |

The uuid-shape mutant matters: `cairn_who.py` records 2 of 41 windows carrying a `ses_…` token and warns that a shape-assuming join "silently matches nothing."

## Activation — both opt-in

- **Claude side:** the `PreToolUse` registration lives in per-host `settings.json`, unmanaged by design.
- **Git side:** `scripts/install-session-stamp.sh`, **dry-run unless `--apply`**. Same shape as `sync-skill-tiers.py`.

🔴 Deliberately **not** wired through `githooks/install.sh`: that sets `core.hooksPath` globally and arms the devrc pre-push test gate — the **#322** shape, where a pre-push hook rewrote the branch it was pushing. The installer also *reports* when `core.hooksPath` is set, since that would make an install into `.git/hooks` silently inert.

Fails open throughout: a broken deploy costs a trailer, never a commit. A human's `git commit` has no Claude ancestor and comes out byte-identical.

https://claude.ai/code/session_01TY12oTR7U9BSzaEe4KoucA


ZacxDev:
## Round 1 fixes — what was claimed, for the round-2 delta audit

Round 1 (`9164486e`) returned **needs rework**. Fixes landed in `29c5556f`, `3e04b203`, `fbd71d71`.

```audit-claims round=1 audited=9164486e..fbd71d71
1. 🔴-1 the hook could REFUSE a commit when its interpreter was unresolvable — claimed fixed by replacing the `#!/usr/bin/env python3` hook with a `/bin/sh` wrapper plus a Python half, every exit 0
2. 🔴-2 four ledger/scan failures (mention-only, shebangs, managed-hooks, state-ownership) — claimed registered; two further ledgers (PATH-clobbers, and the shebang guard's own self-check) were hit by those registrations and claimed registered too
3. 🟡-1 a failed message write truncated the commit message to 0 bytes and lost the commit — claimed fixed by writing to a temp path and renaming
4. 🟡-2 `git -C <other-repo> commit` recorded state against the cwd's repo — claimed fixed by removing the git dir from the design entirely; state now lives under $HOME keyed only by pid
5. 🟡-3 session B amending session A's commit kept A's id — claimed fixed by keying idempotence on (key, value) and rewriting the trailer in place
6. 🟡-4 `prune()` kept any file whose pid merely existed, and the pid space has already wrapped — claimed fixed by pinning /proc field 22 (starttime) and treating a mismatch as a MISS
7. 🟡-5 the recording half had zero tests — claimed fixed by scripts/tests/test_session_stamp_hook.py (24 tests)
8. 🟡-6 the installer printed INSTALLED and exited 0 on an install it had just declared INERT — claimed fixed by refusing with rc 5
9. 🟡-7 the core.hooksPath check read --local/--global and so missed XDG/--system — claimed fixed by reading the effective value
10. 🟡-8 placing the hook in githooks/ auto-armed it under githooks/install.sh — claimed fixed by moving it out of githooks/ entirely
11. 🟡-9 the installer baked an ephemeral worktree path into a symlink that dangles — claimed fixed by GENERATING the wrapper into .git/hooks instead of shipping and symlinking one
12. 🟡-10 no settings.json registration snippet existed anywhere — claimed fixed by printing it on successful install
13. nit: `has_trailer` used .strip() so an INDENTED example in a message body counted as a real trailer — claimed fixed by anchoring at column 0
14. nit: the PR body's 46%/33% coverage figures did not reproduce — claimed corrected to 41%/27%, line-anchored, with a sha
15. nit: `_COMMIT_RE` backtracked super-linearly — claimed fixed by removing the nested quantifier over an optional group
16. a defect found DURING round 2, not reported by round 1: the recording tests passed on the dev host and failed in the nix sandbox, because the hook keys on a Claude ancestor that a sandbox build does not have — claimed fixed by an injected pid, as the git half already had
```

### Two things the delta round should treat as open, not fixed

- **`git commit -v` / `--cleanup=scissors`** — round 1 reported the trailer lands after the scissors line and git discards it. **Not fixed.** It needs a decision about insertion position relative to the scissors marker, and I did not want to guess at it mid-round.
- **The `command -v python3` guard in the generated wrapper is an EQUIVALENT MUTANT.** Deleting it does not fail the suite, because the trailing `exit 0` already swallows a missing interpreter. Documented in the source as not-load-bearing rather than left reading as verified. The `exit 0` is the load-bearing half and has its own test.

### A retraction carried in the history

`3e04b203`'s installer comment blamed nixpkgs' `patchShebangs` for a sandbox-tier failure, from a five-build bisect. **That was wrong and is retracted in `fbd71d71`.** The bisect was confounded: every copy was a `cp -a` with `.git` removed, so each step varied `.git` presence alongside the file under test. The real cause was the build invocation — `nix build path:<a git WORKTREE>` copies the worktree's `.git` *file* into the sandbox, so `src` looks like a repo with an unreachable gitdir and every git call there dies on discovery. Built from a clone, the tier is green.

### Gate state at `fbd71d71`

- dev-host `gate.sh` in an isolated clone: `GATE: RESULT=PASS exit=0`, 10431 collected, 0 failed
- sandbox `checks.pytests`, built from a **clone**: realised (positive control confirms the command shape can report failure)
- sandbox `checks.nodetests`: realised

ZacxDev:
## Round 2 fixes — what was claimed, for the round-3 delta audit

Round 2 (`9164486e..fbd71d71`) returned **merge after fixing the 🟡 set**. Fixes landed in `f605e7c7`.

```audit-claims round=2 audited=fbd71d71..f605e7c7
1. 🟡-B the installer was a SHELL INJECTION — `--source` interpolated into an unquoted heredoc ran arbitrary shell on every commit — claimed fixed by a quoted heredoc plus `printf %q` for the one interpolated value
2. 🟡-A the wrapper was generated with a truncating `cat >`, so a failed re-install left an executable but syntactically invalid hook — claimed fixed by generating to a temp path and `mv`
3. 🟡-G a bare `grep | sed` under `set -euo pipefail` aborted the installer with an undocumented rc 1, including on the read-only dry run — claimed fixed by capturing with `|| true`
4. 🟡-C the atomic-write guard was VACUOUS (reverting the fix left the suite 86/86 green) — claimed fixed by asserting the message file's INODE changes, with a negative control for the no-stamp path
5. 🟡-D three ledger justifications stated facts the same round's fixes had falsified — claimed corrected in place
6. 🟡-E the coverage figures were wrong — claimed corrected to 47%/33%, measured per-commit at 3b1a0477 by two agreeing methods, with the wrong stated cause retracted
7. 🟡-F `--uninstall` deleted a foreign hook that merely mentioned the marker string — claimed fixed by anchoring the marker and additionally requiring an `impl=` line
8. 🟢-H `splitlines()` corrupted CRLF and \x0b bodies on the rewrite path — claimed fixed by splitting on "\n" only
9. 🟢-I duplicate trailers with the SAME id were not collapsed — claimed fixed by also requiring exactly one existing trailer for the early return
10. 🟢-J prior-revision symlink installs were neither recognised nor removable — claimed fixed by `is_legacy_ours`
11. 🟢-K the equivalent-mutant note was claimed documented but did not exist — claimed fixed by emitting it into the generated wrapper
12. the 0600 state file was chmod'ed AFTER creation, leaving a world-readable window — claimed fixed by `os.open(..., 0o600)`
```

### The injection guard took four attempts, and the first three were green

Recorded because each failure is a distinct named trap, and a reviewer should check the final one actually discriminates:

1. **Textbook fixtures** — `quote"root` breaks quoting but executes nothing.
2. **An early `return`** when the install failed — a vacuous pass.
3. **The canary armed in one phase only** — heredoc expansion fires at *install* time, where `$CANARY` was unset, so `touch` ran with no argument.
4. **The mutation was too broad** — unquoting the whole heredoc also expanded `"$impl"`/`"$py"`/`"$@"`, breaking the installer outright, so the mutant died for the wrong reason and the vulnerability was never reproduced.

The final guard kills a faithful mutant (only the `%q` removed) on all three payload shapes, while the benign `space root` case stays green.

### Still open, deliberately

- **`git commit -v` / `--cleanup=scissors`** — the trailer lands after the scissors line and git discards it. Round 2 confirmed this affects **editor-driven commits only**; `-m` with `commit.verbose=true` stamps correctly, and `-m` is what agents use. Not fixed.
- **The `command -v python3` guard is an equivalent mutant** — deleting it leaves the suite green because the trailing `exit 0` already covers it. Now documented as not-load-bearing in the generated wrapper.
- **`transcript_path`** is recorded in the state file and read by nothing.

### Gate state at `f605e7c7`

- dev-host `gate.sh`, isolated clone: `GATE: RESULT=PASS exit=0`, **TOTAL collected=19025 passed=19023 skipped=2 failed=0**
- sandbox `checks.pytests` and `checks.nodetests`, built from a **clone**, one at a time: both realised

⚠ An earlier claims block said the dev-host gate collected **10431**. That was one *target's* count quoted as the whole gate — a scoped number reported without its scope. The gate total is 19025.

ZacxDev:
## Round 3 fixes — what was claimed, for the round-4 delta audit

Round 3 (`fbd71d71..f605e7c7`) returned **merge after fixing F1, F2 + the two guard items**. Fixes landed in `f2dfbfa3`.

```audit-claims round=3 audited=f605e7c7..f2dfbfa3
1. F1 `is_legacy_ours` matched a symlink shape devrc never installed and deleted third parties' symlinks — claimed fixed by REMOVING the migration entirely, on the evidence that 9164486e symlinked `githooks/…`, that `scripts/git-hooks/prepare-commit-msg` was never added in any ref, and that the revision never reached main
2. F2 the 0600 hardening regressed when a leftover .tmp existed at 0644 — claimed fixed by unlink + O_EXCL so the mode always applies on creation
3. G1 the atomic-wrapper-write test used a fixture of eight COMMENT lines that `sh -n` accepts — claimed fixed by cutting at the first `if` and ASSERTING the fixture is genuinely invalid before asserting repair
4. a NEW defect the audit did not report, surfaced by fixing G1's fixture: with a genuinely broken hook present the installer printed "already installed ✓" and exited 0, because ownership matched while the file was unusable — claimed fixed by an end sentinel plus `is_intact`, with a broken hook of ours self-repairing WITHOUT --force
5. G2 reverting session_trailer.py wholesale left all 137 tests green — claimed fixed; that revert now fails 9 tests (exotic separators, CRLF, duplicate collapse, 0600-with-leftover)
6. F5 a docstring still asserted the retracted "55" — claimed corrected to 67 with its sha
7. F6 the anchoring half of `is_ours` was unguarded because the fixture wrote the marker at column 0 — claimed fixed by quoting it MID-LINE and giving the fixture its own `impl=` line
8. F7 a vacuous early return survived in the injection class — claimed fixed by asserting a refusal is one of the documented codes and left no hook behind
9. F8 the dead `[ -n "$_existing" ] && echo` was a `set -e` landmine — claimed fixed by making it an `if`
10. ~180 duplicated test lines shipped in round 3 and were caught only by an unrelated edit reporting "2 occurrences" — claimed deduplicated
```

### Two of my own round-3 self-reports were WRONG and are corrected here

- The commit message claimed the benign `space root` fixture **stays green** under the `%q`-removal mutant. It does not — all four params go red, because a bare space is hostile too. What actually matters is that all four are green at HEAD.
- It implied the injection guard demonstrates **two** phases. The install-phase canary **never fires** on any payload, even against the vulnerable installer; it is untriggered defence-in-depth, not a demonstrated second phase.

### Deliberately NOT fixed

- **`git commit -v` / `--cleanup=scissors`** — trailer lands after the scissors line and git discards it. Round 2 established this affects **editor-driven commits only**; `-m` with `commit.verbose=true` stamps correctly, and `-m` is what agents use.
- **The `command -v python3` guard is an equivalent mutant** — documented as not-load-bearing in the generated wrapper; the trailing `exit 0` is the load-bearing half and has its own test.
- **`transcript_path`** is recorded and read by nothing.
- **F9** (no `trap`, so a failed wrapper write leaves a stray temp file, and rc 153 / rc 1 are undocumented) — **not fixed**, deliberately: the trade the audit itself called right (the original hook survives) is intact, and I did not want to add an untested `trap` this late in the ladder.
- **F10** (a CRLF message whose trailer is already correct is rewritten once, self-converging) — **not fixed**; strictly better than the behaviour it replaced.

### Gate state at `f2dfbfa3`

- dev-host `gate.sh`, isolated clone: `GATE: RESULT=PASS exit=0`, **TOTAL collected=19037 passed=19035 skipped=2 failed=0**
- sandbox `checks.pytests` and `checks.nodetests`, from a **clone**, one at a time: both realised

ZacxDev:
## Round 4 fixes — what was claimed, for the round-5 delta audit

Round 4 (`f605e7c7..f2dfbfa3`) returned **merge after fixing 🔴-1**. Fixes landed in `8dd98e7b`.

```audit-claims round=4 audited=f2dfbfa3..8dd98e7b
1. 🔴-1 self-repair set FORCE=1 on `is_ours` alone, so a third-party hook carrying the marker and an impl= line was overwritten with no --force — claimed fixed by scoping self-repair to a hook whose `impl=` names THIS checkout
2. 🟡-2 the self-repair behaviour shipped with zero coverage (deleting FORCE=1 left the suite green) — claimed fixed by tests in both directions, with the lookalike case and the positive control
3. 🟡-3 the ownership fixture ROTATED coverage rather than widening it (un-anchor survived before, drop-impl survives after) — claimed fixed by keeping BOTH fixtures so each half has a killing mutant
4. 🟡-5 the repair path fell through into the re-point branch, printing three contradictory lines in one run — claimed fixed by an `elif` plus a `_repairing` guard, with a test
5. 🟡-4 the docstring figure 47/67 was reported unreproducible — claimed REFUTED, not fixed: the disagreement is an instrument bug, reproduced below
6. 🟢-6 my "the revert now fails 9 tests" conflated two different reverts — claimed corrected to 8 (revert to round-2) and 1 (revert to round-3)
```

### 🔴 The contested measurement was an INSTRUMENT bug, and it is reproduced

The audit measured 41/55 across ten variants and could not reproduce 47/67. Both of us ran honest commands; one shape silently discards data:

```
set -o pipefail + `git show … | grep -q`   ->  55 of 200
same loop without pipefail                 ->  67 of 200
```

`grep -q` exits at the first match, `git show` dies of SIGPIPE, and `pipefail` promotes that to a failed pipeline — so a commit is dropped **exactly when it matches**. 47/67 re-confirmed by a third method (one subprocess per commit, no pipe anywhere).

This is the same trap behind the 41%/27% figure I adopted from the round-1 auditor and shipped into a docstring. It has now produced a wrong number in **three** independent measurements, so it is documented in `session_trailer.py` rather than left to catch a fourth.

**Please measure it a fourth time, without a pipe, and say your method.** If you get 55 again with a pipe-free method, I am wrong and want to know.

### Mutation results this round

| mutation | killed by |
|---|---|
| un-scope self-repair (the round-4 regression) | `test_a_lookalike_third_party_hook_is_not_overwritten` |
| delete `FORCE=1` | the self-repair positive control **and** the messaging test |
| un-anchor the marker grep | both anchor tests |
| drop the `^impl=` requirement | `test_marker_at_column_zero_but_NO_impl_line_is_not_ours` |

### Still open, deliberately

- **`git commit -v` / `--cleanup=scissors`** — editor-driven commits only; `-m` (what agents use) stamps correctly.
- **The `command -v python3` guard is an equivalent mutant** — documented as not-load-bearing; the trailing `exit 0` is the load-bearing half and has its own test.
- **`transcript_path`** is recorded and read by nothing.
- **F9** — no `trap`, so a failed wrapper write leaves a stray temp file; rc 153 / rc 1 undocumented.
- **F10** — a CRLF message whose trailer is already correct is rewritten once, self-converging.
- **The same-pid `record()` race** the round-4 audit analysed but did not reproduce.

### Gate state at `8dd98e7b`

- dev-host `gate.sh`: `GATE: RESULT=PASS exit=0`, **TOTAL collected=19042 passed=19040 skipped=2 failed=0**
- sandbox `checks.pytests` and `checks.nodetests`, from a **clone**, one at a time: both realised

ZacxDev:
## Round 5 fixes — what was claimed, for the round-6 delta audit

Round 5 (`f2dfbfa3..8dd98e7b`) returned **safe to merge, with 🟡-1 worth closing first**. Fixes landed in `dcc11da7`.

```audit-claims round=5 audited=8dd98e7b..dcc11da7
1. 🟡-1 self-repair still destroyed a VALID hook (derived from ours, naming this checkout, no sentinel, parses fine) with no --force — claimed fixed by requiring `sh -n "$TARGET"` to FAIL before the --force gate is skipped
2. 🟡-1's guard — claimed added: a fixture that asserts it is VALID shell before asserting it survives, plus a positive control that a genuinely broken hook is still repaired; the drop-`sh -n` mutant dies on the intended test
3. 🟡-2 a test docstring still claimed the mid-line fixture exercises "BOTH halves" of the ownership test, which the same commit had measured false — claimed corrected, and it now names the class that provides the other half and must not be deleted as redundant
4. 🟢-3 a comment claimed the messaging fix was an `elif`; the mechanism is a `_repairing` flag — claimed corrected
5. 🟢-4 nix/home.nix carried a FOURTH figure (46%) at a sha where it was never right — claimed corrected to 47%, agreeing with the module, the hook and the tests
6. two overstatements in my own docstring — claimed softened: the SIGPIPE undercount is a RACE (~12 of 67 matches drop), not a rule, and 55 is not a constant (five runs: 55,55,55,56,55). The `grep -c` control (67) is now recorded because it isolates the early exit from the pipe itself
```

### The shape worth auditing, not just the fix

Three consecutive rounds have re-opened the same class one axis narrower:

| round | self-repair condition | outcome |
|---|---|---|
| 4 | any hook matching `is_ours` | clobbered third-party hooks |
| 5 | + must name **this** checkout | still clobbered a derived hook |
| 6 | + must actually **fail to parse** | premise tested directly |

**Please assume round 6 did it a fourth time and look for it.** The specific worry: `sh -n` is now the discriminator, and it is a claim about *syntax*, not about whether the hook works. A hook that parses but is semantically broken (empty, or `exit 1` at the top) still refuses or breaks commits and will now NOT be repaired — is that the right trade, or the next residue?

### Still open, deliberately

- **`git commit -v` / `--cleanup=scissors`** — editor-driven commits only; `-m` (what agents use) stamps correctly.
- **The `command -v python3` guard is an equivalent mutant** — documented as not-load-bearing.
- **`transcript_path`** is recorded and read by nothing.
- **F9** — no `trap`; a failed wrapper write leaves a stray temp file, rc 153 / rc 1 undocumented.
- **F10** — a CRLF message whose trailer is already correct is rewritten once, self-converging.
- **The same-pid `record()` race**, analysed but never reproduced.
- **`test_impl_line_but_marker_only_MID_LINE_is_not_ours` is redundant** with the older fixture (both die under the same mutant) — kept, not wrong.
- **rc 4 prints "already exists and is not ours" right after "this is devrc's own hook, not a foreign one"** — pre-existing, untouched.

### Gate state at `dcc11da7`

- dev-host `gate.sh`: `GATE: RESULT=PASS exit=0`, **TOTAL collected=19044 passed=19042 skipped=2 failed=0**
- sandbox `checks.pytests` and `checks.nodetests`, from a **clone**, one at a time: both realised

ZacxDev:
## Round 6 fixes — what was claimed, for the round-7 delta audit

Round 6 (`8dd98e7b..dcc11da7`) returned **safe to merge; 🟡-1 and 🟡-2 worth a follow-up**. Fixes landed in `6744d641`.

```audit-claims round=6 audited=dcc11da7..6744d641
1. 🟡-1 a truncated devrc hook that happens to PARSE was refused, with two false sentences ("points at another checkout" naming this checkout, "is not ours" about our own hook) — claimed fixed by replacing the `sh -n` proxy with a byte-exact PREFIX comparison against the wrapper this run would generate
2. 🟡-2 `sh -n` asks whatever `sh` is on PATH, not the shell git runs the hook under, and an absent `sh` (rc 127) read as "broken" and destroyed a working hook with no backup — claimed fixed by the same replacement: the prefix test invokes no shell at all
3. 🟢-3 the fixture built `impl=` via `bash -c f'printf "%q" "{IMPL}"'`, itself vulnerable to the metacharacters `%q` exists for — claimed OBSOLETE rather than fixed: that fixture belonged to the `sh -n` guard, which is gone
4. F9's stray-temp-file half — claimed fixed: the wrapper is generated ONCE before the ownership checks, and a trap removes the temp on every exit path
```

### Two defects my own testing found that the audit did not — please verify both

- **Sweeping all 26 cut points instead of sampling one** showed the fix was still wrong for 15 of them: the prefix check sat INSIDE the `is_ours` gate, and `is_ours` requires an `^impl=` line — line 16 of the 27 written. Measured: cuts 1–15 refused (rc 4), cuts 16+ repaired. Hoisted above `is_ours`.
- **The mutation battery caught an unguarded guard of mine.** I added a marker requirement so a bare `#!/bin/sh` (someone's deliberate no-op) is not silently replaced with a working stamper — then excluded that case from the sweep *in a comment* and never tested it. Deleting the requirement left 36 tests green. Now guarded.

That is the fifth vacuous guard in this ladder, and the second I caught myself. **Every one has had the same signature: prose explaining why a case is excluded, standing in for a test of that case.** If you find a sixth, it will probably look like that too — please look for comments in this diff that justify an exclusion, and check each has a test.

### Mutation results this round

| mutation | killed by |
|---|---|
| revert to the `sh -n` proxy | the 26-point sweep |
| prefix check back inside `is_ours` | the 26-point sweep |
| drop the marker requirement | the bare-stub guard |

### Still open, deliberately

- **`git commit -v` / `--cleanup=scissors`** — editor-driven commits only.
- **The `command -v python3` guard is an equivalent mutant** — documented as not-load-bearing.
- **`transcript_path`** is recorded and read by nothing.
- **F9's rc half** — rc 153 / rc 1 on a failed wrapper write are still undocumented codes.
- **F10** — a CRLF message whose trailer is already correct is rewritten once, self-converging.
- **The same-pid `record()` race**, analysed but never reproduced.
- Redundant-but-harmless test pairs noted by rounds 5 and 6.

### Gate state at `6744d641`

- dev-host `gate.sh`: `GATE: RESULT=PASS exit=0`, **TOTAL collected=19047 passed=19045 skipped=2 failed=0**
- sandbox `checks.pytests` and `checks.nodetests`, from a **clone**, one at a time: both realised

ZacxDev:
## Round 7 fixes — what was claimed, for the round-8 delta audit

Round 7 (`dcc11da7..6744d641`) returned **safe to merge; not clean**. Fixes landed in `38039a87`.

```audit-claims round=7 audited=6744d641..38039a87
1. 🟡-1 hoisting generation above the mode branches made read-only paths WRITE — the dry run created .git/hooks, and both the dry run and --uninstall aborted rc 1 on an unwritable hooks dir — claimed fixed by moving the scratch file to $TMPDIR and generating only past the branches that need it
2. 🟡-2 the cleanup trap was an UNGUARDED GUARD (deleting it left 37 tests green while leaking on four paths) — claimed fixed by a parametrized sweep over all five paths asserting no scratch file survives
3. 🟢-3 28 lines of prose describing the deleted is_ours + sentinel + `sh -n` triple survived, asserting the opposite of current behaviour — claimed removed; three test docstrings that named `sh -n` as the live discriminator claimed corrected
4. 🟢-4 a prefix shorter than the `impl=` line is checkout-INDEPENDENT, so announcing it as "this checkout's own hook" was false — claimed fixed by narrowing the CLAIM rather than the behaviour, and naming the checkout it will be completed against
5. 🟢-5 my claim that the `printf %q` fixture was obsolete was WRONG — it is live and still injectable — claimed fixed by passing IMPL as an argument
6. the tautological `assert repaired == len(lines) - 2` — claimed kept as a size bound, annotated to say the per-iteration assertion is what does the work
```

### The signature to hunt, sixth instance

🟡-2 was **this ladder's signature defect landing on a fix for one of its own findings**: round 7's message claimed the trap "closes the stray-file half of F9", and the closure was untested. Six instances now, all the same shape — *prose asserting a case is handled, standing in for a test of that case.*

**This round adds new prose in exactly that register** (why the scratch file is in `$TMPDIR`, why generation is deferred, why the repair message is narrowed, why the sweep starts at 2). Please walk each one and check it has a mutant-killing test. That has been the highest-yield check in the last three rounds.

### Mutation results this round

| mutation | killed by |
|---|---|
| delete `trap cleanup_tmp EXIT` | 4 of the 5 litter-sweep params |
| generate into `.git/hooks` again | the dry-run and read-only guards |
| generate before the `--uninstall` branch | both read-only guards + the dry-run guard |

### A test-name problem worth generalising

`test_dry_run_changes_nothing` passed through the entire dry-run-writes-to-disk defect, because it asserts only that the hook **file** is absent. A name wider than its assertion reads as coverage and provides a slice. If you see others in this feature, they are worth naming — I have not swept for them.

### Still open, deliberately

- **`git commit -v` / `--cleanup=scissors`** — editor-driven commits only.
- **The `command -v python3` guard is an equivalent mutant** — documented as not-load-bearing.
- **`transcript_path`** is recorded and read by nothing.
- **F9's rc half** — rc 153 on a failed wrapper write is still an undocumented code.
- **F10** — a CRLF message whose trailer is already correct is rewritten once, self-converging.
- **The same-pid `record()` race**, analysed but never reproduced.
- **`[ "${_repairing:-0}" -ne 1 ]` is now always true** (the repair branch short-circuits the `is_ours` block) — dead as a condition, kept as belt-and-braces.

### Gate state at `38039a87`

- dev-host `gate.sh`: `GATE: RESULT=PASS exit=0`, **TOTAL collected=19055 passed=19053 skipped=2 failed=0**
- sandbox `checks.pytests` and `checks.nodetests`, from a **clone**, one at a time: both realised

ZacxDev:
## Round 8 fixes — what was claimed, for the round-9 delta audit

Round 8 (`6744d641..38039a87`) returned **merge after deciding on F1**. Fixes landed in `32b19566`.

```audit-claims round=8 audited=38039a87..32b19566
1. F1 staging in $TMPDIR turned the atomic `rename(2)` into a cross-filesystem copy-then-unlink (truncate in place) — claimed fixed by staging beside the target on the apply path only; read-only paths keep $TMPDIR because they never install
2. F1's guard — claimed added: an INODE assertion (a rename changes it, a truncate does not) plus a structural pin that the apply path stages beside the target, since an inode test alone passes on a same-filesystem $TMPDIR and would regress silently on a tmpfs /tmp host
3. F2 the read-only rc-1 abort was RELOCATED, not removed — mktemp still ran before the uninstall branch, so an unwritable or full /tmp reproduced it — claimed fixed by leaving `$_tmp` empty until a branch that needs it says otherwise; `--uninstall` now creates nothing
4. F2's false comment ("IS MADE AFTER THE UNINSTALL BRANCH") — claimed corrected, and the relocation recorded so it is not re-derived
5. F3 the NOTE block had no test (deleting it survived) — claimed guarded in both directions: fires below the impl= line, silent above it
6. F5 the scratch sweep had no positive control (an installer hardcoding /tmp survived) — claimed fixed by a behavioural discriminator; ⚠ my FIRST attempt was itself vacuous (`assert scratch.exists()`), and that is recorded
7. F6 the installed hook's mode had silently changed 0755 -> 0711 — claimed pinned to 0755 with a test
8. F4 deferring `generate_wrapper` past the uninstall branch was an equivalent mutant presented as a pillar of the fix — claimed resolved: under the new structure the deferral IS load-bearing (the apply path stages elsewhere), and the mode pin is now the equivalent mutant, documented as such rather than disguised
```

### The number worth auditing

**Four of the five mutants round 8's audit ran against my work SURVIVED**, and each corresponded to prose I had written asserting a property. Three are now guarded; one is genuinely equivalent and labelled.

**Please run mutants against every property this round's new prose asserts**, including:
- the inode/rename guarantee
- the "read-only paths never create anything" claim
- the NOTE's two branches
- the $TMPDIR discriminator
- the 0755 pin

If any survives, it is the eighth instance of this ladder's signature and I want it named.

### A distinction I am now trying to hold

An **untested** claim and an **untestable** (equivalent) one need different treatment, and both need labelling. I have started writing "this is an equivalent mutant, measured" in place rather than letting a comment read as a guard. Please check I have not used that label to excuse something that *is* testable.

### Still open, deliberately

- **`git commit -v` / `--cleanup=scissors`** — editor-driven commits only.
- **The `command -v python3` guard** — equivalent mutant, documented.
- **`transcript_path`** — recorded, read by nothing.
- **F10** — a CRLF message whose trailer is already correct is rewritten once, self-converging.
- **The same-pid `record()` race** — analysed, never reproduced.
- **The 0711→0755 multi-user consequence** — the mode change is measured; the second-uid effect is inference.
- Names wider than their assertions found by round 8 and left alone: `test_dry_run_changes_nothing`, `test_an_uninstalled_repo_is_left_byte_identical`, `test_install_is_idempotent`.

### Gate state at `32b19566`

- dev-host `gate.sh`: `GATE: RESULT=PASS exit=0`, **TOTAL collected=19061 passed=19059 skipped=2 failed=0**
- sandbox `checks.pytests` and `checks.nodetests`, from a **clone**, one at a time: both realised

PR #1087, opened 2026-08-30
rules: disjoint files are not merge safety — re-run the merged-tree test when the base moves

ZacxDev:
## The gap

The existing bullet reads:

> **Two changes touching one file: TEST-MERGE them** … a *semantic* conflict survives it (both sides edit different lines of one function/config) … **Read the merged result of any overlapping region.**

Both halves presuppose an **overlap**. A break needs none, and when there is none the rule reads as inapplicable — which is the failure mode this file's own preamble warns about (*"when a rule looks inapplicable on a technicality, widen it"*).

## What happened

`civitai/civitai` **#4491** widened the required inputs of a function (two new fields). **#4493**, authored in parallel, added a **second caller** of that same function in a different component. Neither PR touched a file the other touched.

- Both PRs green.
- Merge textually **CLEAN**; `gh pr view --json mergeable` said `MERGEABLE`.
- `main`'s typecheck went **red** on merge.

## Why it's worth widening the rule rather than leaving it to the reader

The merged-tree discipline **was** followed — and then abandoned one step too early. I merged `main` in, re-ran the full unit tier on the merged tree (1601 files green), and then `main` moved again in roughly **90 seconds**. At that point I substituted a cheaper proxy: a check that the two changesets' file lists were disjoint. They were.

File-disjointness and merged-tree-green are not degrees of the same evidence — they are **different claims**. Only one of them can see a new *caller* of a signature the other side changed. Type-level coupling is invisible to a path comparison.

And the window can't be closed by being quick: there's no interval short enough to make "the base hasn't moved" safe to assume. So the rule's trigger is restated as **"the base MOVED"** — observable — rather than "we touched the same thing", which was the premise that didn't hold.

## Shape of the change

- `claude/RULES.md`: **4 lines** of imperative appended to the existing bullet. No existing scope removed or narrowed.
- `claude/RULES-ARCHIVE.md`: the incident narrative as `disjoint-file-merge-break`, plus its TOC entry — per the eviction playbook `test_rules_size.py` prints.

## Verification

- `test_rules_size.py`: **5 passed**.
- 🔴 **Watched RED first.** The untrimmed version left **777 bytes** of headroom against a required **900**, and the test failed with the reclaim figure. The green is a measurement, not an assumption. (Worth noting: running the file directly as `python3 test_rules_size.py` exits **0 silently** — it has no `__main__` runner, so a bare invocation is a false green. Use pytest.)
- Every `→ archive:` anchor in `RULES.md` still resolves: **44 referenced, 0 unresolved.**
- Archive TOC: no label/link mismatches; the new entry is present in both.


ZacxDev:
## `tekton/devrc-pytests` is red, and the evidence says it is the documented capacity artifact — not this change

Recording the diagnosis rather than merging through it.

**Failing test:** `TestTheActorComesFromTheTOKEN.test_a_FORGED_actor_in_the_body_is_DISCARDED[record0-…]` in `scripts/tests/test_subsystem_store_api.py`. This PR changes **two markdown files** (`claude/RULES.md`, `claude/RULES-ARCHIVE.md`) and no code.

**Controls, both run locally:**

- At `origin/main`, without this change: **11 passed**.
- At this PR's head (`5893870b`, main already merged in): **11 passed**.

**Fleet control** — the same check, right now, on unrelated branches:

| PR | `devrc-pytests` |
|---|---|
| #1100, #1099, #1098, #1097, #1096 | pending |
| #1095 | error |
| #1092, #1090 (older, quieter window) | success |

**The repo documents this exact failure mode** — `test_subsystem_store_api.py:99`:

> measured 2026-08-29, the devrc Tekton gate was failing **~60% of runs REPO-WIDE** (6 of 10 in one window, **on unrelated branches**) with `TimeoutError` out of `socket.py` — a localhost round-trip that lost the scheduler for >15 s while 12 pipelineruns shared the node… **The test logic was never reached, so the gate reported a code failure for a capacity problem.**
>
> 🔴 This is the SYMPTOM fix. The cause is a 10-minute parallel suite competing with a saturated cluster, which belongs to Tekton capacity, not to this file.

I observed **five concurrent `devrc-ci` PipelineRuns** on the node while this ran.

**Not retriggering yet** — with five runs already queued, another push adds a sixth and reproduces the contention. Re-run when the cluster has drained.

**Not merging through it either.** A red gate merged past is how a gate becomes meaningless. Leaving this open with the diagnosis attached so whoever picks it up can re-run rather than re-derive.

PR #1095, opened 2026-08-30
docs(handoff): cairn-phase3 rank 2 CLOSED — the backup shipped, and a git-bundle backup would have been green and empty

ZacxDev:
Closes rank 2 of `claudedocs/handoff-cairn-phase3.md`. The backup for the **served** store copy shipped as homelab-infra **#551** (squash `c4e0f82b`), Flux-reconciled and adopted.

## The finding worth keeping

🔴 **A `git bundle` backup of the served copy would have been green and empty.** devrc's `analyze-service-index/backup.py` archives per-scope bundles; re-pointing it at `/data` was the obvious move and is wrong. Measured on the served copy 2026-08-30:

| path | state |
|---|---|
| `devrc/cairn.md` | **untracked** — in no commit of any scope repo |
| `devrc/subsystem-store-api.md` | API-appended bullet in the working tree; **0** occurrences at `HEAD` |

A bundle carries committed objects only, so it is structurally blind to exactly the bytes that exist nowhere else. Hence a whole-tree tar, its own bucket, its own credential.

## What it does and does not fix

Fixes: the served copy now has an off-node, verified, alerted backup with 90-day retention. The "a backup must land before criterion 9" precondition is **satisfied**.

Does not fix: RPO is up to 24 h; the **seed clobber is unchanged** (an API-appended bullet is still overwritten by the next `seed.sh` until criterion 9); and the read/write allowlist split is still criterion 9's own job. The backup makes a destructive `PUT` *recoverable*, not impossible.

## Two lessons from seven audit rounds

- **A comment inside an unquoted heredoc executed on the host.** Backticked prose (`` `mc admin user add` ``) ran at expansion time and its output was spliced into a script that runs as root inside the MinIO tenant pod. Now guarded with a positive control.
- **A pre-flight deleted the credential it existed to protect** — and its *destructive predecessor had been accidentally safe* against a dangling symlink, because `rm` doesn't follow links. Removing a bad mechanism can remove a good side effect with it.

⚠ The ladder stopped on the **payload-attribution gate, not a clean round** — round 7 still found a real 🟡. By then fixes were touching 3 executable lines against ~24 of prose. Recorded so nobody reads "merged" as "audited clean".

### DIFF
--- a/claudedocs/handoff-cairn-phase3.md
+++ b/claudedocs/handoff-cairn-phase3.md
@@ -183,11 +183,33 @@ fresh: `claim-work --slug-for <this doc> <rank>`.
    jsonpath='{.data.token}' | base64 -d` and take field 1 of line 2. 🔴 **Never pipe it into an
    `ssh … bash -s` that is also receiving a heredoc** — see the MULTIOS trap above; that is
    what forced the first rotation.
-2. 🔴 **The backup CronJob, and it is now more urgent than it was.** The served copy can hold
-   bytes that exist nowhere else the moment anything writes through the API (see the seed
-   clobber below), and `zach` is a whole-file `PUT` credential over all 15 scopes. The LOCAL
-   store is covered (`commit.sh` per-scope git + `backup.py` encrypted bundles), so "the store
-   has no backup" is wrong — **"the SERVED copy has none" is right**, and that is the gap.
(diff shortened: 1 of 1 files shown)


ZacxDev:
Superseded — redoing this through `scripts/lib/handoff_doc.py` (the `/handoff` write path) rather than hand-editing the doc. Same content, but it goes through the merge, the forcing-function gate and the drops-durable-lines check instead of around them. The hand-edit was the mistake the write gate exists to catch.

PR #1108, opened 2026-08-30
docs(tekton): rank 6 REFUTED — the hostPath cache was reverted after 2h20m; what was missing is WHY it can't come back

ZacxDev:
Rank 6 of `claudedocs/handoff-skill-chain-usage-audit.md` (devrc#1055) said the `tekton`
skill's devrc-ci claims were stale — *"measured 2026-08-29 against the live gate: no
`nodeSelector`, and a per-node hostPath at `/var/lib/mnt/disk-1/devrc-ci-nix-cache`"* —
and named lines 39/41/45/48/430 for correction. That item also said **re-measure before
editing, it is a live label/volume, not a git fact.** I did, and the premise is **REFUTED**.

## The skill is correct. Live, 2026-08-30T18:5xZ (homelab, ns `tekton-ci`)

| thing | live value |
|---|---|
| `devrc-ci-vchxk-gate-pod` nodeSelector | `kubernetes.io/hostname: talos-xr6-r7p` |
| its `nix-cache` volume | `persistentVolumeClaim: {claimName: nix-store-cache}` |
| pvc `nix-store-cache` | Bound, 30Gi, selected-node `talos-xr6-r7p` |
| pvc `nix-store-cache-2` | Bound, 30Gi, selected-node `talos-uvh-gtj` (`gitops-validate`) |

**So no line named by rank 6 is edited here.** The 08-29 measurement was a correct read of
a state that existed for 2h20m: homelab-infra `6bec075e` (2026-08-29T22:09Z) swapped the
PVC for the hostPath, `7839ef54` (2026-08-30T00:29Z) reverted it, and the read landed
inside that window.

## What was actually wrong, and it is the part that misleads

Gotcha 6(a) ends on *"The narrow fix is a baseline-compatible cache."* That reads as an
open, cheap direction. It is not — the hostPath has a **second, independent blocker with
nothing to do with PodSecurity**, and the skill never recorded it. `DirectoryOrCreate` is
created fresh by kubelet as **root**, so `seed-nix` fills it and then nothing inside the
sandbox can take the nix DB lock:

```
error: opening lock file "/nix/var/nix/db/big-lock": Permission denied
```

**75 occurrences over 42 tests, on EVERY devrc PR**, byte-identical across two different
branches (`devrc-ci-csfzb`/#1057, `devrc-ci-kdcmr`/#1059), against a pre-change run
`devrc-ci-q4d5m` of `failed=0` over 18,557 passed. The PVC works only because earlier runs
populated its `/nix` with ownership the nix build user can use.

Recorded alongside it, so a retry starts from measurements rather than zero: what the
unpin genuinely bought (three nodes reachable instead of one, queue wait **17–22m →
0.1m**, wall clock **39.1m median → 17.4m**), and that `686d6ff0`'s `privileged` namespace
exemption buys nothing while the PVC is back.

## Diff

+21 / -0 to `claude/skills/tekton/SKILL.md`, all inside gotcha 6(a).

## Verification

- Live reads above, from `kubectl -n tekton-ci get pod/pvc -o json` on the homelab cluster.
- Revert provenance from `git log`/`git show` on `ZacxDev/homelab-infra` `origin/trunk`.
- `scripts/tests/test_doc_path_rot.py` → **77 passed**.

https://claude.ai/code/session_01Cwo6fMLaXLfm7jZ9ouKqAj

### DIFF
--- a/claude/skills/tekton/SKILL.md
+++ b/claude/skills/tekton/SKILL.md
@@ -305,8 +305,16 @@ debugging, changing or copying a specific pipeline.
    the dev host and `failed=43` in CI — identical inputs, different output, i.e. impure.
    🔴 **Before debugging any diff against a red `devrc-ci` run, check both:**
    `kubectl get taskrun <run>-gate -n tekton-ci -o jsonpath='{.status.conditions[*].message}'`
-   and `kubectl exec -n tekton-ci <gate-pod> -c step-pytests -- sh -c 'nix config show | grep "^sandbox "'`.
+   and `kubectl exec -n tekton-ci <gate-pod> -c step-pytests -- sh -c 'ls -d /build; nix config show | grep -E "^(sandbox|sandbox-fallback) "'`.
    A red check whose cause is either of these is a **broken gate, not a bad change**.
+   🔴 **`nix config show` alone CANNOT ANSWER THIS — it reports the CONFIGURED value, and the
+   configured value is `true`.** Measured 2026-08-30 in a live gate pod: `sandbox = true`,
+   `sandbox-fallback = true`, **`/build` does not exist**, and the build sits in
+   `/tmp/nix-build-<drv>-0` — nix fell back to running UNSANDBOXED and said nothing. **`/build`'s
+   absence is the tell**, which is why the command leads with it; a grep for `^sandbox ` returns
+   a reassuring `true` over exactly the condition this gotcha exists to catch. Recorded as a 🔴
+   KNOWN LIMITATION at `devrc-ci-pipeline.yaml:147-154`, where all three fixes are blocked by
+   PodSecurity `baseline:latest`.
 
 ## What / where
 
--- a/claude/skills/tekton/SKILL.md
+++ b/claude/skills/tekton/SKILL.md
@@ -149,8 +149,12 @@ debugging, changing or copying a specific pipeline.
    *delays* it — and whether that becomes a lost verdict is decided by the pipeline's own
    headroom, not by the burst. devrc (50m budget) drains and survives; gitops-validate, with
    the tightest budget and the worst latency, is the one that loses. **The lever is scheduling
-   pressure, not the timeout** (#378 family). Note `clawgate-ci` — the one pipeline with **no
-   `nodeSelector`** — has a pod-start max of **18s** rather than 19 minutes.
+   pressure, not the timeout** (#378 family). Note `clawgate-ci` — **no `nodeSelector`** — has a
+   pod-start max of **18s** rather than 19 minutes. ⚠ It was the ONLY unpinned pipeline when
+   this was measured 2026-08-23 and has not been since **2026-08-26**, when the `vetr-*` family
+   landed; **8** are unpinned today. **Derive the set** —
+   `~/workspace/devrc/claude/skills/tekton/reference/pipelines.md` → the clawgate-ci "PVC
+   unpinned" bullet ships the command; never read a pin list out of prose.
 4. **Placeholder imagePullSecret breaks ALL pulls.** A `harbor-cred` dockerconfigjson with a
    non-base64 `auth` placeholder makes every pod fail image pull ("illegal base64 data").
    **Do NOT attach a placeholder imagePullSecret** to the pipeline SA — public images
--- a/claude/skills/tekton/SKILL.md
+++ b/claude/skills/tekton/SKILL.md
@@ -260,6 +260,27 @@ debugging, changing or copying a specific pipeline.
    `pod-security.kubernetes.io/enforce: privileged` on the namespace (`686d6ff0`).
    ⚠ That is a **broader** grant than the 7 infra namespaces already carrying it — this one
    runs webhook-triggered CI. The narrow fix is a baseline-compatible cache.
+   🔴 **THAT hostPath IS GONE — `6bec075e` was REVERTED by `7839ef54` 2h20m later
+   (2026-08-29T22:09Z → 2026-08-30T00:29Z), so gotcha #3's node-pinning and §"Adding a
+   pipeline" step 4 describe the LIVE gate, not history.** Re-measured 2026-08-30T18:5xZ on
+   `devrc-ci-vchxk-gate-pod`: `nodeSelector kubernetes.io/hostname=talos-xr6-r7p`, volume
+   `nix-cache` → `persistentVolumeClaim: nix-store-cache` (30Gi, Bound, selected-node
+   `talos-xr6-r7p`); `gitops-validate` on `talos-uvh-gtj` with its own `nix-store-cache-2`.
+   A read of this section taken *during* that 2h20m window is a correct measurement of a
+   state that no longer exists — check the volume live before quoting either shape.
+   🔴 **The revert was NOT the admission failure above — it is a SECOND, INDEPENDENT failure
+   of the same volume, and it is the one to solve first if the unpin is ever retried.** A
+   `DirectoryOrCreate` hostPath is created fresh by kubelet as **root**, so `seed-nix` fills
+   it and then nothing inside the sandbox can take the nix DB lock:
+   `error: opening lock file "/nix/var/nix/db/big-lock": Permission denied` — **75
+   occurrences across 42 tests, on EVERY devrc PR**, byte-identical on two different branches
+   (`devrc-ci-csfzb`/#1057, `devrc-ci-kdcmr`/#1059), against a pre-change run
+   (`devrc-ci-q4d5m`) of `failed=0` over 18,557 passed. The PVC works only because earlier
+   runs populated its `/nix` with ownership the nix build user can use. ⚠ **What the unpin
+   genuinely bought, recorded so it is not re-litigated from zero:** three nodes reachable
+   instead of one, queue wait **17–22m → 0.1m**, wall clock **39.1m median → 17.4m**. Retry
+   it only with the store's ownership solved FIRST and proven on a scratch pipeline — and
+   then reconsider `686d6ff0`, whose `privileged` exemption buys nothing while the PVC is back.
    **(b) `sandbox = false` in the CI pod.** The `nix build .#checks…` tier is only hermetic
    where nix's sandbox is ON. With it off, the *same derivation hash* produced `failed=1` on
    the dev host and `failed=43` in CI — identical inputs, different output, i.e. impure.
--- a/claude/skills/tekton/SKILL.md
+++ b/claude/skills/tekton/SKILL.md
@@ -260,27 +260,50 @@ debugging, changing or copying a specific pipeline.
    `pod-security.kubernetes.io/enforce: privileged` on the namespace (`686d6ff0`).
    ⚠ That is a **broader** grant than the 7 infra namespaces already carrying it — this one
    runs webhook-triggered CI. The narrow fix is a baseline-compatible cache.
-   🔴 **THAT hostPath IS GONE — `6bec075e` was REVERTED by `7839ef54` 2h20m later
-   (2026-08-29T22:09Z → 2026-08-30T00:29Z), so gotcha #3's node-pinning and §"Adding a
-   pipeline" step 4 describe the LIVE gate, not history.** Re-measured 2026-08-30T18:5xZ on
-   `devrc-ci-vchxk-gate-pod`: `nodeSelector kubernetes.io/hostname=talos-xr6-r7p`, volume
-   `nix-cache` → `persistentVolumeClaim: nix-store-cache` (30Gi, Bound, selected-node
-   `talos-xr6-r7p`); `gitops-validate` on `talos-uvh-gtj` with its own `nix-store-cache-2`.
-   A read of this section taken *during* that 2h20m window is a correct measurement of a
-   state that no longer exists — check the volume live before quoting either shape.
+   🔴 **THAT hostPath IS GONE — `6bec075e` was REVERTED by `7839ef54` 2h19m later
+   (2026-08-29T22:09Z → 2026-08-30T00:29Z), so gotcha #3's node-pinning and
+   §"Adding a new pipeline / new repo" step 4 describe the LIVE gate, not history.**
+   Re-measured 2026-08-30 on `devrc-ci-vchxk-gate-pod`:
+   `nodeSelector kubernetes.io/hostname=talos-xr6-r7p`, volume `nix-cache` →
+   `persistentVolumeClaim: nix-store-cache` (30Gi, Bound, selected-node `talos-xr6-r7p`);
+   `gitops-validate` on `talos-uvh-gtj` with its own `nix-store-cache-2`. A read taken
+   *during* that window is a correct measurement of a state that no longer exists — check the
+   volume live before quoting either shape. ⚠ **And the window you could have OBSERVED it in
+   is shorter than the git interval: ~1h41m, not 2h19m** — for the first 39 minutes
+   (22:09Z→22:48Z) admission rejected every gate pod, so there was nothing to read.
    🔴 **The revert was NOT the admission failure above — it is a SECOND, INDEPENDENT failure
-   of the same volume, and it is the one to solve first if the unpin is ever retried.** A
-   `DirectoryOrCreate` hostPath is created fresh by kubelet as **root**, so `seed-nix` fills
-   it and then nothing inside the sandbox can take the nix DB lock:
+   of the same volume, and it is the one to solve first if the unpin is ever retried.** Nothing
+   in the sandbox could take the nix DB lock:
    `error: opening lock file "/nix/var/nix/db/big-lock": Permission denied` — **75
    occurrences across 42 tests, on EVERY devrc PR**, byte-identical on two different branches
    (`devrc-ci-csfzb`/#1057, `devrc-ci-kdcmr`/#1059), against a pre-change run
-   (`devrc-ci-q4d5m`) of `failed=0` over 18,557 passed. The PVC works only because earlier
-   runs populated its `/nix` with ownership the nix build user can use. ⚠ **What the unpin
-   genuinely bought, recorded so it is not re-litigated from zero:** three nodes reachable
-   instead of one, queue wait **17–22m → 0.1m**, wall clock **39.1m median → 17.4m**. Retry
-   it only with the store's ownership solved FIRST and proven on a scratch pipeline — and
-   then reconsider `686d6ff0`, whose `privileged` exemption buys nothing while the PVC is back.
+   (`devrc-ci-q4d5m`) of `failed=0` over 18,557 passed. *(Those figures are quoted from
+   `7839ef54`; the runs are pruned, so they are not re-derivable — the mechanism below is.)*
+   🔴 **"PVC vs hostPath" IS NOT THE DIFFERENCE, AND `7839ef54`'s OWN STATED CAUSE IS WRONG.**
+   `nix-store-cache` **is** a hostPath: its PV is `hostPath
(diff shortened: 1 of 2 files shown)


ZacxDev:
**Audit round 1** (first, full) returned 2 🟡 and 3 🟢, all against the payload — this is a docs PR, so the `.md` IS the payload. All five are fixed in `bfc782cb`. Every measurement was re-derived rather than accepted from the audit report.

Round 1 independently confirmed the PR's live-state claims (gate pod nodeSelector, the `nix-store-cache`/`nix-store-cache-2` PVCs, the `6bec075e`/`7839ef54` provenance and UTC conversion, the `686d6ff0`-buys-nothing claim against 358 live pods, no contradiction with gotcha #3 or the pipeline section, markdown structure, `test_doc_path_rot.py` 77 passed, and that no SKILL.md byte ceiling applies to `tekton`).

It also found the red `tekton/devrc-pytests` is **not** this PR: the sibling parametrization of the same test failed on PR #1118 with `TimeoutError: timed out` out of `socket.py`, and `scripts/tests/test_subsystem_store_api.py:98-107` documents the mechanism in-tree.

```audit-claims round=2 audited=9cb3e56f..bfc782cb
1. 🟡1 — removed the claim that "the PVC works only because earlier runs populated its /nix"; the doc now states that nix-store-cache IS a hostPath (PV hostPath /var/lib/mnt/disk-1/pvc-aef79024…_tekton-ci_nix-store-cache, type DirectoryOrCreate), that the one measured structural difference is local-path-config's `mkdir -m 0777` vs kubelet's 0755 root:root, and that nix-store-cache-2 was created fresh 2026-08-25T05:33:47Z and works — and directs a retry to test `chmod 0777` first, explicitly labelled unproven.
2. 🟡2 — the perf pair (17–22m→0.1m, 39.1m median→17.4m) is now qualified as measured at requests.cpu 4 (23887675), with bb62668f having put it back to 2 fifteen minutes before the revert and 2 being live today; the doc says grading a retry on this pair over-credits the unpin and to re-take the baseline at cpu 2.
3. 🟢3 — added that 686d6ff0's own in-file comment in homelab-infra namespace.yaml is stale and asserts the opposite ("REQUIRED by the gate's hostPath nix cache") over a precondition already satisfied.
4. 🟢4 — corrected the cross-reference from §"Adding a pipeline" to the actual heading §"Adding a new pipeline / new repo".
5. 🟢5 — "2h20m" corrected to 2h19m, and the OBSERVABLE window separated from the git interval: ~1h41m, because admission rejected every gate pod for the first 39 minutes (22:09Z→22:48Z).
6. Additionally, the 75-occurrences/42-tests/18,557-passed figures are now explicitly marked as quoted from 7839ef54 and not re-derivable (the runs are pruned), so the mechanism carries the weight rather than the numbers.
```

ZacxDev:
**Audit round 2** (delta `9cb3e56f..bfc782cb`) confirmed claims 2–5 clean and found two 🟡 against round 1's own fix — **one of them a regression that fix introduced**. Fixed in `b65f1596`. Every measurement re-derived, not accepted.

Round 2 verified independently: the PV hostPath path and type, `local-path-config`'s `mkdir -m 0777`, the full `23887675`/`6bec075e`/`686d6ff0`/`bb62668f`/`7839ef54` timeline with `6bec075e` sitting between the two cpu commits, live `requests.cpu: "2"`, the 39m/1h41m/2h19m arithmetic, that all four named PipelineRuns are pruned, 3-space continuation indent on every added line, zero fence delimiters in the file, and the doc-rot gate both ways.

Ledger: `round 3 · payload lines changed THIS round: 67 (39 SKILL.md + 51 reference/pipelines.md, 23 removed; for a docs PR both .md files are payload, 0 scaffolding) · since round 1: 21`.

```audit-claims round=3 audited=bfc782cb..b65f1596
1. F2 (the regression round 1 introduced) — restored "proven on a scratch pipeline" to the retry instruction, with 7839ef54's reasoning attached ("the place to find the third failure mode is not production") and the enforce_admins blast radius named; measured `scratch pipeline` 1x at 9cb3e56f, 0x at bfc782cb, 1x now.
2. F1 — the refutation is withdrawn: 7839ef54's stated cause is now labelled UNPROVEN rather than refuted, and gitops-validate is explicitly named as NOT a valid control, because the gate's NIX_CONFIG drops to an unprivileged nixbld user while gitops-validate's sets `build-users-group =` and runs as root (both re-derived off live pods).
3. F1 (second half) — the chmod 0777 theory is retained only as a weak candidate with its own counter-argument stated (cp -a preserves modes below the root, 0755 already grants r-x to other, so 0777 adds write at the root level only, which is not where /nix/var/nix/db/big-lock fails); the reader is pointed instead at the gate's existing per-user-profile heal and its MEASURED probe failures.
4. F3 — "has been green since" removed; gitops-validate retention is ~2h so 2026-08-25 is unobservable.
5. F4 — the volatile "358 live tekton-ci pods" replaced with "12 distinct pipeline shapes" plus an instruction not to quote a pod total; the qualitative zero-baseline-forbidden-features result is re-derived and kept.
6. F5 — the perf pair is now marked as quoted from 7839ef54 and not re-derivable, matching the 75/42/18,557 treatment.
7. F6 — rather than grow again, the diagnosis detail was DEMOTED to reference/pipelines.md ("Retrying the devrc-ci unpin"); SKILL.md is now 37,231 B, SMALLER than round 1's 37,991 and 3,729 B under skill-audit.py's HARD = 40,960.
8. Guards: 255 passed in the nix dev shell; the doc-rot instrument was validated by mutation (new reference pointed at a nonexistent file ⇒ test_no_new_dead_paths FAILED at claude/skills/tekton/SKILL.md:288).
```

ZacxDev:
**Audit round 3** (delta `bfc782cb..b65f1596`) confirmed 6 of 8 prior claims outright, found nothing lost in the SKILL.md → reference demotion beyond one datum, and found one 🟡 of the same class that opened this ladder — a recorded WHY the tree falsifies. Fixed in `5ea60d7e`. Both measurements re-derived, not accepted.

Round 3 independently verified: the `scratch pipeline` restoration (1× → 0× → 1×), both pipelines' `NIX_CONFIG` **from the manifests** rather than the pods, the removal of "has been green since", the perf-figure provenance chain, the doc-rot mutation reproduced at `SKILL.md:288`, and the qualitative security sweep over all 431 live pods with a **positive control on 6 of 8 detectors** (cluster-wide: 105 hostPath / 29 hostNetwork / 9 privileged / 30 added-caps / 17 hostPort).

Ledger: `round 4 · payload lines changed THIS round: 49 (33 added / 16 removed, both .md files — for a docs PR both are payload, 0 scaffolding) · since round 1: 196`.

```audit-claims round=4 audited=b65f1596..5ea60d7e
1. 🟡1 — candidate 1 (root-directory mode) is now marked REFUTED rather than "weak": measured on live gate pod devrc-ci-5nqqx, the WORKING PVC root is `755 root:root /nix` (with `755 /nix/var/nix/db`, `600 root:root /nix/var/nix/db/big-lock`, `755 /nix/var/nix/profiles/per-user`) — i.e. exactly what round 2 claimed only the failing arm gets.
2. 🟡1 mechanism — the doc now states that `cp -a src/. dst/` OVERWRITES the destination root's own mode with the source's, controlled locally on GNU coreutils 9.11 with a 0700 source (dst 0755 → 700 AND dst 0777 → 700), so seed-nix erases whatever the volume root was created with and both arms converge before any test runs; and it warns that a chmod before the seed measures nothing while its null result reads as confirmation.
3. 🟡1 corollary — added that `big-lock` is 0600 root:root on the WORKING volume too, so round 2's "0755 already grants r-x to other" never described the file the error names; the section's UNKNOWN framing is retained and strengthened rather than resolved.
4. The heading no longer says "Three measured differences" — it now says one is refuted and two stand (build-users-group, and the gate-only per-user-profile heal).
5. 🟢 restored the one claim lost in round 2's demotion: "the gate pod sets no securityContext and no fsGroup at all", a measured negative that closes off the kubelet-side remedy.
6. 🟢 the "12 distinct pipeline shapes" count is removed as population-ambiguous and drifting (re-derived: 11 by live pod label, 12 by PipelineRun, 13 by Pipeline CR). Replaced by the sweep itself plus an instruction to name the population; the qualitative zero-baseline-forbidden-features result is re-derived and kept, and the pod total is shown drifting 358 → 457 within one session as the reason not to quote it.
7. Out of range but in a file this commit already touches: reference/pipelines.md:99 listed `gitops-validate 6Gi` among the pipelines pinned to talos-xr6-r7p; homelab-infra #396 moved it to talos-uvh-gtj with nix-store-cache-2. Corrected.
8. Guards: 255 passed (test_doc_path_rot, test_skill_audit, test_skill_descriptions, test_skill_tiers) in the nix dev shell. SKILL.md is 37,507 B against HARD = 40,960 (headroom 3,453), still below round 1's 37,991.
```

ZacxDev:
**Audit round 4** (delta `b65f1596..5ea60d7e`) confirmed **all eight** round-3 claims and — for the first time in this ladder — independently reproduced the central mechanism claim rather than merely checking it was quoted right: the `cp -a` control re-run at two coreutils versions **including inside the CI image itself** (9.5), the seed target confirmed as the volume root (`mountPath: /nix-cache`), and the image's own `/nix` measured at 0755, so "both arms converge on 0755" is measured. It also ran a **positive control** on the security sweep (`kube-system`: 89 hostPath, 22 hostNetwork, 9 privileged, 23 added-caps), so the zero is a real zero.

It then found two 🟡. Both fixed in `35750271`, both re-derived first-hand.

```audit-claims round=5 audited=5ea60d7e..35750271
1. 🟡2 (the premise error) — "the gate's NIX_CONFIG means nix drops to an unprivileged nixbld user" is REMOVED as wrong. Measured live (`kubectl -n tekton-ci exec devrc-ci-4dh7x-gate-pod -c step-pytests -- sh -c 'id; nix config show'`): uid=0(root), build-users-group = nixbld, sandbox = true. The step's own nix client is ROOT; nixbld is who nix drops the SANDBOXED BUILDER to, not who takes the store lock.
2. 🟡2 corollary — round 3's "0600 root:root big-lock on the working volume deepens the mystery" framing is removed; it was a paradox only under the wrong premise. The corollary now cross-references candidate 2 for why root opening a 0600 root:root file is unremarkable.
3. 🟡2 relocation — candidate 2 is restated as MISLOCATED rather than refuted: the population that can hit this is the one the gate's own step comment already names — tests that shell out to `nix-instantiate` from INSIDE the build, where nix HAS dropped to nixbld, reaching the real store. Stated as the proposition a retry should be designed against.
4. 🟡1 — the self-contradiction is removed: round 3 deleted `gitops-validate 6Gi` from the pin list but left "*every* other pipeline pins talos-xr6-r7p" standing. Re-derived by census over live pods (group by tekton.dev/pipeline × spec.nodeSelector): 3 pin talos-xr6-r7p (auditloop-ci, devrc-ci, naida-ux-audit), 1 pins talos-uvh-gtj (gitops-validate), 7 pin nothing (clawgate-ci, clawgate-e2e, clawgate-ux-audit, vetr-api-main, vetr-api-pest, vetr-app-unit, vetr-infra-guards).
5. 🟡1 dependents — "clawgate-ci is the odd one out" removed (floating is the majority, 7 of 11), and "clawgate pods land on talos-jkj-deb" removed as false: live count is talos-deu-s2q 30, talos-uvh-gtj 14, jkj-deb ZERO. The bullet now carries the derivation command instead of a list and states that the old claim was false, so nobody re-reads it as current.
6. 🟡1 hazard added — an explicit warning NOT to "fix" clawgate-ci by pinning it to talos-xr6-r7p, which would concentrate a fourth pipeline onto the node the devrc-ci section is trying to take load off.
7. The restored securityContext/fsGroup datum is sharpened with what round 4 measured: pod-level `{}` and `null` on all six step containers.
8. Guards: 255 passed (test_doc_path_rot, test_skill_audit, test_skill_descriptions, test_skill_tiers) under devrc's own flake. SKILL.md unchanged at 37,507 B against HARD = 40,960; only reference/pipelines.md changed this round (30 added / 17 removed).
```

🔴 **Operator note, from round 4's ledger and carried forward:** the ladder's `two consecutive rounds of zero payload lines` stop gate **cannot fire on a docs-only PR** — every fix round is 100% payload by construction. Ending this ladder is a judgement call, not a gate read. Round 4's nits (the PipelineRun count reproducing as 15 over all runs, and the mode readings being image-derived) were correctly left as nits and are not fixed here.

ZacxDev:
**Audit round 5** (delta `5ea60d7e..35750271`) confirmed the round-4 relocation is **mechanically correct** — devrc really does shell out to `nix-instantiate` from inside the checks build (`scripts/testlib/skills_mapping.py:118-125`, `scripts/tests/test_freeze_instrumentation.py:121-125`), with the gate's own measured nested EPERM beside its heal. So the rounds-1–4 pattern of *"each correction over-claims about the mechanism"* did **not** repeat on that axis. It found two 🟡 on other axes, both fixed in `dc9ffaf8`, both re-derived first-hand.

🔴 **This round took round 5's STRUCTURAL advice rather than restating a third number.** The recurring generator was a literal pin list in prose; it is now a derivation command, and the command was **run before shipping** (rc 0, reproduces 4/1/8).

```audit-claims round=6 audited=35750271..dc9ffaf8
1. F1 — the pin census was over the WRONG POPULATION. Round 4 censused live pods, which silently drops any pipeline with none at that instant: it lost `remix-ux-audit` (which DOES pin talos-xr6-r7p; 20 retained PipelineRuns) and `vetr-app-e2e`. Re-derived over TriggerTemplates, the defining population: 13 rows — 4 pin talos-xr6-r7p (auditloop-ci, devrc-ci, naida-ux-audit, remix-ux-audit), 1 pins talos-uvh-gtj (gitops-validate), 8 pin nothing.
2. F1 structural fix — the bullet no longer carries a pin list to be READ. It carries the `kubectl get triggertemplates -o json | jq …` derivation command, states that this bullet has been wrong twice in two days in opposite directions and why, and keeps only the durable qualitative claim (floating is the majority; clawgate-ci is not the odd one out). The jq command was executed against the live cluster before shipping.
3. F1 dependents — the 🔴 hazard's "fourth pipeline" is corrected to "FIFTH", and the remix section's "both pipelines node-pin to talos-xr6-r7p" is reconciled to point at the same derivation, closing the 50-line self-contradiction.
4. F2 — SKILL.md gotcha 6(b) prescribed `nix config show | grep "^sandbox "` as the detector for "sandbox = false in the CI pod". Measured live: `sandbox = true`, `sandbox-fallback = true`, `/build` DOES NOT EXIST, build in `/tmp/nix-build-devrc-pytests.drv-0` — the pod builds unsandboxed and the configured value reads `true`. The prescribed grep returns a reassuring `true` over exactly the condition it exists to catch.
5. F2 fix — the command now leads with `ls -d /build` and the gotcha states that `/build`'s ABSENCE is the tell, that `nix config show` reports the CONFIGURED value, and cross-references the in-tree 🔴 KNOWN LIMITATION at devrc-ci-pipeline.yaml:147-154 where all three fixes are blocked by PodSecurity baseline:latest.
6. Nits taken — the comparative PVC sizes (naida 8Gi, remix 12Gi, auditloop 10Gi) are RESTORED, since dropping them is what made remix's omission invisible; and the bullet now states that being unpinned is a deliberate WIN with its named benefits (clawgate-ci-pipeline.yaml:58,69), not only a cost.
7. Guards: 255 passed (test_doc_path_rot, test_skill_audit, test_skill_descriptions, test_skill_tiers) under devrc's own flake. SKILL.md 38,236 B against HARD = 40,960, headroom 2,724.
8. Not addressed, deliberately: round 5's out-of-scope note that SKILL.md:304's "failed=1 on the dev host / failed=43 in CI" disagrees with devrc-ci-pipeline.yaml:178-181's "dev host failed=0 / this pod failed=1". Both predate this PR's range and neither was touched here.
```

ZacxDev:
## Audit ladder CLOSED after round 6 — `6343c7dc`

Round 6's two 🟡 are fixed and **the ladder stops here, on its own recommendation and for its reason**: a round 7 would exist to re-read two text edits whose correctness is settled by `ls <path>` and one `grep` — the manufactured round the stop rule exists to prevent.

🔴 **Recording that this stop was a JUDGEMENT call, not a gate read.** The ladder's mechanical gate is *two consecutive rounds of zero payload lines*. On a docs-only PR every fix round is 100% payload by construction, so that gate **could never have fired here**. Round 6 flagged this and it is worth carrying forward to the next docs PR that gets a ladder.

**Round 6's answer to the question it was dispatched on** — did round 5's structural fix close the generator, or move it?

> "Replacing the list with a derivation command genuinely removes it: I ran the shipped command byte-for-byte, it exits 0, reproduces 4/1/8, is robust to a missing `podTemplate` and to differently-nested ones, and — verified independently against Pipeline CRs, `config-defaults` and every affinity block — it censuses the right population. **That is the first correction in this ladder that I could not falsify against the tree.**"

It also confirmed the population choice independently: `tekton-ci` holds **13 Pipeline CRs and 13 TriggerTemplates, 1:1 by name**, no orphan either side; `config-defaults` has no `default-pod-template`; and every `kubernetes.io/hostname` affinity in the corpus is `NotIn` (an exclusion, not a pin), so the 8 `<none>` rows are genuinely unpinned.

### The two fixed in `6343c7dc`

| | finding | why it mattered |
|---|---|---|
| **R6-1** | dead external path — shipped `<homelab-infra>/…/tekton-pipelines/ci-priority-classes.yaml`; the file is under `…/tekton-pipelines/**triggers**/` | This is the **one path class gate 0 is designed not to see** (`test_doc_path_rot.py:64-66` — naming the repo in angle brackets makes the token invisible), so nothing would have caught it. The reader it misleads is the one accepting this bullet's own invitation to go fix the stale in-cluster comment — which **is** there, at `ci-priority-classes.yaml:48`. |
| **R6-2** | the delta flagged the stale twin in another repo and left the same assertion in its own **auto-loaded body** — `SKILL.md` called `clawgate-ci` "the one pipeline with no `nodeSelector`" | Round 6 dated it: **true** when measured 2026-08-23 (6 TriggerTemplates existed), **false since 2026-08-26** when the `vetr-*` family landed; 8 are unpinned today. `SKILL.md` loads on trigger while `reference/pipelines.md` — holding the DERIVE-IT instruction and the whole correction — is read on demand, so an agent loading only the body would read the stale line as the pin census and **reproduce the exact error this PR exists to kill**. |

Guards: **255 passed** (`test_doc_path_rot`, `test_skill_audit`, `test_skill_descriptions`, `test_skill_tiers`) under devrc's own flake. `SKILL.md` **38,580 B** against `skill-audit.py`'s `HARD = 40_960`.

### The ladder, six rounds, every one of which found something real

1. the PR's recorded cause was falsified by the tree
2. the R1 fix **deleted a safety instruction** and leaned on an invalid control
3. the R2 fix asserted a "measured difference" that is **measurably absent**
4. the R3 fix rested on a premise the tree contradicts (`nixbld` vs **root**)
5. the R4 fix **censused the wrong population** and left a self-contradiction 50 lines away
6. two one-line edits; the structural fix held

**Ledger:** `round 6 · payload lines changed THIS round: 43 (since round 1: 307)`. Both files are payload throughout — zero scaffolding was touched in any round of this PR.

### Still blocking the merge, and it is not this PR

`tekton/devrc-pytests` is red on a test this diff cannot touch. Round 1 established the mechanism and it is documented in-tree at `scripts/tests/test_subsystem_store_api.py:99-108`: a localhost round-trip losing the scheduler while ~12 pipelineruns share one node, so *"the test logic was never reached, so the gate reported a code failure for a capacity problem."* `main` is `enforce_admins: true` with both legs required, so the leg needs a re-run before this can land.

PR #1109, opened 2026-08-30
test(browser-bridge): sequence the absent-vs-empty origin pair — `where=` never ordered our OWN two rows

ZacxDev:
test(browser-bridge): sequence the absent-vs-empty origin pair — `where=` never ordered our OWN two rows

`test_an_absent_origin_header_is_not_the_same_as_an_empty_one` issued both of
its `tabs` commands and then waited for two rows, unpacking them POSITIONALLY
while its own docstring says order between them is the signal. `emit_cmd_event`
runs off the critical path, after the HTTP response, so which of the two landed
first was the scheduler's decision. This is the identical mechanism that flaked
one test over in #1074 — a pair that came back REVERSED in the sandbox tier —
and it was NOT closed by the `where=_routed_to(inst)` filter both sites carry:
that predicate separates our rows from a NEIGHBOUR's, and cannot order two rows
that satisfy it equally, which ours do (same op, same routing key).

Fix: wait for row one before issuing command two — `_wait_events`' own sanctioned
"order pinned structurally" form. The single routed row returned before the
second command exists IS the first command's, by observation rather than by
argument. `pair[0] == absent` then asserts the append-only order still holds, so
a future regression says so instead of surfacing as a bogus attribution failure.

CONTROL, because a passing test proves nothing about why it passes: swapping the
two commands (keeping the sequencing) turns it RED at `absent["session"]` with
`KeyError: 'session'`, while `pair[0] == absent` still PASSES — so the failure is
the assertions being genuinely order-dependent, not the new guard firing. File
restored byte-identical afterwards (sha256 1b42b227…).

Also corrected three claims this touched, each false before the change:
  * `_wait_ops`' docstring said the site's `where=` "keeps the order". It does
    not, and that sentence is why the hazard sat unseen — replaced with the two
    hazards named separately and the remedy for each.
  * `_wait_events`' bucket totals (`53 = 39 + 5 + 9`, `7 op-selected`) were left
    behind by #1074 and wrong in three places. Re-derived by AST: `52 = 38 + 5 +
    9`, 11 op-selected.
  * the ratchet's ledger comment said "the same 48 sites" while the pin was 47.

Ratchet impact: `_wait_ops where=` 3 -> 4, total 62 -> 63, RATCHETED number
unchanged at 47 and the ledger untouched — sequencing adds a wait, it does not
convert a positional `_wait_events` site. Recorded there, along with the fact
that that module is structurally blind to this hazard class.

Verified: 461 passed in scripts/browser-bridge/tests/test_server.py; 16 passed in
the ratchet; full sandbox tier reported separately.

---

**Scope of the verification claim.** `461 passed` in `scripts/browser-bridge/tests/test_server.py`
and `16 passed` in the ratchet, on the dev host. The tier the merge is gated on is the
sandbox one (`nix build .#checks.x86_64-linux.pytests`, a `cp -r` store copy with no `.git`);
its result is posted in a comment below rather than asserted here.

**What this does NOT close.** The other 46 positional `_wait_events` sites are untouched —
this is the one site in the file that unpacks a pair, and the ratchet's incremental posture
is deliberate. And the ratchet cannot detect this hazard class at all; that is now written
into its pin comment so a green run there is not misread as coverage.

### DIFF
--- a/scripts/browser-bridge/tests/test_server.py
+++ b/scripts/browser-bridge/tests/test_server.py
@@ -312,15 +312,31 @@ def _wait_ops(spool_dir, op, n=1, where=None, **kw) -> list:
     one test carry the SAME op and the SAME routing key, so no per-row predicate
     can tell them apart; their file order is decided by when each `emit_cmd_event`
     runs, which is after the HTTP response and off the critical path. A caller
-    unpacking a pair positionally must ALSO sequence: wait for row one, then
-    issue command two. `#1074` is the measured case — a pair issued together came
-    back REVERSED in the sandbox tier with `where=` already in place.
+    reading N rows of one op positionally must ALSO do something about order —
+    either SEQUENCE (wait for row one, then issue command two) or make the
+    assertion ORDER-INDEPENDENT.
 
-    THE ONE SITE IN THIS FILE THAT UNPACKS A PAIR DOES BOTH.
-    `test_an_absent_origin_header_is_not_the_same_as_an_empty_one` routes both
-    of its commands to an instance id minted per run and selects with
-    `where=_routed_to(that id)`, so the pair it unpacks is the pair it caused;
-    and it waits for the first row BEFORE issuing the second command, so which
+    🔴 `#1074` IS THE MEASURED CASE, AND READ WHAT ITS FIX ACTUALLY DID — the
+    earlier draft of this paragraph said the reversal happened "with `where=`
+    already in place", and that is FALSE.
+    `test_the_two_origin_tokens_are_distinct_and_recorded_verbatim` flaked while
+    it was a bare positional `_wait_events(spool_dir, len(ORIGIN_TOKENS))`;
+    `#1074` ADDED the `where=`. And `where=` is not what fixed its order: that
+    site also became `sorted(...) == sorted(ORIGIN_TOKENS)`, i.e. it stopped
+    depending on order at all. Two halves, two remedies — which is exactly the
+    point, because a site whose order IS the signal cannot take the sorting one.
+
+    THE TWO SITES IN THIS FILE THAT READ A PAIR POSITIONALLY BOTH SEQUENCE, and
+    neither can sort, because for both of them order between the rows is the
+    assertion:
+      * `test_an_absent_origin_header_is_not_the_same_as_an_empty_one` (absent
+        vs empty `origin` header), and
+      * `test_a_neighbours_row_of_the_same_op_is_not_selected_as_one_of_ours`,
+        the guard for the first — which carried the identical unsequenced race
+        until it was measured red under the same swap-the-commands control.
+    Each routes both commands to an instance id minted per run and selects with
+    `where=_routed_to(that id)`, so the rows it reads are the rows it caused;
+    and each waits for the first row BEFORE issuing the second command, so which
     of the two comes first is fixed by observation rather than by scheduling.
     That is an INVARIANT THE CODE ENFORCES, replacing the corpus property this
     docstring used to assert ("no current test leaves a `tabs` command in flight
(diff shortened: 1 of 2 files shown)


ZacxDev:
Round 1 (blind adversarial audit) returned three findings; all three were independently re-derived here before fixing, and the fixes are in `e4777c58`.

```audit-claims round=1 audited=2579e2f3..e4777c58
1. `_wait_ops`' docstring no longer claims #1074's pair reversed "with `where=` already in place"; it now states that #1074 ADDED the `where=` and that the site's order was fixed by sorting, not by `where=`.
2. The sentence "THE ONE SITE IN THIS FILE THAT UNPACKS A PAIR" is replaced; it now names both pair sites and no longer names a test that (after this PR) does not unpack a pair.
3. `test_a_neighbours_row_of_the_same_op_is_not_selected_as_one_of_ours` is now SEQUENCED (waits for row one before issuing command two) and no longer unpacks `first, second` positionally.
4. That test's `# THE FIX: where= keeps the pair THIS test caused, in order` comment is removed and replaced with text recording that the claim was false.
5. The `pair[0] == absent` guard's comment no longer claims it would detect a lost sequencing; it is narrowed to the append-only-order invariant and states that nothing there can detect the sequencing's removal.
6. `test_the_two_origin_tokens_are_distinct_and_recorded_verbatim`'s docstring no longer says sequencing the sibling "belongs in its own PR"; it is marked resolved and names both sites.
7. The ratchet pin comment's bucket accounting is updated to `_wait_ops where=` 5 and 64 total call sites, with the RATCHETED total unchanged at 47 and the ledger untouched.
```

Ledger: round 1 · payload lines changed THIS round: 78 (`test_server.py` 56/22, ratchet 10/6) · measured with `git log --numstat --format= --remerge-diff 2579e2f3..e4777c58 --not origin/main`, rc 0, non-empty range, git 2.55.0.

ZacxDev:
Round 2 (delta re-audit, blind to round 1's reasoning) dispositioned all seven round-1 claims — six **actually fixed**, one **partially** — and returned one new 🟡. Independently re-derived here before fixing. Fix is in `621ef6c2`.

```audit-claims round=2 audited=e4777c58..621ef6c2
1. The census sentence no longer reads "THE TWO SITES IN THIS FILE THAT READ A PAIR POSITIONALLY"; it is narrowed to the two `_wait_ops(..., where=)` pair sites, which is the scope for which "two" is correct.
2. The docstring now names the four sites the widened verb silently excluded (`test_an_oversized_id_is_dropped_whole_never_truncated`, `test_both_join_sites_answer_the_same_way_for_every_joinable_tier`, `test_activate_telemetry_records_the_consent_decision`, `test_heartbeat_tracks_registry_liveness_across_the_stale_boundary`) and states they take row [1] off a bare `_wait_events(spool_dir, 2)`.
3. The docstring now states which hazard each population is exposed to: the two `where=` sites had the OWN-rows hazard (closed by sequencing); the four bare sites have the FOREIGN-row hazard (not closed here, and counted among the ratchet's 47).
4. The ratchet's pin comment carried the same widened wording and is corrected the same way, with an explicit warning not to read that sentence as a census of the ratcheted population.
```

**Ledger: round 2 · payload lines changed THIS round: 37 (`test_server.py` 19/1, ratchet 13/4) · since round 1: 115.** Measured with `git log --numstat --format= --remerge-diff e4777c58..621ef6c2 --not origin/main`, rc 0, non-empty range, git 2.55.0.

⚠ **Read that ledger with a caveat I want on the record.** This range is **comment-only** — no test behaviour changed (`grep -v '^[+-]#'` on the ratchet diff is empty; the `test_server.py` hunk is entirely docstring). Round 1 genuinely changed behaviour (it sequenced a second test); round 2 did not. Round 1's auditor scored the whole range "100% scaffolding, payload 0" on a file-type reading, which the skill explicitly warns against — but the stricter honest reading here is that this PR exists to ship *both* a sequencing fix and corrections to false docstrings, so the prose is payload while the ladder is still finding false claims in it. **If round 3's fixes are also comment-only, that is two consecutive rounds and the attribution gate fires: the ladder has left the PR, and the remainder should be filed rather than audited further.**

ZacxDev:
## Ladder CLOSED at round 3 — on the attribution gate, not on a clean round

Round 3 dispositioned all four round-2 claims as **actually fixed** and returned one 🟡 + one 🟢, **both artifacts of round 2's own edit** (a severed antecedent and a rotted relative pointer — neither would exist had round 2 not run). Fixes are in `2e6d8dbc`.

**Why stop here.** Rounds 2 and 3 each changed **zero executable lines** — round 3's auditor confirmed this mechanically rather than by eye: `ast.dump()` of both files is identical across the round-2 range once docstrings are stripped. Two consecutive zero-payload rounds is the skill's condition for "the ladder has left the PR", and round 3's independent judgement said the same thing unprompted: *"the marginal defect being found is now generated by the process rather than detected by it."*

**The result that justifies the whole ladder, and indicts it.** `(+ 11 op-selected)` in `_wait_events`' docstring was **false**, and had been for two commits. Measured: `_wait_ops` 9 + `_wait_payload` 3 = **12**. `git log -S` attributes the line to `2579e2f3` — this PR's own first commit — and `e4777c58` staled it by adding a `_wait_ops` call while sequencing the guard test.

🔴 **All three delta rounds missed it, and the reason generalises: a delta round diffs only the newest range, so a claim that an EARLIER commit of the same PR staled is outside every subsequent round's range, permanently.** It sat four lines from the paragraph all three rounds were editing. Nothing pins it either — no assertion reads `op-selected` — so a fully green suite is silent on it. It was found by the audit of the *other* PR, looking at the same number from outside.

**Ledger:** round 3 · payload lines changed THIS round: 0 executable / 32 comment · since round 1: 147 total, of which round 1 alone carried the behaviour change (sequencing the guard test).

**Not fixed, recorded instead:** nothing outstanding on this PR.

PR #1110, opened 2026-08-30
fix(dl-router): save the file that was POSTED, not the proxy thumbnail — and give a signed CDN URL a ledger key it can be found by

ZacxDev:
Discord is **59% of this router's live traffic** (20 of 34 routed downloads; all 5 identity aliases are `discord:` channel rows). The *routing* half has been Discord-specific since the first evening — `route_core.js` says so in its own header. Three things on the **save** side were still wrong for it.

### 1. The context menu saved the thumbnail

`onMenuClicked` read `info.srcUrl || info.linkUrl`. Discord serves one attachment from two hosts: a downscaled, re-encoded webp from its resizing proxy in the ``, and the posted original on the wrapping `<a href>`. `srcUrl` winning meant *Save to library…* on an image saved the **proxy copy** and reported it as the download.

`preferOriginalUrl` swaps only when both URLs are Discord attachments **with the same path** — provably one asset at two resolutions. A link that merely wraps an image is still never followed, and an avatar sharing the same hosts is never swapped.

> ⚠️ The **defect** is proven: on pre-change code the new test downloads `…a.png?format=webp&width=550`. Whether it **matters** rests on a premise I could not confirm in a live client (the bridge lost host access mid-check) — that Discord puts the proxy URL in `src` with the original on the wrapping anchor. If that is wrong this is a harmless no-op, not a regression. Items 2 and 3 do not depend on it.

### 2. The source-URL ledger could only ever miss on Discord

Every Discord CDN URL is signed (`ex`/`is`/`hm`) and re-signed on each page load, so the query names **the request, not the asset**. `store.source_url_key` keeps the query deliberately — and is right to, for the sites it was written for. Discord is the exception, so `discordSourceKey` expresses the exception once rather than weakening that rule for everyone.

All 7 Discord ledger rows today carry a signature, so a re-download could only write a **new** row — the failure `haveUrl` already names in its own comment: *"a badge that never lights actively asserts you do not have this."* **Structural, not yet observed**: no two rows share a path, so nothing has collided in practice.

It beats the capture's own key on purpose — that key is the embed *page*, and one Discord channel URL names thousands of assets.

### 3. A Discord video went to yt-dlp with the page URL

`mediaType === "video"` sent every right-clicked video down the yt-dlp branch, handing it `discord.com/channels/<guild>/<channel>` while the `.mp4` sat in `srcUrl`. The bypass is scoped to CDN attachments only. **Found while implementing 1 and 2 — not part of the original plan**; say the word if you'd rather it were split out.

## Verification

| control | result |
|---|---|
| dl-router node suite | **508 → 524** (+16) |
| **RED at `origin/main`** with these tests | `identity.test.mjs` fails to link (both functions genuinely new); both behavioural menu tests fail on assertions |
| invariant guard, labelled as one | *"a NON-Discord video still takes the yt-dlp path"* **passes on pre-change code** — it pins unchanged behaviour, so it is not regression coverage |
| mutation sweep | **4/4 killed, each by its own named test** — drop the path-equality check · keep the query in the ledger key · flip the `sourceKey` precedence · remove the direct-file bypass |
| dev-host tier | node `1436/1436` · pytest `19290 passed, 0 failed` (floor 18145) |
| **sandbox tier** (what Tekton gates on) | nodetests `1436/1436 floor=1367`, built **alone** per the combined-build contention warning; pytests in flight |

The ledger-key assertion is driven off the **shared** `url_cases.json` rows rather than a second hand-copied list, so a URL shape added there is covered here too.

Manifest `0.3.0 → 0.3.1` — an extension change needs a **full Brave restart**, and the version is how `brave://extensions` proves the restart took.

## Not in this PR

A hover *Save* button on Discord is config-only (`[site_rules."discord.com".player]`) and needs no code — but its selector depends on the same unconfirmed DOM premise above, so it stays an experiment rather than a commit.


ZacxDev:
**Round 1 audit — 6 findings, no 🔴.** Fixed in `f7994dde`. The auditor also independently reproduced the mutation sweep and added a fifth mutant of its own.

I re-measured F1 and F2 myself before fixing rather than taking the report on trust; both reproduced exactly.

Two of the six were defects in **my own guards**, which is the part worth recording:
- **F2** — the table-driven ledger-key guard asserted `=== url.split("?")[0]`, not the actual contract. A legitimate new fixture row with a `#fragment` would have RED-ed it while the implementation was correct — a guard that punishes the fixture growth its own comment advertises.
- **F4** — `"a NON-Discord video still takes the yt-dlp path"` asserted only an absence (`downloads.length === 0`) and stayed **green** when `startFetch` was stubbed to a no-op. It was passing for a reason other than the one it names. Its fixture was also wrong: `auto:false` queues the picker and POSTs nothing, so it could not have exercised the path it was named for.

One finding, **F3**, is deliberately **not** fixed: nothing records whether the proxy→original swap ever fires, and all 20 live Discord downloads are already origin-host, so the premise has no live instance behind it. The audit suggested a log line; this extension has no logging facility and inventing one is out of scope here. The docstring now says plainly that it is unobservable, and names the unhandled full-size-proxy variant rather than leaving it implicit.

Also caught by a repo guard I had not read — `tests/source_hygiene.test.mjs` requires every extension source to be plain ASCII, and my emoji comment markers broke it. `route_core.js:107` documents that rule.

**Verification of the fix round:** 524 → 527. Mutation sweep on the fixes, **4/4 killed, each by its own named test** — including the `startFetch` no-op mutant that previously **survived**. Both extension sources verified byte-identical to their pre-mutation copies afterwards. Sandbox `nodetests` on the fixed tree: `1439/1439 floor=1367` (the count moving 1436→1439 is the positive control that it rebuilt rather than serving a cache).

```audit-claims round=1 audited=27fd6660..f7994dde
1. discordSourceKey now folds the host to the origin, so the proxy copy and the original produce ONE ledger key.
2. The table-driven ledger-key guard asserts properties (no query, no fragment, host folded, path preserved) instead of a re-derived string; exact values moved to hand-written literal tests.
3. "a NON-Discord video still takes the yt-dlp path" now also asserts the POSITIVE half — that /fetch was POSTed with the page url — and its fixture changed auto:false -> auto:true.
4. The manifest (.m3u8/.mpd) test now sits OUTSIDE the direct-file bypass, so a Discord-hosted manifest is still treated as a stream.
5. The mediaType==="video" comment no longer asserts the unverified blob: rationale; it states what is known and marks the rest unverified.
6. F3 was NOT fixed: no observability was added. The docstring instead states the swap is unobservable and names the unhandled full-size-proxy variant.
7. Non-ASCII characters removed from both extension sources to satisfy tests/source_hygiene.test.mjs.
```

ZacxDev:
**Round 2 — 2 🟡 + 2 🟢. The fix round introduced the next finding**, which is the shape the ladder exists to catch. Fixed in `f76a3142`.

**F-A is the one worth reading.** My new manifest guard asserted only `downloads.length === 0`. The auditor measured — and I reproduced on my own tree — that inserting `if (manifest && directFile) return;` at the top of the streaming block, i.e. **a Discord manifest doing nothing at all**, left the suite at **527/527, fully green**. That is the exact absence-only defect the *previous* commit had just fixed in the F4 test, reintroduced one test later, by me, in the same file.

**F-B is a regression this PR created.** The `manifest && directFile` cell only became reachable when I moved the manifest test outside the bypass — and in that cell the code handed yt-dlp `discord.com/channels/<guild>/<channel>`, an authenticated SPA route, while **discarding `target`**, the manifest yt-dlp actually consumes. I verified the failure is invisible: `/fetch` has no URL allowlist so the job is accepted, `submitFetch` toasts "filed to `<dir>`" as soon as the POST returns, and **nothing ever polls the job** — `jobId` appears once across all 10 extension sources, returned and dropped. Success toast, no file. My own comment said the move was made to avoid a silent wrong answer; for that cell the new path *was* the silent one.

**F-C** was a seam: the fold split the write key from the read key, since `haveUrl` still used bare `playerSourceKey`. Not reachable today (no Discord player rule exists) and nothing pinned it. The new test pins the **relationship**, not either component — neither side was wrong alone.

### Two corrections to my own round-1 commit message

1. I wrote that a new fixture row with a `#fragment` red-ed the old table guard. True, but **not the operative reason** — `url_cases.json` already held a `media.discordapp.net` row, so the host fold alone forced the rewrite regardless. I presented a consequence of the payload change as an independent scaffolding fix.
2. "the mutant that previously SURVIVED" is killed by **6** tests, not 1. True of that named test in isolation, never of the suite.

**Verification:** 527 → 528. Mutation sweep on this round's fixes, **3/3 killed, each by its own named test** — including the mutant that survived round 2. `service_worker.js` verified byte-identical to its pre-mutation copy.

```audit-claims round=2 audited=f7994dde..f76a3142
1. The manifest test now uses auto:true and asserts the POSITIVE half - that /fetch was POSTed - so a Discord manifest doing nothing no longer passes.
2. It also asserts the manifest URL itself is handed to yt-dlp, not the Discord page url.
3. onMenuClicked hands `target` (the manifest) to startFetch for the manifest && directFile cell; every other streaming case still gets the page url.
4. haveUrl now folds via discordSourceKey before falling back to playerSourceKey, so the /have read key matches the ledger write key.
5. A new test pins the write-key/read-key relationship using a hand-spelled literal, not a value computed from the function under test.
6. The "THE HOST IS FOLDED" comment now says authority (host + scheme + port) and states that the tests exercise the host axis only.
7. Two claims in the round-1 commit message were corrected in the round-2 message: the #fragment rationale, and "killed by 1 test" vs 6.
```

ZacxDev:
**Round 3 — one 🟢, no 🔴/🟡.** Not a clean round, so the ladder continues. Fixed in `1d84e87c`.

The finding is the **consolidation rule working as a bug-finding instrument**, not as hygiene.

There are **two** sites that turn one URL into a `source_urls` key — `playerDownload` (write) and `haveUrl` (read). Round 2 folded the reader and left `service_worker.js:1321` calling `playerSourceKey` directly. So my round-2 claim *"Both sides fold now"* was true of `route_core.js` and **false of that line**: a `media.discordapp.net` embed url would be written unfolded and read folded — a lookup asking for a string the writer never stored.

Note the shape, because it is worse than a plain miss: **before round 2 those two agreed.** The round-2 fix converted a working sub-case into a broken one while fixing the case it was aiming at. Net-better, not free.

The expression had been spelled three different ways across two files for exactly one round, and that is precisely how the two came apart. It is now one exported `ledgerSourceKey`, called at both sites.

**Deliberately not consolidated:** `buildMatchPayload`'s `sourceKey`. It looks like the same rule and isn't — it arbitrates between *two* candidate keys and yields `""` for an ordinary download so the sidecar falls back to the full URL. Folding it in would mint a key for every download on every site. That is not an opinion in a comment: **mutant Q3 does exactly that and three pre-existing tests catch it**, so the boundary is pinned.

Two comments corrected, both of which had become false — `haveUrl`'s docstring header still described the old `playerSourceKey` rule, and `discordSourceKey` pointed at "the tests below" when they live in `tests/identity.test.mjs`.

**Verification:** 528 → 529. Mutation sweep **3/3 killed, each by its own named test**, including the exact defect this round found. Both extension sources verified byte-identical to their pre-mutation copies.

```audit-claims round=3 audited=f76a3142..1d84e87c
1. A single exported ledgerSourceKey(url) = discordSourceKey(url) || playerSourceKey(url) now exists in route_core.js.
2. playerDownload's sourceKey (service_worker.js) calls ledgerSourceKey instead of playerSourceKey, so the write folds a Discord attachment.
3. haveUrl calls ledgerSourceKey, so the read uses the identical expression rather than a separately-spelled one.
4. buildMatchPayload's sourceKey was deliberately NOT routed through ledgerSourceKey; a mutant that does so is caught by three pre-existing tests.
5. A new test pins the player WRITE's fold using a hand-spelled literal, not a value computed from the function under test.
6. haveUrl's docstring header no longer claims the key is normalised through playerSourceKey.
7. discordSourceKey's comment now names tests/identity.test.mjs instead of "the tests below".
8. playerSourceKey was removed from service_worker.js's imports, having become comment-only there.
```

ZacxDev:
**Round 4 — one 🟢, nothing else.** Fixed in `a7a15880`.

Round 3's insertion put `ledgerSourceKey` **and its own JSDoc** between `discordSourceKey`'s 35-line docstring and `discordSourceKey` itself. Read top-down, that first block then documented the wrong function — and it asserts a contract `ledgerSourceKey` does **not** have (that the result is `""` for a non-Discord input, and that the authority is always folded). The concrete wrong action it invites: "simplifying" `haveUrl`'s `if (!key)` on the belief that a non-Discord URL yields `""`.

**Proven pure relocation, not asserted.** Stripping comments and blank lines from both sides gives **473 code lines each, and the two sets are identical** — the only difference is the position of the three `export function ledgerSourceKey` lines. No executable line changed; function declarations hoist, so call order was never load-bearing. 529/529, unchanged, as a pure move must be. No mutation sweep, deliberately: there is no new behaviour to mutate, and set-equality is the stronger check — a mutant tests what the code *does*, and the claim here is that it does exactly what it did before.

One prose correction from round 4's non-finding list: the docstring said the expression "was spelled three different ways", counting `buildMatchPayload` among the three while the next paragraph declares that expression **not** the same rule. It now says two, and says why the third is excluded, so the paragraph no longer argues with itself.

⚠️ Unrelated but worth recording, since I nearly reported it as a defect: `git diff --stat origin/main..HEAD` on this branch shows ~3,900 deletions across `drift-check.sh`, `run-tests.sh`, `handoff_doc.py` and their tests. **None of that is this PR.** `main` has moved 6 commits past my base, so its newer work reads as removals from the branch's side. Against the merge base the change is **5 files, +486/−16, all under `scripts/dl-router/`**. Same moved-operand trap that bit the reading of #1068 earlier today — diff against the merge base, not the tip.

```audit-claims round=4 audited=1d84e87c..a7a15880
1. ledgerSourceKey and its JSDoc moved to sit BELOW discordSourceKey, so discordSourceKey's docstring abuts discordSourceKey again.
2. The move is claimed to be pure: comment-stripped code lines are the same 473-line SET on both sides, only reordered, and no executable line changed.
3. The ledgerSourceKey docstring now says the rule was spelled TWO ways (the write in playerDownload and the read in haveUrl), not three, and states why buildMatchPayload is excluded.
4. No behaviour change is claimed at all; the suite is unchanged at 529/529.
```

PR #1111, opened 2026-08-30
docs(handoff): rank 1 shipped as #1109 — `where=` never ordered our OWN two rows, and the docstring saying it did is why the audit stopped

ZacxDev:
Updates `claudedocs/handoff-audit-pr-ladder.md` for the rank-1 work, landed as #1109.

- **Rank 1 shipped, not merged.** The sibling ordering race is fixed by SEQUENCING. `where=_routed_to(inst)` separates your rows from a *neighbour's*; it cannot order two rows that satisfy it equally, and that site's two do (same op, same routing key). `_wait_ops`' docstring said it "keeps the order" — that sentence is why the hazard survived an order-safety pass, and it is the second time this same site defeated one by a different mechanism.
- **Recorded honestly as in flight:** the sandbox tier was still building at write time, with a sibling session contending on the nix store. The doc says a green from that run is trustworthy and a red is not until re-run alone, and names the log.
- **New rank 1** carries `forcing: gate`; rank 2 (drift-check rc 17) stays `forcing: none`.
- Landed from its own worktree because the shared clone `~/workspace/devrc` had been switched to another session's branch (`docs/handoff-bb-resume-0830`) mid-session. Recorded in the doc.

### DIFF
--- a/claudedocs/handoff-audit-pr-ladder.md
+++ b/claudedocs/handoff-audit-pr-ladder.md
@@ -16,40 +16,47 @@ sessions, then fix what the measurement exposed. It exposed that the ladder's
 findings-keyed stop rule does not terminate in the guard-hardening regime.
 
 ## State now
-- Work is on `fix/sequence-absent-vs-empty-origin-pair` (worktree
-  `~/workspace/devrc-seq-absent-empty`, commit `2579e2f3`), pushed, open as **`#1109`**.
-  This doc's own update is **`#1111`** (`docs/handoff-audit-pr-ladder-1109`).
-- ✅ **RANK 1's CLOSING CONDITION IS MET. BOTH SANDBOX TIERS ARE GREEN** — built ONE AT A TIME
-  against the branch tree, and the verdict read from each derivation's own log:
-  - `pytests` (`gwzds1c6…-devrc-pytests.drv`) → `RESULT: PASS (exit=0)`, `PASS 48  FAIL 0`,
-    `0` occurrences of `panic: test timed out`.
-  - `nodetests` (`gmaidwminm5hwgw198g3swka1sgjk8c5-devrc-nodetests.drv`) → `RESULT: PASS
-    (exit=0)`, four `# fail 0` blocks (508 / 21 / 188 / 134 tests).
-- **The concurrent sibling build does NOT undermine that green.** A second session was building
-  `pytests` at the same time; CLAUDE.md's rule is directional — a contended run fails loudly
-  rather than faking a pass, so a contended GREEN is trustworthy and only a RED would have needed
-  re-running alone. No re-run was required.
-- ⏳ **NOT MERGED — both PRs' required checks are `PENDING`** (`tekton/devrc-pytests`,
-  `tekton/devrc-nodetests`; `mergeStateStatus: UNKNOWN` while they run). 🔴 Watch for the
-  documented failure mode: a run that hits `timeouts.tasks` never posts, and the check stays
-  `pending` FOREVER — only a fresh push clears it. If these are still pending much later, that is
-  the thing to check, not the diff.
-- **Dev-host tier, separately:** `461 passed` in `scripts/browser-bridge/tests/test_server.py`,
-  `16 passed` in the ratchet.
-- 🔴 **DO NOT ASSUME THE BASE CLONE IS ON `main` — IT WAS NOT.** `~/workspace/devrc` sat on
-  `main` at session start and on **`docs/handoff-bb-resume-0830`** (tracking
-  `origin/docs/handoff-browser-bridge-tab-ref`, one commit `28ff9d2a`) by the time this doc was
-  written — another session's live branch in the shared checkout. Caught by
-  `git branch --show-current` immediately before the handoff write; without it `handoff_doc.py`
-  would have committed this doc onto that session's branch, silently and with no conflict.
-  **Re-check the branch before every write here.**
-- **PROVENANCE CARRIED FORWARD** (all merged, all verified by content; kept because this section
-  is REPLACED on every update): `#1035` → `ccb31628`, `#1060` → `31cd214d`, `#1074` → `e9f8ce14`,
-  `#1023` → `8e33bf1d`, `#1033` → `70eff59c`, `#1009` → `442bde83`, `#1005` → `aaa5514c`.
-  All ranked items from those rounds are CLOSED.
-- **Base:** branched off `2cec1d45`; `origin/main` moved to `9e23c379` during the session.
-  **No open PR of the 30 touches either file** (checked via `gh pr list --json files`).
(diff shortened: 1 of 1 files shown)


ZacxDev:
Round 1 (blind adversarial audit) returned one 🔴 and six 🟡, plus nits. All were independently re-derived here before fixing — including a paired positive/negative control on the `nix build` logging mechanism. Fixes are in `b1d6cb3f`.

```audit-claims round=1 audited=c6810f47..b1d6cb3f
1. The `| tail` gotcha is RETRACTED and rewritten: it now says `nix build` prints no build log without `-L`, that `tail -40` discarded nothing, and that with `-L` a `tail -40` would have KEPT the `RESULT:` line.
2. The "How to verify" sandbox-tier block no longer tells the reader to read a `RESULT:` line from a bare `nix build`; it uses `-L` plus `nix log <drv>` and no longer points at the shared base clone.
3. A new bullet states that an already-built derivation prints nothing, so silence is not a pass.
4. Rank 2 (drift-check rc 17) is REMOVED from the ranked list and recorded in State now as CLOSED, with the re-measured evidence (subtree count 0, clawgatectl 0.8.18 not 0.8.17).
5. The nodetests tally is corrected from four `# fail 0` blocks to FIVE, and now names the omitted browser-bridge target (569 tests).
6. The unverifiable `sha256 1b42b227…` corroboration is removed and replaced with a statement of what is actually checkable.
7. The "keeps the order" sentence is re-attributed from `_wait_ops`' docstring to `_wait_events`'.
8. The State-now / ranked-list contradiction about whether rank 1 is done is removed; State now no longer asserts rank 1's closing condition is met.
9. "Three stale counts" is corrected to FOUR; the "no open PR of the 30" claim is marked non-durable; the base clone's transient sha `28ff9d2a` is removed in favour of the branch NAME.
10. The claim that a sibling session built "the same derivation" is corrected — it was the same check ATTRIBUTE of a different tree, hence a different `.drv`.
```

Ledger: round 1 · payload lines changed THIS round: the doc itself (this is a docs PR, so the `.md` IS the payload) · measured with `git log --numstat --format= --remerge-diff c6810f47..b1d6cb3f --not origin/main`, rc 0, non-empty range, git 2.55.0.

ZacxDev:
Round 2 (delta) dispositioned all ten round-1 claims — six **actually fixed**, three **partially**, **one NOT fixed** — and returned four new 🟡, no 🔴. All independently re-derived here, including re-running the failing-build control. Fixes are in `5b1cffb1` + `25a3c9fd`.

🔴 **The "not fixed" one is the most useful result of the round:** round 1 claimed the unverifiable `sha256` citation had been *removed*. It had not — a retraction was **appended under Gotchas while the original line kept making the claim**, ~140 lines apart. `handoff_doc.py` merges those sections by appending, so a delta cannot remove a sentence there. That is now recorded as a general trap, and those corrections were made by direct edit with the reason in the commit message.

```audit-claims round=2 audited=b1d6cb3f..25a3c9fd
1. The `sha256 1b42b227…` claim at the CONTROL bullet is rewritten; both remaining mentions of that digest are retractions, not claims. (`grep -n 1b42b227` → 2 hits, both retractions.)
2. The `| tail` retraction is qualified to "a build that SUCCEEDS", names `log-lines` (25 here) as what governs the failure case, and states the non-tty and not-already-cached preconditions.
3. A measured counter-example is recorded: a FAILING build with no `-L` prints rc 1, 35 stderr lines, 24 builder lines including the `RESULT:` line.
4. The `op-selected` count is corrected 11 -> 12, with the reason it went stale (my own round-1 fix added a third `_wait_ops` call) and the structural tell (no assertion reads the number).
5. The "How to verify" bucket re-derivation no longer hardcodes the shared base clone; it takes an explicit `<TREE>` argument and asserts `TARGET_DIR` resolves under it.
6. The verify block's greps carry positive controls, and the `RESULT:` grep is annotated as returning TWO hits with the LAST being the verdict.
7. A new gotcha records that an append-only section cannot be corrected by appending elsewhere, and that a number nothing asserts on will drift unnoticed through a green suite.
```

**Ledger: round 2 · payload lines changed THIS round: 96 · since round 1: 293.** Measured with `git log --numstat --format= --remerge-diff b1d6cb3f..25a3c9fd --not origin/main`, rc 0, non-empty range, git 2.55.0. For a docs PR the `.md` IS the payload, so this gate cannot fire here — which is a **recorded structural blind spot** of the ladder on prose PRs, not a clean bill of health. Still outstanding from this round and deliberately not fixed here: O1 (the `devrc-mergegate-1073` attribution is from memory with no surviving artefact) and O2 (a one-host store hash cited for a two-host claim; "37+ open PRs" is now 35).

ZacxDev:
## Ladder CLOSED at round 2 — on the STATED CRITERION, because the arithmetic gate structurally cannot fire here

Round 2's four 🟡 are fixed (`5b1cffb1` + `25a3c9fd`). I am not running a round 3, and the reason is worth stating rather than leaving as a silent choice.

**This doc's own Gotchas already records the condition:** *"The `#900` attribution gate cannot rescue a ladder whose payload IS PROSE. It stops after two zero-payload rounds; here 'fixed a defect' and 'reworded a warning' are frequently the same edit. Structural blind spot on any generator/docs/prompt PR."*

That is exactly this PR. For a docs PR the `.md` **is** the payload, so every round is by construction non-zero-payload and the arithmetic gate can never fire. Meanwhile round 2's findings were mostly about text round 1 wrote — which is the shape the gate exists to detect but cannot see here. Left alone, this ladder does not terminate.

**So it stops on the criterion this thread adopted the last time it hit this regime:** no 🔴, no blast radius beyond "the doc contains a false sentence", and the recurring shape swept at every site rather than at the one that was reported. Round 2 met all three:
- **0 🔴.**
- Every finding was a claim a reader could not rely on — none breaks a system.
- The recurring shape (a corrected count going stale, a retraction that did not remove) was swept: `grep -n 1b42b227` now returns 2 hits and **both are retractions**; the `op-selected` count was corrected in this doc **and** chased into `#1109`, where it turned out all three of that PR's audit rounds had walked past it.

**Deliberately NOT fixed, recorded so the next reader knows they are open:**
- **O1** — the sibling-build attribution to `devrc-mergegate-1073` is from memory; no worktree, directory, store path or `.drv` survives to confirm it. The doc concedes the overlap is unmeasured but states the tree identification as fact. It should say "attributed from memory".
- **O2** — a one-host store hash (`p85k4nyi…`) is cited for a two-host claim; the laptop's is `zc8ms7yf…`, binaries byte-identical, so the substantive claim stands. And "37+ open PRs" is now 35 — though that bullet's operative instruction is already "re-derive, don't quote".

Both are 🟢. Fixing them would start round 3 in a ladder that cannot terminate on its own.

**Ledger:** round 2 · payload lines changed THIS round: 96 · since round 1: 293. The gate is inert here by construction — that is a recorded structural blind spot, not a clean bill of health.

PR #1120, opened 2026-08-30
feat(ci-speedup): measure what scripts/tests actually reads — the ceiling is 1.08x, so do not build rank 3 yet

ZacxDev:
Rank 2 of `claudedocs/handoff-ci-speedup.md`. That doc flags its own input 🔴 **NOT TRUSTWORTHY** — a regex called 32/139 files "repo-wide scanners" and over-classifies, flagging `test_drift_check.py` on a line where `git ls-files` sits inside a **fixture string**. It asked for empirical measurement. This is it.

> ⚠️ **This description was corrected after an adversarial audit** — see [the correction comment](https://github.com/innovation-upstream/devrc/pull/1120#issuecomment-5471441437). The original said 22/18/93 and a 1.49x ceiling. Both were wrong.

## Method

Read-set tracing: a `sys.addaudithook` records what each test file opens, lists, scans and spawns. Attributed to **both** collection and runtest, because module-level reads are real — several files read `nix/home.nix` at import, which a runtest-only hook loses. Timing from a **separate untraced run** so hook overhead is not in the numbers.

## Result — all 144 files classified

🔴 **Two denominators, named.** 144 files classified; 133 have timing pytest will show (`--durations` hides sub-5ms). Conflating them is what produced the original wrong table.

| bucket | files | of which timed | seconds | share of timed |
|---|---|---|---|---|
| ALWAYS-RUN (proven) | 27 | 27 | 185.5s | 14.3% |
| **OPAQUE** (unmeasured) | 61 | 60 | **1010.2s** | **78.1%** |
| scoped (provably skippable) | 56 | 46 | 97.6s | 7.5% |
| total | **144** | **133** | 1293.3s | |

**Best-case ceiling for a perfect path→target map: 1.08x.** Estimate history: 3x → 1.7x → uncertain → 1.49x → **1.08x measured**. The handoff estimated rank 3 at "~1.7x alone, ~3.6x after"; neither is reachable.

**Timing cross-check:** 1293.3s visible + ~74s pytest hides ≈ 1367s, and the handoff independently quotes 1367s by another route.

## 🔴 The recommendation this produces: do not build rank 3 yet

Only **7.5%** of suite time is provably skippable. **78.1%** is OPAQUE — subprocess reads the tracer cannot see. A mapping built today buys ~8% and inherits 61 unmeasured files as chances to skip a test that should have run. Making the opaque children legible is the prerequisite, and it decides more than any directory split.

`test_drift_check.py` — the file the handoff named as the regex's false positive — is **OPAQUE**, not ALWAYS-RUN. "Unproven" is more honest than either prior label.

## Blind spots, stated

- **Reads inside a subprocess** — the whole OPAQUE bucket.
- 🔴 **Stat-only dependencies are invisible.** CPython raises no audit event for `os.stat`/`Path.exists()` (measured, 3.12.14). A guard depending only on `(REPO_ROOT/"x.sh").exists()` records an empty read set. The corpus contains this shape.
- **Symlinks not followed** (`_rel` avoids `resolve()` so the hook cannot re-enter).
- **A test depending on a file it never reads** — no read-tracer can see it.
- **One run** — single-observation read sets.

## This classifier reproduced the bug it exists to fix — seven times

Each matched *characters* or an *enumeration* rather than the operation: bare tool name (`git init /tmp/x`); a read verb anywhere in argv (a stub `bw … status`); the same off **script text** in `bash -c`; first non-flag token as git subcommand (`git -C scripts ls-files` → subcommand "scripts", acquitted); OPAQUE keyed on literal `-c`; argv[0] in `/nix/store` acquitting real scans; and cwd required to be exactly `.`.

A mutation sweep caught three and showed **three guards were unreachable**. An audit caught the rest on real corpus files. 🔴 **Then the fix round introduced its own regression** — a separator-terminated prefix made REPO_ROOT compare as outside the repo, acquitting `git -C <REPO_ROOT> ls-files`; ALWAYS-RUN fell 22 → 9, caught only by diffing the two classification runs.

## Tests

27 guards in `scripts/tests/test_readset_classify.py`, each pinning a defect these files actually shipped, each **watched red** against pre-fix code under `PYTHONDONTWRITEBYTECODE=1`. Two pre-existing tests had assertions **inverted** because they encoded behaviour the audit proved wrong.

https://claude.ai/code/session_01M5swnZSyv5Ae6LPBhJ2gPM

### DIFF
--- a/claudedocs/measurement-scripts-tests-readsets.md
+++ b/claudedocs/measurement-scripts-tests-readsets.md
@@ -56,11 +56,13 @@ bound on skippability until someone has tried to break it again.
 
 ⚠ The last step, 1.01x → 1.02x, is the one exception and it was a POLICY
 change, not a defect fix: round 3 deleted the operand-based acquittal (four
-defects came out of guessing scope from argv) and restored an unambiguous `-C`
-rule. That let three files whose only opacity was `git -C <tmp fixture repo>
-<unknown verb>` leave OPAQUE — a command with `-C` at an absolute outside path
-demonstrably operates on another tree. Precision gained, not safety traded; the
-verdict is unchanged either way.
+defects came out of guessing scope from argv) and kept only `-C`. That let
+three files whose only opacity was `git -C <tmp fixture repo> <unknown verb>`
+leave OPAQUE — **for a command whose grammar defines `-C`**, an absolute
+outside value does name another tree. That qualifier is load-bearing and was
+missing here for one round: stated unqualified, it is the exact generalisation
+defect 10 below disproved. Precision gained, not safety traded; the verdict is
+unchanged either way.
 
 ### 🔴 The recommendation this produces: DO NOT BUILD RANK 3
 
@@ -147,7 +149,7 @@ DEVRC_READSET_OUT=/tmp/rs PYTHONPATH=$DEVRC/scripts \
 python3 scripts/lib/readset_classify.py /tmp/rs.*.json --json /tmp/readsets.json
 ```
 
-## This classifier reproduced the bug it exists to fix — nine times
+## This classifier reproduced the bug it exists to fix — ten times
 
 The pattern is the deliverable's real lesson. Every one matched *characters* or
 an *enumeration* rather than the operation actually performed:
@@ -176,6 +178,14 @@ an *enumeration* rather than the operation actually performed:
    blind spot, re-introduced through the exec path. Neither that set nor
    `_GIT_WRITERS` had any test: mutants emptying `_HARMLESS` and removing
    `init` from `_GIT_WRITERS` both **survived a full green suite**.
+
+10. Consulted `-C` before knowing the command, so ANY argv merely containing
+    the token was acquitted — `bash -c '… git -C /tmp/fx rev-parse …'`, where
+    it sits in **script text**, and `bash run-tests.sh -C /tmp/outdir`, where
+    it is another program's flag. 🔴 **Found inside the single acquittal round
+    3 kept BECAUSE it was supposed to be unambiguous** — it is unambiguous only
+    for a command whose grammar defines it. All 38 guards passed with the hole
+    open.
 
 Findings 1, 2 and 4 came from a mutation sweep that also showed **three guards
 were unreachable** — an earlier acquittal always won, so deleting the guard
--- a/claudedocs/measurement-scripts-tests-readsets.md
+++ b/claudedocs/measurement-scripts-tests-readsets.md
@@ -1,0 +1,139 @@
+# Measured: what `scripts/tests` actually reads — and what decomposing it is worth
+
+**Date:** 2026-08-30 · **Effort:** ci-speedup rank 2 · **Claim:** `ci-speedup-2`
+
+## Why this exists
+
+`claudedocs/handoff-ci-speedup.md` ranks "decompose `scripts/tests`" second and
+attaches a 🔴 to its own input:
+
+> The repo-wide-scanner classifier is NOT trustworthy. A regex for
+> `git ls-files` / `REPO_ROOT.rglob` said 32 of 139 files (394s, 29%); positive
+> and negative controls passed, but it over-classifies … The true always-run set
+> is between ~8s and 394s and needs **empirical** measurement (change an
+> unrelated file, see what fails), not another regex.
+
+This is that measurement. It is not a regex and it is not a perturbation run —
+it is a **read-set trace**: what each test file actually opens, lists, scans and
+spawns, recorded by a `sys.addaudithook` while the suite runs.
+
+## The answer
+
+143 of 143 collectible test files measured (the other 4 files in the directory
+are `conftest.py` and three `mutation_battery_*`/`mutants-*` harnesses, none of
+them collectible tests). Timing measured in a **separate, untraced** run so the
+hook's overhead is not in the numbers.
+
+| bucket | files | seconds | share |
+|---|---|---|---|
+| **ALWAYS-RUN** — proven to read the tree | 22 | 163.1s | 12.6% |
+| **OPAQUE** — read set UNKNOWN | 18 | 704.3s | 54.5% |
+| **scoped** — proven bounded | 93 | 425.9s | 32.9% |
+
+- **The proven always-run set is 163.1s.** The handoff bracketed it at "between
+  ~8s and 394s"; it lands in the lower half, and far below the 394s the regex
+  implied.
+- **Best-case speedup from a perfect path→target mapping: 1.49x.** The handoff
+  estimated rank 3 at "~1.7x alone but ~3.6x after". 1.49x is the measured
+  ceiling *today*, and the 3.6x is not reachable while OPAQUE stands.
+- Estimate history is now **3x → 1.7x → uncertain → 1.49x measured.**
+
+### Cross-check on the timing
+
+Summed per-test durations come to **1293.3s** visible, plus ~74s of sub-5ms
+durations pytest hides (27,012 of them, capped at 0.005s each). That totals
+~1367s — and the handoff independently quotes **1367s of work** for this suite,
+derived by a different route. The timing instrument reproduces a number nobody
+fed it.
+
+## 🔴 The finding that reframes rank 3
+
+**The blocker is not file layout. It is subprocess opacity.**
+
+54.5% of the suite's time sits in 18 files that shell out to `bash -c` /
+`python -c` with a repo-root cwd. The audit hook is per-interpreter and cannot
+see a child process's reads, so those files' dependencies are genuinely
+unmeasured — and a path→target mapping must fail safe and run them all. Two
+files are most of it:
+
+| file | time | bucket |
+|---|---|---|
+| `test_subsystem_store_api.py` | 313.3s | OPAQUE |
+| `test_run_tests_floors.py` | 171.2s | OPAQUE |
+| `test_run_tests_preconditions.py` | 77.3s | OPAQUE |
+| `test_drift_check.py` | 62.7s | OPAQUE |
+
+Resolving the top two alone adjudicates ~37% of total suite time. That is a
+better next move than splitting directories, and it is cheap to state as a
+target: make those tests' repo reads visible (declare them, or run the child
+under a tracer), then re-run this measurement.
+
+**`test_drift_check.py` is the specific file the handoff named as the regex's
+false positive.** It is not ALWAYS-RUN here. It is OPAQUE — its repo-wide-ness
+is *unproven*, which is a different and more honest answer than either the
+regex's "repo-wide" or a bare "scoped".
+
+## Why OPAQUE is its own bucket
+
+Folding unknown into always-run would inflate a number that then reads as
+measured; folding it into scoped would be unsafe. RULES.md is explicit that
+UNMEASURED must not be folded into a clean count, so it is reported separately
+and a consumer must treat it as must-run.
+
+## The tools
+
+- `scripts/testlib/readset_plugin.py` — the tracer. Opt in with
+  `-p testlib.readset_plugin` and `DEVRC_READSET_OUT=<prefix>`; writes one JSON
+  shard per xdist worker. Attributes **both** collection-time (module-level) and
+  runtest-time reads, because several files in this corpus read `nix/home.nix`
+  at import.
+- `scripts/lib/readset_classify.py` — merges shards and assigns buckets.
+- `scripts/tests/test_readset_classify.py` — 15 guards, every one pinning a bug
+  this classifier actually shipped.
+
+Reproduce:
+
+```bash
+DEVRC_READSET_OUT=/tmp/rs PYTHONPATH=$DEVRC/scripts \
+  python3 -m pytest scripts/tests -q -p no:cacheprovider \
+  -p testlib.nolaunch_plugin -p testlib.spool_plugin \
+  -p testlib.gitenv_plugin -p testlib.nogit_plugin \
+  -p testlib.readset_plugin -n 4 --dist loadfile
+python3 scripts/lib/readset_classify.py /tmp/rs.*.json --json readsets.json
+```
+
+## What this cannot see — state these with any claim
+
+- **Reads inside a subprocess.** The whole OPAQUE bucket. Argv and cwd are
+  recorded; the child's own `open` calls are not.
+- **A test whose outcome depends on a file it never reads** — e.g. one asserting
+  a count someone else computed. No read-tracer can see that. Perturbation can,
+  and has the complementary blind spot (an innocuous edit does not trip a
+  scanner that only fails on violations). Neither method is sufficient alone.
+- **One run.** These are single-observation read sets; a test with a
+  data-dependent branch could read more on another run. The direction of that
+  error is under-classification, which is the unsafe one.
+
+## Building the classifier reproduced the very bug it exists to fix — four times
+
+Worth recording, because the pattern is the point. Each draft matched
+*characters* rather than *operations*, exactly as the original regex did:
+
+1. Matched the bare tool name → `git init -q /tmp/x` scored as a repo scan.
+2. Matched a read verb anywhere in argv → a stub binary invoked as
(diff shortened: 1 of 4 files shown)


ZacxDev:
## Gate status: cannot produce a green run on this box — and `main` is equally red

Three sandbox-tier runs (`nix build .#checks.x86_64-linux.pytests`, one at a time, never combined):

| run | tree | verdict | failing test | cause |
|---|---|---|---|---|
| 1 | this branch | FAIL | `test_live_cotenants_does_not_count_this_process` | unreaped `git` child of its own fixture |
| 2 | **`origin/main`, unmodified** | FAIL | `TestSkillDocsArePinned.test_the_pinned_docs_are_the_DEPLOYED_ones` | `SQLite database … is busy` |
| 3 | this branch | FAIL | same as run 2 | same — byte-identical assertion |

**Attribution: not this PR.**

- Run 2 is the control: `origin/main` with none of these changes fails the same way.
- Runs 2 and 3 fail on the *same* test with the *same* error, on trees differing only by this PR's four files.
- Run 1's test (`test_git_repo_isolation.py`) is not among those four files and passes 3/3 in isolation on this branch.
- The plugin added here is **inert in the gate**: it appears in no `conftest.py`, no `run-tests.sh`, no `home.nix`, and loads only via an explicit `-p`. This PR's entire runtime footprint is +15 tests in 0.11s.

`CLAUDE.md` already records `SQLite database is busy` evaluating `nix/home.nix` as a **false** failure from nix store contention. It documents that for one operator building two derivations concurrently; these runs were sequential, so the contention is coming from elsewhere on the box — which widens the rule: **any sandbox-tier RED under concurrent nix load is untrustworthy, not just a self-inflicted combined build.**

Run 1's failure is a genuinely different and pre-existing defect worth its own issue: the test builds a fixture repo with git subprocesses, then asserts no live git sits in it, and caught an unreaped child. Same class as the `test_timeout_reaps_the_whole_process_group` zombie problem already documented in this repo.

**I am not claiming this PR's gate is green.** I could not obtain a green run on any tree, including unmodified `main`.

ZacxDev:
## 🔴 CORRECTION — the numbers in the PR description above are WRONG

Posting rather than silently editing, since a reviewer may already have read them. The description has now also been updated.

An adversarial audit found 2 deploy-blocking and 6 should-fix defects in the classifier, several proven on **real corpus files**. Fixing them changed the headline result:

| | ALWAYS | OPAQUE | scoped | ceiling |
|---|---|---|---|---|
| ~~as described~~ | ~~22~~ | ~~18~~ | ~~93~~ | ~~1.49x~~ |
| **corrected** | **27** | **61** | **56** | **1.08x** |

**What changed, and why it matters:**

- **OPAQUE was keyed on the literal token `-c`.** The corpus's two *dominant* opacity shapes — a nested `python3 -m pytest <repo dir>` and `bash <repo script> <REPO_ROOT>` — matched neither branch and were published as "scoped, **proven bounded**". That is under-classification: a real test skipped when it should have run. Opacity is now the fall-through; only a recognised command can be scored clean.
- **`os.stat` was listed as a traced event and raises nothing.** CPython emits no audit event for `os.stat`/`Path.exists()` (measured, 3.12.14). It read as coverage while providing none — a guard depending only on `(REPO_ROOT/"x.sh").exists()` recorded an empty read set.
- **`scoped` was 103, not 93**, and the table's file counts silently came from the *timing* denominator (133 files with visible durations) under a claimed 143 classified. Both denominators are now printed side by side.
- Four more under-classifications: argv[0] in `/nix/store` acquitting real `git ls-files` scans; `cwd=scripts` acquitting a subdir scan; `_rel` matching **sibling worktrees** (`devrc-readsets` startswith `devrc` — ~45 such dirs on this box); `_rel` discarding relative paths; and self-scoping erasing every same-directory fixture.
- 🔴 **The fix round then introduced its own regression**, also unsafe: a separator-terminated prefix made `REPO_ROOT` itself compare as *outside* the repo, acquitting `git -C <REPO_ROOT> ls-files`. ALWAYS-RUN fell 22 → 9. No test caught it — diffing the two classification runs did.

## The conclusion inverts

**Do not build rank 3 yet.** Only **7.5%** of this suite's time is provably skippable; **78.1%** is OPAQUE. A path→target mapping on today's tree buys ~8%, and inherits 61 unmeasured files as chances to skip a test that should have run. The work that raises its ceiling is making the opaque children legible first.

The qualitative finding is unchanged and *stronger*: layout is not the blocker, subprocess opacity is.

**Ledger:** round 1 · payload lines changed this round: 227 (`readset_classify.py` 69, `readset_plugin.py` 48, the measurement doc 110 — the doc is the deliverable) · scaffolding 152 · 27 guards, each watched red against pre-fix code.

ZacxDev:
```audit-claims round=1 audited=44973f71..0a7335f6
1. OPAQUE is now the fall-through: any exec at a repo cwd that is not a recognised git read verb, a recognised direct scanner, a recognised git write verb, or a member of a small explicit harmless set, is classified OPAQUE rather than clean.
2. The `_OPAQUE_INTERPRETERS` set was deleted; opacity is no longer an enumeration of interpreter names, and the literal-token `-c` condition is gone.
3. `os.stat` was removed from `_PATH_EVENTS`; the stat blind spot is stated in the plugin module docstring and in the published measurement doc.
4. The argv[0]-outside-the-repo acquittal was removed, so a scanner or git reader invoked by absolute path (e.g. /nix/store/.../git) at a repo cwd is scored as a scan.
5. `_is_repo_cwd()` was added; a cwd that is a repo-relative subdirectory now counts as inside the repo, not as an acquittal.
6. `_rel()` compares against a separator-terminated repo prefix, so sibling directories such as `devrc-readsets` are no longer mapped to repo-relative paths.
7. `_rel()` joins non-absolute paths against the cwd instead of discarding them.
8. The operand acquittal treats a token equal to REPO_ROOT itself as inside the repo, so `git -C <REPO_ROOT> ls-files` is scored as a scan.
9. Self-scoping in `classify()` excludes only the test file's own path, not every path under its parent directory.
10. The published buckets changed from 22/18/93 to 27/61/56 and the stated ceiling from 1.49x to 1.08x; the measurement doc now prints the classified-file count (144) and the timed-file count (133) separately.
11. `scripts/tests/test_readset_classify.py` carries 27 guards; each new guard is claimed to have been watched red against the pre-fix code under PYTHONDONTWRITEBYTECODE=1.
12. Two pre-existing tests had their assertions inverted: an absolute-path scanner binary is now asserted to BE a scan, and a same-directory fixture is now asserted to BE a dependency.
```

**Ledger:** round 1 · payload lines changed this round: 227 (`readset_classify.py` 69, `readset_plugin.py` 48, `claudedocs/measurement-scripts-tests-readsets.md` 110 — the doc is this PR's deliverable, so it is payload) · scaffolding: 152 (`test_readset_classify.py`) · non-zero payload, so the attribution gate does not fire.

ZacxDev:
```audit-claims round=2 audited=0a7335f6..ed75ae91
1. The operand acquittal no longer runs as an early `continue`; it is computed as `outside_operand` and applied only inside the recognised `git`-reader and `_DIRECT_SCANNERS` arms, so an unrecognised head reaches the OPAQUE fall-through regardless of where its operands point.
2. `test` and `which` were removed from `_HARMLESS`.
3. `_HARMLESS` and `_GIT_WRITERS` now have guards; the two mutants round 2 reported as SURVIVED (emptying `_HARMLESS`, removing `init` from `_GIT_WRITERS`) are claimed to be killed.
4. `_rel()` restored `.lstrip("/")` on the repo-relative remainder, so a doubled separator no longer yields a leading-slash path.
5. `_SKIP_PARTS` is applied to the NORMALISED repo-relative path rather than the raw string, so relative `.git/...` reads are filtered again.
6. The plugin's `resolve()` comment no longer claims resolve re-enters the audit hook; it states the syscall cost instead, and the doc carries the same correction.
7. The published buckets changed from 27/61/56 to 27/79/38 and the stated ceiling from 1.08x to 1.01x; the doc's title, the opacity share (84.8%) and the defect list (now nine) were updated to match.
8. The doc now states that every correction so far moved files OUT of `scoped`, and that numbers from this tool should be treated as an UPPER bound on skippability.
9. `scripts/tests/test_readset_classify.py` carries 34 guards; each new round-2 guard is claimed to have been watched red against the round-1 code under PYTHONDONTWRITEBYTECODE=1.
```

**Ledger:** round 2 · payload lines changed this round: 103 (`readset_classify.py` 26, `readset_plugin.py` 15, the measurement doc 62 — the doc is this PR's deliverable) · scaffolding: 63 · non-zero payload, so the attribution gate does not fire.

**Round 2's own result, for the record:** it confirmed 10 of 12 round-1 claims outright, independently re-derived the 144/27/61/56 split exactly in its own clone, and independently re-measured that `os.stat`/`Path.exists()`/`Path.resolve()` raise no audit events on 3.12.14. Its 🔴 was that round 1's fix left the operand acquittal outranking the new fall-through — the same defect class round 1 was written to close, reaching the same files by another route.

ZacxDev:
```audit-claims round=3 audited=ed75ae91..30c54328
1. The operand-based acquittal was DELETED entirely; `outside_operand` no longer exists and neither the git-reader arm nor the `_DIRECT_SCANNERS` arm consults any operand test.
2. A new `_effective_cwd(toks, cwd)` resolves `-C` and is the only surviving acquittal: an absolute `-C` outside the repo returns `<outside-repo>`, an absolute `-C` into the repo returns ".", and a relative `-C` returns the caller's cwd unchanged.
3. `git ls-files --exclude-from /tmp/ex`, `find . -newer /tmp/stamp`, `grep -rf /tmp/patterns .`, `git grep -f /tmp/pats -- scripts` and `git log --oneline -- scripts /tmp/x` are now all scored as scans of this tree.
4. A recognised scanner with an outside operand and no `-C` is now deliberately OVER-classified as a scan; two pre-existing assertions were reversed to pin that policy.
5. `_HARMLESS` and `_GIT_WRITERS` are now pinned two-way by an exact frozenset comparison, so adding or removing any member fails the suite; the `stash` and `echo` mutants round 3 reported as SURVIVED are claimed killed.
6. `_rel()` calls `os.path.normpath` when the joined path contains "..", so a relative read escaping the repo is no longer recorded as an in-repo path.
7. The published buckets changed from 27/79/38 to 28/75/41 and the ceiling from 1.01x to 1.02x; the doc's title, opacity share (83.4%), skippable figure (21.4s) and guard count (38) were updated to match.
8. The doc's claim that every correction moved files OUT of `scoped` was corrected: it is now stated as true of the three DEFECT fixes only, with the 1.01x -> 1.02x step explicitly labelled a POLICY change in which three files whose only opacity was `git -C <tmp fixture repo> <unknown verb>` legitimately left OPAQUE.
9. `scripts/tests/test_readset_classify.py` carries 38 guards; each new round-3 guard is claimed to have been watched red against the round-2 code under PYTHONDONTWRITEBYTECODE=1.
10. One test was renamed because its name credited the wrong acquittal mechanism twice, in opposite directions; it now names no mechanism.
```

**Ledger:** round 3 · payload lines changed this round: 83 (`readset_classify.py` 46, `readset_plugin.py` 8, the measurement doc 29) · scaffolding: 79 · non-zero payload, so the attribution gate does not fire.

**Round 3's result, for the record:** it found **no regression** — the first fix round that did not introduce one — confirmed all five round-2 code claims, re-derived 144/27/79/38 exactly in its own clone, and independently re-measured the audit-event behaviour of `os.stat`/`Path.resolve`. Its 🔴 was the fourth occurrence of one defect class: inferring a command's scope from its operands. That inference is now deleted rather than patched again.

ZacxDev:
```audit-claims round=4 audited=30c54328..8934395b
1. `_effective_cwd()` takes a `head` parameter and returns the caller's cwd unchanged unless `head == "git"` or `head in _DIRECT_SCANNERS`, so a `-C` token is only consulted for a command whose grammar defines one.
2. `bash -c '… git -C /tmp/fx rev-parse …'` is now OPAQUE rather than clean — a `-C` appearing inside interpreted script text no longer acquits.
3. `bash scripts/run-tests.sh -C /tmp/outdir`, `python3 -m pytest scripts/tests -C /tmp/out` and `node scripts/x.mjs -C /tmp/out` are now OPAQUE rather than clean — a `-C` that is another program's unrelated flag no longer acquits.
4. The classification buckets are unchanged by this fix: 28 ALWAYS-RUN / 75 OPAQUE / 41 scoped, ceiling 1.02x, identical to the round-3 result.
5. `scripts/tests/test_readset_classify.py` carries 40 guards; the two new ones are claimed to be the only tests that fail when the head gate is removed, and to have been watched red against the round-3 code under PYTHONDONTWRITEBYTECODE=1.
6. The doc's guard count was updated from 38 to 40; no other published figure changed.
```

**Ledger:** round 4 · payload lines changed this round: 16 (`readset_classify.py` 15, the measurement doc 1) · scaffolding: 26 · non-zero payload, so the attribution gate does not fire.

**Round 4's result, for the record:** it re-derived **every** published number from the raw shards and got an exact match — the first round able to reproduce the whole table from the original inputs — and verified per file that the three OPAQUE→scoped movers were each `git -C <tmp fixture repo> {rev-parse,cat-file}` and nothing else. Its 🔴 was the **fifth** occurrence of this PR's one defect class, surviving inside the single acquittal round 3 kept *because* it was supposed to be unambiguous. Latent, not live: the corpus holds one non-git exec with a bare `-C` and its target is an unexpanded `"$SR"`. **All 38 tests passed with the hole open.**

ZacxDev:
## Audit ladder CLOSED after 5 rounds — round 5 reports the code delta clean

**Round 5 verdict: the code delta is clean.** It re-derived the buckets from the raw shards under both round-3 and round-4 code and found them identical **at exec granularity — 0 of 20,064 execs changed verdict**; mutation-proved both new guards fail on their own assertions; and probed every sixth-occurrence candidate, including the one I specifically suspected in my own fix. **`grep -C 3` does not fire** — the context count is not absolute, so `_effective_cwd` returns the cwd and it scores SCAN. `-C` as final token, `-C=value`, `--git-dir`/`--work-tree`, and head-spoofing via `/tmp` binaries all resolve safely.

Its single finding was documentation, and it was the shape this PR's own defect entry #9 criticises: the ledger still said "nine" and never gained round 4's defect, and one sentence asserted flatly that "a command with `-C` at an absolute outside path demonstrably operates on another tree" — **the exact generalisation round 4 disproved, sitting one screen from the entry disproving it.** Fixed in `6ae15f68`: entry 10 added, claim qualified with *for a command whose grammar defines `-C`*.

```audit-claims round=5 audited=8934395b..6ae15f68
1. The defect-ledger heading was changed from nine to ten and an entry 10 was added describing `_effective_cwd` consulting `-C` before the head was known.
2. The estimate-history note's `-C` claim was qualified with "for a command whose grammar defines `-C`", and the paragraph now states that the unqualified form is the generalisation defect 10 disproved.
3. No code file changed in this commit and no published figure moved; the 40 guards remain green.
```

**Ledger:** round 5 · payload lines changed this round: 16, all in the measurement doc's *historical record of this ladder* · scaffolding: 0 · code payload: **0**.

### Why this ends here

The stop rule ends the ladder on a clean round, and the code delta is clean. This last commit changes only the ladder's own account of itself, so a round 6 would be the ladder auditing what the ladder just wrote — the waste the attribution gate exists to stop. Recording that as a judgment call, not a claim that the rule's letter compelled it.

### What five rounds cost and bought

| round | outcome |
|---|---|
| 1 | 2 🔴 + 6 🟡 — headline was wrong (1.49x → 1.08x); `os.stat` traced nothing; `scoped` was 103 not 93 |
| 2 | 🔴 same class, other route — 18,806 execs acquitted across 68 files (1.08x → 1.01x) |
| 3 | no regression; 🔴 4th occurrence — option values read as operands. Deleted the inference (1.01x → 1.02x) |
| 4 | 🔴 5th occurrence, inside the acquittal kept for being unambiguous. Buckets bit-identical |
| 5 | **code clean**; ledger one entry short |

Every round found something real. Three of five fix rounds introduced their own regression. **The measurement's conclusion never wavered and only strengthened:** subprocess opacity is the whole story, and rank 3 should not be built.

PR #1130, opened 2026-08-30
fix(session-trailer): a single-paragraph message got a stamp git does not read as a trailer

ZacxDev:
Closes the remaining defect on clawgate cg#365. The session-id trailer is ARMED and live on
both hosts, so this was producing unreadable stamps in production, not hypothetically.

## The defect

On a **single-paragraph** commit message the stamp was appended with no blank line, so git
did not parse it as a trailer at all — the id was present as text and invisible to every
standard trailer consumer:

```
$ git commit -m "test: does a real commit carry the trailer"
test: does a real commit carry the trailer
Claude-Session-Id: 8cdbb099-…          <-- no blank line

$ git log -1 --format='%(trailers:key=Claude-Session-Id,valueonly)'
                                        <-- EMPTY
```

That is exactly cg#365's closing condition ("git blame a line → resolve the session → wake
it") failing, and failing silently.

## Root cause — the old code tried, and its test was wrong

It judged the separator from the **last line**:

```python
tail = _lines(body)[-1]
is_trailer_tail = bool(re.match(r"^[A-Za-z][A-Za-z0-9-]*:\s", tail))
```

A conventional-commit subject matches that regex. `fix: one line only` looked like a
trailer block, so the stamp was glued on with no separator.

## The fix

The separator is now decided by the **last paragraph**, matching what git actually does:

```python
sep = "\n" if _ends_in_trailer_block(body) else "\n\n"
```

A paragraph is a trailer block only if it is not the subject paragraph *and* every line in
it is a trailer line or an indented continuation. The rule was derived by probing
`git interpret-trailers` 2.55.0 across 19 shapes rather than inferred from docs. Where it
differs it is stricter than git — being stricter only ever *adds* a blank line, which still
yields a paragraph git parses.

> ⚠ **Corrected after audit round 1 (see the PR comment for the measurements).** An earlier
> revision of this line called the divergence "deliberately conservative (git accepts
> `Key:value` with no space; this does not)". That named **one** class where there are
> **five**, and omitted the largest — `Key:` with an empty value. Over a 1688-message
> differential fuzz: 209 shapes where this loses a sibling trailer that base kept, vs 29 the
> other way. **Real-world impact over the last 600 `origin/main` messages is zero** (devrc's
> trailer paragraphs are uniformly well-formed), and the change still takes "git cannot parse
> our stamp" from 6 to 2 there — but the original wording read as coverage it did not have.

## Red at base, green at HEAD

`single-line`: **RED at `a5bc5df6`, GREEN at HEAD**, verified on a real commit object with
everything else held identical:

```
base   'test: … trailer\nClaude-Session-Id: SID-42\n'    -> %(trailers:…) = (empty)
head   'test: … trailer\n\nClaude-Session-Id: SID-42\n'  -> %(trailers:…) = SID-42
```

Four shapes were red at base, not one: `single-line`, `prose-then-trailerish`,
`trailerish-second-line`, `subject-and-blank-only`. Six more are invariant guards (labelled
as such in the test file — they pin behaviour the bug never violated and are **not**
regression coverage).

**A second regression this closes:** on `continuation-line`
(`Co-Authored-By: A\n    <a@b.c>`) the old last-line test read the indented continuation as
prose and split the block, **losing the sibling `Co-Authored-By` trailer**. It survives now.

## Verification, done independently of the implementation

- The oracle is `git interpret-trailers --parse`, carrying **both controls as committed
  tests**: a negative control fed the exact broken base output (must parse to `''`), and a
  positive control fed a known-good message. Without them a `--parse` wired to nothing
  would make every assertion vacuous. Subprocesses run under a hermetic git env so a stray
  global `trailer.*` config cannot move the answer.
- **Mutation sweep, re-run independently rather than taken on report**, under
  `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` cleared between mutants:

  | mutant | result |
  |---|---|
  | always `"\n\n"` (the plausible wrong fix — splits an existing trailer block) | KILLED — sibling-trailer test |
  | revert to the old last-line check (the original bug) | KILLED — `test_a_single_paragraph_message_gains_a_blank_line` |
  | drop the whole-paragraph check | KILLED — `test_a_paragraph_that_only_STARTS_as_trailers_is_not_a_block` |

  Each dies by a **specifically named test**, not merely "a test failed".

## Existing tests

Nothing modified or deleted. None encoded the buggy output — they assert the trailer *text*
is present, true in every shape, which is precisely why the suite passed through this defect
for its whole life.

## Known-open, deliberately not fixed here

- **`append_trailer("")` has the same shape and is left as is**: it returns a one-line
  message git also does not parse as a trailer. Reachable only via
  `commit --allow-empty-message`, so widening the change into it was not worth the risk.
  Measured and recorded in the commit message; not covered by a test.

## Not run

Focused dev-host suites only (189 passed across the trailer/hook/hermetic-git tests, plus
136 across the repo-wide scanners). **The `nix build .#checks…` sandbox tier — the one the
merge actually gates on — was not run here**, and CLAUDE.md documents that the two tiers
legitimately disagree. Tekton will be the first run of that tier on this branch.

### DIFF
--- a/scripts/lib/session_trailer.py
+++ b/scripts/lib/session_trailer.py
@@ -404,6 +404,18 @@ def _ends_in_trailer_block(body: str) -> bool:
     if start == 0:
         return False                 # the last paragraph IS the subject's
     para = lines[start:]
+    if not para:
+        # 🔴 `body` is `rstrip("\n")`, which strips NEWLINES and not a final line
+        # of spaces or tabs. Such a line is falsy under `.strip()`, so the scan
+        # above never moves, `start == len(lines)`, and this slice is EMPTY —
+        # `para[0]` then raised IndexError. Measured: `"fix: v\n\nbody\n   \n"`
+        # stamped fine BEFORE this predicate existed and raised after, so it is
+        # a STRICT regression, and the hook's broad `except` swallowed it into a
+        # commit that succeeded carrying no trailer, with nothing logged.
+        # Reachable via `--cleanup=verbatim` and from any direct library caller;
+        # git's own cleanup strips such lines before the hook on the `-m` path.
+        # A trailing blank paragraph is not a trailer block, so: do not join.
+        return False
     if not _TRAILER_LINE_RE.match(para[0]):
         return False                 # a block cannot open on a continuation
     return all(_TRAILER_LINE_RE.match(text)
--- a/scripts/lib/session_trailer.py
+++ b/scripts/lib/session_trailer.py
@@ -362,6 +362,54 @@ def has_trailer(message: str, key: str = TRAILER_KEY) -> bool:
     return bool(_trailer_lines(message, key))
 
 
+# `Key: value` at column 0 — the shape of a trailer LINE. Not the shape of a
+# trailer BLOCK, which is what `_ends_in_trailer_block` decides.
+_TRAILER_LINE_RE = re.compile(r"^[A-Za-z][A-Za-z0-9-]*:\s")
+# git folds an indented line into the PREVIOUS trailer's value, so it belongs to
+# the block rather than ending it. Measured: `Co-Authored-By: A\n    <a@b.c>` is
+# one trailer to `git interpret-trailers --parse`, not a trailer plus prose.
+_TRAILER_CONTINUATION_RE = re.compile(r"^[ \t]")
+
+
+def _ends_in_trailer_block(body: str) -> bool:
+    """Would git let a new trailer JOIN `body`'s last paragraph?
+
+    🔴 THE TAIL LINE ALONE CANNOT ANSWER THIS, and asking it that way is the bug
+    this replaces. The old test was `re.match(r"^[A-Za-z][A-Za-z0-9-]*:\\s", tail)`
+    on the LAST LINE only — which is TRUE of a conventional-commit SUBJECT
+    (`fix: one line only`). A single-paragraph message therefore got its trailer
+    glued straight onto line 2 with no blank line, and git then read the whole
+    thing as one subject paragraph: `git interpret-trailers --parse` returned
+    NOTHING, and a real commit's `%(trailers:key=Claude-Session-Id)` was EMPTY.
+    Whole commits were silently unresolvable — the failure the trailer exists to
+    prevent.
+
+    git's actual rule, measured against `git interpret-trailers` 2.55.0 rather
+    than inferred, is a property of the LAST PARAGRAPH:
+
+      * it must not BE the first paragraph — git never reads the subject
+        paragraph as trailers, so `fix: x\\nNote: y\\n` gets a blank line;
+      * every line in it must be a trailer line (or an indented continuation),
+        so `Some prose here.\\nNote: something` gets a blank line too.
+
+    Conservative where it differs: git also accepts `Key:value` with no space,
+    and we do not. Being stricter only ever ADDS the blank line, which still
+    yields a paragraph git parses as trailers — it can cost a sibling trailer
+    its block, never cost us our own.
+    """
+    lines = _lines(body)
+    start = len(lines)
+    while start > 0 and lines[start - 1].strip():
+        start -= 1
+    if start == 0:
+        return False                 # the last paragraph IS the subject's
+    para = lines[start:]
+    if not _TRAILER_LINE_RE.match(para[0]):
+        return False                 # a block cannot open on a continuation
+    return all(_TRAILER_LINE_RE.match(text)
+               or _TRAILER_CONTINUATION_RE.match(text) for text in para[1:])
+
+
 def append_trailer(message: str, session_id: str, key: str = TRAILER_KEY) -> str:
     """Return `message` carrying exactly one `key: session_id` trailer.
 
@@ -404,8 +452,6 @@ def append_trailer(message: str, session_id: str, key: str = TRAILER_KEY) -> str
     if not body:
         return line + "\n"
     # A trailer is separated from prose by a blank line, but must not gain one
-    # when the message already ends in a trailer block.
-    tail = _lines(body)[-1]
-    is_trailer_tail = bool(re.match(r"^[A-Za-z][A-Za-z0-9-]*:\s", tail))
-    sep = "\n" if is_trailer_tail else "\n\n"
+    # when the message already ends in a trailer block git would let it join.
+    sep = "\n" if _ends_in_trailer_block(body) else "\n\n"
     return body + sep + line + "\n"
--- a/scripts/tests/test_session_trailer.py
+++ b/scripts/tests/test_session_trailer.py
@@ -35,12 +35,18 @@ from __future__ import annotations
 import importlib.util
 import json
 import os
+import subprocess
+import sys
 from pathlib import Path
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


ZacxDev:
## Correcting this PR's own description — audit round 1

Posting rather than editing the body silently, because a reviewer may already have read the
wrong version.

### 🔴 The body's "deliberately conservative" claim was too narrow

The body says:

> *Where it differs it is deliberately conservative (git accepts `Key:value` with no space;
> this does not)*

That presents **one** divergence class where there are **five**, and it omits the largest.
Measured by the auditor against `git interpret-trailers` 2.55.0 over a 1688-message
differential fuzz, base vs HEAD, git as oracle:

| line HEAD rejects that git accepts | fuzz regressions |
|---|---|
| `Key:` with an **empty value** (`Note:`) | 79 |
| `Key:value`, no space — *the only one the body named* | 39 |
| `#` comment line inside the block | 35 |
| `* bullet` / prose beside a git-generated trailer (`Signed-off-by:`) | 34 |
| key outside `[A-Za-z0-9-]` (`Under_score:`) | 17 |

Net across the fuzz: **209 shapes where HEAD loses a sibling trailer that base kept**, vs 29
where HEAD keeps one base lost. git's real rule is also wider than the docstring says — a
paragraph qualifies when it contains a git-generated trailer even at 50% non-trailer lines.

**Real-world impact measured separately: zero.** Over the last 600 `origin/main` commit
messages, neither version loses a sibling trailer and neither crashes — devrc's trailer
paragraphs are uniformly well-formed `Key: value`. And over those same 600, "git does not
parse our stamp" falls **6 → 2** with this change. So the direction is still a clear
improvement; the body just overstated how narrow the divergence is. The function's own
docstring was honest about the trade ("it can cost a sibling trailer its block"); the PR body
was not.

### 🔴 A real regression the body did not know about — now fixed in `10a5dc0a`

A whitespace-only tail line raised **`IndexError`**. `body` is `rstrip("\n")`, which strips
newlines but *not* a final line of spaces or tabs; such a line is falsy under `.strip()`, so
the paragraph scan never moved and the slice was empty.

```
"fix: v\n\nbody\n   \n"           base a5bc5df6 -> stamped        HEAD a2029da0 -> IndexError
"subject\n\nCo-Authored-By: A\n  \n"  base -> stamped             HEAD -> IndexError
"   \n"                               base -> stamped             HEAD -> IndexError
```

A strict regression, and a **silent** one: `prepare_commit_msg.py` calls `append_trailer`
outside any `try`, so the broad `except Exception: sys.exit(0)` and the wrapper's
`2>/dev/null` turned the crash into a commit that succeeded carrying no trailer, with nothing
logged — contradicting the module header's *"FAILS OPEN, ALWAYS … degrades to 'no trailer'"*.
It degraded by crashing, which is a different thing.

Reachable via `--cleanup=verbatim` and from any direct library caller; git's own cleanup
strips such lines before the hook on the `-m` path, so the mainline agent path could not hit
it. Fixed with a one-line guard plus five tests; all four shapes are now byte-identical to
base output. Removing only that guard kills exactly those five tests (83 → 78), each by the
guard's own `IndexError` at `:407`.

### ✅ Correcting the body's "Not run" section

The body said the `nix build .#checks…` sandbox tier — the tier the merge actually gates on —
had not been run. **It has now, one at a time, at `a2029da0`, and both are green:**

```
pytests   -> /nix/store/mhgwfl4m99c44zllp4c3ffjbdl58zssk-devrc-pytests
nodetests -> /nix/store/7dy7nknbwlxmisfj971b8m5dz1af01iq-devrc-nodetests
```

That was at `a2029da0`; `10a5dc0a` has not been through that tier yet.

### Still open, deliberately

- **`append_trailer("")`** — unchanged, still flagged known-open.
- **The five divergence classes** are *not* being fixed here. The cheapest partial fix (drop
  `\s` from `_TRAILER_LINE_RE`) covers the top two but makes a bare `https://…` line read as a
  trailer opener, so it needs its own oracle work. Real-world impact is zero today.
- **🟢 The test oracle's hermetic env pins global/system git config but not repo-local**, so a
  repo-local `trailer.*` key could move it. devrc has none today; it would surface as a
  dev-host/sandbox tier disagreement. Worth closing separately.

ZacxDev:
Audit round 1 claims, for round 2 to check.

```audit-claims round=1 audited=a2029da0..10a5dc0a
1. 🟡-1 whitespace-only tail IndexError: claimed FIXED — a `if not para: return False` guard in `_ends_in_trailer_block` (session_trailer.py), plus 5 tests in `TestAWhitespaceOnlyTailDoesNotRaise`.
2. 🟡-1: claimed that all four whitespace-tail shapes now produce output byte-identical to base a5bc5df6.
3. 🟡-1: claimed mutation-isolated — removing only that guard kills exactly those 5 tests (83->78) and nothing else.
4. 🟡-2 PR body overstating the divergence classes: claimed corrected in PR comment 5471864821. NO code change; `_TRAILER_LINE_RE` is unchanged and the five divergence classes remain.
5. 🟢-3 (`---` patch-start line): NOT addressed, no code or docstring change.
6. 🟢-4 (oracle hermetic env does not pin repo-local `trailer.*` config): NOT addressed.
7. Claimed the sandbox `checks` tiers were green at a2029da0; NOT re-run at 10a5dc0a.
```

ZacxDev:
## Merged-tree gate — run by hand, because `strict: false`

A green check is a claim about the PR's **branch**, not about the tree the merge creates. So I built an integration tree in a **real clone** (never a worktree — `nix build path:<worktree>` breaks every git call in the sandbox) and ran both merge-gating tiers on it, **one at a time** (a combined invocation makes nested-`nix` tests contend on the store and produces false failures).

**Integration commit `f8500b10` = `origin/main c06a56a1` + `fix/trailer-separator 96d5c315`**, clean merge, both sides verified present in the tree.

```
checks.x86_64-linux.pytests   -> /nix/store/prdbqlp9203nkkhpk1xwbqlhzcfv4nw1-devrc-pytests    GREEN
checks.x86_64-linux.nodetests -> /nix/store/3iqdfnr9lk56w12a3yncfhp3payhn5lh-devrc-nodetests  GREEN
```

Verdict taken from the produced out-paths and rc 0, not a piped exit code.

### Scope of that claim, stated rather than covered

`main` has since advanced 2 commits (`e0f35ce2`, `5324bf47`) — audit-pr skill docs, `audit-dispatch.py` and its tests. **Disjoint from every file this PR touches.** I checked the one way disjoint changes can still collide — a shared counter — and there is none: both sides only *add* tests, and `TARGET_FLOORS` are minimums, so more tests moves further above the floor rather than crossing it; the byte-ceiling hazard needs both sides contributing to the same doc counter, and this PR touches no skill or doc.

I did not rebuild on `5324bf47`. Chasing a base that moved 11+ times in one session is the treadmill `strict: false` exists to avoid. The gap is named above.

### Ladder summary

Round 1: 2 🟡 (one a real `IndexError` regression, fixed in `10a5dc0a`), 2 🟢. Round 2: no payload defect — verified the guard across 10,974 messages with zero divergence from base — 2 🟢 test-comment prose, fixed in `96d5c315`. Round 2's fixes changed **0 payload lines**, so a round 3 would audit the ladder's own scaffolding. Stopping here is a judgment call, not a clean round.

PR #1132, opened 2026-08-30
fix(index-store): the store HAS an off-machine backup — 15 places said it did not

ZacxDev:
## What

`~/.claude/analyze-service-index/` genuinely had no backup until 2026-08-21. `analyze-service-index-backup.service` landed that day and has run daily since — **measured today: workbench 04:30 and laptop 04:58, both `ExecMainStatus=0`, 15 scopes each, every bundle `(verified; pruned 0)`** to the homelab MinIO `minio-archive` tenant. Not one of the **fifteen** sentences asserting the absence was updated.

## Why this is not cosmetic

It is a **"this is impossible" claim sitting in a RECOVERY path**, inside 🔴 blocks. The failure it produces is silent and total: an agent facing a destroyed store reads *"irreplaceable, no off-machine backup"*, concludes recovery is unavailable, and never looks for `restore-verify.py` or the bundles that would have restored it.

A stale comment that hides a **capability** costs more than one that hides a hazard — nobody seeks a second opinion about an impossibility.

## The safety rules are unchanged

None is weakened. They rest on **confidentiality and curation**, never on backup-absence, so each site keeps its conclusion and only its premise is corrected. Two get *sharper*, because the backup's existence is what makes the real exposure legible:

- `reset --hard` / `clean` / `checkout --` destroy **uncommitted** content — exactly the window no hourly commit and no daily bundle holds.
- `seed.sh`'s other copies are **lagging derivatives**, so a write that corrupted the source would be replicated outward, not repaired from them.

## Sites corrected — 15 across 13 files

Three were found only after **widening the scan**: the bare `no backup` / `THE ONLY COPY` phrasings are invisible to a pattern written for `no off-machine backup`, and four more were **line-wrapped across the phrase**, so a line-based grep could not see them at all.

| | |
|---|---|
| docs | `analyze-service/SKILL.md`, `analyze-service/reference/index-store.md` (canonical), `prune-index/SKILL.md` |
| code | `lib/subsystem_recall.py`, `lib/subsystem_touch.py`, `subsystem-audit.py`, `subsystem-store-api/seed.sh` |
| tests | `test_service_recon.py`, `test_subsystem_audit.py` (×2), `test_subsystem_audit_budget.py`, `test_subsystem_recall.py`, `test_subsystem_store_api.py`, `test_subsystem_touch.py` (×2) |

All test-file mentions were **prose** — a docstring, or in one case an assertion *message*. **No assertion changed and no test needed a new expectation.**

## New guard

`scripts/tests/test_index_store_backup_claim.py` pins the doc claim against the nix wiring: `A == B`, where

- **A** = `home.nix` declares the unit + timer + an `ExecStart` naming a `backup.py` that exists on disk
- **B** = the Store-safety section names the unit, MinIO, the bucket, and `restore-verify.py`

A **phrase ban was rejected deliberately** — it is the *"spelled rather than structural"* guard `RULES.md` warns about, walkable by any reword, and it would fire on the ⚠ retraction sentences that legitimately quote the old wording.

## Verification

- **RED at base `ac64ccb4`** with this guard's own message (*"wires … but the 'Store safety' section does not say so in actionable terms"*); **GREEN at HEAD**.
- The opposite branch proven **reachable**, not merely present: with the corrected doc in place, renaming the nix unit fails with the **other** message (*"claims the store is backed up … but nix/home.nix no longer wires it"*).
- Both **controls on the detector itself** pass in both trees — negative (the pre-2026-08-21 paragraph, *and* a vague "backed up daily" naming no restore path, must NOT satisfy it) and positive (a rewrapped, re-cased minimal claim must).
- **2,660 tests green** across the 11 affected files (571 + 2,089, `nix develop` pytest, two batches).
- Full `scripts/gate.sh` running at time of opening — verdict posted as a comment.

## Scope this does NOT cover

Stated so nobody reads coverage into it: the guard never opens the live store, reaches MinIO, or runs systemd. It **cannot** tell you a backup *ran* or that a bundle is *restorable* — only that the repo's claim and the repo's wiring agree. Per-host liveness stays a hand check: `systemctl --user list-timers 'analyze-service-index*'`.

https://claude.ai/code/session_011G1qW4WiqHPNjLtbE3nsbe


ZacxDev:
## Audit round 1 — all six findings claimed fixed

Fix round landed as `804f86f2`, then `origin/main` merged in as `566c42bf` (zero file overlap with the 14 commits that landed on main since `ac64ccb4`).

**Gate:** dev-host tier PASS on both suites at `ac64ccb4` before the fix round; re-running on the merged tree now.

```audit-claims round=1 audited=54b7b593..566c42bf
1. The 19 surviving sites of the false claim in 11 files are corrected — including the `unbacked-up` spelling that the round-0 scan never covered, `prune-index/SKILL.md`'s section heading, and `subsystem-audit.py`'s live RuntimeError message.
2. `writing-and-safety.md`'s "Nothing in this store leaves the machine" is corrected as a CONFIDENTIALITY claim, and the no-remote rule is re-justified on encryption/ownership/audit rather than on nothing leaving the box.
3. `index-store.md`'s reset bullet no longer claims only uncommitted content is at risk: `git reset --hard <ref>` is stated to orphan COMMITTED content that `git bundle create --all` will not reach, in both `index-store.md` and `writing-and-safety.md`.
4. The doc claim is pinned as a WHOLE NORMALISED STRING instead of four tokens, so the auditor's meaning-reversed section no longer passes; that reversed text is now a regression test.
5. The decommission direction has a canonical CLAIM_ABSENT paragraph printed by the failure message, so the gate is satisfiable in both states; a test asserts the two canonical claims cannot satisfy each other's check.
6. `_nix_wires_a_backup` strips comments, requires `WantedBy = [ "timers.target" ]`, requires an ExecStart naming a backup.py that exists, and rejects a `lib.mkIf`-gated unit — consolidated into `scripts/testlib/nix_units.py` rather than reimplemented beside the existing reader in `test_analyze_service_index_backup.py`.
7. The `OnCalendar=*-*-* *:00:00` literal is corrected to `OnCalendar = "hourly"` with `RandomizedDelaySec = 600`.
8. The four-token failure message that advertised four requirements while enforcing three is gone with the token list.
```

**Not fixed, deliberately:** `backup.py:51` still reports that the scope READMEs say the content never leaves the machine — it is quoting those READMEs to explain why age encryption exists, and they live in the store, not in this repo.

**Carried forward, not closed:** the sandbox tier (`nix build .#checks…`) was not exercised by round 1's auditor — another session was building the same derivation concurrently, which produces documented false failures — and the laptop's backup run was not re-measured from that worktree.

ZacxDev:
## Audit round 2 — all 11 findings addressed

Fix round landed as `4038d8a8`. Two of round 2's findings (F2, F4) were round 1's fixes reproducing this PR's own thesis in fresh spellings; both are corrected.

**Payload ledger, round 2:** `54b7b593..566c42bf` was +382/−133 across 17 files; 63 payload lines, the rest scaffolding (`nix_units.py` +94, six test files). Above zero, so the ladder is on the PR rather than on itself.

```audit-claims round=2 audited=566c42bf..4038d8a8
1. F2 — `CLAIM_ABSENT` no longer asserts "nothing leaves this machine" or "not at all if the disk is gone"; it states that retiring the unit stops NEW bundles while retained bundles (ASIB_KEEP) and the served pod's copy still exist, and says to check both before calling anything unrecoverable.
2. F4 — "bare" is scoped to `git reset --hard` alone in both `index-store.md` and `writing-and-safety.md`, with an explicit note that `git checkout --` always takes a pathspec and `git clean` is almost always `-fd`.
3. F3 — the reflog sentence is bounded by the retention window: orphaned commits reach no FUTURE bundle but are still inside bundles already in the bucket, with `restore-verify.py` named; same overstatement corrected in `run-tests.sh`.
4. F1 — `flake.nix:347`'s "the only copy" is corrected to match the `run-tests.sh` twin, with a note recording that the two files disagreed.
5. F5 — `WantedBy` and `ExecStart` are read by DIRECTIVE NAME via a new `nix_units.directive()`; both mutants (`After = [ "timers.target" ]`, ExecStart line deleted while `X-Restart-Triggers` keeps the path) are in the sweep, each with a control asserting the substring a naive check would match is still present.
6. F6 — `is_conditional` spans `=` to the body's opening brace, so a `lib.mkIf` on the `let … in {` service declaration is visible; the sweep now gates both declaration shapes.
7. F7 — the consolidation actually happened: `test_analyze_service_index_backup.py` imports the shared `strip_nix_comments` and its local copy is gone.
8. F8 — `strip_nix_comments` strips trailing comments too, with double-quote parity so a `#` inside a nix string is not treated as one.
9. F9 — retraction lines are dropped before matching, so a quoted-and-denied claim no longer satisfies the pin; a control asserts the filter does not eat the live claim beside one.
10. F11 — `_normalise` strips `_` only at a word boundary, so `__bold__` normalises while `ASIB_KEEP` survives.
11. F10 — `proposal-entry-shape-explicit.md` keeps its "unbacked-up store" reasoning verbatim and carries a dated superseded-fact note above it.
```

**Not changed:** round 2 counts 18 sites across 14 files where round 1's message said 19 across 11. The discrepancy runs toward the fix being *wider* than advertised, so nothing a reader does changes.

**Still open:** the sandbox tier (`nix build .#checks…`) is unexercised for a **third** round. It is building now, sequentially, after waiting out another session's concurrent build of the same derivation — result will be posted here.

ZacxDev:
## Audit round 3 — all six findings addressed

Fix round landed as `e91604ae`. Two of round 3's findings are round-2 *guards* that were measured not to pin what their comments claimed, and one is the decommission paragraph being wrong for a second round running.

**Payload ledger, round 3:** `566c42bf..4038d8a8` was 33 payload lines (111 since round 1), 229 scaffolding. Above zero — the ladder is still on the PR.

Every fix below was **measured**, not reasoned:

| check | result |
|---|---|
| F6 mutant now discriminates | `old=False new=True` — where round 2's was `old=True new=True` (**vacuous**) |
| `is_conditional` over-reach | 41 top-level units, **0** disagreements; live service reads unconditional |
| wedged retraction | delete-filter `True` → sentinel `False` (**green before, red after**) |
| `directive()` multi-line | raises `NotImplementedError`; still reads live `ExecStart` + `WantedBy` |
| `CLAIM_ABSENT` | free of all four refuted phrasings |

```audit-claims round=3 audited=4038d8a8..e91604ae
1. 🟡-1 — CLAIM_ABSENT no longer asserts a retention window: it states that pruning runs only as part of an upload, so retiring the unit retires the pruner and the last ASIB_KEEP bundles per scope stay until deleted by hand; it names restore-verify.py and says to delete deliberately when the data should be gone.
2. 🟡-2 — flake.nix now carries BOTH halves its twin has (the AUTHORITATIVE-copy correction and the "a bundle already in the bucket still holds what was reachable" qualifier), and its note no longer claims in the past tense that the divergence was closed; it says the two comments have drifted twice and must be changed together.
3. 🟡-3 — the is_conditional mutant is built on the line AFTER the `=`, where nix style puts a gate for the `let … in {` shape, with a control asserting the `=` line still ends at the `=` so a shape change rebuilds the mutant rather than voiding it.
4. 🟡-4 — retraction lines are replaced by a sentinel instead of deleted, so a retraction wedged INTO the claim no longer stitches back together; the wedge is a regression test, and the docstring states that the marker set is a spelling rather than implying the hazard is closed.
5. 🟡-5 — directive() raises NotImplementedError naming the line rather than returning a truncated multi-line value, because the truncation produced a false-red that instructed pasting "the unit has been RETIRED" while the unit is live.
6. 🟢-6 — writing-and-safety.md carries the same bounded claim as index-store.md and points at restore-verify.py.
```

**On round 3's unverifiable-claim note:** it could not reproduce round 2's "812 tests green across 8 files" because that line named no files. Fair. The eight are now named in the commit message: `test_index_store_backup_claim`, `test_analyze_service_index_backup`, `test_analyze_service_index_commit`, `test_doc_path_rot`, `test_skill_audit`, `test_store_content_not_copied`, `test_subsystem_resolver`, `test_prune_skill_size` — 812 again on this tree.

**Sandbox tier:** finally running. It waited out 795s of another session's concurrent build of the same derivation, then started `pytests` sequentially. Result posted here when it lands.

ZacxDev:
## Audit round 4 — all five findings addressed

Fix round landed as `7644e087`. The headline finding is the worst shape available to this PR: a sentence it **added** (round 2), **proved false** (round 3), and then corrected in only two of the three places it lives.

**Payload ledger, round 4:** `4038d8a8..e91604ae` was 43 payload lines (337 since round 1), 86 scaffolding. Above zero — the ladder is on the PR.

Measured, in `verify_r4.py`:

| check | result |
|---|---|
| own-line retraction marker | round-3 filter `True` (green) → round-4 `False` (red) |
| wedged marker | still `False` — round 3's regression test still holds |
| live doc | still `True` |
| wrapped `ExecStart` declaration | **accepted** (round 3 wrongly raised) |
| `writeShellScript` / `Environment = { … }` | both **raise** |
| `ExecStart` vs `ExecStartPre` | still disambiguated |
| refuted phrase asserted anywhere | **none** (only inside dated retractions) |

```audit-claims round=4 audited=e91604ae..7644e087
1. 🟡-A — `run-tests.sh` no longer instructs "check the retention window"; it says to check what is in the bucket with restore-verify.py, and carries a dated retraction recording that the window framing is false. `flake.nix`'s note no longer asserts a drift count that was already wrong when written: both blocks now say three drifts, nothing enforces agreement, change both in one commit or delete one.
2. 🟡-B — `writing-and-safety.md` no longer asserts flatly that orphans are in the bucket: it states a bundle holds them only if they were reachable when THAT bundle was written, gives the same-day-orphan case explicitly, and names the reflog as the holder past that.
3. 🟡-C — `_section_states` matches over LIST-ITEM blocks cut at the next bullet, a blank line, or a retraction line, which closes the own-line-marker walk; the docstring now states the limit as adjacency rather than spelling, and names both measured walks. The sentinel constant is DELETED rather than left unused with a docstring describing it.
4. 🟢-D — `directive()` joins continuation lines to the terminating `;` and refuses only on genuine ambiguity (unbalanced ''/{}/[]), so a wrapped declaration no longer turns a benign nix reformat into a red required check.
5. 🟢-E — `CLAIM_ABSENT` says the bundles survive per HOST-and-scope prefix, and that clearing one prefix is not clearing the scope — the service is not `mkIf`-gated, so both hosts write their own.
```

**On round 4's two unverifiable items:** whether a hand-added ILM rule exists on the live `minio-archive` tenant is genuinely outside both repos — round 4 checked devrc and homelab-talos and found the only lifecycle rules target other buckets under a `daily/` prefix the ASIB key layout never produces. I have not checked the live tenant either, and the paragraph no longer depends on it.

**Sandbox tier:** running now against this sha, clean tree, recorded. ⚠ The earlier green is **discarded, not quoted** — that run was launched at `566c42bf` and sat 795 s waiting out another session's build while the round-2 fixes were being edited into the same worktree, so a path flake ref read a tree that is no commit. The re-run refuses a dirty tree, re-checks immediately before building, and re-reads `HEAD` afterwards.

ZacxDev:
## Sandbox tier — GREEN, and attributable, at `7644e087`

This is the tier the merge actually gates on (`tekton/devrc-pytests` + `tekton/devrc-nodetests`), built from a store copy with no `.git`. It had been unexercised for four rounds; here it is.

```
sandbox-tier2: clean tree at 7644e08744ef7d68584c68b5306e48a0eee5fb05
sandbox-tier2: clear after 0s; building sequentially

pytests   NIXBUILD_RC=0   TOTAL collected=19537 passed=19535 skipped=2 failed=0
                          (floor: 18145 = sum of 29 per-target floors)
                          RESULT: PASS (exit=0)

nodetests NIXBUILD_RC=0   TOTAL suites=5 files=41 tests=1441 pass=1441 fail=0 skipped=0
                          (floor: 1367)
                          RESULT: PASS (exit=0)

tree-after=(head=7644e087… dirty=0)
```

Built **one derivation at a time** — a combined invocation produces documented false failures when the tests that shell out to nested `nix` contend on the store.

### ⚠ The earlier green is discarded, not quoted

A previous run of this tier also returned `pytests=0 nodetests=0`. **It does not count and is not being reported as a pass**, because it cannot be attributed to any commit: it was launched with a clean tree at `566c42bf`, then sat 795 s waiting out another session's build of the same derivation — and the round-2 fixes were edited into the same worktree during that wait. A path flake ref reads the *working tree*, so what it built is neither `566c42bf` nor `4038d8a8`. A live probe against a dirty tree is evidence about the artifact, never about the commit.

The run above closes that hole with three checks the first script did not have:

1. **refuses outright** if the worktree is dirty, rather than building something unattributable;
2. **re-checks cleanliness and `HEAD` immediately before the build** — the wait can be long, and a fact measured thirteen minutes ago is a hypothesis about now, not a finding;
3. **re-reads `HEAD` and dirtiness after the build**, and says so explicitly if the tree moved underneath it.

### Gate status for this PR

| tier | base | verdict |
|---|---|---|
| dev-host (`scripts/gate.sh`, both suites) | merged tree at `566c42bf` | **PASS** |
| sandbox (`nix build .#checks…`, both derivations) | `7644e087` | **PASS** |

⚠ The dev-host run above is from an **earlier** sha. It is quoted at the sha it ran on, not carried forward — I will re-run it on the final tree before calling the PR mergeable, since a green is a claim about one tier, one run, one base.

ZacxDev:
## Audit round 5 — all five findings addressed

Fix round landed as `6913b6f7`. Four of the five trace to one root cause: round 4's list-item scoping and balance heuristic each closed their target walk and reached further than intended. Fixed at the design, not outward from each symptom.

**Payload ledger, round 5:** `e91604ae..7644e087` was 31 payload lines, zero executable (166 since round 1). This round's fix is scaffolding-heavy by design — the finding was that the scaffolding did not hold.

### The sweep, re-run against these fixes

Anchor-asserted (an unapplied mutant **aborts** rather than scoring SURVIVED), `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged per mutant, `.git` removed from every `cp -a` copy, positive control first:

| mutant | round 5 | now |
|---|---|---|
| P0 `_section_states` always True (control) | — | **KILLED** |
| M1 whole-string pin → token presence | SURVIVED | **KILLED** |
| M2 `CLAIM_ABSENT` contains `CLAIM_PRESENT` | SURVIVED | **KILLED** |
| M3 delete the retraction filter | SURVIVED | **KILLED** |
| M4 drop the nix-escape stripping | — | **KILLED** (by its own two tests) |
| M5 revert the continuation-line joiner | SURVIVED | **KILLED** |
| M6 balance check back on single-line values | SURVIVED | **KILLED** |
| M7 whole-line comments only | SURVIVED | **KILLED** |
| M8 `is_conditional` back to the `=` line | SURVIVED | **KILLED** |

⚠ **M4 aborted on its first attempt** — the anchor did not match, so nothing ran. Re-run with a working anchor rather than recorded: an unapplied mutant is not a survivor and is not a corpse.

844 tests green across nine files.

```audit-claims round=5 audited=7644e087..6913b6f7
1. F1 — `_section_states` no longer requires a `-` list item: blocks split on blank lines and at `-`/`*`/`1.` starts, and the claim's own formatting is irrelevant, so `CLAIM_ABSENT` pasted as the PARAGRAPH the failure message names now satisfies its own guard; `*` and `1.` bullets are tested.
2. F1 (second half) — the contact rule disqualifies a block adjacent to a marked one in BOTH directions, closing the mirror case (marker on the line BELOW the claim) that round 4's docstring named as uncaught.
3. F3 — M1/M2/M3 are killed: the polarity, mutual-exclusion and retraction-filter assertions run against paragraph fixtures again, so each measures what its name says rather than a formatting reject.
4. F2 — the balance probe strips nix's `'''`, `''${`, `''$` and `''\` escapes before counting, so a `writeShellScript` body using `''${PATH}` raises instead of returning a truncation; the balance check no longer applies to single-line values, restoring cases the previous reader got right.
5. F4 — `scripts/tests/test_nix_units.py` pins every branch of the shared reader: comment stripping (whole-line, trailing, `#`-in-string, escaped quote), directive name matching and prefix collisions, directive-vs-substring, the continuation joiner, each refusal shape, and both `is_conditional` declaration shapes.
6. F5 — the stated limit now names the actual residual (a retraction using none of the marker words AND blank-separated from the claim) rather than one half of a symmetric hole.
```

**Deliberately not changed:** `index-store.md:47` still frames `ASIB_KEEP` as a "window". Round 5 called it defensible in context — that file describes the LIVE unit, where the pruner does run per upload — and outside its range. Recorded so a later round does not read it as a missed instance.

**Gates:** sandbox tier re-running against `6913b6f7`; dev-host gate to follow on the same sha. The previous sandbox PASS is quoted at `7644e087` and is **not** carried forward.

ZacxDev:
## Audit round 6 — all five findings addressed

Fix round landed as `f3a0bbbd`. The decisive finding is about the mechanism round 5 introduced, and the answer was to **delete it**, not patch around it.

**Payload ledger, round 6:** `7644e087..6913b6f7` was 19 payload lines (391 since round 1), ~4 executable. This round is scaffolding-only by design: the finding was that the scaffolding did not hold.

### The contact rule is gone

It was measured to buy nothing and cost a live fragility:

- **Nothing reached it.** Both shipped retraction fixtures collapse into a single *marked* block, so they were caught by the marked-block rule. The entire taint loop could be deleted with the suite green — while an assertion message read *"the contact rule is one-directional"*, naming a mechanism its fixture never exercised.
- **It made the required check fragile.** The claim bullet was untainted only because an unrelated `- hourly, local` bullet sat between it and the section's ⚠ preamble. Three ordinary edits — delete that bullet, swap the two, add a ⚠ to it — each flipped the gate to FAIL with *"does not carry the canonical claim … Expected this text: `<the claim>`"* **while the claim was present verbatim**.

A retraction *about* a section legitimately precedes the claims it corrects; a retraction *quoting* a claim contains it. Only the second is mechanically separable, so only the second is caught.

### The residual is pinned, not described

Every round so far shipped a docstring wider than its body — the limit was called "spelling" when it was adjacency, then one half of a symmetric hole, then an `and` where the blank line alone decides. So this round stops writing the limit in prose: `test_the_shapes_this_pin_does_NOT_catch` pins the five walking shapes and **fails if any becomes caught**, telling you to update the stated limit in the same commit. `test_the_live_doc_is_not_one_edit_from_an_unfollowable_red` pins the three edits above so the contact rule cannot return by accident.

### Round 6's reported survivors, re-run

| mutant | round 6 | now | died at |
|---|---|---|---|
| P0 `_section_states` always True (control) | — | **KILLED** | 3 tests |
| A drop the `\` escape handling | SURVIVED | **KILLED** | `test_an_escaped_quote_does_not_flip_string_state` |
| B drop the `[`/`]` balance clause | SURVIVED | **KILLED** | `test_a_LIST_value_raises` |
| C `_LIST_ITEM` narrowed back to `-` | SURVIVED | **KILLED** | `test_list_item_splitting_is_what_the_pattern_is_for` |

848 tests green across nine files.

```audit-claims round=6 audited=6913b6f7..f3a0bbbd
1. F1/F2 — the contact rule is removed entirely; a block is disqualified only by containing a retraction marker, and the three ordinary docs edits that reddened the required check under the contact rule are pinned green.
2. F3/F5 — the stated limit is replaced by an executable pin over the five shapes that walk it, which fails if any becomes caught rather than drifting in prose.
3. F4a — the escaped-quote fixture uses an ODD number of escapes and asserts the whole string, so a mutant that stops honouring `\` no longer produces byte-identical output.
4. F4b — a `[`/`]` balance test exists; without the clause the reader returns the truncation `[ "A=1"` instead of raising.
5. F1c — `_LIST_ITEM`'s widening is pinned by what it actually buys (block splitting: a retracted sibling item must not contaminate the next, nor shield itself), not by the claim-recognition loop that passed for the wrong reason.
6. Dead `prev_blank` removed with the loop it belonged to.
```

### ⚠ The dev-host gate run at `6913b6f7` is discarded, not quoted

It reported PASS with `dirty=0` at start and **`dirty-after=2`** — this round's edits landed in the same worktree while it ran, so the verdict describes a tree that is no commit.

That is the **second** time this has happened here. The guard written for the sandbox script after the first occurrence was never applied to the dev-host tier, which also reads the working tree. Both tiers now run from one script that asserts a clean tree at start, re-asserts after the concurrency wait, re-asserts between tiers, and re-reads `HEAD` at the end — running now at `f3a0bbbd`.

ZacxDev:
## Ready for review — both tiers green on the MERGED tree, ladder closed

### Gates — every claim at the sha it actually ran on

| tier | sha | verdict |
|---|---|---|
| **dev-host** (`scripts/gate.sh`, both suites) | `228b8cea` (merged) | **PASS** — pytest 19955/19958 (3 skipped, floor 18383), node 1441/1441 |
| **sandbox** (`nix build .#checks…`, one derivation at a time) | `228b8cea` (merged) | **PASS** — same totals, `RESULT: PASS (exit=0)` from each runner |

Both runs assert a clean tree at start, re-assert after the concurrency wait, and re-read `HEAD` + dirtiness at the end; both reported `tree-after=(head=228b8cea… dirty=0)`.

⚠ **Three earlier greens are discarded, not quoted.** A sandbox run built a tree that was neither `566c42bf` nor `4038d8a8` — it was launched clean, then waited 795 s while a fix round was edited into the same worktree, and a path flake ref reads the *working tree*. A dev-host run then repeated it (`dirty-after=2`) because the guard written for the first had never been applied to the second tier. A third pair ran at `3f9c8144`, which is pre-merge and therefore not the tree the merge creates.

⚠ **The merged tree was gated by hand, because `strict: false`.** `origin/main` moved 13 commits during the ladder and one — `203a5582` — touches `scripts/run-tests.sh`, which this PR also changes. Test-merged rather than reasoned about: textually clean, and the overlapping file was **read**, not assumed — main's edit lands in `TARGET_FLOORS` (`scripts/signal/tests|682` → `|920`) and `EXPECTED_SKIPS`; this PR's lands in the rsync rationale comment. Independent regions, both present in the result. The `skipped=3` and floor `18383` above are the merge having actually taken. Three further commits have landed since, with **zero** overlap.

### The ladder — 7 rounds, every one of which found something real

| round | headline |
|---|---|
| 1 | The sweep was ~half done — 19 more sites, 10 in files already edited; the `unbacked-up` spelling no pattern covered |
| 2 | A 🔴 safety rule got **narrower** while being called sharper; the guard was walkable in one pass and unsatisfiable in the other direction |
| 3 | Two round-2 guards did not pin what their comments claimed; `CLAIM_ABSENT` wrong a second time, reassuringly |
| 4 | A false claim this PR introduced, refuted, then left standing in one of three places |
| 5 | The round-4 narrowing silently defanged three sibling guards; `nix_units` had no tests at all |
| 6 | The contact rule bought nothing measurable and put the live doc one ordinary edit from an unfollowable red — removed |
| 7 | Verdict **safe to merge**, no shipped defect; three unpinned guard branches, closed rather than deferred |

**Stopped on the payload-attribution gate, not on a clean round.** Rounds 6 and 7 both changed zero payload lines, so the ladder had moved to auditing its own scaffolding. Round 7 found no defect in shipped behaviour and its three findings are closed in `3f9c8144`, not filed — what stopped is the dispatch of round 8, deliberately.

Two of the most serious findings were mine, not the original code's: a 🔴 rule I made narrower while advertising it as sharper, and a guard that would have instructed an operator to publish *"the unit has been RETIRED"* into a shipped skill reference while the unit was live.

### Scope this does NOT cover

- Whether a hand-added ILM rule exists on the live `minio-archive` tenant — no declarative API in either repo; the shipped prose no longer depends on it.
- Per-host liveness of the backup timer: `systemctl --user list-timers 'analyze-service-index*'`, by hand, on each host. The guard's own docstring says so.
- `_section_states`' residual is pinned by `test_the_shapes_this_pin_does_NOT_catch`, which **fails if a listed shape becomes caught** — so the stated limit cannot drift wider than the body without a red test.

ZacxDev:
## `tekton/devrc-pytests` is red for a reason outside this PR — evidence

**Do not debug a diff against this check.** It is the documented broken-gate condition, not a bad change.

### What the run actually says

`devrc-ci-5hsmf` @ `228b8cea`: `pytests` **exit 0**, `nodetests` **exit 0**, `verdict` exit 1 —

```
pytests  fail  FAILING: TestTextIsValidatedAgainstEVERYLineBreak.test_PAIRED_CONTROL_…_is_ONE_line
               TOTAL collected=19958 passed=19954 skipped=3 failed=1  (floor: 18383)
```

One failing test in 19,958, in `scripts/tests/test_subsystem_store_api.py`.

### Three facts that rule this PR out

1. **The failing test and the server it exercises are byte-identical** between `3f9c8144` (pytests **SUCCESS**) and `228b8cea` (**FAILURE**). Verified with `git diff --quiet` on both paths. The merge brought 41 files; neither is among them — this PR's content did not change across the green→red transition.
2. **This PR does not touch that test.** Its edits to that file are a header docstring and one section banner; the failing test posts to a live HTTP server and asserts line-break validation.
3. **Four other runs, on four shas that are not this branch's, failed the same way in the same window** (05:55–06:24Z) — each on a *different* test in that *same* file:

| run | sha | failing test |
|---|---|---|
| `devrc-ci-gjljc` | `6428f963` | `TestTheActorComesFromTheTOKEN.test_a_FORGED_actor…` |
| `devrc-ci-j7gw5` | `c4a276d7` | `TestTheWritePrimitives.test_the_LOCK_FILE_is_invisible…` |
| `devrc-ci-fk9sv` | `afe8e190` | `TestTheActorComesFromTheTOKEN.test_a_FORGED_actor…` |
| `devrc-ci-ddrxx` | `857fc3f5` | same |
| `devrc-ci-5hsmf` | `228b8cea` | `TestTextIsValidated…` |

### Local control: 4 passes, same sha

`test_subsystem_store_api.py` run three times standalone → **641 passed** each (292 s, 291 s, 291 s), plus the full sandbox tier at `228b8cea` → `failed=0` over 19958. Four green locally against one red in CI, identical inputs.

### The mechanism — measured live, just now

Per the `tekton` skill's gotcha 6(b), in a **live** gate pod (`devrc-ci-gnktv-gate-pod`, Running):

```
$ ls -d /build            → No such file or directory
$ nix config show         → sandbox = true
                            sandbox-fallback = true
```

`/build`'s **absence is the tell**: nix fell back to building **UNSANDBOXED** and said nothing, while the configured value reads a reassuring `true`. The `nix build .#checks…` tier is only hermetic with the sandbox on — with it off, the same derivation hash produces different results, which is precisely how one suite that binds ports and touches the filesystem goes red on one run and green on the next.

This is already recorded as a 🔴 KNOWN LIMITATION at `devrc-ci-pipeline.yaml:147-154`, where all three fixes are blocked by PodSecurity `baseline:latest`. **Not a new finding, and not in this PR's scope** — filing it again would duplicate an existing record.

### Consequence

With `enforce_admins: true` and both contexts required, nothing merges through this. That is a fleet-level condition affecting every open devrc PR right now, not a #1132 condition. Re-running (a fresh push) may well come back green — the per-test rate is ~1 in 15,500 — but that is a decision for the operator, and the node is currently congested, so an extra push would also cost other people's checks.

PR #1133, opened 2026-08-30
fix(audit-pr): the brief told every auditor to read a `RESULT:` line that `nix build` never prints

ZacxDev:
Three fixes to the audit tooling, all from measurements taken while running four blind audit rounds across #1109 and #1111.

## 1. 🔴 A live defect in the dispatch brief

`audit-dispatch.py` emitted `nix build <worktree>#checks.x86_64-linux.{pytests,nodetests}` with **no `-L`**, inside a section instructing the auditor to "read each runner's own `RESULT:` line". For the sandbox tier — the tier the merge actually gates on — that read cannot succeed.

**Paired control, measured**, on a builder emitting 201 lines plus a `RESULT:` line:

| | console lines | builder lines | `RESULT:` |
|---|---|---|---|
| `nix build` (no `-L`) | 3 | 0 | absent |
| `nix build -L` | streams; `\| tail -40` keeps 40 | 39 | **present, last line** |

⚠ **Why it survived:** a *failing* build DOES print its log tail inline, bounded by `log-lines` (25 here). Only the green case was silent, so the omission never looked broken.

Also added, because it defeats the fix on its own: **an already-built derivation prints nothing at all, `-L` or not** — so silence is never a pass. The brief now hands over `nix path-info --derivation` + `nix log <drv>`, which survives both caching and a swallowed console.

Three of the four auditors this session declined to run the sandbox tier. That is *consistent with* this defect; it does not prove it caused them.

## 2. A delta round cannot see a claim the PR's own earlier commit staled

Every round after the first diffs `<audited tip>..HEAD`. A claim TRUE when an early commit wrote it and FALSIFIED by a later commit of the same PR is inside **no round's range**.

Measured on #1109: `(+ 11 op-selected)` was correct at commit 1 and staled by commit 2; **three delta rounds walked past it, four lines from the paragraph all three were editing.** Found only from outside, by the audit of a different PR quoting the same number. Nothing pinned it — no assertion read the value — so the suite was green throughout.

Remedy in the skill: once per ladder, range-free, re-derive every count/version/cross-reference the PR's files assert against the current head.

## 3. When the payload is prose, the attribution gate is structurally inert

For a docs/skill/prompt PR the `.md` IS the payload, so no round is ever zero-payload and the two-zero-rounds gate can never fire — while the ladder does exactly what the gate exists to catch, since "fixed a defect" and "reworded a warning" are the same edit. Measured on #1111. The skill now carries the stated criterion to stop on, and says to record what is NOT being fixed.

---

Rule text in `SKILL.md`, measurements in `reference/round-ladder-evidence.md`, per that file's existing split. **Nothing pinned by `test_audit_ladder_stop_rule.py` was reworded** — additions sit after the sentences it pins.

**Verified:** 185 passed across `test_audit_ladder_stop_rule.py`, `test_audit_dispatch.py`, `test_skill_descriptions.py`, `test_skill_tiers.py`. The **generated brief** was re-read to confirm it carries `-L` and the `nix log` fallback, rather than trusting the source edit. Sandbox `pytests` tier result posted below.


ZacxDev:
Round 1 (blind) returned five 🟡 and three 🟢. **All five 🟡 confirmed here by independent measurement before fixing** — including re-running the stderr control and the unbuilt-derivation control on this repo's own `checks.x86_64-linux.nodetests`. Fixes are in `ea36a489`.

🔴 Two of the five are this PR's own thesis reflected back: the fix for *"a brief that instructs an impossible read"* instructed two more.

```audit-claims round=1 audited=abb0c27e..ea36a489
1. The TOOLCHAIN block no longer claims `-L … | tail -40` keeps the `RESULT:` line; it states that `-L` writes to STDERR, gives the `2>&1` form, and warns that the `nix log` form is the opposite (stdout) so a redirect must not be carried across.
2. The `nix log` fallback now carries `|| { echo 'NO LOG — never built HERE'; exit 1; }` and `test -s /tmp/tier.log || exit 1` before any grep, and the block states why both are load-bearing.
3. The "do NOT read an exit code" instruction that suppressed the one signal separating "never built" from "clean" is gone from that block.
4. `test_audit_dispatch.py` gains `_assert_every_nix_build_prints_its_log`, which scans EVERY emitted `nix build` line for `-L`/`--print-build-logs` rather than pinning two command substrings.
5. `round-ladder-evidence.md` no longer asserts the brief "is why" three auditors declined the sandbox tier; it says plausible-and-unproven and names the empty-result problem.
6. The prose-PR stop criterion in SKILL.md now carries an explicit "does not override the findings-keyed stop rule / not a shortcut out of a converging ladder" disclaimer, and defines "does not terminate" as "the GATE cannot fire".
7. The delta-range section no longer says the stale claim is "invisible for the life of the ladder"; it distinguishes the out-of-range LINE from the in-range FALSIFYING EDIT, restores the per-round habit, and notes a REBASE stales counts identically.
8. `--no-link` is stated rather than silent, with its justification (READ-ONLY mandate, no `result` symlink in the auditor's cwd).
```

**Ledger: round 1 · payload lines changed THIS round: 89 (`audit-dispatch.py` 34/12, `SKILL.md` 18/4, `round-ladder-evidence.md` 9/4) + 37 scaffolding (`test_audit_dispatch.py` 33/1)** · measured with `git log --numstat --format= --remerge-diff abb0c27e..ea36a489 --not origin/main`, rc 0, non-empty range, git 2.55.0.

**Mutation control on the new guard, because a guard nobody watched fail proves nothing:** stripping ` --no-link -L` from the emitted pytests line fails with **this guard's own message** (`1 failed, 111 passed`), not a neighbour's. The auditor's measurement of the OLD guard is the motivation — 123 passed with the mutant SURVIVING, and a `.pytests`→`.PYTESTS` rename as positive control proving the harness reached those lines.

ZacxDev:
## 🔴 CORRECTION to round 1's claims block, posted rather than quietly fixed

**Round 1's claim 8 was FALSE when I wrote it.** It said *"`--no-link` is stated rather than silent, with its justification"*. No such prose existed — `--no-link` appeared only in the two command lines, and `abb0c27e` already carried it byte-identically, so that range did nothing toward it. I described it in a commit message and never wrote it into the file.

This matters more than an ordinary slip: **`audit-dispatch.py` reprints the claims block into the NEXT round's brief**, so a false claim there is served to the next auditor as an established assertion — the framed-audit failure the same file warns about. The prose now exists in `9ff5c9a9`. Round 2 caught it by grepping instead of believing the ledger, which is the correct behaviour and is why the block says to verify each item.

---

Round 2 dispositioned round 1's eight claims: **five actually fixed, two partially, one NOT fixed (claim 8, above)**. It returned two 🟡 and two 🟢, all confirmed here by measurement. Fixes in `9ff5c9a9`.

```audit-claims round=2 audited=ea36a489..9ff5c9a9
1. The `nix log` block gains `[ -n "$DRV" ] || { echo 'NO DERIVATION — this repo has no such tier'; exit 1; }` as its first guard.
2. The blanket "THE TWO GUARD LINES ABOVE ARE LOAD-BEARING" claim is replaced by a per-guard list naming what EACH of the three closes, including that `test -s` guards empty only and claims nothing about truncation.
3. The block states that an empty `$DRV` makes `nix log ""` resolve as `.` and print the cwd flake's default package log, and that a repo without `packages.default` is immune by luck of flake shape rather than by design.
4. `--no-link` now carries one line of prose stating it is deliberate under the READ-ONLY mandate and changes nothing about what is built or logged.
5. The `exit 1` interactive-shell hazard is stated, with the note that the diagnosis is echoed before each exit.
6. SKILL.md's bolded "Stop on this" imperative now requires NAMING why the ladder will not terminate on its own, and sends the reader to the disclaimer before acting.
```

**Ledger: round 2 · payload lines changed THIS round: 47 (`audit-dispatch.py` 42/9, `SKILL.md` 5/4) · scaffolding 0 · since round 1: 136 payload + 31 scaffolding.** Measured with `git log --numstat --format= --remerge-diff ea36a489..9ff5c9a9 --not origin/main`, rc 0, non-empty range, git 2.55.0.

**Control on the new guard:** in a throwaway flake with a `packages.default` and no `checks` outputs, the pre-fix block printed a foreign `RESULT: PASS (exit=0)` with neither existing guard firing; post-fix the new guard fires with its own message and rc 1, while a real tier still resolves its derivation.

ZacxDev:
**Round 3 of this PR's ladder has now been run** — it was never run before merge, and `9ff5c9a9` (round 2's fix) was the only commit here no auditor had seen. Blind delta audit over `ea36a489..9ff5c9a9`.

**Not a clean round: 4 🟡 + 1 🟢.**

- **F4 is live in `main`** and is fixed in #1157. Round 2's fix added the "ONLY once you can NAME" precondition to the prose-payload escape hatch and **deleted "stated in the round's summary and not left implicit" in the same edit** — wider on one axis, narrower on another. It made a report that *ended* a ladder via the escape hatch indistinguishable from one that *converged*. Now restored and pinned; it had been the only stop-related clause in `SKILL.md` that nothing asserted on.
- **F1, F2, F3, F5 are not actionable against `main`**: #1117 (`1f0fe4c1`) rewrote `render_toolchain` 2h36m after this merged and removed the `nix log` fallback entirely (`grep -c path-info` on `origin/main` → 0). They are recorded in #1157's body as **acceptance criteria for the port-back** #1117 deferred, so they are open rather than absent.

Verified rather than accepted: this PR's commit message claims "301 passed across five guard modules" but names no modules, which makes it structurally unfalsifiable. An exact-fitting set exists and reproduces at `9ff5c9a9` — `test_audit_dispatch` 112 + `test_skill_audit` 116 + `test_skill_tiers` 41 + `test_skill_descriptions` 21 + `test_audit_ladder_stop_rule` 11 = **301**. The number is true.

Issue #1115, opened 2026-08-30
test(handoff_doc): four scaffolding gaps from #1046 round 8 — a dead mutation anchor, an unpinned predicate, and unpinned exit values


ZacxDev:
Four scaffolding findings from round 8 of the `/audit-pr` ladder on #1046 — the round that audited the `main` merge resolution (`48c20d18`) and the shadow fix (`a2381ce2`). **Filed rather than fixed**, per the ladder's attribution gate: round 8 found no payload defect, so the ladder closed.

Sibling of #1093 (rounds 1–7's leftovers). All four were measured, not reasoned.

## 1. 🟡 `scripts/tests/mutants-handoff-cap.sh:199-202` — two mutation rows no longer apply

#962's rule-(i) rows anchor on `s|if not doc.exists() and not args.new_effort:|…|`. #1046's `a2381ce2` rewrote that line into a parenthesised three-term condition, so the anchor matches **0** times. Running the real sweep:

```
== rule (i-b): a second doc for an existing effort (must be KILLED) ==
  🔴 new-doc-guard-never-fires        MUTATION DID NOT APPLY — result meaningless
  🔴 new-effort-assertion-ignored     MUTATION DID NOT APPLY — result meaningless
```

The harness's own `cmp -s` guard reports this **loudly** rather than as SURVIVED, so it is a broken instrument, not a false green. Neither gate tier sees it (`.sh`, not collected). The underlying coverage is intact — re-anchored, both rows kill.

**Fix:** re-anchor to `if (not doc.exists() and not args.new_effort`, and add a third row for the term `a2381ce2` introduced (confirmed to kill via `test_a_STALE_BASE_in_a_REALISTIC_repo_is_not_reported_as_a_NEW_effort`).

## 2. 🟡 `scripts/lib/handoff_doc.py:2444-2445` — the new term has no test pinning WHICH predicate

Replacing `not currency.replaces_mainline_doc("")` with the broader `not currency.stale` **SURVIVES all 295 tests** and changes behaviour.

Scenario: repo with other handoff docs, `claudedocs/handoff-<topic>.md` absent locally, mainline copy whitespace-only → at HEAD **rc 7** (cap fires, nothing written); under the mutant **rc 0, the doc is written** and #962's one-doc-per-effort cap is silently bypassed. Both measured.

[...]

PR #1144, opened 2026-08-31
feat(handoff): rule (k) — an elimination must name HOW it was eliminated

ZacxDev:
A `Ruled out:` bullet in a handoff **update** must carry `via: <kind>` — from a
closed vocabulary: `command`, `measurement`, `code`, `change`, `doc`, or the
honest opt-out `assumed`. New refusal `status=unevidenced`, exit **10**.

## Why

MEASURED 2026-08-30. An opencode session on a weaker model wrote this into a
handoff's open-investigations block:

```
- **Ruled out:** Not a per-DIMM issue (all 4 identical)
```

It had run `inxi -CmG` and `inxi -dm`. **Neither prints a part number.** The
sentence was an elimination its own data could not support, and it was false —
one flag away (`inxi -max`) sat two different part numbers from two different
kits, killing the theory the doc then ranked **first**, with a ⚠ and "free
performance sitting on the table". That rank-1 instruction was to reboot a
26-day-uptime workstation into its BIOS.

An elimination is the claim a later session trusts most and re-checks least — it
is what stops the next reader looking. `/resume` already warns that a
mid-diagnosis block "reads as current forever"; this is that hazard one level
down, at the individual bullet.

## Why a closed vocabulary and not a content check

Measured against that bullet, **every cheap heuristic accepts it**: it contains a
digit, it is fluent, it is the same length as bullets that do cite evidence.
What it lacks is not a word but a measurement — so a content sniff is exactly the
"guard on WORDS walkable by REWORDING" that `RULES.md` warns about. Rule (j)'s
shape instead: an allowlist the author must pick from.

`assumed` is **accepted and counted**, for `forcing: none`'s reason. Refusing it
would not stop anyone reasoning their way to an elimination — it would teach them
to type `command` falsely, which moves the population underground. The failure
was never that the session reasoned rather than measured; it was that the doc did
not *say so*.

The gate cannot tell a true elimination from a false one and does not try.

## Why it is not a permanently-red gate

- **It reads the UPDATE, never the merged doc.** `open investigations` is an
  APPEND heading, so the merge carries every elimination ever written — **122
  across 44 of this repo's 90 docs**, none now editable. Checking the merge
  would refuse on run one, forever.
- **The marker needs no punctuation.** The first version demanded a separator
  right after `Ruled out` and **missed nine real bullets**, all the same shape: a
  qualifier before the colon (`Ruled out as writers:`, `Ruled out
  (structurally):`, `Ruled out, and still true:`). Such a gate is escaped by
  typing `Ruled out (obviously):`.

## Controls

- **Positive** — the committed corpus yields **122** elimination bullets, pinned
  at a floor of 90. A pattern matching nothing reads exactly like a gate that
  passes.
- **False-positive** — `via:` followed by a vocabulary member occurs **0 times**
  corpus-wide. That is also why the key is `via` and not the more
  self-documenting `evidence`: `evidence: the ACCESS_DENIED is positive evidence
  it is NOT` is a real shape here, and against it the kind group captures `the`,
  refusing a bullet whose author *did* cite a measurement.

## Two bugs found while building this

- **The EXIT CODES legend stopped at 8** while `EXIT_STALE_BASE = 9` was live,
  reachable and printed at callers. The suite was green throughout — the only
  legend guard reads `SKILL.md`, never the module's own docstring. Added
  `test_the_module_legend_documents_every_nonzero_code`.
- **Two rule (i-b) mutation rows were silently meaningless.** Both targeted the
  guard's one-line form; the condition was wrapped across two lines at some
  point and neither sed was updated, so both had been reporting `MUTATION DID
  NOT APPLY` — rule (i-b)'s two headline guards were UNVERIFIED while the sweep
  looked complete. Re-anchored; both now killed by their named test.

`_item_blocks` is **parameterised** on the item pattern rather than copied, so
rule (k) reuses rule (j)'s measured block-boundary and fence handling instead of
regenerating its bugs.

## Verification

- **Dev-host tier:** `RESULT: PASS (exit=0)` — 19,537 passed, 0 failed, floor
  18,145. Via `run-tests.sh` inside the repo's own dev shell; plain `gate.sh`
  exits 3 outside it, which is a could-not-run, not a pass.
- **Sandbox tier**, built ONE AT A TIME per CLAUDE.md:
  `.#checks.x86_64-linux.pytests` rc=0, then `.nodetests` rc=0. Both saw the
  change (dirty-tree warning, derivation actually built, not a cache hit).
- **Mutation battery:** 46 rows, 0 failures. Seven new rule (k) rows, each
  killed by its **own** named test rather than a neighbour's, plus a
  behaviour-free control that SURVIVED as required.

⚠ The merged tree has **not** been gated — `strict` is false on this repo, so
these are claims about this branch.

## SKILL.md budget

The body had 34 bytes of headroom. Following the size test's own playbook, the
argument went to `reference/write-gate.md` §D (no ceiling) and the refusal itself
prints the vocabulary and remedy, so `SKILL.md` carries one sentence. Paid for by
evicting a clause that restated the forcing vocabulary two sentences after the
list, and which `handoff_doc.py:884` already prints at refusal time. Net **+13
bytes**, still inside the budget.


ZacxDev:
## Round 1 audit → fixed, and re-audited by the battery

An adversarial audit found the payload correct and both merge-gating tiers green, but the **clearability** half missing. Fixed in `a960d43a`:

- **F1** — four remedies, one **per cause**, and only the causes present. The first revision printed one unconditional trailer, reproducing all three unrecoverable-refusal shapes rule (j) had already been fixed for *twice*: a near-miss told to add a field it has, a fenced field told only to "move it out" (which promotes a quote into a false declaration), a flush-left field never told to INDENT.
- **F2** — `MARK_NO_VIA` was printed and appeared **nowhere** in SKILL.md. `REFUSAL_MARKERS`' two-way guard could not see it: it asserts a *length of 4* against rule (j)'s tuple. New `ELIMINATION_MARKERS` + a guard asserting the skill against **that**, not a number.
- **F3** — the step-2 **template** still taught the refused shape. Measured by replaying every past handoff update: **80 of 242 deltas (33%)** add an elimination bullet and **none** would pass. Template now names `via:`.
- **F4** — a `🔴` between the bullet and the marker hid **four real corpus bullets**. Same class as the qualifier fix, mirrored left. The admitted run is non-letter on purpose — `- **NOT ruled out:**` is a real line asserting the opposite. Corpus is **126 / 45 of 90**, and every place that asserted 122/44 (or 121/45) is corrected.
- **F5** — an unrelated **column-0 fence** forged a `[fenced]` verdict, because `_item_blocks` appends fenced lines to `hidden` before its boundary check. The likeliest such paste is *this tool's own refusal* — what a session re-running after rc 10 has in its scratch file.
- **F8/F9/F12/F13** — `NOTHING WRITTEN` + an elided-count line; and three comments corrected against the code, including a pre-existing "stale-base exits 7" that is 9.

### 🔴 The battery's own extractor was broken

`failing()`'s module-level pattern is **end-anchored**, so a module-level *parametrised* test (`name[param]`) matched neither it nor the `class::method` pattern — its failures were invisible and its mutant scored **SURVIVED**. Measured: `leading-decoration-rejected` reported SURVIVED while running the killing test by hand failed 3 of its 4 cases. Third pattern added. Two of this round's own rows were also meaningless until fixed (a stale anchor; a sed producing invalid Python that ran 0 tests) — the harness reported both honestly rather than as passes.

### Verification at `a960d43a`

- Mutation battery: **52 rows, 0 failures** — 13 rule (k) rows, each killed by its **own** named test, control surviving.
- Dev-host tier: `RESULT: PASS (exit=0)` — 19,551 passed, 0 failed.
- Sandbox tier, one at a time: `pytests` rc=0, `nodetests` rc=0.
- SKILL.md 25,834 B; `MAX_BYTES` 26,400 → 27,000 (+600), lever 3, with the ledger entry the option owes a reader. Levers 1 and 2 were spent first, and what could not be evicted is case (a) of the 2026-08-28 entry verbatim.

⚠ Still not gated: **the merged tree** (`strict` is false here).

ZacxDev:
Heads-up: **#1146 just merged (squash `edbc596f`) and it will fail this PR's required checks until one line changes here.** Sorry — I merged first; here is everything you need so it costs you a minute rather than a debugging session.

## The one-line fix

#1146 closed #1115.3, which asked for the `EXIT_*` constants' *values* to be pinned exactly. `scripts/tests/test_handoff_doc.py` now ends `test_no_two_exit_constants_share_a_value` with:

```python
assert len(codes) == 9, f"the EXIT_* constant set changed: {codes}"
```

This PR adds `EXIT_UNEVIDENCED = 10`, a legitimate 10th constant, so **change that `9` to `10`**. The exact pin is the point — it exists so that adding a constant cannot be silent.

## What you will see, and what you will NOT see

**The failure is diagnostic, not a bare count mismatch.** An earlier revision of #1146 had the count assertion sitting *above* the injectivity loop, so a colliding constant died to `assert 10 == 9` and the remedial text never rendered — inviting exactly the wrong fix. Its round-2 audit caught that and moved the assertion below the loop, so:

- a **new, non-colliding** constant (your case) reaches `assert len(codes) == 10` and prints the whole dict;
- a **colliding** constant instead reports which two constants share a value, with the remedy.

**You may also want the `(exit N)` prose check.** `test_the_prose_quotes_the_CONSTANT_not_a_stale_literal` now binds each documented `status=…` token to its code in `SKILL.md`. If rule (k) documents a `status=` token for `EXIT_UNEVIDENCED`, adding its tuple there keeps that guard honest; if it does not, nothing to do.

## The SKILL.md byte ceiling is NOT a problem — please ignore anything you read to the contrary

An earlier revision of #1146's description claimed the merged tree would blow the enforced ceiling by ~2 KB and need an eviction. **That was measured wrong and is retracted** — the 27,510 B figure was `git merge-file`'s *conflicted* output, markers and duplicated region included, which is not a resolution anyone would commit. Re-measured properly:

| | bytes |
|---|---|
| `origin/main` before either PR | 25,466 |
| #1146's delta | +27 |
| this PR's delta | +368 |
| **union resolution** | **25,861** |
| merged budget (this PR raises `MAX_BYTES` to 27,000, minus `MIN_HEADROOM_BYTES` 900) | **26,100** |

**+239 B of headroom. No eviction needed.** Both PRs' `SKILL.md` edits are single-hunk and land in disjoint clauses of line 158, so the additive model holds.

## Conflicts

Merging `main` will conflict textually in `SKILL.md`, `test_handoff_doc.py` and `mutants-handoff-cap.sh` — loud, not silent. Worth reading the merged result of each region rather than taking a clean `git merge` as safe: a *semantic* conflict survives a clean textual merge, and that is precisely what cost #1046 eight audit rounds (two PRs appending to one `main()`, no marker, both suites green).

One thing to look at in `mutants-handoff-cap.sh`: #1146 added rows grouped by *kind* (regression coverage / invariant guard / diagnostic-quality), measured against the merge-base. If rule (k) adds mutation rows, the same distinction applies — a mutant already killed at the base is an invariant guard, and labelling it as regression coverage is the over-claim that took three audit rounds to stop recurring in #1146.

Not touching this branch — it is your claim. Ping me if you would rather I prepared the merge commit.

PR #1148, opened 2026-08-31
fix(nogit_plugin): serialise the config control — a retry only randomises a collision it cannot remove

ZacxDev:
Closes the mechanism behind clawgate cg#445: `scripts/gate.sh --tier pytest` has been permanently red on the workbench for **attribution-only** reasons — GUARD 10 refusing to vouch for itself, which `claude/RULES.md` names as worse than no gate because it trains everyone to click through.

## The mechanism, and a correction to the card

cg#445 attributes this to the git **common dir's** config shared by ~117 worktrees — which frames it as an environmental fact about the box, i.e. unfixable. That is how all **three** prior rediscoveries ended (2026-08-22, 2026-08-28, 2026-08-30).

The contended file is not `.git/config`. It is the guard's own `GIT_CONFIG_GLOBAL` target, and `nogit_plugin.py`'s own comment already said so:

> *every worker runs its own session and fires this control, and they all write the SAME guard file — git takes an exclusive `<cfg>.lock` per write, so the losers exit 255*

The control's key and token are already per-pid, so **only the file lock collides**. `LOCK_RETRIES = 6` with jitter randomises the collision; it cannot remove it, and under enough concurrent workers it exhausts and reports `unmeasured`.

## 🔴 The obvious fix is wrong, and this PR does not ship it

Making the config filename per-pid removes the contention — and **breaks GUARD 10 on every target**, because the shared filename is load-bearing in `scripts/run-tests.sh`, which constructs the path independently in shell as `$NOGIT_CONFIG`:

- **`run-tests.sh:4562`** — `[ "$m_redirect" != "$NOGIT_CONFIG" ]`, an **equality** test. Per-pid names trip *"Something between this script and pytest is re-pointing git's global config"* everywhere.
- **`run-tests.sh:3023`** `_nogit_controls()` — counts `control-*` keys in that **one** file, pinned to the session count at `:4566`. Per-pid files scatter the keys and it reads as *"the redirect is not where global writes land"*.

Neither greps for `guard_config_path`, which is exactly why a Python-symbol search misses them. One path, open-coded in two languages.

## What this ships instead

An advisory `flock` on a **separate** `<guard-dir>/gitconfig.pylock`, held only around the git invocation, never across the backoff sleep. `CONFIG_NAME`, `guard_config_path` and the `GIT_CONFIG_GLOBAL` target are **untouched**, so both bash consumers and the protection half are unchanged.

- 🔴 The mutex suffix is `.pylock`, never `.lock` — that is git's own lock filename for the same file. Pinned by its own test.
- Acquire is a polled `LOCK_EX|LOCK_NB` to a bounded deadline; a blocking acquire inside a session-scoped autouse fixture would hang a target with no output.
- It **fails open** to the pre-existing unlocked write, so it can never itself produce `unmeasured`.
- Retry/jitter kept as defence in depth against writers that never took the mutex.

## Evidence

**RED at `203a5582`, GREEN at HEAD**, same test, same command:

```
GREEN AT HEAD:  2 passed
RED AT BASE:    AssertionError: 10 of 16 concurrent sessions could not measure their own
                `git config --global` control: … still lock-contended after 6 attempts
```

**Re-verified independently of the implementing agent:** removing the serialisation while *keeping* the retry still fails — `11 of 16 concurrent sessions could not measure` — so the fix is the serialisation, not incidental.

**Mutation sweep**, fresh module per mutant, `PYTHONDONTWRITEBYTECODE=1`:

| mutant | result |
|---|---|
| no-op (**positive control**) | SURVIVED — the harness is not killing indiscriminately |
| serialisation removed | KILLED, 10/16, with the lock-contention message |
| mutex named `.lock` | KILLED, 16/16 |
| `CONFIG_LOCK_SUFFIX = ".lock"` in-tree | KILLED by the suffix test, with its own message |

Test fixture note: rather than 160 concurrent writers (measured 74 failures, but expensive), it makes **one write slow** via a 200k-key guard config — the actual precondition under which a backoff loop stops working — and is reliably red at a sixth of the process count.

## The real gate — the original symptom, gone

`gate.sh --tier pytest` on a clean worktree at `f1fa6cfa`:

```
GATE: RESULT=PASS exit=0
TOTAL collected=19784 passed=19781 skipped=3 failed=0   (run=1081s)
scripts/validation/tests  real-config-changed=0/3  config-control=4  protocol=refused
```

The originally-failing target reports `config-control=4` — all four xdist workers measured their own control. Zero GUARD 10 problems.

## Not verified

- **The sandbox tier was not run.** GUARD 10's `NOGIT_REPO_LOCAL` is empty in the nix sandbox, so that tier is not running the same guard — a green there would not be evidence for this.
- One gate run, one base sha; **not gated on the merged tree** against current `main`.
- The gate reported 3 unattributable repo-local deltas from other sessions doing `git branch` in sibling worktrees. Pre-existing, unrelated, not enforced.

### DIFF
--- a/scripts/testlib/nogit_plugin.py
+++ b/scripts/testlib/nogit_plugin.py
@@ -99,6 +99,8 @@ see (an atexit hook, a lingering thread, a fixture finalizer).
 """
 from __future__ import annotations
 
+import contextlib
+import fcntl
 import os
 import random
 import shutil
@@ -164,6 +166,38 @@ LOCK_TIMEOUT = 5
 CONTROL_OK = "emitted"
 CONTROL_UNCONTAINED = "WITHHELD-UNCONTAINED"
 
+# 🔴 THE CROSS-PROCESS MUTEX THAT MAKES THE RETRY LOOP A FALLBACK RATHER THAN THE
+# PLAN. Every session sharing one guard dir fires the control write above; git's
+# own `<cfg>.lock` is exclusive and NON-blocking, so the losers exit 255
+# immediately and the retry loop is a thundering herd racing to re-collide. It
+# only wins while the write is faster than the backoff: MEASURED on this box,
+# 96 concurrent writers to an empty guard file produced 3 exhaustions, 160
+# produced 74, and 16 writers against a config large enough to make one write
+# outlast the jitter produced 11. The operator's box is the loaded case, and
+# GUARD 10 read the exhaustion as `control=unmeasured` — the guard unable to run
+# its own positive control, which the runner correctly refuses to score as a
+# pass.
+#
+# So the writers are SERIALISED here instead, on an advisory `flock` every
+# participant takes before invoking git. There is no collision left to back off
+# from, and nothing about the guard's accounting moves: the redirect still names
+# the ONE file `run-tests.sh` exported (checked for equality at its GUARD 10
+# evaluation), and the control keys still accumulate in that ONE file where
+# `_nogit_controls()` counts them one-per-session.
+#
+# 🔴 `.pylock`, NEVER `.lock`: `<cfg>.lock` is git's OWN lock file name, and
+# creating it ourselves would make every `git config --global` write in the run
+# fail outright — the guard's positive control silenced by the thing meant to
+# protect it.
+#
+# The lock is advisory and released by the kernel when the holder's fd closes,
+# so a killed worker cannot wedge the run. Acquisition is bounded: on timeout the
+# write proceeds UNLOCKED, which is exactly the pre-serialisation behaviour, and
+# the retry loop below is still there to catch it.
+CONFIG_LOCK_SUFFIX = ".pylock"
+CONFIG_LOCK_WAIT = 60.0
+CONFIG_LOCK_POLL = 0.005
+
 # The protocol probe. `.invalid` is reserved by RFC 2606 and can never resolve,
 # so if the allowlist is NOT in effect this fails at DNS — a DIFFERENT error,
 # which is exactly what makes the two states distinguishable. Nothing leaves the
@@ -181,6 +215,59 @@ REFUSAL_TOKEN = "not allowed"
 def guard_config_path(guard_dir: Path) -> Path:
     """The throwaway file `git config --global` writes to under this guard."""
     return Path(guard_dir) / CONFIG_NAME
+
+
+def guard_config_lock_path(guard_dir: Path) -> Path:
+    """The advisory lock serialising this guard's `git config --global` writes.
+
+    A SIBLING of the config file, never the config file itself and never git's
+    own `<cfg>.lock` — see CONFIG_LOCK_SUFFIX. Nothing reads it; it exists only
+    to be flock'd.
+    """
+    return Path(guard_dir) / (CONFIG_NAME + CONFIG_LOCK_SUFFIX)
+
+
+@contextlib.contextmanager
+def _config_write_lock(guard_dir: Path, wait: float = CONFIG_LOCK_WAIT):
+    """Hold the guard dir's config mutex for one `git config --global` write.
+
+    Yields True if the lock is held, False if it could not be taken — and False
+    is not a failure path: the caller writes anyway, because an unserialised
+    write is the behaviour that shipped before this lock existed and the retry
+    loop still covers it. FAILING here must never turn into `unmeasured`, which
+    is the verdict this whole change exists to stop producing.
+
+    Polled `LOCK_NB` rather than a blocking `LOCK_EX` so the wait is BOUNDED. A
+    blocking acquire against a holder that wedged would hang the session inside
+    a session-scoped autouse fixture, i.e. hang the target with no output.
+    """
+    fh = None
+    held = False
+    try:
+        try:
+            Path(guard_dir).mkdir(parents=True, exist_ok=True)
+            fh = guard_config_lock_path(guard_dir).open("a+")
+        except OSError:
+            fh = None
+        if fh is not None:
+            deadline = time.monotonic() + wait
+            while True:
+                try:
+                    fcntl.flock(fh.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
+                    held = True
+                    break
+                except OSError:
+                    if time.monotonic() >= deadline:
+                        break
+                    time.sleep(CONFIG_LOCK_POLL)
+        yield held
+    finally:
+        if fh is not None:
+            if held:
+                with contextlib.suppress(OSError):
+                    fcntl.flock(fh.fileno(), fcntl.LOCK_UN)
+            with contextlib.suppress(OSError):
+                fh.close()
 
 
 def install(guard_dir: Path) -> Path:
@@ -251,14 +338,24 @@ def config_control(guard_dir: Path) -> tuple[str, str]:
     key = f"{CONTROL_SECTION}.{CONTROL_PREFIX}{os.getpid()}"
     token = f"nogit-{os.getpid()}"
 
-    # 🔴 RETRY ONLY ON LOCK CONTENTION. Under pytest-xdist every worker runs its
-    # own session and fires this control, and they all write the SAME guard file
-    # — git takes an exclusive `<cfg>.lock` per write, so the losers exit 255
-    # with "could not lock config file". MEASURED: 8 concurrent
-    # `git config --global` writes to one file -> 4 exit 0, 4 exit 255 with that
-    # message. Before the retry, GUARD 10 reported `control=unmeasured` for 16 of
-    # 27 targets — the guard could not run its own positive control, which it
-    # correctly refuses to score as a pass.
+    # 🔴 RETRY ONLY ON LOCK CONTENTION — DEFENCE IN DEPTH, NOT THE MECHANISM.
+    # Under pytest-xdist every worker runs its own session and fires this
+    # control, and they all write the SAME guard file — git takes an exclusive
+    # `<cfg>.lock` per write, so the losers exit 255 with "could not lock config
+    # file". MEASURED: 8 concurrent `git config --global` writes to one file ->
+    # 4 exit 0, 4 exit 255 with that message. Before the retry, GUARD 10 reported
+    # `control=unmeasured` for 16 of 27 targets — the guard could not run its own
+    # positive control, which it correctly refuses to score as a pass.
+    #
+    # The retry ALONE was not enough, and was never going to be: it randomises a
+    # collision instead of removing one, so it holds only while a write finishes
+    # inside the backoff window. On a loaded box it does not, and GUARD 10 went
+    # back to `unmeasured` (measured on `scripts/validation/tests`). The
+    # `_config_write_lock` below is what removes the collision; this loop is KEPT
+    # because it covers the contention the lock cannot see — a writer that never
+    # took it (a test in the target firing its own `git config --global`, a
+    # concurrent process reaching the same file, a session that failed to open
+    # the lock at all).
     #
     # The retry is deliberately NARROW: any other non-zero exit still returns
     # "unmeasured" on the first try. Widening it to all failures would convert a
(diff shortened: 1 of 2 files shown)


ZacxDev:
Audit round 1 claims, for round 2 to check.

```audit-claims round=1 audited=f1fa6cfa..38975c6e
1. 🟡-1 stale worst-case comment: claimed FIXED in BOTH places (the retry loop at the `LOCK_RETRIES * timeout` sentence, and the CONFIG_LOCK_WAIT constant block), stating the ceiling as LOCK_RETRIES * (CONFIG_LOCK_WAIT + LOCK_TIMEOUT) + jitter ≈ 391s.
2. 🟡-2 exhaustion verdict conflating three mechanisms: claimed FIXED — the verdict now appends (mutex=held|NOT-held, stale-git-lock=PRESENT|absent). `held` was previously computed and discarded; the context manager is now bound with `as held`.
3. 🟡-2: claimed verified by planting a stale `<cfg>.lock` and observing `(mutex=held, stale-git-lock=PRESENT)`, with a negative control on a clean dir returning CONTROL_OK.
4. 🟢-3 errno: claimed FIXED — `except OSError as exc`, fails fast on any errno outside {EWOULDBLOCK, EAGAIN} instead of polling the full wait. `import errno` added.
5. 🟢-5 CONFIG_LOCK_WAIT default-arg capture: claimed FIXED — signature is now `wait: float | None = None`, read from the module attribute at call time; an explicit `wait=` still wins.
6. 🟢-4 child barrier: claimed FIXED — 120s bounded barrier emitting a BARRIER-TIMEOUT status, plus parent-side `finally` that kills survivors and a TimeoutExpired branch that kills and continues.
7. 🟢-6 hardware ratio: claimed DOCUMENTED only, no behaviour change — a comment on _CONTENDERS/_CONTENDED_CONFIG_KEYS stating they are a ratio against the ~0.75s jitter budget.
8. Claimed the full nogit suite is green at 38975c6e (46 passed). NOT re-run: `gate.sh --tier pytest` (last run at f1fa6cfa), and the sandbox tier (never run; it does not exercise this guard).
```

**Ledger — round 1 · payload lines changed THIS round: 69 (+8 deleted) in `scripts/testlib/nogit_plugin.py` · scaffolding: 41 (+8) in `scripts/tests/test_nogit_isolation.py`.** Payload is non-zero, so the ladder has not left the PR.

ZacxDev:
Audit round 2 claims, for round 3 to check.

```audit-claims round=2 audited=38975c6e..a8861d34
1. N1 (🟡, payload) mutex field sampled one attempt: claimed FIXED — `held_count` accumulates across the retry loop and the verdict prints `mutex=held {held_count}/{attempt + 1}` instead of a boolean.
2. N1: claimed covered by a NEW regression test `test_the_exhaustion_verdict_counts_acquisitions_it_does_not_sample_one`, driven to exhaustion by holding the mutex from the test process AND planting a stale `<cfg>.lock`; asserts the literal `mutex=held 0/6` and `stale-git-lock=PRESENT`.
3. N1: claimed mutation-verified — reverting the ratio to the sampled boolean fails that test; unmutated it passes.
4. N2 (🟢, payload) errno rationale factually wrong: claimed FIXED as a COMMENT-ONLY change. Behaviour deliberately unchanged. The comment now says ENOLCK is rare-and-transient per flock(2), that EPERM is not a documented flock error, that fail-open is a deliberate trade rather than a property of the errno, and that neither path is reachable here so it rests on the man page rather than a live repro.
5. N4 (🟢, scaffolding) cumulative child budget: claimed FIXED — a single 600s whole-fixture deadline replaces per-child `timeout=300`, with each `communicate` given the remaining time. 600 was chosen to exceed the documented 391s single-control ceiling; the comment says this, not `_CONTENDED_CONFIG_KEYS`, is the real cap.
6. N4: claimed the killed child’s stdout/stderr are now captured into the CHILD-TIMEOUT detail rather than discarded for a fixed string.
7. N3 (🟢, scaffolding) no tests for the delta: claimed ADDRESSED by two new tests. The second, `test_CONFIG_LOCK_WAIT_is_read_at_call_time_not_captured_at_import`, asserts by ELAPSED TIME and was watched RED — 60.28s with the default-argument capture restored, against its 5s bound; 0.30s when correct. It carries its own positive control that an explicit `wait=` still wins.
8. Claimed 48 passed at a8861d34. NOT re-run since 38975c6e: `gate.sh --tier pytest` (PASS on record at 38975c6e, 19781 passed). Sandbox tier never run and would not be evidence.
```

**Ledger — round 2 · payload lines changed THIS round: 35 (+11 deleted) in `scripts/testlib/nogit_plugin.py` · scaffolding: 113 (+2) in `scripts/tests/test_nogit_isolation.py`.** Payload non-zero, so the ladder has not left the PR.

⚠ **Process note for round 3, because it cost me twice this session:** I twice ran `git checkout -- scripts/testlib/nogit_plugin.py` to restore after a mutation while my fixes were still UNCOMMITTED, which silently discarded them — the second time I caught it only because a brand-new test failed unmutated and the message showed the OLD verdict format. Fixes are now committed before any mutation. The rule that would have prevented it is in this repo already: mutate a `cp -a` copy, never the tracked file.

ZacxDev:
Audit round 3 claims, for round 4 to check.

```audit-claims round=3 audited=a8861d34..21176d1d
1. F1 (MEDIUM, scaffolding) install() leaking five process-wide env vars into 42 later tests: claimed FIXED — both round-2 tests now use `monkeypatch.setenv(nogit_plugin.CONFIG_ENV, ...)` (the pattern the file already used) and create the cfg file themselves; neither calls `install()`. Both signatures now take the `monkeypatch` fixture.
2. F2 (MEDIUM, scaffolding) `held_count += bool(held)` -> `+= 0` SURVIVED: claimed FIXED — the test now also drives the NON-zero side (same stale `<cfg>.lock`, mutex NOT held) and asserts the literal `mutex=held 6/6`, pinning mechanism (a) beside the existing (b)+(c) case.
3. F2: claimed mutation-verified on a `cp -a` copy with `.git` removed — the previously-surviving `+= 0` mutant now FAILS the test; unmutated the copy passes.
4. F4 (LOW) budget comment cited the wrong ceiling: claimed FIXED — the comment now says 391s is the RETRY LOOPs bound, that config_control adds a post-loop read-back at `_git`s DEFAULT timeout of 30, that the calls real ceiling is ~421s, and to re-derive from 421.
5. F3 (LOW) CHILD-TIMEOUT capture empty in the case it targets: claimed ADDRESSED AS DOCUMENTATION ONLY — no behaviour change; the comment records the measurement that an empty tail means "killed before its single JSON write", not "said nothing useful".
6. F5 (LOW) header ledger not updated: claimed FIXED — both round-2 tests are now enumerated in the "WHAT IS REGRESSION COVERAGE HERE AND WHAT IS NOT" block with what each pins.
7. Claimed 48 passed at 21176d1d. NOT re-run since a8861d34: `gate.sh --tier pytest` (PASS on record at a8861d34, 19783 passed, under heavy concurrent load). Sandbox tier never run and would not be evidence.
```

**Ledger — round 3 · payload lines changed THIS round: 0 · scaffolding: 65 (+8) in `scripts/tests/test_nogit_isolation.py`.**

🔴 **This is the FIRST zero-payload round, and the trend is the one the attribution gate watches for.** Payload per round: **113 → 69 → 35 → 0**. Per `audit-pr`s own rule, *two consecutive* zero-payload fix rounds mean the ladder has left the PR and should stop. One more like this and it fires. Round 4 is run because the rule is keyed to findings, not to a count — but if its findings are again scaffolding-only, that is the stop signal, not a reason for round 5.

ZacxDev:
## Audit ladder CLOSED at round 4 — clean

Round 4 returned **zero findings**, and both stop conditions fired together:

- **A clean round ends the ladder**, and per `audit-pr` a clean round is never re-run to confirm it.
- **The attribution gate's stop condition is met**: payload lines per fix round **113 → 69 → 35 → 0 → 0**. Two consecutive zero-payload rounds mean the ladder has left the PR and is auditing its own scaffolding.

### What four rounds actually bought

Every round found a defect **created by the previous round's fix**, which is the pattern `claude/RULES.md` says to budget for:

| round | found |
|---|---|
| 1 | An `IndexError`-class defect plus 5 others in the original fix |
| 2 | 🟡 the `mutex=` field I added in round 1 **sampled one retry attempt** while reading as the whole run — it would print `mutex=held` on a run where 4 of 6 writes went unserialised, sending the reader after a rogue writer that does not exist. The field added to prevent a misdiagnosis was producing one. |
| 3 | 🟡 my round-2 tests called `install()` in-process, leaking a redirect into **42 later tests**; and the mutant `held_count += 0` **SURVIVED**, because my fixture could only ever produce `0` and I asserted that literal — the "fixture derived from the constant under test" trap, walked into *while fixing a mutation-coverage defect*. |
| 4 | nothing |

Round 4 did not take those on trust. It rebuilt the round-2 tree and **reproduced the 42-test leak from scratch** as a positive control before confirming it is now 48/48 clean, and independently re-ran the mutation table — including verifying that `held_count += 0` genuinely survived before the second assertion existed, and that both halves of that test are load-bearing (each kills a mutant the other misses).

It also found the fix is **wider than claimed**: `install()` no longer appears in any in-process test path anywhere in the suite, and dropping it loses nothing, because the other four env vars it sets are guard_dir-independent constants the session fixture already sets identically.

### Outstanding, and being done now

Round 4 correctly flagged that `gate.sh` has not run at `21176d1d`, and that a merge-safety claim needs the **merged** tree rather than this branch. Building that now in a real clone (`origin/main` + this PR), both sandbox tiers one at a time. I will post the result before merging.

### Two process failures of my own, recorded because they nearly cost real work

- I twice ran `git checkout -- scripts/testlib/nogit_plugin.py` to restore after a mutation **while my fixes were uncommitted**, silently discarding them. The second time was caught only because a brand-new test failed *unmutated* and its message showed the old verdict format. Without that test I would have committed a fix that was not there. Fixes are now committed before any mutation, and mutations run on a `cp -a` copy.
- A commit message lost a bullet to backtick command substitution; amended via `-F <file>`.

ZacxDev:
## Merged-tree gate — all three tiers green

`strict: false`, so a branch check is a claim about the branch, not about the tree the merge creates. Built an integration tree in a **real clone** (never a worktree — `nix build path:<worktree>` breaks every git call in the sandbox): **`origin/main 74b427d5` + `pr1148 21176d1d`**, clean merge, both sides verified present.

```
checks.x86_64-linux.pytests    -> /nix/store/gzx25hxsyzz8bj793qqhkmf9w2zfnr52-devrc-pytests     GREEN
checks.x86_64-linux.nodetests  -> /nix/store/s112ljw56ww0z231r8ls7baf3fw5yi4b-devrc-nodetests   GREEN
gate.sh --tier pytest (dev-host)  GATE: RESULT=PASS exit=0   19914 passed / 3 skipped / 0 failed
```

Built one at a time, and the verdict taken from the produced out-paths and the runners own `RESULT:` line rather than a piped exit code.

### 🔴 The two sandbox tiers do NOT prove this fix works — only the third does

`NOGIT_REPO_LOCAL` is empty in the nix sandbox, so GUARD 10 is not the same guard there. Those two greens establish **merge safety**; only the dev-host tier establishes that the thing this PR exists to fix is fixed. Stating them separately because "the gate passed" would otherwise read as a stronger claim than it is.

**The dev-host evidence, on the merged tree:**

- **Zero occurrences of `unmeasured` or `lock-contended` in the whole run.** That is the failure mode, and it is a *pass-shaped* value if you only read the exit code — which is why it is grepped for explicitly rather than inferred.
- **`config-control` equals `plugin` on every one of 38 targets.** A mismatch would mean the guard ran and proved nothing.
- **`scripts/validation/tests` — the exact target named in cg#445 — reports `config-control=4 plugin=4`.** That target is where the incident was observed.

(The `control=0 plugin=0` rows are the non-pytest shell targets where the plugin never loads. Consistent, not vacuous.)

### Scope, stated rather than implied

- Gated at `74b427d5 + 21176d1d`. `main` can move before this merges; the two commits gated here are what was measured.
- Round 4s outstanding item — `gate.sh` never having run at `21176d1d` — is closed by this run, and on the merged tree rather than the branch.
- ⚠ Still true and worth carrying: **which of the three mechanisms caused the original 2026-08-30 red is not recoverable.** No discriminator was recorded at the time, and the string was identical for all three. The fix is verified against the mechanism it addresses; it is not proof about that specific incident.

PR #1149, opened 2026-08-31
feat(dl-router): a player rule may declare an ORDERED LIST of media accessors, so one rule covers an anchor-linked image AND a direct video

ZacxDev:
`media` was a single `{element, attr}` pair, and **`attr` is one name**. That makes a site posting both media kinds inexpressible:

| rule | image | video |
|---|---|---|
| `element="img, video"` `attr="src"` | ⚠️ saves the downscaled proxy copy | ✅ |
| `element="a"` `attr="href"` | ✅ original | ❌ dead |

Comma groups were already allowed, so the limit was never the selector — it was `attr`.

## What changes

`media` now also accepts an **ordered list**, tried in order, first http(s) hit wins:

```toml
media = [
  { element = 'a[href^="https://cdn.example-cdn.test/attachments"]', attr = "href" },
  { element = "video", attr = "src" },
]
```

The image resolves through its anchor — the original **by construction**, so no URL is rewritten and nothing has to reason about signatures.

**Deliberately not done:** reaching the original by swapping a proxy URL's host. The two hosts commonly carry *different* signature parameters, so a rewritten URL drops a signature belonging to the other host.

## Two languages, because the sidecar is the first gate

`config.py` validates `site_rules` **before** the snapshot reaches the extension, so a list it rejected could never render a button whatever `player_buttons.js` accepted. `MAX_MEDIA_ACCESSORS = 8` is mirrored on both sides — and is now **pinned by a test that reads the JS source**, because bumping only the Python side left both suites fully green.

## Back-compat is a contract, not an accident

A single table still loads and normalises internally to a one-entry list, so `readMediaUrl` walks one shape. Both forms are asserted side by side. The single-table and list error messages are each pinned **whole**, which is what makes "identical except for the index" checkable rather than prose.

One malformed accessor rejects the **whole** rule. Keeping the good ones would render a button silently covering only *some* media on the page.

🔴 **Deploy order is asymmetric and is now documented in `SKILL.md`.** The extension loads unpacked from the working tree (git pull + full Brave restart); the sidecar runs from the nix store (switch only). Writing a list rule while the old sidecar is live is `ConfigError` → `load_degraded` → **every routing endpoint 503**. Switch first, write the rule second; remove the rule before rolling back.

## Verification

| control | result |
|---|---|
| node | **529 → 537** |
| python (dl-router) | **1018 passed** |
| **RED at the PR base** | **8** node failures, **5** python failures |
| labelled invariant guards | two back-compat tests **pass on pre-change code** — guards, not regression coverage |
| mutation sweep, round 1 | **6/6 killed**, each by its own named test |
| mutation sweep, audit fixes | **3/3 killed**, incl. the accumulate-nodes mutant that **survived** round 1 |
| sandbox tiers | re-running on the fix commit |

🔴 **The reversal mutant is killed only by the `ORDER decides` control.** The image test alone passes under it, because reversing also finds the anchor.

### Two corrections to the original body

1. It said "node 508 → 536". The baseline at this PR's own base is **529**, not 508 — I carried the number from a different PR. It is **529 → 537**.
2. It said the single-table error message is "byte-identical". That holds for **four** messages (empty `attr`, empty `element`, missing `attr`, non-string `attr`) and **not** the fifth: the wrong-*shape* message deliberately changed to mention the list form.


ZacxDev:
**Round 1 — no 🔴, five 🟡, five 🟢.** Fixed in `8921d6fc`. I reproduced the two headline findings on my own tree before touching anything.

**The two that mattered:**

**A hand-copied constant in two languages, unpinned.** `MAX_MEDIA_ACCESSORS = 8` lives in both `config.py` and `player_buttons.js`. Measured: **bumping only the Python side to 9 leaves both suites fully green** (48/48 python, 536/536 node). The sidecar would then accept a 9-entry rule `normalisePlayerRule` rejects — no button on that host, no error anywhere. This repo has paid for this exact shape before: `fixtures/name_cases.json` exists because two hand-copied cross-language tables agreed with each other while the implementations disagreed on **991** inputs. Now pinned by a test that reads the JS source.

**An untested guard that stops a crash loop.** `load_degraded` catches `ConfigError` and nothing else, and the unit is `Restart=always`/`RestartSec=10`. Without the `isinstance(acc, dict)` check, `media = ["video"]` raises `AttributeError`, which propagates — six restarts a minute with nothing listening, the precise failure `load_degraded`'s docstring exists to prevent. Deleting the guard left 48/48 green.

Also fixed: the test TOML emitter **stringified every scalar**, so `element = 1` reached the validator as `"1"` and was accepted — type assertions through it would have passed vacuously; the `readMediaUrl` node-scoping invariant was unasserted (an accumulating variant survived 536/536 and returned a URL belonging to *neither* pair); the wrong-shape message rendered a literal `{{ … }}` nobody can paste; and the message assertions were substrings, now pinned whole.

**`SKILL.md` — the highest-value non-code finding.** The deployed operator/agent doc still documented only the single-table form, so an agent told to "add a site rule" would keep emitting rules that cannot cover both media kinds. The feature was unreachable from the file that routes to it. Now documented, with the two new silent-null causes and the **asymmetric deploy order**: the extension loads unpacked from the working tree, the sidecar from the nix store, so writing a list rule while the old sidecar is live is `ConfigError` → `load_degraded` → **every routing endpoint 503**, and reverting with the rule still in place takes the sidecar down the same way.

**Merged-tree verification is now complete** (main `2a357c01` + this PR's 6 files, disjoint paths): node `536/536`, `scripts/dl-router/tests` `1013 passed`, `scripts/tests` **`10952 passed, 1 skipped, 0 failed`** — cross-read four ways (summary line, independent `--collect-only` at 10953, zero `FAILED`/`ERROR` lines, no timeout panic), with a negative control confirming the instrument can go red.

**Verification of the fix round:** node 536 → 537, python (dl-router) 1013 → 1018, skill-size gates 61 passed. Mutation sweep **3/3 killed, each by its own named test**, including the accumulate-nodes mutant that **survived** round 1.

**Two claims in the original PR body were wrong and are corrected there:** the node baseline is **529**, not 508 (carried from a different PR), and "byte-identical" holds for four messages but not the wrong-*shape* one, which deliberately changed.

```audit-claims round=1 audited=78442146..8921d6fc
1. MAX_MEDIA_ACCESSORS is now pinned across the two languages by a test that reads player_buttons.js and compares to config.MAX_MEDIA_ACCESSORS.
2. A list entry that is not a table is asserted to raise ConfigError rather than AttributeError, and load_degraded is asserted to return a config.
3. The test TOML emitter preserves bools, numbers, dicts and nested lists instead of stringifying every scalar; a test feeds non-strings through it.
4. readMediaUrl declares `nodes` inside the accessor loop; a new test pins that each accessor sees only its own nodes.
5. The wrong-shape ConfigError message renders single braces, and a test asserts the example is pastable TOML.
6. The single-table and list error messages are each pinned as whole normalised strings instead of substrings.
7. SKILL.md documents the list form, the one-bad-accessor and over-cap silent-null causes, and the asymmetric deploy/rollback order.
8. The PR body's node baseline was corrected 508 -> 529, and the byte-identical claim was narrowed to the four messages it holds for.
```

ZacxDev:
**Round 2 — two 🟢, no blockers.** Fixed in `c3ebaf01`. Both reproduced on my own tree before I touched anything.

**G1 — my round-1 fix covered one of three branches.** `_toml` routed only the array-of-inline-tables branch through `_toml_value`; the other two kept stringifying. So the inversion survived on the **single-table form — the back-compat shape every deployed config actually uses.** Measured side by side:

```
single: element = "1"   attr = "True"
list  : element = 1     attr = true
```

A type-rejection test written for the single form would have passed while asserting nothing — exactly what the comment two lines above it says the helper exists to prevent, left on the branch the comment didn't cover. One rule, one place: every value goes through `_toml_value` now, and the dispatcher is three lines instead of nine.

**G2 — the doc told operators the loud failure was silent.** `SKILL.md` said *"ONE malformed accessor kills the WHOLE rule, silently — no button, no error."* True of the **extension's** grammar checks; **false** of the **sidecar's** structural checks, which this same PR improved. Measured:

```
media = {element="video", attr=""}            -> RAISED: media.attr must be a non-empty string
media = [{...ok...}, {element="a", attr=""}]  -> RAISED: media[1].attr must be a non-empty string
```

An operator reading "no error" stops looking and never sees the message that names the offending index — the very affordance `test_a_bad_entry_in_a_list_names_ITS_INDEX` exists to provide. A confidently wrong doc is worse than a silent one. It now splits the two cases: type/shape errors are caught **loudly** by the sidecar and surface in `dl-route status`; only a **grammar** error the sidecar deliberately does not check is the silent one.

**Verification:** python (config) 53 → 55; node 537 unchanged (no JS this round). The second new guard is a guard **on** the guard — it asserts the emitter's output directly, so a future re-stringification fails there rather than silently making some other test vacuous. Mutant (re-stringify the scalar branch) → **7 tests fail**, including both new ones.

**Ladder note:** this round changed **zero executable payload lines** — `SKILL.md` is prose, `test_config.py` is scaffolding. Round 2's fix had 5. One more scaffolding-only round and the attribution gate says the ladder has left the PR.

```audit-claims round=2 audited=8921d6fc..c3ebaf01
1. _toml now routes EVERY value through _toml_value; the three-way branch is gone and only dicts are still handled separately as sub-tables.
2. A new test asserts the SINGLE-table form refuses non-string element/attr, which the stringifying emitter previously made impossible to test.
3. A new test asserts the emitter's own output preserves ints, bools and nested lists while still quoting strings.
4. SKILL.md's "Buttons don't appear" item 4 now distinguishes sidecar-caught type/shape errors (loud, names the index, visible in dl-route status) from extension-caught grammar errors (silent).
5. The over-cap bullet was folded into the same item rather than claiming its own silent failure.
6. No JS and no config.py behaviour changed this round.
```

ZacxDev:
**Round 3 — no findings. Ladder closed.**

All six claims verified against the diff rather than accepted:

- **The emitter collapse is measured behaviour-preserving.** Both versions of `_toml`/`_toml_value` were AST-extracted from `8921d6fc` and `c3ebaf01` and run over **all 20 fixtures the pre-existing tests feed `_load`** — emission byte-identical on **0 of 20 changed**, every document still parsing under `tomllib`.
- **Both new tests are watched-red at base**, spliced onto the round-2 test file: `element = "1"` / `attr = "True"` then, `element = 1` / `attr = true` now. The round-2 defect was real, not a story.
- **6-mutant battery with a positive control.** Stringifying the int/float branch, the bool branch, the list branch, or `_toml`'s own else-branch each kills the new guards. The one branch the "EVERY branch" name doesn't reach (inline tables) kills five *other* tests loudly, so it is not a coverage hole.
- **The doc's two-bucket claim is measured true on both validators** for every input named — unsupported selectors, `attr = "a.b"`, a 301-char element, a 9-entry list, a non-table entry — plus the **300-char boundary**, confirming no off-by-one. The "surfaces in `dl-route status`" half holds by source (`server.py:218` puts `configError` in the `/healthz` body; `dl-route:140` prints it).

**One correction to my own ladder accounting:** I said round 2's fix was the second zero-payload round. It is the **first** — round 1's fix did touch payload (`config.py` 5 lines, `player_buttons.js` 8). So the attribution gate never fired; the ladder stops on the **clean-round rule**, which is the stronger reason anyway.

**Ladder summary:** round 1 → 10 findings · round 2 → 2 · round 3 → clean. Both fix rounds were caught introducing something: first a hand-copied cross-language constant and an untested crash-loop guard, then my own emitter fix covering one of three branches and a doc that called the loud failure silent.

**Remaining gap, stated plainly:** every green above is the **dev-host tier**. The sandbox tier on this exact commit, and on the merged tree, is still outstanding — four other sessions have been running `checks.x86_64-linux.pytests` concurrently and adding a fifth is the contention CLAUDE.md says produces false failures. Re-running both as soon as the store drains; Tekton's own green is a claim about the branch, not the merged tree (`strict: false`).

PR #1157, opened 2026-08-31
fix(audit-pr): the ladder's escape hatch must state its rationale IN THE SUMMARY — and pin it

ZacxDev:
Found by the **never-run round 3** of #1133's own audit ladder (the handoff's rank 3). Round 3 returned 4 🟡 + 1 🟢; **this PR fixes the only one live in `main`.**

## The defect

#1133's round-2 fix reworded the prose-payload escape hatch in `claude/skills/audit-pr/SKILL.md`:

```diff
-**Stop on this, stated in the round's summary and not left implicit:** no 🔴 · no blast radius
+**Stop on this — but ONLY once you can NAME why the rounds will not stop on their own; read the
```

It **added** the NAME precondition and **deleted** the summary obligation in the same edit — wider on one axis, narrower on another, the exact shape this skill tells auditors to hunt for.

"Can NAME" is a private mental state; no reader can check it. This clause is the one criterion that ends a ladder **without** a clean round, so with the artifact gone, a report that *ended* the ladder on the escape hatch is indistinguishable from one that *converged* — same findings section, same verdict, same ledger line. Those mean opposite things: "no findings remain" vs "real 🟡s are deliberately unfixed".

Verified the requirement was restored nowhere else: one `round's summary` hit in `main`, and it is the ledger line, a different obligation.

## Why the pin, not just the reword

This escape hatch was **the only stop-related clause in the file that nothing asserted on**. The findings-keyed rule beside it carries five pins (`SKILL_HEADING`, `SKILL_STOP_BODY`, `SKILL_NOT_A_CAP`, `SKILL_VERDICT_NOT_STOP`, `SKILL_NOT_WASTE`), so a reword there fails loudly — a reword here was silent. Fixing the prose alone regrows the drift.

Pinned as a **whole normalised string** per `claude/RULES.md` (spelled-guards): the artifact is prose, so a keyword guard is walkable by rewording, which is precisely what happened.

## Controls

- **Red at base / green at HEAD.** Reverting the sentence to its pre-fix wording fails `test_the_prose_escape_hatch_demands_its_rationale_IN_THE_SUMMARY` with **that guard's own message** (`the prose-payload escape hatch's summary requirement is not present exactly once as written (found 0)`) — **1 failed, 11 passed**, so the kill is isolated and not another guard's error.
- **Dev-host tier:** 316 passed across the five modules that read this skill.
- **Sandbox tier — the one the merge gates on**, built one at a time: `devrc-pytests RESULT: PASS (exit=0)` (1466 log lines), `devrc-nodetests RESULT: PASS (exit=0)` (8785 lines), 0 timeouts. Base `origin/main` at `74b427d5`.

## NOT fixed here — acceptance criteria for the deferred port-back

Round 3's other four findings are against `scripts/audit-dispatch.py`'s `nix log` fallback, which **no longer exists in `main`**: #1117 (`1f0fe4c1`) rewrote `render_toolchain` 2h36m after #1133 merged, carrying forward only `--no-link -L` and explicitly deferring the rest to "its own PR and its own round". Verified: `git show origin/main:scripts/audit-dispatch.py | grep -c path-info` → **0**.

Recording them so the port-back does not reintroduce them:

1. **F1** — do not carry the parenthetical claiming `nix path-info` stderr is discarded by the `>` two lines down. Measured: that `>` belongs to a different command and redirects stdout; the flake error lands on the terminal. It also contradicts the same block's correct "`-L` WRITES TO STDERR" rule 28 lines above.
2. **F2** — do not carry the fixed `/tmp/tier.log`. Two parallel audit agents truncate each other's file and each greps the other's tier, with all three guards passing. Use `mktemp`.
3. **F3** — add a structural guard for the guard set (`_assert_the_nix_log_fallback_carries_its_three_guards`). Round 3 measured the existing suite **cannot see those lines**: deleting the `[ -n "$DRV" ]` emitter left the run byte-identical (301 passed), while a `.pytests`→`.PYTESTS` positive control on the same emitter went red.
4. **F5** — "THEY CLOSE THREE DIFFERENT FALSE GREENS" is wider than the bullets deliver; the list supplies two.

⚠ This PR's own fix is **unaudited** — an audit fix resets the verification gate.

### DIFF
--- a/claude/skills/audit-pr/SKILL.md
+++ b/claude/skills/audit-pr/SKILL.md
@@ -211,11 +211,21 @@ what the gate exists to catch, because "fixed a defect" and "reworded a warning"
 edit. Left alone it does not terminate. Measured on devrc #1111: round 2's findings were mostly
 about text round 1 had written, and the gate could not see it.
 
-**Stop on this — but ONLY once you can NAME why the rounds will not stop on their own; read the
-next paragraph before acting on it.** No 🔴 · no blast radius beyond "the document contains a
-false sentence" · and the recurring SHAPE swept at every site rather than at the one that was
-reported. Record what you are NOT fixing, on the PR, so the next reader knows it is open rather
-than absent.
+**Stop on this — but ONLY once you can NAME, IN THE ROUND'S SUMMARY AND NOT LEFT IMPLICIT, why
+the rounds will not stop on their own; read the next paragraph before acting on it.** No 🔴 · no
+blast radius beyond "the document contains a false sentence" · and the recurring SHAPE swept at
+every site rather than at the one that was reported. Record what you are NOT fixing, on the PR,
+so the next reader knows it is open rather than absent.
+
+🔴 **WRITING IT DOWN IS THE WHOLE POINT, AND AN EARLIER REWORD DELETED IT.** "Can NAME" is a
+private mental state; a reader cannot check it. Without the rationale in the summary a report
+that ENDED the ladder on this escape hatch is **indistinguishable from one that converged** — the
+findings, verdict and ledger line all look the same — so an operator cannot tell "no findings
+remain" from "real 🟡s are deliberately unfixed". Those are opposite meanings. Measured: this
+requirement was dropped by `#1133`'s round-2 fix, which added the NAME precondition and removed
+the summary obligation in the same edit — **wider on one axis, narrower on another**, the exact
+shape this skill tells you to hunt for. It is pinned by
+`scripts/tests/test_audit_ladder_stop_rule.py` now because nothing pinned it then.
 
 ⚠ **THIS DOES NOT OVERRIDE THE FINDINGS-KEYED STOP RULE, AND IT IS NOT A LICENCE TO STOP ON A
 ROUND THAT FOUND THINGS.** That rule still governs: a round returning findings that needed fixing
--- a/claude/skills/audit-pr/SKILL.md
+++ b/claude/skills/audit-pr/SKILL.md
@@ -234,11 +234,13 @@ findings, verdict and ledger line all look the same — so an operator cannot te
 remain" from "real 🟡s are deliberately unfixed". Those are opposite meanings. Measured: this
 requirement was dropped by `#1133`'s round-2 fix, which added the NAME precondition and removed
 the summary obligation in the same edit — **wider on one axis, narrower on another**, the exact
-shape this skill tells you to hunt for. 🔴 **This paragraph sits BELOW the ⚠ one on purpose:
-the sentence above says "read the next paragraph", and that pointer means the ⚠ caveat.** An
-earlier draft of this very fix inserted this history here and silently re-pointed it at itself —
-found by the next round. The criteria sentence and the ⚠ caveat are both pinned by
-`scripts/tests/test_audit_ladder_stop_rule.py`; this paragraph is not.
+shape this skill tells you to hunt for. 🔴 **This paragraph sits BELOW the ⚠ one on purpose: the
+"read the next paragraph" pointer is in the CRITERIA paragraph two above, and it means the ⚠
+caveat.** An earlier draft of this very fix inserted this history between the two and silently
+re-pointed it at itself — found by the next round. Both of those paragraphs are pinned WHOLE by
+`scripts/tests/test_audit_ladder_stop_rule.py`, and their ADJACENCY is asserted separately,
+because both pins pass while a paragraph sits between them. This paragraph is pinned by nothing —
+edit it freely.
 
 ## Mutation testing: deletion-mutants are the EASY half
 
--- a/claude/skills/audit-pr/SKILL.md
+++ b/claude/skills/audit-pr/SKILL.md
@@ -217,16 +217,6 @@ blast radius beyond "the document contains a false sentence" · and the recurrin
 every site rather than at the one that was reported. Record what you are NOT fixing, on the PR,
 so the next reader knows it is open rather than absent.
 
-🔴 **WRITING IT DOWN IS THE WHOLE POINT, AND AN EARLIER REWORD DELETED IT.** "Can NAME" is a
-private mental state; a reader cannot check it. Without the rationale in the summary a report
-that ENDED the ladder on this escape hatch is **indistinguishable from one that converged** — the
-findings, verdict and ledger line all look the same — so an operator cannot tell "no findings
-remain" from "real 🟡s are deliberately unfixed". Those are opposite meanings. Measured: this
-requirement was dropped by `#1133`'s round-2 fix, which added the NAME precondition and removed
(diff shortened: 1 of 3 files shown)


ZacxDev:
Round 1 (blind, full) returned 3 🟡 + 3 🟢 and refuted one PR-body claim by measurement. Fixes are in `18131a61`.

```audit-claims round=1 audited=a8e9e50a..18131a61
1. The `🔴 WRITING IT DOWN IS THE WHOLE POINT` paragraph was moved BELOW the ⚠ NOT-A-LICENCE caveat, so the preceding sentence's "read the next paragraph" resolves to the caveat again.
2. A new test `test_the_escape_hatch_pointer_resolves_to_the_NOT_A_LICENCE_caveat` asserts the criteria paragraph is IMMEDIATELY followed by the caveat, as an adjacency check under the pins' own normalisation.
3. `SKILL_PROSE_ESCAPE_STOP` was extended from ending at "acting on it.**" to cover the whole paragraph including the stop CRITERIA tail and the "Record what you are NOT fixing" obligation.
4. The comment claiming the hatch was "the only stop-related clause in the file that nothing asserted on" was corrected; the false version is recorded in place rather than deleted.
5. A new constant `SKILL_HATCH_NOT_A_LICENCE` pins the ⚠ caveat, which was measured unpinned.
6. `MIN_TESTS` in `scripts/tests/mutants-audit-ladder.sh` was re-derived 10 → 12 against the module's current 13 tests, with a note to re-derive rather than remember.
7. The "four commits downstream" figure was replaced with 22 commits on main, of which one (#1117) touched an audit-pr surface.
8. The `POS unmutated copy ... 11 passed` harness-control rows and the "current 11-test module" note were updated to 13, with the historical 11s left as history.
```

ZacxDev:
Round 2 (blind delta) returned 2 🟡 + 2 🟢. Fixes are in `21a98263`.

```audit-claims round=2 audited=18131a61..21a98263
1. `SKILL_HATCH_NOT_A_LICENCE` was extended from the caveat's first two sentences to the whole paragraph, ending at "...why the rounds will not stop on their own."
2. The two clauses measured SURVIVING at 18131a61 — inverting "not a shortcut out of a converging one", and "If in doubt, run the next round" -> "If in doubt, STOP" — were re-run against 21a98263 and both now fail.
3. `SKILL.md:240` no longer says the criteria sentence and the caveat "are both pinned"; it now says both are pinned WHOLE and that their adjacency is asserted separately, and states that the history paragraph itself is pinned by nothing.
4. The M5 row in the mutation matrix was re-measured 9 -> 11 failed, and the sentence claiming the per-mutant counts were unchanged was narrowed to name M5 as the exception, with M2 re-measured as a control at 2.
5. `test_the_escape_hatch_pointer_resolves_to_the_NOT_A_LICENCE_caveat` now splits the file into paragraph blocks and asserts the caveat's block index is the criteria block's index + 1, instead of comparing whitespace-normalised text.
6. The three unscoped "393 passed / 0 failed" figures were replaced with module-scoped counts re-derived at a8e9e50a, and the "40 lines below" cross-reference was corrected to 35.
```

ZacxDev:
🔴 **Correction to the `audit-claims round=2` block above — item 3 was FALSE when I posted it.**

It said "`SKILL.md:240` no longer says the criteria sentence and the caveat 'are both pinned'". It still did. `21a98263` contains **one file**, the test module; the prose edit never landed.

**Mechanism, since it is reusable:** the SKILL.md edit was made *before* the mutation controls ran, and each control restored its mutation with `git checkout -- claude/skills/audit-pr/SKILL.md`. That restores to the last **commit**, so it reverted the mutation and the uncommitted fix together, silently. `claude/RULES.md` says to restore from a `cp -a` copy rather than `git checkout --` for exactly this reason.

**Nothing detected it.** The suite was green either way (that prose is not pinned), the commit succeeded, and `git log` showed what I expected. It surfaced only from grepping the **merged tree** for the sentence I believed I had written — a content check, not a status check. Same shape as this repo's standing rule that a deploy reporting success is a claim about the deploy, not the consumer.

Now landed as `e764c9c4`, which also fixes the "the sentence above says 'read the next paragraph'" nit (the pointer is two paragraphs up).

⚠ **Round 3's audit is in flight against `18131a61..21a98263` and therefore cannot see `e764c9c4`.** If it reports SKILL.md:240 as unfixed, that finding is correct for its range and already closed at head. The next round's range is `21a98263..e764c9c4`.

ZacxDev:
## 🔴 LADDER STOPPED ON THE STATED CRITERION — not on a clean round

Recording this **in the round's summary and not left implicit**, which is the obligation this PR exists to restore. Operator decision, taken with the ledger below in front of them.

**Round 3 returned findings (2 🟡 + 4 🟢), so this is NOT the findings-keyed stop.** It is the prose-payload escape hatch, and the criterion requires naming why the rounds will not stop on their own.

### Why they will not stop on their own
The payload here is prose plus test constants. Every round's fix writes new prose — comments, failure messages, commit messages — and that prose becomes the next round's audit surface. Measured on this PR: round 1 found a defect in the round-0 fix's prose; round 2 found a defect in round 1's fix, *four lines under the banner forbidding it*; round 3 found that round 2's correction got its own history wrong. **"Fixed a defect" and "reworded a warning" are the same edit here**, so the attribution gate cannot distinguish them, and it does not fire: payload per round is 23 → 0 → 12 lines, never two consecutive zeroes.

### The criterion, item by item
- **No 🔴** — none in any of the three rounds.
- **Blast radius bounded to "the document contains a false sentence"** — the shipped guards are verified working; every open item is a comment that reads as more precise than it is.
- **The recurring SHAPE swept at every site, not just the reported one** — the shape is *a number in prose that no assertion reads*. Round 3 named the unswept sibling (the `PREV` row); `ec249258` swept it, re-derived the whole matrix, and marked the `BASE` row **NOT REPRODUCIBLE** rather than inventing a value for it.

### What is NOT fixed, so it reads as open rather than absent
- **🟢 F6** — `test_audit_ladder_stop_rule.py:472` and `:1054` carry three fresh **unpinned SKILL.md line numbers** (`:214`, `:249`, `:219`, and the derived `35`). All correct today; any paragraph inserted above line 214 makes them wrong with a green suite. Deliberately not fixed: each quotes its target verbatim beside it, so a reader recovers by text search, and pinning line numbers is the practice this module argues against. **Closing condition:** they are corrected or deleted in whatever commit next edits that region of `SKILL.md`.
- **The `BASE` row** is now honestly marked unreproducible rather than carrying a number. **Closing condition:** it names a fixed sha, or it is deleted.

### What the ladder actually bought
Four defects that reached `main` or would have: the escape hatch's summary obligation deleted by a reword (live in `main` since #1133); a pin covering 2 of 6 sentences so the only clause forbidding the hatch on a *converging* ladder could be inverted green; an adjacency instruction silently re-pointed at itself; and three mutants that justified a pin living in a comment instead of the executable battery (18 → 21 rows).

PR #1170, opened 2026-08-31
fix(index-store): ONE append protocol — retire the y/N at the last door, and stop hardcoding created_by

ZacxDev:
## The problem

Two skills write `~/.claude/analyze-service-index/`, and each carried a **full protocol**. They **materially conflicted**, and neither named a winner:

| door | prompt | mechanism |
|---|---|---|
| `claude/skills/subsystem-index/SKILL.md` (`/handoff`) | y/N **RETIRED** | `Edit` anchored on `## Nuance / work-history` |
| `claude/skills/analyze-service/reference/write-back.md` (`/analyze-service`) | `append this to the index? (y/N)` | "plain `Write`" |

They disagreed about the one operation that can destroy another session's bullet: a whole-file retype is MEASURED to lose a concurrent append silently (`subsystem-index/reference/index-write.md`, the two-session simulation). Both read as *the* protocol.

## The decision (operator, 2026-08-31)

Retire the y/N everywhere. **One protocol, one document, zero forks.**

## What changed

**`write-back.md` — protocol DELETED, not reworded.** It keeps only what is `/analyze-service`-specific and was never duplicated: what counts as notable, auto-discovered pointers, bloat discipline — plus the two **caller** facts the shared protocol leaves to its caller (its own `created_by: analyze-service` stamp, and where the file goes). Step 4 now routes to `subsystem-index/SKILL.md`. The pointer paragraph sits in a hash-pinned region.

**`subsystem-index/SKILL.md` — the 🔴 "the two MATERIALLY CONFLICT / reconciling them is open work" block becomes a dated record** of what the fork was and how it closed. The history is kept deliberately: a reader who half-remembers the conflict and finds no trace of it re-derives it — or "restores" the deleted copy to fix a protocol that isn't missing.

**`created_by` is no longer hardcoded.** `subsystem-index/SKILL.md`'s own text said `created_by: handoff` "is correct only while this file has one caller". It now has two.

- `new_entry_template(...)` takes `created_by` as a **required keyword** (`TypeError` without it).
- `--template` **refuses** without `--writer {analyze-service,handoff}` — exit 2, naming the choices — rather than defaulting. `created_by:` is stamped once at creation and never edited, so a guessed default is *permanently* wrong for one caller and corrupts the split. `--writer` has **no default**, and a test asserts that.
- Both printed `--template` hints now carry `--writer`, from one `WRITER_PLACEHOLDER` derived from `KNOWN_WRITERS` (not three hand-typed copies).
- `report_json`'s `writer_id` → `known_writers`. A probe run stamps nothing, so naming one writer was a claim the report had no standing to make.

**`--census` needed NO change** — it already buckets by whatever `created_by:` value is present and never folds an unknown one into a known writer. That is asserted, not assumed (`test_a_writer_the_tuple_does_not_KNOW_is_still_counted_separately`).

**Six now-false claims corrected.** A comment is a claim too: `subsystem_touch`, `subsystem_recall`, `service_recon`, `subsystem-audit` all described the append as "confirm-gated"; `handoff/reference/write-gate.md` and `resume/SKILL.md` stated the retired prompt in the present tense. `index-store.md` gained a note for the stale "*confirmed* write-back" wording that sits **inside** the resolver's hashed region and therefore could not be edited in place.

**`prune-index` KEEPS its y/N and its `Write`** — stated explicitly, with the reason, rather than left as an unexplained inconsistency: a prune is a **deletion** that rewrites a whole file; the evidence that retired the append prompt ("the answer was always `y`") was measured on an **append**; and a cut removes bytes that are often their content's only copy. Its false "same contract as `analyze-service`'s write-back" pointer is retargeted at its own contract doc.

## New guard — `scripts/tests/test_index_append_protocol.py` (56 tests)

Four independent layers, because each alone has a known hole:

1. **Substring pins** on each door — catches deletion of a named instruction and drift in its wording.
2. **A LEDGER, not a component check.** The exhaustive set of `claude/skills/**` docs that state the append mechanism, failing when it **grows or shrinks**, each entry carrying its reason. With a positive control that the probe can actually observe a new carrier — a reassuring set is otherwise indistinguishable from a scan wired to nothing.
3. **A whole-region sha256** over the pointer block, whitespace-normalised so re-wrapping is free but a changed *word* is not. A substring pin over prose is walkable by rewording; this is what closes that.
4. **The `created_by` parameterisation end-to-end**, including the template→census **seam**: a value written by one must be counted by the other under the same name.

🔴 **Every fixture uses `analyze-service`, never `handoff`.** `handoff` *was* the hardcoded value, so a fixture spelling it cannot distinguish "the argument was written" from "the old default was" — it would survive the exact regression the class exists to catch. Three pre-existing tests were moved off `handoff` for the same reason.

## Verification

**Red at base / green at HEAD** — base tree extracted clean via `git archive origin/main` into a scratch dir with the new test planted, so nothing in my worktree was mutated to measure it:

| tree | result |
|---|---|
| `edbc596f` (base) | **24 failed**, 29 passed, 3 skipped |
| `f2fe2b38` (HEAD) | **56 passed** |

Sample guard error messages at base:

```
AssertionError: the set of skill documents stating the append mechanism has changed.
    NEW carriers (a second protocol is growing): none
    LOST carriers (the protocol has no home):    ['claude/skills/analyze-service/reference/write-back.md']

AssertionError: marker '<!-- one-append-protocol:begin' is missing from write-back.md

Failed: DID NOT RAISE TypeError
```

**Gate tiers** — see the comment below for the merged-tree run and the sandbox tier.

## Pinned sentences moved

- `test_subsystem_touch.py::TestSkillDocsArePinned` — `"created_by: handoff"` still resolves in `subsystem-index/SKILL.md`, now inside the sentence explaining that `--writer handoff` is what puts it there. Still asserting the right thing: `/handoff`'s own stamp value reaches the written file.
- `test_analyze_service_stamps_itself_on_a_new_file` — `"stamp \`created_by: analyze-service\` in the front matter"` survives in `write-back.md` as a **caller** fact. (It broke once mid-work purely on a markdown line wrap; the paragraph is reflowed so the pinned sentence stays on one line.)
- `test_json_mode_is_machine_readable` — `writer_id` pin replaced by `known_writers`, with the reason in the test.
- `test_repo_path_guard.py` — two `--template` invocations gained `--writer handoff` so they still reach the repo-path check they guard, rather than stopping at the new exit-2 refusal.

## Store safety

🔴 **Nothing here touched `~/.claude/analyze-service-index/`.** No git command was run inside it, no entry content was read into this repo, and every fixture in the new test file is synthetic under `tmp_path`. The new module asserts this about itself.

## Not merged

Left open per instruction. Note `#1132` (`fix/index-store-backup-claim`) was still OPEN at the time of the merge below and touches both doors — see the comment for the conflict I expect and how it should resolve.


ZacxDev:
## Gate verdicts — both tiers, merged tree

Tree asserted **clean** before each run and `HEAD` + dirtiness re-read after each. All four runs are against the **merged** tree `ff146a82` (`fix/one-index-append-protocol` ← `origin/main` @ `c2daa65d`, #1163).

| tier | command | verdict |
|---|---|---|
| dev-host, pre-merge `f2fe2b38` | `nix develop <worktree> --command bash scripts/gate.sh` | `GATE: RESULT=PASS exit=0` — pytest `19996 passed / 3 skipped / 0 failed` (floor 18383), node `1449 pass / 0 fail` (floor 1367) |
| **dev-host, merged `ff146a82`** | same | `GATE: RESULT=PASS exit=0` — pytest `19996 passed / 3 skipped / 0 failed`, node `1449 pass / 0 fail` |
| **sandbox, merged `ff146a82`** | `nix build .#checks.x86_64-linux.pytests --no-link -L` | `BUILD_RC=0`, `devrc-pytests> RESULT: PASS (exit=0)`, `collected=19999 passed=19996 failed=0`, `panic: test timed out` count **0** |
| **sandbox, merged `ff146a82`** | `nix build .#checks.x86_64-linux.nodetests --no-link -L` | `BUILD_RC=0`, `devrc-nodetests> RESULT: PASS (exit=0)`, `suites=5 files=41 tests=1449 pass=1449 fail=0` |

The two sandbox derivations were built **one at a time**, never combined — the documented false-failure hazard. `scripts/tests` alone: `collected=11024 passed=11024 skipped=0 floor=10269`.

⚠ **`test_subsystem_store_api.py` did NOT fire this run.** Its documented sandbox instability was checked for explicitly: the four log hits for that filename are a `DeprecationWarning` and three test *names*, not failures — `failed=0`. So this is a clean sandbox run, not a run whose only failure was the known-broken one.

⚠ **Concurrency, stated rather than hidden.** The `nodetests` build ran while two *sibling agents* were building the `pytests` derivation (`devrc-integ-1157`, `devrc-integ-1168`) — different derivation, shared store. Per `CLAUDE.md`, a **green** under contention is trustworthy (a contended run fails loudly, it does not fake a pass); a red would have needed re-running alone. It came back green, so no re-run was needed. The `pytests` build itself ran with **no** other `nix build` live, verified by resolved PID via `/proc`, never a `pkill -f` pattern. Nothing was killed.

## Mutation battery on the new guard

A green suite only covers the mutations you imagined, so three were planted — each on a **fresh `git archive HEAD` extract** in scratch (my worktree was never mutated; `git status` clean throughout), run under `PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider` so no mutant could be scored SURVIVED off stale bytecode.

| mutant | killed by, with its own error |
|---|---|
| **A** — re-add a forked protocol paragraph to `write-back.md`, *outside* the pinned region | `test_the_pointer_INSTRUCTS_no_mechanism_of_its_own[On decline, discard]` and `[then plain \`Write\`]` — 2 failed, 54 passed |
| **B** — a **reworded** fork using *none* of the banned imperatives | `test_the_pointers_only_occurrence_is_INSIDE_the_pinned_region`: *"write-back.md states the append mechanism OUTSIDE its pinned `one-append-protocol` region (2 occurrences in the file, 1 inside the region)"* — 1 failed, 55 passed |
| **C** — re-hardcode `created_by`, at **both** the function default and the CLI default | `test_the_template_REFUSES_without_a_writer`, `test_the_known_writers_ledger_matches_the_CLI_choices`, `test_the_cli_refuses_a_template_with_no_writer` — 3 failed, 53 passed |

Mutant **B** is the one that matters: it is the failure mode a substring pin cannot see, and it is what layer 3 (the whole-region hash) and the region-scoped ledger exist for. Each mutant killed a *different* guard and left the rest green, so none died for the wrong reason.

## Red-at-base matrix (repeated here for the record)

Base `edbc596f` extracted clean via `git archive` with the new test planted — nothing in the working tree was mutated to take the measurement:

- **`edbc596f`: 24 failed, 29 passed, 3 skipped**
- **`f2fe2b38` (HEAD): 56 passed**

## ⚠ Interaction with #1132

`#1132` (`fix/index-store-backup-claim`) was still **OPEN** at merge time, re-checked immediately before. It touches both doors. Expected conflicts and how they should resolve:

- **`write-back.md` step 5** — #1132 edits the line; this PR **deletes** it. Take the deletion. Its factual correction ("the one path off this machine is the encrypted daily backup") is **not lost**: #1132 carries the same correction in `index-store.md` → Store safety, which this PR does not touch.
- **`subsystem-index/SKILL.md` last paragraph** — #1132 replaces "curated, unbacked-up entry" with "curated entry … the hourly commit behind it can be up to seventy minutes stale". This PR does not touch that paragraph, so **take #1132's wording**; do not revert its correction.
- **`subsystem_recall.py` module docstring** — both PRs edit the same paragraph (#1132 the backup clause, this PR the "confirm-gated" clause). Both corrections are wanted; merge them, do not pick one.

Whichever lands second should read the merged result rather than trusting a clean `git merge` — a semantic conflict survives one.

ZacxDev:
## Your red `tekton/devrc-pytests` was NOT this PR — exonerated by control, and the cause is now fixed on `main`

Picked this up as rank 8 (Cairn phase 0) of `claudedocs/handoff-cairn-task-linkage.md`; claimed as `cairn-task-linkage-8`. Flagging what I found rather than sitting on it.

**The failing test was `test_unknown_instance_fails_fast_not_after_cmd_timeout`**, in `scripts/browser-bridge/tests/test_server.py` — a file this PR does not touch, and whose copy here is **byte-identical to `main`**.

The discriminating control, same environment, varying only the code: it **passes on `origin/main` (`4a7d7c79`) and on this PR's head (`ff146a82`)**, in isolation. Not your defect.

**Root cause:** three tests asserted a flat `elapsed < 1.0` against a **5.0s** timeout — 5× tighter than the path they exist to rule out, so they are load detectors, not timeout detectors. Fixed in **#1179** (merged): each bound is now derived as `TIMEOUT / 2`, mutation-verified so it still catches a 3.0s slowdown.

**What that means for you:** this branch still carries the old flaky test, so its checks will keep rolling that die until it picks up `main`. I am updating the branch so the checks re-run against the fixed test.

⚠ **I did not change any file in this PR** — the update is a merge of `main`, nothing else. If you are mid-stream here, say so and I will back off.

ZacxDev:
## Blind adversarial audit, round 1 — two 🔴, five 🟡. Not merged; leaving the call to you.

Ran this because the PR had no audit round and it is rank 8 (Cairn phase 0) of
`claudedocs/handoff-cairn-task-linkage.md`. The auditor got the diff and the checklist, not
my conclusions. **I changed nothing in this branch** beyond the earlier `main` merge.

The prose consolidation is genuinely well done and the code half is careful. But both
headline claims have a measurable hole.

### 🔴 1 — the door this PR exists to fix now has the UNSAFE mechanism as its pre-approved one

`claude/skills/analyze-service/SKILL.md:5` — `allowed-tools: Bash, Read, Write, Grep, Agent`,
**no `Edit`** — while `write-back.md:15` now routes that caller to
`subsystem-index/SKILL.md:161`, which mandates *"use `Edit` anchored on
`## Nuance / work-history`, not `Write`"*.

`allowed-tools` is a pre-approval, not a restriction, so `Edit` is still reachable — the
defect is the incentive. Measured on this host: `~/.claude/settings.json` is
`defaultMode: default` with 210 allow entries and **zero** `Edit`/`Write` entries. So at that
door the *mandated safe* mechanism raises a permission prompt and the *unsafe* one does not —
and `subsystem-index/reference/index-write.md:17` measures the unsafe one losing a concurrent
append *"silently — no error, and nothing about it appears in the diff that was shown"*.

Before this PR the pairing was coherent (`write-back.md` said "plain `Write`", and `Write`
was pre-approved). The PR changed the protocol and not the frontmatter. Every sibling
index-writing skill lists `Edit` (`subsystem-index:4`, `handoff:5`, `prune-index:5`); the
read-only ones deliberately omit both (`resume:5`). In a headless/opencode or subagent run
there is nobody to answer the prompt.

**One-word fix.** No test reads `allowed-tools` anywhere in the repo.

### 🔴 2 — the stated acceptance criterion is not met: a REWORDED second protocol passes all 56 tests

Phase 0 is *"one write path … with a test that fails if a second protocol reappears."*
Measured in a `cp -a` copy: a complete, materially conflicting second protocol appended to
`write-back.md` — carrying both the retired `append this to the index? (y/N)` prompt and the
banned whole-file `Write` — gives **56 passed**.

Controls proving the harness was wired to the mutated tree, not silently zero:

| mutant | result |
|---|---|
| verbatim re-add of the old text | **3 failed** |
| new third door carrying the pinned `MANDATE` **verbatim** | **1 failed** |
| same third door, mandate **reworded** | **56 passed** |
| contradicting carve-out appended to the **owner** | **56 passed** |

Root cause: every layer keys on the exact string an *agreeing* copy would contain, or on ten
exact phrasings of the *old* text. A fork is by definition a document that says something
**different** — so the ledger detects agreement and misses disagreement, and the whole-region
hash covers one block of one file rather than the file or the owner.

🔴 **And this PR's own mutation evidence for that guard does not hold.** The comment's
mutant B is described as "a **reworded** fork using none of the banned imperatives", killed by
a count check reporting *"2 occurrences in the file, 1 inside the region"*. That count can
only move if the mutant reproduced the pinned `MANDATE` **verbatim** — so the mutation was not
isolated to the property under test. `claude/RULES.md`: *isolate the mutation; a green sweep
covers only the mutations you imagined.*

**Fix direction:** a predicate that fires on *disagreement* — any `claude/skills/**` doc other
than the owner containing an imperative near `Write`/`Edit`/`(y/N)` **and** the store path —
or extend the hashed region to all of `write-back.md` below step 4 plus a region on the
owner's write half.

### 🟡 the rest, briefly

3. **`subsystem_touch.py:4587,:4667` print "confirm-gated" twice**, at exactly the two
   decision points this PR governs, read by the agent immediately before it writes —
   contradicting `subsystem-index/SKILL.md:157` *"Write it — no question"*. The PR corrected
   six prose claims under "A comment is a claim too" and left the two that are *printed
   output*. Two-line fix. Seventh instance: `test_subsystem_recall.py:3592`.
4. **`created_by` became an unvalidated parameter that can overwrite the fail-safe
   `sensitivity:`.** Measured: `created_by='evil\nsensitivity: public'` emits both lines, the
   parser takes the last (`public`), and `validate_entry_file()` returns `None` — VALID. Not
   CLI-reachable today (`--writer` has `choices`), but this PR is what turned a constant into
   an input, and `sensitivity` is what phase 4 gates sharing on.
5. **`subsystem-index/SKILL.md:141`** opens the shared write half with *"read the policy file
   the probe named on its `policy:` line"* — `service_recon.py` emits no `policy:` line, so
   the newly-routed caller cannot satisfy it, while `index-store.md` tells that same caller to
   read the scope README. The two now disagree for the caller just moved.
6. **The `Write`-only-for-a-first-ever-file carve-out is unguarded against a race** — the
   "no prior content to lose" justification is a claim about when the agent *looked*, not when
   it *writes*; with two callers on one protocol both can resolve NO ENTRY for the same slug.
7. **Semantic conflict with the still-open #1132** ("the store HAS an off-machine backup — 15
   places said it did not"): this PR adds **new** "only copy" lines at
   `write-back.md:90`, `prune-index/SKILL.md:73`, `subsystem-audit.py:37-41` that merge
   *cleanly* into a document that then says both things. The PR's expected-conflict list names
   three sites and none of these.

### Verification

`scripts/gate.sh --tier both` on the audited head `77490b72` (base `7f1c81c9`), dev-host tier:
`GATE: RESULT=PASS exit=0`, pytest `collected=20005 passed=20002 failed=0`, node `1449/1449`.
Red-at-base re-measured independently rather than accepted: the new test planted on
`7f1c81c9` gives **24 failed / 32 passed**.

⚠ **The merge-gating SANDBOX tier was not confirmed on this head** — it was still building.
The author reports it green on the pre-merge head `ff146a82`; the only delta since is a merge
of `main`.

⚠ **Out of scope, stated so it is not read as coverage:** `scripts/subsystem-store-api/server.py`
carries a **third** append protocol (flock + temp-file-rename, `PUT` behind `If-Match`). It
writes the pod's `/data`, not the local store, so "one protocol" holds locally today — but the
new ledger scans `claude/skills/**/*.md` only and would never see it, and **phase 1 makes the
pod canonical.**

ZacxDev:
## 🔴 1 fixed here (`97d898e5`); 🔴 2 filed as cg#473. Merging once checks re-run.

**🔴 1 — fixed on this branch.** `analyze-service`'s `allowed-tools` now carries `Edit`, the tool this PR's protocol mandates at that door. One word.

Guarded, because **nothing in this repo read `allowed-tools` at all** — no gate could have caught this. `TestTheMandatedToolIsThePreApprovedOne` asserts every writing door pre-approves `Edit`, plus a **negative control** that the read-only door (`resume`) pre-approves neither write tool — without which the guard would also pass on a tree where every skill declared every tool, i.e. would be testing nothing.

Mutation-verified both directions, each isolated to exactly one failure with its own message:

| mutant | fired |
|---|---|
| revert the one-word fix | `test_a_writing_door_pre_approves_the_mandated_Edit[analyze-service]` |
| give `resume` a write tool | `test_the_read_only_door_pre_approves_NEITHER` |

61 passed (was 56).

**🔴 2 — filed as cg#473, not fixed here.** The reworded-fork gap needs a predicate that fires on *disagreement* rather than on the exact string an agreeing copy would contain, and that is design work rather than a patch — doing it inside this PR would put an unaudited second mechanism into the change that exists to remove a second mechanism. The card carries the three mutants as acceptance criteria and requires each to be demonstrated with an **isolated** mutant, which is the specific thing that went wrong in the original evidence.

🔴 **So this PR should NOT be read as completing Cairn phase 0.** The consolidation is real and worth having; the *"test that fails if a second protocol reappears"* half is not met until cg#473 lands. I will record rank 8 that way rather than closing it.

I changed nothing else on this branch — the `main` merge earlier, and this commit.

PR #1178, opened 2026-08-31
docs(handoff): rank 3 closed — #1157 merged, ladder stopped on the stated criterion, port-back filed as rank 4

ZacxDev:
Closes rank 3 of the audit-pr-ladder handoff and files what the ladder left open, so the next `/resume` neither re-does it nor loses the deferred work.

## Rank 3 — done
`#1133`'s never-run round 3 ran blind over `ea36a489..9ff5c9a9`: **4 🟡 + 1 🟢**. Exactly one was live in `main` — the prose-payload escape hatch's "stated in the round's summary and not left implicit" obligation, deleted by round 2's reword. Fixed and pinned in **#1157** (merged `3e4c447f`), which then ran its own three-round ladder.

**#1157 was stopped on the stated criterion, not on a clean round** — rationale recorded on the PR, which is the obligation that PR restores. The other four findings were not actionable against `main`: `#1117` had already removed the code they describe.

## New rank 4 — the port-back, with criteria already measured
`#1117` deferred the `nix log` fallback to "its own PR and its own round" (`grep -c path-info` on `origin/main` → 0). Round 3's four findings are exactly the acceptance criteria for that port-back, and until now they lived only in a PR body — so a port-back author had no route to them. Now in the ranked list with a closing condition.

## Gotchas added
Six, all measured this session — the most reusable being that **`git checkout -- <file>` used to restore a mutation also reverts your uncommitted fix**, silently, while every status signal reports success (green suite, successful commit, expected `git log`). It surfaced only by grepping the merged tree for a sentence I believed I had written.

Docs only — no code, no test surface.

### DIFF
--- a/claudedocs/handoff-audit-pr-ladder.md
+++ b/claudedocs/handoff-audit-pr-ladder.md
@@ -141,27 +141,43 @@ findings-keyed stop rule does not terminate in the guard-hardening regime.
   other "surfaces a worktree does not hand you" in `claude/RULES.md`.
 
 ## Next steps (ranked)
-🔴 **Numbering is deliberately STABLE — rank 3 keeps its number.** The rank is half a
-`claim-work` slug's identity, so renumbering would silently re-point any live claim. Items 1
-and 2 are retained as DONE markers rather than deleted.
+🔴 **Numbering is deliberately STABLE.** The rank is half a `claim-work` slug's identity, so
+renumbering silently re-points every live claim. Closed items are retained as DONE markers.
 
-1. **DONE (2026-08-31) — ship the laptop.** Both hosts converged at `9a7c4338`; `drift-check.sh`
-   rc 0, `PARITY-RC=0` on both. Nothing to do; do not re-claim.
+1. **DONE (2026-08-31) — ship the laptop.** Both hosts converged; `drift-check.sh` rc 0,
+   `PARITY-RC=0` on both. Do not re-claim.
    forcing: none
-2. **DONE (found already closed 2026-08-31) — the three leftover worktrees.** Gone from
-   `git worktree list`, closed by another session. ⚠ **The successor condition is real and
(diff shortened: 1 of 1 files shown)


ZacxDev:
🔴 **The `tekton/devrc-pytests` red on `dde4f05d` was environmental, not this PR.** Diagnosed rather than re-run blind; rebased onto current `main` (`4a7d7c79`) to re-trigger.

**What failed:** 2 tests, both `TestTheBackstopNeverSendsASecondResponse`, in `scripts/tests/test_subsystem_store_api.py` — the file this handoff already documents as a recurring load-flake site.

**This PR changes one file: `claudedocs/handoff-audit-pr-ladder.md`.** It cannot reach store-api.

**🔴 The assertion message describes the OPPOSITE of the failure, exactly as the handoff warns.** It reads *"a SECOND complete response followed the 200 on one connection"* — but the values are `assert 0 == 1` with `raw = b''`. Zero responses were read; nothing second arrived. That is the CLIENT-bound symptom the handoff records `#996` as never having touched. Anyone debugging the message rather than the values goes looking for a pooling-proxy bug that is not there; the handoff says four cycles were lost to exactly that before someone read a pod in time.

**Discriminating controls, run before concluding:**
- The two tests pass **3/3 locally** on this same tree, ~5s each, versus a **464s (7:44)** `scripts/tests` target in CI.
- Positive control: `--collect-only` confirms the `-k` filter *did* select both CI-failing tests — not a filter that quietly deselected them.
- Negative control: breaking one assertion in that class makes the same invocation report `1 failed`, so it can go red.
- `tekton/devrc-nodetests` **passed** in the same run; the cluster was running 4–5 concurrent pipelineruns.

**Not filed as a work item**, deliberately: this is the Tekton capacity condition the handoff already records as open and unowned, with three fixes already rejected on measurements. It has no closing condition anyone can check, so a ticket would read as covered while nothing could close it.

PR #1179, opened 2026-08-31
test(browser-bridge): three timing bounds were tighter than the thing they discriminate — CI load failed them, not the diff

ZacxDev:
## What went wrong

`test_unknown_instance_fails_fast_not_after_cmd_timeout` went **red on devrc#1170** — a PR touching no browser-bridge file, whose copy of this test file is **byte-identical to main**.

**The control settles it.** Same environment, vary only the code: the test passes on `origin/main` (`4a7d7c79`) *and* on #1170's head (`ff146a82`), run in isolation. #1170 is exonerated.

## Root cause, not a re-run

Three sites assert a flat wall-clock literal against a much larger mechanism:

| site | asserts | mechanism |
|---|---|---|
| `unknown_instance` fails fast | `elapsed < 1.0` | `cmd_timeout = 5.0s` |
| `no_extension` fails fast | `elapsed < 1.0` | `cmd_timeout = 5.0s` |
| a healthy ping is fast | `elapsed < 1.0` | `poll_timeout = 5.0s` |

Each is **5× tighter than the path it exists to rule out**, making it a *load* detector rather than a *timeout* detector — under a contended CI node it fails while nothing is wrong.

`366de091` records this exact shape at ten CLI sites: *"a test's safety net must not be tighter than the bound of the thing it invokes."* Known class in this file, so it is swept at **every** site rather than patched at the one that was reported.

Each bound is now derived as `TIMEOUT / 2`, with the timeout a named local instead of a literal buried in the `_serve` call, so the two cannot drift apart. The failure message names both numbers.

## Not widened into vacuity — mutation-verified

A **3.0s** delay in `_send` (under the 5s timeout — a *partial* regression, not just the full timeout path) makes all three fail on their own assertion:

```
AssertionError: took 3.00s against a 5.0s cmd_timeout — it went down a waiting path
```

## Deliberately NOT touched

`test_ping_fast_fails_even_while_BUSY...`'s `< 1.5`. Its comment shows a real discriminator — *"1.5 sits between 0.3 (correct) and 3.5 (no `- age` term)"* — so widening it would blind the test to the mutant it was written for. Same literal shape, different meaning: the sweep reads each site rather than pattern-matching them.

## Gate

461 passed in `test_server.py`, dev-host tier, base `4a7d7c79`. The sandbox tier is this PR's own required checks.

### DIFF
--- a/scripts/browser-bridge/tests/test_server.py
+++ b/scripts/browser-bridge/tests/test_server.py
@@ -2366,7 +2366,14 @@ def test_a_healthy_ping_is_unaffected_and_answers_in_milliseconds(monkeypatch):
     measured healthy pings averaged 3–4 ms, so the 10 s deadline is ~2500×
     headroom — assert the fast path still returns 200, not a timeout."""
     monkeypatch.delenv("BROWSER_BRIDGE_PING_TIMEOUT", raising=False)
-    srv, _ = _serve(cmd_timeout=25.0, poll_timeout=5.0)
+    # Bound DERIVED from the deadline it discriminates, not a literal. `< 1.0`
+    # against a 5s poll_timeout was five times tighter than the thing under
+    # test, which makes it a load detector rather than a timeout detector; see
+    # the 🔴 block on test_unknown_instance_fails_fast_not_after_cmd_timeout.
+    # The measured healthy ping is 3-4 ms, so 2.5s is still ~600x headroom.
+    POLL_TIMEOUT = 5.0
+    FAST = POLL_TIMEOUT / 2
+    srv, _ = _serve(cmd_timeout=25.0, poll_timeout=POLL_TIMEOUT)
     ext = FakeExtension(srv, executor=lambda c: {"pong": True})
     ext.start()
     try:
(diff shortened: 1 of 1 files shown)


ZacxDev:
## Correction to this PR's own description — the exclusion list was incomplete

The body says the class was "swept at every site" and names **one** deliberate exclusion (`test_ping_fast_fails_even_while_BUSY`'s `< 1.5`). There are **two** `assert elapsed < 1.5` sites, and I had only read one when I wrote that.

Verified after merge — the conclusion is unchanged, but it was unverified when claimed:

`test_ping_timeout_is_env_overridable_and_survives_a_malformed_value` (line ~2408) is **correctly bracketed and rightly untouched**. The env override sets `BROWSER_BRIDGE_PING_TIMEOUT=0.3` and the fallback path is `cmd_timeout=8.0`, so `1.5` sits between the correct value (0.3) and the thing it rules out (≥8s) — 5× headroom over the expected duration, and nowhere near the mechanism. It is not an instance of the class this PR fixed, which was specifically **bound tighter than the mechanism** (`1.0` against `5.0`).

So: the three `< 1.0` sites were the whole class, and both `< 1.5` sites are genuine discriminators. The sweep was correct; the *description* of how thoroughly I had checked it was ahead of the checking.

Recording it here rather than editing the body, because the body has already been read.

PR #1180, opened 2026-08-31
docs(handoff): the store-api flake RECURRED — "Fixed by devrc#996" is false

ZacxDev:
The handoff records `test_subsystem_store_api.py` as **"Fixed by devrc#996"**. It has recurred, so that claim is now false — and a flake marked fixed is the most expensive possible wrong reading, because it tells the next person their diff broke it.

## What happened

`devrc#1162` — a **one-markdown-file** PR — was blocked by `tekton/devrc-pytests` on:

```
TestTheActorComesFromTheTOKEN::test_a_FORGED_actor_in_the_body_is_DISCARDED[record0-…]
```

That is the **same case name this doc already records** from devrc#1056.

The two reds on that branch were unrelated, which is what makes the attribution clean:

| head | failing test |
|---|---|
| `e1d1318f` | `test_no_unallowlisted_public_ip_literal_is_committed` — a real defect of mine, since fixed |
| `74e39bea` | `TestTheActorComesFromTheTOKEN…` — this flake |
| `2fd84888` | none — `collected=19942 passed=19939 skipped=3 failed=0` |

## Attribution

**Ruled out — the diff.** #1162 touches exactly one file, `claudedocs/handoff-tmux-webapp.md`. A markdown file cannot reach a store-api test. That is the doc's own stated discriminator ("the discriminator that settled it was a DOCS-ONLY PR failing"), reproduced.

**Control, with its limit stated.** The file ran **3/3 green on a clean `origin/main` worktree — 641 passed each, 301.6s / 293.8s / 296.8s**; the tight spread rules out load inflation in that run. 🔴 But it is weaker than it looks: it ran on the **dev host** while the failure is in the nix **sandbox tier** under CI concurrency, so it samples a different environment than the one that failed. The structural argument is what actually discriminates; the 3/3 only shows the file is not deterministically broken.

Three distinct cases have now been observed failing across four runs, which keeps the original seed/ordering hypothesis alive: #996 narrowed it, it did not eliminate it.

## Why it matters operationally

devrc is the one repo here with `enforce_admins: true` and two required checks, so this blocks arbitrary PRs — docs-only included — and the only remedy is a fresh push.

The added gotcha generalises the lesson: when closing a flake, record what was **measured** (the fix, and the runs that passed) rather than the verdict — "N consecutive green runs in the failing tier" is falsifiable by the next red, "fixed" silently absorbs it.

The next probe is also corrected: stop re-deriving this on the dev host, where it passes. Reproduce in the tier that fails (`nix build .#checks.x86_64-linux.pytests`, one derivation at a time).

Docs-only; one file.

https://claude.ai/code/session_01GwT3r4UwWbvGj3W59afNSP

### DIFF
--- a/claudedocs/handoff-tmux-webapp.md
+++ b/claudedocs/handoff-tmux-webapp.md
@@ -719,6 +719,45 @@ directions. Read both halves — either one alone is wrong.**
   unchanged: the pin (needs sudo, one host) or clearing the record on `192.168.50.1` (fixes every
   machine on the LAN, and is still untouched).
 
+### 🔴 `test_subsystem_store_api.py` HAS RECURRED — the "Fixed by devrc#996" claim above is now FALSE
+🔴 **This corrects the Gotchas bullet that ends "Fixed by devrc#996 (`1b1f71ad`…)" and the
+open-investigation heading that says "nothing is fixing it".** #996 did not close it.
+
+- **Symptom + exact repro:** `devrc#1162` — a **one-markdown-file** PR — was blocked by
+  `tekton/devrc-pytests` on
+  `TestTheActorComesFromTheTOKEN::test_a_FORGED_actor_in_the_body_is_DISCARDED[record0-…]`.
+  That is the **same case name** this doc already records from `devrc#1056`, in the block above.
+- **Observed (with values):** the failure is attached to head `74e39bea`
+  (`FAILED: pytests — FAILING: TestTheActorComesFromTheTOKEN…`). The **immediately preceding**
+  head of the same branch, `e1d1318f`, failed a DIFFERENT test
+  (`test_no_unallowlisted_public_ip_literal_is_committed` — a real defect of mine, since fixed), so
+  the two reds are unrelated. After a rebase with no content change beyond that fix, head
+  `2fd84888` passed: `TOTAL collected=19942 passed=19939 skipped=3 failed=0`.
+- **Ruled out — the diff.** #1162 touches exactly ONE file, `claudedocs/handoff-tmux-webapp.md`.
+  A markdown file cannot reach a store-api test. This is the doc's own stated discriminator
+  ("the discriminator that settled it was a DOCS-ONLY PR failing"), reproduced.
+- **Control, and its LIMIT:** `scripts/tests/test_subsystem_store_api.py` on a clean
+  `origin/main` worktree ran **3/3 green — 641 passed each — at 301.6s / 293.8s / 296.8s.** The
+  tight spread rules out load inflation *in that run*. 🔴 **But it is a weaker control than it
+  looks:** it ran on the DEV HOST while the failure is in the nix **sandbox tier** under CI
+  concurrency, so it is a second sample of a DIFFERENT environment, not of the failing one. The
+  structural argument (one markdown file) is what actually discriminates here; the 3/3 only shows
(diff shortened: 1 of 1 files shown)


ZacxDev:
Different session here — I diagnosed this same flake in parallel today (claim `devrc-ci-flake-population`) and reached a **named mechanism plus a reproducer**, which I think supersedes two lines in this PR. Flagging before it merges rather than landing a competing doc.

Your attribution is right and your control's caveat is right. The part I can add:

## The mechanism, from the suite's own instrumentation

On the failing run (`devrc-ci-86zxj`, sha `5de43017`) **all 3 failures** printed:

```
MECHANISM = SERVER_BLOCKED_IN_FSYNC   (handler threads=1 [Thread-566 (process_request_thread)=SERVER_BLOCKED_IN_FSYNC], accept loop parked=True)
E   TimeoutError: timed out
    .../python3.12/socket.py:720: TimeoutError
```

`server.py:_replace_bytes` issues an `fsync` **before** the response is written; fsync blocks in uninterruptible sleep. So the client's 60s bound expires and the gate reports a code failure for an I/O stall. Note the check description names only ONE test — the run actually had `failed=3`.

## Why it is capacity, and specifically DISK

- `devrc-ci` is pinned to one node (`nodeSelector: kubernetes.io/hostname: talos-xr6-r7p`).
- **12 pipelineruns overlapped** that window (5 devrc + 4 gitops-validate + auditloop) — the same figure the file's 2026-08-29 comment cites.
- The gate workspace is `emptyDir medium=disk` and the nix caches are `local-path` PVCs, so every one of those runs contends on **one physical disk**.

## Two corrections I would offer to this PR

1. 🔴 **The seed/ordering hypothesis is refuted.** No reordering is required to produce this. Delaying a single `fsync` past `HANG_TIMEOUT` reproduces it exactly.
2. 🔴 **"Reproduce in the tier that fails" is no longer the cheapest next probe** — I reproduced it **on the dev host**, by inducing the mechanism rather than waiting for the environment:

```
control  (normal fsync):  8 passed in 4.63s, rc=0
repro    (one fsync stalled 65s > HANG_TIMEOUT=60):  1 failed, 7 passed, rc=1
```

and it failed on the **identical test with the identical parametrisation** as CI:

```
CI:    TestTheActorComesFromTheTOKEN::test_a_FORGED_actor_in_the_body_is_DISCARDED[record0-…-kelp-forest-zach]
repro: TestTheActorComesFromTheTOKEN::test_a_FORGED_actor_in_the_body_is_DISCARDED[record0-…-kelp-forest-zach]
```

via an `LD_PRELOAD` shim that delays exactly one `fsync` — no repo file modified, only the test process affected. I validated the shim delayed as intended before trusting its verdict.

## One trap worth recording

The obvious infra fix is wrong: the gate task has `computeResources: null`, so "give it CPU/memory requests" is the natural reading — but Kubernetes requests govern CPU and memory, **not disk I/O**, and this is an fsync stall. (Also, `null` is not a devrc oversight: **all 449** taskruns in that namespace declare no resources.) Raising `HANG_TIMEOUT` again is worse — the file itself computes ~320 hung-call sites × 60s ≈ 5.3h against a 45m task budget.

Happy to either fold this into your doc or land it separately as a correction — your call, it's your effort's doc. I'll hold off editing `handoff-tmux-webapp.md` either way.

PR #1181, opened 2026-08-31
fix(store-api flake): it is fsync CONTENTION, not seed/ordering — with a dev-host reproducer

ZacxDev:
The store-api "flake" has been re-diagnosed three times because nobody could make it fail on demand. This lands the mechanism and a ~70 s reproducer that runs **on the dev host**, so the next recurrence is a measurement rather than a hypothesis.

Relates to #1180 (same flake, concurrent effort) — evidence also posted there rather than editing that effort's doc.

## Mechanism
`server.py:_replace_bytes` fsyncs **before** the response is written, and fsync blocks in uninterruptible sleep. When one fsync exceeds `HANG_TIMEOUT` (60.0), the client raises `TimeoutError` at `socket.py:720` and the gate reports a **code failure for an I/O stall**.

The suite already classifies this correctly and says so unprompted — all 3 failures on `devrc-ci-86zxj` (sha `5de43017`) printed:
```
MECHANISM = SERVER_BLOCKED_IN_FSYNC   (handler threads=1 [… =SERVER_BLOCKED_IN_FSYNC], accept loop parked=True)
```
⚠ That run had `failed=3`; the check description names only one, which is how this keeps getting read as a single mystery test.

## Why CI and not here
`devrc-ci` is pinned to one node (`talos-xr6-r7p`); the gate workspace is `emptyDir medium=disk` and the nix caches are `local-path` PVCs — so every concurrent pipelinerun contends on **one physical disk**. **12 pipelineruns overlapped** the failing window, matching the figure this file's own 2026-08-29 note cites.

## The reproducer
`scripts/ci-repro/slowfsync.c` — an `LD_PRELOAD` shim delaying exactly one fsync past the bound. Modifies no repo file; affects only the process it is preloaded into. Instrument validated before its verdict was read.

| run | result |
|---|---|
| control (normal fsync) | `8 passed in 4.63s`, rc 0 |
| reproduction (one fsync stalled 65 s) | `1 failed, 7 passed`, rc 1 |

Failing on the **identical test and parametrisation** as CI: `TestTheActorComesFromTheTOKEN::test_a_FORGED_actor_in_the_body_is_DISCARDED[record0-…-kelp-forest-zach]`.

**This refutes the seed/ordering hypothesis** — no reordering is required.

## Two fixes that look right and are not
- 🔴 **CPU/memory requests cannot fix it.** The gate task has `computeResources: null`, so "declare requests" is the natural reading — but k8s requests govern CPU and memory, **not disk I/O**. `null` is also not a devrc oversight: **all 449** taskruns in that namespace declare none.
- 🔴 **Raising `HANG_TIMEOUT` again is worse than nothing.** 60 is already the symptom fix (from 15) and it did not hold; ~320 hung-call sites × 60 s ≈ 5.3 h against a 45 m budget — the documented state where nothing posts and required checks stay `pending` forever.

The real levers are infra and are **not** applied here: unpin the node or spread disk-heavy pipelines, cap concurrent runs per node (distinct from `tekton-supersede`, which only collapses redundant runs of the *same* PR), or isolate the workspace storage.

## Scope
The change to `test_subsystem_store_api.py` is **comment-only** — verified, no non-comment line in the diff. `scripts/ci-repro/` is deliberately outside `scripts/tests/` so it cannot perturb `run-tests.sh`'s two-way-pinned target/floor table (confirmed: `test_run_tests_targets.py` 31 passed after adding it).

## Gate
230 passed across `test_no_captured_text`, `test_no_captured_markup`, `test_run_tests_targets`, `test_ci_claim_matches_reality`; plus the store-api class itself (8 passed).

https://claude.ai/code/session_01Sm5tDmE4y8PrpEPdJGaVCC

### DIFF
--- a/scripts/ci-repro/README.md
+++ b/scripts/ci-repro/README.md
@@ -37,8 +37,11 @@ node. Measured 2026-08-31: gitops-validate is pinned to a **different** node
 (42/42 pods on `talos-uvh-gtj`) and the one auditloop run was on `talos-deu-s2q`.
 Neither mounts anything devrc-ci mounts. Sizing a concurrency cap against 12 would
 size it against runs that were never there.
-⚠ `claude/skills/tekton/SKILL.md:48` says "every run lands on `talos-xr6-r7p`" — true
-of *devrc-ci* runs, misleading as written, and the likely source of the 12.
+⚠ `claude/skills/tekton/SKILL.md` says "every run lands on `talos-xr6-r7p`" in its
+node-pinning paragraph — true of *devrc-ci* runs, misleading as written, and the likely
+source of the 12. **That same file already resolves it** further down, recording the
+`devrc-ci` / `gitops-validate` node split from homelab-infra #396 — so read both before
+editing either. (`grep -n 'talos-uvh-gtj' claude/skills/tekton/SKILL.md`.)
 
 🔴 **AND THE FAILING WRITE IS ON NEITHER NAMED VOLUME.** The gate pod mounts
 `nix-store-cache` at `/nix` and the per-run `source` PVC (a `volumeClaimTemplate`,
@@ -107,10 +110,17 @@ mechanism testable on demand rather than to eliminate rivals.
 ### Why LD_PRELOAD and not the narrower tool already in the repo
 
 `test_subsystem_store_api.py` monkeypatches `api._fsync_dir` inside
-`TestAHungRoundTripSAYSWhichSideBlocked` — `grep -n '_fsync_dir", _stall' ` finds it —
-and carries a 🔴 comment noting that patching the stdlib's `os.fsync` is process-global
-and would stall any other thread. (Cited by name, not by line: that file's comment
-block shifts on every edit, and a line citation into it shipped wrong twice.) This shim is **strictly wider** than what that comment warns
+`TestAHungRoundTripSAYSWhichSideBlocked`, and carries a 🔴 comment noting that patching
+the stdlib's `os.fsync` is process-global and would stall any other thread. Find it:
+
+```bash
+grep -n '_fsync_dir", _stall' scripts/tests/test_subsystem_store_api.py
+```
+
+(By name, never by line. A line citation into that file shipped wrong **three times** in
+this PR alone — twice from counting against the pre-PR tree, and once because the very
+commit that "fixed" it inserted lines above its own citation.) This shim is **strictly
+wider** than what that comment warns
 against. It is used anyway because it needs no repo edit and therefore tests the
 shipped code path exactly as CI runs it — but the narrower monkeypatch is the right
 tool if you only need `_fsync_dir`, and the warning above applies to this shim too.
@@ -118,10 +128,18 @@ tool if you only need `_fsync_dir`, and the warning above applies to this shim t
 ## Two fixes that look right and are not
 
 🔴 **Raising `HANG_TIMEOUT` again is worse than doing nothing.** 60.0 is already the
-symptom fix (raised from 15 on 2026-08-29) and it did not hold. The test file's own
-arithmetic (`:155-158`) is ~320 hung-call sites × 60 s ≈ 5.3 h; re-counted live in
-this PR's tree it is 211 `fetch(` + 113 `await_audit(` = **324**, so ≈ 5.4 h. Either
-way it dwarfs the budget.
+symptom fix (raised from 15 on 2026-08-29) and it did not hold. The test file carries
+this arithmetic itself, next to the constant — search for `per-hung-call` in
+`test_subsystem_store_api.py`. Re-derive the multiplier rather than trusting either
+copy:
+
+```bash
+F=scripts/tests/test_subsystem_store_api.py
+echo $(( $(grep -c 'fetch(' $F) + $(grep -c 'await_audit(' $F) ))   # hung-call sites
+```
+
+At the time of writing that is ~324 (× 60 s ≈ 5.4 h); the file's own text says ~320
+(≈ 5.3 h). Either way it dwarfs the budget.
 ⚠ **That budget is NOT 45 m** — the test file says 45 m and the live values are the
 gate task's `timeout: 1h0m0s` inside pipeline `timeouts: {tasks: 1h10m0s, pipeline:
 1h25m0s}`. The conclusion survives (5.4 h ≫ 1 h, and even 15 s × 324 ≈ 81 m exceeds
@@ -133,17 +151,31 @@ CPU and memory, **not** disk IOPS, so requests would not make fsync faster. But 
 devrc-ci run is `nodeSelector`-pinned to one node, so non-zero requests are exactly the
 standard mechanism for making excess runs **Pending instead of co-scheduled** — i.e.
 they are one way to implement the concurrency cap listed below. (`computeResources:
-null` is not a devrc oversight: **479/479** taskruns in that namespace declare none, so
+null` is not a devrc oversight: **every** taskrun in that namespace declares none — the
+ratio is what matters and has held at every reading; the absolute count drifts — so
 this is a platform-wide default, and changing devrc alone would not stop under-declared
 neighbours oversubscribing the node.)
 
 The real levers are infra and are **not** this repo's to apply, ranked by whether they
 can move the write that actually stalls:
 
-1. **Cap concurrent devrc-ci runs on the pinned node, or unpin it.** Sizes against
-   **7**, not 12. Distinct from `tekton-supersede`, which only collapses redundant
-   runs of the *same* PR. Non-zero `computeResources` is one way to implement this —
-   it does not touch IOPS, but it makes excess runs Pending instead of co-scheduled.
+1. **Bound the disk-heavy work on that NODE, or unpin devrc-ci.** 🔴 **Capping
+   devrc-ci alone does not bound the node** — the mechanism is node-local *device*
+   contention, so any pod there competes regardless of which volumes it names.
+   `talos-xr6-r7p` is not devrc-ci's: measured 2026-08-31 it also hosted `auditloop`
+   38, `clawgate` 13, `vetr` 7, `naida` 6, `remix` 2 pods, and `auditloop-ci` is **not**
+   hostname-pinned (its template only excludes one other node), so it lands there
+   routinely. A cap set at "7 devrc-ci runs" bounds one pipeline out of five that
+   schedule there. Re-derive the co-residents before sizing anything:
+
+   ```bash
+   kubectl -n tekton-ci get pods --field-selector spec.nodeName=talos-xr6-r7p \
+     -o json | jq -r '.items[].metadata.name | split("-")[0]' | sort | uniq -c | sort -rn
+   ```
+
+   Distinct from `tekton-supersede`, which only collapses redundant runs of the *same*
+   PR. Non-zero `computeResources` is one way to implement a cap — it does not touch
+   IOPS, but it makes excess runs Pending instead of co-scheduled.
 2. **Give the gate's ephemeral layer (`/tmp`) faster or isolated storage** — that is
    where the stalling fsync lands. Requires first measuring whether it shares a device
    with the local-path PVs; see the ⚠ above.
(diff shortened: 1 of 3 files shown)


ZacxDev:
## 🔴 Correcting this PR's own description — an adversarial audit found the diagnosis wrong in one place

Posting rather than silently editing the body, because a reviewer may already have read the wrong version.

**The body says "the gate workspace is `emptyDir medium=disk`". That is FALSE.** Each run's `source` workspace is a `volumeClaimTemplate` — a per-run `local-path` PVC holding a ~13 MB clone. I had read the POD's `tekton-internal-workspace` emptyDir, which is Tekton's own plumbing, not the pipeline's workspace. It also contradicted a correct statement already in this repo (`claude/skills/tekton/SKILL.md:41`).

It matters beyond accuracy: it **misdirected the fix**. The volume every concurrent run actually *shares* is the static `nix-store-cache` PVC (30 Gi, `local-path`, same pinned node), so "give the workspace isolated storage" pointed at the wrong volume.

Five more corrections to body claims, all now fixed in `843ca63b`:

| body claim | corrected |
|---|---|
| "affects only the process it is preloaded into" | FALSE — `LD_PRELOAD` survives `exec()`; parent **and** child each stall 65 s (measured) |
| "This refutes the seed/ordering hypothesis" | sufficiency sold as necessity — the reproducer shows fsync latency *suffices*; the CI classifier is what refutes ordering |
| "CPU/memory requests cannot fix it" | too strong, and it contradicted this PR's own recommended lever — requests don't govern IOPS, but node-pinned runs + requests = Pending, i.e. a concurrency cap |
| "all 3 failures … printed" (implying one flake) | the third failed against `CLIENT_BOUND = 0.25`, **not** `HANG_TIMEOUT` — the do-not-raise-the-constant advice does not address it |
| "(5 devrc + 4 gitops-validate + auditloop)" | sums to 10, not 12; actual is **7** devrc-ci + **4** gitops-validate + **1** auditloop |

Counts also drifted already: 449 taskruns → **479**; ~320 call sites → **~324**. Nothing asserts on them, and the README now says so.

## Reproducer hardening (also from the audit)
`sleep()`'s return was discarded, so a signal could shorten the stall while printing an identical line — an under-delivered stall would produce a **passing** run reading as "not reproducible". Now `nanosleep`-on-`EINTR`, reporting **measured** elapsed. Latch is atomic (was a C11 race against a multi-threaded server); `dlsym` checked; scratch paths made unique; `-p no:randomly` dropped (a no-op — `pytest-randomly` isn't in `gatePyEnv`, so it implied an ordering control that doesn't exist).

Re-verified after the fixes: clean under `-Wall -Wextra`; instrument reports 65.0 s; control `8 passed`/rc 0; reproduction `1 failed`/rc 1 with `TimeoutError` + `MECHANISM = SERVER_BLOCKED_IN_FSYNC`. The `.py` change is still comment-only (verified). Gate: 248 passed.

## One thing the audit surfaced that this PR does NOT fix
`CLIENT_BOUND = 0.25` / `SERVER_STALL = 1.2` (`test_subsystem_store_api.py:15319`) is the same "timing bound tighter than the thing it discriminates" shape as commit `f4a3d69b`. It is **unfixed and unflagged elsewhere** — recorded in the file's comment so it is open rather than absent, not silently absorbed here.

```audit-claims round=1 audited=f874ff2e..843ca63b
1. The `emptyDir medium=disk` workspace claim is replaced with the measured one: per-run `source` volumeClaimTemplate vs the shared static `nix-store-cache` PVC, and the storage lever now names `nix-store-cache`.
2. The "affects only the process it is preloaded into" claim is replaced with the measured LD_PRELOAD-survives-exec behaviour, plus an explicit instruction to preload for one test selection only.
3. The seed/ordering claim is restated as sufficiency-not-necessity, attributing the refutation to the CI classifier rather than to the reproducer.
4. The "CPU/memory requests cannot fix it" claim is narrowed to "requests do not speed up fsync but do cap co-scheduling on a pinned node".
5. The "3 failures" claim is split: the third failed against CLIENT_BOUND=0.25, not HANG_TIMEOUT, and that 0.25/1.2 pair is recorded as unfixed.
6. The overlap breakdown is corrected to 7 devrc-ci + 4 gitops-validate + 1 auditloop, and the drifting counts (479 taskruns, ~324 sites) are marked unpinned.
7. slowfsync.c: sleep-return discarded -> nanosleep-on-EINTR reporting measured elapsed; non-atomic latch -> __atomic_test_and_set; unchecked dlsym -> explicit failure; fixed /tmp paths -> unique.
8. The README now states why LD_PRELOAD was chosen over the narrower _fsync_dir monkeypatch at :15342, and that nothing builds or lints the .c.
```

ZacxDev:
## Round 2 (delta) — NOT clean; my round-1 fix introduced a new false claim

Full disposition of round 1's eight items: **six actually fixed**, one partially (the storage claim — corrected, then wrong in a *new* way), one fixed with a new defect. Round 2's own findings, all now fixed in `e7f655f9`:

- 🟡 **The contention set is 7, not 12.** Round 1's "the volume every concurrent run shares is `nix-store-cache`" is false for 5 of the 12 runs it cites as evidence: gitops-validate is pinned to a **different** node (42/42 on `talos-uvh-gtj`), the auditloop run was on `talos-deu-s2q`. Only the **7 devrc-ci** runs were ever on the pinned node. I re-verified both against the live cluster. Sizing a cap against 12 sizes it against runs that were never there. Also: tekton-ci PVs spread over four nodes (100/52/51/43), so "all `local-path`" ≠ "all one disk".
- 🟡 **The lever was still wrong.** "Give `nix-store-cache` isolated storage" cannot move the failing write — the CI traceback puts the stalling fsync in the step container's **ephemeral layer under `/tmp`**, and the gate sets no `TMPDIR` and mounts nothing at `/tmp` (verified on the pod spec). Levers re-ranked by whether they can move that write; the "shared device" premise is now marked **inferred, not measured**.
- 🟡 **Every in-file line citation I added was wrong — twice.** Computed against the pre-PR file, and my *recount* went stale before it was committed because the comment block itself shifts them. Recounting was not the fix: all in-file citations are now **by name** and cannot drift.
- 🟢 `dlerror()` clears on read, so `dlerror() ? dlerror() : …` printed `(null)` and was formally UB — the "fail loudly" path discarded the reason it exists to print.
- 🟢 The README attributed "~324 / 5.4 h" to the test file, which says ~320 / 5.3 h — two documents in one PR disagreeing where they had agreed.
- 🟢 "a 45 m task budget" is not live: gate task `timeout: 1h0m0s`, pipeline `timeouts {tasks: 1h10m0s, pipeline: 1h25m0s}`. Conclusion survives; the number doesn't.

**Ledger:** round 2 · payload lines changed THIS round: **+77/−36** (since round 1: +242/−102) · all three files are this PR's payload, so the attribution gate does not fire.

Re-verified after the fixes: clean under `-Wall -Wextra`; instrument reports 65.0 s measured; control `8 passed`/rc 0; reproduction `1 failed`/rc 1 with `TimeoutError` + `MECHANISM = SERVER_BLOCKED_IN_FSYNC`; `.py` still comment-only across the full PR range; gate 248 passed.

```audit-claims round=2 audited=843ca63b..e7f655f9
1. The "every concurrent run shares nix-store-cache" claim is replaced with the measured contention set of 7 devrc-ci runs on the pinned node, and the 12-run figure is explicitly marked as NOT the contention set.
2. The per-node PV spread (100/52/51/43 across four nodes) replaces the "all local-path so all one disk" inference.
3. The storage lever is replaced: the stalling fsync is located in the step container's ephemeral /tmp layer, neither nix-store-cache nor the source PVC, and the levers are re-ranked by whether they can move that write.
4. The "shared device" premise between the ephemeral layer and the local-path PVs is labelled inferred-not-measured.
5. All in-file line citations are replaced with by-name references (CLIENT_BOUND, the assertion message, `_fsync_dir", _stall`); cross-file server.py line citations are retained deliberately.
6. slowfsync.c: the double dlerror() call is replaced with a single stashed read.
7. The ~324/5.4h figure is attributed to a live re-count rather than to the test file, and the file's own ~320/5.3h is quoted alongside it.
8. The 45m budget figure is corrected to the live gate task timeout (1h) and pipeline timeouts (tasks 1h10m, pipeline 1h25m), with the test file's stale copy left untouched and called out.
```

ZacxDev:
## Round 3 (delta) — NOT clean; the citation fix shipped a THIRD instance of its own bug

Round 2's six items: **five actually fixed**, one partially — and the fix for "every line citation was wrong" introduced another wrong line citation, invalidated by the very commit that added it. Third instance of one class.

Round 3's findings, all fixed in `8a93f2dc`:

- 🟡 **`:155-158` was wrong at head** — the same commit inserted 5 lines above its own citation. Recounting has now failed twice, so this round stopped recounting: **every in-file line citation is gone**, replaced by names or printed `grep`s. Cross-file `server.py` citations are kept (that file is untouched here) and were re-verified against the CI traceback.
- 🟡 **"the commands are in this PR's comments" was FALSE** — 0 `kubectl` in any comment, and it had *replaced* a working pointer. That sentence was the doc's entire drift mitigation. The commands are now **in the file**.
- 🟡 **The sizing rule was wrong**, in the direction round 2's own re-framing should have caught. Round 2 established the mechanism as node-local **device** contention — under which any pod on that node competes regardless of which volumes it names. `talos-xr6-r7p` is not devrc-ci's: measured today it also hosts `auditloop` **38**, `clawgate` **13**, `vetr` **7**, `naida` **6**, `remix` **2**, and `auditloop-ci` is **not** hostname-pinned. A cap at "7 devrc-ci runs" bounds one pipeline out of five that schedule there.
- 🟢 `479/479` in the body vs `499` in the ledger — inside one commit. Now expressed as a **ratio** ("every taskrun declares none"), which has never moved, while the count has read 449/479/499/468.
- 🟢 The `grep` locator didn't run as printed (no path → read stdin). Fixed and executed.
- 🟢 The `SKILL.md` characterisation was incomplete — that same file already records the `devrc-ci`/`gitops-validate` split from homelab-infra #396, 28 lines below the sentence I quoted. Both are now pointed at.

**The through-line:** this round applied the repo's own lesson — *ship the derivation command, not the restated number*. Every count in the file is now either a runnable command or an explicitly-unasserted record.

**Verified:** every command added this round runs as printed — grep locator → exactly **1** hit; call-site sum → **324**; co-resident query → **8** rows; `SKILL.md` locator → **2** hits. No in-file line citation remains. `slowfsync.c` untouched this round. Still comment-only in the `.py`. Gate: **256 passed**.

**Ledger:** round 3 · payload lines changed THIS round: **+81/−20** (since round 1: **+323/−122**) · both files are payload, so the attribution gate does not fire.

I considered stopping here on the prose-payload criterion — no 🔴, and the blast radius is "the document contains a false sentence". **Not invoking it**: round 3's fix was structural (numbers → commands), which attacks the generator, so I cannot claim the rounds won't stop on their own. Round 4 tests that.

```audit-claims round=3 audited=e7f655f9..8a93f2dc
1. All in-file line citations are removed; targets are located by name or by a printed grep command, and the cross-file server.py citations are retained deliberately with that scope stated.
2. The call-site multiplier is replaced by a runnable two-grep sum rather than a literal figure.
3. The "commands are in this PR's comments" pointer is replaced with the actual kubectl/jq commands inline, plus a warning that pruned pipelineruns make old windows non-re-derivable.
4. The 479/479 taskrun figure is replaced by the ratio claim (every taskrun declares none), removing the body-vs-ledger contradiction.
5. The concurrency-cap lever is rewritten: capping devrc-ci alone does not bound the node, the co-resident pipelines are named, and the query to re-derive them is printed.
6. The SKILL.md reference now points at both the misleading sentence and the later paragraph in that same file which already resolves it.
```

ZacxDev:
## Round 4 (delta) — verdict "safe to merge"; one seam + three drifts, all fixed. **Ending the ladder here, with the reason stated.**

Round 3's six items: **five actually fixed, one fixed-in-substance with new inaccuracies.** The structural fix held — round 4 executed every command the README prints and all but one ran verbatim and returned what the prose claims.

Fixed in `2af7941b`:

- 🟡 **The doc contradicted itself 130 lines apart.** The 🔴 heading still said "the contention set is 7, NOT 12" and excluded the other five runs because "neither mounts anything devrc-ci mounts" — reasoning round 2's own re-framing repudiates, since node-local *device* contention doesn't care which volumes a pod names. A reader hits the heading first, sizes a cap at 7, stops. Now says 12 is wrong in one direction and 7 in the other.
- 🟢 **The co-resident command omitted `--kubeconfig`** — alone among the file's `kubectl` commands. There's no default `KUBECONFIG` here by design: bare `kubectl` errors loudly, but with another cluster exported the `--field-selector` matches nothing and returns a **silent zero** reading as "devrc-ci is alone on the node". Swept the class, not the site: every `kubectl` now passes it.
- 🟢 **`keep: 100`/daily → live `keep: 20`, hourly** (changed 2026-08-28; two stale comments in homelab-talos still say 100). Now derived by a printed command — and the correction cuts *against* this doc: the window is 5× less recoverable than the superseded figure implied.
- 🟢 **"one pipeline out of five" → eight are resident** (devrc-ci 73, auditloop-ci 38, vetr-infra-guards 7, clawgate-ci 7, naida-ux-audit 6, clawgate-e2e 5, remix-ux-audit 2, clawgate-ux-audit 1), and the query now groups by `tekton.dev/pipeline` instead of name prefix, which split one pipeline across rows.

Also recorded: `auditloop-ci` and `remix-ux-audit` lost their hostname pins on 2026-08-31 (homelab-infra #597), so their pod counts are **pre-un-pin fossils** that *understate* future co-residency; and **`naida-ux-audit` is still pinned to this node (6/6)** — a concrete lever this section's own logic calls for, which the doc had not named.

**Ledger:** round 4 · payload **+36/−16** (rounds 1–4: **+359/−138**) · one file, payload, zero scaffolding — the attribution gate does not fire.

---

## 🔴 Why I am stopping the ladder here rather than running round 5

Stating this explicitly, because a report that *ended* on this criterion must not be indistinguishable from one that converged.

**The prose-payload stop criterion applies**, and its three conditions are met: no 🔴 in any round; the blast radius is "the document contains a false sentence" (nothing in `scripts/ci-repro/` is executed by CI, collected by any runner, or referenced by any floor table or check derivation); and each recurring **shape** was swept at every site rather than at the reported one — all in-file line citations removed in round 3, all `--kubeconfig` omissions in round 4, all load-bearing counts replaced by runnable derivations.

**Why the rounds would not stop on their own:** this document's subject is *live infrastructure that changes faster than the audit cycle*. Round 4's findings were largely drift introduced by cluster changes made hours earlier — `auditloop-ci` was un-pinned the same day, retention changed three days prior — not by my edits. The taskrun count alone read 449 → 479 → 499 → 468 → 482 across this one session. A prose document describing that surface will yield a real finding on every future round indefinitely, which is precisely the non-terminating case this criterion exists for.

**What I am NOT fixing, so it is open rather than absent:**
1. The two stale `keep: 100` comments in `homelab-talos` — different repo, out of this PR's scope.
2. Whether the ephemeral `/tmp` layer and the local-path PVs share one physical device — labelled **inferred, not measured**; needs node-level access, not `kubectl`.
3. `naida-ux-audit`'s live pin on the contended node — named as a lever, not acted on; infra changes are not this repo's to apply.
4. `grep -c 'fetch('` counts *lines*, not call sites, and matches `prefetch(` — the figure is written "~324" with a re-derivation command beside it.

If a future round wants to reopen this, the honest trigger is a **🔴 or a changed conclusion**, not another drifted count.

```audit-claims round=4 audited=8a93f2dc..2af7941b
1. The 🔴 contention-set heading is rescoped so it no longer asserts 7 as the node's contention set and no longer uses the volume-based exclusion the lever section repudiates; it now points at that section.
2. Every kubectl command in the file passes --kubeconfig, with KC defined before first use, and the silent-zero failure mode is stated.
3. The prune retention figure is replaced with the live keep:20 hourly value and a printed derivation command, and the non-re-derivability warning is strengthened rather than softened.
4. The co-resident count is corrected to 8 distinct pipelines, the query groups by tekton.dev/pipeline label instead of name prefix, and the 2026-08-31 un-pinning of auditloop-ci and remix-ux-audit is recorded as making those counts understate rather than overstate.
5. naida-ux-audit's surviving hostname pin is named as an unactioned lever.
```

PR #1185, opened 2026-08-31
fix(audit-dispatch): a CACHED nix build prints nothing at rc 0 — port back the nix log recovery, with guards

ZacxDev:
Rank 4 of the audit-pr-ladder handoff — the `nix log` recovery block #1117 deferred to "its own PR and its own round".

## First: is it still needed?

The handoff item allowed closing this as *"a named reader records that the fallback is not coming back"*, so I checked before porting. **It is needed.** Measured on one derivation, three runs of the *same* command:

| run | rc | lines | `RESULT:` |
|---|---|---|---|
| uncached, `-L` | 0 | 5 | 1 |
| **cached, `-L`** | **0** | **0** | **0** |
| `nix log <drv>` | 0 | 2 | 1 |

`#1117` carried forward `--no-link -L`, which fixes the **succeeding** build. It does nothing for the **already-built** one: nix prints nothing and exits **0**, while the brief tells the auditor to "read each runner's own `RESULT:` line". The silence is indistinguishable from a pass. Sibling sessions and CI build the same tree constantly here, so the cached case is the common one.

## The four acceptance criteria (measured by #1133's round 3, recorded in the handoff)

1. **No stderr parenthetical.** The old block claimed `nix path-info`'s stderr was discarded by the `>` two lines down. That `>` belongs to a different command and redirects stdout; the flake error lands on the terminal. It also contradicted the same block's correct "`-L` writes to stderr" rule 28 lines above.
2. **`mktemp`, not a fixed `/tmp/tier.log`.** Audit agents run concurrently here by design; a shared path means each truncates the other's file and greps the other's tier — with every guard still passing.
3. **A structural guard.** `_assert_the_cached_build_fallback_carries_its_guards` checks each guard by name. A substring assert on the *command* stayed green for the entire life of a missing `-L`; this is the same argument its sibling `_assert_every_nix_build_prints_its_log` makes.
4. **The hazard count is not asserted in the prose.** The old block claimed "THREE DIFFERENT FALSE GREENS" while its bullets delivered two, and `test -s` speaks to emptiness only — it claims nothing about truncation.

## Design notes

Guards are **named constants**, so a mutant can remove exactly one and the harness can say which. New mutant **V62** removes the empty-`$DRV` guard, whose absence is an *affirmative* false green rather than silence: `nix log ""` resolves as `.` and prints the cwd flake's **default package** log, which may read `RESULT: PASS (exit=0)`.

🔴 **The block is emitted only where check lines were actually fenced.** Not cosmetic — with no check names it is advice about a command the brief never gave, and its own text contains `#checks.`, which made a presence check circular. Measured: it stole a kill from **V41**, whose mutant empties the check-name list.

## Verification

- **Watched RED at `7de5b0bd`** by grafting the test onto that tree (1 failed). Registered as `RED_AT_BASE_R17` with fix-matrix row `r17/N1`.
- **Mutant harness: 138 rows, all as expected**; V62 killed by *exactly* its named test.
- **Dev-host tier:** `GATE: RESULT=PASS exit=0` (pytest + node).
- **Sandbox tier** — the one the merge gates on — built one at a time: `devrc-pytests RESULT: PASS (exit=0)` (1468 lines), `devrc-nodetests RESULT: PASS (exit=0)` (8833 lines), 0 timeouts. Base `7de5b0bd`.

Three defects of mine were caught by this repo's own ledgers rather than by me, and are worth noting for anyone reviewing: I named a mutant `V62` that did not exist (the fix-matrix ledger refused it); my guard was over-broad and stole kills from C3 and V41; and my precondition was circular. All fixed above.

### DIFF
--- a/scripts/audit-dispatch.py
+++ b/scripts/audit-dispatch.py
@@ -2834,6 +2834,16 @@ CHECKS_DISCRIMINATOR = (
 )
 
 
+# 🔴 The cached-build fallback's guard lines, each a SINGLE named constant so a
+# mutation can remove exactly one and the harness can name which. Each closes a
+# DIFFERENT measured false green; the number of them is deliberately not
+# asserted in the emitted prose (`test -s` speaks to emptiness only and claims
+# nothing about truncation).
+NIX_LOG_TMPFILE = 'LOG=$(mktemp -t audit-tier-XXXXXXXX.log)   # per-agent: a FIXED path is truncated by every sibling agent'
+NIX_LOG_DRV_GUARD = '[ -n "$DRV" ] || { echo "NO DERIVATION — wrong attr, or this repo has no such tier"; exit 1; }'
+NIX_LOG_RUN_GUARD = 'nix log "$DRV" > "$LOG" 2>/dev/null || { echo "NO LOG — never built HERE (nix log is per-machine)"; exit 1; }'
+NIX_LOG_EMPTY_GUARD = '[ -s "$LOG" ] || { echo "EMPTY LOG — cannot vouch"; exit 1; }'
+
 def _toolchain_checks_lines(tc):
     """The sandbox tier — only when the flake really declares one."""
     if not tc.flake:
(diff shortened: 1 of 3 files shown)


ZacxDev:
Round 1 (blind, full) returned **4 🟡 + 2 🟢**. Every finding reproduced independently before fixing; fixes are in `2eaa3c62`.

```audit-claims round=1 audited=90202ce5..2eaa3c62
1. `_assert_the_cached_build_fallback_carries_its_guards` no longer tests substrings against the whole brief: it slices the ```bash fence and asserts each guard as a whole line, in order, each terminating in `exit 1; }`, with the log path required to come from a command substitution.
2. The three mutants measured surviving at 90202ce5 — dropping the $DRV guard's `; exit 1`, a fixed log path with the word mktemp left in a comment, and demoting the guards out of the fence into prose — were re-run against 2eaa3c62 and all three now fail, each naming test_the_cached_build_fallback_is_emitted_with_its_guards as the killer, with an unmutated positive control at 127 passed.
3. The remediation note no longer offers "replace the exits with `return`"; it says to wrap the block in a function and call it, and states that at top level `return` is not a stop, measured in interactive zsh, interactive bash and `bash <script>`.
4. The `if tc.check_names else []` conditional was removed as unreachable-false (an earlier `if not tc.check_names: return` fires ~80 lines above), and the comment claiming it "keeps the guard isolating" was replaced with what actually does — the test's own precondition.
5. The emitted block's last command is now `grep -n "RESULT:"` rather than `grep -c`, so a healthy tier exits 0 and a log with no verdict line exits 1; previously healthy exited 1 and a timing-out tier exited 0.
6. The docstring no longer claims "dropping any ONE of them restores it": the run guard buys a better diagnosis, not an independent false-green closure, because `test -s` still catches the empty file.
7. `nix path-info` stderr is captured to `$ERR` and echoed by the $DRV guard instead of being discarded, so a wrong attribute and a `flake.nix` syntax error no longer share one diagnosis.
8. The first draft of fix 1 matched `^grep -c` against the last LINE rather than the last COMMAND and failed a block that was already correct; it now splits on `;`.
```

**Gates after the fixes:** mutant harness `✅ 138 row(s), all as expected` with V62 killed by exactly its named test; sandbox tier (the one the merge gates on) built one at a time — `devrc-pytests RESULT: PASS (exit=0)`, `devrc-nodetests RESULT: PASS (exit=0)`, 0 timeouts.

⚠ The auditor did **not** run the sandbox tier and said so; the numbers above are mine, at base `7de5b0bd`.

ZacxDev:
## Ladder stopped after round 1 — operator decision, on cost

Round 1 returned findings, so this is **not** the findings-keyed stop and **not** a clean round. Recording it explicitly so nobody reads the merge as convergence.

🔴 **`2eaa3c62` — the fix round — is UNAUDITED.** This session's rate makes that a real risk, not a formality: a fix has introduced a new defect in five of six rounds across the two PRs, twice by committing the exact error the same commit was fixing elsewhere. The first draft of this PR's own F1 fix did it again (matched the last LINE where only the last COMMAND sets `$?`).

**What is verified, and how:**
- All six findings independently reproduced before fixing — not accepted from the auditor.
- The three mutants that walked the old spelling pin now die, each naming `test_the_cached_build_fallback_is_emitted_with_its_guards` as killer, with an unmutated positive control at 127 passed.
- The emitted block walked **as a shell** on every path: bogus attr → `NO DERIVATION` + nix's real error (rc 1); built-elsewhere drv → `NO LOG` (rc 1); empty log → `EMPTY LOG` (rc 1); healthy → rc 0; log with no verdict line → rc 1.
- Mutant harness `✅ 138 rows, all as expected`, V62 killed by exactly its named test.
- Sandbox tier — the one the merge gates on — built one at a time: both `RESULT: PASS (exit=0)`, 0 timeouts.

**What is NOT verified, stated plainly:** the fix round itself has had no adversarial read. Round 2's range, for anyone who picks it up, is `90202ce5..2eaa3c62`.

⚠ `strict: false` on this branch, so the required checks speak for the PR branch, not the tree the merge creates.

PR #1186, opened 2026-08-31
test(index-append): guard DISAGREEMENT, not agreement — a reworded second protocol left all 61 tests green (cg#473)

ZacxDev:
Closes clawgate card **cg#473**. Test-only; **no skill document changed**.

## The problem, re-measured on this base

`scripts/tests/test_index_append_protocol.py` exists to satisfy Cairn phase 0's criterion —
*"one write path … with a test that fails if a second protocol reappears."* It did not.

Re-measured myself on a fresh `cp -a` copy of `50bfd91f` (`.git` removed first,
`PYTHONDONTWRITEBYTECODE=1`, one copy per mutant), against the module **as it is on `main`**:

| mutant, appended | base module |
|---|---|
| a complete, materially conflicting second protocol in `write-back.md` | **61 passed** |
| a contradicting per-caller carve-out in `subsystem-index/SKILL.md` | **61 passed** |
| a new third door under `claude/skills/**` | **61 passed** |

All three are **isolated** — verified mechanically against the base module's own literals:
none reproduces `MANDATE` (`anchored on \`## Nuance / work-history\``) and none contains any of
the 10 `BANNED_IMPERATIVES`. The isolation check runs against the literals parsed out of the
base module source, not out of my head.

**Root cause** (as the card states, and confirmed): layers 1-4 all key on what an *agreeing*
copy would contain. A fork is by definition a document that says something **different**, so the
module detected AGREEMENT and was blind to DISAGREEMENT.

## Design chosen: key on the MECHANISM a fork cannot avoid naming

The one thing a second append protocol cannot do is describe a write mechanism without naming a
write mechanism. So layer 5 detects the **co-occurrence, within one paragraph**, of

* a **MECHANISM** token — `` `Write` ``/`` `Edit` ``, `Write tool`, `use|then|via|plain <tool>`,
  `(y/N)`, `y/N`, `yes/no`; and
* a **STORE CONTEXT** token — `analyze-service-index`, `## Nuance / work-history`,
  `nuance heading`, `subsystem-index/SKILL.md`, `index entry`, `entry file`, `index store`.

🔴 It makes **no attempt to tell an instruction from a description**, and no docstring claims it
does. Both doors quote the retired prompt verbatim while recording its retirement, and a guard
that banned the spelling would make the resolution record unwritable — which is how the next
session re-derives the fork from scratch. Instead the predicate feeds three *structural*
questions:

1. **`TestTheStoreWriteCarrierLedgerIsEXACT`** — the exhaustive set of `claude/skills/**/*.md`
   that state a mechanism for this store, five entries each with its reason. Fails when the set
   GROWS *or* SHRINKS. Unlike the existing `MANDATE` ledger it does **not** require a new door to
   agree with the owner, which is the whole point. `prune-index/SKILL.md` is in it *deliberately*
   and with its argument written down: a prune is a DELETION, it keeps its y/N and its `Write` on
   blast radius, and it says so in its own text. (On `main` its absence from any ledger was luck,
   not design — a slightly different wording there would have false-positived.)
2. **`TestThePointerStatesNoMechanismOUTSIDEThePinnedRegion`** — positional, so no wording walks
   it: every mechanism token in `write-back.md` must sit inside the already-hashed
   `one-append-protocol` region. This is the existing
   `test_the_pointers_only_occurrence_is_INSIDE_the_pinned_region` argument, widened from one
   agreeing string to the whole token class.
3. **`TestTheOwnersMechanismProseIsPinnedWHOLE`** — the owner's three mechanism paragraphs,
   whitespace-normalised, under one hash. **CONTENT-selected, not position-delimited**: a hash
   over the write half alone would miss a carve-out inserted into the read half, and the read
   half is where the historical record lives — precisely the text a re-opener would edit.

### Rejected alternative

The card's other suggestion: extend the existing region hash over **all of `write-back.md` below
step 4**, plus a marker-delimited hash on the owner's write half. Worse on three counts:

* it **cannot satisfy criterion 3 at all** — a NEW file under `claude/skills/**` lies inside no
  hashed region, so mutant (c) stays green;
* it freezes `## Notable`, `## Auto-discovered pointers` and `## Bloat discipline`, which are
  genuinely `/analyze-service`-specific and churn independently of the protocol. A high-churn
  hash trains people to paste the new sha without reading — which the region's own failure
  message says is the one way to make it worthless;
* a *positional* hash on the owner's write half misses a carve-out inserted above the separator.

## Result: each mutant dies from EXACTLY ONE test, a different one each

| mutant | base | with this PR | killing test |
|---|---|---|---|
| **(a)** reworded second protocol in the POINTER | 61 passed | **1 failed**, 70 passed | `TestThePointerStatesNoMechanismOUTSIDEThePinnedRegion::test_every_mechanism_token_is_inside_the_region` |
| **(b)** contradicting carve-out in the OWNER | 61 passed | **1 failed**, 70 passed | `TestTheOwnersMechanismProseIsPinnedWHOLE::test_the_owners_mechanism_paragraphs_are_UNCHANGED` |
| **(c)** new third door under `claude/skills/**` | 61 passed | **1 failed**, 70 passed | `TestTheStoreWriteCarrierLedgerIsEXACT::test_the_carrier_set_is_EXACTLY_the_ledger` |
| **harness positive control** — a literal `BANNED_IMPERATIVE` | 1 failed | 2 failed | fires on BOTH module versions, proving the harness read the mutated tree in both configurations |
| unmutated baseline | 61 passed | **71 passed** | — |

Assertion messages, verbatim (excerpted):

* (a) `claude/skills/analyze-service/reference/write-back.md states a write mechanism OUTSIDE its pinned \`one-append-protocol\` region:` → `line 94: '(y/N)'`, `line 95: 'use the \`Write'`
* (b) `The MECHANISM PROSE of claude/skills/subsystem-index/SKILL.md CHANGED. expected sha256 2ed3ce83… actual sha256 6419e7fe… 4 paragraph(s) currently selected:` then lists them, the fourth being the planted carve-out
* (c) `the set of skill documents stating a WRITE MECHANISM for the analyze-service index store has changed. NEW carriers (a second protocol is growing): ['claude/skills/handoff/reference/index-append.md']`

## False-positive control

The risk here is the opposite of the usual one. Both doors legitimately quote the retired
`(y/N)` while *describing* the retirement, and a guard that makes the resolution record
unwritable is how the fork comes back.

* **On the real tree: 71 passed** — no new guard fires on either door's historical quotations.
* **On a plausible re-wording of `write-back.md`'s pinned historical block** (still quoting
  `(y/N)`, `Write`, `Edit` and the anchor heading verbatim): the ONLY failure is the
  pre-existing `test_region_hash` — **none of the three new guards fires**. With `EXPECTED_SHA`
  repinned per that test's own documented procedure: **71 passed**.

## Every control was watched to fail

Seven controls ship alongside the three guards. They are **not** regression coverage for the
three mutants — they are instrument controls, green before and after by construction — so each
was mutation-tested against the guard it controls. 8 guard mutants, **0 survivors**:

| guard mutant | controls killed |
|---|---|
| `MECHANISM` matches nothing | 7 tests, incl. `test_the_predicate_can_OBSERVE_a_new_carrier`, `test_the_selection_is_NOT_EMPTY` |
| `STORE_CONTEXT` matches nothing | 6 |
| predicate uses `or` instead of `and` | `test_the_predicate_needs_BOTH_halves` |
| front matter not stripped | `test_front_matter_is_NOT_a_mechanism_statement` |
| `_normalise` is identity | `test_a_REWRAP_does_not_move_the_hash` (+ the pre-existing rewrap/region tests) |
| confinement filter reports nothing | `test_the_confinement_check_can_SEE_an_outside_occurrence` |
| owner digest is a constant | `test_the_hash_MOVES_for_an_appended_carve_out` |
| `MECHANISM` narrowed to `(y/N)` only | 7 |

🔴 **The sweep found two real defects in my own first draft** — both the exact shapes
`claude/RULES.md` warns about, and neither visible from a green suite:

* `test_front_matter_is_NOT_a_mechanism_statement` was **vacuous**: its fixture spelled
  `allowed-tools: …, Write, Edit`, which `MECHANISM` does not match anyway, so it passed with the
  strip deleted. The fixture now spells both halves inside the front matter, and the test
  asserts its own premise.
* `test_the_confinement_check_can_SEE_an_outside_occurrence` carried **its own copy** of the
  filter, so replacing the *guard's* copy with `if False` left all 71 tests green — a control
  proving a duplicate reachable while the real guard was dead. Both now call
  `_mechanism_outside_region`. One rule, one place.

## Out of scope, stated so it is not read as covered

`scripts/subsystem-store-api/server.py` carries a **THIRD** append protocol (flock +
temp-file-rename, `PUT` behind `If-Match`). It writes the pod's `/data`, not
`~/.claude/analyze-service-index/`, so "one protocol" holds for the local store today — but this
ledger scans `claude/skills/**/*.md` only and would never see it, and **phase 1 of the Cairn
plan makes the pod canonical**. Recorded in the module docstring's non-coverage list; **not**
touched here.

## What this still cannot see, plainly

* A fork that describes a whole-file retype **without naming a tool** ("re-type the entry
  yourself") and without a confirm-gate spelling. The `MECHANISM` half is a token list; it is the
  narrowest thing a real fork cannot avoid, not a proof of completeness.
* A fork stated in a document **outside `claude/skills/**`** — the scan root is unchanged.
* It reads documents. It cannot make an agent follow them. (Already stated in the module
  docstring; unchanged.)

## Gate

Base sha: `50bfd91f` (`origin/main` at branch point); head `4f8a3427`. Both tiers run on the final committed tree, verdicts verbatim:

**Dev-host tier** — `nix develop <worktree> --command bash scripts/gate.sh --tier both`, on the
committed tree:

```
  TOTAL collected=20070  passed=20067  skipped=3  failed=0  (floor: 18383 = sum of 29 per-target floors)
RESULT: PASS (exit=0)          <- pytest
RESULT: PASS (exit=0)          <- node
GATE: RESULT=PASS exit=0
```

**Merge-gating sandbox tier** — `nix build ...#checks.x86_64-linux.pytests` and `...nodetests`,
built **one at a time** (a combined invocation produces false failures from store contention):

```
devrc-pytests>   TOTAL collected=20070  passed=20067  skipped=3  failed=0  (floor: 18383 = sum of 29 per-target floors)

```

See the comment thread for `tekton/devrc-pytests` going red on `TestAHungRoundTripSAYSWhichSideBlocked.
test_a_stall_in_the_FSYNC_region_is_NAMED` — a real-socket timing test in a file this diff cannot
reach, green in the identical local `nix build` of the same derivation on the same head.

WARNING: `strict: false` on this repo's branch protection, so a green check is a claim about this
branch, not about the tree the merge creates. Not gated on the merged tree here.

https://claude.ai/code/session_016dXi6YHVk9m75kDQAaJviv

### DIFF
--- a/scripts/tests/test_index_append_protocol.py
+++ b/scripts/tests/test_index_append_protocol.py
@@ -41,6 +41,21 @@ WHAT THIS MODULE DOES **NOT** COVER, stated so nobody reads coverage into it
   its wording. It cannot see a hostile rewrite that keeps the substrings — which
   is what layer 2 (the whole-region hash) is for, and why the region is pinned
   whole rather than by keyword.
+* 🔴 Layers 1-4 all key on what an AGREEING copy contains — the exact `MANDATE`
+  fragment, or ten exact phrasings of the retired text. A fork is by definition a
+  document that says something DIFFERENT, so those layers detect AGREEMENT and
+  are blind to DISAGREEMENT. That gap was MEASURED (clawgate cg#473): a complete,
+  materially conflicting second protocol appended to `write-back.md` left the
+  module at 61 passed. Layer 5 is the disagreement half, and it is keyed on the
+  MECHANISM TOKENS a second protocol cannot avoid naming rather than on any
+  particular wording.
+* `scripts/subsystem-store-api/server.py` carries a THIRD append protocol
+  (flock + temp-file-rename, `PUT` behind `If-Match`). It writes the POD's
+  `/data`, not `~/.claude/analyze-service-index/`, so "one protocol" holds for
+  the local store today — but this module scans `claude/skills/**/*.md` only and
+  would never see it. Phase 1 of `claudedocs/plan-cairn-integration.md` makes the
+  pod canonical; that is when this has to be reconciled. Recorded so nobody reads
+  coverage of it into a green run here.
 
 🔴 NO TEST HERE READS THE REAL STORE. Nothing in this file opens
 `~/.claude/analyze-service-index/`; it reads tracked documents in this repo only.
@@ -102,6 +117,23 @@ BANNED_IMPERATIVES: tuple[str, ...] = (
 # --- the region extractor ------------------------------------------------------
 
 
+def _region_span(doc: str, name: str) -> tuple[int, int]:
+    """Half-open `[start, end)` CHARACTER offsets of the region's body.
+
+    Split out from `_region()` so the confinement guard in layer 5 can ask
+    "is this occurrence inside the region?" against the same extraction the hash
+    uses. One rule, one place: two extractors would drift, and the whole point of
+    this module is that a duplicated predicate disagrees with itself eventually.
+    """
+    begin, end = f"<!-- {name}:begin", f"<!-- {name}:end"
+    i = doc.find(begin)
+    assert i != -1, f"marker {begin!r} is missing from {POINTER.name}"
+    i = doc.index("-->", i) + len("-->")
+    j = doc.find(end, i)
+    assert j != -1, f"marker {end!r} is missing from {POINTER.name}"
+    return i, j
+
+
 def _region(doc: str, name: str) -> str:
     """Bytes between `<!-- <name>:begin -->` and `<!-- <name>:end -->`.
 
@@ -111,12 +143,7 @@ def _region(doc: str, name: str) -> str:
     inside each marker stays editable without changing which bytes are hashed.
     Same contract as `test_subsystem_resolver.py`'s extractor, deliberately.
     """
-    begin, end = f"<!-- {name}:begin", f"<!-- {name}:end"
-    i = doc.find(begin)
-    assert i != -1, f"marker {begin!r} is missing from {POINTER.name}"
-    i = doc.index("-->", i) + len("-->")
-    j = doc.find(end, i)
-    assert j != -1, f"marker {end!r} is missing from {POINTER.name}"
+    i, j = _region_span(doc, name)
     body = doc[i:j].strip()
     assert body, f"region {name!r} is EMPTY — the hash would guard nothing"
     return body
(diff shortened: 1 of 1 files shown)


ZacxDev:
## ⚠ `tekton/devrc-pytests` is RED, and I do not believe it is this diff — evidence, not a story

The check failed on `TestAHungRoundTripSAYSWhichSideBlocked.test_a_stall_in_the_FSYNC_region_is_NAMED`, in `scripts/tests/test_subsystem_store_api.py`. Flagging it rather than waving it through, with the controls I actually ran:

1. **Same tier, same tree, green here.** `nix build .#checks.x86_64-linux.pytests` — the *identical* derivation Tekton builds, from the same source — passed on this exact head `4f8a3427`: `TOTAL collected=20070 passed=20067 skipped=3 failed=0` / `RESULT: PASS (exit=0)`. Also passed on the previous head `4a186526`. `nodetests` green in both tiers.
2. **The diff cannot reach it.** One file changed, `scripts/tests/test_index_append_protocol.py`. Nothing in the repo imports it (`find scripts -name '*.py' | xargs grep -l test_index_append_protocol` → empty), and `test_subsystem_store_api.py` shares no fixture with it.
3. **A sibling in that same file failed the same way on someone else's PR.** #1175 went red on `TestTheActorComesFromTheTOKEN.test_a_FORGED_actor_in_the_body_is_DISCARDED` — and that test's CI `TimeoutError` is, by its own docstring, the reason `TestAHungRoundTripSAYSWhichSideBlocked` was written. These are real-socket, thread-parking timing tests, and #1181 was running concurrently.

**What I could NOT verify:** I have no Tekton step log (`target_url` is null on the status), so I could not compare wall times to separate load from a genuine assertion failure the way `claude/RULES.md` prescribes. So this is "three controls point away from the diff", not "proved a flake". **A re-run is the remaining check, and it is not mine to force** — a required check can only be re-satisfied by a fresh push, and I am not pushing a no-op commit to launder a red gate.

Not merging.

ZacxDev:
## Rebased onto `329f0d40` — gates the merged tree, and re-triggers the red check

`main` moved (#1184) while this was in flight. Rebased; the only file main gained is a handoff doc, so no overlap with this diff — but a clean merge is not a clean merge, so the module was re-run on the merged tree: **71 passed**.

### The red `tekton/devrc-pytests` — attribution, honestly

It failed on `TestAHungRoundTripSAYSWhichSideBlocked.test_a_stall_in_the_FSYNC_region_is_NAMED` in `test_subsystem_store_api.py`. **Controls run, all pointing away from this diff:**

- this PR touches **one** file, `test_index_append_protocol.py`, and nothing imports it
- the failing test passes on `main` (`329f0d40`) **and** on this head, same environment, in isolation
- the whole `TestAHungRoundTripSAYSWhichSideBlocked` class: **5/5 pass locally**, ~5s each, no variance
- the identical `nix build` sandbox derivation is green on this head

🔴 **But I am not calling it a flake.** `target_url` is null on that status, so there is no step log and I cannot tell WHICH assertion failed — and this class has two very different ones (`assert stalled.is_set()` vs the mechanism assertions), which fail for different reasons. An absence cannot distinguish two mechanisms. Passing 5/5 in isolation is also exactly what cg#439 did, and that was real.

So: strong attribution away from this diff, **no diagnosis of the CI failure itself**. The class's own docstring says it is an instrument for a flake that has never been made to reproduce. If it goes red again on the same test after this push, that is new information and worth its own card.

### Independent re-verification of the headline claim

I re-ran mutant (a) myself rather than accepting the report. Isolation checked **mechanically** against literals parsed from the module source — `reproduces MANDATE: False`, `banned imperatives: none` — and it dies with:

```
write-back.md states a write mechanism OUTSIDE its pinned `one-append-protocol` region:
    line 94: '(y/N)'
    line 95: 'use the `Write'
```

That is phase 0's criterion — *a test that fails if a second protocol reappears* — actually met, on a fork that shares no word with the old one.

PR #1191, opened 2026-09-01
docs(handoff): rank 4 closed — all four ranked items done, and rc17's second recurrence DID reach the binary

ZacxDev:
Closes the last gap from the audit-pr-ladder thread: **rank 4 was merged and shipped while the ranked list still read OPEN.**

## Why this PR exists at all

`#1185` merged as `76d20386` and shipped to both hosts — but the handoff still listed rank 4 as workable. A merged PR is invisible to `/resume`; **the ranked list is what it reads**, so the next session would have claimed rank 4 and re-done finished work. That is the exact duplicate-work hazard rank 1's own update fixed earlier in the same session, walked into at the last step.

All four ranked items are now marked DONE, with a note at the top of the list that nothing there is workable.

## Findings added

- 🔴 **rc 17 recurred within hours, on both hosts — and the second instance had real blast radius.** The first touched only `.bats`/`_test.go` files (binary unaffected); the second touched `cmd/clawgatectl/client.go`, so both hosts were building `clawgatectl` *without* the `#468` deeplink fix `0.8.20` carries. Fixed: both now at `eed7db5a` running `0.8.20`, `drift-check` rc 0. **"rc 17 fired" and "the binary is wrong" are independent claims — read the diff every time.**
- 🔴 **`pull.rebase = true` makes `git pull --ff-only` print a failure it did not suffer.** Measured: `error: cannot pull with rebase: You have unstaged changes` — and HEAD moved anyway. `git reflog` is the arbiter (`merge origin/trunk: Fast-forward`). The other session's tracked modifications were byte-identical before and after.
- 🔴 **The `| tail` trap, reproduced in that same command.** `… | tail -3; echo "PULL_RC=$?"` printed `PULL_RC=0` — `tail`'s status, which happened to agree with a success it could not observe. Documented in `RULES.md` *and* in this very doc, and still hit by habit.
- ⚠ **Same version, same subtree commit, different store hash — not a fault.** The workbench holds an untracked file inside the built source dir, and the build reads the tree, not the commit.

Docs only — no code, no test surface.

### DIFF
--- a/claudedocs/handoff-audit-pr-ladder.md
+++ b/claudedocs/handoff-audit-pr-ladder.md
@@ -144,40 +144,30 @@ findings-keyed stop rule does not terminate in the guard-hardening regime.
 🔴 **Numbering is deliberately STABLE.** The rank is half a `claim-work` slug's identity, so
 renumbering silently re-points every live claim. Closed items are retained as DONE markers.
 
+**ALL FOUR RANKED ITEMS ARE CLOSED.** Nothing here is workable; a `/resume` picking from this
+list should read the Gotchas and stop, not claim a rank.
+
 1. **DONE (2026-08-31) — ship the laptop.** Both hosts converged; `drift-check.sh` rc 0,
    `PARITY-RC=0` on both. Do not re-claim.
    forcing: none
 2. **DONE (found already closed 2026-08-31) — the three leftover worktrees.** ⚠ The successor
    condition is real and unowned: **137 registered worktrees**, most on merged branches. Not
-   filed — no named owner and no checkable closing condition, so a ticket would read as covered
-   while nothing could close it.
+   filed — no named owner and no checkable closing condition.
    forcing: none
-3. **DONE (2026-08-31) — `#1133`'s never-run round 3.** Run blind over `ea36a489..9ff5c9a9`. It
-   returned 4 🟡 + 1 🟢. Exactly one was live in `main` — the escape hatch's summary obligation,
-   deleted by a reword — fixed and pinned in **`#1157`**, which then ran its own three-round
-   ladder and was **stopped on the stated criterion, not on a clean round** (rationale recorded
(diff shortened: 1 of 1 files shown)


ZacxDev:
🔴 **The `tekton/devrc-pytests` red on `969b9fd5` was environmental, not this PR.** Diagnosed before re-triggering; new head `157922e8` also records the finding in the doc.

**This PR changes ONE file: `claudedocs/handoff-audit-pr-ladder.md`.** It cannot reach store-api code.

**What failed:** `TestAHungRoundTripSAYSWhichSideBlocked.test_a_stall_in_the_FSYNC_region_is_NAMED`, `scripts/tests/test_subsystem_store_api.py:15766`.

🔴 **The assertion is self-diagnosing and says what to conclude:**

> `AssertionError: the server never reached the stall site, so the hang under test was NOT the one this test set up — the report below would be about some other mechanism`

The test detected that **its own setup had not taken effect** — a scheduling failure under load, not a defect in anything.

**Controls, run before concluding:**
- 3/3 pass locally on this tree, ~5s each, against a **530.30s** `scripts/tests` target in CI.
- Positive control: `--collect-only` confirms the `-k` filter selected the CI-failing test (1 hit), so the local green is not a filter quietly deselecting it.
- `tekton/devrc-nodetests` passed in the same run.

**Second occurrence this session, on a different test.** `#1178` died the same way on `TestTheBackstopNeverSendsASecondResponse` (464s target, also a docs-only diff, also 3/3 locally). Two docs-only PRs, two different tests, one file.

Worth contrasting the two messages: this one hands over the diagnosis; `#1178`'s said the **opposite** of its own values (`assert 0 == 1` with `raw == b''`, under the text "a SECOND complete response followed the 200"). A self-diagnosing assertion is worth writing.

**Still deliberately NOT filed** as a work item: the Tekton-capacity condition has no closing condition anyone can check, and three fixes are already rejected on measurements. Now recorded in the handoff as routine rather than occasional.

PR #1207, opened 2026-09-01
docs(handoff): rank 5 worked example — evict terminal content from the skill-chain doc (6.5x -> 4.7x target)

ZacxDev:
The bloat auditor merged in #1064 immediately indicted the doc this arc had been growing: **80,397 B, third-worst of 94 docs, 33.5% evictable**. This is the `EVICT_HISTORY` half of rank 5, done by hand on one doc so the contract can be designed from evidence instead of proposed from theory.

**What was evicted:** the 8 resolved investigation blocks, each replaced by a one-line **tombstone** naming the verdict and where the durable lesson now lives; plus the 8 completed ranked items, collapsed to stubs. Before deleting anything I verified every block lesson has a home in the subsystem store (`devrc/tests.md`, `devrc/skills.md`) or in `State now` carried-forward headline.

**Result: 80,397 -> 57,356 B (-23,041 B, 28.7%), 6.5x -> 4.7x target.**

### Three contract requirements, all discovered by doing it

1. **A tombstone, not a deletion.** Append-only exists so a reader can see a prior reading was *corrected*. The capacity block proves the case: it is terminal **and partly refuted**, so its line stays and points at the correction that overturned it.
2. **The rank NUMBER must survive.** It is half a `claim-work` slug identity — deleting it re-points every live claim. The auditor already charges 200 B per rank for exactly this reason.
3. **Any eviction tool must assert structure BEFORE writing.** My first attempt reported *"59.7% saved"* and had in fact **destroyed three whole sections**. A regex split on `###` runs to the next `###`, which **spans later `##` headings**, so the last block in a section swallows the rest of the file. The impressive number *was* the damage. This version asserts the H2 list is unchanged and the ranks are still 1..11, and refuses to write otherwise.

### Not claimed

This does **not** fix the doc. It is still **over** the 40,960 B hard cap at 57,356 B — exactly as the auditor predicted (`net 25,306 -> 55,091 B, OVER HARD CAP`). Getting under needs the 9 superseded-but-unmarked investigation blocks and the 19,343 B `Gotchas` section, both judgment calls rather than mechanical evictions. Deliberately left for a decision.

### DIFF
--- a/claudedocs/handoff-skill-chain-usage-audit.md
+++ b/claudedocs/handoff-skill-chain-usage-audit.md
@@ -45,7 +45,8 @@ survives adversarial re-derivation, and fix whatever it exposes.
 - ✅ **MERGED — rank 6: devrc#1108 as `57b010fb`** (squash, 2026-08-31T05:15:15Z; **verified by
   CONTENT on `origin/main`, never by ancestry** — ancestry is false forever after a squash).
   9 commits: the correction, six audit rounds, two gate re-triggers. **Every round found something
-  real and FIVE of six found the PREVIOUS round's correction wrong** — see the ladder block below;
+  real and FIVE of six found the PREVIOUS round's correction wrong** — detail in store `devrc/skills.md`
+  (the ladder block itself is now a tombstone);
   that ratio is the durable finding, not the PR.
 - ✅ **CLOSED — rank 3**, committed on #1055: the drift question does not discriminate (95.5% of
   abandoned docs declare open work vs a **90.3% control**). Real finding: **13 / 26** abandoned and
@@ -95,7 +96,7 @@ survives adversarial re-derivation, and fix whatever it exposes.
 EVICTED 2026-09-01 (terminal). Verdict: **8 cleanly-ended · 4 interrupted-at-end · 2 context-exhausted · 2 never-started** of 16 — carried in `## State now`'s headline, which survives every update.
 
 ### ✅ CLOSED — the topic-drift sessions, and why the question as posed does not discriminate
-EVICTED 2026-09-01 (terminal). Verdict: the question does NOT discriminate — **95.5%** of abandoned docs declare open work vs a **90.3%** control. Instrument `claudedocs/skill-chain-drift-audit.py`; see rank 3.
+EVICTED 2026-09-01 — **terminal EXCEPT its open probe, which is now rank 12** (it was wrongly labelled wholly terminal; caught by audit). Verdict: the question does NOT discriminate — **95.5%** of abandoned docs declare open work vs a **90.3%** control. Instrument `claudedocs/skill-chain-drift-audit.py`; see rank 3.
 
 ### 🔴 WALK-BACK — "queue depth is what moved" was UNDER-CONTROLLED, and the store already knew
 **Corrected 2026-08-31, by reading the subsystem store instead of re-deriving.** This doc briefly
@@ -133,10 +134,14 @@ does not name which of ~28 targets failed), which is why attempt 1's named test 
 `failed=0` are different shapes and must not be merged into one story.
 
 ### ✅ CLOSED — the audit ladder on #1108, and what SIX rounds on a 21-line docs PR actually taught
-EVICTED 2026-09-01 (terminal). Verdict: **five of six rounds found the PREVIOUS round's correction wrong**; an ATTRIBUTION GATE, not a clean round, is what ends such a ladder. Durable in store `devrc/tests.md` + `devrc/skills.md`.
+EVICTED 2026-09-01 (terminal). Verdict: **five of six rounds found the PREVIOUS round's correction wrong** — and 🔴 **the zero-payload ATTRIBUTION GATE CANNOT FIRE on a docs-only PR**: the payload IS the `.md`, so every fix round is 100% payload by construction and **stopping was a JUDGEMENT CALL, not a gate**. (A clean round still ends a ladder — `claude/RULES.md`. This tombstone previously asserted the inverse on both counts; corrected after audit.) Durable in store `devrc/skills.md` and `claude/skills/audit-pr/reference/round-ladder-evidence.md`.
 
 ### ✅ CLOSED — the red `devrc-ci` checks are a CAPACITY problem, and the repo had already diagnosed it
 EVICTED 2026-09-01 (terminal, and PARTLY REFUTED — that is why this line stays). Capacity is real and measured in `devrc/tests.md`, but it was **not the right class for #1055**: see the CORRECTION block below, which is KEPT.
+🔴 Also carried out of that block, because it is a RETRACTION and this pass's own contract says a
+correction must be tombstoned rather than deleted: **"An equal test COUNT is not an equal TREE."**
+The "same `collected` count, different verdict" control was RETRACTED before write-up — two count
+groups each held BOTH verdicts, and each run carried a distinct `refs/pull/N/merge` preview sha.
 
 ### Both PRs are red on tests neither one touches — environment-sensitive, not code
 - **Symptom + exact repro:** `gh pr checks 1055` → `tekton/devrc-pytests FAILURE`; same for #1064.
@@ -405,10 +410,12 @@ rather than removed and renumbered. New work is appended at the end.
 2. ✅ **DONE (2026-08-30)** — the handoff-write `Stop` hook. **MERGED as `ad891a5c`.**
    forcing: none
 
-3. ✅ **DONE (2026-08-30)** — the question as posed does NOT discriminate: 95.5% of abandoned
+3. ✅ **DONE (2026-08-30)** — the question as posed does NOT discriminate: 95.5% of abandoned docs
+   declare open work vs a 90.3% control. Its one open thread is now rank 12.
    forcing: none
 
-4. ✅ **DONE (2026-09-01) — NO deploy was needed and none was run.** The store copy at
+4. ✅ **DONE (2026-09-01) — NO deploy was needed and none was run.** The store copy was already
+   byte-identical to `origin/main`; traps are durable in store `devrc/skills.md`.
    forcing: none
 
 5. **Act on the handoff-doc bloat proposal** — `claudedocs/proposal-handoff-doc-bloat.md`.
@@ -420,10 +427,12 @@ rather than removed and renumbered. New work is appended at the end.
    2026-08-31: the clawgate skill ratchet DID catch a real unpaid 405 B addition — one case, on a
    SKILL body not a handoff doc, so it does not settle the replay.
    forcing: none
-6. ✅ **DONE (2026-08-31) — the item's own PREMISE was REFUTED by the re-measure it demanded, and
+6. ✅ **DONE (2026-08-31) — the item's own PREMISE was REFUTED by the re-measure it demanded.**
+   **MERGED as `57b010fb`.**
    forcing: none
 
 7. ✅ **DONE (2026-08-30)** — merged, shipped to both hosts, registered on both, and probed live
+   against the deployed copy.
    forcing: none
 
 8. 🔴 **Grade the hook against the number it was built to move — THE CLOSING CONDITION OF THIS
@@ -439,11 +448,23 @@ rather than removed and renumbered. New work is appended at the end.
    rank 4 was skipped partly to keep it that way. (b) *The grading population must exist across
    the whole pre-window*: the pre-period is the 2026-08-15 → 08-29 corpus, already measured and on
    disk, so this half is satisfied — say so rather than re-deriving it.
+   🔴 **RANK 1's CONTROLS AND BAND — restored here after audit, because rank 1's block is
+   evicted, its classifier was throwaway and is GONE, and this item requires re-deriving that
+   measurement. Grading `8/16` as a point estimate is a defect.** The probe positive control:
+   `handoff_doc.py` fired on **32 of 48** readable loser transcripts, so the 16 zeros are
+   absences, not a dead probe. The compaction-marker detector is real but **weak — 20 of 5,961**
+   corpus transcripts. The two context-exhausted peaks were **944,856 (0.94)** and **965,819
+   (0.97)** against 1M; every other session peaked 0.29–0.60. **Three sessions have an
+   unresolvable ceiling, so at most one D could flip: the floor is `cleanly-ended ≥ 7`, not
+   exactly 8.** Denominator caveat: 3 loser transcripts live on the laptop and are unreadable
+   from the workbench, and 29 sessions were dropped because their resumed doc would not resolve —
+   **treat the bucket SHARES as the finding, never the denominator.**
    🔴 **And name the confound before cutting:** the guard changes the behaviour it measures, so a
    session that writes a handoff BECAUSE it was blocked is a success, not a contaminated sample —
    count blocks (`~/.cache/claude-handoff-write/s/*/fires-*`) and report them beside the rate.
    forcing: none
 9. ✅ **DONE (2026-08-31), premise REFUTED twice; #1055 MERGED as `8c108d8b`.** The durable output
+   is the two-arm local discriminator (branch vs `main`), not the PR — see rank 11.
    forcing: none
 
 10. **Right-size the devrc-ci gate pod, or establish it cannot be.** 🔴 **Its premise is REFUTED:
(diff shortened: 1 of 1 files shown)


ZacxDev:
Adversarial audit run (blind, first round). **It found two 🔴 and five 🟡, and the PR body was wrong** — correcting publicly rather than silently editing it, since it may already have been read.

**The claim that failed:** *"Before deleting anything I verified every block lesson has a home."* That does not hold at **two of eight** blocks.

- **🔴-1** — the tombstone replacing the audit-ladder block asserted *"an ATTRIBUTION GATE, not a clean round, is what ends such a ladder."* That is the **inverse** of the block it replaced, of `devrc/skills.md`, of `audit-pr`s own `round-ladder-evidence.md`, and of `claude/RULES.md` ("A CLEAN round ENDS the ladder"). The real lesson: the zero-payload gate **cannot fire** on a docs-only PR, so stopping is a judgement call. A tombstone exists to stop you reading further, so a wrong one is worse than the deletion.
- **🔴-2** — the topic-drift block was labelled `(terminal)` but carried a **declared-open probe** (the unmeasured RENAME-vs-SCOPE-MOVE split that decides which drift remedy to build) plus its 5-pair read set, surviving nowhere. Re-opened as **rank 12** with a closing condition.
- **🟡-1** — rank 8, the arc closing condition, requires re-running rank 1s measurement, whose classifier is gone. Its controls were evicted with it: the 32/48 probe positive control, the 20/5,961 detector weakness, and the **`cleanly-ended >= 7` floor** — so the `8/16` rank 8 grades against is not a point estimate. Restored into rank 8.
- 🟡-2 two unclosed `**` spans · 🟡-3 six stubs ending mid-sentence · 🟡-4 a self-flagged **retraction** deleted against this PRs own requirement 1 · 🟡-5 a cross-reference pointing at a tombstone.

**Corrected numbers:** 80,397 → **60,964 B** (−19,433, **24.2%**), 6.5x → **5.0x**. Not the 28.7% / 4.7x in the description — restoring the over-evicted content cost 3,608 B, which is the honest price. And **6** items were collapsed, not 8 (ranks 1 and 2 were already single-line).

**A fourth contract requirement, from the audit:** `handoff-audit.py` judges a resolved investigation on its **H3 heading alone**, so it still reports "8 resolved investigations" after eviction — **the signal is not idempotent**. An automated `EVICT_HISTORY` driven off it would re-evict tombstoned blocks and delete the tombstones.

**Unresolved, flagged not swept:** one combined run of `test_handoff_audit.py` + `test_handoff_doc.py` showed 6 failures in `TestPushabilityCasesTheFetchVersionGotWRONG`. Not reproduced in 4 further runs, base clean 5/5, wall time normal so not load. Those are git-pushability tests and this diff is one markdown file — but one event is not enough to attribute either way.

ZacxDev:
```audit-claims round=1 audited=21ee51d7..1127e620
1. 🔴-1 the audit-ladder tombstone was rewritten to state that the zero-payload attribution gate CANNOT fire on a docs-only PR and that stopping is a judgement call, replacing the inverted claim.
2. 🔴-2 the topic-drift tombstone no longer claims the block was wholly terminal, and its declared-open RENAME-vs-SCOPE-MOVE probe plus its 5-pair read set were re-opened as ranked item 12 with a CLOSES WHEN.
3. 🟡-1 rank 1 controls were restored into rank 8: the 32/48 probe positive control, the 20/5,961 detector weakness, the two context peaks, the cleanly-ended >= 7 FLOOR, and the denominator caveat.
4. 🟡-2 the two unclosed ** spans in ranked items 6 and 11 were closed.
5. 🟡-3 the six stubs that ended mid-sentence were given completed second lines.
6. 🟡-4 the retracted "An equal test COUNT is not an equal TREE" control was tombstoned into the capacity tombstone instead of being deleted.
7. 🟡-5 the State now cross-reference that pointed at the ladder block now points at the subsystem store and says the block is a tombstone.
8. The PR body numbers were corrected publicly in a PR comment: 80,397 -> 60,964 B, -24.2%, 6.5x -> 5.0x, and 6 items collapsed rather than 8.
9. A fourth contract requirement was recorded: handoff-audit.py judges a resolved investigation on its H3 heading alone, so the resolved-investigation signal is not idempotent after eviction.
```

ZacxDev:
**Round 2 (blind delta).** It verified all 9 of round 1s claims — 8 actually fixed, 1 partial — and then found **one 🔴 that round 1s own correction created**. That is the five-of-six shape this arc keeps measuring.

- **🔴-A** — round 1 added "MERGED as `f71ff648`" to rank 11. True, but `## State now` still said #1064 was *"GREEN and awaiting a merge decision"* and *"merging it was NOT part of rank 11 and was not done"*, and rank 5 still called it *"GREEN and mergeable"*. A resuming session reads `State now` first by this docs own convention, so it would re-litigate a decision already made. **Self-contradiction is worse than uniform staleness** — nothing tells the reader which arm is current. All three propagated.
- **🟡-B** — rank 8s *"THIS IS NOW THE ONLY SUBSTANTIVE ITEM LEFT"* was falsified by round 1 adding rank 12, in the sentence that tells a resuming agent what to pick up.
- **🟡-C** — the fourth contract requirement existed **only in a PR comment**, and `/resume` does not read those. All four now live in rank 5.
- **🟡-D** — a tombstone claimed the rank-1 headline is *"carried in `State now`, which survives every update"*. A merge round-trip **refutes** it: `State now` buckets as REPLACE and the tool flags that very line among 9 durable lines a replace would drop. It survives only if each session carries it forward by hand.
- 🟢-E the topic-drift H3 read `✅ CLOSED` while its body said otherwise — and `handoff-audit.py` buckets on the H3 alone, so it counted as fully resolved · 🟢-F a weak pointer in rank 9.

**Corrections to the PR description above, which is now wrong in three ways** (posting rather than silently editing, since it may already have been read):
- it says **"Three contract requirements"** — there are now **four**, and the fourth is the important one: `handoff-audit.py` buckets a resolved investigation on its **H3 heading alone**, so it still reports "8 resolved investigations" over 8 one-line tombstones. **The signal is not idempotent** — an automated `EVICT_HISTORY` driven off it would re-evict tombstoned blocks and delete the tombstones.
- it says **"ranks are still 1..11"** — they are **1..12**; rank 12 is the re-opened drift probe.
- its headline **"6.5x -> 4.7x"** and **28.7%** are superseded: now **62,298 B, 5.1x**, still over the hard cap.

**The flake is characterised, still not attributed.** 13 clean runs against 1 failure — and the 6 failures are the **entire** `TestPushabilityCasesTheFetchVersionGotWRONG` class (it contains exactly 6 tests), not 6 scattered ones. That is the signature of a **shared precondition** in a fixture that inherits `os.environ`, not per-test nondeterminism. This diff touches one markdown file and no fixture. Unexplained is not benign; leaving it stated.

```audit-claims round=2 audited=1127e620..ef49ac82
1. 🔴-A the #1064-merged fact was propagated into State now and rank 5; the contradictory "awaiting a merge decision" and "was not done" sentences were removed.
2. 🟡-B rank 8 no longer claims to be the only substantive item left, and names ranks 5 and 12 as also open while keeping its CLOSING CONDITION status.
3. 🟡-C all four contract requirements were moved into rank 5, including the non-idempotent-measurement one that had lived only in a PR comment.
4. 🟡-D the rank-1 headline tombstone now states that State now buckets as REPLACE and that the headline survives only by being carried forward by hand.
5. 🟢-E the topic-drift H3 was retitled from "✅ CLOSED" to "⚠ MOSTLY CLOSED (open probe → rank 12)" so the H3-only bucketing does not count it fully resolved.
6. 🟢-F rank 9 now names the block that describes the two-arm discriminator instead of pointing at rank 11.
```

ZacxDev:
**Round 3 (blind delta): NO 🔴.** It swept **~95 sites** — 17 PR/issue refs, 23 shas, 13 repo paths, ~14 in-doc cross-references, ~20 arithmetic assertions, 5 structural invariants — and returned 3 🟡 / 5 🟢. **Two of the three 🟡 were introduced by round 2s own fixes**, the same shape for the third round running, at reduced amplitude each time.

- **🟡-1 — I replaced an over-claim with a *checkable false enumeration*.** Round 2 removed rank 8s "THIS IS NOW THE ONLY SUBSTANTIVE ITEM LEFT" and wrote "(Ranks **5** and **12** are also open)". The open set is **{5, 8, 10, 12}** — the docs own auditor scores `8/12 ranked items done`. The omission hid **rank 10, which carries a live `claim-work` lock (`ci-speedup-7`, verified still held)**, so an agent reading the docs designated pickup point could open a second front on locked work. That is a real-world action, not a wrong read. Fixed with a **derivation command** instead of a third hand-maintained list — a list of open ranks goes stale by construction, which is exactly what happened twice.
- **🟡-2 — a precise pointer to the wrong block is worse than a vague one, because it gets followed.** Round 2 repointed rank 9 at *"The local suite s 76 failures…"*, a different control (ranks 2/7) that never mentions #1055. Repointed at the CORRECTION block, with the wrong target named so it is not re-chosen.
- **🟡-3 — a SECOND cross-reference-to-a-tombstone**, at a site round 1 never swept. Same class round 1 fixed at the `State now` site only; this is what the sweep-every-site mandate was for.
- **🟢-3 (pre-existing, real arithmetic bug):** *"11700m (82%) is the gate itself — 5 × 2250m"* contradicts its own ceiling line by 450m. **11,250m (79%)** is the five gate pods; **11,700m** is the whole `tekton-ci` namespace share. Corrected.
- **🟢-1:** the verification recipe hardcoded *"93 for #1064"* — a figure that has since read 93 / 124 / 130 / 138. Replaced with an instruction to re-derive.

Now **63,374 B, 5.2x**, still over the hard cap. Suite 371 passed. Net vs base is **−21.2%**, not the descriptions −28.7% — each audit round has correctly bought back bytes I should not have evicted.

```audit-claims round=3 audited=ef49ac82..59e85907
1. 🟡-1 rank 8s hand-written enumeration of the open set was replaced by a DERIVATION COMMAND, with the {5,8,10,12} reading dated and rank 10s live ci-speedup-7 lock called out.
2. 🟡-2 rank 9 now points at the CORRECTION block where the two-arm discriminator is derived, and explicitly names the "76 failures" block as NOT the right target.
3. 🟡-3 the second cross-reference-to-a-tombstone now says the target is a tombstone, states the mechanism inline, and points at store devrc/tests.md for the durable form.
4. 🟢-3 the 11700m/82% gate figure was corrected to 11,250m/79% for the five gate pods, with 11,700m identified as the tekton-ci namespace share.
5. 🟢-1 the How-to-verify recipe no longer hardcodes 93 for #1064 and instructs the reader to re-derive a figure that has read 93/124/130/138.
```

ZacxDev:
**Round 4: one 🔴 — and it is the worst defect of this ladder, because it was a CERTIFICATION rather than an omission.**

Round 3 rewrote the superseded-hypothesis line and, instead of merely repointing it, promoted an inherited framing into a checkable claim: *"the mechanism is stated inline here and the durable form is in store `devrc/tests.md`"*, where the mechanism given was *"a loopback socket starved by the scheduler … any test that makes a localhost call while the node is oversubscribed."*

**Verified with a positive control: `devrc/tests.md` contains ZERO loopback mentions** (6 other store files do, so the zero is a measurement, not a dead pattern). Its **newest** bullet, dated the same day, names a different mechanism entirely and says to read **`scripts/ci-repro/README.md`** — which is **in this PR own tree**, landed ~19 h before my commit:

> `server.py:_replace_bytes` fsyncs the file and then the parent dir **INSIDE the request, before the response is written**; `devrc-ci` is pinned to one node, so stacked runs contend **on one disk**.

So I named the wrong **resource** (CPU scheduler, not disk I/O), the wrong **population** ("any localhost call" vs the fsync-in-request path), and cited a store file that **contradicts** me. Blast radius beyond a false sentence: *"the node is oversubscribed"* routes a reader at capacity work — i.e. at **rank 10, which this same doc marks LOCKED** (`ci-speedup-7`, verified live). The socket timeout is the **surface**; the fsync is the **cause**. Corrected, with the README and the store bullet cited.

**A lead, not a conclusion:** that store bullet records the failure *"hits PRs whose diff cannot reach it, docs-only included"* — 4 reds across 3 different tests, each passing 3/3 locally. That is the signature of this PR own unattributed whole-class `TestPushabilityCasesTheFetchVersionGotWRONG` failure. Mine was a **local** run and theirs is devrc-ci, so it is not the same observation — but it is the first plausible mechanism anyone has offered for it, and it is written down now rather than lost.

- **🟡-1** — fixing `82%` at :297 **stranded `82%` at :307**, ten lines below, re-asserting the exact mis-attribution the edit above had just corrected. **79%** (11,250m of 14,220m) is the five gate pods; 11,700m is the namespace share.
- **🟡-2** — *"the 5-line tombstone below"* was 26 lines **above**; round 3 made it uniquely checkable while leaving the direction wrong.
- **🟢-1** — my derivation command yields the **count** while the sentence promised the **set**. It now says so and gives the membership method.

⚠ **Correction to the commit message of `34924645`:** it says `63,374 -> 64,690 B`. The measured size is **64,397 B** (5.2x, still over the hard cap). I misread the tool output when writing the message; correcting here rather than force-pushing over a pushed commit.

```audit-claims round=4 audited=59e85907..34924645
1. 🔴-1 the loopback/scheduler framing was retired; the line now names the fsync-in-request mechanism, cites scripts/ci-repro/README.md and the 2026-09-01 devrc/tests.md bullet, warns the failure hits PRs whose diff cannot reach it, and flags that the old text pointed at a store file containing no loopback bullet.
2. 🟡-1 the stranded "82% ... are the gate own pods" at :307 was corrected to 79% (11,250m of 14,220m) with 11,700m identified as the namespace share.
3. 🟡-2 the "5-line tombstone below" pointer was removed entirely by the 🔴-1 rewrite; no "tombstone below" phrasing remains.
4. 🟢-1 the derivation command sentence now states it yields the COUNT only, gives the membership method, and says to run it from the devrc repo root.
5. 🟢-2 the bare "93 commits behind" at :387 now carries a re-derive caveat naming the 124 reading.
```

ZacxDev:
**Round 5: claim 1 fully verified — the certification shape did NOT recur.** All seven factual assertions checked independently against the repo and store: `scripts/ci-repro/README.md` exists and describes the mechanism; `_replace_bytes` fsyncs the file at `server.py:2012` and the parent dir at `:2015`, both **before** the `:4012` respond; `devrc-ci` is pinned to `talos-xr6-r7p` in `devrc-ci-triggertemplate.yaml:244-245` (verified from the definition, not the doc); the store bullet is transcribed faithfully; `ci-speedup-7` is live. Size **64,397 B** confirmed three ways — so the PR comment was right and commit `34924645` message was the misread.

**But it found a 🔴 in the FIX.** Round 4 answer to a stale hand-written list was *"derive it, do not enumerate"* — and the derivation I shipped was broken in the one way a derivation must not be: **it matched itself.**

I wrote *"scan for items carrying no `✅ **DONE**` marker"*. Two defects:
- that literal appears **nowhere** in the ranked list — real items are spelled `✅ **DONE (2026-08-30)**`, date inside the bold — so a strict reader gets **1 done, 11 open** against the tool 8;
- the instruction **lives inside rank 8**, so a charitable unanchored search finds a done-marker within rank 8 and marks it done → open set `{5,10,12}`, and **rank 8 is the arc CLOSING CONDITION**. It vanishes three lines below the sentence calling it exactly that.

That is round 4 own harm shape reproduced by the fix for it — round 4 finding was a stated set hiding **rank 10** and its live lock; this hid **rank 8**. Fifth consecutive round in which the round under audit created the next finding.

**Fixed by anchoring on line start** (`an item is done iff its opening line matches ^<rank>. ✅`) and by naming the self-match trap in the text so the next editor does not reintroduce it.

**Verified, not asserted.** Applying the shipped method mechanically now yields `done=8, open={5,8,10,12}` — agreeing with the tool `8/12 ranked items done` **and** with the set the doc asserts — and the self-matching literal now occurs **0** times. That is the *"check the two agree"* control the doc itself mandates, actually executed.

```audit-claims round=5 audited=34924645..384e1c37
1. 🔴-1 the membership method was re-anchored to line start (an item is done iff its opening line matches ^<rank>. ✅), replacing the unanchored marker scan.
2. 🔴-1 the self-matching literal was eliminated: the string that previously matched only the instruction itself now occurs 0 times in the file.
3. 🔴-1 the text now names the self-match trap explicitly, stating that the instruction sits inside rank 8 and that an unanchored search would mark rank 8 done and drop the arc closing condition.
4. The method was executed mechanically against the document and produces done=8 open={5,8,10,12}, agreeing with both the tool N and the docs asserted set.
```

ZacxDev:
**Round 6: CLEAN. Zero findings. The ladder ends here — no round 7.**

The auditor executed the shipped method itself rather than trusting the claim, and all three sources agree:

| source | result |
|---|---|
| mechanical run of the rule as written | `done=8` → `{1,2,3,4,6,7,9,11}`; **`open={5,8,10,12}`** |
| `handoff-audit.py --sections 1` | **`8/12 ranked items done`** |
| the set the doc asserts | **`{5, 8, 10, 12}`** |

It validated the instrument before quoting it: on a mutated copy (rank 5 prefixed with a done marker) the tool moved to `9/12` and the parser to `done=9` — neither is wired to a constant. And **the real document is its own negative control**: rank 8 contains ✅ in its body and is still correctly reported open, which is exactly the defect round 5 found.

It also established the cross-check is **structural, not coincidental**: `handoff-audit.py:145-162` searches `DONE_MARK` over the item first line only, so the doc `^<rank>. ✅` is a strict **subset** of the tool predicate. Doc-done ⊆ tool-done always, so any divergence can only show up as membership-count **<** `N` — and "the method is wrong, not the tool" is the safe direction to send a reader.

Attempts to break the anchored rule all failed: no numbered line inside any of the 6 fence lines, no indented numbered items, `^1. ✅` cannot cross-match `11. ✅`, and no rank opens with anything before the ✅.

**Stop conditions: (a) no 🔴 — confirmed, zero findings of any severity · (b) blast radius — n/a, no false sentence remains · (c) recurring shape swept — confirmed, the membership method exists at exactly one site and repo-wide sweeps found no second instance.**

**Ladder summary: 6 rounds, and 5 of them found a defect created by the previous round fix** — the same ratio #1108 measured. Round 6 is the first where the fix did not create the next finding. Findings over the ladder: 3 🔴 (all in text I authored: an inverted tombstone, a certified-but-superseded mechanism, a self-matching derivation), 12 🟡, 10 🟢. `round 6 · payload lines changed THIS round: 8 · since round 1: 151`.

Final state: **64,397 B**, 5.2x the 12,288 B target, **still over** the 40,960 B hard cap — the eviction bought **−21.2%**, not the −28.7% this PR description still claims, because every audit round correctly bought back bytes I should not have evicted. Getting under the cap needs the 9 superseded-but-unmarked blocks and the 19,343 B `Gotchas` section, both left deliberately as judgement calls.

PR #1210, opened 2026-09-01
feat(cairn): the store write goes through the API — both writers, including the one §5 says everyone forgets

ZacxDev:
§8 step 7 of `claudedocs/plan-cairn-phase1-cutover.md`, landed **before** the cutover so
that phase 5's `0444` freeze breaks no writer. Per §5's table there are **two** writers, not
one — and the doc says the second is the one everyone forgets.

## 1. `claude/skills/subsystem-index/SKILL.md` — the ONE append protocol

- bullet append → `cairn append --scope … --ref … --session … --text '<one line, no leading
  "- ", no leading date>'`. The server supplies the `- ` and the date; the skill says so,
  because the first production append through this route double-dated itself.
- `OPEN:` → `RESOLVED <sha>:` rewrite (and a `## Pointers` fix) → `cairn put` against a
  scratch copy of the live bytes.
- **Order is load-bearing and now stated**: the `put` goes FIRST, then the `append`.
  Reversed, the caller's own append moves the entry out from under its own `If-Match` and
  the put 412s — a refusal that reads exactly like *someone else's* concurrent edit.
- exit vocabulary carried into the prose: 6 refused · 7 unreachable · 8 the entry moved.
  A write during an outage is refused, never queued.
- an entry with no `## Nuance / work-history` is refused by the server (422), so the heading
  is a `cairn put` that comes **before** the append — the opposite ordering from the old
  "created as part of the append".
- `--validate` now has two forms and the doc says which is which: after an API write, `cairn
  sync` and validate the refreshed cache (the read-only mirror does not move, so validating
  the local path parses the *pre-append* bytes); a brand-new local file is the mirror image.

**Dropped, and named in place as replaced rather than forgotten** — the API's per-entry
`flock` plus an `If-Match` derived from a live sync is the arbiter neither rule had:

- "re-read the file and re-apply to current bytes"
- "use `Edit` anchored on `## Nuance / work-history`, not `Write`"

**Kept**: "show the diff first" (that was never the prompt's job — the transcript is the only
record of what landed), the `already there` comparison (the actual content filter), and
`Write` for a first-ever file.

**Not in scope, stated rather than discovered**: creating a NEW entry file is still a local
`Write` — scope dirs stay `0755` and the API has no create route — so a locally-created
entry is **local-only until the next cutover run classifies it `ADD`**. Known follow-up, not
a break. The hourly commit and 6-hourly backup timers are untouched; the freeze targets
entry files only.

## 2. `claude/skills/prune-index/` — the writer §5 calls "most likely to be forgotten"

- the whole-file cut → `cairn put`; the §5 ref-collision alias drop is the same write (it is
  a one-line edit *inside a frozen entry file*, not an exemption).
- **its y/N STAYS.** The retirement evidence ("the answer was always `y`") was measured on an
  APPEND; a cut removes bytes that are often their content's only copy. Only the mechanism
  moved, and the skill now says exactly that.
- 🔴 **audit / back up / verify now read the SYNCED CACHE, not the frozen mirror.** This is a
  correctness fix, not tidying: the mirror no longer receives writes, so a cut built from its
  bytes would silently delete every bullet appended since the freeze — and the audit's
  `OPEN:` count, the one number §6 compares before and after, would be under-counted on both
  sides.
- `reference/writing-and-safety.md`: step 5 and the "Landing" paragraph updated ("the writes
  are local and final" became false at the cutover).

## Tests — updated in the same commit, each pin watched to fail

`scripts/tests/test_index_append_protocol.py` still sees **exactly one protocol**;
`write-back.md` is untouched, still routes at step 4, and its region hash is unchanged.

- 🔴 the **disagreement predicate was widened** to name the `cairn` verbs. Without that, a
  second protocol phrased in the *new* mechanism named no tool and opened no gate, so it
  would have been invisible to every layer while the ledgers reported a clean set — the
  module would have gone on guarding a mechanism nobody uses.
- `MANDATE` deliberately keeps the retired `Edit`-anchor fragment: the relationship its
  ledger asserts is unchanged, only the *reason* a document may carry it, which is recorded
  in the ledger's reason strings.
- `MANDATED_TOOLS` is now `Bash` + `Write`, not `Edit` — an `Edit` on a `0444` entry is
  EACCES, and pinning it would read as coverage of a rule nothing follows. The negative
  control's width is stated: `resume` has `Bash`, so it discriminates the `Write` half.
- new `TestThePruneDoorMovedItsMechanismAndKeptItsGate` — nine pins across both prune docs,
  asserting the two claims that pull in opposite directions (the mechanism MOVED, the gate
  did NOT). Measured while adding it: deleting the `cairn put` command line from that skill
  left every pre-existing ledger and layer green.
- `test_subsystem_touch.py`: three pins superseded, each with its replacement's own `why`.
- `test_subsystem_recall.py`: the retraction's operational consequence is *inverted* by the
  arbiter, so its pin moved from "no error is not evidence you were alone" to "exit 8 IS the
  other writer".

**Controls**: ten mutants, one at a time, each restored byte-exact (checksummed) and each
killed by ITS OWN pin's error. One found a real defect while being built — a bare `entry
files are 0444` pin SURVIVED deleting the write half's whole sentence, because the owner
spells the freeze twice; it is now pinned as the whole clause.

## Gate

Base sha `049bf036` (`origin/main`).

- dev-host tier — `nix develop <worktree> -c bash scripts/gate.sh`:
  `GATE: RESULT=PASS exit=0`; pytest `RESULT: PASS (exit=0)`, node `RESULT: PASS (exit=0)`
  (5 suites / 41 files / 1449 tests, floor 1367).
- merge-gating tier — `nix build .#checks.x86_64-linux.{pytests,nodetests}`, **one at a
  time**, results in the thread below.

Nothing in this PR was run against the live pod or either host's store.


ZacxDev:
Merge-gating tier, run **one derivation at a time** (a combined invocation yields false REDs via nix store contention), base sha `049bf036` + this branch's two commits:

```
TIER=pytests   NIX_EXIT=0   devrc-pytests> RESULT: PASS (exit=0)
  TOTAL collected=20297  passed=20294  skipped=3  failed=0  (floor: 18404 = sum of 30 per-target floors)

TIER=nodetests NIX_EXIT=0   devrc-nodetests> RESULT: PASS (exit=0)
```

Read from the runners' own `RESULT:` lines in a redirected log, not a pipe.

⚠ Two caveats, stated rather than implied:
- `strict` is false on branch protection, so this is a claim about **this branch**, not about the tree the merge creates. Gating the merged tree is still a manual step.
- other sessions were running their own `nix build` of the same derivation name concurrently on this box. Both tiers came back **green**, and a contended run fails loudly rather than faking a pass — so the greens stand; a red would have needed a serial re-check.

PR #1219, opened 2026-09-01
fix(store tests): one shared store siting + a ledger, because #1211 reached one of three sites

ZacxDev:
`#1211` sited the store on tmpfs in `test_subsystem_store_api.py` **and nowhere else**. The very next PR gated after it merged went red on `TestAppendLands` in `test_cairn_write.py` — a docs-only diff, a file `#1211` never touched, and the same `api.build_server` writing through the same `_replace_bytes` fsync.

Three test files stand up that server. `#1211` fixed one. A predicate open-coded per call site is wrong at every site it wasn't copied to, so this consolidates rather than copying the fix twice more.

| file | before | after |
|---|---|---|
| `scripts/testlib/store_siting.py` | — | new: `mount_fstype` / `tmpfs_dir` / `store_root` |
| `test_subsystem_store_api.py` | local copies | imports the shared module; copies deleted |
| `test_cairn_write.py` | **disk-backed** | sited |
| `test_cairn_cli.py` | **disk-backed** | sited |

Still falls back to the caller's `tmp_path` on every failure mode, so it cannot make any suite worse. That fallback is load-bearing rather than defensive: CI builds **unsandboxed**, so its `/dev/shm` is the container's own mount — 64Mi by default and possibly absent.

## The ledger is the point of the change

Every one of those files was individually fine. What was broken was the **relationship** between them, which no per-file test could see — so `test_store_siting_ledger.py` asserts that relationship and fails when the set **grows** (a new `build_server` caller starts disk-backed and silently rejoins the flake population) or **shrinks** (a stale name quietly weakens the guard).

It checks the **source**, deliberately. A behavioural check ("the store ended up on tmpfs") cannot run where no tmpfs exists — CI may well be such a place — so it would pass vacuously in exactly the environment that matters. `test_subsystem_store_api.py` keeps the behavioural half; the two are different claims and both are needed.

## 🔴 My first version of that guard was a spelled guard, and a mutant survived it

It was `re.search(r"\bstore_siting\b|\bstore_root\s*\(")`. The mutation that matters — strip the import **and** the call, leave the explanatory **comment** that says *"the root comes from `testlib.store_siting`"* — left the regex matching. The file read as compliant with its store back on the contended disk: **3 passed**.

Replaced with an AST check requiring a real import binding **and** a real `store_siting.store_root(...)` call. Both halves, because either alone is satisfiable without the other (an unused import; an unrelated `store_root()`). Re-run with the comment deliberately retained, the mutant now dies.

I only found this by running the mutation rather than trusting the guard — which is the argument for running it every time, not the argument that this one is now fine.

## Verification

Every mutant killed by its **own** test:

| mutant | test that went red |
|---|---|
| new unledgered `build_server` caller | `test_the_ledger_names_exactly_the_files_that_stand_up_the_store_server` |
| ledgered file stops using the siting | `test_every_ledgered_file_takes_its_store_root_from_the_shared_siting` |
| scan regex broken | `test_the_scan_can_actually_SEE_a_build_server_call` |

Baseline and restored: 3 passed.

- all four affected files, `__pycache__` cleared: **746 passed**, rc 0
- tier 2 pytests: `TOTAL collected=20360 passed=20357 skipped=3 failed=0` (floor 18404), `RESULT: PASS (exit=0)`, `NIX_RC=0`
- tier 2 nodetests: `TOTAL suites=5 files=41 tests=1449 pass=1449 fail=0 skipped=0` (floor 1367), `RESULT: PASS (exit=0)`, `NIX_RC=0`

Built one at a time, redirected to files, never piped — a pipe replaces nix's exit status with `tail`'s and truncates the counts, which cost me the pytests signal on `#1211`.

## ⚠ Scope

This does **not** demonstrate the flake is fixed. The verifier is still the flake **rate** over later PRs — and `claudedocs/handoff-gate-flake-store-api.md` rank 1's baseline was measured against a suite where only one of three sites was fixed, so that number needs re-reading before anyone acts on it. Follow-up doc update to come.

https://claude.ai/code/session_015wc9LzSF9voiSPhY4wpb4q


ZacxDev:
Round 1 returned four 🟡 findings and no 🔴. Round 2's fixes are pushed as `22dd33df`. What that round CLAIMS to have addressed, one line each — stated as claims to check, not as arguments that they are correct:

```audit-claims round=2 audited=7d616485..22dd33df
1. F4 — `mount_fstype` reported the shadowed fstype at an equal-depth mount point; claimed fixed by `depth > best[0]` -> `depth >= best[0]` in scripts/testlib/store_siting.py.
2. F3 — a nearly-full tmpfs passed the 5-byte probe and the caller's writes then raised ENOSPC; claimed fixed by an 8 MiB `_MIN_FREE_BYTES` floor plus wrapping the `tempfile.mkdtemp` call that sat outside the try.
3. F3b — the module docstring's "can never make a suite worse" guarantee; claimed corrected rather than restated, and now claims to name the two residual windows it does not close (a tmpfs filling mid-run, and SIGKILL skipping the finally so /dev/shm/devrc-store-* survives).
4. F2 — `_calls_build_server` was a regex that an aliased import walked through and that flagged comment-only mentions; claimed fixed by an AST scan resolving import-as and assignment rebindings plus `getattr(x, "build_server")`.
5. F1 — `scoped_store` (test_subsystem_store_api.py) was still disk-backed while feeding ~110 `running(...)` sites; claimed fixed by siting it through `store_siting.store_root`.
6. F1b — `test_every_ledgered_file_takes_its_store_root_from_the_shared_siting` claimed coverage its body did not provide; claimed corrected by renaming it to `..._IMPORTS_AND_CALLS_the_shared_siting_at_least_once`.
7. F1c — the residual inline `tmp_path / "store"` roots in test_subsystem_store_api.py; claimed NOT closed, and instead held at 18 by a new ratchet test that fails if the count grows.
8. Nits — the dead `found.discard()` kept with its false comment corrected; unused `import tempfile` removed.
9. Deliberately NOT taken — `_uses_siting` still rejects `import store_siting as siting`, `import testlib.store_siting`, and `from testlib.store_siting import store_root`; claimed to fail safe (false offender, never a false pass).
```

Reproducers I ran for each are in `22dd33df`'s commit message with their before/after values. Treat those numbers as claims to re-derive, not as evidence.

Residual scope I could not close and am not implying is covered: the 18 inline sites above, and whether the 8 MiB floor is the right threshold for a container `/dev/shm` under real CI load — that is a judgement, not a measurement.

ZacxDev:
Round 2 verified all nine of round 2's claims and reproduced every self-reported number, then found three new 🟡s. Round 3's fixes are pushed as `06853164`. What that round CLAIMS to have addressed — stated as claims to check, not as arguments they are correct:

```audit-claims round=3 audited=22dd33df..06853164
1. A — `>=`, the free-space floor and the mkdtemp fallback all SURVIVED round 2's mutation battery because nothing pinned them; claimed fixed by a new `TestTheSitingRULESThemselvesArePinned` class, with `_MOUNTS_PATH` added as a module-level seam so the shadowed-mount case is reachable without `unshare -Urm`.
2. A2 — `scoped_store` had no positive control of its own; claimed fixed by `test_scoped_store_is_ALSO_sited_off_the_disk`.
3. B — the ratchet was `.count('tmp_path / "store"')` and four of five spellings walked through it; claimed fixed by an AST predicate counting `tmp_path / <"store"|"src"|Name>` and `tmp_path.joinpath(...)`, with the constant moving 18 -> 20 because the AST sees two variable-spelled sites the text count missed.
4. B2 — the ratchet's `or actual == 0` escape left it fully slack at zero; claimed removed in favour of an equality assertion.
5. C — `_calls_build_server` (AST-only) was NARROWER than the regex it replaced and missed `build_server(` inside a string literal, i.e. a subprocess-driven server; claimed fixed by OR-ing in a scan over `ast.Constant` strings, and claimed NOT to reintroduce the comment-only false positive because a comment is not a Constant.
6. D — a fallback comment restored the absolute "never worse than disk" guarantee the module docstring retracted 120 lines above; claimed reworded to cover only that branch.
7. E — `tmpfs_dir() is None` gained a second cause (under the floor) while the skip message named only the first; claimed corrected to name all four causes and to warn that the skip is also what a silently-inert fix looks like.
8. F — the 8 MiB floor rejected a measured-usable 4 MiB tmpfs, making the fix inert on a good mount; claimed fixed by lowering `_MIN_FREE_BYTES` to 1 MiB.
9. Not taken — the docstring's "two residual windows" still omits cgroup memory charging for tmpfs pages; claimed to be reasoned rather than measured, and immaterial because these stores are kilobytes.
```

Self-reported numbers to re-derive rather than accept: the re-run battery (`>=`→`>` 2 failed, floor 1 failed, mkdtemp 2 failed, baseline/restored 14 passed), the four ratchet spellings all failing, the C reproducer failing, the corpus scan showing no new false positives, the ratchet constant 20, and `755 passed, rc 0` across the four affected files.

Residual scope I am not claiming to have closed: the 18→20 inline sites themselves (still a ratchet, not a conversion), whether 1 MiB is the right floor under real CI load rather than on this host, and the cgroup-charging window in item 9.

ZacxDev:
Round 3 found that round 3's own fix made things worse, and it was right. Round 4's fixes are pushed as `6efca01e`. What that round CLAIMS to have addressed — claims to check, not arguments that they are correct:

```audit-claims round=4 audited=06853164..6efca01e
1. F1 — lowering `_MIN_FREE_BYTES` to 1 MiB was a regression: it sat BELOW the measured 1.199 MiB page-allocated peak and opened an ENOSPC window at 1 MiB <= free < ~1.3 MiB. Claimed fixed by raising it to 4 MiB and recording the peak as a named constant `_LARGEST_STORE_BYTES = 1_257_472`, independently re-measured here (apparent 53,985 B vs page-allocated 1,257,472 B).
2. F2 — both floor guards were structurally blind (one monkeypatches `_MIN_FREE_BYTES` before measuring, the other skips exactly when the defect is present), so `1 << 60` gave 95 passed / 4 skipped / rc 0. Claimed fixed by `test_the_floor_CONSTANT_clears_the_largest_store_this_suite_builds`, which asserts the constant against `_LARGEST_STORE_BYTES`, takes no fixture, monkeypatches nothing, and cannot skip. Bounded both ways (below the peak fails; above a container's 64Mi /dev/shm fails).
3. F3 — the NESTED mount fixture was ordered shallowest-first, so longest-match and last-match agreed on every line and a depth-dropping mutant survived hermetically, being killed only by this host's incidental real `/proc/mounts` order (reversed in the sandbox tier). Claimed fixed by moving `/` last in that fixture.
4. F4 — `test_store_siting_ledger.py`'s comment calling `found.discard()` "provably unnecessary / dead" was falsified by the same commit that wrote it, because widening to `ast.Constant` made the file match its own docstrings. Claimed fixed by rewriting the comment to say the discard is load-bearing, and by narrowing the "a COMMENT is not a Constant" claim to `#` comments only, noting a DOCSTRING is prose AND a Constant.
5. F5 — the string-literal half of `_calls_build_server` had no test (disabling it reverted green at 99 passed). Claimed fixed by two tests: a `build_server(` inside a textwrap block must count, and a `#`-comment mention must not.
6. F6 — the class header "EVERY FIX IN THIS MODULE HAS A TEST" was false (write probe, `is_dir()`, tmpfs `rmtree` all still survive). Claimed rescoped to what it actually pins.
7. F7 — the ratchet's shrink message told the reader to lower the constant, which cannot distinguish "sites converted" from "predicate narrowed". Claimed fixed by making the message require checking which happened.
8. Not taken — F8 (strict equality fires on any `tmp_path / <Name>`), the nits about `"src"` being dead in `_ROOT_NAMES`, `tmp_path / "data"` escaping the ratchet, and the AST predicate remaining walkable by f-string / `os.path.join` / string concatenation. All strictly wider than the text count they replaced, so claimed not regressions.
9. Not taken — the docstring's "two residual windows" still omits cgroup memory charging for tmpfs pages.
```

Self-reported numbers to re-derive rather than accept: the round-3 survivors now dying (floor 1 MiB → 1 failed; floor `1<<60` → 1 failed; string half disabled → 1 failed; baseline and restored 35 passed), the depth-drop and `>=`→`>` mutants both dying hermetically (1 failed and 2 failed), and `758 passed, rc 0` across the four affected files.

Residual scope I am not claiming to have closed: the 20 inline sites (still a ratchet), whether 4 MiB is right under real CI load rather than on this host, item 8's walkarounds, and item 9.

ZacxDev:
Round 4 found one MEDIUM (N1) and three LOW. Round 5's fixes are pushed as `1eafc40c`. What that round CLAIMS to have addressed — claims to check, not arguments that they are correct:

```audit-claims round=5 audited=6efca01e..1eafc40c
1. N1 — `_LARGEST_STORE_BYTES` was pinned to nothing, so growing the fixture silently re-armed the ENOSPC hazard through the guard added to prevent it. Claimed fixed by `test_the_largest_store_constant_still_covers_the_biggest_fixture`, which AST-reads the entry count out of `test_cairn_cli.py` and asserts `count * 4096 + slack <= _LARGEST_STORE_BYTES`; claimed to fail LOUDLY (returning None, not 0) when the scan finds nothing.
2. N1b — the first version of that scan took the max `range(N)` in the whole file and picked up an unrelated `range(60_000)`, demanding a 245 MB floor. Claimed fixed by narrowing to loops whose body contains a `write_text` call, claimed verified to separate the 303-entry store loop (True) from `range(60_000)` and `range(80)` (both False).
3. N2 — the shipped "apparent 53,985 B" and "~23x" were measurements of a SYNTHETIC store, not the real fixture. Claimed corrected to apparent 72,602 B / page-allocated 1,265,664 B / 17.4x, re-measured against `_populate_source_store` + `range(303)` = 306 files, with `_LARGEST_STORE_BYTES` raised to 1,318,912.
4. N3 — the class header carried "(99 passed)", falsified by the very commit that added three tests to that class. Claimed deleted rather than re-stated.
5. N4 — both tests for the ledger's `_calls_build_server` lived in `test_subsystem_store_api.py`, so `pytest test_store_siting_ledger.py` was 4 passed under both regressions. Claimed moved to the ledger file.
6. Nit — the floor's upper bound `<= 64 MiB` admitted exactly the value its own message calls fatal. Claimed changed to `<`.
7. Not taken, unchanged — F8's walkarounds (f-string, `os.path.join`, concatenation), `"src"` dead in `_ROOT_NAMES`, `tmp_path / "data"` escaping the ratchet, and cgroup memory charging for tmpfs pages.
```

Self-reported numbers to re-derive rather than accept: the audit's own N1 scenario now failing (`range(303)` → `range(1200)`, no constant touched → 1 failed), `_LARGEST_STORE_BYTES = 0` → 1 failed (it SURVIVED at round 4), breaking the scan → 2 failed, baseline and restored 8 passed on the ledger file, and `760 passed, rc 0` across the four affected files.

🔴 Two errors of mine this round are recorded in the commit message and are themselves worth checking: the synthetic-vs-real measurement mix-up (item 3), and the over-broad first scan (item 2). Both were caught before push, but both are the same shape as everything this ladder has found.

Residual scope I am not claiming to have closed: the 20 inline sites (still a ratchet), whether 4 MiB is right under real CI load rather than on this host, item 7, and whether the `write_text` discriminator holds if a future fixture populates entries some other way.

ZacxDev:
Round 5 found F1 (the derivation under-reported silently in three shapes) and my replacement measurement was also wrong. Round 6's fixes went out as `734cb0bc`, **and the PR has since merged** — squash `b4fde334`, now on `main`. This block is posted late: the round-6 delta was never audited before the merge, so the audit below runs against **mainline**, and any finding becomes a fix-forward PR rather than a push to this branch.

What round 6 CLAIMS to have addressed — stated as claims to check, not as arguments that they are correct:

```audit-claims round=6 audited=1eafc40c..734cb0bc
1. F8 — the ledger's `tmp_path / <Name>` arm was unrestricted, so it accused any scratch write. Narrowed to expressions actually USED as a store root: bound to a root-ish name, or passed to `_build_store`/`running`/`build_server`. CLAIM: the two real variable-spelled sites (`root = tmp_path / name`, `root = tmp_path / kind`) still count, main's `(tmp_path / name).write_text(body)` scratch write does not, and the merged tree reads 20 == 20.
2. F1a — the peak-bytes derivation took the MAX of one `range(<int>)` + `write_text` loop in ONE file. CLAIM: it now SUMS rather than maxes, and the three under-reporting shapes are closed — two write loops in one test (reported 303, truth 603), nested loops (reported 203, truth 1003), and a `write_bytes` loop beside a `write_text` one (ratchet green while the true need was 21.7 MB against a 4 MiB floor).
3. F1b — the derivation counted only `write_text`. CLAIM: `write_bytes` is now counted too, and every file in `EXPECTED_SERVER_TESTS` is scanned rather than one.
4. F1c — `_SEEDED_ENTRIES = 3` was a bare literal about another file's function, pinned by nothing, added by the commit whose subject was that a constant checked against a constant is not a guard. CLAIM: it is gone, and seeds are derived by counting write calls outside any loop.
5. F1d — `_LARGEST_STORE_BYTES` was a transcribed measurement. CLAIM: it is now the derived requirement — 442 entries + 16 slack pages = 1,875,968 B — computed from the fixture that produces it, not copied from a run.
6. F2 — round 5's replacement measurement ("72,602 B / 1,265,664 B / 17.4x") double-counted DIRECTORY pages; tmpfs dirs cost zero and 1,253,376 + 3*4096 = 1,265,664 exactly. CLAIM: the prose now carries 74,613 B / 1,253,376 B / 16.80x, which four independent methods agreed on (`sum(st_size)`, `st_blocks*512`, `du -sB1`, statvfs delta) and which matches round 4 to the byte.
7. F3 — the "~23x" round 5 said it deleted still shipped in a LIVE assertion message in `test_subsystem_store_api.py`, four lines from an edit that same commit made in the same hunk, alongside "page-allocates 1.199 MiB" — the constant that commit deleted as wrong. CLAIM: both corrected.
8. F4 — the moved test's docstring claimed "99 passed" for a file whose suite is 8, falsified by the move the same commit performed. CLAIM: the count is removed, not restated with a new number.
9. Nit — the dead `import test_store_siting_ledger as led` left by the move. CLAIM: removed.
10. Merge — `origin/main` was merged (the PR had gone CONFLICTING). CLAIM: the conflict was ONE hunk, both sides adding an import, and both were kept.
```

Verification round 6 self-reports, to be re-run rather than believed: a second 600-entry loop and a 5,000-entry `write_bytes` loop each now fail the ratchet; the four affected files on the merged tree with `__pycache__` cleared give **782 passed, rc 0**.

🔴 Two things this round did NOT establish, and which the audit should treat as open: the summing-across-files bound is claimed conservative on the argument that under `-n 4 --dist loadfile` the files do not all hold stores at once — that is an argument, not a measurement; and F8's narrowing predicate is new code introduced by the fix, in exactly the place the last five rounds found things.

PR #1220, opened 2026-09-01
docs(handoff): rank 13 — the CI mitigation is PARTIAL at FIXTURE granularity, not file

ZacxDev:
Rank 13 of `claudedocs/handoff-skill-chain-usage-audit.md`. One doc, +38/-9.

The item was filed as *"add a MITIGATED caveat — the mechanism should no longer fire."* I measured the premise before writing it, and it is wrong. The mechanism still fires, so the caveat says **PARTIALLY**.

## What the measurement found

`#1211` (`1a4350f3`) sites the store on tmpfs so fsync cannot stall inside the request. It reached **one of the three** files that stand up `api.build_server`. On `origin/main` at `b59b0475`:

| file | hits for `tmpfs\|/dev/shm\|store_siting` |
|---|---|
| `scripts/tests/test_subsystem_store_api.py` | 44 |
| `scripts/tests/test_cairn_write.py` | **0** |
| `scripts/tests/test_cairn_cli.py` | **0** |
| `scripts/testlib/store_siting.py` | does not exist on `main` |

Demonstrated rather than predicted: the first PR gated after `#1211` merged went red on `TestAppendLands` in `test_cairn_write.py` against a docs-only diff. The consolidation is **#1219**, still OPEN — so this caveat is deliberately written to expire when that merges.

The caveat also carries the inverse warning, which is the half that is easy to miss: a red on `test_subsystem_store_api.py` is **no longer this mechanism either**, because that file *is* sited. It needs a fresh diagnosis, not this block.

Mechanism text kept verbatim, as the item required — `scripts/ci-repro/README.md`, the reproducer and the code sites are the durable half and none of it is touched.

## Verification

- **Structural control, base vs branch** — `handoff-audit.py --sections 1`: `9/13 → 10/13` ranked items done, `9 → 10` resolved investigations. Both edits register; a green with unmoved counters would have meant the parser never saw them.
- `test_handoff_doc.py` + `test_handoff_audit.py`: **413 passed**, including `TestPushabilityCasesTheFetchVersionGotWRONG` — the class this doc records as intermittently red.
- 🔴 **`test_doc_path_rot.py` passes 77, and that green is VOID for this file.** I planted a dead path in the same doc and it *still* passed 77 ⇒ the suite does not scan `claudedocs/`. Reporting the pair, not the green.

## What I did not do

`scripts/ci-repro/README.md` carries **no mitigation note** (zero hits for `tmpfs|1211|mitigat|shm` on `main`), so the reproducer reads as fully live at the exact place this doc sends readers. That is a devrc-repo edit rather than a handoff-doc edit, so it is out of rank 13's scope and is written into the item with a closing condition instead of being silently absorbed.

## Size

`65,506 → 68,336 B` (+2,830). Over the 40,960 B hard cap either way. Closing one investigation raises evictable bytes `5,328 → 7,906`, so rank 5 gets more to work with, not less.

https://claude.ai/code/session_01W1yap7m8dQtUgtJqZsvSNg

### DIFF
--- a/claudedocs/handoff-skill-chain-usage-audit.md
+++ b/claudedocs/handoff-skill-chain-usage-audit.md
@@ -194,12 +194,12 @@ groups each held BOTH verdicts, and each run carried a distinct `refs/pull/N/mer
   CANCELLATION.** `grep -c 'running(store'` reads **133**, which is the correct answer reached by
   two equal and opposite errors: it **misses 3 line-wrapped** `running(\n store, …)` sites
   (`:8764, :8792, :8832`) and **over-counts 3 lines inside a `textwrap.dedent("""…""")` string**
-  (`:8017, :8033, :8300`). Full partition of the 272 lines containing `running(`: **256** real
+  (`:8017` in `misplaced`, `:8033` in `correct`, `:8300` in a different test's helper). Full partition of the 272 lines containing `running(`: **256** real
   sites · 4 comments · **11 inside string literals** · 1 `def running(`.
   🔴 **This paragraph previously published 136/259 and claimed a walker produced them — both were
   wrong, and the mechanism is the joke of the arc:** a *text* walker that resolved the wrap but not
-  the string counted a `textwrap` block as production code, and that block is **another walker's
-  positive-control fixture**, sitting under the comment *"the detector must be able to SEE a
+  the string counted `textwrap` string bodies as production code — and `:8017` sits in **another walker's
+  positive-control fixture**, four lines under the comment *"the detector must be able to SEE a
   misplaced call, or the empty list above is a fact about the walker and nothing else."* The
   instrument was fooled by the fixture built to test exactly that. Use `ast.walk` over
   `With`/`AsyncWith` items — it cannot see comments **or** strings, and resolves wraps for free.
@@ -508,7 +508,7 @@ EVICTED 2026-09-01 (terminal). Verdict: #1064 green, MERGED `f71ff648`; the two-
   reads the checkout PATH) and **#1214 `07a22f14`**.
 - **Ruled out:** *"the doc is simply wrong"* — it is not; the mechanism and the reproducer are real
   and `scripts/ci-repro/README.md` still documents them. What is missing is the **mitigation**, so a
-  reader will expect a failure mode that ~~should no longer fire~~ **mostly still fires** (see the fixture
+  reader will expect a failure mode that ~~should no longer fire~~ **mostly still fires** (see the fixture/call-site
   table in the CI block — this line's premise is the one that was refuted). via: change
 - **Leading hypothesis:** this is the arc's own recurring shape operating ACROSS PRs rather than
   within one — a claim true at write time, falsified by a sibling change, inside no audit round's
@@ -600,7 +600,7 @@ rather than removed and renumbered. New work is appended at the end.
    `ci-speedup-7` (verified live 2026-09-01). 🔴 **And re-read the premise first: #1211 `1a4350f3`
    sited the store on tmpfs**, so the contention this item was motivated by may already be gone.
    ⚠ **CORRECTED 2026-09-01 — do NOT act on the sentence above without reading the CI block's
-   fixture table first.** #1211 sited **one fixture of two** in one file of three; **123 of that
+   fixture/call-site table first.** #1211 sited **one fixture of two** in one file of three; **123 of that
    file's 256** `running()` sites — the 110 in `scoped_store` PLUS 13 that belong to no fixture,
    three of them write-path — plus both cairn suites are still disk-backed, and the tmpfs probe is UNVERIFIED on
    `talos-xr6-r7p`. So the contention is **not** established as gone, and de-prioritising this item
@@ -635,8 +635,11 @@ rather than removed and renumbered. New work is appended at the end.
    undone; it also leaves the *second* half of the caveat standing, because `store_siting.py`'s
    probe keeps a disk fallback and nobody has verified it engages on `talos-xr6-r7p`.
    **(a) = that README names `1a4350f3` AND says what is STILL unsited — and "still unsited" is
-   three separate populations, so it is three greps.** 🔴 An alternation would certify all three by
-   naming one, which is this arc's own recurring bug expressed as a regex:
+   THREE populations (`scoped_store`, the two cairn suites, the 13 unfixtured sites), so with the
+   sha that is FOUR required greps.** 🔴 An alternation would certify all four by naming one, which
+   is this arc's own recurring bug expressed as a regex. ⚠ **This sentence said "three greps" until
+   round 5 — one round AFTER a4 was added below it, so the definition of (a) contradicted its own
+   check for a whole round. Count the `# (aN)` lines in the block; they are the authority.**
    ```bash
    DEVRC=~/workspace/devrc
    git -C "$DEVRC" fetch origin main -q      # or a stale clone answers for a ref you do not have
--- a/claudedocs/handoff-skill-chain-usage-audit.md
+++ b/claudedocs/handoff-skill-chain-usage-audit.md
@@ -46,9 +46,10 @@ survives adversarial re-derivation, and fix whatever it exposes.
   `home-manager switch` was deliberately skipped rather than run as a no-op riding another
   session's uncommitted `nix/programs/alacritty/default.nix`. Rank 11 forward-merged `main` into
   #1064 and its gate went green.
-- **OPEN: ranks 5, 8, 10, 12.** Derive, don't trust this line —
+- **OPEN: ranks 5, 8, 10, 12** (rank 13 DONE 2026-09-01). Derive, don't trust this line —
   `python3 scripts/handoff-audit.py --sections 1 <this doc>` from the devrc repo root prints
-  `N/12 ranked items done` (8/12 at 2026-09-01).
+  `N/13 ranked items done` (**10/13** at 2026-09-01; the `8/12` this line carried was already stale
+  by two items and one denominator when read — the count and the DENOMINATOR both move).
 - 🔴 **CARRIED FORWARD — durable specifics the condensed status above would otherwise drop.**
   *Ranks 2+7, the hook:* devrc**#1092** `ad891a5c`; `ship.sh` converged BOTH hosts to **`bd1572f3`**
   (verified by ancestry of the shipped sha PLUS a byte check of the deployed blob **`6d25558e`**);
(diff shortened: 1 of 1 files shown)


ZacxDev:
## 🔴 Correction to the PR description above — the "inverse warning" paragraph was wrong

A blind adversarial audit refuted a claim I made in this PR's description, so correcting it here rather than silently editing the body — a reviewer may already have read the wrong version.

**What the description said:**

> The caveat also carries the inverse warning, which is the half that is easy to miss: a red on `test_subsystem_store_api.py` is **no longer this mechanism either**, because that file *is* sited.

**That is false.** The file has **two** store fixtures and `#1211` sited one:

| site | sited on tmpfs? | `running(...)` sites |
|---|---|---|
| `store` (`:393`, calls `_tmpfs_dir()`) | **yes** | 133 |
| `scoped_store` (`:9097`, `_build_store(tmp_path / "store", …)`) | **NO** | 110 |

So the granularity is the **fixture**, not the file — meaning the fix for a "one of three files" error committed the same error one level down, inside the very PR written to warn about it.

`#1219`'s own code states it outright (`22dd33df:scripts/tests/test_subsystem_store_api.py:9033-9037`): *"It was missed when #1211 sited `store` alone."* That is the PR this description cites as the consolidation.

**The harm was live, not hypothetical.** `#1216` (`1e5942a8`) — a diff of one markdown file — went `devrc-pytests FAILURE` on `TestTheActorComesFromTheTOKEN`, a class that runs on `scoped_store` and is the same class `scripts/ci-repro/README.md:79,83` uses as its reproducer. A reader hitting that red and following my paragraph would have opened an investigation into a defect that does not exist — the exact harm the surrounding block exists to prevent.

Fixed in `d02027ce`, along with five smaller findings (a `10 + 4 = 14 ≠ 13` count contradiction traceable to a `\bDONE\b`-on-first-line parser false positive that silently closes rank 5; a surviving `N/12` at a second site; an unsupported "~3 h" that re-derives to 41m36s; "two items" → one; and a closing-condition command that tested something other than its own condition).

One more correction while I am here: my description's verification list is otherwise reproduced exactly — the auditor re-derived every count, sha, timestamp and byte figure and all matched — but I want to flag that the audit also cited `#1214` as a second instance of the misrouting. **`#1214` currently reads `SUCCESS`**, so I have not used it as evidence; only `#1216` is cited.

ZacxDev:
Round 1 audit claims — what the fix commit `d02027ce` claims to have addressed. Stated as claims to check, not as reasons they are correct.

```audit-claims round=1 audited=afe68628..d02027ce
1. The false inverse warning ("test_subsystem_store_api.py is now sited, so a red there needs a fresh diagnosis") is removed and replaced with a fixture-granularity table asserting store (:393) sited / scoped_store (:9097, 110 running sites) NOT sited.
2. The claim "removes the mechanism where it was applied" is downgraded to "wherever the tmpfs probe succeeds", with _tmpfs_dir()'s disk fallback and the unverified CI-node state stated.
3. #1216 (1e5942a8) is cited as the live docs-only misrouting instance; #1214 is NOT cited as evidence.
4. The State-now ranked count now states 9 done / 4 open counted from the list, and names the handoff-audit.py \bDONE\b-first-line false positive on rank 5 as why the tool prints 10/13.
5. The second "N/12" site in the How-to-verify section no longer restates the ranked-item number and defers to the single caveat in State now.
6. The "~3 h before anyone worked it" figure is replaced with measured timestamps: filed 21:00:22Z, worked 21:41:58Z (41m36s), #1219 opened 21:11:38Z.
7. The "stale by two items" clause is corrected to one item and one denominator.
8. Rank 13's closing condition now has two explicitly labelled alternatives and a fenced command block that fetches first, greps for 1a4350f3|scoped_store, and checks #1219's state.
9. Two now-false sentences inside the CLOSED cross-PR block ("i.e. the mechanism is mitigated", "a failure mode that should no longer fire") are struck through and annotated rather than left unmarked.
10. Nothing durable was removed: the mechanism text, the ci-repro/README.md pointer and the code sites are unchanged.
```

ZacxDev:
Round 2 audit claims — what fix commit `69fa62b9` claims to have addressed. Claims to check, not reasons they are correct.

```audit-claims round=2 audited=d02027ce..69fa62b9
1. Rank 10 (:561) no longer carries the refuted file-granularity framing un-caveated; a CORRECTED note now points at the fixture table and says the contention is NOT established as gone.
2. Rank 13's closing condition (a) is now two separate greps that must BOTH be non-zero (1a4350f3, and scoped_store|unsited), matching its AND prose, with a stated positive control of grep -c fsync = 25.
3. Alternative (b) "#1219 merges" is REMOVED as a closing condition, on the ground that #1219 does not touch scripts/ci-repro/README.md and leaves the unverified-probe half standing.
4. Both strikethrough annotations now name the fixture table in the CI block directly instead of deferring to the CLOSED verdict, which itself defers.
5. The durable evidence round 1 dropped is restored: the 44/0/0 hit counts, the absence of scripts/testlib/store_siting.py, the b59b0475 measurement anchor, and the TestAppendLands / #1213 063b02be instance for the cairn half.
6. "still live nearly everywhere it was live before" is replaced by the measured "110 of that file's 243 running() sites plus both cairn suites".
7. The PR title no longer says "one of three sites".
```

ZacxDev:
Round 3 audit claims — what fix commit `ea664c31` claims to have addressed. Claims to check, not reasons they are correct.

```audit-claims round=3 audited=69fa62b9..ea664c31
1. The table now partitions by CALL SITE: 259 real `with running(...)` sites = 136 store (tmpfs) + 110 scoped_store (disk) + 13 UNFIXTURED (disk), with all 13 line numbers listed and 3 of them named as write-path (post_bullet/PUT).
2. The doc now states 123 of 259 sites are disk-backed, replacing "110 of that file's 243".
3. The doc now states #1219 does NOT reach the 13 unfixtured sites because it re-sites fixtures only, and tells the reader to assume a fourth granularity.
4. The naive-grep trap is recorded: `grep -c 'running(store'` reads 133 and misses 3 line-wrapped sites; 4 of the 272 `running(` lines are comments.
5. Closing condition (a) is now THREE separate greps (1a4350f3, scoped_store, cairn) that must all be non-zero, replacing the alternation that could certify three populations by naming one.
6. The positive control (`grep -c fsync` = 25) now runs INSIDE the fenced block and runs FIRST.
7. The H3 heading of the closed cross-PR block now says PARTIALLY MITIGATED and carries a correction note stating headings are machine-read.
8. That block's "Next probe, verbatim" is replaced with "none — closed" plus an explicit warning not to run the old instruction, which told the reader to add a flat MITIGATED caveat.
9. Rank 13's `forcing:` line now says PARTIALLY mitigated.
10. The size trail is pinned to fixed shas only; `384e1c37`/`main` 64,837 is corrected, `66eee1d7` 65,506 added, and the stale moving-ref label is called out. The `## State now` 64,837/5.3x figures are explicitly flagged as describing #1207's tree, not the reader's.
```

ZacxDev:
Round 4 audit claims — what fix commit `5dc20213` claims to have addressed. Claims to check, not reasons they are correct.

```audit-claims round=4 audited=ea664c31..5dc20213
1. The partition is corrected to 256 real `with running(...)` sites = 133 store + 110 scoped_store + 13 unfixtured, and 272 = 256 sites + 4 comments + 11 string-literal + 1 def.
2. The naive-grep note now states BOTH errors — grep -c 'running(store' = 133 is right by cancellation, missing 3 line-wrapped sites and over-counting 3 inside a textwrap.dedent block at :8017/:8033/:8300 — and directs the reader to an AST walker.
3. The false provenance sentence ("the 136/110/13 split comes from a walker that resolves the wrap and excludes comments") is replaced with an account of why the text walker was wrong, naming the positive-control fixture it was fooled by.
4. The CI block's lead bullet headline now says the unit is the CALL SITE, not the FIXTURE, and flags that it read FIXTURE until round 4.
5. The superseded-label cross-reference at :504 now says CALL-SITE granularity and names the fixture/call-site table.
6. Rank 10 now says 123 of 256 disk-backed, naming the 13 unfixtured sites and the three write-path ones, replacing the 110 that understated the population.
7. Rank 13's DONE text no longer re-asserts FIXTURE as the granularity and states the call-site figure.
8. Closing condition (a) now requires a FOURTH grep (a4, 'unfixtured|11755') for the unfixtured population, promoted from an advisory to a requirement, with the prose updated to "a1, a2, a3 AND a4".
9. A mechanical sweep was run across 259/136/256/133/123/110; the only surviving 259/136 in the doc is the correction note that quotes them.
10. All operationally load-bearing figures are unchanged and re-confirmed: 123 disk-backed, the 13 line numbers, and the three write-path sites :11755/:11852/:12584.
```

ZacxDev:
Round 5 audit claims — what fix commit `b6601873` claims to have addressed. Claims to check, not reasons they are correct.

```audit-claims round=5 audited=5dc20213..b6601873
1. The bolded sentence defining closing condition (a) now says FOUR required greps (the sha plus three unsited populations), matching the four (aN) lines in the block below it, and carries a note that it said "three greps" until round 5.
2. "An alternation would certify all three" is corrected to "all four".
3. The over-attribution is fixed per-line: :8017 in `misplaced` (the positive control), :8033 in `correct` (that test's negative control), :8300 in a different test's helper — replacing "that block is another walker's positive-control fixture".
4. The claim that the walker "counted a textwrap block as production code" is narrowed to "counted textwrap string bodies as production code", with only :8017 attributed to the positive-control fixture.
5. Two sibling cross-references that still read "the fixture table" now read "the fixture/call-site table"; all five sites in the document agree.
6. No live self-contradiction about the grep count remains — the two surviving "three greps" strings are historical correction notes.
7. All figures are unchanged from round 4 and still correct: 256/133/110/13, 123 disk-backed, the 13 line numbers, the three write-path sites, 9 done / 4 open / 13.
```

ZacxDev:
Round 6 audit claims — what fix commit `3873c654` claims to have addressed. Claims to check, not reasons they are correct.

```audit-claims round=6 audited=b6601873..3873c654
1. The drifted citation is sha-pinned to `8c108d8b:scripts/tests/test_subsystem_store_api.py:100-105` and names the string `Why 60 and not 15` as a non-moving anchor, with the drift itself recorded (at b59b0475 the old range holds a DRAIN-bounds comment and the capacity text is at :106).
2. Rank 5 now says the tool reports 10 resolved investigations, not 8, and states that only 8 of the 10 are one-line tombstones while two are full live content — so an automated EVICT_HISTORY would delete two blocks that were never tombstones, making the hazard wider than the item previously stated.
3. Rank 5 records that this PR's own first commit (afe68628) moved that count from 9 to 10 by adding the rank-13 H3.
4. The "i.e. `_replace_bytes` itself" gloss is replaced: the three sites are described as driving the WRITE endpoints, with only :11755 asserted to reach `_replace_bytes` (200/appended), :11852 refused at 412 and :12584 at 422 entry-shape, both before the call at server.py:2211.
5. The former over-certification is recorded in place rather than silently removed, and the SET of three is still named as correct.
6. The closing-condition control is now a POSITIVE control ("must be NON-ZERO (25 today)") rather than the equality "must be 25", and the pass condition says "the control is non-zero".
7. A note records why the equality was wrong: the edit that satisfies (a) is a mitigation note about an fsync bug, which plausibly contains "fsync" and moves the count to 26.
8. A note records that #1219 re-sites scoped_store AND both cairn suites, so a2/a3 name populations that stop being "still unsited" once it merges, and identifies a4 as the durable half of the check.
9. All previously verified figures are unchanged: 256/133/110/13, 123 disk-backed, the 13 line numbers, the size trail, 9 done / 4 open / 13.
```

ZacxDev:
## Audit ladder CLOSED at round 7 — and why it stopped here rather than converging to zero

Recording this explicitly, because a report that ends on the stated criterion is otherwise indistinguishable from one that ran out of findings, and those are opposite meanings.

### The ladder

| round | findings | payload lines fixed |
|---|---|---|
| 1 | 🔴 1 · 🟡 4 · 🟢 0 | 92 |
| 2 | 🟡 3 · 🟢 2 | 56 |
| 3 | 🟡 3 · 🟢 2 | 67 |
| 4 | 🟡 3 | 48 |
| 5 | 🟡 1 · 🟢 1 | 17 |
| 6 | 🟡 2 · 🟢 2 | 33 |
| 7 | 🟢 1 | 13 |

**Round 6's delta was clean** — all seven of round 5's claims fixed, no regression at any sibling. Its four findings came entirely from the once-per-ladder whole-document re-derivation, and three of those were claims staled by commits sitting inside *no round's range* (one by this PR's own first commit). That is the pre-merge mandate working as designed, not an eighth round of the chain.

### Why it stops now

The payload is prose, so the zero-payload attribution gate is **structurally inert** — every round is 100% payload by construction and the gate cannot fire. The stop criterion is therefore the stated one, and all three parts are met:

- **no 🔴** — round 7 returned one 🟢;
- **no blast radius beyond "the document contains a false line number"** — the conclusion was unaffected either way: neither site reaches `_replace_bytes`, so a4 stays mandatory;
- **the recurring shape swept at every site** — both write sites are now named per verb, from function bounds I re-read out of `server.py` rather than by patching the one number the audit named.

### The finding that recurred every single round

**A fix lands where it was reported and not at its siblings; and claims certify rather than omit.** Seven for seven. The granularity of the central claim was wrong four times running — file → fixture → call site → the call-site *count itself* — and each correction was then found un-swept at one to four sibling sites. Round 7's 🟢 is the purest instance: round 6's fix for a citation-into-the-wrong-function bug reintroduced that same bug in the sentence that fixed it.

The sharpest single artefact: my round-3 walker counted three lines inside a `textwrap.dedent("""…""")` block as production call sites. That block is **another walker's positive-control fixture**, sitting four lines under the comment *"the detector must be able to SEE a misplaced call, or the empty list above is a fact about the walker and nothing else."* The instrument was fooled by the fixture built to catch exactly that.

### Open, not absent

- Two of the eight "one-line tombstones" are 3 and 5 lines.
- The sha-pinned range `:100-105` truncates its paragraph by one line (the grep anchor makes it self-correcting).
- 🔴 **The "495 passed" figure is a true claim about the tree and is NO evidence about this file's prose.** All four suites are synthetic-fixture or `claude/`-corpus based, and `test_doc_path_rot.py` deliberately excludes `claudedocs/` (`:165-170`). No gate in this repo validates this document's prose or its path references; every citation in it was checked by hand.
- The document is **79,511 B**, up from 65,506. This ladder added ~14 KB of corrections to a doc whose open rank 5 exists to shrink it. That tension is real and unresolved.

Nothing here is a merge blocker. Recording them so the next reader knows they are open rather than absent.

PR #1222, opened 2026-09-01
fix(subsystem-store-api): the byte-identity verifier compared an MTIME-ORDERED render — it could not pass for any scope with more than two entries

ZacxDev:
Rank 11 of `claudedocs/handoff-cairn-phase3.md` — "🔴 OPEN — `verify-byte-identity.sh` STILL cannot pass for a multi-entry scope: it is mtime-ordered".

## The defect

`verify-byte-identity.sh` compared the **whole-scope digest**. `subsystem_recall` orders its INDEX newest-first by entry-file **mtime** and picks the digest's one featured BODY the same way (`select_featured`'s most-recent fallback), and the transport (`seed.sh`: `rsync` into a stage, `tar` into the pod) does not carry mtime. So two stores holding byte-identical entries render a different row order **and** a different featured entry.

Measured 2026-09-01 against the live pod over a `kubectl port-forward`, store `~/.claude/analyze-service-index`:

```
FAIL scope=devrc              raw-diff-lines=45  accounted-for=6
FAIL scope=cli                raw-diff-lines=8   accounted-for=6
PASS scope=storage-resolver   (2 entries)
FAIL scope=homelab-infra      raw-diff-lines=108 accounted-for=6
FAIL scope=datapacket-talos   raw-diff-lines=336 accounted-for=6
verify: scopes=5 pass=1 fail=4
```

Every PASS had 2 entries; the `cli` scope's entire unaccounted difference was one row's POSITION. #1214 closed the `host:` half only. Until now the cutover's P4 was a **permanently-red gate** — which `claude/RULES.md` calls worse than no gate.

## What changed

The claim is restated at the level it can hold. Per scope:

1. **the `mode=list` render** — no body, so no mtime-selected featured entry — with the index **rows compared as a sorted set** and the rest of the stream compared verbatim, in order. Only the rows are reordered, because only the rows are ordered by mtime; sorting the whole stream would make it a multiset that passes when a line moves between sections.
2. **the entry SET**, by `comm` over the refs the two index blocks list. A ref on one side only FAILS and **names it**. It runs *before* the row comparison on purpose: the index head names the entry count, so a set difference would otherwise always surface as a render difference and this arm would be an unreachable guard printing a message nobody sees.
3. **each entry's bytes**, as its own `--ref` / `?ref=` render — a narrowing that prints no index, so it carries no order. There is no route serving raw entry bytes; the API only ever returns renders.

A scope whose index **paginates** is REFUSED, not partly compared: past `LISTING_PAGE_SIZE` both the page membership and the ref column's width (`max(len(ref) …)`, computed per page) are mtime-derived, and the refs past page 1 are never read at all. No scope in the store is near the cap today (largest measured: 50 entries).

## Accounting

`index-order-lines` joins the printed sum **in the same commit as the sort** — a rule that erases a difference without a matching count widens the blind spot rather than the gate. It is **measured** (the diff reduction the sort bought), not assumed, and the identity is now four-term and exact on a PASS:

```
raw-diff-lines == store-root-lines + host-lines + snapshot-block-lines + index-order-lines
```

The counts are **sums over every stream compared for the scope**, so a pod-shaped remote reads `store-root-lines = 2 * (1 + entries)`. The tests assert that multiple rather than a constant, so a per-entry stream that skipped a `sed` cannot hide behind the scope-level stream's count. `entries=` is printed for the `drift-check.sh` reason: a scope that compared no bodies must read as one.

The numbers stay **evidence, not a second gate**.

## Test matrix

Both pre-existing controls kept. Five tests added.

**Red at base `66eee1d7`, green at HEAD** — all five. The base failure for the regression test is the defect itself:

```
FAIL scope=cable-tray canonicalised-cmp=differs raw-diff-lines=12 store-lines=2 host-lines=0 snapshot-block-lines=2 accounted-for=4
  7,8d8
  <   duct-west      1 nuance   client-confidential (declared: internal)
  <   duct-east      1 nuance   client-confidential (declared: internal)
  9a10
  >   duct-east      1 nuance   client-confidential (declared: internal)
  10a12
  >   duct-west      1 nuance   client-confidential (declared: internal)
  14c16
  <   ### duct-west  (cable-tray/duct-west.md, …)
  ---
  >   ### clamp-south  (cable-tray/clamp-south.md, …)
```

— rows reordered **and** a different featured entry, on stores whose bytes are identical.

mtimes are set with explicit `os.utime`, never inferred from write order. The regression test carries its **own positive control**: it asserts the two index orders genuinely differ *and* the two sets genuinely match before running the comparator, so it cannot go quietly vacuous.

**Mutation sweep** — 9 narrow mutants, run under `PYTHONDONTWRITEBYTECODE=1`, control green first, **9/9 KILLED**, each by its expected test with that guard's own assertion:

| mutant | guard | verdict |
|---|---|---|
| M1 local row `sort` → `cat` | the reordering rule | KILLED — `MULTI_ENTRY…still_PASSES` |
| M2 `comm -13` → `""` | the set arm's pod-only mirror | KILLED — `ONE_SIDE_ONLY` |
| M3 entry `cmp` → `if false` | per-entry bytes **(positive control)** | KILLED — `ONE_CHARACTER_divergence` + 2 |
| M4 pagination `if` → `if false` | the paginated refusal | KILLED — `PAGINATED_index_is_REFUSED` |
| M5 `residual` drops `frame_diff` | the scope-level frame comparison | KILLED — `LEADING_BLANKS` |
| M6 `index_order=0` | the measured order accounting | KILLED — `MULTI_ENTRY…` (`index-order-lines=0 — the sort accounted for nothing`) |
| M7 per-entry `store_lines` not accumulated | per-stream canonicalisation | KILLED — `POD_SHAPED…PASSES` (`no store-root difference on all 2 streams`) |
| M8 `rows_diff=0` | the index-row content comparison | KILLED — `VISIBLE_IN_THE_INDEX_ROW` |
| M9 `entries=0` | the `entries=` evidence field | KILLED — `ACCOUNTED_FOR` + 2 |

M8 **survived the first sweep** — every field on an index row is derived from the body, so the per-entry arm catches the same difference a step later and `returncode == 1` proves nothing there. A test was added that asserts the difference was seen **by the row comparison** (`sorted-row-lines > 0`), because that count is what keeps a genuine row difference out of the `index-order-lines` excuse. The script now says so where the comparison happens, rather than claiming coverage it does not have.

## Flagged, not fixed

After the phase-1 cutover the **pod is canonical and each host's local store is a read-through cache that can legitimately lag** — measured 2026-09-01, scope `devrc` at 26 entries locally against 29 on the pod, with nothing wrong. So an entry-set difference is an ordinary operational state *in general*, and this script is only a meaningful acceptance check **immediately after a seed/push**, which is exactly where `cairn-cutover.py` runs it (P4). Said plainly in the script header, in the P4 docstring, in the README, and in the set-arm test's docstring.

The set arm is **deliberately not weakened** for it: a check that tolerated a missing entry could not detect a half-copied seed, which is the one thing P4 exists to catch.

⚠ Cost: the verifier is now **N+1 requests per scope**, not 1 (a 50-entry scope issues 51). One local `--ref` render measures ~73 ms; the whole real store is ~400 entries. P4's `timeout=600` is what bounds it.

## Prose updated in the same commit

Nine of eleven findings in this effort's audit ladder were prose contradicting code, so:

- `cairn-cutover.py` P4 docstring — named **three** differences, and did not mention the digest was no longer compared.
- `server.py` header — claimed it "asserts the raw diff is exactly one line", **false since `host:` shipped** and false through two more causes.
- `README.md` — "Why byte-identity is asserted modulo **THREE** named differences" → four, plus the arms, the pagination refusal, the lagging-cache caveat and the request cost.
- `claudedocs/plan-cairn-phase1-cutover.md` — the instrument's control-table row.

## Gates run

See the comment below for the tier + base sha of each run.

https://claude.ai/code/session_01CnDQ8q2rop6dPiDbxX1twi

### DIFF
--- a/claudedocs/plan-cairn-phase1-cutover.md
+++ b/claudedocs/plan-cairn-phase1-cutover.md
@@ -603,7 +603,7 @@ answer. A green from an unvalidated instrument is a claim about the instrument.
 
 | instrument | negative control | positive control |
 |---|---|---|
-| `verify-byte-identity.sh` | one entry mutated by a single character → **FAIL, naming the scope** | identical stores → **PASS**, printing `raw-diff-lines`/`store-root-lines` beside the verdict |
+| `verify-byte-identity.sh` | one entry mutated by a single character → **FAIL, naming the scope**; the same mutation under a shuffled mtime order → still FAIL; an entry on one side only → FAIL **naming the ref** | identical stores → **PASS**; identical stores whose entry-file **mtimes are in a different order** → **PASS** (the mtime-ordering red), printing `raw-diff-lines`/`store-root-lines`/`index-order-lines`/`entries` beside the verdict |
 | the backup precondition | absent CronJob, absent `kubectl`, unparseable JSON, unparseable timestamp, and a `lastSuccessfulTime` that is **absent** → all refuse with their own sentence | a fresh timestamp passes and prints the measured age beside the ceiling; the boundary is measured from **both** sides (35.5 h passes, 36.5 h refuses) |
 | the write-route probe | a `405 read-only` server → reported as an **operator** problem | a `400 bad-request` server → reported as deployed. The probe body is `{}`, refused by the server's validator *before* any ref is resolved, so it cannot write |
 | the freeze | `probe_writable` returns `writable` at 0644 and `refused` at 0444, on the same file, twice | a partially-applied freeze (one file left writable) exits 16 **and** restores every mode bit |
--- a/scripts/cairn-cutover.py
+++ b/scripts/cairn-cutover.py
@@ -1550,13 +1550,39 @@ def _acceptance(args: argparse.Namespace, cache_dir: Path) -> int:
     about the push and this is a claim about the STATE.
 
     `verify-byte-identity.sh` is then run unmodified over every scope. Its own
-    header documents the two controls that make its green meaningful, and its
-    PASS lines print `raw-diff-lines` beside the verdict, decomposed into
-    `store-root-lines` + `host-lines` + `snapshot-block-lines` and summed as
-    `accounted-for`, so a reader can see the canonicalisation was spent where it
-    claims. 🔴 THREE named differences, not one: this sentence used to name only
-    the store root, which was already false when `host:` shipped and is exactly
-    the difference that made that script permanently red.
+    header documents the controls that make its green meaningful, and its PASS
+    lines print `raw-diff-lines` beside the verdict, decomposed into
+    `store-root-lines` + `host-lines` + `snapshot-block-lines` +
+    `index-order-lines` and summed as `accounted-for`, so a reader can see the
+    canonicalisation was spent where it claims. 🔴 FOUR named differences, not
+    one: this sentence used to name only the store root, which was already
+    false when `host:` shipped, and it still named three after `index-order`
+    joined — each omission being exactly the kind of difference that made that
+    script permanently red.
+
+    🔴 IT NO LONGER COMPARES THE WHOLE-SCOPE DIGEST, AND P4's READING CHANGES
+    WITH IT. The reader's index is newest-first by entry-file MTIME and the
+    digest's one featured body is chosen the same way, while the transport
+    (`seed.sh`: `rsync` to a stage, `tar` into the pod) carries no mtime — so
+    the old whole-scope `cmp` failed on stores that were byte-identical
+    (measured 2026-09-01 against the live pod: `scopes=5 pass=1 fail=4`, the
+    lone PASS being the only two-entry scope). It now compares, per scope, the
+    `mode=list` render with its index ROWS sorted, the entry SET via `comm`,
+    and then EACH entry's own single-ref render. A `FAIL … entry SET differs`
+    names the refs.
+
+    ⚠ THE SET ARM IS AN ACCEPTANCE CHECK, AND IT IS ONLY ONE HERE. After this
+    cutover the pod is canonical and each host's store is a read-through cache
+    that may legitimately lag — measured the same day, scope `devrc` at 26
+    entries locally against 29 on the pod, with nothing wrong. Run at P4,
+    immediately after the push, a set difference means the push was lossy; run
+    at an arbitrary moment it may only mean the cache is behind. The arm is
+    deliberately not weakened for that: a check that tolerated a missing entry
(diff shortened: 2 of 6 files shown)


ZacxDev:
## Gates run — tier and base sha named

All three ran against commit **`d2fa8a88`**, which is this branch's single commit on top of **`b59b0475`** (`origin/main` at rebase time).

| tier | command | verdict |
|---|---|---|
| dev-host | `nix develop <worktree> --command bash scripts/gate.sh` | `GATE: RESULT=PASS exit=0` — pytest `RESULT: PASS (exit=0)`, node `TOTAL suites=5 files=41 tests=1449 pass=1449 fail=0` |
| nix sandbox — **the tier Tekton gates on** | `nix build .#checks.x86_64-linux.pytests` | `devrc-pytests> RESULT: PASS (exit=0)`, 0 `FAILED` lines, no timeout panic |
| nix sandbox | `nix build .#checks.x86_64-linux.nodetests` | `devrc-nodetests> RESULT: PASS (exit=0)` — `TOTAL suites=5 files=41 tests=1449 pass=1449 fail=0 (floor: 1367)` |

The two check derivations were built **one at a time**, never in one invocation — a combined call contends on the nix store and produces false failures (CLAUDE.md, measured 2026-08-30). Verdicts read from the runners' own `RESULT:` lines, not from a piped exit code.

Targeted subset, run separately: `test_subsystem_store_api.py` + `test_cairn_cli.py` → **719 passed**.

### What these runs do NOT cover

- ⚠ `origin/main` moved to `836bac03` while these were running. The delta is **one file, `claudedocs/handoff-tmux-webapp.md`**, with no overlap against anything this PR touches — but the runs above are a claim about `b59b0475 + this commit`, not about that tree. Branch protection has `strict: false`, so a green check is a claim about the branch, never about the merge result.
- No run against the **live pod**. The measurements quoted in the diff (`scopes=5 pass=1 fail=4`; `devrc` 26 local vs 29 served) were taken before this change; the fixed verifier has not been pointed at the pod. That is the acceptance step, and it belongs to whoever next runs `cairn-cutover.py --apply`.
- The `--ref`-per-entry request cost (N+1 per scope) was measured only for the **local CLI leg** (~73 ms/render, 10 runs). The pod's per-request cost over a port-forward was not measured.

PR #1227, opened 2026-09-01
fix(tmp-churn): the /tmp retention rules aged no directory at all — plus the disk-accounting instrument that found it

ZacxDev:
## What this is

The workbench root filesystem sat at 82% with an unexplained ~900 GB. A prior session closed that as **\"ext4 metadata overhead — SOLVED/CONFIRMED\"**. That was wrong, and the real answer turned out to be `/tmp`.

## The refutation

ext4 metadata consumes **blocks**, never **used inodes** — `df -i` IUsed counts files, directories and symlinks (plus ~11 system inodes). So 96–97M used inodes are 96–97M real files. Static metadata computed from the superblock, with `Inode size: 256` read from `dumpe2fs`:

| | |
|---|---|
| inode tables (122,036,224 × 256 B) | 29.1 GiB |
| block bitmaps | 0.06 GiB |
| journal | 1.0 GiB |
| **total** | **~30 GiB — not 902 GiB** |

**The answer:** `/tmp` holds **78,501,285 entries / 469 GiB — 81% of the filesystem's inodes**, on the root partition rather than tmpfs.

**How the wrong answer happened — two mechanisms, the second larger.** (a) `scripts/diagnose-nix-disk.sh` ran `find "$d" 2>/dev/null` unprivileged, so every root-only tree was skipped silently: `/var` measured 980 inodes / 6.1 GiB with 32 denials, against 2,785,594 / 535.9 GiB as root. (b) `/home` had **zero** denials and was still recorded at ~200 GB against 691 GiB — those came from **truncated scans recorded as totals**.

## The real bug: `m:7d` aged no directory

`nix/system/apply-tmp-churn-retention.sh` was written 2026-08-15 and **had never been run**. Its diagnosis was right; its rule spelling was not.

Per `tmpfiles.d(5)`: *\"The lower-case letter signifies that the given timestamp type should be considered for FILES, while the upper-case letter signifies ... for DIRECTORIES.\"* Default `abcmABM`. So `m:` selects mtime for files and **no criterion at all for directories**. Measured on a fixture:

```
m:7d   -> removes ./file.txt, ./sub/inner.txt   ... LEAVES ./sub
mM:7d  -> removes those AND ./sub
7d     -> removes nothing (default includes ctime)
```

It emptied every directory and removed none — on a filesystem whose problem is inodes.

## Other defects fixed

- **Appending was not an upgrade path.** systemd takes the *first* line per path (`Duplicate line for path, ignoring`). An old `m:7d` above a new `mM:7d` made the fix fully inert. The script's verifier could not see it: it proves the **text** is on disk, never that the rule is **in effect**. Now evicts stale lines first.
- **One rule ledger.** The rule set was hardcoded twice (a Python inserter, a bash verifier) with nothing tying them together.
- **`--emit-rules`** — measured on the workbench: the config edit built correctly but `nixos-rebuild switch` was blocked by an unrelated `dbus -> broker` channel migration, so nothing went live. The printed rollback was also unexecutable (its own switch hits the same inhibitor). The cleanup can now be taken with no system change.

## Safety, measured with a real `--clean`
An old directory containing a **recent** file is **not** destroyed — tmpfiles uses `rmdir` semantics. `mM` is more thorough, not more dangerous; the glob scoping carries the safety.

## Two measurement traps recorded
- `--dry-run` prints \"Would remove\" on **stderr** — `2>/dev/null | grep -c` returns a clean 0 for a rule removing 1,066 entries.
- `--dry-run` **over-reports**, flagging directories the real run leaves alone. Its counts are upper bounds, so the header's \"4,255,261 entries / ~47 GB\" is not a reclaim estimate.

## Also included
`scripts/diagnose-disk-accounting.sh` — root-required (refuses non-root, rc 2), dedupes hardlinks by inode, excludes foreign mounts, and reports its own denial count and inode residual rather than hiding them.

## Verification
- Dev-host node tier on the **merged tree**: PASS, 1449/1449.
- Dev-host pytest tier on the merged tree: running at time of writing.
- Merge is textually clean **and** the two sides' file sets are disjoint (main touched none of these three files in the 44 commits since the merge-base), so there is no semantic-conflict surface.

https://claude.ai/code/session_01BRy2Xxw22fU9zdDe455dKA


ZacxDev:
## Audit round 1 — fixes landed

Round 1 (blind, full) reported 2 🔴, 12 🟡, 4 🟢. Six findings were fixed in
`cbdae4f5` and `463ad54d`. Recorded here as CLAIMS to be checked against the diff,
not as conclusions.

```audit-claims round=1 audited=d915f362bf11e815f9dc42ca44fbe876884b2df2..463ad54d
1. F1 root command injection in diagnose-disk-accounting.sh section 6d is closed by replacing `xargs -I{} sh -c '… "{}" …'` with `find -exec sh -c '… "$1"' _ {} \;`
2. F2 the E2BIG abort from `du -sh -x /tmp/*` is closed by `find -print0 | xargs -0 du -sh -x`, so the script no longer stops mid-section under set -euo pipefail
3. F3 the lsof byte total no longer reads a fixed column: SIZE/OFF is resolved from the header and a missing column prints COULD NOT MEASURE instead of 0
4. F4 section 6c's foreign-mount list is no longer assigned in a subshell, so foreign mounts are listed rather than reported as "none"
5. F5 rule eviction is scoped to the `systemd.tmpfiles.rules` block, so a rule in `systemd.user.tmpfiles.rules` is no longer deleted
6. F7 a trailing comment after a rule's closing quote no longer defeats eviction, and the printed count is a count of removals rather than matches
7. F8 an unrecognised argument exits 64 without touching /etc/nixos, and --help prints usage instead of performing the live edit
8. F9 the dead `homelab-talos-prs-*` glob is removed from the ledger and the coverage table no longer claims it, since it matched 821 plain files and 0 directories
9. F14 the comment inserted into /etc/nixos no longer says "mtime-ONLY ageing (`m:`)" above mM: rules
10. F10 the eviction block now has five tests; mutation-verified that deleting the block, unscoping the removal, reverting the regex, and re-adding the dead rule each fail
```

Not fixed, deliberately, and open: F6 (eviction cannot distinguish a deliberate human rule from a stale one — it now warns and names each evicted line, with the backup as undo), F15 (the `--emit-rules` recipe's `/tmp` handoff — mitigated by using `/run`), and the nit that `diagnose-disk-accounting.sh` has no test file and the repo has no shellcheck gate, which is why F1–F4 were invisible to the merge gate.

ZacxDev:
## Audit round 2 — fixes landed

Round 2 (delta, `d915f362..463ad54d`) returned **needs rework**: one 🔴 regression
introduced by round 1's own fixes, two 🟡, and six 🟢. Fixed in `990b9fe4` and
`be6744ac`. Recorded as CLAIMS to check against the diff, not as conclusions.

```audit-claims round=2 audited=463ad54dcd4c433ac2fb70c23d5d4636446b43b8..be6744ac
1. R2-1 the removal loop no longer splices by unanchored substring — it splices on the regex match's own start/end offsets, right-to-left, so it can only ever remove a line the pattern actually matched
2. R2-1 an unrelated live rule below a comment that quotes a ledger rule is no longer commented out, and the stale rule that was surviving is now removed
3. R2-2 the anti-duplication guard is keyed on the stable MARKER instead of HEADER.splitlines()[0], so rewording the header no longer makes it read absent and prepend a second block
4. R2-2 an already-applied host carrying an older header has it REPLACED rather than accumulated, so the superseded `mtime-ONLY ageing (m:)` sentence does not survive
5. R2-2 the old assertion `text.count("/tmp churn retention (2026-08-15)") == 1` is replaced by a structural count of header blocks, so a future reword cannot vacate it
6. R2-3 the eviction block's closing bracket is found by scanning for a line whose only content is `];`, so a `];` inside a comment no longer truncates the scope
7. R2-5 `bash script ""` no longer edits /etc/nixos — argc is checked before the case, and an empty single argument exits 64
8. R2-6 both mktemp files in diagnose-disk-accounting.sh section 6c are created before the trap widens, so a second-mktemp failure does not leak the first
9. R2-9 the --emit-rules recipe now uses `sudo sh -c '… > /run/churn.conf'`, so root never reads a file written into a 1777 directory by an unprivileged shell
10. Three regression tests added and mutation-verified: reverting the splice fails 1, reverting the block close fails 1, reverting the header guard fails 2; control green and byte-identical after
```

⚠ Correction to round 1's block: claim F15 there said "mitigated by using `/run`".
That was **false against the tree** — the shipped file still documented `/tmp` and
contained no `/run`. Round 2 caught it; `be6744ac` makes it true. Flagging because
a claim that was wrong once is the kind a later round takes on trust.

Still open and deliberately unfixed: **F6** — eviction cannot distinguish a
deliberate human rule from a stale one. It warns and names each evicted line, and
the backup is the undo. Round 2 noted that this is only defensible while the
eviction removes exactly what it names, which R2-1 broke and `990b9fe4` restores.
Also open: `diagnose-disk-accounting.sh` still has no test file, and the repo has
no shellcheck gate — the reason the round-1 🔴s were invisible to the merge gate.
Round 2 singled out the lsof awk as the one piece that is cheap to cover, being a
pure text transform; that is not done.

PR #1233, opened 2026-09-02
fix(subsystem-store): repoint the host-local READ path off the frozen mirror, and refuse an undateable store

ZacxDev:
## The defect

The Cairn cutover made a hosted pod the canonical subsystem store, **froze** the per-host
mirror at `~/.claude/analyze-service-index` (entry files `0444`, nothing refreshes it), and
introduced a synced read-through cache at `~/.cache/subsystem-store` that `cairn sync`
maintains. **Nobody repointed the READ path.**

Measured on the workbench, 2026-09-02:

| command | store it read | verdict it printed |
|---|---|---|
| `subsystem_recall.py --repo ~/workspace/devrc --list` (what `/resume` step 4 ran) | `~/.claude/analyze-service-index` | **ALL 26** entries in `devrc/`, none omitted |
| `cairn recall --scope devrc` | `~/.cache/subsystem-store` | **ALL 29** entries in `devrc/`, none omitted |

Two directories. The frozen one is three entries short, **claims completeness anyway**, and
carries no staleness stamp. Every `/resume` since the cutover oriented on a store that had
stopped moving while being told it was complete.

**The defect was wider than the reader.** `service_recon.py` (the `/analyze-service` recon)
had the same `DEFAULT_STORE_ROOT` default at `read_index` / `_read_index_one` / `recon`, so
*both* prescribed read surfaces were frozen.

## The fix

**One resolver, one place** — `scripts/lib/subsystem_read_store.py`, a stdlib-only leaf:

* Owns `DEFAULT_CACHE_ROOT` and `SYNC_STAMP`, which `scripts/cairn` used to declare and now
  imports. One definition each, so the writer and every reader move together.
* **The discriminator is the STAMP, not the path.** A store carrying `.sync-stamp` can say how
  fresh it is; one that cannot is REFUSED rather than silently served. That is strictly wider
  than "don't read that path": it also catches a host that has never synced, and it cannot be
  walked by moving the frozen tree somewhere else.
* **No clock.** It reads a file and returns its lines verbatim. `subsystem_recall` documents
  itself "READ-ONLY. No clock, no network, no git, no prompt" and `cairn.cache_age` owns age
  computation; a second, cheaper age here would be a second answer to one question.

**`subsystem_recall.py` — CLI only:**

* `--store` defaults to the synced cache.
* The **DEFAULT** resolution exits **4** with `REFUSING to read … Run \`cairn sync\` and
  re-run` when the store is unstamped, and renders **no index at all**.
* An **EXPLICIT** `--store <path>` stays permissive. That distinction is recorded by an
  argparse `Action`, not derived by comparison — `--store <the cache>` is byte-identical to
  passing nothing, and those two must not behave the same.
* A stamped read prints the stamp **unparsed** (trailing whitespace stripped, nothing else)
  beside `store:` / `store host:`, and under `read_store_stamp` in `--json` (the one output
  with no header block to read). `service_recon`'s brief prints the same lines under its
  `index:` line, through the same `_read_store.stamp_header`.

**The library is deliberately untouched.** `subsystem-store-api/server.py` and `scripts/cairn`
both `import subsystem_recall as rc` and call `rc.recall` / `rc.search` / `rc.load_store`
against a directory with no stamp — `/data`, or an unsynced cache. A refusal there would take
the store **and its own sync client** offline in one commit. Verified by grep before a line was
written (neither shells the CLI; neither calls `rc.main`) and pinned by
`TestThePodContractIsUnchanged`, including a source-region assertion that the library half of
the file never consults the resolver at all.

**Round 1 of the audit ladder** found no 🔴 and four 🟡, all real, all fixed here:

- **F1** — `Brief.store_root` reported the literal string `"None"`: `recon` stringified its own
  `store_root=None` sentinel, so the field that answers *which store did this brief read?*
  answered with the sentinel meaning "work it out". `recon` now resolves once and reports the
  **resolved** root; on a refusal the value is `NOT READ (store-unstamped) — <path>`, because a
  bare path there is a claim the brief opened it. The suite was structurally blind: all ~40
  `sr.recon(...)` calls pass `store_root=` explicitly and nothing asserted the field. Now tested
  on the **default** path, stamped and refused.
- **F2** — the pod-contract guard split the source at `def _build_parser(` and scanned only
  above it, missing **`_exit_for`**, which `server.py` calls on every `/recall` and `/search`
  and `cairn` on every read. A refusal grown there SURVIVED the guard. It is no longer
  positional: it enumerates by AST every top-level definition mentioning the resolver and
  asserts that set **equals** an explicit CLI-only allowlist. The behavioural backstop now
  repoints the default first — on this host the real cache is stamped, so the mutant would
  otherwise never execute and be scored SURVIVED for want of a reachable input.
- **F3** — two more expectation-derived-from-implementation survivors, pinned to literals built
  independently of the constants (details in the sweep section).
- **F4** — `/resume` step 4 said "never prompts and never writes" (`cairn recall` fetches and
  unpacks the cache) and told the reader a `4` means "run `cairn sync`". **`cairn recall` can
  never return 4** — it reaches the reader as a library — and `cairn`'s own 4 is
  `EXIT_REFRESH_FAILED`, sync-only, meaning *the store was not reached but a cache survived*.
  The instruction sent the reader to re-run the command that had just failed. Both sentences
  corrected, and the collision pinned from the code: a test asserts the two 4s are equal **and**
  that `EXIT_REFRESH_FAILED` is returned from `cmd_sync` and nowhere else.

**`service_recon.py` degrades gracefully** — the index section reports
`store-unstamped — the index could not be read: … Run \`cairn sync\`` while roots, config, git
log and live state still run. Asserted by comparing the degraded brief block-for-block against
a named-store brief: exactly one block may differ, and it must be the index.

**Prescriptions updated:** `/resume` step 4 and the `/handoff` doc template now say
`cairn recall --repo <path>`; `/analyze-service` prefixes `cairn sync`; the two live cairn
handoff docs' "Run this first" blocks follow. The `Dockerfile` and `Dockerfile.dockerignore`
gain the new module — the image enumerates its `COPY`s and the ignore file is an allowlist.

## Verification

**Live, on this host** — the original symptom reproduced and gone:

```
$ python3 scripts/lib/subsystem_recall.py --scope devrc --list
subsystem-recall: status=recalled scope=devrc
  store: /home/zach/.cache/subsystem-store          <- was ~/.claude/analyze-service-index
  host: nixos-…
  stamp: synced=1788363567
  stamp: revision=unknown
  stamp: snapshot=seeded=… entry-files=201
  stamp: entries=201
```

…and the refusal, against a never-synced cache: `EXIT=4`, refusal on stderr, no index.

**Red-at-base / green-at-HEAD. RE-MEASURED FROM SCRATCH IN ROUND 3, because this figure
had been wrong three times.** Base **`a6e50641`** (the merge-base). Against the *pure* base the
new test file cannot import (`ModuleNotFoundError: subsystem_read_store`), and a bare red is
not evidence when a **name** is what is missing — so the matrix is taken against base **plus
the new leaf module plus name-only shims** (`INDEX_UNSTAMPED`, `NOT_READ_PREFIX`,
`brief_store_root`, `Brief.store_root`, `Brief.store_stamp`, `EXIT_UNSTAMPED_READ_STORE`, and
an accepted-but-unused `extra_header=` keyword). No behaviour. Anything still red after that is
behavioural.

**PR-wide: 34 red at base / 31 pass at base and are labelled invariant guards**, not counted as
regression coverage. Measured, not derived:

| file | red at `a6e50641`+shims | green at base (invariant) | collected |
|---|---|---|---|
| `test_subsystem_read_store.py` | 31 | 29 | 60 |
| `test_subsystem_recall.py` (the 5 cases this PR adds) | 3 | 2 | 5 |

⚠ **The round-3 audit proposed 25/30 and that correction does NOT hold at this base.** It found
`test_the_status_literal_matches_the_skill_prose_that_quotes_it` passing "at base" — true of the
**`c05b2df8`** payload it measured, which is a *mid-PR commit* that already contains this PR's
own SKILL.md change. At `a6e50641` that file carries no `store-unstamped` anywhere, so the test
is **red**. Two true observations, two different bases, and the arithmetic only works on one of
them. The re-measurement reproduces the previous per-file numbers exactly (23/27 + 3/2 = 26/29
before round 3's 8 new tests and 2 new parametrize cases), so the earlier figure was right for
the base it named.

⚠ **The audit's underlying point stands, and the category is bigger than one test.** Four of
the 34 are red at base only because they pin **prose or a command shape this PR introduces** —
they guard future drift rather than catch a pre-existing defect:
`test_the_REMEDY_names_a_subcommand_cairn_ACTUALLY_HAS`,
`test_the_status_literal_matches_the_skill_prose_that_quotes_it`,
`test_resume_step_sentence[the prescribed command …]`, and
`test_the_analyze_service_step_does_NOT_chain_the_sync_with_double_ampersand` (which caught a
defect this PR *itself introduced* in commit 1 and fixed in round 2 — a real catch, but not a
pre-existing one). **Thirty catch a defect that existed before this PR.**

**Round 3's own 8 new tests, measured against the round-2 head `1b5d2e43` + name-only shims —
7 red, 1 green:**

| test | at `1b5d2e43` |
|---|---|
| `test_a_stamped_default_prints_EVERY_stamp_line_INSIDE_the_index_block` | RED |
| `test_the_stamp_precedes_the_ENTRY_BODY_on_a_hit` | RED |
| `test_the_stamp_prefix_is_the_ONE_spelling_both_renderers_use` | RED |
| `test_the_brief_json_carries_the_stamp_UNPARSED` | RED |
| `test_an_explicitly_named_store_is_dated_when_it_can_be_and_not_when_it_cannot` | RED |
| `test_the_skill_prose_says_what_the_recon_ACTUALLY_does` | RED |
| `test_the_prose_token_is_the_one_the_code_emits` | RED |
| `test_a_REFUSED_default_invents_no_stamp` | **GREEN — invariant guard** (nothing rendered a stamp then either) |

⚠ **Two of round 3's fixes are NOT red-at-base and must not be counted as regression coverage.**
Widening the ledger sweep (F1) and the `cairn` re-declaration arm (F5) makes two guards that
were *green while covering nothing* actually cover it. A widened guard is green before and
after by construction; the only instrument that can show the difference is the mutation sweep,
and M1–M5 / M15–M16 below are that evidence.

**Mutation sweep — 27 mutants + 1 positive control + 1 equivalence control. 27 KILLED, 0
unexplained survivors**, each by the *expected* test rather than a neighbour's error; control
green before and after (635 both times); run under `PYTHONDONTWRITEBYTECODE=1`. Beyond the
refusal guard it covers all four round-1 fixes, all three round-2 fixes, and each of the three
refusal sites separately so they cannot cover for one another.

The one mutant that survives is **declared equivalent** and the declaration is itself a
control: reading `_read_store.DEFAULT_CACHE_ROOT` inside `build_parser` is a call-time
attribute access, so it follows a repointed global exactly as `read_store_root()` does. The
hazard is the *module-scope* `from … import` binding, and that mutant is KILLED. If the
equivalence one is ever killed, the sweep reports the reasoning as wrong rather than as a pass.

🔴 **Two sweep results were findings rather than harness bugs, and both are reported because
the sweep is only worth running if its survivors are believed.**

1. The first positive control SURVIVED: every fixture wrote the stamp through `rs.SYNC_STAMP`,
   so `SYNC_STAMP = ".sync-stampX"` renamed the fixture with it and 615 tests stayed green.
   Expectation derived from the implementation. Now pinned to the literal, with a behavioural
   half that writes `.sync-stamp` by hand. Round 1 found two more of the same shape
   (`DEFAULT_CACHE_ROOT`'s `$HOME` anchor, `REMEDY`); both pinned the same way.
2. **Re-running the sweep after the round-1 fixes turned a previously-KILLED mutant into a
   SURVIVOR** — which is the entire reason to re-run it. Fixing F1 moved the store resolution
   up into `recon`, so `read_index`'s own refusal branch stopped being reachable through any
   normal path and went silently uncovered. It is live surface (`read_index` is exported), so
   it is now pinned by three direct-caller tests plus its own mutant.

## Round 2 — the same defect class, for the fourth time

🔴 **The F1 fix re-opened the class the F3 fix had just closed, in the same session.**
`NOT_READ_PREFIX` was asserted only as `blob["store_root"].startswith(srec.NOT_READ_PREFIX)` —
the constant agreeing with itself. Measured: `NOT_READ_PREFIX = "READ"` survived all 42 tests,
which would have made an unstamped default emit
`"store_root": "READ (store-unstamped) — …"` — the field asserting the brief *read* the store
it explicitly refused to open, the exact inversion the field exists to prevent, on a green
suite. `INDEX_UNSTAMPED` was the same shape (pre-existing).

Four one-off pins is what allowed the fourth, so the remedy is a **ledger**: `WIRE_CONSTANTS`
maps every boundary-crossing string — a filename on disk, a command a human types, a status
another tool matches, a prefix a JSON consumer parses — to its literal, and
`test_the_ledger_is_two_way_over_the_new_module` sweeps every module-level `str` constant in
`subsystem_read_store` and fails on one nobody pinned. That two-way half is the only part that
can stop the fifth.

**`analyze-service/SKILL.md` chained `cairn sync && …`, and `&&` deletes the whole brief.**
`cmd_sync` returns 4 whenever the pod is unreachable but a usable cache survives — its stated
contract, not an edge case — so during an outage the recon never ran and `/analyze-service`
emitted nothing, while the paragraph eight lines below promised "a failed sync costs you the
index block, never the brief". Now `;`, pinned on the command itself (the `REMEDY` test only
required the substring `cairn sync`, which `&&` satisfies).

## Round 3 — the ledger's own blind spot, and the `;` that handed back an undated index

Round 3 audited the round-2 delta and returned five 🟡. All five are fixed here.

🔴 **F1 — the fifth instance of this PR's recurring class was hiding INSIDE the guard built to
stop the fifth.** `test_the_ledger_is_two_way_over_the_new_module` promised the new module's
"whole surface is in scope" and implemented `name.isupper() and isinstance(value, str)`.
MEASURED as SURVIVORS against that sweep: `FROZEN_MIRROR = Path.home() / ".claude" /
"analyze-service-index"` (a `Path`), `STAMP_FIELDS = (...)` (a tuple), and
`Wire_Fact_v2 = "another-status"` (mixed case). Not hypothetical either —
**`DEFAULT_CACHE_ROOT`, the most load-bearing wire fact in the module, is a `Path`** and was
therefore invisible to the sweep, surviving only on a hand-written literal pin in one test:
exactly the one-off pattern the ledger exists to replace. The sweep now enumerates module-scope
**assignments off the AST**, so neither the name's case nor the value's type is an escape
hatch, and `DEFAULT_CACHE_ROOT` and the new `STAMP_PREFIX` are ledger rows.

🔴 **F2 — `;` was the right fix and it exposed a gap the new prose flatly denied.** The refusal
fires on an **UNDATEABLE** store, never on a **STALE** one. Pod down three days, cache stamped
three days ago: `cairn sync` prints `⚠ cairn: cached … 3 days old` **on stderr** and exits 4,
`;` correctly lets the recon run, and the reader gets a complete brief whose `index:` block is
three days old **with no freshness field anywhere in the text output** and nothing telling them
to carry the sync banner forward. Under `&&` you got nothing, loudly; under `;` an unlabelled
stale index, silently — which is this PR's own thesis one layer up.

Fixed by rendering the date rather than by weakening `;`:

* `render_brief` prints the stamp as `stamp:` lines **directly under the `index:` line**, on
  every branch (inserted at a remembered index, so a new branch cannot forget it).
* `brief_json` carries it under **`read_store_stamp`** — the same key `subsystem_recall --json`
  already uses, so a consumer needs one spelling, not two.
* Both renderers go through `_read_store.stamp_header`, and `STAMP_PREFIX` is a ledger row:
  the skill tells a reader to relay "the `stamp:` lines", which is true of one output and false
  of the other the moment the two spellings drift.
* No clock, still: the epoch is echoed, never converted. `cairn` owns age arithmetic.
* The two false SKILL.md claims ("REFUSES an undateable store rather than serving a stale one",
  "would have served the index block at full fidelity") are corrected, and
  `reference/index-store.md`'s `index:` status table — which never gained the `store-unstamped`
  row this PR added, and still said every status comes from `recall` — is corrected too.
* 🔴 The replacement paragraph is pinned as a **WHOLE NORMALISED STRING**, not by keyword. This
  file's prose has now carried a false freshness claim twice, and each time the keyword pins
  stayed true. A reword now goes red on purpose; that is the price of a machine-readable claim.

**F3** — the red-at-base matrix, re-measured; see the ⚠ notes above.

**F4** — the earlier reword retired the word VERBATIM in the module and left two claims
standing (`test_the_header_carries_every_stamp_line_UNPARSED`'s docstring and an inline
`# verbatim, unparsed` comment asserting both halves of a contradiction). The two places that
quote the historical error *deliberately* are untouched.

**F5** — one arm of `test_the_constants_have_exactly_one_definition` matched the text
`DEFAULT_CACHE = Path.home()`, a spelling the previous round deleted from `scripts/cairn`, so
the arm could never match again. MEASURED: re-declaring `DEFAULT_CACHE_ROOT = Path.home() /
".cache" / "subsystem-store"` at `cairn`'s module scope **SURVIVED** it. The arm is now an AST
walk (immune to whitespace and to prose containing the same characters) whose forbidden set is
**derived from `subsystem_read_store.__all__`**, so a constant added there is covered the same
day with no second list to keep in step.

**Round-3 mutation sweep — 16 mutants + 1 positive control, 16 KILLED, 0 survivors, each by the
EXPECTED test.** `cp -a` copy with its `.git` removed FIRST, `PYTHONDONTWRITEBYTECODE=1`,
control green before and after (645 both).

| mutant | killed by |
|---|---|
| `FROZEN_MIRROR` (a `Path`) added | `test_the_ledger_is_two_way_over_the_new_module` |
| `STAMP_FIELDS` (a tuple) added | same |
| `Wire_Fact_v2` (mixed case) added | same |
| `KNOWN_SCOPES` (a frozenset) added | same |
| `DEFAULT_CACHE_ROOT` anchored at `/var/tmp` | `test_the_constant_equals_its_literal[DEFAULT_CACHE_ROOT]` |
| `REMEDY` renamed (positive control) | `test_the_constant_equals_its_literal[REMEDY]` |
| `render_brief`'s stamp insert deleted | `test_a_stamped_default_prints_EVERY_stamp_line_INSIDE_the_index_block` |
| stamp appended to the brief instead of under `index:` | `test_the_stamp_precedes_the_ENTRY_BODY_on_a_hit` |
| `_resolve_store` returns an empty stamp | `test_a_stamped_default_prints_EVERY_stamp_line_INSIDE_the_index_block` |
| `STAMP_PREFIX = "  stamped: "` | `test_the_constant_equals_its_literal[STAMP_PREFIX]` |
| `stamp_header` returns nothing | `test_the_header_carries_every_stamp_line_UNPARSED` |
| `read_store_stamp` JSON key renamed | `test_the_brief_json_carries_the_stamp_UNPARSED` |
| one word changed in the freshness paragraph | `test_the_skill_prose_says_what_the_recon_ACTUALLY_does` |
| the retired false claim re-added to the doc | same |
| `cairn` re-declares `DEFAULT_CACHE_ROOT` | `test_the_constants_have_exactly_one_definition` |
| `cairn` re-declares `SYNC_STAMP` | same |

**Tests run on the shipped tree** (dev-host pytest tier, targeted): the analyze-service /
subsystem / cairn / skill-doc suites, **green**. The `nix build .#checks…` sandbox tier is the
dispatching session's to run on the merged tree.


ZacxDev:
## Audit ladder — round 1 claims

Round 1 was a first, full blind audit. It returned **no 🔴 and four 🟡**. All four are claimed fixed below, plus two extras and one finding the fix round introduced and caught itself.

```audit-claims round=1 audited=b49938aa..c05b2df8
1. F1 — `service_recon.py` reported `"store_root": "None"` in `brief_json()` on the default path; `recon` now resolves the store once and reports the resolved root, and a refusal reads `NOT READ (store-unstamped) — <path>` via a single `brief_store_root()` spelling. Two new tests exercise the DEFAULT path and assert that field.
2. F2 — the pod-contract guard split the source positionally at `def _build_parser(` and so excluded `_exit_for` (line 3186), which `server.py:3573` and `cairn:730` call on every read. The guard is no longer positional: it walks the AST and asserts the set of top-level defs mentioning `_read_store.` EQUALS `{_build_parser, _with_stamp, main}`, pinned both ways; its premise is pinned separately; the behavioural backstop now repoints `DEFAULT_CACHE_ROOT` before calling `_exit_for`, so it no longer depends on this host's cache being stamped.
3. F3 — two constants had expectations derived from the implementation. `DEFAULT_CACHE_ROOT`'s expectation is now built from `Path.home()` plus two literals and never read back off the constant; `REMEDY` is pinned to the literal `"cairn sync"` and cross-checked against `scripts/cairn`'s own `add_parser` declarations and both prescribing SKILL.mds.
4. F4 — `resume/SKILL.md:138` claimed the step "never prompts and never writes" about a command that networks and unpacks a tar, and documented an exit 4 that the prescribed command can never return while `cairn`'s own 4 means something else. Both sentences rewritten; the two-different-4s hazard is now machine-checked by `test_cairns_exit_4_is_SYNC_ONLY_and_is_not_the_readers_refusal`.
5. EXTRA — the Dockerfile's "only four .py files" count is removed rather than corrected, now pointing at the transitive-closure test that is the authority; the stdlib enumeration names `subsystem_read_store`.
6. EXTRA — `read_stamp`'s docstring said VERBATIM while the body `rstrip()`s; the strip is kept (it keeps a CRLF `\r` out of the header) and every affected docstring now states what happens, backed by a test feeding CRLF, blank lines, leading whitespace, a duplicate key and an unknown field.
7. SELF-CAUGHT — fixing F1 moved resolution into `recon`, which made `read_index`'s own refusal branch unreachable by any normal path: mutant M13 went KILLED -> SURVIVED between commits. Three direct-caller tests plus a permissive control were added, and the sweep gained M13/M13b/M13c so the three refusal sites cannot cover for one another.
8. Corrected matrix, twice: the claim is now **20 behavioural red-at-base / 22 invariant guards**, derived by applying name-only shims at base rather than by reasoning. One of the 20 (`test_the_REMEDY_names_a_subcommand_cairn_ACTUALLY_HAS`) pins a new contract rather than catching a pre-existing defect, so **19** catch a real defect.
```

### Ledger — round 1

`round 1 · payload lines changed THIS round: 101 added / 22 removed across 5 payload files (service_recon.py 61/5, subsystem_read_store.py 18/8, subsystem_recall.py 2/2, resume/SKILL.md 5/1, Dockerfile 15/6) · scaffolding: test_subsystem_read_store.py 277/14`

Measured with `git log --numstat --format= --remerge-diff b49938aa..c05b2df8 --not a6e50641`, rc 0, non-empty range, git 2.55.0. **The attribution gate does not fire** — this round changed substantial payload, so the ladder is still auditing the PR rather than its own guards.

### 🔴 The base moved during this round

`origin/main` went `a6e50641` → `3d0695c7` (4 commits). **No file overlaps this PR** — and per `claude/RULES.md` that is explicitly not safety. The relevant one is **#1219 / `b4fde334`, "one shared store siting"**, which added `scripts/testlib/store_siting.py` and modified `test_cairn_cli.py`, `test_cairn_write.py`, `test_subsystem_store_api.py`. This PR adds a *new* store-siting constant (`DEFAULT_CACHE_ROOT`) and its tests. Two sides both deciding where a test store lives, with zero shared files, is the disjoint-file merge hazard verbatim.

Both gate tiers passing at `c05b2df8` are therefore claims about the **branch**, not about the merged tree. A merged-tree gate off `3d0695c7` is being run separately and its result will be posted here.

ZacxDev:
## Audit ladder — round 2 claims

Round 2 was a delta re-audit over `b49938aa..c05b2df8`, blind to the implementer's reasoning. It confirmed all four round-1 findings genuinely fixed and mutation-verified, and returned **no 🔴, two 🟡, three 🟢**. All are claimed fixed below.

```audit-claims round=2 audited=c05b2df8..1b5d2e43
1. N1 — `NOT_READ_PREFIX` was asserted only as `startswith(srec.NOT_READ_PREFIX)`, the constant agreeing with itself; `NOT_READ_PREFIX = "READ"` survived all 42 tests. Fixed as a LEDGER rather than a fifth one-off pin: `WIRE_CONSTANTS` maps each boundary-crossing string to its literal written out in the test file, and `test_the_ledger_is_two_way_over_the_new_module` enumerates every module-level `str` constant in `subsystem_read_store` and fails on one nobody pinned. `service_recon`'s two are pinned by name, deliberately not swept.
2. N2 — `analyze-service/SKILL.md:17` chained `cairn sync && …`, so a pod outage with a surviving cache (exit 4) short-circuited and produced no brief at all, contradicting the file's own "a failed sync costs you the index block, never the brief". Now `;`, pinned on the command line itself rather than on the `cairn sync` substring, plus an assertion that the "never the brief" promise is still present.
3. N3 — `INDEX_UNSTAMPED` had the same unpinned shape (`"store-unstampedX"` survived all 42). Pinned via the ledger plus a code/prose cross-check requiring `index: store-unstamped` in `analyze-service/SKILL.md`.
4. N4 — the red-at-base matrix was scoped to one file. Re-measured at the new head: **26 behavioural / 29 invariant guards** (8 new tests: 3 red, 5 green). The round-2 auditor's 23/24 reproduced exactly at `c05b2df8`. This figure has now been wrong twice, both times conservative.
5. N5 — the Dockerfile asserted "there are seven", the same artefact that had just gone stale as "four". The number is removed rather than corrected, pointing at the transitive-closure test that is the authority.
6. NIT — `scripts/cairn:90` did `from subsystem_read_store import DEFAULT_CACHE_ROOT as DEFAULT_CACHE`, the binding pattern `read_store_root()`'s docstring names as defeating test repointing. It now calls the accessor, pinned behaviourally through cairn's own parser. This change would have silently neutered `test_cairn_who.py:1028`, which patched `cli.DEFAULT_CACHE` with `raising=False`; that patch now targets the seam at the default `raising=True`.
7. NIT — `resume/SKILL.md`'s "verbatim" and three `_verbatim` test names are now `UNPARSED`, matching the documented `rstrip()`. The "exits non-zero only when NOTHING readable came back" rule is now attributed to the reader, with cairn's wrapping code 5 called out as not following it.
```

### Ledger — round 2

`round 2 · payload lines changed THIS round: 28 added / 8 removed across 4 payload files (analyze-service/SKILL.md 10/1, cairn 11/2, resume/SKILL.md 4/3, Dockerfile 3/2) · scaffolding: 155/5 · since round 1: 129/30 payload`

Measured with `git log --numstat --format= --remerge-diff c05b2df8..1b5d2e43 --not a6e50641`, rc 0, non-empty range, git 2.55.0. **The attribution gate does not fire** — payload is non-zero. But it flattened 101 → 28, which is the trend to watch: if round 3 returns findings that are mostly about guards this ladder wrote, that is the signal to stop and file the remainder rather than keep going.

### Merged-tree gate — green at the *previous* head, and already stale

Run against merge commit `edcc0fdf` = `main@3d0695c7` + PR `c05b2df8` (clean merge, no textual conflict):

| tier | verdict | counts |
|---|---|---|
| dev-host `gate.sh --tier both` | `GATE: RESULT=PASS exit=0` | both runners' own `RESULT:` lines |
| sandbox `pytests` | `RESULT: PASS (exit=0)` | collected=20537 passed=20534 skipped=3 **failed=0**, floor 18404 |
| sandbox `nodetests` | `RESULT: PASS (exit=0)` | 1449/1449, floor 1367 |

Zero timeout panics; the two sandbox derivations were built one at a time. This clears the disjoint-file hazard against **#1219 / `b4fde334`** ("one shared store siting") *by measurement* rather than by the absence of shared files.

🔴 **It is nonetheless stale twice over**: the head has moved to `1b5d2e43`, and `origin/main` has moved `3d0695c7` → `946a51f0`. The merged-tree gate will be re-run against the final head and a re-derived base before merge.

### An unrelated flake this ladder measured, filed rather than fixed here

The pytest tier went red once at `1b5d2e43` on `test_subsystem_touch.py::TestCommitNegativeControls::test_the_LENGTH_bound_is_not_vacuous_git_WOULD_have_expanded_it` — **with a mechanism, not an "it's flaky"**: the assertion is `expanded == [sha]` and the run returned two objects, `5875580177…` and `587be6b4c3…`, both starting `587`. `--disambiguate` matches any object type, and commit shas re-randomise per run because they are timestamp-dependent. Rate is order 0.2 %/run (objects/4096) — **estimated, not measured**.

Attribution: this PR touches neither `subsystem_touch.py` nor its test file (`git diff --stat` empty), and that test file imports none of this PR's modules. Clean re-run: `RESULT: PASS (exit=0)`, `11531 passed`, up exactly one from `11530 passed, 1 failed`. Deliberately **not** fixed here — different subsystem, outside the audit delta, and widening the PR mid-ladder hands the next round unrelated surface. The one-line fix is `assert sha in expanded`.

ZacxDev:
## Audit ladder — rounds 3 and 4, and why the ladder STOPS HERE

Four audit rounds ran. **No round ever returned 🔴.** Rounds 3 and 4 both returned "safe to merge" *and* real findings, all of which are now fixed.

```audit-claims round=4 audited=eb947328..cf1ff4da
1. N1 — the WIRE_CONSTANTS ledger sweep iterated `tree.body` only and required `ast.Name` targets, while its docstring claimed the module's whole surface. `_module_scope_assignments` now follows module scope through if/try/with/for/while (bodies, orelse, finalbody, handlers) and collects every durable binding form: `=`, annotated `=`, tuple/list/starred unpack, `for … in`, `with … as`, `:=`. `except … as` is deliberately NOT collected — Python unbinds it, so collecting it would false-RED an ordinary `except OSError as exc:`.
2. N2 — the cairn forbidden set derived from `__all__` alone, which nothing keeps in step and which the ledger sweep excludes because it starts with `_`. It is now `set(rs.__all__) | _module_scope_bindings(subsystem_read_store.py)`, where `_module_scope_bindings` adds top-level def/class names — kept separate from the assignment sweep so the ledger does not demand a WIRE_CONSTANTS row per function.
3. N3 — the cairn clash check used `ast.walk`, so a function-LOCAL named like an exported constant went red with a re-declaration message. Moved to module scope; `_assigned_anywhere` had no other caller and is deleted.
4. N4 — `reference/index-store.md`'s provenance sentence was still false for `not-attempted` (built in `service_recon` before `recall` is asked, the same justification given one row down for excepting `store-unstamped`). Replaced with three groups that are true; `store-unreadable` row added; the reintroduced "verbatim" replaced with the real normalisation, which also drops blank lines.
```

### 🔴 Why the ladder stops here — recorded, because this is NOT a clean round

`claude/RULES.md` and the `/audit-pr` skill both say a clean round ends the ladder. **Round 4 was not clean** — it returned four findings, and they are fixed above. The ladder stops on the **attribution gate** instead, and that has to be written down or a reader cannot tell this closure from convergence.

| round | payload changed by its fixes | scaffolding | ratio |
|---|---|---|---|
| 1 | 101 lines | 277 | 1 : 2.7 |
| 2 | 28 lines | 155 | 1 : 5.5 |
| 3 (round 4's measurement) | **~22 executable** (58 of the 80 added lines are comments/docstrings) | 348 | **1 : 16** |

- **Round 3: 4 of 5 findings were in guards this ladder itself wrote.**
- **Round 4: 4 of 4.** Both substantive ones (N1, N2) were holes in *the test that guards the test ledger*.
- Round 4's fixes touched **no runtime file at all** — verified: `git diff --name-only eb947328 cf1ff4da` is exactly `test_subsystem_read_store.py` + `reference/index-store.md`.

The shipped payload, meanwhile, survived 36 mutants clean, is byte-identical on the library path, has no `index:` branch that renders without a stamp, and added no clock. The rounds have left the PR and are auditing their own scaffolding, which is the documented signal to stop.

**Nothing is being left unfixed.** All four of round 4's findings are closed in `cf1ff4da`. No round 5 will run; its expected yield is another guard-on-a-guard.

### The recurring defect class — five occurrences, and why the fifth fix is wider than its finding

This PR's characteristic defect was *an assertion satisfied by a constant agreeing with itself*: `SYNC_STAMP`, then `REMEDY` + `DEFAULT_CACHE_ROOT`, then `NOT_READ_PREFIX` (introduced by the fix for the first three), then N1/N2 (introduced by the fix for the fourth). Each fix was narrower than the class it closed.

The round-4 fix deliberately went **past** its prescribed hunk twice, each time measured first:
- The prescribed widening left `for [LOOP_STATUS] in [["store-frozen"]]:` **surviving** — N1's own defect one level in.
- The prescribed `__all__` union killed the `SECOND_STAMP` probe but not the `stamp_header` one, because `stamp_header` is a `def`, not an assignment.

Shipping a fix narrower than its own finding is exactly what produced occurrences 4 and 5, so both were widened. Measured before widening: cairn's 21 top-level defs and `subsystem_read_store`'s 6 have **zero** overlap, so no false-positive surface is added.

### Mutation matrix, base `eb947328` → HEAD `cf1ff4da`

11 hazard mutants **SURVIVED at base, KILLED at HEAD** (`try:`, `if True:`, tuple unpack, `for … in`, `with:` body, `with … as`, walrus, walrus-in-comprehension, `SECOND_STAMP` omitted from `__all__`, `stamp_header` dropped from `__all__`, `stamp_header` reassigned in cairn). Positive control (`SYNC_STAMP` rename) killed at both. Three negative controls correctly survive, including `except OSError as exc`. **The N3 case inverts as intended:** a cairn function-local named `stamp_header` was KILLED at base (the false red) and SURVIVES at HEAD.

Each mutant died on the expected guard carrying **its own** message, not the `hasattr` arm, with no collateral failure. `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` cleared between mutants, `cp -a` copy with `.git` removed first, control green before and after.

No new test functions were added, so there is no red-at-base matrix for this round — two existing guards were widened, and the matrix above is the equivalent evidence.

### Merged-tree gate

`main@5a82aaa9` + `eb947328` → merge commit `7ac5b465`, clean merge:

| tier | result |
|---|---|
| dev-host `gate.sh --tier both` | `collected=20555 passed=20551 skipped=3 **failed=1**` · node 1449/1449 |
| sandbox `pytests` | `collected=20555 passed=20551 skipped=3 **failed=1**` — the same single test |
| sandbox `nodetests` (at the previous base) | 1449/1449, floor 1367 |

🔴 **That one failure is `main`'s, not this PR's**, and it is bisected with controls at both ends: `scripts/collector/keylog/tests/test_espanso_detect.py::test_live_existing_resolutions_not_made_ambiguous` **passes at `146770ef`** and **fails at `a720d30d`** — a direct-to-main one-line commit that changed `:acq`'s label so `recom`/`recommend` match both `:acq` and `:rna` and resolve to nothing. It reproduces on both tiers and on plain `main` with no PR merged. Being fixed separately, on the operator's ruling ("the gate is overly strict, relax it") via the `_AMBIGUOUS_TERM_OWNER` declaration the detector already provides.

**#1233 adds zero failures to the merged tree.** This also clears, by measurement rather than by absence-of-shared-files, the disjoint-file hazard against #1219 / `b4fde334` ("one shared store siting") — which landed mid-ladder and touches the same subsystem with no file in common.

⚠ The gate above will be re-run against the final head and a re-derived base before merge; `origin/main` moved three times during this ladder (`a6e50641` → `3d0695c7` → `946a51f0` → `5a82aaa9`).

### A correction on the record

Round 2's auditor reported the red-at-base matrix as wrong and I relayed that as fact. **Both of us were wrong.** It measured "red at base" against `c05b2df8` — a *mid-PR* commit that already contained this PR's own SKILL.md change. Verified independently: `analyze-service/SKILL.md` contains `store-unstamped` **0 times at the PR's true base `a6e50641`**, once at `c05b2df8`, once at head (positive control). The implementer re-measured rather than complying, and was right to. The auditor's underlying point survives and is the better one: some tests are red only because they pin prose this PR itself introduces, which is not the same as catching a pre-existing defect — that distinction, and the base name, are now in the PR body.

PR #1239, opened 2026-09-02
fix(store-siting): the entry census was a syntactic sweep and six shapes walked through it — measure the real store instead (round-7 audit)

ZacxDev:
Round 7 of the store-siting audit ladder. Round 6's delta audit returned **7 findings (4 🔴, 3 🟡)**, six of them in the ratchet rather than the payload. Base: `3d0695c7`; `origin/main` merged in at `946a51f0`.

The findings interact — F5/F6 **widen** the site predicate, F2 **raises** the derived entry requirement, F1 says the budget must stop being a tripwire — so this is one coherent change, not seven patches.

> **Round 2 (2026-09-02)** — the round-1 audit of this PR returned **8 findings**, seven of them claims in this PR's own prose that were FALSE about the code beside them. Commit `5a3d7fe7` corrects them and this body is corrected in place. **No mechanism changed in round 2** — the operator's decision is to land an accurate residual list and end the ladder, not to keep widening a test-only ratchet. The three claims below that were falsified are struck and rewritten: the 13-site enumeration (it summed to 15), the "448 things" positive control (a one-off hand measurement, not a standing control), and one row of the mutation table (it does not reproduce, and cannot). See "Round 2 corrections" at the end.

---

## The central move: the census stopped reading the source

Rounds 4–6 kept `_LARGEST_STORE_BYTES` honest by DERIVING it from a syntactic sweep of the ledgered test files. The reasoning was right — a constant checked against a constant is not a guard — but a syntactic sweep can only ever see the shapes its author imagined, and every revision was walked through by one it had not.

So the enforcement left the scan and moved into the thing itself. **`store_siting._check_store_budget` walks the real directory at every `store_root` teardown** and raises `StoreBudgetExceeded` over budget. There is no spelling of a write loop that produces files a directory walk cannot see, so F2/F3/F4/F9 are **gone rather than patched**.

A ONE-OFF, HAND-INSTRUMENTED run over the three ledgered files saw the check called on **448 stores**, peak **1,253,376 B / 306 entries** (`test_cairn_cli.py`'s concurrency fixture), runner-up **176,128 B / 43 entries** — a 7x gap, so the peak is not a close call.

⚠ **That was a manual measurement and it is NOT the positive control this body originally called it.** The instrumentation it used (three `PEAK_STORE_*` module globals) was read by nothing and has been deleted; nothing re-runs the 448 count. The standing controls on this instrument are two tests that run every time and were mutation-verified: `test_store_root_INVOKES_the_budget_check_on_the_root_it_yielded` (it is reached at all) and `test_a_store_over_the_budget_RAISES_with_the_budget_checks_own_message` (it can go red, on its own wording).

---

## Per-finding disposition (round 6's audit)

| # | Finding | Disposition |
|---|---|---|
| 🔴 F1 | zero headroom on a census of the wrong population | **FIXED** |
| 🔴 F2 | nested loops summed, not multiplied | **FIXED** (census removed) |
| 🔴 F3 | three silent under-report shapes + missing `AsyncFor` | **FIXED** (census removed) |
| 🔴 F4 | the `entries >= 100` positive control is defeated | **FIXED** (control removed with the thing it guarded; replaced per-instrument) |
| 🟡 F5 | `_ROOT_BINDINGS` is SPELLED, not structural | **FIXED — and it was larger than reported** |
| 🟡 F6 | keyword arguments never indexed | **FIXED** |
| 🟡 F7 | the margin comment was falsified by round 6 | **FIXED, and the claim is now machine-read** |
| 🟢 F8 | two stale descriptions of the removed scan | **FIXED** |
| 🟢 F9 | `_seeded_entry_count`'s docstring lies | **FIXED** (function deleted) |

### F1 — zero headroom, wrong population
`_LARGEST_STORE_BYTES` was `1,875,968 = (442 + 16) * 4096`, where 442 was the sweep's own output: **the requirement and the constant were the same number, slack 0 bytes.** And 442 was not a store-entry count — `_seeded_entry_count` counted every `write_text`/`write_bytes` file-wide, scoring `test_cairn_cli.py`'s 3-file fixture as 15.

The budget is now `ceil(measured_peak * 1.5)` to a page — **computed, not transcribed**, so `_BUDGET_HEADROOM` cannot read 1.5 beside a budget carrying 1.26x. The 1.5 is pinned at `>= 1.25`, and the derivation itself is asserted.

⚠ **It also stopped modelling concurrency, and round 2 is where that got written down.** `main`'s constant was a SUM over every ledgered file's fixtures, documented as conservative under `-n 4 --dist loadfile`. This one is a single measured peak enforced **per store**, so adding a second large fixture no longer moves it, and `_MIN_FREE_BYTES` is a margin over ONE store while up to `PYTEST_JOBS` workers may hold one each. Nothing regresses in value (1,880,064 > the old sum 1,875,968) and `tmpfs_dir()` re-reads `statvfs` at every `store_root` entry, so the floor is a live check rather than a static reservation. Both constants now say which population they describe.

### F5 — the spelled predicate was bigger than the audit found
The audit reported the binding-**name** half and reproduced it with a rename. The directory-**name** half was the same defect and nobody had looked: `_ROOT_NAMES = {"store", "src"}` meant

```python
served = tmp_path / "served"          # a `cp -a` of the store fixture
subprocess.run(["cp", "-a", str(store), str(served)], check=True, ...)
path = served / SCOPE / "thing-alpha.md"
path.write_text(...)                  # written into
with running(served) as (base, _):    # and served by the real store server
```

did not count. **Measured: 13 such sites on `test_subsystem_store_api.py` while the ratchet read a confident 20.** The breakdown, **re-measured in round 2 because the original one summed to 15**:

| directory | sites |
|---|---|
| `served` | **3** (this body originally said five) |
| `served-elsewhere` | 3 |
| `stage` (the output of `run_seed`, served back) | 2 |
| `absent` (`running(tmp_path / "absent")`) | 2 |
| `big` | 1 |
| `at-the-cap` | 1 |
| `unambiguous` | 1 |
| **total** | **13** — plus the 20 `store`/`name`/`kind` sites = **33** |

`_DISK_ROOTED_SITES` therefore goes **20 → 33 with not one site added**. The assertion message and the ledger comment now both say the UP direction is ambiguous — "13 sites were written" and "the predicate got 13 sites wider" demand opposite actions — the way the DOWN direction already did.

The other side of that widening is pinned too: a cache dir, a scratch file and a non-consumer call still count **0**, so this is not "count every path", which would re-create the false accusation round 3 measured.

Of the 33, **26 flow into a store consumer and 7 count only via `_ROOT_NAMES`**. ⚠ Round 2 re-measured what those 7 are and the original description of them was wrong — see residual 3.

---

## Red → green matrix

All runs `PYTHONDONTWRITEBYTECODE=1`, `-p no:cacheprovider`, each mutation in a **freshly created isolated tree** (never mutated in place).

### The site predicate (F5, F6) — probes replayed against each ref's own index

| control | `origin/main` | this PR |
|---|---|---|
| rename `root` → `base_dir`, nothing else | 1 vs **0** ❌ | 1 vs 1 ✅ |
| same expression positional vs `store_root=` | 1 vs **0** ❌ | 1 vs 1 ✅ |
| directory named `store` vs `served` | 1 vs **0** ❌ | 1 vs 1 ✅ |
| tuple target | **0** ❌ | 1 ✅ |
| attribute target (`self.holder`) | **0** ❌ | 1 ✅ |
| two-hop chain | **0** ❌ | 1 ✅ |
| returned from a helper | **0** ❌ | 1 ✅ |
| walrus | 1 ✅ | 1 ✅ — **invariant guard, not regression coverage** (it already counted, for an unrelated reason: the consumer's argument is walked recursively) |
| a cache dir / scratch file / non-consumer call | 0 ✅ | 0 ✅ — the no-false-accusation control |
| **the same binding moved into a `@pytest.fixture`** | **0** | **0** — ⚠ **round 2: still 0. NOT closed.** See residual 6 |
| real site count on `test_subsystem_store_api.py` | 20 | 33 |

### Round 2's own matrix

| control | before | after |
|---|---|---|
| `test_a_store_root_bound_in_a_pytest_FIXTURE_is_NOT_counted`, `_ROOT_NAMES` widened to include `"served"` (closes the hole by the spelled route) | — | 🔴 RED with its own message: *"the ratchet now SEES a store root bound in a fixture"* |
| same guard, `"running"` dropped from `_ROOT_CONSUMERS` | — | 🔴 RED on its positive-control assertion first: *"the inline form of this probe counts 0, not 1"* |
| the four affected files, merged tree | 794 passed | **795 passed** (+1, the new gap guard) |

Everything else in round 2 is prose. The one code change with no red→green story is deleting the `PEAK_STORE_*` globals — nothing read them, which is precisely the finding.

### The entry census (F1, F2, F3) — real fixtures, real stores

Each row mutates `test_cairn_cli.py`'s bulk-write loop and then reports **what the ledger suite said at base** vs **what the runtime guard says now**.

| mutation | real files | `origin/main` ledger suite | this PR, running the fixture |
|---|---|---|---|
| *(unmutated)* | 306 | 8 passed | 1 passed |
| one 5-byte scratch write, no store/server/tmpfs | 306 | 🔴 **1 failed** — "now 443 entries, needing 1,880,064… but is 1,875,968" | ✅ 20 passed |
| the same 5 bytes, in `test_cairn_write.py` | 306 | 8 passed — that file's **whole** seed is dropped (F3's LOUD→SILENT half) | ✅ green |
| nested `range(4)`×`range(300)` | 1,203 | red **by one page**, reporting **304** — a 2.7x under-report that only bit because of F1's zero slack | 🔴 caught: **4,927,488 B across 1,203 entries** |
| `range(0, 1200)` | 1,203 | **8 passed — SILENT** | 🔴 caught, same exact count |
| list comprehension, 1,200 | 1,203 | **8 passed — SILENT** | 🔴 caught, same exact count |
| loop body extracted to a helper, 1,200 | 1,203 | **8 passed — SILENT** | 🔴 caught, same exact count |
| plain `range(1200)` — the batch's positive control | 1,203 | red (caught, reporting 1,339) | 🔴 caught |

The positive control being caught **at base** is what says the harness was running the code it edited.

Note the F2 row: fix F1's headroom without fixing F2 and that row goes green at base — the ENOSPC guard silently stops working. That is why this is one change.

---

## Mutation results

Every mutant was required to be killed **by its designated test**, and that test was then re-run **in isolation** and its own message checked — "something went red" was not accepted as a kill. A known-caught positive control rode in the batch.

**17 defect mutants killed by their designated test; one row that was reported as a kill does not reproduce and is corrected below.**

| mutant | verdict |
|---|---|
| `page_allocated_bytes` sums `st_size` instead of pages | KILLED — `test_page_allocated_bytes_counts_PAGES_not_apparent_size` |
| drop `max(1, …)` so an empty file costs nothing | KILLED — same |
| count directories as entries | KILLED — same |
| `<=` → `<` at the budget boundary | KILLED — `test_a_store_just_UNDER_the_budget_is_accepted` |
| `store_root` stops calling `_check_store_budget` | KILLED — `test_store_root_INVOKES_the_budget_check_on_the_root_it_yielded` |
| enforce on the tmpfs branch only | KILLED — `test_the_budget_check_runs_on_the_DISK_FALLBACK_branch_too` |
| guard runs even when the body raised (masks the real failure) | KILLED — `test_the_budget_check_does_NOT_mask_the_bodys_own_failure` |
| budget re-set to the peak (zero headroom, the F1 regression) | KILLED — `test_the_budget_keeps_real_HEADROOM_over_the_measured_peak` |
| `_BUDGET_HEADROOM` below the pinned 1.25 floor | KILLED — same |
| ~~budget re-literalised so it can disagree with the headroom~~ | 🔴 **CORRECTED — SURVIVES, and must.** See below |
| `_MIN_FREE_BYTES` lowered so the margin drops under 2x | KILLED — `test_the_free_space_floor_keeps_the_MARGIN_its_comment_claims` |
| index stops walking `node.keywords` | KILLED — `test_a_store_root_passed_BY_KEYWORD_counts_like_a_positional_one` |
| index reverts to a spelled name set | KILLED — `test_the_site_index_keys_on_USE_not_on_the_variables_NAME` |
| Constant arm reverts to `_ROOT_NAMES`-only | KILLED — `test_the_site_index_does_not_key_on_the_DIRECTORY_being_spelled_store` |
| Constant arm counts every constant (the false accusation) | KILLED — `test_a_path_that_is_never_used_as_a_store_still_does_NOT_count` |
| index leaks names across function scopes | KILLED — `test_one_tests_store_root_does_not_vouch_for_ANOTHERS_scratch_directory` |
| index drops the fixpoint (multi-hop chains) | KILLED — `test_the_site_index_sees_the_BINDING_FORMS_a_plain_assignment_is_not` |
| **positive control**: `_PAGE_BYTES = 1` | KILLED — the batch was running the code it edited |

🔴 **The corrected row.** Replacing the computed `_LARGEST_STORE_BYTES` with the literal `1880064` was reported as KILLED. Re-run in an isolated tree under `PYTHONDONTWRITEBYTECODE=1`: **20 passed — it SURVIVES.** It must: at today's constants the literal *equals* the computed value, so the two programs are semantically identical and no test can separate them. The harness was validated on the same tree in the same session — `_BUDGET_HEADROOM 1.5 → 1.1` is KILLED by `test_the_budget_keeps_real_HEADROOM_over_the_measured_peak`, the very test the row credited — so this is a mis-scored row, not a broken sweep. **The hazard the row claimed to cover is still genuinely closed**, by that test's `budget == expected` assertion: a headroom reading 1.5 beside a budget carrying 1.26x fails it. What is not true is that the re-literalisation mutant demonstrates it.

**Two further rows that are not kills, reported because a sweep that hides them is not a sweep:**

* `_BUDGET_HEADROOM = 1.5 → 1.3` **SURVIVES, and that is correct.** 1.3 is still above the pinned 1.25 floor, and because the budget is *computed* from the headroom the two cannot disagree — so this is a legal value, not a defect.
* One earlier run reported `ANCHOR matched 2 times` for that mutant — the harness saying **the mutation never applied**, not that it survived. Fixed and re-run rather than counted.

---

## What I could NOT close — stated, not hidden

1. **The runtime guard is a claim about stores that are actually BUILT by a run.** A fixture nobody executes is not measured. The gate runs the whole suite and `test_every_ledgered_file_IMPORTS_AND_CALLS_the_shared_siting_at_least_once` keeps the ledgered files routed through the seam, so together they cover the population — but a deselected subset run is not that claim, and I have written that into the source rather than left it implied.

2. **`_MEASURED_PEAK_STORE_BYTES` is still a transcribed measurement, and nothing re-runs it.** It sets a *lower* bound on the budget (budget ≥ peak × 1.5); the *upper* bound on reality is the runtime guard, which cannot be fooled. So a real store growing to 1.4× the peak is silent — under budget, and correctly so. Round 2 deleted the `PEAK_STORE_*` instrumentation that produced the number, so re-deriving it means adding a print and re-running the three ledgered files; the source now says so.

3. **`_ROOT_NAMES` is still a spelled set, and it is doing more work than "a shortcut".** 🔴 **This body originally called its 7 sites "built and used without ever reaching a consumer this file knows about". Round 2 measured them and that is false of all seven.** They are every `tmp_path / "store"` at :11310, :11765, :13099, :14086, :14201, :14227, :14803 of `test_subsystem_store_api.py`, and they split into two *masked gaps*:
   * :11310 and :13099 are `root = tmp_path / "store"` in a `_phases` helper whose NESTED `present()`/`absent()` closures call `_build_store(root, …)` — a consumer `_ROOT_CONSUMERS` **does** name. Invisible only because `_walk_scope` stops at a nested function boundary, so no one scope sees both the binding and the call.
   * the other five reach `api.append_bullet` / `api.rc.load_index` — real store consumers `_ROOT_CONSUMERS` does not name.

   Widening `_ROOT_NAMES` to "any constant" is still measured-and-rejected (it re-creates round 3's false accusation), but deleting it would not "lose seven non-flowing sites" — it would expose two structural gaps.

4. **`_ROOT_CONSUMERS` closes RENAMES, not GROWTH.** The rename argument is sound: a callee name is a real referent, so renaming `_build_store` renames its definition too. But a ratchet's job is to notice growth, and growth arrives as a NEW consumer name. Measured with this PR's own control probe: `running(served)` counts **1**, `serve_store(served)` counts **0** — same expression, same copy, same writes. Only half the spelled-guard class is closed, and residual 3 is what that already costs on the real file.

5. **The 20 → 33 sites are now visible debt, not converted.** This PR does not convert them to `store_root()` — a scripted mass conversion was attempted in an earlier round and silently skipped 9 of 19 edits while its own assertion passed. The ratchet records them.

6. 🔴 **A store root bound in a `@pytest.fixture` is NOT counted — this replaces the residual this body used to disclose.** Measured with this PR's own paired control: `served = tmp_path / "served"` written into and handed to `running(...)` counts **1** inline and **0** when the identical binding moves into a fixture. Fixture-binding is the dominant idiom in this file. **Two live instances**, both real `cp -a` copies of a store, written into and served in-process:
   * `served = tmp_path / "ordered-served"` — fixture `shuffled_pair` (:3945), served at :4392 and four more `running(served)` sites;
   * `served = tmp_path / "ambig-served"` — fixture `ambiguous_pair` (:4294), served at :4758 and one more.

   Not closed, deliberately. A pytest fixture crosses the scope boundary through pytest's own name injection, which is not a Python binding an AST can see; and **both live instances return a TUPLE the consumer destructures** (`local, served = shuffled_pair`), so following it needs a new cross-scope resolver *plus* tuple-element tracking. **File-wide indexing is not the shortcut**: measured on the real file, per-scope counts **33** and file-wide **39**, and of those six extra sites four are `tmp_path / "stage"` (:2658, :3619, :3647, :3750) that are `--stage` arguments to a `seed.sh` SUBPROCESS, never served in-process — file-wide counts them only by colliding with the genuinely-served `stage` at :5710/:5726, which is round 3's false accusation exactly. Two accidental catches for four false accusations.

   Pinned by a new gap guard, `test_a_store_root_bound_in_a_pytest_FIXTURE_is_NOT_counted`, so closing the hole fails the suite and forces this residual and `_DISK_ROOTED_SITES` to move in the same commit. **The honest statement of what the predicate covers is therefore: the flow case WITHIN ONE FUNCTION SCOPE — not the flow case generally.**

7. **The budget does not cover a concurrent writer filling the mount.** That half of the module docstring's residual window is still open and cannot be closed from inside this module; the docstring now says which half moved and which did not.

---

## Round 2 corrections — what this body used to say

| claim as written | measured | fix |
|---|---|---|
| the 13 sites enumerate to `served ×5 + served-elsewhere ×3 + stage ×2 + big + at-the-cap + unambiguous + absent ×2` | that sums to **15**, and `served` occurs **3** times | breakdown re-measured; now a table that sums to 13 |
| "the instrument was watched observing 448 things" — offered as a positive control | a **one-off hand measurement** using globals nothing read; the suite never re-runs it | reframed as a manual measurement; the two standing, mutation-verified controls are named instead, and the dead globals are deleted |
| mutation table: "budget re-literalised so it can disagree with the headroom \| KILLED" | **SURVIVES** (20 passed) and must — the literal equals the computed value | row corrected; the hazard's real guard named; harness validated on the same tree |
| residual 3: the 7 `_ROOT_NAMES`-only sites "never reach a consumer this file knows about" | false of **all seven** — 2 reach `_build_store` across a nested-function boundary, 5 reach `api.append_bullet` / `api.rc.load_index` | rewritten as two masked gaps |
| the disclosed residual was about `_ROOT_NAMES` coverage | the larger, undisclosed one is fixture-bound roots | residual 6 added, pinned by a gap guard |

Also corrected in the source, without a body claim to match: `_ROOT_CONSUMERS`'s comment overclaimed (residual 4 now states the growth half); `_assignments`'s docstring listed four binding shapes as "here" when the helper-return case lives in `flowing_callees` and `ast.withitem` — which the body handles — was unmentioned; `"os.fspath"` removed from `_PATH_WRAPPERS`, where it could never match (`_unwrap` compares `expr.func.attr`, i.e. `"fspath"`); a stale "the 20 inline `tmp_path / \"store\"` sites" now reads 33.

---

## Tier — name it, because "the gate passed" is true of one run, one tier, one base

**Dev-host `pytest` tier**, the four affected files, merged tree (`origin/main` `946a51f0` merged in), commit `5a3d7fe7`: **795 passed, rc 0**. Round 7's baseline was 794; the +1 is the new gap guard. Command: `nix develop -c env PYTHONDONTWRITEBYTECODE=1 python3 -m pytest <4 files> -q -p no:cacheprovider`.

**`nix build` sandbox tier** — the one Tekton runs and the merge is gated on — `pytests` and `nodetests` built **one at a time**, never combined (a combined invocation contends on the nix store and produces false failures). Results posted as a comment on this PR. **A dev-host green is not a claim about that tier**, and the two are structurally different environments.

⚠ Whether either check actually *blocks* is branch protection, not something this PR can assert — and #1232 is open reporting that `main` is protected in name only. Check it rather than reading a green tick as a gate.

### DIFF
--- a/scripts/testlib/store_siting.py
+++ b/scripts/testlib/store_siting.py
@@ -42,6 +42,10 @@ docstring claimed that.** Two residual windows, both real and both narrower than
 free-space floor above rather than closed by it:
   * a tmpfs that passes the checks and then FILLS mid-run — a concurrent writer, or a
     suite far larger than these stores — surfaces ENOSPC where disk would have passed.
+    Half of that is now closed and half is not, and the halves are different in kind:
+    a store of OURS growing past what `_MIN_FREE_BYTES` was sized for is caught by
+    `_check_store_budget`, which walks the real tree at teardown; a CONCURRENT writer
+    filling the mount is not, and cannot be from in here.
   * a run killed by SIGKILL (a gate timeout, `panic: test timed out`) skips the
     `finally`, so `/dev/shm/devrc-store-*` survives; on a persistent container
     `/dev/shm` repeated kills accumulate toward the first case. Nothing reaps them.
@@ -70,40 +74,97 @@ _DEFAULT_CANDIDATE = "/dev/shm"
 # one exist.
 _MOUNTS_PATH = "/proc/mounts"
 
-# The largest store this suite builds, in tmpfs PAGE-ALLOCATED bytes.
-#
-# 🔴 DERIVED, NOT TRUSTED: this literal is checked against the fixture that produces
-# it by `test_store_siting_ledger.py::test_the_largest_store_constant_still_covers_
-# the_biggest_fixture`, which reads the `range(N)` out of `test_cairn_cli.py`. Without
-# that, growing the fixture silently re-opens the ENOSPC hazard with every guard green
-# — MEASURED by an audit: `range(303)` -> `range(1200)`, all guards pass, and a
-# /dev/shm of 4500k then dies `Errno 28`. A constant checked only against another
-# constant is not a guard.
-#
-# Measured 2026-09-01 against the REAL fixture (`_populate_source_store` + 303 bulk
-# entries = 306 files, on tmpfs): apparent **74,613 B**, page-allocated
-# **1,253,376 B**, ratio **16.80x**. Confirmed four independent ways — `sum(st_size)`,
-# `st_blocks*512`, `du -sB1`, and the `/dev/shm` statvfs used-delta — all agreeing.
-#
-# 🔴 TWO EARLIER REVISIONS OF THIS COMMENT WERE MEASUREMENTS OF SOMETHING ELSE, and
-# the second was mine correcting the first. "53,985 B / 23x" came from a SYNTHETIC
-# store, not this fixture. "72,602 B / 1,265,664 B / 17.4x" then double-counted
-# DIRECTORY pages: tmpfs directories cost zero, and 1,253,376 + 3*4096 = 1,265,664
-# exactly. In a module whose subject is a number being wrong, three revisions were
-# needed to get this one right — which is why the value below is DERIVED and pinned
-# rather than transcribed.
-#
-# 🔴 APPARENT BYTES UNDERSTATE TMPFS COST ~17x. tmpfs charges whole 4 KiB pages, so
-# entry COUNT drives this, not text size; a floor reasoned from `st_size` lands an
-# order of magnitude low.
-#
-# The value is the DERIVED requirement, not the single-fixture measurement: the
-# ratchet sums every ledgered file's fixtures (conservative — under `-n 4 --dist
-# loadfile` they do not all hold stores at once) and adds slack pages.
-_LARGEST_STORE_BYTES = 1875968
+# tmpfs charges whole 4 KiB pages per file, and directories cost nothing. Both halves
+# are load-bearing: an earlier revision of the measurement below double-counted
+# directory pages and landed 12,288 B high (1,253,376 + 3*4096 = 1,265,664 exactly).
+_PAGE_BYTES = 4096
 
-# Free space a candidate must have before we will site a store on it: the measured
-# peak above, with better than 3x margin.
+# The peak page-allocated footprint any store in this suite has been MEASURED to reach.
+#
+# 🔴 THIS IS A MEASUREMENT, AND IT IS NOT WHAT MAKES THE BUDGET HONEST. Rounds 4-6 tried
+# to keep `_LARGEST_STORE_BYTES` truthful by DERIVING it from a syntactic sweep of the
+# ledgered test files — counting `write_text` calls and `range(N)` loops. Every revision
+# of that sweep was walked through by a shape it could not see: two write loops in one
+# test, nested loops summed instead of multiplied (`for a in range(4): for b in
+# range(300)` reported 304 for 1,200 real files), `range(303)` respelled as
+# `range(0, 303)`, a for-loop rewritten as a list comprehension, a loop body extracted
+# into a helper. Each of those under-reported SILENTLY with the whole suite green —
+# re-arming the ENOSPC hazard through the guard added to prevent it.
+#
+# A syntactic sweep can only ever see the shapes its author imagined. So the enforcement
+# moved to `_check_store_budget` below, which WALKS THE REAL STORE at teardown and
+# raises when it exceeds the budget. That cannot be fooled by a spelling, a nesting or a
+# refactor, because it is not reading the source at all.
+#
+# Measured 2026-09-02 by `store_root` itself over a full run of the three ledgered files
+# (`scripts/tests/test_{subsystem_store_api,cairn_write,cairn_cli}.py`): **448 stores**
+# walked, largest 1,253,376 B / 306 entries, second largest 176,128 B / 43 entries. The
+# peak is `test_cairn_cli.py`'s concurrency fixture — `_populate_source_store`'s 3 seed
+# entries + 303 bulk entries = 306 files, apparent 74,613 B — a 16.8x ratio, confirmed
+# four independent ways (`sum(st_size)`, `st_blocks*512`, `du -sB1`, and the `/dev/shm`
+# statvfs used-delta).
+#
+# ⚠ SAY WHAT WAS MEASURED, because this constant has been wrong twice by being a
+# measurement of something else. The raw `PEAK_STORE_BYTES` from that run reads
+# 1,884,160 — bigger than the number below — and that peak is the ledger suite's OWN
+# deliberately-over-budget probe, not a fixture. 1,253,376 is the largest store the
+# suite builds in the course of testing something else, which is what this constant is
+# for. The gap to the runner-up (176,128) is 7x, so the peak is not a close call.
+#
+# 🔴 APPARENT BYTES UNDERSTATE TMPFS COST ~17x. Entry COUNT drives this, not text size;
+# a floor reasoned from `st_size` lands an order of magnitude low.
+_MEASURED_PEAK_STORE_BYTES = 1253376
+
+# Headroom the budget carries over that measured peak.
+#
+# 🔴 THE PREVIOUS VALUE HAD ZERO SLACK AND THAT MADE IT A TRIPWIRE, NOT A BUDGET.
+# `_LARGEST_STORE_BYTES` was 1,875,968 = (442 + 16) * 4096, where 442 was the sweep's
+# own output — so the required value and the constant were the same number and the gate
+# sat exactly on its own boundary. Appending a single unrelated five-byte scratch write
+# to a ledgered file turned the required merge check RED, on a machine with no store,
+# no server and no tmpfs involved. Measured base rate: that derived requirement moved
+# SIX times in nineteen commits, three of them on one day.
+#
+# 1.5x is chosen so a fixture can grow by half again — the kind of change an author
+# makes without thinking about tmpfs — before anyone has to touch a constant, while
+# still leaving `_MIN_FREE_BYTES` (4 MiB) more than 2x above the budget. It is pinned
+# by `test_store_siting_ledger.py::test_the_budget_keeps_real_HEADROOM_over_the_
+# measured_peak`, so shrinking the slack to buy room is a visible act, not a quiet one.
+_BUDGET_HEADROOM = 1.5
+
+# The budget: no store this suite builds may page-allocate more than this.
+#
+# 🔴 ENFORCED AT RUNTIME, against the real directory tree, by `_check_store_budget`.
+# It is not a floor reasoned from source and it is not checked against another constant:
+# every store `store_root` yields is walked at teardown and compared to this number, so
+# a fixture that grows past it fails LOUDLY on the test that grew it, whatever spelling
+# the growth arrived in.
+#
+# 🔴 COMPUTED, NOT TRANSCRIBED — and that is not tidiness. Written as the literal
+# 1,880,064 it would be a THIRD number that can disagree with the two above it, so
+# `_BUDGET_HEADROOM = 1.5` could read 1.5 while the budget carried 1.26x and nothing
+# would notice. That is the same defect as the "better than 3x margin" comment one
+# constant further down, which was false for a whole round. Rounded UP to a whole page
+# because a budget is a page count.
+_LARGEST_STORE_BYTES = (
+    -(-int(_MEASURED_PEAK_STORE_BYTES * _BUDGET_HEADROOM) // _PAGE_BYTES) * _PAGE_BYTES
+)  # 1,880,064 at the two values above — a snapshot, not a second definition
+
+# Free space a candidate must have before we will site a store on it: the budget above,
+# with better than 2x margin. 🔴 **2x IS THE CLAIM AND IT IS THE CLAIM A TEST READS**
+# (`test_the_free_space_floor_keeps_the_MARGIN_its_comment_claims`). Today the ratio
+# happens to be 2.23x (4,194,304 / 1,880,064), and that figure is a SNAPSHOT — do not
+# promote it to the guarantee, because a snapshot in this position is precisely what
+# went stale last time.
+#
+# ⚠ The margin used to be stated as "better than 3x" and round 6 falsified that without
+# updating the sentence — it was 2.24x by then, and had been 3.18x when written. The
+# fix is not a better sentence — it is that the RATIO IS NOW READ BY A TEST.
+# `test_the_floor_CONSTANT_clears_the_largest_store_this_suite_builds` in
+# `test_subsystem_store_api.py` enforces the bare inequality;
+# `test_store_siting_ledger.py::test_the_free_space_floor_keeps_the_MARGIN_its_comment_
+# claims` enforces the 2x promised above. A prose margin nothing reads is how the "3x"
+# survived being false, and rewording it would have left that mechanism intact.
 #
 # 🔴 THIS WAS 1 MiB FOR EXACTLY ONE COMMIT AND THAT WAS A REGRESSION — BELOW the peak.
 # It was lowered from 8 MiB on the strength of an observation that 8 MiB rejects a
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Sandbox tier verdict — `nix build .#checks.x86_64-linux.pytests`

**`RESULT: PASS (exit=0)`**

```
TOTAL collected=20503  passed=20500  skipped=3  failed=0
       (floor: 18404 = sum of 30 per-target floors)
```

Read from the runner's own `RESULT:` line and its per-target `collected/passed/skipped/floor` counts, not from a piped exit code. `NIXBUILD_PYTESTS_RC=0` agrees with it. Built **alone**, not combined with `nodetests` — a combined invocation contends on the nix store and produces false failures.

This is the tier Tekton runs and the merge is gated on, and it is a different environment from the dev-host run (`cp -r ${./.}` store copy with no `.git`). The dev-host tier reported **794 passed, rc 0** over the four affected files. Both are stated because neither is evidence for the other.

The `nodetests` derivation is building separately; this diff changes no `.mjs`, so it cannot reach that tier, but it is a separate required check and I will post its verdict rather than assume it.

---

⚠ **A green tick here is not a gate right now.** Measured on this PR:

```
$ gh api /repos/innovation-upstream/devrc/branches/main/protection \
    --jq '{required: .required_status_checks, admins: .enforce_admins.enabled}'
{"admins":false,"required":null}
```

`required_status_checks` is **null** and `enforce_admins` is **false** — `main` currently requires nothing, and `gh pr checks` reports **no checks at all** on this branch while `mergeStateStatus` reads `CLEAN`. That `CLEAN` is the absence of a gate, not the passing of one. This corroborates open PR #1232 ("`main` is protected in NAME ONLY"). The two runs above are the only gate evidence this PR has.

ZacxDev:
## Node sandbox tier — `nix build .#checks.x86_64-linux.nodetests`

**`RESULT: PASS (exit=0)`**

```
```

Built alone, after the pytests derivation finished — never concurrently. `NIXBUILD_NODETESTS_RC=0` agrees with the runner's own `RESULT:` line.

Both required tiers are now green **on this branch**. Per the repo's own note, `strict` is false, so that is a claim about the branch and **not** about the tree a merge creates — gating the merged tree is still a manual step, and `main` has moved since `3d0695c7`.

ZacxDev:
## Merged-tree gate — `main` moved after the runs above

`origin/main` advanced `3d0695c7` → `30b0e7dc` (#1238) while this was in flight, so the tier results above are claims about a base that no longer exists. Rather than reason about it, I merged current `main` into this branch (`54ca5ee2`) and re-ran.

`git merge-tree --write-tree` exits **0** (branching on the exit code, not grepping for conflict markers, which `merge-tree` never emits).

The delta is one file — `claudedocs/handoff-tmux-webapp.md`, +115/−11. **Disjoint from my diff, which is not the same as safe**, and here it genuinely is not: `scripts/tests/test_no_public_ips.py:105` carries an allowlist entry keyed on that exact path, so a test does have a stake in it. Run on the merged tree:

```
test_store_siting_ledger.py + test_no_public_ips.py -> 35 passed
```

The pushed merge commit is what Tekton now gates, so its `devrc-pytests`/`devrc-nodetests` runs are against the merged content rather than the pre-merge branch.

⚠ Correction to my earlier comment: the checks **do run** — `tekton/devrc-pytests` and `tekton/devrc-nodetests` are posting and currently `pending`. What they do not do is **block**: `required_status_checks` is still `null`. "A check runs" and "a check gates" are separate facts and I conflated them by reporting `no checks reported` before Tekton had picked the branch up.

PR #1243, opened 2026-09-02
docs(handoff): rank 1 deployed and verified live — and three of the four ranks were stale on re-measurement

ZacxDev:
Rank 1 of `claudedocs/handoff-resume-state-worktree-resolution.md` is done, and re-measuring the other ranks found three of the four stale. One doc, +139/-24, written through `handoff_doc.py` (not by hand — see the gotcha below).

## What changed

**Deployed (rank 1).** `ship.sh` converged both hosts to `aa1c2abf`; a later run took them to `946a51f0`. Read per-host, not by verdict: 0 dangling managed symlinks on either, both switched, no host skipped. The two halves deploy on different triggers, so both were checked with the arbiter before and after — `scripts/resume-state.sh` resolves to itself (live on the pull, `worktrees_holding` ×6), and `~/.claude/skills/resume/SKILL.md` moved `/nix/store/p900zd0v…` → `iy8zczaj…` and is now byte-identical to `origin/main`. It **differed** before the switch, so the skill half was genuinely stale. Behaviour verified on the defect itself, not just the artifact: a handoff-shaped path absent from the base clone but present in 28 linked worktrees now refuses to choose and reconciles nothing, instead of falling back to newest-of-N.

**dnsmasq (was rank 3) is already applied — the script must not be run.** `/etc/nixos/configuration.nix:76` already carries the `/docker.io/1.1.1.1` route and its precondition now matches 0. It is live, not merely edited: TTL 49 s and 8/8 `*.docker.com` via `127.0.0.1`. But the root cause is live and **worse** — `dig @192.168.50.1` still returns the pin at 41,697,864 s (≈482 d), and its eight addresses are now **0 correct / 2 third-party certs / 6 no handshake**, against 4/2/2 when measured on 2026-08-29.

**The CI-flake mechanism is confirmed, and this doc's retention claim was wrong.** It read "Tekton retains ~14 pipelineruns … which is why none of these four has a preserved log". Measured: **235 pipelineruns**, `devrc-ci` back 16 h, gate-pod logs readable. Two retained reds named the arm themselves — `MECHANISM = SERVER_BLOCKED_IN_FSYNC` → `os.fsync` at `server.py:1996` — and both **predate** `#1219` (`b4fde334`), which sites the fixture that stalled. `store_siting.py` flags that CI may silently fall back to disk; probed on the pinned node, it does not (`/dev/shm` tmpfs, rw, 64 MiB vs a 4 MiB floor, ~9× headroom at `-n 4`).

⚠ **2 of the 4 named tests remain unexplained** — `test_no_test_writes_a_usr_bin_env_shebang_at_runtime` and `test_release_deletes_the_ref_and_the_slug_becomes_claimable_again` contain 0 references to `build_server`/`store_siting`. Recorded as *2 of 4 explained*, deliberately not as "solved".

**Unprotected `main` recorded as deliberate.** Per `CLAUDE.md`, the operator turned the gate off until Tekton capacity is addressed; `drift-check.sh` rc 24 is expected, not a finding. This session first read it as a botched break-glass restore and filed it as a rank to restore — the API shape is identical either way, which is itself the lesson now written down.

## Gate

`main` has no required checks today, so this was gated by hand rather than by the merge. Both tiers, one at a time, on the merged tree — tiers and base sha named in the comments below.

## Notes

- Nothing here changes code. The doc's own `## How to verify` block gains a post-switch byte-identity check and a `drift-check` step that says never `| tail`.
- New gotchas: a handoff is a mutable file on a moving branch (this session acted on a rank already retracted on `origin/main` — 5th occurrence of that lesson); `drift-check.sh | tail` reports rc 0 for a run that exited 24; `| tail -20` of a 235-row list reads as the whole population; and `handoff_doc.py`'s buckets are not uniform — REPLACE sections lose what you omit, APPEND sections duplicate if you re-supply them whole.

### DIFF
--- a/claudedocs/handoff-resume-state-worktree-resolution.md
+++ b/claudedocs/handoff-resume-state-worktree-resolution.md
@@ -37,14 +37,31 @@ happened.**
 - **Session artifacts cleaned**: three worktrees (`devrc-rsw`, `devrc-merged`,
   `devrc-scaffold`) removed, `/tmp` fixtures removed, 0 of mine remaining.
 
-🔴 **NOT DEPLOYED — and the two halves deploy on DIFFERENT triggers.** Measured with the
-arbiter, not inferred:
-| path | `readlink -f` | goes live on |
-|---|---|---|
-| `scripts/resume-state.sh` | **itself** | a plain `git pull` — 0 references in `nix/` |
-| `claude/skills/resume/SKILL.md` | `/nix/store/…-devrc-claude-skills/resume/SKILL.md` | `home-manager switch` |
-Pull without switching and the narrowed resolver is LIVE while the deployed prose still
-describes the old rule — the one state that actively misleads an agent.
+✅ **DEPLOYED to both hosts 2026-09-02.** `ship.sh` converged workbench + laptop to
+`aa1c2abf` (workbench `1cc720ba`→, laptop `26e16eca`→ — the laptop was 4 commits further
+behind), 0 dangling managed symlinks on either, both switched, no host skipped.
+
+The two halves go live on DIFFERENT triggers, which is why a partial deploy was the hazard.
+Measured with the arbiter, not inferred — BEFORE the switch and AFTER:
+| path | `readlink -f` | goes live on | after the switch |
+|---|---|---|---|
+| `scripts/resume-state.sh` | **itself** | a plain `git pull` — 0 references in `nix/` | `worktrees_holding` ×6 |
+| `claude/skills/resume/SKILL.md` | `/nix/store/…-devrc-claude-skills/resume/SKILL.md` | `home-manager switch` | store path moved `p900zd0v…`→`iy8zczaj…`, now byte-identical to `origin/main` |
+It DIFFERED from the repo copy before the switch, so the skill half was genuinely stale.
+The misleading half-deployed state never existed on either host.
+
+🔴 **Behaviour verified end to end on the DEFECT, not just the artifact.** Naming a
+handoff-shaped path absent from the base clone but present in linked worktrees now refuses
+to choose instead of falling back to newest-of-N:
+```
+! requested handoff "…/handoff-discord-embed-ext-rescue.md" — NO SUCH FILE, and … exists
+  in 28 worktrees of the clone that path resolves against (…, and 24 more), so NONE was
+  chosen. NOTHING was reconciled; the DRIFT section below is about no document at all.
+```
+⚠ **The verifier recipe below asks for a doc in EXACTLY ONE worktree and no such doc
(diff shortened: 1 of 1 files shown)


ZacxDev:
## Gate — both tiers, one derivation at a time, on the MERGED tree

`main` has no required checks today, so this was gated by hand. Base `5a82aaa9`, PR head `cc246793`, merged tree `6a74d41c`. Read from each runner's own `RESULT:` line, never a piped exit code.

| tier | suite | verdict | counts |
|---|---|---|---|
| dev host (`run-tests.sh`) | pytest | `RESULT: FAIL (exit=1)` | collected=20491 passed=20487 skipped=3 **failed=1** (floor 18404) |
| dev host (`gate.sh`) | node | `RESULT: PASS (exit=0)` | suites=5 files=41 tests=1449 pass=1449 fail=0 (floor 1367) |
| sandbox (`nix build .#checks…pytests`) | pytest | `RESULT: FAIL (exit=1)` | collected=20491 passed=20487 skipped=3 **failed=1** (floor 18404) |
| sandbox (`nix build .#checks…nodetests`) | node | `RESULT: PASS (exit=0)` | suites=5 files=41 tests=1449 pass=1449 fail=0 (floor 1367) |

### The one failure is PRE-EXISTING on `main`, not this PR

`scripts/collector/keylog/tests::test_live_existing_resolutions_not_made_ambiguous`

Established with a control, not by assertion: it fails **identically on a pristine detached checkout of base `5a82aaa9`** with this diff absent —

```
assert not {'recom': (':rna', None, [':acq', ':rna']),
            'recommend': (':rna', None, [':acq', ':rna'])}
```

An espanso snippet's search terms now make `:rna` ambiguous with `:acq` for the terms "recom"/"recommend". This PR touches one markdown file under `claudedocs/` and cannot reach `nix/home.nix` snippets. **`main` is red on this test independently — worth a separate fix.**

### What the gate DID catch in this PR, and which is now fixed

The first run failed `test_no_unallowlisted_public_ip_literal_is_committed`: the doc committed routable public IP literals to a PUBLIC repo, plus two real third-party hostnames used as examples — all five introduced by this branch, none present on base. `cc246793` scrubs them (measured: 0 residual occurrences whole-doc), and that test is now `1 passed` on the merged tree. The gate was right and the addresses/hostnames are deliberately not recorded; the doc now says to re-measure and read the certs yourself.

⚠ Tier 1's pytest leg initially reported `exit=3` — `logrotate` missing from PATH, i.e. no test had run at all. Re-run inside the flake dev shell as the runner itself instructs. `exit=3` is a missing environment, not a red suite.

PR #1245, opened 2026-09-02
docs(handoff): ferdium-server SHIPPED and verified live — plus a gateway outage from a landmine that had been armed for days

ZacxDev:
Closes out the ferdium-server effort. Everything is deployed and verified live.

## Merged this session

devrc **#1240** (client packages) · devrc **#1241** (prior handoff) · homelab-infra **#647** (manifests, both gateways, Authelia rule, relay firewall) · homelab-infra **#651** (incident fix).

Open and deliberately unmerged: homelab-infra **#653** — it rolls the gateway pod on each cluster, which is a window you schedule.

## What the doc now records

**Verification with its controls, not just outcomes.** The 8117 internet-facing check is recorded as a three-way probe (new port vs a known-guarded port vs an unguarded closed port) rather than "it's firewalled", because the first two look identical and only the third proves the path reaches the host.

**A section for what is verified-adjacent but NOT verified.** `{"api":"success","db":"success"}` is hardcoded in `HealthController` and never queries the DB — it proves the route. A Ready pod means the same thing. The desktop client's background sync through Authelia is untested and is written up as an open investigation with its ruled-outs and next probe.

## The incident

Restarting the homelab gateway put it in CrashLoopBackOff for ~6 min across every mesh service. **Not caused by ferdium** — `promptver` was torn down days earlier but its `proxy_pass` blocks stayed, and two things hid that: no `resolver` directive (nginx resolves static upstreams at startup), and a `subPath` mount (which never receives ConfigMap updates). The running pod served a config resolved *before* the teardown, so every reconcile and every health check was green and the fault was reachable only by restarting.

Both halves are now written down as rules rather than as a story: deleting a service means deleting its gateway block in the same change, and prefer `nginx -s reload` over replacing the pod because a reload fails safe.

## Two method notes worth keeping

- **A PR can be red for a reason outside its diff.** #647's failure was purely that it branched off the commit before #646 "fix(ci): unbreak trunk". A rebase fixed it with zero code change; the tell was other open PRs passing the same leg.
- **Two probes returned confident numbers while measuring nothing**, both caught only by a control: a `/dev/tcp` check that returned identically for every port, and a config grep against a file nginx does not read — where the control port `8115` also scored 0.

### DIFF
--- a/claudedocs/handoff-ferdium-server.md
+++ b/claudedocs/handoff-ferdium-server.md
@@ -4,25 +4,28 @@
 Deploy Ferdium Server (self-hosted multi-messenger backend) on the homelab Talos cluster, exposed at `ferdium.zacx.dev` behind Authelia, so the Ferdium desktop client can sync without a cloud account.
 
 ## State now
-- **Branch / PR**: devrc `fix/ferdium-client-packages` → **PR #1240 OPEN** (client packages). devrc `docs/handoff-ferdium-server-pickup` → **PR #1241 OPEN** (this doc). homelab-infra `feat/ferdium-server` → **PR #647 OPEN**, commit `71c7d475`, rebased onto `origin/trunk` `8cfa319d` — 16 files, +633/-2. **Not merged. Nothing applied to any cluster or node.**
-- **Clawgate task**: none recorded. `clawgate_handoff.sh resolve` exited 5 (0 tasks for this session). That cannot distinguish "touched no task" from "wrong session id", so no `clawgate-task:` field was written — it is not a clean bill of health.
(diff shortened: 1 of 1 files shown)


ZacxDev:
Closing as an obsolete duplicate. This branch (`docs/handoff-ferdium-shipped`, opened 2026-09-02) carried an earlier revision of `claudedocs/handoff-ferdium-server.md`; a richer version of the same doc already landed on `main` via #1266 (`f972992e`) and #1278 (`a7dac5bd`), which is why this went `CONFLICTING`.

The current update to that doc is #1281. Nothing here is lost — the content this PR would have added is a subset of what is already on `main`.

Branch left in place.

PR #1250, opened 2026-09-03
feat(drift-check): rc 24 fires on DISAGREEMENT with a declared expectation; rc 25 catches the declaration going stale

ZacxDev:
## The problem, measured

`drift-check.service` exits **24** on every fire. Measured on the workbench 2026-09-02:

- `systemctl --user show drift-check.service` → `Result=exit-code`, `ExecMainStatus=24`
- 5 failures in 3 days, on a 6h `OnUnitActiveSec` timer — it would have continued indefinitely
- live API: `required_status_checks` absent from the protection object, `enforce_admins: false`, while `GET /branches/main` still reports `protected: true`

The unit's `OnFailure` is `notify-failure@%n.service` — the **one** toast class deliberately wired to bypass do-not-disturb (`zz_notify_failure_bypass`, `override_pause_level = 100`). `nix/home.nix` justifies that bypass by a **measured rate of ~1 firing in 9 days**. 4/day is three orders of magnitude past it.

But the finding is **true**: the merge gate really is off. It is off *deliberately* — until the Tekton capacity issue is addressed, which a different session owns. So this is a permanently-red gate reached by a third route: not a broken instrument, not a false positive, but a true finding about a state nobody intends to change.

And rc 24 could not simply be excused: it is the only thing that catches a **botched break-glass restore** — the measured 2026-08-29 shape where `DELETE` opened the window, the restore trap *ran*, and `PATCH` could not close the sub-resource.

## The design

The arm now fires on the **disagreement** between a declared expectation and what it measured:

| declared | live | verdict |
|---|---|---|
| ON | OFF | **rc 24** — the alarm, preserved whole |
| ON | ON | no code |
| OFF | OFF | no code; reported plainly, with the reason |
| OFF | ON | **rc 25** — STALE DECLARATION |
| — | `gh` cannot answer | COULD NOT MEASURE, no code, under **either** declaration |

### Why `expected=off / live=on` is rc 25 and not silence

A stale "off" **disarms rc 24**. Declared-off/live-off is the excused cell, so once protection is restored and the declaration still says off, a gate that is removed *again* lands in that same cell and passes silently — the botched-restore detector is switched off while the thing it detects is live again. Nothing is lost yet; what is lost is the ability to notice.

It ranks **between rc 17 and rc 14**, derived from this table's own two published principles:

- **below 17** — *"a hazard that has already cost something outranks one that is merely open"* (severity()'s own argument for 8 > 24). rc 17's measured failure had already shipped a binary that ran, exited 0 and silently did nothing. Here the gate is **up**.
- **above 14** — 14 is LOUD at the moment of use (`command not found`); this is silent by construction, since its whole symptom is that nothing says anything.

No escalation ladder, deliberately: unlike rc 13/16/18, its repair is a one-line edit by whoever restored the protection, so it is closable the moment it is seen.

### Where the declaration lives

`bp_declared_off_reason()` in `scripts/drift-check.sh` — an enumeration of literal `if [ "$1" = <slug> ]` arms, each carrying its reason, in the same shape and for the same stated reason as the `settings.json` per-host allowlist. It **fails closed** (an unenumerated repo is expected ON), and there is **no env override and no ledger path**: an expectation a run can set is one the run can use to excuse itself.

`DRIFT_PROTECT_SLUG` is not that hole — it changes *which* repo is asked about, never what is expected of it, so it cannot turn a finding off.

### One rule, one place

`note_rc 24` used to sit at **three** sites. Each branch now decides only `BP_LIVE=on|off` and prints its measured facts; the alarm, the repair text and the code are decided **once**, at the bottom. Adding the declaration to two of the three would have left an arm that still fires on a declared-off gate, correct at both sites anyone would think to read.

## Verification

**Live, before/after** (`--no-remote`, real `gh`, real remote):

- base `bf5516fc`: `rc=24`, `🔴 DRIFT — … ZERO required status checks`
- HEAD: rc 24 gone; prints `DECLARED OFF, and live OFF — expectation and reality agree. No rc is set.` with the reason. (Run still exits 10 — the base clone is a commit behind. Pre-existing, unrelated.)

**19 new tests**, 16 red at base `bf5516fc` / green at HEAD. 3 are **invariant guards** (green at base, labelled as such): the two COULD-NOT-MEASURE-under-a-declaration cases and the env-override refusal — the discipline pre-existed; these pin that the declaration did not create a second way to manufacture a verdict from an absence.

**Mutation sweep**: 7 targeted mutants + 1 positive control, each in a fresh `git archive HEAD` tree (no `.git` at all), `PYTHONDONTWRITEBYTECODE=1`, control green before and after, each mutation asserted to apply exactly once. Results in the PR thread.

`test_only_16_is_excused_from_failing_the_unit` is **untouched and green** — rc 24 and rc 25 are real divergences with real fixes and must reach `OnFailure`, so the excuse list stays at exactly one code.

## Prose fixed in the same commit

- `nix/home.nix` — the "what does/does not toast" ledger did not list rc 24 **at all**, which is what let this happen. rc 24 and rc 25 are now entered with the measurement.
- `nix/home.nix` — the dunst DND-bypass comment's `~1 firing in 9 days` justification is annotated with the breach it was measured against and what closed it.
- `CLAUDE.md` — the paragraph claiming rc 24 "will report an unprotected main every run: that report is EXPECTED, not a finding" is now false and is replaced; it also now says that **deleting the declaration is part of restoring protection**, not a follow-up.

The `<!-- merge-gate: other -->` marker is untouched and still the only one.

https://claude.ai/code/session_01XPiQDMvSzEQXRwmEHi4ZxK

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -166,8 +166,16 @@ Repo-level facts that are NOT in any skill — they live here on purpose:
   `protected: true`. A PR merges with both Tekton checks red, or with none posted at all.
   🔴 **DELIBERATE AND CURRENT — not drift, and not yours to "restore".** The operator turned
   it off because the gate was slowing work down; it stays off until the Tekton capacity
-  issue is addressed, which a different session owns. `drift-check.sh` rc 24 will report an
-  unprotected `main` every run in the meantime: that report is EXPECTED, not a finding.
+  issue is addressed, which a different session owns. 🔴 **That decision is now DECLARED IN
+  CODE, and the declaration is load-bearing.** `bp_declared_off_reason()` in
+  `scripts/drift-check.sh` names this repo with the reason above, so rc 24 fires on
+  DISAGREEMENT rather than on the bare state: declared-off/live-off is printed plainly and
+  sets no code (it exited 24 on every 6-hourly fire before that — 5 DND-bypassing toasts in
+  3 days), while an UNDECLARED unprotected `main` is still the alarm, unchanged. 🔴 **When
+  protection is restored, DELETE that arm in the same change** — a declaration left saying
+  "off" over a live gate DISARMS rc 24, which is `drift-check.sh` rc 25 and toasts until it
+  is removed. The declaration is deliberately not env-overridable and has no ledger path: a
+  run must not be able to excuse itself.
   ⚠ **Tekton still RUNS** — both checks post on a PR head, they just do not gate. That is
   exactly why the marker stays `other`: it records that something runs at merge time, never
   that it blocks. There is still no `.github/workflows`.
@@ -224,7 +232,10 @@ Repo-level facts that are NOT in any skill — they live here on purpose:
   lags by up to a full interval, and only where the deadman is actually wired in
   (`serverMode && enableDriftDeadman`; the timer runs on the workbench). It catches a
   botched restore after the fact; it does not undo one, and it is not a substitute for
-  step 4.
+  step 4. ⚠ **And it is only armed for a repo whose declared expectation is ON** — while
+  the arm above declares this repo's gate off, a botched restore lands in the same cell as
+  the intended state and passes silently. That is the price of the declaration, and it is
+  why deleting the arm is part of restoring protection rather than a follow-up.
   ⚠ **`strict` is FALSE on purpose.** `strict: true` would force every PR to be up to date
   with `main` before merging — correct in principle, and unworkable here: `main` moved 11+
   times in one session and each move would re-queue a ~20-minute gate for every open PR.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -175,10 +175,11 @@ Repo-level facts that are NOT in any skill — they live here on purpose:
   `nix/home.nix`'s `zz_notify_failure_bypass` block — do not restate it; it was wrong at
   every copy.) 🔴 **When protection is restored, DELETE that arm in the same change** — a
   declaration left saying "off" over a live gate DISARMS rc 24, which is `drift-check.sh`
-  rc 25 and toasts until it is removed. **A PARTIAL restore fires rc 25 too**: put the
-  required checks back and omit `enforce_admins` — the exact shape step 4 below warns a
-  partial `PUT` produces — and the checks existing again is evidence the declaration is
-  stale, so it is a finding rather than agreement. The declaration is deliberately not
+  rc 25 and toasts until it is removed. **A HALF-finished restore fires rc 25 too**, in
+  either spelling: required checks back with `enforce_admins` omitted (the exact shape step 4
+  below warns a partial `PUT` produces), or an ACTIVE ruleset listing a check but keeping
+  `bypass_actors`. A gate existing again is evidence the declaration is stale, so both are
+  findings rather than agreement. The declaration is deliberately not
   env-overridable and has no ledger path: a run must not be able to excuse itself. ⚠ The
   SUBJECT can still be redirected (`DRIFT_PROTECT_SLUG`), but the run prints the slug it
   asked about, so that silence names a repo nobody chose.
@@ -238,13 +239,16 @@ Repo-level facts that are NOT in any skill — they live here on purpose:
   lags by up to a full interval, and only where the deadman is actually wired in
   (`serverMode && enableDriftDeadman`; the timer runs on the workbench). It catches a
   botched restore after the fact; it does not undo one, and it is not a substitute for
-  step 4. ⚠ **And while this repo's gate is DECLARED off, rc 24 cannot fire for it** — a
-  restore that is later undone lands in the same cell as the intended state. Two things
-  narrow that: a PARTIAL restore (checks back, `enforce_admins` dropped — the failure step 4
-  exists for) fires **rc 25**, not silence; and a completed restore fires rc 25 until the
-  declaration is deleted. What is genuinely uncovered is a full restore, the declaration
-  deleted, and a *later* deletion — which is ordinary rc 24 again. Deleting the arm is part
-  of restoring protection, not a follow-up.
+  step 4. ⚠ **And while this repo's gate is DECLARED off, rc 24 cannot fire for it.** What
+  narrows that is rc 25, which fires on any OFF state carrying evidence a gate was
+  re-established: a completed restore, a PARTIAL one (checks back, `enforce_admins` dropped —
+  the failure step 4 exists for), and an ACTIVE ruleset that lists a check but keeps bypass
+  actors. 🔴 **The one state still uncovered is a ruleset parked in `evaluate`/`disabled`
+  mode** — not a gate anybody built, and indistinguishable at the endpoint drift-check reads
+  from the org-level evaluate rollout it must not fire on. (Closable, not structural:
+  `ruleset_source_type` is on `/rules/branches/main` and is simply not read yet.) Note a full
+  restore followed by a *later* deletion is NOT uncovered — deleting the declaration is part
+  of restoring protection, and after that it is ordinary rc 24 again.
   ⚠ **`strict` is FALSE on purpose.** `strict: true` would force every PR to be up to date
   with `main` before merging — correct in principle, and unworkable here: `main` moved 11+
   times in one session and each move would re-queue a ~20-minute gate for every open PR.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -170,12 +170,18 @@ Repo-level facts that are NOT in any skill — they live here on purpose:
   CODE, and the declaration is load-bearing.** `bp_declared_off_reason()` in
   `scripts/drift-check.sh` names this repo with the reason above, so rc 24 fires on
   DISAGREEMENT rather than on the bare state: declared-off/live-off is printed plainly and
-  sets no code (it exited 24 on every 6-hourly fire before that — 5 DND-bypassing toasts in
-  3 days), while an UNDECLARED unprotected `main` is still the alarm, unchanged. 🔴 **When
-  protection is restored, DELETE that arm in the same change** — a declaration left saying
-  "off" over a live gate DISARMS rc 24, which is `drift-check.sh` rc 25 and toasts until it
-  is removed. The declaration is deliberately not env-overridable and has no ledger path: a
-  run must not be able to excuse itself.
+  sets no code, while an UNDECLARED unprotected `main` is still the alarm, unchanged. (The
+  measured breach rate against the DND-bypass justification is stated ONCE, in
+  `nix/home.nix`'s `zz_notify_failure_bypass` block — do not restate it; it was wrong at
+  every copy.) 🔴 **When protection is restored, DELETE that arm in the same change** — a
+  declaration left saying "off" over a live gate DISARMS rc 24, which is `drift-check.sh`
+  rc 25 and toasts until it is removed. **A PARTIAL restore fires rc 25 too**: put the
+  required checks back and omit `enforce_admins` — the exact shape step 4 below warns a
+  partial `PUT` produces — and the checks existing again is evidence the declaration is
+  stale, so it is a finding rather than agreement. The declaration is deliberately not
+  env-overridable and has no ledger path: a run must not be able to excuse itself. ⚠ The
+  SUBJECT can still be redirected (`DRIFT_PROTECT_SLUG`), but the run prints the slug it
+  asked about, so that silence names a repo nobody chose.
   ⚠ **Tekton still RUNS** — both checks post on a PR head, they just do not gate. That is
   exactly why the marker stays `other`: it records that something runs at merge time, never
   that it blocks. There is still no `.github/workflows`.
@@ -232,10 +238,13 @@ Repo-level facts that are NOT in any skill — they live here on purpose:
   lags by up to a full interval, and only where the deadman is actually wired in
   (`serverMode && enableDriftDeadman`; the timer runs on the workbench). It catches a
   botched restore after the fact; it does not undo one, and it is not a substitute for
-  step 4. ⚠ **And it is only armed for a repo whose declared expectation is ON** — while
-  the arm above declares this repo's gate off, a botched restore lands in the same cell as
-  the intended state and passes silently. That is the price of the declaration, and it is
-  why deleting the arm is part of restoring protection rather than a follow-up.
+  step 4. ⚠ **And while this repo's gate is DECLARED off, rc 24 cannot fire for it** — a
+  restore that is later undone lands in the same cell as the intended state. Two things
+  narrow that: a PARTIAL restore (checks back, `enforce_admins` dropped — the failure step 4
+  exists for) fires **rc 25**, not silence; and a completed restore fires rc 25 until the
+  declaration is deleted. What is genuinely uncovered is a full restore, the declaration
+  deleted, and a *later* deletion — which is ordinary rc 24 again. Deleting the arm is part
+  of restoring protection, not a follow-up.
   ⚠ **`strict` is FALSE on purpose.** `strict: true` would force every PR to be up to date
   with `main` before merging — correct in principle, and unworkable here: `main` moved 11+
   times in one session and each move would re-queue a ~20-minute gate for every open PR.
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -539,15 +539,36 @@ in
       # opt-in subset — justified by the measured firing rate (7 activations in
       # ~6 months of laptop journal, 1 in 9 days of workbench journal), which is
       # far too low to make DND feel broken.
-      # ⚠ THAT RATE IS A MEASUREMENT WITH A DATE, AND IT HAS BEEN BREACHED ONCE.
-      # 2026-08-30..09-02: drift-check.service exited 24 on every 6-hourly fire —
-      # 5 failures in 3 days through this rule — because the branch-protection
-      # arm reported a merge gate the operator had deliberately turned off. The
-      # justification above is what the breach was measured AGAINST, not
-      # something the breach retired; it was closed in the SCRIPT (see the rc 24 /
-      # rc 25 note in the drift-check block below), not by narrowing this rule.
-      # Anything wired to OnFailure=notify-failure@ inherits this bypass, so a
-      # unit that can fail on a standing condition breaches it again.
+      #
+      # 🔴 THAT RATE IS A MEASUREMENT WITH A DATE, IT IS CURRENTLY BREACHED BY
+      # ~32×, AND THIS IS THE ONE PLACE THE ARITHMETIC IS STATED. Measured from
+      # `journalctl --user -u drift-check.service`, 2026-08-25..2026-09-03:
+      #   32 unit failures in 9 days — every one of them through THIS rule —
+      #   against a justification of ~1 firing in 9 days. That is 32×, not the
+      #   "three orders of magnitude" an earlier wording claimed at three
+      #   different sites; 4/day against 1-per-9-days is 36× (~1.6 orders), and
+      #   no restatement of it was ever right. Verdicts in that window:
+      #   rc 10 ×13, rc 17 ×11, rc 24 ×5, rc 12 ×2, rc 8 ×1.
+      # Do not restate this number elsewhere — point here. It was wrong at every
+      # copy, in both directions, and a figure nobody can locate is a figure
+      # nobody re-measures.
+      #
+      # 🔴 AND THE BREACH IS NOT CLOSED. The rc-24 arm (see the drift-check block
+      # below) was the STRUCTURALLY PERMANENT contributor — the one finding that
+      # could never be repaired because nobody intended to repair it — and it is
+      # fixed. It was not the only one: rc 24 was the verdict on only 5 of those
+      # 32 runs, all inside a single 24-hour span, and on all 5 at least one other
+      # DRIFT finding was co-present (`local main is BEHIND` on 5 of 5, `BUILT
+      # SOURCE homelab-talos/containers/clawgate is NOT current` on 4 of 5).
+      # Removing rc 24 lowers those runs to rc 17 or rc 10 — still non-zero, still
+      # OnFailure, still a toast through this rule, still ~4×/day. 🔴 A reader who
+      # stops here concludes the breach was closed and stops looking, which is the
+      # exact "learns to ignore the one alert that must keep its meaning" outcome
+      # this rule's scope argument exists to prevent. rc 17 and rc 10 on this
+      # fleet are a separate, still-open job.
+      #
+      # Anything wired to OnFailure=notify-failure@ inherits this bypass, so any
+      # unit that can fail on a STANDING condition breaches it again.
       # Deliberately NOT keyed on
       # `urgency = critical`: other tools send critical toasts, and those must
       # still respect DND.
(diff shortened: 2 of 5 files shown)


ZacxDev:
## Verification results

### Live before/after — the original symptom reproduced

Base `bf5516fc`, real `gh`, real remote, `--no-remote`, `DRIFT_STATE_DIR` pinned to a scratch dir:

```
BASE rc=24
[protect] 🔴 DRIFT — innovation-upstream/devrc main has ZERO required status checks (protected=true),
```

HEAD:

```
[protect] innovation-upstream/devrc main has ZERO required status checks (protected=true),
[protect]   by CLASSIC protection and by RULESETS alike — both were checked.
[protect] DECLARED OFF, and live OFF — expectation and reality agree. No rc is set.
[protect]   why: operator decision 2026-09-02: the Tekton merge gate is off until the Tekton
              capacity issue is addressed (a different session owns that)
[protect]   🔴 The alarm is QUIET, not DISABLED: it fires the moment the two disagree.
```

The HEAD run exits **10**, not 0 — the base clone is one commit behind `origin/main`. Pre-existing and unrelated; rc 24 is gone.

### Red-at-base / green-at-HEAD, base `bf5516fc`

Base tree built with `git archive bf5516fc` (no `.git`), with **only** the new test file overlaid, so each test runs against the unmodified base script. Per-test, one pytest invocation each.

| test | base | HEAD |
|---|---|---|
| `test_a_declared_off_gate_that_is_live_off_sets_no_rc` | RED | GREEN |
| `test_a_declared_off_run_says_the_alarm_is_still_armed` | RED | GREEN |
| `test_a_declared_off_gate_that_is_live_ON_is_rc25` | RED | GREEN |
| `test_a_stale_declaration_is_reached_through_the_RULESET_path_too` | RED | GREEN |
| `test_an_UNDECLARED_repo_with_no_gate_still_fires_rc24` | RED | GREEN |
| `test_enforce_admins_false_on_a_DECLARED_off_repo_sets_no_rc` | RED | GREEN |
| `test_an_undeclared_repo_with_a_live_gate_says_so_explicitly` | RED | GREEN |
| `test_the_declaration_names_exactly_the_repos_argued_for` | RED | GREEN |
| `test_every_declared_repo_carries_a_written_reason` | RED | GREEN |
| `test_the_declaration_is_a_pure_function_of_its_argument` | RED | GREEN |
| `test_DRIFT_PROTECT_SLUG_changes_the_subject_and_never_the_expectation` | RED | GREEN |
| `test_the_arm_decides_its_code_in_exactly_one_place` | RED | GREEN |
| `test_a_stale_declaration_is_outranked_by_unpushed_commits` | RED | GREEN |
| `test_a_stale_declaration_outranks_a_merely_behind_host` | RED | GREEN |
| `test_rc25_is_ranked_between_rc17_and_rc14_in_the_severity_table` | RED | GREEN |
| `test_the_rc25_legend_is_printed_with_the_verdict` | RED | GREEN |
| `test_a_gh_that_answers_nothing_is_still_could_not_measure_for_a_declared_repo` | **GREEN** | GREEN |
| `test_no_gh_binary_at_all_is_could_not_measure_for_a_declared_repo` | **GREEN** | GREEN |
| `test_the_declaration_cannot_be_set_from_the_environment` | **GREEN** | GREEN |

**16 regression tests. The last 3 are INVARIANT GUARDS, not regression coverage** — the COULD-NOT-MEASURE discipline and the absence of an env override both pre-existed, and these pin that the declaration did not create a second way to manufacture a verdict from an absence. They are labelled as such and are not counted as coverage of the change.

### Mutation sweep — 8/8 killed

Each mutant tree built fresh with `git archive HEAD` — a tree with **no `.git` at all**, which is stronger than `cp -a` + `rm -f <copy>/.git` (there is never a moment when a `git` command inside the copy could reach the real branch). `PYTHONDONTWRITEBYTECODE=1` and `-p no:cacheprovider` on every run; a fresh tree per mutant, so a stale `.pyc` cannot score a mutant SURVIVED without executing it. Each mutation is an exact-string replace whose occurrence count is asserted `== 1`.

|---|---|---|
| M1 declaration fails **open** (`else` returns a reason) | KILLED (26 failed) | `test_an_UNDECLARED_repo_with_no_gate_still_fires_rc24`, `test_DRIFT_PROTECT_SLUG_changes_the_subject…`, + 24 |
| M2 declared slug renamed | KILLED (9 failed) | `test_the_declaration_names_exactly_the_repos_argued_for`, all six cell-3/cell-4 tests |
| M3 `note_rc 25` → `:` | KILLED (5 failed) | `test_a_declared_off_gate_that_is_live_ON_is_rc25`, `test_the_rc25_legend…`, `test_the_arm_decides_its_code_in_exactly_one_place` |
| M4 COULD-NOT-MEASURE stops short-circuiting | KILLED (183 failed) | `test_a_gh_that_answers_nothing_is_still_could_not_measure_for_a_declared_repo` + 182 |
| M5 enforce_admins site drops `BP_LIVE=off` | KILLED (2 failed) | `test_enforce_admins_false_on_a_DECLARED_off_repo_sets_no_rc`, `test_required_checks_with_enforce_admins_FALSE_is_rc24` |
| M6 ruleset-gates-it site drops `BP_LIVE=on` | KILLED (1 failed) | `test_a_stale_declaration_is_reached_through_the_RULESET_path_too` |
| M7 `25) echo 66` → `echo 20` | KILLED (2 failed) | `test_a_stale_declaration_outranks_a_merely_behind_host`, `test_rc25_is_ranked_between_rc17_and_rc14…` |
| **PC** (positive control) `note_rc 24` → `note_rc 4` | KILLED (19 failed) | the pre-existing rc-24 suite |

`CONTROL: 450 passed, 0 failed` before **and** after every batch.

**Reachability, with each arm's own message.** M4 run in isolation against only the tests that own that cell:

```
mutation applied (1 occurrence)
E  AssertionError: an unreadable gh produced a verdict: 25      <- declared repo, empty gh
E  assert 25 == 0                                                <- declared repo, no gh binary
E  AssertionError: an EMPTY gh answer was read as a count of zero and fired rc 24   <- undeclared repo
```

Each fails with **its own** arm's message, not a neighbour's, and no earlier check wins first.

M5 and M6 are the reachability proofs for the *second* live-OFF site (`enforce_admins=false`) and the *second* live-ON site (ruleset-gated). M6 is the seam case: the pre-existing `test_a_ruleset_protected_branch_is_not_reported_as_drift` still passes under M6 (rc 0 either way) — only the new test sees it.

### Two harness defects found and fixed before any number was quoted

Reported honestly because both would have produced a false SURVIVED/blank:

1. **M1's pattern matched twice** (`else\n echo ""\n fi\n}` also matches `perhost_reason`). The harness refused to score it — `MUTATION DID NOT APPLY CLEANLY (occurrences=2)` — rather than reporting SURVIVED. Re-anchored on the declared reason string.
2. **The summary parser was the instrument and was not validated.** `re.search(r"(?:(\d+) failed, )?(\d+) passed", out)` took the **first** match anywhere in the output — assertion text included — so M4 first scored **SURVIVED (0 failed)** while printing 178 killing tests. Fixed to anchor on the last summary line **and** cross-check the parsed count against the counted `FAILED` lines; that cross-check then caught a second-order bug (`\S+` truncates a parametrized id at its first space, collapsing 183 failures to 178 unique names) and again refused to score rather than report a number.

### Suites

- `scripts/tests/test_drift_check.py`: **450 passed** (431 before this change).
- Full dev-host pytest tier, `nix develop … -m pytest scripts/tests`: **11589 passed, 1 skipped, 0 failed** (18m07s).
- `test_only_16_is_excused_from_failing_the_unit`: **green and untouched**. `git diff bf5516fc..HEAD -- scripts/tests/test_drift_check.py | grep '^-'` yields only the `--- a/…` header — the test-file diff is purely additive, zero existing lines changed.

I have **not** run `scripts/gate.sh` or the two `nix` check derivations — the merged-tree gate is the reviewer's.

ZacxDev:
## Round-1 audit fixes — `e158c10c`

All six findings addressed. Every measurement below was re-taken here rather than carried over from the audit.

---

### 🟡 F1 — the headline claim was wrong

Re-measured independently from `journalctl --user -u drift-check.service`, window **2026-08-25 → 2026-09-03**:

```
32 unit failures in 9 days
verdicts:  rc 10 ×13   rc 17 ×11   rc 24 ×5   rc 12 ×2   rc 8 ×1
rc 24 timestamps:  2026-09-01T20:47 … 2026-09-02T20:47   (5 consecutive fires, one 24h span)
```

Co-presence, checked on each of the 5 rc-24 runs individually: `local main is BEHIND` on **5 of 5**, `BUILT SOURCE homelab-talos/containers/clawgate is NOT current` on **4 of 5**. So removing rc 24 lowers those runs to **rc 17 or rc 10** — still non-zero, still `OnFailure`, still a DND-bypassing toast.

The audit's numbers reproduce exactly. "Exited 24 on every fire" was false (5 of 32). Both blocks now say plainly that this change removes the *structurally permanent* contributor and **does not stop the toasts on this host**, with the explicit note that a reader who stops there is the failure the dunst block exists to prevent. rc 17 / rc 10 are named as a separate, still-open job.

### 🟡 F2 — "cannot turn a finding off" was false

Replaced with the claim that actually holds, in `drift-check.sh`'s header and in `nix/home.nix`:

> An env var **can** silence this alarm by redirecting the subject. The defensible distinction is narrower: the run **prints the slug it asked about**, so that silence names a repo the operator never chose. An expectation override would leave the correct subject and change only the verdict — nothing in the output would differ from a healthy run. **Redirecting the subject is loud; flipping the expectation is silent.** Not that one is harmless.

Scaffolding: the old test asserted only that an unenumerated slug still fires, so it never covered its own name. It is renamed to `test_a_declaration_applies_to_its_own_slug_and_not_to_a_neighbour` (what it checks), and two new tests carry the claim, driven by a new **slug-aware** `gh` stub:

- `test_DRIFT_PROTECT_SLUG_changes_the_SUBJECT_and_the_run_names_it` — asserts `rc == 0` for the gated fixture repo (**the silencing, asserted rather than denied**) *and* that the run names `fixture-owner/gated-repo` and never `fixture-owner/ungated-repo`.
- `test_the_expectation_is_still_read_from_the_enumeration_for_any_subject` — the pass comes from the enumeration, not from anything declaring it off.

### 🟡 F3 — the partial-restore regression (behaviour change)

Reproduced at HEAD before fixing: declared-off repo, `true 2 tekton/devrc-pytests,tekton/devrc-nodetests` with `enforce_admins=false` → rc 0 while printing "Restore protection and this becomes rc 25". **Fixed as you preferred**: `BP_OFF_KIND=enforce-admins` under a declaration now fires **rc 25**, with a repair naming *both* halves (finish the `PUT`; delete the declaration). The agreement cell's text no longer promises something false — it now says a partial restore fires too.

**I did not extend it to `ruleset-nonbinding`, and the asymmetry is argued in-line rather than assumed.** `/rules/branches/main` returns every *applying* ruleset, repo- and **org**-level mixed (the script's own loop comment documents this), and an org ruleset parked in `evaluate` mode is the standard rollout pattern — nothing this repo created or can remove. Firing on it would be a permanently-red gate on a state nobody can close, which is the one thing this arm refuses everywhere else. It stays in the agreement cell and converts to rc 25 the moment any ruleset **binds**. The discriminator is *positive evidence that someone restored on **this** repo* — required checks in the repo's own protection object — which the ruleset case does not carry.

**Known bound, stated rather than hidden**: a *repo-level* ruleset the operator left in evaluate mode is indistinguishable from the org case at the endpoint this arm reads. `ruleset_source_type` on `/rules/branches/main` is the field that would separate them; it is not read today, and closing it needs that read plus its own tests. Pinned by `test_a_NONBINDING_RULESET_under_a_declaration_stays_in_the_agreement_cell`.

### 🟢 F4 — wrong ladder codes

`nix/home.nix` said "unlike rc 13/16/18". Corrected to **rc 13/18/23**, each named with its env var (`DRIFT_UNREACHABLE_ESCALATE`, `DRIFT_UNMEASURED_ESCALATE`/`_FETCH_ESCALATE`, `DRIFT_NIXDIRT_ESCALATE`), with a note that rc 16 has no ladder and is the single `SuccessExitStatus` code — so naming it muddled the paragraph arguing about the excuse list, and disagreed with `drift-check.sh:563` in the same commit.

Also recorded there, per your note: `ExecStart` is `%h/workspace/devrc/scripts/drift-check.sh` — the **working tree** — so the rc-25 repair lands on merge + pull with no `home-manager switch`. That is now stated in `nix/home.nix` *and* printed in the rc-25 finding itself.

### 🟢 F5 — the figure was wrong at three sites

`4/day ÷ 1-per-9-days` = **36×** (~1.6 orders), not three orders. The measured figure is **32 unit failures in the same 9-day window the justification is stated in — 32×**. Now stated **once**, in `nix/home.nix`'s `zz_notify_failure_bypass` block, with the verdict breakdown and an explicit *"do not restate this number elsewhere — point here"*. The two copies this PR added (`drift-check.sh`, the drift-check block) and the `CLAUDE.md` "5 in 3 days" copy now point there instead.

### 🟢 F6 — `classic-zero` was a dead value

Branched on explicitly. The chain is now `enforce-admins` → `ruleset-nonbinding` → `classic-zero` → **`else` = a refusal, not a default** (`⚠ NO SPECIFIC REPAIR … it will not guess one`). Pinned two-way by `test_every_BP_OFF_KIND_assigned_is_branched_on_in_the_verdict`: assigned set == compared set, both with positive controls, plus the literal vocabulary written out in the test file.

---

## Verification

### Mutation sweep — 12 real mutants killed, 1 deliberate survivor

Fresh `git archive HEAD` tree per mutant (no `.git`), `PYTHONDONTWRITEBYTECODE=1`, each mutation asserted to apply exactly once.

|---|---|---|
| **F3** partial restore reads as agreement (the regression restored) | **KILLED** (2) | `test_a_PARTIAL_restore_under_a_declaration_is_rc25_not_agreement`, `test_the_enforce_admins_site_REACHES_the_verdict_under_a_declaration` |
| **F3** asymmetry inverted — ruleset-nonbinding fires too | **KILLED** (1) | `test_a_NONBINDING_RULESET_under_a_declaration_stays_in_the_agreement_cell` |
| **F6** `classic-zero` branch removed | **KILLED** (1) | `test_every_BP_OFF_KIND_assigned_is_branched_on_in_the_verdict` |
| **F6** `classic-zero` assignment removed — *the exact mutant that SURVIVED your sweep* | **KILLED** (4) | the ledger test + `test_a_KNOWN_off_kind_does_not_take_the_unrecognised_arm` + 2 pre-existing rc-24 tests |
| M1 declaration fails open | KILLED (30) | |
| M2 declared slug renamed | KILLED (12) | |
| M3 `note_rc 25` → `:` | KILLED (7) | |
| M4 COULD-NOT-MEASURE stops short-circuiting | KILLED (183) | |
| M5 enforce_admins site drops `BP_LIVE=off` | KILLED (4) | |
| M6 ruleset site drops `BP_LIVE=on` | KILLED (1) | |
| M7 rc 25 ranked below rc 10 | KILLED (2) | |
| PC positive control `note_rc 24`→`4` | KILLED (22) | |
| **NC** negative control: a no-op trailing space on the pass line | **SURVIVED**, as designed | — |

The negative control is there because a sweep that reports KILLED for everything is indistinguishable from one wired to the control tree. `CONTROL: 458 passed, 0 failed` **before and after**.

### Red-at-base / green-at-HEAD — base `bf5516fc`

**27 tests: 21 RED at base, 6 GREEN.** All 27 green at HEAD. Base tree = `git archive bf5516fc` with only the test file overlaid; one pytest invocation per test.

The **6 green-at-base are INVARIANT GUARDS**, labelled as such in their docstrings and not counted as regression coverage:

| invariant guard | what pre-existed |
|---|---|
| `test_a_gh_that_answers_nothing_is_still_could_not_measure_for_a_declared_repo` | the COULD-NOT-MEASURE discipline |
| `test_no_gh_binary_at_all_is_could_not_measure_for_a_declared_repo` | same |
| `test_the_declaration_cannot_be_set_from_the_environment` | there was no override at base either |
| `test_an_UNDECLARED_partial_restore_is_still_rc24` | pre-PR behaviour this must not weaken |
| `test_a_KNOWN_off_kind_does_not_take_the_unrecognised_arm` | see below |
| `test_DRIFT_PROTECT_SLUG_changes_the_SUBJECT_and_the_run_names_it` | the run always named its subject |

🔴 **One self-caught repeat of the F2 defect class.** `test_a_KNOWN_off_kind_does_not_take_the_unrecognised_arm` was first written as `test_an_unrecognised_OFF_kind_refuses_to_guess_a_repair` — a name claiming it covered the `else` arm. It does not: that arm is **unreachable by construction**, because the two-way ledger asserts every value `BP_OFF_KIND` can hold is branched on, so no fixture can drive it without mutating the script. Renamed to what it checks, with the unreachability and the guard that maintains it stated in the docstring. Reporting it because it is exactly the defect you flagged in F2, found by reading my own base matrix.

### Suites

- `test_drift_check.py`: **458 passed** (450 before this round, 431 pre-PR).
- Full dev-host pytest tier (`nix develop … -m pytest scripts/tests`): **11597 passed, 1 skipped, 0 failed** (20m48s), on `55d183dc`.
- `test_only_16_is_excused_from_failing_the_unit`: **green and untouched**. `git diff bf5516fc..HEAD -- scripts/tests/test_drift_check.py | grep '^-'` still yields **nothing** — the test-file diff is purely additive across both commits, including this round's two rewrites. `SuccessExitStatus = 16` unchanged.
- `<!-- merge-gate: other -->`: still exactly one marker, untouched.

I have not run `nix build` or `gate.sh` — the merged-tree gate is yours, and F3 changed behaviour.

ZacxDev:
## Audit ladder — round 1 claims

Round 1 was a first, full blind audit at `a9f0d6c2`. **No 🔴.** It could not break the state machine, the declaration's tamper-resistance, the consolidation, or the severity table, and it independently reproduced the red-at-base matrix exactly. It returned **three 🟡 and three 🟢**, all fixed below.

```audit-claims round=1 audited=a9f0d6c2..55d183dc
1. F3 (behaviour, a real regression) — a PARTIAL protection restore (required checks back, `enforce_admins=false`) landed in the declared-off cell and passed silently at rc 0 while printing "Restore protection and this becomes rc 25". Pre-PR that state fired rc 24. Fixed: `BP_OFF_KIND=enforce-admins` under a declaration now fires rc 25, with repair text naming both halves. Deliberately NOT extended to `ruleset-nonbinding` — argued in-line, with its known bound stated (a repo-level evaluate-mode ruleset is indistinguishable at that endpoint; `ruleset_source_type` would separate them and is not read).
2. F1 — the DND-bypass comment claimed "exited 24 on every 6-hourly fire — 5 failures in 3 days". Re-measured 2026-08-25..09-03: 32 unit failures in 9 days (rc 10 x13, rc 17 x11, rc 24 x5, rc 12 x2, rc 8 x1); the 5 rc-24 runs are one 24h span; co-presence checked individually (`local main is BEHIND` 5/5, `BUILT SOURCE … NOT current` 4/5). Both blocks now state plainly that this removes the STRUCTURALLY PERMANENT contributor and does NOT stop the toasts, naming rc 17/rc 10 as a separate open job.
3. F2 — "`DRIFT_PROTECT_SLUG` … cannot turn a finding off" was false: a slug-aware `gh` stub silenced the alarm by redirecting the subject. Reworded to the defensible claim — an env var CAN silence it, but the run PRINTS the slug it asked about, so the silence names a repo nobody chose, whereas an expectation override would leave the correct subject and change only the verdict. The old test is renamed to what it actually checked; two new tests assert the silencing rather than deny it.
4. F4 — `nix/home.nix` named the escalation ladders as "rc 13/16/18"; corrected to rc 13/18/23 with each env var named, and why rc 16 (no ladder, the single code on `SuccessExitStatus`) was wrong there. The two new blocks had disagreed with `drift-check.sh:563`, which was already correct.
5. F5 — "three orders of magnitude" was wrong (4/day vs 1-per-9-days is 36x, not 1000x) and this PR had added two more copies of it. The measured figure is now 32x, stated ONCE in the dunst block with "do not restate — point here"; all other sites point at it.
6. F6 — `BP_OFF_KIND=classic-zero` was a dead value that no branch matched; it was the SOLE SURVIVOR of the auditor's 15-mutant sweep. Now branched on explicitly, with an `else` that refuses rather than guesses.
7. SELF-CAUGHT — a test named `test_an_unrecognised_OFF_kind_refuses_to_guess_a_repair` claimed coverage of an `else` arm that is UNREACHABLE by construction (the two-way ledger asserts every `BP_OFF_KIND` is branched on). Found by reading its own red-at-base matrix (green at base); renamed with the unreachability and the guard that maintains it documented.
```

### Ledger — round 1

`round 1 · payload lines changed THIS round: 193 added / 54 removed across 3 payload files (drift-check.sh 113/25, nix/home.nix 61/19, CLAUDE.md 19/10) · scaffolding: test_drift_check.py 252/18`

Measured with `git log --numstat --format= --remerge-diff a9f0d6c2..55d183dc --not bf5516fc`, rc 0, non-empty range, git 2.55.0. **The attribution gate does not fire** — payload is substantial and F3 was a behaviour change, so round 2 is warranted rather than ceremonial.

### Verification of the fix round

- **Mutation: 12/12 real mutants KILLED**, including both F6 mutants — the assignment-removal one that survived round 1's sweep is now killed by 4 tests. A deliberate **negative control** (a no-op trailing space) **SURVIVED as designed**, because a sweep reporting KILLED for everything is indistinguishable from one wired to the wrong tree. `CONTROL: 458 passed, 0 failed` before *and* after.
- **Matrix, base `bf5516fc`: 27 tests, 21 RED / 6 GREEN**, all 27 green at HEAD; the 6 greens are labelled invariant guards in their docstrings.
- `test_only_16_is_excused_from_failing_the_unit` green and untouched; the test-file diff remains purely additive; `SuccessExitStatus` unchanged; exactly one `merge-gate` marker.

### Merged-tree gate — green on the FIXED tree

`main@2c6b2ac9` + `55d183dc` → merge commit `8fe99b7e`, clean merge:

| tier | verdict | counts |
|---|---|---|
| dev-host `gate.sh --tier both` | `GATE: RESULT=PASS exit=0` | `collected=20620 passed=20617 skipped=3 **failed=0**` · node 1449/1449 |

⚠ The **pre-fix** tree (`a9f0d6c2`) was separately green on all three tiers — dev-host, sandbox `pytests` (20612/0 failed) and sandbox `nodetests` (1449/1449), built one at a time. That run is **superseded** by the F3 behaviour change and is recorded only so that a red after the fix round would be attributable to the fix rather than to the merge. The sandbox tiers are being re-run against `8fe99b7e`.

### An omission on my side, recorded

This block should have been posted when round 1's fixes landed. It was not, and `audit-dispatch.py --round 2` correctly **refused to assemble a delta brief without it** — which is the guard working as designed, catching a missing intermediate block rather than silently widening the range.

ZacxDev:
## Round-2 audit fixes — `33bc930b`

---

### 🟡 R2-2 — rc 25 now covers the ruleset spelling of a half-gate

**I agree with your preference and extended it.** The reason the bypass-actors case was silent is that one OFF kind carried two states that are not the same thing:

| ruleset state | is it a gate? |
|---|---|
| `enforcement != active` | **no** — explicitly a non-enforcing dry run |
| `active` + `bypass_actors > 0` | **yes** — somebody built it; a listed actor can step around it |

Only the first was ever excluded on purpose. The second was excluded *by accident*, because `ruleset-nonbinding` collapsed both — and my discriminator for `enforce_admins` was "positive evidence that a gate was re-established on this repo", which an active ruleset listing a check plainly is. The script's own loop already called bypass actors *"the ruleset spelling of enforce_admins=false"*; the verdict block just wasn't reading it that way.

**What changed:**

- Split into **`ruleset-bypassed`** / **`ruleset-inactive`**. The old name became a lie once the states were separated — a bypassed ruleset is *also* non-binding, for some actors.
- The declared-OFF branch now turns on **one predicate**, `BP_RESTORED_GATE`, instead of an open-coded `!= enforce-admins`. That open-coding is precisely how the second spelling got missed; the predicate is defined once, above the branch, with both arms named.
- The rc-25 finding gets its own wording for this shape, naming the measured bypass count and the ruleset id, and a **both-halves** repair (remove the bypass actors *and* delete the declaration).
- The agreement cell's promise now names every state that fires and the one that does not.

**The remaining exclusion is only `ruleset-inactive`, and your `ruleset_source_type` finding is why it had to shrink.** Since the org-rollout objection is *closable, not structural*, it cannot carry an exclusion any wider than the case it actually describes. That is now written into the bound, with your `astral-sh/uv` measurement (`Organization` vs `Repository`) cited as the closing path.

**Test scaffolding**: `test_a_NONBINDING_RULESET_...` did generalise over a relationship while exercising one side. Renamed to `test_an_INACTIVE_ruleset_...` and scoped to what it checks, plus three new tests — declared/bypassed → rc 25, undeclared/bypassed → rc 24, and a bypassed ruleset **alongside an inactive sibling** (a non-enforcing sibling must not soften the finding; the verdict must come from the one that *is* a gate).

Also: `test_the_agreement_cell_no_longer_promises_something_false` pinned a single phrase, and that sentence has now been reworded twice. It now pins the **relationship** — every firing state named, plus the non-firing one — because a guard on one word is walkable by rewording.

### 🟡 R2-1 — the consolidation now actually happened

You were right that I removed the two copies I added and left the one I didn't. Removed `nix/home.nix:2991`, **and a fourth** I found on the re-sweep: my own restatement in the test file's block comment (`"5 failures in 3 days"`).

Re-run sweep, enumerated (`find … | xargs grep`, not `grep -r`, which honours `.gitignore` here):

```
$ find nix scripts claude CLAUDE.md -type f \( -name '*.nix' -o -name '*.sh' \
    -o -name '*.py' -o -name '*.md' \) -print0 \
  | xargs -0 grep -nE "orders of magnitude|failures in 3 days"
nix/home.nix:548:      #   "three orders of magnitude" an earlier wording claimed at three
```

**One occurrence: the owning block, calling it wrong.** The claim is now true rather than asserted.

### 🟢 R2-3 — `CLAUDE.md` fixed in both directions

Now names **evaluate/disabled rulesets** as the single uncovered state (with the closable-not-structural note), and says explicitly that a full restore followed by a later deletion is *not* uncovered, because deleting the declaration is part of restoring protection.

---

## Verification

### Mutation — 4/4 new mutants killed, negative control survived

|---|---|---|
| **R2** bypassed collapses back into inactive (*the regression restored*) | **KILLED** (4) | `..._BYPASS_ACTORS_under_a_declaration_is_rc25`, `..._ALONGSIDE_an_inactive_one_...`, `..._on_an_UNDECLARED_repo_...`, the kind ledger |
| **R2** `BP_RESTORED_GATE` drops the ruleset arm (*the exact original defect*) | **KILLED** (2) | `..._BYPASS_ACTORS_under_a_declaration_is_rc25`, `..._ALONGSIDE_an_inactive_one_...` |
| **R2** the argued exclusion inverted — inactive also fires | **KILLED** (1) | `test_an_INACTIVE_ruleset_under_a_declaration_stays_in_the_agreement_cell` |
| **R2** the loop never records `BP_RS_BYPASSED` | **KILLED** (3) | the three bypass tests |
| **NC** no-op trailing space on the pass line | **SURVIVED**, as designed | — |

`CONTROL: 461 passed, 0 failed` before **and** after. Each mutant is killed by the test that owns its arm, not a neighbour — the inactive-exclusion mutant is killed *only* by the inactive test, and the bypass mutants *only* by the bypass tests.

### Red-at-base / green-at-HEAD — base `bf5516fc`

**30 tests: 24 RED at base, 6 GREEN.** All 30 green at HEAD. Base tree is `git archive bf5516fc` with only the test file overlaid; one pytest invocation per test.

The **6 green-at-base are INVARIANT GUARDS**, labelled in their docstrings and not counted as regression coverage: the two COULD-NOT-MEASURE-under-a-declaration cases, the env-override refusal, `test_an_UNDECLARED_partial_restore_is_still_rc24`, `test_a_KNOWN_off_kind_does_not_take_the_unrecognised_arm`, and `test_DRIFT_PROTECT_SLUG_changes_the_SUBJECT_and_the_run_names_it`.

⚠ **One honest qualification.** `test_an_ACTIVE_bypassed_ruleset_on_an_UNDECLARED_repo_is_still_rc24` is RED at base, but **its `rc == 24` assertion is invariant** — that behaviour is unchanged; it goes red at base only on the new *"NOTHING DECLARES"* wording. So it is regression coverage for the finding text and an invariant guard for the exit code. Counting it wholesale as regression coverage would overstate it.

### Suites

- `test_drift_check.py`: **461 passed** (458 before this round, 450, 431 pre-PR).
- Full dev-host pytest tier: **11600 passed, 1 skipped, 0 failed** (19m31s), on `33bc930b`. The only change after that run is `62fcb75b`, a comment-only commit; `test_drift_check.py` re-run on it is **461 passed**.
- `test_only_16_is_excused_from_failing_the_unit` green and untouched; `SuccessExitStatus` unchanged. Test-file diff against base is still **purely additive — 0 deleted lines** across all four commits. Exactly one `merge-gate` marker.

### Live, against the real remote

`rc=0`, the declared-off report, and the corrected promise:

```
[protect] DECLARED OFF, and live OFF — expectation and reality agree. No rc is set.
[protect]   🔴 The alarm is QUIET, not DISABLED. It fires as rc 25 as soon as a gate
[protect]   is re-established on this repo — a full restore, a PARTIAL one that puts
[protect]   the required checks back and omits enforce_admins, or an ACTIVE ruleset
[protect]   listing a check but carrying bypass actors (the ruleset spelling of the
[protect]   same half-gate).
[protect]   ⚠ It stays quiet for a ruleset parked in evaluate/disabled mode — that is
[protect]   not a gate anybody built.
```

---

### On the two you deliberately deferred

- **`BP_LIVE=unknown` when the protection endpoint is unreadable with `BP_N>0`** — I agree it does not belong in this PR. It is byte-identical at `bf5516fc`, so nothing here caused or worsened it, and fixing it means deciding what a push-scoped token *should* yield, which is a separate argument with its own tests.
- **`bp_declared_off_reason`'s contract now reads "off, except the re-established spellings of off"** — took the one-line note (`62fcb75b`). The declaration function now says explicitly that a repo whose half-gate is *intentional* would be permanently rc 25, and that the two-way ledger is what forces that through a human diff where the note sits one line above the enumeration.

I have not run `nix build` or `gate.sh` — R2-2 changed behaviour, so all three merged-tree tiers are worth your re-run.

PR #1255, opened 2026-09-03
feat(cairn): a two-way ledger over store-root resolvers, cairn doctor, and the skill

ZacxDev:
## What this is

Three things, in order of what matters:

1. **`scripts/tests/test_store_root_ledger.py`** — a two-way ledger over every file
   that resolves a subsystem-store root. This is the deliverable.
2. **`cairn doctor`** — one call for what a reader currently checks by hand.
3. **`claude/skills/cairn/`** — tier B, a router to those commands.

## 1. The ledger guard

The same defect shipped **three times**: a reader open-coding the FROZEN pre-cutover
mirror instead of asking `subsystem_read_store` where to read — `subsystem_recall`'s
CLI, `service_recon`, `subsystem-audit.py` — each found by a human noticing, one at a
time, after the previous one was fixed.

Two ledgers, each failing when the set **grows** *or* **shrinks**:

| ledger | population | the direction that matters |
|---|---|---|
| `ROUTED` | files that CALL `read_store_root` / `resolve_read_store` | **SHRINKS** — a reader that stops routing is the regression, and it takes its own per-file tests with it |
| `SITED` | files that compute a store-root path themselves, each with a written reason | **GROWS** — the fourth open-coded reader, on the commit that introduces it |

A third guard, `TestTheScannerVocabularyMatchesTheRealPaths`, pins the scanners'
vocabulary two-way against `DEFAULT_CACHE_ROOT` and `DEFAULT_STORE_ROOT`. Without it,
renaming the cache directory would leave both ledgers green over a scan that can no
longer see the thing they police.

### Red at base

The scanner cannot be *run* at the pre-change base — the module it pins did not exist.
So it was graded against the tree where the defects were live:

| tree | `subsystem_recall.py` | `service_recon.py` | `subsystem-audit.py` |
|---|---|---|---|
| `b9b2493d` (parent of #1233, pre-resolver) | **SITED**, not routed | **SITED**, not routed | **SITED**, not routed |
| this branch | ROUTED, not sited | ROUTED, not sited | ROUTED, not sited |

### Mutation sweep

Isolated `cp -a` copy with `.git` removed and re-inited, `PYTHONDONTWRITEBYTECODE=1`,
`__pycache__` cleared between mutants.

| mutant | outcome | killed by |
|---|---|---|
| an unledgered reader sites the frozen mirror | **KILLED** | `TestTheSitedLedgerIsTwoWay::test_the_ledger_is_exactly_the_scan`, NEW siting naming the planted file |
| a ledgered router stops routing | **KILLED** | `TestTheRoutedLedgerIsTwoWay::test_the_ledger_is_exactly_the_scan`, LOST routers naming the file |
| a `SITED` row is deleted | **KILLED** | the SITED ledger, NEW siting naming the un-ledgered file |
| the scanner's vocabulary drifts from a live root | **KILLED** | `TestTheScannerVocabularyMatchesTheRealPaths` |
| **negative control** — a no-op comment | **SURVIVED** (correct) | — |

Controls: 49 passed before the sweep and 49 after. No survivors other than the
negative control.

### Exemptions, and the two judgement calls

* **`cairn-cutover.py` — EXEMPT.** The mirror is its *subject*, not a store it reads
  for an answer: it plans the delta from that tree, pushes it, and P5 chmods it
  `0444`; `--unfreeze` restores the recorded modes on the same tree. Pointed at the
  synced cache it would plan an empty delta (a copy of the pod, against the pod) and
  then freeze a directory the next `cairn sync` deletes and recreates.
* **`backup.py` — EXEMPT, with a recorded open question.** It bundles the mirror's
  per-scope **git history**, which exists nowhere else; the synced cache is an
  extracted tar with no repositories in it, so it cannot be repointed and still do its
  job. But the pod is canonical now and has its own backup CronJob, so this is an
  archive of a frozen tree rather than the store's backup — and its docstring still
  says a disk failure *"loses the whole thing permanently"*, which is no longer true.
  Recorded in the ledger rather than fixed here.
* **`present/measure.py` — NEW, and the closest thing to a fourth instance.**
  `Env.live()` resolves the frozen mirror for the explainer page's store measurements,
  and its own provenance string still reads *"read via `subsystem_recall.load_store()`"*.
  Post-cutover it measures a frozen tree and presents the numbers as the store's — the
  same shape as the three regressions, in a page rather than a recall. Ledgered with
  the question open; **not fixed here**, because the fix is a judgement about what the
  page claims, not a repoint.
* Also ledgered: `subsystem_touch.py` (the writer's target), `escrow-verify.py` and
  `restore-verify.py` (transitive `B.DEFAULT_STORE` re-exports — invisible to a
  component scan, which is why the ledger looks for the names too), `commit.sh` (shell),
  `server.py` (pod-side `/data`), and `cairn` itself (its one sited path is the
  credential file).

## 2. `cairn doctor`

Seven checks in one call, each in one of four states — and the fourth state is the
point:

* `OK` / `PROBLEM` — measured.
* `UNMEASURED` — it could not look, and the reason is mandatory (a `Check` with an
  empty detail is refused at construction).
* `NOT-OBSERVABLE` — no client can answer it at all. It contributes **nothing** to the
  exit code, deliberately: the pod exposes no identity route, so folding it into
  `UNMEASURED` would make `doctor` non-zero on every healthy run forever.

Unauthorised is separated from unreachable **structurally** — `StoreUnreachable` now
carries `http_status`, set from the `HTTPError`, so nothing greps an f-string. The read
path is unchanged; `resolve_state` reads neither.

It **installs nothing** — it fetches, reads headers and the member list, and discards
the bytes, so it can be run twice without repairing the staleness it was run to
measure. Exit codes are documented by the command itself (`EXIT_LEGEND`, printed on
every run and in `--help`), never by a skill.

**It found two real problems on the workbench the moment it existed**, both reported
in the run, neither fixed here:

* `frozen-mirror` **PROBLEM** — 6 of 160 entry files under the "frozen" mirror are
  still mode 644. A write to one of those lives on one host and the pod never sees it.
* `token-scopes` **PROBLEM** — two scopes present on this disk are absent from this
  token's snapshot. The API cannot say which of the two readings that is (a refused
  scope is byte-identical to an absent one, deliberately), so the report names both
  and guesses neither.

## 3. The skill

`claude/skills/cairn/SKILL.md` (4,290 B) + `reference/operator-surface.md` (6,000 B).
The body is a router: a command table, the two different exit 4s side by side, where a
host reads from, and — for a new consumer — a pointer at the ledger guard rather than a
paragraph of advice. The operator detail (the pod, `seed.sh`'s overwrite semantics,
`verify-byte-identity.sh`'s measured one-directional coverage gap, `build-push.sh`,
the freeze, and the once-at-startup token load with its `exit 78` / `kubectl delete pod`
consequences) is in `reference/`, which costs nothing until loaded.

**Tier B** in `claude/skill-tiers.json` — the tool is named, never described.

### Paying for the listing entry

`test_skill_descriptions.py` had **12 chars** of headroom, so a tier-B skill is not
free there (that gate does not read the tier ledger). Eviction, per its own playbook
step 1 — mechanism prose out of the costliest descriptions, no trigger phrase or
disambiguation clause dropped:

| skill | change | Δ |
|---|---|---|
| `signal` | dropped the pipeline parenthetical | −71 |
| `mailbox` | dropped the pipeline parenthetical (`inbox.zacx.dev` survives as a trigger) | −53 |
| `session-manager` | dropped the gloss on UNSENT PROMPTS | −42 |
| `prune-index` | trimmed; it no longer routes on the frozen-mirror path, which `cairn` claims | −148 |
| `cairn` | new | +308 |

Listing total **12,911 / 12,929** — 18 chars of headroom. `TIER_A_CEILING_CHARS`
**lowered 9,200 → 8,935**, because the tier-A cut would otherwise have been banked as
519 chars of slack, more than a whole average entry.

## Departures from the brief, reported rather than implemented

1. **`prune-index` and `subsystem-index` have no byte ceilings.**
   `git grep -l MIN_HEADROOM_BYTES scripts/` names `prune-skill`, `handoff`,
   `session-manager`, `browser-bridge` and `RULES.md` — not these two. There was no
   ceiling to re-check.
2. **The absorption is smaller than "absorb the duplicated prose".** Roughly twenty
   literal sentences in those two files are pinned by four test modules, and
   `subsystem-index`'s cutover header is inside a whole-paragraph **hash** pin
   (`TestTheOwnersMechanismProseIsPinnedWHOLE`) that exists precisely so the owner's
   protocol prose cannot be quietly reworded. That prose is the owner stating its own
   mechanism, not cairn documentation misfiled — so it stays, and each door gains a
   pointer instead. Byte deltas: `prune-index` **13,884 → 13,729 (−155)`;
   `subsystem-index` **37,204 → 37,469 (+265)**, a growth, because the pointer was
   added and nothing pinned was removed.
3. **The measured population was incomplete.** All the line numbers in the brief were
   right (`present/measure.py:381` is the last of a three-line resolution). Missing:
   `escrow-verify.py`, `restore-verify.py`, `commit.sh`, `present/measure.py`,
   `server.py`. `build-push.sh` names no store path at all. `seed.sh` and
   `verify-byte-identity.sh` require `--store` with no default and are correctly out.
4. **The pod exposes no identity route** (`API_ROUTES` = recall/search/snapshot), so
   "token identity and which scopes it may write" is answered in two parts: the
   credential's **fingerprint** (`sha256[:12]`, the same handle the pod's audit log
   carries — pinned against `server.token_id` through a literal digest), and the scopes
   it can actually reach, derived from the snapshot. The declared identity string is
   `NOT-OBSERVABLE`, with the exact operator command in the detail. `doctor` never
   reads a secret and never shells `kubectl`.

## Not done here

`nix/home.nix` and `CLAUDE.md` are untouched — #1250 is mid-gate on both. The skills
tree deploys via `cp -R ${../claude/skills}`, so the new skill needs no nix change, and
the opencode commands are generated at build time from the same tree.

---

### DIFF
--- a/claude/skill-tiers.json
+++ b/claude/skill-tiers.json
@@ -53,6 +53,10 @@
     },
     "bar": { "tier": "A" },
     "browser": { "tier": "A" },
+    "cairn": {
+      "tier": "B",
+      "why": "the tool is NAMED, never described -- 'cairn doctor', 'cairn sync', 'cairn who'. The symptoms that could route here already belong to skills that name it in their own bodies: a refused read to `resume`, a stale index block to `analyze-service`, an entry that got huge to `prune-index`"
+    },
     "check-clickup-addressed": { "tier": "A" },
     "clawgate": { "tier": "A" },
     "clickup": { "tier": "A" },
--- a/claude/skills/activity/SKILL.md
+++ b/claude/skills/activity/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: activity
-description: "Operate the personal activity-telemetry pipeline: 9 sources -> collector -> homelab ClickHouse activity.events -> Grafana + a per-source deadman. Query it, revive a stalled or DEAD source, deploy, validate. Use for: activity tracking, the keylogger, \"where my time goes\", the activity dashboard, activity.events, the collector, a stale telemetry source, the `tlm` bar pill."
+description: "Operate the personal activity-telemetry pipeline: homelab ClickHouse activity.events -> Grafana. Query it, revive a stalled or DEAD source, deploy, validate. Use for: activity tracking, the keylogger, \"where my time goes\", the activity dashboard, activity.events, the collector, a stale telemetry source, the `tlm` bar pill."
 ---
 
 # activity-telemetry operations
--- a/claude/skills/cairn/SKILL.md
+++ b/claude/skills/cairn/SKILL.md
@@ -1,0 +1,85 @@
+---
+name: cairn
+description: "The hosted subsystem store (pod in ns `subsystem-store`) and its client `scripts/cairn`. Use for: `cairn doctor`, cairn sync/recall/search/ls-entries/who, a stale or unstamped store, a `cairn` exit 4, seeding the pod, a scope a token cannot reach. Writes are `subsystem-index`; pruning is `prune-index`."
+allowed-tools: Bash, Read, Grep, Glob
+---
+
+# cairn — the hosted subsystem store, and the client that reads it
+
+**This file is a router. Anything a command can answer, a command answers** —
+`claude/RULES.md`, "prefer deterministic/structural fixes over prose".
+
+## Start here, always
+
+```
+cairn doctor            # pod, credential, cache stamp, counts, scope visibility, reader resolution
+cairn doctor --json     # the same facts, machine-readable
+cairn doctor --no-sync  # same, without touching the network
+```
+
+It prints its own exit-code legend and a state per check. Four states, and the
+last two are not the same fact: `OK` · `PROBLEM` (measured, and it is not fine) ·
+`UNMEASURED` (it could not look, and says why) · `NOT-OBSERVABLE` (no client can
+answer it; the detail names who can). **It installs nothing** — it fetches, reads
+headers, and throws the bytes away, so it can be run twice without repairing the
+staleness it was run to measure.
+
+## The verbs
+
+| you want | run |
+|---|---|
+| refresh the local cache from the pod | `cairn sync` |
+| this repo's digest | `cairn recall` (`--scope X` / `--repo PATH`) |
+| find a hunk by text | `cairn search '<query>'` (`--all-scopes` to search every scope) |
+| what does the cache actually hold | `cairn ls-entries` |
+| parse-check the cached entries | `cairn validate` |
+| which sessions/windows/transcripts worked a task | `cairn who <task>` (`--json`, `--host`, `--no-windows`) |
+| diagnose anything above going wrong | `cairn doctor` |
+
+`cairn who` is about a **task**, not a store entry: it touches no store and never
+syncs, and it has its own longer `--timeout` because it shells into tmux on two
+hosts.
+
+**Writes are not this skill's.** `subsystem-index` owns the one protocol for
+every writer; `prune-index` owns deletion, with its own confirmation gate. Load
+whichever applies rather than reconstructing their steps here.
+
+## 🔴 The two different exit 4s
+
+Confusing them sends you to re-run the command that just failed.
+
+| whose 4 | means | do |
+|---|---|---|
+| `cairn sync`'s `EXIT_REFRESH_FAILED` | the store was **not reached**, but a usable cache survived | nothing to re-run — read the banner; the cache still serves reads |
+| the **reader's** `EXIT_UNSTAMPED_READ_STORE` | the store it resolved carries no `.sync-stamp`, so it cannot date itself and refuses | `cairn sync` |
+
+`claude/skills/resume/SKILL.md` step 4 carries the long form, including why
+`cairn recall` never returns 4 (it reaches the reader as a *library*). Read it
+there; it is not restated here.
+
+## Where a host reads from
+
+`scripts/lib/subsystem_read_store.py` is the ONE answer, and
+`cairn doctor`'s `reader-resolution` check prints it. Two directories exist and
+they are not interchangeable: `~/.cache/subsystem-store` is the synced
+read-through cache, stamped by `cairn sync`; `~/.claude/analyze-service-index`
+is the pre-cutover per-host mirror, frozen and refreshed by nothing.
+**The discriminator is the stamp, not the path** — a store that cannot date
+itself is refused rather than served.
+
+## Adding a consumer that reads the store
+
+Call `subsystem_read_store.resolve_read_store()`. Do **not** compute a path.
+`scripts/tests/test_store_root_ledger.py` enumerates every file that resolves a
+store root and requires each to route through that module or carry a written
+exemption; it fails when the set grows **or** shrinks. That test — not this
+paragraph — is what stops the fourth open-coded reader, and its ledger reasons
+are where the exemptions are argued. Read them before adding a row.
+
+## Operator surface
+
+`reference/operator-surface.md` — the pod, `seed.sh`, `verify-byte-identity.sh`
+and its measured coverage gap, `build-push.sh`, `cairn-cutover.py`, the backup
+CronJob, and 🔴 **the token file is read ONCE at startup**: editing the secret
+changes nothing until the pod is replaced, and the wrong replacement command
+costs two rollouts. Load it before touching the deployment.
--- a/claude/skills/cairn/SKILL.md
+++ b/claude/skills/cairn/SKILL.md
@@ -44,6 +44,15 @@ hosts.
 every writer; `prune-index` owns deletion, with its own confirmation gate. Load
 whichever applies rather than reconstructing their steps here.
 
+🔴 **There is no CREATE route, and the failure does not say so.** `PUT` demands
+an `If-Match` the caller derives from bytes that must already exist (`428` with
+none, and `If-Match: *` is refused because it is no precondition at all), and
+`POST …/bullets` appends to something. A ref the pod has never held answers
+`404 ref-unknown` — the same bytes as a ref outside your allowlist. So a scope's
+FIRST record cannot be made through the API at all; it is an operator step, and
+until somebody takes it the record does not exist anywhere the pod can serve.
+Measured 2026-09-03, on real entries stranded by exactly this.
+
 ## 🔴 The two different exit 4s
 
 Confusing them sends you to re-run the command that just failed.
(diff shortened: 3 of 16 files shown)


ZacxDev:
**What was run, and what was not** — base `14ace732`, head `9c37f6ed`.

Dev-host pytest tier only, 16 modules covering the blast radius: `test_cairn_doctor`, `test_store_root_ledger`, `test_cairn_{cli,who,write,cutover}`, `test_subsystem_{read_store,touch,recall,audit}`, `test_index_append_protocol`, `test_service_recon`, `test_skill_{tiers,descriptions,audit}`, `test_present_measure` — **2,263 passed**.

**NOT run here, deliberately:** `nix build .#checks…`, `scripts/gate.sh`, and the whole `scripts/tests` directory. The merged-tree gate is yours; a green dev-host run is a claim about one tier, one tree and one base, and I am naming all three rather than implying more.

Two later commits after the PR body was written, both fixing the same class the PR is about, found by re-reading the module against its own docstring:
- `dd8411ca` — absent-vs-unreadable was decided by `reason.endswith("does not exist")`. A guard spelled rather than structural: reword the message and an UNREADABLE mirror reports OK.
- `9c37f6ed` — `token-scopes` returned OK with the hole stated in its tail when a local root could not be read, and the `pod` UNMEASURED heading said "no answer from the store" over a reason that was an answer.

ZacxDev:
## Round 1 addressed — all 10, plus a defect in my own sweep harness

Head `d54721b3`. Corrections to the PR body are at the bottom.

### 🔴 1 — two-tier enumerator
`tracked_sources()` follows `testlib/public_ip_scan.py::repo_files`: git when `.git` exists, filesystem walk otherwise, both filtered to the same set. **No skip** — a skip would make both ledgers vacuous in the tier that gates, which is worse than the crash.

Verified with a real no-`.git` simulation (`cp -a` the tree, remove `.git`), carrying its own positive control that `git ls-files` genuinely exits **128** there: **152 passed**.

### 🟡 2 — the false row, fixed structurally
You were right that the row was the defect and not the code: the resolver deliberately does not name the mirror, so `cmd_doctor` taking `subsystem_touch.DEFAULT_STORE_ROOT` **is** taking the one definition. Row corrected to name both hits.

But a corrected sentence would rot the same way, so the class is closed: **`SELF_RESOLVED_KINDS`** pins which KINDS each ledgered file performs, line numbers stripped. A new kind inside an already-ledgered file now fails even though the file set does not move — mutant M11 confirms.

### 🟡 3 — four survivors caught, and the widening's own over-fire fixed
All four are now caught and kept as parametrized fixtures. ⚠ My first widening (substring) **over-fired** onto `subsystem-store-api`, `analyze-service-index-backups`, `subsystem-store-backup` and the `subsystem-store-client/1` User-Agent, and pulled `build-push.sh` in as a phantom row. The rule is now whole path **segments** (with shell punctuation stripped from segment ends — `commit.sh`'s `${POSITIONAL[0]:-…}` leaves a trailing brace, which segment equality alone missed). Pinned by 14 literal cases where every FALSE shares a prefix with a TRUE one.

Residuals rewritten as measured: `str.join` corrected to CAUGHT, and the `scripts/`-only scope plus `nix/home.nix` now lead the list. I did **not** build a `.nix` arm — named as a gap instead.

### 🟡 4, 5, 6 · 🟢 7, 8 — all as directed
SKILL.md now carries the module's caveats. `PodFacts.__post_init__` enforces the reason. `token-scopes` tags each scope with its tree — live, both are `[mirror]` and get the leftover reading spelled out. `_describe` re-checks `root.exists()`. `_doctor_epilog` fails soft and its comment no longer claims laziness it does not have.

### 🔴 10 — you were right, and the justification was wrong too
Both phrases restored; measured, each appears in **zero** other descriptions. The 148 chars re-taken from architecture prose: `initiatives` (component list), `window-triage` (address examples — "Gold" survives in its own Use-for clause, only `scratch2:@81` went), `activity` (pipeline arrows). Listing **12,892/12,929**, headroom 37. `TIER_A_CEILING_CHARS` lowered again **8,935 → 8,821**.

### 🔴 A defect in my own mutation harness, found by its negative control
The first re-run reported the **negative control KILLED**. Cause: `snap()` wrote `<file>.orig` *inside* the tree, so the walk tier saw an untracked file the git tier did not — the sweep was measuring itself. Harness fixed, and the test relaxed to what actually matters: the walk must not be **blind** to anything git tracks, and no untracked file may resolve a store root. A stray scratch file is not a defect in this ledger.

### Sweep, re-run: 13 mutants
| mutant | result |
|---|---|
| unledgered reader / router stops routing / SITED row deleted | KILLED (set ledger) |
| **new KIND in an already-ledgered file** | KILLED (kind ledger) — the #2 class |
| **segment rule → substring** | KILLED (8 tests) |
| **alias-import arm removed** | KILLED (survivor fixtures) |
| **walk skip-list emptied** | KILLED (no-git enumerator) |
| **PodFacts reason unenforced** / **absent inferred from exception** | KILLED |
| scanner vocabulary drift / `_HASHBANG` ordinal / shebang literal returns | KILLED |
| **no-op comment (negative control)** | **SURVIVED** ✓ |

Controls **152 passed before and after**. No unexpected survivors.

### Local runs
- `test_runtime_shebangs.py` ✅ — and both old lines still fire under the same scanner, so the guard was right and is still capable.
- no-`.git` sandbox simulation ✅ 152 passed, with its 128 positive control.
- repo-wide structural guards (20 modules incl. `test_store_siting_ledger`, `test_repo_path_defaults`, `test_conditional_skip_pins`, `test_hermetic_git`, the four content gates): **743 passed**.
- topical blast radius (15 modules): **2,855 passed**.

## Corrections to the PR body
- **Freeze numbers are stale and now fixed** in `test_cairn_doctor.py`: **7** files at 644 (not 6), an append watched to succeed, six entries stranded. Live now: 159 files all `0444`, `frozen-mirror OK`, `cache-vs-pod 205/205`.
- **Ledger population is ROUTED 5, SITED 10** — the body's "9 exemptions" undercounted by listing `subsystem_read_store.py` separately as THE OWNER.
- The eviction table's "no trigger phrase dropped" was **false**; see 🔴 10.

## One thing you did not flag
`scripts/tests/test_store_siting_ledger.py` already existed (`b4fde334`) and I did not find it — my population-sweep subagent was asked exactly that question and I proceeded without its answer. **Not a duplicate**: theirs ledgers TEST files that stand up the store server (tmpfs vs disk); mine ledgers PRODUCTION files that resolve a store root, and explicitly excludes `scripts/tests/`. Zero overlap, both verified.

But the names collide badly — `test_store_root_ledger` vs `test_store_siting_ledger`. I renamed my predicate `sites_a_store_root` → `resolves_a_store_root` and kept `SITED` as the dict name since your report uses that word. **Your call** whether the module itself should be renamed; I did not touch the sibling.

### `cairn put` cannot create — added
Verified in `server.py` before writing it: `If-Match` is mandatory (428 without), `If-Match: *` refused, and the handler resolves an existing entry. The skill now says a scope's first record cannot be made through the API at all and that it is an operator step.

PR #1256, opened 2026-09-03
docs(handoff): #1219 moved every figure rank 13 measured — 9 h after rank 13 merged

ZacxDev:
End-of-session handoff update for the skill-chain-usage-audit arc. One file, +64/−13, written by `handoff_doc.py` (merge classified `State now → REPLACE · Open investigations → APPEND · Next steps → REPLACE`).

## What changed since the doc was written

**Rank 13 merged** as #1220 `aa1c2abf`, verified by content (blob-identical on `origin/main`, never by `--is-ancestor` — squash makes that permanently false).

**Then #1219 merged (`b4fde334`, 2026-09-02T16:48:50Z) and moved every figure rank 13 had measured — 9 hours after rank 13 landed.** That is the arc's signature failure recurring on its own output, which is why it is recorded rather than quietly corrected.

Re-derived with my own `ast.walk` over `With`/`AsyncWith` items on current `origin/main` — not the doc's numbers, not #1219's:

| | before #1219 | after |
|---|---|---|
| `with running(...)` sites | 256 | **270** |
| `store` | 133, disk | **135, SITED** |
| `scoped_store` | 110, disk | **110, SITED** |
| unfixtured | 13, disk | **25, still disk** |
| **disk-backed total** | **123 of 256** | **25 of 270** |

All 25 checked individually: 17 bind `tmp_path` directly, 8 arrive via `shuffled_pair`/`ambiguous_pair` which are themselves `tmp_path`-rooted — **0 sited**. Both cairn suites now carry `store_siting`, and `scripts/testlib/store_siting.py` now exists.

Two traps recorded so the next reader does not walk into them:
- **The 25 are not the 13 grown by attrition** — the *file* grew, 256 → 270 sites. The two counts are not differenceable.
- **#1219's own ratchet `_DISK_ROOTED_SITES = 33`** counts store-root *constructions*, a different unit from `with running(...)` call sites. Equating them would be this arc's certify-by-resemblance bug.

## The consequence for rank 13's closing condition

Rank 13's condition (a) is **still UNMET** — `scripts/ci-repro/README.md` returns 0 for all four greps, with the positive control `grep -c fsync` = 25, so those zeros are readings and not a dead probe.

But it is now **REVISED and smaller**: #1219 sited `scoped_store` and both cairn suites, so `a2`/`a3` are **obsolete and must not be required** — demanding those words would force a false sentence into the README. `a1` + `a4` survive. The doc predicted this in writing *before* #1219 merged ("re-read a2/a3 after #1219 merges"), which is the one thing that went right here.

Rank 10 is likewise superseded: the remaining exposure is ~5× smaller and entirely unfixtured, though the tmpfs probe is **still unverified on `talos-xr6-r7p`**, so the contention is still not established as gone.

## Notes

- **No `clawgate-task:` field.** `clawgate_handoff.sh resolve` exited **5** — 0 tasks. An unknown session id answers 200 with an empty array, so that cannot distinguish "touched no task" from "wrong id". Not a clean bill of health; no field written, no task created.
- The `DROPS 1 durable line` warning was read and explained: line 33 is not dropped, it was **re-wrapped** — the `All verified by CONTENT` text is carried forward on its own line.
- `forcing: none` on 12 of 13 items is accurate — 9 are ✅ DONE and the open ones are not externally forced. Rank 13 carries `forcing: regression`.
- Subsystem index: PR #1220's only file is the handoff doc itself, so the `--pr` window is legitimately empty after `--exclude`. The durable lesson from this session went to `devrc/tests` instead (appended, validated rc 0).

### DIFF
--- a/claude/skills/audit-pr/SKILL.md
+++ b/claude/skills/audit-pr/SKILL.md
@@ -26,6 +26,19 @@ generates the range, the cross-repo worktree directive, checkout state and toolc
 prior round's claims from the fenced `audit-claims` block ONLY, and carries the invariant clauses
 verbatim. **A delta round with no parseable block is REFUSED.**
 
+🔴 **PASS `--repo <owner>/<name>` WHENEVER THE PR IS NOT IN YOUR CWD'S REPO — without it the PR
+NUMBER IS RESOLVED AGAINST THE CWD'S REPO, SILENTLY.** This is a dispatch hub: sessions routinely
+sit in one repo and audit a PR in another, and PR numbers collide across repos by construction.
+MEASURED 2026-09-01: `audit-dispatch.py 1220` run from `civitai/talos-infra` for
+`innovation-upstream/devrc#1220` assembled a **complete, well-formed, confident brief for
+talos-infra#1220** — a real, unrelated mongo-alerts PR. Nothing errored; the only tell is the
+repo name on the brief's own title line. An auditor handed that brief audits the wrong PR and
+reports findings about work you never touched. **Read the title line of every brief before
+dispatching it.** With `--repo` the script also emits the CROSS-REPO section that tells the agent
+**not** to use `isolation: "worktree"` (which would worktree the cwd's repo — the same class of
+error one level down); without it, that section is silently absent because the script believes the
+PR is local.
+
 🔴 **That refusal covers NO block at all — it does NOT cover a MISSING INTERMEDIATE one, which
 proceeds and SILENTLY WIDENS the range.** The anchor is the newest block the script can PARSE, not
 the previous round's, so when round N−1 posted nothing, round N anchors on round N−2's tip and the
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Added a second commit: `5d6c73a1` — the `audit-dispatch.py --repo` trap

Flagging rather than folding it in silently, since it is a different file from the handoff doc this PR is named for. Both are documentation, in the same repo, and this lesson came out of this same arc — but say the word and I will split it into its own PR.

**`claude/skills/audit-pr/SKILL.md`, +13/−0.** Measured this session: `audit-dispatch.py 1220` run from `civitai/talos-infra` while intending `innovation-upstream/devrc#1220` assembled a **complete, well-formed brief for `talos-infra#1220`** — a real, unrelated mongo-alerts PR. Nothing errored. The only tell is the repo name on the brief's own title line.

The second-order effect is what makes it worse than a wrong title: with `--repo`, the script emits the CROSS-REPO section telling the agent **not** to use `isolation: "worktree"` (which would worktree the cwd's repo — the same class of error one level down). Without `--repo` that section is silently absent, because the script believes the PR is local. **The omission removes the very warning that would have caught its own consequence.**

Verification: `doc-path-rot` 77 passed — and unlike the handoff-doc case, that green is *meaningful* here, because `CORPUS_DIRS = ("claude", "CLAUDE.md")` so `claude/skills/` is genuinely scanned. `test_handoff_skill_size.py` + `test_skill_tiers.py`: 50 passed; no size ratchet exists for `audit-pr` specifically.

PR #1274, opened 2026-09-04
feat(main-green): a deadman that reports a RED main, because nothing else does (rank 22)

ZacxDev:
Closes rank 22 of `claudedocs/handoff-cairn-phase3.md`. Claim `cairn-phase3-22`.

## The gap

Branch protection is OFF by a live operator decision, so a commit can land straight on `main` with nothing running against it. **MEASURED 2026-09-03: that happened twice in one session** — `a720d30d` and `a451abc0`, both one-line direct pushes titled `espanso`. `main` was red for hours each time, and **both times the only detector was a human running the gate by hand, incidentally.**

`drift-check` answers *"is a host still receiving changes?"*. This answers the independent question *"is what they are receiving any good?"*.

## What it does

On a 4h timer: fetch, and **if the mainline tip has moved since the last recorded verdict**, run the real suite against it.

| rc | meaning | systemd |
|---|---|---|
| 0 | GREEN — passed / already-verified / red-did-not-reproduce. Each prints **which**. | success |
| 10 | **RED, REPRODUCED** — ran twice, failed twice | fails → toast |
| 11 | COULD NOT MEASURE — quiet, recorded, laddered | success |
| 12 | **BLIND** — 6 consecutive unmeasured runs (~24h) | fails → toast |

## Three design calls, each from a measurement rather than a preference

**It runs the hermetic tier, not `scripts/gate.sh`.** On one tree (main + #1261) the dev-host tier reported **4 failures** and the nix sandbox tier reported **0**. All four were `subprocess.TimeoutExpired` at a fixed 300 s cap while the box sat at load 22–29; an *untouched* target inflated 4.49 s → 177.26 s in the same window. Wiring a DND-bypassing toast to the tier that emits those would build the permanently-red gate `RULES.md` calls worse than no gate.

**rc 11 is a systemd success and rc 12 is not.** Making 11 quiet stops a network blip toasting — and, alone, lets the deadman go blind forever in silence, which is the same shape as the bug it exists to catch. `drift-check`'s rc 18 is the precedent (*"setting no code was right per run and wrong forever"*), so the ladder is copied rather than reinvented. One measured run resets it.

**It runs the whole suite.** The second break was a **guard whose premise died**, not a regression — the collision it asserted was gone at the source. A distilled *"did attribution regress?"* check would have sailed past it.

⚠ Rank 22's own proposed fix — trigger the espanso test when `nix/home.nix` changes — is **void if #1265 lands**, since that deletes all nine live-config tests and `_live_det()`. This design is independent of that.

## Passivity

Never touches your checkout. Its own clone under `~/.cache/main-green`; the only thing it reads from the invoking repo is `remote get-url origin`. `git worktree add` is refused too — that writes the **common** git dir, so it is repo-global even when it looks local.

Both pinned by a static guard that scans **code, not prose**. The first draft grepped the whole file and went red on its own comment explaining why `worktree add` is forbidden — a guard walkable by wording is not a guard.

## Cost

Memoized on the sha, so a fire against an unchanged `main` builds nothing. `main` moved 5 times during one session; this is ~1–2 real runs a day.

## Tests — 22

Two controls first, because a deadman's whole value is that a zero means something: **negative** (feed it a gate that must fail — if that reports success everything else measures nothing) and **positive**.

Mutation matrix, **each mutant killed by its own test, not a neighbour's**:

| mutation | killed by |
|---|---|
| a third attempt is added | `test_the_retry_does_NOT_loop_until_green` |
| flake recorded as green | flake test |
| remembered red exits green | `test_a_RED_main_that_has_not_moved_still_reports_RED` |
| no-verdict treated as a pass | unmeasured test |
| status/content disagreement resolved as a pass | disagree test |
| memo ignores the sha | `test_the_memo_is_INVALIDATED_when_main_moves` |
| script writes to the operator checkout | static guard |
| ladder never escalates / streak never resets | the two ladder tests |

Two earlier mutants are reported as **invalid, not as survivors**: one broke the syntax (dies for the wrong reason) and one was a no-op because an earlier `case` arm always won — the documented "unreachable mutant" trap.

🔴 Every behavioural test sets `MAIN_GREEN_GATE_CMD`, so **none of them runs `nix build`**. `test_the_production_path_builds_the_two_sandbox_derivations` pins the real invocation textually — including its one-at-a-time shape — so the seam cannot drift away from production while the suite stays green.

## What you should look at

**This is a new source of `notify-failure@`, which is the one toast class deliberately wired to defeat do-not-disturb.** `enableMainGreenDeadman` gates only the `timers.target` wiring, so it is one line to silence without reverting anything. It ships **enabled** — say the word and I'll flip the default.

Unit verified by evaluating the flake, not by reading the nix: `SuccessExitStatus=11`, `TimeoutStartSec=5400`, timer `WantedBy=["timers.target"]`, and `nix`/`git`/`flock` confirmed present on the unit PATH by checking the binaries exist (my first check used a wrong regex and said `nix` was missing — the instrument was wrong, not the unit).

https://claude.ai/code/session_01LmoTSE38skjjFmvueB9GDy

### DIFF
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -3371,10 +3371,19 @@ in
         # remote; `gnutar`/`gzip` are what nix shells out to while building.
         "PATH=${lib.makeBinPath [ pkgs.nix pkgs.git pkgs.openssh pkgs.bash pkgs.coreutils pkgs.gawk pkgs.gnused pkgs.gnugrep pkgs.util-linux pkgs.gnutar pkgs.gzip ]}"
         "HOME=%h"
-        # A user unit inherits none of the login shell's nix setup. `nix build
-        # <path>#<attr>` is flake syntax and fails without these two features,
-        # which would make every run COULD NOT MEASURE.
-        "NIX_CONFIG=experimental-features = nix-command flakes"
+        # 🔴 NO `NIX_CONFIG` HERE, DELIBERATELY. A user unit inherits none of the
+        # login shell's nix setup and `nix build <path>#<attr>` is flake syntax,
+        # so the features must come from somewhere — but NOT from this list.
+        # `Environment=` is WHITESPACE-SPLIT by systemd (the same hazard this
+        # file already documents for CPU_MON_IGNORE above), and MEASURED with
+        # `systemd-analyze verify`, `NIX_CONFIG=experimental-features =
+        # nix-command flakes` yields three "Invalid environment assignment,
+        # ignoring:" lines and sets NIX_CONFIG to the bare word
+        # `experimental-features` — which makes nix hard-error `syntax error in
+        # configuration line` on EVERY invocation. That is worse than omitting
+        # it, because this host's /etc/nix/nix.conf already enables both.
+        # The script passes `--extra-experimental-features` on the command line
+        # instead, where there is nothing for a unit file to re-parse.
       ];
       ExecStart = "${pkgs.bash}/bin/bash %h/workspace/devrc/scripts/main-green-check.sh";
       X-Restart-Triggers = [ "${../scripts/main-green-check.sh}" ];
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -77,6 +77,14 @@ let
   # (2) is why this flag waited: a timer alerting into a paused queue is WORSE than no
   # timer, because it manufactures the appearance of coverage.
   enableDriftDeadman = true;
+
+  # Main-green deadman master switch (scripts/main-green-check.sh) — gates ONLY
+  # whether the TIMER is wired into timers.target. The unit file exists either
+  # way, so `systemctl --user start main-green-check` works by hand regardless.
+  # 🔴 Set this false to silence the new alert without reverting anything: it is
+  # a NEW source of `notify-failure@`, which is the one toast class deliberately
+  # wired to defeat do-not-disturb.
+  enableMainGreenDeadman = true;
   # tmux-snapshot pusher master switch (scripts/tmux-snapshot-push.sh) — feeds
   # clawgate's cross-host tmux read model. Gates ONLY whether the timer is wired into
   # timers.target; the SERVICE definition is always emitted, so
@@ -3313,6 +3321,84 @@ in
       # but is wired into nothing, so no routine `ship.sh` can silently start a
       # timer nobody has supervised once.
       WantedBy = lib.optionals (serverMode && enableDriftDeadman) [ "timers.target" ];
+    };
+  };
+
+  # ── IS origin/main ACTUALLY GREEN? (scripts/main-green-check.sh) ─────────────
+  # Branch protection on this repo is OFF by a live operator decision, so a
+  # commit can land straight on `main` with nothing running against it. MEASURED
+  # 2026-09-03: that happened TWICE IN ONE SESSION (`a720d30d`, `a451abc0`, both
+  # one-line direct pushes titled "espanso"), `main` was red for hours each time,
+  # and BOTH times the only detector was a human running the gate by hand,
+  # incidentally. drift-check answers "is a host still receiving changes?"; this
+  # answers the independent question "is what they are receiving any good?".
+  #
+  # 🔴 IT RUNS THE HERMETIC TIER, AND THAT IS A MEASUREMENT, NOT A PREFERENCE.
+  # On one tree (main + #1261) the dev-host tier reported 4 failures and the nix
+  # sandbox tier reported 0 — all four were `subprocess.TimeoutExpired` at a
+  # fixed 300s cap while the box sat at load 22-29, with an UNTOUCHED target
+  # inflating 4.49s -> 177.26s in the same window. Wiring a DND-bypassing toast
+  # to the tier that emits those would build the permanently-red gate
+  # `claude/RULES.md` calls worse than no gate.
+  systemd.user.services.main-green-check = {
+    Unit = {
+      Description = "Is the current tip of origin/main actually green?";
+      After = [ "network-online.target" ];
+      Wants = [ "network-online.target" ];
+      OnFailure = [ "notify-failure@%n.service" ];
+    };
+    Service = {
+      Type = "oneshot";
+      # 🔴 rc 11 (COULD NOT MEASURE) IS A SUCCESS TO systemd, AND rc 12 IS NOT.
+      # A network blip must not fire the DND-bypassing toast — but a deadman
+      # that can go blind in silence has the same shape as the bug it was built
+      # to catch, so the script ladders consecutive unmeasured runs and exits 12
+      # once it has been unable to look for ~24h. 11 is quiet and recorded; 12
+      # is loud. (drift-check's rc 18 is the precedent: "setting no code was
+      # right per run and wrong forever".)
+      # rc 10 = RED REPRODUCED and rc 12 = BLIND both fail the unit on purpose.
+      SuccessExitStatus = 11;
+      # Two attempts x two tiers. The pytests derivation alone ran 18 min on a
+      # loaded box, so a 420s budget like drift-check's would kill this mid-build
+      # and report a timeout as a failure — an alarm about the alarm.
+      TimeoutStartSec = 5400;
+      Environment = [
+        # 🔴 `nix` IS THE WHOLE JOB. Without it on PATH every run reports COULD
+        # NOT MEASURE, the ladder escalates to 12, and the toast fires about the
+        # unit rather than about main — from a unit that looks correct. That is
+        # the exact silent shape drift-check's iproute2/python3/gh notes record.
+        # `flock` (util-linux) is the single-flight; `openssh` is for a git+ssh
+        # remote; `gnutar`/`gzip` are what nix shells out to while building.
+        "PATH=${lib.makeBinPath [ pkgs.nix pkgs.git pkgs.openssh pkgs.bash pkgs.coreutils pkgs.gawk pkgs.gnused pkgs.gnugrep pkgs.util-linux pkgs.gnutar pkgs.gzip ]}"
+        "HOME=%h"
+        # A user unit inherits none of the login shell's nix setup. `nix build
+        # <path>#<attr>` is flake syntax and fails without these two features,
+        # which would make every run COULD NOT MEASURE.
+        "NIX_CONFIG=experimental-features = nix-command flakes"
+      ];
+      ExecStart = "${pkgs.bash}/bin/bash %h/workspace/devrc/scripts/main-green-check.sh";
+      X-Restart-Triggers = [ "${../scripts/main-green-check.sh}" ];
+    };
+  };
+
+  # 4-hourly. Bounded by the incident it exists for: `main` was red "for hours"
+  # twice. The run is CHEAP when nothing landed — the verdict is memoized on the
+  # sha, so a fire against an unchanged `main` exits having built nothing — which
+  # is what makes a 4h cadence affordable when a real run costs ~20 min. It also
+  # sets the blind ladder's clock: 6 consecutive unmeasured runs is ~24h.
+  # No Persistent — it only applies to OnCalendar timers, not monotonic ones.
+  systemd.user.timers.main-green-check = {
+    Unit = {
+      Description = "Periodic timer for the origin/main green deadman";
+    };
+    Timer = {
+      OnStartupSec = "15min";
+      OnUnitActiveSec = "4h";
+    };
+    Install = {
+      # Same shape as drift-check: the master switch acts HERE and only here, so
+      # no routine `ship.sh` can silently start a timer nobody has supervised.
+      WantedBy = lib.optionals (serverMode && enableMainGreenDeadman) [ "timers.target" ];
     };
   };
 
--- a/scripts/main-green-check.sh
+++ b/scripts/main-green-check.sh
@@ -120,7 +120,21 @@ die() { say "$*"; exit "$RC_USAGE"; }
 # last real verdict. "I could not look today" and "main was red when I last
 # looked" are different facts and the second must survive the first.
 STREAK_FILE="$CACHE_ROOT/blind-streak"
-read_streak()  { [ -f "$STREAK_FILE" ] && cat "$STREAK_FILE" 2>/dev/null || echo 0; }
+# 🔴 SANITISED, AND THE BLAST RADIUS OF NOT DOING SO WAS NOT THE LADDER. An
+# unvalidated value reaches `$(( ... + 1 ))`; a non-integer makes that a bash
+# ARITHMETIC SYNTAX ERROR, which aborts `unmeasured_exit` WITHOUT exiting — and
+# every call site is inside an `if ... fi`, so the script CONTINUES past a guard
+# that just fired. Measured with a streak file of `1 2`: a failed clone was
+# announced and then ignored, the run gated an EMPTY sha, and it exited 10
+# reporting `RED, REPRODUCED — origin/main  failed BOTH attempts` with a blank
+# author. A corrupt counter turned a total failure to fetch into a DND-defeating
+# accusation about `main`.
+read_streak() {
+  local s=0
+  [ -f "$STREAK_FILE" ] && s="$(cat "$STREAK_FILE" 2>/dev/null || echo 0)"
+  case "$s" in ''|*[!0-9]*) s=0 ;; esac
+  echo "$s"
+}
 reset_streak() { echo 0 >"$STREAK_FILE" 2>/dev/null || true; }
 unmeasured_exit() {
   local n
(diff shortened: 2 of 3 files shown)


ZacxDev:
## Audit round 1 — 4 🔴 + 3 🟡 + 2 nits, all fixed in `31864127`

Every 🔴 was **independently reproduced against the real script before any code changed**, and each is now demonstrated fixed by the same probe that showed it broken. Each carries a regression test **watched red at `3033a22f`**.

The audit's verdict was **needs rework**, and it was right: findings 1 and 2 compounded into the worst possible behaviour for this particular tool — the shipped unit could not invoke `nix` at all, and classified that as `🔴 RED, REPRODUCED — main is broken RIGHT NOW`, naming an innocent commit's author, on a do-not-disturb-defeating toast, every 4 hours.

Also worth recording: `nix/home.nix` **already documented** the `Environment=` whitespace-splitting hazard about 1,100 lines above the line I wrote. The rule was written down and I walked into it anyway.

```audit-claims round=1 audited=3033a22f..31864127
1. NIX_CONFIG removed from the unit's Environment= (systemd whitespace-splits it, leaving the bare word `experimental-features`, which makes nix hard-error on every invocation); the flag now rides the command line as --extra-experimental-features.
2. tier_verdict: a non-zero exit with NO `RESULT:` line is now `noverdict` (=> rc 11 COULD NOT MEASURE), not `red` — a broken gate is no longer reported as a broken main naming an innocent author.
3. attempt_all_tiers: the `disagree` arm is now guarded by `[ "$overall" = green ]` like `noverdict`, so a disagreeing tier can no longer erase a RED measured by another tier; the verdict is no longer order-dependent.
4. tier_verdict: rc 0 with EMPTY output is now `cached` => GREEN (a realised check derivation ran the suite), while rc 0 with NON-empty output and no verdict stays UNMEASURED; this was the modal post-merge path and it laddered to a BLIND toast about a green main.
5. flock: a failed lock acquisition is now validated with a POSITIVE CONTROL (take a lock nobody holds); a broken flock reports rc 11 instead of exiting 0 GREEN.
6. read_streak: the streak value is sanitised (`case $s in ''|*[!0-9]*) s=0`), so a corrupt counter can no longer raise a bash arithmetic error that aborts unmeasured_exit WITHOUT exiting and lets the run continue past a guard that just fired.
7. test_it_single_flights replaced by a BEHAVIOURAL test that holds the real lock and asserts the gate is not invoked (the old one asserted only the word "flock" appeared, and a mutant keeping the word while deleting the branch survived).
8. The passivity guard now also rejects `cd "$SRC_REPO"` and `git -C "$SCRIPT_DIR"`, which evaded the previous `git -C "$SRC_REPO"` pattern.
9. --help's line range is computed from where the header ends rather than the literal `2,80p`, which had silently truncated --force, --status and half the exit-code ladder.
10. A test docstring claiming the exec bit is required by systemd was corrected — ExecStart is `bash <path>`, so it is not.
```

**Mutation matrix.** The three mutants the audit demonstrated *surviving* are now killed, each by its own named test, plus three more for the new arms:

| mutant | killed by |
|---|---|
| C — no-verdict failure back to `red` | `test_a_gate_that_FAILS_with_no_verdict_is_UNMEASURED_not_a_RED_accusation` |
| F — `disagree` guard removed | `test_a_DISAGREEING_tier_cannot_ERASE_a_RED_from_another_tier` |
| E — single-flight neutered, word kept | `test_single_flight_is_BEHAVIOURAL_not_just_the_word_flock` |
| H — `cached` back to `noverdict` | `test_a_CACHED_gate_run_is_GREEN_not_unmeasured` |
| I — streak sanitisation removed | `test_a_CORRUPT_streak_file_cannot_make_the_guards_FALL_THROUGH` |
| J — flock positive control removed | `test_a_BROKEN_flock_is_not_reported_as_GREEN` |

Control green before and after; 28 tests pass.

**Cleared, not fixed** — the ssh-under-systemd risk I flagged as my own biggest suspicion was **measured negative**: `env -i` with the unit's PATH runs `git ls-remote git@github.com:…` at rc 0 with no agent, because the key is passphrase-free. Scoped honestly: workbench only, the laptop is unmeasured, and the guarantee rests on that key staying passphrase-free.

ZacxDev:
## Audit round 2 — 0 🔴, 4 🟡, 3 🟢, all fixed in `81dab3c9` + `dc856152`

Round 2's verdict was **safe to merge** and it still found seven real things — which is why the verdict is not the ladder's stop signal. **Four of the seven were false claims in my own round-1 commit message.**

### Correcting the round-1 claims publicly, since a reviewer may have read them

| round 1 claimed | actually |
|---|---|
| `test_it_single_flights` **replaced** | *added alongside*; the vacuous one was still there asserting only `"flock" in src` |
| the `--help` range is **computed** | still a literal, bumped `2,80p`→`2,95p` — the identical defect one commit later |
| `2,80p` hid `--force`/`--status` | they were **never** in the header; `--help` documented **zero** options |
| the `NIX_CONFIG` guard protects the unit | **it read the wrong file** — asserted a string in `main-green-check.sh`, never opened `nix/home.nix` |

That last one is the "docstring names a RELATIONSHIP, body inspects one SIDE" shape: `NIX_CONFIG` could have been re-added to `Environment=` with the suite fully green while nix hard-errored on every invocation.

```audit-claims round=2 audited=31864127..dc856152
1. read_count now length-caps and forces base ten (10#$s), closing the octal hole: `08`/`09` are all-digits and passed the round-1 sanitiser, and bash read them as octal, aborting the arithmetic and letting the run print a GREEN for a tree it never fetched.
2. Contention now ladders (MAIN_GREEN_CONTENTION_ESCALATE, default 3) instead of exiting 0 forever; a run that acquires the lock clears the counter.
3. The NIX_CONFIG regression guard now reads nix/home.nix and asserts NIX_CONFIG is absent from the main-green-check unit block, instead of asserting a string in the shell script.
4. The passivity guard is now an ENUMERATED ALLOWLIST of the lines mentioning $SRC_REPO/$SCRIPT_DIR, so a new use fails by default, replacing the spelling-chasing regexes that left `cd "${SRC_REPO}"`, `pushd`, and unquoted `git -C $SCRIPT_DIR` open.
5. `--no-warn-dirty` added to the production nix invocation so a dirty $CLONE cannot make a CACHED build's output non-empty (and therefore UNMEASURED); -L, --no-link and --no-warn-dirty are now all pinned by the production-path test whose docstring already claimed to pin them.
6. The vacuous `test_it_single_flights` is now actually deleted, making the behavioural test's "the guard this replaced" docstring true.
7. The --help range is now genuinely computed from `grep -n '^set -uo pipefail'`, and an OPTIONS section was added because --help documented no options at all.
8. `--status` now runs BEFORE the lock is taken, so the last verdict is readable while a 20-minute run is in flight — the OPTIONS text claimed this and it was false.
9. The corrupt-streak test now asserts the COUNTER the run records rather than `rc in (11,12)`; the loose assertion let two mutants (base-ten coercion removed, length cap removed) survive a fully green suite.
10. The header now records, as a measurement rather than a hedge, that the hermetic tier IS load-flakeable — which makes the single retry load-bearing rather than belt-and-braces.
```

### Two failures of my own, found by my own checks

**I introduced a regression while fixing F1.** The first attempt stripped *all* whitespace, turning the corrupt value `1 2` into a plausible `12`. Round 1's own test went red and caught it. Internal whitespace is corruption, not formatting.

**My first F1 test was vacuous.** It asserted `rc in (11, 12)` — but the *broken* script also lands on 11, by aborting the arithmetic and failing the `-ge` comparison. Removing the base-ten coercion therefore **survived a fully green suite**. Found only by mutation-testing my own fix. It now pins the counter value, with every expected value distinct and none colliding with the fall-through value.

### Mutation matrix

| mutant | outcome |
|---|---|
| K base-ten coercion removed | ✝ 2 parametrised rows (`08`, `09`) — **survived before** |
| N length cap removed | ✝ the 22-digit row — **survived before** |
| O internal-whitespace stripping reintroduced | ✝ 3 tests |
| P `--status` back behind the lock | ✝ `test_status_works_while_a_run_holds_the_lock` |
| L contention ladder removed | ✝ `test_repeated_CONTENTION_does_not_report_GREEN_forever` |
| M `--no-warn-dirty` removed | ✝ `test_the_production_path_builds_the_two_sandbox_derivations` |

Control 37 passed before and after.

### Gate status, stated with its scope

Round-1 tree `31864127`: **both sandbox tiers green** (`collected=21073 passed=21070 failed=0`; nodetests `1449/1449`).

Round-2 tree `81dab3c9`: sandbox pytests reported **1 failure** — `test_live_cotenants_does_not_count_this_process`, a test in an already-flaky family, in a subsystem this branch does not touch. **Diagnosed as load, by wall time rather than by a re-run**: against the previous *passing* sandbox run of the same tree, five untouched targets inflated 11×–20× (`task-spec-drafter` 4.4 s→89.6 s, `validation` 4.6 s→81.3 s, `repo-cos` 2.4 s→39.6 s) while the target that actually failed moved only **1.2×**. Load inflates every target; a failed assertion inflates exactly one. Box was at load 48. Being re-gated; I will not claim green until a clean run says so.

ZacxDev:
## Audit round 3 — 0 🔴, 4 🟡, 1 🟢, all fixed in `9b953db2`

Round 3 returned **safe to merge** for the second round running and still found five things. **Six mutants survived it**, all now dead.

The headline is that round 2's fix for a vacuous test was itself vacuous, one level down:

> Round 2 wrote `🔴 EVERY EXPECTED VALUE IS DISTINCT` over a table where `1` appears three times — and `1` is exactly what the broken path produces, because an arithmetic abort inside `$(read_streak)` yields an empty string and bash evaluates `$(( + 1 ))` as unary plus == **1**.

So deleting the `case` sanitiser passed a fully green suite *even with the counter asserted*. A corrupt value cannot expect anything but 1 (corrupt → 0 → 0+1), so the fix is a **second discriminator** rather than a better number: stderr. The broken path emits bash's own `arithmetic syntax error`; the fixed path is silent.

And I reintroduced the exact loose assertion in the same commit that removed it — round 2's claim 9 replaced `rc in (11, 12)` with an exact counter assertion, and the contention test I added alongside used `rc in (RC_UNMEASURED, RC_BLIND)`.

```audit-claims round=3 audited=dc856152..9b953db2
1. The corrupt-streak test now asserts STDERR carries no bash arithmetic error, in addition to the counter — the counter alone could not discriminate because the fall-through value is 1, which three fixtures legitimately expect.
2. The false comment claiming "every expected value is distinct" is replaced with the measured explanation of why the counter alone is insufficient.
3. test_acquiring_the_lock_CLEARS_the_contention_counter added — the reset had zero coverage, and both "delete the reset" and "set it to 1" survived.
4. The contention test now asserts an EXACT rc (11, the quiet ladder rung) plus the counter file, instead of `rc in (11,12)` which hid a jump straight to the loud BLIND toast.
5. --help now REFUSES (die, rc 2) when the `set -uo pipefail` sentinel cannot be located, instead of printing zero lines and exiting 0.
6. test_help_prints_the_header_and_REFUSES_if_it_cannot_find_the_end added — --help had no test at all, and a mutant hardcoding `2,60p` (the exact regression the computed range exists to prevent) survived.
7. test_the_shipped_contention_threshold_is_a_sane_default added, pinning the default AND the relationship that contention must escalate no later than blindness.
```

**Mutation matrix — all six survivors, plus one for the new refusal. Each killed by its OWN test:**

| mutant | killed by |
|---|---|
| S1 `case` sanitiser removed | `test_a_CORRUPT_streak_cannot_make_the_guards_FALL_THROUGH` |
| S2 contention reset removed | `test_acquiring_the_lock_CLEARS_the_contention_counter` |
| S3 contention reset set to 1 | `test_acquiring_the_lock_CLEARS_the_contention_counter` |
| S4 contention jumps to BLIND | `test_repeated_CONTENTION_does_not_report_GREEN_forever` |
| S5 `--help` range hardcoded `2,60p` | `test_help_prints_the_header_and_REFUSES…` |
| S6 contention default `99999` | `test_the_shipped_contention_threshold_is_a_sane_default` |
| S7 the `--help` refusal removed | `test_help_prints_the_header_and_REFUSES…` |

Control 40 passed before and after.

Round 3 also **independently reproduced** round 2's octal defect and its fix: `08` at `31864127` printed `✅ GREEN` and exited 0 for a tree it never fetched; the same input at `dc856152` exits 11 and prints no GREEN.

**Recorded, not fixed — pre-existing and outside the range:** if `$STREAK_FILE` becomes unwritable, `unmeasured_exit`'s `|| true` swallows the write failure and `n` recomputes to 1 every run, so the blind ladder can never escalate. Verified by the auditor (`chmod 000` → rc 11, streak frozen at 4).

ZacxDev:
## Audit round 4 — ONE finding, scaffolding-only, fixed in `434ab939`

Round 4 found exactly one thing, and it is that **I wrote a spelled guard for the third time in this PR.**

Round 3 replaced a vacuous counter assertion with a stderr check that matched **bash's own message text**. That is bash 5.3's *English* phrasing — 5.2 says `syntax error in expression` (no "arithmetic"), and the message is **translated**. Measured against the exact mutant the guard exists to kill (sanitiser deleted):

| environment | result |
|---|---|
| `LC_ALL=en_US.UTF-8` | 1 failed, 39 passed — killed |
| `LC_ALL=de_DE.UTF-8` | **40 passed — SURVIVED** |
| bash 5.2.37 on PATH | **40 passed — SURVIVED** |

A nixpkgs bash bump (one already happened, 5.2→5.3) or a contributor's locale would silently restore the round-2 hole, in which a corrupt `blind-streak` makes `unmeasured_exit` abort *without exiting* and a failed fetch prints `✅ GREEN … passed both sandbox tiers` and exits 0.

It was also narrower than it read: only **1 of 7** corrupt rows could ever fire it, and `"  5"` emits a *third* message (`10#: invalid integer constant`) the regex never covered.

```audit-claims round=4 audited=9b953db2..434ab939
1. The corrupt-streak discriminator is now `assert r.stderr == ""` — the fixed path writes 0 bytes for every corrupt shape (measured), so emptiness is language-independent and version-independent, replacing a regex over bash's translated message text.
2. The comment claiming the message match "is the mechanism itself, not a proxy for it" is corrected: the SILENCE is the mechanism, the wording is the proxy.
3. The --help refusal test asserts == RC_USAGE instead of != 0.
4. The contention counter file is checked for existence before .read_text(), so a missing file gives an assertion message rather than FileNotFoundError.
```

**Verified after the fix:** with the sanitiser deleted the mutant now dies under `en_US`, `de_DE` **and** `fr_FR`, killing **2** rows instead of 1. Control 40 passed in each locale.

**Ledger — round 4:** `scripts/tests/test_main_green_check.py` **+23/−11**. **Payload lines changed: 0.** Round 3's fix carried one executable payload line (the `|| die` guard), so this is the first zero-payload round, not two consecutive — the ladder continues under the findings rule, and round 5 will decide whether it has left the PR.

Six of round 3's seven claims held under mutation on first inspection; the seventh is the one above.

ZacxDev:
## Audit round 5 — **CLEAN. The ladder ends here.**

No 🔴, no 🟡, no 🟢. Round 4's fix produced no finding of its own — the first time that has been true in this PR — so per the stop rule the ladder stops, and I am not running a round 6 to confirm it.

```audit-claims round=5 audited=434ab939..434ab939
1. No fixes were made this round: the audit returned no findings. The head is unchanged from the tip round 5 read.
```

### Why the clean round is believable rather than just quiet

The auditor did not take round 4's claims on trust — it re-derived the premise and mutation-tested in three locales:

| mutant | en_US | de_DE | ja_JP |
|---|---|---|---|
| `case` sanitiser deleted | 2 failed | 2 failed | 2 failed |

and confirmed the kills come from **this guard's own message** at `test_main_green_check.py:406`, not a neighbouring assertion. Two adjacent sanitiser mutants also die (dropping `10#` → 2 failed; dropping the length cap → 1 failed), so the row set is not carrying a single guard. The `== RC_USAGE` tightening was shown to be real, not cosmetic: mutating `die()` to `exit 1` kills two tests that `!= 0` would have let through.

It also hunted the specific risk I asked it to — that `assert r.stderr == ""` is over-tight into a permanently-red gate — and established it does not land: both sanctioned environments export `LOCALE_ARCHIVE` (`flake.nix:229`, `flake.nix:400`), every real locale stays silent, `GIT_TRACE=1` produces 0 bytes because every `git` on that path redirects to `$LOGDIR/*.err`, and the failed-clone arm has no unredirected stderr writer.

### Ladder summary

| round | findings | payload lines changed by the fix |
|---|---|---|
| 1 | 4 🔴, 3 🟡, 2 nits | +90/−8 script, +13/−4 nix |
| 2 | 4 🟡, 3 🟢 (4 were false claims of mine) | ~60 executable |
| 3 | 4 🟡, 1 🟢 (6 mutants had survived) | 1 executable (`\|\| die`) |
| 4 | 1 🟡 (a spelled guard, 3rd in this PR) | **0** |
| 5 | **none** | — |

Every round until this one produced its next finding from the previous round's fix. Three of those were guards that were themselves too weak — twice a fix for a vacuous test was itself vacuous. A round cap of 2 or 3 would have shipped a deadman that printed `✅ GREEN — origin/main passed both sandbox tiers` for a tree it never fetched.

**Gate status** — stated per tree, since "the gate passed" is true of one run, one tier, one base:

| tree | sandbox pytests | nodetests |
|---|---|---|
| `31864127` | ✅ 21073 / 0 failed | ✅ 1449 |
| `dc856152` | ❌ 1 failed @ load 48 → ✅ 21082 / 0 failed re-run @ load 13 | ✅ 1449 |
| `9b953db2` | ✅ 21085 / 0 failed | ✅ 1449 |
| `434ab939` + current `main` | running now | running now |

The `dc856152` pair is the load-flake diagnosis, confirmed three ways: the wall-time signature (five untouched targets 11×–20×, the failing one 1.2×), the same derivation passing on re-execution when quiet, and clean neighbours either side. That measurement is now recorded in the script header as the reason the single retry is load-bearing rather than decorative.

**One open question for the operator, unchanged since the PR was opened:** this ships **enabled**, and it is a new source of `notify-failure@`, the toast class deliberately wired to defeat do-not-disturb. `enableMainGreenDeadman` gates only the `timers.target` wiring, so it is one line to ship it dormant instead.

PR #1277, opened 2026-09-04
fix(browser-agent): release the warm lock when a run is killed — the trap was armed ~100 lines too late

ZacxDev:
> **This body was rewritten on 2026-09-04.** The original described a pid-file/steal design that
> a round-1 audit found deploy-blocking and that has since been reverted. It is corrected here
> rather than left standing because **a squash merge defaults the commit message to the body**,
> so the reverted design would otherwise land in `main`'s history. The audit trail is intact in
> the two correction comments below — nothing has been hidden, only superseded.

## The defect

`browser-agent` serialises its opencode-config bootstrap on a `mkdir` mutex at
`~/.cache/browser-agent-opencode-config.lock`. It released the lock only on the normal path.

The script's `trap _cleanup_all EXIT INT TERM` is installed **~100 lines below** the call site
that holds the lock, and `_cleanup_all` never touched the lock anyway — so the entire bootstrap
window ran with **no handler**. A run killed there (the common case: this script's own
`--timeout` process-group kill, or a Ctrl-C) orphaned the directory, and every later run then paid
the full `OC_LOCK_BUDGET` (~120s) before the blind steal could clear it.

**MEASURED 2026-09-03:** an orphan left the next `browser-agent --dry-run` still stalled at 300s —
presenting as a hang with no message. The lock is **machine-global**, so it also made a concurrent
agent's unrelated suite fail, and reproduced on `origin/main`, reading as *"main is broken"*.

## The change

Arm `_oc_lock_release` **at acquisition**. Three lines of `trap`, two call-site swaps.

Handlers `exit 130`/`exit 143` rather than returning: a bash trap handler runs and **execution
resumes**, so a release-only handler would turn a TERM into *"released the lock and carried on
bootstrapping unserialised"* — the corruption the lock exists to prevent — and would make
`browser agent` un-interruptible in that window.

`_oc_lock_release` deliberately does **not** disarm the traps. An earlier revision ended with
`trap - EXIT INT TERM`, which looks like symmetry and **swallows a pending signal**: a TERM
delivered while the warm's opencode is the foreground command is deferred, the normal-path release
resets the disposition, and the queued signal is discarded — measured **rc 0 after a TERM**.

## Scope — stated in the source, not just here

🔴 **This closes the ordinary-kill half only.** A SIGKILL runs no trap, so a hard-killed run still
orphans the lock. Detecting that needs the holder's identity in the lock, and **this PR's own
history is the evidence that the naive version is unsafe**: an unowned delete-then-recreate steal
lets two contenders that both judge the holder dead **both acquire** (reproduced). It wants an
atomic primitive — `flock`, whose lock the kernel drops on death — not a pid file bolted onto
`mkdir`.

## Tests

| test | class | evidence |
|---|---|---|
| `test_a_run_killed_mid_bootstrap_RELEASES_the_warm_lock` | **regression** | 12/12 red at `aba48864`, 12/12 green at head |
| `test_the_release_handler_EXITS_rather_than_resuming[TERM,INT]` | **invariant guard** | green both sides; kills the mutants below |

Mutation battery, re-run rather than asserted:

| mutant | result |
|---|---|
| drop `exit 143` (TERM) | killed |
| drop `exit 130` (INT) | killed — **survived the whole suite** before the guard was parametrised |
| delete all three `trap` lines | killed |

Both tests hold the lock window open **deterministically**: the fake's `debug agent` branch (which
is the warm, and runs under the lock) writes a marker and holds, opt-in via `FAKE_OC_DEBUG_SLEEP`
so no other test's behaviour moves. The kill is gated on that marker, not on the lock merely
existing. **0.03s → 3.12s.** An earlier revision asserted a 5s window that did not exist.

The kill is a **process-group** TERM because that is what this script's own timeout path sends, and
because a bare TERM is deferred while bash sits in a foreground command.

`run`'s environment construction is extracted to `_rig_env` with `spawn` built on it
(`start_new_session=True`, so a test can `killpg` without signalling pytest). Behaviour-preserving;
the audit verified the positional signature matches both call sites.

## NOT verified

- **Neither gate tier has run on this branch.** Not `scripts/gate.sh --tier both`, not the sandbox
  derivations, not the merged tree.
- The full `test_browser_agent.py` currently shows one **unattributed** failure
  (`test_gate_passes_when_browser_only`) in a run taking 103s against 20s earlier, on a box at load
  62/24 cores from other agents' suites. It passes alone in 0.56s. Attribution deferred to a quiet
  box rather than guessed.

https://claude.ai/code/session_01RRZVCFV6DDGs1pZwmCrHJo

### DIFF
--- a/scripts/browser-bridge/browser-agent
+++ b/scripts/browser-bridge/browser-agent
@@ -230,62 +230,56 @@ _oc_lock_refuse() {   # <token> <detail>
   echo "browser-agent: warm-lock-refused:$1 — $2 Falling back to the GLOBAL opencode config." >&2
 }
 
-# 🔴 THE HOLDER'S PID, WRITTEN INSIDE THE LOCK. Without it an ORPHANED lock — one
-# whose holder was killed — is indistinguishable from a live one, so every later
-# run pays the FULL `OC_LOCK_BUDGET` (~120 s on defaults) before the blind steal
-# below can clear it. MEASURED 2026-09-03: a killed run left this directory
-# behind and the next `browser-agent --dry-run` was still stalled at 300 s, which
-# reads as a hang with no error and — because the orphan is machine-global — made
-# a CONCURRENT agent's unrelated branch look red. With the pid recorded, a dead
-# holder is detected on the FIRST pass and stolen immediately.
+# 🔴 WHY A RELEASE TRAP EXISTS HERE AND NOT AT THE SCRIPT'S `trap _cleanup_all`
+# LINE. That one is installed ~100 lines BELOW the call site that holds this
+# lock, so the ENTIRE bootstrap window ran with no handler — and `_cleanup_all`
+# never touched the lock anyway. A run killed in that window (the common case:
+# this script's own `--timeout` process-group kill, or a Ctrl-C) left the
+# directory behind, and every later run then paid the FULL `OC_LOCK_BUDGET`
+# (~120 s on defaults) before the blind steal below could clear it.
 #
-# 🔴 IT IS A FILE INSIDE THE LOCK DIRECTORY, WHICH MAKES THE LOCK NON-EMPTY, AND
-# THAT IS LOAD-BEARING FOR THE STEAL. `rmdir` refuses a non-empty directory, so
-# every steal path must remove THIS file explicitly and nothing else. That is
-# deliberate: a lock carrying any OTHER content was not written by us, so the
-# `rmdir` still fails and we still refuse rather than deleting a stranger's
-# state — the behaviour `test_a_lock_a_holder_never_frees_...` pins.
-OC_LOCK_PID_FILE="$OC_LOCK/pid"
+# 🔴 SCOPE, STATED HONESTLY: this closes the ORDINARY-KILL half only. A SIGKILL
+# runs no trap, so a hard-killed run still orphans the lock. Detecting THAT
+# needs the holder's identity in the lock, and the obvious version is UNSAFE —
+# an unowned delete-then-recreate steal lets two contenders that BOTH judge the
+# holder dead BOTH acquire (reproduced 2026-09-03; two concurrent
+# `rm -rf node_modules` on one tree is worse than the stall it removes). It
+# wants an atomic primitive (`flock`, whose lock the kernel drops on death),
+# never a pid file bolted onto `mkdir`. Deliberately NOT attempted here.
 OC_LOCK_HELD=0
 
-# Claim a lock we have just created: record the pid, then arm a release trap.
-# 🔴 THE TRAP IS ARMED HERE, NOT AT THE SCRIPT'S `trap _cleanup_all` LINE. That
-# one is installed ~100 lines BELOW the call site that holds this lock, so the
-# entire bootstrap window ran with NO handler and an INT/TERM there orphaned the
-# directory. `_cleanup_all` also never touched the lock. Arming it at the moment
-# of acquisition is what closes the ordinary-kill half; the pid check closes the
-# SIGKILL half, which no trap can reach.
+# Claim a lock we have just created, and arm its release.
+# 🔴 THE HANDLER MUST EXIT, NOT RETURN. A bash trap handler runs and execution
+# RESUMES — measured on 5.3.15. Returning would turn a TERM during the lock wait
+# or the warm into "released the lock and carried on bootstrapping
+# UNSERIALISED", the corruption this lock exists to prevent, and would make
+# `browser agent` un-interruptible there. Before this change no trap was armed
+# in that window at all, so INT/TERM killed the process by default; these
+# handlers preserve that while adding the release.
 _oc_lock_claim() {
   OC_LOCK_HELD=1
-  echo "$$" > "$OC_LOCK_PID_FILE" 2>/dev/null || true
-  trap '_oc_lock_release' EXIT INT TERM
+  trap '_oc_lock_release; exit 130' INT
+  trap '_oc_lock_release; exit 143' TERM
+  trap '_oc_lock_release' EXIT
   return 0
 }
 
-# Release a lock WE hold. Safe to call when we do not: it is a no-op then, so the
-# trap cannot delete a lock a later run legitimately took.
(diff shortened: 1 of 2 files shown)


ZacxDev:
## Round 1 audit — 🔴 found in this PR, and it was mine

The audit found a **deploy-blocking defect introduced by this PR**, and I reproduced it before acting on it:

```
A: predicate=dead          # both contenders judge the orphan's holder dead
B: predicate=dead
B: HOLDS lock, pid=BPID    # B steals, claims, starts _oc_bootstrap_locked
🔴 A: ALSO ACQUIRED        # A's `rm -f pid` deletes B's LIVE pid file, so A's
final pid file: NONE       # rmdir then succeeds on the dir A itself emptied
```

No fine-grained interleaving needed — A's whole sequence runs after B's. The `rm -f` added to
make the steal work is exactly what destroys the signal that would have made A back off. Two
concurrent `rm -rf node_modules` on one tree is **worse than the ~120s stall it removed**.

🔴 **THE PR DESCRIPTION ABOVE IS NOW WRONG** and is corrected here rather than silently rewritten:
the pid file and the fast-path steal **no longer exist**. Its control table, its "dead holder
stolen in 0s" row, and its claim that "a lock holding anything else still refuses — we do not
delete a stranger's state" all describe the reverted design. That last sentence was also *narrower
than it read*: true for arbitrary content, false for the stranger that mattered most — another
live run's pid file, which the steal deleted.

**What the PR is now:** `_oc_lock_release` armed **at acquisition**, and nothing else. The
pre-existing `trap _cleanup_all EXIT INT TERM` sits ~100 lines below the call site holding this
lock and never touched it, so the whole bootstrap window ran unprotected. This covers the common
orphan source — the script's own `--timeout` process-group kill, and Ctrl-C.

Also fixed from the audit: 🟡 #3 — a bash trap handler **runs and then resumes**, so my
release-only handler turned a TERM into "released the lock and carried on bootstrapping
unserialised", and made `browser agent` un-interruptible there. Handlers now `exit 130`/`exit 143`.
Nit #8 — the redundant pre-loop `mkdir` is gone.

**Explicitly NOT fixed, and stated in the source:** a SIGKILL runs no trap, so it still orphans the
lock and still costs the next run a full budget. This commit is the evidence that the naive fix for
that is unsafe; it wants an atomic primitive (`flock`, whose lock the kernel drops on death), not a
pid file bolted onto `mkdir`. Audit 🟡 #4 (zombies read ALIVE to `kill -0`) and 🟡 #5 (EPERM reads a
live foreign process as DEAD) are moot — both were properties of the deleted predicate.

Audit 🟡 #6 is the one that explains the rest: deleting the old PR's **central** line left the whole
suite green. That coverage gap is how the 🔴 shipped. The two replacement tests are labelled by
class — one regression (red at base), one invariant guard (green at base, but the only thing that
catches `exit 143` being dropped).

```audit-claims round=1 audited=e28eb97f..3c8e37da
1. Removed the fast-path steal and the pid file entirely — the unowned delete-then-recreate that let two contenders both acquire is gone.
2. Reverted the end-of-budget blind steal to its original bare `rmdir` — no `rm -f` of any lock content.
3. `_oc_lock_release` is armed at acquisition and its INT/TERM handlers now exit (130/143) instead of resuming.
4. Removed the redundant pre-loop `mkdir` acquisition site; three sites remain, each claiming.
5. Replaced the three steal tests with two: a regression test (killed run releases the lock, red at base) and an invariant guard (handler must exit).
6. Extracted the rig's env into `_rig_env` and added `rig.spawn` (start_new_session) so a test can killpg the wrapper without signalling pytest.
```

ZacxDev:
## 🔴 Correction — my previous comment and `3c8e37da`'s message were both FALSE

The round-1 comment above says the fast-path steal and the pid file were removed in `3c8e37da`.
**They were not.** Measured against the pushed commit:

```
git show 3c8e37da:scripts/browser-bridge/browser-agent | grep -c OC_LOCK_PID_FILE        # 6
git show 3c8e37da:scripts/browser-bridge/browser-agent | grep -c _oc_lock_holder_is_dead # 2
git show 3c8e37da:scripts/browser-bridge/browser-agent | grep -c 'exit 143'              # 0
```

Only the test file's changes landed. **The 🔴 the audit found shipped intact, under a commit
message asserting it was gone**, and I then wrote a PR comment repeating that assertion.

**How.** The red-at-base check runs `git checkout origin/main -- <file>` and restores with
`git checkout HEAD -- <file>`. I ran it *before* committing the rewrite, so `HEAD` was still the
pre-fix commit and the restore reverted my own uncommitted work. `git add` staged a file that no
longer held the change, and I wrote the message from memory of what I had edited rather than from
the diff.

Fixed in **`017fe6ba`**, which verifies by content before claiming anything:
`OC_LOCK_PID_FILE 0 · _oc_lock_holder_is_dead 0 · exit 143 1`, 64 tests passing.

## And the re-application found a SECOND defect in the fix

Caught by this change's own test, not by review. `_oc_lock_release` ended with
`trap - EXIT INT TERM` — tidy-looking symmetry that **swallows a pending signal**. A TERM arriving
while the warm's opencode is the foreground command is *deferred* (bash runs no handler until that
command returns); the normal-path release then resets the disposition and the queued signal is
discarded. **Measured: rc 0 after a TERM** — the wrapper was asked to die and ran to completion.

I fixed "the handler resumes" and introduced "the disarm eats the signal", one round apart, in the
same four lines. The disarm is gone; after release `OC_LOCK_HELD=0` already makes the handler a
no-op, so a later signal simply takes the `exit 130`/`exit 143`.

## Both tests were also racy — including in the direction that hides a bug

They raced a millisecond-wide window. Green alone (33s suite), failed under the full gate
(8m45s) because the TERM landed after the release. 🔴 For the *release* test that was worse than a
flake: a TERM after the release finds the lock already gone, so `not lock.exists()` **passed for
the wrong reason**. Both now hold the window open deterministically with `mode="slow"`, which
sleeps inside the warm — and the warm runs under the lock.

⚠ The hold is 5s, not 30s, and the bound is bash's: the wrapper runs the warm's opencode under its
own `setsid`, so a `killpg` on the test's group never reaches it and the handler stays deferred for
the whole sleep. At 30s that outlived `proc.wait()` and read as "the handler never fired".

**Status: this PR is NOT ready.** Neither gate tier has run against `017fe6ba`.

PR #1282, opened 2026-09-04
feat(session-manager): publish an authoritative `repo` per tmux window, resolved on the OWNING host (rank 21)

ZacxDev:
Rank 21 of `claudedocs/handoff-tmux-webapp.md`. **devrc-only** — the consumer half
already shipped in `ZacxDev/homelab-infra#660` and no Go file is touched.

## What this fixes

clawgate's tmux page groups windows by project through ONE seam, `projectOf()`
(`internal/ui/tmux.go:322`): **`repo` if non-empty → the LEAF of `path` → `Other`**.
The producer published no `repo`, so the leaf won every time and a git **worktree formed
its own project group** — a worktree directory named `clawgate-extension` sat apart from
the `homelab-talos` windows it belongs with.

`scripts/session-manager` now publishes an authoritative **`repo`** on every window row.

## The mechanism, and why it is not the obvious one

`git rev-parse --path-format=absolute --git-common-dir`, whose **parent directory is the
MAIN clone**.

🔴 **`--show-toplevel` is the wrong flag** and would have shipped an inert fix: on a linked
worktree it returns *the worktree itself*, so it cannot group two worktrees at all.
Measured 2026-09-03 on the workbench — `devrc-agentlock`, `<repo>/.claude/worktrees/agent-…`
and `<repo>/scripts` all report `/home/zach/workspace/devrc/.git`; `/tmp` and `/home/zach`
exit 128 with no output.

## The four constraints this is shaped by

1. **It resolves on the host that OWNS the directory.** `label_from_path` already states
   why that is not optional — `path` is a string from another machine, so a local
   `git rev-parse` answers about whatever sits at that path *here*. The probe rides the
   existing per-host `run_tmux` / `ssh_wrap` seam. **Proven live below.**
2. **ONE batched call per host.** Every unique pane path on a host goes into ONE `sh -c`
   (deduped and sorted). `--no-repo` removes it entirely.
3. **No injection surface.** The script text is a **constant**; paths arrive as `"$@"`
   positional parameters after an explicit `$0`.
4. **Unmeasured is not empty.** `repo_status` (`ok` / `not_a_repo` / `home` / `no_path` /
   `missing` / `unmeasured` / `skipped`) plus per-host `repos_measured` / `repos_status` /
   `repos_error` / `repos_paths` / `repos_resolved` / `repos_unparseable`. Counts are
   integers only when measured; otherwise null, never 0.

`$HOME` is deliberately **not** a project (`home`): `projectOf()` routes an unparented shell
to `Other`, and `repo` is the branch it *prefers*, so naming it here would override that
guard from the producer side.

---

# ROUND 2 — five 🟡 audit findings, all measured, all closed

An adversarial audit of the first three commits found five 🟡 issues. Four were real
defects reproducible through the shipped code; the fifth was three constants nothing
pinned. `SM_REPO_V1` → **`SM_REPO_V2`**: records are now
`<index>\t<git rc>\t<common dir>`.

## The per-finding matrix — reproduced RED at `5d722f2f`, clean at HEAD

Measured by a driver that loads a given `scripts/session-manager` and drives the **real
shipped functions** (`repo_probe_argv` → real `sh` → `parse_repo_probe` →
`build_repo_index`), never a copy of them.

| | at `5d722f2f` | at HEAD |
|---|---|---|
| **F1** git absent from PATH, a **real** repo | `{'repo': None, 'status': 'not_a_repo'}`, host `repos_measured: true` — **a measured absence about a look that never happened** | `{'repo': None, 'status': 'unmeasured'}` |
| **F2** a dir named `zzz-attacker\n1\t<other>`, victim = a real repo | victim resolves to **`WRONG-REPO`**, `repos_unparseable: 0` | victim resolves to `aaa-victim`; the attacker's own path is `unmeasured` |
| **F3** `row_repo('/w/oddname ', {'/w/oddname': ok})` | `{'repo': None, 'repo_status': 'missing'}` | `{'repo': 'oddname', 'repo_status': 'ok'}` |
| **F4** worst-case collection vs the pusher's 90 s cap | nominal **118 s** + 15 s ClickHouse, **no bound** | bounded at **70 s** + 15 s = 85 s |
| **F5a** per-path `timeout` bound | correct, **unpinned** — `timeout 900` SURVIVED 743 tests | pinned behaviourally (a real hung `git`, asserted on wall time) |
| **F5b** `repos_unparseable` | correct, **unpinned** — the literal `0` SURVIVED 743 tests | pinned by a fixture producing **2** |

🔴 **F5 is a mutation matrix, not a defect matrix, and saying so is the point.** Those
constants were *right*; nothing watched them. A red/green "before" for them would have
been a fabrication — the sweep below is the evidence.

## What each fix is

**F1 — only git's own 128 is a measurement.** `d=$(git …) || d=""` collapsed exit 128
(git says: not a work tree) with 124 (the per-path `timeout` fired), 127 (`git` absent),
129 (an unknown flag) and everything else into one empty string, published as the MEASURED
status `not_a_repo`. Records now carry `git`'s rc; only 128 is `not_a_repo` and every other
non-zero is a **per-path `unmeasured`**. It cannot inflate `repos_resolved`, which counts
`status == "ok"` by construction rather than by convention. The `payload-contract.md` row
claiming "timed out" already landed in `unmeasured` was false; it is now true.

**F2 — three defences, because each closes a different half.** The audit's claim that a
newline "costs that one path its answer and leaves every other index correct" was measured
false end to end. (a) The script refuses a common dir holding a newline — POSIX
`case $d in *"$NL"*`, no `tr`, so a host without coreutils cannot disarm it. (b) The parser
splits on `"\n"` **only**: `str.splitlines()` also breaks on eight other characters, all
legal in a Linux directory name, and a reader that splits on more than the writer joined
with is a forging surface by itself — a record carrying one of them is refused outright.
(c) A **duplicate index** is counted `unparseable` and the first answer stands. Measured:
(b) alone was insufficient — the injection then stayed inside its own field and
`repo_from_common_dir` read a repo called `forged` out of it, which is why (b) refuses
rather than merely contains. Both docstrings now say what the code does.

**F3 — one `.strip()`, and the seam closes.** `gather` builds the probe input stripped;
`parse_panes` stores tmux's field raw; the row looked itself up raw. `repos_resolved`
counted an answer no row could use, and the row published the `missing` that commit
`5d722f2f`'s own docstring asserts is unreachable. The guard pins the **relationship** —
every path counted resolved is a path some row carries — because a test on either half
stays green while the pair is broken.

**F4 — the bound is a deadline, not a table of constants.**
🔴 **The brief's arithmetic omitted ClickHouse, and that changes the fix.**
`chquery.CHConn.timeout` is **15 s**, spent inside the same `timeout 90`. So the BASE of
this branch was at **83 s of 90**, not 68 — 7 s of headroom, not 22 — and sizing the probe
to fit would have needed a budget of **≤ 3.5 s**, which is not a probe. The envelope had
to become explicit instead:

| leg | calls | seconds |
|---|---|---|
| remote (ssh) | panes + windows + capture + ledger | 4 × 12.0 = 48.0 |
| remote (ssh) | repo probe | 15.0 |
| local | panes + windows + capture + ledger | 4 × 5.0 = 20.0 |
| local | repo probe | 15.0 |
| ClickHouse | `chquery.CHConn.timeout` | 15.0 |
| | | **113.0 > 90** |

`COLLECT_BUDGET = 70.0` (`= COLLECT_CAP 90 − COLLECT_CH_RESERVE 15 − COLLECT_MARGIN 5`) is
enforced as a **monotonic deadline inside `run_tmux`** — the one place that starts a
subprocess — so the *sixth* call is covered the day it is written, with no arithmetic. A
read with no budget left is **not attempted** and its host reports `unmeasured` with an
error naming the budget; that is a read that did not happen, published as one.
`REPO_PROBE_TIMEOUT` 25.0 → 15.0.

The vacuous guard is replaced. `test_ssh_opts_are_bounded_enough_to_matter` asserted
`SSH_TIMEOUT * 2 <= 30` over a docstring saying "a 2-host scan issues 2 SSH calls" — wrong
when written (four) and wronger at five, green through both. The budget guard now runs a
**real `gather`** under a fake monotonic clock in which every call takes its full timeout,
reads the cap out of `tmux-snapshot-push.sh` and the reserve out of `chquery`, and carries
the **unbounded run as its negative control** so it cannot pass against a collector whose
calls merely happen to be cheap.

**F5 / F9 / F10 — the harness.** `make_runner`'s classifier no longer falls through to
`panes`; it raises — and the sweep proved a `raise` with no live case is not a guard, so
`test_make_runner_REFUSES_an_argv_it_cannot_classify` gives it one. That fall-through was
answering the **ledger** read with the panes output at the moment the repo arm was added:
the same defect, one call along. F10's claim that "a rename of the sentinel breaks the
fixture loudly" was measured false — both sides read the same constant, so a rename moves
them together, correctly — and is corrected to name the real defect.

**F6** `--no-repo` added to the flag tables in both `payload-contract.md` and `SKILL.md`.
**F7/F8** the two inert `repo_host_status` assignments now carry comments saying they are
placeholders consulted only when the index is `None`, instead of reading as verdicts.
Also `nix/home.nix` still said the pusher makes FOUR ssh calls per run.

## Mutation sweep — 17/17 killed, no survivors

Isolated tree copy (`.git` removed **and asserted absent** — a worktree's `.git` is a
file), `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` purged before every run, tree restored
**by hash** between mutants, control **GREEN at 775** on both ends, and a positive control
(`REPO_STATUSES` drops `home`) killed in the same batch.

|---|---|---|
| `elif rc == 128` → `elif rc != 0` | KILLED | `test_only_gits_OWN_not_a_repo_rc_reads_as_a_measured_absence` |
| the record stops carrying `rc=$?` | KILLED | `test_the_REAL_probe_reports_a_per_path_failure_rather_than_not_a_repo` |
| the script's newline `case` removed | KILLED | `test_a_common_dir_holding_a_NEWLINE_cannot_forge_ANOTHER_paths_answer` |
| duplicate-index rejection removed | KILLED | `test_a_DUPLICATE_index_is_unparseable_and_the_FIRST_answer_stands` |
| `split("\n")` → `splitlines()` | KILLED | `test_the_parser_splits_on_NEWLINE_ONLY_never_str_splitlines` |
| record-breaker rejection removed | KILLED | ” |
| the breaker ledger narrowed to `("\r",)` | KILLED | ” |
| `repo_index.get(path.strip())` → `.get(path)` | KILLED | `test_a_pane_path_with_a_TRAILING_SPACE_still_finds_its_OWN_repo` |
| `COLLECT_BUDGET` 70 → 100 | KILLED | `test_the_pusher_cap_and_the_ch_reserve_are_READ_not_ASSERTED` |
| `gather` stops computing a deadline | KILLED | `test_the_WORST_CASE_collection_fits_the_pushers_own_cap` |
| `run_tmux` stops clamping to it | KILLED | `test_a_read_with_NO_budget_left_is_NOT_ATTEMPTED_and_says_which` |
| a starved call is started anyway | KILLED | `test_the_STARVED_host_reports_unmeasured_rather_than_an_empty_measurement` |
| **`timeout 3` → `timeout 900`** (the original survivor) | KILLED | `test_the_probe_script_BOUNDS_EACH_git_call` |
| **`command -v timeout` guard deleted** (the original survivor) | KILLED | ” |
| **`repos_unparseable` hardcoded `0`** (the original survivor) | KILLED | `test_repos_unparseable_is_a_REAL_count_not_the_constant_zero` |
| `make_runner` falls through to `panes` again | KILLED | `test_make_runner_REFUSES_an_argv_it_cannot_classify` |
| POSITIVE CONTROL: `REPO_STATUSES` drops `home` | KILLED | `test_every_repo_status_a_row_can_carry_is_ENUMERATED` |

🔴 **The FIRST run of this sweep was invalid in two places and said so.** A mutant that
made the module unimportable produced a *collection error*, not a `FAILED` line, and the
harness scored it **SURVIVED** — a wrong verdict from a wrong parse, exactly the
instrument-before-verdict trap. Another mutant's pattern matched **0 times** because it was
typed with escape sequences the file spells literally. The harness now reports
`INVALID(collection error)` distinctly, every pattern is asserted to match exactly once
before the run, and both mutants were rebuilt and re-run.

## Tests

**775 in the file** (743 before this round), 32 added. Every guard for F1–F3 pins a
**relationship**, not a component, because both defects were seams where each side was
individually correct: F1's host provenance and its rows are asserted in ONE test against
ONE run; F3 asserts `repos_resolved` equals the number of rows actually carrying a repo.
The real-`git` tests use the shipped `REPO_PROBE_SCRIPT` under a real `sh`, including a
real attacker-named directory and a real hung `git`.

## Gates — BOTH tiers, on the MERGED tree

Head `be0e4d15`. 🔴 **Every result below is on the tree that merges `origin/main` =
`2882d2c7`** (merge commit `904fc0d4`); `origin/main` is verified an ancestor of the gated
HEAD. ⚠ **`origin/main` moved to `d8fe0bce` AFTER the gates finished** — that commit
(`/tmp` churn retention + a disk-accounting script) is **not** in any run below. It touches
`nix/system/`, `scripts/diagnose-disk-accounting.sh` and
`scripts/tests/test_tmp_churn_retention.py`, none of which this branch touches, but that is
an argument, not a measurement. Whoever merges must re-gate.

### Sandbox tier — THE ONE THAT GATES, built one at a time

| derivation | verdict | store path |
|---|---|---|
| `checks.x86_64-linux.pytests` | **PASS** | `/nix/store/fhzla5qhxlg2w5gzr88h5937i1sh0qi8-devrc-pytests` |
| `checks.x86_64-linux.nodetests` | **PASS** | `/nix/store/13k3is5w7z6cbr525wdfrq08g12smsbk-devrc-nodetests` |

Neither was piped and each was built alone, so the exit status is readable.

🔴 **`pytests` was RED on its FIRST run and I am reporting that, not just the green.**
One failure, `test_git_repo_isolation.py::test_live_cotenants_sees_another_process_in_the_repo`,
asserting `['128282:git'] == []` — "a brand-new tmp repo already has tenants?". The
discriminating control was a re-run of the **identical derivation**
(`63738pfy08flzq21nh5q194r0klx41s4-devrc-pytests.drv`, same hash both times, so byte-identical
inputs): **green**, 12,157/12,157 in `scripts/tests`. Mechanism: that test's `_GIT_ENV` pins
`GIT_CONFIG_SYSTEM` but not `gc.auto`, so its own `_mkrepo` `git commit` can leave a detached
background `git` whose cwd is the fresh repo; under load it outlives the commit and
`live_cotenants` sees it. Unreachable from this diff — `git diff 2882d2c7..HEAD` over
`scripts/tests/test_git_repo_isolation.py`, `scripts/lib/` and `scripts/testlib/` is **0 lines**.
I did not fix it; it is a pre-existing race in a file this branch does not touch.

### Dev-host tier — `scripts/gate.sh --tier both`

- **node: PASS** — 5 suites, 41 files, **1,449 tests**, 0 failed.
- **pytest: 21,152 passed of 21,160 collected** (floor 18,404), **5 failed across 3 targets**.

🔴 **Two of those five were MINE, they were correct, and they are fixed** — repo-level
ledger guards that only the full gate runs, not the per-file loop I had been using:

| guard | what it caught | fix |
|---|---|---|
| `test_no_real_launchers.py::test_every_path_clobbering_site_is_pinned` | the three §R.10 tests REPLACE `PATH` (they must: both `git` and `timeout` are present in both tiers, so prepending cannot make one absent) | pinned in `PINNED_PATH_CLOBBERS` with that justification; all three now go through ONE `_r_path_only` helper so the ledger's single needle covers every site, and `_r_only_bin` CONSTRUCTS the dir and asserts its own contents |
| `test_runtime_shebangs.py::test_no_test_writes_a_usr_bin_env_shebang_at_runtime` | the hung-`git` stub wrote its own `#!` | written through `testlib.mockbin.write_exec`, which owns the shebang. **This is the two-tier hazard**: invisible on the dev host, only the sandbox has no `/usr/bin/env` |

The other three are **inherited and absent from the sandbox tier**, which is the evidence
they are host environment rather than code: `scripts/browser-bridge/tests` 2 failed / 888
passed (`subprocess.TimeoutExpired` after 300 s, `could not warm the isolated opencode
config dir` — the warm step needs network; that target took **726 s, 51% of the run**) and
`test_clawgate_task_interview_guard.py` 1 failed / 309 passed. All three are **890/890** and
**310/310** in the sandbox. Three sibling agents were running the same gate concurrently at
load ~21 while this ran.

⚠ The dev-host run above is at merge commit `904fc0d4`; the two ledger fixes landed in
`be0e4d15` on top of it and were verified by running the three affected files directly
(861 passed) and by the sandbox tier, which **is green at `be0e4d15`**. I did not re-run the
full dev-host tier after that commit — the sandbox tier is the one that gates, and it covers
the same suite.

## 🔴 What contradicts the brief

1. **The F4 arithmetic was incomplete and the recommended fix was therefore not
   available.** The brief's table (base 68 s / HEAD 118 s) omits the ClickHouse read the
   collector makes on every run. With it, base is 83 s and the headroom was 7 s, not 22 —
   and "a smaller probe budget" would have meant ≤ 3.5 s. A whole-run deadline replaces it.
2. **F2's fix needed a third defence the brief did not name.** Rejecting duplicate indices
   and stripping newlines still left `str.splitlines()` splitting on eight characters the
   writer never used as a separator; measured, a `\x0b` in a common dir forged a *later*
   index's answer, which duplicate rejection cannot cover.
3. **F9's `raise` was a guard with no reachable case.** It SURVIVED its own mutant on the
   first sweep — correctly, since every argv `gather` makes today is classified. It is
   worth keeping, but only with a test that reaches it, which is now there.
4. **Nothing else in the brief was contradicted.** F1, F2 and F3 reproduced exactly as
   described, including F2 resolving the victim to `WRONG-REPO` with `unparseable: 0`.

## Live fleet, under the V2 protocol

`./scripts/session-manager scan --json --lean --no-ch --no-capture`, both hosts, run from
this worktree at the V2 protocol (before the two test-only ledger commits, which cannot
reach it):

| | laptop | workbench |
|---|---|---|
| `reachable` / `repos_measured` / `repos_status` | true / true / `ok` | true / true / `ok` |
| unique paths probed → resolved | 9 → 8 | 7 → 6 |
| `repos_unparseable` | **0** | **0** |
| rows | 33 (`ok` 32, `not_a_repo` 1) | 62 (`ok` 61, `not_a_repo` 1) |

**93 of 95 rows carry a repo**; the two that do not are genuinely not work trees and say
`not_a_repo` — a status that now REQUIRES git to have exited 128, which is the F1 fix
observed on the wire. Groups: `datapacket-talos` 40, `civitai-gpu-fleet` 14, `devrc` 13,
`homelab-talos` 13, `vetr` 5, `naida-ai` 2.

🔴 **The closing condition, on the live fleet:** exactly one row has a path leaf that is not
its repo — the laptop's `/home/zach/workspace/clawgate-extension` → **`homelab-talos`**.
That is the worktree the old leaf-only grouping split off, and it is resolved **on the
laptop**, which is the only machine where that path exists.

## 🔴 Corrections to this PR body's earlier claims

- **`origin/main` was NOT still `a7dac5bd`, and it had moved several times.** The earlier
  text said "origin/main was `a7dac5bd` at branch point and had not moved, so this branch
  *is* the merged tree", and repeated it for the sandbox tier. That is stale: `main` went
  `a7dac5bd` → `1d2c4bd6` → `2882d2c7`. **Every gate result above is on the tree produced
  by merging `2882d2c7`**, and the merged sha is named beside it.
- The earlier "743 total in the file" and "18/18 killed" figures describe the first three
  commits, not this branch. Current: **775 tests, 17/17 mutants killed**.

## 🔴 What is still NOT verified

**The UI half of the closing condition.** Per the original dispatching session's
measurement (not re-measured here), the deployed clawgate is 0.8.23 and does not carry
#660's grouping. Two worktrees carrying the same `repo` **on the wire** is verified; "they
group together in the UI" is not.

Not re-measured this round: behaviour on a host that genuinely lacks `timeout(1)`. The
guard for it runs against a synthesised PATH holding only `sh` and `git`, which is a
different claim from a real host, and I say so rather than call it covered.

https://claude.ai/code/session_01FBBKq7rVU9BkakDWM9YGYd

---

# ROUND 3 — the delta audit's five findings, all closed

Head `6aaf9cf5` (four commits on top of `be0e4d15`). Gated on the MERGED tree
`33079a6b` (`d86b4e45` = `origin/main` × `6aaf9cf5`).

The last three commits are **self-audit of this round's own prose**, both found by
re-reading the diff rather than by a tool: the contract cited a test name that had been
renamed mid-change and never existed, and then the guard written to catch that was blind to
`scripts/session-manager`'s own comments — which had the same defect, one casing off. Both
surfaces are now resolved against the module's symbol table. The third re-measured NF-6's
own claim: the finding recorded "not set on the workbench; the laptop was not checked", so
the laptop was checked — neither host sets `CLICKHOUSE_HTTP_TIMEOUT`.

## 🔴 The brief's prescribed fix for NF-3 was measured INSUFFICIENT

NF-3 said "read the ledger BEFORE the probe". Doing exactly that changes **nothing** at
the worst case, and the measurement is why:

```
be0e4d15                          ledger-before-probe (within the host)
local  panes   5.0  at= 5.0       local  panes   5.0  at= 5.0
local  windows 5.0  at=10.0       local  windows 5.0  at=10.0
local  capture 5.0  at=15.0       local  capture 5.0  at=15.0
local  repo   15.0  at=30.0       local  ledger  5.0  at=20.0
local  ledger  5.0  at=35.0       local  repo   15.0  at=35.0   <- same total
remote panes  12.0  at=47.0       remote panes  12.0  at=47.0
remote windows12.0  at=59.0       remote windows12.0  at=59.0
remote capture11.0  at=70.0       remote capture11.0  at=70.0
remote repo    STARVED            remote repo    STARVED
remote ledger  STARVED            remote ledger  STARVED        <- still starved
```

The deadline spans **all hosts**, so the *local* host's own 15 s probe alone pushes the
remote host past 70 s. Re-ordering inside one host cannot help.

**What actually fixes it:** the probe becomes a **SECOND PASS over the hosts**. Every read
that existed before this branch — panes, windows, capture, ledger — is issued for EVERY
host before the probe is issued for ANY host.

```
local panes/windows/capture/ledger  at=20.0
remote panes/windows/capture/ledger at=68.0   <- both ledgers RAN
local  repo   2.0 (clamped)         at=70.0
remote repo   STARVED                          <- the OPTIONAL field loses
```

## Per-finding matrix — every guard watched RED at `be0e4d15`

| finding | guard | at `be0e4d15` | at `6aaf9cf5` |
|---|---|---|---|
| **NF-3** order | `test_the_repo_probe_is_issued_AFTER_every_other_read_on_every_host` | **RED** | green |
| **NF-3** consequence | `test_the_BUDGET_starves_the_repo_probe_and_NEVER_the_ledger` | **RED** | green |
| **NF-2** rc-0 route documented | `test_the_contract_enumerates_EXACTLY_the_rcs_that_reach_not_a_repo` | **RED** | green |
| **NF-2** two-way pin | `test_the_contract_repo_status_TABLE_is_pinned_TWO_WAY_to_REPO_STATUSES` | green* | green |
| **NF-1** `\r\n` forge | `test_a_CRLF_PAIR_cannot_forge_a_LATER_index` | **RED** | green |
| **NF-1** widened parametrisation | `test_the_parser_splits_on_NEWLINE_ONLY_never_str_splitlines` (18 cases) | **RED ×18** | green |
| **NF-1** adjacency is walkable | `test_the_POISON_is_NOT_narrowed_to_ADJACENCY_because_that_is_WALKABLE` | **RED** | green |
| **NF-4** unicode digit | `test_a_UNICODE_DIGIT_index_or_rc_is_unparseable_NEVER_a_ValueError` | **RED** | green |
| **NF-4** helper vs `int()` | `test__is_ascii_int_is_exactly_the_set_int_accepts` | **RED** | green |
| **NF-7c** V1 measurement struck | `test_the_contract_carries_NO_V1_ERA_FLEET_MEASUREMENT_as_current` | **RED** | green |
| **NF-7b** budget numbers pinned | `test_the_contract_docs_budget_numbers_are_PINNED_to_the_constants` | green* | green |
| **NF-6** reserve enforced | `test_the_CH_RESERVE_is_ENFORCED_on_the_client_that_SPENDS_it` | **RED** | green |
| **NF-6** helper | `test__clamp_ch_timeout_never_raises_on_a_client_without_a_conn` | **RED** | green |
| *(self-audit)* every cited test name resolves, in BOTH the doc and the script | `test_every_TEST_NAME_the_docs_and_the_SCRIPT_cite_actually_EXISTS` | **RED** on each surface's own defect | green |

Total at `be0e4d15` with this commit's test file: **28 failed, 775 passed**.
At `6aaf9cf5`: **804 passed** in this file, and **12,184 passed / 1 skipped** across the
whole of `scripts/tests` on the dev host (18:48) — no collateral breakage anywhere.

`*` = green at `be0e4d15` **and that is correct** — these two close a gap rather than fix a
regression. There was no pin, so nothing could be red; both were watched failing against
deliberate mutants instead (below).

## The two-way pin, watched failing in BOTH directions

| mutant | killed by |
|---|---|
| a row **added** to the contract table with no constant (`probably_a_repo`) | `…TABLE_is_pinned_TWO_WAY_to_REPO_STATUSES` |
| a constant **added** to `REPO_STATUSES` with no row (`invented`) | `…TABLE_is_pinned_TWO_WAY_to_REPO_STATUSES` |

The second was initially green *for the wrong reason* — a pre-existing enumeration test
caught it first — so it was re-scored with a `-k` selecting only this guard.

## Mutation results — 24 mutants, 0 survived, 0 invalid

**Sweep 1 (broad, 17 mutants, all KILLED):** poison flag not set / not read / not counted;
`isascii` dropped; `_is_ascii_int` always true; the guard bypassed; the CH clamp not called
/ inert / inverted; the pass-2 host order reversed; the doc's budget number, cap number,
V1 measurement, table row and rc cell; `REPO_STATUSES` grown; plus a known-caught positive
control (`REPO_PROBE_RC_NOT_A_REPO` 128→127).

**Sweep 2 (5 isolated mutants, each scored against its OWN named guard, dying on an
`AssertionError` and not a traceback):**

| mutant | guard that killed it |
|---|---|
| the probe folded back into loop 1 (the faithful regression) | order **and** budget guards |
| **a SIXTH `_read` inserted ahead of the ledger** | order guard |
| `REPO_STATUSES` grew with no doc row | two-way pin |
| a doc row added with no constant | two-way pin |
| the rc cell drops the rc-0 route | rc-enumeration guard |

**Sweep 3 (2 mutants):** the contract citing the old, never-existing test name; and the
script's second-pass comment citing the budget guard with the wrong casing (the ORIGINAL
defect, restored on purpose). Both killed by
`test_every_TEST_NAME_the_docs_and_the_SCRIPT_cite_actually_EXISTS`.

Method: `cp -a` copy with `rm -f <copy>/.git`; `PYTHONDONTWRITEBYTECODE=1` **and**
`__pycache__` purged per mutant; every pattern asserted to match exactly once; restore
verified by SHA-256; collection errors scored INVALID; the verdict reader itself given a
control (it must report SURVIVED / KILLED-BY-ANOTHER-TEST / KILLED-FOR-THE-WRONG-REASON on
synthetic input); control green at the full file before **and** after each sweep.

## Judgement calls

**NF-6 — took "enforce", not "document".** `_clamp_ch_timeout` holds the ClickHouse client
to `COLLECT_CH_RESERVE` at the one place that spends it. Making `COLLECT_BUDGET` read the
runtime value would make the whole budget host-dependent and computed from a file read at
import; clamping is the deterministic version of the same claim, and it converts
`COLLECT_BUDGET = CAP - RESERVE - MARGIN` from a belief into an invariant. A *lower*
`CLICKHOUSE_HTTP_TIMEOUT` is left alone — it only buys headroom. ⚠ The residual the finding
names is stated in the source and not closed: `urlopen`'s timeout is per socket operation,
so a connect-then-stall can still exceed the reserve; `COLLECT_MARGIN` is the only slack.
Re-measured 2026-09-04 on **both** hosts: neither sets the key. That is a fact about
today's env files, which is exactly why it is now a bound and not a note.

**NF-5 — deliberately NOT fixed; NF-1 wins.** A CRLF-carrying reply loses the whole field
(every row `missing`). It is unreachable — `SSH_OPTS:411` allocates no pty — and the
failure direction is honest. Fixing it means accepting a trailing `\r` as transport noise,
which is *precisely* NF-1's forging vector: the measured forge is a record ending in `\r`
followed by an attacker-controlled line. Hardening and forgiving the same byte are not
compatible, and a wrong `repo` is worse than a missing one. Recorded in the contract's
`missing` cell.

**Commit message "eight" vs "nine" breakers — left as is.** The commit is public on a
shared branch and an amend means a force-push that would rewrite a reviewed history for one
word. The *code* was already right; the *test* was the thing actually covering seven of
nine, and that is fixed properly — the parametrisation is now **derived from
`REPO_PROBE_RECORD_BREAKERS`**, so the count can no longer be wrong anywhere.

## What NF-1's fix actually is, and what it deliberately is not

A record carrying a `str.splitlines()` break character now **poisons the remainder of the
reply** — it and every line after it are `unparseable`, so those paths report `missing`.

Narrower rules were considered and rejected as *walkable*:

* "refuse the next line too" (adjacency) — evaded by `\r`, space, `\n`, which puts the
  breaker mid-line while the newline still starts an attacker-controlled record. There is a
  test for this exact string.
* index-equals-position — would close the bare-`\n` case too, but only reduces it from
  "forge any path" to "forge the next path", and it makes the duplicate-index defence
  structurally unreachable, deleting a pinned behaviour for a partial win.

⚠ **Residual, stated rather than papered over:** `parse_repo_probe` is still forgeable
through a bare `\n` if fed directly (the `repo_outputs` seam). Nothing in a parser can tell
two well-formed records apart from one record that was split. Only the **script** can, and
it does (`case $d in *"$NL"*` → rc 201). The "THREE **independent** defences" comment was
therefore false in the direction that gets code deleted; it now reads FOUR **layered**
defences with defence 1 marked **SOLE COVER** for a bare newline.

## Round-3 close-out — two 🟢 findings fixed (`f1568d4f`)

The audit ladder ended clean: round 3 found no regression and returned only these two.
Both are minimal; neither changes the happy path.

### 1. A false equivalence claim that licensed a wrong refactor

`parse_repo_probe` stated that `isascii() and isdigit()` is *"exactly the set `int()`
accepts without a sign"*, and the helper's test carried that sentence in its **name**
(`test__is_ascii_int_is_exactly_the_set_int_accepts`).

**Measured false.** `int()` also converts non-ASCII decimal digits — `'١٢'`, `'１２'`,
`'۱'`, `'٠'`, `'᱀'`, **100+ characters below U+3000** — every one of which the guard
rejects. The test's own body already contradicted its name: it asserts
`_is_ascii_int("１") is False` while `int("１") == 12`. The existing sweep only pins
*guard-accepts ⇒ int-accepts*, which is the one direction a wrong refactor preserves.

**The hazard:** a maintainer who believes the two sets are equal replaces the guard with
`try: int(x) / except ValueError` to remove the duplication, widening the accepted
index/rc alphabet to every Unicode decimal digit **with nothing going red**.

**Fixed:** the comment now states the SUBSET and *why it is correct* — the field is `$n`
out of the probe's POSIX arithmetic, ASCII by construction, so a non-ASCII digit is a
corrupted reply and not an index. The helper's docstring carries the same one-way note.
The test is renamed `test__is_ascii_int_accepts_only_what_int_converts`, which is what
its body pins. **No behaviour change.**

### 2. The CH-reserve clamp was silent

`gather` called `_clamp_ch_timeout(client)` and **discarded the return**. This is the one
place in a file whose stated discipline is *"a read that did not happen is published as
one"* that **changes** a number instead of measuring and publishing it — and it changes a
number the operator set. An operator who put `CLICKHOUSE_HTTP_TIMEOUT=60` in
`~/.config/activity-collector/env` because their CH read is slow got a 15 s read, an
ordinary-looking timeout in the `clickhouse` block, and an env file still saying 60 —
i.e. they debug ClickHouse, which is not where the 15 came from.

**Fixed:** `report["clickhouse"]` now carries

| key | meaning |
|---|---|
| `timeout_secs` | what the read actually ran with |
| `timeout_clamped_from` | the configured value it was reduced **from**; `null` when nothing was reduced |

Both are present in **every** branch (`ok`, `skipped`, `unavailable`) and `null` when no
client carrying a timeout was built — unmeasured, which is not the same claim as "not
clamped", so a consumer never has to tell an absent key from a null one. The `getattr`
chain is factored into `_ch_timeout`, **the one reader**, so the clamp and the field that
publishes it cannot disagree about what they saw. `payload-contract.md` documents both.

### Red/green matrix — every guard watched RED at `6aaf9cf5`

| guard | red at `6aaf9cf5` (its own assertion) | green at `f1568d4f` |
|---|---|---|
| `test_the_ascii_int_guard_is_documented_as_a_SUBSET_not_an_equivalence` | AssertionError: *parse_repo_probe (the call site) still claims the guard is `exactly the set int() accepts` — it is a strict subset* | ✅ |
| …its **name-half**, isolated (old test file + new guard, new script) | `AssertionError: ['test__is_ascii_int_is_exactly_the_set_int_accepts']` | ✅ |
| `test_the_CH_TIMEOUT_CLAMP_is_PUBLISHED_and_never_SILENT` | `KeyError: 'timeout_secs'` | ✅ |
| `test__ch_timeout_is_the_ONE_reader_of_the_clients_timeout` | `AttributeError: module 'session_manager' has no attribute '_ch_timeout'` | ✅ |

⚠ **Residual, stated rather than hidden:** the second half of fix 1's guard is a guard on
**words**, so a *reworded* equivalence claim would walk it. What it cannot walk is the
first half, which is a measurement (the 100+ character sweep), and the requirement that
each surface still say SUBSET.

### The ladder's evidence, re-confirmed after the change

| claim | how | result |
|---|---|---|
| the two-way contract pin fails in **both** directions | added a `bogus_status` row to the real table; separately deleted the `home` row | ✅ red both ways, by the pin's **own** message: `documented-but-not-a-status=['bogus_status']` / `undocumented=['home']` |
| `_is_ascii_int` reverted to bare `.isdigit()` still kills its two tests | mutated the helper in an isolated `cp -a` copy | ✅ `test_a_UNICODE_DIGIT_index_or_rc_is_unparseable_NEVER_a_ValueError` dies with the exact `ValueError: invalid literal for int() with base 10: '²'` it exists to prevent; `test__is_ascii_int_accepts_only_what_int_converts` dies on `_is_ascii_int('²') is False` |
| `test_the_repo_probe_is_issued_AFTER_every_other_read_on_every_host` still dies when a probe `_read` moves into pass 1 | inserted `_read(repo_probe_argv([...]), host, …)` after `wins_res` | ✅ red on its **own** claim 2: `a repo probe (index 2) is issued before another read (index 9)` |

Mutation protocol: `cp -a` copy with `rm -f <copy>/.git` first, `PYTHONDONTWRITEBYTECODE=1`,
`__pycache__` purged per mutant, every mutant restored **by md5 hash**, control run green
before each battery. The real worktree's hashes were re-checked unchanged afterwards.

## Both tiers, on the MERGED tree `18601071` (= `origin/main` `d86b4e45` × head `f1568d4f`)

`origin/main` re-fetched immediately before gating; it was still `d86b4e45`. The merge was
clean — the only overlap is a comment block in `nix/home.nix` (four→five ssh invocations),
read and confirmed coherent rather than assumed from a clean `git merge`.

🔴 **Sandbox tier — the one that gates. Built ONE AT A TIME, both green.** Numbers read
from the derivation logs (`-L`), not from an exit code:

| derivation | result |
|---|---|
| `nix build .#checks.x86_64-linux.pytests` | `RESULT: PASS (exit=0)` — **collected 21,261, passed 21,258, skipped 3, failed 0** (floor 18,404 = sum of 30 per-target floors); `panic: test timed out` × **0** |
| `nix build .#checks.x86_64-linux.nodetests` | `RESULT: PASS (exit=0)` — suites 5, files 41, **tests 1,449, pass 1,449, fail 0, skipped 0** (floor 1,367); `not ok` × **0**, `panic: test timed out` × **0** |

Neither was the cached-and-silent case: pytests emitted 1,505 log lines, nodetests 437 KB
of TAP, both from a real `building '…drv'`.

**Dev-host tier — `scripts/gate.sh --tier both`:** `GATE: RESULT=FAIL exit=1`, node
**PASS** (1,449/1,449), pytest **21,257 passed / 1 failed** of 21,261 collected.

| dev-host failure | attribution (my own control, not an inherited claim) | sandbox |
|---|---|---|
| `test_a_body_file_written_by_a_heredoc_on_the_same_line_is_read` | `/tmp/body.md` **exists on this host** — 21,181 B, mtime 2026-09-03 19:41, content is another session's PR body about a "mute-review queue". The test writes a heredoc to that path; the guard reads the real file instead. | `test_clawgate_task_interview_guard.py` **310 collected, 310 passed, 0 failed** |

The second historically-known dev-host failure (the browser-bridge test whose stderr names
a missing npm-install) **did not occur this run** — `scripts/browser-bridge/tests` was
890/890 in the dev-host tier and 890/890 in the sandbox. No other failures appeared.

Neither failure is reachable from this diff, which touches `scripts/session-manager`, its
test file and one skill reference doc. `scripts/tests` itself is **12,258 collected /
12,258 passed / 0 failed** in both tiers.

⚠ Branch protection on `main` is **declared off** in this repo, so these two local tiers are
the only gate — run on the merged tree `18601071`, base `d86b4e45`, tiers named above.

### DIFF
--- a/claude/skills/session-manager/SKILL.md
+++ b/claude/skills/session-manager/SKILL.md
@@ -33,6 +33,7 @@ column — **how it is counted**) are different, and `caveats.kind_scope` carrie
 | `--no-ch` | skip ClickHouse — drops the **largest non-row block** (17.4 KB below), which answers a different question |
 | `--no-capture` | skip the pane scrape; **every** `waiting_probable` AND `unsent_prompt` becomes `null` (both roll-ups `null`, never `0`) |
 | `--no-ledger` | skip the ledger read → **no age, no session id** on any row |
+| `--no-repo` | skip the per-host repo probe; every `repo` becomes `null` with `repo_status: skipped` — never `not_a_repo` |
 | `--fuzzyclaw` | the task-file join, **OFF by default** (see below) |
 
 `--json`, `--no-fuzzyclaw`, `--plain`, `--stale-threshold`, `--lines`, and what each drops:
--- a/claude/skills/session-manager/reference/payload-contract.md
+++ b/claude/skills/session-manager/reference/payload-contract.md
@@ -227,6 +227,46 @@ repo resolve to the SAME label, so quote both. New non-scratch sessions get a re
 at creation from `scripts/tmux-autoname-session.sh`; `label` is what covers the ones that
 predate it.
 
+## 🔴 `repo` — the PROJECT key, and it is not `label`
+
+`label` answers "where is this" for an OPERATOR (the leaf of the cwd, or a codename).
+`repo` answers "which project is this" for a CONSUMER: the **MAIN CLONE** of the pane's
+cwd, so **every linked worktree of a repo carries ONE name**. That is a thing no string
+operation on `path` can produce — a worktree called `clawgate-extension` is a worktree of
+`homelab-talos`, and `ht-r11-930492` is not derivable from `homelab-talos` at all.
+
+Its consumer is clawgate's tmux page, whose `projectOf()` prefers `repo` when non-empty,
+falls back to the LEAF of `path`, then to `Other`. Before this field existed the leaf won
+every time, so each worktree formed its own project group.
+
+🔴 **IT IS RESOLVED ON THE HOST THAT OWNS THE DIRECTORY**, by one batched `sh -c` per host
+over the same SSH transport as the tmux calls — `git rev-parse --path-format=absolute
+--git-common-dir`, whose parent directory is the main clone. Roughly half these rows come
+off the laptop, and a LOCAL `git rev-parse` would answer about whatever happens to sit at
+that path on the workbench, or about nothing. `--show-toplevel` is the wrong flag: on a
+linked worktree it returns the worktree itself, so it cannot group two worktrees at all.
+
+🔴 **`repo_status` IS WHAT MAKES A NULL `repo` READABLE.** Never read a null as "not in a
+repo" without it:
+
+| `repo_status` | means |
+|---|---|
+| `ok` | resolved; `repo` names the main clone. The ONLY status carrying a name. |
+| `not_a_repo` | MEASURED: the owning host says this path is not in a work tree (or its common dir cannot be honestly named — a bare repo, a submodule). |
+| `home` | MEASURED and deliberately withheld: the main clone IS the owning host's `$HOME`. `projectOf()` routes an unparented shell to `Other`, and `repo` is the branch it PREFERS, so a name here would override that guard. |
+| `no_path` | the pane reported no cwd. |
+| `missing` | the probe answered for this host but not for this path (a partial reply). |
+| `unmeasured` | **nobody looked** — the probe failed, timed out, or came back without its sentinel. NOT a measured absence. |
+| `skipped` | `--no-repo`. |
+
+Per host, `repos_measured` / `repos_status` / `repos_error` / `repos_paths` /
+`repos_resolved` / `repos_unparseable` are the FOURTH independent measurement beside
+`reachable`, `windows_measured` and `captures_measured`. The counts are integers only when
+`repos_measured` is true; otherwise they are null, never 0.
+
+Measured on the live fleet 2026-09-03: 90 of 92 windows resolved (laptop 31/32, workbench
+59/60); the two that did not are `/home/zach` and a non-repo directory, both `not_a_repo`.
+
 ## The caveats are in the OUTPUT, not just in this file
 
 `report["caveats"]` (structured) + one footer line each in the table, printed
--- a/claude/skills/session-manager/reference/payload-contract.md
+++ b/claude/skills/session-manager/reference/payload-contract.md
@@ -22,6 +22,7 @@ _Moved verbatim out of `SKILL.md` on 2026-08-21, when the body was cut from 23,2
 | `--pane-preview` | publish each Claude pane's **visible screen** as `pane_preview`. Costs no extra tmux work (the capture already runs — `waiting`/`unsent` are derived from it and it used to throw the screen away), but makes the document **2.63x** larger: measured live back to back, 122,731 B without / 322,204 B with. Off by default for that reason; `--lean` drops it, so do not pass both |
 | `--fuzzyclaw` / `--no-fuzzyclaw` | the task-file join is **OFF by default** (see below) |
 | `--no-ledger` | skip the per-host agent-ledger read. Rows then have **no age and no session id** — the #419 view, reproducible on demand |
+| `--no-repo` | skip the per-host repo probe (one batched `sh -c` per host, the FIFTH ssh call). Every row's `repo` becomes `null` with `repo_status: **skipped**` and every per-host `repos_*` count becomes `null` — never `not_a_repo`, never `0`. This is the cost control for the probe; the field is otherwise ON by default |
 | `--plain` | `tail` only: strip ANSI at the source instead of `sed`-ing it out |
 | `--stale-threshold <secs>` | default 3600; `age >= threshold` is stale |
 | `--lines N` | `tail` scrollback depth (default 100) |
@@ -252,17 +253,30 @@ repo" without it:
 | `repo_status` | means |
 |---|---|
 | `ok` | resolved; `repo` names the main clone. The ONLY status carrying a name. |
-| `not_a_repo` | MEASURED: the owning host says this path is not in a work tree (or its common dir cannot be honestly named — a bare repo, a submodule). |
+| `not_a_repo` | MEASURED: the owning host's `git` exited **128** for this path — it is not in a work tree (or its common dir cannot be honestly named: a bare repo, a submodule). 🔴 It requires that exit status. Until 2026-09-04 every OTHER failure — `git` timing out, `git` absent, an unknown flag — arrived here too, so a real repo could read as a measured absence. |
 | `home` | MEASURED and deliberately withheld: the main clone IS the owning host's `$HOME`. `projectOf()` routes an unparented shell to `Other`, and `repo` is the branch it PREFERS, so a name here would override that guard. |
 | `no_path` | the pane reported no cwd. |
-| `missing` | the probe answered for this host but not for this path (a partial reply). |
-| `unmeasured` | **nobody looked** — the probe failed, timed out, or came back without its sentinel. NOT a measured absence. |
+| `missing` | the probe answered for this host but not for this path — a partial reply, or a record refused because it carried a character `str.splitlines()` would treat as a line break. |
+| `unmeasured` | **nobody looked.** TWO granularities, and both are real: the whole HOST's probe failed / timed out / came back without its sentinel, **or** THIS path's `git` exited non-zero for a reason that is not 128 (the per-path `timeout` fired, `git` is not installed, the flag is unsupported). NOT a measured absence either way. |
 | `skipped` | `--no-repo`. |
 
 Per host, `repos_measured` / `repos_status` / `repos_error` / `repos_paths` /
 `repos_resolved` / `repos_unparseable` are the FOURTH independent measurement beside
 `reachable`, `windows_measured` and `captures_measured`. The counts are integers only when
 `repos_measured` is true; otherwise they are null, never 0.
+
+🔴 **A host can be honestly `repos_measured: true` while one of its ROWS is `unmeasured`,
+and reading only the host block hides that.** The probe answering is a fact about the
+probe; each path's `git` exit status is a separate fact about that path.
+`repos_resolved` counts `ok` rows and nothing else, so a per-path `unmeasured` can never
+inflate it — but `repos_paths - repos_resolved` is **not** a count of non-repo panes.
+
+🔴 **The whole collection is bounded by `COLLECT_BUDGET` (70 s), not by the per-call
+timeouts.** `tmux-snapshot-push.sh` wraps the collector in `timeout 90`, and rc 124 there
+is `exit 3` — no snapshot for EITHER host. Five reads per host plus the ClickHouse query
+sum past that, so a read with no budget left is **not attempted** and its host reports
+`unmeasured` with an error naming the budget. That is a read that did not happen, and it
+is published as one.
 
 Measured on the live fleet 2026-09-03: 90 of 92 windows resolved (laptop 31/32, workbench
 59/60); the two that did not are `/home/zach` and a non-repo directory, both `not_a_repo`.
--- a/claude/skills/session-manager/reference/payload-contract.md
+++ b/claude/skills/session-manager/reference/payload-contract.md
@@ -278,14 +278,18 @@ sum past that, so a read with no budget left is **not attempted** and its host r
 `unmeasured` with an error naming the budget. That is a read that did not happen, and it
 is published as one. *(Both numbers above are pinned to `COLLECT_BUDGET` / `COLLECT_CAP` by
 `test_the_contract_docs_budget_numbers_are_PINNED_to_the_constants`; the count of five
-reads and their ORDER are pinned by
-`test_the_per_host_call_ORDER_puts_the_optional_probe_LAST`.)*
+reads and their ORDER by
+`test_the_repo_probe_is_issued_AFTER_every_other_read_on_every_host`.)*
 
-🔴 **ORDER MATTERS AND IT IS PART OF THE CONTRACT.** The reads share one deadline, so the
-LAST call on the LAST host is the one that starves. The **agent ledger is read before the
-repo probe**, because the probe is the optional one (`--no-repo`) while a starved ledger
-costs every row on that host `age_secs`, its `stale` bucket and `claude_session_id`. With
-both hosts at their bounds, the laptop's `repo` is the field that is dropped — deliberately.
+🔴 **ORDER MATTERS AND IT IS PART OF THE CONTRACT.** One deadline spans ALL hosts, so the
+LAST call issued is the one that starves. The repo probe is therefore a **second pass over
+the hosts**: every pre-existing read — panes, windows, capture, ledger — is issued for
+EVERY host before the probe is issued for ANY host. The probe is the read a consumer can
+turn off (`--no-repo`) with the payload still whole, while a starved ledger silently costs
+every row on that host `age_secs`, its `stale` bucket and `claude_session_id`. Moving the
+probe later *within* a host is not enough and was measured so: the local host's own 15 s
+probe alone pushed the remote host past the deadline. With both hosts at their bounds it is
+the laptop's `repo` that is dropped — deliberately.
 
 ⚠ **There is no current fleet measurement of the resolve rate.** The only one taken (90 of
 92 windows, 2026-09-03) was measured under protocol **V1**, which collapsed every `git`
--- a/claude/skills/session-manager/reference/payload-contract.md
+++ b/claude/skills/session-manager/reference/payload-contract.md
@@ -281,6 +281,17 @@ is published as one. *(Both numbers above are pinned to `COLLECT_BUDGET` / `COLL
 reads and their ORDER by
 `test_the_repo_probe_is_issued_AFTER_every_other_read_on_every_host`.)*
 
+🔴 **The ClickHouse read's own timeout is CLAMPED to `COLLECT_CH_RESERVE` (15 s), and the
+clamp is PUBLISHED.** `CHConn.from_env` resolves `CLICKHOUSE_HTTP_TIMEOUT` out of the
+collector's per-HOST env file, so an operator can configure a value the budget arithmetic
+cannot afford — and this is the one place in the collector that CHANGES a number instead of
+measuring one. `clickhouse.timeout_secs` is what the read actually ran with;
+`clickhouse.timeout_clamped_from` is the configured value it was reduced FROM, and is
+**`null` when nothing was reduced**. Both are `null` when no client carrying a timeout was
+built (`skipped`, `unavailable`, an injected double) — unmeasured, which is not the same
+claim as "not clamped". Without them a 60 s setting produced a 15 s read that read as an
+ordinary ClickHouse timeout, and the operator debugs ClickHouse.
+
 🔴 **ORDER MATTERS AND IT IS PART OF THE CONTRACT.** One deadline spans ALL hosts, so the
 LAST call issued is the one that starves. The repo probe is therefore a **second pass over
 the hosts**: every pre-existing read — panes, windows, capture, ledger — is issued for
--- a/claude/skills/session-manager/reference/payload-contract.md
+++ b/claude/skills/session-manager/reference/payload-contract.md
@@ -253,11 +253,11 @@ repo" without it:
 | `repo_status` | means |
 |---|---|
 | `ok` | resolved; `repo` names the main clone. The ONLY status carrying a name. |
-| `not_a_repo` | MEASURED: the owning host's `git` exited **128** for this path — it is not in a work tree (or its common dir cannot be honestly named: a bare repo, a submodule). 🔴 It requires that exit status. Until 2026-09-04 every OTHER failure — `git` timing out, `git` absent, an unknown flag — arrived here too, so a real repo could read as a measured absence. |
+| `not_a_repo` | MEASURED: `git` ANSWERED for this path and what it answered does not name a clone. 🔴 **TWO exit statuses reach it, not one** — **128** (git: not in a work tree) *and* **0 with a common dir whose parent is not a clone root** (empty, a bare repo `/srv/foo.git`, a submodule `<super>/.git/modules/<n>`); naming those would publish a confident wrong group, so they fail soft here. This cell used to say it "requires" 128, which contradicted its own parenthetical and was measured false. What it excludes is a **failure** to measure: until 2026-09-04 `git` timing out, `git` absent or an unknown flag arrived here too, so a real repo could read as a measured absence. |
 | `home` | MEASURED and deliberately withheld: the main clone IS the owning host's `$HOME`. `projectOf()` routes an unparented shell to `Other`, and `repo` is the branch it PREFERS, so a name here would override that guard. |
 | `no_path` | the pane reported no cwd. |
-| `missing` | the probe answered for this host but not for this path — a partial reply, or a record refused because it carried a character `str.splitlines()` would treat as a line break. |
-| `unmeasured` | **nobody looked.** TWO granularities, and both are real: the whole HOST's probe failed / timed out / came back without its sentinel, **or** THIS path's `git` exited non-zero for a reason that is not 128 (the per-path `timeout` fired, `git` is not installed, the flag is unsupported). NOT a measured absence either way. |
+| `missing` | the probe answered for this host but not for this path — a partial reply, or a record refused because it carried a character `str.splitlines()` would treat as a line break. 🔴 Such a record also **poisons every record after it**: once a non-`\n` line break is inside a record, no later boundary in that reply is knowable, so the rest of the host's paths are `missing` too. Honest and bounded — never a wrong name. |
+| `unmeasured` | **nobody looked.** TWO granularities, and both are real: the whole HOST's probe failed / timed out / came back without its sentinel, **or** THIS path's `git` exited non-zero for a reason that is not 128 — the per-path `timeout` fired (124), `git` is not installed (127), the flag is unsupported (129), or the probe itself **refused to transmit** a common dir containing a newline (**201**, `REPO_PROBE_RC_NEWLINE` — see `missing` for the sibling case the parser refuses). NOT a measured absence either way. |
 | `skipped` | `--no-repo`. |
 
 Per host, `repos_measured` / `repos_status` / `repos_error` / `repos_paths` /
@@ -276,10 +276,22 @@ timeouts.** `tmux-snapshot-push.sh` wraps the collector in `timeout 90`, and rc 
 is `exit 3` — no snapshot for EITHER host. Five reads per host plus the ClickHouse query
 sum past that, so a read with no budget left is **not attempted** and its host reports
 `unmeasured` with an error naming the budget. That is a read that did not happen, and it
-is published as one.
+is published as one. *(Both numbers above are pinned to `COLLECT_BUDGET` / `COLLECT_CAP` by
+`test_the_contract_docs_budget_numbers_are_PINNED_to_the_constants`; the count of five
+reads and their ORDER are pinned by
+`test_the_per_host_call_ORDER_puts_the_optional_probe_LAST`.)*
 
-Measured on the live fleet 2026-09-03: 90 of 92 windows resolved (laptop 31/32, workbench
-59/60); the two that did not are `/home/zach` and a non-repo directory, both `not_a_repo`.
+🔴 **ORDER MATTERS AND IT IS PART OF THE CONTRACT.** The reads share one deadline, so the
+LAST call on the LAST host is the one that starves. The **agent ledger is read before the
+repo probe**, because the probe is the optional one (`--no-repo`) while a starved ledger
+costs every row on that host `age_secs`, its `stale` bucket and `claude_session_id`. With
+both hosts at their bounds, the laptop's `repo` is the field that is dropped — deliberately.
+
+⚠ **There is no current fleet measurement of the resolve rate.** The only one taken (90 of
+92 windows, 2026-09-03) was measured under protocol **V1**, which collapsed every `git`
+failure into `not_a_repo` — i.e. under the exact bug the rc field was added to fix — so its
+`not_a_repo` counts are not comparable to what V2 publishes. It has been struck rather than
+carried forward; re-measure under V2 before quoting a rate.
 
 ## The caveats are in the OUTPUT, not just in this file
 
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -3470,9 +3470,10 @@ in
   #
   # ⚠ THE PER-RUN COST IS MORE THAN ONE SSH, and this comment used to say "~720
   # ssh round trips a day" as though it were. `session-manager --json` makes up
-  # to FOUR ssh invocations to the laptop per run (list-panes, list-windows, the
-  # capture batch, the ledger read) with no ControlMaster, so it is nearer
-  # ~2,880 handshakes/day; it also runs one ClickHouse query per run (~720/day),
+  # to FIVE ssh invocations to the laptop per run (list-panes, list-windows, the
+  # capture batch, the repo probe, the ledger read) with no ControlMaster, so it
+  # is nearer ~3,600 handshakes/day; it also runs one ClickHouse query per run
+  # (~720/day),
   # which the old figure omitted entirely. Store side: ~720 upserts/day of
   # ~100 KB jsonb onto a two-row table, i.e. real TOAST/WAL churn for data that
   # currently has no reader. If that cost ever matters, `--lean` is the lever —
--- a/scripts/session-manager
+++ b/scripts/session-manager
@@ -201,11 +201,32 @@ and the roll-ups a caller usually wants are:
                                         which skill owns each
 
 Each row carries: kind, host, session, window_index, window_id, window_name,
-codename, label, label_source, hotkey, hotkey_display, pane_id, path, command,
+codename, label, label_source, hotkey, hotkey_display, pane_id, path, repo,
+repo_status, command,
 task, claude, busy, age_secs, age_source, status, waiting_probable,
 waiting_signals, waiting_status, unsent_prompt, unsent_prompt_status,
 pane_preview, pane_preview_status, claude_session_id, runtime, ledger,
 fuzzyclaw, panes. The row ledger is pinned by a test.
+
+🔴 `repo` IS THE PROJECT KEY, AND IT IS RESOLVED ON THE HOST THAT OWNS THE
+DIRECTORY
+---------------------------------------------------------------------------
+`label` answers "where is this" for an operator: the LEAF of the pane's cwd, or
+a scratch-slot codename. `repo` answers "which project is this" for a consumer:
+the MAIN CLONE, via `git rev-parse --git-common-dir` — so every linked worktree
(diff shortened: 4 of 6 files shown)


ZacxDev:
Round 1 audit findings and the fixes claimed for them. Posted as an issue comment (not a review) so the ladder's own parser can read it.

```audit-claims round=1 audited=5d722f2f..be0e4d15
1. F1 — a per-path `git` failure that is NOT "not in a work tree" (timeout, `git` absent, any other rc) no longer publishes the MEASURED status `not_a_repo`; the remote script now carries git's exit code as a third field and anything other than 128 becomes a per-path `unmeasured` that cannot inflate `repos_resolved`.
2. F2 — a forged or duplicate reply index can no longer overwrite another row's answer: duplicate in-range indices are counted `unparseable`, and the `str.splitlines()` character class (eight separators the writer never used) is handled, which duplicate rejection alone structurally could not cover.
3. F3 — the row lookup now strips the path so it matches the probe key, which also makes commit `5d722f2f`'s stated claim ("the row-level `missing` branch is unreachable from `gather`") true rather than false.
4. F4 — the collection envelope is bounded by an explicit `COLLECT_BUDGET` monotonic deadline enforced inside `run_tmux`, so the worst case stays inside `tmux-snapshot-push.sh`'s 90s cap; the stale `SSH_TIMEOUT * 2 <= 30` guard is replaced by one deriving the real call count and budget.
5. F5 — the probe timeout literal, the `command -v timeout` guard and `repos_unparseable` are now pinned behaviourally; all three previously SURVIVED the full file.
6. F6/F9/F10 — `--no-repo` documented in both flag tables; `make_runner` raises on an unrecognised argv with a reachable case; the sentinel docstring's false "breaks the fixture loudly" claim corrected.
```

Two things the fix round found that the audit did not, both recorded because they changed the fix:

- **F4's arithmetic omitted ClickHouse (15s).** The base was at **83s of 90**, not 68 — 7s of headroom, not 22 — so "a smaller probe budget" would have needed ≤3.5s. Hence the explicit monotonic envelope instead, which also covers a *sixth* call with no arithmetic.
- **F2 needed a third defence.** Duplicate-index rejection plus newline stripping still left `str.splitlines()` splitting on eight characters; measured, a `\x0b` forged a *later* index's answer, which duplicate rejection cannot reach.

⚠ `origin/main` moved to `d8fe0bce` after these gates finished and is not in any run recorded here. Whoever merges must re-gate on the merged tree.

ZacxDev:
Round 2 (delta) findings and the fixes claimed for them.

```audit-claims round=2 audited=be0e4d15..6aaf9cf5
1. NF-3 — the new 5th call starved the pre-existing ledger read under budget pressure. 🔴 The prescribed fix (read the ledger before the probe, within a host) was MEASURED INSUFFICIENT — the deadline spans all hosts, so the local host's own probe alone pushes the remote past the budget. Shipped instead: a TWO-PASS over hosts — every pre-existing read (panes, windows, capture, ledger) for EVERY host before the probe for ANY host — so the optional field is what starves, by construction.
2. NF-1 — `\r\n` forged a later index through `parse_repo_probe`, and the "three INDEPENDENT defences" comment was false. The parametrised guard is widened from single characters to sequences and adjacency-walkability. 🔴 The residual is stated rather than closed: a bare `\n` can still forge if the parser is fed directly, so only the script can prevent it and its guard is now marked SOLE COVER.
3. NF-2 — `not_a_repo` was documented as requiring git rc 128 while three rc-0 routes reach it (bare repo, submodule). The prose is corrected in both the code comment and the agent-facing contract, and the contract's `repo_status` table is now pinned TWO-WAY against `REPO_STATUSES` — the gap the previous round filed as not-done, which is why the false row shipped.
4. NF-4 — `str.isdigit()` is true for `²`/`①` but `int()` rejects them, so the parser could raise `ValueError` where its docstring promised `unparseable`; an unhandled exception in `gather` reaches `tmux-snapshot-push.sh` as exit 3, i.e. no snapshot for either host. Now `_is_ascii_int`.
5. NF-6 — `COLLECT_CH_RESERVE` and its guard both read a class default the runtime can override via `CLICKHOUSE_HTTP_TIMEOUT`. Enforced rather than documented: `_clamp_ch_timeout` holds the CH client to the reserve where it is spent. Both hosts re-measured — neither sets the variable.
6. NF-7 — the contract's `unmeasured` list omitted the rc-201 newline refusal; `COLLECT_BUDGET` was restated as a literal with nothing pinning it; and a "measured on the live fleet" line taken under the V1 protocol (i.e. under the exact bug F1 fixed) was carried as current. First two fixed and pinned, third struck.
7. NF-5 — DELIBERATELY NOT FIXED, with the reason recorded: forgiving a trailing `\r` is exactly NF-1's forging vector. Unreachable today (no pty is allocated), and it fails honestly.
```

Method notes worth carrying:

- **24 mutants, 0 survived, 0 invalid.** Sweep 2 re-scored five of them against their **own named guard**, requiring an `AssertionError` rather than any failure — **two had been green for the wrong reason**, killed by a different guard's error. The two-way pin was watched failing in **both** directions. `cp -a` + `rm -f .git`, `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` purged per mutant, every pattern asserted to match exactly once, SHA-256 restore, and the verdict reader itself controlled.
- **Two of the round's own prose defects were found and fixed mid-round:** the contract cited a test name that never existed, and the guard written to catch that class was blind to the script's own comments — the same defect one level down. Both surfaces are now pinned to the module's symbol table.
- The commit message says "eight" record-breaker characters where the code says nine; the message is left alone, because the real defect was the *test* covering 7 of 9, and it is now **derived from `REPO_PROBE_RECORD_BREAKERS`** so the count cannot be wrong again.

Gates on merged sha `33079a6b` (`origin/main` `d86b4e45` × `6aaf9cf5`), sandbox tier built one at a time: `pytests` **PASS** 21,255 passed / 0 failed; `nodetests` **PASS** 1,449/1,449. The two dev-host failures were re-attributed with fresh controls — `/tmp/body.md` genuinely exists (21,181 B, another session's content), and the browser test's stderr names a missing npm-install rather than load; both pass in the sandbox (310/310 and 890/890).

PR #1283, opened 2026-09-04
fix(mention-open): the repo mapping generator was lossy and the API fallback was inert

ZacxDev:
## What

`talos-infra#1065` only resolved when the repo happened to be checked out under `~/workspace`. This adds two more owner sources — a generated static mapping of every repo the operator can see, and an exact-name GitHub search as a last resort — plus the two defects found while exercising them.

## The two defects (both measured)

**1. `regen-known-repos.sh` was a lossy generator.** It lowercased keys only for LOCAL checkouts, so simply running it dropped 33 lowercase aliases the committed mapping had (`comfyui`, `civitui`, `spm`, `gcodeide`, `idm-vton`, …) — while its own header comment claimed *"Keys are lowercased"*. `mention_scan._resolve_repo` does an exact `repos.get(repo)` with **no case folding anywhere**, so those aliases are the only thing making `comfyui#100` resolve. Both spellings are now emitted for API-sourced repos too.

Re-measured after the fix: keys 454 → 420, **0 real repos lost, 0 values changed**. The only losses are 34 transient agent-worktree directory names (the generator now skips linked worktrees — their `.git` is a *file* — since they share the base clone's remote and add no owner); the only gain is `civitai-feed`.

**2. The GitHub API fallback shipped INERT.** Its jq filter lived in a Python string with no `f` prefix, so it compared `.name` against the literal text `"$name"` and selected nothing, for any input, ever. Verified against the real jq binary:

```
$ jq -r '.items[] | select(.name == "$name") | .full_name' < payload.json
        # 0 lines, rc 0
```

`gh api --jq` accepts no `--arg`, so the fix moves the name match into Python and leaves the jq program a constant (`.items[].full_name`) — the whole interpolation class is gone. It also now takes the exact (case-insensitive) name match instead of `splitlines()[0]`: search results are relevance-ordered, so the old first-row rule would have opened a stranger's repo under the operator's repo name had the filter ever worked.

Before → after, live:

| click | before | after |
|---|---|---|
| `kubernetes#1` | cannot resolve | `kubernetes/kubernetes` |
| `nixpkgs#1` | cannot resolve | `NixOS/nixpkgs` |
| `zzz-not-a-real-repo-xyz#1` | cannot resolve | cannot resolve (unchanged — nothing is guessed) |

## Tests

8 new. **7 are red against the pre-fix `_gh_api_repo_search`** (restored into a scratch copy and run) and green here; the 8th is labelled in its own docstring as an invariant guard, not regression coverage.

Two of them pin the **seam** rather than the component, because every stubbed-`gh` test is structurally incapable of catching the original bug — with `gh` faked, a filter that selects nothing in production still passes, which is exactly how it shipped. One asserts the jq program contains neither `$` nor the searched name (so jq *cannot* be the selector); the other runs the emitted program through the real `jq` binary.

That second test was itself vacuous when first written: `monkeypatch.setattr(MO.subprocess, "run", fake)` patches the *shared* `subprocess` module, so its own `jq` call reached the fake and it passed against the inert original. Caught by running it against the pre-fix code; fixed with a `_REAL_RUN` bound at import time.

## Gate

Run on the **merged tree** (`integ/mention-merge` = `feat/mention-system-repos` merged onto `origin/main` `a7dac5bd`), both tiers, the two nix derivations built **one at a time**:

- `nix build .#checks.x86_64-linux.pytests` — collected 21044, **0 failed**, rc 0
- `nix build .#checks.x86_64-linux.nodetests` — 1449 tests, **0 failed**, rc 0
- dev-host tier: node PASS (1449); pytest 2 failures, both accounted for and neither reachable from this diff — `test_clawgate_task_interview_guard.py:429` **fails identically on untouched `origin/main`**, and browser-bridge `test_agent_without_any_tab_is_untouched` is a load-sensitive 300s timeout that passes alone at 223s.

## Not covered

`known_repos.py` is a generated snapshot with no automated refresh — a repo created after the last regen now resolves only via the (working) API fallback. Filed as a ranked next step in `claudedocs/handoff-mention-system-repos.md`.


ZacxDev:
Closing without merging: the generated repo mapping must not be tracked in this repository. Reworking so the mapping is produced per-host and untracked.

Issue #1284, opened 2026-09-04
browser agent --dry-run runs a full model session (90s warm + up to 120s) — name vs cost


ZacxDev:
## What

`browser agent --dry-run "<goal>"` runs a **full opencode model session** — and on a cold cache pays a 90 s warm first. A flag named `--dry-run` reads as "validate and tell me what you would do"; this one costs up to ~210 s and real model tokens.

## Measured (2026-09-04)

Timings from `pytest --durations` on `scripts/browser-bridge/tests/test_browser_tab_ref.py`, where each case invokes the real wrapper under a fresh `HOME`:

| invocation | wall time |
|---|---|
| `agent --dry-run <goal>` | **155.98 s** |
| `agent --dry-run <goal>` (second pass via `--continue`) | **234.22 s** |
| every non-agent test in the same file | ~0.5 s |

Direct reproduction of the warm half, fresh `HOME`, no goal:

```
browser-agent: WARNING: could not warm the isolated opencode config dir
'…/.cache/browser-agent-opencode-config' (opencode …, timeout 90s).
… the warm step needs to npm-install @opencode-ai/plugin into it
```

Two costs compose: `OC_WARM_TIMEOUT` (default **90 s**, unreachable without a writable dir + network) and the agent's own `--timeout` (default **120 s**) around `run_opencode`.

## Why this may be intended — read before "fixing" it

`BROWSER_AGENT_DRY_RUN` is a **tool-level** flag, not a wrapper-level one. `browser-agent` only exports it into opencode's environment (the `BROWSER_AGENT_DRY_RUN=` line in `run_opencode()`), and `browser_tool_impl.mjs` reads it to suppress *mutating* browser ops. `README.md` states `emulate` counts as MUTATING for it, and `reference/agent.md` says it skips the auto-wake.

So "the model still reasons, browser mutations are suppressed" is a coherent design. **This issue is about whether the flag's NAME and cost match that design**, not a claim that the behaviour is a bug. Do not short-circuit `--dry-run` without deciding that question — it would change shipped semantics that three docs describe.

## Not fixed by the ordering change

PR (this session) moved argument validation above the warm block, so a *refused* invocation now costs 9 ms instead of 91 s. That does **not** touch this: with a goal present, the warm and the model run both still happen.

## Closing condition

`browser agent --dry-run "<goal>"` either returns in under 5 s on a cold config dir, **or** its `--help` and `README.md` state that it runs a full model session and roughly what that costs.

[...]

PR #1285, opened 2026-09-04
fix(browser-agent): validate args before the warm block — 508s -> 29s, and the CI flake it caused

ZacxDev:
## What this is

Rank 1 from `claudedocs/handoff-signal-skill-eval.md` was *"add the per-test `pytest-timeout` budget `cli_budget.py` prescribes"*. **That turned out to be the wrong fix**, so this PR does something else — the reasoning is below, because the divergence matters more than the diff.

## The flake

`tekton/devrc-pytests` failed on devrc#1271 with a docs-only diff. The traceback ends in `subprocess.py:1209 in communicate` under `subprocess.run(["bash", str(CLI), *args], timeout=CLI_TIMEOUT_S)` — a `TimeoutExpired`, so **no assertion ever executed**. That reads as a code failure and is not one.

It reproduces locally on identical code: three runs of `scripts/browser-bridge/tests` gave **0, 2 and 1** failures.

## Why `pytest-timeout` was the wrong remedy

`cli_budget.py` prescribes it for a different failure — stopping a run of hangs from hitting the CI Task ceiling and aborting the pipeline. It would fire on a starved-but-healthy test too: it changes *which* timeout reports, not whether one does. It would have masked what follows.

## Actual root cause

`browser-agent` parsed its arguments **~300 lines below** the config-dir warm/bootstrap block. So an invocation it was always going to refuse ran the warm first. That warm is bounded by `OC_WARM_TIMEOUT` (90 s) and needs a writable dir plus network to npm-install `@opencode-ai/plugin`; under a pytest `tmp_path` HOME it can do neither, so it burned the full budget every time.

`--durations` on `test_browser_tab_ref.py` — bimodal, not load:

| test | before |
|---|---|
| `test_agent_without_any_tab_is_untouched` | **234.22 s** |
| `test_the_free_text_list_covers_agent_too` (the CI failure) | **155.98 s** |
| `test_POSITIVE_CONTROL_the_missing_goal_error_is_producible` | **90.10 s** |
| the other 51 tests | ~0.5 s each |

480 of the file's 509 s sat in three tests. Two of them also ran a **real opencode model session** (`--timeout` 120 s, twice via `--continue`). Against `CLI_TIMEOUT_S = 300` that left a 66 s margin, and CI load erased it.

## The fix

**1. Ordering** (`browser-agent`). Move arg parsing + validation above the warm block — the same fix, for the same reason, already applied to the `command -v opencode` preflight.

Measured, fresh HOME, no goal: **91,000 ms → 9 ms**, same `rc=2`, byte-identical error. `--steps`, `--timeout`, unknown-flag and two-goals refusals now take 5–9 ms each.

Verified I reordered rather than skipped: with a goal present the warm block **still runs** (its warning still appears).

**2. Stub** (`test_browser_tab_ref.py`). The two goal-supplying tests now run beside a stub `browser-agent` that echoes the goal, so they assert **positively** that the reference arrived instead of that two error strings are absent. The positive control still drives the **real** wrapper — it costs 9 ms now.

| | before | after |
|---|---|---|
| `test_browser_tab_ref.py` | 508.93 s | **28.76 s** |
| `browser-bridge` target | ~950 s | **177.79 s** (891 passed) |
| slowest single test | 234.22 s | **0.51 s** |

Margin against the 300 s budget: **1.28× → 588×**.

## Both new assertions have a killing mutant

Neither is an invariant guard — each was watched to fail, for **its own** reason:

- Drop `agent` from `REF_FREE_TEXT_SUBCOMMANDS` → `test_the_free_text_list_covers_agent_too` fails on its own `STUB_GOAL` assertion (line 564).
- Restore the old ordering → `test_a_refusal_never_pays_for_the_opencode_warm` fails on its own config-dir assertion (line 614).

🔴 **The new guard's observable is the config dir, not the opencode call log.** My first draft used `_oc_calls()` and would have been **vacuous**: the fake opencode deliberately does not touch `FAKE_OC_LOG` for `debug agent`, so it passes on the old code too. Reachability measured on both arms — `origin/main` creates the dir, the fixed tree does not. The rig's own `oc_config_dir` is pre-warmed, hence the deliberately cold override in the test.

## Deliberately not fixed

`browser agent --dry-run <goal>` still runs a full model session. `BROWSER_AGENT_DRY_RUN` is **tool-level by design** — only `browser_tool_impl.mjs:1111` consumes it, and `README.md:748` / `reference/agent.md:329` document it as suppressing mutating ops while the model still reasons. Short-circuiting it would change shipped semantics three docs describe. Filed as **#1284** with a mechanical closing condition instead.

## Verification

```bash
nix develop . --command bash -c 'python3 -m pytest scripts/browser-bridge/tests -q -n 4 --dist loadfile'
# 891 passed in 177.79s
```

Scope is contained: nothing outside `scripts/browser-bridge/` references `browser-agent`.

https://claude.ai/code/session_01GS7vJg4fSS3rbounyRmepL

### DIFF
--- a/scripts/browser-bridge/browser-agent
+++ b/scripts/browser-bridge/browser-agent
@@ -79,6 +79,61 @@ die() { echo "browser-agent: $*" >&2; exit 2; }
 # binary can never leave an orphaned tab. Both constraints point the same way.)
 command -v "$OPENCODE_BIN" >/dev/null 2>&1 || die "opencode not found on PATH ('$OPENCODE_BIN') — install it or set BROWSER_AGENT_OPENCODE"
 
+# --- parse args (MUST stay above the config-dir bootstrap and the WARM block) - #
+# 🔴 THIS BLOCK USED TO LIVE ~300 LINES BELOW, after the warm/bootstrap block, and
+# that ordering cost 90 s on EVERY invocation that was going to be refused anyway.
+# The warm step is bounded by OC_WARM_TIMEOUT (90 s) and needs a writable config
+# dir plus network to npm-install @opencode-ai/plugin; under a fresh HOME it can
+# do neither, so it burns the FULL timeout and only THEN did `a goal is required`
+# fire. MEASURED 2026-09-04 with a fresh HOME: `browser-agent --dry-run` (no goal)
+# took 91 s to emit a refusal that needs no opencode at all. Argument validation is
+# free; do it before anything expensive. This is the SAME fix, for the same reason,
+# that moved the `command -v opencode` preflight above the bootstrap (see the
+# comment on that check) — cheap refusals must not pay for setup they never use.
+#
+# Nothing between here and the warm block reads a parsed variable (the only
+# `$STEPS` in it are prose and an ESCAPED `x\$STEPS` inside a warning string), and
+# nothing in the warm block reads a positional — every `$1` there is function-local.
+GOAL=""; GOAL_SEEN=0; INSTANCE=""; ALLOW=""; DENY=""; STEPS=12; TIMEOUT=120; DRYRUN=0
+_csv_to_ws() { printf '%s' "$1" | tr ',' ' '; }
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --instance) shift; [ $# -ge 1 ] || die "usage: --instance <key>"; INSTANCE="$1" ;;
+    --instance=*) INSTANCE="${1#--instance=}" ;;
+    --allow-domains) shift; [ $# -ge 1 ] || die "usage: --allow-domains a,b"; ALLOW="$(_csv_to_ws "$1")" ;;
+    --allow-domains=*) ALLOW="$(_csv_to_ws "${1#--allow-domains=}")" ;;
+    --deny-domains) shift; [ $# -ge 1 ] || die "usage: --deny-domains a,b"; DENY="$(_csv_to_ws "$1")" ;;
+    --deny-domains=*) DENY="$(_csv_to_ws "${1#--deny-domains=}")" ;;
+    --steps) shift; [ $# -ge 1 ] || die "usage: --steps N"; STEPS="$1" ;;
+    --steps=*) STEPS="${1#--steps=}" ;;
+    --timeout) shift; [ $# -ge 1 ] || die "usage: --timeout S"; TIMEOUT="$1" ;;
+    --timeout=*) TIMEOUT="${1#--timeout=}" ;;
+    --dry-run) DRYRUN=1 ;;
+    -h|--help) grep -E '^#( |$)' "$0" | sed -E 's/^# ?//'; exit 0 ;;
+    # `--` ends the flags: everything after it is the goal, even if it starts with
+    # `-`. It used to take only the FIRST remaining arg and silently DROP the rest
+    # (`browser agent -- g1 g2` ran with goal "g1"), the same drop-after-`--` shape
+    # fixed in the `browser` CLI's nine loops. GOAL_SEEN (not [ -z "$GOAL" ]) so an
+    # empty first goal cannot be silently overwritten by a second.
+    --) shift
+        while [ $# -gt 0 ]; do
+          [ "$GOAL_SEEN" -eq 0 ] || die "only one goal is accepted (got extra: $1)"
+          GOAL="$1"; GOAL_SEEN=1; shift
+        done
+        break ;;
+    -*) die "unknown flag: $1" ;;
+    *) [ "$GOAL_SEEN" -eq 0 ] || die "only one goal is accepted (got extra: $1)"; GOAL="$1"; GOAL_SEEN=1 ;;
+  esac
+  shift
+done
+
+[ -n "$GOAL" ] || die "a goal is required: browser agent \"<goal>\""
+case "$STEPS" in (*[!0-9]*|"") die "--steps must be a positive integer (got: $STEPS)";; esac
+[ "$STEPS" -ge 1 ] 2>/dev/null || die "--steps must be >= 1 (got: $STEPS)"
+case "$TIMEOUT" in (*[!0-9]*|"") die "--timeout must be a positive integer of seconds (got: $TIMEOUT)";; esac
+[ "$TIMEOUT" -ge 1 ] 2>/dev/null || die "--timeout must be >= 1 (got: $TIMEOUT)"
+
+
 # --- isolated opencode CONFIG dir (token containment) ------------------------ #
 # devrc now deploys a ~43 KB global ~/.config/opencode/AGENTS.md (PRINCIPLES +
 # RULES + addendum). MEASURED on 1.18.4: that file IS injected into a run whose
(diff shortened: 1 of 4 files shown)


ZacxDev:
## Round 1 audit — two claims in the PR body above are FALSE

Correcting these publicly rather than silently editing the body, since anyone who already read it read the wrong version. Both are now fixed in the code that carries them (`321d73c4`), not only here.

### ❌ "Scope is contained: nothing outside `scripts/browser-bridge/` references `browser-agent`."

Wrong. `scripts/tests/test_opencode_config.py` holds **24 references** and ~12 assertions over this file's comment-stripped source — **including an ordering assertion** (`preflight < bootstrap_call`) guarding the `rm -rf` defect this reorder moves code around. `flake.nix` and `nix/home.nix` reference the wrapper too.

My check was itself defective: a `find … | xargs command grep -l` for the hyphenated literal returned nothing, while a direct grep on that same file returns 24. A spelling check, not a structural one. And `scripts/browser-bridge/tests` does **not** collect that file, so the PR's stated Verification never ran it.

Now run:
- `scripts/tests -k browser_agent` → **11 passed**
- `scripts/tests` (full) → **12,033 passed**

The parse block's comment now names that gate and the command to re-run it after any reorder.

### ❌ "same `rc=2`, byte-identical error text"

True only for inputs that already reached the goal check. An input that used to die **earlier** — in the code this block now precedes — is refused by this block instead. Measured on both trees:

| input (no goal) | base | PR |
|---|---|---|
| `env -u HOME … --dry-run` | `HOME: unbound variable`, **rc 1** | `a goal is required`, **rc 2** |
| `BROWSER_AGENT_WARM_TIMEOUT=abc` | its own validation error | `a goal is required` |
| stray `AGENTS.md` in the config dir | `unexpected instruction file at …` | `a goal is required` |

`OC_CONFIG_DIR` derives from `$HOME` under `set -u`, so the first is an **exit-code** change — reachable from exactly the contexts the preflight comment names (systemd user units, cron, `env -i`). Also, where the warm fails, the base emitted a `could not warm …` WARNING before every refusal and this PR does not, so full stderr is not byte-identical either.

All three are improvements. The claim was still wrong, and the comment I wrote reasoned only about forward dependence — never about which `die` wins.

### Also fixed this round

- **The stub was strictly more permissive than the real wrapper** — `--*` swallowed unknown flags, `-h`/`-x` became goals, a second positional silently last-won. That masks the exact regression the pair exists to catch. All three now refuse, mirroring `browser-agent`'s own loop.
- Three docstrings claimed wider coverage than their code: the positive control no longer backs the two absence assertions (they read the *stub's* stderr now); the free-text docstring said the wrapper refuses the mutant by name, when measured the **CLI** refuses first and the stub is never invoked; and the new guard pins **dir-creation, not cost** — a wall-clock budget was considered and rejected, because this suite's own CI flake *was* a wall-clock timeout under load.

Mutants re-verified **after** the stub change, since a fix round can break its own guard: free-text mutant fails at `test_browser_tab_ref.py:592` on its own `STUB_GOAL` assertion; ordering mutant fails on its own config-dir assertion.

### Not re-derived — flagged, not covered

The base wall-clock figures (91 s, 508.93 s, ~950 s, and the 155.98/234.22 durations). The auditor could not reproduce them: the box was at load 46–68 from other sessions for the whole window. They remain single measurements, and the *direction* (91,000 ms → 4–17 ms) was independently confirmed.

```audit-claims round=1 audited=967c5592..321d73c4
1. Corrected the false scope claim; ran the previously-missed gate scripts/tests (12033 passed) and scripts/tests -k browser_agent (11 passed); the parse-block comment now names that gate.
2. Corrected the false "same rc, byte-identical error" claim; the comment now states the three input classes whose diagnostic changed, one of them an exit-code change (rc 1 -> rc 2 under `env -u HOME`).
3. Tightened the test stub's grammar so it refuses unknown flags, -h/-x, and a second positional — it was more permissive than the real wrapper and could mask a forwarded-argv regression.
4. Corrected three docstrings that claimed wider coverage than their code (positive control's scope, the free-text mutant's actual refuser, the new guard pinning dir-creation not cost).
5. Re-verified both mutants AFTER the stub change; both still fail on their own assertions.
6. De-lined issue #1284's pointers (symbol names, not line numbers) — this PR staled them within the hour.
```

ZacxDev:
## Round 2 delta re-audit — and the ladder stops here

Range `967c5592..321d73c4`. No 🔴. The headline finding is against **round 1's own fix**.

### 🟡 The round-1 comment claimed a gate it does not have

Round 1 added a 🔴-marked comment saying that moving the parse block above the `command -v opencode` preflight "would break" `scripts/tests/test_opencode_config.py` §7.

I built exactly that mutant and ran the command the comment prescribes: **11 passed**. `scripts/browser-bridge/tests` on the same mutant: **891 passed**. Nothing catches it.

§7 asserts `preflight < bootstrap_call` on the comment-stripped source. Moving this block reorders neither member of that pair, so both assertions stay true. It pins the **preflight**, not this block.

The comment now states it per direction — below the warm → caught by the behavioural guard; above the preflight → **caught by nothing, measured**. A false coverage claim is worse than none, because it stops the next reader looking. That this appeared *in the round whose purpose was removing false coverage claims* is the point worth remembering.

### 🟡 The stub still diverged — in both directions

Round 1 said the stub now "MIRRORS the real grammar's REFUSALS". Measured, it did not:

| argv | real wrapper | stub (round 1) |
|---|---|---|
| `g --steps` (value-taking flag, final position) | rc 2 `usage: --steps N` | **rc 0** — accepted |
| `--instance=x g` (and the other `=` forms) | accepted | **rc 2 `unknown flag`** |
| `-h` / `--help` | prints help, **rc 0** | **rc 2 `unknown flag`** |

So round 1's "`-h`/`-x` … all now refuse, mirroring browser-agent's own loop" was false for `-h` — the wrapper does not refuse it. Too-permissive masks the regression the pair exists to catch; too-strict invents a CLI regression that isn't there.

### The claim is now *checked*, not just reworded

New drift-guard `test_the_stub_agent_agrees_with_the_real_wrapper_on_what_it_refuses`: a 21-case matrix driving the **real** wrapper (stub opencode + stub browser binary, so an accepted argv dies at the tool-set gate in ~50 ms instead of warming and running a model) alongside the stub, asserting both reach the same accept/refuse/help decision. Each case validates the matrix against the real wrapper **first**, so the expectations are measured rather than assumed.

Mutation-verified in both directions: drop the `[ $# -ge 1 ]` guard → **4 killed**; drop the equals-forms and `-h` → **6 killed**.

The stub is now written by one function (`_write_agent_stub`) — with two copies the guard would certify a stub nothing uses.

### 🟢 The input-class list read as exhaustive and was not

Added the missed `BROWSER_AGENT_LOCK_BUDGET` class and the `XDG_CACHE_HOME` conditional. And one class does not fit the "refused by this block instead" framing at all: `env -u HOME … -h` goes **rc 1 → rc 0** — a failure flipped to a *success*, not a sideways refusal.

Issue #1284: round 1 de-lined only the Pointers section; `browser-agent:722` survived elsewhere. Zero line-number pointers now.

### Verification

- `scripts/browser-bridge/tests` → **912 passed** (was 891; +21 matrix cases)
- `scripts/tests -k browser_agent` → **11 passed**
- `scripts/tests` (full, round 1) → **12,033 passed**

### 🔴 Why the ladder stops — this is NOT a clean round

Round 2 found real defects and they were fixed. The stop is the **attribution gate**, not convergence:

```
round 2 · payload lines changed THIS round: 0 executable (27/6 in browser-agent, ALL comments) · since round 1: 0 executable
```

Proof rather than assertion — comment-stripped `browser-agent` at both ends of the round: **360 lines each, byte-identical** (`cmp` clean, both files non-empty as the positive control). Round 1 was likewise 0 executable payload.

Two consecutive zero-payload rounds means the ladder has left the PR and is auditing scaffolding it wrote itself. Recording it explicitly so a reader can tell this from a converged ladder — the two look identical in the findings list otherwise.

**Left open deliberately, not absent:** the base wall-clock figures (91 s, 508.93 s, ~950 s, and the 155.98/234.22 durations) were never independently reproduced. Both auditors declined — the box sat at load 26–68 from other sessions throughout. The *direction* is confirmed (91,000 ms → 4–17 ms, re-measured); the absolute baselines remain single measurements by one observer.

```audit-claims round=2 audited=321d73c4..bb62c335
1. Corrected the false gate claim in the parse-block comment; it now states per-direction what pins the block (behavioural guard below the warm; NOTHING above the preflight, measured).
2. Closed the stub's remaining divergences in both directions: added the [ $# -ge 1 ] guard for value-taking flags; accepted the --flag=value equals forms and -h/--help.
3. Added test_the_stub_agent_agrees_with_the_real_wrapper_on_what_it_refuses — a 21-case matrix driving the REAL wrapper and the stub and asserting the same decision; mutation-verified both directions (4 killed / 6 killed).
4. Extracted the stub to a single _write_agent_stub() so the drift-guard checks the same stub the forwarding tests use.
5. Made the input-class enumeration non-exhaustive-by-admission; added the LOCK_BUDGET class, the XDG_CACHE_HOME conditional, and the -h rc 1 -> rc 0 success flip.
6. Removed the last line-number pointer from issue #1284.
7. LADDER STOPPED on the attribution gate (two consecutive rounds of zero executable payload), not on a clean round — round 2 found real defects.
```

PR #1286, opened 2026-09-04
fix(dl-router): reach the Discord original by rewriting the proxy URL, not by reading a link that is never there

ZacxDev:
`preferOriginalUrl` has been **inert in production since it shipped**. It reads `info.linkUrl`, which Chrome populates only from an **ancestor** `<a>` — and a Discord image attachment has none.

## Measured, against the live client

3 image attachments, 2 channels, 2 message shapes (single image + mosaic):

| | measured |
|---|---|
| `` host | `media.discordapp.net` — 3/3 |
| ancestor `<a>` around the image | **0 of 3** (`div[role=button].clickableWrapper`) |
| `cdn.discordapp.com` anchor in the message | 3/3, a **sibling at depth 9** |
| same pathname (unambiguous pairing) | true |
| `ex`/`hm`/`is` values across the two hosts | **byte-identical** |

So every real right-click lands on `if (!linkUrl) return srcUrl` and downloads the resized webp thumbnail — the defect #1110 set out to fix, still live for images. Every existing unit test supplied a `linkUrl`, so both suites stayed green over a feature that could not fire: the defect lived in the **seam**, not in either component.

## Why a rewrite, not a DOM read

A prior session eliminated host-rewriting because "the two hosts carry different query params". They do not — the proxy URL carries `ex`/`hm`/`is` **plus** the resize params. Probed from the page context of a live Discord tab, with both controls:

| | status |
|---|---|
| positive control — the message's own cdn anchor | 206 |
| **under test** — proxy url, host swapped, resize params dropped | **206** |
| negative control — same url, signature removed | network error |

That also closes the gap the handoff flagged as never probed: whether a cdn URL without a valid signature actually fails. It does.

Chosen over a content script, which would need a `discord.com` host permission and couple to Discord's hashed class names (they rotate every deploy). `originalFromPreview` is pure and rewrites only what `discordChannelId` recognises as an attachment, so an avatar or emoji on the same proxy host comes back untouched — pinned by a test that reaches that guard by a path no earlier check rejects.

## Behaviour changes, both deliberate

Two existing assertions move toward the clicked image's own original: the no-link case, and the mismatched-path case (the live mosaic shape, where the only anchor on offer belongs to a sibling image).

## Test matrix

The production path is pinned at the service-worker seam, asserting **which URL** was downloaded (not merely that something happened):

- **red** at `aba48864` — downloads `media.discordapp.net/...?format=webp&width=550`
- **green** at HEAD — downloads `cdn.discordapp.com/...?ex=1&is=2&hm=3`

## Gate, both tiers, base `a7dac5bd`

- `nix build .#checks.x86_64-linux.nodetests` — **rc 0**
- `nix build .#checks.x86_64-linux.pytests` — **rc 0** (built one at a time)
- dev-host `gate.sh` node tier — 1455 pass / 0 fail (dl-router 543, was 537)

One earlier pytests run went red on `test_six_writers_with_a_tiny_busy_timeout_still_land_every_row`. Attributed to load, not to this change: the whole 1020-test dl-router target took **391 s** in that run against **115 s** at base — load inflates every test, an assertion inflates one — the test is itself a deliberate 5 ms busy-timeout load reproduction, this diff contains no Python, and a genuine re-run was green.

The dev-host pytest tier also shows 3 failures; all 3 were confirmed to fail identically at base.

`manifest` 0.3.2 -> 0.3.3, because 0.3.2 was never actually loaded by the browser — so the outstanding restart check becomes "`brave://extensions` reads **0.3.3**".

https://claude.ai/code/session_01BePLrAzQrg983BX2FSdEjR

### DIFF
--- a/scripts/dl-router/SKILL.md
+++ b/scripts/dl-router/SKILL.md
@@ -205,24 +205,38 @@ media = [
 ]
 ```
 
-An image's own `src` is often a **downscaled copy from a resizing proxy**.
-🔴 **Do NOT expect the original on a wrapping `<a>` — on the chat site this was
-built for there is no wrapping anchor at all.** MEASURED 2026-09-03 on the live
-client, 3 image attachments across 2 channels and 2 message shapes: the number
-of ancestor `<a>` elements between the image and its message was **0 of 3**. The
-origin anchor exists in the same message but as a **sibling**, and a browser
-fills the context menu's `linkUrl` only from an ANCESTOR link — so a rule that
-waits for one waits forever.
+An image's own `src` is often a **downscaled copy from a resizing proxy**, which
+is why the anchor accessor goes FIRST in the list above: it is how this rule
+reaches the original. 🔴 **That accessor is a DESCENDANT query and it does not
+need a wrapping `<a>`.** `safeQuery` resolves each `element` with
+`container.querySelectorAll` (`player_buttons.js`), so an anchor that is merely
+somewhere inside `container` — a SIBLING of the image, which is the live shape —
+resolves fine. Do not delete it on the strength of the context-menu note below:
+🔴 **the player-button path does NOT rewrite anything.** `playerDownload` hands
+its `mediaUrl` straight to the download API and never calls
+`originalFromPreview()`, so for THIS path the anchor accessor is the only route
+to the original.
 
-🔴 **Rewriting the proxy URL's host IS the supported route, and the rule here
-used to say the opposite.** It claimed the two hosts carry different signature
-parameters; they do not. Measured for one asset: the proxy URL carries the
-signature parameters **plus** the resize knobs, with the signature **values
-byte-identical** to the origin anchor's, so the rewrite keeps a valid signature.
-Probed with both controls — the message's own origin anchor **206**, the
-rewritten URL **206**, the same URL with the signature stripped **fails**.
-`originalFromPreview()` in `route_core.js` is the one implementation; do not
-open-code a second.
+### The "Save to library…" CONTEXT MENU is a different path with a different rule
+
+🔴 **Scoped to `info.linkUrl`, and it does NOT apply to the `media` list above.**
+A browser fills a context menu's `linkUrl` only from an ANCESTOR link, and on
+the chat site this was built for an image attachment has none. MEASURED
+2026-09-03 on the live client, 3 image attachments across 2 channels and 2
+message shapes: ancestor `<a>` elements between the image and its message,
+**0 of 3**. The origin anchor is a sibling — reachable by a descendant query,
+invisible to `linkUrl`. So the menu path cannot wait for a link and must
+rewrite instead.
+
+🔴 **For the menu path, rewriting the proxy URL's host IS the supported route,
+and this file used to say the opposite.** It claimed the two hosts carry
+different signature parameters; they do not. Measured for one asset: the proxy
+URL carries the signature parameters **plus** the resize knobs, with the
+signature **values byte-identical** to the origin anchor's, so the rewrite keeps
+a valid signature. Probed with both controls — the message's own origin anchor
+**206**, the rewritten URL **206**, the same URL with the signature stripped
+**fails**. `originalFromPreview()` in `route_core.js` is the one implementation
+**of the rewrite**; do not open-code a second.
 
 ⚠ **A video's `src` is USUALLY the origin already, but that is not guaranteed
 and was measured at zero points** — every video in the live route log was
--- a/scripts/dl-router/SKILL.md
+++ b/scripts/dl-router/SKILL.md
@@ -205,11 +205,29 @@ media = [
 ]
 ```
 
-An image's own `src` is often a **downscaled copy from a resizing proxy**, with
-the original on the wrapping `<a>`; a video's `src` already *is* the original.
-🔴 **Never reach the original by rewriting a proxy URL's host** — the two hosts
-carry different signature parameters, so the rewrite drops a signature belonging
-to the other host. The anchor's href is already signed for where it points.
+An image's own `src` is often a **downscaled copy from a resizing proxy**.
+🔴 **Do NOT expect the original on a wrapping `<a>` — on the chat site this was
+built for there is no wrapping anchor at all.** MEASURED 2026-09-03 on the live
+client, 3 image attachments across 2 channels and 2 message shapes: the number
+of ancestor `<a>` elements between the image and its message was **0 of 3**. The
+origin anchor exists in the same message but as a **sibling**, and a browser
+fills the context menu's `linkUrl` only from an ANCESTOR link — so a rule that
+waits for one waits forever.
+
+🔴 **Rewriting the proxy URL's host IS the supported route, and the rule here
+used to say the opposite.** It claimed the two hosts carry different signature
+parameters; they do not. Measured for one asset: the proxy URL carries the
+signature parameters **plus** the resize knobs, with the signature **values
+byte-identical** to the origin anchor's, so the rewrite keeps a valid signature.
+Probed with both controls — the message's own origin anchor **206**, the
+rewritten URL **206**, the same URL with the signature stripped **fails**.
+`originalFromPreview()` in `route_core.js` is the one implementation; do not
+open-code a second.
+
+⚠ **A video's `src` is USUALLY the origin already, but that is not guaranteed
+and was measured at zero points** — every video in the live route log was
+already on the origin host, so the rewrite simply does not fire for them. It
+*will* fire on a proxy-host video src, and that shape is unverified.
 
 **Important details**
 - The media URL is **signed and rotates** — `player_buttons.js` reads it **at
--- a/scripts/dl-router/SKILL.md
+++ b/scripts/dl-router/SKILL.md
@@ -209,9 +209,13 @@ An image's own `src` is often a **downscaled copy from a resizing proxy**, which
 is why the anchor accessor goes FIRST in the list above: it is how this rule
 reaches the original. 🔴 **That accessor is a DESCENDANT query and it does not
 need a wrapping `<a>`.** `safeQuery` resolves each `element` with
-`container.querySelectorAll` (`player_buttons.js`), so an anchor that is merely
-somewhere inside `container` — a SIBLING of the image, which is the live shape —
-resolves fine. Do not delete it on the strength of the context-menu note below:
+`container.querySelectorAll` (`player_buttons.js`), so the anchor only has to be
+a descendant of `container` — it does not have to be an ancestor of the image.
+⚠ **Confirm your `container` actually encloses it.** What was measured is the
+NEGATIVE half only (0 ancestor `<a>` of 3); the origin anchor was nine levels
+away from the image, and whether it falls inside any particular `container`
+selector has never been measured. Do not delete this accessor on the strength
+of the context-menu note below:
 🔴 **the player-button path does NOT rewrite anything.** `playerDownload` hands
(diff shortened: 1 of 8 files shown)


ZacxDev:
Round 1 adversarial audit complete; fixes landed in `261a4cf5`. Both sandbox tiers green on the rebased tree at base `2882d2c7`: `nodetests` `RESULT: PASS` (dl-router 550/550, total 1462/0 fail), `pytests` `RESULT: PASS` (dl-router 1020/1020, total 21073 passed / 0 failed).

```audit-claims round=1 audited=3bcb503b..261a4cf5
1. SKILL.md's live 🔴 "never reach the original by rewriting a proxy URL's host" rule rewritten; its three constituent claims (original on a wrapping <a>, hosts carry different signature params, a video src is always the origin) replaced with the measured position.
2. correlateCapture tier 1 now also matches a capture's `downloadUrl`, and onMenuClicked stamps the outgoing url onto the capture it came from — claimed to restore tier-1 binding for rewritten Discord downloads, which had fallen to tier 3.
3. stampDownloadUrl matches on the clicked element's `srcUrl` rather than on recency — claimed to bind a download to its own capture, not to the newest one, in a multi-image message.
4. The `""` resize-param comment rewritten: the stray-`&` mechanism it asserted is false (the urlencoded parser skips empty sequences); the entry actually covers `?=value`, which is unobserved and defensive.
5. The `KNOWN UNHANDLED VARIANT` docstring rewritten: that variant is now handled, and the widening it invited would replace a correct answer with a downscaled proxy copy.
6. The `// (A <video> is unaffected)` comment replaced: proxy-host video srcs ARE rewritten, video was measured at zero points, and there is no fallback if the rewrite is wrong for it.
7. Five new tests claimed to kill mutants the round-1 audit reported SURVIVED or newly introduced: M1 (`""` entry), M10 (wrong-link-host exit), M11 (non-attachment-pair exit), M13 (tier 1 drops downloadUrl), M14 (stamp takes newest capture).
8. Emoji removed from the two dl-router extension sources to satisfy source_hygiene.test.mjs's ASCII rule, which the round-1 fixes had broken.
```

Not addressed, deliberately, with reasons: audit M7 (preview-host guard) is near-inert — the attachment guard already rejects everything it would catch. Audit M12 is genuinely unreachable dead code (both `hostOf` calls already parsed both URLs, so the `new URL` cannot throw) and is pre-existing, not introduced here. Neither is a coverage gap.

ZacxDev:
Round 2 delta audit complete; fixes landed in `54d2a654`. Both sandbox tiers green: `nodetests` `RESULT: PASS` (dl-router 551/551, total 1463/0 fail), `pytests` `RESULT: PASS` (21073 passed / 0 failed).

Round 2's own summary of the code half: "the fix does what it claims, the mutation results are real, and I found no regression it introduced in executable code." Both its findings were prose.

```audit-claims round=2 audited=261a4cf5..54d2a654
1. F1 — the round-1 SKILL.md rule was rewritten AGAIN, because it had landed inside `## Player buttons / embedded video downloads`, a section whose TOML `media` list is resolved by container.querySelectorAll (a DESCENDANT query) and whose path never calls originalFromPreview. The context-menu path now has its own subsection; the `info.linkUrl` ancestor fact is scoped to it; the player path states plainly that it does not rewrite and that its anchor accessor is the only route to the original.
2. F2 — the refuted "do NOT rewrite a proxy URL's host" claim was removed from `config.example.toml`, the second of its sites, which round 1 missed.
3. A THIRD site the round-2 grep did not name was found and fixed: `service_worker.test.mjs`'s existing menu-test comment still asserted the original sits "on the wrapping <a href>", which is why that test's supplied linkUrl looked sufficient. It now states why both menu tests exist.
4. F3 — the `downloadUrl` comment no longer claims onMenuClicked stamps "the url it hands to chrome.downloads.download"; it names the streaming branch, where the url goes to the sidecar's /fetch instead.
5. F4 — the preview-host guard (audit M7) is now pinned by a test rather than declined. The decline reason was wrong in mechanism: DISCORD_CDN_HOSTS holds BOTH hosts, so an origin-host attachment would have had its resize knobs stripped without the guard. The mutant now dies on that test's own assertion, fail 1.
```

Not changed, and why: round 2's two 🟢 non-findings (the `preferOriginalUrl:407` bare-`srcUrl` return, semantically identical since `originalFromPreview` is the identity for a non-preview host; and `stampDownloadUrl` applying no tab filter, where an identical `srcUrl` already implies the same message) were left alone — both change nothing a reader does. Audit M12 remains unreachable pre-existing dead code.

ZacxDev:
🔴 **Retraction of a claim in the `audit-claims round=2` block above.** Item 4 of that block said the `downloadUrl` comment "no longer claims onMenuClicked stamps the url it hands to `chrome.downloads.download`; it names the streaming branch." **That was false against the tree.** The edit was made and then destroyed before the commit: the M7 mutation battery restored `route_core.js` with `git checkout --` while the fix was still uncommitted, and the reverted file was staged and committed. Round 3 caught it by reading the tree instead of the claim. It is landed now in `852585a0` and verified present in the commit itself, not only in the working tree.

Round 3 fixes landed in `852585a0`. Both sandbox tiers green: `nodetests` `RESULT: PASS` (dl-router 551/551, total 1463/0 fail), `pytests` `RESULT: PASS` (21073 passed / 0 failed).

```audit-claims round=3 audited=54d2a654..852585a0
1. F3 (round 2's lost edit) is now actually in the tree and in the commit: the `downloadUrl` comment states that the stamp runs BEFORE the streaming branch, so the url reaches chrome.downloads.download on the ordinary path and goes to the sidecar's /fetch for a manifest, where the stamp is inert rather than wrong.
2. 🟡-1 — the `###` context-menu subsection introduced in round 2 was left open, so 45 lines of player-button guidance (Important details, the "Buttons don't appear" list, the DEPLOY ORDER note) sat under a context-menu heading. Closed with an explicit `### Player button details` heading that states why it exists.
3. 🟢-3 — the "sibling, inside the container" claim was dropped at BOTH prose sites. The measurement (0 ancestor <a> of 3) supports only the negative half; whether the anchor falls inside any given `container` was never measured, and this repo's own note records it nine levels from the image. Both sites now state the descendant rule and tell the reader to confirm their container encloses it.
4. The "this exercise the swap" typo in service_worker.test.mjs was corrected.
```

Round 3's own verdict on the code: no executable code in its range, both tiers green, the M7 mutant reproduced dying `fail 1` on its own assertion, and its fourth-site sweep found **no** further copy of the refuted claim — reported with a positive control (the same patterns matched 3 lines at `261a4cf5`, proving the patterns can match).

Not changed: round 3's two stated non-findings (`SKILL.md:215` being marginally wider than the code, scoped correctly by the next sentence; `preferOriginalUrl:407`'s bare `srcUrl` return, semantically identical).

ZacxDev:
Round 4 returned **safe to merge, no 🔴, no 🟡** — but four 🟢 findings, each a false sentence naming an action a reader would take. Fixed rather than shipped, in `6de885d7`. Both sandbox tiers green: `nodetests` `RESULT: PASS` (dl-router 551/551, total 1463/0), `pytests` `RESULT: PASS` (21073 passed / 0 failed).

```audit-claims round=4 audited=852585a0..6de885d7
1. F1 — "the stamp is inert there rather than wrong" was FALSE and the opposite error to the one it replaced. Verified against the code: startFetch runs its own correlateCapture on the same url and tier 1 reads c.downloadUrl; a stamp implies the rewrite fired, which implies a cdn attachment target, which implies directFile, so `streaming` reduces to `manifest` and that fetch correlates at tier 1 against the capture just stamped. The comment now says the stamp is DECISIVE there, and names what believing "inert" would license.
2. F2 — config.example.toml's bare "on a sibling <a> INSIDE the container" claim removed; it sat four lines above the round-3 hedge that contradicted it. SKILL.md had already been swept correctly.
3. F3 — the round-3 heading note said the context-menu path "has no rules". It has no PLAYER rules but IS governed by [site_rules."<host>".context], read by content_capture.js on the contextmenu event. Corrected, and it now points at that rule as the fix when menu downloads land in the catch-all.
4. F4 — the round-3 typo repair changed the subject's number instead of the verb's, making "these exercise the swap" attribute swap coverage to both menu tests; the second supplies no linkUrl and cannot reach the swap. Rewritten to say which test is the swap branch's only coverage and why deleting it would remove that coverage.
5. Emoji removed from route_core.js again — dl-router extension sources are ASCII-only and the suite enforces it. Second time in this ladder.
```

Round 4's own position on the ladder, recorded because it matters: it explicitly declined to stop on the "prose could be reworded forever" basis, on the grounds that all four findings misdirect a specific concrete action rather than being stylistic. It also verified claim 1 in the tree AND the commit separately — the distinction that caught round 2's lost edit — and confirmed the round-3 heading fix held, with no fifth instance of correct-content-wrong-section.

Gap it flagged rather than covered: dev-host `gate.sh --tier both` was not run this round (no executable line and no Python changed; both sandbox tiers green).

ZacxDev:
Round 5 fixes landed in `22474fa8`. Both sandbox tiers green: `nodetests` `RESULT: PASS` (dl-router 551/551, total 1463/0 fail), `pytests` `RESULT: PASS` (21073 passed / 0 failed).

```audit-claims round=5 audited=6de885d7..22474fa8
1. The round-4 comment claimed THIS test is "the only one that reaches the swap" and that deleting it would remove "the swap branch's only coverage". FALSE: identity.test.mjs's "a proxy thumbnail is swapped for the original behind it" also reaches the swap exit and kills the swap mutant alone — verified present on origin/main (from #1110), so it predates this PR. Now scoped to this file, and it names the unit-level test explicitly.
2. The F3 pointer ("if menu downloads land in the catch-all, add a context rule") sat UNDER the heading "everything below is about the BUTTON, not the menu", so the operator it names was told by the heading to skip it. Moved to the end of the context-menu subsection.
3. "decides where a menu download files" softened: a context rule contributes subject/tags as the first and most specific signal, but the sidecar's /match weighs them against title, Open Graph, link text and url-derived signals. Strongest lever, not the decision.
```

## 🔴 THE LADDER STOPS HERE, and this is the ESCAPE HATCH, not convergence

Round 5 was **not** a clean round — it returned one 🟡 and one 🟢, both fixed above. Under the findings-keyed rule that would mean a round 6. I am stopping instead, and recording why so the next reader knows this is open rather than absent:

- **No 🔴 in rounds 3, 4 or 5.** Round 4's verdict was "safe to merge"; round 5's was "safe to merge".
- **Blast radius is bounded by "the document contains a false sentence."** No finding since round 2 has touched an executable line. Round 5 confirmed the range contains no assertion, guard or code-path change.
- **The recurring shape was swept at every site, not only where reported.** Round 5 checked the correct-content-wrong-section shape at all four edited sites and cross-checked both prose sites (`SKILL.md`, `config.example.toml`) against each other after four rounds of edits, finding no self-contradiction.
- **Round 6 would be a prose round auditing a prose fix to a prose defect** — the non-terminating loop. Rounds 2, 3, 4 and 5 each found a defect introduced by the previous round's *fix*, every one of them a sentence rather than code.

**What convergence DID look like, separately:** round 5 verified F1 — the item most likely to be a third wrong correction of the same comment — link by link against `service_worker.js`, and proved unreachable the specific exception it was asked to hunt (a stamp coexisting with `streaming` true via `mediaType === "video"`, foreclosed by `stampDownloadUrl`'s `!srcUrl` guard). It also validated the ASCII hygiene guard with a negative control. The code half of this PR is settled.

**Left open, deliberately:** the two 🟢 non-findings round 5 declined to raise, and any further prose polish. If a future reader finds another false sentence in these files, it is unaudited, not cleared.

PR #1296, opened 2026-09-04
fix(browser-agent): don't warm opencode before argv validation; stub the two argv tests

ZacxDev:
Unblocks `tekton/devrc-pytests`, which has been red on **4 of the last 6 PRs** on a repo with `enforce_admins`.

## Root cause

`browser-agent` warms an isolated opencode config dir by npm-installing `@opencode-ai/plugin`. That block ran at **script load** — above argv parsing, and far above `[ -n "$GOAL" ] || die "a goal is required"`. So every invocation paid a network round trip, **including ones that die two steps later having never used the dir**.

`devrc-pytests` runs in a nix build sandbox with **no network**, where the warm stalls to `OC_WARM_TIMEOUT` (90s) + `OC_LOCK_BUDGET` (+30s).

Reproduced under `unshare -rn`: **287.68s against a 300s ceiling.** That thin margin is why it failed intermittently and never reproduced for anyone running the suite with network. CI reported `TimeoutExpired … returncode -9` plus the warm warning on exactly the two tests that invoke `agent` — matching the verdict’s `failed=2`.

## Two changes

**1. Move the warm below argv validation.** No behaviour change for a run that proceeds — every path reaching opencode still passes through it. It only stops runs that are about to `die` on their arguments from doing network I/O first. The `command -v "$OPENCODE_BIN"` preflight still runs before the bootstrap, preserving the ordering constraint documented there.

**2. Drive both agent tests against the stub `browser-agent`** already used by their neighbours. `--dry-run` does **not** avoid opencode — `BROWSER_AGENT_DRY_RUN` is passed as an env var into `setsid "$OPENCODE_BIN"`, so the agent genuinely runs and the flag only stops it *acting*. A unit test of argv routing was launching an LLM agent to find out whether a token survived.

This is a **correctness** improvement too. Both tests asserted the *absence* of error strings; both now assert **positively** that the goal reached `browser-agent`. An absence pair passes if the wording changes, if the agent dies earlier for an unrelated reason, or if the stub never runs. The real error strings stay covered by `test_..._is_reachable`, which still runs the real `browser-agent` and is now cheap because it dies at goal validation before the warm.

## Rejected first attempt (recorded so it is not retried)

Setting `BROWSER_AGENT_WARM_TIMEOUT=1` in the fixture. Measured — it made the tests **3.5× slower** with network (22.14s → 77.81s), because killing the warm forces the expensive global-config fallback that CI already pays anyway. It traded one stall for another.

## Verification

Both tests were **mutation-tested**, not merely observed green:

| mutation | result |
|---|---|
| drop `agent` from `REF_FREE_TEXT_SUBCOMMANDS` | first test **FAILS** |
| widen the tab guard onto the subcommand | second test **FAILS** |

Both restored clean afterwards.

| measurement | before | after |
|---|---|---|
| no-network control (same harness) | 287.68s | **1.05s** |
| `test_browser_tab_ref.py` | 74–80s | **27.60s**, 54 passed |
| full `scripts/browser-bridge/tests/` | — | **890 passed** |

Also verified directly: `browser-agent --dry-run` with no goal now exits 2 with `a goal is required`, emits **no** warm warning, and creates **no** config dir.

https://claude.ai/code/session_018aHWYAKJvN3ewGjJ2eKiN9

### DIFF
--- a/scripts/browser-bridge/browser-agent
+++ b/scripts/browser-bridge/browser-agent
@@ -348,22 +348,6 @@ OCJSON
   return 1
 }
 
-_oc_ver="$(oc_warm_version)"
-if oc_is_warm "$_oc_ver" || oc_bootstrap "$_oc_ver"; then
-  OC_ISOLATED=1
-else
-  echo "browser-agent: WARNING: could not warm the isolated opencode config dir '$OC_CONFIG_DIR' (opencode $_oc_ver, timeout ${OC_WARM_TIMEOUT}s). Falling back to the GLOBAL opencode config — this run works but costs ~7.7k extra input tokens PER REQUEST (x\$STEPS) for an AGENTS.md this agent cannot act on. Fix: make '$OC_CONFIG_DIR' writable and re-run ONCE with network available (the warm step needs to npm-install @opencode-ai/plugin into it)." >&2
-fi
-
-# Fail loud if an instruction file ever appears here — that would silently
-# reintroduce the whole cost this isolation exists to remove.
-for _stray in "$OC_CONFIG_DIR/AGENTS.md" "$OC_CONFIG_DIR/CLAUDE.md"; do
-  [ -e "$_stray" ] && die "unexpected instruction file at '$_stray' — the isolated opencode config dir must stay free of AGENTS.md/CLAUDE.md (that is the entire point; see the comment above). Remove it."
-done
-
-if [ "$OC_ISOLATED" = 1 ]; then
-  export OPENCODE_CONFIG_DIR="$OC_CONFIG_DIR"
-fi
 
 # Emit a compact schema result on stdout and exit (0 for ok/partial, 1 for
 # blocked). NEVER prints raw page content — only the structured schema.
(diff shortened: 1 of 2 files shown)


ZacxDev:
Closing as **superseded by #1285** (`2b1d3552`, "validate args before the warm block — 508s -> 29s, and the CI flake it caused"), which landed while this was being written.

Same diagnosis and the same two fixes — move the opencode warm below argv validation, and drive the agent tests against the stub instead of a real model session — plus `test_the_stub_agent_agrees_with_the_real_wrapper_on_what_it_refuses`, which guards the stub against drifting from the real wrapper. That guard is a genuine improvement on what this PR had, and #1285 also covers `test_browser_agent.py` and `run-tests.sh`, which this did not touch.

Nothing here is worth salvaging on top of it.

My process failure, recorded so it is not repeated: I claimed `devrc-pytests-red` but did **not** sweep `gh pr list --state open` before starting. The claim namespace cannot see work that is already an open PR — the sweep is the only thing that catches an unclaimed duplicate, and I skipped it.

One measurement from this branch may still be worth having, since it is a different instrument from #1285's: reproduced under `unshare -rn`, the two tests took **287.68s against a 300s ceiling** before the fix and **1.05s** after. That thin margin is why the failure was intermittent and never reproduced for anyone running the suite with network.

PR #1304, opened 2026-09-04
fix(seed): refuse to overwrite a pod entry whose bytes differ — the cutover inverted the authority

ZacxDev:
## The defect

`seed.sh` pushes local → pod and its extract **"adds and overwrites but never deletes"**. That was
safe while the local store was authoritative. Since the Cairn cutover the **pod** is, and the same
push is a *derivative overwriting the authority*: a shared entry whose pod copy has moved on — a
bullet appended via `cairn append`, an `OPEN:` rewritten `RESOLVED <sha>:` — is replaced by this
host's older copy, and the verdict still prints `seed: OK` because the **name** landed.

**MEASURED 2026-09-02/03 on the real store:** of 25 bullets present locally but not on the pod,
**five were the pod being newer** — including two `OPEN:` → `RESOLVED` closures carrying ~20 lines
of later corrections. A re-seed would have reverted all five and reported success.

## The guard

A **pre-flight before the tar**. For exactly the paths in `$staged_list`, ask the pod whether it
holds different bytes; refuse with **exit 8, pushing nothing**, and name them.
`--allow-overwrite` proceeds and still prints what it replaced.

- 🔴 **Pre-flight, not a louder verdict.** The existing containment check runs *after* the extract,
  so by the time it can speak the bytes are gone.
- 🔴 **Bytes, not mtime or "newness".** Two copies cannot be ordered — a local edit not yet pushed
  and a pod edit not yet pulled both read as different. Post-cutover the pod is the authority, so
  *any* difference means this push overwrites it with a derivative. No clock needed.
- 🔴 **No third `find`.** It reuses the list `_shippable_entries` already wrote, so this comparison
  and the verdict are one population rather than two walks that agree by luck — and
  `test_the_two_find_expressions_are_IDENTICAL` still pins exactly two listings. `test -f` on the
  pod yields the intersection for free: a staged path the pod lacks is a pure addition.

**Header corrected.** It opened `🔴 THE LOCAL STORE IS AUTHORITATIVE` — false since the cutover, and
the sentence outlived the fact for the whole life of it. What still holds (this script never
*writes* to the local store) is kept, and separated from what changed.

## Two things the tests caught in my own work

**`|| :` in the probe is load-bearing.** `xargs` exits 123 when any child exits non-zero, and a
staged path the pod does *not* have is the ordinary case. Without it the probe returned 123,
`set -euo pipefail` aborted, and **seeding a genuinely new entry died with exit 123 and an empty
stderr** — the guard against data loss breaking the script's primary function. All four tests
failed that way, including the two that must pass.

**Red-at-base was proving the wrong thing.** Asserting the exit code before the bytes made the base
run fail on `0 vs 8` and never reach the clobber assertion. Reordered, red at base is now
`THE POD'S BYTES WERE REPLACED`.

## Test matrix — base `46264622`

| test | at base | class |
|---|---|---|
| different bytes → refuses, **pod untouched** | **RED** (on the bytes) | regression |
| identical bytes → proceeds | green | invariant guard |
| absent from pod → proceeds, addition lands | green | invariant guard |
| `--allow-overwrite` names what it replaced | red — *only because the flag is new* | **not** regression coverage, labelled |

The two invariant guards are the ones that matter for over-firing: a guard that fired on mere
*presence*, or that broke pure additions, would be switched off within a day.

`test_subsystem_store_api.py`: **717 passed**.

## Not verified

Neither gate tier has run on this branch yet, and nothing has been exercised against the **live**
pod — the tests drive a fake `kubectl` whose `exec` rewrites `/data` to a temp dir.

https://claude.ai/code/session_01RRZVCFV6DDGs1pZwmCrHJo

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -118,7 +118,7 @@
       gateTools = [
         gatePyEnv pkgs.bash pkgs.ripgrep pkgs.git pkgs.util-linux pkgs.jq
         pkgs.gnugrep pkgs.curl pkgs.nodejs pkgs.nix pkgs.opencode pkgs.logrotate
-        pkgs.rsync pkgs.zsh pkgs.age
+        pkgs.rsync pkgs.zsh pkgs.age pkgs.dash
       ];
     in
     {
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -479,10 +479,24 @@ cd "$ROOT" || { echo "run-tests: cannot cd to ROOT=$ROOT" >&2; exit 3; }
 #           cannot occur in. Both hosts run zsh as the login shell, and
 #           flake.nix's `gateTools` carries it for the sandbox.
 #
+# dash:     scripts/tests/test_subsystem_store_api.py
+#           (TestTheSeedProbeAnswersSurviveTheRealPodShell, 5 tests). The SAME
+#           shape as zsh above, one layer out. `seed.sh`'s pod probe runs in the
+#           store-api pod, whose `/bin/sh` is dash (`Dockerfile`:
+#           `FROM python:3.12-slim` -> Debian); dash's `echo` INTERPRETS
+#           backslash escapes and bash's does not. Every test in that file drives
+#           a fake `kubectl` that runs the probe under THIS host's shell, so the
+#           whole suite is structurally blind to the difference — and the
+#           difference is a SILENT one (the two sides' join keys diverge and a
+#           pod entry that DIFFERS is overwritten as though it were new).
+#           Measured 2026-09-05: without dash on PATH all 5 skipped, which is
+#           the "green having measured the one shell the defect cannot occur in"
+#           failure the zsh note describes. They FAIL rather than skip now.
+#
 # 🔴 `python` is listed as well as `python3` because THIS SCRIPT invokes
 # `python -m pytest`, not `python3`. Asserting only `python3` checked a binary
 # the runner never calls.
-REQUIRED_TOOLS=(bash curl node rg git awk jq grep setsid python python3 nix-instantiate opencode logrotate rsync zsh)
+REQUIRED_TOOLS=(bash curl node rg git awk jq grep setsid python python3 nix-instantiate opencode logrotate rsync zsh dash)
 missing_tools=()
 for t in "${REQUIRED_TOOLS[@]}"; do
   command -v "$t" >/dev/null 2>&1 || missing_tools+=("$t")
--- a/scripts/subsystem-store-api/seed.sh
+++ b/scripts/subsystem-store-api/seed.sh
@@ -1,10 +1,22 @@
 #!/usr/bin/env bash
 # Seed the cluster store from the LOCAL one. Phase 1 (proposal §4).
 #
-# 🔴 THE LOCAL STORE IS AUTHORITATIVE, AND THIS SCRIPT NEVER WRITES TO IT.
-# `~/.claude/analyze-service-index/` is client-confidential, not re-derivable by
-# re-running recon, and phase 1's whole premise is "local stays authoritative and
-# untouched".
+# 🔴 THE LOCAL STORE IS *NOT* AUTHORITATIVE ANY MORE — THE POD IS. This header
+# said the opposite until 2026-09-04, and the sentence outlived the fact by the
+# whole life of the Cairn cutover, which made the hosted store canonical and
+# froze `~/.claude/analyze-service-index/` to a per-host mirror.
+#
+# What still holds, unchanged: THIS SCRIPT NEVER WRITES TO THE LOCAL STORE. It is
+# client-confidential and not re-derivable by re-running recon, so it is read
+# ONLY, for the reasons enumerated below.
+#
+# What CHANGED is the direction of the risk. The push can no longer be reasoned
+# about as "authoritative source refreshing a derivative": it is a DERIVATIVE
+# overwriting the AUTHORITY, and the extract adds and overwrites but never
+# deletes. That is why the pre-flight before the tar refuses any staged entry
+# whose bytes differ from the pod's — measured 2026-09-02/03, a re-seed would
+# have reverted five pod-newer bullets, two of them `OPEN:` -> `RESOLVED`
+# closures, and reported success.
 #
 # ⚠ This block used to open "THE LOCAL STORE IS THE ONLY COPY … has no
 # off-machine backup". Both halves are now false — daily age-encrypted bundles go
@@ -38,6 +50,7 @@ STORE=""
 STAGE=""
 PUSH=""
 DEST="/data"
+ALLOW_OVERWRITE=0
 
 while [[ $# -gt 0 ]]; do
   case "$1" in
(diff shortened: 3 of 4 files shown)


ZacxDev:
## Round 1 audit — 🔴 fixed, plus five 🟡/🟢. And the 🔴 was the guard's own clean path.

The audit drove the pre-flight end-to-end with a probe returning silence at rc 0: the push
**destroyed the pod's newer bytes and printed `seed: OK`** — the exact defect this PR exists to
prevent, with the guard installed. The probe emitted a line only for files the pod *has*, so
"the pod holds none of them" and "the probe never ran" were the same observation.

🔴 **Refusing on "0 present" was the wrong fix, and I tried it first** — a pod holding none of the
staged entries is the ordinary first-seed case, and that rule failed **18 legitimate tests**. The
answerable question is not "did it find anything" but **"did it SEE the whole list"**. Every path is
now answered — a hash, `ABSENT`, or `UNREADABLE` — and a short reply is the fault (exit 9). Counts
print on every path, which is this file's own silent-zero rule applied to the guard itself.

Also fixed:

- 🟡 the local side was bare `xargs -r sha256sum`, which word-splits and parses quotes:
  `sc/two words.md` → rc 123, `sc/it's.md` → rc 1, both with **no `seed:` line**. Base handled both
  at rc 0. Now `-I{}`, matching the remote.
- 🟡 **the refusal's remedy could not clear the refusal.** `cairn sync` refreshes
  `~/.cache/subsystem-store`, not the `--store` tree; `cairn put` writes the *pod*, moving it
  further from the mirror. With 52 of 157 shared entries differing today, the old text pointed at
  `--allow-overwrite` as the only exit — the permanently-red-gate shape. It now says a push is
  probably no longer the right verb and names the per-entry API.
- 🟡 the header read as "the push can no longer revert the pod". It covers **211 of 1,755 files
  (~12%)**; the tar carries whole scope directories including 15 `.git` repos, one measurably
  divergent. Stated rather than left to imply a guarantee.
- 🟢 an unreadable pod file was classified a pure addition → now answers `UNREADABLE`, which cannot
  equal a hex digest, so it always lands in the clobber set.
- 🟢 `seed: pushing …` printed before the guard could refuse → `preparing push`.
- 🟢 `--allow-overwrite` documented in Usage.

## Mutation matrix — re-run, not asserted

| mutant | before | now |
|---|---|---|
| `LC_ALL=C join` → `join` | **SURVIVED** all 40 seed tests | **killed** |
| drop the answered-count gate | n/a (new) | **killed** |

🔴 **My first locale test was vacuous and I caught it only by mutating.** It planted the
`README`/lowercase adjacency on the *pod* — but the probe answers exactly the *staged* paths, so a
pod-only file never reaches the join. Both sides of the inversion are now staged, with one `ABSENT`
to supply the unpairable line GNU join needs before it arms the order check.

## Process note, recorded because it happened three times

I destroyed my own uncommitted work again with `git checkout -- <file>` after a mutation run — it
restores from the **index**, so mutating uncommitted work and reverting takes the work with it.
All seven fixes had to be re-applied. Commits now precede every mutation run.

```audit-claims round=1 audited=3542060..ec16cce8
1. The pre-flight probe now answers EVERY staged path (hash / ABSENT / UNREADABLE); a short reply is exit 9, and the counts print on every path including success.
2. Refusing on "0 present" was tried and reverted — it broke 18 legitimate first-seed pushes; the gate is answered-count, not presence-count.
3. The local hash side uses `-I{}` so a path with a space or quote no longer aborts with rc 123/1 and no diagnostic.
4. The refusal message no longer prescribes `cairn sync`/`cairn put`, neither of which can clear it; it names the per-entry API and says a push is probably the wrong verb from this host.
5. The header states the guard covers ~12% of what the push overwrites (211 of 1755), naming the .git repos it does not cover.
6. An unreadable pod file answers UNREADABLE and lands in the clobber set instead of being treated as a pure addition.
7. `seed: pushing` is now `seed: preparing push`, and `--allow-overwrite` is documented in Usage.
8. Two new test classes; the LC_ALL=C join mutant that previously survived all 40 seed tests is now killed.
```

**Still not verified:** neither gate tier on the merged tree, and nothing against the live pod.

ZacxDev:
## Round 2 — not clean. My own round-1 claim was refuted, and both halves were wrong.

Round 1 claimed `-I{}` made a path with a space **or a quote** survive. Neither had a test, which
is how the claim survived being written down.

- **QUOTE — not fixed.** `-I{}` stops word-*splitting* but leaves xargs's input **quote parsing**
  on. `sc/it's.md` still died `xargs: unmatched single quote`, rc 1, with no `seed:` line —
  byte-identical to base. Only `-d`/`-0` disables it. Now `-d '\n'` on both sides.
- **SPACE — stopped aborting and started LYING**, which is worse. `awk '{print $2" "$1}'` rebuilt
  the line from *fields*, truncating the key at its first blank. Two **byte-identical** entries
  collapsed to one key, the join degenerated into a cross-product, and the guard printed
  `differing=2` and **refused — naming a path that does not exist**, offering `--allow-overwrite`
  as the remedy. A confident false refusal is worse than the crash it replaced. Key is now
  tab-separated, with `join -t <tab>` so join's own splitting can't re-introduce it.

Also fixed: `grep -v … || :` swallowed grep's **exit 2 (I/O error)** as well as its exit 1
(no-match) — an error left the join empty, `differing=0`, and the push proceeding, i.e. the
silent-zero shape one line below the gate that removes it. And a probe comment still described a
bare `|| :` that two rounds ago became an `if/else`.

`--help` ended **mid-sentence** and has never reached the Usage block, so round 1's
"`--allow-overwrite` is documented in Usage" was true of the file and invisible where an operator
looks. Now prints the whole leading comment block.

### Mutation matrix — re-run

| mutant | result |
|---|---|
| local loses `-d '\n'` (quote case) | **killed** |
| `awk` truncates the join key (space case) | **killed** |
| `LC_ALL=C join` → `join` | **killed** (survived all 40 base tests before round 1) |
| drop the answered-count gate | **killed** |

`test_subsystem_store_api.py`: **723 passed**.

### ⚠ Needs an operator decision, not a code change from me

**`cairn-cutover.py` P3 invokes `seed.sh` without `--allow-overwrite`**, and its shippable set is
`ADD + SUPERSEDES + MERGED` — where `SUPERSEDES`/`MERGED` are *by definition* entries whose pod
bytes differ. So the pre-flight refuses (exit 8) and **P3 can never complete**. The refusal predates
this round; what this round changed is the guidance it prints, which now tells that caller the local
tree is "a FROZEN pre-cutover mirror" and to "send it entry-by-entry" — false for a curated delta
that has already written its rollback set. Nothing exercises P3 against the real `seed.sh`, so 85
green cutover tests say nothing about it.

Options: pass `--allow-overwrite` from P3 (it *is* a reviewed delta with a rollback set), or accept
that P3 is dead post-cutover and say so. I have not chosen for you.

```audit-claims round=2 audited=ec16cce8..HEAD
1. `-d '\n'` on both xargs sides — a quoted path no longer aborts rc 1 with no diagnostic.
2. The join key is TAB-separated and `join -t <tab>` — a spaced path is no longer truncated into a cross-product false refusal.
3. `grep -v … || [ $? -eq 1 ]` — an I/O error is no longer swallowed as a no-match.
4. The probe comment now describes the if/else that is there, not the `|| :` it replaced.
5. `--help` prints the whole leading comment block, so it ends on a complete line and reaches Usage.
6. Three new tests cover identical-spaced (no false refusal), differing-spaced (still caught, named in full), and quoted (no abort).
7. NOT FIXED, flagged for a decision: cairn-cutover.py P3 is blocked by the exit-8 refusal.
```

**Still not verified:** neither gate tier on the merged tree, and nothing against the live pod.

PR #1310, opened 2026-09-05
feat(clawgate): host-side feeder for the Claude Code transcript read model (rank 25)

ZacxDev:
The host half of rank 25 of the tmux-webapp effort. The server half — the read model, the ingest routes and the chat view — is **ZacxDev/homelab-infra#695**.

🔴 **This is a no-op without #695**, by design: the feeder treats a 404 from either route exactly as it treats an unreachable server, so deploying in either order degrades rather than errors.

## What this adds

- `scripts/transcript-push.sh` — credentials, the two HTTP calls, a distinct exit code per condition.
- `scripts/lib/build_transcript_push.py` — the one decision worth testing: **which** sessions to send and **how much** of each.
- `systemd.user.services.transcript-push` + a 5-minute timer in `nix/home.nix`, behind `enableTranscriptPush`.

## Decisions a reviewer should want to challenge

**1. 🔴 ONE UNIT PER HOST — the opposite of its sibling.** `tmux-snapshot-push` is `serverMode`-gated (workbench only) as a *correctness* requirement: its collector already reaches both machines over ssh, so a second reporter would fight over every row. Transcripts have no such collector — the files are local and unreadable from the other host — so a `serverMode` gate here would feed exactly half the fleet and **every laptop session would render "no transcript stored" for ever, silently**. The read model is keyed on a globally-unique Claude Code session id, so two hosts pushing disjoint sets cannot collide.

This is the single thing a copy-paste of the sibling unit would get wrong, so `test_the_unit_is_NOT_serverMode_gated` pins it — *and* asserts the sibling's gate is still different, so the contrast cannot go stale.

**2. 🔴 The walk is the SHARED enumerator, and that fix is a correctness fix.** The first revision open-coded its own `*.jsonl` walk. `test_transcript_search.py::test_the_jsonl_glob_site_ledger_is_pinned_two_way` caught it — that ledger SCANS the tree for walk sites rather than naming files, precisely so a fourth hand-rolled walk cannot pass unseen.

The fix is not hygiene. `transcript_search.is_corpus_member` excludes `subagents/`, which holds transcripts that are **not resumable sessions**: measured 2026-09-04, **4,884 of the 5,788** `.jsonl` files under `~/.claude/projects` — **84%** — live there, and nothing can ever join them to anything the chat view is reached from. My walk missed them only because it descended one level, i.e. it was right *by accident of its depth*. A later "make this recursive" edit would have fed clawgate thousands of unreachable rows and nothing would have gone red.

**3. A bounded TAIL, never a transcript.** 315 files modified in 24h, 456 MB, largest 23 MB. Client defaults (192 KiB, 6 sessions) sit deliberately **under** the server's caps (256 KiB, 8) — a client tuned exactly to the limit turns any rounding disagreement into a feeder that fails every tick while looking correctly configured. `test_the_client_bounds_sit_UNDER_the_servers_not_at_them` reads the server's numbers as literals, so a clawgate-side change fails loudly here.

**4. Dedupe via a SERVER-AUTHORITATIVE digest.** A host-side "what I sent last time" cache is wrong in both directions: a DB restore leaves the server empty while the host believes everything is delivered (silent, permanent, invisible); a wiped cache re-pushes everything. Asking the server is correct both ways and needs no state on this machine.

**5. A failed pre-flight is FATAL, not "assume the server has nothing".** Treating an unreadable digest as empty is the expensive direction — every recent session re-pushed every tick — and it fires precisely when the server is least able to absorb it.

**6. "Nothing changed" is rc 0, signalled by an exit code, not an empty document.** The server rejects a push carrying no sessions (correctly), so emitting one would make the ordinary steady state an HTTP 400 every tick: a feeder reporting failure exactly when it works.

**7. No `OnFailure` toast.** `notify-failure@` defeats do-not-disturb and is justified by a measured ~1 firing in 9 days; at a 5-minute cadence any sustained outage would fire one every tick. The compensating control is `Type=oneshot` with distinct codes landing in the failed-unit list `/standup` reads — **plus**, unlike the sibling, a consumer that renders its own staleness (`fed <n> ago`).

**8. The `PATH` is deliberately shorter than the sibling's** — no tmux/openssh/gawk, because this reads local files. Pinned.

## Verification matrix

Base: merged with `origin/main` @ **`f75c92d4`**. Tree under test: **`6da75cb1`**.

| tier | command | result |
|---|---|---|
| dev-host, both tiers | `nix develop … scripts/gate.sh --tier both` | **`GATE: RESULT=PASS exit=0`** |
| ↳ pytest | (from the runner's own summary) | **collected=21508, passed=21505, skipped=3, failed=0** (floor 18555) |
| ↳ node | (from the runner's own summary) | **suites=5 files=41 tests=1449 pass=1449 fail=0 skipped=0** (floor 1367) |
| this suite alone | `pytest scripts/tests/test_transcript_push.py` | **32 passed** |
| nix syntax | `nix-instantiate --parse nix/home.nix` | clean |
| sandbox tier | `nix build .#checks.x86_64-linux.pytests` / `.nodetests`, **one at a time** | see the comment below |

🔴 **The sandbox tier is a DIFFERENT tier from `gate.sh`, not a second spelling of it** — it builds from a `cp -r` store copy with no `.git`, and it is what CI runs. Its result is posted as a comment rather than asserted here.

### Three tests were watched to FAIL first — all three were my own bugs

- `test_the_unit_does_not_wire_the_DND_defeating_failure_toast` failed against the very unit it was written to approve, because the block *explains at length why there is no `OnFailure`*. A guard that reads prose is not a guard on configuration — it would equally pass a unit that wired `OnFailure` while calling it something else in a comment. Now strips comment lines, with a positive control that the stripper can still see a real setting.
- `test_the_builder_never_emits_invalid_utf8` used the default `json.dumps`, which escapes non-ASCII to six ASCII bytes — the fixture was pure ASCII, no byte offset could land mid-rune, and it would have passed while exercising nothing. It now writes real multi-byte bytes and **searches** for an offset that provably splits a rune, failing loudly if none exists. Its own control then caught a hardcoded offset that happened to be rune-aligned.
- The glob-site ledger failure above, which was a real defect and not a test bug.

## What is NOT covered

- **Nothing is deployed.** No `home-manager switch`, no `ship.sh`. The timer is armed in config only.
- **The two halves have never run against each other.** This suite drives the real script against a stub HTTP server on loopback; #695's drives real routes against fake stores. Neither exercises the live pod.
- **No measurement of the real steady state.** The dedupe saving is argued from the design and asserted against a stub; the actual bytes/tick on the live fleet are unmeasured.
- **`project_of` is a LABEL, not an identifier.** The directory name is a slugified cwd and the slug is lossy, so it is deliberately not un-slugified; the server prefers the `cwd` the records carry.
- **A session resumed on the other host** flips the row's `host` column. No test covers that transition.
- **`TRANSCRIPT_PUSH_HOST` falls back to `ACTIVITY_HOST` then `uname -n`.** Nothing pins that `ACTIVITY_HOST` is set on both machines.
- **The tail is read while the file is being appended to.** That is the normal case, not an edge one; the code says so and claims no consistency beyond "a tail as of some instant".

https://claude.ai/code/session_01LBuEtZn6AaCQFy6PjWUrjQ

### DIFF
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -105,6 +105,24 @@ let
   # SSHes to the laptop), so running it on the laptop too would have each host pushing
   # a full two-host document and the two would fight over every row.
   enableTmuxSnapshotPush = true;
+  # Claude Code TRANSCRIPT feeder master switch (scripts/transcript-push.sh) — feeds
+  # clawgate's chat view. Gates ONLY whether the timer is wired into timers.target; the
+  # SERVICE definition is always emitted, so `transcript-push` can be started by hand
+  # regardless.
+  #
+  # 🔴 NOT serverMode-GATED, AND THAT ASYMMETRY WITH enableTmuxSnapshotPush ABOVE IS THE
+  # WHOLE POINT. That one is workbench-only as a CORRECTNESS requirement: its collector
+  # already reaches both hosts over ssh, so a second reporter would have the two fighting
+  # over every row. Transcripts are ordinary files on each host's OWN filesystem, readable
+  # from nowhere else, so a workbench-only unit would feed exactly half the fleet and the
+  # laptop's sessions would render as "no transcript stored" for ever. The read model is
+  # keyed on a Claude Code session id, which is globally unique, so two hosts pushing
+  # disjoint session sets cannot collide.
+  #
+  # ON from the start: the write path is a latest-per-session upsert, a bad push is
+  # rejected outright (the server 4xx's and CHANGES NOTHING, leaving the previous tail in
+  # place), and the server sweeps rows older than 7 days.
+  enableTranscriptPush = true;
   # Graphical host = runs X/i3 (both current NixOS hosts do; only a genuinely headless
   # box would not). Approximated as isNixOS, mirroring graphical.nix — deliberately NOT
   # !serverMode, which is true on the graphical workbench.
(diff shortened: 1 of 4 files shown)


ZacxDev:
## Sandbox tier — the one CI actually gates on — PASS, run ONE AT A TIME

Base: merged with `origin/main` @ `f75c92d4`. Tree: `6da75cb1`.

```
nix build .#checks.x86_64-linux.pytests    -> rc 0
  devrc-pytests>   TOTAL collected=21508  passed=21505  skipped=3  failed=0  (floor: 18555)

nix build .#checks.x86_64-linux.nodetests  -> rc 0
  devrc-nodetests> TOTAL suites=5 files=41 tests=1449 pass=1449 fail=0 skipped=0  (floor: 1367)
```

🔴 **Two invocations, deliberately, not one combined `nix build a b`.** A combined call builds both concurrently and the tests that shell out to nested `nix` then contend on the store, producing measured FALSE failures (`SQLite database … is busy`, `database is locked`). A combined *green* is trustworthy; a combined *red* is not — so the honest thing is to never produce one.

🔴 **Read from the runners' own `RESULT:` lines, not from a piped exit code.** Both agree with the build status.

This is a **different tier** from the `gate.sh` run in the PR body, not a second spelling of it: `gate.sh` runs on the dev host against the live checkout, while these build from a `cp -r` store copy with **no `.git`** — so anything keyed on the repo being a git checkout evaluates differently. Both are now green on the same merged tree.

⚠ The box was at load ~24 on 24 cores throughout. That inflates wall time; it does not affect the counts, and the two derivations were run sequentially so neither contended with the other.

PR #1311, opened 2026-09-05
feat(tmux-restore): bind a window by the per-pane LEDGER, not a 145-file grep

ZacxDev:
## The problem

`scripts/tmux-session-restore.py` snapshots which claude conversation is in which tmux window so they can be resumed after a reboot. Binding a window to its session id was done by **fuzzy pane-content matching**: capture the last 200 pane lines, keep fragments >=40 chars, and `grep -lF` each across every `*.jsonl` in that cwd's project dir, keeping only fragments that hit exactly one file.

It is slow and it misses. Measured on the workbench 2026-09-04: **145 competing transcripts** in `-home-zach-workspace-devrc` alone, and a `save` over 50 windows took **2m15s**, essentially all of it in that grep.

## The fix

A deterministic per-pane ledger **already existed and was not being read**. `scripts/claude-hooks/agent-ledger-hook.py` is deployed, wired on 4 hook events, and writes one JSON file per pane to `~/.cache/agent-ledger/claude-p<N>.json` carrying `{pane_id, session_id, transcript_path, window_id, tmux_pid, last_activity_ts}`. The hook receives the real `session_id` from Claude Code and reads `TMUX_PANE` from its own environment — a record, not an inference.

`build_plan()` now runs **two passes**: pass 1 reads the ledger; pass 2 runs the existing grep only for panes pass 1 could not answer. Ledger claims first, so a guess can never take an id the ledger owns, and the grep is never *called* for a ledger-bound pane.

**Four validations**, each with its own reason token: `no-session-id`, `transcript-missing`, `generation-mismatch`/`generation-unmeasured`, `project-mismatch` (the cross-repo guard). `last_activity_ts` deliberately does **not** gate — see below.

## Measured, same-instant head-to-head (44 live claude panes, one `tmux list-panes` snapshot)

```
ledger lookup, all 44 panes:      0.002s  ->  34/44
pane-content grep, all 44 panes: 176.606s ->  34/44

both  27  (agree 27, disagree 0)   ledger only 7   fuzzy only 7   neither 3
union 41/44
```

🔴 **The ledger is NOT better at coverage — it ties at 34/44**, and the two miss different panes. Its wins are that it is a record rather than an inference, ~5 orders of magnitude cheaper, and **complementary**: together they bind 41/44 where either alone binds 34. Where both answered they agreed 27/27, an independent cross-check neither gives alone.

`save` wall time, full runs to a scratch plan: **180.5s / 35 bound -> 37.8s / 42 bound** — 4.8x, +7 windows. (Runs minutes apart, so counts differ from the snapshot above by window churn; stated as measured.)

⚠ **The 10 `no-record` panes are pre-hook sessions — but that set does NOT simply "shrink on its own".** `agent_ledger.DEFAULT_MAX_AGE` is 7 days and prune sweeps the whole directory, so it keeps re-opening for exactly the long-idle windows. `no-record` has a permanent floor.

⚠ **`tmux_pid` equality is NOT an exact generation check.** A reboot resets the pid counter, so a colliding record for the same pane id in the same cwd would pass all four validations. Narrow in practice, documented, and **accepted** — closing it needs a writer-side change to `agent_ledger.py`.

**Post-reboot path, exercised rather than reasoned:** with a foreign server pid, `{'generation-mismatch': 33, 'no-record': 10}`. The fallback fires for 100% of panes and the first post-reboot `save` costs the full grep — no regression, no improvement in that window.

## Why `last_activity_ts` does not gate

Any age threshold would reject only *correct* bindings, hardest for the long-idle windows most worth restoring. Corroborated: 18 ledger-bound panes cross-checked against the real grep — **18 agree, 0 disagree**, including the oldest record at 5.3 days, where a 24h gate would have rejected 3 correct bindings.

## Verification

- **RED -> GREEN**: 23 failed/13 passed at `cc409f82` -> **47 passed** at head.
- **Mutation**: every guard killed by its own assertion, under `PYTHONDONTWRITEBYTECODE=1` with a fresh tree per mutant, a green control and a known-fatal positive control.
- **Both gate tiers on the merged tree** (`origin/main` `f75c92d4` is an ancestor, so head IS the merged tree), run one at a time: `pytests` -> `RESULT: PASS (exit=0)`, `collected=21509 passed=21506 skipped=3 failed=0`; `nodetests` -> `RESULT: PASS (exit=0)`, 0 fail; 0 timeout panics in either.

## Four adversarial audit rounds

No 🔴 at any round. Rounds 1-3 each found real defects, none of which a green suite could catch; **round 4 came back clean**, which is what ended the ladder.

1. The loader promised graceful degradation but caught only import *failure* — a module importing without `pane_filename` raised `AttributeError` out of `cmd_save`, **silently freezing the restore plan**. Fixed with a borrowed-symbol guard.
2. The four reason tokens were computed and discarded; `cmd_save` now prints them.
3. `LEDGER_DIR` was a *second* borrowed symbol, unguarded, with a restated literal default — and its failure rendered as `no-record`, the token the legend reserves for the benign case. Now `_BORROWED = ("pane_filename", "LEDGER_DIR")`, read directly off the module, literal default gone.
4. Two prose claims were false about the code they described (a pid-collision inference its own measurement refuted; a prune rule false about one of the two writers). Corrected at all three copies.
5. The `_BORROWED` two-way pin was **blind to `getattr(_AL, "X", …)`** — the exact spelling `LEDGER_DIR` had been written in — while its docstring claimed it closed that hole. Now walks the AST; the surviving mutant is killed and prose is no longer scanned as code.

Payload lines changed across the fix rounds: 75, 67, 34.

https://claude.ai/code/session_01D2HwZGfRQSHQTJxjh7mkfy

### DIFF
--- a/claude/skills/initiative-scan/SKILL.md
+++ b/claude/skills/initiative-scan/SKILL.md
@@ -59,9 +59,9 @@ skill instead. This skill is the ephemeral scan that feeds it.
 If `$ARGUMENTS` is **`snapshot`** (or `save`), **`restore`** (optionally `restore --dry-run`), or **`show`**, run the workspace snapshot helper instead of the scan — it binds each live claude tmux window to its exact session id (the per-pane agent ledger first — a RECORD; a pane-content grep only for what the ledger cannot answer — a GUESS) so you can bring the whole workspace back after a reboot. tmux-continuum already restores the sessions/windows/cwds; this relaunches the right `claude --resume <id>` in each.
 
 ```bash
-python3 ~/workspace/devrc/scripts/tmux-session-restore.py <snapshot|restore|show> [--dry-run]
+python3 ~/workspace/devrc/scripts/tmux-session-restore.py <save|restore|show> [--dry-run]
 ```
-- **`snapshot`** — run BEFORE rebooting: writes `~/.config/initiatives/restore-plan.json` + a readable `restore-cheatsheet.md` (survives reboot). Present the cheat-sheet. Each entry carries `bind_source` and the sheet prints it per line (`(ledger)` = certain, `(fuzzy)` = a content guess to eyeball); the summary's `bound: N ledger, M pane-content` + `ledger reasons:` tally say why anything is unbound.
+- **`save`** (the script's verb for a snapshot) — run BEFORE rebooting: writes `~/.config/initiatives/restore-plan.json` + a readable `restore-cheatsheet.md` (survives reboot). Present the cheat-sheet. Each entry carries `bind_source` and the sheet prints it per line (`(ledger)` = certain, `(fuzzy)` = a content guess to eyeball); the summary's `bound: N ledger, M pane-content` + `ledger reasons:` tally say why anything is unbound.
 - **`restore`** — run AFTER reboot (once tmux-continuum has restored the shells): relaunches `claude --resume <id>` in each window; windows already running claude are skipped, and windows with no certain match fall back to the interactive picker. Use `--dry-run` first to preview.
 - **⚠ host-local:** run it on the host you're rebooting — it reads that host's live tmux + `~/.claude/projects`. The plan is per-host.
 
--- a/claude/skills/initiative-scan/SKILL.md
+++ b/claude/skills/initiative-scan/SKILL.md
@@ -56,12 +56,12 @@ skill instead. This skill is the ephemeral scan that feeds it.
 
 ## Session snapshot / restore (survive a reboot)
 
-If `$ARGUMENTS` is **`snapshot`** (or `save`), **`restore`** (optionally `restore --dry-run`), or **`show`**, run the workspace snapshot helper instead of the scan — it binds each live claude tmux window to its exact session id (by matching pane content) so you can bring the whole workspace back after a reboot. tmux-continuum already restores the sessions/windows/cwds; this relaunches the right `claude --resume <id>` in each.
+If `$ARGUMENTS` is **`snapshot`** (or `save`), **`restore`** (optionally `restore --dry-run`), or **`show`**, run the workspace snapshot helper instead of the scan — it binds each live claude tmux window to its exact session id (the per-pane agent ledger first — a RECORD; a pane-content grep only for what the ledger cannot answer — a GUESS) so you can bring the whole workspace back after a reboot. tmux-continuum already restores the sessions/windows/cwds; this relaunches the right `claude --resume <id>` in each.
 
 ```bash
 python3 ~/workspace/devrc/scripts/tmux-session-restore.py <snapshot|restore|show> [--dry-run]
 ```
-- **`snapshot`** — run BEFORE rebooting: writes `~/.config/initiatives/restore-plan.json` + a readable `restore-cheatsheet.md` (survives reboot). Present the cheat-sheet.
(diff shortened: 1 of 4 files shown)


ZacxDev:

`origin/main` moved during the first round (`cc409f82` -> `f75c92d4`), so `origin/main` was merged in (`771735dc`) and everything was re-run against that tree. Base sha for the claims below: **`f75c92d4`**; head **`771735dc`**.

**Tier 1 — dev host, `scripts/gate.sh --tier both`**
```
GATE: RESULT=PASS exit=0
```

**Tier 2 — sandbox, `nix build .#checks.x86_64-linux.{pytests,nodetests}`, built ONE AT A TIME**, verdicts read from `nix log`, not from a piped exit code:
```
pytests   RESULT: PASS (exit=0)   collected=21498 passed=21495 skipped=3 failed=0  (floor 18555)
nodetests RESULT: PASS (exit=0)   suites=5 files=41 tests=1449 pass=1449 fail=0    (floor 1367)
```

The known non-hermetic dev-host failure (`test_clawgate_task_interview_guard.py::test_a_body_file_written_by_a_heredoc_on_the_same_line_is_read`) did **not** fire on this host in these runs — `/tmp/body.md` was absent — so it is not among the results above in either direction.

### Not verified

- **The post-reboot path was not exercised.** Restoring after a real tmux server restart would make every ledger record fail the `tmux_pid` check, and the whole workspace would fall back to the grep — which is the correct and intended behaviour, but it was reasoned about, not observed. Rebooting to test it was out of scope for a read-only check.
- The live measurements are **one host, one instant** (workbench, 2026-09-04, 44 panes). The laptop was not measured.

ZacxDev:
## ⚠ Correction: this PR body is stale against the current head

Two adversarial audit rounds have run since the description was written. Posting this rather than silently editing the body, because the body is what gets squashed into `main` and anyone who already read it read the wrong version. **The body will be corrected before merge; this comment is the record of what changed.**

Six claims in the description are no longer true at head `2eb06a6d`:

1. 🔴 **"that set shrinks on its own"** (of the 10 `no-record` panes). True only of panes predating the ledger hook. `agent_ledger.DEFAULT_MAX_AGE` is 7 days, so prune keeps re-opening the set for exactly the long-idle windows — `no-record` has a permanent floor. Corrected in the docstring; the body was not touched.
2. 🔴 **"so equality is an exact generation check"** (validations table). This is the round-1 finding verbatim, uncorrected. A reboot resets the pid counter, so equality is *not* exact across generations. The residual is documented and accepted, not eliminated.
3. **"GREEN at HEAD: 36 passed"** — head is now **44 passed**.
4. **"Every mutant killed"**, over a table listing none of the three that had in fact SURVIVED (`or "fuzzy"` → `or "ledger"`; dropping `isinstance(rec, dict)`; `cmd_save` counting ledger as fuzzy). All three are killed now, each by its own assertion — but the table never said they had survived.
5. **"If that module cannot be *loaded* there is deliberately no fallback spelling"** — pre-`hasattr` wording. The `imports-but-unusable` case (module present, symbol missing) was the round-1 finding and is not mentioned.
6. **"`build_plan` entries gain a `bind_source` field"** — they now also gain `ledger_reason`. The test list omits all 8 tests added since.

### Also corrected in the round-1 fixes, for the record

- The loader promised graceful degradation but only caught import *failure*, not a module importing without `pane_filename` — which raised `AttributeError` out of `cmd_save`, silently freezing the restore plan. Fixed with a `hasattr` guard.
- The four reason tokens were computed and then discarded (`sid, _reason = …`). `cmd_save` now prints them.

### Still open at the time of this comment

A round-2 delta found the prose fixes for the two items above were themselves wrong in places, plus a real guard gap: `pane_filename` is not "the one symbol this file borrows" — `LEDGER_DIR` is borrowed too, unguarded, and its failure mode renders as `no-record`, the token the legend reserves for the benign case. Fixes in flight.

**No behaviour defect has been found in either round.** Every finding has been documentation-of-a-guard, a stale claim, or the guard gap above.

PR #1313, opened 2026-09-05
feat(mentions): attribute a bare #N to its repo, widen detection for TELEMETRY only, and stop the terminal handler dead-ending

ZacxDev:
## What and why

92% of the mentions detected in a measured 24h window were an ambiguous bare
`#N` — and the commonest shape, `<repo> PR #1291`, threw away the repo name two
words to its left. That **misattribution**, not the genuinely missed references,
is roughly 10x the volume. This PR fixes the attribution first and widens
detection second, and it does the widening for the **telemetry** consumer only.

## 1. Attribution (the primary change)

Every candidate and span now carries `repo` (an `owner/repo`, or `""`) and
`repo_source` naming HOW it was resolved, in a documented priority order:

| source | evidence | profile |
|---|---|---|
| `explicit` | an `owner/repo#N` written out | both |
| `mapped` | a bare `repo#N` — the text named the REPO, the OWNER came from the caller-MEASURED `repos` mapping | both |
| `adjacent` | a repo token immediately before the ref, looked up in the caller-MEASURED `repos` mapping | telemetry |
| `url` | a `github.com/owner/repo/(pull\|issues)/N` in the same text | telemetry |
| `flag` | a `--repo owner/repo` in the same text | telemetry |
| `default` | the caller-supplied `default_repo` | both |

🔴 **`explicit` means the TEXT wrote the owner out, and nothing else.**
`trowelcast#591` and `gardenersguild/trowelcast#591` are different evidence: the
first is a name our mapping resolved, the second is a repository the operator
stated. Reporting both as `explicit` — which round 1 of this PR did — makes an
analysis that cannot separate "the text said so" from "our mapping said so", i.e.
one measurement pretending to be two. `mapped` is the second of them.

🔴 **Nothing is guessed, and the no-guessing rule is *extended*, not relaxed.**
A repo token absent from the mapping leaves the ref unattributed — there is no
default org. A block naming **two different** repositories attributes **nothing**:
several is an absence of an answer, not a tie to break by position. And a ref
that NAMES a repo is never re-attributed from the block, because substituting a
different repository for the one the operator wrote is worse than admitting the
owner is unknown.

Attribution does **not** settle the platform. Which repository a `#N` belongs to
and whether it is a GitHub issue or a clawgate task are different questions;
answering the first is not a claim about the second, so the span stays ambiguous.

### 1b. 🔴 The dedupe key carries the attribution, and the migration is CHOSEN

The tailer keys its per-session ledger on the mention's identity. Keying that on
`platform:raw` alone throws away the very thing this PR computes: 92% of mentions
are a bare `#N`, so `raw` is `#1291` for nearly all of them and the FIRST
occurrence in a session wins. Measured on real transcripts: **34 rows/day shipped
`repo=""` although an attribution was available** (15% of the rows that had one),
and **6 rows/day dropped a second repository's reference outright** —
`trowelcast PR #1291` and `plotwidget PR #1291` collapsing to one row. Because
the bare form usually appears first, the loss was biased systematically downward.

🔴 **THE SUFFIX IS CONDITIONAL, AND THAT IS THE MIGRATION DECISION.** An
unattributed mention keys **byte-identically to before** (`platform:raw`); only
an attributed one takes the `@owner/repo` suffix. The deployed tailer passes NO
mapping (`collect_mentions(objects)`), so **every key in every host's
`session-summary-state.json` today was written with `repo == ""` and still
matches** — nothing already emitted re-emits. The obvious alternative, an
unconditional `platform:repo:raw`, would have re-keyed the lot and re-emitted
every already-shipped mention once on both hosts. What DOES re-emit once under
the chosen format is exactly the set that newly gains an attribution — a row
whose content genuinely changed, which is the row this PR exists to ship.
(`@` is safe as the separator: no pattern in `mention_scan` can put one in `raw`.)

## 2. Detection — telemetry profile only, ENUMERATED

`github.com/owner/repo/(pull|issues)/N` · `/audit-pr N` and `audit-pr N` ·
`gh (pr|issue) <enumerated-subcommand> N` · `task N` **only** behind the literal
`clawgate` · the legacy `#task-N` anchor.

Deliberately **not** added, measured rather than assumed: a bare `task N` (179
occurrences in one 24h window, overwhelmingly prose), a bare `PR N`, and git
shas — a `[0-9a-f]{7,12}` probe over the same window returned ~520,000 hits.
Each of those is a generic pattern, which is the shape the module's own
`DEV-123` note already refuses.

## 3. The profile split is EXPLICIT — and now ENFORCED at the consumer

This deliberately relaxes the module's "one set of regexes, they can never drift
apart" invariant, along one axis only. `PATTERN_LEDGER` names every compiled
pattern with the profiles it is consulted in and is pinned **two-way** by the
suite — a pattern with no entry fails, an entry naming no pattern fails.
`NON_SCAN_PATTERNS` is the only other home a pattern may have and is pinned too,
so "in neither" is a failure rather than a gap. The default is `terminal`, so
`scripts/mention-open.py` keeps today's narrow click-safe surface without saying
anything, and every pre-existing negative case still holds.

🔴 **That default was pinned at the module and NOWHERE at the four sites that
consult it.** An independent mutation sweep deleted each of the three
`if "<NAME>" in on` attribution guards, and added `profile="telemetry"` to
`mention-open.py`'s only `scan_mention_spans` call — putting the WIDE profile on
the terminal click surface — with **all 313 tests green** for each. All four are
now pinned behaviourally, each with a positive control proving the same fixture
DOES attribute (or DOES match) at the telemetry profile, so none of them can pass
by scanning nothing. The click-surface pin is driven off `PATTERN_LEDGER` itself,
so a telemetry-only pattern added later is covered without anyone remembering.

## 4. 🔴 The inert-prefilter trap

`session-tailer.py` had `_MENTION_HINTS = ("#", "868")`, a cheap short-circuit
applied **before** the regex pass that skipped ~4 assistant text blocks in 5.
Every new shape — `audit-pr 1291`, `gh pr view 1291`, `clawgate task 370`, a
github URL — contains **neither** literal, so adding the regexes alone would have
shipped a **completely dead feature** that still passed any unit test calling
`scan_mentions()` directly.

It is now **derived** from the same ledger, at the **telemetry** profile (using
the default would return the old two literals and look correct). The suite proves
each shape reachable through `collect_mentions`, with a control asserting each
fixture really was invisible to the old value — without that control the
reachability test could pass against a prefilter that never moved.

⚠ **The percentage is a property of the WINDOW, not of the code**, and this PR
previously quoted two different windows as if they were one. Re-measured over the
24h to **2026-09-04** (10,118 assistant text blocks across 312 transcripts): the
OLD two literals skip **8,169 = 81%**, the derived telemetry set skips
**7,990 = 79%** — the short-circuit survives the widening. The earlier
development window (6,052 blocks) read 82% → 80%. Both source comments now carry
the counts and the caveat.

## 5. 🔴 Disclosure: BOTH consumers of `known_repos.json` are now guarded

`known_repos.json` names private repositories — the file whose committed ancestor
disclosed 232 of them into this PUBLIC repo (#1283). This PR newly wires the
**tailer** to it, and the asymmetry that shipped in round 1 was backwards: the
hint handler, which sends the mapping to a rofi window that closes, had a guard;
the tailer, which sends it into a **durable ClickHouse table**, had none.

* `mention-open.py`'s guard **stubbed `notify`** — the only function in the module
  that writes to stderr or `notify-send`. MEASURED: with the universe added to the
  **refusal** path's `notify` body (reachable — `--print` skips PASS 4 while
  `discovered` holds every private name), the pre-fix suite reported **92 passed**.
  It now stubs `subprocess.run` instead, keeps the real `notify`, enumerates every
  sink in one helper, and covers the refusal path as well as the picker path.
* `session-tailer.py` gains its first such guard, at the SPOOL: the run's raw
  spool bytes, the decoded events, stdout and stderr, on an attributed run and on
  an unattributed one. MEASURED: the pre-fix suite reported **106 passed** with
  the whole mapping appended to `b64:repo`, and again with it in `context`.

Both guards lead with their positive controls — the attributed repo IS present,
the mapping the run held really did have the other entries in it — so a clean
verdict is never a scan wired to nothing.

## Also: the terminal handler no longer dead-ends (operator request)

Separate axis from the detection split — this is resolution **UX**, not what gets
underlined. Three dead ends now open a rofi picker with `-matching fuzzy` instead
of refusing:

1. **The namesake cap is gone.** It refused above 8, arguing "a 100-row list of
   URLs differing only by owner is not a choice, it is a wall". True of a list you
   can only SCROLL, false of one you can TYPE AT — `kubernetes#1` is now four
   keystrokes from the right owner. That comment is replaced in the same change,
   because a comment left asserting a hazard the code has closed is how the
   refusal comes back.
2. **A search that matched nothing** offers the host's known-repo universe, so
   `talos-inf#12` can still reach `talos-infra`.
3. **A bare `#N` nothing could attribute** appends the universe *below* the
   clawgate row, so the 92% case is still one Enter away.

A picker is a CHOICE, not a guess. 🔴 A universe row is **never opened without a
selection, even when it is the only one** — the first version of this did open it,
and a one-repo host would have answered `zzznosuchrepo#77` by opening issue 77 in
an unrelated repository. Dismissal still opens nothing. `--print` and
`--no-discovery` never reach the universe. An unreadable or empty universe
degrades to the original refusal, which still names WHICH empty it is; and a
search that could not RUN is announced **alongside** the picker rather than being
swallowed by it.

⚠ **Removing the cap changed `--print`, and that is DECLARED rather than
discovered.** `--print widget#12` with nine namesakes used to exit 1 with a named
refusal; it now prints nine URLs and exits 0. **Intended**: `--print`'s documented
contract is "print every candidate when ambiguous", a namesake set is exactly that
ambiguity, and every row is an EXACT-name search hit — evidence about the name the
operator typed. It is **not** the PASS 4 universe, which is evidence about nothing
and stays barred from `--print`. A consumer wanting one answer writes
`owner/repo#N`. Pinned by a test, at nine rather than eight so a fixture cannot
confuse "no cap" with "a cap nobody reached".

## Measured — the development 24h window, counts only

```
assistant text blocks (non-sidechain): 6,052   chars: 3,068,092

spans detected   BEFORE 2,088   ->   AFTER 2,294        (+206, +9.9%)
   ambiguous      1,926              1,926   (unchanged, by design)
   github           151                347   (+196)
   clickup           11                 11
   clawgate           0                 10   (new)

spans carrying an owner/repo   BEFORE 27  ->  AFTER 475   (17.6x)
   by source: url 323 · explicit 103 · adjacent 35 · flag 14
```

⚠ The `by source` split predates the `explicit`/`mapped` separation, so its
`explicit 103` is the MERGED figure the fix above exists to break apart. It has
not been re-measured; the number is left as-is rather than restated with a
distinction it does not carry.

## Deploy note

🔴 **CORRECTED — the earlier version of this note was wrong in the direction that
misleads verification.** It said both runtime files are `home.file` targets, so
nothing changes until a switch. That is true of the telemetry half only.

| half | deployed as | when it changes |
|---|---|---|
| telemetry — `mention_scan.py` + `session-tailer.py` under `~/.config/activity-collector/` | `home.file` **store copies** (verified `readlink -f` → `/nix/store/…`) | only on `home-manager switch` (`scripts/ship.sh`) |
| terminal — the Alacritty hint | a `writeShellScript` wrapper that **execs the working tree**, `nix/programs/alacritty/default.nix:26` | **on `git pull`, no switch** |

And the terminal half drags the scanner with it: `mention-open.py` does
`sys.path.insert(0, <its dir>/collector)`, so a click imports **the repo's**
`scripts/collector/mention_scan.py`, not the store copy. One file, two deploy
lifetimes depending on which consumer reads it — which is why `readlink -f`, and
never a diff, is the arbiter. So the picker, the removed cap, the universe
fallback and the `--print` behaviour go live the moment `main` is pulled on a
host; the ClickHouse rows still need the switch.

## Gate

Run on a TRUE **MERGED** tree — ``7d94c568``, this branch merged with current
`origin/main` `f887e958`. (Round 1 claimed `316ccf74` as "the merged tree"; that
was the branch head and its merge base was `cc409f82`, so that merge was a no-op
and the gate never saw `main`.) The nix derivations were built **one at a time**.

| tier | leg | result |
|---|---|---|
| dev host (`scripts/gate.sh --tier both`) | pytest | PASS — collected 21,642 / passed 21,639 / skipped 3 / failed 0 (floor 18,610) |
| dev host | node | PASS — 1,449 tests, 41 files, 0 fail (floor 1,367) |
| sandbox (`nix build .#checks.x86_64-linux.pytests`) | pytest | PASS — collected 21,642 / passed 21,639 / skipped 3 / failed 0 |
| sandbox (`nix build .#checks.x86_64-linux.nodetests`) | node | PASS — 1,449 tests, 0 fail |

## Test discipline

**Mutation sweep — COMMITTED this time.**
`scripts/tests/mutation_battery_mentions.py` is re-runnable from the tree:

```
python3 scripts/tests/mutation_battery_mentions.py
```

Round 1's sweep lived in `/tmp`, so the auditor could not re-run it and built
their own — which found four guards the first had not imagined. This one spans
**three files**, because the defects it exists for live at the SEAMS between the
scanner, the tailer and the click handler rather than inside any one of them;
`scripts/tests/test_mutation_battery_anchors.py` (which IS collected) gains
per-row `TARGETS` support so a row whose anchor stops occurring exactly once
fails the push instead of silently scoring SURVIVED.

Latest run: **16/16 killed for the stated reason, 0 survived**, positive control
`P1` KILLED, restore verified by hash. Fourteen of the sixteen rows carry an
`expected` phrase and are scored `KILLED(attributed)` — the row's own assertion
went red, not merely "a test failed". Every run sets `PYTHONDONTWRITEBYTECODE=1`.

**One survivor found and closed during this round.** `repo_source: source if repo
else SOURCE_NONE` is observable on a CANDIDATE and **not on a span** —
`scan_mention_spans` takes its `repo_source` from a candidate that HAS a repo, so
it reports `""` for that case with or without the guard. A span-level version of
the test SURVIVED; the assertion moved to `scan_mentions` and the source comment
now states that scope instead of claiming the guard is reachable everywhere.

**Regression matrix, measured against the pre-fix tree rather than asserted:**

| mutant | old suite | new suite |
|---|---|---|
| round-1 `mention_key` (drops the repo) | 106 passed | 3 failed |
| whole mapping → `b64:repo` | 106 passed | 2 failed |
| whole mapping → `context` | 106 passed | 2 failed |
| universe → refusal-path `notify` | 92 passed | 1 failed |
| universe → stdout on the picker path | 1 failed | 1 failed |

The last row is stated because it is the one place the round-1 guard was NOT
blind: a `print()` leak was already caught. It was the `notify` paths — the two
the guard stubbed out — that were invisible.

**Disclosure check:** every token this branch adds that is **new to the tree**
(329) diffed against `gh api user/repos`. Private list asserted **non-empty
first** (233 private / 155 public / 388 total) so a clean verdict is not a scan
wired to nothing; negative control confirms the matcher fires on a planted
private name and full_name. **0 matches against private names, 0 against public
ones.**

https://claude.ai/code/session_016A9pA8WLHsvyeaAcuEsdPi

### DIFF
--- a/scripts/collector/claude/session-tailer.py
+++ b/scripts/collector/claude/session-tailer.py
@@ -877,7 +877,10 @@ MENTIONS_PER_SESSION_CAP = 200
 
 # Cheap pre-filter before the regex pass: a text block containing none of these
 # literals cannot match any pattern this tailer scans for, and does not need
-# scanning. Measured over one 24h window it skipped 81% of assistant text blocks.
+# scanning. Re-measured 2026-09-04 over the preceding 24h (10,118 assistant text
+# blocks): the OLD two literals skipped 81% of them, this derived telemetry set
+# skips 79%. ⚠ That percentage is a property of the WINDOW, not of the code —
+# `mention_scan.mention_hints`' docstring carries the same caveat and the numbers.
 #
 # 🔴 DERIVED FROM THE SCANNER'S OWN LEDGER, NEVER HAND-MAINTAINED. This used to
 # be the literal `("#", "868")`, and that made it the exact place a widening goes
@@ -926,6 +929,17 @@ def load_mention_repos(path: Path | None = None) -> dict:
     attribution, never the telemetry pass. The value SHAPE rule is
     `mention_scan.clean_repo_map`, shared with the hint handler so the two
     cannot disagree about what a usable entry is.
+
+    🔴 WHAT COMES BACK NAMES PRIVATE REPOSITORIES — the same file whose committed
+    ancestor disclosed 232 of them into this PUBLIC repo. `scripts/mention-open.py`
+    carries this warning for its rofi window; this reader is the one that matters
+    MORE, because its consumer is a durable ClickHouse table rather than a window
+    that closes. ONE value from this mapping may leave here per mention: the
+    repository that mention was ATTRIBUTED to. Never the mapping, never a second
+    entry, never a count of it. Pinned by
+    `test_session_tailer.py::test_the_repo_mapping_never_reaches_the_SPOOL_beyond_
+    the_ONE_repo_a_mention_was_attributed_to`, which asserts the attributed name
+    IS present (a positive control) before asserting the others are not.
     """
     p = path or mention_repos_path()
     try:
@@ -935,13 +949,43 @@ def load_mention_repos(path: Path | None = None) -> dict:
 
 
 def mention_key(m: dict) -> str:
-    """The dedupe identity of a mention: platform + the exact matched text.
+    """The dedupe identity of a mention: platform + the exact matched text + the
+    ATTRIBUTED repo.
 
     🔴 The RAW text, not the bare id. `devrc#370` and `talos-infra#370` are
     different references that share an id, and keying on the id alone would emit
     only the first of them and silently drop the second for the life of the
-    session."""
-    return f"{m['platform']}:{m['raw']}"
+    session.
+
+    🔴 AND THE REPO, for exactly the same reason one level up. 92% of mentions
+    are a bare `#N`, so the raw text is `#1291` for every one of them: without
+    the repo in the identity, `trowelcast PR #1291` and `plotwidget PR #1291`
+    collapse to ONE row, and the FIRST occurrence wins — which, because the bare
+    unattributed form usually appears first, systematically discards the
+    attribution this whole change exists to compute. Measured on real
+    transcripts before the fix: 34 rows/day shipped `repo=""` although an
+    attribution was available, and 6 rows/day dropped a second repository's
+    reference outright.
+
+    🔴 THE FORMAT IS VERSIONED BY OMISSION, AND THAT IS A DELIBERATE MIGRATION
+    CHOICE, NOT AN ACCIDENT OF STRING BUILDING. An unattributed mention keys
+    EXACTLY as before — `"<platform>:<raw>"`, byte for byte — and only an
+    attributed one takes the `@<owner>/<repo>` suffix. The alternative
+    (`"<platform>:<repo>:<raw>"` unconditionally) would have re-keyed EVERY
+    entry in every host's `session-summary-state.json`, re-emitting every
+    already-shipped mention once on both hosts. The deployed tailer passes NO
+    mapping at all (`collect_mentions(objects)`), so every key on disk today was
+    written with `repo == ""` and every one of them still matches. What DOES
+    re-emit once is precisely the set that newly gains an attribution — a row
+    whose content genuinely changed, which is the row this PR exists to ship.
+
+    ⚠ `@` IS SAFE AS THE SEPARATOR because no pattern in `mention_scan` can put
+    one in `raw`: `_OWNER`, `_REPO` and `_URL_REPO` are all alphanumerics plus
+    `-`, `_` and (for URLs) `.`, and the wordy forms are literal keywords. So a
+    key cannot be spelled two ways.
+    """
+    repo = m.get("repo") or ""
+    return f"{m['platform']}:{m['raw']}@{repo}" if repo else f"{m['platform']}:{m['raw']}"
 
 
 def collect_mentions(objects: list[dict], *, repos: dict | None = None) -> list[dict]:
@@ -1022,6 +1066,14 @@ def build_mention_emit_args(ev: dict, m: dict) -> list[str]:
     `repo_source` rides beside it so a consumer can tell a repo the text stated
     from one resolved through the operator's mapping — an analysis that cannot
     separate those is one measurement pretending to be two.
+
+    🔴 THAT SEPARATION IS `explicit` vs `mapped`, and it did not exist until it
+    was measured. `explicit` means an `owner/repo#N` the operator wrote out;
+    `mapped` means a bare `repo#N` whose OWNER came from
+    `~/.config/mention-open/known_repos.json`. Both used to report `explicit`,
+    so the field promised a distinction it did not carry. `adjacent`, `url`,
+    `flag` and `default` are the block-level routes — see the ATTRIBUTION block
+    in `mention_scan`'s module docstring, which owns the full ladder.
     """
     args = [
         "source=mentions",
--- a/scripts/collector/claude/session-tailer.py
+++ b/scripts/collector/claude/session-tailer.py
@@ -875,11 +875,63 @@ def emit_event(emit: str, ev: dict) -> None:
 # state file grow without limit, not that 200 is a meaningful threshold.
 MENTIONS_PER_SESSION_CAP = 200
 
-# Cheap pre-filter before the regex pass. Every pattern in mention_scan requires
-# a literal '#' or the literal '868', so a text block containing neither cannot
-# possibly match and does not need scanning. This is the same short-circuit the
-# rejected hook design needed, applied where it is nearly free.
-_MENTION_HINTS = ("#", "868")
+# Cheap pre-filter before the regex pass: a text block containing none of these
+# literals cannot match any pattern this tailer scans for, and does not need
+# scanning. Measured over one 24h window it skipped 81% of assistant text blocks.
+#
+# 🔴 DERIVED FROM THE SCANNER'S OWN LEDGER, NEVER HAND-MAINTAINED. This used to
+# be the literal `("#", "868")`, and that made it the exact place a widening goes
+# to die: every telemetry-only shape the scanner learned — `audit-pr 1291`,
+# `gh pr view 1291`, `clawgate task 370` — contains neither literal, so the
+# regexes would have been reachable in a unit test calling `scan_mentions()`
+# directly and DEAD in production, with the whole suite green. Deriving it means
+# a pattern added to `PATTERN_LEDGER` widens the filter in the same commit,
+# because there is nowhere else to put the fact.
+#
+# 🔴 AND IT IS THE TELEMETRY PROFILE, not the default. `mention_hints()` with no
+# argument returns the TERMINAL profile's two literals — the old value — which
+# would look correct, pass a "derived from the ledger" review, and still skip
+# every new shape.
+_MENTION_HINTS = MS.mention_hints(MS.PROFILE_TELEMETRY)
+
+# Where the operator's generated repo mapping lives — the same per-host file
+# `scripts/mention-open.py` reads, and for the same reason: it names PRIVATE
+# repositories, so it lives OUTSIDE every checkout and this repo is public.
+# Nothing breaks when it is absent; a `#N` simply stays unattributed.
+#
+# 🔴 THIS EXPRESSION IS THE THIRD COPY OF ONE PATH — the generator writes it, the
+# hint handler reads it, and now so does this tailer. They agree only by
+# coincidence, so `scripts/tests/test_mention_open.py` imports all three modules
+# and pins the three answers equal. Move one and the mapping is generated into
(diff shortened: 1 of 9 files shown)


ZacxDev:
## 🔴 Correction: the "Deploy note" in the body above is WRONG, in the direction that misleads verification

The body says both runtime files are `home.file` targets, so "merging changes
nothing until `scripts/ship.sh` runs a `home-manager switch` on both hosts".
**That is true of the telemetry half only. The operator-visible half goes live on
`git pull` alone.** Leaving that uncorrected would send a reviewer to verify the
picker against an artifact they believe is frozen, and it points the wrong way:
it says "you have time", when in fact the click surface changes the moment `main`
is pulled on a host.

Measured on the workbench, `readlink -f` as the arbiter — not a diff:

```
$ readlink -f ~/.config/activity-collector/mention_scan.py
/nix/store/arkxlclfjr0kms7yvhxbgnc1icp2xxk2-hm_mention_scan.py          # store copy
$ readlink -f ~/.config/activity-collector/claude/session-tailer.py
/nix/store/rwvr9apmkxrs30c0xdxrd9043lh45y7b-hm_claude/session-tailer.py # store copy

$ cat /nix/store/lp1hjldfay4mcvf638d1j8gvbj3rz0rl-alacritty-mention-open
  …
  exec …/bin/python3 \
    /home/zach/workspace/devrc/scripts/mention-open.py "$@"             # LIVE TREE
```

So:

| half | deployed as | when it changes |
|---|---|---|
| telemetry — `mention_scan.py`, `session-tailer.py` under `~/.config/activity-collector/` | `home.file` store copies | only on `home-manager switch` |
| terminal — the Alacritty hint | a `writeShellScript` wrapper that **execs the working tree** (`nix/programs/alacritty/default.nix:26`) | on `git pull`, no switch |

And the terminal half drags the scanner with it: `mention-open.py` does
`sys.path.insert(0, <its own dir>/collector)`, so a click imports **the repo's**
`scripts/collector/mention_scan.py`, not the store copy. One file, two deploy
lifetimes, depending on which consumer reads it — which is exactly why
`readlink -f` and not a diff is the arbiter.

**Consequence for review:** the picker, the removed namesake cap, the universe
fallback and the `--print` behaviour are live on a host the moment `main` is
pulled there. The ClickHouse rows — the new shapes, the attribution, the dedupe
key — still need the switch.

The body has been corrected too, but this comment is the record: anyone who read
the earlier version read a false claim about half the change.

_Posted by the round-1 fix pass; the four code findings are in the commit above._

ZacxDev:

```audit-claims round=1 audited=316ccf74..7d94c568
1. mention_key now carries the attributed repo as a conditional `@owner/repo` suffix; claims no already-emitted mention re-keys.
2. A new SOURCE_MAPPED value distinguishes a repo resolved through the operator's mapping from an owner written out in the text; source is decided by whether the text wrote an owner.
3. All four profile-split enforcement sites are pinned behaviourally, with the click-surface pin driven off PATTERN_LEDGER.
4. The mention-open disclosure guard keeps the real notify and stubs subprocess.run; session-tailer gains a spool-level guard over raw bytes, decoded events and stdout/stderr.
5. The PR body deploy note is corrected, and a correction comment is posted separately.
6. `--print` above the retired namesake cap prints every namesake and exits 0 — declared intended and pinned at nine rows.
7. A re-runnable mutation battery is committed at scripts/tests/mutation_battery_mentions.py, reported 17/17 killed with a positive control.
8. The scripts/collector/claude/tests floor is re-pinned from 224 to 219.
9. The prefilter percentage is restated at both sites with counts and the window it was measured over.
```

PR #1324, opened 2026-09-05
feat(clawgate): host-side terminal WRITE agent — outbound short poll, SHIPPED DISABLED

ZacxDev:
The **host half** of rank 29 of the tmux-webapp effort. The server half is `ZacxDev/homelab-infra#711`.

## What this is

clawgate's pod cannot see either host's tmux socket — no hostPath, no hostNetwork, no hostPID, no nodeName. The read model solved that by pushing outbound (`tmux-snapshot-push.sh`); this is the **return direction**, solved the same way rather than by opening a port. `scripts/tmux-reply-agent` holds an outbound short poll (~5s), claims queued writes for **its own host**, runs `tmux send-keys` locally, and reports the outcome. No inbound access to either machine, and no credential for either machine inside a pod on an unauthenticated LAN surface.

That closes the third missing layer. Host→pod was push-only; **nothing pulled commands down**, which is why "jump into this pane" is currently a clipboard copy of `tmux switch-client -t %N`.

## 🔴 IT SHIPS DISABLED. NO SECRET IS PROVISIONED.

`enableTmuxReplyAgent = false` in `nix/home.nix`. The **service unit is emitted on both hosts and wired into nothing** — `Install.WantedBy` is `lib.optionals enableTmuxReplyAgent [ "default.target" ]`, so a `home-manager switch` after this merges starts no process.

Arming needs **two independent acts**, and neither implies the other:

1. `CLAWGATE_TERMINAL_TOKEN` provisioned into the **pod's** secret — until then the server's routes answer 503 and the pod says so at boot, unconditionally and in both directions;
2. this flag set true, with the same secret readable on the host from `~/.claude/clawgate.env`.

Flip only the flag and you get a unit that exits **2** saying the surface is not armed. Provision only the secret and you get a live route nobody polls. Disarming is removing the secret; it is reversible.

Both halves are pinned: `test_the_agent_unit_SHIPS_DISABLED` reads the flag through `testlib.nix_units.strip_nix_comments`, **not** a substring search — this file's blocks quote directives verbatim in prose, and that exact failure once left a guard green over a deleted unit.

## What one delivery actually does, and what guards it

🔴 **`tmux send-keys` into a named pane, then Enter. That is arbitrary command execution as the operator on this host**, on text that arrived over a LAN route with no human auth. Your decisions — free text into any pane, and clawgate presses Enter — are what make it that; everything below exists because of them, not instead of them.

- 🔴 **The text never reaches a shell.** It is passed to `subprocess.run` as one element of an **argv list**. `test_the_text_is_delivered_as_one_argv_element_with_no_shell` feeds `-n $(id) \`id\` "q" 'p' ;echo x |tee /dev/null && true \n $HOME` and asserts the byte-identical string arrives as a single argument — through a stub `tmux` that records **argv**, not a command line, because a stub recording `"$*"` would flatten the exact distinction under test.
- 🔴 **`-l` and `--` are the command, not decoration.** Without `-l`, `send-keys` reads its argument as *key names*: a reply containing `C-c` interrupts whatever is running instead of typing three characters. Without `--`, text starting with `-` is a flag. The test pins the exact argv, because presence without ordering proves nothing.
- **One submission, never two.** The server rejects any text containing a control character, so the literal send plus a separate `Enter` is the only submission a delivery can produce. (A newline would otherwise submit line 1 and *run line 2 as a fresh command*.)
- **The write-time guard runs before the write.** `expectTmuxServerId` — the collector's own `<pid>:<start_time>` token — is compared against this host's live tmux server identity. A mismatch is `refused` **with no `send-keys` at all**; the test asserts zero sends, not merely a refusal status, because a guard that reports after the keys have gone is not a guard.
- **An unmeasurable identity refuses.** tmux renders an unknown format verbatim, so an old tmux yields a bare `:`; that reads as "cannot measure", which the guard turns into a refusal.
- **An unknown `kind` refuses rather than guesses.** Falling back to the default would run a command nobody on this host reviewed.
- **`new-session` addresses the new pane by id.** `new-window -P -F '#{pane_id}'` prints it, and the text is sent to *that* id — between the two calls the operator may have switched windows, and a relative target would type into whatever they moved to. If tmux reports no pane id, the agent **fails rather than guessing**, with zero sends.

## Idempotency: at-most-once, and the agent's part in it

A retried `send-keys` runs a command twice. The server's claim moves a row out of `pending` permanently; **this agent never retries a send, never re-claims, and never replays**. When the result POST fails it does *not* re-execute — the row is relabelled `abandoned`, which means **delivery UNKNOWN, never "did not run"**, and the agent logs it in those words. `test_a_failed_result_post_does_NOT_re_execute_the_write` pins both halves.

## The reply text is credential-grade and is never logged here

You chose a full audit log **including the text**; the cost stated at the time was that a reply may contain a secret. That log lives in the server's table behind the terminal token. The journal is a much wider readership, so:

- `test_the_reply_text_is_never_written_to_the_journal` drives a delivery that **fails with tmux's stderr echoing the text back** and asserts a synthetic marker appears nowhere in the agent's output — with positive controls that the agent logged *something*, named the write, and really did send the marker.
- the agent **redacts the exact text** from any diagnostic before reporting it (`detail` lands in the audit log *and* a log line, and tmux quotes its arguments in its own errors). It is an exact substitution, not a secret-shaped pattern, so it cannot miss the way a heuristic misses.
- 🔴 The same hazard bit the server side for real: its first draft logged the agent's `detail` and the equivalent guard caught it. See #711.
- **This repo is public.** Never paste a real reply, a real pane preview, or a row from that table anywhere in it.

## The seams a disagreement would break silently

- **Host label.** The queue is keyed on the label the *read model* stores, which the collector computes from `ACTIVITY_HOST` (`hostname` is `nixos` on both machines). An agent computing a different label would poll for a host nobody enqueues to and deliver nothing, with no error anywhere. `test_the_host_label_follows_the_same_rule_as_the_collector` asserts the rule **and** reads `HOST_NAMES` / `DEFAULT_LOCAL_HOST` out of `scripts/session-manager`'s own source rather than restating them.
- **tmux server-id spelling.** Two spellings of one token would compare apples to oranges and **refuse every write** — which reads exactly like "the feature does not work", with nothing naming the cause. `test_the_server_id_format_matches_the_collectors` pins `TMUX_SERVER_ID_FORMAT` against the collector's `WINDOW_FORMAT` and its `f"{pid}:{started}"` join, read from that file.
- **Env beats file.** `clawgate-stop-hook.sh` sources the same credential file with `set -a`, which makes the *file* win — and the measured consequence over there was a probe aimed somewhere harmless silently POSTing to production. On this surface that mistake sends a reply into a live pane. Regression-guarded, along with the `export `-and-leading-whitespace anchor that `tmux-snapshot-push.sh` had to learn.
- **The token is never in argv or a URL** — asserted through the stub tmux's recorded argv and the stub server's received request, not by reading the source.

## The unit

`Type = "simple"`, not the oneshot-plus-timer shape its two siblings use: a ~5s poll driven by a timer would fork a process, a Python interpreter and a TCP connection **17,280 times a day** for a queue that is empty almost always.

- 🔴 **No `OnFailure`**, deliberately. `notify-failure@` toasts are wired to *defeat* do-not-disturb, justified by a measured ~1 firing in 9 days. This unit restarts on failure and polls constantly, so any sustained condition — above all "the surface is not armed", the state it ships in — would fire a DND-bypassing toast on every restart and burn down the one alert channel that has to keep its meaning. What surfaces a genuinely broken agent instead is `StartLimitBurst=5` / `StartLimitIntervalSec=600` marking the unit **failed**, which `/standup` reads. Everything transient is backed off from *inside* the loop rather than exited on, precisely so a restart storm is not the failure mode.
- **A repeating condition is logged once, not per tick** — asserted (`out.count("NOT ARMED") == 1`), because at the real cadence a per-tick line is 17,280 a day and buries every real event.
- **PATH carries `coreutils`, `python3`, `tmux` and nothing else.** Deliberately *not* curl/openssh/gawk: the agent is stdlib Python and never leaves this host. That is the opposite trim from `tmux-snapshot-push`'s list, and copying it wholesale would be carrying binaries to justify later. Pinned two ways (present *and* absent).
- **`TMUX_TMPDIR=%t`** — this host's socket is at `$XDG_RUNTIME_DIR/tmux-UID/`, not tmux's compiled-in `/tmp/tmux-UID/`. Here it is load-bearing rather than defensive: if this unit cannot connect, every delivery fails *loudly* and is recorded as `failed`, which is the safe direction — unlike the feeder, where the same miss produces a silent destructive zero.
- **NOT `serverMode`-gated**, and the asymmetry with `enableTmuxSnapshotPush` is the same one `transcript-push` has. That feeder is workbench-only as a *correctness* requirement (its collector already reaches both hosts over ssh, so a second reporter would fight over every row). A tmux socket is visible only from the machine it lives on, so a workbench-only agent would leave every laptop pane permanently unanswerable. The queue is keyed on the host label, so two agents claiming disjoint scopes cannot collide.

Exit codes are distinct on purpose: **2** = no usable credentials (a configuration error that cannot self-heal — and the state this ships in), **3** = tmux unusable, **0** = clean shutdown. An unreachable server, a 503, a 401 or a malformed body are **backed off from, not exited on**.

## 🔴 A gap I found reviewing my own agent: it trusted the server's payload

The agent validated nothing the server had already validated. That reads like good hygiene — "one rule, one place" — and it is **the wrong reading of that rule**: it governs two copies on ONE side of a trust boundary. Here the server is a pod reachable from an unauthenticated LAN NodePort and this process is the thing that *runs commands as the operator*. The two sides fail independently, and the side that executes is the one that must refuse.

Re-checked on the host now, each of which the server also rejects:

- **a pane that is not `%N`** — a name-shaped target does not *fail* in tmux, it resolves fuzzily and runs the command in some other pane;
- **a control character in the text** — one write becoming two submissions;
- **a relative or `..`-bearing `cwd`** — resolves against *this process's* directory, which differs per host and after every restart, on the one path that creates its own target;
- **an oversized claim batch** — the server caps a claim at 4; a thousand writes from a server that is wrong, compromised or simply newer is a thousand commands, and nothing downstream would stop them. The batch is refused **whole**, not truncated: executing four and dropping the rest is a partial delivery nobody can reason about, whereas refusing leaves every row to be relabelled `abandoned` (delivery UNKNOWN), which is the honest record.

It also bounds the response body it will read — a bare `.read()` takes whatever the peer sends into a process on the operator's workstation.

**Mutation-tested:** deleting the `validate()` call reds **12** cases; deleting the batch cap reds its own, with a control that a batch *at* the cap still executes.

## 🔴 The text gate was a DENYLIST, at the one site reachable from the network

```python
def has_control(value):
    return any(ord(ch) < 0x20 or ord(ch) == 0x7F for ch in value)
```

It refuses NUL, `\n`, `\r`, ESC and DEL — and **passes** every C1 control (U+0080–U+009F), **U+0085 NEL** (a line break to some terminals), U+2028/U+2029, and every invisible. Including **U+000F**, the exact character `scripts/session-write`'s own docstring records as having got through *its* earlier denylist and executed a pane's buffer **with no Enter sent**.

So this repo had already found and fixed this class, written down why, and mutation-tested the fix — and I shipped a second, weaker copy at the site an attacker can reach. That is `claude/RULES.md`'s "a predicate open-coded at N sites is typically wrong at N−1 of them, in the same direction", measured.

**The rule now lives in `scripts/lib/tmux_text_policy.py` and both programs import it.** `session-write` re-exports the names its 185 tests already read, so none of them changed, and it pins the tmux argv separator against `session-resolve`'s own constant at import — an assertion, not a convention. Neither program falls back if the module is missing: a fallback is exactly how the weaker check would return.

Guarded two ways, because either alone decays:

- **structural** — both predicates' `__code__.co_filename` must be the shared file, and an AST scan (not a grep — this file's prose names `has_control` repeatedly) fails if either script defines its own again;
- **behavioural** — a sweep across U+0020–U+20FF asserts the agent's gate answers exactly what the shared allowlist answers, with a positive control that the sweep saw both answers.

The allowlist is `ch in ("\t",) or ch.isprintable()`. Tab is the one control character deliberately permitted: it has an ordinary typographic use and cannot accept a line.

## 🔴 Three guards that pinned words and passed over live mutants

The audit found these; all three now fail on the mutant that walked through them.

| guard | the mutant that survived | why it mattered |
|---|---|---|
| `WantedBy` | `lib.optionals (!enableTmuxReplyAgent) [ "default.target" ]` | both asserted substrings survive an **inversion** that starts the agent on both hosts on the next switch — the single thing this change claims cannot happen |
| unit `PATH` | deleting `pkgs.python3` from `makeBinPath` | the assertion was satisfied by the **ExecStart** line, which names python3 for a different reason; the child lost its interpreter directory |
| new-session `submit` | replacing `submit = True` with `pass` | neither direction was pinned, because every fixture set submit true |

The `WantedBy` guard now extracts the condition and pins it as a **whole normalised expression** — `!enableTmuxReplyAgent` is not `enableTmuxReplyAgent` — and asserts the flag is `false`. It is deterministic, so it runs identically in the check sandbox, which has no recursive nix; a `nix eval` sits beside it as **labelled confirmation** that skips where nix is unavailable, because a skipped test is not a passing one.

## 🔴 Two defects a stubbed suite structurally could not see

Both were in shipped payload, both on the path that **always presses Enter**, and both were invisible to 86 green tests because **the stub tmux exits 0 for everything** — it models neither resolution rule. Only a real-tmux probe finds them.

**1. tmux PREFIX-MATCHES a session name.** `-t "<name>:"` does not fail when the session is absent — measured on tmux 3.7c, it returns rc 0 and picks a longer match. The operator's live server is exactly the shape that triggers it: `scratch`, `scratch2` … `scratch20`, and `datapacket-talos`, `datapacket-talos-2`. So `-t scratch2:` — *the literal this suite's own fixtures used* — opened a window in `scratch20`, typed the command, pressed Enter and reported `delivered`. My docstring asserted the opposite. Fixed with `=`, which forces an exact match.

**2. `-c <absolute path that does not exist>` also returns rc 0**, with the window in `$HOME`. Three shape-valid inputs reach it: a path absent **on this host** (the queue is host-keyed, so a workbench path can be enqueued for the laptop), a trailing space, and a trailing `;` — which tmux eats off the `-c` word, and which the *text* path already refused while the *cwd* path did not. Same character, same file, opposite treatment. A build or a recursive delete ran in the home directory and was reported `delivered`.

Both are now checked twice: `os.path.isdir` up front for a legible refusal, and a **read-back** of the session and working directory tmux actually chose, printed by the same `-P -F` at no extra round trip. The read-back is what catches a resolution rule nobody has thought of.

⚠ **Both read-backs SURVIVED their first mutation** — `=` and `isdir` each fire first in the ordinary case — so the stub can now *lie* about where the window landed, and two tests pin them independently.

`tmux` is in `REQUIRED_TOOLS` and `flake.nix` `gateTools` so those guards cannot silently skip in the tier that gates. ⚠ Adding it without a written justification is itself what the sandbox tier caught: `test_devshell_satisfies_required_tools.py` requires one, because the FATAL it prints promises the reader one.

## Verification matrix

**Base sha `a75ffc6a`** (`origin/main`; the branch is rebased onto it, so this is the merged tree).

**Tier 1 — `scripts/gate.sh --tier both`, run inside the flake devShell** (`nix develop <worktree> -c bash scripts/gate.sh --tier both`; a bare run fails rc 3 with `required tool(s) missing from PATH: logrotate`, because `.envrc` is `use opencode` and does not carry the gate toolchain):

Both tiers PASS, and the counts are the runners' own summary lines rather than my arithmetic:

```
  TOTAL collected=21779  passed=21776  skipped=3  failed=0  (floor: 18610 = sum of 30 per-target floors)
RESULT: PASS (exit=0)
RESULT: PASS (exit=0)
GATE: RESULT=PASS exit=0
```

`gate.sh` cross-checks its own exit status against those `RESULT:` lines and exits 90 when they disagree; it exited 0, so the status and the verdicts agree.

**Tier 2 — the sandbox tier the merge gate runs**, `nix build .#checks.x86_64-linux.pytests` and `.nodetests`, **one at a time** (a combined invocation produces false failures under store contention):

```
devrc-pytests>   TOTAL collected=21779  passed=21776  skipped=3  failed=0  (floor: 18610 = sum of 30 per-target floors)

```

Exit codes as the runs themselves reported them — not a piped status, which `| tail` destroys:

```
GATE_RC=0
NIXPY_RC=0
NIXNODE_RC=0
```

⚠ **Only the final run counts, and the earlier ones are catalogued above.** Each superseded run measured a tree that has since moved; none of their numbers appear here.

**This suite alone:** `scripts/tests/test_tmux_reply_agent.py` — **105 passed**, 0 failed, 0 skipped. `scripts/tests/test_session_write.py` — **185 passed**, unchanged by the predicate extraction.

⚠ **An earlier version of this body claimed 46 and described code that was not in the pushed commit.** The push had never landed: I pushed once before rebasing and never again, so the PR showed the pre-rebase head while the body was written from my worktree. `git status` was clean, which is what I checked, and it cannot see that. **Checking `git rev-list --left-right origin/<branch>...HEAD` is what would have.** The head is **`1adc0d47`**, and I verify the PUSHED head's own file list against every claim here — `git show origin/<branch>:<path>` — rather than the worktree's. The same failure recurred one layer out later in this PR: an edit to this body's *file* that was never `gh pr edit`-ed, so the published body lacked a section I had reported as present. I now re-read the live body with `gh pr view --json body` after every edit.

## Red-at-base, honestly labelled

🔴 **Nothing in this PR is regression coverage, and I am not presenting it as such.** Every test here covers a script that did not exist at `origin/main`; there is no pre-change code to watch them fail on, and a compile/import error is not a red I would report as one.

The one guard that *did* earn its keep by going red is on the server side (#711): the equivalent text-never-logged test caught a real credential leak in that handler's first draft.

Two tests here are **invariant guards** rather than coverage, and are labelled so in their own docstrings: `test_the_server_id_format_matches_the_collectors` and `test_the_host_label_follows_the_same_rule_as_the_collector` pin cross-file relationships no bug has yet violated.

## ⚠ A property of the transport, stated so it is not later inferred to have been considered

**This agent talks to the pod over plain HTTP to a LAN NodePort.** No TLS, no server authentication, and the terminal token travels in cleartext on every poll (~every 5s). That is the pattern `tmux-snapshot-push.sh` and `transcript-push.sh` already use — **but the stake is different**: those tokens gate *reads*, this one gates `send-keys` plus Enter. Not fixed here, and not in scope; written down so arming is a decision taken with it on the table.

## What I could NOT verify

- 🔴 **The agent has never run against a real clawgate, and structurally cannot until the surface is armed.** Every test drives it against a stub HTTP server and a stub `tmux`. It has never touched a real pane.
- 🔴 **The closing condition in the handoff — "a reply typed in the web UI appears in the target pane" — is NOT met by this PR and is not claimed to be.** It needs the token provisioned, rank 30's UI, and a live check on both hosts.
- **No `home-manager switch` was run.** The unit is not wired into anything, so a switch would start nothing, but I did not run one and cannot say the module evaluates on a real host beyond `nix-instantiate --parse` and whatever the nix check tier exercises.
- The unit's `Environment` PATH list is asserted against the source, **not** by running the child under it — the one mistake `tmux-snapshot-push`'s comment warns about (its `gawk` was trimmed as "unused" and its absence was silent). Here the exposure is smaller: the agent is stdlib Python plus `tmux`, and a missing `tmux` exits 3 loudly rather than fabricating a zero. Still: unmeasured.
- Branch protection on this repo was not re-measured by me.

## Not done, deliberately

- **No `CLAWGATE_TERMINAL_TOKEN` created, written to `~/.claude/clawgate.env`, encrypted or applied.**
- Master switch left `false`. No timer, no `default.target` want.
- No `ship.sh`, no deploy, no merge.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -119,6 +119,14 @@
         gatePyEnv pkgs.bash pkgs.ripgrep pkgs.git pkgs.util-linux pkgs.jq
         pkgs.gnugrep pkgs.curl pkgs.nodejs pkgs.nix pkgs.opencode pkgs.logrotate
         pkgs.rsync pkgs.zsh pkgs.age
+        # 🔴 tmux, because two of tmux-reply-agent's guards CANNOT be written
+        # against a stub. A stub tmux always exits 0, so it models neither the
+        # session PREFIX-MATCH (`-t scratch2:` opening a window in `scratch20`)
+        # nor the `-c <missing path>` fallback to the home directory -- both
+        # measured live, both shipped, both invisible to a stubbed suite. Those
+        # tests drive a REAL tmux on a private `-L` socket, and this entry is what
+        # stops them SKIPPING in the check sandbox, where a skip is an error.
+        pkgs.tmux
       ];
     in
     {
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -137,6 +137,37 @@ let
   # rejected outright (the server 4xx's and CHANGES NOTHING, leaving the previous tail in
   # place), and the server sweeps rows older than 7 days.
   enableTranscriptPush = true;
+  # clawgate TERMINAL WRITE agent master switch (scripts/tmux-reply-agent) — the
+  # host half of the reply channel. Gates ONLY whether the long-running SERVICE is
+  # wired into `default.target`; the unit definition is always emitted, so it can
+  # be started by hand regardless.
+  #
+  # 🔴 IT SHIPS **false**, AND THAT IS THE POINT OF THE WHOLE SHAPE — not caution
+  # about an unfinished feature. What this agent delivers is `tmux send-keys`
+  # followed by Enter, i.e. arbitrary command execution as the operator on this
+  # host, driven by a route on a LAN NodePort that has no human auth. The operator
+  # took that decision deliberately, with the blast radius stated; what was ALSO
+  # decided is that building it and ARMING it are separate acts, so the write path
+  # can be merged, deployed, read and audited before it can execute anything.
+  #
+  # 🔴 TWO INDEPENDENT SWITCHES, AND NEITHER IMPLIES THE OTHER. Arming needs
+  # (1) CLAWGATE_TERMINAL_TOKEN provisioned into the POD's secret — until then the
+  # server's write routes answer 503 and the pod says so at boot, unconditionally
+  # and in both directions ("terminal write surface: DISABLED (fail-closed)") —
+  # and (2) this flag true, with the same secret readable on this host from
+  # ~/.claude/clawgate.env. Flipping only this one gets a unit that exits 2 saying
+  # the surface is not armed; provisioning only the secret gets a server with a
+  # live route and nobody polling it. Disarming is removing the secret, which is
+  # reversible and takes effect on the pod's next boot.
+  #
+  # 🔴 NOT serverMode-GATED, and the asymmetry with enableTmuxSnapshotPush is the
+  # same one transcript-push has. That feeder is workbench-only as a CORRECTNESS
+  # requirement (its collector already reaches both hosts over ssh, so a second
+  # reporter would fight over every row). A tmux socket, by contrast, is visible
+  # ONLY from the machine it lives on, so a workbench-only agent would leave every
+  # laptop pane permanently unanswerable. The queue is keyed on the host label, so
+  # two agents claiming disjoint host scopes cannot collide.
+  enableTmuxReplyAgent = false;
   # Graphical host = runs X/i3 (both current NixOS hosts do; only a genuinely headless
   # box would not). Approximated as isNixOS, mirroring graphical.nix — deliberately NOT
   # !serverMode, which is true on the graphical workbench.
@@ -3759,6 +3790,87 @@ in
     Install = {
       # 🔴 NO serverMode GATE — both hosts, on purpose. See enableTranscriptPush.
       WantedBy = lib.optionals enableTranscriptPush [ "timers.target" ];
+    };
+  };
+
+  # ── CLAWGATE TERMINAL WRITE AGENT (scripts/tmux-reply-agent) ────────────────
+  # The host half of clawgate's reply channel: a long-running OUTBOUND short poll
+  # that claims queued writes for this host, runs `tmux send-keys` locally, and
+  # reports the outcome. It is the return direction the two feeders above never
+  # had — and it is outbound for the same reason they are: the pod cannot see
+  # either host's tmux socket, and the alternative (an inbound port, or an ssh
+  # credential inside a pod on an unauthenticated LAN surface) is strictly worse.
+  #
+  # 🔴 SHIPPED DISABLED. `enableTmuxReplyAgent` is false — see its comment above
+  # for the two independent switches that arm this and why building it and arming
+  # it were deliberately separated. The unit below EXISTS on both hosts and is
+  # wired into nothing.
+  #
+  # 🔴 Type = "simple", NOT the oneshot-plus-timer shape its two siblings use, and
+  # the difference is the cadence. A ~5s poll cannot be a timer: the user manager
+  # would fork a process, a Python interpreter and a TCP connection 17,280 times a
+  # day for a queue that is empty almost always. A resident process holding one
+  # loop is the cheaper and simpler thing, and it is what lets a reply land within
+  # seconds of being typed — which is the entire point of the feature.
+  systemd.user.services.tmux-reply-agent = {
+    Unit = {
+      Description = "Deliver clawgate's queued terminal writes into this host's tmux panes";
+      After = [ "network-online.target" ];
+      Wants = [ "network-online.target" ];
+      # 🔴 NO OnFailure, DELIBERATELY, and for the reason spelled out at
+      # tmux-snapshot-push: notify-failure@ toasts are wired to DEFEAT
+      # do-not-disturb, and that bypass is justified by a MEASURED rate of about
+      # one firing in nine days. This unit polls every few seconds, so any
+      # sustained condition — the surface not armed (which is the state it ships
+      # in), clawgate mid-redeploy, the laptop's network down — would fire a
+      # DND-bypassing toast on every restart and burn down the one alert channel
+      # that has to keep its meaning.
+      #
+      # What surfaces a genuinely broken agent instead: the two exit codes it can
+      # actually take (2 = no credentials, 3 = tmux unusable) are both permanent
+      # configuration faults, and StartLimit below turns a repeat of either into a
+      # FAILED unit, which `/standup` reads. Everything transient is backed off
+      # from inside the loop rather than exited on, precisely so a restart storm
+      # is not the failure mode.
+      StartLimitIntervalSec = 600;
+      StartLimitBurst = 5;
+    };
+    Service = {
+      Type = "simple";
+      Restart = "on-failure";
+      # Long enough that five restarts cannot happen inside the 600s window above
+      # unless the fault is permanent — which is what should mark the unit failed.
+      RestartSec = 60;
+      Environment = [
+        # 🔴 EVERY ENTRY IS FOR THE CHILD — under systemd there is no login-shell
+        # PATH to fall back on, and drift-check.service already paid for that
+        # lesson (it reported COULD NOT MEASURE on every run forever, from a unit
+        # that looked completely correct).
+        #
+        # The agent is pure stdlib Python plus `tmux`. It needs NEITHER curl (it
+        # uses urllib) NOR openssh/gawk (it never leaves this host) — copying
+        # tmux-snapshot-push's list would be carrying binaries to justify later.
+        # Pinned by `test_tmux_reply_agent.py::test_the_unit_PATH_carries_tmux_and_python`.
+        "PATH=${lib.makeBinPath [ pkgs.coreutils pkgs.python3 pkgs.tmux ]}"
+        "HOME=%h"
+        # 🔴 THE SOCKET DIRECTORY, AND HERE IT IS LOAD-BEARING RATHER THAN
+        # DEFENSIVE. This host's tmux socket is at $XDG_RUNTIME_DIR/tmux-UID/,
+        # not tmux's compiled-in /tmp/tmux-UID/. tmux-snapshot-push sets this
+        # defensively — if IT cannot connect, a host is reported with an empty
+        # window list, which is silent and destructive. If THIS unit cannot
+        # connect, every delivery fails loudly and is recorded as `failed` in the
+        # audit log, which is the safe direction; the pin is here so the ordinary
+        # case works at all rather than to prevent a silent one. `%t` is the user
+        # runtime dir (/run/user/UID).
+        "TMUX_TMPDIR=%t"
+      ];
+      ExecStart = "${pkgs.python3}/bin/python3 %h/workspace/devrc/scripts/tmux-reply-agent";
+      X-Restart-Triggers = [ "${../scripts/tmux-reply-agent}" ];
+    };
+    Install = {
+      # 🔴 SHIPPED DISABLED — see enableTmuxReplyAgent. A `default.target` want,
+      # not `timers.target`: this is a resident service, not a timer.
+      WantedBy = lib.optionals enableTmuxReplyAgent [ "default.target" ];
     };
   };
 
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -3865,7 +3865,15 @@ in
         "TMUX_TMPDIR=%t"
       ];
       ExecStart = "${pkgs.python3}/bin/python3 %h/workspace/devrc/scripts/tmux-reply-agent";
-      X-Restart-Triggers = [ "${../scripts/tmux-reply-agent}" ];
+      # 🔴 THE POLICY MODULE IS A TRIGGER TOO. This is a RESIDENT service, not a
+      # timer: it imports scripts/lib/tmux_text_policy.py once at startup and then
+      # runs for weeks. Without this line, TIGHTENING the text allowlist would
+      # leave the running agent on the OLD predicate indefinitely -- a security
+      # change that appears deployed and is not, on the process that executes.
+      X-Restart-Triggers = [
+        "${../scripts/tmux-reply-agent}"
+        "${../scripts/lib/tmux_text_policy.py}"
+      ];
     };
     Install = {
       # 🔴 SHIPPED DISABLED — see enableTmuxReplyAgent. A `default.target` want,
--- a/scripts/lib/tmux_text_policy.py
+++ b/scripts/lib/tmux_text_policy.py
@@ -1,0 +1,100 @@
+"""THE policy for what may be delivered to a tmux pane as literal text.
+
+🔴 ONE PREDICATE, ONE PLACE — AND THIS MODULE EXISTS BECAUSE THE SECOND COPY WAS
+ALREADY WRONG. `session-write` reached this rule the hard way and wrote the
+reason down; `tmux-reply-agent` then shipped its own version at the
+NETWORK-FACING site, and that version was a DENYLIST:
+
+    def has_control(value):
+        return any(ord(ch) < 0x20 or ord(ch) == 0x7F for ch in value)
+
+It refuses NUL, `\\n`, `\\r`, ESC and DEL — and passes every C1 control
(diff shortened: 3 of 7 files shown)


ZacxDev:
Round-1 audit addressed at `3fa3096ab`. The round-1 audit measured `d19a613a9`, which predated the validation commit — the branch had been rebased locally and never pushed. Claims are WHAT was claimed, trimmed of reasoning.

```audit-claims round=1 audited=d19a613a9..3fa3096a
1. Finding 1 (blocking): claimed fixed — the validation layer the PR body described is now actually pushed; the head's own file list was verified against every claim.
2. Finding 2 (blocking): claimed fixed — a pane that is absent or not of the form %N is refused rather than delivered into whatever pane tmux picks.
3. Finding 3 (blocking): claimed fixed — text is screened by an allowlist rather than a C0+DEL denylist, so C1 controls and U+0085 no longer pass.
4. Finding 4 (blocking): claimed fixed — the ships-disabled guard pins the relationship rather than the words, so an inverted WantedBy no longer survives.
5. Finding 5: claimed fixed — duplicate delivery within a claim batch is bounded on the host side rather than relying only on the server.
6. Finding 6: claimed fixed — the response body read is bounded.
7. Finding 8: claimed fixed — the new-session path no longer hardcodes submit, and both directions are pinned.
8. Finding 9: claimed fixed — the unit PATH test asserts PATH rather than being satisfied by ExecStart.
9. Findings 10 and 11: claimed fixed — server-controlled pane/wid are bounded before reaching the journal and validated before reaching the result URL.
10. Finding 7: claimed NOT fixed, deliberately — plaintext HTTP and a cleartext execution-grade token are flagged in the body as a known transport property.
11. New in this round: tmuxSessionName carried and re-validated host-side; and a nix-eval test was REMOVED rather than pinned as an expected skip, because it can never run in the gating sandbox tier.
```

ZacxDev:
```audit-claims round=2 audited=3fa3096ab..1adc0d47
1. 🔴1: open_window passes `-t "=<name>:"`, so tmux matches the session EXACTLY instead of prefix-matching; the docstrings that asserted a missing session would fail are corrected to state the measured behaviour.
2. 🔴1: a real-tmux test on a private `-L` socket, with a server shaped like the operator's (scratch20 live, scratch2 absent), asserts a missing session is refused and NO window is created; a control asserts the exact name still opens one.
3. 🔴2: validate() refuses a cwd that is not a directory on this host, and refuses a trailing separator or surrounding whitespace in it — the same character the text path already refused.
4. 🔴2: open_window reads back the session and pane_current_path tmux actually chose, from the same `-P -F`, and refuses on either mismatch before any keys are sent.
5. 🔴1+2: both read-backs are pinned INDEPENDENTLY by a stub that reports a landing different from the request, because `=` and isdir each fire first in the ordinary case and both read-backs survived their first mutation.
6. 🟡3: tmux_text_policy.first_disallowed is tested with the disallowed character at OFFSET 0, through the shared scan and through the agent's own gate, with a control that an allowed character at offset 0 still passes.
7. 🟡4: post_json's bounded `resp.read(MAX_RESPONSE_BYTES)` has a test asserting the limit argument is passed.
8. 🟡5: a non-string kind, expectTmuxServerId or any other field returns a refusal instead of raising; a raising write and every LATER write in the same batch are still reported, and the validator test covers eight fields.
9. 🟡6: every reported detail is clamped by redact() at the single point where it becomes output, so a validate() refusal no longer bypasses it.
10. 🟢10: scripts/lib/tmux_text_policy.py is in the unit's X-Restart-Triggers, so tightening the allowlist restarts the resident agent.
11. 🟢11: tmux_text_policy.py names the two guards that actually exist instead of a test file that does not.
12. tmux is in REQUIRED_TOOLS and flake.nix gateTools, with the written justification test_devshell_satisfies_required_tools.py requires, so the real-tmux guards cannot skip in the tier that gates.
13. Eight mutants run against this tree, each killed by the test named for it. Two — the session read-back and the path read-back — SURVIVED first time and the TESTS were changed, not the claim.
14. Verification on this tree: dev-host gate PASS (21779 collected / 21776 passed / 3 skipped / 0 failed; node 1449/1449) and both nix checks PASS, run one at a time (GATE_RC=0 NIXPY_RC=0 NIXNODE_RC=0). test_tmux_reply_agent.py 105 passed; test_session_write.py 185 passed, unchanged by the predicate extraction.
15. NOT fixed, and stated rather than claimed: 🟢9 (the ships-disabled guard pins one route, so a SECOND unit wanting tmux-reply-agent.service would survive it) and the plain-HTTP transport, which is flagged in the PR body as a property of the channel rather than a defect of this change.
```

PR #1326, opened 2026-09-05
docs(proposal): session capture attached to cairn entries

ZacxDev:
Proposes shipping a session's whole transcript to object storage at handoff time and attaching it to the cairn entries that session touched, so a future reader of an entry can reach the receipts instead of re-deriving them.

**Nothing is built.** No code, no bucket, no credential, no front-matter field. This is the argument and the shape.

## What's in it

**Eight settled decisions (§2)**, each recorded with the consequence it accepts so they are not re-litigated: whole file not tail · per-session trigger not the 5-minute timer · pointer in cairn, bytes in MinIO · ship raw with the storage boundary as the control · handoff ships + `SessionEnd` re-ships · pointer on every touched entry · fail-open on push failure · indefinite retention.

Decisions 4 and 8 compound into a permanent unredacted asset. §6 therefore builds the boundary like it is the only control — because it is.

**It argues against the letter of one decision, once, and says so.** The pointer should be an **opaque id, not a resolvable URL**. Fanning a URL across every touched entry puts a reference to a multi-client object inside a `client-confidential` entry, colliding with the already-taken decision that *tenancy is a security boundary, not a cache key*. An id keeps the fan-out exactly as chosen while making the shared artifact an identifier rather than content. Cheap to accept, expensive to retrofit.

**Why not the shipper that already exists.** `#1310`'s clawgate feeder is good code and the wrong instrument here — structurally, not stylistically: 192 KiB tail vs whole file, ambient timer vs explicit act, inline JSON vs object storage. The duplication is named rather than hidden, with the specific parts that should become shared.

## Measured, not estimated

| | |
|---|---|
| transcripts | 6.6 GB, 5,840 files |
| median / p90 / max | 739 KB / 2.5 MB / 23 MB |
| opencode | 2.4 GB of SQLite, **no jsonl** — separate project |

`SessionEnd` was verified to be a wired hook event on this host rather than assumed, since one option depended on it.

## Seven controls the implementation must be watched to FAIL

Positive and negative controls on the shipper · idempotency across two handoff runs in one session · a `SessionEnd` overwrite that actually changes the digest · the crash path still attaching the handoff-time floor · anonymous GET denied against the **real** bucket · cross-credential separation, both directions.

## Still open

Five questions in §9, led by the pointer-shape recommendation above.

---

No scope, entry, bucket or endpoint name appears, per the precedent `plan-cairn-phase1-cutover.md` set for this public repo.

**Gate:** content + skill-listing gates, 294 passed. Docs-only; other tiers not run and not claimed.

https://claude.ai/code/session_014wTuXEuj8Hs89X6H8xYpjq

### DIFF
--- a/claudedocs/proposal-cairn-session-capture.md
+++ b/claudedocs/proposal-cairn-session-capture.md
@@ -59,10 +59,25 @@ It is nonetheless **the wrong instrument here, and the reasons are structural, n
 will hold two answers to "which sessions matter", and those answers will drift — the exact
 shape `claude/RULES.md` calls out as regenerating the same bug at every site. The mitigation
 is not to merge them (their triggers genuinely differ) but to **share the parts that are one
-rule**: transcript discovery and `project_of`. **Not the byte-boundary logic** — it exists to
-drop the leading partial JSON record from a *tail*, and decision 1 ships whole files, so the
-new shipper has no caller for it. Extracting it would move dead surface for one of the two
-callers.
+rule**. **Not the byte-boundary logic** — it exists to drop the leading partial JSON record
+from a *tail*, and decision 1 ships whole files, so the new shipper has no caller for it.
+
+🔴 **AND TRANSCRIPT DISCOVERY IS ALREADY SHARED — REUSING IT IS A DECISION, NOT A FREEBIE,
+AND IT ANSWERS §9.6 IN THE OPPOSITE DIRECTION.** `scripts/lib/transcript_search.py` already
+exposes `iter_transcripts` / `is_corpus_member`; the existing feeder imports it, and
+`test_transcript_search.py`'s two-way site ledger scans the tree so a fourth hand-rolled walk
+cannot pass unseen — i.e. the repo actively rewards reuse. But that module carries
+`EXCLUDED_DIR_NAMES = ("subagents",)` and `is_corpus_member` rejects anything beneath it.
+Measured 2026-09-05: the exclusion and the `agent-*` prefix **coincide exactly** — 4,955 of
+4,955 `agent-*` files are under a `subagents/` directory and 0 of 913 real-session files are.
+
+So an implementer who takes "share discovery" at face value, reuses `iter_transcripts` — the
+correct-looking move, and the one the ledger test pushes toward — **silently ships parent
+transcripts only**, which is precisely the "drops most of the evidence this proposal exists to
+preserve, while looking complete" outcome §5.0 names. The ledger test stays green throughout,
+because reusing the shared walk is what it exists to reward. **If §9.6 resolves to "ship the
+set", the shared walk must be widened or deliberately bypassed, and that is a change to a
+module with an enforced site ledger — not a free adoption.**
 
 ## 4. Volume — measured, not estimated
 
@@ -80,28 +95,40 @@ Measured on the workbench, 2026-09-05. All three populations, because they diffe
 |---|---|---|---|---|---|
 | all `.jsonl` — **NOT what ships** | 5,864 | 0.74 MB | 2.53 MB | 23.48 MB | 6.73 GB |
 | real sessions (excluding `agent-*`) | 913 | **2.62 MB** | 4.97 MB | 23.48 MB | 2.45 GB |
-| **sessions that produced a handoff** — what decision 2 selects | 37 | **3.82 MB** | 5.22 MB | 10.59 MB | 148 MB |
+| **sessions that produced a handoff** — what decision 2 selects, **parent file only** | 37 | **3.82 MB** | 5.22 MB | 10.59 MB | 148 MB |
+| the same 37 sessions, **parent + their subagent transcripts** | 37 | **8.68 MB** | — | 31.00 MB | 375 MB |
 
-The third row is the population decision 2 actually selects. It was derived from the
+Rows 3 and 4 are the population decision 2 selects, counted two ways — see the next
+paragraph but one for which of them to size off. Both were derived from the
 `Claude-Session-Id` trailer on commits touching `claudedocs/handoff-*.md` since 2026-07-01:
 37 distinct ids, **all 37** resolving to a transcript on disk, none of them an `agent-*`
 file.
 
 🔴 **The selection effect runs the WRONG WAY, and that is the point of the row.** Sessions
 that produce a handoff are the long ones, so the per-handoff filter picks the **large tail**,
-not the median — 3.82 MB against 2.62 MB against 0.74 MB. `SessionEnd` re-ship makes it
-larger still, so **3.82 MB is a floor, not an estimate.** Anyone sizing a bucket, a PUT
-timeout or a per-handoff cost must use that row.
+not the median — 3.82 MB against 2.62 MB against 0.74 MB.
+
+🔴 **AND THE ROW YOU SIZE OFF DEPENDS ON AN ANSWER §9.6 HAS NOT GIVEN YET.** Row 3 is the
+parent transcript alone. If §5.0 resolves to "ship the set", row 4 is the real unit: **2.3×
+the median and 2.5× the total**, with a largest single session of **31.00 MB** — which
+*exceeds* the 23 MB figure §5.2 uses as its worked example of a payload too big for the pod.
+33 of the 37 sessions have subagent bytes at all; the worst measured is a 3.79 MB parent with
+60 subagent files totalling 43.20 MB, where shipping one object captures **8%** of the
+session. Two further reasons both rows are floors: `SessionEnd` re-ship only grows them, and
+the sample below is partial.
+
+**So: size off row 4 unless and until §9.6 resolves to parent-only.** Sizing off row 3 while
+§9.6 is open is how a bucket, a PUT timeout or a per-handoff cost estimate comes out low by
+more than 2×.
 
 ⚠ **Scope of the 37-session sample, stated rather than buried:** devrc handoffs only, and
 only commits carrying the trailer. Handoffs in other repos are not counted, so the true
 population is larger and the totals are floors.
 
-| | |
-|---|---|
-| opencode session data | **2.4 GB** — a SQLite DB plus `storage/`, `snapshot/`, `tool-output/` |
+**opencode**, for scale only: **2.4 GB** — a SQLite DB plus `storage/`, `snapshot/` and
+`tool-output/`. Not comparable to the rows above and not summed with them.
 
-⚠ The `6.73 GB` above is the `.jsonl` population summed in decimal bytes. A `du -sh` of the
+⚠ The `6.73 GB` in row 1 is the `.jsonl` population summed in decimal bytes. A `du -sh` of the
 directory reads `6.6 GB` because it is GiB and includes ~7,000 non-`.jsonl` files. Two
 different measurements; do not pair them.
 
@@ -127,7 +154,8 @@ work at all."* For an `agent-*.jsonl` the stem is `agent-<hash>`, which is **not
 id. Sampled 6 subagent transcripts: in every one the stem is absent from the `sessionId`
 values inside the file, and the single `sessionId` present is the **parent's**.
 
-So "keyed on session id" (§5.1) has two readings and neither is safe by default:
+So keying the object "on session id" — which is what §5.1 said before it was rewritten to
+defer to this section — has two readings, and neither is safe by default:
 
 - key on the **in-record `sessionId`** → parent and all N subagent files collide on one
   object key, each push overwriting the last. §8 control 4 ("assert the sha256 changed")
@@ -179,8 +207,9 @@ imply "over the mesh" — an implementer who accepts only the capacity argument 
 MinIO through a public ingress satisfies it completely while putting unredacted multi-client
 transcripts through a third party. **Name the rule when writing this down.**
 
-The capacity point is true and secondary: the pod is single-replica, `Recreate`, PVC-backed
-and designed to render markdown (all three verified), so a 23 MB PUT through it is a category
+The capacity point is true and secondary: the pod is single-replica, `Recreate` and
+PVC-backed (all three verified) and is designed to render markdown, so a 23 MB PUT — or a
+31 MB one, per §4 row 4 — through it is a category
 change. But that is a reason not to use the pod — it is not the reason the route must stay on
 the mesh.
 
@@ -192,9 +221,16 @@ Front-matter on each touched entry gains a session reference carrying `id`, `sha
 🔴 **THE FIELD SHAPE IS CONSTRAINED, AND THE OBVIOUS SPELLING IS SILENTLY DESTRUCTIVE.**
 Cairn front matter is parsed by `parse_front_matter` in `lib/subsystem_resolver.py`, which is
 hand-rolled and **line-based**. It handles `key: value`, an inline flow list `key: [a, b, c]`,
-and a block list of one-line `- item`s. **There is no case for a nested mapping**, and its own
-docstring records the measured result of feeding it one: the key comes back empty and every
-child is promoted to a phantom top-level key by its own internal colon.
+and a block list of one-line `- item`s. **There is no case for a nested mapping**: the key
+comes back empty and every child is promoted to a phantom top-level key by its own internal
+colon.
+
+⚠ **Do not read the parser's docstring as documenting this** — an earlier revision of this
+paragraph cited it, and the citation was wrong in a way that would close the case for a reader
+who checked it. The docstring records the same corruption for a **block list**, under the
+heading *"THE BLOCK FORM IS PARSED BECAUSE NOT PARSING IT CORRUPTED THE MAPPING"* — i.e. that
+shape was **fixed**. The nested-mapping case is unfixed and undocumented there; the evidence
+for it is the reproduction below and nothing else.
 
 Reproduced against the live parser while writing this, with the exact five fields above under
 a `session:` key:
@@ -281,15 +317,19 @@ depth now rests on storage. Concretely, the implementation must:
   write-capable key whose secret exists nowhere — the exact window #683 closed. The
   stdout half matters too (these scripts are run by agents, so a printed secret lands in a
   transcript), and with decision 8 it would land there **permanently**;
-- carry **no anonymous access policy** — see §8 control 6 for how to prove that, which is
-  harder than it looks;
+- carry **no anonymous access policy**, proven by probing it rather than by reading the
+  policy JSON — and see §8 control 6, because probing it correctly is harder than it looks;
 - keep the **cairn read token unable to reach the bucket**, and vice versa — pinned by a
   test, not asserted by a comment. ⚠ Note the limit stated in §5.3: this separates
   credentials, not the host they both sit on.
 - 🔴 **register the credential in `SECRETS.md` with a rotation coupling, in the same change
   that mints it.** That file exists to make a new-host bootstrap deterministic instead of
-  manual archaeology, and every comparable entry carries one. A credential minted outside it
-  is invisible at bootstrap and at rotation.
+  manual archaeology. ⚠ **Naming the population, since this document's own §4 is about not
+  doing that:** of `SECRETS.md`'s 8 host-file rows, 6 bear a secret and **2 of those 6** carry
+  a rotation-coupling clause — both of them hosted-service bearer tokens with a k8s source of
+  truth, which is exactly what this credential would be. So the precedent is strong for this
+  shape and is *not* a universal convention. A credential minted outside that file is
+  invisible at bootstrap and at rotation either way.
 
 🔴 **The bucket has NO tenancy boundary, and the store it attaches to does.** Under
 `plan-cairn-integration.md` the store's future is multi-tenant with sharing gated on
@@ -372,9 +412,10 @@ invariant the bug never violated is an invariant guard and must be labelled as o
 3. **Which entries count as "touched"?** The handoff run knows what it wrote; whether that
    set is the right one, or too wide, has not been measured.
 4. **opencode** — schedule, or park indefinitely.
-5. **Does the shared-module extraction in §3 happen now or later?** Later is defensible;
-   never is how the two shippers drift. Scope it to discovery and `project_of` — not the
-   byte-boundary logic.
+5. **What does the new shipper do about `transcript_search`?** Not "does the extraction
+   happen" — discovery is **already** a shared module with an enforced site ledger. The
+   question is narrower and sharper: reuse it and inherit its `subagents/` exclusion, widen
+   it, or bypass it deliberately. §3. This is downstream of question 6, not independent of it.
 6. 🔴 **Does a session ship as ONE object or a SET?** §5.0. This is the largest open
    question, it decides the object key, and the goal in §1 depends on the answer — 64% of the
    bytes are in subagent transcripts, and one-object-per-session drops them.
--- a/claudedocs/proposal-cairn-session-capture.md
+++ b/claudedocs/proposal-cairn-session-capture.md
@@ -50,7 +50,8 @@ and a distinct exit code per failure condition.
 
 It is nonetheless **the wrong instrument here, and the reasons are structural, not stylistic**:
 
-- it ships a **192 KiB tail**; decision 1 wants whole files up to 23 MB;
+- it ships a **192 KiB tail**; decision 1 wants whole files, up to 23.48 MB — §4 row 1's max,
+  i.e. the largest single transcript FILE on the host;
 - it is **ambient and periodic**; decision 2 wants an explicit per-session act;
 - it posts the payload **inline in JSON** to an HTTP endpoint; decision 3 wants bytes in
   object storage and only a pointer in the API.
@@ -59,8 +60,20 @@ It is nonetheless **the wrong instrument here, and the reasons are structural, n
 will hold two answers to "which sessions matter", and those answers will drift — the exact
 shape `claude/RULES.md` calls out as regenerating the same bug at every site. The mitigation
 is not to merge them (their triggers genuinely differ) but to **share the parts that are one
-rule**. **Not the byte-boundary logic** — it exists to drop the leading partial JSON record
-from a *tail*, and decision 1 ships whole files, so the new shipper has no caller for it.
+rule**.
+
+**The one clearly-shareable surface is `project_of`** (`build_transcript_push.py`), which maps
+a transcript to its project label. It exists at exactly one site, is **not** in the shared
+discovery module, and its own docstring warns that it is *"a LABEL, NOT AN IDENTIFIER"* that
+must not be un-slugified. A second shipper needing a project label for object metadata or a
+key prefix would hand-roll it in four lines, and the two copies would disagree the first time
+a directory name contains a literal `-`. ⚠ An earlier revision of this section deleted this
+sentence as collateral while rewriting the paragraph below, leaving the section with no
+positive advice at all — which reads as "nothing is safe to share" and is the opposite of what
+was found.
+
+**Not the byte-boundary logic** — it exists to drop the leading partial JSON record from a
+*tail*, and decision 1 ships whole files, so the new shipper has no caller for it.
 
 🔴 **AND TRANSCRIPT DISCOVERY IS ALREADY SHARED — REUSING IT IS A DECISION, NOT A FREEBIE,
 AND IT ANSWERS §9.6 IN THE OPPOSITE DIRECTION.** `scripts/lib/transcript_search.py` already
@@ -89,14 +102,19 @@ was right; the population was wrong. Kept here rather than silently corrected, b
 error is the instructive part: **a percentile is a claim about a population, so name the
 population or the number means nothing.**
 
-Measured on the workbench, 2026-09-05. All three populations, because they differ by 5×:
+Measured on the workbench, 2026-09-05. Four populations, because the answer spans ~12×
+between them — and which one applies is decided by a question §9 has not closed:
 
 | population | files | median | p90 | max | total |
 |---|---|---|---|---|---|
 | all `.jsonl` — **NOT what ships** | 5,864 | 0.74 MB | 2.53 MB | 23.48 MB | 6.73 GB |
 | real sessions (excluding `agent-*`) | 913 | **2.62 MB** | 4.97 MB | 23.48 MB | 2.45 GB |
 | **sessions that produced a handoff** — what decision 2 selects, **parent file only** | 37 | **3.82 MB** | 5.22 MB | 10.59 MB | 148 MB |
-| the same 37 sessions, **parent + their subagent transcripts** | 37 | **8.68 MB** | — | 31.00 MB | 375 MB |
+| the same 37 sessions, **parent + their subagent transcripts** | 288 files / 37 sessions | **8.68 MB** *(per session)* | 20.23 MB | 31.00 MB *(per session)* | 375 MB |
+
+⚠ Every percentile in this table uses one convention — the sorted value at `int(n × 0.9)`,
+no interpolation. Stated because a different convention gives a visibly different p90 for
+row 4 (16.65 MB rather than 20.23 MB) and nothing in the table would show which was used.
 
 Rows 3 and 4 are the population decision 2 selects, counted two ways — see the next
 paragraph but one for which of them to size off. Both were derived from the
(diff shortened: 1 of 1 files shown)


ZacxDev:
## Round 1 — fixes landed

Auditor returned 3 🔴, 9 🟡, 4 🟢. All three 🔴 were re-verified against the live system before being acted on, not accepted from the report:

- **population split** — reproduced exactly: 5,864 `.jsonl`, 4,951 (84%) `agent-*`, real-session median 2.62 MB vs the 0.74 MB the doc quoted;
- **parser corruption** — reproduced against the live `parse_front_matter` with the exact five proposed fields: `session: ''` plus five phantom top-level keys;
- **handoff-session sizing** — reproduced independently (37 ids, all resolving, median 3.82 MB). My first attempt returned 0 from a wrong pathspec; I discriminated that rather than reporting the zero.

```audit-claims round=1 audited=eda003cc..4f0f1708
1. §4 re-measured against three named populations; the ~5x-wrong "well under a megabyte" conclusion replaced with the 3.82 MB handoff-session floor, retraction kept in place, sample scope stated.
2. §5.0 added: a session's transcripts are a SET; subagent stems are not session ids (verified on 6 files); the two readings of "keyed on session id" named; filed as open question 6 rather than decided.
3. §5.3 now states the front-matter shape constraint, with the live parser's corrupting output reproduced inline, and flags the parser widening as a two-repo change.
4. §2 decision 5's consequence corrected to "whole transcript on a clean exit; handoff-time floor on a crash".
5. §5.1 records that SessionEnd is wired on one host only, is unmanaged per-host state, and that drift-check cannot see its absence.
6. §5.3's leak argument re-aimed: decision 11 forecloses the client-entry case; the live hazard is a SHAREABLE entry pointing into client-confidential content.
7. §5.3 downgrades "separate by construction" to the threat model it actually covers (one credential, no host access).
8. §6 now requires record-before-create per #683's actual finding, not just the never-print half.
9. §5.2 names the store's third-party rule as the binding constraint; the pod-capacity argument demoted to secondary.
10. §8 control 6 now requires an EXISTING key and asserts 403 specifically, plus LIST and PUT; control 7 given named observables; new control 8 covers decision 7's fail-open behaviour.
11. §6 adds the SECRETS.md row requirement, the bucket's absent tenancy boundary, and the missing retraction path.
12. §3's shared-module list drops the byte-boundary logic; §5.3's "against the letter of decision 6" framing removed; §4's unit basis separated; one quotation made exact.
```

**Gate:** content, skill-listing and path-rot gates on the merged tree — 371 passed. Docs-only; other tiers not run and not claimed.

Round 2 (delta) dispatching now — round 1 produced findings that needed fixing, so the ladder continues.

ZacxDev:
## Round 2 — fixes landed

2 🟡, 3 🟢 — every one of them in prose round 1 had just written, which is the documented failure mode for a prose payload. Both 🟡 re-verified against the live system before acting:

- **parent vs set sizing** — reproduced exactly: median 8.68 MB vs 3.82, total 375 MB vs 148, largest session 31.00 MB vs 10.59, 33 of 37 sessions carrying subagent bytes. The 31.00 MB exceeds the 23 MB §5.2 uses as its own "too big for the pod" example;
- **shared discovery already exists and excludes the bytes** — `transcript_search.py` carries `EXCLUDED_DIR_NAMES = ("subagents",)`, the existing feeder imports it, a two-way site ledger test rewards reuse, and the exclusion coincides exactly with the `agent-*` prefix: 4,955/4,955 and 0/913.

The second is the one worth calling out: round 1's own §3 advice would have pushed an implementer toward the single move that silently ships parent transcripts only — with the ledger test green throughout, because reusing the shared walk is precisely what it exists to reward.

```audit-claims round=2 audited=4f0f1708..f966fbd7
1. §4 row 4 added (parent + subagents: 8.68 MB median, 375 MB total, 31.00 MB max) and the size-off-this instruction now names which row and why it depends on §9.6.
2. §3 and §9.5 rewritten: transcript_search is ALREADY the shared discovery module and excludes subagents/, so adopting it answers §9.6 in the opposite direction rather than being a free win.
3. §5.0 no longer quotes a phrase that the same commit deleted from §5.1.
4. §6's "every comparable entry carries one" replaced with the named population: 2 of 6 secret-bearing SECRETS.md rows, both the same shape as this credential.
5. §5.3 no longer attributes the nested-mapping corruption to the parser docstring, which records a fixed BLOCK-LIST case; the claim now rests on the inline reproduction alone and says so.
6. Three declined nits fixed: "all three verified" trailed four properties; the opencode row was an orphaned table; control 6's "rather than by reading the policy JSON" was restored.
```

**Ledger:** round 2 · payload lines changed by round 2's fixes: **67 insertions / 26 deletions** (round 1: 224/57). Measured `git log --numstat --format= --remerge-diff 4f0f1708..f966fbd7 --not origin/main` — rc 0, stderr silent, range non-empty (1 commit). 100% payload: for a prose PR the `.md` is the deliverable, so the two-zero-rounds attribution gate is structurally inert here.

**Gate:** content, skill-listing and path-rot gates — 371 passed. Docs-only; other tiers not run and not claimed.

Round 3 dispatching — round 2 produced findings that needed fixing, so the ladder continues.

ZacxDev:
## Round 3 — fixes landed

2 🟡, 4 🟢, **no 🔴**. Both 🟡 re-verified before acting.

The 🟡 worth naming: §4's worst-case example — *"a 3.79 MB parent with 60 subagent files, 8% captured"* — is an exact figure **from the wrong population**. Measured `in37=False`; only two sessions on disk have 60 subagent files and neither is in the selected 37. Its 46.99 MB total also contradicted the 31.00 MB row-4 max stated three lines above it. I had imported that number from the round-2 report and placed it without checking which population it came from — committed *inside* the section whose entire thesis is that a number without a named population means nothing. True worst within the 37 is 12.6%.

The other: `project_of` was deleted from §3 and §9.5 as collateral when round 2 rewrote the discovery paragraph, leaving the section with no positive advice at all — "nothing is safe to share", the opposite of what round 1 found. Wider on one axis, narrower on another.

**Two things this round found that the audit did not**, both from doing the sweep rather than fixing only what was reported:

- the recurring shape is *a quantity stated without its population*, so I enumerated **every** quantitative claim in the document and checked each names one. Two more sites had the same defect — §3's "whole files up to 23 MB" and §5.2's "a 23 MB PUT" — neither reported;
- the row-4 p90 I was about to copy from the report (16.65 MB) **does not reproduce** under the convention every other row in the table uses. It is 20.23 MB. The table now states its percentile convention, since nothing in it would have shown which was used.

```audit-claims round=3 audited=f966fbd7..269376b0
1. §4's worst-case example replaced with the true in-37 worst (12.6%, 3.82 MB parent, 19 subagent files, 30.39 MB session); the wrong-population import recorded rather than swapped out.
2. project_of restored to §3 and §9.5 as the one unconditionally-shareable surface, with the round-2 collateral deletion recorded.
3. §4 caption corrected: four populations, ~12x spread.
4. Row 4's files column now reads 288 files / 37 sessions, and its median/max are labelled per-session.
5. The 31.00 MB session-total vs object-size conflation fixed on both sides (§4 and §5.2); per-object sizing now points at row 1's max, per-handoff cost at row 4's total.
6. §5.0 gained a note that the corpus counts are a live snapshot that drifts (4,951 / 4,955 / 4,957), and that the durable claims are the ratio and the coincidence, not the absolutes.
7. SWEEP beyond the findings: every quantitative claim checked for a named population; two unreported sites fixed (§3 and §5.2's bare "23 MB").
8. Row-4 p90 corrected from a report figure that did not reproduce (16.65 MB) to the measured 20.23 MB, and the table now states its percentile convention.
```

**Ledger:** round 3 · payload lines changed by round 3's fixes: **59 insertions / 19 deletions**. Rounds 1 and 2 changed 224/57 and 67/26. Measured `git log --numstat --format= --remerge-diff f966fbd7..269376b0 --not origin/main` — rc 0, stderr silent, 1 commit. 100% payload.

**Gate:** content, skill-listing and path-rot gates — 371 passed. Docs-only; other tiers not run and not claimed.

Round 4 dispatching — round 3 produced findings that needed fixing, so the ladder continues.

ZacxDev:
## Round 4 — fixes landed

1 🟡, 2 🟢, **no 🔴**.

**The 🟡 is the sharpest finding of the ladder.** Round 3's fix retracted a worst-case figure from the wrong population — and then closed with *"the largest sessions on the whole host do reach ~47 MB"*. That is 46.99 MB: **the total of the very session being retracted**, promoted from "an example that does not belong here" to "the host maximum". Measured: the host maximum is **94.02 MB** (11.72 MB parent + 82.30 MB across 29 subagent transcripts), and **nine sessions exceed 47 MB**. A reader sizing for the widened population §4's own caveat anticipates would have been **2× low** — from a sentence written to warn against exactly that.

Three consecutive revisions of that one paragraph made the same mistake in three different places, so the fix records the rule rather than only the number: **retracting a number is not the same as not using it.**

🟢 Row 1 is labelled *NOT what ships* and was then used three times to size what ships. The number is a deliberate over-estimate (largest single file among the 37 selected sessions is 10.59 MB, 2.2× smaller) — but the stated justification was wrong, and it now cites §4's own coverage caveat instead. 🟢 §4's table needed the same live-population caveat §5.0 already had.

**One of the audit's own numbers did not reproduce.** It reported the id set as 43 today under the stated method — a 16% move. Re-measured here: **still 37**, with 0 ids from commits dated today. (My first attempt returned 0 from a format-string change; I discriminated that against a known-good control rather than reporting it.) The finding's *principle* holds and is fixed; its *number* is recorded as non-reproducing — which is precisely why the method is stated in the table.

```audit-claims round=4 audited=269376b0..8a1b8e87
1. The "~47 MB host maximum" removed; measured host max 94.02 MB and 9 sessions above 47 MB stated, with the re-import recorded as its own lesson.
2. §5.2 and §3's row-1 justification corrected: 23.48 MB is a deliberate over-estimate for headroom under §4's coverage caveat, not "the right population"; the 10.59 MB selected maximum stated plainly.
3. §4's table gained the live-population caveat §5.0 already carried, distinguishing staleness from wrongness.
4. The audit's non-reproducing "43" recorded as non-reproducing, with the re-measurement (37, 0 from today).
5. Sweep applied to the fix's OWN new prose this time: every magnitude re-enumerated, three bare medians annotated with their rows, remaining hits confirmed scoped on adjacent lines.
6. Three nits: orphaned wrap rejoined, vague back-reference now names its distance, 84% / ~85% reconciled to one measurement (84.4%).
```

**Ledger:** round 4 · payload lines changed by round 4's fixes: **36 insertions / 11 deletions**. Rounds 1-3: 224/57, 67/26, 59/19 — monotonically declining. Measured `git log --numstat --format= --remerge-diff 269376b0..8a1b8e87 --not origin/main`, rc 0, silent stderr, 1 commit.

**Gate:** content, skill-listing and path-rot gates — 371 passed. Docs-only; other tiers not run and not claimed.

Round 5 dispatching — round 4 produced findings that needed fixing, so the ladder continues.

ZacxDev:
## Round 5 — fixes landed, and a correction to my own round-4 comment

### 🔴 Correction, public rather than silent

My round-4 comment above said the id set measures *"(37, 0 from today)"*. **The "0 from today" is wrong.** Measured: 37 total, 36 before 2026-09-05, and **1 id enters solely from a commit dated today**. The document itself says only "still 37, so no drift was observed", which does reproduce — so this was a wrong claim in my comment, not a defect in the payload. Leaving the original comment standing and correcting it here, since a reviewer may already have read it.

### The findings

2 🟡, 2 🟢, no 🔴 — both 🟡 inside the paragraph about number-import discipline, which had acquired a wrong count *and* a wrong distance.

- *"three separate revisions … in three different places"* → it was **two**. `eda003cc` and `4f0f1708` contain neither the paragraph nor any of the figures; only `f966fbd7` and `269376b0` carry the mistake.
- *"the coverage caveat two paragraphs below"* → it is **five** below, and round 4 itself inserted the two blocks that widened the gap before naming a distance smaller than either count. Following that pointer lands on the **94.02 MB whole-host** paragraph — so a reader would import a host figure into a 37-session claim, the precise hazard §4 exists to prevent, licensed by a pointer added to be *more* precise.

### The fix is structural, because the instance is not the problem

A paragraph **distance** and a **self-count** are maintenance burdens that the next inserted block falsifies — and this ladder has now falsified one of each, in consecutive rounds. Correcting them to "five" and "two" would leave the generator running.

Both are replaced by **names**: the block is titled `THE COVERAGE CAVEAT` at its own site and referenced by that name, and the count is replaced by naming what each of the two revisions did. **Swept at every site rather than the two reported** — two further positional pointers ("the paragraph below", "three lines above it") turned up and are gone. The document now contains **zero** positional or self-counting pointers.

🟢 The `84%` / `~85%` pair: only one measurement ever existed (84.4%); the round-4 reconciliation had left the `~85%` standing one sentence above it. 🟢 The audit's `43` is now **explained rather than non-reproducing** — it is what the trailer yields when the **pathspec is dropped**. The method now names its ref too (`origin/main`; `--all` varies by clone and measures nothing shared).

```audit-claims round=5 audited=8a1b8e87..aef34fa7
1. "three separate revisions in three different places" corrected to two, and replaced by naming what each revision did rather than counting them.
2. "two paragraphs below" replaced by a NAME: the block is titled THE COVERAGE CAVEAT at its own site and referenced by that name everywhere.
3. The sweep extended to every positional pointer in the document; two unreported ones ("the paragraph below", "three lines above it") eliminated. Zero remain.
4. The ~85% left standing by round 4's reconciliation removed; one measurement (84.4%) stated once.
5. The audit's 43 explained as the pathspec-dropped count, and the stated method now names its ref as origin/main.
6. Cosmetic: a ⚠ that rendered inside the preceding paragraph now has its blank line.
```

**Ledger:** round 5 · payload lines changed by round 5's fixes: **29 insertions / 15 deletions**. Rounds 1-4: 224/57, 67/26, 59/19, 36/11. Measured `git log --numstat --format= --remerge-diff 8a1b8e87..aef34fa7 --not origin/main`, rc 0, silent stderr, 1 commit.

**Gate:** content, skill-listing and path-rot gates — 371 passed. Docs-only; other tiers not run and not claimed.

Round 6 dispatching — round 5 produced findings that needed fixing, so the ladder continues. It is pointed at whether the structural fix actually removed the generator or merely relocated it.

ZacxDev:
## Round 6 — 1 🔴 refuted, 2 🟡 fixed

### The 🔴 does not reproduce

The audit claims the attribution is inverted — that the document's stated method yields **43** and the pathspec-dropped count is **59-60**. Measured the opposite, at four refs including the exact one it names:

| ref | stated method (pathspec) | pathspec dropped |
|---|---|---|
| `origin/main` `4d249c0c` | **37** | 43 |
| `c4e76a10` | **37** | 42 |
| this branch's head | **37** | 43 |
| base clone HEAD | **37** | 43 |

I also tested the obvious mechanism for their number — a **shell-expanded** rather than git-globbed pathspec — and that gives 37 too. Their figure does not reproduce here by any variant I can construct.

**But the signal behind it is real, and is now fixed properly.** Two competent independent re-derivations *from the prose description of the method* disagreed with each other and with the table. A method described in prose is not a method. The **exact command is now in the document**, fenced, with the refs it was reproduced at — so the next re-derivation either matches or finds a real change, instead of landing somewhere unattributable.

### 🟡 The sweep's instrument was the problem, not the diligence

Round 5 claimed zero positional pointers remained. **False** — *"see the next paragraph but one"* has been in §4 since round 2. The reason is worth more than the fix: **the sweep grepped line by line, and the phrase wraps across a newline.** The instrument was structurally incapable of seeing the survivor, so a clean sweep meant nothing. Re-swept newline-insensitively — that found it, and nothing else. Replaced with a name, `THE SIZING DIRECTIVE`, defined once at its own site.

### 🟡 The paragraph about not counting things made a false causal claim

It said the count and the distance *"were falsified by the next edit that inserted a block."* Measured across every revision: **both were wrong at the revision that introduced them.** Drift is the more comfortable story because it blames time rather than the writing. The lesson is now the sharper thing the evidence supports: nobody counts paragraphs before writing "two paragraphs below" — a name can be checked when it is written; an ordinal is a measurement nobody takes.

```audit-claims round=6 audited=aef34fa7..be36fb5d
1. The 43-vs-37 inversion claim refuted by measurement at four refs, including origin/main 4d249c0c, and with the pathspec both quoted and shell-expanded.
2. The exact derivation command is now IN the document as a fenced block with the refs it reproduced at, replacing the prose description two auditors re-derived differently.
3. The surviving ordinal "see the next paragraph but one" replaced with the name THE SIZING DIRECTIVE, defined once at its own site.
4. The sweep instrument corrected to be newline-insensitive and re-run; one survivor found, nothing else.
5. The false causal claim ("falsified by the next edit") corrected to what the revisions show: both were wrong at introduction.
6. The coverage-caveat reference no longer quotes a heading round 5 renamed; both names now resolve, each defined once.
```

**Ledger:** round 6 · payload lines changed by round 6's fixes: **23 insertions / 12 deletions**. Rounds 1-5: 224/57, 67/26, 59/19, 36/11, 29/15.

**Gate:** content, skill-listing and path-rot gates — 371 passed. Docs-only; other tiers not run and not claimed.

Round 7 dispatching, because round 6 produced findings that needed fixing. It is the last round I intend to run unless it surfaces something about the **design** rather than about the document's own retraction records — see the stopping note I will post with its result.

ZacxDev:
## Round 7 — fixes landed, and **the ladder stops here**

### The carried 🔴 is refuted, independently

Round 7 ran the command exactly as the document states it, at three distinct commits and five invocation variants: **37** every time, with **43** as the pathspec-dropped figure — exactly what the document says. Round 6's audit figure reproduces under no construction either of us could build. That dispute is closed.

### 1 🟡, 5 🟢 — all fixed

- 🟡 **`THE SIZING DIRECTIVE`'s only reference wrapped across a newline**, so a line-oriented grep saw the name once: defined, apparently unreferenced, one tidy-up away from deletion. That is the *exact* instrument failure round 6 diagnosed, reintroduced by the fix written to close it.
- 🟢 The "exact command" was not exact — verbatim it prints 407 lines, and `sort -u` alone gives **38**, because commits without the trailer emit an empty line that counts as a value. The dedupe was still prose; it is now in the block.
- 🟢 The 43/`--all` explanation was stated twice in two spellings, 70 lines apart. Now once.
- 🟢 The record mislabelled its own artifact as "a paragraph COUNT" when it was a **revision** count. Both now quoted verbatim.
- 🟢 **Control 3 was wrong under one branch of §9.6 — a design defect, not prose.** It asserted "two handoff runs → one object", false under "ship the set" (288 files across 37 sessions), so an implementer writing it as specified would assert `len(objects)==1` against a *correct* implementation. Restated as NO GROWTH, true under both branches.
- 🟢 **Partial fan-out had no owner**, now §9.7. Decision 6's pointer write is N-way and can land on 3 of 5; §5.5 covers the push, §6 covers deletion, and §5.4's ledger validates pointers that *exist* — so an entry that silently got none is indistinguishable from one the session never touched.

🔴 **Structural:** the volume table now carries a **written row-number column**. Fourteen sites say "row 1".."row 4" against a table that had none, so every one was an unwritten ordinal a future inserted row would silently falsify — including THE SIZING DIRECTIVE itself. Same class as the paragraph ordinals, at the site neither the sweep's patterns nor any round had named.

```audit-claims round=7 audited=be36fb5d..404ec208
1. THE SIZING DIRECTIVE's reference unwrapped so the name greps as defined AND referenced.
2. The dedupe pipeline (grep . | sort -u | wc -l) moved into the fenced command, with why: raw prints 407 lines and sort -u alone gives 38.
3. The 43/--all explanation consolidated to one site; the second now refers to it rather than restating it.
4. The retraction record now quotes both artifacts verbatim instead of mislabelling the revision count as a paragraph count.
5. Control 3 restated as NO GROWTH, correct under both branches of 9.6, instead of "one object" which is false under ship-the-set.
6. Partial fan-out filed as 9.7 with why the digest ledger cannot detect it.
7. The volume table gained a written row-number column, retiring the unwritten-ordinal class at all 14 reference sites.
```

**Ledger:** round 7 · payload lines changed by round 7's fixes: **44 insertions / 20 deletions**. Rounds 1-6: 224/57, 67/26, 59/19, 36/11, 29/15, 23/12.

---

## Why I am stopping, named rather than implied

The stop conditions for a prose payload are met, and I am stating the reasoning so that "the ladder ended" is distinguishable from "the ladder converged":

- **No 🔴 survives.** The only two 🔴s the ladder produced (rounds 6 and 7) were the same claim, and both were refuted by measurement rather than argument.
- **Blast radius is bounded by "the document contains a false sentence."** Nothing is built; nothing runs.
- **The recurring shape has been swept at every site, not just where reported** — paragraph ordinals, self-counts, the table-row ordinal class nobody had named, and the sweep *instrument* itself, which was line-oriented and structurally blind to wrapped prose until round 6.

**And here is why the rounds would not stop on their own.** Since round 3, every finding has been about prose *the ladder itself generated* — the retraction records. Each fix to a retraction record writes new checkable historical prose, which the next round checks. Rounds 6 and 7 each caught the fix round **reintroducing the very failure mode it was closing**. That is a self-sustaining generator, and it is the prose analogue of a ladder auditing scaffolding it just wrote.

**What I am NOT fixing, so it reads as open rather than absent:** the retraction records stay. They are the most valuable thing in §4 — they are why nobody re-derives the wrong-population figure — and they are also the generator. I am accepting residual risk of further imprecision in them rather than deleting them or continuing to round 8. The design content, by round 7's own assessment, is sound: §5.0's set-vs-file framing, §5.2's mesh rule, §5.3's live-parser reproduction, §6's before-not-after credential ordering and §8's DENIED-vs-ABSENT control are each specific, falsifiable and correctly reasoned.

**Not merging.** This is a proposal with nine open questions that are the operator's to answer — §9.1 (opaque id vs URL) and §9.6 (one object or a set) being the two that change what gets built.

ZacxDev:
## Eight of nine open questions answered — §9 is nearly closed

The operator answered §9.1, .2, .3, .5, .6, .7, .8 and .9. They are now **decisions 9-16 in §2**, stated once there, with every conditional they resolved unwound in the body rather than left hanging.

| # | decision |
|---|---|
| 9 | a session ships as a **SET** — one object per transcript file |
| 10 | the pointer is an **opaque id**, never a resolvable URL |
| 11 | the expected **fan-out set is recorded** |
| 12 | front matter is a **block list of ids**; digest/size/host as object metadata |
| 13 | `transcript_search` widened with an **opt-in flag** |
| 14 | 🔴 **no retraction path** — rotate the leaked credential instead |
| 15 | `SessionEnd` wired per host, **drift-check extended** to detect its absence |
| 16 | "touched" = entries the handoff run **wrote or updated** |

### 🔴 Decision 14 diverged from the recommendation, and it changes what the rest is for

With no deletion there is no retraction, so §5.4's digest ledger stops being a retraction-detector and becomes an integrity check. More importantly, **the report-only scan is promoted from a follow-on to a v1 control**: redaction was declined (decision 4) and deletion is now declined (decision 14), which leaves detection as the only remaining link in the chain. If the sole response to a leak is rotation, then *knowing* is the entire response — and nothing else in this design would tell you.

§6 also now records the limit of that policy: rotation only works on credentials that **can** be rotated. A client's secret, a token you don't own, or personal data in a transcript has no rotation, and for those the policy is the whole of the response.

### What decision 9 forced

It settled the sizing (row 4, not row 3) and forced decision 13. It also leaves one constraint it does not itself settle: an `agent-<hash>` stem joins to nothing, so the object key must be **(parent session id, transcript filename)** — using the parent id verified present and singular inside every subagent file sampled.

### §9 and the new §10

§9 now carries one question — **opencode**, with a recommendation to park it: no `.jsonl`, 2.4 GB of SQLite plus three storage dirs, so "the whole session" means an exporter with its own fidelity question.

A new **§10** collects what implementation must decide that no decision settles: the object key; where the fan-out set is recorded; that an unresolvable id must read as NOT AUTHORISED *distinctly* from NOT FOUND — which matters more now, because under decision 14 nothing is ever deleted, so "missing" means something went wrong rather than something was retracted; and that the opt-in flag must default to today's behaviour, with a test pinning the existing feeder **before** the module is touched.

Swept after the rewrite: zero positional pointers, both names still defined once and referenced, and all eight new decisions actually referenced in the body rather than only tabled — **decision 16 was tabled-but-unapplied** until this commit, which is the same defect class the ladder kept finding.

**Gate:** content, skill-listing and path-rot gates — 371 passed. Docs-only; other tiers not run and not claimed.

ZacxDev:
## Round 8 — the decisions rewrite audited: 7 🟡, 5 🟢, no 🔴

This round was **not** a continuation of the prose ladder (stopped at round 7). It covers `be36fb5d..b9848f9d` — two commits, both unaudited: round 7's fixes, and the eight-decision unwind.

### 🔴 The worst finding is a design defect in a prescription

§5.1 told an implementer to build decision 15's detector as a **cross-host diff** of hook-event sets, citing `drift-check.sh` rc 22 as precedent. Re-verified against the source: **rc 22 is not a cross-host comparison**, and the script says so in capitals —

> `🔴 THIS IS NOT A CROSS-HOST COMPARISON. Both hosts can agree perfectly and both be wrong; the reference is the ledger, not the other machine.`

rc 15 *is* the cross-host arm — the one that same paragraph already calls too narrow. So the cited precedent argues **against** the prescribed design, and the design inherits the exact failure decision 15 exists to end: drop `SessionEnd` from both hosts and the sets agree while every session silently attaches only the floor. Now prescribes a git-tracked ledger of required hook events, each host checked against it.

### Decision 9 created an N-way object fan-out; the integrity story didn't follow

Decision 12's consequence and §5.4 both said checking a pointer costs **one HEAD request** — true for one object per session, false at ~7.8. Worse: **a member that never landed is undetectable.** A `sha256` proves each object that *exists* is intact and says nothing about one that does not, because nothing records how many objects a session should have. That is precisely the gap decision 11 closes for entries, left open for the object fan-out decision 9 created — and the reasoning that named it was deleted along with §9.7 in the same commit and never relocated.

### The control decision 14 made load-bearing had no control

§8 is the list of "ways this silently does nothing", and it had nothing for the report-only scan — the one mechanism now standing between a leak and nobody knowing. Added as control 8: plant a synthetic credential, watch the count move **0 → 1**, report the pair.

### The rest

- *"the only remaining control in the chain"* contradicted §6's own heading (*"the boundary is the only control"*). Narrowed to the claim that holds — the only response available **once bytes are already in the bucket** — which leaves the conclusion intact.
- *"the one decision that diverged from the recommendation"* named a recommendation this document never makes; it was made in the interview. Same shape the document itself retracted for "against the letter of decision 6".
- **"decision 11" named two different decisions inside §5.3** — this document's new one, and `plan-cairn-integration.md`'s sharing gate. Qualified.
- **Two open sub-questions were closed with no decision covering them**: who owns *writing* the front-matter field, and whether `project_of` is shared. Both were halves of questions whose other halves became decisions 12 and 13. Restored to §10 so they read as open rather than absent.
- Three headings still asserted their question was OPEN, 35–45 lines above their own ✅ DECIDED markers.
- 🟢 §10 listed three things decisions 1-16 *do* settle, two paragraphs after §9 said §2 is the single place they live. Rewritten to carry only what is genuinely undecided — plus **who reads the fan-out set**, without which decision 11 is inert.
- 🟢 Control 3 said the same thing twice, and "NO GROWTH" is wrong when a subagent is dispatched between two handoff runs; the property that holds is per-key.
- 🟢 The stated reason decision 12 is safe was **false against the live parser**: `- clickup:868abc123` parses to a clean list member because block-item promotion was *fixed* — which is what the docstring quoted 20 lines earlier records. Conclusion right, reason a hazard that no longer exists.
- 🟢 "Fourteen places say row N" was wrong when written (16 then, 21 now). Number **deleted**, not corrected — the document's own rule.
- 🟢 The un-rotatable class now says the response is *nothing* and the exposure permanent, rather than implying a response exists.

```audit-claims round=8 audited=b9848f9d..2398fcba
1. §5.1's detector re-specified as a git-tracked ledger checked per host, with the rc 22 misattribution recorded and the cross-host blind spot named.
2. §5.4 and decision 12's consequence corrected for the N-way object fan-out; the expected OBJECT set is now recorded alongside the entry set, with the deleted reasoning restored.
3. §8 gained control 8 for the report-only scan: plant a credential, watch 0 to 1, report the pair.
4. "the only remaining control in the chain" narrowed to the claim that holds, resolving the contradiction with §6's heading.
5. The decision-14 divergence framing now names where the recommendation was made instead of implying it is in this document.
6. §5.3's second "decision 11" qualified as plan-cairn-integration.md's.
7. The field-writer owner and project_of restored to §10 as genuinely open, with a note that they were dropped when their sibling halves became decisions.
8. Three stranded OPEN headings and lead sentences corrected to match their own DECIDED markers.
9. §10 rewritten: settled constraints removed, and "who READS the fan-out set" added because recording is not detecting.
10. Control 3 de-duplicated and its property restated per-key, since the set legitimately grows when a subagent is dispatched mid-session.
11. Decision 12's safety rationale corrected against the live parser; the colon claim was a fixed hazard.
12. The "Fourteen places" self-count deleted rather than updated, and the un-rotatable class stated as having no response.
```

**Ledger:** round 8 · payload lines changed by round 8's fixes: **101 insertions / 47 deletions** (the range it audited was 255 lines across 2 commits). Rounds 1-7: 224/57, 67/26, 59/19, 36/11, 29/15, 23/12, 44/20.

**Gate:** content, skill-listing, path-rot and rules-size gates — 376 passed, base `2398fcba`, dev-host tier. Docs-only; sandbox tier not run and not claimed.

⚠ **What this round could not check, stated because it matters:** the auditor verified §2 against the *body*, never against what you actually answered. If §2 mis-records one of your eight answers, no round can see it — only you can.

ZacxDev:
## Round 9 — the new prescriptions attacked, not their wording

3 🟡, 1 🟢, no 🔴. Two are defects in prescriptions round 8 newly wrote — which is what this round was dispatched to find.

### 🔴 The ledger fix carried rc 22's SHAPE but not the property that makes it work

Round 8 correctly replaced the cross-host detector with a ledger-based one. But rc 22's ledger is pinned **two-way** by `test_skill_tiers.py`, whose docstring calls that *"THE ONE PROPERTY THAT MAKES IT SCALE — without it the mechanism silently stops covering the tree the moment a skill is added."*

That pin is possible because **skills exist as files**, so the expected set can be generated. **Required hook events have no generator** — verified: nothing in the tree enumerates them, and the only hook-event list (`register-nudge-hook.py`'s `MATCHER_EVENTS`) says which events *accept a matcher*, not which must be *registered*. So no two-way pin can be written for this ledger — and round 8 named the residual for the design it rejected while naming none for the one it prescribed.

**The failure that survived:** seed the ledger from a host — the laptop, the host being fixed, or a fresh one — and `SessionEnd` is omitted. Both hosts then match the ledger, no rc fires, and the run prints a clean tiers-style line forever. The same failure decision 15 exists to end, one level up.

Fixed by stating the three things the absent gate cannot do for you: **author each entry from the requirement, never seed from a host**; **state why each event is required**, since that review is all the missing generator leaves; and **treat a missing or empty ledger as COULD NOT MEASURE, never clean** — which `drift-check.sh` already says for rc 22 (*"an empty expectation would make every host look compliant, in silence"*) and defends with a test this arm cannot have.

### 🟡 Control 8 was satisfiable with 64% of the bytes unscanned

It said "plant a credential in **a** fixture transcript" and "run **the scan**". Under decision 9 the shipped unit is ~7.8 files and `agent-*` files hold **64% of the bytes**; §3 already shows the correct-looking implementation ships parents only. Plant in the parent and the count moves 0 → 1 while the majority of bytes are never scanned — **a control that passes on the case that was never in doubt.**

It also never asserted the **shipper invokes the scanner**, so a scanner unit-tested against a fixture and never called on the real path reports `0 findings` forever — verbatim the failure the control's own closing sentence names. Now: run the shipper, plant in a subagent transcript.

### The other two

- 🟡 Round 8's N-way cost correction reached §2 and §5.4 and **missed §5.3 — where decision 12 is declared** — and the fix had rewritten the sentence two lines above it.
- 🟢 Control 1 spelled "0 to 1", the literal assertion control 3 forbids two bullets later; under decision 9 `== 1` fails a correct implementation.

```audit-claims round=9 audited=2398fcba..fb1b55fc
1. §5.1 now states the three constraints the absent generator leaves: author from the requirement not from a host, say why each event is required, and treat a missing or empty ledger as COULD NOT MEASURE.
2. The asymmetry with rc 22 is stated explicitly — its two-way pin needs a generator, skills have one as files, required hook events do not.
3. Control 8 rewritten to run the SHIPPER rather than the scanner, and to plant the credential in a subagent transcript rather than the parent.
4. §5.3's third copy of the pointer-check cost corrected to the N-way figure.
5. Control 1 no longer spells the "== 1" assertion control 3 forbids.
6. §10's owed-work bullet now carries the ledger's three gate constraints rather than just naming the detector.
```

**Ledger:** round 9 · payload lines changed by round 9's fixes: **48 insertions / 11 deletions**. Rounds 1-8: 224/57, 67/26, 59/19, 36/11, 29/15, 23/12, 44/20, 101/47.

**Gate:** content, skill-listing, path-rot and rules-size gates — 376 passed, base `fb1b55fc`, dev-host tier. Docs-only; sandbox tier not run and not claimed.

⚠ Round 9 could not run any gate on this PR — its worktree did not contain the file — and said so rather than reporting a run against a tree without the payload as coverage. The 376 above is mine, on the merged content.

PR #1328, opened 2026-09-05
perf(mention-open): delete the GitHub-wide search; offer the picker instead of refusing

ZacxDev:
## The problem

Clicking `dashboard#12` in Alacritty took **3.1–3.6s** (measured; ~10s through the real hint path, at 6% CPU). All of it was network wait in `gh_api_repo_search`, which searched **all of GitHub** for repos named `dashboard` and returned `vzhong/dashboard`, `yorkie-team/dashboard`, `zce/dashboard` — strangers' repositories the operator would never pick.

It was not slow *and* useful. It was slow *because* it searched a corpus that structurally cannot hold the answer.

The correct universe was local the whole time. `~/.config/mention-open/known_repos.json` (369 entries, built by `regen-known-repos.py` from `gh api user/repos --paginate` — owner + collaborator + organization_member — plus local checkouts) **is** "the repos the operator contributes to", and reading it costs 0.3ms.

So the search is **deleted**, not deferred or cached.

Second change, same root cause — a refusal that reads like a failure. `#282828` produced the toast `no mention in the clicked text`. That was the `_NUM = \d{1,5}` guard working correctly, but a guard that reads as a bug is one the next maintainer deletes. Every refusal path that *can* offer a choice now opens the fuzzy picker over the local universe.

A toast survives in exactly one case: **the picker cannot be shown** — the mapping absent, unreadable, or holding no usable rows, or `--print`/`--no-discovery` barring it by contract. It still names *which*.

## Latency

Interleaved A/B against the pre-change file (A,B,A,B… so a load spike hits both arms), `rofi`/`xdg-open`/`notify-send` stubbed, **min/median of 9 runs each**. `rofi` exits 1, so the timing is click → picker on screen.

| case | before | after | |
|---|---|---|---|
| `click dashboard#12` | 3141 / 3572 ms | **69 / 79 ms** | ~45× |
| `click #370` (bare id) | 169 / 174 ms | **69 / 81 ms** | ~2.2× |
| `click owner/repo#12` | 26 / 29 ms | **26 / 28 ms** | unchanged |
| `click #282828` | 29 / 32 ms, rc 1 | **68 / 71 ms, rc 0** | toast → picker |

`#282828` got **slower in wall time** because it now does work instead of refusing. That is the point of the change, not a regression.

## A second finding the brief did not have

`discover_repos` ran **100 serial `git remote get-url` children — 149–184 ms**, which is 60% of a 100 ms budget on its own and would have missed the target even with the network call gone. It is now a 16-thread fan-out (**26 ms**), semantically identical.

`git remote get-url` is **kept** rather than parsing `.git/config` directly: a hand-rolled reader measured 8 ms and agreed with git on all 100 checkouts here (56 of them linked worktrees, 0 disagreements) — but it would have to re-implement `commondir` resolution, submodule gitdirs, `include`/`includeIf` and `url.<base>.insteadOf` to *keep* agreeing. 26 ms is already far inside budget; reimplementing a slice of git to save 18 ms buys a class of divergence bugs for nothing.

`concurrent.futures` is imported **lazily**. At the top of the file it cost ~5 ms of interpreter startup on *every* click, including `owner/repo#N` and bare `#N`, which never reach the fan-out: 29.6 → 35.1 ms for an explicit owner, back to 25.5 ms once moved. Caught by the interleaved measurement; a before/after run done separately had blamed it on load.

## Preserved — pinned, not asserted

- **No owner is ever guessed.** An unresolvable `repo#N` still yields no URL.
- **Dismissing the picker opens NOTHING.**
- **A six-digit number is OFFERED, never AUTO-OPENED.** `_NUM = \d{1,5}` is untouched and still decides what may be *opened*. The new `offer_number` is a strictly weaker thing: its output can only ever become a *universe row*, and `offered_universe` unconditionally excludes universe rows from the single-candidate auto-open shortcut. The distinction is spelled out in both the code and the comments.
- **No universe row reaches a log, spool, stdout, stderr, a notify body, or a fixture.** `refuse()` does not even have `discovered` in scope, which is a structural improvement over the previous inline refusal. No test reads the real mapping.

## `--print`, decided and stated

`--print` **never offers the picker** and now exits 1 with a named reason where it used to exit 0 with one URL per namesake.

That is a correction, not a loss: those namesakes were strangers' repos, so the old exit 0 handed a consumer a plausible URL into somebody else's issue tracker. Real ambiguity — several candidates for a reference the text or the host actually attributes — still prints, unchanged.

`--no-discovery` is unchanged: "resolve only what the text itself carries".

## New guards, each mutation-verified

- `test_the_resolution_path_spawns_ONLY_these_local_commands` — a **ledger** of `(argv[0], argv[1])` that fails when the set grows *or* shrinks, so `git ls-remote` cannot take the deleted call's place. Its assertion carries its own positive control (`test_the_no_network_assertion_can_actually_FIRE`), which watches it reject both a re-added `gh` call and an empty set.
- `test_the_module_no_longer_carries_a_repository_SEARCH_at_all` — reads the **AST**, not the file text, so the docstring can keep *naming* `gh api search/repositories` in the paragraph recording why it was deleted. A substring check would have forced that note to be softened to make the test pass.
- `test_the_handlers_OFFER_bound_matches_the_hints_own` — the seam between the Alacritty hint regex's `{1,6}` and `_OFFER_NUM_RE`. Two files, two languages, two suites; each hermetically green while they disagree, silently in both directions.

## Mutation battery

8 new rows (K17–K24) in the committed `scripts/tests/mutation_battery_mentions.py`.

```
positive control P1: KILLED — the battery can observe
25/25 killed for the stated reason; problems: none
restore: OK
```

`test_alacritty_hints.py` joins its `SUITES` — K20 scored `KILLED-WRONG-REASON` until it did, because the battery was not collecting the guard's own suite. Three further rows scored `KILLED-WRONG-REASON` on the first run because the phrase sat on an assertion that never evaluated; the messages were moved to the assertion that actually fires.

### DIFF
--- a/claudedocs/handoff-mention-system-repos.md
+++ b/claudedocs/handoff-mention-system-repos.md
@@ -53,17 +53,27 @@ diagnosis in progress.)
 1. **Exercise the real Alacritty click path** — STILL the one thing never done end-to-end,
    and now more valuable than before because #1313 changed it. Plain left-click (the hint is
    `mouse.enabled = true` with no mods; `Ctrl+Shift+M` is the keyboard route):
-   `talos-infra#1065` (opens), `dashboard#12` (rofi picker), `#282828` (underlines, opens
-   nothing). 🔴 **The widest-blast-radius change to check deliberately:** a bare `#N` clicked
+   `talos-infra#1065` (opens), `dashboard#12` (rofi picker, now with an explanatory line
+   above the list naming the clicked text, the row count and the mapping's age), `#282828`
+   (underlines; since #1328 it opens NO picker and shows a named toast saying it looks like a
+   colour literal). 🔴 **The widest-blast-radius change to check deliberately:** a bare `#N` clicked
    in a pane with NO resolvable repo now shows a ~370-row fuzzy picker with the clawgate task
    first, where it previously opened the clawgate task directly. Type to narrow. If that is
    wrong in practice it is a one-line ordering change.
    🔴 NEEDS A HUMAN — clicking raises windows, a `pkill`-class action for an agent.
    forcing: none
-2. **Add a staleness signal for `~/.config/mention-open/known_repos.json`** (devrc;
-   `scripts/regen-known-repos.py` + a test, or a systemd-user timer in `nix/home.nix`).
-   Re-measured 2026-09-04: still zero timers reference it (`present-regen.timer` belongs to
-   the `present` skill and is unrelated). Degrades to the API fallback rather than breaking.
+2. ~~**Add a staleness signal for `~/.config/mention-open/known_repos.json`**~~ — **DONE** in
+   PR #1328. `mention-open.py` now measures the mapping's age (`mapping_age_days`) and
+   surfaces it in BOTH places the operator can see it: the note above the fuzzy picker
+   (`universe_note`) and the refusal body (`staleness_note`, past `STALE_MAPPING_DAYS = 7`).
+   🔴 **A SIGNAL, NOT A REPAIR, AND THAT WAS THE CHOICE.** A systemd-user timer was the other
+   option and was rejected: `regen-known-repos.py` shells out to `gh api user/repos` and
+   REFUSES below its 25-repo floor with exit 3, so a host without `gh auth` would take a
+   failing unit plus a failure toast on every fire — a permanently-red timer, which trains
+   everyone to ignore it. The signal instead fires at the exact moment the staleness bites:
+   the click that did not resolve. Re-measured 2026-09-05: still zero timers reference the
+   generator. Pinned at four ages (0, 6, 8, 90 days) by `test_the_mapping_age_is_measured_at_
+   FOUR_points`, and mutation-verified (K32).
    forcing: none
 3. **Commit the workbench's `save_to_clipboard` WIP** in
    `nix/programs/alacritty/default.nix`. Now the ONLY substantive difference between the two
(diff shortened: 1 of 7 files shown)


ZacxDev:

```audit-claims round=1 audited=c1075a5d..a3bcd537
1. The fan-out is extracted as `_fan_out()` and covered by a FOUR-checkout fixture with pairwise-distinct owners and repo halves; the neighbour-transposition and dropped-`if full:` mutants are claimed to die on their own assertions (K25, K26).
2. The no-network ledger key is now the verb path (argv[0], argv[1], plus argv[2] when a bare word), claimed to reject `git remote update`, `git fetch --all`, `git ls-remote` and `curl` while the real argv still passes (K27).
3. `_fan_out` failure degrades to a serial fan-out with the same pairing, and `guarded_main` is the `__main__` entry point (K33, K34).
4. `notify-send` is invoked with `--` before SUMMARY/BODY; claimed reproduced against the real binary, without which a refusal showed no toast at all (K28).
5. `--print` names both causes — the flag and an absent mapping — and `--no-discovery` is claimed still not to blame a mapping it never read (K35).
6. The picker's diagnosis is carried above the list as rofi `-mesg` (clicked text, row count, mapping generation date/age, remedy), Pango-escaped; dismissal still opens nothing and says nothing (K31).
7. Ranked item 2 is closed as a staleness SIGNAL rather than a timer, justified by `regen-known-repos.py` exiting 3 below its 25-repo floor; `mapping_age_days()` feeds the picker note and a refusal note past 7 days, measured at four ages plus absent (K32).
8. Six-digit numbers get a named toast and skip discovery entirely; the picker is retained for every other shape, pinned at five digits (K29, K30).
9. `pkgs.gh` is removed and the wrapper's PATH is pinned against the handler's AST, failing in both directions; K36 mutates the nix file.
10. The battery's verdict is a pure `classify()` with a NOT-OBSERVED arm; a syntax-error mutant that previously scored SURVIVED is claimed to land in `problems`.
11. Two tests that were reading the operator's real `known_repos.json` now use a tmp mapping.
```

PR #1332, opened 2026-09-06
feat(resume): move the corpus query into step 4 — it fired 1/14 where it was

ZacxDev:
Rank 1 of `handoff-handoff-search-index`. The measurement that motivated it merged as #1329.

## What changed
The handoff section index had exactly one consumer: a conditional buried in step 3's prose (*"before working any open item…"*). It is now the **second command of step 4**, keyed on the handoff's **topic** rather than an open item — every resume has a topic, and keying on an item *is* the conditional being removed.

## Why — measured, not argued
Over the 34 h after #1295 merged, **1 of 14 `/resume` runs across both hosts** invoked it, and that one was reading `origin/main` off the staleness alarm rather than firing the step. Yield: 0.

The non-firing was **not** correct restraint — all six workbench non-firing runs ran `claim-work`, five made edits, and five of six resumed a doc with an `## Open investigations` section.

**Discriminator — placement, not the tool:** step 3's *sibling* check (`git log --since`; same trigger, same block, an ordinary command) fired **0/6**, while step 4 (`cairn recall`; numbered, unconditional, fenced) fired **5/6**.

## Not a new numbered step, deliberately
~15 prose cross-references across skills and scripts pin "step 4" (cairn) and "step 6" (claim-work), none machine-checked. A renumber would falsify all of them.

## The guard is structural, not spelled
A grep for "unconditional" passes on a body that says the word and still gates the command. The new test asserts **which numbered step the command physically sits in**. Watched to fail two ways: red at `origin/main` (count arm), and an isolated mutant — query moved back into step 3, wording untouched — fires the step-identity arm with its own message.

## Residual, stated in both the skill and the test
The 5/6 was measured with **one** command in that step. Co-locating a second is a prediction, not a measurement. Re-measure with check 1 of the handoff's "How to verify".

## Gate
Both tiers on the merged tree at base `f0b9c474`:
- `nix build .#checks.x86_64-linux.pytests` — collected=21837 passed=21834 failed=0 (floor 18610)
- `nix build .#checks.x86_64-linux.nodetests` — suites=5 tests=1449 pass=1449 fail=0 (floor 1367)

Run one at a time. Base re-confirmed unmoved after the run.

https://claude.ai/code/session_01G7eZPAdX7nAeamShBnYNg2


ZacxDev:
## Audit round 1 — six 🟡, no 🔴, all fixed in `e8a273c8`

Both sandbox tiers green on the rebased tree (base `e8143452`): pytests 21868/21865/0 failed (floor 18610); nodetests 1449/1449/0 (floor 1367).

```audit-claims round=1 audited=87989efb..e8a273c8
1. F1 — test renamed off the UNCONDITIONAL claim it did not check, and widened from a shared step DIGIT to a shared FENCE; mutant M4 (second list item re-labelled `4.`, gated, own fence) now dies with its own message. M2 and M3 are claimed STILL UNCOVERED and named as such in the docstring, not fixed.
2. F2 — "a few hundred bytes" replaced with re-measured figures: 3941 B and 3999 B for two real --limit 3 queries, 1251 B no-match floor.
3. F3 — step 3 reduced to a pointer carrying no action; the item-level re-run instruction and the "Ruled out: bullets" purpose statement both moved into step 4.
4. F4 — step 4 now states that handoff_search is corpus-wide over four repos (two of them client repos) while cairn recall is repo-scoped, and tells the reader to check each hit repo before reuse. Scope deliberately NOT narrowed to --repo.
5. F5 — step 5 now carries both surfaces and their distinct banner labels (`from index` vs `from handoff docs`), both verified by running the tools.
6. F6 — "that doc's How to verify" replaced with the named path claudedocs/handoff-handoff-search-index.md, with an explicit "not the handoff you read in step 3".
7. F7 — denominator accounting stated (8 = 1 fired + 6 analysed + 1 measuring session excluded as the instrument), with an as-of timestamp, in the skill, the test docstring and the handoff doc.
8. Nit — docstring claiming the recall posture sits "four steps earlier" corrected; it was wrong both before and after the move.
```

Round 1 produced findings that needed fixing, so a delta round follows.

ZacxDev:
## Audit round 2 — all 8 round-1 claims verified fixed; 2 🟡 + 1 🟢 found and fixed in `00e803a2`

Round 2 re-ran the mutation matrix at both guard revisions and re-derived the adoption count live rather than accepting the self-reports. Both sandbox tiers green on the rebased tree (base `10d437c9`): pytests 21871/21868/0 failed (floor 18610); nodetests 1449/1449/0 (floor 1367).

```audit-claims round=2 audited=e8a273c8..00e803a2
1. 🟡 doc status — claudedocs/handoff-handoff-search-index.md said "No fix is built; awaits operator direction" and ranked next-step 1 still described THIS PR as pending. Both survivability mitigations (held claim, open PR in step 6 sweep) end at merge, so a post-merge /resume would have re-implemented a merged change. State now + rank 1 rewritten; rank 1 is now RE-MEASURE ADOPTION.
2. 🔴-in-that-item — the same item asserted "SKILL.md is byte-capped, so promoting text means evicting text". FALSE; the caps cover browser/handoff/prune-skill/RULES.md, not resume. Corrected in place and LABELLED as a correction rather than silently deleted.
3. 🟡 M5 — a fifth mutant round 1 did not anticipate: command left in the fence but COMMENTED OUT. Every check here is a substring test, so a commented line satisfied the string pin, the step check AND the fence check over dead wiring. Now killed by requiring the fence line to be live. Docstring no longer presents the mutant list as closed.
4. 🟢 scope warning restated handoff_index.REPO_ENV_HANDLES membership instead of deriving it. Now pinned against the module. The pin covers the LABELS only; the "two of which are client repos" count is a human judgement and is stated as unpinnable.
5. Mutants re-run under PYTHONDONTWRITEBYTECODE=1 — control 8 passed; M5 killed with its own message; M6 (drop a handle from the warning) killed by the derived pin naming DATAPACKET. M2 and M3 STILL SURVIVE and remain named as uncovered.
```

Round 2 produced findings that needed fixing, so round 3 follows.

ZacxDev:
## Audit round 3 — three findings, all fixed in `0d0f0b6f`. **Ladder STOPS here, on the attribution gate.**

Round 3 caught a real hole in the guard round 2 added — see claim 1; it is the sharpest finding of the whole ladder.

```audit-claims round=3 audited=00e803a2..0d0f0b6f
1. 🟡 the round-2 derived scope pin had a STRUCTURALLY UNREACHABLE arm. It searched the whole wiring block, and EXPECTED_COMMAND contains `~/workspace/devrc/...`, so "devrc" in block was guaranteed true on every green run; round 2 swept with DATAPACKET, a fixture that could only die, and read the kill as proof. Narrowing to the warning PARAGRAPH was measured INSUFFICIENT (the paragraph itself says "a devrc-topic query"). Fixed by matching BACKTICKED tokens. Now measured: control 8 passed; M6-devrc -> 1 failed naming [DEVRC]; M7 (handle dropped from the list but mentioned in an unrelated sentence, which passed before) -> 1 failed naming [CIVITAI]; M5 and M1 still die. All four arms reachable; two of four were not.
2. 🟡 the `Claim:` line asserted a release that had not happened and dropped the command that performs it. Restored as STILL HELD with the explicit `claim-work --release`, plus why an unreleased global claim blocks the next rank 1 under a different identity.
3. 🟡 a THIRD site still told the reader to build the fix — Open investigations -> Next probe. Rounds 1-2 fixed State now and rank 1 and missed this one, so the doc contradicted itself at the exact section /resume step 3 reads. Reworded.
4. 🟢 the correction paragraph`s own number was wrong: "~1.9 KB" was one commit`s delta. Measured +3,688 B (41,852 -> 45,540 via git cat-file -s) and labelled as a correction of a correction.
5. 🟢 rank 1 now says to raise check 1`s CUT to #1332`s merge time; left at #1295`s, the 14 pre-fix runs stay in the denominator and a successful fix reads as ~10/24, i.e. as a failure.
```

### Why the ladder stops here — the attribution gate, measured

Payload for this PR is `claude/skills/resume/SKILL.md`. Measured with the prescribed command (`git log --numstat --format= --remerge-diff <audited>..HEAD --not <base>`, rc 0, non-empty range):

| round | fixes touched | payload lines |
|---|---|---|
| round 2 | test + handoff doc | **0** |
| round 3 | test + handoff doc | **0** |

**Two consecutive rounds whose fixes changed zero payload lines** — the skill body has been untouched since round 1. The ladder is auditing the guard and the notes it wrote, not the change. That is the documented stop condition, and it is a measurement, not a judgement call or a round cap.

Nothing is being left unfixed: all three round-3 findings are fixed above. The two known residuals (mutants M2 and M3 — a gating sentence above the fence, a conditional comment inside it) are **not** silently dropped: they are named as uncovered in the test docstring, which also states the mutant list is not closed.

### Gate — merged tree, base `f58d2df0`, run one at a time
- `pytests` — collected=21903 passed=21900 failed=0 (floor 18672)
- `nodetests` — suites=5 tests=1449 pass=1449 fail=0 (floor 1367)

Base re-confirmed unmoved after the run. Re-gated because `f58d2df0` changed `scripts/run-tests.sh`, which owns the per-target floor table — the floor moved 18610 -> 18672, so that re-run was substantive.

PR #1334, opened 2026-09-06
feat(tmux-reply-agent): ARM the host half — enableTmuxReplyAgent = true, and every guard that pinned the disabled state flips with it

ZacxDev:
Rank 32 of `claudedocs/handoff-tmux-webapp.md`. The server half of the reply loop
has been merged, deployed and armed as **0.8.26** since 2026-09-06
(`ZacxDev/homelab-infra#711`/`#715`/`#712` + `innovation-upstream/devrc#1324`).
The host half was still wanted by nothing, so a reply typed in the web UI queued
and was never executed. Measured before this change:
`systemctl --user is-active tmux-reply-agent` → `inactive`.

The functional change is one line. Everything else is here because that one line
makes six other statements in the tree false.

## 🔴 What this arms, stated plainly

`tmux send-keys` followed by Enter is **arbitrary command execution as the
operator, on both hosts**, driven by a route on a LAN NodePort with **no human
auth**, gated solely by `CLAWGATE_TERMINAL_TOKEN`.

That is an operator decision taken with the blast radius stated (handoff rank 29
records four such decisions, including "free text into ANY pane, unrestricted" —
offered and declined the narrower scoping). The deliberate part was that
**building and arming are separate acts**, so the write path could be merged,
deployed, read and audited across seven rounds while it could execute nothing.
That ordering has happened. This is the arming half, and it is one line to
reverse.

## Measured, with a paired control

| tree | flag | `Install.WantedBy` |
|---|---|---|
| base clone | `false` | `[]` |
| this branch | `true` | `["default.target"]` |

Both read from `nix eval` on the real flake — the flag is demonstrably what
drives the install, in **both** directions, not one reading asserted twice.

## The guards flip rather than being deleted

`test_the_agent_unit_SHIPS_DISABLED` pinned the flag `false`; it is now
`test_the_agent_unit_IS_ARMED` and pins it `true`. **The symmetry is the point.**
What that guard has always enforced is that the armed state is a deliberate,
reviewed edit in both directions: while it read `false`, arming meant editing the
test; now it reads `true`, disarming means editing the test. An accidental revert
would stop every queued reply from executing while the server kept accepting
writes and the UI kept looking healthy — a silent reopening of the loop this rank
closes.

The condition assertion (`cond == "enableTmuxReplyAgent"`, the bare flag) did
**not** change meaning and is the more important half: it keeps the flag the only
thing deciding whether the agent runs. `(enableTmuxReplyAgent || isNixOS)` would
want the unit regardless of the switch — and it mentions the flag's name, so it
walks past any word-pinning guard.

### The positive control — and the two corrections it took

This section has been WRONG TWICE, and the history is the useful part.

**Draft 1** claimed the control, left mutating `false → true`, would have gone
"green and blind". **Audit round 1 refuted it**: `assert flipped != src` already
existed and `enableTmuxReplyAgent = false` occurred zero times in the armed file,
so the stale direction would have failed loudly.

**Draft 2 retracted that — and was falsified by its own fix.** The retraction
commit also wrote a disarm example into `nix/home.nix`'s comments:

```
  #     enableTmuxReplyAgent = false;   # then merge + scripts/ship.sh, THEN:
```

That prefix ends in spaces, so the two-space-prefixed literal matches **inside the
comment**. **Audit round 2 measured it**: reverting the control to its pre-flip
direction on that tree gives **`1 passed`** — the vacuous green draft 2 had just
finished explaining could not occur. A fix for a claim created the hazard the
claim denied.

**Now:** the mutation is ANCHORED — `re.subn(r"^  enableTmuxReplyAgent = true;$",
flags=re.M)` with the substitution count asserted `== 1`, the same shape
`_wanted_by_expr` always used, which is why *it* was never fooled by the comment.
The disarm example stays: it is the only written rollback for a surface that
executes commands. Fix the test, not the sentence.

## Mutation battery — each mutant killed by ITS OWN assertion

Controls green either side; `PYTHONDONTWRITEBYTECODE=1`; tree verified clean
after restore.

| mutant | result | the assertion text it produced |
|---|---|---|
| control (unmutated) | **105 passed** | — |
| flag `true` → `false` | **3 failed** | `the terminal-write agent's master switch is not true` · ``enableTmuxReplyAgent is `false` `` · `the flag declaration this control mutates was not found verbatim` |
| cond → `(!enableTmuxReplyAgent)` | **2 failed** | ``the WantedBy condition is `(!enableTmuxReplyAgent)`, not the bare flag`` |
| cond → `(enableTmuxReplyAgent \|\| true)` | **2 failed** | ``the WantedBy condition is `(enableTmuxReplyAgent \|\| true)`, not the bare flag`` |
| post-restore control | **105 passed** | — |

A green run was not treated as evidence until each guard had been watched to go
red for its own reason.

⚠ **This table first reported `3 failed` for mutant 3. That was wrong** — the two
condition mutants change only `cond`, so they cannot differ. The third failure in
that run was an UNRELATED intermittent test (`test_the_EXACT_session_still_works`),
which later failed again on a completely unmutated tree: 2 occurrences in 10 runs.
Independently re-measured as `2` by audit rounds 1 and 2 and by a re-run here. The
intermittent test is a pre-existing defect in `#1324`'s `open_window` read-back —
`out.strip()` eats an empty trailing `#{pane_current_path}`, collapsing 3 fields to
2 — recorded under "Not closed by this PR", not fixed here.

## Gate — leg by leg, on the merged tree

🔴 **Re-run from scratch after the base moved TWICE mid-review** (`f0b9c474 -> c5a445d8 -> 88f1bda4`).
The numbers below are against **`c25ef63c`**, merged into this branch, with
`origin/main` re-confirmed as an ancestor of HEAD afterwards — so the branch tree
IS the merged tree. The earlier green runs at `f0b9c474` and `88f1bda4` are superseded and are NOT
quoted as evidence for this head — each was a real four-leg green, against a base
that has since moved and a tree without the later audit fixes. "The gate passed"
is true of one run, one tier, one base; it is not a property of the change.

| tier | leg | verdict |
|---|---|---|
| 1 (dev host) | `run-tests.sh` (pytest) | ✅ **PASS** — `RESULT: PASS (exit=0)`, **21,867 / 21,870**, 3 skipped, **0 failed**, floor 18,610 |
| 1 (dev host) | `run-node-tests.sh` | ✅ **PASS** — 5 suites, 41 files, **1,449 / 1,449**, floor 1,367 |
| 2 (nix sandbox) | `checks.x86_64-linux.pytests` | ✅ **PASS** — `RESULT: PASS (exit=0)`, **21,867 / 21,870**, 0 failed |
| 2 (nix sandbox) | `checks.x86_64-linux.nodetests` | ✅ **PASS** — `RESULT: PASS (exit=0)`, **1,449 / 1,449** |

Each verdict is the runner's own `RESULT:` line, not a shell exit code. The two
sandbox derivations were built **one at a time**; `database is locked` /
`SQLite database is busy` appear **0** times in either log, despite other sessions
running the same check derivation concurrently throughout.

🔴 **An intermediate run of this gate was RED, and it was this PR's own fault.**
Adding `systemctl --user stop tmux-reply-agent` to the agent docstring (the round-1
rollback fix) made `test_no_real_launchers.py`'s text scan see a new file reaching
an acknowledged binary — `scripts/tests` 12,740 passed / **1 failed**. Fixed by
ACKNOWLEDGING the prose mention with grep evidence of no call site, **not** by
rewording the docstring: that ledger's own `session-write` entry says "Deleting the
word to get green would delete the guarantee", and here the word is the only
written rollback for a surface whose payload is arbitrary command execution. That
red run is the acknowledgement's positive control — the guard was watched failing
for exactly this reason before it was satisfied.

⚠ **The first tier-1 attempt of all exited `3` — a PRECONDITION failure, not a red
suite** (`gate.sh` from a plain shell has no toolchain on PATH and prints the
`nix develop` form to re-run). In that same output `| tail` printed a reassuring
`GATE_RC=0` directly beneath the runner's own `GATE: RESULT=FAIL exit=1`; the
status belonged to `tail`.

## Comments are claims too

The five sentences this line falsified are corrected, not left:

- `nix/home.nix` — "🔴 IT SHIPS **false**, AND THAT IS THE POINT OF THE WHOLE SHAPE"
- `nix/home.nix` — "🔴 SHIPPED DISABLED … The unit below EXISTS on both hosts and is wired into nothing"
- `nix/home.nix` — "🔴 SHIPPED DISABLED — see enableTmuxReplyAgent" (on `Install`)
- `scripts/tmux-reply-agent` — "both are deliberately off in the change that introduced them"
- `scripts/tmux-reply-agent` — the no-token exit describing itself as "the state this ships in … neither has been done"

The last now says what reaching it **means** on an armed host: the host's copy of
the credential has gone (`~/.claude/clawgate.env` rewritten, or a unit
environment that cannot read it), not that the feature is unbuilt.

## Not closed by this PR

- The rank-32 closing condition is `systemctl --user is-active tmux-reply-agent`
  reading `active` on **both** hosts — that needs `scripts/ship.sh` after merge,
  reading every per-host line rather than the final verdict.
- **Rank 33** — a real reply typed in the UI landing in a real pane, exercising
  the `=` session-target prefix specifically — is untouched, and is the closing
  condition for the whole effort. `devrc/tests` in the subsystem index records
  that the `=` fix and the `isdir` check each *survived their first mutation* and
  have never run against real tmux outside a test.
- **Rank 34** — the two residuals the audit recorded as ACCEPTED (the browser
  tier authenticates nobody; the host agent posts an execution-grade token over
  plain HTTP to the LAN NodePort every ~5 s) are unchanged and now live.

https://claude.ai/code/session_014bDkLZ3BgR1c5wKKqc1G5G

### DIFF
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -149,9 +149,24 @@ let
   # decided is that building it and ARMING it are separate acts, so the write path
   # was merged, deployed, read and audited across seven rounds while this flag was
   # still false and it could execute nothing. It shipped `false` for exactly that
-  # reason — that ordering is history now, not a live guard, and the SHAPE it left
-  # behind is what still matters: setting this back to `false` is a one-line,
-  # reviewable disarm of the host half, independent of the pod's secret.
+  # reason — that ordering is history now, not a live guard.
+  #
+  # 🔴 DISARMING IS TWO STEPS, NOT ONE, AND THE SECOND IS NOT OPTIONAL. Setting
+  # this back to `false` and shipping does NOT stop a RUNNING agent. The unit
+  # DEFINITION is emitted unconditionally, so the flag only removes the
+  # `[Install]` block — home-manager's `sd-switch` then sees a CHANGED unit and
+  # plans **Stop/Start**, restarting the very process you meant to stop. It goes
+  # on polling and executing until logout, reboot, or an explicit stop. MEASURED
+  # with `sd-switch --dry-run` against three fixtures: identical generations =>
+  # `No action`; unit deleted outright => `Stop`; `[Install]` removed with the
+  # unit still present => `Stop/Start (SwitchMethod(StopStart))`.
+  #
+  #     enableTmuxReplyAgent = false;   # then merge + scripts/ship.sh, THEN:
+  #     systemctl --user stop tmux-reply-agent          # on BOTH hosts
+  #
+  # The flag alone is a durable disarm (nothing wants the unit at next login);
+  # the `stop` is what makes it take effect NOW. Removing CLAWGATE_TERMINAL_TOKEN
+  # from the POD is the other lever and is immediate for every host at once.
   #
   # 🔴 TWO INDEPENDENT SWITCHES, AND NEITHER IMPLIES THE OTHER. Arming needs
   # (1) CLAWGATE_TERMINAL_TOKEN provisioned into the POD's secret — until then the
@@ -3825,8 +3840,9 @@ in
       # tmux-snapshot-push: notify-failure@ toasts are wired to DEFEAT
       # do-not-disturb, and that bypass is justified by a MEASURED rate of about
       # one firing in nine days. This unit polls every few seconds, so any
-      # sustained condition — the surface not armed (which is the state it ships
-      # in), clawgate mid-redeploy, the laptop's network down — would fire a
+      # sustained condition — the surface not armed (which since arming means the
+      # pod's secret was REMOVED), clawgate mid-redeploy, the laptop's network
+      # down — would fire a
       # DND-bypassing toast on every restart and burn down the one alert channel
       # that has to keep its meaning.
       #
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -165,8 +165,14 @@ let
   #     systemctl --user stop tmux-reply-agent          # on BOTH hosts
   #
   # The flag alone is a durable disarm (nothing wants the unit at next login);
-  # the `stop` is what makes it take effect NOW. Removing CLAWGATE_TERMINAL_TOKEN
-  # from the POD is the other lever and is immediate for every host at once.
+  # the `stop` is what makes it take effect NOW, on the host you run it on.
+  #
+  # Removing CLAWGATE_TERMINAL_TOKEN from the POD is the other lever. It is
+  # fleet-wide rather than per-host — but 🔴 NOT INSTANT: per the paragraph
+  # below and the pod's own boot line, it takes effect on the POD'S NEXT BOOT,
+  # so the route keeps answering until then. Neither lever is immediate on its
+  # own; `systemctl --user stop` on both hosts is the one that stops execution
+  # at the moment you run it.
   #
   # 🔴 TWO INDEPENDENT SWITCHES, AND NEITHER IMPLIES THE OTHER. Arming needs
   # (1) CLAWGATE_TERMINAL_TOKEN provisioned into the POD's secret — until then the
--- a/nix/home.nix
+++ b/nix/home.nix
@@ -142,13 +142,16 @@ let
   # wired into `default.target`; the unit definition is always emitted, so it can
   # be started by hand regardless.
   #
-  # 🔴 IT SHIPS **false**, AND THAT IS THE POINT OF THE WHOLE SHAPE — not caution
-  # about an unfinished feature. What this agent delivers is `tmux send-keys`
-  # followed by Enter, i.e. arbitrary command execution as the operator on this
-  # host, driven by a route on a LAN NodePort that has no human auth. The operator
+  # 🔴 THIS IS NOW **true** — THE AGENT IS ARMED, AND WHAT IT DELIVERS IS ARBITRARY
+  # COMMAND EXECUTION AS THE OPERATOR ON THIS HOST. `tmux send-keys` followed by
+  # Enter, driven by a route on a LAN NodePort that has no human auth. The operator
   # took that decision deliberately, with the blast radius stated; what was ALSO
   # decided is that building it and ARMING it are separate acts, so the write path
-  # can be merged, deployed, read and audited before it can execute anything.
+  # was merged, deployed, read and audited across seven rounds while this flag was
+  # still false and it could execute nothing. It shipped `false` for exactly that
+  # reason — that ordering is history now, not a live guard, and the SHAPE it left
+  # behind is what still matters: setting this back to `false` is a one-line,
+  # reviewable disarm of the host half, independent of the pod's secret.
   #
   # 🔴 TWO INDEPENDENT SWITCHES, AND NEITHER IMPLIES THE OTHER. Arming needs
   # (1) CLAWGATE_TERMINAL_TOKEN provisioned into the POD's secret — until then the
@@ -167,7 +170,7 @@ let
   # ONLY from the machine it lives on, so a workbench-only agent would leave every
   # laptop pane permanently unanswerable. The queue is keyed on the host label, so
   # two agents claiming disjoint host scopes cannot collide.
-  enableTmuxReplyAgent = false;
+  enableTmuxReplyAgent = true;
   # Graphical host = runs X/i3 (both current NixOS hosts do; only a genuinely headless
   # box would not). Approximated as isNixOS, mirroring graphical.nix — deliberately NOT
   # !serverMode, which is true on the graphical workbench.
@@ -3801,10 +3804,11 @@ in
   # either host's tmux socket, and the alternative (an inbound port, or an ssh
   # credential inside a pod on an unauthenticated LAN surface) is strictly worse.
   #
-  # 🔴 SHIPPED DISABLED. `enableTmuxReplyAgent` is false — see its comment above
-  # for the two independent switches that arm this and why building it and arming
-  # it were deliberately separated. The unit below EXISTS on both hosts and is
-  # wired into nothing.
+  # 🔴 ARMED. `enableTmuxReplyAgent` is true — see its comment above for the two
+  # independent switches that arm this and why building it and arming it were
+  # deliberately separated. The unit below is wanted by `default.target` on both
+  # hosts and RUNS; it executes what the queue hands it. It shipped disabled and
+  # was armed only after the server half was deployed and audited.
   #
   # 🔴 Type = "simple", NOT the oneshot-plus-timer shape its two siblings use, and
   # the difference is the cadence. A ~5s poll cannot be a timer: the user manager
@@ -3876,8 +3880,8 @@ in
       ];
     };
     Install = {
-      # 🔴 SHIPPED DISABLED — see enableTmuxReplyAgent. A `default.target` want,
-      # not `timers.target`: this is a resident service, not a timer.
+      # 🔴 ARMED — see enableTmuxReplyAgent. A `default.target` want, not
+      # `timers.target`: this is a resident service, not a timer.
       WantedBy = lib.optionals enableTmuxReplyAgent [ "default.target" ];
     };
   };
--- a/scripts/tests/test_no_real_launchers.py
+++ b/scripts/tests/test_no_real_launchers.py
@@ -373,9 +373,14 @@ ACKNOWLEDGED_UNSTUBBED = {
         "behind sudo and neither script is executed by scripts/tests"),
     "wmctrl": (
         {"session-write"},
-        "The FOURTH occurrence of the prose-mention shape already justified "
-        "three times under `home-manager` above, and justified here rather "
-        "than reworded away — this scan is a TEXT scan "
+        "Another occurrence of the prose-mention shape justified under "
+        "`home-manager` above (that entry numbers its own), and justified here "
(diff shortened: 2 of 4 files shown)


ZacxDev:
**Audit round 1** — findings and what the fixes claim to have addressed. Round 2 audits the delta below.

```audit-claims round=1 audited=b32a65a9..d46ca292
1. 🔴 The documented rollback was FALSE: flipping enableTmuxReplyAgent false and shipping does NOT stop a running agent (sd-switch plans Stop/Start, not Stop). Claimed fixed by documenting disarm as TWO steps (flag + `systemctl --user stop` on both hosts) at three sites: the flag comment in nix/home.nix, the agent module docstring, and the guard docstring most likely to be mistaken for a kill switch.
2. 🟡 SIX stale statements survived, not the five the first commit fixed. Claimed fixed: the agent's `IT IS INERT` headline, the BACKOFF_SECONDS comment, home.nix's OnFailure rationale, two test-file sites, and — the one that is not a comment — the RUNTIME 503 message that told the operator "This is the shipped state, not a fault."
3. 🟡 The first commit's own justification for inverting the positive control was FALSE (it claimed the old direction would go "green and blind"; `assert flipped != src` already existed, so it fails loudly). Claimed fixed by RETRACTING it in place rather than deleting it, naming `!= src` as the line that does the work.
4. 🟢 The first commit's mutation table reported 3 failures for mutant 3; the true count is 2. Claimed fixed by re-running the battery and correcting the number.
5. Round 1's own fix introduced a NEW gate failure: adding `systemctl` to the agent docstring made a text-scanning launcher guard see a new file reaching an acknowledged binary (test_no_real_launchers.py went red on the merged tree). Claimed fixed by ACKNOWLEDGING the prose mention in the ledger with grep evidence of no call site — deliberately NOT by rewording the docstring, since that sentence is the only written rollback.
6. The base moved twice during the round (f0b9c474 -> c5a445d8 -> 88f1bda4). Claimed addressed by merging current main into the branch and re-running both tiers on the resulting merged tree.
```

**Not claimed fixed, and deliberately out of scope for this PR** — recorded so round 2 does not read it as an omission:

- An intermittent failure in `test_the_EXACT_session_still_works`, measured **2 failures in 10 runs**, one on a completely unmutated tree. Mechanism identified: `scripts/tmux-reply-agent:573` calls `out.strip()` before `.split("\t")`, so a `NEW_WINDOW_FORMAT` response whose trailing `#{pane_current_path}` is empty collapses from 3 fields to 2 and is rejected — after the window has already been created. The observed failure string `'%2\tscratch20'` is byte-identical to `'%2\tscratch20\t'.strip()`. ⚠ The empty-field trigger is **inferred, not forced**: 8/8 direct probes returned it populated. This is pre-existing (`#1324`), lives in `open_window` (rank 31's path, not the send-keys reply path), and does not block arming.
- `ExecStart` runs `%h/workspace/devrc/scripts/tmux-reply-agent` — the shared working tree, not the audited store copy. Pre-existing house pattern (~20 units); arming is what makes it live.

ZacxDev:
**Audit round 2** — delta `b32a65a9..d46ca292`. Not clean; three findings, two of them defects in round 1's own corrective prose. Round 3 audits the delta below.

```audit-claims round=2 audited=d46ca292..ac4372cc
1. 🔴 Round 1's RETRACTION was false about the tree it landed in, and round 1's own fix is what falsified it: the disarm example it added to nix/home.nix comments (`  #     enableTmuxReplyAgent = false;`) has a `  #     ` prefix ending in spaces, so the two-space-prefixed literal matches INSIDE the comment. Reverting the control to its pre-flip direction gives `1 passed`, not the `1 failed` the retraction asserted — the exact "green and blind" the retraction said could not happen. Claimed fixed by ANCHORING the mutation (`re.subn(r"^  enableTmuxReplyAgent = true;$", flags=re.M)` with substitution count asserted == 1) and rewriting the retraction to record that a fix for a claim created the hazard the claim denied. The disarm example was deliberately KEPT.
2. 🔴 Round 1's launcher-guard acknowledgement named a compensating control that did not exist ("test_tmux_reply_agent.py pins that the agent shells out to nothing else"). Because hazard_hits returns a FILE set and the ledger asserts set equality, once the file is in the set a real call site is invisible: injecting `subprocess.run(["systemctl", ...])` left test_no_real_launchers.py at 77 passed. Claimed fixed by WRITING the pin — test_the_agent_SHELLS_OUT_TO_TMUX_AND_NOTHING_ELSE, an AST walk asserting the agent's spawn argv[0] set is exactly {tmux_bin} — and by rewriting the ledger justification to name it and record the measured gap. The first draft of that walker failed on a CLEAN tree by matching os.getpid(); it was narrowed to exec/spawn verbs and both controls re-run.
3. 🟡 A sentence round 1 added to nix/home.nix ("Removing CLAWGATE_TERMINAL_TOKEN from the POD ... is immediate for every host at once") contradicted the same file ten lines below ("takes effect on the pod's next boot"). Claimed fixed by stating the pod lever is fleet-wide but NOT instant, and that `systemctl --user stop` is the only lever that stops execution at the moment it runs.
4. 🟢 Two stale figures in round 1's ledger prose: "three times" under `home-manager` is FIVE (that entry numbers its own fourth and fifth), and session-write's entry is under `wmctrl`, not under the systemctl key. Claimed corrected.
5. The base moved again (88f1bda4 -> c25ef63c). Claimed addressed by merging current main and re-running all four gate legs on the resulting merged tree.
```

**Round 2 items NOT claimed fixed, recorded so round 3 does not read them as omissions:**

- The PR body still carried round 1's un-retracted "green and blind" section and the wrong mutant-3 count (`3 failed`; true count `2`). Being corrected in the body alongside this comment — the body is prose, not code, and no commit carries it.
- Round 2 found the pre-existing `out.strip()` defect in a **second** test (`test_a_cwd_that_does_not_exist_is_REFUSED_not_opened_in_HOME:1609`, same 3-fields-to-2 collapse), so that surface is wider than the single test named in round 1's block. Still pre-existing (`#1324`), still `open_window`, still out of scope for this PR.
- `ExecStart` runs the working tree rather than the store copy — unchanged, pre-existing, and proven untouched by this delta (the comment-stripped nix diff since `b32a65a9` is **0 lines**).

ZacxDev:
**Audit round 3** — delta `d46ca292..ac4372cc`. Not clean; three findings, the first being that round 2's new pin was narrower than its own docstring. Round 4 audits the delta below.

```audit-claims round=3 audited=ac4372cc..cc25c0fe
1. 🟡 Round 2's AST pin (test_the_agent_SHELLS_OUT_TO_TMUX_AND_NOTHING_ELSE) matched only a literal `subprocess.<verb>(...)` / `os.<verb>(...)` attribute call and never read import bindings, so it was narrower than its own docstring AND than the launcher-ledger sentence staking itself on it. MEASURED four survivors: `from subprocess import run as _r`, `from subprocess import run`, `import subprocess as _sp`, and `os.posix_spawn` — with one live, the two files were 183 passed. Claimed fixed by resolving import bindings (ast.Import/ImportFrom honouring asname; a bare Name bound out of subprocess counts as a spawn; posix_spawn/posix_spawnp named explicitly). Claimed AFTER state: negative control passes on a clean tree and ALL SEVEN injections are KILLED.
2. 🔴 The first run of that survivor battery was INVALID and is recorded rather than hidden: the scratch root was never created, so every `cp -a` failed, pytest ran against nonexistent paths, and all seven injections scored KILLED — together with the negative control. Four survivors becoming zero reads exactly like a working fix; only the control being wrong distinguished it. Claimed fixed by making the battery assert the copy exists, assert the agent file exists, and require a `1 passed`/`1 failed` verdict line before scoring anything.
3. 🟢 A sentence in round 2's retraction overstated the anchor: it said a comment at COLUMN 0 would be "a failure rather than a silent extra substitution". MEASURED `1 passed` — `^  ` cannot match a line beginning `#`, so such a comment is INVISIBLE, not detected. Claimed fixed by stating that plainly and keeping the second-declaration half, which is true and carries its own assertion message.
4. 🟢 Round 2 corrected the prose-mention count at the `systemctl` entry (to FIVE) but left the `wmctrl` entry saying "three times", so one dict held two figures for one population. Claimed fixed by REMOVING the running total rather than renumbering a third time — the entry now points at the home-manager entry, which numbers its own.
5. The base moved again (c25ef63c -> 43f86a14). Claimed addressed by merging current main and re-running all four gate legs on the resulting merged tree.
```

**Refuted by round 3, recorded so it is not re-investigated:** I had flagged a suspected asymmetry — that the `inverted` half of the same control still uses a plain `str.replace` on the `WantedBy` line and might carry the comment vulnerability the anchored half had just fixed. It does not: `_wanted_by_expr` strips comments before reading, so a literal planted in a comment leaves the control failing loudly. Round 3 verified this with a discriminating probe and discarded a confounded first attempt before concluding. No change was made.

**Still not claimed fixed, and still out of scope** — the pre-existing `out.strip()` 3-fields-to-2 collapse at `scripts/tmux-reply-agent:573` (two affected tests), and `ExecStart` running the working tree rather than the store copy. Both proven untouched by this delta: `scripts/tmux-reply-agent` has **zero** changed lines in `d46ca292..ac4372cc`.

PR #1336, opened 2026-09-06
fix(mention): pin the disclosure guard on KEYS too, keep the staleness arm the audit called dead, and make `audit-pr N` clickable

ZacxDev:
Three things, in one branch because they touch the same four files and separate PRs would conflict at merge.

## F1 — the staleness note is NOT dead code, and the audit's deletion was wrong

The round-2 audit proposed deleting the `if not reason and (stale := staleness_note())` arm in `refuse()`, reasoning that `universe_reason() == ""` implies a non-empty `clean_repo_map`, that `repo_universe` filters on the same `OWNER_REPO_VALUE_RE`, and that `discover_repos` only ADDS rows — so the universe must be non-empty, the picker is shown, and `refuse()` is never reached.

Every step is true except the last. **`discover_repos` also OVERWRITES**, and `parse_owner_repo` is looser than `OWNER_REPO_VALUE_RE`: it asks only for two `/`-separated segments, so `https://github.com/-acme/widget.git` yields the non-empty `-acme/widget`, which the value regex rejects. A checkout whose directory name shadows a mapping key replaces a valid row with an invalid one.

Measured end-to-end — mapping `{"widget": "acme/widget"}`, a `~/workspace/widget` checkout whose remote parses to `-acme/widget`:

```
universe_reason  : ''
repo_universe    : []
rc               : 1
notify body      : no repository owner is known for it; the repo mapping is 400 days old ...
```

So the branch is KEPT, and what was actually missing is a test — deleting it left the suite green because the only refusal test naming an age used `--print`, which takes the other arm. Added `test_a_NON_print_refusal_CAN_still_name_the_mapping_AGE`, its premise assertion, and a fresh-mapping negative control. The comment now states the reachability condition so the next reader does not re-derive "dead code", and the module-level claim no longer reads "the refusal body" (which invited "every refusal past 7 days").

## F3 — the disclosure tripwire pinned VALUES, never KEYS

`FAKE_UNIVERSE` spelled every key as a substring of its own value (`trowelcast -> gardenersguild/trowelcast`), so guards iterating `.values()` looked total while being blind to a key leak. The audit's mutant — a "did you mean?" line carrying `", ".join(sorted(load_known_repos()))`, i.e. every private repo NAME, onto stderr and `notify-send` — survived all 127 tests.

- Keys, owners and repo halves are now **pairwise distinct** — **12 tokens over three rows** (key, full `owner/repo`, bare owner, bare repo-half; nine of the twelve carry no slash, which is the substring sweep's count and not the ledger's) — pinned by `test_the_fixtures_KEYS_and_VALUES_are_pairwise_distinct`.
- One `_no_universe_token_anywhere()` helper asserts on the whole token set, used by all three disclosure guards plus the staleness note.
- `_mapping_aged()` writes `FAKE_UNIVERSE` so the mapping `refuse()` re-reads holds the names the guards look for — without this the key half was vacuous.
- The tailer's attributed-run guard now derives its forbidden set instead of hand-listing it.

**Measured hole nobody had counted: NINE tests in `test_mention_open.py` read the operator's real `~/.config/mention-open/known_repos.json`** (seven `stat`, two `read_text`), which is why they asserted different things on the dev host than in the nix sandbox. An autouse fixture redirects `KNOWN_REPOS_PATH` per test, with `test_the_autouse_redirect_is_IN_FORCE` as its positive control.

## `audit-pr N` is clickable — and it took TWO files

The operator clicked `audit-pr 1291` and nothing happened. Two independent blockers, both measured:

1. `PATTERN_LEDGER` had `AUDIT_PR_RE` as telemetry-only, so `scan_mention_spans` returned `[]` at the terminal profile.
2. The Alacritty hint regex never underlined it, and unhighlighted text cannot be clicked.

Fixing either alone ships a feature that is 100% dead and unit-tested green. Hint regex:

```
-  ...#[0-9]{1,6}|868[a-z0-9]{6}
+  ...#[0-9]{1,6}|868[a-z0-9]{6}|/?audit-pr[ \t]+[0-9]{1,6}
```

`{1,6}` mirrors the existing `#` bound for the same reason: `{1,5}` would underline `audit-pr 12345` inside `audit-pr 123456` and open a confident wrong page.

**A third blocker the brief did not name:** the terminal surface is pinned as unchanged in five places (the terminal hint set, the unknown-profile fallback, the telemetry-hints invisibility check, and both residual false-positive sets). Each is a deliberate decision record, so each was updated to keep asserting something real rather than relaxed — the terminal hint set is now `("#", "868", "audit-pr")`, three literals, not the two it had been — and `/audit-pr 12` moved from `_KNOWN_FALSE_POSITIVES_TELEMETRY` to `_KNOWN_FALSE_POSITIVES`, because an instructional example in a runbook now gets underlined. The two sets are now asserted disjoint.

Scope held exactly: only `audit-pr N`. `gh pr view N`, `gh issue close N`, `clawgate task N` and `#task-N` stay telemetry-only, asserted by absence.

The URL hint is **byte-identical** — `test_the_url_hint_regex_is_alacrittys_own_default` still passes and the diff does not touch that line.

---

# Round-1 audit fixes (three findings, each mutation-verified)

## 1. The tailer's disclosure guard was blind to the OWNER half — the half that names the client

`forbidden` was `set(FAKE_REPOS) | set(FAKE_REPOS.values())`. The audit's mutant leaking `{v.split('/')[0] for v in load_mention_repos().values()}` **SURVIVED** while the KEYS mutant written beside it died. The sink is a durable ClickHouse table, and the 2026 incident disclosed 232 private repo names, **167 of them one client's** — the client identity *is* the owner half.

`FAKE_REPOS` also still spelled every key as a substring of its value — the exact fixture shape F3 above diagnoses, left standing in the sibling file. Now pairwise distinct (12 tokens over 3 rows) with its own `test_the_fixtures_KEYS_OWNERS_and_REPOS_are_pairwise_distinct`, and both spool guards check all four spellings through `_repo_tokens`.

Measured, battery filtered to the new rows plus the P1 control:

| | K43 (OWNERS) | K44 (repo-halves) | P1 control |
|---|---|---|---|
| at `59a18db4` (pre-fix guard) | **SURVIVED** `f=0 p=390` | KILLED(attributed) | KILLED |
| at HEAD | **KILLED(attributed)** `f=2 p=389` | KILLED(attributed) | KILLED |

K43 is the audit's mutant verbatim. Both rows are committed in `scripts/tests/mutation_battery_mentions.py`; with K13 (values) and K14 (keys) the four exercise every arm of `_repo_tokens`.

## 2. The autouse real-mapping redirect was in-process only; a subprocess walked through it

`monkeypatch.setattr(MO, "KNOWN_REPOS_PATH", p)` rebinds an attribute on the test process's module object. The same file's `_run()` spawns `mention-open.py` as a real child, which re-resolves the constant at its own import. Four tests use `_run`; all four happened to pass `--no-discovery` (under which the mapping is never read). **Nothing enforced that.**

Fix: `monkeypatch.setenv("MENTION_OPEN_KNOWN_REPOS", str(p))` in the same fixture — `mention-open.py:136` already reads it and a child inherits the environment.

**And the control is fixed too**, because a control that cannot observe the failure it names is not a control. `test_the_autouse_redirect_is_IN_FORCE` inspects a module attribute the child never consults, so it stayed green through the whole gap; it now says so. Two new guards:

- `test_the_autouse_redirect_REACHES_A_SUBPROCESS` — asserts the path a **child** resolved, plus a positive control that the child actually READ the fake mapping.
- `test_a_DISCOVERING_subprocess_resolves_through_the_FAKE_mapping` — the behavioural half, on the one flag combination that was unguarded, with `DEVRC_WORKSPACE` pointed at an empty directory so the host's real checkouts cannot decide the answer.

Measured with the `setenv` deleted: both go red — the child reporting `/home/zach/.config/mention-open/known_repos.json` — while the old in-process control stays **green**.

## 3. `audit-pr N` auto-opened on a repo guessed from the tmux pane

`repo_source=default` is the weakest rung of the attribution ladder — the text named no repository and the answer came from `tmux_pane_repo()`, which answers for the most recently *active* client, not necessarily the clicked pane. `main()`'s "one candidate → just open it" shortcut fired on it. A bare `#N` was protected only **incidentally**, by having a clawgate sibling.

Suppressed at the `repo_source` level, **not** on `audit-pr`, so any future no-owner shape inherits the rule. Measured with the pane forced to `WRONGORG/wrongrepo`:

| text | before | after |
|---|---|---|
| `audit-pr 1291` | **OPEN-DIRECTLY** `WRONGORG/wrongrepo/issues/1291` | picker, 1 row |
| `/audit-pr 1291` | **OPEN-DIRECTLY** | picker, 1 row |
| `#1291` | picker (2 rows) | picker (2 rows) |
| `loamfield#12` (`mapped`) | open | open |
| `civitai/talos-infra#1065` (`explicit`) | open | open |

`tmux_pane_repo`'s docstring promised a wrong guess "shows up as an extra row in the picker, **never as the wrong page opening**". That is true again — and now enforced where it is decided rather than promised where it is guessed. The one-row picker carries a `guessed_note` saying why (rofi cannot report a reason after a dismissal); the ordinary two-row `#N` picker is deliberately untouched, with a negative control asserting so. `SOURCE_DEFAULT` is imported from `mention_scan`, never spelled, so a rename cannot silently disarm the branch.

## Also closed

- `mention_scan.mention_hints` and `session-tailer` both enumerated `audit-pr 1291` as telemetry-only, contradicting the ledger row 46 lines away. Both corrected; the terminal profile's literal SET is now asserted rather than described.
- `FAKE_UNIVERSE_TOKENS` is 12, not nine (the nine was attached to the wrong object).
- `test_alacritty_hints.py` did a global `sys.path.insert` in a test body and never undid it — now a context manager.
- The alacritty residual named `myaudit-pr 12` but not the likelier path form. Measured on the same pattern: `scripts/audit-pr 12` and `claude/skills/audit-pr 12` both underline `/audit-pr 12` and resolve — and this repo *has* a `claude/skills/audit-pr/` path.
- `test_the_universe_never_reaches_a_LOG_a_SPOOL_or_stderr` covered neither a log nor a spool — renamed to the three sinks `_every_sink` actually reads, with the reason recorded.

## The battery's own parser had a false-`SURVIVED` path

Found while scoring K43. `re.search(r"(\d+) failed", out)` takes the first match **anywhere**, and the tailer under test prints `session-tailer: scanned=1 emitted=1 failed=0 …` — which contains the substring `1 failed`. Pytest's real summary said `2 failed, 389 passed`; the battery read `nfail=1`.

Harmless in that direction. The one that is not: the same line with `emitted=0` reads `nfail=0` over a **red** suite and reports `SURVIVED` for a mutant every guard caught. Counts now come from pytest's own summary line only (`counts_from`), unit-tested in the *collected* `test_mutation_battery_anchors.py` against the real line that broke it, with a no-summary case that resolves to `NOT-OBSERVED` rather than `SURVIVED`.

A `--only ID[,ID]` selector was added so one row can be re-checked without a ~25-minute sweep. It never drops the `P1` positive control and prints a FILTERED banner above **and** below the count, because `3/3 killed` from three rows reads like a verdict on the battery.


ZacxDev:

```audit-claims round=1 audited=59a18db4..c0b3f44b
1. The tailer's spool guards now use a derived four-spelling ledger (keys, full owner/repo, bare owners, bare repo-halves) applied to BOTH guards; FAKE_REPOS is claimed pairwise-distinct and enforced by its own test.
2. The audit's OWNERS mutant is committed as K43 and claimed to move SURVIVED at 59a18db4 to KILLED(attributed) at HEAD; K44 adds the fourth spelling.
3. The autouse fixture now sets MENTION_OPEN_KNOWN_REPOS so the redirect reaches a subprocess, and its control is replaced by one that asserts the path a CHILD resolved — claimed to go red when only the setenv line is deleted, while the old control stays green.
4. A discovery test that spawned a real `tmux display-message` now passes --default-repo instead.
5. The one-candidate auto-open shortcut is suppressed at the repo_source level whenever repo_source == SOURCE_DEFAULT, not special-cased to audit-pr; a bare #N is claimed unchanged, with negative controls for mapped and explicit.
6. tmux_pane_repo's docstring is corrected and points at the enforcing test; a one-row picker carries guessed_note, which is claimed to name the clicked text only and never a repo.
7. A third stale copy of the audit-pr-is-telemetry-only claim (mention_scan.py's module docstring table) is fixed, and the terminal hint set is asserted as a set of three rather than described in prose.
8. The battery's failure count is now read from pytest's own summary line rather than the first `N failed` match anywhere — the tailer's `emitted=1 failed=0` output was being misread; a no-summary case resolves to NOT-OBSERVED.
9. Full battery reported 45/45 killed for the stated reason, P1 killed, restore hash-verified.
10. Both tiers green on merged sha c0b3f44b containing origin/main c19fb289.
```

⚠ Two corrections to the round-1 brief, from the fix round: Tekton's `devrc-pytests` was **not** red on `59a18db4` (both checks passed), and a red gate round was attributed to a pre-existing failure owned on another branch (`e6770754`), not to this diff.

ZacxDev:
Round-2 delta audit found one 🔴 and it is now fixed: **both sides were under the `scripts/tests` drift ceiling and the MERGE crossed it.**

| tree | collected | vs ceiling 12836 |
|---|---|---|
| `origin/main` alone | 12818 | under by 18 |
| this branch alone | 12806 | under by 30 |
| **merged** | **12843** | **over by 7** |

Neither side is over alone, so nothing on either branch could see it — only a merged-tree gate, which is what caught it, in both tiers, before it could land a red `main`. Same shape as the 2026-08 case recorded directly above the floor entry (`alone (10137 and 10112); the SUM crossed it`).

Floor raised to `"scripts/tests|12793"` — **the number the gate printed itself**, on the tree I measured, and pinned **after** merging `main` into the branch per the ORDER note beside it.

### Gate — all four legs, merged tree containing `origin/main` `1ecc03c1`

| tier | leg | result |
|---|---|---|
| dev host `gate.sh --tier both` | pytest | `RESULT: PASS (exit=0)` |
| dev host | node | `RESULT: PASS (exit=0)` — 1449/1449 |
| sandbox `nix build …pytests` | pytest | `RESULT: PASS (exit=0)` |
| sandbox `nix build …nodetests` | node | `RESULT: PASS (exit=0)` — 1449/1449 |

⚠ **A gating trap worth recording.** The sandbox pytest leg first failed `exit=2` on `run-tests: FATAL — 'git config --global' does NOT write to this run's isolated file` — *before any test ran*. Cause was the **method, not the tree**: `nix build path:<a worktree>` copies the worktree's `.git`, which is a FILE pointing at the real git dir, and that defeats the sandbox's git-config isolation. Same content extracted with `git archive` (no `.git`) passes. The rules document this hazard for `cp -a` copies of worktrees; `nix build path:` is a second door to it.

PR #1338, opened 2026-09-06
feat(opencode): serialise one session to one canonical, byte-deterministic artifact (clawgate #511)

ZacxDev:
Closes clawgate **#511**. Session id in, one byte-deterministic NDJSON artifact out — one record per part — built on the reader that already exists rather than a new one.

## 🔴 The task's own first assumption is refuted

It assumed `_shared.iter_parts`' `text` field carries the receipts, with an explicit stop condition if not. Measured across 12 sessions / 1,055 parts **before starting**:

| part type | count | has `text` | has `_data` |
|---|---|---|---|
| **tool** | 378 | **0** | 378 |
| patch | 35 | 0 | 35 |
| step-start / step-finish | 188 / 188 | 0 | ✓ |
| reasoning | 177 | 177 | ✓ |
| text | 89 | 89 | ✓ |

`text` is populated on **25%** of parts and on **none** of the tool parts — which are the receipts. An exporter written to the assumption's letter ships an artifact three-quarters empty containing zero tool calls, *while looking like it worked*.

It does **not** invert the task: `_data` is already exposed on all 378, and criterion 2 says "the fields the existing reader **already yields**". The assumption was wrong; the criteria were right.

## 🔴 Determinism needed a total order the reader does not provide

`iter_messages`/`iter_parts` both `ORDER BY time_created` alone, and SQLite leaves ties unspecified — two rows sharing a millisecond could reorder between runs. Everything is re-sorted on `(time_created, id)`.

⚠ **The hazard is latent, and that decides the test design.** Zero ties across 617 messages and 2,907 parts on this host — so *real data cannot exhibit the bug*, and a determinism test reading only the real store would pass **with the sort deleted**. The fixture therefore carries a deliberate tie with ids sorting opposite to insertion order, plus a positive control asserting the tie actually ties.

## Matrix (criterion 7)

| | |
|---|---|
| RED at base `88f1bda4` | collection error — `export.py` absent there; all 11 fail |
| GREEN at HEAD | 11 passed |
| **Mutation, isolated** | dropping the `id` tiebreaker kills **exactly** `test_ordering_is_total_when_time_created_ties` at its own assertion — 1 failed, 10 passed. Under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` cleared; source restored byte-identical (sha verified) |
| Regressions | full `scripts/collector/opencode/tests` — **214 passed** |
| Live exercise | a real session: 210 parts, 1.13 MB, two runs byte-identical, **55/55** tool parts carrying `_data` |

Criterion 4 is an **AST** assertion, not a grep, with a negative control proving the pattern matches a real offender — a check that cannot go red is not a check. The CLI separates the two empty cases (rc 3 unknown session, rc 4 known-but-empty), because both produce no output and exiting 0 on a typo would let a caller record an empty artifact as success.

## Non-goals held

No wiring into any shipper, bucket or cairn field; no second DB reader; no change to the tailer/backfill/plugin; no opencode session-end hook hunt; `tool-output/`, `snapshot/` and `storage/` excluded **and named as excluded** in the module docstring, per criterion 5.

https://claude.ai/code/session_014wTuXEuj8Hs89X6H8xYpjq

### DIFF
--- a/scripts/collector/opencode/export.py
+++ b/scripts/collector/opencode/export.py
@@ -16,51 +16,86 @@ session" is a claim about a population and an unnamed one is worthless:
     are real files a session produced and they are DELIBERATELY excluded — this
     reads the DB only. A session that wrote a large tool output has that output
     referenced here and not embedded.
+  * 🔴 **a message with NO parts produces NO record at all**, because the unit is
+    the part. Measured 2026-09-06: 38 of 17,679 messages store-wide. Such a
+    message is invisible in the artifact — not empty, absent.
   * whatever `_shared.iter_*` drops. Those functions project the `data` JSON blob
     onto named fields; this module carries `_data` (the whole parsed blob) so the
     projection is not additionally lossy, but a column `_shared` does not SELECT is
     invisible here too.
-  * any session-level file outside the `session`/`message`/`part` tables.
 
 🔴 THIS MODULE OPENS NO DATABASE OF ITS OWN. It takes a connection from
-`_shared.get_db()`, which opens `mode=ro`. Three readers of this store already
-exist (`_shared`, the tailer, `find-session`); a fourth connection path is what
-`tests/test_export.py::test_export_opens_no_connection_of_its_own` exists to
-prevent, and it asserts over the AST rather than trusting this comment.
+`_shared.get_db()`, which opens `mode=ro`. Counting actual `sqlite3.connect` sites
+against this store there are exactly TWO — `_shared.py` (`mode=ro`) and
+`scripts/lib/opencode_search.py`, which is **not** read-only — so this would be a
+third. `tests/test_export.py::test_export_opens_no_connection_of_its_own` is what
+prevents it, over the AST rather than trusting this comment.
 
 🔴 DETERMINISM RESTS ON A TOTAL ORDER, AND THE READER DOES NOT PROVIDE ONE.
 `_shared.iter_messages` and `iter_parts` both `ORDER BY time_created` alone. That
 is not a total order: SQLite leaves ties unspecified, so two rows sharing a
-millisecond can come back in either order and the artifact's bytes would change
+timestamp can come back in either order and the artifact's bytes would change
 between runs for reasons nothing controls. Every sequence here is therefore
 re-sorted on `(time_created, id)`.
 
-⚠ MEASURED 2026-09-06: **zero** ties across 617 messages and 2,907 parts on this
-host. So the hazard is LATENT, not live — which is precisely why the determinism
-test must use a fixture carrying a DELIBERATE tie. Real data cannot exhibit the
-bug, so a test that only reads real data would pass whether or not this sort
-exists, and would report coverage it does not have.
+⚠ MEASURED 2026-09-06, STORE-WIDE (691 sessions / 17,679 messages / 77,671 parts):
+**5 part tie-groups covering 10 rows in 5 distinct sessions**; zero message-level
+ties. So ties are RARE — about 1 part in 7,767 — but they are REAL, and an earlier
+revision of this comment said "zero ties … real data cannot exhibit the bug" on the
+strength of a 2,907-part sample, i.e. 3.7% of the store, stated at host scope. That
+was false, and it was the dangerous direction: it invited deleting the sort below as
+unfalsifiable defensive code. **The rarity is why the test fixture carries a
+DELIBERATE tie — a sampled session will not supply one — not evidence that the sort
+is unnecessary.**
 """
 from __future__ import annotations
 
 import argparse
 import json
+import os
+import sqlite3
 import sys
 from pathlib import Path
 from typing import Any, Iterator
 
-_HERE = str(Path(__file__).resolve().parent)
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Round 1 — 8 🟡, 6 🟢, no 🔴. All fixed.

The two must-fix findings were re-verified here before acting, and both were defects in **my own** work rather than in the payload's logic.

### 🔴 F1 — my measurement was false at the scope I stated it

I claimed *"zero ties across 617 messages and 2,907 parts … real data cannot exhibit the bug"*. That sampled **3.7%** of the store and was stated at **host** scope. Re-derived store-wide (691 sessions / 17,679 messages / **77,671 parts**): **5 part tie-groups covering 10 rows in 5 distinct sessions.** The message half held (zero ties).

The direction of the error is what matters: *"real data cannot exhibit the bug"* invites the next maintainer to delete the sort as unfalsifiable defensive code. Restated at the scope actually measured, with the rarity (~1 part in 7,767) given as the reason the **fixture** manufactures a tie — not as evidence the sort is unneeded.

### 🔴 F6 — the criterion-4 control certified a copy of itself

`test_the_ast_check_can_actually_fire` re-implemented the matcher inline instead of driving the guard. **Verified: mutating the guard's own pattern (`{"connect"}` → `{"connect_NEVER"}`) left the suite at 11 passed** — the "instrument can go red" control stayed green while the instrument was fully disarmed. The copy had already drifted, walking only `ast.Attribute` where the guard walked `Attribute` *and* `Name`.

Both now run through one `_connect_offenders()` helper, with the control parametrised over three real refactors — `sqlite3.connect`, `from sqlite3 import connect as _c`, `getattr(sqlite3, "conn"+"ect")`. **The last two survived before.**

### The rest

- **F2/F3** `ensure_ascii=False` had two destructive latent failures (0 of 77,671 live parts, both real): a **lone surrogate** raised `UnicodeEncodeError` *after* truncating the output file — destroying the artifact it was refreshing — and **U+2028/2029/0085** passed through raw, so `splitlines()` shattered a record into unparseable fragments while the file stayed valid NDJSON.
- **F8** the artifact was **0644 from a 0600 source**, followed a symlink onto its target, and truncated before writing. Now `O_NOFOLLOW` + 0600 temp + `os.replace`. ⚠ The test asserts the **property** (target survives), not a mechanism — my first draft demanded `OSError` and failed against the *correct* implementation, since replacing the symlink is the safe outcome and doesn't raise.
- **F7** giving emptiness a *meaning* made a renamed table report `no such session`, blaming the caller for a broken store. **11 of 691** live sessions genuinely have zero parts, so rc 4 really fires and must not be overloaded. New rc 5.
- **F4/F5/F9/F10/F11/F12/F13/F14** — two surviving mutants (`sort_keys`, the message sort), a `TypeError` on heterogeneous `time_created`, the zero-part-message omission (38 of 17,679) now named, connection closed, `.resolve()` reverted to the siblings' documented nix-store reasoning, the reader enumeration corrected (two actual connect sites, one **not** read-only), vacuous assertions cut.

### Mutation matrix — all four previously-SURVIVING mutants now killed and isolated

| mutant | before | now |
|---|---|---|
| disarm the guard's matcher | SURVIVED | 1 failed / 19 passed |
| `sort_keys=True` → `False` | SURVIVED | 1 failed / 19 passed |
| drop the message sort | SURVIVED | 1 failed / 19 passed |
| `ensure_ascii=True` → `False` | — (new fix) | 2 failed / 18 passed |
| unmutated control | — | 20 passed |

`PYTHONDONTWRITEBYTECODE=1`, `__pycache__` cleared between runs, sources restored byte-identical (sha verified).

**Also worth recording: the audit re-derived every measurement in the PR body and confirmed the rest at larger scale** — `text` on **0 of 21,749** tool parts store-wide (I claimed 0 of 378), `_data` on 21,749/21,749, and cross-process determinism with differing `PYTHONHASHSEED` on a 1,389-record session.

```audit-claims round=1 audited=4a1ef3fd..c7a78cc1
1. The tie measurement restated at store-wide scope: 5 tie-groups / 10 rows / 5 sessions across 77,671 parts, replacing a false "zero ties, real data cannot exhibit the bug" from a 3.7% sample.
2. The criterion-4 guard and its negative control now share one _connect_offenders() helper, so disarming the guard cannot leave the control green; the control is parametrised over three refactors including the two that survived.
3. ensure_ascii=True, with the lone-surrogate and U+2028/2029/0085 hazards each pinned by a test.
4. The artifact is written O_NOFOLLOW + 0600 to a temp file and os.replace'd; the test asserts the target survives rather than that an exception is raised.
5. New rc 5 for an unreadable store, so schema drift is no longer reported as an unknown session; DatabaseError and IndexError caught at the main() boundary.
6. sort_keys and the message-level sort each gained a test that kills their mutant; _order_key type-ranked so a heterogeneous time_created cannot raise.
7. The docstring names the zero-part-message omission (38 of 17,679), the reader enumeration corrected to two connect sites with one not read-only, .resolve() reverted, connection closed, vacuous assertions removed.
```

Suite: 20 here, 223 across the opencode dir. Live re-exercise after the encoding change: 210 records, two runs byte-identical, every line parses, output pure ASCII.

Round 2 dispatching — round 1 produced findings that needed fixing.

PR #1342, opened 2026-09-06
fix(audit-dispatch): the stop-note prescribed a no-op, and three pins were narrower than their docstrings

ZacxDev:
Fixes the three 🟡 findings from #1185's round-2 delta re-audit, plus the two 🟢 nits in the same files — and then, after a blind audit of this PR, the two ways **round 18's own repairs were still narrower than their docstrings**. Handoff `claudedocs/handoff-audit-pr-ladder.md` rank 6.

---

# Round 19 — the audit of this PR

Round 18's two scaffolding repairs were both walkable by the very hazard they were written against. Reproduced here before fixing, at `74cb7409`:

| mutant | attack | result |
|---|---|---|
| **V68** | `[ -n "$DRV" ] \|\| { …; cat "$ERR"; }   # exit 1; }` | **SURVIVED, 128 passed** |
| **V69** | `…; grep -n "RESULT:" "$LOG" \|\| true` | **SURVIVED, 128 passed** |

**V68** — `endswith("exit 1; }")` anchors at the **END** of the line, which is exactly where a trailing comment lives. `re.match` excludes comments because it anchors at the START; `endswith` cannot. So the check written to refuse *"the stop removed, the words kept in a comment"* accepted the words kept in a comment, one character from the mutant it does kill. The rendered guard carries no `exit`, and the block walks into `nix log ""` — the CWD flake's default-package log. Byte-for-byte V63's hazard.

**V69** — the last command was `lines[-1].split(";")[-1]`, so a command reached by `||` was invisible, under a failure message that named a trailing `true` explicitly. Against a log with no `RESULT:` line: shipped block → rc 1; V65 (`; true`, killed) → rc 0; this (`|| true`, survived) → rc 0.

Round 18 pinned the **symptom shape** (a suffix, a `;`) where the hazard is a **class**. That is verbatim the criticism round 18 filed against round 17.

## Both offered one-line repairs were rejected, each measured overshooting

- **`"#" not in lines[i]` is narrower than the language.** This payload stands one line from proving it: `nix path-info --derivation <w>#checks.<system>.<name>` carries a `#` that is part of a **flake ref**, and a guard may legitimately echo a `#` inside a quoted diagnosis. `#` opens a comment only at the **start of a word** and only **outside quotes**. The clause buys V68 at the price of a false RED on correct code.
- **`re.split(r'[;&|]+', …)` is wider than the language, twice.** It splits **inside quotes** — `grep -nE "RESULT:|panic" "$LOG"` is an ordinary spelling of this line and a character split returns `panic" "$LOG"` — and it splits on the `&` of a **`2>&1` redirection**, returning `1`. Both false RED on correct code.

Taking either would have been the ladder's shape again: wider on one axis, narrower on another.

## Fixed as the class

- **`shell_code()`** strips a trailing comment quote-aware, at a word start only, and runs **where `lines` is built** — so no assertion in the block can be satisfied by a comment. That also closes round 18's own `re.search(r"\bERR=\$\(mktemp\b")`, which a comment satisfied just as happily (a third instance of the same class, in my own commit).
- **`last_command()`** scans for an **unquoted** separator (`;`, `|`, `||`, `&&`) instead of splitting on characters. A bare `&` is deliberately *not* a separator — it is the character in `2>&1`.
- The last command is pinned **WHOLE** against a restated constant, not by prefix. That additionally closes a **wrong file argument** (`"$ERR"` is the stderr capture: greping it for `RESULT:` finds nothing and exits 1 on a *healthy* tier) and a **pipeline** (`| head` donates its status) — neither visible to `^grep -n "RESULT:"`.
- Six control assertions in the body pin both overshoot directions and the four separators the scanner must see, so the rejected spellings cannot come back silently.

## FIX_MATRIX

- **`r18/F1`'s claim corrected** — it read *"Now `endswith('exit 1; }')`"*, which overstated what it closed.
- **`r18/F3`'s likewise** — it closed the prefix, past a `;`, not the class.
- Rows **`r19/A1`** and **`r19/A2`** added. `MIN_FIX_MATRIX_ROWS` 99 → 101 at m = 106, counted from an import. `MIN_TESTS` unchanged: no test was added, m is still 128.

---

# Round 18 — the original three findings

## F2 (🟡, PAYLOAD) — the remediation prescribed an action that does not work

The brief told every auditor: *"🔴 **Wrap the block in a function and call it** — `run_tier() { … }; run_tier`. Do NOT simply swap `exit` for `return`…"*

**`exit` inside a function still exits the shell.** Measured under a real pty (`script -qec "<sh> -i" /dev/null`) with a survival marker terminal echo cannot forge — a **file** the surviving shell writes:

| form | `bash -i` | `zsh -i` |
|---|---|---|
| `run_tier(){ …; exit 1; }; run_tier` | no marker — **shell dead**; guard stopped | no marker — **shell dead**; guard stopped |
| `run_tier(){ …; return 1; }; run_tier` | marker `rc=1`; **guard stopped** | marker `rc=1`; **guard stopped** |
| top-level `return 1` (no function) | marker `rc=0`; **block CONTINUED** | marker `rc=0`; **block CONTINUED** |
| `bash <file>` / `zsh <file>` | calling shell alive, block `rc=1`, guard stopped | same |

A **positive control** where the guard PASSES was run too: the "block continued" file *is* created, so the detector can observe both states.

The note's own cited measurement was **correct** — and it was about `return` **at top level**, a different thing from `return` **inside the function it prescribed in the same sentence**. The same edit also **withdrew** the working alternative, "run the block as a script".

Fixed: function **plus `return`**, *or* run it as a script; the top-level refutation kept and scoped to top level. **Nothing pinned this prose** (0 hits for `Wrap the block` / `run_tier` / `INTERACTIVE shell`). Now a named constant pinned **whole and whitespace-normalised**, located in the brief **structurally** — an extractor keyed on the note's own words would fail to find a reworded note and pass.

## F1 (🟡) — the `exit 1` stop-check was a bare substring

`assert "exit 1" in lines[i]`, under a docstring claiming each guard is asserted "as a WHOLE LINE … each terminating in `exit 1; }`". Comments were not stripped. See round 19 above for how the first repair was still incomplete.

## F3 (🟡) — an inverted comment over a negative-only pin

The comment read *"🔴 The verdict grep must NOT be last"*. **That is the inverse** — contradicted by the payload's own comment and by the assertion's own failure message three lines below. A maintainer acting on it reinstates the exit-status inversion verbatim. Corrected. The assertion was also negative-only; widened (and widened again in round 19).

## 🟢 also done

- **🟢4** — `$ERR` pinned to `$(mktemp …)` beside `$LOG`.
- **🟢5** — the constants' comment no longer claims "Each closes a DIFFERENT measured false green", which the emission site 90 lines down already refuted.

**Declined: 🟢6 and 🟢7** — judgement calls, not defects; out of scope per the brief.

---

# Evidence

## Regression matrix

Branch point / mutation base **`10d437c9`**; measured by grafting HEAD's test module onto that script.

| test | at `10d437c9` | at HEAD |
|---|---|---|
| `test_the_stop_note_prescribes_something_that_actually_stops` | **RED** — 1 failed, whole-string mismatch | GREEN |
| `test_the_cached_build_fallback_is_emitted_with_its_guards` (strengthened) | **GREEN** | GREEN |

The second row is deliberate: F1, F3, 🟢4 and both round-19 findings are **width** findings against an existing detector, not regressions — the payload at base is correct. A width claim has no base ref; its evidence is the mutants. Same shape as the existing `r6/4` row against `r5/2`'s detector.

## Mutants

Run under `PYTHONDONTWRITEBYTECODE=1`.

| id | mutation | before | after — killing test @ line, and its own error |
|---|---|---|---|
| V63 | `exit 1` demoted to a trailing comment | SURVIVED @ `10d437c9` | `…with_its_guards` @ `:3013` — *"this guard does not END in `exit 1; }`"* |
| V64 | verdict grep deleted, ends on `echo done` | SURVIVED @ `10d437c9` | same @ `:3057` — *"the block's LAST COMMAND is not, character for character, the verdict grep"* |
| V65 | `; true` after the verdict grep | SURVIVED @ `10d437c9` | same @ `:3057` |
| V66 | fixed `ERR=` while `LOG=` keeps `mktemp` | SURVIVED @ `10d437c9` | same @ `:2985` — *"the stderr path is not assigned from `$(mktemp ...)`"* |
| V67 | stop-note reworded back to "wrap it in a function" | n/a (new target) | `…stop_note_prescribes…` @ `:3202` |
| **V68** | `}   # exit 1; }` | **SURVIVED @ `74cb7409`** | `…with_its_guards` @ `:3013` |
| **V69** | `\|\| true` after the verdict grep | **SURVIVED @ `74cb7409`** | same @ `:3057` |

Positive controls, both directions: V62 (`$DRV` guard deleted wholesale) KILLED in the pre-fix batch at `10d437c9`; V63 and V65 KILLED in the pre-fix batch at `74cb7409` — so each sweep is shown able to go red before its SURVIVEDs are believed. Every kill lands on the **new** assertion, not a neighbour's, and leaves the other 127 tests passing.

Full battery: **✅ 145 rows, all as expected** — no WRONG-KILLER, no EXTRA-KILLER, `POS unmutated copy … 128 passed`.

## Gate

One merged tree, base **`8b173b05`** (`origin/main`), merge **`a08bf109`**. All verdicts read from `RESULT:` lines and collected counts in redirected files, never a piped exit code. Both `nix build`s run **one at a time**.

| tier | verdict |
|---|---|
| `scripts/gate.sh --tier both` (dev-host) | pytest `RESULT: PASS (exit=0)` — collected 22003, failed 0 · node `RESULT: PASS (exit=0)` — 1449/1449 · `GATE: RESULT=PASS exit=0` |
| `nix build .#checks.x86_64-linux.pytests` (sandbox) | `RESULT: PASS (exit=0)` — collected 22003, failed 0 |
| `nix build .#checks.x86_64-linux.nodetests` (sandbox) | `RESULT: PASS (exit=0)` — suites 5, tests 1449, fail 0 |

⚠ **One non-reproducing failure, recorded rather than hidden.** On an *earlier* merged tree (base `f58d2df0`), `gate.sh` reported `1 failed` in `scripts/tests`: `test_tmux_reply_agent.py::test_a_cwd_that_does_not_exist_is_REFUSED_not_opened_in_HOME`. Mechanism: `#{pane_current_path}` expanded **empty** right after `new-window`, so tmux returned `'%3\tscratch20'` — two fields where `NEW_WINDOW_FORMAT` asks for three — and the agent refused a window it had opened correctly. A second full run of the same tier on the same tree returned `failed=0`; it passes in isolation; and it did not recur on the final tree above. Wall time was identical across the two runs (`run=1072s`; `scripts/tests` 577s vs 567s), so it is neither the load signature nor an inflated assertion — a real race in a real-tmux test. This diff cannot reach it (three files, none imported by that test or by `scripts/tmux-reply-agent`). **I did not run a pristine base**, so I am not claiming it pre-exists on `main`.

### DIFF
--- a/scripts/audit-dispatch.py
+++ b/scripts/audit-dispatch.py
@@ -2835,14 +2835,56 @@ CHECKS_DISCRIMINATOR = (
 
 
 # 🔴 The cached-build fallback's guard lines, each a SINGLE named constant so a
-# mutation can remove exactly one and the harness can name which. Each closes a
-# DIFFERENT measured false green; the number of them is deliberately not
-# asserted in the emitted prose (`test -s` speaks to emptiness only and claims
-# nothing about truncation).
+# mutation can remove exactly one and the harness can name which. They do NOT
+# each close an independent false green — the reason each one is here is stated
+# once, at the emission site below, and deliberately not restated here: this
+# comment used to say "Each closes a DIFFERENT measured false green", which the
+# ledger 90 lines down had already refuted ("dropping the run guard alone still
+# leaves `test -s` catching the empty file"). Two comments about the same three
+# constants, disagreeing. The number of them is deliberately not asserted in the
+# emitted prose either.
 NIX_LOG_TMPFILE = 'LOG=$(mktemp -t audit-tier-XXXXXXXX.log); ERR=$(mktemp -t audit-tier-XXXXXXXX.err)   # per-agent: a FIXED path is truncated by every sibling agent'
 NIX_LOG_DRV_GUARD = '[ -n "$DRV" ] || { echo "NO DERIVATION — nix said:"; cat "$ERR"; exit 1; }'
 NIX_LOG_RUN_GUARD = 'nix log "$DRV" > "$LOG" 2>/dev/null || { echo "NO LOG — never built HERE (nix log is per-machine)"; exit 1; }'
 NIX_LOG_EMPTY_GUARD = '[ -s "$LOG" ] || { echo "EMPTY LOG — cannot vouch"; exit 1; }'
+
+# 🔴 HOW TO RUN THE BLOCK WITHOUT CLOSING YOUR SHELL. Pinned WHOLE by
+# `test_the_stop_note_prescribes_something_that_actually_stops`, because a
+# guard on WORDS here is walkable by rewording (claude/RULES.md, spelled-guards)
+# and this note has already shipped one prescription that does not work.
+#
+# 🔴 THE PRESCRIPTION IT REPLACES WAS "Wrap the block in a function and call it"
+# ALONE, with "Do NOT simply swap `exit` for `return`" beside it. `exit` inside
+# a function still exits the shell, so wrapping alone changes NOTHING. MEASURED
+# under a real pty (`script -qec "<sh> -i" /dev/null`) with a survival marker
+# terminal echo cannot forge — a FILE the surviving shell writes:
+#     run_tier(){ …; exit 1;   }; run_tier  -> no marker file:  bash -i DEAD,
+#                                             zsh -i DEAD; guard did stop
+#     run_tier(){ …; return 1; }; run_tier  -> marker file rc=1 in BOTH; the
+#                                             block stopped at the failed guard
+#     top-level `return 1`, no function     -> marker file rc=0 in BOTH, and the
+#                                             line AFTER the failed guard ran
+# The old note's own measurement was CORRECT and was about `return` AT TOP
+# LEVEL — a different thing from `return` inside the function it prescribed in
+# the same breath. Wider on one axis, narrower on another. The top-level
+# refutation is kept, scoped to top level; "run the block as a script", which
+# that edit withdrew, is restored as the other working option (MEASURED:
+# `bash <file>` and `zsh <file>` both stop at the failed guard, exit 1, and
+# leave the calling shell alive).
+NIX_LOG_STOP_NOTE = (
+    "⚠ Each guard echoes its diagnosis before `exit 1`, so the reason "
+    "survives; but a bare `exit 1` pasted into an INTERACTIVE shell closes it. "
+    "🔴 **Two forms stop the block without closing your shell: run it AS A "
+    "SCRIPT (`bash <file>`), or wrap it in a function AND swap `exit` for "
+    "`return` — `run_tier() { … return 1; }; run_tier`.** Wrapping ALONE is "
+    "not one of them: `exit` inside a function still exits the shell, MEASURED "
+    "under a pty in interactive bash and interactive zsh — `run_tier() { … "
+    "exit 1; }; run_tier` killed both. 🔴 The swap works only INSIDE the "
+    "function: `return` AT TOP LEVEL is not a stop, and MEASURED in "
+    "interactive zsh, interactive bash and `bash <script>` the block then "
+    "CONTINUES past a failed guard and prints a foreign log's `RESULT: PASS` — "
+    "the exact false green this block exists to prevent."
+)
 
 def _toolchain_checks_lines(tc):
     """The sandbox tier — only when the flake really declares one."""
@@ -2956,14 +2998,7 @@ def _toolchain_checks_lines(tc):
         'grep -c "panic: test timed out" "$LOG"; grep -n "RESULT:" "$LOG"',
         "```",
         "",
-        "⚠ Each guard echoes its diagnosis before `exit 1`, so the reason "
-        "survives; but a bare `exit 1` pasted into an INTERACTIVE shell closes "
-        "it. 🔴 **Wrap the block in a function and call it** — `run_tier() { … }; "
-        "run_tier`. Do NOT simply swap `exit` for `return`: at top level "
-        "`return` is not a stop, and MEASURED in interactive zsh, interactive "
-        "bash and `bash <script>` the block then CONTINUES past a failed guard "
-        "and prints a foreign log's `RESULT: PASS` — the exact false green this "
-        "block exists to prevent.",
+        NIX_LOG_STOP_NOTE,
         "",
     ]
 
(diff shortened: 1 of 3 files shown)


ZacxDev:
Round 1 (first, full) audit of this PR returned 2 🟡. Both were reproduced at head `74cb7409` before any fix, and the fixes are in `ee201067`.

```audit-claims round=1 audited=74cb7409b4bbb5028c15744838ec8a84ce1f2d79..ee201067
1. The `endswith("exit 1; }")` pin, which a trailing comment `# exit 1; }` still satisfied (V68 SURVIVED at 128 passed), is replaced by comment-stripping at the point `lines` is built — `shell_code()` strips quote-aware at a word start, so no assertion below it can be satisfied by a comment. This also closed a third instance shipped in the same PR: the `ERR=$(mktemp` check was comment-satisfiable too.
2. The verdict-grep pin, which `grep -n "RESULT:" "$LOG" || true` walked because `last_cmd` split on `;` only (V69 SURVIVED at 128 passed), is replaced by `last_command()` scanning for an UNQUOTED separator (`;`, `|`, `||`, `&&`), with bare `&` deliberately excluded as `2>&1`'s character.
3. The last command is now pinned WHOLE against a restated constant rather than by prefix, which additionally closes a wrong file argument (`"$ERR"` exits 1 on a healthy tier) and a trailing pipeline (`| head` donates its status) — neither of which `^grep -n "RESULT:"` could see.
4. Both offered one-line fixes were REJECTED with counter-examples rather than applied: `"#" not in lines[i]` false-REDs on a flake ref (`nix path-info --derivation <w>#checks.<system>.<name>`, one line away in this payload) and on a `#` inside quotes; `re.split(r'[;&|]+', …)` false-REDs by splitting inside quotes and on the `&` of `2>&1`.
5. Six control assertions were added pinning both overshoot directions and the four separators the scanner must recognise, so the rejected spellings cannot return silently.
6. `r18/F1` and `r18/F3` FIX_MATRIX claims were corrected — both overstated what they closed (F1 said "Now `endswith('exit 1; }')`"; F3 closed the prefix past a `;`, not the class). Rows `r19/A1` and `r19/A2` added; `MIN_FIX_MATRIX_ROWS` 99 → 101 at m=106, counted from an import.
```

🔴 Ledger for the NEXT round: this fix round changed **ZERO payload lines** — `scripts/audit-dispatch.py` is untouched (`74cb7409..ee201067` = `test_audit_dispatch.py` 249/32, `mutants-audit-dispatch.py` 55/0; rc 0, silent stderr). Per the attribution gate, one further zero-payload round means the ladder has left the PR and must stop.

PR #1344, opened 2026-09-06
feat(tmux-restore): make ONE reboot answer the boot-race question

ZacxDev:
Rank 1 of `claudedocs/handoff-tmux-restore-chain.md` is **reboot and observe** — the only test
that closes both open investigations. This does not reboot anything. It removes the way that
one shot gets wasted.

## Why

The comparison the post-boot state has to be made against is the **pre-reboot tmux inventory**,
and that stops existing the moment the machine goes down. So does the session holding it in
context. `pre` writes it to `~/.cache/tmux-restore-observe/` — outside the repo, so neither the
reboot nor a branch switch can take it away; `post` re-reads the same facts plus the unit's
journal for that boot and renders a verdict against it.

It is a **READER**. It starts nothing, orders nothing and adds no boot-path unit — deliberately:
three defects in this arc were introduced by changing a boot path nobody had observed, so the
observation must not perturb what it observes.

## What the verdict REFUSES to say is the load-bearing part

- **Identical `boot_time` across the pair ⇒ INCONCLUSIVE**, never "no race". Two captures of the
  same boot are not evidence about a boot, and that is the exact shape of a green that observed
  nothing.
- **Either side UNMEASURED ⇒ INCONCLUSIVE**, never a comparison against a missing number.
- **Duplicate `(session, window_name)` pairs are their own arm**, asserted with the counts
  deliberately EQUAL — continuum replacing one window while the script creates another leaves
  the total unchanged, so a count-only check is blind to exactly the signature it is hunting.
  The inventory is TAB-separated because agent-generated window names contain `;`.
- **Fewer windows than the layout is a restore failure, not the duplication race**, so it gets
  its own rc rather than being folded into a clean pass.
- **rc 0 states in its own output** that one clean boot is ONE negative sample of an unguarded
  timing assumption, not a closure.

rc `0` clean · `1` race evidence · `3` could-not-decide · `4` windows missing.

## The second commit is the gate catching me

`test_no_real_launchers.py` pins the `systemctl` acknowledgement against the tree in **both**
directions, so a NEW file reaching an acknowledged binary reds the suite instead of being
absorbed by an existing entry's sentence. It fired. The script is acknowledged on the same
ground as `syshealth` — the **VERB**: `systemctl --user show …`, and `show` is on
`nolaunch.SYSTEMCTL_READ_VERBS`, so the verb-splitting stub passes it through as a read.

That justification is pinned rather than asserted: `test_every_systemctl_call_site_in_the_script_uses_a_READ_verb`
reads every invocation-shaped occurrence's positional verb out of the script and checks it
against `SYSTEMCTL_READ_VERBS` itself. The extractor is scoped to invocation shape, which is
what excludes the script's own `reason=systemctl-show-returned-nothing` diagnostic — that is
deliberately the fix rather than rewording the message, because rewording to dodge a scanner is
how a justification stops describing the tree it claims to.

## Verification

**13 tests**, driven through a `verdict` subcommand against fixture captures — no reboot, no
real tmux, no systemd.

**Mutation-swept, 10 mutants, each killed by the test named for it**, control green either side,
every file restored byte-identical:

| mutant | killed by |
|---|---|
| drop the same-boot refusal | `test_an_identical_boot_time_refuses_to_render_a_verdict` |
| drop the duplicate arm | `..._a_duplicate_pair_is_race_evidence_even_when_the_counts_MATCH` |
| split inventory on `;` instead of TAB | same |
| treat below-layout as clean | `..._fewer_windows_..._is_its_own_finding_not_a_clean_pass` |
| drop the UNMEASURED guard | `..._an_unmeasured_side_is_inconclusive_never_a_comparison` (both params) |
| drop the one-sample caveat | `..._a_clean_boot_says_so_AND_says_it_is_only_one_sample` |
| drop the baseline-required check | `..._post_without_a_baseline_refuses_...` |
| script `show` → `restart` | `..._every_systemctl_call_site_in_the_script_uses_a_READ_verb` |
| new file removed from the ledger | `test_every_hazardous_binary_the_scripts_reach_is_stubbed_or_acknowledged` |
| extractor regex blinded | BOTH the positive control and the extractor control |

**Both tiers, on the merged tree** (`origin/main` = `254b487d`, unmoved; branch head is the
merge):

- dev host — `nix develop … -c bash scripts/gate.sh --tier both` → `GATE: RESULT=PASS`
  (pytest `21882 collected / 21879 passed / 3 skipped / 0 failed`, floor 18610; node
  `1449/1449`, floor 1367)
- sandbox — `nix build .#checks.x86_64-linux.pytests` and `.nodetests` **one at a time**
  (a combined red is not trustworthy): both `RESULT: PASS (exit=0)`, same counts.

The pre-reboot baseline was captured live before the first commit and is recorded in the
handoff's rank 1.

https://claude.ai/code/session_019o7RBCqvjQgnYyVTFm8LTB


ZacxDev:
## Round 1 — fixes landed

Round 1 measured the instrument against the operator's live workspace and found it returns `RACE EVIDENCE` on a **perfect restore**. The discriminator was redesigned rather than patched. Independently re-verified before acting: 9 duplicate `(session, name)` groups in the real baseline, 8 live.

```audit-claims round=1 audited=95478474..1231117d
1. F1 — the duplicate arm no longer keys on window NAMES; the discriminator is (session, window_index) against the layout ACTUALLY REPLAYED (newest resurrect save older than this boot, resolved at post time).
2. F2 — the missing-windows arm is no longer nested under `if rc == 0`; it is reported independently of the race arm and only weakens the rc when nothing worse fired.
3. F3 — the expectation is no longer taken from the baseline's layout_windows, so the baseline cannot go stale; it now supplies only the previous boot_time and the host identity.
4. F6 — misplacement is detected by comparing each (session, window_index)'s working directory between the replayed layout and the live workspace.
5. F4 — a unit that never ran is stated on its own line naming the OnActiveSec=45s timer, instead of leaving `Result=success ExecMainStatus=0` to be read as success.
6. F5 — no tmux server at all returns rc 5 TOTAL RESTORE FAILURE instead of rc 3 INCONCLUSIVE.
7. F7 — captures record host_machine_id and host_name, and a verdict across two different hosts refuses.
8. F8 — `pre` installs the ~/.cache copy of the script that the handoff advertises, so that path has a producer.
9. Nits — the INVENTORY header no longer claims a `;` separator; the dead `= UNMEASURED` string comparisons are gone; `pre`/`post` fail if the capture wrote nothing; rc 2 and rc 5 are named in the usage legend and pinned by a test; emit_plan writes its whole block or none of it.
10. NEW (found while writing the tests, not by the audit) — `exp` is an awk BUILT-IN, so the misplacement arm was a syntax error emitting nothing while the verdict read NO RACE OBSERVED; renamed to want/got.
11. NEW — the pane cwd is NOT at a fixed field index (measured: 49 lines at field 8, 5 at field 7, all 11 fields); it is located by content, and pane lines it cannot read are COUNTED and disclosed rather than silently dropped.
12. NEW — `uptime -s` is a procps extension that returns nothing under coreutils' uptime (which shadows it in this repo's dev shell), leaving boot_time EMPTY; boot time now comes from /proc/stat btime, verified byte-identical on this host.
13. NEW — replayed_layout seeded its best-so-far at 0, so a save whose mtime is exactly the epoch could never be selected; seeded below every possible mtime now.
14. The `moved_rc` awk-failure branch is deliberately left UNPINNED and says so in the source: it is unreachable today because section() is also awk and its failure trips the vacuity guard first. The real degradation path is pinned by test_a_broken_awk_degrades_to_INCONCLUSIVE_not_to_clean.
```

**Verification.** 26 tests. Six controls driven through the real script on real data — including the negative control that was missing in round 1: a healthy 54-window workspace with 9 duplicate-name groups now returns rc 0. Extra window → 1, missing → 4, misplacement → 1, no server → 5, unit-never-ran → 0 with the notice.

**Two corrections to round 1's own mutation scoring**, stated because it was reported as evidence: two mutants had deleted a closing `fi` and died of a syntax error rather than of the guard — rescored INVALID and redone isolated; and one genuinely **SURVIVED** — `capture` no longer emitting a field `verdict` reads — because both halves were pinned and the seam between them was not. Closed by `test_capture_post_EMITS_the_fields_the_verdict_READS`, which drives a real capture against a fixture resurrect dir.

ZacxDev:
## Round 3 fixes landed — and round 4 was NOT run

Merging on the operator's explicit instruction with the audit ladder **still open**. Recording that plainly, because a merged PR with two audit rounds on it otherwise reads as audited-clean, and this one is not.

**What round 3 changed** (round 2's two 🔴, three 🟡, five 🟢 — all reproduced on the shipped script before fixing):

- 🔴 The verdict blamed the boot race for what `tmux-session-restore.py` itself does. It is a *second writer*: `new-window -t <sess>:<idx>` and `send-keys "cd <plan cwd> …"`. The expectation came from the replayed layout alone, so legitimate restore work read as `RACE EVIDENCE` — measured at 46 min plan/layout skew: 1 false race id + 4 false misplacement rows. The capture now carries the plan's ids and cwds; a window in neither layout nor plan still fires.
- 🔴 An unmeasured cwd comparison rendered as `in the right place`, rc 0 — including the case where `section()` strips the emitter's own `# UNMEASURED — no tmux server responding` marker as a comment. Both now return rc 3.
- 🟡 `ExecMainStatus` was never branched on. Measured on this host: `Result=success` with `ExecMainStatus=1`. Now its own `RAN AND FAILED` line.
- 🟡 The seam test's docstring claimed it pinned "every key and section the verdict consumes" while stubbing tmux/systemctl to *fail* — three verdict-deciding emits were unpinnable against it and their mutants survived. Scoped docstring + a second fixture with succeeding stubs.
- 🟡 The handoff said the cached script copy "cannot rot" four lines before saying re-running `pre` is not required. Scoped to what is true.
- 🟢 `section()` re-opened its block at any later matching header (a JOURNAL line spelling `--- OBSERVED-IDS` injected a row and produced a false race); the `last`-link note asserted an unmeasured cause; the awk no longer discards stderr; the `moved_rc` unreachability claim narrowed to "no fixture reaches it"; `[ -s "$out" ] || die` is now pinned; the duplicate-group count (9 → 10 during the audit) is stated as volatile rather than as the claim.

**Verification.** 37 tests. Mutation sweep: every mutant killed by the test named for it, control green either side, script restored byte-identical — including all three that SURVIVED round 2.

Both tiers, on the merged tree (`origin/main` = `c19fb289`, confirmed an ancestor of `4a986ee8` immediately before merging):

| tier | pytest | node |
|---|---|---|
| dev host (`gate.sh --tier both`) | `RESULT: PASS` | `1449/1449`, floor 1367 |
| sandbox (`nix build`, one derivation at a time) | `21976 collected / 21973 passed / 0 failed`, floor 18672, 0 timeout panics | `1449/1449`, floor 1367 |

🔴 **The residual.** Rounds 1 and 2 each found real 🔴s in the *preceding* round's work, and round 3 substantially reworked the discriminator — so the code being merged is the first version no audit has seen. The instrument's own controls were re-run on real data (a healthy 54-window workspace with duplicate names returns rc 0; extra → 1, missing → 4, misplacement → 1, no server → 5), but that is not the same as an adversarial round. If the one reboot produces a surprising verdict, suspect the instrument before the race.

⚠ Also: `rerere` recorded this branch's resolution of the `ACKNOWLEDGED_UNSTUBBED["systemctl"]` conflict. It is correct for that exact conflict and will now replay automatically — worth checking if that ledger conflicts again, since a replayed resolution from a different merge is how a wrong two-way result gets written.

PR #1351, opened 2026-09-07
fix(tmux-restore): REFUSE when there is no tmux server, instead of manufacturing one systemd then kills

ZacxDev:
## What changed and why

This PR was opened on a **refuted** diagnosis. A round-1 audit caught it, the
mechanism was re-derived from the journal and confirmed by an isolated
experiment, and this revision replaces the fix, the prose and the tests
accordingly. The old title ("the resumes were sent 24s before the panes
existed") described the wrong mechanism and has been changed.

### The refuted story

> The unit fired at boot+63s and sent 43 `claude --resume` lines; resurrect
> created every pane shell at boot+87s, so the sends landed in panes that were
> not ready and were DISCARDED.

The only evidence for it was an **absence** — the sent text appears nowhere in
any pane's scrollback. That observation is equally consistent with a second
story, so it cannot choose between them.

### What actually happened

`Started tmux child pane N launched by process <pid>` arrived in **two
cohorts**: 17 lines at the unit's own timestamp naming the pid the unit
started, then 56 lines 24s later naming a **different** pid. Two servers, not
one unready one.

On a cold boot nothing else starts tmux, so `cmd_restore`'s own
`tmux new-session` created the server — **inside the unit's cgroup**. The 43
sends were delivered successfully. Then `ExecStart` returned and, with
`Type=oneshot` / `RemainAfterExit=no` / `KillMode=control-group`, systemd tore
the cgroup down, taking the server and every claude process with it. The unit
reported `Result=success`.

Confirmed by experiment on a private `-L` socket, real server untouched:

| unit config | server after ExecStart returns |
|---|---|
| `Type=oneshot`, no `RemainAfterExit` | **gone** — `no server running on …` |
| `Type=oneshot` + `RemainAfterExit=yes` | **survives** |

So a duration was never the variable, and no amount of waiting could have
fixed it.

## The fix

`no_tmux_server_to_restore_into()`, checked **before** the send loop — because
the loop's own `tmux new-session` is the destructive step. With no server,
refuse: a server this process creates cannot outlive it.

**The refusal exits 0, and that is deliberate.** The unit is
`OnFailure=notify-failure@%n`, and that toast bypasses DND. Until the unit
triggers on the tmux socket appearing rather than a fixed `OnActiveSec=45s`,
"no server" is the **standing** cold-boot state — a non-zero exit would fire
the alarm on every boot forever, which `nix/home.nix` explicitly calls out as
re-breaching the bypass. It logs loudly instead, and says what to re-run.

The settle-wait and `_verify_sends` from the original revision are **kept**,
re-described honestly as a secondary guard and a detector rather than as the
fix.

## 🔴 Scope — what this does NOT do

This stops the destruction and makes the failure legible. It does **not** make
cold-boot restore work: after this, a cold boot refuses and the operator still
has to re-run the restore once attached. Making it actually restore needs the
socket trigger in `nix/home.nix`, which is deliberately not in this PR.

`RemainAfterExit=yes` alone is measured to work and is **not** the answer —
with `KillMode=control-group` the unit would then own the operator's server,
and stopping the unit would kill the workspace.

## Audit round-1 follow-ups, all fixed

- **`[ "$live" = UNMEASURED ]` never matched production.** The emitter writes
  `UNMEASURED reason=no-tmux-server-responding` and `get` returns everything
  after `key=`. Measured by reverting the fix: the failed `-lt` does not abort
  — the shell prints `integer expected`, the test evaluates false, and control
  falls into the `else`, which printed
  `resumes: UNMEASURED reason=... pane(s) running claude vs 43 send(s)` and
  returned **RC_CLEAN**. The path whose whole job is to say "I do not know"
  returned a confident pass. Now a `case` prefix match.
- **The existing test could not see that bug** — its fixture used a bare
  `UNMEASURED`, a value production never writes. Added a test using the real
  emitted string; kept the old one.
- **Both exit-1 branches in `cmd_restore` were untested** (mutants survived).
  Pinned, plus a settled-and-landed control so an unconditional `return 1`
  fails.
- **`claude_panes_live` is a whole-host count** — includes hand-started panes
  and ones the unit skipped as already running. The verdict now prints that
  caveat, so `live >= sends` is not misread as "every send landed".
- **Four tests assigned `tsr.pane_fingerprint` with no teardown**, leaking
  module state into whatever ran next. Now `monkeypatch`.

## Prose corrections

The refuted mechanism was asserted at four sites, all corrected:
`tmux-session-restore.py` (settle docstring, `_verify_sends` stderr) and
`tmux-restore-observe.sh` (arm comment, `RC_RACE` message). The failure
message now names the **observation** ("these panes are not running claude")
rather than a cause it cannot determine.

**This branch's rewrite of the handoff doc is dropped.** `main`'s corrected
ANSWERED block stands; the merge conflict was resolved by taking main's copy
wholesale.

## Verification

- Mutation battery, 6 mutants, positive control green, each killed by **its
  own** named assertion (not a neighbour's): refusal removed; predicate
  inverted; refusal exits 1; unsettled penalty dropped; `UNMEASURED*` reverted
  to equality; whole-host caveat removed. Run under
  `PYTHONDONTWRITEBYTECODE=1`.
- Every new `cmd_restore` test monkeypatches the server predicate rather than
  reading the real tmux, because the dev-host tier **has** a server and the nix
  sandbox has **none** — a test that reads the real thing asserts a different
  branch in each tier and cannot fail in one of them. This arc has already
  shipped two defects through exactly that gap.
- Both tiers run on the merged tree; see the comment below for the base sha
  and results.


ZacxDev:
## Gate results — naming the tier and the base sha

Merged tree: **`83697d30`** (branch merged with `origin/main` at `969f0581`). Both nix
derivations were built **one at a time** — a combined invocation produces false failures
through store contention, and a combined red is not trustworthy.

| Tier | Derivation / suite | Verdict |
|---|---|---|
| nix sandbox | `checks.x86_64-linux.pytests` | **PASS** — `RESULT: PASS`, collected 22026 / passed 22023 / failed 0 |
| nix sandbox | `checks.x86_64-linux.nodetests` | **PASS** — `RESULT: PASS`, 5 suites / 41 files / 1449 tests / 0 fail |
| dev host | `gate.sh --tier both` on `3b348542` | **PASS** — pytest 22023 passed / 0 failed; node 1449 / 0 fail |
| dev host | `gate.sh --tier both` on `83697d30` | **PASS** — `GATE: RESULT=PASS`, pytest 22023 passed / 0 failed; node 1449 / 0 fail |

`scripts/session-analysis/tests` collected **562** against the floor of **525** this PR
raises.

**All four tier/tree combinations are green on `83697d30`.** `gate.sh`'s exit status
agreed with both runners' own `RESULT:` lines, so this is not an rc-90 could-not-vouch.

🔴 **Read out of `nix log`, not the exit code.** A `nix build` sends its test output to the
daemon and prints none of it; a derivation that ran *zero* tests exits 0 identically. Both
verdicts above are the runners' own `RESULT:` and `TOTAL` lines.

⚠ **`nix build path:<a worktree>` does not work here** and fails `exit=2` *before any test
runs*: a worktree's `.git` is a FILE pointing at the real git dir, which defeats the
sandbox's git-config isolation. Extract with `git archive HEAD | tar -x -C <dir>` and build
from there.

⚠ **A merged-tree gate is a claim about the tree it ran on.** `main` moved again after
`83697d30` — including real test changes, not just docs — so whoever merges this should
re-gate the tree the merge actually creates.

## Mutation battery

6 mutants, positive control green, each killed by **its own** named assertion rather than a
neighbour's, run under `PYTHONDONTWRITEBYTECODE=1` (a same-length edit landing in the same
second as the last import is invisible to CPython's mtime-in-whole-seconds cache, and the
mutant is scored SURVIVED without ever executing).

| Mutant | Killed by |
|---|---|
| refusal block removed | `test_a_restore_with_no_tmux_server_REFUSES` — "waited instead of refusing" |
| `has-session` predicate inverted | `test_the_server_guard_reads_has_session_and_treats_failure_as_absent` |
| refusal returns 1 instead of 0 | `test_a_restore_with_no_tmux_server_REFUSES` — "a standing cold-boot state must not fire the OnFailure toast" |
| unsettled penalty dropped | `test_an_unsettled_workspace_exits_1_even_when_every_send_landed` |
| `UNMEASURED*` reverted to equality | `test_the_unmeasured_marker_AS_PRODUCTION_SPELLS_IT_is_not_an_integer_comparison` |
| whole-host caveat removed | `test_the_verdict_says_the_live_pane_count_is_WHOLE_HOST` |

The `UNMEASURED` mutant is worth reading: reverting the fix does **not** produce a loud
error. The shell prints `integer expected`, the `-lt` test evaluates **false**, and control
falls into the `else`, which rendered the reason string *as a pane count* and returned
`RC_CLEAN`. The arm whose entire job is to say "I do not know" returned a confident pass —
worse than the "integer-expected error" the round-1 audit described.

PR #1357, opened 2026-09-07
docs(handoff): ranks 34/39 and 41 closed — the browser tier authenticates, and the `=` prefix is exercised

ZacxDev:
Closes tmux-webapp ranks **34** and **39** — they were one item.

`ZacxDev/homelab-infra#743`, squash `3f6ef8ae7`, live as **0.8.29**.

Closed by **authenticating**, not by disarming: the free-form send-keys box and the tool-input disclosure are still armed, now behind a login.

Verified by re-running the ORIGINAL probe against the live pod, not by reading a rollout:

| probe | 0.8.28 | 0.8.29 |
|---|---|---|
| uncredentialed `POST /ui/term/send-keys` | 200, host agent ran it | **401** |
| anonymous `GET /` | 200, 147 KB | **303** to `/login` |
| anonymous `GET /session/<id>` | 200 w/ tool inputs | **303** to `/login?next=` |
| correct password | — | 303 + cookie, dashboard 200 |
| wrong password | — | 401, no cookie |
| `/health`, `/metrics` | 200 | 200 |

The machine tier was proven unharmed rather than assumed: both host agents active, 0 backing-off lines, and a bounded control write to a nonexistent pane was claimed and attempted within seconds.

Docs-only.

### DIFF
--- a/claudedocs/handoff-tmux-webapp.md
+++ b/claudedocs/handoff-tmux-webapp.md
@@ -1072,11 +1072,41 @@ drop, so a typo’d rank can no longer collapse two items onto one lock in silen
     and `#1353` (pre-deploy handoff). `#1353`'s only red is the contention flake
     under "Open investigations"; confirm the re-run before merging.
     forcing: none
-41. **Rank 31's `=` session-target prefix has still never run against real tmux.**
-    Rank 33's brief conflated it with the reply path, which targets a pane id and
-    does no name matching. `=` is used only by `new-window -t "=" + tmux_session +
-    ":"` (devrc `scripts/tmux-reply-agent` line 568) — the start-a-session path.
-    The live server has `scratch2` beside `scratch20`, so the hazard is reachable.
+41. ✅ **DONE 2026-09-07 — EXERCISED AGAINST REAL tmux, and the control reproduced the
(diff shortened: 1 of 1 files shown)


ZacxDev:
Added **rank 41** to this PR (`9fe5895a`): the `=` exact-match session target is now exercised against real tmux through the agent's own `open_window()`, not just as a raw-`tmux` measurement.

| probe | result |
|---|---|
| raw `new-window -t <SHORT>:` (no `=`) | rc 0, **landed in `<LONG>`** — wrong session |
| agent `open_window(cwd, <SHORT>)` | **refused**, `can't find session` |
| agent `open_window(cwd, <LONG>)` | pane `%61`, confirmed in `<LONG>` |
| windows in probe session | 1 → **3** exactly (refusal created nothing) |
| teardown | 22 sessions before, 22 after, no strays |

No code change — `=` was correct as written; the evidence was what was missing.

⚠ **The item's own premise had gone stale:** it said `scratch2` sits beside `scratch20` so the hazard is reachable. Both now exist, so that pair resolves exactly and is *not* a live hazard. The probe built its own prefix-ambiguous pair instead.

PR #1360, opened 2026-09-07
docs(handoff): rank 42 built and in flight as homelab-infra#749 — two guards walkable by a NAME, one banner carrying the DSN

ZacxDev:
Records rank 42 of the tmux-webapp arc: all three UI residuals closed in `ZacxDev/homelab-infra#749` (OPEN at time of writing).

**Status** replaced (0.8.28 → the current 0.8.29 authenticated posture), with the "loop is closed and proven" evidence carried forward across the replace after the write gate flagged it. **Open investigations** and **Gotchas** appended.

Worth reading even if you skip the rest:
- Two clawgate guards, both with docstrings asserting they were unwalkable, were walkable by a **name** — `aaOffScreen` by a CSS unit, `aaEvictsIn` by an import alias. Both read a *derived* surface while a *defining* one sat next to them.
- Redacting the pgx DSN out of the operator banner would have **destroyed the only copy** of the driver error: the three toggles that build that notice logged it nowhere.
- The devrc clone was 2 commits behind on this doc, so the working copy still described 0.8.28 and an open rank 39. Read the doc from the ref, or fast-forward first.

---

https://claude.ai/code/session_01Mi9GPkd2dFWCBcEPCk8v1q

### DIFF
--- a/claudedocs/handoff-tmux-webapp.md
+++ b/claudedocs/handoff-tmux-webapp.md
@@ -9,23 +9,31 @@ and an **attention queue** that surfaces sessions needing a human so Zach can ju
 
 ## Status
 
-🔴 **SHIPPED AND VERIFIED LIVE — clawgate `0.8.28`, pod `clawgate-c8f7c4f6f-qhkqr`,
-0 restarts.** The whole arc is done: rank 32 armed the host agent, rank 33 drove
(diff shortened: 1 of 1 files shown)


ZacxDev:
Superseded by the arc-wide handoff, which contains this commit plus the full record of the tmux-page round.

PR #1370, opened 2026-09-07
feat(nix): classify a host's /tmp-churn state instead of reading "nothing to do" as "patched"

ZacxDev:
`apply-tmp-churn-stale-lines-2026-09-04.sh --dry-run` exits **0** with
"neither edit applies: no stale header, no dead rule" on a host that has never
run the retention script at all. The script is honest about it — it prints the
discriminating grep itself — but the exit code and the wording both read as
success, which is the shape `claude/RULES.md` names: a wrapper that reports
"nothing to do" instead of erroring is how a green gets believed.

This runs that grep, plus the other reads that decide whether
`apply-tmp-churn-retention.sh` can work on a given host, and classifies:

```
applied-and-live      rules in config AND live      -> nothing to do
applied-not-rebuilt   rules in config, none live    -> needs a switch
unapplied-can-apply   no rules, anchor present      -> the script can run
unapplied-no-anchor   no rules, no anchor           -> REFUSE (exit 3), or --init
unapplied-no-script   retention script unusable     -> REFUSE (exit 3)
```

Read-only by default; `--apply` edits and switches; `--init` creates the
attribute on a config that has none.

---

## 🔴 STATUS CORRECTION — supersedes the commit messages on this branch

The first commit message says *"NOT verified: `--init` has never run against a
real `/etc/nixos`"*. **That is out of date.** `--init` **has** since run live on
the operator's laptop — 2026-09-07 17:54, generation 247, 7 rules verified live.
Commit messages cannot be edited, so this section is the authority.

**That live run was safe on two axes by luck, not by check**, and both are now
checked:

| axis | why the run survived | now |
|---|---|---|
| the retention script's version | the laptop's devrc was at `38bd8edd`, which *contains* `d8fe0bce` — confirmed by the absence of any `bak-tmp-churn-*` | probed before execution |
| the config being a symlink | the laptop's `configuration.nix` is a regular file | `cp -aL`, and refused-through checks |
| the config's mode | it was already `0600`, so the mode-clobber changed nothing | `cat >` preserves mode and inode |

---

## Round-1 audit rework

A round-1 audit returned 3 🔴 and 9 🟡. Every one is addressed, and every one now
has a fixture test that was **watched red at `d5421b4d` and green at HEAD**.

### 🔴 1 — READ-ONLY mode could rewrite `/etc/nixos` as root

`expected=$(bash "$RETENTION" --emit-rules 2>/dev/null | ...)` ran on the
**no-argument** invocation the header documents as *"READ-ONLY report, changes
nothing"*, against a path derived from a `$SUDO_USER` lookup whose version was
never probed — the `LSOF_BIN` shape from #1366. Measured here: `06597151`
(2026-08-23, the revision before `d8fe0bce`) contains **zero** occurrences of
`--emit-rules` **and zero** of `TMP_CHURN_CFG`, and has no argument handling at
all, so that call performs the full config edit with every diagnostic swallowed
by the command substitution.

Fixed by probing for the `--emit-rules)` **case arm** — not for a mention of the
string, which a comment would satisfy — and then validating that what it printed
is actually a set of `/tmp` rules. The ledger is now read **once** and reused, so
`$RETENTION` is executed at most once per run instead of twice.

### 🔴 2 — `RESULT: PASS`, exit 0, over a completely inert edit

The verify asserted `after_live -eq expected`: a count over the **whole** live
tmpfiles config against a number from the repo ledger. Reproduced at
`d5421b4d`, in a fixture: seven *unrelated* `mM:7d` rules already live and an
activation that changes nothing printed `RESULT: PASS` and exited **0**.

Replaced by two independent required claims:

* **IDENTITY** — every rule we intend to apply is matched **as a string** in the
  live config. A count of unrelated rules can no longer mask it.
* **DELTA** — the live count moved by exactly `|want| - already`, where `already`
  is measured *before* the edit so a rule that legitimately pre-exists is not a
  false red.

### 🔴 3 — a symlinked config made the backup worthless

`cp -a` implies `-d`. Measured: on a symlinked `configuration.nix` the backup was
a second symlink to the file about to be edited, the write went through the link,
and the restore was a no-op — config left edited-and-broken with no recoverable
copy. Now `cp -aL` for every backup this file takes, and `cat "$new" > "$CFG"`
for the write.

🔴 **This also makes `apply-tmp-churn-retention.sh`'s documented rollback a lie on
such a host** — its `cp -a "$CFG" "$BAK"` has the same defect. That file is out
of scope here, so `--apply` takes its own dereferenced backup before delegating
to it, and the report says so.

### The nine 🟡

| # | finding | fix |
|---|---|---|
| 4 | *"the restore is itself re-parsed"* was false — under `set -e` a failing `cp -a` exits first | restore is guarded; a failed restore now says so and names the backup |
| 5 | `case "${1:-}"` ignored later args, so `--init --dry-run` did the full apply | branch on `$#`, matching `apply-tmp-churn-retention.sh:230` |
| 6 | `TMP_CHURN_CFG` was honoured for reads and the edit, then `nixos-rebuild switch` ran anyway | honoured fully: no rebuild, and no root requirement — which is what makes the whole file testable |
| 7 | `getent passwd <unknown>` exits 2; under `set -e`+`pipefail` the run died at exit 2 with zero output | diagnosed; an empty/relative home field is refused rather than becoming an absolute path root would `bash` |
| 8 | `applied-not-rebuilt` + unreadable retention script → `expected=0` → 🔴 FAIL on a fully successful operation | in that state the expectation comes from the **config**, not the ledger |
| 9 | `$CFG` was never established as the file the host evaluates | a `flake.nix` beside it is reported, and blocks `--init` pending `TMP_CHURN_FLAKE_ACK=1` |
| 10 | precondition 1 was file-scoped while `systemd.tmpfiles.rules` is a `listOf` merged across modules | scans every `*.nix` under the config directory (the same-file case stays caught by `nix-instantiate --parse`) |
| 11 | no tests, next to a 355-line harness for the sibling | `scripts/tests/test_preflight_tmp_churn_host.py`, 54 tests |
| 12 | `cp -a "$new" "$CFG"` put the mktemp's `0600` onto a `0644` config on **success** only | `cat "$new" > "$CFG"` — measured to preserve mode and inode |

Nits: `--help` is now explicit text that names `--init` (it previously printed a
line-range of the header and mentioned `--init` **zero** times); `nix-instantiate`
stderr is shown so exit 127 is distinguishable from a syntax error; the trailing
`systemctl` is guarded so it cannot turn `RESULT: PASS` into exit 1; the
`=== applying (this EDITS …) ===` banner no longer prints on the path that edits
nothing; both temp files are on one EXIT trap.

## Testing

`scripts/tests/test_preflight_tmp_churn_host.py` drives the **real script**
end-to-end against fixture hosts — no extracted fragment, no namespaces, no root,
no skips. `TMP_CHURN_CFG` supplies the config; `systemd-tmpfiles`,
`nixos-rebuild`, `nix-instantiate`, `systemctl`, `hostname` and `findmnt` are
stubbed on `PATH`.

The `systemd-tmpfiles` stub is **stateful** — call 1 returns the before state,
call 2+ the after state — because a stub with one fixed answer cannot tell a
working activation from an inert one, which is exactly 🔴 2.
`test_the_live_stub_is_actually_reached` is the positive control that the script
calls it at all.

**Red-before is reproducible**, and is how each guard above was checked:

```
git show d5421b4d:nix/system/preflight-tmp-churn-host.sh > /tmp/base.sh
PREFLIGHT_SCRIPT_UNDER_TEST=/tmp/base.sh PREFLIGHT_RUN_PREFIX='unshare -r' \
  python -m pytest scripts/tests/test_preflight_tmp_churn_host.py
```

The `unshare -r` prefix is required because the old script demanded root even
under `TMP_CHURN_CFG` (finding 6); without it every case would go red on the root
check instead of on the defect it is about — red for the wrong reason, which
proves nothing.

**Matrix, measured:**

| tree | result |
|---|---|
| `d5421b4d` (base), under `unshare -r` | **37 failed, 16 passed, 1 skipped** |
| HEAD | **54 passed, 0 skipped** |

The 16 green at base are invariant guards and positive controls, and each says so
in its own docstring — e.g. `test_init_preserves_the_configs_mode[0o600]` is the
mode the operator's laptop happened to be in, where the defect changed nothing;
the `[0o644]` case is the one that can tell a fix from the bug, and it is red at
base.

A mutation sweep over the new guards found the **delta** assertion SURVIVED every
fixture: the identity check killed them all first. `test_an_activation_that_
drops_an_UNRELATED_rule_does_not_pass` was written to kill it, and does.

**Both tiers, measured one at a time, counts not verdicts:**

| tier | `scripts/tests` | total |
|---|---|---|
| dev host (`run-tests.sh --targets scripts/tests`) | collected=12898 passed=12891 skipped=0 **failed=7** | floor 12848 accepted |
| hermetic (`nix build .#checks.x86_64-linux.pytests`) | collected=12898 **passed=12898** skipped=0 floor=12848 | **collected=22057 passed=22054 skipped=3 failed=0, exit 0** |

The dev host's 7 are the known toolchain drift another branch is fixing —
`test_every_decrypt_family_VERDICT_is_pinned_WHOLE`,
`test_the_LEAKING_manglers_really_DO_make_age_keygen_echo_the_secret`,
`test_resolve_recipient_NEVER_quotes_age_keygens_INPUT_ECHOING_stderr` (age) and
`test_engine_is_the_version_every_measurement_is_keyed_to` (opencode). None is in
this diff, and all four pass in the hermetic tier.

🔴 **The full-target run is also what caught this suite's own two-tier defect**:
its first draft wrote stubs with an env-based shebang, which execs on a NixOS dev
host and ENOENTs in the build sandbox. `scripts/tests/test_runtime_shebangs.py`
went red on the first gate run; the stubs now go through
`testlib.mockbin.write_exec` with POSIX-sh bodies. Running the new module alone —
which was green — could not have seen it.

Registration: `scripts/tests` is already a `HERMETIC_TARGETS` directory target, so
the new file needs no entry there; the `TARGET_FLOORS` line is the only evidence
the gate runs it at all. `12793 -> 12848`, which is `_suggested_floor 12898`
sourced out of `run-tests.sh` itself and fed the count the gate printed.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


ZacxDev:
## `--init` has now run against a real `/etc/nixos` — the commit message's caveat no longer holds

`d5421b4d`'s message says *"NOT verified: --init has never run against a real
/etc/nixos."* That was true when written and is now false. Correcting it here,
since the commit message cannot be.

Ran on the laptop (`192.168.50.155`), 2026-09-07 17:54 CDT, from
`unapplied-no-anchor` — the state the mode exists for:

- both preconditions passed at runtime: no `systemd.tmpfiles` anywhere, and the
  last non-blank line was a bare `}` (line 564)
- backup written to `/etc/nixos/configuration.nix.bak-20260907-175418` —
  independently confirmed afterwards as **16,722 B, mtime Aug 24**, i.e. byte-size
  and mtime identical to the pre-edit file measured before the run, so `cp -a`
  preserved the original rather than stamping a fresh copy
- `nix-instantiate --parse` clean; 7 rules inserted
- `nixos-rebuild switch` — `Checking switch inhibitors... done`, built 8
  derivations, started `systemd-tmpfiles-resetup.service`
- self-verify reported `RESULT: PASS — 7 live, 0 stale`

**Verified independently of the script**, over ssh, reading the live config rather
than trusting the run's own verdict:

| check | result |
|---|---|
| `systemd-tmpfiles --cat-config \| grep -c 'mM:7d'` | **7** |
| `... \| grep -c ' m:7d'` | **0** |
| the rule lines themselves | the same 7 as the workbench |
| `/run/booted-system` vs `/run/current-system` | **differ** — the switch took, no reboot |
| `systemctl is-active systemd-tmpfiles-clean.timer` | `active` |

Not verified: `NextElapseUSecRealtime` returned empty, so the timer is confirmed
active but its next-fire time is relayed from the run's own output
(`Mon 2026-09-07 18:54:51 CDT`), not independently read.

Both hosts are now applied and live. The laptop was at 57% disk / 24% inodes with
117,236 `/tmp` top-level entries, so this was preventive rather than a fix.

PR #1375, opened 2026-09-08
docs(gates): inventory all 81 gates/tiers/batteries/hooks with evidence — 67 have a named catch, 0 earn a DROP

ZacxDev:
Closes clawgate task **#525**.

An inventory of **all 81** gates, tiers, batteries and hooks devrc runs. Every row carries
either **named evidence of a real catch** — a session id, a CI run, or a commit sha — or the
explicit string `no evidence found` with the search that produced it, plus a `KEEP`/`DROP`/`TIER`
verdict.

🔴 **Nothing that runs is changed by this PR.** No gate, target, battery or hook is dropped,
disabled or made conditional, and the per-path selection rule in §10 is **proposed, not
implemented**. Both are the task's explicit non-goals; the verdict column is a recommendation
the operator decides on.

## How the enumeration was derived

From the repo at scan time, not from the task's own list (which the document then contradicts in
two places — see below). Commands and counts are in §1 so they can be re-run and disagreed with:
31 pytest targets, 5 node suites, 2 nix check derivations, 10 mutation batteries, 16
claude-hooks modules, 13 drift-check arms, the githooks pre-push gate, 2 Tekton checks.

## Method, and why the numbers are worth anything

Four transcript scanners over **5,836** `.jsonl` (6.8 GB), **each with a negative and a positive
control, all reported in §2**. This session's own transcript is excluded from every count.

🔴 **One control FAILED, and the table was not trusted until it was explained.** `scan_drift.py`'s
`rc=0` positive control returned 0 hits. Cause read out of the source rather than assumed:
`scripts/drift-check.sh:3554` guards the `(rc=$rc)` line behind `if [ "$rc" != 0 ]`, so a clean
run is **structurally incapable** of emitting that format — it prints
`drift-check: no drift on the host(s) CHECKED` instead. Re-run against that sentence: **34
sessions**. Control passes.

**The `toolDenialKind` co-requirement was load-bearing, and here is the measurement that shows
it:** on the same search string, text-only finds **97** sessions and text-plus-marker finds
**75**. Those 22 sessions *discussed* a guard without being blocked by it, and would have
inflated every hook row.

## What the evidence shows

- **67 of 81 rows carry a named catch.** All 10 mutation batteries; 11 of 14 `bash-guard` arms;
  11 of 13 drift-check arms; 28 of 31 pytest targets red at least once; 4 of 5 node suites; 8 of
  8 hooks that are actually guards.
- **14 read `no evidence found`**, each with its search recorded, and **none on idleness
  grounds**: six are modules that are not gates and emit nothing to find; two are drift-check
  arms structurally unable to fire yet; three are never-red targets covering guards that *have*
  fired; three have the evidence in a sibling tier.
- **0 rows earn a DROP.**

The clearest case for keep-despite-no-evidence is §7's three zero-firing `bash-guard` arms
(`mkfs`, `dd` to a block device, `rm -r $HOME`): their cost is a regex on a string already being
matched for eleven other arms, and the loss they prevent is total and unrecoverable. **Frequency
is the wrong axis for a guard whose downside is unbounded** — which is the same reason a
run-count could never have answered this question.

## Three findings, none of them a drop

1. 🔴 **`scripts/devhost-tests` is in `TARGET_FLOORS` (31 entries) and not in `HERMETIC_TARGETS`
   (30).** It appears in **4** sessions where its neighbours appear in 200–590, while its floor
   still counts toward the derived global floor. This is already the failure mode the task's
   non-goals warn about — a target that silently stopped running — arrived at by omission rather
   than by anyone choosing it. Decide it explicitly, either way.
2. 🔴 **No mutation battery is wired into any gate.**
   `grep -rn "mutants-" scripts/gate.sh scripts/run-tests.sh scripts/run-node-tests.sh flake.nix githooks/`
   returns nothing. Defensible as manual one-shot instruments, but currently undocumented — worth
   a line in each battery's header so the next reader does not assume they run.
3. 🔴 **The proposed tiering saves less than it looks, and the saving is unmeasured.** The
   always-run set is **14,053 of the 21,202** pytest floor-sum (66%), capping the benefit at
   ~34% and only on single-subsystem diffs. Floors are not wall-clock. §11 also lists two risks
   worth answering before building it: a path rule is itself a guard that can fail silently, and
   conditional selection interacts badly with the derived global floor.

## Also in this PR

`scripts/check-gate-inventory.py` asserts criteria 3 and 4 mechanically over the table —
**81 rows checked, 0 problems**. It carries its own controls (`--self-test`: four malformed row
shapes each rejected, a well-formed row accepted, plus a scoping control), and it **errors rather
than passing when it matches zero rows**, because a vacuous green there would certify an empty
table. Its first run reported 11 false positives against §10's rule table, which reuses row ids;
fixed by scoping to the inventory sections rather than by loosening the check.

It is deliberately **not** wired into any gate and **not** under `scripts/tests/` — a file there
would change that target's collected count and force a `TARGET_FLOORS` re-pin, which this
document does not need.

## Follow-on

Clawgate task **#528** — add `emit_invocation` to the gates and the five blocking guard hooks,
recording the **outcome** (and, for `bash-guard`, **which arm**), so this is answerable from
`activity.events` directly next time instead of by transcript proxy. An event recording only
"the gate ran" would reproduce the exact gap this document had to work around, so #528 makes the
outcome field a criterion and requires the emitter to be fail-open and demonstrated so.
(#526 was already taken by the session that filed #525.)

## Note on criterion 1

The task says "a PR against `ZacxDev/devrc`". This repo's origin is
`innovation-upstream/devrc` (`git remote -v`). Read as "the devrc repo" and targeted at origin.
Flagged rather than silently resolved.

https://claude.ai/code/session_01HXxTANDijZTVxj398NrWBU


ZacxDev:
## 🔴 Correcting this PR's description — round 1's headline finding is RETRACTED

Posting rather than silently editing the body, because a reviewer may already have read the
wrong version.

**The PR body's finding #1 — that `scripts/devhost-tests` had "silently stopped running,
arrived at by omission" — is false on all three of its claims**, and the remedy it proposed
would have made things worse:

| body claim | measured |
|---|---|
| "arrived by omission rather than by anyone choosing it" | `scripts/run-tests.sh:876-898` is a dedicated `DEVHOST_TARGETS` array under a 🔴 comment explaining the choice — it needs a real `nvim`, added 2026-08-29 |
| "a target that has silently stopped running" | `scripts/tests/test_nvim_clipboard_osc52.py` parses that array and **asserts the registration exists** |
| "its floor of 6 is counted into the derived global floor regardless" | **False.** `run-tests.sh:900` sets `TARGETS=("${HERMETIC_TARGETS[@]}")` and the floor sums over `$TARGETS`. The PR's own gate output said `floor: 21281 = sum of **30** per-target floors` against 31 targets — the refuting evidence was in hand and I misread it |

🔴 **And "add it to `HERMETIC_TARGETS`" would have redded the sandbox tier** — these tests need
`nvim`, which the nix sandbox does not carry, which is exactly why they were moved out
(`ERROR: 3 UNPINNED skip group(s)`). No action is needed; the design is correct as it stands.

**Second retraction: the `bash-guard` arm table was wrong in both directions.** It was assembled
from the guard's deny *messages* instead of from `guard_core._CLAUDE_CODE_CHECKS`, the list
`bash-guard.py:60` actually runs:

- **`check_git_commit_to_main` was missing entirely** — re-measured independently of the audit at
  **435 blocks across 270 sessions**, the *third most-fired arm in the repo*, and the one
  enforcing this repo's own 🔴 never-commit-to-main rule.
- **`rm -r $HOME` is not a `bash-guard` arm at all** — it is in `_IRREVERSIBLE_CHECKS`, an
  opencode-only policy, under a 20-line comment arguing the exclusion. It has fired 3 blocks / 1
  session, from *opencode's* guard.

That one mattered beyond bookkeeping: the body's keep-without-evidence argument read *"a regex on
a string already being matched for eleven other arms"*. For `rm -r $HOME` under `claude-code`
that is false — it is not matched at all, so its zero was a **policy** fact, not a rarity fact.
An operator persuaded by that argument might have moved `check_rm_rf_critical` into the
claude-code policy, reversing a deliberate and argued decision. **Corrected: 12 of 14 arms have
fired; the zero-firing set is two, not three.**

**Also corrected in round 2**, each measured rather than reasoned:

- **Row C3 counted readers, not firings.** `gate.sh:250` emits `GATE: RESULT=UNVOUCHED`; it emits
  `could-not-vouch` nowhere — that string lives in `CLAUDE.md`, which carries *both* of round 1's
  needles in one sentence. Re-measured: **18 sessions, not 241**. The `KEEP` survives; the figure
  overstated ~13×. A fabricated-name negative control structurally cannot catch this, and that
  limit is now stated in §2.
- **The documented search needle was wrong** — the runners emit **two** spaces
  (`FAIL  <target>  (`). Round 1's regex used `\s+` so its counts were right, but a reader
  following the appendix would have got 0 for nearly every row and read that as confirmation.
- **Instrument 3 cannot separate the two tiers** — both runners emit an identical
  `PASS  <dir>  (` and three dirs are both a pytest target and a node suite. Six rows are now
  marked `AMBIGUOUS`, and C2's reason no longer leans on a separation the instrument cannot make.
- **§10's always-run triggers omitted `scripts/testlib/**`** (122 tracked files, 8+ targets) —
  a live instance of the risk §11 itself states. Added.
- **The base moved twice under this PR.** #1377 retired `initiatives` and `repo-cos`, so round 1
  carried rows for two *deleted* targets; then two new mutation batteries landed. Re-anchored
  `112a5225` → `01956bf0`: **81 rows** (was 81 → 79 → 81 for different reasons), pytest
  floor-sum 21,202 → 20,348.

### The checker had three defects of its own, all now mutation-verified

- **Its scoping control was vacuous.** The fixture contained no `## 3.` line, so `in_scope` was
  `False` from the start regardless of the branch it claimed to cover; deleting that branch left
  the self-test fully green. Fixed, and *mutation-checked*: with the branch deleted the control
  now fails with **its own** error, so it is reachable.
- **Its evidence check accepted vocabulary, not evidence.** The bare word `CONFIRMED` passed.
  Decisive control: replacing all 81 evidence cells with `CONFIRMED` yielded `RESULT: PASS`. It
  now requires a lookup-able identifier; the same control fails all 81 rows.
- **Its zero-row guard was narrower than its docstring** — a document truncated to one row
  passed. The document now declares its size *and* its verdict tally in machine-read markers; a
  mismatch or missing marker exits 2.

🔴 **The tally marker immediately earned its keep: the hand-written tally was wrong again this
round** (47/32 written, 49/30 counted), as it was in round 1 (51/30 written, 49/32 counted) —
while the row *lists* beside it were correct both times. It is now machine-checked rather than
proofread.

### The pattern, which is the part worth keeping

Both 🔴 were **a set inferred from its members' behaviour instead of read from the registry that
defines it** — the arm table from deny messages rather than `_CLAUDE_CODE_CHECKS`, row A29 from
two array names without reading the 40 lines of comment above them. Every remaining error was a
variant: a needle inferred from prose rather than the emitting `echo`, a tier separation assumed
rather than checked against the runners' output.

What held up was the mechanical arithmetic — all floors, both floor-sums, 12/12 battery shas,
11/13 drift arms, every quoted line number. **The controls did their job where they were pointed;
what they could not do was tell the method it was pointed at the wrong thing.** That is recorded
in §13 of the document rather than left in this thread.

PR #1381, opened 2026-09-08
refactor(cairn): `who` becomes its own `cairn-who` binary, and the shared timeout predicate moves to the module both import

ZacxDev:
`cairn who` was the one subcommand about a different noun. Every other verb acts on an ENTRY in the hosted store and takes `--scope`/`--repo`/`--no-sync`; `who` took none of them, synced nothing, and needed its own longer timeout because it shells into tmux on two hosts rather than fetching an HTTP snapshot.

It is also the seam `github.com/ZacxDev/cairn` chose independently — the OSS extraction ships the store client and excludes `who`, because session forensics on named hosts cannot go in a public repo. `claudedocs/handoff-cairn-oss-multi-instance.md` ranks this as rank 3's second half; claimed as `cairn-oss-multi-instance-3`.

**Scope:** the split only. NOT pinning devrc to the cairn flake, NOT moving `~/.local/bin/cairn` into `/nix/store`, NOT the `entry_shape`/writer-vocabulary question — the handoff doc says to put that last one to the operator before building, and it is untouched here.

## The non-obvious part

The STORE path called `_cairn_who().unbounded_timeout_reason(timeout)` — it borrowed the predicate by importing the whole `who` module, which only worked while the two were one program. That rule was open-coded at two sites once before and the copies DISAGREED (only one excluded `bool`; the weaker copy ran with a silent one-second bound). Re-open-coding it on the store side would recreate exactly that state, so it moved to `scripts/lib/timeouts.py`, which both binaries import at module scope. The `bool`-subclasses-`int` trap and its measurement are carried over verbatim.

No shared `DEFAULT_TIMEOUT`: the store's 20s and who's 60s differ for measured reasons, and one constant would erase both.

## Test matrix

| | result |
|---|---|
| HEAD, the three touched suites | **112 passed** |
| fork point, new suite only | **12 failed / 8 passed** |

The 12 reds are the regression coverage. The 8 that pass at base are `test_the_STORE_fetch_STILL_refuses_an_unbounded_timeout_after_the_split` parameterized over `None/0/-1/60/True/False/1.5`, plus `test_a_POSITIVE_bound_is_NOT_refused` as its positive control — the behaviour predates the split, so these are **invariant guards** and are not counted as regression coverage.

Authoritative gate on the rebased tree, counts read from the log rather than the exit code:

```
PASS  scripts/tests  (collected=12870 passed=12870 skipped=0 floor=12820)
TOTAL collected=20915  passed=20913  skipped=2  failed=0
```

## Live verification

Run against the worktree's own binaries:

- `cairn doctor --no-sync` → rc 10, `frozen-mirror OK` — the store path is intact
- `cairn ls-entries` → live fetch, 225 entries — the network path still works
- `cairn who foo` → **exit 2**, `invalid choice: 'who'` (the value the SKILL now claims)
- `cairn-who --help` → rc 0, `--json`/`--host`/`--no-windows`/`--timeout` preserved
- `cairn-who <id>` → reached clawgate and returned a well-formed refusal

## A stale comment this uncovered

`nix/home.nix` justified `mkOutOfStoreSymlink` by naming `cairn_who` importers at `scripts/cairn:68/756/808/818`. Checked at the fork point: line 68 is blank and 756/808/818 are unrelated comments about recall and silent zeros — the real references were at 285/289/957/1334+. The comment was already wrong before this change. Line numbers are gone; the requirement is restated on each script's own `__file__` lookup, and `cairn-who` gets the same deploy mode for the same reason.

## Floor ratchet (second commit)

`scripts/tests` collected 12870 against a pinned floor of 12793 — 77 of slack, more than this branch's 20-test suite could lose with the gate still green. Re-pinned to 12820 per the file's documented rule, after the rebase per its ORDER note. Floor sum 20208 → 20235.

## Deploy note

🔴 **Merging this requires a `home-manager switch`** — corrected after the round-1 audit; the original wording here was wrong and is retracted in a PR comment.

Only the `who`-REMOVAL half is live on merge: `~/.local/bin/cairn` already resolves to `scripts/cairn`, so `cairn who` starts exiting 2 as soon as the checkout is pulled. The new `~/.local/bin/cairn-who` symlink does **not** exist yet and is created by home-manager ACTIVATION — `mkOutOfStoreSymlink` describes the link's target, not who creates the link. `claude/skills/cairn/SKILL.md` is likewise a `home.file` COPY into `/nix/store`, and the deployed copy still says `cairn who`.

Between merge and switch the capability is therefore unavailable and the skill misroutes. **No switch was run.**

⚠ An earlier version of this section listed a deliberately-left nit — a `# noqa: E402,F401` whose `F401` could never fire. **Round 1 fixed it** in `02fcc1a3`, and this paragraph went stale describing it (with two wrong line numbers). The file now carries `# noqa: E402` alone. ⚠ An earlier version of this sentence said `02fcc1a3` was "the same commit that corrected the deploy note" — there is no such commit. The deploy note was corrected in a PR comment plus an edit to this body; `02fcc1a3` does not touch `nix/home.nix`. A provenance claim with nothing in the tree to find is the same defect this section is about.

On the floor-sum figure above: `20235` is the gate's OWN printed line, `floor: 20235 = sum of 28 per-target floors` — the sum over the **28 hermetic targets**, not over all 29 `TARGET_FLOORS` entries (the remainder are dev-host-only). A round-2 audit could not place it and reported a non-matching total; measured here, `HERMETIC_TARGETS` has 28 entries and `TARGET_FLOORS` has 29.

https://claude.ai/code/session_01E3KEUsTLLUhS37an52rn28

### DIFF
--- a/claude/skill-tiers.json
+++ b/claude/skill-tiers.json
@@ -55,7 +55,7 @@
     "browser": { "tier": "A" },
     "cairn": {
       "tier": "B",
-      "why": "the tool is NAMED, never described -- 'cairn doctor', 'cairn sync', 'cairn who'. The symptoms that could route here already belong to skills that name it in their own bodies: a refused read to `resume`, a stale index block to `analyze-service`, an entry that got huge to `prune-index`"
+      "why": "the tool is NAMED, never described -- 'cairn doctor', 'cairn sync', 'cairn-who'. The symptoms that could route here already belong to skills that name it in their own bodies: a refused read to `resume`, a stale index block to `analyze-service`, an entry that got huge to `prune-index`"
     },
     "check-clickup-addressed": { "tier": "A" },
     "clawgate": { "tier": "A" },
--- a/claude/skills/cairn/SKILL.md
+++ b/claude/skills/cairn/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: cairn
-description: "The hosted subsystem store (pod in ns `subsystem-store`) and its client `scripts/cairn`. Use for: `cairn doctor`, cairn sync/recall/search/ls-entries/who, a stale or unstamped store, a `cairn` exit 4, seeding the pod, a scope a token cannot reach. Writes are `subsystem-index`; pruning is `prune-index`."
+description: "The hosted subsystem store (pod in ns `subsystem-store`), client `scripts/cairn`. Use for: `cairn doctor`, cairn sync/recall/search/ls-entries, cairn-who, a stale or unstamped store, a `cairn` exit 4, seeding the pod, a scope a token cannot reach. Writes are `subsystem-index`; pruning is `prune-index`."
 allowed-tools: Bash, Read, Grep, Glob
 ---
 
@@ -33,12 +33,15 @@ staleness it was run to measure.
 | find a hunk by text | `cairn search '<query>'` (`--all-scopes` to search every scope) |
 | what does the cache actually hold | `cairn ls-entries` |
 | parse-check the cached entries | `cairn validate` |
-| which sessions/windows/transcripts worked a task | `cairn who <task>` (`--json`, `--host`, `--no-windows`) |
+| which sessions/windows/transcripts worked a task | `cairn-who <task>` (`--json`, `--host`, `--no-windows`) — a SEPARATE binary |
 | diagnose anything above going wrong | `cairn doctor` |
 
-`cairn who` is about a **task**, not a store entry: it touches no store and never
-syncs, and it has its own longer `--timeout` because it shells into tmux on two
-hosts.
+🔴 **`cairn-who` is a separate command, not a `cairn` subcommand.** It is about a
+**task**, not a store entry: it touches no store and never syncs, takes none of
+the `--scope`/`--repo`/`--no-sync` flags, and has its own longer `--timeout`
+because it shells into tmux on two hosts rather than fetching an HTTP snapshot.
+Typing it as a `cairn` subcommand is no longer valid: argparse exits 2 with an
+`invalid choice` naming the verbs that remain.
 
 **Writes are not this skill's.** `subsystem-index` owns the one protocol for
 every writer; `prune-index` owns deletion, with its own confirmation gate. Load
--- a/claudedocs/handoff-cairn-oss-multi-instance.md
+++ b/claudedocs/handoff-cairn-oss-multi-instance.md
@@ -252,15 +252,34 @@ countable.
    forcing: none — done
 
 3. 🔨 **HALF DONE — Phase A3, devrc consumes cairn as a pinned flake input.**
-   🔴 **IN FLIGHT ELSEWHERE — `claim-work` slug `cairn-oss-multi-instance-3` is HELD by another
-   session** (taken 2026-09-08 ~01:45Z) for the **first slice only**: re-home `cairn who` as its
-   own `cairn-who` binary + extract `unbounded_timeout_reason` into devrc `lib/timeouts.py`.
-   That session explicitly did NOT pin the flake and did NOT touch the `entry_shape`/writer fork.
+   🔴 **IN FLIGHT — `claim-work` slug `cairn-oss-multi-instance-3`, taken 2026-09-08 ~01:45Z
+   for the FIRST SLICE ONLY**: re-home `cairn who` as its own `cairn-who` binary + extract
+   `unbounded_timeout_reason` into devrc `lib/timeouts.py`. That slice explicitly does NOT pin
+   the flake and does NOT touch the `entry_shape`/writer fork.
    - **Half 1 — ✅ MERGED 2026-09-08: `ZacxDev/cairn`#4, squash `218b6c1`.**
-   - **Half 2 — the rest is still open** after that slice: pin the input in `flake.nix`, move
-     `~/.local/bin/cairn` into `/nix/store`, point the writer at the pinned `entry_shape`, and
-     decide the fate of devrc's five duplicated `lib/` modules. 🔴 **THE OPERATOR FORK IS STILL
-     UNANSWERED and must be put before building** — see the fork investigation above.
+   - **Half 2 — one slice of four is in review; the rest is still open.** Remaining: pin the
+     input in `flake.nix`, move `~/.local/bin/cairn` into `/nix/store`, point the writer at
+     the pinned `entry_shape`, and decide the fate of devrc's five duplicated `lib/` modules.
+     🔴 **THE OPERATOR FORK IS STILL UNANSWERED and must be put before building** — see the
+     fork investigation above.
+     - **The slice is devrc PR #1381.** Extracting `unbounded_timeout_reason` is not optional
+       there: the store path reached that predicate by importing the whole `who` module, a
+       reach that breaks the moment the two are separate binaries, and re-open-coding it
+       recreates the two-copies bug the predicate already caused once.
+       🔴 **Merging it requires a `home-manager switch`.** The `who` REMOVAL is live on merge
+       (`~/.local/bin/cairn` is an out-of-store symlink), but `~/.local/bin/cairn-who` is
+       created by ACTIVATION, and the deployed skill copy still says `cairn who` until the
+       switch. Between merge and switch the capability is unavailable.
+     ⚠ Every `cairn who` spelling elsewhere in this doc — and in
+     `handoff-cairn-task-linkage.md` and `proposal-cairn-session-capture.md` — becomes the
+     DEAD spelling once #1381 merges, and exits 2. They are left as written because they
+     record what was true when written; do not copy a command out of them.
+   🔴 `nix/home.nix` deploys `scripts/cairn` as an `mkOutOfStoreSymlink` and the comment above
+   that entry says it is REQUIRED, not preferred, because `.resolve()` must land beside `lib/`.
+   (⚠ This bullet cited `:1474` and `:1467`; both were already wrong at `c5e425c7`, landing on
+   unrelated `shell-env-nudge` comments — so no line numbers. Grep for
+   `mkOutOfStoreSymlink is NOT a preference`.) **#4 solves exactly that** by installing script
+   and `lib/` together under `libexec`. Client edits will then need a `home-manager switch`.
    **Closing condition:** a merged devrc PR in which `flake.nix` names cairn as an input and
    `readlink -f ~/.local/bin/cairn` resolves into `/nix/store`. Re-verified NOT met 2026-09-08.
    forcing: none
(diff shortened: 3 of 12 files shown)


ZacxDev:
## Correction to the PR description (round-1 audit finding, 🟡)

The deploy note in the body above was **wrong**, and a reviewer may have already read it. Correcting publicly rather than silently editing.

It said:

> The two scripts are `mkOutOfStoreSymlink`, so they take effect on merge without one.

That is true for the *removal* half and false for the *new binary*. `mkOutOfStoreSymlink` describes the link's **target**, not who creates the link — the link itself is created by home-manager activation. Measured just now:

```
readlink -f ~/.local/bin/cairn      -> /home/zach/workspace/devrc/scripts/cairn   (live; removal lands on merge+pull)
ls ~/.local/bin/cairn-who           -> No such file or directory                  (does not exist yet)
grep -c 'cairn who' ~/.claude/skills/cairn/SKILL.md -> 2                          (deployed store copy, still the dead spelling)
```

**So there is a real gap window.** On merge + `git pull`, before any `home-manager switch`:

- `cairn who <task>` → exit 2, `invalid choice: 'who'`
- `cairn-who <task>` → `command not found`
- the loaded `cairn` skill still routes readers to `cairn who`

The capability is unavailable until a switch, and the skill actively misroutes during that window.

**Corrected note:** merging this requires a `home-manager switch` to restore the capability. The switch creates `~/.local/bin/cairn-who` and refreshes the skill's store copy; only the `who`-removal half is live-on-merge. The body has been updated to say this.

Nothing else in the audit was deploy-blocking. Fixes for the remaining 🟢 findings follow in a delta commit.

ZacxDev:
## Round 1 — fixes landed

Round 1 returned **no 🔴**. One 🟡 (the PR description) and seven 🟢. Fixes are in `02fcc1a3`.

```audit-claims round=1 audited=a57f0a0025fa4df11ed97e04eedb3d0389b8d7ea..02fcc1a3
1. The PR description's deploy note is corrected publicly: merging requires a `home-manager switch`, because `~/.local/bin/cairn-who` is created by activation while the `who` removal is live on merge.
2. `test_the_timeout_predicate_has_exactly_ONE_implementation` now counts AST `isinstance(_, int)` CALL nodes over the three consumers, with `timeouts.py` as a positive control rather than a fourth zero-asserted subject.
3. `timeouts.py`'s docstring no longer claims that guard pins the rule "structurally ... and this file".
4. The nix-comment guard is renamed `test_the_nix_deploy_comment_still_states_why_the_MODE_is_required` and narrowed to assert only that the deploy comment still states why the mode is required; two stronger drafts were withdrawn and both are recorded in its body.
5. `_store_timeout`'s docstring no longer asserts a caller count; it says "every site" instead of "four".
6. The `scripts/tests` floor entry states what the ratchet does NOT close: 50 of slack remains by the rule, still more than this branch's own suite.
7. The shared-predicate import guard accepts `import timeouts` + attribute access as well as the `from` form.
8. `cairn_who.py`'s module-scope `sys.path` insert is guarded with `if _LIB not in sys.path`, like its siblings.
9. `claudedocs/handoff-cairn-oss-multi-instance.md` no longer marks rank 3's half 2 NOT STARTED, and flags the now-dead `cairn who` spellings in the sibling handoffs.
```

Two notes for whoever reads this next, because both are the kind of thing that reads as covered when it is not:

- **Two of my own fixes were defective and the mutation testing is what found them.** The first "structural" nix-comment guard was *vacuous* — it asked whether `scripts/cairn` reaches `cairn_who` by substring, which this script's own docstring satisfies (`_store_timeout` names `cairn_who._run`), so its assertion was unreachable and the reworded justification stayed green. The AST replacement was accurate about the code and still wrong, because it forbids the *retraction* the comment legitimately carries. Hence the narrowing in claim 4 rather than a third clever predicate.
- **The mutation harness itself produced a false green first.** `open(p,"w").write(open(p).read()…)` truncates before it reads, so both mutation targets became empty files. Re-run reading into a variable first, with byte deltas as the receipt: owner `3486 -> 3482 B` (positive control fires), `cairn-who` `1940 -> 1998 B` (the case the old regex missed), plus the deploy-comment WHY stripped while keeping the block (fails with its own message).

Gate unchanged on the fix tree: `scripts/tests collected=12870 passed=12870 floor=12820`, `TOTAL collected=20915 passed=20913 skipped=2 failed=0`.

ZacxDev:
## Round 2 — fixes landed

No 🔴, two 🟡, four 🟢. Fixes in `43f86f3f`. The first 🟡 was against round 1's own fix — the third failed predicate on the same target — so the paragraph is now pinned whole instead.

```audit-claims round=2 audited=02fcc1a37db2d8a1fb3b52b80310aa943d5ac3f0..43f86f3f
1. The nix deploy comment's WHY paragraph is pinned WHOLE and whitespace-normalised as `NIX_DEPLOY_WHY`; the test's name, docstring, inline comment and assertion message now all describe that same check.
2. `claudedocs/handoff-cairn-oss-multi-instance.md:29` and `:41` no longer contradict the ranked entry at `:211`/`:220` — half 2 reads STARTED, and the slug is described as re-claimed with `claim-work --list` named as the authority.
3. `test_BOTH_sides_import_the_predicate_from_the_module_that_owns_it` parses the AST instead of matching substrings, so a comment mentioning the import cannot satisfy it.
4. `scripts/run-tests.sh:1084`'s "~1900" is corrected to 12870 and dated, with a note to re-derive rather than trust it.
5. The no-copies assertion message no longer claims to know whether a matched `isinstance(_, int)` is a timeout check.
6. The PR description's "deliberately left" nit is retracted (round 1 fixed it), and the `20235` floor sum is attributed to the gate's own printed `sum of 28 per-target floors`.
```

**Ledger — round 2's fixes changed ZERO payload lines.** Range `02fcc1a3..43f86f3f --not origin/main` (base `18bc1500`, rc 0, silent stderr, non-empty range): `claudedocs/handoff-cairn-oss-multi-instance.md` 10/6, `scripts/run-tests.sh` 4/1, `test_cairn_split.py` 99/47, `test_cairn_who.py` 6/3 — all scaffolding. Round 1's fixes were 45 payload lines, so this is the **first** zero-payload round, not the second; the attribution gate does not fire yet. If round 3's fixes are also zero-payload, the ladder has left the PR and I will stop there and file the remainder rather than run a round 4.

Mutation receipts for this round, re-run against the auditor's exact mutants (`cp -a` copy, `rm -f <copy>/.git`, `PYTHONDONTWRITEBYTECODE=1`):

| mutant | before | now |
|---|---|---|
| WHY paragraph deleted, word `mkOutOfStoreSymlink` kept (−468 B) | GREEN | **RED** |
| import replaced by re-open-coded predicate + prose naming it (29689 → 29917 B) | GREEN | **RED** |
| control, unmutated | — | **GREEN** (20/20) |

One audit item resolved as *not a defect*: round 2 could not place the body's `20235` and reported `TARGET_FLOORS` as 39 entries summing to 29093. Measured here — `HERMETIC_TARGETS` has **28** entries, `TARGET_FLOORS` has **29**, and `20235` is the gate's own printed hermetic-tier sum. The figure is correct and is pinned by that line.

Gate on the fix tree: `RESULT: PASS (exit=0)`, `scripts/tests collected=12870 passed=12870 floor=12820`, `TOTAL collected=20915 passed=20913 skipped=2 failed=0`.

ZacxDev:
## Round 3 — fixes landed, and the ladder ENDS here

No 🔴, one 🟡, three 🟢. Fixes in `0cae9fd3`.

```audit-claims round=3 audited=43f86f3f7dc681021cccc4303cb9530f0f2dec4c..0cae9fd3
1. The import guard now requires the CALL to come from the IMPORT — `(from_form and bare_calls) or (module_form and attr_calls)`, attribute receiver pinned to `timeouts` — and refuses a local `def unbounded_timeout_reason` outright.
2. The withdrawn-draft list header says FOUR and tells the reader to count the items; each item names whether it failed for being TOO WEAK or TOO STRICT, and draft 3 is explicitly marked do-not-re-derive.
3. The mutant's size is stated as 468 characters / 475 bytes, with the char-vs-byte confusion in the earlier receipts named.
4. The PR description no longer claims `02fcc1a3` corrected the deploy note; that correction is a PR comment plus a body edit.
5. The no-copies message tells a reader to make the guard DISCRIMINATE on a false positive, not to widen it.
6. `claudedocs/handoff-cairn-oss-multi-instance.md`'s `nix/home.nix:1474`/`:1467` citations are removed — both were already wrong at the merge base.
```

### Why this is the last round

**Not a clean round, and not a cap — the attribution gate.** Round 2's fixes changed 0 payload lines. Round 3's fixes touch `test_cairn_split.py`, `test_cairn_who.py` and a handoff doc: 0 payload lines again (measured, `43f86f3f..0cae9fd3 --not origin/main`, base `18bc1500`, rc 0, silent stderr, non-empty range — 7/3, 55/20, 4/2, all scaffolding). Two consecutive zero-payload fix rounds means the rounds are auditing scaffolding the ladder itself wrote, and a findings-keyed stop rule cannot fire on that.

This criterion was **declared in the round-2 comment, before round 3's result was known**. Round 3 reached the same conclusion independently, and all four of its findings were scaffolding.

**Nothing is filed as outstanding** — every round-3 finding is fixed in `0cae9fd3`. Two nits remain and are deliberately not fixed: the `#:` doc block for `NIX_DEPLOY_WHY` sits ~20 lines from the constant, and the pin test's docstring is narrower than its body (the safe direction; the assertion message says what to do).

### The pattern worth recording

Rounds 1→3 each found a defect in the **previous round's fix**, all the same shape — *wider on one axis, narrower on another*:

| round | the fix | what it broke |
|---|---|---|
| 1 | spelled guard → "does the code reach it" | vacuous: satisfied by the script's own prose |
| 2 | → AST | over-strict: forbade the retraction the comment carries → pinned the paragraph whole |
| 3 | substring import check → AST | closed prose, opened a real re-open-coded copy (103 passed / 0 failed) |

No round found a defect in the shipped behaviour of the split. `f869e170`'s `who` re-homing and `timeouts.py` extraction have been clean since round 1, verified live and by mutation in every round. What kept producing findings was the prose *about* the guards.

### Still true at merge time

- 🔴 **Merging requires a `home-manager switch`** — `~/.local/bin/cairn-who` is created by activation; only the `who` removal is live on merge, and the deployed skill copy says `cairn who` until the switch.
- `origin/main` is independently RED (`age` 1.3.2 / `opencode` 1.18.29 vs pinned 1.18.21, from #1382's lock bump). Controlled at both the fork point and this head — identical failures with and without this branch. GitHub reports `MERGEABLE`/`CLEAN`, so that red is invisible from this page.
- The claim `cairn-oss-multi-instance-3` stays **held** until this merges; releasing it now would let another session redo work that is already in review.

Gate on the final tree: `RESULT: PASS (exit=0)`, `scripts/tests collected=12870 passed=12870 floor=12820`, `TOTAL collected=20915 passed=20913 skipped=2 failed=0`.

PR #1388, opened 2026-09-08
fix(find-session): a 12-day default archive window that says it is one, plus the crash and the degenerate --tail a live trace found

ZacxDev:
Three findings from tracing `find-session` end to end, plus the default window the archive leg's cost has been asking for.

## 1. `live_scan` could crash out of the contract that says it never does

`isinstance(report, dict)` proved the top level was an object — then the next line called `.items()` on whatever `hosts` was and `.get()` on each value. Measured against the function, all four raising past the `except` that only wraps the subprocess call:

```
{"hosts": [1, 2]}        -> AttributeError: 'list' object has no attribute 'items'
{"hosts": {"wb": null}}  -> AttributeError: 'NoneType' object has no attribute 'get'
{"hosts": {"wb": []}}    -> AttributeError: 'list' object has no attribute 'get'
{"hosts": {"wb": "..."}} -> AttributeError: 'str' object has no attribute 'get'
```

A function whose whole job is to report *"the live fleet was NOT measured"* cannot report it by crashing. Discriminated at the source, so no publisher re-derives it. A session-manager version skew is enough to trigger any of the four.

## 2. `--tail` below 1 was accepted where `--limit` below 1 is refused

Measured against a real window before the fix: `--tail 0` and `--tail -5` both exited **0** and printed a scrollback block headed `(last 0 lines)` / `(last -5 lines)` — a header stating a line count nothing honoured, over a value `session-manager tail` had silently clamped to zero. Output-line deltas at four values: `0 -> +0`, `-5 -> +0`, `5 -> +5`, `40 -> +40`. Refused now, not clamped — the same call `--limit` gets one flag over.

## 3. A 12-day default archive window that is never silent

Measured 2026-09-08, warm, `--claude-only`, back to back, quoting **the walk's own `skipped_stale` counter**:

| window | elapsed | skipped unopened | |
|---|---|---|---|
| `--since 3d` | 2.40 s | 780 of 924 | 5.5x |
| **`--since 12d`** | **7.35 s** | **467 of 924** | **1.8x** — the default |
| `--since 30d` | 13.35 s | **0 of 924** | 1.0x |
| `--all-time` | 13.09 s | — | |

🔴 **A 30-day default would be a no-op** — it skips *zero* of 924 files, because nothing in this corpus is older than that. It would cap the answer and buy nothing. 12 days skips about half.

Whole-tool, both corpora: **7.67 / 7.99 s** windowed against **14.26 / 17.06 s** under `--all-time` warm, and **42.96 s** cold.

**The bound is never silent.** Every archive run prints its window and the number of transcripts it skipped unopened — from `search()`'s own counters, so it is a measurement rather than a second copy of the prefilter — and `--json --live` carries the same facts as `archive.window`. `--all-time` lifts it. Naming both windows is a usage error rather than a silently-inert flag. A count nothing measured publishes as `null`, never as a reassuring `0`.

```
ARCHIVE window: the last 12 days (since 2026-08-27) — DEFAULT, not a corpus-wide
search. Anything older was NOT looked at; pass --all-time for the whole corpus
(~15s warm, ~43s cold) or --since YYYY-MM-DD for a different window.
(local transcripts skipped unopened: 467 of 924)
```

🔴 **`--skill` is exempt and stays corpus-wide.** `adoption-scan` routes *"has skill X ever been used"* here by name. Windowing it would turn a corpus-wide "no recorded use" into "not in the last 12 days" with nothing on screen to say so — the reverse-of-the-truth failure that skill's own SKILL.md documents, one layer down and this time with no keyword to blame. The `skills_used` rollup is itself forward-only from 2026-08-29, so the evidence base is young enough that a short window would start discarding real uses within days. An explicit `--since` still applies under `--skill`: the exemption removes a default nobody asked for, it does not override an instruction somebody gave.

## A denominator this PR also corrects

The docstring and SKILL.md would otherwise have shipped **5,954** as the corpus size — that is `find ~/.claude/projects -name '*.jsonl'`, and ~5,030 of those are `<project>/<id>/subagents/agent-*.jsonl`, which `iter_transcripts` excludes **by name**. The walked set is **924**. Every ratio above is quoted against the counter the walk publishes, not against a `find` this code never performs. The 2026-08-28 perf figures (30.1 s / 1.82 s) are re-measured too, with the ratio named as the argument rather than the constants.

## Test fixtures that were pinned to a wall-clock date

Two fixtures stamped records `2026-08-21` and fell outside the new window — a correct result, answering a question the tests were not asking. `test_find_session_skill_cli.py` now stamps relative to `now` (a fixture anchored to a calendar date is a test that expires); `test_transcript_search.py`'s CLI helper appends `--all-time` **only when the caller named no window**, since several of its tests pass `--since` deliberately and the two together are a usage error.

## Tests

**23 new nodes** in `test_find_session_live.py`, replayed against a detached worktree at `18bc1500` with only this file copied in:

- **20 red at base**, green at HEAD.
- **3 green at base by design**, named in `R4_BOUNDARY_CONTROLS` so nobody counts them as bugs fixed: `test_a_REAL_hosts_object_still_parses` and `test_tail_of_exactly_ONE_line_is_accepted` are boundary controls (a guard that also broke the happy path would show up only here), plus the ledger self-check.

**8 end-to-end nodes** in `test_find_session_skill_cli.py` that drive the real walk over a real fixture corpus — including the **positive control** that `--all-time` returns what the default window hid, without which "the default returned nothing" is indistinguishable from "the fixture never matched".

Every assertion in that class reads session ids out of `--json`, not words out of the human output: the first draft asserted `"ancient" not in out` and **failed against correct code**, because the human path echoes the query (`No sessions matched: zzancient` contains the word the guard was looking for).

## Gate

Base sha `18bc1500`; `origin/main` had not moved when this was run, so **the branch is the merged tree**.

| tier | result |
|---|---|
| `scripts/gate.sh --tier both` (dev host) | pytest `7 failed, 12886 passed`; node **PASS** 1449/1449 |
| `nix build .#checks.x86_64-linux.pytests` (sandbox, alone) | `7 failed, 12886 passed` |
| `nix build .#checks.x86_64-linux.nodetests` (sandbox, alone) | **PASS** 1449/1449 |

The two `nix` derivations were built **one at a time** — a combined invocation contends on the store and produces false failures.

🔴 **The seven failures are inherited, not introduced.** A detached worktree at `18bc1500` with **no** changes reproduces all seven by identical node name (`7 failed, 406 passed`), and they appear with the same names and counts in both tiers. Six are `age-keygen: error: failed to parse input: … unknown identity type` (age 1.3.2 rejecting a test identity) in `test_analyze_service_index_{backup,escrow_verify}.py`; the seventh is `test_engine_is_the_version_every_measurement_is_keyed_to` — opencode is `1.18.29` where the pin says `1.18.21`, and it fails in the **sandbox** too, so that is `flake.lock` drift and not this host's PATH. `main` is red in both tiers today, before this branch.

A `dl-router` sqlite `database is locked` failure appeared in one gate run under full parallel load and not in the next two; it passes in isolation at base. Load contention, not a defect here.

## Live smoke test

On the real corpus, streams read separately (`2>&1 >/dev/null` under zsh's MULTIOS hands you *stdout*, and `$?` after a pipe reads `tail` — both bit the first attempt):

- default run: notice on **stderr**, `skipped unopened: 467 of 924`
- `--tail 0`: exit **2**, stdout empty
- `--since 2026-01-01 --all-time`: exit **2**, stdout empty
- `--skill find-session`: `ARCHIVE window: the WHOLE corpus — the 12-day default is NOT applied to --skill …`

https://claude.ai/code/session_01HhbfcZJEExMvfv8w457x2S


ZacxDev:
## Audit round 1 — 4 🟡, 5 🟢, no 🔴. All fixed in `ccb974bf`.

### Two corrections to the PR description above — it is wrong in two places

1. **"origin/main had not moved … so the branch is the merged tree" is stale.** `main` moved while the audit ran. The branch has been **rebased onto `39c31521`**; the audited commit `0d44c562` is now `033ff988`, content-identical (the rebase touched none of this PR's 5 files — verified by `git diff --name-only 18bc1500..origin/main`, no overlap).
2. **"8 end-to-end nodes in `test_find_session_skill_cli.py`" is wrong — pytest collects 7** in `TestTheDefaultArchiveWindow`. Found by the audit; re-counted with `--collect-only`. The `23 new nodes / 20 red at base` claim for `test_find_session_live.py` reproduces exactly.

Correcting these here rather than editing the body, since a reviewer may already have read the wrong version.

### What round 1 found

Every finding was a **disclosure** defect in a change whose entire thesis is disclosure.

- **🟡1 the exit-2 table stopped enumerating its own codes.** `main` has nine `return EXIT_USAGE` sites; the sentence named five causes covering six. Two of the unnamed were added by this PR (`--tail < 1`, `--since` + `--all-time`), one predates it (`--live` with no terms). The existing gate rebuilds the sentence from a ledger and checks **sentence → code** only — its own comment block says the other direction is open — so the suite stayed green. Closed with a gate that **traces** each probe to the line it returns from; structural counting cannot do it, because probes share sites.
- **🟡2 the "measured size of the cap" measured one leg of three.** The local Claude corpus loses **466 of 926 (50%)**; the local **opencode** corpus loses **487 of 707 (69%)** and appeared in no number, and the peer hosts are unmeasured. `local` was the wrong word too — everywhere else in this repo it means *this host vs a peer*, and an opencode session on this machine **is** a local transcript.
- **🟡3 the window was disclosed on stderr only**, so `find-session.py redis 2>/dev/null` printed `No sessions matched: redis` with no hint it was bounded.
- **🟡4 `ARCHIVE_STATS.clear()` sat inside `if run_archive:`**, so `archive.window.message` could publish a stale `skipped unopened: 0 of 2` beside `archive.ran: false`.

### The new gate was mutation-verified, in isolation

Dropped `--tail below 1` from the sentence, the ledger **and** the shipped doc together, so every other gate stays green: **1 failed, 31 passed** — the new gate red alone, by its own assertion message, naming the exact line. Run in a `cp -a` copy with `rm -f <copy>/.git` first, under `PYTHONDONTWRITEBYTECODE=1`.

### One of this round's fixes falsified a test this PR wrote last round

`test_the_WINDOW_IS_ANNOUNCED_on_the_classic_path` pinned **stderr** and carried the same wrong reasoning in its docstring. Updated with that history, plus `test_the_window_goes_to_STDERR_under_json_so_stdout_stays_PARSEABLE` as its boundary control — a fix that merely moved the print would satisfy the first and corrupt the JSON path.

### Also recorded: why the 30-day no-op is permanent

The audit established the mechanism behind the load-bearing "a 30-day default skips 0 of 924": Claude Code prunes `~/.claude/projects` on a **30-day retention** and no `cleanupPeriodDays` override is set, so that corpus can never be deeper and any default at or above 30 days is inert **by construction**. Explicitly **not** true of the opencode store, which has no such pruning — which is exactly why its larger cut must be named as unmeasured.

### Gate — base `39c31521`

| tier | result |
|---|---|
| `scripts/gate.sh --tier both` (dev host) | pytest `7 failed, 12922 passed`; node **PASS** 1449/1449 |

The 7 are the inherited failures (6× age-keygen 1.3.2 rejecting a test identity, 1× `flake.lock` opencode `1.18.29` vs a `1.18.21` pin). Reproduced on a clean detached worktree at base with zero changes, and present in the sandbox tier too. **No 8th failure.**

### Ledger

`round 1 · payload lines changed THIS round: 155 (find-session.py +121/-28, SKILL.md +4/-2) · scaffolding: +149/-17 across two test files · since round 1: 155`

Payload changed, so the attribution gate does not fire and the ladder continues to a delta round.

```audit-claims round=1 audited=033ff988..ccb974bf
1. The exit-2 sentence in EXIT_CONTRACT now names `--tail` below 1, `--since` with `--all-time`, and `--live` with no terms; EXIT_2_CAUSES and the SKILL.md `2` row were updated to match.
2. A new gate `test_every_EXIT_USAGE_SITE_is_a_cause_the_sentence_NAMES` closes the code->sentence direction by tracing each probe to the `return EXIT_USAGE` line it leaves from, with `test_the_TRACER_can_actually_SEE_a_return_line` as its positive control.
3. The window notice now reads "Claude transcripts on THIS host" and states that the opencode corpus and the peer hosts were windowed too and are not in the count.
4. The window disclosure now goes to stdout on the human classic path and stderr under `--json`; SKILL.md states that `--json` without `--live` has no machine-readable window.
5. `ARCHIVE_STATS.clear()` was hoisted to the top of `main`, so `archive.window.message` can no longer publish a stale count on a run where the archive did not execute.
6. The docstring no longer calls it "the 30-second archive search".
7. A `--since` carrying a time component is now disclosed with that time instead of the wider bare date.
8. The `hosts` shape error names the offending type separately for a bad `hosts` and a bad host entry, and its test asserts that type rather than only the word "hosts".
9. `--live --all-time` no longer suggests passing `--all-time`, and `--live --deep` prints the window once rather than once per stream.
10. The 30-day no-op is recorded as structural (Claude Code 30-day retention on `~/.claude/projects`), and explicitly not true of the opencode store.
11. `test_the_WINDOW_IS_ANNOUNCED_on_the_classic_path` was corrected (it pinned the wrong stream) and given a boundary control for the `--json` path.
12. The previous commit message's and PR body's "8 end-to-end nodes" is corrected to 7.
13. The branch was rebased onto 39c31521; `033ff988` is the content-identical rebase of the audited `0d44c562`.
```

ZacxDev:
## Audit round 2 (delta, blind, `033ff988..ccb974bf`) — 4 🟡, 3 🟢, no 🔴. Fixed in `407ae705`.

Round 1's fixes held — 8 of 13 claims verified outright, and every self-reported number re-derived (the `487 of 707` opencode figure exactly; the Claude denominator had grown 926 → 929 in the intervening hours). What round 2 found is that four of them were **narrower than the sentences describing them**.

### The gate round 1 added was walkable three ways — all measured GREEN

| bypass | disposition |
|---|---|
| a usage error spelled `sys.exit(EXIT_USAGE)` — verified reachable on the mutant, and it slips the sibling bare-literal gate too, which matches `ast.Constant` only | **closed**, watched go SURVIVED → KILLED |
| `return EXIT_USAGE` in a module-level helper `main` returns through — invisible to a collector scoped lexically to `main` | **closed**, watched go SURVIVED → KILLED |
| a new site whose argv is filed under an **existing** cause — traced union stays complete, sentence never names the new reason | **ratchet only** (`EXIT_USAGE_SITE_COUNT = 9`) |

The third is not structurally distinguishable from a cause legitimately spanning two sites — the "names nothing" cause really does span two — so it gets a count ratchet with a message telling the next author to give a new site its own cause. **A tripwire, not a proof**, and labelled as one in the source. Worth saying plainly: `find-session.py` documents that one site *used to be* spelled `sys.exit(2)`, so bypass (a) was a live path rather than a hypothetical.

### Three sentences that were wider than their code

- **F3 — the scope clause was false under two flags.** Appended unconditionally, so `--opencode-only` told the operator that the corpus it had just searched was excluded from the run, and `--claude-only` asserted an unreported cut in a corpus nobody opened. Now derived from the same conditions `archive_search` branches on. ⚠ Its own first draft then picked the verb from the *number of legs* and printed "the peer hosts **was** windowed too" — replaced with a colon list that agrees with nothing.
- **F2 — the machine-readable window still named a wider one than ran.** Round 1 fixed the human string and left `archive.window.since` on its own `.date()` call, so one object published `since: "2026-09-01"` beside a message naming `18:30:00`. Two fields of one window disagreeing is worse than the bug it half-fixed. One `window_stamp()` now serves both.
- **F1 — the shipped doc named the wrong stream.** SKILL.md still said the run "prints its window on stderr", true for one of four shapes after round 1 moved it, and sitting inside the `--live` block — the one shape where it is never on stderr.

### F5: five of round 1's nine fixes had no guard at all

Mutation-swept by the auditor, all five SURVIVED with the whole suite green: the scope clause, the `--since` stamp, the hoisted `ARCHIVE_STATS.clear()`, the conditional `--all-time` hint, and the `--deep` de-duplication. All five now have regression guards, plus boundary controls for the midnight case and for the subject-verb draft above.

🔴 **The round-1 commit reported a mutation sweep for one change without saying it covered only that one.** That is the finding behind the finding, and this commit says which fixes its sweep covers.

### Two self-inflicted repeats, recorded rather than buried

1. A guard written for F5 was **itself a spelled guard** on its first draft: `assert "7" not in message` failed against correct code, because the notice names the cutoff `2026-08-27`. Second sighting of that trap inside this one change; it now reads the state (`NOT MEASURED`), not a character.
2. SKILL.md's own measurement table wrote windows as `--since 3d`. **There is no relative-date parser** — every one of those exits 2 with `bad --since date`. They are durations in prose and now say so.

### Gate — base `39c31521`, tree `407ae705`

| tier | result |
|---|---|
| `scripts/gate.sh --tier both` (dev host) | pytest `7 failed, 12934 passed`; node **PASS** 1449/1449 |

The 7 are the inherited failures, unchanged. **No 8th.** The auditor independently ran both tiers on `ccb974bf` including the sandbox derivations one at a time, and confirmed the same 7 with `panic: test timed out` count 0.

### Ledger

`round 2 · payload lines changed THIS round: 90 (find-session.py +79/-32, SKILL.md +11/-6) · scaffolding +204/-11 · since round 1: 245`

Payload changed again, so the attribution gate does not fire.

```audit-claims round=2 audited=ccb974bf..407ae705
1. `_exit_usage_sites` now collects EXIT_USAGE sites module-wide and in three spellings (`return` in main, `return` in a helper, `sys.exit(EXIT_USAGE)` anywhere), closing two measured bypasses of the round-1 gate.
2. `EXIT_USAGE_SITE_COUNT = 9` ratchets the number of exit-2 sites, as a tripwire for the third bypass (a new site filed under an existing cause) which is not structurally detectable.
3. `unmeasured_legs()` derives the notice's excluded-leg list from the same conditions `archive_search` branches on, so `--claude-only` and `--opencode-only` no longer name legs that did not run.
4. The excluded-leg list is rendered as a colon list with no subject-verb agreement to get wrong.
5. `window_stamp()` is the single spelling of a cutoff, used by both the human notice and `archive.window.since`, so the two cannot disagree.
6. SKILL.md now says the window follows the reader and names all three destinations (stdout human, stderr under bare `--json`, `archive.window` under `--live --json`).
7. SKILL.md's measurement table no longer writes windows as `--since 3d`; it states that `--since` takes YYYY-MM-DD only and that `3d` exits 2.
8. Regression guards added for all five round-1 fixes that survived mutation: scope clause, `--since` stamp, hoisted `ARCHIVE_STATS.clear()`, conditional `--all-time` hint, `--deep` de-duplication.
9. A multi-bad `hosts` report names every offender's own type, not just the alphabetically first.
10. `--help` and a comment no longer advertise the retired 30 s figure; the invariant comment about the printed window now covers time-of-day, not only the date.
11. A guard written this round was corrected from a character search to a state assertion after it failed against correct code.
```

ZacxDev:
## Audit round 3 (delta, blind, `ccb974bf..407ae705`) — 4 🟡, no 🔴. Fixed in `adb181e9`.

All eleven round-2 claims verified **fixed or improved**, and every self-reported number re-derived exact (the payload ledger, `EXIT_USAGE_SITE_COUNT = 9`, and the "both KILLED where both previously SURVIVED" matrix all reproduced independently).

### F2 is a real bug, and it PREDATES this PR

`find-session.py <term> --claude-only --opencode-only` **searches no corpus at all and exits 0** with `No sessions matched` — for a term with hundreds of real hits. `archive_search` skips opencode under the first flag and skips both the local Claude walk *and* `search_peers` under the second. That is a false corpus-wide absence, the one thing this skill's own 🔴 rule forbids, and it is present unchanged at `39c31521`.

What **this PR** added was a notice that then described the never-opened opencode corpus as "windowed too and NOT in this count" — i.e. as searched. Now **refused as a usage error**, the same call `--since` + `--all-time` gets. That also makes `[--claude-only | --opencode-only]` in the usage synopsis true for the first time: argparse never enforced that notation.

🔴 **And it retracts a claim round 2 made.** That round's docstring said `unmeasured_legs` was "derived from the same conditions `archive_search` branches on, **so the two cannot drift**". False as a mechanism claim — they are two open-coded copies, and they already disagreed on exactly this pair. Refusing the pair removes the disagreement rather than re-deriving it.

### Three guards narrower than the sentences describing them — all three measured surviving

| mutant | at `407ae705` | at `adb181e9` |
|---|---|---|
| `unmeasured_legs` → `return ()` (disclosure deleted wholesale) | 177 passed | **3 failed** |
| `DEFAULT_SINCE_DAYS 12 → 9` (doc left saying 12 in four places) | 177 passed | **1 failed** |
| `raise SystemExit(EXIT_USAGE)` — a fourth spelling | 177 passed | **2 failed** |

Unmutated control: **180 passed**. Each mutant is killed by its own guard's assertion.

- **F1** — the scope-clause guard asserted only *absences*, and its `default` case asserted **nothing at all** (no absences to check). Deleting the whole disclosure left it green, while the code comment calls that unreported cut "both the larger fraction and the longer reach". Asserts presence first now.
- **F3** — 🔴 **the number this change is named after was the one number in the shipped doc that nothing enforced**, in a file that already pins the exit-code table verbatim and the `ARCHIVE_ONLY_FLAGS` list. Now pinned by shape, so a fifth mention is covered without anyone remembering the gate exists.
- **F4** — `raise SystemExit(EXIT_USAGE)` walked through **both** gates at once, because the count ratchet reads the same collector: a blind spot there blinds both. Added. The docstring's "every line in the MODULE that can hand `EXIT_USAGE` back to the shell" was a completeness claim the body did not deliver; it now says FOUR spellings and **names the three it still cannot see** so nobody reads silence as coverage.

### One of my own fixes was permanently-red on its first draft

The F3 gate's first regex (`\d+-day`) also matched the measurement table's **comparison** rows — "a 3-day window", "a 30-day default would be a NO-OP" — which must never equal the default. It went red on correct prose the moment it was written. It now matches the default's own two phrasings and carries a **negative control** proving a comparison row is not picked up.

### From the once-per-ladder re-derivation

Ranges cannot see a claim an early commit wrote true and a later commit of the same PR falsified, so this was run against the whole head:

- Two **superseded** live-scan figures (`1.82 s`, `"1.8 s"`) corrected. One had been **half-edited last round** — the `30.1 s` replaced, the `1.8 s` beside it left.
- The corpus counts are now labelled **dated snapshots**: the walked set was 924 on 2026-09-08 and passed 930 within hours, so a comment stating 924 as a flat fact goes stale by the day. The **ratios** the argument rests on are unaffected.
- Everything else held: the "nine sites" count, the cross-referenced test names (this round fixed a misspelling), `5,954 − 924 = 5,030`, the ratios, `487 of 707`, all nine referenced paths, and SKILL.md's table matching the code's digit for digit.

⚠ Adding the tenth exit-2 cause required editing `EXIT_CONTRACT`, `EXIT_2_CAUSES`, the shipped doc **and** `EXIT_USAGE_SITE_COUNT` together — the ratchet working as designed. It caught a doubled "or" in the joined sentence on the first attempt.

### Gate — base `39c31521`, tree `adb181e9`

| tier | result |
|---|---|
| `scripts/gate.sh --tier both` (dev host) | pytest `7 failed, 12937 passed`; node **PASS** 1449/1449 |

The 7 are the inherited base-red set, unchanged. **No 8th.** ⚠ Log-reading trap the auditor named: `grep "RESULT:"` on the pytest log hits `test_bash_guard.py`'s own `RESULT: all good` before the suite verdict — read the summary block, not the first match.

### Ledger

`round 3 · payload lines changed THIS round: 33 (find-session.py +32/-8, SKILL.md +1/-1) · scaffolding +99/-10 · since round 1: 278`

Payload per round: 155 → 90 → 33. Shrinking, but non-zero, so the attribution gate does not fire and the ladder continues.

```audit-claims round=3 audited=407ae705..adb181e9
1. `--claude-only` with `--opencode-only` is now refused as a usage error instead of searching no corpus and reporting "no sessions matched" at exit 0.
2. The exit-2 sentence, `EXIT_2_CAUSES`, the shipped doc row and `EXIT_USAGE_SITE_COUNT` were updated together for that tenth cause; the count is now 10.
3. `_exit_usage_sites` also collects `raise SystemExit(EXIT_USAGE)`, and its docstring now states FOUR spellings and names the three it cannot see instead of claiming to cover every line.
4. `test_every_WINDOW_LENGTH_in_the_doc_is_DEFAULT_SINCE_DAYS` pins the shipped doc's default-window literals to `DEFAULT_SINCE_DAYS`, matching the default's own two phrasings and not the table's comparison rows.
5. That gate carries a positive control (it can see a wrong number) and a negative control (it does not match a comparison row).
6. `test_the_scope_clause_names_only_legs_that_actually_RAN` now asserts which legs ARE named, not only which are absent, in all three parametrised cases.
7. Round 2's claim that `unmeasured_legs` and `archive_search` "cannot drift" is retracted in the source; they are two open-coded copies and the disagreement was removed by refusing the flag pair.
8. The superseded 1.82 s / 1.8 s live-scan figures are corrected to the current measurement.
9. The corpus counts are labelled dated snapshots of a growing corpus rather than flat facts.
10. The ratchet's failure message and the site-collector's empty-set assertion no longer say "main", since the collector is module-wide.
```

ZacxDev:
## Audit round 4 (delta, blind, `407ae705..adb181e9`) — 3 🟡, 3 🟢, no 🔴. Fixed. **Gate is now fully GREEN.**

### The headline finding was a sentence I wrote last round asserting coverage the same round had measured as absent

Round 3 wrote *"the count ratchet is the backstop if one appears"* about the spellings the site collector cannot see — **four lines below its own sentence** saying the ratchet *"uses this same collector, so a blind spot here blinds both gates at once"*. Measured: `_USAGE_ALIAS = EXIT_USAGE` + `sys.exit(_USAGE_ALIAS)` is a reachable exit 2, named nowhere, with all 180 tests green.

🔴 **The fix is not a third justification and not a wider enumeration.** Two rounds running, a spelling nobody had thought of walked through — the next one is always outside the list. So the hazard is *removed*: `test_the_module_EXITS_ONLY_BY_RETURNING_from_main` forbids `sys.exit` / `raise SystemExit` / bare `exit` anywhere but the `__main__` guard, which the module already satisfied. `return EXIT_USAGE` becomes the only spelling that can reach the shell, so enumerating it is **complete** rather than merely wide. The false clause is deleted with its history kept, so nobody derives a fourth.

### The seam: two mutation-clean components, one edit from deleting the feature

Round 3 hardened `unmeasured_legs` and asserted on `window_notice` **directly**, so no test ever built the combined state `main → _window_line → unmeasured_legs → window_notice → printed line`. Measured: `legs=()` at the one production call site left **all 283 tests across six files green** while every real archive run stopped naming the peer-host and opencode legs — the disclosure the code comment itself calls *"both the larger fraction and the longer reach"*. Now pinned through `main`, on the printed line **and** on `archive.window.message`, which a different call produces.

### The doc gate was walkable by an innocent reword

Measured in two steps: reword two of four mentions into an equally natural third shape (`"a 12-day default"`) → gate stays green, nothing warns the editor; then move the constant to 9 and update only the shapes the old regex could see → **180 pass** while the shipped headline says 12 and the tool prints 9. Exactly the defect the gate was written for. It now matches every window length and excuses the comparison rows through an enumerated `WINDOW_COMPARISON_DAYS`, each with its reason, and asserts the doc states the real default at all.

| mutant | before | after |
|---|---|---|
| `sys.exit(<alias>)` | 180 passed | **1 failed** (new structural gate) |
| `legs=()` at the call site | 283 passed | **3 failed** (three seam cases) |
| reword + partial constant update | 180 passed | **1 failed** (widened doc gate) |

Unmutated control: **185 passed**.

### ⚠ One finding deliberately NOT fixed

Round 3's claim that `unmeasured_legs` and `archive_search` "cannot drift" is still in the source. The audit found it imprecise but **not load-bearing** — `test_opencode_only_and_claude_only_partition_the_corpus` does catch behavioural drift between them, verified by mutating `archive_search`'s opencode condition. Rewriting that sentence a third time is what the "stop reaching for a better rationale" rule forbids. Recorded rather than re-drafted.

### Gate — and the base has moved twice

🔴 **`GATE: RESULT=PASS exit=0`.** pytest **13,139 passed / 0 failed** (floor 12,927), node **PASS** 1449/1449. Base `5b564844`. **The first fully green run on this branch.**

Getting there took two rebases and turned up two things worth recording:

1. **The 7 "inherited" failures are gone** — `main` fixed them in #1392 (`the 7 gate failures are nixpkgs drift, not a PR — re-derive the opencode pin, and re-key the age tamper verdict that 1.3.2 broke`). An independent session reached the same diagnosis from the other direction. Every prior comment on this PR reporting "7 failed, N passed" was correct at the time and is now superseded.
2. **A run at `bd796162` showed an 8th failure and then a content-gate failure — neither was this PR's.** The 8th (`test_a_launched_pane_gets_a_PATH_THAT_CAN_FIND_claude`) passes 3/3 in isolation, is in a file this diff cannot reach, and failed only in a run whose wall time inflated to 26.5 min under load average 65 — a load flake, and the test has a real timing dependency worth removing separately. The content-gate failure was a **client subdomain committed to this public repo** in `claudedocs/handoff-civitai-app-fleet.md`; it was introduced upstream, and a later commit on `main` had already removed it. Rebasing forward cleared it.

⚠ **Two measurement errors of my own, stated because either would have produced a confident wrong report:** I checked the refs with an **unbraced** `git show $r:path`, hitting the zsh history-modifier trap — it reported `0 matches` for every ref while the literal was plainly in the file (`${r}` returns 1). And I nearly reported the breach as live on `main` off a `merge-base --is-ancestor` check, which answers whether the *introducing* commit is an ancestor and says nothing about whether a later commit removed the line.

### Ledger

`round 4 · payload lines changed THIS round: 12 (find-session.py +7/-1, SKILL.md +5/-2) · scaffolding +170/-15`

Payload per round: 155 → 90 → 33 → **12**. Converging, non-zero, so the attribution gate does not fire.

⚠ **Sha mapping after the rebases** — the branch was replayed twice, so the shas in earlier claims blocks no longer resolve. Round 4 audited `adb181e9`, now `0e4876d0`; its fixes were `3f6341aa`, now `993a7564`. Contents are identical (the rebases touched none of this PR's files; `git diff origin/main...HEAD` is exactly the 6 files).

```audit-claims round=4 audited=0e4876d0..993a7564
1. `test_the_module_EXITS_ONLY_BY_RETURNING_from_main` forbids `sys.exit`, `raise SystemExit` and bare `exit` outside the `__main__` guard, making the `return EXIT_USAGE` enumeration complete rather than merely wide.
2. The false "the count ratchet is the backstop" clause is deleted, and its history recorded so no replacement justification is invented.
3. The `main` -> printed-notice seam is pinned by three cases that drive `main` and read the printed line and `archive.window.message`, rather than calling `window_notice` directly.
4. `_WINDOW_LITERAL_RE` now matches every window length in the shipped body, with comparison rows excused by name through `WINDOW_COMPARISON_DAYS`.
5. That gate also asserts the shipped body states the real default at all, so rewording every mention away fails.
6. The source comment no longer states a `return EXIT_USAGE` site count; it points at `EXIT_USAGE_SITE_COUNT`, which is asserted.
7. The collector docstring no longer says THREE spellings in one line and FOUR in another.
8. The shipped SKILL.md now carries the dated-snapshot warning for its corpus counts, not just the private source comment.
9. Round 3's "cannot drift" sentence is deliberately left unrewritten, having been found imprecise but not load-bearing.
10. The branch was rebased onto 5b564844; the full gate is RESULT=PASS with 13,139 pytest passed and 0 failed, and the previously-reported 7 inherited failures were fixed upstream in #1392.
```

ZacxDev:
## Audit round 5 (delta, blind, `0e4876d0..993a7564`) — 5 🟡, no 🔴. Fixed in `35fabed8`. Gate **PASS**, 13,144 passed / 0 failed.

### 🔴 The completeness claim in this spot has now been false THREE ROUNDS RUNNING

- **Round 3** wrote *"the count ratchet is the backstop if one appears"* — false; the ratchet reads the same collector.
- **Round 4** caught that, and replaced it with a **stronger** false claim — *"`return EXIT_USAGE` is the only spelling that can reach the shell, so enumerating it is complete"* — **deleting the honest disclosure list in the same edit.**
- Round 5 measured seven live exit-2 routes, suite green at 185/185 for all but the control.

🔴 **The decisive one is `return EXIT_USAGE if … else …`** — it *is* a `return`, the shape the sibling gate permits by design, so round 4's claim failed on its own terms.

**No fourth justification has been written.** The claim now reads "narrower, not complete", the disclosure list is restored, and the docstring carries the before/after matrix. Widening the detector to *any* receiver — an aliased `import os as _os` and `build_parser().exit()` both slipped a by-name check — moved three rows:

| spelling | before | after |
|---|---|---|
| `_e = SystemExit(EXIT_USAGE); raise _e` | SURVIVED | **KILLED** |
| `_os._exit(EXIT_USAGE)` (aliased import) | SURVIVED | **KILLED** |
| `build_parser().exit(EXIT_USAGE, "…")` | SURVIVED | **KILLED** |
| `_bye = sys.exit; _bye(EXIT_USAGE)` | SURVIVED | SURVIVED — **named** blind spot |
| `build_parser().error("…")` | SURVIVED | SURVIVED — **named** blind spot |
| `return EXIT_USAGE if a.any else EXIT_OK` | SURVIVED | SURVIVED — **named** blind spot |

The last three stay uncaught deliberately: catching them means flagging every call to a local name, every `.error(`, and every conditional return — breadth this module cannot carry. Naming them is the point.

### 🔴 argparse exits 2 with no mutation at all, and nine enumerated causes missed it

`--nope`, `--limit abc`, `--tail x` — all rc 2 today, measured. The exit-2 sentence an agent branches on enumerated nine causes and named none of them: the one a caller is likeliest to hit. Now in `EXIT_CONTRACT`, in `EXIT_2_CAUSES` with three probes, and in the shipped doc.

⚠ `_run` had to learn to catch `SystemExit` to observe it — **the harness could not previously see the one exit class this file exists to reason about.**

### The positive control was validating a hand-copy of the gate

`test_the_EXIT_SPELLING_gate_can_SEE_a_stray_exit` re-declared the detector inline. Measured: blinding the real gate (`"exit"` → `"exitt"`) left **all 185 passing**, while a genuine stray `sys.exit(EXIT_AMBIGUOUS)` — for which that gate is the **sole** catcher, since the site collector only matches an arg named `EXIT_USAGE` — went unreported. One `_stray_exit_lines`, two callers.

The `__main__` allowance was also `ast.dump(...).find("__main__")`, a substring search, so `if '__main__' not in str(argv): sys.exit(…)` was excused as the entry point. Matched by shape now, with a spoofing case in the control.

### A seam argument pinned on one side, and a test that derived its expectation from the implementation

`legs=` was pinned; its sibling `claude_leg_ran=` **on the same call** was not — flipping it to `True` left 185/185 green while `--opencode-only` reported the Claude corpus as merely *uncounted* when it was never opened. Worse, its only coverage **computed the expectation from the implementation** (`claude_leg_ran=not a.opencode_only`), so the two could never disagree. Expectation is a literal now, and the seam cases gained the `--opencode-only` row they were missing.

### "the body states the real default at all" was satisfied by a comparison row

Measured, both surviving: set the constant to 3 (or to 30) and reword the four real mentions to name no number — the body then states the bound **nowhere**, while the measurement table's own foil (`a 3-day window`, `🔴 A 30-day default would be a NO-OP`) satisfies the check *about the actual default*. `DEFAULT_SINCE_DAYS` may no longer BE a comparison value, and the default must now appear in the default's own phrasing.

Plus `R5_GREEN_AT_AUDITED_TIP` (four guards shipped without the file's own green-at-base ledger, so a reader consulting R3's and R4's would take them for red-at-base regression coverage) and two paragraphs contradicted by the code directly beneath them.

### Gate — base `5b564844`, tree `35fabed8`

| tier | result |
|---|---|
| `scripts/gate.sh --tier both` (dev host) | pytest **PASS**, `13,144 passed / 0 failed`; node **PASS** 1449/1449 |

The auditor independently ran the same tier on the **merged** tree (`993a7564` + `origin/main 0477e52a`) and also got PASS, 13,139/0.

⚠ **The sandbox tier is still UNMEASURED since the rebase** — load has been 40–65 with several sessions' gates in flight, and a contended red there is not trustworthy. It must be run one derivation at a time before merge.

### Ledger

`round 5 · payload lines changed THIS round: 8 (find-session.py +7/-4, SKILL.md +1/-1) · scaffolding +223/-59`

Payload per round: 155 → 90 → 33 → 12 → **8**. Converging, non-zero.

```audit-claims round=5 audited=993a7564..35fabed8
1. The "enumerating `return EXIT_USAGE` is complete" claim is replaced with "narrower, not complete", the deleted disclosure list is restored, and a before/after mutation matrix is recorded in the docstring.
2. `_stray_exit_lines` matches `.exit`/`._exit` on ANY receiver, moving `raise <name>`, aliased `os._exit` and `parser.exit()` from SURVIVED to KILLED.
3. Three spellings remain uncaught and are NAMED as the standing blind spot rather than covered by a claim.
4. argparse's own rejection of a malformed command line is named in `EXIT_CONTRACT`, in `EXIT_2_CAUSES` with three probes, and in the shipped doc row.
5. `_run` catches `SystemExit` so a raised exit is observed as an rc rather than erroring the probe.
6. `_stray_exit_lines` is one function called by both the gate and its positive control, so blinding the gate blinds its control too.
7. The `__main__` guard allowance is matched structurally, and the control includes a spoofed-`__main__` case.
8. `claude_leg_ran` is pinned by a literal expectation rather than one computed from the implementation, and the seam cases cover `--opencode-only`.
9. `DEFAULT_SINCE_DAYS` is asserted not to be a `WINDOW_COMPARISON_DAYS` value, and the default must appear in the default's own phrasing.
10. `R5_GREEN_AT_AUDITED_TIP` records the four green-at-base guards, pinned two-way like R3's and R4's.
```

ZacxDev:
## Audit round 6 (delta, blind, `993a7564..35fabed8`) — 3 🟡, 2 🟢, no 🔴. Fixed in `fb1aef02`.

🔴 **THE SANDBOX TIER IS NOW MEASURED AND GREEN** — the outstanding gap since the rebase.

### The fix for the third false closure claim was itself the FOURTH

Round 5 replaced *"enumerating is complete"* with *"THE LAST THREE ROWS ARE THE STANDING BLIND SPOT"* — naming three and implying the set closed. Round 6 measured two more, neither caught **nor named**, both exiting 2 from a real script:

| spelling | why it was missed |
|---|---|
| `quit(EXIT_USAGE)` | the sibling builtin of `exit`, which the detector already matched **by name** |
| `raise <recv>.SystemExit(…)` | the same receiver-blindness round 5 had just fixed for CALLS, left open for RAISES |

Both now caught — one token and one clause, **neither costing the breadth the docstring cited as the reason the residue had to stay uncaught.**

🔴 **The list is no longer asserted complete, in either direction.** Four consecutive rounds wrote a closure claim in this one docstring and all four were false. It now records what was measured, states that what is uncaught is whatever nobody has thought of yet — not a set anyone can enumerate — and instructs successors not to assert one.

### An over-reaching guard that was green only because the module has no `raise`

The `raise <Name>` arm flagged **every** bare-name raise: `raise NotImplementedError`, `raise err`, and a re-raised `except OSError as e: raise e`. None is an exit, and the failure message told the author to "spell an exit as `return <EXIT_*>` from main" — incoherent advice for those. `find-session.py` has no `raise` today, so this was **latent**: one ordinary `raise e` around the `RUN` subprocess would have turned the gate red for the wrong reason, which is how a gate gets deleted. Narrowed to `SystemExit`, with three ordinary raises as negative controls.

### The else-branch trap this file already documents for its sibling helper

`_main_guard_spans` used `(n.lineno, n.end_lineno)`, and `ast.If.end_lineno` spans the `orelse` — so `else: sys.exit(9)` beneath a real `__main__` guard reported `stray=[]`, while the identical statement under `if True:` was caught.

🔴 **This file already carries a 🔴 comment documenting that exact trap for `_tail_path_lines`, with its own regression test.** Round 5 rewrote this helper from scratch to close the substring-spoof hole and did not carry the lesson across. Spans are the body now, with a control.

Plus: the R5 ledger gate was **vacuous when emptied** while claiming to be "two-way, like R3's and R4's" (measured: emptying R3's or R4's fails theirs; emptying R5's passed), and the exit-2 row said argparse rejects "before `main` runs" — false, `main`'s first statement *is* `parse_args`. It now carries the fact round 5 learned in its own harness and didn't write down: this is the only exit 2 the module **raises** rather than returns.

Measured against an unmutated control of **190 passed**:

```
quit(EXIT_USAGE)                 SURVIVED -> KILLED
raise <recv>.SystemExit(…)       SURVIVED -> KILLED
raise NotImplementedError        negative control: NOT flagged
```

### Gate — both tiers, on `fb1aef02`

| tier | result |
|---|---|
| sandbox `checks.x86_64-linux.pytests` | **PASS** — `collected=21189 passed=21187 skipped=2 failed=0` (floor 20,342), 1,719 lines of streamed log so not a cached silence |
| dev-host `scripts/tests` + node | `13,144 passed / 0 failed`; node **PASS** 1449/1449 |

⚠ **The dev-host verdict is ASSEMBLED FROM THREE RUNS, not one sweep.** The box has sat at load 80–90 for hours: `gate.sh` was SIGTERMed at its **own internal 3600s cap** (`exit=124` → `RESULT: FAIL (exit=143)`) after 38m21s, having completed 20 of 28 targets with **zero failures anywhere in the log**. The 7 targets it never reached were then run directly (`RESULT: PASS`, 0 failures), and the one it killed mid-run was re-run (`RESULT: PASS`, 0 failures). Every declared hermetic target has now run green on this commit — but it took three invocations, and saying otherwise would overstate it.

Two earlier failures under that load were investigated rather than waved off: `test_a_partial_run_is_declared_where_gate_sh_actually_LOOKS` (`subprocess.TimeoutExpired`, `orig_timeout=120`, nested `run-tests.sh`) and `test_a_scope_FILTERED_snapshot_of_a_denied_scope_ships_nothing` (tar-bytes). Both pass in isolation; neither is in this PR's six-file diff.

⚠ Two `run-tests.sh` gotchas that cost a cycle: `--targets` takes a **space-separated** list (a comma list is rejected as "not a target in the hermetic set", which reads like a missing target rather than a syntax error), and `scripts/devhost-tests` is in `DEVHOST_TARGETS`, needing `--set all`.

### Ledger

`round 6 · payload lines changed THIS round: 2 net (find-session.py +6/-4 comment + EXIT_CONTRACT string, SKILL.md +1/-1) · scaffolding +90/-15`

Payload per round: 155 → 90 → 33 → 12 → 8 → **2**. Non-zero, so the attribution gate does not fire — but the trend, and the fact that every 🟡 since round 3 has been a guard defect rather than shipped behaviour, is the real signal.

```audit-claims round=6 audited=35fabed8..fb1aef02
1. `_stray_exit_lines` now catches `quit(...)` and `raise <recv>.SystemExit(...)`, the two spellings round 6 measured as neither caught nor named.
2. The docstring no longer asserts any closure over the blind-spot set, records the measurement instead, and instructs successors not to assert one.
3. The `raise <Name>` arm is narrowed to `SystemExit`, with `raise NotImplementedError`, a re-raised `except ... as e: raise e`, and `raise err` as negative controls.
4. `_main_guard_spans` uses the guard's BODY span, so a stray exit in the `__main__` guard's `else:`/`elif` body is no longer excused; a control covers it.
5. The R5 ledger gate asserts the tuple is non-empty, matching R3's and R4's behaviour it claimed to share.
6. The R5 ledger comment no longer says round 5 "added" four guards; it authored one, modified two and left one untouched.
7. The exit-2 row says argparse rejects inside `main`'s first statement, and states that this is the only exit 2 the module raises rather than returns.
8. Both sandbox derivations were built one at a time on this branch and are green; the dev-host tier is green but assembled from three runs under sustained load.
```

PR #1391, opened 2026-09-08
feat(skills): civitai-app-release — the author-side release path, as scripts

ZacxDev:
Adds `civitai-app-release` — the author-side release path for the seven Civitai App Blocks, with the logic in **scripts rather than prose**.

## Why it exists

Releasing and submitting those apps had **no skill coverage**. The substantial `app-blocks` skill in talos-infra is platform-OPERATOR oriented — Tekton, Flipt, the publish-request state machine — and only loads with *that* repo as cwd, so it never fires while working in an app repo.

It is scripts, not prose, because prose telling the reader to be careful had already failed twice in a single session:

1. **`civitai app status <slug>` returns only the NEWEST submission.** A withdrawn duplicate of the same version sat on top of a healthy `approved/building` row, so the per-slug view reported `withdrawn` for an app that was building fine.
2. **A hand-rolled parser fell through on a state it had never heard of.** It enumerated `live|failed|rejected|building|approved|pending`, met the real state `deploying`, and silently reported `approved` — under-reporting progress for an hour without ever erroring.

## What's in it

**`app_state.py`** reads the LIST view and **raises** on an unrecognised state token. The failure mode in (2) was *silence*, so being loud is the entire fix.

**`preflight.py`** refuses a submit the platform builder would fail on:

| Check | The defect it corresponds to |
|---|---|
| version lockstep | a 7-app batch bumped the manifest and left `package.json`; one shipped that way |
| `buildCommand` ↔ lockfile | the builder picks its install command from `buildCommand`'s **first word**, so a pnpm manifest with only a `package-lock.json` is a guaranteed failure **CI cannot see** — `.github/` isn't in the bundle |
| exactly one lockfile | a repo carried both |
| the platform's anchored `BUILD_COMMAND_RE` | `pnpm  run build` (two spaces) is rejected by the platform while first-token parsing accepts it |
| above the submit floor | the floor is the highest version **on record** — a withdrawn submission still occupies it |

Both read `CIVITAI_STATUS_FILE`, so the tests need no network and no auth — and can't go accidentally green in a sandbox lacking the CLI.

## Verification

**19 tests, every case a defect that actually occurred.** Mutation battery: **6 mutations, each killing exactly one test — its own — with the other 18 green**, controls green either side, tree restored byte-identical.

Full-suite attribution was done by comparing failing **files** against a pristine `origin/main` control run. Pristine fails **7** (`analyze_service_index_backup`, `analyze_service_index_escrow_verify`, `opencode_engine`) — pre-existing and unrelated. The six gates this touches pass (291 tests).

⚠️ **The final full-suite run carrying these fixes was still in flight when this PR was opened.** I'll post the number as a comment rather than assert it here. Target is 7.

### Two wrong diagnoses on the way, worth recording

My first id-level comparison was scraping test names out of **tracebacks**, not failures, and produced a different plausible set each run. I built two diagnoses on it — an xdist-redistribution theory and a shared-`subprocess`-monkeypatch theory — and **both were wrong**. A same-named, same-count dummy file on pristine main caused *zero* extra failures, which killed the first; the second survived a fix that changed nothing (18 before, 18 after). Comparing failing **files** named the real cause immediately: two gates I had simply never run.

The `_RUN` seam introduced for the second theory stays — patching a shared module from a test is wrong regardless — and now carries the positive control it lacked: with the env seam cleared, `read_status` must actually reach `_RUN`.

## Two of this repo's gates caught real defects

- **`test_no_client_hostnames.py`** — my test fixture committed a live client subdomain to a **public** repo. Now `*.example.test`; nothing parses that column, so the fixture loses nothing.
- **`test_skill_tiers.py`** — the missing ledger entry, and the tier-A ceiling breach below.

## 🔴 Two ratchets raised, both deliberate

`LISTING_TOTAL_CEILING_CHARS` 10,800 → 11,192 and `TIER_A_CEILING_CHARS` 7,496 → 7,639.

**Both files say "do NOT raise it to make a new description fit."** Zach took both decisions with that sentence quoted back to him, after declining the demotion and tier-B alternatives. Each is documented at the line: what was weighed, that the rule still binds everyone else, and headroom pinned back to **0** so the next addition of any size reds the gate.

The tier-A comment records the honest cost: **that budget overflows by dropping the least-invoked skills' descriptions first**, and a brand-new skill is by definition the least invoked — so the first thing at risk is the skill the raise was taken for. If a third raise is ever proposed, the right answer is almost certainly a demotion.

Reviewers should feel free to push back on either raise; a demotion is still a one-line change.

## Note

This needs a **`home-manager switch`** to go live — global skills are `home.file` copies, so editing devrc alone does not install it.

https://claude.ai/code/session_01DULuahpBaAwuYb89Hg2pEv

### DIFF
--- a/claude/skill-tiers.json
+++ b/claude/skill-tiers.json
@@ -58,6 +58,7 @@
       "why": "the tool is NAMED, never described -- 'cairn doctor', 'cairn sync', 'cairn who'. The symptoms that could route here already belong to skills that name it in their own bodies: a refused read to `resume`, a stale index block to `analyze-service`, an entry that got huge to `prune-index`"
     },
     "check-clickup-addressed": { "tier": "A" },
+    "civitai-app-release": { "tier": "A" },
     "clawgate": { "tier": "A" },
     "clickup": { "tier": "A" },
     "devrc-dx": { "tier": "A" },
--- a/claude/skills/civitai-app-release/SKILL.md
+++ b/claude/skills/civitai-app-release/SKILL.md
@@ -1,0 +1,99 @@
+---
+name: civitai-app-release
+description: "Release and submit Civitai App Blocks (gen-matrix, sensei, model-benchmarking, playable-collections, custom-generators, app-requests, generate-from-model). Use for: civitai app submit, bump an app version, a refused submit, submission/deploy state, a pass across the app repos. Platform-side ops — Tekton, Flipt, deploy diagnosis — is the `app-blocks` skill in talos-infra."
+argument-hint: "<action> — state <app> [version] | floor <app> | preflight <dir> [floor] | submit <repo> | fleet"
+allowed-tools: Bash, Read, Edit, Write, Grep, Glob
+---
+
+# Civitai App Blocks — author-side release
+
+You are the app AUTHOR here. Platform operation — Tekton, Flipt flags, the
+publish-request state machine, deploy diagnosis — is `app-blocks` in
+`datapacket-talos`, and it only loads with that repo as cwd.
+
+## Use the scripts. They exist because prose failed.
+
+Two reads of platform state were wrong in one session, each looking like an
+answer rather than an error. Do not re-derive these by eye.
+
+```bash
+SKILL=~/.claude/skills/civitai-app-release
+
+python3 $SKILL/app_state.py <app> <version>   # authoritative state; exit 0 iff live
+python3 $SKILL/app_state.py <app>             # the submit floor
+python3 $SKILL/preflight.py <export-dir> [floor]   # refuse a doomed submit
+```
+
+`CIVITAI_STATUS_FILE=<dump>` makes `app_state.py` read a captured
+`civitai app status` instead of calling the CLI — use it to reason about a
+past state, and it is how the tests run offline.
+
+## The three facts that cost the most time
+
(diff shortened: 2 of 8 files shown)


ZacxDev:
**Full-suite result — the number the PR body deliberately did not assert.**

```
TOTAL collected=20928  passed=20919  skipped=2  failed=7
```

**7 failures, and the identical three files as the pristine `origin/main` control:**

| | Failing files | Count |
|---|---|---|
| pristine `origin/main` | `analyze_service_index_backup`, `analyze_service_index_escrow_verify`, `opencode_engine` | 7 |
| **this branch** | **same three** | **7** |

So this change contributes **zero** failures. Collection went 12,863 → 12,883 (+20: the 19 tests plus the positive control added during review of my own seam test).

Two notes for whoever merges:

1. **`main` is red today**, independently of this PR — 7 pre-existing failures across those three files. Worth someone's attention on its own: a permanently-red suite is the kind people learn to merge through.

2. **`rc=1` on the runner is expected** given (1), so don't read a non-zero exit as this branch failing. The number that matters is the file-level comparison above. Both full runs were also reported by my harness as "exit code 0" while the captured `rc` was 1 — a trailing command in my wrapper swallowed the status the first time, which is why the attribution here is by counted content rather than by exit code.

PR #1393, opened 2026-09-08
refactor(skills): civitai-app-release → civitai-app-fleet, and add the inventory

ZacxDev:
Follow-up to #1391, which framed this skill as *the release path*. That was too narrow.

## The session it came from disagrees with the framing

| That afternoon's work | Repo-passes |
|---|---|
| ready-to-work pass | 7 |
| guard backport | 6 |
| guard hole fix | 6 |
| version bumps | 5 |
| **submits** | **6** ← all #1391 covered |

~30 passes; the skill covered 6 of them. And release was the *easy* part. What was hard — and hand-rediscovered repeatedly — was fleet mechanics:

- `isolation: "worktree"` builds a worktree of the **CWD's** repo, not the target's
- base clones sit on other people's branches with uncommitted work
- `sensei`'s default branch is `trunk`; six are `main`
- one repo runs a **single** vitest project, six run two — `--project node` against it matches nothing and **runs nothing**
- the version-lockstep guard lives in `manifest.test.ts` in some repos, `version-lockstep.test.ts` in others

None of it was in the skill.

## `fleet.py` — a script, not a table, on purpose

```
app                   branch  checked-out                dirty ver     proj guard
gen-matrix            main    main                           0 0.9.0      2 version-lockstep
model-benchmarking    main    main                           0 0.4.4      2 manifest.test
app-requests          main    zach/app-requests-0.2.0-ga     0 0.4.1      2 version-lockstep
sensei                trunk   feat/orchestrator-adapter      1 0.1.21     2 manifest.test
generate-from-model   main    zach/wildcard-parent-origi     3 0.2.23     1 version-lockstep
```

Every column was a wrong assumption before it was a column, and a table recording them would have been stale within the day — three of those checked-out branches changed *during* the session that discovered them. Reading it live means the failure mode is "the script errors", not "the doc lies".

## 🔴 Its first real run found a bug in what #1391 merged an hour earlier

`preview-live` is a real deploy state (`w6-ui-dogfood`) that `DEPLOY_STATES` omitted, so `app_state.py` **raised on the actual `civitai app status` output**. The guard firing loudly on its own author is the design working exactly as intended.

Re-deriving the vocabulary from the platform rather than guessing a second time showed the set was wrong in **both directions** — it also contained `queued`, which occurs nowhere in the listing. A set that is too wide fails *silently*, accepting a renamed state as valid.

Both directions now have regression tests, mutation-verified: dropping `preview-live` kills two tests, re-adding `queued` kills one, tree byte-identical after. Membership is pinned with per-member provenance and a one-liner to re-derive it.

## `reference/fan-out.md`

The dispatch mechanics, all from things that went wrong: the cross-repo isolation trap, branch-off-freshly-fetched-origin (a live session moved a base clone underneath the work **four times** in one afternoon), one brief file rather than N paraphrased prompts, per-agent scratch dirs, evidence-not-"done", and attributing a suite failure by comparing failing **files** against a pristine control rather than scraping ids out of tracebacks.

It also records the most useful thing that happened: **two agents disagreeing.** One argued the canonical guard the other six had adopted could be walked around; testing that claim instead of reconciling it by preference found a real hole — the guard read what CI *installed* rather than what it *ran*, and passed 5/5 against a workflow that never invoked pnpm.

## Both ratchets re-pinned DOWNWARD

The reframed description is 22 chars shorter, so `TIER_A_CEILING_CHARS` 7,639 → **7,617** and `LISTING_TOTAL_CEILING_CHARS` 11,192 → **11,170**, headroom back to 0.

Lowering on a cut is the rule those files state, and it pays back part of yesterday's raise rather than banking the slack. The two override notes stay, since the raise still happened.

## Verification

- 22 tests pass (2 new state guards, both mutation-checked)
- all six affected gates pass — 293 tests
- **full suite: 7 failures**, the identical three files (`analyze_service_index_backup`, `analyze_service_index_escrow_verify`, `opencode_engine`) as a pristine `origin/main` control — **this contributes zero**

`rc=1` on the runner is expected: `main` is red today for those pre-existing failures, independent of this PR.

## Note

Still needs a **`home-manager switch`** to go live, and the directory rename means the old `~/.claude/skills/civitai-app-release/` will linger until then.

https://claude.ai/code/session_01DULuahpBaAwuYb89Hg2pEv


ZacxDev:
**Round 1 audit — findings and fixes.** Full adversarial audit run against `eab2ceb6`; fixes in `d6815fd1`.

The audit's headline is worth stating plainly: **`fleet.py`'s whole thesis is that an inventory must never state what it could not read, and it did exactly that in three columns — while being the only file in the PR with no tests.**

## 🔴 Both fixed

1. **`lockfile()` returned the literal `"none"` on git failure**, where every sibling reader returned `"?"`. A repo with a `pnpm-lock.yaml` it simply could not read reported as having *no* lockfile — the column that decides whether `buildCommand` and the committed lockfile agree, wrong in the direction that breaks a platform build. `_git` now returns `None` for FAILED, distinct from `""` for "ran, empty output", and every reader converts that to `UNREADABLE`.

2. **The exit code promised coverage it did not have.** The docstring said "non-zero if any repo could not be inspected"; the flag was set in exactly one place (`isdir`). A directory that existed but was not a git repository produced a full row of plausible defaults — `checked_out: "(detached)"`, `dirty_files: "0"` — and exit **0**. `0 dirty` is precisely the value that makes a caller proceed. Any unreadable column, non-repository, or failed fetch now sets it.

## 🟡 / 🟢 also fixed

- The unconditional **`git fetch` in seven SHARED clones** is now documented rather than discovered, reported per row (`ok`/`FAILED`/`skipped`), and opt-out with `--no-fetch`. A failed fetch left every column describing a stale ref while the file claimed the inventory "cannot rot".
- The broad `except Exception` **caught `app_state.UnknownState`** — muting the guard this skill exists to celebrate, through the entry point it tells you to run first. `preview-live` was found *only* because that guard reached a human. It now re-raises.
- `--no-platform` rendered as `-`, **a real deploy state** meaning "no deploy"; now `not-consulted`.
- `vitest_projects` claimed to be a **count** while testing for a key that cannot tell two from three → reports the class, `1` / `2+`.
- `REVIEW_STATES` had neither pin nor provenance while its sibling argued at length that a too-wide set fails silently — widening it passed a green suite. Provenance added.
- `failed` sat in the same uncovered fixture position `preview-live` had been in → now a fixture row.
- `resolve()`'s tie-break for two non-withdrawn rows: documented and pinned.
- Three prose claims corrected — the **"22 chars shorter"** attribution (20 were the description, 2 the shorter name), a present-tense **"254 chars of headroom"** that this very PR had moved to 0, and `SKILL.md` dropping the half of the `.env*` rule that **keeps** `.env.production`.

## Verification

`fleet.py` now has **11 tests**, every one driving a real git-failure path through the `_RUN` seam — the happy path hid all of these. **Six mutations, each killed by its own test**, controls green either side, tree byte-identical after. **302 pass** across the six affected gates. Live run reproduces the fleet shape, populates the platform column, and exits 0 on success / 1 against a workspace it cannot read.

## Not fixed, deliberately

- `test_control_the_tier_a_ratchet_can_go_red` is degenerate at 0 headroom (it passes trivially and cannot distinguish 0 from 300). **Pre-existing**, unchanged in shape by this PR, and fixing it means changing a control this PR does not own.
- `building` / `deploying` / `pending` could not be re-confirmed from today's listing — they were observed on submissions on 2026-09-08, which a later snapshot can neither confirm nor refute. The provenance comment says exactly that.

```audit-claims round=1 audited=eab2ceb65f0bcdf610baa54dcdf78076daed1687..d6815fd1
1. lockfile() and every other fleet.py reader now return UNREADABLE on git failure instead of a plausible default; _git returns None for failure, distinct from "" for empty output.
2. inspect() sets "error" for any unreadable column, a non-repository, or a failed fetch — not only for a missing directory — so main()'s exit code is non-zero whenever a row was not fully read.
3. The git fetch is documented as a write to shared clones, reported per row as ok/FAILED/skipped, and opt-out via --no-fetch.
4. The platform block re-raises app_state.UnknownState instead of catching it under a broad except.
5. --no-platform renders "not-consulted" rather than "-", which is a real deploy state.
6. vitest_projects reports the class 1 / 2+ rather than asserting a count it cannot measure.
7. REVIEW_STATES carries provenance for each member, matching the standard DEPLOY_STATES already had.
8. The STATUS fixture now contains a `failed` row; resolve()'s multi-non-withdrawn-row tie-break is documented and pinned by a test.
9. Prose corrections: the 22-char attribution, the present-tense 254-chars-of-headroom claim, SKILL.md's truncated .env* rule, and app_state.py's docstring path to the renamed test file.
10. fleet.py gained 11 tests, all exercising git-failure paths through the _RUN seam.
```

ZacxDev:
**Round 2 delta audit — findings and fixes.** Audited `eab2ceb6..d6815fd1`; fixes in `db08c7f0`. Round 2 was **not clean**, so the ladder continues.

## The headline finding was an over-correction round 1 introduced

**F1 🟡 — a failed `git fetch` erased the entire row.** Round 1 folded it into `error`, and `main()` prints an error row as a bare `!!` *instead of* the row. One unreachable remote — no SSH agent, VPN down, a single bad URL — collapsed all seven rows, and the tool `SKILL.md` says to start every bulk pass with emitted **no inventory at all**. Every column had been read successfully; they were merely possibly-stale, which is exactly what the warning said.

Round 1's behaviour was *silently stale*. Round 2's was *no data*. The second is worse.

A failed fetch is now a staleness **warning**: the row prints in full, `fetch` is a column in the table (it was a `--json` key only, while the docstring claimed "reported per row"), a stderr note names every stale row, and the exit code is still non-zero. `error` is reserved for a row that could not be **read**. Verified live against a clone with readable columns and a broken remote — full row, `FAILED`, exit 1.

## The rest

- **F2** — `--no-fetch` produces byte-identical staleness to a failed fetch; round 1 warned about one and left the other indistinguishable from fresh data. *Wider on the failure axis, narrower on the deliberate-skip axis, for the same hazard* — the exact shape the audit brief says to hunt for. `skipped` now appears in the same column and note. The docstring also now says running without touching anything needs **both** `--no-fetch` and `--no-platform`.
- **F3** — the round-1 `vitest_projects` test asserted the result was in `{"1","2+","?"}` — every value the function can return, so tautological — plus a spelled check on a docstring. A mutant returning `"1"` unconditionally **survived a green suite**: the exact `--project` trap the module documents.
- **F4** — five of round 1's own fixes were unpinned, proven by mutation: `main()`'s exit code (**no test called `main()` at all**), the `checked_out` / `dirty_files` / `lockstep_guard` None-guards, the `UnknownState` re-raise, and `not-consulted`. **All eight of the audit's survivors now die**, each by its own test, controls green either side, tree byte-identical.
- **F5** — round 1 replaced a stale "254 chars of headroom" claim with a **new false one**, crediting the loss to this PR when #1391's raise consumed it and both ceilings have been at 0 since. A false claim replaced by a differently false claim, in the file whose stated purpose is that the audit trail "cannot rot into a lie".
- **F6** — `REVIEW_STATES` got a comment saying "same standard as DEPLOY_STATES" while that standard is provenance **plus** an exact-set pin. Widening it with `queued` or `escalated` survived a green suite. The round closed a named hazard with a sentence; it now has the pin.
- **F7** — my numbers were wrong: "11 tests" was 8, and "302 pass" matched no file combination. This round's figures are measured, below.
- **F8** — the `# noqa: PLC0415 — optional` comment was false once the import moved out of the `try`; `fan-out.md` still said "vitest project count".

## Two of this round's own test fixes are worth naming

The round-1 test asserting `"fetch failed" in row["error"]` **went red on the fix — correctly**, because it had pinned the over-correction rather than the behaviour.

And `fleet.main()` does a plain `import app_state` while the test module loads the same file under a private name: **two module objects, two distinct `UnknownState` classes**, so `pytest.raises` did not catch what fleet raised. The test now pins the module name.

## Verification (measured, not estimated)

38 tests in `test_civitai_app_fleet.py` · **309 pass** across the six affected gates · all 8 previously-surviving mutants killed, one per test · live run against the seven real clones renders the full table and exits 0 · the failed-fetch case verified against a purpose-built clone.

## Ledger

`round 2 · payload lines changed THIS round: 77 (fleet.py 75, fan-out.md 2) · scaffolding: 149 · since round 1: 612`

Non-zero payload, so the two-consecutive-zero-payload gate does not fire.

```audit-claims round=2 audited=d6815fd1edaf60fbfc3b24c434be2c223e99a795..db08c7f0
1. A failed git fetch no longer sets "error" and no longer replaces the row; it is a staleness warning with the row printed in full.
2. `fetch` (ok|FAILED|skipped) is a column in the human table, not only a --json key, and a trailing stderr note names every possibly-stale row.
3. --no-fetch staleness is surfaced as loudly as a failed fetch, and the docstring states that touching nothing requires both --no-fetch and --no-platform.
4. main() exits non-zero for an unreadable row OR a failed fetch, and zero when --no-fetch merely skipped.
5. vitest_projects has a real test distinguishing 1 from 2+ from unreadable, replacing a tautological set-membership assertion.
6. The five previously-unpinned round-1 fixes are pinned: main()'s exit code, the checked_out/dirty_files/lockstep_guard None-guards, the UnknownState re-raise, and not-consulted.
7. REVIEW_STATES has an exact-set pin matching the standard its comment claims.
8. The false "254 headroom consumed by this PR" attribution is corrected to #1391's raise; the noqa comment and fan-out.md's "count" phrasing are fixed.
9. The round-1 test that pinned the over-correction was rewritten to pin the behaviour instead; the two-module-object hazard in the UnknownState test is pinned by name.
```

ZacxDev:
**Round 3 delta audit — findings and fixes.** Audited `d6815fd1..db08c7f0`; fixes in `f769372c`. Round 3 was **not clean**, so the ladder continues.

Two of the six findings are this ladder's recurring shapes, and both are mine.

## 🟡 The three that matter

**🟡2 — I overstated the bug I fixed by 7×.** Round 2's docstring said "a single unreachable remote collapsed all seven rows". The auditor built a seven-repo fixture with exactly ONE broken remote: **one** row collapsed, six printed in full. All seven collapse only under a cause common to all seven — no network, no SSH agent, VPN down. The bug was real; the blast radius written down for it was not, and a reader told the rows are coupled when they are independent will mis-triage the next occurrence. Corrected, with the measurement recorded beside it.

**🟡3 — round 2's own rule, applied to one axis only.** It wrote "discarding probably-correct data is a worse failure than serving it with a caveat" and then applied it to the *fetch* axis alone: a row with ONE unreadable column still lost its other ten readable facts to a `!!` line. A repo with no `vite.config.ts` on the default ref was enough. Now only a row with NO columns — a missing checkout or a non-repository — is replaced; an unreadable column prints `?` in place and still sets `error` and a non-zero exit. The platform loop likewise no longer skips on `error`, only on an unreadable VERSION, so rows that had platform state available stop being denied it.

Same finding's second half: `vitest_projects` returned `UNREADABLE` for a file that was merely **absent** — the exact absent-vs-unreadable conflation the sentinel exists to prevent. It now checks the ref first and reports `no-vite`.

**🟡1 — four of the five behaviours round 2 shipped were unpinned**, which is the same F4 defect round 2 was fixing. The audit killed them by mutation: dropping `failed_fetch` from the exit code (a stale inventory would exit 0 — the round-1 bug this docstring exists to prevent), deleting the stale stderr note, removing the fetch value from the row, removing it from the header (silent column misalignment), and narrowing the stale set to `{"FAILED"}` so `--no-fetch` staleness went unsaid. All five now die, plus both new behaviours above. The pair of exit-code tests now use a **real row** rather than `REPOS=[]` — an empty list cannot exercise anything that depends on a row's contents, which is how the `--no-fetch`-exits-0 half went unpinned while looking covered.

## 🟢 The three nits

- **🟢4** — three docstring sentences held only for a subset of invocations: the `fetch` column/key is absent on rows that returned early, the stale note is suppressed under `--json`, and "as loudly as FAILED" is contradicted three lines later by rc 1 vs rc 0. All three now say what is true, including why the rc difference is deliberate.
- **🟢5** — the tier-A ceiling's rationale block never recorded THIS PR's lowering (7,639 → 7,617) while the sibling file's did, leaving the **raise** as the newest entry and naming a skill this PR renames out of existence. A maintainer reading it as the documented value would "restore" 7,639 and bank 22 chars of the regrowth the ratchet refuses.
- **🟢6** — "309 pass across the six affected gates" was not reproducible and never named the six. Named and re-measured below. Also correcting round 2's "each by its own test": **8 mutants were killed by 6 distinct tests, not 8** — three of the None-guard mutants share one.

## Verification (measured)

45 tests in `test_civitai_app_fleet.py` · the seven mutants above each killed, controls green either side, tree byte-identical after · live run against a purpose-built clone shows a failed fetch yielding the **full row** with `FAILED` and exit 1.

**316 passed** across the six named gates: `test_civitai_app_fleet.py`, `test_skill_tiers.py`, `test_skill_descriptions.py`, `test_skill_audit.py`, `test_doc_path_rot.py`, `test_no_client_hostnames.py`.

## Ledger

`round 3 · payload lines changed THIS round: 94 (fleet.py, +67/−27) · scaffolding: 113 (test_civitai_app_fleet.py 97, test_skill_tiers.py 16) · since round 1: 824`

Measured over `f769372c` alone. The script's own ledger renders `d6815fd1..HEAD`, which spans **two** commits (round 2's fix and round 3's), so its 433-line figure is not this round's payload — the 824 "since round 1" is the script's cross-commit sum and is carried forward unchanged.

Non-zero payload, so the two-consecutive-zero-payload gate does not fire.

```audit-claims round=3 audited=db08c7f00925dad574ff37abfea94f42e71fe124..f769372c
1. The "one unreachable remote collapses all seven rows" blast-radius claim is corrected to one row, with the seven-repo/one-broken-remote measurement recorded beside it.
2. Only a row with NO readable columns — a missing checkout or a non-repository — is replaced by the `!!` line; a row with one unreadable column prints `?` in place, keeps its other columns, and still sets `error` and a non-zero exit.
3. The platform loop no longer skips a row on `error`; it skips only on an unreadable VERSION.
4. vitest_projects reports `no-vite` for a vite config that is ABSENT on the ref, rather than UNREADABLE.
5. The five round-2 behaviours are pinned by mutation: failed_fetch in the exit code, the stale stderr note, the fetch value in the row, the fetch column in the header, and `skipped` in the stale set.
6. The two exit-code tests drive a REAL row instead of REPOS=[], so they exercise the row-dependent half they previously could not.
7. Three docstring sentences are narrowed to what is true: the fetch column/key is absent on early-return rows, the stale note is suppressed under --json, and the rc 1 (FAILED) vs rc 0 (skipped) difference is stated and justified.
8. test_skill_tiers.py's rationale block records this PR's lowering 7,639 -> 7,617 as the newest entry and stops naming the skill this PR renames away.
9. The "309 pass" figure is replaced by 316, with the six gate files named; round 2's "8 mutants each killed by its own test" is corrected to 8 mutants killed by 6 distinct tests.
```

ZacxDev:
**Round 4 delta audit — findings and fixes.** Audited `db08c7f0..f769372c`; fixes in `b8cac6a8`. Round 4 was **not clean**, so the ladder continues.

Seven of the nine round-3 claims came back **actually fixed**, and the auditor watched the two behavioural ones red at `db08c7f0` rather than accepting the commit message. Claim 3 was behaviourally right but unpinned. Claim 9's "8 mutants / 6 distinct tests" recount it could **not** verify — that is a claim about round 2's sweep, whose mutant set cannot be reconstructed — and it said so instead of reporting it covered.

## The theme: round 3 traded one class of fabrication for another

Round 3's fix was to stop replacing a partly-readable row with `!!`. That was right. But printing a row it used to suppress **exposed two cells that assert things no measurement supports** — in the file whose entire premise is that a column which could not be read must never look like one that was.

**🟡2 — an unreadable version printed a fabricated lockstep-break marker.** `!` is documented one line above its own cell as *"lockstep broken — loud"*, and it was emitted whenever the two version strings **differed**. `?` differs from every real version, so a row whose manifest version could not be read rendered `?/0.8.8!` — an assertion that two versions disagree, from a comparison that never happened. Both readings still print; only the assertion is suppressed, and only when one side is `?`.

**🟡3 — `not-consulted` acquired a second meaning.** The printer's `!!` discriminator (`"default_branch" not in r`) and the platform loop's skip (`manifest_version == UNREADABLE`) were both `"error" in row` before this delta and are now *different predicates* — so a row can reach the table **and** be skipped by the loop, printing `not-consulted` on a run where the platform **was** consulted. That is the exact double meaning the cell's own comment forbids for `-`.

Not fixed by dropping the guard: `resolve(parsed, app, "?")` answers `none/-`, a real-looking platform state for a version nobody read, which is worse. The skip gets its own word — three cases, three strings: `not-consulted` (`--no-platform`), `unread` (the CLI failed for every row), `version-unread` (consulted, this row had no version to ask about). The submit floor is keyed on the app slug alone, so it is answerable even when the version is not, and is now read *before* the version guard rather than skipped with it — reporting it unavailable would be the same fabrication in the other direction.

**🟡1 — `no-vite` overflowed the `proj` column and shifted every column after it.** The cell was `{...:>4}` and round 3's own new sentinel is 7 characters; Python's width specifier pads but never truncates. Measured before the fix: `guard` at column 112 in the header and in a `2+` row, at **115** in a `no-vite` row. This is the "silent column misalignment" defect that same commit wrote a test to prevent, on a column it introduced a value for.

The `ver` cell has the same shape and is **deliberately not changed**: a two-version lockstep-break string is unbounded, so no width fixes it and truncating would turn a read version into a fabricated one. Written down beside the cell; the new alignment guard is scoped to the columns after `proj`. It reproduces at `f769372c` too and is not introduced by this delta.

## 🟡4 — the recurring shape, fourth consecutive round

**The entire platform-enrichment branch was executed by zero tests** — including round 3's own new guard. Three controls, all reproduced: reverting to `if "error" in row:`, replacing it with `if False:`, and planting a bare `raise` inside the loop body each left the suite at **45 passed**. The `else:` branch never ran anywhere, because the only test setting `CIVITAI_STATUS_FILE` also set `REPOS = []` — an empty fleet cannot enter a per-row loop.

Four of the six new tests now drive `main()` with a captured status dump **and** real rows. Two are invariant guards, labelled as such in the file and **not counted as regression coverage** for this delta: they are green at `f769372c` because round 3's behaviour is correct there — what was missing was anything that ran it.

## Verification

**322 passed** across the six gates (316 at `f769372c`), `test_civitai_app_fleet.py` 45 → 51. Neither byte ceiling touched; both still green at 0 headroom.

Red/green against `f769372c`'s `fleet.py`: **4 red** (proj alignment, the `!` marker, `version-unread`, the floor on a version-unread row), 2 green (the invariant guards above).

**10 isolated mutants, each killed by its own guard's assertion**, swept under `PYTHONDONTWRITEBYTECODE=1` in a `cp -a` copy with its `.git` removed. Two were re-verified independently by the operator rather than taken from the agent's report: the planted-`raise` reachability control (survived at `f769372c` with 45 passed; now fails **4** tests with `PLATFORM ENRICHMENT LOOP REACHED`), and spelling `version-unread` back to `not-consulted` (killed by `test_the_three_platform_answers_are_three_distinct_strings` on `assert 'version-unread' in ...` — its own assertion, not a bystander's).

**Not pinned, stated rather than implied:** the fake `inspect` these tests install means the `inspect`→`main` seam is still unexercised. They pin what `main()` does with a row, not that `inspect()` produces one of that shape.

## The red gate is `main`'s, not this PR's — measured two ways

`tekton/devrc-pytests` is red on this PR. It is **not** this PR's:

| tree | collected | failed |
|---|---|---|
| pristine `origin/main` `4f49f5dc` | 21,127 | 2 |
| `origin/main` + `f769372c` (clean merge) | 21,152 | 2 |

**Identical failing test set in both**, and neither failure belongs to #1393:

1. `test_run_tests_targets.py::test_a_pinned_skip_whose_TARGET_did_not_run_does_not_count` — a **load flake**. It shells out to a nested `run-tests.sh` its own docstring costs at ~13s, and hit the 120s ceiling (`returncode: -9`) because two full 21k-test suites were running concurrently on the box. Alone, with the box quieter: **1 passed in 45.48s**.
2. `test_no_client_hostnames.py::test_no_client_subdomain_literal_is_committed` — real, and **`main`'s**. `claudedocs/handoff-civitai-app-fleet.md` spells six client subdomains into this public repo; landed by #1402, fixed in #1405.

Separately: this branch is **18 commits behind `origin/main`** and carries stale copies of three escrow/opencode test files that fail on it and pass on `main`. That is what the PR-branch gate is actually reporting. The merged tree is the tree that matters, and it is clean apart from the two above.

## Ledger

`round 4 · payload lines changed THIS round: 54 (fleet.py +47/−7) · scaffolding: 263 (test_civitai_app_fleet.py +256/−7) · since round 1: 1,011 over 7 files (git diff --numstat eab2ceb6..b8cac6a8)`

Non-zero payload, so the two-consecutive-zero-payload gate does not fire.

```audit-claims round=4 audited=f769372c804425b05eec0881022f4978f4886241..b8cac6a8
1. The proj column is 7 wide in BOTH the header and the row, so a `no-vite` value no longer shifts guard/fetch/platform three columns right.
2. The `ver` cell's same-class overflow is deliberately NOT fixed, is documented beside the cell, and the new alignment guard is scoped to the columns after proj rather than silently widened.
3. The `!` lockstep-break marker is suppressed when either version reading is UNREADABLE; both readings still print.
4. The platform loop's row-level skip writes its own string `version-unread`, distinct from `not-consulted` (--no-platform) and `unread` (the CLI failed for every row).
5. submit_floor is read BEFORE the version guard, so a version-unread row still reports the floor it is able to answer.
6. Four new tests drive main() with a real status dump AND real rows, so the platform-enrichment loop body executes; the planted-raise mutant that survived at f769372c now fails four tests.
7. Two of the six new tests are invariant guards, labelled as such in the file, and are NOT counted as regression coverage for this delta.
8. The inspect->main seam remains unpinned: the new tests install a fake inspect, so they pin what main() does with a row, not that inspect() produces one of that shape.
```

ZacxDev:
**Round 5 delta audit — findings and fixes.** Audited `f769372c..b8cac6a8`; fixes in `6cbaca03`. Round 5 was **not clean**, so the ladder continues.

Round 5's advisory verdict was *safe to merge*. The stop rule is about **findings**, not the verdict — and the round found a hole that a fully green suite would have shipped.

## 🟡1 — the third sentinel that nothing ran

Round 4's own thesis was *"three cases, three strings"*, and the test named `test_the_three_platform_answers_are_three_distinct_strings` pinned **two**. The `unread` arm — the `except Exception` branch, where the CLI failed for every row — was entered by no test in the file.

Three isolated mutants each **survived a green 51**: rewriting `unread` to `not-consulted`, to `version-unread`, and rewriting the `submit_floor` sentinel to `none`. The first is the one that matters — it makes a failed-CLI run **indistinguishable from a `--no-platform` run**, which is precisely the double meaning round 4 existed to remove, on the one arm its guard could not reach.

Fixed by driving the arm for real: a `CIVITAI_STATUS_FILE` pointing at a nonexistent file raises `FileNotFoundError`, which is not `UnknownState`, so it lands exactly where an absent or unauthed CLI lands. Both keys are asserted through `--json`, and the test ends in a case→answer ledger.

Also added `_platform_cells()` — the old substring checks were **walkable, because `version-unread` contains `unread`**.

## 🟡2 — the admission that was only ever in a commit message

Round 4 claimed the `inspect`→`main` seam was unpinned and said so "plainly". It said so **only in the commit message**, which is not what a maintainer reads. Worse, the nearest text in the tree pointed the other way: `_fleet_rows`'s docstring said *"so `REPOS` and the fake `inspect` cannot drift apart"* — fixture-internal consistency, reading as reassurance about the open seam.

The hazard was live and measured: renaming `inspect()`'s `manifest_version` key left the suite at **51 passed**, while a real run died `KeyError: 'manifest_version'` at `fleet.py:366` on the first row. The tool dies on every invocation, suite fully green.

Closed with **both halves**, because neither alone is sufficient:
- a **key ledger** that AST-walks the module, skips only the `inspect()` subtree, and pins the produced and consumed sets as literals plus an orphan check — so the assertion fails when the set **grows** *or* **shrinks**;
- a **behavioural case** driving the real `inspect()` through the `_RUN` seam, pinning the rendered row as a whole normalised line.

The split earns itself: a mutant that reads `package_version` from the wrong manifest field is **invisible to the ledger** (the key set is unchanged) and is caught only by the behavioural guard. Its mirror — an added consumed key — is caught only by the ledger.

Independently re-verified by the operator, not taken from the agent's report: the `manifest_version` rename now fails **two** tests, the ledger on `AssertionError: ['manifest_ver', 'manifest_version']` and the behavioural guard on `KeyError: 'manifest_version'`. Each on its own error.

The seam guard's limits are stated **in the file**: names only, and only rows reached via `r`/`row`.

## 🟢3 / 🟢4

- **`branch` is a second unbounded cell** the round-4 width comment did not name — it framed `ver` as *the* exception while `default_branch` comes from `origin/HEAD` and is equally untruncated (`platform` moves from column 142 to 146 with a `development` branch). Zero blast radius today, but **a comment is a claim**: it now names `branch` and names the three cells that genuinely are bounded. No column widened and no existing guard's scope changed; a separate new guard pins that a long branch value prints **in full**.
- **The floor reorder gave never-inspected rows two `--json` keys they never had.** `submit_floor` stays — it is slug-keyed and genuinely answerable. But `platform: "version-unread"` had come to span two structurally different cases, so the never-inspected case gets its own sentinel `not-inspected` rather than a docstring saying one string means two things — which would contradict the principle three lines above it. The duplicated "was this row inspected?" predicate is consolidated into `_inspected()`, used by both the loop and the printer; **those two disagreeing is what produced round 4's finding 3.**

## Verification

**325 passed** across the six gates (322 at `b8cac6a8`), `test_civitai_app_fleet.py` 51 → 54. Neither byte ceiling touched; both green at 0 headroom.

**12 isolated mutants, all killed**, swept under `PYTHONDONTWRITEBYTECODE=1` in a `cp -a` copy with `.git` removed — including the three that survived at `b8cac6a8`, now dying on the new arm's own equalities. Plus two controls: a positive control reddening round 4's alignment guard (the harness can go red), and an orphan-assertion reachability check confirming that assertion is not shadowed by the two before it.

**Honest coverage accounting:** three of the four new guards are **invariant guards**, labelled as such in the file. Only the `not-inspected` arm is regression coverage — it is red at `b8cac6a8` on `assert 'version-unread' == 'not-inspected'`.

Round 4's "each by its own guard's assertion" is corrected: three of those ten mutants kill **two** tests each.

**Not verified, stated rather than implied:** nothing here ran `civitai app status` against the real platform. Every platform path goes through `CIVITAI_STATUS_FILE`, so the `except Exception` arm is exercised by a **missing file**, not by an absent or unauthed Go CLI — the same code path, not the same observation.

## Ledger

`round 5 · payload lines changed THIS round: 52 (fleet.py +46/−6) · scaffolding: 313 (test_civitai_app_fleet.py +300/−13) · since round 1: 1,338 over 7 files (git diff --numstat eab2ceb6..6cbaca03)`

Non-zero payload, so the two-consecutive-zero-payload gate does not fire.

```audit-claims round=5 audited=b8cac6a895cad5acb5aa290ab07d6460a0bb0d1a..6cbaca03
1. The `unread` arm of the platform trichotomy is driven by a test — a nonexistent CIVITAI_STATUS_FILE raises FileNotFoundError, which is not UnknownState — asserting both `platform` and `submit_floor` exactly through --json.
2. The renamed four-answers test covers every arm and ends in an explicit case->answer ledger, so its body is as wide as its name.
3. `_platform_cells()` replaces substring checks that were walkable because `version-unread` contains `unread`.
4. The inspect->main seam is pinned by a key ledger that AST-walks the module, skips the inspect() subtree, pins produced and consumed key sets as literals, and fails when the set GROWS or SHRINKS.
5. A behavioural case drives the real inspect() through the _RUN seam and pins the rendered row as a whole normalised line, catching wrong-VALUE mutants the ledger cannot see.
6. `_fleet_rows`'s docstring states what the fixture does and does NOT cover, and the seam guard's limits (names only; only rows reached via r/row) are stated in the file rather than only in a commit message.
7. The width comment names `branch` as a second unbounded cell and names the three that are bounded; no column was widened and no existing guard's scope changed.
8. A never-inspected row gets its own sentinel `not-inspected` instead of sharing `version-unread`; submit_floor stays because it is slug-keyed real data.
9. The duplicated "was this row inspected?" predicate is consolidated into `_inspected()`, used by both the platform loop and the printer.
10. Three of the four new guards are invariant guards, labelled as such in the file; only the `not-inspected` arm is counted as regression coverage.
11. Round 4's "each by its own guard's assertion" is corrected: three of those ten mutants kill two tests each.
```

PR #1404, opened 2026-09-08
feat(bar): runaway-process pill — block, poller source, fzf TUI and two-way pinned tests

ZacxDev:
## What this is

A runaway-process bar pill: a `custom` i3status-rust block, a poller source, an fzf
click-through TUI, and tests. **411 insertions.**

🔴 **This is not my work and it has not been reviewed by its author.** It was found as
*uncommitted work in the laptop's tree*, where it was blocking `ship.sh` at rc7. The
previous session preserved it onto `zach/i3-runaways-bar-block` (`90ab4b61`) rather than
discard it. This branch is that commit with current `main` merged in.

Base `d8e2bbe8`. The merge is clean (no conflicts).

| tier | result |
|---|---|
| dev-host `gate.sh --tier both` | `GATE: RESULT=PASS exit=0` — pytest `collected=21117 passed=21115 skipped=2 failed=0` (floor 20342), node `tests=1449 pass=1449 fail=0` (floor 1367) |
| nix sandbox `checks.x86_64-linux.pytests` | `RESULT: PASS (exit=0)` — `failed=0` |
| nix sandbox `checks.x86_64-linux.nodetests` | `RESULT: PASS (exit=0)` — `fail=0` |

The two nix derivations were built **one at a time**, never combined. All verdicts are read
from the runners' own `RESULT:` lines in `nix log`, not from an exit code — both derivations
actually built rather than returning a cached 0. No `panic: test timed out` in either.

**Delta against `main`: zero.** The handoff this came from set the bar at "delta against
`main`'s known 7 failures", but those 7 were already fixed by #1392 (`94f82796`), which is in
this branch's merge base — so the comparison is against a green `main`, not a 7-failure one.

The base moved 4 commits after the original merge was tested, so that earlier result described
a tree that no longer existed; this is a fresh run on the current base.

## Test coverage

The block is pinned **two-way** into every ledger in `test_bar_status.py` — `BLOCK_SOURCE_FILES`
vs `ALL_BLOCKS` (`assert ==`), `ALARM_CACHES` (`assert sorted(...) == sorted(...)`), the
`SOURCES` ledger, `UNMEASURED_PILL`, `MEASURED_ALL_CLEAR_PILL` — plus the measured counts
`_EXPECTED_SCRIPT_BLOCKS` 12→13 and `_EXPECTED_BLOCK_DEFS` 21→22. A half-added block fails the
suite rather than passing silently.

## 🔴 Known defects — found in review, NOT fixed here

Deliberately left alone so the gate result above describes exactly the preserved commit. Each
should be a follow-up.

1. **`RUNAWAY_IGNORE` is inert.** It is built from the environment and its comment states the
   thresholds are "env-tunable so a host can retune without code changes" — but
   `parse_runaways` never reads it. The exclusion is hardcoded to
   `comm in ("ps", "python3", "bar-status-poll")`. A declared knob that nothing branches on:
   no test can catch this, because there is no code path to fail.
2. **That hardcoded exclusion is over-broad.** It drops *any* runaway whose `comm` is
   `python3` — on this host (agents, collectors) a plausible real runaway, silently unreported.
3. **Two copy-paste leftovers.** The `runaways` toast `action` opens Grafana, inherited from
   the `telemetry` spec it was copied from; and the right-click `cmd` uses a literal
   `~/workspace/devrc/scripts/syshealth` where every other working-tree reference in
   `graphical.nix` interpolates `${home}` (lines 261, 271, 350). A literal `~` is only expanded
   if the click is spawned through a shell.

## Deploy note

Merging is not deploying, and shipping is not rendering: a **new** bar block needs
`i3-msg restart` — a `home-manager switch` alone leaves the host reporting complete success
with no pill (measured previously in `handoff-load-monitor-reap`). The block is workbench-only
(`!isLaptop`), matching its `home.file` gates.


ZacxDev:
## 🔴 Correction to this PR's own description, plus audit round 1 — verdict: NEEDS REWORK

**Correcting my own PR body first, publicly, because a reviewer may already have read it.**

The body's "Test coverage" section says the block "is pinned **two-way** into every ledger" and
that "a half-added block fails the suite rather than passing silently". Both sentences are
literally true and both are **misleading about what is actually covered.** Measured this round:

```
parse_runaways replaced wholesale with a constant {"count": 0, "state": "Idle", "processes": []}
  -> 563 passed          MUTANT SURVIVED
positive control: i3status-runaways `count <= 0` -> `count < 0`
  -> 4 failed, 559 passed   (so the harness CAN go red on this file)
baseline: 563 passed        (run under PYTHONDONTWRITEBYTECODE=1)
```

The suite constrains the **renderer** and nothing else. `parse_runaways`, `_scan_runaway_procs`,
`fetch_runaways`, the `--mock-runaways` flag and all 146 lines of `runaway-menu` are unconstrained.
The ledger insertions are real work, but every one feeds a **hand-written cache payload** to the
block script — none reaches the code that produces those payloads. My description implied coverage
this change does not have.

Both gate tiers were green over every finding below.

## 🔴 Deploy-blocking

1. **The left-click is a silent no-op.** `nix/graphical.nix:426` launches the fzf TUI with no
   controlling terminal. The live `i3status-rs` (PID 2859788) has TTY `?`. Driving it:
   `setsid runaway-menu </dev/null` → rc 0, no stdout, no stderr, nothing happens; bare `fzf` with
   no ctty → rc 2 `inappropriate ioctl for device`. Every other TUI in this file is wrapped in
   `alacritty --class float,float -e`, including this PR's own right-click; the bare left-clicks
   are all rofi, which needs no tty.
2. **The kill confirmation can never return yes.** `scripts/runaway-menu:130-136` runs fzf with no
   `input=`, so stdin is inherited and fzf falls back to `$FZF_DEFAULT_COMMAND` (`fd …`, set in
   `nix/sessionVariables.nix:4,9`) and emits the *selected line*, never typed text — `--print-query`
   is absent. Measured under a pty, typing `y`↵: stdout `'.nix-profile/lib/lua/5.3/yaml.so'` →
   `in ("y","yes")` is False. The prompt also renders a recursive listing of `$HOME`.
3. **Every `python3` process is structurally unflaggable.** `scripts/bar-status-poll:1340` excludes
   the whole runtime; the comment claims self/ps exclusion. `comm=python3, pcpu=100, etimes=7200` →
   **count=0**; `node` control → count=1. **18 python3 processes are live on this box.** The
   collector, dl-router, the poller itself and every agent script are Python — the most likely
   runaway class here is the one class it cannot see. `syshealth:246` does this correctly via
   `os.getpid()`, not a comm name.

## 🟡 Should-fix

4. **`RUNAWAY_IGNORE` is inert** (`bar-status-poll:1310`) — zero consumers, verified by exhaustive
   `find … | xargs grep` (not `grep -r`, which is ugrep here and honours `.gitignore`).
   `comm=annobin` is flagged despite `anno` being in the default list. ⚠ Not a one-line fix: it is a
   `frozenset` (exact membership) while both implementations it says it copies — `syshealth:83` and
   `cpu-monitor.sh:78-90` — do case-insensitive **substring** matching, under which `annobin` still
   would not match `anno`. The semantics need deciding.
5. **This is a third implementation, and `cpu-monitor.sh` is LIVE right now**
   (`systemctl --user is-active cpu-monitor` → active, MainPID 1470) alerting on the same event with
   sampled CPU, a 6-sample sustain, a working ignore list, a cooldown, a **daily cap** and a recovery
   toast — none of which this has. `cpu-monitor.sh:43-54` records that its cap exists *because* an
   uncapped toast path was a measured problem. `cc1plus` at 100% for 20 min (an ordinary `nix build`)
   satisfies both and produces two toasts.
6. **Zombies are flagged and cannot be cleared.** `syshealth:248` excludes them; `parse_runaways` does
   not. A zombie at `pcpu=95, etimes=3600` is flagged, `kill_process` SIGTERMs it forever, the pill
   stays red. 5 zombies are live now (all at 0% today, so latent). Sub-issue at `:1374`: `ps -o comm=`
   emits `<name> <defunct>` **with a space**, so `split(None, 8)` misaligns `args` for exactly these.
7. **No test constrains the payload** — see the correction above.
8. **Level latch on a source that can sit permanently non-zero** (`:860`). This is the defect the repo
   already paid for and fixed once — `_toast_specs()`'s own comment at `:835-841` documents that a
   permanently-above-zero source leaves the latch always set, *"the one event worth interrupting for
   was the one that could not interrupt"*, which is why `clawgate_stuck` exists. One persistent entry
   (the zombie from #6) pins count at 1 and every later runaway is a silent 1→2.

## 🟢 Nits

9. Toast action opens Grafana (`:860`), violating `_toast_specs()`'s stated invariant at `:821-824`
   that the action opens the same target as the left-click. Copy-paste from `telemetry`.
10. `runaway-menu:41-46` renders `proc.get("cwd")`, which `parse_runaways` never emits.
11. `bar-status-poll:1428`'s comment enumerates cache contents; `runaways.json` now also persists the
    full argv of every flagged process. Not a leak — verified `-rw-------` — but the enumeration is stale.

## Re-verified, and correct

`_EXPECTED_BLOCK_DEFS = 22` and `_EXPECTED_SCRIPT_BLOCKS = 13` both measure **exactly** right, not
merely `>=`. Signal 19 is free and `SIGNALS` agrees. Block and script `!isLaptop` gates agree.
The renderer half is good and worth keeping as-is: the `?`-carries-the-count unmeasured pill,
hide-at-zero, and the `bar_freshness` delegation are all correct and consistent with sibling blocks.

## Could not verify

The pill has **never been seen on a live bar** — the workbench has no `runaways.json`, as the running
poller predates this change. All findings come from driving the functions directly. The with-tty half
of #1 was not driven inside an `alacritty` wrapper, because #1 fires first.

## Verdict

**Needs rework** — not merge-after-small-fixes. The click does nothing, the kill it exists to perform
is unreachable even once the click works, the detector is blind to the most likely runaway class on
this host, its one knob is inert, and no test constrains any of it.

🔴 **Before another round, #5 wants an operator decision.** The cheapest coherent version of this is
probably a pill that *renders* what `cpu-monitor` or `syshealth --json` already decides, rather than a
third detector — which dissolves #3, #4, #5 and #6 together. Worth settling before paying for tests on
a predicate that may not survive.

<sub>Audited at head `1d74f343`, merge-base `d8e2bbe8`. Note `origin/main` has moved three times since
the green gate reported in the body, so that result is now stale.</sub>

ZacxDev:
## Reworked to render-only — 8 of 11 findings addressed by DELETING the code that carried them

Round 1 returned **needs rework**. Rather than fix a private predicate, this removes it: the pill
now renders what `scripts/syshealth` already decides. **Net −240 payload lines.**

| finding | how |
|---|---|
| 3 python3 blindness · 4 inert `RUNAWAY_IGNORE` · 6 zombies flagged and unkillable | **by construction** — syshealth excludes self by RESOLVED PID (never a comm name), excludes zombies from its eligible set, and honours a working case-insensitive ignore list |
| 1 dead left-click · 2 unreachable kill confirm · 10 phantom `cwd` column | **by removal** — `scripts/runaway-menu` deleted (146 lines). Both clicks now open syshealth in a float terminal, which is what the working right-click already did |
| 8 level latch | the toast gates on `new_count` (pids absent from the previous poll), the `clawgate_stuck` pattern. **One** spec, not two — a second latch on `count` would double-toast the first runaway |
| 9 Grafana action | now opens what the left-click opens, as `_toast_specs`'s docstring requires |

Also: a broken detector is now an `error` payload rather than a zero, so `bar_freshness.is_marker`
renders the visible `?` instead of a confident empty block. A detector that cannot run and one that
ran and found nothing must not be indistinguishable.

### Finding 7 — the measurement, before and after

Finding 7 was that nothing constrained the payload. Same mutation (`parse_runaways` replaced
wholesale with a constant), same harness:

| | result |
|---|---|
| **before** (round 1) | `563 passed` — **MUTANT SURVIVED** |
| **after** (this commit) | `4 failed` — **KILLED** |

Second mutant, `count_key` reverted to `"count"`: killed by **its own** assertion
(`assert spec["count_key"] == "new_count"`), not a neighbour's. Both swept under
`PYTHONDONTWRITEBYTECODE=1` in a `cp -a` copy with `.git` removed first (a worktree's `.git` is a
FILE — a commit in the copy would land on this branch). Harness positive control: the suite went red
on a wrong assertion of mine during development, so it demonstrably reaches this file.

### 🔴 A trap found while wiring it, now pinned by a test

**`syshealth`'s exit code is not the signal and is deliberately never checked.** It folds every
section into one status, so it returns **1** whenever anything warns — load, zombies, cpu hogs —
while `runaways` is empty. Measured on this host: `runaways: []` at rc 1. Gating on the status would
fail every such poll and pin the unmeasured `?` pill permanently, *while looking exactly like a
working freshness guard*. `test_fetch_runaways_IGNORES_syshealths_EXIT_CODE` pins it, with a control
proving the same rc plus unreadable stdout still IS a failure.

### Gates — and the dev-host tier flaked three ways, so read this before the numbers

Head `e2b338f2`, merge base **`e8392c57`**.

| tier | result |
|---|---|
| **sandbox** `checks.x86_64-linux.pytests` | `collected=21166 passed=21164 skipped=2 failed=0` · `RESULT: PASS (exit=0)` |
| **sandbox** `checks.x86_64-linux.nodetests` | `tests=1449 pass=1449 fail=0` · `RESULT: PASS (exit=0)` |
| dev-host `gate.sh --tier both` | **FAIL both runs — see below** |

🔴 **Two runs of the IDENTICAL tree produced THREE distinct failures with ZERO repeats:**

| run | failed |
|---|---|
| 1 | `test_subsystem_store_api.py` — `OSError: [Errno 98] Address already in use` · `test_nebula_relay_apply.py::test_no_predictable_tmp_path_is_live_while_the_verifier_runs` |
| 2 | a tmux `open_window` PATH assertion — *"the launch environment REPLACED this unit's PATH instead of prepending to it"* |

None is in, or reachable from, the four files this PR touches. All three are resource-contention
shapes — a bound port, a predictable `/tmp` name, a replaced PATH — and sibling agents are live on
this box (`gitenv` reported 8 processes in protected repos, "none of them is ours"). A defect
reproduces; these did not.

Controls run rather than assumed: the nebula test **alone** at `origin/main` passed; both failing
files together under the suite's own `-n 4 --dist loadfile` at `origin/main` passed (774). Neither
reproduces the full gate's pressure, which is why the **sandbox tier is the arbiter here** — it is
hermetic, it is what Tekton gates on, and it collected the **same 21166 tests** and passed every one.

I am reporting this as *the dev-host tier being unreliable on this host right now*, not as a green
gate. Both statements are in front of you; the tier disagreement is real and is already tracked as an
open investigation in the handoff — these are three more instances of it, none involving
`browser-bridge`, which suggests that item is broader than the single test it names.

⚠ **`main` moved 7 commits during the gate run**, so the numbers above describe base `e8392c57`, not
current `main`. Re-gate the merged tree immediately before merging — `strict` is deliberately off on
this repo, so a green check is a claim about a branch, never about the tree a merge creates.

### One deliberate omission

syshealth offers `cwd`, `critical` and `user` per row and the payload drops them. Emitting a field
nothing reads is the same defect as the `RUNAWAY_IGNORE` knob this commit deletes — so every field
that survives has a consumer: `pid` feeds `new_count`, `cpu` and `cmd` feed the toast body.

### Not addressed

**Finding 5 stands** and is not a code question: `cpu-monitor.sh` is live (`MainPID 1470`) and still
alerts on this same event with sampled CPU, a 6-sample sustain, a cooldown and a daily cap. This
change removes the *third* predicate, leaving two. Whether the pill should instead render
`cpu-monitor`'s verdict is an operator call — and it would need new plumbing, because `cpu-monitor.sh`
persists no state a pill could read (measured).

ZacxDev:
Round 1 fixes are pushed as `e2b338f2`. Claims block for the next round:

```audit-claims round=1 audited=1d74f343..e2b338f2
1. Finding 3 (every python3 process unflaggable): claimed addressed by deleting the local predicate entirely; the set now comes from `syshealth --json`.
2. Finding 4 (RUNAWAY_IGNORE inert): claimed addressed by deleting the constant; the ignore list is now syshealth's `--ignore`.
3. Finding 6 (zombies flagged, unkillable): claimed addressed by delegation; syshealth excludes zombies from its eligible set.
4. Finding 1 (left-click a silent no-op) and 2 (kill confirm can never return yes): claimed addressed by DELETING scripts/runaway-menu; both clicks now run syshealthCmd, a float-terminal wrapper.
5. Finding 10 (runaway-menu rendered a `cwd` nothing emits): claimed addressed by the same deletion.
6. Finding 8 (level latch on a permanently non-zero source): claimed addressed by gating the toast on `new_count` via `count_key`, with exactly one spec rather than two.
7. Finding 9 (toast action opened Grafana): claimed addressed by pointing the action at syshealth, matching the block's left-click.
8. Finding 7 (no test constrained the payload): claimed addressed by 5 new tests; claimed measured as mutant SURVIVED 563-passed before vs KILLED 4-failed after, plus a second mutant on count_key killed by its own assertion.
9. Claimed: a broken detector now yields an `error` payload rather than count 0, so the pill renders `?` instead of a confident empty block.
10. Claimed: syshealth's exit code is never consulted, pinned by test_fetch_runaways_IGNORES_syshealths_EXIT_CODE.
11. Claimed NOT addressed: finding 5 (cpu-monitor.sh is live and alerts on the same event). Left open as an operator decision.
```

ZacxDev:
Round 2's fixes landed as `339e078d` (then merged to main via `c604ddc0` / squash `53c9f664`). Posting the block that was missing, so a round 3 anchors on the right range instead of silently widening to span both fix rounds.

```audit-claims round=2 audited=e2b338f2..339e078d
1. Finding 1 (the toast's Open was a bare `alacritty -e syshealth`, which closes when syshealth exits ~0.16s): claimed addressed by emitting the same `bash -c '...; read -n 1'` hold the click uses.
2. Finding 1's prose half (two comments asserting the toast action and the click were "shared"/the SAME string): claimed addressed by stating they are two spellings of one rule that cannot be collapsed, and naming the test that pins them.
3. Finding 2 (the guard asserted only `"syshealth" in action` and never opened graphical.nix): claimed replaced by a test reading BOTH files and asserting terminal + hold + syshealth on each side; claimed measured as mutant D SURVIVED before / KILLED after, and mutant E (real terminal, no hold) KILLED as the reachability control.
4. Finding 3 (`i3status-runaways:13` still described the deleted runaway-menu): claimed corrected to say the click opens syshealth.
5. Finding 4 (`--mock-runaways` help described the pre-delegation contract): claimed corrected to say it takes a `syshealth --json` report dict.
6. Finding 5 (X-Restart-Triggers omitted the new working-tree dependency): claimed addressed by adding `scripts/syshealth` to the list.
7. Finding 7 ("every field the payload emits has a consumer" was false for `cpu` and `cmd`): claimed restated as a deliberate exception rather than a rule without exceptions.
8. Finding 6 (the commit message's "Net: -240 payload lines"): claimed corrected to +122/-238, net -116, in the following commit message.
9. Claimed corrected alongside: "~0.09s" for syshealth --json -> 0.156-0.187s; "18 python3 processes" -> 14 at re-measurement.
10. Claimed NOT addressed: finding 5 of round 1 (cpu-monitor.sh is live and alerts on the same event) — left as an operator decision.
```

⚠ **Item 10 has since been ACTED ON, and it was a shipped regression rather than a preference.** `cpu-monitor.sh` fires `⚠ Runaway process: <comm>` at critical urgency with a 6-sample sustain, a daily cap (`CPU_MON_MAX_ALERTS_PER_DAY`, default 8) and a recovery toast; the pill added a second, uncapped announcer for the same process ~7 minutes later. Fixed on branch `fix/runaways-toast-duplication` (`85c62a70`) by removing the pill's toast spec entirely, along with the `new_count`/`prev_pids` and `_syshealth_action` machinery that existed only to arm it.

ZacxDev:
## Ladder CLOSED at round 2 — by operator decision, with the one unaudited item MEASURED rather than assumed

Round 2 returned two real 🟡 findings and its fix changed payload, so the findings-keyed stop rule
and the attribution gate both say round 3 was indicated. It is not being run. Recording why, and
what was checked instead, so this reads as a decision rather than as a ladder that trailed off.

**What a round 3 could still have seen on this PR, and what it now says.** Round 2 audited
`e2b338f2..339e078d` (claims block posted above). The only change after that tip attributable to
this PR is the merge `c604ddc0`, whose own resolution is two constants. This ladder never ran the
"re-derive every COUNT the PR's files assert, against the CURRENT head" step, so it was run now,
against `origin/main` at `a2b74e4b` — i.e. *after* #1430 landed `fansBlock`'s COOLING view, not on
the tree the merge was resolved on:

| constant | re-derived on current main, with the test's own regex | value in the file |
|---|---|---|
| `_EXPECTED_BLOCK_DEFS` (`_BLOCK_DEF`) | **23** | 23 ✔ |
| `_EXPECTED_SCRIPT_BLOCKS` (`_SCRIPTS_DIR_CMD`) | **14** | 14 ✔ |

`scripts/tests/test_bar_status.py` — **569 passed in 79.8s**. These are `>=` floors, so a green run
alone would not have distinguished "correct" from "one short"; the numbers above are the
measurement, and the semantic-conflict resolution in `c604ddc0` is exact and still exact.

**Why the rest of a round 3 would have audited the ladder rather than the PR.** Everything else in
that range is round 2's own fix: the `bash -c '…; read -n 1'` hold added to `_syshealth_action`,
the two comments rewritten to say the toast action and the click are two spellings of one rule, and
the replacement guard that reads both files. **PR #1435 deletes all of it** — the pill's toast spec
goes entirely, along with `_syshealth_action` and the `new_count`/`prev_pids` machinery that existed
only to arm it (`nix/graphical.nix +10/-12`, `scripts/bar-status-poll +41/-69`,
`scripts/tests/test_bar_status.py +42/-63`). Auditing it here reads code with hours to live.

**Where the outstanding scrutiny went instead — not dropped.** #1435 is open and has never been
audited. Its round 1 is briefed with the round-2 claims that SURVIVE the deletion, so they are
re-checked against the tree that will actually ship: the click path's own hold, the surviving half
of the two comments, the `X-Restart-Triggers` entry for `scripts/syshealth`, the `--mock-runaways`
help text, and the `+122/-238` line count.

⚠ **One claim from the merge is corroborated but NOT confirmed.** This PR was merged over a red
**nix sandbox** pytests tier attributed to `test_git_repo_isolation.py::test_live_cotenants` as a
co-tenant flake, with the confirming re-run still in flight. That file is now **115 passed**,
including `test_live_cotenants` — but on the **dev-host** tier, which is not the tier that was red.
The attribution is still unconfirmed on its own tier.

PR #1409, opened 2026-09-08
fix(escrow): age has more than three refusals — three ordinary corruptions were reaching "CANNOT SAY WHY"

ZacxDev:
Rank 13 of the `nix-disk-cleanup` handoff asked for an adversarial audit of the
merged `#1392` — the change that governs whether a corrupt backup means "rotate
your escrow key", which merged without one. This is the audit's one finding,
fixed.

## The audit first confirmed `#1392`

Re-measured independently rather than read off its table: 84 decrypt runs, 7
payload sizes × 6 fault shapes × age v1.3.1 (login shell) and v1.3.2 (dev shell).
v1.3.2 really does create `--output` lazily, its three re-keyed marker strings
really are byte-identical across both versions, and an 11-mutant battery on the
guards it shipped killed 10 (the 11th, deleting `classify_age_refusal`'s
`.lower()`, is guarded here). The three load-bearing mutants were re-run to name
their killer and each died to the guard that owns it. **The re-key was the right
repair.**

## The finding: a coverage claim measurement refutes

`escrow-verify.py` said of its unclassified verdict: *"DEFENSIVE, NOT OBSERVED …
on both measured age versions every refusal classifies"*, and the new test's
docstring repeated it. Driving really-damaged artifacts through the **real**
verifier, on **both** binaries, three ordinary corruption shapes produced a
refusal the three-entry table did not name:

| damage | age said | verdict it got |
|---|---|---|
| header MAC bit flipped | `bad header MAC` | `DECRYPT-FAILED` — "CANNOT SAY WHY" |
| payload stripped entirely | `failed to read nonce: EOF` | same |
| payload truncated to 8 B | `unexpected EOF` | same |

That message tells the operator age has probably **reworded** and to teach
`classify_age_refusal` the new string — wrong guidance for an artifact that is
simply damaged — and it lists a wrong key among its three open causes. It fails
**safe** (no rotation advice), so this is a coverage defect, not an inversion.
But `ARTIFACT-CORRUPT`'s own sentence and `SECRETS.md` row 33 both claim
TRUNCATED coverage that held only for truncations landing inside a payload chunk.

**Root cause of the miss, recorded because it is reusable:** `#1392`'s sweep
varied payload SIZE across seven values and mangling SHAPE across five, and none
of the five produced a header-MAC-only or a payload-stripped case.

## The repair

age authenticates the whole header, then reads a 16-byte nonce, then the payload
chunks. Truncating *inside* the header and *after* it produce disjoint messages —
measured at 8 offsets spanning the boundary on both versions, every in-header
truncation carries `failed to read header`. So the table is keyed on **where age
got to**, which is what the verdict actually turns on:

- **pre-auth** → `25`: `no identity matched any of the recipients`,
  `failed to read header`, `bad header MAC`, `failed to parse X25519 recipient`
- **post-auth** → `33`: `failed to decrypt and authenticate payload chunk`,
  `failed to read nonce`, `last chunk is empty`, `unexpected EOF`

`AGE_REFUSED_TRUNCATED` joins the closed set. `AGE_REFUSALS_POST_AUTH` and
`AGE_REFUSALS_PRE_AUTH` are published so `escrow-verify.py` branches on a set
owned by the module that classifies rather than on a value it spells itself.

🔴 **Order is now load-bearing, and the rule is total**: every pre-auth marker
precedes every post-auth one, because age v1.3.2 nests the clauses (`failed to
read header: parsing age header: unexpected EOF reading intro`). My first draft
ordered only the EOF family after the header markers — the overlap that had been
*observed* — which is the same "fix the instance, not the class" mistake one
level down; a control run against the other installed binary caught it.

## Controls, reported as pairs

- **RED AT BASE** — this branch's tests against `origin/main`'s two source files:
  **18 failed, 418 passed**. Every new guard names the absence of this fix.
- **GREEN ON BOTH BINARIES** — **436 passed** in the dev shell (age v1.3.2) and
  **436 passed** with v1.3.1 forced ahead on `PATH`. The gate runs 1.3.2; the
  login shell and the operator's `check-escrow.sh` nix-shell run 1.3.1, so one
  green is a claim about one binary.
  ⚠ The first version of that control expanded `$PATH` in the *outer* shell and
  so swapped the **interpreter** too — three unrelated tests went red on a
  missing `minio`. It now asserts `python3` is still the dev shell's before it
  scores anything.
- **GATE, both tiers, dev-host** — `GATE: RESULT=PASS exit=0`.
  pytest `scripts/tests` **13084 passed, 0 failed** (33:49); node
  **suites=5 files=41 tests=1449 pass=1449 fail=0** (floor 1367).
- **MUTATION, 4 new guards + a control** — all four KILLED, each by the guard
  that owns it (`failed to read nonce` marker; `bad header mac` marker; the
  marker ORDER; `_corrupt`'s set membership). The behaviour-free comment reword
  **SURVIVED**, so the scorer is shown able to emit both verdicts. The tree was
  checksummed before the battery and verified byte-identical to the gated state
  after it.
- Every new fixture is **built from the real binary**, never spelled, and
  `test_the_real_binarys_in_header_truncations_are_NEVER_post_auth` walks real
  in-header offsets on whatever age is installed — the live guard on the one
  genuinely weak marker, bare `unexpected EOF`.

## Scope — two audit findings deliberately NOT fixed here

1. **`no identity matched any of the recipients` is not exclusively a wrong-key
   signature.** One flipped bit in the recipient stanza produces it on both
   versions (on v1.3.2, a bit in the wrapped-file-key body does too). Only its
   code comment is corrected — it sits inside the block this change rewrites and
   leaving a known-false sentence was not an option. No operator-facing text
   changes: the verdict already names header damage as equally consistent and
   says do not rotate.
2. **`plain_present is None` (unobservable) is still reported downstream as
   "without leaving plaintext"**, contradicting the probe's own comment that
   `None` must never be read as "age wrote nothing". Narrow reachability
   (requires `Path.exists()` to raise EACCES mid-run).

https://claude.ai/code/session_018EdUuMkYb4BghDqQTbSi9r

### DIFF
--- a/SECRETS.md
+++ b/SECRETS.md
@@ -360,7 +360,7 @@ disaster-recovery key gets rotated, or a tampered backup gets waved through:
 | `29` `AGE-MISSING` | `age` is not on PATH | environment fault; says nothing about the escrow |
 | `30` `ARTIFACT-UNREADABLE` | failed before the key was used | diagnose the object, not the key |
 | `25` `DECRYPT-FAILED` | age refused **before the payload**: wrong key **or** damaged header — **not separable**. ⚠ The same code also carries a second message, *"CANNOT SAY WHY"*, when age refuses in a way this tool cannot read — then **three** causes stay open, corruption included. Read the sentence, not just the number. | try a **different** artifact (`--scope <other>`) with the same escrowed copy: if another opens, the key is fine and this object's header is damaged. **Do not rotate first.** |
-| `33` `ARTIFACT-CORRUPT` | age authenticated the header (**the key worked**) then failed the payload | 🔴 **the backup is TAMPERED/CORRUPT/TRUNCATED.** Check the other retained objects. Do not rotate. |
+| `33` `ARTIFACT-CORRUPT` | age authenticated the header (**the key worked**) then failed past it — a payload chunk that would not authenticate, or an artifact that ran out of bytes | 🔴 **the backup is TAMPERED/CORRUPT/TRUNCATED.** Check the other retained objects. Do not rotate. |
 | `31` `ARTIFACT-EMPTY` | age exited **zero** on an empty payload (**the key worked**) | the artifact holds nothing; do not rotate |
 | `26` `RESTORE-FAILED` | decrypted fine, the git bundle is bad | artifact fault; do not rotate |
 
@@ -383,20 +383,45 @@ which is every artifact this subsystem produces. Left alone, that reported a
 TAMPERED backup as `25` — the row that points at your key.
 
 The distinction **survives**, but on weaker footing: it is now taken primarily
-from age's own refusal message (`failed to decrypt and authenticate payload
-chunk` vs `no identity matched any of the recipients` vs `failed to read
-header`), which is byte-identical on v1.3.1 and v1.3.2. File presence is kept as
-a second signal — still **sufficient**, no longer **necessary**. Both are
-combined in `escrow-verify.py`; the classification itself is
-`restore-verify.py::classify_age_refusal`.
+from age's own refusal message, which is byte-identical on v1.3.1 and v1.3.2.
+File presence is kept as a second signal — still **sufficient**, no longer
+**necessary**. Both are combined in `escrow-verify.py`; the classification itself
+is `restore-verify.py::classify_age_refusal`.
+
+🔴 **THE FIRST VERSION OF THAT TABLE HAD THREE ENTRIES AND MISSED THREE ORDINARY
+FAULTS (corrected 2026-09-09).** An adversarial audit of the change above drove
+really-damaged artifacts through the real verifier on both age versions and found
+that a flipped header-MAC bit, a payload stripped to nothing, and a payload
+truncated to 8 bytes each produce a refusal none of the three entries named — so
+they landed in the *"CANNOT SAY WHY"* message, which told the operator age had
+probably reworded when the artifact was simply damaged. The table is now keyed on
+**where age got to**, which is the thing the verdict actually turns on:
+
+| age said | reached | classification |
+|---|---|---|
+| `no identity matched any of the recipients` | header not opened | pre-auth → `25` |
+| `failed to read header`, `bad header MAC`, `failed to parse X25519 recipient` | header not opened | pre-auth → `25` |
+| `failed to decrypt and authenticate payload chunk` | **past the header** | post-auth → `33` |
+| `failed to read nonce`, `unexpected EOF`, `last chunk is empty` | **past the header** | post-auth → `33` |
+
+The pre-auth/post-auth split is measured, not reasoned: age authenticates the
+whole header before it reads the nonce, so truncating *inside* the header and
+truncating *after* it produce disjoint messages (every in-header truncation
+carries `failed to read header`). A post-auth refusal therefore proves the
+escrowed identity opened the header, which is exactly what `33` claims. Note the
+first line of `classify_age_refusal`'s marker tuple order is load-bearing for
+this: age's in-header EOF messages nest the post-auth clause inside the pre-auth
+one, so the pre-auth markers are matched first.
 
 ⚠ **Say plainly what that costs:** a substring match on another tool's prose is
 weaker than the phase observation it replaced, and this subsystem's own design
 notes argue against exactly that shape. The mitigation is that an
-**unrecognised** message is its own outcome — the verifier then says it *cannot
-say why* and names all three causes, rather than picking one. If you ever see
-that sentence, age has reworded: teach `classify_age_refusal` the new string
-rather than acting on a guess.
+**unrecognised** message is still its own outcome — the verifier then says it
+*cannot say why* and names all three causes, rather than picking one. If you see
+that sentence, age has said something no measured fault produces: teach
+`classify_age_refusal` the new string rather than acting on a guess. ⚠ And do not
+read "no measured fault reaches it" as "nothing can" — that is a claim about a
+sweep, and the sweep above is what the previous one's wording got wrong.
 
 ⚠ It cannot unlock the vault and will not try: every `bw` call runs with stdin on
 `/dev/null`, `--nointeraction`, and a timeout, so an unattended run **fails fast
--- a/SECRETS.md
+++ b/SECRETS.md
@@ -399,10 +399,22 @@ probably reworded when the artifact was simply damaged. The table is now keyed o
 
 | age said | reached | classification |
 |---|---|---|
-| `no identity matched any of the recipients` | header not opened | pre-auth → `25` |
-| `failed to read header`, `bad header MAC`, `failed to parse X25519 recipient` | header not opened | pre-auth → `25` |
+| `no identity matched any of the recipients` | no identity shown to work | pre-auth → `25` |
+| `failed to read header`, `failed to parse X25519 recipient` | no identity shown to work | pre-auth → `25` |
+| `bad header MAC` | **a stanza unwrapped** — the key WORKS — but the header failed its integrity check | key proven → `33` |
 | `failed to decrypt and authenticate payload chunk` | **past the header** | post-auth → `33` |
 | `failed to read nonce`, `unexpected EOF`, `last chunk is empty` | **past the header** | post-auth → `33` |
+
+🔴 **`bad header MAC` is the one to read twice.** It was filed pre-auth in the
+first draft of this table, where row `25` tells the operator *"the escrowed
+identity does not match … if none open, the key is the likely cause"* — a
+rotation-shaped sentence about a key the message vindicates. The discriminating
+control, measured on both versions: decrypt the **same** damaged blob with the
+right identity and with a wrong one — `bad header MAC` vs `no identity matched`.
+age reaches the header's integrity check only after an identity has unwrapped a
+recipient stanza, so that message is positive evidence the escrow **works**. It
+gets `33` with its own sentence, because `33`'s usual wording asserts age
+*authenticated* the header, which is what a MAC failure means it did not.
 
 The pre-auth/post-auth split is measured, not reasoned: age authenticates the
 whole header before it reads the nonce, so truncating *inside* the header and
--- a/SECRETS.md
+++ b/SECRETS.md
@@ -360,7 +360,7 @@ disaster-recovery key gets rotated, or a tampered backup gets waved through:
 | `29` `AGE-MISSING` | `age` is not on PATH | environment fault; says nothing about the escrow |
 | `30` `ARTIFACT-UNREADABLE` | failed before the key was used | diagnose the object, not the key |
 | `25` `DECRYPT-FAILED` | age refused **before the payload**: wrong key **or** damaged header — **not separable**. ⚠ The same code also carries a second message, *"CANNOT SAY WHY"*, when age refuses in a way this tool cannot read — then **three** causes stay open, corruption included. Read the sentence, not just the number. | try a **different** artifact (`--scope <other>`) with the same escrowed copy: if another opens, the key is fine and this object's header is damaged. **Do not rotate first.** |
-| `33` `ARTIFACT-CORRUPT` | age authenticated the header (**the key worked**) then failed past it — a payload chunk that would not authenticate, or an artifact that ran out of bytes | 🔴 **the backup is TAMPERED/CORRUPT/TRUNCATED.** Check the other retained objects. Do not rotate. |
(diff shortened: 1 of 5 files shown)


ZacxDev:
## Round 1 audit — findings, and what round 2 (`deee8a8e`) claims to have done

The round-1 audit read `d72fb128`. It re-derived all six of this PR body's numbers (they held), and attacked the inversion exhaustively on both installed age binaries — every in-header truncation offset × 8 payload sizes, all 168 single-byte header corruptions at two masks, wrong-key, truncated-payload × wrong-key, non-age inputs, and all 5887 armored-artifact truncation offsets, each scored against a wrong-key control. **No inversion found.** It also measured the new marker-ordering rule to be load-bearing on **21 of 168** real v1.3.2 in-header offsets (0/168 on v1.3.1), so it is not theoretical.

It returned three 🟡 and three 🟢. Round 2 addresses all six.

```audit-claims round=1 audited=d72fb128..deee8a8e
1. `bad header MAC` was classified pre-auth; it is now its own published value `AGE_REFUSED_HEADER_MAC`, in neither the pre-auth nor the post-auth set, with `AGE_REFUSALS_KEY_PROVEN` published as the predicate the "do not rotate" advice rests on.
2. The consumer gives `AGE_REFUSED_HEADER_MAC` the `ARTIFACT-CORRUPT` token and exit code with a SECOND message, rather than the pre-auth verdict that offers a non-matching identity as an open cause.
3. `failed to parse x25519 recipient` and `last chunk is empty` shipped with no provoking fixture; both now have deterministic fixtures built from the real binary.
4. A fourth unnamed refusal, `invalid X25519 recipient block`, was found while writing fixture 3 and is now a marker with its own fixture.
5. The `header-mac` fixture flipped a byte of base64 TEXT and was vacuous about 1 run in 20; it now decodes, flips and re-encodes.
6. The claim "no CURRENTLY KNOWN fault reaches [the unclassified verdict]" was still false; it is replaced with a statement that the branch is reachable, naming the three run-environment faults measured to reach it and why they are deliberately not given markers.
7. The `.lower()` docstring's claim to be the only guard, and the v1.3.1 quotation that dropped its `failed to read header:` prefix, are both corrected.
```

### Round 2 controls

- **Red at the round-1 tip `d72fb128`:** 10 failed, 432 passed.
- **Green on both binaries:** 442 passed on age v1.3.2 (dev shell) and 442 on v1.3.1, with `python3` asserted to still be the dev shell's.
- **Gate, both tiers, dev-host, on the round-2 tree:** `GATE: RESULT=PASS exit=0` — pytest `scripts/tests` **13090 passed, 0 failed** (21:51); node **suites=5 files=41 tests=1449 pass=1449 fail=0**.
- **Mutation, 7 mutants + a control:** all 7 KILLED, each by the guard that owns it — including the two the audit scored SURVIVED (`failed to parse x25519 recipient`, `last chunk is empty`). The behaviour-free comment reword SURVIVED.

### Ledger

`round 2 · payload lines changed THIS round: 176 (since round 1: 427) · scaffolding this round: 234`

Measured with `git log --numstat --format= --remerge-diff d72fb128..deee8a8e --not origin/main` (git 2.55.0, rc 0, silent stderr, non-empty range). Payload = the two `analyze-service-index` sources plus `SECRETS.md`, which for this PR is payload rather than scaffolding: the operator-facing verdict table is a thing this change exists to correct. Scaffolding = `scripts/tests/`. The attribution gate does not fire — this round's fixes are overwhelmingly in the shipped classifier, not in the guards the previous round wrote.

### Not fixed, and still open

The two findings this PR deferred in round 1 stand: `no identity matched any of the recipients` is not exclusively a wrong-key signature beyond its corrected comment, and `plain_present is None` is still reported as "without leaving plaintext". The round-1 audit confirmed both are genuinely out of scope and that the second is *less* reachable after this change.

ZacxDev:
## Round 2 delta audit — findings, and what round 3 (`ce4d7973`) claims to have done

The round-2 audit read `deee8a8e` and attacked the new third state directly. It held: **336 wrong-identity single-byte header corruptions produced ZERO `bad header MAC`**, while the right-identity positive control produced 39 (v1.3.1) and 37 (v1.3.2) — the sweep can see the message and never sees it without a matching identity. It also confirmed by the same discriminating control that `invalid X25519 recipient block` is genuinely pre-auth, and re-derived every number round 2 reported.

It returned 9 findings — no inversion, one real guard gap round 2 created, and four items of prose the round itself wrote or left stale. Round 3 addresses all of them.

```audit-claims round=2 audited=deee8a8e..ce4d7973
1. The marker-ordering invariant was stated over `AGE_REFUSALS_POST_AUTH`, which left `AGE_REFUSED_HEADER_MAC` (in neither set) with its tuple position pinned by nothing; it is now stated over `AGE_REFUSALS_KEY_PROVEN`, plus an assertion that no marker kind may be ungrouped.
2. The `AGE_REFUSED_HEADER_MAC` branch in escrow-verify was keyed on `age_refusal` alone; it now carries the same `cause == DECRYPT_AGE_REFUSED` test its neighbour `_corrupt` calls mandatory.
3. `restore-verify.py`'s own decrypt-failure message said "the identity does not open it, or the ciphertext is damaged" for every refusal; it now reads the classification and says which of the three situations it is in.
4. `invalid X25519 recipient block` was missing from restore-verify's summary table and from the SECRETS.md table; both now list it and both are pinned against `_AGE_REFUSAL_MARKERS` in both directions.
5. SECRETS.md row `33` described only its first message in the round that gave `33` a second one; it now carries the same "read the sentence, not just the number" note row `25` already had.
6. The seam ledger's sentence claimed it walks every published value while walking three hand-named groups; it now walks `AGE_REFUSALS` and requires the strong token to be reachable only from `AGE_REFUSALS_KEY_PROVEN`.
7. The destructive-advice positive control asserted `>= 12` beside a message naming 13, against a real count of 15; the floor is now derived from the pin set rather than spelled.
8. The v1.3.1 in-header-truncation quotation enumerated 2 of the 4 measured shapes; all four are now listed with their `failed to read header:` prefixes.
9. `AGE_REFUSALS_KEY_PROVEN`'s comment claimed a production-consumer role it does not have; it now states narrowly that it owns the marker ORDER and is not read by a branch.
```

### Round 3 controls

- **Red at the round-2 tip `deee8a8e`:** 6 failed, 442 passed.
- **Green on both binaries:** 448 passed on age v1.3.2 and 448 on v1.3.1.
- **Mutation, 5 mutants + a control:** Q3 (delete a summary-table row), Q4 (message ignores the classification), Q5 (an ungrouped kind) **KILLED**. Q1 (move the MAC marker to the end) and Q2 (drop the new branch's cause test) **SURVIVED** and are labelled in the code as equivalent-today, with the measurement behind each. Behaviour-free control **SURVIVED**.

🔴 **Two of round 3's own guards initially passed for the wrong reason, and the battery is what caught them** — both fixed before this commit. The summary-table test sliced the whole comment block and asked whether each marker appeared *anywhere* in it, so deleting a row survived because a ⚠ note below happens to *name* that marker; it now extracts rows structurally. And the new conditional diagnosis had no test at all — `if False:` survived the whole suite.

🔴 **The first battery run was invalid and is reported rather than quietly rerun.** Its restore step used `git checkout -- SECRETS.md`, which reverted round 3's *uncommitted* edits to that file instead of the temporary round-2 copy the control had written; every mutant afterwards died on an unrelated doc guard, **including the must-survive control**. The control dying is the only thing that exposed it. The script now restores by copy and proves it with `cmp`.

### Gate — reported as a union of two runs, not as a gate pass

`scripts/gate.sh --tier both` was SIGTERM'd at its own 3600 s wall-clock cap under heavy host contention (six concurrent gates, load ~70), inside `scripts/claude-hooks/tests`. That is an infrastructure timeout, not a test failure.

- pytest tier: **24 of 29 targets completed, zero failures**, `scripts/tests` **13096 passed / 0 failed** (37:08)
- the 5 targets it never reached plus the one it was killed inside, run separately with the runner's own plugins and `PYTHONPATH`: **701 passed, 0 failed**
- node tier: **suites=5 files=41 tests=1449 pass=1449 fail=0**, `RESULT: PASS`

The union covers all 29 pytest targets with no failure. That is not the same claim as one green gate run, and is not written as one.

### Ledger

`round 3 · payload lines changed THIS round: 137 (since round 1: 564) · scaffolding this round: 203`

Measured with `git log --numstat --format= --remerge-diff deee8a8e..ce4d7973 --not origin/main` (git 2.55.0, rc 0, silent stderr, non-empty range). `SECRETS.md` counted as payload. The attribution gate does not fire — this round changed the shipped classifier, its consumer and the operator-facing doc.

ZacxDev:
## Round 3 delta audit — findings, and what round 4 (`2435b6bb`) claims to have done

The round-3 audit read `ce4d7973`. It re-ran the discriminating control at scale — **336 wrong-identity single-byte header corruptions produced ZERO `bad header MAC`**, against 39 (v1.3.1) and 37 (v1.3.2) from the right-identity positive control — and confirmed by the same method that `invalid X25519 recipient block` is genuinely pre-auth. It returned one 🔴 and eight more, and it did the once-per-ladder pass re-deriving every count and cross-reference against the head.

```audit-claims round=3 audited=ce4d7973..2435b6bb
1. `test_the_real_binarys_in_header_truncations_are_NEVER_post_auth` — the only guard that reads the real binary — still asserted `not in AGE_REFUSALS_POST_AUTH` while its two siblings were widened; it now asserts `not in AGE_REFUSALS_KEY_PROVEN`.
2. The SECRETS.md half of the table guard was a substring test against the whole file, so deleting or inverting the `bad header MAC` row survived; SECRETS.md is now parsed as a table and its rows compared.
3. Both table guards discarded the classification column, so a row could name the right marker and give the wrong answer; both now compare the classification, with a separate directional assertion that a pre-auth marker is never described in key-proven language.
4. `AGE_REFUSALS_KEY_PROVEN`'s comment said it is "NOT read by a production branch" while the same commit added one 560 lines below; its readers are now enumerated, including the one that deliberately does not read it.
5. The key-proven diagnosis asserted the identity "DID open it", which is false for a MAC failure; that arm now says the identity UNWRAPPED a recipient stanza and the header then failed its integrity check.
6. The guard for that message asserted the "DID open" overclaim for the `header-mac` case, i.e. it pinned the wrong sentence; it now asserts the corrected wording and that "DID open" is absent.
7. `AGE_REFUSED_UNRECOGNISED` had no fixture in the message test; `identity-malformed` provokes it from the real binary.
8. The destructive-advice floor was derived from the pin set and therefore implied by the two assertions below it; it is reverted to an independent literal with no count quoted in prose.
9. Three stale claims are corrected: the `decrypt()` cause comment describing `age-refused` as "the identity does not open the artifact", the "only the witness sentence would differ" understatement, and a cross-reference naming a test that does not exist under that name.
```

### Round 4 controls

- **Red at the round-3 tip `ce4d7973`:** 1 failed, 448 passed. 🔴 **That one red is claim 5 only.** Claims 1, 2, 3 and 8 are **invariant guards** — the base's tables are correct, so a guard that starts checking them has nothing to go red against. This matrix must not be read as covering the round.
- **Green on both binaries:** 449 passed on age v1.3.2 and 449 on v1.3.1.
- **Mutation, 5 mutants + a control:** R1 (delete the SECRETS.md MAC row), R2 (**invert** it to pre-auth), R3 (invert restore-verify's summary row) and R5 (revert the DID-open overclaim) **KILLED** — R1 and R2 are the audit's own surviving mutants. R4 (revert claim 1's widening) **SURVIVED** and is labelled in the commit: every in-header truncation on both binaries emits `failed to read header`, so the widening is future-proofing against a reworded age, not a live defect. Behaviour-free control **SURVIVED**.

### Gate — a union of runs, with one contention failure diagnosed rather than assumed

`scripts/gate.sh --tier both` hit its own 3600 s cap under load 78–90, reaching 8 of 29 targets.

- `scripts/tests`: **13094 passed, 3 failed** in 53:19. All three are `subprocess.TimeoutExpired: … timed out after 120 seconds` on a **nested** `run-tests.sh` — those tests spawn their own gate runs on a 120 s budget. All three **pass in isolation** on this tree (3 passed in 200 s), and the target inflated 22:51 → 37:08 → 53:19 across runs, which is the load signature rather than one assertion failing.
- the 21 targets the cap never reached, run **one invocation per target** the way the runner does: **6161 passed, 2 skipped, 0 failed**.
- node tier: **1449/1449**, `RESULT: PASS`.

### Ledger

`round 4 · payload lines changed THIS round: 88 (since round 1: 652) · scaffolding this round: 181`

Measured with `git log --numstat --format= --remerge-diff ce4d7973..2435b6bb --not origin/main` (git 2.55.0, rc 0, silent stderr, non-empty range). Payload = `scripts/analyze-service-index/restore-verify.py`; scaffolding = the two test files. The attribution gate does not fire.

PR #1410, opened 2026-09-08
fix(docs): a CLIENT subdomain literal is committed to this PUBLIC repo — resolve it through bar-url

ZacxDev:
## The problem

`main` fails `scripts/tests/test_no_client_hostnames.py::test_no_client_subdomain_literal_is_committed`, on **both** tiers:

```
a CLIENT subdomain is committed to a PUBLIC repo — this is internal topology
  claudedocs/handoff-civitai-app-fleet.md:204: h.civit.ai
    | curl -sS -o /dev/null -w "$h %{http_code}\n" https://$h.civit.ai/
```

Added by `6d488a1b` (#1402). `CLAUDE.md`: *this repo is PUBLIC — never commit a real third-party hostname used as an example.* The apex is client topology.

## The fix

The snippet is a **live check the operator actually runs**, so placeholdering it would leave a command that goes nowhere. Per the guard's own guidance the *value* moves out of tracked source instead:

```sh
apex=$(bar-url civitai_apps_apex) || exit 3
… curl … "https://$h.$apex/"
```

`bar-url` reads `~/.config/bar/urls.env` (0600, untracked) — the same indirection the `civitai` bar block already uses for its Grafana link, for exactly this reason (`scripts/bar-url`'s own docstring cites the public-repo rule). An unset key **exits 3 naming the key and the file**, so the loop cannot silently check nothing.

## Verification — red→green, control watched failing

Same test, same instrument, `nix develop … -c python3 -m pytest`:

| tree | result |
|---|---|
| `origin/main` @ `27d5028d`, unfixed | **1 failed**, 17 passed |
| this branch | **18 passed** |

**Scope checked, not assumed.** The other tracked `civit.ai` occurrences are all `<slug>.civit.ai` placeholders, which the guard tolerates by design. This changes only the one literal it flags — `git grep` output reviewed rather than counted.

## ⚠ Forward-only

This fixes it going forward. **The literal remains in git history**, and all four content gates read `git ls-files`, so they are blind to it (`SECRETS.md` → *"Dead credentials in reachable history"*). Whether to rewrite history on a public repo is the operator's call, not this PR's.

## Context — `main` is red for TWO independent reasons

This is one of them. The other is #1407 (20 nebula tests failing on the sandbox tier, `/usr/bin/env` absent in the nix build sandbox). They are unrelated and neither fixes the other:

| tree | pytest failures |
|---|---|
| `main` @ `27d5028d` | 21 |
| `main` + #1407 | 1 ← this leak |
| `main` + this PR | 20 |
| both | 0 (verifying) |

### DIFF
--- a/claudedocs/handoff-civitai-app-fleet.md
+++ b/claudedocs/handoff-civitai-app-fleet.md
@@ -198,10 +198,17 @@ done
 git -C ~/workspace/civit/civitai-app-sensei show \
   origin/trunk:src/toolchain-lockstep.test.ts | grep -cE '^\s*it\('
 
-# all six released apps serve
+# all six released apps serve.
+# 🔴 The apex is a CLIENT hostname and this repo is PUBLIC, so it is not
+# committed. Put `civitai_apps_apex=<apex-domain>` (bare domain, no scheme) in
+# ~/.config/bar/urls.env (0600, untracked) and `bar-url` resolves it — the same
+# indirection the civitai bar block uses for its Grafana link. An unset key
+# exits 3 naming the key and the file, so this loop can never silently check
+# nothing. See scripts/bar-url.
+apex=$(bar-url civitai_apps_apex) || exit 3
 for h in app-requests generate-from-model custom-generators \
          playable-collections model-benchmarking sensei; do
-  curl -sS -o /dev/null -w "$h %{http_code}\n" https://$h.civit.ai/
+  curl -sS -o /dev/null -w "$h %{http_code}\n" "https://$h.$apex/"
 done
 
 # the skill's own gates (six named files)


ZacxDev:
### Combined verification complete — and which rows are MEASURED vs DERIVED

The body's last row said "(verifying)". It is now measured.

Built `origin/main` @ `27d5028d` + this branch + #1407 (both merges clean, no conflicts) → merge `eb705ae2`, then `nix build .#checks.x86_64-linux.pytests` on that tree:

```
TOTAL collected=21127  passed=21125  skipped=2  failed=0  (floor: 20342)
RESULT: PASS (exit=0)
```

**Correcting the body's table** — two of its rows were arithmetic, not measurements, and it did not say so:

| tree | pytest failures | provenance |
|---|---|---|
| `main` @ `14126d94` | 20 | **measured** — sandbox tier |
| `main` @ `27d5028d` | 21 | *derived* (20 + this leak) — **not measured directly** |
| `main` + #1407 | 1 | **measured** — round-1 audit, merge `b1460808` |
| `main` + this PR | 20 | *derived* — **not measured directly** |
| `main` + **both** | **0** | **measured** — merge `eb705ae2` |

The endpoints are measured and the arithmetic between them is consistent, but a derived number is a claim and should be labelled as one rather than sitting in a table that reads as five measurements.

ZacxDev:
Obsolete — closing without merging.

`131aa3b0` (#1405) rewrote `claudedocs/handoff-civitai-app-fleet.md` and the `h.civit.ai` literal went with it. Verified against current `origin/main`: no `civit.ai` occurrence remains in that file, and `test_no_client_hostnames.py` is **18 passed** there — the failure this PR fixed no longer exists, which is also why this PR now reports CONFLICTING.

Nothing to salvage: the `bar-url civitai_apps_apex` indirection proposed here only mattered for the snippet that was deleted.

⚠ Two things this closure does **not** resolve, recorded so they are not lost:
- **The literal is still in git history** (`6d488a1b`). All four content gates read `git ls-files` and are blind to history — see `SECRETS.md` → "Dead credentials in reachable history". Whether to rewrite is the operator's call.
- **Nothing pinned this.** The doc was fixed incidentally by an unrelated rewrite, not because anyone was chasing the guard. The next handoff doc that pastes a live `curl` against a client host reintroduces it identically.

`main`'s remaining red is #1407 (20 nebula tests on the sandbox tier), unaffected by this.

ZacxDev:
### Correcting my own closure comment

I wrote that the doc "was fixed **incidentally** by an unrelated rewrite, not because anyone was chasing the guard." **That is false**, and I should have read their replacement before characterising it.

`131aa3b0` (#1405) fixed it deliberately and left the reasoning in place:

```
# 🔴 APEX is the app-hosting apex domain and is DELIBERATELY not committed —
# devrc is a PUBLIC repo and scripts/tests/test_no_client_hostnames.py fails on a
# client subdomain literal. This line spelled it until 2026-09-08 and was RED on
# main for it. Read the apex off `civitai app doctor`, then:
APEX=<the app-hosting apex domain>
```

So they hit the same guard, diagnosed the same cause, and landed a fix while this PR was open — two sessions on one failure again, neither aware of the other. (I did take a `claim-work` claim on it, `claim/civitai-fleet-doc-client-hostname`, but only *after* they had already started; a claim only helps whoever checks it first.)

**This also retracts the second caveat in my closure comment.** I said "nothing pinned this" and that the next doc pasting a live `curl` reintroduces it identically. The first half is wrong in spirit: `test_no_client_hostnames.py` *is* the pin, it fired correctly, and it is what both of us were responding to. The second half stands only in the weak sense that a guard catches a leak after it is committed rather than before — which is a property of a repo-wide `git ls-files` scan, not a gap in this one.

Their approach (`APEX=<the app-hosting apex domain>`, read from `civitai app doctor`) is also better suited to a handoff doc than the `bar-url` indirection this PR proposed — `bar-url` exists for values a *bar button opens*, and a documented one-off check does not need a resolver.

The history caveat is unaffected and still stands: the literal remains in `6d488a1b`, and the content gates are blind to history.

PR #1415, opened 2026-09-08
fix(guard): deny tmux kill-server on a shared socket — the command that destroyed 47 live conversations

ZacxDev:
Deny `tmux kill-server` / `kill-session` unless an explicitly-named private socket makes it safe — and stage the OOM-score protection that was originally asked for.

## Why this exists — a real incident, tonight

At **21:54:15.802 CDT on 2026-09-07** an agent dispatched by a Claude Code session ran, inside a tmux pane, against the operator's live workspace:

```
TMUX_TMPDIR=$SCRATCH/run tmux kill-server
```

42 `tmux-spawn-*.scope` units tore down 1.2s later. **47 live claude conversations were destroyed.** 46 panes were eventually recovered — 20 of them only by manual forensic identification against the agent-ledger and restored pane scrollback — and one was never recovered.

🔴 **`TMUX_TMPDIR` does not isolate a tmux client.** A process running inside a pane reads `$TMUX`, and that socket path wins. Only `-L <socket>` / `-S <path>` isolates.

🔴 **The agent that ran it had MEASURED the non-isolation 72 seconds earlier** — an empty `TMUX_TMPDIR` directory, no socket in it, and yet `tmux list-sessions` returned the operator's real sessions marked `(attached)` — and did not read its own output as the warning it was. Its dispatch brief warned about this exact hazard, in capitals, citing an earlier instance in the same arc that destroyed 43 panes.

**That is the argument for this PR.** A written warning had already failed twice. This is the structural version.

## The guard

`check_tmux_kill_shared_server` in `scripts/claude-hooks/guard_core.py`, wired into `bash-guard.py`. It denies `kill-server` / `kill-session` unless `-L` or `-S` names a non-default socket.

- Matched on the **`kill-s` prefix**, because tmux resolves command abbreviations — a guard matching only the full spelling is walkable by typing less.
- **Measured blast radius: it denies nothing that exists.** Every `kill-server` call in this repo already passes `-L`. Re-measured after merging current main, which brought two more such calls in #1379 — both carry `-L`, so the claim still holds.

## Tests

**The regression test pins the exact command, not a tidied equivalent:**

`test_the_exact_command_that_killed_47_conversations_is_denied` — asserts the literal `TMUX_TMPDIR=$SCRATCH/run tmux kill-server` is denied. Three properties are load-bearing and a cleaned-up fixture would lose all three:

1. **The `TMUX_TMPDIR=` env prefix is kept** — that prefix is the whole reason the command looked safe. A guard tested only against a bare `tmux kill-server` is untested against the form that actually occurred.
2. **`$SCRATCH/run` stays unexpanded** — the guard sees the pre-expansion string.
3. **It asserts this guard's OWN reason reaches the caller** (`evaluate()`'s result must equal this check's result), so a neighbouring deny cannot satisfy it.

`test_the_isolated_spelling_the_deny_message_prescribes_is_still_allowed` — the discriminating control. `tmux -L my-probe-9182 kill-server` must still be ALLOWED. Without it, a guard hardcoded to refuse everything would pass the regression test while breaking every legitimate use.

A third added test passes at base and is labelled an **invariant guard**, not counted as regression coverage.

**Mutation:** two sweeps. The blanket-deny mutant is killed by the `-L` ALLOW control — exactly what that control exists for. One mutant was **mis-designed** (`commands(cmd)[:1]` does not model env-stripping, since `_peel_variants` strips `VAR=` during primary peeling); it was replaced by mutating `_ASSIGN` itself, and the incident test is among the killers — so the env prefix is genuinely load-bearing rather than decorative.

## The OOM half — staged, not applied, and it does NOT address the measured cause

The dispatching brief hypothesised an OOM kill. **That hypothesis is refuted**, with instrument-validated negatives rather than bare absences:

- **Kernel OOM killer did not run.** Zero `oom-kill:` / `Killed process` lines for the incident boot. **Positive control:** the same grep over the same journal *does* match real kernel OOM kills from 2026-08-28, so the pattern can fire and didn't.
- **systemd-oomd is not installed at all**, so `ManagedOOMPreference` would have been decoration.
- **tmux did not crash** — core limit `unlimited`, `core_pattern` pipes to systemd-coredump, no tmux core exists.
- **The alarming `57.7G / 38.2G / 29.8G` figures are not what they look like.** They are systemd's per-scope **lifetime** `memory peak`, printed at teardown, over 7–23h wall clocks — high-water marks for scopes that mostly no longer overlapped. Not concurrent usage, and not summable.
- **The V8 `FatalProcessOutOfMemory` was `tsserver.js`**, hitting its own per-process heap cap in a different scope **24ms after** the real cause.

`nix/system/apply-tmux-oom-protection.sh` is therefore **staged and honestly labelled**. It cannot be home-manager: unprivileged processes cannot lower `oom_score_adj` (EPERM, measured), and the tmux server is not in an HM unit — it lives in a per-login `session-N.scope`.

🔴 **On what it displaces:** raising `OOMScoreAdjust` on the pane scopes was **rejected** — it reaches the same relative ordering by making the conversations *more* killable, which buys an empty workspace. It lowers one process only, and to **−500 rather than −1000**, so tmux stays reclaimable. For context on whether it is worth applying at all: tmux currently scores 668 against a box top of 1332–1333.

## 🔴 Gate status — owed, and stated plainly

Earlier runs on base `5d6fffa7` reported: dev-host node **PASS** (1449); dev-host pytest 8 failures, **0 attributable to this diff** — 6 were the `age` 1.3.2 nixpkgs drift **since fixed on main by #1392**, and the other two did not reproduce; sandbox **nodetests PASS** read from `nix log`.

**That is not sufficient and I am not claiming it is.** `main` has moved substantially since, this branch has been merged with it (head `24c77099`), and **no tier has been run on that merged tree**. The box is at load ~62 with two other sessions gating; adding a fourth concurrent suite is the exact condition that preceded tonight's incident, so the run is deliberately deferred rather than rushed.

**Do not merge until both tiers are green on the merged tree** — `scripts/gate.sh --tier both`, then the two `nix build` derivations **one at a time**, with verdicts read from `nix log` rather than build exit codes.

## Blast radius

**Reversible** — 793 insertions, 0 deletions; purely additive. The guard denies a command class that no existing caller uses. The sudo script changes nothing until someone runs it.


ZacxDev:
## audit-claims — round-2 fix pass on #1415

Per prior finding: **fixed / partially / not / made worse**. Severity words, never coloured circles.

| # | finding | status |
|---|---|---|
| BLOCKER 1 | `pgrep -u 1000 -x tmux` selects nothing; unit reports success forever | **FIXED** |
| BLOCKER 2 | import-wiring awk cannot match this host's `imports =` shape | **FIXED** |
| SHOULD-FIX 3 | unresolved `-L`/`-S` accepted as proof of isolation | **FIXED** (denied; argued below) |
| SHOULD-FIX 4 | `ssh <host> tmux kill-server` is ALLOW | **FIXED** (targeted arm, blast radius measured) |
| SHOULD-FIX 5 | prose "MEASURED BLAST RADIUS" ledger already wrong | **FIXED** (derived + test-enforced) |
| NIT 6 | red-at-base is an import-surface red | **FIXED** |
| NIT 7 | `mv` inode, orphan module, backup churn, lying `restore()` | **FIXED** (all four) |
| NIT 8 | remediation text does not fit `kill-session`; no stated escape hatch | **FIXED** |
| CONTEXT | `tekton/devrc-pytests` red on old head | **RE-BASED**; named a concrete candidate cause; awaiting the run on the new head |

---

### BLOCKER 1 — selector

Selector is now `pgrep -u 1000 -x 'tmux: server'`. Re-measured on the workbench 2026-09-09:

```
pgrep -u 1000 -x tmux             -> rc=1, no output
pgrep -u 1000 -x 'tmux: server'   -> 1111077, rc=0
cat /proc/1111077/comm            -> tmux: server
```

Three things changed, because the selector was only half the defect — the other half was that
an empty match set was the **silent** path:

1. **Selector corrected**, and the in-loop `/proc/<pid>/comm` re-read tightened from a `tmux*`
   prefix (which also accepts `tmux: client`) to the exact string.
2. **Apply-time positive control.** The script now runs the selector itself, under sudo, and
   **refuses to install** when it matches nothing. That is the control the every-2-minutes unit
   structurally cannot perform — there, `found 0` is indistinguishable from "no tmux running".
3. **Run-time counters split.** The unit counts `found` separately from `adjusted`, reads the
   written value **back** out of procfs, and exits non-zero when it found servers it could not
   lower. `found 0` is deliberately *not* escalated: after a clean shutdown there genuinely is no
   server, and failing on that would toast four times an hour forever.

**Behavioural verification, not just a text check.** The generated unit body was extracted and
executed as the unprivileged user:

```
tmux-oom-protect: found 1 tmux server process(es), adjusted 0 to oom_score_adj=-500
tmux-oom-protect: FAILED to lower 1 of 1 server process(es)
rc=1
```

`found 1` is the corrected selector locating the real server (the old one found 0); `adjusted 0`
is the documented EPERM for an unprivileged lower; `rc=1` is the new loud path firing. Afterwards
`cat /proc/1111077/oom_score_adj` still reads `0` and the server is alive — nothing was mutated.

`test_process_selection_never_uses_a_full_commandline_pattern` no longer certifies the bug: it now
says in its own docstring that it checks flag SHAPE only, and the selector's *behaviour* is a
separate test asserting it against the realistic `comm` (`tmux: server`), with the shipped-broken
`tmux` pattern kept as the negative control so a "tidied" fixture goes red.

### BLOCKER 2 — import wiring

Copied from `apply-airvpn-host.sh:88-116` rather than reimplemented, **including the
multiple-`imports =` guard this script had dropped**. Also reordered: the import is wired *before*
`$MODULE` is written, so a refused wiring leaves no orphan in `/etc/nixos`.

Verified by running the awk, not by reading it — against the real `/etc/nixos/configuration.nix`
(copied read-only) and against both list shapes:

```
new awk on the real config      -> inserts at line 25, inside the list; nix-instantiate --parse OK
OLD awk on the same real config -> 0 insertions   (negative control: the blocker reproduces)
```

Duplication across two standalone `sudo bash` scripts is deliberate (neither can import from the
other), so a test pins that both still carry the same load-bearing pieces — the `[`-character
split, the assignment count, the comparison that makes the count a guard, the refusal and its
non-zero exit — **as five separate claims**. That granularity was not gratuitous: see the mutation
table, where a one-fragment version let a mutant through.

### SHOULD-FIX 3 — unresolved `-L`/`-S`: **DENY**, and here is the argument

I took your lean. `_socket_is_proven_isolated` now requires the socket's **last path component**
to retain a literal character after every recognised expansion is deleted, and that component must
not be `default`.

- `-S "${TMUX%%,*}"`, `-S "$TMUX_SOCK"`, `-L "$sock"`, `-L ''`, `-S ''` — all now **DENY**.
  Measured, and it is not theoretical: `$TMUX` on this box is
  `/run/user/1000/tmux-1000/default,1111077,20`, so `${TMUX%%,*}` *is* the operator's own socket.
- `-L my-probe-$$` — still **ALLOW**. The rule is not "contains no variable"; that would deny the
  very command the message prescribes.

**Argued against the repo's real callers, as asked.** All four `-L`-carrying sites are Python
`subprocess.run([...])` argument lists. This hook gates Bash-tool **text**, so it can never see
them — denying the shell spelling `-L "$sock"` breaks nothing that exists in this tree today, and
the workaround is one word (write the label literally), stated in the deny message.

Two more fail-opens surfaced while building this and are closed:

- **Leftover metacharacters.** `` tmux -L `echo default` kill-server `` tokenises with an
  unbalanced backtick; the socket became the literal-looking string `` `echo ``, which survived
  the expansion strip and read as proven-isolated. Anything the expansion table could not consume
  is now unproven.
- **Elided command substitution.** `commands()` lifts `$(…)`/backticks into their own argv and
  **deletes** them from the outer one, so `-L` swallowed the *subcommand* and the guard read a
  socket named `kill-server`. Now dropped as incoherent.

Also: the action is detected by scanning **every** token, with only the socket read positionally.
Resolving the subcommand positionally is exactly how a token miscount fails open.

The scope is now stated in code: this protects **one socket name**, `default`.
`tmux -L work kill-server` is ALLOW by design.

### SHOULD-FIX 4 — ssh

**Blast radius measured before choosing, as asked — and adding `ssh` to `_WRAPPERS` is the wrong
fix for a mechanical reason, not a taste one.** `_peel_variants` treats a wrapper's first non-flag
token as the command; for ssh that token is the **host**. `ssh laptop tmux kill-server` peels to
`['laptop', 'tmux', 'kill-server']` — argv[0] `laptop` — so the check would *still* miss it while
every other check in the policy started evaluating hostnames as commands. Wider radius, hole open.
That is pinned by a test that fails if `ssh` ever starts peeling.

So: a targeted arm that scans an ssh argv for a `tmux` **token** at any position and reads flags
from there. It needs no ssh option-arity table (getting one wrong is the fail-open mode), and
`ssh -L 8080:localhost:80 host tmux -L probe kill-server` reads correctly — the port-forward `-L`
is skipped, the tmux `-L` is the socket. Cost is a bounded over-block on
`ssh host echo "tmux kill-server"`, which is the right side to err on.

Measured on the tree: the only shell-text `ssh … tmux …` in this repo are reads
(`list-panes`, `is-active`) — both still ALLOW.

### SHOULD-FIX 5 — the ledger

The prose enumeration is gone. In its place:

- `test_every_kill_server_call_site_in_the_repo_is_classified` — a **two-way** ledger of every
  tracked file mentioning a wide kill, keyed on **paths, not line numbers** (the line number is
  what rotted last time). A new or removed file fails the suite.
- `test_every_executed_tmux_kill_argv_names_an_isolating_socket` — every tmux argv list either
  carries `-L` or is named in an exemption dict **with the reason** (the reason must be that the
  surrounding assertion is that the call is *refused*; "it's a test" is not a reason).
- `test_no_tracked_shell_text_writes_a_kill_this_guard_would_deny` — the population the guard
  *can* see.
- Both scanners carry a **positive control** (`test_the_kill_site_scanner_can_see_anything_at_all`)
  so a broken regex cannot report a clean tree as a reassuring zero.

The ledger earned its keep on its first run: it flagged a site I had just written (a scanner
fixture in the test file itself) that I had not classified.

**And the honesty you asked for is now in the docstring:** the ALLOW measurement is reassurance
about the **wrong population**. Those sites are Python argument lists this hook structurally never
sees; the guard could deny every one of them and nothing here would break. The population that
matters is shell text an agent types, and this repo contains none of it — so the measurement is a
true fact about a set that cannot be affected, not a clean bill of health for the set that can.

These scanners reuse `public_ip_scan.repo_files` rather than growing a third copy of the
git-ls-files/walk logic — which also buys the sandbox tier, where a bare `git ls-files` errors and
a test that caught that error would skip its way to green in exactly the tier the merge is gated
on. **Verified in a git-less copy of the tree** (1298 tracked files, no `.git`): 24 passed.

### NIT 6

Assertion order swapped: `evaluate(_INCIDENT_COMMAND, "claude-code") is not None` is now line one,
with the `AttributeError` reasoning written into the docstring so it does not get "tidied" back.

### NIT 7 — all four

- `mv "$CFG.new" "$CFG"` → `cat "$CFG.new" > "$CFG"` (inode + `0644 root:root` preserved), matching
  `apply-airvpn-host.sh:114-115`. `$MODULE` also gets an explicit `chown`/`chmod`.
- `$MODULE` is written **after** the wiring succeeds.
- Backup name is now **fixed** (`.bak.tmux-oom`), taken only on the branch that actually edits, so
  a re-run overwrites one file instead of adding a nineteenth to the 18 already in `/etc/nixos`.
- `restore()` checks the backup exists and otherwise prints
  `"$CFG was not modified by this run; nothing to restore"`.

### NIT 8 — remediation text

Rewritten as three branches by what the caller wanted: (a) experiment with a server → make your own
and kill that; (b) remove one session/window/pane → `kill-pane`/`kill-window` are allowed, and
there is deliberately **no** allowed spelling of `kill-session` on the default socket, because it
takes panes you did not name; (c) target a real non-default server → allowed, but write the socket
name **literally**, and here is why `-L "$sock"` is not.

Plus the sentence you asked for: the escape hatch for a genuinely-needed shared `kill-server` is
that this hook gates the Bash **tool**, and the operator's own terminal is not hooked — ask them.
Do not go looking for a spelling that slips past the check.

---

### Mutation results — 18 mutants, 18 KILLED, each by its own guard's assertion text

Control run GREEN (106 tests) before and between mutants; final full run of both affected
files on the committed tree: **1556 passed, 0 failed** (284s, dev-host tier, base `c507d71d`). `PYTHONDONTWRITEBYTECODE=1` throughout.
Node IDs, never `-k tmux` (which deselects
`test_the_exact_command_that_killed_47_conversations_is_denied`). Both files restored from `cp -a`
copies afterwards and confirmed byte-identical.

|---|---|---|
| **M1 selector back to the shipped-broken `-x tmux`** | **KILLED** | `test_the_pgrep_pattern_matches_a_REALISTIC_comm` |
| **M2 selector → `-x zzz-no-such-proc`** (the mutant that **SURVIVED** in round 1) | **KILLED** | same |
| M3 awk `[`-split disabled | KILLED | `test_the_wiring_awk_actually_inserts_in_both_config_shapes` |
| M4 multiple-`imports =` guard deleted | KILLED¹ | `test_the_import_wiring_is_the_SAME_logic_…` |
| M4b comparison neutralised (count taken, never acted on) | KILLED | same |
| M4c refusal removed from the guard branch | KILLED | same |
| M5 `cat` → `mv` | KILLED | `test_the_config_is_overwritten_in_place_not_replaced_by_mv` |
| M6 fixed backup → timestamped | KILLED | `test_the_backup_does_not_accumulate_one_file_per_run` |
| M7 `found>0` escalation disabled | KILLED | `test_an_empty_match_set_is_not_allowed_to_be_the_silent_path` |
| M8 apply-time pre-flight removed | KILLED | same |
| M9 `/proc` comm re-check loosened to `tmux` | KILLED | `test_the_comm_recheck_in_the_loop_uses_the_same_realistic_value` |
| M10 empty-residue check removed | KILLED | `test_an_unresolvable_socket_is_not_an_isolation_claim` |
| M11 leftover-metacharacter check removed | KILLED | `test_socket_isolation_is_proven_not_assumed` |
| M12 elided-substitution guard removed | KILLED | `test_an_unresolvable_socket_is_not_an_isolation_claim` |
| M13 prefix-OF relaxed to bare `startswith` | KILLED | `test_tmux_kill_near_misses_stay_allowed` (`kill-server-test`) |
| M14 ssh arm removed | KILLED | `test_tmux_kill_over_ssh_is_denied` |
| M15 default-socket comparison neutralised | KILLED | `test_tmux_kill_shared_server_is_denied` (`-L default`) |
| **PC** positive control: drop `-f` from `_TMUX_VALUE_FLAGS` (known fatal) | **KILLED** | `test_tmux_value_flag_table_covers_every_value_taking_server_flag` |

¹ **M4 SURVIVED on the first sweep** and that was a real finding, not noise: the test asked only
for the token `n_imports`, which still appeared in the `if` that compared it — the guard was gone
while the test read as covering it. A guard on a word is walkable by rewording; the check now names
the count, the comparison and the refusal as separate claims, and M4/M4b/M4c all die.

### CI

`tekton/devrc-pytests` was red on `24c77099` on
`test_the_SUMMARY_BANNER_names_the_real_selection_source`
(`scripts/tests/test_run_tests_targets.py`), a file this PR does not touch. `origin/main` merged;
new base `c507d71d` (was 19 commits behind).

A concrete candidate rather than "base/environment": **`9300f234` (#1406) touched that exact test
file and `run-tests.sh`, and is absent from the old base.** The four most recent merged PRs on
bases that include it (#1432, #1427, #1426, #1425) all have `pytests=SUCCESS`. That is
**correlation and a plausible mechanism, not a proof** — the measurement is the tekton run on this
new head. If it stays red on a current base I will report it rather than merge through it.

### What was NOT run, and why

The box was at load 72-99 throughout (six other sessions gating). Per the standing instruction not
to gate above ~25, **I did not run `scripts/gate.sh --tier both` or the two `nix build` check
derivations locally.** What was run: the affected test files directly (dev-host tier), plus a
git-less-tree run of the new corpus-scanning tests to cover the sandbox tier's file-listing path.
Naming this rather than implying a full gate.

### Split into two PRs?

**Recommendation: keep them together — but it is your call, and I did not split unilaterally.**

The auditor's reasoning was "the guard is sound, both blockers are in the script". Both blockers
are now fixed *and* gated, so the thing that made splitting attractive has largely gone. And
merging the script **applies nothing**: it is staged, inert until someone runs it under `sudo`, so
carrying it costs no operational risk.

The honest asymmetry that remains is **verification confidence, not correctness**: the guard is
verified behaviourally end to end, while the script's real runtime effect (`nixos-rebuild` + the
timer firing every 2 minutes) cannot be exercised from here at all. If you would rather ship the
guard *now* and keep the OOM prophylaxis decision open — its own header says it is staged for an
explicit decision, against a mechanism that has never fired on this host — splitting is the right
call and I will do it on a word.

ZacxDev:
### CI re-check — resolved

Both tekton legs are green on the re-based head `a8c47b05`:

```
tekton/devrc-nodetests: SUCCESS
tekton/devrc-pytests:   SUCCESS
```

So the `test_the_SUMMARY_BANNER_names_the_real_selection_source` failure was **base-related, not
this diff** — as hypothesised, and now measured rather than argued. That closes the CONTEXT row in
the comment above, which said "awaiting the run on the new head".

⚠ **Neither of those checks blocks a merge.** `main` is protected in name only and the block is
DECLARED off in `scripts/drift-check.sh` — a green check here is a claim about this branch, not
about the tree a merge creates. Gating the merged tree is still a manual step.

⚠ **And this fix is not live on the workbench.** Measured, with a positive control: the deployed
`~/.claude/hooks/guard_core.py` (`/nix/store/y9dwn6x81xk5sdkma2svd5qrkzj9xxan-hm_guard_core.py`)
has 3 hits for `check_pkill_full_pattern` and **0** for `check_tmux_kill_shared_server`. The guard
only starts protecting anything after merge → pull → `home-manager switch` / `ship.sh`.

ZacxDev:
Machine-readable restatement of the round-1 fix pass above, in the fenced form
`scripts/audit-dispatch.py` parses. The prose comment used a `## audit-claims` **heading**,
which the parser does not accept, so round 2 could not be dispatched. Content is unchanged —
this is a format fix, not a new claim.

```audit-claims round=1 audited=24c77099..a8c47b05
1. BLOCKER 1 — the every-2-minutes unit's process selector was changed from `pgrep -u 1000 -x tmux` (matches nothing; `comm` is `tmux: server`) to `-x 'tmux: server'`, the in-loop `/proc/<pid>/comm` re-read was tightened from a `tmux*` prefix to the exact string, an apply-time positive control was added that refuses to install when the selector matches nothing, and run-time counters were split into `found` vs `adjusted` with a readback and a non-zero exit when servers are found but cannot be lowered (`found 0` deliberately not escalated).
2. BLOCKER 1 (cont.) — `test_process_selection_never_uses_a_full_commandline_pattern` was claimed to have stopped certifying the bug: its docstring now says it checks flag SHAPE only, and selector behaviour moved to a separate test asserting against the realistic `comm`, with the shipped-broken `tmux` pattern kept as a negative control.
3. BLOCKER 2 — the `/etc/nixos` import-wiring awk was replaced with the one copied from `apply-airvpn-host.sh:88-116`, including the multiple-`imports =` guard this script had dropped, and the ordering was changed so the import is wired BEFORE `$MODULE` is written so a refused wiring leaves no orphan.
4. SHOULD-FIX 3 — `_socket_is_proven_isolated` now DENIES an unresolved `-L`/`-S` socket: the socket's last path component must retain a literal character after every recognised expansion is deleted and must not be `default`; leftover metacharacters (e.g. an unbalanced backtick) and elided command substitution are both treated as unproven; the action is detected by scanning every token with only the socket read positionally.
5. SHOULD-FIX 4 — `ssh <host> tmux kill-server` is denied by a targeted ssh arm that scans the ssh argv for a `tmux` token at any position and reads flags from there, rather than by adding `ssh` to `_WRAPPERS` (which was measured to still miss, because `_peel_variants` reads the wrapper's first non-flag token — the HOST — as the command); a test pins that `ssh` never starts peeling.
6. SHOULD-FIX 5 — the prose "MEASURED BLAST RADIUS" enumeration was replaced by a two-way ledger keyed on paths not line numbers (`test_every_kill_server_call_site_in_the_repo_is_classified`), an argv-level check that every executed tmux kill names an isolating socket or is exempted with a reason, a check that no tracked shell text writes a kill this guard would deny, and a positive control that the scanner can see anything at all; the docstring now states the ALLOW measurement covers a population this hook structurally cannot see.
7. NIT 6 — assertion order in the red-at-base test was swapped so `evaluate(_INCIDENT_COMMAND, "claude-code") is not None` is first, with the `AttributeError` reasoning written into the docstring.
8. NIT 7 — all four: `mv "$CFG.new" "$CFG"` became `cat "$CFG.new" > "$CFG"` (inode and mode preserved) with explicit `chown`/`chmod` on `$MODULE`; `$MODULE` is written after the wiring succeeds; the backup name is fixed (`.bak.tmux-oom`) and taken only on the branch that edits; `restore()` checks the backup exists and otherwise says the config was not modified.
9. NIT 8 — the remediation text was rewritten as three branches by caller intent (make your own server; `kill-pane`/`kill-window` with deliberately no allowed `kill-session` on the default socket; a real non-default server with the socket name written literally), plus a stated escape hatch that the hook gates the Bash tool only and the operator's own terminal is not hooked.
10. Mutation battery — 18 mutants claimed KILLED including round 1's survivor (selector `-x zzz-no-such-proc`), each by its own guard's named assertion, with a green control run before and between mutants and `PYTHONDONTWRITEBYTECODE=1` throughout; M4 is claimed to have survived a first sweep because the test asked only for the token `n_imports`, and to have been split into three separate claims (count, comparison, refusal) so M4/M4b/M4c all die.
11. CI — the `tekton/devrc-pytests` red on `24c77099` (`test_the_SUMMARY_BANNER_names_the_real_selection_source`) is claimed to be base-related rather than caused by this diff, on the basis that both legs are SUCCESS on the re-based head `a8c47b05`.
```

⚠ Two things the round-1 pass states it did NOT do, carried forward so round 2 does not
assume them: `scripts/gate.sh --tier both` and the two `nix build` check derivations were
**not** run locally (the box was at load 72–99); and the SHOULD-FIX 3 argument that denying
the shell spelling `-L "$sock"` breaks no real caller — because all four `-L`-carrying sites
are Python `subprocess.run([...])` lists the hook never sees — is **unaudited**.

ZacxDev:
## audit round 2 — delta `24c77099..a8c47b05` (the round-1 fix pass)

Scope: the delta only. `a8c47b05` is the fix commit; the rest of that range is a merge of
`origin/main`. Verified behaviourally where possible rather than by reading.

### What I re-verified rather than took on the claims block

**The guard itself: 29/29 adversarial cases correct, with controls.** Ran
`check_tmux_kill_shared_server` from the MERGED tree (`origin/main` + `a8c47b05`) against a
battery covering every claim in round 1: the incident command, `-S "${TMUX%%,*}"`,
`-S "$TMUX_SOCK"`, `-L "$sock"`, `-L ''`, `-S ''`, both elided-substitution spellings,
`-L default`, `-S …/default`, all four kill abbreviations, the `kill-server-test` session-name
near-miss, `kill-pane`/`kill-window`, the ssh arm (bare, quoted, and with a port-forward `-L`
that must not be read as the tmux socket), and the two documented blind spots asserted as
STILL blind. **0 mismatches.** Controls: the deny path fires, the allow path fires, and the
check is reachable through the real `evaluate(cmd, "claude-code")` entry point — not just
callable directly.

**SHOULD-FIX 3's unaudited argument holds.** Round 1 argued that denying the shell spelling
`-L "$sock"` breaks nothing because every real call site is a Python `subprocess.run([...])`
list the hook never sees, and flagged that reasoning as the thing round 2 should check first.
Re-derived independently over `git ls-files`: every executable `kill-s…` site is a Python argv
list, and the only shell-TEXT occurrences in the tree are prose in handoff docs and this
script's own header comment. The one `-L`-less site
(`scripts/tests/test_session_manager.py:114`) is classified in the ledger, and its surrounding
assertion genuinely *is* that the call is refused (`pytest.raises(_Forbidden)`), which is the
ledger's own stated bar for an exemption. **Claim verified.**

**All 10 spot-checked tests named in the mutation table exist** at the paths claimed.

---

### 🟡 SHOULD-FIX 1 — a re-run whose rebuild fails silently reverts `/etc/nixos/configuration.nix`, including unrelated operator edits

`nix/system/apply-tmux-oom-protection.sh`. Introduced by NIT 7's fixed-backup change, and not
present in the timestamped spelling it replaced.

`BACKUP="${CFG}.bak.tmux-oom"` is a **fixed** name, but the backup is only taken inside the
`else` branch that does the wiring. On a re-run the script takes the already-wired path and
creates no backup — while `restore()`, armed by `trap restore ERR` for the whole script,
still finds the **previous run's** backup and `cp -a`s it over the live config.

`nixos-rebuild switch` runs at the end under that trap, so any rebuild failure — for a reason
having nothing to do with this module — triggers it.

**Measured, not reasoned.** Ran the real script with only its environment couplings patched
(paths, the `$EUID` check, the selector pre-flight, and `nixos-rebuild`); the backup, trap and
`restore()` logic are untouched:

```
RUN 1 (rebuild succeeds)  -> import wired, configuration.nix.bak.tmux-oom created, rc=0
operator then hand-edits configuration.nix   (services.openssh.enable — an unrelated edit)
RUN 2 (already wired; rebuild FAILS)
  "FAILED — restoring …/configuration.nix from …/configuration.nix.bak.tmux-oom"
  import present in CFG:  0     <- the live, correctly-applied import is gone
  operator edit present:  0     <- the unrelated edit is gone
  CFG identical to the pre-wiring backup: YES
```

So the message says it is unwinding this run, and it is actually reverting the file to a state
from an arbitrarily earlier run. The `cp -a` also restores the old mtime, which removes the
obvious tell.

**Why the existing test cannot see it.**
`test_restore_does_not_claim_to_restore_a_backup_that_does_not_exist` asserts that the literal
`-f "$BACKUP"` appears inside `restore()` and that the string `nothing to restore` is present.
That is a guard on the SPELLING of the source, and the spelling it pins is exactly what makes
the stale restore fire — the `-f` test passes because a *stale* backup exists. The test reads
as covering restore honesty while providing none of it, which is worse than no test because it
stops anyone looking.

**Fix direction:** gate `restore()` on whether *this run* took the backup (a flag set next to
the `cp -a`), not on whether the file exists — and pin it behaviourally (run 1 succeeds, run 2
fails, assert the config still carries the import) rather than by grepping the source.

### 🟢 NIT 2 — the refusal path leaves a backup behind

On the `could not find an 'imports =' list` exit, `cp -a "$CFG" "$BACKUP"` has already run, so a
refusal that modified nothing still deposits `configuration.nix.bak.tmux-oom` in a directory the
script's own comment notes already holds 18 such files. Cosmetic on its own; it also arms
SHOULD-FIX 1 for the next run.

---

### Ledger

Round 1 claimed 11 items. Verified: the guard behaviour (claims 1–5, 7–9 by battery and by
reading), the blast-radius ledger and its exemption (claim 6), and the existence of every test
named in claim 10. **Round 2 attributes SHOULD-FIX 1 and NIT 2 to claim 8** (NIT 7's fixed
backup name + honest `restore()`) — both are defects the round-1 fix introduced, which is the
pattern this arc's handoff already records for #1311, #1317 and #1344.

Not re-run by me: the 18-mutant sweep (re-running it requires mutating files the gate is
currently reading). Claim 11 (the pytests red was base-related) is superseded — I am gating the
merged tree directly.

### Verdict

**The guard half is safe to merge** — it is the only half that does anything on merge, and it is
now verified behaviourally end to end rather than on the fix pass's own authority.

**Both findings are in the staged sudo script, which merging does not run.** Its runtime effect
cannot be exercised from here at all, and it has now produced a real defect in each of two
consecutive audit rounds. Merging it changes nothing on disk; the findings are being carried
straight into a follow-up rather than held against the guard, because the guard is currently
NOT DEPLOYED and the failure mode it prevents is live for every agent running right now.

PR #1417, opened 2026-09-08
docs(handoff): #1406 round-1 audit fixes are COMMITTED, and the branch now carries a TRUE merged tree

ZacxDev:
Interim update to `handoff-cairn-oss-multi-instance.md`, forced by one sentence that had become false in the dangerous direction.

**The correction.** The doc said *"SIX FILES ARE UNCOMMITTED in that worktree — the round-1 audit fix round, which exists nowhere else."* They are now `f9ff05c4`, and `~/workspace/devrc-flake-pin` is clean. A reader acting on the old sentence would treat committed work as at-risk, or re-do it.

**The base moved.** `feat/cairn-flake-pin` had drifted **3 ahead / 17 behind** `origin/main`. `git merge-tree --write-tree` exits **0** — no textual conflict, read by EXIT CODE and not by a marker grep — but main had touched `SECRETS.md`, three `scripts/tests/` files and two skills, all sharing ratchets with that diff, so the merge is in (`b79cf63a`, `origin/main` is now an ancestor) and the gate is being re-run on the merged tree rather than on the branch.

**Two new investigation blocks**, both appended rather than replacing:

1. `tekton/devrc-pytests` is red on #1406 at the now-stale head `f98be263`, naming a test the diff does not touch. Ruled out that the tier is broken (#1411 passed it in the same window, `collected=21163 failed=0`) and that the other five red devrc SHAs share a cause (four distinct verdict classes, two of them `KILLED`/`BROKEN GATE` congestion). **Not ruled out and recorded as the honest gap:** the run has been pruned, so which assertion fired is unrecoverable — the same shape as rank 6.
2. Round 1's two behavioural findings, with a both-ways mutation matrix. The decoy `cairnPackage = pkgs.hello;` + `cairnUnused = cairn.packages.${system}.cairn;` **SURVIVED at `f98be263`** and is killed now; a legal multi-line `nix/home.nix` header was **RED at base with a false diagnosis** and is green now. 6/6 mutants killed by their intended guard's own message.

No `clawgate-task:` field: `clawgate_handoff.sh resolve` exits **5** with its positive control green, and a wrong session id also answers 200 with an empty array — that is not a clean bill of health.

https://claude.ai/code/session_01NTRPN6dhkdf8kZJch3mne5

### DIFF
--- a/claudedocs/handoff-cairn-oss-multi-instance.md
+++ b/claudedocs/handoff-cairn-oss-multi-instance.md
@@ -33,35 +33,52 @@ is the PRIVATE proposal, not this doc.
 - **devrc #1394 — ✅ MERGED `65d8bfba`.** Records the operator decision of 2026-09-08:
   **CONSOLIDATE ONTO THE PIN.** The fork is CLOSED; it said "unanswered" in three places and
   all three were fixed, because one stale copy re-opens a settled question.
-- 🔴 **RANK 3 HALF 2, SLICE 2 IS CLAIMED AND IN FLIGHT — PUSHED as devrc PR #1406**
-  (`feat/cairn-flake-pin`, head `f98be263`, worktree `~/workspace/devrc-flake-pin`).
-  `claim-work cairn-oss-multi-instance-3` is HELD for exactly that slice, NOT the
-  entry_shape/writer consolidation. ⚠ **SIX FILES ARE UNCOMMITTED in that worktree** — the
-  round-1 audit fix round, which exists nowhere else. Do not delete the worktree.
-  ⚠ Earlier revisions of this doc described commit `4c77daab` as "unpushed, no PR yet"; that
-  is superseded — do not read it as current or conclude the slice stalled.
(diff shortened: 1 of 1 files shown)


ZacxDev:
Closing as **superseded, not abandoned** — and the reason is the finding.

While this sat open, four commits touched `claudedocs/handoff-cairn-oss-multi-instance.md` on `main`, including `c756efe0`, which makes the **same** "six files uncommitted is now FALSE" correction this PR exists for. Another session got there first.

`main`'s copy is now **1,189 lines**; this branch's is **1,129**. Merging would have replaced the newer document with the older one and silently dropped the other sessions' content — the stale-base hazard `handoff_doc.py` refuses at `status=stale-base`, arriving here by the back door because a PR merge does not run that check.

Nothing is lost: the content unique to this branch — the round-1 mutation matrix and the Tekton investigation block, neither of which is on `main` — is re-landed on top of current `main` in **#1425**, together with the round-3 and merge outcomes.

⚠ The transferable bit: **a docs PR left open across a busy day is a stale-base hazard, not a harmless one.** Land a handoff delta in the same session that writes it, or re-derive it against current `main` before merging.

PR #1428, opened 2026-09-09
fix(cairn-cutover): retire P3 post-cutover — its exit-8 refusal is the last guard, not a bug

ZacxDev:
fix(cairn-cutover): retire P3 post-cutover — its exit-8 refusal is the last guard, not a bug

P3 pushes LOCAL -> POD. That is correct exactly once — while a host still holds
bytes the pod has never seen — and destructive afterwards. This script's own
header already states the post-condition: "After this cutover the hosts' stores
are caches of the pod."

What it LOOKED like: `plan.shippable` is ADD + SUPERSEDES + MERGED, and
SUPERSEDES/MERGED are by definition entries whose pod bytes differ, which is
exactly what `seed.sh`'s pre-flight refuses. So post-cutover P3 exits 8 the
moment anything supersedes, and reads as a shipped code path that can never
complete — one `--allow-overwrite` away from working.

🔴 IT IS NOT. That refusal is the LAST GUARD between a stale mirror and
authoritative content, and `--allow-overwrite` is the single change that would
disarm it. `seed.sh`'s tar adds and overwrites but never deletes, so a push from
a frozen mirror silently reverts every entry the pod has moved on and REPORTS
SUCCESS. Measured 2026-09-02/03: of 25 local-only bullet candidates, 5 were
POD-NEWER, two of them `OPEN:` -> `RESOLVED` closures carrying ~20 lines of
later corrections.

So P3 now refuses with RC_CUTOVER_COMPLETE (19) instead. Measured on the live
store 2026-09-08: 147 entry files, 0 writable (P5 freeze applied, cutover DONE)
against 218 entries in the pod cache — so a push today would ship a strict
subset and revert the difference.

🔴 THE SIGNAL IS THE FREEZE, NOT A FLAG OR A DATE. A store whose entries all
refuse a write has completed the cutover; one still writable has not. So a
genuine first cutover passes straight through and P3 stays live for it — this
is a retirement conditioned on state, not a deletion.

Tests — the FIRST in this file to drive P3 through the real `main`. The prior
suite only re-extracted `seed.sh`'s `find` expression, so its 85 green tests
said nothing about whether this phase runs:
  - a FROZEN store returns 19 AND `seed.sh` is never invoked (asserting the rc
    alone would pass for a run that pushed and then refused)
  - a WRITABLE store is NOT refused — the over-refusal direction, labelled an
    INVARIANT GUARD because it passes at base by construction
  - the refusal names the remedy (`cairn append`/`put`) and warns off
    `--allow-overwrite`, because a refusal that does not say what to do instead
    gets routed around, and here the obvious route around IS the data-loss bug

Red-at-base: deleting the guard reds 2 of the 3; the third is the labelled
invariant guard and stays green, as it should. File 85 -> 88 tests.

⚠ No `examined == 0` branch at P3, deliberately: P0 already refuses an empty
store with RC_NO_STORE and `survey` walks the same `read_store` population, so a
zero is unreachable there — and a guard that cannot execute reads as coverage
while providing none. P5 keeps its own check because `--freeze` reaches it
without passing P0.

## Gating — read the tier split, it is not a clean green

Merged tree `77677af6` (base `origin/main` = `53c9f664`, unmoved at push).

- **sandbox `pytests`** — `RESULT: PASS (exit=0)`, 21332 collected / 21330 passed / **0 failed**
- **sandbox `nodetests`** — `RESULT: PASS (exit=0)`, 1449/1449
- **dev-host node** — `RESULT: PASS (exit=0)`, 1449/1449
- 🔴 **dev-host pytest — DID NOT COMPLETE.** `exit=124 (timeout after 3600s)`,
  `RESULT: FAIL (exit=143)` — SIGTERM at the wall clock, not an assertion. It was
  killed part-way through `scripts/browser-bridge/tests`.

**Why that is load and not this diff, by control rather than by theory:**

1. The killed target run STANDALONE on this exact tree: **915 passed, rc 0** in 422s.
2. The sandbox tier ran the whole suite on the same tree and passed.
3. This diff touches `scripts/cairn-cutover.py` and `scripts/tests/test_cairn_cutover.py`
   only — neither is reachable from browser-bridge.
4. Load average was **74–78 on 24 cores** throughout, all of it other sessions
   (checked for my own orphans: none).

⚠ **Not claimed:** a completed dev-host pytest run. One is owed when the box is
quieter. The repo's own rule is that a red above ~load 20 needs a control before
it means anything; the control above is what I have, and it is not the same
claim as a green full run.

---

### DIFF
--- a/scripts/cairn-cutover.py
+++ b/scripts/cairn-cutover.py
@@ -49,9 +49,14 @@ PHASES AND THEIR ROLLBACKS
                                  is modified. Roll back by deleting the run dir.
   P1  plan the delta           — read-only; writes only into the run directory
   P2  ref-collision check      — read-only
-  P3  push the delta           — rollback: `--rollback-push <run-dir>` re-PUTs the
-                                 pre-push bytes this script saved for every entry
-                                 it was about to overwrite
+  P3  push the delta           — 🔴 RETIRED once P5 has frozen the store: it
+                                 refuses with RC_CUTOVER_COMPLETE (19) rather
+                                 than pushing a frozen mirror back over a pod
+                                 that has moved on. Live only for a host that
+                                 has NOT been cut over. Rollback for a run that
+                                 did push: `--rollback-push <run-dir>` re-PUTs
+                                 the pre-push bytes this script saved for every
+                                 entry it was about to overwrite.
   P4  acceptance + byte check  — read-only
   P5  freeze local disk        — records every mode, then chmods. Rollback:
                                  `--unfreeze`, which RESTORES the recorded modes
@@ -142,6 +147,7 @@ RC_FREEZE_INEFFECTIVE = 16  # the freeze was applied and a write STILL succeeded
 # editing this sentence, and count.
 RC_ACCEPTANCE = 17
 RC_COULD_NOT_MEASURE = 18   # an instrument did not answer; never folded into a pass
+RC_CUTOVER_COMPLETE = 19    # P3 is RETIRED on this store — the freeze is already applied
 
 # 🔴 THE DISCRIMINATOR THIS WHOLE MERGE RULE TURNS ON. `server.render_bullet`
 # terminates every API-appended bullet with exactly this shape, and nothing else
@@ -1031,7 +1037,10 @@ def build_parser() -> argparse.ArgumentParser:
                         "divergence is refused rather than silently superseded")
     p.add_argument("--run-dir", type=Path, default=None)
     p.add_argument("--push", default=None, metavar="NS/DEPLOY",
-                   help="hand the delta tree to seed.sh --push")
+                   help="hand the delta tree to seed.sh --push. RETIRED on a "
+                        "store whose entries already refuse a write: the cutover "
+                        "has completed there and P3 exits 19 rather than reverting "
+                        "the pod. Use `cairn append`/`cairn put` instead.")
     p.add_argument("--dest", default="/data")
     p.add_argument("--alias-owner", action="append", default=[], metavar="S:A=FILE",
                    help="acknowledge one LIVE ref collision by naming its owner")
@@ -1353,7 +1362,53 @@ def main(argv: list[str] | None = None) -> int:
             say(f"P2 {len(owners)} live collision(s) acknowledged by --alias-owner: "
                 f"{sorted(f'{s}:{a}' for s, a in owners)}")
 
-        # ---- P3 the push --------------------------------------------------
+        # ---- P3 the push — RETIRED ONCE THE FREEZE IS APPLIED ---------------
+        # 🔴 P3 IS DEAD POST-CUTOVER, AND THE OBVIOUS "FIX" IS THE DATA-LOSS BUG.
+        # This phase pushes LOCAL -> POD. That is correct exactly once: while the
+        # host still holds bytes the pod has never seen. After P5 freezes local
+        # disk the direction inverts — this header's own ordering argument says
+        # so ("After this cutover the hosts' stores are caches of the pod") — and
+        # the pod accumulates API-appended content the mirror will never have.
+        #
+        # What it looked like instead: `plan.shippable` is ADD + SUPERSEDES +
+        # MERGED, and SUPERSEDES/MERGED are BY DEFINITION entries whose pod bytes
+        # differ, which is what `seed.sh`'s pre-flight refuses. So P3 exits 8 the
+        # moment anything supersedes, and that reads as a shipped code path that
+        # can never complete — one `--allow-overwrite` away from working.
+        #
+        # 🔴 IT IS NOT. That refusal is the LAST GUARD between a stale mirror and
+        # authoritative content, and `--allow-overwrite` is the single change that
+        # would disarm it. `seed.sh`'s tar "adds and overwrites but never
+        # deletes", so a push from a frozen mirror silently reverts every entry
+        # the pod has moved on — MEASURED 2026-09-02/03: of 25 local-only bullet
+        # candidates, 5 were POD-NEWER, two of them `OPEN:` -> `RESOLVED` closures
+        # carrying ~20 lines of later corrections. It would report success.
+        #
+        # So this refuses instead, by the same instrument P5 uses. The signal is
+        # the freeze, not a flag or a date: a store whose entries all refuse a
+        # write has completed the cutover, and a store that is still writable has
+        # not — so a genuine first cutover is untouched and passes straight
+        # through. To get bytes to the pod now, use the write route (`cairn
+        # append` / `cairn put`), which is what replaced this phase.
+        # ⚠ NO `examined == 0` BRANCH HERE, DELIBERATELY. `survey` walks the same
+        # `read_store` population P0 already refused as RC_NO_STORE, so a zero is
+        # unreachable by the time this runs — and a guard that cannot execute is
+        # worse than none, because it reads as coverage. P5 keeps its own check
+        # because `--freeze` reaches it without passing P0.
+        frozen = survey(args.store)
+        say(f"P3 local store: {frozen}")
+        if frozen["writable"] == 0:
+            return refuse(RC_CUTOVER_COMPLETE, (
+                f"P3 is RETIRED on this store — all {frozen['examined']} entry "
+                f"file(s) already refuse a write, so the cutover has completed "
+                f"and local disk is a read-through CACHE of the pod. Pushing it "
+                f"back would overwrite every entry the pod has moved on since, "
+                f"and `seed.sh` never deletes, so it would report success. "
+                f"🔴 Do NOT pass --allow-overwrite to get past this: that is the "
+                f"silent revert, not the fix. Write to the pod through `cairn "
+                f"append` / `cairn put`. NOTHING was pushed."
+            ))
+
         if not args.apply:
             say(f"DRY RUN — {len(plan.shippable)} entr(ies) would be pushed "
                 f"({counts[ADD]} new, {counts[SUPERSEDES]} superseding, "
--- a/scripts/cairn-cutover.py
+++ b/scripts/cairn-cutover.py
@@ -1452,20 +1452,41 @@ def main(argv: list[str] | None = None) -> int:
             ))
         if frozen["refused"] > 0:
             # 🔴 THE MESSAGE STATES WHAT WAS MEASURED, NOT A CONCLUSION IT CANNOT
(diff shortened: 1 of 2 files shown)


ZacxDev:
Round 1 audit complete; fixes pushed as `1b54f14e`. Claims block for the delta round.

```audit-claims round=1 audited=77677af6..1b54f14e
1. 🟡1+2 FIXED — the predicate changed from `writable == 0` to `refused > 0`. Claimed to fix BOTH directions: a post-freeze-created entry no longer disarms the guard (the freeze leaves scope dirs writable by design), and a never-cut-over store on a read-only mount no longer trips it (EROFS is not a PermissionError, so it lands in `other`, not `refused`).
2. 🟡3 FIXED — the refusal now names `cairn create` for the ADD case alongside `append`/`put`, and states that append/put both 404 on a ref that does not resolve.
3. 🟡4 FIXED — `set_entry_mode`'s docstring no longer claims the API has "NO create route" (false since `34d00d90`); what remains true — a writable scope dir is why `survey` can see a writable file on a cut-over store — is kept and tied to the new predicate.
4. 🟡5 FIXED — `assert "cairn" in text` was VACUOUS (P0 logs `cairn sync` every run) and a mutant survived it; the test now asserts rc 19 and pins `cairn create`, `cairn append`, `cairn put` separately.
5. 🟡6 FIXED — the control now passes the SAME argv as the frozen case (`--apply --push`) so the fixtures differ only in mode bits, and additionally asserts the un-frozen run REACHES seed.sh (`rc != 19` alone is satisfied by any earlier refusal).
6. 🟢7 FIXED — the dry-run behaviour change (bare invocation on a frozen store exits 19 instead of printing a plan) is now documented in USAGE and pinned by a test that also asserts it still changes nothing.
7. 🟢9 FIXED — "exit codes are disjoint" is now enforced; the `RC_CUTOVER_COMPLETE = 18` collision mutant that survived round 1 is dead.
8. Mutation claim — 5 mutants, control 91 either end: drop-`cairn create` KILLED (was SURVIVED), rc-collision KILLED (was SURVIVED), revert-to-`writable == 0` KILLED, always-fires KILLED, never-fires KILLED 4.
9. NOT FIXED, deliberately: finding 8 (a refused run still writes a `cairn sync` copy of the store first, because the guard sits after P0) and finding 10 (the handoff doc still lists this item as open — it is updated on merge, as its sibling item was). File 88 -> 91 tests.
10. ⚠ Correction carried: the previous commit cited "147 entry files" live; re-measured 159, all 0444. The count MOVES (hourly appends), so the load-bearing half is "0 writable", never the total.
```

ZacxDev:
Round 2 audit complete; fixes pushed as `10337671`.

`round 1 · payload 76` · `round 2 · payload 63` (cairn-cutover.py; the test file is scaffolding). Gate has not fired — no zero-payload round yet.

Gating: sandbox tier green on MERGED tree `8fbc6563` (base `a2b74e4b`) — `pytests` 21478/21476/**0 failed**, `nodetests` 1449/1449. ⚠ A completed full **dev-host** `gate.sh` run is still owed; the box has been at load 80–90 from other sessions all session and that tier hits its 3600s wall clock. Scoped dev-host runs over the cairn files pass.

```audit-claims round=2 audited=1b54f14e..10337671
1. F1 FIXED — `other > 0` now refuses with RC_COULD_NOT_MEASURE ABOVE the retirement, so the `other` bucket is no longer folded into a pass. Claimed to close the measured case where a cut-over store on a read-only mount (EROFS -> `other`) was PUSHED, and to honour the doctrine written on that constant.
2. F2 FIXED — the refusal states the measured COUNT instead of asserting "P5 has frozen it"; a MIXED store is told it may be a PARTIAL or interrupted freeze and pointed at `--freeze --apply`, which is now also in USAGE.
3. F3 FIXED — the USAGE sentence no longer claims a dry run "changes nothing"; it says what is and is not changed (a run dir holding a plaintext copy of the served store), and the test now asserts the run dir rather than globbing the store only.
4. F4 FIXED — `fake_run` intercepts `seed.sh`, so the hermetic target no longer shells out to `kubectl`; the `seen` ledger still proves reachability.
5. F5 FIXED — the stale paragraph above the guard (old predicate, `append`/`put`-only remedy) is rewritten.
6. F6 FIXED — the USAGE note is pinned on both claims it carries; deleting it now fails.
7. Mutation claim, control 94 either end: drop-the-`other`-branch KILLED (the round-1 SURVIVOR), drop-the-`--freeze`-route KILLED, weaken-the-USAGE-note KILLED. Fold-`other`-into-retirement SURVIVED and is reported as an EQUIVALENT mutant, not a gap — the `other` branch returns first so the added `or` clause is unreachable; the `other` branch itself is reachable (deleting it kills a test).
8. NOT FIXED, deliberately: the handoff doc still lists this item as open (updated on merge, as its sibling was). File 91 -> 94 tests.
9. ⚠ Evidence scope: F1's unreadable state is injected by stubbing `probe_writable`, not by mounting a real read-only filesystem (needs root). The errno mapping underneath is asserted in the test rather than assumed.
```

ZacxDev:
⚠ **Correction to the ledger line in my previous comment.** It says `round 2 · payload 63`. The measured figure is **123** (79+/44− in `scripts/cairn-cutover.py`; the 120+/2− in the test file is scaffolding):

```
$ git log --numstat --format= --remerge-diff 1b54f14e..10337671 --not origin/main
79      44      scripts/cairn-cutover.py
120     2       scripts/tests/test_cairn_cutover.py
```

I wrote the number before running the command in the same breath as running it. It changes no conclusion — both figures are far from zero, so the attribution gate does not fire either way and the ladder continues — but a count in prose is a claim, and that one was invented.

ZacxDev:
Round 3 audit complete; fixes pushed as `9ac728e8`.

`round 1 · payload 76` · `round 2 · payload 123` · `round 3 · payload 79` (cairn-cutover.py; the test file is scaffolding). Gate has not fired.

Gating: sandbox tier green on MERGED tree `9205587d` (base `8a9ebba6`) — `pytests` 21491/21489/**0 failed**, `nodetests` 1449/1449. A full dev-host `gate.sh` run is now in flight (load finally dropped from ~90 to 63); it has been owed since round 1.

```audit-claims round=3 audited=10337671..9ac728e8
1. 🟡3 FIXED — the MIXED-state advice now carries `--unfreeze --mode-ledger <the INTERRUPTED run>/…`, because `--freeze --apply` writes a SECOND ledger recording 0444 and `--unfreeze` takes the newest, so a 0600 entry was "restored" to 0444 with a success message.
2. 🟡2 FIXED — the false absolute ("which only P5's freeze produces here") is gone; the message now names all three states a `refused > 0` store can be in, and puts `cairn create` BEFORE the freeze route with "pushes NOTHING" stated, so following it in order cannot strand local-only content behind a retired P3.
3. 🟡1 FIXED — the RUNTIME dry-run message on the ordinary path no longer says a flat "Nothing was changed."; it says "IN EITHER STORE", names the run dir P0 wrote, and says to delete it.
4. 🟢4 FIXED — the ordering (`other` decided before the retirement) is pinned by a store no other test builds: partly frozen AND partly unreadable. The F1 test stubs `probe_writable` for every entry so it cannot construct that bucket.
5. Mutation claim, control 97 either end: revert-dry-run-message KILLED, drop-ledger-caveat KILLED, drop-"pushes NOTHING" KILLED, move-`other`-below-the-guard KILLED (SURVIVED last round).
6. NOT CLAIMED: no test drives the recovery END TO END (freeze -> unfreeze -> compare modes). The harm was reproduced by the round-3 auditor; what is pinned here is that the message says the right thing. File 94 -> 97 tests.
7. NOT FIXED, deliberately: the handoff doc still lists this item as open (updated on merge, as its sibling was).
```

ZacxDev:
✅ **The owed dev-host run is DONE — all four tiers now green on one merged tree.**

Merged tree `9205587d` (base `origin/main` = `8a9ebba6`):

| tier | verdict | totals |
|---|---|---|
| dev-host `pytest` | `RESULT: PASS (exit=0)` | 21491 collected / 21489 passed / **0 failed** (floor 20441) |
| dev-host `node` | `RESULT: PASS (exit=0)` | 1449/1449 (floor 1367) |
| sandbox `pytests` | `RESULT: PASS (exit=0)` | 21491 / 21489 / **0 failed** |
| sandbox `nodetests` | `RESULT: PASS (exit=0)` | 1449/1449 |

`gate.sh` verdict `GATE: RESULT=PASS exit=0`, and both runners' own `RESULT:` lines agree with it — no exit-90 could-not-vouch, no `panic: test timed out`.

This retires the caveat carried on every previous comment. The dev-host tier had been hitting its 3600s wall clock since round 1 because the box sat at load 80–90 from other sessions all session; it dropped to ~63 and the run completed. Nothing about the change moved — what changed is that the measurement finally finished, which is a different claim from the earlier standalone-target control and is why it was tracked as owed rather than waved through.

ZacxDev:
**Final gate, merged tree `65fccebb` (base `4a67ea73`) — and it is NOT a clean four-tier green. The honest accounting:**

| tier | tree | verdict |
|---|---|---|
| sandbox `pytests` | **this** | `RESULT: PASS (exit=0)` — 21522 / 21520 / **0 failed** |
| sandbox `nodetests` | **this** | `RESULT: PASS (exit=0)` — 1449/1449 |
| dev-host `node` | **this** | `RESULT: PASS (exit=0)` — 1449/1449 |
| dev-host `pytest` | **this** | 🔴 **DID NOT COMPLETE** — `exit=124 (timeout after 3600s)`, `RESULT: FAIL (exit=143)` = SIGTERM at the wall clock |

**Why that is load, established by control rather than asserted:**

1. It was killed in `scripts/claude-hooks/tests/test_gh_issue_closing_condition_guard.py` — a target this PR does not touch.
2. That target **plus** `test_cairn_cutover.py`, run standalone on this exact tree: **573 passed, rc 0, in 78s**.
3. The sandbox tier ran the whole suite on this tree and passed.
4. A full dev-host `pytest` run **did** complete on the immediately-prior tree `9205587d` (`RESULT: PASS`, 21491/21489/0). The entire delta since then is **28 lines in one test file** — `test_cairn_cutover.py`, the file the control above exercises directly.
5. Load was 65–75 on 24 cores throughout, from other sessions.

⚠ **Not claimed:** a completed full dev-host `pytest` run on THIS tree. Points 1–4 are what I have, and they are a different claim from a green full run. Merging on the sandbox tier (the one a merge gates on), the completed prior-tree run, and the standalone control.

PR #1429, opened 2026-09-09
perf(gate): size xdist workers from the cgroup quota, and re-enter the dev shell instead of printing the command

ZacxDev:
Two changes to how the pytest tier runs, plus a rewrite of the merge-gate policy that used to sit on top of them.

> 🔴 **RE-SCOPED. The slot limiter that was a third of this PR is DELETED, not fixed.**
>
> The operator applied the algorithm to the process this PR was optimizing. **The requirement that every session run the full suite locally before merging is removed from `CLAUDE.md`** — branch protection stays off (a knowing trade of safety for speed), CI stays advisory, and the inner loop becomes a change-scoped run (a separate PR).
>
> A limiter that bounds concurrent full-suite gate runs is a step-3/4 optimization of a step the first two questions say should not exist. So it is gone: `acquire_slot`, `DEVRC_GATE_SLOTS`/`_SLOT_WAIT`/`_SLOT_DIR`/`_SLOT_POOL`, the `slot:` line in the GATE report, the fd-close subshell in `run_tier` (it existed only to protect the slot lock), and all of their tests and docs. **The concurrency table that justified it is deleted, not reworded.** For the record, since this PR quoted it as measured fact: bucketing 237 runs by how many others overlapped them is length-biased — overlap count is partly a *function* of the run's own duration — and a null Monte Carlo with zero interaction between runs reproduced the shape and about two thirds of the effect. Nothing left in the tree needs the number, so it is gone rather than corrected.

## 1. Worker sizing from the cgroup quota (`run-tests.sh`)

`min(nproc, 4)` read as "adapt to the machine". It wasn't: the constant 4 did all the work and merely *happened* to equal the `devrc-ci` pod's CPU limit. **`nproc` reports the node's cores, not the container's**, so every formula built on it oversubscribes inside a pod by that factor.

The budget is now the narrowest cgroup v2 quota, falling back to `nproc`, capped at 8 — still 4 in that pod, now for the right reason, and 8 on a 24-core box where the flat 4 left 20 cores idle.

Two walk bugs found in review, each fixed with a test that dies on its own mutant:

- a cgroup path of `/` (the k8s cgroupns-private shape) **escaped the seamed root**, because `${root}/` never string-equals `$root` and `dirname` then returns the root's parent;
- `cut -d: -f3 | head -1` took a **cgroup v1** line on a hybrid v1+v2 host, and truncated any cgroup path containing a `:`. Now `sed -n 's/^0:://p'`.

## 2. `gate.sh` re-enters `nix develop` itself

`run-tests.sh` refuses to run without its REQUIRED_TOOLS and prints the exact `nix develop …` line that fixes it. The message was never wrong; making a correct instruction be re-typed by hand is the defect.

⚠ **The "100 of 100" this PR originally claimed was a population defined by the failure's own cause.** `LOG_DIR` defaults to `mktemp -d`, so a run *inside* `nix develop` lands under nix's per-shell `TMPDIR` and one *outside* lands in bare `/tmp` — globbing `/tmp/devrc-gate-*` selects exactly the runs launched outside the dev shell, i.e. the failing kind, by construction. Re-measured across both locations:

| population | dirs | died on missing-tools FATAL |
|---|---:|---:|
| bare `/tmp` (outside `nix develop`) | 101 | 101 (100%) |
| nix-shell `TMPDIR` (inside) | 281 | 0 (0%) |
| **all** | **382** | **101 (26%)** |

**101 of 382 (26%)**, not 100 of 100 — still roughly one gate invocation in four thrown away. The sub-claim "the node tier ran anyway in all 100" is corrected to **91 of 101**.

Three guards keep it honest — `DEVRC_GATE_NO_REEXEC`, `DEVRC_GATE_ENV`, and a `DEVRC_GATE_REEXEC` loop guard whose failure mode the script itself calls an unkillable fork bomb — plus a blanket `PYTEST_CURRENT_TEST` suppression so a nested run can never re-enter the shell. **All four are now individually reachable and each dies to its own mutant**; before this round a `GATE_SEAMED` short-circuit stood in front of them and all three could be deleted with the suite still green.

## 3. `CLAUDE.md`: the gate policy now says what is true

Replaces "run BOTH tiers yourself before merging":

- **the local full-suite ritual is deleted, with its reason** — it produced 27–50 concurrent full-suite runs on one 24-core box, the same ~22,000 tests twice per change, while gating nothing, and contended badly enough that the dev-host tier repeatedly hit its own 3600s cap and produced *no verdict at all*;
- **CI is advisory** — the only automated signal *on a PR*, and read as such: `gh pr checks <n>`; merging through a red one is a decision you own and should say out loud;
- **but a bad merge is DETECTED, not undetected** — `scripts/main-green-check.sh` is a live 4-hourly deadman (`main-green-check.timer` active, `OnUnitActiveSec=4h`, verified 2026-09-09) that re-runs **both nix sandbox tiers** against `origin/main`'s tip, **reproduces a red before alerting**, and escalates through the DND-defeating toast class. ⚠ A round-1 draft of this bullet claimed CI was "the only automated signal that exists" and that the backstop was "one advisory check plus whoever is reading the PR" — **false, and it understated the net by a whole subsystem**; round 2 caught it;
- **the CI noise rate ships with it** — re-derived 2026-09-09 over the 60 most-recently-updated PR heads (newest `tekton/devrc-pytests` status per head): **21 of 50 terminal verdicts not success (42%)**, plus **10 of 60 heads (17%) stuck `pending`**. So act on the failing test's name, not on the colour;
- **this is less safe, deliberately** — prevention was traded for **detection**, and the doc says that rather than implying the arrangement has been made safe.

## Audit findings — status

| finding | status |
|---|---|
| 🔴-1 gating tier RED — tests derived expectations from `os.cpu_count()` while the code floors on `nproc` (16 vs 4 on the CI builder) | **FIXED.** Every expectation and both skip guards now come from `_nproc()`, a subprocess mirroring the code |
| 🟡-2 the limiter could change a verdict | **DELETED WITH THE LIMITER** |
| 🟡-3 all three re-exec guards unreachable; mutants survived a green run | **FIXED.** The `GATE_SEAMED` short-circuit is deleted; each guard's test pops `PYTEST_CURRENT_TEST` and sets only its own variable; a new test pins the `GATE_NESTED` suppression they all now bypass |
| 🟡-4 relative-path invocation died `rc=127` with no verdict | **FIXED.** `GATE_SELF` is resolved before any `cd`; the test drives gate.sh from `<repo>/scripts` through a fake `nix` that actually `exec`s, so the failure reproduces end to end |
| 🟡-5 the 3.4x concurrency table | **DELETED, not restated** |
| 🟡-6 "100 of 100" | **RESTATED** as 101 of 382 (26%), with the selection bias named, in both surviving places |
| 🟡-7 `--help` guard narrower than its name | **FIXED.** The expected set is derived from `gate.sh`'s source, with positive and negative controls |
| 🟡-8 stale `-n 4` comments | **FIXED** in `run-tests.sh`, `testlib/store_siting.py`, `testlib/nolaunch.py`. The `/dev/shm` budget is **verified still correct at 8**: 8 × 1.88 MB ≈ 15 MB against a 62 GB `/dev/shm`, no constant changed |
| 🟡-9 ~11 un-serialised nested full runs added to the gating tier | **FIXED**, and it matters more now that CI is the only signal. New `DEVRC_TEST_BUDGET_ONLY` seam stops the runner right after the banner these tests read: **~150s → ~3s per case**. It exits **3**, never 0, so an ambient value reds a real gate loudly instead of emptying it |
| 🟢 cgroup-walk nits | **FIXED**, both with tests |
| 🟢 slot-related nits | **DELETED WITH THE LIMITER** |
| second-order question about `DEVRC_GATE_SLOTS=2` | **MOOT** |

## Two things this round found that the audit did not

- **Five host-conditional skips would have red-ed the gate at `failed=0`.** `run-tests.sh` GUARD 2 pins the expected-skip *set*, and each skip was `if nproc <= N`, so all five fired on the gating builder — where `nproc` is 4 — leaving the cgroup branch untested exactly where it is gated. Every "the quota wins" case now uses a quota of **1**, which is below `nproc` on any multi-core host, so the same assertion is exact and discriminating everywhere. **25 passed / 0 skipped at nproc 4, 8 and 24.**
- **One test was vacuous and the sweep caught it.** `test_a_quota_wider_than_nproc_does_not_widen_the_budget` asserted `min(nproc, 8)`; on a 24-core box that is `8` for *every* quota above 8, so a mutant letting the quota win outright was capped to 8 and produced the identical answer. It now narrows the child's CPU set with `taskset` so the wrong answer sits strictly between the right one and the cap, with a positive control for the narrowing.

## Verification

- **Mutation sweep**, one mutation at a time, whole file each time: 13 mutants, **12 killed by their own test**, 1 SURVIVED and labelled. Table in the `audit-claims` comment.
- **Measured at three points on the dimension that broke the blocker** — `nproc` 4, 8 and 24 via `taskset` — because the blocker *was* a single-point measurement.
- **CI is the arbiter, not a local run** — a green `nix build …#pytests` on the dev box does not close 🔴-1, since it was green here while `tekton/devrc-pytests` was red on the same commit, twice. **Both checks are now GREEN on `8fc65c0b`:** `collected=21535 passed=21533 skipped=2 failed=0` (pytests) and `tests=1449 pass=1449 fail=0` (nodetests).
- ⚠ **The previous head's single unrelated failure was investigated, not waved through.** The mechanism I most suspected — the 4→8 cap doubling the sandbox tier's parallelism — is **measured absent**: the CI log's own banner reads `parallelism =4 pytest worker(s)`, because the builder's `nproc` is 4 and `min(nproc, quota, 8)` is unchanged there. Across the 13 retained `devrc-ci` runs with a verdict, 8 failed and 6 of those on a *single, different* test each — 5 distinct tests on 6 branches, with `origin/main`'s own run clean. The control: re-pushing with my 8 files **byte-identical** went green. A green here is evidence this PR does not break the suite; it is **not** evidence the tier is healthy.

https://claude.ai/code/session_019zKmAWntvACVnzP6kbxW9b


ZacxDev:
## Round 1 fixes — what was done

Full detail in the PR body and in commit `a8d2aa01`. Highlights a re-audit should check hardest:

- **The blocker's fix is verified at three points on the dimension that broke it** (`nproc` = 4, 8, 24 via `taskset`), because the bug *was* a single-point measurement. A green `nix build …#pytests` on the dev host is explicitly NOT treated as evidence here — it was green while `tekton/devrc-pytests` was red on the same commit.
- **Mutation table** — every guard touched this round, one mutation at a time, whole file run each time:

| # | mutation | result |
|---|---|---|
| — | baseline, gate tests | 20 passed |
| — | baseline, jobs tests | 16 passed |
| M1 | drop `DEVRC_GATE_NO_REEXEC` guard | **KILLED** by `test_no_reexec_opt_out_is_honoured` (+ the scrub-ledger test, which fails when gate.sh's var set shrinks) |
| M2 | drop `DEVRC_GATE_ENV` guard | **KILLED** by `test_it_does_not_re_exec_when_already_in_a_gate_environment` (+ ledger) |
| M3 | drop `DEVRC_GATE_REEXEC` loop guard | **KILLED** by `test_the_loop_guard_stops_a_second_re_exec` (+ ledger; also reds the relative-path test, which is the fork-bomb symptom showing up) |
| M4 | re-exec hands the raw `${BASH_SOURCE[0]}` through | **KILLED** by `test_a_relative_path_invocation_survives_the_re_exec`, alone |
| M5 | drop the slot-fd close in `run_tier` (positive control) | **KILLED** by `test_an_orphan_spawned_by_the_tier_does_not_keep_holding_the_slot`, alone |
| M6 | drop the ancestor-pool inertness check | **KILLED** by `test_a_descendant_does_not_queue_behind_its_own_ancestors_slot` (+ ledger) |
| M7 | do not export `DEVRC_GATE_SLOT_POOL` | **KILLED** by `test_a_gate_run_names_the_pool_it_holds_for_its_descendants`, alone |
| M8 | treat an unopenable pool as contention | **KILLED** by `test_an_unopenable_pool_is_not_reported_as_contention`, alone |
| M9 | drop the trailing-slash normalisation | **KILLED** by `test_a_root_level_cgroup_path_does_not_escape_the_root`, alone |
| M10 | back to `cut -d: -f3 \| head -1` | **KILLED** by `test_a_hybrid_v1_v2_proc_self_cgroup_reads_the_unified_line` + `test_a_cgroup_path_containing_a_colon_is_not_truncated` |
| M11 | drop the `q=""; p=""` reset in the walk | ⚠ **SURVIVED** — see below |
| M12 | drop the walk terminator (`break` at the root) | **KILLED** by `test_the_walk_stops_at_the_root_it_was_given` (+ the `0::/` test). ⚠ On the first sweep it was killed by the *wrong* test — the named one could not see the escape because nothing sat above its fake root. Fixture strengthened, re-swept |
| M13 | cap back at 4 (control) | **KILLED** by `test_an_unquota_d_host_gets_more_than_the_old_hardcoded_four` + 5 others |
| M14 | `DEVRC_TEST_BUDGET_ONLY` seam exits 0 instead of 3 | **KILLED** by `test_the_budget_only_seam_runs_no_tests_and_cannot_report_a_pass`, alone |

- **M11 SURVIVED and is reported rather than hidden.** Bash's `read` assigns its variables even when it returns non-zero at EOF, and the walk returns on the first numeric quota, so no fixture reachable through the `DEVRC_TEST_CGROUP_*` seam can make the reset matter. It is kept as an **invariant guard**; the comment in `run-tests.sh` and the test docstring both now say so — the previous comment asserted it was load-bearing.
- **File-level red/green:** the round-1 gate tests run against `scripts/gate.sh` as it was at `d640356a`: **9 failed, 11 passed**. At HEAD: **20 passed**. The three re-exec-guard tests pass in *both* directions, which is the finding — they were vacuous, and the mutation table above is the evidence that they are not any more.
- **Deliberately left open:** (a) the two withdrawn measurements are corrected in the PR body, `CLAUDE.md`, `gate.sh` and the test docstring, but the *original commit message* (`eead6811`) still carries them — rewriting it would move the sha the next round anchors its delta on. (b) `DEVRC_GATE_SLOT_WAIT` 3600s→600s is a judgement, argued in the source and the PR body, not a measurement.

```audit-claims round=1 audited=d640356a8d7949beb9b6bfb4c5d6b9154f5df8f5..a8d2aa01
1. BLOCKER: test_run_tests_jobs.py derived expectations from os.cpu_count() while run-tests.sh floors on nproc — every expectation and both skip guards now come from a `_nproc()` subprocess; verified green at nproc 4, 8 and 24 via taskset.
2. The two tests that assert an exact worker count against the real machine are now seamed to a no-quota fake cgroup, so an unknown CI-builder quota cannot move them; one new test keeps the unseamed path exercised as a bound (1 <= budget <= min(nproc, 8)).
3. gate.sh now enforces one slot per gate TREE — it exports DEVRC_GATE_SLOT_POOL and a descendant that would join the same pool goes inert instead of queueing behind its own ancestor.
4. Both gate test files scrub the operator-facing DEVRC_GATE_* variables out of the inherited environment; the scrub list is pinned two-way against gate.sh's own source.
5. The "FAIL-OPEN, ALWAYS … nothing here can change the verdict" comment is replaced with one that names the exit-2 exception and the DELAY the limiter can still cause.
6. The GATE_SEAMED short-circuit in front of the three re-exec guards is deleted, and each guard's test now pops PYTEST_CURRENT_TEST and sets only its own variable; all three mutants now die, each by its own test.
7. gate.sh resolves its own path into GATE_SELF before any cd, so `cd <repo>/scripts && bash gate.sh` no longer dies rc=127; the test drives it from that directory through a fake nix that actually execs.
8. The --help env-var test derives its expected set from gate.sh's source instead of four literals, with its own positive and negative controls; all 11 vars are documented.
9. The 3.4x concurrency claim is RETRACTED as length-biased sampling in all four places it appeared (PR body, CLAUDE.md, gate.sh, test docstring); only the unbiased count (35 of 117 reds died on SIGTERM at the 3600s cap) is kept.
10. The "100 of 100" re-exec claim is RETRACTED and restated as 101 of 382 (26%), with the node-tier sub-claim corrected from 100/100 to 91/101, in the same four places.
11. Stale `-n 4` modelling corrected in run-tests.sh, testlib/store_siting.py and testlib/nolaunch.py; the /dev/shm budget is verified still correct at 8 workers (~15 MB against 62 GB) with no constant changed.
12. A new DEVRC_TEST_BUDGET_ONLY seam stops the runner right after the parallelism banner, cutting the file's ~11 nested full runs from ~150s each to ~3s; it exits 3, never 0, and a test pins that it cannot report a pass.
13. NITS: the cgroup walk no longer escapes its root on a `0::/` path; `sed -n 's/^0:://p'` replaces `cut -d: -f3 | head -1` (v1-line and colon-in-path bugs); a failed lock-file open is reported as a pool problem rather than contention; `slot: NONE HELD` now names which of five states it is; CLAUDE.md's "per box" is corrected to per (uid, slot pool) with all four exclusions stated.
14. DEVRC_GATE_SLOT_WAIT default changed 3600s -> 600s and the queue now prints a 60s heartbeat; the second-order question (are the defaults right, should the nix tier take a slot) is answered in the PR body.
15. One mutant SURVIVED and is labelled rather than fixed: deleting the `q=""; p=""` reset in the cgroup walk kills no test, and it is now documented as an invariant guard in both the code comment and the test docstring.
```

ZacxDev:
## Round 1 fixes — SUPERSEDES the earlier claims comment on this PR

🔴 **Read this block, not the one above it.** Between the two, the operator re-scoped the PR: the slot limiter is **deleted rather than fixed**, along with the concurrency measurement that justified it and every finding that only existed because of it. The earlier comment described a tree that no longer exists.

### Mutation table — reduced scope, one mutation at a time, whole file run each time

| # | mutation | result |
|---|---|---|
| — | baseline, `test_gate_reexec.py` | 9 passed |
| — | baseline, `test_run_tests_jobs.py` | 17 passed |
| M1 | drop `DEVRC_GATE_NO_REEXEC` guard | **KILLED** by `test_no_reexec_opt_out_is_honoured` (+ the scrub/derivation ledger tests, which fail when gate.sh's var set shrinks) |
| M2 | drop `DEVRC_GATE_ENV` guard | **KILLED** by `test_it_does_not_re_exec_when_already_in_a_gate_environment` (+ ledger) |
| M3 | drop `DEVRC_GATE_REEXEC` loop guard | **KILLED** by `test_the_loop_guard_stops_a_second_re_exec` (+ ledger; also reds the relative-path test, which is the fork-bomb symptom surfacing) |
| M3b | drop the `GATE_NESTED` short-circuit | **KILLED** by `test_the_limiter_free_gate_does_not_re_exec_inside_pytest`, alone |
| M4 | re-exec hands the raw `${BASH_SOURCE[0]}` through | **KILLED** by `test_a_relative_path_invocation_survives_the_re_exec`, alone |
| M5 | `--help` back to the hardcoded `sed 2,70p` | **KILLED** by `test_the_help_text_documents_every_env_var_the_script_reads`, alone |
| M9 | drop the trailing-slash normalisation in the cgroup walk | **KILLED** by `test_a_root_level_cgroup_path_does_not_escape_the_root`, alone |
| M10 | back to `cut -d: -f3 \| head -1` | **KILLED** by `test_a_hybrid_v1_v2_proc_self_cgroup_reads_the_unified_line` + `test_a_cgroup_path_containing_a_colon_is_not_truncated` |
| M11 | drop the `q=""; p=""` reset in the walk | ⚠ **SURVIVED** — labelled, see below |
| M12 | drop the walk terminator (`break` at the root) | **KILLED** by `test_the_walk_stops_at_the_root_it_was_given` (+ the `0::/` test). ⚠ On the first sweep it was killed by the *wrong* test; fixture strengthened, re-swept |
| M13 | cap back at 4 (control) | **KILLED** by `test_an_unquota_d_host_gets_more_than_the_old_hardcoded_four` + 4 others |
| M14 | `DEVRC_TEST_BUDGET_ONLY` seam exits 0 instead of 3 | **KILLED** by `test_the_budget_only_seam_runs_no_tests_and_cannot_report_a_pass`, alone |
| M15 | let the cgroup quota WIDEN past `nproc` | ⚠ **SURVIVED at first**, then **KILLED** by `test_a_quota_wider_than_nproc_does_not_widen_the_budget` after the fixture was fixed — see below |

### The two mutants worth reading in full

**M15 — a fixture that landed on its own boundary, found by the sweep and fixed.** The test asserted `min(nproc, 8)`. On a 24-core box that is `8` for *every* quota above 8, so a mutant that let the quota win outright was still capped to 8 and produced the identical answer: fully green suite, mutant survived. The test now narrows the **child's** affinity mask to 2 CPUs and uses a quota of 5, so the wrong answer (5) sits strictly between the right one (2) and the cap (8) — plus `test_the_narrowed_cpu_seam_actually_narrows_nproc` as a positive control that fails if the narrowing ever stops taking effect.

**M11 — still SURVIVES, and is labelled rather than papered over.** Bash's `read` assigns its variables even when it returns non-zero at EOF, and the walk returns on the first numeric quota, so no fixture reachable through the `DEVRC_TEST_CGROUP_*` seam can make the reset matter. It stays as an **invariant guard**; the comment in `run-tests.sh` and the test docstring both say so — the previous comment asserted it was load-bearing.

### Red/green

- `test_run_tests_jobs.py` + `test_gate_reexec.py` measured at **three points on the dimension that broke the blocker**: `nproc` 4 → 21 passed / 5 skipped, `nproc` 8 → 25 passed / 1 skipped, `nproc` 24 → 26 passed / 0 skipped (`taskset`).
- `test_gate_exit_truthfulness.py` + `test_ci_claim_matches_reality.py`: 18 passed.
- **The blocker is closed by CI, not by a local run.** On head `a8d2aa01`, `tekton/devrc-pytests` reported `collected=21365 passed=21357 skipped=7 failed=1` — the five `test_run_tests_jobs.py` failures are **gone**. The one remaining failure is `test_mjs_parses[attachments.mjs]`, in `claude/skills/clickup/api/` — a file this branch does not touch, last modified in #438, and which `node --check` parses clean locally. Watching the re-scoped head for whether it recurs.

```audit-claims round=1 audited=d640356a8d7949beb9b6bfb4c5d6b9154f5df8f5..6597318a
1. RE-SCOPE: the slot limiter is DELETED, not fixed — acquire_slot, DEVRC_GATE_SLOTS/_SLOT_WAIT/_SLOT_DIR/_SLOT_POOL, GATE_SLOT_HELD/_FD/_STATE, the SLOT LIMITER comment block, the `slot:` report line, the run_tier fd-close subshell (it only ever protected the slot lock), and the whole test file are gone.
2. The 3.4x concurrency table is DELETED rather than restated, in every place it appeared (PR body, CLAUDE.md, gate.sh, test docstring); it existed only to justify the limiter.
3. BLOCKER: test_run_tests_jobs.py derived expectations from os.cpu_count() while run-tests.sh floors on nproc — every expectation and both skip guards now come from a `_nproc()` subprocess; measured green at nproc 4, 8 and 24 via taskset.
4. The two tests that assert an exact worker count against the real machine are seamed to a no-quota fake cgroup so an unknown CI-builder quota cannot move them; one new test keeps the unseamed path exercised as a bound (1 <= budget <= min(nproc, 8)).
5. The GATE_SEAMED short-circuit in front of the three re-exec guards is deleted, and each guard's test now pops PYTEST_CURRENT_TEST and sets only its own variable; all three mutants now die by their own tests, and a new test pins the GATE_NESTED suppression they bypass.
6. gate.sh resolves its own path into GATE_SELF before any cd, so `cd <repo>/scripts && bash gate.sh` no longer dies rc=127; the test drives it from that directory through a fake nix that actually execs.
7. The --help env-var test derives its expected set from gate.sh's source instead of four literals, with positive and negative controls; all six remaining vars are documented.
8. The "100 of 100" re-exec claim is restated as 101 of 382 (26%) with the selection bias named, and the node-tier sub-claim corrected from 100/100 to 91/101.
9. Stale `-n 4` modelling corrected in run-tests.sh, testlib/store_siting.py and testlib/nolaunch.py; the /dev/shm budget is verified still correct at 8 workers (~15 MB against 62 GB) with no constant changed.
10. A new DEVRC_TEST_BUDGET_ONLY seam stops the runner right after the parallelism banner, cutting the file's ~11 nested full runs from ~150s each to ~3s; it exits 3, never 0, and a test pins that it cannot report a pass.
11. NITS: the cgroup walk no longer escapes its root on a `0::/` path, and `sed -n 's/^0:://p'` replaces `cut -d: -f3 | head -1` (v1-line and colon-in-path bugs). Slot-related nits are deleted with the limiter.
12. CLAUDE.md: the "run BOTH tiers yourself before merging" mandate is DELETED with its reason stated; CI is named as advisory and as the only automated signal; and the bullet says plainly that the new arrangement is LESS SAFE and deliberately so.
13. test_a_quota_wider_than_nproc_does_not_widen_the_budget was found VACUOUS by the sweep (the 8-cap swallowed the mutant on a 24-core box) and now narrows the child's CPU affinity so the wrong answer sits between the right one and the cap, with a positive control for the narrowing.
14. The CPU narrowing in claim 13 is done with `taskset`, not subprocess `preexec_fn`: this suite runs under pytest-xdist and forking a threaded process to run Python before exec is a deadlock risk, which would turn a red test into a hang.
15. One mutant SURVIVES and is labelled rather than fixed: deleting the `q=""; p=""` reset in the cgroup walk kills no test, and it is documented as an invariant guard in both the code comment and the test docstring.
```

ZacxDev:
## Round 1 fixes — SUPERSEDES both earlier claims comments on this PR

🔴 **Read this block, not the two above it.** Between them the operator re-scoped the PR (the slot limiter is **deleted**, not fixed) and the branch was rebased onto `origin/main`. The earlier comments describe trees that no longer exist.

⚠ **Range note for the next round.** `<from>` is `d640356a`, the tip the round-1 audit actually read. The branch has since been **rebased**, so that sha is no longer an ancestor of the head — `git diff d640356a..8fc65c0b` therefore also contains the 7 upstream commits the rebase pulled in. The PR's own change is exactly these 8 files: `CLAUDE.md`, `scripts/gate.sh`, `scripts/run-tests.sh`, `scripts/testlib/nolaunch.py`, `scripts/testlib/store_siting.py`, `scripts/tests/test_gate_exit_truthfulness.py`, `scripts/tests/test_gate_reexec.py`, `scripts/tests/test_run_tests_jobs.py`. `git diff origin/main...HEAD` is the clean view.

### Mutation table — one mutation at a time, whole file run each time, after the fixture rewrite

| # | mutation | result |
|---|---|---|
| — | baseline, `test_gate_reexec.py` | 9 passed |
| — | baseline, `test_run_tests_jobs.py` | 16 passed |
| M1 | drop `DEVRC_GATE_NO_REEXEC` guard | **KILLED** by `test_no_reexec_opt_out_is_honoured` (+ the two ledger tests, which fail when gate.sh's var set shrinks) |
| M2 | drop `DEVRC_GATE_ENV` guard | **KILLED** by `test_it_does_not_re_exec_when_already_in_a_gate_environment` (+ ledger) |
| M3 | drop `DEVRC_GATE_REEXEC` loop guard | **KILLED** by `test_the_loop_guard_stops_a_second_re_exec` (+ ledger; also reds the relative-path test, which is the fork-bomb symptom surfacing) |
| M3b | drop the `GATE_NESTED` short-circuit | **KILLED** by `test_the_limiter_free_gate_does_not_re_exec_inside_pytest`, alone |
| M4 | re-exec hands the raw `${BASH_SOURCE[0]}` through | **KILLED** by `test_a_relative_path_invocation_survives_the_re_exec`, alone |
| M5 | `--help` back to the hardcoded `sed 2,70p` | **KILLED** by `test_the_help_text_documents_every_env_var_the_script_reads`, alone |
| M9 | drop the trailing-slash normalisation in the cgroup walk | **KILLED** by `test_a_root_level_cgroup_path_does_not_escape_the_root`, alone |
| M10 | back to `cut -d: -f3 \| head -1` | **KILLED** by `test_a_hybrid_v1_v2_proc_self_cgroup_reads_the_unified_line` + `test_a_cgroup_path_containing_a_colon_is_not_truncated` |
| M11 | drop the `q=""; p=""` reset in the walk | ⚠ **SURVIVED** — labelled, see below |
| M12 | drop the walk terminator (`break` at the root) | **KILLED** by `test_the_walk_stops_at_the_root_it_was_given` (+ the `0::/` test) |
| M13 | cap back at 4 (control) | **KILLED** by `test_an_unquota_d_host_gets_nproc_capped_at_eight` + 2 others |
| M14 | `DEVRC_TEST_BUDGET_ONLY` seam exits 0 instead of 3 | **KILLED** by `test_the_budget_only_seam_runs_no_tests_and_cannot_report_a_pass`, alone |
| M15 | let the cgroup quota WIDEN past `nproc` | **KILLED** by `test_a_quota_wider_than_nproc_does_not_widen_the_budget`, alone |

Two entries in that table changed verdict during this round and are worth reading:

- **M12** was first killed by the *wrong* test — the one whose docstring claimed to cover the escape could not see it, because nothing sat above its fake root. Fixture strengthened, re-swept.
- **M15 first SURVIVED.** The test asserted `min(nproc, 8)`; on a 24-core box that is `8` for *every* quota above 8, so the cap swallowed the mutant — a fixture landing exactly on its own boundary. It now narrows the child's CPU set with `taskset` and uses a quota of 5, so the wrong answer (5) sits strictly between the right one (2) and the cap (8), with `test_the_narrowed_cpu_seam_actually_narrows_nproc` as the positive control.
- **M11 still SURVIVES and is labelled rather than papered over.** Bash's `read` assigns its variables even at EOF and the walk returns on the first numeric quota, so no fixture reachable through the `DEVRC_TEST_CGROUP_*` seam can make the reset matter. It stays as an **invariant guard**; the comment in `run-tests.sh` and the test docstring both now say so — the previous comment asserted it was load-bearing.

### Red/green

| leg | result |
|---|---|
| `test_gate_reexec.py` at `scripts/gate.sh` as of `d640356a` (pre-round-1) | **9 failed, 11 passed** (measured on the pre-re-scope file, which was the superset) |
| `test_run_tests_jobs.py` + `test_gate_reexec.py`, `nproc` = 4 (`taskset -c 0-3`) | **25 passed, 0 skipped** |
| same, `nproc` = 8 (`taskset -c 0-7`) | **25 passed, 0 skipped** |
| same, `nproc` = 24 (unrestricted) | **25 passed, 0 skipped** |
| `test_gate_exit_truthfulness.py` + `test_ci_claim_matches_reality.py` | **18 passed** |

**The blocker is closed by CI, not by any of the above.** A local `nix build …#pytests` was green on this box while `tekton/devrc-pytests` was red on the same commit — twice, for two different reasons — which is exactly why 🔴-1 existed.

| head | `tekton/devrc-pytests` |
|---|---|
| `a8d2aa01` | fail — `collected=21365 passed=21357 skipped=7 failed=1`, `test_mjs_parses[attachments.mjs]` |
| `6597318a` | fail — `collected=21495 passed=21488 skipped=7 failed=0` → **GUARD 2**, five unpinned skips |
| `aa406499` | fail — `collected=21535 passed=21532 skipped=2 failed=1`, `test_control_variable_resolution_never_unblocks_a_blocked_repo` |
| **`8fc65c0b`** | **PASS — `collected=21535 passed=21533 skipped=2 failed=0`** (nodetests PASS too) |

🔴 **The `aa406499` failure was investigated, not waved through.** Three things, in order:

1. **Structural** — `test_guard_core.py` imports `guard_core` only, invokes neither `gate.sh` nor `run-tests.sh`, and sits in a directory this PR does not touch.
2. **The mechanism I most suspected is measured absent.** The 4→8 worker cap could plausibly have doubled the sandbox tier's parallelism and shaken out timing-sensitive tests. It did not: the CI log's own banner reads **`parallelism =4 pytest worker(s) (-n 4 --dist loadfile)`** — the builder's `nproc` is 4, so `min(nproc, quota, 8)` is 4, exactly as before. Read from `devrc-ci-ntpgm`'s `step-pytests` log, not inferred.
3. **The population** — across the 13 retained `devrc-ci` runs carrying a verdict, **8 failed, 6 of them on a single test each, and those were 5 DISTINCT tests on 6 different branches** (`test_mjs_parses`, `test_the_SUMMARY_BANNER_names_the_real_selection_source`, `test_every_path_clobbering_site_is_pinned`, `TestARefusedWriteIsIndistinguishableFromAnAbsentOne...` ×2, and mine). `origin/main`'s own head run (`devrc-ci-9vll7`, `4a67ea73`) is clean. This matches the rate the `tekton` skill already documents (~1 test in ~15,500, congestion as amplifier not cause).
4. **The control** — the branch was re-pushed with my 8 files **byte-identical** (`git diff aa406499..8fc65c0b -- CLAUDE.md scripts/` is empty; only two upstream commits differ) and it went **green**. Same tree, different run, opposite verdict.

⚠ **Stated as a limitation, not as reassurance:** a green here is one sample from a tier that is failing ~46% of runs on unrelated single tests. It is evidence this PR does not break the suite; it is not evidence the tier is healthy.

```audit-claims round=1 audited=d640356a8d7949beb9b6bfb4c5d6b9154f5df8f5..8fc65c0b
1. RE-SCOPE: the slot limiter is DELETED, not fixed — acquire_slot, DEVRC_GATE_SLOTS/_SLOT_WAIT/_SLOT_DIR/_SLOT_POOL, GATE_SLOT_HELD/_FD/_STATE, the SLOT LIMITER comment block, the `slot:` report line, the run_tier fd-close subshell (it only ever protected the slot lock), and the whole test file are gone.
2. The 3.4x concurrency table is DELETED rather than restated, in every place it appeared (PR body, CLAUDE.md, gate.sh, test docstring); it existed only to justify the limiter.
3. BLOCKER: test_run_tests_jobs.py derived expectations from os.cpu_count() while run-tests.sh floors on nproc — every expectation now comes from a `_nproc()` subprocess, measured at nproc 4, 8 and 24 via taskset.
4. Every host-conditional `pytest.skip` in that file is REMOVED, because run-tests.sh GUARD 2 pins the expected-skip SET and five unpinned skips red the gate at failed=0 (measured on head 6597318a: collected=21495 passed=21488 skipped=7 failed=0, FAILED). Each "the quota wins" case now uses a cgroup quota of 1, below nproc on any multi-core host.
5. The two host-gated cap tests are merged into one that asserts min(nproc, 8) and never skips, labelled as regression coverage above 4 cores and an invariant guard at or below it.
6. The GATE_SEAMED short-circuit in front of the three re-exec guards is deleted, and each guard's test now pops PYTEST_CURRENT_TEST and sets only its own variable; all three mutants die by their own tests, and a new test pins the GATE_NESTED suppression they bypass.
7. gate.sh resolves its own path into GATE_SELF before any cd, so `cd <repo>/scripts && bash gate.sh` no longer dies rc=127; the test drives it from that directory through a fake nix that actually execs.
8. The --help env-var test derives its expected set from gate.sh's source instead of four literals, with positive and negative controls; all six remaining vars are documented.
9. The "100 of 100" re-exec claim is restated as 101 of 382 (26%) with the selection bias named, and the node-tier sub-claim corrected from 100/100 to 91/101.
10. Stale `-n 4` modelling corrected in run-tests.sh, testlib/store_siting.py and testlib/nolaunch.py; the /dev/shm budget is verified still correct at 8 workers (~15 MB against 62 GB) with no constant changed.
11. A new DEVRC_TEST_BUDGET_ONLY seam stops the runner right after the parallelism banner, cutting the file's ~11 nested full runs from ~150s each to ~3s; it exits 3, never 0, and a test pins that it cannot report a pass.
12. NITS: the cgroup walk no longer escapes its root on a `0::/` path, and `sed -n 's/^0:://p'` replaces `cut -d: -f3 | head -1` (v1-line and colon-in-path bugs). Slot-related nits are deleted with the limiter.
13. CLAUDE.md: the "run BOTH tiers yourself before merging" mandate is DELETED with its reason stated; CI is named as advisory and as the only automated signal; and the bullet says plainly that the new arrangement is LESS SAFE and deliberately so.
14. test_a_quota_wider_than_nproc_does_not_widen_the_budget was found VACUOUS by the sweep (the 8-cap swallowed the mutant on a 24-core box) and now narrows the child's CPU set via taskset so the wrong answer sits between the right one and the cap, with a positive control for the narrowing.
15. tekton/devrc-pytests is GREEN on head 8fc65c0b (collected=21535 passed=21533 skipped=2 failed=0), and nodetests too; the previous head's single unrelated failure was shown to be the fleet-wide flake by a same-tree re-push that went green.
16. The CPU narrowing uses `taskset`, not subprocess `preexec_fn`: this suite runs under pytest-xdist and forking a threaded process to run Python before exec risks a deadlock, which would turn a red test into a hang.
17. One mutant SURVIVES and is labelled rather than fixed: deleting the `q=""; p=""` reset in the cgroup walk kills no test, and it is documented as an invariant guard in both the code comment and the test docstring.
```

ZacxDev:
## Round 2 fixes

All seven round-2 findings addressed. No code defect was reported in the round-1 delta and none was introduced here: the only behaviour change is 🟢-3's redirection on a test-seam-only path, and it now has a test that can see the defect it fixes.

**Payload** = `CLAUDE.md`, `scripts/run-tests.sh` · **Scaffolding** = `scripts/tests/test_run_tests_jobs.py`. (`scripts/gate.sh` is untouched this round.)

### 🟡-A — "the only automated signal that exists" was false, and it was the sentence that most needed to be true

I re-verified it on this host rather than taking the report: `main-green-check.timer` `ActiveState=active`, `OnUnitActiveSec=4h`, last 10:17:35 / next 14:17:35, enabled by `enableMainGreenDeadman = true` at `nix/home.nix:95`. `scripts/main-green-check.sh` runs **both `nix build` sandbox tiers** against `origin/main`'s tip, one at a time, **reproduces a red before alerting** (rc 10 = `RED, REPRODUCED — failed BOTH attempts`), and escalates through `notify-failure@`. Its header names the two commits that landed straight on `main` and broke it on 2026-09-03, found hours later by a human incidentally.

The bullet now says a bad merge is **detected within ≤4h, not never**, names the script, and states the two limits: it **reports only** (never fixes, reverts or pushes), and it is a `serverMode` unit — running on the workbench, not to be assumed elsewhere. The trade is reframed as what it actually is: **prevention was traded for detection**, not for nothing.

### 🟡-B — the CI noise rate now ships, and I re-derived it

Not transcribed. Over the 60 most-recently-updated PRs, newest `tekton/devrc-pytests` status per head from `/repos/…/commits/{sha}/statuses`:

> **21 of 50 terminal verdicts not success (42%)** — 12 `failure`, 9 `error` — **plus 10 of 60 heads (17%) still `pending`**, never resolved.

🔴 **Validating the instrument caught a real bug in my own script.** A dict comprehension over the status list is **last-wins**, i.e. *oldest*, and it reported a head I knew was green as `pending`. The positive control now asserts `8fc65c0b` reads `success` before any count is believed. The audit's independent same-day numbers were 45% / ~10%; the pair is quoted as "roughly two in five reds are noise", with date and method, **not as a constant**.

### 🟡-D — the boundary is 2 CPUs, and it is now stated instead of implied

Round 1 removed five host-conditional *skips* and left a host-conditional *failure* at a different boundary. Re-measured at four points:

| child CPUs | result |
|---|---|
| 1 (`taskset -c 0`) | **1 failed, 13 passed, 2 skipped** |
| 2 | 16 passed |
| 4 | 16 passed |
| 24 | 16 passed |

Stated in the file header, with why it is **not** "fixed": at 1 CPU the quota branch and the `nproc` fallback both answer 1, so the only green options are a vacuous assertion or a skip GUARD 2 cannot forgive — a loud, self-explaining failure beats both. `run-tests.sh`'s "anticipates 1–2-core CI nodes" is reconciled in place: the **runner** is fine at 1 (it goes serial); the 2-CPU floor is a property of the **test file's ability to discriminate**. The header's false "exact everywhere" / "never vacuous and never wrong" claims are gone.

### 🟢s

- **🟢-1** — the "parent's 6-core quota" docstring is corrected; the fixture had moved to `100000 100000` in the same round, and a maintainer restoring 6 would have reintroduced the round-1 defect through a comment.
- **🟢-2** — both stale counts re-derived and re-phrased so they **cannot drift**: "one `_run` per test" instead of "eleven of them — ~28 minutes" (16 cases now), and "the whole file passed and the mutant SURVIVED" instead of "15 passed" (16 now). Called out in-line as measurements that went stale inside the round whose point was retracting an overstated measurement.
- **🟢-3** — all three warning lines moved to one stream (stderr; `RESULT:` stays on stdout). **Mutation-verified**: restoring the old split reds the new per-stream guard, alone. That guard exists because every other assertion in the file reads `stdout + stderr` concatenated, which is exactly why none of them could see it.
- **🟡-C** — PR title: done by the coordinator, not redone.

### Verification

`test_run_tests_jobs.py` green at 2, 4 and 24 CPUs and red-as-documented at 1. `test_gate_reexec.py` + `test_ci_claim_matches_reality.py` + `test_doc_path_rot.py`: 90 passed. Full tiers **not** re-run — this round is prose plus one redirection on a seam-only path, and per the brief a green from a tier measured at 42% not-success is one sample, not an upgrade.

```audit-claims round=2 audited=8fc65c0b23ca5f0b4e845ab580b9b24492c41e0c..d0ef2135
1. CLAUDE.md no longer claims CI is "the only automated signal that exists" — main-green-check.sh is named as a LIVE 4-hourly deadman that runs both nix sandbox tiers against origin/main's tip, reproduces a red before alerting, and escalates through the DND-defeating toast class; verified on this host (timer active, OnUnitActiveSec=4h, enableMainGreenDeadman=true).
2. The same bullet now states the deadman's two limits: it REPORTS ONLY, and it is a serverMode unit running on the workbench, not to be assumed on another host.
3. The "less safe" paragraph is reframed as prevention traded for DETECTION rather than for nothing, and records that an earlier draft understated the backstop by a whole subsystem.
4. The measured CI not-success rate now ships in CLAUDE.md rather than only in the PR body: 21/50 terminal verdicts not success (42%) plus 10/60 heads stuck pending (17%), re-derived 2026-09-09 with the method and date stated, and paired with the audit's independent 45%/~10%.
5. That measurement's instrument was validated first, and the validation caught a last-wins/oldest-status bug in the measuring script; the positive control now asserts a known-green head reads success before any count is believed.
6. CLAUDE.md now tells the reader to act on the failing TEST rather than the colour, citing six recent reds across six branches that were six different unrelated tests with origin/main's own run clean.
7. test_run_tests_jobs.py's header states the 2-CPU boundary explicitly, with the four-point measurement (1 CPU = 1 failed/13 passed/2 skipped; 2, 4, 24 = 16 passed) and why a skip is not the remedy (GUARD 2 cannot forgive it) — replacing the false "exact everywhere" and "never vacuous and never wrong" claims.
8. run-tests.sh's "anticipates 1–2-core CI nodes" is reconciled in place: the runner is fine at 1 core and goes serial; the 2-CPU floor belongs to the test file's ability to discriminate, not to the code under test.
9. The "parent's 6-core quota" docstring is corrected to 1, with a warning that restoring 6 would pass on a 24-core host and fail on the 4-core builder — the round-1 defect reintroduced by a comment.
10. Three self-reported counts that went stale inside round 1 are re-derived and re-phrased as rates that cannot drift ("one _run per test", "the whole file passed") instead of totals that can.
11. The DEVRC_TEST_BUDGET_ONLY warning is no longer split mid-sentence across stdout and stderr; all three lines go to stderr, and RESULT: stays on stdout.
12. That split now has a test that can SEE it — a per-stream assertion, added because every other assertion in the file reads stdout+stderr concatenated — and it is mutation-verified: restoring the old split kills it, alone.
```

PR #1430, opened 2026-09-09
feat(bar): clicking the fans pill opens a COOLING view, not btop

ZacxDev:
Clicking the fans pill opened **btop** — a CPU/memory view. When that pill goes red the two questions are *"is the pump dead"* and *"how hot is the thing it was cooling"*, and btop answers neither: it shows no fan or pump data at all.

`scripts/fans-detail` is the cooling equivalent of the existing memory/disk/media detail floats:

```
  AIO pump    2758 rpm  █████████░  87%  floor 500
  Case fan    1641 rpm  ██████░░░░  60%  no floor
  8 unconnected header(s): fan2, fan4 … fan10

  CPU  Tctl      79.9 C  ████████░░
       Tccd1     74.8 C  ███████░░░
       Tccd2     78.0 C  ████████░░
  GPU  edge      48.0 C  █████░░░░░
  board AMD TSI  79.0 C  ████████░░
        VRM/chip 35-44 C  (6 sensors)
  disk  NVMe     28-55 C  (11 sensors)
```

Live-redraws every 2s (the `media-detail --watch` idiom); `--dump` prints one frame for tests and piping.

## The predicate is not re-implemented

`fans-detail` loads `i3status-fans` **by path as a co-located sibling** to reuse the chip-location and tacho-reading logic, rather than open-coding a second copy. That drift would be silent here, because only one of the two is ever on screen. It makes the sibling a **deploy requirement**: both carry the same `!isLaptop` gate, pinned by `test_fans_detail_and_its_SIBLING_are_deployed_together`, and a `fans-detail` deployed alone renders a red banner rather than an empty screen.

## A bug that only existed after a nix switch

🔴 The first version used `Path(__file__).resolve()`. Under nix that is wrong: `home.file` deploys **each file as its own store path**, so the deployed script is a symlink to `/nix/store/<hash>-hm_fansdetail` — a *file* directly in `/nix/store`, not a directory containing anything. The sibling's computed location became `/nix/store/i3status-fans`, which does not exist.

A correctly-deployed pair rendered the "sibling did not load" banner with **both fan rows `unreadable`**, on a host where the pill beside it was reading 2758 rpm.

**All 24 tests passed through this**, because they run from the repo where the two files genuinely *are* siblings — the layout the bug depends on only exists after a switch. `abspath` (never `realpath`/`resolve`) is the repo idiom for exactly this reason; `i3status-clawgate` loading `bar_freshness.py` does the same.

The new test builds that layout — two lone files in separate directories plus a symlink dir that is the only place they are siblings — and carries a **positive control** that deletes the sibling and asserts the banner *does* appear, so it cannot pass vacuously. Watched red on the pre-change code: reverting to `.resolve()` fails **only** that test (1 failed / 24 passed), confirming it is the sole guard for this class.

The banner is also why this was caught at all: it announced itself instead of showing an empty screen.

## Two more defects my own tests caught

- 🔴 `except SystemExit: args = p.parse_args([])` **discards every good argument alongside the bad one** — so `--dump --nonsense` dropped `--dump`, fell into the watch loop and **hung forever**. The test *timed out* rather than failed, which is how that ships unnoticed. Now `parse_known_args`.
- 🔴 A test asserting `"0 rpm" not in out` over the whole frame is satisfied by nothing and broken by everything: the case fan's legitimate `1650 rpm` **contains** `0 rpm`. Fixed to assert on the pump's own row.

Also: sysfs `pwmN` is **raw 0-255, not a percentage** — unscaled it renders a pump at "201%". Pinned at five points including both endpoints.

## Verification

**25 tests**, hermetic against a fake hwmon tree — no test reads the builder's real `/sys`, so no verdict depends on the runner's hardware.

**Mutation: 4 mutants, 4 killed**, each by its own named guard — click reverted to btop, sibling gate diverged, pwm scaling removed, missing-sibling banner deleted.

**Live**: deployed via `home-manager switch`, ran the click command verbatim, confirmed a real float opened (i3 `floating=user_on`), screenshotted it rendering against the real NCT6687D, then closed it and restored focus + workspace.

### Gate — dev-host tier, merged tree at base `53c9f664`

🔴 **Stated precisely, because it was NOT one clean run.** The box is at load ~70 on 24 cores (other sessions' concurrent suites), and the pytest tier hit its 3600s timeout **twice** — both times with **zero failing tests**, killed mid-run rather than failing.

| | result |
|---|---|
| node tier | **PASS** — 1449/1449, 0 fail |
| pytest, 22 of 29 targets (gate run) | **PASS**, 0 failures |
| pytest, remaining 7 targets (run directly) | **PASS** — 1014 passed, floor sum 940 |

So **all 29 targets are measured green, but as 22 + 7 rather than as a single completed tier.** The target ledger comes from `run-tests.sh --check-floors` (the runner's own two-way-pinned enumeration), after a hand-rolled grep of mine returned 39 entries with duplicates and falsely listed `scripts/tests` as never-started.

⚠ **The nix sandbox tier was not run on this tree.** That is the tier a merge nominally gates on and it is blind to different things; it is not measured here.

⚠ `main` has moved since `53c9f664`; nothing in those commits shares a file with this diff, but the merged-tree claim above is against that base, not against current `main`.

https://claude.ai/code/session_019LDvGed1jgLsz2UXSZNTdm


ZacxDev:
## 🔴 Correcting this PR's own description — the headline claim was FALSE

An adversarial audit (round 1) refuted the central architectural claim in the body above. Posting it here rather than silently editing the body, because a reviewer may already have read and believed the wrong version.

**What the body said:** *"fans-detail does NOT re-implement the chip-location / tacho-reading predicate — it loads `i3status-fans` by path, so the pill and the view cannot disagree."*

**What was true:** only `read_rpm` was reused. `CHIP`, the chip-location walk, and the entire fan table (`KNOWN_FANS`) were **second copies**. Both were measured drifting with **84 of 84 tests green**:

| drift | result before the fix |
|---|---|
| chip renamed in the sibling only | pill `{"text": "2448·1650", "state": "Idle"}` beside view `Super I/O NOT FOUND — the nct6683 driver is not loaded / persist it: sudo bash …` |
| pump moved to `fan2` in nix only | pill `2448·1650 Idle` beside view `AIO pump ? ░░░░░░░░░░ ?% unreadable` |

The first is the worse one: an **actively wrong diagnosis** pointing the operator at a `sudo` script that fixes nothing, next to a healthy pill.

### Fixed at the mechanism, not the sentence (`9ae4fdb4`)

- The fan mapping is now **one `fanArgs` binding** in `nix/graphical.nix`, passed verbatim to both the pill and the view. `fans-detail` gained `--chip`/`--fan` and parses them with the **sibling's own `parse_fan`**.
- The chip name comes from the sibling's `DEFAULT_CHIP`.
- A guard resolves the binding and asserts **both** commands interpolate it. Both reproductions above now **fail the suite** instead of passing it.

### Also fixed from the same round

- **The footer advertised a `q` key that did nothing.** The watch loop read stdin nowhere — measured under a pty, `q` left it running past 8 s, so the keypress echoed, got wiped by the next redraw, and the view read as hung. 🔴 The *fix* then introduced the mirror defect, which the same harness caught: with cbreak alone `\x03` is eaten by the line discipline and becomes a SIGINT that only arrives if the process is in the tty's foreground process group — true under alacritty, not under a bare pty — so **Ctrl-C stopped working while the footer still promised it**. ISIG is now cleared so the byte is readable and the promise holds in both. Pinned by six pty cases including **two negative controls** (an unrelated key, and no key, must *not* exit).
- **`--dump` could exit 0 with zero bytes on stdout.** `read_str` caught only `OSError` while `read_int` also caught `ValueError`; a non-UTF-8 hwmon `name` raised `UnicodeDecodeError` straight past `render` into the failsafe.
- Temperatures were gauged with `bar()`, which takes a **percentage** — 80 °C rendering as 8/10 looked plausible and meant nothing. Now a 30–100 °C scale with threshold colouring.
- `tlabel[:8]` rendered `AMD TSI Addr 98h` and `…9Ah` **identically**.
- The frame was 74 columns against `columns=76`; widened to 84.

### Not fixed, deliberately

`sys.dont_write_bytecode` is unguarded and process-global — inert today, and the audit confirmed it is earning its place: without it a `.pyc` **is** written into `~/.config/i3status-rust/scripts`, where every nix store path has `mtime=1` and CPython would honour that cache across any same-size redeploy forever.

### Verification after the fix

666 tests green. Both audit reproductions re-run and now caught. Six pty cases for the keys. Rendered against the real NCT6687D.

⚠ Unchanged from the body: **the nix sandbox tier has still not been run on this tree**, and the dev-host evidence is a 22+7 split rather than one completed tier (the box has been at load ~70 and the tier timed out twice, both times with zero failing tests).

PR #1432, opened 2026-09-09
test(harness): devrc never received cairn #3 — the spawn port pick is a TOCTOU here too

ZacxDev:
devrc's `_free_port()` binds port 0, reads the number, **closes the socket**, and returns. Between that and the spawned child's `bind()` — one interpreter startup — the kernel may hand the same port to anything else on the box, including this very process: the health probe `connect()`s, and an outbound connect draws its local port from the same ephemeral range.

`ZacxDev/cairn` closed this in **#3 (`8e4ef84`)**. devrc's fork has carried the unfixed copy since the split — it has **zero** occurrences of `SPAWN_ATTEMPTS` or `_lost_the_port_race`. **This is a port, not a cherry-pick**: `diff -u` between the two copies is 5,819 lines over 28 hunks, so `git apply` of `8e4ef84` cannot land.

**Re-measured here rather than importing cairn's number:** 1000 trials × 20 subsequent `bind(("127.0.0.1", 0))` recycled the just-released port **4 times**; the same loop with the socket still **OPEN**, **0 times**. The control is what makes the 4 a measurement instead of noise.

## 🔴 This is NOT a fix for the CI intermittent, and the class docstring says so

The recurring `TestARefusedWriteIsIndistinguishableFromAnAbsentOne` failure is now **diagnosed** — `MECHANISM = SERVER_BLOCKED_IN_FSYNC`, the handler parked in an unbounded `fsync` inside the request, surfacing as `TimeoutError` in `recv_into`: **established and unanswered**. This race's signature is EADDRINUSE, i.e. connection **refused**. Different mechanism. Whether closing this moves that rate is **unknown and written as unknown**.

What this fixes is a real, independently-worth-closing race devrc has run with since the fork.

## Scope, and what deliberately is not here

Retries **only** the lost port race, and only `SPAWN_ATTEMPTS = 3` times — an unbounded retry turns a box with no free ports into a hang instead of a failure. `_lost_the_port_race` derives its needle from `errno.EADDRINUSE` rather than spelling it. `_never_healthy_message` names host:port, the budget, the last probe exception verbatim, earlier lost races, and both drained streams.

⚠ **devrc has no `_run_to_completion`.** There are exactly three `server.py` spawn sites here, and the two exit-78 tests reach `EXIT_CONFIG` **before** `build_server` ever binds — so an occupied port is invisible to them. That is written into `_free_port`'s docstring as the trap it is, rather than inventing a second retry loop for a caller that does not exist.

⚠ **No `child_was_alive=` field.** Cairn's ladder deleted one after a mutant proved it structurally pinned to `True`. The field that *can* vary — `exit=<rc> is <how>` — is pinned by three tests across four causes.

## Matrix — red at `8970142d`, green at HEAD

| test | at base |
|---|---|
| retried-and-survived | `server exited 1: … server_bind → self.socket.bind → OSError: [Errno 98] Address already in use` |
| a thief that ACCEPTS AND NEVER ANSWERS | `server never became healthy` |
| the message names what the old one dropped | `assert 'server never became healthy on 127.0.0.2:' in 'server never became healthy'` |

That first traceback is the evidence the fixture **genuinely reaches `bind()`** — not the `EXIT_CONFIG`-before-bind trap. Invariant guards are labelled as such and not counted as regression coverage.

## Mutation battery — 16 mutants, 16 killed by their intended guard's own message

`PYTHONDONTWRITEBYTECODE=1`, `__pycache__` cleared between mutants, tree byte-compared against the snapshot after.

🔴 **One survivor, recorded in the code rather than hidden.** `test_the_retry_is_BOUNDED…` **survives** `SPAWN_ATTEMPTS 3→1` in isolation: every assertion is written in terms of the constant, so its fixture can only ever produce the constant's own value. The docstring says so and names the two guards that do kill it.

⚠ **Re-verifying that by hand caught a sibling trap worth knowing:** `SPAWN_ATTEMPTS = 3` matches **twice** in the file — the definition and that docstring — so a text-anchored mutant scores **INVALID, not SURVIVED**, and the pristine green reads exactly like a surviving mutant. Mutate line 7026, not the string.

758 passed in the file (744 at base, **+14**). No stray race warnings leaked. Collected-test floor untouched — +14 is a rise against a minimum.

⚠ Merged onto current `main` (`78a52387`) so the gate judges the merged tree.

https://claude.ai/code/session_01NTRPN6dhkdf8kZJch3mne5

### DIFF
--- a/scripts/tests/test_subsystem_store_api.py
+++ b/scripts/tests/test_subsystem_store_api.py
@@ -67,17 +67,20 @@ import io
 import os
 import re
 import shutil
+import signal
 import socket
 import socketserver
 import stat
 import tarfile
 import secrets
+import select
 import subprocess
 import sys
 import textwrap
 import threading
 import time
 import traceback
+import warnings
 import urllib.error
 import urllib.request
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


ZacxDev:
⚠ **`/audit-pr 1432` was OFFERED, RECOMMENDED, and NOT RUN — recorded so this reads as skipped, not clean.**

This ships on its own evidence, which is real but is **not** an adversarial read:

- regression matrix red at `8970142d` / green at HEAD, with the base ref's own stack (`server_bind → socket.bind → OSError: [Errno 98]`) proving the fixture reaches the racing line rather than the `EXIT_CONFIG`-before-bind trap;
- 16 mutants, 16 killed by their intended guard's own message, with **one survivor reported rather than hidden**;
- both gate tiers green on a **merged** tree (`78a52387`): `collected=21426 passed=21424 skipped=2 failed=0`, which is `21412 + 14` — the `+14` confirmed from a direction independent of the author's own count.

Two of those mutation results were re-verified by hand rather than accepted, and doing so caught a trap worth recording: **`SPAWN_ATTEMPTS = 3` matches TWICE** in the file — the definition and the docstring recording the survivor — so a text-anchored mutant scores **INVALID, not SURVIVED**, and the pristine green reads exactly like a surviving mutant. The harness refused to write; a harness that did not would have reported a false SURVIVED. Mutate line 7026, not the string.

The precedent this follows is rank 12 of the cairn effort, where `/audit-pr 6` was likewise offered and never run. That is a known, repeated gap in this effort — not a claim that this change was reviewed.

PR #1435, opened 2026-09-09
fix(bar): the runaways pill must not toast — cpu-monitor already owns that event, with a cap

ZacxDev:
## A shipped regression from #1404, and it is mine

`scripts/cpu-monitor.sh` is live as a systemd user unit (`active`, `MainPID 1470`) and **already
owns the runaway toast**: `⚠ Runaway process: <comm>` at **critical** urgency after a 6-sample
sustain (`:276`), capped at `CPU_MON_MAX_ALERTS_PER_DAY` (default 8, the `alert_capped`
predicate). 🔴 **CORRECTED — it has no recovery TOAST**: `clear_alert` is journal-only unless
`CPU_MON_CLEAR_TOASTS=1`, which nothing in this repo sets. Neither announcer has one.

#1404 added a **second announcer for an OVERLAPPING but not identical set** — normal urgency,
**no cap**, firing ~7 minutes later for a process hot since birth. 🔴 **CORRECTED — the original
claim of "the same process" was too strong**: cpu-monitor reads `top`'s first data row
instantaneously, the pill reads a lifetime-average `pcpu >= 80` with `etimes >= 600` over every
eligible process, so the sets overlap rather than coincide.

🔴 **CORRECTED — the duplicate DID fire, and the earlier claim here was false.** This paragraph
previously said `~/.cache/bar-status/runaways.toast-state` does not exist, so the count never
crossed. MEASURED on the workbench: it **exists**, `{"latched": false}`, written 2026-09-09
10:32:10. The only writer is `bar-status-poll.evaluate_edge_toast`, which writes **only on a latch
transition**, and `read_latch` returns `False` for a missing file — so a stored `false` requires a
prior `true`, and writing `true` is on the same code path as the `fire(...)` call. The rising edge
happened. ⚠ Whether the toast reached the screen is a **separate, unestablished** claim:
`dunstctl history` carries no `bar-status` appname at all, so that instrument cannot confirm
delivery for *any* poller toast. Use the latch as the evidence; do not restate "never observed".
(The original claim was true when an audit first read it and became false afterwards — the live
poller runs from the shared checkout, which still carries the spec.)

🔴 **That cap is not incidental — but the measurement originally cited here was the wrong one.**
The `123-267/day → 11-32/day` cut belongs to a **threshold raise** (#351, `4c3e9132`), not to the
cap, which arrived six days later (#430, `d71bf18f`). The cap earns its keep on its own evidence:
over the last 30 days it demoted runaway alerts on **8 separate days**, 283 in total, including
101, 73 and 67 on 08-30, 08-31 and 09-06. ⚠ `alert()` journals only on the fallback path, so
demotions are countable and deliveries are not.

## The fix

Remove the pill's toast spec. The division is now explicit and written down: **cpu-monitor owns
the INTERRUPT, the pill owns the AMBIENT display.** `media` and `airvpn` are already polled
sources with no toast spec, so a source without one is the normal shape, not a gap.

Removed with it, because each existed **only** to arm that latch — and a field nothing reads is
the inert-`RUNAWAY_IGNORE` defect this same source shipped once already:

- `new_count` and the `prev_pids` plumbing (`parse_runaways` is pure again; `fetch_runaways` no
  longer reads the previous cache)
- `SYSHEALTH_TERM` and `_syshealth_action()`, the toast's own click target

The block's own click is untouched: `syshealthCmd` still opens syshealth in a float terminal
that holds open.

## Guards — an ABSENCE needs one

Nothing else goes red when a spec is added back, and adding one is the obvious "improvement".

| mutant | result |
|---|---|
| **F** — a `runaways` toast spec added back | **KILLED** by its own assertion, which names *why* (cpu-monitor already owns it, with a cap) |

`test_the_runaways_source_has_NO_TOAST_SPEC_because_cpu_monitor_owns_it` also carries a control
proving the guard can still see a source that legitimately *has* a spec — otherwise it would
pass against a `_toast_specs()` that returned nothing at all.
`test_the_runaways_CLICK_still_holds_its_terminal_open` keeps the click's terminal+hold shape
pinned from `graphical.nix` after the toast's own copy went away.

## Gate

Sandbox tier, base `f4a42927`, both derivations actually built, verdicts read from the runners'
own `RESULT:` lines:

- `pytests` — `collected=21411 passed=21409 skipped=2 failed=0` · `RESULT: PASS (exit=0)`
- `nodetests` — `tests=1449 pass=1449 fail=0` · `RESULT: PASS (exit=0)`

🔴 **CORRECTED — the test count was not parity.** `test_bar_status.py` collects **569** at
`f4a42927` and **568** at `3dd1491a`: a net −1 (3 removed, 2 added), presented here originally as
"568 before and after". After the round-1 fixes it collects **570**.

⚠ **The gate numbers above are stale** — they are scoped to base `f4a42927` and to the tree at
`3dd1491a`, before the round-1 fixes. See the round-1 findings comment and the re-gate on the
merged tree.

⚠ Re-gate the merged tree at merge time — `strict` is deliberately off here, so a green check is
a claim about this branch, not about the tree a merge creates.


ZacxDev:
## Audit round 1 — verdict: MERGE AFTER FIXING

Ten findings. No deploy-blocker: the payload is correct, the behaviour change is host-scoped
(workbench only — both the poller and the block are `mkIf (!isLaptop)`) and reversible. But
findings 1–3 are **false statements in the three surfaces a maintainer reads to decide whether to
undo this change**, and finding 4 is a pin that does not pin what its name says.

**Gate, on the MERGED tree** (`3dd1491a` merged into `origin/main` `b79ccfbe`, merge clean, `ort`,
no conflict) — the sandbox tier, built **one derivation at a time**:
`pytests` → `RESULT: PASS (exit=0)`, `collected=21478 passed=21476 skipped=2 failed=0`, no timeout panic.
`nodetests` → `RESULT: PASS (exit=0)`, `suites=5 files=41 tests=1449 pass=1449 fail=0`.
⚠ The **dev-host tier (`scripts/gate.sh`) was NOT run** — this claim is scoped to the sandbox tier
and to base `b79ccfbe`.

---

### SHOULD-FIX 1 — the guard's failure message asserts a property cpu-monitor does not have, six lines under the correction
`scripts/tests/test_bar_status.py:3039` · scaffolding

The assert message says cpu-monitor *"already owns this event with a sustain, a daily cap **and a
recovery toast**."* Line 3033 of the **same docstring** says the opposite, and is correct: that
`✓ Runaway process cleared` is JOURNAL-ONLY unless `CPU_MON_CLEAR_TOASTS=1`.

MEASURED: `cpu-monitor.sh:284` calls `clear_alert`, which at `:178-184` toasts only when
`CLEAR_TOASTS=1`; default `0` at `:70`; no setter anywhere in the tree (positive control: the
`CPU_MON_` pattern matches 6 lines in `nix/home.nix`). This is the exact claim `3dd1491a` was
written to retract — it survived in the one surface that prints **at the moment a maintainer is
deciding whether to re-add the toast**. Failure scenario: a future session adds a `runaways` spec,
the guard fires, they read "cpu-monitor has a recovery toast", conclude the removal was safely
redundant, and delete the guard instead of the spec.

### SHOULD-FIX 2 — the cited measurement belongs to a different knob, shipped six days earlier
`scripts/bar-status-poll:865-868`, `scripts/tests/test_bar_status.py:3024-3025` · payload + scaffolding

Both credit the `123-267/day → 11-32/day` cut to the cap. MEASURED with `git log -S`:
`nix/home.nix:538-539` — the source being cited — credits it to raising
`CPU_MON_THRESHOLD`/`RUNAWAY_PCT`, i.e. **#351 `4c3e9132`, 2026-08-06**.
`CPU_MON_MAX_ALERTS_PER_DAY` arrived **six days later** in **#430 `d71bf18f`, 2026-08-12** — the
same commit that wrote the 123-267 sentence retrospectively, about the earlier change. #430's own
headline is a different number (~330 → ~60/day).

🔴 **The conclusion is TRUE and the evidence for it is better than the evidence quoted.** Measured
independently over 30 days: the runaway cap demoted alerts on **8 separate days**, including
**101, 73 and 67** demotions on 08-30, 08-31 and 09-06. Cite that, or #430 — not #351's threshold
measurement. ⚠ Caveat on that number: `alert()` (`:159-166`) journals only on the *fallback* path,
so delivered toasts leave no journal line. Demotions are countable; deliveries are not.

### SHOULD-FIX 3 — the PR description still carries three claims this PR's own second commit measured false
payload

`3dd1491a` retracted them in its commit message; the body was never updated, and a reviewer reads
the body.

| body says | measured |
|---|---|
| "with a matching `✓ Runaway process cleared` on recovery (`:284`)" | FALSE as deployed — finding 1 |
| "#1404 added a **second announcer for the same process**" | retracted as F3 by `3dd1491a`: the predicates are not the same set |
| "568 tests in `test_bar_status.py` before and after" | FALSE — **569** at `f4a42927`, **568** at `3dd1491a` (`--collect-only` on both trees; 3 removed, 2 added) |

The gate section also pins base `f4a42927`; the base is now `b79ccfbe`, 8 commits ahead, one of
them touching `nix/graphical.nix`.

### SHOULD-FIX 4 — the click "pin" pins a string, not the click. MUTANT SURVIVED
`scripts/tests/test_bar_status.py:3049-3065`, claimed by `nix/graphical.nix:50` and `:59` · scaffolding

MEASURED — mutant `I`: change `nix/graphical.nix:518` from
`{ button = "left"; cmd = syshealthCmd; }` to `cmd = btopCmd;` → **568 passed, SURVIVED.**
The runaways pill's left-click can be rewired to a different view with the full suite green.

`:50` asserts *"The runaways pill's clicks — BOTH buttons, and the only consumer"*; `:59` says
*"Pinned by `test_the_runaways_CLICK_still_holds_its_terminal_open`"*. That test regexes the
`syshealthCmd` let-binding and asserts three properties **of the string**; it never opens
`runawaysBlock`. The docstring names a relationship, the body inspects one side.

Reachability controls confirm the guard is otherwise live: mutant `CTRL_hold` (drop `read -n 1`)
→ KILLED by that test's own message; mutant `H` (no longer a terminal) → KILLED by the same.
The repo already has the missing idiom — `test_fans_detail.py`, arrived via #1430, is the only
file asserting on `button = …` wiring.

### SHOULD-FIX 5 — a count in one of this PR's own files, staled by #1404 and still wrong
`scripts/bar-status-poll:17`, `:565`, `:596`; also `scripts/i3status-civitai:29` · payload

All four say **seven** block scripts use `bar_freshness.py`. MEASURED: **eight** call
`fresh.unmeasured(...)` — `airvpn, alerts, civitai, clawgate, mail, media, runaways, telemetry`.
`runaways` is the eighth, and `i3status-runaways:83-84` renders `?` exactly like the others. The
`SOURCES` ledger at `test_bar_status.py:3107-3108` pins eight names; **nothing pins the prose.**
The enumeration of `?` spellings at `:562-564` lists six and omits this pill.

This is the class a delta round structurally cannot see — true when written, falsified by #1404 —
and it sits in the file this PR edits.

### SHOULD-FIX 6 — the enumerated trade omits a fourth class, and it is live on this host
`scripts/bar-status-poll:876-882` · payload

The comment enumerates what removing the interrupt gives up as (a) the 80-95% band, (b) the
runner-up, (c) rank churn — and reads as complete. There is a **(d)**: `cpu-monitor.sh:257-261`
runs `is_ignored` on the *top* process and on a match sets `cand=""` **without falling through to
the runner-up**; the `else` branch then resets `runaway_streak` to 0 and clears
`runaway_alerting`. So while an ignored game holds the top slot, cpu-monitor is blind to *every*
runaway and any in-progress streak is destroyed. `nix/home.nix:2352-2358` records 12 real alerts
naming a game, so this is a state the host actually reaches.

(a), (b) and (c) are themselves all accurate — verified.

### SHOULD-FIX 7 — the operator-facing pill ledger has no `runaways` row
`claude/skills/bar/SKILL.md:32` and the table at `:106-116` · payload

The `bar` skill's per-pill table omits `runaways` entirely, as does `:32`'s block-script
enumeration. #1404 shipped a pill without a ledger row; this PR establishes an invariant about it
("deliberately no toast — widen cpu-monitor instead") that a maintainer operating the bar through
the skill will never see. ⚠ That skill has an **enforced byte ceiling**, so this needs an eviction
in the same commit, not an append — which makes it a decision rather than a typo.

### NIT 8 — a cross-reference in one of these files has already rotted by 241 lines
`nix/graphical.nix:37` · payload, pre-existing

`:37` cites `nix/home.nix:298` for the 123-267 measurement. That text is at `nix/home.nix:539`;
line 298 is `services.espanso`. Not introduced here — but it is the empirical proof that the four
refs this PR adds (`home.nix:539`, `home.nix:2345`, `cpu-monitor.sh:209`, `cpu-monitor.sh:284`)
will rot the same way. All four are correct today; nothing pins any of them.

### NIT 9 — "Every field here is consumed" re-asserts the sentence the previous revision recorded as false
`scripts/bar-status-poll:1379-1381` · payload

The removed comment called `cpu`/`cmd` *"a deliberate, **stated exception**"* and noted that an
earlier revision had claimed every field had a consumer, "which was false for these two". The
replacement opens *"Every field here is consumed"* and supplies a human reader as `pid`'s consumer.

MEASURED whole-tree: `count`/`state` are read by `i3status-runaways`; `cpu`/`cmd` are read
in-function to build `detail`; **`pid`, `detail` and the `processes` container have no machine
reader.** It sits three lines from this PR's own stated deletion rule (*"a field nothing reads is
the inert-`RUNAWAY_IGNORE` defect"*). Either restore "stated exception", or apply the rule
symmetrically.

### NIT 10 — `_toast_specs`'s docstring invariant now holds for no spec that needed it
`scripts/bar-status-poll:825-826` · payload, pre-existing

*"The action opens the SAME target as the block's left-click."* Per spec: `alerts` ✔, `civitai` ✔,
`clawgate`/`clawgate_stuck` ✔, `mail` action is `None` (vacuous), **`telemetry` ✗** (`xdg-open`
a dashboard URL vs an `alacritty … read -n 1` block click, `:344`). Already false before this PR —
noted only because the deleted `runaways` spec was the one whose action existed expressly to
satisfy this sentence, so nothing now enforces it.

---

## Verified clean

- **Deleted-symbol sweep** (`find … -print0 | xargs -0 grep`; positive control `syshealthCmd` → 2 files): `SYSHEALTH_TERM`, `prev_pids`, `test_the_toast_action_and_the_CLICK_are_the_same_shape` — **zero** references. `_syshealth_action` and `new_count` survive only in comments that say they were deleted.
- **No dead code**: `read_prev_status` still used at `:1040`, `:1465`, `:1568`.
- **A missing spec is safe**: `dispatch_edge_toast:551` uses `specs.get(name)` → `None`, no `KeyError`. `media` and `airvpn` already prove the shape.
- **Mutants F and G KILLED, verified independently**, under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` cleared between mutants and `CTRL_hold` as a known-caught positive control in the same batch. `F` (plain `runaways` spec re-added) → killed by `…_has_NO_TOAST_SPEC_because_cpu_monitor_owns_it`. `G` (a derived spec under a *different* name — the obvious walk around an absence guard) → killed by `…SOURCES_table_is_a_LEDGER_of_every_polled_source`.
- **No coverage regression from the deleted toast test**: mutating `:913` to `payload.get("count")` kills **8** tests on the merged tree (9 at base) — the `clawgate_stuck` gate covers it independently.
- **`X-Restart-Triggers`**: the `scripts/syshealth` entry is still correct (`SYSHEALTH` exec'd at `:1341`). ⚠ The list was never *complete* by its own stated rule — the poller also loads `bar_freshness.py`, `mail-actions/_db.py`, `airvpn-menu` and `data/airvpn-servers.json` by working-tree path, none listed, while `graphical.nix:865-867` claims the ledger enumerates the class. Runtime consequence nil (ExecStart is itself a working-tree path on a 45s oneshot), so not filed separately — but the completeness sentence is false.
- **`--mock-runaways` help text** and **`i3status-runaways:13`**: both accurate.
- **Base movement**: #1430 added ~38 lines to `nix/graphical.nix` around the fans pill; the merged diff touches only the runaways comment block. Read by eye — no semantic overlap. #1430's fans-detail click needs no hold (it live-redraws), so it is not a competing spelling of the rule `graphical.nix:53` claims sole ownership of.
- No leaks, no secrets, no new resource handles. Net deletion.

⚠ **Two contextual notes, not findings.** (1) `~/.cache/bar-status/` has a `.toast-state` latch for every spec'd source **except** `runaways`, so this toast has not fired on the workbench — but #1404 merged only ~5 h before that reading, so it is weak evidence. It does mean the body's "produces two desktop toasts" is derived from the predicates, not observed. (2) A dozen orphaned `sleep` processes under a pytest tmp path belong to another session's run and self-reap; left alone rather than guessed at.

ZacxDev:
Round 1 fixes are pushed as `0e9c469c` (fixes `2337eea5` + `da2a73c3`, then a merge of current
`main` `4374ebad`). Claims block for the next round:

```audit-claims round=1 audited=3dd1491abc14a0d9c7c1b29d774658bfaa08ecf3..0e9c469c
1. Finding 1 (the guard's assert message claimed cpu-monitor has a recovery toast, contradicting its own docstring six lines up): claimed addressed by rewriting the message to claim only the 6-sample sustain and the per-day cap, and to state that NEITHER announcer has a recovery toast, so nobody re-adds the spec to obtain one.
2. Finding 2 (the 123-267/day -> 11-32/day cut was credited to the daily cap; it belongs to a threshold raise six days earlier): claimed addressed at BOTH sites by keeping the conclusion and replacing the evidence — the cap demoted alerts on 8 of the last 30 days, 283 total, 101/73/67 on 08-30, 08-31 and 09-06 — with the caveat that `alert()` journals only on the fallback path, so demotions are countable and deliveries are not. Claimed measured: `4c3e9132` = #351 (threshold raise), `d71bf18f` = #430 (the cap's first appearance, via `git log -S MAX_ALERTS_PER_DAY --reverse`).
3. Finding 3 (the PR DESCRIPTION still asserted a recovery toast, "the same process", the wrong PR for the cap measurement, and "568 tests before and after"): claimed corrected in the PR body itself, each marked as a correction rather than silently rewritten, and the stale gate numbers labelled as scoped to base `f4a42927` / head `3dd1491a`.
4. Finding 4 (the click "pin" regexed the syshealthCmd STRING and never opened runawaysBlock; mutant I SURVIVED at 568 passed): claimed addressed by a new `test_the_runaways_pill_CLICKS_are_wired_to_syshealthCmd` following `test_fans_detail.py`'s button-wiring idiom, plus corrected `graphical.nix:50`/`:59` naming which test pins the string and which pins the wiring. Claimed measured: mutant I KILLED after, 1 failed / 569 passed, by that test's own message; `CTRL_hold` KILLED both sides as the reachability control; a negative control reads fansBlock's left-click so a green cannot be the regex matching nothing.
5. Finding 5 (four sites said SEVEN freshness consumers; there are EIGHT): claimed corrected at all four named sites PLUS two the finding did not name (`bar_freshness.py:14-19` and `:178`, same wrong number), and claimed pinned by a new `test_the_PROSE_COUNT_of_freshness_consumers_matches_the_SOURCES_ledger` that derives the count, asserts set-equality with `SOURCES`, and matches whitespace-normalised sentence templates so a stale count is distinguishable from a reword.
6. Finding 6 (the enumerated trade listed three given-up classes and read as complete): claimed addressed by adding (d) — `is_ignored` on the top row sets `cand=""` with no fall-through, and the else branch resets the streak — so an ignored top process blinds cpu-monitor to every runaway. (a)(b)(c) claimed re-verified.
7. Finding 7 (the `bar` skill's pill table and block-script enumeration had no `runaways` row): claimed addressed by adding both, with the NO-TOAST invariant and what it costs stated in the row. 🔴 Claimed CORRECTION to the finding's own caveat: `claude/skills/bar/SKILL.md` has NO enforced byte ceiling, so no eviction was required — the ceilinged docs are RULES.md plus browser, prune-skill, session-manager, handoff and clawgate. Claimed NOT addressed and filed instead: that body is 27,844 B against a 12,038 B advisory budget and nothing gates it.
8. Finding 8 (a cross-reference already rotted 241 lines): claimed corrected to `~:539` plus the quoted phrase, and the four line refs this PR added claimed re-spelled to name symbols (`alert_capped`, `runaway_proc`, `find_runaways`, the `clear_alert` call) beside any number.
9. Finding 9 ("Every field here is consumed" re-asserted a sentence a previous revision recorded as false): claimed restored as a stated exception; claimed measured whole-tree with GNU grep that `pid`, `detail` and the `processes` container have no machine reader, with `evaluate_edge_toast:975` named as a real reader that is unreachable for `runaways` because it has no spec.
10. Finding 10 (`_toast_specs`' docstring invariant was false for `telemetry`): claimed made true by naming telemetry as the exception, and claimed DELIBERATELY NOT justified — an earlier draft blamed the poller's env, which `fire_toast` disproves, so the divergence is recorded as undecided rather than explained.
```

⚠ **Scope and verification, stated plainly so round 2 does not have to infer it.**
- **No gate result backs this push.** The two-tier run on the merged tree was **skipped by operator instruction**; a run was still in flight at push time and its verdict is not in this comment. The only suite evidence is `test_bar_status.py` **570 passed** in the fix worktree plus 271 passed across eight adjacent bar/nix files, and `nix-instantiate --parse nix/graphical.nix` clean. **Treat the merged tree as UNGATED.**
- The mutation matrix for finding 4 was **re-run independently** rather than accepted from the agent that produced the fix.
- Finding 5's fix touches `scripts/i3status-civitai` and `scripts/bar_freshness.py`, **outside this PR's original three files** — deliberate, on the N−1 argument, and called out rather than slipped in.

ZacxDev:
## Round 2 — findings fixed as `a56c4c71`, and 🔴 **THE LADDER STOPS HERE**

Round 2 was **not clean** (3 SHOULD-FIX, 4 NIT), so the findings-keyed rule would continue it.
**The attribution gate overrides that, and it has now fired.**

```audit-claims round=2 audited=0e9c469c21fb27040eceb1fdf455e8f83f178075..a56c4c71
1. Finding 1 (the PR BODY claimed `~/.cache/bar-status/runaways.toast-state` does not exist, so the duplicate never fired): claimed corrected in the body; claimed measured that the file EXISTS, `{"latched": false}`, written 2026-09-09 10:32:10, and that only a latch TRANSITION writes it — the same code path as the `fire(...)` call — so the rising edge happened. Claimed left explicitly unestablished: whether the toast reached the screen, because `dunstctl history` carries no `bar-status` appname for ANY poller toast.
2. Finding 2 (`bar/SKILL.md` said the missing spec is pinned by the NO_TOAST guard "+ the SOURCES ledger guard"): claimed corrected to name that guard ALONE, and to state that the ledger guard is NOT redundant cover; claimed measured that re-adding a spec fails the NO_TOAST guard alone at 1 failed / 569 passed while the ledger guard stays green, because `spec.get("source", spec_name) in names` is satisfied by a `runaways` spec.
3. Finding 3 (the bar-status-poll systemd unit `Description` enumerated seven sources, omitting runaways — the only stale enumeration a human sees at runtime): claimed corrected in `nix/graphical.nix`.
4. Finding 4 (a comment about unpinned numbers said SEVEN appeared "at four sites"): claimed corrected to seven times across three files, with the reason stated — four was round 1's finding count, carried forward without re-counting after the fix found three more. Claimed additionally recorded: `_COUNT_PROSE` is NOT pinned two-way, so a prose site it omits is unguarded.
5. Finding 6 (the new test's docstring said `test_fans_detail.py` "lives on main; this branch is based on a commit before it", which this PR's own merge falsified): claimed corrected to say it arrived via #1430 and is now on the branch.
6. Finding 7 (`i3status-dnd` does not exist; the script is `i3status-notifs`): claimed corrected at both `bar/SKILL.md` sites; claimed measured that the file is absent, `i3status-notifs` is present, and `graphical.nix:399` records the notifs pill as replacing the retired dndBlock.
7. Sub-nit (the "2026-08-05 ... six days later" arithmetic lands on 08-11 while the cap landed 08-12): claimed addressed by naming both dates rather than deriving one from a rounded prose date.
8. Claimed NOT FIXED, corrected here instead: finding 5 — the round-1 claims comment quotes `bar/SKILL.md` at 27,844 B, which is `origin/main`'s tree. The PR's own row makes it 29,345 B at this head. The direction (2.3x the advisory budget, ungated) is unchanged.
```

## 🔴 Why this is the last round — the ATTRIBUTION GATE, measured both times

Round 2 measured round 1's fix range (`3dd1491a..0e9c469c`) as **zero behaviour-bearing payload
lines** — every `+`/`-` in the four non-test files is a comment or docstring.

I measured this round's fix range (`0e9c469c..a56c4c71`) the same way: **one** non-comment line,
the systemd unit `Description` string. It changes no behaviour, and by the REVERT TEST it is not
payload — revert it and the PR's deliverable still ships.

**Two consecutive fix rounds that changed no behaviour ⇒ the ladder has left the PR.** Both rounds
found real things, and none of them was in the code — findings 1, 2 and 4 above were *introduced
or left standing by round 1's own fix*, which is the shape the gate exists to catch. Continuing
would audit the ladder, not the change.

**The findings-keyed rule does not override this**, and neither does the fact that round 2 was
unclean: the gate measures what the fixes CHANGE, never what the rounds FIND.

## Filed rather than fixed — one follow-up, with a closing condition

**`claude/skills/bar/SKILL.md` is 29,345 B against a 12,038 B advisory budget and NOTHING gates
it.** `skill-audit.py` reports every skill against the *browser* skill's constants, so its own
"ENFORCED" label is about browser's ceiling, not the file being audited. The size-gated docs are
`claude/RULES.md` plus browser, prune-skill, session-manager, handoff and clawgate — measured, and
the correction to round 1's finding 7, which had asserted a ceiling here that does not exist.
**Closing condition: a merged PR that either brings this body under budget (`/prune-skill bar`) or
adds a gating test for it.** Out of scope for a PR about removing a toast.

## Gate — read this before merging

- **Dev-host tier at this head:** `test_bar_status.py` + `test_doc_path_rot.py` → **646 passed**;
  `nix-instantiate --parse nix/graphical.nix` clean. Round 2 separately ran 570 + 308 passed at
  `0e9c469c`.
- 🔴 **The sandbox tier has NOT been run at `a56c4c71`.** Do it before merging, **one derivation at
  a time** — a combined invocation produces false failures.
- ✅ **The two unattributed merged-tree failures are now DISCRIMINATED AS ENVIRONMENTAL.** Run alone,
  one target per invocation, on the PR head: `scripts/dl-router/tests` **1020 passed** (454s),
  `scripts/browser-bridge/tests` **915 passed** (459s). The browser-bridge suite passed *while a
  second agent was executing the identical suite concurrently* — which is precisely the load that
  produces `assert 1.5600753400067333 < 1.5`. ⚠ Instrument note: two target dirs in ONE pytest
  invocation yields 8 collection errors — an artifact, not a finding; `run-tests.sh` runs each
  target separately for this reason.

ZacxDev:
## ⚠ Correction to my own "Gate" section above — the repo's merge policy changed mid-session

My previous comment ended with *"🔴 The sandbox tier has NOT been run at `a56c4c71`. Do it before
merging."* **That instruction is now contrary to this repo's own guidance**, which changed while
this ladder was running.

`CLAUDE.md` was rewritten by **#1429 (`a0839ec4`, 2026-09-09 13:23)**:

> 🔴 **THE LOCAL FULL-SUITE RITUAL BEFORE EVERY MERGE IS NO LONGER EXPECTED — this instruction
> used to say "run BOTH tiers yourself before merging" and that requirement is DELETED.**

The stated reason is that it produced **27–50 concurrent full-suite runs on one 24-core box**,
each running ~22,000 tests twice per change, while gating nothing — and that the contention made
the dev-host tier hit its own 3600s cap and produce *no verdict at all*, which is worse than not
running it. What replaces it: **read CI**, and run a **change-scoped subset** locally.

🔴 **That is independent confirmation of this PR's two unattributed failures.** I reported
`dl-router` → `database is locked` and `browser-bridge` → `assert 1.5600753400067333 < 1.5` on the
merged tree, and round 2 discriminated them as environmental by re-running each suite alone (1020
passed / 915 passed) — noting the browser-bridge suite passed *while a second agent ran the
identical suite concurrently*. The policy change names exactly that contention as the mechanism,
measured across the whole box rather than from this one PR. Two independent derivations, same
cause.

**So the standing evidence for this PR is:**
- **Change-scoped local runs** at head `a56c4c71`: `test_bar_status.py` + `test_doc_path_rot.py`
  → **646 passed**; `nix-instantiate --parse nix/graphical.nix` clean. At `0e9c469c`: 570 + 308
  passed, plus the full mutation battery.
- **CI is the automated signal and it is ADVISORY** — `tekton/devrc-pytests` and
  `tekton/devrc-nodetests` post on the head and do not block. Both are **pending** as of this
  comment. `gh pr checks 1435`. **Read them before merging; nothing will stop you if they are red.**
- The two failures above are attributed and are **not** this diff.

I am not running the full tiers again. Under the new guidance that is a judgement call rather than
a checklist item, and for a change that ships **zero behaviour-bearing lines** it is not the right
call.

PR #1436, opened 2026-09-09
docs(journald-migration): retract a runtime claim #1412 shipped, and close out the merged handoff

ZacxDev:
Follow-up to #1412 (merged as `f06b106f`). Three corrections found **after** that merge, two of them to claims #1412 itself shipped.

## 1. A runtime claim #1412 shipped is unsupported — retracted

`scripts/diagnose-nix-disk.sh`'s header said **"THIS TAKES TENS OF MINUTES"** and cited "section 3 alone had not finished after 13 minutes".

The run launched to verify that finally returned: **4h05m wall clock, having reached section 3 of 10.** So the shipped figure understates it by roughly an order of magnitude — but it is a **floor, not a runtime**, for two reasons that both matter:

- the box was running the test gate throughout and the process sat at **3% CPU** (95s user + 452s system across four hours) — I/O-starved, not working;
- the run did not stop of its own accord. It died on a syntax error at a line that is **fine on disk** (`bash -n` on the committed copy is clean; that line is a `df -hT /`). Bash reads a script by **byte offset** as it executes, and the file was edited five times during those four hours, shifting underneath the running interpreter. The artefact looked exactly like a defect in the script. The header now warns about it.

**The honest state, and the header now says so: no clean full run has ever been observed**, and the script's own closing `all 10 sections attempted` banner is untested end to end. The paragraph explicitly tells the next reader not to quote a runtime from it as if measured, and to replace it once someone completes a run on an idle box.

## 2. The merged handoff doc still described the PR as open

It said "PR #1412 OPEN" with a gate "IN FLIGHT" pointing at a superseded run. It now records the squash (verified by CONTENT — a squash is never an ancestor of its base), both tiers passing on the merged tree `53d8b962`, and the five-round audit whose ladder stopped on a stated criterion.

## 3. The rename list was wrong, and the fix is deliberately NOT a script

Next-step 4 listed **three** deprecated renames; only two are live — `services.dnsmasq.servers` is commented out at `configuration.nix:71`. The list came from an eval warning that was true when captured; `/etc/nixos` changed underneath it, and a too-narrow first grep missed it.

The two live ones (`gnome.tracker` → `tinysparql`, `gnome.tracker-miners` → `localsearch`) get the exact edit written into the doc and **no `apply-*.sh` script**. They are warning-only (`mkRenamedOptionModule`). A second root-privileged `/etc/nixos` mutator would ship a **copy of the rebuild-and-rollback trap that took five audit rounds to get right** — three of its four blockers introduced by the previous round's own fix. `RULES.md` → "One rule, one place": a duplicated predicate is typically wrong at N−1 sites in the same direction. The doc says to extract a shared `nix/system/lib/` helper first if a third such change ever appears. **That is a judgement to overrule, not a gap to fill.**

## Gate: NOT GREEN, and not because of this change

🔴 **Do not merge this on my say-so — the full pytest tier could not be obtained.** Two consecutive runs hit the **3600s timeout** and were `Terminated` (exit 124 / RESULT exit=143). That is load, not an assertion, by the documented discriminator:

- node tier **PASS 1449/1449** on both runs;
- every pytest target that completed before the wall **passed**;
- load average went **72 → 92** across the two runs, with 257 concurrent `python3.12` processes and 54 other Claude session wrappers on the box;
- the same suite passed twice earlier today on this branch's base (`collected=21164 failed=0`).

The first timeout was **my fault** — I ran a `/nix`-walking diagnostic (~80M inodes) concurrently with the gate. I stopped it by resolved PID and confirmed the PIDs were gone rather than zombies. The second timeout happened with that gone, on other sessions' load alone.

What I *can* vouch for: this diff is one comment block in a `.sh` and one `.md`, so it cannot reach Python. Targeted runs of every test that reads those paths — the three content gates, the CI-claim marker test, and the journald suite — are **254 passed**. That is a narrower claim than "the gate passed", and I am not substituting one for the other.

**Someone should run `scripts/gate.sh --tier both` on a quiet box before merging.**


ZacxDev:
### Targeted evidence completed — 861 tests, still NOT a substitute for the gate

The PR body listed 254. The remaining tests that read `claudedocs/` have now finished too:

```
test_closing_condition_single_source.py + test_drift_check.py
  + test_claim_work.py + test_claude_sessions.py     607 passed in 427.40s
content gates + CI-claim marker + journald suite     254 passed in  47.31s
                                                     ---
                                                     861 passed, 0 failed
```

That is every test in `scripts/tests/` that reads either path this diff touches (`scripts/diagnose-nix-disk.sh`, `claudedocs/`), found by `grep -rl`.

🔴 **This still is not "the gate passed."** It is a claim about the targets I could enumerate as reachable, on one tier, run individually — not the full suite on the merged tree, and not the sandbox tier at all. The full `pytest` tier remains unobtainable on this box: two consecutive 3600s timeouts under load 72→92 from other sessions.

**The merge condition is unchanged: run `scripts/gate.sh --tier both` on a quiet box, and `nix build .#checks.x86_64-linux.{pytests,nodetests}` one at a time.** Neither has been done for this branch.

PR #1427, opened 2026-09-09
fix(audit-pr): a NIT-only round stops the ladder — and blocker-only is rejected

ZacxDev:
## What

The audit ladder's stop rule now reads:

> Rounds continue **only** while the previous round produced a **deploy-blocking or should-fix** finding that required a fix.

plus an explicit carve-out — **a round whose findings are all NITs is a stopping round** — and, pinned beside it, the rejection of the obvious next edit.

## Why

`scripts/audit-dispatch.py` has shipped a `nit-is-not-a-finding` clause into every brief, while `SKILL.md`'s Output section solicits findings in three tiers *including* nit. An auditor reading the skill reports nits as findings, and the ladder spends another round on them. The rule now says what the brief already enforces.

## Why not "high-severity only"

That was the starting proposal and the measurements reject it:

| case | deploy-blockers | what a blocker-keyed ladder ships |
|---|---|---|
| `homelab-infra` #702 | **zero across six rounds** | four of six findings were a false claim the previous round's own fix had written |
| devrc #804 | rounds 5-7 each "safe to merge" | a real defect per round, last a latch vacuous both ways |

Both end at round 1 under a blocker-keyed rule. So **nit is the only severity that cannot extend a ladder**, and #702 is recorded next to the carve-out so the bar cannot be quietly raised later.

`claude/RULES.md` is unchanged and stays true: a nit is not a finding, so "produced a finding that needed fixing" already excludes a nit-only round.

## Coverage — both new guards watched RED on pre-change code, isolated

| control | result |
|---|---|
| `SKILL.md` restored to `origin/main` | **2 failed** (stop-rule body pin, carve-out pin), 13 passed |
| `"nit-is-not-a-finding"` renamed one char in `audit-dispatch.py`, nothing else | **1 failed** (seam test alone), 14 passed |
| restored | 15 passed |

The seam test asserts the clause **id** the skill cites still exists in the dispatcher — deliberately not the clause text, which `test_audit_dispatch.py` owns.

## Gate status — NOT clear to merge

- **node tier: PASS** — 1449 tests, floor 1367.
- **pytest tier: INCOMPLETE** — main suite PASS (1020), then `scripts/browser-bridge/tests` hit the 3600s budget under load average 66 with five concurrent pytest runs from other sessions on the box. The same directory passes in **420s standalone on a pristine `origin/main` worktree**, so the timeout is load, not this diff.
- **nix-check tier: NOT RUN** — a contended `nix build` reports false failures by documented measurement.

[...]

PR #1440, opened 2026-09-09
feat(audit-pr): round 0 asks whether the change should EXIST — the nine axes only ever asked whether it is correct

ZacxDev:
## What this is

`/audit-pr` audits for risks, regressions, assumptions, gaps, bugs, issues, behaviour changes, leaks and second-order consequences. **Every one of those asks whether the change is CORRECT. None asks whether it should EXIST** — and an audit scoped to a diff will not raise it on its own, which is how a 145 KB webhook listener nothing had ever run survived every round that read it.

This adds **ROUND 0**, a requirements & deletion pass, as a section of the skill and a `--round 0` mode of the assembler. It is the only round that can conclude *"close this PR, do not audit it."*

It is an integration of "the algorithm" (question the requirement → delete → simplify → accelerate → automate) into the existing ladder.

## Why it is a separate round and not checklist items 10–11

The ordering **is** the mechanism. Steps 3–5 spent on something step 2 would have deleted is the waste the pass exists to catch — so appending "can this be deleted?" after the nine correctness axes reproduces exactly the failure being prevented.

Round 0 therefore emits its section **instead of** the nine axes, not before them. An auditor handed both answers "is it correct?" first. Both halves are asserted, and the negative half (`CHECKLIST not in out`) is as load-bearing as the positive one.

## What it deliberately does NOT add

Steps 3–5 grow no machinery:

- step 3 routes to `/simplify` and `/code-review`
- step 2 routes to `/prune-skill` and `/prune-memory`
- step 2's tie-breaker is the **revert test the attribution gate already defines**

One rule, one place. Automation is step 5 for a reason.

## It cannot move the ladder

Stated explicitly in the section: round 0 is **not** a finding for the findings-keyed stop rule, it cannot end a ladder, and it cannot license skipping a round. A new pass that quietly becomes a stop condition is the *"wider on one axis, narrower on another"* shape this skill hunts for.

## It ships with its own retirement condition

> Run it on the next 3–5 PRs and count the ones where its verdict CHANGED what happened. **If that count is zero, DELETE this section.**

The assembler inlines that condition into the brief, so the auditor sees it. This PR is trial #1.

## Two defects fixed along the way

1. **Found by a live smoke test against #1427, not by the suite.** `--round 0` printed the delta-anchor warning (*"Using it anyway"*, *"check you are not re-auditing a round that already ran"*) over a claims block round 0 never consults — `render_claims` returns `""` below round 2 and the range comes from the base. It fired on **every** round 0 of any PR with a ladder, on the one stream carrying the missing-intermediate-block warning. Scoped to round 0 only; round 1 keeps both warnings and a positive control pins that.

2. **A negative round assembled a round-1 brief in silence.** Every gate is spelled `>= 2` or `< 2`, so negatives fell through the `< 2` side. Harmless while `0` was meaningless; not harmless now that `0` names a different pass. Refused, rc 4.

## Test evidence

6 **invariant guards**, not regression coverage — `--round 0` exists at no base in `RED_AT_BASE_REFS`, so a red there would only restate that the flag was added. Evidence is the mutation battery plus in-test controls:

| mutant | killed by |
|---|---|
| M1 guard-off (`== 99`) | `INSTEAD_OF_the_nine_axes`, unreadable-skill |
| M2 title → round 1 | `INSTEAD_OF_the_nine_axes` |
| M3 ledger branch off | `ledger_rather_than_claiming_round_one` |
| M4 negative bound loosened | `negative_round_is_refused` |
| M5 reader's answer unused | `INSTEAD_OF_the_nine_axes`, unreadable-skill |
| M6 heading renamed | `section_the_script_reads_is_the_one_the_skill_ships` |
| M7 suppression off | `does_not_warn_about_a_claims_block` |

- **negative control:** unmutated GREEN (133 passed)
- **positive control:** an *unrelated* pre-existing guard mutated → RED, proving the harness can go red at all

Run under `PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`, restored from a `cp -a` copy between mutants.

The **seam guard** is the one that matters most: `_read_round_zero` anchors on a heading literal and the skill owns that heading. Both sides are hermetically tested and neither notices a rename — the script's tests inject a fixture reader on purpose, the skill's tests pin prose the script never parses. So one test reads the real file with the real reader.

## 🔴 Gate status — INCOMPLETE, read before merging

Measured on the isolated worktree at each named sha, not on a shared checkout.

| tier | result |
|---|---|
| node (`scripts/gate.sh`) | **PASS** — 5 suites, 41 files, 1449 tests, 0 fail (run twice: `d16bfd7a`, `97b03d8a`) |
| pytest — targeted (`test_audit_dispatch.py` + `test_audit_ladder_stop_rule.py`) | **PASS** — 151 passed, at `dcd84c73` |
| pytest — FULL dev-host tier | 🔴 **NO VERDICT** — timed out at the 3600 s cap (`exit=124`, `RESULT: FAIL (exit=143)` = SIGTERM). Not an assertion failure |
| `nix build .#checks.x86_64-linux.pytests` | **NOT RUN** |
| `nix build .#checks.x86_64-linux.nodetests` | **NOT RUN** |

**Why the missing tiers are missing, stated plainly:** this box has been continuously saturated by other agent sessions for the whole life of this PR — load average 88–96 at the first attempt and still **60 with 36 concurrent `run-tests.sh` runs and another `nix build .#checks` in flight** nine hours later. A second concurrent check derivation is the store contention `CLAUDE.md` says yields *false* failures, so it was never started; a red from it would have been evidence about the box, not about this diff. The one full run that did start hit the hour cap.

**Do not merge on what is green above.** The node tier and a targeted pytest selection are not the gate. The sandbox tier is the one Tekton runs and the one the dev-host tier is structurally blind to (it builds from a `cp -r` store copy with no `.git`, so anything keyed on being a git checkout evaluates differently there).

**What remains:** `scripts/gate.sh --tier both` to completion, then the two `nix` check derivations **one at a time**, on a quiet box.

### DIFF
--- a/claude/skills/audit-pr/SKILL.md
+++ b/claude/skills/audit-pr/SKILL.md
@@ -93,7 +93,7 @@ audit scoped to a diff will never raise it on its own: that is how a 145 KB webh
 nothing had ever run survived every round that read it. Round 0 is the only round that can
 conclude **close this PR, do not audit it**.
 
-The order is the mechanism, not a preference — steps 3–5 spent on something step 2 would have
+The order is the mechanism, not a preference — steps 3–4 spent on something step 2 would have
 deleted is the waste this exists to catch. Work them in order; do not skip ahead.
 
 1. **Question every requirement, and NAME its author.** For each behaviour the diff introduces,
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Round 0 trial record — PR #1440 is trial #1

`round 0 · ran: 1 · changed the outcome: 1`

Round 0 was dispatched **blind** against this PR's own diff (`d16bfd7a`), per the skill's framed-audit rule. It returned findings on its first run, and its verdict — `requirement questioned` — changed what shipped.

**Ledger it produced:** `requirements: 21 (unattributed: 8) · deletion candidates: 5`

### What it caught that the nine correctness axes would not have

Two were verified empirically before being accepted, and both were defects the change's own tests could not see:

1. **Round 0 could move the ladder** — the one thing its 🔴 sentence forbids. `if args.emit_claims:` sits below every round gate and had none of its own, so `--round 0 --emit-claims` emitted a valid `audit-claims round=0` block that a later `--round 2` would anchor its delta on. Measured at `d16bfd7a`: `audit-claims round=0 audited=d16bfd7a..d16bfd7a`, rc 0. The prose was a comment, not a guard. Fixed in `97b03d8a`, rc 4, with a positive control pinning that round 1 still emits.

2. **The `## THE CHECKLIST` heading is load-bearing and read as decoration.** `_read_round_zero` captures to the next `## `; delete that heading and the capture runs 3,367 → 4,057 chars, pulling the nine axes into the round-0 brief and silently defeating the entire INSTEAD_OF design. The seam guard stayed **green** through it — it only checked steps were *present*, which stays true at the head of a longer capture. Now asserts the checklist is *absent*; mutant M9 confirms it catches the deletion the old guard slept through.

Plus: a false ledger provenance (`--numstat` cannot identify *re-added* lines, so the add-back percentage was undefined by the instrument it named), a retirement condition that could not distinguish "ran and was useless" from "nobody ran it", step 3 pointing a read-only auditor at mutating skills, and three prose counts that disagreed with each other and the tree.

### Operator decisions taken (`dcd84c73`)

- **Step 2 no longer uses the revert test.** It answers "payload or scaffolding", not "should this exist" — applied to a PR's own payload it always says *keep*, applied to surrounding code it always says *deletable*. Replaced with the three questions the cited incident actually turned on: *is it running? has it ever caught anything? does something else already check it?* The 145 KB listener this section cites had no token configured on either host — question (a) is what killed it, and the old text could not have reached it.
- **Steps 4 and 5 merged into one ordering rule.** Nothing read them; the ordering claim is the half that does work.

### Honest note on what this proves

One trial is one trial. It says round 0 found real things on a PR **it was the subject of**, which is the most favourable possible case — the auditor had a fresh, self-referential target. Trials 2–5 need to be ordinary PRs before the retirement condition means anything.

PR #1442, opened 2026-09-09
feat(nixos): clear 5 of the workbench's 6 nixos-rebuild deprecation warnings — and why the sixth deliberately stays

ZacxDev:
Adds `nix/system/apply-nixos-option-renames-2026-09-09.sh` — a staged sudo script that clears **5 of the workbench's 6** `nixos-rebuild` evaluation warnings — plus 31 hermetic tests.

## What it changes

| # | from | to |
|---|---|---|
| 1 | `services.dnsmasq.servers` | `services.dnsmasq.settings.server` |
| 2 | `services.gnome.tracker.enable` | `services.gnome.tinysparql.enable` |
| 3 | `services.gnome.tracker-miners.enable` | `services.gnome.localsearch.enable` |
| 4 | `xorg.xrandr` | `xrandr` (i3 `extraPackages` **only**) |
| 5 | — | `xdg.portal.config.common.default = "*"` |
| 6 | `wineWowPackages` | **deliberately left alone — see below** |

## Measured, not assumed

Evaluated the patched config against root's channel:

* **Patch A (1–4) is a no-op.** 0 derivations, same system derivation. `settings` is a key-value format with `listsAsDuplicateKeys = true`, so the resolver list renders as identical repeated `server=` lines. `xorg.xrandr` and `xrandr` are the same derivation.
* **Patch B (5)** builds exactly **5 trivial derivations** and adds exactly one file, `/etc/xdg/xdg-desktop-portal/portals.conf`. Realised its content: `[preferred]` / `default=*`.
* Warning count goes **6 → 1**, and the survivor names `wineWowPackages`.

Edit 4 touches only the package-list entry — `sessionCommands` already uses the correct `${pkgs.xrandr}` and is untouched (asserted).

## Why warning 6 stays

`wineWowPackages` → `wineWow64Packages` is **not a rename**. `wineWow` is a genuine 32+64-bit pair; `wineWow64` is a single 64-bit binary relying on upstream's new WoW64 thunking. Different mechanism, different bug surface — and the config's own comment records that the current spelling was tuned until Lutris / Ubisoft Connect worked. Its nixpkgs alias also carries no `# Added <date>`, so `remove-old-aliases.py` structurally skips it: it is the least likely of the six to ever break. One warning is the correct steady state; a test pins the prose that says so.

## Two measured hazards it handles, both silent

1. **`NIX_PATH` poisoning would roll the system backwards.** The interactive `NIX_PATH` starts with `$HOME/.nix-defexpr/channels`, searched *before* the explicit `nixpkgs=` entry, and that channel is an **older** nixpkgs than root's — which the running system was built from. `sudo -E nixos-rebuild switch` (the natural form, since it preserves the PATH hazard 2 needs) evaluates against the older tree and rebuilds ~34 derivations.

   The script **exports an explicit `NIX_PATH`** pinned to root's channel rather than unsetting. `unset` is *not* equivalent: NixOS's `nix.nixPath` default is what supplies the `nixpkgs=` alias, and root's channel directory is named `nixos`, not `nixpkgs` — so with it unset, `<nixpkgs>` resolves via nix's built-in default whose first component is `$HOME/.nix-defexpr/channels` again. Under `sudo -E` that reintroduces the exact hazard.

2. **root has neither `python3` nor `dig`** in `/run/current-system/sw/bin`; both come from the user profile. `python3` is preflighted **before the backup**, using shell builtins only (the branch fires precisely when PATH is impoverished), and prints the exact `sudo env "PATH=$PATH" bash …` remedy rather than half-applying. A missing `dig` degrades the two DNS checks to printed commands instead of skipping them silently.

## Public-repo safety

The script **never spells a resolver address**. It captures the bracketed `servers = [...]` list out of the live config and re-emits it verbatim, so:

* the `/docker.io/` pin cannot be dropped or retyped from a stale copy, and
* no address literal lands in this public repo — `test_no_public_ips.py` needed no allowlist entry.

The test fixture is a structural likeness of the live config using TEST-NET-3 (`203.0.113.0/24`) and RFC 1918 addresses only.

## Refuses on drift

Every edit must match **exactly once** or the script exits 2 having written nothing and tells the operator to apply by hand. Six drift cases are tested. Anchors are line-anchored past whitespace only, so a `#` kills them — which matters here: the live config carries a large commented-out `xdg.portal` experiment including `#xdg.portal.config.common.default = "*";`. A plain substring idempotency check matched that dead line and made the script skip patch B entirely. Caught on the first fixture run against the real config; there is a named regression test.

## Verify section

Commands **with expected values**, not "it worked":

* exactly one `evaluation warning:`, naming `wineWowPackages`
* `dig +short @127.0.0.1 workbench.lan` → the node's own LAN address
* `dig @127.0.0.1 registry-1.docker.io A` → TTL in the **tens** of seconds. ~42,000,000 means the LAN router's 487-day record is back, the pin did **not** survive, and the operator must roll back
* `/etc/xdg/xdg-desktop-portal/portals.conf` contains `default=*`
* printed for the operator (root cannot see them): `systemctl --user list-unit-files | grep -E 'tinysparql|localsearch|tracker'`, and `sudo nixos-rebuild dry-build 2>&1 | grep -c '^evaluation warning:'` → 1
* rollback: `sudo nixos-rebuild switch --rollback`, plus the timestamped `.bak-*`

## Testing

31 hermetic tests. Mutation-swept with `PYTHONDONTWRITEBYTECODE=1`; each mutant was confirmed to fail with the targeted test's own assertion:

| mutant | result |
|---|---|
| comment walk-back loses its trailing space (swallows `#servers =`) | killed |
| portal idempotency back to a plain substring test | killed |
| drift check accepts >1 `servers` lines | killed |
| `rc=0` removed before the `python3` heredoc | killed |
| `xorg.xrandr` replaced by a global `xorg.` swap | killed |
| gnome rename target typo (positive control) | killed |
| byte-delta "nothing else moved" guard deleted | **survived — equivalent** |

The last one is redundant with the external line-level test given a correct implementation, so it is an equivalent mutant. It is proven **reachable** by hand: mis-accounting the xrandr edit by 5 bytes makes it fire with its own message and refuse, and the same bug is written out silently with the branch removed. Labelled as defence-in-depth in the source.

The `xorg.` mutant originally **survived a fully green sweep** — the fixture had only one `xorg.` occurrence, so a global swap was equivalent. Fixed by adding a sibling `xorg.xkill` entry that must survive untouched.

Run it: `sudo bash nix/system/apply-nixos-option-renames-2026-09-09.sh` (or `--dry-run` first, which needs no root).

### DIFF
--- a/nix/system/apply-nixos-option-renames-2026-09-09.sh
+++ b/nix/system/apply-nixos-option-renames-2026-09-09.sh
@@ -1,0 +1,550 @@
+#!/usr/bin/env bash
+# Clear 5 of the workbench's 6 `nixos-rebuild` evaluation warnings.
+#
+#   sudo bash nix/system/apply-nixos-option-renames-2026-09-09.sh
+#   bash nix/system/apply-nixos-option-renames-2026-09-09.sh --dry-run   # no root needed
+#
+# ── WHY ──────────────────────────────────────────────────────────────────────
+#
+# MEASURED 2026-09-09 on the workbench (NixOS 26.11 "Zokor", plain channels, not
+# flakes). `nixos-rebuild dry-build` emits SIX lines beginning
+# `evaluation warning:`. Five are pure option/attribute RENAMES that nixpkgs has
+# already aliased; the sixth is not a rename at all and is deliberately left.
+#
+#   1. services.dnsmasq.servers        -> services.dnsmasq.settings.server
+#   2. services.gnome.tracker.enable   -> services.gnome.tinysparql.enable
+#   3. services.gnome.tracker-miners.enable -> services.gnome.localsearch.enable
+#   4. xorg.xrandr                     -> xrandr
+#   5. xdg-desktop-portal 1.17 wants an explicit `xdg.portal.config`
+#   6. wineWowPackages                 -> wineWow64Packages    ← NOT TOUCHED
+#
+# ── PATCH A (items 1-4): PROVEN BYTE-IDENTICAL ───────────────────────────────
+# Verified 2026-09-09 by evaluating the patched config: the SAME system
+# derivation (`q97dj70ph64md5vjsf0x757jq6vjbmni`), 0 derivations built, no new
+# generation, no unit restarts. These four are the alias being spelled out.
+#
+#   * `settings` is a `pkgs.formats` key-value format with
+#     `listsAsDuplicateKeys = true`, so the list becomes repeated `server=`
+#     lines in dnsmasq.conf — byte-for-byte what `servers` already produced.
+#   * `xorg.xrandr` and `xrandr` are the same derivation. NOTE the config
+#     ALREADY uses the correct `${pkgs.xrandr}` in `sessionCommands`; only the
+#     i3 `extraPackages` entry is stale, and this script changes only that one.
+#
+# ── PATCH B (item 5): 5 TRIVIAL DERIVATIONS, ONE NEW FILE ────────────────────
+# Adding `config.common.default = "*";` inside the existing `xdg.portal` block
+# creates `/etc/xdg/xdg-desktop-portal/portals.conf` (which does not exist on
+# this host today — the directory is empty). `"*"` reproduces the pre-1.17
+# behaviour: "use the first portal that implements the interface", which on this
+# host is the single installed `xdg-desktop-portal-gtk`. It is not a behaviour
+# change; it is writing down the behaviour that was previously implicit.
+#
+# ── WHY WARNING 6 STAYS: wineWowPackages IS NOT A RENAME ─────────────────────
+# 🔴 DO NOT "FINISH THE JOB" BY ADDING IT. `wineWowPackages` is a genuine 32-bit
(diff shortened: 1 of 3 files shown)


ZacxDev:
Follow-up commit `871bcf6c`: the repo's own launcher ledger (`test_no_real_launchers.py::test_every_path_clobbering_site_is_pinned`) went red on `1c314fe6` and it was right — the python3-preflight test REPLACES `PATH` instead of prepending, dropping the suite-wide no-launcher stub dir. Reproduced standalone in 6.4 s, so it was a deterministic ledger violation, not the load-starve that truncated the first full-gate run.

Fixed by pointing the clobber at a dedicated **empty** directory (strongest justification shape, matching `test_claude_log_rotate.py` / `test_session_stamp_seam.py`) and adding the pin entry. Emptiness makes the assertion *stronger*: the preflight's error branch is builtins-only so it can speak in an impoverished environment, and an empty PATH is what makes that observable. Negative control on the new pin: swapping the needle makes the ledger report `still clobbers PATH but not in the pinned shape`, so it is live, not a rubber stamp.

PR #1445, opened 2026-09-09
feat(gate): change-scoped test runs, and a narrowed run that names the work it left for CI

ZacxDev:
## The problem

The pytest tier's median wall time is **20.1 min over 237 real runs** (p90 38 min). No suite is
slow — the box is: **~27 concurrent agent sessions each run the FULL suite** on 24 cores, and runs
bucketed by overlap go **14.5 min (alone) → 48.9 min (6+)**.

The lever is **fewer full suites**: make an iteration run cheap and change-scoped.
`run-tests.sh --targets` already existed and buys little, because the target that matters is a
monolith — `scripts/tests` is **13,510 of 21,947 collected tests (62%)** in 185 files.

🔴 **CI is the only automated check.** Branch protection is off and the local full-suite ritual is
deleted, so `tekton/devrc-*` are advisory and land minutes after a push. That decides what a
narrowed run is *for*: not an approximation of a gate, but the inner loop — whose most useful
output is a statement of **which work it handed to CI**.

---

## Design chosen: (B) a change-scoped mode. (A) rejected, with measurement.

Re-measured at load 58 (the auditor flagged the originals as unverified):

| what | tests | wall | CPU |
|---|---|---|---|
| `pytest scripts/tests --collect-only` | 13,518 | **47.7s** | 27.5s |
| `pytest scripts/tests/test_drift_check.py --collect-only` | 461 | **0.32s** | — |
| one pytest invocation **with the 4 guard plugins** (×10, load 68) | 0 | **1.46s** | — |

**Collecting the monolith costs ~48s before a single test runs.** *Target*-granularity selection
avoids none of it — a sub-target still collects itself. Only **file** granularity does:
47.7s → 0.32s, **149×**.

### Why (A) — splitting `scripts/tests` into sub-targets — was rejected

1. **At file granularity it is a full-suite regression.** ~185 targets (one per file), each a
   separate pytest invocation at a measured **1.46s** with the guard plugins → **~270s added**
   to a run whose sandbox wall time is 15m33s — ~29%, landing on **CI**, the actual bottleneck.
   Plus 185 mutually-pinned `TARGET_FLOORS` entries. (The first measurement gave 0.95s at lower
   load; a re-measure can only come in higher, which strengthens the rejection.)
2. **At coarser granularity it does not deliver the win.** Ten sub-targets add ~15s but a scoped
   run still collects ~1,350 tests instead of 0.32s — an order of magnitude worse — for a change
   that physically **moves 185 files** into subdirectories, altering `sys.path`/conftest
   resolution for every one.
3. **It fragments the xdist pool where the win is concentrated.** `--dist loadfile` assigns whole
   FILES to workers, so a pool of 3 files cannot use 8. ⚠ Worker count is host-dependent — this
   run used **8** (main's new `_devrc_cpu_budget`), earlier runs used 4; fragmenting hurts more
   the more workers there are.
4. **(B) touches no full-run code path.**

### What (B) is

* **`run-tests.sh --files "<paths>"`** — run named test FILES inside their owning targets. **One
  pytest invocation per owning target**, never per file, so per-target guard accounting and
  `_markers_ok`'s worker bound are unchanged. `--dist loadfile` untouched; nesting still
  serialises on `PYTEST_CURRENT_TEST`.
* **`scripts/scoped-tests.sh`** — maps a git diff → test files, reading the declared target list
  from `run-tests.sh --check-targets` (never a second copy). The mapping stays *outside* the
  runner, as that file's `--targets` header requires.
* **`SCOPE: FULL|NONE|PARTIAL|SCOPED|UNKNOWN`** on every exit path of both runners.
* **The CI-gap block** — the unrun targets BY NAME, the unrun files inside the targets that did
  run, the skipped families, and the node tier. Every number derived from what the run selected.

---

## Round-1 audit: status per finding

| # | finding | status |
|---|---|---|
| 🔴 1 | GUARD 7's `NOLAUNCH_ACK` failed a scoped run of `scripts/tests` | **FIXED** — required direction suspended for a slice, PERMITTED direction kept |
| 🔴 2 | GUARD 2's `EXPECTED_SKIPS` failed a scoped run of `scripts/signal/tests` | **FIXED** — skip TOTAL suspended, UNPINNED check kept |
| 🔴 3 | the suite was structurally blind to both | **FIXED** — `runner_patch` gained `expected_skips`; 6 new tests build the combined state, each with a non-scoped control |
| 🟡 3 | three zero-test invocations printed `SCOPE: FULL` + `RESULT: PASS` | **FIXED** — new `NONE` state; NONE wins over narrowing |
| 🟡 4 | "no `--files` selection can name them" false for SHELL_TESTS | **FIXED** — refused by family, naming it |
| 🟡 5 | CLAUDE.md "every GUARD still applies" false three ways | **FIXED** — states which ledgers are suspended and why |
| 🟡 6 | gate.sh's refusal narrower than the hazard | **FIXED** — all four vars refused |
| 🟡 7 | one 53-test file, ~7 CPU-min, one xdist worker | **FIXED** — split into 5 files: **187.6s (load 58) → 59.5s (load 79.75)**, ~3.1× |
| 🟢 | `_scope_line` first-match | **FIXED** — last match, matching `gate.sh` |
| 🟢 | `scoped-tests.sh` honoured ambient `DEVRC_TARGETS` | **FIXED** — refused (wrong *diagnosis*, not wrong verdict) |
| 🟢 | `SCOPE:` has no structural grammar reservation | **OPEN** — see "left open" |

**The insight I got wrong the first time:** GUARD 3's floors, GUARD 7's ack and GUARD 2's skip
total are all expectations about a target **run in full**. I gave the treatment to one of three.
Reproduced, both with every selected test PASSING:

```
--files scripts/tests/test_ship_detect_role.py   → 12 passed, "intercepted NOTHING", exit 1
--files scripts/signal/tests/test_search.py      → 15 passed, "2 of 3 pinned entries apply", exit 1
```

Green iff the selection happened to include one of three seam files out of 181, or one specific
file out of a 920-test target. Both printed advice that would have disabled a guard repo-wide or
reddened the FULL gate if followed.

🔴 **Only the whole-target half is suspended, and suspension is ANNOUNCED** under
`whole-target expectations SUSPENDED by this SCOPED run`. Kept, because they are per-observation:
GUARD 7's PERMITTED direction, GUARD 2's UNPINNED check, GUARD 8, GUARD 10, and every
plugin-marker count.

Verified on the rebased head — green **with the suspension stated**, not with the guard inert:

```
--files scripts/tests/test_ship_detect_role.py
  12 passed · "ACKNOWLEDGED — required direction SUSPENDED" · RESULT: PASS (exit=0)
--files scripts/signal/tests/test_search.py
  15 passed · "GUARD 2's skip TOTAL … The UNPINNED-skip check still ran." · RESULT: PASS (exit=0)
```

---

## Wall-time deltas

| run | before | after |
|---|---|---|
| **full suite** (sandbox `pytests`, base `eeea9025`) | 21,947 collected, PASS | **15m33s**, 21,947 collected / 21,945 passed / **0 failed**, load ~38, `-n 8` |
| **single-subsystem change** (`dl-router/server.py` → 3 files, 59 tests) | 20.1 min median | **49s wall / 16.8s CPU** |
| one-file change (13 tests) | 20.1 min median | **13s** |

⚠ **The no-regression claim rests on a STRUCTURAL reading, not a benchmark.** An earlier revision
offered a millisecond-scale A/B; its runs overlapped and inverted, and the auditor accepted the
conclusion while rejecting the measurement. Dropped. With `SCOPED_MODE=0` the diff adds two
`echo`s, one scope-resolution `if`, an array copy, a `shift`+`paths` per target and two
`${arr[@]+…}` guards; `_print_ci_gap` is not called at all.

A second lever: a `--targets` subset of ONE target measured **172s**, of which the target was 17s
and the five `SHELL_TESTS` were **101s**. No `--files` selection can name those, so SCOPED mode
skips them and says so.

## What a scoped run does NOT prove

1. **Not that the change is safe.** CI is the signal.
2. **Not that the selected files cover the change.** The mapping finds tests that NAME what you
   changed (full path first, then basename **within the changed file's own subsystem** —
   measured: `server.py` matched 27 files repo-wide vs 3 in `scripts/dl-router/tests`). Coverage
   through an import chain is invisible; unmapped files print under `🔴 UNMAPPED`.
3. **Not that the unselected files still pass** — no target floor was checked.
4. **Not the hook/shell families**, and **nothing about the node tier**.
5. **Not the three suspended whole-target ledgers** — that is CI's job, and the run names them.

---

## Verification

**Base gated on: `eeea9025`. Head: `c912bc95`.** Both derivations built **one at a time**:

| tier | verdict |
|---|---|
| sandbox `pytests` | `SCOPE: FULL (28 of 28 hermetic target(s))` · `RESULT: PASS (exit=0)` · collected=21947 passed=21945 skipped=2 **failed=0** |
| sandbox `nodetests` | `SCOPE: FULL (this runner has no selection flag…)` · `RESULT: PASS (exit=0)` |

**Red/green matrix** — the 5 scoped suites, planted into a `git archive` of the pre-branch base
`8970142d`: **45 RED, 9 green**. All 9 are labelled INVARIANT GUARD or POSITIVE CONTROL in their
own docstrings and are not counted as regression coverage.

**New-fixture matrices** (the audit's 🔴-3 — fixtures that load a REAL ack entry and a REAL pinned
skip), against the pre-fix commit `e3166a99`:

| test | pre-fix | post-fix |
|---|---|---|
| `test_a_scoped_run_suspends_GUARD_7s_required_ack_direction` | **RED** | green |
| `test_a_scoped_run_suspends_GUARD_2s_skip_TOTAL` | **RED** | green |
| `test_a_FULL_run_still_ENFORCES_GUARD_7s_required_ack_direction` | green | green (control) |
| `test_a_FULL_run_still_ENFORCES_GUARD_2s_skip_TOTAL` | green | green (control) |
| `test_a_scoped_run_still_FAILS_an_UNPINNED_skip` | green | green (control) |
| `test_a_scoped_run_still_FAILS_an_unacknowledged_real_launcher` | green | green (control) |

The four controls pass both ways **by design** — they are what separates *suspended for a slice*
from *switched off*, which "the scoped run is green" alone cannot distinguish.

**Mutants** — 19 across three rounds, each mutating the narrowest expression that can be wrong.
Round 2 (this round's guards):

| # | mutation | outcome |
|---|---|---|
| N1 | GUARD 7 required-ack applied to a slice again | KILLED — `…suspends_GUARD_7s_required_ack_direction` |
| N2 | GUARD 2 skip TOTAL applied to a slice again | KILLED — `…suspends_GUARD_2s_skip_TOTAL` |
| N3 | GUARD 7 **PERMITTED** direction also suspended (over-correction) | KILLED — `…FAILS_an_unacknowledged_real_launcher` |
| N4 | GUARD 2 **UNPINNED** check also suspended (over-correction) | KILLED — `…FAILS_an_UNPINNED_skip` |
| N5 | `--check-targets` reports FULL again | KILLED — `…reports_NONE_never_FULL` |
| N6 | node `--check-suites` reports FULL again | KILLED — same |
| N7 | `--files` accepts a SHELL_TEST again | KILLED — `…refused_by_family_not_misdiagnosed` |
| N8 | gate.sh refuses only `DEVRC_TARGETS` again | KILLED — `…refuses_an_ambient_devrc_targets…` |
| N9 | `scoped-tests.sh` honours ambient `DEVRC_TARGETS` again | 🔴 **SURVIVED** → test added → N9′ **KILLED** |
| — | CONTROL (behaviour-preserving) | **SURVIVED** (correct) |

Round 1's ten are unchanged, including its own survivor: M6 stayed green with the `--files` empty
check deleted, because the run fell through to the target-subset block whose empty check emits a
**byte-identical first line**. Fixed by asserting the follow-up line only that guard emits.

🔴 **Two mutants found gaps in my own work** (M6, N9). Both are in the table because a battery that
only confirms what you believed is not a battery.

**Sandbox-only defects fixed** — none reachable on the dev host:
* `test_every_pytest_target_gets_the_detector` selected the pytest line by requiring `$d`, i.e. by
  the *name* of a shell variable. `--files` renamed it and the population went **empty**; only its
  own `assert pytest_lines` positive control made that a red. Selector widened; re-mutated,
  deleting `-p testlib.gitenv_plugin` fails it with **its own** message.
* `SCOPED_MODE: unbound variable` took four tests in `test_conditional_skip_pins.py` red — that
  file extracts the GUARD 2 block and runs it standalone under `bash -uo pipefail`. The block
  above carries a comment warning about exactly this for `${ONLY_TARGETS:-}`; I reintroduced the
  defect its comment documents.
* a fixture wrote `#!/usr/bin/env bash`, which the sandbox has no interpreter for.

**#1429's own two-way pins caught three more** after the rebase, each naming its fix: an
undocumented `DEVRC_GATE_ALLOW_AMBIENT`, its absence from `_AMBIENT_GATE_VARS`, and a fake runner
that printed no `SCOPE:` line.

⚠ `test_store.py::test_six_writers_with_a_tiny_busy_timeout_still_land_every_row` failed one
sandbox build and passed the next **on the same tree**; it uses a deliberate 5ms sqlite
`busy_timeout` and its own docstring names an I/O-stalled node as the shape. My diff touches no
dl-router file. Load flake.

## Left open

* **`SCOPE:` has no structural grammar reservation** the way `RESULT:` does. Non-exploitable
  today because `gate.sh` takes the last match and the runner emits from the EXIT trap — pinned by
  `test_a_forged_scope_line_earlier_in_the_stream_cannot_win`. Extending
  `test_result_grammar_is_reserved.py`'s population scan to a second grammar is a separate change.
* **`DEVRC_GATE_ALLOW_AMBIENT` is itself an ambient escape.** It is documented as test-only,
  scrubbed by `_AMBIENT_GATE_VARS`, and needed because the suite must drive the runner seam. An
  ambient value would silently disable the refusal.
* **`gate.sh`'s 91/refusal was built before the re-scope** removed the local full-gate mandate. It
  is cheap, correct and fully tested, so it is kept rather than unpicked — but it is now
  belt-and-braces, not the point.


ZacxDev:
Round-1 audit fixes are pushed. Payload class for this round, decided once and recorded:
**payload** = `scripts/run-tests.sh`, `scripts/gate.sh`, `scripts/run-node-tests.sh`,
`scripts/scoped-tests.sh`, `CLAUDE.md`. **scaffolding** = `scripts/tests/**`, `scripts/testlib/**`.

```audit-claims round=1 audited=e3166a99..c912bc95
1. GUARD 7's NOLAUNCH_ACK REQUIRED direction is suspended for a --files run and announced; its PERMITTED direction still fails an unacknowledged target that reaches a real launcher.
2. GUARD 2's skip TOTAL is suspended for a --files run and announced; its UNPINNED-skip check still fails an unexplained skip, and pin CONDITIONS still gate forgiveness.
3. Suspensions are printed under "whole-target expectations SUSPENDED by this SCOPED run" rather than applied silently.
4. testlib/runner_patch.py gained `expected_skips`, and six new tests build a scoped run against a NON-EMPTY ack ledger and a REAL pinned skip — four of them non-scoped controls proving the guards are suspended, not disabled.
5. --check-targets, --check-floors and --check-suites now report SCOPE: NONE instead of SCOPE: FULL, and NONE takes precedence over narrowing.
6. --files refuses a SHELL_TESTS entry by family instead of handing a .sh to pytest and calling the 0 collected "a collection error or an import breakage".
7. gate.sh refuses all four gate-weakening variables (DEVRC_TARGETS, DEVRC_GATE_PYTEST_RUNNER, DEVRC_GATE_NODE_RUNNER, MIN_TESTS), not only the first; DEVRC_GATE_ALLOW_AMBIENT is the documented test-only escape.
8. scoped-tests.sh refuses an ambient DEVRC_TARGETS rather than mapping against a silently shrunken universe and printing a wrong diagnosis.
9. CLAUDE.md no longer claims a scoped run applies every GUARD; it names the three suspended ledgers, and no longer instructs a local full-gate run before merging.
10. scope_line() in the shared harness takes the LAST SCOPE match, matching gate.sh's tail -1.
11. The 53-test file is split into five files under scripts/testlib/scoped_harness.py; measured 187.6s (load 58) -> 59.5s (load 79.75) under -n 4 --dist loadfile.
12. Mutation round 2 (N1-N9 + control) is in the PR body; N9 SURVIVED and is recorded as a gap this round introduced and then closed with a test.
13. Rebased onto eeea9025 (post-#1429): _devrc_cpu_budget and the nix develop re-exec are kept from main, and #1429's own two-way pins (help-text ledger, ambient-scrub ledger, fake-runner scope) are satisfied rather than relaxed.
```

PR #1449, opened 2026-09-09
docs(handoff): two ranked items were telling the next session to redo MERGED work

ZacxDev:
docs(handoff): two ranked items were telling the next session to redo MERGED work

Closes both, and records what the ladders found — the lessons, not the status.

✅ Old rank 2, `cairn-cutover.py` P3 — `devrc#1428` -> `13c0791a`. 🔴 The item's
own framing was the wrong half of the choice: "either pass the flag or declare
P3 dead" called `--allow-overwrite` the one-line fix, and that flag IS the
data-loss path. `seed.sh`'s tar adds and overwrites but never deletes, so a push
from a frozen mirror silently reverts every pod-newer entry and reports success.
The exit-8 refusal was the last guard, not the bug.

✅ Old rank 3, the opencode blindness — `devrc#1365` -> `14126d94`. Verified on
`origin/main` by CONTENT: 9 occurrences of `OPENCODE_SESSION_ID` where the item
asserted 0. It had been telling the next session to do work that merged hours
earlier.

🔴 The ranked list is NOT self-closing, and this is the second time in this doc
that a closed item sat open — the previous one named its own closing condition
and still needed somebody to come back. The next writer is the only moment anyone
looks.

The durable lessons, recorded in Gotchas where a `State now` replace cannot drop
them:
  * FOUR audit rounds on #1428, and the last three found MY PROSE, not my code —
    each round's fix wrote the next round's finding.
  * The worst defect was ADVICE: a recovery route I recommended writes a second
    mode ledger recording 0444, and `--unfreeze` takes the newest, so a 0600
    entry is "restored" to 0444 and the rollback exits 0. Reproduced end to end.
    Prose that tells an operator what to do is PAYLOAD — audit it like code.
  * A mutation batch that reports green may have applied NO mutant: a
    non-matching anchor printed "97 passed", indistinguishable from a survival.
  * `cairn create` exists, so retiring P3 strands nothing — two comments in the
    tree still claimed otherwise, and that stale sentence was what made the
    retirement look costly.

⚠ #1365 buys CORRECTNESS, not capability: `clawgatectl` has no opencode tier
(measured, `grep -rl OPENCODE containers/` = 0 vs 5), so opencode sessions
resolve exit 5 until a Go change lands in `homelab-talos`. Recorded rather than
left implicit.

Ranked list re-based 1..4. The doc remains over its byte cap — pre-existing, and
it wants a dedicated prune rather than being folded into a closure commit.

---
Docs-only: one file in `claudedocs/`, 0 code files. Gated on the merged tree with the sandbox tier (the one a merge gates on).

### DIFF
--- a/claudedocs/handoff-index-store-claims-accuracy.md
+++ b/claudedocs/handoff-index-store-claims-accuracy.md
@@ -387,38 +387,64 @@ item, not the drift one.
    forcing: gate — it reds the sandbox tier non-deterministically, and the only reason
    #1304 merged through it was a human re-running and reading both results.
 
-2. **Decide `cairn-cutover.py` P3.** It invokes `seed.sh` WITHOUT `--allow-overwrite`
-   (`cairn-cutover.py:1379-1382`) over an ADD + SUPERSEDES + MERGED set, where
-   SUPERSEDES/MERGED are BY DEFINITION entries whose pod bytes differ. 🔴 #1304 made this
-   WORSE: the new NAME-CHECK is a SECOND refusal P3 can hit. Either pass the flag or
-   declare P3 dead post-cutover.
-   forcing: regression — a shipped code path that can never complete.
-
-3. **Fix the opencode blindness in `scripts/lib/clawgate_handoff.sh`.** Diagnosed (squash
-   `13775144`), NOT fixed. It reads only `CLAUDE_CODE_SESSION_ID`;
-   `grep -c OPENCODE_SESSION_ID` is **0**. Detached opencode ⇒ exit 3 forever; NESTED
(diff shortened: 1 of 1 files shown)


ZacxDev:
**Gate — merged tree `5eec4297`, and it is now fully green.**

- sandbox `pytests` — `RESULT: PASS (exit=0)`, 21893 collected / 21891 passed / **0 failed**
- sandbox `nodetests` — `RESULT: PASS (exit=0)`, 1449/1449

The earlier red on this PR was the co-tenant flake (`git maintenance run --auto --detach`), now fixed and merged as `eeea9025` (#1453). The 10 further failures inherited from `605b29ac`/#1439 have also been resolved upstream since my last measurement — `main` is no longer red.

PR #1450, opened 2026-09-09
docs(subsystem-index): the OPEN: block is read from the FROZEN MIRROR, so it can name already-RESOLVED bullets

ZacxDev:
The skill already says the local tree is a read-only mirror, and that *writing* there — or validating a path under it after a write — is wrong. It does not say that the probe's own **reads** come from it. Those reads are what the writer acts on.

`subsystem_touch.py` reads `~/.claude/analyze-service-index` and prints that path on its `store:` line; `cairn` reads `~/.cache/subsystem-store`. So the `OPEN:` block, the `already there` duplicate-check display, and the `journal: N bullets, newest dated …` line are all as stale as that mirror — and nothing in the output says the reading is second-hand. The `store:` line is the only tell, and it reads as provenance rather than a warning.

## Measured 2026-09-09, while following this skill

The probe reported an entry had **2** bullets still declaring `OPEN:`, oldest unverified for 23 days. The live store already carried that exact bullet rewritten as `RESOLVED <ref> (<date>)` — closed five days earlier by another session.

| | mirror | synced cache |
|---|---|---|
| that entry | 162 lines | 170 lines |
| newest file in scope | 2026-09-02 | 2026-09-09 |

Following the block literally would have re-resolved an already-resolved bullet, or re-opened settled work. The remedy is one command: `cairn sync`, then read `~/.cache/subsystem-store/<scope>/<entry>.md` before acting on any `OPEN:` row.

Also noted, because it cuts the other way: the same staleness makes the `already there` comparison miss bullets appended since the last refresh, so the duplicate-check can pass on content that **is** already recorded.

## Instrument validated before trusting the suite

🔴 `python3 <test>.py` on these pytest modules **collects nothing and exits 0** — a green from a command that never ran. Under pytest the protocol module runs 94 tests.

Negative control, aimed at the file the test actually forbids the string in (`write-back.md` — not the owner doc, which legitimately *quotes* the retired prompt):

- injected `append this to the index? (y/N)` → **rc=1, 1 failed / 93 passed**
- restored → **rc=0, 94 passed**
- `test_subsystem_touch.py` → **914 passed**

Docs-only; no behaviour change.


ZacxDev:
**Independent confirmation, 3 days later and larger — and the drift has grown.**

Hit this while following the same step, without having seen this PR (my duplicate-sweep ran too late; closing #1580 in favour of this one).

Measured 2026-09-12, `homelab-talos` scope:

| | mirror | pod |
|---|---|---|
| bullets in the entry | **11** | **29** |
| newest bullet | 2026-08-31 | 2026-09-11 |
| entry files in scope | **11** | **36** |

So 12 days and 18 bullets of drift, against the 7 days / 8 lines measured here on 09-09.

Two additions to what this PR already documents:

1. **`NEAR-MISS` is affected too**, not just `OPEN:` and `already there`. The probe flagged a marker as non-parsing; the live copy already reads `RESOLVED <sha>: (also …)` and parses fine. So all three displays rendered from the mirror are suspect, and the fix-the-LINE instruction on a NEAR-MISS row can send you to rewrite prose that is already correct.
2. **Both flagged items were already fixed upstream in my run** — the NEAR-MISS above, and an `OPEN:` bullet the pod carried as `RESOLVED 841d6fc4:` *plus* a later bullet explicitly closing its open half. I verified that one on its own terms before believing the mirror was wrong (PR merged, and the deployed Pipeline/TriggerTemplate read `clawgate-ci 40m0s`, envelope `tasks: 45m` / `pipeline: 60m`), which is what made the staleness unambiguous rather than a guess.

Worth noting the failure mode this closes is self-inflicted by the protocol: the `already there` comparison exists to stop a writer re-recording a lesson the entry already carries, and a stale input makes it do the opposite.

PR #1452, opened 2026-09-09
docs(tekton): retract "devrc gates on Tekton" — measured false, and it inverts reasoning about a red check

ZacxDev:
devrc `main` has `required_status_checks: ABSENT`, `enforce_admins: false`, zero rulesets (measured 2026-09-09 17:07Z). PR #1430 merged 26 minutes before its pytests failure posted, which a required check cannot allow.

The skill said devrc was the one repo where a Tekton check blocks a merge. Believing that turns a red detector into a perceived block, so a session stops and hunts a defect that may not exist — on devrc a red has repeatedly been a resource stall (`test_mjs_parses[attachments.mjs]`, 5+ false failures, real assertion `TimeoutExpired` on a command costing 0.01-0.02s; same revision failed one run and passed the next).

Fixes both the gotcha and the dependent claim in the trigger list.

🔴 Note: `~/.claude/skills/tekton/SKILL.md` resolves into `/nix/store` (a `home.file` copy), so this needs a `home-manager switch` to go live locally.

### DIFF
--- a/claude/skills/tekton/SKILL.md
+++ b/claude/skills/tekton/SKILL.md
@@ -242,15 +242,33 @@ debugging, changing or copying a specific pipeline.
    🔴 Bound EVERY task, not just the slow one — the task deadline is
    `taskStart + timeout` while the budget is `runStart + tasks`, so an unbounded early task
    (devrc's `notify` inherited the cluster's 1h default) lets them cross and re-opens this.
-9. ⚠ **Gotcha 6 is scoped to `homelab-infra`, not to Tekton.** `innovation-upstream/devrc` is a
-   DIFFERENT repo on a plan where protection works, and since 2026-08-23 it requires **both**
-   `tekton/devrc-nodetests` and `tekton/devrc-pytests` (measured — re-measure, this moved
-   twice in one day). A required check `ERROR`/`PENDING` ⇒ `mergeStateStatus=BLOCKED`,
-   both `SUCCESS` ⇒ `CLEAN`. So on devrc a Tekton check **is** a gate — on either tier;
-   the earlier nodetests-only window let pytests-red PRs read `UNSTABLE` and merge.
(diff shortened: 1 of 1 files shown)


ZacxDev:
Not my PR and I have not touched the branch — flagging two sites the sweep looks to have missed, since this is the PR that owns the retraction.

Re-measured just now, agreeing with your figures: `required_status_checks: ABSENT`, `enforce_admins: false`, `rulesets: 0` (2026-09-11T21:59:58Z).

**`:330` is the one I'd fix.** It is not narrating a past incident — it is forward-looking advice, and the false premise is what makes the advice strong:

> 🔴 Retry only with the ownership question answered FIRST and **proven on a scratch pipeline** … and with `enforce_admins: true` on devrc `main` a wrong guess blocks **everyone's** merges.

A wrong guess does not block everyone's merges. The caution may still be right for other reasons — the `7839ef54` "the place to find the third failure mode is not production" argument stands on its own — but it is currently resting on the claim this PR retracts, which is the shape where a reader reasons *from* the stale half.

**`:296` is defensible as history** (`…both required checks post COULD NOT RUN, so with enforce_admins: true nothing could merge. Fixed by …`) — it describes a state that held at the time. Worth a date or a past-tense marker so "both **required** checks" is not read as current, but it is not wrong.

Context for why this came up: I hit the same staleness from the other side. `handoff-gate-flake-store-api.md` rank 1 briefly cited this skill as corroboration for "advisory, not required" — and the skill said the opposite, so the pointer sent a reader to a 🔴 contradicting the line that sent them. That cross-reference is now removed and replaced with the live measurement plus a note that this file is stale pending your PR (`#1477`).

One reason to sweep the remaining sites rather than land the headline retraction alone: **a skill loads deterministically on trigger**, so it is the copy most likely to be read, and a reader who acts on `:330` will over-cautious their way out of a retry that is actually cheap now.

PR #1454, opened 2026-09-09
feat(mention-open): fzf in a float terminal replaces rofi — the eponymous repo ranked 8th of 230, now 1st

ZacxDev:
## What changed

`scripts/mention-open.py`'s `pick()` no longer runs rofi. It spawns a dedicated
float **Alacritty** window running **`fzf -i --tiebreak=end`**, feeds it the rows
over a **FIFO pair** in a 0700 `mkdtemp`, and reads the selection back the same
way. What the picker offers, in what order, what the note says, and the rule that
a dismissal opens NOTHING are all unchanged.

## Why — the ranking, with numbers

rofi's `-sorting-method fzf` produced the right *scores* and then broke every tie
by **row length**. In `owner/repo` every row of one owner matches at the same
offset, so all of them score identically and the **longest loses**.

Measured on the operator's real 392-row universe (locally, never committed),
query `civitai`, rows in the exact shape `picker_rows()` builds:

| query | rows matching | rank, rofi / fzf default | rank, `--tiebreak=end` |
|---|---|---|---|
| `civitai` (→ `civitai/civitai`) | 230 | **8** | **1** |
| `devrc` | 16 | 1 | 1 |
| `talos-infra` | 1 | 1 | 1 |
| `comfyui` | 5 | 1 | 1 |

The three controls do not move. "rofi" and "fzf default" are one column because
they agree: fzf's own default tiebreak **is** `length`, so rofi's fzf sorting was
faithfully reproducing upstream fzf's default — 8th, both. The fix is a flag rofi
does not expose, not a better algorithm.

Reproduced on a **synthetic** corpus of the same shape (392 rows,
`nimbusworks/nimbusworks` with shorter siblings), which is what the committed
test uses: rank **27 of 41** by default, **1 of 41** with `--tiebreak=end`.

`--tiebreak=end` is principled rather than lucky: the row ends
`…/owner/repo/pull/<id>`, the suffix after the repo is a constant, so "prefer the
match nearest the end" *is* "prefer the repo name over the owner".

### Flags considered and rejected, by measurement

* **`--exact`** — does **not** fix it. Same corpus, same query: rank **27 with it
  and 27 without**. The defect is a tiebreak, and exact matching leaves the
  scores just as tied. It also narrows the match set (a 6-char prefix: 56 rows →
  21), which is the fuzzy narrowing that makes a 392-row universe usable at all.
* **`--select-1`** — would re-open the defect #1336 closed. It auto-accepts when
  the initial query matches exactly one row, and the one-row picker is precisely
  the guessed-repo confirmation case that must never open unconfirmed.

Both absences are asserted, with the measurement in the failure message.

## The invariants, and how each was verified

1. **Clawgate row stays FIRST.** fzf ranks a *scored* match set; an empty query
   scores nothing, so input order survives. Verified three ways: a `--filter ""`
   test, a **pty-driven interactive fzf**, and an **end-to-end run under
   `xvfb-run`** with a real Alacritty + real fzf — a bare Enter returned
   `https://clawgate.zacx.dev/tasks/370`, pointer on the clawgate row.
2. **Disclosure.** Rows never touch argv, an env var or a file. A terminal does
   not proxy stdin, so `input=` had nowhere to go; a FIFO holds nothing at rest.
   The **note** goes down the same pipe as `--header-lines`, not `--header` on
   argv, so "nothing that could carry a universe token is ever an argument" is
   structural. Guards: `test_NO_ROW_reaches_the_terminal_ARGV`,
   `test_the_rows_go_down_a_FIFO_that_does_not_outlive_the_pick`,
   `test_the_NOTE_survives_all_the_way_from_main_to_the_picker`,
   `test_no_toast_body_can_name_a_REPOSITORY`. The existing `_every_sink` /
   `_no_universe_token_anywhere` guards are untouched and green.
3. **The no-network ledger.** `RESOLUTION_PATH_COMMANDS` unchanged and still
   two-way. That test stubs `pick`, so the picker was never in its ledger and
   putting it there would change what the ledger means; the picker is covered by
   the AST ledgers plus tests driving the real `pick`.
4. **The AST spawn pin.** `EXPECTED_ARGV0` / `SPAWNABLE_EXECUTABLES` move
   `rofi → alacritty`, still grows-or-shrinks with the `<computed>` sentinel; the
   list-literal property is re-pinned on the `Popen`.
5. **The launcher ledger.** `alacritty` is now **stubbed** (`HAZARD_VOCABULARY` +
   `HOST_LAUNCHERS`), not acknowledged — so no `ACKNOWLEDGED_UNSTUBBED` entry and
   none of the blinding that table's `tmux-reply-agent` entry records. It also
   closes a pre-existing hole: ten top-level scripts already reached `alacritty`
   unstubbed. `fzf` is deliberately not added — it takes nothing from the
   operator's session and cannot run without the terminal that is stubbed.
6. **The note.** `-mesg` → `--header-lines`, wrapped by `picker_header()`. Still
   names the clicked text and a count, never a repository. Pango escaping
   removed — fzf's header is plain text.
7. **`-no-custom` equivalent.** fzf cannot return the query; that needs
   `--print-query`, whose absence is pinned. Measured live: typing
   `zzznotarepoxyz` and pressing **Enter** returned `""`, as did Esc.
8. **`row_to_url` matches on URL suffix** — unchanged; round trip pinned.
9. **The wrapper's PATH.** `pkgs.alacritty` and `pkgs.fzf` added. `fzf` sits one
   level down (inside the `-c` script), so the wrapper-PATH test grew a second
   AST reader, `_shell_child_commands`, taking the first word of that script.
10. **No window raised.** Unit tests replace `Popen` with a fake terminal thread;
    `alacritty` is stubbed besides. The only real windows were on `xvfb-run -a`.

**Nothing in the brief had to be given up.**

## Round-1 audit — six findings, all fixed

Verdict was "merge after considering the two 🟡s"; the payload's core claims held
under independent measurement. Every fix below arrives with a mutation row that
was watched to kill.

* 🟡 **`-i` was dropped, so the picker became SMART-CASE.** rofi's `-i` was
  unconditional; fzf's default makes any query containing an uppercase letter
  case-SENSITIVE. Reproduced independently: query `NimbusWorks` → **0 rows**
  without `-i`, **41** with. `pick()` cannot tell an empty list from a dismissal,
  so a capital letter produced SILENCE — the exact wall this PR removes — and
  `repo_universe()` preserves original casing, so mixed-case rows are real.
  Ranking unaffected (1/41 either way). Now pinned by a flag test *and* a
  real-fzf behavioural test with a negative control. **(K54)**
* 🟡 **`BrokenPipeError` escaped the write loop and discarded the selection.**
  fzf answers on the first keypress and closes its read end; the next `os.write`
  raises EPIPE, an `OSError`, which fired a FALSE error toast and returned `""`
  while the chosen row was already in the choice FIFO. Measured: fine at 392 rows
  (~21 KB), broken at 1,200 (~67 KB) and 3,000 (~172 KB) — the 64 KiB pipe-buffer
  boundary, the same threshold the open-order fix exists for. 🔴 The fake had to
  change to see it: `_FakeTerminal` drained the whole FIFO before answering, i.e.
  it was a **friendlier peer than fzf**, so every transport test was structurally
  blind. `answer_early=True` makes the case reachable. **(K55)**
* 🟡 **The `-c` script is a surface rofi did not have, and its only reader saw
  one word.** `fzf … | tee /tmp/picker.log >"$2"` would write private rows to
  disk with every ledger green. Now pinned three ways: the whole normalised
  string, a metacharacter ban with a reason per character, and the redirection
  set read through `shlex` — not a character count, which the first version got
  wrong because `>` also appears inside `--prompt` and `--pointer`. The positive
  control records **which** guard catches the leaky mutant, since the redirection
  check alone does not. **(K56)**
* 🟢 **The picker terminal inherited `selection.save_to_clipboard = true`** — a
  drag would copy private repo names into the X CLIPBOARD, which outlives the
  pick and is readable by every X client, and the module's disclosure enumeration
  does not list a clipboard. Now `-o selection.save_to_clipboard=false`. **(K57)**
* 🟢 **A comment is a claim.** `test_no_real_launchers.py` still said the spawn
  pin asserts `{…, rofi}` and three PATH-clobber justifications did not enumerate
  `alacritty`. Both corrected, the first restated rather than softened.
* 🟢 **An undeclared behaviour change, restored.** rofi's `timeout=120` produced a
  toast; the rewrite went silent while the body claimed "same behaviour as
  before". `run_picker` now reports an outcome
  (`selected`/`dismissed`/`timeout`/`never-shown`) and `pick()` toasts for the
  last two. **A dismissal stays silent** — a toast there fires after every
  dismissal. `never-shown` is a case rofi had no analogue for: the terminal
  started and died before it could be fed, which is what a missing `fzf` on the
  wrapper's PATH looks like. **(K58)**

## Two defects found while building it

**A latent deadlock.** `/bin/sh` applies `<"$1" >"$2"` before exec'ing fzf, and
opening a FIFO for writing blocks until a reader exists — so opening the choice
FIFO last parks the child with nothing draining `$1`. Invisible at today's ~27 KB.
`run_picker` opens the choice FIFO first (`O_RDWR`). Regression test matrix:

| tree | the big-payload test | the three small-payload transport tests |
|---|---|---|
| pre-fix open order | **FAILED** after 120 s, its own message | passed |
| HEAD | passed | passed |

**Four tests that self-skipped in the tier that gates the merge.** They were
`skipif(fzf missing)`; the dev-host tier ran all four, the nix sandbox skipped all
four, and `run-tests.sh` failed the derivation on the unpinned skips — every
target PASSED and the build still failed. The authoritative tier had **zero**
coverage of the property this change exists for. Fixed the repo's own way
(`gateTools`' `pkgs.tmux` entry documents the identical case): `pkgs.fzf` →
`gateTools`, `fzf` → `REQUIRED_TOOLS`, and the four marks became `_require_fzf()`,
which asserts and names both files.

## Mutation results

11 rows added across two commits (K48–K58). **10/10 killed for the stated
reason**, plus the P1 control, across two runs.

K48–K51 were **re-anchored** after the `-i` and outcome-tuple edits moved their
lines — `test_mutation_battery_anchors.py` caught that before the battery could
score them as silent survivors, which is exactly the failure that guard exists
for. K58 first scored **KILLED-WRONG-REASON**: the right test caught it, but
`assert said and "timed out" in said[0][0]` short-circuits on an empty list, so
the phrase lived only in the SOURCE line and never in pytest's `E ` lines. That is
a scoring blind spot rather than a weak guard; both toast tests now put the token
in the assertion **message**, and K58 re-ran KILLED(attributed).

## Gate

⚠ **`main`'s merge policy changed while this PR was open** (`CLAUDE.md`, merged
into this branch): the local full-suite ritual before every merge is **deleted**,
CI is advisory, and a **change-scoped subset** plus reading CI is what replaces
it. Running the full tiers is now a judgement, not a checklist item.

**Base: `origin/main` `cdfd14ab` is an ancestor of head `1e5bc9a1`, 0 commits
behind, verified twice at gate time.** That merge is a genuine test-merge, not a
formality: main's #1445 rewrote `scripts/run-tests.sh`, the same file this branch
edits (`REQUIRED_TOOLS`), and the `fzf` entry survived it — checked by reading the
merged line, not by trusting a clean `git merge`.

* **Advisory CI on head `51ae8463`: BOTH GREEN.**
  `tekton/devrc-pytests` → `TOTAL collected=21934 passed=21932 skipped=2 failed=0`;
  `tekton/devrc-nodetests` → `TOTAL suites=5 files=41 tests=1449 pass=1449 fail=0`.
  CI runs the `nix build` sandbox tier, so this is the authoritative tier
  reporting the same numbers below — measured by something other than me.
* ⚠ **Advisory CI on head `1e5bc9a1` (the newer merge): `devrc-pytests` RED on
  ONE test, `devrc-nodetests` and `devrc-cairn-client-runs` green.** The failing
  test is
  `test_run_tests_targets.py::test_a_partial_run_is_declared_where_gate_sh_actually_LOOKS`.
  I did **not** dismiss it on "my diff cannot reach it" — this branch edits
  `scripts/run-tests.sh`, and that is a meta-test *of* `run-tests.sh`/`gate.sh`.
  It gets a control instead:

  | | branch `1e5bc9a1` | pristine `origin/main` `c6700447` |
  |---|---|---|
  | isolated, 3 runs | **3/3 passed** (45–49 s) | **3/3 passed** (64–80 s) |

  Positive control: `1 selected` for the `-k` filter on every run, so those are
  not six vacuous zeros. It passes on both trees, and the branch is **faster**
  than pristine main — my `REQUIRED_TOOLS` entry adds one `command -v`. The
  mechanism is the one already measured on this family: the test spawns a
  **nested `run-tests.sh`** under a **120 s cap** and takes 45–80 s standalone,
  so it has no headroom under CI contention. My own sandbox build of **this exact
  commit** passed it (`21991 passed, 0 failed`), which is the same tier CI runs.
  Recorded rather than re-run to green — and `CLAUDE.md` puts this repo's red-noise
  rate at roughly two in five.
* **Sandbox tier locally on the merged tree `1e5bc9a1`, both derivations built
  ONE AT A TIME from a `.git`-free `git archive` extract, plain `nix build`:**

  ```
  devrc-pytests>   RESULT: PASS (exit=0)   TOTAL collected=21993 passed=21991 skipped=2 failed=0
  ```

  The 2 skips are the pre-existing pinned Postgres pair.
  🔴 **The fzf-skip claim, read out of the build log rather than inferred from
  PASS:** `PASS scripts/tests (collected=13833 passed=13833 skipped=0 …)` and
  **zero** `fzf is not on PATH` lines — against **13457 passed / 4 skipped**
  before the fix. Run at all because this branch edits the **runners themselves**
  (`scripts/run-tests.sh`, `flake.nix` gateTools), which the new policy names as
  a case where the full tier is still right, and because "four tests stopped
  skipping" is only observable there.
* **Change-scoped subset, merged tree `51ae8463`: 470 passed, 0 failed** —
  `test_mention_open.py`, `test_no_real_launchers.py`,
  `test_mutation_battery_anchors.py`, `test_alacritty_hints.py`,
  `test_mention_scan.py`.

### Round ledger

`round 2 · payload lines changed THIS round: 110 (since round 1: 110) · working
elapsed: 4 h 14 m`

Counted with the skill's own command —
`git log --numstat --format= --remerge-diff 97e01d15..HEAD --not cdfd14ab` — rc 0,
silent stderr, non-empty range. Round 2 touched four files: `scripts/mention-open.py`
(110 lines, **PAYLOAD**), `scripts/tests/test_mention_open.py` (325),
`scripts/tests/mutation_battery_mentions.py` (51),
`scripts/tests/test_no_real_launchers.py` (20) — all three **scaffolding** by the
revert test: revert any of them and the picker still ships. The two counts are
equal because round 2 is the first fix round. Wall-clock from the audited sha is
~22 h, but that includes an API-limit stop with no work in it; **4 h 14 m** is the
figure that measures the ladder.

## Things the brief did not anticipate

* **`fzf --filter` and interactive fzf had to be checked separately.** An Xvfb run
  once picked the wrong row, which read as "interactive ranks differently". It
  does not — that run read the working tree **while the mutation battery was
  rewriting it in place**, so it ran mutant K49. `--filter` is a sound proxy here
  *only because it was then checked*, which is why the pty pair exists.
* **`pty.fork()` leaves the terminal 0×0 and fzf draws nothing** — while still
  filtering and still returning the right row. A pty test that only checked the
  return value would pass while measuring a picker that rendered no prompt.
* **A terminal emulator does not proxy stdin** — the whole reason the transport
  had to change.
* **`alacritty --class A,B` sets class=A, instance=B**, so
  `--class float,mention-open` is caught by the existing
  `for_window [class="float"]` rule. Verified with `xdotool search --classname`
  under Xvfb; no new i3 rule needed.
* **`ROFI_THEME` is gone** — fzf takes its palette from the terminal's ANSI
  colours (`--color=16`), which this same nix file paints gruvbox.

## Residual risk

* The picker window is 120×22 and `PICKER_TIMEOUT` is 120 s (inherited from the
  rofi call). Both are taste, not measurement.
* The three PATH-clobber justifications in `test_no_real_launchers.py` enumerate
  HAZARD_VOCABULARY names in prose and were already **not** exhaustive before
  this PR (they omit `zenity`, `loginctl`, `ddcutil`, `brightnessctl`, `xset`,
  `pactl`, `playerctl`, `dunstify`, `dunstctl`). I added `alacritty` to keep them
  consistent, but the real guard there is each fixture asserting its own
  directory contents — the prose list is decoration and a future reader should
  not trust it as a ledger.


ZacxDev:
Round 1 audit findings and the fixes that claim to address them, recorded so the next round can diff against them.

legend: `<from>` = the tip round 1's audit READ · `<to>` = the head round 1's FIXES produced.

```audit-claims round=1 audited=97e01d15..1e5bc9a1
1. 🟡 `-i` restored to PICKER_SH — a capitalised query no longer returns an empty list under fzf's default smart-case; claimed pinned by a test, where the old rofi test never asserted `-i`.
2. 🟡 BrokenPipeError caught in run_picker's write loop — a selection made before the payload is fully fed is no longer discarded, and no false error toast fires; claimed the regression test can now see it, where _FakeTerminal previously drained the FIFO first.
3. 🟡 PICKER_SH's SHAPE pinned, not just its first word — _shell_child_commands previously read text.split()[0], so an added `;`, `|` or `$(` kept every guard green.
4. 🟢 selection.save_to_clipboard=false passed to the picker terminal — rows can no longer reach the X CLIPBOARD, which outlived the pick.
5. 🟢 the stale `{git, tmux, notify-send, xdg-open, rofi}` claim in test_no_real_launchers.py corrected to alacritty, plus the four PATH-clobber enumerations.
6. 🟡 the timeout toast RESTORED via a run_picker outcome enum (selected/dismissed/timeout/never-shown); pick() toasts for timeout and never-shown, stays silent on dismissal. The PR body no longer claims "same behaviour as before".
7. residual-risk section: the `[INT]` browser-bridge test withdrawn as an inherited red — it was measured on a loaded box and both trees were contended.
8. K48-K51 mutation rows re-anchored after the edits moved their lines; battery re-run 10/10 killed plus the P1 control.
```

Round 1 was a first, full audit, so it has no ledger. Round 2's fixes reported: `round 2 · payload lines changed THIS round: 110 (since round 1: 110)`.

⚠ Verified independently by the coordinating session, not taken on the author's word: the `-i` regression (`Civitai` → 0 rows without `-i`, 2 with) and the git-maintenance root cause underlying the sibling flake PR.

⚠ One CI red on head `1e5bc9a1` — `test_a_partial_run_is_declared_where_gate_sh_actually_LOOKS` — is contention, not a defect: it passes 3/3 on this branch and, measured separately by the coordinator, 2/2 on pristine `origin/main` at **84–89s against its own 120s cap**. It has no headroom on a loaded box regardless of this PR.

ZacxDev:
## Round 2 — fixes landed

```audit-claims round=2 audited=1e5bc9a1f8efa0d3a23bd72472e0ee249452d0cd..3ae6051e
1. A missing `fzf` is now caught by a `shutil.which` pre-flight in `pick()` BEFORE the terminal is spawned, and toasts naming the nix attribute to add — it no longer reaches `run_picker` as a silent dismissal.
2. `PICKED_NEVER_SHOWN`'s comment, toast body, test docstring and commit prose no longer claim a missing `fzf` can produce it; the toast now names the cause that can (the terminal exiting before the shell ran) and no longer sends the operator to check a PATH that is fine.
3. The read-loop timeout arm — the picker OPEN with nobody answering, which is what rofi's `timeout=120` was — is now covered by `test_a_timeout_with_the_LIST_ON_SCREEN_toasts` using a peer that opens both FIFOs and drains the rows; the audit's own probe mutant on that line now dies (K59, KILLED-attributed).
4. `test_the_two_TIMEOUT_arms_are_different_code_paths` asserts `run_picker` keeps BOTH deadline checks, so the two arms cannot collapse into one test again.
5. The `-i` measurement is corrected in all four places to the committed corpus: 195 rows, `NimbusWorks` 0 matched without / 15 with, and the ranking claim is re-stated on the lowercase query instead of the self-contradictory "1/41 with and without".
6. The redirection guard is widened to `^\d*[<>]`, so fd-prefixed redirections (`2>/tmp/leak`, `1>/tmp/leak`) are seen; a positive control asserts both shapes are caught, and the docstring's "both, and no more" is now true of the body.
7. `PICKED_OUTCOMES` pins the outcome set two-way and `pick()` has a catch-all, so a fifth outcome cannot be silently dropped.
8. The "removed in a `finally`" claim is qualified in the source: measured 0 survivors from a clean suite run and exactly 1 from a process SIGKILLed inside `run_picker`, so the 18 orphaned dirs are killed test subprocesses, not clicks — and are empty FIFOs in a 0700 dir either way.
```

**Ledger:** `round 2 · payload lines changed THIS round: 98 (round 1's fixes: 110) · working elapsed: ~2 h`

`git log --numstat --format= --remerge-diff 1e5bc9a1..HEAD --not 719519fa` — rc 0,
silent stderr, non-empty range. Three files: `scripts/mention-open.py` **98 =
payload**; `test_mention_open.py` 216 and `mutation_battery_mentions.py` 37 =
scaffolding by the revert test. Both fix rounds are non-zero, so the two-consecutive-zero
stop has **not** fired.

**Mutation:** K59–K61 added; **8/8 killed for the stated reason** across two runs with the
P1 control. K60 first scored KILLED-WRONG-REASON for the same reason K58 did last round —
the mutant trips an *earlier* assertion than the one whose message carried the expected
token. That is a scoring blind spot in how I write these rows, not a weak guard, and it has
now cost two rows in two rounds.

**Not taken:** the auditor's suggestion to add a discriminator *inside* `PICKER_SH` — the
pre-flight achieves the same detection without touching a surface that is now pinned three
ways, and it never raises a window for a picker that cannot run.

ZacxDev:
## Round 3 — fixes landed

```audit-claims round=3 audited=3ae6051e..47c62402
1. The missing-`fzf` pre-flight toast no longer names `nix/programs/alacritty/default.nix` as the remedy; it names the DEPLOY gap and `home-manager switch --flake ~/workspace/devrc --impure`, because that file's `pkgs.fzf` entry is test-enforced and therefore already correct in any passing tree.
2. `mention-open.py` is added to the `home-manager` ACKNOWLEDGED_UNSTUBBED file set as a PROSE-MENTION (the new toast string), re-justified rather than reworded, and carrying its pin — the AST spawn ledger asserting argv[0] is exactly {git, tmux, notify-send, xdg-open, alacritty}.
3. The redirection guard and its control now call ONE function, `_redirections()`; narrowing that function back to the round-2 defect turns three of the four control cases RED, where the previous inline-copy control stayed green at 232 passed.
4. The redirection reader is widened to `^(\d*|&)[<>]` so `&>` is seen — measured honoured by this /bin/sh (bash 5.3p15) — and `&>` is added to the control set.
5. The comment claiming `^\d*[<>]` is "every redirection shape sh accepts" is replaced with what was measured: the shapes covered, plus `{fd}>` named as excluded because this shell does not honour it.
6. The "alacritty exits 0 … or does not exist at all" claim is corrected in both places — `-e zzznosuchbinary` exits 1 — with the unchanged conclusion and its reason (`-e` here is always the existing /bin/sh) stated.
7. The unknown-outcome catch-all no longer interpolates `str(outcome)` into a notification body; the body is fixed, and a test drives the branch via a stubbed `run_picker` to prove no supplied string reaches either field.
8. The mutation battery header now states all three ways an `expected` token gets mis-scored (short-circuited assertion, earlier assertion firing, case mismatch), after K63 became the third such row in three rounds.
```

**Ledger:** `round 3 · payload lines changed THIS round: 43 (round 2: 98 · round 1: 110) · payload 16.9% of added lines`

`git log --numstat --format= --remerge-diff 3ae6051e..HEAD --not ce9b55c3` — rc 0,
silent stderr, non-empty range. Four files: `scripts/mention-open.py` **43 = payload**;
`test_mention_open.py` 142, `mutation_battery_mentions.py` 31,
`test_no_real_launchers.py` 20 = scaffolding by the revert test. Non-zero, so the
two-consecutive-zero stop has not fired — but the trend across three fix rounds is
**110 → 98 → 43**, and every round-3 finding was prose-about-the-fix rather than
behaviour.

**Mutation:** K62–K63 added, both KILLED(attributed) with the P1 control. K63 first scored
KILLED-WRONG-REASON — the third row in three rounds — because my message spelled the token
in a different case. All three causes are now written into the battery header.

**Two things I did NOT do, and why:**
- The auditor's F5-adjacent note that `mention-open.py:835` still says "its `else` is
  SILENCE" in the present tense: left as-is, since they marked it not-a-finding and the
  operative instruction beside it is correct and test-enforced. Say the word and I will
  reword it.
- No guard was added to catch the `expected`-token mis-scoring, because a text check would
  have caught only one of the three cases — the other two are runtime ordering, which only
  running the row detects.

**Gate — merged tree `47c62402`, base `origin/main` `ce9b55c3` (ancestor, 0 behind, checked
at gate time):**

```
scoped-tests.sh            SCOPE: SCOPED (70 files across 5 of 28 targets)   RESULT: PASS (exit=0)
devrc-pytests   (sandbox)  SCOPE: FULL (28 of 28 hermetic targets)           RESULT: PASS (exit=0)
devrc-nodetests (sandbox)  SCOPE: FULL                                       RESULT: PASS (exit=0)
```

Both sandbox derivations built ONE AT A TIME from a `.git`-free `git archive` extract, plain
`nix build`. That closes the first of the two things the round-3 auditor could not verify.

**The second, re-measured:** a process parked INSIDE `run_picker` and SIGKILLed leaves
exactly **1** dir (`dirs while parked: 1` → `dirs after SIGKILL: 1`), against **0** from a
clean 236-test run. So the `finally` qualification holds in both directions.

⚠ **Advisory CI is RED on `test_subsystem_touch.py::TestCommitMutationKillMatrix::
test_kills_the_LENGTH_guard`**, which this branch **does not touch** (0 changes to that file
vs `origin/main`). Isolated both directions with a selection control (`1 selected` each run):

| | branch `47c62402` | pristine `origin/main` `ce9b55c3` |
|---|---|---|
| isolated, 3 runs | **3/3 passed** (~0.5 s) | **3/3 passed** (~0.5 s) |

⚠ I am **not** claiming the mechanism here. The earlier reds on this PR were a nested-runner
120 s cap; this test runs in half a second, so that diagnosis does not transfer and I have
not established what does. What I can say: it passes on both trees, my diff cannot reach it,
and `CLAUDE.md` puts this repo's red rate at roughly two in five. `devrc-nodetests` and
`devrc-cairn-client-runs` are green on the same head.

ZacxDev:
## The CI red is explained, and it is NOT this branch — it is a latent flake on `main`

`tekton/devrc-pytests` is red on head `47c62402` for exactly one test:
`test_subsystem_touch.py::TestCommitMutationKillMatrix::test_kills_the_LENGTH_guard`
(`collected=22013 passed=22010 skipped=2` → 1 failure). This branch changes that
file **0 times**.

### Mechanism

The test truncates a freshly-created commit sha to **three characters** and
asserts the mutated module resolves it back to that commit:

```python
wrong = mod.collect_commit_paths(repo, [sha[:3]])
assert wrong.commits == (sha,), "wrong thing removed"
```

Three chars is deliberate — `COMMIT_SHA_MIN_CHARS = 4`, so the prefix must be
*below* the bound to exercise the LENGTH guard. But a 3-hex prefix has only 4096
values and the fixture repo holds **9 objects**, so the prefix is sometimes
**ambiguous**. When it is, the mutant raises instead of returning, and the
assertion is never reached:

```
mc_len.CommitAmbiguousError: commit sha is ambiguous: `cc7` names 2 objects in this repo
(cc77442a1600d94ab2df7a243ac5965c8f861add, cc7c779ea8a0b424eb9437982a7847c1ac14d077)
```

### Measured, on pristine `origin/main` (`ce9b55c3`), with none of this branch present

| experiment | N | hits | rate |
|---|---|---|---|
| 3-char prefix ambiguous in a fixture-shaped repo | 600 | 2 | **0.333%** |
| the REAL test failing, run in isolation on pristine `main` | 600 | 1 | **0.167%** |

The failure above is from iteration 113 of the pristine-`main` run. The test has
been in the tree since #432 (2026-08-12).

### What this corrects — including my own earlier report

* **It does NOT need the full run.** The hypothesis was cross-test interference
  from this branch's terminal/FIFO tests. That is refuted: it fails in
  **isolation**, on **pristine `main`**, with no other test in the session.
* 🔴 **My earlier "3/3 passed both directions" was not evidence.** Three runs
  against a ~0.2% event cannot distinguish "passes" from "fails rarely", and I
  reported it as though it settled the question. It did not. Sizing N to a
  measured rate is what made the difference here.
* `#1457` and `#1441` being green is consistent with this, not evidence against
  it: at ~0.2% per run, a green is the expected outcome ~99.8% of the time.

### Not fixed here, deliberately

The fix is a design decision inside another subsystem's test — the prefix must
stay below `COMMIT_SHA_MIN_CHARS`, so "use more characters" is not available; it
needs to either regenerate until the prefix is unambiguous or assert that
precondition explicitly. Doing that in this PR would add unrelated payload to a
branch three audit rounds deep. **Closing condition if someone picks it up:**
`test_kills_the_LENGTH_guard` no longer depends on a probabilistic property of a
generated sha — checkable by re-running the 600-iteration loop above and getting
0 failures.

**CI was not re-run.** A re-run would very likely go green and would be
information about the flake rate, not a fix.

ZacxDev:
## Merge decision — gate on the merged tree, and a red I am merging through

**Base: `origin/main` `bfb36b14` is an ancestor of head `c5e3123e`, 0 behind, checked at gate time.**

```
scoped-tests.sh            SCOPE: SCOPED (70 files across 5 of 28 targets)   RESULT: PASS (exit=0)
devrc-pytests   (sandbox)  SCOPE: FULL (28 of 28 hermetic targets)           RESULT: PASS (exit=0)
devrc-nodetests (sandbox)  SCOPE: FULL                                       RESULT: PASS (exit=0)
```

Both sandbox derivations built ONE AT A TIME from a `.git`-free `git archive`
extract, plain `nix build`, per #1445.

### The conflict resolution was not just textual

`origin/main` brought #1457, whose point is that `_no_universe_token_anywhere`
knows `FAKE_UNIVERSE` and the pane guess is deliberately **not** in it — so that
guard covers every repository except the one at issue. **Every sink guard this
branch added used that universe guard alone** (terminal argv, end-to-end header,
toast bodies). A textually clean merge would therefore have re-created, in the
surfaces this branch introduces, exactly the blindness #1457 was written to close
— with the whole suite green.

`test_the_GUESSED_note_reaches_the_picker_and_no_other_surface` is the join: it
drives the guessed path through the real transport and asserts the SPLIT — the
guess **must** appear in the candidate rows (offering it is the feature) and must
appear in the note, the terminal argv and every published sink **nowhere**, each
with **both** token guards, behind three positive controls.
`test_the_guessed_seam_guard_can_actually_FIRE` pins the blindness itself, so the
pair cannot be "simplified" back to one call.

Control watched: mutating `main()` to put the guessed repo into the note turns
that seam test **red** (among 22), so it is not vacuous.

### 🔴 Merging through a red, and saying so

`tekton/devrc-pytests` is red on
`test_run_tests_targets.py::test_the_SUMMARY_BANNER_names_the_real_selection_source`.
`devrc-nodetests` and `devrc-cairn-client-runs` are green on the same head.

This is the nested-runner family already characterised on this PR — those tests
spawn a **nested `run-tests.sh` under a 120 s cap** and take 45–80 s standalone,
so they have no headroom under CI contention. Evidence, both directions, measured
earlier in this ladder:

| | branch | pristine `origin/main` |
|---|---|---|
| `test_run_tests_targets.py`, run 1 | 1 failed (`SUMMARY_BANNER`) | 1 failed (`subset_note`) |
| run 2 | **31 passed** | **31 passed** |

A *different* test fails each time and both trees go green on a re-run — the
signature of contention, not of a defect. I did **not** dismiss it on
reachability (this branch does edit `scripts/run-tests.sh`), and the decisive
check is same-tier: **my local `nix build` sandbox run of this exact tree passed
`scripts/tests` FULL**, which is the tier CI runs.

**CI was not re-run.** A green re-run would be information about the flake rate,
not a fix, so the red stands on the record with this explanation attached.

PR #1469, opened 2026-09-10
feat(main-green): start the deadman EARLY on main's own CI status, instead of polling harder

ZacxDev:
# 🔴 READ THIS FIRST: `/commits/{sha}/status` LIES ABOUT THIS REPO'S CI

**If you are reaching for a GitHub status API in this repo, use
`/commits/{sha}/statuses` (plural). Not `/commits/{sha}/status` (singular).**

The singular endpoint is the **roll-up**, and it collapses `error` into
`failure`. This repo's main-push pipeline reports every *non-code* outcome as
`error` — `superseded by a newer run`, `KILLED: the gate pod died`,
`NO GATE POD` — each of which says *"Not a code failure."* in its own
description. Measured over the 100 newest commits to `main`
(2026-09-08 → 09-10):

| read via | `main` looks red on |
|---|---|
| `/commits/{sha}/status` (roll-up `.state`) | **48 of 60 pushes (~80%)** |
| `/commits/{sha}/statuses` (per-status rows) | **9 of 200 rows** |

A watcher built on the roll-up would have fired on ~80% of main pushes **on day
one** — which is precisely the *"a fast alert that fires on a flake is worse than
a slow one"* failure this work exists to prevent, and the permanently-red gate
`claude/RULES.md` calls worse than no gate.

**This PR uses the plural endpoint**, which returns only the status array and
carries **no roll-up field at all** — so the mistake is *unavailable*, not merely
avoided. `test_the_script_never_reads_the_rollup_status_endpoint` pins it
structurally (`/status` not followed by `e` or `/` must not appear in code).

This is the same family as the `/check-runs`-returns-nothing trap: Tekton posts
**commit statuses**, not check-runs, so a check-runs query returns an empty set
that reads as "no CI" rather than as "wrong API". Both are *the endpoint is
answering a different question than you asked*. Neither is discoverable from a
green run — only from comparing two reads of the same commit.

---

## The problem this PR solves

46% of CI failure heads trace to windows where **`main` itself was red**, not to
flaky tests. Every PR branched during such a window inherits the red and a human
triages a failure their diff cannot reach.

Both halves of the detector already exist: `devrc-ci-push-main` posts
`tekton/devrc-main-*` on every push, and `main-green-check.sh` reproduces a red
before alerting. The gap is **latency** — the deadman runs on
`OnUnitActiveSec=4h` from its own last *activation* with `TimeoutStartSec=5400`,
so a healthy cycle is up to **~5.5h**.

## Premise validation — it held, with a second correction

| fact | value |
|---|---|
| main pushes carrying a `tekton/devrc-main-*` status | **100 / 100** |
| authoritative verdict latency after the commit | **p50 19.6 min · p90 26.1 min** |
| commits that actually **get** an authoritative verdict | **21 / 100** |
| rows: `superseded by a newer run` | 119 / 200 (59.5%) |
| rows: `KILLED: the gate pod died` | 30 / 200 (15%) |
| rows: `NO GATE POD` | 10 / 200 (5%) |
| rows: **real `failure`** | 9 / 200 |
| median gap between consecutive authoritative verdicts | **1.43h** (max observed 13h) |

So the leg is reliable and fast **as a signal** — but **"main's status" is
usually not a verdict at all.** Only 21% of commits are ever validated; a newer
push cancels the older run.

### 🔴 What the design does with a non-verdict

**It claims nothing and takes no action** — absence is never treated as green.
The walk classifies `superseded` / `KILLED` / `NO GATE POD` / `pending` as
NOT-A-VERDICT and continues past them; if the whole window holds none, it prints
`no authoritative tekton/devrc-main-* verdict in the newest N commits` and
`Nothing is claimed about main. The 4-hourly deadman still covers it`, rc 0.

Pinned by three tests, and it is the half that would rot silently:
`test_an_error_state_is_never_a_green_either`,
`test_a_pending_only_commit_is_not_a_verdict`, and
`test_a_NON_VERDICT_is_never_treated_as_green` (5 consecutive superseded
commits). Mutant M2 — "a commit with no verdict reads as green" — is killed.

Consequently this **accelerates detection; it does not make it continuous**, and
the 4-hourly deadman remains the backstop, not a fallback.

## Design: an event-driven trigger, not a second alerting path

`scripts/main-status-watch.py` runs every 10 min, makes a handful of API reads,
and on an authoritative red **starts `main-green-check` early**. It never decides
anything and never notifies a human.

1. **A flake reaching it costs compute, not attention.** The deadman already
   separates flake from break by re-running — `VERDICT=flake` when attempt 1 is
   red and attempt 2 green, and only rc 10 (red on *both*) reaches
   `notify-failure@`. The click-through hazard is answered by **reproduction**,
   not by name-matching in the fast path.
2. **If it breaks, coverage is exactly what it was yesterday** — hence **no
   `OnFailure=notify-failure@`**. It still ladders (rc 12 after 12 consecutive
   unmeasured runs) so it cannot go blind in permanent silence.
3. **No `--force`** on the trigger: the deadman memoises on the sha, so it can
   never be made to overrule its own reproduction with a status row.

**Rejected.** *Lower `OnUnitActiveSec`* — each run is both sandbox tiers (~20
min) on a box carrying 20–40 concurrent runs; the "accelerate before deleting"
mistake. *Two consecutive reds* / *same failing test twice* — with 21% of commits
validated and a 1.43h median verdict gap, that buys hours of latency for a
distinction reproduction already makes better. *Trigger on every main push* —
~100 pushes in 2.5 days × 20 min each.

### Flake-vs-broken: made by reproduction, not by name-matching

The screen in the fast path is **sound and almost never satisfiable, and both
halves are the point.** Descriptions cap at 140 chars, and real rows prove the
list is untrustworthy: one reported `failed=7` while naming **one** test; one
named **none**; the row naming the known flake was cut **mid-word**
(`…_CAN_see_the_dif`). A description can prove *"at least one test failed"*; it
can never prove *"these are all of them"*.

So it skips only on **proven completeness** (every failure named, count agreeing,
every name on the ledger) and triggers on any doubt — **including on the real
known-flake row**. On the 100 commits measured it would have fired **zero**
times. That is correct, not a defect.

## Verification

**Red/green matrix.** The script is new at base `cdfd14ab`, so **nothing here is
regression coverage** — no test was ever red at a base containing it. All 40 are
invariant guards and behavioural contracts, labelled as such in the file header,
with two controls first (`test_NEGATIVE_CONTROL_the_harness_can_go_red`;
`test_POSITIVE_CONTROL_the_trigger_receipt_can_move_off_zero` — reported as a
pair: **1 on the positive control, 0 on the green case**).

**Mutation table — 24/24 killed, each by its intended test.** Control 43 passed /
0 failed; `PYTHONDONTWRITEBYTECODE=1`; whole file per mutant (no `-k` filter);
M1 as a known-caught positive control.

| # | mutation | killed by |
|---|---|---|
| M1 | `superseded` classified as red *(positive control)* | `test_an_error_state_is_never_a_red` |
| M2 | a commit with no verdict reads as green | `..._is_never_a_green_either` |
| M3 | keep oldest row per context, not newest | `test_a_stale_pending_row_does_not_outvote...` |
| M4 | missing `failed=` treated as a complete list | `test_a_LEDGER_MATCHING_name_with_no_count_still_triggers` |
| M5 | count disagreeing with names no longer blocks | `..._with_MORE_failures_than_named_triggers` |
| M6 | a red naming no test screened as all-flake | `test_a_red_naming_no_test_triggers` |
| M7 | debounce removed | `test_the_same_red_sha_is_handed_over_exactly_once` |
| M8 | walk does not stop at a green | `test_a_newer_green_wins_over_an_older_red` |
| M9 | foreign-context filter removed | `test_a_foreign_context_is_ignored` |
| M10 | blind ladder off by one | `test_the_blind_ladder_escalates...` |
| M11 | unreadable API reports OK not UNMEASURED | `test_an_unreadable_api_is_UNMEASURED_not_green` |
| M12 | failed trigger swallowed as success | `test_a_failing_trigger_is_UNMEASURED...` |
| M13 | a good run no longer resets the streak | `test_one_good_run_resets_the_blind_ladder` |
| M14 | known-flake ledger emptied | `test_the_ledger_contains_exactly_the_documented_flake` |
| M15 | total-budget check bypassed | `test_running_out_of_the_total_budget_is_UNMEASURED_not_a_kill` |
| M16 | budget raised above the unit timeout | `test_the_total_budget_is_under_the_units_timeout` |
| M17 | script's `TimeoutStartSec` belief drifts from `home.nix` | same |
| M18 | commit-entry `isinstance` check removed | `test_a_commit_list_of_non_objects_is_UNMEASURED_not_a_traceback` |
| M19 | loud status-row `isinstance` check removed | `test_a_status_row_that_is_not_an_object_...` |
| M20 | no-verdict return goes back to `None` | `test_the_no_verdict_return_carries_a_SLICEABLE_sha` |
| M21 | top-level exception net removed | `test_an_unexpected_exception_is_REPORTED_and_LADDERED...` |
| M22 | escalation env var read a second time (duplication returns) | `test_the_blind_ladder_default_lives_in_exactly_one_place` |
| M23 | `BUDGET` dropped from the documented env list | `test_every_env_var_the_code_reads_is_documented_in_the_header` |
| M24 | consolidated default changes value | `test_blind_escalate_parses_and_defaults` |

🔴 **Five mutants SURVIVED across the sweeps, and every one was a genuine test
defect** — all the same shape, *green for the wrong reason*:

- **M4, M5** — both count guards were **unreachable**: the real fixtures' names
  don't match the ledger anyway, so the *name* check decided the case and the
  count checks never executed. Deleting either left all 30 tests green. Fixed
  with fixtures whose name *does* match.
- **M14** — the test built its fixture by reading the ledger out of the
  implementation, so replacing the ledger with nonsense kept it green. *"Never
  derive a test's expectation from the implementation it tests."* Fixed with an
  independent literal plus a two-way pin.
- **M19** — the guard I had added to `newest_per_context` was **unreachable**
  (the walk rejects a non-object row first) *and* it `continue`d, silently
  dropping a row that could be the red. **Deleted** rather than tested: one rule,
  one place, and the surviving one fails loudly.
- **M21** — the unexpected-exception test fed a shape `find_newest_verdict`
  already rejects cleanly, so it passed with the net deleted. Rebuilt around a
  description that is a nested object — it passes every `isinstance` check and
  then explodes inside `classify`.

⚠ **Two mutants came back `BAD-PATCH` rather than SURVIVED** (M10, M22): my own
new code made their target expressions ambiguous, and the sweep **refused to
patch** rather than silently mutating the wrong occurrence — the `count=1`
replace hazard, caught by the harness instead of by me. Both re-run with
disambiguating patterns and killed.

### One rule, one place

The blind-ladder escalation default was open-coded at **both** ladder sites —
literal `12` *and* the parsing guard. The copies agreed, so this is not a bug
fix; the hazard is the next edit touching one of them, and the ladder's failure
mode is **silence**, so a divergence would not announce itself.
`blind_escalate()` now owns it. **Pure extraction, no behaviour change** —
`test_blind_escalate_parses_and_defaults` pins the behaviour it preserved
(including a non-numeric value falling back rather than raising, since it is
called on the failure path).

That read also found a real gap: **`MAIN_STATUS_WATCH_BUDGET` was read for a full
commit while nothing documented it** — the `MIN_TESTS` shape. Now in the header
(which *is* the `--help` text) and pinned **two-way**: read-but-undocumented
fails, documented-but-unread fails.

⚠ The remaining pyright hits (`float()`/`int()` over `str | None`) are **false
positives, deliberately not chased**: `(x or "")` absorbs `None` and `.isdigit()`
guarantees the conversion; pyright cannot narrow because the guard is on a
different expression than the conversion. A `# type: ignore` would be noise.

**End-to-end, production `gh`, real GitHub JSON, only the trigger stubbed:**

| arm | result |
|---|---|
| live `main` | walked past superseded/killed commits to the real verdict — `GREEN at c6700447 (walked 2)`, **rc 0**, no trigger |
| real historical red `504ea983` | detected, printed the real failing test name, **rc 10**, receipt written **once** |
| same evidence re-run | `already handed 504ea983 … not re-triggering`, **rc 0**, receipt still 1 line |

⚠ **On "does not fire on a single isolated red":** by design a single red *does*
trigger a **confirmation**, because triggering costs compute, not attention. The
"does not *alert*" property belongs to the deadman and is pre-existing and green:
`test_a_red_is_retried_EXACTLY_once_and_a_clearing_red_is_a_FLAKE_not_a_pass`.
This PR deliberately adds **no** new alerting path.

**Unit syntax.** `SuccessExitStatus=10 11` verified with `systemd-analyze verify`,
reported as a pair: **1 diagnostic on a deliberately-bad control**
(`notanumber` → `Failed to parse value`), **0 on the subject**.

### A launcher ledger caught this branch, and it was right

`test_no_real_launchers.py::test_every_hazardous_binary_the_scripts_reach_is_stubbed_or_acknowledged`
went red: `systemctl` sits in `ACKNOWLEDGED_UNSTUBBED` with a **two-way pinned
file set**, and `scripts/main-status-watch.py` is a new file reaching it.
Attributed rather than assumed — it reproduced on **both** trees of this branch,
`origin/main` is green over the same files, and it fails **standalone** as well
as under `--files`, so it is this diff and not a flake or a scoped-mode artefact.

The entry is re-justified on the **opposite ground** from every existing one.
`syshealth` and `tmux-restore-observe.sh` are acknowledged because their verbs
*are* reads, so the verb-splitting stub passes them through. This verb is
`start` — mutating on purpose. Two independent legs: **(1) seam** — the command
sits behind `MAIN_STATUS_WATCH_TRIGGER`, which every behavioural test sets, so no
test executes the production argv; **(2) fail-closed** — `start` is not on
`SYSTEMCTL_READ_VERBS`, so the stub *blocks* it. Leg 2 protects against a test
nobody has written yet.

🔴 **The pin arrives with the entry**, which is the one lesson that ledger's own
`tmux-reply-agent` paragraph asks anyone to carry — there, an acknowledgement
*blinded* the guard it was filed under. Controls watched, and the middle row
reproduces that blinding live:

| control | result |
|---|---|
| clean tree | 121 passed across both files |
| **second injected `systemctl` call site** | file set unchanged → `test_no_real_launchers` **stays green**; only the new pin fails |
| verb changed to `status` (a read) | pin fails, naming leg 2 as void |

The pin reads `SYSTEMCTL_READ_VERBS` itself rather than a copied literal, and
carries a positive control — an empty or renamed list would make the `not in`
assertion pass for every possible verb.

**Gate.** Base sha `cdfd14ab`. Iteration via `run-tests.sh --files` in the flake
devShell (`nix develop`) — 40 passed / 0 failed on the changed file. Repo-wide
`scripts/scoped-tests.sh` on the final tree `c066aafd`: **8980 collected, 8980
passed, 0 failed**, 83 files across 15 of 28 hermetic targets. ⚠ That is the
**dev-host tier**; `nix build .#checks…` (the sandbox tier) was not run locally —
Tekton runs it, and posted **success on all three checks** for `c066aafd`
(`devrc-pytests` 21994 passed, `devrc-nodetests` 1449 passed,
`devrc-cairn-client-runs`). The four public-repo content gates (captured text, captured markup,
public IPs, client hostnames) ran clean: 227 passed. ⚠ `tekton/devrc-*` on this
head is one sample from a tier measured ~30% genuine-failure; it is not upgraded
to a merge warrant.

## Round-1 adversarial audit — 5 findings, all fixed

No 🔴; the design core survived. **Every defect was in a *failure* path** — which
for unattended automation on a timer is where they matter most, because that is
the code that only runs once something else has already gone wrong.

| # | defect | measured before | measured after |
|---|---|---|---|
| **F1** | blind ladder could **never** escalate for a trigger failure — `reset_streak()` ran when the *walk* succeeded, so the bump could only return 1 | escalate=2 → `11,11,11,11,11,11`, all `streak 1/2`, all systemd **successes** | `11,12,12,12,12,12` |
| **F2** | unwritable receipt re-started the deadman **every run, forever** (written *after* a successful trigger, no error handling) | starts = `1,2,3` | writable `1,1,1` · unwritable `0,0,0` (fails closed) |
| **F3** | uncreatable state dir → permanently rc 11, permanently silent; sweep showed **nothing covered it** | `11,11,11,11` | `12,12,12,12` |
| **F4** | pin named 2 numbers; there are 3 — the trigger timeout isn't charged to the budget | worst case `120`, real `150` | asserts `budget + trigger < TimeoutStartSec` |
| **F5** | debounce keyed on the **verdict sha**, the wrong unit | 5 red verdicts → 5 starts | 5 red verdicts → **1 start** |

**F1 falsified a sentence I had written into `nix/home.nix`** claiming it ladders
"rather than going blind in permanent silence." It was that exact shape, in the
one arm where the trigger *is* the whole job. The comment now records that it was
false and names the test that enforces it.

**F2 was the only finding that made things worse than not having this PR at
all** — and the trigger is *correlated*: `ENOSPC` on `~/.cache`, which shares the
root filesystem with the nix store, so the disk pressure breaking the write is
produced by the very builds least worth re-kicking every 10 minutes.

### F5 — the judgement call, decided: key on the episode

Only ~22% of main commits get a verdict, so the newest *verdicted* sha changes
repeatedly inside **one** sustained red window — the 46%-of-failures scenario
this PR targets. Keying on it meant ~**47% duty cycle** of the most expensive job
on the box (1.43h median verdict gap vs ~40 min per deadman run) against ~17%
before.

And the amplification landed on **attention**, not just compute:
`main-green-check` carries `OnFailure = notify-failure@`, and its memo's red
branch exits `RC_RED` **without re-running anything** — so a commit flipping
`pending → failure` inside one interval fires the DND-defeating toast having
measured nothing. My "a flake costs compute, not attention" was true of one
isolated flake and **false of a sustained red**, which would have rebuilt the
exact click-through hazard this PR exists to remove.

So the debounce is now keyed on the **red episode**: one trigger, opened at the
first non-flake red, closed **only** by an observed green verdict. Worst case is
one extra deadman run per red window — the whole point of the unit — and the
steady state on a healthy `main` is zero. The duty cycle is now **stated** in the
unit comment rather than left silent.

### Two nits that changed what a reader would do

- *"The header IS the `--help` text"* was **false** — `--help` printed `__doc__`,
  one line, naming **zero** of the seven knobs. Rather than delete the sentence,
  `--help` now prints the header (as `main-green-check.sh` already does), so the
  claim is true **by construction**; the test refuses a stub rather than
  accepting a silent zero.
- The launcher ledger's leg (2) said a forgetful test *"gets an error"*. Half
  right: the blocked stub **exits 0**, so such a test would pass **vacuously**
  rather than fail loudly. The safety claim is unaffected — nothing on the host
  starts — but the wording would mislead the next author, which is precisely the
  failure that ledger's own `tmux-reply-agent` paragraph warns about.

**7/7 new mutants killed** by their intended tests (control 50 passed / 0
failed); each fix arrived with the test that detects its regression.

## 🔴 Not live until deployed

Adds a systemd **user** unit — nothing needs root, so nothing is staged under
`nix/system/`. It becomes live only after:

```
home-manager switch --flake ~/workspace/devrc --impure   # or scripts/ship.sh after merge
systemctl --user daemon-reload
```

**I ran neither.** Deploying from an unmerged branch would put un-merged code on
the live host. The timer is gated on the existing
`serverMode && enableMainGreenDeadman`, so it arms on the workbench and nowhere
else. All four files are `git add`ed (2 new, 2 modified) — a new file the flake
does not see is silently omitted from the deploy.

## Left open

- The 79% of status rows that are **not** verdicts is a blind spot in the
  *signal*, not in this change; the deadman covers it by construction. Making
  `devrc-ci-push-main` validate the tip after a burst is a separate, larger
  change.
- The flake screen is expected to fire ~never at current description lengths. It
  is kept because it encodes *why* name-matching cannot carry the flake decision,
  next to a ledger someone will otherwise grow. If still dead in a month, delete
  it — the deadman's double-run is the real defence.


ZacxDev:
Round-1 adversarial audit ran against `c066aafd` and returned 5 findings (no 🔴; verdict "merge after fixing F1 and F2"). The fixes below produced `b5308c73`.

Posting the claims block as an **issue** comment so the round-2 dispatcher can read it — `gh pr view --json comments` does not return review comments, and a missing block silently widens the next round's range.

```audit-claims round=1 audited=c066aafd..b5308c73
1. F1 — `reset_streak()` no longer runs before the red branch, so a repeated trigger failure escalates instead of pinning the streak at 1. Claimed measured: escalate=2 goes `11,12,12,12,12,12` where it previously gave `11,11,11,11,11,11`.
2. F1 follow-on — the `nix/home.nix` unit comment that asserted the watcher "still ladders … rather than going blind in permanent silence" is claimed corrected to record that the assertion was FALSE as written, and to name the test that now enforces it.
3. F2 — a failed receipt write no longer leaves the deadman started with no receipt. Claimed measured: unwritable receipt now gives `0,0,0` starts (fails closed) where it previously gave `1,2,3`.
4. F3 — the state-dir-creation failure path now escalates rather than returning rc 11 forever. Claimed measured: `12,12,12,12` where it previously gave `11,11,11,11`. Claimed deliberate that this arm goes straight to rc 12 rather than laddering, because the streak file lives inside the directory that could not be created.
5. F3 follow-on — a test is claimed to arrive with the fix, covering the arm whose `return RC_UNMEASURED` → `return RC_OK` mutant previously SURVIVED a fully green 44-test suite.
6. F4 — the budget/timeout pin is claimed widened from `TOTAL_BUDGET_S < TimeoutStartSec` to include `trigger_deadman()`'s own 30s timeout, which was not charged against `_DEADLINE`.
7. F5 — the trigger debounce is claimed re-keyed from the newest verdicted sha to the red EPISODE, closed only by an observed green. Claimed measured: 5 red verdicts now produce 1 start (was 5); a green closes the episode; a new red window produces a second start. The resulting duty cycle is claimed to be stated in the unit comment rather than left unanalysed.
8. Nit — `--help` is claimed to print the script header, so the two-way env-var guard's stated rationale ("the header IS the --help text") is true by construction rather than merely asserted; the test is claimed to refuse a stub rather than only a non-zero exit.
9. Nit — the launcher-ledger entry is claimed reworded to say plainly that the blocked `systemctl` stub EXITS 0, so a future test lacking the seam would pass vacuously rather than fail loudly. The safety claim (nothing real runs) is unchanged.
10. Verification claimed for the fix round: 31/31 mutants killed each by its intended test, control 50 passed / 0 failed, 167 passed across the three affected files, and `scoped-tests.sh` on `b5308c73` reporting 8986/8986 with 0 failed on the dev-host tier.
```

⚠ Not claimed, and open: no round-2 delta audit has been run over `c066aafd..b5308c73`. An audit fix resets the verification gate, so this ladder is not closed.

PR #1477, opened 2026-09-11
docs(gate-flake): rank 7 — a second gate flake sharing this doc's shape and nothing else

ZacxDev:
Files the `#1445` timeout flake where the gate-flake population is tracked, and corrects rank 1, which cannot be measured without both fixes.

## Rank 7 — the new flake

`test_the_subset_note_reports_N_of_the_FULL_set_not_N_of_N` (`test_run_tests_targets.py:756`) spawns a **nested full `run-tests.sh` run** as a subprocess under a hard `timeout=120` and is **SIGKILLed at the bound**: `subprocess.TimeoutExpired`, returncode `-9`, traceback ending in `_check_timeout`.

🔴 **It must not be folded into the store-api population.** It shares the observable — a check red on a diff that cannot reach it — and nothing else. The store-api flake is an `fsync` stalling inside a request and announces itself with `MECHANISM =`; grepping for that here finds nothing, because no store server is involved.

Added by **`#1445`**'s own audit ladder (the test's docstring opens "🟡 round-2 F5") and first observed red hours later. The property it pins is real — it kills a mutant that survived all 23 tests in its file. The **120 s bound on a nested full run** is the part that is not survivable on a contended node.

Ruled out, with values:

| | |
|---|---|
| any diff | `#1458`'s four files (store-api suite, siting ledger, `ci-repro/**`) cannot reach that file |
| determinism | `#1462` — same base, same test, docs-only — **passed minutes earlier**; locally `31 passed` in 205 s |
| general node load | failing vs passing run, CI-to-CI: **1030.34 s vs 1065.64 s** and **155.60 s vs 154.73 s** — the failing run was *faster*; one target ran ~23 % long, the rest flat |

⚠ **Not established:** what the nested run blocked on. No disk/CPU/PSI figure at the moment of the kill, and the pipelinerun and pods were **pruned within the hour** (`keep: 20`). `#1458`'s log was captured in time; `#1462`'s was not — the recovery recipe is in the item.

## Rank 1 — two corrections it cannot be measured without

1. **The anchor moved.** `#1458` (squash `ce9b55c3`) found that `#1211`/`#1219`/`#1239` sited **5** store roots and left **18** open-coded on disk — including the test that kept reddening the gate — and sited all 18. A rate measured against heads postdating `65f7325b` but predating `ce9b55c3` spans **two** interventions and credits the first.
2. **The population is no longer one mechanism.** Counting reds without classifying them will read rank 7 as this doc's flake failing to close.

Also retracts "a required check" there: measured 2026-09-10, `main` has **no required status checks, no rulesets, `enforce_admins: false`**. Advisory. The line now says to re-read the setting rather than cite the doc.

https://claude.ai/code/session_01C4p3xnvfqhRRjAUGKLK96v

### DIFF
--- a/claudedocs/handoff-gate-flake-store-api.md
+++ b/claudedocs/handoff-gate-flake-store-api.md
@@ -233,19 +233,26 @@ Not a bug — a measurement that would mislead if run as written.
    roots in `test_subsystem_store_api.py` and left **18** open-coded on disk, including the one
    test that kept reddening the gate — and sited all 18. **Verified: `ce9b55c3^` has exactly 5
    `store_siting.store_root(` sites and 18 `tmp_path / "store"`; after it, 1 remains (a
-   docstring).** Use **`ce9b55c3`** as the anchor. 🔴 **The contaminated window is heads that
-   postdate `65f7325b` with NO upper bound** — that set mixes heads carrying one intervention
-   with heads carrying two. Heads between the two shas carry only the first and are fine to
-   count *as* the first. Say which anchor any recorded number is against. Full evidence:
-   `handoff-cairn-oss-multi-instance.md` rank 22.
+   docstring).** Use **`ce9b55c3`** as the anchor. 🔴 **The contaminated set is "carries
+   `65f7325b`" with NO upper bound** — it mixes heads carrying one intervention with heads
+   carrying two. Heads carrying the first and not the second are fine to count *as* the first.
+   🔴 **AND THE PREDICATE IS ANCESTRY, NOT DATE — every "postdates" in this doc, including rank
+   1's own sentence above, is the wrong test.** A branch cut before an intervention and never
+   rebased has commits dated after it and does **not** carry it, so a date filter gives a wrong
+   denominator on the one measurement this rank exists to produce. Use
+   `git merge-base --is-ancestor <sha> <head>`. Say which anchor any recorded number is against.
+   Full evidence: `handoff-cairn-oss-multi-instance.md` rank 22.
    ⚠ **AND THE POPULATION IS NO LONGER ONE BOUND — see rank 7.** A store-api red and a
    `run-tests.sh` subprocess timeout are both "a check red on a diff that cannot reach it", so a
    rate that counts reds without classifying them will read rank 7's flake as this one failing
    to close. **Classify by the failing TEST before counting.** ⚠ Do **not** read that as "and
    the causes are unrelated" — rank 7 records why that is unestablished.
-   forcing: gate — a check has been failing PRs whose diff cannot reach it. ⚠ Advisory, not
-   required — measured, with the qualifiers that matter, in this doc's Gotchas and in
-   `claude/skills/tekton/SKILL.md` (`#1452`). Re-read the setting; do not cite either line.
+   forcing: gate — a check has been failing PRs whose diff cannot reach it. ⚠ **Advisory, not
+   required: measured 2026-09-10 and re-measured 2026-09-11 — `main` has no required status
+   checks, no rulesets, `enforce_admins: false`.** See this doc's Gotchas for the qualifiers
+   (the state is deliberate and is not yours to restore). 🔴 **`claude/skills/tekton/SKILL.md`
+   says the OPPOSITE** — "requires both", "`enforce_admins: true`" — and is STALE; an earlier
+   draft of this line cited it as corroboration. Re-read the setting live; cite no doc for it.
 2. **Fix the hung-server classifier's path sensitivity** — devrc,
    `scripts/tests/test_subsystem_store_api.py`, `_HUNG_SERVER_RULES` /
    `_why_the_server_did_not_answer`. Scan frames' SOURCE LINES, not filenames. Reproduction in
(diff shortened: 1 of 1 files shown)


ZacxDev:
## Round 0 + round 1 — verdict was **needs rework**, and rank 7 has been replaced, not patched

Both rounds ran blind of each other, read-only. Round 1 returned **four 🔴** and I re-derived every one. Most of the original item was not true, so `486d9c8c` replaces it: **37 lines / 440 words**, down from 56 / 683.

### What round 1 found

- **The unit was wrong.** Six 120 s bound sites, not one — `_run:107`, `_run_env:862`, inline `:658`/`:804`/`:830`/`:961`, covering **26 of 29** tests. `_run_env` is behind **2 of the 3** occurrences, so the old closing condition was satisfiable by fixing `_run` alone while the test that motivated the re-keying stayed killable — the identical walk the item's own ⚠ claimed to have closed.
- **"Ruled out: general node load" is now the LEADING hypothesis.** The flaking spawns are **one-target** runs at **57.20 s** and **47.10 s** against a 120 s bound; the same file on the same tree measures **205 s** and **428.40 s** — **2.09× ambient variance**. Amplifier: `run-tests.sh` runs the outer suite `-n` up to 8 while forcing the nested run to `PYTEST_JOBS=1`, same cgroup quota.
- **"Ruled out: determinism" rested on a false premise.** `8df1117a` does not contain `ce9b55c3`; `dc972398` does — the branch was rebased across `#1458`'s merge between the green and the red. It also contradicted rank 1 *in this same PR*.
- **"A nested FULL run" is false for both tests** — argv `--targets`, or `DEVRC_TARGETS` injected via `ENV_ONLY:855`, narrows each to one target. Half the Next probe had nothing to do.
- **A third occurrence was uncounted:** `#1454`@`c5e3123e`, and its diff **does** touch `run-tests.sh`, so the "cannot reach it" rule-out never covered it.
- **The provenance sentence has been wrong three times.** `#1445` never touched the file (`gh pr view 1445 --json files` → empty); `#289` (`ca088e70`) created it *with* `_run` and the bound, `#1073` (`809486fa`) added both flaking tests. **Deleted rather than corrected a fourth time** — nothing depended on it. The file has been **byte-identical since 2026-08-30**: no code changed before the reds, the load did.

### What round 0 found that no correctness round asks

- **It inverted the fix I was leaning toward.** I came in ready to close this PR and "just raise or remove the bound". Pricing that found **six** sites and two docstrings (`:74`, `:274`) citing the bound as the instrument that *caught* the original defect. 🔴 **Raise it; do not remove it** — a full set is ~28× a one-target run, so a bound with real headroom still catches that regression. Removing or "guarding" it deletes the detector.
- **A live cross-initiative claim collision**, still open: another session holds *"establish the `test_run_tests_targets.py` flake mechanism, then remove or guard it"* under a slug from an unrelated initiative, while `gate-flake-store-api-7` sits **FREE**. The exact-slug lock cannot see it. That session needs both halves — the bound is a detector **and** 2.1× headroom is too tight.
- **A duplicate**: the branch-protection retraction already exists in open `#1452` (`claude/skills/tekton/SKILL.md`), measured a day earlier. Rank 1 now cross-references instead of restating, keeping the qualifiers that copy carries.

### Also corrected in rank 1

The anchor sentence named the **wrong contaminated window** — heads between `65f7325b` and `ce9b55c3` carry only the first intervention and are fine to count as such; the hazardous set is "postdates `65f7325b`" with *no* upper bound.

### Added, because the item lacked them

The detector argument, and a **closing condition with a positive control** — the old one keyed on an absence in a 140-char GitHub status description that already truncates mid-token, and was satisfied by a rename, a skip, a deselect, or simply not looking.

**Round 0 trial: `ran: 2 · changed the outcome: 2`.**

```audit-claims round=1 audited=d10dab2b..486d9c8c
1. Rank 7 is keyed to all SIX 120 s bound sites in the file (`_run:107`, `_run_env:862`, inline `:658`/`:804`/`:830`/`:961`) covering 26 of 29 tests, not to `_run` alone.
2. "Ruled out: general node load" is removed; the measured headroom (57.20 s / 47.10 s against 120 s) and the 2.09x same-tree ambient variance are stated as the leading hypothesis.
3. "Ruled out: determinism" is removed; the green/red pair it rested on straddles `ce9b55c3` and is no longer cited as a same-base comparison.
4. "A nested FULL run" is removed everywhere; both flaking spawns are stated as one-target runs, and the Next-probe half that asked for scoping is deleted.
5. The `#1445` provenance claim is DELETED rather than corrected; the file's two commits (`ca088e70` #289, `809486fa` #1073) and its byte-identity since 2026-08-30 are stated instead.
6. The third occurrence `#1454`@`c5e3123e` is named, and the "diff cannot reach it" rule-out is gone.
7. The closing condition requires all six sites and is verified by a positive control (run under 2x load, watch every spawn finish) rather than by absence from a `FAILING:` line.
8. Rank 7 states that its relationship to the store-api flake is NOT established, and that the two are counted separately because their BOUNDS differ — not because their causes are known to.
9. Rank 1's anchor sentence names the unbounded window as the contaminated one, and its branch-protection note cross-references the Gotchas entry and `#1452` instead of restating it.
```

PR #1485, opened 2026-09-11
feat(tmux,bar): 90% scratchpad popups; move the slot colour legend into the i3status-rust bar

ZacxDev:
Two UX changes to the tmux scratchpad system: bigger popups, and the slot colour legend moves from tmux's `status-left` onto the i3status-rust bar, where all 20 slots fit instead of 14.

Three audit rounds have landed on top of that. **This body describes `7c449cdd`, the current head** — an earlier version of it described the opening commit and, by round 2, was telling readers that a fixed 🔴 was still open. Round 2's largest change (the shared-grammar marker) appeared in it nowhere.

## The round-2 change, first, because it is the largest

The legend and the generated tmux key bindings are two readers of one file, and they used to carry **a regex each**. They disagreed in both directions — a 9- or 12-digit hex colour (valid pango) got a key binding and no legend entry; a commented-out slot line got a legend entry advertising a hotkey bound to nothing.

So the grammar now lives in **one place**: a `# SLOT_ENTRY_RE:` marker line inside `scripts/tmux-scratch-slots.sh`, which both readers read out of the same bytes they already open for the table itself.

Round 2 found the guard on that property was **spelled, not structural**. It was a regex over nix source text — `re.search(r"^\s*slotRe\s*=\s*$", …) or not re.search(r'^\s*slotRe\s*=\s*"', …)` — whose first disjunct matched the then-current *formatting*, so the second never evaluated. **MEASURED at `ffe4e5d0`: two mutants that bound `slotRe` to a string literal both SURVIVED a full 47-test run**, one of them carrying the original disagreeing grammar, i.e. the bug this PR exists to fix, fully regressed. (Reproduced here, not quoted: `git archive ffe4e5d0` → apply → `47 passed` twice.)

The fix is structural:

* the nix reader moved to **`nix/programs/tmux/slot-table.nix`** — a pure-`builtins` function of a path, no nixpkgs, no channel — so the grammar it *uses* is an **evaluable value**;
* `slotRe` is `nix-instantiate --eval`'d and compared to the string Python compiled, **and to a moved marker**, which a literal cannot follow;
* the `bind -n M-<key>` block the real module **generates** is compared tuple-for-tuple to the legend's own parse of the same table;
* `nix/programs/tmux/default.nix` is asserted to contain **no `builtins.match` and no `builtins.split`** — the only two ways nix can classify a string — so it cannot carry a grammar at all;
* the nix/Python corpus cross-check now runs **both real readers** over a table on disk. It used to build its own `builtins.match` expression and so never opened `default.nix`.

Round 2 also found the shared grammar had **narrowed**: it used `[ ]` where the old nix regex used `[[:space:]]`, and the CANDIDATE patterns (the shortfall floor's yardstick) were `[ ]` too — so the floor narrowed in lockstep and was structurally blind. **MEASURED with one entry re-indented with a tab: bash 20, the legend 19 slots and no `?`, nix 19 bindings and no throw.** The marker now carries a literal TAB byte inside the brackets (the one spelling both engines accept — measured; the table's comment had claimed `[ \t]` and `[[:blank:]]` were the only alternatives and omitted it), and both candidate patterns are now deliberately **wider** than the grammar, pinned both ways.

## Per-file

| file | change |
|---|---|
| `nix/programs/tmux/default.nix` | `mkBind`: `-w 80% -h 80%` → `-w 90% -h 90%`. Derives every binding from `slot-table.nix`; carries no regex primitive. `slotTablePath` is overridable so tests can run the real module over a synthetic table. |
| `nix/programs/tmux/slot-table.nix` | **NEW.** The nix slot-table reader: reads the marker, classifies lines, checks the capture-group arity. Pure `builtins`, evaluable in isolation. |
| `.tmux.conf` | Alt+Shift+T picker popup 80%→90%; Alt+`[` "most recent `scratch-*`" popup 80%→90%; `status-left` loses the `#(scratch-status.sh)` interpolation (keeps `#S` + the prefix-indicator colouring verbatim); `status-right` loses `%H:%M` and `#H`, keeps `idle-update.sh`. |
| `scripts/i3status-scratchpads` | **NEW.** The bar block: all 20 slots as `<hotkey><window count>` in the slot's colour + bold when live, dim `#504945` and no count when absent, `scratch ?` when unmeasured. Finds tmux's socket itself; reads the grammar from the marker. |
| `scripts/tmux-scratch-slots.sh` | Carries the `# SLOT_ENTRY_RE:` marker — the single copy of the entry grammar. |
| `scripts/tmux-scratch-picker.sh` | Round 1: the detach branch now requires `$TMUX`. Round 2: finds the tmux socket the same way the legend does, and branches on **fzf's exit code** rather than on the emptiness of its output. |
| `nix/graphical.nix` | `scratchPickerCmd` + `scratchpadsBlock` in the `let` block; block inserted after `diskBlock`; two **unconditional** `home.file` entries (the block script, and the canonical slot table symlinked beside it as `scratch-slots.sh`). |
| `scripts/tests/test_scratchpads_block.py` | **NEW.** **59 tests** at head (27 at the opening commit, 47 after round 1), fixture-driven — no real tmux. |
| `scripts/tests/test_no_real_launchers.py` | Acknowledgement rows for `i3status-scratchpads` and `tmux-scratch-picker.sh`, each with a PIN. |
| `scripts/tmux-scratch-status.sh`, `scripts/README.md`, `claude/skills/devrc-dx/reference/{config-stack,verification}.md` | comment/doc claims corrected — nothing may still assert the legend lives in `status-left`, and the verification checklist no longer pins **6** of the 20 slot bindings. |

`scripts/tmux-scratch-status.sh` is **not deleted**; it is retained as a debugging/fallback renderer and its header now says exactly that.

### Line ~104 was widened beyond the original plan

The plan named only line 89 (the Alt+Shift+T picker). I also changed line ~104 — the popup inside the `bind -n M-[` "toggle most recent `scratch-*`" run-shell. It is the same class of scratchpad popup, and leaving it at 80% would have made one scratchpad popup visibly smaller than its twenty siblings for no reason. `fzf-tmux -p 80%,50%` at line 66 is **deliberately untouched** — that is the window switcher, a different feature.

### Design notes

* **`state` is always `Idle`.** A legend never demands attention; the per-slot colour rides inside pango markup, not in the block state.
* **No poller / cache / signal.** One local `tmux list-sessions` per 30 s tick — nothing can go stale, so `bar_freshness` does not apply. Same shape as `loadBlock` / `claudeRunsBlock`.
* **Unconditional on both hosts, in both places** (block + both `home.file`s). Scratchpads are plain tmux sessions and the Alt-key bindings are generated on every host, unlike the credential-backed count pills.
* **The `scratch ?` discriminant is the point.** "No scratchpad sessions running" is a *real* reading and renders the full dim legend, so a reading that could **not** be taken must look different. Unloadable slot table, a **partially**-parseable one (a shortfall against the file's own declared count), a missing `tmux`, a non-zero exit that is not `no server running`, and a **zero** exit whose output nothing parsed → all `?`. `no server running` with no socket found anywhere → the dim legend, because that genuinely is zero.
* **A partially-parseable table fails the nix BUILD** rather than shipping fewer keys. The legend's counterpart is `?`; here the only honest "unmeasured" is a build error.
* **The socket search is defence, not a fix for an observed defect** — see below.
* **Left-click** opens the existing fzf picker wrapped in a float terminal, modelled on `syshealthCmd` — i3status-rust runs clicks through `sh -c` with no controlling terminal, so a bare TUI is a silent no-op.
* **No icon glyph**, deliberately — no unverified nerd-font codepoint that could render as tofu.

## The `TMUX_TMPDIR` claim, corrected

Round 1 added socket-finding to the legend and justified it by asserting that `i3status-rust` "does not run in that shell" and faced "a PERMANENT `?`". **That premise was wrong.** MEASURED on the **live** `i3status-rs` process, `/proc/<pid>/environ`: it carries `TMUX_TMPDIR=/run/user/1000`, and under exactly that environment a bare `tmux list-sessions` answers correctly (19 sessions, 16 of them `scratch*`). The round-1 measurement was taken with the variable **unset** — a condition the bar is not in, and the same `env -i`-is-stricter-than-systemd error `nix/home.nix` already records making and retracting once.

The socket-finding code is **kept** — it is a strict improvement — and re-justified in all four places it was described. What is actually true, and no more:

* `TMUX_TMPDIR` is undeclared runtime state with **no established source**;
* the live bar **does** currently carry it, so **no live bug was demonstrated**;
* both failure modes (a persistent `?`, and a false zero via a stale `/tmp` socket) are real and reachable if it ever goes missing;
* therefore this is defence against a state **not currently observed**.

The picker now finds the socket the same way, so the pill and the click behind it cannot name different servers.

## Verification

**Nix.** `nix-instantiate --parse` OK on `default.nix`, `slot-table.nix` and `graphical.nix`. Both new `home.file` sources evaluate off this branch's flake:

```
.config/i3status-rust/scripts/i3status-scratchpads → /nix/store/jpjka7fvvmcc5gby5n9yqm94jliy7as2-i3status-scratchpads
.config/i3status-rust/scripts/scratch-slots.sh     → /nix/store/ph21bw358h22qsf2721xx12398k7y1hb-tmux-scratch-slots.sh
```

…and the **real home-manager config's** `programs.tmux.extraConfig` generates **20** scratch bindings.

**The generated bindings are byte-identical to `ffe4e5d0`.** The whole generated block was extracted from both trees (`cmp`, 5133 bytes each) — and `cmp` was shown to report a difference on a one-byte change, so the green is not an instrument that cannot go red.

Earlier round's bar-TOML evidence still holds — the block and its click land between `disk_space` and `net`, and a positive control through a throwaway `i3status-rs 0.36.1` showed the spans reach the i3bar line **unescaped** with `"markup":"pango"`, which is why every interpolated value goes through `html.escape`.

⚠ Worth recording from the first round: **before** `git add`, the eval failed with `path '/nix/store/…-source/scripts/i3status-scratchpads' does not exist` — a live reproduction of the "a NEW file must be `git add`ed or the flake silently omits it" hazard.

**The edited block, run against the real server** (read-only; nothing attached, nothing killed):

| environment | result |
|---|---|
| `TMUX_TMPDIR=/run/user/1000` (the live bar's actual value) | 20 spans, `state=Idle`, no `?` |
| `TMUX_TMPDIR` unset (the socket search doing its job) | 20 spans, `state=Idle`, no `?` |
| `TMUX_TMPDIR=/nonexistent-root` (a declared root with no socket) | `scratch ?` — it does **not** fall through and report `/tmp`'s server |

**Tests.** `test_scratchpads_block.py` + `test_no_real_launchers.py` + `test_bar_status.py` + `test_tmux_continuum_save_interpolation.py` + `test_doc_path_rot.py` + `test_skill_descriptions.py` + `test_skill_tiers.py` → **851 passed**.

`scripts/scoped-tests.sh` (inside the repo dev shell — the first attempt outside it exited 3 on a missing-toolchain FATAL, an environment problem, not a code one):

```
SCOPE: SCOPED (22 file(s) across 2 of 28 hermetic target(s) via --files)
RESULT: PASS (exit=0)
TOTAL collected=3975  passed=3975  skipped=0  failed=0
```

**This is not a gate** and is not quoted as one. It printed three UNMAPPED changed files it says nothing about: `config-stack.md`, `verification.md`, `tmux-scratch-status.sh`. It also suspended GUARD 7's required direction and GUARD 2's skip total, and replaced GUARD 3's floors with a one-test-per-file floor. CI closes the rest.

### Mutation matrix (round 2)

Every mutant is materialised into its **own** pristine `git archive` export (no `.git` at all), mutated with the narrowest edit that can be wrong, and run under `PYTHONDONTWRITEBYTECODE=1` with every `__pycache__` removed first. A mutant counts as KILLED only if the run fails **and the named test is among the failures** — "a test failed" is not enough.

**Baseline at `ffe4e5d0` (the round-1 tree), reproduced here:**

| mutant | round-1 verdict |
|---|---|
| A — `slotRe =` bound to a string literal on the next line, marker read left dead | **SURVIVED** — 47 passed |
| B — same shape carrying the **original disagreeing grammar** (`[[:space:]]*` + `#[0-9a-fA-F]+`) | **SURVIVED** — 47 passed |

**At head (control: pristine export → 59 passed, 0 failed):**

| # | mutation | verdict | killed by |
|---|---|---|---|
| A | `default.nix` inlines the reader and binds `slotRe` to a literal (today's grammar) | KILLED | `test_default_nix_CANNOT_CLASSIFY_A_LINE_ITSELF` |
| B | same, carrying the **original disagreeing grammar** | KILLED | `test_default_nix_CANNOT_CLASSIFY_A_LINE_ITSELF` (+ `…REFUSES_a_table_whose_colour_the_OLD_grammar_would_have_taken`) |
| A′ | `slot-table.nix` binds `slotRe` to a literal equal to today's marker | KILLED | `test_the_grammar_NIX_USES_is_the_MARKER_STRING_ITSELF_evaluated` |
| B′ | `slot-table.nix` binds `slotRe` to the original disagreeing grammar | KILLED | `test_the_grammar_NIX_USES_is_the_MARKER_STRING_ITSELF_evaluated` (+ 5 more) |
| C | the shared grammar loses the TAB (revert to `[ ]*`) | KILLED | `test_a_TAB_INDENTED_entry_is_a_slot_to_ALL_THREE_readers` |
| D | the **Python** candidate yardstick narrows back to the grammar's width | KILLED | `test_the_CANDIDATE_yardsticks_are_WIDER_than_the_grammar_on_BOTH_sides` |
| E | the **nix** candidate yardstick narrows back to the grammar's width | KILLED | `test_the_CANDIDATE_yardsticks_are_WIDER_than_the_grammar_on_BOTH_sides` |
| F | nix stops stripping trailing whitespace off the marker line | KILLED | `test_TRAILING_WHITESPACE_on_the_marker_line_is_stripped_by_BOTH` |
| G | `slot_pattern` stops checking the capture-group arity | KILLED | `test_the_GROUP_INDEX_contract_is_CHECKED_on_both_sides` |
| H | `slot-table.nix` stops checking the capture-group arity | KILLED | `test_the_GROUP_INDEX_contract_is_CHECKED_on_both_sides` |
| I | `_socket_roots` falls through past an explicit `TMUX_TMPDIR` again | KILLED | `test_an_explicit_TMUX_TMPDIR_wins_the_OUTCOME_not_merely_the_ORDER` |
| J | the picker stops finding the socket (bare `tmux` again) | KILLED | `test_the_picker_FINDS_THE_SAME_SOCKET_THE_LEGEND_DOES` |
| K | the picker treats **every** fzf exit as deliberate | KILLED | `test_the_picker_HOLDS_when_FZF_FAILS_TO_LAUNCH` |
| L | nix reads the slot NAME out of the wrong capture group (`elemAt m 4` → `3`) | KILLED | `test_the_BINDINGS_NIX_GENERATES_are_exactly_the_LEGENDS_OWN_PARSE` |

None died for the wrong reason. Several killed **extra** tests as well — e.g. A/B also fail the group-arity test because the inlined reader drops that check — which is legitimate collateral, not a substitute for the named kill.

The earlier rounds' matrices (M1–M6 on the discriminant branches, F1–F9 on the round-1 findings) are unchanged and still pass; the "Baseline before: 27 passed" figure in the original body was a snapshot of the opening commit and is superseded by the 59 above.

⚠ One limitation stated rather than implied: a **second grammar that agrees with the first today** is invisible to every behavioural test — it only bites when someone edits one copy. That is what `test_default_nix_CANNOT_CLASSIFY_A_LINE_ITSELF` covers, and it covers it by asserting the absence of a regex *primitive*, not the absence of a *spelling*. nix could in principle classify lines by hand with `substring`/`stringLength`; nobody has, and it would be far louder in review than a regex.

## 🔴 DEPLOY NOTE — restart-class, and deliberately NOT deployed

Adding a bar **block** is restart-class, not reload-class: i3status-rs reads its TOML once at startup. Landing this needs `home-manager switch` **and a bar restart** before anything appears on screen. That has **not** been done here, on purpose — raising or restarting the WM takes the operator's screen, and deployment is the operator's step. Nothing on the live bar, the live tmux server, or the operator's focus was touched by this work.

## Not verified — stated plainly

* The block has **never been seen on the real bar**, only through a throwaway i3status-rs instance reading a throwaway config. Colours/widths on the actual bar are unverified.
* The **left-click has not been clicked.** The float-terminal wrapper is reasoned from `syshealthCmd`'s measured precedent (no controlling terminal under `sh -c`); the picker behind it has been driven end to end with fake binaries and, in round 1, on a real pty against a throwaway server — but not from the bar.
* The picker's **hold** on an fzf that failed to launch is asserted structurally, not observed: with stdin on `/dev/null` bash suppresses `read -p`'s prompt and returns immediately, so the test proves the branch was taken, not that the terminal stayed open.
* The 90% popups were not visually confirmed — the tmux config change is not live until a switch + `source-file`.
* The 14-of-20 figure for the old tmux renderer is read off its `for (i = 7; i <= n; i++)` loop over a 20-entry table, not measured on screen.
* What `TMUX_TMPDIR=/run/user/1000` on the bar process **comes from** is unestablished. It is recorded as unknown rather than guessed.

### Fixed since an earlier version of this body said otherwise

* ~~"No `read -n 1` hold: every exit path is either long-lived or a deliberate instant exit."~~ **Both halves were false.** The picker carries its own hold on two paths (attach-and-create both failed; fzf failed to launch), and an fzf that never ran is *not* a deliberate exit.
* ~~"the click could dismiss a scratchpad instead of opening the picker … Unchanged by this PR … the fix belongs in the picker."~~ **That defect is fixed**, in the picker, in round 1 — and the body understated it: `tmux detach-client` run outside a client targets the server's best client, so it detached a live client of an **unrelated** session (measured: client-count 1 → 0, exit 0, no output). The detach branch now requires `[ -n "$TMUX" ]`.

https://claude.ai/code/session_01TpHAkviTvJNEybpsDFBHEJ


ZacxDev:
## Round-1 audit fixes — per-finding disposition

Head is now `ffe4e5d0` (two commits + a merge of `origin/main` at `c38c5484`).
Nothing here is deployed: i3status-rs reads its TOML once at startup, so a
`home-manager switch` **and a bar restart** are still the operator's step.

| # | disposition | one-line |
|---|---|---|
| F1 | **fixed** (click KEPT) | detach branch now requires `[ -n "$TMUX" ]`; outside-a-client path exercised end to end on a real pty |
| F2 | **fixed** | slot-count floor derived from the file, not a literal 20 |
| F3 | **fixed** | one grammar, read by both readers from the table's own marker line; cross-artifact test |
| F4 | **fixed** | rc 0 + non-empty stdout + nothing parsed → `None` |
| F5 | **fixed** (and the finding CORRECTED) | the block finds tmux's socket itself; the symptom was a permanent `?`, not a false zero |
| F6 | **fixed** (comment only, removal kept, no host pill) | |
| F7 | **fixed** | the two-test pair + the sibling-data-file pin |
| F8 | **fixed** | `sys.dont_write_bytecode` deleted |
| F9 | **fixed**, and the unverified claim **withdrawn** rather than restated | |
| F10 | **fixed**, plus two more stale claims the sweep found | |
| F11 | **no action**, as instructed | |

---

### 🔴 F1 — the click detached a live tmux client

**Reproduced before fixing.** Private `TMUX_TMPDIR` server, `$TMUX` unset (the
bar's click environment), one client attached to the **non-scratch** session
`work`, `scratch2` created last so it was MRU:

```
== PRECONDITION
   TMUX=<unset>
   MRU (tmux display-message -p '#{session_name}') = scratch2
   client=/dev/pts/58 session=work
   client-count-before=1
== RUN picker            rc=0   stdout=   stderr=
== AFTER
   client-count-after=0
   VERDICT: DETACHED a live client (the F1 defect)
```

After the fix, same harness, same run:

```
== AFTER
   client=/dev/pts/58 session=work
   client-count-after=1
   VERDICT: client SURVIVED
   stderr=inappropriate ioctl for device   <- fzf, reached instead of the detach
```

**The Alt+Shift+T path is unchanged — measured, not assumed.** A
`display-popup -E` child *does* get `$TMUX` (`TMUX=<socket>,<pid>,<sid>`;
`TMUX_PANE` is **not** set, which is why nothing in the script reads it), and
the inside-a-scratchpad toggle detaches identically before and after
(`client-count 1 → 0` both times).

**The click is KEPT because the outside-a-client path was verified to work**, on
a real pty rather than asserted:

* fzf renders — prompt `scratch>`, header `enter: toggle | type to filter/create`,
  the scratch sessions, `[+ new scratchpad]`, and `work` correctly **absent**;
* typing filters (`3/3` → `1/3`);
* Enter attaches — `client=/dev/pts/62 session=scratch5`, measured **while the
  picker was still running**.

The same run from inside a **non-scratch** session with `$TMUX` set also
attaches (a second client on `scratch2`) — tmux refuses nesting only when the
target contains the calling pane, and a popup is its own pty.

`exec` is gone from the final line so the both-failed path can print why and
hold the terminal open. `graphical.nix`'s comment — which claimed every instant
exit was *deliberate* — now records the reproduction instead.

### F2/F3/F4 — the discriminant narrower than its prose

**F2.** `if slots:` → a floor: entries the grammar accepted vs entries the file
*declares* (any quoted string alone on a line). Derived from the file, so a 21st
slot needs no edit. Measured on the shipped code first: corrupting 17 of 20
colour fields rendered `g3 G v`.

**F3/F9 — one grammar, not three.** The pattern now lives on the slot table's
own `# SLOT_ENTRY_RE:` marker line and **both** readers read it from there — the
same bytes they already open for the table, so no new deploy surface. Written in
the intersection of POSIX ERE and Python `re`, applied whole-line by both.
Colour widths are the ones pango accepts (3/6/9/12), so the old `{3,8}`'s 5-, 7-
and 8-digit admissions are gone.

Both measured disagreements are closed, and a corpus test proves it by running
the **real** marker through nix's own engine (`nix-instantiate --eval`, pure
builtins) and through Python:

| line | old nix | old python | now (both) |
|---|---|---|---|
| `#    "scratch21:z:#123456:zed"` (commented out) | reject | **accept** → a legend entry for a hotkey bound to nothing | reject |
| `"scratch21:z:#123456789:zed"` (9 hex, valid pango) | **accept** → a binding | reject → no legend entry | accept |
| `"scratch21:z:#12345:zed"` (5 hex) | accept | accept | reject |
| `"scratch21:z:#12345678:zed"` (8 hex) | accept | accept | reject |

A third test pins **bash `source`** against the regex on the real table, and the
`len == 20` literal is gone — the expected count now comes from bash, an
independent parser. Generated bindings verified **byte-identical** to the ones
this PR shipped. nix now `throw`s rather than generating a partial binding set
(its only honest spelling of "unmeasured"); both nix guards were watched failing.

**F4.** rc 0 + non-empty stdout + nothing parsed → `None`. Genuinely empty
stdout with rc 0 is still a real zero and is pinned separately.

### F5 — `TMUX_TMPDIR`, and a correction to the finding

The audit predicted a false zero. **Measured on the workbench (tmux 3.7c,
`TMUX_TMPDIR` unset): tmux says `error connecting to /tmp/tmux-1000/default (No
such file or directory)` — not `no server running`** — so the shipped code
rendered **`scratch ?` permanently** while 20 scratchpads were live. Confirmed
end to end in that exact environment:

```
BEFORE (HEAD version):  {"text": "scratch ?", "state": "Idle"}
AFTER:                  {"text": "<span foreground=\"#504945\">g</span> <span foreground=\"#d79921\" weight=\"bold\">G8</span> …
```

The false-zero variant is real but needs a stale socket — also measured: a
socket file left by a dead server makes tmux say `no server running`. Both are
closed: the block searches `$TMUX_TMPDIR`, `$XDG_RUNTIME_DIR`,
`/run/user/<uid>` (a **literal**, because that variable is as absent from a bar
block's environment as `$TMUX_TMPDIR` is) and `/tmp`, points tmux at what it
found, and accepts `no server running` as a zero only when no socket was found
anywhere.

### F6 — comment only; removal kept, no host pill

Verified: `timeBlock` renders `%a, %b %d | %H:%M` and **no hostname**, and there
is no host block anywhere in `graphical.nix`. The comment now says the clock
survives, the hostname is shown **nowhere**, and that this is a deliberate
accepted loss. Also recorded: `status-right` expands to a single space
(`idle-update.sh` prints **0 bytes**, measured; `continuum_save.sh` prints
nothing) so the right edge is blank by design. `status-right-length 70` /
`status-left-length 90` are ceilings, not widths — neither is wrong, both left
alone.

### F7 — the two nix-seam guards, plus the data-file pin

Both added, in the `runawaysBlock` idiom, plus a pin on the sibling
`scratch-slots.sh` `home.file`. Deleting that entry fails leg 1, finds no leg 2,
and **silently succeeds on leg 3** off `~/workspace/devrc` — the failure nobody
would see.

### F8 / F10

`sys.dont_write_bytecode` deleted (it ran after `exec_module` and the runtime
script imports nothing local). `config-stack.md`'s six-slot row fixed; the sweep
found two more stale claims in the same file — the monitor HUD covers all 20
(it sources the table), and the `N🔄 N⏸ N●` status-right counters were removed in
2026-08 along with the status-left `●` they were documented as pairing with.

### F9's second half — the unverified claim is WITHDRAWN, not restated

Measured (i3status-rs 0.36.1, throwaway TOML, `timeout 6`, rc 124): the block's
`<span …>` spans reach the i3bar line with `"markup":"pango"` — positive control
confirmed. And a block emitting a raw `a < b & c` reaches the line **verbatim**,
unescaped, unvalidated, with the other blocks still emitted. So i3status-rs will
not escape a stray `<` for us — reason enough to escape here — but **"breaks the
whole bar line, not just this block" was never verified and is withdrawn** from
both the test docstring and the script. What pango/i3bar do with invalid markup
was not observed, because observing it means rendering the operator's real bar.

---

### Mutation matrix — every new guard watched failing

`PYTHONDONTWRITEBYTECODE=1`, `__pycache__` deleted between mutants, every file
restored from a `copy2` pristine and byte-compared after each.

| mutation | failing test | its assertion |
|---|---|---|
| picker: drop the `$TMUX` conjunct | `test_the_picker_DETACH_branch_requires_TMUX` | "the detach branch is no longer guarded by `[ -n "${TMUX:-}" ]`…" |
| picker: restore `exec tmux new-session` | `test_the_picker_HOLDS_the_terminal_open_when_it_can_do_NEITHER` | "`exec` replaces this shell, so the failure hold below it can never run" |
| nix: re-point the click at `syshealthCmd` | `test_the_scratchpads_pill_CLICK_is_wired_to_scratchPickerCmd` | "…does not run `scratchPickerCmd` — it is wired to 'syshealthCmd'" |
| nix: drop the float terminal | `test_the_scratchPickerCmd_STRING_runs_the_picker_in_a_TERMINAL` | "the click must run in a TERMINAL…" |
| nix: delete the sibling `home.file` | `test_the_SLOT_TABLE_is_deployed_BESIDE_the_block_script` | "…would fall back to ~/workspace/devrc's working tree (leg 3)" |
| blk: revert the floor to `if slots:` | `test_a_PARTIALLY_parseable_table_is_UNMEASURED_not_a_short_legend` | "a table that parsed to 3 of 20 entries was accepted as a reading" |
| blk: drop the rc-0 guard | `test_fetch_sessions_rc_ZERO_with_unreadable_output_is_UNMEASURED` | `assert got is None` on 4 payloads |
| blk: accept `no server running` unconditionally | `test_no_server_running_is_a_real_zero_ONLY_WHEN_NO_SOCKET_EXISTS` | `assert … is None` with a socket present |
| blk: drop the literal `/run/user` root | `test_socket_roots_name_run_user_LITERALLY_not_only_via_XDG` | `AssertionError: ['/tmp']` |
| blk: stop handing tmux the found socket dir | `test_run_tmux_POINTS_TMUX_AT_THE_SOCKET_IT_FOUND` | `env["TMUX_TMPDIR"] == "/run/user/4242"` |
| table: widen the colour grammar to `{3,8}` **(isolated — group structure unchanged)** | `test_the_colour_grammar_is_the_ONE_PANGO_ACCEPTS` + the nix/python corpus test | "`"scratch21:z:#123456789:zed"` -> False, expected True (9 hex digits IS valid pango…)" |
| table: un-anchor the leading whitespace to `.*` | same two | "`#    "scratch21:z:#123456:zed"` -> True, expected False (COMMENTED OUT…)" |
| blk: inject `subprocess.run(["home-manager","switch"])` | `test_the_block_SPAWNS_TMUX_AND_NOTHING_ELSE` | "spawn set is ['home-manager', 'tmux'], not {'tmux'}" |
| blk: argv[0] from a variable | same | "spawn set is ['<computed>']" |
| blk: `os.system("true")` | same | "reaches a raw exec/system primitive: ['system']" |
| **positive control** — render a live slot in `DIM` | `test_render_colours_and_bolds_present_slots_and_dims_absent_ones` | (pre-existing test, killed as expected) |

nix-side guards, watched the same way with a pristine control before *and* after:

* one entry given a 5-hex colour → `error: … declares 20 slot entries but only
  19 match the `# SLOT_ENTRY_RE:` grammar — refusing to generate a partial set`
* marker line deleted → `error: … must carry EXACTLY ONE `# SLOT_ENTRY_RE:`
  marker line … Found 0.`

⚠ Two honest notes on the battery. My **first** attempt at the colour mutant
changed the group *count*, so several tests died for the wrong reason; it was
redone isolated. And my first attempt at the nix partial-table mutant
(`#cc2:poppy:extra`) **did not fire** — 3 hex is valid and `[^"]+` swallowed the
extra colon. A mutant that dies, or survives, for the wrong reason proves
nothing either way.

### Tests

* `test_scratchpads_block.py` + `test_no_real_launchers.py` on the merged tree:
  **127 passed**.
* Change-scoped run on the merged tree (`scoped-tests.sh`, in the dev shell):
  **`RESULT: PASS`, collected=3963 passed=3963 skipped=0 failed=0** across
  `scripts/tests` (21 files) and `test_guard_core.py`. `SCOPE: SCOPED` — not a
  gate, and it says so; CI's full run is what closes the gap.
* The explicitly-named files also green: `test_bar_status.py`,
  `test_tmux_continuum_save_interpolation.py`, `test_doc_path_rot.py`,
  `test_skill_descriptions.py`, `test_skill_tiers.py`,
  `test_ci_claim_matches_reality.py`.
* `nix-instantiate --parse` clean on both nix files; `nix eval` resolves **both**
  `home.file` sources to store paths, and `programs.tmux.extraConfig` still
  evaluates with the 20 generated slot bindings.

### Merged tree — one SEMANTIC conflict, found by running the tests

The git merge of `origin/main` (`c38c5484`) was clean; the tests were not.
main had widened `test_no_real_launchers.py` into a two-way pin on
`ACKNOWLEDGED_UNSTUBBED`'s file **sets**, while this branch added one clause of
comment prose naming `home-manager` to `scripts/i3status-scratchpads`.
`launcher_scan` is a text scan, so the file became a hit in a set that now has
to be exact → `Extra items in the left set: 'i3status-scratchpads'`. Neither
side is wrong and neither side's own suite could see it. Fixed the way that row
itself prescribes — re-justified rather than reworded, **with** the AST pin,
because the row records an acknowledgement *blinding* the guard it is filed
under. ⚠ Reproduced here too: with the acknowledgement in place and no pin, an
injected real `home-manager` spawn left `test_no_real_launchers.py` **green**.

### Not done / not verified — stated plainly

* **Nothing is deployed.** No `home-manager switch`, no bar restart, no
  `i3-msg`. Every claim above is about the source tree and about processes I
  started on a private tmux socket — never the operator's server, which was
  confirmed untouched at the end.
* **The scope of bad pango markup is unknown**, not asserted — see F9 above.
* **The laptop was not measured.** F5's environment finding was taken on the
  workbench; the block is unconditional on both hosts and the laptop's socket
  layout was not checked.
* `claudedocs/handoff-tmux-scratchpad-bar-statusline.md` (touched by main, not
  by me) still says this branch has "no commits yet" and "No PR yet". Stale, out
  of scope for this round, flagged rather than edited.

ZacxDev:
Round 1 audit findings and what the fix round claims to have addressed. Recorded so round 2 can diff against the right tip.

```audit-claims round=1 audited=bbaa4fa8..ffe4e5d0
1. F1 (🔴) the bar click ran `tmux detach-client` with `$TMUX` unset, detaching a live client — including one on an unrelated session. CLAIMED: detach branch now gated on being inside a client; click KEPT; `exec` dropped so the both-failed path holds the terminal; the "every instant exit is deliberate" comment corrected.
2. F2 (🟡) a partially-parseable slot table rendered a plausible short legend instead of `?`. CLAIMED: a slot-count floor derived from the file itself rather than a literal 20.
3. F3/F9 (🟡/🟢) three readers of the slot table used three different colour grammars, and the Python one admitted hex lengths pango rejects. CLAIMED: one grammar, read by both readers from a `# SLOT_ENTRY_RE:` marker line in the slot table; a cross-artifact nix-vs-python corpus test and a bash three-way test added.
4. F4 (🟡) `tmux` exiting 0 with unparseable stdout produced a full dim legend — a false zero. CLAIMED: rc 0 + non-empty stdout + nothing parsed now yields `None` (unmeasured).
5. F5 (🟡) the block depended on ambient `TMUX_TMPDIR`. CLAIMED: fixed by locating the socket and pointing tmux at it. ALSO CLAIMED: round 1's stated consequence was wrong — with `TMUX_TMPDIR` unset, tmux 3.7c emits `error connecting to …`, not `no server running`, so the shipped block rendered `scratch ?` permanently rather than a false zero.
6. F6 (🟡) a new comment falsely claimed the bar carries the hostname. CLAIMED: comment corrected only; the `%H:%M`/`#H` removal from tmux `status-right` is kept and no host pill was added (operator's explicit decision).
7. F7 (🟡) nothing pinned the block's `click` binding or the sibling slot-table `home.file` entry. CLAIMED: the two-test pair added, plus a pin on the sibling data-file deploy.
8. F8 (🟢) `sys.dont_write_bytecode = True` sat after `exec_module` and did nothing. CLAIMED: deleted.
9. F10 (🟢) `config-stack.md` still described six scratch slots. CLAIMED: fixed, plus two further stale claims found by a sweep.
10. F11 was left alone deliberately (bar width measured adequate; unconditional gating internally consistent).
11. CLAIMED: `origin/main` merged into the branch; one SEMANTIC conflict surfaced and fixed — main widened `test_no_real_launchers.py` into a two-way pin on `ACKNOWLEDGED_UNSTUBBED` while this branch added prose naming `home-manager`; resolved by re-justifying the row and adding an AST spawn pin.
12. NOT VERIFIED by the fix round, by its own statement: nothing is deployed (no switch, no bar restart); the laptop was never measured, so F5's environment finding is workbench-only; and the scope of malformed pango markup on a real bar is unknown — the earlier "breaks the whole bar line" claim was WITHDRAWN, not restated.
```

ZacxDev:
## Round-2 delta audit — per-finding disposition (`7c449cdd`)

| # | finding | disposition |
|---|---|---|
| 🟡1 | the flagship property was UNGUARDED (spelled guard, 2 mutants survived) | **FIXED, structurally.** See below. |
| 🟡2 | a narrowing regression (`[ ]` vs `[[:space:]]`), floor blind because the candidate narrowed in lockstep | **FIXED** — literal TAB in the grammar, both candidate yardsticks deliberately wider, pinned both ways. False "bash would not put it in the array either" corrected. |
| 🟡3 | F5's premise was wrong; re-derived a retracted `env -i` error | **FIXED in all four places**, code KEPT. |
| 🟡4 | the PR body presented a fixed 🔴 as open | **REWRITTEN** to match head. |
| 🟡5 | the verification checklist pinned 6 of 20 bindings | **FIXED**, and every touched file swept. |
| 🟢6 | one trailing space on the marker breaks the nix build, blaming the wrong thing | **FIXED** — nix strips trailing space *and* tab now. |
| 🟢7 | `load_slots`'s "Never raises" was false; group-index contract unguarded | **FIXED** — arity checked on both sides; `UnicodeDecodeError` also caught. |
| 🟢8 | `graphical.nix` enumerated the instant exits incompletely | **FIXED** in the comment **and** in the picker. |
| 🟢9 | "an explicit `TMUX_TMPDIR` still wins" won the ORDER, not the OUTCOME | **BEHAVIOUR fixed** — a declaration now ends the search. |

The five cleared items (marker ReDoS, absent/duplicate/invalid marker, commented-out entries, `mkBind` byte-identity, the merged `test_no_real_launchers.py` region) were left alone — except that `mkBind` byte-identity is now re-measured (below) because the module was refactored underneath it.

---

### 🟡1 — mutants A and B, before and after

Not quoted from the audit: reproduced here. Each mutant goes into its **own** pristine `git archive` export (no `.git` at all), gets the narrowest edit that can be wrong, and runs under `PYTHONDONTWRITEBYTECODE=1` with every `__pycache__` removed first.

**At `ffe4e5d0` (the tree the audit ran on):**

```
CONTROL                                    rc=0  47 passed
A (current grammar as a literal)           rc=0  fails=NONE -> SURVIVED   47 passed
B (the ORIGINAL DISAGREEING grammar)       rc=0  fails=NONE -> SURVIVED   47 passed
```

**At head (`7c449cdd`), control in the same batch:**

```
CONTROL (pristine)                         rc=0  59 passed
A   rc=1  KILLED by  test_default_nix_CANNOT_CLASSIFY_A_LINE_ITSELF
B   rc=1  KILLED by  test_default_nix_CANNOT_CLASSIFY_A_LINE_ITSELF
A'  rc=1  KILLED by  test_the_grammar_NIX_USES_is_the_MARKER_STRING_ITSELF_evaluated
B'  rc=1  KILLED by  test_the_grammar_NIX_USES_is_the_MARKER_STRING_ITSELF_evaluated
```

A/B are the audit's mutants re-created on the post-refactor tree (`default.nix` stops importing the reader and spells its own grammar, marker read left dead). A′/B′ apply the same two mutations **where the binding now lives**, `nix/programs/tmux/slot-table.nix`, which is what the evaluated-value guard is for.

Twelve more mutants, one per guard added or changed, are in the body's matrix — all fourteen KILLED by their named test, none for the wrong reason. Two of them fired for real during development rather than as a rehearsal: the picker's PATH jail caught `date`, a binary I had left out of the pinned set, and the first version of the jail's shims used `#!/usr/bin/env bash`, which cannot start inside the jail — that would have logged nothing while every "did it reach anything unexpected" assertion stayed green.

### 🟡1 — why it is three guards and not one

* `slotRe` is `nix-instantiate --eval`'d and compared to the marker string Python compiled — **and to a moved marker**, so a literal that happens to equal today's grammar cannot pass by agreeing.
* the `bind -n M-<key>` block the real module **generates** is compared tuple-for-tuple to the legend's own parse. That also pins the group-INDEX contract the marker cannot carry (nix `elemAt m 4`, Python `group(5)`).
* `default.nix` must contain **no `builtins.match` and no `builtins.split`**. This is the one part that is unavoidably textual, and it is deliberately a property of the artifact (it cannot classify a string at all) rather than a spelling of the grammar. ⚠ Stated plainly in the test: nix could in principle classify lines by hand with `substring`/`stringLength`; nobody has, and a second grammar that *agrees* today is invisible to any behavioural test.

### 🟡2 — the tab case, before and after

Measured on the real table with one entry re-indented with a TAB:

| reader | before (`ffe4e5d0`) | after |
|---|---|---|
| bash `source` | 20 entries | 20 entries |
| the legend | **19 slots, no `?`** | 20 slots, no `?` |
| nix | `nSlots=19 nCandidates=19`, **no throw**, 19 bindings | 20 bindings |

The spelling that works in both engines is a **literal tab byte inside the bracket** — measured: nix `builtins.match "[ <TAB>]*x[ <TAB>]*" "<TAB>x<TAB>"` matches, while `[ \t]` written with a backslash does **not** (a backslash inside a POSIX bracket expression is literal), and `[[:blank:]]` works in nix but not in Python. The table's justification named the two failures and omitted the one that works; it now records all three with the measurements.

Both candidate yardsticks are now wider than the grammar — `[[:space:]]`/`[ \t]` for the indent, and a trailing `.*` so an entry with a **trailing comment** counts as declared. That last one is the corrected claim: the corpus said such a line "is not a slot because bash would not put it in the array either", and **bash does** — measured, `${#SCRATCH_SLOTS[@]}` counts it. What is given up is stated in the table's own comment: a trailing-comment entry is deliberately not a slot, and it now trips the floor **loudly** (nix throws, the legend renders `?`) instead of being silently dropped by two of three readers.

### 🟡3 — what I measured, and what I am not claiming

`/proc/934021/environ` on the live `i3status-rs`: **`TMUX_TMPDIR=/run/user/1000`**. Under exactly that environment a bare `tmux list-sessions` returns 19 sessions (16 `scratch*`). With the variable unset it returns `error connecting to /tmp/tmux-1000/default`.

So: the variable is undeclared runtime state with **no established source**; the live bar **does** carry it; **no live bug was demonstrated**; both failure modes are real if it ever goes missing; therefore the socket search is defence against a state **not currently observed**. That is now the wording in the module docstring, the `_socket_roots` comment block, the `…ONLY_WHEN_NO_SOCKET_EXISTS` test docstring, and the `ACKNOWLEDGED_UNSTUBBED` justification. I did **not** go looking for a better-sounding reason, and I have not tried to establish where the variable comes from — that is written down as unknown.

**Corollary taken, not declined:** `scripts/tmux-scratch-picker.sh` now finds the socket the same way, so the pill and its click cannot name different servers. The duplication (Python + bash) is deliberate and flagged in both files; it is pinned by a test that runs the picker with a fake tmux and checks the root it picks against `_socket_roots()` under the same environment.

### 🟢9 — behaviour, not the claim

I fixed the behaviour rather than the claim: an explicitly declared `TMUX_TMPDIR` is now the **only** root searched, so `_run_tmux` can never override the operator's declaration with another server's socket. Measured against the live server with `TMUX_TMPDIR=/nonexistent-root`: the pill renders `scratch ?`, not `/tmp`'s sessions.

⚠ And a claim I wrote and then had to correct within the same session: I first documented that case as ending in tmux's `no server running` → a real zero. Measured, tmux says `error connecting to …`, so it is a `?`. Same direction, different mechanism; the docstring now says what was measured.

### Byte-identity re-check

The module was refactored, so `mkBind`'s output was re-measured rather than inherited: the whole generated binding block extracted from `ffe4e5d0` and from head is **byte-identical**, 5133 bytes each — with `cmp` shown to report a difference on a one-byte change, so the green is not an instrument that cannot go red. The real home-manager config's `programs.tmux.extraConfig` generates 20 bindings, and both `home.file` sources evaluate.

### Tests

`test_scratchpads_block.py` (59) + `test_no_real_launchers.py` + `test_bar_status.py` + `test_tmux_continuum_save_interpolation.py` + `test_doc_path_rot.py` + `test_skill_descriptions.py` + `test_skill_tiers.py` → **851 passed**.

`scripts/scoped-tests.sh` → `SCOPE: SCOPED (22 file(s) across 2 of 28 hermetic target(s))`, `RESULT: PASS (exit=0)`, `collected=3975 passed=3975 failed=0`. **Not a gate**, and it printed three UNMAPPED files it says nothing about.

`origin/main` was re-merged before this work (`3eb8332b`, disjoint files) and has not moved since. **The full gate was not run** and nothing here is quoted as one; CI is the automated signal.

### Not verified

* The bar has not been restarted and nothing has been deployed — that is the operator's step, and it takes the screen.
* The picker's **hold** is asserted structurally, not observed: with stdin on `/dev/null` bash suppresses `read -p`'s prompt, so the test proves the branch was taken, not that the terminal stayed open.
* The left-click has still never been clicked from the real bar.
* Where the bar's `TMUX_TMPDIR` comes from is unestablished.

https://claude.ai/code/session_01TpHAkviTvJNEybpsDFBHEJ

ZacxDev:
Round 2 delta-audit findings and what the round-2 fix claims to have addressed. Recorded so round 3 can diff against the right tip.

```audit-claims round=2 audited=ffe4e5d0..7c449cdd
1. 🟡1 the "one grammar" guard was SPELLED, not structural — its first disjunct matched the current formatting so the second never evaluated, and two mutants (one fully regressing F3/F9 to the original disagreeing grammar) SURVIVED a 47-test green run. CLAIMED: the nix reader moved to a new `nix/programs/tmux/slot-table.nix` (a pure-builtins function of a path) so the grammar nix USES is an evaluable value; three guards replace the spelled one — `slotRe` is `nix-instantiate --eval`'d and compared to the marker string Python compiled AND to a moved marker; the `bind -n M-<key>` block the real module GENERATES is compared tuple-for-tuple to the legend's parse; `default.nix` is asserted to contain no `builtins.match`/`builtins.split` at all. CLAIMED the corpus cross-check now runs both REAL readers over a table on disk, where before it built its own expression and never opened `default.nix`.
2. 🟡2 the new shared grammar used `[ ]` where the old nix regex used `[[:space:]]`, so a TAB-indented entry shipped 19 of 20 bindings with no throw and no `?` — a narrowing regression introduced by the round-1 fix, invisible because the candidate yardstick narrowed in lockstep. CLAIMED: a literal TAB byte inside the grammar's brackets (stated to be the one spelling both engines accept); both candidate yardsticks deliberately WIDENED and pinned both ways; the false "bash would not put it in the array either" claim corrected.
3. 🟡3 F5's justification asserted the bar "does not run in that shell" and that the pre-fix block was a PERMANENT `?`, but the live `i3status-rs` process carries `TMUX_TMPDIR`, so no live bug was demonstrated — an `env -i` probe measuring a condition the bar does not run in, which this repo already records as a mistake made and retracted once. CLAIMED: reason corrected in all four places, socket-finding code KEPT as defence against a state not currently observed.
4. 🟡4 the PR body was stale at head and still presented the fixed 🔴 as an open caveat, with a wrong test count. CLAIMED: body rewritten to match head.
5. 🟡5 `verification.md` still instructed a reader to check for 6 scratch-slot bindings. CLAIMED: fixed (6→20, derived from the table rather than a literal); sweep of every touched file found one further stale `== 20` literal in the test file.
6. 🟢6 a trailing space on the marker line broke the nix build with a message blaming 20 correct slot entries. CLAIMED: nix now strips trailing space and tab.
7. 🟢7 `load_slots`'s "Never raises." was falsified by a marker with the wrong capture-group count. CLAIMED: arity checked on both sides; `UnicodeDecodeError` also caught.
8. 🟢8 the instant-exit enumeration omitted an fzf that FAILS TO LAUNCH. CLAIMED: fixed in the comment and in the picker, which now branches on fzf's exit code.
9. 🟢9 an explicitly-declared `TMUX_TMPDIR` won the search ORDER but not the OUTCOME. CLAIMED: a declared `TMUX_TMPDIR` now ends the search.
10. CLAIMED: mutants A and B reproduced SURVIVING at `ffe4e5d0` (47 passed each) and both KILLED at head, along with A′/B′ applied where the binding now lives; 14 mutants total, all killed by their named test, none for the wrong reason; generated tmux bindings byte-identical to `ffe4e5d0` (5133 B, `cmp`, with a control proving `cmp` can report a difference).
11. NOT VERIFIED by the fix round, by its own statement: nothing is deployed and the left-click has still never been clicked from the real bar; the picker's hold is pinned STRUCTURALLY only (stdin on /dev/null suppresses `read -p`, so the test proves the branch was taken, not that the terminal stayed open); where the bar's `TMUX_TMPDIR` comes from is written down as UNESTABLISHED rather than guessed; a second nix grammar that AGREES with the marker today is invisible to every behavioural test, covered only by asserting the absence of a regex primitive; the full gate was not run.
```

PR #1494, opened 2026-09-11
fix(tmux-restore): display-message answers about the WRONG window, and the restore trusted it

ZacxDev:
**2026-09-11 incident.** A tmux server died holding **52 bound conversations**. The automated restore reported `relaunched 50 windows, skipped 0` and recovered **one**.

### Root cause, with a control
`tmux display-message -t <session>:<MISSING-INDEX>` does not fail — it answers about the session's **current** window and exits 0. Measured live, where `scratch3` had exactly one window:

```
display-message -p -t scratch3:1  -> '1:zsh'  rc=0
display-message -p -t scratch3:87 -> '1:zsh'  rc=0   <- the lie
```

continuum restored the sessions but only **one window each**, so 33 of 52 windows were absent. `window_state` said they existed, `new-window` was never called, and `send-keys` resolves the same way — so 50 resumes collapsed onto the ~18 windows that did exist. `_verify_sends` then re-read through the same lying predicate. `list-windows` **enumerates** rather than resolving, so there is nothing for tmux to be helpful about.

### Two more failures from the same night
- Every continuum-restored pane answered `claude: command not found` — a restored pane doesn't re-run the login profile, so its PATH can predate the current generation. Now resolved in *this* process, with a fallback to the bare name.
- The 15-minute autosave fired **mid-recovery**, saw a half-restored workspace, and repointed the plan from 52 entries to 37. `richer_generation()` now warns with the exact `--plan` command, and `--best` switches automatically. Warning rather than silently switching is deliberate.

### Coverage — red at base, green at HEAD
**Behavioural** reds (4), including the incident reproduced exactly (`['scratch3:1','scratch3:1','scratch3:1','scratch3:1']`). **New-symbol** reds (3) are labelled as such and *not* counted as behavioural coverage, per the NIT-6 lesson from #1415. One is a positive control that the new warning stays **quiet** when nothing was lost. 125 passed at HEAD.

🔴 **The double had to be made less forgiving than it first was** — the end-to-end test **passed at base** until `_FakeTmux` reproduced `send-keys`' fallback as well as `display-message`'s. Second time this week a too-permissive double hid a real defect (#1467 was the first).


ZacxDev:
## audit-claims — round-1 fix pass on #1494

Per prior finding: **fixed / partially / not / made worse**.

| # | finding | status |
|---|---|---|
| 🔴 1 | `claude_command()` inert under the unit's PATH | **FIXED** in `46a03152` — which landed AFTER the audit read head `19948061`, so it was already fixed when the finding was filed |
| 🔴 2 | `richer_generation` is a permanently-red alarm | **FIXED** — predicate narrowed to the immediately preceding generation |
| 🔴 3 | session half of the resolution bug still open; docstring claimed otherwise | **FIXED** — exact `#{session_name}` match + `list-sessions`; docstring corrected |
| 🟡 4 | `--best` measured staleness against the plan it was not using | **FIXED** — suppresses the gate, like `--plan` |
| 🟡 5 | `_verify_sends` now honest ⇒ new exit-1 into a DND-bypassing toast | **NOT FIXED — stated in the PR body instead.** Correct behaviour; the consequence is now documented so the first toast is not read as a new bug |
| 🟢 6 | `--best` missing from the flags block | **FIXED** |
| 🟢 7 | unreachable `except OSError`; `AttributeError` can escape | **PARTIALLY** — now catches the `AttributeError`/`TypeError` that CAN escape |
| 🟢 8 | test stubs return a list where `_verify_sends` returns an int | **NOT FIXED** — harmless, output discarded |
| 🟢 9 | `claude_command()` called per entry | **FIXED** — hoisted |
| 🟢 10 | commit-message control quotes a different format string | **NOT FIXED** — history; conclusion was independently reproduced by the auditor |
| 🟢 11 | `new-window` without `-d` moves the active window | **FIXED** |

```audit-claims round=1 audited=19948061..345c4fbd
1. `claude_command()` resolves PATH, then `~/.nix-profile/bin/claude` through `realpath` to an immutable store path, then the bare name — verified to return a `/nix/store/...` path under the unit's own three-entry PATH.
2. `richer_generation()` compares ONLY against the immediately preceding generation, so it is silent on a healthy history; measured `None` on the live plan where the previous predicate reported 18 lost ids across 137 of 137 generations.
3. `window_state()` carries `#{session_name}` and compares it exactly, so a prefix session (`scratch2` vs `scratch20`) is no longer answered from the wrong session; its docstring no longer claims `list-windows` cannot fall back.
4. `tmux_session_exists()` enumerates via `list-sessions` and compares exactly, instead of `has-session`, which prefix- and fnmatch-matches.
5. `--best` sets `staleness_hours = None`, so an explicitly chosen older generation is not refused for being old.
6. `new-window` gains `-d`; `claude_command()` is hoisted out of the per-entry loop; `--best` is documented in the flags block; `richer_generation` catches the `AttributeError`/`TypeError` that can escape `bound_ids`.
7. `_FakeTmux` now PREFIX-resolves session names like tmux, having previously matched exactly — it modelled tmux as stricter than tmux and could not have shown finding 3.
8. Two mutants killed by their own named guards: reverting the predicate to the oldest generation, and dropping the `got_sess == sess` comparison. 128 passed.
9. Verified live and read-only against the operator's real 55-window workspace: 55/55 existing windows classified correctly with zero disagreements; `scratch2:87` and `scratch:1` both report MISSING; `tmux_session_exists('scratch')` is False; `richer_generation` on the live plan is None.
```

PR #1495, opened 2026-09-11
fix(scoped-tests): the header justified the script with a figure CLAUDE.md had already retracted

ZacxDev:
## The defect

`scripts/scoped-tests.sh`'s **WHY THIS EXISTS** header stated, as the measured reason for the script's own existence:

> `runs bucketed by overlap go 14.5 min (0 others) -> 48.9 min (6+). The lever is not a faster suite. It is FEWER FULL SUITES.`

`CLAUDE.md`, **in the same tree**, retracts exactly that figure:

> *An earlier version of this line said "14.5 min alone → 48.9 min at 6+ overlapping"; that is **RETRACTED**, and re-deriving it is the trap. Bucketing runs by how many others overlapped them is length-biased — a long run overlaps more runs BY CONSTRUCTION — and a null Monte Carlo with zero interaction reproduces the shape, the 14.5-min baseline and ~2.06x of the 3.4x. Contention is real and its mechanism is uncontroversial; that dataset cannot size it.*

So one tree carried two opposite claims about one number, and the live one was the justification for a script. `claudedocs/handoff-gate-speed-and-ci-signal.md` retracts it too — `scoped-tests.sh` was the only site still asserting it. Nothing noticed, because a retraction in prose cannot reach a comment in a shell script.

By this repo's own doctrine — *"a comment is a claim too"* — that is a defect, not a nit.

## The fix

The header now says "dozens of concurrent full suites" with **no magnitude**, and records the retraction in place so it is not re-derived.

**The two numbers that actually justify the script are untouched**, and are now explicitly labelled as not retracted:
- the 20.1-min median over 237 real runs
- `pytest scripts/tests --collect-only` = **60.0s** before a single test runs, vs **0.64s** for one file

Neither depends on the overlap bucketing. The case for scoping is unaffected.

## The guard, and why it took four versions

Each version was defeated by a *different* mechanism, and each was found by mutation rather than by reading.

**v1 pinned proximity** — "a retraction appears within 20 lines". It survived re-inserting the bare claim into `scoped-tests.sh`'s header, because the retraction note the same commit added sat ten lines below and satisfied the window. No window size fixes that: an in-place edit is always adjacent to the note that excuses it.

**v2 pinned a per-file count.** It survived too, for an unrelated reason — the killing mutant *replaces* the quoting line with an assertion built from the same two tokens, so the count is identical whether you count lines or matches. A count cannot tell quotation from assertion.

**v3 pinned normalised text**, which is RULES.md's rule for a prose artifact. That closed the in-place case, but narrowed `FIGURE` to require the literal `min` — re-opening the regression class the guard exists for.

**v4** fixes that (`min(?:ute)?s?`) and makes the comment-lead strip's `*` branch **require a following space** — it was eating one star of a markdown `**BOLD**` run at a line start. The branch is kept, not dropped: `* ` is still stripped as a bullet.

| mutant | v1 | v2 | v3 | v4 |
|---|---|---|---|---|
| M3 in-place re-assert, note retained | **SURVIVED** | **SURVIVED** | KILLED | KILLED |
| M2 full header revert | killed | **SURVIVED** | KILLED | KILLED |
| M5 spelled-out unit (`48.9 minutes`) | (blind) | (blind) | **blind** | KILLED |
| M4 hyphenated (`14.5-min`) | (blind) | (blind) | KILLED | KILLED |
| N1 bare re-assert, new line | KILLED | KILLED | KILLED | KILLED |
| N3 re-assert in an unledgered file | KILLED | KILLED | KILLED | KILLED |
| N2 blinded scanner (positive control) | KILLED | KILLED | KILLED | KILLED |
| control | GREEN | GREEN | GREEN | GREEN |

⚠ **Two claims made in earlier commit messages here are RETRACTED**, both caught by the round-2 delta audit:

- *"reflow-safe by construction"* — **false**. The `**BOLD**` cause is fixed; hyphen-break sensitivity is real, unfixed, and now stated in the docstring and in the failure message rather than denied. ⚠ A further retraction, round 3: this bullet used to carry "v3's digest moved at 3 of 7 reflow widths, the count version at 1 of 7" and **named none of the seven**, so nobody could check it. Those numbers are **withdrawn, not replaced** — an independent sweep agreed on the direction and disagreed on the proportions, and I am not publishing numbers I did not measure. Reproducible substitute: rewrap CLAUDE.md to width 80 and the normalised output changes. The direction is what the retraction rests on, and it stands.
- *"strictly wider than what a proximity test could see"* — **false**. A ledgered site whose retraction sits beyond `CONTEXT` chars can lose that retraction with the digest unchanged. Coverage overlaps, it does not nest. Latent for the three current sites; a fourth entry inherits it, and the note says so.

What the text pin genuinely buys over a count is catching an in-place reword. That claim survives.

## Tests

🟢 **CI green at `760c8769` — including the tier this PR's headline finding was about.**

| check | result |
|---|---|
| `tekton/devrc-pytests` (the sandbox tier) | **pass** — collected 22,083 · passed 22,081 · failed **0** |
| `tekton/devrc-nodetests` | pass — 1,449 tests, 0 fail |
| `tekton/devrc-cairn-client-runs` | pass |

That is the authoritative answer to round 1's 🔴: the guard no longer reds the tier that gates.

**12 collected, 12 passed** — `test_scoped_mapper.py` + `test_retracted_contention_figure.py`, measured at `1c3b59b2`.

⚠ An earlier version of this body said "13 passed … at `c38c5484`". Both halves were wrong and are corrected here: `c38c5484` is this PR's **base**, where the test file does not exist, and the count dropped to 12 when round 1 deleted the subsumed proximity test. A commit message also claimed this correction had been made when it had not — the round-2 audit caught the claim-about-a-claim.

Sandbox-tier repro, with a discriminating control, in a no-`.git` replica of the head tree:

| | before | after |
|---|---|---|
| `test_retracted_contention_figure.py` | **3 failed** (exit 128) | 2 passed |
| `test_doc_path_rot.py` (carries the fallback) | 76 passed | 76 passed |

Scope-limited by design: a comment change plus one test file, with the full-tier ritual deleted per the re-scoped merge policy.

## Found by

Round 0 (the requirements & deletion pass added in #1440), running on #1445. Round 0's job is to ask whether a thing should exist — which surfaces the case where the stated reason for its existence is no longer true.

### DIFF
--- a/scripts/scoped-tests.sh
+++ b/scripts/scoped-tests.sh
@@ -5,10 +5,23 @@
 #
 # 🔴 WHY THIS EXISTS — the measured problem, not a hunch. The pytest tier's
 # median wall time is 20.1 min over 237 real runs (p90 38 min). It is not slow
-# because any one thing is slow: it is slow because ~27 concurrent agent
-# sessions each run the FULL ~21k-test suite on one 24-core box, and runs
-# bucketed by overlap go 14.5 min (0 others) -> 48.9 min (6+). The lever is not
-# a faster suite. It is FEWER FULL SUITES.
+# because any one thing is slow: it is slow because dozens of concurrent agent
+# sessions each run the FULL ~21k-test suite on one 24-core box. The lever is
+# not a faster suite. It is FEWER FULL SUITES.
+#
+# ⚠ DO NOT PUT A SIZE ON THE CONTENTION EFFECT HERE. This header used to read
+# "runs bucketed by overlap go 14.5 min (0 others) -> 48.9 min (6+)", and that
+# figure is RETRACTED — `CLAUDE.md` retracted it and says in terms that
+# re-deriving it is the trap, while this file went on asserting it as the
+# measured reason for its own existence. Bucketing runs by how many others
+# overlapped them is LENGTH-BIASED: a long run overlaps more runs BY
+# CONSTRUCTION, and a null Monte Carlo with zero interaction reproduces the
+# shape, the 14.5-min baseline and ~2.06x of the 3.4x. Contention is real and
+# its mechanism is uncontroversial; that dataset cannot size it.
+#
+# The two numbers above and below this note are NOT retracted and are what
+# justify the script: the 20.1-min median, and the collection cost measured
+# next. Neither depends on the overlap bucketing.
 #
 # `run-tests.sh --targets` has existed for a while and buys little on its own,
 # because the target that matters is one monolith. MEASURED 2026-09-08 in this
--- a/scripts/tests/test_retracted_contention_figure.py
+++ b/scripts/tests/test_retracted_contention_figure.py
@@ -38,7 +38,7 @@ from pathlib import Path
 
 REPO = Path(__file__).resolve().parents[2]
 sys.path.insert(0, str(REPO / "scripts"))
-from testlib.public_ip_scan import repo_files  # noqa: E402
+from testlib.public_ip_scan import SKIP_SUFFIXES, repo_files  # noqa: E402
 
 # The figure, in the spellings it has actually been written in.
 #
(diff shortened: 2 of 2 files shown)


ZacxDev:
## Round 1 — blind adversarial audit, findings fixed in `2d89291a`

Verdict was **merge after fixing 🔴 1**. The 🔴 was reproduced before being accepted, and the headline 🟡 was reproduced as a *surviving mutant* before being fixed.

| mutant | v1 (`d6a5aa42`) | v2 | v3 (`2d89291a`) |
|---|---|---|---|
| M3 in-place re-assert, retraction note retained | **SURVIVED** | **SURVIVED** | KILLED |
| M2 full header revert | killed | **SURVIVED** | KILLED |
| N1 bare re-assert on a new line | KILLED | KILLED | KILLED |
| M4 hyphenated spelling | (blind) | (blind) | KILLED |
| N3 re-assert in an unledgered file | KILLED | KILLED | KILLED |
| N2 blinded scanner (positive control) | KILLED | KILLED | KILLED |
| control | GREEN | GREEN | GREEN |

Sandbox-tier repro, with a discriminating control, in a no-`.git` replica of the head tree:

| | before | after |
|---|---|---|
| `test_retracted_contention_figure.py` | **3 failed** (exit 128) | 2 passed |
| `test_doc_path_rot.py` (carries the fallback) | 76 passed | 76 passed |

⚠ Correction to this PR's body: it labelled its "13 passed" run `c38c5484`, which is the PR's **base**, where the test file does not exist. The number was measured at the head.

```audit-claims round=1 audited=d6a5aa42..2d89291a
1. the test no longer calls `git ls-files` directly — `_tracked_text_files` goes through `testlib.public_ip_scan.repo_files`, which carries the no-.git fallback, so the sandbox tier that gates is no longer red
2. the binding assertion pins the normalised TEXT around each occurrence instead of a per-file count, so an in-place rewrite of a quoting line into an assertion (same tokens, same count) now fails
3. the text pin is reflow-safe: each file is collapsed to one whitespace-normalised string with comment markers stripped before matching
4. `FIGURE` requires the `min` unit on both branches, so an unrelated `48.9%`/`48.9 MB`/`$48.9` no longer produces a false accusation
5. `FIGURE` accepts the HYPHENATED spelling (`14.5-min`), which CLAUDE.md's own retraction uses and which was previously invisible to the guard
6. the module docstring states the different-digits blind spot under its own heading instead of implying both directions are covered
7. the binary-suffix skip was widened (.sqlite/.db/.woff/.zip/...) since `repo_files` filters directories only
8. the proximity test was DELETED as subsumed by the ledger's unledgered check, rather than widened
```

ZacxDev:
## Round 2 — delta re-audit of `d6a5aa42..2d89291a`, findings fixed in `1c3b59b2`

Verdict was **safe to merge**, with two 🟡s the auditor recommended fixing anyway because they are the guard's own correctness. Both were reproduced here before being accepted. Round 1's two closures were independently re-verified with controls and hold.

🔴 **PAYLOAD CLASSIFICATION, decided once and recorded here:** `scripts/tests/test_retracted_contention_figure.py` is **PAYLOAD** for this PR, not scaffolding. Revert test — the stated deliverable is "correct a false claim *and* guard it", so reverting the test ships half of it. This is the standing devrc exception: a scanner that IS a repo gate, on a PR whose deliverable is that gate. The attribution gate therefore does not fire on these rounds; the ladder is governed by findings.

```audit-claims round=2 audited=2d89291a..1c3b59b2
1. `FIGURE` matches the spelled-out unit (`minutes`/`mins`/`minute`), closing a false-negative the round-1 narrowing introduced — `48.9 minutes at 6+ overlapping` was invisible to v3 and is caught now
2. the three named false positives (`48.9%`, `48.9 MB`, `$48.9`) are still excluded after that widening
3. `_COMMENT_LEAD`'s `*` branch requires a following space, so a markdown `**BOLD**` run at a line start is no longer half-eaten by normalisation
4. the "reflow-safe by construction" claim is RETRACTED in the docstring, in the failure message and in the commit message, and replaced with the measurement that contradicts it (digest moved at 3 of 7 reflow widths; the count version moved at 1)
5. the failure message no longer tells a reader that reflow cannot be the cause
6. the "strictly wider than a proximity test" claim is RETRACTED; the deletion note now records the measured latent gap and warns a fourth ledger entry inherits it
7. the deletion note names NEAR_LINES=20 lines rather than CONTEXT chars, and no longer asserts a false failure that was never observed
8. `_BINARY_SUFFIXES` derives from `testlib`'s `SKIP_SUFFIXES` with a named local addition, instead of being a fourth divergent copy
9. the PR body is ACTUALLY corrected this time — sha, count (12 not 13), mutation table and the stale "counts lines" paragraph
10. digests for all three ledgered sites are unchanged, so none of the above moved any pinned text
```

ZacxDev:
## Round 3 — delta re-audit of `2d89291a..1c3b59b2`: **no code defect. The ladder stops here.**

Round 3 found nothing wrong with the code. Every mechanism it touched was re-verified with its own control, both axes (wider pattern, narrower normalisation) were checked in both directions, and mutants M3 / M5 / N3 were independently re-killed. Its two 🟡s were prose, and both are fixed in `760c8769`.

### 🔴 Correction to the round-2 comment above — my payload classification was wrong

That comment argued the test file is **PAYLOAD** by the revert test, and concluded "the attribution gate therefore does not fire on these rounds". **The conclusion does not follow, and the gate does fire.**

The revert test is sound about the **file across the whole PR** — reverting it ships half the deliverable. I applied it to the **range**, which is a different unit. Measured:

| range | files changed | `scripts/scoped-tests.sh` |
|---|---|---|
| `d6a5aa42..2d89291a` | test file only, +175/−125 | **untouched** |
| `2d89291a..1c3b59b2` | test file only, +57/−28 | **untouched** |

Revert either range and the shipped artifact is still fixed, the guard is still present, and it still passes — v3 and v4 produce identical hit sets and identical digests on this tree. Round 2's auditor called the range "100% scaffolding" and was right *about the range*; I was right *about the file*; we were measuring different things and I used my answer to rule out a gate that was scoped to the other one.

**Two consecutive rounds changed zero payload lines, so the ladder has left the PR.** That is the stop condition, and it agrees with round 3's own verdict.

### Fixed in `760c8769`

- 🟢 The docstring claimed "the digest moved at 3 of 7 reflow widths, the count version at 1 of 7" and **named none of the seven**. RULES.md requires naming the points measured, so it was unreproducible as written. **Withdrawn, not restated** — an independent sweep agreed on the direction and disagreed on the proportions, and I am not publishing numbers I did not measure. Replaced with a reproducible instruction (rewrap CLAUDE.md to width 80; the normalised output changes on a hyphen break), which I ran.
- 🟢 Two blocks credited the same reflow failures to different causes. The star block now says it fixes the star cause only.
- 🟡 The PR body said v4 "drops the `*`". It does not — the branch is kept and now requires a following space. Corrected above.

### Ladder summary

| round | outcome |
|---|---|
| 0 | `requirement questioned` — produced this PR |
| 1 | 🔴 red in the sandbox tier + 🟡 guard walkable by proximity |
| 2 | 🟡 pattern narrowed past the unit; "reflow-safe by construction" measured backwards |
| 3 | **no code defect; stop** |

Merging.

PR #1501, opened 2026-09-11
fix(obs-read): --json carried labels only as a rendered display string

ZacxDev:
## The defect

`rows[].metric` in `--json` is the output of `_fmt_labels()` — `{a=1, b=2}` — a
string built for the human-readable table. It was the **only** representation of
a label set in JSON output, so any machine consumer had to regex a rendered
display string to recover a single label.

## How it was found

Querying `kube_job_status_start_time` for job durations, I wrote a parser against
Prometheus's native `data.result[].metric` dict — the shape you'd reasonably
expect from a tool that proxies Prometheus. It returned `[]`, and I reported
"0 series" **three times** before catching it.

🔴 **The silent-zero guard was not at fault and did its job**: the very same
document said `row_count: 5` and `matched_nothing: false`. The shape is what
pushed the caller into hand-parsing, and the hand-parse is what failed.

What finally caught it was a contradiction no data can produce: `count(...)`
returning **5** while `topk(3, ...)` returned **0 series** on a *less* restrictive
selector. A filter cannot increase results, so the fault had to be in my parser.

## The change

Adds `rows[].labels` — the label set as a dict — to prometheus vector/matrix and
loki matrix rows.

Strictly **additive**:
- `rows[].metric` unchanged (the table still reads it)
- `qr.columns` unchanged, and `render_table` selects by column name, so it
  cannot see the new key — table output is unchanged by construction
- no flag, no behaviour change to any existing consumer

```
row_count: 3
  job_name = verify-b2-29815500  ns = storage-resolver     # no regex
```

## Tests

Six new, **verified RED at `origin/main`** (`_raw_labels` does not exist there)
and green at HEAD. Suite goes **74 → 80 pass**.

They pin the dict is present and correct, that it is *additive* rather than a
replacement, that it survives into the rendered JSON document (the defect was in
`--json` specifically, so one test drives `render()` rather than the parser),
that values are coerced to `str` to match both backends' wire promise, and that
a junk/`None` label set does not raise — scalar and string Prometheus results
carry no metric dict at all.

## Not done

`parse_pyroscope` builds its own row shape (`function`/`self_pct`) and has no
label set, so it is untouched.

### DIFF
--- a/claude/skills/obs-read/SKILL.md
+++ b/claude/skills/obs-read/SKILL.md
@@ -177,10 +177,16 @@ validated preset; treat unvalidated ones as starting points.
 🔴 **`rows[].metric` (prometheus) and `rows[].stream` (loki streams) are RENDERED
 DISPLAY STRINGS** — `{a=1, b=2}` — built for the human-readable table. They are
 NOT objects and they are NOT Prometheus's `data.result[].metric`, despite the
-name. Field-accessing them RAISES — `jq` prints `Cannot index string with string`
-and exits **5**, Python raises `TypeError` — and under a suppressor (`.metric.job?`,
-`// empty`, a bare `except`) it instead returns nothing silently, while `row_count`
-sits in the same document saying the query matched.
+name. Field-accessing them RAISES — `jq` prints `Cannot index string with string` and
+exits **5**; Python raises `TypeError` on `row["metric"]["job"]` and
+`AttributeError` on `row["metric"].get("job")`. Only `?` / `try…catch` (jq) or a
+bare `except` (Python) make it silent, and a silent miss is the dangerous one:
+`row_count` sits in the same document saying the query matched.
+
+🔴 **`// empty` is NOT an error suppressor** — jq's `//` is a null/false
+alternative, so the indexing error propagates before it can apply and the
+pipeline still exits 5 (measured, jq 1.8.1; `jq -n 'error("boom") // 1'` errors
+too). `.labels? // {}` is safe because of the **`?`**, not the `//`.
 
 **`rows[].labels` is the label set as a `{str: str}` dict.** Use it:
 
--- a/claude/skills/obs-read/SKILL.md
+++ b/claude/skills/obs-read/SKILL.md
@@ -177,21 +177,32 @@ validated preset; treat unvalidated ones as starting points.
 🔴 **`rows[].metric` (prometheus) and `rows[].stream` (loki streams) are RENDERED
 DISPLAY STRINGS** — `{a=1, b=2}` — built for the human-readable table. They are
 NOT objects and they are NOT Prometheus's `data.result[].metric`, despite the
-name. Field-accessing them returns nothing, silently, while `row_count` sits in
-the same document saying the query matched.
+name. Field-accessing them RAISES — `jq` prints `Cannot index string with string`
+and exits **5**, Python raises `TypeError` — and under a suppressor (`.metric.job?`,
+`// empty`, a bare `except`) it instead returns nothing silently, while `row_count`
+sits in the same document saying the query matched.
 
 **`rows[].labels` is the label set as a `{str: str}` dict.** Use it:
 
 ```bash
-obs-read --cluster dpprod --backend prometheus --kind instant --json \
+OBS=~/workspace/devrc/scripts/obs-read   # NOT on $PATH — see the rule above
+$OBS --cluster dpprod --backend prometheus --json \
   --query 'kube_job_status_start_time{cluster="dp-1"}' \
   | jq -r '.rows[] | "\(.labels.job_name) \(.labels.namespace)"'
 ```
 
 Present on prometheus **vector** and **matrix** rows and on loki **matrix** and
-**streams** rows. Absent on prometheus **scalar**/**string** results, which have
-no label set at all — so read it with `.get("labels", {})` / `.labels? // {}`
-rather than assuming every row carries one.
+**streams** rows. Absent on prometheus **scalar**/**string** results and on ALL
+**pyroscope** rows (`{function, self_samples, self_pct}` — a flamebearer has no
+label set) — so read it with `.get("labels", {})` / `.labels? // {}` rather than
+assuming every row carries one.
+
+⚠ **Cost, measured, so it is a choice and not a surprise:** each label set is
+serialised twice — rendered into `metric`/`stream` for the table, structured into
+`labels` for parsers — so `--json` is roughly **2x** larger on a label-rich query
+(measured on production loki: 235 streams, 141,096 -> 289,342 bytes). The
+duplication is what keeps `metric`/`stream` byte-identical for existing readers.
+Pipe through `jq` and select what you need; do not cat a wide `--json` raw.
 
 🔴 **Read `row_count` and `matched_nothing` before concluding anything from an
 empty parse.** A zero from your own parser is a fact about your parser; those two
--- a/claude/skills/obs-read/SKILL.md
+++ b/claude/skills/obs-read/SKILL.md
@@ -171,3 +171,29 @@ validated preset; treat unvalidated ones as starting points.
   `Handling connection for P` line per connection and blocks at 64 KiB unread).
 - Known limitation (documented, unchanged): a matched-nothing result still exits
   0 — check the `--json` `matched_nothing`/`warning` fields to fail a pipeline.
+
+## The `--json` row schema — read a label from `labels`, NEVER from `metric`
+
+🔴 **`rows[].metric` (prometheus) and `rows[].stream` (loki streams) are RENDERED
+DISPLAY STRINGS** — `{a=1, b=2}` — built for the human-readable table. They are
+NOT objects and they are NOT Prometheus's `data.result[].metric`, despite the
+name. Field-accessing them returns nothing, silently, while `row_count` sits in
+the same document saying the query matched.
+
+**`rows[].labels` is the label set as a `{str: str}` dict.** Use it:
+
+```bash
+obs-read --cluster dpprod --backend prometheus --kind instant --json \
+  --query 'kube_job_status_start_time{cluster="dp-1"}' \
+  | jq -r '.rows[] | "\(.labels.job_name) \(.labels.namespace)"'
+```
+
+Present on prometheus **vector** and **matrix** rows and on loki **matrix** and
+**streams** rows. Absent on prometheus **scalar**/**string** results, which have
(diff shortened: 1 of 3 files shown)


ZacxDev:
## Round 1 audit — findings and disposition

Docs-validated first (Prometheus HTTP API + Loki HTTP API, both "latest" channel,
fetched 2026-09-11; neither page carries a recoverable build date).

**Phase 0 result:** prometheus handles all four documented `resultType`s; labels
are a flat `{str: str}` object in both APIs, so `rows[].labels` is the
conventional shape; `str()` coercion is identity on any conforming payload and
correct as a defence otherwise. `parse_loki` does NOT conflate `.stream` and
`.metric` — each branch reads its own key.

| # | finding | disposition |
|---|---|---|
| 🟡-1 | `rows[].metric` still means a display string, not the API's label object | **WONTFIX, documented.** Renaming is the breaking change this PR avoids. The skill now warns explicitly. |
| 🟡-2a | loki matrix `labels` untested — mutant survived 80/80 | **FIXED** `5b197f53` — reproduced the surviving mutant, added a test, mutant now dies against that test alone |
| 🟡-2b | loki **streams** still emitted a rendered string — the unfixed instance of the bug being fixed | **FIXED** `5b197f53` — reproduced, fixed, 3 tests, all watched RED |
| 🟡-3 | new key undocumented in the skill | **FIXED** `5b197f53` |
| 🟡-4 | `--json` 2.28x larger on a label-rich query | **ACCEPTED, now stated.** Re-measured independently: **2.30x** (60 rows, 8 labels, 14034 -> 32274 bytes). Duplication is what buys the zero-breakage guarantee. |

**Re-derived, not accepted:** the first commit's "RED at main / 74->80" both hold.
The audit corrected my PR body in my favour — I wrote "(the helper does not exist
there)", implying import-shaped failures, but **4 of the 6 fail with
`KeyError: 'labels'`** (behavioural); only the 2 private-helper unit tests are
import-shaped.

**Pre-existing, NOT introduced here, recorded so they are on the record:**
- Prometheus native histograms (`histogram`/`histograms`) are unhandled → `value:
  null` with `matched_nothing: false`.
- Loki `vector` resultType has no branch and would produce a **false MATCHED
  NOTHING**. Currently **unreachable** — `build_url` only ever emits
  `query_range`, which returns `matrix` or `streams`. Latent; flag before anyone
  adds a Loki instant-query mode.

**Cross-repo follow-up, not fixable here:**
`civitai/talos-infra` `.claude/skills/inventory-disks/reference/manual-audit-queries.md:126`
instructs agents to parse `rows[].metric` **with a regex, never with field
access**. This PR makes that advice actively wrong. Closing condition: a merged
talos-infra PR amending that paragraph to point at `rows[].labels`.

```audit-claims
round=1
from=bfb925fc
to=5b197f53
findings=4 fixed, 1 wontfix-documented
payload-files=scripts/obs-read
scaffolding-files=scripts/tests/test_obs_read.py, claude/skills/obs-read/SKILL.md
```

ZacxDev:
Re-posting the round-1 claims block in the format `audit-dispatch.py` parses — my
earlier comment put `round=`/`from=`/`to=` inside the fence instead of on the
header line, so the script refused round 2. The prior comment's prose stands; this
is the machine-readable half.

```audit-claims round=1 audited=bfb925fc..5b197f53
1. Loki STREAMS rows now carry a structured `labels` dict, not only the rendered `stream` display string.
2. The Loki matrix branch's `labels` is now covered by a test; previously a mutant deleting it passed the whole suite.
3. Four Loki tests added (one matrix, three streams), claimed watched-RED with both Loki branches reverted.
4. The `--json` row schema is now documented in `claude/skills/obs-read/SKILL.md`, including which result types omit `labels`.
5. A comment in `parse_loki` states that Loki sends streams labels under `stream` and matrix labels under `metric`.
6. `rows[].metric` was deliberately NOT renamed; the display-string meaning is unchanged and the skill warns about it instead.
```

ZacxDev:
## Round 2 — findings and disposition

Round 2 was dispatched **blind** (claims stated as *what was claimed*, never *why*).
It re-derived all six round-1 claims independently — including against the **real
production Loki wire format** rather than the repo's fixtures — and ran a 10-mutant
battery with a positive control and `__pycache__` purged between mutants.

**Confirmed by that round, not by me:** claims 1, 2, 3, 5, 6 actually fixed. The
`stream`→`metric` conflation mutant IS caught. The matrix mutant dies against
exactly one test. All four Loki tests go red on revert, every failure behavioural
(`KeyError`), none import-shaped. Table output byte-identical across **11** branches.
No surviving mutant.

| # | finding | disposition |
|---|---|---|
| 🟡 F1 | the new example used a bare `obs-read`, against this file's own 🔴 `$PATH` rule 116 lines above — the copy-paste path fails on line 1 | **FIXED** `4e158109`. Reproduced (`command -v obs-read` → rc 1). Fence now sets `OBS=` **inside itself**, so it is self-contained rather than depending on one 118 lines up. Ran the example verbatim: works against the patched copy; the base copy returns `has("labels") == false` as the control. |
| 🟡 F2 | `--json` ~2x larger on the streams path, undisclosed | **ACCEPTED, now STATED in the skill.** Round 2's production measurement (235 streams, 141,096 → 289,342 B) is in the doc, with why the duplication is the price of byte-identical `metric`/`stream`. |
| 🟢 F3 | "returns nothing, silently" measurably false | **FIXED** `4e158109`. Re-measured myself: jq exits **5** with `Cannot index string with string`; Python raises `TypeError`; only `?` / `// empty` / bare-except are silent. Telling a reader it fails silently is worse than saying nothing — they stop looking for the loud error. |
| 🟢 F4 | absence list omitted **all pyroscope rows** | **FIXED** `4e158109`. |
| nit | redundant `--kind instant` (prometheus default) | dropped in the same edit |

**Not reproduced, recorded as the auditor left it:** the skill's *"disagreed with a
hand-written parser three times in a row once"* anecdote is unverifiable from the
range. It is my own account of this PR's origin; harmless either way.

**Ledger (round 2's own measurement, rc 0, stderr silent, non-empty range):**
`round 2 · payload lines changed THIS round: 34 (since round 1: 31)` — SKILL.md 26/0
and `scripts/obs-read` 7/1 both payload; `test_obs_read.py` 54/0 scaffolding. The
attribution gate does **not** fire: this round changed payload.

```audit-claims round=2 audited=5b197f53..4e158109
1. The `--json` example fence now sets `OBS=` inside itself and no longer invokes a bare `obs-read`.
2. The "field access fails silently" sentence now says it RAISES (jq exit 5 / Python TypeError) and names the suppressors that do make it silent.
3. The `labels`-absent enumeration now includes all pyroscope rows, not only prometheus scalar/string.
4. The ~2x `--json` size cost is now stated in the skill with its measured figures.
5. No code changed this round — `scripts/obs-read` is untouched; suite still 84.
```

ZacxDev:
## Round 3 — findings, and STOPPING THE LADDER on a stated criterion

Round 3 was blind. It re-derived all five round-2 claims and **both numeric claims
it was told to distrust reproduced**: the ~2x figure landed within 1% of its own
independent production measurement (147 streams, 88,137 → 179,693 B, ratio 2.039,
against the doc's 2.051), and jq exit 5 / Python TypeError are real.

| # | finding | disposition |
|---|---|---|
| 🟡 F1 | `// empty` named as an error suppressor — it is not | **FIXED** `ab29e44f`. Re-measured: every spelling exits 5; `jq -n 'error("boom") // 1'` errors, so `//` catches nothing. The doc now says so explicitly, because `.labels? // {}` is safe due to the **`?`** — the sentence was teaching the wrong half of its own idiom as load-bearing. |
| 🟢 F2 | "Python raises TypeError" true only for subscript; the doc's own `.get()` idiom raises `AttributeError` | **FIXED** `ab29e44f`. Both now named with the access each belongs to. |
| — | commit message transposed two line distances (116/123, not 116/118) | **ACKNOWLEDGED, not amended** — a pushed commit message, no reader consequence. Recorded here instead. The same message's "no gate catches it" claim was verified correct. |

---

## 🔴 STOPPING HERE — and naming why the rounds would not stop on their own

Round 3 returned a finding that needed fixing, so the findings-keyed rule would
normally mean a round 4. I am invoking the **prose-payload escape hatch** instead,
and the skill requires that be stated rather than left implicit — an unqualified
stop is indistinguishable from convergence.

**Why they will not stop by themselves:** the payload here is prose, so every round
changes payload by construction and the two-zero-rounds attribution gate is
**structurally inert**. Round 2's finding and round 3's finding are the *same defect
class* — a false claim about whether this failure is loud or silent — in the *same
six-line paragraph*, now rewritten three times. Round 3's own verdict: *"the ladder
is no longer finding new classes, it is finding residue of its own last pass… a
round 4 on this paragraph is overwhelmingly likely to find only the next
parenthetical."*

**The stop conditions are met:** no 🔴 in any round · blast radius is "a doc sentence
is wrong", nothing executable · and the recurring SHAPE was swept rather than the
reported instance — every factual claim in the replacement paragraph was re-run
before commit (jq rc for four spellings, both Python exception types, the `//`
control), not just the one the audit named.

**What I am NOT fixing, so the next reader knows it is open rather than absent:**
nothing outstanding in this PR. The two items deliberately carried forward are
recorded on the round-1 comment — `rows[].metric` keeps its display-string meaning
(renaming is the breaking change this PR avoids), and the ~2x `--json` size cost is
accepted and now documented.

**Ladder summary:** 3 rounds · 🔴 0 · 🟡 5 · 🟢 3 · all fixed or explicitly accepted.
Payload since round 1: 88 lines (`scripts/obs-read` 39, `SKILL.md` 49); scaffolding
114. Round 1 found the Loki gap that made the original fix half a fix; rounds 2 and
3 found defects in the prose written to document it.

**Cross-repo follow-up, unchanged and still open** (cannot be fixed in this repo):
`civitai/talos-infra` `.claude/skills/inventory-disks/reference/manual-audit-queries.md:126`
instructs agents to parse `rows[].metric` with a regex and never with field access.
This PR makes that advice wrong. Closing condition: a merged talos-infra PR pointing
it at `rows[].labels`.

```audit-claims round=3 audited=4e158109..ab29e44f
1. `// empty` is no longer described as an error suppressor; the doc now states it is not one and that `?` is what makes `.labels? // {}` safe.
2. The Python exception sentence now names TypeError for subscript access and AttributeError for the `.get()` idiom the same paragraph prescribes.
3. No code changed this round; suite 84 and doc-path-rot 76 both green.
4. LADDER STOPPED on the prose-payload criterion — no round 4 is planned.
```

PR #1504, opened 2026-09-11
docs(cairn): retract "there is no CREATE route" — it shipped in #1254, and FIVE tracked files still asserted it

ZacxDev:
## What

`claude/skills/cairn/SKILL.md` asserted **"There is no CREATE route, and the failure does not say so."** `cairn create` (`PUT` + `If-None-Match: *`) has shipped since devrc#1254 / `34d00d90`.

🔴 **Scope grew after a round-0 audit: it was not one file but FIVE.** See the [round-0 comment](#issuecomment-5640877644) — my first sweep was case-sensitive and I reported it clean. It was not.

## 🔴 Why it survived 8 days — the conclusion was still true

A scope's FIRST entry genuinely cannot be made through the API. But not for the reason given. The real mechanism: the index is built by **walking the store root narrowed by the caller's allowlist** (`subsystem_recall.load_store`), so a scope with no directory on the pod's disk resolves to nothing and the write is refused at the index, before any filesystem write.

Anyone checking the conclusion found it held and moved on; nothing checked the premise. **A doc claim that is right by accident reads exactly like one that is right.**

## Measured 2026-09-11, with the control that makes the reading mean something

| `cairn create` target | answer |
|---|---|
| absent + non-allowlisted scope | **rc 6** `[not-found] — not found` |
| allowlisted, present scope, ref that already exists | **rc 9** `[already-exists]` |

Two different codes, so the `not-found` is a fact about that scope rather than a probe wired to nothing. **Both wrote nothing.** This also re-derives the sub-claim `handoff-index-store-claims-accuracy.md` had flagged UNVERIFIED (the earlier probe omitted `--file`).

## Sites corrected

`claude/skills/cairn/SKILL.md` · `claudedocs/plan-cairn-phase1-cutover.md` · `claudedocs/handoff-cairn-phase3.md` ×2 · `scripts/tests/test_index_append_protocol.py` ×2 · `scripts/tests/test_cairn_cutover.py` · plus retraction markers moved inside the scanner's window in two handoff docs.

Each keeps the still-TRUE conclusion. ⚠ `test_cairn_cutover.py` is the sharp one — `cairn-cutover.py` had **already** retracted that reason, so the code and its own test disagreed. And it **straddles a newline**, so no line-based grep could see it.

## `reference/operator-surface.md`

Gains the two ways an **operator** can tell ABSENT from REFUSED, which no client can: the count gap (store-wide `entry-files=` minus your `X-Store-Entries`; **gap 0 ⇒ nothing is hidden from you**) and the pod's own `ls /data`. Measured **237 / 237 / 237**, so the token's allowlist was refusing nothing.

## Two guards, because they catch different failures

1. **`_RETRACTED_BOUNDARY` needle** (existing repo-wide scanner — normalises case, emphasis and line wraps; marker-window exemption with a negative *and* positive control). Catches the **sentence** reappearing anywhere.
2. **`test_cairn_skill_verb_ledger.py`** — a two-way ledger of every subcommand, read off the **real argparse parser**. Catches **forward drift**: a verb the CLI gains that the skill never mentions.

## Test matrix

- **Red at `origin/main`**: `1 failed, 9 passed` — exactly `test_a_named_verb_appears_in_the_skill_body[create]`. Re-verified after the parser refactor.
- **Green at HEAD**: `10 passed`.
- **Needle mutation-tested**: reintroducing the sentence unmarked turns the scanner red, naming file and needle; tree restored clean.
- Scanner suite **16 passed** (both controls). Affected suites **297 passed**. Five doc/skill-integrity suites **221 passed** (+ this PR's 10 = 231).
- `test_doc_path_rot` caught a real defect in my own new prose (a bare relative ref where a SKILL.md needs the deployed `~/.claude/…` path).

⚠ **Scope of the claim**: change-scoped dev-host runs. `test_subsystem_store_api.py` full run is large; the scanner tests within it were run directly. Neither full tier was run; CI is the advisory signal.

⚠ **Not deployed.** The skill is a `home.file` copy, so the corrected text is not live until a `home-manager switch` / `ship.sh`.

https://claude.ai/code/session_01Ae2foszv754sLoi2RcZmGh

### DIFF
--- a/claude/skills/cairn/SKILL.md
+++ b/claude/skills/cairn/SKILL.md
@@ -73,11 +73,17 @@ directory and all.** `create_entry` runs `path.parent.mkdir(exist_ok=True)`
 `test_a_scopes_FIRST_entry_creates_the_directory` asserts **201** plus the bytes
 on disk for a scope with no directory.
 
-🔴 **THE ONLY GATE IS YOUR TOKEN'S SCOPE ALLOWLIST.** A scope outside it answers
-**404**, byte-identical to one that never existed — deliberately, so an error
-cannot enumerate the store. A scope INSIDE it with no directory yet is created.
-So the remedy for "I cannot create into scope X" is an allowlist edit, **not**
-seeding.
+🔴 **THE ONLY SCOPE-LEVEL GATE IS YOUR TOKEN'S SCOPE ALLOWLIST.** A scope outside
+it answers **404**; a scope INSIDE it with no directory yet is created. So the
+remedy for "I cannot create into scope X" is an allowlist edit, **not** seeding.
+
+🔴 **That 404 is byte-identical for FOUR different causes** — a scope outside your
+allowlist, a scope that never existed, a ref that resolves to nothing, and an
+entry the loader could not parse (`server.py`, and the same list in its README).
+Deliberately, so an error cannot enumerate the store. ⚠ Do not read a 404 as
+"widen the allowlist": that is one of four, and an earlier version of this block
+listed only two after being rewritten to widen the CREATE claim — wider on one
+axis, narrower on another.
 
 ⚠ **THIS FILE HAS NOW BEEN WRONG TWICE ABOUT THE SAME SENTENCE, IN OPPOSITE
 DIRECTIONS.** It used to say there was no create route at all (false since
--- a/claude/skills/cairn/SKILL.md
+++ b/claude/skills/cairn/SKILL.md
@@ -89,8 +89,8 @@ probe wired to nothing. ⚠ Both wrote nothing.
 
 🔴 **That 404 is byte-identical for "outside your allowlist", "never existed" and
 "ref resolves to nothing"** — deliberately, so an error cannot enumerate the
-store. `reference/operator-surface.md` carries the two ways an OPERATOR can still
-tell absent from refused; no client can.
+store. `~/.claude/skills/cairn/reference/operator-surface.md` carries the two
+ways an OPERATOR can still tell absent from refused; no client can.
 
 ## 🔴 The two different exit 4s
 
--- a/claude/skills/cairn/SKILL.md
+++ b/claude/skills/cairn/SKILL.md
@@ -80,10 +80,15 @@ remedy for "I cannot create into scope X" is an allowlist edit, **not** seeding.
 🔴 **That 404 is byte-identical for FOUR different causes** — a scope outside your
 allowlist, a scope that never existed, a ref that resolves to nothing, and an
 entry the loader could not parse (`server.py`, and the same list in its README).
-Deliberately, so an error cannot enumerate the store. ⚠ Do not read a 404 as
-"widen the allowlist": that is one of four, and an earlier version of this block
-listed only two after being rewritten to widen the CREATE claim — wider on one
-axis, narrower on another.
+Deliberately, so an error cannot enumerate the store. ⚠ That four-way ambiguity
+is about a **write 404 generally** (it reaches `append` and `If-Match` `PUT`
+through `_resolve_writable`). **On the CREATE path specifically only the
+allowlist arm can fire** — a ref resolving to nothing is the success case, a bad
+normalisation is 400, and the malformed-entry arm is not on that route — so for
+`cairn create`, "widen the allowlist" IS the right reading. ⚠ An earlier version
+of this block listed only two causes after being rewritten to widen the CREATE
+claim, and its replacement then told the reader to distrust the one reading that
+is correct here: wider on one axis, narrower on another, twice running.
 
 ⚠ **THIS FILE HAS NOW BEEN WRONG TWICE ABOUT THE SAME SENTENCE, IN OPPOSITE
 DIRECTIONS.** It used to say there was no create route at all (false since
--- a/claude/skills/cairn/SKILL.md
+++ b/claude/skills/cairn/SKILL.md
@@ -67,30 +67,37 @@ every writer; `prune-index` owns deletion, with its own confirmation gate. Load
 whichever applies rather than reconstructing their steps here.
 
 🔴 **A CREATE route EXISTS — `cairn create --scope S --ref R --file F` (`PUT`
-with `If-None-Match: *`, devrc#1254). It still cannot create a scope's FIRST
-entry, and the reason is NOT the one this file used to give.** The old text said
-there was no create verb at all; that has been false since #1254, and the
-conclusion it drew survived only by accident. The real mechanism is that the
-index is built by WALKING THE STORE ROOT narrowed by your allowlist
-(`subsystem_recall.load_store`), so a scope with no directory on the pod's disk
-resolves to nothing and the write is refused at the index, before any filesystem
-write. A new scope's first entry is therefore still an operator step — seeding —
-and until somebody takes it the record exists nowhere the pod can serve.
+with `If-None-Match: *`, devrc#1254) — and it DOES create a scope's first entry,
+directory and all.** `create_entry` runs `path.parent.mkdir(exist_ok=True)`
+(`server.py`), commented *"how the store gained EVERY scope it has"*, and
+`test_a_scopes_FIRST_entry_creates_the_directory` asserts **201** plus the bytes
+on disk for a scope with no directory.
 
-MEASURED 2026-09-11, with the control that makes the reading mean something:
+🔴 **THE ONLY GATE IS YOUR TOKEN'S SCOPE ALLOWLIST.** A scope outside it answers
+**404**, byte-identical to one that never existed — deliberately, so an error
+cannot enumerate the store. A scope INSIDE it with no directory yet is created.
+So the remedy for "I cannot create into scope X" is an allowlist edit, **not**
+seeding.
 
-| `cairn create` target | answer |
-|---|---|
-| absent + non-allowlisted scope | **rc 6** `[not-found] — not found` |
-| allowlisted, present scope, ref that already exists | **rc 9** `[already-exists]` |
+⚠ **THIS FILE HAS NOW BEEN WRONG TWICE ABOUT THE SAME SENTENCE, IN OPPOSITE
+DIRECTIONS.** It used to say there was no create route at all (false since
+#1254). The correction then said a first entry "is still an operator step —
+seeding", blaming an index walk — **also false**, and caught by a round-1 audit.
+The measurement behind it probed a scope that was absent *and* non-allowlisted,
+on a pod where those two sets are identical, so it could not tell the allowlist
+gate from the index walk and credited the wrong one. `claude/RULES.md`: *"An
+EMPTY RESULT cannot distinguish two mechanisms — go find the step that differs."*
+The discriminating control is: allowlist a scope, do NOT seed it, `cairn create`
+→ expect **201**.
 
-The two codes differ, so the `not-found` is a fact about that scope rather than a
-probe wired to nothing. ⚠ Both wrote nothing.
+⚠ The failure a caller actually sees, measured 2026-09-11: `cairn create` into a
+non-allowlisted scope is **rc 6** `[not-found]`; into an allowlisted scope at a
+ref that exists is **rc 9** `[already-exists]`. Both wrote nothing. The codes
+differ, so rc 6 is a real reading — **but it is about the ALLOWLIST**, which is
+what the earlier version of this block got wrong.
 
-🔴 **That 404 is byte-identical for "outside your allowlist", "never existed" and
-"ref resolves to nothing"** — deliberately, so an error cannot enumerate the
-store. `~/.claude/skills/cairn/reference/operator-surface.md` carries the two
-ways an OPERATOR can still tell absent from refused; no client can.
+`~/.claude/skills/cairn/reference/operator-surface.md` carries the two ways an
+OPERATOR can tell a refused scope from an absent one; no client can.
 
 ## 🔴 The two different exit 4s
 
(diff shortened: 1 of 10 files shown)


ZacxDev:
## ROUND 0 (requirements & deletion) — verdict, and the fixes it produced

`requirement questioned — R2 "leave a guard behind so the class cannot drift back silently", and its sub-requirement R3 that the guard pin verb NAMES *instead of* the refuted sentence`

`round 0 · requirements: 6 (unattributed: 2) · deletion candidates: 4`

🔴 **The round was right, and it caught a false claim of mine.** I reported this PR's sweep as clean. **It was not — my sweep was case-sensitive and I quoted its zero.**

| sweep | result |
|---|---|
| what I ran and quoted (`grep -ln 'There is no CREATE route\|no CREATE route'`) | **0 files** |
| case-insensitive | **6 files** |
| the repo's own normalising scanner (`_unmarked_retractions`) | **10 occurrences in 7 files** |

One of them **straddles a newline** inside a docstring in `test_cairn_cutover.py`, where no line-based grep can ever see it. `RULES.md`: *"Grep's answer is a claim about grep's VIEW, never the file's contents."*

### Fixed in `04ce4a20` + `993242c6`

- **The guard is now the existing instrument, not only a new one.** Added one needle (`"no create route"`) to `_RETRACTED_BOUNDARY` in `test_subsystem_store_api.py`, which already normalises case, emphasis and line wraps, exempts a quote only when a retraction marker sits within 60 chars, and carries a negative **and** a positive control. Its needle list had already absorbed a non-boundary claim for exactly this reason.
- **Four live false assertions corrected** — `plan-cairn-phase1-cutover.md` (its own closing condition was met by #1254 on 2026-09-03), `handoff-cairn-phase3.md` ×2 (rank 24 + the Goal note), `test_index_append_protocol.py`, `test_cairn_cutover.py`. Each keeps the still-TRUE conclusion: a new **scope's** first entry remains uncreatable via the API, because the index is built by walking the store root, so an absent scope resolves to nothing.
- ⚠ `test_cairn_cutover.py` was the sharp one: `cairn-cutover.py` had **already** retracted that reason, so the code and its own test disagreed.
- **R3's justification was false and is corrected in place.** I argued a sentence guard "would trip on this change's own retraction". The scanner exempts a quote carrying a marker — that is what the marker window is *for*. Both guards now run: the ledger catches forward drift on a new verb, the needle catches the sentence reappearing anywhere.
- **D1 — consolidated.** The ledger now reads the **real argparse parser** (`build_parser()._actions`), the way `test_cairn_split.py` already did, instead of an AST scan of `add_parser` literals. Same 9 verbs; strictly stronger.
- **D2 — disclosed.** 9 of the file's 10 tests were already green at base; only `[create]` was red. The rest are **invariant guards**, not regression coverage, and now say so.

### 🔴 Correction to this PR's description (posting rather than silently editing)

The body said **"Related skill/doc-integrity suites: 231 passed"** and then listed five suites. **221** is the figure for those five; the remaining 10 are this PR's own new tests. The total was right, the label was 10 tests wider than the list.

### Verification

- Needle **mutation-tested**: reintroducing the sentence unmarked in a tracked file turns `test_NO_TRACKED_FILE_ASSERTS_…` **red**, naming the file and the needle. Tree restored clean afterwards.
- Scanner suite **16 passed** including both its controls.
- Affected suites (`test_cairn_skill_verb_ledger`, `test_index_append_protocol`, `test_cairn_cutover`, `test_cairn_split`, `test_doc_path_rot`): **297 passed**.
- Red-at-base matrix **re-verified after the parser refactor**: red at `origin/main` on exactly `[create]`, green at HEAD.

### Not fixed — open, not absent

**D3:** nothing checks that a `DELEGATED` ledger reason is *true*. Measured churn of the guarded surface is 6 verb-set-changing commits in 16 days, so the cheapest way past the ledger is to type `DELEGATED` and a sentence. That is the maintenance cost the premise asked about, and it is where it leaks.

**Round 0 trial ledger:** `ran: 1 · changed the outcome: 1` — it changed the outcome materially (a false clean-sweep claim, four live false assertions, and a duplicated predicate).

ZacxDev:
## ROUND 1 — findings acted on, and what was NOT

Round 1's verdict was **needs rework**, on a 🔴: this PR had replaced one false premise with another. That is fixed in `7746f9b7`, along with two of the four lesser findings.

🔴 **The 🔴 was correct and the correction is confirmed in-tree.** `cairn create` **does** create a scope's first entry — `create_entry` runs `path.parent.mkdir(exist_ok=True)`, the handler's `except rc.UnknownScopeError: existing = None` is commented *"THE ONE PLACE THIS IS NOT AN ERROR"*, and `test_a_scopes_FIRST_entry_creates_the_directory` asserts **201** for a scope with no directory — a test that shipped in `34d00d90`/#1254, the very commit this PR cites. The only gate is the token's scope allowlist.

The probe that produced the wrong mechanism used a scope that was absent **and** non-allowlisted, on a pod whose allowlist enumerates exactly the scopes it holds — the two sets coincide, so it could not separate them. The 404 comes from the allowlist check, which returns before the index is loaded.

**Operational consequence:** seeding is NOT required to bring a new scope into existence. `reference/operator-surface.md` had prescribed "both, then replace the pod" on the strength of the false premise; corrected to allowlist → replace pod → `cairn create`.

### Not addressed, deliberately — open, not absent

- **F4** — the scanner's assertion text and test name still say "two-entry boundary" while covering three unrelated claim families. Pre-existing; widening the message is a separate change to a shared guard.
- **F5** — the `NAMED` half is a substring match, and the file header calls the guard structural. The two-way ledger halves genuinely are; the `NAMED` half is not.
- **F6** — `_declared_verbs()` execs `scripts/cairn`, whose unconditional `sys.path.insert(0, …/scripts/lib)` mutates the pytest process. No shadowing observed across 1,000+ tests.
- **F7** — merged-tree behaviour against #1508 / #1516 is still unrun.

```audit-claims round=1 audited=993242c6..7746f9b7
1. The "index is built by walking the store root" mechanism is retracted at all 7 sites and replaced with the token scope allowlist as the only gate.
2. claude/skills/cairn/SKILL.md now states that `cairn create` creates a scope's first entry, directory included, and records that this file has now been wrong twice in opposite directions.
3. reference/operator-surface.md's prescribed order is corrected: seeding is not required; allowlist edit -> replace pod -> `cairn create`.
4. The `_RETRACTED_BOUNDARY` comment instructing future readers NOT to correct the sentence is deleted, and its falsity recorded in place.
5. The "four live sites in four tracked files" count is replaced with a scanner-measured figure at base 8b2b960b: 10 occurrences across 8 files, 6 live assertions in 4 files besides SKILL.md.
6. The new needle's false-positive range (a short generic phrase that trips on a TRUE statement about another API) is documented at the needle.
7. NOT addressed this round: F4 (assertion text names one claim family of three), F5 (NAMED half is a word match), F6 (exec mutates sys.path), F7 (merged tree vs #1508/#1516 unrun).
```

ZacxDev:
## ROUND 2 — delta `993242c6..7746f9b7`, findings acted on

Round 2 **independently verified the core correction is true** — the only scope-level gate on create is the token allowlist, and `test_a_scopes_FIRST_entry_creates_the_directory` is itself the discriminating control round 1 said was never run. It also found the fix round had reproduced its own failure mode twice.

🔴 **F1 was the important one and it is mine twice over.** The round-1 sweep searched by *mechanism* ("walking the store root") and found 7 sites. An **8th** spelled only the *conclusion* — *"a new SCOPE's first entry remains an operator step, for a different reason"* — in `handoff-cairn-phase3.md:87`, **a sentence this PR wrote**, 860 lines above the rank 24 the same commit corrected to say the opposite.

**And the structural half is worse than the instance:** the second retraction had **no needle**, so it was enforced by nothing — precisely the state the first claim was in while it drifted for eight days, in a PR whose whole thesis is that hand sweeps return false zeros.

Added `"remains an operator step"` / `"is still an operator step"`. ⚠ Deliberately **not** needling `"walking the store root"` — that phrase truthfully describes `snapshot_freshness`, so it would fire on correct writing.

**Both new needles mutation-tested:** each mutant dies naming its own needle; tree restored clean.

| finding | disposition |
|---|---|
| 🔴 F1 8th site + missing needle | fixed, needles added and mutation-tested |
| 🟡 F2 404 narrowed from four causes to two | fixed — `server.py` enumerates **four**; wider on one axis, narrower on another |
| 🟡 F3 `operator-surface.md` self-contradiction 12 lines apart | fixed, both older claims retracted in place |
| 🟡 F4 count wrong again (6-in-4 vs **8 live in 6 files**) | fixed, re-derived with the repo's own scanner + positive control |
| 🟢 F5 cutover docstring justification | now says **nothing justifies this**, per audit-pr's own "a guard that has lost its reason should say so" |
| 🟢 F6 verb-ledger antecedent false after the correction | restated direction-neutrally |

**F4 re-derivation** (loading `_normalise_for_scan` / `_RETRACTION_MARKERS` / `_MARKER_WINDOW` against an extracted `8b2b960b` tree, positive control = 1): **10 occurrences / 8 files · 9 live+unmarked / 7 files · 8 live / 6 files besides SKILL.md.** The auditor's figures reproduce exactly. That bullet's count has now been wrong three times, and it now says so.

Verified: affected suites **297 passed**; `TestByteIdentityVerifier` **35 passed**.

```audit-claims round=2 audited=7746f9b7..73a94e65
1. The 8th site at handoff-cairn-phase3.md:87 no longer asserts that a new scope's first entry remains an operator step; it states the allowlist is the only gate.
2. Two needles ("remains an operator step", "is still an operator step") now pin the SECOND retracted claim; both were mutation-tested and each kills its mutant naming its own needle.
3. "walking the store root" was deliberately NOT needled, because it truthfully describes snapshot_freshness; the reason is recorded at the needle.
4. SKILL.md again states all FOUR causes of the 404, restoring what the create-claim rewrite had narrowed to two.
5. reference/operator-surface.md no longer contradicts itself: the "remedies are opposite" framing and "widening the allowlist would change nothing" are both retracted in place.
6. The site count is re-derived with the repo's own scanner plus a positive control: 10 occurrences / 8 files, 9 live in 7, 8 live in 6 besides SKILL.md.
7. test_cairn_cutover.py's docstring now states that nothing justifies the asymmetry as a design choice, rather than supplying a fourth rationale.
8. The verb-ledger docstring rule is restated direction-neutrally so its antecedent is no longer false.
```

ZacxDev:
## ROUND 3 — delta `7746f9b7..73a94e65` — no 🔴, six 🟡/🟢, all acted on

Round 3 independently re-derived four claims and all four held — including the site count, **correct for the first time in four attempts**. It also verified the four causes against `server.py` and confirmed both needles were reachable. No deploy-blocking finding.

Two findings were worth fixing regardless of severity:

🟡 **F1 — "All sites are corrected" was false when I wrote it.** Two more live, present-tense sites in spellings neither round-2 needle could see — one **115 lines below that very claim, in the same file**. Both point a reader at `seed.sh`, which is the *destructive* remedy (rank 23 of that doc records it reverting pod content in 10 of 10 divergent entries).

🔴 **Four sweeps, four spellings, every one by STRING**: mechanism → one conclusion phrase → two more. The claim has no canonical wording, so a needle cannot close the class. That is now recorded at the needle and in the handoff, with the instruction to sweep by *meaning* and never to write "all sites are corrected" again without showing the sweep.

🟡 **F2 — my round-2 needles forbade TRUE writing.** `"remains an operator step"` carries no subject. **Measured: 9 true occurrences of "operator step" across 7 unrelated files** (signal, nix disk cleanup, tmux-webapp, browser-bridge, cairn-oss-multi-instance), each one word from turning the gate red — the audit turned it red by changing "is" to "remains" in a doc about `sudo`. That is the *exact objection I had used two lines above to reject a third needle*, applied harder to what I added. `RULES.md`: a permanently-red gate is worse than no gate.

**Paired verification after narrowing:** the audit's exact false-positive sentence now **passes**, and **4/4 mutants still die**.

| finding | disposition |
|---|---|
| 🟡 F1 9th + 10th live site; false "all corrected" | fixed; 3 needles added; class limitation recorded |
| 🟡 F2 needles fired on true writing | fixed — subject-bearing; false-positive control + mutants both verified |
| 🟡 F3 docstring lede survived its own retraction; I quoted ⅓ of the code's docstring | fixed — the code's FIRST paragraph *does* assert a rationale |
| 🟡 F4 `operator-surface.md` overshoot | fixed — two-row table; the old text would have caused a pointless prod pod restart |
| 🟢 F5 unqualified "only gate" in the handoff | qualified to scope-level |
| 🟢 F6 ⚠ told readers to distrust the one correct reading on the create path | scoped to write-404s generally |

Verified: `TestByteIdentityVerifier` **35 passed**, affected suites **297 passed**.

```audit-claims round=3 audited=73a94e65..cd5b23a2
1. handoff-cairn-phase3.md:956 and handoff-index-store-claims-accuracy.md:544 no longer assert in present tense that a first record can only reach the pod via seed.sh; both are retracted in place with a do-not-run-seed.sh warning.
2. The false "All sites are corrected" claim is replaced by a statement that four string sweeps found four different spellings and that a needle cannot close this class.
3. The two round-2 needles are narrowed to carry a subject ("first entry remains/is still an operator step") because the unsubjected versions fired on 9 true occurrences across 7 unrelated files.
4. Three further needles were added for the spellings F1 found: "structurally cannot accept a new entry", "only path that ever created one", "can only reach the pod through an operator".
5. The narrowed needle set was verified as a PAIR: the audit's false-positive sentence passes, and 4/4 mutants are killed.
6. test_cairn_cutover.py's lede no longer promises a reason its body denies, and its characterisation of cairn-cutover.py is corrected to acknowledge that the code's FIRST paragraph does assert a rationale.
7. reference/operator-surface.md now distinguishes REFUSED (edit secret, replace pod, create) from ABSENT (cairn create alone, no restart).
8. handoff-cairn-phase3.md:88 is qualified to "only SCOPE-LEVEL gate"; SKILL.md's four-causes warning is scoped to write-404s generally rather than the create path.
```

PR #1508, opened 2026-09-11
chore(cairn): consolidate onto the pinned client — delete the five forked reader modules

ZacxDev:
Rank 3 slice 3 of the `cairn-oss-multi-instance` arc. devrc carried its own
copies of five reader modules beside the `ZacxDev/cairn` originals and the two
had **forked** — measured 2026-09-07/08, widening on its own while the item sat
open. **Operator decision 2026-09-08: consolidate onto the pin.** devrc deletes
its copies and takes them from the locked `cairn` flake input; the writer takes
its shared vocabulary from the pinned `entry_shape` instead of its own copies.

## 🔴 AUDIT ROUND 1 FOUND TWO DEPLOY-BLOCKERS. Both are fixed; read this first.

They share one root cause, and it is the reusable part: **every environment
claim in the first version of this PR was measured from a shell that has `cairn`
on PATH. The three environments that decide whether this repo's SCHEDULED work
runs do not.**

**1 — three live systemd user timers could not have started.**
`analyze-service-index-backup`, `handoff-index-sync` and `present-regen` each set
`Environment=PATH=${lib.makeBinPath [...]}` — a CLOSED list. Measured live with
`systemctl --user show <unit> -p Environment`: **none contains a cairn path**
(positive control on the same read — `coreutils` appears in all three). Their
programs now import the pinned reader at module scope, and `cairn_pin.ensure()`
raises by design, so they do not degrade — they fail to start. Matrix, same
closed env both arms: base `IMPORT OK` → head `CairnPinUnresolved`, for
`backup.py` and `handoff_doc.py` alike.

🔴 **And it breaks on `git pull`, not on a switch.** Each unit `ExecStart`s the
WORKING-TREE copy, so the first version of this body put the risk in the wrong
place when it scoped it behind "no `home-manager switch` was performed".

**Fix:** `CAIRN_LIB=${cairnPackage}/libexec/cairn/lib` on all three. Widening
PATH is *not* an alternative for the backup unit — `%h/.local/bin/cairn` is under
$HOME and that unit runs `ProtectHome=tmpfs`. Measured here with `systemd-run
--user`, both arms: with the directive a `[ -e … ]` probe prints **ABSENT**,
without it **PRESENT**.

🔴 **And the guard that let it ship green is fixed too.**
`test_analyze_service_index_backup.py::_unit_shaped_env` — the only probe
claiming to model this environment — had a docstring saying *"`env -i` plus
exactly what nix/home.nix sets … NOT `dict(os.environ)`"* over a body that passed
`os.environ["PATH"]` straight through. It now builds a closed PATH from the
unit's own tool list, **with a positive control asserting `cairn` is not
reachable from it**, so the modelling cannot go vacuous the same way twice.

**2 — my own new code was unreachable on every host.** `m_index_store` still
guarded on `(lib / "subsystem_recall.py").is_file()`, a file this PR deletes: the
row was permanently UNMEASURED, the `cairn_pin.ensure()` behind it was dead, and
its reason text pointed a reader at a deliberately-deleted file. Fixed **with**
#1 rather than after it — removing the precondition exposes `present-regen`'s
closed PATH, so alone it would have split the row by tier.

## 🔴 ROUND 3 (blind delta): behaviour merge-ready; three sentences and one structure

The auditor could not break a single claim this delta makes about code — every
count, size, date, path and cross-reference re-derived true, including a
self-caught inflation and a self-reported RED-at-base it reproduced
independently. All four findings were prose in files already touched, and one of
them named a failure mode that **cannot occur**.

* **🟡1 — a comment describing the wrong failure, on the arm that names this
  guard.** Re-measured both arms myself, caches purged: a real `git init`ed tree
  gives `Failed: DID NOT RAISE AssertionError`; a `.git`-less `cp -a` copy gives
  `entry_shape.GitError`. My comment claimed the first dies on the guard's own
  `AssertionError: git was invoked` — it **cannot**: with the patch deleted
  `entry_shape.scope_for_repo(ROOT)` returns normally, so the sentinel is
  structurally unable to arrive and its **absence** is the mechanism. Corrected
  to the measurement, not to a third description.
* **🟡2 — a sweep that reached one claim and not its sibling, in the same
  commit.** `PIN_ABSENCE_CONSEQUENCE` was added to retract the blanket "this unit
  will not start"; seventeen lines above, the ledger docstring still carried it.
  The docstring now states only what is uniform and names the map as the
  authority. Swept the repo for other survivors — the rest are unit-specific and
  true.
* **🟡3 — a RENDERED string naming a file this PR deletes.** `source=` for the
  index-store row said "read via `scripts/lib/subsystem_recall.load_store()`",
  and `render.py` prints it verbatim as the row's `from:` line. Swept **all 25**
  `source=` literals rather than patching one: 13 live repo paths, **one dead**
  (this), one deliberate glob. Fixing it closes the class.
* **🟢4 — the hoist had traded one latent shape for another, and my first fix for
  that reintroduced the first.** The outer `try` now carries **only** a
  `finally`; handlers naming `cairn_pin` live in the `else`, where the import has
  provably succeeded. Both shapes are now **exercised** by a new guard with a
  planted import failure — RED at the audited tip, GREEN at head. ⚠ That guard's
  own positive control was miswired on the first attempt (it called the real
  import, bypassing the plant) and failed with DID NOT RAISE — the control
  working on itself.

⚠ **This round is NOT zero-payload.** It changes executable structure in
`measure.py` (the `try`/`else`/`finally` restructure) and one rendered string, so
the ladder has not left the PR.

## 🔴 ROUND 2 (blind delta): no logic defect — every finding was a SENTENCE

Both round-1 blockers were re-derived independently and confirmed closed, and the
`CAIRN_LIB` ledger was verified complete against all 18 units that run a devrc
script. What round 2 found was prose, and in four of six cases **the edit that
asserted a number or a state is the edit that invalidated it**. One reached
shipped output.

* **🟡1 — shipped output, and mine from round 1.** Replacing the two dead
  `readers` entries took the row from `0 local reader(s) can speak to it` to
  **`1`** (measured at both refs), because `scripts/cairn` imports urllib — while
  the `detail` three lines below still said the store has *"no reader"*.
  `present-regen` republishes that page daily, and the comment directly above the
  list warns about this exact failure. `detail` now states what the row measures:
  the client exists (shipped 2026-08-26), and the readerless period is history
  rather than the present tense. **A guard was added**, since a human caught it
  both times — RED at the audited tip on its own message, GREEN at head.
* **🟡2** — the vacuity fix's docstring asserted the state it had just left:
  "BOTH `_git`s ARE PATCHED" eleven lines above "ONLY `entry_shape._git` …", "the
  two patches" over one, and a `MEASURED:` claim that was the pre-fix reading.
  Re-measured (mutant RED, control GREEN) and rewritten as one account — which
  now also records that the kill *reason* differs by environment.
* **🟡3** — claim 7's sweep reached 2 of 4 surfaces and skipped the one that
  ACTS: `cairn-cutover.py` P4 runs the comparator and hard-refuses on non-zero.
  Its refusal text now names the expected class at the moment it refuses, and the
  store-api README says the copy-paste recipe is expected to fail today. Swept
  all 17 files that mention the script rather than the four named.
* **🟢4** — a `grep` figure that my own comment expansion invalidated. **Dropped
  rather than refreshed**: any number written there rots on the next edit to the
  same paragraph, so it has no stable figure to quote.
* **🟢5** — `timeouts` called "devrc-ONLY" when it exists in both trees — in a PR
  titled *delete the five forked reader modules*, the sentence that would stop
  the next person noticing a sixth duplicate.
* **🟢6** — "this unit will not start" applied to all three units; false for
  `present-regen`, because my own new `except CairnPinUnresolved` degrades it to
  `Unmeasurable`. Now per-unit, and it says the weaker case is the **more**
  dangerous one.
* Sub-bar, taken: an `except` clause naming a module imported inside its own
  `try`. Hoisted.

⚠ **I swept my own new sentences too, and two were already wrong** — the new
guard's docstring claimed "the third time" where the record supports two, and two
pin-derived figures are now marked as of-the-pin so a bump reads them as dated
rather than false.

## 🔴 THE FIX ROUND RESET THE GATE, AND THE TIER CAUGHT THREE MORE — mine

An audit fix is a code change like any other, so the tiers were re-run and they
were not green. All three were introduced BY the blocker fix, and two of them are
the same two-tier split the fix was about — committed by the fix for it:

* **the literal token `Exec`+`Start` inside a comment** I added to the
  `handoff-index-sync` Environment block. Guards read these blocks as TEXT:
  `test_handoff_index.py` counts lines containing it (my comment made the count
  2) and slices the Environment block at its first occurrence (my comment
  truncated the block, hiding the handle lines from the check that they are
  DERIVED rather than listed — so a correct unit failed). Reworded, and the
  hazard is written down beside the entry.
* **`_closed_unit_path` required EVERY unit tool to resolve**, with a message
  asserting the tool "is in flake.nix's gateTools". MEASURED FALSE: `kubectl`
  appears **0** times in `flake.nix` (positive control: `pkgs.git` appears 2). So
  every probe in that file aborted in the sandbox while passing on the dev host.
  The property is CLOSEDNESS, and a missing tool can only make the modelled PATH
  *more* closed — so tools the probed program needs (`git`, `cp`) are required by
  name, anything else is a named omission, and a new assertion refuses an empty
  modelled PATH.
* **the new probe's PATH replacement had to be classified.** `_unit_shaped_env`
  used to pass `os.environ["PATH"]` through — which `launcher_scan` correctly
  does not count as a clobber, and which is precisely why the suite could not see
  that the unit's PATH has no cairn in it. Modelling the closed PATH makes it a
  real replacement, so the two-way clobber ledger fired. Classified, with the
  justification that replacing is required: prepending would leave the pinned
  client findable and every pin-absence assertion would pass having measured
  nothing.

⚠ **Reported rather than smoothed over:** the round-2 tier read
`failed=32`; this is what those 32 were, and they were found by the gating tier
rather than by me.

## New guards this round, and why each is not a paraphrase of the others

| guard | what it would have caught |
|---|---|
| `test_the_producer_CANNOT_IMPORT_under_the_units_env_without_the_pin` | behavioural pair: `backup.py` must REFUSE under the modelled closed env and RUN with `CAIRN_LIB`. Its positive arm is what stops the negative arm passing for an unrelated reason |
| `test_the_backup_unit_DECLARES_the_cairn_lib_its_program_needs` | the behavioural test models an environment, so deleting the line from `nix/home.nix` leaves it green. This reads what ships |
| `test_every_scheduled_consumer_of_the_pin_declares_CAIRN_LIB` + its shrink guard | the above covers ONE unit — the one whose test file happened to exist. This is the two-way ledger over all three |
| `test_the_index_store_row_can_actually_be_MEASURED` + its negative control | every existing `Env` in `test_present_measure.py` points at a nonexistent store, so 50 tests passed over a permanently-dead MEASURED branch |
| `test_the_pinned_readers_validate_remedy_is_a_NON_COMMAND_here` | the second, previously unstated wording regression (below) |

## 🔴 A SECOND wording regression, previously unstated

Round 1 declared one regression and shipped two. The pinned `subsystem_recall`
replaces devrc's *"check a file with `subsystem_touch.py --validate <path>`"*
with **`a writer --validate <path>`** at two sites — correct for a repo that
ships no writer, wrong here, and it is the store's most common failure path (an
entry that will not parse). devrc's real equivalent is **`cairn-validate <path>`**
— on PATH, and MEASURED to accept a file path, not just `--scope`. devrc cannot
rewrite a string printed from inside the package, so it is **pinned as observed**
(both sites, counted, with a length floor as a positive control) and documented.
**Closing condition:** the same upstream hook the `repo_path_missing_message`
regression needs, merged and the pin bumped.

## 🟡 A design gap, RECORDED and deliberately NOT fixed here

Nothing checks that the lib `cairn_pin` resolves agrees with `flake.lock`. Route
2 guarantees *client/lib coherence* — the modules always sit beside the client
that was asked — but not *lock coherence*: a host whose `~/.local/bin/cairn`
predates the current lock resolves an older lib and every guard here stays green.
**Not fixed in this round on purpose** — widening scope mid-ladder is its own
hazard, and this is a new property rather than a defect in what changed.
**Closing condition:** a guard that compares the resolved lib's provenance
against `flake.lock`'s `cairn` entry (`nix flake metadata --json`), merged; or an
explicit decision that client/lib coherence is the property devrc wants and lock
coherence is `ship.sh`'s job, written down where `cairn_pin` states its routes.

## The mechanism, and why the pinned dir is APPENDED

`scripts/lib/cairn_pin.py` is the one seam. Resolution order:

1. `$CAIRN_LIB` — an explicit override.
2. The deployed client: `shutil.which("cairn")` → `os.path.realpath` →
   `<store-path>/libexec/cairn/lib`. (`bin/cairn` is a `makeWrapper` shell
   wrapper; `realpath` is load-bearing because `~/.local/bin/cairn` is a
   home-manager symlink and `parents[1]` of the *symlink* is `~/.local`.)
3. **Refuse** — `CairnPinUnresolved`, whose message names both routes and the
   remedy. There is no silent fallback **because there is nothing to fall back
   to**: the devrc copies are deleted, and an `ImportError` naming
   `subsystem_recall` would send a reader hunting for a file removed on purpose.

A candidate is accepted on a **positive content check** (`entry_shape.py` *and*
`subsystem_resolver.py` present), never on `is_dir()` — an empty
`libexec/cairn/lib` is exactly what a half-fetched store path looks like.

**A `CAIRN_LIB` that is SET but unusable REFUSES rather than falling through to
route 2.** The hermetic checks can set that variable precisely so the leg tests
the pinned client; falling through would hand them whatever `cairn` the builder
carries — green, measuring something else. An *unset* value is the ordinary case
on a host and does fall through. (This repo's `checks.pytests` takes route 2:
the package is on `gateTools`, so the gating tier exercises the route the hosts
actually use.)

**Why APPEND, not `insert(0, …)`:** every consumer bootstraps with
`sys.path.insert(0, <repo>/scripts/lib)` and then calls `ensure()`, so a name
present on both sides resolves to devrc's copy. Exactly one name is in that
position — `timeouts` — and it is not an accident to tidy away: devrc's copy is
imported by `scripts/cairn`, `scripts/cairn-who` and `scripts/claim-work`, none
of which are cairn's to govern.

🔴 **But appending is only half the protection, and the other half is not ours.**
MEASURED against `cairn-c84c142`: the pinned `subsystem_recall` runs
`sys.path.insert(0, <its own dir>)` at import time. Control, one interpreter
each — importing `entry_shape`/`subsystem_resolver`/`host_identity`/
`cairn_doctor`/`subsystem_read_store` leaves `timeouts` resolving to devrc's copy
and `sys.path[0]` as devrc's lib; adding `import subsystem_recall` moves **both**
to the store path. So the ordering `ensure()` sets is the state *before* the
reader loads, not an invariant after it. That is why the overlap is **guarded**
rather than merely arranged: `test_cairn_pin.py` pins the set two-way *and*
requires the pinned `timeouts` to carry every public name devrc's exports.

## Deleted

`scripts/lib/{host_identity,subsystem_resolver,subsystem_recall,cairn_doctor,subsystem_read_store}.py`.

They are re-derived, code-only (`ast.unparse` after stripping docstrings; control
A = file against itself → 0 lines, control B = two different modules → 39), as:
`subsystem_resolver` **0**, `subsystem_read_store` **0**, `host_identity` **19**
(pin adds `HOST_LABEL_ENV`, a superset), `cairn_doctor` **42** (pin's
`mirror_root: Path | None` + a `NOT_OBSERVABLE` frozen-mirror check, a superset),
`subsystem_recall` **98** (pin refactors `main` into `reject_recall_flags`, and
imports the shared vocabulary from `entry_shape` where devrc's imported the same
names from `subsystem_touch`).

## The writer half

`scripts/lib/subsystem_touch.py` now imports `CairnError`/`TouchError`/
`GitError`/`StoreMissingError`/`RepoPathMissingError`/`STORE_IS_PER_HOST`/
`SHAPE_HEADINGS`/`store_host`/`store_host_line`/`derive_scope` from the pinned
`entry_shape` and deletes its own copies. Its ~25 writer errors re-base onto
`CairnError` (which `TouchError` aliases), so every `except TouchError` in the
repo keeps matching what it used to. **Class identity is the point** — the pinned
reader catches `entry_shape.StoreMissingError`, so a same-named look-alike would
sail past that `except`; `test_cairn_pin.py` asserts `is`, not name equality.

**The test injection seam MOVED, and that is a behaviour change worth stating.**
`store_host_line()` → `store_host()` → `this_host()` are now all looked up in
`entry_shape`'s globals, so a test needing byte-stable `host:` output patches
`entry_shape.this_host`. Patching `subsystem_touch.this_host` moves that module's
own direct callers and nothing else. It is still ONE seam shared by reader and
writer — it is in the module both import from, which is a better place for it.

## The knowing regression, and what I did instead of accepting it

The pinned `repo_path_missing_message` is the OSS-sanitised one. Against devrc's,
it drops **two** things, not one:

* the sentence naming the pre-exported repo handles (`$DEVRC`, `$HOMELAB`,
  `$DATAPACKET`, `$CIVITAI`);
* devrc's `_scope_hint`, which distinguishes **four** readings (the store HAS the
  scope / has no such scope / `NOT CHECKED: no store root given` / `NOT CHECKED:
  store root is not a directory`) against the pin's single "Did you mean
  `--scope X`?" that appears only when the directory exists.

**Remedy — a deviation from the brief's preferred spelling, stated plainly.** The
brief suggested catching `RepoPathMissingError` at devrc's CLI boundary and
appending the handles sentence. I did **not** do that, for two reasons: there are
four such boundaries (`subsystem_touch.main`, `cairn-validate`, `scripts/cairn`,
`service_recon`), so it duplicates a predicate at N sites; and the pinned message
already carries its own hint, so appending devrc's would concatenate two
different, sometimes contradictory, hint sentences. Instead `scope_for_repo` is a
**thin wrapper in one place**: it answers the non-directory case with devrc's own
message *before* delegating, and raises the **pinned** `RepoPathMissingError`
class. Nothing is lost on devrc's own CLIs, and the derivation rule still comes
from the pin (`test_cairn_pin.py` asserts `st.derive_scope is
entry_shape.derive_scope` and that the two agree on a real repo's scope).

🔴 **What IS a real regression, and it is not fixable from here** (one of TWO —
the other is the `--validate` remedy wording above; round 1 stated one and
shipped two, which is the shape the ladder punishes): `cairn recall`
— the packaged client — prints the pinned wording. devrc cannot reach inside the
package, so the "ONE spelling, TWO CLIs" property `test_repo_path_guard.py` was
written for now holds *per half*. Rather than weaken that file's assertions, the
divergence is pinned as a relationship: the reader's six literals are updated to
upstream's exact wording, a new `TestTheWriterKeepsTheFULLERRefusal` pins the
three branches devrc still distinguishes, and the old
differ-ONLY-by-prefix test becomes
`test_the_two_cli_bodies_DIVERGE_and_the_divergence_is_ENUMERATED`, which asserts
the shared diagnosis, enumerates exactly what devrc carries and upstream does
not, and **fails if upstream ever regains the handle sentence** — telling the
next reader to simplify.
**Closing condition:** a `ZacxDev/cairn` PR that lets a consumer inject an extra
remedy/hint into `repo_path_missing_message`, merged and the pin bumped — at
which point that enumeration loop goes red and the reader literals here go back
to devrc's wording.

## 🔴 A defect in the pinned client, found by re-pointing these guards

`normalize_ref("!!!")` folds to the **empty string**, and upstream's hint asks
`(store_root / "").is_dir()` — which is the store root itself, so it is always
true. The message therefore ends **`Did you mean --scope ?`**, naming no scope at
all; `--scope ""` recalls nothing. devrc's own version guards `if not name` and
says `NOT CHECKED: that value does not normalize to a scope name`.

It is pinned **as it behaves**, so the suite states the truth about the deployed
client rather than the truth we would prefer.
**Closing condition:** a `ZacxDev/cairn` PR making the hint require a non-empty
normalised name, merged and the pin bumped — at which point that literal goes red
and is updated in the same commit.

## Tests: kept, not deleted — with the reasoning

The brief said to delete devrc test files that exist **only** to test a
now-deleted module. I checked each of the four candidates and **none of them
does**:

| file | devrc-specific coverage it also carries |
|---|---|
| `test_subsystem_resolver.py` | pins the hashed `resolver-rules` block in `claude/skills/analyze-service/reference/index-store.md`, and the `nix/home.nix` skills mapping |
| `test_subsystem_read_store.py` | pins `service_recon`/`scripts/cairn` constants and the `/analyze-service` SKILL.md remedy line |
| `test_subsystem_recall.py` | pins `scripts/resume-state.sh` resolution order and `service_recon`'s `index:` block |
| `test_cairn_doctor.py` | cross-checks `scripts/cairn`'s `IDENTITY_REMEDY` against `subsystem-store-api/server.py`'s `DEFAULT_TOKEN_FILE` |

So **nothing was deleted.** They are repointed at the pinned source through one
seam, `scripts/testlib/cairn_lib.py` (`pinned("<module>")`), which also buys
something `flake.nix` explicitly laments the absence of: *"every other cairn guard
in this repo reads text and stays green while the pinned client is thoroughly
broken"*. These now execute it.

Expectations were updated to the **pinned** strings/spellings rather than
weakened:

* `test_subsystem_recall.py` — `"from entry_shape import"` (was
  `"from subsystem_touch import"`); the two mutation anchors that the pin's
  `reject_recall_flags` refactor reworded (`if listing and limit is not None:`,
  `if page is not None and (ref is not None or limit is not None):`) — the
  harness's anchor-must-occur-exactly-once assert is what surfaced them, and it
  was **not** relaxed to a substring.
* `test_subsystem_recall.py` / `test_subsystem_touch.py` — the host fixture
  patches `entry_shape.this_host`.
* `test_subsystem_touch.py::test_kills_worktree_stable_scope` — mutates the
  pinned `entry_shape` through a new `_load_pinned_mutant` helper that keeps the
  same anchor-uniqueness rule (a 0-occurrence anchor is a FAILURE, so a pin bump
  that rewords the guard cannot silently produce a no-op mutant).
* `test_cairn_doctor.py::test_the_doctor_subcommand_is_TRACKED_and_the_module_ships`
  — now asserts `scripts/lib/cairn_pin.py` and `scripts/cairn` are tracked;
  `cairn_doctor.py` arrives through `flake.lock`, not through `git add`.
* `test_store_root_ledger.py` — `ROUTED`/`SITED` named two files this repo no
  longer tracks. Removed **with a note** recording that a row leaving that ledger
  is normally the regression it exists to catch, and that every devrc surface
  which could reintroduce the frozen-mirror defect is still ledgered. The
  `ROUTED ∩ SITED` literal drops 2 → 1; the enumerator's positive control now
  names a file that exists.
* `test_analyze_service_index_backup.py` — the transplant fixture copied
  `lib/host_identity.py`, so both halves of the pointer-ledger pair died with
  `FileNotFoundError` **in the fixture**: the pair had stopped being a
  measurement rather than failing its claim. It transplants `lib/cairn_pin.py`
  now, which resolves the real module itself.
* `test_subsystem_resolver.py::TestCommandDocIsPinned` — the doc's "EXECUTABLE
  authority" bullet named a `scripts/lib/` path this branch deletes, so the
  substring pin held an unresolvable spelling and the doc-rot gate would have
  called it dead. Both layers (substring + region hash) updated in one commit,
  and **the hash was not pasted blind**: the block was re-read against the pinned
  code first — `normalize_ref("External DNS")` → `external-dns`,
  `normalize_ref("image_ingestion")` → `image-ingestion`,
  `split_kind("repo-cos.process")` → `("repo-cos", "process")`,
  `split_kind("notes.md")` → `("notes.md", None)`, each exactly as its bullet
  describes, with a 0-line code-only diff for that module.

## The store-api pod image — scope I extended, deliberately

Not in the brief's file list, and **not optional**: the pod's `Dockerfile` COPYs
those five modules by name from `scripts/lib/`, so deleting them breaks the image
build outright. There is only one reading consistent with the operator's
decision — the Dockerfile's own header says the pod must run *the same reader the
CLI runs*, and after this change that reader is the pinned one.

* `build-push.sh` STAGES the pinned modules into a gitignored
  `scripts/subsystem-store-api/.cairn-lib/` (by name, never `cp -r`), emptying it
  first so a module dropped by a pin bump cannot linger.
* `Dockerfile` COPYs them from there and sets `ENV CAIRN_LIB=/app/cairn-lib` —
  route 1, the only one available in a container with no `cairn` on PATH.
* `subsystem_touch.py` and `git_mainline.py` are **no longer copied**: the pinned
  reader imports its vocabulary from `entry_shape`, not from the writer. That is
  a consequence, re-derived by the guard below, not a trim by eye.
* `test_the_image_copies_every_module_it_needs` now walks the closure across
  **both** trees from `server.py` itself (not a hardcoded `subsystem_recall` —
  the server also imports `cairn_pin`, which the old entrypoint could not see),
  and asserts **three** lists agree: Dockerfile COPYs, the `.dockerignore`
  allowlist, and `build-push.sh`'s stager — plus that `CAIRN_LIB` is set. Two
  lists agreeing was the original bug's shape; three is the hazard now.
* `verify-byte-identity.sh` resolves the local reader via
  `python3 scripts/lib/cairn_pin.py`.
* `test_subsystem_store_api.py`'s in-process `host:` seam moves to
  `sys.modules["entry_shape"]` — the pod image no longer carries
  `subsystem_touch` at all, so the old spelling would `KeyError` at import.

**No image was pushed.** `build-push.sh 0.0.0-cairnpin-test --no-push` ran
locally: `/data` holds 0 files, `pin resolves to /app/cairn-lib`,
`subsystem_recall imported, 3 modes`. The test image was then deleted. **The live
pod is untouched by this PR** and will pick this up only on a deliberate rebuild.

## Regression matrix

Base = `f76f3c11` (the merge base; this branch was rebased onto `origin/main`
twice while it was in flight, and the matrix was re-measured at each).
`test_cairn_pin.py`, `cairn_pin.py` and `testlib/cairn_lib.py` were copied into a
clean worktree at that ref and run there.

**RED at base, GREEN at HEAD — 6 of 17, each failing on its OWN sentence:**

| guard | what it printed at base |
|---|---|
| `test_exactly_one_module_name_is_present_on_BOTH_sides` | `overlap is ['cairn_doctor', 'host_identity', 'subsystem_read_store', 'subsystem_recall', 'subsystem_resolver', 'timeouts'], expected ['timeouts']` |
| `test_the_writer_raises_the_classes_the_PINNED_reader_catches` | `subsystem_touch.CairnError is None, not entry_shape.CairnError …` |
| `test_the_writer_takes_its_shared_vocabulary_from_the_pin_not_from_itself` | `` `subsystem_touch.CairnError` is None, not the pinned `entry_shape.CairnError` `` |
| `test_scope_derivation_still_comes_from_the_pin` | `` `subsystem_touch.derive_scope` is <function …>, not the pinned `entry_shape.derive_scope` `` |
| `test_the_five_consolidated_modules_are_not_back_in_scripts_lib` | `['host_identity', …] reappeared in scripts/lib/` |
| `test_every_consolidated_module_resolves_to_the_pin` | `` `host_identity` resolves to …/scripts/lib/host_identity.py, not to the pinned lib … `` |

Base run: **6 failed, 11 passed**. HEAD run: **17 passed**.

⚠ **Two of those six initially died on an `AttributeError` and a third on a bare
`assert a is b`** — red for the right reason, but naming no claim. A red that does
not state its own property is how a matrix stops being evidence, so the lookups
take a `None` default and the bare `is` got a message; the table above is the
re-measurement after that fix.

**The other 11 are INVARIANT / mechanism guards, labelled as such** — they pass at
base because `cairn_pin.py` was copied in and the pin is deployed there too. They
are guards on *new* code (resolution routes, the refusal, `ensure()` semantics,
the shell surface, the pinned-reader-prepends fact, the overlap compatibility
half), not regression coverage, and this PR does not count them as such.

## Mutation results — survivors included

Run under `PYTHONDONTWRITEBYTECODE=1`, each mutant restored from a `cp -a` copy
(`git status` clean between rounds), with a **no-op positive control**. The
harness asserts each anchor occurs exactly once and reports `ANCHOR-ERROR`
otherwise, so a mutation that failed to apply is never scored.

| # | mutation | verdict |
|---|---|---|
| M0 | comment-only edit (control — must survive) | **SURVIVED**, 17 passed ✅ expected |
| M1 | `_accepts` → `candidate.is_dir()` | KILLED, `DID NOT RAISE` |
| M2 | `sys.path.append` → `insert(0, …)` | KILLED, `ensure() did not APPEND` |
| M3 | refusal message drops the remedy | KILLED, `the remedy is not named` |
| M4 | set-but-invalid `CAIRN_LIB` falls through to route 2 | KILLED, `DID NOT RAISE` |
| M5 | `__main__` prints nothing | KILLED, shell-surface test |
| M6 | writer redefines `StoreMissingError` locally | KILLED, `not entry_shape.StoreMissingError` |
| M7 | writer adopts the pinned (sanitised) refusal message | KILLED, `no longer names $DEVRC` |
| M8 | `EXPECTED_OVERLAP` emptied | KILLED, `the … overlap is …` |
| M9 | `MARKER_MODULES` shrunk to one entry | **SURVIVED → FIXED → KILLED** |
| M10 | writer stops importing `derive_scope` from the pin | KILLED, names `derive_scope` |
| M11 | writer stops exposing `store_host` at all | KILLED, on the `is None` branch |

🔴 **M9 is the finding worth reading.** It survived a fully green run because the
content-check test *iterated `cairn_pin.MARKER_MODULES`* — deriving its
expectation from the implementation it tests, so the loop shrank with the tuple
and never built the case the deletion exposes. The test now spells the marker set
literally and asserts equality; M9 re-run is KILLED with that assert's own
message.

⚠ **M6 kills two guards, not one** — the class-identity guard and the vocabulary
ledger are deliberately redundant at that name. Both messages name
`StoreMissingError` specifically, so neither is standing in for the other, but
"each died for its own reason" is the weaker claim here and I am not making the
stronger one.

### Round 2 mutants (the blocker fixes)

| # | mutation | verdict |
|---|---|---|
| R2-M0 | comment-only edit (control) | **SURVIVED**, 20 passed ✅ expected |
| R2-M1 | backup unit drops `CAIRN_LIB` | KILLED, `does not set …` |
| R2-M2 | `_unit_shaped_env` goes back to the ambient PATH | KILLED — by the behavioural pair, with its own message (`backup.py STARTED with no pinned cairn client reachable … or this probe's PATH is not the unit's`). **This is the mutant that reproduces the original defect** |
| R2-M3 | `measure.py` precondition reverts to the deleted file | KILLED, on the MEASURED-branch test's own message |
| R2-M4 | `handoff-index-sync` drops `CAIRN_LIB` | KILLED |
| R2-M5 | `present-regen` drops `CAIRN_LIB` | KILLED |
| R2-M6 | drop the `entry_shape._git` patch | **SURVIVED → FIXED → KILLED** |
| R2-M7 | drop the `st._git` patch instead | **SURVIVED → the patch was REMOVED** |
| R2-M8 | delete the writer-side reachability assertion | SURVIVED — **uninformative by construction**, see below |

🔴 **R2-M6 and R2-M7 are the finding of this round.** My first fix for the vacuous
git guard patched *both* `_git`s "for strength". R2-M6 survived: a
not-a-directory input never reaches either one, so the patches were unobservable
from the assertion — the guard was still not measuring what it said. It now
drives an input that MUST call git and requires the sentinel first, which proves
the instrument is live; R2-M6 re-run is KILLED. R2-M7 then survived on the *other*
half, and that one is a fact about the code rather than the test: after the
consolidation `subsystem_touch._git` is no longer on the scope path at all (the
wrapper delegates). A patch whose removal changes nothing is not extra safety, so
it was **deleted**, not annotated.

⚠ **R2-M8 is reported and not chased.** Deleting an assertion and finding the
rest of the test green is tautological — the mutation removes the observer, not
the behaviour. It is listed so the table is the whole set I ran, not the subset
that flattered it.

Three rounds were reported `ANCHOR-ERROR` rather than scored (a wrong quoting of
the remedy string; `\n` not being interpreted by the first harness version) and
were re-run correctly — that is the harness working, and it is why the anchor
assert exists.

## What the AUTHORITATIVE tier found that the dev host did not

🔴 Worth reading, because it is the two-tier blind spot `claude/RULES.md` names.
Every suite I touched was green file-by-file in `nix develop`, and the first
`nix build .#checks.x86_64-linux.pytests` still reported
`scripts/tests (collected=13933 passed=13927 failed=6)`. **Five of the six were
mine**, and all five were ledgers or fixtures that named something this branch
moved — not logic:

* `test_subsystem_touch.py` — `_load_pinned_mutant` called `pinned()` without
  importing it (a helper added and its test not re-run after a later edit).
* `test_subsystem_touch.py` — the PR sentinel map derives its universe from
  `st.__all__`, which now exports `CairnError`. Excluded alongside `TouchError`
  for the reason `TouchError` always was: it is the BASE, not a condition, and
  the two names are now ONE class — demanding a sentinel for it would add a
  ledger entry nothing can ever exercise.
* `test_subsystem_recall.py` — the `/resume` doc pin held the literal
  `scripts/lib/subsystem_recall.py`. It pins the MODULE name now.
* `test_no_real_launchers.py` — `test_autouse_is_what_protects_a_test_that_never_asks`
  copies the REAL conftest into a control tree, and the conftest now loads
  `scripts/lib/cairn_pin.py` by path. **Both halves** died with
  `FileNotFoundError` while loading it: the control/mutant pair had stopped
  being a measurement rather than failing its claim. The file is copied in.
* `test_no_real_launchers.py` — `test_cairn_pin.py` replaces PATH and had to be
  added to the two-way clobber ledger, with the justification that replacing is
  the point (the case under test is `cairn` NOT on PATH, and the pinned client
  IS on PATH in both tiers, so prepending could never make the refusal branch
  execute — the unreachable-guard trap).

🔴 **The sixth failure is NOT mine, and it is reported rather than swept.**
`test_mention_open.py` drives a real `fzf` through a pty with two 20 s budgets
and reported `fzf never drew its prompt in 20s`. My diff cannot reach
`mention_scan` or fzf — it names `flake.nix`/`run-tests.sh` only inside a
docstring about why fzf is in `REQUIRED_TOOLS`. **Control: 242 passed at the
merge base `21f2c162` AND 242 passed at this HEAD, both at box load ~55.** That
is recorded as *unattributable*, not as a flake I proved; its timing dependency
is left alone as out of scope.

## Tier verdicts (counts, not exit codes)

Run **one at a time** (a combined invocation contends on the nix store and
produces false failures — `CLAUDE.md`), on the tree this PR pushes:

| tier | verdict (counts, not exit codes) |
|---|---|
| `pytests` | `TOTAL collected=22199 passed=22195 skipped=2 failed=2` (floor 20441) · `SCOPE: FULL (28 of 28)` · `RESULT: FAIL (exit=1)` — both failures are `main`'s, below |
| `nodetests` | `TOTAL suites=5 files=41 tests=1449 pass=1449 fail=0 skipped=0` (floor 1367) · `RESULT: PASS (exit=0)` |
| `cairn-client-runs` | built with `--rebuild` (forced re-run, not a cache hit) → exit 0 |

🔴 **`scripts/tests` — the target this PR actually changes — is
`PASS (collected=14038 passed=14038 skipped=0 floor=13026)`.**

🔴 **THE TWO FAILURES ARE PRE-EXISTING ON `main`, AND THE EVIDENCE IS A BASE RUN
WITH THIS BRANCH ABSENT.**

```
scripts/claude-hooks/tests/test_guard_core.py::test_every_kill_server_call_site_in_the_repo_is_classified
scripts/claude-hooks/tests/test_guard_core.py::test_no_tracked_shell_text_writes_a_kill_this_guard_would_deny
```

* **Base arm** — clean worktree at `550de40a`, the current `origin/main` and this
  branch's merge base, with none of my code in it: **`2 failed, 1534 passed`**,
  those exact two ids. Re-measured at **every** base this branch has been rebased
  onto (`058c90c2`, `e5f4e2f0`, `550de40a`) — same two, every time.
* **Branch arm** — the only failing target is `test_guard_core.py` with
  `failed=2`, the same two ids.
* Both assertions name `claudedocs/handoff-tmux-webapp.md`. **This PR touches no
  `claudedocs/` file and no `claude-hooks/` file**, so it cannot reach either.
* Being fixed upstream: `0f0736f7` (#1520) classified the *sibling* file and says
  in its own message that "`main` is RED and has been since that doc merged".

**Not fixed here, deliberately** — another guard's ledger, a classification
judgement over a doc this branch has nothing to do with, and someone visibly
mid-fix. A base that is already red is a fact about the tier, not this diff.

## Live surfaces exercised

From a non-interactive `bash -c` (no interactive profile, so no
`sessionVariables`; route 2 — `cairn` on PATH — is what answered):

* ✅ `cairn recall --repo <worktree> --no-sync` → `status=recalled scope=devrc`, rc 0
* ✅ devrc `scripts/cairn --help` → rc 0; `scripts/cairn recall --no-sync` → `status=recalled`
* ✅ devrc `scripts/cairn doctor --no-sync` → full report, `OK=3 PROBLEM=0 UNMEASURED=3 NOT-OBSERVABLE=1`, exit 10 (the correct verdict with `--no-sync`)
* ✅ `scripts/cairn-validate --scope devrc` → 32 entry files, `0 out-of-reach marker(s)`, rc 0
* ✅ `scripts/cairn-who --help` → rc 0
* ✅ `scripts/subsystem-audit.py --help`, `scripts/cairn-cutover.py --help` → rc 0
* ✅ `subsystem-store-api/server.py` imports clean; `handoff_doc`, `service_recon`,
  `subsystem_touch`, `analyze-service-index/{backup,restore-verify}.py`,
  `present.measure` all import clean
* ✅ the pod image builds and passes its own three controls (above)

**NOT exercised:** `cairn sync` or any write verb against the live pod;
`verify-byte-identity.sh` end-to-end (needs a port-forward to the pod); the
laptop host (a second, independent `home-manager switch`, still unverified —
that is rank 3's other carried-forward residual, untouched here).

## Also in scope, briefly

* `run-tests.sh` `REQUIRED_TOOLS` gains `cairn`, with its justification beside the
  others — a missing binary is now a named FATAL rather than five
  `ModuleNotFoundError`s or a quiet skip.
* `flake.nix` `gateTools` gains `cairn.packages.${system}.cairn`, so `nix develop`
  and **both** check tiers satisfy that precondition from one list, and the
  gating tier exercises route 2 (the route the hosts use). ⚠ `cairn` is a private
  input absent from `cache.nixos.org`, so on a cold CI store this leg may pay a
  real build — unmeasured, same caveat already recorded on
  `checks.cairn-client-runs`.
* `scripts/tests/conftest.py` calls `cairn_pin.ensure()` once for the whole
  directory, loading the module **by path** rather than prepending `scripts/lib`
  (several test modules depend on the ordering their own `sys.path.insert` sets).
* Five doc sites that named the deleted paths are fixed — `test_doc_path_rot.py`
  caught four of them and is green; the fifth (`prune-index`) was an absolute
  path the gate does not scan.
* `nix/home.nix`'s `cairn-validate` comment named `subsystem_read_store`; the
  argument there was never about the reader modules (it is about
  `subsystem_touch.py`, which no package ships), so it is corrected and says what
  moved.

## Deliberately out of scope

* `subsystem_touch`'s private `_git`/`_toplevel` are kept local rather than
  imported from `entry_shape`. They are byte-identical in behaviour, but they are
  leading-underscore names — depending on another package's non-public API is a
  coupling a pin bump can break with no API change.
* `scripts/cairn-who` and `scripts/cairn-validate` keep `mkOutOfStoreSymlink`
  deploy mode, untouched in either direction.
* `claudedocs/handoff-cairn-oss-multi-instance.md` is **not** touched — the
  handoff update for this arc is #1492.
* No image pushed, no `home-manager switch`, no pin bump.


ZacxDev:
Round 1 of the audit ladder ran blind against `acc9ee6a` and returned two 🔴 plus seven 🟡. Fixes landed at `0a331066`. Claims block for round 2:

```audit-claims round=1 audited=acc9ee6a..0a331066
1. The three scheduled systemd units (analyze-service-index-backup, handoff-index-sync, present-regen) now carry CAIRN_LIB in their Environment, so their closed PATH no longer breaks the module-level cairn_pin.ensure().
2. _unit_shaped_env in test_analyze_service_index_backup.py now models the unit's closed PATH instead of passing through os.environ["PATH"], with a positive control that cairn is unreachable from it.
3. present/measure.py no longer gates on the deleted scripts/lib/subsystem_recall.py, so the code below that precondition is reachable; a missing pin degrades with its own reading, and a test drives the MEASURED branch.
4. The three sentences in cairn_pin.py and test_cairn_pin.py asserting that the hermetic nix checks set CAIRN_LIB were corrected; behaviour unchanged, only the justification was wrong.
5. test_repo_path_guard.py's vacuous git-sabotage guard was re-pointed, found still vacuous by the author's own mutant, and the redundant st._git patch was deleted rather than annotated.
6. The second wording regression (the pinned reader's "a writer --validate <path>" replacing devrc's real command) is now pinned by a test and listed in the PR body's regression list.
7. verify-byte-identity.sh and operator-surface.md now record that the local-vs-pod comparison is expected to mismatch until an image built from this tree is pushed.
8. measure.py's m_store_api_clients readers list no longer names the two deleted modules, so the page stops asserting two devrc readers are missing files.
9. The resolver-rules marker in claude/skills/analyze-service/reference/index-store.md no longer names scripts/lib/subsystem_resolver.py as the implementing code.
10. Three nits taken: the inert ensure() in scripts/cairn was deleted rather than re-justified; test_cairn_cli.py's stale docstring and tautological assert were replaced; subsystem_touch.py's "several call sites" comment was corrected to the measured zero.
11. Three defects the authoritative tier found in the blocker fix itself were fixed: an Exec+Start token in a new comment broke two text-scanning guards, _closed_unit_path required kubectl (0x in flake.nix), and the probe's new PATH replacement was classified in the clobber ledger.
12. 🟡6 (nothing checks the resolved lib agrees with flake.lock) is RECORDED with a closing condition, deliberately NOT fixed in this round.
```

Not fixed and not ours: `tekton/devrc-pytests` is red on `test_guard_core.py` for two assertions that both name `claudedocs/handoff-tmux-webapp.md`. Measured red at the merge base with this branch absent (`2 failed, 1534 passed`), and this PR's 47-file diff touches no `claudedocs/` or `claude-hooks/` file. #1520 fixed the sibling doc and #1525 states the other was deliberately left on `main`.

ZacxDev:
Round 2 (blind delta) returned three 🟡 and three 🟢 — no logic defect; every finding was prose or a rendered string. Fixes landed at `15b17ae3`.

🔴 **Range note for round 3, because the literal one is wrong again.** The branch was rebased, so `0a331066` is **not** an ancestor of the head. Its rebase counterpart is **`e992dba4`** (verified: ancestor of HEAD; `0a331066` is not). Round 3's true delta is `e992dba4..15b17ae3` — 9 files, +261/−49. The uncorrected range attributes main's movement to this round.

```audit-claims round=2 audited=e992dba4..15b17ae3
1. present/measure.py's index-store detail paragraph now states what the row actually measures (a client exists, shipped 2026-08-26) instead of asserting the subsystem has no reader; the readerless period is kept as history, not present tense.
2. The justification comment beside the readers list now explains scripts/cairn-validate as well as scripts/cairn, naming it the discriminating case (0 urllib refs) without which the predicate cannot be shown to read anything.
3. A new guard pins count-vs-rows with a positive control plus prose/count coherence; it is RED at 0a331066 on its own message and GREEN at head.
4. test_repo_path_guard.py's docstring was rewritten as one account instead of two contradictory paragraphs; the "MEASURED: deleting the entry_shape patch leaves this test green" claim was re-measured (mutant RED, control GREEN) and corrected, and it now records that the kill reason differs by environment.
5. The expected-mismatch note for verify-byte-identity.sh was swept across all 17 files naming the script, not the 4 reported: the four that act or direct now carry it, including cairn-cutover.py P4's refusal text and the store-api README's new "A FIFTH difference" section.
6. subsystem_touch.py's pre-emptive grep figure was fixed BY DELETION rather than refreshed, on the grounds that any number there rots on the next edit to the paragraph; the code-only measurement stands.
7. test_cairn_cli.py no longer calls timeouts a devrc-only module, and records the measured reason there is no live shadowing risk (0 pinned modules import it; control: 2 import subsystem_resolver).
8. The "will not start" consequence in test_cairn_pin.py and nix/home.nix is now stated per-unit, and says the present-regen case is the more dangerous one because a dead timer is noticed and a daily page with one quietly unmeasured row is not.
9. measure.py's cairn_pin import was hoisted out of the try, so the except clause can no longer raise UnboundLocalError past the following except Exception.
10. Two self-inflicted defects were caught and corrected before commit: the new guard's docstring claimed "the third time on this row" where the record supports two, and two pin-derived figures are now marked as of the pin current when written.
```

`tekton/devrc-pytests` remains red on `main`'s two `test_guard_core.py` failures, re-measured at the *current* base `e5f4e2f0` with this branch absent (`2 failed, 1534 passed`). `scripts/tests` — the target this PR changes — is `collected=14037 passed=14037 skipped=0`.

ZacxDev:
Round 3 (blind, scoped to SENTENCES) returned three 🟡 and one 🟢 — **no behavioural defect; it could not break a single claim the delta made about code.** Fixes landed at `918e62e5`.

🔴 **Range for round 4, corrected — and the correction is now a per-round ritual.** The branch was rebased again after the round-3 fixes were measured, so the audited tip `15b17ae3` is **NOT** an ancestor of the head (`0a331066` is not either). Its counterpart is **`c0a55a32`** — same subject, verified ancestor. The true delta is `c0a55a32..918e62e5`, 4 files / +159/−36; the literal `15b17ae3..918e62e5` wrongly pulls in `claude/skills/activity/SKILL.md`, which is main's movement.

```audit-claims round=3 audited=c0a55a32..918e62e5
1. test_repo_path_guard.py's mutation-reading comment is corrected to the measurement: in a real git tree deleting the patch gives DID NOT RAISE (the sentinel is structurally unable to arrive, which IS the mechanism), and only the .git-less copy gives entry_shape.GitError.
2. test_cairn_pin.py's ledger docstring no longer carries the blanket "they fail to start" that its own sibling retraction had already disproven for present-regen; it states only what is uniform and names PIN_ABSENCE_CONSEQUENCE as the authority, recording that the blanket sentence survived its own retraction.
3. The dead rendered path in present/measure.py's index-store source string is fixed class-wide: all 25 source= literals were swept (13 live repo paths, 1 dead — this one, 1 deliberate glob false positive), independently reaching the same single real instance.
4. The cairn_pin import hoist is closed properly rather than traded: the outer try now carries only a finally and the handlers naming cairn_pin live in the else, so an import failure yields Unmeasurable rather than UnboundLocalError and sys.path is fully restored — RED at the audited tip (raw ImportError escapes), GREEN at head.
5. The "beside this file" wording is corrected — cairn_pin.py is in scripts/lib/, measure.py in scripts/present/.
6. Two self-inflicted defects were caught during this round and are reported rather than hidden: the FIRST fix for finding 4 nested the blocks but left the handlers outside, reintroducing the shape it believed it had closed; and the new guard's own positive control was miswired (it called the real import, bypassing the plant, and reported DID NOT RAISE).
```

`tekton/devrc-pytests` remains red on `main`'s two `test_guard_core.py` failures — now re-measured at **all three** bases this branch has been rebased onto (`058c90c2`, `e5f4e2f0`, `550de40a`): `2 failed, 1534 passed`, same two ids, branch absent every time. `scripts/tests` is `14038 passed / 0 failed`.

**Ledger, round 3: payload +57/−27 · tests +102/−9. NOT zero-payload** — `measure.py` carries a real `try`/`else`/`finally` restructure (41 code-only lines), so the attribution gate has not fired and the ladder has not left the PR.

PR #1509, opened 2026-09-11
feat(mention-open): order the picker by whether a repo could plausibly have the clicked #N

ZacxDev:
The mention picker offers ~392 `owner/repo` rows for a clicked `#N`, in one fixed
alphabetical order **whatever the number is**. This makes the order a function of the
number.

## Why — measured

Over a 120-repo sample of the operator's universe, by highest issue/PR number:

| max ref | repos | share |
|---|---|---|
| **0 (none at all)** | **54** | **45%** |
| 1–9 | 40 | 33% |
| 10–99 | 19 | 16% |
| 100–999 | 5 | 4% |
| 1000+ | 2 | 2% |

So `#1291` is plausible for ~2 repositories and `#12` for ~66 — and **nearly half the
list can satisfy no numeric click at all**, which is the bigger and cheaper half of the
win. 92% of detected mentions are bare `#N`, which is exactly this population.

## The design

**Tier A — plausibility (hard).** `regen-known-repos.py` gains a third output,
`known_ranges.json`: `{"owner/repo": <highest issue-or-PR number>}`, built from batched
GraphQL on the same daily timer. At click time rows sort:

`PLAUSIBLE` (`max_ref >= N`) → `BELOW` (`0 < max_ref < N`) → `UNKNOWN` (no entry) →
`IMPOSSIBLE` (`max_ref == 0`), and within PLAUSIBLE by `max_ref - N` ascending.

**Tier B — learned (soft).** A successful pick appends `{"t","repo","n"}` to
`picks.jsonl`. A recency-weighted (30-day half-life), proximity-warmed score orders rows
**within** a Tier A class and can never cross one.

## Where the right repo actually lands

Measured end-to-end on a **390-row** corpus with the distribution above (synthetic names;
this is the same construction the suite asserts against, so the figures and the guards
cannot drift apart):

| clicked | wanted repo's rank | alphabetical baseline | |
|---|---|---|---|
| `#1291` (high) | **1 of 390** | 196 of 390 | only 4 rows are plausible at all |
| `#12` (low) | — | — | the **159** zero-ref rows are *exactly* the tail, so all **231** rows that can answer come first |

Nothing is dropped in either case — asserted as a multiset equality, not a length.

## Two measurements that changed the plan

**1. `--tiebreak=end,index` is a NO-OP, so `PICKER_SH` is UNCHANGED.**
fzf appends `index` implicitly, so `end` already *is* `end,index`. Re-measured on the
merged tree with **typed queries only** (an empty query proves nothing — fzf preserves
input order when there is nothing to score): **560 typed trials, `end` vs `end,index`
differed in 0**, while the positive control `--tiebreak=length` differed in **518**. A
discriminating control on rows built to *tie* under `end` returns input order, and
reversing the input reverses the output.

`PICKER_SH` is pinned by a whole-string test, a metacharacter ban and a redirection
ledger, so changing a disclosure-relevant constant for zero behaviour change would be
pure churn. The property is pinned instead — on a **typed** query — by
`test_adding_index_to_the_TIEBREAK_would_change_NOTHING_on_a_TYPED_query`, which carries
its own positive control so a zero cannot mean a blind comparison.

**2. `gh api graphql` exits 1 when ANY alias errors and still returns `data` for the
rest.** A 21-alias batch containing one deleted repo exited **1**, printed
`errors: [{type: NOT_FOUND, path: ["r20"]}]`, and carried all **20** other answers.
Branching on the exit code would discard a whole batch on one bad row, so
`read_api_ranges` parses stdout regardless of it. (`claude/RULES.md`'s
parsing-tool-output rule, in a new shape.)

Batch size measured flat: 25/50/75/100 aliases all answered in full in ONE request at
**2.15 / 2.14 / 2.27 / 2.28s**, so ~400 repos is ~8 requests ≈ 20s — well inside the
unit's `TimeoutStartSec=300`, which did not have to move.

## How each invariant was preserved, and how it was verified

| # | Invariant | How | Verified by |
|---|---|---|---|
| 1 | **Order only, never auto-open** | the ordering never touches `offered_universe` | `test_a_PLAUSIBLE_universe_row_that_is_the_ONLY_one_is_still_NOT_auto_opened` — one repo, PLAUSIBLE, no clawgate sibling → still a picker |
| 2 | **Clawgate row stays first** | ordering applies to `universe_repos` only; the arms append | `test_main_ORDERS_the_universe_rows_and_the_CLAWGATE_row_STAYS_FIRST` (through `main()`, not the sorter) |
| 3 | **No click-time network** | two local reads; no new spawn | the existing two-way `RESOLUTION_PATH_COMMANDS` ledger, still green and still two-way |
| 4 | **Disclosure** | 0600, outside every checkout; guard extended | two new detectors + a wider file filter, each with a planted-tree control — see below |
| 5 | **Staleness is a deadman** | stale ⇒ order nothing, and say so in the header | `test_a_STALE_range_table_is_IGNORED_rather_than_TRUSTED`, `test_main_says_the_rows_are_UNORDERED_when_the_table_is_STALE` |
| 6 | **Cold start** | key is `(class, -score, distance)` with **no name**, and `sorted` is stable | `test_COLD_START_leaves_the_incoming_order_EXACTLY_as_it_was`, on a deliberately unsorted input |

### Invariant 4 in detail — the guard was blind in two new ways

* `known_ranges.json` names repositories in its **KEYS** and maps them to integers. The
  existing mapping detector asks whether a *value* is a repo name, so it scores this file
  **0** — the exact mirror image of the shape it was built for.
* `picks.jsonl` is **JSON-LINES**: `json.loads` of the whole file raises, so every
  detector that starts by parsing returns 0 — **and `.jsonl` was not in the sweep's file
  filter at all**, so no detector would ever have been handed its text.

Both are now detected, both on the same sweep, both proved able to fire on a planted tree
(`test_the_incident_guard_CAN_GO_RED_on_the_TWO_NEWER_shapes`,
`test_the_sweep_actually_LOOKS_at_jsonl_files` — the latter is a separate claim, because a
working detector the loop never calls is still no guard).

`picks.jsonl` is also the **first file this handler writes**, so the autouse redirect grew
a fourth arm plus a two-way ledger (`HOST_STATE_CONSTANTS`): without it, a test reaching a
selection would have appended the operator's *real* pick log. Every previous hole in that
fixture was a test *reading* the operator's data.

## Verified against the real host (writing nothing)

`regen-known-repos.py --print` on the workbench: **393 repos in the picker universe → 8
GraphQL requests**, exactly the prediction from the batch measurement. The host has
`known_repos.json` and `known_universe.json` (0600, parent 0700, refreshed today) and no
`known_ranges.json` yet — so after merge + `ship.sh` the picker shows the quiet
cold-start clause until the timer next fires (≤ ~17h), then orders.

## Defects found while building this, by my own instruments

| Found by | Defect |
|---|---|
| the cold-start test | `order_universe` **alphabetised** instead of refining — a trailing `name.lower()` in the sort key, under a docstring claiming the opposite |
| a `main()` end-to-end test | the header clause was keyed on `offered_universe`, which the **bare-`#N` arm never sets** — the whole universe was ordered and the header said nothing, on the commonest path |
| a latency probe | the **fast path paid for the ordering**: `may_offer_universe and num` is true for `owner/repo#N`, which carries a number and bars no picker but never shows one. Median **1.94 ms** (0.26 ranges + 1.09 picks + 0.20 sort) on a ~30 ms click, under a comment asserting it did not |
| the mutation sweep (M9) | the staleness test's fixture names had an alphabetical order **identical** to their plausibility order, so the degrade and the mutant produced byte-identical lists |
| the mutation sweep (M18) | `build_ranges`' case-fold test fed already-lowercase keys, making the fold a no-op |
| the mutation sweep (M25) | my own docstring claimed the memo was load-bearing; the three call sites are **mutually exclusive**, so it is an invariant guard. Both the comment and the test are relabelled rather than the unkillable mutant being dropped |
| the **sandbox tier** | `test_the_jsonl_glob_site_ledger_is_pinned_two_way` — adding `.jsonl` to the sweep created a new `*.jsonl` walk site. **1 failure in 22,182, invisible to every dev-host run I did**, because those selected two files and the ledger lives in a third |

## Base-vs-head matrix

Base `8b2b960b`, head as below.

* `test_regen_known_repos.py` (HEAD tests on BASE source): **18 failed, 55 passed**.
  Same file at HEAD: **73 passed**.
* `test_mention_open.py` on BASE source: **collection ERROR** (a module-level
  `parametrize` names `MO.CLASS_PLAUSIBLE`) — red, but uninformative, which is why the
  mutation sweep below is the real matrix.
* **Mutation sweep: 25 KILLED, 1 SURVIVED-as-declared, 0 unaccounted, of 26.** Green
  control on the unmutated tree (146 selected tests), a known-fatal positive control
  KILLED, `PYTHONDONTWRITEBYTECODE=1` throughout, verdicts COUNTED from pytest's own
  result lines, and each row asserting the *expected failure message* so a mutant dying
  for the wrong reason is reported as such.

## Audit ladder, and a deletion

Six adversarial rounds ran, each auditing the previous round's delta. **Rounds 1–2
found real defects in shipped behaviour and stand. Rounds 3–6 were a ladder that had
left the PR, and their subject has been deleted.**

**Rounds 1–2 — kept.** A non-UTF-8 `picks.jsonl` was a **dead click**
(`UnicodeDecodeError` is a `ValueError`, and `load_picks` caught only `OSError`); the
ranges leg could **redden the systemd unit** because `write_ranges` raising escaped
`main()`; **50 private repo names rode the `gh` child's argv**, world-readable in
`/proc/<pid>/cmdline` — a surface the pre-existing `gh api user/repos` does not have;
and the disclosure guard was blind to both new artefacts. Round 2 also caught a
**BLOCKER against my own prose** — a discussions justification that was false at all
three of its cited measurement points.

**Rounds 3–6 — deleted.** They were about a lock on the `picks.jsonl` append, and each
round found a defect in the *previous round's own* scaffolding. Accounting: **272
payload insertions against 788 test insertions**, almost all of it lock code. The
operator challenged it; measured, and it does not survive:

| question | measured |
|---|---|
| do concurrent appends need a lock? | **no** — `O_APPEND` single writes are atomic; 8 processes × 400 appends with the lock removed gave **3200/3200 rows, 0 torn, 0 lost** |
| how wide is the one real window? | **0.28 ms**, opening **once per 500 picks** — it needs a second *human* click inside a quarter-millisecond |
| what does it cost if it strikes? | **one Tier-B learning row**, a soft recency signal the next pick re-establishes |

So `_picks_lock`, its wait budget, its errno filter, the read-offset carry-over, 8 tests
and 2 battery rows are **gone** — `mention-open.py` drops 2698 → 2541 lines, the commit
is **−513 net**. Kept, both justified *without* concurrency: `_compact_picks` itself
(nothing else bounds a 0600 file that only grows) and tmp-then-replace (`os.replace` is
atomic, so a SIGKILL mid-compaction leaves the old log rather than a truncated one).

### Mutation coverage

🔴 **Eight rows are now in the COMMITTED battery** (`mutation_battery_mentions.py`
K64–K71), because round 4 pointed out that every sweep I had quoted ran from a scratch
harness nobody else can run — and that file's own header calls that *"a claim, not
evidence"*. The committed battery previously had **zero** mutants touching any line this
PR changed.

`--only K64,…,K71` → **9/9 KILLED for the stated reason**, P1 positive control killed.
They cover: ordering removed, IMPOSSIBLE filtered instead of ranked, Tier B crossing a
class, the sort alphabetising, a stale table trusted, the UTF-8 dead click, the append
losing its lock, and the mode mask destroying setuid. `test_mutation_battery_anchors.py`
pins their anchors, so a future edit that moves one fails the push.

⚠ Six of the eight first scored `KILLED-WRONG-REASON` — the battery's documented token
trap, in a **third shape its header does not list**: pytest *wraps* a long `E ` line, so
a phrase spanning the wrap is never found. Tokens are now taken from measured output,
and four bare assertions gained real failure messages in the process.

A 43-mutant scratch sweep also ran each round (green control, known-fatal positive
control, `PYTHONDONTWRITEBYTECODE=1`, verdicts counted from result lines). Its one
declared survivor is `universe_rows()`'s memo — an invariant guard, not a live
optimisation: the three call sites are mutually exclusive, so it is **unkillable by
design**, and both the comment and the test say so.

## Gate

Both derivations built **one at a time** (a combined invocation produces false failures),
from a `git archive HEAD | tar -x` export rather than the worktree, reading the runners'
own `RESULT:` lines rather than a piped exit code.

| tier | sha | verdict |
|---|---|---|
| sandbox `pytests` | `facccbd1` | `collected=22195 passed=22193 skipped=2 failed=0`, `SCOPE: FULL (28 of 28)`, **PASS** |
| sandbox `nodetests` | `5c648d3c` | **PASS**, `SCOPE: FULL` |
| Tekton CI | `facccbd1` | pytests **pass**, nodetests **pass**, cairn-client-runs **pass** |
| dev-host, affected suites | `5c648d3c` | `589 passed` |

The tiers at the round-2 head `5c648d3c` are posted as a comment below when they land.

⚠ **Two earlier runs were RED, and both were real.** At `2960108c`, **1 failure in
22,182** — the `.jsonl` glob-site ledger, which adding `.jsonl` to the disclosure sweep
tripped. It was **invisible to every dev-host run I did**, because those selected two
files and the ledger lives in a third; the sandbox tier and Tekton found it
independently and agreed. Fixed in `eb3fd6ab`. At `99ef2996`, the mutation-battery
anchor `K12` had been orphaned by the laziness restructure and was scoring
`PATTERN OCCURS 0x — NOT APPLIED`, i.e. surviving without testing anything. Re-anchored;
`--only K11,K12,K37` confirms it kills again, with the P1 control killed.

⚠ **`scripts/scoped-tests.sh`** at `2960108c` gave `collected=653 passed=653`, but
`SCOPE: SCOPED` — recorded as iteration evidence only. `gate.sh` exits 91 = PARTIAL off
any run that is not `SCOPE: FULL`, so it is not a gate and is not quoted as one.

## Residual risk

* **The `no-table` note appears on every picker until the first refresh — and forever on a
  host where `gh` is absent or not logged in**, which `regen-known-repos.py` exits 4 for as
  a legitimate state. It is deliberately the quietest of the three wordings and names the
  generator rather than a unit, but it is a line that was not there before, on the
  commonest path. Silencing it entirely is a one-line change if you would rather.
* **Tier B ranks above the `max_ref - N` distance** within a class. The brief did not
  settle that; the argument is that a real past pick is a measurement while head-distance
  is a heuristic, and only repos the operator actually chose carry a non-zero score.
* **The table is a daily snapshot.** A repo that advances past `N` within the day is filed
  BELOW, and one that gains its first reference is filed IMPOSSIBLE. Both are *ranked*, not
  hidden — that is the whole reason for ranking.
* **The 45% figure is the brief's 120-repo sample, not re-derived here** (doing so means
  pulling private data). The corpus test asserts its own shape so it cannot silently
  flatten.
* **The click path has not been exercised by a human.** Clicking raises windows, which is a
  `pkill`-class action for an agent. Rank 1 of the mention handoff already needs this.

A pre-push scan of every name this branch ADDS against `gh api user/repos` reports **0**
real repository names, with both controls firing (389 repos known; the extractor proved
able to match a planted real name).

## Overlap with open PRs

Three open PRs touch `nix/home.nix` (#1451, #1177, #1038). My change there is one
`Description` string and one comment block inside the `mention-known-repos-refresh`
service, so a textual conflict is unlikely — but ⚠ **disjoint files are not safety and
neither is a clean textual merge**; no PR touches any of the other five files.

https://claude.ai/code/session_01TcAdJ9nztsud113yVKfWhQ


ZacxDev:
## Merged-tree gate at `63846b81`

Merged `origin/main` (26 commits) into the branch — clean merge, and **none of the seven
files this PR touches had moved on main**, so there were no semantic conflicts to resolve.
(#1454, the fzf picker, was already in the branch point; this PR was built on top of it.)

Both derivations built **one at a time**, from a `git archive HEAD | tar -x` export, reading
the runners' own `RESULT:` lines rather than a piped exit code.

| tier | verdict |
|---|---|
| sandbox `nodetests` | **PASS**, `SCOPE: FULL` |
| sandbox `pytests` | `collected=22326 passed=22322 skipped=2 **failed=2**`, `SCOPE: FULL (28 of 28)` |
| Tekton CI | nodetests **pass**, cairn-client-runs **pass**, pytests **fail** |

### 🔴 Both failures are INHERITED — `main` is currently red

```
test_no_tracked_shell_text_writes_a_kill_this_guard_would_deny
```

Both name `claudedocs/handoff-tmux-webapp.md`, a file this PR does not touch. It landed on
`main` in `60194765` without being classified in `guard_core.py`'s `_KILL_MENTION_LEDGER`.

**Controlled, not assumed:** the two tests were run against a clean `git archive origin/main`
export and fail there with byte-identical messages. This PR contributes **0** failures; the
local sandbox tier and Tekton agree with each other and with the control.

That is a real break on `main` and someone should classify the file — `main-green-check`
should reproduce it on its next 4-hourly run.

### Earlier reds on this branch, both real and both fixed

* `2960108c` — the `.jsonl` glob-site ledger, tripped by adding `.jsonl` to the disclosure
  sweep. **Invisible to every dev-host run** (those selected two files; the ledger lives in a
  third); the sandbox tier and Tekton found it independently. Fixed in `eb3fd6ab`.
* `99ef2996` — mutation-battery anchor `K12`, orphaned by the laziness restructure and
  scoring `PATTERN OCCURS 0x — NOT APPLIED`, i.e. surviving without testing anything.
  Re-anchored; `--only K11,K12,K37` kills again with the P1 control killed.

Issue #1431, opened 2026-09-09
audit-pr battery: MIN_TESTS pin sees only a literal assignment, and one commit message mis-names a control


ZacxDev:
Filed from PR #1427's audit ladder, which stopped on the payload-attribution gate (rounds 4 and 5 both changed zero payload lines). These are the findings that were **not** fixed there, recorded so they read as open rather than absent.

Both are `scaffolding`, both bounded, neither deploy-blocking.

## 1. The floor pin reads a literal, so a conditional override is invisible to it

`scripts/tests/test_audit_ladder_stop_rule.py` — `test_the_batterys_floor_is_re_derived_from_this_modules_size` matches `^MIN_TESTS=(\d+)` in `scripts/tests/mutants-audit-ladder.sh`. Measured during round 5: with

```sh
MIN_TESTS=15
if [ -n "${QUICK:-}" ]; then
  MIN_TESTS=3
fi
```

the guard reports `1 passed` while the shell would apply 3. Same for `MIN_TESTS=$LOW`.

Covered already: a second *numeric* assignment fires the `len(literals) == 1` check, and a commented-out assignment yields 0 and fails loudly. **Residual exposure is collapse-without-deletion only** — deleting tests moves `m`, which the pin does catch.

Possible fix: have the battery print its effective `MIN_TESTS` and assert on that, rather than parsing the source.

## 2. A control in commit `5ab273bf`'s message names the wrong mechanism

It reads *"floor 12 + both PR tests deleted (round 3's) → HARNESS could not run"*. Reproduced: at floor 12 with 14 tests, the floor does **not** fire (14 ≥ 12) — what stopped the battery was the new pin. The outcome (refuse, rc 2, nothing scored) is right; the attribution is wrong, and it reads as evidence the floor mechanism catches test deletion at that setting, which it does not.

Not correctable in place (the message is committed). Recorded here so the claim is not inherited by a later reader.

## Closing condition

Item 1: a PR that makes the pin read the battery's **effective** `MIN_TESTS` (not its source text) is merged, and this control passes — with a `QUICK`-style conditional override present in `mutants-audit-ladder.sh`, `test_the_batterys_floor_is_re_derived_from_this_modules_size` FAILS. Until that control has been watched red, item 1 is open.

Item 2 is a record, not a repair: it closes when a reader confirms in writing on this issue that the mis-named control in `5ab273bf` will not be inherited.

[...]

PR #1511, opened 2026-09-11
docs(handoff): the arc closed out — plus a ranked item of mine that was false and a carry-forward my own update dropped

ZacxDev:
Closes out the `audit-pr-ladder` arc and corrects three stale claims this session left behind.

## Why this exists

The doc on `main` said `#1495` was **"OPEN, unmerged and unaudited"** and ranked **"Audit and merge #1495"** as the next step — hours after it merged, was audited through a three-round ladder, and shipped. A `/resume` session would have re-done finished work.

## What it corrects

- **State now / Next steps / How to verify** → final state: all three PRs merged (`34da597d`, `e1cd9a38`, `cc278b7a`), both hosts verified at `3a0c77dd`, `tekton/devrc-pytests` green at `760c8769` (22,083 collected, 0 failed).
- 🔴 **Ranked item "`ship.sh` cannot reach the laptop off-LAN" is CLOSED — my own item was false.** `#1439` had already shipped the fallback. I hit a LAN timeout early in the arc, wrote the item from that observation, then passed `LAPTOP_SSH=` as an override on both later ship runs *out of habit*, which made my own claim untestable. **Measured with the override removed:** `ship: zach@192.168.50.155 did not answer — falling back to zach@10.42.0.100`, then `VERIFIED`. A prior session had already struck this item; it was right and I was about to re-assert it.
- ✅ **Restores a prior session's `#1431` carry-forward that my own previous commit dropped.** That bullet's text said in terms *"This bullet stays"*, and the `State now` REPLACE ate it while re-asserting the superseded "`#1431` is CLOSED" it existed to correct. `#1431` is **REOPENED** (2026-09-10).

## Recorded as gotchas

- A claim corrected on one surface stays false on the others — this arc did it **six** times (the original `scoped-tests.sh` figure, a commit message, the PR body, the cairn entry, this handoff, and the dropped carry-forward). Every one was caught by a sweep, none by reading.
- A lesson recorded **mid-ladder** records the wrong remedy — the `devrc/tests` cairn bullet named the count ledger as "the remedy that held" when round 1's own next finding was that it did not. Corrected at revision `64a2bba6`.
- `git checkout -- <path>` to restore a mutated file reverts the **committed** version and destroys uncommitted work — done here mid-battery, caught only because a control printed `3 passed` where `2` was expected.
- A `State now` REPLACE will eat another session's deliberate carry-forward; the tool warns in a line that is easy to skim. **Grep the DROPS block for `stays` / `carried forward` before confirming.**

Written by `scripts/lib/handoff_doc.py`; committed on a branch, not `main`.


ZacxDev:
Heads-up on the queue, from the session that just landed #1519 — **not a request to change this PR's approach**, just two pointers it would be good to carry.

The renumbering between the old list (ranks 1–16) and this one dropped **three** items with no closure note. Two of them were genuinely resolved and are visible in `State now` (#1431 reopened, #1287 decided → #1505), so those are fine. The other three have **zero traces** in the doc on `main` today — I grepped for `range-coverage`, `Churn-measure the ladders OUTSIDE`, `stop-rationale` and `42 block-carrying`:

- **old rank 10 — fix the range-coverage hole.** ✅ **DONE**, merged as `766c1295` (#1519). Worth a line in the list so the next `/resume` doesn't re-derive it: the instrument is `scripts/ladder-range-coverage.py`, and it found **interior 655 lines in 2 of 20 ladders, tail 3,727 in 11** — including a second interior gap (#998) that the 2026-09-04 review never named.
- **old rank 8 — churn-measure the ladders OUTSIDE devrc** (`homelab-talos`, `civit-datapacket-talos`, `vetr`, `auditloop`, `civitai-gpu-fleet`, `naida-ai`). **Still open and untouched.** I have just claimed it as `audit-pr-ladder-8` and am working it now, so it needs no rank — but it should not be invisible.
- **old rank 9 — mine the stop-rationale prose across all 42 block-carrying PRs.** **Still open and untouched.** This one survives durably as open item **4** of `claudedocs/audit-ladder-review-2026-09-04.md`, so it is not lost — but nothing in the handoff points at it any more.

⚠ The reason this is worth a note rather than a shrug: the list's own preamble used to say *"Numbering is STABLE — the rank is half a `claim-work` slug's identity"*, and I was holding `audit-pr-ladder-10` against a rank that had already ceased to exist. The slug namespace and the list drifted apart silently. No action needed on your side beyond deciding whether 8 and 9 deserve rows.

Also, unrelated and inherited — **`tekton/devrc-pytests` is red on `main` itself** right now, independent of any PR: `test_every_kill_server_call_site_in_the_repo_is_classified` fails because `_KILL_MENTION_LEDGER` does not classify newly-added handoff docs. Measured on a clean `origin/main` worktree at `50e8a71a` (flagging `claudedocs/handoff-tmux-webapp.md`) and on my own branch (flagging `claudedocs/handoff-tmux-scratchpad-bar-statusline.md`). Since every new handoff doc that mentions a wide tmux kill breaks it, it is on its way to being a permanently-red gate — which `claude/RULES.md` rates worse than no gate. Not mine to fix, but somebody should own it.

PR #1518, opened 2026-09-11
feat(resume): retire the corpus-search step — yield measured 0 of 16, and the falsifier was honoured

ZacxDev:
## What this is

The `/resume` skill no longer runs `handoff_search.py`. The step is retired on the
re-measurement its own handoff doc said to wait for — and on the falsifier that doc
wrote down in advance, which fired.

**This is a negative result being honoured, not a bug fix.**

## The measurement

16 `/resume` sessions across 5 repos queried the corpus after `#1399` (2026-09-09 → 2026-09-11).

| | pre-fix | post-fix |
|---|---|---|
| adoption | 2 of 11 → 20 of 22 | **16 of 16**, all with `--exclude-slug` |
| self-hit slots | 23 of 60 (38%) | **0 of 48** |
| own doc as #1 hit | 13 of 20 | **0 of 16** |
| **yield** — a hit changed what the session did | 1 of 20 | **0 of 16** |

Every fix in this arc did exactly what it promised. `#1332` moved the command out of
step 3's prose and adoption went 2/11 → 20/22. `#1399` added `--exclude-slug` and the
self-hit defect went to zero. The thing they were proxies for never moved.

## The cause — recall, not an empty corpus

This correction matters, because the falsifier's *stated reason* was wrong even though
its *condition* was right. It predicted low yield would mean "the corpus has little to
offer". The corpus has plenty; the ranker cannot reach it.

- **44 of 48 hits were `[gotcha#0]`** — the same first section of a Gotchas block.
- 29 distinct docs filled 48 slots; two appeared **5× each** across unrelated repos.
- ranks clustered 0.97–1.75 — no separation between a good match and a bad one.

**The decisive case:** session `d775cf17` asked about *"store-api test fails on fsync
stall in Tekton pytests tier"*. The corpus contains `devrc/gate-flake-store-api`, about
exactly that. Reproduced live: **it ranks 11th** — unreachable at the prescribed
`--limit 3`. That session got the doc anyway, from `cairn search 'fsync'`, thirteen
records earlier. Two retrieval surfaces, same question; the surviving one answered it.

So this is **not** evidence that better retrieval over this corpus would fail. Nobody
has tested that, and the doc now says nobody should start without a new reason.

## Instrument controls (all watched to work)

- **Positive control, mandatory:** the same parser and needle over the **pre-fix** window
  returns **37% self-hit slots**, reproducing the independently-recorded 38%. Without it
  a post-fix `0%` is indistinguishable from a dead needle.
- My first extractor had an off-by-one read window (`recs[i:i+6]`; every result sits at
  delta **6**) and manufactured **four false "no result" sessions**. Widened; all 16 found.
- Two apparent "acted on a hit" readings were substring collisions — `clickup-mirror` is a
  prefix of `clickup-mirror-check`, each session's own handoff, and also a script directory
  both were working in. On exact filename match it is 0 of 16.
- Laptop measured over SSH, not assumed: **0 resume runs**. All numbers are workbench-only.

## 🔴 The guard is inverted, and its first draft was broken

`test_resume_handoff_search_wiring.py` pinned that the skill **calls** the search. It is
replaced by `test_resume_handoff_search_retired.py`, which fails if the command returns
and hands over the measurement that must be redone first. That matters because a
retirement decided on data is exactly what a later session re-adds in good faith from the
true observation *"the index is live and nothing calls it"* — the very sentence that
motivated the original guard.

**The first draft of the new guard survived its own mutation test.** It scoped the scan to
fenced blocks with `^``` ` anchored at column 0 — but every fence in `resume/SKILL.md` is
indented three spaces inside a numbered list item, so it matched **zero** fences and "no
invocation among the blocks" was vacuously true forever. It went green on a mutant that
re-added the command verbatim. Fixed, plus a scanner positive control
(`test_the_fence_scanner_actually_finds_this_skills_fences`) that fails if the block list is
ever empty again.

Mutation verified **after** the fix: RED with this guard's own error message on the
re-added command, green once reverted. 10 tests pass.

## What is NOT changed

`scripts/lib/handoff_search.py`, the index, the sync timer and `--exclude-slug` all remain
and are supported for ad-hoc use. What was retired is paying ~4 KB of every `/resume` for it.

## Verification

- `test_resume_handoff_search_retired.py`: 10 passed; mutation watched RED then green.
- Change-scoped suite via `scripts/scoped-tests.sh` (10 files selected).
- ⚠ Not a full-gate claim: this is `SCOPE: SCOPED`, so CI's two tiers are what run everything
  it skipped. Branch protection is off on this repo by standing operator decision.

https://claude.ai/code/session_016V17twsscizyqxZohvzVBa


ZacxDev:
## Closing — the premise is falsified. The measurement behind this PR was wrong.

Two independent audits (round 0 and round 1) each found the same counterexample, and I verified it from the raw transcript. **The yield figure this PR is built on — "0 of 16" — is false**, and so is the "1 of 20" baseline it was compared against, because both came from the same broken instrument.

### The counterexample

Session `5320d1bf` (2026-09-09T19:34Z, inside this PR's own window) ran the prescribed step-4 command. Its #1 hit was `datapacket-talos/cnpg-alerting-hardening [investigation#3]`. It then:

- reported it to the operator under step 5's own `from handoff docs` label — the same `maxmemory` asymmetry from the other side, **with ticket `868kurzhm` attached**, noting "a thread this arc's docs don't reference";
- **wrote that pointer into its committed handoff doc** as a durable "⚠ Recall pointer, NOT verified this session", with an action attached.

Control: the string appears **zero** times in that session before the query. The search is the only possible source. That is "mentioned and acted on" by the claim's own wording — and it is the highest-value thing this step can do.

### Why my instrument could not see it

Three compounding flaws in the "did the session act on a hit" half:

1. the needle required a `handoff-` prefix that the sessions' reports don't use (they write `<repo>/<slug>`);
2. assistant text was truncated at 400 chars — the mentions sit at offset ~3,100;
3. for `Write`/`Edit` it captured only `file_path`, never the content — so a hit written *into* a handoff doc was structurally invisible.

The claim was also stated wider than the instrument: the doc's own method step reads *"was a hit DOCUMENT **opened**"*, while the sentence quoted everywhere is *"opened, **mentioned** or acted on"*. `claude/RULES.md` → *"a guard's DESCRIPTION claims COVERAGE — check the implementation is as wide as the sentence."*

### Corrected numbers, same criterion on both windows

| | sessions | referenced a fresh hit afterwards |
|---|---|---|
| pre-fix (`#1332`..`#1399`) | 38 | **12 (32%)** |
| post-fix (`#1399`..now) | 17 | **9 (53%)** |

⚠ That mechanical criterion **overcounts**: step 5 *requires* reporting what step 4 recalled, so several of the nine are compliance, and their own text says "nothing bears on this". The strict "a hit changed a decision or a durable doc" count is **3 firm, ~5 generous, of 17** — against 1 of 20 previously believed. Hand-verified clear cases: `5320d1bf` (above), `8152b7fc` (a hit told it a *different* doc's "one line to public-launch" item is now **false** after #4685), `9b6235dd` (supplied a positive-control argument).

Either way it is **not zero**, and the `--exclude-slug` fix did not reduce it.

### Consequence

The falsifier's condition — *"yield stays ~1/20 with foreign hits filling every slot"* — **did not fire**. The retirement rests on a number that is wrong, so this PR is closed unmerged and **the step stays wired**.

The descriptive half of the measurement was sound and is worth keeping: 16 sessions, 48 slots, **0 self-hit slots** (against 23 of 60 before), own doc the #1 hit **0 of 16** times. `--exclude-slug` works exactly as designed. That is the part that survives.

### Also found, and not lost with this PR

- 🟡 The replacement guard's needle missed three realistic re-wiring shapes (bare `./handoff_search.py`, `$DEVRC/…`, relative path) — all three **survived** a mutation sweep. Cause was one non-load-bearing lookbehind. Moot now, but recorded.
- 🟡 The deleted 558-line guard held the only `bash -n` check on step 4's fence, and the only statement of the cross-repo/client-content hazard. Both would have been lost silently. **This is the strongest argument against the deletion as authored.**
- ⚠ Independent of this PR: `tekton/devrc-pytests` is red on `test_every_kill_server_call_site_in_the_repo_is_classified` (1 of 22,149) — **pre-existing on `main`**, `claudedocs/handoff-tmux-webapp.md` is not in `_KILL_MENTION_LEDGER`. Not caused here; someone should fix the ledger.

A follow-up PR corrects the false figures in `claudedocs/handoff-handoff-search-index.md`, since "1 of 20" and "0 of 16" are currently written there as established fact.

PR #1524, opened 2026-09-11
feat(ci-signal): name the commit that already fixed a PR's inherited red

ZacxDev:
## Why

**MEASURED:** of the **6** genuine `tekton/devrc-pytests` failures in the whole post-#1458 window, **3 — half — were the same test** on PRs **17, 40 and 40** commits behind `main`, all since merged, and that test passes on current `main`. An earlier sample had 8 of 8 failing open PRs 5–42 commits behind, with a rebase curing 4 outright.

Nothing blocks a merge here, so the gate's only value is whether a human *believes* a red — and half the genuine reds are already-fixed failures a human then debugs against a diff that cannot reach them. `strict: true` is deliberately off and correctly so, so this attacks the *triage* cost rather than the staleness.

## What

`scripts/stale-base-triage.py --pr N` / `--sweep`. It **runs no tests** and **writes nothing anywhere**. For each failing test named in the newest `tekton/devrc-pytests` status it emits a deterministic evidence pair:

1. the PR head's copy of the file defining that test is **byte-identical to the merge-base's** — so this change never touched it; and
2. `origin/main` **HAS moved that file** since the merge-base, in commits **absent from the head**.

Verdict per PR: `INHERITED — likely cured by rebase` (naming the commit and the test) · `NOT EXPLAINED BY STALENESS` · `COULD NOT MEASURE` with a reason.

---

## 🔴 Round 0 found the headline verdict fired on 0 of the 3 cases that justified it. Fixed.

`names_provably_complete` could prove completeness only by reading `failed=N` out of the status description. `failed=` is the **last** field in the gate's banner and a status description is capped at **140 BYTES** (measured: every real row sampled is 138 *characters*, which is 140 bytes once the `—` in `FAILED: pytests —` counts as three — ⚠ whether that byte cut is GitHub's or the posting pipeline's own truncation is **not** determined here, because every row sampled carries exactly one multi-byte character), so whether it survives is decided by **how long the failing test's name is**. On the three PRs this tool's requirement was measured on — #1454, #1462, #1499 — it did not survive. All three resolved to `COULD NOT MEASURE` **with the correct answer already in the row**.

**The fix is a second route, and it is a proof rather than a heuristic.** `scripts/run-tests.sh` GUARD 4 defines

```
collected   = passed + skipped + failed + errors + xfailed + xpassed
TOT_FAILED += failed + errors                    (the banner's failed= is f+e)
```

so `collected − passed − skipped == failed + xfailed + xpassed ≥ failed`, and names are a subset of the failures, so `len(names) ≤ failed ≤ derived`. When **`derived == len(names)`** the inequality is squeezed shut and completeness is **proven**. `collected=`, `passed=` and `skipped=` all *precede* `failed=`, so they are exactly the fields that survive the cut. The direct `failed=N` route is kept and still tried first, because it is the stronger claim.

### Yield, measured read-only on real status rows

| population | provably complete, before | after |
|---|---|---|
| 100 most-recently-updated closed PRs | 5 / 28 red heads | **18 / 28** |
| — the three justifying PRs #1454 #1462 #1499 | **0 / 3** | **3 / 3** |
| 58 open PRs | 4 / 26 red heads | **7 / 26** |
| — #1518, #1515 (9 commits behind, actively worked) | **0 / 2** | **2 / 2** |

### 🔴 The soundness control, because this is the half that can be wrong in the dangerous direction

An **over**-count only ever makes the tool withhold. An **under**-count would certify a completeness the row does not have and dismiss a real red — the one error this tool must not make.

- On the same real rows, **17 carried both a visible `failed=N` and a derivable bound. The two agreed exactly on all 17. Zero under-counts.**
- Genuine multi-failure reds stay withheld: **#1177** derives **39** against one name, **#1280** `failed=5`, **#1440** `failed=3` — all still `COULD NOT MEASURE`, with the `HINT:` line.
- The bound is structurally an over-count: `xfailed`/`xpassed` can only inflate it, and the three fields are matched as **one adjacent, ordered group** so a number the cap cut *short* cannot shrink it. `collected=` and `passed=` are proven intact by having matched banner text after them; `skipped=`, the only one that can be short, *subtracts less* and so errs toward refusing.

### The coupling is pinned mechanically, not asserted

`test_the_run_tests_collected_arithmetic_this_derivation_rests_on_is_pinned` reads the `collected=$((…))` and `TOT_FAILED=$((…))` expressions out of the shell source and checks the **containment relation** — so renaming the shell locals is fine, and changing *what is summed* is not. It also requires **every** `TOTAL` banner line to print the fields in order (there are two; a first version searched the whole file and was satisfied by the other one — the mutation sweep caught that). ⚠ What it *cannot* check — that `xfailed`/`xpassed` are non-negative counts — is **stated in the test, not asserted**.

### Verified against the original symptom, on live data

```
old code (5ad08e5a), --pr 1518:   VERDICT: COULD NOT MEASURE    rc 0
this tree,           --pr 1518:   VERDICT: INHERITED            rc 10
```

A four-PR read-only sweep returns `INHERITED` for #1518/#1515/#1450 naming the fix commits, and `COULD NOT MEASURE` for #1177.

---

## 🔴 Prior art, and what this adds to it — honestly

`scripts/main-status-watch.py`'s `screen_all_known_flakes` is **the same completeness-proving screen in 15 lines**, and it shipped first. Its own comment records that it would have fired **zero** times on the 100 commits measured. Round 0 was right that this PR rebuilt that gate and cited it nowhere. It now cites it, in the header, in both files.

**What this tool adds:**

1. **The derived route explains that zero.** The screen is not inert because the *idea* is unsatisfiable — it is inert because it reads only `failed=N`, the field the cap eats. That is a measurable answer to handoff rank 6's open question ("is that screen inert?", due 2026-10-11): **it is inert for a reason, and the reason is fixable.** On the rows measured the derived route moves provable completeness from 5/28 to 18/28.
2. **The evidence half, which the screen has no equivalent of.** The screen answers *may I stay silent about this red?* from a name ledger. This answers *which commit on `main` already fixed it?* from blob OIDs and commit lists — `resolve_test_file`'s two-phase search, the merge-base/head/main blob compare, `commits_touching`, `prove_candidates`. That is the ~150 lines round 0 kept, and it is where the value is.

**What it does not add, said plainly:** the completeness gate itself is prior art. It is not a new idea here, and the honest summary of the *gate's* contribution is "the same 15 lines, plus the derivation that makes them fire".

**The derived route is deliberately NOT back-ported to the live watcher, and the reason is the direction of the error.** In the triage tool, proving completeness makes it **speak**. In the watcher, it makes it stay **silent** about a red on `main`. Widening a silence over main's only automated detector is a change to make on purpose with its own measurement, not a free win inherited from a sibling. `main-status-watch.py` now says so where the screen is defined.

---

## 🔴 Report only by default

`COMMENT_MODE_DEFAULT` is the literal `"off"`, pinned by `test_the_comment_mode_default_is_the_LITERAL_off` **both as a value and as source text** — so arming the PR-comment writer costs a visible line in the arming commit. `dry-run` says what it would post; `on` posts once, behind an idempotence marker. **Nothing has been posted to GitHub at any point, and every live run in this work was read-only with `mode=off`.**

## One rule, one place (D6) — and the two copies had already diverged

`classify` here was **byte-identical** to `main-status-watch.py`'s; `_FAILING_RE`, `_FAILED_COUNT_RE`, `parse_failing_names` and `parse_failed_count` were duplicated **and already divergent**. All five now live in **`scripts/lib/ci_status.py`**, imported by both. Each divergence was reconciled deliberately, because `main-status-watch.py` is **live on both hosts every 10 minutes**:

- **The fold.** The live copy kept the *first* row per context, which is right only because GitHub returns the array newest-first — an assumption about someone else's response ordering, load-bearing and unstated. ⚠ *Correction to the round-0 framing:* first-wins over a newest-first array yields the **newest** row, not the oldest. The defect is order-**dependence**, not a wrong row today. It now folds on `max(created_at)`, which agrees with first-wins on GitHub's current ordering and stays right if that changes; with no `created_at` anywhere, first-wins is the fallback. `test_newest_per_context_is_ORDER_INDEPENDENT_which_first_wins_was_not` reimplements the old rule and shows it returning the superseded `pending` row on the order it does not expect.
- **The context filter** moved out of the fold into `commit_verdict`, where the policy belongs, and *before* the emptiness check — a commit carrying only a foreign pipeline's rows must have **no** verdict, never green.
- **The `isinstance` skip.** The live copy deliberately had none, arguing the arm was unreachable and that dropping a row could drop the red. The shared function keeps the triage tool's skip: still unreachable for the watcher (its walk validates every row first), unchanged for the triage tool.
- **The fragment strip** is now shared and is *provably inert* for the watcher: a row cut at `| TOTA` was necessarily cut before `failed=N` too.

Duplication is refused mechanically: `test_the_shared_predicates_are_NOT_re_declared_in_either_consumer` fails on a local `def` of any of the five, in either file, with a positive control so the scan cannot pass vacuously.

Two pre-existing tests broke, both for the right reason, both fixed rather than weakened: one used `def classify(...)` as its positive control for the docstring-stripper, and one copies the script to a tmp dir and runs it (no sibling `lib/`; it now passes `PYTHONPATH` and says why).

**Deployment:** the unit runs the script straight out of the checkout, so a `git pull` delivers both files and no home-manager switch is involved. ⚠ `X-Restart-Triggers` in `nix/home.nix` still names only `main-status-watch.py` and was deliberately **not** edited — for a timer-driven oneshot that re-reads the file every run it is cosmetic.

## Deletion pass, re-measured *after* F2 (D1–D5)

| candidate | decision | evidence |
|---|---|---|
| `--fetch` | **DELETE** | The only git subcommand here that **writes**, in a tool whose safety claim is that it writes nothing. Measured: this repo's PRs are same-repo branches and `origin` fetches `+refs/heads/*`, so **58 of 58 open PR heads already resolved in the base clone** with no `--fetch`. |
| `--json` | **DELETE** | No consumer. A whole-repo search for `stale-base-triage`/`STALE_BASE_TRIAGE` returns this script and its test file, nothing else. It cost a global `_JSON_MODE`, a `say()` indirection on every human line, and a second exit path. |
| `--sweep` | **KEEP** | Its rationale is exactly what F2 reversed: the sweep now surfaces **7** provably-complete open reds where it surfaced 4, and the three it gained are 9–94 commits behind and **actively worked** — not the 259/357/508-behind abandoned ones. |
| self-parsing `--help` | **KEEP** | What it prints is the `#` header — env ledger, exit codes, measured rationale — which argparse cannot see. The alternative is copying the ledger into an epilog, i.e. reintroducing the duplication D6 just removed. |
| the 3 env seams + ledger test | **KEEP** | `_GH`/`_REPO` are the seams the test harness points at a stub — they are *why* no test here can reach real GitHub. `_BUDGET` is the only way to drive the budget guard to zero. `_COMMENT_MODE` is the arming seam and fails closed across seven spellings. |
| duplicated predicates (D6) | **DONE** | Above. |

🔴 Deleting `--fetch` made the safety guard **stronger, not weaker**: `test_only_refs_under_its_own_namespace_are_ever_written` (a *relative* claim about a repo-**global** surface) is replaced by `test_no_git_subcommand_that_WRITES_is_ever_invoked`, which enumerates the read-only subcommands as an **allowlist** off the AST — so a subcommand nobody enumerated is a write by default.

## Traps handled, each pinned by a test

- `/commits/{sha}/status` (**singular**) maps `error` onto `failure` — never read. Only the plural list, which carries no roll-up field at all.
- The list is **newest-first** — folded on `max(created_at)`, so it is order-independent rather than merely order-lucky.
- `error` is a **broken gate, not a code failure** — its own summary line, never inside the red total.
- A **squash** merge never makes a head an ancestor of its base. Ancestry is asked only about the PR head; every "this landed" claim is made by **blob OID**.
- `git diff --quiet <ref> -- <path>` exits 0 when the path exists on **neither** side. Existence is proved with `git cat-file -e` first — pinned by an **AST scan** of the git invocations, because a string search would have been satisfied by the comment explaining the hazard.
- The **140-byte description cap**: `INHERITED` at PR level requires provable completeness by *either* route above. Deliberately conservative — the `HINT:` line still reports a partially-inherited red.
- The test-name → file mapping is **searched, never guessed**, with a distinct outcome for not-found, ambiguous, class-qualified, parametrised and cap-truncated names. Two phases, and the order is the guard.

## Testing

**Red/green for the F2 change:** **13 of the 14 new tests fail at the PR's previous head `5ad08e5a` and pass at HEAD.** The 14th is a negative control that is green at base too, and its docstring says so — it is an *invariant* guard, not regression coverage, and is not counted as such.

**Suites:** `test_stale_base_triage.py` 56 → **71**. `test_main_status_watch.py` 96 → **101**. Together with `test_no_real_launchers.py`: **252 passed.**

🔴 **A repo-wide guard went red, and the change-scoped mapping could not see it.** `test_no_real_launchers.py::test_every_hazardous_binary_the_scripts_reach_is_stubbed_or_acknowledged` failed on the D6 commit: `launcher_scan.hazard_hits` is a **text** scan over `scripts/` — a prose mention counts, deliberately and fail-closed — and the new import comment says "no home-manager switch is involved". **Re-justified, not reworded**, which is this repo's stated convention written into the table itself, and **pinned**, because that entry says in its own words that an acknowledgement "would otherwise blind the guard" (it has been blinded before: a real `home-manager switch` added to an already-acknowledged script once passed 54 guard tests). Two new tests carry the row — an AST argv[0] set that grows-or-shrinks, and a mention-exists-but-is-never-spawned pair. **Both controls watched, both directions:** injecting `subprocess.run(["home-manager", "switch"])` fails both new tests with *their own* messages; rewording the mention away fails the mention-must-exist half **and** the ledger's file-set pin. ⚠ Found only by running the whole `scripts/tests` target — `scoped-tests.sh` selected just the two suites named in the diff, and a change that adds only a *comment* can go red here.

**Mutation sweep on the new completeness logic**, `PYTHONDONTWRITEBYTECODE=1`, every edit verified to have changed the file, verdicts read from the runner's own result lines:

| round | mutants | killed | survived | negative control | positive control |
|---|---|---|---|---|---|
| 1 | 15 | 10 | 5 | GREEN | KILLED |
| 2 (delta) | 8 | 6 | 2, both argued **equivalent** | GREEN | KILLED |

Round 1's five survivors were **four real gaps in the tests** (fixed) and one equivalent mutant:

- **M14** — reordering `passed=`/`skipped=` in the `TOTAL` banner survived, because the pin searched the whole *file* and `run-tests.sh` prints **two** banners; mutating one was satisfied by the other.
- **M11** — `(\d*)` for the `skipped` group survived: no fixture fed a row cut exactly on the `=`, which makes `int("")` raise on the reporting path.
- **M10** — replacing the underivable-row refusal with a bound of 0 survived, because `names` is non-empty by then so the *verdict* is identical; only the operator-facing reason distinguishes them, and both reasons are now pinned as whole strings.
- **M5b** — the naive three-separate-searches spelling is now driven explicitly and is killed.

The two round-2 survivors are argued equivalent rather than fixed, and labelled as such in the source: **M5** (`.*` between the fields still requires the *order*, and a single-banner description has exactly one of each field) and **M13** (`run-tests.sh` *shrinking* what it sums into `failed=` keeps the bound an upper bound — the pin checks containment **on purpose**; **M15**, a term *leaving* `collected`, is killed by the same pin). Round 2 produced no new finding, so the ladder ends there.

## 🔴 This PR's own CI is red, and the tool in it triaged that red

`tekton/devrc-pytests` fails on `test_every_kill_server_call_site_in_the_repo_is_classified`. Run `stale-base-triage.py --pr 1524` against this head and it:

- resolves the name to `scripts/claude-hooks/tests/test_guard_core.py`, shows that blob is **byte-identical at the head and the merge-base** (`490ed52e57a3`) while `main` has moved it in two commits absent from the head, and
- **withholds the PR-level verdict anyway** — `collected − passed − skipped = 2` against **one** name, so `COULD NOT MEASURE` plus the `HINT:` line.

That is the conservative arm working on its author. And the withholding was **right**, which is the part worth reading: a locally built **merged tree** does *not* cure it, and a pristine `origin/main` worktree (`f3e27aa3`) **fails the same test on its own** — a second unclassified doc, `claudedocs/handoff-cairn-oss-multi-instance.md`, is missing from `_KILL_MENTION_LEDGER`. **`main` is red, independently of this PR.** Had the tool asserted `INHERITED — likely cured by rebase` here, it would have been wrong; it says "likely" and it declined to say even that.

**Attribution of the red, from this PR's own arithmetic:** previous head `collected=22204 passed=22200 skipped=2`; this head `collected=22224 passed=22220 skipped=2`. **+20 collected, +20 passed** — exactly the 15 + 5 tests added here, all passing — with the failure count unchanged at 2. My diff touches six files, all under `scripts/`, and no `claudedocs/` file at all. On the merged tree the three suites still report **252 passed**.

## Not verified

- **No gate tier was run locally** — neither `scripts/gate.sh` nor the `nix build` sandbox. A full `scripts/tests` target run was attempted twice and **produced no verdict either time**: the box was at load ~55 with six other full-target runs from another session, and it hit its 90-minute cap. Per the repo's own rule that is *could-not-vouch*, **not a pass**, and it is recorded that way. What did run to completion, on both the branch and the merged tree, is the three affected suites (**252 passed**) plus a 1,150-test subset covering every repo-wide scanner identifiable by grep — through the dev-shell pytest door, which bypasses GUARDs 7/8/9/10. **CI is the authoritative signal, and it is read above.**
- The tool says "**likely** cured by rebase" and means it: a commit touching the failing test's file is a *candidate* fix. Only re-running the suite on the merged tree proves one.
- The write path was exercised only against a stub `gh`. **No real comment has ever been posted.**
- `main-status-watch.py` was **not** exercised against live GitHub — only its 99 tests plus a `--help` smoke run. The deployed copy is untouched: this work happened in a worktree, and the live unit runs the base clone's `main`.
- The blob half of the verdict is unmeasured for #1454/#1462/#1499 specifically — their heads are not in the local clone. It **is** verified live on #1518, #1515, #1450 and #1177.

### DIFF
--- a/scripts/lib/ci_status.py
+++ b/scripts/lib/ci_status.py
@@ -1,0 +1,208 @@
+"""GitHub commit-status reading for this repo's `tekton/devrc-*` gate.
+
+🔴 ONE RULE, ONE PLACE. Every predicate here was open-coded in BOTH
+`scripts/main-status-watch.py` (live, every 10 minutes, on both hosts) and
+`scripts/stale-base-triage.py`, and the copies had ALREADY DIVERGED before
+anybody noticed — only the newer one stripped the `TOTAL` truncation fragment,
+and only the newer one folded statuses by timestamp instead of by array
+position. Consolidation is what made that disagreement audible; the divergence
+was the finding, not the tidiness.
+
+WHAT LIVES HERE, AND WHAT DELIBERATELY DOES NOT. This module owns the rules that
+are about GITHUB'S DATA and about the GATE'S OWN OUTPUT FORMAT — how a status row
+maps to a verdict class, how a commit's rows fold to a current answer, and how
+the `FAILING: … | TOTAL collected=… ` banner parses. It owns no policy: which
+CONTEXT to look at, whether a foreign context may speak, what to do about a red,
+and whether a named set is provably complete are each the CALLER'S decision and
+stay in the caller, because the two callers genuinely answer them differently.
+
+🔴 AND EACH CALLER KEEPS ITS OWN ROW-VALIDITY POLICY, WHICH IS NOT THE SAME
+POLICY. `main-status-watch.py` validates every decoded row is a dict in its walk
+and names the commit when one is not — so it must NEVER silently drop a row,
+because a dropped row could be the red it exists to catch. `stale-base-triage.py`
+has no such walk and skips a malformed row. `newest_per_context` below therefore
+skips non-dict rows: for the watcher that arm is UNREACHABLE (its caller already
+proved every row is a dict), and for the triage tool it is the existing
+behaviour. Neither caller's contract changes; see
+`test_newest_per_context_skips_a_malformed_row_rather_than_raising`.
+"""
+import re
+
+# ── status classification ─────────────────────────────────────────────────────
+# 🔴 ONLY `state == "failure"` IS A RED. Every non-code outcome this pipeline
+# reports — superseded, KILLED, NO GATE POD, clone rc 128 — arrives as
+# `state == "error"`. Treating `error` as red is the `/commits/{sha}/status`
+# roll-up mistake in a second spelling; treating it as GREEN would be worse
+# still, so an unrecognised state is classified as NOT-A-VERDICT and the caller
+# walks past it. That is the narrowest rule that can be wrong in one direction.
+#
+# ⚠ THE FINER CLASSES NEED A TEST THAT DRIVES THIS FUNCTION DIRECTLY. Both
+# callers fold `superseded`/`killed`/`no-gate-pod`/`error-other` together, so an
+# end-to-end test can only ever see "not a verdict" — measured, 35 of this
+# function's 60 enumerated mutants survived a fully green 61-test suite, and the
+# two fall-through `return "error-other"`s were reached by no fixture at all:
+# turning either into `return "green"` survived. That is a state GitHub adds
+# tomorrow folded into "main is green", closing a red episode.
+
+
+def classify(state, description):
+    """One status row -> a verdict class. Pure, and the tests drive it directly.
+
+    🔴 UNRECOGNISED MEANS `error-other`, WHICH MEANS NOT-A-VERDICT — never
+    `red`, and never `green`. Both fall-through returns below are the arm an
+    unknown `state` lands on, and "absence reads as green" is the single failure
+    the watcher that consumes this is built against.
+    """
+    desc = (description or "").strip()
+    if state == "success":
+        return "green"
+    if state == "failure":
+        return "red"
+    if state == "pending":
+        return "pending"
+    if state == "error":
+        if desc.startswith("superseded"):
+            return "superseded"
+        if desc.startswith("KILLED"):
+            return "killed"
+        if desc.startswith("NO GATE POD"):
+            return "no-gate-pod"
+        return "error-other"
+    return "error-other"
+
+
+def newest_per_context(rows):
+    """Fold status rows to the NEWEST row per context, by `created_at`.
+
+    A commit accumulates a `pending` row and then its verdict under the SAME
+    context, so folding by context is what turns the history into a current
+    answer. Without it a stale `pending` from 20 minutes ago outvotes the
+    verdict that replaced it.
+
+    🔴 NOT first-wins AND NOT last-wins — BY TIMESTAMP, so the array's order is
+    not an input at all. GitHub returns this list newest-first today, which
+    makes first-wins accidentally correct and last-wins catastrophically wrong:
+    an agent shipped last-wins and reported a known-green head as `pending`. The
+    max-`created_at` fold agrees with first-wins on the order GitHub uses today
+    AND stays right if that order ever changes, so the hazard is removed rather
+    than avoided. `test_newest_per_context_is_order_independent` feeds both
+    orders and pins that they agree.
+
+    ⚠ WITH NO `created_at` ANYWHERE — a shape GitHub does not produce, but a
+    fixture might — every stamp compares equal and the FIRST row per context
+    wins, which is exactly the previous behaviour. The fallback is the old rule,
+    never an arbitrary one.
+
+    Returns EVERY context it saw. Filtering to the contexts a caller is allowed
+    to believe is the caller's job, and `main-status-watch.py` does it in
+    `commit_verdict`: a foreign pipeline's green must not read as this gate's.
+    """
+    best = {}
+    for row in rows:
+        # See the module docstring: skipping is the triage tool's existing
+        # behaviour, and unreachable for the watcher, whose walk has already
+        # proved every row is a dict and named the commit if one was not.
+        if not isinstance(row, dict):
+            continue
+        ctx = row.get("context") or ""
+        if not ctx:
+            continue
+        stamp = row.get("created_at") or ""
+        if ctx not in best or stamp > best[ctx][0]:
+            best[ctx] = (stamp, row)
+    return {ctx: row for ctx, (_, row) in best.items()}
+
+
+# ── the gate's own banner ─────────────────────────────────────────────────────
+# The line `run-tests.sh` prints and the pipeline posts:
+#   FAILED: pytests — FAILING: a | b | TOTAL collected=N  passed=N  skipped=N
+#                                     failed=N  (floor: …)
+# GitHub caps a status description at 140 BYTES, so most real rows arrive cut.
+FAILING_RE = re.compile(r"FAILING:\s*(.+?)(?:\s*\|\s*TOTAL\b|$)")
+FAILED_COUNT_RE = re.compile(r"\bfailed=(\d+)\b")
+
+# The only fragments a cut can leave behind where `TOTAL` was starting.
+# Enumerated rather than pattern-matched: a `startswith` test would also eat a
+# real name, and every test in this repo begins `test_` or `Test`, neither of
+# which is a prefix of `TOTAL`.
+TOTAL_FRAGMENTS = ("T", "TO", "TOT", "TOTA")
+
+
+def parse_failing_names(description):
+    """The test names the description NAMES. May be INCOMPLETE — see the callers.
+
+    A description cut inside ` | TOTAL` leaves a trailing fragment of the word
+    TOTAL sitting where a name would be. Left in, it resolves to no file and the
+    PR is reported as a real red on the strength of a truncation artefact.
+
+    ⚠ THE FRAGMENT STRIP IS INERT FOR A COMPLETENESS SCREEN AND LOAD-BEARING FOR
+    A NAME->FILE LOOKUP, which is why the two copies could diverge unnoticed: a
+    row cut at `| TOTA` was necessarily cut before `failed=N` too, so a screen
+    that also demands the count refuses either way. Only a caller that RESOLVES
+    the names can be hurt by the fragment.
+    """
+    m = FAILING_RE.search(description or "")
+    if not m:
+        return []
+    names = [n.strip() for n in m.group(1).split("|") if n.strip()]
+    if names and names[-1] in TOTAL_FRAGMENTS:
+        names.pop()
+    return names
+
+
+def parse_failed_count(description):
+    """The banner's own `failed=N`, or None when the cap ate it.
+
+    It is the LAST field in the banner, so whether it survives is decided by the
+    failing test's NAME LENGTH. `derived_failure_upper_bound` is the route that
+    does not depend on that.
+    """
+    m = FAILED_COUNT_RE.search(description or "")
+    return int(m.group(1)) if m else None
+
+
+# 🔴 THE THREE FIELDS ARE MATCHED AS ONE ADJACENT, ORDERED GROUP, NOT SEPARATELY,
+# AND THAT IS A SOUNDNESS REQUIREMENT RATHER THAN TIDINESS. The hazard is a
+# number the 140-byte cap cut SHORT — `collected=22204` arriving as
+# `collected=2220` would read as a perfectly well-formed integer and make the
+# derived bound far too SMALL, which is the one direction that can certify a
+# completeness a row does not have. Requiring `passed=` and then `skipped=` to
+# follow proves `collected=` and `passed=` are complete numbers: the cap
+# truncates a SUFFIX, so a field with more banner text after it was not cut.
+# Only `skipped=` — the last of the three — can itself be short, and a short
+# `skipped` SUBTRACTS LESS and so inflates the bound, i.e. errs toward refusing.
+TOTALS_RE = re.compile(r"\bcollected=(\d+)\s+passed=(\d+)\s+skipped=(\d+)")
+
+
+def derived_failure_upper_bound(description):
+    """An UPPER bound on the banner's own `failed=`, or None if underivable.
+
+    `scripts/run-tests.sh` GUARD 4 computes, per target,
+
+        collected = passed + skipped + failed + errors + xfailed + xpassed
+        TOT_FAILED += failed + errors          (the banner's `failed=` is f+e)
+
+    and accumulates each term into the TOTAL banner this description quotes. So
+
+        collected − passed − skipped == failed + xfailed + xpassed >= failed
+
+    🔴 THE INEQUALITY ONLY EVER POINTS ONE WAY. `xfailed`/`xpassed` are
+    non-negative, so this can equal `failed=` but never fall below it. That is
+    what makes it usable as a completeness proof: a caller comparing it against
+    a count of NAMES is comparing against a ceiling, and a ceiling that happens
+    to equal the number of names squeezes `failed` to that same number. An
+    under-count would instead certify completeness on a row that had more
+    failures than it named — the single error a triage tool must not make.
+
+    The coupling to `run-tests.sh` is pinned by
+    `test_the_run_tests_collected_arithmetic_this_derivation_rests_on_is_pinned`.
+    """
+    m = TOTALS_RE.search(description or "")
+    if not m:
+        return None
+    collected, passed, skipped = (int(g) for g in m.groups())
+    derived = collected - passed - skipped
+    # Negative is arithmetically impossible under the definition above, so it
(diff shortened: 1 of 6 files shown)


ZacxDev:
## ROUND 0 — requirements & deletion pass

`round 0 · requirements: 10 (unattributed: 6) · deletion candidates: 6`

**`ran: 1 · changed the outcome: 1`** — recorded for the round-0 trial. This round's F1/F2 are not reachable from any of the nine correctness axes; it is the only round that could have produced them.

**Verdict: `requirement questioned — the provable-completeness gate on the PR-level verdict (R3)`.** Not `proceed to the checklist`: auditing the correctness of code whose headline verdict fires on 0 of the 3 cases that justify it would be steps 3–4 spent on something step 1 catches.

### F1 — the verdict is suppressed on exactly the population the requirement names

The three PRs rank 4 was measured on (#1454, #1462, #1499) all carry `len(desc)=138`, cut before `failed=N`, so `names_provably_complete` is False and each resolves to `COULD NOT MEASURE`. The bot comment would never have fired on any of them.

The discriminator is the **failing test's name length** — `#1286` clears the 140-char cap because its test name is 39 chars; `#1518` does not because its is 58. Of 16 open red heads, 14 have `any_explained=True` but only 3 reach `INHERITED`, and those 3 are 259 / 357 / 508 commits behind (abandoned). `#1518` and `#1515` — 9 behind, actively worked — are suppressed with the correct answer in hand.

### F2 — the fix is ~5 lines and is a proof, not a heuristic

`run-tests.sh` defines `collected = p+s+f+e+x+xp` and `failed = f+e`, so `collected − passed − skipped = failed + xfailed + xpassed ≥ failed ≥ len(names)`. Hence `collected − passed − skipped == len(names)` **proves** completeness. `collected=`, `passed=` and `skipped=` all precede `failed=` in the banner, so they survive the cut.

| population | current | with derived count |
|---|---|---|
| historical red heads (n=16) | 3/16 | **13/16** |
| — the 3 justifying PRs | **0/3** | **3/3** |
| live open red heads (n=16) | 4/16 | 7/16 |
| — recent/triageable (#1518, #1515) | **0/2** | **2/2** |

*Soundness control:* on all 7 rows where `failed=N` is visible the derived value equals it exactly (incl. `#1440 failed=3 derived=3`, `#1280 failed=5 derived=5`, both correctly still withheld). Zero under-counts. Genuine multi-failure reds stay `COULD NOT MEASURE`.

### Prior art this PR neither cites nor reconciles with

`scripts/main-status-watch.py:232-286` already implements the identical completeness-proving screen **in 15 lines**, and its own comment records that it would have fired ZERO times on the 100 commits measured. This PR rebuilds that gate at 889 lines. Handoff rank 6 carries an open question about whether that screen is inert, due 2026-10-11 — the two are the same question.

`classify` (`:158-181`) is **byte-identical** to `main-status-watch.py:163-186`; `_FAILING_RE` / `_FAILED_COUNT_RE` / `parse_failing_names` / `parse_failed_count` are duplicated **and already divergent** (only the new copy strips `_TOTAL_FRAGMENTS`).

### Not deletion candidates

The evidence engine — `resolve_test_file`'s two-phase search, the blob compare, `commits_touching`, `prove_candidates` (~150 lines) — is correct on live data and carries the value. The per-test evidence block is right on 14 of 16 live reds and prints regardless of the PR-level verdict.

⚠ The 6 deletion candidates (`--sweep`, `--fetch`, `--json`, self-parsing `--help`, env seams, duplicated predicates) were assessed against the **broken** tool; at least `--sweep`'s rationale weakens once F2 lands, since the sweep then surfaces actively-worked PRs rather than abandoned ones. Re-evaluate after F2, not before.

### Not verified in this round
No gate tier was run — neither dev-host nor the `nix build` sandbox. The PR body's mutation tables (74/82/81) were not re-run. Whether #1454/#1462/#1499 produce `INHERITED` end-to-end under F2 is proven only for the completeness half; their head shas are not in the local clone, so the blob half is unmeasured for those three (it *is* verified on #1518, #1515, #1286).

ZacxDev:
## ROUND 1 — claims block

Round 1 ran the nine correctness axes (a first, full audit; round 0 ran separately and does not move the ladder). It returned **no 🔴, three 🟡, three 🟢**. All six were fixed.

legend: `<from>` = the tip THIS round's audit READ · `<to>` = the head THIS round's FIXES produced.

```audit-claims round=1 audited=ebd6b436..d1c4d39f
1. F1 — the citation guard in test_stale_base_triage.py now FOLLOWS its script's scripts/lib imports (derived from the AST, not enumerated), so the shared module is covered; the cited test test_newest_per_context_skips_a_malformed_row_rather_than_raising is now WRITTEN, and the test_main_status_watch.py copy names the sibling as owner rather than implying coverage it lacks.
2. F2 — _git_subcommands emits <computed>/<not-a-list>/<empty-argv>/<no-argv> sentinels instead of silently dropping what it cannot constant-fold, reads argv from the second positional, and every git spawn must now be lexically inside the one _git helper.
3. F3 — the gh write path is pinned structurally rather than by the literal word POST: an AST ledger of reachable gh verbs, plus gh_write_calls deciding read-vs-write from the method AND the field flags; all four "POST" in c sites replaced.
4. F4 — TOTALS_RE is anchored on \bTOTAL\s+collected=, and FAILED_COUNT_RE (which had the identical defect and was NOT filed by the auditor) is anchored too; both err toward REFUSING.
5. F5 — the 140-BYTE correction was swept across all five sites including the test name, replacing prose with a control asserting the two real truncated rows at 138 chars / 140 bytes.
6. F6 — sys.path.append replaces sys.path.insert in both scripts, so scripts/lib no longer precedes stdlib for later imports.
```

⚠ **Two things the next round must know.**

**The branch was REBASED after these fixes** (26 commits behind → 0). The claims above were measured on `d1c4d39f`; the rebase re-parents those same 8 commits onto current `main`. `test_no_real_launchers.py` was changed on **both** sides, so that file's merged result is the one place a semantic conflict could hide.

**Round 1's red CI was INHERITED, not caused by this PR** — `test_every_site_writing_its_OWN_runner_bound_is_in_the_ledger`, which was failing on a pristine `origin/main` worktree. Fixed on main by `#1567` (`6f1867b1`); the rebase picks it up.

PR #1542, opened 2026-09-12
docs(handoff): round-0 trials 3-5 are DONE — ran 6 / changed the outcome 3, so the section STAYS and the TRIGGER is the defect

ZacxDev:
Rank 1 of `claudedocs/handoff-audit-pr-ladder.md`. Claim: `audit-pr-ladder-1`. One commit, `2b8ab137`, re-applied onto current `main`.

## Verdict: the section stays. Every zero was a post-decision dispatch.

`ran: 6 · changed the outcome: 3` — trials 1–2 (2/2) + my trials 3–5 (**3/0**) + the peer's round 0 on `#1518` (1/1). The retirement condition fires only on *ran **and** changed nothing*.

| trial | PR | verdict | could it act? |
|---|---|---|---|
| 3 | `#1523` | `deletion candidate` | ❌ merged **27 min before** dispatch |
| 4 | `#1518` | `deletion candidate` | ❌ PR **closed 5 min before** the report returned |
| 5 | `#1510` | `requirement questioned` | ❌ merged **+6 min** after dispatch, inside the audit |

Runtimes 459 / 920 / 776 s, so no speedup reaches any of them. **Nothing fires round 0 at PR-open time** — it runs as ladder round 0, after a PR is already shipping. Rank 1 is now "build the trigger", not "run more trials". The proof it works is a peer's round 0, which closed `#1518` unmerged.

## Three rounds of adversarial audit on this PR, and two of my own errors

**🔴 I certified a false figure.** Trial 4 was told to re-verify `#1518`'s "0 of 16" rather than accept it. It wrote its own extractor *from the method description* — believing that made it independent — matched **13 of 13 figures including a 36.1% positive control**, and concluded the requirement holds. It doesn't. All three inherited flaws were in the *method*. An N-of-13 agreement between two implementations of one flawed method is replication, not corroboration; the peer's approach — hunt one counterexample by hand — is what worked.

**🔴 I discredited a correct auditor twice, with zeros I never controlled.** Round 2 cited `endswith("exit 1; }")` and `THIS COMMENT USED TO STATE THE INVERSE` as proof `#1185` F1/F3 were fixed. I grepped `audit-dispatch.py`, got nothing, and published *"that string does not exist on `origin/main`"* — in a document that teaches instrument discipline. Both live in `test_audit_dispatch.py`; `:3036` is the **live assertion**, i.e. the auditor's best proof, which I threw away. I searched one half of a two-file subject and published a negative with no positive control.

**🔴 And I fabricated an attribution.** Asked to replace "a peer session" with an id, I wrote that `b6315603` closed `#1518`. It mentions `1518` **zero times** (controls in the same file: `1532` → 65, `1287` → 59). The closer is `a21e60e0…`, the only transcript containing `gh pr close 1518`, two seconds before GitHub's `closedAt`. A checkable id that is wrong is worse than vagueness — it stops the reader looking.

Round 3 also caught a **needle incapable of observing its target**: I measured `#1532`'s gap with `grep -c 'shared-surface'` → 0, and that needle returns 0 **even now the refusal is live**, because the code spells it `SHARED_SURFACE`. Spelled, not structural.

## Why this was re-applied rather than merged

`#1532` merged **6 seconds** after my previous fix commit and `#1533` **36 seconds** after, and `#1533` wrote a 25-line measured block into this same file. `gh pr view --json mergeable` returned `CONFLICTING`. Taking my side would have deleted a peer's merged correction; taking theirs would have dropped measurements only mine had. So I reset onto current `main` and re-applied my unique content through `handoff_doc.py`, which merges section-wise — and cleared the `rerere` preimage the aborted merge recorded. Verified first that `main` carries **none** of my content, so nothing is duplicated; `handoff_doc.py` reported **no dropped durable lines**.

⚠ The force-push momentarily emptied the branch and GitHub auto-closed this PR; reopened, comments intact. Ranks 2–9 are carried forward verbatim from `main` — numbering is preserved because live `claim-work` refs are keyed to it.


ZacxDev:
## Round 1 — applied. 3×🟡 + 2×🟢, all five fixed in `0f341aec`.

Blind, read-only audit. Each finding was re-verified here before editing rather than taken on the auditor's word; the commit message records the commands.

```audit-claims round=1 audited=00c8a6f0..0f341aec
1. 🟡 Rank 1's ledger pair was corrected from `ran: 5 · changed the outcome: >=3` to `ran: 6 · changed: 3`, with its decomposition (trials 1-2 = 2/2, trials 3-5 = 3/0, the peer's round 0 on #1518 = 1/1) written beside it.
2. 🟡 Rank 2 was changed from "✅ CLOSED by #1532 ... is implemented ... Do not re-open" to an OPEN 🟡 item carrying a forcing function and a content-based closing condition, because #1532 is OPEN and origin/main:scripts/scoped-tests.sh has zero occurrences of shared-surface / scripts/testlib / gate-inventory.
3. 🟡 Eight duplicated Gotchas bullets were deleted by EDITING the file, keeping the longer first copy of each; the three bullets genuinely new in 00c8a6f0 were kept.
4. 🟢 The "^## After the fixes" control claim was corrected: a --round 1 brief returns 0 for that pattern too, so a real positive control (origin/main:claude/skills/audit-pr/SKILL.md, both headings present) is now named in the text and added to the verify block.
5. 🟢 Two verify recipes that read a working tree to establish a claim about `main` were rewritten to `git -C $R show origin/main:<path>`.
6. Minors: "~27 KB" for audit-pr/SKILL.md -> 34,112 B measured; a dangling "elsewhere in this doc" pointer to a claim this update deleted; "~8 min" conflating trial 3's 459 s with trial 5's 776 s; a State-now line naming only the first of the PR's commits.
```

### Two notes on the round itself

**The auditor's count was right and mine was wrong.** It reported eight duplicated bullets; my commit-diff instrument reported six, because its lead-extraction regex normalised away wording changes between the two copies (`, AND THE SLUG` vs `; THE SLUG`). I used the auditor's figure after confirming it line by line.

**Finding 2 is the one that mattered.** Marking a gate `✅ CLOSED` while it is not on `main` would have left a session expecting `scoped-tests.sh` to refuse a `scripts/testlib/` diff and instead getting a scoped `RESULT: PASS` off 10 of 69 referencing files — with nothing tracking the gap if #1532 never landed.

### Round 0 trial ledger, recorded here as the retirement condition requires

`ran: 6 · changed the outcome: 3` — trials 1-2 (2/2), my trials 3-5 (3/0, every zero a post-decision dispatch), and the peer's round 0 on #1518 (1/1, which closed that PR unmerged). **The section stays; the defect is the dispatch trigger.**

ZacxDev:
## Round 2 — applied. 5×🟡 + 2×🟢 in `84f5d69b`. Not a clean round, so the ladder continues.

Round 2 dispositioned round 1's findings as **3 actually fixed / 2 partially fixed**, and named the shape of the partials precisely: **I corrected each claim at the site round 1 pointed to and left the document's other copy asserting the old thing.** That is "one rule, one place" applied to prose, and it produced three of this round's five 🟡s.

```audit-claims round=2 audited=0f341aec..84f5d69b
1. 🟡 State now:25 still said in bold 🔴 "RANK 2 IS CLOSED BY ANOTHER SESSION" while rank 2 at :108 (rewritten in round 1) said the opposite; both copies now agree and each points at the other. Re-verified first: #1532 OPEN, mergedAt null, origin/main:scripts/scoped-tests.sh 0 occurrences of shared-surface / scripts/testlib / gate-inventory, control 394 lines and 4 `exit 4` hits.
2. 🟡 :1503 restated the same population as C alone ("changed an outcome at least 3 times") inside the block rank 1 points at; now the pair `ran: 6 · changed: 3` with its decomposition.
3. 🟡 The `git -C $R fetch origin -q` my round-1 fix added to a recipe has been removed — a fetch into a clone you did not make is a write with cross-session blast radius, which audit-dispatch.py renders a prohibition against into every brief it emits. The staleness caveat it was meant to address is now stated once for all three recipes.
4. 🟡 Rank 6 now leads with `IN FLIGHT as #1547`; the PR appeared in the doc only inside a recipe comment 940 lines from the queue. Its sibling half — the new comment had falsified the pre-existing "(fails once rank 6 lands)" parenthetical — is corrected.
5. 🟡 Six investigation blocks under "live diagnosis state" that prescribed already-done work are now marked RESOLVED with measurements: #1035 MERGED, #1185 F1/F2/F3 fixed by #1342 (08ef1d5a), #1287 CLOSED where the block asserts OPEN, and the trigger-list block whose heading states the figure this PR retracts.
6. 🟢 Restored the clause naming session b6315603 as the one that closed #1518 — the single piece of evidence rank 1's KEEP verdict rests on, which survived only in a bullet the de-duplication deleted.
7. 🟢 The dangling-pointer fix named the wrong deleter (de8e36cd, not this update) and claimed "nothing to compare against" when origin/main still carries the reading; both corrected.
```

### 🔴 Two of the auditor's three `#1185` citations do not reproduce

`lines[i].endswith("exit 1; }")` and `THIS COMMENT USED TO STATE THE INVERSE` exist **nowhere** on `origin/main`. The *conclusion* — all three `#1185` findings are fixed by `#1342` — holds; the evidence for it did not. So each resolution note records what I measured instead: the bare `"exit 1"` pin is gone, `test_audit_dispatch.py:2930` names it in the **past** tense, and the guards are now module-level constants at `audit-dispatch.py:2907-2909` (one rule, one place — a better shape than the auditor described).

Also recorded while there: **`#1185` F2's own `Ruled out:` bullet is now FALSE.** It claimed greps for `Wrap the block` / `run_tier` "return zero hits"; measured, they return **1** and **5** — and the fix is what made the elimination false. A ruled-out bullet can expire.

### Verification

Every recipe in the document was re-run after editing: recipe A `1/0/1` with the positive control at 2 hits and `--emit-claims` rc 4; recipe B `0/9/0/1` as written, with no fetch. A fence-aware scan confirms no code fence in the file contains a `git fetch` — the one remaining mention is inside the prohibition's own prose.

ZacxDev:
## Round 3 — applied, and the PR was re-based rather than merged. Ladder continues.

Round 3 returned 3×🔴 + 2×🟡 + 3×🟢 against `0f341aec..84f5d69b`. All are addressed in `2b8ab137`, which re-applies this work onto current `main`.

```audit-claims round=3 audited=84f5d69b..2b8ab137
1. 🔴 #1532 had MERGED (6fa09466, 6s after the previous fix commit), falsifying four sites that said it was OPEN and the gap live; State now and rank 2 now carry main's merged status, and the needle that measured it is recorded as incapable of observing its target (`shared-surface` returns 0 even with the refusal live — the code spells it SHARED_SURFACE).
2. 🔴 #1533 had MERGED (f3e27aa3) and written 25 measured lines into this same file, making the PR CONFLICTING; resolved by resetting onto current main and re-applying only my unique content through handoff_doc.py, after verifying main carries none of it. The rerere preimage from the aborted merge was cleared.
3. 🔴 The restored attribution named the wrong session: b6315603 mentions 1518 zero times (controls: 1532 -> 65, 1287 -> 59); the closer is a21e60e0-7bc4-41b6-af4e-44c3af9ff3e0, the only transcript containing `gh pr close 1518`. Corrected, and recorded as its own gotcha.
4. 🟡 Three measurements inside the RESOLVED notes were false because I grepped scripts/audit-dispatch.py while the citations live in scripts/tests/test_audit_dispatch.py; both cited strings DO exist (endswith at :3036 is the live assertion, the INVERSE comment at :3052), so the round-2 auditor was right and I was wrong. Recorded as a gotcha about publishing a negative with no positive control.
5. 🟡 The self-satisfying recovery recipe and the backwards "below" pointer are gone with the section rewrite; the staleness caveat is stated once, and the no-fetch prohibition is stated where the recipe lives.
6. 🟢 Rank 4's "~27 KB" corrected to 34,112 B measured at 6395ce1f.
7. 🟢 The stale Tekton-capacity hypothesis and the one-directional cross-reference are not re-introduced: ranks 2-9 come from main verbatim and the two spent blocks are not re-asserted by this delta.
```

### Deliberately not done

I did **not** re-mark the six spent investigation blocks RESOLVED in this pass. My previous attempt's markers carried three false measurements, and `main`'s copy of those blocks is untouched by this delta — so they remain unmarked rather than marked wrongly. That is a known gap, stated here rather than papered over; it needs a fresh sweep that greps **both** files of each two-file subject.

### Ledger

`round 3 · payload lines changed THIS round: 42 (since round 1: 93)` as measured by round 3. This commit is a re-application rather than a delta on the same base, so a payload count against the previous tip is not comparable — stated rather than computed, since the attribution gate is structurally inert for a prose payload anyway.

PR #1549, opened 2026-09-12
fix(guard tests): scope the wide-kill ledger to executable text — prose red-lined main 3x in one day

ZacxDev:
fix(guard tests): scope the wide-kill LEDGER to executable text — prose was red-lining main once per session

The mention half of `_KILL_MENTION_LEDGER` scanned EVERY tracked file, so a doc
that merely said the words red-lined `main` until a human classified it.

MEASURED 2026-09-11 — it fired THREE times in one day, all three from handoff
docs:
  1. handoff-tmux-scratchpad-bar-statusline.md  classified by another session
  2. handoff-tmux-webapp.md                     a FALSE positive (`skill-session`
                                                contains `kill-session`), fixed in #1534
  3. handoff-mention-system-repos.md            red now — and its offending line is
                                                DOCUMENTING THIS GUARD

Writing about the guard tripped the guard. That is the permanently-red-gate
shape: every PR in the repo blocked on an unrelated doc, which trains everyone
to click through.

Composition at that point: 11 of 13 entries were real `scripts/` call sites, and
all three failures came from the 2-entry `claudedocs/` half.

WHAT STILL GUARDS PROSE, AND WHY IT IS BETTER AT IT.
`test_no_tracked_shell_text_writes_a_kill_this_guard_would_deny` scans every
tracked file INCLUDING claudedocs, matches only text in real shell-command shape,
and runs each hit through `check_tmux_kill_shared_server` itself. A doc that
SPELLS a dangerous command is still caught, by the check built for that; a doc
that merely discusses one is not.

PROVEN, not asserted. Planting a literal wide kill in a TRACKED claudedocs file
turns that guard RED; the tree restored byte-identical afterwards.
🔴 My first attempt at this control was VACUOUS and is worth recording: I appended
to `claudedocs/handoff-comic-flex.md`, which does not exist in this repo, so the
scanner skipped it and the test passed. A mutant that never runs scores as
SURVIVED. The real control uses a file `git ls-files` returns.

The narrowing immediately paid for itself: with the ledger scoped, the shell-text
guard caught #3 above as a TRUE positive — it really does write `tmux kill-server`
— so it is added to `quoting_is_the_point`, the allowlist built for files where
quoting is the point. That is the right home for it; the ledger never was.

Also drops the two now-out-of-scope `claudedocs/` ledger entries.

Verified: all three kill guards green; the coverage control red-then-restored.

### DIFF
--- a/scripts/claude-hooks/tests/test_guard_core.py
+++ b/scripts/claude-hooks/tests/test_guard_core.py
@@ -2523,8 +2523,6 @@ def test_tmux_kill_prefix_matches_both_wide_kills_and_neither_narrow_one():
 # Two-way: a new file, or a removed one, fails this test and forces a human to
 # classify it. Paths only — line numbers are what rotted last time.
 _KILL_MENTION_LEDGER = {
-    "claudedocs/handoff-tmux-restore-chain.md": "prose: the incident write-up",
-    "claudedocs/handoff-tmux-scratchpad-bar-statusline.md": "prose: a gotcha warning AGAINST it — \"Never `kill-server`; that destroys every session\"",
     "scripts/claude-hooks/bash-guard.py": "prose: the guard's own ban list",
     "scripts/claude-hooks/guard_core.py": "prose: this check, its docstring and its message",
     "scripts/claude-hooks/tests/test_guard_core.py": "prose: these tests' fixtures + this ledger",
@@ -2590,6 +2588,34 @@ def _tracked_files():
     return [str(p.relative_to(_REPO)) for p in _public_ip_scan.repo_files(_REPO)]
 
 
+# 🔴 THE LEDGER IS SCOPED TO EXECUTABLE TEXT, AND PROSE IS NOT LEFT UNGUARDED.
+#
+# The mention half used to scan EVERY tracked file, so any doc that merely says
+# the words "kill-server" red-lined `main` until a human classified it. Measured
+# 2026-09-11: it fired THREE times in one day, all three from handoff docs, and
+# one of them was a doc whose only content was DOCUMENTING THIS GUARD — writing
+# about it tripped it. That is the permanently-red-gate shape: every PR in the
+# repo blocked on an unrelated doc, which trains people to click through.
+#
+# Composition at that point: 11 of 13 entries were real `scripts/` call sites;
+# all three failures came from the 2-entry `claudedocs/` half.
+#
+# What still guards prose, and why it is BETTER at it:
+# `test_no_tracked_shell_text_writes_a_kill_this_guard_would_deny` scans every
+# tracked file INCLUDING claudedocs, matches only text in real shell-command
+# shape, and runs each hit through `check_tmux_kill_shared_server` itself. So a
+# doc that actually spells a dangerous command is still caught — by the check
+# built for that — while a doc that merely discusses one is not. It already
+# allowlists `claudedocs/handoff-tmux-restore-chain.md` and carries its own
+# positive control.
+_LEDGER_SCOPE_EXCLUDES = ("claudedocs/",)
+
+
+def _in_ledger_scope(rel):
+    """Is `rel` executable text the ledger is meant to classify?"""
+    return not rel.startswith(_LEDGER_SCOPE_EXCLUDES)
+
+
 def _scan_kill_sites():
     mentions, argvs = {}, {}
     for rel in _tracked_files():
@@ -2597,7 +2623,7 @@ def _scan_kill_sites():
             body = (_REPO / rel).read_text()
         except (OSError, UnicodeDecodeError):
             continue
-        if _MENTION_RE.search(body):
+        if _in_ledger_scope(rel) and _MENTION_RE.search(body):
             mentions[rel] = body
         for m in _TMUX_ARGV_RE.finditer(body):
             argvs.setdefault(rel, []).append(" ".join(m.group(0).split()))
@@ -2680,6 +2706,11 @@ def test_no_tracked_shell_text_writes_a_kill_this_guard_would_deny():
     shell_text = re.compile(r"tmux(?:\s+-{1,2}[^\s'\"]+)*\s+kill-s[a-z-]*")
     quoting_is_the_point = {
         "claudedocs/handoff-tmux-restore-chain.md",
+        # Quotes the command while WRITING UP the incident where a handoff doc
+        # tripped this very guard. Caught here after the ledger was scoped to
+        # executable paths — which is the coverage argument working: prose that
+        # SPELLS a dangerous command is still caught, by the check built for it.
+        "claudedocs/handoff-mention-system-repos.md",
         "scripts/claude-hooks/bash-guard.py",
         "scripts/claude-hooks/guard_core.py",
         "scripts/claude-hooks/tests/test_guard_core.py",


ZacxDev:
This is the right fix and it supersedes what I was doing — I reached the same diagnosis independently and was about to add a **seventh** ledger row when I found this PR. Three things I measured that bear on landing it, one of which is my fault.

## 1. A fourth firing, live right now

Your description says prose red-lined `main` three times in one day. It is **four**: `origin/main` is red at this moment on `claudedocs/handoff-gate-speed-and-ci-signal.md`.

```
  added:   ['claudedocs/handoff-gate-speed-and-ci-signal.md']
  removed: []
```

Not mine, landed while I was fixing the previous one. The treadmill is still turning.

## 2. 🔴 This branch now CONFLICTS with `main`, and that is my doing

`#1556` merged (`b62d1bf1`) about an hour ago and added `claudedocs/handoff-cairn-oss-multi-instance.md` to **both** allowlists. `git merge origin/fix/kill-ledger-scope` into current `main` fails with conflicts in exactly the two hunks where I added rows you delete:

```
scripts/claude-hooks/tests/test_guard_core.py
  2526 <<<<<<< HEAD ... 2538 >>>>>>> origin/fix/kill-ledger-scope
  2738 <<<<<<< HEAD ... 2747 >>>>>>> origin/fix/kill-ledger-scope
```

Resolution is take-yours in both. Apologies — I was fixing a red I had caused and did not know this PR existed.

## 3. The merged tree needs MORE rows dropped than your diff drops

After resolving take-yours and running the guards on the merged tree, the census **inverts** — `added: []`, and now three rows are stale:

```
removed: ['claudedocs/handoff-cairn-oss-multi-instance.md',
          'claudedocs/handoff-mention-system-repos.md',
          'claudedocs/handoff-tmux-webapp.md']
```

Which follows from your own change: once the mention scan is scoped to executable text, **every** `claudedocs/` row is unreachable, and `main` accumulated three more of them after your branch was cut. Worth a rebase so the diff drops all of them rather than the two that existed when you wrote it.

⚠ **I am deliberately not handing you a verified edit for this.** I tried removing the `claudedocs/` rows programmatically, over-removed into `quoting_is_the_point`, and took the suite to `2 failed`. So treat the three names above as the census's output — which I did verify — and not as a recipe I got working.

## Why I think this is the right shape

I hit the same thing from three sides today: a fix that populated one of the two allowlists and left the suite red (three separate attempts, mine included); a doc that tripped the ledger **because it documented the ledger**; and then my own merged handoff becoming the offender. Scoping to executable text removes the whole class rather than the instance — and the ARGV scanner is the half that can catch a real call site anyway.

ZacxDev:
I was asked to rebase and merge this, started, and **stopped** — it has been superseded and merging it would replace a newer fix with an older one.

## `main` already carries this change, and is green

`#1561` (`c0bbd6d9`) landed the same idea more generally — a prefix predicate rather than row deletions:

```
scripts/claude-hooks/tests/test_guard_core.py (on origin/main)
  2624: _PROSE_ONLY_PREFIXES = ("claudedocs/",)
  2634:     return rel.startswith(_PROSE_ONLY_PREFIXES)
```

Measured just now at `origin/main` `b1abf6b1`, `__pycache__` cleared, `PYTHONDONTWRITEBYTECODE=1`:

```
1537 passed in 238.95s   (0 failed)
```

So the red this PR exists to fix is closed.

## What stopped me mid-rebase

`git rebase origin/main` conflicted in three hunks, and the HEAD side of the second one was **main's own version of this fix** — the `🔴 PROSE-ONLY TREES THE TWO KILL SCANNERS DO NOT READ` block, which cites six offending files in two hours and names `#1556` explicitly. Resolving that by taking this branch's side would have deleted the landed implementation.

The PR is `CONFLICTING`, and the conflict is *with the thing it wanted to achieve*.

## Recommendation

Close as superseded by `#1561`. Two pieces of this branch may still be worth salvaging as a follow-up if they are not already covered:

- its **`#1549`-specific rationale text**, which is better written than what landed — the "permanently-red gate trains people to click through" framing in particular;
- whether `_PROSE_ONLY_PREFIXES` should be a tuple of one forever, or whether the ARGV scanner should keep reading prose while only the *mention* census skips it. `#1561` scopes **both** scanners; this PR scoped the ledger. That difference is real and I have not checked which is intended.

It is your PR, so I have not closed it — and apologies for the conflict in the first place: `#1556` was mine, and it added rows to both allowlists an hour before `#1561` made them unnecessary.

ZacxDev:
Closing as **superseded by #1561** (`c0bbd6d9`), on the operator's instruction — not abandoned, and the branch is left in place.

#1561 landed the same idea at a wider scope: one `_is_prose_only()` predicate over
`_PROSE_ONLY_PREFIXES = ("claudedocs/",)` consulted by **both** kill scanners. This PR
scoped the *mention* half only and added a `quoting_is_the_point` row for the other —
which is the two-allowlists-over-one-hazard shape that kept `main` red on two tests for
one root cause. The single predicate is what actually ends the loop.

State at close: `CONFLICTING` against `main` in three hunks, and the HEAD side of one of
them **is** #1561's own text — so there is nothing to rebase toward.

Verified on `origin/main` before closing, so this is not being closed on the strength of a
commit message:

- 6 of 189 tracked `claudedocs/` files match the mention pattern today; **0** reach either
  scanner. 11 executable-tree files still scanned by the mention half, 7 by the argv half.
- `pytest -k kill` → 98 passed.
- `457a5dc7` landed a handoff doc carrying the trigger spelling 74 seconds after
  `c0bbd6d9`, and nothing has touched `test_guard_core.py` since — this PR's goal, met.
- The residual both PRs create (neither static scanner covers docs now) is carried by the
  runtime hook, exercised directly: `check_tmux_kill_shared_server` refuses both wide
  spellings and returns `None` for the literal-socket form; `bash-guard.py` is untouched.

Written up in #1522's rank 11. The claim `devrc-kill-ledger-scope-executables` is still
live and is worth releasing by whoever holds it.

ZacxDev:
SUPERSEDED BY #1561 — measured, not inferred. Recommend closing unmerged.

#1561 ("stop the kill scanners reading claudedocs/") merged 2026-09-12T03:16:37Z and fixes the same defect this PR was opened for. This PR meanwhile went `CONFLICTING/DIRTY`, and its three gates are `KILLED` (the gate pod died — not a code failure), so its red says nothing about its content either way.

## The defect is closed on main

Clean worktree at `origin/main`, the three tests this PR targets:

    3 passed, 1534 deselected

and the whole module on a tree with #1515 merged in: `1537 passed`.

## The two implementations, side by side

|  | this PR (#1549) | #1561 (merged) |
|---|---|---|
| predicate | `_in_ledger_scope()` | `_is_prose_only()` |
| consulted by | the mention/ledger scanner only | **both** scanners |
| dead `claudedocs/` entries removed | 2 | 11 |
| positive control that executable offenders are still caught | — | `test_the_kill_scanners_still_catch_an_executable_offender` |

#1561 is the stronger version of this change on every axis this PR argued for — including the one-predicate-one-place point this PR's own comment makes, which #1561 actually implements and this PR does not (it leaves the two scanners on separate lists).

## The one genuine difference, and why it does not block closing

They are not identical. This PR deliberately kept the **shell-text** scanner reading `claudedocs/`, so a doc that SPELLS a runnable wide kill would still be caught; #1561 exempts `claudedocs/` from both scanners, so it would not. That is a real reduction in scanner coverage and this PR's comment argues for it explicitly.

It does not matter here, because the scanner was never the gate. Measured against the live hook just now:

    'tmux kill-server'              -> deny
    'tmux kill-session -t scratch'  -> deny
    'echo hello'                    -> (no decision)   <- negative control, so "deny" means something

Both kills are denied unless a non-default `-L <socket>` is proven; the harmless command passes, so the probe is not wired to always-deny. `bash-guard.py` is untouched by both PRs. A doc under `claudedocs/` can therefore spell the command, and executing it is still refused at runtime — the layer that actually prevents the harm is intact, and the scanner difference costs a second-line warning, not the control.

## Recommendation

Close this unmerged. Nothing in it needs porting: its coverage delta is covered by the runtime hook, and its cleanup is a strict subset of #1561's.

NOT VERIFIED: I did not run the full 22k-test suite on either tree — I ran the guard-core module (1537 tests) and the `-k kill` subset (96 tests), which is where both PRs live and where the red was. I did not re-run #1561's own mutation battery; I read its diff on main and confirmed both scanners consult the one predicate (lines 2640, 2741, 2812).

⚠ A concurrent session holds the claim `ci-flakes-and-misattribution-11` ("verify it is closed by #1561 and record the closure in #1522's doc"). This comment is about THIS PR's disposition only; I have deliberately not touched #1522's doc.

PR #1551, opened 2026-09-12
fix(guard): the wide-kill guard read only the command line, so a kill inside an executed script walked past it

ZacxDev:
## What this closes

The wide-kill guard from #1415 was deployed and **correct**, and it did not stop the
same accident happening **twice more on 2026-09-11** — 52 and 53 live Claude
conversations. It gates Bash-tool **TEXT**; both times the command was in a file:

```
Write -> /tmp/.../repro.sh   containing  TMUX_TMPDIR=... ; tmux kill-server
Bash  -> bash /tmp/.../repro.sh
```

Control pair, run **before** writing anything: the inline spelling **denies**, the
identical bytes inside a file **allowed at rc 0**.

## Attribution — not "it happened around then"

| | crash 1 | crash 2 |
|---|---|---|
| script written | 00:05:42.8 | 02:36:32.8 |
| executed | 00:05:46.3 | 02:36:42.9 |
| **server dies** | **00:05:47** | **02:36:44** |
| unrelated sessions interrupted at that instant | **4** | **3** |

Those sessions were doing unrelated work in other repos. A per-command denial cannot
interrupt four unrelated sessions; a dying shared server can.

🔴 **The handoff had the times wrong** (00:08 / 02:23 — the second by 13 minutes),
which is why repeated searches concluded "nothing logged" and why this was recorded
as UNATTRIBUTED.

Also ruled out properly: `oom-kill:` is **0 across every boot** against a **working**
positive control (a contained 50 MB cgroup OOM does log on this host — the prior
"positive control of 54" does not reproduce); `segfault` fires 6× elsewhere; both
windows hold **zero** kernel entries.

## Why only the tmux check runs over file contents

The obvious design — run the whole policy over the file — was **measured and
rejected**:

| scope | population | denied |
|---|---|---|
| whole policy over file bodies | 103 top-level `scripts/` | **2** — `ship.sh`, `drift-check.sh` |
| this arm (tmux check only) | ~~135 scripts, 5 dirs~~ | ~~**0**~~ |
| ⚠ **that row was measured NON-RECURSIVELY and is RETRACTED** | 819 files, recursive | **5**, all Python — see round 1 below |
| this arm, after the round-1 interpreter fix | 819 files, recursive, both invocation shapes | **0** |

Both denials are false positives from prose. Denying `bash scripts/ship.sh` is exactly
the permanently-red gate `_IRREVERSIBLE_CHECKS` argues against. Widening this arm needs
the same two numbers.

## Two gaps, stated rather than papered over

- **`source x.sh` / `. x.sh` is NOT read** — even though sourcing runs in the *current*
  shell and so deserves it more. A different invariant outranks it here:
  `test_the_hot_path_reads_no_files_and_execs_no_git` pins that this hook must not open
  a file just because a command names one, and `. wt.env; ls` is the commonest shape on
  that path. Pinned by its own test so it cannot close silently.
- **a bare `x.sh` through `$PATH`** — resolving it means guessing a PATH this process
  does not share with the command; a wrong guess denies on an unrelated file.

## The repo's own invariants caught two things review did not

Both fixed, not worked around:

1. **The hot-path `open()` ledger.** My first version read every sourced file on every
   Bash call. That is why `source` is out of scope.
2. **A mutation survivor.** `p = path` — ignoring `cwd` entirely — **SURVIVED a green
   suite**, because every test wrote an absolute path. That is a docstring claiming
   coverage the tests did not provide. `bash repro.sh` after a `cd` is the shape agents
   actually type; now pinned in both directions.

## Verification

- **3145 passed, 1 failed** in `scripts/claude-hooks/tests/` on the tree with
  `origin/main` MERGED IN (round 3). The one failure is INHERITED —
  `test_clawgate_writeback_guard.py::test_the_skill_names_the_same_DEPLOYED_path_the_hook_prints`
  fails on `origin/main` alone, measured in a clean detached worktree of it.
- ~~The **2 failures are INHERITED**~~ — byte-identical on `origin/main` (the kill-mention
  ledger over two `claudedocs` handoffs). This branch adds **zero** new offenders; the
  added/removed sets match `main` exactly. #1543 / #1549 fix that.
- ~~**Mutation sweep 8/8 KILLED**~~ — 🔴 **RETRACTED.** A 13-mutant re-run found **SIX
  survivors**. Replaced by a 24-mutant sweep, **24/24 KILLED**, `CONTROL-GREEN`, a
  known-fatal positive control KILLED, under `PYTHONDONTWRITEBYTECODE=1` with
  `__pycache__` cleared between mutants and the tree restored byte-identically
  (`cmp`). Matrix in the round-1 comment.
- 11-case control matrix: every invocation shape denies; `kill-pane`, a caller-made
  `-L my-probe-$$` server, a missing file, an unexpanded variable, `ship.sh` and
  `drift-check.sh` all still allow.

## 🔴 Blast radius, said out loud

This adds the **sixteenth** check to a policy that runs on **every Bash call in every
Claude Code session on both hosts**, and the **second** check that reads the world.
Both facts are pinned by name in `test_guard_core.py` so neither can grow unremarked.
It is a behaviour change to the operator's primary tool and is vetoable on that basis.

⚠ **Overlaps `test_guard_core.py` with #1543 / #1549.** Disjoint files are not safety
and neither is a clean merge, so the merged tree wants a run before this lands —
whichever of us is second.

---

## 🔴 ROUND 1 — an adversarial audit found two 🔴, and three claims in this body were false

The audit read `09f07aa7`. Everything below was **re-measured**, not argued.

### F1 🔴 the arm added 0.2–1 s to EVERY Bash call, on commands this repo mandates

In-process `evaluate()` medians (21 runs each), workbench, same tree:

| command | `origin/main` | `09f07aa7` | fixed |
|---|---|---|---|
| `scripts/claim-work.sh --list` | 0.21 ms | **1195 ms** | **1.02 ms** |
| `scripts/drift-check.sh` | 0.18 ms | **820 ms** | **1.59 ms** |
| `bash scripts/gate.sh --tier both` | 0.36 ms | 50.0 ms | 0.71 ms |
| `./scripts/claude-hooks/guard_core.py` | 0.53 ms | 844 ms **(DENY)** | 0.70 ms (allow) |
| `ls -la` | 0.11 ms | 0.14 ms | 0.16 ms |

`claim-work.sh` is **mandated by the resume skill** and this hook fires on every
Bash call in every session on both hosts.

The cost was never the read (88 KiB ≈ 0.5 ms). It was `commands()` — 187 ms on
that body, paid **twice** (once by the check, once to find nested invocations) —
and then again for every path-like token in the script's **prose** that happened
to name a real file. Traced: one `scripts/claim-work.sh --list` read
`claude/RULES.md`, `claude/RULES-ARCHIVE.md` (106 KiB), `/etc/machine-id`,
`~/.gitconfig`, `scripts/task-spec-drafter/drafter.sh` and 20 more.

Three gates fix it, and they are different gates:

1. **A sentinel pre-gate implied by the matcher.** `_is_wide_kill_word` requires
   `tok.startswith(_TMUX_KILL_MANY_PREFIX)`, so every spelling this check can
   deny carries `kill-s`. `_SCRIPT_TEXT_CHECKS` is now a tuple of
   **(check, sentinel) PAIRS** — structural, because a pre-gate keyed to one
   check's sentinel becomes silently wrong the moment a second check joins.
   Pinned two-way by name; the implication is *proved* at the matcher, not
   asserted. Coverage loss: **zero**, and quote-split spellings
   (`tmux 'kill-'server`, `tmux kill\-server`) still deny because the body is
   also tested with quoting stripped. One residual — a line-continuation splice
   — is declared and pinned.
   ⚠ `str.translate` for that strip cost **20 ms** on a 216 KiB body; three
   `str.replace` passes cost 0.6 ms. The pre-gate must not become the cost it
   exists to avoid.
2. **A descent cap** (`_SCRIPT_DESCEND_MAX_BYTES = 8 KiB`). The body is still
   **checked** at any size up to the read cap; a nested `bash <other>` inside a
   body over 8 KiB is **not followed**. That one IS a coverage trade and is
   stated as one, with both arms pinned. The parser costs ~5 ms at 8 KiB, 34 ms
   at 16 KiB, 211 ms at 88 KiB.
3. **The interpreter rule** from F7 below, which also stops binaries being
   parsed.

A structural regression pin counts `commands()` calls rather than wall clock: a
big benign body must cost exactly ONE parse, with a positive control proving the
counter can move.

### F2 🔴 a comment claimed a test prevented F1. It did not.

`test_the_hot_path_reads_no_files_and_execs_no_git` and
`test_the_hot_path_gate_is_the_same_walk_it_guards` assert only
`not [o for o in opened if str(env) in o]` — that *the sourced env file* was not
opened. Neither pins a total. **Measured: both pass with this PR applied while
`evaluate('…/scripts/ship.sh --no-laptop')` opens `ship.sh`.** The block citing
them as the reason `source` is excluded was false, and it is now corrected in
place rather than quietly reworded.

The `source` exclusion is **kept**, on a reason that survives re-measurement and
is about the POPULATION, not about an existing invariant: `. wt.env; ls` is the
commonest shape on this hook's hot path and an env file is not a program. It is
now pinned by a test that actually constrains it —
`test_the_script_arm_opens_only_the_files_it_will_execute` asserts the exact SET
of paths opened for a realistic mix (`. env; ls` → **nothing**), with a positive
control so no assertion can pass by observing nothing.

### F3 🟡 real bypasses, all measured ALLOW with a file containing a kill

**Fixed** (each verified ALLOW before, DENY after): `busybox sh x.sh`,
`busybox ash x.sh`, `bash -O extglob x.sh`, `bash +O extglob x.sh`,
`bash +o histexpand x.sh`, `bash 2>/dev/null x.sh`, `bash >/tmp/o.log x.sh`,
`bash 2> /tmp/e.log x.sh`, and — newly caught rather than merely declared —
`bash < x.sh` and `bash -s < x.sh`, where the shell reads its PROGRAM from the
redirect.

**Declared and pinned as gaps** (a ledger, not a passing grade):
`cat x.sh | bash`, `tmux run-shell`, `tmux source-file`, `make -f`, `python3`,
`watch bash`, `xargs`, `find -exec`, `nix develop … -c bash`, `bash -c repro.sh`.

🔴 The deny message said *"Do not go looking for a spelling that slips past this
check"* while `cat x.sh | bash` works. That sentence is **deleted**; it now says
gaps exist, are listed in the source, and using one is the same accident.

### F4 🟡 the `seen` set swallowed a real kill, order-dependently

`seen` was global to the walk while `depth` is per-branch. Measured on
`A→B→C→X→K`: `bash A.sh; bash X.sh` → **ALLOW**, `bash X.sh; bash A.sh` →
**DENY**. Same files, verdict decided by order. Keyed on `(realpath, depth)`
now; both orders pinned.

### F6 🟡 "8/8 KILLED" was wrong — 6 of 13 mutants survived

The claim in the original body is **retracted**. Each survivor was a guard some
docstring here claimed to cover, and each now has a test that kills it:
the path-separator test, the `-c` early return, the value-flag skip, the
`*?$\``-rejection (whose old test was vacuous — `_read_script('$SCRIPT')` finds
no file either way), and `if not os.path.isfile(p)` — the **anti-hang guard**
for FIFOs, which was neither commented nor tested, and whose failure mode is
"every Bash call in every session blocks forever", not "this command is denied".
The replacement sweep and its matrix are in the round-1 comment.

### F7 🟡 the false-positive sweep walked the wrong population

It called `d.iterdir()` — **non-recursive** — over five directories, missing 671
files under `scripts/` alone. Recursively, the arm as written denied **FIVE**
real files: `scripts/claude-hooks/guard_core.py` (mode 0755, so
`./scripts/claude-hooks/guard_core.py` is a real invocation),
`scripts/opencode/lib/oc_permissions.py`, `scripts/tests/test_opencode_engine.py`,
`scripts/tests/test_opencode_config.py`, and a `__pycache__/*.pyc`. Every one is
**Python**.

Fixed by the principled rule: **the INVOCATION decides the interpreter.**
`bash x` parses as shell whatever the file says (anything else would be a
bypass); `./x` honours the shebang, and a file with no shebang is shell only if
it is executable at all. A NUL sniff keeps binaries out of the parser.

The sweep is now `rglob`, excludes `__pycache__`, sweeps **both** invocation
shapes, and has a floor of 600 (the old floor of 80 was satisfied by the walk
that missed 671 files). Honest figure, 2026-09-12: **819 files, 0 denied**
direct-execution; **86 shell programs, 0 denied** under `bash <file>`.

### F8 🟡 the escape hatch's advice was inverted

`_QUOTING_ESCAPE_HATCH` tells the caller to write the text to a FILE — backwards
when what is denied is RUNNING a file. This arm now appends its own tail saying
so, and pointing at the real remedy (`-L my-probe-$$`, or move the prose out of
the executable file). Ordering asserted, so the correction cannot precede the
advice it corrects.

### F10 🟢 nits

"one hop" → `_SCRIPT_RECURSION_LIMIT = 3` gives **three**; measured, hops 1–3
deny and hop 4 allows, and the boundary is now pinned on both sides.
`bash -- -weird.sh` was ALLOW (`--` did `continue` without forcing the next
token to be the operand) and is now DENY.

### Scope table, corrected

| scope | population | denied |
|---|---|---|
| whole policy over file bodies | 103 top-level `scripts/` | **2** — `ship.sh`, `drift-check.sh` |
| this arm, as first written | 135 files, 5 dirs, **non-recursive** | 0 — **wrong population** |
| this arm, as first written | **819 files, recursive** | **5**, all Python |
| this arm, after F7 | **819 files, recursive**, both invocation shapes | **0** |

---

## 🔴 ROUND 2 — the round-1 `seen` fix introduced two 🔴 of its own

A delta audit cleared the sentinel pre-gate (92,274 constructed spellings, 0
coverage losses) and the interpreter rule (19 shebang attacks), and then found
that the three-line edit which fixed the round-1 order bug had created:

1. **A NEW order-dependent ALLOW on the exact shape this PR exists to block.**
   The visited key omitted `direct` and was written BEFORE the read, so a path
   whose read returned `None` poisoned the set. With the incident body in a
   mode-0644 no-shebang file — what the Write tool produces —
   `./repro.sh || bash repro.sh` (the ordinary "not executable, fall back"
   idiom) **ALLOWED**, while `bash repro.sh || ./repro.sh` denied.
2. **An uncaught `ValueError: lstat: embedded null character in path` escaping
   `evaluate()`**, reachable from 19 tracked `icon-*.png` files with no crafted
   input. Failed CLOSED, so a confusing false deny rather than a bypass.

Both fixed; both pinned in both orders. Also in round 2: fan-out is now bounded
(two-level 30×30 fan-out **16,381 ms → 183 ms**) by a budget that **never gates
the check**; bundled short flags (`bash -eo pipefail x.sh` and four more) closed;
the line-continuation gap re-attributed to the tokeniser with its wrong remedy
deleted; the 8 KiB cap's stated cost corrected ~4×; the deny's remedy no longer
names a `#` comment as a cause that can fire.

🔴 **Two of the audit's own prescriptions did not survive re-measurement** and
are recorded as negatives: `direct` in the visited key is NOT an independent
second half (its mutant SURVIVED — it is defence in depth, labelled as such with
the invariant pinned), and two of my own mutants first SURVIVED because I had
constructed them badly, not because the tests were weak.

Full matrices in the round-2 comment.

---

## 🔴 ROUND 3 — the fan-out budget was itself an order-dependent ALLOW

A delta audit confirmed both round-2 🔴 as fixed and **wider than claimed** (an
exhaustive pairwise sweep found 0 order-dependent pairs out of 660, against 64
before), confirmed the merge lost nothing, and then found that the *budget* added
in round 2 had re-opened order-dependence in a new coordinate:

    bash b0.sh; …; bash b7.sh; bash last.sh   ALLOW  🔴   (kill two hops down)
    bash last.sh; bash b0.sh; …; bash b7.sh   DENY        same files, reversed

— while `guard_core.py` headed that paragraph *"THE BUDGET DELIBERATELY NEVER
GATES THE CHECK"* and the test was named `..._NEVER_gates_the_CHECK`. **Third
round running in which the fix for the previous round's order-dependence
re-opened it elsewhere, and the second in which the sentence written to justify
it denied the behaviour.**

**The walk is now breadth-first and spends its budget in a canonical order**
(smallest body first, then resolved path), so the verdict is a function of the
file SET, not of the order the command names them in. Measured over 40 orderings
of the audit's own fixture: **8 ALLOW / 32 DENY → 0 ALLOW / 40 DENY**. The
residual — a kill below a level whose budget ran out is still missed,
deterministically — is declared in the same paragraph rather than described as an
absence.

🔴 **The direction the audit offered was built and measured, and it does not
hold**: charging only sentinel-bearing bodies is not a bound at all (30×30
fan-out **187 ms → 17,855 ms**, worse than *no* budget), because almost nothing
bears the sentinel.

Also round 3: the value-flag cluster rule was **factually wrong** and its comment
certified it (`bash -eoO pipefail extglob x.sh` consumes two words and still runs
the script — verified against GNU bash 5.3.15); the round-2 `&`-redirect
"retraction" was itself false in two ways and is corrected in both; two docstrings
claimed coverage the suite does not have; and the fan-out timings were two runs
presented as one fact.

⚠ **Four guards now have SURVIVING mutants**, all subsumed by the level
ordering — `depth` and `direct` in the visited key, the last-level `break`, and
`break`-vs-`continue`. They are kept, labelled, and reported as survivors rather
than dressed up as covered; one test pins the ordering that makes them redundant.

Full matrices in the round-3 comment.

### DIFF
--- a/scripts/claude-hooks/guard_core.py
+++ b/scripts/claude-hooks/guard_core.py
@@ -2359,7 +2359,13 @@ _TMUX_KILL_DENY = (
     "own socket. "
     "🔴 AND IF A SHARED `kill-server` IS GENUINELY NEEDED, THE ESCAPE HATCH IS NOT YOURS: this "
     "hook gates the Bash TOOL, and the operator's own terminal is not hooked. Ask them to run "
-    "it. Do not go looking for a spelling that slips past this check."
+    # 🔴 THIS USED TO SAY "Do not go looking for a spelling that slips past this check", full
+    # stop — which read as a claim that none exists. It does: `cat x.sh | bash` reaches a shell
+    # through a pipe this parser does not model, and the blind-spot list in
+    # `check_tmux_kill_shared_server` names three more. An overclaiming deny message is worse
+    # than an honest one, because the first agent who finds a gap learns the guard lies.
+    "it. Gaps in this check exist and are listed in its source; using one is the same accident "
+    "as the incident above, not a workaround — the answer is still to ask the operator."
     + _QUOTING_ESCAPE_HATCH
 )
 
@@ -2508,42 +2514,130 @@ def check_tmux_kill_shared_server(cmd):
 # workflow, which `_IRREVERSIBLE_CHECKS` argues at length is worse than no gate
 # because it gets routed around AND reports safety.
 #
-# The narrow arm was measured on the same population: 135 real scripts across
-# `scripts/`, `githooks/`, `collector/`, `dl-router/` and `opencode/` — ZERO
-# denied — while the actual crash-1 body denies. That is the whole argument:
-# this arm is all upside on the tree as it stands. Widening it to another check
-# is an operator decision that must bring the same two numbers, not an
+# ⚠ THE "135 SCRIPTS, ZERO DENIED" FIGURE THAT USED TO SIT HERE WAS MEASURED ON
+# THE WRONG POPULATION AND WAS WRONG. The sweep walked five directories
+# NON-RECURSIVELY and missed 671 files under `scripts/` alone. Re-run
+# recursively, the arm as first written denied FIVE real files — `guard_core.py`
+# itself (mode 0755), `scripts/opencode/lib/oc_permissions.py`, two tests and a
+# `.pyc` — every one of them PYTHON, read as shell because the arm took any
+# path containing `/` as a script. Fixed by `_is_shell_program`: the INVOCATION
+# decides the interpreter.
+#
+# The honest figure, measured 2026-09-12 over a RECURSIVE walk of `scripts/` and
+# `githooks/`: 819 files, ZERO denied by direct execution, and ZERO denied by
+# `bash <file>` across the 86 of them that are shell programs — while the actual
+# crash-1 body denies. Widening this arm to another check is an operator
+# decision that must bring the same numbers from the same recursive walk, not an
 # assumption that more checks are more safety.
-_SCRIPT_TEXT_CHECKS = (check_tmux_kill_shared_server,)
+#
+# 🔴 EVERY ENTRY IS A (check, SENTINEL) PAIR, AND THE SENTINEL IS NOT OPTIONAL.
+# The sentinel is a substring that the check's OWN MATCHER makes NECESSARY for a
+# deny, and it is what keeps this arm off the hot path: a body that cannot
+# contain the sentinel cannot be denied, so it is never parsed. The pairing is
+# structural rather than a comment because a tuple of bare functions would make
+# the pre-gate silently WRONG the moment a second check joined — it would still
+# skip on the tmux sentinel while the new check's own spellings walked past.
+# `test_every_script_text_check_carries_a_sentinel_its_matcher_implies` fails if
+# an entry has no sentinel, and the implication itself is proved (not asserted)
+# by `test_the_tmux_sentinel_is_IMPLIED_by_the_matcher`.
+_SCRIPT_TEXT_CHECKS = (
+    # `_is_wide_kill_word` requires `tok.startswith(_TMUX_KILL_MANY_PREFIX)`, so
+    # EVERY spelling this check can deny carries that literal. Derived from the
+    # constant, never retyped, so the pre-gate cannot drift from the matcher.
+    (check_tmux_kill_shared_server, _TMUX_KILL_MANY_PREFIX),
+)
 
 # A script big enough to be a data file is not one an agent hand-wrote to run a
 # kill; the cap keeps a pathological read off the hot path of EVERY Bash call.
 _SCRIPT_READ_MAX_BYTES = 256 * 1024
 _SCRIPT_RECURSION_LIMIT = 3
 
+# 🔴 THE DESCENT CAP IS THE LATENCY FIX, AND IT IS A SEPARATE NUMBER FROM THE
+# READ CAP ON PURPOSE. MEASURED on the first shipped version of this arm, with
+# `evaluate()` timed in-process on the workbench:
+#
+#     scripts/claim-work.sh --list   960–1195 ms   (0.2 ms before this arm)
+#     scripts/drift-check.sh         552–820 ms
+#     bash scripts/gate.sh …          41–50 ms
+#
+# `claim-work.sh` is MANDATED by the resume skill, and this hook runs on every
+# Bash call in every session on both hosts. The cost is not the read (88 KiB is
+# ~0.5 ms); it is `commands()`, which took 187 ms on that body and was paid
+# TWICE — once by the check and once to find nested invocations — and then again
+# for every path-like token in its PROSE that happened to name a real file
+# (`claude/RULES.md`, `RULES-ARCHIVE.md`, `/etc/machine-id`, … all read).
+#
+# Two gates fix it, and they are different gates:
+#   * the SENTINEL pre-gate skips the check's parse when the body cannot be
+#     denied — no coverage loss, because the sentinel is implied by the matcher;
+#   * this cap skips the DESCENT parse for a body over 8 KiB. That one IS a
+#     coverage trade, stated plainly: a nested `bash <other>` inside a script
+#     larger than 8 KiB is not followed. The script itself is still read and
+#     still checked at any size up to the read cap. 8 KiB is where the parser
+#     costs ~5 ms; 16 KiB already costs ~34 ms and 88 KiB ~211 ms.
+# Agent-written scratch scripts — the population both 2026-09-11 incidents came
+# from — are a few hundred bytes, so multi-hop still works where it was needed.
+_SCRIPT_DESCEND_MAX_BYTES = 8 * 1024
+
 # 🔴 KNOWN GAP, AND A DELIBERATE ONE: `source x.sh` / `. x.sh` is NOT read.
 #
 # Sourcing runs in the CURRENT shell, so on danger alone it deserves reading
-# more than `bash <file>`, not less. It is excluded because of a DIFFERENT
-# invariant that outranks it here: `test_the_hot_path_reads_no_files_and_execs_
-# no_git` pins that this hook — which runs on EVERY Bash call on both hosts —
-# must not open files just because a command mentions one, and the dominant real
-# use of `.`/`source` is ENV FILES (`. wt.env; ls`), which the guard already
-# resolves through its own separately-gated path. Reading every sourced file
-# would put an open() on the hot path for the commonest shape there is.
+# more than `bash <file>`, not less.
 #
-# `bash <file>` is not that population: it is only the commands that actually
-# spawn a shell to RUN a script, which is both rare and exactly what the two
-# 2026-09-11 incidents did. So the arm covers the measured shape and declines
-# the one that would tax every call.
+# ⚠ THE REASON THIS BLOCK USED TO GIVE WAS FALSE, and the correction is the
+# point. It said `test_the_hot_path_reads_no_files_and_execs_no_git` "pins that
+# this hook must not open a file just because a command names one". That test
+# pins no such thing: it asserts only that THE SOURCED ENV FILE is not opened
+# for commands carrying no real `git … commit`, and it passed unchanged while
+# this very arm opened `ship.sh` for `evaluate('…/ship.sh --no-laptop')`. No
+# total open() count was pinned anywhere. Measured, not argued.
 #
-# What would close it without paying that cost: a cheap pre-gate that can tell
-# an env file from a program without reading it. There isn't one, so this is
-# recorded as open rather than papered over.
+# The reason that survives re-measurement is narrower, and it is about the
+# POPULATION rather than about an existing invariant: `. wt.env; ls` is the
+# commonest shape on this hook's hot path, an env file is not a program, and
+# reading every sourced file would put an open() on that shape on every Bash
+# call. `bash <file>` is not that population — it is only the commands that
+# spawn a shell to RUN a script, which is rare and is exactly what the two
+# 2026-09-11 incidents did.
+#
+# That claim is now PINNED by a test that actually constrains it:
+# `test_the_script_arm_opens_only_the_files_it_will_execute` asserts the exact
+# SET of paths opened for a realistic command mix — `. wt.env; ls` opens
+# nothing, `bash repro.sh` opens exactly `repro.sh` — with a positive control so
+# the assertion cannot pass by observing nothing.
 
 
-def _script_operand(argv):
-    """The path this argv would EXECUTE as a script file, or None.
+# Value-taking shell flags whose ARGUMENT is not the script. `-O`/`+O` and `+o`
+# were missing and each one was a live bypass: `bash -O extglob x.sh` returned
+# `extglob` as the script path (no such file → ALLOW), and `+`-prefixed tokens
+# were not recognised as flags at all, so `+O` itself was returned.
+_SHELL_VALUE_FLAGS = ("-o", "+o", "-O", "+O", "--rcfile", "--init-file")
+
+# A redirection token as this tokeniser hands it over: `>`, `2>`, `>>`, `&>`,
+# `<`, `<<EOF`, `2>&1`, `>/tmp/o.log`. Group 2 is the operator, group 3 whatever
+# was attached to it. Nothing that is not a redirection can match: an ordinary
+# operand would have to begin with digits followed by `<` or `>`.
+_REDIR_RE = re.compile(r"^(\d*)(<<-?|<&|>&|&>>?|<>|>>|>\||>|<)(.*)$")
+
+
+def _redirection(tok):
+    """`(kind, attached-target)` for a redirection token, else `(None, None)`.
+
+    `kind` is `"in"` only for a plain `<`, where the shell READS ITS PROGRAM
+    from the target — `bash < x.sh` and `bash -s < x.sh` execute `x.sh`, so the
+    redirect target IS the script. Every other operator redirects a stream the
+    shell does not interpret, so its target must be SKIPPED rather than read:
+    `bash 2>/dev/null x.sh` used to return `2>/dev/null` as the script path and
+    ALLOW the file that actually ran.
+    """
+    m = _REDIR_RE.match(tok)
+    if not m:
+        return (None, None)
+    return ("in" if m.group(2) == "<" else "other", m.group(3))
+
+
+def _script_operand(argv, _depth=0):
+    """`(path, executed_directly)` for the script this argv RUNS, or None.
 
     Two shapes, and deliberately no third:
       * `bash x.sh` / `sh -x x.sh`  — a shell with a FILE operand. `-c` is
@@ -2551,42 +2645,122 @@ def _script_operand(argv):
         this arm is for the case where the code is not in the command line.
       * `./x.sh` / `/abs/x.sh`      — executed directly.
 
-    Two gaps, both stated rather than papered over:
-      * `source`/`.` — see the block above; excluded to keep the hot-path
-        promise, not because it is safe.
+    🔴 THE SECOND ELEMENT IS LOAD-BEARING: THE INVOCATION DECIDES THE
+    INTERPRETER, NOT THE FILE. `bash x` means "read x as shell" whatever x
+    contains; `./x` means "run x the way its shebang says", which for a Python
+    file is not shell at all. Reading every directly-executed file as shell
+    denied FIVE real files in this repo — `scripts/claude-hooks/guard_core.py`
+    (mode 0755, so `./scripts/claude-hooks/guard_core.py` is a real invocation),
+    `scripts/opencode/lib/oc_permissions.py`, two tests and a `.pyc` — whose
+    bytes were never shell commands. See `_is_shell_program`.
+
+    KNOWN GAPS, stated rather than papered over, each pinned by a test in
+    section 16 of test_guard_core.py so it cannot close silently and then
+    reopen:
+      * `source`/`.` — see the block above.
       * a BARE `x.sh` resolved through `$PATH` — resolving it means guessing a
         PATH this process does not share with the command, and a wrong guess
         reads an unrelated file and denies on ITS contents.
+      * `cat x.sh | bash`, `tmux run-shell x.sh`, `tmux source-file x.sh`,
+        `make -f x.sh`, `python3 x.sh`, `watch bash x.sh`, `xargs … bash`,
+        `find -exec bash {} \\;`, `nix develop <dir> -c bash /abs/x.sh` — the
+        file reaches a shell through a program this parser does not model. Each
+        would need its own option-arity table, and getting one wrong is how a
+        check fails OPEN (see `_TMUX_VALUE_FLAGS`).
+      * `busybox busybox sh x.sh` — one applet hop is modelled, not two.
     """
     if not argv:
         return None
     base = os.path.basename(argv[0])
+    if base == "busybox":
+        # `busybox` is in `_SHELLS` because the binary can BE the shell, but the
+        # applet-dispatch spelling puts the real shell in argv[1] — and the flag
+        # scan below then returned `sh`, a path that never exists, so EVERY
+        # `busybox sh x.sh` was ALLOWED. That is the only way busybox is ever
+        # used as a shell here, so the missed spelling was the whole feature.
+        if _depth or len(argv) < 2:
+            return None
+        return _script_operand(argv[1:], _depth + 1)
     if base in _SHELLS:
-        skip_next = False
-        for tok in argv[1:]:
-            if skip_next:
-                skip_next = False
-                continue
+        i, force_operand = 1, False
+        while i < len(argv):
+            tok = argv[i]
+            if force_operand:
+                # After `--` the next token is the operand even when it looks
+                # like a flag: `bash -- -weird.sh` runs a file called
+                # `-weird.sh`. The old scan `continue`d past `--` and then
+                # skipped `-weird.sh` as a flag.
+                return (tok, False)
             if tok == "--":
+                force_operand = True
+                i += 1
                 continue
-            if tok.startswith("-"):
+            kind, attached = _redirection(tok)
+            if kind == "in":
+                if attached:
+                    return (attached, False)
+                return (argv[i + 1], False) if i + 1 < len(argv) else None
+            if kind == "other":
+                i += 1 if attached else 2
+                continue
+            if tok.startswith("-") or tok.startswith("+"):
                 # `-c` carries inline text, not a file; `_nested_shell_text` owns it.
-                if "c" in tok[1:] and not tok.startswith("--"):
+                if tok.startswith("-") and not tok.startswith("--") and "c" in tok[1:]:
                     return None
-                if tok in ("-o", "--rcfile", "--init-file"):
-                    skip_next = True
+                i += 2 if tok in _SHELL_VALUE_FLAGS else 1
                 continue
-            return tok
+            return (tok, False)
         return None
     # Direct execution. Require a path separator: a bare word here is either a
     # PATH lookup (see the docstring) or not a script at all.
     if "/" in argv[0]:
-        return argv[0]
+        return (argv[0], True)
     return None
 
 
-def _read_script(path, cwd):
-    """The file's text, or None when it is not a readable local regular file.
+def _is_shell_program(p, body):
+    """Would running `p` DIRECTLY execute `body` as shell commands?
+
+    Only asked of a directly-executed file (`./x`, `/abs/x`), never of a `bash
+    x` operand — there the invocation has already said "shell", and second-
+    guessing it from the file's own content would be a bypass: write a
+    `#!/usr/bin/env python3` line above a kill and run it with `bash`.
+
+    Three answers, in the order the kernel and the shell actually resolve them:
+      * a NUL byte near the start means a binary, not a text script;
+      * a `#!` line is authoritative — `execve` hands the file to THAT
+        interpreter, so a Python file is Python however it is named;
+      * no shebang at all means `execve` fails ENOEXEC and the calling shell
+        re-runs the file as a SHELL script — but only if it is executable in the
+        first place, because otherwise the invocation simply fails.
+    """
+    if "\x00" in body[:4096]:
+        return False
+    if body.startswith("#!"):
+        words = body.split("\n", 1)[0][2:].strip().split()
+        if not words:
+            return False
+        interp = os.path.basename(words[0])
+        if interp == "env":
+            # `#!/usr/bin/env -S VAR=1 bash` — the interpreter is the first word
+            # that is neither a flag nor an assignment.
+            interp = None
+            for w in words[1:]:
+                if w.startswith("-") or "=" in w:
+                    continue
+                interp = os.path.basename(w)
+                break
+            if interp is None:
+                return False
+        return interp in _SHELLS
+    try:
+        return bool(os.stat(p).st_mode & 0o111)
+    except OSError:
+        return False
+
+
+def _read_script(path, cwd, direct=False):
+    """The file's text, or None when it is not a shell program this can read.
 
     Never raises. A guard that throws on a weird path would fail the Bash call
     it was asked about, which is a worse outcome than the hole it is closing.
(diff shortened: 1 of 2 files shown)


ZacxDev:
Round-1 audit fixes. Every finding re-measured before being changed; the two 🔴
are fixed, the 🟡s are fixed or declared-and-pinned, and three claims in the
original PR body were false and are retracted in it.

```audit-claims round=1 audited=09f07aa7..5aedd14c
1. F1 🔴 latency — a sentinel pre-gate derived from `_TMUX_KILL_MANY_PREFIX` (never a retyped literal), an 8 KiB descent cap, and the F7 interpreter rule take `evaluate('scripts/claim-work.sh --list')` from 1195 ms to 1.02 ms and `scripts/drift-check.sh` from 820 ms to 1.59 ms (main: 0.21 / 0.18 ms).
2. F1 — `_SCRIPT_TEXT_CHECKS` is now a tuple of (check, sentinel) PAIRS pinned two-way by name, so a second check cannot join without its own sentinel; the sentinel's implication by `_is_wide_kill_word` is proved at the matcher, and quote-split spellings (`tmux 'kill-'server`, `kill\-server`) still deny.
3. F2 🔴 — the comment claiming `test_the_hot_path_reads_no_files_and_execs_no_git` pins "must not open a file just because a command names one" is corrected in place as FALSE (measured: it passes while `ship.sh` is opened); the `source` exclusion is KEPT on a population argument and is now pinned by `test_the_script_arm_opens_only_the_files_it_will_execute`, which asserts the exact SET of opened paths with a positive control.
4. F3 🟡 — `busybox sh|ash x.sh`, `bash -O|+O|+o <opt> x.sh`, `bash 2>/dev/null x.sh`, `bash >/log x.sh`, `bash 2> /log x.sh`, `bash < x.sh` and `bash -s < x.sh` were each measured ALLOW and now DENY.
5. F3 — ten further shapes (`cat x.sh | bash`, `tmux run-shell`, `tmux source-file`, `make -f`, `python3`, `watch bash`, `xargs`, `find -exec`, `nix develop -c bash`, `bash -c repro.sh`) are DECLARED gaps with reasons and pinned by `test_the_DECLARED_gaps_stay_declared`.
6. F3 — the deny message no longer claims no bypass exists; "Do not go looking for a spelling that slips past this check" is deleted and asserted absent.
7. F4 🟡 — the visited set is keyed on `(realpath, depth)`; `bash A.sh; bash X.sh` and `bash X.sh; bash A.sh` over an `A→B→C→X→K` chain both DENY (measured ALLOW/DENY before).
8. F6 🟡 — the "8/8 KILLED" claim is retracted in the PR body; the five surviving guards each have a test, including the FIFO anti-hang `isfile` guard (asserted by forbidding the open, not by timing a hang) and a non-vacuous replacement for the `*?$\`` rejection test.
9. F7 🟡 — the invocation now decides the interpreter (`bash x` parses regardless of shebang; `./x` honours it; no shebang ⇒ shell only if executable; NUL ⇒ binary), and the sweep is recursive over both invocation shapes with a floor of 600: 819 files, 0 denied, positive controls in both shapes.
10. F8 🟡 — the arm appends `_SCRIPT_ARM_REMEDY`, which says the inherited "write the text to a file" advice is backwards here and gives the real remedy; ordering asserted.
11. F10 🟢 — `_SCRIPT_RECURSION_LIMIT` documented and pinned as THREE hops (1–3 DENY, 4 ALLOW); `bash -- -weird.sh` was ALLOW and now DENIES.
12. NOT addressed: a line-continuation splice (`kill-\`+NL+`server`) hides the sentinel from the pre-gate — declared and pinned; `chmod +x f && ./f` on a file with no shebang — declared and pinned; the 8 KiB descent cap is a real coverage trade, both arms pinned.
```


### Mutation sweep — 24/24 KILLED, control GREEN

Selection is **explicit node ids** (every section-16 test plus seven named
hot-path/ledger tests), not a `-k` filter; `PYTHONDONTWRITEBYTECODE=1`,
`__pycache__` cleared between mutants, tree restored from a `cp -a` copy and
verified byte-identical with `cmp`. Scoring is against the CONTROL's own failure
set, because the control run is not vacuously green — see the inherited note.

|---|---|---|
| M0 control, no edit | **CONTROL-GREEN** | — |
| M1 positive control: disable `_is_wide_kill_word` | KILLED | 46 tests |
| M2 drop the path-separator test | KILLED | `test_a_bare_word_is_not_treated_as_a_script_path` |
| M3 drop the `-c` early return | KILLED | `test_bash_dash_c_names_a_COMMAND_not_a_file` |
| M4 drop the value-flag skip | KILLED | `test_a_value_flags_ARGUMENT_is_not_the_script` |
| M5 stop recognising `+`flags | KILLED | same + `…MEASURED_allow_now_deny[bash +O …]` |
| M6 drop the busybox applet hop | KILLED | `…MEASURED_allow_now_deny[busybox sh …]` |
| M7 treat `<` like any other redirect | KILLED | `…MEASURED_allow_now_deny[bash < …]` |
| M8 stop recognising redirections | KILLED | `…MEASURED_allow_now_deny[bash 2>/dev/null …]` |
| M9 restore the old `--` continue | KILLED | `test_a_dash_dash_operand_that_looks_like_a_flag_is_the_script` |
| M10 drop the `*?$\`` rejection | KILLED | `test_a_path_carrying_an_unexpanded_expansion_is_NEVER_read` |
| M11 drop the `isfile` anti-hang guard | KILLED | `test_a_FIFO_is_never_OPENED` |
| M12 drop the read size cap | KILLED | `test_an_oversized_file_is_not_read` |
| M13 `_is_shell_program` always True | KILLED | the PYTHON / no-shebang / sweep tests |
| M14 drop the NUL/binary sniff | KILLED¹ | `test_a_binary_is_never_parsed_as_shell` |
| M15 ignore the shebang | KILLED | `test_a_directly_executed_PYTHON_file_is_not_read_as_shell` |
| M16 no-shebang files always shell | KILLED | `test_a_no_shebang_file_executed_directly_is_still_read_as_shell` |
| M17 honour the shebang for `bash x` (the bypass) | KILLED | `test_a_directly_executed_PYTHON_file_is_not_read_as_shell` |
| M18 remove the sentinel pre-gate | KILLED | `test_the_arm_does_not_PARSE_a_body_that_cannot_be_denied` |
| M19 drop the quote-stripping half | KILLED | `test_the_pre_gate_survives_a_QUOTE_SPLIT_spelling` |
| M20 remove the descent cap | KILLED | `test_a_body_over_the_DESCENT_cap_is_still_checked_but_not_followed` |
| M21 key `seen` on the path alone | KILLED | `test_the_visited_set_does_not_swallow_a_kill_by_ORDER[order0]` |
| M22 recursion limit 3 → 1 | KILLED | `test_the_recursion_limit_is_THREE_hops_not_one[2,3]` |
| M23 drop `_SCRIPT_ARM_REMEDY` | KILLED | `test_the_script_arm_deny_carries_its_OWN_remedy` |
| M24 `_SCRIPT_TEXT_CHECKS` loses its sentinel | KILLED | `test_every_script_text_check_carries_a_sentinel_its_matcher_implies` |

¹ **M14 SURVIVED on the first run, and that was a fixture bug worth recording.**
The binary fixture was `b"\x7fELF\x00\x00\x00tmux kill-server\x00"`; with the NULs
glued to the word, the tokeniser produced `argv[0] == '\x7fELF\x00\x00\x00tmux'`,
so the guard would have allowed it **either way** — the assertion was UNREACHABLE
and the test passed for the wrong reason. The fixture now puts the kill on its own
line and carries a control asserting those bytes DO deny when read as shell.

### Test RED/GREEN matrix

New tests run against the PRE-FIX `guard_core.py` (09f07aa7) with the new test
file: **25 failed, 1584 passed**. Against HEAD: **3104 passed, 2 failed** over
all of `scripts/claude-hooks/tests/`.

The 23 that go RED→GREEN: the recursive sweep, the two interpreter tests, the
sentinel ledger, the parse-count pin, the descent cap, the open() ledger, ten
`MEASURED_allow_now_deny` shapes, the `--` operand, both deny-message tests, the
`ORDER` case, the bare-word mutant test and the value-flag mutant test.

⚠ Honest labelling: **`test_the_tmux_sentinel_is_IMPLIED_by_the_matcher`,
`test_the_pre_gate_survives_a_QUOTE_SPLIT_spelling`, the declared-gap ledger, the
hop-boundary test and three of the five mutant tests are GREEN at 09f07aa7 too.**
They are INVARIANT GUARDS and mutation-kill tests, not regression coverage, and
are not counted as the latter.

### The 2 failures are still INHERITED, unchanged

`test_every_kill_server_call_site_in_the_repo_is_classified` and
`test_no_tracked_shell_text_writes_a_kill_this_guard_would_deny` fail on
`claudedocs/handoff-mention-system-repos.md` and
`claudedocs/handoff-tmux-webapp.md` — neither touched by this branch, and the
offender list is byte-identical to the pre-fix run above. This round adds **zero**
new offenders (`guard_core.py` and `test_guard_core.py` are both already in
`quoting_is_the_point`).

ZacxDev:
Round-2 delta audit fixes. Both 🔴 were the same three-line edit that fixed the
round-1 order bug; both reproduced here before anything changed. Two of the
audit's own prescriptions did not survive re-measurement and are recorded as
negatives rather than quietly followed.

```audit-claims round=2 audited=5aedd14c..133a79cd
1. 🔴-1 seen-poisoning — the visited set is now written only AFTER a successful read, so a path whose read returns None never enters it; `./repro.sh || bash repro.sh`, `./repro.sh 2>/dev/null || bash repro.sh`, `./repro.sh; bash repro.sh` and `./repro.sh && bash repro.sh` were each measured ALLOW at 5aedd14c with the incident body in a mode-0644 no-shebang file and all DENY now, in both orders, plus the one-level-down wrapper case.
2. 🔴-1 HONEST NEGATIVE — the audit's two "independent halves" are NOT independent: a mutation sweep scored "visited key omits `direct`" SURVIVED, because a failed read never enters `seen` and a successful read returns the same bytes either way. `direct` is KEPT as defence in depth and LABELLED as such, not counted as covered; the invariant that makes it redundant is pinned by `test_the_direct_element_of_the_visited_key_is_DEFENCE_not_a_fix`.
3. 🔴-2 NUL crash — `os.path.realpath` now runs only on a path `_read_script` has already opened; `evaluate("bash …/icon-48.png")` and `evaluate("bash a\0b.sh")` raised `ValueError: lstat: embedded null character in path` out of `evaluate()` at 5aedd14c and both return None now. Pinned with a non-vacuity test over the 19 tracked `icon-*.png` files, separately from the repo sweep whose `shellish` restriction excluded them.
4. 🟡-3 fan-out — `_SCRIPT_DESCEND_BUDGET_BYTES` (64 KiB) bounds total parsed bytes per `evaluate()`: 1→100×8 KiB 1,610→133 ms, two-level 30×30 (900 distinct leaves) 16,381→183 ms, 300 inline 1,559→337 ms. The docstring's "bounded three ways" claim is replaced by an enumeration of what is actually bounded.
5. 🟡-3 soundness — the budget NEVER gates the check; every file reached is still read and sentinel-tested. Pinned by a test whose killer file is deliberately LARGER than the leftover budget, because a small one let the "gate the check too" mutant survive.
6. 🟡-4 splice attribution — corrected to a tokeniser/matcher gap that predates the pre-gate, with the control asserted (`check_tmux_kill_shared_server` called directly on the spliced text still returns None). The old prescribed remedy ("normalise continuations before the substring test") is DELETED because it would leave the check running and still allowing.
7. 🟡-5 bundled flags — the rule is now bash's own (a value letter consumes the next word only when LAST in the cluster): `bash -eo pipefail x.sh`, `bash -euo pipefail x.sh`, `bash -xo pipefail x.sh`, `sh -eo pipefail x.sh`, `bash -eO extglob x.sh` all measured ALLOW at 5aedd14c and DENY now, with `-eu`/`-x` asserted NOT to eat the operand.
8. 🟡-5 `&` redirects — `bash 2>&1 x.sh`, `bash 3>&1 x.sh`, `bash &>/log x.sh` are DECLARED gaps with the mechanism asserted (`split_commands` cuts on `&` one layer below this parser); `_redirection`'s false claim to handle `2>&1` is retracted, and the direct-execution branch no longer accepts the `&`-split residue `>/tmp/o` as a script path.
9. 🟡-6 cap cost — re-measured at exactly 8 KiB: echo 7.1 ms, heredocs 21.2 ms, here-strings 39.2 ms, real `test_release_wrapper.sh` 23.5 ms; the "~5 ms" figure the cap was argued from is corrected. `test_the_arm_does_not_PARSE_a_body_that_cannot_be_denied` gained an UNDER-cap case (exactly 2 parses), since its old fixture was over the cap and its assertion was guaranteed by the cap alone.
10. 🟢-7 deny remedy — (b) no longer names the `#` comment as a cause: measured, a comment ALLOWS and a heredoc BODY denies, and the message now says which is which.
11. Merged `origin/main` (63 commits) into the branch: no textual conflict, both sides' edits to `test_guard_core.py` verified present in the merged result, and the MERGED tree runs 3135 passed / 0 failed — main's own `_KILL_MENTION_LEDGER` fixes resolved the two failures this branch had been carrying.
12. NOT fixed, declared and pinned: the `&`-redirect family (needs a change to the shared segment splitter); the line-continuation splice (needs `_tokenise` to splice continuations); the 8 KiB per-body descent cap and the 64 KiB fan-out budget, both real coverage trades with both arms pinned; `chmod +x f && ./f` on a no-shebang file; `cat x.sh | bash` and the other nine declared invocation gaps.
```

### Mutation sweep — control GREEN, 30/31 KILLED, 1 SURVIVED-and-declared

`PYTHONDONTWRITEBYTECODE=1`, `__pycache__` cleared between mutants, restored from
a `cp -a` copy and verified byte-identical with `cmp`, explicit node ids (not a
`-k` filter), scored against the control's own failure set.

Round-1's 24 mutants were re-run against the round-2 code and all still die.
New this round:

|---|---|---|
| M0 control, no edit | **CONTROL-GREEN** | — |
| M1 positive control | KILLED | 46 tests |
| N1 visited key omits `direct` | **SURVIVED** | — see claim 2; declared redundant, with the invariant pinned |
| N2 record visited BEFORE the read | KILLED | `test_a_NUL_bearing_path_does_not_raise_out_of_evaluate`, `…RELATIVE_script_path…` |
| N3 remove the fan-out budget | KILLED | `test_the_walk_bounds_TOTAL_parsed_bytes_not_just_one_body` |
| N4 let the budget gate the CHECK | KILLED¹ | `test_the_fan_out_budget_NEVER_gates_the_CHECK` |
| N5 drop the redirection test in the direct branch | KILLED | `test_a_redirection_token_is_not_a_directly_executed_script` |
| N6 drop the `#` sentence from the remedy | KILLED¹ | `test_a_COMMENT_only_script_is_allowed_and_the_deny_says_so` |
| M4b regress to exact-spelling value flags | KILLED | `test_a_value_flags_ARGUMENT_is_not_the_script` |
| M4c drop the LONG value-flag skip | KILLED | same |

¹ **N4 and N6 SURVIVED their first run and both were MY mutant-construction
errors, not test gaps** — the exact "isolate the mutation" trap. N4's condition
(`budget > 0`) could never be false at the sizes the fixture used; N6 deleted
only the tail of the sentence the assertion reads. Re-spelled to the narrowest
expression that can be wrong, both die. N4's fixture also needed a killer file
LARGER than the leftover budget, which is now asserted in the test.

⚠ Four round-1 mutants first reported `PATTERN-MISS` because this round's edits
re-spelled the lines they patch. Re-spelled and re-run: M2, M3, M20, M21 all
KILLED. A `PATTERN-MISS` is not a pass, and is reported rather than dropped.

### Test RED/GREEN matrix

New/updated tests against the pre-fix `guard_core.py` (`5aedd14c`), with this
round's test file: **17 failed, 1609 passed**. On the merged tree: **3135 passed,
0 failed** across all of `scripts/claude-hooks/tests/`.

RED→GREEN (14): the four `SKIPPED_read` poisoning cases, the one-level-down
wrapper, the read-then-key ordering pin, both NUL cases, the icon non-vacuity
test, the fan-out budget bound, the budget-never-gates-the-check case, the
redirection-token case, the comment/heredoc remedy case, and the extended
value-flag test.

⚠ Honest labelling, two ways. **Green at `5aedd14c` too**, so invariant guards
rather than regression coverage: the three `&`-redirect declared-gap rows, the
splice attribution test, the under-cap parse-count assertion, the
`bash repro.sh || ./repro.sh` order, and the `direct`-is-defence invariant.
**And one RED in that run was a harness artifact, not a finding**:
`test_a_directly_executed_PYTHON_file_is_not_read_as_shell` failed only because
my baseline harness used `shutil.copy2`, which copied the 0644 mode of the
`git show` output onto `guard_core.py`, tripping that test's own `os.access(…,
X_OK)` fixture check. It passes at `5aedd14c` in isolation; the harness now uses
`shutil.copy`.

ZacxDev:
Round-3 delta audit fixes. NEW-1 reproduced exactly before anything was changed.
The direction the round-3 brief offered was **built and measured, and it does not
hold** — the fallback honesty repair is done too, and the real fix is a
restructure. Four guards now have SURVIVING mutants; all four are reported as
survivors rather than dressed up as covered.

```audit-claims round=3 audited=133a79cd..d57cbd39
1. 🔴 NEW-1 reproduced verbatim at 133a79cd (`k.sh` 17 B behind `last.sh` 3 KB behind eight 8,190 B bodies): bulk-then-killer ALLOW, killer-then-bulk DENY, `&&`-chain-bulk-first ALLOW. All DENY now, and my own sweep over 40 orderings of that fixture (every rotation, each reversed, both `;` and `&&`) went from 8 ALLOW / 32 DENY to 0 ALLOW / 40 DENY.
2. 🔴 NEW-1 fix — the walk is BREADTH-FIRST and spends its budget in a canonical order (smallest body first, then resolved path). Two separate claims, stated separately: no file the walk REACHES is ever check-gated, and which files are reached below level 1 is a function of the file SET rather than of the command's order.
3. 🔴 NEW-1 honesty repair, done as well as the fix — the headline "THE BUDGET DELIBERATELY NEVER GATES THE CHECK" and the test name `..._NEVER_gates_the_CHECK` were both false as written; renamed to `test_no_file_the_walk_REACHES_is_ever_check_gated`, and the residual (a kill below a level whose budget ran out is still missed, deterministically) is DECLARED in the same paragraph rather than described as an absence.
4. 🔴 NEW-1 direction REJECTED ON MEASUREMENT, not on preference: charging only sentinel-bearing bodies is not a bound at all — built it, and the two-level 30×30 fan-out went from 187 ms back to 17,855 ms, WORSE than the 16,381 ms with no budget whatsoever, because almost nothing bears the sentinel. Recorded in the source so it is not re-proposed.
5. 🔴 NEW-1 pinned by `test_bulk_ahead_of_a_killer_parent_does_not_hide_it` (4 shapes, the multi-file TOP-LEVEL form, not the depth-1 case) and `test_the_verdict_does_not_depend_on_COMMAND_ORDER` (every rotation + reverse). Both RED at 133a79cd.
6. 🟡 NEW-5 fixed — the level loop breaks before charging at the last level, so `_SCRIPT_DESCEND_BUDGET_BYTES`'s "total bytes handed to the parser" is true again. ⚠ HONEST ATTRIBUTION: what makes the regression test green is the breadth-first walk, NOT the break — removing the break SURVIVES mutation, because a level-ordered walk has no level after the last.
7. 🟡 NEW-2 fixed — one skipped word per value letter, wherever it sits in the cluster. Verified against GNU bash 5.3.15 myself: `-oe`, `-uoe`, `-ox`, `-Oe`, `-eoO` (two letters, two words) and `sh -oe` all RUN `x.sh`. Last-letter missed 5 of 16 shapes; counting misses 0. The FALSE-DENY direction is pinned too (a real `pipefail` file in cwd must not be read as the script).
8. 🟡 NEW-2 ⚠ measuring it is itself a trap, recorded in the source: a first sweep passed the flags through an unquoted zsh variable, which does not word-split, so bash got `-oe pipefail` as ONE argument, errored, and every row read as "bash refuses this shape" — the opposite conclusion, stated with measurements.
9. 🟡 NEW-3 fixed — the two docstrings claiming the `direct` half is asserted "structurally" now say what is true and point at the test that says so.
10. 🟡 NEW-4 fixed, both halves: the stale `2>&1` listing was in the MODULE COMMENT (untouched by round 2, hence the 17-line contradiction) and is corrected there; and "NEVER reach this function" is retracted — only the UNQUOTED spelling is cut by `split_commands`. `bash '2>&1' x.sh`, `bash "2>&1" x.sh` and the escaped form all DENY and are now pinned in the DENY direction, so the `<&|>&|&>>?` alternatives cannot be deleted as dead code.
11. 🟢 NEW-6 fixed — the three fan-out timings were two different runs presented as one fact; unified to one run (1,610 / 16,381 / 1,559 ms) with the qualifier stated. 🟢 NEW-7 fixed — the budget test's 120/12/10 disagreement is gone; the count is derived and named once.
12. ⚠ FOUR SURVIVING MUTANTS, ALL SUBSUMED BY THE LEVEL ORDERING, none claimed as covered: `depth` in the visited key (round 1), `direct` in the visited key (round 2), the `break` before the last level (round 3), and `break`-vs-`continue` on the first misfit (provably inert given the ascending sort). All four are KEPT and LABELLED; `test_the_walk_is_BREADTH_first_which_is_what_makes_three_guards_redundant` pins the ordering itself and goes red first if it is lost.
13. Merged `origin/main` again (49 further commits): no textual conflict, both sides verified present. Merged tree runs 3145 passed / 1 failed, and that one failure is INHERITED — `test_clawgate_writeback_guard.py::test_the_skill_names_the_same_DEPLOYED_path_the_hook_prints` fails on `origin/main` alone, measured in a clean detached worktree of `origin/main`.
```

### Measured before → after

```
                                          133a79cd        d57cbd39
bash last.sh                              DENY            DENY
bash b0..b7; bash last.sh                 ALLOW  🔴        DENY
bash last.sh; bash b0..b7                 DENY            DENY
b0 && … && b7 && last.sh                  ALLOW  🔴        DENY
40 orderings of the same clauses          8 ALLOW/32 DENY  0 ALLOW/40 DENY

bash -oe  pipefail x.sh  (runs x.sh)      ALLOW           DENY
bash -uoe pipefail x.sh                   ALLOW           DENY
bash -ox  pipefail x.sh                   ALLOW           DENY
bash -Oe  extglob  x.sh                   ALLOW           DENY
bash -eoO pipefail extglob x.sh           ALLOW           DENY
sh   -oe  pipefail x.sh                   ALLOW           DENY
bash -oe pipefail <benign>  (false deny)  DENY   🔴        ALLOW

bash '2>&1' x.sh / "2>&1" / escaped       DENY            DENY   (pinned now)
bash 2>&1 x.sh   (unquoted)               ALLOW           ALLOW  (declared gap)

fan-out, one evaluate():
  1 -> 100 helpers x 8 KiB                1,610 ms        176 ms
  two-level 30x30, 900 distinct leaves   16,381 ms        192 ms
  300 scripts on one command line         1,559 ms        322 ms
  120 benign then a KILL                  2,828 ms         11 ms  (DENY both)
sentinel-charging variant (rejected):    30x30 = 17,855 ms — worse than no budget

per-call latency, unchanged:
  scripts/claim-work.sh --list            1.07 ms         1.02 ms
  scripts/drift-check.sh                  1.76 ms         1.58 ms
  ls -la                                  0.16 ms         0.16 ms
```

### Mutation sweep — control GREEN, 31/35 KILLED, 4 SURVIVED-and-declared

`PYTHONDONTWRITEBYTECODE=1`, `__pycache__` cleared between mutants, restored from
a `cp -a` copy and verified byte-identical with `cmp`, explicit node ids (not a
`-k` filter), scored against the control's own failure set.

New this round: **P1** (spend in traversal order — the round-3 bug) KILLED by
`test_bulk_ahead_of_a_killer_parent_does_not_hide_it` and
`test_the_verdict_does_not_depend_on_COMMAND_ORDER`; **P2** (sort by path only,
losing smallest-first) KILLED by all three bulk shapes; **M4d** (regress to the
round-2 last-letter rule) KILLED by the new bash-cluster test; **N3** (remove the
budget) KILLED; **M20**, **N2**, **N4**, **N6** re-spelled for the restructure
and all KILLED.

⚠ Four mutants SURVIVED and are listed in claim 12. Three round-2 mutants first
reported `PATTERN-MISS` because this round re-spelled the lines they patch;
re-spelled and re-run, all KILLED. A `PATTERN-MISS` is not a pass.

### Test RED/GREEN matrix

New/updated tests against `133a79cd`: **4 failed, 1634 passed** —
`test_bulk_ahead_of_a_killer_parent_does_not_hide_it[bulk-then-killer]` and
`[and-chain-bulk-first]`, `test_the_verdict_does_not_depend_on_COMMAND_ORDER`,
and `test_every_bash_value_flag_CLUSTER_shape_is_measured_against_real_bash`.
On the merged tree: **3145 passed, 1 failed (inherited, see claim 13)**.

⚠ `test_a_body_at_the_recursion_limit_is_not_CHARGED_for_a_parse_that_cannot_happen`
needed its fixture rebuilt to go red at all: a first version used 9 chains of
~8 KiB and was GREEN at `133a79cd`, because the charge is all-or-nothing and the
big debits were REFUSED once the budget dropped below 8 KiB. Many small phantom
charges are what actually drain it. Reported because a test that cannot go red
is the thing this ladder keeps finding.

⚠ The harness artifact from round 2 is fixed properly: `shutil.copy` also copies
the mode, so the baseline run still produced one bogus RED
(`test_a_directly_executed_PYTHON_file_is_not_read_as_shell`, which asserts
`guard_core.py` is executable). Now `shutil.copyfile`; that row is gone from
this round's matrix.

PR #1553, opened 2026-09-12
test(cairn): guard the operator-surface PROBE against the code it routes off

ZacxDev:
## What this closes

`claude/skills/cairn/reference/operator-surface.md` routes an operator off a
**probe**: run `cairn create` against the scope, and read the answer.

| answer | remedy |
|---|---|
| **201** | done — that WAS the fix |
| **rc 6 `[not-found]`** | edit the secret, replace the pod, re-run |

That block is payload a human executes against a live k8s store, and a round-4
audit found it **guarded by nothing** — `git grep -l operator-surface -- scripts/`
returned no test. It has already been wrong **twice, in opposite directions**:

* the pre-#1504 draft made the allowlist edit the remedy in BOTH readings (a safe
  superset, merely wasteful);
* the round-3 draft (`cd5b23a2`) replaced it with a **classification** table whose
  `ABSENT` row prescribed `cairn create` alone, on a precondition ("allowlisted
  already") that neither of its own discriminators can establish — wrong *and*
  stuck, for the exact two scopes the page names as live.

## What the guard pins — state, not wording

One new file, `scripts/tests/test_cairn_operator_surface_probe.py`. **No other
file is touched.** Nothing here asserts a sentence: the doc's table is parsed as
a **routing object**, and every answer it quotes is cross-checked against what
the code actually produces.

1. **The success code is read out of the code, never typed in the test.** AST over
   `server.py::_create_entry` collects every `self._respond(<int>, …)`; the doc's
   success answer must be a 2xx that handler can send. Today that is `{201}`.
2. **The refusal pair is COMPUTED, not quoted.** `_token_codes` derives
   `X-Store-Status` token → the HTTP code(s) `server.py` pairs it with, and the
   doc's `[token]` is fed through the **real** `scripts/cairn::_classify` on those
   codes. The doc must equal the `(exit_code, status)` that comes back. This is
   the **seam** — `server.py` and `scripts/cairn` are each well tested alone, and
   `rc 6 [not-found]` is a claim about the two of them *together*.
3. **The mechanism the whole procedure rests on.** `create_entry` must still do
   `path.parent.mkdir(...)` — it is the only reason an allowlisted scope with no
   directory answers 201 instead of 404.
4. **"the allowlist arm is the ONLY 404"**, pinned structurally: zero direct 404
   responses in `_create_entry`, exactly one `_not_found` arm, and the `if`
   reaching it tests a name bound from `visible_scope_set`. Pinning merely *that a
   404 exists* would pass if the arm were rewired to any other condition — the
   "different shape" a spelled guard walks past.
5. **Citation integrity.** The doc cites `test_a_scopes_FIRST_entry_creates_the_directory`
   as its evidence; that test must exist and its source must mention the success code.

Reword the prose freely. Move any of those facts and this goes red.

## Red / green matrix

Runner: `nix develop ~/workspace/devrc -c python3 -m pytest <file> -q -p no:cacheprovider`
— **dev-host tier, not the `nix build` sandbox**. Base sha `9a4411c8`.

| tree | result |
|---|---|
| HEAD (this branch) | **19 passed** |
| doc content @ `cd5b23a2` — the round-3 *classification* table | **16 failed, 3 passed** |
| doc content @ `3161851b^` — pre-#1504, the section did not exist | **17 failed, 2 passed** |

Both reds land on the two load-bearing tests with their own messages —
`test_the_probe_table_was_actually_found` and
`test_the_probe_table_agrees_with_the_code_it_routes_off`
(*"…has no `| answer | meaning | remedy |` table — the operator has no probe to
route off"*). The pre-#1504 red additionally kills the citation test. The
remaining reds are the mutation batteries reporting **a broken harness, not a
clean result** — their anchors are absent from the historical doc, and they say
so rather than scoring survivors against a doc they cannot mutate.

## Mutation results — watched to fail, on BOTH sides

Three live source mutations. Each was applied after committing, then restored and
verified `cmp` byte-identical:

| mutation | result |
|---|---|
| doc: `rc 6 `[not-found]`` → `rc 7` | **KILLED** — *"the doc quotes `rc 7 [not-found]`, but `scripts/cairn` produces [(6, 'not-found')] for that token on the HTTP code(s) `server.py` pairs it with ([404])"* |
| `server.py`: delete `create_entry`'s `path.parent.mkdir(...)` | **KILLED** — *"…no longer creates the entry's parent directory, so an allowlisted scope with no directory cannot answer 201 — the doc's success row, and the whole probe, are false"* |
| `scripts/cairn`: add `"not-found": EXIT_WRITE_PRECONDITION` to `_WRITE_STATUS_TOKEN_EXITS` | **KILLED by 4 tests** — that table is consulted **before** the code table, so the mutation silently changes the rc an operator sees while every isolated test stays green. Exactly the seam case. |

Plus **12 permanent in-suite mutants**, each asserting THIS guard's own finding
(the needle is checked, not merely a non-empty list):

* 5 doc mutations — rc drift, token drift, success-code drift, an unreadable
  answer cell, reverting to a classification table;
* 7 code drifts — the server stops sending 201, `create_entry` stops making the
  directory, a second 404 arm appears, the 404 arm stops being the allowlist, the
  404 becomes reachable from two arms, the server renames the wire token, and the
  token starts riding on a second HTTP code the client maps differently.

Driving **both** sides is deliberate: a relationship guard that can only ever fire
on one side is the "docstring claims a relationship, body inspects one side"
defect this subsystem produced six of in one session.

## Positive / negative controls

* `test_the_probe_table_was_actually_found` — a zero from a scan that walked
  nothing would make every assertion below it vacuous.
* `test_the_code_facts_were_actually_read` — the same for the **other** side; an
  empty AST read is not agreement.
* `test_an_unmutated_doc_is_the_batterys_negative_control` — unmutated input must
  report **exactly zero**, or a `disagreements()` that flagged everything would
  score every mutant as killed while checking nothing.
* Each battery asserts its own anchor is present in the file **before** mutating it.
* A table row the parser cannot classify is **reported**, never skipped.

## A false-red found and fixed mid-review (second commit)

The first cut classified **every** refusal row as if it were the create path's
404. That would have made this a permanently-red gate: adding a correct third row
(`rc 9 [already-exists]`, which `_entry_exists` really does answer at 412) would
have failed the guard against a doc that had just got **better**.

`_token_codes` now classifies each row on **its own** token's codes.
`test_a_CORRECT_extra_refusal_row_does_not_go_red` is the control that pins it,
and two of the seven code drifts exist only because of the new map.

## Deliberately NOT built here

A second retracted-claim needle scan. That class is already owned repo-wide by
`test_subsystem_store_api.py::_unmarked_retractions`, whose site list is derived
from `git ls-files` — one rule, one place.

## Collision surface

**Zero file overlap** with #1508 (`chore/cairn-consolidate-onto-pin`, 48 files) or
#1516 (`claude/skills/cairn/SKILL.md`). This PR adds exactly one new file and
edits nothing else.

⚠ But **disjoint files are not safety**: #1508 touches
`claude/skills/cairn/reference/operator-surface.md`, `scripts/cairn`,
`scripts/subsystem-store-api/server.py` and `scripts/tests/test_subsystem_store_api.py`
— all four of which this guard **reads**. So the merged tree was tested, not
reasoned about: an integration branch off this PR with #1508's head merged in has
**no textual conflict**, and this guard is **19 passed** on that tree. ⚠ One
sample, against one head, and #1508 is in flight — re-run if either branch moves.

## Not verified

* Not run: either `nix build` sandbox tier, or a full `gate.sh` run. Every claim
  above is scoped to the dev-host tier, the named files, base `9a4411c8`.
* Nothing was deployed, shipped, or `home-manager switch`ed. The Kubernetes
  secret, the `subsystem-store` pod and `seed.sh` were **not touched** — that half
  is the operator's.
* This guard checks the probe table's **answers** against the code that produces
  them, plus the structural facts that make those answers true. It does **not**
  check the remedy prose, the rest of the page, or what the cited test asserts at
  run time — the docstring says so explicitly, so it cannot read as wider coverage
  than it provides.

https://claude.ai/code/session_01Ae2foszv754sLoi2RcZmGh

### DIFF
--- a/scripts/tests/test_cairn_operator_surface_probe.py
+++ b/scripts/tests/test_cairn_operator_surface_probe.py
@@ -230,6 +230,8 @@ class CodeFacts:
     not_found_wire_token: str
     allowlist_gates_the_404: bool
     creates_the_parent_directory: bool
+    # `X-Store-Status` token -> every HTTP code `server.py` pairs it with.
+    token_codes: dict[str, frozenset[int]]
     classify: Callable[[int, str, str], object]
 
 
@@ -261,8 +263,30 @@ def read_code_facts() -> CodeFacts:
         not_found_wire_token=wire,
         allowlist_gates_the_404=_allowlist_gates_the_404(handler),
         creates_the_parent_directory=_makes_the_parent_directory(create_fn),
+        token_codes=_token_codes(tree),
         classify=_load_cairn_cli()._classify,
     )
+
+
+def _token_codes(tree: ast.Module) -> dict[str, frozenset[int]]:
+    """Every `X-Store-Status` token `server.py` sends, and with which codes.
+
+    🔴 THIS IS WHAT KEEPS THE GUARD FROM BECOMING A PERMANENTLY-RED GATE, which
+    `claude/RULES.md` calls worse than no gate. The first cut classified every
+    refusal row the doc quotes as if it were the create path's 404 — so adding a
+    perfectly correct third row (`rc 9 [already-exists]`, which the create path
+    really does answer, at 412) would have gone red against a doc that had just
+    got BETTER. The token is now looked up in the server's own emissions, so each
+    row is checked against the code that row's token actually rides on.
+    """
+    out: dict[str, set[int]] = {}
+    for call in _calls(tree, "_respond"):
+        code = _const_int(call)
+        token = _header_literal(call, "X-Store-Status")
+        if code is None or not token:
+            continue
+        out.setdefault(token, set()).add(code)
+    return {k: frozenset(v) for k, v in out.items()}
 
 
 def _const_int(call: ast.Call) -> int | None:
@@ -370,14 +394,23 @@ def disagreements(doc_text: str, facts: CodeFacts) -> list[str]:
             )
 
     for rc, token in table.refusals:
-        refused = facts.classify(404, facts.not_found_wire_token, "")
-        actual = (refused.exit_code, refused.status)
-        if (rc, token) != actual:
+        codes = facts.token_codes.get(token)
+        if not codes:
             out.append(
-                f"the doc quotes `rc {rc} [{token}]` for the create path's 404, "
-                f"but `scripts/cairn` produces rc {actual[0]} [{actual[1]}] from "
-                f"the token `server.py::{NOT_FOUND_HANDLER}` actually puts on the "
-                f"wire ({facts.not_found_wire_token!r})"
+                f"the doc tells the operator to expect `[{token}]`, but "
+                f"`server.py` never sends that `X-Store-Status` token — it sends "
+                f"{sorted(facts.token_codes)}"
+            )
+            continue
+        produced = {
+            (r.exit_code, r.status)
+            for r in (facts.classify(code, token, "") for code in sorted(codes))
+        }
+        if produced != {(rc, token)}:
+            out.append(
+                f"the doc quotes `rc {rc} [{token}]`, but `scripts/cairn` "
+                f"produces {sorted(produced)} for that token on the HTTP "
+                f"code(s) `server.py` pairs it with ({sorted(codes)})"
             )
 
     if not facts.creates_the_parent_directory:
@@ -455,6 +488,10 @@ def test_the_code_facts_were_actually_read(facts: CodeFacts) -> None:
     assert facts.create_not_found_audit_tokens, (
         f"`{CREATE_HANDLER}` appears to reach `{NOT_FOUND_HANDLER}` from no arm "
         f"at all — suspect the AST read, not the server"
+    )
+    assert facts.not_found_wire_token in facts.token_codes, (
+        f"the token map does not even contain "
+        f"{facts.not_found_wire_token!r} — suspect `_token_codes`, not the server"
     )
     assert callable(facts.classify), "`scripts/cairn::_classify` did not load"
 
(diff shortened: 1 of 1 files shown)


ZacxDev:
## CI: the pytests red is INHERITED from `main`, not from this diff

`tekton/devrc-pytests` — **fail**, `collected=22361 passed=22357 skipped=2`, one
failing test: `test_every_kill_server_call_site_in_the_repo_is_classified`.
`tekton/devrc-nodetests` **pass** (1449/1449), `tekton/devrc-cairn-client-runs` **pass**.

Per `CLAUDE.md` — *"read the failing test's name and ask whether your diff can
reach it"* — I ran the cheap discriminating control rather than reasoning about it.

**This branch** (base `9a4411c8`):

```
added:   ['claudedocs/handoff-mention-system-repos.md']
removed: []
```

**`origin/main` tip, unmodified checkout** (`f9415d93`, detached worktree, same runner):

```
added:   ['claudedocs/handoff-cairn-oss-multi-instance.md']
removed: []
```

So `main` is red on this test **on its own**, and the two runs name *different*
unclassified docs only because `main` moved between them. Neither file is in this
PR's diff — `git diff --name-only origin/main...HEAD` is exactly one path,
`scripts/tests/test_cairn_operator_surface_probe.py`, which contains no `tmux …
kill-s*` mention at all. `claudedocs/handoff-mention-system-repos.md` was added to
`main` in `79338677` and last touched by #1535.

It is also already known and under repair by other PRs — #1522 (*"devrc main is
red on the kill-mention ledger, and a docs-only PR was not the cause"*), #1549
(*"scope the wide-kill ledger to executable text — prose red-lined main 3x in one
day"*) and #1551. Nothing for this PR to do; flagging it so the red is not debugged
against this diff.

⚠ Stated at the scope measured: two runs, dev-host tier, one test selected by
`-k kill_server_call_site`. I did not re-run the full sandbox tier.

PR #1559, opened 2026-09-12
fix(handoff): un-red main — my own handoff documenting this recursion caused the next instance of it

ZacxDev:
🔴 **`main` is RED at `b62d1bf1`** on `test_every_kill_server_call_site_in_the_repo_is_classified`, and the offender is `claudedocs/handoff-gate-speed-and-ci-signal.md` — a doc I pushed one turn ago as `58c03a77`. One line, one file.

## What happened

The paragraph that broke it is the paragraph **describing this exact recursion**. It spelled the session-kill token literally while explaining that `#1534` had fixed a scanner bug about that token; `_MENTION_RE` matched it; the doc became the next unclassified offender.

## Three sessions did this concurrently, inside one hour

| PR | what its own title says |
|---|---|
| `#1557` | main is RED again, one hour later, from a doc quoting **THIS SCANNER'S OWN OUTPUT** |
| `#1556` | un-red main — my own merged handoff became **the sixth ledger offender**, and eliding it was not enough |
| **this** | the doc explaining the above became the next one |

Each fixed the previous offender and created the next. **The ledger is at six offenders and the count is a function of how many people have written about it** — that is the finding, not the anecdote. Three independent sessions reproducing it in an hour is stronger evidence than any argument in the doc.

## The fix, and why it is not a seventh ledger row

Uses the `sk`+`ill-session` split the guard's own file already uses for this purpose.

⚠ **Not evasion, and the distinction is load-bearing**: the ledger enumerates files that could **execute** a wide kill, and a handoff cannot. One row per write-up forever turns a list of call sites into a list of documents — that is the ledger degrading, not the doc. Re-justify on a real call site; use the split form for a prose quotation. **If you would rather every mentioning file be classified regardless, say so — it is a one-line change.**

## Verification

- Both previously-red guards pass on this tree.
- The doc matches `_MENTION_RE` **zero** times.
- 🔴 **Positive control** — re-introducing the literal makes the guard fail again with `added: ['claudedocs/handoff-gate-speed-and-ci-signal.md']`. So this **fixes rather than hides**.

## Supersedes #1558

That PR's ledger rows for `handoff-cairn-oss-multi-instance.md` are now **redundant** — `#1556`/`#1557` classified it while #1558 was open, which is also why #1558 went `CONFLICTING`. Closing it in favour of this.

## Does not close the loop

The next handoff about any of this reopens it. The design options — scope the scanner off `claudedocs/`, or auto-classify a prose-only file that executes no tmux command — are **rank 9** of `claudedocs/handoff-gate-speed-and-ci-signal.md`.

https://claude.ai/code/session_01PQx16RMr8YQYBfsXaBEsSm

### DIFF
--- a/claudedocs/handoff-gate-speed-and-ci-signal.md
+++ b/claudedocs/handoff-gate-speed-and-ci-signal.md
@@ -436,8 +436,22 @@ bash ~/workspace/devrc/scripts/ship.sh
 - **Observed (with values):** `_KILL_MENTION_LEDGER` (`test_guard_core.py:2525`) is a two-way pin over
   every file mentioning a wide tmux kill. Four `claudedocs/` rows now; **three exist purely because
   someone documented the guard.** `#1520` classified one, `#1534` classified two more **plus a real
-  scanner bug** (it matched `kill-session` inside `sk`+`ill-session`), and `main` went red again on a
-  new doc **within minutes of #1534 merging**.
+  scanner bug** (it matched the session-kill token inside `sk`+`ill-session`), and `main` went red
+  again on a new doc **within minutes of #1534 merging**.
+  🔴 **AND THIS PARAGRAPH BROKE `main` ITSELF, ONE TURN AFTER BEING WRITTEN — measured, not a
+  hypothetical.** Its first draft spelled that token literally; `_MENTION_RE`
+  (`test_guard_core.py`, `(?<![A-Za-z0-9_])kill-s(?:erver|ession)`) matched it, and THIS DOC became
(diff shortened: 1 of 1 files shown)


ZacxDev:
Obsolete — closing unmerged. #1561 closed this structurally by stopping the kill scanners reading `claudedocs/` at all, which is strictly better than my fix: it ends the treadmill instead of paying one more round of it. Verified on `origin/main` 4e970998 — both kill guards pass with my doc's literal still present, so nothing here is needed. The three-session recursion this PR documented is going into the handoff instead, where it belongs.

PR #1569, opened 2026-09-12
feat(mention-open): the auto-open was invisible to Tier B — record it tagged, and report every click outcome

ZacxDev:
Two coupled changes to `mention-open`, shipped together because the second is only readable if the first happened.

## (A) The single-candidate AUTO-OPEN now records a pick — tagged by path

`main()`'s "exactly one candidate → just open it" shortcut returned `open_url(...)` **without calling `record_pick`**. That is the common, fast shape — an explicit `owner/repo#N`, a mapping hit, a measured checkout — so Tier B's pick log only ever saw the clicks the handler could **not** resolve. The learned preference described the ambiguous minority and was read as describing the operator.

Both paths record now, and the row says which:

```
{"t": 1757…, "repo": "owner/repo", "n": 1291, "via": "auto"|"picker"}
```

* `via` is a **required keyword** on `record_pick` — no default. A default would make the tag a thing a call site can forget, and the mislabelling would be silent and permanent in an append-only 0600 log. A `via` outside `PICK_VIA_RECORDED` is a quiet `False` and no write.
* `pick_scores` is **unchanged** — it still counts every row the same.

### On-disk backwards compatibility

**The log format is additive and old rows keep working.** A row with no `via` is **loaded, not dropped**, and classifies as `PICK_VIA_UNKNOWN` (`"unknown"`).

They are *not* back-labelled `"picker"`. They were in fact picker picks — that was the only writer — but the **row** does not say so, and a reader cannot tell a pre-change row from a future writer that forgot the field. The same coercion covers a `via` of the wrong type and a tag written by a *newer* handler: `pick_via` is total, and none of those is ever a reason to discard a row whose `repo`/`n`/`t` are good.

### 🔴 This is NOT a pure addition, and it changes the live picker order

An earlier draft said recording auto-opens was "strictly more data". **That was wrong at the caps.** The log is bounded twice and both bounds are FIFO with no `via` filtering: `load_picks` scores the last `PICKS_MAX_ROWS` rows and `_compact_picks` truncates the file to the same tail. Auto-opens are the common shape, so at real volume they do not sit beside the picker rows — they **push them out**.

And since `pick_scores` feeds `order_universe`, recording auto-opens **re-weights what the operator is offered on their next ambiguous click**. That is defensible — it is still evidence about what this operator means by a number — and it is a behaviour change that must be stated rather than discovered. If it crowds out the picker signal, the fix is a `via`-aware cap or weight, which is exactly what the tag exists to make possible.

## (B) Click-path outcomes are reported to the activity telemetry

One row per click that reached a decision, through the pipeline's existing best-effort helper (`scripts/collector/invocation.py` → `spool_emit` → the collector daemon):

```
source=tool  kind=invocation  text=mention-open
payload = {"tool":"mention-open", "outcome":…, <dims>}
```

**Outcomes:** `auto-open`, `picked`, `dismissed`, `no-selection`. Deliberately not every exit — the refusal paths (`refuse()`, `--print`, the six-digit colour toast) emit nothing.

### The dims

| field | when | meaning |
|---|---|---|
| `repo` | always (may be `""`) | plain `owner/repo` of the row that was **opened** |
| `platform` | always (may be `""`) | `github` / `clawgate` / `clickup` |
| `picker_shown` | when measured | **three-valued**: `true`, `false`, or **absent** when not measured |
| `offered_total` | when a picker was shown | rows in the list as presented |
| `rank` | when a row was chosen | **0-based** position in the list as presented |
| `ordered` | when a row was chosen | was **this row** placed by `order_universe`? |
| `pinned_above` | when there was an ordered block | rows sitting above it |
| `plausibility` | only when `ordered` | `plausible` / `below` / `unknown` / `impossible` |
| `reason` | on every picker outcome | why `pick()` returned what it did |
| `surface` | when something was opened | `browser` / `tui` |

`repo` is emitted **plain, not hashed** — matching the `source=mentions` precedent. The line is: *emit what the operator TOUCHED, never what they were OFFERED.*

### 🔴 Absences a consumer must not read as zero

1. **An auto-open has no `rank`, `plausibility` or `offered_total` — the keys are OMITTED.** Nothing was offered, so there is no list the chosen row can have been ranked in. `rank=0` would read as "took the top row", the *success* value for the very ordering this data exists to evaluate. **Filter `picker_shown = true`**; `JSONExtractInt(payload,'rank')` returns `0` for a missing key and cannot detect the absence.
2. **A missing `plausibility` ≠ `plausibility = "unknown"`.** Missing = the ordering did not place this row. `"unknown"` = it did, and the repo had no table entry. The class ships as a **name**, never its ordinal, because `CLASS_PLAUSIBLE` is `0` and so is an absent key.
3. **`rank` mixes two populations unless you filter `ordered`.** `candidates` is `[evidence rows] + [universe rows]`: the clawgate task, the pane guess and a mapping hit are **pinned** above the ranked block. Without `ordered`, "chosen rank clusters near 0" would read as "the ordering works" on the commonest picker shape whether or not it does. `rank - pinned_above` is the position *within* the ranked block.
4. **A dismissal carries no `surface`** — nothing was opened, which is a different fact from "opened somewhere unrecorded".

### `dismissed` means a dismissal

`pick()` returns `""` from **six** endings and in three of them **no picker was ever displayed** (fzf missing, the spawn raising, a terminal that exited first). Reporting those as `dismissed picker_shown=true` put clicks that saw nothing into the ordering's own denominator. `pick()` now records *why* (`PICK_REASON_*`, 8 values, ledger-pinned); only a real dismissal is `dismissed`, the rest are `no-selection` carrying `reason`, and `picker_was_shown` is three-valued so an unmeasured case omits the field.

### `surface` — the seam with #1582

#1582 routes a GitHub mention to a neovim review buffer instead of `xdg-open`. Without this dim a TUI open and a browser open would emit an **identical row**, making the new surface invisible to the telemetry built to answer "is this used?".

The value is **measured, not asserted**: it comes back from `open_reference(url) -> (rc, surface)`, so it is reported by the code that did the opening. A literal at the call site would keep saying `browser` from a branch that no longer uses one. That wrapper is the **single integration point** for #1582. It degrades honestly: without #1582 it reports `browser` because the browser genuinely opened it — never a "not asked" placeholder. Clawgate/ClickUp rows are `browser` and always will be.

## Disclosure

🔴 **The offered set never reaches the sink, structurally.** `click_dims` takes **scalars only**, so there is no argument through which an unchosen row can get there. `emit_click(outcome, **dims)` hands its keywords to `click_dims`, so a field nobody added to that signature is a `TypeError` inside the guard and **writes nothing**.

The guard is `test_no_universe_token_can_reach_the_TELEMETRY_payload`, and its line is narrower than the notes': a telemetry row **must** name one repository, so the assertion is *"every universe token except that one's three spellings is absent"*. `touched` is captured from the picker's return value, not read out of the payload, so a leaking mutant cannot widen its own allowance. Its positive control drives the real emitter, spool, decoder and guard with a realistic leak and asserts it raises. `test_the_spool_DECODER_can_see_a_name_the_raw_line_HIDES` is the guard on the guard — a v1 spool line base64-encodes its fields, so a substring check on the raw line is vacuously true.

## Cost to the click

* The telemetry import is **lazy**: median **3.7 ms** (range 3.3–6.8, box load) on the first emit of a click; 0.05 ms after, which no click reaches. Refusals pay nothing.
* Emitted **after** `open_reference` — a non-blocking `Popen` — so it lands behind the browser launch. `record_pick` stays in front: it is what the next click reads.
* **No network, no subprocess, no new verb** on the click path.
* Nothing here can cost the click. This found a real defect in the first draft: `click_dims(...)` was built at the call site, outside `emit_click`'s `try` — measured with it raising, the browser opened and then `guarded_main` returned **1** with a "mention-open failed" toast.

## Test evidence

**Red-at-base / green-at-HEAD.** Base `6ff4d215`, measured in a detached worktree with this PR's test file dropped in:

* `test_a_single_candidate_AUTO_OPEN_RECORDS_the_pick` — **RED at base** with `AssertionError: the AUTO-OPEN path recorded 0 picks` (`assert 0 == 1`), **green at HEAD**. The only genuine regression test: it names no new API and fails on behaviour alone.
* Everything else red at base by `AttributeError`/`KeyError` = coverage for new behaviour, **not** regression coverage, labelled as such in the file.
* `test_a_clawgate_or_clickup_AUTO_OPEN_records_NOTHING` is an **invariant guard**, labelled.

**Mutation battery** (`mutation_battery_mentions.py`, rows K72–K92, `PYTHONDONTWRITEBYTECODE=1`): **22/22 KILLED, all attributed**, P1 control killed, `problems: none`, `restore: OK` by hash.

Three sweeps each found real defects in this PR's own guards:
* **K74** — the tag assertions read their expectation from the module's own constants, so `PICK_VIA_AUTO = "picker"` passed the one test whose subject is *"the two values differ"*. All spellings are literals now.
* **K87 / K89 SURVIVED** — every dismissal test stubbed `pick`, so nothing exercised the seam where the real `pick()` records the reason; and the F3 test drove only the guessed arm. Closed with a seam test over all seven of `pick()`'s exits (ledger-pinned to the reason vocabulary) and a dead-end-2 test.
* **K90 scored KILLED-WRONG-REASON** for a reason worth recording: the assertions **subscripted** the missing dim, so the mutant raised `KeyError` and pytest rendered no message at all. An assertion that indexes the thing a deletion-mutant deletes can never be attributed — both sites use `.get` now.

The anchors test (`test_mutation_battery_anchors.py`) caught **six** rows at `0x` across this PR, each time preventing a silent `SURVIVED`.

## 🔴 A pre-existing flake this PR did NOT cause — notes for whoever fixes it

`test_REAL_INTERACTIVE_fzf_opens_on_the_FIRST_INPUT_ROW_with_no_query` failed once in CI on this branch (`fcd82015`) and **passed on the re-trigger** (`1012466a`). Recorded here so it is not lost in a task thread:

* **Pre-existing** — present at this branch's merge base `6ff4d215`.
* **Untouched** — this diff adds/removes **0** lines matching `_fzf_interactive_first_row|REAL_INTERACTIVE|pty\.fork|_PTY_PROMPT|_drain`.
* **Not reproducible standalone** — 20 consecutive runs at system load **84–92** (no load generator spawned): **20 passed, 0 failed**. It passes in ~0.95s.
* ⚠ **That does not prove it fine.** A standalone run cannot reproduce the failing condition: CI runs it under **xdist (`gw1`) beside 14,399 tests in one target**. What is narrowed is that *ordinary system load is insufficient — it needs the concurrent-worker context.*
* **The mechanism, from the gate log**: fzf **did draw its prompt** (the earlier 20s assertion passed, or the message would have been *"fzf never drew its prompt in 20s"*). Enter was written; the selection pipe then returned `b''` inside a **second** 20s deadline. The test's own message says it: *"the measurement did not happen, so neither did the assertion."*
* **The remedy is removing the timing dependence, not re-running** — two 20s pty deadlines around `pty.fork()` + `execvp("fzf")`. Not done here: it is not this PR's code, and one observation does not justify touching it.

## A seam that is deliberately left open

`emit_click`'s guard covers the emitter **and** the dim assembly. Outside it, enumerated rather than summarised: the picker arm's `picked_rank`, `picked_platform`, `picked_class` and `picked_ordered` lookups; the AUTO arm's `candidates[0]["platform"]` subscript; and `repo_of_github_url(...)` on both arms. None is moved in, because doing so would mean handing `emit_click` the **candidate list** — the one thing the disclosure rule forbids. A narrower guarded surface was chosen over a wider disclosure surface. All are pure and total on the values `main()` can hold there; that is an argument, not a guarantee, and it is stated as one.

## Not in scope

* No `adoption-scan.py` REGISTRY row — a different question. Rows are queryable directly (`source='tool' AND kind='invocation' AND text='mention-open'`).
* No collector/ClickHouse schema change — the dims land in the existing `payload` JSON.
* **A repo-wide `monkeypatch.undo()` guard.** The one added here is scoped to `test_mention_open.py`; six other files under `scripts/` call `undo()` and most are probably fine (the hazard needs an autouse fixture redirecting *host state*). Widening it is a follow-up, closed when a repo-wide PR merges or a named reader dismisses it on this thread.

## Merge sequencing

**#1569 first, then #1582.** When #1582 lands, the only edit needed here is the branch inside `open_reference` — which is why it is documented as the single integration point.

### DIFF
--- a/scripts/mention-open.py
+++ b/scripts/mention-open.py
@@ -95,15 +95,27 @@ move. A silent empty picker would be the same silent zero one layer up.
 repos into this PUBLIC repository. It may go to the operator's own screen and
 NOWHERE ELSE: never to a log, never to activity.events, never to a test fixture,
 never to stderr. `notify()` prints, so the refusal paths below name only the
-clicked text — never a row from the universe.
+clicked text — never a row from the universe. ⚠ THE TELEMETRY SINK ADDED IN
+2026-09 DOES NOT WEAKEN THIS. It reports the one row the operator opened and
+three scalars about the list (rank, size, class); `click_dims` is never handed
+the candidate list at all, so there is no argument through which an offered row
+could reach it.
 
 🔴 THE SAME RULE COVERS TWO MORE FILES THIS HANDLER NOW TOUCHES, and both are
 0600 under the same directory, outside every checkout:
   * `known_ranges.json` — `{"owner/repo": <highest issue-or-PR number>}`, whose
     KEYS are the private names (the mirror of the two files above, where the
     values are). Written by `regen-known-repos.py`.
-  * `picks.jsonl`       — one `{"t","repo","n"}` line per repository the
-    OPERATOR chose from the picker. Written HERE, by `record_pick`.
+  * `picks.jsonl`       — one `{"t","repo","n","via"}` line per repository the
+    OPERATOR opened, from the picker AND from the single-candidate auto-open,
+    tagged with which. Written HERE, by `record_pick`.
+
+🔴 AND ONE SINK THAT IS *NOT* A FILE: the handler reports the OUTCOME of a click
+to the activity-telemetry spool (`source=tool`, `kind=invocation`,
+`text=mention-open`). The rule above is unchanged and the distinction is the
+whole of it — what leaves this process is what the operator TOUCHED (the repo
+whose page opened, the rank their row sat at, how many rows were offered), never
+what they were OFFERED. See the CLICK-PATH TELEMETRY block.
 
 ORDERING, AND WHY IT IS ONLY ORDERING
 The picker offers ~392 rows for a clicked `#N`, and they used to arrive in one
@@ -500,6 +512,45 @@ PICKS_HALF_LIFE_DAYS = 30.0
 # keeps is always at least what a reader would use.
 PICKS_COMPACT_AT = PICKS_MAX_ROWS * 2
 
+# 🔴 WHICH CODE PATH PRODUCED A PICK. Until 2026-09-11 the log recorded only the
+# rows the operator selected in the PICKER, and `main()`'s single-candidate
+# shortcut — the common, fast path — returned without recording anything. So
+# Tier B learned from the AMBIGUOUS clicks alone: every click the handler could
+# answer outright was invisible to it, and the scores described the minority of
+# clicks that were hard rather than what this operator means by a number.
+#
+# Both paths are recorded now, and the row says WHICH, because they are not the
+# same evidence. A picker pick is a CHOICE among rows the operator read; an
+# auto-open is the handler's own resolution, which the operator only implicitly
+# ratified by not complaining. A later scoring change may weight or exclude one
+# of them — the tag is what makes that possible without having thrown the data
+# away. Nothing reads it today: `pick_scores` is unchanged and counts every row
+# the same, which is the pre-change behaviour for picker rows and strictly more
+# data for auto rows.
+PICK_VIA_AUTO = "auto"
+PICK_VIA_PICKER = "picker"
+
+# 🔴 WHAT AN UNTAGGED ROW BECOMES, AND WHY IT IS NOT BACK-LABELLED `picker`.
+# Rows written before the tag existed carry no `via` at all (7 of them on the
+# laptop when this shipped; the workbench had no log). They were in FACT picker
+# picks — that was the only writer — but the ROW does not say so, and a reader
+# cannot tell a pre-change row from a future writer that forgot the field. So
+# they load as `unknown`: a third value a consumer must handle deliberately,
+# rather than a claim about provenance the data cannot support. An unrecognised
+# value (a corrupt row, or a tag written by a NEWER version of this handler)
+# lands here too — it is never a reason to DROP the row, because `repo`/`n`/`t`
+# are what the scoring reads and those are still good.
+PICK_VIA_UNKNOWN = "unknown"
+
+# What `record_pick` may WRITE. Deliberately narrower than what `load_picks` may
+# RETURN — a writer must name one of the two real paths, while a reader must be
+# total over whatever is on disk.
+PICK_VIA_RECORDED = (PICK_VIA_AUTO, PICK_VIA_PICKER)
+# What `load_picks` may return. Pinned two-way by
+# `test_the_pick_VIA_vocabulary_is_pinned_two_way`.
+PICK_VIA_VALUES = (PICK_VIA_AUTO, PICK_VIA_PICKER, PICK_VIA_UNKNOWN)
+
+
 # How far apart two reference numbers have to be before proximity stops helping.
 # `#1290` and `#1291` in one repo are the same week's work; `#3` and `#1291` are
 # not. A soft 1/(1+d/SCALE) curve rather than a window, so nothing falls off a
@@ -625,8 +676,27 @@ def ordering_state(ranges: dict[str, int], age: float | None) -> str:
     return ORDER_APPLIED
 
 
+def pick_via(row: dict) -> str:
+    """Which path produced a pick row — one of `PICK_VIA_VALUES`. Pure, total.
+
+    🔴 TOTAL ON PURPOSE, AND THAT IS THE BACKWARDS-COMPATIBILITY RULE IN ONE
+    PLACE. Every way a row can fail to name a path it wrote — the field absent
+    (a row written before the tag existed), the wrong type, a value this version
+    does not know (a NEWER writer) — is `unknown`. None of them is a reason to
+    discard the row: `load_picks` already decided the row is well-formed on the
+    three fields the scoring actually reads, and dropping it over a label would
+    silently delete history the operator earned.
+
+    🔴 ONE DEFINITION, because `load_picks` is not the only reader that will
+    want this. A predicate open-coded at two sites is wrong at one of them.
+    """
+    via = row.get("via") if isinstance(row, dict) else None
+    return via if via in PICK_VIA_VALUES else PICK_VIA_UNKNOWN
+
+
 def load_picks(path: Path | None = None, now: float | None = None) -> list[dict]:
-    """The operator's recent picks as `[{"t": float, "repo": str, "n": int}]`.
+    """The operator's recent picks as
+    `[{"t": float, "repo": str, "n": int, "via": str}]`.
 
     Newest LAST, capped by `PICKS_MAX_ROWS` and filtered to
     `PICKS_MAX_AGE_DAYS`. 🔴 EVERY failure is [] — absent, unreadable, and
@@ -638,6 +708,11 @@ def load_picks(path: Path | None = None, now: float | None = None) -> list[dict]
     per pick, so the rows that matter are the last ones; taking the first 500
     would freeze the preference at whatever the operator liked when the file was
     new and never move again.
+
+    🔴 `via` IS ALWAYS PRESENT ON THE WAY OUT AND NEVER REQUIRED ON THE WAY IN.
+    A row that does not name a path is `unknown`, not dropped — see `pick_via`,
+    which owns that rule. Every caller therefore gets a four-key row whatever the
+    file's vintage, so nothing downstream has to `.get` around a missing field.
     """
     path = path or PICKS_PATH
     now = time.time() if now is None else now
@@ -673,7 +748,7 @@ def load_picks(path: Path | None = None, now: float | None = None) -> list[dict]
             continue
         if (now - float(t)) / 86400.0 > PICKS_MAX_AGE_DAYS:
             continue
-        out.append({"t": float(t), "repo": repo, "n": n})
+        out.append({"t": float(t), "repo": repo, "n": n, "via": pick_via(row)})
     return out
 
 
@@ -812,8 +887,17 @@ def narrow_dir(directory: Path) -> None:
 
 
 def record_pick(repo: str, num: str, path: Path | None = None,
-                now: float | None = None) -> bool:
-    """Append one `{"t","repo","n"}` line to the pick log.
+                now: float | None = None, *, via: str) -> bool:
+    """Append one `{"t","repo","n","via"}` line to the pick log.
+
+    🔴 `via` IS A REQUIRED KEYWORD AND HAS NO DEFAULT, WHICH IS THE WHOLE POINT
+    OF THE FIELD. A default would make the tag a thing a call site can forget:
+    the auto-open path — the one this field exists to make visible — would then
+    record itself as whatever the default said, and the mislabelling would be
+    silent and permanent in a 0600 append-only log. Two call sites exist and
+    both must state which they are. A value outside `PICK_VIA_RECORDED` is a
+    quiet `False` and NO write, for the same reason a malformed repo is: a row
+    whose provenance is wrong is worse than no row.
 
     Returns True if the line was WRITTEN — which is not quite "survived". A
     compaction racing this call carries over anything appended past its read
@@ -859,6 +943,8 @@ def record_pick(repo: str, num: str, path: Path | None = None,
     reader would discard is not worth writing, and a log full of them would
     push real history past `PICKS_MAX_ROWS`.
     """
+    if via not in PICK_VIA_RECORDED:
+        return False
     if not (isinstance(repo, str) and _OWNER_REPO_RE.match(repo)):
         return False
     try:
@@ -869,7 +955,7 @@ def record_pick(repo: str, num: str, path: Path | None = None,
         return False
     path = path or PICKS_PATH
     row = {"t": round(time.time() if now is None else now, 3),
-           "repo": repo, "n": n}
+           "repo": repo, "n": n, "via": via}
     try:
         path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
         narrow_dir(path.parent)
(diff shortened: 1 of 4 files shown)


ZacxDev:
## CI triage — both pytest failures are inherited, and my PR body named the wrong pair

`tekton/devrc-pytests` is **red**; `devrc-nodetests` (1449/1449, floor 1367) and `devrc-cairn-client-runs` are green.
`collected=22378 passed=22374 skipped=2 failed=2`. GitHub's status description only had room for **one** of the two names, so I pulled the gate pod's own log (`devrc-ci-zjrlq-gate-pod`, 604 KB) instead of guessing at the second.

### The two failures, each with a discriminating control

**1. `scripts/claude-hooks/tests/test_guard_core.py::test_every_kill_server_call_site_in_the_repo_is_classified`**
`added: ['claudedocs/handoff-gate-speed-and-ci-signal.md']` — a file with no `_KILL_MENTION_LEDGER` entry. This diff does not touch it and adds no `kill-server`/`kill-session` text anywhere.
✅ **Already fixed upstream.** Detached worktree of `origin/main` @ `61f41adf`, same file, `-k kill`: **96 passed, 0 failed** (and the whole file was 1537/1537 at `5dac6ec2`). It is red on this branch only because the branch is based on `6ff4d215`, which predates the fix.

**2. `scripts/tests/test_runner_bound_ledger.py::test_every_site_writing_its_OWN_runner_bound_is_in_the_ledger`**
`{('scripts/tests/test_scoped_tests_shared_surface.py', 'ABSENT'): 1}` — a file this diff does not touch. Introduced on `main` by `622fc2d8`.
🔴 **Still red on a clean `origin/main`.** Same detached `61f41adf` worktree: **identical assertion, identical offending file**.

**And "fails on both trees" is not on its own proof it is inherited** — so I also ran the test's *own* `_discover()` against both trees and diffed the result rather than reasoning about it:

```
MY TREE      : 8 discovered site(s), ledger has 7
CONTROL(main): 8 discovered site(s), ledger has 7
sites ONLY in my tree     : []
sites ONLY in the control : []
UNLEDGERED in my tree     : [('scripts/tests/test_scoped_tests_shared_surface.py', 'ABSENT')]
UNLEDGERED in the control : [('scripts/tests/test_scoped_tests_shared_surface.py', 'ABSENT')]
IDENTICAL unledgered set? True
files THIS PR changed that the scan discovers at all: NONE
```

The scan's population is "files that spawn a test RUNNER with a timeout bound of their own". **Not one of the four files this PR changes is in it**, so the test is structurally unable to see this diff — which is a stronger claim than "it was red before too".

### Correction to the PR body

The body says `main` is "independently RED … on two tests in `scripts/claude-hooks/tests/test_guard_core.py`" and names `test_no_tracked_shell_text_writes_a_kill_this_guard_would_deny` as the second. **That is wrong.** Only *one* guard_core test failed; the real second failure is the runner-bound ledger, in a different file — and the guard_core one has since been fixed on `main` while the ledger one has not. I carried that pair from a briefing without checking it, which is exactly the thing this PR's own tests are about. Correcting it here rather than silently editing the body, since the body may already have been read.

### Not done, deliberately

I have **not** merged current `main` into this branch. Doing so would clear failure 1 and would gate the *merged* tree rather than the branch — `strict` is off, so a green check here is only ever a claim about this branch. Your call; it is one command and one more CI cycle.

ZacxDev:
## Audit round 1 — F1 (blocker), F2, F3 fixed in `67453477`; F4–F8 too

`audit-claims` for the next round is at the bottom.

### F1 🔴 — it was writing to the operator's live data, and I reproduced it

`monkeypatch.undo()` in `test_the_TELEMETRY_can_never_cost_the_CLICK` reverted the AUTOUSE `_mapping_is_never_the_operators` fixture, not just its own patches — `monkeypatch` is one function-scoped object shared with every fixture in the test. The `guarded_main` call after it ran against real host state, took this PR's **new auto-open arm**, and appended to `~/.config/mention-open/picks.jsonl`.

**Measured both directions, same trap, same command** (`XDG_CONFIG_HOME` redirected, `GIT_CONFIG_GLOBAL/SYSTEM=/dev/null` so the trap cannot move git):

| tree | suite | rows in trap |
|---|---|---|
| `9d15aac0` (buggy) | **375 passed** | **1** |
| fixed | **394 passed** | **0 (file absent)** |

A fully green suite was writing to the operator's dataset — which is the whole point: nothing in the suite could see it.

Fixed with `pytest.MonkeyPatch.context()`. **The guard**: `test_no_test_in_this_file_calls_monkeypatch_undo`, an AST guard keyed on `.undo()` called on a function's own `monkeypatch` **parameter** (a `MonkeyPatch.context()` object calling its own `.undo()` is scoped and harmless — banning that would push people back onto the shared fixture). It has a **positive control** (a planted `undo()` must be found) and a **negative control** (the scoped form must not be flagged).

**Why no existing guard could have caught it, as you said:** `test_the_autouse_redirect_is_IN_FORCE` and `..._REACHES_A_SUBPROCESS` both assert state at *their own* runtime, and pytest tears the fixture down between tests — so no runtime assertion inside test A can observe test B's `undo()`. Only reading the source can. **Scoped to this file deliberately**: measured, six files under `scripts/` call `monkeypatch.undo()` (`test_subsystem_store_api.py`, `test_clawgate_writeback_guard.py`, `dl-router/tests/test_dedupe.py`, …) and most are probably fine — the hazard needs an autouse fixture redirecting *host state*. Widening it is a real follow-up, not something to impose on other people's files from inside this PR. **FOLLOW-UP, closing condition: a repo-wide version merges, or a named reader dismisses it in writing on this thread.**

I did **not** touch the operator's `picks.jsonl` — agreed, it is theirs.

### F2 — `dismissed` now means a dismissal

`pick()` records **why** it returned nothing (`PICK_REASON_*`, 8 values, ledger-pinned). Only a real dismissal is `dismissed`; the rest are `no-selection` carrying `reason`. `picker_was_shown` is **three-valued** — `True`/`False`/`None` — because both guesses are wrong for an unmeasured case in the direction that matters, so an unmeasured picker now **omits** `picker_shown` rather than asserting one.

### F3 — `rank` no longer mixes two populations

Rows carry `ordered` (was **this** row placed by `order_universe`) and `pinned_above` (rows above the ranked block), and `plausibility` is gated on `ordered` rather than on "the ordering ran". `rank - pinned_above` is the position within the ranked block. **A consumer measuring the ordering must filter `ordered = true`** — that is in the `click_dims` docstring and the PR body.

### F4–F8

- **F4** — you are right and the comment was wrong: both bounds are FIFO with no `via` filtering, so at real auto-open volume this means permanently **fewer** picker rows on disk, not "strictly more data". Corrected, and the live-ordering change is now stated as a behaviour change in both the source and the PR body.
- **F5** — three mechanisms, not two; `conftest.py`'s `no_real_activity_spool` already covered the bare-pytest case. The line buys **per-test isolation**, and the comment now says that and records the mutation result.
- **F6** — the other half of `unknown` is now driven end-to-end (ordering active, chosen repo absent from the table), with a positive control that the table really was `ORDER_APPLIED`.
- **F7** — the outcome ledger is two-way for real: it parses the handler's AST for every `CLICK_*` constant and every name handed to `emit_click`. It previously compared the tuple only to literals the test file itself spells, so an emitted-but-unledgered outcome stayed green.
- **F8** — the paragraph now enumerates all six unguarded expressions, including `candidates[0]["platform"]`.

### Battery

5 new mutants (K85–K89); K80 and K82 re-anchored — **the anchors test caught both at 0x** rather than letting them score silent SURVIVEDs.

🔴 **First sweep: K87 and K89 SURVIVED — two real gaps in the guards I had just written.** K87: every dismissal test stubbed `pick`, so nothing exercised the seam where the *real* `pick()` records the reason — the verified-in-isolation shape exactly. K89: the F3 test drove the *guessed* arm, and K89 mutates *dead end 2*. Closed with a seam test over all seven of `pick()`'s exits (ledger-pinned against the reason vocabulary, so a new reason with no case fails) and a dead-end-2 test with a negative control on the pinned clawgate row.

Re-run: **19/19 KILLED, all attributed, P1 control killed, `problems: none`, `restore: OK`.**

### Verification

- `test_mention_open.py` **394 passed**; the 8 affected suites **783 passed** (dev-host tier, base `6ff4d215`).
- CI red remains inherited — see the triage comment above; re-confirmed against `origin/main` @ `61f41adf`, including running the failing test's own `_discover()` on both trees (identical sets; **no file this PR changes is in its population**).

```audit-claims
round: 1
audited: 9d15aac0c3d1ba4c00b6fed0e1e0e46dc0fca3b6
fixed_at: 67453477
findings:
  F1: blocker, scaffolding — monkeypatch.undo() reverted the autouse host-state redirect; wrote to the operator's real picks.jsonl. FIXED (MonkeyPatch.context) + new AST guard with positive and negative controls. Reproduced 1 row -> 0 rows.
  F2: should-fix, payload — `dismissed` emitted for 4 non-dismissals. FIXED (PICK_REASON_* + no-selection + three-valued picker_shown).
  F3: should-fix, payload — rank mixed pinned rows with ordered rows. FIXED (ordered + pinned_above; plausibility gated on ordered).
  F4: nit->real, payload comment — "strictly more data" false at the caps. FIXED; live-ordering change now stated.
  F5: nit, scaffolding comment — named 2 of 3 mechanisms. FIXED.
  F6: gap, scaffolding — untested half of plausibility=unknown. FIXED.
  F7: gap, scaffolding — outcome ledger pinned one way. FIXED (AST-driven).
  F8: nit, payload comment — seam enumeration incomplete. FIXED.
open:
  - repo-wide monkeypatch.undo() guard (6 other files) — deliberately NOT in this PR. Closes when a repo-wide PR merges or a named reader dismisses it here.
mutation: 19/19 killed attributed, P1 killed, problems none (K87+K89 survived the first sweep; both gaps closed)
```

ZacxDev:
## F1 end-to-end: byte-identical against the operator's REAL paths — and I added 4 of those rows

The trap measurement in the previous comment proves the fix under a redirected `XDG_CONFIG_HOME`. That is one step removed from the configuration that did the damage, so here is the direct one — suite run with the operator's real host-state paths in scope, no trap, `picks.jsonl` read only:

```
BEFORE: 40 rows  sha=737f73df25d0
394 passed in 60.97s
AFTER:  40 rows  sha=737f73df25d0
RESULT: NO GROWTH, BYTE-IDENTICAL
```

**Accounting, because it is mine to own:** you measured that file at **36 rows**; it is **40** now. The extra four are from **my own test runs before F1 landed** — every `pytest scripts/tests/test_mention_open.py` I ran while iterating had `ACTIVITY_SPOOL_DIR` redirected but *not* `XDG_CONFIG_HOME`, so `undo()` restored the real `PICKS_PATH` and the auto-open arm appended. I was running the suite to verify the feature and contaminating the data with every run, which is precisely the shape of the defect: **the leak is invisible to the thing you would use to find it.**

I have **not** truncated or edited the file — agreed that it is the operator's. It is now stable: the sha above is unchanged across a full post-fix run.

⚠ What I cannot tell you: which 4 of the 40 are mine. The rows carry `{t, repo, n, via}` and nothing distinguishes a test-run row from a real click, so attribution is by timestamp and inference, not by the data. That is a property of the log, not of this measurement — and it is an argument for the operator simply deleting the file, since all 40 rows are fixture residue by your measurement (it was absent earlier that night) and Tier B cold-starts cleanly from nothing.

ZacxDev:
## `surface` dim added — `6f6a0101`

Agreed this is a real seam finding and not tidiness: without it this PR closes one blind spot and opens another on the exact axis it was built to instrument.

### What landed

`CLICK_SURFACES = ("browser", "tui")`, emitted as a `surface` dim.

🔴 **It is measured, not asserted.** The value comes back from a new `open_reference(url) -> (rc, surface)`, so it is reported by the code that *did the opening* rather than by a literal at the call site. A literal is a claim about what happened; when a second opener arrives it keeps saying `browser` from a branch that no longer uses one, with every suite green. That wrapper is also the **single integration point for #1582** — the branch goes there, both emit sites report it with no edit, and `test_every_surface_open_reference_can_return_is_LEDGERED` fails if the new value is not in the ledger.

**Honest degradation**, exactly as you asked: with #1582 absent it reports `browser` **because the browser is what opened it** — never a placeholder implying the question was not asked. Clawgate/ClickUp rows are `browser` and always will be, so the dim is a real measurement on that arm **today**. A **dismissal opened nothing and carries no surface at all** — absent, because "opened somewhere unrecorded" is a different fact, and defaulting to `browser` would put a never-opened click into every per-surface count.

### The ledger pin, two-way

- `surface` in `CLICK_DIM_FIELDS`, which is itself pinned against `click_dims`' own signature (`inspect.signature`), so a parameter added without a ledger row fails.
- `CLICK_SURFACES` pinned against the **handler's AST** for every `CLICK_SURFACE_*` constant — a constant with no ledger row, or a ledger row naming no constant, fails.
- An **unledgered value is dropped, not shipped** (`click_dims(surface="teletype")` emits no `surface`), with `test_every_surface_open_reference_can_return_is_LEDGERED` as the other half so the drop can never be silent.
- The outcome ledger's AST scan now excludes the non-outcome `CLICK_*` constants **by name**, not by pattern — a new `CLICK_*` constant is an outcome *by default* and must be classified deliberately.

### Battery

K90–K92 added. K77/K82/K84 **re-anchored** — the anchors test caught all three at `0x`; that guard has now stopped a row from scoring a silent `SURVIVED` six times in this PR.

⚠ **K90 first scored `KILLED-WRONG-REASON`, and the cause was not wording.** The assertions **subscripted** the missing dim, so the mutant raised `KeyError: 'surface'` and pytest rendered *no assertion message at all*. Both sites use `.get` now and lead with a short `NO SURFACE DIM` token. Worth recording as a battery-authoring trap in its own right: an assertion that indexes the thing a deletion-mutant deletes cannot be attributed.

**Re-run: 22/22 KILLED, all attributed, P1 control killed, `problems: none`, `restore: OK`.**

### Re-verification

| check | result |
|---|---|
| `test_mention_open.py` + anchors | **417 passed** |
| the 8 affected suites | **785 passed** |
| host-state leak check (real paths in scope) | **byte-identical, no growth** |

⚠ Noted while re-running the leak check: `~/.config/mention-open/picks.jsonl` is now **absent** — the operator appears to have deleted it, which is the clean outcome given all 40 rows were fixture residue. The check confirms it **stays** absent across a full suite run.

### Sequencing

Understood — **#1569 first, then #1582**, and I am not touching the merge resolution. Nothing here depends on #1582 existing; when it lands, the only edit needed is the branch inside `open_reference`.

```audit-claims
round: 2
audited: 67453477
fixed_at: 6f6a0101
findings:
  F9: should-fix (missing requirement), payload — no SURFACE/TARGET dim; a browser open and #1582's TUI open emit an identical row. FIXED: `surface` dim measured from `open_reference`, two-way ledger pin, absent on a dismissal, `browser` derivable today for clawgate/ClickUp.
open:
  - repo-wide monkeypatch.undo() guard (6 other files) — still deliberately out of scope. Closes when a repo-wide PR merges or a named reader dismisses it here.
mutation: 22/22 killed attributed, P1 killed, problems none (K90 re-attributed after the KeyError-renders-no-message trap)
```

ZacxDev:
## Merged `main` — and the `runner_bound` red is gone. But it was `failed=2`, not 4, and the second one is `main`'s.

Merged current `origin/main` (`e8ee277e` at the time, not the `f013ac50` measured for me — it had moved again) into this branch: `fcd82015`, clean, **zero conflicts**.

### Overlap, and why I re-ran anyway

`git diff --name-only <base>..HEAD` vs `<base>..origin/main`: my side 4 files, `main`'s side **84**, **overlap 0**. That is not safety — `RULES.md` is explicit that *disjoint files are not safety* (one side widens a signature, the other adds a caller; both green, zero overlap, red on merge), and the trigger is that **the base moved**. So everything was re-run on the merged tree rather than reasoned about.

| check, on the merged tree | result |
|---|---|
| `test_runner_bound_ledger.py` (the previously-red one) | **5 passed** ✅ |
| the 8 affected suites | **785 passed** |
| mutation battery, 22 rows | **22/22 KILLED, all attributed** · P1 control KILLED · `problems: none` · `restore: OK` |

Your merge prediction was right and I confirmed it on **my** tree at the **current** main, not by trusting the `f013ac50` reading.

### 🔴 Two corrections to the CI read

**1. It is `failed=2`, not 4.** The gate log for `devrc-ci-ghhcc` says `collected=22590 passed=22586 skipped=2 failed=2`. The 4 came from subtracting without the 2 skipped.

**2. The second failure is not a picker test — it is `main`'s own breakage, and a re-trigger cannot clear it.**

```
FAILURES
_____ test_REAL_INTERACTIVE_fzf_opens_on_the_FIRST_INPUT_ROW_with_no_query _____
____________________________ test_no_new_dead_paths ____________________________
```

`test_no_new_dead_paths`:
```
1 NEW dead path reference(s) -- a doc claims a file that does not exist:
    claude/skills/audit-pr/SKILL.md:144  DEAD PATH: reference/round-ladder-evidence.md
```

🔴 **It reproduces byte-identically on a CLEAN `origin/main` worktree** — same file, same line, same assertion. **`main` is red on this right now.** It arrived with #1587 (*"cut 36% of what every --round 0 brief inlines"*), which edited that SKILL.md. My merge imported it; my diff does not touch `claude/skills/`.

Consequences worth acting on, none of them mine to fix here:
- The **main-green deadman should be firing** on this.
- **#1582 will hit exactly the same failure** — it has merged `main` too.
- I deliberately have **not** fixed it in this PR: it is another PR's regression, and patching it here would collide with whoever fixes it properly.

### On the fzf test — my read agrees with yours, with the evidence

- **Pre-existing**: present at my merge base `6ff4d215`.
- **Untouched**: my diff adds/removes **0** lines matching `_fzf_interactive_first_row|REAL_INTERACTIVE|pty\.fork|_PTY_PROMPT|_drain`.
- **Passes locally**: all 25 fzf-related tests green on the merged tree, the named one in **0.95s**.
- **The failure mode is a timeout, and the log narrows it**: fzf **did draw its prompt** (the earlier 20s assertion passed, or the message would have been *"fzf never drew its prompt in 20s"*). Enter was written; the selection pipe then returned `b''` inside a second 20s deadline, on `gw1` under xdist. The test's own message says it: *"the measurement did not happen, so neither did the assertion."*

Two 20s pty deadlines is a real timing dependence. Agreed that if it fails again it stops being a flake hypothesis — I have pushed an empty commit (`1012466a`) specifically to get that second observation, and I have said in its message that this **cannot** produce a green run while the dead-path failure stands.

⚠ So: **do not read the next run's red as a regression** — read the failing test names. Green is not reachable from this branch until `main`'s dead path is fixed.

ZacxDev:
## Re-triggered CI on `1012466a`: the fzf test PASSED. One failure left, and it is `main`'s.

| | run 1 — `fcd82015` | run 2 — `1012466a` |
|---|---|---|
| `devrc-pytests` | **failed=2** | **failed=1** |
| `test_REAL_INTERACTIVE_fzf_opens_on_the_FIRST_INPUT_ROW_with_no_query` | ❌ | ✅ **passed** |
| `test_no_new_dead_paths` | ❌ | ❌ (unchanged) |
| `devrc-nodetests` | pass 1449/1449 | pass 1449/1449 |
| `devrc-cairn-client-runs` | pass | pass |

### The fzf test: one hit, not two — so it does not escalate

It passed on the re-run. Per the standing condition, one hit is a flake hypothesis and two would be a finding; this is one. Independently, I measured it **20 consecutive standalone runs at load 84–92** (higher than CI saw) — **20 passed, 0 failed**, no load generator spawned.

⚠ **That does not "prove it fine", and I am not claiming it does.** Standalone runs cannot reproduce the condition it failed under: CI runs it through **xdist (`gw1`)** alongside 14,399 tests in one target. So what I have narrowed is that ordinary system load is not sufficient — the failure needs the concurrent-worker context. The timing dependence is real and still there (two 20s pty deadlines, `pty.fork` + `execvp("fzf")`), it is **pre-existing** (present at my merge base, and my diff changes **0** lines of its machinery), and it remains fixable by removing the timing dependence rather than by re-running. I have not done that here because it is not this PR's code and one observation does not justify touching it.

### The remaining failure is `main`'s, and this branch cannot go green until it is fixed

```
claude/skills/audit-pr/SKILL.md:144  DEAD PATH: reference/round-ladder-evidence.md
```
Reproduces byte-identically on a clean `origin/main` worktree. **`main` is red on this right now**, and #1582 will hit it too.

🔴 **The referenced file EXISTS** — `claude/skills/audit-pr/reference/round-ladder-evidence.md`, 40,916 bytes. The gate is objecting to how the path is *written*: a bare relative `reference/…` resolves against `<bare-ref-unopenable-write-the-deployed-~-.claude-path>`, i.e. the checker wants the deployed form. **The top of that very file already uses it** (`~/.claude/skills/audit-pr/reference/round-ladder-evidence.md`), so the fix is making line 144 match line 2.

### Your call — I have not made this change

It is one line and it unblocks `main`, this PR and #1582. But it is **#1587's file and #1587's regression**, and editing it from inside this PR risks colliding with whoever fixes it properly — exactly the kind of cross-PR edit I was told to stay out of for the `open_reference` seam.

**Recommendation:** fix it on `main` directly (or in #1587's follow-up), not here. Say the word and I will do it here instead; it is a two-minute change.

Everything this PR owns is green: **785 passed** across the 8 affected suites on the merged tree, and **22/22 mutants killed and attributed**, P1 control killed, `problems: none`.

ZacxDev:
## PR body rewritten — it had drifted, and one of its claims was the one I had already corrected in a comment

I was asked to record the fzf finding in the body rather than leave it in a task thread. On fetching the live body I found appending would have been the wrong move: several sections had gone stale across the audit rounds, including a claim I had **already publicly corrected**.

What was wrong in the body until now, and is fixed:

1. 🔴 *"`main` is independently RED … on two tests in `test_guard_core.py`"*, naming `test_no_tracked_shell_text_writes_a_kill_this_guard_would_deny`. **Wrong** — corrected in a comment hours ago but never in the body, so the body and the thread disagreed. Now removed entirely; the inherited-red story is over.
2. The dims table listed **6** fields. There are **10** — it was missing `reason`, `ordered`, `pinned_above` and `surface`, i.e. everything the audit rounds added.
3. Outcomes listed **3**; there are **4** (`no-selection` was missing).
4. The "seam deliberately left open" section named **3** expressions; the enumeration is **6**.
5. Mutation evidence said 13 mutants; it is **22**, across three sweeps that each found a real defect in this PR's own guards.
6. No `surface` section, no `dismissed`-vs-`no-selection` section, no fzf note.

The fzf note is now a first-class section — **notes for whoever fixes it**, including the part that matters most: the 20/20-at-load-84–92 result **does not prove it fine**, because a standalone run cannot reproduce the xdist `gw1`-beside-14,399-tests condition. What is narrowed is *"ordinary load is insufficient; it needs the concurrent-worker context"*, plus the log detail that fzf drew its prompt and the selection pipe then returned `b''` inside the **second** 20s deadline. The remedy is removing the timing dependence, not re-running.

Also recorded there: the K90 lesson, which generalises past this PR — **an assertion that subscripts the thing a deletion-mutant deletes can never be attributed**, because `KeyError` renders no message. `.get` is what makes such a guard mutation-testable.

Merge sequencing is stated in the body too: **#1569 first, then #1582**, and when #1582 lands the only edit needed here is the branch inside `open_reference`.

PR #1568, opened 2026-09-12
docs(gate-flake): rank 1 is MEASURED — 0/99 on heads carrying the sha, and the gate's red is no longer a flake

ZacxDev:
Rank 1 of `claudedocs/handoff-gate-flake-store-api.md` existed to produce one number — the flake rate against a baseline carrying `#1458`'s store-siting fix — and had never been run. It is run now. Closes `handoff-cairn-oss-multi-instance.md` **rank 22**, whose closing condition was exactly this reading.

## The reading

Predicate is **ancestry**, not date: `git merge-base --is-ancestor ce9b55c3 <head>`.

| window (ancestry) | verdicts | success | failure | error | pending | store-api test named |
|---|---|---|---|---|---|---|
| does **not** carry `ce9b55c3` | 298 | 192 | 64 | 41 | 1 | **12** (4.03%) |
| **carries** `ce9b55c3` | 99 | 29 | 37 | 29 | 4 | **0** |

`P(observing 0 | the pre-window rate) ≈ 0.017`, against **0.23** for the only prior reading (`#1512`, 4/125 → 0/45, split on the anchor's *timestamp*).

Population: 400 devrc PR heads `#1162`–`#1566`, every state; 397 carried a `tekton/devrc-pytests` verdict; **0 ancestry-unmeasurable**. Source is the GitHub combined-status endpoint — the only surviving record, since pipelineruns are pruned hourly (`keep: 20`). Heads came from `refs/pull/*/head` fetched into a **throwaway** repo, so the shared clone's ref namespace was not written to.

🔴 **Collector positive-controlled before its verdict was read**: `#1458`@`a6dd11eb`, `#1462`@`dc972398`, `#1454`@`c5e3123e` all come back `failure` **with a test named**, so a zero elsewhere is a reading and not a harness wired to nothing.

## Three things stated rather than glossed

**The zero is not what establishes the fix.** The mechanism being gone is, and that is content-checked on `origin/main`: the class and test are present (`:14543`), `sited_root` is on 106 lines, and the 1 surviving `tmp_path / "store"` is prose. The 0 therefore is not a deleted test.

[...]

PR #1570, opened 2026-09-12
docs(gate-flake): #1568 asserted a count it got wrong and a window that cannot be falsified — plus the retirements it owed

ZacxDev:
Follow-up to **#1568**, which merged **6 minutes** after it opened — so its round-0 audit landed **10 minutes after the merge**, and every candidate it found is a follow-up rather than a change to that PR. Round 0 re-verified #1568's self-reported numbers and **every checkable one verified exactly**: `sited_root` on 106 lines, the surviving `tmp_path / "store"` at `:449`, the class at `:14543`, `ce9b55c3`/`c0bbd6d9` both real, rank 8's subject real, the claim genuinely held. This PR fixes the one that did not, and pays the retirements.

## The claim that was wrong

> *"`main`'s last **8** commits carry no completed `tekton/devrc-main-pytests` verdict: **six** `superseded…`, two `NO GATE POD`, newest pending."*

Re-derived over the **named range** `457a5dc7~1..8114a124`: **four** superseded, two `NO GATE POD`, one **`KILLED`**, one pending. The headline — zero completed verdicts — held; the breakdown did not.

🔴 **The frame was worse than the count.** "The last N commits" is a moving window, and `main` took two commits between writing the sentence and checking it. It was unfalsifiable by construction. **Name a sha range or do not quote a window.**

## …and it was a worse duplicate of a sibling doc

That paragraph re-measured at n=8, badly, something `handoff-gate-speed-and-ci-signal.md` already owns at **n=100** with the decomposition (`:54`), and split on this very fix (`:363`: *"of 43 post-fix verdicts on `main`, **40 were artefacts** … leaving **3**. Use PR heads."*). `KILLED` is not an undiagnosed mode either — the four artefact strings are a **shipped mechanism** (`handoff-cairn-task-linkage.md:37`). Replaced with a pointer to both.

## Retirements the measurement owed

| site | was | now |
|---|---|---|
| gate-speed rank 3 | rested on `#1512`'s timestamp split (P≈0.23) | names the ancestry read (P≈0.017), while stating `#1512` was *underpowered, not corrupted* |
| gate-speed rank 9 | `forcing: gate`, two options open | **CLOSED** by `#1561`, with rate evidence; successor named (`test_runner_bound_ledger.py`, live) |
| gate-speed rank 6 | "probably inert — decide on/after 2026-10-11" | gets the datum to decide **on**, without pre-empting the date |
| cairn-oss rank-22 bullet | retained an **instruction** to go run the finished work | struck per `supersede.md`; instruction deleted, unique provenance shas kept |
| gate-flake rank 1 | `forcing: gate` | `forcing: none` — its own payload (0 of 99) disproves it; the forcing claim *moved* to ranks 8/9 |
| the retained "superseded text" | justified as "a caveat still true of any FUTURE baseline" | justification **withdrawn** — it is a warning about one past baseline; the number is kept |
| `State now` | `main @ 65f7325b`, two merges stale | **no tip sha at all** — see below |
| open investigation | *"tmpfs siting … UNMEASURED"*, still prescribing the probe | CLOSED, with its residual stated: the hypothesis's *second* half is still unmeasured |

🔴 **The tip-sha line is the self-demonstrating one.** I replaced `65f7325b` with `8114a124` — and it was stale **before it was pushed**, because `main` took `#1564` in the interval. So the line is gone rather than corrected; every sha in the doc is now pinned to the commit it is *about*, never to "the tip".

## One residual narrowed — measured, not argued

Whether the gate container has a usable tmpfs was "the first thing to check". `store_siting.py:221` sets `_MIN_FREE_BYTES = 4 MiB`, `:67` makes `/dev/shm` the only candidate, and the gate pipeline (2,482 lines) has **0** occurrences of `shm` and **0** of `medium:` — positive controls on the same read: `pytests` 59, `emptyDir` 3. So nothing overrides `/dev/shm` for the `pytests` step. ⚠ Stated in two halves because only one is a reading of *this* cluster: "nothing overrides it" is measured; "therefore 64 MiB" is the runtime default, and **no node was read**.

## New rank 9 — and it is not new

`#1508` deleted `scripts/lib/subsystem_recall.py`; **61** `claudedocs/` files still prescribe running it (**65** invocations). The bare path appears 71 times in 63 files, which is why the closing condition is keyed to the **invocation** — otherwise the item makes its own condition unsatisfiable. Six source references, all comments, none broken; the Dockerfile's is already correct. ⚠ An earlier count said **5** and missed the Dockerfile, because the sweep was an extension-filtered `find` rather than a `git grep` over the ref.

🔴 **The index has carried this as an `OPEN:` bullet since 2026-09-01 naming this exact remedy**, and `cairn recall` surfaced it at resume time. Rank 9 points there instead of presenting it as new.

## Outside this diff

The index entry `devrc/subsystem-store-api` still stated the verifier as **OPEN** against the 2026-09-01 baseline — and the index is read **before** the doc, so the next session met the opposite of the truth first. Corrected by `cairn append` (revision `68a1bc6fb145b460`). Round 0 called this its highest-value single finding.

## Round-0 trial ledger — transcribed verbatim, as its owner requires

**`ran: 6 · changed the outcome: 3`**

This dispatch makes `R=7` and **cannot move `C`**: it was a post-merge ordinary trial, which `handoff-audit-pr-ladder.md:31` already diagnosed as the cause of every zero, and `:116` says not to run more of. The remaining work is the **trigger** (**#1565**), not more trials. Recorded here rather than quietly skipped.

## Verification

Three `claudedocs/` files, prose only. `test_handoff_audit.py` + `test_handoff_doc.py` + `test_handoff_index.py` + `test_doc_path_rot.py`: **802 passed**. ⚠ Those gates parse headings but **do not validate path citations** — proven in #1568 by inserting a bogus path and watching them stay green — so citations here were checked by hand, which is how the `:236-239` → `:232-239` and `:365` → `:363` corrections were found.

Rank 8's red re-measured at **three** successive `main` tips (`4e970998`, `8114a124`, `61f41adf`) because the base moved twice under this work. `#1567` still `OPEN`.

https://claude.ai/code/session_018rg3dNxWP65JYxjRujF23a

### DIFF
--- a/claudedocs/handoff-cairn-oss-multi-instance.md
+++ b/claudedocs/handoff-cairn-oss-multi-instance.md
@@ -2097,11 +2097,17 @@ covers; pin it with `--config`, do not `cd`.
   those AFTER it** — the same design class in a second ledger, and the live one). Table and
   residuals in `handoff-gate-flake-store-api.md` rank 1.
   The pre-verification wording, kept for the provenance it names:
-- 🔴 **RANK 22 CLOSED-PENDING-VERIFICATION, 2026-09-11.** `#1458` squash **`ce9b55c3`** merged and
-  content-verified; recorded by `#1462` (`60033d1e`) and corrected by `#1525` (`018e483b`). The
-  second flake it uncovered is filed as **`handoff-gate-flake-store-api.md` rank 7** (`#1477`,
-  `50e8a71a`), which also corrects that doc's rank 1. **The verifier is the flake RATE against
-  heads that CARRY `ce9b55c3` — `git merge-base --is-ancestor`, never a date comparison.**
+- ~~**RANK 22 CLOSED-PENDING-VERIFICATION, 2026-09-11.**~~ **SUPERSEDED 2026-09-12 — the
+  verification it asks for HAS BEEN RUN; see the bullet directly above.** Retained only for the
+  provenance shas it uniquely names: `#1458` squash **`ce9b55c3`** merged and content-verified;
+  recorded by `#1462` (`60033d1e`) and corrected by `#1525` (`018e483b`). The second flake it
+  uncovered is filed as **`handoff-gate-flake-store-api.md` rank 7** (`#1477`, `50e8a71a`), which
+  also corrects that doc's rank 1. 🔴 **Its closing sentence — an INSTRUCTION to go run the
+  flake-rate read — is DELETED rather than preserved**, per
+  `claude/skills/handoff/reference/supersede.md`: keeping a corrected *reading* is the point,
+  keeping a corrected *instruction* arms a landmine for whoever greps `rank 22` in a 2,200-line doc
+  and lands on this hit instead of the bullet above. The predicate it stated was correct and
+  survives above — ancestry, never a date.
 - 🔴 **`cairn recall --repo <cairn>` is `scope-absent`; the scope is `devrc`.** The OSS repo has no
   store scope. `cairn search --scope devrc '<term>'` is what surfaced `ci-repro/` and the
   `#1211`/`#1219`/`#1239` history that made rank 22's whole diagnosis possible.
--- a/claudedocs/handoff-cairn-oss-multi-instance.md
+++ b/claudedocs/handoff-cairn-oss-multi-instance.md
@@ -2094,10 +2094,11 @@ covers; pin it with `--config`, do not `cd`.
   worth carrying forward: the gate's remaining red is not a flake.** Deterministic ledger
   censuses over tracked text account for **27 of 99** post-fix verdicts (22 the kill-mention
   ledger, all of them before `#1561` exempted `claudedocs/`; **5 the runner-bound ledger, 4 of
-  those AFTER it** — the same design class in a second ledger, which is the THIRD instance in two
-  days). ⚠ **Both instances are now fixed** (`#1561` `c0bbd6d9`, `#1567` `6f1867b1`); this bullet
-  called the second "the live one" and that was true for about an hour. **The CLASS is what
-  survives: nothing stops a fourth census reddening `main` for everyone.** Table and residuals in
+  those AFTER it** — the same design class in a second ledger, the SECOND instance enumerated). ⚠ **Both instances are now fixed** (`#1561` `c0bbd6d9`, `#1567` `6f1867b1`); this bullet
+  called the second "the live one" and that was true for about an hour. ⚠ **It also called this the
+  THIRD instance while saying "both instances are now fixed" two lines later; only TWO are
+  enumerated anywhere** (rank 1 of the gate-flake doc). **The CLASS is what survives: nothing stops
+  the NEXT census reddening `main` for everyone.** Table and residuals in
   `handoff-gate-flake-store-api.md` rank 1.
   The pre-verification wording, kept for the provenance it names:
 - ~~**RANK 22 CLOSED-PENDING-VERIFICATION, 2026-09-11.**~~ **SUPERSEDED 2026-09-12 — the
--- a/claudedocs/handoff-cairn-oss-multi-instance.md
+++ b/claudedocs/handoff-cairn-oss-multi-instance.md
@@ -2094,8 +2094,11 @@ covers; pin it with `--config`, do not `cd`.
   worth carrying forward: the gate's remaining red is not a flake.** Deterministic ledger
   censuses over tracked text account for **27 of 99** post-fix verdicts (22 the kill-mention
   ledger, all of them before `#1561` exempted `claudedocs/`; **5 the runner-bound ledger, 4 of
-  those AFTER it** — the same design class in a second ledger, and the live one). Table and
-  residuals in `handoff-gate-flake-store-api.md` rank 1.
+  those AFTER it** — the same design class in a second ledger, which is the THIRD instance in two
+  days). ⚠ **Both instances are now fixed** (`#1561` `c0bbd6d9`, `#1567` `6f1867b1`); this bullet
+  called the second "the live one" and that was true for about an hour. **The CLASS is what
+  survives: nothing stops a fourth census reddening `main` for everyone.** Table and residuals in
+  `handoff-gate-flake-store-api.md` rank 1.
   The pre-verification wording, kept for the provenance it names:
 - ~~**RANK 22 CLOSED-PENDING-VERIFICATION, 2026-09-11.**~~ **SUPERSEDED 2026-09-12 — the
   verification it asks for HAS BEEN RUN; see the bullet directly above.** Retained only for the
--- a/claudedocs/handoff-gate-flake-store-api.md
+++ b/claudedocs/handoff-gate-flake-store-api.md
@@ -8,11 +8,11 @@ cairn recall --repo ~/workspace/devrc
 reader into the pinned package, so the old spelling now fails with a file-not-found. **It is not
 just this doc: see rank 9 for the population, which is the single place those counts live.** Only
 this doc's invocation is fixed here; fixing one site and implying the rest is the partial-sweep
-failure. ⚠ **This block previously restated the counts and got two of them wrong in the direction
-rank 9 exists to prevent** — `63`/`71` are the BARE-PATH figures, not the *prescribing* ones
-(`61`/`65`), and it listed `5` source files while rank 9, in the same commit, said `6` and named the
-Dockerfile the list had dropped. **Two copies of a count are how one of them becomes wrong; rank 9
-owns them.**
+failure. ⚠ **This block deliberately carries NO counts.** It previously restated them and got two wrong in
+the direction rank 9 exists to prevent — quoting the bare-path figures for *prescribing*, and listing
+five source files where rank 9, in the same commit, said six and named the Dockerfile the list had
+dropped. **A second copy of a count is how one of them becomes wrong, and the fix is to have one
+copy, not two agreeing ones: rank 9 owns these.**
 Terse pointers this doc does not carry, curated by past sessions and outliving it.
 🔴 RECALL, NOT LIVE OBSERVATION — every line is a pointer to VERIFY, never a current
 reading, and it may describe a gotcha already fixed. `scope-absent`/`scope-empty` means
@@ -30,10 +30,17 @@ number, and the number relocates the problem.** The store-api fsync flake is nam
 `tekton/devrc-pytests` verdicts on heads that CARRY `ce9b55c3` against **12 of 298** that do not
 (P(0) ≈ 0.017), measured by **ancestry**. 🔴 **And what the same read found instead: the gate's
 remaining red is dominated by DETERMINISTIC ledger censuses over tracked text — 27 of 99 post-fix
-verdicts, ~7× this flake at its worst — of which 22 were closed at the source by `#1561` and
-5 are the runner-bound ledger, on which `main` IS RED RIGHT NOW — already owned by `#1567`.**
-That is the new rank 8. Full
-table, controls and residuals in rank 1; claim `gate-flake-store-api-1` was held for the read.
+verdicts, ~7× this flake at its worst — of which 22 were the kill-mention ledger, closed at the
+source by `#1561` (`c0bbd6d9`), and 5 the runner-bound ledger, closed by `#1567` (`6f1867b1`).**
+Both reddened `main` itself; both are fixed, confirmed at `origin/main` `337114e0` (**5 passed** on
+`test_runner_bound_ledger.py`, **1 passed** on the kill guard). 🔴 **The CLASS is not fixed: at least
+two instances in two days, and nothing prevents the next.** That is rank 8 — closed as an instance,
+retained for the class. ⚠ **The 27 / 22 / 5 figures are a READ-TIME population that cannot be
+re-derived** (rank 1 says why); do not treat a later disagreement as a refutation.
+⚠ **This paragraph said `main` IS RED RIGHT NOW for about an hour after `#1567` merged**, while rank
+8 five hundred lines below already said CLOSED — and `State now` is the block a resuming session
+reads first. **Four other sites were swept for that phrasing and this one, the most-read, was
+missed.** Full table, controls and residuals in rank 1; claim `gate-flake-store-api-1` released.
 
 🔴 **THE AUDIT LADDER IS CLOSED — by decision, not by a clean round.** `#1219` and its
 successor `#1239` are both MERGED. The ladder ran **8 rounds on `#1219` + 2 on `#1239`**;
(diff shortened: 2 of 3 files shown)


ZacxDev:
## Round 1 — 3 🔴 / 4 🟡 / 2 nits, all fixed in `551a58b3`

Every finding was re-verified first-hand before fixing, not taken on the auditor's report. All of them were in prose this PR exists to correct — the skill's "a false claim replaced by a differently false one" shape, confirmed.

```audit-claims round=1 audited=cf4c87a1..551a58b3
1. Rank 9's closing condition was unsatisfiable while asserting it was not — the item discusses the dead path using the literal invocation the grep keys on (3 occurrences, grep returned 67). Now carries :(exclude) for this file; measured satisfiable, 64 across 60 files remain.
2. The corrected verdict census was wrong again and stale before commit — FIVE superseded not four, and ZERO pending; 8114a124 flipped pending->superseded at 04:32:10Z, 24 min before the commit was authored. Root cause recorded: a sha range pins COMMITS, not their mutable VERDICTS.
3. Two cross-doc line citations were invalidated by this PR's own +25-line insertion (:363 -> :388). All cross-doc citations are now quoted by OPENING WORDS, never by line number.
4. Five "main is red RIGHT NOW" claims across three docs went false when #1567 merged (6f1867b1, 05:07:06Z) — including a How-to-verify recipe reading `expect 1 failed`. Rank 8 closed with its condition run, not assumed: 5 passed. Rank 6's keylog expectation was stale too (104 passed).
5. The top block restated rank 9's counts and got two wrong in rank 9's own failure direction (bare-path figures used for "prescribe"; 5 source files where rank 9 said 6). Counts now live in rank 9 only.
6. "Its grep -c classifier names the right test families" was false — it greps descriptions, which never carry a filename, so it matches 0 of the 12 store-api rows that ARE rank 1's pre-window count. Positive control: 4 of 101 match, so it is mis-tokenised rather than wired to nothing.
7. The tmpfs read overstated its coverage three ways: /dev/shm is the DEFAULT candidate not the only one (DEVRC_TEST_TMPFS first, :270); the pytests pod spec also lives in devrc-ci-triggertemplate.yaml, unread; the manifest ref was unpinned and has moved (2,482 -> 2,584 lines, emptyDir 3 -> 12). Conclusion re-measured across both manifests and holds.
8. The orphan "doc-surgery task with a real closing condition" aside named no doc, owner or mechanical check — now stated plainly as an observation, not dressed as a work item.
9. Every number in the 27/99 split is marked read-time-only and non-re-derivable: a re-read hours later gave 20 and 9/8 rather than 22 and 5/4, same direction, original rows destroyed by superseding runs.
```

**Verification:** 802 passed on `test_handoff_audit` + `test_handoff_doc` + `test_handoff_index` + `test_doc_path_rot`. ⚠ Those gates parse headings but **do not validate path citations** — proven in #1568 — so citations were checked by hand again, which is how findings 1, 3 and 7 were confirmed.

**Ledger:** round 1 · payload lines changed THIS round: 146 (all three files are `claudedocs/*.md`, and for a docs PR the `.md` **is** the payload — decided at round 1 and recorded here so it cannot be re-decided later) · scaffolding: none, this PR ships no tests.

ZacxDev:
## Round 2 (delta) — 2 🔴 / 5 🟡 / 4 nits, all fixed in `e09956d1`

Rebased onto `origin/main` `337114e0` in the process; `gh` now reports **MERGEABLE / CLEAN** (it was `CONFLICTING / DIRTY`).

```audit-claims round=2 audited=551a58b3..e09956d1
1. The PR was CONFLICTING — main's c0325f83 rewrote this sibling doc's State now and ranked list, conflicting in rank 9, the hunk this work wrote. Resolved by READING the merged result: main's rank 9 tombstone and its "main IS GREEN AGAIN" (verified on 6f1867b1, 20 passed) are better than mine and were kept; only the rate evidence and the class point survive from my side; main's rank 10 rewrite taken whole. gh now reports MERGEABLE/CLEAN.
2. State now still said "main IS RED RIGHT NOW" 528 lines above the rank 8 the same commit marked CLOSED — the most-read block, missed by a sweep that covered four other sites. Fixed, and the miss is recorded in place.
3. Rank 6 kept `forcing: regression — main is red` on a suite measured green (104 passed). forcing: is what a claim-work session sorts on, so the stale value advertised a gate-forcing regression with nothing behind it. Rank 6 closed and the forcing retired in the same edit.
4. Four new measurements were attributed to a bare `origin/main` in the commit that writes the rule against unpinned reads. All four now name 337114e0.
5. "pytests 59 -> 62" mixed a line count with an occurrence count (real movement +1) and attributed the post-move figures to a moving origin/trunk. Now one metric, pinned to 3c53d618.
6. "the third census in this family" was 1 unsourced site before this work and 6 after, while only 2 instances are enumerated anywhere — and cairn-oss said "both instances are now fixed" two lines after calling one the THIRD. Now "at least two, both enumerated in rank 1".
7. The classifier paragraph asserted 0 and 4 about the same population. The 12 rows are the FSYNC-FLAKE rows; the 4 matches lie in the 21 rows naming any class from that file. "0 in both arms" was wrong — pre-window returns 4, post 0, which is worse than a flat zero. And "a description carries Class.test_name" describes 27 of 100 rows; 73 are bare, so adding the missing class token does not fix it either.
8. Two "quoted by opening words" citations matched NOTHING (an ASCII hyphen for an en-dash, two missing backticks) — the same rot as the line numbers in a new spelling. Both now match exactly once; old spellings confirmed at 0.
9. "56 minutes" was 1 h 08 m 32 s, contradicting the same item's "1 h 10 m". Both now 1 h 08 m.
10. The closing condition now states what it COVERS (60 of 61 docs, structurally blind to this one) rather than implying "no doc prescribes the dead path"; both-direction controls re-confirmed at the rebased head.
11. The top block now carries NO counts rather than a second agreeing copy of rank 9's.
```

### Ledger, and a correction to round 1's

```
round 2 · payload lines changed THIS round: 103 (66+/37−, claudedocs/*.md = payload by round 1's recorded decision; scaffolding 0) · elapsed: ~45 min
```

🔴 **Round 1's ledger said `146` and that number is not re-derivable** — the prescribed command gives **168+/80−** over `cf4c87a1..551a58b3` (sum 248, net 88), and 146 matches none of added, deleted, sum or net. Round 2 caught it; recording the correction rather than quietly restating. Round 2's own figure above is `added+deleted` from the same command over `4466b379..e09956d1`, rc 0, range non-empty (1 commit), stderr clean — named so the next round can check it the same way.

**The gate does not fire:** both rounds changed payload, so the two-zero-rounds rule is not in play.

**Verification:** 802 passed on the four doc-gate files at the rebased head. Those gates still do not validate citations, which is why findings 8, 5 and 7 were found by hand.

ZacxDev:
## Round 3 (delta) — 2 🔴 / 5 🟡 / 6 nits, all fixed in `70dcf5a5`

Both blockers were created by **round 2's own fix**. The first is the most serious defect this ladder has produced, and I reproduced its control before fixing.

```audit-claims round=3 audited=e09956d1..70dcf5a5
1. Rank 6's closure cited "the keylog suite -> 104 passed" as "the criterion this item itself names as the deterministic check". That suite cannot see nix/home.nix: controls on a cp -a copy (.git removed first) give 104 passed with the collision re-introduced in full, and 104 passed with nix/home.nix DELETED OUTRIGHT. A vacuous green used as a closing condition. Rank 6 now cites the direct config read (nix/home.nix:419/:431, disjoint) and records that the suite is blind BY DECISION — #1265 (68d10b19) dropped the 10 live-config guards, so no automated check covers this interface.
2. Closing rank 6 left its 30-line Open-investigations block present-tense — "This is a REAL user-facing regression" plus a "Next probe, verbatim" telling the next session to go fix it — in the same commit that wrote "a verify-block expectation must track the closing condition". That is round 2's own finding #2, one rank over. The block also cited a test #1265 DELETED, at a line that is now an unrelated docstring. Closed, both errors kept as the worked example.
3. A bash comment still said "rank 6 itself still says main IS RED" after rank 6 said the opposite — the very comment round 2 reasoned about and did not edit.
4. "the THIRD instance in two days" survived in cairn-oss, two lines above "both instances are now fixed", while the sibling doc told readers not to write a third. Claim 6 was fixed at two of three sites; now three of three.
5. "1 h 08 m" was #1567's open-to-merge window pasted onto a DIFFERENT quantity. Filing to closed is 57 m 04 s (rank 8 filed in 029e801c, authored 04:10:02Z). A third site still said 70 minutes; now all consistent.
6. One measurement still read "at origin/main" unpinned, 32 lines above a pinned copy of the same number. Pinned, and the pre-merge side now names 61f41adf.
7. The new rank-6 evidence block in gate-speed quoted read-time-only figures without the caveat rank 1 mandates, and called the 140-byte cap 138. Both fixed.
8. Swept the doc's OTHER closing conditions for finding 1's defect: rank 7 already does it right and says so; rank 8's was run; rank 9's was exercised in both directions. One instance, not a pattern — swept and recorded anyway, because a defect at one site is a hypothesis about the others.
```

### Ledger

```
round 3 · payload lines changed THIS round: 100 (67+/33−, claudedocs/*.md = payload by round 1's decision; scaffolding 0) · elapsed: ~35 min
```

✅ **Round 2's ledger figure re-derived and it reproduces exactly** (66+/37− = 103 over `4466b379..e09956d1`, rc 0, 1 commit) — so the ledger defect round 2 found in round 1 did **not** recur. Round 3 also corrected a scaffolding claim of mine: the rebase landed on **`1bf1fa3d`**, not `337114e0`. `337114e0` is the sha the *measurements* are pinned to, which is a different thing and is correct.

**The gate does not fire** — every round changed payload.

### Two things I am flagging rather than claiming

**The status-derived figures cannot be verified by anyone, now or later** — 0/99, 12/298, 27/99, 22, 5, 27/100, 73, 12→0, 21→4, 100/101. GitHub keeps one status per context and supersedes overwrite it. Round 3 confirmed internal consistency (22+5=27, 27+73=100) and that rank 1 declares the population non-re-derivable; it did not confirm the values, and said so. The docs now carry that caveat wherever the numbers appear.

**This ladder has run three rounds on prose and every round found real defects in the previous round's fix.** Round 3's two 🔴 were both mine. The shape is now swept at every site I can find — present-tense reds, unpinned readings, duplicated counts, line-number citations, and closing conditions resting on blind instruments — so if round 4 comes back with only reword-class findings, that is the stopping criterion for a prose ladder and I will say so explicitly rather than leaving it implicit.

ZacxDev:
## Round 4 (delta) — 1 🔴 / 1 🟡, fixed in `06f81c0d`

Round 3 replaced a vacuous green with a **vacuous read**. Round 4 caught it; I verified across four trees before fixing.

```audit-claims round=4 audited=70dcf5a5..06f81c0d
1. Rank 6's replacement evidence — "a DIRECT READ of :acq/:rna's search_terms — disjoint" — is vacuous: measured with `git show <ref>:nix/home.nix`, :acq's search_terms are ["ask" "clarify" "clarifying" "questions"] on a720d30d, a451abc0^, a451abc0 AND 337114e0. Byte-identical on the red trees and the green one; they never contained recom/recommend. The prescribed read returns "disjoint — closed" on the tree the doc calls RED.
2. The real fix is a451abc0 swapping :acq's `replace` and `label` — the phrase moved OUT of label, and _token_matches (espanso_detect.py:440-452) reads trigger + label + search_terms and never `replace`. a720d30d put it in the label. The repo already said so at test_espanso_detect.py:728-732 and espanso_detect.py:350-357, which no earlier round read. Rank 6 now names the label as the discriminating field.
3. The class is ALSO closed in code, independently of the config: _attribute's declared-interface precedence (espanso_detect.py:349-353), guarded by four named tests. So the residual is narrower than "this interface is unguarded" — what nothing covers is a genuine DECLARED search_terms collision, which resolves to None and leaves the suite at 104 passed. Scoped accordingly.
4. The hypothesis was labelled CONFIRMED. Both its arms were refuted — neither narrowing :acq's terms nor disambiguating :rna's happened; the fix was a third option outside the fork. Now marked REFUTED IN BOTH ARMS, with "do not mark a hypothesis confirmed from its outcome".
5. Round 3's closing-condition sweep reported "rank 6 was the only one resting on an instrument that cannot observe its subject" — and exempted rank 6's own replacement evidence, which was the defect's new home. The sweep's blind spot is now recorded in place: when you fix a vacuous check, put the replacement through the sweep too.
6. "TWO THINGS IN THIS BLOCK ARE NOW WRONG" was a closed count that was short. Two further present-tense claims stood unstruck ("both recom and recommend now resolve to nothing"; "This is a REAL user-facing regression"). Driving the real _attribute against configs rebuilt from the actual nix/home.nix lines, the RED-shaped config resolves recom to :rna anyway — neither claim describes a reachable state. Now FOUR, enumerated, both new ones marked at their own sites.
7. #1265's title says "the 10 live-config guards" while the commit deletes 9 def test_ functions. Quoted as the commit's own count, flagged rather than silently corrected.
```

### Ledger

```
round 4 · payload lines changed THIS round: 109 (75+/34−, claudedocs/*.md = payload by round 1's decision; scaffolding 0) · elapsed: ~30 min
```

Round 3's figure re-derived by round 4 and **reproduces exactly** (100 = 67+/33−). The ledger has been stable since round 2 caught round 1's.

### On whether this ladder should continue

Round 4 returned a 🔴, so by the findings-keyed rule round 5 follows, and I am running it. But I want to name the pattern rather than discover it again: **the last two generations of defect have both been in the espanso archaeology attached to rank 6, which is not what this PR is for.** Rank 6 needed one thing — a false `forcing: regression` retired. Everything beyond that is history I wrote while sweeping, and it has now been wrong three times in three different ways.

So, stated in advance as a decision rather than taken as a reaction: **if round 5 finds another defect in that same espanso material, the fix is to CUT it to the one-line retirement, not to correct it a fourth time.** The label-vs-`search_terms` trap is worth keeping — it goes in the index, not in a ranked item's body. The rest is `supersede.md`'s "if a sentence cannot be made true and precise, delete the claim."

What has never been challenged across four rounds: the measurement in #1568 itself.

ZacxDev:
## Round 5 (delta) — 2 🔴 / 2 🟡. **Acted on by CUTTING, in `c65bde28`**

The trigger I pre-committed to in round 4 fired, and round 5 independently recommended the same cut on its own reasoning. Both 🔴 verified first-hand before acting.

```audit-claims round=5 audited=06f81c0d..c65bde28
1. The fix attribution was wrong for the THIRD consecutive round, in the same direction each time (crediting a config edit over a code change). Measured by driving the real _attribute against the RED-shaped config while varying ONLY the detector code: a720d30d -> None (negative control), b9b2493d (#1247, 09-02 17:16) -> ':rna', de677683 (#1252) -> ':rna', a451abc0 -> ':rna', origin/main -> ':rna'. So #1247 fixed it in the DETECTOR 1 h 39 m after the break; a451abc0 — credited by rounds 3 AND 4 — landed 23 h later, after the collision was already fixed, and is what #1265's own body says reddened main next. The answer sat two lines past where round 4 stopped reading (test_espanso_detect.py:734).
2. "guarded by four tests" named four tests that cannot observe the mechanism. Deleting the precedence block outright: the four named tests 4 PASSED; the full suite 5 failed. The five real guards are named nowhere in the doc, and one of the four carries an in-source instruction at the cited line — "AN INVARIANT GUARD, NOT REGRESSION COVERAGE ... Do not count it as evidence the ordering is tested" — which the doc counted anyway. A named, resolving, passing test that stops anyone looking is worse than the vague claim it replaced.
3. The residual was scoped too narrowly: precedence is INERT when no candidate declares the term, so a label-only collision still resolves to None — two routes, not one, and the second is the cosmetic-reword class the block is about.
4. "FOUR THINGS IN THIS BLOCK ARE WRONG" was a short closed count for the third round running (the heading itself still said "the terms are disjoint"; a struck bullet's tail stood present-tense).
5. RESOLUTION: all of the above was CUT rather than corrected. Investigation block 65 -> 29 lines; rank 6 49 -> 8. What survives is only what is mechanically checkable, and it was re-measured rather than carried forward: recom -> ':rna' by driving the detector against the live config (matched=[':rna']); fixed by #1247 then #1252 in code; the narrowed residual. The durable lesson moved to the index under devrc/espanso-audit, where it loads on trigger instead of on every read of this doc.
6. One surviving citation was widened: espanso_detect.py:344-357 -> :344-372. The old range covered the rationale comment but stopped short of the code it describes.
```

### Ledger

```
round 5 · payload lines changed THIS round: 149 (36+/113−) — the only NEGATIVE-net round
of the ladder · scaffolding 0 · elapsed: ~40 min
```

Round 4's figure re-derived by round 5 and reproduces exactly (109 = 75+/34−). **Stable for four consecutive rounds** since round 2 caught round 1's.

### Why this ends differently from "fix and re-audit"

Four rounds, four replacements, each written more carefully than the last, each wrong in a new way: a suite that cannot read the file → a field identical on the red and green trees → the wrong commit → four blind guards. **The error rate did not fall.** That is the signature of a claim nobody can satisfy, not of insufficient diligence.

Rank 6 needed exactly one thing — retire a false `forcing: regression` — and that is mechanically checkable. The archaeology had **no closing condition at all**, so it could not be audited to a stop; its correctness condition was "every sentence is true", which is unbounded. Meanwhile the record it was duplicating already lives where it stays correct as the code moves: `espanso_detect.py:344-372` and `test_espanso_detect.py:727-757`. "One rule, one place" — the second copy regenerated the same bug four times.

🔴 **One thing the ladder established that is worth more than the espanso history:** in this repo the doc gates are blind to **paths**, not merely to line numbers. Round 5's control injected a nonexistent path into this payload file and the gate stayed green. So every citation in any `claudedocs/` file is unverified by construction — which is why five rounds of hand-checking found what 802 passing tests could not see.

**Next:** round 6 on `c65bde28`. The cut is itself an edit to the material that was wrong four times, so it gets audited like any other — but its surviving claims are now two measurements and two citations rather than a narrative.

ZacxDev:
## Round 6 (delta) — **CLEAN. The ladder ends here.**

0 🔴 · 0 🟡 · 1 🟢. Round 6 re-derived every surviving claim of the cut first-hand: the attribution matrix **arm-for-arm**, both citations on `origin/main` as well as the PR head, each blindness claim with a watched positive control, all three self-reported counts (149 · 65→29/49→8 · 802), the index bullet genuinely holding the durable lesson, and **no dangling reference anywhere in the repo**.

🔴 **A clean round ends the ladder, and I am not running a round 7 to confirm it.** Per `claude/RULES.md`: *"A CLEAN round ENDS the ladder — never run another to confirm one."*

### The one 🟢, fixed in `d1110fb4` — it was a real precision defect

*"Fixed in the DETECTOR, not the config"* was a **one-armed measurement**: round 5 varied only the detector code. Round 6 ran the other arm, and I reproduced it:

| detector | config | `recom` |
|---|---|---|
| `a720d30d` (pre-fix) | RED | **None** ← control |
| `a720d30d` (pre-fix) | **today's** | `:rna` |
| today's | RED | `:rna` |

The closure is **over-determined** — the config edit is independently sufficient. So *"the code, not the config"* was a claim about **order** dressed as one about **sufficiency**. Both sites now say "closed twice over, independently; the detector fix landed 23 h first", and the measured matrix is in the doc. Also took two items off round 6's reword list: the block header claimed "FOUR TIMES IN FOUR AUDIT ROUNDS" while its own enumeration labels three, and one timestamp lacked a zone.

### What the ladder cost and what it bought

| round | findings | payload |
|---|---|---|
| 0 | requirement questioned + the index contradiction | — |
| 1 | 3 🔴 / 4 🟡 | 248 (168+/80−) |
| 2 | 2 🔴 / 5 🟡 | 103 (66+/37−) |
| 3 | 2 🔴 / 5 🟡 | 100 (67+/33−) |
| 4 | 1 🔴 / 1 🟡 | 109 (75+/34−) |
| 5 | 2 🔴 / 2 🟡 | **149 (36+/113−)** — the cut |
| 6 | **clean** | — |

Every round's ledger figure was re-derived by the next round and reproduced exactly from round 2 onward; round 1's stated `146` was unreproducible and is superseded by `248`.

🔴 **Not one finding across six rounds was in the measurement this PR exists to record.** All of them were in prose *about* it — attribution, counts, citations, closing conditions, present-tense claims about mutable state. That is the durable result, and it is why the two `🔴 the doc gates are blind to PATHS in claudedocs/` controls matter more than anything else here: 802 tests pass on every revision of this file, including ones asserting a nonexistent path, a wrong fix commit and four blind guards.

**Ready to merge.** `MERGEABLE / CLEAN`.

PR #1576, opened 2026-09-12
feat(audit-pr): classify gap commits by the round they DECLARE — 40 UNLEDGERED rounds, split interior/tail

ZacxDev:
Follow-on filed by #1564. Claim: `ladder-tail-round-classifier`.

⚠ **This body was rewritten after rounds 0 and 1.** An earlier version headlined **40** and claimed this "supersedes sampling with a census" — both retracted by the diff's own later commits. The correction history is in the two audit comments below; this text is the current claim.

## The signal

#1564 found, as a by-product, that tail fixes often **name the audit round they belong to in their own commit subject**. That is machine-checkable, so the classifier lists the commits contributing each GAP's churn and buckets them.

## What it asserts — and the number to quote is 4, not 40

| repo | gap commits | INTERIOR (r·m·u) | TAIL (r·m·u) |
|---|---|---|---|
| devrc | 69 | **1** · 0 · 1 | 10 · 23 · 34 |
| homelab-talos | 79 | **3** · 0 · 2 | 13 · 17 · 44 |
| civit-datapacket-talos | 25 | 0 · 0 · 0 | 7 · 2 · 16 |
| civitai-gpu-fleet | 10 | 0 · 0 · 0 | 5 · 0 · 5 |
| vetr-app | 1 | 0 · 0 · 0 | 1 · 0 · 0 |
| vetr-api | 2 | 0 · 0 · 0 | 0 · 0 · 2 |
| **TOTAL** | **186** | **4** · 0 · 3 | **36** · 42 · 101 |

🔴 **40 commits name a round; only 4 are INTERIOR**, the class where the reading is unambiguous. The other 36 are TAIL, which conflates a fix posted after the final block with work that simply continued after the ladder ended. Summing them is what the file's own line-count banner forbids in capitals. **Quote the 4.**

## Three limits, all in the output rather than only here

🔴 **What it measures is an UNLEDGERED ROUND, not unaudited churn.** A subject reading `audit round 5 — …` is evidence the round **RAN**; the gap proves it posted no two-sha `audited=` block. That is a *ledgering* defect. Round 0 caught the code asserting the opposite.

🔴 **"Names a round" is not "is that round's fix."** A `docs(handoff)` commit **narrating** a round matches too. Measured on devrc `main`: **2 of 4 matched subjects are narration** — half the population on that corpus. Deliberately not patterned around (excluding `docs(` is a guess over words, and wrong both ways — for a prose-payload ladder every fix is a `docs(` commit). Each row prints the matched **span** so a reader can judge.

🔴 **It is a floor, ~40% sensitive, never a rate.** Against the 32 hand-classified commits it sees 8 where the hand pass found 20 fixes. The 104 `unclassified` are **not** development — they are unread, and the tool refuses to guess. An earlier draft had five heuristic buckets keyed on commit types and correction verbs; deleted, because a guard spelled over *words* is walkable by rewording.

⚠ **It does not supersede the hand sampling** — 63 of 79 adjacencies remain unread by a human. The census classifies a proxy over a different unit (commits, not adjacencies). Both are needed; neither is complete.

## Evidence for what it does claim

- **Zero false positives** on the real corpus, verified by reading all 11 devrc hits — including one whose reference is a trailing `(round-2 audit)`.
- **Base-rate control:** matches **4 of 1,377** devrc `main` squash subjects (0.29%) vs 16% of gap commits — **~55× enriched**. Command in the doc.
- `eb947328` (#1233's round 3) is among the hits. ⚠ **It is NOT an independent control** — `_ROUND_REF_POSITIVES` already pins the identical `round-3 audit` form, so the regex matched a shape its own ledger carries. It confirms one case and validates nothing about the inference. An earlier version of this body claimed otherwise.

## Verification

- 27 tests; battery **30 rows, all as expected**, both `SURVIVES` controls green.
- `--not <base>` is asserted by a test, not remembered — omitting it makes a tail read as the base branch's own squash commits, which *confirms* the hypothesis under test using commits that contribute no churn.
- A listing failure is reported, never swallowed as "zero round-refs".
- ⚠ No full tier run. `tekton/devrc-pytests` is red for an inherited `main`-side reason (`test_every_site_writing_its_OWN_runner_bound_is_in_the_ledger`, naming a file this PR does not touch).

## Audit record

`ran: 2 · changed the outcome: 2`. Round 0 questioned two requirements (both right); round 1 returned 5 🟡 / 2 🟢 / **no 🔴**, including a **demonstrated vacuous guard half** (`re.compile(r"\d")` survived the suite) and a census TOTAL row that did not sum. Both audit comments are below.

### DIFF
--- a/claudedocs/audit-ladder-review-2026-09-04.md
+++ b/claudedocs/audit-ladder-review-2026-09-04.md
@@ -557,34 +557,70 @@ Both directions, and they do not cancel.
   ✅ **SAMPLING SUPERSEDED BY A CENSUS, 2026-09-11 — `--find-carriers` now classifies every gap
   commit in every repo that carries ladders.** The sample's own by-product is the instrument:
   a commit whose subject names the audit round it belongs to, sitting in a range no round
-  covered, is self-declared missed audit surface. **186 gap commits:**
+  covered, is an UNLEDGERED ROUND (see below — not unaudited churn). **186 gap commits:**
 
-  | repo | gap commits | ROUND-REF | MERGE | unclassified | gaps with a ROUND-REF |
-  |---|---|---|---|---|---|
-  | devrc | 69 | **11** | 23 | 35 | 9 |
-  | homelab-talos | 79 | **16** | 17 | 46 | 16 |
-  | civit-datapacket-talos | 25 | **7** | 2 | 16 | 7 |
-  | civitai-gpu-fleet | 10 | **5** | 0 | 5 | 5 |
-  | vetr-app | 1 | **1** | 0 | 0 | 1 |
-  | vetr-api | 2 | 0 | 0 | 2 | 0 |
-  | **TOTAL** | **186** | **40** | 42 | 104 | **38** |
+  Split INTERIOR / TAIL as `round-ref · merge · unclassified`, because only the interior half is
+  unambiguous — see the 🔴 under the table:
 
-  **40 commits across 38 gaps say in their own subject which audit round they belong to, while
-  sitting in a range no round's ledger covers.** That is the finding this whole arc was after,
-  and it needs no judgement: the commits assert it themselves.
-  🔴 **ZERO false positives on the real corpus, verified by reading all 11 devrc hits** — every
-  one is a genuine round commit, including one whose reference is a trailing parenthetical
-  (`… (round-2 audit)`). And the control that matters: **`eb947328` is among them** — the exact
-  commit this review identified BY HAND as #1233's uncovered round 3. The classifier found it
-  independently.
-  🔴 **THE CENSUS IS A FLOOR, NEVER A RATE, AND THE OUTPUT SAYS SO.** It sees only a round
-  reference a commit chose to write down; a round's fix with an ordinary subject lands in
-  `unclassified`, indistinguishable from a feature. Measured against the 32 hand-classified
-  commits: **8 self-declared where the hand pass found 20 fixes.** So the 104 `unclassified`
-  are **not** development — they are unread, and the tool refuses to guess. An earlier draft
-  had five heuristic buckets keyed on commit types and correction verbs; it was deleted,
-  because `claude/RULES.md` is explicit that a guard spelled over WORDS is walkable by
-  rewording, and a classifier is no different.
+  | repo | gap commits | INTERIOR | TAIL | gaps w/ round-ref |
+  |---|---|---|---|---|
+  | devrc | 69 | **1** · 0 · 1 | 10 · 23 · 34 | 9 |
+  | homelab-talos | 79 | **3** · 0 · 2 | 13 · 17 · 44 | 16 |
+  | civit-datapacket-talos | 25 | 0 · 0 · 0 | 7 · 2 · 16 | 7 |
+  | civitai-gpu-fleet | 10 | 0 · 0 · 0 | 5 · 0 · 5 | 5 |
+  | vetr-app | 1 | 0 · 0 · 0 | 1 · 0 · 0 | 1 |
+  | vetr-api | 2 | 0 · 0 · 0 | 0 · 0 · 2 | 0 |
+  | **TOTAL** | **186** | **4** · 0 · 7 | **36** · 42 · 101 | **38** |
+
+  🔴 **FOUR, NOT FORTY — and an earlier revision of this section headlined the 40.** Forty
+  commits name their round; only **4 are INTERIOR**, the class where the reading is unambiguous.
+  The other **36 are TAIL**, which conflates a fix posted after the final block with work that
+  simply continued after the ladder ended — nothing in the ranges separates them. Summing the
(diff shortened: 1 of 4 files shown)


ZacxDev:
## ROUND 0 — requirements & deletion pass (reports only; does not move the ladder)

`round 0 · requirements: 8 (unattributed: 1) · deletion candidates: 2`
`ran: 1 · changed the outcome: 1`

**Verdict: `requirement questioned`** — R2 (*"a commit naming its round, in a gap, IS missed audit surface; no judgement needed"*) and R3 (*"this supersedes sampling with a census"*). Not `close, do not audit`: the mechanism is real, the devrc census row was independently re-derived byte-for-byte, and the `--not <base>` assertion, the listing-failure refusal and the five-bucket deletion are all attributed, cheap and correct.

**Both findings are about this PR's PROSE, not its code. Both are now fixed, and two claims in the PR body above are wrong — corrected here rather than by a silent edit.**

### 🟡 1 — the label inverts its own evidence *(fixed)*

A subject reading `audit round 5 — <what was fixed>` is **evidence that round 5 RAN**; the commit *is* that round's fix. What the gap proves is that the round posted no two-sha `audited=` block. That is a **ledgering** defect — and the code asserted the opposite: *"IS missed audit surface … no judgement is needed"*.

🔴 **The case that settles it is the one the PR body calls "the control that matters".** `eb947328` is #1233's round 3 — and `claudedocs/audit-ladder-review-2026-09-04.md:438` says in terms that #1233's round-4 comment is **titled "rounds 3 and 4"**. Round 3 *was* audited, inside round 4's comment. The review's own mechanism hypothesis says the same. So it is a real find and a real defect, and **not** evidence the code went unread.

It is also **not an independent control**: `_ROUND_REF_POSITIVES` already pins the identical `round-3 audit` form, so the regex matched a shape its own ledger carries. It confirms one case; it validates nothing about the inference. **The PR body overstates it — treat that sentence as retracted.**

Relabelled to **UNLEDGERED ROUND**, with the narrow residual stated: *round N's fix has no successor block, so the gate-reset rule was not honoured and no round ever diffed that delta.*

### 🟡 2 — the census summed interior and tail, which the file forbids in capitals *(fixed)*

`ladder-range-coverage.py` carries `INTERIOR AND TAIL ARE DIFFERENT CLAIMS AND MUST NOT BE SUMMED INTO ONE HEADLINE` for line counts — and the census then did exactly that for commits. **Measured on devrc: interior round-ref 1, tail round-ref 10.** The "40 across 38 gaps" headline is ~91% the ambiguous class.

Compounding it, the doc edit **deleted the `⚠ Still 63 of 79 unclassified` caveat** and replaced it with the census table, which makes coverage read as complete. The census classifies a *proxy* at ~40% sensitivity over a **different unit** (commits, not adjacencies).

Fixed: the census prints the split, and the caveat is restored with the distinction stated. **"Supersedes sampling" in the PR title/body is withdrawn** — it is a weaker, different claim wearing the stronger one's words. 63 of 79 adjacencies remain unread by a human.

### 🟢 D1 — a duplicated paragraph in the battery *(fixed)*

`mutants-ladder-range-coverage.sh` carried the same 🔴 paragraph twice, and the **first copy described a mutation the row does not perform** (*"the reason itself going None"* vs the row's `if rc != 0:` → `if False:`). A superseded draft of my own, left in place. Deleted.

### Filed, because the census had no owner

**Open item 6** in the review doc. Its own output says *"Read them; the subjects are printed above"* and nothing owned doing that — `claude/RULES.md` requires a closing condition and a named checker. It closes when the **interior** ones are read (1 of 40 on devrc, so it is small) **or** when the operator writes that they will not be read. It explicitly does **not** close by re-running the census with a bigger number.

### Kept deliberately, and recorded

`OVERLAP`/`UNRELATED` have **never fired** — 0 occurrences across every recorded run plus a 64-carrier re-derivation. Kept because `_is_ancestor` returns `None` on an unresolvable sha and deleting that path makes such a sha read as a GAP *with a size*; a never-fired refusal is cheaper than the number it stops anyone inventing. No redundancy found between `ladder-depth-sweep.py` (depth, from transcripts), `audit-dispatch.py::measure_ledger` (per-round churn) and this (cross-PR range coverage) — they share `measure_range_churn` by design.

### New evidence the round produced

**Base-rate control**, which I had not run: the pattern matches **4 of 624** devrc `main` squash subjects (0.6%) against 11 of 69 gap commits (16%) — **~26× enriched**. Good evidence the signal is real for what it now claims to measure.

### Uncertainty

The round did not run either `nix build` tier or the head tree's suite (no worktree available without writing to the shared git dir), so the 25-test / 26-row counts are re-derived by inspection. The five non-devrc census rows remain **unverified** — re-deriving them needs `fetch refs/pull/<n>/head` into client-owned shared clones.

⚠ **An audit fix resets the verification gate**, so these fixes are re-verified live and the nine correctness axes have not run yet.

PR #1578, opened 2026-09-12
feat(bar): relay the workbench's whole bar to the laptop as one pill

ZacxDev:
## The problem
SSH'd from the laptop to the workbench, the i3 bar on screen is the **laptop's**. The laptop omits **12 pills** (`telemetry alerts civitai mail clawgate media airvpn runaways claudeRuns rigcontrol fans gpu`), so the machine actually being worked on is invisible.

## The shape
A single `wb` pill on the laptop bar, fed by a snapshot pulled over nebula, with a drill-down TUI behind its click.

🔴 **This is a transport problem, not a bar problem.** The bar is deliberately CALM — blocks never query anything remote per tick, they render local JSON that `bar-status-poll` fetches out of band. So the existing blocks work unchanged if the data arrives. Nebula is ~138 ms RTT, which is also why per-tick remote queries would be the wrong shape.

Running a second poller on the laptop was the cheaper-looking option and is worse: it doubles standing port-forwards into **client prod** every 45 s, and still carries no workbench *machine* state (load/fans/runaways/claude-runs) — most of what "how is the workbench doing" means.

## The gather runs the bar's own block commands
Every red line already lives in two places (`graphical.nix` argv + the poller's toast env). Re-deriving verdicts here would make a **third**, and the one furthest from anyone's eyes.

So the gather reads the **deployed** `config-top.toml`, takes each custom block's `command` verbatim, runs it, and records what it emitted. **The verdict travels; the predicate does not** — the relayed pill cannot disagree with the bar it mirrors, because it is the same opinion relayed, not a second opinion. New blocks and retuned thresholds come along for free.

Built-in blocks (cpu/memory/disk/net) have no command, so they are gathered separately under `vitals`, kept in a separate key so a reader can tell a relayed verdict from a locally-computed number.

## Failure semantics reuse what the bar already has
`bar_freshness.unmeasured` works on the payload, not the path — so an unreachable peer is indistinguishable from a dead poller, and renders the bar's existing trailing `?`. On an outage the puller preserves the prior snapshot under `last_good` and the pill carries its alarms behind the `?`.

**Pull, not push** — a push fires at a laptop that is asleep or roaming, every 60 s, toasting and filling the workbench journal with the operator's own lid being shut. Pulling puts the schedule on the machine that knows whether it is awake. The unit exits 0 for an unreachable peer and has **no** `OnFailure` toast; a stderr line is the only thing separating a transient outage from a permanently broken pull.

**The pill does not hide at zero** — alone on this bar. It describes a machine nobody is looking at, so "quiet" and "I have not heard from it in an hour" would otherwise both render as nothing.

## Three defects the controls caught
| | |
|---|---|
| `json` assumed for every custom block | MEASURED: 15 custom blocks, only **14** declare `json = true`. `i3blocks-rigcontrol` prints a bare glyph as plain text, and the gather scored that healthy block `unparseable` — **inventing an alarm** for a block the remote bar renders fine. |
| label defaulted to `gethostname()` | 🔴 **Both hosts are named `nixos`**, so it names neither — the same trap RULES records from the claim-lock's `uname -n`. Identity is now declared, with `label_source` recorded so a guessed name is reported as guessed. |
| disk deduped by value | `/` and `/home` are one filesystem here (same `st_dev`), so the detail view printed the same free space twice as two disks. |

## Two existing guards widened — both from a spelling to a relationship
- `test_every_block_that_loads_the_sibling_is_DEPLOYED_beside_it` asserted the literal `!isLaptop`, while its own docstring states the invariant as "not **narrower** than its consumers". A laptop-only consumer made the correct widening fail the word. Now extracts each entry's gate and asserts containment — with a positive control proving the helper **refuses** an unrecognised condition rather than defaulting to permissive.
- `test_the_PROSE_COUNT_of_freshness_consumers_matches_the_SOURCES_ledger` equated consumers with poller `SOURCES` — one sentence while there was one writer. There are now two. The second writer is named explicitly, pinned two-way, and the named timer must exist. 🔴 **The prose count stays the polled population**: I read all six pinned sentences and every one is about the poller's blocks, so renumbering them to nine would have made three outright false — the N−1 defect, committed in the act of satisfying the guard.

## Tests
36 new, plus a **9-mutant battery** under `PYTHONDONTWRITEBYTECODE=1` with a positive control. 8 killed singly; the click-cmd guard needs **both** redundant mechanisms broken, shown explicitly.

⚠ Two mutants initially SURVIVED and were **bad mutants, not gaps** — they targeted expressions that do not encode the behaviour (the `json` *parse* line rather than the *default*; the `[` branch rather than the `command`/`cmd` regex). Re-run against the narrowest wrong-able expression, the real one is killed.

Full gate (`--tier both`) running; result in a follow-up comment.

## 🔴 Not yet verified live, and it needs the operator
Every piece is proven end to end **by hand**: gather against the live workbench bar (15 blocks, real verdicts), a real pull from the laptop over nebula, the pill in all 7 states, the detail view rendering.

What is **not** proven is the nix wiring deploying and the bar showing the pill — **adding a block is restart-class** (the bar holds the block list it parsed at startup), and a bar restart takes the screen of whoever is sitting at that machine. That is the laptop, where the operator is. Handing it over rather than doing it.


ZacxDev:
## Round 1 — fixes landed

Ten findings; the 🔴 was a real merge-blocker this PR introduced. Fix commit `381f7725`.

⚠ **Gate status, stated honestly:** the dev-host `gate.sh --tier both` run on `ff5c590c` produced **no pytest verdict** — `RESULT: FAIL (exit=143)`, `pytest exit=124 (timeout after 3600s)`. It never reached the hazard test. Box load was 59 with 24 concurrent full suites. The node tier passed (1449 tests, floor 1367). The three touched suites pass locally: **704 tests**. Tekton is the authoritative signal and is pending on `381f7725`.

```audit-claims round=1 audited=ff5c590c..381f7725
1. test_every_hazardous_binary_the_scripts_reach_is_stubbed_or_acknowledged went RED at ff5c590c and GREEN at the merge base; bar-remote-snapshot and remote-host-detail are now entered in ACKNOWLEDGED_UNSTUBBED['systemctl'] and re-justified separately as two different cases (a real spawn vs a help-string mention).
2. The same scanner then flagged the fix's own comment, which named a window-control binary literally; the family is now described rather than spelled, with a note recording why.
3. test_every_block_that_loads_the_sibling_is_DEPLOYED_beside_it checked eight of nine sibling consumers (it iterated BLOCK_SOURCE_FILES, which i3status-remote-host cannot join); consumers are now DERIVED from the tree via _load_freshness, with a floor asserting the derivation covers at least the old list.
4. The narrowing mutant that previously PASSED was rebuilt against the fixed guard and watched go PASS -> FAIL -> PASS, with nix/graphical.nix byte-identical after restore.
5. _home_file_gates classified on the home.file key's own line, so a wrapped entry read as the permissive "both"; it now reads the continuation line too.
6. gather_poller_cache filtered on ".json" and carried airvpn.json and runaways.json, whose blocks are in RELAY_BLOCKS; the carried set is now DERIVED from NATIVE_BLOCKS and pinned two-way.
7. The comment justifying ~/.cache/bar-remote claimed the relay mixes nothing into ~/.cache/bar-status, which install_poller_cache contradicts; corrected in place.
8. Three tests set XDG_CACHE_HOME but not XDG_CONFIG_HOME and executed the operator's live deployed bar; they are now hermetic, with a positive control watched RED when the fixture is removed and GREEN with it.
9. Four prose claims falsified by this diff are corrected (nix/graphical.nix x2, claude/skills/bar/SKILL.md), including the one ~50 lines above the code that falsified it.
10. not_relayed had no reader anywhere; the detail view now renders it.
11. SCHEMA was a DTO field with no branch (a payload claiming schema 999 rendered "wb ok"/Idle); pull now reports a newer-than-understood snapshot as unreadable, accepting older.
12. The DISPLAY/XAUTHORITY injection was dead and its stated justification false (it cited game mode, which is LOCAL_ONLY and never gathered); removed.
13. An alarming block with empty text produced a bare "wb +1" with nothing to be more than (reachable: airvpn's stale pill is literally {"text": "", "state": "Warning"}); it now falls back to the block name.
14. local_poller_is_running's docstring said it "fails SAFE" when with respect to its hazard it fails OPEN; the choice stands, the label is corrected.
```

**NOT fixed, needs the operator (R1-3):** off nebula the laptop gains six Warning `?` pills (measured with an empty cache), the mirror image of the defect this PR was rewritten to remove. A knowing behaviour trade, named nowhere in the PR or the skill, and not an auditor's call.

**Round 0 ledger:** `round 0 · requirements: 11 (unattributed: 4) · deletion candidates: 8`.

PR #1580, opened 2026-09-12
docs(subsystem-index): the probe's already-there/OPEN/NEAR-MISS displays read the FROZEN mirror

ZacxDev:
The step tells you to compare your proposed bullet against the `already there` lines, and to act on the `OPEN:` and `NEAR-MISS` blocks. All three render from `~/.claude/analyze-service-index` — the probe names it on its own `store:` line — and since the Cairn cutover that tree does not move. The comparison the protocol leans on can be arbitrarily far behind the pod.

**Measured 2026-09-12, one scope:** mirror held **11** bullets (newest 2026-08-31) against **29** on the pod (newest 2026-09-11) — 12 days and 18 bullets of drift; the mirror listed 11 entry files where the pod had 36.

Both items that run's probe flagged for repair were **already fixed upstream**: a NEAR-MISS marker that parses fine live, and an `OPEN:` bullet the pod already carried as `RESOLVED <sha>:` plus a later bullet closing its open half. Acting on the display would have rewritten correct prose and re-opened a closed action — the exact failure the `already there` comparison exists to prevent, arriving through its own input.

Remedy is one line the write half already prescribes for `put`: `cairn sync` first, then diff `~/.cache/subsystem-store/<scope>/<ref>.md` against what the probe printed. Same read, one step earlier.

Text only, no behaviour change. `test_index_append_protocol.py` and `test_subsystem_touch.py` (which pins this file's sentences verbatim) both green — **1008 passed**.

https://claude.ai/code/session_01FqmKj7nMiGwPxUdtUWRJSn


ZacxDev:
Duplicate of #1450, which filed the same finding on 2026-09-09. My independent measurement (11 vs 29 bullets, 12 days of drift, plus the NEAR-MISS block being affected too) is now a comment there: https://github.com/innovation-upstream/devrc/pull/1450#issuecomment-5643744054

Closing rather than leaving two PRs for one fix. My duplicate-sweep ran in the same command as the create, so its answer arrived too late to stop it — the sweep has to be its own step immediately before `gh pr create`.

PR #1582, opened 2026-09-12
feat(mention-open): open a clicked GitHub mention in octo.nvim, behind a browser fallback

ZacxDev:
A clicked GitHub `repo#N` mention opens in **octo.nvim** — a full review-and-merge TUI — instead of a browser tab. Everything else (clawgate tasks, ClickUp ids, any GitHub URL that is not `/issues/N` or `/pull/N`) opens exactly as before, and the browser stays reachable.

> **`/audit-pr` ROUND 0 ran and its deletions are applied** (commit `421631b3`). Three things came out before the correctness ladder spent a round on them; the detail is at the bottom. The headline: **`main()` is now byte-identical to base**, so this change is purely additive to the decision path and the cross-PR conflict with #1569 is gone — verified, not assumed.

## What changed

**`scripts/mention-open.py`** — `open_url` keeps its exact pre-TUI signature and becomes a dispatcher:

```python
def open_url(url):
    if open_target(url) == TARGET_TUI:
        return open_tui(url)
    return open_browser(url)          # the old body, renamed
```

`open_target` rungs: **(0)** the URL's own shape — anything `github_ref` cannot turn into `(repo, number)` is browser, and no marker can override that; **(1)** a marker file at `~/.config/mention-open/target`; **(2)** the TUI iff this host has it, silently browser otherwise.

**The number is passed, not the URL.** `openable()` builds `/pull/{id}` for *every* GitHub mention because the handler cannot know whether `#N` is an issue or a PR — github.com resolves that by redirecting. `Octo <N> <owner/repo>` fires one `issueOrPullRequest` GraphQL query and dispatches on the `__typename` the server returns. A `/pull/` URL would have asserted the kind, wrongly, for every issue.

**A marker file, not an env var.** Alacritty spawns the hint handler with the *display manager's* environment, so a shell-set variable can never reach a click. `~/.server-mode` is the precedent.

**`shutil.which` pre-flight is mandatory.** alacritty exits 0 whether its `-e` command exits 0 or 127, and `Popen` never waits — a missing `nvim-octo` is a window that flashes and vanishes, invisible to the caller. A post-spawn fallback is structurally impossible. On failure: browser + exactly one toast.

**`nix/pkgs/tools/nvim-octo/`** — a `writeShellApplication` wrapping `pkgs.neovim.override { configure = … }` with octo-nvim, plenary, fzf-lua, nvim-web-devicons and gruvbox on the packpath (octo declares no runtime deps), plus `gh` pinned by store path. Exposed as `pkgs.nvim-octo` through an overlay in `flake.nix` — that spelling is load-bearing for the AST ledger.

## Merge safety

`mappings_disable_default = true`, then only non-destructive mappings re-declared. Merging requires typing `:Octo pr merge`.

Three findings beyond the brief, each measured against octo 2026-08-28:

1. **There are seven merge mappings, not four.** Besides `merge_pr`, `squash_and_merge_pr`, `rebase_and_merge_pr`, `merge_pr_stack`, upstream also binds `merge_pr_queue`, `squash_and_merge_queue`, `rebase_and_merge_queue`. All seven withheld.
2. **`pr_options` is `<CR>` and its menu contains "Merge PR", "Squash and Merge PR" and "Delete Branch".** Re-declaring it would have left merge one keystroke away with every merge *keymap* gone — a guard that reads as coverage and provides none. Not re-declared. `issue_options` is, because its menu carries neither.
3. **`mappings_disable_default` does NOT clear `picker_config.mappings`.** Measured on the built derivation: after setup, `picker_config.mappings.merge_pr.lhs` is still `<C-r>`. What closes it is `picker = "fzf-lua"` — that picker tree contains no occurrence of "merge", while telescope wires that lhs to a merge action in two places. The picker choice is a safety setting and the suite asserts it as one.

Also `default_merge_method = "squash"` (upstream defaults to a merge commit), `default_delete_branch = false` (deleting a stacked parent's branch auto-closes the child PR, which GitHub then refuses to reopen), `use_local_fs = false`.

**One upstream bug worked around.** octo's `gh` guard is `if not vim.fn.executable(cfg.gh_cmd)`, and `vim.fn.executable()` returns the *number* `0`, truthy in Lua — so the guard never fires and `setup()` throws inside plenary instead. With the setup call last, that traceback aborted the rest of the init: no colorscheme, no options. Options and colours now run first and the call is wrapped, with a message naming the cause.

## Ledgers

- **`test_mention_open.py`'s wrapper-PATH seam.** `nvim-octo` is neither argv[0] nor inside a `sh -c` script — it is alacritty's `-e` payload, invisible to both existing readers. A third reader, `_exec_payload_commands`, reads the first word after a literal `-e` out of the syntax tree (resolving the `REVIEW_EXE` module constant), with its own positive *and* negative controls. `PROVIDER` gains `pkgs.nvim-octo`.
- **`launcher_scan.HAZARD_VOCABULARY`** gains `nvim-octo`.
- **`nolaunch.ACKNOWLEDGED_UNSTUBBED`** gains it bound to `{"mention-open.py"}`. **Not** `HOST_LAUNCHERS`: the stub dir is on PATH, so a `nvim-octo` stub would make `shutil.which` succeed and flip `tui_available()` true for every test in the file that does not override it — silently moving the whole suite off the browser path. Ships with its pin, per the tmux-reply-agent lesson.

`EXPECTED_ARGV0` and `SPAWNABLE_EXECUTABLES` did **not** move: argv[0] is still `alacritty`. Their comments now say there are two alacritty call sites, so the set cannot be read as a count.

## Verification

**Mutation: every guard killed by its own named guard.** Run under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` swept between mutants, against a `.git`-free `git archive` copy. Negative control (clean tree green) and positive control (a mutant caught by a pre-existing unrelated test) both watched. Covers every `open_target` rung, the pre-flight, argv order, `REVIEW_EXE`, `github_ref`, both `main()` call sites routing *around* `open_url`, the deletion pin, all five octo settings, all four wrapper guards, and the ledgers (dropping `pkgs.nvim-octo` from the wrapper PATH, dropping the vocabulary name, a direct `Popen(["nvim-octo", …])`).

**Red-at-base found a vacuous green in my own tests** (`cf519a1d`). Against `61f41adf`, where `nvim-octo.sh` does not exist, 18 rejection tests **passed** — `bash` exits non-zero with an empty argv log for a missing script exactly as a correct wrapper does for a bad argument. Each rejection now pins the wrapper's own exit code (64/65/66, distinct per guard so a kill is attributable) and its own message. Red-at-base went from 20 passed to 2 — the two self-contained instrument controls.

**Live measurement, headless, no window and no network** (`gh` stubbed so octo returns before `gh auth status`): 44 `pull_request` mappings with all seven merge keys and `pr_options` absent; `default_merge_method=squash`, `default_delete_branch=false`, `use_local_fs=false`, `picker=fzf-lua`; `exists(':Octo')` → 2; colorscheme gruvbox; every other mapping table non-empty. Re-run after the deletions: unchanged.

## Concurrent work — #1569, before and after the deletions

| | conflicts in `mention-open.py` | resolution needed | result |
|---|---|---|---|
| **before** (`cf519a1d`) | 3, two of them **modify/modify** at the call sites | naive = **silently drops the flag**; correct = re-apply keyword + widen 5 stubs | naive: 418 pass, **2 seam guards red**; correct: 546 pass |
| **after** (`421631b3`) | 1, **add/add with an empty base** | **none — union only** | **564 pass, 0 fail** |

The union resolver *refuses* on any non-empty base, so "purely additive" is a checked property rather than a claim. Whoever merges second has nothing to re-apply.

## What ROUND 0 deleted

- **D1 — `--browser` and its `open_target` rung.** Unreachable in production: the only production caller is the Alacritty hint, whose `command` gets the matched text and nothing else, and whose regex cannot produce a token starting with `-`; the diff adds no new `[[hints.enabled]]` entry. Its sole caller was a human typing it in a shell, while its docstring called it "the operator's per-click escape hatch". The marker file already does this job and *is* reachable. It was also the sole cause of the #1569 conflict. The deletion is **pinned** (`test_open_url_takes_NO_argument_but_the_url` asserts both signatures and the parser help) because re-adding a keyword re-opens the cross-PR hazard. The picker-call-site seam guard survives as `test_main_routes_a_PICKED_github_row_through_the_review_TUI`.
- **D2 — the `home.packages` entry.** Its own comment conceded it was not what the click uses. The overlay already forces the build; it bought a hand-typed invocation nobody asked for at the cost of a known collision surface with imperative `nix profile` installs and a second wrapped-neovim closure.
- **D3 — `pkgs.git` from `runtimeInputs`.** octo reads a remote only for a buffer opened *without* an explicit repo, and `nvim-octo.sh` refuses that invocation (exit 64). It served a call shape that cannot happen. Verified gone from the built wrapper's PATH.

**Declined, with reasons** (D4/D5, marked marginal): the `discussion` mapping table *is* reachable from `Octo <N> <repo>` — `utils.open_buffer` dispatches on `__typename` and has a `Discussion` arm, so round 0's premise is wrong there; deleting `repo`/`release` would redden `test_every_other_mapping_table_the_flag_cleared_is_re_declared`, which pins table *names*; and `open_tui`'s `if not repo` arm is 3 lines guarding against spawning `nvim-octo "" ""` in a module-public function.

Net: 10 files, and the **total set of pre-existing lines this diff removes is 12**, six of which are one comment block being rewritten.

## The #1569 surface seam — found here, and now INTEGRATED

#1569 is merged, and this branch now carries the integration (`afb0d3d2`).

**The defect.** #1569's `open_reference` ended `return open_url(url), CLICK_SURFACE_BROWSER` — a measurement wrapper whose measurement is a literal, which is exactly what its own docstring warns against one level up. Correct while the browser was the only opener; wrong the moment this PR adds a second. Union-merged it made **every TUI click record `surface=browser`** with the whole suite green. Measured on the merged tree, before and after:

```
before   A  spawned=alacritty  reported='browser'   WRONG
         B  spawned=xdg-open   reported='browser'   open_target()='tui'
after    A  spawned=alacritty  reported='tui'       OK
         B  spawned=xdg-open   reported='browser'   open_target()='tui'
```

**The fix.** `open_tui` now returns `(rc, surface)` and every arm reports the surface that *actually ran*. `open_reference` propagates that tuple on the TUI arm and delegates to `open_url` on the browser arm. The surface is never re-derived from `open_target()` — row B is why: the routing decision says `tui` while `xdg-open` is what opened the link, so a derived value would be a claim rather than a measurement.

⚠ **`open_url` still routes, deliberately.** Making `open_reference` the sole dispatch broke 30 tests, the decisive one being #1569's *own* seam guard, which stubs `open_url` and asserts the browser arm delegates to it — bypassing it turns a real guard inert. Both route; the duplication is documented and paid for by `test_open_url_and_open_reference_ROUTE_THE_SAME_WAY`, a 6-row table driving both and asserting the spawn and the reported surface name the same opener.

**#1569's invariants kept, verified not asserted:** the exit code is still `open_url`'s and a failure names the surface it *tried*; a dismissal emits no `surface=` at all; both emit sites pass `surface=surface`, never a literal.

**Rebuilt against the real merge, not replayed.** The conflict was *seen* — git printed `Recorded preimage`, never "Resolved using previous resolution". `rr-cache` was checked preimage-only beforehand, and the commit used `rerere.enabled=false` so no resolution was recorded after.

**Mutation: 7/7 killed by a behavioural case**, twice, independently, on an uncontended tree; negative control `passed=454 failed=0`. A structural-only kill was scored as a failure. All four `open_tui` arms are covered and each is uniquely anchored — three share a byte-identical line, so a mutant anchored on it alone is refused and an arm can silently end up untested. ⚠ **S5 survived round 1**: the spawn-failure arm had no behavioural surface assertion and the suite stayed green at 453 while that arm reported the wrong surface.

## Merging with #1569 — the record of how this was found

Re-measured against #1569's newer head `6f6a0101`, which adds `open_reference` — a wrapper that returns `(rc, surface)` for click telemetry. **All four conflict hunks are add/add with an empty base, the union merge applies with no manual decision, and 486 tests pass — and the result is WRONG.** Measured on the merged tree:

```
A  spawned=alacritty  reported_surface='browser'   WRONG
B  spawned=xdg-open   reported_surface='browser'   open_target()='tui'
```

`open_reference` hardcodes `return open_url(url), CLICK_SURFACE_BROWSER`, so **every TUI click is recorded as a browser click** while the whole suite stays green. That is precisely the failure its own docstring predicts ("the literal version would keep reporting `browser` from a branch that no longer uses one, and every suite would stay green"), arriving through the merge rather than through an edit.

Good news: #1569 pre-wired the slot — `CLICK_SURFACE_TUI` exists and is in `CLICK_SURFACES`, and the docstring names `open_reference` as the single integration point for #1582.

⚠ **But it is not a one-liner.** Case B above is the trap: deriving the surface from `open_target()` would report `'tui'` while `xdg-open` actually opened it, because `open_tui`'s pre-flight falls back. #1569's rule is "the surface is MEASURED FROM THE OPENER THAT RAN, NOT ASSERTED AT THE CALL SITE" — so the integration must have `open_tui` report which surface it actually used. Whoever merges second should do that, not union-and-go.

This corrects my own earlier "union only, nothing to re-apply" claim, which was textually right and semantically incomplete — a clean `git merge` is not a clean merge.

## CI caught a real defect in this branch

`tekton/devrc-pytests` failed `test_every_mutation_anchor_occurs_exactly_once_in_its_target[mutation_battery_mentions.py]`, reproduced locally. **Mine, not inherited.** The repo-tracked battery's row **K53** anchors on the WHOLE `makeBinPath` line, so adding `pkgs.nvim-octo` took that anchor to **0 occurrences** — and an anchor at 0x does not fail loudly: the row reports `!! PATTERN OCCURS 0x — NOT APPLIED` and scores as a **SURVIVOR without testing anything**. Exactly the silent green `test_mutation_battery_anchors.py` exists to catch, and it caught it.

Re-anchored on the line as it now reads, with the mutation unchanged in kind (drop `pkgs.fzf`, keep the rest). Per that test's own instruction the **battery row** was fixed, never the test. Then verified it still *kills* rather than merely parses:

```
P1   control      KILLED              f=20  p=634   positive control — the battery can observe
K53  deletion     KILLED(attributed)  f=1   p=653   killed by test_the_alacritty_wrapper_PATH_covers_every_executa…
2/2 killed for the stated reason; problems: none        restore: OK
```

`tekton/devrc-nodetests` passed (1449 tests) and `tekton/devrc-cairn-client-runs` passed on the same head.

## Not verified — stated plainly

- That alacritty maps a window, that the review terminal is usable, that i3 floats it.
- That `Octo <N> <repo>` resolves the correct buffer kind against real GitHub — that needs a live GraphQL call.
- That the merge mappings are absent from a **live PR buffer**. The headless probe measured the merged *config*; only an open buffer proves the keymaps.
- That a broken `nvim-octo` is noticed in production — the pre-flight is tested, the deployed failure mode is not.
- **Nothing is deployed.** Verified `nvim-octo` is not on PATH and no marker file exists, so rung 2 yields the browser: **this merges inert** until `home-manager switch` / `ship.sh`, which is the operator's call.
- `test_runner_bound_ledger.py::test_every_site_writing_its_OWN_runner_bound_is_in_the_ledger` is **inherited red** from `622fc2d8` (#1536) on `main` — measured failing identically at this branch's base, in a file this diff never touches.

https://claude.ai/code/session_012MK5m7J2Z3h698wMPczJH4

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -82,9 +82,36 @@
       system = "x86_64-linux";
       # Explicit allowUnfree so unfree pkgs (elixir-ls, playwright browsers)
       # build without relying on an ambient NIXPKGS_ALLOW_UNFREE / --impure.
+      # ---------------------------------------------------------------------
+      # 🔴 THE OVERLAY EXISTS SO ONE PACKAGE IS SPELLED `pkgs.nvim-octo`, AND
+      # THAT SPELLING IS LOAD-BEARING RATHER THAN COSMETIC.
+      #
+      # `nvim-octo` is consumed in two places: `nix/pkgs/tools/default.nix`
+      # (so the operator has it on PATH) and — the one that matters — the hint
+      # wrapper's `lib.makeBinPath` in `nix/programs/alacritty/default.nix`.
+      # That list is pinned against `mention-open.py`'s own syntax tree by
+      # `test_mention_open.py::test_the_alacritty_wrapper_PATH_covers_every_
+      # executable_the_handler_spawns`, which reads the block TEXTUALLY and
+      # matches `pkgs\.[A-Za-z0-9_-]+`. A local `let`-bound derivation is not
+      # spelled that way, so it would be INVISIBLE to the reader: the wrapper
+      # would look like it pins nothing for `nvim-octo`, and the seam that test
+      # exists to hold — "everything the handler spawns is on the wrapper's
+      # PATH" — would quietly stop covering the review TUI.
+      #
+      # So the package is put into the package set instead of being threaded as
+      # an argument. `callPackage` is not used: the derivation takes `pkgs`
+      # whole (it reaches `pkgs.vimPlugins`, `pkgs.neovim` and
+      # `pkgs.writeShellApplication`), so it is called with `final` directly —
+      # which also means a later overlay can still override it.
+      # ---------------------------------------------------------------------
+      nvimOctoOverlay = final: _prev: {
+        nvim-octo = import ./nix/pkgs/tools/nvim-octo { pkgs = final; };
+      };
+
       pkgs = import nixpkgs {
         inherit system;
         config.allowUnfree = true;
+        overlays = [ nvimOctoOverlay ];
       };
       # Same allowUnfree treatment for the frozen 1.57 nixpkgs — the browser
       # bundle is unfree there too, and an --impure fallback would make the
--- a/nix/pkgs/tools/default.nix
+++ b/nix/pkgs/tools/default.nix
@@ -58,6 +58,19 @@ with pkgs; [
   # discriminating check is in the assertion message of
   # test_engine_is_the_version_every_measurement_is_keyed_to.
   opencode
+
+  # nvim-octo — neovim + octo.nvim as a single-purpose GitHub review TUI, the
+  # surface a clicked `repo#N` mention opens in. Supplied by the overlay in
+  # `flake.nix` (see the comment there for why it must be spelled `pkgs.X`),
+  # built from `nix/pkgs/tools/nvim-octo/`.
+  #
+  # ⚠ THIS ENTRY IS FOR THE OPERATOR'S OWN PATH AND IS NOT WHAT THE CLICK USES.
+  # The hint handler is spawned by alacritty with the display manager's
+  # environment and resolves the binary through the wrapper's pinned
+  # `lib.makeBinPath` in `nix/programs/alacritty/default.nix` — which also
+  # survives the ~1s window where a `home-manager switch` blanks
+  # `~/.nix-profile`. Both entries are real; neither substitutes for the other.
+  nvim-octo
 ]
 ++ (import ./tmux-fuzzyclaw.nix { inherit pkgs workspace; })
 # clawgatectl — machine client for the clawgate JSON API. Built from the
--- a/nix/pkgs/tools/default.nix
+++ b/nix/pkgs/tools/default.nix
@@ -58,19 +58,6 @@ with pkgs; [
   # discriminating check is in the assertion message of
   # test_engine_is_the_version_every_measurement_is_keyed_to.
   opencode
-
-  # nvim-octo — neovim + octo.nvim as a single-purpose GitHub review TUI, the
-  # surface a clicked `repo#N` mention opens in. Supplied by the overlay in
-  # `flake.nix` (see the comment there for why it must be spelled `pkgs.X`),
-  # built from `nix/pkgs/tools/nvim-octo/`.
-  #
-  # ⚠ THIS ENTRY IS FOR THE OPERATOR'S OWN PATH AND IS NOT WHAT THE CLICK USES.
-  # The hint handler is spawned by alacritty with the display manager's
-  # environment and resolves the binary through the wrapper's pinned
-  # `lib.makeBinPath` in `nix/programs/alacritty/default.nix` — which also
-  # survives the ~1s window where a `home-manager switch` blanks
-  # `~/.nix-profile`. Both entries are real; neither substitutes for the other.
-  nvim-octo
 ]
 ++ (import ./tmux-fuzzyclaw.nix { inherit pkgs workspace; })
 # clawgatectl — machine client for the clawgate JSON API. Built from the
--- a/nix/pkgs/tools/nvim-octo/default.nix
+++ b/nix/pkgs/tools/nvim-octo/default.nix
@@ -30,6 +30,18 @@
 #   gruvbox-nvim       — the colorscheme `octo-init.lua` sets, matching the
 #                        alacritty palette in nix/programs/alacritty.
 #
+# ⚠ IT IS NOT IN `home.packages`, AND THAT IS DELIBERATE. `nix/pkgs/tools/
+# default.nix` carried an entry for a while; it bought only a hand-typed
+# invocation nobody asked for, while adding a `home.packages` entry (a known
+# collision surface with an imperative `nix profile install` on this host) and a
+# second wrapped-neovim closure on the operator's PATH. The CLICK does not use
+# it: the hint handler resolves `nvim-octo` through the wrapper's pinned
+# `lib.makeBinPath` in `nix/programs/alacritty/default.nix`, which is the entry
+# the AST ledger pins and the one that survives the ~1s window where a
+# `home-manager switch` blanks `~/.nix-profile`. The overlay in `flake.nix`
+# already forces this derivation to build, so nothing is un-gated by its
+# absence.
+#
 # 🔴 `gh` IS A HARD REQUIREMENT, NOT A NICETY. octo's `setup()` checks
 # `vim.fn.executable(config.values.gh_cmd)` and REFUSES to initialise without
 # it — the `Octo` user command is created by `commands.setup()`, which runs
@@ -68,11 +80,14 @@ in
 pkgs.writeShellApplication {
   name = "nvim-octo";
 
-  # `git` is here because octo reads the remote host/name for buffers opened
-  # WITHOUT an explicit repo. `mention-open.py` always passes one, so this is
-  # belt-and-braces for a hand-typed invocation rather than a path the click
-  # depends on — stated plainly rather than justified as essential.
-  runtimeInputs = [ nvimWithOcto pkgs.gh pkgs.git ];
+  # ⚠ `pkgs.git` IS DELIBERATELY ABSENT, AND IT WAS HERE. octo reads a remote's
+  # host/name only for a buffer opened WITHOUT an explicit repo — and this
+  # wrapper REFUSES such an invocation: `nvim-octo.sh` requires both arguments
+  # and exits 64 without them. So git could only ever have served a call shape
+  # that cannot happen, which is dead weight in the closure plus a comment
+  # justifying a code path that does not exist. If a future `nvim-octo` grows a
+  # one-argument or zero-argument form, add it back WITH that form.
+  runtimeInputs = [ nvimWithOcto pkgs.gh ];
 
   text = builtins.readFile ./nvim-octo.sh;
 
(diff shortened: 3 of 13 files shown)


ZacxDev:
## Note for whoever merges this alongside #1569

Measured against #1569 head `6f6a0101`. The merge conflicts are **all add/add with an empty base**, so `git` will let a union resolution through and **486 tests pass** — but the result is wrong:

```
A  spawned=alacritty  reported_surface='browser'   WRONG
B  spawned=xdg-open   reported_surface='browser'   open_target()='tui'
```

`open_reference` hardcodes `return open_url(url), CLICK_SURFACE_BROWSER`, so every TUI click is logged as a browser click with the suite green — the exact failure its own docstring predicts.

#1569 already pre-wired `CLICK_SURFACE_TUI` and names `open_reference` the integration point, so the change belongs there. ⚠ It is **not** a one-liner: deriving the surface from `open_target()` would report `tui` in case B while `xdg-open` actually opened it (`open_tui`'s pre-flight falls back), which is the "CLAIM vs MEASUREMENT" distinction #1569 is explicit about. `open_tui` needs to report the surface it actually used.

Nothing to change in this PR — it is correct standalone and based on `main`.

ZacxDev:
## Superseding my earlier note — the integration is now IN this PR

That comment ended *"Nothing to change in this PR — it is correct standalone and based on `main`."* That was true when #1569 was still open. It is **no longer true** and should not be read as current.

#1569 merged, and `afb0d3d2` on this branch is the real merge plus the surface integration:

- `open_tui` returns `(rc, surface)`; every arm reports the surface that actually ran.
- `open_reference` propagates it on the TUI arm, delegates to `open_url` on the browser arm — never re-derives from `open_target()`.
- `open_url` still routes (deliberate, documented), because #1569's own seam guard stubs it and asserts the browser arm delegates to it; bypassing that turns a real guard inert. The duplication is paid for by `test_open_url_and_open_reference_ROUTE_THE_SAME_WAY`.

Measured before/after on the merged tree:

```
before   A  alacritty -> reported 'browser'   WRONG
after    A  alacritty -> reported 'tui'       OK
         B  xdg-open  -> reported 'browser'   (open_target()='tui' — why it must propagate, not re-derive)
```

Mutation **7/7 behavioural**, twice independently, negative control `passed=454 failed=0`. The conflict was *seen*, not replayed: `Recorded preimage`, `rr-cache` verified preimage-only beforehand, commit made with `rerere.enabled=false`.

PR #1583, opened 2026-09-12
fix(cairn): rank 23 — the recovery command named a path true only on the machine that printed it

ZacxDev:
Rank 23(a) and (b) of `claudedocs/handoff-cairn-oss-multi-instance.md`. (c) is **not** here — it lives in `ZacxDev/cairn` and still needs an upstream PR.

## The defect

One class, three places. A write protocol whose whole point is that agents work in **other** repos was quoting commands that are true for the machine that printed them and false for anyone who pastes them elsewhere, with nothing in the text to tell those apart:

- `validate_command()` emitted `python3 <absolute path to subsystem_touch.py> …`
- `claude/skills/resume/SKILL.md` ×2 spelled a bare `subsystem_touch.py --validate` — **not on PATH, exits 127**

All three now name `cairn-validate`, a bare command on `home.sessionPath`, and the spelling that launcher's own docstring declares as its interface. This also brings `resume/` into agreement with `cairn/SKILL.md:36` and `subsystem-index/SKILL.md:163`, which already used these two forms — `resume/` was the odd one out.

⚠ **The item's line numbers were stale** (`:128`/`:150`; actually `:135`/`:157`). Found by grepping the string, not by opening the named line.

## `--store` is emitted explicitly, and that is not redundant

The launcher prepends the **synced cache** as its default store, while `validate_command`'s contract is to check the store the refusal actually came from. Passing it means argparse's last-occurrence-wins gives that value, so the command cannot silently report a *different* store clean — the exact failure `malformed_refusal` exists to prevent.

Measured on the deployed pin, not reasoned: with an explicit `--store`, the run's own `store:` line names it rather than the launcher's default, and a malformed entry still exits **3**, untranslated.

## The cost rank 23 did not anticipate — the reusable part

Six neighbouring tests broke, because `test_the_recovery_command_ACTUALLY_RUNS_and_reproduces_the_diagnosis` **executes** the emitted command.

`python3 <abs path>` is runnable in both tiers. A bare `cairn-validate` is a home-manager symlink the `nix build` tier has no reason to carry. **Exec'ing the real launcher would have made that guard structurally incapable of passing in one tier while staying green on a dev host — the defect this repo already shipped once.**

So the launcher's prepend is modelled in exactly one helper (`_writer_argv`), and a new seam ledger `test_the_LAUNCHER_still_prepends_what_this_command_omits` reds if the launcher stops prepending `--validate`, stops prepending `--store`, or stops appending the caller's argv. A relationship between two files that nothing else asserted.

## Evidence

**Closing condition MET:** `grep -c 'subsystem_touch.py --validate' claude/skills/` → **0** at this tree.

**Mutation battery: 6/6 KILLED BY THEIR INTENDED TEST**, each required to fail with its *own* assertion's message (not merely "a test failed"); harness positive-control watched green on the pristine tree first; `PYTHONDONTWRITEBYTECODE=1`; tree diffed byte-identical afterwards.

🔴 **One mutant SURVIVED the first run, reported rather than hidden, and the fix is the lesson.** Dropping `--store` was invisible because the test modelled the launcher's default with the **same value** the command emits — the parse yielded the right root either way. It dies only against a sentinel the caller's store can never equal. A fixture whose fields are not pairwise distinct cannot see the mutant that collapses them.

**End-to-end on the real operator path** (which the unit tests deliberately cannot do): a real malformed entry produced a refusal, and the command it printed was executed as a shell command — rc **3**, `store:` naming the passed store, and it named both the offending file and the reason.

**Full suite** (`nix develop` → `scripts/run-tests.sh`): `scripts/tests` **1 failed, 14237 passed**; collector 273/104/231/13/13/235, dl-router 1022 — all green.

🔴 **That 1 failure is INHERITED, proven rather than assumed.** `test_every_site_writing_its_OWN_runner_bound_is_in_the_ledger` fails identically at the base `8114a124` in a clean worktree (`1 failed, 4 passed`). It concerns `test_scoped_tests_shared_surface.py`, which this diff never touches, and it is already claimed (`devrc-main-red-runner-bound-ledger`) with **#1567** open for it.

## Also: this handoff had drifted from the tree

`State now` still said rank 3 slice 3 was NOT MERGED with its claim HELD — it merged as `44bd8b0e` and the claim is released. Rank 3's own half-2 line still said NOT STARTED. Both corrected, verified **by content, not ancestry** (all five forked reader modules `ABSENT` on `origin/main`).

Rank 18 predicted slice 3 landing would close both (a) and (b). Re-measured: it closed **(b) only** — `host_identity.py` is gone. **(a) is not closed and slice 3 could not have closed it**: `nix/sessionVariables.nix:36` is a `.nix` literal, a surface that diff never touched. Left open with a revised condition rather than closed unread.

⚠ A **third site of the same class**, found while fixing (b) and deliberately **not** fixed: `subsystem_touch.py:3991` and `:4712` emit `python3 {SELF_PATH} --template …`. Outside this item's scope, and unlike `--validate` it has no one-word remedy — there is no `cairn-template` launcher — so closing it means first deciding whether to add one. Recorded with a closing condition so it reads as known-and-open rather than missed.

https://claude.ai/code/session_015rr3KiEvhZdjW9YZR6cb3t


ZacxDev:
## 🔴 Correction to this PR's own description — round 0 refuted my stated rationale

Posting rather than silently editing the body, since a reviewer may already have read it.

**The code change stands. The REASON I gave for it was wrong**, in the body above, in the first commit message, and in the docstring. Retracted in `e4e3b930`.

I claimed `Path(__file__)` makes the command *"true for the machine that printed it and false for anyone who pastes it anywhere else"*. Both halves fail:

- The old spelling emitted an **absolute** path — so cwd was never the failure mode.
- `nix/home.nix:1521-1522` deploys the launcher as `mkOutOfStoreSymlink "${workspace}/devrc/scripts/cairn-validate"` with `workspace = "${home}/workspace"`. **Any host where `cairn-validate` resolves at all necessarily has this checkout at that same absolute path** — the old command would have worked there too. If anything the new spelling's precondition is *stronger*: it needs a home-manager switch and a deployed pin, where the old one needed only python.

**The real defect, measured:** `Path(__file__).resolve()` names the **running copy**. Run out of a throwaway worktree — this repo's standing default for any file-modifying agent — the old code emitted `python3 /tmp/wt-cairn-rank23/scripts/lib/subsystem_touch.py …`, a path about to be `worktree remove`d. The recovery command went stale the moment the session that printed it ended.

## A guard I wrote has been deleted, on measurement

`test_the_LAUNCHER_still_prepends_what_this_command_omits` claimed *"nothing else asserts it … which is a silent green"*. **False.** Each of its three mutations, run against `test_cairn_flake_pin.py` **alone** with `test_subsystem_touch.py` deselected:

| mutation | result |
|---|---|
| launcher stops prepending `--validate` | **3 failed** |
| launcher stops prepending `--store` | **1 failed** |
| launcher stops appending caller argv | **2 failed** |
| pristine control | 15 passed |

Those tests run the **real** launcher as a subprocess and read both streams, so they hold in the `nix build` tier too. Mine was **spelled** — baked double quotes, so falsely red on a legal refactor and green on a literal in a comment. A second, weaker copy of an existing guard reads as coverage while providing none.

## `--store` kept, but no longer claimed to be free

`store` is `args.store`, whose default is `DEFAULT_STORE_ROOT` — the **frozen pre-cutover mirror**, not the synced cache the launcher would otherwise pick (measured: mirror **161** entries, cache **244**), and `subsystem-index/SKILL.md`'s mandated invocations pass no `--store`. Keeping it is *faithful* — the malformed file really is in the store that was read — but it inherits an unanswered question (why does the writer default to the frozen mirror?) that this PR must not be read as settling.

## Filed, not fixed — the closing condition is SPELLED and the class is open

`grep -c 'subsystem_touch.py --validate' claude/skills/` → 0 is genuinely met. But `command grep -rn '/home/zach/workspace/devrc' claude/skills/` returns **21 occurrences across 9 files**, including four literal `python3 /home/zach/workspace/devrc/scripts/lib/subsystem_touch.py …` invocations in **`claude/skills/subsystem-index/SKILL.md:73, 93, 113, 218`** — the write-protocol skill itself. No scanner gates this class. Recorded in the handoff with a closing condition (a mechanical gate over `claude/skills/**`, watched red-then-green on a planted violation) and **no owner** — it is a new ranked item, not part of rank 23.

Ledger: `round 0 · requirements: 9 (unattributed: 4) · deletion candidates: 5`. Round 0 reports and does not move the ladder.

ZacxDev:
## Round 1 — claims block

Round 1 returned **zero 🔴, zero 🟡, two 🟢** and a verdict of safe to merge. Both nits were fixed rather than filed, because one was a false claim this PR itself introduced. Block for the next round to read:

```audit-claims round=1 audited=899f0de8..7fb5e00e
1. The refusal now names a remedy for the emitted command ITSELF failing: a host that has not deployed the launcher gets exit 127, and the RECOVER block now says so and names home-manager switch.
2. That remedy line spells NO flag and NO checkout path — a first draft offered `python3 <devrc>/scripts/lib/subsystem_touch.py --store … --scope … --validate` as a fallback and was withdrawn before commit.
3. The two raw line counts in the handoff (subsystem_touch.py 6,654 / entry_shape.py 264) were REMOVED rather than corrected, and replaced with the order-of-magnitude claim plus the commands to derive them.
4. Rank 20 half one recorded as MET and OBSERVED (tekton/devrc-cairn-client-runs reported pass on this PR); half two explicitly recorded as still UNMET with a closing condition, and the "never executed in-cluster" caveat retired.
5. The branch was rebased onto origin/main 663bc86a; the previously-red tekton/devrc-pytests leg was the inherited test_runner_bound_ledger failure, fixed upstream by #1567.
```

**Anchor note for round 2.** `<from>` is `899f0de8`, not the `e4e3b930` round 1 literally read — the branch was rebased between the two and `e4e3b930` is no longer reachable on it. `899f0de8` is the post-rebase equivalent of the tree round 1 read, so `899f0de8..7fb5e00e` is exactly "what changed since round 1 read this PR", with main's bring-in excluded. Round 1's own report is at the comment above; its scope was the pre-rebase content, which is byte-identical for the four files this PR touches.

ZacxDev:
## Round 2 — claims block

Round 2 audited only `899f0de8..7fb5e00e` (my round-1 fix commits) and found **three real 🟡s in them** — all in prose or in a claim about a guard, which is the shape this ladder says a fix round produces. Fixed in `2dfc5f7b`.

```audit-claims round=2 audited=7fb5e00e..2dfc5f7b
1. The remedy is now spelled `home-manager switch --flake ~/workspace/devrc --impure` — the exact spelling cairn_pin.ensure() emits — replacing a bare `home-manager switch` that cannot evaluate nix/home.nix (required cairnPackage, no default).
2. "this checkout" removed from the remedy: the launcher is an mkOutOfStoreSymlink to a hardcoded ~/workspace/devrc, so which tree you switch is irrelevant, and the printing tree is typically a throwaway worktree.
3. The comment claiming test_the_command_is_BUILT_not_typed forbids the withdrawn fallback was NOT softened — the guard was WIDENED instead: new test_the_refusal_NAMES_NO_running_copy_path pins the rendered refusal rather than one function's return value, and kills the re-added fallback.
4. New test_the_refusal_NAMES_THE_REMEDY_for_the_command_itself_failing guards the remedy line (nothing did), and asserts cairn_pin.py still spells the remedy the same way so the two operator-facing messages cannot drift.
5. The surviving `6,654` / `264` at the fork-decision bullet were removed — the round-1 block declaring them "GONE ON PURPOSE / do not re-insert" was false about its own file 27 lines below it.
6. The replacement figures were corrected as unreliable in their own right: 6,693 was re-parented by the rebase one commit earlier in the same range, and entry_shape is 330 at a local clone HEAD vs 314 at the rev flake.lock pins. The paragraph now requires naming the rev.
7. Narrowed the overclaim "the absolute-checkout spelling this whole change removes" — true of the RECOVER command, false of the file, since `python3 {SELF_PATH} --template` survives and two tests REQUIRE it.
```

**Evidence.** 948 passed. New-guard kill matrix **3/3 by their intended test**: the withdrawn fallback re-added (*the regression the old guard let through* — N1), the remedy line deleted, the remedy losing its flags; positive control green on the pristine tree first, `PYTHONDONTWRITEBYTECODE=1`, tree restored byte-identical. Refusal rendered live to confirm the corrected line.

⚠ **The widened guard immediately caught my own comment describing it** — quoting the token literally reds the suite, because the guard reads that function's source. The comment says so rather than quoting it.

**Not fixed, and deliberately so:** the two 🟢s round 2 left out as nits (bash-vs-zsh word order in the parenthetical; the bare `#786` cross-repo reference qualified 25 lines above). Both change nothing a reader does.

PR #1585, opened 2026-09-12
fix(browser-bridge): the [INT] split verdict is a STARVED INSTRUMENT, not a handler defect

ZacxDev:
Closes rank 7 of `handoff-alert-recall-and-skill-consumers.md` — *"decide the devrc split-verdict test `test_browser_agent.py::test_the_release_handler_EXITS_rather_than_resuming[INT]`"*.

## The decision: it is a TEST defect, not a code defect

The handoff's leading hypothesis was *"genuine parallelism-sensitivity in the signal handler's lock release"*, with the caveat *"NOT established: nobody has read the handler against the failure."* That hypothesis is **refuted**.

## Mechanism, reproduced verbatim

The test spawns the wrapper, waits for the warm marker, then `killpg`s it. Injecting a stall between those two points splits the outcome into three bands — only the first is a verdict on the handler:

| stall | rc | what actually happened |
|---|---|---|
| < ~3 s | 143 / 130 | the trap ran — the real observation |
| ~3 s .. ~12 s | 2 | the wrapper had **left the warm**; the signal killed its tool-set gate instead, so `rc != 0` **passes for the wrong reason** |
| > ~12 s | **0** | the wrapper had **already exited**. It is a **ZOMBIE**, so `os.getpgid` still resolves and `killpg` still succeeds (the signal is discarded), and `proc.wait()` hands back the stored status **0** |

The third band emits the reported message verbatim — `exited 0 after a INT — the handler released the lock and let the run CONTINUE unserialised instead of terminating` — naming a regression that did not happen. The test never established that its signal reached a **live** wrapper still inside the warm.

## Why "reproduced 2/2, so not a flake" did not settle it

The handoff ruled out load by reproducing twice in the same minutes. That is two samples of one box state, not a discriminator. Corroboration that no code moved: **neither `browser-agent` nor this test file has a commit since before the failure was recorded**, and the documented repro command at `origin/main` `337114e0` today gives **914 passed** — which is the target's full collection, so the test ran (not a vacuous green).

## The fix

`_warm_window_lost()` runs immediately before the `killpg` and reads `/proc/<pid>/stat` — **never `proc.poll()`**, which reaps the child and manufactures a fourth failure shape (`ProcessLookupError`); that one bit me while instrumenting. A lost window is retried up to 3 times; giving up is a **loud** failure naming starvation and stating that it is not a verdict on the handler. This also closes the silent rc-2 band.

Nothing here can green a real regression: a wrapper that resumes after the signal is alive and inside the warm at kill time, which is exactly when an attempt counts.

## Verification

- **control** — `2 passed in 6.91s`; unchanged cost on the happy path.
- **mutation, isolated to INT** — `trap '_oc_lock_release; exit 130' INT` → `trap '_oc_lock_release' INT`: **KILLED**, `1 failed, 1 passed`, by **this test's own** `exited 0 after a INT` message, with the new line reading *"the INT landed inside the warm (attempt 1; hold 3s), so this IS a verdict on the handler"*. TERM stayed green.
- **negative controls, injected stall** — 6 s: all 3 attempts report *"the warm lock was already released"* (previously a silent wrong-reason PASS). 12 s: all 3 report *"already a ZOMBIE"* (previously the false handler regression). Both fail as **STARVED INSTRUMENT**, neither as a handler defect.
- **full target, the documented repro command** — `pytest scripts/browser-bridge/tests -n 4 --dist loadfile` → **914 passed**.
- `scripts/scoped-tests.sh` → `SCOPE: SCOPED (1 file(s) across 1 of 29 hermetic target(s))`, `RESULT: PASS (exit=0)`. ⚠ **Not a gate** — `gate.sh` exits 91 = PARTIAL off any non-`FULL` scope. Neither tier was run in full; CI evaluates the whole-target expectations this run suspended.

## Left open, deliberately — filed, not fixed

The sibling `test_a_run_killed_mid_bootstrap_RELEASES_the_warm_lock` has the **same** exposure and its own comment already names it: *"a TERM landing after the release finds the lock already gone, so `not lock.exists()` passes for the WRONG REASON."* It asserts absence, so a starved window makes it pass **silently** rather than fail loudly — strictly worse to read, and invisible to any count of reds. `_warm_window_lost()` is what it needs. Out of rank 7's scope.

**Closing condition:** a merged PR in which that test consults `_warm_window_lost()` before its `killpg`, watched to go red when the window is starved out.

https://claude.ai/code/session_01NRDTqApwXvg5JwrLkYEdd5

### DIFF
--- a/scripts/browser-bridge/tests/test_browser_agent.py
+++ b/scripts/browser-bridge/tests/test_browser_agent.py
@@ -1636,6 +1636,67 @@ def test_a_run_killed_mid_bootstrap_RELEASES_the_warm_lock(rig):
         "the ~120 s stall this change removes")
 
 
+# How long the fake holds the warm open, and how many times a starved attempt is
+# retried before the test gives up. 🔴 THE RETRY IS NOT FLAKE-PAPERING — a lost
+# window produces NO observation of the handler at all, so retrying is what keeps
+# a machine-load event from being reported as a code defect. The give-up path is
+# a LOUD failure that says which it was; nothing here can turn a real regression
+# green, because a wrapper that resumes after the signal is alive and inside the
+# warm at kill time, which is exactly when the attempt counts.
+_WARM_HOLD_S = 3
+_WARM_WINDOW_ATTEMPTS = 3
+
+
+def _warm_window_lost(proc, lock: Path):
+    """-> None if a signal sent NOW lands on a wrapper still inside the warm,
+    else the reason it would not.
+
+    🔴 MEASURED 2026-09-12, and it is the whole reason this function exists.
+    With the wrapper's `FAKE_OC_DEBUG_SLEEP` hold at 3 s, injecting a stall
+    between "the warm marker appeared" and the `killpg` splits into three bands,
+    and only the first is a verdict on the handler:
+
+      stall < ~3 s   rc 143 / 130   the trap ran — the real observation
+      ~3 s .. ~12 s  rc 2           the wrapper had LEFT the warm; the signal
+                                    killed its tool-set gate instead, so the
+                                    `rc != 0` assertion PASSES for the wrong
+                                    reason and proves nothing
+      > ~12 s        rc 0           the wrapper had already exited. It is a
+                                    ZOMBIE, so `os.getpgid` still resolves and
+                                    `killpg` still succeeds — the signal is
+                                    discarded — and `proc.wait()` hands back the
+                                    stored status 0. The assertion then fails
+                                    with "exited 0 … the handler released the
+                                    lock and let the run CONTINUE", naming a
+                                    regression that did not happen.
+
+    That third band is the reported `[INT]` split verdict: red under the suite's
+    own `-n 4 --dist loadfile` on a loaded dev host, green alone, green in the
+    nix sandbox — with no commit to the wrapper or this file in between.
+
+    ⚠ `proc.poll()` MUST NOT be used for this. It REAPS an exited child, which
+    removes the zombie and makes the very next `os.getpgid` raise
+    ProcessLookupError — a third failure shape, manufactured by the instrument.
+    Read `/proc/<pid>/stat` instead; it observes without reaping.
+    """
+    stat = Path(f"/proc/{proc.pid}/stat")
+    try:
+        # `comm` may contain spaces and parens, so the state field is the first
+        # token AFTER the final ')' — never `split()[2]`.
+        state = stat.read_text().rsplit(")", 1)[1].split()[0]
+    except (FileNotFoundError, ProcessLookupError, IndexError):
+        return "the wrapper was already gone before the signal"
+    if state == "Z":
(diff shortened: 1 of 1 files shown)


ZacxDev:
## Merging — ladder record, and what is NOT established

Six rounds (0 + five delta). Merging at `d0b75715` on the operator's call **with CI pending**, not green — stated plainly because those are different things.

### What the ladder found

| round | findings | the defect |
|---|---|---|
| 0 | 3 🟡 | the fix silently **deleted a hard precondition** (`assert lock.is_dir()`), rerouting a real lock-ordering regression into "read this as machine load"; a predicate open-coded over `_process_gone`; machine load asserted, not measured |
| 1 | 1 🔴 | the `defect` verdict **fired on purely starved runs** — the fake's `debug agent` branch serves the warm *and* the tool-set gate, so the gate reset the mtime clock after the lock was released |
| 2 | 4 🟡 | scoping created an **unpinned coupling** to the wrapper's argv; the retry multiplied the pathological wall time ×3; a "second sleep removed" saving that never existed |
| 3 | 1 🔴 | the guard added in round 2 was **spelled, not structural** — green in 4 of 6 mutant shapes, with a docstring claiming to assert "both sides" while reading one |
| 4 | 3 🔴 | **CI was red for three commits and nobody looked**; the whole diagnostic subsystem had **no positive control** (4 mutations deleted it with the suite green); a load-sensitive assert whose failure text is a false accusation |
| 5 | 1 🔴 1 🟡 | **three edits asserted in a commit message were never written** (multi-edit script wrote only at the end); deleting the timing assert **did** lose coverage — M9 proved it |

🔴 **Every round after round 0 found its defect in code or prose the previous round wrote while fixing the round before it.** That is the documented pattern, hit six times consecutively, and it is the most reusable thing this PR produced.

### Final state, measured

- Mutation matrix, re-run at head: **M5 / M7 / M8 / M9 all killed**, control 68 passed. M9 is the one that survived at the round-5 audited head.
- `test_runtime_shebangs.py` **9 passed** — the CI red this PR caused is fixed **at the root**, with no allowlist exemption added (`git diff origin/main...HEAD -- scripts/tests/test_runtime_shebangs.py` is empty).
- browser-bridge target **916 passed**.

### 🔴 NOT established — read this before trusting the merge

1. **CI was PENDING at merge time.** Not green. The prior head was red on `test_tmux_reply_agent.py::test_a_cwd_that_does_not_exist_is_REFUSED_not_opened_in_HOME` — a `real_tmux` test with no import path from this diff, passing here on the dev-host tier, with `main` green (22821 passed). That makes it **unattributed**, not cleared: the red tier is the sandbox one and it was never run against this head.
2. **The merged tree is ungated.** The branch was ~39 commits behind `origin/main`; `strict` is off by design, so a green check is a claim about the branch, never about the tree the merge creates.
3. The three-band table describes the **unscoped** fake and is historical; current bands are beside it.

Detection after the fact is `main-green-check` (4-hourly, reproduces a red before alerting) plus `main-status-watch`.

PR #1586, opened 2026-09-12
fix(tmux-restore): make a mid-session no-server refusal LOUD (once per incident), and stop the staleness gate counting the crash's own damage

ZacxDev:
> **This body was rewritten after round 1.** The original version described two items and asserted two things that turned out to be false — that a tmux socket disappears with its server, and that agent processes all stop writing to the ledger when the server dies. Both were refuted by measurement; both are fixed in `b627b094`. The per-finding ledger is in [this comment](https://github.com/innovation-upstream/devrc/pull/1586#issuecomment-5644401406).
>
> **Retention (192 → 672) has been SPLIT OUT to #1588** and reverted here (`e6af2432`). It is the operator's explicit ask, the audit found nothing against it, and it has no dependency on this rework — it should not wait on it.

🔴 `scripts/tmux-session-restore.py` **runs from the working tree** (the systemd unit and the continuum post-save hook both exec `%h/workspace/devrc/scripts/tmux-session-restore.py`), so `ship.sh`'s fast-forward *is* the deploy. Developed in a separate worktree; nothing written to `~/workspace/devrc`.

---

## The incident

MEASURED 2026-09-11: the tmux server died at 02:36:44 taking ~52 live claude conversations. The restore chain failed to recover them for **8.5 hours**, because two individually-correct refusals compounded.

```
02:36:43  Starting Resume claude conversations when the tmux server's socket appears...
02:37:14  no tmux server is running — REFUSING to restore.           (rc 0, silent)
...nothing for 8.5 hours...
11:00:02  restore plan is out of step with the saved layout by 8.6h
          (limit 2.0h, basis=layout) — too stale, skipping. Run `save` first.
```

## The one measurement that settles both fixes

Identical fixture, shaped like the preserved evidence, driven through three versions of the module:

| version | staleness on the incident shape | three consecutive mid-session activations |
|---|---|---|
| `origin/main` | 8.580h `basis=layout` → **REFUSES** | `[0, 0, 0]` — the 8.5h of silence |
| round 1 (`d42e78ae`) | 8.580h `basis=layout` → **REFUSES** *(the fix did not fix it)* | `[75, 75, 75]` — standing DND-bypass toast |
| **HEAD (`b627b094`)** | 0.223h `basis=ledger` → **passes** | `[75, 0, 0]` — one alert per incident |

---

## (a) A mid-session no-server refusal is LOUD — once per incident

**The refusal itself is unchanged and still correct.** Creating a server inside the unit's cgroup is what lost 43 conversations on 2026-09-06.

**The diagnosis.** `no_server_is_mid_session()` compares the pointer's mtime against uptime. `cmd_save` writes a generation only when it found live claude panes, so a plan written *inside this boot* proves a workspace existed during this boot and does not now — a cold boot cannot produce it. Uptime alone is a fact about the host, not the workspace; resurrect's `last` is the wrong witness for the same reason (b) is about.

### 🔴 What round 1 got wrong, and what makes the arm safe now

`nix/home.nix` wires `OnFailure=notify-failure@%n.service`, whose toast defeats do-not-disturb, and its rule is precise: *"any unit that can fail on a STANDING condition breaches it again"*. Round 1 argued the loud arm was an **event** because `PathChanged=` fires once per socket change. **That rests on the socket disappearing when the server dies, and it does not:**

```
tmux -L rgprobe1 new-session -d ...   ->  socket present, has-session rc 0
tmux -L rgprobe1 kill-server
ls /run/user/1000/tmux-1000/rgprobe1  ->  STILL THERE
tmux -L rgprobe1 has-session          ->  rc 1, "no server running on ..."
```

7 of the 8 sockets in the operator's runtime dir were orphans. So `ConditionPathExists=` keeps passing for the rest of the boot, every activation reaches the refusal, and `plan-written-this-boot` stays true — with `X-Restart-Triggers` restarting the unit on every home-manager switch (14 activations on 2026-09-11, four inside 9 seconds) and `notify-failure.sh` having no throttle. That is the bypass trained away.

**The latch is what makes it an event.** `claim_midsession_alert()` claims an `O_CREAT|O_EXCL` marker in `$XDG_RUNTIME_DIR` — a tmpfs the OS destroys with the login session, so it is self-scoping and needs no cleanup path that could itself fail. **Keyed on the plan's mtime, not the boot:** 2026-09-11 had *two* deaths (00:05 and 02:36), and a boot-scoped latch would have silenced the second. After a recovery the save chain writes again, so the next incident carries a different mtime and gets its own single alert.

- **Cold boot still exits 0**, unchanged, on the standing condition — with a positive-control test so a hardcoded-75 guard cannot pass.
- **No runtime dir, or an unwritable one, WITHHOLDS the alarm** rather than firing it. "Once" cannot be promised there, and an unbounded toast through the bypass is worse than a missed one.
- **`SuccessExitStatus=75` was considered and rejected** in-source: it suppresses the toast entirely, which is the silence this arm exists to end.

Exit **75** (`EX_TEMPFAIL`), not 1: `ExecMainStatus` already means three unrelated things at 1 here, and `tmux-restore-observe.sh` records that field.

## (b) The staleness gate no longer counts the crash's own damage

The crash brought up a new **empty** server; continuum autosaved that degraded layout at 10:59; the good 02:23 plan then looked 8.6h "out of step" and the gate refused. The inverted-basis shape #1317 already fixed once for powered-off time.

`plan_staleness_hours` cross-checks the *"the plan stopped being refreshed"* story (`state > mt` only) against a witness the crash cannot refresh.

### 🔴 What round 1 got wrong

Round 1 used *the newest pane record* on the invariant "when the server dies they all stop writing at once". **Calling the shipped function on the evidence directory its own docstring cited returns `11:03:34`** (`claude-p2.json` — a *recovery* pane), giving `worked` 8.672h against `gap` 8.6h, so `worked < gap` was false and **the gate still refused**. An agent in a pane of the *new* server writes `claude-p<N>.json` like any other, and pane ids restart at `%0` so it even overwrites the dead generation's file.

**The generation is what makes it freeze.** `build_plan` now records `tmux_pid` per entry; the witness keeps only records of that generation — the same fact `ledger_binding`'s generation check already rests on. Grouping the snapshot:

| tmux_pid | records | newest |
|---|---|---|
| 4025325 | 60 | 2026-09-06 17:25:30 |
| **4063373** | **56** | **2026-09-11 02:36:40** ← the plan's; server died 02:36:44 |
| 1111077 | 31 | 2026-09-11 00:05:46 *(crash 1)* |
| 627687 | 9 | 2026-09-07 21:54:13 |
| 1820124 | 2 | 2026-09-11 11:03:34 ← post-crash |

A post-crash agent is in a different generation and **cannot** move the witness, however long the recovery takes. Degradation is one-directional: as the new server overwrites more pane files the surviving record can only move *earlier* (more permissive), and if all are gone the witness is `None` and the gate falls back to refusing.

Still pane-records-only — `claude-s-<id>.json` records come from agents with no `$TMUX_PANE`, i.e. from the investigation of the outage itself (**158 pane / 94 non-pane** in the snapshot; an earlier "125/105" was the *live* ledger, the wrong directory).

### ⚠ Scope of this fix, stated in the source

A plan that does **not** record its generation — every plan written before this change, **including the preserved 02:23 plan** — measures nothing, and the gate behaves exactly as before. This makes a *replay* of 2026-09-11 recoverable, not that specific saved plan; its remedy is `restore --plan <generation>`, which bypasses the gate. Recovering the generation by session-id vote was measured (**54 / 11 / 4 / 1** on the real snapshot, contaminated by conversations resumed across generations) and **rejected as a heuristic**.

### Cases

| case | gap | witness | verdict |
|---|---|---|---|
| 2026-09-11 crash | 8.6h | 0.22h (gen froze at the crash) | **PASSES** — the fix |
| 2026-08-05 outage (plan Jul 5, layout Jul 29, same generation still live) | 576h | ~576h | REFUSES — preserved |
| same, mid-session | 576h | ~576h | REFUSES |
| layout OLDER than the plan (`state < mt`) | — | not applied | REFUSES — untouched |
| unmeasurable (no generation / no module / no records) | — | `None` | unchanged |

## ⚠ Not fixed, deliberately

The **`wall` basis** (unreadable `<resurrect-dir>/last`) returns before the cross-check, so the 2026-09-11 shape still refuses there — re-measured: `(8.600h, 'wall')`. Discounting it was tried and is unsafe: that path has no liveness term to pair with, so `max(gap, live)` is circular, and `min(wall, worked)` alone would wave through a plan whose workspace died weeks ago. There is genuinely no signal there saying "recover me". The reason is now in the source rather than the gap being silent.

The **documented inert liveness case** (`uptime <= limit`) is untouched — this change narrows contemporaneity only and does not subsume it.

## 🔴 A hazard I introduced and caught with my own control

The first latch tests wrote `alerted-<plan-mtime>` files into the operator's **live** `/run/user/1000/tmux-session-restore/` — three keyed on real plan mtimes, because those tests also read the real plan. Deployed, that is a test run **pre-claiming the latch and silencing a real incident's alert**. The four leaked files were removed (the deployed script has zero occurrences of `claim_midsession_alert`, so nothing else could have written them), and the fix is structural: an **autouse fixture** redirects `XDG_RUNTIME_DIR` for every test in the file, plus a guard pinning that the latch writes nowhere else.

## Red / green

Base **`663bc86a`**. `nix develop ~/workspace/devrc -c python3 -m pytest … -q`, `PYTHONDONTWRITEBYTECODE=1`.

| | result |
|---|---|
| base script + new tests | **50 failed**, 123 passed |
| HEAD, this file | **173 passed** (137 at base) |
| HEAD, restore chain (`+ observe + trigger + no_real_launchers + runner_bound_ledger`) | **327 passed** |

⚠ **15 of those 50 are pre-existing tests reddened by a fixture artifact**, not regressions — `_staleness_fixture` pins a symbol base lacks, so they die with `AttributeError` before reaching an assertion. Stated because the raw count flatters.

### 🔴 The two tiers, and the ambient-state control

The dev host has a live tmux server, a fresh restore plan, a populated agent ledger **and a live `$XDG_RUNTIME_DIR`**; the nix sandbox has none of them. `_boot_shape` pins mtime + uptime, `_staleness_fixture(pane_age_s=…)` pins the witness, and the autouse fixture pins the runtime dir. The mechanical control: **the whole file passes with `HOME` pointed at an empty directory**, and the real runtime dir is still absent after a full run.

## Mutation

29 mutants across the latch, the witness, the narrowing and the diagnosis. `PYTHONDONTWRITEBYTECODE=1`, restored from a `cp -a` pristine copy between mutants (never `git checkout --`), whole file per mutant (no `-k` filter), occurrence count asserted before each replacement. Green control **173** → restored control **173**. **29/29 killed**, including the two the round-1 sweep let survive (the `pane_filename` shape-change guard, and the 🟡-7 boundary).

⚠ One initially survived — **W2, "an empty generation matches every record"** — because every fixture record carried a pid, so the *inner* comparison rejected them and the outer guard was never the thing under test. That is the pairwise-distinct rule in its other form. Fixed by adding a record with no generation; re-run; killed. The hazard was real: the ledger holds records written before `tmux_pid` existed, and for a legacy plan `"" == ""` would have matched that whole cohort.

## Still not done

No reboot test — the cold-boot arm is fixture-covered only, and the operator has the boot-path reboot on hold. No nix change: `OnFailure=` was already wired and only the exit code moved.


ZacxDev:
Round-1 findings addressed in `b627b094`. Both 🔴 were independently re-measured here before being fixed; both held. Retention is **split out to #1588** and reverted on this branch (`e6af2432`).

The one measurement that settles both, same fixture, three versions of the module:

| version | staleness on the incident shape | three consecutive mid-session activations |
|---|---|---|
| `origin/main` | 8.580h `basis=layout` → **REFUSES** | `[0, 0, 0]` — the 8.5h of silence |
| round 1 (`d42e78ae`) | 8.580h `basis=layout` → **REFUSES** (the fix did not fix it) | `[75, 75, 75]` — standing DND-bypass toast |
| **HEAD (`b627b094`)** | 0.223h `basis=ledger` → **passes** | `[75, 0, 0]` — one alert per incident |

```audit-claims round=1 audited=d42e78ae..b627b094
🔴-1 FIXED — socket-outlives-server re-measured (`-L rgprobe1`: kill-server leaves the socket, has-session rc 1; 7 of 8 runtime sockets orphans). Loud arm now gated by `claim_midsession_alert()`, an O_CREAT|O_EXCL latch in $XDG_RUNTIME_DIR keyed on the PLAN'S MTIME (not the boot: 2026-09-11 had two deaths, a boot latch silences the second). Repeat activations exit 0. Measured [75,75,75] -> [75,0,0].
🔴-1 FIXED — the false paragraph rewritten: "PathChanged= is an event, so it fires at most once per death" deleted and replaced with the measurement that refutes it; the comment now records that this reason has been wrong three times and that the design no longer rests on a frequency claim.
🔴-1 FIXED — no runtime dir / unwritable latch WITHHOLD the alarm rather than firing it; "once" cannot be promised there. SuccessExitStatus=75 considered and rejected in-source (it suppresses the toast entirely).
🔴-2 FIXED — witness scoped to the plan's TMUX GENERATION. build_plan records tmux_pid; newest_pane_ledger_activity(generation, ...) keeps only that generation's records. A post-crash agent is in a different generation and cannot move it. Snapshot grouped by pid: 4063373 -> 56 recs, newest 02:36:40 (died 02:36:44); 1820124 -> 2 recs, newest 11:03:34.
🔴-2 FIXED — the false invariant ("when the server dies they all stop writing at once") is deleted and the refutation is recorded in the docstring WITH the 11:03:34 measurement and the worked=8.672h vs gap=8.6h arithmetic.
🔴-2 PARTIAL, SAID PLAINLY IN-SOURCE — a plan with no recorded generation (every plan written before this change, INCLUDING the preserved 02:23 plan) measures nothing and the gate behaves exactly as before. This makes a REPLAY recoverable, not that saved plan; its remedy is `restore --plan <generation>`, which bypasses the gate. Session-id-vote recovery was measured (54/11/4/1 on the real snapshot, contaminated by conversations resumed across generations) and REJECTED as a heuristic.
🔴 regression fixture REBUILT — INCIDENT_PANE_AGE_S is now derived from what the function actually returns for the plan's generation, with the per-pid grouping table in the test file. Round 1 hand-picked it from a record the function would not have returned.
🟡-3 FIXED — all three artifacts: the cmd_restore paragraph, tmux-restore-observe.sh's refusal block (now explains both arms and warns that ExecMainStatus=0 on a refusal no longer means cold boot), and test_tmux_restore_observe.py's two docstrings.
🟡-4 FIXED — "the unit's ConditionPathExists= catches the ordinary shape of that" deleted as false, with the measurement in its place.
🟡-5 FIXED — new `ledger-skew` basis; the skew arm no longer reports a ledger-discounted number under layout wording. Vocabulary is six arms, `why` covers all six, and a control test pins that an UNdiscounted skew still says `skew`.
🟡-6 NOT FIXED, DOCUMENTED — re-measured: same shape with no readable resurrect state file returns (8.600h, 'wall') and still refuses. Discounting the wall number was tried and is unsafe (no liveness term to pair with, so max(gap,live) is circular; min(wall,worked) alone waves through a plan whose workspace died weeks ago). The reason is now in the source instead of the gap being silent.
🟡-7 FIXED — pane_age_s=0.0 so the case overshoots the discount boundary instead of sitting on it (worked 600h vs gap 599h), and it asserts basis == "layout".
🟡-8 FIXED — test_the_refusal_is_checked_BEFORE_the_send_loop_creates_anything now calls _boot_shape and pins XDG_RUNTIME_DIR; it was the one call in the file reading the real restore-plan.json and the real /proc/uptime.
🟡-9 FIXED — sweep widened to 29 mutants across the latch, the witness, the narrowing and the diagnosis, including the pane_filename shape-change guard and the 🟡-7 boundary. 29/29 killed.
🟡-10 FIXED — "125 pane records against 105 session-keyed" was the LIVE ledger. Corrected to the snapshot's 158 pane / 94 non-pane, in both the source and the test.
🟡-11 FIXED — the remedy no longer offers `restore --best` (a dead end: same refusal, another 30s wait, same exit). It says start tmux, and notes that doing so re-triggers the unit.
🟢-13 FIXED on #1588 — 146 generations / 4 MiB corrected to the measured 192 / 6.53 MiB (the dir is at the old cap).
🟢-14 FIXED on #1588 — stated in the source, the README and a new invariant guard: retention widens MANUAL recovery only; --best still scans RECOVERY_ALERT_WINDOW=4 generations.
🟢-15 FIXED — the mtime-not-content caveat is stated in the witness docstring, along with why the record's own last_activity_ts is not used as the clock.
SELF-FOUND, NOT IN THE AUDIT — the first latch tests wrote alerted-<plan-mtime> files into the operator's LIVE /run/user/1000/tmux-session-restore/, three keyed on real plan mtimes. Deployed, a test run could pre-claim the latch and silence a real incident. Four leaked files removed (the deployed script has 0 occurrences of claim_midsession_alert, so nothing else wrote them); fixed structurally with an autouse fixture redirecting XDG_RUNTIME_DIR for every test in the file, plus a guard pinning the latch writes nowhere else.
CONTROL — the whole file passes with HOME pointed at an empty directory, so no test depends on the operator's plan, ledger or projects; the real runtime dir stays absent after a full run.
```

**Evidence trail**

- 🔴-1, causal: `tmux -L rgprobe1 new-session -d` → socket + `has-session` rc 0; `kill-server` → **socket still present**, `has-session` rc 1. Fleet: 7 of 8 sockets in `/run/user/1000/tmux-1000/` orphaned.
- 🔴-2, on the preserved snapshot: shipped `newest_pane_ledger_activity()` → `2026-09-11 11:03:34` (`claude-p2.json`). `worked` 8.672h vs `gap` 8.6h ⇒ discount never applied.
- Snapshot record counts: **158 pane / 94 non-pane** (252 files).
- Generation grouping (the fix's basis): `4063373` 56 recs newest `02:36:40`; `1820124` 2 recs newest `11:03:34`; `1111077` 31 recs newest `00:05:46` (crash 1).
- Legacy-generation recovery by session-id vote: `4063373`×54, `1111077`×11, `4025325`×4, `1820124`×1 — plurality, not certainty. Rejected.

**Tests** — 173 in `test_tmux_session_restore.py` (137 at base), 327 across the restore chain (`+ observe + trigger + no_real_launchers + runner_bound_ledger`). 50 of the new/changed tests are red at `663bc86a`; ⚠ 15 of those 50 are pre-existing tests reddened by a **fixture artifact** (`_staleness_fixture` pins a symbol base lacks → `AttributeError`), not regressions — same caveat as round 0, restated because the raw count flatters.

**Mutation** — 29 mutants, `PYTHONDONTWRITEBYTECODE=1`, restored from a `cp -a` pristine copy, whole file per mutant (no `-k`), occurrence count asserted. Green control 173 → restored control 173. **29/29 killed.** One initially survived — W2, "an empty generation matches every record" — because every fixture record carried a pid, so the *inner* comparison rejected them and the outer guard was never the thing under test. That is the pairwise-distinct rule in its other form; fixed by adding a record with no generation, re-run, killed.

Not done: no reboot test (the cold-boot arm is fixture-covered only), and no nix change — `OnFailure=` was already wired and only the exit code moved.

PR #1590, opened 2026-09-12
docs(gate-flake): file the prose-census class as rank 10 — and a third instance landed while writing it

ZacxDev:
Follow-on to **#1570** (merged `f013ac50`), whose closing observation was that the census class behind ranks 8 and 9 is owned by nobody. This files it with a closing condition — and the discovery step turned up more than expected.

## The class

A hand-maintained ledger asserted over **tracked text** reddens the shared branch for everyone. The cost is not the red; it is that **a document nobody has classified yet blocks a branch nobody else can unblock.**

## Three instances, measured — and the third arrived mid-filing

| | instance | status |
|---|---|---|
| (a) | `test_guard_core.py` kill-mention ledger — **22** PR-head reds plus `main`, ≥4 sessions in ~2 h | closed by **#1561** (`c0bbd6d9`), scoping the scanner off `claudedocs/` |
| (b) | `test_runner_bound_ledger.py` — **5** reds, 4 after `c0bbd6d9`, plus `main` | closed by **#1567** (`6f1867b1`), routing the unbounded spawn |
| (c) | `test_doc_path_rot::test_no_new_dead_paths` — a census over tracked prose in `claude/` | 🔴 **OPEN — `main` is RED on it now** |

🔴 **(c) is a FALSE POSITIVE, which makes it the best evidence in the item.** It reports `claude/skills/audit-pr/SKILL.md:144 DEAD PATH: reference/round-ladder-evidence.md` — and **the target exists on `origin/main`** (`git cat-file -e` → 0). The gate resolved the relative reference against a placeholder segment after **#1587** (`e8ee277e`) reworded around it, rendering `…/<bare-ref-unopenable-write-the-deployed-~-.claude-path>/reference/…`. **So the guard reddens the shared branch over a reference that is correct.** Measured in a clean worktree whose only diff is this `claudedocs/` file — which that gate does not read. Not claimed by anyone; no open PR addresses it (swept).

That widens the class: a census can redden the branch over something **not wrong at all**, which no amount of classifying would have prevented. And it needs a *third kind* of fix — #1561 narrowed a scanner's scope, #1567 removed the thing being counted, this one needs a resolution fix. **That is why none of the three generalises.**

## The discovery step failed by grep, twice over — and that is the finding

**47** test files under `scripts/` enumerate tracked files via `git ls-files`; a regex for the ledger constant matches **2** of them. Known instance (b) is caught by **neither**:

1. its constant is `_OWN_BOUND_LEDGER: dict[...] = {` — a **type annotation** between the name and the `=`, so the regex cannot see it;
2. it does not use `git ls-files` at all — `_discover()` walks `REPO_ROOT.glob("scripts/**/*.py")`, so it is not in the 47.

Two orthogonal escapes from one grep, found only by checking the one file I already knew belonged. **So no population figure for this class is defensible today**, and arm (a) below cannot be satisfied by a pattern match — it needs an AST walk over what the guards *do*.

## Closing condition — either arm, both checkable

**(a)** A guard that enumerates the repo's ledger-style censuses and **fails** when one scans tracked prose without declaring an exemption — shown RED on today's tree and GREEN after, with the enumeration **derived, not hand-listed** (rank 1's `leakscan` lesson: a hand-written allowlist is the same defect one level up). **(b)** OR the operator records in writing that per-instance fixing is the accepted cost, having read the three instances — then it closes as a decision and `forcing` drops to `none`.

**Filed rather than fixed**, because the remedy is a judgement about what these guards are *for*: a census over prose catches a real hazard — #1561's own history is docs quoting a wide kill command — and scoping every census off `claudedocs/` would close this class and reopen that. Not a trade to make silently.

## Scope / verification

One `claudedocs/` file, prose only. Every figure re-derived on the merged `origin/main`: 47, 2, the annotation and glob escapes, and (c)'s red with the target's existence both checked. ⚠ `main` is currently red on (c), so a gate run on this branch inherits it — that is instance (c), not this diff.

https://claude.ai/code/session_018rg3dNxWP65JYxjRujF23a


ZacxDev:
## Correction: instance (c) is a TRUE positive. I called it false from the file's existence.

The PR body above says (c) is a census **false positive** because the target file exists on `origin/main`. **That is wrong**, and the error changes what this item concludes — so it is corrected in the doc and here rather than quietly edited.

`_resolve`'s rule 1c (`test_doc_path_rot.py:306-315`) resolves a `reference/<file>` token to `REPO_ROOT / BARE_REF_MARKER / token` **deliberately**, where

```python
BARE_REF_MARKER = "<bare-ref-unopenable-write-the-deployed-~-.claude-path>"
```

— an unopenable segment **whose name states the remedy**. The in-source comment: resolving it against the citing doc's directory *"would always succeed and reproduce exactly the false-clean this rule exists to kill"*. A `SKILL.md` is loaded as authoritative guidance from an arbitrary cwd, so the property is **followability, not existence**.

🔴 **I asserted the verdict from the file's existence without reading the resolver** — the same shape as rank 6's two vacuous checks, which this very doc spent four audit rounds on. Fixed in **#1591** (one line; red at base, green at HEAD).

### What it does to the item

(c) is the census **working**: it caught a real, one-line, agent-facing defect **22 minutes** after it merged. So two of three instances were the guard over-reaching and the third was the guard earning its place.

That flips (c) from the item's strongest complaint to **evidence for arm (b)** of the closing condition, and makes the item's substance a *decision* rather than a grievance. The doc now states plainly that **a remedy which silences the class silences (c) too** — which is the actual trade. `forcing:` drops to **none**: all three instances are fixed (#1561, #1567, #1591) and nothing is red on this class today.

Everything else in the item stands and was re-derived: 47 / 2, the two orthogonal grep escapes, and the three instances' counts with their read-time caveat.

ZacxDev:
## Narrowed: rank 10 closes as a DECISION (arm **b**). Arm (a) proposed and rejected.

Recommendation asked for and given. Classifying the three instances by **what each actually scans** collapsed the class from three to one — and the collapse is the finding, so it is recorded rather than quietly rewritten.

| instance | scans | verdict |
|---|---|---|
| (a) kill-mention ledger | `claudedocs/` **prose**, ledger grows per document | **the only real instance** — 22 reds + `main`, ≥4 sessions, ~2 h |
| (b) runner-bound ledger | `scripts/**/*.py` by AST — **code, not prose** | true positive: a genuine new unbounded spawn |
| (c) `test_doc_path_rot` | `claude/` + `CLAUDE.md`, already **excludes** `claudedocs/` | true positive: a genuinely unfollowable reference |

And the convention is already applied everywhere it applies: `_PROSE_ONLY_PREFIXES` in (a) after #1561, `CORPUS_DIRS` in (c) which never scanned `claudedocs/`.

### The discriminator — the durable content of the item

> A census over tracked prose is **fine** when it enforces a CONTENT RULE — a new document simply must not contain the thing. It is a **hazard** when a new document requires a hand LEDGER ENTRY, because then someone must *classify* each new mention, and until they do a doc nobody has triaged blocks a branch nobody else can unblock. **A ledger whose population grows with every document must declare its prose exemption at birth.**

Measured **with a positive control**, so the zeros are readings and not artefacts: `test_no_public_ips` and `test_doc_path_rot` carry **0** hand-ledger constants; `test_guard_core` — the known growth-ledger — carries **1**. The grep distinguishes the two kinds.

### Why arm (a) is rejected

- 🔴 **It would be a census over censuses — this class one level up.** The item's own arm (a) forbade hand-listing the population, and the grep evidence shows why that is the hard part: it escaped a pattern match in **two orthogonal ways** (a type annotation before the `=`, and a `glob()` instead of `git ls-files`). A real arm (a) needs an AST walk over what each test *does*.
- **It would police a population of one, already remediated.** Nothing left to catch today.
- 🔴 **(c) is the argument against silencing the class at all** — a prose census caught a real agent-facing defect **within half an hour** of it merging. A remedy scoping every census off prose would have lost that.

### Residual — what the decision rests on and cannot prove

The population is **not enumerable by grep**, so *"no second growth-ledger exists"* is a claim about the instances checked — the three known plus the obvious prose scanners — **not a census**. A fourth behind a third spelling would not have been found. **If one surfaces it does not reopen this decision; apply the discriminator to it at birth.**

`forcing:` → **none**. All three instances fixed (#1561, #1567, #1591). 802 passed on the four doc gates. `MERGEABLE / CLEAN`.

PR #1594, opened 2026-09-12
fix(tests): the 3-hex fixture prefix was AMBIGUOUS 1 run in 585 — 0.1709% over a 4,096,000-timestamp enumeration (devrc#432)

ZacxDev:
## The defect

`scripts/tests/test_subsystem_touch.py::TestCommitMutationKillMatrix::test_kills_the_LENGTH_guard`
feeds a **three-character** sha prefix on purpose. 3 is below
`COMMIT_SHA_MIN_CHARS` (4), which is the entire point of the guard it exercises,
so *"use more characters"* was never an available remedy — it would have deleted
the coverage rather than fixed anything.

But a 3-hex prefix has 4096 values and the fixture repo holds **eight** objects,
so the draw was sometimes **ambiguous**. git then raised `CommitAmbiguousError`,
the mutant died of the **wrong** guard, and the assertion under test never ran.

It reddened PRs whose diffs could not reach this file — #1582 most recently.

## Rate — measured, and stated at the scope it was measured

**0.1709%** — 7001 hits in a window of **4,096,000 consecutive commit
timestamps**.

The commit timestamp is the only thing that moves the sha here (content,
identity and message are all fixed), so enumerating timestamps enumerates the
randomness rather than sampling the test. The object model behind the
enumeration was validated **object-for-object against real git 2.55.0** before
any of it was believed.

⚠ The window is still a sample of all timestamps. The analytic value is
`7/4096 = 0.17090%`; the window's expectation is exactly 7000 with σ ≈ 83.59, so
the observed 7001 sits **0.012σ** away. That is *consistent with* the analytic
value, **not** a confirmation of it to four digits — re-run the window and expect
a number near 7000, not 7001. An earlier revision of this PR body said it
"matches the analytic 7/4096 exactly"; that was wrong and is corrected here and
in the source comment (`0f021210`).

## The remedy — both halves

| half | what it buys |
|---|---|
| `_remint_until_short_sha_is_UNIQUE` | **corrects** an ambiguous draw by amending HEAD until its 3-hex prefix names exactly one object. A precondition alone would have turned a 0.17% *silent* flake into a 0.17% *loud* one — still red CI for changes that cannot reach it. |
| `_assert_short_sha_is_UNIQUE` | **refuses by name**, with its own distinct sentinel, so a draw the re-mint somehow failed to fix can never masquerade as the guard passing or failing. |

Amending rewrites the commit object only — tree and parent carry over — so the
commit's **diff**, which every caller asserts on, cannot move. Pinned by
`test_the_REMINT_does_NOT_change_what_the_commit_CHANGED`.

**Applied at three sites, not one** (`claude/RULES.md` "one rule, one place").
The same latent ambiguity sat open-coded in
`test_the_LENGTH_bound_is_not_vacuous_git_WOULD_have_expanded_it` (same 0.17%)
and `test_the_boundary_is_a_boundary_not_a_slope` (0.011%, via its 4-hex
prefix). Only one of the three had ever been seen to fire.

## Evidence

**Reproduction — forced, not waited for.** 600 timestamps *known* to draw an
ambiguous prefix, driving the verbatim body of `test_kills_the_LENGTH_guard`:

| tree | result |
|---|---|
| `origin/main` fixture construction | **600/600 FAIL**, all `CommitAmbiguousError` |
| this branch's `_repo_with_unique_short_sha` | **600/600 PASS** |

Power 1.0 against this defect — the adversarial draw is constructed, not hoped
for. This is the load-bearing evidence, not the wall-clock loop below.

**Red-at-base / green-at-HEAD**, same forced collision, same test:

| tree | result |
|---|---|
| `origin/main@10ae4da0` + forced collision | 1 failed — `CommitAmbiguousError: '68d' names 2 objects` |
| this branch + the identical forced collision | 15 passed |

**The 600-iteration wall-clock loop** (the stated closing condition): **600/600
pass, 0 fail, 2534 s wall.**

⚠ **State its power honestly: 64.2%, not ~100%.** At p = 0.001709, 600
*independent* draws have `1 − (1−p)^600 = 0.642` probability of showing at least
one failure, so a green run of this shape had a **35.8%** chance of happening
even with the bug still present. It corroborates; it does not establish.

⚠ And **600 iterations are not automatically 600 draws.** The fixture's content
is fixed, so iterations landing in the same second mint the *same* sha. This run
averaged 4.2 s per iteration, which makes distinct seconds very likely — but I
did **not** record the fixture commit timestamps, so "600 independent draws" is
inferred from spacing and is not measured. The number I reported from the
script's own `distinct_seconds_spanned` field is just wall-clock seconds under a
misleading label; it is not a count of distinct draws.

That caveat is not hypothetical: a 3000-build real-git sample returned **0/3000**
over roughly 412 distinct seconds — expected failures 0.70, `P(0) = 49.4%` —
which is entirely consistent with the rate and is evidence of nothing on its own.

**Mutation sweep** — 8 mutants, `PYTHONDONTWRITEBYTECODE=1`, no-op control
GREEN, **7/7 KILLED**, each by its named test. Re-run after the prose correction
in `0f021210`, with identical results:

| mutant | killed by | reason |
|---|---|---|
| precondition branch → `if False` | `..._refuses_an_ambiguous_fixture_BY_NAME` | `DID NOT RAISE` |
| sentinel reworded | same | pinned literal |
| re-mint loop → `range(0)` | `test_the_REMINT_CONVERGES_from_a_forced_collision` | the precondition's **own** sentinel |
| re-mint never re-reads HEAD | same | same sentinel |
| prefix reader sees nothing | `test_the_COLLISION_fixture_really_DOES_collide` | positive control moves |
| collision search returns junk | same | same |
| object oracle narrowed to HEAD | same | same |

The two re-mint mutants dying to the precondition's own sentinel is what proves
the guard **reachable in fact**, not merely present.

One survivor was found and closed: the refusal test asserted the sentinel
*constant* against a message built from that same constant, so rewording it
stayed green — an expectation derived from the implementation. The literal is
pinned now.

**CI**, on head `79f3e8f5` (the tree before the prose-only correction):
`tekton/devrc-pytests` **pass** (22551 passed, 0 failed), `tekton/devrc-nodetests`
**pass** (1449/1449), `tekton/devrc-cairn-client-runs` **pass**. Not yet
re-reported for `0f021210`.

## What this does NOT do

The prefix handed to the guard is still **three** characters and the production
module still refuses it — `test_the_reminted_fixture_STILL_feeds_a_THREE_character_prefix`.
No coverage was traded for determinism.

## Considered and not taken — measured, then rejected

Pinning `GIT_AUTHOR_DATE`/`GIT_COMMITTER_DATE` in `_git_env` deletes the entropy
source outright in ~3 lines instead of ~290. **It was measured, not waved away:**
patched onto `origin/main` (`ce4aee8e`) in a scratch worktree with the anchor
asserted to occur once, the file then ran **933 passed**. The alternative works.

It is rejected on a stated ground:

- **A pinned date makes the fixture sha deterministic *forever*, so an unrelated
  fixture-content edit that happens to draw an ambiguous 3-hex prefix fails
  100%, not 0.17%.** That trades a rare flake for a latent permanently-red gate —
  the failure `claude/RULES.md` calls worse than no gate, and the one this PR
  exists to remove.
- It pins a dimension across all 938 tests in the file, and a suite whose config
  pins a dimension is structurally blind to that dimension's bugs.
- It removes **one** entropy source; the re-mint corrects **any**.

The counter-argument — that RULES prefers deterministic fixes over probabilistic
ones — is real and does not reach this case: determinism that can lock red is not
the determinism that rule asks for.

## D3 — the precondition is now *inside* the search

Review found `_remint_until_short_sha_is_UNIQUE` falling through to a silent
`return sha` on exhaustion, with six call sites not asserting afterwards.
Probability ~1e-66, so not a defect — but the wrong **shape** for a function whose
whole job is to guarantee a property. A guarantee living in a separate helper is
one you can forget to call.

`_assert_short_sha_is_UNIQUE` and `AMBIGUOUS_FIXTURE_SENTINEL` are deleted; the
search itself raises, and reaching its `return` **is** the guarantee.

🔴 **The property that would otherwise have been lost.** The old loop returned the
final amend's result **unexamined** — it checked `attempts` draws and handed back
an `attempts+1`-th nobody had looked at, and the deleted assert was silently
covering that. So the collapse adds a post-loop check: `attempts` bounds the
**amends**, `attempts + 1` **draws** are checked, the last included. Pinned
behaviourally by `test_the_LAST_draw_is_checked_and_the_amends_are_BOUNDED`,
which replaces the oracle with one that always reports ambiguity and then
**counts** — the oracle must be consulted 4 times at `attempts=3`, and the object
database must grow by exactly 3 (each amend orphans the commit it replaced, so
the amend count does not rely on the function reporting its own work).

**Two tests restructured so the oracle mutants stay attributable.** The positive
control and the seam test both called the re-mint first, which after this change
means a broken *oracle* raises before the control ever runs — the mutant would be
killed by the raise, and a guard that kills everything distinguishes nothing.
Both now skip the re-mint and assert a **delta** rather than an absolute count,
which also removes their own residual 1-in-585 dependency.

**Re-run battery — 9 arms, `PYTHONDONTWRITEBYTECODE=1`, control GREEN, 8/8 KILLED,
each by its OWN named victim** (checked against each mutant's full failure list,
not just the score):

| mutant | own named victim | its own reason |
|---|---|---|
| post-loop check removed | `..._RAISES_rather_than_returning_an_AMBIGUOUS_sha` | `DID NOT RAISE` |
| sentinel reworded | same | the literal assertion |
| re-mint loop `range(0)` | `test_the_REMINT_CONVERGES_from_a_forced_collision` | the sentinel |
| re-mint never re-reads HEAD | same | the sentinel |
| oracle sees nothing | `test_the_COLLISION_fixture_really_DOES_collide` | its delta assertion |
| collision search returns junk | same | its delta assertion |
| oracle narrowed to HEAD | `test_the_two_object_oracles_AGREE` | set comparison |
| amend bound off by one | `test_the_LAST_draw_is_checked_and_the_amends_are_BOUNDED` | the counts |

The two properties named in review, confirmed individually rather than inferred
from the score:

- **`attempts=0` reaches the raise** — refusal reads `after 0 re-mint(s)` and
  `names 2 object(s)`.
- **An empty oracle reaches it after the FULL 24 amends** — refusal reads
  `after 24 re-mint(s)` and `names 0 object(s)`.
- **The sentinel reword breaks the LITERAL assertion**, quoted from the run:
  `assert "FIXTURE PRECONDITION: short sha is AMBIGUOUS" in str(exc.value)`.

⚠ **It did not save ~50 lines, and that estimate was mine to check.**
`git show --numstat`: **136 added, 65 deleted — net +71.** The deletion was real
but smaller than the structural pin that preserving the last-draw property
required. The guarantee is stronger; the file is longer.

⚠ **And my first counts here were wrong, from my own instrument** — `125/62/+63`,
produced by `grep -c "^+[^+]"`, a pattern that silently drops every added or
removed **blank** line. Two methods, two answers, and the prettier one was mine.
`--numstat` is the authority.

**Full file after D3: 938 passed, 0 failed** (277 s).


ZacxDev:
Corrected the PR body in place. Two claims I had written were overstated and are retracted here rather than silently edited, since a reviewer may have read the earlier version:

1. **"matches the analytic `7/4096` exactly"** — it does not. The 4,096,000-timestamp window has expectation exactly 7000 with σ ≈ 83.59; the observed 7001 is **0.012σ** away. Consistent with the analytic value, not a four-digit confirmation. Fixed in the source comment too (`0f021210`), along with "exhaustive enumeration", which was exhaustive over the randomness *source* and not over the timestamp space.

2. **The 600-iteration loop's power was never stated.** It is **64.2%**, not ~100% — a green run of that shape had a 35.8% chance of occurring with the bug still present. The load-bearing evidence is the adversarial run (600/600 fail at base, 600/600 pass at HEAD, power 1.0), not the wall-clock loop. Relatedly, my script printed `distinct_seconds_spanned` which was just wall-clock seconds under a misleading label — it is **not** a measured count of distinct draws, and I have said so rather than quoting it.

Also corrected: the source comment cited RULES.md's *"a permanently-red gate is worse than no gate"*, which is the wrong bullet for an intermittent red. The fitting one is *"A flaky test is also fixable — remove the timing dependency rather than re-running.""

PR #1599, opened 2026-09-12
feat(peer-host): answer WHICH MACHINE a peer session is on, derived from live state

ZacxDev:
Closes clawgate task 562 — *"peer session listing has no host column, so a session cannot route work to the right machine"*.

## What shipped

`scripts/peer-host <selector>` — prints `workbench` or `laptop` on **stdout**, exits 0. Diagnostics go to stderr, so a caller can capture the answer directly:

```sh
host=$(scripts/peer-host datapacket-talos-ef) || exit
```

The brief allowed either shape (a host column in the listing, or a command that takes a name/ref and prints its host) and asked for whichever is cheaper against whatever already holds the cross-host view. This is the second shape, because the first is not ours to change — the listing is the harness's own agent list.

### Key set accepted (a superset of the brief's "name or ref")

| kind | example |
|---|---|
| peer **name** — the string the listing shows | `datapacket-talos-ef` |
| **session id**, full or a ≥6-char hex prefix | `a58cd02c-820f-…` / `a58cd0` |
| **pid** | `1394874` |
| **tmux address**, full or the `session:@window` prefix | `scratch2:@62.%62` / `scratch2:@62` |

Kinds are unioned, so one record hit by two kinds is still one match.

### Exit codes

| code | meaning |
|---|---|
| `0` | resolved — the host is on stdout |
| `2` | ambiguous — more than one live peer matched; candidates on stderr |
| `3` | unmatched — zero matched, and **every** in-scope host was read successfully |
| `4` | incomplete — zero matched but a host could not be read |
| `64` | usage (`EX_USAGE`) |

🔴 `64` rather than argparse's default `2` is deliberate: `2` is `ambiguous` here, so before the remap a mistyped flag was indistinguishable from a real *"this name matches two peers"* answer.

## Why not `session-manager`

The task pointed there first. It cannot answer this question, and the reason is **structural, not a coverage gap**:

* `grep -c 'claude/sessions' scripts/session-manager` → **0**. It never reads the harness session registry, and its rows carry `session` / `window_id` / `label` / `claude_session_id` but **no harness peer name**. There is nothing in its output to match a peer name against on any host, so no amount of coverage would make the lookup possible.
* `claude_session_id` is the only field that could bridge to a registry record, and I measured its coverage in one `--json --no-ch` run: **laptop 7 of 31 windows, workbench 50 of 53** — confirming the brief's figure exactly. Even at 100% it would still need the registry to map name → sessionId.

The registries are both complete (**49/49** local, **9/9** laptop in the same measurement) and cheap — no tmux scan, no ClickHouse, one SSH round trip, ~1.9 s end to end.

`session-resolve` already reads this registry and documents that it is local-only (its `PRESENCE_EXCLUDED` state is exactly *"this window is on another machine, so it can never carry a record here"*). This is the cross-host half of that.

## Criterion 4 — derived from live state, not a map

A record's host is simply **the host whose filesystem holds it**. There is no name→host table anywhere in the change, so there is nothing to edit when a session starts. Demonstrated below.

## 🔴 Two hazards the design answers rather than documents

**A wrong SSH address does not fail loudly.** `10.42.0.100` is the laptop; `10.42.0.10` is the homelab gateway — and it answers just as happily, reporting the wrong machine's sessions under the label you asked for. A constant cannot catch that, because the constant is the thing that would be wrong. So the remote program **computes its own label**, using `scripts/lib/host_label.py`'s source shipped over stdin rather than a second copy, and prints it in a sentinel line. A leg that self-identifies as another machine is **refused** (`label_mismatch`), never recorded. The sentinel doubles as the positive control that distinguishes "answered with nothing" from "the command was swallowed" (`no_sentinel`).

**An unreachable laptop must never become "workbench" by default** — the exact failure criterion 3 forbids. An unsearched host contributes **no matches**, so a laptop-only name yields zero matches, and zero matches with an unsearched host is `incomplete` (4), a distinct code from `unmatched` (3). When a *unique* match is found while another host went unread, that is a positive measurement rather than a default, so it is returned (0) with a loud stderr warning and `hosts_unsearched` set in `--json`. A caller that must refuse on incomplete coverage reads that field — there is deliberately no flag for it (a `--strict` flag was written and then deleted in round 0: it was a second spelling of a fact already published).

## 🔴 The bracketed `[ref]` is NOT supported

Two independent derivation sweeps found nothing:

* the task author's — 11 candidate registry inputs × 6 hash algorithms × 3 offsets, across all 48 records;
* mine — 9 inputs including the raw UUID bytes in both byte orders and `bridgeSessionId`, × 11 algorithms including CRC32/Adler32, against the one ground-truth `(name=datapacket-talos-ef, ref=ef79ae)` pair recoverable from the task text.

Zero hits in either. One observation I deliberately did **not** build on: in that single pair the ref's first two characters equal the name's suffix (`ef79ae` / `…-ef`). That is n=1 and 1-in-256 by chance — and if wrong it answers the **wrong host silently**, the one outcome this tool exists to prevent. `test_no_ref_selector_is_implemented` pins the absence on purpose, with instructions to delete it deliberately if a real derivation is ever established.

## One rule, one place

The peer SSH table moved into `scripts/lib/host_label.py` — the module that already owns "which box is this" and whose docstring exists because that rule was open-coded twice and the copies disagreed. `opencode_search.PEERS` is now **the same object**, not an identical literal (`test_opencode_search_and_peer_host_share_ONE_peer_table` asserts identity, not equality).

`scripts/session-manager` **derives** its `LAPTOP_SSH_TARGET` from the same table. The first cut left it alone behind an equality seam guard; round 0 showed that guard was the wrong instrument — both copies were already literal-pinned by their own tests, so equality only fires for an edit that changes exactly one of them. It is asserted **structurally** now (`test_session_manager_DERIVES_its_laptop_target_from_the_peer_table` requires an `ast.Call`, not an `ast.Constant`), so re-declaring the literal fails even if the address typed is correct.

The replacement ledger scan — `test_no_module_redeclares_a_peer_address_literal` — then found a **fourth copy on its first run**: `scripts/session-analysis/espanso-usage.py`, a hardcoded `DEFAULT_REMOTE`. That is derived now too. The scan reads the **syntax tree**, not the text, because the first cut flagged a comment that quoted the literal while explaining why the literal had been removed; it carries its own negative control (`test_that_ledger_guard_can_actually_fire`).

🔴 **CORRECTED — an earlier version of this PR body claimed "the peer SSH address table has exactly one spelling", and `host_label.py`'s comment claimed to list "every place these addresses are spelled". Both were FALSE when written**, and an over-claiming ledger is worse than none: it is what stops the next reader from looking. Two spellings remain **within the guard's scope** (`scripts/`, non-test, as a string constant), both now NAMED in that comment and in the guard's docstring — and many more outside it:

- `scripts/lib/host-role.sh:32-35,47-48` — bare IPs in shell, composed into `zach@10.42.0.100` at **runtime**. It is live (sourced by `ship.sh` and `drift-check.sh`) and `scripts/README.md` calls it "the ONE host-identity predicate" — so two modules currently each claim to be the single home. It is shell and answers a different question (which role am I, from a list of interface addresses), so folding it in is a design change, not a rename.
- `scripts/browser-bridge/server.py:653-654` — bare IPs in `_HOST_IP_ORDER`. Deliberately untouched: another agent is working in that directory.

The values all agree today, so this is a duplication hazard rather than a live defect.

🔴 **No unscoped census is asserted anywhere, deliberately.** An earlier version of this paragraph said "reduces independent spellings from 5 to 3" — but that `5` was the second draft's own count, and the *next* audit round measured 11+ once `nix/` is included. Stating a total is what kept regenerating the defect, so the claim is now scoped and the total is dropped: **within `scripts/`, non-test, as a string constant, this PR takes the count from 5 to 3 and enforces that mechanically.** Outside that scope — runtime-composed targets, bare IPs, `.md`, and everything under `nix/` — nothing is claimed and nothing is enumerated. Grep.

### Round 0 — requirements and deletion (`/audit-pr`)

Five findings, all acted on. Verdict was `deletion candidate`, not `proceed`.

1. 🔴 **Reachability — the deliverable did not reach its caller.** `scripts/peer-host` had no PATH entry, was named in no skill and had no caller, so the session the task was written for could not invoke it: a cross-repo caller cannot resolve an absolute devrc path. That is the rationale `nix/home.nix` already states for `claim-work` and `cairn`. **Fixed** — `.local/bin/peer-host`, `mkOutOfStoreSymlink` (required, not stylistic: `peer-host` reaches `lib/` through `Path(__file__).resolve().parent`, so a store copy dies on import). Verified by invoking it through a symlink from an unrelated repo's cwd.
2. **`--strict` deleted.** Unattributed, and `--json` already publishes `hosts_unsearched`.
3. **`--local-host` deleted.** Unattributed, and the only hand-hold for the mislabelling in the limitations below.
4. **The session-manager seam guard was the wrong instrument.** Folded the constant in instead; see above.
5. **Selector kinds KEPT, attribution recorded.** The auditor put `pid` / `session_id` / `tmux` up for deletion as beyond "name or ref". They are attributed: the task's dispatch brief asked for them by name. The source now says so, so the next reader does not re-open it.

Round 0 also flagged the duplicate registry reader, which is **filed as #1602** rather than folded in — merging it with `session-resolve`'s reader is a design question, and `peer-host`'s copy is shipped over SSH so it must stay self-contained.

### The CI tier this branch had not read was RED, and it was a real defect

I flagged the nix-build tier as unverified; CI read it and found one failure — **in my own test**. `test_the_module_is_executable_and_has_a_python_shebang` pinned the env-form interpreter line exactly, and `flake.nix` runs `patchShebangs src/scripts` over the copied tree, so that string exists in only one of the two tiers. The test passed on the dev host and *could not* pass in the sandbox: a test whose environment decides its answer. It checks by **shape** now (`#!` + a python3 interpreter). Fixing it then tripped the repo-wide `test_runtime_shebangs.py` scan twice, both resolved the way the repo prescribes — the prose stopped carrying the literal, and the one remaining assertion is allowlisted as shape (b) with a reason.

**Ledger** — `round 0 · payload lines changed THIS round: 321 insertions / 85 deletions across 5 payload files (nix/home.nix, scripts/peer-host, scripts/session-manager, scripts/session-analysis/espanso-usage.py, scripts/lib/host_label.py) + 2 scaffolding files (the two test files).` The attribution gate does not fire: this round changed payload substantially.

## Test matrix

**85 tests** in `scripts/tests/test_peer_host.py`.

### Red at base

Honestly: **vacuous.** At `origin/main` the whole module fails to import, because `scripts/peer-host` does not exist there:

```
E   FileNotFoundError: .../basetree/scripts/peer-host
1 error in 0.30s
```

That is a red, but it is not evidence about any individual assertion, and it is not claimed as such. The tests pinning behaviour that was **measured wrong during this branch's own development** are named in `RED_AT_BASE_FOR_A_REAL_DEFECT` and in the sections above — the argparse exit-code collision, the spliced `from __future__` line, and the two-tier shebang pin.

### Mutation sweep — the real evidence

Because the base red is vacuous, the coverage claim rests on mutation sweeps — **one per round, reported per round rather than as a running total.** An earlier version of this section carried a hand-summed "47 distinct mutants over 63 runs" and went stale the moment the next round ran; a total kept beside what it counts will drift, so the per-round figures below are each a number a sweep printed.

Every mutant is restored by `cp` from a pristine copy with a sha256 check, under `PYTHONDONTWRITEBYTECODE=1` — a same-length edit inside one second of the last import is invisible to CPython's mtime-in-seconds bytecode cache and would score SURVIVED without ever executing. Every sweep carries both controls (`CONTROL-none` must be green, `CONTROL-broken` must go red **via its own named test**).

A mutant counts as **KILLED only if the test the matrix names for it actually fails**; "something went red" is green for the wrong reason. `MUTATION_MATRIX` records every mutant and its killing test, and `test_every_mutation_matrix_entry_names_a_real_test` keeps the matrix from becoming a claim nobody checks.

| sweep | mutants | result |
|---|---|---|
| round 1 | 38 | 35 killed · **3 problems**, all defects in the *tests* |
| round 1, re-check | 4 | all killed |
| round-0 fixes (delta) | 12 | 11 killed + 1 `PATCH-FAILED` (bad pattern), re-run and killed |
| round-1 fixes (delta) | 11 | 10 killed + 1 `KILLED-BY-WRONG-TEST` → the guard was **deleted** as unreachable |
| round-2 fixes (delta) | 11 | all killed |
| round-3 fixes (delta) | 12 | all killed |

Every sweep's controls behaved: `CONTROL-none` green, `CONTROL-broken` red via its own named test.

**The sweeps are the reason this PR is not just a green suite.** Four times they refuted a coverage claim the suite could not:

1. **Round 1 — the negative control did not go red.** Renaming `STATUS_RESOLVED`'s value broke nothing, because every test compared `r["status"]` against `ph.STATUS_RESOLVED` — against itself. `--json` publishes those strings, so they are a contract.
2. **Round 1 — `raw.get("tmux")` → `raw["tmux"]` SURVIVED.** The only tmux-less fixture arrived through the *stubbed SSH leg* as an already-shaped dict, so the reader never parsed it. The docstring claimed coverage the test did not provide.
3. **Round 2 — `import tomllib` SURVIVED** the guard round 1 had just added to prevent exactly that. It asked "is this in `sys.stdlib_module_names`?" against the **local** interpreter; `tomllib` is stdlib on 3.11+ and absent on 3.9, the version the constraint names. Replaced with an exact-equality import ledger.
4. **Round 3 — the ledger's own `==` was unprotected.** Mutating it to `<=` or `>=` SURVIVED, while a `MUTATION_MATRIX` row claimed the mutant was killed: the negative control re-performed the comparison on synthetic strings instead of invoking the guard, so it could not see the guard's operator. The control now calls the real guard and asserts it goes red on a GROWN *and* a SHRUNK set.

Round 3 also measured three more survivors, all now killed: an import added to `remote_program()`'s **trailer** (a third import-bearing contributor the ledger did not cover), and a `home.file.".config/…"` entry wedged between the cairn entries (the adjacency regex only indexed `.local/bin/`).

### Green at HEAD, on the merged tree

* `test_peer_host.py` — **85 passed**.
* Merged `origin/main` at each round (5, then 9, then 3 commits) and re-ran the affected suites on each merged tree — **1002 → 1024 → 1098 passed**. Where main touched a file this PR also edits (`nix/home.nix`, `test_runtime_shebangs.py`), I read the merged region rather than trusting the clean merge.
* `scripts/run-tests.sh --check-floors` — **PASS**.

⚠ **One run of those six suites reported 7 failures and it was MY OWN CONCURRENCY, not a defect.** It overlapped the delta mutation sweep by 58 s (sweep 13:57:14–14:00:24, suite 13:59:26–14:01:45), and the sweep rewrites those exact files in place. The rival mechanism — real cross-suite pollution — predicts reproduction; it did not reproduce: the same six files alone gave 1002 passed at that commit, and every pairwise combination is clean (84 / 85 / 84 / 151 / 898). Recorded rather than quietly re-run.

**`TARGET_FLOORS` deliberately untouched.** `scripts/tests` collects **14 675** at head (re-derived; it was 14 426 when this line was first written) against a floor of 13 026 — 1 649 of slack, inside the gate's own drift allowance of `max(60, floor/4)` = 3 256, and `test_ordinary_growth_needs_no_hand_edit` says growth needs no hand edit. Bumping it would require re-deriving the global sum and would conflict with every concurrent branch, for no gate signal.

## Live verification (re-run against the FIXED code, sha256 `9a2641d0…`)

🔴 An audit fix RESETS the verification gate. The first round's transcripts were measured against a build that predates both the `proc_dir` seam and the round-0 fixes, so **everything below was re-run against the committed code**, not carried over.

**Criteria 1 + 2 — one known-workbench and one known-laptop session** (ground truth read from the registries directly, not from the tool):

```
$ peer-host cairn-b9            -> workbench   exit=0
$ peer-host auditloop-1d        -> laptop      exit=0
$ peer-host 3587236             -> workbench   exit=0     (pid)
$ peer-host e081c4              -> workbench   exit=0     (session-id prefix)
```

A single-host check cannot distinguish a working lookup from one hardcoded to the local machine; `auditloop-1d` exists only on the laptop and the local host is `workbench`.

**Criterion 3 — no-match and ambiguous, both live:**

```
$ peer-host zzz-not-a-peer
peer-host: no live peer matches 'zzz-not-a-peer' on workbench or laptop
           (searched: workbench=51 records, laptop=9 records)
  exit=3

$ peer-host datapacket-talos-d3
peer-host: 'datapacket-talos-d3' matches 2 live peers — refusing to guess:
  workbench datapacket-talos-d3  a65d47b0  pid=1207483  tmux=scratch12:@23.%23
  workbench datapacket-talos-d3  83cad032  pid=3424794  tmux=scratch15:@70.%70
  exit=2

$ peer-host --bogus x        -> exit=64   (usage, NOT 2)
$ peer-host --strict x       -> exit=64   (deleted flag is really gone)
$ peer-host --local-host … x -> exit=64   (deleted flag is really gone)
```

Both ambiguity cases were **live, not constructed** — and they were *different* duplicate names across runs hours apart, so this is a recurring condition rather than a one-off.

**Criterion 4 — a session started 37 seconds earlier, no file edited:**

Started a fresh `claude` on the **laptop**, in a dedicated tmux server (`tmux -L ph-final` — never the operator's default socket), in a directory the laptop already trusts:

```
newest laptop record: name=civitai-gpu-fleet-5c  pid=1510612  sid=fba158  age=37s

$ peer-host civitai-gpu-fleet-5c  -> laptop
$ peer-host 1510612               -> laptop
$ peer-host fba158                -> laptop
```

The working tree was unchanged between starting that session and asking for it — a diffstat over the branch showed only the fix-round edits, none of which is a name-to-host map.

…and the same lookup **through the PATH symlink, from an unrelated repo's cwd** → `laptop`, which is the round-0 reachability finding closed end to end.

Torn down afterwards: server killed, socket files removed, laptop registry back to its baseline 9, operator's 19 sessions untouched.

**Safety controls — the failure modes criterion 3 forbids, driven live against the fixed code:**

```
CONTROL B — laptop pointed at a blackhole address
  status=incomplete  host=None  exit=4   laptop leg: unreachable
  => a laptop-only name did NOT come back as "workbench"

CONTROL C — laptop down, workbench name still resolves
  status=resolved  host=workbench  exit=0
  stderr: WARNING — answered from workbench, but laptop could not be searched;
          a same-named peer there would be invisible

CONTROL A2 — dial the LAPTOP under the label "workbench" (the host ANSWERS)
  workbench leg: label_mismatch | reported_label=laptop
  => refused, not recorded

NEW — the local leg reports NO self-label (honest asymmetry)
  local(workbench).reported_label = None
  remote(laptop).reported_label   = 'laptop'
```

Control A2 matters because the first attempt at it (`10.42.0.30` under the laptop label) hit `unreachable` first — self-SSH is refused — so the `label_mismatch` guard was **breakable but not proven reachable**. A2 reaches it with its own specific error.

## Not verified / limitations

* **The bracketed `[ref]` selector.** Not implemented, not derivable — see above. `test_no_ref_selector_is_implemented` is labelled honestly as an **invariant guard**, not regression coverage: it cannot distinguish a correct ref derivation from a wrong one, because no derivation is known. What it pins is the deliberate absence.
* 🔴 **`peer-host` has only ever run *from* the workbench, and there is a latent wrong-answer path when run from the laptop.** The local/remote split comes from `host_label.local_host_label()`, which **defaults to `workbench`** when neither `ACTIVITY_HOST` nor the collector env file supplies a value. On a laptop in that state the local leg would read the laptop's registry and label it `workbench`, while the `laptop` leg would be a self-connection that fails — so laptop peers would come back as `workbench`: a **wrong answer, not a refusal**. The remote leg's self-identification cannot catch this, because the mislabelled leg is the local one.

  Measured: the hazard is **latent, not live** — both hosts positively set the label today (`ACTIVITY_HOST=laptop` in the laptop's `~/.config/activity-collector/env`, `ACTIVITY_HOST=workbench` on the workbench), and the laptop's own `host_label` demonstrably returns `laptop` (that is what control A2 measured). Deliberately **not** fixed here: the defaulting behaviour is `host_label`'s, shared by `transcript-push`, `tmux-reply-agent`, `session-manager` and `espanso-usage`, so a guard belongs there — cross-checking the label against the machine's own Nebula address, an independent signal — not as a private copy in this script. **Filed as #1601** with a closing condition. The local leg no longer *pretends* to be checked: its `reported_label` is `None`, and `test_the_local_leg_reports_NO_self_label` pins that a future edit cannot tidy the two columns into looking uniform.
* **Two independently-written readers of `~/.claude/sessions/*.json` now exist** (`session-resolve`'s and this one), and they already disagree on stale records, tmux-less records and error accounting. **Filed as #1602.**
* **The peer SSH address is still spelled in several places, and this PR does not claim otherwise.** An earlier version of this body said "the peer SSH address table has exactly one spelling" — retracted; it was false. What is enforced is scoped and stated in `host_label.PEER_SSH`'s comment: no `user@addr` string constant for a peer, in any non-test file under `scripts/`, outside `host_label.py`. Out of that scope and deliberately **not** enumerated anywhere: runtime-composed targets (`host-role.sh`), bare IPs (`browser-bridge/server.py`), and everything under `nix/` (espanso snippets, a systemd `ExecStart`, `nix/system/` shell). Two drafts of that comment tried to list every spelling — one caught at 3 of 5, the next at 5 of 11+ — so the third states the scope and tells the reader to grep.
* **A stale registry record** (pid gone) is dropped and counted. Measured 0 stale of 49 locally and 0 of 9 on the laptop, so the harness prunes; the filter is a guard, not a hot path, and pid reuse can only make it too permissive, never too strict.
* **Cross-host name collisions** were 0 at every measurement. If one occurred it surfaces as `ambiguous` (2), not a wrong answer.
* **The nix-sandbox tier is read via CI, not locally.** I have not run `nix build .#checks.x86_64-linux.pytests` on this machine; the local `run-tests.sh` attempt was killed by my own timeout (exit 143) and is **inconclusive — not a pass**. What is measured is the Tekton `devrc-pytests` check, which runs that tier: it caught the shebang defect on the previous push, and the fix is pushed for it to re-run. Treat the CI result on this PR as the authoritative tier signal, not anything I ran locally.

https://claude.ai/code/session_01UESXKDWC4HXPoaYYUGaG48


ZacxDev:
Round 1 (nine correctness axes) returned 4 should-fix findings and 5 nits. All were verified against the code before acting; all are fixed in `5ca02ba0`, merged with `origin/main` at `4bda23c9`.

Round-1 delta mutation sweep: 11 mutants, all killed by the test the matrix names for them; `CONTROL-none` green and `CONTROL-broken` red via its own named test; restores sha256-verified, run under `PYTHONDONTWRITEBYTECODE=1`. 1024 passed across the seven affected suites; 82 in `test_peer_host.py`.

One guard was **deleted rather than kept**: the relative-import branch of `test_host_label_stays_SHIPPABLE_over_the_wire`. The sweep showed it can never execute — a `from . import x` in `host_label.py` breaks the flat `import host_label` at collection, so the mutant kills the whole file and the assertion is never reached. The hazard is real but is caught earlier and louder by every importer at once, so a branch for it would be dead code claiming coverage it cannot provide.

```audit-claims round=1 audited=c41632ba..4bda23c9
1. A non-ASCII digit selector no longer crashes out of match_kinds: `sel.isascii() and sel.isdigit()` replaces `sel.isdigit()`, so "²" is a non-match (exit 3) instead of a ValueError traceback (exit 1), and "١٢٣" no longer resolves as pid 123.
2. An empty or unrecognised host scope now raises ValueError instead of returning UNMATCHED (exit 3) from a search that dialled no host.
3. The peer-address ledger in host_label.py no longer claims to list "every place these addresses are spelled"; it is narrowed to what the guard enforces (a `user@addr` string constant under scripts/) and NAMES the two remaining independent spellings, scripts/lib/host-role.sh and scripts/browser-bridge/server.py.
4. The same over-claim was corrected in the PR description, which had said "the peer SSH address table has exactly one spelling"; it now states 5 spellings reduced to 3.
5. test_the_remote_program_actually_runs_and_emits_the_protocol no longer asserts `stale_dropped == 2` against the machine's real /proc; it asserts `stale_dropped + len(records) == 2`, so it cannot pass or flake by accident of the live pid table.
6. The vacuous `assert proc` in that same test is removed.
7. read_host no longer escapes as a traceback when host_label.py cannot be read; it returns a leg ERROR, making its "never raises for a host problem" docstring true.
8. parse_leg_output's docstring now states that leading noise is tolerated but the tail must parse as pure JSON.
9. The nix/home.nix peer-host entry was moved below cairn-validate, so that block's "the THIRD member of the pair above" and "read all three lines together" resolve to the cairn trio again.
10. A new guard, test_host_label_stays_SHIPPABLE_over_the_wire, pins that host_label.py imports only stdlib, because it — not _READER_SOURCE — is the module spliced into the remote SSH program.
```

ZacxDev:
Round 2 (delta) confirmed all ten round-1 claims as fixed or partially fixed, and returned 4 findings. All verified against the code before acting; all fixed in `4d81ecfc`, merged with `origin/main` at `4a8a981e`.

Three of the four were this branch's own **prose** — now the dominant defect class in this PR, not an incidental one. The ledger comment overclaimed for the **third consecutive round** (3 of 5, then 5 of 11+), so the fix this time was to stop enumerating and state the guard's **scope** instead, which is checkable and stable. Extending the list is what kept regenerating the defect.

Round-2 delta mutation sweep: 11 mutants, all killed by the test the matrix names for them; both controls correct; restores sha256-verified under `PYTHONDONTWRITEBYTECODE=1`. Notably `host_label-imports-tomllib` — which **SURVIVED** the guard round 1 added — is now killed, and so are `reader-source-imports-tomllib` and `cairn-trio-re-wedged`. 1098 passed across the seven affected suites; 85 in `test_peer_host.py`.

```audit-claims round=2 audited=4bda23c9..4a8a981e
1. The peer-address ledger in host_label.py no longer enumerates where the addresses appear; it states the guard's SCOPE (no user@addr string constant in a non-test file under scripts/, outside host_label.py) and names runtime-composed targets, bare IPs and everything under nix/ as out of scope and deliberately not enumerated, with an instruction to grep.
2. The four file:line references that ledger carried are removed; nix/home.nix already states that a line number is a claim that rots silently, and one of the four had gone stale inside this PR.
3. The PR description no longer contains the retracted sentence "the peer SSH address table has exactly one spelling", which had survived verbatim further down the body after round 1 claimed to retract it.
4. test_host_label_stays_SHIPPABLE_over_the_wire is replaced by test_both_spliced_sources_stay_SHIPPABLE_over_the_wire, which pins BOTH sources remote_program() splices (host_label.py and _READER_SOURCE) against an exact SHIPPABLE_IMPORTS ledger checked for equality, so `import tomllib` — which survived the previous sys.stdlib_module_names check — is now rejected.
5. That guard's unreachable `assert stdlib, "cannot judge"` is removed; `| {"__future__"}` made it always truthy, and on a <=3.11 interpreter the test would have failed claiming `os` does not exist on the remote.
6. That guard's docstring no longer says the stdlib constraint was "stated about the WRONG module"; remote_program() splices both sources, so the constraint was INCOMPLETE, not misplaced.
7. The MUTATION_MATRIX count and test count in the PR body are re-derived at head (46 -> 52 entries, 75 -> 85 tests), by AST and by pytest collection rather than by eye.
8. New guard test_the_cairn_home_nix_entries_stay_ADJACENT pins that nothing is inserted between the three cairn .local/bin entries, which round 1's positional fix left unguarded; test_cairn_split.py pins that comment's text, nothing pinned the adjacency the text depends on.
```

ZacxDev:
Round 3 (delta) confirmed round 2's eight claims and measured **four mutants that SURVIVED a fully green suite**. All four are now killed by their named test. Fixed in `592baef0`.

Every one of the four was a guard *this branch added in an earlier round to close a coverage gap* — the same defect class the ladder has been chasing, one level up: not "the code is wrong" but "the guard does not check what it says".

Round-3 delta sweep: 12 mutants, all killed by their named test, both controls correct — including all four survivors. 1029 passed across the six affected suites; 86 in `test_peer_host.py`. Live re-verified against the committed code: workbench → `workbench` (0), laptop → `laptop` (0), no-match → 3, ambiguous → 2.

## 🔴 The ladder stops here, and this is the stated reason

Round 3's fixes changed **59 payload lines, all of them comments** — the executable payload has now gone two rounds unchanged, and 6 of round 3's 9 findings were guard/prose. The attribution gate is one round from firing. Round 3's own recommendation was to apply these fixes and verify **the specific deltas directly rather than run a fourth full round**, which is what was done: each of the four surviving mutants was re-run individually and is now killed by the test the matrix names for it.

Remaining round-3 items deliberately **not** fixed, so the next reader knows they are open rather than absent: F9 (a "3 of 4" wording nit in `host_label.py`, fixed) — nothing else is outstanding. The two design follow-ups are filed as #1601 and #1602.

```audit-claims round=3 audited=4a8a981e..592baef0
1. test_that_shippable_ledger_can_actually_fire now invokes the real parametrized guard instead of re-performing its comparison on synthetic strings, so the guard's own `==` operator is covered; mutating it to `<=` or `>=` (both of which SURVIVED before) is now killed.
2. SHIPPABLE_IMPORTS gained a "remote_program()" entry pinning the import union over the ASSEMBLED remote program, closing the unpinned third contributor — remote_program()'s trailer, which emits `import sys as _s`; adding `import tomllib as _t` there SURVIVED before and is now killed.
3. test_the_cairn_home_nix_entries_stay_ADJACENT indexes every `home.file."..."` key rather than only `.local/bin/` ones, so a `home.file.".config/..."` wedged between the cairn entries — which SURVIVED before — is now killed.
4. That guard's docstring no longer claims test_cairn_split.py pins the cairn-validate comment's text; test_cairn_split anchors on the .local/bin/cairn entry, the positional phrases are pinned by nothing, and this is the only pin and guards position rather than wording.
5. The SCOPE sentence in host_label.py and in the guard's docstring is measured against _scan_for_address_literals rather than inferred: it catches a string constant EQUAL to a peer's user@addr and misses one that merely CONTAINS it, an f-string, a bare address in a .sh, and anything .md; the predicate is "not under a tests/ directory", not "non-test file".
6. "all were literals" over the four DERIVE targets is corrected to three of four; scripts/peer-host is new and never carried one.
7. The PR body no longer asserts an unscoped census; the "from 5 to 3" total is dropped and the claim is scoped to scripts/, non-test, string constants, with nothing claimed or enumerated outside that scope.
8. The PR body's hand-summed sweep total ("47 distinct mutants over 63 runs") is replaced by per-round figures, each a number a sweep printed, and the scripts/tests collect count is re-derived at head as 14675.
```

PR #1613, opened 2026-09-12
docs(handoff): the triage bot's first live sweep produced a FALSE INHERITED — arming is blocked

ZacxDev:
Updates `claudedocs/handoff-gate-speed-and-ci-signal.md`.

- **rank 7 — `ZacxDev/homelab-infra#799` MERGED** `0b14768a`, verified by CONTENT on `trunk` (a squash merge never makes the head an ancestor). Test-only claim re-verified independently before merging: 2 files, both `scripts/tests/*`, zero under `clusters/`/`triggers/`/`apps/`.
- **rank 11 — `devrc#1603`'s red was REAL.** The PR had added `test_the_runner_is_tracked_and_executable`, which calls `git ls-files` — unpassable in the `nix build` sandbox tier, which has no `.git`. It reported `not tracked by git: []` against a file tracked at mode 100755. Merged, that is a permanently-red gate on `main`: the exact failure `#1603` exists to prevent. Reproduced via `git archive` into a `.git`-free tree, fixed at `8a88f255`, verified in both tiers (29 passed each), tier guard mutation-killed by its own error string.
- **rank 10 — `devrc#1600` HELD, not merged.** Green on its own branch but 24 commits behind `main`; the merged-tree run over the 72 test files reading `nix/home.nix` did not finish. Recorded as an open investigation with the exact repro, because its verdict is UNREAD.

https://claude.ai/code/session_01298sLVSsAeZJpGNJTftdWv


ZacxDev:
Second commit (`621074c4`) — the state moved materially after the first:

- **`devrc#1600` MERGED `f99d3c1b`.** Its merged-tree run came back **8184 passed / 0 failed** over the 72 test files that read `nix/home.nix` (33 min). The merged-tree claim was re-checked *at merge time* and held: `origin/main` was `7e000e6b` both when the tree was built and when the merge ran, and the tested tree's second parent was the PR's live head `18701e53`.
- 🔴 **But it runs on NO host.** `ship.sh` exited **rc 7**: the laptop fast-forwarded and verified, the **workbench was SKIPPED** on a locally-changed `claude/skills/clawgate/SKILL.md`. The timer is `serverMode`-gated to the workbench, so the laptop's copy is `UnitFileState=linked`, 0 timers, `journalctl` `-- No entries --`. Merged ≠ deployed ≠ running; this broke at deploy.
- 🔴 **The blocker was NOT cleared, deliberately.** The file is live WIP — working-copy sha1 matches none of the last 8 commits of that path, and its mtime was **2 minutes old** when read. `ship.sh`'s printed remedy (`git checkout origin/main -- <file>`) would have destroyed another session's in-progress security retractions to the clawgate skill. Handed over instead.

Rank 10 stays OPEN with a deploy, not a code change, as its remaining work.

ZacxDev:
Third commit (`5ee83fcf`) — rank 10 is now COMPLETE, and it immediately produced the most important finding of the session.

**Rank 10 done, all three claims made separately:** `#1600` merged `f99d3c1b`; `ship.sh` rc **0** converged both hosts to `b55720e8` (the earlier rc 7 cleared once another session finished with `clawgate/SKILL.md` — it was never mine to clear); consumer **verified running** on the workbench — timer `enabled`/`active` on a 2h interval, one clean run, `Result=success` `ExecMainStatus=10`, and the deployed `ExecStart` still carries `--comment-mode dry-run` read off the unit rather than the source.

🔴 **Its first live sweep produced a CONFIRMED FALSE `INHERITED`** — which `#1600`'s own unit comment calls disqualifying for arming (criterion (i): *"ONE false INHERITED is disqualifying: this bot would be telling a human to ignore a real red"*).

It ruled `devrc#1603` INHERITED, citing that `test_runtime_shebangs.py` is byte-identical at head and merge-base while `main` moved it in `cfdb3899`. **That commit's entire change to that file is an 18-line allowlist row for `test_nvim_octo.py`, an unrelated file.** A rebase would not have cured anything — the red was `#1603`'s own diff, and replacing the offending writes turns the guard green with `main` held constant.

The mechanism is structural, not bad luck: the heuristic *"the failing test file is unchanged in my branch"* is sound for a test that exercises code it names, and **backwards for a repo-wide census/scanner guard**, where the offender lives in a different file and an unchanged guard file is the normal state of a genuine breakage. This repo is dense with that class — it is the same class rank 9 records as unaddressed.

Filed as **rank 13**; the bot is inert meanwhile, so this is a decision, not an incident.

Also: `#1603` had a **second** real red, caused by the fix for the first — the new controls wrote runtime stubs with an env-resolved shebang, absent in the sandbox for the same reason `.git` is. Fixed at `323c6b6b` via `mockbin.write_exec`; **38 passed in both tiers**.

PR #1619, opened 2026-09-12
fix(mention-open): size the review TUI per host in percent of the OUTPUT rect, and delete the cell hint

ZacxDev:
> **Revised three times, after three audits of this PR's own earlier heads.**
>
> - **Round 0** audited `d1f80d4a` (both hosts at `90 ppt 90 ppt`, a lowered `140x40`
>   cell hint).
> - **Round 1** audited `451ead89` and found the central arithmetic was **against the
>   wrong rect** — `ppt` resolves against the **OUTPUT** rect, not the workspace — so
>   every height *derived from a `ppt` percentage* was short by the status bar. It
>   concluded the workbench's `78 ppt` was a **visible** size change and moved it to
>   `77`. **Round 2 retracted that conclusion** (see below); the rect correction itself
>   stands.
> - **Round 2** audited `13a2a0ce` and found that round 1, while documenting a term it
>   had just discovered it was ignoring, **asserted the wrong mechanism for it in five
>   places, two of them payload files**: the float's decoration is `pixel 2` (BS_PIXEL),
>   not i3's default `normal 2` (BS_NORMAL) — there is no titlebar. That **voids the
>   stated rationale for round 1's only code change**: `77` and `78` both render
>   `199x50`. `77` is kept because it is already committed, and for no better reason —
>   the PR now says so rather than supplying a fresh justification. It also retires this
>   PR's "the workbench renders what it renders now" promise, which the arithmetic does
>   not keep in the width axis.
>
> See
> [the round-0 findings](https://github.com/innovation-upstream/devrc/pull/1619#issuecomment-5650169649)
> and the round-1 / round-2 findings comments below. Everything here describes the PR
> **as it now stands**.

## The defect

The operator: **"the new tui opens far too big and overflows the screen, its unusable."**

`scripts/mention-open.py` had `REVIEW_COLUMNS = 200` / `REVIEW_LINES = 50`, passed to alacritty as `-o window.dimensions.{columns,lines}`. That is a size in **character cells**, while the constraint it must satisfy is a screen in **pixels**. A cell is not a length — its pixel size is a function of font size and DPI — so one constant cannot fit both hosts.

**And the complaint was the LAPTOP's.** A TUI opening implies a GitHub reference, which implies `record_pick` ran; the workbench's `~/.config/mention-open/picks.jsonl` **does not exist**, while the laptop's does. The review TUI has never opened on the workbench. (⚠ The row count is deliberately not quoted: it grows with use, nothing asserts on it, and the load-bearing claim is presence vs absence. An earlier revision said "39 rows"; it was 40 a day later.)

## 🔴 `ppt` is a percentage of the OUTPUT rect, not the workspace

This is round 1's finding and it moved every number below. Established from **i3 4.25.1 source** (the version running on both hosts), two independent artifacts:

- `src/commands.c`, `cmd_resize_set()`, floating branch:
  ```c
  const Con *output = con_get_output(floating_con);
  … cwidth  = output->rect.width  * ((double)cwidth  / 100.0);
  … cheight = output->rect.height * ((double)cheight / 100.0);
  ```
- `testcases/t/252-floating-size.t`: with `fake-outputs 1333x999`, `resize set 33 ppt 20 ppt` is asserted to land on `int(0.33*1333) x int(0.2*999)` — the **output**. i3's test outputs carry no bar, so workspace == output there, which is why this can never surface in i3's own suite.

⚠ And **`move position center` in the same chain centres against the WORKSPACE rect** — `cmd_move_window_to_center()` → `floating_center(…, con_get_workspace(…)->rect)`; only `move absolute position center` uses the root rect. So one `for_window` chain sizes against one rect and places against another. Harmless at these percentages; it would not be near `100 ppt` of the height.

## Measured

Read-only, both hosts, 2026-09-12, i3 4.25.1. Rects from `i3-msg -t get_outputs` and `-t get_workspaces`. Cell size taken off each host's **running alacritty pty** with `TIOCGWINSZ`, which reports the grid in rows/cols **and** its size in pixels — so the cell size is a division, not a DPI estimate; the pty has to be a *direct child of alacritty*, since a tmux pane's winsize is synthesized by tmux and says nothing about a font.

| | workbench | laptop |
|---|---|---|
| **OUTPUT rect** (what `ppt` multiplies) | **`3440x1440`** | **`2256x1504`** |
| workspace rect (what it must fit, and what centring uses) | `3440x1413` | `2256x1480` |
| status bar | 27 px | 24 px |
| **measured cell** | **`11.0 x 22.0` px** | **`19.0 x 37.0` px** |
| largest grid observed | `312 x 63` | `118 x 39` |
| old `200x50` cells | `2200x1100` — fits | `3800x1850` = 168% x 125% of the workspace — **overflows** |
| `90 ppt 90 ppt` of the output | `3096x1296` — **+41% on the width the operator called unusable** | `2030x1353` — fits |
| `140x40` cells | `1540x880` — fits | `2660x1480` = **118% wide — still overflows** |

**Width is unaffected by the rect correction** — the bar is a full-width strip, so output width == workspace width. **Only the height was ever wrong**, by 27 px / 24 px.

The laptop's cell is **1.73x wider, 1.68x taller, 2.90x the area**. An earlier version of this PR said *"roughly twice in each axis — roughly four times the area"* in four places; that was never measured and **over-stated** the gap, which is how one percentage came to look like it could serve both hosts.

## The fix, part 1 — per-host percentages

`nix/i3/config.nix` is already `{ isLaptop ? false }:` and the file has per-host precedent (laptop brightness keys; workbench output pinning), so the numbers go through that mechanism as `reviewSizePpt` — only the operands vary, so the rule's structure cannot diverge between hosts by accident:

```
for_window [class="float" instance="mention-review"] floating enable, resize set ${reviewSizePpt}, move position center
```

| host | value | i3 RECT (of the OUTPUT) | fits workspace | CLIENT rect | rendered cells |
|---|---|---|---|---|---|
| **laptop** | `90 ppt 90 ppt` | `2030x1353` px | `2256x1480` ✅ | `2026x1349` | ≈`106x36` |
| **workbench** | `64 ppt 77 ppt` | `2201x1108` px | `3440x1413` ✅ | `2197x1104` | ≈**`199x50`** |

⚠ **The rect and the client rect are different numbers and the distinction is load-bearing** — a floating con's `rect` includes its border, which is `pixel 2` here, so the client area is 4 px smaller in **both** axes (derivation below). The bare rect is `200x50` cells; what renders is `199x50`. Earlier revisions of this table quoted the bare-rect grid as the rendered one, and one of them applied the correction to the width axis only.

🔴 **`77` AND `78` ARE INDISTINGUISHABLE, AND `77` IS KEPT ONLY BECAUSE IT IS WHAT IS ALREADY COMMITTED — round 2's correction.** Round 1 argued `78 ppt` of the output's `1440` is ≈`1123` px = **51 rows**, one row more than today, i.e. a visible change. **`51` is a row count of the BARE RECT.** The decoration the same commit documented absorbs it:

| `ppt` | rect | client (BS_PIXEL bw 2) | cells |
|---|---|---|---|
| `64 ppt 77 ppt` | `2201 x 1108` | `2197 x 1104` | **`199 x 50`** |
| `64 ppt 78 ppt` | `2201 x 1123` | `2197 x 1119` | **`199 x 50`** |

So that rationale was **void**. 🔴 **It has not been replaced with a better one.** No purpose for `77` over `78` could be found; `77` is retained because it is the value already committed and deployed on this branch, not because it renders differently. "I could not find a purpose" is the finding — a purpose produced under pressure to supply one would be a hypothesis, and this ladder has already watched a claim go through several drafts, each retracted by the next round.

🔴 **AND THE NO-CHANGE PROMISE IS RETIRED, NOT RE-JUSTIFIED.** `199x50` is not today's `200x50`, so *"the workbench must come out of this change rendering what it renders now"* **is not met in the width axis either — for `77` or `78`**. The height axis matches (50 rows); the width is one column short. Round 1 applied the border correction to the **width axis only** ("the true grid is likely ~`199x50`") — the claims-a-relationship-inspects-one-side shape, in prose — and it is now applied to both axes everywhere it appears.

**The stakes, honestly:** the workbench's `~/.config/mention-open/picks.jsonl` does not exist, so this window has never opened on that host and **no rendered grid has ever been observed there**; the laptop — the actual defect — fits comfortably either way. Width stays `64 ppt` (`2201` px of `3440`).

⚠ **The cell figures above are BARE-RECT grids; the rendered grid is 4 px smaller in EACH axis.** One term is now modelled and one still is not.

🔴 **(a) DECORATION — and round 1's claim here was WRONG.** It said `nix/i3/config.nix` "sets `default_border pixel 2` but never `default_floating_border`, so floats take i3's default `normal 2` … a titlebar plus borders". **That does not hold for this window.** Verified against the realised i3 4.25.1 source the running binary was built from (`/proc/<pid>/exe` → its derivation's `src`):

- `config.default_floating_border` is applied **only** inside `floating_enable()` under `if (automatic)` — `src/floating.c:353-354`.
- `for_window … floating enable` is a **command**, not the automatic path: `cmd_floating()` calls `floating_enable(con, false)` (`src/commands.c:1157`), and `run_assignments()` runs at `src/manage.c:588` (map time; `:746` on the remanage path) — **after** the `want_floating` decision at `src/manage.c:462-546`.
- `want_floating` is set only by window-type atoms / `_NET_WM_STATE_MODAL` / sticky / `transient_for` / a fixed `min==max` size hint. An alacritty toplevel matches none.
- So the container keeps what `con_new_skeleton()` gave it (`src/con.c:44`, `config.default_border`) = **BS_PIXEL with `logical_px(2)`**, and `con_border_style_rect_without_title()`'s non-BS_NORMAL branch is `{bw, bw, -2bw, -2bw}` (`src/con.c:1846-1849`) — 2 px on all four sides, so the client area is the rect **minus 4 px in BOTH axes**. **No titlebar.**

**Live, read-only** (`i3-msg -t get_tree` / `-t get_config`, i3 4.25.1): the running alacritty windows report `border=pixel`, `current_border_width=2`, and the config carries no `default_floating_border`, no `new_float` and no `for_window … border`. ⚠ Those live windows sit in a **tabbed** parent, where `con_border_style()` (`src/con.c:1942`) overrides a non-BS_NORMAL style to BS_NORMAL for a >1-child tabbed container — which is why their client *height* is rect−2, not rect−4. A float's parent is a `CT_FLOATING_CON` with `layout = L_SPLITH` (`src/floating.c:291`), so no override applies and −4/−4 is the float's case. The live read is evidence for the **border style** and the **cell size**; the float's own inset comes off the C. The model is validated against the live window: `floor(3436/11)=312`, `floor(1391/22)=63`, matching the observed `312x63` grid.

**(b) SIZE-INCREMENT SNAPPING — still not modelled.** `floating_resize` (`src/floating.c`) upscales the decorated rect to a multiple of the window's increments. Measuring it needs a window opened or resized on the operator's live desk.

No pixel figure in this PR is quoted to a tenth, because the model cannot support that precision.

- **`instance=` and NOT the shared `class="float"` rule.** That general half is shared by every float terminal in the system — a dozen bar-click detail windows, `media-menu`, `airvpn-menu` — plus the fzf picker. Sizing there would move windows nobody asked to move. The shared rule is untouched.
- **`floating enable` is repeated.** `resize set` on a still-TILED window is a silent no-op, so the rule must not depend on another rule being reached first.

## The fix, part 2 — the cell hint is DELETED, not lowered

`REVIEW_COLUMNS`, `REVIEW_LINES` and **both** `-o window.dimensions.{columns,lines}` flags are gone from the review spawn. **i3 is the sole authority on this window's geometry.**

`140x40` — the lowered hint the first version of this PR shipped — is still `2660x1480` px on the laptop, **118% of its width**. A hint that can only ever be right on one host is a second bug, not a fallback.

Without a hint, alacritty maps at its own default (~`80x24` = `880x528` px on the workbench, `1520x888` on the laptop — inside both workspaces) and i3 resizes **up**. So the pre-resize flash goes **small → right** instead of **oversized → right**. If the i3 rule is ever absent you get a **small** window, not a broken one.

`PICKER_COLUMNS`/`PICKER_LINES` and the picker's spawn are **untouched**: different window, sized in cells deliberately, and its i3 rule carries no `resize`.

## 🔴 DEPLOY ORDER — `i3-msg reload` is a REQUIRED step, not a verification nicety

```
merge  ->  scripts/ship.sh  ->  i3-msg reload   (on EACH host)
```

Round 1's finding 3: the two halves of this change deploy on **different schedules**.

- `scripts/mention-open.py` is exec'd **straight out of the working tree** (`nix/programs/alacritty/default.nix` runs `~/workspace/devrc/scripts/mention-open.py`), so a plain `git pull` makes the hint **deletion** live with **no switch**.
- `nix/i3/config.nix` becomes `xdg.configFile."i3/config"` (`nix/graphical.nix`), so it needs a `home-manager switch` **and then an explicit `i3-msg reload`** — i3 does not re-read its config when the file changes, and **nothing in `nix/graphical.nix` or any activation script reloads it** (verified: the workbench's running config has no `mention-review` rule).

**So between `ship.sh` and the reload, neither half sizes the TUI**: the script passes no dimensions, the running i3 has no rule, and the review window opens at alacritty's default **~`80x24` on BOTH hosts**. That is small but usable (not overflowing) — and on the workbench it *is* a large change from today's `200x50`, persisting until somebody reloads. ⚠ That window is not the only change to the workbench: the **resized** size is `199x50` client cells, one column narrower than today. This PR does not keep the workbench's grid identical, and this paragraph used to say it "promises not to" change it.

An automatic reload is deliberately **not** wired into the activation script: that would reload i3 on every future `home-manager switch`, a change to shared graphical infrastructure well beyond this window. Recorded here as an **optional follow-up** for the operator to decide, not done.

## New finding, deliberately NOT fixed here

The **picker** is also marginally too wide on the laptop: `120x22` cells is `2280x814` px against a `2256` px width — ~101%, over by about 24 px of grid before any decoration is counted. Pre-existing, not introduced by this PR, and cosmetic. Narrowing `PICKER_COLUMNS` changes a layout the operator reads constantly (`format_row` wraps to `PICKER_COLUMNS - 2`), so it is a separate decision with its own before/after. Recorded in the source beside the constants.

## Coverage

Round 1's fixes **reset the verification gate**, so every assertion below was re-derived and the whole mutation sweep was re-run. Nothing here is carried over on an auditor's authority.

Guards read through the existing `for_window` engine in `scripts/tests/test_i3_picker_centering.py`, which asks *"what happens to a window with THESE properties"* rather than grepping config text. The review window's identity is derived from `REVIEW_CLASS`, never spelled twice. Every geometry guard is parametrized over both hosts.

**Regression guards — RED at `451ead89`** (this PR's own previous head; it is an amendment, so that is the meaningful base). Method: that ref's **production** files (`nix/i3/config.nix`, `scripts/mention-open.py`) with **these** tests.

| guard | at `451ead89` | what it pins |
|---|---|---|
| `test_the_WORKBENCH_keeps_rendering_the_size_it_ALREADY_renders` | 🔴 **RED** | `78 ppt` of the OUTPUT's 1440 maps to 200x**51** cells **of the bare rect**, not 200x50. The red is real and the arithmetic holds. ⚠ **Read the name narrowly** — round 2 found it over-states what is established: it pins a BARE-RECT derivation against a CLIENT-area cell count, so a green here is *not* "the workbench is unchanged" (it is one column narrower). Its docstring now says so, at length. |
| `test_the_review_windows_percentages_are_THIS_HOSTS[workbench]` | 🔴 **RED** | the workbench must receive exactly `(64, 77)`. |

Full matrix, that ref's production files against these tests, centering file unfiltered:

| base | result | RED |
|---|---|---|
| **`451ead89`** (previous head — the meaningful base) | `2 failed, 31 passed` | the two above |
| `d1f80d4a` (round-0 head, shared `90 ppt 90 ppt`) | `3 failed, 30 passed` | the two above **+ `test_the_two_HOSTS_review_percentages_DIFFER`** |
| `origin/main` (no review rule at all) | `14 failed, 19 passed` | every review-geometry guard on both hosts |

`test_mention_open.py` is not in that matrix: round 1 changed **no assertion** in it — only a docstring and **two** message strings — so its negative `window.dimensions` pin carries the same red-at-`main` evidence it had at round 0. (⚠ An earlier revision said "three message strings". `git diff --numstat 451ead89 13a2a0ce` on that file is **9 insertions / 3 deletions**: one docstring reword, six new docstring lines, two message strings. The load-bearing half — *changed no assertion* — is **true**.) It is in the HEAD-green run and in the mutation sweep (M16, M20).

**Invariant guards, labelled as such and not counted as regression coverage:**

- `test_the_MEASURED_ledger_is_INTERNALLY_COHERENT` — **new**, and the direct answer to round 1. Every other guard reads `_MEASURED` as ground truth, so a wrong ledger is invisible to all of them; this PR shipped green with the *workspace* height sitting in the slot the arithmetic used as the *output* height. It cannot measure a display, but it can assert the relationship a bar-on-an-output must produce: equal widths (an i3 bar is a full-width strip) and a height gap that is positive and bar-sized. A zero gap is exactly the substitution the audit found. The ledger's docstring now says plainly that it is a transcribed **input**, not a derived fact.
- `test_the_review_window_FITS_the_workspace_it_is_CENTRED_in` — **new**; the only assertion that checks the laptop's `90x90` against something other than the `_REVIEW_PPT` ledger. Pins that a size taken as a percentage of the *output* still fits the *smaller* workspace it is centred in.
- `test_a_DIFFERENT_float_instance_is_NOT_resized`, `test_the_PICKER_is_still_not_resized_by_anything` — unchanged anti-simplification guards.

**Kept:** `test_the_sizing_rule_names_the_review_instance_EXACTLY` (the unanchored-PCRE hazard is real — i3 keeps matching `mention-review` against `mention-review-v2`), with its docstring still stating the cost it imposes: it forbids the legitimate `instance="mention-(open|review)"` alternation. Also kept: the negative `window.dimensions` pin in `test_mention_open.py`, the instance-exactness guard, and the engine's own positive/negative controls.

**Green at HEAD:** **603 tests passed** (0 failed, 771s under a box at load ~100) across `test_i3_picker_centering`, `test_mention_open`, `test_i3_render_seam`, `test_i3_game_mode`, `test_no_real_launchers`. `nix-instantiate --parse nix/i3/config.nix` clean — it caught a real bug during this round (a `${home}` in a comment inside the raw nix string, which the renderer tried to interpolate).

### Mutation sweep — re-run in full

Run under `PYTHONDONTWRITEBYTECODE=1` in `cp -a` copies with the worktree's `.git` pointer file removed first, `__pycache__` cleared between mutants. Each mutation is the narrowest expression that can be wrong. Every verdict comes from **two independent reads** — pytest's exit code and the counts parsed from its own terminal summary line — and a disagreement is reported `UNKNOWN`, never as a verdict.

|---|---|---|
| **M01** ledger: workbench OUTPUT height → the workspace height | KILLED | `MEASURED_ledger_is_INTERNALLY_COHERENT`, `WORKBENCH_keeps_rendering` |
| **M02** ledger: laptop OUTPUT height → the workspace height | KILLED | `MEASURED_ledger_is_INTERNALLY_COHERENT` **alone** — the proof that guard is *reachable*, not merely breakable |
| **M19** ledger: swap the workbench's output/workspace heights | KILLED | `MEASURED_ledger_is_INTERNALLY_COHERENT`, `WORKBENCH_keeps_rendering` |
| **M03** collapse both hosts onto `64/77`, ledger unchanged | KILLED | `percentages_are_THIS_HOSTS[laptop]`, `two_HOSTS_DIFFER` |
| **M04** …**and edit the ledger to match** | KILLED | `two_HOSTS_DIFFER` **only** |
| **M05** collapse both hosts onto `90/90`, ledger unchanged | KILLED | `percentages_are_THIS_HOSTS[workbench]`, `two_HOSTS_DIFFER`, `WORKBENCH_keeps_rendering` |
| **M06** …**and edit the ledger to match** | KILLED | `two_HOSTS_DIFFER`, `WORKBENCH_keeps_rendering` |
| **M07** swap the two hosts' pairs | KILLED | `percentages_are_THIS_HOSTS` both arms, `WORKBENCH_keeps_rendering` (**not** `DIFFER` — they still differ) |
| **M08** swap the workbench's axis pair (`64x77` → `77x64`) | KILLED | `percentages_are_THIS_HOSTS[workbench]`, `WORKBENCH_keeps_rendering` |
| **M09** drop `ppt` on the WIDTH axis | KILLED | `is_sized_in_PERCENT_OF_OUTPUT[workbench]` + 6 more |
| **M10** drop `ppt` on the HEIGHT axis | KILLED | `is_sized_in_PERCENT_OF_OUTPUT[workbench]` + 6 more |
| **M11** drop `floating enable` | KILLED | `the_rule_that_RESIZES_…_also_FLOATS_it`, both hosts |
| **M12** drop `move position center` | KILLED | `the_review_window_is_CENTRED_on_both_hosts`, both hosts |
| **M13** move the resize onto the shared `class="float"` rule | KILLED | `no_OTHER_float_…_is_moved_or_resized`, `a_DIFFERENT_float_instance_is_NOT_resized`, `the_PICKER_is_still_not_resized`, `names_the_review_instance_EXACTLY` — 8 cases |
| **M14** rename the instance in the i3 rule only | KILLED | 14 cases — the rule goes inert |
| **M15** rename `REVIEW_CLASS` to a non-prefix value | KILLED | 17 cases, incl. `names_itself_and_sets_NO_geometry` |
| **M20** rename `REVIEW_CLASS` to a **prefix-sharing** value (`mention-review-v2`) | KILLED | `names_the_review_instance_EXACTLY` both hosts + `no_OTHER_float_…` — **this is the over-match case that guard was built for**: i3's unanchored PCRE keeps the rule firing, every sizing guard stays green, and only the exactness guard sees it |
| **M17** laptop height → `100 ppt` | KILLED | `percentages_are_THIS_HOSTS[laptop]`, `review_window_FITS_the_workspace[laptop]` |
| **M18** …**and edit the ledger to match** | KILLED | `review_window_FITS_the_workspace[laptop]` **alone** |
| **M16** re-add a `window.dimensions` flag to the review argv | KILLED | `names_itself_and_sets_NO_geometry` |

**20 mutants, 20 killed**, each with the assertion that caught it.

**Harness validation, before any verdict was read** — two traps from earlier rounds of this PR are specifically guarded: a verdict regex that matched pytest's `short test summary info` banner and printed `SURVIVED` over 8 failures, and a `-k review` filter that silently deselected `test_the_WORKBENCH_keeps_rendering_…`. So the sweep runs **unfiltered**, passes `--color=no` (pytest wraps `FAILED` in ANSI, so `grep '^FAILED'` finds nothing on a coloured log), reads counts only from a line ending `in <n>s`, and cross-checks the parsed `FAILED` name set against the count.

**The control pair, which is what makes the kills above readable:**

| control | result |
|---|---|
| collection positive control | `collected=33`, floor 25 — the selection is wired to something |
| known-GREEN control (no edit) | **SURVIVED** — `0 failed, 33 passed, rc=0` |
| known-FATAL control (delete the review `for_window` rule) | **KILLED** — `14 failed`, and the parsed `FAILED` name set matched the count |
| **comment-only mutant** | **SURVIVED** — `0 failed, 33 passed` |

🔴 **And the harness REFUSED to print a verdict table on its first run of this round** — the `FAILED`-name parser returned 0 names over a log with 14 failures, because `^` was compiled without `re.MULTILINE`. The cross-check between the two reads is what caught it; I would otherwise have quoted killer names from an empty set. That same bug existed in the red/green matrix script and was found the same way. Both are fixed and both runs were redone.

**Known blind spot:** the laptop's pair is `90x90`, so an axis-**order** swap there is a textual no-op and structurally unobservable. The same mutation on the workbench, where the axes are distinct, is in the batch and died.

## 🔴 What is NOT verified

- **Live end-to-end.** The *mechanism* is now read off i3 4.25.1's own source and test suite (above), which is a real upgrade over the previous head's "NOT VERIFIED ANYWHERE". What remains unverified is that this `for_window` **fires at map time on these hosts** and the window comes up at that size. Confirming it needs an `i3-msg reload` and an opened window on the operator's desk.
- 🔴 **The RENDERED GRID — never observed, on either value.** The decoration term is now derived from i3's C (see above) and gives `199x50` for **both** `77 ppt` and `78 ppt`, but nothing has ever been seen on screen: the workbench's `picks.jsonl` does not exist, so this window has never opened there. **And a floating `deco_rect` was unavailable in all three rounds** — the live i3 tree held **zero** floating containers every time it was read, so no real float's inset was ever measured. `floating_resize`'s increment snapping remains a second unmodelled term.
- **The earlier version of this bullet was itself wrong**, and in two ways: it asserted `normal 2` / a titlebar (it is `pixel 2`, no titlebar), and it applied the correction to the **width axis only**. Both are fixed in the comments and here.
- **The laptop's deployed state.** The laptop was read read-only (rects, cell size, i3 version); nothing was deployed to it.

This session changed no window state and ran no state-changing `i3-msg` command — reads only (`get_outputs`, `get_workspaces`, `get_version`, `get_config`, `get_tree`). **Not deployed** — no `home-manager switch`, no `ship.sh`.


ZacxDev:
## ⚠ Round-0 audit findings — this PR changed materially, and **the body you may have read was wrong**

Pushed `451ead89`. If you reviewed `d1f80d4a`, three things you were told are no longer true. Flagging them here rather than only editing the body silently, because two of them were *load-bearing claims in the justification*, not wording.

### 1. One shared `90 ppt 90 ppt` was wrong on the workbench

`ppt` is the right **unit** on any display. It is **not a host-independent size** — and the first version treated it as one.

Re-measured with `TIOCGWINSZ` on each host's own running alacritty pty (it reports the grid in rows/cols **and** in pixels, so the cell size is a division, not a DPI estimate; the pty must be a direct child of alacritty, since a tmux pane's winsize is synthesized by tmux):

| | workbench | laptop |
|---|---|---|
| usable workspace | `3440x1413` | `2256x1480` |
| measured cell | `11.0 x 22.0` px | `19.0 x 37.0` px |
| `90 ppt 90 ppt` renders | **`3096x1272`** | `2030x1332` |

`2200x1100` is the size the operator called *"far too big … unusable"*. `90 ppt` on the workbench is **+41% on that width** — on the host where the review TUI **has never opened**: a TUI opening implies a GitHub reference, which implies `record_pick` ran, and the workbench's `~/.config/mention-open/picks.jsonl` does not exist while the laptop's does (39 rows). The defect was the laptop's throughout.

**Now:** per-host percentages via the `{ isLaptop }` mechanism the file already uses. Laptop `90 ppt 90 ppt`; workbench `64 ppt 78 ppt`, which maps to the **same `200x50` grid it renders today** (`2202x1102` vs `2200x1100` — sub-cell in both axes). The workbench's rendered size does not change at all.

### 2. The lowered `140x40` cell hint **still overflowed the laptop**

`140 x 19.0 = 2660` px against a `2256` px width — **118%**. 140 columns cannot fit a panel whose full-screen terminal is 118 columns.

**Now:** the cell hint is **deleted**, not lowered. `REVIEW_COLUMNS`, `REVIEW_LINES` and both `-o window.dimensions.*` flags are gone; i3 is the sole authority. Alacritty maps at its own default (~`80x24`, inside both workspaces) and i3 resizes **up**, so the flash goes small→right instead of oversized→right. If the rule is ever absent you get a small window, not a broken one.

### 3. 🔴 The body claimed a "conservative standalone fallback" that **no longer exists**, and a ratio that was **never measured**

Both are corrected. The fallback is gone by design (see 2) — there is no second number any more. And the claim repeated in four places —

> *"the laptop's cell is roughly twice the workbench's in each axis — one constant asking for roughly four times the area"*

— is **wrong**: measured, it is **1.73x wider, 1.68x taller, 2.90x the area**. It **over-stated** the gap, which is precisely how one percentage came to look like it could serve both hosts. All three in-tree sites (`nix/i3/config.nix`, `scripts/mention-open.py`, `scripts/tests/test_i3_picker_centering.py`) now carry the measured cell sizes and the measurement method instead of a ratio — a ratio is what went wrong.

### Also new, and deliberately not fixed

The **picker** is marginally too wide on the laptop too: `120x22` cells is `2280x814` px against `2256` — **101.1%, over by 24 px**. Pre-existing, cosmetic, and `format_row` wraps to `PICKER_COLUMNS - 2`, so narrowing it changes a layout you read constantly. Left alone and recorded in the source as a separate decision.

### Gate

The audit fixes reset the verification gate, so nothing here rides on the auditor's word:

- **RED at `d1f80d4a`** (the meaningful base for an amendment) on 4 assertions: the workbench per-host pin, the new hosts-must-DIFFER guard, the new workbench-size-unchanged guard, and the argv negative pin. `percentages_are_THIS_HOSTS[laptop]` is **green** at that base — the shared `90 ppt` happens to equal the laptop's right answer — so it is labelled an **invariant** guard, not counted as regression coverage.
- **Green at HEAD:** 600 tests across `test_i3_picker_centering`, `test_mention_open`, `test_i3_render_seam`, `test_i3_game_mode`, `test_no_real_launchers`.
- **11 mutants, all killed**, each by a named assertion — including the sneaky one (collapse the hosts *and* edit the ledger to match), which only the two guards that don't read the ledger can catch. Control green; a comment-only mutant confirmed **SURVIVED**, which is what makes the kills readable. Two earlier versions of my sweep harness were themselves broken — reporting `SURVIVED` over runs with 8 failures, because the verdict regex matched pytest's `short test summary info` banner — and a three-way cross-check on the failure count is what caught that.
- Blind spot stated: the laptop's `90x90` makes an axis-**order** swap there a textual no-op. The same mutant on the workbench (`64x78`→`78x64`) is in the batch and died.

### 🔴 Still unverified, and the fix rests on it

**Whether i3 honours `resize set <n> ppt <n> ppt` in a `for_window` on a floating container.** It is the only `resize set` directive in `nix/i3/config.nix` outside `mode "resize"`, and that is the only i3 config in the repo — no prior art to argue from. Confirming it needs a live `i3-msg reload` and a real click, which is yours to run: this session touched no window state and ran no state-changing `i3-msg`. **Not deployed** — no `home-manager switch`, no `ship.sh`.

ZacxDev:
## Round-1 audit findings, and what changed

🔴 **If you read this PR before this push, the central arithmetic was wrong and one of its loudest claims was false.** Recording that plainly, because a reviewer may have read the now-corrected numbers.

### FINDING 1 🟡 — `ppt` resolves against the OUTPUT rect, not the workspace. **FIXED.**

Every number and comment in the PR said "workspace". It is the **output**. Established from i3 4.25.1 source (the version on both hosts), and re-verified independently in this round:

- `src/commands.c`, `cmd_resize_set()`, floating branch: `const Con *output = con_get_output(floating_con); … cwidth = output->rect.width * ((double)cwidth / 100.0);`
- `testcases/t/252-floating-size.t`: `fake-outputs 1333x999`, then `resize set 33 ppt 20 ppt` asserted as `int(0.33*1333) x int(0.2*999)` — the **output**. i3's test outputs carry no bar, so workspace == output there, which is why this cannot surface in i3's own suite.

Re-measured read-only, both hosts, i3 4.25.1:

| host | OUTPUT rect (what `ppt` uses) | workspace rect (what the PR used) | bar |
|---|---|---|---|
| workbench | `3440x` **`1440`** | `3440x1413` | 27 px |
| laptop | `2256x` **`1504`** | `2256x1480` | 24 px |

**Width is unaffected** (the bar is a full-width strip). **Height was understated by 27 px / 24 px.**

**Consequence, and the PR's loudest promise was false:** *"the workbench's size does not change at all … +1.6 × +2.1 px, sub-cell in both axes"*. `78 ppt` of **1440** is ≈`1123` px, not `1102` — **+23 px over today's 1100, larger than one 22 px cell**, rendering **200×51** cells, not 200×50.

Reproduced before fixing, as the proof the guard is load-bearing: substituting the two output heights into `_MEASURED` as the **only** change flipped `test_the_WORKBENCH_keeps_rendering_the_size_it_ALREADY_renders` to RED, verbatim —

```
AssertionError: the workbench's `resize set 64 ppt 78 ppt` maps to 200x51 CHARACTER CELLS
(2202x1123 px of a 3440x1440 workspace at a measured 11.0x22.0 px cell). It has to come to 200x50 …
```

**Fixed:**
- **Workbench is now `64 ppt 77 ppt`.** Re-derived, not copied: `77 ppt` of 1440 = `1108.8` px → 50 rows at the measured 22.0 px cell, preserving today's 200×50. (`76 ppt` → 1094.4 = 49 rows; `78 ppt` → 1123.2 = 51. The admissible window for 50 rows is 76.39–77.92 ppt.) Width stays `64 ppt` — `2201.6` px of 3440 → 200 cols; the width was never affected.
- **Laptop stays `90 ppt 90 ppt`** — **confirmed, not assumed**: `2030.4 × 1353.6` px of the `2256x1504` output, inside the `2256x1480` workspace in both axes, and centring (which uses the workspace rect) puts it at y≈87..1441 of 24..1504. A new guard asserts this mechanically rather than leaving it to arithmetic in a comment.
- `_MEASURED` now carries **both** rects plus the cell, per host, keyed `output` / `workspace`, with the wrong-rect class of error pinned by a new coherence guard (below).
- **The claim was swept for, not fixed only at the named sites.** Corrected in `nix/i3/config.nix` (let-block + rule comment), `scripts/mention-open.py` (both comment blocks + the inline `Popen` comment), `scripts/tests/test_i3_picker_centering.py` (module docstring, guard-3 header, every assertion message, the parser docstring, two test names), `scripts/tests/test_mention_open.py`, and the PR body. Each site now cites `cmd_resize_set` and the bar height so the next reader can check it.

### FINDING 5 — `move position center`'s basis. **ESTABLISHED FROM SOURCE, and it differs.**

`cmd_move_window_to_center()` in the same file: for `position` it calls `floating_center(floating_con, con_get_workspace(floating_con)->rect)` — the **WORKSPACE** rect. Only `move absolute position center` uses `croot->rect`. So this one chain **sizes against the output and places against the workspace**. That is now stated at every site rather than guessed, and it is what the new fit guard tests: a size taken as a percentage of the larger rect has to fit inside the smaller one.

### FINDING 3 🟡 — the deploy window where neither half sizes the TUI. **DOCUMENTED (not automated).**

Verified: `scripts/mention-open.py` is exec'd straight out of the working tree (`nix/programs/alacritty/default.nix`), so `git pull` makes the hint deletion live with **no switch**; `nix/i3/config.nix` is `xdg.configFile."i3/config"` (`nix/graphical.nix`), needing a switch **and** an explicit reload — and **nothing in `nix/` reloads i3** (the workbench's running config, read via `i3-msg -t get_config`, carries no `mention-review` rule).

So the ordered sequence is now explicit in the PR body **and** in the config's comment block: **merge → `scripts/ship.sh` → `i3-msg reload` on each host**, with the reload stated as a **REQUIRED deploy step**, and the intermediate state named: ~`80x24` on both hosts, small-but-usable, and on the workbench a temporary change from 200×50. An automatic reload in the activation script was **deliberately not added** — it would reload i3 on every future `home-manager switch` — and is recorded as an optional follow-up for the operator.

### FINDING 2 🟡 — decoration and increment snapping are unmodelled. **RECORDED, not chased to a number.**

`nix/i3/config.nix` sets `default_border pixel 2` and never `default_floating_border`, so floats take i3's default: `src/config.c` sets `config.default_floating_border = BS_NORMAL` and `default_floating_border_width = logical_px(2)` — a titlebar plus borders. A floating con's `rect` **includes** that, so the client area is smaller than the rect and the true grid is likely ~199×50 rather than 200×50. `floating_resize` (`src/floating.c`) adds a third term, snapping the decorated rect up to the window's size increments.

Quantifying either needs a window opened or resized on the live desk, which this session must not do — and there were **no floating windows in the tree at all** when it was read, so even an indirect `deco_rect` measurement was unavailable. So: a clearly-labelled ⚠ now sits in the config's comment block, in `mention-open.py`, and in the test file, saying the arithmetic models the **rect only** and the grid is therefore **approximate** — and **no pixel figure anywhere in this PR is quoted to a tenth any more**. The over-precision was the defect, independent of the true value.

### FINDING 4 🟢 — a mutation-table row overstated its killers. **CORRECTED, and re-measured.**

The old table claimed the collapse-and-edit-the-ledger mutant was killed by both `two_HOSTS_DIFFER` and `WORKBENCH_keeps_rendering`. That is direction-dependent, and the re-run sweep in this round reports the killers **per direction** rather than merged. See the table in the PR body.

### The gate was reset

Every assertion was re-verified and the **whole mutation sweep re-run** — nothing above is shipped on the auditor's authority. Two **new** guards came out of it:

- **`test_the_MEASURED_ledger_is_INTERNALLY_COHERENT`** — the direct answer to finding 1. Every other guard reads `_MEASURED` as ground truth, so a wrong ledger is invisible to all of them; this PR was green with the workspace height in the output slot. It cannot measure a display, but it asserts the relationship a bar-on-an-output must produce: equal widths, and a height gap that is positive and bar-sized. A **zero** gap is exactly the substitution the audit found. The ledger's own docstring now says it is a transcribed **input**, not a derived fact, and says to re-measure rather than nudge a number to make a guard pass.
- **`test_the_review_window_FITS_the_workspace_it_is_CENTRED_in`** — the only assertion checking the laptop's `90x90` against something other than the `_REVIEW_PPT` ledger.

Both are labelled **invariant guards**, not regression coverage: the pre-fix tree had no `output` key and both hosts already fit, so neither could be red against an earlier commit. Their killing power is shown by mutation instead — and the laptop-rect mutant is killed by the coherence guard **alone**, which is what proves it reachable rather than merely breakable.

`nix-instantiate --parse` earned its keep this round: it caught a `${home}` written into a comment **inside the raw nix string**, which Nix tried to interpolate.

No window state was changed and no state-changing `i3-msg` was run — reads only. Not deployed.

### Verification, measured this round

- **Red/green at `451ead89`** (that ref's production files, these tests): `2 failed, 31 passed` — `test_the_WORKBENCH_keeps_rendering_the_size_it_ALREADY_renders` and `test_the_review_windows_percentages_are_THIS_HOSTS[workbench]`. At `d1f80d4a`: `3 failed, 30 passed` (+ `two_HOSTS_DIFFER`). At `origin/main`: `14 failed, 19 passed`.
- **Green at HEAD**: `603 passed` across `test_i3_picker_centering`, `test_mention_open`, `test_i3_render_seam`, `test_i3_game_mode`, `test_no_real_launchers`.
- **Mutation sweep re-run in full: 20 mutants, 20 killed**, each with the assertion that caught it. Controls: collection `33` (floor 25), no-edit control **SURVIVED** green, known-fatal control **KILLED** with the parsed name set matching the count, and a **comment-only mutant SURVIVED**. Table in the PR body.
- **Finding 4 reproduced independently**: collapse-onto-`64/77` + ledger edit dies by `two_HOSTS_DIFFER` **only**; collapse-onto-`90/90` + ledger edit dies by **both**. The old table was right about one direction and overstated the other, exactly as the audit said.
- 🔴 **The sweep harness REFUSED to print verdicts on its first run** — the `FAILED`-name parser returned 0 names over a log with 14 failures, because `^` was compiled without `re.MULTILINE`. The two-read cross-check is what caught it; the same bug was in the red/green matrix script and was found the same way. Both fixed, both runs redone. This is the third harness-parsing trap on this PR, in a new shape.
- **M20 was added** because the batch had a hole: the over-match direction (`REVIEW_CLASS` renamed to a **prefix-sharing** `mention-review-v2`, rule unchanged) is the case `names_the_review_instance_EXACTLY` exists for — i3's unanchored PCRE keeps the rule firing and every sizing guard green. It dies by that guard on both hosts.

ZacxDev:
## Audit ladder — round 1 claims block

Round 0 (requirements & deletion) reported and fixed nothing; its verdict is prose in an earlier comment. Round 1 was the first correctness round (nine axes, dispatched **blind**). Its fixes are `13a2a0ce`.

**Payload/scaffolding call, made once for this ladder (revert test):** payload = `nix/i3/config.nix` + `scripts/mention-open.py` (revert either and the deliverable does not ship). Scaffolding = `scripts/tests/test_i3_picker_centering.py` + `scripts/tests/test_mention_open.py`.

**Ledger — round 1 · payload lines changed THIS round: 218** (`nix/i3/config.nix` +115/-38, `scripts/mention-open.py` +50/-15) · since round 1: 218. Measured `451ead89..13a2a0ce --not origin/main --remerge-diff`, rc 0, range non-empty.

```audit-claims round=1 audited=451ead89..13a2a0ce
1. `resize set … ppt` on a floating container resolves against the OUTPUT rect, not the workspace rect — corrected at every site that said "workspace", with the `cmd_resize_set` source citation and both rects tabulated per host.
2. The workbench pair changed `64 ppt 78 ppt` -> `64 ppt 77 ppt`, because 78 ppt of the 1440 output height is 1123.2 px and renders 51 rows, not the 50 the PR claimed.
3. `_MEASURED` now carries `{output, workspace, cell}` per host and its docstring says it is a transcribed INPUT that must be re-measured, not a derived fact.
4. `move position center` was established from source to centre against the WORKSPACE rect while resize sizes against the OUTPUT — the asymmetry is stated at every site and a new fit guard tests it.
5. Two guards added: `test_the_MEASURED_ledger_is_INTERNALLY_COHERENT` and `test_the_review_window_FITS_the_workspace_it_is_CENTRED_in`. Both are labelled INVARIANT guards, not regression coverage, because the pre-fix tree had no `output` key and both hosts already fit.
6. The required deploy sequence (merge -> `ship.sh` -> `i3-msg reload` per host) is now stated as an ordered deploy step rather than only a verification step, with the ~80x24 intermediate state named.
7. Every pixel figure stopped being quoted to a tenth; decoration (`default_floating_border` unset, i3 default `BS_NORMAL`) and increment snapping are recorded as NOT modelled.
8. The round-0 mutation row that overstated its killers was corrected to name the killer per collapse direction.
```

### Still NOT established, and no audit round can reach it

Whether the `for_window` actually fires at map time and the window comes up at that size. That needs `i3-msg reload` plus an opened window on a machine the operator is using. The *mechanism* is now read off i3 4.25.1 source and i3's own implementing testcase, which is a real upgrade over the earlier "not verified anywhere" — but it is not a live observation and is not reported as one.

Finding 2 could not be measured even indirectly: there were **zero floating containers in the live tree**, so no `deco_rect` was obtainable.

ZacxDev:
# Round-2 audit findings, and what changed (`13a2a0ce` → `70e2ba07`)

🔴 **Reviewers who read this PR before `70e2ba07`: several claims you read were wrong and have been corrected.** This round is documentation-correctness. The payload is unchanged — **no assertion changed, no percentage changed, one failure-message string reworded.** That is the point: this PR's payload is ~5 lines of code and ~150 lines of comment, and round 1, *while documenting a term it had just discovered it was ignoring*, asserted the wrong mechanism for it in **five places, two of them payload files**.

---

## F1 — the decoration note named the wrong border style, and that voids the stated rationale for round 1's only code change

**What was claimed** (`nix/i3/config.nix:128-131`, `scripts/mention-open.py:1741-1742`, `scripts/tests/test_i3_picker_centering.py:431-434`, and the PR body twice):

> this file sets `default_border pixel 2` but never `default_floating_border`, so floats take i3's default — `config.default_floating_border = BS_NORMAL` … i.e. a titlebar plus borders

**It does not hold for this window.** Verified against the realised i3 4.25.1 source the running binary was built from — `/proc/$(pgrep -x i3)/exe` → `/nix/store/g8p0…-i3-4.25.1`, then that store path's derivation's `src`:

| leg | source |
|---|---|
| `config.default_floating_border` is applied **only** under `if (automatic)` inside `floating_enable()` | `src/floating.c:353-354` |
| `for_window … floating enable` is a **command**: `cmd_floating()` → `floating_enable(con, false)` | `src/commands.c:1157` |
| `run_assignments()` runs **after** the `want_floating` decision | `src/manage.c:588` (map time; `:746` on the remanage path) vs `:462-546` |
| `want_floating` is set only by window-type atoms / `_NET_WM_STATE_MODAL` / sticky / `transient_for` / a fixed `min==max` hint — an alacritty toplevel matches none | `src/manage.c:462-546` |
| so the con keeps `con_new_skeleton()`'s `config.default_border` = **BS_PIXEL / `logical_px(2)`** | `src/con.c:44` |
| `current_border_width` comes from the **non-floating** arm | `src/manage.c:546` |
| the non-BS_NORMAL inset is `{bw, bw, -2bw, -2bw}` → client = rect **− 4 px in BOTH axes**, no titlebar | `src/con.c:1846-1849` |

**Live, read-only** (`i3-msg -t get_tree` / `-t get_config`, i3 4.25.1): the running alacritty windows report `border=pixel`, `current_border_width=2`; the config carries no `default_floating_border`, no `new_float`, no `for_window … border`.

⚠ **One nuance worth stating because it nearly inverted the reading.** Those live windows sit in a **tabbed** parent, and `con_border_style()` (`src/con.c:1942`) overrides a non-BS_NORMAL style to BS_NORMAL for a >1-child tabbed container — which is why their client *height* is rect−2, not rect−4. A float's parent is a `CT_FLOATING_CON` with `layout = L_SPLITH` (`src/floating.c:291`), so no override applies and −4/−4 is the float's case. **The live read is evidence for the border STYLE and the CELL SIZE; the float's own inset comes off the C.** The model is validated against the live window: `floor(3436/11)=312`, `floor(1391/22)=63`, matching the observed `312x63`.

### The consequence

| `ppt` | rect | client (BS_PIXEL bw 2) | cells |
|---|---|---|---|
| `64 ppt 77 ppt` | `2201 x 1108` | `2197 x 1104` | **`199 x 50`** |
| `64 ppt 78 ppt` | `2201 x 1123` | `2197 x 1119` | **`199 x 50`** |

1. **`51 rows` was a count of the BARE RECT.** `1123 // 22 = 51`; the client height is `1119`, and `1119 // 22 = 50`. The decoration the same commit documented absorbs the row. So *"`78 ppt` … is 51 rows: one cell MORE than today, i.e. a visible change on the host this fix promises not to touch"* was **false**.
2. **`77` and `78` are indistinguishable on screen.** 🔴 **`77` was NOT changed back, and no replacement rationale was manufactured.** It is retained because it is the value already committed and deployed on this branch — and the comments now say that, plainly. *"I could not find a purpose"* is the finding; a purpose found under pressure to supply one is a hypothesis, and this ladder has already watched a claim go through several drafts, each retracted by the next round.
3. **The one-sided correction is fixed.** Round 1's own estimate (*"the true grid is likely ~`199x50`"*) applied the border correction to the **width axis only** — the claims-a-relationship-inspects-one-side shape, in prose. It is now applied to both axes everywhere it appears.
4. 🔴 **The over-promise is RETIRED, not re-justified.** `199x50` ≠ today's `200x50`, so *"the workbench must come out of this change rendering what it renders now"* **is not met in the width axis either — for `77` or `78`.** The height axis matches (50 rows); the width is one column short. Every site that made that promise now says so.
5. **The rendered grid is NOT ESTABLISHED.** It has never been observed, because the review window has never opened on the workbench: `~/.config/mention-open/picks.jsonl` **does not exist** there (measured). The laptop — the actual defect — fits comfortably either way.

### The sweep

`grep` on this box is a function wrapping ugrep and `-r` honours `.gitignore`, so the sweep enumerates instead:

```
find <repo> -path '*/.git' -prune -o -type f -print0 \
  | xargs -0 /run/current-system/sw/bin/grep -nIE 'BS_NORMAL|normal 2|titlebar|default_floating_border'
```

**Positive control** (same enumeration, a pattern known to match): `default_border pixel 2` → 4 files, counts `1/2/1/2`. **Sites found: exactly the three prose blocks named above** — `nix/i3/config.nix`, `scripts/mention-open.py`, `scripts/tests/test_i3_picker_centering.py` — plus one unrelated hit (`loop_wedge.test.mjs`, "normal 204 idle"). A second sweep for the *consequence* claims (`51 row|one row MORE|one cell MORE|SAME grid|renders what it renders now|must keep rendering|renders TODAY`) found the two remaining sites in `config.nix` and the test file's guard docstring plus failure message, and the `_DEPLOYED_REVIEW_CELLS` / `_REVIEW_PPT` docstrings. **No sixth border-style site existed.** Post-fix re-sweep: every remaining hit is corrected text that quotes the old claim as retracted.

---

## The smaller findings — all fixed

| # | site | finding |
|---|---|---|
| **F2** | `test_i3_picker_centering.py:635-636` | *"both read as the same grid"* had **no antecedent** and is false on its nearest reading (against `1413`: `77 ppt` → 49 rows, `78 ppt` → 50). The intended "both" is {`78 ppt`, today's `200x50`}; `config.nix` already said it unambiguously and that wording is mirrored. |
| **F3** | `test_i3_picker_centering.py:33` (and the GUARD 3 header, found by sweep) | *"every height in it was wrong by the status bar"* **over-states it**. Measured at `451ead89`: only heights **derived from a `ppt` percentage** were wrong (`1332`, `1272`, and the `"78%"` fit). The transcribed rects (`1413`, `1480`), cell heights (`22.0`, `37.0`), observed row counts (`63`, `39`) and the `1100` px of a 50-row grid were all correct. As written it **invited converting the FIT basis to the output rect**, which this same PR forbids elsewhere (*"Do not carry one rect over to the other"*). |
| **F4** | `test_i3_picker_centering.py:440` | *"The guards below pin FIVE things"* over a block holding **nine** test functions (round 1 added two while incrementing the total by one). 🔴 **Fixed by FORM, not by the number:** the list is numbered and the reader is told to count it. A total kept beside its own list drifts again. |
| **F5** | `test_i3_picker_centering.py:688` | *"the only assertion … other than the `_REVIEW_PPT` ledger"* is overstated — `test_the_two_HOSTS_review_percentages_DIFFER` and the sizing guard's `0 < v <= 100` bound read no ledger either. The true claim is "the only assertion that checks it against a **measured rect**". |
| **F8** | the `bar > 0` message in `test_the_MEASURED_ledger_is_INTERNALLY_COHERENT` | rendered *"a negative value means the same rect has been written into both slots"* — **false**; a negative gap means swapped or wrong rects. Each branch now states what **its own** condition implies, and the docstring enumerates all three failing shapes (zero / negative / implausibly large) as **different faults**. |
| **F6** | PR body | *"three message strings"* → **two**. `git diff --numstat 451ead89 13a2a0ce -- scripts/tests/test_mention_open.py` is **9 insertions / 3 deletions**: one docstring reword, six new docstring lines, two message strings. ✅ The load-bearing half — *"changed no assertion"* — is **true**. |
| **F7** | PR body | `picks.jsonl` *"(39 rows)"* is now **40**. 🔴 **It has stopped being quoted rather than been pinned** — the count grows with use, nothing asserts on it, and the load-bearing claim is presence vs absence. |

### F9 — `nix/graphical.nix:14-15` was measured FALSE, and it mattered

The note said writing `~/.config/i3/config` is *"INERT until the system stops forcing `i3 -c /etc/i3.conf` — run `sudo bash nix/system/apply-i3-to-hm.sh` to cutover"*. Outside this PR's payload and pre-existing, so it was **re-measured before editing** — read-only, **both hosts**:

```
ls /etc/i3.conf                       → No such file or directory   (workbench AND laptop)
tr '\0' ' ' < /proc/$(pgrep -x i3)/cmdline
                                      → i3 -a --restart /run/user/1000/i3/restart-state.<pid>
                                        (no -c, both hosts)
i3-msg -t get_version                 → loaded_config_file_name: /home/zach/.config/i3/config
                                        (both hosts, 4.25.1)
grep -n 'configFile|i3.conf' /etc/nixos/configuration.nix   → no matches (workbench)
```

**The cutover already happened.** Left alone the note would have stopped someone following this PR's own `merge → ship.sh → i3-msg reload` step, by telling them the edit cannot take effect. Corrected, and it now says it was measured. 🔴 **What still applies is stated in its place:** `~/.config/i3/config` is a nix-store copy (`readlink -f` → `/nix/store/…-hm_i3config`), so an edit needs a `home-manager switch`; and i3 does not re-read its config on change, so the switch needs an explicit `i3-msg reload`.

---

## Tests

🔴 **The code is unchanged by this round, so there is no new regression coverage — and none was manufactured.** Saying that plainly is the honest report; a guard invented to fill the row would pin nothing this round introduced.

**Suite: 603 collected, 603 passed** across `test_i3_picker_centering`, `test_mention_open`, `test_i3_render_seam`, `test_i3_game_mode`, `test_no_real_launchers` — exactly round 2's number, **no difference**. Two independent reads, unpiped: `603` ` PASSED ` result lines and the verdict line `603 passed, 2 warnings in 97.06s`; `0` FAILED/ERROR; collection control `collected 603 items` (non-zero). (The 2 warnings are pre-existing `SyntaxWarning`s in `test_no_real_launchers.py`, untouched.)

### M02 re-verified — because F8 edited that guard's message

In a `cp -a` copy with the worktree's **`.git` pointer file removed first** (confirmed absent), under `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` cleared per mutant, **unfiltered**, `--color=no`, `^` compiled with `re.MULTILINE`, and every verdict taken from **two independent reads** (the tail summary counts **and** a re-derived count of parsed `FAILED` name lines) with a disagreement reported `UNKNOWN`:

| run | verdict | detail |
|---|---|---|
| control: no edit (known-GREEN) | **SURVIVED** | `collected=33`, `0 failed`, rc=0 |
| control: delete the review `for_window` rule (known-FATAL) | **KILLED** | `14 failed`, name set matched the count |
| **M02** laptop OUTPUT height → workspace height | **KILLED** | by `test_the_MEASURED_ledger_is_INTERNALLY_COHERENT` **ALONE** |
| **M19** swap the workbench's output/workspace heights (added, to reach the **negative** branch) | **KILLED** | `MEASURED_ledger_is_INTERNALLY_COHERENT` + `WORKBENCH_keeps_rendering` |

M02 dies at **that guard's own assertion** (`test_i3_picker_centering.py:682`) rendering the **new** zero-branch text: *"A gap of EXACTLY ZERO means the same rect has been written into both slots."* M19 reaches the other branch and renders *"A NEGATIVE gap means the two rects are SWAPPED, or one of them is not this output's rect at all — it does NOT mean they are identical; that is the zero case."* **Both branches were exercised; neither was asserted from reading the code.**

🔴 **And the harness refused to print a verdict on its first run — a fourth harness bug on this PR.** It ran with `-q`, which prints neither `collected N items` nor a `=====`-decorated tail, so **both** regexes missed and it reported `UNKNOWN` plus a failed collection positive control rather than guessing. Format is a dependency you do not pin by wanting to; the fix is in the harness and noted in its source.

**`nix-instantiate --parse` clean** on `nix/i3/config.nix` and `nix/graphical.nix`.

`scripts/gate.sh` was not run (~20 min, and nothing here can reach a new target); `scripts/scoped-tests.sh` refuses with exit 4 on a `nix/**` diff, as expected.

---

## What I could NOT verify

- 🔴 **The rendered grid, on either value.** Never observed. The workbench's `picks.jsonl` does not exist, so this window has never opened there.
- 🔴 **A real floating `deco_rect` — still unavailable, now in all three rounds.** The live i3 tree held **zero** floating containers (3 windows: `i3bar`, `Alacritty`, `Brave-browser`, all `floating: auto_off`). **Nothing was worked around by opening a window**: the float's −4/−4 inset is derived from i3 4.25.1's C and the live read supplies only the border style and the cell size.
- **`floating_resize`'s size-increment snapping** — a second unmodelled term, unmeasurable without resizing a window on the live desk.
- **End-to-end:** that the `for_window` fires at map time and the window comes up at that size. Needs `i3-msg reload` plus an opened window.

**This session raised, focused, moved and resized nothing, and ran no state-changing `i3-msg`** — reads only (`get_tree`, `get_version`, `get_config`). **Not deployed:** no `home-manager switch`, no `ship.sh`. The laptop was read over SSH under the same rules.

ZacxDev:
## Audit ladder — DELIBERATELY ENDED after round 2's fixes (`70e2ba07`)

🔴 **Read this as *ended*, not as *converged*.** Round 2 returned findings that were fixed, so the findings-keyed rule would ordinarily schedule a round 3. I am stopping instead, on the prose-payload criterion — and per that criterion's own requirement, here is why the rounds would not stop on their own, stated rather than left implicit.

### Why this ladder cannot terminate by itself

**The payload is prose.** This PR ships **~5 lines of code** (`reviewSizePpt`'s two operands, the `for_window` rule, two deletions) and **~150 lines of comment** explaining them. Consequences:

- **The attribution gate is structurally inert here.** It counts payload lines, and comment lines inside `nix/i3/config.nix` and `scripts/mention-open.py` *are* payload lines — so every round is non-zero by construction and the two-zero-rounds stop can never fire, while the ladder does exactly what that gate exists to catch.
- **"Fixed a defect" and "reworded a warning" are the same edit.** Round 1's fix introduced a false claim *while documenting the very term it had just discovered it was ignoring* (`default_floating_border` → it asserted `BS_NORMAL`; the real style is `BS_PIXEL`/2px, because `for_window … floating enable` reaches `floating_enable(con, false)` and the `automatic` branch never runs). Round 2's fix was **entirely prose** correcting that claim. A round 3 would audit prose that round 2 wrote. That is the non-terminating shape, not a hypothetical.

### The criterion's conditions, each met

- **No 🔴.** Round 2: one 🟡 (F1) and seven 🟢. No round in this ladder produced a deploy-blocker.
- **Blast radius is "a comment contains a false sentence."** The deployed value `77` renders **identically** to `78` under the corrected border model (client 2197×1104 vs 2197×1119 → **199×50 both**), the laptop's reported defect is fixed with room to spare, and the workbench's review window **has never opened at all** (`~/.config/mention-open/picks.jsonl` does not exist on that host).
- **The recurring SHAPE was swept at every site, not just the reported one.** Enumerated with `find … -print0 | xargs -0 grep` rather than `grep -r` (the wrapper honours `.gitignore`), with a positive control (`default_border pixel 2` → 4 files, 1/2/1/2). It found the 3 prose blocks, **2 sites beyond the brief's list** (the GUARD 3 header's "every height in this block", and `_DEPLOYED_REVIEW_CELLS`'s "the one it must keep rendering") and **2 more in the PR body**. Post-fix re-sweep returned zero against an in-regex positive control.

### 🔴 WHAT IS NOT FIXED — open, not absent

1. **`test_the_WORKBENCH_keeps_rendering_the_size_it_ALREADY_renders` still carries that name, and the name over-states what it pins.** Its docstring now opens by telling the reader to read the name narrowly and enumerates the three things it does not establish. Renaming is a code change that would ripple into the red-at-base matrix and the 20-mutant ledger, so it was deferred deliberately. **Worth renaming in a follow-up.**
2. **`77` has no justification better than "it is what is already committed and deployed."** Round 1's stated reason (78 ppt → 51 rows → a visible change) is **false** — 51 counts the bare rect and the decoration absorbs it. The rationale is recorded as void **and unreplaced**; no fresh one was invented, deliberately.
3. **The promise that the workbench "renders what it renders now" is NOT met in the width axis**, for `77` or `78`: 199×50, one column narrower than today's 200×50. Retired in the text rather than left standing.
4. **The rendered grid has never been observed**, on either value or either host.
5. **A floating `deco_rect` was unavailable in all three rounds** — zero floating containers in the live tree every time. The −4/−4 inset is derived from i3's C source, not measured. ⚠ The live alacritty windows sit in a **tabbed** parent where `con_border_style()` overrides to `BS_NORMAL`, so their client height is rect−2, not rect−4; a float's parent is `L_SPLITH` so no override applies. The live read is evidence for the border *style* and the *cell size* only.
6. **`floating_resize`'s increment snapping is unmodelled.**
7. 🔴 **The live end-to-end is unverified and no audit round can reach it** — whether the `for_window` fires at map time and the window comes up at that size needs `i3-msg reload` plus an opened review window on the operator's desk. The *mechanism* is now read off i3 4.25.1's source and i3's own implementing testcase, which is a real upgrade over "not verified anywhere" — but it is not a live observation and is not reported as one.

### Ledger

`round 2 · payload lines changed THIS round: 320 insertions / 100 deletions across 4 files, of which ALL are comment/docstring prose except one failure-message string · since round 1: 538` — and that number is precisely why the gate cannot fire here.

### What the verification actually cost, recorded because it recurs

**Four separate sweep-harness bugs on this one PR**, each of which would have inverted a verdict: a `^` compiled without `re.MULTILINE` returning 0 names over a log with 14 failures; a verdict regex matching pytest's `short test summary info` banner and printing SURVIVED over 8 failures; a `-k review` filter silently deselecting the workbench guard; and `-q` printing neither `collected N items` nor a decorated tail, so both regexes missed. Every one was caught by a control or a cross-check, not by reading more carefully. **Validate the instrument before reading its verdict** — and keep a control pair (a no-edit mutant that must stay green, a comment-only mutant that must SURVIVE) so the kills are readable at all.

### Deploy — required, ordered, and not yet done

`merge → scripts/ship.sh → i3-msg reload` on each host. `nix/i3/config.nix` is a `home.file` target so **merging changes nothing**; and between the switch and the reload the TUI opens at alacritty's default ~80×24 on both hosts, because the new script passes no dimensions while the old i3 config has no sizing rule. Small and usable, not overflowing — but it is a real intermediate state and the reload is a deploy step, not a verification step.

PR #1621, opened 2026-09-12
gate(docs): a skill may not spell out a checkout path a handle already names

ZacxDev:
`claude/skills/**` is loaded VERBATIM into an agent's context as authoritative
instructions. 21 sites across 9 files spelled out the literal checkout path
`/home/zach/workspace/devrc/...` inside runnable commands — a second copy of a value
`nix/agent-handles.nix` already exports into every agent shell (zsh + opencode),
existence-guarded, from one source. Nothing reconciled the two.

This is the CLASS that rank 23 (#1583) did not close. Rank 23's own closing condition
was a `grep` for one spelling, which is why it could be met while the class stood.

## What this does

**A gate** — `scripts/tests/test_absolute_handle_paths.py`. Rejects any absolute
checkout path in the `claude/` + `CLAUDE.md` corpus that a handle already names, and
prints the remedy:

```
claude/skills/resume/SKILL.md:266  /home/zach/workspace/homelab-talos/scripts/foo.sh  ->  use `$HOMELAB/scripts/foo.sh`
```

Three design points worth review:

- **The handle table is parsed out of `nix/agent-handles.nix`**, not hardcoded. That
  file is already the acknowledged single source for both consumers (the zsh
  `envExtra` block is generated from it, and its header says adding a handle by hand
  is "exactly the drift this replaced"). So the gate cannot drift from the handles,
  and it emits the right remedy for free.
- **It rejects the SPELLING, never resolves the path.** The sibling gate
  `test_doc_path_rot.py` deliberately skips absolute paths — "Absolute paths are never
  claims this repo can settle" — which is precisely why these 21 went unseen. Matching
  text is host-independent; stat'ing would not be.
- **`${home}` matches any prefix**, so `/home/alice/workspace/devrc/x` is caught too.
  A gate keyed on `/home/zach` would be a spelled guard, walkable by any other user.

**Two tiers.** `scripts/tests` also runs inside `flake.nix` `checks.pytests` over a
copy of the flake source with no `.git`, where an unguarded `git ls-files` exits 128.
The corpus falls back to a filesystem walk, as `test_doc_path_rot.py` does. A guard
that can only pass in one tier is a defect this repo has already shipped twice.

**21 sites cleared, 1 exemption.** 16 → `$DEVRC`; `prune-index:90`'s
`DEVRC=/home/...` assignment deleted outright (it redefined the handle that already
exists); `mailbox:133` is Python, so `os.environ["DEVRC"]` (`$VAR` does not expand
there); doc pointers → the deployed `~/.claude/skills/...` spelling.

The one exemption is `cross-session-reach.md`'s two recorded `clawgatectl` JSON
payloads, which show `path`/`cwd` as the API actually emits them. Rewriting a wire
payload would misrepresent the format. The ignore list is keyed on `(doc, literal)`
so an exemption cannot leak to a doc that copies the string, **requires a written
reason** (`test_every_ignore_entry_carries_a_reason`), and **can only shrink**
(`test_no_stale_ignore_entries`) — an escape hatch, not the coverage mechanism.

## Evidence

**Red → green on a planted violation, verified independently of the implementing
session**: a *different* handle (`$HOMELAB`) planted into
`claude/skills/resume/SKILL.md`, a file this diff does not touch → 1 failed / 35
passed, failing on the gate's own assertion and naming the remedy. Reverted → 36
passed. Identical in the index tier and in a `.git`-stripped copy.

**Mutation battery: 13 mutants, 12 killed by their intended test on that test's own
message**; `PYTHONDONTWRITEBYTECODE=1`, `__pycache__` wiped per mutant, tree restored
and sha256-verified byte-identical, pristine positive control watched green first.
Mutants included the handle table emptied, `${home}` narrowed to a literal
`/home/zach`, the corpus narrowed so `CLAUDE.md` escapes, the ignore list swallowing
everything, and the two-tier `.git` fallback removed.

🔴 **One mutant SURVIVED and is reported rather than hidden**: lowering the
corpus-size floor to 0 on a healthy corpus is vacuous by construction — nothing can
observe it while the corpus is non-empty. M5 (corpus collapsed) proves that guard
reachable and attributable, so the floor is not unguarded; the specific mutant is
simply unobservable. Two further mutants survived their FIRST run and the fixes are
the interesting part: the shared ignore-filter and the reason check each had no
killing mutation until a test was added that makes them fire.

Also found while building: a latent `rfind`-vs-`find` bug in the remedy builder that
would have silently deleted the middle of a repeated-suffix path.

## Two places the brief was wrong, recorded rather than worked around

- **`$KC_WORKBENCH` / `$HOMELAB` are NOT a valid remedy for the
  `ls .../homelab-{talos,infra}/...` probes.** The checkout is `homelab-talos` on the
  workbench and `homelab-infra` on the laptop, and the handle is existence-guarded —
  so it is absent exactly where the other name is correct, and an empty `KUBECONFIG=`
  falls back to `~/.kube/config`, i.e. a silently wrong cluster. Those lines keep the
  probe; only the hardcoded user was dropped.
- **The test-count floor did not need re-pinning.** A floor can only be breached from
  below, and adding tests cannot do that (`collected=14625 floor=13026`). Re-pinning
  would have written a number measured across other branches' tests.

`deploy.md` also picked up a real fix in passing: the recipe named `homelab-talos` at
steps 0/5/6 and `homelab-infra` at step 4, so it could not run end-to-end on either
host.

## Verification status, stated separately

The full `scripts/gate.sh --tier pytest` was last run **before** the final rebase onto
`22ddd8dc`: `SCOPE: FULL (29 of 29)`, 28 of 29 targets PASS including
`scripts/tests (collected=14625 passed=14625 skipped=0 floor=13026)`. The single
failure was `browser-bridge::test_the_release_handler_EXITS_rather_than_resuming[INT]`
— a module this diff does not touch, which passes in isolation in both this tree and a
pristine base worktree, and which `origin/main`'s own HEAD commit (#1585) is titled
for: *"the [INT] split verdict is a STARVED INSTRUMENT, not a handler defect"*. The
box was at load ~50 with four concurrent suites.

On the final rebased tree a targeted **1162 passed** was run over
`test_absolute_handle_paths` + `test_index_append_protocol` + `test_prune_skill_size` +
`test_doc_path_rot` + the whole `browser-bridge` target. **The full gate has not been
re-run since the rebase** — CI is the authoritative run.

The two test re-pins in `f4241883` are derived-figure updates, not weakenings:
`prune-skill/SKILL.md` got 60 B *smaller* and moved further inside its budget with
`MAX_BYTES` untouched, and the prose-sha guard exists precisely to force a doc edit to
be acknowledged — the acknowledgement it demands (re-reading the other door) was
performed and is recorded above the constant.

https://claude.ai/code/session_01WfAUmKX4fQykF5GSoEztFv


ZacxDev:
## Round 1 — audit + fixes

Blind full audit (nine axes) against `22ddd8dc..2b4d805e`. Result: **🔴0 · 🟡2 · 🟢5**, verdict safe to merge. Both 🟡s and two of the 🟢s were fixed in `5a299f2c`; a fifth claim of the same class was found by the sweep and fixed with them.

```audit-claims round=1 audited=2b4d805e..5a299f2c
1. Claimed fixed: a ONE-SEGMENT handle suffix parsed from nix/agent-handles.nix wildcarded into any absolute parent and made the gate emit a wrong-FILE remedy; an `assert "/" in rel` in test_the_handle_table_is_parsed_from_the_nix_source is claimed to reject it with its own message before the collateral failures.
2. Claimed fixed: the module docstring's "MEASURED" KUBECONFIG transcript asserted a silent ~/.kube/config fallback that does not reproduce on this host; the block is claimed to now state the precondition (a NON-EMPTY current-context in the default kubeconfig) rather than the unconditional result.
3. Claimed fixed: CORPUS_DOC_FLOOR's comment and test_corpus_is_not_empty's failure message said "measured 88" for a corpus of 99; claimed re-derived via the gate's own _corpus().
4. Claimed fixed: the tilde census said 183 tokens "of which 13" name a $KC_* file when 3 of those 13 are ~/.kube/ tokens outside the 183, and a second figure of 184 for the same quantity sat at :521; claimed restated once against a named ref, with the duplicate deleted.
5. Claimed fixed: the docstring's "21 occurrences across 9 files" reproduced under no framing; claimed replaced with a ref-named figure.
6. Claimed UNCHANGED and deliberately out of scope: the ~176 repo-handle `~/workspace/...` sites, and CORPUS_DOC_FLOOR's VALUE (40), whose comment was wrong rather than the floor.
```

### Verified independently of the implementing agent

- The one-segment repro: planting `NIXCFG = "${home}/nixos";` into `nix/agent-handles.nix` with no edit to the gate reddens it on 22 corpus sites with the remedy `$NIXCFG/...`, naming a different file. After the fix the new assertion fires first, on its own message (`assert '/' in 'nixos'`). Reverted clean.
- The kubeconfig arm is scoped as intended: a planted `KUBECONFIG=~/workspace/homelab-talos/production-kubeconfig` goes red naming `$KC_PROD`, while a planted repo-handle `~/workspace/devrc/...` stays green — the deferred half is genuinely untouched.
- The `current-context: ""` finding was reached independently by the round-1 auditor and by me before reading its report.

### Reported, not fixed — outside this PR

`scripts/tests/test_doc_path_rot.py` carries the same stale-census defect twice over the identical corpus: `CORPUS_DOC_FLOOR = 40  # measured 80` against a real 99, and `REFERENCE_FLOOR = 155  # measured 310` against a real 593, each restated in a failure message (4 sites). The 310→593 gap is wide enough to be worth an owner.

ZacxDev:
## Round 2 — delta re-audit + fixes, and the ladder STOPS here

Delta re-audit over `2b4d805e..5a299f2c`: **🔴0 · 🟡2 · 🟢1**, all six round-1 claims verified against the code (five *actually fixed*, one *partially* — that partial is 🟡1 below). Sandbox tier built directly (`nix build …#checks.pytests`): `RESULT: PASS`, `scripts/tests collected=14645 passed=14645 skipped=0`.

```audit-claims round=2 audited=5a299f2c..7976af6c
1. Claimed fixed: the retracted count "21 occurrences across 9 files" survived 12 lines below its own retraction as "these 21 sites"; the census is claimed to now carry the absolute/tilde split (30/14 absolute + 9/4 tilde raw; 28/13 + 9/4 net, at 22ddd8dc) and that sentence to refer to it in words with no figure of its own.
2. Claimed fixed: the `leaked` ledger's stated consequence ("a grown table would arm ~ for EVERY ~/workspace/... site and make the gate unsatisfiable on Read-tool pointers") was measurably false; it is claimed replaced by a measured statement naming the KUBECONFIG= ANCHOR as what arms ~ broadly, with the ledger retained as a scope pin.
3. Claimed fixed: the one-segment-handle message offered "or exclude it here and say why", a remedy with no implementation; the clause is claimed dropped and the message claimed to state why exclusion needs a code change in _handle_table.
4. Claimed swept: every count, ref and stated CONSEQUENCE in the module re-derived or established by planting — 13 counts re-derived, 23 consequences established, and SIX further false consequence claims found and fixed beyond the three above.
5. Claimed UNCHANGED and deliberately out of scope: the ~176 repo-handle sites; test_doc_path_rot.py's own stale censuses; CORPUS_DOC_FLOOR's value.
```

### Verified independently of the implementing agent

- Planting two repo handles into the `kubeconfigs` block turns **exactly 2** tests red — `test_a_tilde_kubeconfig_outside_a_shell_assignment_stays_green[repo-handle-tilde]` and `[sibling-file]`. This **refutes the instruction I gave** (that the probe stays green) and confirms the sentence the fix round actually wrote. That sentence is on its third draft; the first two were false.
- `these 21 sites` and `or exclude it here and say why` are both gone from the module; `THE HAZARD IS THE ANCHOR, NOT THE TABLE` is present.
- 48 passed, both tiers. No scratch file committed; the range touches exactly one file.

### 🔴 Why the ladder stops here rather than running a round 3

Stated explicitly, because a report that ended on the prose escape hatch is otherwise **indistinguishable from one that converged**.

The normal rule is that a round returning findings is followed by another. Round 2 returned findings, so strictly round 3 follows. It is not being run because **this PR's payload is now prose inside a test module**: the attribution gate is structurally inert (every round touches the payload by construction), and "fixed a defect" and "reworded a warning" are the same edit. Left alone this does not terminate — round 2's two 🟡s were both sentences round 1's fix had written, 🟡2 was already the *second* draft of one sentence, and the round-3 fix round's own first replacement for a boundary claim was **also wrong**, caught by its own sweep.

The escape hatch's three conditions hold: **no 🔴**; **blast radius bounded to "the document contains a false sentence"** — none of the findings can change the gate's verdict on any input, and both tiers are green; and **the recurring shape swept at every site** rather than at the ones reported — 13 counts re-derived, 23 consequences established by planting, 6 further false claims found that way.

### Recorded as OPEN, not absent

- **The sentences this last round wrote have not been read by an adversarial round.** That is the accepted cost of stopping here, and given that this round's own first replacement was wrong, the residual error rate on new prose is non-zero rather than assumed zero.
- The **~176 repo-handle `~/workspace/…` sites** — deferred by an explicit scope decision, to be filed as its own ranked item.
- **`scripts/tests/test_doc_path_rot.py`** carries the same stale-census defect twice over the identical corpus: `CORPUS_DOC_FLOOR = 40  # measured 80` against a real **99**, and `REFERENCE_FLOOR = 155  # measured 310` against a real **593**, each restated in a failure message — 4 sites, in a file this PR does not own.
- **The handle table has two hand-maintained copies, both drifted from `nix/agent-handles.nix` today**: `scripts/claude-hooks/shell-env-nudge.py` is missing `KC_PROD`, and `scripts/lib/handoff_index.py`'s `REPO_ENV_HANDLES` is missing `CIVITAI_CLI`. The `KC_PROD` gap is live but minor; the `CIVITAI_CLI` one is latent — that checkout exists but holds 0 handoff docs, so it has no victim today.

PR #1625, opened 2026-09-13
docs(handoff): arc closed — and rank 13's probe measures the triage bot wrong 2 of 4

ZacxDev:
Close-out for `claudedocs/handoff-gate-speed-and-ci-signal.md`.

**All four inherited PRs are merged and shipped**, both hosts converged and verified at `14daa42a`:
`homelab-infra#799` `0b14768a` · `#1600` `f99d3c1b` · `#1613` `264de70d` · `#1603` `14daa42a`.

**`#1603` was gated on a merged-tree run twice.** The first came back clean (3665 passed / 0 failed, 24 files) and `main` moved during it, gaining a commit touching `scripts/lib/transcript_search.py` — a declared shared surface. The rule is that the trigger is *the base moved*, not *we touched the same file*, so it was re-run rather than reasoned past. The second held.

**The deliverable is consumer-verified on `main`, not merely merged:** `ledger-check.sh` derives 244 nodeids and reports `392 passed` / `RESULT: PASS (exit=0)`.
⚠ **It takes 229s, not the 166s the doc has carried since rank 11 was written** — corrected in place, with the reason (different tree and load, and a loaded box is the real usage condition).

**The arc now has exactly one actionable open item: rank 13.** `#1600`'s triage bot produced a confirmed false `INHERITED` on its first live sweep, which its own unit comment calls disqualifying for arming. The bot is inert, so this is a decision rather than an incident. A note was added that `#1603`'s AST derivation is a ready-made oracle for the exact predicate rank 13's likeliest fix needs — the two were built independently in one session and did not know about each other.

https://claude.ai/code/session_01298sLVSsAeZJpGNJTftdWv


ZacxDev:
Second commit (`a29901d7`) — **rank 13's probe ran, and it turns one anecdote into a measured rate with a named mechanism.**

The bot's `INHERITED — likely cured by rebase` is a falsifiable claim, so I tested it instead of reasoning about it: for each of the five verdicts, build the merged tree (PR head + current `origin/main`) and run **only the named failing test**.

| PR | failing test | merged tree | verdict |
|---|---|---|---|
| #1450 | `test_a_partial_run_is_declared_…` | 1 passed | bot RIGHT |
| #1286 | `test_agent_without_any_tab_is_untouched` | 1 passed | bot RIGHT |
| #1603 | shebang census guard | 1 failed | 🔴 FALSE |
| #1194 | shebang census guard | 1 failed | 🔴 FALSE |
| #1038 | — | merge conflict | UNTESTABLE |

**2 of 4 testable verdicts are false (50%).** Both false ones are the same repo-wide census guard; both correct ones are ordinary unit tests. The predicate proposed when there was only one data point — *is the failing test a census guard over files it does not name?* — **separates this sample 4/4**, and it earned that by predicting the two TRUE positives too, which is the half that could have falsified it.

The mechanism is sharper than "census guard": **in both false cases the offender is a NEW FILE the PR itself adds.** #1194's failure names `scripts/tests/test_break_glass_merge.py` — one of #1194's own files, confirmed absent from `main`, so a rebase would carry the offending file along with the red. The guard file is byte-identical in the branch *because the PR never touched the guard*.

**The fix is identified and its oracle is already on `main`:** check whether the failing test is in `scripts/testlib/census_scan.py::census_nodeids()` — the AST derivation #1603 merged as `14daa42a` — and demote to NOT EXPLAINED if so. The two were built independently in one session; the probe is what connected them. Rank 13 is now an implementation task with a self-writing regression test, not an investigation.

⚠ #1038 is excluded from the rate rather than counted — its merged tree does not build. A denominator of 4, stated, beats a denominator of 5 that hides one.

PR #1637, opened 2026-09-13
fix(handoff-search): a blank --exclude-slug filtered NOTHING and exited 0 — refuse it at the input

ZacxDev:
## The defect, reproduced live

```
$ python3 scripts/lib/handoff_search.py --offline --query "handoff search index" \
      --limit 3 --exclude-slug "  "
indexed_docs=418 indexed_sections=5861 backend=memory in_scope_docs=418 in_scope_sections=5861
query='handoff search index'  excluded=
…
rc=0
```

`excluded=` with nothing after it, `in_scope_docs` **equal to** `indexed_docs`,
exit **0**, and the document the caller meant to drop right there in the hits.
`"  "` strips to `""`, `""` matches no row, and a filter that silently declines
to filter is indistinguishable from one that worked.

This is the last reachable spelling of the class three earlier audit rounds
closed everywhere else in this module (an absolute path, a `handoff-x` basename
with no `.md`, a trailing space). Those each had a slug to derive and were fixed
by normalising harder. This one has none — so normalisation cannot repair it and
the only correct answer is to refuse the argument.

## The fix

An input rejection at `main()`, beside the existing `--limit` and no-repos rc-2
paths and phrased like them. It keys on **the derived slug, not the spelling**:

```
unusable = tuple(v for v in args.exclude_slug if not exclusion_slug(v))
```

`claudedocs`, `claudedocs/`, `/` and `handoff-.md` are not blank and reach the
identical empty slug by a different road, so a guard keyed on whitespace would
be walkable by a different spelling of the same state.

```
$ … --exclude-slug "  "
handoff-search: --exclude-slug takes a slug or a doc path, and got a value that
names NO slug — blank, or nothing beyond a `claudedocs/` prefix ('  '); it
normalises to the EMPTY slug, which excludes NOTHING while printing a confident
`excluded=` line and handing back the very document you meant to drop.
rc=2
```

**Not** in a renderer. A `label or "(unnamed)"` there would put a falsy string
back on the decision path the earlier rounds cleared, and would make the run
*look* filtered while filtering nothing.

Also updated in the same commit: the module docstring's exit-code table (which
claimed rc 2 was only `bad --limit, no repos`) and the `--exclude-slug` help
text. A comment is a claim too.

### Positive control — a real slug is untouched

```
$ … --exclude-slug handoff-handoff-search-index.md
indexed_docs=418 … in_scope_docs=417 in_scope_sections=5848
query='handoff search index'  excluded=handoff-search-index
rc=0
```

## Decision: the LIBRARY layer is left alone, and it is a documented hole

`_exclusion_list` in `handoff_index.py` still accepts `exclude=[""]`, which
filters nothing on both backends exactly as the CLI used to. **Deliberately not
changed**, and named in a comment beside the new guard so it is not a silent
omission:

- The closing condition was the CLI, and argparse cannot reach the library hole
  — `main()` is now the only door from argv into `run_search`, and it is shut.
- `_exclusion_list` is a shared coercion in front of **two** backends whose
  agreement is the thing that module exists to protect. Widening it is a
  behaviour change for every programmatic caller, and it deserves its own change
  with its own cross-backend evidence rather than riding along here.
- There is a live test (`test_the_refusal_does_NOT_offer_the_empty_string_as_the_fix`)
  whose whole argument is built on `[""]` being accepted; rewriting that premise
  in passing is how an audited invariant gets lost.

My recommendation is that it should eventually reject too — the hazard is
identical one layer down — but as a separate, scoped change.

## Tests — red at base, green at HEAD

New `TestABlankExclusionIsRefusedAtTheINPUT` in `scripts/tests/test_handoff_index.py`,
next to the class that closed the three earlier shapes.

Base sha **`30bad70c`** (this branch is rebased onto it).

```
nix develop <repo> -c python3 -m pytest scripts/tests/test_handoff_index.py \
    -q -p no:cacheprovider -k TestABlankExclusionIsRefusedAtTheINPUT
```

| | base `30bad70c` | HEAD |
|---|---|---|
| `test_a_value_that_derives_no_slug_is_a_usage_error` | **FAILED** (`'claudedocs' exited 0`) | passed |
| `test_the_refusal_says_WHY_rather_than_dumping_usage` | **FAILED** (`assert 0 == 2`) | passed |
| `test_a_blank_among_GOOD_values_is_still_refused` | **FAILED** (`assert 0 == 2`) | passed |
| `test_a_REAL_slug_is_unaffected_END_TO_END` | passed | passed |

3 failed / 1 passed at base; 4 passed at HEAD. The fourth is the **negative
control** — a guard that refused every `--exclude-slug` value would satisfy the
other three — so it passes at both by design and is *not* counted as regression
coverage.

At base the run also prints the leak itself: `excluded=widget-relay,` with a
trailing comma, and `⚠ 2 slug(s) were EXCLUDED (widget-relay, )`.

### Mutation check — 5 mutants, all killed, control green

Run under `PYTHONDONTWRITEBYTECODE=1` with `__pycache__` deleted between
mutants, each applied to a pristine copy (the harness aborts if a mutant changes
no bytes).

| mutant | killed by | the assertion that fired |
|---|---|---|
| *(control: no change)* | — | 4 passed |
| predicate → `not v.strip()` (spelled, not structural) | `…usage_error` | `'claudedocs' exited 0, so a scripted caller reads it as an answer` / `assert 0 == 2` |
| `return 2` → `return 0` | 3 tests | `'' exited 0…` / `assert 0 == 2` |
| guard deleted (`if False`) — the positive control | 3 tests | `assert 0 == 2` |
| refuse only when EVERY value is unusable | `…blank_among_GOOD_values…` | `assert 0 == 2` |
| refuse EVERY value (deletes the feature) | `…blank_among_GOOD…` + the negative control | `assert 2 == 0`, and the refusal naming `'widget-relay'` |

Every kill is on **this** guard's own observables — its rc 2 and its own message
text — not on an earlier check or another guard's error. Reachability: the
control run reaches the guard with a valid `--limit 3` and a resolvable
`--offline-repo`, i.e. no earlier check rejects that input.

## Scope of the test runs

- `scripts/tests/test_handoff_index.py` + `test_resume_handoff_search_wiring.py`
  + `test_runtime_shebangs.py` (the three files that reference `handoff_search`
  besides the module itself): **335 of 335 passed**.
- Earlier round, before the docstring/help edits:
  `test_handoff_index.py` + `test_resume_handoff_search_wiring.py` 326/326, and
  `test_runtime_shebangs.py` + `test_handoff_doc.py` + `test_handoff_audit.py`
  422/422.
- These are three- and two-file denominators against a repo of ~21,000 tests.
  They are not a gate verdict.
- `scripts/scoped-tests.sh` **refuses** this diff — `scripts/lib/**` is a shared
  surface — which is the correct behaviour, not an error. The advisory Tekton
  checks are what run everything these did not.

## Behaviour change worth knowing

`/resume` step 4 passes `--exclude-slug` the value step 2 printed on its
`handoff:` line. If that line is ever empty, the session now gets rc 2 with an
actionable message instead of a silent no-op that spends its top hits on the doc
it just read. That is the intended direction. I did **not** edit
`claude/skills/resume/SKILL.md` to mention it — out of scope here, and skill
bodies carry byte ceilings.


ZacxDev:
## Audit fixes applied — `d3cd51fe`. And one retraction.

### 🔴 Retracting a false claim in the PR body above

The body says (line ~150):

> I did **not** edit `claude/skills/resume/SKILL.md` to mention it — out of scope here, and skill bodies carry byte ceilings.

**The byte-ceiling half of that is FALSE, and I am retracting it rather than quietly editing it away.** Measured: no byte-ceiling test names `resume`. The ceilinged set is `browser-bridge`, `clawgate`, `handoff`, `prune-skill`, `session-manager`, `RULES.md` and `handoff_doc` — and `scripts/tests/test_resume_handoff_search_wiring.py:277` says in as many words that *"`claude/skills/resume/SKILL.md` has no byte ceiling"*. I invoked a constraint that does not apply to the file I was declining to edit. The file is now edited.

### What changed

No behavioural change. The rc-2 guard is correct and no argv value escapes it; every fix below is a **claim-accuracy** fix — a sentence that claimed more coverage than the code delivers.

1. **`scripts/README.md`** — the `lib/handoff_search.py` row enumerated this CLI's rc-2 conditions by name and omitted the new one. Added. ⚠ `TestTheReadmeRowMatchesTheExitCodeLedger` derives its expectation from `EXIT_CODES`/`STATUSES`, which are status-keyed; rc 2 is a usage code in neither, so that test is structurally blind here and would not have caught it either way.
2. **`claude/skills/resume/SKILL.md`** — the "exits non-zero for the four that are not readings: 3 / 4 / 6 / 7" block now carries rc 2, flagged as a non-zero that is *not* a zero (so the "four", scoped to zeros, stays true). The catch-all `Non-blocking:` sentence is untouched — it already covered rc 2 behaviourally, which is why nothing was operationally broken.
3. **"the last reachable spelling of the SILENT NO-OP"** — overreaching. **Three occurrences found, two fixed**: the `main()` comment in `scripts/lib/handoff_search.py` and the test class docstring. The third is in `8602595d`'s own commit message, which is immutable history — hence this comment. The claim that holds is *"the last spelling that derives the EMPTY slug"*. Measured counter-example: `resume-state.sh` prints `handoff: (none found — git-only)`, the resume skill tells the caller to pass that printed value, and it derives the **non-empty** slug `'(none found — git-only)'` — passes the guard, prints a confident `excluded=`, filters nothing. Benign (no doc was loaded, so there is nothing to exclude) and deliberately allowed, since `test_an_unmatched_slug_still_says_what_it_excluded` pins that an unknown slug must not filter. Both sites narrowed, the exception named in each.
4. **A dead assertion.** `test_a_REAL_slug_is_unaffected_END_TO_END` asserted `"relayrepo/widget-relay" not in out` — copy-paste from the sibling class, whose fixture repo *is* `relayrepo`. Ours is `blankrepo`, so that string could never appear and the assertion could not fail. Replaced with the check that was missing: the doc **is** returned *before* exclusion, giving the later `"── blankrepo/widget-relay" not in out` a baseline. The `readouterr()` call is kept — it clears the buffer.
5. **Two narrow labels.** The `UNUSABLE` comment said "The last two are NOT whitespace"; there are three (`claudedocs`, `claudedocs/`, `/`). And `TestABlankExclusionIsRefusedAtTheINPUT` understated its own ledger, so it is renamed `TestAValueThatDerivesNoSlugIsRefusedAtTheINPUT` — it had exactly one in-tree reference (its own definition), confirmed by grep.
6. **Help text + error message.** Both glossed the refusal as "blank, or nothing beyond a `claudedocs/` prefix", while `/`, `handoff-.md` and `handoff-` are refused and match neither clause. Generalised to "a bare prefix/affix with no topic left", naming the shapes. The governing sentence and the `repr()` of the offending value are unchanged.

### Verification

`scripts/tests/test_handoff_index.py` + `scripts/tests/test_resume_handoff_search_wiring.py` — **326 passed** (flake devShell, pytest 9.1.1, `-p no:cacheprovider`). That is the scope of the count; it is not a claim about the suite.

Behaviour re-confirmed unmoved, exit codes read from `$?` and never through a pipe: `--exclude-slug` of `"  "`, `""` and `"claudedocs"` each **exit 2** with empty stdout; `--exclude-slug handoff-handoff-search-index.md` **exits 0** and still narrows `in_scope_docs=418` against `indexed_docs=419`.

### Not fixed, on purpose

The library layer (`_exclusion_list` in `handoff_index.py`) still accepts `exclude=[""]`. Unchanged and still named in code beside the guard — out of scope, and widening a shared two-backend coercion is not what this PR is.

PR #1646, opened 2026-09-13
feat(handoff): rules (m) and (n) — an arc must name what ENDS it, and its self-generated rank queue may not GROW

ZacxDev:
## What this is for

`claudedocs/audit-arc-rabbit-holes-2026-09-13.md` in **homelab-talos** (`841cf63b3`) read 75 days of session telemetry — **745 doc-linked kickoff sessions across 299 handoff ARCS** — and deep-read the five longest.

In **all five**, the round-1 objective was met within **1–7 rounds**. The arcs ran **13–23**.

They did not stop shipping: commits/session stayed flat at ~7 and tokens/commit flat at ~80–86k all the way to round 7+. What changed is *what* they shipped — audit-fixes to audit-fixes. Two mechanisms carried it:

1. **No document said what would END the arc.** "Is this done?" had no object, so every round answered it from the ranked list.
2. **The ranked list grows.** Each round's audits minted 2–6 new items — `0→76`, `9→84`, `1→55`, `11→54` in the worst arcs — faster than rounds closed them. One doc recorded its own state as *"54 items… Of the 29 live items, only nine carry a forcing function."*

Of the **185** close-checks that landed on an arc session, **21 (11%)** closed one; median close-check → re-kickoff **1.0 h**.

## What lands

**Rule (m) — `status=undefined-done` (11).** `## Goal` carries `closing-condition: check|judgement — <the thing itself>`. A NAMED FIELD with a CLOSED vocabulary, taken from `claude/RULES.md`'s own object-leak split (a mechanical check, or a named human reading named evidence — never "someone will decide"). **REFUSED** on a NEW doc that lacks one, and on an update that DELETES one the doc had. **ADVISORY** on a legacy doc: measured with the parser itself, 0 of 119 devrc handoff docs and 0 of 64 homelab-talos ones comply, and refusing them all would be red-by-construction from run one.

**Rule (n) — `status=rank-growth` (12).** An anti-regrowth ratchet on the queue's SELF-GENERATED half: an update may not carry more items answering to nothing external than the document does. **Items with an EXTERNAL forcing kind are not counted at all** — this is not a queue-length limit. `--rank-growth-approved` is the operator opt-in. ⚠ **This is NARROWER than the report asked for** — it asks for *all* ranks to be opt-in and names `forcing: user` items as part of the problem. Operator decision 2026-09-13: keep the exemption; blocking work you asked for is the wrong failure mode, and the tag is unverifiable either way. Recorded in write-gate §G with its two costs — the ratchet binds only an honest author, and it does **not** address the `forcing: user` finding, which is a *drain* problem this *additions* ratchet cannot touch. the default route for an audit finding is the new `## Defects (batched)` section, which drains.

**The `DOD` block in `resume-state.sh`.** Rule (m) makes the field *exist*; nothing about a field existing stops round 14. This block puts the finish line on screen in the digest every `/resume` reads **before** it reads the doc, with a different instruction per kind — and `/resume` step 5 must now open its report with an **ADDRESSED / NOT ADDRESSED / UNMEASURABLE** verdict instead of an inventory.

## What was proposed and NOT taken

The report's §8 carries **five** proposed fixes; the operator's ask named three of them. Fix 5 (kickoff template anchors the round-1 goal) is outside that ask and is **not** implemented here. The third fix was *"max 3 delta rounds per PR, then ship-with-known-findings"*. That contradicts `claude/RULES.md`'s audit-fix-resets-gate rule in bold — *"A CLEAN round ENDS the ladder… **Not a cap** — the count is set by FINDINGS, never by a number."* Both cite the same evidence.

**Operator decision 2026-09-13: disclosure, not a cap.** `RULES.md` is unchanged. The ladder's rounds are RECORDED in `## State now` so a long ladder is visible and costed, and its findings may no longer mint ranks (that is rule (n)). Nothing forbids round 4 when round 3 found a real defect.

Fix 4 (close-check must end in a verdict) is not a separate mechanism — it is what rule (m)'s field gives the close-check something to be a verdict *about*, so the two ship together.

## `MAX_BYTES` 27,000 → 28,700 (+1,700)

Lever 3, with 1 and 2 spent first: every remedy is printed **by the tool**, and the evidence went WHOLE to `reference/write-gate.md` §F and §G. What could not be evicted, per the ledger — the template's field and its two kinds (the executor writes its scratch file before step 5 exists to refuse anything), the rank-ratchet clause (its fix is a DIFFERENT section the author must have seen), and the two status names (`test_every_exit_code_the_module_can_return_is_documented` derives them). **381 B free above the floor.** Round 0 deleted the `Audit ladder:` template line (91 B) and the ceiling was ratcheted DOWN by 100 rather than banking that as silent headroom. Round 0 also re-derived the arithmetic independently and found the raise **not avoidable by eviction** — cutting that line still leaves a ~1,319 B need.

## Three defects this round found in its own work

Each is recorded where it bit, with a mutation row.

- **`is_declared` did not require the field to be under `## Goal`.** A well-formed field under `## State now` satisfied the writer while `resume-state.sh`, which parses the Goal section, could not see it — `RULES.md`'s spelled-guard shape exactly.
- **The python pattern could not parse the TEMPLATE'S OWN SPELLING** (`- **closing-condition:** \`check\` — …`): `_MARKUP\s*` admits one markup run and the template writes two. The shell parser accepted it. Caught by the cross-language seam class and by nothing else — both parsers were green in isolation.
- **The DOD block raised a `!` gap for a missing field.** It fired on every run (0 of 183 docs comply) and took **49 tests red** — a gap on every document turns the `!! GAPS` banner into furniture. Also the wrong channel by the script's own rule: the `CLAWGATE` block already treats "no field" as nothing having been asked.

## Verified, not inferred

- **Mutation battery** — **19 new rows, 0 failures** over rules (m) and (n) in `mutants-handoff-cap.sh`, each naming the test that must kill it, plus the forced positive control and two behaviour-free `SURVIVES` controls. The battery gains a `MUTANT_FILTER` so a block can be re-derived without a four-hour run; **the two controls are forced through any filter**, and a filtered run prints `🔴 PARTIAL SWEEP` with the skipped count.
- **The 49-test regression was DIAGNOSED by a control run at `origin/main` (200 passed)**, not by reasoning about it.
- **End-to-end against a REAL handoff** (`homelab-talos claudedocs/handoff-arc-rabbit-hole-drift.md`): the legacy advisory renders, a 4th `forcing: none` rank is refused with `status=rank-growth`, and the DOD block prints its 🔴 line. Neither probe wrote anything.
- **New coverage**: 28 tests across two rule classes in `test_handoff_doc.py`, plus a new `test_resume_state_dod.py` whose `TestTheTwoParsersAgree` runs ONE 15-case matrix through both the python and the awk parser — with a positive control that the shell parser can be shown to ANSWER, since half its assertions are "the shell says nothing".

## Round 0 ran, before the merge decision

`round 0 · requirements: 9 (unattributed: 2) · deletion candidates: 3` · trial pair: **ran: yes · changed the outcome: yes**. Verdict: *requirement questioned* + *deletion candidate* — **not** *close, do not audit*; the premise holds (the arcs measurably do not terminate, nothing else declares a finish line, and rule (m)'s REFUSE arm is reachable ~twice a day at devrc's measured rate of ~2 new handoff docs/day).

What it changed: the dead cross-repo path that was failing CI, three false counts (the close-check denominator, 24→19 mutation rows, four→five fixes in §8), a broken `MUTANT_FILTER` example that selected 4 of 19 rows, and the deletion of the `Audit ladder:` template line. What it questioned and the operator kept: rule (n)'s external-kind exemption, `## Defects (batched)` (a convention, not a mechanism — worth an adoption check in a month), and `MUTANT_FILTER` itself.

One item left for `/simplify` or `/code-review`, not fixed here: `self_generated_report` (pre-existing, `kind == "none"`) and the new `is_self_generated` (`kind not in EXTERNAL_FORCING_KINDS`) are two predicates under one name in one module. They cannot currently print in the same run, so it is a maintenance hazard rather than a live defect.

## CI status — one failure was mine and is fixed; two are `main`'s

CI ran the full suite. **node: PASS** (1449/1449). **pytest: 23,285 passed, 2 failed** — down from 3.

- **Mine, FIXED** — `test_no_new_dead_paths`. Two devrc docs cited `claudedocs/audit-arc-rabbit-holes-2026-09-13.md` as a bare path; that doc lives in homelab-talos, so the scanner resolved it locally and found nothing. Now written `<homelab-talos>/…`, the form `doc-path-ignore.list` itself prescribes, at all **six** sites rather than the two the gate flagged.
- **NOT mine — `main` is red** — the other two. Both are in **one file this PR never touches**, `claudedocs/handoff-index-store-claims-accuracy.md`:
  - `test_no_handoff_doc_exceeds_its_budget` — 72,598 B, over the 65,536 B ceiling by 7,062 B
  - `TestByteIdentityVerifier::test_NO_TRACKED_FILE_ASSERTS_the_RETRACTED_two_entry_boundary`

  **Measured, not inferred:** a detached worktree at `origin/main` fails **the same two tests, on the same file, with the same 7,062 B overage** — re-checked at `3a3bede7` and again at `a2c84a1c`.

  ⚠ **#1641 has since MERGED and did NOT fix it** — the doc is still 72,598 B on current `main`, and **no open PR touches it any more**. This red is now unowned. Closing condition if anyone picks it up: `test_no_handoff_doc_exceeds_its_budget` goes green.

⚠ **CORRECTION to an earlier version of this section, which implied a gate.** `devrc/CLAUDE.md:174`: *"NOTHING BLOCKS A MERGE TODAY — `main` is protected in NAME ONLY… CI is advisory; read it."* `gh pr view` confirms `mergeable=MERGEABLE state=UNSTABLE`. So there is no gate to merge "through": this PR is mergeable whenever you want it, and the red is information, not an obstacle. What `claude/RULES.md` does say still applies to the red itself — *"a permanently-red gate is worse than no gate; it trains everyone to click through"* — which is an argument for someone owning that document, not for this PR waiting on it.

I have not cut another session's live handoff doc — that is precisely the cross-arc scope creep this PR exists to curb.

## Adjacent in-flight work

**#1643** (*"measure WHY every audit ladder stopped — 63.4% of 191 state no reason at all"*) is on the same theme and is not a duplicate: it measures ladder stop-rationale, this changes what a handoff doc must declare. Worth reading the two together — if #1643's finding argues for a cap after all, that is a change to `claude/RULES.md`, which this PR deliberately does not make.

https://claude.ai/code/session_01NDxbcnzMeMRe665mUdrvJF


ZacxDev:
## Round 1 fixes landed — `4e90d28c`

Round 1 (blind, nine axes) returned 0 🔴 / 4 🟡 / 4 🟢, verdict *safe to merge*. Every 🟡 was fixed rather than deferred, because two of them land on the default proposal path with counts that are false about the document.

Three of those fixes created the next finding, and one of my own new tests was vacuous. Both are recorded in `4e90d28c`'s message rather than smoothed over.

**Regression matrix: 6 red at `HEAD^`, 12 green at `HEAD`.** Mutation: 20 rows ok, plus the two rows this round's edits invalidated, repointed and re-killed by their own named tests.

```audit-claims round=2 audited=22892179..4e90d28c
1. F1 — rule (n) and rule (m) no longer judge a base this checkout does not hold: both now test `base_readable` directly instead of inheriting it from sitting below rule (h), whose refusal is gated on `--confirm`.
2. F2 — rule (n) skips when the base carries no canonical `## Next steps`, so migrating a reworded queue onto that heading is no longer counted as growth.
3. F3 — an emptied or absent working copy of a doc still tracked at HEAD is no longer treated as a new arc; `doc_tracked_at_head` asks HEAD, and keeps `None` distinct from `False`.
4. F4 — #1648's size-budget warning was moved below the refusals, so it no longer asserts "will go RED on main" and "not a refusal" on a run whose stderr says NOTHING WRITTEN.
5. F5 — the python and awk parsers now agree on all four inputs where they diverged (digit after the kind, markup run past three, NBSP after the colon, `~~~` inside a ``` fence); all four are pinned in the seam matrix.
6. F6 — the two hand-rolled round-4 rows now honour `MUTANT_FILTER`, and an accounting guard fails any sweep whose ROWS+SKIPPED does not equal the row sites counted from the file.
7. F7 — the legacy advisory no longer claims the field "costs a line"; 42 of 183 handoff docs have no heading that resolves to `goal`.
8. F8 — NOT fixed, recorded as an open limitation in write-gate §G: nothing counts, gates or reads `## Defects (batched)`, so rank growth can relocate there and nothing would measure it.
9. Both of rule (n)'s skips are DISCLOSED (`RULE (n) DID NOT RUN`, naming which reason) rather than silent, because a skip nobody sees is a pass.
10. `doc_tracked_at_head` uses `ls-tree`, not `cat-file -e`, which exits 128 for both "path absent" and "no HEAD".
```

ZacxDev:
## Round 2 fixes landed — `493b8b7f`

Round 2 (blind delta, `22892179..4e90d28c`): 0 🔴 / 5 🟡 / 5 🟢. Its sharpest finding was that **my previous round's headline fix was mutation-proven dead** — collapsing the `None`/`False` distinction survived a full green 472-test run, because the only guard on it fed an input `main()` refuses a screen earlier.

Ledger from that round: `round 2 · payload lines changed: 200 (since round 1: 200)`.

```audit-claims round=3 audited=4e90d28c..493b8b7f
1. The `None` arm of `doc_tracked_at_head` is now reachable from `main()` and covered: `git_could_not_answer` is its own predicate with its own skip reason and its own advisory wording, and three mutation rows kill it through an unborn-HEAD repo.
2. The legacy advisory no longer asserts a doc "was written before rule (m)" when git could not read the history — an unborn HEAD now says the history is unknown.
3. Rule (n)'s skip in that case no longer borrows the stale-base wording, which told the author to consult a count in a document that may not exist.
4. Two conjuncts added in the previous round were provably dead and are removed: `base_readable`'s `replaces_mainline_doc` operand, and rule (m)'s `base_readable and` operand. Neither changed any result.
5. awk `exit` became `next` at the digit-boundary guard, so a malformed field above a well-formed one no longer makes the reader show the NO-closing-condition banner for a doc the writer accepts. Seam divergences 5 -> 3.
6. write-gate §F's "22 docs" claim is replaced: it did not reproduce (38 wide / 28 narrow). What is stated now is the live count, 3, named individually, plus the corpus size 183.
7. write-gate §G no longer describes the relocation hazard as one heading wide: it is every heading `is_next_steps_heading` does not recognise, 28-38 documents, verified end to end.
8. The round-4 mutation gate names both rows it controls, so filtering by either name runs it.
9. RECORDED, NOT FIXED: a base carrying both a canonical and a reworded ranked queue is still counted from the canonical one alone, so a partial migration can still be refused for shrinking.
10. RECORDED, NOT FIXED: the NBSP fix admits NBSP in `trim` only and is locale-dependent; under LC_ALL=C the bracket matches bytes. `dod_row` sets no locale.
```

ZacxDev:
## Round 3 fixes landed — `968840b1`

Round 3 (blind delta, `4e90d28c..493b8b7f`): 0 🔴 / 4 🟡 / 4 🟢, ledger `payload 86 (since round 1: 286)`. Every 🟡 was a one-liner. Its sharpest observation was structural rather than any single defect: **each round's own fix has created the next one, three rounds running** — a retraction landing in one of two places, a predicate one notch too narrow, a payload change with no test, and mutation rows that duplicated each other instead of isolating the thing they named.

```audit-claims round=4 audited=493b8b7f..968840b1
1. The retracted "22 docs / 19 dated / 3 live" figure is gone from `handoff_doc.py`; the module now carries no population figure and points at write-gate §F, which holds the one that reproduces.
2. `history_unknown` is keyed on `tracked_at_head is not True` rather than `git_could_not_answer`, so a doc absent from HEAD — untracked, template-pasted, added-but-not-committed, copied into a fresh worktree — is no longer told it "was written before rule (m)".
3. That fix is covered by `test_an_UNTRACKED_doc_is_not_told_it_PREDATES_rule_m` plus a negative control asserting a genuinely tracked doc IS still told it predates the rule.
4. The `exit` -> `next` seam fix now has a test: `digit-boundary-first-then-good` in `TestTheTwoParsersAgree.CASES`, watched red at `exit` and green at `next`.
5. A new mutation row `legacy-advisory-loses-its-history-qualifier` isolates the advisory wording; it dies on a message assertion with a single killer, where the two prior rows were the same mutant dying on a returncode.
6. Three narrations that still said rule (n) has TWO silent skips now say three.
7. A comment quoting `tracked_at_head is not False` — an expression the previous round deleted — is rewritten, and grandfathering is called the PERMISSIVE direction in both places rather than "FAIL-CLOSED" in one.
8. write-gate §G carries §F's caveat with the 28-38 figure (all but 3 are unreachable behind `status=dated-topic`) and names the unbounded future case as the half that matters.
9. A fourth seam divergence class — two `closing-condition:` spellings on one line — is recorded as a residual so the list is not read as closed. Contrived, pre-existing, present under both `exit` and `next`.
10. RECORDED, NOT FIXED, unchanged from last round: the partial-migration count, the NBSP locale dependence, and `## Defects (batched)` remaining uncounted.
```

ZacxDev:
## Ladder closed at round 4 — `1841a00b`

Round 4 (blind delta, `493b8b7f..968840b1`): 0 🔴 / 4 🟡 / 4 🟢, all fixed. **I am stopping the ladder here, on `/audit-pr`'s prose escape hatch**, and naming the criterion rather than leaving it implicit — because a report that ENDED a ladder on that hatch is otherwise indistinguishable from one that converged.

### Why the rounds will not stop on their own

The **attribution gate is structurally inert** on this PR. It ends a ladder after two consecutive rounds of zero *payload* lines, and the payload here is `scripts/lib/handoff_doc.py` — a file where "fixed a defect" and "reworded a comment" are the same edit. Round 4 measured **56 payload lines of which exactly one was executable**; this round is the same shape. The gate will read non-zero forever while the rounds keep finding prose.

The other two conditions the hatch requires:

- **No 🔴 in four rounds.** Not one round found a defect that could produce a wrong *write*. Every refusal writes nothing, and the behavioural deltas of the last two rounds are single sentences with unchanged verdicts.
- **Blast radius: "the document contains a false sentence."** That is the honest ceiling on 7 of round 4's 8 findings and most of round 3's.

### The shape was swept, not spot-fixed

The recurring defect across all four rounds is one thing: **a claim that has gone stale against the tree.** Each round's own fix produced the next instance — a retraction landing in one of two files, a predicate one notch too narrow and then one notch too wide, mutation rows duplicating each other, self-reported numbers measured on a tree that no longer existed.

Rather than fix the reported instance again, this round made the class **mechanical** where it could be: `scripts/tests/test_mutants_handoff_cap.py` now checks every mutation row's expression against the live source on every run, with no filter — *the check that finds a dead row must not be subject to the filter that hid it*. That found a second dead row of mine, dead since round 3 and missed by **two successive adversarial audits**, because every sweep they ran was filtered and it was never selected.

Negative control watched: planting a dead expression turns the guard red and names the row. ⚠ The first attempt at that control **did not apply and passed** — recorded, because a control that silently fails to plant is indistinguishable from a guard that works.

The prose half stays human. There is no mechanism here for "this sentence is false", and inventing one is the failure this PR exists to curb.

### What is open — recorded so it reads as open rather than absent

None of these is fixed. All are in `write-gate.md` §F/§G.

1. A base carrying **both** a canonical and a reworded ranked queue is counted from the canonical one alone, so a **partial migration can still be refused for shrinking**.
2. The NBSP admission is in `trim` only, and is **locale-dependent** — under `LC_ALL=C` the bracket matches bytes. `dod_row` sets no locale.
3. `## Defects (batched)` is uncounted, ungated and unread: **growth can relocate there and nothing would measure it.** The honest follow-up is an `/adoption-scan` in a month, not another guard now.
4. An **update** whose queue sits under an unrecognised heading is invisible to rule (n) and raises no disclosure — inherited from rule (j)'s deliberate synonym gap, not introduced here.
5. `elision-omits-the-MAINLINE-source` measures nothing. **Pre-existing on `origin/main`**, now named in `KNOWN_DEAD` instead of silent; not repointed, because it belongs to rule (i) and not to this arc.

### Merge state

`devrc-pytests` is red for `claudedocs/handoff-index-store-claims-accuracy.md` — a file this PR never touches, red on `origin/main` independently, and concurrently edited by another session (its byte count moved between audit readings). devrc's `CLAUDE.md:174`: *"NOTHING BLOCKS A MERGE TODAY — `main` is protected in NAME ONLY… CI is advisory."* So this is mergeable whenever you want it; that red is information, not an obstacle, and not this PR's to clear.

PR #1650, opened 2026-09-13
docs(handoff): index-store doc was 7,062 B over its ceiling — evict what closed, demote what is dated, and drop the stale grandfather entry

ZacxDev:
`main` is RED on `test_no_handoff_doc_exceeds_its_budget` for two independent reasons. Both are fixed here.

| | before | after |
|---|---|---|
| `claudedocs/handoff-index-store-claims-accuracy.md` | **72,598 B** (over the 65,536 B ceiling by 7,062) | **59,735 B** — 5,801 B of headroom |
| `claudedocs/handoff-handoff-search-index.md` | 59,805 B, still carrying a GRANDFATHERED entry | 59,805 B, **entry deleted** |
| `claudedocs/refs/index-store-claims-accuracy.md` | — | **25,279 B**, new |

**No allowance was raised.** The gate's own playbook order held: step 1 (evict what closed) plus step 2 (demote dated evidence) were the whole answer; steps 3 and 4 were never reached.

## The ledger entry is DELETED, not re-pinned

`handoff-handoff-search-index.md` fell back under `MAX_BYTES` on its own, which is check (c) in the gate's docstring: *an entry whose doc now fits under MAX_BYTES → FAIL, delete the entry.* The ledger is a ratchet, not a permanent exemption — leaving the entry at 81,920 would have licensed 22 KB of regrowth nobody would see. A comment in `handoff_budget.py` records why the twelfth entry is gone.

## What closed — verified on `origin/main` BY CONTENT

Not by believing the doc's own status text, which is exactly the thing that goes stale.

| block | evidence |
|---|---|
| `cairn-cutover.py` P3 is blocked | `gh pr view 1428` MERGED, squash `13c0791a`; `RC_CUTOVER_COMPLETE` present in `scripts/cairn-cutover.py` |
| co-tenant sandbox flake + the REFUTED `gc --auto` theory | `#1453` MERGED, squash `eeea9025`; `maintenance` ×11 in `test_git_repo_isolation.py` |
| `seed.sh --help` "has ZERO test coverage" | false today — `test_subsystem_store_api.py` asserts `--allow-overwrite` is in its output |
| why devrc-ci went red | the block declares its own `Next probe: none needed`; `#1211` and `#1179` are both on `main` |
| ranked item 2 | `#1554` MERGED, squash `663bc86a`; already marked ✅ CLOSED |

## How the demotion was done

Every evicted block was cut into `claudedocs/refs/index-store-claims-accuracy.md` by **verbatim line-range slicing** — a python slice of the original lines, never retyped, never re-ordered within a block — so content survival is structural rather than something a reviewer has to trust. Each slice carries its source line range and the evidence that closed it.

Ten Gotchas bullets that appeared **twice** in the doc were sliced there too. Those are not evictions: only the second copy was removed, and the first is still in the doc and still indexed.

## Gap audit

Of **893** non-blank lines in the original, **892** are byte-identical in either the trimmed doc or the refs file. The single exception is `3. ` → `2. `, the ranked-list renumber after the closed item came out.

Structure was checked as well, because this doc's own Gotchas record that *a bullet-level survival check is structurally blind to a whole section vanishing — count the headings*:

```
H2      7 → 7      fences  4 → 4
H3     10 → 9      ranked  3 → 2
```

H3 drops by one because the two co-tenant-flake blocks (the observation and its REFUTED theory) collapse into a single closed pointer; ranked drops by one because the ✅ CLOSED item came out.

## Nothing open was deleted

Every open investigation, gotcha and ruled-out theory stays in the **indexed** doc. `handoff_index.SECTION_BOOST` ranks those highest for exactly that reason, and a `refs/` file is not indexed by `handoff_search` at all — which is why only DATED, CLOSED material went there.

Where a demoted block was **guarding** an instruction that stayed, the guard went with the instruction or left a cross-reference:

- the `gc --auto` refutation is restated beside the flake's closure, so nobody re-derives a theory that was correctly killed twice;
- the DASH-premise bullet now points at the join-key block above it, which still carries the open literal-TAB route;
- *"the dev-host tier has no known inherited red any more"* stays beside the `"INHERITED RED" IS NOT A REUSABLE VERDICT` bullet it qualifies.

The `RESOLVED — entries are still being WRITTEN` block was deliberately **not** demoted on two grounds: its root cause is still live (`subsystem_touch.DEFAULT_STORE_ROOT` still points at the frozen mirror on `main`), and it names a client scope that must not be propagated into a new file in a public repo.

## Verification

`scripts/tests/test_handoff_doc_size.py` — **9 passed** (was 1 failed / 8 passed).

Also green on every file that reads these paths: `test_handoff_doc.py`, `test_handoff_index.py`, `test_handoff_audit.py`, `test_resume_handoff_search_wiring.py`, `test_resume_state_handoff_resolution.py`, `test_no_captured_text.py`, `test_no_captured_markup.py` — **752 passed** across those seven plus the gate, and **603 passed** over the four re-run against the final tree. Dev-host tier only, base `d35c8655`; the `nix build` sandbox tier is CI's.


ZacxDev:
Pushed two commits here (fast-forward, no force) after confirming the branch tip was unmoved at \`783c8b3a\`:

1. **`fix(refs): mark the quoted needle in the new refs heading`** — this PR's new `claudedocs/refs/index-store-claims-accuracy.md` carried a **third**, unmarked occurrence of the retracted `"no create route"` phrase in a section heading. On the merged tree it was the *only* remaining finding, so merging as-is would have traded the doc-size red for a fresh red on `test_NO_TRACKED_FILE_ASSERTS_the_RETRACTED_two_entry_boundary`. Same one-marker remedy the needle's own comment prescribes.
2. **`origin/main` merged in** — the branch predated #1655 (`18a95e23`), which fixed the two occurrences in the core doc. That red you saw on this PR's CI was **inherited from `main`**, not caused by your diff.

### Verified on this branch
- Repo's **own** scanner (`_normalise_for_scan` + `_RETRACTION_MARKERS` + `_MARKER_WINDOW`), behind a positive control that still reports a bare assertion as a finding: **5/5 occurrences marked**.
- `test_handoff_doc_size.py` + `test_subsystem_store_api.py` in full: **769 passed**.
- Both gates green together, so `main` should be green on merge.

Nothing else in your diff was touched — the eviction, the `refs/` demotion and the grandfather-ledger drop are all as you wrote them.

PR #1651, opened 2026-09-13
docs(handoff): the arc is CLOSED — rank 1 shipped, and trimming this doc reddened main via the ratchet ledger

ZacxDev:
Close-out for `handoff-search-index`.

## What landed

**Rank 1 shipped** — #1637 merged (`67295ffba`): a blank `--exclude-slug` now exits 2, keyed on the *derived slug* rather than the spelling, so `claudedocs`, `/` and `handoff-.md` are caught too. Verified on `origin/main` by content with a negative control. Claim released. #1639 merged (`d35c86558`).

Both were merged through a red `devrc-pytests` with the operator's explicit go-ahead. The red was **proven inherited**, not assumed: `main` itself was red on the same two tests, and both branches were 7–8 commits behind the fixes.

## The lesson this PR exists to record

**Shrinking a doc can redden `main`.** The grandfather ledger is a *ratchet* — a doc listed in it that now FITS must have its entry deleted, and going under is a failure exactly like going over. A routine `/handoff` update cut this doc 67,076 → 59,805 B and the fleet-wide gate went red on a finding nobody had touched.

Also recorded: how to prove an inherited red in three commands; that I called two failures "mine" before checking and both were inherited; and that I confirmed a claim using the grep `CLAUDE.md` itself documents as incomplete — right answer, wrong method, caught by CI rather than by the verification.

## New rank 7 — a live root cause

`subsystem_touch.DEFAULT_STORE_ROOT` still points at the frozen mirror. Found while trimming a different doc, filed nowhere else; it is the mechanism that stranded five entries and 24 bullets on one machine.

## Note on the ledger

The proposal run advised deleting this doc's `GRANDFATHERED` entry in the same commit. **#1650 already does that** — duplicating it here would make the two PRs conflict, so #1650 owns it. This doc lands at 60,364 B, under the ceiling.


ZacxDev:
Merged `origin/main` into this branch (fast-forward push, no force) after confirming the tip was unmoved at `5dca8af3`. **No change to your diff** — this only refreshes the base.

### Your red check was inherited, not yours

`tekton/devrc-pytests` was failing on `TestByteIdentityVerifier::test_NO_TRACKED_FILE_ASSERTS_the_RETRACTED_two_entry_boundary`, which **your diff cannot reach** — you touch only `claudedocs/handoff-handoff-search-index.md`.

Verified rather than assumed: that test was **already red on `origin/main`**, and `18a95e23` (#1655, the fix) was **not** in this branch's history — it was 5 commits behind. With `main` merged in, the gate and its two controls pass here.

### ⚠ One thing this PR does not do

It does **not** green `test_no_handoff_doc_exceeds_its_budget`. It shrinks this doc to **60,364 B** — under the 65,536 B ceiling — but the gate then reports the other half:

> `handoff-handoff-search-index.md`: 60,364 B now fits under the ceiling — **DELETE its GRANDFATHERED entry** (the ledger is a ratchet, not a permanent exemption)

Deleting that entry is part of **#1650**, so the doc-size gate stays red until that one lands. I test-merged `main` + this + #1650: **clean, no conflicts, both gates green**, so merge order does not matter.

PR #1657, opened 2026-09-14
fix(shell-env-nudge): KC_PROD was missing from the hook handle table, unpinned for six weeks

ZacxDev:
Closes the live half of rank 26 of `claudedocs/handoff-cairn-oss-multi-instance.md`.

## The defect

`scripts/claude-hooks/shell-env-nudge.py` hand-maintains two dicts — `REPO_VARS` and `KC_VARS` — that restate `nix/agent-handles.nix`, the single source of truth every other consumer parses. **Nothing read them.** `KC_VARS`/`REPO_VARS` occur in that one file and nowhere else in the tree, so the copy had no ledger of any kind.

It had drifted: `KC_VARS` carried **4 of the 5** kubeconfig handles. `KC_PROD` was absent — and its kubeconfig exists on this host.

| | |
|---|---|
| hook landed | `15233e3c` (#50), 2026-07-02 — one commit, never touched again |
| `KC_PROD` declared in nix | `e21985a0` (#276), 2026-08-02 |
| gap | 2026-08-02 → 2026-09-14, **~6 weeks**, unreportable |

## Why it needed a gate, not just a fix

The failure is silent **by construction**. A handle absent from these tables is not an error: it is a `.get()` returning `None`, one `if var:` that does not fire, and a nudge that is never emitted. The hook keeps working and reports nothing. There is no symptom, no log line, and nothing to grep — so the only thing that can catch it is an assertion that the table **is** the source.

The hook's header also named `nix/programs/zsh/default.nix` as the thing it "must match". That file *imports* `agent-handles.nix` — it is a generated consumer exactly like this one. **Matching a copy against a copy is the mechanism**, and the header is corrected to name the real source and the ledger that now enforces it.

## What is in the change

- `KC_PROD` added to `KC_VARS`; header corrected.
- **`scripts/tests/test_shell_env_nudge_handles.py`** pins both dicts to the nix source in **both directions**, as whole `path -> handle` **mappings** rather than name sets — a table with the right names and drifted paths matches no command and nudges nobody, the same silent nothing. A one-way check ("every hook entry is real") stays green forever against exactly the state this ends.
- The nix parser is **imported** from `test_absolute_handle_paths`, not rewritten. A third regex over the file whose duplication is the defect under test would be the bug (RULES: one rule, one place).
- A **basename-shadowing** guard. `KC_BASENAMES` is `{basename(path): handle}`, so two kubeconfigs sharing a file name collapse and the loser becomes unreachable through the relative-path arm — it would nudge toward the **wrong cluster**, which is worse than no hint. This change adds a fifth kubeconfig, so it *enlarges* that hazard rather than leaving it where it was.
- `test_absolute_handle_paths`'s docstring said "unifying the hook with it is separate work". That is no longer outstanding; the paragraph now states what holds instead.

## Regression matrix

**RED** at `06287019` on `test_the_kubeconfig_table_is_exactly_the_nix_kubeconfigs_block`:

```
AssertionError: nix/agent-handles.nix declares kubeconfig handle(s) that
scripts/claude-hooks/shell-env-nudge.py's KC_VARS does not carry: ['KC_PROD'].
```

**GREEN** at HEAD (5 passed).

## Mutation battery — 7/7

Each mutant killed **by its intended test** and required to fail on **that test's own assertion message**; harness positive-control watched green on the pristine tree first; `PYTHONDONTWRITEBYTECODE=1`; both mutated files diffed byte-identical afterwards.

| # | mutant | killed by |
|---|---|---|
| M1 | `KC_PROD` dropped (the base defect) | kubeconfig equality |
| M2 | undeclared kubeconfig added | kubeconfig equality (`extra` arm) |
| M3 | repo handle dropped | repo equality |
| M4 | undeclared repo added | repo equality (`extra` arm) |
| M5 | correct path rebound to a **wrong handle name** | repo mapping equality |
| M6 | two kubeconfigs share a basename | basename shadow guard |
| M7 | the **nix source** loses `KC_PROD` | parse positive control |

- **M5 is the reachability proof** for the `got == want` mapping assertion. Both set differences are empty in that state, so without M5 that assertion would be a guard nobody had watched execute.
- **M6 is isolated**: nix and the hook are moved *together*, so basename uniqueness is the only property that changes. Mutating the hook alone also made the path undeclared, and the guard was then killed alongside a mutant it did not need to catch — green for the wrong reason.
- **M7 is the positive control.** Every assertion here is `hook_dict == parsed_nix`; a parse that silently yielded nothing would make them all vacuous and pass against *any* table, including the drifted one. The zero is refused first.

🔴 **The battery's first run was wrong, and the harness was the reason.** `grep -cF` with a multi-line pattern splits it one-pattern-per-line, so a trailing newline became an empty pattern matching every line: three valid mutants were scored INVALID at "150 / 38 occurrences". The count is done in Python now. Recorded because the instrument, not the mutation, was the thing that failed — and it failed *reassuringly*.

## Not done, deliberately

Rank 26 also called `handoff_index.REPO_ENV_HANDLES` drifted for omitting `CIVITAI_CLI`. **It is not drift.** That omission is a deliberate exclusion, already pinned in both directions with its reason recorded in source, by `test_handoff_index.py::TestTheUnitEnvironmentMatchesTheHandlesTheIndexerReads` — which passes today. Adding the handle would add a zero-doc repo to the corpus and narrow the hosts `--prune` can run from.

So the brief's "two hand-maintained copies, both drifted" is **one** drifted copy and one correctly-gated decision. The handoff will be corrected separately, after this merges.


ZacxDev:
## Round 0 (requirements & deletion) — verdict: `requirement questioned`, plus a guard of mine that could not fail

Dispatched blind against `57f93df0`. Not deploy-blocking, but three findings were about prose or guards **this PR introduced**, and one was a real vacuity. Each re-verified independently before acting — an auditor's self-report is a claim.

**Fixed in `fa2cb23b`:**

1. 🔴 **`test_the_nix_source_parses_to_a_usable_table`'s disjointness assertion was VACUOUS — it could not fail.** It asserted `not (repo_names & kc_names)`, but `_repos()` builds its half by *subtracting* the kubeconfig names, so that intersection is empty by set-difference identity whatever the nix file says. **Reproduced:** `SHARED` planted in **both** nix blocks with the hook carrying only the kubeconfig one — so `agent-handles.nix` declares a repo handle `REPO_VARS` does not carry, exactly the silent half `test_the_repo_table_is_exactly_the_nix_repos_block` calls undetectable — and the module reported **5 passed**. Now reads the raw whole-file parse, where a name declared twice comes back as two entries. Same planted tree: **RED**, on its own message.

2. 🟡 **A false sentence in a guard's sole justification.** The basename-shadow guard said the hazard was *"a live hazard rather than a hypothetical"*. Measured false — the five basenames are pairwise distinct, and were before this PR. It is a **forward** guard and now says so.

3. 🟡 **This ledger pins the DECLARATION, and a declaration is not an EXPORT.** Both consumers existence-guard (`exportIf "-d"` / `"-f"`), so a handle whose file is absent is declared and *not* exported. Measured on this host: `~/.kube/homelab-nebula.yaml` does not exist, `$KC_NEBULA` is **unset**, and the hook nudges `KUBECONFIG=$KC_NEBULA` anyway — precisely the failure this module's own `extra`-arm message describes. **Pre-existing**, but this ledger makes it mandatory, so it is now recorded in the module with the better design named (`{p: n for n in NAMES if (p := os.environ.get(n))}`) and a closing condition. Deriving from the nix file at import is *not* that fix — the deployed hook is a `home.file` store copy, so reaching the file needs `$DEVRC`, the very mechanism it teaches.

4. 🟢 **Deleted `test_the_hook_file_is_where_this_module_thinks_it_is`**, on measurement: the loader already raises `FileNotFoundError` naming the exact path.

**Re-verified after the fixes** (an audit fix resets the gate): new guard RED-then-GREEN on the planted tree; 52 passed across this module + `test_absolute_handle_paths`; mutation battery re-run **7/7** still killed by their intended test on that test's own message, harness control green first, both files byte-identical after.

**Round 0 independently confirmed the half this PR declined.** It ran `test_handoff_index.py::TestTheUnitEnvironmentMatchesTheHandlesTheIndexerReads` (1 passed) and agrees `CIVITAI_CLI`'s absence is a gated decision, not drift.

**Correction to this PR's body.** Its "Not done, deliberately" section said *"The handoff will be corrected separately, after this merges"* — an object with no closing condition and no owner, while the **ranked item** (the queue `claim-work` draws from) still told the next session both tables were drifted. That is fixed now, not deferred: #1659 amends rank 26 in place, including its closing condition, which was unmeetable as written.

`round 0 · requirements: 7 (unattributed: 2) · deletion candidates: 3`

ZacxDev:
## Round 1 (nine axes) — audited blind at `fa2cb23b`. Two findings fixed in `3b9bdbf2`; no 🔴.

Both re-verified independently before acting, and both watched RED before GREEN.

### 🟡 1 · payload · an absolute path was matched by BASENAME alone — naming the wrong cluster

`KC_BASENAMES` is `{basename(path): handle}` and the lookup consulted it for **any** path whose exact match missed, absolute ones included. An absolute path that merely *ends* in a known kubeconfig name is a different file, so the nudge named the wrong cluster — and since `$KC_*` is existence-guarded in nix, where that handle is unset the suggestion expands to **empty** and `KUBECONFIG= kubectl` silently takes the **default context**.

Measured, and the realistic case is the damaging one: `claude/skills/auditloop/SKILL.md` documents the laptop's kubeconfigs at `~/workspace/homelab-infra/{workbench,homelab,production}-kubeconfig`, while `$KC_PROD` is guarded on the `homelab-talos` spelling — so an absolute `.../homelab-infra/production-kubeconfig` was nudged as `$KC_PROD`.

**Pre-existing**, confirmed against the base copy (`/tmp/prod-kubeconfig` already yielded `KC_DPPROD` there). Adding `KC_PROD` completed it to three of three, which is why it is fixed rather than filed. The hook's own comment already said the map was for *"Relative kubeconfig references"* — the code was wider than its comment.

### 🟡 2 · scaffolding · `_repos()`'s derivation was unsound, and printed a WRONG remedy

It was "whole-file parse minus the kubeconfig names", justified as exact because the blocks are disjoint. Disjointness gives `repos ∩ kc = ∅`; exactness *also* needs `_handle_table() ⊆ repos ∪ kc`, which nothing asserted. Measured: a third block — `caches = { CACHE_X = "${home}/.cache/agent-x"; };` — was swept into the repo half, so the repo-equality test failed telling the reader to **add a never-exported, non-repo handle to `REPO_VARS`**. Now parses the `repos` block directly (symmetric with `_kubeconfig_table()`, reusing the imported `_NIX_ENTRY`), plus a partition assertion whose message says explicitly *not* to add it to a consumer dict to go green.

### 🟢 Correcting this PR's own body, publicly rather than silently

The body says *"GREEN at HEAD (5 passed)"* and carries a **7/7** mutation table. Round 0 deleted one test, so that count was already stale when written, and this round adds one — the module is now **6** tests. The substance was re-derived and holds; only the counts drifted. Read the commit messages for the current figures rather than the body's table.

### 🔴 Environment, not this PR: `main` is red on TWO gates

Both measured at `origin/main` with the discriminating control, not inferred:

- `test_runtime_shebangs.py::test_no_test_writes_a_usr_bin_env_shebang_at_runtime` — fails at merge-base `06287019`, at this PR's head, **and** at `origin/main`. Offender is `scripts/claude-hooks/tests/test_guard_core.py` (unpinned `#!/usr/bin/env bash` writes) from #1551. This PR touches none of it.
- `test_handoff_doc_size.py::test_no_handoff_doc_exceeds_its_budget` — fails identically at `origin/main` on `handoff-index-store-claims-accuracy.md` (7,091 B over) and `handoff-handoff-search-index.md` (stale ledger entry). Already claimed by another session.

So `devrc-pytests` will be red here for reasons that predate this branch. Flagging rather than merging quietly through it.

### Deployment gap (second-order, not a defect)

`nix/home.nix` ships this hook as a `home.file` /nix/store copy. Merging does **not** make the fix live — that needs a `home-manager switch`.

```audit-claims round=1 audited=fa2cb23b..3b9bdbf2
1. The KC_BASENAMES fallback no longer fires for absolute paths; relative refs and exact absolute matches still resolve.
2. _repos() parses the `repos` block directly instead of subtracting kubeconfig names from a whole-file parse.
3. A partition assertion fails when a handle is declared in neither the `repos` nor the `kubeconfigs` block.
4. A behavioural regression test covers the absolute-path case, carrying its own positive control.
```

ZacxDev:
## Round 2 (delta, `fa2cb23b..3b9bdbf2`) — all four round-1 claims verified fixed. Three 🟡, no 🔴. **Ladder STOPS here; rationale named below.**

| # | round-1 claim | disposition |
|---|---|---|
| 1 | basename fallback no longer fires for absolute paths; relative + exact-absolute still resolve | **actually fixed** (18-spelling differential; 5-mutant battery) |
| 2 | `_repos()` parses the `repos` block directly | **actually fixed** (7 nix shapes probed) |
| 3 | partition assertion fails on a handle in neither block | **actually fixed**, and not circular |
| 4 | behavioural regression test with its own positive control | **actually fixed**, non-vacuous |

### 🟡 1 — fixed in `28b711e0` · I restated a claim this repo had already RETRACTED, *in the file that retracts it*

My round-1 comment said a handle that fails to export "expands to EMPTY and `KUBECONFIG= kubectl` **silently falls back to the default context**". `scripts/tests/test_absolute_handle_paths.py` — which I was editing in the same commit — already carries a block retracting that sentence and saying in terms: *"do not repeat the fallback sentence as the mechanism, and do not quote the transcript without its precondition."*

Re-measured, kubectl v1.36.3, this host's `~/.kube/config` carrying `current-context: ""`:

```
KUBECONFIG=''                  -> error: current-context is not set   rc 1
KUBECONFIG='/absent/nope.yaml' -> error: current-context is not set   rc 1
```

Both arms **loud** and character-identical. The silent arm needs a non-empty default `current-context`; this host has none. Corrected at all four sites this PR introduced — hook comment, module docstring, the regression test's assertion **message** (which a maintainer would act on), and the `extra`-arm quote — plus the pre-existing site. Each now carries the precondition **and** names the case that needs none: where the wrongly-named handle *is* exported, the command runs against the **wrong cluster with no error at all**. That is the worse failure and my wording had buried it under the one that does not reproduce.

### 🟡 2 — fixed in `28b711e0` · my docstring said "the parser is IMPORTED, not re-written"; round 1 made that false

Round 1 added `_REPOS_BLOCK`, so three copies of one predicate existed over one file — and they **disagree**: on an inline `repos = { A = "..."; };` the local copy over-captures into the next block while `test_repo_path_guard`'s `split("repos = {")` copy silently drops an entry. `_NIX_SECTION` is now `nix_block(name, why)`, parameterised where it already lived, called by both consumers; the old constant is **removed** rather than left beside its successor.

Two limits named rather than implied, each with a closing condition in the docstring: `nix_block` inherits the inline-shape limit, and the **third copy in `test_repo_path_guard.py` is not consolidated here**.

### 🟡 3 — **FILED, not fixed — operator decision**

The relative arm keeps the same hazard: the hook is cwd-blind, so `KUBECONFIG=./production-kubeconfig` → `$KC_PROD`, newly reachable because this PR added `KC_PROD` to `KC_VARS`. Pre-existing in kind (`KC_HOMELAB`/`KC_WORKBENCH` already had it). The proper remedy threads `cwd` from the PostToolUse payload — `bash-guard.py`, `git-add-provenance-nudge.py` and `guard_core.py` all already read it — resolves relative paths, exact-matches `KC_VARS`, and makes `KC_BASENAMES` unnecessary.

Deliberately **not** done here: that changes `analyze()`'s signature and the core matching of a hook that fires on every Bash call, which deserves its own PR and its own audit rounds rather than being bolted onto a ladder already three rounds deep. Filed as a ranked item with a mechanical closing condition. Frequency is **unmeasured** — flagged, not resolved.

### 🔴 Why the ladder stops here, stated rather than left implicit

Required by the skill, because a report that ends on this escape hatch is otherwise indistinguishable from one that converged:

- **The payload is now prose, so the attribution gate cannot fire.** Measured: round 2's fixes changed 12 hook lines and **0 of them are executable** — all comments. "Fixed a defect" and "reworded a warning" are the same edit here, so the rounds do not terminate on their own.
- **No 🔴 in any round**, and no blast radius beyond "a comment contains a false sentence".
- **The recurring shape was swept at every site**, not just the reported one: all four new sites of the retracted claim plus the pre-existing one.
- **Recorded as open, not absent:** 🟡 3 (filed), `nix_block`'s inline-shape limit, and the un-consolidated third copy in `test_repo_path_guard.py`.

### Gates

`devrc-pytests` is red here for two reasons that predate this branch, both confirmed at `origin/main` with the discriminating control: `test_runtime_shebangs.py` (offender `scripts/claude-hooks/tests/test_guard_core.py`, from #1551) and `test_handoff_doc_size.py` (two other handoff docs; claimed by another session). Neither names a file this branch touches. Merging on that evidence, documented rather than quietly.

⚠ The full local `scripts/tests` run is **void as evidence** and is not quoted as a pass or a failure: the worktree was edited throughout its 89 minutes, so it is attributable to no commit. For the record, none of its 130 failures name either file this PR touches, and they cluster in `test_nogit_isolation` / `test_run_tests_jobs` / `test_scoped_ledgers` / `test_activity_spool_isolation` — four suites contending on one box.

```
round 2 · payload lines changed THIS round: 12 (0 executable) · since round 1: 17 · elapsed: ~57 min
```

```audit-claims round=2 audited=3b9bdbf2..28b711e0
1. The retracted "silently falls back to the default context" claim is removed from all five sites and each carries its precondition.
2. _NIX_SECTION is parameterised as nix_block(name, why); both consumers import it and the old constant is deleted.
3. nix_block's inline-shape limit and the un-consolidated third copy in test_repo_path_guard.py are named with closing conditions.
4. The relative-arm cwd hazard is FILED as a ranked item, not fixed, by operator decision.
```

PR #1665, opened 2026-09-14
fix(battery): main is RED — 6 mutation anchors went 0x when #1648 moved the ceiling's constants out of the suite

ZacxDev:
## The red

`test_every_mutation_anchor_occurs_exactly_once_in_its_target` is red on `origin/main` — **6 anchors at 0x**. This is the last of three reds `main` was carrying; the other two closed via #1655 and #1650.

🔴 **The severity is not "a test is red".** Per the battery's own docstring a 0x anchor prints `NOT-APPLIED` and **scores as a SURVIVOR while testing nothing**. So `main` has carried **six silently-inert mutants** since the cause landed — the ceiling's whole data layer (`MAX_BYTES`, the grandfather ledger, the quantiser) was being "mutation tested" by rows that never applied.

## Cause

`#1648` (`a2c84a1c`) moved `MAX_BYTES`, `GRANDFATHERED` and `tightest_allowance` out of `scripts/tests/test_handoff_doc_size.py` into `scripts/lib/handoff_budget.py`. The battery's `TARGETS` map still named the file they came from.

**A constant extraction has a third ledger: the mutation battery's `TARGETS`** — alongside the two this repo already documents (`test_no_public_ips.py`'s ALLOWLIST, `census_scan.py`/`ledger-check.sh`). It does not fail loudly at the battery; the tell is the separate anchors test.

## Fix

Added a `BUD` target; re-pointed **C1/C4/C6/C8/C11** at it. Split **per row, not per section** — the ceiling's DATA and arithmetic live in `BUD`, while the checker reading them and the scan wiring stay in `CAP`, and the section spans both files.

**C5 could not simply be re-pointed.** It anchored on the `handoff-handoff-search-index.md` grandfather entry, which `#1650` correctly DELETED once that doc was pruned under `MAX_BYTES`. A row anchored on a ledger entry is only as durable as that entry, and this ledger is *designed* to shrink. Re-anchored to `handoff-cairn-task-linkage.md` — 76,743 B quantises to 81,920, so 114,688 is exactly the two steps of slack the row's description names.

## Verification — the anchors test is only half the claim

An anchor at 1x proves the mutant is **APPLIED**, not that it **KILLS**; the failure message demands the other half. Full battery re-run (>25 min, detached):

```
CONTROL scripts/tests/test_handoff_doc_size.py:                9 passed
CONTROL scripts/tests/test_resume_state_handoff_resolution.py: 200 passed
21/21 killed for the named reason
```

Every row `KILLED(attributed)` — by the test its own row names, not a neighbouring guard — **including C5** and the **C11 positive control**.

- `test_mutation_battery_anchors.py`: **29 passed, 2 skipped** (was 28 passed / 1 failed).
- **Matrix: RED at `22ae0817`, GREEN at HEAD.**
- **Merged-tree gated** against `9254cded`: clean merge, anchors **29 passed**, `test_handoff_doc_size.py` **9 passed**.

## ⚠ Gotcha worth keeping

The battery takes **>10 min** and a timeout kill **skips its `finally` restore** — a foreground run here left `scripts/resume-state.sh` carrying row A6's mutant, i.e. tracked source silently modified. Its docstring's "check `git status` anyway if it dies hard" is not boilerplate. Run it detached.


ZacxDev:
Pushed `a7f69d8b` — two findings from a **round-0 audit** of this PR, both re-verified against the tree before I acted on them.

### 1. C4 was anchored on the ledger's *least*-headroom entry

A row anchoring on a ledger line binds to its **trailing measured comment** too — the mutation deletes the whole line, so the comment can't be trimmed out of the anchor. That comment is re-measured whenever the allowance is bumped, and an allowance is bumped when the doc runs out of headroom. So the right entry to anchor on is the one with the **most** headroom — and C4 had the **least of the eleven**.

Measured at `9254cded`:

| entry | size / allowance | headroom |
|---|---|---|
| **`handoff-tmux-webapp.md`** (C4's old anchor) | 326,750 / 327,680 | **930 B** |
| `handoff-nix-disk-cleanup.md` (new anchor) | 99,215 / 114,688 | 15,473 B |

It grew **314,233 → 323,642 → 326,750** across the two preceding commits on `main`, both of which touched it. The next `/handoff` append bumps the allowance, re-measures the comment, and takes C4 to **0x** — SURVIVED-while-testing-nothing, i.e. *the exact defect this PR exists to fix*, reproduced within days. Same mutation, same named killer; only the entry changed.

### 2. `BUD` was missing from the outer restore net

`saved` listed `SH`/`CAP`/`IDX`, so the target this PR adds was the one file the net could not restore. The per-row `finally` covers each mutant, but a run **killed between rows** skips it and only the outer snapshot can put the tree back. Measured while writing this change: a foreground run hit a 10-minute cap and the SIGTERM left `scripts/resume-state.sh` carrying row A6's mutant — tracked source, silently modified.

### Not taken, deliberately

Deriving C4/C5/C11's anchors from `handoff_budget.GRANDFATHERED` at import time. It would make the anchors test's 0x direction **near-vacuous** for those rows — comparing a string against the file it was generated from — and it's an operator-level redesign, not something to smuggle into this PR. Recording it as open rather than absent.

### ⚠ Verification state — do not read this push as merge-ready

- `test_mutation_battery_anchors.py`: **29 passed, 2 skipped**.
- 🔴 The **full battery re-run proving C4 still KILLS after the re-anchor is still in flight.** An anchor at 1x is **APPLIED**, not **KILLING** — different claims, and the whole point of this PR is that the weaker one reads like the stronger. I'll post the verdict when it lands.
- The earlier `21/21 killed for the named reason` was measured **before** this re-anchor, so it does not cover C4's new anchor.

ZacxDev:
✅ **The pending verification landed — C4 still KILLS after the re-anchor.**

Full battery re-run against the re-anchored table:

```
CONTROL scripts/tests/test_handoff_doc_size.py:                9 passed
CONTROL scripts/tests/test_resume_state_handoff_resolution.py: 200 passed
21/21 killed for the named reason
```

The three rows that bind to ledger data, all `KILLED(attributed)` — by the test each row names, not a neighbouring guard:

- `C4` — silently drop a document from the ledger ← **the re-anchored one**
- `C5` — loosen one allowance by two steps (slack entry)
- `C11` — POSITIVE CONTROL: a malformed ledger MUST die

This supersedes the earlier `21/21`, which was measured before the C4 re-anchor and therefore did not cover its new anchor. The tree restored cleanly this time (`git status` empty), so the `finally` was not skipped.

### Re-gated against a moved base

`main` advanced `9254cded → d4179fbd` while the battery ran, so the earlier merged-tree result was stale. Re-merged (clean) and re-run on the new tree:

- `test_mutation_battery_anchors.py` + `test_handoff_doc_size.py`: **38 passed, 2 skipped**
- C4's new anchor still occurs exactly **1x** in `handoff_budget.py` on the merged tree

### Status

Every claim in this PR is now verified, including the one I flagged as pending. Remaining open item is the one I declined on purpose — deriving C4/C5/C11's anchors from `GRANDFATHERED` at import time — which is an operator-level redesign, recorded above as open rather than absent.

